PR #1184, opened 2018-06-27
[unittests][graph] Add a test case for the uselist

qcolombet:
This currently fails with:
../tests/unittests/graphTest.cpp:77: Failure
Value of: conv->getFilter()->hasOneUse()
  Actual: false
Expected: true
../tests/unittests/graphTest.cpp:78: Failure
      Expected: conv->getFilter()->getNumUsers()
      Which is: 2
To be equal to: 1
[  FAILED  ] Graph.useList (1 ms)

Put it up for review just to illustrate the problem.

NFC.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -45,30 +45,43 @@ private:
   Node *node_{nullptr};
   /// Specifies the node result number to use.
   unsigned resNo_{0};
+  /// Is this value an operand of some instruction or just a temporary object
+  /// flying around.
+  bool isOperand_{false};
 
 public:
   /// Create a new value and register the node we reference.
   /*implicit*/ NodeValue(Node *N);
 
   /// Create a new value for result \p resNo and register the node we reference.
-  NodeValue(Node *N, unsigned resNo);
+  NodeValue(Node *N, unsigned resNo, bool isOperand = false);
 
   /// Create a new operand and register it as a new user to the node.
-  NodeValue(const NodeValue &that) { setOperand(that.getNode(), that.resNo_); }
+  NodeValue(const NodeValue &that, bool isOperand = false)
+      : NodeValue(that.getNode(), that.resNo_, isOperand) {}
 
   /// Unregister old value, assign new NodeValue and register it.
   NodeValue &operator=(const NodeValue &that) {
-    setOperand(that.getNode(), that.resNo_);
+    if (isOperand_) {
+      setOperand(that.getNode(), that.resNo_);
+    } else {
+      node_ = that.node_;
+      resNo_ = that.resNo_;
+    }
     return *this;
   }
 
   /// When deleting an operand we need to unregister the operand from the
   /// use-list of the node it used to reference.
-  ~NodeValue() { setOperand(nullptr, 0); }
-  /// Sets the operand to point to \p N. This method registers the operand as a
-  /// user of \p N.
+  ~NodeValue() {
+    if (isOperand_) {
+      setOperand(nullptr, 0);
+    }
+  }
+  /// Sets the operand to point to \p v. This method registers the operand as a
+  /// user of \p v.
   void setOperand(Node *v, unsigned resNo);
-  /// Get the index which selects a specific result in the SDNode
+  /// Get the index which selects a specific result in the Node.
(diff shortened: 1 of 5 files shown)


rdzhabarov:
Returning const reference from those getters (getFilter, etc) seems to resolve these types of problems.

qcolombet:
Indeed, returning const reference would solve this particular case.
But that’s fragile, because:
1. We claim those are lightweight and copy them around quite a lot.
2. If I slightly change the test to keep the returned value of getFilter into a variable, this will create a copy and the problem will still be here.

I believe we want to tweak/split NodeValue to only do the tracking when it makes sense.
I have a hacky prototype working. @opt-mix said it did exactly that and his implementation is arguably in a better shape.

Bottom line, the fix will come soon

qcolombet:
Quick and dirty patch attached for the record.

tools/ClassGen/NodeBuilder.cpp
```diff
@@ -74,6 +74,16 @@ void NodeBuilder::emitCtor(std::ostream &os) const {
   for (auto &RT : nodeOutputs_) {
     os << "    addResult(" << RT.first << ");\n";
   }
+  for (const auto &op : members_) {
+    if (op.first != MemberType::VectorNodeValue) {
+      continue;
+    }
+    os << "    int i = 0;";
```
opti-mix: \n is missing at the end of this line.

tools/ClassGen/NodeBuilder.cpp
```diff
@@ -74,6 +74,16 @@ void NodeBuilder::emitCtor(std::ostream &os) const {
...
+    os << "    for(const auto &nodeVal : " << op.second << ") {\n";
+    os << "      " << op.second
+       << "_[i++].setOperand(nodeVal.getNode(), nodeVal.getResNo());\n";
+    os << "     }\n";
```
opti-mix: Indentation is a bit off for this line.

include/glow/Graph/Node.h
```diff
@@ -45,30 +45,43 @@ struct NodeValue {
   Node *node_{nullptr};
   /// Specifies the node result number to use.
   unsigned resNo_{0};
+  /// Is this value an operand of some instruction or just a temporary object
+  /// flying around.
+  bool isOperand_{false};
```
opti-mix: This bool isOperand_ will result in consuming an additional word probably. I wonder if we should try to use `llvm::PointerIntPair`. But it is purely an optimization to reduce memory.  No real need to think about it in this PR.

nadavrot:
I think that somewhere in the optimizer we have a map of NodePtr and operand integer to remove the Handle problem. Could be a good idea to remove that as well. 

I like Quentin’s suggestion to create two types: NodeValue and NodeValueHandke, where the first one is just a pod and the second one is the current handle.

qcolombet:
For the record, while taking with @nadavrot off-line, he rightly pointed out that while we are rethinking this part we should step back and see what other use cases we may want to solve as well.
I think @opti-mix mentioned walking through the users.

In any case, that would require more work than just this dirty patch! :P

qcolombet:
Talking with @rdzhabarov and @opti-mix this morning, we thought the hacky patch may not be that bad in the end. One of my concern was how easy it could be to misuse NodeValue in the "operand mode". @opti-mix suggested to make this functionality only accessible to the *Node classes by making the related constructors protected and add a friend relationship.
The newly pushed commit does just that. 

If people think that is the right way to do it, then I'll add more checks in the unittest and squash all the commits.

nadavrot:
@qcolombet If we only use the NodeValue (NodeHandle) inside the Node operands then why do we need the hack? Can't we just change every non operand usage to some POD struct, simular to SDValue? My point is that the handle does not need to be the same struct/class as the POD.

qcolombet:
@nadavrot We definitely could do that. @opti-mix has such patch, but he was liking this patch better because it was making the diff smaller and the changes more contained :).
I don't mind either way.

nadavrot:
@qcolombet @opti-mix what do you think is the cleanest design?

qcolombet:
Test case cleaned up and put in its own PR #1249.
The proper NodeValue refactoring is out thanks to @opti-mix in #1198.

Closing this one.

PR #1154, opened 2018-06-19
non-square padding for Convolution operator

beicy:

PR #1185, opened 2018-06-28
[OpenCLSpecificNodesVerification] Fix broken pad for conv output dims

jfix71:
Fixes missed change from from https://github.com/pytorch/glow/pull/1154 which has broken the OCL unit tests.


rdzhabarov:
Any specific reason we do not have OpenCL tests on CI? (apart from not being a priority so far)

opti-mix:
@rdzhabarov To run the OpenCL tests on CI, the CI machine should have OpenCL drivers installed and in ideal case it should have an OpenCL-enabled graphics card, though sometimes you may use CPUs as OpenCL devices with a special driver.

rdzhabarov:
"the CI machine should have OpenCL drivers installed" - sure.
Yeah, Travis machines normally do not have graphics cards installed, but good idea about the CPUs as OpenCL device.

PR #1188, opened 2018-06-28
Getting Glow compiling and running under Win32.

lygstate:
Getting OpenCL works under Win32.
CPU backend still not works and disabled now for Win32.


lygstate:
What's wrong with the CLA? Why I have the following issue:
```
There were some errors submitting your CLA:
GitHub Username is already associated with an existing contributor.
Email is already associated with an existing contributor.
If you would like to update an existing contributor account, please email cla@fb.com with your details.
```

opti-mix:
@compnerd You may want to have a look at this.

nadavrot:
@lygstate Thanks for doing this work!  I am not sure what's wrong with the CLA.

lib/Graph/Graph.cpp
```diff
@@ -581,7 +581,7 @@ Node *Function::createBroadcast(llvm::StringRef name, NodeValue input,
   // new shape (no action taken) or == 1 (broadcast in that direction). Else
   // the original shape had no dimensions here (after considering axis), so
   // add the new dimension and broadcast in that direction.
-  size_t reshapeDims[newShape.size()];
```
nadavrot: GLOW_VLA is not very readable. I prefer to just use array of size "max_tensor_dimensions" here.

lib/Base/Tensor.cpp
```diff
@@ -219,8 +220,8 @@ template <class ElemTy>
 static bool tryTransposeFastImpl(Handle<ElemTy> &src, Handle<ElemTy> &dest,
                                  llvm::ArrayRef<unsigned> shuffle) {
   const size_t numDims = dest.dims().size();
-  size_t srcCoorArr[numDims];
```
nadavrot: GLOW_VLA is not very readable. I prefer to just use array of size "max_tensor_dimensions" here.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -6,20 +6,26 @@ typedef unsigned cl_uint32_t;
 /// This type is always 64 bits.
 typedef unsigned long cl_uint64_t;
 
+#if GLOW_HOST_SIZEOF_SIZE_T == 4
```
nadavrot: This is not correct. The host does not have to match the device.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -692,8 +713,9 @@ void OCLBackend::doForwardPass() {
           createKernel(useTiledMatMul ? "matmul_tiled" : kernelName);
       setKernelArg(kernel, 0, deviceBuffer_);
 
-      unsigned numArgs = I.getNumOperands();
-      for (unsigned arg = 0; arg < numArgs; arg++) {
+      /* unsigned numArgs = I.getNumOperands(); */
```
nadavrot: What is this?

include/glow/Support/Support.h
```diff
@@ -23,6 +23,21 @@
 #include <iostream>
 #include <sstream>
 
+#ifdef _MSC_VER
```
nadavrot: I prefer to work around the dynamic use and not declare GLOW_VLA which is not very readable.

cmake/modules/FindOpenCL.cmake
```diff
@@ -0,0 +1,184 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
nadavrot: Where did you get this file from? Isn't this the same file as the one that comes with CMake? Why do we need it here?

CMakeLists.txt
```diff
@@ -86,5 +99,5 @@ add_custom_target(dependency_graph
 # Fetch the dependencies for all the tests.
 get_property(GLOW_TEST_DEPENDS GLOBAL PROPERTY GLOW_TEST_DEPENDS)
 
-add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND}
+add_custom_target(check-glow COMMAND ${CMAKE_CTEST_COMMAND}
```
nadavrot: Why did you change this?

tests/unittests/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
+if (MSVC)
```
compnerd: This seems like the result of not building googletest

tools/loader/CMakeLists.txt
```diff
@@ -10,3 +10,7 @@ target_link_libraries(image-classifier
                         ExecutionEngine
                         IR
                         Quantization)
+
+target_include_directories(image-classifier
+  PRIVATE
```
compnerd: Please fix the indentation

tests/unittests/CMakeLists.txt
lygstate: llvm also have googletest and test target

cmake/modules/FindOpenCL.cmake
lygstate: The one in CMake doesn't works properly under win32， I have to copy and modify

lib/Graph/Graph.cpp
lygstate: VLA means Variable Length Array， msvc doesn't support for that， so I place it in support.h， the name comes form 
https://github.com/jerryscript-project/jerryscript/blob/a74bf7dc91e7358f0e6c7e111e727110377fca82/jerry-core/include/jerryscript-compiler.h#L62,
Also are my patch, I hope this would have an agreement

lib/Backends/OpenCL/OpenCL.cpp
lygstate: An bug, test case show it was 4, but we expected 3

nadavrot: @opti-mix Do you mind looking at this bug?

lib/Graph/Graph.cpp
nadavrot: I know what VLA is. I am saying that GLOW_VLA is not very readable. I prefer to use the static size: int foo[max_tensor_dimensions];

lygstate: so does max_tensor_dimensions a constants or a macro？

nadavrot: It does not matter if it’s a macro or a constant number. It is not dynamic.

lygstate: I'll try constant number first, if doesn't works, macro would used.

lib/Backends/OpenCL/OpenCL.cpp
opti-mix: @lygstate I checked under debugger and I see I.getNumOperands() == 3 for MatMulInst instructions. It is also defined with 3 arguments in InstrGen.cpp. 

If you have discovered a case, where it reported 4 arguments, could you provide a description how to reproduce this?

CMakeLists.txt
```diff
@@ -54,26 +57,36 @@ if (GLOW_WITH_OPENCL)
...
-endif()
-if (NOT LLVM_FOUND)
-  message(FATAL_ERROR "Could not find LLVM. Build LLVM manually and configure the project with -DCMAKE_PREFIX_PATH=/path/to/llvm/install")
+if(MSVC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4710 /wd4711 /wd4514 /wd4571 /wd4820 /wd4774 /wd4625")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4526 /wd4061 /wd4365 /wd4242 /wd4996 /wd5026 /wd4626")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd5027 /wd4668 /wd4305 /wd4826 /wd4574")
+else()
+  # Prefer LLVM 6.
+  find_package(LLVM 6 CONFIG)
```
bertmaher: Something doesn't make sense here... llvm is a required dependency of glow.  I don't understand how you can possibly be building without it.

lib/Backends/OpenCL/OpenCL.cpp
lygstate:
```

TEST_P(Operator, simplePredication) {
  auto *inputs = mod_.createVariable(ElemKind::FloatTy, {10, 10, 10}, "inputs");
  auto *counters = mod_.createVariable(ElemKind::FloatTy, {10}, "counters");

  counters->getHandle() = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
  inputs->getHandle().randomize(-10, 10, mod_.getPRNG());

  Node *C5 = F_->createSplat("C5", counters->getType(), 5.0);
  Node *pred = F_->createCmpLTE("cmp", C5, counters);

  auto *FC0 = F_->createFullyConnected("FC0", inputs, 128);
  auto *RL0 = F_->createRELU("RL0", FC0);
  auto *FC1 = F_->createFullyConnected("FC1", RL0, 64);
  auto *RL1 = F_->createRELU("RL1", FC1);
  auto *FC2 = F_->createFullyConnected("FC2", RL1, 32);
  auto *RL2 = F_->createRELU("RL2", FC2);

  F_->createSave("ret", RL2);

  FC0->setPredicate(pred);
  FC1->setPredicate(pred);
  FC2->setPredicate(pred);

  EE_.compile(CompilationMode::Infer, F_);
  EE_.run({}, {});
}
```
The OperatorTest.cpp
simplePredication

CMakeLists.txt
lygstate: @bertmaher , Win32 have no llvm-devel package, I building glow with llvm source.
[

lib/Backends/OpenCL/OpenCL.cpp
lygstate: I have already use the CI to trigger the issue

lib/Backends/OpenCL/kernels.cl
lygstate: For quick fix the inconsistence between Host and Device.

lib/Backends/CPU/libjit/libjit_defs.h
```diff
@@ -26,7 +26,7 @@ typedef __declspec(align(64)) float float16;
 #else
 typedef float float4 __attribute__((ext_vector_type(4)));
 typedef float float8 __attribute__((ext_vector_type(8)));
-typedef float float16 __attribute__((ext_vector_type(16)));
+typedef float float16 __attribute__((aligned(64)));
```
nadavrot: This change is not correct. Float16 would be a regular float that's aligned to 64b.

nadavrot:
@lygstate This PR has a ton of good stuff. Would it be possible to extract more changes out of this PR and start submitting them as small focused PRs?  I think that this work is awesome and it would be great to see these PRs land.

lygstate:
@nadavrot I am waiting for feedback, which part I should split out as smaller patch:)

PR #1192, opened 2018-06-28
Use llvm::StringRef for dumpDAG methods

bertmaher:
I keep wanting to build filenames like `std::string(F->getName()) + "_suffix"`, and it's kind of tedious to wrap that in `(...).c_str()`.  I picked `llvm::StringRef` since it's lightweight compared to `std::string`

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -155,7 +155,7 @@ public:
   void dumpDAG();
 
   /// Dump a dotty graph that depicts the Module.
-  void dumpDAG(const char *dotFilename);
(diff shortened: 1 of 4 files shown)


nadavrot:
Bert, can you check if we can still call dump in the debugger? In LLDB I run “p M->dumpDAG(“file.dot”). “ I am not sure if this PR broke this workflow. 

bertmaher:
Indeed, this breaks that behavior.  How would you feel about #1196, to enable both workflows?  (I keep hoping debuggers will be a true REPL for C++, but alas, not yet.)

PR #1193, opened 2018-06-29
Add a missing include for stdio.h

opti-mix:
GLOW_ASSERT and GLOW_UNREACHABLE were using printf without declaring it first.

### DIFF
--- a/include/glow/Support/Compiler.h
+++ b/include/glow/Support/Compiler.h
@@ -16,6 +16,8 @@
 #ifndef GLOW_SUPPORT_COMPILER_H
 #define GLOW_SUPPORT_COMPILER_H
 
+#include <stdio.h>
+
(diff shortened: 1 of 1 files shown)


include/glow/Support/Compiler.h
```diff
@@ -16,6 +16,8 @@
 #ifndef GLOW_SUPPORT_COMPILER_H
 #define GLOW_SUPPORT_COMPILER_H
 
+#include <stdio.h>
```
rdzhabarov: while we are here, GLOW_GNUC_ATLEAST seems to be unused. Let's purge this as well.

opti-mix: Pushed another commit in this PR that removes GLOW_GNUC_ATLEAST

PR #1194, opened 2018-06-29
[quantization] Fold quantize node into variable with multiple usages.

rdzhabarov:
It appears that tests for the folding were missing :).
Added the base one + multiple usages.


nadavrot:
Nice!

PR #1196, opened 2018-06-29
Allow dumpDAG(const char *) as well

bertmaher:


bertmaher:
It Works On My Machine(TM), so hopefully it'll work for others :-).  If there are problems, I'll revert the StringRef change.

Issue #1178, opened 2018-06-26
Assumptions baked in the Graph optimizer not checked anywhere


qcolombet:
Some of the Graph optimizations assume that some nodes have variables as their inputs (e.g., batch normalization, convolution) and that these variables are used only in those nodes.
Although this currently matches what the importers emit, this creates an unwelcome coupling. In essence, we could spot usage of `cast<Variable>` sprinkled in the code base.
We should either:
1. Check these assumptions in the related verify methods, or
2. Fix the optimizer (and the hand written test cases)

I am tempted to go with #2 because:
A. Only the optimizer, importer and test cases have these assumptions baked in (for instance, the CPU optimization for convolution optimizeCPUConv is fine without those, same for the lowering and the interpreter.)
B. It is not unlikely that constant Variables with the same value would be uniqued and be shared across different uses and that's never checked.
C. With the introduce of the load node, those optimizations will need to be taught how to look through loads anyway.

nadavrot:
I agree. The semantics should be that operators work on activations. Only special nodes such as save and load work on weights.

nadavrot:
This is fixed now. Variables are gone, and save notes only operate on placeholders.

PR #1197, opened 2018-06-29
[GraphOptimizer] Fix batchNorm optimization

qcolombet:
The batchNorm optimization used to assume that the inputs of the related
convolution node were all variables. This is not enforced anywhere and may
actually be wrong.

The fix consists in checking that the inputs of the convolution node are indeed
variables. Moreover, we check that these variables are used only by this
convolution node before doing the transformation. Otherwise we will incorrectly
change the inputs of the other users.

This fixes issue #1178 for the graph optimizer.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -763,13 +763,24 @@ static void optimizePool(Function *F) {
   } // For all nodes in the graph.
 }
 
+/// \returns The uniquely used variable from node or nullptr
+/// if node has more than one user or is not a variable.
+static Variable *getUniquelyUsedVariable(Node &node) {
+  // If that node has more than one use, it may not
+  // be okay to modify the underlying variable.
+  if (!node.hasOneUse()) {
+    return nullptr;
+  }
+  return dyn_cast<Variable>(&node);
+}
+
 static void optimizeBatchNorm(Function *F) {
   auto &nodes = F->getNodes();
 
   // For each node:
   for (auto &node : nodes) {
     // Merge the Batch Normalization operation into the convolution that comes
-    // before it by updating the weights of the filter.
+    // before it by updating the weights of the filter and bias.
     if (auto *BN = dyn_cast<BatchNormalizationNode>(&node)) {
       auto *CV = dyn_cast<ConvolutionNode>(BN->getInput());
       if (!CV) {
@@ -815,13 +826,33 @@ static void optimizeBatchNorm(Function *F) {
       // Q = W * A
       // C = b * A + B
 
-      auto filterH = cast<Variable>(CV->getFilter())->getHandle<>();
-      auto cbiasH = cast<Variable>(CV->getBias())->getHandle<>();
+      // FIXME: We need to use a temporary node to hold on
+      // the variables, because using the result directly
+      // will create a temporary NodeValue that is going
+      // to be otherwise alive at the call site and that
+      // messes up with the number of users.
+      Node *tmp = CV->getFilter().getNode();
+      Variable *filterV = getUniquelyUsedVariable(*tmp);
+      tmp = CV->getBias().getNode();
+      Variable *cbiasV = getUniquelyUsedVariable(*tmp);
 
-      auto scaleH = cast<Variable>(BN->getScale())->getHandle<>();
-      auto biasH = cast<Variable>(BN->getBias())->getHandle<>();
-      auto meanH = cast<Variable>(BN->getMean())->getHandle<>();
-      auto varH = cast<Variable>(BN->getVar())->getHandle<>();
+      if (!filterV || !cbiasV) {
+        continue;
+      }
+
+      Variable *scaleV = cast<Variable>(BN->getScale());
+      Variable *biasV = cast<Variable>(BN->getBias());
+      Variable *meanV = cast<Variable>(BN->getMean());
+      Variable *var = cast<Variable>(BN->getVar());
+
+      auto filterH = filterV->getHandle<>();
+
+      auto cbiasH = cbiasV->getHandle<>();
+
+      auto scaleH = scaleV->getHandle<>();
+      auto biasH = biasV->getHandle<>();
+      auto meanH = meanV->getHandle<>();
+      auto varH = var->getHandle<>();
 
       // Update the filter/bias variables of the Conv node.
       auto epsilon = BN->getEpsilon();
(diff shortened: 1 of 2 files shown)


qcolombet:
Note: This pull request depends on a fix for PR #1184.
Otherwise the batchNorm optimization doesn't kick in at all because getFilter creates a temporary NodeValue that increases the #users by 1.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -763,6 +763,18 @@ static void optimizePool(Function *F) {
...
+static Variable *getModifiableVariable(Node &node) {
+  // If that node has more than one use, it may not
+  // be okay to modify the underlying variable.
+  if (!node.hasOneUse()) {
+    return nullptr;
+  }
+  if (!isa<Variable>(&node)) {
+    return nullptr;
+  }
+  return cast<Variable>(&node);
```
bertmaher: Could replace the isa/cast pair with dyn_cast.  Up to you though, it's pretty clear this way.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -815,13 +827,24 @@ static void optimizeBatchNorm(Function *F) {
...
 
-      auto filterH = cast<Variable>(CV->getFilter())->getHandle<>();
-      auto cbiasH = cast<Variable>(CV->getBias())->getHandle<>();
+      Variable *filterV = getModifiableVariable(*CV->getFilter());
+      Variable *cbiasV = getModifiableVariable(*CV->getBias());
+
+      Variable *scaleV = getModifiableVariable(*BN->getScale());
+      Variable *biasV = getModifiableVariable(*BN->getBias());
+      Variable *meanV = getModifiableVariable(*BN->getMean());
+      Variable *var = getModifiableVariable(*BN->getVar());
```
bertmaher: The BatchNorm variables don't technically have to be single-use, right?  Since we're actually modifying the Conv inputs?

qcolombet: Good catch!

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -763,6 +763,15 @@ static void optimizePool(Function *F) {
   } // For all nodes in the graph.
 }
 
+static Variable *getModifiableVariable(Node &node) {
```
nadavrot: Doxygen comments please.

Also, the name is not great, because "Modifiable" is subjective and depends on the context.

nadavrot:
@qcolombet To remove the dependency on the NodeValue refactoring you could use getFilter().getNode().

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -815,13 +824,24 @@ static void optimizeBatchNorm(Function *F) {
...
 
-      auto filterH = cast<Variable>(CV->getFilter())->getHandle<>();
-      auto cbiasH = cast<Variable>(CV->getBias())->getHandle<>();
+      Variable *filterV = getModifiableVariable(*CV->getFilter());
+      Variable *cbiasV = getModifiableVariable(*CV->getBias());
+
+      Variable *scaleV = dyn_cast<Variable>(BN->getScale());
+      Variable *biasV = dyn_cast<Variable>(BN->getBias());
+      Variable *meanV = dyn_cast<Variable>(BN->getMean());
+      Variable *var = dyn_cast<Variable>(BN->getVar());
```
rdzhabarov: based on the way we build batch normalization node we expect all those 4 be a variable, maybe place an assert on those instead of if checking.
This way if it changes we could adopt implementation of optimization otherwise this optimization will not trigger at all (would be easy to skip).

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -815,13 +824,24 @@ static void optimizeBatchNorm(Function *F) {
       // Q = W * A
       // C = b * A + B
 
-      auto filterH = cast<Variable>(CV->getFilter())->getHandle<>();
-      auto cbiasH = cast<Variable>(CV->getBias())->getHandle<>();
+      Variable *filterV = getModifiableVariable(*CV->getFilter());
```
rdzhabarov: can you also update comment above:
```
 // Merge the Batch Normalization operation into the convolution that comes
     // before it by updating the weights of the filter.
```
It changes convolution filter and bias.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: That's the thing, this is going to change, because only load and save would be able to work directly with variables.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: Good catch!

qcolombet:
Hi @nadavrot,

Thanks for the review.

One comment:
> @qcolombet To remove the dependency on the NodeValue refactoring you could use getFilter().getNode().

I've tried, but that doesn't work either. We still have a temporary NodeValue alive at the call site.
To really workaround this limitation, we would need to use a temporary variable like so:
```
Node *tmp = CV->getFilter().getNode();
Variable *filterV = getModifiableVariable(*tmp);
```

I can go with that to unblock this, but it feels unnatural.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: I'll go back to the cast, because I'll have to change this code anyway when the change for loads land. It used to be consistent when I was using getModifiableVariable for all, but now that I rolled back to a straight dyn_cast, that doesn't make much sense.

Anyway, reverting to the straight cast here.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -815,13 +826,33 @@ static void optimizeBatchNorm(Function *F) {
...
       // C = b * A + B
 
-      auto filterH = cast<Variable>(CV->getFilter())->getHandle<>();
-      auto cbiasH = cast<Variable>(CV->getBias())->getHandle<>();
+      // FIXME: We need to use a temporary node to hold on
+      // the variables, because using the result directly
+      // will create a temporary NodeValue that is going
+      // to be otherwise alive at the call site and that
+      // messes up with the number of users.
+      Node *tmp = CV->getFilter().getNode();
```
nadavrot: I wonder why getFilter returns NodeValue instead of NodeValue&. But that's a different story.

qcolombet: Agreed :).

nadavrot:
Nice!

PR #1198, opened 2018-06-29
Make sure that use-lists of Nodes only contain uses by other Nodes

opti-mix:
Until now, NodeValue was used to refer to the results of Nodes. But unfortunately, every NodeValue would add itself to the use-list of a Node it references, which is rather expensive. In many cases, NodeValue objects are created as temporary objects and it makes no sense to consider them to be uses. Moreover, having these extra uses in the use-list prevents some optimizations, etc.

We want to count only references from Nodes to the results of other Nodes as uses. To achieve that this commit introduces a new type NodeValueHandle. It is used instead of NodeValue inside Nodes to refer to the results of other nodes. All NodeValueHandles add themselves as uses to the use lists. And the former NodeValue is changed in such a way that it does not register itself in the use-lists anymore.

To ensure that NodeValueHandle cannot be used in the code outside of Node and its derived classes some kind of a type system  based access control is introduced by means of introducing a "namespace" PrivateNodeTypes, whose members can be accessed only by Node and its derived classes.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -31,47 +31,50 @@ namespace glow {
 
 class Node;
 class NodeWalker;
+struct NodeUse;
 
 /// Unlike LLVM values, graph nodes may return multiple values as the result of
 /// a computation. Gradient-calculating nodes such as conv-grad return multiple
 /// values. As such, each use of a node computation must indicate the node that
 /// computes it as well as which return value to use from that node. This pair
-/// of information is represented in this class. This class also manages the
-/// node use-def chain, by registering and removing the address of the value
-/// from the use-list. This data structure is similar to LLVM's SDValue.
+/// of information is represented in this class.
+
+/// NodeValue is a simple POD struct that contains a reference to a node and a
+/// result number.
 struct NodeValue {
-private:
+protected:
   /// A pointer to the node (owned by the graph).
   Node *node_{nullptr};
   /// Specifies the node result number to use.
   unsigned resNo_{0};
 
 public:
-  /// Create a new value and register the node we reference.
+  /// Create a new value.
   /*implicit*/ NodeValue(Node *N);
 
-  /// Create a new value for result \p resNo and register the node we reference.
+  /// Create a new value for result \p resNo.
   NodeValue(Node *N, unsigned resNo);
 
-  /// Create a new operand and register it as a new user to the node.
-  NodeValue(const NodeValue &that) { setOperand(that.getNode(), that.resNo_); }
+  /// Create a new value from an existing one.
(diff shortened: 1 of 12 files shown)


opti-mix:
This is an alternative implementation of #1184. The main difference is that it uses two different types now. NodeValue is a simple POD, which does not register itself in the use-list. NodeValueHandle is the new type, which is used to reference nodes from other nodes. Only NodeValueHandle register itself in the use-list, i.e. only references to nodes from other nodes are registered there.

Some type-system magic is used to make sure that NodeValueHandle is invisible outside the Node type and its derived classes.

No dynamic bits are required to ensure a proper behavior as opposed to the @qcolombet's PR.

BTW, since NodeValueHandle is a separate class now, we can add more fields in the future. E.g. a back-reference to the Node that contains this NodeValueHandle. This would allow for iteration over all  Nodes that contain NodeValueHandle from the use-list, i.e. one could iterate over all user nodes.

opti-mix:
I'm not very happy with the amount of code in this PR. Defining a new type and "protecting" it from misuse needs more code than I'd like ...

opti-mix:
Hmm. The build failure is due to `gcc` that is used to collect the coverage info. Apparently, it does not respect some of the `friend` declarations. I reduced the test-case to this:

```c++
namespace glow {

template <typename UserTy, typename UseTy, typename Use> class UseDef {
};

class Node;
struct NodeUse;

struct NodeValue {
};

class PrivateNodeTypes {
  friend class Node;
  struct NodeValueHandle : NodeValue {
  };
};


struct NodeUse {
};

class Node : public UseDef<Node, PrivateNodeTypes::NodeValueHandle, NodeUse> {
};

}
```

I tried it on https://godbolt.org/ and it seems like newer version of GCC do not have any problems with it. But older ones produce the same error as our CI. Looks like it is a bug in GCC that was fixed in the meantime.

@artemrakhov @rdzhabarov Can we update the version of GCC used by CI?

nadavrot:
@opti-mix Thanks for doing this work. The general direction makes sense to me. 

I am not sure why we need the type protection, for three reasons. First, the name Handle already signals the properties of the class. And second, it could be useful to use a handle not a part of an operand (similar to LLVM's ValueHandle). And third, as evident by this PR, this extra code is hurting us in terms of complexity and portability (gcc does not build, who knows what MSCV does). 

I would approach this work in two steps. First, rename NodeValue to NodeHandle. This is a small mechanical change. Then, in a separate commit I would create NodeValue (similar to SDNode) and change the relevant interfaces one at a time.

rdzhabarov:
@opti-mix after https://github.com/pytorch/glow/pull/1251 merged, you should be good.

opti-mix:
Fixed by #1274. Closing.

PR #1200, opened 2018-06-30
Force to verify MatMulInst operands count.

lygstate:
Used to trigger the error.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1204,6 +1204,7 @@ void Interpreter::fwdElementSelectInst(const glow::ElementSelectInst *I) {
 }
 
 void Interpreter::fwdMatMulInst(const glow::MatMulInst *I) {
+  GLOW_ASSERT(I->getNumOperands() == 3 && "MatMulInst should have 3 operands");
   if (getTensor(I->getLHS())->getType().isQuantizedType()) {
     auto lhs = getTensor(I->getLHS())->getHandle<int8_t>();
     auto rhs = getTensor(I->getRHS())->getHandle<int8_t>();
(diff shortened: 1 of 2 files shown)


nadavrot:
@lygstate Thanks for doing this work.  Can you please use a regular assert()? GLOW_ASSERT is used for runtime checks.  Glow uses assert() in a similar way to how LLVM uses asserts to check for internal errors.

lygstate:
@nadavrot Regular assert for what? do you want me to print something or some other reason?

lygstate:
@nadavrot We already have the test result:
```
../lib/Backends/Interpreter/InterpreterNodes.cpp:1207: failed assertion `I->getNumOperands() == 3 && "MatMulInst should have 3 operands"'

```

jfix71:
@lygstate `GLOW_ASSERT()` is not removed for release builds. We prefer these sorts of checks to be used with regular `assert()` which is.

nadavrot:
@lygstate Ah, sorry. I missed the fact that this is just a test commit. Thanks.

opti-mix:
I have an idea how this could happen. Probably this happened during the execution of some predication tests.

Look at the following code:
```c++
void Instruction::setPredicate(Value *p) {
  // Push a new predicate.
  if (!hasPredicate()) {
    predicateIndex_ = getNumOperands();
    pushOperand({p, OperandKind::In});
  }

  setOperand(predicateIndex_, p);
}
```

It seems like it would add a new operand to any instruction which doesn't have a predicate yet and thus increase the reported number of operands, because  `getNumOperands` currently reports the size of the `ops` vector, including the predicate.

One way to solve it may be to change getNumOperands() to take the existence of a predicate into account and ignore it, i.e. something like this:

```c++
unsigned getNumOperands() const { return ops_.size() - (hasPredicate() ? 1 : 0) ; }
```

Does it make sense?

nadavrot:
When we discussed the predicate design we decided to add it as a formal operand to simplify the scheduler and other optimizations that scan the operand list.

lygstate:
Add extra function getInstNumOperands()?

bertmaher:
Fixed by #1206 -- thanks so much for reporting & reproing this!

PR #1201, opened 2018-06-30
[OCL/Backends] Templatize NCHW conversion for use by other backends

jfix71:
This allows for other backends to do NCHW conversion more simply. Note that because OCL currently only supports equal padding, it does not use the templatized version of the conversion loop for Conv.

### DIFF
--- a/include/glow/Backends/LayoutConverter.h
+++ b/include/glow/Backends/LayoutConverter.h
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


include/glow/Backends/LayoutConverter.h
```diff
@@ -0,0 +1,69 @@
...
+ */
+#ifndef GLOW_BACKENDS_LAYOUTCONVERTER_H
+#define GLOW_BACKENDS_LAYOUTCONVERTER_H
+
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Nodes.h"
+
+#include "llvm/Support/Casting.h"
+
+using llvm::dyn_cast;
```
nadavrot: I don't think that the dyn_cast is used in this file.

include/glow/Backends/LayoutConverter.h
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_LAYOUTCONVERTER_H
+#define GLOW_BACKENDS_LAYOUTCONVERTER_H
+
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Nodes.h"
+
+#include "llvm/Support/Casting.h"
```
nadavrot: We are not using any casting in this file.

PR #1202, opened 2018-07-02
Make graph scheduler dtor virtual

bertmaher:
We use `Wnon-virtual-dtor` internally at FB and it's generally a good practice, so let's conform to it here too.


rdzhabarov:
Does this trigger any warnings for class Node : public Named?

I think there are many cases when a base class does not have virtual dtor in the code base.

bertmaher:
Nope, this doesn't trigger for `Node`.  `Node` doesn't have any virtual methods, so I think it doesn't need a virtual destructor.

PR #1205, opened 2018-07-02
[quantization/optimization] Merge rescales into min, max nodes.

rdzhabarov:
@jfix71 mentioned that there are some redundant rescales on the fr2en graph.
This PR should help to eliminate those as well.


jfix71:
Great, thanks!

PR #1206, opened 2018-07-03
[opencl] Re-factor and improve the logic to pass buffer operands to OpenCL kernels

opti-mix:
Move the often repeated logic for providing the buffers arguments into a dedicated function.
Teach this code to ignore the predicate operands as they are not supported by the OpenCL backend yet.

This PR should fix the issue demonstrated by #1200

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -201,6 +201,36 @@ void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
   GLOW_ASSERT(err == CL_SUCCESS && "Unable to set parameter");
 }
 
+/// Set OpenCL \p kernel arguments using the buffer operands of the
+/// instruction \p I. The first of these arguments should be passed to the \p
+/// kernel at index \p nextKernelArgIdx. The \p tensors map provides a mapping
+/// from Values to on-device buffers addresses of these values.
+///
+/// \returns the index of the last set OpenCL kernel argument.
+static size_t
+setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
+                        size_t nextKernelArgIdx,
+                        std::unordered_map<const Value *, size_t> &tensors) {
+  // Number of instruction operands.
+  auto numArgs = I.getNumOperands();
+  // The predicate of the instruction if available.
+  Value *predicate = I.hasPredicate() ? I.getPredicate() : nullptr;
+  // The index of the kernel argument to be set.
+  unsigned kernelArgIdx = nextKernelArgIdx;
+  // Go over all operands and pass buffer operands to the kernel.
+  for (unsigned arg = 0; arg < numArgs; arg++) {
+    auto *value = I.getOperand(arg).first;
+    // Ignore predicate operands as they are not supported by the OpenCL backend
+    // yet.
+    if (value == predicate)
+      continue;
+    // The value is a buffer that should be passed as a kernel argument.
+    setKernelArg<cl_uint>(kernel, kernelArgIdx, tensors[value]);
+    kernelArgIdx++;
+  }
+  return kernelArgIdx - 1;
+}
+
 void OCLBackend::fillBuffer(cl_mem buffer, size_t start, size_t len,
                             float value, ElemKind elemKind) {
   auto kernel = createKernel(getKernelName("splat", elemKind));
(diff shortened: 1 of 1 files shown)


opti-mix:
I checked. The `simplePredication` tests fails if we add this assert. We should probably disable this test for OpenCL.

opti-mix:
@bertmaher I'm a bit concerned doing it in the OpenCL backend only. It seems like none of our backends support the code generation for the predicated instructions. But they do not explicitly assert or bail on them. They seem to silently ignore it and just get lucky by not crashing on them during the codegen, because they use a different way of accessing the operands (e.g. by name, not by index). 

So, I wonder if we should handle the predication the same way in all of our backends, rather than asserting in the OpenCL one and ignoring them in others.

No strong opinion on this though.

bertmaher:
Ah, yeah, I kind of suspected that was the case.  I favor asserting (or actually implementing) predication over silently pretending to work.  But I'm cool with this diff since it solves a real problem and doesn't make things weirder than they currently are :-).

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -167,7 +168,11 @@ class OCLBackend final : public Backend {
                      cl_device_id device, llvm::ArrayRef<size_t> global,
                      llvm::ArrayRef<size_t> local,
                      std::vector<KernelLaunch> &kernelLaunches);
-
+  /// Set OpenCL \p kernel arguments using the buffer operands of the
+  /// instruction \p I. The first of these arguments should be passed to the \p
+  /// kernel at index \p nextArgIdx.
+  size_t setKernelArgsForBuffers(cl_kernel kernel, const glow::Instruction &I,
```
bertmaher: Can you make this a static in the cpp file?  It doesn't look it needs to be exported.

opti-mix:
@nadavrot 

We have two predication tests `simplePredication` and `nodesWithPredicates`, who use `ExecitionEngine::compile` and `ExecutionEngine::run`. They currently pass just because all of our backends are unaware of predicates and simply ignore them and the tests do not check if the backends do anything reasonable with predicates. The only thing that is really exercised by these tests as part of the `compile` call is that we can perform the IRGen and produce a predicated IR. So, may be the tests should just call `generateIR` for now, until we really support the predication in our backends? Or we could even disable these tests for now.

What's your opinion about these predication tests & asserting in the backends if they see an instruction with a predicate, as we do not support the predication in any of the backends?

If we go this route of asserting on predicated instructions, this PR could be simplified as there would be no need for any special-casing on predicate operands.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -201,6 +201,30 @@ void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
   GLOW_ASSERT(err == CL_SUCCESS && "Unable to set parameter");
 }
 
+size_t OCLBackend::setKernelArgsForBuffers(cl_kernel kernel,
```
nadavrot: Doxygen comments please.

nadavrot:
@opti-mix @bertmaher Thanks for doing this work. I think that the refactoring part of the PR is awesome.

About the predicates, this is work in progress. The semantics of the predicates are defined in such a way that the result is unspecified if the predicate is set to zero. In other words, the backend is free to return garbage if the predicate is set to zero. This means that it's always legal to drop the predicates. It could be useful for the implementation of byzantine operators that we don't want to predicate (like, how do you predicate topk?).

I think that the question should be around the best way to make progress on the implementation of predication. At the moment we simply wire the predicated operand but ignore it on all backends. What do you think would be the best way to move forward with the implementation? 

CC @qcolombet

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -773,19 +755,14 @@ void OCLBackend::doForwardPass() {
...
       auto odim = ShapeNHWC(CC->getDest()->getType()->dims());
       auto idim = ShapeNHWC(CC->getSrc()->getType()->dims());
 
-      setKernelArg<cl_uint>(kernel, 5, CC->getKernel());
-      setKernelArg<cl_uint>(kernel, 6, CC->getStride());
-      setKernelArg<cl_uint>(kernel, 7, CC->getPads()[0]);
+      setKernelArg<cl_uint>(kernel, numArgs + 1, CC->getKernel());
+      setKernelArg<cl_uint>(kernel, numArgs + 2, CC->getStride());
+      setKernelArg<cl_uint>(kernel, numArgs + 3, CC->getPads()[0]);
       setKernelArg(kernel, 8, odim);
```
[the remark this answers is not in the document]
opti-mix: Good catch, @jfix71 ! Will fix.

opti-mix:
Pushed the update:
- The function `setKernelArgsForBuffers` is now a static function in OpenCL.cpp (based on @bertmaher's comment)
- A couple of `setKernelArg` calls that I missed initially are properly updated now (thanks @jfix71 for noticing it)

I left the logic for ignoring the predicate operands in place. We do not assert on them, because per @nadavrot's explanation it's always legal to drop the predicates.

bertmaher:
shipitwithfire

PR #1207, opened 2018-07-03
Remove "emulator" binary

bertmaher:
This looks like something unused, but please correct me if I'm wrong.


rdzhabarov:
Nice emulator :)

PR #1208, opened 2018-07-03
[graph] Fix layering violations

bertmaher:
Graph shouldn't depend on headers from IR or Optimizer.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -18,7 +18,6 @@
 
 #include "glow/Base/Type.h"
 #include "glow/Graph/Nodes.h"
-#include "glow/Optimizer/Optimizer.h"
 
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/DenseMap.h"
@@ -27,7 +26,6 @@
 #include "llvm/ADT/ilist_node.h"
 
 #include <list>
-#include <unordered_map>
 #include <vector>
 
 namespace glow {
--- a/include/glow/Graph/UseDef.h
+++ b/include/glow/Graph/UseDef.h
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef GLOW_IR_USEDEF_H
-#define GLOW_IR_USEDEF_H
+#ifndef GLOW_GRAPH_USEDEF_H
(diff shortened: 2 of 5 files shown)


include/glow/IR/IR.h
```diff
@@ -19,7 +19,7 @@
 #include "glow/Base/Traits.h"
 #include "glow/Base/Type.h"
 #include "glow/Graph/Graph.h"
-#include "glow/IR/UseDef.h"
+#include "glow/Graph/UseDef.h"
```
rdzhabarov: can you change #ifdef inside the IR/UseDef also? (after file rename)

include/glow/Graph/Graph.h
```diff
@@ -18,7 +18,6 @@
 
 #include "glow/Base/Type.h"
 #include "glow/Graph/Nodes.h"
-#include "glow/Optimizer/Optimizer.h"
```
rdzhabarov: are these in use?
"llvm/ADT/ilist.h"
"llvm/ADT/ilist_node.h"

lib/Optimizer/Quantization.cpp
```diff
@@ -17,6 +17,7 @@
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Node.h"
 #include "glow/Graph/Nodes.h"
+#include "glow/Optimizer/Optimizer.h"
```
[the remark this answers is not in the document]
bertmaher: It just contains the prototype for profileQuantization.  We could also just wrap the def (below) in `namespace glow { ... }`, but it seemed cleaner to include the header.

include/glow/Graph/UseDef.h
```diff
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef GLOW_IR_USEDEF_H
-#define GLOW_IR_USEDEF_H
+#ifndef GLOW_GRAPH_USEDEF_H
+#define GLOW_GRAPH_USEDEF_H
```
rdzhabarov: also at the very bottom

bertmaher: sigh, thanks :).  There's a reason I <3 `#pragma once`

nadavrot:
❤️

PR #1209, opened 2018-07-03
[importer] Use Variable through Load node

qcolombet:
This patch wires the load node in the importer.
This is one step closer for making the load and save nodes the
only users of Variable.

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -101,9 +101,9 @@ protected:
   /// \returns the tensor that was registered under the name \p name.
   Tensor *getTensorByName(llvm::StringRef name);
 
-  /// Create a new variable \p name initialized with \p tensor.
-  /// The created variable is also registered under \p name.
-  /// \returns the newly created variable.
+  /// Create a new variable \p name initialized with \p tensor and load it.
+  /// The load is also registered under \p name.
+  /// \returns A load to the newly created variable.
   /// \pre !hasNodeByName(name)
   Node *createAndRememberVariable(
       llvm::StringRef name, Tensor &tensor,
@@ -120,8 +120,16 @@ public:
   Node *getNodeByName(llvm::StringRef name) const;
 
   /// \returns the node that was registered with the name \p name or create a
-  /// new Variable node for a tensor with this name.
+  /// new Variable node for a tensor with this name and load it.
+  /// In case a new variable is returned, this method registers
+  /// the returned load node, not the variable itself.
+  /// Only save and load node are allowed to use variable directly.
   Node *getOrCreateVariableByName(llvm::StringRef name);
+
+  /// \returns The variable referenced by the load registered
+  /// under \p name.
+  /// \pre isa<LoadNode>(getNodeByName(name))
+  Variable *getVariableByName(llvm::StringRef name) const;
 
   /// Create a new variable \p name initialized with \p tensor.
   /// \returns the newly created variable.
--- a/lib/Importer/ProtobufLoader.cpp
+++ b/lib/Importer/ProtobufLoader.cpp
@@ -64,9 +64,10 @@ Node *ProtobufLoader::createAndRememberVariable(llvm::StringRef name,
                                                 VisibilityKind visibilityKind,
                                                 Variable::TrainKind trainKind) {
   assert(!hasNodeByName(name) && "Creating an already existing node?!");
-  auto *V = createVariable(name, tensor, visibilityKind, trainKind);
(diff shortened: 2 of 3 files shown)


qcolombet:
Note: Although I don't believe this PR will change a lot, I won't land it yet, because I am still investigating the codegen impact of this. In particular, we fail to eliminate some of the copies introduced by the loads and for some workloads it impacts codegen. For instance, resnet50 is identical with this patch, but densenet121 is not.
I believe the fix will be in the optimization that coalesces the allocactivations, thus sending this patch early anyway for feedback.

qcolombet:
It's because the loader creates some variables to feed the inference engine.
I didn't think it makes sense to force loads here.

An alternative would be to have the loader creates these variables directly (createVariable) instead of using the getOrCreate mechanism. That way we wouldn't need this concept of naked.

That said, I thought it may be useful to have an API to access the variables directly, but other that this use case, it may not be that useful.

What do you think?

qcolombet:
Bringing a bit more details, like you guessed @bertmaher, the naked is indeed for the input activations, it tells the inference engine where to copy the input tensor. So, this is really a location (as in address) and the variables make sense here I would say.

bertmaher:
Makes sense, thanks for the explanation.  I'd ideally like to avoid the boolean param (just from a code-complexity standpoint) but it's not really a huge deal.  Happy to accept whenever you get the codegen optimizations worked out to your liking.

qcolombet:
While looking at getting rid of the boolean, I put the pieces together of what we said :).

Only input activations need to access the variable directly, that is when we use the related variables for inference, and that means the graph has already been constructed.
Therefore, we only need a way to access them, not to create them.

I believe the new version is much simpler and it demonstrates more clearly what is the intent of each calls.

Regarding the optimization problem, I think I understood what is going on. In a nutshell, our buffer optimization only looks at merging one segment at a time. In the case of the lowering of the load node, we have for the first time potentially more than one copy on a chain to a use (e.g., with the lowering of reshape(load)) .
Now to illustrate the problem with chain of copy, let us consider:
1: a = copy b
2: c = copy a
3: = c
We have the following segments:
a = (1, 2); b = (-,1); c (2, 3)

First we merge c and a:
1: a = copy b
2: a = copy a
3: = a
Now the segments look like:
a = (1, 2), (2, 3); b = (-,1)

When considering merging a with b, we only consider updating the segment of a that encloses the segment of b, thus (1,2).
The updated code looks like:
1: b = copy b
2: a = copy b
3: = a
We don't update the old c segment and end up with a copy in 2 now.

Interestingly, if you delete the useless copies as you go, you actually end up with incorrect code, e.g., for the previous example:
1: b = copy b
---2: a = copy b---- deleted
3: = a  <--- a never defined

This is because when we update the instruction, we don't preserve the last definition of the outgoing variable and given we don't update the uses not in the segment, they potentially reference an undefined variable.

It looks like it works just by accident but maybe it was intended this way?

Long story short we need to lift the one-segment-at-a-time limitation (maybe by merging the segments?).

I'll look closer when I come back.

opti-mix:
@qcolombet Regarding your analysis. It could indicate a real bug in the buffers sharing optimization. Could you provide/construct a small test-case that reproduces it?

qcolombet:
If I could feed the optimization directly with some IR maybe, but from the graph I am not sure. I am not sure how to expose two copies in a row without the load node (which I don’t want to land until the code gen look sane :)). The obvious candidate would be two reshapes but I think we optimize them into one.

FWIW, resnet50 does not expose the problem only densenet121 as far as I checked and I don’t know yet what’s special in densenet to expose this problem.

opti-mix:
@qcolombet We have some tests that create the IR directly, not from the graph, IIRC. This should be enough to construct the test-case, right?

jfix71:
@qcolombet @opti-mix Yeah, we should be able to write a test in `IROptTest.cpp` to reproduce this

qcolombet:
Thanks @jfix71, @opti-mix, trying to write a test case from IR.

qcolombet:
@opti-mix, @jfix71, kind of painful (because you have to survive through the optimizations), but I was able to write test cases demonstrating the issue.
See #1237.
The first test case goes smoothly through the optimization, whereas the second ends up with a useless copy. The second test case is the same as the first test case except it has a chain of two copies instead of just one for the variable xxx_loaded.

opti-mix:
@qcolombet Thanks for creating the self-contained test reproducing the issue!

qcolombet:
With #1240 fixed, the generated LLVM IR is identical with and without this PR.

PR #1210, opened 2018-07-03
[opencl] Create shader program text with xxd

bertmaher:
While I like the R("...") string trick, it's often bugged me that I don't get syntax highlighting in these .cl files since it's just one big string.  Importing the string using `xxd -i` solves that issue, albeit at the cost of a very slightly weirder implementation.  WDYT?

### DIFF
--- a/lib/Backends/OpenCL/CMakeLists.txt
+++ b/lib/Backends/OpenCL/CMakeLists.txt
@@ -1,4 +1,17 @@
+set(KERNELS_HDR ${GLOW_BINARY_DIR}/kernels_cl.h)
+set(KERNELS_FWD_CONV_HDR ${GLOW_BINARY_DIR}/kernels_fwd_conv_cl.h)
+
+add_custom_command(
+        OUTPUT "${KERNELS_HDR}"
+        COMMAND xxd -i < ${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl > ${KERNELS_HDR})
+
+add_custom_command(
+        OUTPUT "${KERNELS_FWD_CONV_HDR}"
+        COMMAND xxd -i < ${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_conv.cl > ${KERNELS_FWD_CONV_HDR})
+
 add_library(OpenCL
+            ${KERNELS_HDR}
+            ${KERNELS_FWD_CONV_HDR}
             OpenCL.cpp
             Transforms.cpp)
 
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -39,10 +39,17 @@ using llvm::isa;
 
 typedef uint32_t cl_size_t;
 
-// This defines the string "SHADER_CODE".
-#include "kernels.cl"
+// This defines the OpenCL kernels.
+const char kernels_cl[] = {
+#include "kernels_cl.h"
+  , 0x00
+};
+
 // This defines kernels for optimized convolutions.
-#include "kernels_fwd_conv.cl"
+const char kernels_fwd_conv_cl[] = {
+#include "kernels_fwd_conv_cl.h"
+  , 0x00
+};
 
 namespace {
 llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
@@ -102,7 +109,7 @@ OCLBackend::OCLBackend(IRFunction *F) : F_(F), allocator_(0xFFFFFFFF) {
 
   err = CL_SUCCESS;
   /// Create the program from the source.
-  createProgram(SHADER_CODE, {}, commands_);
+  createProgram(&kernels_cl[0], {}, commands_);
 }
 
 OCLBackend::~OCLBackend() {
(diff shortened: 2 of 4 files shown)


rdzhabarov:
I do not have a strong opinion and usually do not work with opencl code :P
But when I did I just commented out the string part "static const char* SHADER_CODE = R"(" to get IDE to recognize the format.

lygstate:
make sense for me

lygstate:
no xxd on Windows.

opti-mix:
@lygstate It should be trivial to replace xxd with something similar, e.g. a small C program or a Python script that would be part of Glow. Such a utility just needs to read the input file and produce a hex array. In fact, even the xxd itself with all its options is https://github.com/lwilletts/xxd/blob/master/xxd.c is self-contained and rather small.

lygstate:
@opti-mix There is pure CMake version, I am working on it:
```
include(CMakeParseArguments)

# Function to wrap a given string into multiple lines at the given column position.
# Parameters:
#   VARIABLE    - The name of the CMake variable holding the string.
#   AT_COLUMN   - The column position at which string will be wrapped.
function(WrapString)
    set(oneValueArgs VARIABLE AT_COLUMN)
    cmake_parse_arguments(WRAP_STRING "${options}" "${oneValueArgs}" "" ${ARGN})

    string(LENGTH ${${WRAP_STRING_VARIABLE}} stringLength)
    math(EXPR offset "0")

    while(stringLength GREATER 0)

        if(stringLength GREATER ${WRAP_STRING_AT_COLUMN})
            math(EXPR length "${WRAP_STRING_AT_COLUMN}")
        else()
            math(EXPR length "${stringLength}")
        endif()

        string(SUBSTRING ${${WRAP_STRING_VARIABLE}} ${offset} ${length} line)
        set(lines "${lines}\n${line}")

        math(EXPR stringLength "${stringLength} - ${length}")
        math(EXPR offset "${offset} + ${length}")
    endwhile()

    set(${WRAP_STRING_VARIABLE} "${lines}" PARENT_SCOPE)
endfunction()

# Function to embed contents of a file as byte array in C/C++ header file(.h). The header file
# will contain a byte array and integer variable holding the size of the array.
# Parameters
#   SOURCE_FILE     - The path of source file whose contents will be embedded in the header file.
#   VARIABLE_NAME   - The name of the variable for the byte array. The string "_SIZE" will be append
#                     to this name and will be used a variable name for size variable.
#   HEADER_FILE     - The path of header file.
#   APPEND          - If specified appends to the header file instead of overwriting it
#   NULL_TERMINATE  - If specified a null byte(zero) will be append to the byte array. This will be
#                     useful if the source file is a text file and we want to use the file contents
#                     as string. But the size variable holds size of the byte array without this
#                     null byte.
# Usage:
#   BinToHeader(SOURCE_FILE "Logo.png" HEADER_FILE "Logo.h" VARIABLE_NAME "LOGO_PNG")
function(BinToHeader)
    set(options APPEND NULL_TERMINATE)
    set(oneValueArgs SOURCE_FILE VARIABLE_NAME HEADER_FILE)
    cmake_parse_arguments(BIN2H "${options}" "${oneValueArgs}" "" ${ARGN})

    # reads source file contents as hex string
    file(READ ${BIN2H_SOURCE_FILE} hexString HEX)
    string(LENGTH ${hexString} hexStringLength)

    # appends null byte if asked
    if(BIN2H_NULL_TERMINATE)
        set(hexString "${hexString}00")
    endif()

    # wraps the hex string into multiple lines at column 32(i.e. 16 bytes per line)
    WrapString(VARIABLE hexString AT_COLUMN 32)
    math(EXPR arraySize "${hexStringLength} / 2")

    # adds '0x' prefix and comma suffix before and after every byte respectively
    string(REGEX REPLACE "([0-9a-f][0-9a-f])" "0x\\1, " arrayValues ${hexString})
    # removes trailing comma
    string(REGEX REPLACE ", $" "" arrayValues ${arrayValues})

    # converts the variable name into proper C identifier
    string(MAKE_C_IDENTIFIER "${BIN2H_VARIABLE_NAME}" BIN2H_VARIABLE_NAME)
    string(TOUPPER "${BIN2H_VARIABLE_NAME}" BIN2H_VARIABLE_NAME)

    # declares byte array and the length variables
    set(arrayDefinition "const unsigned char ${BIN2H_VARIABLE_NAME}[] = { ${arrayValues} };")
    set(arraySizeDefinition "const size_t ${BIN2H_VARIABLE_NAME}_SIZE = ${arraySize};")

    set(declarations "${arrayDefinition}\n\n${arraySizeDefinition}\n\n")
    if(BIN2H_APPEND)
        file(APPEND ${BIN2H_HEADER_FILE} "${declarations}")
    else()
        file(WRITE ${BIN2H_HEADER_FILE} "${declarations}")
    endif()
endfunction()
```

opti-mix:
@lygstate Something similar: https://github.com/stevenlovegrove/Pangolin/blob/master/CMakeModules/EmbedBinaryFiles.cmake Seems to be a bit shorter. But it is probably essentially the same.

lygstate:
@opti-mix Too simple, that we will have a very long single line code.
That's would cause the editor not happy./

PR #1215, opened 2018-07-05
[cmake] Fix cyclic dependencies

rdzhabarov:
Cleanup unnecessary dependency. NFC, only some code movement.

There are some unnecessary dependency cycles which should be fixed before adding BUCK support.
screen shot 2018-07-04 at 9 50 49 pm

Fixed: Interpreter -> Quantization -> EE -> Backends -> Interpreter.
Fixed: EE -> Optimizer -> Quantization -> EE


jfix71:
Doesn't GraphOptimizer actually depend on Quantization? E.g. `optimizeQuantizedMaxSplat()` calls `quantization::dequantize()`.

rdzhabarov:
Good finding! "quantization::dequantize" is indeed in use. Will update this PR with the Interpreter change as well, as it touches the same pieces of quantization library (quantize, dequantize, etc).

rdzhabarov:
screen shot 2018-07-05 at 11 54 27 am

nadavrot:
@rdzhabarov Thank you for doing this. It looks like a good solution.

PR #1217, opened 2018-07-05
[cpu] Improve the ability to debug JITted code

opti-mix:
* Add the option --dump-jit-symbol-info to dump addresses of JITted functions
  This option can be very useful e.g. for debugging crashes inside the code produced by the CPU JIT.

* Add debugger events listener
   This allows to the JIT to inform the debugger about the dynamically generated object file. Using this information, the debugger can show proper stack traces, etc.
   This feature currently works only on Linux, because LLVM's MachO RuntimeDyld does not support debug info loading yet.

The main motivation for this PR was the debugging experience last week, when @beicy was hunting a bug in the JITted code. This PR should make it easier to debug the JITted code.

### DIFF
--- a/lib/Backends/CPU/GlowJIT.cpp
+++ b/lib/Backends/CPU/GlowJIT.cpp
@@ -15,12 +15,72 @@
  */
 
 #include "GlowJIT.h"
+#include "CommandLine.h"
+#include "llvm/Object/SymbolSize.h"
 
 using GlowJIT = llvm::orc::GlowJIT;
 
+namespace {
+/// An option to enabling the dump of the symbol information for the JITted
+/// functions. It dumps e.g. the names of the functions, their start addresses
+/// and their end addresses.
+static llvm::cl::opt<bool> dumpJITSymbolInfo(
+    "dump-jit-symbol-info",
+    llvm::cl::desc("Dump the load addresses and sizes of JITted symbols"),
+    llvm::cl::init(false), llvm::cl::cat(CPUBackendCat));
+
+/// This is a callback that is invoked when an LLVM module is compiled and
+/// loaded by the JIT for execution.
+class NotifyLoadedFunctor {
+  /// JIT object whose events are received by this listener.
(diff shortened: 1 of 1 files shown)


lib/Backends/CPU/GlowJIT.h
```diff
@@ -46,6 +46,9 @@ class GlowJIT {
   const DataLayout DL_;
   RTDyldObjectLinkingLayer objectLayer_;
   IRCompileLayer<decltype(objectLayer_), SimpleCompiler> compileLayer_;
+  /// The listener for debugger events. It is used to provide debuggers with the
+  /// information about JITted code.
+  llvm::JITEventListener *dbgRegistrationListener = nullptr;
```
bertmaher: Could this be a member of `NotifyLoadedFunctor` instead?  Also, nit: end member name with _.

lib/Backends/CPU/GlowJIT.cpp
```diff
@@ -15,14 +15,81 @@
...
+    }
+  }
+
+public:
+  NotifyLoadedFunctor(GlowJIT *jit) : jit_(jit) {}
+  void operator()(llvm::orc::RTDyldObjectLinkingLayerBase::ObjHandleT,
+                  const llvm::orc::RTDyldObjectLinkingLayerBase::ObjectPtr &obj,
+                  const llvm::RuntimeDyld::LoadedObjectInfo &objInfo) {
+    auto loadedObj = obj->getBinary();
+    // Inform the debugger about the loaded object file. This should allow e.g.
```
rdzhabarov: nit: can you fix the second sentence?

lib/Backends/CPU/GlowJIT.h
opti-mix: Done

lib/Backends/CPU/GlowJIT.cpp
opti-mix: Done.

PR #1218, opened 2018-07-05
Support group arg for conv grad in CPU backend

beicy:


tests/unittests/gradCheckTest.cpp
```diff
@@ -199,11 +200,29 @@ static void gradientCheckGroupConv(size_t numInputChan, size_t group) {
...
+TEST(Network, gradientCheckDepthwiseConv) {
+  gradientCheckGroupConv(4, 4, BackendKind::Interpreter);
+}
+
+TEST(Network, gradientCheckGroupConv) {
+  gradientCheckGroupConv(4, 2, BackendKind::Interpreter);
+}
 
-TEST(Network, gradientCheckDepthwiseConv) { gradientCheckGroupConv(4, 4); }
+TEST(Network, gradientCheckConvCPU) {
```
bertmaher: Consider using a parameterized test fixture (`TEST_P`) for the different backends, like we use in OperatorTest and MLTest.  It can be a separate PR if you'd like.

tests/unittests/gradCheckTest.cpp
```diff
@@ -199,11 +200,29 @@ static void gradientCheckGroupConv(size_t numInputChan, size_t group) {
   performGradCheck(IP, result, A, Ex, &inputs, &outputs, 0.001, 0.004);
 }
 
-TEST(Network, gradientCheckConv) { gradientCheckGroupConv(1, 1); }
+TEST(Network, gradientCheckConv) {
```
rdzhabarov: can you use a similar technique as in operatorTest when we support multiple backends for tests?

tests/unittests/gradCheckTest.cpp
rdzhabarov: that was fast :)

rdzhabarov:
There is an exception while running under GCC: https://travis-ci.org/pytorch/glow/jobs/400583898

tests/unittests/gradCheckTest.cpp
```diff
@@ -26,6 +26,14 @@
 
 using namespace glow;
 
+class GradCheckBase : public ::testing::TestWithParam<BackendKind> {
+public:
+  ExecutionEngine IP_{GetParam()};
```
rdzhabarov: nit: IP seems to be coming from "interpreter", maybe call this backend_.

beicy: "IP" stands for Execution Engine in the code.  Will change it to EE_.

lib/Backends/CPU/libjit/libjit_conv.cpp
```diff
@@ -492,49 +492,58 @@ void libjit_convolution_grad_f(float *inG, const float *outG, const float *inW,
...
+                }
+
+                for (size_t c = 0; c < inCperG; c++) {
+                  inG[libjit_getXYZW(inWdims, n, (size_t)ax, (size_t)ay,
+                                     g * inCperG + c)] +=
+                      filterW[libjit_getXYZW(filterGdims, d, kx, ky, c)] * grad;
+                  filterG[libjit_getXYZW(filterGdims, d, kx, ky, c)] +=
+                      inW[libjit_getXYZW(inWdims, n, (size_t)ax, (size_t)ay,
+                                         g * inCperG + c)] *
+                      grad;
```
rdzhabarov: nit: fix format.

tests/unittests/gradCheckTest.cpp
```diff
@@ -196,22 +201,22 @@ static void gradientCheckGroupConv(size_t numInputChan, size_t group) {
   inputsH.initXavier(1, mod.getPRNG());
   outputsH.initXavier(1, mod.getPRNG());
 
-  performGradCheck(IP, result, A, Ex, &inputs, &outputs, 0.001, 0.004);
+  performGradCheck(EE_, result, A, Ex, &inputs, &outputs, 0.001, 0.004);
 }
 
-TEST(Network, gradientCheckConv) { gradientCheckGroupConv(1, 1); }
+TEST_P(GradCheck, gradientCheckConv) { gradientCheckGroupConv(1, 1, EE_); }
```
rdzhabarov: are not all those groupConv tests? (names of the tests seems odd, is it really gradientCheckDepthwiseConv)

lib/Backends/CPU/libjit/libjit_conv.cpp
beicy: Actually, it is the format after running format.sh

rdzhabarov: :) haha.

tests/unittests/gradCheckTest.cpp
beicy: The test are updated by https://github.com/pytorch/glow/commit/361dc5a156496f1d25083cec70f82335cdf6c673#diff-1c1a40d1f65512890dcbc5053e5dd5a3. 
I think the first one is just keep the original test (non-group). The second one is called DepthwiseConv, for depethwise conv, all I know so far is that the number of group should be equal to the input channel.  I need to do more research on this part.

nadavrot:
Nice!   +1 on @artemrakhov 's suggestion to benchmark the perf.

PR #1220, opened 2018-07-05
Unbreak/cleanup fr2en 

jfix71:
Some places used the Node instead of NodeValue to get the type and dims. This issue was uncovered due the opti recently landed by @ZchiPitt , as the inputs were previously Reshapes but are now TopKs. Also small cleanup to use `getIndices()` directly instead of the implicit NodeValue via `{topk, 1}`.


bertmaher:
If you want to add said unit test in a separate PR that's fine, given that this is unbreaking things.

PR #1221, opened 2018-07-05
Non-Square padding for OpenCL backend

beicy:
So far, the first half of the output values are correct, but the rest are all zero. The test is NonSquarePaddingConvolution in OperatorTest.cpp

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -394,6 +394,7 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
   auto odim = ShapeNCHW(CC->getDest()->getType()->dims());
   auto idim = ShapeNCHW(CC->getSrc()->getType()->dims());
   auto fdim = ShapeNCHW(CC->getFilter()->getType()->dims());
+  PaddingTLBR pads(CC->getPads());
   // Create options for compiling the program.
   // Don't use names M, N, K as they are defined in precompiled headers.
 
@@ -405,8 +406,8 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
   // Parameters for kernel size, padding and stride
   addIntOption(options, "v_k_0", CC->getKernel());
   addIntOption(options, "v_k_1", CC->getKernel());
-  addIntOption(options, "v_p_0", CC->getPad());
-  addIntOption(options, "v_p_1", CC->getPad());
+  addIntOption(options, "v_p_0", pads.top);
+  addIntOption(options, "v_p_1", pads.left);
   addIntOption(options, "v_s_0", CC->getStride());
   addIntOption(options, "v_s_1", CC->getStride());
 
@@ -499,6 +500,12 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
   std::vector<size_t> global = {((N_FW_ - 1) / fw_div_N + 1) * fw_wgs0,
                                 ((M_FW_ - 1) / fw_div_M + 1) * fw_wgs1,
                                 idim.n * group};
+
+  // The global work for the N dimension (which covers all pixels) should be big
+  // enough to cover all outputs for the correctness.
+  if (global[0] * fw_div_N < odim.h * odim.w) {
+    global[0] = (odim.h * odim.w - 1) / fw_div_N + 1;
+  }
 
   enqueueKernel(commands_, kernel, deviceId_, global, local, kernelLaunches_);
 }
@@ -765,10 +772,10 @@ void OCLBackend::doForwardPass() {
 
       auto odim = ShapeNHWC(CC->getDest()->getType()->dims());
       auto idim = ShapeNHWC(CC->getSrc()->getType()->dims());
-
+      auto pads = PaddingTLBR(CC->getPads());
       setKernelArg<cl_uint>(kernel, numArgs + 1, CC->getKernel());
       setKernelArg<cl_uint>(kernel, numArgs + 2, CC->getStride());
-      setKernelArg<cl_uint>(kernel, numArgs + 3, CC->getPads()[0]);
+      setKernelArg(kernel, numArgs + 3, pads);
       setKernelArg(kernel, numArgs + 4, odim);
       setKernelArg(kernel, numArgs + 5, idim);
       setKernelArg(kernel, numArgs + 6,
@@ -795,10 +802,10 @@ void OCLBackend::doForwardPass() {
       auto destGradDim = ShapeNHWC(destGrad->dims());
       auto srcDim = ShapeNHWC(src->dims());
       auto filterGradDim = ShapeNHWC(filterGrad->dims());
-
+      auto pads = PaddingTLBR(CG->getPads());
       setKernelArg<cl_uint>(kernel, numArgs + 1, CG->getKernel());
       setKernelArg<cl_uint>(kernel, numArgs + 2, CG->getStride());
-      setKernelArg<cl_uint>(kernel, numArgs + 3, CG->getPads()[0]);
+      setKernelArg(kernel, numArgs + 3, pads);
       setKernelArg(kernel, numArgs + 4, srcDim);
       setKernelArg(kernel, numArgs + 5, destGradDim);
       setKernelArg(kernel, numArgs + 6, filterGradDim);
--- a/lib/Backends/OpenCL/Transforms.cpp
+++ b/lib/Backends/OpenCL/Transforms.cpp
@@ -25,27 +25,6 @@ using llvm::dyn_cast;
 
 using namespace glow;
 
-/// Optimize regular Convolution nodes (that use NHWC) into an
-/// OCL-backend-specific convolution that uses NCHW.
-static Node *convertConvToOCLConv(ConvolutionNode *CN, Function *F) {
-  // Convert filter and input from NHWC (Glow's default) into NCHW.
-  auto *NI = F->createTranspose("conv.input", CN->getInput(), NHWC2NCHW);
-  auto *NF = F->createTranspose("conv.filter", CN->getFilter(), NHWC2NCHW);
-
-  auto dimsNHWC = ShapeNHWC(CN->getResult().getType()->dims());
-  auto dimsNCHW = {dimsNHWC.n, dimsNHWC.c, dimsNHWC.h, dimsNHWC.w};
-  auto outTy = F->getParent()->uniqueType(ElemKind::FloatTy, dimsNCHW);
-
-  // For OpenCL, we currently only support equal padding.
-  PaddingTLBR pads(CN->getPads());
-  assert(pads.equalPadding() && "OpenCL Conv requires equal padding");
-  auto *NC = F->addNode(new OCLConvolutionNode(CN->getName(), outTy, NI, NF,
-                                               CN->getBias(), CN->getKernel(),
-                                               CN->getStride(), pads.top));
-  auto NR = F->createTranspose("conv.result", NC, NCHW2NHWC);
-  return NR;
-}
-
 /// Perform OpenCL specific post-lowering graph transformation.
 bool OCLBackend::transformPostLowering(Function *F, CompilationMode mode) {
   // NCHW transformation is not supported in training mode yet, because of some
@@ -56,9 +35,7 @@ bool OCLBackend::transformPostLowering(Function *F, CompilationMode mode) {
   bool changed = false;
   for (auto &node : F->getNodes()) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
-      // Note: Using OCL-specific convolution conversion here since OCL only
-      // currently supports equal padding.
-      auto *NR = convertConvToOCLConv(CN, F);
+      auto *NR = convertConvToNCHWConv<OCLConvolutionNode>(CN, F);
       NodeValue(&node, 0).replaceAllUsesOfWith(NR);
       changed = true;
       continue;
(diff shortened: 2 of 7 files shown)


lib/Backends/OpenCL/kernels.cl
```diff
@@ -637,14 +637,14 @@ __kernel void convolutiongradW(__global void *mem,
                            cl_uint32_t biasGrad,
                            cl_uint32_t filterSize,
                            cl_uint32_t stride,
-                           cl_uint32_t pad,
+                           __global unsigned *pads,
```
opti-mix: And here.

tools/ClassGen/Backends/OpenCL/OpenCLSpecificInstrs.h
```diff
@@ -23,7 +23,7 @@ BB.newBackendSpecificInstr("OCLConvolution")
     .addOperand("Bias", OperandKind::In)
     .addMember(MemberType::SizeT, "Kernel")
     .addMember(MemberType::SizeT, "Stride")
-    .addMember(MemberType::SizeT, "Pad")
+    .addMember(MemberType::VectorSizeT, "Pads")
```
opti-mix: This is OK. 

BWT,  I noticed that no Group parameter was added to OpenCL convolutions, when we added group convolutions support. May be we should add this parameter for the sake of completeness. In a different PR, of course.

tests/unittests/OperatorTest.cpp
```diff
@@ -2107,7 +2107,7 @@ TEST_P(InterpAndCPU, GroupConvolution) {
   EXPECT_FLOAT_EQ(result.at({0, 1, 0, 5}), (13 + 14 + 15 + 16) * 100000);
 }
 
-TEST_P(InterpAndCPU, NonSquarePaddingConvolution) {
+TEST_P(Operator, NonSquarePaddingConvolution) {
```
opti-mix: Please add comments about what this test does and how.

tests/unittests/OperatorTest.cpp
```diff
@@ -2139,13 +2139,24 @@ TEST_P(InterpAndCPU, NonSquarePaddingConvolution) {
...
+  auto filter1 =
+      mod_.createVariable(ElemKind::FloatTy, {2, 2, 2, 1}, "filter1");
+  auto FH1 = filter1->getHandle();
+  for (size_t i = 0; i < 2 * 2 * 2; i++) {
+    FH1.raw(i) = pow(2.0, i);
+  }
+  auto *zeroBias1 = mod_.createVariable(ElemKind::FloatTy, {2}, "bias1");
+  zeroBias1->getPayload().zero();
+
+  CN = F_->createConv("Conv1", input1, filter1, zeroBias1, outTy, 2, 1,
```
opti-mix: You probably want to use a different function for it. F_ was already used for the first part of the test.

tools/ClassGen/Backends/OpenCL/OpenCLSpecificInstrs.h
beicy: Will work on group parameter.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -22,6 +22,13 @@ typedef struct {
   cl_uint64_t w; // Width
 } ShapeNCHW;
 
+typedef struct {
```
opti-mix: Please don't use size_t in OpenCL kernels. Its size is target dependent. Use cl_uint64_t instead, like other structs in this file.

opti-mix:
Looking good now. Have you tested locally this change does not break any other OpenCL functionality? You should try to run `run.sh` with `-opencl` option. Also, please check that mnist, cifar10, ptb, etc are still working with OpenCL.

We need to be careful here, as OpenCL is not tested on the CI.

beicy:
@opti-mix I've run ./tests/images/run.sh with -opencl and the result is the same with the current repo(i.e. w/o my change): shufflenet and zfnet512 model failed, and others passed.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -500,6 +501,14 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
                                 ((M_FW_ - 1) / fw_div_M + 1) * fw_wgs1,
                                 idim.n * group};
 
+  if (fw_wgs0 == 1 && fw_wgs1 == 1) {
+    // Special case for the execution on Apple's CPU OpenCL backend, where
```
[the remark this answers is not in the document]
opti-mix: It is needed for correctness.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -500,6 +501,14 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
...
+    // fw_wgs0 = 1 and fw_wgs1 = 1. The global work for the N dimension should
+    // be big enough to cover all outputs.
+    while (global[0] * fw_div_N < odim.h * odim.w) {
```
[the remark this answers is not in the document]
opti-mix: May be. Just keep in mind that all the operands are integers here. So, one needs to be very careful with division if you want to replace the loop by an expression.

lib/Backends/OpenCL/OpenCL.cpp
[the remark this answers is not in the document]
opti-mix: @beicy Please update the comment to reflect the fact that it is needed for correctness.

lib/Backends/OpenCL/OpenCL.cpp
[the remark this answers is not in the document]
rdzhabarov: this should work: (h * w + fw_div_N - 1)/ fw_div_N

lib/Backends/OpenCL/OpenCL.cpp
artemrakhov: I don't understand this comment and following code. Why is it specific to Apple? Why do we only touch dimension N, but not M?
Can you simplify loop by using division?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -765,10 +774,10 @@ void OCLBackend::doForwardPass() {
```
artemrakhov: Isn't this a dead code branch? Because we replace Conv with OCLConvolution? If yes, then I suggest to delete this codepath first, maybe in separate PR.

tests/unittests/OperatorTest.cpp
```diff
@@ -2190,13 +2195,24 @@ TEST_P(InterpAndCPU, NonSquarePaddingConvolution) {
       IH1.at({0, i, j, 0}) = i * 4 + j - 2 + 1;
     }
 
-  CN = F_->createConv("Conv1", input1, filter, zeroBias, outTy, 2, 1,
-                      {0, 0, 0, 0}, 1);
-  S = F_->createSave("save1", CN);
-  EE_.compile(CompilationMode::Infer, F_);
+  auto filter1 =
```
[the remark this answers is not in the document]
beicy: No. For OpenCL, there is an optimization (forget which one, need to re-check it :P)  makes the var "filter" to be an unused var and deleted.

tests/unittests/OperatorTest.cpp
```diff
@@ -2190,13 +2195,24 @@ TEST_P(InterpAndCPU, NonSquarePaddingConvolution) {
...
+  Function *refF = mod_.createFunction("mainRef");
+  CN = refF->createConv("Conv1", input1, filter1, zeroBias1, outTy, 2, 1,
+                        {0, 0, 0, 0}, 1);
+  S = refF->createSave("save1", CN);
+  EE_.compile(CompilationMode::Infer, refF);
   EE_.run({}, {});
   Tensor &result1 = S->getVariable()->getPayload();
 
+  for (size_t i = 0; i < 4 * 8 * 2; i++)
+    EXPECT_EQ(result.getHandle().raw(i), result1.getHandle().raw(i));
```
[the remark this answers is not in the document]
beicy: Forget to delete this one -- used for debugging.

jfix71:
You should be able to remove `convertConvToOCLConv()` entirely and replace it with a templatized call into `convertConvToNCHWConv()` from LayoutConverter.h, very similar to `convertPoolToNCHWPool()`

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
artemrakhov: So, if filter has 2 users it is considered to be an unused var? This sounds like a bug. Could you please check which optimization does this and why?

lib/Backends/OpenCL/OpenCL.cpp
opti-mix: It is not dead. It is still used for training.

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
opti-mix: > For OpenCL, there is an optimization (forget which one, need to re-check it :P) makes the var "filter" to be an unused var and deleted.

I'd be interested to hear what optimization does this. AFAIK, there are no OpenCL-specific optimizations except for Transforms.cpp, which changes the layout from NHWC to NCWH for convolutions and pooling operators.

lib/Backends/OpenCL/OpenCL.cpp
artemrakhov: I understand that ConvolutionGradInst is still used.
But forward ConvolutionNode is converted to OCLConvolutionNode, and then to OCLConvolutionInst, isn't it?

opti-mix: No, ConvolutionNode is not replaced by OCLConvolutionNode in the the training mode. There are some issues with this, so this transformation is disabled in this case.

rdzhabarov: yeah, i was looking at the same recently: https://github.com/pytorch/glow/blob/master/lib/Backends/OpenCL/Transforms.cpp#L53

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
beicy: @artemrakhov @opti-mix : for OpenCL, after the transform, the original "conv->save" is now :name : Conv
Input : float<1 x 4 x 4 x 1>
Filter : float<2 x 2 x 2 x 1>
Bias : float<2>
Kernel : 2
Stride : 1
Pads : [0, 2, 1, 3]
Group : 1
users : 0
Result : float<1 x 4 x 8 x 2>

Save
name : save
Input : float<1 x 4 x 8 x 2>
Output : float<1 x 4 x 8 x 2>
users : 0

Transpose
name : conv_input
Input : float<1 x 4 x 4 x 1>
Shuffle : [0, 3, 1, 2]
users : 1
Result : float<1 x 1 x 4 x 4>

Transpose
name : conv_filter
Input : float<2 x 2 x 2 x 1>
Shuffle : [0, 3, 1, 2]
users : 1
Result : float<2 x 1 x 2 x 2>

OCLConvolution
name : Conv1
Input : float<1 x 1 x 4 x 4>
Filter : float<2 x 1 x 2 x 2>
Bias : float<2>
Kernel : 2
Stride : 1
Pads : [0, 2, 1, 3]
users : 1
Result : float<1 x 2 x 4 x 8>

Transpose
name : conv_result
Input : float<1 x 2 x 4 x 8>
Shuffle : [0, 2, 3, 1]
users : 1
Result : float<1 x 4 x 8 x 2>

And the the first "conv" node (and the vars "conv_input", "conv_filter" etc.) has no user and be deleted as dead code.  The final layers are: 
Graph structure main:
Save
name : save
Input : float<1 x 4 x 8 x 2>
Output : float<1 x 4 x 8 x 2>
users : 0

OCLConvolution
name : Conv1
Input : float<1 x 1 x 4 x 4>
Filter : float<2 x 1 x 2 x 2>
Bias : float<2>
Kernel : 2
Stride : 1
Pads : [0, 2, 1, 3]
users : 1
Result : float<1 x 2 x 4 x 8>

Transpose
name : conv_result
Input : float<1 x 2 x 4 x 8>
Shuffle : [0, 2, 3, 1]
users : 1
Result : float<1 x 4 x 8 x 2>

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -500,6 +501,12 @@ void OCLBackend::executeConvolution(const OCLConvolutionInst *CC) {
                                 ((M_FW_ - 1) / fw_div_M + 1) * fw_wgs1,
                                 idim.n * group};
 
+  // The global work for the N dimension (which covers all pixels) should be big
+  // enough to cover all outputs for the correctness.
+  if (global[0] * fw_div_N < odim.h * odim.w) {
+    global[0] = (odim.h * odim.w - 1) / fw_div_N + 1;
```
rdzhabarov: that's the same as (w * h + fw_div_n - 1) / fwdivN :P, but looks good.

lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -58,7 +37,7 @@ bool OCLBackend::transformPostLowering(Function *F, CompilationMode mode) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
       // Note: Using OCL-specific convolution conversion here since OCL only
```
jfix71: This comment is stale and can be removed now that `convertConvToNCHWConv()` is used

lib/Backends/OpenCL/OpenCL.cpp
beicy: Yes, I just converted it from your expression :)

PR #1222, opened 2018-07-05
[cleanup] Make code more defensive to multiple outputs.

rdzhabarov:
Inspired by https://github.com/pytorch/glow/pull/1220.
Mostly to be defensive in case node will have multiple outputs.


rdzhabarov:
For these, I was just grepping. For the next batch, it's removal of a default value of the index on getType. There are a bunch of those, I'll look into those tomorrow.

nadavrot:
Should we remove the implicit conversion between NodeValue and Node?

jfix71:
@rdzhabarov Perhaps we could also remove the default value from `dims()` and `getElementType()`. I would imagine it would be no more painful than removing from `getType()`?

PR #1223, opened 2018-07-06
[OperatorTest] Add test for concat of operators with multiple results (TopK)

jfix71:
Tests the fix from https://github.com/pytorch/glow/pull/1220


jfix71:
@rdzhabarov Nope, not supported by OpenCL.

PR #1224, opened 2018-07-06
[cleanup] getType should have explicit index.

rdzhabarov:
There were several bugs exposed by missing index.
This PR changes the behavior to use an explicit index.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -231,7 +231,7 @@ public:
   ~Node() { releaseUsers(); }
 
   /// \returns the n'th result type of the node.
-  TypeRef getType(unsigned idx = -1) const;
+  TypeRef getType(unsigned idx) const;
 
   /// Methods that forward to the result type (that must be valid):
   /// @{
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -86,6 +86,9 @@ public:
   /// \returns the original training mode of the variable.
   TrainKind getTrainKind() const { return train_; }
 
+  /// \returns result type of the variable.
+  TypeRef getType() const { return Node::getType(0); }
+
   /// \returns the value used during initialization.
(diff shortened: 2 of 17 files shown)


jfix71:
Thanks for working on this!

It seems most if not all the places you have to provide index `0` are with Variables. I'm wondering if it might make sense to add a `getType()` in `Variable` that calls `Node::getType(0)`, since Variables can only have one result.

opti-mix:
In general, if we were calling/still want to call getType() without an index, we should probably assert if the node has more than one result.

nadavrot:
@opti-mix We had an assertion that checked that there is only one result if we use the default argument version.

rdzhabarov:
"I'm wondering if it might make sense to add a getType() in Variable that calls Node::getType(0), since Variables can only have one result."
@jfix71 right, that was for vars.
I think it's a good suggestion. Let me see what I can do here.

rdzhabarov:
Majority of the getType(0) is gone now.

rdzhabarov:
squashed second commit.

qcolombet:
> "I'm wondering if it might make sense to add a getType() in Variable that calls Node::getType(0), since Variables can only have one result."
> @jfix71 right, that was for vars.
> I think it's a good suggestion. Let me see what I can do here.

Actually, now that I am replacing the direct uses of Variable with Load, I have more place to change. I could play the same trick though (have a getType() on the Load node).

include/glow/Graph/Node.h
```diff
@@ -231,7 +231,7 @@ class Node : public Named,
   ~Node() { releaseUsers(); }
 
   /// \returns the n'th result type of the node.
-  TypeRef getType(unsigned idx = -1) const;
+  TypeRef getType(unsigned idx) const;
```
qcolombet: Node::getType has a special case for idx == -1.
This is not used anymore, right?
In other words, we should remove that code (I don't think this PR does that).

PR #1231, opened 2018-07-06
[Graph/fr2en] Put back Variable::TrainKind::None everywhere, and fix visibility on output

jfix71:
I added in a `createVariable()` that takes a required TrainKind parameter before an optional VisibilityKind, so that we don't need to specify the default VisibilityKind every time we're creating Variables with a non-default TrainKind. Then used it for fr2en, and added back Public visibility on output. Patch up for https://github.com/pytorch/glow/pull/1225

### DIFF
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -178,7 +178,8 @@ private:
     auto &mod = EE_.getModule();
     Variable *result =
         mod.createVariable(ElemKind::FloatTy, {langSize, EMBEDDING_SIZE},
-                           "embedding." + langPrefix.str());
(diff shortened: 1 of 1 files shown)


bertmaher:
IMO, as a user it's more annoying to have to remember the order of TrainKind and VisibilityKind.  

Although I could see an argument to have no default values, because I frankly can't remember what the defaults are most of the time, anyways.

jfix71:
Just to unblock this for now, I removed the `createVariable()` with swapped params. If we decide in the future to add it back in we can, or remove default params entirely, I don't have a strong opinion here.

PR #1233, opened 2018-07-07
[cleanup] Fix a couple of places missed by the "getType should have explicit index" cleanup

opti-mix:
Fix a couple of places that were missed by #1224


opti-mix:
The reason why this was missed in #1224 is most likely because we do not test OpenCL on CI and @rdzhabarov probably does not build locally with OpenCL backend enabled.

rdzhabarov:
Yes, my bad, thanks for quick fix!

PR #1235, opened 2018-07-09
[Graph] Combined lookup table variables for Tanh/Sigmoid

jfix71:
When we have many Tanh/Sigmoid lookup tables, we replicate the mapping data Variable many times. This PR reuses a single mapping data Variable for Tanh and Sigmoid.

To do so I added reserved names for these variables to the Module's `uniqueVariableNames_`, and then `createVariableWithReservedName()` to create a variable using this reserved name.

Any opinions on if this is a reasonable approach?


rdzhabarov:
This approach looks reasonable to me and similar in nature to static variables. One suggestion would be to rename from reserved to static.

include/glow/Graph/Graph.h
```diff
@@ -42,6 +42,9 @@ using FunctionList = std::list<Function *>;
 using VariablesList = std::list<Variable *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 
+constexpr char intTanhMappingVarName[] = "int_tanh_mapping";
```
opti-mix: I'd suggest moving these specific reserved names into the cpp file. We may need more later. No need to enumerate them here and force client to recompile if we change this set.

For the same reason, I'd move the `Module` constructor below into a cpp file as well.

jfix71:
I've updated reflecting comments. `createVariableWithReservedName()` is now `createStaticVariable()` -- I am not 100% positive I like the naming here, but happy to go with it if everyone prefers it, or open to other suggestions.

opti-mix:
@jfix71 Quick idea:  Is it "static" or may be "singleton" is a more precise name?

jfix71:
@opti-mix I like that much better -- updated.

lib/Graph/Graph.cpp
```diff
@@ -32,6 +32,16 @@ using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
+constexpr char intTanhMappingVarName[] = "int_tanh_mapping";
```
opti-mix: should they be `static`?

jfix71: Yep! Done.

lib/Graph/Graph.cpp
```diff
@@ -338,6 +348,23 @@ Variable *Module::createVariable(ElemKind T, llvm::ArrayRef<size_t> dims,
...
+                                float scale, int32_t offset,
+                                llvm::StringRef name, VisibilityKind visibility,
+                                Variable::TrainKind train, float val) {
+  for (auto *V : getVars()) {
+    if (V->getName() == name) {
+      assert(false && "Variable with this name has already been created.");
+    }
+  }
+
+  auto FT = uniqueType(T, dims, scale, offset);
```
opti-mix: Once we checked that a variable with a given name does not exist, can't we delegate to the usual `createVariable` instead of duplicating the logic here?

nadavrot:
Jordan, why does the graph, which is a generic container, need to know about sigmoid? This design looks a big strange to me. 

nadavrot:
Oh, I noticed that you changed the PR. The singeleton solution looks better because it moves the node specific code to the graph builder. 

lib/Graph/Graph.cpp
jfix71: `createVariable()` calls `addVar()` which uniquifies the name of the Variable. Here, we avoid the uniquify. The name has already been added in the Module constructor to `uniqueVariableNames_ ` to ensure it's not taken by another variable. So if we go through `createVariable()` the singleton variable will be given a different (unique) name, and future calls to `getVariableByName()` would not find the singleton.

opti-mix: Oh, I see.

lib/Graph/Graph.cpp
```diff
@@ -338,6 +348,23 @@ Variable *Module::createVariable(ElemKind T, llvm::ArrayRef<size_t> dims,
   return createVariable(FT, name, visibility, train, val);
 }
 
+Variable *
+Module::createSingletonVariable(ElemKind T, llvm::ArrayRef<size_t> dims,
```
rdzhabarov: can you also add a test that verifies this behavior.

jfix71:
@nadavrot We never change the original Variable, correct? We replace all uses of it with a new Variable. So if DCE removes the original Variable, then the next call to `createIntTanh()` would not find it via `getVariableByName()` and would re-create the singleton variable with the same name. If DCE does not remove the original variable then it would still be there unchanged, and `createIntTanh()` would find it correctly. So I don't think there's a bug.

jfix71:
For moving the table creation to the quantizer, I have https://github.com/pytorch/glow/pull/1236 open which moves the int lookup table creation to be done during lowering, to allow a backend to prevent lowering and do something else with the quantized tanh/sigmoid if it prefers. So with that change, the table would need to be created in lower instead of the quantizer. I think it makes the most sense to just keep it in Graph so all tanh/sigmoid lookup tables would all try to reuse the same table regardless of where it's being called from.

nadavrot:
Jordan, optimizations change the variable weights without checking if they are registered in some map or static variable that gives them special semantics such as the tanh table. 

The other problem that I mentioned is that the graph starts to depend on specific node semantic. 

If you want to unify variables why not let CSE or some other special pass do it?

include/glow/Graph/Graph.h
```diff
@@ -140,6 +140,15 @@ class Module final {
...
   ///@}
 
+  /// Variable builder for a singleton Variable with a reserved \p name,
+  /// reserved in the Module constructor. A singleton Variable with this name
+  /// should only be created once.
+  Variable *createSingletonVariable(
+      ElemKind T, llvm::ArrayRef<size_t> dims, float scale, int32_t offset,
+      llvm::StringRef name, VisibilityKind visibility = VisibilityKind::Private,
+      Variable::TrainKind train = Variable::TrainKind::Broadcast,
+      float val = 0.0);
```
qcolombet: What happens if you try to create a singleton variable twice?
It's easy to see in the code, but we shouldn't have to check the code to have the answer :).

jfix71: Absolutely can update the comment!

But, I was under the impression @nadavrot preferred to only do https://github.com/pytorch/glow/pull/1263 to deduplicate these Variables, and wanted to abandon this PR.

Of course, this PR would deduplicate these Vars at creation time and avoid some optimization time, so I do think it has value aside from https://github.com/pytorch/glow/pull/1263.

qcolombet:
You’re right. Sorry for the noise. We should close this one then, right?

Issue #1227, opened 2018-07-06
Refactor Backend interface


stoklund:
The `glow::Backend` interface is current storing a lot of state, and it needs to be refactored. We want to clearly separate the hardware abstraction layer from the different kinds of state we're storing.

Over in #1176 we want to compile multiple functions to run on different CPUs. In the future, we also want to support multiple GPUs and other accelerators, so we need a backend interface that separates state related to different functions and different execution units.

I'll elaborate on a design in the comments below.

stoklund:
# Prior art

One responsibility of the `Backend` class is to provide a common *hardware abstraction layer* interface. Each supported target architecture provides a subclass of `Backend` that implements virtual functions which specialize compilation and execution to the particular target.

Old-school re-targetable compilers used `#ifdef`s sprinkled all over their source code to specialize the compiler for a target architecture. The compiler could only be built for one architecture at a time. This meant that unit tests would only test the architecture you built, and you couldn't even detect syntax errors if they happened to be invisible in your current configuration. This made continuous integration very painful.

LLVM can be re-targeted at runtime. It uses a [`Target`](http://llvm.org/doxygen/classllvm_1_1Target.html) class which has virtual functions that provide further information about a target architecture. LLVM's target configuration system is more complex than Glow needs.

The Cretonne code generator uses a [`TargetIsa`](https://docs.rs/cretonne/0.13.0/cretonne/prelude/codegen/isa/trait.TargetIsa.html) trait which encapsulates both the target architecture *and* any compiler flags and settings that can affect the code generation. This means that the generated code is a pure function of the input IR and the `TargetIsa` instance. No secret command line flags can affect the code generation and cause hard-to-reproduce bugs.

Common to the LLVM and Cretonne designs is that their `Target` and `TargetIsa` objects are constant once they have been created. All the virtual methods are declared as `const`. This means that one target object can be reused for multiple (concurrent) compilations, and the state related to target configuration is clearly separated.

stoklund:
# Use cases

To inform the design, it's useful to look at a couple use cases for Glow. These all assume that Glow is used as a JIT compiler, i.e. we're not concerned with cross compilation or saving compiled models to disk.

## Parallel CPU inference

Running parallel inferences with a single fixed graph on a multicore CPU:

- Compile graph once.
- Execute compiled graph on many threads.

## Parallel inference on NUMA CPU

Same as above, but running on a multi-socket NUMA server:

- Compile graph once.
[...]

PR #1238, opened 2018-07-09
[cpu] Factor out BundleSaver

bertmaher:
A step towards #1227.  Saving bundles adds a bunch of stuff to CPUBackend's API, so it's hard to tell what can be safely split out.  It's actually quite nicely self-contained, so this diff is mostly a bunch of code motion.


opti-mix:
@bertmaher Looking good. Have you checked that our bundle examples still work as expected after this change?

bertmaher:
Ah yes, meant to say so in the description.  I checked resnet50:
```
cd examples/bundles/resnet50/
LOADER=`pwd`/../../../build/Release/bin/image-classifier GLOW_SRC=`pwd`/../../../ make run
```

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -33,10 +37,10 @@
 #include "llvm/Support/raw_ostream.h"
 
 using namespace glow;
+using llvm::StringRef;
```
opti-mix: You are trying to get rid of `llvm::StringRef` in the other PR, right?

bertmaher: Yup :-).  Will update this one once it lands.

PR #1241, opened 2018-07-09
[ee] Remove getIR from API

bertmaher:
To preserve debuggability, I added -dump-ir and -dump-ir-dag to ExecutionEngine, which also saves us re-implementing it in every example program :-).

I did have to work around the lack of getIR in one unit test, which is the most substantial code change in this diff.


soumyarooproy:
👍 Nice!

PR #1242, opened 2018-07-09
[IROptimizer] Enclosing intervals don't need to be strict

qcolombet:
Prior to this patch, we were considering that an interval A is
enclosed inside B iff B strictly started before A and ended after
A, including A's end point.

This patch removes the *strictly* part of the condition. Now, A is
enclosed inside B iff B starts before A and ends after A.

Drive by review.


opti-mix:
Good catch!

Issue #1234, opened 2018-07-09
Separate HAL, temporary state, and compiled functions


stoklund:
This is part of the `Backend` refactoring in #1227.

The goal is to have the `Backend` sub-classes represent the *hardware abstraction layer*. The instances hold no mutable state, and all actions are performed through const virtual functions. 

Proposed interface:

```c++
class Backend {
  virtual bool transformPreLowering(Function *F, CompilationMode mode) const;
  virtual bool transformPostLowering(Function *F, CompilationMode mode) const;
  virtual bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const;
  virtual bool shouldLower(const Node *N) const;
  virtual bool shouldShareBuffers() const;

  /// Compile an IR function, consuming it in the process.
  /// Returns a new object representing the compiled function.
  virtual unique_ptr<CompiledFunction> compileFunction(unique_ptr<IRFunction>) const;
}

class CompiledFunction {
  virtual void execute() =0;
}
```

Refactoring steps:

- Change `transformPreLowering` and friends to be `const`.
- Create a `CompiledFunction` sub-class per backend which holds the result of compiling a function.
- Move remaining temporary compilation state out of the `Backend` sub-classes into a locally defined class that only exists during `compileFunction()`.
- Remove the `IRFunction ` pointer from backend instances and the `createBackend()` function.

At this stage, the `CompiledFunction` sub-classes will still reference the original module, and their `execute` method can modify variables in the module.

bertmaher:
I believe this is done; Backend is now a pure stateless abstraction layer.

PR #1243, opened 2018-07-09
[cpu] Make codegen not retain state

bertmaher:
We don't need to hang on to LLVMIRGen or AllocationsInfo after generating code.  It turns out it's fairly easy to make these stack allocated in `init()`.


opti-mix:
Nice! But there are still a couple of fields, that should better be hidden from the include file.

I wonder if we could use something like a pimpl idiom (https://en.cppreference.com/w/cpp/language/pimpl) to hide all of the backends state in the implementation files. WDYT?

stoklund:
@opti-mix When #1234 is implemented, the `Backend` sub-classes should not contain any remaining mutable state. I think it can all be on the stack of the `compileFunction()` method.

PR #1244, opened 2018-07-09
[IROptimizer] Teach the share buffers opti how to propagate copies

### LINKED ISSUE #1240: shareBuffers fails to merge chain of copies
As found in #1209, our shareBuffers optimization fails to merge chain of copies.
The problem is illustrated with the test case in #1237.

Copy/pasting the analysis made in #1209:
In a nutshell, our buffer optimization only looks at merging one segment at a time. In the case of the lowering of the load node, we have for the first time potentially more than one copy on a chain to a use (e.g., with the lowering of reshape(load)) .
Now to illustrate the problem with chain of copy, let us consider:
1: a = copy b
2: c = copy a
3: = c
We have the following segments:
a = (1, 2); b = (-,1); c (2, 3)

First we merge c and a:
1: a = copy b
2: a = copy a
3: = a
Now the segments look like:
a = (1, 2), (2, 3); b = (-,1)

When considering merging a with b, we only consider updating the segment of a that encloses the segment of b, thus (1,2).
The updated code looks like:
1: b = copy b
2: a = copy b
3: = a
We don't update the old c segment and end up with a copy in 2 now.

Interestingly, if you delete the useless copies as you go, you actually end up with incorrect code, e.g., for the previous example:
1: b = copy b
---2: a = copy b---- deleted
3: = a <--- a never defined

This is because when we update the instruction, we don't preserve the last definition of the outgoing variable and given we don't update the uses not in the segment, they potentially reference an undefined variable.

It looks like it works just by accident but maybe it was intended this way?

Long story short we need to lift the one-segment-at-a-time limitation (maybe by merging the segments?).

qcolombet:
When coalescing two interval through copies, merge then into one big
interval. This is legal because the buffers have the same value for
that interval.
Given our share buffers optimization only considers one interval at
a time, this effectively teaches it how to consider intervals that
have been copy-coalesced.
    
This fixes #1240.

### DIFF
--- a/lib/Optimizer/IROptimizer.cpp
+++ b/lib/Optimizer/IROptimizer.cpp
@@ -781,24 +781,6 @@ static void eraseInstructions(IRFunction &M,
 /// outside.
 static bool isObservable(Value *V) { return isa<WeightVar>(getOrigin(V)); }
 
-/// Substitute all uses of \p oldBuffer by \p newBuffer inside the
-/// \p oldInterval.
-static void reuseBufferInsideInterval(
-    Value *oldBuffer, Value *newBuffer, Interval oldInterval, IRFunction &M,
-    const LiveIntervalsInstructionNumbering &instrNumbering,
-    Intervals &oldIntervals, Intervals &newIntervals) {
-  DEBUG_GLOW(llvm::dbgs() << "\n\nReuse buffers: use buffer of "
-                          << newBuffer->getName() << " as a buffer for "
-                          << oldBuffer->getName() << "\n"
-                          << "in live interval " << oldInterval << "\n");
-  // Replace oldBuffer with newBuffer.
-  replaceAllUsesInsideIntervalWith(oldBuffer, newBuffer, oldInterval, M,
-                                   instrNumbering);
-  // oldInterval does not belong to oldIntervals anymore. It should belong
-  // to newIntervals now.
-  moveInterval(oldIntervals, newIntervals, oldInterval);
-}
-
 /// A helper class for performing a sharing of buffers used by a given
 /// instruction.
 class BufferSharingOptimizer {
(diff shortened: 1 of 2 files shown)


qcolombet:
Note: The first two commits are respectively covered by #1237 and #1242.

opti-mix:
@qcolombet Very nice! The fix is rather small and elegant! I'm happy it didn't require any major re-factoring.

Questions/requests:
- Have you run all of our run.sh and other big tests after your change to see if it breaks anything? CI won't catch it. We need to do it locally.
- I'm not very happy with the test. As far as I can see, you took it as is from a real example. But it seems to be too long and noisy as it contains a lot of unrelated nodes. Could you try to shorten and simplify it?

qcolombet:
> Have you run all of our run.sh and other big tests after your change to see if it breaks anything? CI won't catch it. We need to do it locally.

I've already done the run.sh and everything was fine. Do you have any other big tests I should try?

> Could you try to shorten and simplify it?

I may be able to get rid of all the cruft but one tensor insert. Renaming the variables in something sensible should help too :).

opti-mix:
> Do you have any other big tests I should try?

Our big standalone tests: ptb, mnist, cifar10, fr2en, etc

qcolombet:
Thanks for the review @opti-mix !

Could you approve the two pre-requisite PR #1237 and #1242? (Which you already review indirectly as the first two commits of this PR).

Also, ran mnist, cifar10, fr2en, ptb, and char-rnn, results looked good. (Would be nice if the PR was doing that... my laptop doesn't like running those :P)

opti-mix:
> Could you approve the two pre-requisite PR #1237 and #1242? (Which you already review indirectly as the first two commits of this PR).

Done.

> Also, ran mnist, cifar10, fr2en, ptb, and char-rnn, results looked good. (Would be nice if the PR was doing that... my laptop doesn't like running those :P)

You can always try to use `-cpu`. Usually it is quite a bit faster. But yes, if CI could do it, it would be much nicer ;-)

nadavrot:
Awesome 
@qcolombet. 

PR #1246, opened 2018-07-10
[travis] Make jobs instead of matrix.

rdzhabarov:
Matrix does not support conditional switch off for subjobs, which would be needed if we want to run cron jobs (for time consuming runs of run.sh etc).
Currently, matrix capabilities are not used anyways.


nadavrot:
Cool

PR #1247, opened 2018-07-10
[unittest] Properly scope IRBuilder before invoking backend

bertmaher:
I was looking into making backends own their passed-in IRFunction (i.e. the unique_ptr interface in #1234 ) and hit a weird memory bug stemming from this issue. In these cases we invoke the backend before we've finished building our IR, since the dtor still needs to `deallocateActiveInstructions()`.  We could manually call that, but it still leaves us with a dangling pointer.

Also note the irOpt tests all call ::optimize on partially built IR for this same reason, but I don't think it's a problem there since the IR is still under construction anyways.


qcolombet:
Looks like one more argument for getting rid of the dealloc instructions :).

PR #1248, opened 2018-07-10
[backends] Refactor Backend

bertmaher:
Per #1234 and #1227 we want a cleaner separation of concerns from our Backends.  Supposedly commits within a PR can be reviewed individual, and there might be a few diffs here, so let's try it out.  First one up makes `init()` the recipient of an IRFunction, rather than stashing it in the Backend.


include/glow/Backends/Backend.h
```diff
@@ -84,7 +84,7 @@ class Backend {
 };
 
 /// Create a backend of kind \p kind, to run the IR function \p M.
-Backend *createBackend(BackendKind backendKind, IRFunction *M);
+Backend *createBackend(BackendKind backendKind);
```
rdzhabarov: do not forget to update the comment.

lib/Backends/Interpreter/Interpreter.h
```diff
@@ -41,7 +41,7 @@ class Variable;
 /// execute the instructions one at a time.
 class Interpreter final : public Backend {
   /// The Module that holds the IR. This does not own the module.
-  const IRFunction *F_;
+  std::unique_ptr<const IRFunction> F_;
```
rdzhabarov: Is that really a "Module"? Seems that module contains list of Functions* but not IRFunction.

include/glow/Backends/Backend.h
```diff
@@ -44,10 +44,10 @@ class Backend {
   virtual ~Backend() = default;
 
   /// Prepare the interpreter for execution of new code.
-  virtual void init() = 0;
+  virtual void init(std::unique_ptr<const IRFunction> IR) = 0;
```
stoklund: Since the `unique_ptr` transfers ownership, there's no need for `const`. If `init` owns the IR, it can do whatever it wants to it, including mutating it.

include/glow/Backends/Backend.h
```diff
@@ -44,10 +44,10 @@ class Backend {
...
 
   /// Save the bundle for a later standalone execution.
-  virtual void save(llvm::StringRef outputDir) {
+  virtual void save(std::unique_ptr<const IRFunction> IR, llvm::StringRef outputDir) {
```
stoklund: Presumably, this would become a method on `CompiledFunction`, but this is a fine intermediate step.

include/glow/Backends/Backend.h
opti-mix: Overall, I like the direction.

Question: This `init` API states that this backend becomes the owner of the IR. Can we ever have a situation where we want to compile the same IR using different backends? If so, this unique_ptr may make it rather difficult. Of course, we could always clone the IR and provide each backend with its own copy as an option...

lib/Backends/Interpreter/Interpreter.h
stoklund: Looks like a wrong comment got copy-pasted everywhere.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -162,30 +153,32 @@ void ExecutionEngine::generateIR(CompilationMode mode, Function *F) {
   }
 
   /// Prepare the IR container to handle our function.
-  IR_->setGraph(F);
+  auto IR = llvm::make_unique<IRFunction>(F);
```
stoklund: Is there a reason we use `llvm::make_unique` instead of `std::make_unique`?

include/glow/Backends/Backend.h
stoklund: The backends would want to do different things to the graph during lowering, right? It seems we would have to even clone the graph to use multiple backends.

lib/ExecutionEngine/ExecutionEngine.cpp
rdzhabarov: I thought we'd like to keep c++11 support, but if it's c++14 std:: makes sense.

bertmaher: I seem to remember the g++ version we use for coverage on Travis didn't support std::make_unique.  If we can bump that to something newer, though, I'm all in favor of c++14

rdzhabarov: I'm pushing this: https://github.com/pytorch/glow/pull/1251
I think that should resolve the issue you have ATM.

rdzhabarov: not so fast .... :) failing.

include/glow/Backends/Backend.h
bertmaher: I kind of liked the `const` here as an indication that the backend/init doesn't modify the IR, even though ownership would indicate that it might.  So, for instance, if you `dump` it beforehand you know it's not being changed inside the backend.  But I'm not too attached to it, so I'll change it if you feel strongly.

bertmaher:
Fixed the stale comments, thanks for pointing them out.

include/glow/Backends/Backend.h
bertmaher: I could maybe see unit tests wanting to run the same IR with different backends...but it's telling that we don't have any now that do :).
I agree with @stoklund that any real scenario would probably do different lowering so we'd want to clone it.

stoklund: I think it would be reasonable for a backend to do its own transformations on the IR. That might delay the introduction of a backend-specific IR, or avoid it altogether.

Such a backend should provide command line options to dump the IR after the modifications.

bertmaher: Sounds good.  Done!

lib/ExecutionEngine/ExecutionEngine.cpp
rdzhabarov: Issues are resolved now, I'll merge https://github.com/pytorch/glow/pull/1251
After that, you should be able to switch.

PR #1249, opened 2018-07-10
[unittests][graph] Add a test case for the uselist

qcolombet:
This test illustrates our issue when there are temporary NodeValues
around. Those temporary NodeValues contributes to the use list whereas
they should not.

The test reflects the current state of our use lists but we should fix
that eventually.

NFC.


qcolombet:
Thanks for the review!

PR #1250, opened 2018-07-10
[IROptimizer] Reuse copy propagation check and fix a bug

qcolombet:
Refactor the code to have one shared way of checking if we are in a copy
propagation case. This also fixes a bug where we were not checking that two
intervals had the same values and thus wrongly assuming this was a copy
propagation.

The added test case demonstrates the issue and would fail without the change in
the IR optimizer.


lib/Optimizer/IROptimizer.cpp
```diff
@@ -812,6 +812,14 @@ class BufferSharingOptimizer {
   /// instruction.
   Interval *destInterval_;
 
+  /// Check if instr_ is a copy propagation.
+  /// That is, instr_ is a copy and both the source and destination
+  /// have the same value on the related interval.
```
opti-mix: Use a more precise wording for the "same value"

tests/unittests/IROptTest.cpp
```diff
@@ -451,3 +451,65 @@ TEST(Optimizer, chainOfTwoForwardCopies) {
       instrs.begin(), instrs.end(),
       [](const Instruction &I) -> bool { return isa<CopyInst>(&I); }));
 }
+
+/// Make sure we don't coalesce two buffers that don't share the same
```
opti-mix: "Same" here :-D

PR #1252, opened 2018-07-10
[cleanup] Remove default index on dims.

rdzhabarov:
This is the followup from the https://github.com/pytorch/glow/pull/1224

Need to check OpenCL backend as well (looks good).

### DIFF
--- a/lib/Graph/Node.cpp
+++ b/lib/Graph/Node.cpp
@@ -80,11 +80,6 @@ void Node::setPredicate(const NodeValue &P) { predicate_ = P; }
 bool Node::hasPredicate() const { return predicate_.getNode(); }
 
 TypeRef Node::getType(unsigned idx) const {
-  if (idx == (unsigned)-1) {
-    assert(numRes_ == 1 && "Did not specify the result number for a node "
-                           "with multiple results.");
-    return types_[0];
-  }
   assert(idx < numRes_ && "Result number does not exist.");
   return types_[idx];
 }
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -539,7 +539,7 @@ static void mergeMatMul(Function *F) {
     size_t R = MM->getNthResult(0).dims()[1];
     size_t start = 0;
     for (auto *origMM : MMs) {
-      size_t H = origMM->dims()[0];
+      size_t H = origMM->getResult().dims()[0];
       auto *ex = F->createSlice("extract", MM, {start, 0}, {start + H, R});
       start += H;
       NodeValue(origMM).replaceAllUsesOfWith(ex);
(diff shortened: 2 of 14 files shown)


qcolombet:
Did you expose any bug while doing this?
If so, I am curious what we could have done to catch them earlier.

rdzhabarov:
no bugs, but there were some bugs with the getType() with the default index exposed earlier :)
those bugs would be caught during debug build if the network exposes certain structure, but it's tedious every time to fix those places. So, I went ahead and fixed all those now.

qcolombet:
> those bugs would be caught with the debug build run if the network exposes certain structure, but it's tedious every time to fix those places. So, I went ahead and fixed all those now.

Sounds good. I was mostly wondering if we could add some tests that would have catch the issues (same for getType).

lib/Graph/Node.cpp
```diff
@@ -80,11 +80,6 @@ void Node::setPredicate(const NodeValue &P) { predicate_ = P; }
 bool Node::hasPredicate() const { return predicate_.getNode(); }
 
 TypeRef Node::getType(unsigned idx) const {
-  if (idx == (unsigned)-1) {
-    assert(numRes_ == 1 && "Did not specify the result number for a node "
-                           "with multiple results.");
-    return types_[0];
-  }
```
qcolombet: That's the code I was talking about :).
Thanks!

Out-of-curiosity, do you know why the coverage bot, reports this as covered? (That was what I was checking when I asked you about our coverage checking.)

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1492,7 +1492,7 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       }
 
       auto *newReshape = F->createReshape(reshape->getName(),
-                                          rescale->getInput(), reshape->dims());
+                                          rescale->getInput(), reshape->getResult().dims());
```
jfix71: nit: make sure you run clang-format before merging

qcolombet:
> I'm wondering if we should change all the get/create Variables functions in the proto loaders to return Variable instead of Node.

We create loads now, so no we can’t + in theory went you call get you can get a random node (whatever was registered under that name)

qcolombet:
@rdzhabarov correct!

lib/Graph/Node.cpp
rdzhabarov: Not sure, will look into this.

lib/Optimizer/GraphOptimizer.cpp
rdzhabarov: damn :( forgot! will run real quick.

rdzhabarov: we need to have a clang format checker before merge :) (seperate story).

https://github.com/pytorch/glow/pull/1253

PR #1254, opened 2018-07-11
[backend] Factor out CompiledFunction from Interpreter

bertmaher:
Finally I think things are in shape to bite off #1234.  This PR creates the CompiledFunction interface, and modifies Backend::init (now known as `compile`) to produce it, and implements that interface in the Interpreter.  (All other tests and backends are not ported yet).

What do you all think?  (In retrospect maybe the interpreter was a weird place to start, because its "forward pass" is very heavyweight.  But hopefully that doesn't cloud things too much.)

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -18,7 +18,7 @@
 
 namespace glow {
 
-/// Interface containing state necessary to execute a compiled network.
+/// Interface for executing a compiled function.
 class CompiledFunction {
 public:
   /// Dtor.
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
+#define GLOW_BACKENDS_COMPILEDFUNCTION_H
+
+namespace glow {
+
+/// Interface containing state necessary to execute a compiled network.
+class CompiledFunction {
+public:
+  /// Dtor.
+  virtual ~CompiledFunction() = default;
+
+  /// Execute the network.
+  virtual void doForwardPass() = 0;
+};
+
+} // end namespace glow
+
+#endif // GLOW_BACKENDS_COMPILEDFUNCTION_H
--- a/lib/Backends/CPU/CPUFunction.h
+++ b/lib/Backends/CPU/CPUFunction.h
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_CPU_CPUFUNCTION_H
+#define GLOW_BACKENDS_CPU_CPUFUNCTION_H
+
+#include "GlowJIT.h"
+
+#include "glow/Backends/CompiledFunction.h"
+
+namespace glow {
+
+/// A function compiled for the CPU using LLVM.
+class CPUFunction final : public CompiledFunction {
+  /// The LLVM JIT engine. The jit must be initialized after the ctor
+  /// initializes the LLVM backends.
+  std::unique_ptr<llvm::orc::GlowJIT> JIT_;
+  /// This represents the heap, that stores the activations at runtime.
+  void *heap_;
+
+ public:
+  /// Ctor.
+  CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT, void *heap);
+
+  /// \name CompiledFunction interface
+  ///@{
+  ~CPUFunction() override;
+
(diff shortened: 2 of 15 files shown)


stoklund:
This PR could become a pretty large diff that needs to land atomically. If you want, you could avoid that:
- First define the CompiledFunction interface, but leave Backend as is.
- Start using CompiledFunction inside the init() function of each backend; save it as a member that doForwardPass() can use. This could be a PR per backend.
- Finally make the switch in the Backend and ExecutionEngine classes.

bertmaher:
Refactored the refactor.  This version should be incrementally correct.

include/glow/Backends/CompiledFunction.h
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
+#define GLOW_BACKENDS_COMPILEDFUNCTION_H
+
+namespace glow {
+
+/// Interface containing state necessary to execute a compiled network.
```
rdzhabarov: nit: maybe change wording slightly that implementation contains state etc. And this is an interface to execute a compiled network.

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "Interpreter.h"
+#include "InterpreterFunction.h"
```
rdzhabarov: Can you check if any of the headers below need to be removed? (probably glow/IR/IRUtils.h for example)

lib/Backends/Interpreter/Interpreter.h
```diff
@@ -16,6 +16,8 @@
 #ifndef GLOW_BACKENDS_INTERPRETER_INTERPRETER_H
 #define GLOW_BACKENDS_INTERPRETER_INTERPRETER_H
 
+#include "InterpreterFunction.h"
```
rdzhabarov: Tensor.h should be gone from here.

include/glow/Backends/CompiledFunction.h
opti-mix: Nit: "containing" is a bit misleading here, as this interface does not define any state at all. It just defines the API to be used for executing a compiled network. 

"network" in the comment and "Function" in the name also do not quite match.

include/glow/Backends/CompiledFunction.h
```diff
@@ -0,0 +1,33 @@
...
+class CompiledFunction {
+public:
+  /// Dtor.
+  virtual ~CompiledFunction() = default;
+
+  /// Execute the network.
+  virtual void doForwardPass() = 0;
```
opti-mix: We actually do not have a separate backward pass since a while, because we just generate it after the forward pass into the same IRFunction. And we use `doForwardPass` even for training these days. We should probably rename this method into `run` or `execute` or something like that. But this could be done here and elsewhere in a separate PR.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef GLOW_BACKENDS_CPU_CPUFUNCTION_H
+#define GLOW_BACKENDS_CPU_CPUFUNCTION_H
+
+#include "GlowJIT.h"
+
+#include "glow/Backends/CompiledFunction.h"
+
+namespace glow {
+
+/// A function compiled for the CPU using LLVM.
```
opti-mix: Nit: `function` in the comment is a bit ambiguous, because LLVM has a concept of functions too. May be it should say "Glow IR function" or "compiled network model"?

include/glow/Backends/CompiledFunction.h
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
+#define GLOW_BACKENDS_COMPILEDFUNCTION_H
+
+namespace glow {
+
+/// Interface containing state necessary to execute a compiled network.
+class CompiledFunction {
```
opti-mix: I've got a question thinking about this API.

I've been thinking about the following possible use-case: compile the function once, store the compiled function,  but then load and run it many times. I think this could be useful, as it does not require permanent re-compilations (which may require a lot of time and memory spent on compilations) and allows for caching/storing of compiled functions.

In its current form this interface does not provide any methods to set/change the "execution environment/configuration" of the compiled function. E.g. you cannot provide inputs or outputs and may be set any other kinds of parameters.

I guess it is always the task of the `ExecutionEngine` to set up those execution/configuration parameters, e.g. in its `run` or `runBatch` methods ? If this is the case, we may need to allow for creating of an  `ExecutionEngine` based on a compiled function (i.e. no IR or graph is available).

stoklund: I have a phase 2 in mind for #1227 which separates the compiled function from the runtime state. I haven’t written it up yet.

The idea is to have a `CompiledFunction::bind()` method which returns a third object: a bound function. This function is bound to concrete buffers for inputs and outputs, so multiple threads can use the same compiled function. Each thread would call bind(), and the execute() method would move from the compiled function to he bound function.

opti-mix: @stoklund OK, that makes sense! It will make the configuration part much cleaner.

lib/Backends/Interpreter/Interpreter.h
```diff
@@ -25,37 +27,20 @@
...
 class Interpreter final : public Backend {
-  /// The IR to be executed.
-  std::unique_ptr<IRFunction> F_;
-  /// Maps values to Tensors, that are owned by this class.
-  std::unordered_map<const Value *, Tensor *> tensors_;
-
-  /// Maps values to Tensors, that are *not* owned by this class.
-  std::unordered_map<const Value *, Tensor *> externalTensors_;
+  /// State necessary to execute a function using the interpreter.
+  std::unique_ptr<CompiledFunction> function_;
```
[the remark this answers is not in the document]
bertmaher: Fair point, but step 2 of this refactor is going to make init() return a `CompiledFunction` anyways, so I'm inclined to leave this alone for now.

include/glow/Backends/CompiledFunction.h
bertmaher: I'm confused.  We still have the ability to compile a function once and run multiple times. EE.run copies new input tensors and executes the compiled function; no recompilation is necessary.  Right?

bertmaher:
I think all comments should be addressed now.  I'll do s/doForwardPass/execute/g when I change the Backend interface.

include/glow/Backends/CompiledFunction.h
stoklund: @bertmaher I updated #1227 with more details, let's discuss it there.

If I understand correctly, the current scheme in `ExecutionEngine::run()` is not thread safe since data is copied into the variables which are owned by the module.

bertmaher: Ah, ok, sorry, I thought the concern was that this diff had broken the ability to run with new data without recompiling.

PR #1256, opened 2018-07-11
[ProtobufLoader] Make createVariable a protected API

qcolombet:
Users of ProtobufLoader should not need to create variables directly.

NFC.

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -106,13 +106,20 @@ protected:
   /// \returns A load to the newly created variable.
   /// \pre !hasNodeByName(name)
   Node *createAndRememberVariable(
-      llvm::StringRef name, Tensor &tensor,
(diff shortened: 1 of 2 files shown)


include/glow/Importer/ProtobufLoader.h
```diff
@@ -114,6 +114,13 @@ class ProtobufLoader {
   /// if no node has been registered with this name.
   Node *getNodeByNameOrNull(llvm::StringRef name) const;
 
+  /// Create a new variable \p name initialized with \p tensor.
+  /// \returns the newly created variable.
+  Node *createVariable(
```
rdzhabarov: why Tensor &tensor, but not const one? Seems like createVar should not change Tensor values.

qcolombet: Good catch! I'd say historical accident :).

Fixing that.

PR #1258, opened 2018-07-11
[travis] Use g++5.

rdzhabarov:
Relax a bit requirement on the compiler, g++5 should have required c++14 support.


rdzhabarov:
I'll put on hold this one until we decide on the c14 vs c11.

PR #1260, opened 2018-07-11
[GraphIR] Have LoadNode::getVariable return Variable

qcolombet:
Change the name of the input to Address so that we can use getVariable
to return the properly typed variable.

NFC.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -677,7 +677,7 @@ void ConcatNode::verify() const {
 
 void LoadNode::verify() const {
   // Loads are restricted to Variables.
-  assert(llvm::isa<Variable>(getVariable().getNode()));
+  assert(llvm::isa<Variable>(getAddress().getNode()));
 }
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -677,7 +677,7 @@ void ConcatNode::verify() const {
 
 void LoadNode::verify() const {
   // Loads are restricted to Variables.
-  assert(llvm::isa<Variable>(getVariable().getNode()));
+  assert(llvm::isa<Variable>(getAddress().getNode()));
```
rdzhabarov: why not directly: getVariable? (the method you introduced in the LoadNode class in NodeGen).

qcolombet: llvm::cast<Variable> actually already asserts if node is not a variable.
I thought that way it was more explicit. Don't have a strong opinion either way.

Longer term plan, I want xxx::verify to return bool, so that you can still verify in the debugger while built without assert and in that case the `isa` implementation is better.

That said, we don't have to do it now.

What do you prefer?

Note: I don't know exactly what I would put in the verify with getVariable  :) ((void)getVariable()?)

rdzhabarov: makes sense. This LG, thanks for the explanation.

PR #1261, opened 2018-07-11
Use CompiledFunction in OpenCL Backend

bertmaher:
The OpenCL backend references/mutates quite a bit of state during execution, so I'm essentially wrapping everything in the current backend inside a CompiledFunction.  At least this subdivision lets us maintain some separation between backend creation and function compilation.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.h
+++ b/lib/Backends/OpenCL/OpenCL.h
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_OPENCL_OPENCL_H
 
 #include "glow/Backends/Backend.h"
+#include "glow/Backends/CompiledFunction.h"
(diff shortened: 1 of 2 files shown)


lib/Backends/OpenCL/OpenCL.h
```diff
@@ -92,29 +92,18 @@ class OCLBackend final : public Backend {
 
 public:
   /// Ctor.
-  explicit OCLBackend();
+  explicit OpenCLFunction(std::unique_ptr<IRFunction> F);
 
-  /// @name Backend methods.
-  /// This is the implementation of the Backend interface.
+  /// \name CompiledFunction interface
```
rdzhabarov: should it be @name ... to place a comment on the whole block.

PR #1262, opened 2018-07-11
Non-square padding for Pooling

beicy:
For both AveragePool and MaxPool operators.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -286,11 +286,13 @@ void InterpreterFunction::fwdConvolutionGradInst(const ConvolutionGradInst *I) {
 //===----------------------------------------------------------------------===//
 template <class T>
 static void fwdPoolMax(Tensor *inW, Tensor *outW, Handle<size_t> *SXY,
-                       size_t filterSize, size_t stride, size_t pad) {
+                       size_t filterSize, size_t stride,
+                       llvm::ArrayRef<size_t> pads) {
   ShapeNHWC odim(outW->dims());
   ShapeNHWC idim(inW->dims());
   Handle<T> inHandle = inW->getHandle<T>();
   Handle<T> outHandle = outW->getHandle<T>();
+  PaddingTLBR pdim(pads);
 
   // For each input in the batch:
   for (size_t n = 0; n < odim.n; n++) {
@@ -298,9 +300,9 @@ static void fwdPoolMax(Tensor *inW, Tensor *outW, Handle<size_t> *SXY,
     // For each layer in the output tensor:
     for (size_t z = 0; z < idim.c; z++) {
       // For each convolution 'jump' in the input tensor:
-      ssize_t x = -ssize_t(pad);
+      ssize_t x = -ssize_t(pdim.top);
       for (size_t ax = 0; ax < odim.h; x += stride, ax++) {
-        ssize_t y = -ssize_t(pad);
+        ssize_t y = -ssize_t(pdim.left);
         for (size_t ay = 0; ay < odim.w; y += stride, ay++) {
           size_t maxX = x;
           size_t maxY = y;
@@ -329,7 +331,6 @@ static void fwdPoolMax(Tensor *inW, Tensor *outW, Handle<size_t> *SXY,
             }
           }
 
-          assert(!first && "Max value is uninitialized");
           outHandle.at({n, ax, ay, z}) = max_value;
 
           if (SXY) {
@@ -348,10 +349,10 @@ void InterpreterFunction::fwdPoolMaxInst(const PoolMaxInst *I) {
 
   if (inW->getType().isQuantizedType()) {
     fwdPoolMax<int8_t>(inW, outW, nullptr, I->getKernel(), I->getStride(),
-                       I->getPad());
+                       I->getPads());
   } else {
     fwdPoolMax<float>(inW, outW, nullptr, I->getKernel(), I->getStride(),
-                      I->getPad());
+                      I->getPads());
   }
 }
 
@@ -362,10 +363,10 @@ void InterpreterFunction::fwdPoolMaxWithXYInst(const PoolMaxWithXYInst *I) {
 
   if (inW->getType().isQuantizedType()) {
     fwdPoolMax<int8_t>(inW, outW, &SXY, I->getKernel(), I->getStride(),
-                       I->getPad());
+                       I->getPads());
   } else {
     fwdPoolMax<float>(inW, outW, &SXY, I->getKernel(), I->getStride(),
-                      I->getPad());
+                      I->getPads());
   }
 }
 
@@ -373,7 +374,7 @@ void InterpreterFunction::fwdPoolAvgInst(const PoolAvgInst *I) {
   ShapeNHWC odim(I->getDest()->dims());
   ShapeNHWC idim(I->getSrc()->dims());
 
-  auto pad = I->getPad();
+  PaddingTLBR pdim(I->getPads());
   auto filterSize = I->getKernel();
   auto stride = I->getStride();
   // Implement the avg pooling operation as defined here:
@@ -393,9 +394,9 @@ void InterpreterFunction::fwdPoolAvgInst(const PoolAvgInst *I) {
       // For each layer in the output tensor:
       for (size_t z = 0; z < idim.c; z++) {
         // For each convolution 'jump' in the input tensor:
-        ssize_t x = -ssize_t(pad);
+        ssize_t x = -ssize_t(pdim.top);
         for (size_t ax = 0; ax < odim.h; x += stride, ax++) {
-          ssize_t y = -ssize_t(pad);
+          ssize_t y = -ssize_t(pdim.left);
           for (size_t ay = 0; ay < odim.w; y += stride, ay++) {
             int32_t sum = 0;
 
@@ -434,9 +435,9 @@ void InterpreterFunction::fwdPoolAvgInst(const PoolAvgInst *I) {
     // For each layer in the output tensor:
     for (size_t z = 0; z < idim.c; z++) {
       // For each convolution 'jump' in the input tensor:
-      ssize_t x = -ssize_t(pad);
+      ssize_t x = -ssize_t(pdim.top);
       for (size_t ax = 0; ax < odim.h; x += stride, ax++) {
-        ssize_t y = -ssize_t(pad);
+        ssize_t y = -ssize_t(pdim.left);
         for (size_t ay = 0; ay < odim.w; y += stride, ay++) {
           float sum = 0;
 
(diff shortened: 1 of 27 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -329,7 +331,7 @@ static void fwdPoolMax(Tensor *inW, Tensor *outW, Handle<size_t> *SXY,
             }
           }
 
-          assert(!first && "Max value is uninitialized");
+          // assert(!first && "Max value is uninitialized");
```
bertmaher: Why is this assertion commented out now?

lib/Graph/Graph.cpp
```diff
@@ -455,29 +456,48 @@ ConvolutionNode *Function::createConv(llvm::StringRef name, NodeValue input,
...
+  return createPoolMax(name, input, kernel, stride, pads);
+}
+
+PoolAvgNode *Function::createPoolAvg(llvm::StringRef name, NodeValue input,
+                                     size_t kernel, size_t stride,
+                                     llvm::ArrayRef<size_t> pads) {
   ShapeNHWC idim = ShapeNHWC(input.dims());
-  assert(idim.w >= kernel && idim.h >= kernel &&
+  PaddingTLBR pdim(pads);
+  (void)pads;
```
bertmaher: This `(void)pads;` shouldn't be necessary since `pads` is used in non-debug code... right?  Oh, wait -- I think you want `(void)pdim`, since that's only used in the assertion.

tests/unittests/OperatorTest.cpp
```diff
@@ -2213,6 +2213,78 @@ TEST_P(Operator, NonSquarePaddingConvolution) {
...
+/// padding, while the second one has zero padding. The second pool op's input
+/// is the same as the first one's after-padding input. All other parameters of
+/// the two convs are the same.
+TEST_P(Operator, NonSquarePaddingAveragePool) {
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {1, 4, 4, 1}, "input");
+  auto IH = input->getHandle();
+  for (size_t i = 0; i < 4 * 4; i++) {
+    IH.raw(i) = i + 1;
+  }
+  auto *Pool = F_->createPoolAvg("Pool", input, 2, 1, {0, 2, 1, 3});
```
bertmaher: nit: lowercase `pool` since it's not an abbreviation.

tests/unittests/basicIRTest.cpp
```diff
@@ -131,7 +131,7 @@ TEST(IR, allInstrs) {
 
     builder.createCopyInst("", I1, I0);
     builder.createConvolutionInst("", I3, I1, F0, B0, 7, 2, {3, 3, 3, 3}, 1);
-    builder.createPoolMaxInst("", I4, I0, 7, 2, 3);
+    builder.createPoolMaxInst("", I4, I0, 7, 2, {3, 3, 3, 3});
```
bertmaher: Don't we still have a version that takes scalar pads?

tests/unittests/basicIRTest.cpp
```diff
@@ -155,7 +155,7 @@ TEST(IR, casting) {
     auto *input = bb.createWeightVar(ElemKind::FloatTy, {1, 224, 224, 3});
     auto *res = bb.createAllocActivationInst("sigmoid.res", input->getType());
     auto *sig = bb.createSigmoidInst("sigmoid", res, input);
-    auto *pool = bb.createPoolAvgOp(sig->getDest(), 7, 2, 3);
+    auto *pool = bb.createPoolAvgOp(sig->getDest(), 7, 2, {3, 3, 3, 3});
```
bertmaher: Same comment.

lib/Importer/ONNX.cpp
```diff
@@ -84,19 +84,6 @@ std::vector<size_t> getPads(const ArgumentDictionaryTy &dict) {
   return {0, 0, 0, 0};
 }
 
-size_t getPad(const ArgumentDictionaryTy &dict) {
-  if (dict.count("pads")) {
-    return getConstantArrayHead(dict.at("pads"));
-  }
-  if (dict.count("auto_pad")) {
-    auto padStr = loadStr(dict.at("auto_pad"));
-    if (padStr == "VALID")
```
bertmaher: Is this stuff about auto_pad and VALID relevant anymore?  It looks like we lost some functionality here.

tests/unittests/OperatorTest.cpp
```diff
@@ -2213,6 +2213,78 @@ TEST_P(Operator, NonSquarePaddingConvolution) {
...
+  SaveNode *S = F_->createSave("save", Pool);
```
jfix71: nit: I'd prefer consistently use either auto or [X]Node, otherwise it seems like there's a deliberate reason for the inconsistency. Same for the other test case as well.

tests/unittests/OperatorTest.cpp
```diff
@@ -2213,6 +2213,78 @@ TEST_P(Operator, NonSquarePaddingConvolution) {
...
+  Function *refF = mod_.createFunction("mainRef");
+  Pool = refF->createPoolAvg("Pool1", input1, 2, 1, 0);
+  S = refF->createSave("save1", Pool);
+  EE_.compile(CompilationMode::Infer, refF);
+  EE_.run({}, {});
+  Tensor &result1 = S->getVariable()->getPayload();
+
+  for (size_t i = 0; i < 4 * 8; i++)
+    EXPECT_EQ(result.getHandle().raw(i), result1.getHandle().raw(i));
+  EXPECT_TRUE(result.isEqual(result1));
```
jfix71: Why are you looping on the raw pointers and also calling `isEqual()`? I'd just drop the loop and use the latter. Same for the other test case as well.

tools/ClassGen/NodeGen.cpp
```diff
@@ -82,7 +82,7 @@ int main(int argc, char **argv) {
       .addInput("Input")
       .addMember(MemberType::SizeT, "Kernel")
       .addMember(MemberType::SizeT, "Stride")
-      .addMember(MemberType::SizeT, "Pad")
+      .addMember(MemberType::VectorSizeT, "Pads")
       .addResultFromCtorArg()
       .addGradient()
       .setDocstring("Performs a Max Pool operation on the Input given provided "
```
jfix71: nit: Update the comment here to say Pads? Same on PoolAvg below too.

tests/unittests/basicIRTest.cpp
beicy: Yes, we do have scalar pads, but that is for Node level.  At instruction level,  I don't think it is necessary to keep the scalar pad :)

tests/unittests/basicIRTest.cpp
beicy: Same, it creates an instruction.

lib/Importer/ONNX.cpp
beicy: Yes, in function getPads(). This "getPad" function was used to only support equal-padding. Now since non-square padding is supported, this one is replaced by "getPads", which wad added in conv non-square padding PR already.

tests/unittests/OperatorTest.cpp
beicy: You are right :)  The loop is for debugging purpose and I forgot to remove it...

lib/Backends/Interpreter/InterpreterNodes.cpp
bertmaher: I see.  Please delete the assert then (rather than leaving it commented).

I guess it still seems like a mistake if someone chooses filter/pad such that it pulls a bunch of zeros into the output activation... but probably a user error rather than an assert condition.

lib/Importer/ONNX.cpp
bertmaher: Ah, got it, thanks!  I was confused by the getPads() above in the Caffe2 importer :)

tools/ClassGen/NodeGen.cpp
beicy: Comments updated!

lib/Graph/Graph.cpp
```diff
@@ -455,29 +456,48 @@ ConvolutionNode *Function::createConv(llvm::StringRef name, NodeValue input,
 }
 
 PoolMaxNode *Function::createPoolMax(llvm::StringRef name, NodeValue input,
-                                     size_t kernel, size_t stride, size_t pad) {
+                                     size_t kernel, size_t stride,
+                                     llvm::ArrayRef<size_t> pads) {
   ShapeNHWC idim = ShapeNHWC(input.dims());
   assert(idim.w >= kernel && idim.h >= kernel &&
```
bertmaher: Shouldn't the assertion here look the same as for `createPoolAvg` below?

tests/unittests/OperatorTest.cpp
```diff
@@ -2213,6 +2213,74 @@ TEST_P(Operator, NonSquarePaddingConvolution) {
...
+/// padding, while the second one has zero padding. The second pool-op's input
+/// is the same as the first one's after-padding input. All other parameters
+/// of the two convs are the same.
+TEST_P(Operator, NonSquarePaddingMaxPool) {
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {1, 4, 4, 1}, "input");
+  auto IH = input->getHandle();
+  for (size_t i = 0; i < 4 * 4; i++) {
+    IH.raw(i) = i + 1;
+  }
+  auto *Pool = F_->createPoolMax("Pool", input, 2, 1, {0, 2, 1, 3});
```
jfix71: nit: variable name `Pool -> pool`

lib/IR/IRBuilder.cpp
```diff
@@ -55,13 +55,15 @@ void IRBuilder::deallocateActiveInstrs() {
 //===----------------------------------------------------------------------===//
 //                        High level operators.
 //===----------------------------------------------------------------------===//
-PoolMaxWithXYInst *IRBuilder::createPoolMaxWithXYOp(Value *input, size_t kernel,
-                                                    size_t stride, size_t pad) {
+PoolMaxWithXYInst *
+IRBuilder::createPoolMaxWithXYOp(Value *input, size_t kernel, size_t stride,
+                                 llvm::ArrayRef<size_t> pads) {
   ShapeNHWC idim = ShapeNHWC(input->dims());
   assert(idim.w >= kernel && idim.h >= kernel &&
```
artemrakhov: You have the same assert in Graph.h. Maybe it makes sense to remove this one?

lib/Graph/Graph.cpp
```diff
@@ -455,29 +456,51 @@ ConvolutionNode *Function::createConv(llvm::StringRef name, NodeValue input,
...
+PoolAvgNode *Function::createPoolAvg(llvm::StringRef name, NodeValue input,
+                                     size_t kernel, size_t stride,
+                                     llvm::ArrayRef<size_t> pads) {
   ShapeNHWC idim = ShapeNHWC(input.dims());
-  assert(idim.w >= kernel && idim.h >= kernel &&
+  PaddingTLBR pdim(pads);
+  (void)pdim;
+  assert((idim.w + pdim.left + pdim.right) >= kernel &&
+         (idim.h + pdim.top + pdim.bottom) >= kernel &&
          "buffer too small for selected stride");
```
artemrakhov: word "stride" here may be a bit confusing

beicy: This is not the only place this "buffer ... stride" msg shows. Will re-condition the word, and if necessary, will create a PR for it.

PR #1263, opened 2018-07-12
Perform Variable CSE on Variables of small enough size

jfix71:
This is an alternative to https://github.com/pytorch/glow/pull/1235

For Variables that are 256 elements or smaller and of the same type, are never written into, are private, and have no train kind, this will attempt to combine them into a single shared variable. This is particularly useful for IntLookupTables of the same "kind" (e.g. Tanh lookup tables), which often use the same data as their lookup table Var. Note that 256 is a heuristic used to ensure compile time isn't too long.

Also, note that I needed to add this after all other passes are complete, and not part of `optimize()` -- I added it to `optimize()` initially, but there are potential issues with other passes adding in new nodes that write to Variables. For example, `lower()` was creating new Save nodes for Variables that had been combined when lowering BatchNormalizationGrad, which breaks the program.

I haven't added a test yet -- if this is preferred to https://github.com/pytorch/glow/pull/1235 then I will add one.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -21,9 +21,17 @@
 #include "glow/Quantization/Base/Base.h"
 
 #include "llvm/Support/Casting.h"
+#include "llvm/Support/CommandLine.h"
 
 #include <unordered_map>
 #include <unordered_set>
+
+llvm::cl::OptionCategory graphOptCat("Graph Optimizations Options");
+llvm::cl::opt<unsigned> constVarDedupSizeOpt(
+    "const_var_dedup_size",
+    llvm::cl::desc(
+        "Max number of elements allowed for deduplicating constant variables"),
+    llvm::cl::Optional, llvm::cl::init(256), llvm::cl::cat(graphOptCat));
 
 using namespace glow;
 using llvm::cast;
@@ -1219,14 +1227,122 @@ struct CSEVisitor : NodeWalker {
   }
 };
 
+/// A helper type for hashing Variable pointers when they are used as keys in
+/// hash maps for deduplication. The hash is based on the type of the Variable
+/// (element type, dimensions), as well as a constant number of elements from
+/// the Variable to balance collisions with hash calclulation time.
+struct VarsHasherDedup {
+  size_t operator()(Variable *V) const {
+    auto hash = llvm::hash_value(V->getType());
+    auto &T = V->getPayload();
+    // Only use the first 8 elements in the hash. It's likely that if two
+    // tensors have different content they will diverge quickly. Fall back to
+    // full equality check in VarsEqDedup.
+    constexpr size_t maxNumEls = 8;
+    size_t numEls = std::min(T.getType().size(), maxNumEls);
+    size_t bufSize = T.getType().getElementSize() * numEls;
+    auto *data = T.getUnsafePtr();
+    for (size_t i = 0; i < bufSize; i++) {
+      hash = llvm::hash_combine(hash, data[i]);
+    }
+    return hash;
+  }
+};
+
+/// A helper type implementing the Variable equality predicate that can be used
+/// when Variable pointers are used as keys in hash maps for deduplication. It
+/// is assumed the Visibility and TrainKinds are the same, as deduplication
+/// only inserts if Private and None, respectively.
+struct VarsEqDedup {
+  bool operator()(const Variable *lhs, const Variable *rhs) const {
+    // Only consider Vars for deduplication if they have the same type. The
+    // train kind and visibility must already be the same.
+    if (lhs->getType() != rhs->getType()) {
+      return false;
+    }
+    assert(lhs->getVisibilityKind() == rhs->getVisibilityKind() &&
+           "Should only be comparing Variables with same VisibilityKind.");
+    assert(lhs->getTrainKind() == rhs->getTrainKind() &&
+           "Should only be comparing Variables with same TrainKind.");
+    // Only combine Vars if their data matches exactly, so allowed error is 0.0.
+    return lhs->getPayload().isEqual(rhs->getPayload(), /* allowedError */ 0.0);
+  }
+};
+
 } // namespace
+
+/// \returns true if Variable \p V is written into, either as a result, or as an
+/// overwritten input.
+static bool hasWriters(Variable *V) {
+  for (auto &U : V->getUsers()) {
+    auto *N = U.getUser();
+
+    // See if V is used as a result anywhere.
+    for (unsigned i = 0; i < N->getNumResults(); i++)
+      if (V == N->getNthResult(i))
+        return true;
+
+    // See if V is used as an overwritten input anywhere.
+    for (unsigned i = 0; i < N->getNumInputs(); i++)
+      if (N->isOverwrittenNthInput(i))
+        if (V == N->getNthInput(i))
+          return true;
+  }
+  return false;
+}
+
+/// Deduplicates constant variables in the Module \p M. Applicable constant
+/// variables for deduplication must have the same data, have
+/// VisibilityKind::Private, have TrainKind::None, and have no writers.
+static void deduplicateConstants(Module *M) {
+  // Map from Variables to other Variables that are equivalent for purposes of
+  // deduplication.
+  std::unordered_map<Variable *, Variable *, VarsHasherDedup, VarsEqDedup>
+      duplicateVars;
+
+  for (auto &V : M->getVars()) {
+    // Only perform deduplication on vars of small enough size. Otherwise just
+    // skip them. constVarDedupSizeOpt defaults to 256 as a heuristic, to keep
+    // compile time reasonable.
+    size_t maxNumEls = constVarDedupSizeOpt;
+    size_t numEls = V->getType()->size();
+    if (numEls > maxNumEls) {
+      continue;
+    }
+
+    // Only perform deduplication on private vars that have no train kind.
+    if (V->getVisibilityKind() != VisibilityKind::Private ||
+        V->getTrainKind() != Variable::TrainKind::None) {
+      continue;
+    }
+
+    // Only perform deduplication on vars that have no writers.
+    if (hasWriters(V)) {
+      continue;
+    }
+
+    // Try to find a var that has the same data as the current one.
+    auto foundI = duplicateVars.find(V);
+    if (foundI == duplicateVars.end()) {
+      // No node equivalent to the current one has been seen yet. Remember this
+      // variable, so that the next occurrence can be replaced by this one.
+      duplicateVars.emplace(V, V);
+      assert(duplicateVars.find(V) != duplicateVars.end());
+      continue;
+    }
+    Variable *foundV = foundI->second;
+    assert(V != foundV && "Variables should not be visited multiple times.");
+
+    // Replace current var by a found var, which is equivalent to it.
+    V->getOutput().replaceAllUsesOfWith(foundV);
+  }
+}
 
 /// Common Subexpression Elimination.
 static void CSE(Function *F) {
   CSEVisitor visitor;
 
-  // No need to perform CSE on variables because
-  // all variables are distinct from each other.
+  deduplicateConstants(F->getParent());
 
   // Perform CSE on all nodes.
   for (auto &N : F->getNodes()) {
--- a/tests/unittests/graphOptzTest.cpp
+++ b/tests/unittests/graphOptzTest.cpp
@@ -1114,3 +1114,72 @@ TEST_F(GraphOptz, concatReshapes) {
       llvm::dyn_cast<ConcatNode>(O->getInput()->getNthInput(0)->getNthInput(0));
   ASSERT_TRUE(newCN);
 }
+
+/// Check that Variable CSE works correctly, combining small Variables that have
+/// the same data.
+TEST_F(GraphOptz, VarsCSE) {
+  // Create three variables that are Private, have TrainKind None, and have no
+  // writers. The first two variables have the same data, and so should be
+  // combined via variable CSE. The third variable differs by the last value,
+  // and so should not be combined.
+  auto *input1 =
+      mod_.createVariable(ElemKind::FloatTy, {10}, "input1",
+                          VisibilityKind::Private, Variable::TrainKind::None);
+  auto *input2 =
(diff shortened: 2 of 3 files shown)


jfix71:
I stripped out the Visitor pattern as I'm not sure it's really useful in this case, can go back to it if preferred though.

Also considering renaming this to `mergeVars()` or something, as it's not working on subexpressions so CSE doesn't really make sense as a name.

nadavrot:
Jordan, thanks for working on this. Every optimization is special in some way but we try to fit all of them into the standard pipe. What is exactly the problem with adding CSE to the opt pipeline? Optimizations need to preserve the semantics of the program so I am not sure why CSEing breaks the code. Are we only applying CSE to const vars? 

jfix71:
@nadavrot If I add it to the end of `optimize()`, there is a problem when optimizing the lowered version of some gradient nodes, e.g. `BatchNormalizationGrad` and `SGD`. However I'm realizing this is probably due to a bug in NodeGen of not marking overwritten input correctly.

The lowering for `BatchNormalizationGrad` includes adding two `SaveNodes` to the graph which write to variables which previously were not written into; they were just inputs into `BatchNormalizationGrad` (mean and variance). `BatchNormalizationGradNode` does not mark these inputs as an overwritten. SGD also does not mark `Gsum` as overwritten even though it may be written to by a Save node too after lowering.

I suppose that if these two things are fixed, and we are sure we are not going to add other Nodes that write into Vars into the graph, it should be OK to add it back into `optimize()`.

bertmaher:
With respect to compile time, would it make sense to hash the variable contents as well as the type?

jfix71:
@bertmaher Good suggestion, I think it makes sense to hash some constant number of values, as it's likely the contents diverge very quickly if they're not equal. This way if they're not equal you don't have to iterate over all the contents to calculate the hash. Experimentally this was maybe 1% faster than either hashing all contents or none at all. Doesn't have a large impact because of the heuristic that only applies this to variables smaller than 256 elements. But if we decide it makes sense to increase 256 in the future then I would imagine hashing all the contents would grow comparatively worse.

jfix71:
I've updated this by adding in the missing overwritten inputs to NodeGen, a test case, and the first 8 elements of the Variable's contents to the hash.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1199,14 +1199,112 @@ struct CSEVisitor : NodeWalker {
...
+  for (auto &V : M->getVars()) {
+    // Perform CSE on all vars that have no writers.
+    if (varsWithWriters.find(V) != varsWithWriters.end()) {
+      continue;
+    }
+
+    // Only look for var CSE on vars of small enough size. Otherwise just skip
+    // them. 256 is used as a heuristic, to keep compile time reasonable.
+    size_t numEls = V->getType()->size();
+    if (numEls > 256) {
```
opti-mix: Could you introduce a named constexpr for 256 instead of hard-coding it here.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1199,14 +1199,112 @@ struct CSEVisitor : NodeWalker {
...
 
+/// Common Subexpression Elimination for Variables.
+static void varsCSE(Module *M) {
+  // Create a set of Vars that have writers.
+  std::unordered_set<Node *> varsWithWriters;
+
+  // Determine which vars are written into, either as a result, or as an
+  // overwritten input. This needs to be done across all functions, as this
+  // combines variables which may be used across multiple functions.
+  for (auto *F : M->getFunctions()) {
```
opti-mix: Move this for loop and nested ones into a separate function?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1199,14 +1199,112 @@ struct CSEVisitor : NodeWalker {
...
 };
 
+struct VarsHasherCSE {
+  size_t operator()(Variable *V) const {
+    auto hash = llvm::hash_value(V->getType());
+    auto &T = V->getPayload();
+    // Only use the first 8 elements in the hash. It's likely that if two
+    // tensors have different content they will diverge quickly. Fall back to
+    // full equality check in VarsEqCSE.
+    size_t numEls = std::min(T.getType().size(), 8ul);
```
opti-mix: Could you introduce a named constexpr for 8 instead of hard-coding it here.

jfix71:
@qcolombet There are a couple problems with that approach currently. One, this optimization only applies to constant Variables, but Variables don't have a Mutability kind. So we don't know if a Variable is const until the entire graph is constructed, when we can then inspect to see if anyone writes into it.

Two, `createVariable()` doesn't take the Variable's data -- the way it's currently done is basically:
```
auto *var = createVariable(ElemKind::FloatTy, {3}, "name");
var->getHandle<float>() = {1,2,3};
```

Maybe we should consider adding a Mutability kind to Variable, and a `createVariable()` that takes the data of the Variable. I'm not sure that makes this PR obsolete, though.

jfix71:
I squashed the commits and made requested changes. Happy to address any more issues/continue the discussion if this is the best path forward/what is otherwise.

qcolombet:
> @qcolombet There are a couple problems with that approach currently. One, this optimization only applies to constant Variables, but Variables don't have a Mutability kind. So we don't know if a Variable is const until the entire graph is constructed, when we can then inspect to see if anyone writes into it.

I agree that this only makes sense for variables with data, not necessarily constants.
If we modify the data at compile time, we can create a new variable if it is used somewhere else.
If we modify the data at runtime, then we have some save nodes and that just means that we need to schedule these save nodes after the (soon to land™) load nodes.

> Two, createVariable() doesn't take the Variable's data -- the way it's currently done is basically:
>
> auto *var = createVariable(ElemKind::FloatTy, {3}, "name");
> var->getHandle<float>() = {1,2,3};
> Maybe we should consider adding a Mutability kind to Variable, and a createVariable() that takes the data of the Variable.

I am actually thinking that we could, like the loader do, have an API that creates a variable from a tensor. E.g., for this example `auto *var = createVariable("name", tensor123);`. At this point, this is easy to check if the variable needs to be created or not. Most of our code does it the other way around like you showed, so that means a lot of refactoring. Even if we would stick to the createVariable then getHandle to set the value, I am confident we could find a way to still share the variables at build time.

>  I'm not sure that makes this PR obsolete, though.

I have to admit I haven't looked at the PR, so I cannot really comment on that front :P. That said, if we only merge identical constant Variable, I believe we shouldn't need an optimization.

Anyhow, my main concern is that we are going to pay some compile time to solve an issue we first create and the worse part is that we would only do it for "small" tensor otherwise that's too expensive. That's why I am trying to see if there are not more efficient alternative.

The bottom line, I don't want to hold on this if that already giving some benefits!

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1199,14 +1199,117 @@ struct CSEVisitor : NodeWalker {
...
+      continue;
+    }
+
+    // Try to find a var that has the same data as the current one.
+    auto foundI = cseVars.find(V);
+    if (foundI == cseVars.end()) {
+      // No node CSE-equivalent to the current one has been seen yet. Remember
+      // this variable, so that the next occurrence can be replaced by this
+      // one.
+      cseVars.insert({V, V});
```
rdzhabarov: super nit: emplace would avoid unnecessary copies.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1199,14 +1199,117 @@ struct CSEVisitor : NodeWalker {
...
+      assert(cseVars.find(V) != cseVars.end());
```
rdzhabarov: why is this needed? (we just inserted V into the map)

opti-mix: @rdzhabarov I think it is mostly copy/pasted from the CSE pass that I've written a while ago. IIRC, the initial idea of doing such a defensive check was to make sure that our custom hash and equality functions work properly when the insertion is performed. There were some issues with that at the beginning.

tools/ClassGen/NodeBuilder.h
```diff
@@ -142,6 +145,18 @@ class NodeBuilder {
                      "input of a node");
   }
 
+  NodeBuilder &addOverwrittenGradientInput(const std::string &name) {
```
nadavrot: Doxygen comment please.

jfix71:
> I agree that this only makes sense for variables with data, not necessarily constants.
> If we modify the data at compile time, we can create a new variable if it is used somewhere else.
> If we modify the data at runtime, then we have some save nodes and that just means that we need to schedule these save nodes after the (soon to land™) load nodes.

Yeah, I meant constant at runtime, but I agree that once the load nodes land we should be able to support this if they have a single writer.

> I am actually thinking that we could, like the loader do, have an API that creates a variable from a tensor. E.g., for this example `auto *var = createVariable("name", tensor123);`.

Yeah I'm a fan of that. If we don't do this we would need some way to signal that the Variable is complete (e.g. `var->finalize()` or something) to know when it's OK to try to combine it with others. 

However, I don't think doing this at build time would make any difference for overall compile time compared to the optimization from this PR (assuming build time is considered part of compile time). This PR checks if the data is the same in tensors of 256 elements or less. We would need to perform essentially this same logic at build time, checking if the data in other Variable's tensors is the same.

The main way I think we can avoid the compile time overhead is doing something like what I had done in the alternative I put up earlier this week with singleton variables https://github.com/pytorch/glow/pull/1235. This only works if we know ahead of time the Variable we're repeatedly creating has the same data. Otherwise I don't see how we can avoid having to compare the data in each Variable's tensor.

lib/Optimizer/GraphOptimizer.cpp
jfix71: Yep, what @opti-mix said. I can remove it if desired, but I did think it was useful.

tools/ClassGen/NodeGen.cpp
```diff
@@ -393,6 +395,7 @@ int main(int argc, char **argv) {
       .addMember(MemberType::Float, "Momentum")
       .addMember(MemberType::Unsigned, "BatchSize")
       .addOverwrittenInput("Weight")
+      .addOverwrittenInput("Gsum")
```
qcolombet: This should be done as a separate PR IMHO.

tools/ClassGen/NodeGen.cpp
```diff
@@ -125,6 +125,8 @@ int main(int argc, char **argv) {
       .addMember(MemberType::Float, "Epsilon")
       .addMember(MemberType::Float, "Momentum")
       .addResult("Input.getType()")
+      .addOverwrittenGradientInput("Mean")
+      .addOverwrittenGradientInput("Var")
```
qcolombet: This is only true for when we compile for training, thus I am not sure we want to put that unconditionally.
Put differently, although that's correct, I wonder if we are not leaving performance on the table for inference, which is what we care the most.

What do you think?

jfix71: This is `addOverwrittenGradientInput()`, not `addOverwrittenInput()`, so it does not impact the non-gradient node `BatchNormalizationNode`. It only marks `Mean` and `Var` as overwritten  in `BatchNormalizationGradientNode`. So it should not change anything/negatively impact performance for inference.

qcolombet: Ah right, but that's not accurate, isn't it?

I mean those inputs are going to be modified by the lowering of BN when we are in training mode not just by the lowering of BNG. (Admittedly, having one without the other would be pretty strange.)

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1219,14 +1219,119 @@ struct CSEVisitor : NodeWalker {
...
+    for (unsigned i = 0; i < N->getNumInputs(); i++)
+      if (N->isOverwrittenNthInput(i))
+        if (V == N->getNthInput(i))
+          return true;
+  }
+  return false;
+}
+
+/// Common Subexpression Elimination for Variables in the Module \p M.
+static void varsCSE(Module *M) {
```
nadavrot: I think that a better name would be "deduplicate constants".

jfix71: Done.

jfix71:
FYI we can now revisit this since https://github.com/pytorch/glow/pull/1326 fixes the BN issue and has landed.

tests/unittests/graphOptzTest.cpp
```diff
@@ -1114,3 +1114,39 @@ TEST_F(GraphOptz, concatReshapes) {
...
+  auto *CN = F_->createConcat("concat", {TN, SN, RN}, /* axis */ 0);
+  F_->createSave("ret", CN);
+
+  // Initially there are inputs 1, 2, and 3, and the variable for the save.
+  EXPECT_EQ(mod_.getVars().size(), 4);
+
+  ::glow::optimize(F_, CompilationMode::Infer);
+
+  // Now input1 and input2 have been combined, but input 3 has not.
+  EXPECT_EQ(mod_.getVars().size(), 3);
```
qcolombet: Could you check that we are indeed left with the variables we expect?
Also, check that the users have been properly updated.

jfix71: Done.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1219,14 +1219,121 @@ struct CSEVisitor : NodeWalker {
...
+static void deduplicateConstants(Module *M) {
+  // Map from Variables to other Variables that are equivalent for purposes of
+  // deduplication.
+  std::unordered_map<Variable *, Variable *, VarsHasherDedup, VarsEqDedup>
+      duplicateVars;
+
+  for (auto &V : M->getVars()) {
+    // Only perform deduplication on vars of small enough size. Otherwise just
+    // skip them. 256 is used as a heuristic, to keep compile time reasonable.
+    constexpr size_t maxNumEls = 256;
```
qcolombet: Could we make this a cl::opt with the default being 256?
That way we can easily play with different sizes for tuning or if we have a bug we can easily rule that optimization out by putting maxNumEls to 0.

jfix71: Done.

nadavrot:
Nice!

PR #1264, opened 2018-07-12
[backends] Finish CompiledFunction refactoring

bertmaher:
I think this is the light at the end of the tunnel for #1234

### DIFF
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -24,12 +24,7 @@
 
 namespace glow {
 
-class Context;
 class IRFunction;
-class Value;
-class Tensor;
-class Variable;
-class Function;
 class Node;
 
 enum class BackendKind {
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -44,17 +44,14 @@ public:
   /// Dtor.
   virtual ~Backend() = default;
 
-  /// Prepare the interpreter for execution of new code.
-  virtual void init(std::unique_ptr<IRFunction> IR) = 0;
+  /// Generate code for input function \param IR.
+  virtual std::unique_ptr<CompiledFunction>
+  compile(std::unique_ptr<IRFunction> IR) const = 0;
 
   /// Save the bundle for a later standalone execution.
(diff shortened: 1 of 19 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -45,6 +46,8 @@ class ExecutionEngine final {
   std::unique_ptr<Module> M_;
   /// The network execution backend.
   std::unique_ptr<Backend> backend_;
+  /// A glow function compiled for this ExecutionEngine's backend.
+  std::unique_ptr<CompiledFunction> function_;
```
rdzhabarov: we should probably reset this function on setBackend() call.

tests/unittests/BackendTest.cpp
```diff
@@ -145,8 +145,8 @@ TEST_P(BackendTest, debugPrint) {
   IRBuilder(IR.get()).createDebugPrintInst("print", *IR->getWeights().begin());
 
   std::unique_ptr<Backend> backend(createBackend(GetParam()));
-  backend->init(std::move(IR));
-  backend->doForwardPass();
+  auto function = backend->compile(std::move(IR));
```
jfix71: nit: I prefer explicitly using * for auto pointers

include/glow/Backends/Backend.h
```diff
@@ -44,29 +39,26 @@ class Backend {
...
 
-  /// Prepare the interpreter for execution of new code.
-  virtual void init(std::unique_ptr<IRFunction> IR) = 0;
+  /// Generate code for input function \param IR.
+  virtual std::unique_ptr<CompiledFunction>
+  compile(std::unique_ptr<IRFunction> IR) const = 0;
 
   /// Save the bundle for a later standalone execution.
-  virtual void save(std::unique_ptr<IRFunction> IR, llvm::StringRef outputDir) {
+  virtual void save(std::unique_ptr<IRFunction> IR, llvm::StringRef outputDir) const {
```
jfix71: nit: make sure you clang-format!

tests/unittests/BackendTest.cpp
bertmaher: It's a `unique_ptr` though, so I don't think `auto *` is correct.

PR #1265, opened 2018-07-12
Add the 1to1 mode (-1:1 range) to the image classifier application

tlepley:
We are working with models trained for inputs in the -1:1 range.
This small PR then proposes to add the "1to1" mode to the image classifier application.

### DIFF
--- a/tests/images/run.sh
+++ b/tests/images/run.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 ./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=resnet50 $@
-./bin/image-classifier tests/images/imagenet/*.png -image_mode=128to127 -m=vgg19 $@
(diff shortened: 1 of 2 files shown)


jfix71:
@tlepley Thanks for working on this!

I think I would prefer this mode to be called something like `neg1to1` instead of `1to1`. Any objections?

tlepley:
@jfix71 I agree  `neg1to1`  is more explicit, but I picked this `1to1` name simply to be consistent with the existing `128to127` mode (values are in the range: -128 .. 127). If I rename `1to1` into `neg1to1`, I think we should also rename `128to127` into `neg128to127`. Any opinion ?

jfix71:
@tlepley Yeah I agree, let's change both then. Thanks!

bertmaher:
If you change 128to127 can you also update tests/images/run.sh?  Thanks!

tlepley:
@jfix71 @bertmaher Done.

PR #1266, opened 2018-07-12
[ee] A few more API cleanups

bertmaher:


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -38,20 +38,12 @@ static llvm::cl::opt<bool> dumpIR("dump-ir",
                                   llvm::cl::desc("Prints IR to stdout"));
 } // namespace
 
-ExecutionEngine::ExecutionEngine(BackendKind backendKind) {
-  backendKind_ = backendKind;
-  M_.reset(new Module());
-  backend_.reset(createBackend(backendKind_));
-}
+ExecutionEngine::ExecutionEngine(BackendKind backendKind)
+    : backend_(createBackend(backendKind)) {}
```
rdzhabarov: right now, one has to make sure that compile is run prior to run or runBatch.
Can we assert on function_ being non-nullptr at the very beginning of those methods?

Should we consider having run and runBatch on CompiledFunction instead?

bertmaher: Good idea re: assertions.

I think I want to keep the variable initialization outside of CompiledFunction for now since it's currently uniform across backendsm, and CF is there to virtualize the backend-specific execution details.

But I could see a separate "binding" operation like #1227 proposes.

rdzhabarov: "But I could see a separate "binding" operation like #1227 proposes." not necessary for the purpose of this PR. Thanks!

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -38,27 +38,20 @@ static llvm::cl::opt<bool> dumpIR("dump-ir",
                                   llvm::cl::desc("Prints IR to stdout"));
 } // namespace
 
-ExecutionEngine::ExecutionEngine(BackendKind backendKind) {
-  backendKind_ = backendKind;
-  M_.reset(new Module());
```
rdzhabarov: once the module is created, where is it used?

PR #1267, opened 2018-07-12
[format] Speed up clang-format with xargs -P

bertmaher:
Maybe if I can format my PRs faster I'll forget to do it less often :-p.  The magic of parallelism takes format.sh from 9 to 2 seconds on my Macbook.

### DIFF
--- a/utils/format.sh
+++ b/utils/format.sh
@@ -9,11 +9,7 @@ if [ $(which clang-format) ]; then
   FARRAY=( $FILES ) # count the number of files to process
   echo  Formatting ${#FARRAY[@]} files
 
-  for F in $FILES; do
-    clang-format -i $F
-    echo -n .
-  done
-  echo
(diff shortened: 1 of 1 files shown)


opti-mix:
Nice speedup!

BTW, I personally prefer to use `git-clang-format -f`, which formats only the files you touched in my current working copy. It does not try to format any other files, e.g. those ones not correctly formatted by other people in other commits. And since it only formats the files you have changed, it is also much faster because you typically touch a small number of files.

rdzhabarov:
The only downside of git-clang-format -f is that it works only on currently modified/changed files.
For example, if you commit your changes without running git-clang-format -f then after commit the command will do nothing.

opti-mix:
@rdzhabarov git-clang-format can also take a `--commit commit_hash` parameter, which should in theory allow for something like `git-clang-format --commit HEAD -f`. Never tried to use it this way though ;-)

PR #1269, opened 2018-07-12
[quantization][OpenCL]: Quantization support for quantize/dequantize and element-wise instructions.

ZchiPitt:
Quantization support for quantize/dequantize and element-wise instructions.
- Support for QuantizeInst/DequantizeInst and implementing their OpenCL kernels.
- Support for the quantized element-wise instructions
- Introduced one test case that test the instructions above, and compare with the results of Interpreter backend.


tests/unittests/OperatorTest.cpp
```diff
@@ -2450,6 +2452,83 @@ TEST_P(InterpOnly, SparseLengthsSum) {
   EXPECT_TRUE(expected.isEqual(result));
 }
 
+/// A helper function to test quantized element-wise ops on the specific
+/// backend.
+void quantizedElementwiseOpsOnSpecificBackend(BackendKind backend, Tensor *lhs,
```
rdzhabarov: what stops these test to be run on all backends, TESTP_P(Operator ... ?

tests/unittests/OperatorTest.cpp
```diff
@@ -44,6 +44,8 @@ class InterpAndCPU : public Operator {};
 
 class InterpOnly : public Operator {};
 
+class OpenCLOnly : public Operator {};
```
rdzhabarov: not sure if we want this specific one, seems it's better to run quantization ops tests on all backends.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -176,8 +176,21 @@ class OCLBackend final : public Backend {
   bool transformPostLowering(Function *F, CompilationMode mode) const override;
 
   bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const override {
+    // Check quantization support.
     if (elementTy == ElemKind::Int8QTy) {
```
rdzhabarov: have you tried profiling Resnet50 graph with Interpreter and then load_profile with OpenCL backend?
Would be a good thing to look at and also dump DAG and see what was actually quantized.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -22,6 +22,7 @@
 #include "glow/Graph/Nodes.h"
 #include "glow/IR/IRUtils.h"
 #include "glow/IR/Instrs.h"
+#include "glow/Quantization/Quantization.h"
```
rdzhabarov: why does it need Quantization.h? Depending on Quantization lib will introduce circular dependency which was fixed recently :)

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -561,12 +571,60 @@ void OpenCLFunction::execute() {
...
-        setKernelArg(kernel, numArgs + 1, SI->getValue());
+        setKernelArg(kernel, ++numArgs, SI->getValue());
       } else if (auto *EPI = dyn_cast<ElementPowInst>(&I)) {
         // Pass the exp as a parameter.
-        setKernelArg(kernel, numArgs + 1, EPI->getExp());
+        setKernelArg(kernel, ++numArgs, EPI->getExp());
       }
 
+      if (isQuantized) {
+        int32_t destOffset = I.getOperand(0).first->getType()->getOffset();
```
rdzhabarov: use names instead of getOperand(0) etc.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -528,6 +529,15 @@ void OpenCLFunction::execute() {
                                      : ElemKind::FloatTy;
     std::string kernelName = getKernelName(I.getKindName(), elemTy);
 
+    //  Check if the instruction is quantized.
+    bool isQuantized = false;
+    if (I.getNumOperands() &&
+        I.getOperand(0).first->getType()->isQuantizedType()) {
```
rdzhabarov: checking "isQuantized" should be enough, you could remove "if" also and do this in one statement.

tests/unittests/OperatorTest.cpp
```diff
@@ -2450,6 +2452,83 @@ TEST_P(InterpOnly, SparseLengthsSum) {
...
+  case glow::Kinded::Kind::ElementMaxInstKind:
+    F = mod.createFunction("maxFunc");
+    opNode = F->createMax("max", X, Y);
+    break;
+  default:
+    F = mod.createFunction("main");
+    break;
+  }
+  auto *result = F->createSave("save", opNode);
+  Function *PF = glow::profileQuantization(F);
```
rdzhabarov: do not profile quantization in operator test.

ZchiPitt: thx!

lib/Backends/OpenCL/OpenCL.cpp
ZchiPitt: Thx! But i cant call getDest() here as instruction I was not casted into any specific instruction, and it does not have a getDest() member function

rdzhabarov:
@ZchiPitt also please make sure to prioritize the PR for bugfix here: https://github.com/pytorch/glow/pull/1273 before addressing comments on this PR.

tests/unittests/OperatorTest.cpp
ZchiPitt: updated!

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -561,12 +566,69 @@ void OpenCLFunction::execute() {
...
+          setKernelArg(kernel, ++numArgs, rhsScaleParams);
+        }
+      }
+      enqueueKernel(commands_, kernel, deviceId_, {global}, kernelLaunches_);
+      continue;
+    }
+
+    // Quantize floating point tensor. Scale and Offset are based on return type
+    // of the instruction \p I.
+    if (auto *QI = dyn_cast<QuantizeInst>(&I)) {
```
rdzhabarov: This is fine, but smaller granularity PR is faster to review/ship.
Would be good to keep one quantized operation per PR unless the other one really touches few lines.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -69,6 +81,50 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
...
 
+/// Scales a 32-bit integer using the integer shift-mult-shift method.
+cl_int32_t scale_i32i8(cl_int32_t input, cl_int32_t pre, cl_int32_t post,
+                       cl_int32_t scale, cl_int32_t offset) {
+  // See more details in libjit_defs.h:libjit_scale_i32i8()
+  int rtn = (post > 0) ? (1 << (post - 1)) : 0;
+  return ((((input >> pre) * scale) + rtn) >> post) + offset;
+}
+
+// Clips int32_t into int8_t
```
rdzhabarov: doxygen.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -561,12 +566,69 @@ void OpenCLFunction::execute() {
...
         // Pass the splat as a parameter.
-        setKernelArg(kernel, numArgs + 1, SI->getValue());
+        setKernelArg(kernel, ++numArgs, SI->getValue());
       } else if (auto *EPI = dyn_cast<ElementPowInst>(&I)) {
         // Pass the exp as a parameter.
-        setKernelArg(kernel, numArgs + 1, EPI->getExp());
+        setKernelArg(kernel, ++numArgs, EPI->getExp());
       }
 
+      if (isQuantized) {
```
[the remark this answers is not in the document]
rdzhabarov: I think this is kinda fragile.
Can you explicitly state which instructions need to run quantized here and only call setKernelArgs for those.

tests/unittests/BackendTestUtils.cpp
```diff
@@ -57,6 +58,58 @@ void inferBatchedAddNet(Tensor *batch, Tensor *slice, Tensor *out,
   out->copyFrom(&result->getVariable()->getPayload());
 }
 
+/// A helper function to test quantized element-wise ops on the specific
+/// backend.
+void quantizedElementwiseOpsOnSpecificBackend(BackendKind backend, Tensor *lhs,
```
rdzhabarov: I think this is not needed.

OperatorTest already covers quantized math for CPU and Interpreter. There is a parameter in the test which defines the backend. So the only thing you need to do is to enable test on OpenCL, e.g., TEST_P(Operator) instead of TEST_P(InterpAndCPU).

lib/Backends/OpenCL/OpenCL.cpp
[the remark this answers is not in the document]
ZchiPitt: sounds right. Updated!

tests/unittests/BackendTestUtils.cpp
ZchiPitt: moved it back to OperatorTest.cpp. I've changed the test to compare quantized op result with non-quantized op results. This saves me from previous problem of explicitly specifying backend.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -183,6 +241,65 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
     name##K(&mem[dest], &mem[lhs], &mem[rhs]);                                 \
   }
 
+/// Macro to define a kernel for data-parallel binary quantized operations. The
+/// body of the kernel is auto-generated by the macro.
+/// \p name the name of the kernel 
+/// \p type the type of the tensor elements and of the return value 
+/// \p body the operation to be performed 
+/// The naming follows the convention of its corresponding implemention in CPU 
+/// baclend.
```
rdzhabarov: typo

lib/Backends/OpenCL/kernels.cl
```diff
@@ -183,6 +241,65 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
...
+  }
+
+/// Macro to define a mini-kernel for data-parallel multiplicative quantized
+/// operations. The body of the kernel is auto-generated by the macro.
+/// \p name the name of the kernel
+/// \p type the type of the tensor elements
+/// \p body the operation to be performed
+/// The naming follows the convention of its corresponding implemention in CPU 
+/// baclend.
+#define DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL_QUANTIZED_M(name, type,      \
```
rdzhabarov: why do you need to parametrize this by type?

lib/Backends/OpenCL/kernels.cl
```diff
@@ -183,6 +241,65 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
...
+    size_t i = get_global_id(0);                                               \
+    cl_int32_t LHS = lhs[i] - lhsOffset;                                       \
+    cl_int32_t RHS = rhs[i] - rhsOffset;                                       \
+    dest[i] =                                                                  \
+        clip(scale_i32i8((body), mulPre, mulPost, mulScale, destOffset));      \
+  }                                                                            \
+  __kernel void name##_i8W(                                                    \
+      __global void *mem, cl_uint32_t dest, cl_uint32_t lhs, cl_uint32_t rhs,  \
+      cl_int32_t destOffset, cl_int32_t lhsOffset, cl_int32_t rhsOffset,       \
+      QuantizationTransform32To8 mulScaleParams) {                             \
```
rdzhabarov: maybe name it resultScaleParams.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -328,11 +445,24 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
 
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementadd, float, LHS + RHS)
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementsub, float, LHS - RHS)
-DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmul, float, LHS *RHS)
+DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmul, float, LHS * RHS)
```
rdzhabarov: maybe that what clang format wanted (lhs *rhs), can you run clang-format for this PR.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -69,6 +81,52 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
...
+  return ((((input >> pre) * scale) + rtn) >> post) + offset;
+}
+
+/// Clips int32_t into int8_t.
+inline cl_int8_t clip(cl_int32_t val) {
+  return (cl_int8_t)min(max(val, -128), 127);
+}
+
+/// Quantizes \p input from float to int8. 
+inline cl_int8_t quantize(float input, float scale, cl_int32_t offset) {
```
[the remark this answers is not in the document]
ZchiPitt: was copying from the convention of the corresponding implementation in libjit. But it seems i should follow the same coding style here with other func in this file. Removed inline.

lib/Backends/OpenCL/kernels.cl
ZchiPitt: changed

lib/Backends/OpenCL/kernels.cl
ZchiPitt: it can be used by both ElementMul and ElementDiv, so it's better to parameterize it.

lib/Backends/OpenCL/kernels.cl
ZchiPitt: sounds good

lib/Backends/OpenCL/kernels.cl
ZchiPitt: oh, this is after i run the clang-format, and it should be okay for this.

ZchiPitt:
@rdzhabarov updated QuantizedArithmeticUnrescaled with Operator. It didnt fail on my mac

lib/Backends/OpenCL/kernels.cl
rdzhabarov: If i read this correctly "type" is always int8

tests/unittests/OperatorTest.cpp
rdzhabarov: I do not think we need this test. TEST_P(Operator, QuantizedArithmeticUnrescaled) tests the OpenCL quantized operations, make sure that QuantizedArithmeticUnrescaled has all arithmetic operations you are interested in.

After you add the Rescale operator support you could change QuantizedArithmetic*Rescaled* test to be Operator also.

ZchiPitt: You are right. I just remove my test.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -561,12 +566,73 @@ void OpenCLFunction::execute() {
...
+          auto rhsScaleParams = quantization::quantizeScaleOffset32To8(
+              I.getOperand(2).first->getType()->getScale() / destScale,
+              I.getOperand(2).first->getType()->getOffset());
+          setKernelArg(kernel, ++numArgs, destOffset);
+          if (isa<ElementMulInst>(I) || isa<ElementDivInst>(I)) {
+            auto resultScaleParams = quantization::quantizeScaleOffset32To8(
+                I.getOperand(1).first->getType()->getScale() *
+                    I.getOperand(2).first->getType()->getScale() / destScale,
+                0);
+            setKernelArg(kernel, ++numArgs, lhsScaleParams.offset_);
```
rdzhabarov: why do we not handle ElementAddInst inst and the rest here? The code as of now it only supports quantized ElementMulInst and ElementDivInst.

ZchiPitt:
Removed my test implementation as QuantizedArithmeticUnrescaled basically does the same thing

lib/Backends/OpenCL/kernels.cl
```diff
@@ -328,11 +442,18 @@ size_t getNCHW(ShapeNCHW s, cl_uint32_t n, cl_uint32_t c, cl_uint32_t h,
...
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementsub, float, LHS - RHS)
-DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmul, float, LHS *RHS)
+DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmul, float, LHS * RHS)
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementdiv, float, LHS / RHS)
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmax, float, max(LHS, RHS))
 DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(elementmin, float, min(LHS, RHS))
 
+DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL_QUANTIZED(elementadd, LHS + RHS)
+DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL_QUANTIZED(elementsub, LHS - RHS)
+DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL_QUANTIZED_M(elementmul, LHS * RHS)
```
opti-mix: Could you group these macro-calls in such a way that we first call all non `_M` macros and then all `_M` macros?

lib/Backends/OpenCL/OpenCL.cpp
rdzhabarov: you could pass rhsScaleParams and lhsScaleParams in both cases and remove else condition, might look cleaner.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -176,8 +176,21 @@ class OCLBackend final : public Backend {
...
-      return false;
+      switch (opKind) {
+      case Kinded::Kind::AddNodeKind:
```
rdzhabarov: can you sort that in alpha order?

lib/Backends/OpenCL/kernels.cl
ZchiPitt: done

lib/Backends/OpenCL/OpenCL.h
ZchiPitt: done

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -22,6 +22,7 @@
 #include "glow/Graph/Nodes.h"
 #include "glow/IR/IRUtils.h"
 #include "glow/IR/Instrs.h"
+#include "glow/Quantization/Base/Base.h"
```
rdzhabarov: you need to modify the CMake file accordingly since you introduce a dependency.

ZchiPitt: done. Thanks a lot for your reviews!

PR #1270, opened 2018-07-12
[Importer] Support "axis" and "axis_w" params of  "FC" op in Caffe2.

beicy:
Since only Caffe2 support FC op, we move it to Caffe2.cpp. In ONNX, if the input is 2-dims, Gemm op can be used to represent FC. If the input has > 2dims, we can use Reshape+Gemm for FC layer.
According to Caffe2 FC defs, the input and weight don't have to be 2-dims, and they will be coerced into 2D dims for calculation. In Glow's FC node, the input will be flatten to 2-dims, but the weight needs to be 2D. Therefore, the weight loaded from Caffe2 model should be coerced into 2-dims before the FC node is created.
The "axis" and "axis_w" params of FC operators in Caffe2 are supported. https://caffe2.ai/docs/operators-catalogue.html#fc . The params are used in inception_v1 models.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -132,14 +132,18 @@ struct PaddingTLBR {
   }
 };
 
-/// Collapse a tensor shape into two sizes: the first dimension and the size of
-/// the rest of the dimensions.
-/// For example, [7, 3, 4, 2] -> [7, 24]
-inline std::pair<size_t, size_t> flattenCdr(llvm::ArrayRef<size_t> dims) {
-  assert(dims.size() > 1);
+/// Collapse a tensor shape into two sizes: the first n dimensions and the size
+/// of the rest of the dimensions. For example, ([7, 3, 4, 2], 1) -> [7, 24]
+inline std::pair<size_t, size_t> flattenCdr(llvm::ArrayRef<size_t> dims,
+                                            size_t n = 1) {
+  assert(1 <= n && n < dims.size());
   size_t first = dims[0];
-  size_t rest = dims[1];
-  for (size_t i = 2; i < dims.size(); i++) {
+  for (size_t i = 1; i < n; i++) {
+    first *= dims[i];
+  }
+
+  size_t rest = dims[n];
+  for (size_t i = n + 1; i < dims.size(); i++) {
     rest *= dims[i];
   }
 
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -266,6 +266,42 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "FC") {
+    // Load the inputs:
+    auto *in = getOrCreateVariableByName(op.input(0));
+    if (in->getType(0)->dims().size() > 2) {
+      size_t axis = dict.count("axis") ? loadInt(dict["axis"]) : 1;
+      auto xDim = flattenCdr(in->getType(0)->dims(), axis);
+      in = G_.createReshape("fc.in", in, {xDim.first, xDim.second});
+    }
+
+    // Load weights.
+    Tensor *w = getTensorByName(op.input(1));
+    Tensor *b = getTensorByName(op.input(2));
+    size_t axis_w = dict.count("axis_w") ? loadInt(dict["axis_w"]) : 1;
+
+    // Caffe2 stores the transposed W matrix. In here we first coerce W to a 2D
+    // matrix size if necessay and then transpose it back.
+    Tensor wtag;
+    if (w->dims().size() > 2) {
+      auto wDims = flattenCdr(w->dims(), axis_w);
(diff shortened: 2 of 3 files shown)


lib/Importer/Caffe2.cpp
```diff
@@ -279,6 +279,38 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+    Tensor *w = getTensorByName(op.input(1));
+    Tensor *b = getTensorByName(op.input(2));
+
+    // Caffe2 stores the transposed W matrix. 
+    // In here we first coerce W into 2D tensor if necessary, 
+    // and then transpose W back.
+    Tensor wtag;
+    if (w->dims().size() != 2) {
+      size_t second = w->size() / b->size();
+      Tensor tmp(ElemKind::FloatTy, {b->size(), second});
```
artemrakhov: After reading comment here:
https://github.com/caffe2/caffe2/blob/0dd3284525079f3870df92f61bed3b94eb45ff53/caffe2/operators/fully_connected_op.cc#L167
I just think that we need to call `flattenCdr()`, like we do for `X` input upon lowering.

Also, turns out that they also have `axis` and `axis_w` arguments. While you are here, it may be good idea to assert that these are either not present, or are equal to 1.

tools/ClassGen/NodeGen.cpp
```diff
@@ -105,6 +105,8 @@ int main(int argc, char **argv) {
       .addInput("Input")
       .addInput("Weights")
       .addInput("Bias")
+      .addMember(MemberType::SizeT, "Axis")
```
artemrakhov: Have you seen nets where Axis is not 1?

tools/ClassGen/NodeGen.cpp
```diff
@@ -105,6 +105,8 @@ int main(int argc, char **argv) {
...
+      .addMember(MemberType::SizeT, "AxisW")
```
artemrakhov: Why do we need this AxisW? As far as I understand, it is always =1, since you do flattening in Caffe2 Importer.

lib/Optimizer/Lower.cpp
```diff
@@ -94,13 +94,14 @@ void lowerRegressionGradNode(Function *F, RegressionGradNode &node) {
 }
 
 void lowerFullyConnectedNode(Function *F, FullyConnectedNode &FC) {
-  auto xDim = flattenCdr(FC.getInput().getType()->dims());
-  auto wDim = FC.getWeights().dims();
+  auto xDim = flattenCdr(FC.getInput().getType()->dims(), FC.getAxis());
```
artemrakhov: If you end up leaving Axis and AxisW, then you'll also need to support them in `lowerFullyConnectedGradNode()`

nadavrot:
@beicy @artemrakhov I wonder if we can use the reshape node before the FC and get rid of the new arguments. The loader could insert the reshape.

tools/ClassGen/NodeGen.cpp
beicy: No. "Axis" uses default value 1 in Caffe2 zoo model.

tools/ClassGen/NodeGen.cpp
beicy: Yes, in Caffe2 importer, "weight" is flatten. But I think we should also support the param for FC in C++ ?

beicy:
@nadavrot  Inception_v1 Caffe2 model uses axis_w. So do you mean why we need axis and axis_w in Glow FC node?

nadavrot:
@beicy Can we use a sequence of FC and reshape instead of adding parameters to the FC node? 

What's the semantics of the new parameters?

beicy:
@artemrakhov @nadavrot I will avoid adding new parameters to Glow FC node. Yes, we can do some flatten before creating a FC node.

include/glow/Base/Type.h
```diff
@@ -132,14 +132,18 @@ struct PaddingTLBR {
...
-/// Collapse a tensor shape into two sizes: the first dimension and the size of
-/// the rest of the dimensions.
-/// For example, [7, 3, 4, 2] -> [7, 24]
-inline std::pair<size_t, size_t> flattenCdr(llvm::ArrayRef<size_t> dims) {
-  assert(dims.size() > 1);
+/// Collapse a tensor shape into two sizes: the first n dimensions and the size
+/// of the rest of the dimensions. For example, ([7, 3, 4, 2], 1) -> [7, 24]
+inline std::pair<size_t, size_t> flattenCdr(llvm::ArrayRef<size_t> dims,
+                                            size_t n = 1) {
+  assert((dims.size() > 1 && n < dims.size()));
```
artemrakhov: No need in extra brackets.

Also, please check that `1 <= n`. Then `dims.size() > 1` check is not needed, because `1 <= n && n < dims.size()` implies that `dims.size()` is at least 2.

nadavrot:
@beicy Nice!

Issue #1212, opened 2018-07-04
image-classifier behaves differently with one or several images for densenet121


qcolombet:
I noticed that the image-classifier returns different results if invoked with one file at a time and all the same file at once and was wondering if this is expected.

To reproduce:
$ ./bin/image-classifier  tests/images/imagenet/* -image_mode=0to1 -m=densenet121/model.onnx
$ ./bin/image-classifier  tests/images/imagenet/cat_285.png -image_mode=0to1 -m=densenet121/model.onnx

Result:
With all the files at once:
Model: densenet121/model.onnx
 File: tests/images/imagenet/cat_285.png Result:281
 File: tests/images/imagenet/dog_207.png Result:207
 File: tests/images/imagenet/zebra_340.png Result:340

With one at a time:
Model: densenet121/model.onnx
 File: tests/images/imagenet/cat_285.png Result:674
Model: densenet121/model.onnx
 File: tests/images/imagenet/dog_207.png Result:674
Model: densenet121/model.onnx
 File: tests/images/imagenet/zebra_340.png Result:674

Also when there is only one file given, the process takes much longer to produce a result.

Is this expected?

bertmaher:
Good catch.  I'm pretty sure this is a regression.

bertmaher:
I bisected this to "Optimization on ConcatNode, Concat(Reshape(x) * N) -> Reshape(Concat(x * N))" 
 https://github.com/pytorch/glow/commit/710cc3b0db91854a36c44080c360ca751e37a393
using
```
git bisect start 45abbe8 3a9f9da --
git bisect run regress.sh
```
With `regress.sh`:
```
cd `dirname $0`;
ninja image-classifier
./bin/image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m=densenet121/model.onnx | grep -q 'Result:28[15]'
exit $?
```

@ZchiPitt can you please take a look and fix forward?

qcolombet:
I think @beicy also looked at this issue. @beicy, did you make any progress on this?

beicy:
@qcolombet  Not yet. Haven't spent much time on this one yet :)

qcolombet:
Thanks for the update @beicy. I just wanted to be sure @ZchiPitt and you were not looking at the same thing.

ZchiPitt:
Sorry just saw it. Will look into it later today.

opti-mix:
[...]

PR #1273, opened 2018-07-13
[fix]: Fix the issue[1212] that densenet produces inconsistent results

ZchiPitt:
- Fixed the [issue](https://github.com/pytorch/glow/issues/1212) that densenet cannot produce consistent results. 
- Modified the optimization of Concat(Reshape(x) * N)->Reshape(Concat(x * N)) by setting up more strict requirements for the optimization: 
-- previously we determine if a reshape node can be optimized by checking if we can concatenate based on same size sub-tensor after the specified dimension
-- Now we check not only the trailing dimensions but also the leading dimensions to be the same size.


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -903,20 +903,28 @@ static bool checkConcatNodeUniformDims(llvm::ArrayRef<NodeValue> inputs,
...
-// Given one tensor and the desired sub-tensor size, \returns the dimension,
-// after which the sub-tensor has the desired size, otherwise return -1.
-// Example: Given a tensor <2,3,4,5>, and a desired size 20, this function will
-// return dimension 1 as the sub-tensor after it is <4,5>, which has the
-// size 20.
-static ssize_t
-findMatchingConcatDimForSameSizeSubTensor(llvm::ArrayRef<size_t> firstDims,
-                                          size_t desiredSubTensorSize) {
-  size_t currentSubTensorSize = 1;
+/// Given one tensor and the desired leading/trailing dimensions sizes.
```
jfix71: Can you specify the arguments here in the description for doxygen? E.g. `Given a tensor's dims \p firstDims, ...` etc.

tests/unittests/graphOptzTest.cpp
```diff
@@ -1047,25 +1047,26 @@ TEST_F(GraphOptz, concatElim) {
...
-  const size_t shape1[] = {10, 20};
-  const size_t shape2[] = {10, 1, 20};
+  const size_t shape1[] = {2, 5, 1, 20};
+  const size_t shape2[] = {10, 1, 2, 10};
   const size_t shape3[] = {5, 40};
   llvm::SmallVector<NodeValue, 10> inputs1;
   llvm::SmallVector<NodeValue, 10> inputs2;
   for (size_t i = 0; i < 10; i++) {
-    // 10 reshape nodes that transform from {10,20} to {10, 1, 20}.
+    const size_t shape2[] = {10, 1, 2, 10};
```
jfix71: Seems this was duplicated unnecessarily?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -925,33 +933,40 @@ findMatchingConcatDimForSameSizeSubTensor(llvm::ArrayRef<size_t> firstDims,
...
 /// 1) Input tensors are concatenate-able along this dimension.
-/// 2) The sub-tensors after this dimension in the input tensors, are
-///    of the same size as the sub-tensor of the input of the original
-///    Concat node after the concatenation dimension.
+/// 2) The trailing/leading dimensions sizes after/before this dimension in the
+///    input tensors, are of the same size as the trailing/leading dimensions of
+///    the input of the original Concat node after/before the concatenation
+///    dimension.
 /// \returns this dimension if found, otherwise -1.
 static int tryToConcatAlongSameSizeSubTensor(llvm::ArrayRef<NodeValue> inputs,
```
jfix71: I'm realizing I don't like the this function being named `tryToConcat...`, as it doesn't actually do the concat itself, it just tries to find the concat dimension. Maybe it can be `findConcatDim...` or something similar?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -925,33 +933,40 @@ findMatchingConcatDimForSameSizeSubTensor(llvm::ArrayRef<size_t> firstDims,
...
   }
 
-  // Try to find the dimension in the first input such that the sub-tensor size
-  // based on this direction is the same as the size of sub-tensors based on the
-  // concatenation dimension used by the original ConcatNode.
+  // Try to find the dimension in the first input such that the trailing/leading
+  // dimensions sizes are the same as the sizes of the trailing/leading
+  // dimensions based on the concatenation dimension used by the original
+  // ConcatNode.
   ssize_t dimension = findMatchingConcatDimForSameSizeSubTensor(
```
opti-mix: You use "dim" for dimension almost everywhere. May be the variable should be called "dim" as well to be consistent?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -925,33 +933,40 @@ findMatchingConcatDimForSameSizeSubTensor(llvm::ArrayRef<size_t> firstDims,
...
-/// 2) The sub-tensors after this dimension in the input tensors, are
-///    of the same size as the sub-tensor of the input of the original
-///    Concat node after the concatenation dimension.
+/// 2) The trailing/leading dimensions sizes after/before this dimension in the
+///    input tensors, are of the same size as the trailing/leading dimensions of
+///    the input of the original Concat node after/before the concatenation
+///    dimension.
 /// \returns this dimension if found, otherwise -1.
 static int tryToConcatAlongSameSizeSubTensor(llvm::ArrayRef<NodeValue> inputs,
                                              ConcatNode *origConcatN) {
```
opti-mix: "origConcatN" stays for "original Concat Node" I guess? Later you use names like "trailingDimsProdOriginalConcatNode", which is inconsistent. 

Either use "ConcatNode" in such cases everywhere, or even better "CN" to make the names a bit shorter as they are getting really long. 

In any case, please pick one way of naming things and use it consistently in the code.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -925,33 +933,40 @@ findMatchingConcatDimForSameSizeSubTensor(llvm::ArrayRef<size_t> firstDims,
 /// there is a dimension in the input tensors, with which we can meet
 /// two requirements:
 /// 1) Input tensors are concatenate-able along this dimension.
-/// 2) The sub-tensors after this dimension in the input tensors, are
-///    of the same size as the sub-tensor of the input of the original
-///    Concat node after the concatenation dimension.
+/// 2) The trailing/leading dimensions sizes after/before this dimension in the
+///    input tensors, are of the same size as the trailing/leading dimensions of
+///    the input of the original Concat node after/before the concatenation
+///    dimension.
```
opti-mix: Can you explain why we need to meet these conditions? E.g. because they ensure that the payload of the new concat node should be the same as the payload of the original concat node.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -903,20 +903,28 @@ static bool checkConcatNodeUniformDims(llvm::ArrayRef<NodeValue> inputs,
...
+static ssize_t findMatchingConcatDimForSameSizeSubTensor(
+    llvm::ArrayRef<size_t> firstDims, size_t leadingDimsProdOriginalConcatNode,
+    size_t trailingDimsProdOriginalConcatNode) {
+  size_t trailingDimsProdCurNode = 1;
   for (ssize_t i = firstDims.size() - 1; i >= 0; i--) {
-    if (currentSubTensorSize == desiredSubTensorSize) {
-      return i;
+    if (trailingDimsProdCurNode == trailingDimsProdOriginalConcatNode) {
+      size_t leadingDimsProdCurNode = 1;
+      for (size_t j = 0; j < i; j++) {
```
jfix71: It's not necessary for this to be a nested loop; I'd prefer it to be refactored into two non-nested loops to be more readable. The first loop looks for `i` where `trailingDimsProdCurNode == trailingDimsProdOriginalConcatNode`. If it doesn't find such an `i` then it returns `-1`. The second loop is then essentially then the inner loop here, given this `i`.

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: sounds right!

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: thx. done

tests/unittests/graphOptzTest.cpp
ZchiPitt: thx!

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: changed to findConcatDimForSameTrailingAndLeadingDims

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: changed to originalConcatNode

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: Thx for the advice! However it's possible that there are multiple i where `trailingDimsProdCurNode == trailingDimsProdOriginalConcatNode`. 
say for example like <2,3,1,1,1,1,3>, with `trailingDimsProdOriginalConcatNode = 3`, `leadingDimsProdOriginalConcatNode=2`, if we early break the first loop when i = 6, then the current leading dims size is 6, which is not desired. But if we can iterate to i=1, we then can return the correct dim.

ZchiPitt:
Sorry for the late response on this PR, i was delayed by another PR.

lib/Optimizer/GraphOptimizer.cpp
jfix71: Ah ok cool got it. I suggest adding a unit test for this case then, as it seems this is currently not covered by the unit test you added.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -886,8 +886,8 @@ static void optimizeBatchNorm(Function *F) {
   } // For all nodes in the graph.
 }
 
-// \returns true if all dimensions of the input tensors are the same except
-// for the provided dimension, otherwise return false.
+/// \returns true if all dimensions of the input tensors are the same except
```
jfix71: nit: add `\p` for `inputs` and `dimension`

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -903,70 +903,90 @@ static bool checkConcatNodeUniformDims(llvm::ArrayRef<NodeValue> inputs,
...
     return -1;
   }
-  return dimension;
+  return dim;
 }
 
-// Given the inputs of one Concat Nodes, \returns true if they are all
-// ReshapeNode, and the input tensors of these input nodes have same number of
-// dimensions, otherwise returns false.
+/// Given the inputs of one Concat Nodes, \returns true if they are all
```
jfix71: nit: Add doxygen `\p` for inputs here too

PR #1274, opened 2018-07-13
Introduce a NodeHandle type to make sure that use-lists of Nodes only contain uses by other Nodes

opti-mix:
This PR consists of three major changes:
- It introduces a new type NodeHandle that is used instead of NodeValue inside Nodes to reference the results of other Nodes. NodeHandles register themselves as users in the use-lists. NodeValue becomes a simple POD which does not register itself in the use-lists.
- The affected APIs are updated to use NodeHandle instead of NodeValue where appropriate
- The UnownedNodeValueMap type is removed as it is not needed anymore


opti-mix:
This is a follow-up of #1198

PR #1276, opened 2018-07-14
Provide the functionality to iterate over all nodes referencing a given node

opti-mix:
It is now possible to iterate over a use-list of a Node and see which nodes refer to it. NodeUse has now a getUser() method to get this information.

To provide this functionality, we store in each NodeHandle a pointer to the parent node that contains it.

One possible concern regarding this PR is that we add a pointer per NodeHandle, which increases the amount of memory used by Nodes. I was playing with the idea of using (small) offsets from the beginning of the object instead of pointers. But this does not work with nodes like Concat, which contain `std::vector<NodeHandle>`. Such handles inside vectors are allocated on heap and can be very far away from the beginning of the parent object. Therefore, their offsets may require as much memory as a pointer. Hopefully, this possible memory increase is not a big concern, because we don't have that many nodes even for big NN models.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -31,6 +31,7 @@ namespace glow {
 
 class Node;
 class NodeWalker;
+struct NodeUse;
 
 /// Unlike LLVM values, graph nodes may return multiple values as the result of
 /// a computation. Gradient-calculating nodes such as conv-grad return multiple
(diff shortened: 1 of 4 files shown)


nadavrot:
Maybe we should add a use list verifier?

opti-mix:
@nadavrot You mean checking that all uses from the use-list really refer to the node and that all nodes referring to the same node are also the use-list of this node? Yes, we could add such checks to the graph verifier.  Good idea!

PR #1277, opened 2018-07-14
[GraphIR] Make the overwrites explicit for SGD through Save

qcolombet:
Instead of using overwrites of input arguments for the weight and
gsum arguments of the SGD node, have the node produces two results
and save them at the right places.

This is one more step toward having save and load nodes the only
direct users of variables.

NFC.

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -71,8 +71,6 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
 
   // A list of nodes to add to the Function.
   std::vector<Node *> toAppend;
-  // A list of vars to add to the Function.
-  std::vector<Variable *> newVars;
 
   // Generate the gradient nodes for each one of the nodes in the function.
 
@@ -232,9 +230,6 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
   // Add all of the new variables and instructions.
   for (auto &I : toAppend) {
     G->addNode(I);
-  }
-  for (auto &I : newVars) {
-    G->getParent()->addVar(I);
   }
 
   return G;
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -220,17 +220,13 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       continue;
     }
 
-    TypeRef Ty = conf.momentum > 0 ? V->getType() : G->getParent()->getVoidTy();
-    Variable *gsum = new Variable("gsum", Ty, VisibilityKind::Private,
-                                  Variable::TrainKind::Broadcast, 0,
-                                  G->getParent()->getPRNG());
-
-    newVars.push_back(gsum);
-
-    auto X = new SGDNode(V->getName(), map.getGradient(V), V, gsum,
-                         conf.L1Decay, conf.L2Decay, conf.learningRate,
-                         conf.momentum, conf.batchSize);
+    auto X = new SGDNode(V->getName(), map.getGradient(V), V, conf.L1Decay,
+                         conf.L2Decay, conf.learningRate, conf.momentum,
+                         conf.batchSize);
     toAppend.push_back(X);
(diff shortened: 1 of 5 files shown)


nadavrot:
Quentin, why do we need to know about gsum at the graph level? We can just create the gsum buffer during lowering. 

nadavrot:
Keep in mind that other trainers have other parameters. For example, Adam has xsum. It does not make sense to represent them at the graph level if we don’t care about the details of the buffers. Lowering can create all of the details by adding new gsum and xsum variables. 

qcolombet:
Removing the argument something sounds good to me.

qcolombet:
Given we have the same problem with variance and mean for batch normalization. Should we remove those as well?

I am not sure what makes sense to expose or not. We do have a construtor that exposes them but it’s used only in a unit test.
What do you think?

qcolombet:
Reopen, the weight update will still need to be fixed. I’ll update the pr with removing gsum.

qcolombet:
Updated the PR to remove Gsum from the graph.

qcolombet:
@nadavrot Ping, with Gsum removed.

tools/ClassGen/NodeGen.cpp
```diff
@@ -386,13 +386,12 @@ int main(int argc, char **argv) {
...
       .addInput("Gradient")
       .addInput("Weight")
-      .addInput("Gsum")
       .addMember(MemberType::Float, "L1Decay")
       .addMember(MemberType::Float, "L2Decay")
       .addMember(MemberType::Float, "LearningRate")
       .addMember(MemberType::Float, "Momentum")
       .addMember(MemberType::Unsigned, "BatchSize")
-      .addOverwrittenInput("Weight")
+      .addResult("Weight.getType()", "UpdatedWeight")
```
nadavrot: Is the result the updated weight or the delta that we need to add to the weight? We need to clarify that in a comment, I think.

lib/Graph/Grad.cpp
```diff
@@ -220,26 +218,20 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
-    newVars.push_back(gsum);
-
-    auto X = new SGDNode(V->getName(), map.getGradient(V), V, gsum,
-                         conf.L1Decay, conf.L2Decay, conf.learningRate,
-                         conf.momentum, conf.batchSize);
+    auto X = new SGDNode(V->getName(), map.getGradient(V), V, conf.L1Decay,
+                         conf.L2Decay, conf.learningRate, conf.momentum,
+                         conf.batchSize);
     toAppend.push_back(X);
+    // Now update the weight and gsum, if we care, with the values
```
nadavrot: The comment is incorrect. We don't update gsum here.

tools/ClassGen/NodeGen.cpp
artemrakhov: What if we make Gsum to be another output of SGDNode?

tests/unittests/graphGradTest.cpp
```diff
@@ -137,5 +137,22 @@ TEST(GraphAutoGrad, cloneAndDiff) {
...
 
   EXPECT_EQ(M.getFunctions().size(), 3);
-  EXPECT_EQ(M.getVars().size(), 7);
+  EXPECT_EQ(M.getVars().size(), 5);
+  // Check that we have as many SGD node as variables that need to be trained.
+  unsigned nbSGDs = 0;
+  unsigned nbSGDA = 0;
+  unsigned nbSGDB = 0;
+  for (auto &node : diffF->getNodes()) {
+    SGDNode *SGD;
```
artemrakhov: nit: it's probably more readable this way:
```
SGDNode *SGD = llvm::dyn_cast<SGDNode>(&node);
if (!SGD)
  continue;
```

lib/Graph/Grad.cpp
artemrakhov: We don't update gsum here. Stale comment?

PR #1279, opened 2018-07-16
[Importer] Use addNodeAsOutput function to the save outputs code for each operator

beicy:


beicy:
@jfix71  Yes, the run.sh works.

rdzhabarov:
@beicy I think "dropout", "squezee" could also be changed accordingly. Could you fix those and check there are no more cases to clean up?

beicy:
@rdzhabarov Thanks for pointing out the missing ones. I will double check it.

rdzhabarov:
also, nit, there comments "// Save the outputs:" which seem useless.
It's better to have a method name such that it self-explanatory so you do not have to place comments on it.

PR #1280, opened 2018-07-16
[graph-verifier] Add more checks for checking the consistency of Nodes use-lists

opti-mix:
- Each node referring to another node N (i.e. having N as its input) should have a proper entry in the use-list of the node N.

- Each use from the use-list of a node N should refer to node N.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -207,6 +207,7 @@ struct NodeUse {
   /// \returns the instruction that the use refers to.
   NodeHandle *get() const { return site_; }
   /// Get the node containing this use.
+  const Node *getUser() const { return site_->parent_; }
   Node *getUser() { return site_->parent_; }
   /// Sets the operand to a new value.
   void setOperand(NodeHandle &site);
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1838,6 +1838,22 @@ Function *Function::clone(llvm::StringRef newName,
   return newF;
 }
 
+/// Verify the input \p idx of a node \p N. Check that the node \p N is in the
+/// use-list of the corresponding input node.
+static void verifyNodeInput(const Node &N, size_t idx) {
+  auto input = N.getNthInput(idx);
+  auto refN = input.getNode();
+  // Check that N is in the use-list of the input node and there is a proper
+  // entry for it.
+  for (auto &U : refN->getUsers()) {
+    if (U.getUser() == &N && *U.get() == input) {
+      return;
+    }
+  }
+  GLOW_UNREACHABLE(
+      "Any node referencing another node N be in the use-list of the node N");
+}
+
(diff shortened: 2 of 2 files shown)


opti-mix:
This is a follow-up of #1276 to improve the graph verifier as suggested by @nadavrot .

include/glow/Graph/Node.h
```diff
@@ -207,7 +207,7 @@ struct NodeUse {
   /// \returns the instruction that the use refers to.
   NodeHandle *get() const { return site_; }
   /// Get the node containing this use.
-  Node *getUser() { return site_->parent_; }
+  Node *getUser() const { return site_->parent_; }
```
rdzhabarov: could this return const Node *? It's a bit tricky to expose the internals of the const object and one could modify the const object this way.

opti-mix: @rdzhabarov   We could. I wonder though what happens if you want to get a user of a node and then mutate that user. If we make `getUser` const, then that would require an explicit cast. Or we can have two `getUser` methods for const and non-const uses. What do you prefer?

opti-mix: I added two versions of `getUser`, one that is const returning  `const Node *` and another which is non-const.

lib/Graph/Graph.cpp
```diff
@@ -1838,6 +1838,22 @@ Function *Function::clone(llvm::StringRef newName,
   return newF;
 }
 
+/// Verify the input \p idx of a node \p N. Check that the node \p N is in the
+/// use-list of the corresponding input node.
+static void verifyNodeInput(const Node &N, size_t idx) {
+  auto input = N.getNthInput(idx);
+  auto refN = input.getNode();
```
rdzhabarov: do we have any standard on a format of `auto` with the pointers? (would be good to stick to a common style).

opti-mix: You mean we should use `auto *` here? Yeah. I wish the tooling could warn us in cases like `auto *` or `auto &` if we forget it.

lib/Graph/Graph.cpp
```diff
@@ -1838,6 +1838,22 @@ Function *Function::clone(llvm::StringRef newName,
...
+  // Check that N is in the use-list of the input node and there is a proper
+  // entry for it.
+  for (auto &U : refN->getUsers()) {
+    if (U.getUser() == &N && *U.get() == input) {
+      return;
+    }
+  }
+  GLOW_UNREACHABLE(
```
rdzhabarov: btw, i think verifiers mostly use normal asserts in the codebase.

lib/Graph/Graph.cpp
rdzhabarov: yeah, like auto * in this case.

opti-mix:
@rdzhabarov Thanks for your comment! I addressed them in #1285. Please have a look.

PR #1281, opened 2018-07-16
[GraphIR] Add a proper copy constructor for NodeHandle

qcolombet:
Prior to this patch, copying a NodeHandle around (like with a vector::resize)
would not properly set the tracking of the value.

For instance, this was impossible to pretty print nodes with vector of inputs
like concat, because our pretty printer was doing creating a local NodeHandle.

This patch features a proper copy constructor for NodeHandle and at the same
time, make sure pretty printing vectors of NodeHandle doesn't create temporary
objects.

NFC.

### DIFF
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -359,6 +359,9 @@ TEST(Graph, cloneTest2) {
 
   auto *newF = F->clone("new_main");
   newF->verify();
+  F->dump();
+  newF->dump();
+
   EXPECT_EQ(newF->getNodes().size(), F->getNodes().size());
   EXPECT_EQ(newF->getParent(), F->getParent());
 }
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -143,6 +143,8 @@ public:
     setOperand(that.getNode(), that.getResNo());
   }
 
+  NodeHandle(const NodeHandle &that) : NodeHandle(that.parent_, that) {}
+
   /// Create an empty handle.
   NodeHandle() : NodeValue(nullptr), parent_(nullptr) {}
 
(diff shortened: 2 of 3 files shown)


opti-mix:
@qcolombet Nice catch! Thanks!

tests/unittests/graphTest.cpp
```diff
@@ -359,6 +360,11 @@ TEST(Graph, cloneTest2) {
 
   auto *newF = F->clone("new_main");
   newF->verify();
+  F->dump();
+  F->dumpDAG();
```
opti-mix: Do we really need `dumpDAG` or may be just `dump` is enough? I'm asking because `dumpDAG` creates a new file each time it runs. If we can avoid it, we should not call it.

tests/unittests/graphTest.cpp
```diff
@@ -52,6 +52,7 @@ TEST(Graph, simpleTestConv) {
   K = F->createRELU("Relu", K);
   K = F->createSoftMax("SoftMax", K, S);
   F->createSave("Save", K);
+
```
opti-mix: Why does it show any change here?

qcolombet: Historical accident :)

tests/unittests/graphTest.cpp
qcolombet: dump is enough. Switching to this. The reason I went with both was because the other tests use both.

bertmaher: Slightly troll-ish suggestion: `dumpDAG("/dev/stdout")`  :-D

qcolombet:
@bertmaher, hehe.
I was actually thinking of being even more aggressive about it with "/dev/null", then I thought, that's not going to end well on Windows when we support it, right? (That's even more troll-ish :P)

PR #1284, opened 2018-07-16
[unittest] Add a test case exposing a bug in our scheduler

qcolombet:
Our graph does not model the memory dependencies explicitly.
As a result, if we have a variable is written and read on different
data paths, the scheduler may be unlucky and we read the wrong
version of the variable.

Essentially, if you have:
     A
    / \
write read

The scheduler can produce either:
write A
read A
or
read A
write A

Obviously this does not produce the same output.
Glow intended semantic is that write should happen last, but with
our current algorithm, this is actually possible to fool the
scheduler not to do that.

### DIFF
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -460,3 +460,89 @@ TEST(Graph, disableUnrollingGroupConv) {
   EXPECT_GT(numberOfNodesOpenCL, numberOfNodesInterpreter);
 #endif // GLOW_WITH_OPENCL
 }
+
+/// Check that save nodes are properly scheduled.
+/// That is, they happen after the last use of the related variable.
+/// In that test, the order of the creation of the nodes give a valid schedule.
+TEST(Graph, schedulingOfSavesOrderProvided) {
+  ExecutionEngine EE;
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  auto *A = mod.createVariable(ElemKind::FloatTy, {3, 32}, "A",
+                               VisibilityKind::Public,
+                               Variable::TrainKind::Xavier, 1.0);
+  auto *zero = mod.createVariable(A->getType(), "zero", VisibilityKind::Public,
+                                  Variable::TrainKind::Broadcast, 0.0);
+
+  auto *B = mod.createVariable(A->getType(), "B", VisibilityKind::Public,
+                               Variable::TrainKind::Xavier, 1.0);
+
+  auto *addAB = F->createAdd("addAB", A, B);
+
+  auto *saveNode = F->createSave("ret", addAB);
+  F->createSave("resetA", zero, A);
+
+  // Copy the value of A.
+  Tensor AOrig = A->getPayload().clone();
+
+  EE.compile(CompilationMode::Infer, F);
+  EE.run({}, {});
+  auto *ret = saveNode->getVariable();
+  auto handleAOrig = AOrig.getHandle<>();
+  auto handleB = B->getPayload().getHandle<>();
+  auto handleRet = ret->getPayload().getHandle<>();
+  bool allEqual = true;
+  for (unsigned row = 0; row != 3; ++row) {
+    for (unsigned column = 0; column != 32; ++column) {
+      allEqual &= handleAOrig.at({row, column}) + handleB.at({row, column}) ==
+                  handleRet.at({row, column});
+    }
+  }
+  EXPECT_TRUE(A->getPayload().isEqual(zero->getPayload(), 0.0));
+  EXPECT_TRUE(allEqual);
+}
+
+/// Same as schedulingOfSavesOrderProvided except the order in which the nodes
+/// are added to the function don't form a valid schedule.
+/// In other words, the scheduler won't get away with scheduling
+/// using only the order of the nodes in the list of nodes.
+TEST(Graph, schedulingOfSaves) {
+  ExecutionEngine EE;
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  auto *A = mod.createVariable(ElemKind::FloatTy, {3, 32}, "A",
+                               VisibilityKind::Public,
+                               Variable::TrainKind::Xavier, 1.0);
+  auto *zero = mod.createVariable(A->getType(), "zero", VisibilityKind::Public,
+                                  Variable::TrainKind::Broadcast, 0.0);
+  F->createSave("resetA", zero, A);
+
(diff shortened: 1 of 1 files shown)


tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
 #endif // GLOW_WITH_OPENCL
 }
+
+/// Check that save nodes are properly scheduled.
+/// That is, they happen after the last use of the related variable.
+/// In that test, the order of the creation of the nodes give a valid schedule.
+TEST(Graph, schedulingOfSavesOrderProvided) {
+  ExecutionEngine EE;
+
+  Tensor inputs(ElemKind::FloatTy, {1, 32, 32, 3});
```
jfix71: nit: This seems to be unused

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+}
+
+/// Same as the previous test, except the order in which the nodes
+/// are added to the function don't form a valid schedule.
+/// In other words, the scheduler won't get away with scheduling
+/// using only the order of the nodes in the list of nodes.
+TEST(Graph, schedulingOfSaves) {
+  ExecutionEngine EE;
+
+  Tensor inputs(ElemKind::FloatTy, {1, 32, 32, 3});
```
jfix71: nit: Also seems to be unused

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+    for (unsigned column = 0; column != 32; ++column) {
+      EXPECT_EQ(handleA.at({row, column}), 0.0);
+      allEqual &= handleAOrig.at({row, column}) + handleB.at({row, column}) ==
+                  handleRet.at({row, column});
+    }
+  }
+  EXPECT_TRUE(allEqual);
+}
+
+/// Same as the previous test, except the order in which the nodes
```
jfix71: nit: maybe say `Same as schedulingOfSavesOrderProvided` in case ordering of tests changes.

tests/unittests/graphTest.cpp
qcolombet: Indeed! Interestingly the compiler didn’t complain about unused variable.

qcolombet:
The idea is to merge this and keep the test.
The test would still be valid because we have one save to A and one save to ret.

The only thing that will change when I fix the scheduler is the final expect_false that will become except_true.

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+  F->createSave("resetA", zero, A);
+
+  // Copy the value of A.
+  Tensor AOrig = A->getPayload().clone();
+
+  EE.compile(CompilationMode::Infer, F);
+  EE.run({}, {});
+  auto *ret = saveNode->getVariable();
+  auto handleAOrig = AOrig.getHandle<>();
+  auto handleA = A->getPayload().getHandle<>();
```
rdzhabarov: nit: getHandle() should be enough.

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+/// are added to the function don't form a valid schedule.
+/// In other words, the scheduler won't get away with scheduling
+/// using only the order of the nodes in the list of nodes.
+TEST(Graph, schedulingOfSaves) {
+  ExecutionEngine EE;
+
+  Tensor inputs(ElemKind::FloatTy, {1, 32, 32, 3});
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
```
rdzhabarov: same thing here.

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+  auto *ret = saveNode->getVariable();
+  auto handleAOrig = AOrig.getHandle<>();
+  auto handleA = A->getPayload().getHandle<>();
+  auto handleB = B->getPayload().getHandle<>();
+  auto handleRet = ret->getPayload().getHandle<>();
+  bool allEqual = true;
+  for (unsigned row = 0; row != 3; ++row) {
+    for (unsigned column = 0; column != 32; ++column) {
+      EXPECT_EQ(handleA.at({row, column}), 0.0);
+      allEqual &= handleAOrig.at({row, column}) + handleB.at({row, column}) ==
```
rdzhabarov: there is also a method for checking equality of handles

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+                  handleRet.at({row, column});
+    }
+  }
+  // This should be true, but right now it is not!
```
rdzhabarov: maybe a TODO instead of a simple comment for easier discoverability

tests/unittests/graphTest.cpp
```diff
@@ -460,3 +460,97 @@ TEST(Graph, disableUnrollingGroupConv) {
...
+/// That is, they happen after the last use of the related variable.
+/// In that test, the order of the creation of the nodes give a valid schedule.
+TEST(Graph, schedulingOfSavesOrderProvided) {
+  ExecutionEngine EE;
+
+  Tensor inputs(ElemKind::FloatTy, {1, 32, 32, 3});
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  F->setName("interpret");
```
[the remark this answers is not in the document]
qcolombet: Indeed!

tests/unittests/graphTest.cpp
qcolombet: I went with FIXME.

tests/unittests/graphTest.cpp
qcolombet: Maybe you meant for zero :P

qcolombet: Did that for zero == A.

qcolombet:
Hi @jfix71 - My take away of my conversation with @nadavrot is that we don't want to model memory dependency, at least for now, because we don't have any use case for this.
Thus, save is always "save final".

nadavrot:
@qcolombet Thank you for adding this testcase. I really like the practice of adding unit tests that communicate and demonstrate the problem.

PR #1290, opened 2018-07-18
[ONNX] ONNX operator version 

beicy:
Some ONNX ops have different versions. For example, Gemm-1 and Gemm-6 has "broadcast" attr whose default value is 0, while Gemm-7 and Gemm-8 version have no such attribute and always broadcast Tensor C. So we get op_version for each model to handle different op versions if necessary. This PR is triggered by inception_v2 ONNX model and only handles Gemm op versions. I think we could support different op version by demand in the future.

### DIFF
--- a/include/glow/Importer/Caffe2.h
+++ b/include/glow/Importer/Caffe2.h
@@ -38,6 +38,9 @@ class Value;
 /// Loads caffe2 models.
 class caffe2ModelLoader
     : public CommonOperatorLoader<caffe2::OperatorDef, caffe2::Argument> {
+  /// Get the broadcast attribute.
+  bool getBroadcast(const ArgumentDictionaryTy &dict) override;
+
   /// Load the weight tensors from the 'init' file and register them in the map
   /// \p tensors.
   void loadWeights(caffe2::NetDef &net);
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -39,6 +39,8 @@ protected:
   using ArgumentDictionaryTy =
       std::unordered_map<std::string, const AttrType *>;
 
+  virtual bool getBroadcast(const ArgumentDictionaryTy &dict) { return true; }
+
   void addNodeAsOutput(const OpType &op, Node *R) {
     for (int i = 0, e = op.output_size(); i < e; i++) {
       nodeByName_[op.output(i)] = R;
@@ -124,11 +126,11 @@ protected:
     auto *in0 = getOrCreateVariableByName(op.input(0));
     auto *in1 = getOrCreateVariableByName(op.input(1));
 
-    int broadcast = loadInt(dict["broadcast"]);
+    bool broadcast = getBroadcast(dict);
 
     Node *finalIn1 = nullptr;
-    if (broadcast == 1) {
-      int axis = loadInt(dict["axis"]);
+    if (broadcast) {
+      int axis = dict.count("axis") ? loadInt(dict["axis"]) : -1;
       // In ONNX, if axis == -1 then it sets the axis so that the
       // trailing-most dimensions are aligned like this.
       if (axis == -1) {
--- a/include/glow/Importer/ONNX.h
+++ b/include/glow/Importer/ONNX.h
@@ -28,6 +28,7 @@ namespace onnx {
 class AttributeProto;
 class NodeProto;
 class GraphProto;
+class ModelProto;
 } // namespace onnx
 
 namespace glow {
@@ -35,6 +36,12 @@ namespace glow {
 /// Loads ONNX models.
 class ONNXModelLoader
     : public CommonOperatorLoader<onnx::NodeProto, onnx::AttributeProto> {
+  /// Get the broadcast attribute based on different ONNX op versions.
+  bool getBroadcast(const ArgumentDictionaryTy &dict) override;
+
+  /// Set ir verion and op version.
+  void setVersion(onnx::ModelProto MP);
+
   /// Load the network operators from the GraphProto.
   void loadNetwork(onnx::GraphProto &net);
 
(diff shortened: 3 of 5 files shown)


lib/Importer/ONNX.cpp
```diff
@@ -364,7 +375,10 @@ void ONNXModelLoader::loadOperator(const onnx::NodeProto &op) {
     Node *B = getOrCreateVariableByName(op.input(1));
     Node *C = getOrCreateVariableByName(op.input(2));
 
-    bool broadcastC = dict.count("broadcast") && loadInt(dict["broadcast"]);
+    bool broadcastC =
+        (opset_version == 1 || opset_version == 6)
```
artemrakhov: what about opset_version 2 or 3, etc? I think it's more readable this way:
```
bool broadcastC = true;
if (opset_version <= 6)
  broadcastC = dict.count("broadcast") && loadInt(dict["broadcast"]);
```

lib/Importer/ONNX.cpp
```diff
@@ -63,6 +63,17 @@ bool ONNXModelLoader::loadProtoFile(onnx::GraphProto &net,
   onnx::ModelProto MP;
   bool parseNet = MP.ParseFromCodedStream(&codedstr);
   net = MP.graph();
+  ir_version = MP.ir_version();
+  opset_version = 0;
+  if (ir_version >= 3) {
+    // Model with IR version < 3 doesn't have opset_import specified.
+    for (const auto &imp : MP.opset_import()) {
+      if (imp.has_domain() || imp.domain() == "") {
```
artemrakhov: What does this if check? Please add a comment on what are we trying to skip.

Did you mean `!` before first operand here? Because `imp.has_domain()` covers all cases: when `imp.domain() == ""` and also when it's not equal to ""

lib/Importer/ONNX.cpp
```diff
@@ -63,6 +63,17 @@ bool ONNXModelLoader::loadProtoFile(onnx::GraphProto &net,
   onnx::ModelProto MP;
   bool parseNet = MP.ParseFromCodedStream(&codedstr);
   net = MP.graph();
+  ir_version = MP.ir_version();
```
artemrakhov: Let's move new code to new function. We don't want `loadProtoFile()` to be too long.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -124,11 +124,11 @@ class CommonOperatorLoader : public ProtobufLoader {
     auto *in0 = getOrCreateVariableByName(op.input(0));
     auto *in1 = getOrCreateVariableByName(op.input(1));
 
-    int broadcast = loadInt(dict["broadcast"]);
+    int broadcast = dict.count("broadcast") ? loadInt(dict["broadcast"]) : 1;
```
artemrakhov: Does this also depend on opset version?

include/glow/Importer/ONNX.h
```diff
@@ -49,6 +49,12 @@ class ONNXModelLoader
   /// file.
   bool loadProtoFile(onnx::GraphProto &net, const std::string &filename);
 
+  /// ONNX model ir_version;
+  size_t ir_version;
```
artemrakhov: Any plans on using this? If not, then I'd prefer to keep is a local var where it's extracted.

include/glow/Importer/CommonOperatorLoader.h
beicy: Yes.  I will add the version check.

include/glow/Importer/ONNX.h
beicy: For ONNX "constant", if ir_version < 3, the "type" is not required. But if ir_version is >=3, the "type" is required. Since so far we don't have such  a model, I didn't modify this part.

nadavrot: Please follow the variable naming conventions. irVersion_

include/glow/Importer/CommonOperatorLoader.h
artemrakhov: So seems like "broadcast" is one thing, that ONNX changed between versions 6 and 7. Maybe we can create virtual method `getBroadcast()` (or `loadBroadcast()`), which will be different for ONNX and Caffe2 importers. Caffe2's one will be trivial, and ONNX method will look at opset_version.

include/glow/Importer/Caffe2.h
```diff
@@ -38,6 +38,9 @@ class Value;
 /// Loads caffe2 models.
 class caffe2ModelLoader
     : public CommonOperatorLoader<caffe2::OperatorDef, caffe2::Argument> {
+  /// Get the broadcast attribute.
+  int getBroadcast(const ArgumentDictionaryTy &dict) override;
```
artemrakhov: I'd prefer to have Broadcast as `bool`. I think it is `int` in ONNX solely for storage purposes.

lib/Importer/ONNX.cpp
```diff
@@ -50,6 +50,26 @@ static ArgumentDictionaryTy loadArgumentMap(const onnx::NodeProto &op) {
...
+  irVersion_ = MP.ir_version();
+  opsetVersion_ = 0;
+  if (irVersion_ >= 3) {
+    // Model with IR version < 3 doesn't have opset_import specified.
+    for (const auto &imp : MP.opset_import()) {
+      if (!imp.has_domain() || imp.domain() == "") {
+        opsetVersion_ = imp.version();
+        break;
+      }
+    }
```
artemrakhov: Is it an error if opsetVersion_ was not set after the loop?

lib/Importer/ONNX.cpp
artemrakhov: Well, opset_version is not coming from Gemm's description. It's common for all operators. I think that it can be 0 or 3 (because of some other op's changes), and we can totally see any number here.

lib/Importer/ONNX.cpp
beicy: So according to ONNX team, there should be an assertion for irVersion_ < 3.  If there is no domain or domain is "", it means it is ONNX default op. So as long as irVersion_ >=3, the opsetVerson_ should be set for the models we used.

artemrakhov: OK. Then let's add an assert for irVersion_ >= 3, and another one for opsetVersion_ is found.

PR #1291, opened 2018-07-18
[GraphIR] Add a parent link to the nodes

qcolombet:
It is now possible to access the function that holds a node directly
from a node.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -547,6 +547,9 @@ public:
   Node *getNodeByName(llvm::StringRef name);
 
   const NodesList &getNodes() const { return nodes_; }
+
+  /// \returns pointer to the class member for the nodes list.
+  static NodesList Function::*getNodesMemberPtr() { return &Function::nodes_; }
 };
 
 struct TrainingConfig;
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -264,7 +264,7 @@ public:
   /// \returns the function holding this node.
   /// If that node does not belong to any function, this
   /// is nullptr.
-  Function *getParent() const { return parent_; }
+  const Function *getParent() const { return parent_; }
   /// Set the link to the function that holds this node.
   void setParent(Function *parent) { parent_ = parent; }
 
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -29,6 +29,7 @@
 
 namespace glow {
 
+class Function;
 class Node;
 class NodeWalker;
 struct NodeUse;
@@ -239,9 +240,12 @@ protected:
   /// nodes.
   static void destroyNode(Node *N);
 
+  /// Link to the function holding this node.
+  Function *parent_;
+
 public:
   Node(Kinded::Kind k, llvm::StringRef name)
-      : Named(name), Kinded(k), predicate_(this, nullptr) {}
+      : Named(name), Kinded(k), predicate_(this, nullptr), parent_(nullptr) {}
 
   /// \returns the nullable predicate of the current node.
   const NodeValue getPredicate() const;
@@ -256,6 +260,13 @@ public:
   NodeValue getNthResult(unsigned idx);
   /// \returns the n'th result of the node.
   const NodeValue getNthResult(unsigned idx) const;
+
+  /// \returns the function holding this node.
+  /// If that node does not belong to any function, this
+  /// is nullptr.
+  Function *getParent() const { return parent_; }
+  /// Set the link to the function that holds this node.
+  void setParent(Function *parent) { parent_ = parent; }
 
   /// Getters/setters to access Node's inputs and outputs.
   unsigned getNumInputs() const;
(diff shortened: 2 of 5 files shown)


include/glow/Graph/Graph.h
```diff
@@ -183,6 +183,7 @@ class Function final : public Named {
   /// Inserts the node \p N to the list of nodes, and returns the inserted node.
   template <class NodeTy> NodeTy *addNode(NodeTy *N) {
     N->setName(Module::uniqueName(N->getName(), uniqueNodeNames_));
+    N->setParent(this);
```
opti-mix: Besides adding the nodes, we also need to handle the removal of nodes and set the parent to nullptr. And we need to the corresponding check to the test you introduced.

qcolombet: What do you mean?
We don't have a removeNode API, do we?

include/glow/Graph/Node.h
```diff
@@ -257,6 +261,13 @@ class Node : public Named,
   /// \returns the n'th result of the node.
   const NodeValue getNthResult(unsigned idx) const;
 
+  /// \returns the function holding this node.
+  /// If that node does not belong to any function, this
+  /// is nullptr.
+  Function *getParent() const { return parent_; }
```
rdzhabarov: since it's a const method, I think it's a bit confusing to return internals of the Node in here. (is it possible to make that const Function *?)

include/glow/Graph/Node.h
```diff
@@ -257,6 +261,13 @@ class Node : public Named,
...
+  /// Set the link to the function that holds this node.
+  void setParent(Function *parent) { parent_ = parent; }
+
```
opti-mix: I wonder if `setParent` should really be a public function. 

IMHO, it could be safer if it would be private and Graph is a friend of Node and can invoke setParent. If it stays public it is possible to set it without actually moving the node to the nodes list of the parent function.

opti-mix:
@qcolombet Should we also add `removeFromParent` and `eraseFromParent` to the `Node` class like we have e.g. for instructions?

opti-mix:
Another question: For `glow::Instruction` we set parent links in `struct ilist_traits<glow::Instruction>` in IR.h. May be we should follow a similar approach for `Node`?

include/glow/Graph/Node.h
qcolombet: We could. Technically, changing the function does not change the node and returning a pointer does not change the value of parent_ either.
Anyhow, I agree it can be confusing.

include/glow/Graph/Graph.h
opti-mix: We have `eraseNode`

opti-mix: And may be we should have `removeNode` as well.

qcolombet:
> Should we also add removeFromParent and eraseFromParent to the Node class like we have e.g. for instructions?

@opti-mix, given we don't have a use case for these now, I would say no.

> Another question: For glow::Instruction we set parent links in struct ilist_traits<glow::Instruction> in IR.h. May be we should follow a similar approach for Node?

Sure.

include/glow/Graph/Node.h
qcolombet: I've followed what Instruction does. We also have a public setParent in the class Instruction and it is not protected at all :).

Given @nadavrot comment on NodeHandle, I don't think we want the extract protection.
In particular, notice that I've updated the Node::verify and that will fail if we attach a node to a function that it didn't belong too.

opti-mix:
>> Should we also add removeFromParent and eraseFromParent to the Node class like we have e.g. for instructions?

> given we don't have a use case for these now, I would say no.

Semantically, we almost have it in the DCE in GraphOptimizer.cpp. Of course, due to the lack of these functions, it is currently using `F->eraseNode(it);` instead of `it->eraseFromParent()`. But I don't have a strong opinion here. I just wanted Node and Instruction to be more similar.

tests/unittests/graphTest.cpp
```diff
@@ -549,3 +549,46 @@ TEST(Graph, schedulingOfSaves) {
...
+
+  // Cloned nodes don't belong to anything.
+  auto *clonedAddNode = addNode->clone();
+  EXPECT_EQ(clonedAddNode->getParent(), nullptr);
+
+  // Check that the setter properly sets things.
+  clonedAddNode->setParent(F);
+  EXPECT_EQ(clonedAddNode->getParent(), F);
+  clonedAddNode->setParent(nullptr);
+  EXPECT_EQ(clonedAddNode->getParent(), nullptr);
```
opti-mix: Add an example for `eraseNode` and check that the parent is reset to nullptr.

include/glow/Graph/Node.h
opti-mix: Ah, OK. Never mind then ;-)

include/glow/Graph/Node.h
qcolombet: (Bottom line I am going to use const :))

qcolombet: @rdzhabarov, looking at the IR Instruction, the getParent method has exactly the same signature:
  /// \returns parent of current instruction.
  IRFunction *getParent() const { return F_; }

I can do the clean-up in both places in a follow-up PR.

What do you think?

qcolombet:
>[...] it is currently using F->eraseNode(it); instead of it->eraseFromParent(). But I don't have a strong opinion here. I just wanted Node and Instruction to be more similar.

Ah, a use case :).
I'll do that in my next commit then.

tests/unittests/graphTest.cpp
opti-mix: `Graph::eraseNode` exists today, so I'd suggest we should add the check for it in this PR.

include/glow/Graph/Node.h
rdzhabarov: I'd fix here, just on a separate commit, but up to you. Good catch!

qcolombet:
Good point! I keep forgetting that PRs don’t squash the commits!

nadavrot:
@qcolombet Thank you for doing this work. 

I think that it would be useful to add a check to the verifier that all nodes in a graph belong to the function that owns them.

qcolombet:
Good catch @nadavrot!

We indeed check the node to function link, but not the other way around.
Fixed by #1297.

PR #1292, opened 2018-07-19
[Caffe2 importer] Improve weights loading support.

artemrakhov:


lib/Importer/Caffe2.cpp
```diff
@@ -391,9 +393,13 @@ void caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
       }
 
       auto dim = getShape(dict["shape"]);
+      if (dim.empty())
```
rdzhabarov: does onnx have the same behavior?

jfix71:
I'm not sure it's actually supposed to be single dimensional = `{1}`, at least not in the case I'm seeing. For me it's occurring from operators in the predict_net, not being created as a weight from init_net as you have here.

For my case, a 1d tensor is squeezed into having zero dimensions. I looked at the [C2 code for squeeze](https://github.com/caffe2/caffe2/blob/master/caffe2/operators/expand_squeeze_dims_op.h#L70) and it looks like in this case it would pass an empty dims vector to reshape, and [reshape](https://github.com/caffe2/caffe2/blob/master/caffe2/core/tensor.h#L326) basically just sets the new size to 1 and sets the tensor's dims to this empty dims vector.

The corresponding part of the code in ONNX, which is the source of the C2 proto, does:
```
%313 : Dynamic = onnx::Shape(%92)
%314 : Dynamic = onnx::Slice[axes=[0], ends=[1], starts=[0]](%313)
%315 : Long() = onnx::Squeeze[axes=[0]](%314)
```

I.e. `%313` is a 1D tensor, which is sliced into having a single element (`%314`), which is squeezed into a `Long` (`%315`). Also, [the 0.4 release notes](https://github.com/pytorch/pytorch/releases) talked about adding zero-dim tensors. So it seems to me like they really are meant to have no dimensions.

lib/Importer/Caffe2.cpp
beicy: ONNX has zero-dim tensor with dim_size() = 0, and it is viewed as a scalar. In ONNX doc, the zero-dim is written like: "shape(A) = (2, 3, 4, 5), shape(B) = (,), i.e. B is a scalar ==> shape(result) = (2, 3, 4, 5)".

artemrakhov:
I have removed `dim = {0}` part. Created a separate PR to support 0 dimensional Tensors.

lib/Importer/Caffe2.cpp
```diff
@@ -445,6 +449,42 @@ void caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
+      const auto &name = op.output(0);
+      auto *T = new Tensor();
+      tensors_[name] = T;
+      auto dim = getShape(dict["shape"]);
+      T->reset(ElemKind::FloatTy, dim);
+      auto TH = T->getHandle<>();
+      float tensorMin = loadFloat(dict["min"]);
+      float tensorMax = loadFloat(dict["max"]);
+
+      // Uniformly generate numbers in [tensorMin; tensorMax], inclusive.
```
jfix71: Isn't this supposed to be random, with a uniform distribution? https://github.com/caffe2/caffe2/blob/master/caffe2/operators/filler_op.h#L124

artemrakhov: Yes. But can we simulate Caffe2's random generator? Results will be different anyway.

I thought that for the purpose of testing, having this code is as good as having our own random generator in here. And I prefer not to have any randomness, to help with maintenance.

bertmaher: Our PseudoRNG is deterministically random so I don't think it'd cause maintenance issues.

jfix71: Their random generator, at least for the CPU, is just `std::uniform_real_distribution()`: https://github.com/caffe2/caffe2/blob/master/caffe2/utils/math_cpu.cc#L831

artemrakhov:
Changed UniformFill to generate random numbers. Also added a warning: it may lead to discrepancy between C2 and Glow loader.

lib/Importer/Caffe2.cpp
```diff
@@ -445,6 +449,44 @@ void caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
+      const auto &name = op.output(0);
+      auto *T = new Tensor();
+      tensors_[name] = T;
+      auto dim = getShape(dict["shape"]);
+      T->reset(ElemKind::FloatTy, dim);
+      auto TH = T->getHandle<>();
+      float tensorMin = loadFloat(dict["min"]);
+      float tensorMax = loadFloat(dict["max"]);
+
+      llvm::outs() << "The model contains UniformFill operator, which generates"
```
jfix71: Do we want to always print this? Maybe use `NDEBUG` to only print on Debug builds?

artemrakhov: Done.

PR #1293, opened 2018-07-19
[Optimizer] Fix `deleteDeadAllocs`

boncheolgu:
This fix enables `deleteDeadAllocs` to remove chained dead `TensorViewInst` instructions.

### DIFF
--- a/lib/Optimizer/IROptimizer.cpp
+++ b/lib/Optimizer/IROptimizer.cpp
@@ -344,43 +344,34 @@ static void sinkTensorViews(IRFunction &M) {
 static void deleteDeadAllocs(IRFunction &M) {
   auto &instrs = M.getInstrs();
 
-  llvm::SmallVector<Instruction *, 16> erasedInstructions{};
-
-  // Remove all unused tensorviews.
-  for (auto &I : instrs) {
-    if (isa<TensorViewInst>(I) && I.getNumUsers() == 0) {
-      erasedInstructions.push_back(&I);
+  // Remove all unused tensor views tracking back their dependencies, which are
+  // in a topological order.
+  // Note that this should precede to remove dependencies on allocs.
+  for (auto it = instrs.rbegin(); it != instrs.rend();) {
+    // Remember the current instruction and advance the iterator.
+    auto *I = &*it++;
+    if (isa<TensorViewInst>(I) && I->getNumUsers() == 0) {
+      // Remove a tensor view. It may make other tensor views preceding it
+      // eligible for a removal as well.
+      M.eraseInstruction(I);
     }
   }
 
-  for (auto I : erasedInstructions) {
-    M.eraseInstruction(I);
-  }
+  // Remove all of unused allocs and their corresponding deallocs.
+  // Iterate instructions in a reverse order to erase deallocs before
+  // their respective allocs, otherwise `DeallocActivationInst::getAlloc()` will
+  // return erased allocs.
+  for (auto it = instrs.rbegin(); it != instrs.rend();) {
+    auto *I = &*it++;
 
-  erasedInstructions.clear();
-
-  // Remove all of the DeallocActivationInst that close unused allocs.
-  for (auto &I : instrs) {
-    const auto *DA = dyn_cast<const DeallocActivationInst>(&I);
+    const auto *DA = dyn_cast<const DeallocActivationInst>(I);
     if (DA && DA->getAlloc()->getNumUsers() < 2) {
-      erasedInstructions.push_back(&I);
+      M.eraseInstruction(I);
+      continue;
     }
-  }
-
-  for (auto I : erasedInstructions) {
-    M.eraseInstruction(I);
-  }
-  erasedInstructions.clear();
-
-  // Remove the unused allocs.
-  for (auto &I : instrs) {
-    if (isa<AllocActivationInst>(&I) && I.getNumUsers() < 2) {
-      erasedInstructions.push_back(&I);
+    if (isa<AllocActivationInst>(I) && !I->hasUsers()) {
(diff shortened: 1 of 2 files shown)


opti-mix:
@boncheolgu Thanks for the PR! The idea makes sense.

Currently, this PR re-scans the whole list of instructions on every iteration of the while-loop. 

I think it should be possible to do it more efficiently. If I understand correctly, you only need to re-process the operand of the TensorView being removed, because this operand may be another TensorView, which becomes unused after the removal of the current one. A simple worklist algorithm or something like that should easily handle this.

boncheolgu:
@opti-mix Thanks for the comment. I removed the while loop.

lib/Optimizer/IROptimizer.cpp
```diff
@@ -347,18 +347,14 @@ static void deleteDeadAllocs(IRFunction &M) {
   llvm::SmallVector<Instruction *, 16> erasedInstructions{};
 
   // Remove all unused tensorviews.
-  for (auto &I : instrs) {
-    if (isa<TensorViewInst>(I) && I.getNumUsers() == 0) {
-      erasedInstructions.push_back(&I);
+  for (auto it = instrs.rbegin(); it != instrs.rend();) {
+    if (isa<TensorViewInst>(*it) && it->getNumUsers() == 0) {
```
opti-mix: I think this if-then-else-statement could be simplified. May be something like this:

```c++
for (auto it = instrs.rbegin(); it != instrs.rend();) {
  // Remember the current instruction and advance the iterator.
  auto *I = &*it++;
  // Erase unused tensor views.
  if (isa<TensorViewInst>(I) && I->getNumUsers() == 0) {
    // Remove a tensor view. 
    // It may make other tensor views preceding it eligible for a removal as well.
    M.eraseInstruction(I);
  }
}
```

opti-mix: You could also provide a small comment explaining why we iterate in a reverse order.

tests/unittests/IROptTest.cpp
```diff
@@ -125,6 +125,33 @@ TEST(Optimizer, shareBuffers) {
...
+  Function *F = mod.createFunction("DeleteDeadViews");
+  IRFunction M(F);
+  IRBuilder bb(&M);
+
+  auto *input = bb.createWeightVar(glow::ElemKind::FloatTy, {1}, "input",
+                                   WeightVar::MutabilityKind::Constant);
+  auto *output = bb.createWeightVar(glow::ElemKind::FloatTy, {1}, "ouput",
+                                    WeightVar::MutabilityKind::Mutable);
+
+  auto tensor_view1 = bb.createTensorViewInst(
```
opti-mix: Is this `createTensorViewInst` clang-formatted? The indentation looks a bit strange.

tests/unittests/IROptTest.cpp
```diff
@@ -125,6 +125,33 @@ TEST(Optimizer, shareBuffers) {
...
+      mod.uniqueType(Type{glow::ElemKind::FloatTy, {1, 1}}), {0, 0});
+
+  bb.createTensorViewInst("tensor_view2", tensor_view1,
+                          mod.uniqueType(Type{glow::ElemKind::FloatTy, {1}}),
+                          {0});
+  bb.createCopyInst("copy", output, input);
+
+  optimize(M, CompilationMode::Infer, MockBackend());
+
+  // Check that all view instructions are eliminated, because they are never
```
opti-mix: all view instructions -> all tensor_view instructions

lib/Optimizer/IROptimizer.cpp
rdzhabarov: +1.

Also, now we can remove the worklist (erasedInstructions).

tests/unittests/IROptTest.cpp
rdzhabarov: nit: tensorView1

lib/Optimizer/IROptimizer.cpp
```diff
@@ -344,44 +344,30 @@ static void sinkTensorViews(IRFunction &M) {
...
-    M.eraseInstruction(I);
-  }
-  erasedInstructions.clear();
-
-  // Remove the unused allocs.
-  for (auto &I : instrs) {
-    if (isa<AllocActivationInst>(&I) && I.getNumUsers() < 2) {
-      erasedInstructions.push_back(&I);
+      M.eraseInstruction(I);
+    } else if (isa<AllocActivationInst>(I) && I->getNumUsers() < 2) {
```
opti-mix: I think that this condition `I->getNumUsers() < 2` can be formulated more precisely in this re-written version of the code. Since we perform a reverse iteration and first remove deallocs, when we reach the alloc it would have no uses at all, or?

And again, please explain before the loop why we iterate in a reverse order and that it means that deallocs are eliminated before their respective allocs.

boncheolgu: The original code removes deallocs first without explanation, But I think that it is to prevent `DeallocActivationInst::getAlloc()` from returning erased allocs. 
What do you mean by formulating the condition?

opti-mix: I mean that once the the dealloc is erased, the getNumUsers() of the alloc becomes 0. So, the condition may be written as:

```c++
     if (DA && DA->getAlloc()->getNumUsers() < 2) {
       M.eraseInstruction(I);
       continue;
     }
     if (isa<AllocActivationInst>(I) && !I->hasUsers()) {
       M.eraseInstruction(I); 
     }
```

Also note, that my example does not use `else`, as LLVM coding style prefers avoiding else if it possible by using early exist, etc.

nadavrot:
Nice!

PR #1294, opened 2018-07-19
[quantization][opencl]QuantizeAndDequantize

ZchiPitt:
Fix this test so that there are actual quantize and dequantize instructions executed, as they were optimized out previously.

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -928,21 +928,27 @@ TEST_P(InterpAndCPU, ScatterAssignQuantized) {
 }
 
 TEST_P(Operator, QuantizeAndDequantize) {
-  Tensor inputs(ElemKind::FloatTy, {1, 4});
-  inputs.getHandle() = {1, 1.2f, 0.5f, 1.3f};
-
   auto *A = mod_.createVariable(ElemKind::FloatTy, {1, 4}, "A",
                                 VisibilityKind::Public);
+  auto *B = mod_.createVariable(ElemKind::FloatTy, {1, 4}, "B",
+                                VisibilityKind::Public);
+  A->getPayload().getHandle() = {1, 1.2f, 0.5f, 1.3f};
+  B->getPayload().getHandle() = {1.8f, 5.2f, 3.5f, 11.3f};
 
   auto qType = mod_.uniqueType(ElemKind::Int8QTy, {1, 4}, 0.05, -138);
-  auto *quantize = F_->createQuantize("quantize", A, qType);
-  auto *dequantize = F_->createDequantize("dequantize", quantize);
+  auto *quantizeA = F_->createQuantize("quantize", A, qType);
+  auto *quantizeB = F_->createQuantize("quantize", B, qType);
+  auto *add = F_->createAdd("add", quantizeA, quantizeB);
+  auto *dequantize = F_->createDequantize("dequantize", add);
   auto *result = F_->createSave("save", dequantize);
+  auto *fpAdd = F_->createAdd("fpAdd", A, B);
+  auto *fpResult = F_->createSave("fpSave", fpAdd);
 
   EE_.compile(CompilationMode::Infer, F_);
-  EE_.run({A}, {&inputs});
+  EE_.run({}, {});
 
-  EXPECT_TRUE(inputs.isEqual(result->getVariable()->getPayload()));
+  EXPECT_TRUE(result->getVariable()->getPayload().isEqual(
+      fpResult->getVariable()->getPayload()));
 }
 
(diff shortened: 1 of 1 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -928,21 +928,35 @@ TEST_P(InterpAndCPU, ScatterAssignQuantized) {
...
+  Tensor input2(ElemKind::FloatTy, {1, 4});
+  input1.getHandle() = {1, 1.2f, 0.5f, 1.3f};
+  input2.getHandle() = {1.9f, 5.2f, 3.5f, 6.3f};
 
   auto *A = mod_.createVariable(ElemKind::FloatTy, {1, 4}, "A",
                                 VisibilityKind::Public);
+  auto *B = mod_.createVariable(ElemKind::FloatTy, {1, 4}, "B",
+                                VisibilityKind::Public);
 
   auto qType = mod_.uniqueType(ElemKind::Int8QTy, {1, 4}, 0.05, -138);
```
rdzhabarov: make sure that values from the fp32 variables can be represented by {s,o} = {0.05, -138}

tests/unittests/OperatorTest.cpp
```diff
@@ -928,21 +928,35 @@ TEST_P(InterpAndCPU, ScatterAssignQuantized) {
...
-  auto *quantize = F_->createQuantize("quantize", A, qType);
-  auto *dequantize = F_->createDequantize("dequantize", quantize);
-  auto *result = F_->createSave("save", dequantize);
+  auto *quantizeA = F_->createQuantize("quantize", A, qType);
+  auto *quantizeB = F_->createQuantize("quantize", B, qType);
+  auto *int8Add = F_->createAdd("int8Add", quantizeA, quantizeB);
+  auto *dequantize = F_->createDequantize("dequantize", int8Add);
+  auto *int8Result = F_->createSave("int8Save", dequantize);
```
rdzhabarov: a bit misleading name as the result in fact fp32 here.

tests/unittests/OperatorTest.cpp
```diff
@@ -928,21 +928,35 @@ TEST_P(InterpAndCPU, ScatterAssignQuantized) {
...
+  auto *fpResult = F_->createSave("fpSave", fpAdd);
 
   EE_.compile(CompilationMode::Infer, F_);
-  EE_.run({A}, {&inputs});
+  EE_.run({A, B}, {&input1, &input2});
+
+  Tensor int8Out;
+  Tensor fpOut;
+  
+  int8Out.copyFrom(&int8Result->getVariable()->getPayload());
```
rdzhabarov: do not need to copy around, just use the payload from "save" nodes.

tests/unittests/OperatorTest.cpp
ZchiPitt: changed to the same values as in input2, so i can ensure they can be dequantzied

tests/unittests/OperatorTest.cpp
ZchiPitt: removed the prefix

tests/unittests/OperatorTest.cpp
ZchiPitt: done

tests/unittests/OperatorTest.cpp
rdzhabarov: my point is that {0.05, -138} covers floating point range [0.05*(-128 + 138), 0.05 *(127+138)].
I was just asking to make sure that the values from input1 and input2 are within this range :).
I think it's better to have different values for input1 and input2 though.

tests/unittests/OperatorTest.cpp
```diff
@@ -928,21 +928,30 @@ TEST_P(InterpAndCPU, ScatterAssignQuantized) {
...
 
 TEST_P(Operator, QuantizeAndDequantize) {
-  Tensor inputs(ElemKind::FloatTy, {1, 4});
-  inputs.getHandle() = {1, 1.2f, 0.5f, 1.3f};
+  Tensor input1(ElemKind::FloatTy, {1, 4});
+  Tensor input2(ElemKind::FloatTy, {1, 4});
+  input1.getHandle() = {1, 1.2f, 0.5f, 1.3f};
+  input2.getHandle() = {1, 1.2f, 0.5f, 1.3f};
 
   auto *A = mod_.createVariable(ElemKind::FloatTy, {1, 4}, "A",
```
rdzhabarov: super nit: you can remove Tensor input1 and just directly initialize variable A,
A->getPayload().getHandle()={1, 1.2f, 0.5f, 1.3f}.
And if so when you run E_.run just omit additional initialization of tensors.

nadavrot:
cool!

Issue #1190, opened 2018-06-28
Add support for ONNXIFI interface


rdzhabarov:
Support [ONNXIFI interface](https://github.com/onnx/onnx/blob/master/docs/ONNXIFI.md)

rdzhabarov:
https://github.com/pytorch/glow/pull/1369
Putting more details into the implementation: define event, backendid, backend, graph and other entities required for implementing onnxifi interface.

yinghai:
One of the last remaining issue is `onnxGetBackendCompatibility` tries to do a full parse on the input onnx model, which like to lead to failure because we use onnx model as a vessel for op info. It usually only has one node in the model and no input/output. We can add input/output but doing a full parse for each `onnxGetBackendCompatibility` is likely inefficient. Plus in the end, we are going to do the full parse on the whole model anyway. It's better that we have a short-cut for this. (ref: https://github.com/onnx/onnx/blob/219aaf91f3ae62762628aff4f68d84412202bcac/onnx/onnxifi.h#L1032-L1036) 

BTW, I noticed that in the importer, we have a lot of `assert` which will just crash the program. Why don't we throw?

rdzhabarov:
In order to create a specific Op Glow needs to know all inputs to the Op ahead of time (that's why parsing of inputs happens). In the compatibility case, we do not actually allocate memory for input tensors (for tensor elements) so that should have very low overhead. In addition that does not happen on a hot path.
Would it be problematic to add inputs to the Op on compatibility mode call?

> assert 

Those are removed in release build, but used in debug builds

yinghai:
I added inputs but it failed at the same place because it's looking for weight but we didn't supply it in the initialization list. So we still need to avoid full parse. 

As per `assert`, I don't think it's a proper way to check things here. `assert` means something that should not happen. This might not be the case when we check input because we cannot control it. In debug build, it will crash the program. In the opt build, it will mask the issue, and may lead to later crashes if your predicate doesn't hold.

yinghai:
CC @nadavrot on use of `assert`. In Glow code base, what's the policy of using `assert`. Or do we allow throwing exceptions at all?

opti-mix:
@rdzhabarov @Yangqing May be `GLOW_ASSERT` should be used? This one is always checked at run-time, even in Release builds. And it aborts if the assert fails. At least it does not mask the issue and does not let the program to proceed.

yinghai:
[...]

PR #1295, opened 2018-07-19
[onnx] Make ONNX loader more robust

rdzhabarov:
This PR is towards fixing https://github.com/pytorch/glow/issues/1190.

****
Currently, there is an issue with the way how the graph is loaded and the expectation of all input tensors to be initialized.
****

As part of the ONNXIFI interface, Glow would be required to test whether a specific model/node is supported by the backend.

There are two pieces involved:
1) Make sure that Glow supports particular ONNX graph
2) Make sure that specific backend supports operators from the graph

This PR makes checking (1) easier. Note, as part of the ONNXIFI interface there will be calls on the compatibility of the graph and backend. The graph is represented as a pointer to serialized proto + size.

For the testing, as of now, I made sure that the ONNX loader works fine as part of the ImageClassifier.

### DIFF
--- a/include/glow/Importer/ONNX.h
+++ b/include/glow/Importer/ONNX.h
@@ -43,18 +43,43 @@ class ONNXModelLoader
   void setVersion(onnx::ModelProto MP);
 
   /// Load the network operators from the GraphProto.
-  void loadNetwork(onnx::GraphProto &net);
+  /// \returns true if network can be loaded.
+  bool loadNetwork(onnx::GraphProto &net);
+
+  /// Set the root_ as an output node.
+  void setOutputNode(onnx::GraphProto &net);
 
   /// Load the network initializers from the GraphProto.
   void loadInitializers(onnx::GraphProto &net);
 
+  /// Load the inputs from the GraphProto. This is useful when the
+  /// initializers are not available.
+  void loadInputs(onnx::GraphProto &net);
+
+  /// \returns true if operator \p op can be loaded.
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  void loadOperator(const onnx::NodeProto &op);
+  bool loadOperator(const onnx::NodeProto &op);
 
-  /// Reads a network (weights or structure) from the serialized protocol buffer
-  /// file.
-  bool loadProtoFile(onnx::GraphProto &net, const std::string &filename);
+  /// \returns true if \p net can be constructed from the content of the
+  /// file \p filename.
+  /// Loads GraphProto \p net from the file containing serialized protobuf.
+  bool loadProto(onnx::GraphProto &net, const std::string &filename);
+
+  /// \returns true if \p net can be constructed from the in-memory
+  /// serialized protobuf.
+  /// Loads GraphProto \p net from the in-memory serialized protobuf \p
+  /// onnxModel with the model size \p onnxModelSize.
+  bool loadProto(onnx::GraphProto &net, const void *onnxModel,
+                 size_t onnxModelSize);
+
+  /// \returns true if GraphProto \p net can be loaded from the stream \p
+  /// iStream.
+  bool loadProto(onnx::GraphProto &net,
+                 google::protobuf::io::ZeroCopyInputStream &iStream);
+
+  /// Creates a ONNX model loader to build \p F.
+  ONNXModelLoader(Function &F);
 
   /// ONNX model ir_version;
   size_t irVersion_;
@@ -70,6 +95,12 @@ public:
   ONNXModelLoader(const std::string &modelDescFilename,
                   llvm::ArrayRef<const char *> tensorNames,
                   llvm::ArrayRef<Tensor *> tensors, Function &F);
+
+  /// \returns unique pointer to ONNXModelLoader if \p onnxModel can be parsed,
+  /// e.g., the model is a valid ONNX model and Glow supports all of the
+  /// operators in the network. \returns nullptr otherwise.
+  static std::unique_ptr<ONNXModelLoader>
+  parse(const void *onnxModel, size_t onnxModelSize, Function &F);
 };
 
 } // namespace glow
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -72,22 +72,34 @@ void ONNXModelLoader::setVersion(onnx::ModelProto MP) {
               "The opset of this ONNX model is not supported.");
 }
 
-bool ONNXModelLoader::loadProtoFile(onnx::GraphProto &net,
-                                    const std::string &filename) {
+bool ONNXModelLoader::loadProto(
+    onnx::GraphProto &net, google::protobuf::io::ZeroCopyInputStream &iStream) {
+  // Construct and configure a Coded Input Stream
+  google::protobuf::io::CodedInputStream codedStream(&iStream);
+
+  // Don't warn about large file sizes.
+  codedStream.SetTotalBytesLimit(1e+9, 1e+9);
+  onnx::ModelProto MP;
+  bool parseNet = MP.ParseFromCodedStream(&codedStream);
+  net = MP.graph();
+  setVersion(MP);
+
+  return parseNet;
+}
+
+bool ONNXModelLoader::loadProto(onnx::GraphProto &net, const void *onnxModel,
+                                size_t onnxModelSize) {
+  google::protobuf::io::ArrayInputStream arrayStream(onnxModel, onnxModelSize);
+  return loadProto(net, arrayStream);
+}
+
+bool ONNXModelLoader::loadProto(onnx::GraphProto &net,
+                                const std::string &filename) {
   std::ifstream ff(filename, std::ios::in | std::ios::binary);
   GLOW_ASSERT(ff && "Can't find the model or network files.");
 
-  // Construct and configure a Coded Input Stream
-  google::protobuf::io::IstreamInputStream filestr(&ff);
-  google::protobuf::io::CodedInputStream codedstr(&filestr);
-  // Don't warn about large file sizes.
-  codedstr.SetTotalBytesLimit(1e+9, 1e+9);
-  onnx::ModelProto MP;
-  bool parseNet = MP.ParseFromCodedStream(&codedstr);
-  net = MP.graph();
-  setVersion(MP);
-  GLOW_ASSERT(parseNet && "Failed to parse the network descriptor.");
-  return true;
+  google::protobuf::io::IstreamInputStream fileStream(&ff);
+  return loadProto(net, fileStream);
 }
 
 std::vector<size_t> getPads(const ArgumentDictionaryTy &dict) {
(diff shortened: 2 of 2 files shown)


include/glow/Importer/ONNX.h
```diff
@@ -41,22 +41,41 @@ class ONNXModelLoader
...
   ONNXModelLoader(const std::string &modelDescFilename,
-                  llvm::ArrayRef<const char *> names,
+                  llvm::ArrayRef<const char *> tensorNames,
                   llvm::ArrayRef<Tensor *> tensors, Function &F);
+
+  /// \returns unique pointer to ONNXModelLoader if \p onnxModel can be parsed, e.g., the model is the
+  /// valid ONNX model and Glow supports all of the operators in the network.
+  /// \returns nullptr otherwise.
+  static std::unique_ptr<ONNXModelLoader> parse(void const *onnxModel, size_t onnxModelSize, Function &F);
 };
```
beicy: Is there any test to use this function so far?

rdzhabarov: Yeah, not now. I wish we'd have general Loader test coverage.
Once i'll make more progress on ONNXIFI we'll have integration test calling in here.

rdzhabarov:
Just pushed the version with resolved conflicts on ONNX.(h|cpp) files.

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// \p tensorNames and used as inputs to the network.
   ONNXModelLoader(const std::string &modelDescFilename,
-                  llvm::ArrayRef<const char *> names,
+                  llvm::ArrayRef<const char *> tensorNames,
                   llvm::ArrayRef<Tensor *> tensors, Function &F);
+
+  /// \returns unique pointer to ONNXModelLoader if \p onnxModel can be parsed, e.g., the model is the
+  /// valid ONNX model and Glow supports all of the operators in the network.
+  /// \returns nullptr otherwise.
+  static std::unique_ptr<ONNXModelLoader> parse(void const *onnxModel, size_t onnxModelSize, Function &F);
```
qcolombet: I think we want `const void *`, like you did on the definition in ONNX.cpp.

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// \returns true if serialized protobuf can be read from the \p filename.
+  /// Loads GraphProto \p net containing network (weights or structure)
+  /// from the serialized protobuf file.
+  bool loadProto(onnx::GraphProto &net, const std::string &filename);
+
+  /// \returns true if serialized protobuf can be constructed from the in-memory
+  /// serialized protobuf.
+  /// Loads GraphProto \p net from the in-memory serialized protobuf \p onnxModel
+  /// with the model size \p onnxModelSize.
+  bool loadProto(onnx::GraphProto &net, void const *onnxModel, size_t onnxModelSize);
```
qcolombet: const void *

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// \returns true if operator \p op can be loaded.
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  void loadOperator(const onnx::NodeProto &op);
+  bool loadOperator(const onnx::NodeProto &op);
 
-  /// Reads a network (weights or structure) from the serialized protocol buffer
-  /// file.
-  bool loadProtoFile(onnx::GraphProto &net, const std::string &filename);
+  /// \returns true if serialized protobuf can be read from the \p filename.
```
qcolombet: if *the* serialized protobuf

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// Loads GraphProto \p net containing network (weights or structure)
+  /// from the serialized protobuf file.
+  bool loadProto(onnx::GraphProto &net, const std::string &filename);
+
+  /// \returns true if serialized protobuf can be constructed from the in-memory
```
qcolombet: *the* serialized

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  void loadOperator(const onnx::NodeProto &op);
+  bool loadOperator(const onnx::NodeProto &op);
 
-  /// Reads a network (weights or structure) from the serialized protocol buffer
-  /// file.
-  bool loadProtoFile(onnx::GraphProto &net, const std::string &filename);
+  /// \returns true if serialized protobuf can be read from the \p filename.
+  /// Loads GraphProto \p net containing network (weights or structure)
```
qcolombet: I am not sure I understand the meaning here.
Maybe `loads into \p net the network...`

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// from the serialized protobuf file.
+  bool loadProto(onnx::GraphProto &net, const std::string &filename);
+
+  /// \returns true if serialized protobuf can be constructed from the in-memory
+  /// serialized protobuf.
```
qcolombet: The first sentence sounds strange, because we construct a serialized protobuf from a serialized protobuf. That should always work.
I am guessing we meant to say that we manage to build \p net from the in-memory protobuf.

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
+  /// Loads GraphProto \p net from the in-memory serialized protobuf \p onnxModel
+  /// with the model size \p onnxModelSize.
+  bool loadProto(onnx::GraphProto &net, void const *onnxModel, size_t onnxModelSize);
+
+  /// \returns true if GraphProto \p net can be loaded from the stream \p iStream.
+  bool loadProto(onnx::GraphProto &net, google::protobuf::io::ZeroCopyInputStream& iStream);
+
+  /// Creates the ONNX model loader with the network which will be stored
+  /// in the \p F.
```
qcolombet: Suggestion:
Creates a ONNX model loader to build \p F.

Otherwise, no *the* before \p F.

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,47 @@ class ONNXModelLoader
...
   /// The tensors in \p tensors are stored with the names in the list of names
-  /// \p names and used as inputs to the network.
+  /// \p tensorNames and used as inputs to the network.
   ONNXModelLoader(const std::string &modelDescFilename,
-                  llvm::ArrayRef<const char *> names,
+                  llvm::ArrayRef<const char *> tensorNames,
                   llvm::ArrayRef<Tensor *> tensors, Function &F);
+
+  /// \returns unique pointer to ONNXModelLoader if \p onnxModel can be parsed, e.g., the model is the
+  /// valid ONNX model and Glow supports all of the operators in the network.
```
qcolombet: the model is *a* valid ONNX model.

include/glow/Importer/ONNX.h
rdzhabarov: indeed!

include/glow/Importer/ONNX.h
```diff
@@ -48,28 +48,46 @@ class ONNXModelLoader
...
-
 public:
   /// Loads the ONNX model that's represented by a model description file,
   /// serialized in \p modelDescFilename and populates the network into \p F.
   /// The tensors in \p tensors are stored with the names in the list of names
-  /// \p names and used as inputs to the network.
+  /// \p tensorNames and used as inputs to the network.
   ONNXModelLoader(const std::string &modelDescFilename,
-                  llvm::ArrayRef<const char *> names,
+                  llvm::ArrayRef<const char *> tensorNames,
```
artemrakhov: Maybe it's worth to change names -> tensorNames in all 3 constructors: also in ProtobufLoader and Caffe2ModelLoader.

lib/Importer/ONNX.cpp
```diff
@@ -427,21 +437,47 @@ void ONNXModelLoader::loadNetwork(onnx::GraphProto &net) {
   /// Load the network operators:
   for (int i = 0; i < net.node_size(); i++) {
     auto &op = net.node(i);
-    loadOperator(op);
+    if (!loadOperator(op)) {
+      unexpectedNodeError(op, "Unsupported operator.");
+    }
   }
 
   assert(net.output_size() && "Network needs external outputs defined.");
```
artemrakhov: Can we move this code, so that `loadNetwork` could be used in both cases?

rdzhabarov:
@qcolombet thanks for comments!

lib/Importer/ONNX.cpp
```diff
@@ -427,21 +440,49 @@ void ONNXModelLoader::loadNetwork(onnx::GraphProto &net) {
...
+                                                        size_t onnxModelSize,
+                                                        Function &F) {
+  std::unique_ptr<ONNXModelLoader> onnxLoader(new ONNXModelLoader(F));
+
+  onnx::GraphProto modelDef;
+  if (!onnxLoader->loadProto(modelDef, onnxModel, onnxModelSize)) {
+    return nullptr;
+  }
+
+  for (int i = 0; i < modelDef.node_size(); i++) {
```
artemrakhov: How does this work? Note that we eagerly work with Weights now. We extract them from tensors_ map. Who will populate it for `onnxLoader`?

rdzhabarov: The parsing part is needed in order to verify that certain operators can be loaded.
We do not need to provide the tensor values at this moment.
As long as operators can be loaded and then Glow backend (CPU for example) confirms that it can execute it we should be good.

include/glow/Importer/ONNX.h
rdzhabarov: Yup, will change in the follow-up PR.

lib/Importer/ONNX.cpp
artemrakhov: Could you explain what happens upon call to `tensors_` map? Say we have FC in our model. Then `loadFC()` is called. Then it calls `getTensorByName()` to get weights. Which in turn asserts in Debug mode, or continues with `Tensor* = NULL` in Release mode, right?

rdzhabarov: That seems to be correct without explicit initializers this would fail. Looks like we need to simulate those tensors artificially. (Note, initializers are not provided as part of the ONNXIFI call).

Surgical intervention into loader was not successful!

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -593,7 +593,7 @@ void OpenCLFunction::execute() {
                     I.getOperand(2).first->getType()->getScale() / destScale,
                 0);
             setKernelArg(kernel, ++numArgs, resultScaleParams);
-          } 
+          }
```
rdzhabarov: ah, this is annoying. Will fix it first.

include/glow/Importer/ONNX.h
rdzhabarov: fixed it here: https://github.com/pytorch/glow/pull/1301

lib/Importer/ONNX.cpp
rdzhabarov: fixed this

include/glow/Importer/ONNX.h
```diff
@@ -43,18 +43,38 @@ class ONNXModelLoader
   void setVersion(onnx::ModelProto MP);
 
   /// Load the network operators from the GraphProto.
```
nadavrot: Please add doxygen:

//returns True if one two three ..

PR #1298, opened 2018-07-19
[Operators] Allow Tensor to have 0 dimensions.

artemrakhov:
Now Tensor with 0 dimensions is valid thing, representing a scalar. Size=1 for this Tensor, and you can access element[0]

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -29,9 +29,9 @@ namespace {
 template <class ElemTy>
 static void libjit_dump_tensor_impl(ElemTy *tensor, size_t *dims,
                                     size_t numDims) {
-  // Check for empty tensor.
+  // Check for 0-dimensional tensor.
   if (!numDims) {
-    printf("[ Empty tensor ]\n");
+    printf("[ Scalar containing: %.3f ]\n", (float)tensor[0]);
     return;
   }
 
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -521,7 +521,7 @@ public:
   /// the first dimension).
   Tensor extractSlice(size_t idx) const {
     auto sizes = tensor_->dims();
-    assert(sizes.size() > 1 && "Tensor has only one dimension");
+    assert(sizes.size() > 1 && "Tensor must have at least two dimensions");
     assert(idx < sizes[0] && "Invalid first index");
 
     Tensor slice{Type::newShape(tensor_->getType(), sizes.slice(1))};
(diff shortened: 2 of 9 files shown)


jfix71:
Does dumping dot files still work correctly/look reasonable? Does it just show `<>` for the output type?

stoklund:
Some of the deleted comments seem to indicate that a 0-order tensor is the same as an empty tensor. It isn’t of course.

Can we still distinguish between an empty 0-order tensor and a non-empty one?

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -31,7 +31,7 @@ static void libjit_dump_tensor_impl(ElemTy *tensor, size_t *dims,
                                     size_t numDims) {
   // Check for empty tensor.
   if (!numDims) {
-    printf("[ Empty tensor ]\n");
+    printf("[ Empty tensor containing: %.3f ]\n", (float)tensor[0]);
```
nadavrot: Empty tensor containing:  -> Scalar containing:

artemrakhov:
@jfix71 : added a bunch of tests. Everything worked as expected.

Dumping produces `<void>` as output type.

Uncovered one corner case, not related to 0-dims:
```
TEST(Tensor, compareDiffentElemKinds) {
  Tensor T1(ElemKind::FloatTy, {2});
  Tensor T2(ElemKind::IndexTy, {2});
  T1.zero();
  T2.zero();

  EXPECT_TRUE(T1.isEqual(T2));
}
```
This passes in release, but fails in debug, because we don't explicitly compare ElemKind.

jfix71:
Cool, looking good!

I don't think I like something like `float<void>` as the output type -- if I saw that without context I think I'd think it's buggy. Maybe it could be something like `float`, or `float<>`, or `float (scalar)`, etc.?

And nice catch on the `isEqual()`. Should be a pretty simple fix of returning false on ElemKind mismatch?

artemrakhov:
@jfix71 : changed output format to `float<>`. I think it's the best, but we can change it later if it turns out to be confusing.

nadavrot:
@artemrakhov How do you access the scalar data? Using X.raw(0) is strange, and we really want to reduce the use of the unsafe Raw API.

artemrakhov:
@nadavrot : please take a look at tests, e.g. `ZeroDimensionalTensor.handleAt`. I test different access options there. I think the best one is to do `X.at({})`.

nadavrot:
@artemrakhov Ah, yes. Got it. Thank you. :)

jfix71:
Awesome, looks good to go to me.

artemrakhov:
Changed `AllocationsInfo::allocateTensorViews()` to account for the case when `Offsets()` returns empty vector. Otherwise `ninja test_unopt` was failing. Also need to test on OpenCL.

artemrakhov:
Also made change to `OpenCL.cpp`, so OpenCL tests also pass.

PR #1302, opened 2018-07-20
[quantization][opencl]: Support quantization for TransposeInst in OpenCL

ZchiPitt:
Add quantization support for TransposeInst in OpenCL


lib/Backends/OpenCL/kernels.cl
```diff
@@ -1102,9 +1102,36 @@ __kernel void transposeK_u(__global cl_uint64_t *dest,
 
 __kernel void transpose_uW(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
                          ShapeNHWC odim, ShapeNHWC idim, ShapeNHWC shuffle) {
-  transposeK_u(&mem[dest], &mem[src], odim, idim, shuffle);
+  transpose_uK(&mem[dest], &mem[src], odim, idim, shuffle);
 }
 
+__kernel void transpose_i8K(__global cl_int8_t *dest,
```
rdzhabarov: please avoid code duplication. Transpose of fp32 vs transpose of int8 should not be different. The only difference is a type.

ZchiPitt: so you mean i can write a macro to cover the cases for all these types

tests/unittests/OperatorTest.cpp
```diff
@@ -1346,6 +1346,23 @@ TEST_P(InterpAndCPU, QuantizedArithmeticRescaled) {
...
+                                VisibilityKind::Public);
+  auto *B = mod_.createVariable(ElemKind::FloatTy, {3, 2}, "B",
+                                VisibilityKind::Public);
+  A->getPayload().getHandle() = {1, 1.2f, 0.5f, 1.3f, 2.7f, 5.8f};
+  B->getPayload().getHandle() = {1, 1.3f, 1.2f, 2.7f, 0.5f, 5.8f};
+  auto qType = mod_.uniqueType(ElemKind::Int8QTy, {2, 3}, 0.05, -138);
+  auto *quantizeA = F_->createQuantize("quantize", A, qType);
+  auto *tr = F_->createTranspose("tr", quantizeA, {1, 0});
+  auto *dequantize = F_->createDequantize("dequantize", tr);
+  auto result = F_->createSave("ret", dequantize);
```
rdzhabarov: auto * for consistency

rdzhabarov:
LGTM, please fix the code duplication problem. Test looks great, thanks for adding this.

tests/unittests/OperatorTest.cpp
```diff
@@ -1346,6 +1346,23 @@ TEST_P(InterpAndCPU, QuantizedArithmeticRescaled) {
   }
 }
 
+TEST_P(Operator, QuantizedTranspose) {
+  auto *A = mod_.createVariable(ElemKind::FloatTy, {2, 3}, "A",
+                                VisibilityKind::Public);
+  auto *B = mod_.createVariable(ElemKind::FloatTy, {3, 2}, "B",
+                                VisibilityKind::Public);
+  A->getPayload().getHandle() = {1, 1.2f, 0.5f, 1.3f, 2.7f, 5.8f};
+  B->getPayload().getHandle() = {1, 1.3f, 1.2f, 2.7f, 0.5f, 5.8f};
```
opti-mix: Should we use `Tensor::transpose` here to get the payload of B instead of transposing by hand?

rdzhabarov: +1. Maybe even make the transpose operation as a node on fp32 (A var). This way you'll test transposes on both types for openCL and other backends.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -1056,54 +1056,40 @@ __kernel void oclpoolavgW(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
...
-    for (size_t d3 = 0; d3 < idim.c; d3++) {
-      res[3] = d3;
-      size_t dstIdx = getNHWC(odim, res[shuffle.n], res[shuffle.h],
-                              res[shuffle.w], res[shuffle.c]);
-      size_t srcIdx = getNHWC(idim, d0, d1, d2, d3);
-      dest[dstIdx] = src[srcIdx];
-    }
-  }
-}
+DEFINE_OPENCL_TRANSPOSE_KERNEL(_i8, cl_int8_t)
```
opti-mix: either pass a full name here or name the parameter of the macro `suffix` instead of `name`. Otherwise it looks a bit odd. I'd prefer passing a full name here, because `_i8` looks strange.

ZchiPitt: make sense. Changed to pass full name here.

tests/unittests/OperatorTest.cpp
ZchiPitt: use Tensor::transpose now. I also added a transpose operation on fp32 var A. It tests both types now.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -187,6 +187,7 @@ class OCLBackend final : public Backend {
       case Kinded::Kind::MulNodeKind:
       case Kinded::Kind::QuantizeNodeKind:
       case Kinded::Kind::SubNodeKind:
+      case Kinded::Kind::TransposeNodeKind:
```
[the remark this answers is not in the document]
ZchiPitt: i tried to run "end2end" on OpenCL. It can work now. But end2endGRU cannot as it still needs topKInst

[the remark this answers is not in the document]
rdzhabarov: yes, GRU will require few more operators (gather, etc)

PR #1303, opened 2018-07-20
[IROptimizer] Teach the shareBuffers opt how to extend a dest buffer …

qcolombet:
…when the source doesn't have a definition

Let us consider the following pattern:
= useA var
tmp = allocactivation
tmp = useB var

Prior to this patch, if we merge tmp into var while using tmp buffer
the first use of tmp will be undefined because var didn't have any
definition to be replace by tmp:

tmp = allocactivation
= useA tmp    <--- this use of tmp is undefined
tmp = useB tmp

This patch teachs the shareBuffers optimization how to recognize this situation
and insert the proper definition for the buffer:

tmp = allocactivation
tmp = copy var <-- tmp has now var's value
= useA tmp    <--- this use is now fine
tmp = useB tmp

### DIFF
--- a/lib/Optimizer/IROptimizer.cpp
+++ b/lib/Optimizer/IROptimizer.cpp
@@ -676,6 +676,37 @@ static bool hasOverlappingIntervals(Intervals &intervals, Interval I) {
   return false;
 }
 
+/// Helper function to get a compatible value to replace \p val with \p with.
+/// This function casts \p with if necessary. When a cast needs to be created
+/// it will be inserted before \p Before. Moreover, when that happens, the
+/// second element of the returned pair is true, false otherwise.
+///
+/// \returns A pair with the first element being the Value matching val's type,
+/// using \p with's content and the second element being the status of
+/// whether a cast was inserted or not.
+static std::pair<Value *, bool>
+getCompatibleValueForReplacement(IRBuilder &B, Instruction *Before,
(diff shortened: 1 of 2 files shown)


nadavrot:
@qcolombet Good catch!  Thank you for looking at this. This is probably the most sophisticated part of the compiler and it's great that you took over it.

qcolombet:
Talked with @opti-mix offline about maybe moving the firstUse "analysis" directly in the interval building.
After giving a bit more thoughts I don't think that's desirable, for three reasons:
- One and most importantly, we will have to do this kind of checks when we replace the uses, because we may create tensorviews that will change when the first use happens
- We'll need to add some logic for updating this information in the intervals, for instance, when we merge intervals through copy propagation or when the first use changes when inserting tensorviews (that part is also problematic in itself because new instructions are not added to the instruction numbering map, which is what the intervals use)
- Finally, although it seems we do this analysis over and over when we merge the buffers as opposed to once if we were to do it when we build the intervals, we won't save much because we have to loop over the interval to do the updates no matter what and in practice we just add a couple of comparisons

opti-mix:
@qcolombet Thanks for considering the alternative solution we discussed. 

Your code is looking good.

opti-mix:
BTW, an idea for a separate PR: May be we should move the shareBuffers implementation into a separate file as it's getting really big and complex over time.

PR #1304, opened 2018-07-20
[fix]Replace the use of nearbyintf() with round()

ZchiPitt:
Replace the use of nearbyintf() with round(), as it was not fully supported by OpenCL


ZchiPitt:
It always works perfect with OpenCL enabled on my mac😿 Hope this time it will also work on linux system.

PR #1305, opened 2018-07-20
[onnx] Add onnx as a submodule.

rdzhabarov:
That contain some onnxifi declarations as well as we could get the onnx.proto.

Also probably worth to consolidate submodules in ThirdParty folder (gtest for example).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "tests/googletest"]
 	path = tests/googletest
 	url = https://github.com/google/googletest.git
+[submodule "thirdparty/onnx"]
+	path = thirdparty/onnx
+	url = https://github.com/onnx/onnx.git


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "tests/googletest"]
 	path = tests/googletest
 	url = https://github.com/google/googletest.git
+[submodule "ThirdParty/onnx"]
```
nadavrot: Should we use upper case for the directory name? I think that all other directories are lower case.

rdzhabarov: @nadavrot It's pretty inconsistent in the code base, e.g., /lib has Backends, ExecutionEngine, etc.
I'm fine either way, thirdparty also would work for me. Let me know.

rdzhabarov: Yes, this sounds good!

Issue #1283, opened 2018-07-16
Wrong scheduling because of the lack of memory dependencies


qcolombet:
Our graph does not model the memory dependencies explicitly.
    As a result, if we have a variable is written and read on different
    data paths, the scheduler may be unlucky and we read the wrong
    version of the variable.
    
    Essentially, if you have:
         A
        / \
    write read
    
    The scheduler can produce either:
    write A
    read A
    or
    read A
    write A
    
    Obviously this does not produce the same output.
    Glow intended semantic is that write should happen last, but with
    our current algorithm, this is actually possible to fool the
    scheduler not to do that.

I'm going to send a PR to show how we can fool the scheduler.

qcolombet:
Problem illustrated in #1284.

This PR adds two test cases that build exactly the same graph. However, by just changing the order of the creation of the nodes, we can trick the scheduler to produce an invalid schedule.

qcolombet:
After talking with @nadavrot this morning, we are going to explicit state that save nodes happen after the last use of the related variable. In other words, referencing a variable will always get you the old value.

Similarly we are going to enforce that every variable has at most one save nodes. That way it will remove the ambiguity of which save should happen last (all the others would have been dead because of the other restriction.)

qcolombet:
As expected while reviewing #1176, I realize that our implicit memory dependencies are going to bite us again during graph partitioning.

The problem there is that the post order visiting methods don't see the memory dependencies edges. So, we will have to do one of the following fix the visiting scheme, come up with targeted fix for the graph partitioning stuff, or model the memory dependencies.

qcolombet:
Fixed by #1307.

PR #1307, opened 2018-07-20
[GraphScheduler] Honor memory dependencies implied by the SaveNode

qcolombet:
When a variable is used as output in a save node, this is expected
to be its last use. Make sure the GraphSchedule looks at the implicit
dependencies between all the uses of a variable and its save node
when walking the dependencies.

This fixes issue #1283.

### DIFF
--- a/lib/IR/GraphScheduler.cpp
+++ b/lib/IR/GraphScheduler.cpp
@@ -140,6 +140,27 @@ class ChildMemSizeBasedScheduler : public Scheduler {
       orderedChildren.push_back(N->getPredicate());
     }
 
+    // SaveNode hack:
+    // We don't model memory dependencies, but we still need to honor them.
+    // Make sure the SaveNode happens after the last use of the output variable.
+    if (auto *save = dyn_cast<SaveNode>(N)) {
+      Variable *output = save->getVariable();
+      for (NodeUse &use : output->getUsers()) {
+        Node *user = use.getUser();
+        if (user == save) {
+          continue;
+        }
+        // Variables may have users scattered across different functions.
+        // Only accounts for the ones in that function.
+        if (&G_ != user->getParent()) {
+          continue;
+        }
+        assert(!isa<SaveNode>(user) &&
+               "Variables must be saved at most once in each function");
+        orderedChildren.push_back(user);
+      }
+    }
+
     // Order children by (maxSize - resultSize). It gives more
     // priority to the nodes that free more memory after
     // their computation.
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -546,8 +546,7 @@ TEST(Graph, schedulingOfSaves) {
     }
   }
   EXPECT_TRUE(A->getPayload().isEqual(zero->getPayload(), 0.0));
-  // FIXME: This should be true, but right now it is not!
-  EXPECT_FALSE /*TRUE*/ (allEqual);
+  EXPECT_TRUE(allEqual);
 }
 
 /// Check that the parent link is properly updated while tweaking


lib/IR/GraphScheduler.cpp
```diff
@@ -140,6 +140,26 @@ class ChildMemSizeBasedScheduler : public Scheduler {
       orderedChildren.push_back(N->getPredicate());
     }
 
+    // SaveNode hack:
+    // We don't model memory dependencies, but we still need to honor them.
+    // Make sure the SaveNode happens after the last use of the output variable.
+    if (SaveNode *save = llvm::dyn_cast<SaveNode>(N)) {
```
jfix71: nit: this file is already `using llvm::dyn_cast`

lib/IR/GraphScheduler.cpp
```diff
@@ -140,6 +140,26 @@ class ChildMemSizeBasedScheduler : public Scheduler {
...
+        Node *user = use.getUser();
+        if (user == save) {
+          continue;
+        }
+        // Variables may have users scattered across different functions.
+        // Only accounts for the ones in that function.
+        if (&G_ != user->getParent()) {
+          continue;
+        }
+        assert(!llvm::isa<SaveNode>(user) && "More than one save");
```
jfix71: nit: same, already using

lib/IR/GraphScheduler.cpp
```diff
@@ -140,6 +140,26 @@ class ChildMemSizeBasedScheduler : public Scheduler {
...
+    // We don't model memory dependencies, but we still need to honor them.
+    // Make sure the SaveNode happens after the last use of the output variable.
+    if (SaveNode *save = llvm::dyn_cast<SaveNode>(N)) {
+      Variable *output = save->getVariable();
+      for (NodeUse &use : output->getUsers()) {
+        Node *user = use.getUser();
+        if (user == save) {
+          continue;
+        }
+        // Variables may have users scattered across different functions.
```
opti-mix: Sounds like it could be useful to have an iterator type that iterates only over users inside a specific function.

lib/IR/GraphScheduler.cpp
opti-mix: Use `auto *save` as we usually use this style with `if (.. dyn_cast<Ty>(expr))` in our code?

lib/IR/GraphScheduler.cpp
```diff
@@ -140,6 +140,26 @@ class ChildMemSizeBasedScheduler : public Scheduler {
       orderedChildren.push_back(N->getPredicate());
     }
 
+    // SaveNode hack:
+    // We don't model memory dependencies, but we still need to honor them.
+    // Make sure the SaveNode happens after the last use of the output variable.
+    if (SaveNode *save = llvm::dyn_cast<SaveNode>(N)) {
+      Variable *output = save->getVariable();
+      for (NodeUse &use : output->getUsers()) {
```
opti-mix: `const auto &`?

qcolombet: We cannot go with const here because we push non-const node to orderedChildren.
As for auto, I am not a fan of using it unless the type name is already nearby (like with the dyn_cast you've mentioned). To me getUsers does not obviously refers to NodeUse (espicially because getUser after that is Node :P).

lib/IR/GraphScheduler.cpp
```diff
@@ -140,6 +140,26 @@ class ChildMemSizeBasedScheduler : public Scheduler {
...
+        Node *user = use.getUser();
+        if (user == save) {
+          continue;
+        }
+        // Variables may have users scattered across different functions.
+        // Only accounts for the ones in that function.
+        if (&G_ != user->getParent()) {
+          continue;
+        }
+        assert(!isa<SaveNode>(user) && "More than one save");
```
nadavrot: "Variables must be saved at most one time in each function. "

nadavrot:
Nice!

PR #1308, opened 2018-07-20
[quantization][opencl]: Support quantization for SplatInst

ZchiPitt:
Support quantization for SplatInst on OpenCL backend


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -566,7 +566,9 @@ void OpenCLFunction::execute() {
 
       if (auto *SI = dyn_cast<SplatInst>(&I)) {
         // Pass the splat as a parameter.
-        setKernelArg(kernel, ++numArgs, SI->getValue());
+        if (!isQuantized) {
```
rdzhabarov: would be easier to follow if you handle both quantized and non-quantized splat here.

tests/unittests/OperatorTest.cpp
```diff
@@ -1556,14 +1556,17 @@ TEST_P(Operator, TestQuantizedRescaleSequence) {
   auto T4 = mod_.uniqueType(ElemKind::Int8QTy, {len}, 0.1, 7);
   auto T5 = mod_.uniqueType(ElemKind::Int8QTy, {len}, 0.3, -3);
 
+  auto T6 = mod_.uniqueType(ElemKind::FloatTy, {len});
```
rdzhabarov: why do you need to change this test?

ZchiPitt: ahh, you are right.  I was confused why it worked before my implementation, then i realized that previously it might just pass float value to the kernel

lib/Backends/OpenCL/OpenCL.cpp
ZchiPitt: changed

tests/unittests/OperatorTest.cpp
rdzhabarov: can you check why "RescaleNode" test does not fail (in the operatorTest).
Please, fix that and enable only on CPUandInterpreter after the fix.

rdzhabarov:
@ZchiPitt Please check the "RescaleNode" test. This PR looks good to me, but make sure to fix that on a separate PR.

tests/unittests/OperatorTest.cpp
ZchiPitt: sure

ZchiPitt: it seems that all the rescale nodes got optimized out before running

nadavrot:
Nice!

nadavrot:
🎉 🎂

PR #1311, opened 2018-07-23
[GraphIR] Remove the Load node

qcolombet:
We decide not to use the load node, at least for now.
Instead, we'll stick to the following constraints:
- Save nodes are the last use of a Variable
- There is at most one save node per variable

This commit mainly reverts 67ec1cb4c030c145c9b40d5094fedc6aa5c419a2.

NFC.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -508,8 +508,6 @@ public:
                   unsigned hiddenSize, unsigned outputSize,
                   std::vector<NodeValue> &outputs);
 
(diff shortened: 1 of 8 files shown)


opti-mix:
> There is at most one save node per variable

@qcolombet : Do we have a check for this in the verifier?

qcolombet:
>> There is at most one save node per variable

> @qcolombet : Do we have a check for this in the verifier?

@opti-mix : Not yet, but it's planned.

I want to kill all the overwrite operands first to be left with only the save node having the overwrite. Otherwise, it would still be possible to sneak a write to a variable while complying to the only one save node rule.

PR #1312, opened 2018-07-23
[graph-verifier] Add stricter checks for gradient nodes

opti-mix:
Check that types of inputs/outputs and their gradients are the same.


opti-mix:
@jfix71 The only reason is to have a more custom error message ;-)

PR #1313, opened 2018-07-23
[cpu] Remove AllocationsInfo::clear. NFC

bertmaher:
We don't currently reuse the AllocationsInfo object anywhere, and it seems better to discourage that anyways.


qcolombet:
Interesting that we would introduce a method without any use :).

Thanks for the clean up.

rdzhabarov:
might be good to assert on those default values if we remove cleaning part (to guarantee no reuse of the same allocation info).

PR #1314, opened 2018-07-23
[Function] Check that each variable is written at most once

qcolombet:
Update the verifier code to check that at most one write happens
for each variable.

NFC.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1907,9 +1907,31 @@ void Function::verify() const {
     }
   }
 
+  std::unordered_map<const Variable *, const Node *> variablesWrittenTo;
   for (const auto &N : nodes_) {
     assert(N.getParent() == this &&
            "Node is not linked to the function it belongs");
     N.verify();
+    // Make sure all the variables are at most written once.
+    for (size_t idx = 0, e = N.getNumInputs(); idx < e; ++idx) {
+      if (!N.isOverwrittenNthInput(idx)) {
+        continue;
+      }
+      auto &input = N.getNthInput(idx);
+      if (!isa<Variable>(input)) {
+        continue;
+      }
+      const auto *var = cast<Variable>(input);
+      auto varToFirstDef = variablesWrittenTo.find(var);
+      if (varToFirstDef != variablesWrittenTo.end()) {
+        llvm::errs() << "Variable " << var->getDebugDesc() << '\n';
+        llvm::errs() << "has more than one write:\n";
+        llvm::errs() << N.getDebugDesc() << '\n';
+        llvm::errs() << varToFirstDef->second->getDebugDesc() << '\n';
+      }
+      assert(varToFirstDef == variablesWrittenTo.end() &&
+             "Variable has more than one write");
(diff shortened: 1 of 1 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1907,9 +1907,31 @@ void Function::verify() const {
...
     assert(N.getParent() == this &&
            "Node is not linked to the function it belongs");
     N.verify();
+    // Make sure all the variables are at most written once.
+    for (size_t idx = 0, e = N.getNumInputs(); idx < e; ++idx) {
+      auto &input = N.getNthInput(idx);
+      if (!isa<Variable>(input)) {
+        continue;
+      }
+      if (!N.isOverwrittenNthInput(idx)) {
```
opti-mix: This check should be the first. If it fails, we don't care of the input is a variable or not.

lib/Graph/Graph.cpp
```diff
@@ -1907,9 +1907,31 @@ void Function::verify() const {
...
+        continue;
+      }
+      const auto *var = cast<Variable>(input);
+      auto varToFirstDef = variablesWrittenTo.find(var);
+      if (varToFirstDef != variablesWrittenTo.end()) {
+        llvm::errs() << "Variable " << var->getDebugDesc() << 'n';
```
jfix71: nit: missed backslash on `\n`

lib/Graph/Graph.cpp
qcolombet: That's true the other way around too.

lib/Graph/Graph.cpp
qcolombet: Good catch!

lib/Graph/Graph.cpp
```diff
@@ -1907,9 +1907,31 @@ void Function::verify() const {
     }
   }
 
+  std::unordered_map<const Variable *, const Node *> variablesWrittenTo;
```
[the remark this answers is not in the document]
qcolombet: I thought about a set at first but decided to used a map to be able to print where another write happens.
Arguably, this is partially useful because we would only print two writes. I could either refactor the code to print all the writes or none and use a set.

What do you think?

lib/Graph/Graph.cpp
opti-mix: My point was that `N.getNthInput(idx)` and `isa<Variable>(input)` are probably more expensive than `isOverwrittenNthInput`. 
But on the other hand, it is a verifier, so who cares ;-)

lib/Graph/Graph.cpp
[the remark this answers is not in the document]
rdzhabarov: I do not see where you'd use the value from the map ;)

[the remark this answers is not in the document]
rdzhabarov: ah, that's fine. I missed the error output from the varToFirstDef.

qcolombet:
> we are only requiring a Variable be written to once per Function, correct?

@jfix71 that's correct

qcolombet:
I'll update the commit message.

PR #1316, opened 2018-07-24
[quantization][opencl]: Support quantization for RescaleQuantizedInst

ZchiPitt:
Support quantization for RescaleQuantizedInst
-- Modified the "RescaleNode" test so that createRescaleQuantized() didn't get optimized out. 
-- RescaleNode test stands as the test for this PR

### DIFF
--- a/lib/Backends/OpenCL/kernels.cl
+++ b/lib/Backends/OpenCL/kernels.cl
@@ -114,6 +114,26 @@ __kernel void quantize_i8W(__global void *mem, cl_uint32_t dest,
   quantize_i8K(&mem[dest], &mem[src], scale, offset);
 }
 
+__kernel void rescalequantized_i8K(__global cl_int8_t *dest,
+                                   __global cl_int8_t *src, cl_int32_t destOffset,
+                                   cl_int32_t srcOffset, cl_int32_t rescalePre,
+                                   cl_int32_t rescalePost,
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/kernels.cl
```diff
@@ -114,6 +114,26 @@ __kernel void quantize_i8W(__global void *mem, cl_uint32_t dest,
...
+                             rescaleScale, destOffset);
+  dest[i] = clip(s);
+}
+
+__kernel void rescalequantized_i8W(
+    __global void *mem, cl_uint32_t dest, cl_uint32_t src,
+    cl_int32_t destOffset,
+    cl_int32_t srcOffset, QuantizationTransform32To8 rescaleParams) {
+  rescalequantized_i8K(&mem[dest], &mem[src], destOffset, srcOffset,
+                        rescaleParams.pre_, rescaleParams.post_,
```
opti-mix: Nit: The `post_` naming convention with an underscore follows our C++ coding style, but does not follow the `kernels.cl` coding style, as we do not use underscores for the fields there. We probably want to have a consistent style inside  a single file. But this is not something we need to fix in this PR.

stoklund: Actually, public struct fields are not supposed to have a trailing underscore.

PR #1317, opened 2018-07-24
[OpenCL] Fix InsertTensor kernel to expect cl_uint32_t at input.

artemrakhov:
My Linux's OpenCL wasn't happy about some args being sent out as `cl_uint`, but expected to be received as `size_t` on the other side. I also changed the whole kernel's code to use `cl_uint64_t`. I'm not sure what is current plan. Maybe other kernels need to be changed in the same way.


opti-mix:
@artemrakhov In principle, we only need to use fixed-sized integers in the function signatures, because size_t is not guaranteed to have the same size on the host and a GPU.  You found one more place, where I forgot to do it. Thanks!

Inside the kernels, there is no real need to enforce a fixed size representation.

Even if we would like to fix the use of size_t or ssize_t inside kernels, it should be done in a separate PR and for all kernels. So, I'd suggest that you only fix the signature of the kernel in this PR.

artemrakhov:
@opti-mix , I did what you suggest and reverted changes inside kernel's code. But this looks wrong to me: we create `size_t` variable and assign it `cl_uint64_t`, which could be immediate overflow. Of course it is very unlikely. But such assignments happen all over OpenCL code.

PR #1322, opened 2018-07-24
[CPU] Fix DKKC8 filter out-of-bounds issue.

beicy:
Before generating libjit_convDKKC8_f function, we can guarantee that (output_channel / group) is divisible by 64. However, in libjit_convDKKC8_f function, we handle "8 * numDepthRegs * depthStrips" channels per time (i.e. d += 8 * numDepthRegs * depthStrips). The  (output_channel / group) may not be divisible by 8 * numDepthRegs * depthStrips. Under such circumstance, the calculated filter index may be out-of-bounds. (An example is in Inception_v1 Caffe2 model, the outChannel is 192, numDepthRegs is 8 and depthStrips is 2.)
One solution is in node level, we apply DKKC8 only if output_channel / group is divisible by 512, since numDepthRegs could be only 2 or 8 and depthStrips could be only 1, 2, 4, 8. But I think this would affect the performance. 
The second way is add boundary check in libjit, which is applied in this PR.

### DIFF
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -1458,7 +1458,7 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     auto *group = emitConstSizeT(builder, CI->getGroup());
 
     size_t inChannels = src->dims()[3];
-    size_t outChannels = src->dims()[3];
+    size_t outChannels = dest->dims()[3];
 
     // Select a method for iterating on the image in the pixel (filter-first, or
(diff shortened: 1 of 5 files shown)


artemrakhov:
Can we have a unit test, that would expose this issue?

beicy:
Testcase is added.

beicy:
@nadavrot For resnet50 -cpu -iterations=100 : Actually, I run the 100-iterations 3 times for both. With my PR, the numbers are (0.3633, 0.3608, 0.3556), and without my PR, the numbers are: (0.3744, 0.3667, 0.3597). So I think resnet50  has not regressed. Will check others as well.

nadavrot:
LGTM!

PR #1323, opened 2018-07-25
[WIP] Implement the Caffe2 batched operator.

nadavrot:
In this PR we update the gather node to support Batches. We add the
'batches' operand that marks some indices in the beginning of the tensor
as 'batch' indices. The gather operation is executed on each sample in
the batch independently. The batches are later concatenated. This allows
us to implement the Gather and BatchedGather operation in a single node.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -269,6 +269,19 @@ struct Type final {
     return s;
   }
 
+  /// \returns the number of elements in a slice in the tensor. Calculate the
+  /// size of the slice starting at \p startDim. For example, the tensor with
+  /// the shape [10, 10, 3] and startDim 1 would have the size 30, because this
+  /// is the size of the slice [10, 3] that starts at index 1.
+  size_t getSlicesize(unsigned startDim) const {
+    assert(startDim <= numSizes_ && "Invalid start dim");
+    size_t s = 1;
+    for (unsigned i = startDim; i < numSizes_; i++) {
+      s *= size_t(sizes_[i]);
+    }
+    return s;
+  }
+
   /// \returns true if the templated parameter \p ElemTy matches this type.
   template <class ElemTy> bool isType() const {
     return isType<ElemTy>(elementType_);
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -729,18 +729,38 @@ void InterpreterFunction::fwdExtractTensorInst(
 
 void InterpreterFunction::fwdGatherInst(const glow::GatherInst *I) {
   Tensor *dataT = getTensor(I->getData());
+  auto &dataTy = dataT->getType();
+  auto iDims = dataT->dims();
   Tensor *indicesT = getTensor(I->getIndices());
   Tensor *outT = getTensor(I->getDest());
+  unsigned batchDims = I->getBatchDims();
 
   size_t out_p = 0;
-  size_t dataSliceSize =
-      dataT->size() / dataT->dims()[0] * dataT->getType().getElementSize();
-  for (size_t i = 0, end = indicesT->size(); i < end; i++) {
-    size_t slice = indicesT->getHandle<size_t>().raw(i);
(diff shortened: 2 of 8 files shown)


nadavrot:
This PR is still work in progress. Next, I'll add the OCL and CPU support. Later I will update the C2 and ONNX loaders.

include/glow/Base/Type.h
```diff
@@ -269,6 +269,19 @@ struct Type final {
     return s;
   }
 
+  /// \returns the number of elements in a slice in the tensor. Calculate the
+  /// size of the slice starting at \p startDim. For example, the tensor with
+  /// the shape [10, 10, 3] and startDim 1 would have the size 30, because this
+  /// is the size of the slice [10, 3] that starts at index 1.
+  size_t getSlicesize(unsigned startDim) const {
```
opti-mix: `s/getSlicesize/getSliceSize/`?

nadavrot:
@artemrakhov  I fixed all of the issues except for the code duplication in Tensor::sliceSize(). I'd like to clean that up in a separate PR.

artemrakhov:
Why can't you use `Tensor::sliceSize()` ?

nadavrot:
@artemrakhov Because the slice size should be a property of the Type not the Tensor. We should be able to operator on nodes that have operands that are not backed by tensors. We should be removing the backing Tensor from Variable. We should be able to operate on operands (that are activations) and are not backed by tensors. I'd like to clean this up, but prefer not to do it as part of this PR.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -729,18 +729,34 @@ void InterpreterFunction::fwdExtractTensorInst(
...
   size_t out_p = 0;
-  size_t dataSliceSize =
-      dataT->size() / dataT->dims()[0] * dataT->getType().getElementSize();
-  for (size_t i = 0, end = indicesT->size(); i < end; i++) {
-    size_t slice = indicesT->getHandle<size_t>().raw(i);
-    std::copy(&dataT->getUnsafePtr()[dataSliceSize * slice],
-              &dataT->getUnsafePtr()[dataSliceSize * (slice + 1)],
-              &outT->getUnsafePtr()[out_p]);
-    out_p += dataSliceSize;
+  unsigned elementSize = dataT->getType().getElementSize();
```
artemrakhov: nit: `unsigned elementSize = dataTy.getElementSize();`

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1933,12 +1934,26 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
-    auto *sliceSize =
-        emitConstSizeT(builder, dataType->size() / dataType->dims()[0]);
+
+    // The size of the sample in the batch.
+    size_t sampleSize = dataType->getSliceSize(batchDims);
+    // The size of the slices that we gather.
+    size_t sliceSize = dataType->getSliceSize(batchDims + 1);
+
+    // Calculate the size of each sample in the batch.
+    size_t numSamples = 1;
```
artemrakhov: `size_t numSamples =  dataType->size() / sampleSize;`

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1050,12 +1050,35 @@ void OpenCLFunction::execute() {
...
       size_t numIndices = GI->getIndices()->size();
+
+      // The size of the sample in the batch.
+      size_t sliceSize = dataType->getSliceSize(batchDims + 1);
+      // The size of the slices that we gather or pack.
+      size_t srcSampleSize = dataType->getSliceSize(batchDims);
+      size_t destSampleSize = numIndices * sliceSize;
+
+      // Calculate the size of each sample in the batch.
+      size_t numSamples = 1;
```
artemrakhov: ditto

nadavrot:
@artemrakhov Thank you for the review.

Issue #1321, opened 2018-07-24
[quantization] Add symmetric quantization schema support to Glow


rdzhabarov:
## Current state
The latest version of Glow compiler provides a certain level of quantization support.

Quantization process consists of the following steps:
1) Instrument an original floating-point graph with the special [Quantization Profile](https://github.com/pytorch/glow/blob/master/tools/ClassGen/NodeGen.cpp#L393) node

2) Run inference on the instrumented graph many times with good representative inputs (see [dump_profile flag](https://github.com/pytorch/glow/blob/ace1b2beb6a98f8ccff663963984752de99d5552/tools/loader/Loader.cpp#L64)).
During this step Glow automatically captures the distribution of all activations in the compute graph. The distribution contains min/max value seen from the given output of the graph node as well as detailed frequency per floating point range withing [min, max] interval. There are 2000 floating point ranges kept as part of the Quantization Profile node.
As the result of the profiling procedure, Glow generates [Scale and Offset](https://github.com/pytorch/glow/blob/master/lib/Quantization/Quantization.cpp#L31) parameters for all activations and dumps it in the file (Note, currently Glow uses linear quantization). The scale is a positive fp32 number while Offset is an int32. Note, Glow does not use the distribution of floating point numbers but relies only on values of min and max (this is a separate issue and out of scope here).

3) Transform computation graph based on the captured profile and specific execution backend. Note, not all nodes are quantized as not all backends support certain quantized op. See CPU backend op support [here](https://github.com/pytorch/glow/blob/master/lib/Backends/CPU/CPUBackend.cpp#L142) for example.

4) Perform computation of quantized graph. See Interpreter implementation [here](https://github.com/pytorch/glow/blob/master/lib/Backends/Interpreter/InterpreterNodes.cpp#L941).

## What needs to be enhanced
There are backends that tight to a specific quantization schema, e.g., symmetric quantization.
The difference between symmetric and assymetric quantization is that the "offset" parameter equals to 0 in symmetric linear quantization. In this case the dequantization follows the formula: fp32_number = scale * quantized_number (fp32_number = scale * (quantized_number - offset) for the asymmetric case).

## Requirement for the symmetric quantization
* Need to make sure that the accuracy loss is comparable with the asymmetric case on Resnet50. Important note, currently, activations are signed int8 and in case of symmetric quantization, outputs of Sigmoid/ReLU will effectively use only 7 bits out of 8 (which is undesirable).

The issue could be tackled in two steps:
[...]

PR #1324, opened 2018-07-25
Add quantization schema

qcolombet:
Plumb quantization schema up to the load command line so that we can choose which kind of quantization (symmetric/asymmetric) we want to use.

This is related to #1321.

### DIFF
--- a/lib/Quantization/Base/Base.cpp
+++ b/lib/Quantization/Base/Base.cpp
@@ -136,7 +136,8 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
                                     offset);
 }
 
-TensorQuantizationParams chooseQuantizationParams(float min, float max) {
+TensorQuantizationParams chooseQuantizationParams(float min, float max,
+                                                  Schema schema) {
   assert(min <= max && "min must not be bigger than max");
 
   // Given 8 bit precision.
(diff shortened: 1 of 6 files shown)


lib/Quantization/Base/Base.cpp
```diff
@@ -149,6 +150,18 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max) {
   min = std::min(min, 0.f);
   max = std::max(max, 0.f);
 
+  if (schema == quantization::Schema::Symmetric) {
```
rdzhabarov: can you add some tests to make sure that you'll get the 0 offsets

qcolombet: Good point! Forgot to do that after testing the loader.

qcolombet: @rdzhabarov, I've added a test case for the quantization schema if you want to take a look.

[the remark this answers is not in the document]
qcolombet: Done

tests/unittests/quantizationTest.cpp
```diff
@@ -407,6 +407,113 @@ TEST(Quantization, optimizeRescaleQuantize) {
...
+      chooseQuantizationParams(-3.0, 3.0, quantization::Schema::Asymmetric);
+  // Dequantization formula is scale(X - offset).
+  // So in theory, we should get
+  // 1. scale(-128 - offset) == -3.0
+  // 2. scale(127 - offset) == 3.0
+  // Given scale != 0, #1 + #2 gives scale(-128 + 127 - 2*offset) == 0.0
+  // offset == -1 / -2 == 0.5
+  // Then #2 or #1, gives scale == 3.0 / 127.5.
+  // However, when we get symmetric ranges (i.e., [-X; X]),
+  /// we actually force the zero point to map to 0.
```
rdzhabarov: nit: ///

jfix71:
Perhaps add documentation about this to the docs? Either here or in another PR.

PR #1325, opened 2018-07-25
[partition] Introduce FunctionGraph and produce it via partition()

bertmaher:
The FunctionGraph representation and the partitioning process seemed like the least controversial parts of the partitioning PR, so I've extracted them, cleaned up and simplified the logic, and put them up for review.

`glow::partition()` is now a standalone function that produces a `FunctionGraph`.  How this graph is executed is an orthogonal question; currently the test simply uses serial execution of the subgraphs, which is the simplest approach.

### DIFF
--- a/include/glow/Optimizer/Partition.h
+++ b/include/glow/Optimizer/Partition.h
@@ -25,7 +25,7 @@ namespace glow {
 /// Maps a set of functions to the set of functions it depends on.  The
 /// execution of a function cannot begin until all its dependencies have
 /// executed.
-class FunctionGraph {
+class FunctionDAG {
   /// Dependency map type.
   using Map = llvm::DenseMap<Function *, FunctionList>;
 
@@ -40,18 +40,18 @@ class FunctionGraph {
 
 public:
   /// Ctor.
-  FunctionGraph(const FunctionList &functions);
+  FunctionDAG(const FunctionList &functions);
 
   /// Dtor.
-  ~FunctionGraph() = default;
+  ~FunctionDAG() = default;
 
   /// Move ctor.
-  FunctionGraph(FunctionGraph &&G) = default;
-  FunctionGraph &operator=(FunctionGraph &&G) = default;
+  FunctionDAG(FunctionDAG &&G) = default;
+  FunctionDAG &operator=(FunctionDAG &&G) = default;
 
-  /// FunctionGraph is non-copyable to ensure efficiency.
-  FunctionGraph(const FunctionGraph &G) = delete;
-  FunctionGraph &operator=(const FunctionGraph &G) = delete;
+  /// FunctionDAG is non-copyable to ensure efficiency.
+  FunctionDAG(const FunctionDAG &G) = delete;
+  FunctionDAG &operator=(const FunctionDAG &G) = delete;
 
   /// Get list of functions in this graph.
   const FunctionList &getFunctions() const { return functions_; }
@@ -70,8 +70,8 @@ public:
   bool verify() const;
 };
 
-/// Split an input Function into a FunctionGraph.
-FunctionGraph partition(Function *F);
+/// Split an input Function into a FunctionDAG.
+FunctionDAG partition(Function *F);
 
 } // namespace glow
 
--- a/include/glow/Optimizer/Partition.h
+++ b/include/glow/Optimizer/Partition.h
@@ -36,7 +36,7 @@ class FunctionGraph {
   FunctionList functions_;
 
   /// Input dependencies of each function.
-  Map inputs_;
+  Map dependencies_;
 
 public:
   /// Ctor.
@@ -56,15 +56,18 @@ public:
   /// Get list of functions in this graph.
   const FunctionList &getFunctions() const { return functions_; }
 
-  /// Get input dependences for a function.
-  const MappedType &getInputs(Function *F) const {
-    auto it = inputs_.find(F);
-    assert(it != inputs_.end() && "No inputs found for function in graph");
+  /// Get dependencies for a function.
+  const MappedType &getDependencies(Function *F) const {
+    auto it = dependencies_.find(F);
+    assert(it != dependencies_.end() && "No dependencies found for function in graph");
     return it->second;
   }
 
   /// Record that \p F depends on \p inputF.
-  void add(Function *F, Function *inputF) { inputs_[F].push_back(inputF); }
+  void add(Function *F, Function *inputF) { dependencies_[F].push_back(inputF); }
+
+  /// Verify that function graph is well-formed (acyclic, topologically sorted).
+  bool verify() const;
 };
 
 /// Split an input Function into a FunctionGraph.
--- a/include/glow/Optimizer/Partition.h
+++ b/include/glow/Optimizer/Partition.h
@@ -53,7 +53,7 @@ public:
   FunctionGraph &operator=(const FunctionGraph &G) = delete;
 
   /// Get list of functions in this graph.
-  FunctionList &getFunctions() { return functions_; }
+  const FunctionList &getFunctions() const { return functions_; }
 
   /// Get input dependences for a function.
   const MappedType &getInputs(Function *F) const {
--- a/include/glow/Optimizer/Partition.h
+++ b/include/glow/Optimizer/Partition.h
@@ -22,8 +22,9 @@
 
 namespace glow {
 
-/// Data structure representing a set of interdependent Functions and the
-/// dependences between them.
+/// Maps a set of functions to the set of functions it depends on.  The
+/// execution of a function cannot begin until all its dependencies have
+/// executed.
 class FunctionGraph {
   /// Dependency map type.
   using Map = llvm::DenseMap<Function *, FunctionList>;
--- a/include/glow/Optimizer/Partition.h
+++ b/include/glow/Optimizer/Partition.h
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


tests/unittests/partitionTest.cpp
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  // Infer using the un-partitioned graph.
+  Tensor in(ElemKind::FloatTy, {1, 32});
+  ExecutionEngine EE;
+  EE.compile(CompilationMode::Infer, F_->clone("clone"));
+  EE.run({input}, {&in});
+  Tensor ref = res.clone();
+
+  // Infer using the partitioned graph.
```
opti-mix: Add comment that the test simply uses serial execution of the subgraphs, which is the simplest approach.

opti-mix: Also, may be create a helper function for serial execution of FunctionGraphs. Or even a helper function for executing a FunctionGraph that takes a mode of execution (an enumeration of serial, parallel, pipelined, etc) as a parameter. For the moment, it would only support the serial execution.

tests/unittests/partitionTest.cpp
```diff
@@ -0,0 +1,143 @@
...
+  auto *save = F_->createSave("ret", mul);
+  auto &res = save->getVariable()->getPayload();
+
+  auto G = glow::partition(F_);
+  ASSERT_EQ(G.getFunctions().size(), 4);
+
+  auto it = G.getFunctions().begin();
+  {
+    auto *F = *it++;
+    EXPECT_EQ(F->getNodes().size(), 3);
```
opti-mix: In your tests, you check the number of nodes in the partition and the number of inputs. Do we need to check also for some structural properties? E.g. that certain variables or Save nodes are present in the partition?

include/glow/Optimizer/Partition.h
```diff
@@ -0,0 +1,74 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <llvm/ADT/DenseMap.h>
+
+namespace glow {
+
+/// Data structure representing a set of interdependent Functions and the
+/// dependences between them.
+class FunctionGraph {
+  /// Dependency map type.
```
opti-mix: Could you expand the comment here to something like "Maps a function to the set of functions it depends on. The execution of a function cannot start until all of its dependencies are executed."

include/glow/Optimizer/Partition.h
```diff
@@ -0,0 +1,74 @@
...
+  /// Move ctor.
+  FunctionGraph(FunctionGraph &&G) = default;
+  FunctionGraph &operator=(FunctionGraph &&G) = default;
+
+  /// FunctionGraph is non-copyable to ensure efficiency.
+  FunctionGraph(const FunctionGraph &G) = delete;
+  FunctionGraph &operator=(const FunctionGraph &G) = delete;
+
+  /// Get list of functions in this graph.
+  FunctionList &getFunctions() { return functions_; }
```
opti-mix: Should it be a const method?

lib/Optimizer/Partition.cpp
```diff
@@ -0,0 +1,194 @@
...
+  void create(Node *N, Function *F) {
+    functions_.emplace_back(F);
+    nodeToFunction_[N] = F;
+  }
+
+  /// Add a new Node->Function mapping.
+  void add(Node *N, Function *F) { nodeToFunction_[N] = F; }
+
+  /// Get list of functions contained in this map.
+  FunctionList &getFunctions() { return functions_; }
```
opti-mix: Should it be a const method?

lib/Optimizer/Partition.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+#include "glow/Optimizer/Partition.h"
+
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Utils.h"
+
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/Support/Casting.h"
+
+using namespace glow;
```
opti-mix: Add `using llvm::isa;` and replace `llvm::isa` by `isa` in the code. This is what we usually do in the Glow codebase.

tests/unittests/partitionTest.cpp
bertmaher: I can add something simple to the test harness; I kind of want to leave general execution mode support for another PR.

tests/unittests/partitionTest.cpp
bertmaher: Might be reasonable.  In development I found the simple structural checks plus the execution check caught more or less everything and are pretty simple to look at :)

bertmaher:
Addressed comments (I'll squash the fixup commits before merging, of course.)

tests/unittests/partitionTest.cpp
```diff
@@ -0,0 +1,163 @@
...
+public:
+  PartitionTest() : F_(mod_.createFunction("main")) {}
+
+protected:
+  Module mod_;
+  Function *F_;
+};
+
+/// Execute a graph of functions serially, which is the simplest approach.
+void executeSerial(const FunctionGraph &G, llvm::ArrayRef<Variable *> vars,
```
opti-mix: static? Though it probably doesn't matter for a test.

include/glow/Optimizer/Partition.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /// Get input dependences for a function.
+  const MappedType &getInputs(Function *F) const {
+    auto it = inputs_.find(F);
+    assert(it != inputs_.end() && "No inputs found for function in graph");
+    return it->second;
+  }
+
+  /// Record that \p F depends on \p inputF.
+  void add(Function *F, Function *inputF) { inputs_[F].push_back(inputF); }
```
opti-mix: I wonder if the names should be more descriptive, e.g. `addInputDependency` instead of `add` and `getInputDependencies` instead of `getInputs`?

lib/Optimizer/Partition.cpp
```diff
@@ -0,0 +1,195 @@
...
+FunctionGraph::FunctionGraph(const FunctionList &functions)
+    : functions_(functions) {
+  for (auto *F : functions_) {
+    inputs_[F] = FunctionList();
+  }
+}
+
+FunctionGraph glow::partition(Function *F) {
+  NodeFunctionMap partitionMap = selectPartitions(F);
+  return doPartitioning(F, partitionMap);
```
jfix71: Maybe there should be a verifier here, e.g. to check that there are no cycles between Functions in the FunctionGraph returned from `doPartitioning()`?

include/glow/Optimizer/Partition.h
```diff
@@ -0,0 +1,75 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <llvm/ADT/DenseMap.h>
+
+namespace glow {
+
+/// Maps a set of functions to the set of functions it depends on.  The
+/// execution of a function cannot begin until all its dependencies have
+/// executed.
+class FunctionGraph {
```
nadavrot: Maybe a better name would be FunctionDAG or FunctionDependencyDAG?

nadavrot: How are you planning to AOT this class?

include/glow/Optimizer/Partition.h
```diff
@@ -0,0 +1,75 @@
...
+
+  /// FunctionGraph is non-copyable to ensure efficiency.
+  FunctionGraph(const FunctionGraph &G) = delete;
+  FunctionGraph &operator=(const FunctionGraph &G) = delete;
+
+  /// Get list of functions in this graph.
+  const FunctionList &getFunctions() const { return functions_; }
+
+  /// Get input dependences for a function.
+  const MappedType &getInputs(Function *F) const {
```
nadavrot: Would getDependencies be a better name?

include/glow/Optimizer/Partition.h
nadavrot: Generally I agree, except that I think that "inputs" is not a good name. Maybe deps?

include/glow/Optimizer/Partition.h
nadavrot: How are you going to visualize this data structure?

tests/unittests/partitionTest.cpp
```diff
@@ -0,0 +1,164 @@
...
+protected:
+  Module mod_;
+  Function *F_;
+};
+
+/// Execute a graph of functions serially, which is the simplest approach.
+static void executeSerial(const FunctionGraph &G,
+                          llvm::ArrayRef<Variable *> vars,
+                          llvm::ArrayRef<Tensor *> inputs) {
+  for (auto *F : G.getFunctions()) {
```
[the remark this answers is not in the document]
bertmaher: There's a currently-unstated invariant that the function-list be topologically sorted, which is a nice property for the executor to be able to assume.  I was debating whether to explicitly sort the list at construction time or assert that it is sorted.  Currently it's true by construction.

include/glow/Optimizer/Partition.h
bertmaher: There are lots of options for AOT.  The trivial thing would be to handle it externally (like the serial JIT executor): dump N bundles, then simply invoke their exported functions serially.

To actually take advantage of the partitioning, I'd serialize the `FunctionGraph` into `BundleConfig` so we can reconstitute it as part of the network state.  From there I see at least two options:
1. Leave it up to the "glue" code to decide how to execute the graph.  To make this easier, we could provide a library of executors that's independent of the Glow core.
2. Compile-in the execution strategy by making the executor part of libjit.  That would take a bit more codegen to emit the glue, but makes the external interface easier.
3. (Maybe a bit of a research project) Drop the notion of an executor entirely and compile-in the dependences to the individual functions.

bertmaher: Re: visualization, I've found it useful to dump the individual partitions to dot.  It would be extremely to cool to dump the whole thing to a single dot file that uses subgraphs to delineate the partitions, but the graphviz wizardry for that is pretty high level (if even possibly for complicate graphs).  Maybe I'll try it :-p.

nadavrot: I wonder if turning FunctionGraph into a function with Call nodes would solve the AOT and the visualization problem. AOT would be done by emitting 'call' instructions, and you wull get visualization for free.

lib/Optimizer/Partition.cpp
bertmaher: Great idea!  Done :-)

include/glow/Optimizer/Partition.h
bertmaher: Hm, interesting idea.  I kind of think the hardest part exists either way... for AOT you have to figure out how to compile a bunch of functions and then refer to them from some caller.  I don't see a huge gap between a function-dag (containing functions) and a function (containing call nodes).

Issue #1309, opened 2018-07-23
Should lower/optimize behave differently based on the compilation mode?


qcolombet:
As of today our API takes the CompilationMode for both the lowering and optimization phase (glow::lower and glow::optimize).
Except for one thing in the lowering, which I will come back later, this argument does change the behavior of these functions and thus, I was wondering if we should pass it at all.
My way of thinking is if we don't use it, let's not pass it and if/when we need it we can add it.

Now going back to the only user of this argument in lowering.
The lowering of batch normalization is the only user of that argument, but I believe we could avoid that use and that would be a better design. Basically, we lower BN differently between inference and training, whereas I think the right thing would be to have a different representation when we differentiate BN. In essence, what I am suggesting for this use case is to introduce a new node for mean and variance normalization (i.e., for what we special case with the training mode during lowering.) In other words, differentiation for BN would produce BNGrad(BN(meanVarNorm)) and that would completely eliminate the need for a special lowering.

All in all, as of today, I don't see a reason why we pass this mode.

What do people think?

qcolombet:
Hi @nadavrot,

In essence, yes that's what I am suggesting with one difference.
Instead of creating a BN-train node that would replace the BN-inference during differentiation, I was thinking of replacing BN with two nodes BN + meanVarNorm during differentiation.

The effects are the same, the difference is that we wouldn't have to change the optimizations.
I.e.,
Input:
BN-inference

Differentation:
- Today
BN-inference
BN-gradient

- Your proposal
BN-train
BN-gradient

- My suggestion
meanVarNorm
BN-inference
BN-gradient

Essentially, BN-train is equal to (BN-inference(meanVarNorm)).

nadavrot:
@qcolombet Ah, got it. We can split BN-train into two sequential parts. I think that this proposal makes sense.

jfix71:
@qcolombet I'll take a look today!

PR #1326, opened 2018-07-25
Add MeanVarNormalization for use by BatchNormalization for training

jfix71:
I removed the CompilationMode from `lower()` in a separate commit here as well. This resolves https://github.com/pytorch/glow/issues/1309.

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -97,7 +97,6 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
     CONVERT_TO_GRAD_NODE(PoolMaxNode)
     CONVERT_TO_GRAD_NODE(PoolAvgNode)
     CONVERT_TO_GRAD_NODE(FullyConnectedNode)
-    CONVERT_TO_GRAD_NODE(BatchNormalizationNode)
     CONVERT_TO_GRAD_NODE(LocalResponseNormalizationNode)
     CONVERT_TO_GRAD_NODE(SoftMaxNode)
     CONVERT_TO_GRAD_NODE(CrossEntropyLossNode)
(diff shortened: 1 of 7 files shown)


tools/ClassGen/NodeGen.cpp
```diff
@@ -120,6 +120,17 @@ int main(int argc, char **argv) {
                     "Momentum. Similar to Caffe2 SpatialBN, and ONNX "
                     "BatchNormalization operator.");
 
+  BB.newNode("MeanVarNormalization")
+      .addInput("Input")
+      .addInput("Mean")
+      .addInput("Var")
+      .addMember(MemberType::SizeT, "ChannelIdx")
+      .addMember(MemberType::Float, "Momentum")
+      .addResultFromCtorArg("NewMean")
```
artemrakhov: I think you can write here:
```
       .addResult("Mean.getType()")
       .addResult("Var.getType()")
```
Then you won't have to pass types to the constructor.

jfix71: I don't think this is currently supported for Nodes with multiple outputs. They're both called "Result" and so `getResult()` is defined twice. Could fix this in a future PR.

qcolombet: You’ll need to give them a name:
.addResult(“type”, “name”)

jfix71: I didn't realize we supported that! Done.

jfix71:
I realized that the save nodes should have been added during differentiation and not lowering, otherwise the variables would incorrectly be seen as not written into post-differentiation/pre-lowering.

lib/Graph/Grad.cpp
```diff
@@ -190,6 +189,43 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
+      // variables. These will be used during the next iteration of training.
+      G->createSave("saveMean", MVN->getNewMean(),
+                    llvm::cast<Variable>(mean.getNode()));
+      G->createSave("saveVar", MVN->getNewVar(),
+                    llvm::cast<Variable>(var.getNode()));
+
+      // Replace the BN's mean and variance with the new mean and variance
+      // calculated from MVN.
+      for (size_t i = 0; i < BN->getNumInputs(); i++) {
+        if (BN->getNthInput(i) == mean) {
```
qcolombet: Could we directly modify the operands we believe are mean and var while asserting on them being equal instead of using a loop.
I.e.,
assert(BN->getNthInput(idxMean) ==  mean);
BN->setNthInput(idxMean, mean);

jfix71: Yeah, I could make that change if desired. Seems a little fragile as its based on the order of operands in NodeGen but the assert should catch any issues.

jfix71: Done.

nadavrot:
@jfix71 I notices that we don't have any neural networks that use batchednormalization, except for the grad checks. I think it would be a good idea to add at least one compile test that checks that things work as expected across the compilation lifetime (creation->differentiation->execution)

PR #1327, opened 2018-07-25
[quantization][opencl]: Support quantization for the slow ConvolutionInst

ZchiPitt:
Support quantization for the slow implementation of ConvolutionInst.
-- Changed the Transform.cpp so that we execute the slow convolution inst when input type is quantized.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -878,6 +878,16 @@ void OpenCLFunction::execute() {
       setKernelArg(kernel, numArgs + 5, idim);
       setKernelArg(kernel, numArgs + 6,
                    ShapeNHWC(CC->getFilter()->getType()->dims()));
+      if (isQuantized) {
+        setKernelArg(kernel, numArgs + 7, CC->getDest()->getType()->getOffset());
+        setKernelArg(kernel, numArgs + 8, CC->getDest()->getType()->getScale());
+        setKernelArg(kernel, numArgs + 9, CC->getSrc()->getType()->getOffset());
+        setKernelArg(kernel, numArgs + 10, CC->getSrc()->getType()->getScale());
+        setKernelArg(kernel, numArgs + 11, CC->getFilter()->getType()->getOffset());
+        setKernelArg(kernel, numArgs + 12, CC->getFilter()->getType()->getScale());
+        setKernelArg(kernel, numArgs + 13, CC->getBias()->getType()->getOffset());
+        setKernelArg(kernel, numArgs + 14, CC->getBias()->getType()->getScale());
+      }
 
       // Use a 3D grid where the first dimension is the depth and the second
       // dimension is the slice index in the batch.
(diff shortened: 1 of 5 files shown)


lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -36,9 +36,11 @@ bool OCLBackend::transformPostLowering(Function *F,
   bool changed = false;
   for (auto &node : F->getNodes()) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
-      auto *NR = convertConvToNCHWConv<OCLConvolutionNode>(CN, F);
-      NodeValue(&node, 0).replaceAllUsesOfWith(NR);
-      changed = true;
+      if (!CN->getInput()->getType(0)->isQuantizedType()) {
```
opti-mix: Better use an early-exit here, i.e.
```c++
if (CN->getInput()->getType(0)->isQuantizedType()) {
  continue;
}
```

lib/Backends/OpenCL/kernels.cl
```diff
@@ -735,6 +735,69 @@ __kernel void convolutionW(__global void *mem, cl_uint32_t dest,
...
+          sum += (filter[getNHWC(filterDim, d, fx, fy, fd)] - filterOffset) *
+                 (src[getNHWC(idim, n, (size_t)ox, (size_t)oy, fd)] - srcOffset);
+        }
+      }
+    }
+
+    sum += round((float)(bias[d] - biasOffset) * (biasScale / matMulScale));
+    dest[getNHWC(odim, n, ax, ay, d)] = clip(round(float(sum) * (matMulScale 
+                                        / destScale) + destOffset));
+  } // N
```
rdzhabarov: this seems to be inconsistent :) i'd remove that

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -878,6 +878,16 @@ void OpenCLFunction::execute() {
       setKernelArg(kernel, numArgs + 5, idim);
       setKernelArg(kernel, numArgs + 6,
                    ShapeNHWC(CC->getFilter()->getType()->dims()));
+      if (isQuantized) {
```
rdzhabarov: you need to enable isOpSupported for the convolution node.

rdzhabarov: this is somewhat unrelated... Since you added Max support to openCL you should enable ReLU as well as it's lowered to the Max node.

ZchiPitt: cool. Will do after OCLConvolution

lib/Backends/OpenCL/Transforms.cpp
ZchiPitt: changed

lib/Backends/OpenCL/OpenCL.cpp
rdzhabarov: sounds good, that will be one line change :)

PR #1330, opened 2018-07-26
Remove TrainingConfig from the execution engine.

nadavrot:
TrainingConfig should not be a part of the execution engine and this commit just moves it out. Once this commit lands I will continue to clean things in this area.

### DIFF
--- a/examples/mnist.cpp
+++ b/examples/mnist.cpp
@@ -98,11 +98,14 @@ void testMNIST() {
   ExecutionEngine EE(executionBackend);
   llvm::Timer timer("Training", "Training");
 
+  /// The training configuration.
+  TrainingConfig TC;
+
   // Construct the network:
-  EE.getConfig().learningRate = 0.001;
-  EE.getConfig().momentum = 0.9;
-  EE.getConfig().L2Decay = 0.001;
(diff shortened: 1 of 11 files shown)


examples/mnist.cpp
```diff
@@ -98,11 +98,14 @@ void testMNIST() {
   ExecutionEngine EE(executionBackend);
   llvm::Timer timer("Training", "Training");
 
+  /// The training configuration.
```
rdzhabarov: should not be a Doxygen comment here.

tests/unittests/HyphenTest.cpp
```diff
@@ -279,12 +279,13 @@ struct HyphenNetwork {
   // Run `inputs` through the inference function and check the results against
   // `hyphens`. Return the number of errors.
   unsigned inferenceErrors(ExecutionEngine &EE, llvm::StringRef name,
-                           Tensor &inputs, const vector<bool> &hyphens) {
+                           Tensor &inputs, const vector<bool> &hyphens,
+                           TrainingConfig &TC) {
```
rdzhabarov: this seems to be a const &TC.

bertmaher:
Very nice, I'm a fan :)

PR #1331, opened 2018-07-26
[Caffe2/ONNX] Allow max proto size

jfix71:
Allow for the maximum possible proto size.


include/glow/Importer/ProtobufLoader.h
```diff
@@ -30,6 +30,8 @@
 #include <unordered_map>
 #include <vector>
```
nadavrot: Comment:  This is the maximum size that ...

PR #1337, opened 2018-07-26
[onnxifi] Add onnxifi backend skeleton.

rdzhabarov:
Currently just implement dummy ONNXIFI interface with some basic error handling of params.
Need some basic coverage for the interface, will do that on separate PR.

PR towards ONNXIFI support in glow: https://github.com/pytorch/glow/issues/1190.


opti-mix:
@rdzhabarov Is the content of `onnxifiGlow.cpp` completely copy/pasted from an existing template or is there something Glow specific?

rdzhabarov:
The onnxifi interface itself does not have the implementation.

But there is nothing Glow specific in the implementation yet.

nadavrot:
Nice!

PR #1339, opened 2018-07-26
[trivial] Use GLOW_UNREACHABLE where appropriate

bertmaher:
I saw a "control reaches end of non-void function" error here when compiling with `-Wreturn-type`.  We could also use `GLOW_ASSERT(false && ...)`, but either way we need to do something that actually terminates the function here.

### DIFF
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -485,7 +485,7 @@ llvm::Function *LLVMIRGen::getFunction(const std::string &name,
   case ElemKind::IndexTy:
     return get("libjit_" + name + "_u");
   default:
-    GLOW_ASSERT("Unsupported element type");
+    GLOW_UNREACHABLE("Unsupported element type");
   }
 }
 


lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -485,7 +485,7 @@ llvm::Function *LLVMIRGen::getFunction(const std::string &name,
   case ElemKind::IndexTy:
     return get("libjit_" + name + "_u");
   default:
-    GLOW_ASSERT("Unsupported element type");
+    GLOW_UNREACHABLE("Unsupported element type");
```
rdzhabarov: do we have any types not covered in the switch? (for enum classes compiler will ensure that if something is missing  from "case" there will be a compilation error, so we should be good to remove "default")

bertmaher: We cover all the types in the switch, so I guess we could remove it.  But if we add new types (like half-precision or bfloat16) then we'd either have to add them here or put the default back.  So I think I'll err on the side of leaving this UNREACHABLE here :-)

PR #1342, opened 2018-07-26
[Operators] Support Flatten.

artemrakhov:


include/glow/Graph/Graph.h
```diff
@@ -314,6 +314,9 @@ class Function final : public Named {
   ReshapeNode *createExpandDims(llvm::StringRef name, NodeValue input,
                                 llvm::ArrayRef<size_t> axes);
 
+  ReshapeNode *createFlatten(llvm::StringRef name, NodeValue input,
```
nadavrot: Please add a doxygen line that describes the semantics of the operator and the parameter.

artemrakhov:
Added doxygen, fixed unit tests.

PR #1343, opened 2018-07-26
[OpenCL] Fix compilation error in a kernel.

artemrakhov:
doing `float(5)` doesn't work in `C99`. We should always use `(float)5`


opti-mix:
@ZchiPitt Please keep this in mind for the future PRs.

PR #1349, opened 2018-07-27
**WIP**[quantization][opencl]: Support quantization for OCLConvolution

ZchiPitt:
Support quantization for OCLConvolution.
While there are still some things about this PR:
-- All tests passed (including conv tests)
-- However resnet50 cannot produce correct results right now, not sure if it's the precision issues.
-- I have to disable some sinkNode optimization for the arithmetic ops that follows the OCLConvolution node. Due to the type inconsistency caused by the transformation from ConvolutionNode to OCLConvolutionNode (different scale and offset)

Please comment if you have any idea to solve the problems above.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -393,6 +393,7 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   auto odim = ShapeNCHW(CC->getDest()->getType()->dims());
   auto idim = ShapeNCHW(CC->getSrc()->getType()->dims());
   auto fdim = ShapeNCHW(CC->getFilter()->getType()->dims());
+  bool isQuantized = output->getType()->isQuantizedType();
   PaddingTLBR pads(CC->getPads());
   // Create options for compiling the program.
   // Don't use names M, N, K as they are defined in precompiled headers.
@@ -466,16 +467,32 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   // VWN and VWM should be tunable.
   addStringOption(options, "VWM", "4");
   addStringOption(options, "VWN", "4");
+  if (isQuantized) {
+    addStringOption(options, "QUANTIZED", "");
+  }
 
   // Generate a tailor-made convolution kernel using the provided options based
   // on the parameters of the current convolution.
   auto prog = createProgram(FWD_CONV_CODE, options, commands_);
-  auto kernel = createKernel("conv_forward_mem", prog);
+  auto kernelName = isQuantized ? "conv_forward_mem_i8" : "conv_forward_mem";
+  auto kernel = createKernel(kernelName, prog);
   setKernelArg(kernel, 0, deviceBuffer_);
   setKernelArg<cl_uint>(kernel, 1, tensors_[input]);
   setKernelArg<cl_uint>(kernel, 2, tensors_[weights]);
   setKernelArg<cl_uint>(kernel, 3, tensors_[bias]);
   setKernelArg<cl_uint>(kernel, 4, tensors_[output]);
+
+    // Extra options for quantized kernel
+  if(isQuantized) {
+    setKernelArg(kernel, 5, CC->getFilter()->getType()->getOffset());
+    setKernelArg(kernel, 6, CC->getFilter()->getType()->getScale());
+    setKernelArg(kernel, 7, CC->getSrc()->getType()->getOffset());
+    setKernelArg(kernel, 8, CC->getSrc()->getType()->getScale());
+    setKernelArg(kernel, 9, CC->getDest()->getType()->getOffset());
+    setKernelArg(kernel, 10, CC->getDest()->getType()->getScale());
+    setKernelArg(kernel, 11, CC->getBias()->getType()->getOffset());
+    setKernelArg(kernel, 12, CC->getBias()->getType()->getScale());
+  }  
 
   // Compute proper parameters for global work and workgroups.
   int group = 1;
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -466,17 +467,33 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   // VWN and VWM should be tunable.
   addStringOption(options, "VWM", "4");
   addStringOption(options, "VWN", "4");
+  if (isQuantized) {
+    addStringOption(options, "QUANTIZED", "");
```
opti-mix: `QUANTIZED` was required when you had a version that had a single kernel with a bunch of "ifdefs" to handle quantized and non-quantized convolutions. Once you introduced two separate kernels, there is no need for all this macro-magic. Each kernel can have its own set of macro-definitions inside the kernels source file. No need to pass something from here.

lib/Backends/OpenCL/kernels_fwd_conv.cl
```diff
@@ -29,12 +29,25 @@
 // VWM - vector width in dimension M.
 
 static const char* FWD_CONV_CODE = R"(
+#ifndef QUANTIZED
```
opti-mix: As I commented above, most of this macro-based magic is not needed anymore, because you have two kernels now.

lib/Backends/OpenCL/kernels_fwd_conv.cl
```diff
@@ -319,4 +335,221 @@ conv_forward_mem(__global void *mem, unsigned im_in_offset, unsigned wg_offset,
...
+  const int_tp tidn = get_local_id(0);
+  // Local col ID (max: RTSN=TSN/WPTN).
+  const int_tp tidm = get_local_id(1);
+  // Work-group offset.
+  const int_tp offN = TSN * get_group_id(0);
+  // Work-group offset.
+  const int_tp offM = TSM * get_group_id(1);
+  // Local tile memory.
+  // Asub for loading weights & shuffling the output.
+  volatile __local Dtype Asub[TSM][TSK + v_pad_A];
```
opti-mix: I wonder of Asub and BSub should be arrays of "int" rather than "char" for performance reasons. But this is probably something to be benchmarked later.

lib/Backends/OpenCL/kernels_fwd_conv.cl
```diff
@@ -319,4 +335,221 @@ conv_forward_mem(__global void *mem, unsigned im_in_offset, unsigned wg_offset,
...
+              VEC_4_0(Areg) = Asub[row + 0 * RTSM][k];
+              VEC_4_1(Areg) = Asub[row + 1 * RTSM][k];
+              VEC_4_2(Areg) = Asub[row + 2 * RTSM][k];
+              VEC_4_3(Areg) = Asub[row + 3 * RTSM][k];
+#pragma unroll
+
+
+              for (int_tp wn = 0; wn < WPTN / VWN; ++wn) {
+                VEC_4_0(Creg[wm * VWM + 0][wn]) +=
+                    A_OFFSET(VEC_4_0(Areg)) * B_OFFSET(VEC_4_0(Breg[wn]));
```
opti-mix: It seems like there are a lot of repeating A_OFFSET or B_OFFSET computations with the same value of the argument. It could be a good idea to compute them just once. It is probably easier to actually perform these computations when initializing the values of `Areg` and `Breg` above. Or even when initializing the values of `Asub` and `Bsub`.

lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -36,9 +36,9 @@ bool OCLBackend::transformPostLowering(Function *F,
   bool changed = false;
   for (auto &node : F->getNodes()) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
-      if (CN->getInput()->getType(0)->isQuantizedType()) {
-        continue;
-      }
+      // if (CN->getInput()->getType(0)->isQuantizedType()) {
```
jfix71: Remove this commented out code?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -341,6 +341,10 @@ static bool sinkCode(Function *F) {
         continue;
       }
 
+      if (node->getNthResult(0)->getType(0)->isQuantizedType()) {
```
jfix71: Was this just for debugging purposes?

opti-mix:
> I have to disable some sinkNode optimization for the arithmetic ops that follows the OCLConvolution node. Due to the type inconsistency caused by the transformation from ConvolutionNode to OCLConvolutionNode (different scale and offset)

Can you elaborate regarding the type inconsistency? How does it manifest itself? Does it result in a verifier assertion? Or may be it crashes the compiler?

lib/Backends/OpenCL/kernels_fwd_conv.cl
```diff
@@ -29,12 +29,25 @@
...
 #define Dtype16 float16
+#define ZERO 0.0
+#else
+#define Dtype char
+#define Dtype1 char
+#define Dtype2 char2
+#define Dtype4 char4
+#define Dtype8 char8
+#define Dtype16 char16
+#define ZERO 0
```
rdzhabarov: zeros in quantized version correspond to the offset. Given fp = scale (q - offset), if you want to get fp equal to 0 you should use q == offset.

opti-mix:
> I have to disable some sinkNode optimization for the arithmetic ops that follows the OCLConvolution node. Due to the type inconsistency caused by the transformation from ConvolutionNode to OCLConvolutionNode (different scale and offset)

@rdzhabarov and I looked at the issue and found a problem. I'll submit the fix soon.

opti-mix:
@ZchiPitt The fix for the sinkCode is merged as #1352

ZchiPitt:
Thx guys for the review and bug fixes. Close it and submit in a new PR later.

PR #1351, opened 2018-07-27
[trivial] Add ReLU to the list of supported operations for openCL.

rdzhabarov:
This should help to make element-wise max operation to be quantized in the Resnet50 with openCL backend.
The conversion logic/execution is covered by quantization end2end test. cc: @ZchiPitt


ZchiPitt:
Add ReLU to the list rn would cause the slow convolution node to produce wrong result. Thats why i didnt include it in my latest pr.

ZchiPitt:
Not sure if it is because not all nodes lowered from ReLU have the support for quantization now

rdzhabarov:
ok, we need to check why this does not work.

PR #1352, opened 2018-07-27
[quantization] A couple of bug fixes to support quantization

opti-mix:
- Fix the implementation of createCmpEQ
  The implementation should be aware of the quantization. Use the same approach that we use for createCmpLTE.

- Fix a quantization related bug in sinkCode
  When sinkCode performs the arith(transpose(x), transpose(y)) -> transpose(arith(x, y)) transform, it should take the quantization into account. Specifically, it should create correct types for the transformed arith operation.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -959,31 +959,30 @@ ARITHMETIC_FUN_DEF(Max);
 ARITHMETIC_FUN_DEF(Min);
 #undef ARITHMETIC_FUN_DEF
 
+/// \returns the result type for a logical operation.
+static TypeRef getResultTypeOfLogicalOp(Module &M, TypeRef T) {
+  TypeRef OT;
+  if (T->isQuantizedType()) {
+    OT = M.uniqueType(T->getElementType(), T->dims(), 1.0, 0);
+  } else {
+    OT = M.uniqueType(*T);
+  }
+  return OT;
+}
+
 // For the quantized CmpLTE instruction, we require that the scale params be
 // (1.0, 0), so that the actual value and comparison value match.
 CmpLTENode *Function::createCmpLTE(llvm::StringRef name, NodeValue LHS,
                                    NodeValue RHS) {
   assert(LHS.dims() == RHS.dims() && "Invalid operand shapes");
-  TypeRef OT;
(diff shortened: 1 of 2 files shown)


opti-mix:
cc @ZchiPitt

opti-mix:
@rdzhabarov, thanks again for your help with fixing these bugs.

ZchiPitt:
looking cool!

lib/Graph/Graph.cpp
```diff
@@ -977,7 +977,13 @@ CmpLTENode *Function::createCmpLTE(llvm::StringRef name, NodeValue LHS,
 CmpEQNode *Function::createCmpEQ(llvm::StringRef name, NodeValue LHS,
                                  NodeValue RHS) {
   assert(LHS.dims() == RHS.dims() && "Invalid operand shapes");
-  auto OT = getParent()->uniqueType(*LHS.getType());
+  TypeRef OT;
```
rdzhabarov: could you extract the piece of code for getting OT type into a separate static method? (reuse that for all boolean ops).

opti-mix: Done.

opti-mix:
@qcolombet I'll add a test later. Merging it now to unblock @ZchiPitt.

lib/Graph/Graph.cpp
```diff
@@ -959,25 +959,30 @@ ARITHMETIC_FUN_DEF(Max);
 ARITHMETIC_FUN_DEF(Min);
 #undef ARITHMETIC_FUN_DEF
 
+/// \returns the result type for a logical operation.
```
nadavrot: I am not sure what you mean buy "logical op".  When I read the code it looks like it's uniquing the type, which is something that that name does not specify. I am not sure what's going on here.

opti-mix: This helper function is used only by the functions creating logical operations like `createCmpEQ` and `createCmpLTE` to compute their result types. Hence the name. The result type is computed based on the type of the operand of `createCmpEQ`. In quantized cases we provide concrerte scale and offset for the type as we know that the result of a logical operation could be only 0 or 1. I can extend the function comment with this information.

Issue #1334, opened 2018-07-26
Need to redesign the Variable interface.


nadavrot:
## Motivation

Currently Glow variables are backed by Tensors, and are tied to the graph. This gives us the ability to perform optimizations such as constant propagation that mutate the content of Tensors. However, it also ties the input/output tensors to the module. This design prevents us from running multithreaded programs simply because the input/output variables (which need to be different for each thread) are a single entity. 

## Proposed design

The proposed solution is to separate the concept of symbolic variables from graph variables. For example, an input variable X will not be backed by a Tensor until the execution-engine executes the compiled program. To execute the network the caller would need to bind Tensors to variables. The execution engine actually already does this, so we won't need to change the APIs that are used by our unit tests.

To preserve the ability to perform constant propagation (and other tensor-content optimizations) we would need to be able to make some variables concrete during the compilation phase. We would need to have two kinds of variables:

*Constant* - this is a tensor-backed variable. Similar to LLVM's ConstantArray it's a data structure that lives at the module-level and can be manipulated by transformations.

*Variable* - this is a typed variable that isn't backed by a real tensor until runtime. The runtime allocates the tensor (either in device memory or on the host) and manages the input/output to and from the variable.

stoklund:
This sounds great! A couple comments:

The name *ConstantVariable* is a bit of an oxymoron. A constant is something that doesn't change, and a variable is something that does change. I would suggest *Constant* and *Variable* instead.

Second, I think we should also look at separating constants from the module. The module represents the compiler's intermediate representation, but right now it isn't intermediate at all — the constants in the module are needed at runtime, so you have to keep it around.

More generally, we should make sure we are economical with memory management in a production environment. I imagine something like:

- The *Module* only exists during compilation, then it is deleted. We don't need the IR any more.
- After compilation, constants that are needed at runtime are moved into a constant pool. Any constants that are not needed at runtime get deleted along with the module.
- A constant pool can be shared by multiple runtime variable bindings, and also by multiple compiled functions. (e.g., compiled functions representing different batch sizes of the same graph).
- For constants we didn't need to modify during compilation, this should all happen without any copies of tensor data.

bertmaher:
This proposal sounds great to me.  A closely-related issue for multithreaded programs is where we allocate activations.  We should hopefully be able to use the same mechanism for `RuntimeVariable`s and activations.

nadavrot:
@stoklund Thank you for the feedback. 

[...]

PR #1353, opened 2018-07-27
Final push to clean our overwritten arguments

qcolombet:
The quantization profile node was the last use (aside from save nodes) of the overwritten attribute and thus, was also sensible to the same kind of scheduling bugs that was exposed in #1283.

There are several ways we could have fixed/worked around that:
1. Teach the scheduler how to handle the implicit memory dependencies for that node too
2. Check in the verifier that the overwritten inputs of the quantization profile nodes are not used anywhere else
3. Model quantization profile with producing two outputs

I don't like #1 because it pushes more hacks in the scheduler (eventually I'd like we model the memory dependencies period.). #2 is really just a check that the situation does not happen, but if it does, we're back to square 1.
Therefore, I gave a quick stab at #3 with this PR.

The nice thing I was pursuing and that PR achieves is that we would have the invariant that only save nodes write to variable.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1572,14 +1572,28 @@ void InterpreterFunction::fwdDebugPrintInst(const DebugPrintInst *I) {
 void InterpreterFunction::fwdQuantizationProfileInst(
     const glow::QuantizationProfileInst *I) {
   auto inputTensor = getWeightHandle(I->getInputTensor());
-  auto currentHistogram = getWeightHandle(I->getHistogram());
-  auto computationInfo = getWeightHandle(I->getComputationInfo());
+  // Copy the input value into the output.
+  // Only do the copy if the optimizer didn't coalesce the buffers together.
+  auto *updatedHistogramTensor = getTensor(I->getUpdatedHistogram());
+  const auto *histogramTensor = getTensor(I->getHistogram());
+  if (updatedHistogramTensor != histogramTensor) {
+    updatedHistogramTensor->copyFrom(histogramTensor);
+  }
+  auto *updatedComputationInfoTensor =
+      getTensor(I->getUpdatedComputationInfo());
+  const auto *computationInfoTensor = getTensor(I->getComputationInfo());
+  if (updatedComputationInfoTensor != computationInfoTensor) {
+    updatedComputationInfoTensor->copyFrom(computationInfoTensor);
+  }
 
-  float &min = computationInfo.raw(0);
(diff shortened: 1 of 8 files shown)


jfix71:
> Is there a problem with enforcing a rule that operators that update some variables 'own' the variables?

@nadavrot By own, do you mean only the owner node can read from or write to the variable?

qcolombet:
> Implementing the embedding tables using the method that nodes produce new values is impractical

I agree and I don't think we should expose them as variables (at least not how it is done right now).

>  Is there a problem with enforcing a rule that operators that update some variables 'own' the variables?

I think that would be the way to go, but then I believe it should be some special kind of variables. One thing that we need to enforce is that "owned" variables don't have any other reads (which is where we introduce the problematic memory dependencies). The best way to do that would be to not expose these variables for other to grab, and we miss that today. Indeed, variables are available at a module level and this cannot be changed easily because of how we use the Module at runtime #1334.

nadavrot:
@qcolombet Do you know how XLA handles variable mutability?

@qcolombet @jfix71 How would you feel if we split variable into 3 categories: 
1. StorageVariable: this storage kind will be used by ProfileQuantize, EmbeddingTable and other nodes that need database-like access pattern where you can't just copy the whole thing every time. 
2. Constant: this variable kind will store weights.
3. IO: This kind of variables will be linked to actual thread-local buffers at runtime. 

What do you think?

qcolombet:
> Do you know how XLA handles variable mutability?

No, looking is still on my todolist :).

> How would you feel if we split variable into 3 categories:
> StorageVariable [...]

As long as those are not "gettable" to form other links that would work. The thing I want to be sure is that we don't create patterns that could lead into bugs into the scheduler.

qcolombet:
I wonder if we could have the variable being own (I mean litterally in the code) by the node itself.

qcolombet:
Note: I am happy to just trash this PR and go with the verifier check for now, until we land a better designed for the variables. I just hacked this this morning to see how difficult it would be to finally clean-up the overwritten problem and turned out it was easy and not too hacky, so I figured I shared :).

qcolombet:
It does prevent us for running into memory dependencies issues as of today though, which is nice IMHO.

qcolombet:
Abandoning this PR in favor of checks in the verifier.

qcolombet:
@nadavrot regarding mutability in XLA (hlo_instruction.h):
// HLO is pure (mostly).  It has no concept of mutable state.  Instead, data
// values are produced by one HLO and flow into consumers across dependency
// edges.

I was incidentally going in that direction :).

PR #1355, opened 2018-07-27
[quantization][opencl]: Support quantization for OCLConvolution

ZchiPitt:
The OCLConvolutionInst works perfect with the resnet50 now. However i have to comment out ReLUNode in isOpSupported() in openCL.h, otherwise it would cause wrong results.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -45,6 +45,8 @@ typedef uint32_t cl_size_t;
 #include "kernels.cl"
 // This defines kernels for optimized convolutions.
 #include "kernels_fwd_conv.cl"
+// This defines kernels for quantized optimized convolutions.
+#include "kernels_fwd_quantized_conv.cl"
 
 namespace {
 llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
@@ -393,6 +395,7 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   auto odim = ShapeNCHW(CC->getDest()->getType()->dims());
   auto idim = ShapeNCHW(CC->getSrc()->getType()->dims());
   auto fdim = ShapeNCHW(CC->getFilter()->getType()->dims());
+  bool isQuantized = output->getType()->isQuantizedType();
   PaddingTLBR pads(CC->getPads());
   // Create options for compiling the program.
   // Don't use names M, N, K as they are defined in precompiled headers.
@@ -469,13 +472,31 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
 
   // Generate a tailor-made convolution kernel using the provided options based
   // on the parameters of the current convolution.
-  auto prog = createProgram(FWD_CONV_CODE, options, commands_);
-  auto kernel = createKernel("conv_forward_mem", prog);
+  auto &programName = isQuantized ? FWD_CONV_QUANTIZED_CODE : FWD_CONV_CODE;
+  auto prog = createProgram(programName, options, commands_);
+  auto kernelName = isQuantized ? "conv_forward_mem_i8" : "conv_forward_mem";
+  auto kernel = createKernel(kernelName, prog);
   setKernelArg(kernel, 0, deviceBuffer_);
   setKernelArg<cl_uint>(kernel, 1, tensors_[input]);
   setKernelArg<cl_uint>(kernel, 2, tensors_[weights]);
   setKernelArg<cl_uint>(kernel, 3, tensors_[bias]);
   setKernelArg<cl_uint>(kernel, 4, tensors_[output]);
+
+  // Extra options for quantized kernel
+  if (isQuantized) {
+    auto inputTy = CC->getSrc()->getType();
+    auto outputTy = CC->getDest()->getType();
+    auto biasTy = CC->getBias()->getType();
+    auto weightsTy = CC->getFilter()->getType();
+    setKernelArg(kernel, 5, weightsTy->getOffset());
+    setKernelArg(kernel, 6, weightsTy->getScale());
+    setKernelArg(kernel, 7, inputTy->getOffset());
+    setKernelArg(kernel, 8, inputTy->getScale());
+    setKernelArg(kernel, 9, outputTy->getOffset());
+    setKernelArg(kernel, 10, outputTy->getScale());
+    setKernelArg(kernel, 11, biasTy->getOffset());
+    setKernelArg(kernel, 12, biasTy->getScale());
+  }
 
   // Compute proper parameters for global work and workgroups.
   int group = 1;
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -537,21 +554,25 @@ void OpenCLFunction::execute() {
...
-          // Start less kernels and let each kernel do more work using vector
-          // instructions.
-          global /= 16;
-          kernelName += "16";
-        } else if (global % 8 == 0) {
-          // Start less kernels and let each kernel do more work using vector
-          // instructions.
-          global /= 8;
-          kernelName += "8";
+      if (!isQuantized) {
```
opti-mix: May be move this check below `global = I.getOperand(0).first->getType()->size();` to make things a bit simpler.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -469,14 +472,28 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
...
+  auto kernel = createKernel(kernelName, prog);
   setKernelArg(kernel, 0, deviceBuffer_);
   setKernelArg<cl_uint>(kernel, 1, tensors_[input]);
   setKernelArg<cl_uint>(kernel, 2, tensors_[weights]);
   setKernelArg<cl_uint>(kernel, 3, tensors_[bias]);
   setKernelArg<cl_uint>(kernel, 4, tensors_[output]);
 
+  // Extra options for quantized kernel
+  if (isQuantized) {
+    setKernelArg(kernel, 5, CC->getFilter()->getType()->getOffset());
```
opti-mix: At the beginning of the function we have the following definitions:
```c++
 auto input = CC->getSrc();
  auto output = CC->getDest();
  auto bias = CC->getBias();
  auto weights = CC->getFilter();
```
Please use them here to shorten the code, instead of writing `CC->get...`

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -187,7 +187,7 @@ class OCLBackend final : public Backend {
       case Kinded::Kind::MinNodeKind:
       case Kinded::Kind::MulNodeKind:
       case Kinded::Kind::QuantizeNodeKind:
-      case Kinded::Kind::ReluNodeKind:
+      //case Kinded::Kind::ReluNodeKind:
```
opti-mix: If you don't use it in this PR, remove this line.

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
```diff
@@ -0,0 +1,329 @@
...
+/// Clips int into char.
+char clip(int val) { return (char)min(max(val, -128), 127); }
+
+__kernel
+__attribute__((reqd_work_group_size(workgroup_size_0, workgroup_size_1, 1)))
+__attribute__((vec_type_hint(Dtype4)))
+void
+conv_forward_mem_i8(__global void *mem, unsigned im_in_offset, unsigned wg_offset,
+                 unsigned bias_offset, unsigned im_out_offset,
+                 int Aoffset, float Ascale, int Boffset, float Bscale,
```
opti-mix: The naming of offsets and scales does not match the naming of other parameters, which start lowercase and use underscores.

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
```diff
@@ -0,0 +1,329 @@
...
+              VEC_4_0(Areg) = Asub[row + 0 * RTSM][k];
+              VEC_4_1(Areg) = Asub[row + 1 * RTSM][k];
+              VEC_4_2(Areg) = Asub[row + 2 * RTSM][k];
+              VEC_4_3(Areg) = Asub[row + 3 * RTSM][k];
+#pragma unroll
+
+
+              for (int_tp wn = 0; wn < WPTN / VWN; ++wn) {
+                VEC_4_0(Creg[wm * VWM + 0][wn]) +=
+                    A_OFFSET(VEC_4_0(Areg)) * B_OFFSET(VEC_4_0(Breg[wn]));
```
opti-mix: Can you move these repeating A_OFFSET and B_OFFSET computations into the loops above. where you assign to Areg and Breg?

lib/Backends/OpenCL/OpenCL.cpp
ZchiPitt: cool

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
ZchiPitt: changed

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
ZchiPitt: weird that when i moved B_OFFSET up to the loops, the results are wrong. But i dont see the reasons why

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -873,14 +896,19 @@ void OpenCLFunction::execute() {
       setKernelArg(kernel, numArgs + 6,
                    ShapeNHWC(CC->getFilter()->getType()->dims()));
       if (isQuantized) {
-        setKernelArg(kernel, numArgs + 7, CC->getDest()->getType()->getOffset());
+        setKernelArg(kernel, numArgs + 7,
```
opti-mix: Can you do the same trick with introduce a var for type and then using it to make the setKernelArg lines shorter?

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
opti-mix: > weird that when i moved B_OFFSET up to the loops, the results are wrong. But i dont see the reasons why

That is very strange. It doesn't make any sense. Try to figure out why it does not work. IMHO, it should work without any issues.

rdzhabarov:
"However i have to comment out ReLUNode in isOpSupported() in openCL.h, otherwise it would cause wrong results." Is this still the case after opti-mix fixes?

ZchiPitt:
@rdzhabarov yea i merged with the latest commnets, and its still the case after the fixes

lib/Backends/OpenCL/OpenCL.cpp
ZchiPitt: done

lib/Backends/OpenCL/kernels_fwd_quantized_conv.cl
ZchiPitt: changed the type of Areg/Breg from char4 to int4

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -187,7 +187,6 @@ class OCLBackend final : public Backend {
       case Kinded::Kind::MinNodeKind:
       case Kinded::Kind::MulNodeKind:
       case Kinded::Kind::QuantizeNodeKind:
-      case Kinded::Kind::ReluNodeKind:
```
rdzhabarov: Please, investigate why it impacts the results. We have all needed support in OpenCL backend to allow this, there must be a bug ....

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -539,16 +560,18 @@ void OpenCLFunction::execute() {
...
-          // Start less kernels and let each kernel do more work using vector
-          // instructions.
-          global /= 16;
-          kernelName += "16";
-        } else if (global % 8 == 0) {
-          // Start less kernels and let each kernel do more work using vector
-          // instructions.
-          global /= 8;
-          kernelName += "8";
+        if (!isQuantized) {
```
opti-mix: Can you add a comment that this a temporary workaround until the corresponding kernels are implemented for the quantized operations? Or may be you do not plant to implement them?

ZchiPitt: Comments added. I probably would submit a following pr to add the corresponding operations

lib/Backends/OpenCL/OpenCL.h
ZchiPitt: sure, working on it

PR #1356, opened 2018-07-28
Support multiple outputs in proto loaders and runners

jfix71:
The loaders were pretty specific to image classification before this, and so assumed/only supported loading networks with a single output.

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -98,8 +98,8 @@ protected:
   std::unordered_map<std::string, Node *> nodeByName_;
   /// A list of weight tensors indexed by name.
   std::unordered_map<std::string, Tensor *> tensors_;
-  /// The external output of the network.
-  SaveNode *root_{nullptr};
+  /// A map from names of the external outputs of the network to SaveNodes.
+  std::unordered_map<std::string, SaveNode *> outputsByName_;
 
   /// \returns the tensor that was registered under the name \p name.
   Tensor *getTensorByName(llvm::StringRef name);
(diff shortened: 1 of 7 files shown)


include/glow/Importer/ProtobufLoader.h
```diff
@@ -149,9 +149,17 @@ class ProtobufLoader {
...
+  /// classification, this single final output is usually the result of the last
   /// softmax or regression layer.
-  SaveNode *getRoot() { return root_; }
+  SaveNode *getSingleOutput() {
+    assert(outputsByName_.size() == 1);
+    return outputsByName_.begin()->second;
+  }
+
+  /// \returns the SaveNode for the external output with \p name.
+  SaveNode *getOutputByName(llvm::StringRef name) const;
```
qcolombet: We should add a precondition (\pre) that the output with this name must exist or fix the implementation to return nullptr when it doesn't.

jfix71:
@qcolombet I tested this on a proto with multiple outputs. However there are other changes required stacked on this PR to get that proto running which I will be upstreaming this week. I believe @beicy is working on unit tests for our loaders which would be great and allow us to test changes like these!

include/glow/Importer/ProtobufLoader.h
jfix71: Done

lib/Importer/ONNX.cpp
```diff
@@ -465,10 +465,14 @@ void ONNXModelLoader::loadInitializers(onnx::GraphProto &net) {
   }
 }
 
-void ONNXModelLoader::setOutputNode(onnx::GraphProto &net) {
+void ONNXModelLoader::setOutputNodes(onnx::GraphProto &net) {
   assert(net.output_size() && "Network needs external outputs defined.");
-  auto *r = getNodeByName(net.output(0).name());
-  root_ = G_.createSave("output", r);
+
+  for (unsigned i = 0; i < net.output_size(); i++) {
```
rdzhabarov: nits: all those protobuf methods for sizes return "int" type.

PR #1357, opened 2018-07-28
Work around OrcJIT API instability

bertmaher:
I'm not super-happy about this PR, but the API for Orc JIT seems to have churned a lot.  I've tried llvm stable, llvm master, and FB's internal stable branch, and they *all* have little differences from each other, and of course from llvm6.  So... I'm thinking -DFACEBOOK is about the best way to handle this for the present, and at some point we can migrate fully to llvm7 once that's a stable release.

### DIFF
--- a/lib/Backends/CPU/BundleSaver.cpp
+++ b/lib/Backends/CPU/BundleSaver.cpp
@@ -147,7 +147,11 @@ void BundleSaver::produceBundle(llvm::StringRef outputDir) {
               "Could not open the output file for saving the bundle code");
   if (fileName.endswith(".bc")) {
     // Emit the bitcode file.
+#ifdef FACEBOOK
+    llvm::WriteBitcodeToFile(M, outputFile);
+#else
     llvm::WriteBitcodeToFile(&M, outputFile);
+#endif
   } else if (fileName.endswith(".o")) {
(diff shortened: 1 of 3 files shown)


nadavrot:
Thank you for doing this work. It's very important for the success for the project. 

I understand that supporting multiple versions of LLVM will be tricky. Maybe we can refactor the code and write it in a way that will minimize the @ifdefs? I wonder if @compnerd or @qcolombet have advice here.

I have one code-review comment:  maybe FACEBOOK_INTERNAL or FBCODE_BUILD would be a better name for the DEF name?

qcolombet:
Could we use the C-API instead? (OrcBindings.h)
They are supposed to be stable between releases.

bertmaher:
Using the C API is an interesting idea.  It's kind of funny though that OrcBindings.h has a disclaimer at the top: "Note: This interface is experimental. It is *NOT* stable, and may be changed without warning."

qcolombet:
I guess it is too new for the C-API to be stable. But that should be idea at some point.

bertmaher:
I was able to get rid of the #ifdef on initializeAllTargets, so, progress at least.  I'll also do FACEBOOK -> FACEBOOK_INTERNAL.

PR #1358, opened 2018-07-30
Minor cleanups addressing the code review feedback for #1352. NFC

opti-mix:
- Rename getResultTypeOfLogicalOp into getResultTypeOfBooleanOp and improve the comments

- Add a test to check that Cmp nodes are created with proper output types

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -965,8 +965,15 @@ ARITHMETIC_FUN_DEF(Max);
 ARITHMETIC_FUN_DEF(Min);
 #undef ARITHMETIC_FUN_DEF
 
-/// \returns the result type for a logical operation.
-static TypeRef getResultTypeOfLogicalOp(Module &M, TypeRef T) {
+/// This helper function is used only by the functions creating boolean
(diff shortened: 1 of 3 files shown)


lib/Graph/Graph.cpp
```diff
@@ -965,8 +965,15 @@ ARITHMETIC_FUN_DEF(Max);
...
-static TypeRef getResultTypeOfLogicalOp(Module &M, TypeRef T) {
+/// This helper function is used only by the functions creating boolean
+/// operations like createCmpEQ and createCmpLTE to compute their output types.
+/// The output type is computed based on the type \p T of the operand of the
+/// logical operation. In case of a quantized type we provide concrete scale and
+/// offset for the type as we know that the output of a boolean operation could
+/// be only 0 or 1.
+///
+/// \returns the output type for a boolean operation.
+static TypeRef getResultTypeOfBooleanOp(Module &M, TypeRef T) {
```
rdzhabarov: consider renaming in the GraphOptimizer as well (LOGICAL_OP_CASE) to be consistent with this naming schema.

opti-mix: Done.

PR #1360, opened 2018-07-30
[Graph] Check that variables are either written by save node or "owned"

qcolombet:
The concept of an owned variable is a variable that is read and written
by the node that owns it (like QuantizationProfile with its Histogram
and ComputationInfo variables). Technically an owned variable is
something that should be exposed for other nodes to use, but because
variables are owned by the module for proper runtime allocation and so
on, we still have to register them there.

This patch teaches Function::verify how to recognize owned variables
and makes sure they are properly used. This basically papers over
our lack of higher level design for such concept at the moment.

NFC.


qcolombet:
This is an alternative for #1353.

nadavrot:
Nice!

PR #1361, opened 2018-07-30
[InstrGen] Teach inplaceOperand how to deal with several results

qcolombet:
Prior to this commit, it was impossible to describe more than one
inplace property. This would limit our ability to set the inplace
property on instruction with more than one result.

With this patch, it is now possible to define one inplace property
list for each output of an instruction. To use this feature
simply invoke inplaceOperand method for each output on the
declaration of the instruction.

### DIFF
--- a/tools/ClassGen/InstrBuilder.cpp
+++ b/tools/ClassGen/InstrBuilder.cpp
@@ -101,13 +101,15 @@ void InstrBuilder::emitIRBuilderMethods(std::ostream &osH,
 void InstrBuilder::emitInplaceMethod(std::ostream &os) const {
   os << "\n  bool isInplaceOp(unsigned dstIdx, unsigned srcIdx) const {\n";
   if (!inplaceOperands_.empty()) {
-    assert(inplaceOperands_.size() > 1 &&
-           "We don't have a pair of inplace args");
-    for (int i = 1, e = inplaceOperands_.size(); i < e; i++) {
-      auto F0 = getOperandIndexByName(inplaceOperands_[0]);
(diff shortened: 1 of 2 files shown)


jfix71:
Up to you. It seems like something we should support and I don't think it's highly likely to be unintentionally broken without also breaking current inplace usage.

qcolombet:
Hi @jfix71,

I spent some time to see how difficult it would be to add a proper/independent tester for our instr/node generators. I made an additional commit on top of the inplaceOperand fix for a proof-of-context.

In a nutshell, this commit makes our builders libraries that we can reuse with testers and it adds expected output to check with.

The path forward would be to update the expected outputs (header.h in that case) as we add more features/fix bugs.

This is far from being commitable, in particular because it does not enforce the proper dependencies for the check target and uses unix like script for checking the output.

I wanted to still share it in case people think it would be a useful feature to have. I believe it could be useful, but on the other hands we also don't modify this code often enough that it is really worth spending a lot of time on this, so I am torned.

What do you think?

jfix71:
Cool! Thanks for taking a look at this. 

I agree it might not make much sense to add the test and its added complexity given that we don't modify the builders very often. I also would imagine in practice it's highly unlikely this would catch a bug that is not caught in other unit tests. That said, it would make debugging much easier in case of such a bug, and more tests are generally a good thing. Overall I'd probably lean against having this, but am also a bit torn.

qcolombet:
I've ripped the testing part off this commit.
I'll keep it on the back burner if I have some spare time or someone is interested into it again.

PR #1362, opened 2018-07-30
Start to untangle Variable: move init into Tensor

nadavrot:
[Tensor] move the init method from Variable to Tensor.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -47,6 +47,15 @@ ShapeVector expandDimsToMax(llvm::ArrayRef<size_t> currDims);
 
 /// A class that represents a contiguous n-dimensional array (a tensor).
 class Tensor final {
+public:
+  /// Specifies the kind initialization for the tensor.
+  enum class InitKind {
+    Zero,      // The tensor is initialized to zero.
+    Broadcast, // Broadcast a single value to all elements.
+    Xavier,    // Init the variable with random values using the Xavier method.
+  };
+
+private:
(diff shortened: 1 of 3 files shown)


qcolombet:
Hi @nadavrot ,

I like the direction of this patch! Could you share what are the steps going forward?
In particular, I've just posted #1363 and I can totally see it being completely relevant (in particular if we have variables taking of the tensor being past instead of just being copied) or totally useless :).

nadavrot:
@qcolombet 

In #1334 we started the discussion of Variable redesign. The idea is to have multiple kinds of variables (constants and mutable interface variables). In this commit I am only moving junk out of variable, because sooner or later we'll need to clean things around here.  

https://github.com/pytorch/glow/issues/1334

include/glow/Base/Tensor.h
```diff
@@ -47,6 +47,15 @@ ShapeVector expandDimsToMax(llvm::ArrayRef<size_t> currDims);
 
 /// A class that represents a contiguous n-dimensional array (a tensor).
 class Tensor final {
+public:
+  /// Specifies the kind initialization for the tensor.
+  enum class InitKind {
+    Zero,      // The tensor is initialized to zero.
```
rdzhabarov: It seems like Zero is a special case of Broadcast with 0 value. Do we really need to have this extra init type for tensor?

nadavrot: You are right. This part will go away when we finish the refactoring. I am trying to minimize code changes in this massive refactoring.

PR #1363, opened 2018-07-30
[Graph] Add a method to create a variable directly from a tensor

qcolombet:
@jfix71, we talked about having a way to create variables directly from a tensor when we were discussing the CSE optimization for variables (#1263).
That PR does that. Given how few uses there are, I am not convince we should push it, but I thought I'll share it in case people are interested into doing that.


nadavrot:
LGTM.

PR #1365, opened 2018-07-31
[Variable] Remove the tensor initialization from Variable. NFC.

nadavrot:
This commit removes the initialization code from Variable and moves it to the code that creates the variable. After this change Tensors needs to be initialized externally and not by the Variable constructor like they are today. This is a pretty big patch that touches many many files, but most of the changes are mechanical. 

I tried to make this patch as simple as possible and there are still a few things that we need to clean up. For example, the default values for arguments don't make sense in a few places. We should change the boolean training flag to an enum, etc.


rdzhabarov:
Feel free to merge after resolving conflicts, it's easy to got conflicts with this PR.

PR #1374, opened 2018-07-31
[fix] Fix the problem ReLU node does not work with OpenCL quantization

ZchiPitt:
Found the bug that splat int8 kernel always receives 0 as the input value. Fixed it and ReLU works with the opencl quantization now.


ZchiPitt:
@rdzhabarov bc the type of val in the kernel is set to be float. When we pass int8 value into it, it would get messed up.

PR #1375, opened 2018-08-01
[opencl][quantization]: Support quantization for OCLPoolMax

ZchiPitt:
Support quantization for OCLPoolMax on OpenCL backend


opti-mix:
@ZchiPitt I wish we could use C++ instead of C for OpenCL ;-) This would allow us to get rid of all these macro-based definitions that we use to model templates.

opti-mix:
@ZchiPitt What about the unit tests? I do not see anything in this PR.

ZchiPitt:
@opti-mix oops,forgot about the test, will add soon. But i did test it with the resnet50

ZchiPitt:
@rdzhabarov there're still some operators left for end2end, i will enable it for opencl later, after i finished poolavg

ZchiPitt:
Added a test case for now

rdzhabarov:
@ZchiPitt for end2end test you do not need to implement all operators. Operators that are not implemented (not isOpSupported) will be present as fp32 operators in the graph and test should work as expected.

PR #1377, opened 2018-08-01
Assorted Caffe2 importer support for seq2seq

jfix71:
Each commit is pretty self explanatory.

### DIFF
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -389,6 +389,35 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Slice") {
+    auto data = getNodeValueOrCreateVariableByName(op.input(0));
+
+    auto starts = getShape<ssize_t>(dict["starts"]);
+    auto ends = getShape<ssize_t>(dict["ends"]);
+
+    std::vector<size_t> newStarts, newEnds;
+    assert(starts.size() == ends.size());
+    for (size_t i = 0; i < starts.size(); i++) {
+      ssize_t newStart = starts[i];
+      if (newStart == -1) {
+        newStart = data->dims(0)[i];
+      }
+      assert(newStart >= 0 && "Indices should never be negative.");
+      newStarts.push_back(newStart);
+
+      ssize_t newEnd = ends[i];
+      if (newEnd == -1) {
+        newEnd = data->dims(0)[i];
+      }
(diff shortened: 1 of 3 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1299,6 +1299,14 @@ ScatterAssignNode *Function::createScatterAssign(llvm::StringRef name,
   return addNode(new ScatterAssignNode(name, data, indices, slices));
 }
 
+Variable *Function::createShape(llvm::StringRef name, NodeValue input) {
```
artemrakhov: I am not convinced that this should be in Graph API. Would you use it if you were to create graph using C++ API?

lib/Importer/Caffe2.cpp
```diff
@@ -415,6 +447,65 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Tanh") {
```
artemrakhov: Can we have this and other ops in CommonOperatorLoader? I think they have the same semantics in ONNX.

lib/Importer/Caffe2.cpp
```diff
@@ -470,12 +561,23 @@ void caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
       if (dict["values"]->floats_size()) {
         assert(typeName != "GivenTensorIntFill" &&
                typeName != "GivenTensorInt64Fill");
+
+        // Assume a 1D vector of length of input values.
+        if (dim.size() == 0) {
+          dim.push_back(dict["values"]->floats_size());
```
artemrakhov: This is 0-dims Tensor, right? Why can't we leave it as 0 dims? Have you seen cases where `dict["values"]->floats_size()` is more than 1?

lib/Importer/Caffe2.cpp
jfix71: I can move it/others, but I have no ONNX proto with these operators to make sure they work correctly, so I was conservatively adding only for C2.

jfix71:
I'm going to just directly create and set the values for the Shape loader. I think `createVariableFromValue()` is a good idea but it would likely be used very widely and so it's not in scope for this PR.

lib/Importer/Caffe2.cpp
jfix71: I moved all but Slice -- there is a slight difference in the ONNX/C2 implementations for the start and end indices and negative values, so I'd rather not add it in without having a proto with it to test that it works.

Issue #1359, opened 2018-07-30
Minimize #ifdef FACEBOOK_INTERNAL in GlowJIT


bertmaher:
PR #1357 introduced some #ifdefs to GlowJIT.{h,cpp} to work around the shifting ORC JIT API in llvm 7.0.  We can probably minimize those by:
* switching to the C API, if that is stable (OrcBindings.h, as mentioned by @qcolombet in #1357)
* thinning out the GlowJIT API (with a bit of work we can actually remove some of its API functions and types, so there's less to guard)
* figuring out the llvm 7.0 API for GDB registration listeners so we can do something sensible with `NotifyLoadedFunctor`

And at some point the LLVM  7.0 API will stabilize and we can shift to major version guards instead of `FACEBOOK_INTERNAL`.

vedanuj:
@bertmaher Can I take a stab at this if no one else is working on it?

bertmaher:
@vedanuj Be my guest!

PR #1379, opened 2018-08-01
Change FACEBOOK_INTERNAL to generic LLVM_VERSION_MAJOR and reduce som…

vedanuj:
This PR starts to reduce some of the `ifdefs` as requested in issue #1359. 

This is the first version to clean up some of the ifdefs and make it as generic as possible. I have switched from `FACEBOOK_INTERNAL` to `LLVM_VERSION_MAJOR` to make it a bit more generic. Please let me know if that is ok or I should keep the `FACEBOOK_INTERNAL` check as it is.

I am not sure how to test this properly with Facebook's version of llvm. Some guidance how to do it will be really helpful.

I am not entirely sure how to use the `OrcBindings.h` api. If I can get some pointers here it will be really helpful. I tried to check if there is anything for `NotifyLoadedFunctor ` in LLVM 7.0 but couldn't find anything yet. I will keep looking if I can find something. 

cc @bertmaher @nadavrot @rdzhabarov please review this version

### DIFF
--- a/lib/Backends/CPU/BundleSaver.cpp
+++ b/lib/Backends/CPU/BundleSaver.cpp
@@ -147,7 +147,7 @@ void BundleSaver::produceBundle(llvm::StringRef outputDir) {
               "Could not open the output file for saving the bundle code");
   if (fileName.endswith(".bc")) {
     // Emit the bitcode file.
-#ifdef FACEBOOK_INTERNAL
+#if LLVM_VERSION_MAJOR > 6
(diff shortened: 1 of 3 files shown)


bertmaher:
This looks awesome, thanks for doing it!  Does this work with llvm master or llvm stable?

vedanuj:
I followed the macOS setup for `glow` and `homebrew` installed `llvm` stable 6.0.1. I tested on that. I can test with FACEBOOK's version as well if you can point me how to do it.

PR #1382, opened 2018-08-01
[IR] Add a test that we can compile code without graph deps.

nadavrot:
This commit adds a unit test that checks that we can compile functions
without depending on the IR.  We compile some function and then delete
the function. Later we execute the code and check that things work.

### DIFF
--- a/tests/unittests/BackendTest.cpp
+++ b/tests/unittests/BackendTest.cpp
@@ -146,6 +146,33 @@ TEST_P(BackendTest, debugPrint) {
   function->execute();
 }
 
+/// This test checks that we can compile a function without depending on the
+/// graph representation. We compile some function and then delete the function.
+/// Later we execute the code and check that things work.
+TEST_P(BackendTest, decoupleCodegenFromGraph) {
+  Module mod;
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createVariable(ElemKind::FloatTy, {3}, "X");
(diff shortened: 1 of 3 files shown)


tests/unittests/BackendTest.cpp
```diff
@@ -146,6 +146,32 @@ TEST_P(BackendTest, debugPrint) {
...
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createVariable(ElemKind::FloatTy, {3}, "X");
+  X->getPayload().getHandle() = {1., 2., 3.};
+  auto *pow = F->createPow("Pow1", X, 2.0);
+  auto *save = F->createSave("save", pow);
+  Variable *res = save->getVariable();
+  EE_.compile(CompilationMode::Infer, F);
+
+  // Erase all of the functions to ensure that the compiled code does not
+  // depend on the IR.
```
opti-mix: Do you mean "graph representation" when you say "IR" here? Our interpreter or OpenCL cannot run without the low-level IR. But they can run without the graph.

tests/unittests/BackendTest.cpp
```diff
@@ -146,6 +146,32 @@ TEST_P(BackendTest, debugPrint) {
...
+  mod.eraseFunctions();
+
+  // We can run the compiled code without having the IR around.
```
opti-mix: Same here.

nadavrot:
@opti-mix You are correct. Fixed.

PR #1388, opened 2018-08-02
[Importers] Support two trivial operators: Cast and StopGradient.

artemrakhov:


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -241,6 +241,15 @@ class CommonOperatorLoader : public ProtobufLoader {
     nodeValueByName_[op.output(0)] = NodeValue(in, 0);
   }
 
+  void loadCast(const OpType &op, ArgumentDictionaryTy &dict) {
```
nadavrot: Does Cast imply a possible reshape?

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -241,6 +241,15 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    assert(in.getElementType() == ElemKind::FloatTy &&
+           "Only float to float cast is currently supported.");
+    int to = loadInt(dict["to"]);
+    assert(to == 1 && "Only float to float cast is currently supported.");
```
jfix71: Can we create an enum of the element types with their id values from Caffe2/ONNX? Instead of hardcoding `1` here.

include/glow/Importer/CommonOperatorLoader.h
jfix71: [Cast only changes the type](https://caffe2.ai/docs/operators-catalogue.html#cast)

include/glow/Importer/CommonOperatorLoader.h
artemrakhov: Done. This enum already exists in proto, but unfortunately it is different for Caffe2 and ONNX. Moved the code to Caffe2 importer.

qcolombet:
Random comment, what would it take to have unitests for each change we do in the importer?

artemrakhov:
@qcolombet : tests are coming soon. We are going to have more coverage for both importers.

PR #1390, opened 2018-08-02
[cmake] Copy .pb.h headers to namespaced paths

bertmaher:


lib/Importer/CMakeLists.txt
```diff
@@ -5,17 +5,39 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
...
+#   #include "glow/caffe.pb.h"
+#
+# instead of:
+#
+#   #include "caffe.pb.h"
+#
+# Please don't optimize away the "useless" copies!
+add_custom_command(
+  OUTPUT ${GLOW_BINARY_DIR}/glow/caffe.pb.h
+  COMMAND cp ${CAFFE_HDRS} ${GLOW_BINARY_DIR}/glow/caffe.pb.h
```
compnerd: It is better to use `${CMAKE_COMMAND} -E copy_if_different` to make this portable.  Particularly since there is already a PR to make this build/work on Windows.  That also has the benefit of not causing a rebuild all the time since the timestamps on the file shouldn't change unless the content changes.

lib/Importer/CMakeLists.txt
```diff
@@ -5,17 +5,39 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
...
+  DEPENDS ${CAFFE_HDRS})
+
+add_custom_command(
+  OUTPUT ${GLOW_BINARY_DIR}/onnx/onnx.pb.h
+  COMMAND cp ${ONNX_HDRS} ${GLOW_BINARY_DIR}/onnx/onnx.pb.h
```
compnerd: Same.

PR #1392, opened 2018-08-02
A couple softmax-related updates for the importer/loader

jfix71:


tools/loader/ImageClassifier.cpp
```diff
@@ -137,22 +137,20 @@ int main(int argc, char **argv) {
 
   // Create the model based on the input format, and set the Softmax save node
   // expected to come at the end of image inference.
-  Tensor expectedSoftmax(ElemKind::IndexTy, {1, 1});
```
beicy: The comment above should be updated?

jfix71: I think the comment is still correct -- it's referring to `SM` below, which we still look for and set with `SM = LD.getSingleOutput();`

PR #1393, opened 2018-08-02
[cpu] Fix CPUMaxSplat transformation

aturetsk:
Recently sinkRescaleQuantizeNode was extended for Min/Max nodes (https://github.com/pytorch/glow/commit/1e5fdff1c90ea22795641fea52a10c485a09c730). This broke CPUMaxSplat transformation, because now Max nodes can have different quantization scale or offset for input and output, which is not supported by CPUMaxSplat node.
This patch fixes this issue by pulling Rescale node out from Max node, before replacing it with CPUMaxSplat.

### DIFF
--- a/lib/Backends/CPU/Transforms.cpp
+++ b/lib/Backends/CPU/Transforms.cpp
@@ -79,6 +79,40 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
       CN->getBias(), CN->getKernel(), CN->getStride(), CN->getPads(), group));
 }
 
+/// Merge Max and Splat nodes into target-specific CPUMaxSplat node.
+/// For quantized network, sinkRescaleQuantizedNode transformation might have
+/// merged Rescale into Max node. In this case we need to pull it out, since
+/// CPUMaxSplat requires input and output to be quantized the same way.
+static Node *optimizeCPUMaxSplat(MaxNode *MN, Function *F) {
+  SplatNode *splat;
+  NodeValue input;
+
+  // One of the inputs must be Splat.
+  if ((splat = dyn_cast<SplatNode>(MN->getLHS()))) {
+    input = MN->getRHS();
+  } else if ((splat = dyn_cast<SplatNode>(MN->getRHS()))) {
+    input = MN->getLHS();
+  } else {
+    return nullptr;
+  }
+
+  // Pull out Rescale (for quantized types only).
+  if (input.getType() != MN->getResult().getType()) {
+    assert(input.getType()->isQuantizedType() &&
+           MN->getResult().getType()->isQuantizedType() &&
+           "Types should be quantized");
+    auto *RS = F->createRescaleQuantized(MN->getName(), input,
+                                         MN->getResult().getType());
+    input = NodeValue(RS, 0);
+  }
+
+  assert(input.getType() == splat->getResult().getType() &&
+         "Both inputs of Max node must have the same type");
+
+  return F->addNode(
+      new CPUMaxSplatNode(MN->getName(), input, splat->getValue()));
+}
+
 bool CPUBackend::transformPostLowering(Function *F,
                                        CompilationMode mode) const {
   bool changed = false;
@@ -91,17 +125,10 @@ bool CPUBackend::transformPostLowering(Function *F,
         continue;
       }
     }
+
+    // Merge Max and Splat nodes into CPUMaxSplat.
     if (auto *MN = dyn_cast<MaxNode>(&node)) {
-      if (auto *splat = dyn_cast<SplatNode>(MN->getLHS())) {
-        auto MSN = F->addNode(new CPUMaxSplatNode(MN->getName(), MN->getRHS(),
-                                                  splat->getValue()));
-        NodeValue(&node, 0).replaceAllUsesOfWith(MSN);
-        changed = true;
-        continue;
-      }
-      if (auto *splat = dyn_cast<SplatNode>(MN->getRHS())) {
-        auto MSN = F->addNode(new CPUMaxSplatNode(MN->getName(), MN->getLHS(),
-                                                  splat->getValue()));
+      if (Node *MSN = optimizeCPUMaxSplat(MN, F)) {
         NodeValue(&node, 0).replaceAllUsesOfWith(MSN);
         changed = true;
         continue;
--- a/tests/unittests/BackendCorrectnessTest.cpp
+++ b/tests/unittests/BackendCorrectnessTest.cpp
@@ -770,6 +770,19 @@ TEST_P(BackendCorrectnessTest, tinyResnet) {
   EXPECT_TRUE(out1.isEqual(out2, 0.001));
 }
 
+// Test MaxSplat transformation in CPU backend.
+TEST_P(CPUOnly, maxSplatTest) {
+  PseudoRNG PRNG;
+  Tensor input(ElemKind::Int8QTy, {5, 5}, 0.001, -10);
+  input.getHandle<int8_t>().randomize(-128, 127, PRNG);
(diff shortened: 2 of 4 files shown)


jfix71:
Hi @aturetsk, thanks for the fix! Can you write a test in OperatorTest that illustrates the bug that is fixed by the PR? Thanks!

lib/Backends/CPU/Transforms.cpp
```diff
@@ -93,14 +93,30 @@ bool CPUBackend::transformPostLowering(Function *F,
     }
     if (auto *MN = dyn_cast<MaxNode>(&node)) {
       if (auto *splat = dyn_cast<SplatNode>(MN->getLHS())) {
-        auto MSN = F->addNode(new CPUMaxSplatNode(MN->getName(), MN->getRHS(),
+        NodeValue MNInput = MN->getRHS();
```
rdzhabarov: Thanks for fixing this.

is it possible to consolidate the code from both branches in one place?
Seem like it's pretty much the same, once you know which NodeValue is from Splat and which is the other one.

rdzhabarov: also please sign the CLA above :)

aturetsk:
@jfix71, shouldn't the test be in BackendCorrectnessTest (CPUOnly), since I fixed a transformation in CPU-backend?

jfix71:
@aturetsk Oops yeah much better idea 😃

lib/Backends/CPU/Transforms.cpp
```diff
@@ -93,14 +93,30 @@ bool CPUBackend::transformPostLowering(Function *F,
...
+        // Rescale might have been sunk into MaxNode, pull it back.
+        if (MNInput.getType() != MN->getResult().getType()) {
```
rdzhabarov: could you also add an 'assert' before doing a transformation that output of the splat has the same type as the other input to the max operator? (we need to ensure that scale and offset of LHS and RHS are the same for quantized numbers).

Also, you can/should "Rescale" only for quantized operators.

aturetsk:
@jfix71, @nadavrot, @rdzhabarov 
Extracted CPUMaxSplat transformation into a separate function (with description). Consolidated the code from both branches in one place and added asserts, as requested.
Added some comments in transformPostLowering.
Added a test.

lib/Backends/CPU/Transforms.cpp
```diff
@@ -79,29 +79,56 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
       CN->getBias(), CN->getKernel(), CN->getStride(), CN->getPads(), group));
 }
 
+// Merge Max and Splat nodes into target-specific CPUMaxSplat node.
+// For quantized network, sinkRescaleQuantizedNode transformation might have
+// merged Rescale into Max node. In this case we need to pull it out, since
+// CPUMaxSplat requires input and output to be quantized the same way.
```
qcolombet: I don't think this is the right fix, unless I miss something.
Indeed,  as far as I remember, the rescale node is not going to guarantee that its output will have the same offset and scale than the Max node.
E.g., consider
res = max(lhs, rhs)
It is possible to derive different ranges for lhs, rhs and res. In particular, res is likely to have to cover a smaller range than lhs and rhs.
In particular, there is nothing that ties the parameters of the quantization of rescale(lhs) to the parameters of the quantization of res.
This is all theoretical, in the sense that I am not versed into the gory details of our implementation of rescale and quantization in general, but just from the API (and checks in the verifier) it is possible to write something with different scale and offset for each operand (to be fair, we can also put a quantized and non-quantized operand, but that would crash somewhere :P). From the IR perspective, I believe we want this flexibility and the interpreter does the right thing as far as I can tell (dequantize individually each operand and re-quantize the result independently).

In fact, this issue highlights a bigger problem in Glow: what do we do with the legality of each operator?
Right now, each backend has to manually lower the IR to something that works for them (hence the proposed fix, though it wouldn't work I believe), however, it looks like we will want to describe such constraints in a high level way so that they can be fixed by a generic pass.

For instance, I wouldn't be surprised if a lot of backends require that both LHS and RHS have the same quantization parameters for all ALU, right now we cannot express that AFAICT.

@rdzhabarov Am I missing something or is it true that we cannot express something like `a = b + c` must have the same quantization parameters for a, b and c.

rdzhabarov: "Right now, each backend has to manually lower the IR to something that works for them (hence the proposed fix, though it wouldn't work I believe), however, it looks like we will want to describe such constraints in a high-level way so that they can be fixed by a generic pass."
Interpreter, for instance, does not require any specialized handling of the max operator (quantized).
I do not have a good insight into other specific backends but could imagine that there might be ways to set {s,o} per LHS/RHS.

"Am I missing something or is it true that we cannot express something like a = b + c must have the same quantization parameters for a, b and c."
There is no direct way to do this, but in most of the cases, as you pointed out, quantization params would be different for a,b,c.
This specific CPU optimization is applied in order to avoid extra memory allocations for Splat but may introduce additional rescales (some rescales may cancel out with other rescales though). In general, it's not super clear if it's always worth it to enforce this behavior.

lib/Backends/CPU/Transforms.cpp
```diff
@@ -79,29 +79,56 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
...
+static Node *optimizeCPUMaxSplat(MaxNode *MN, Function *F) {
+  SplatNode *splat;
+  NodeValue input;
+
+  // One of the inputs must be Splat.
+  if ((splat = dyn_cast<SplatNode>(MN->getLHS()))) {
+    input = MN->getRHS();
+  } else if ((splat = dyn_cast<SplatNode>(MN->getRHS()))) {
+    input = MN->getLHS();
```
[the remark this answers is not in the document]
aturetsk: Then Max node can be replaced with one (biggest by value) of the splats. But this is not a CPU specific transformation, thus it should not be optimized here. It should be handled in lib/Optimizer/GraphOptimizer.cpp (there are some Splat-related optimizations there), but that's a subject for a separate PR, I think.

lib/Backends/CPU/Transforms.cpp
qcolombet: > This specific CPU optimization is applied in order to avoid extra memory allocations for Splat but may introduce additional rescales (some rescales may cancel out with other rescales though). In general, it's not super clear if it's always worth it to enforce this behavior.

I agree and I thought it was even worse than that, I missed that the rescale node takes the need offset and scale that we want.

If that's just an optimization, then this patch is fine.

lib/Backends/CPU/Transforms.cpp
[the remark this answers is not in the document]
qcolombet: Could you assert on that?

qcolombet:
@aturetsk LGTM too.

lib/Backends/CPU/Transforms.cpp
aturetsk: You have a good point here. At the moment CPUMaxSplat requires input and output to be quantized the same way. It is so at the moment, but in the future this may change (as you described, the output of max is likely to have a smaller range and may be quantized differently in order to achieve better accuracy). So putting an assert here which checks that input type is equal to output type before creating CPUMaxSplat node maybe a good idea, to prevent people from debugging issues with it when/if quantization changes.

lib/Backends/CPU/Transforms.cpp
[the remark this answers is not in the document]
qcolombet: Oh sorry my bad. For some reasons, I thought we had to do this for all the splats feeding the max node, whereas this is just an optimization so if we do only one of them this is fine.

PR #1397, opened 2018-08-03
Add support for more operators to the model loader

tlepley-cadence:
Using some ONNX models, we figured out that support for some operators were missing in the loader:
- sqrt
- reciprocal
- min
- max

The IR nodes already existed so only the loader has been extended.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -79,6 +79,23 @@ protected:
     addNodeAsOutput(op, shape);
   }
 
+  /// Loads Sqrt operator, given its protobuf representation and parsed args.
+  void loadSqrt(const OpType &op, ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    auto *R = G_.createPow(opName, in, 0.5f);
+    addNodeAsOutput(op, R);
+  }
+
+  /// Loads Reciprocal operator, given its protobuf representation and parsed
+  /// args.
+  void loadReciprocal(const OpType &op, ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    auto *R = G_.createPow(opName, in, -1.0f);
+    addNodeAsOutput(op, R);
+  }
+
   void loadSum(const OpType &op, ArgumentDictionaryTy &dict) {
     // TODO: support variadic arguments
     assert(op.input_size() == 2 && "Only Sum of 2 inputs is supported.");
@@ -352,6 +369,14 @@ protected:
     }
     if (typeName == "Shape") {
       loadShape(op, dict);
+      return true;
+    }
+    if (typeName == "Sqrt") {
+      loadSqrt(op, dict);
+      return true;
+    }
+    if (typeName == "Reciprocal") {
+      loadReciprocal(op, dict);
       return true;
     }
     if (typeName == "Sum") {
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -152,6 +152,24 @@ protected:
     nodeValueByName_[op.output(0)] = NodeValue(N, 0);
   }
 
+  void loadMinMax(llvm::StringRef typeName, const OpType &op,
+                  ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in0 = getNodeValueOrCreateVariableByName(op.input(0));
+    auto in1 = getNodeValueOrCreateVariableByName(op.input(1));
+
+    Node *node = nullptr;
+    if (typeName == "Min") {
+      node = G_.createMin(opName, in0, in1);
+    } else if (typeName == "Max") {
+      node = G_.createMax(opName, in0, in1);
+    } else {
+      llvm_unreachable("Invalid min or max operator");
+    }
+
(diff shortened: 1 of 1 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -79,6 +79,22 @@ class CommonOperatorLoader : public ProtobufLoader {
...
 
+  /// Loads Sqrt operator, given its protobuf representation and parsed args.
+  void loadSqrt(const OpType &op, ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    auto *R = G_.createSqrt(opName, in);
+    addNodeAsOutput(op, R);
+  }
+
+  /// Loads Reciprocal operator, given its protobuf representation and parsed args.
```
jfix71: Please run `clang-format`

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -135,6 +151,23 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+                      ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in0 = getNodeValueOrCreateVariableByName(op.input(0));
+    auto in1 = getNodeValueOrCreateVariableByName(op.input(1));
+
+    Node *node = nullptr;
+    if (typeName == "Min") {
+      node = G_.createMin(opName, in0, in1);
+    }
+    else if (typeName == "Max") {
```
jfix71: Make this just `else`, as it must be `"Max"` if the function is called correctly. Otherwise this gives the impression `node = nullptr` is a possibility.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -165,7 +198,9 @@ class CommonOperatorLoader : public ProtobufLoader {
       node = G_.createSub(opName, in0, finalIn1);
     } else if (typeName == "Div") {
       node = G_.createDiv(opName, in0, finalIn1);
-    } else {
+    }
```
jfix71: Please undo this spacing change -- clang-format should fix it.

lib/Graph/Graph.cpp
```diff
@@ -1021,6 +1021,14 @@ PowNode *Function::createPow(llvm::StringRef name, NodeValue Base, float exp) {
   return addNode(new PowNode(name, Base.getType(), Base, exp));
 }
 
+PowNode *Function::createSqrt(llvm::StringRef name, NodeValue Base) {
+  return addNode(new PowNode(name, Base.getType(), Base, 0.5f));
```
jfix71: nit: here and in `createReciprocal()`, I would just directly do `return createPow(name, base, 0.5f)`;

tlepley-cadence: I simplified further. Since there is no `reciprocal` and `sqrt` nodes in the IR, I now simply call `createPow` from the loader.

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: Done

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: I added a `else` statement with an `assert`. I think this is the safest approach.

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: Done

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -135,6 +152,24 @@ class CommonOperatorLoader : public ProtobufLoader {
     nodeValueByName_[op.output(0)] = NodeValue(N, 0);
   }
 
+  void loadMinMax(llvm::StringRef typeName, const OpType &op,
+                  ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    auto in0 = getNodeValueOrCreateVariableByName(op.input(0));
```
rdzhabarov: nit: lhs and rhs for inputs.

nadavrot:
Nice!

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: I simply reused the naming I found in other functions. For instance `loadArithmetic` uses `in0` and 'in1'. I didn't see any usage of `lhs` and `rhs` in this loader file.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -135,6 +152,24 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+    auto in1 = getNodeValueOrCreateVariableByName(op.input(1));
+
+    Node *node = nullptr;
+    if (typeName == "Min") {
+      node = G_.createMin(opName, in0, in1);
+    } else if (typeName == "Max") {
+      node = G_.createMax(opName, in0, in1);
+    } else {
+      assert(false && "Invalid min or max operator");
```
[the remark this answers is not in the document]
tlepley-cadence: Done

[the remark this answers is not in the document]
tlepley-cadence: For @rdzhabarov 's comment, I'm not sure to understand the point since I didn't see any reference to rhs and lhs in the importer that I extended

include/glow/Importer/CommonOperatorLoader.h
jfix71: Ok, I'll just merge this as is, and if we want to make them all consistent we can do so in a future PR.

PR #1398, opened 2018-08-03
Add cmake options to disable examples and/or tests build.

tlepley-cadence:
The link with LLVM consumes a substantial amount of resources (cpu time, memory), and I find very convenient to have an option to disable the build of tests and examples when developing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ enable_testing()
 
 option(GLOW_WITH_CPU "Build the LLVM-based JIT CPU backend" ON)
 option(GLOW_WITH_OPENCL "Build the OpenCL backend" ON)
+option(GLOW_BUILD_EXAMPLES "Build the examples" ON)
+option(GLOW_BUILD_TESTS "Build the tests" ON)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -77,12 +79,15 @@ if(NOT EXISTS "${GLOW_THIRDPARTY_DIR}/onnx")
...
-
 add_subdirectory(lib)
-add_subdirectory(examples)
 add_subdirectory(tools)
-add_subdirectory(tests)
+if (GLOW_BUILD_EXAMPLES)
+  add_subdirectory(examples)
+endif()
+if (GLOW_BUILD_TESTS)
+  add_subdirectory(tests/googletest/)
```
compnerd: Please drop the trailing `/`.

tlepley-cadence: @compnerd done

nadavrot:
@tlepley-cadence  @compnerd  Nice!

CMakeLists.txt
```diff
@@ -77,12 +79,15 @@ if(NOT EXISTS "${GLOW_THIRDPARTY_DIR}/onnx")
...
-
 add_subdirectory(lib)
-add_subdirectory(examples)
 add_subdirectory(tools)
-add_subdirectory(tests)
+if (GLOW_BUILD_EXAMPLES)
+  add_subdirectory(examples)
+endif()
+if (GLOW_BUILD_TESTS)
+  add_subdirectory(tests/googletest)
```
rdzhabarov: there are couple of things at the very end:
```
 # Fetch the dependencies for all the tests.
 get_property(GLOW_TEST_DEPENDS GLOBAL PROPERTY GLOW_TEST_DEPENDS)
 
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND}
                     DEPENDS ${GLOW_TEST_DEPENDS} USES_TERMINAL)
```
Just move it here behind the if (GLOW_BUILD_TESTS).

CMakeLists.txt
```diff
@@ -77,12 +79,15 @@ if(NOT EXISTS "${GLOW_THIRDPARTY_DIR}/onnx")
   message(FATAL_ERROR "No onnx git submodule. Run: git submodule update --init --recursive")
 endif()
 
-add_subdirectory(tests/googletest/)
-
 add_subdirectory(lib)
-add_subdirectory(examples)
 add_subdirectory(tools)
-add_subdirectory(tests)
+if (GLOW_BUILD_EXAMPLES)
```
rdzhabarov: this is also only needed for build examples: find_package(PNG). Would be good just group those things together.

tlepley-cadence: I think the PNG package is also needed for the tools (image classifier), so we may need to keep it outside GLOW_BUILD_EXAMPLES

rdzhabarov: Indeed, it's used in image classifier. Disregard my previous comment :)

CMakeLists.txt
tlepley-cadence: Done

PR #1400, opened 2018-08-03
-1 and 0 values correctly handled by Reshape when the shape is given as a constant tensor

tlepley-cadence:
With the Reshape operator, the output shape information can be passed as protobuf attribute, but also as input tensor. The protobuf attribute case was complete, but the management of the input input tensor giving the output shape was incomplete (shape dimension with value 0 and -1) was not handled.
This PR fixes this issue.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -192,38 +192,57 @@ protected:
 
   void loadReshape(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
-    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    NodeValue in = getNodeValueOrCreateVariableByName(op.input(0));
 
-    std::vector<size_t> newDim;
-    if (dict.count("shape")) {
-      std::vector<int64_t> protoDims = getShape<int64_t>(dict["shape"]);
-
-      auto oldDim = in->dims(0);
-      int64_t product = 1;
-      for (size_t i = 0, e = protoDims.size(); i != e; i++) {
-        if (protoDims[i] == 0)
-          // shape[i] == 0 means that corresponding element should remain
-          // the same.
-          protoDims[i] = oldDim[i];
-        if (protoDims[i] != -1)
-          product *= protoDims[i];
+    // Get the requested shape from the model.
+    // First look at input tensors, then at the "shape" attribute.
+    std::vector<int64_t> requestedDims;
+    if (op.input_size() > 1) {
+      Tensor *constShapeTensor = getTensorByName(op.input(1));
+      auto TH = constShapeTensor->getHandle<size_t>();
+      for (size_t i = 0, e = constShapeTensor->size(); i != e; i++) {
+        requestedDims.push_back(TH.at({i}));
       }
+    } else if (dict.count("shape")) {
+      std::vector<int64_t> protoDims = getShape<int64_t>(dict["shape"]);
       for (size_t i = 0, e = protoDims.size(); i != e; i++) {
-        if (protoDims[i] == -1)
-          // shape[i] == -1 means that corresponding element should be inferred
-          // from all other elements, so that Tensor size remains the same.
-          protoDims[i] = in->getType(0)->size() / product;
-        newDim.push_back(protoDims[i]);
+        requestedDims.push_back(protoDims[i]);
       }
     } else {
-      Tensor *T = getTensorByName(op.input(1));
-      auto TH = T->getHandle<size_t>();
-      for (size_t i = 0, e = T->size(); i != e; i++) {
-        newDim.push_back(TH.raw(i));
-      }
+      assert(false &&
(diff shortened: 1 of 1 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -218,8 +218,31 @@ class CommonOperatorLoader : public ProtobufLoader {
     } else {
       Tensor *T = getTensorByName(op.input(1));
       auto TH = T->getHandle<size_t>();
+
+      std::vector<int64_t> protoDims(T->size());
+      auto oldDim = in->dims(0);
+      int64_t product = 1;
       for (size_t i = 0, e = T->size(); i != e; i++) {
-        newDim.push_back(TH.raw(i));
+        int64_t new_shape = TH.raw(i);
```
jfix71: Can we use `TH.at({i})` here?

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -218,8 +218,31 @@ class CommonOperatorLoader : public ProtobufLoader {
     } else {
       Tensor *T = getTensorByName(op.input(1));
       auto TH = T->getHandle<size_t>();
+
```
jfix71: Can we add an assertion here that the shape of T is `{in.dims()}`?

include/glow/Importer/CommonOperatorLoader.h
jfix71: nit: prefer `newShape` over `new_shape`

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -218,8 +218,31 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+      std::vector<int64_t> protoDims(T->size());
+      auto oldDim = in->dims(0);
```
jfix71: `in->dims(0)` gets the `dims` of the first output of `in`. Instead, use `in.dims()` here, which will use whatever output of `in` is being referred to here, since `in` is a `NodeValue`.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -218,8 +218,31 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+
+        if (new_shape != -1) {
+          product *= new_shape;
+        }
+      }
+      for (size_t i = 0, e = protoDims.size(); i != e; i++) {
+        if (protoDims[i] == -1)
+          // shape[i] == -1 means that corresponding element should be inferred
+          // from all other elements, so that Tensor size remains the same.
+          protoDims[i] = in->getType(0)->size() / product;
```
jfix71: Same as above -- use `in.getType()` to ensure you're getting the correct output of the `NodeValue`.

nadavrot:
@tlepley-cadence Thanks!

tlepley-cadence:
@jfix71 Sure, I'll rework the PR. I agree that avoiding code duplication is better. I had also noticed some other issues I had planed to fix later, but it's now a good opportunity to do it. For instance, the 'shape' attribute should only be looked if the second input tensor is not present (it's explicitly described in caffe2, I need to check for ONNX). Also, there should be at most one -1 value in the new shape. We should also check that the tensor is a constant tensor with a value, but one of your comments may already cover this.

jfix71:
Sounds great, thanks!

RE: if the tensor is constant/valued -- `getTensorByName()` only returns weight Variables that were loaded from the init_net. I think we should change the naming there to make it clear, but that's for another PR.

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: I don't think we can assume that the number of elements in T will be the same as the number of dimensions in 'in'. The description in ONNX and Caffe2 don't request this. For instance, we may want to reshape [1, 1, 1, 200] into [200], like does the Squeeze operator.

jfix71: Oops yeah you're right, never mind 😄

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -192,38 +192,61 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+      if (newDim != -1) {
+        dimProduct *= newDim;
+        outputDims.push_back(newDim);
+      } else {
+        // The -1 case is handled later.
+        outputDims.push_back(0);
+      }
+    }
+    bool gotMinusOne = false;
+    for (size_t i = 0, e = outputDims.size(); i != e; i++) {
```
jfix71: Looking great! One thing -- `loadReshape()` is getting a bit long. I think it is possible to replace this last loop here, which would make the function a bit shorter and more efficient.

E.g. you could introduce something like `ssize_t negOneDim = -1` before the previous loop, and then set it inside the loop inside `if (newDim != -1)`. And then just replace this last loop with directly setting `outputDims[negOneDim]`.

tlepley-cadence:
@jfix71 I think I addressed all your comments

PR #1401, opened 2018-08-03
Check the consistency between  image and declared ONNX input

tlepley-cadence:
In the image classifier the input image shape is not checked with regard to the ONNX input declaration.

In case of input image shape mistake, the wrong image shape is propagated accross the model and some indirect errors are usually generated (for instance for a convolution operator inside the model). 

Since the ONNX model declares inputs, it is possible to check the input image shape before any shape propagation and generate an accurate error message. This is what does this PR.

### DIFF
--- a/include/glow/Importer/ONNX.h
+++ b/include/glow/Importer/ONNX.h
@@ -89,6 +89,13 @@ class ONNXModelLoader
   size_t opsetVersion_;
 
 public:
+  /// Checks that the inputs tensors are compatible with the inputs declared in
+  /// the ONNX model. The input tensors in \p tensors are stored with the names
+  /// in the list of names \p tensorNames.
+  void checkInputs(onnx::GraphProto &net,
+                   llvm::ArrayRef<const char *> tensorNames,
+                   llvm::ArrayRef<Tensor *> tensors);
+
   /// Loads the ONNX model that's represented by a model description file,
   /// serialized in \p modelDescFilename and populates the network into \p F.
   /// The tensors in \p tensors are stored with the names in the list of names
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -507,6 +507,35 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
 ONNXModelLoader::ONNXModelLoader(Function &F)
     : CommonOperatorLoader({}, {}, F) {}
 
+void ONNXModelLoader::checkInputs(onnx::GraphProto &net,
+                                  llvm::ArrayRef<const char *> tensorNames,
+                                  llvm::ArrayRef<Tensor *> tensors) {
+  for (size_t i = 0; i < tensorNames.size(); i++) {
+    // Look if a corresponding input exists.
+    for (int j = 0; j < net.input_size(); j++) {
+      const onnx::ValueInfoProto &valueInfo = net.input(j);
+      const std::string &inputName = valueInfo.name();
+
+      if (inputName != tensorNames[i]) {
+        continue;
+      }
+
+      llvm::ArrayRef<size_t> dims = tensors[i]->dims();
+      const onnx::TensorShapeProto &shape =
+          valueInfo.type().tensor_type().shape();
+
+      // Check if the number of dimensions is consistent.
+      assert(dims.size() == shape.dim_size() &&
+             "Mismatch between input image and ONNX input shape");
+      // Allow batch dimensions to be different.
+      for (size_t k = 1; k < dims.size(); k++) {
+        assert(dims[k] == shape.dim(k).dim_value() &&
+               "Mismatch between input image and ONNX input shape");
+      }
+    }
+  }
+}
+
 ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
                                  llvm::ArrayRef<const char *> tensorNames,
                                  llvm::ArrayRef<Tensor *> tensors, Function &F)
@@ -516,6 +545,8 @@ ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
   if (!loadProto(modelDef, modelDescFilename)) {
     GLOW_ASSERT("Cannot load the network.");
   }
+
+  checkInputs(modelDef, tensorNames, tensors);
 
   loadInitializers(modelDef);
   if (loadNetwork(modelDef)) {


lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
 ONNXModelLoader::ONNXModelLoader(Function &F)
     : CommonOperatorLoader({}, {}, F) {}
 
+void ONNXModelLoader::CheckIO(onnx::GraphProto &net,
+                              llvm::ArrayRef<const char *> names,
+                              llvm::ArrayRef<Tensor *> tensors)
```
nadavrot: Clang format please.

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+{
+  for (unsigned i = 0; i < names.size(); i++) {
+    // Look if a corresponding input exists
```
nadavrot: Period at the end of the sentence.

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+    for(int j = 0; j < net.input_size(); j++) {
+      const onnx::ValueInfoProto& value_info = net.input(j);
+      const std::string& input_name = value_info.name();
+
+      if (input_name == names[i]) {
+        llvm::ArrayRef<size_t> dims = tensors[i]->dims();
+        const onnx::TensorShapeProto& shape = value_info.type().tensor_type().shape();
+
+        // Check
```
nadavrot: Check what?

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+        if (dims.size() != shape.dim_size()) {
+          llvm::outs().flush();
+          llvm_unreachable("Mismatch between input image and ONNX input shape");
+        }
+        // Allow batch dimensions to be different.
+        for(unsigned k = 1; k < dims.size(); k++) {
+          if (dims[k] != shape.dim(k).dim_value()) {
+            llvm::outs().flush();
```
nadavrot: Why do we need the flush?

include/glow/Importer/ONNX.h
```diff
@@ -89,6 +89,10 @@ class ONNXModelLoader
   size_t opsetVersion_;
 
 public:
+  void CheckIO(onnx::GraphProto &net,
```
nadavrot: Doxygen comment please.

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
 
+void ONNXModelLoader::CheckIO(onnx::GraphProto &net,
+                              llvm::ArrayRef<const char *> names,
+                              llvm::ArrayRef<Tensor *> tensors)
+{
+  for (unsigned i = 0; i < names.size(); i++) {
+    // Look if a corresponding input exists
+    for(int j = 0; j < net.input_size(); j++) {
+      const onnx::ValueInfoProto& value_info = net.input(j);
+      const std::string& input_name = value_info.name();
```
jfix71: nit: prefer `inputName` over `input_name`

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
     : CommonOperatorLoader({}, {}, F) {}
 
+void ONNXModelLoader::CheckIO(onnx::GraphProto &net,
+                              llvm::ArrayRef<const char *> names,
+                              llvm::ArrayRef<Tensor *> tensors)
+{
+  for (unsigned i = 0; i < names.size(); i++) {
+    // Look if a corresponding input exists
+    for(int j = 0; j < net.input_size(); j++) {
+      const onnx::ValueInfoProto& value_info = net.input(j);
```
jfix71: nit: prefer `valueInfo` over `value_info`

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+      if (input_name == names[i]) {
+        llvm::ArrayRef<size_t> dims = tensors[i]->dims();
+        const onnx::TensorShapeProto& shape = value_info.type().tensor_type().shape();
+
+        // Check
+        if (dims.size() != shape.dim_size()) {
+          llvm::outs().flush();
+          llvm_unreachable("Mismatch between input image and ONNX input shape");
+        }
+        // Allow batch dimensions to be different.
```
jfix71: I'm not sure we want to allow batch dimensions to be different... Also, it's not necessarily the case that we're loading a proto with batched input -- the loader is supposed to load any proto of any domain.

include/glow/Importer/ONNX.h
tlepley-cadence: @jfix71 It only checks the input shape, so checkInput() would be more accurate.

lib/Importer/ONNX.cpp
tlepley-cadence: Models are usually flexible with regard the batch dimension. Currently, when we convert a TensorFlow model into an ONNX model, we get a model specified for a batch of 1. But this model would also work in Glow with a different batch value. This is why I don't think we should check the input batch dimension for now .
- Tensorflow: input shape=(?, 224, 224, 3)
- ONNX:  input shape=(1, 224, 224, 3)

lib/Importer/ONNX.cpp
tlepley-cadence: Done.

lib/Importer/ONNX.cpp
tlepley-cadence: Done.

lib/Importer/ONNX.cpp
tlepley-cadence: Improved the comment.

lib/Importer/ONNX.cpp
tlepley-cadence: Removed. It was not supposed to be here.

include/glow/Importer/ONNX.h
tlepley-cadence: Renamed : checkInputs()

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
 ONNXModelLoader::ONNXModelLoader(Function &F)
     : CommonOperatorLoader({}, {}, F) {}
 
+void ONNXModelLoader::CheckIO(onnx::GraphProto &net,
+                              llvm::ArrayRef<const char *> names,
+                              llvm::ArrayRef<Tensor *> tensors)
+{
+  for (unsigned i = 0; i < names.size(); i++) {
```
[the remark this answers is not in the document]
tlepley-cadence: Done

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,37 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+        llvm::ArrayRef<size_t> dims = tensors[i]->dims();
+        const onnx::TensorShapeProto& shape = value_info.type().tensor_type().shape();
+
+        // Check
+        if (dims.size() != shape.dim_size()) {
+          llvm::outs().flush();
+          llvm_unreachable("Mismatch between input image and ONNX input shape");
+        }
+        // Allow batch dimensions to be different.
+        for(unsigned k = 1; k < dims.size(); k++) {
```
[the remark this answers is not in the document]
tlepley-cadence: Done.

lib/Importer/ONNX.cpp
tlepley-cadence: Done

lib/Importer/ONNX.cpp
tlepley-cadence: Done

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,36 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+        // Check if the number of dimensions is consistent.
+        if (dims.size() != shape.dim_size()) {
+          llvm_unreachable("Mismatch between input image and ONNX input shape");
+        }
+        // Allow batch dimensions to be different.
+        for (size_t k = 1; k < dims.size(); k++) {
+          if (dims[k] != shape.dim(k).dim_value()) {
+            llvm_unreachable(
+                "Mismatch between input image and ONNX input shape");
+          }
```
nadavrot: This code can be simplified. Instead of increasing the nest level please use early exits. 

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,36 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+      const onnx::ValueInfoProto &valueInfo = net.input(j);
+      const std::string &inputName = valueInfo.name();
+
+      if (inputName == names[i]) {
+        llvm::ArrayRef<size_t> dims = tensors[i]->dims();
+        const onnx::TensorShapeProto &shape =
+            valueInfo.type().tensor_type().shape();
+
+        // Check if the number of dimensions is consistent.
+        if (dims.size() != shape.dim_size()) {
```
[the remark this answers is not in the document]
tlepley-cadence: Done.

lib/Importer/ONNX.cpp
tlepley-cadence: Removed one nesting level (replaced by assert). I assume this is what you meant.

lib/Importer/ONNX.cpp
```diff
@@ -502,6 +502,33 @@ std::unique_ptr<ONNXModelLoader> ONNXModelLoader::parse(const void *onnxModel,
...
+void ONNXModelLoader::checkInputs(onnx::GraphProto &net,
+                                  llvm::ArrayRef<const char *> names,
+                                  llvm::ArrayRef<Tensor *> tensors) {
+  for (size_t i = 0; i < names.size(); i++) {
+    // Look if a corresponding input exists.
+    for (int j = 0; j < net.input_size(); j++) {
+      const onnx::ValueInfoProto &valueInfo = net.input(j);
+      const std::string &inputName = valueInfo.name();
+
+      if (inputName == names[i]) {
```
jfix71: Related to the nest level as mentioned by @nadavrot, this `if` could be replaced by:
```
if (inputName != names[i]) {
  continue;
}
```
And then the current body of the `if` follows this new `if`, with one fewer nest level.

tlepley-cadence: OK, got it. Done.

include/glow/Importer/ONNX.h
```diff
@@ -89,6 +89,12 @@ class ONNXModelLoader
   size_t opsetVersion_;
 
 public:
+  /// Checks that the inputs tensors are compatible with the inputs declared in
+  /// the ONNX model. The input tensors in \p tensors are stored with the names
+  /// in the list of names \p tensorNames.
+  void checkInputs(onnx::GraphProto &net, llvm::ArrayRef<const char *> names,
```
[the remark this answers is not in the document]
tlepley-cadence: Done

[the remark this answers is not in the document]
jfix71: It's `tensorNames` in the comment above and in ONNX.cpp, but still `names` here in the function declaration.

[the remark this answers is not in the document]
tlepley-cadence: Missed this one :) Done.

lib/Importer/ONNX.cpp
jfix71: What if we're loading a proto that does not have batched input? Or, what if the model is very weird and wants to batch input on the last dimension instead of the first dimension? It might not be an image classification model that's using this class -- e.g. it could be text translation, or anything else. So I don't think always ignoring the first dimension is what we want.

Instead, I'm wondering if perhaps it would make sense to add another argument to the `ONNXModelLoader()` constructor: an `ArrayRef<size_t> allowedDifferentDims`, which is passed into `checkInputs()`, which is used to know when it's OK to skip verifying that a dimension has the same size as in the proto. This way, whatever is creating the `ONNXModelLoader` (e.g. `ImageClassifier`) can pass in dimensions that it knows don't need to match.

What do you think?

tlepley-cadence: I see your point: `checkInputs()` is currently in the generic ONNXModelLoader while it's code is actually more specific to `ImageClassifier`. Since the model can have multiple inputs with different numbers of dimensions and since batching is the only use-case I currently see, we can indeed pass an array of `size_t `to express that at most one dimension can be ignore per input tensor during the check.

jfix71: I'm realizing since we don't know the order of the inputs, it probably wouldn't work to just pass the array of size_t -- you can't tell what inputs go with what dimensions. Perhaps it could be an array of pairs with the name of the input and the ignored dimension? So that you can associate different inputs by name with what dimension is OK to be different.

PR #1403, opened 2018-08-03
[unittests] Add the first unnitest for Caffe2 importer.

beicy:


tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+/// Test loading conv op from a Caffe2 model.
+/// The input is N*C*H*W (1*1*3*3), the kernel is 2,
+/// stride is 1, pad is 1, group is 1.
+TEST(caffe2, importConv) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename("tests/models/caffe2Models/predict_net.pbtxt");
```
nadavrot: Would it simplify things if we include the protobuf files as inline text?

rdzhabarov:
This work is awesome! Would be cool to have ONNX loader tests in the follow-up PR.

beicy:
@rdzhabarov Thanks! Yes, ONNX loader tests will be added to our repo as well.

artemrakhov:
Great start! Instead of introducing `isFileName`, could we have proper pbtxt files inside the repo, so that the loader would always read files?

beicy:
@artemrakhov  Actually, the first version was having loader to read the pbtxt files. However, there is one issue related to the path of pbtxt files.  For "ninja test", we can set the "WORKING_DIRECTORY" in CMakeLists file to make sure it is run under "tests/unittests". However, users can run the test unit alone (i.e. ./tests/caffe2ImporterTest) under any file directory, and we need extra code to manipulate the path of the models rather than simply writing some relative path in the source code. Therefore, I prefer avoiding reading files in our repo.   Btw, the inline text is just copied from the pbtxt file :)

artemrakhov:
Does it simplify things, if we assume that test is always ran from "root" (i.e. using `./tests/caffe2ImporterTest`) ? I think it is reasonable requirement, and we do have it for `run.sh` and all examples (they assume that data is in root).

artemrakhov:
I'd prefer to split test data from test code.

beicy:
@artemrakhov I see. Yes, it will be much easier if we assume that test is always ran from "root". Thanks!

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+  std::string NetWeightFilename("tests/models/caffe2Models/init_net.pbtxt");
+
+  // Make sure to destroy loader as soon as Graph is loaded.
+  {
+    Tensor data;
+    getImageData(&data, 1, 1, 3, 3);
+    caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"data"},
+                               {&data}, *F);
+  }
+  EE.compile(CompilationMode::Infer, F);
```
artemrakhov: Let's actually run inference and check that output is what we expect it to be.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/IR/IR.h"
+#include "glow/Importer/Caffe2.h"
+#include "gtest/gtest.h"
+
+using namespace glow;
+
+void getImageData(Tensor *result, size_t n, size_t c, size_t h, size_t w) {
```
rdzhabarov: static or in the anonymous namespace.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/IR/IR.h"
```
rdzhabarov: why do we need IR here?

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+/// stride is 1, pad is 1, group is 1.
+TEST(caffe2, importConv) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename("tests/models/caffe2Models/predict_net.pbtxt");
+  std::string NetWeightFilename("tests/models/caffe2Models/init_net.pbtxt");
+
+  // Make sure to destroy loader as soon as Graph is loaded.
```
rdzhabarov: can you expand this comment as to why we are doing so.

tests/models/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Copy the models to the build directory.
+
+file(GLOB files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} caffe2Models/*.pbtxt)
```
rdzhabarov: how do you plan to keep more test models under caffe2Models folder?

beicy: In general, I would like to 1) group ops; 2) try to share the weights file. 
After the conv op testing is done, I would like to create small networks starting from a conv layer for testing (i.e based on some real models but reduce the layers). Therefore, the weights file can be shared. When an new op is ready for testing, if it is proper, it can be added into the existing models as a layer. Otherwise, a new model file should be created.

tests/unittests/caffe2ImporterTest.cpp
beicy: No, we don't need IR here :)

tests/unittests/caffe2ImporterTest.cpp
beicy: Our image classifier and previous testing sample works in this way. Since the loader is not used anymore, it is reasonable to to free the memory.

nadavrot:
Nice!

tests/models/caffe2Models/predict_net.pbtxt
```diff
@@ -0,0 +1,25 @@
+name: "predict"
```
artemrakhov: You probably want to name this file differently, because future tests will add more predict_nets.

artemrakhov: What I meant here was renaming the file itself (e.g. `predict_net.pbtxt` -> `conv.pbtxt`), not changing name field of the proto. The name field is probably not important, because it's not used anywhere.

jfix71:
@beicy I just noticed the commit message is just "Importer" -- please make sure future commits have a descriptive message explaining the change. Thanks!

beicy:
@jfix71 Yes, I noticed it after the PR has been pushed. I just checked the PR title... I will make sure to double check the description in the future. Thanks!

PR #1405, opened 2018-08-03
Support ConstantFill, GivenTensorIntFill, GivenTensorInt64Fill in predict_net

jfix71:
Refactored `loadWeights()` to load each operator via `loadWeight()`, and then call it from the `loadNetwork()` loop when encountering ConstantFill, GivenTensorIntFill, or GivenTensorInt64Fill.

### DIFF
--- a/include/glow/Importer/Caffe2.h
+++ b/include/glow/Importer/Caffe2.h
@@ -45,6 +45,9 @@ class caffe2ModelLoader
   /// \p tensors.
   void loadWeights(caffe2::NetDef &net);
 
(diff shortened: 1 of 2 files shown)


include/glow/Importer/Caffe2.h
```diff
@@ -45,6 +45,9 @@ class caffe2ModelLoader
   /// \p tensors.
   void loadWeights(caffe2::NetDef &net);
 
+  // Loads an individual weight \p op.
```
rdzhabarov: nit: ///

PR #1408, opened 2018-08-04
[build] Fix compilation warning. NFC

artemrakhov:


bertmaher:
Does this happen to solve the `-Warray-bound` warning, because if so, that's awesome.

artemrakhov:
Yes, this does fix `-Warray-bound`.

PR #1411, opened 2018-08-06
More assorted support in Caffe2/CommonOperatorLoader

jfix71:
All related to seq2seq support.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -72,11 +72,14 @@ protected:
   void loadShape(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     auto in = getNodeValueOrCreateVariableByName(op.input(0));
-    auto *shape =
-        G_.getParent()->createVariable(ElemKind::IndexTy, {in.dims().size()},
-                                       opName, VisibilityKind::Private, false);
-    shape->template getHandle<size_t>() = in.dims();
-    addNodeAsOutput(op, shape);
+
+    // This is statically known data, and so we create a Tensor for it and
+    // register it in tensors_.
+    auto *T = new Tensor(ElemKind::IndexTy, {in.dims().size()});
+    tensors_[opName] = T;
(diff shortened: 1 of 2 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -195,19 +198,20 @@ class CommonOperatorLoader : public ProtobufLoader {
     NodeValue in = getNodeValueOrCreateVariableByName(op.input(0));
 
     // Get the requested shape from the model.
-    // First look at input tensors, then at the "shape" attribute.
+    // First look for the "shape" argument, otherwise search for an input tensor
+    // with the shape.
     std::vector<int64_t> requestedDims;
-    if (op.input_size() > 1) {
+    if (dict.count("shape")) {
+      std::vector<int64_t> protoDims = getShape<int64_t>(dict["shape"]);
```
artemrakhov: Why isn't it a mistake to have both `shape` and `op.input(1)` present?

jfix71: Ah, this was actually occurring only because of my hand-hacked proto for seq2seq. You're right, it's an error. I replaced that commit with one that asserts it doesn't happen.

jfix71:
@artemrakhov I'm running into that issue exactly right now. I think using `Int32QTy` is a hacky workaround. Instead I'm working on moving `IndexTy` to use `uint64_t`, and then will rename it to `UInt64ITy`, and then add `Int64ITy` which uses `int64_t`.

artemrakhov:
Alternative solution could be to store shapes in different map: so we'll have separate `tensors_` and `shapes_`. Strictly speaking, shape is not a tensor. It's just an array that we need on construction phase, and never need on IR phase.

jfix71:
The problem with that is we can't know how a tensor is going to be used on the initial loading/building of the graph from the protos, so we don't know which are shapes and which are tensors. We would need to know this as we build the graph. So I think we would need to parse the protos twice -- a first pass to just determine which tensors are only used as shapes, and then a second pass to actually load the tensors and shapes to the correct maps and build the graph.

PR #1415, opened 2018-08-07
[quantization]: Support div node for quantization

ZchiPitt:
Don't know why Div node was not supported for quantization when given the profiling scale and offset.  But some network do need support for div node.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -226,10 +226,11 @@ static Node *quantizeNode(Function *F, Node *node,
     break;                                                                     \
   }
     CASE_QUANTIZE_NODE(Add);
-    CASE_QUANTIZE_NODE(Mul);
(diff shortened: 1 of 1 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -226,10 +226,11 @@ static Node *quantizeNode(Function *F, Node *node,
     break;                                                                     \
   }
     CASE_QUANTIZE_NODE(Add);
-    CASE_QUANTIZE_NODE(Mul);
-    CASE_QUANTIZE_NODE(Sub);
+    CASE_QUANTIZE_NODE(Div);
```
rdzhabarov: LGTM. Can you run some networks containing "div" through the quantization procedure?

ZchiPitt: Didn't break with densenet with cpu and opencl

PR #1416, opened 2018-08-07
Support IndexTy for Div and Gather

jfix71:
More seq2seq support, to get running on the Interpreter and CPU backends. Note that gather for the interpreter was already supported, as its implementation works for any type.


jfix71:
It's somewhere in the decoder of seq2seq. It's hard for me to tell exactly where from in the decoder, because I'm working with a fully decomposed proto (e.g. I don't see LSTMs in the decoder, just their components).

qcolombet:
Could you add a test for the added support in OperatorTest?

PR #1417, opened 2018-08-07
[ProtobufLoader] Set vars created from protos to default untrainable

jfix71:
Variables loaded from protos should not be trainable by default. This improves constant deduplication.


jfix71:
Sure, makes sense to me, I removed it entirely.

PR #1419, opened 2018-08-07
[onnxifi] Adding more functionality to the onnxifi prototype.

rdzhabarov:
* Support init Graph and add more mock functions to support end2end onnxifi run;
* Move onnxifi model loading logic to a separate class.

Note, I was running through the toy example which does processing of conv node (has a dependency on C2). Integration test for this code (C2 independent) is being worked on onnx side (once it's done, I'll integrate it here).

cc: @Maratyszcza @yinghai

### DIFF
--- a/include/glow/Importer/ONNXIFILoader.h
+++ b/include/glow/Importer/ONNXIFILoader.h
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef GLOW_IMPORTER_ONNXIFILOADER_H
+#define GLOW_IMPORTER_ONNXIFILOADER_H
+
+#include "onnx/onnxifi.h"
+
+#include "glow/Importer/ONNX.h"
+
+namespace glow {
+namespace onnxifi {
+
+class ModelLoader : public ONNXModelLoader {
+private:
+  ModelLoader(Function &F) : ONNXModelLoader(F) {}
+
+  /// Load the inputs from the GraphProto. This is useful when the
+  /// initializers are not available.
+  void loadInputs(onnx::GraphProto &net);
+
+  /// Load pre-trained weights from \p weightDescriptors.
+  bool loadWeights(uint32_t weightsCount,
+                   const onnxTensorDescriptorV1 *weightDescriptors);
+
+public:
+  /// \returns unique pointer to ModelLoader if \p onnxModel can be parsed
+  /// and static weights can be loaded from the \p wightDescriptors.
+  /// \returns nullptr otherwise.
+  static std::unique_ptr<ModelLoader>
+  parse(const void *onnxModel, uint32_t onnxModelSize, uint32_t weightsCount,
+        const onnxTensorDescriptorV1 *weightDescriptors, Function &F);
+
+  /// \returns unique pointer to ModelLoader if \p onnxModel can be parsed,
+  /// e.g., the model is a valid ONNX model and Glow supports all of the
+  /// operators in the network. \returns nullptr otherwise.
+  static std::unique_ptr<ModelLoader> parse(const void *onnxModel,
+                                            size_t onnxModelSize, Function &F);
+};
+
+} // namespace onnxifi
+} // namespace glow
+
+#endif // GLOW_IMPORTER_ONNXIFILOADER_H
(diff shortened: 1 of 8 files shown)


tests/unittests/BackendTestUtils.cpp
```diff
@@ -858,8 +858,8 @@ void inferMaxSplat(Tensor *input, Tensor *out, BackendKind kind) {
 
   auto *var = VarFrom(input);
   auto T = mod.uniqueType(ElemKind::Int8QTy, var->getType()->dims(),
-                           2 * var->getType()->getScale(),
-                           -var->getType()->getOffset());
+                          2 * var->getType()->getScale(),
```
rdzhabarov: will submit PR for this first (some clang-formatting thing).

rdzhabarov: should be fixed by https://github.com/pytorch/glow/pull/1420

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -142,15 +142,30 @@ onnxGetBackendCompatibility(onnxBackendID backendID, size_t onnxModelSize,
...
+      glow::onnxifi::ModelLoader::parse(onnxModel, onnxModelSize, *F);
 
+  // TODO: Make better error reporting.
   if (!loader) {
     return ONNXIFI_STATUS_UNSUPPORTED_OPERATOR;
   }
 
+  // Make sure that every node from the parsed function F
+  // can be executed by the specific backend.
+  for (const auto &op : F->getNodes()) {
```
yinghai: So model parse can be successful with unsupported ops inside the model?

lib/Onnxifi/Base.h
```diff
@@ -81,6 +94,7 @@ class Graph {
 
 private:
   BackendPtr backendPtr_;
+  Function *function_;
```
yinghai: Just curious, what does `Function` in glow mean?

yinghai:
I think we have `setGraphIO` and `Run` to go.

include/glow/Importer/ONNXIFILoader.h
```diff
@@ -0,0 +1,58 @@
...
+#define GLOW_IMPORTER_ONNXIFILOADER_H
+
+#include "onnx/onnxifi.h"
+
+#include "glow/Importer/ONNX.h"
+
+namespace glow {
+namespace onnxifi {
+
+class ModelLoader : public ONNXModelLoader {
```
jfix71: I'm not sure about the naming here -- I would initially assume from the names that a ModelLoader would not be ONNX specific, and that ONNXModelLoader would inherit from ModelLoader, adding in ONNX-specific functionality.

lib/Onnxifi/onnxifiGlow.cpp
rdzhabarov: Yes, model parsing is generic and not specific to the backend.
Like we could add an Op support on Interpreter backend but do not support that on CPU backend for example.

lib/Onnxifi/Base.h
rdzhabarov: It represents the compute graph: variables + actual operations. It's backend independent and needs to be compiled down to a specific backend.
That should be done on graphInit as well, I'll add that later.

include/glow/Importer/ONNXIFILoader.h
rdzhabarov: it's in glow::onnxifi namespace hence onnxifi specific.

include/glow/Importer/ONNXIFILoader.h
```diff
@@ -0,0 +1,58 @@
...
+  bool loadWeights(uint32_t weightsCount,
+                   const onnxTensorDescriptorV1 *weightDescriptors);
+
+public:
+  /// \returns unique pointer to ModelLoader if \p onnxModel can be parsed
+  /// and static weights can be loaded from the \p wightDescriptors.
+  /// \returns nullptr otherwise.
+  static std::unique_ptr<ModelLoader>
+  parse(const void *onnxModel, uint32_t onnxModelSize,
+               uint32_t weightsCount,
```
yinghai: weird indent?

lib/Onnxifi/Base.cpp
```diff
@@ -38,6 +45,16 @@ void Event::wait() {
 onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
                             uint32_t weightCount,
                             const onnxTensorDescriptorV1 *weightDescriptors) {
+  // TODO: support multiple functions here.
+  function_ = backendPtr_->getEE().getModule().createFunction("inference");
+
+  std::unique_ptr<ModelLoader> loader = ModelLoader::parse(
```
yinghai: loader should be a member function of Graph, right?

include/glow/Importer/ONNXIFILoader.h
rdzhabarov: yup, pretty sure I ran the clang-format. Let me see.

lib/Onnxifi/Base.cpp
rdzhabarov: I'll likely need to compile at this point (on initGraph), not sure if loader will stay at that point. We can revisit this later.

include/glow/Importer/ONNXIFILoader.h
rdzhabarov: updating...

PR #1421, opened 2018-08-07
[quantization]: Support splat node for quantization

ZchiPitt:
Support splat node for quantization

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.h
+++ b/lib/Backends/OpenCL/OpenCL.h
@@ -183,6 +183,7 @@ public:
       case Kinded::Kind::ConvolutionNodeKind:
       case Kinded::Kind::DequantizeNodeKind:
       case Kinded::Kind::DivNodeKind:
+      case Kinded::Kind::FullyConnectedNodeKind:
       case Kinded::Kind::MaxNodeKind:
       case Kinded::Kind::MinNodeKind:
       case Kinded::Kind::MulNodeKind:
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -296,6 +296,14 @@ static Node *quantizeNode(Function *F, Node *node,
     quantizedNode = F->createSigmoid(SN->getName(), quantizedInputs[0]);
     break;
   }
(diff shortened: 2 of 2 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -296,6 +296,14 @@ static Node *quantizeNode(Function *F, Node *node,
     quantizedNode = F->createSigmoid(SN->getName(), quantizedInputs[0]);
     break;
   }
+  case Kinded::Kind::SplatNodeKind: {
+    auto *SPN = cast<SplatNode>(node);
```
rdzhabarov: please add a test to end2end quantization test to cover splat node.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -183,6 +183,7 @@ class OCLBackend final : public Backend {
       case Kinded::Kind::ConvolutionNodeKind:
       case Kinded::Kind::DequantizeNodeKind:
       case Kinded::Kind::DivNodeKind:
+      case Kinded::Kind::FullyConnectedNodeKind:
```
opti-mix: Why do we need this for the splat support?

rdzhabarov: @ZchiPitt please make that change in a separate commit for clarity.

lib/Quantization/Quantization.cpp
rdzhabarov: I think it would be better to implement this as a graph optimization by merging quantization node into splat. This potentially will cover more scenarios for splat -> quantize sequences.

ZchiPitt: i just realize the code here is useless, as the splat would still be floating op as i dont have the quantized type for the splat here.

ZchiPitt: maybe i just disable the support for SplatNode in isOpSupported() ?

rdzhabarov:
@ZchiPitt can you submit
```
"case Kinded::Kind::FullyConnectedNodeKind:"
```
change independently though?

PR #1423, opened 2018-08-08
[onnxifi] Add support for the newly added onnxGetEventState onnxifi method.

rdzhabarov:
cc: @yinghai @Maratyszcza

### DIFF
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -60,7 +60,7 @@ onnxGetBackendIDs(onnxBackendID *backendIDs, size_t *numBackends) {
 /// (onnxBackend, onnxGraph, onnxEvent) before calling this function.
 ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
 onnxReleaseBackendID(onnxBackendID backendID) {
-  auto *backendId = reinterpret_cast<glow::onnxifi::BackendIdPtr>(backendID);
+  auto *backendId = static_cast<glow::onnxifi::BackendIdPtr>(backendID);
   if (!backendID) {
     return ONNXIFI_STATUS_INVALID_ID;
   }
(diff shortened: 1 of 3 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -245,6 +245,23 @@ onnxWaitEvent(onnxEvent event) {
   return ONNXIFI_STATUS_SUCCESS;
 }
 
+/// Query ONNXIFI event state without blocking.
+ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
+  onnxGetEventState(
+    onnxEvent event,
+    onnxEventState* state) {
+  auto *glowEvent = reinterpret_cast<glow::onnxifi::EventPtr>(event);
```
Maratyszcza: `static_cast` may be a better option. `onnxEvent` is just `void*`

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -245,6 +245,23 @@ onnxWaitEvent(onnxEvent event) {
...
+  if (!glowEvent) {
+    return ONNXIFI_STATUS_INVALID_EVENT;
+  }
+  if (!state) {
```
Maratyszcza: This function must set `*state` to `ONNXIFI_EVENT_STATE_INVALID` if it fails. I suggest that you do the check `state != nullptr`, and immediately set `*state` to `ONNXIFI_EVENT_STATE_INVALID` if the check succeeds. You will later overwrite `*state` with the proper value before returning.

lib/Onnxifi/onnxifiGlow.cpp
rdzhabarov: @Maratyszcza Seems like the majority (all) onnxifi definitions for (event, graph, backend, etc) are void*. Is it safe to assume that it's not going to change? (There are other places in this implementation where reinterpret_cast could be replaced by static_cast, I'll change that in a separate commit).

lib/Onnxifi/onnxifiGlow.cpp
rdzhabarov: @Maratyszcza good point!

PR #1424, opened 2018-08-08
[OpenCL] Support conv non-square strides and kernels for OpenCL backend

beicy:


lib/Backends/OpenCL/kernels.cl
```diff
@@ -844,17 +849,17 @@ __kernel void convolutionK(__global float *dest, __global float *src,
...
 __kernel void convolutionW(__global void *mem, cl_uint32_t dest,
                            cl_uint32_t src, cl_uint32_t filter,
-                           cl_uint32_t bias, cl_uint32_t filterSize,
-                           cl_uint32_t stride, PaddingTLBR pads, ShapeNHWC odim,
+                           cl_uint32_t bias, ShapeHW filterSizes,
+                           ShapeHW strides, PaddingTLBR pads, ShapeNHWC odim,
                            ShapeNHWC idim, ShapeNHWC filterDim) {
-  convolutionK(&mem[dest], &mem[src], &mem[filter], &mem[bias], filterSize,
-               stride, pads, odim, idim, filterDim);
+  convolutionK(&mem[dest], &mem[src], &mem[filter], &mem[bias], filterSizes,
```
artemrakhov: @opti-mix , @ZchiPitt : what is the status of this code? Does it only run for training? Any plans to enable OCLConv for training and remove this?

opti-mix: @artemrakhov Yes, there are such plans. But we are not there yet. Once OCLConv can be used for training as well, this code is not needed anymore.

As for testing, one needs to run our tests like "mnist -opencl" which would use training with OpenCL backend and thus exercise the changes. Of course, mnist does not use non-square strides. To test that, one would need to write a new unit tests that uses training & non-square strides and kernels.

opti-mix:
Also, may be you could add a small test for training with non-square strides and kernels? As @artemrakhov correctly mentioned in his comment, we only check the inference in unit tests at the moment, IIRC.

beicy:
@opti-mix Thank you so much for reminding the tests. Actually, I forgot to test different devices. Now all items on your list are tested.

beicy:
@artemrakhov @opti-mix : The convGradTest in BackendCorrectnessTest.cpp and OCLTest.cpp is modified to test the non-square kernels and strides. Actually, the convGradTest in OCLTest.cpp is duplicated since it has been tested in BackednCorrectnessTest.cpp already. Can it be removed ?

opti-mix:
@beicy If it is exactly the same test and it is tested in BackendCorrectnessTest with OpenCL backend, then we could remove it from OCLTest.

opti-mix:
> The convGradTest in BackendCorrectnessTest.cpp and OCLTest.cpp is modified to test the non-square kernels and strides

Great! Thanks!

PR #1425, opened 2018-08-08
Extend pow operator

artemrakhov-glow:
I need ability to represent Exp(x) in Glow graph. Decided to extend existing Pow node, so that both LHS and RHS are arbitrary Tensors. Later backends can pattern match cases of Pow(Tensor, Splat) and Pow(e, Tensor) to generate more efficient code.

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -44,7 +44,7 @@ class InterpAndCPU : public Operator {};
 
 class InterpOnly : public Operator {};
 
-TEST_P(InterpAndCPU, pow) {
+TEST_P(Operator, pow) {
   auto *X = mod_.createVariable(ElemKind::FloatTy, {1, 1, 3}, "X");
   auto *Y = mod_.createVariable(ElemKind::FloatTy, {2}, "Y");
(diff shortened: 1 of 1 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1043,8 +1043,13 @@ CmpEQNode *Function::createCmpEQ(llvm::StringRef name, NodeValue LHS,
   return addNode(new CmpEQNode(name, OT, LHS, RHS));
 }
 
+PowNode *Function::createPow(llvm::StringRef name, NodeValue Base, NodeValue Exp) {
```
jfix71: nit: While we're refactoring, maybe change this to `base` and `exp`? Here and in the other `createPow()` too.

artemrakhov: Done. But I'm not sure what is out convention here. I see both `RHS` and `rhs`, also `Cond` and `Ty` and `ty`

artemrakhov:
Added Pow to `Node::isArithmetic()`

lib/Graph/Graph.cpp
jfix71: Yeah it's a bit inconsistent in some places. I think `RHS` makes sense over `rhs` since it's an acronym. I think all others should be camel case, e.g. `cond` and `ty`.

PR #1426, opened 2018-08-08
[opencl]: Support several nodes for int8 operation

ZchiPitt:
Support several nodes for int8 operation:
-- FullyConnectedNode (lowered to matmul and batchedAdd)
-- ConcatNode (lowered to inserttensor)
-- ReshapeNode (lowered to copyinst)
-- SliceNode (lowered to extracttensor)

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.h
+++ b/lib/Backends/OpenCL/OpenCL.h
@@ -180,9 +180,11 @@ public:
     if (elementTy == ElemKind::Int8QTy) {
       switch (opKind) {
       case Kinded::Kind::AddNodeKind:
+      case Kinded::Kind::ConcatNodeKind:
       case Kinded::Kind::ConvolutionNodeKind:
       case Kinded::Kind::DequantizeNodeKind:
       case Kinded::Kind::DivNodeKind:
+      case Kinded::Kind::FullyConnectedNodeKind:
       case Kinded::Kind::MaxNodeKind:
       case Kinded::Kind::MinNodeKind:
       case Kinded::Kind::MulNodeKind:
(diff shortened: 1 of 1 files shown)


opti-mix:
@ZchiPitt  Do we need to explicitly enable any tests to check that these operations work with quantized tensors now?

ZchiPitt:
just checked that end2end tests all these nodes, so we should be fine with the test now

qcolombet:
As much as possible I would prefer simple unittests that always run on top of end2end. @ZchiPitt how hard would it be to add such tests?

ZchiPitt:
@qcolombet do you mean separate unittest for each node? Im not sure i understand what "on top of end2end" means?

qcolombet:
>  do you mean separate unittest for each node? 

Yes

>  Im not sure i understand what "on top of end2end" means?

I mean that you've already done end2end testing, so on top of that :), I would like to see unittests. But it can be done as a follow-up PR.

qcolombet:
Oooh, thanks for the clarification @opti-mix!

rdzhabarov:
yeah, e2e in this context means:
1) Add profiling nodes to the graph
2) Capture profiles
3) Convert graph to quantized form. 
4) Make sure that results from the fp32 graph match results of quantized graph.

Note, not all nodes should/will be quantized from the graph. And the number of quantized nodes would depend on whether backend supports specific operation. But overall it seems useful to add additional check on # of quantized nodes (but that will be backend dependent).

ZchiPitt:
cool, there can be improvement for the end2end test that checks how many nodes got quantized in the future pr and it can also show the usage of how each node got quantized

PR #1427, opened 2018-08-09
Rename MemberType's SizeT to UInt64, and move it from size_t to uint64_t

jfix71:
I tried to do this in only places that make sense. However there's a good chance there are mistakes in here. If anyone has tips on ways to find them let me know. It's a really large diff so it will be hard to review, but not sure how to break it up in a sane way.

One place I made the change to `uint64_t` was for `size()` in `Type`, `Handle`, and `Tensor`. This is because these are generally compared against with other indices and also calculated using the indices, and so their type should match. Basically I tried to only leave `size_t` when it was used for a length of an object, e.g. `dims().size()`.

Note: after this PR I was planning to rename `ElemKind::IndexTy` to `ElemKind::UInt64ITy`, where `ITy` represents this is an index type. This is to be more consistent with our quantized types which end in `QTy` (e.g. `Int8QTy`).


qcolombet:
One high level question:
- What is the motivation for the change?

One high level remark:
- In commit messages, I like to see the rational for a change. Usually we put that in the PR, but when doing `git log`, we only get the commit message and that's not super convenient to track it back to the PR.

jfix71:
>What is the motivation for the change?

There are a couple motivations. First, `size_t` has a platform-dependent bit width. Since we're compiling for other architectures, I didn't think it made sense for us to use `size_t` for our `IndexTy`. I thought it made more sense to have explicitly sized types.

Second, we load ONNX and Caffe2 protos with different integer types -- for example, Caffe2 has `GivenTensorIntFill` (represented in Caffe2 as a signed 32-bit integer) and `GivenTensorInt64Fill`(represented in Caffe2 as a signed 64-bit integer). Currently we load both of these as tensors with `ElemKind::IndexTy`, implemented as `size_t` prior to this PR. This has not been a problem because we have always seen positive values for these fills, but I'm currently loading protos with `-1` as their value. So after this I was going to rename `ElemKind::IndexTy` to `ElemKind::UInt64ITy`, and then I could add a new `ElemKind::Int64ITy` and/or `ElemKind::Int32ITy` to represent these `-1`s.

Note that as far as I've seen, `-1`s are only used for tensors used exclusively as shape info to infer the (positive) dimension from the rest of the shape. So theoretically they could be made positive while building the graph, or we could use their data to inform our shape info and never load them as tensors at all, avoiding the need for a signed `ElemKind`. @artemrakhov had mentioned this as a possibility before. However, it's not possible to tell if a tensor is exclusively used as shape info without iterating over the proto we're loading a second time. So I felt that when combined with the first reason, it made more sense to move to explicitly sized types instead, to clear the way for signed `ElemKind`s.

Also, this change would enable us to add `ElemKind`s with sizes consistent with Caffe2, though assuming we use 64 bit types we should be able to avoid inconsistent results with Caffe2 (though if they overflow we wouldn't...). So this is only a benefit depending on how closely we want to replicate Caffe2 results. Note, ONNX also [supports different data types](https://github.com/onnx/onnx/blob/master/onnx/defs/data_type_utils.cc#L381), including different sizes and both signed/unsigned integers.

jfix71:
Also no problem on the commit message, I didn't realize that for `git log` -- I always use the web interface here when looking at old commits, which makes it easy to find the PR from which a commit came.

opti-mix:
@jfix71 You replace size_t by uint64_t, but both types are concrete types which are used at a huge number of places for different reasons. The names of these types do not reflect their semantics, they describe only their value range. Combined, this makes renaming like this very difficult because you really have to look at each use and think what is semantically required there :-( And if you decide later that you want to use a different type again, you'll have too look at all uses of uint64_t in the code base :-(

 If you are going to perform this grand-renaming, I'd suggest introducing type aliases (and there can be multiple aliases for the same underlying type, because you may have multiple semantic contexts) with meaningful names (e.g. IndexTy, TensorSizeTy, TensorDimTy, etc) and use those names at the right places. This will make the code more understandable and it will also make any change of these types in the future much easier, should we need it.

qcolombet:
Hi @jfix71, thanks for the detailed explanation.
Makes sense. Make sure to reflect that rational in the commit message.

Finally, I agree with @opti-mix, while we are doing this big renaming, let us introduce proper types for the different use-cases.

jfix71:
Closing this in favor of different approach as discussed offline.

Issue #1346, opened 2018-07-27
Our def-use chains are counter intuitive (and error-prone IMHO)


qcolombet:
When we have nodes with several outputs (like TopK or MeanVarNormalization), we don't track the def-use chain per output but instead we do it per node.

In effect, imagine we have something like:
auto *topK = createTopK
auto *values = topK.getValues();
auto *indices = topK.getIndices();

Walking the users for all three variables will give us the exact same def-use chain.

So for instance, walking through the users of indices and doing some changes without checking that the use is of the right NodeValue (i.e., checking that the ResNos are the same) will produce incorrect code.

Another instance is checking hasOneUse for values could be false because indices could also be used.

replaceAllUsesOfWith does the right thing, but one would expect that the number of updates that function does is equal to the number of users the related value has. Right now, it is not the case.

This bit me when I was writing some asserts where I thought the NodeValue would have one use.

I'll write a test case to show the issue.

opti-mix:
@qcolombet Yes, we currently keep the def-use chain per node. 

Do you suggest to keep it per output? Or do you suggest that iteration over users of a NodeValue should yield only relevant uses from the def-use chain?

Yielding only the relevant elements of the def-use chain should be rather easy, I think. We would just need a simple wrapper around the current def-use chain iterator, which would check if the use is a use of the expected output.

Having a def-use chain per output could be more difficult to implement.

qcolombet:
> Do you suggest to keep it per output? Or do you suggest that iteration over users of a NodeValue should yield only relevant uses from the def-use chain?

Having NodeValue filtering the relevant uses should be fine.

qcolombet:
BTW I wonder if we should kill the operator* that directly uses the node, because it gets easy to use ->getUsers() vs. .getUsers() from NodeValue yet it would yield different results.

opti-mix:
> BTW I wonder if we should kill the operator* that directly uses the node, because it gets easy to use ->getUsers() vs. .getUsers() from NodeValue yet it would yield different results.

I agree that `*` and `->` operators of `NodeValue` are pretty dangerous and it is too easy to use it in a wrong way.

stoklund:
A filtering iterator makes sense to me. Could there be a need for a non-filtering variant too?

What about `getNumUsers()`? I think most of the time you only care if there are 0, 1, or multiple users, not the actual number of users.

opti-mix:
[...]

PR #1428, opened 2018-08-09
[unittest][Graph] Check that the use list of multi-res node behave in a reasonable way

qcolombet:
… a reasonable way

Right now, we have no way to check the use list of each result individually.
This test illustrates how confusing it is and this issue is tracked in #1346.

NFC.


qcolombet:
> You mentioned in your comment the "->" vs ".". Maybe it's time to get rid of the implicit conversion between node and nodevalue?

Definitely! PR's coming.

PR #1429, opened 2018-08-09
[graphOptimizer]: SinkCode again before the end of optimization

ZchiPitt:
By profiling the squeezenet and vgg, i found that we get the pattern that"one transpose right after another transpose" during the optimizations, after sinkNode has finished. Which means some other optimization (after sinkNode) rewrites the graph or removes something so that suddenly we have transpose(transpose). 

I propose to sink code again before the end of optimization, and it did remove the extra transpose.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1794,6 +1794,11 @@ void glow::optimize(Function *F, CompilationMode mode) {
     optimizeQuantization(F);
   }
 
+  while (sinkCode(F)) {
+    // Perform Dead Code Elimination between rounds of code sinking.
+    DCE(F);
+  }
+
   // Perform Dead Code Elimination.
   DCE(F);
 }


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1794,6 +1794,11 @@ void glow::optimize(Function *F, CompilationMode mode) {
     optimizeQuantization(F);
   }
 
+  while (sinkCode(F)) {
```
opti-mix: Let's say that sinkCode returns true. Then DCE inside the loop-body will be executed. Later, at some point sinkCode would return false and the loop would terminate. The last executed transformation performed by the will be DCE. And then after the loop you have another DCE call, which would be useless in such a case. 

To avoid this useless DCE call, you should probably place the loop after the last DCE call.

opti-mix:
It is cool that this additional sinkCode calls improve the code!

> Which means some other optimization (after sinkNode) rewrites the graph or removes something so that suddenly we have transpose(transpose).

Have you identified which specific optimization results in those transpose/transpose opportunities? I think it would be interesting to know.

May be your sinkCode calls should be inserted right after that optimization rather than at the very end, because it could create more optimization opportunities? 

Or may be there is a cheaper way to solve this issue? After all sinkCode traverses all nodes on each iteration of your loop. It could have impact on the compile time, though it is probably unlikely to be noticeable.

rdzhabarov:
does that happen on openCL backend or on all backends for the networks you posted in the description?

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: good catch!

opti-mix:
@ZchiPitt Ping!

PR #1430, opened 2018-08-09
[quantization]: Quantize splat node

ZchiPitt:
Quantize splat node in Quantization.cpp:quantizeNode()
As i talked with @rdzhabarov, this support is anyhow needed if the original graph has splat node, and we have to quantize it based on profiling information

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -255,6 +255,16 @@ static Node *quantizeNode(Function *F, Node *node,
         F->createConcat(node->getName(), quantizedInputs, C->getDim(), outTy);
     break;
   }
+  case Kinded::Kind::SplatNodeKind: {
+    auto *SPN = cast<SplatNode>(node);
+    assert(quantizedInputs.size() == 0 && "Invalid number of inputs");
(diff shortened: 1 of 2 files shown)


opti-mix:
@ZchiPitt Nice! Do we also need to add any graph optimizations for quantized splats, like: `quantize(splat)->quantized_splat` or something like this, i.e. basically merging quantization or rescaling directly into splat? Or may be this is handled by our existing optimizations already?

lib/Quantization/Quantization.cpp
```diff
@@ -255,6 +255,15 @@ static Node *quantizeNode(Function *F, Node *node,
         F->createConcat(node->getName(), quantizedInputs, C->getDim(), outTy);
     break;
   }
+  case Kinded::Kind::SplatNodeKind: {
+    auto *SPN = cast<SplatNode>(node);
+    assert(qParams.size() == 1 && "Invalid number of quantized outputs");
```
rdzhabarov: assert also on no quantized inputs here.

ZchiPitt: good catch

ZchiPitt:
the optimization could be a follow-up pr as it's not directly needed by this pr

PR #1432, opened 2018-08-09
[caffe2] Fix out_of_range exception in Caffe2 importer while loading a ConstantFill tensor

aturetsk:
Some Caffe2 models we use seem to have ConstantFill tensors, which don't have "value" attribute specified, so GLOW crashes with out_or_range exception in caffe2ModelLoader::loadWeight.
The fix addresses this.


nadavrot:
Nice!

PR #1433, opened 2018-08-09
[NodeValue] Get rid of the implicit conversion when using the operator->

qcolombet:
The change is pretty mechanical. Namely, this changes the pattern
NodeValue-><Node_method>
Into
NodeValue.getNode()-><Node_method>

Also, as part of this change, we actually get cleaner accesses to
dimensions, types, etc. Indeed, instead of hardcoding the index of
the result, the NodeValue directly gets us the right value.

Related to #1346.

NFC.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -82,8 +82,6 @@ public:
   /// Replace all of the uses of this value with \p v.
   void replaceAllUsesOfWith(NodeValue v);
 
-  /// Provide a smart-pointer interface.
-  Node *operator->() const { return node_; }
   /// Return the TypeRef of the referenced return value.
   TypeRef getType() const;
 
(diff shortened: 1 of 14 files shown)


qcolombet:
Hi @jfix71, thanks for the review and for the suggestion, I was wondering the same thing when I did the patch and I decided against it.

I think the difference with `getName` is unlike `dims` and `getElementType`, the returned value does not change from when we use different value from the same node.

What am I saying is so far, we did this "forwarding" when Node had <method>(ResNo) which thus would translate into NodeValue::<method>() (i.e., the ResNo argument gets eliminated).
`getName` doesn't take ResNo in Node, so it does not seem we should forward that method.

Another candidate though would be getOutputName, but it does not seem to be used often enough to be useful.

Finally, I think we will want to do that for hasOneUser, getNumUsers and the like, while adding a ResNo parameter for the Node itself so that we finally have something sensible regarding #1346.

jfix71:
Agreed, that all makes sense. The others are more for convenience, whereas `dims()` and `getElementType()` prevent accidental bugs from incorrect ResNo usage.

nadavrot:
Ah, missed the discussion above. Never mind. Thanks.

PR #1434, opened 2018-08-09
[quantization] Add quantization support for LRN and SoftMax in middle-end

aturetsk:
Add quantization support for LRN and SoftMax in GLOW middle-end, making possible for any back-and to implement them, if needed.

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -50,9 +50,12 @@ public:
 
   ~ExecutionEngine();
 
-  // Set the code generator kind to \p backendKind. New code will be generated
-  // using this backend.
+  /// Set the code generator kind to \p backendKind. New code will be generated
+  /// using this backend.
   void setBackend(BackendKind backendKind);
+
+  /// Set the code generator to a custom \p backend.
+  void setBackend(Backend *backend);
 
   /// \returns the internal graph.
   Module &getModule() { return M_; }
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -41,9 +41,15 @@ static llvm::cl::opt<bool> dumpIR("dump-ir",
 ExecutionEngine::ExecutionEngine(BackendKind backendKind)
     : backend_(createBackend(backendKind)) {}
 
-// Set the code generator kind to \p backendKind.
+/// Set the code generator kind to \p backendKind.
 void ExecutionEngine::setBackend(BackendKind backendKind) {
   backend_.reset(createBackend(backendKind));
+  function_.reset();
+}
+
+/// Set the code generator kind to \p backend.
+void ExecutionEngine::setBackend(Backend *backend) {
+  backend_.reset(backend);
   function_.reset();
 }
 
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -99,6 +99,10 @@ static bool canBeQuantized(const Node *node) {
     auto *gather = cast<GatherNode>(node);
     return gather->getData().getElementType() == ElemKind::FloatTy;
   }
+  case Kinded::Kind::SoftMaxNodeKind: {
+    auto *SMN = cast<SoftMaxNode>(node);
+    return SMN->getInput().getElementType() == ElemKind::FloatTy;
+  }
   default:
     // Let the general procedure handle this node kind.
     break;
@@ -305,6 +309,24 @@ static Node *quantizeNode(Function *F, Node *node,
     // Note: This should either be lowered into an IntLookupTable, or
     // implemented via a backend-specific Node/Inst.
     quantizedNode = F->createSigmoid(SN->getName(), quantizedInputs[0]);
+    break;
+  }
+  case Kinded::Kind::LocalResponseNormalizationNodeKind: {
+    auto *LRN = cast<LocalResponseNormalizationNode>(node);
+    assert(quantizedInputs.size() == 1 && "Invalid number of inputs");
+    assert(qParams.size() == 1 && "Invalid number of quantized outputs");
+    quantizedNode = F->createLocalResponseNormalization(
+        LRN->getName(), quantizedInputs[0], LRN->getHalfWindowSize(),
+        LRN->getAlpha(), LRN->getBeta(), LRN->getK());
+    break;
+  }
+  case Kinded::Kind::SoftMaxNodeKind: {
+    auto *SMN = cast<SoftMaxNode>(node);
+    // SoftMax node has 2 inputs, but only one should be quantized.
+    assert(quantizedInputs.size() == 1 && "Invalid number of inputs");
+    assert(qParams.size() == 1 && "Invalid number of quantized outputs");
+    quantizedNode = F->createSoftMax(SMN->getName(), quantizedInputs[0],
+                                     SMN->getSelected());
     break;
   }
   default:
--- a/tests/unittests/quantizationTest.cpp
+++ b/tests/unittests/quantizationTest.cpp
@@ -17,6 +17,7 @@
 #include "glow/Quantization/Quantization.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/Graph/Graph.h"
+#include "glow/IR/IR.h"
 #include "glow/Quantization/Serialization.h"
 
 #include "gtest/gtest.h"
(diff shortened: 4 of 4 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -296,6 +300,30 @@ static Node *quantizeNode(Function *F, Node *node,
...
     break;
   }
+  case Kinded::Kind::LocalResponseNormalizationNodeKind: {
+    auto *LRN = cast<LocalResponseNormalizationNode>(node);
+    assert(quantizedInputs.size() == 1 && "Invalid number of inputs");
+    assert(qParams.size() == 1 && "Invalid number of quantized outputs");
+
+    // Note: This should either be lowered into an IntLookupTable, or
+    // implemented via a backend-specific Node/Inst.
+    quantizedNode = F->createLocalResponseNormalization(
```
rdzhabarov: how LRN is supposed to be implemented via IntLookupTable?

lib/Quantization/Quantization.cpp
```diff
@@ -296,6 +300,30 @@ static Node *quantizeNode(Function *F, Node *node,
...
+  }
+  case Kinded::Kind::SoftMaxNodeKind: {
+    auto *SMN = cast<SoftMaxNode>(node);
+    // SoftMax node has 2 inputs, but only one should be quantized.
+    assert(quantizedInputs.size() == 1 && "Invalid number of inputs");
+    assert(qParams.size() == 1 && "Invalid number of quantized outputs");
+
+    // Note: This should either be lowered into an IntLookupTable, or
+    // implemented via a backend-specific Node/Inst.
+    quantizedNode = F->createSoftMax(SMN->getName(), quantizedInputs[0],
```
rdzhabarov: Do you have a plan to add softMax support to the interpreter?

lib/Quantization/Quantization.cpp
```diff
@@ -99,6 +99,10 @@ static bool canBeQuantized(const Node *node) {
     auto *gather = cast<GatherNode>(node);
     return gather->getData().getElementType() == ElemKind::FloatTy;
   }
+  case Kinded::Kind::SoftMaxNodeKind: {
```
rdzhabarov: as long as we quantize node, we should ensure proper support at least by interpreter backend.

lib/Quantization/Quantization.cpp
nadavrot: I am not sure how we can quantize SoftMax because the intermediate sum can be huge, but if some backends need it then we can go for it.

rdzhabarov: I think @jspark1105 had some ideas about this. Where did we end up with the quantized softmax?

jspark1105: We haven't quite finished our work on quantizing softmax. gemmlowp has a quantized softmax implementation: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/lite/kernels/internal/optimized/optimized_ops.h#L3907

rdzhabarov:
I think this PR is OK to ship to unblock you if you are blocked on this (without implementation of quantized softmax). Please, make sure to update the comments properly though. 
Another thing, can you ensure code coverage for these new operators. You can create mock backend and report isOpSupported for softmax and/or LRN in order to cover this path. See some examples in QuantizationTest.

aturetsk:
> Please, make sure to update the comments properly though.

Yes, comments are a copy-paste error. I'll remove them.

> Another thing, can you ensure code coverage for these new operators. You can create mock backend and report isOpSupported for softmax and/or LRN in order to cover this path. See some examples in QuantizationTest.

I'm not sure if it can be done with current interfaces. quantizeFunction (which I need to test), has ExecutionEngine argument and takes Backend object from it. However, I can't initialize EE object with a custom back-end, because EE constructor and setBackend function take a enum rather than a Backend object.
Or am I missing something?

> I think this PR is OK to ship to unblock you if you are blocked on this (without implementation of quantized softmax). 

Yes, currently we are working on quantized LRN and Softmax in our back-end, so I had to add this code.
I just though that since the code is in middle end it might be a good idea to commit it in master (and it'd help to avoid rebase conflicts in the future). If you find it unacceptable to have the changes without the test coverage (due to the issues I described above), I can cancel the PR and keep the changes in our local repo.

rdzhabarov:
> I'm not sure if it can be done with current interfaces. quantizeFunction (which I need to test), has ExecutionEngine argument and takes Backend object from it. However, I can't initialize EE object with a custom back-end, because EE constructor and setBackend function take a enum rather than a Backend object.
Or am I missing something?

Quantization conversion procedure relies on the isOpSupported from the ExecutionEngine and you can just pass MockEE to ```F2 = quantization::quantizeFunction(backendSpecificEE, QI, F2);```
Make mock to return true for certain operations and you should get F2 conversion as expected (Note, you'd need to make isOpSupported on the EE level virtual).

> I just though that since the code is in middle end it might be a good idea to commit it in master (and it'd help to avoid rebase conflicts in the future)

Certainly, let's do it. Do you think you could add a simple test for the conversion?

aturetsk:
> Quantization conversion procedure relies on the isOpSupported from the ExecutionEngine and you can just pass MockEE to F2 = quantization::quantizeFunction(backendSpecificEE, QI, F2);
Make mock to return true for certain operations and you should get F2 conversion as expected (Note, you'd need to make isOpSupported on the EE level virtual).

But I can't derive MockEE from EE, since EE is 'final' class.
I was able to overcome the restriction by overloading setBackend method and allowing it to have a pointer to Backend object as argument (making it possible to pass a custom mock backend to it).
Are you OK with this approach?

rdzhabarov:
I'm fine either with that (as long as backendKind is not kept in the internal state which is not) or just remove "final" as I do not see a strong reason to make that final as well.

aturetsk:
@rdzhabarov, then let's stick to the current implementation.
Is there anything else I need to address before the patch can be committed?

rdzhabarov:
oh, I did not realize that you updated the PR :) Will look @aturetsk

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -54,6 +54,9 @@ class ExecutionEngine final {
   // using this backend.
   void setBackend(BackendKind backendKind);
 
+  // Set the code generator to a custom \p backend.
```
rdzhabarov: /// doxygen comments. If you could please fix the one above as well.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -47,6 +47,12 @@ void ExecutionEngine::setBackend(BackendKind backendKind) {
   function_.reset();
 }
 
+// Set the code generator kind to \p backend.
```
rdzhabarov: same here.

tests/unittests/quantizationTest.cpp
```diff
@@ -577,6 +578,53 @@ TEST(Quantization, chooseQuantizationSymmetricWithUInt8) {
   EXPECT_NEAR(symmetricParams.scale, 16.0 / 255, 0.001);
 }
 
+/// This is a mock backend which extended support of quantizad operators.
```
rdzhabarov: typo in quantized.

tests/unittests/quantizationTest.cpp
```diff
@@ -577,6 +578,53 @@ TEST(Quantization, chooseQuantizationSymmetricWithUInt8) {
...
+  F->createSave("ret", SM);
+
+  std::vector<NodeQuantizationInfo> QI{
+      {NodeQuantizationInfo::generateNodeOutputName(input->getName()),
+       {0.2f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(LRN->getName()), {0.3f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(SM->getName()), {0.4f, 0}},
+  };
+
+  F = quantization::quantizeFunction(EE, QI, F);
```
rdzhabarov: LGTM. Can you check that inputs to the LRN and SM is actually int8.

aturetsk:
@rdzhabarov, thanks for review!
I fixed your remarks.

tests/unittests/quantizationTest.cpp
```diff
@@ -577,6 +578,60 @@ TEST(Quantization, chooseQuantizationSymmetricWithUInt8) {
...
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  auto *input = mod.createVariable(ElemKind::FloatTy, {1, 10}, "input");
+  auto *selected = mod.createVariable(ElemKind::IndexTy, {1, 10}, "selected");
+  auto *LRN =
+      F->createLocalResponseNormalization("LRN", input, 2, 1.0, 0.0001, 0.75);
+  auto *SM = F->createSoftMax("softmax", LRN, selected);
+  F->createSave("ret", SM);
```
rdzhabarov: this is fine, but FYI, if you do not compile "save node" is not necessary.

tests/unittests/quantizationTest.cpp
```diff
@@ -577,6 +578,60 @@ TEST(Quantization, chooseQuantizationSymmetricWithUInt8) {
...
+  std::vector<NodeQuantizationInfo> QI{
+      {NodeQuantizationInfo::generateNodeOutputName(input->getName()),
+       {0.2f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(LRN->getName()), {0.3f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(SM->getName()), {0.4f, 0}},
+  };
+
+  F = quantization::quantizeFunction(EE, QI, F);
+
+  auto *qLRN = cast<LocalResponseNormalizationNode>(F->getNodeByName("LRN1"));
```
rdzhabarov: "LRN1" and "softmax1" is very fragile (depends on the naming schema used, we could change that potentially etc).

Could you just iterate through the nodes in quantized F and check for LRN or SM nodes and then verify that they have quantized type input.

rdzhabarov: Otherwise, if you iterate there will be one non-quantized and one quantized node, also kinda messy...

rdzhabarov:
Thanks @aturetsk!

PR #1438, opened 2018-08-10
[ClassGen] Remove MemberType::SizeT, move all to Unsigned

jfix71:
First baby step toward moving away from using `size_t` in our NodeGen/InstrGen. As discussed in https://github.com/pytorch/glow/pull/1427 and offline, we want to do this as `size_t` is not portable.


jfix71:
Done

PR #1439, opened 2018-08-10
[Operators] Support SparseLengthsWeightedSum

artemrakhov:


artemrakhov:
@nadavrot : I think we can't. In case of 1-D data the formula is:
```
result[0] = weights[0] * data[indices[0]] + weights[1] * data[indices[1]] + ... + weights[lengths[0] - 1] * data[indices[lengths[0] - 1]]
result[1] = weights[lengths[0]] * data[indices[lengths[0]]] + ...
```
So each slice is associated with its own multiplier. It's also possible to have duplicate indices with different weights: `i != j && indices[i] == indices[j] && weights[i] != weights[j]`

@jfix71 : we had this question before, when I first introduced `SparseLengthsSum`. When you first do `Gather`, you get a table of size `len(Indices) * sliceSize`, and then immediately reduce it to `len(Lengths) * sliceSize`. In case we fuse these, you can avoid creating intermediate table and save memory.

nadavrot:
@artemrakhov LGTM. Thank you.

jfix71:
@artemrakhov Perhaps we should consider making that a backend-specific decision? I.e. by default, lower `SparseLengthsWeightedSum` to `Gather` + `LengthsWeightedSum`. Then a backend can return false for `SparseLengthsWeightedSum` in `shouldLower()`, and then provide its own backend-specific inst for the fused `SparseLengthsWeightedSum`. This would be more consistent with how we treat other operators, e.g. FC. Though perhaps it's not so important, since this is a somewhat specialized operator and the wins from fusing would likely be preferred by all backends.

PR #1443, opened 2018-08-10
Move Conv and Pools to use VectorUnsigned instead of VectorSizeT

jfix71:
Going to move these over in chunks until VectorSizeT is entirely unused.


jfix71:
Note that as discussed with @nadavrot I'm leaving the kernels inside the backends to use `size_t` as they currently are. This meant I needed to convert all the unsigned vectors to size_t vectors in LLVMIRGen, as otherwise we have assertion failures on function signature mismatch.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1301,8 +1305,10 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     ShapeVector eDestDims = eBatchDims;
     eDestDims[BR->getAxis()] = 1;
 
-    auto *batchDims = emitConstArray(builder, eBatchDims);
-    auto *destDims = emitConstArray(builder, eDestDims);
+    auto *batchDims =
+        emitConstSizeTArray(builder, llvm::makeArrayRef(eBatchDims));
```
qcolombet: Why these specific uses need to have an explicit makeArrayRef?
(I am not saying it is wrong, just curious :))

include/glow/Graph/Graph.h
```diff
@@ -197,40 +197,40 @@ class Function final : public Named {
   ///@{
 
   ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
-                              size_t depth, llvm::ArrayRef<size_t> kernels,
-                              llvm::ArrayRef<size_t> strides,
-                              llvm::ArrayRef<size_t> pads, unsigned group);
+                              size_t depth, llvm::ArrayRef<unsigned> kernels,
```
qcolombet: Instead of directly using `unsigned`, do we what to use a type alias, like @opti-mix suggested, in case we want to change it in the future?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1120,8 +1120,8 @@ void OpenCLFunction::execute() {
         continue;
       }
 
-      auto odim = ShapeNHWC(odim_vec);
-      auto idim = ShapeNHWC(idim_vec);
+      auto odim = ShapeNHWC(llvm::makeArrayRef(odim_vec));
+      auto idim = ShapeNHWC(llvm::makeArrayRef(idim_vec));
```
qcolombet: Same question for the arrayRef thing.

include/glow/Graph/Graph.h
jfix71: Yeah that probably would have been a good idea!

PR #1445, opened 2018-08-10
[utils] Add two script to help generate simple caffe2/onnx models used for importer unittest.

beicy:
The 2 files in this PR is to provide an sample code for generating simple Caffe2/ONNX models used for our Caffe2/ONNX importer unit testing. The generated model is in text format.  Currently, the files generate a simple CNN with a convolution layer only. Users should customize the code to generate their only models.

### DIFF
--- a/utils/scripts/gen_caffe2_model.py
+++ b/utils/scripts/gen_caffe2_model.py
@@ -1,0 +1,58 @@
+## This is a helper script that generates Caffe2 models.
+## The generated model will be used for Caffe2 importer unittest:    
+## ./tests/unittests/caffe2ImporterTest.cpp 
+## Run $>python gen_caffe2_model.py to get the model files.
+
+from caffe2.proto import caffe2_pb2
+from caffe2.python import utils
+from google.protobuf import text_format
+# Define a weights network
+weights = caffe2_pb2.NetDef()
+weights.name = "init"
+
+op = caffe2_pb2.OperatorDef()
+op.type = "GivenTensorFill"
+op.output.extend(["conv_w"])
+op.arg.extend([utils.MakeArgument("shape", [1, 1, 2, 2])])
+op.arg.extend([utils.MakeArgument("values", [1.0 for i in range(4)])])
+weights.op.extend([op])
+
+op = caffe2_pb2.OperatorDef()
+op.type = "GivenTensorFill"
+op.output.extend(["conv_b"])
+op.arg.extend([utils.MakeArgument("shape", [1])])
+op.arg.extend([utils.MakeArgument("values", [2.0 for i in range(1)])])
+weights.op.extend([op])
+weights.external_output.extend(op.output)
+
+# Define an inference net
+net = caffe2_pb2.NetDef()
+net.name = "predict"
+
+op = caffe2_pb2.OperatorDef()
+op.type = "Conv"
+op.input.extend(["data"])
+op.input.extend(["conv_w"])
+op.input.extend(["conv_b"])
+op.arg.add().CopyFrom(utils.MakeArgument("kernel", 2));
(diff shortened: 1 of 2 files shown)


utils/gen_caffe2_model.py
```diff
@@ -0,0 +1,54 @@
...
+weights = caffe2_pb2.NetDef()
+weights.name = "init"
+
+op = caffe2_pb2.OperatorDef()
+op.type = "GivenTensorFill"
+op.output.extend(["conv_w"])
+op.arg.extend([utils.MakeArgument("shape", [1, 1, 2, 2])])
+op.arg.extend([utils.MakeArgument("values", [1.0 for i in range(4)])])
+weights.op.extend([op])
+#weights.external_output.extend(op.output)
```
rdzhabarov: cleanup dead code.

utils/gen_caffe2_model.py
```diff
@@ -0,0 +1,54 @@
+## This is a helper script that generates simple Caffe2 models.
```
qcolombet: Could you add a comment showing what the invocation of the script looks like?
I.e., what parameters you need to pass.

nadavrot:
Thanks for adding the scripts. Should this go into /utils or /utils/scripts?

yinghai:
Note that @zrphercule is working on C++ ONNXIFI test driver for all the onnx unittest. It will greatly increase the coverage.

beicy:
@nadavrot  Thanks for pointing out the /utils/scripts folder. I didn't notice that. Yes, it is better in /utils/scripts.

utils/scripts/gen_onnx_model.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+graph_def.initializer.extend([weight_tensor])
+graph_def.initializer.extend([bias_tensor])
+graph_def.initializer[0].name = 'W'
+graph_def.initializer[1].name = 'B'
+# Create the model (ModelProto)
+model_def = helper.make_model(graph_def, producer_name='onnx-conv')
+
+with open('simpleConv.onnxtxt', 'wb') as f:
```
artemrakhov: Is it still onnxtxt, even if we write it in binary format (`wb`)?

utils/scripts/gen_onnx_model.py
```diff
@@ -0,0 +1,66 @@
...
+  f.write(format(model_def));
```
artemrakhov: nit: I think `format()` with only one arg doesn't do anything. Let's change it to `str(model_def)`

utils/scripts/gen_onnx_model.py
beicy: Yes, it is in text format. The context of the file to be written is a string.

utils/scripts/gen_caffe2_model.py
```diff
@@ -0,0 +1,58 @@
+## This is a helper script that generates simple Caffe2 models.
+## The generated model will be used for Caffe2 importer unittest:    
+## ./tests/unittests/caffe2ImporterTest.cpp 
+## Run $>python gen_caffe2_model.py to get the model files.
```
qcolombet: Oh I realize now, that the model we want to generate is actually hardcoded in the script ><.
Could you say in the first sentence that the issued model is currently a simple convolution network.

Also could you add a comment to explain what simple means in this context?
For instance, I don't see what are the limitations in the graphs we can write at first glance.

beicy: Since all ops and weights are added manually in the scripts, I don't think we would like to use this script to generate models with a lot of layers or with the input using the real image data. Therefore, I just put the "simple" here.  You are right,  there is no limitations in the graph so far:)  So I think it is better to remove "simple" rather than why it is "simple".

PR #1447, opened 2018-08-11
[GradientTests] Fix Convolution Gradient tests to be more strict.

artemrakhov:
One day @opti-mix pointed out, that if we deleted body of `fwdConvolutionGradInst()`, the test would still pass. The problem was in net architecture: if Conv is not willing to learn, FC would take all work on itself, and FC's weights could learn required thing, so that resulting net as a whole would perform correctly. I changed the net and made sure that if we remove gradient's code, the test fails.

### DIFF
--- a/tests/unittests/gradCheckTest.cpp
+++ b/tests/unittests/gradCheckTest.cpp
@@ -169,36 +169,33 @@ TEST_P(InterpreterGrad, gradientCheckFCConcatRELU) {
   performGradCheck(EE_, result, A, Exp, &inputs, &outputs, 0.0001, 0.001);
 }
 
-static void gradientCheckGroupConv(size_t numInputChan, size_t group,
+static void gradientCheckGroupConv(size_t depth, size_t group,
                                    ExecutionEngine &EE_) {
   size_t numDim = 10;
-  size_t numOutputElem = 10;
 
   auto &mod = EE_.getModule();
   Function *F = mod.createFunction("main");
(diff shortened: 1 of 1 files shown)


tests/unittests/gradCheckTest.cpp
```diff
@@ -169,36 +169,33 @@ TEST_P(InterpreterGrad, gradientCheckFCConcatRELU) {
...
   auto *A =
-      mod.createVariable(ElemKind::FloatTy, {1, numDim, numDim, numInputChan},
+      mod.createVariable(ElemKind::FloatTy, {1, numDim, numDim, depth},
                          "A", VisibilityKind::Public, false);
-  auto *Ex = mod.createVariable(ElemKind::FloatTy, {1, numOutputElem}, "exp",
+  auto *Ex = mod.createVariable(ElemKind::FloatTy,
+                                {1, numDim + 1, numDim + 1, depth}, "exp",
                                 VisibilityKind::Public, false);
 
-  Node *O = F->createConv("conv", A, 4, 5, 1, 2, group);
```
rdzhabarov: should we also fix: createAvgPool and other places?

artemrakhov: Yes. It is good idea to iterate over all grads covered by the test, and check that if grad is not doing anything, then test is indeed failing.

nadavrot:
Impressive! Thank you Artem!

PR #1448, opened 2018-08-11
[OpenCL] Assert that we don't run into Group Convolution.

artemrakhov:


include/glow/Backends/LayoutConverter.h
```diff
@@ -25,6 +25,9 @@ namespace glow {
 /// convolution nodes using NCHW.
 template <class NCHWConvNode>
 Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
+  assert(CN->getGroup() == 1 &&
```
jfix71: This header is backend independent and meant to be shared, so the assert shouldn't go here. I think the other asserts you added are sufficient.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -950,6 +953,9 @@ void OpenCLFunction::execute() {
     }
 
     if (auto *CG = dyn_cast<ConvolutionGradInst>(&I)) {
+      assert(CG->getGroup() == 1 &&
+        "Group Convolution is not supported by OpenCL backend.");
```
qcolombet: Should we use GLOW_ASSERTs instead?

include/glow/Backends/LayoutConverter.h
artemrakhov: Well, this is a place where we lose group parameter. It was present in initial node, but not in resulting node. So I think it's a good idea to have an assertion. I changed the text not to mention OpenCL.

lib/Backends/OpenCL/OpenCL.cpp
artemrakhov: No, I think we usually keep going if dimensions don't match or something minor like this is wrong. We want as less checks as possible in release mode, right?

include/glow/Backends/LayoutConverter.h
jfix71: Ah ok got it. Perhaps we should support the group parameter in this function then, since other backends/users of this function might support it.

E.g. we could add group to the `OCLConvolutionNode`, and assert it's 1 when constructing the input to the OCL conv kernel.

Or we could pass another parameter to `convertConvToNCHWConv()` to signify if the group parameter should be included in the constructor for the new node. Then in `OCLBackend::transformPostLowering()` we could put this assertion that `group == 1`, and pass false for this parameter.

opti-mix:
@beicy Could you remind me why we cannot support group convolutions in the OpenCL backend? IIRC, the OCLConvolution kernels support group convolutions already. We just need to pass the parameters down, i.e. we would need to add group to the OCLConvolutionNode, etc just like we did it for non-square strides and kernels. Or do I forget something and there were issues with adding group convolutions support?

include/glow/Backends/LayoutConverter.h
artemrakhov: @jfix71 : it makes sense to support Group Convolution both for NCHW layout and OpenCL, and any other layout where it is possible. The purpose of this PR is to add assert ASAP while support is not yet there :) It's much better than seeing wrong results as of now.

PR #1451, opened 2018-08-13
[ONNXIFI] Fix incorrect return value in loadWeight

yinghai:
Simple bug. This will get us through the initGraph phase. 

@rdzhabarov


yinghai:
Looks like the CI has some infra errors. apt-get failed.

PR #1453, opened 2018-08-13
[onnxifi] Load output nodes, perform setIO and run the Graph.

rdzhabarov:
towards issue: https://github.com/pytorch/glow/issues/1190
* Set output nodes
* Handle set graph IO ONNXIFI method
* Perform inference as part of the Run() method

cc: @yinghai @Maratyszcza 

running ```pytest -s test_onnxifi.py```, it passes e2e for ReLU and Conv nets.

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -21,6 +21,7 @@
 #include "glow/Graph/Graph.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringMap.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/raw_ostream.h"
 
@@ -99,7 +100,7 @@ protected:
   /// A list of weight tensors indexed by name.
   std::unordered_map<std::string, Tensor *> tensors_;
   /// A map from names of the external outputs of the network to SaveNodes.
-  std::unordered_map<std::string, SaveNode *> outputsByName_;
+  llvm::StringMap<SaveNode *> outputsByName_;
 
   /// \returns the tensor that was registered under the name \p name.
   Tensor *getTensorByName(llvm::StringRef name);
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -29,7 +29,6 @@
 #include <cstddef>
 #include <cstdint>
 #include <fstream>
-#include <iostream>
 #include <sstream>
 #include <string>
 #include <vector>
@@ -457,7 +456,7 @@ bool ONNXModelLoader::setOutputNodes(ONNX_NAMESPACE::GraphProto &net) {
   }
 
   for (int i = 0; i < net.output_size(); i++) {
-    auto &outputName = net.output(i).name();
+    const auto &outputName = net.output(i).name();
     auto r = getNodeValueByName(outputName);
     outputsByName_[outputName] = G_.createSave("save_" + outputName, r);
   }
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -451,14 +451,18 @@ void ONNXModelLoader::loadInitializers(ONNX_NAMESPACE::GraphProto &net) {
   }
 }
 
-void ONNXModelLoader::setOutputNodes(ONNX_NAMESPACE::GraphProto &net) {
-  assert(net.output_size() && "Network needs external outputs defined.");
+bool ONNXModelLoader::setOutputNodes(ONNX_NAMESPACE::GraphProto &net) {
+  if (net.output_size() == 0) {
+    return false;
+  }
 
   for (int i = 0; i < net.output_size(); i++) {
     auto &outputName = net.output(i).name();
     auto r = getNodeValueByName(outputName);
     outputsByName_[outputName] = G_.createSave("save_" + outputName, r);
   }
+
+  return true;
 }
 
 bool ONNXModelLoader::loadNetwork(ONNX_NAMESPACE::GraphProto &net) {
@@ -520,7 +524,11 @@ ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
   checkInputs(modelDef, tensorNames, tensors);
 
   loadInitializers(modelDef);
-  if (loadNetwork(modelDef)) {
-    setOutputNodes(modelDef);
+  if (!loadNetwork(modelDef)) {
+    GLOW_ASSERT("Cannot load the model.");
+  }
+
+  if (!setOutputNodes(modelDef)) {
+    GLOW_ASSERT("Cannot load external outputs.");
   }
 }
--- a/lib/Importer/ONNXIFILoader.cpp
+++ b/lib/Importer/ONNXIFILoader.cpp
@@ -118,6 +118,10 @@ std::unique_ptr<ModelLoader> ModelLoader::parse(
   if (!loader->loadNetwork(modelDef)) {
     return nullptr;
   }
+  
+  if (!loader->setOutputNodes(modelDef)) {
+    return nullptr;
+  }
 
   return loader;
 }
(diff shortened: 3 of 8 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -48,22 +48,64 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
...
+
+  // TODO: copy outputs to the addresses specified in the outputs_.
+  return ONNXIFI_STATUS_SUCCESS;
+}
+
+static onnxStatus
+processDescriptors(uint32_t count, const onnxTensorDescriptorV1 *descriptors,
+                   std::unordered_map<glow::Variable *, onnxPointer> &mapping,
+                   const ModelLoader &loader) {
+  for (unsigned i = 0; i < count; ++i) {
```
yinghai: It's OK for now. As a TODO, we should check the dimensions of input/output tensors and track whether it has changed.

rdzhabarov:
Push formatting changes here: https://github.com/pytorch/glow/pull/1455

lib/Importer/ONNX.cpp
```diff
@@ -504,7 +510,11 @@ ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
...
 
   loadInitializers(modelDef);
-  if (loadNetwork(modelDef)) {
-    setOutputNodes(modelDef);
+  if (!loadNetwork(modelDef)) {
+    GLOW_ASSERT("Cannot load the model.");
+  }
+
+  if (!setOutputNodes(modelDef)) {
+    GLOW_ASSERT("Cannot load external outputs.");
```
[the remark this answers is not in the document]
rdzhabarov: ONNX.cpp is not part of the prod flow through the ONNXIFI. But I do agree we should make error handling better and more detailed.

rdzhabarov:
@qcolombet thanks for reviewing, I'm still iterating on this PR. Appreciate comments!

lib/Importer/ONNXIFILoader.cpp
```diff
@@ -18,6 +18,8 @@
 
 #include "onnx/onnx.pb.h"
 
+#include <iostream>
```
nadavrot: I know that this is WIP, I just want to point out that in most parts of the code we use llvm::outs() or llvm::errs().

rdzhabarov: yup, those std::cout will go away, just using that for debugging. Will change for llvm::outs() for current debugging.

lib/Onnxifi/Base.cpp
```diff
@@ -48,22 +48,84 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
...
+  }
+
+  return ONNXIFI_STATUS_SUCCESS;
+}
 
 onnxStatus Graph::setIO(uint32_t inputsCount,
                         const onnxTensorDescriptorV1 *inputDescriptors,
                         uint32_t outputsCount,
                         const onnxTensorDescriptorV1 *outputDescriptors) {
+  // Process inputs.
```
Maratyszcza: You should check tags before you access any members in `inputDescriptors` or `outputDescriptors` and return `ONNXIFI_STATUS_UNSUPPORTED_TAG` if tag does not match `ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1`. Because if `tag` is not `ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1`, `inputDescriptors[i]` is not a `onnxTensorDescriptorV1`, and even reading any member may SEGFAULT.

yinghai:
OK. This PR has developed into more than the title says. @rdzhabarov, let us know when this is ready for review.

rdzhabarov:
Yes, @yinghai i'll ping you. WIP for now.

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -302,12 +302,32 @@ ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI onnxInitGraph(
   return ONNXIFI_STATUS_SUCCESS;
 }
 
+static bool verifyDescriptors(uint32_t count,
+                              const onnxTensorDescriptorV1 *descriptors) {
+  for (unsigned i = 0; i < count; i++) {
+    const auto &descriptor = descriptors[i];
+    /*if (descriptor.tag != ONNXIFI_TAG_TENSOR_DESCRIPTOR_V1) {
```
rdzhabarov: @yinghai seems like C2 does not set the tag here properly on onnxSetGraphIO (i have to comment out this piece of code in order to process descriptors). descriotpr.tag equals to 0 at this point.

yinghai: https://github.com/pytorch/pytorch/pull/10502 fixed it and is already merged.

lib/Onnxifi/Base.cpp
```diff
@@ -48,22 +48,80 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
...
+  for (unsigned i = 0; i < inputsCount; ++i) {
```
rdzhabarov: @yinghai While running the "conv" ONNXIFI test I noticed that both W and X are passed from the C2 through the ONNXIFI on setGraphIO call. I suppose only X should be passed. (W is a static weights which was initialized as part of the initGraph).

lib/Onnxifi/onnxifiGlow.cpp
rdzhabarov: Nice! Will update it!

lib/Onnxifi/Base.cpp
yinghai: Yeah, that's not right. Let me check.

yinghai: OK. It doesn't break things but I'll figure out a way to prune this.

rdzhabarov: It does not break because I had to explicitly filter out weights from here ;) But yes, it needs to be fixed, not urgent.

yinghai: It might be an artifact of the test: https://github.com/pytorch/pytorch/blob/45e5c17ecfbe00708e2c61488d688b3d651ab8ca/caffe2/python/onnx/test_onnxifi.py#L91

lib/Onnxifi/Base.cpp
```diff
@@ -55,15 +55,74 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
...
+  // Emit IR for the graph and compile it.
+  backendPtr_->getEE().compile(CompilationMode::Infer, function_);
+
   return ONNXIFI_STATUS_SUCCESS;
 }
 
-onnxStatus Graph::run() { return ONNXIFI_STATUS_SUCCESS; }
+onnxStatus Graph::run() {
+  // Copy tensors from the input addresses to the Glow tensors.
+  std::vector<Tensor *> tensors;
```
yinghai: Just curious, when do we use std containers and when do we use llvm containers?

include/glow/Importer/ProtobufLoader.h
```diff
@@ -99,7 +100,7 @@ class ProtobufLoader {
   /// A list of weight tensors indexed by name.
   std::unordered_map<std::string, Tensor *> tensors_;
   /// A map from names of the external outputs of the network to SaveNodes.
-  std::unordered_map<std::string, SaveNode *> outputsByName_;
+  llvm::StringMap<SaveNode *> outputsByName_;
```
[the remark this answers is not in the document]
rdzhabarov: mostly StringMap is more efficient and I'll change `tensors_` and others here in the follow-up.
I'm changing only outputsByName_ as I'm touching that in some changes in this PR.

lib/Onnxifi/Base.cpp
rdzhabarov: llvm one would be preferable, i'll change that to the corresponding vector type.

yinghai: Wait, this one is for external function (RunInference). Maybe std is better.

rdzhabarov: Run inference takes takes llvm types:
```
void ExecutionEngine::run(llvm::ArrayRef<Variable *> vars,
                          llvm::ArrayRef<Tensor *> inputs)
```
which should be fine.

PR #1456, opened 2018-08-13
[NodeValue] Provide methods to access the users of a NodeValue

### LINKED ISSUE #1346: Our def-use chains are counter intuitive (and error-prone IMHO)

qcolombet:
Prior to this patch the only way to get to the users of a node value
was to get through the users of the related node and filtered out
the uses of that did not match the result number.

This feature is now directly accessible through the NodeValueIterator
class and exposed via the new methods NodeValue::getUsers,
NodeValue::getNumUsers, and NodeValue::hasOneUse.

This fixes #1346.

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -19,6 +19,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/ilist.h"
 #include "llvm/ADT/ilist_node.h"
+#include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/Casting.h"
 
 #include "glow/Base/Traits.h"
@@ -33,6 +34,9 @@ class Function;
 class Node;
 class NodeWalker;
 struct NodeUse;
+template <bool is_const_iter> class NodeValueIteratorImpl;
+using NodeValueIterator = NodeValueIteratorImpl<false>;
+using NodeValueConstIterator = NodeValueIteratorImpl<true>;
 
 /// Unlike LLVM values, graph nodes may return multiple values as the result of
 /// a computation. Gradient-calculating nodes such as conv-grad return multiple
@@ -100,6 +104,15 @@ public:
       return resNo_ < O.resNo_;
     return (node_ < O.node_);
   }
+
+  /// Check if this NodeValue has exactly one use.
+  bool hasOneUse() const { return getNumUsers() == 1; }
+  /// Get the number of users of this NodeValue.
+  unsigned getNumUsers() const;
+
+  /// Get the list of users of this NodeValue.
+  llvm::iterator_range<NodeValueIterator> getUsers();
+  llvm::iterator_range<NodeValueConstIterator> getUsers() const;
 };
 
 /// A handle type for a NodeValue. This type should be used only by the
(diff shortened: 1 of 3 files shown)


tests/unittests/graphTest.cpp
```diff
@@ -641,6 +643,19 @@ TEST(Graph, cmpOutputTypes) {
...
 }
 
+/// Check that the users of value are equal to expectedUsers.
+static bool
+hasAllTheseUses(const llvm::SmallPtrSetImpl<const Node *> &expectedUsers,
+                const NodeValue &value) {
+  llvm::SmallPtrSet<const Node *, 4> uses;
+  for (const NodeUse &users : value.getUsers()) {
+    const Node *user = users.getUser();
+    EXPECT_TRUE(expectedUsers.count(user));
```
jfix71: It's a little funny that this function could still return true even if there's a failure here. Maybe you want to just return false here instead of `EXPECT_TRUE()`? Or you could turn this into a `void` function and `EXPECT_TRUE()` on the size comparison at the bottom here.

include/glow/Graph/Node.h
```diff
@@ -352,6 +365,84 @@ class NodeWalker {
...
+  NodeValueIteratorImpl(NodeValueTy &parent, iterator it)
+      : parent_(parent), it_(it) {
+    if (it_ != getEnd() && !hasSameParent()) {
+      ++(*this);
+    }
+    assert((it_ == getEnd() || hasSameParent()) &&
+           "operator++ should return the next valid iterator");
+  }
+
+  const NodeValueTy &getParent() const { return parent_; }
```
jfix71: This doesn't seem to be used -- should it be removed, or be made private? Otherwise I think it should be included in the unit test.

qcolombet: Good catch, I meant to test it :)

tests/unittests/graphTest.cpp
qcolombet: Returning false would be the right thing to do.
I don't know what I was thinking.

qcolombet:
Thanks @jfix71 for the review! I'll add a test case for node with only one result.

include/glow/Graph/Node.h
qcolombet: Finally I decided to drop it. I meant to add some asserts in the test, but I actually already did the checks I wanted within the class (::hasSameParent).

tests/unittests/graphTest.cpp
qcolombet: Done.

qcolombet:
Test added for single result nodes.

nadavrot:
Nice!

Issue #1441, opened 2018-08-10
vgg19: Some FP weights are not used after quantization anymore, but they are not removed


opti-mix:
Sometimes, when an NN model is quantized, the FP weights are replaced by Int8 weights. But it looks like the old FP weights are not removed, even though they are not used.

![vgg19_unused_weights](https://user-images.githubusercontent.com/33707249/43972189-6df0e412-9c88-11e8-9f27-325ca07d5f89.png)

Most likely the DCE pass in the GraphOptimizer.cpp decides to keep these weights around for some reason (e.g. they are public, etc).

This issue needs to be fixes, because keeping these weights around increases the amount of memory required to run a quantized model, whereas one of the main goals of quantization is to reduce the memory consumption and the amount of data transfers.

opti-mix:
The issue could be fixed either in DCE or in the backends, but fixing it in DCE would be preferable.

The first step would be to debug when these weights become unused and why they are not removed by the DCE.

ZchiPitt:
done

PR #1458, opened 2018-08-13
[fix][quantization]Erase the old function after we create the new quantized function

ZchiPitt:
This pr is due to the [issue](https://github.com/pytorch/glow/issues/1441) that we found some FP weights are not used after quantization anymore but they are not removed. 
The reason behind this bug, is that when we perform quantization on an existing function, we create a new quantized function, but we didn't remove the old, which still uses some non-quantized useless variables.
We propose the fix here to remove the old function after quantizing it, and it successfully helped the DCE remove those unused variables.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -432,6 +432,9 @@ quantizeFunction(const ExecutionEngine &EE,
     }
   } while (nodeIt != stopIt);
 
+  // Erase the original function so that the redundant variables that are only 
+  // referenced by the original function will be removed.
+  F->getParent()->eraseFunction(F);
   return G;
 }
 
--- a/tests/unittests/MLTest.cpp
+++ b/tests/unittests/MLTest.cpp
@@ -908,7 +908,8 @@ TEST_P(InterpreterAndCPU, convNetForImageRecognition) {
   auto *TANH = F->createTanh("tanh", CV);
   auto *FCL = F->createFullyConnected("fc", TANH, 2);
   auto *SM = F->createSoftMax("sm", FCL, ex);
-  auto *result = F->createSave("ret", SM);
+  auto *SN = F->createSave("ret", SM);
+  auto *resultVar = SN->getVariable();
   Function *TF = glow::differentiate(F, TC);
   EE.compile(CompilationMode::Train, TF);
 
@@ -938,7 +939,7 @@ TEST_P(InterpreterAndCPU, convNetForImageRecognition) {
   Tensor testLabels(ElemKind::IndexTy, {batchSize, 1});
   generateImageData(testImages, testLabels, mod.getPRNG());
   EE.run({input}, {&testImages});
-  auto SMH = result->getVariable()->getHandle<>();
+  auto SMH = resultVar->getHandle<>();
   for (size_t i = 0; i < batchSize; i++) {
     bool isLine = testLabels.getHandle<size_t>().at({i, 0}) == 0;
     auto lineWeight = SMH.at({i, 0});
@@ -995,7 +996,8 @@ TEST_P(InterpreterAndCPU, testFindPixelRegression) {
   auto *RL0 = F->createRELU("relu0", FC0);
   auto *FC1 = F->createFullyConnected("fc1", RL0, 2);
   auto *R = F->createRegression("regression", FC1, ex);
-  auto *result = F->createSave("ret", R);
+  auto *SN = F->createSave("ret", R);
+  auto *resultVar = SN->getVariable();
   Function *TF = glow::differentiate(F, TC);
   EE.compile(CompilationMode::Train, TF);
 
@@ -1038,7 +1040,7 @@ TEST_P(InterpreterAndCPU, testFindPixelRegression) {
   EE.run({input}, {&testImages});
 
   // A handle to the projected result.
-  auto RH = result->getVariable()->getHandle<>();
+  auto RH = resultVar->getHandle<>();
   // A handle to the true label.
   auto LH = testLabels.getHandle<>();
 
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -387,6 +387,26 @@ TEST(Graph, NodeValue) {
       24);
 }
 
+/// Check that by deleting one function, the variables that refernced
+/// by this function, will reduce its number of uses by one.
+TEST(Graph, deleteFunction) {
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  Function *F1 = mod.createFunction("f1");
+  auto *inputX = mod.createVariable(ElemKind::FloatTy, {1}, "input",
+                                    VisibilityKind::Public, true);
+  F1->createLog("log1", inputX);
+  Function *F2 = mod.createFunction("f2");
+  F2->createLog("log2", inputX);
+  // We check the number of user of inputX to be 2 as only F1 and F2 are 
+  // using it.
+  EXPECT_EQ(inputX->getNumUsers(), 2);
+  // Erase this function here to see if we can see the number of user of inputX
+  // reduce to 1.
+  mod.eraseFunction(F1);
+  EXPECT_EQ(inputX->getNumUsers(), 1);
+}
+
 TEST(Graph, nodesWithPredicates) {
   ExecutionEngine EE;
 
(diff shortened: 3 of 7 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -432,6 +432,7 @@ quantizeFunction(const ExecutionEngine &EE,
     }
   } while (nodeIt != stopIt);
 
+  F->getParent()->eraseFunction(F);
```
opti-mix: Please add a comment here explaining why we erase the old function.

tests/unittests/MLTest.cpp
```diff
@@ -908,7 +908,7 @@ TEST_P(InterpreterAndCPU, convNetForImageRecognition) {
   auto *TANH = F->createTanh("tanh", CV);
   auto *FCL = F->createFullyConnected("fc", TANH, 2);
   auto *SM = F->createSoftMax("sm", FCL, ex);
-  auto *result = F->createSave("ret", SM);
+  auto *resultVar = F->createSave("ret", SM)->getVariable();
```
opti-mix: Can you do it in two lines like this?
```c++
auto *SN = F->createSave("ret", SM);
auto *resultVar = SN->getVariable();
```
Otherwise this lines does a bit too much for one line ;-)

tests/unittests/MLTest.cpp
```diff
@@ -995,7 +995,7 @@ TEST_P(InterpreterAndCPU, testFindPixelRegression) {
   auto *RL0 = F->createRELU("relu0", FC0);
   auto *FC1 = F->createFullyConnected("fc1", RL0, 2);
   auto *R = F->createRegression("regression", FC1, ex);
-  auto *result = F->createSave("ret", R);
+  auto *resultVar = F->createSave("ret", R)->getVariable();
```
opti-mix: Same here. Please use 2 lines.

tests/unittests/graphTest.cpp
```diff
@@ -387,6 +387,20 @@ TEST(Graph, NodeValue) {
...
+TEST(Graph, varWithTwoFunctions) {
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  Function *F1 = mod.createFunction("f1");
+  auto *inputX = mod.createVariable(ElemKind::FloatTy, {1}, "input",
+                                    VisibilityKind::Public, true);
+  F1->createLog("log1", inputX);
+  Function *F2 = mod.createFunction("f2");
+  F2->createLog("log2", inputX);
+  EXPECT_EQ(inputX->getNumUsers(), 2);
```
opti-mix: Add a comment explaining what we check and why we expect 2 uses.

tests/unittests/graphTest.cpp
```diff
@@ -387,6 +387,20 @@ TEST(Graph, NodeValue) {
...
+  mod.eraseFunction(F1);
```
opti-mix: Explain the intent of the removal and what we expect afterwards.

tests/unittests/graphTest.cpp
```diff
@@ -387,6 +387,20 @@ TEST(Graph, NodeValue) {
       24);
 }
 
+TEST(Graph, varWithTwoFunctions) {
+  ExecutionEngine EE;
```
opti-mix: Explain the idea of the test

tests/unittests/quantizationTest.cpp
```diff
@@ -201,7 +201,8 @@ TEST_P(Operator, end2end) {
   // STEP1 - Generate the first network to record the quantization parameters.
   Function *F1 = createSimpleGraphForQuantization(mod, A, B, "main");
   Function *F2 = F1->clone("main2");
-  SaveNode *result1 = cast<SaveNode>(F1->getNodeByName("save"));
+  Variable *result1Var =
+      cast<SaveNode>(F1->getNodeByName("save"))->getVariable();
```
opti-mix: Use 2 lines.

tests/unittests/quantizationTest.cpp
```diff
@@ -214,15 +215,16 @@ TEST_P(Operator, end2end) {
       quantization::generateNodeQuantizationInfos(F1);
 
   // STEP2 - Use the profile to quantize a network.
-  SaveNode *result2 = cast<SaveNode>(F2->getNodeByName("save"));
+  Variable *result2Var =
+      cast<SaveNode>(F2->getNodeByName("save"))->getVariable();
```
opti-mix: Same.

tests/unittests/quantizationTest.cpp
```diff
@@ -336,7 +338,8 @@ TEST_P(Quantization, end2endGRU) {
   auto *mod = &interpreterEE.getModule();
   Function *F1 = createGRUForQuantization(mod, "main");
   Function *F2 = F1->clone("main2");
-  SaveNode *result1 = cast<SaveNode>(F1->getNodeByName("save"));
+  Variable *result1Var =
+      cast<SaveNode>(F1->getNodeByName("save"))->getVariable();
```
opti-mix: Same.

tests/unittests/quantizationTest.cpp
```diff
@@ -349,15 +352,16 @@ TEST_P(Quantization, end2endGRU) {
       quantization::generateNodeQuantizationInfos(F1);
 
   // STEP2 - Use the profile to quantize a network.
-  SaveNode *result2 = cast<SaveNode>(F2->getNodeByName("save"));
+  Variable *result2Var =
+      cast<SaveNode>(F2->getNodeByName("save"))->getVariable();
```
opti-mix: Same.

tools/loader/ImageClassifier.cpp
```diff
@@ -156,6 +156,8 @@ int main(int argc, char **argv) {
     i1 = LD.getVariableByName("data_0");
   }
 
+  auto *SMVar = SM->getVariable();
```
opti-mix: Add a comment explaining the meaning of SMVar.

tests/unittests/graphTest.cpp
```diff
@@ -387,6 +387,20 @@ TEST(Graph, NodeValue) {
       24);
 }
 
+TEST(Graph, varWithTwoFunctions) {
```
rdzhabarov: maybe something like: deleteFunction

tests/unittests/MLTest.cpp
rdzhabarov: why do we have all these changes? is it anyhow related to the bug_fix?

ZchiPitt: because after erasing the old function, the save nodes that belong to original function also got removed. We need to a way to still be able to reference the variable of the save node

lib/Quantization/Quantization.cpp
ZchiPitt: all done

nadavrot:
@ZchiPitt @opti-mix Why are we deleting the original floating-point function inside the quantization procedure? Won't it be a better API to let the caller decide if it wants to delete the function? For example, what if we decide to write a test that compares two quantization modes or if we want to compare the result of the quantized function to the original function?

ZchiPitt:
sounds good. But i think its the best that by default the original function got deleted after quantization, and the caller can pass some parameter to avoid deleting it

nadavrot:
I don’t think that adding a flag deleteAfter is a good idea. Are you trying to avoid changing unit tests?

ZchiPitt:
cool, will do

qcolombet:
Hi guys,

I just stumbled on some code that this PR changed and I am concerned by the coupling this creates.

Namely, when compiling a function, we now have to know that a `save` to a variable may be deleted and thus we have to grab the related variable before running compile.

Admittedly the previous behavior of assuming that an instruction will survive to get to the variable was bad, but needing to know that one needs to grab the variable before running compile doesn't seem better.

Could we instead feature something to get the outputs of a network?
For starter, we could use getVariableByName instead of grabbing the variable before compile.

What do you think?

jfix71:
@qcolombet I agree -- in ProtobufLoader we have `outputsByName_` which is a map from variable names to SaveNodes. There's no reason for this, as all uses just get the Variable immediately anyway. I'm going to kick off a PR to move it to map from the names to Variables instead.

nadavrot:
@qcolombet This change was reversed by @rdzhabarov. We now don't delete the function inside the quantizer.

rdzhabarov:
For the above comment, see here: https://github.com/pytorch/glow/pull/1485

Issue #1452, opened 2018-08-13
Issue for quantized max pooling in the CPU backend (?)


tlepley-cadence:
My understanding is that in Glow the quantization parameters, scale/offset, are computed per tensor. Inputs and outputs tensors of an operator can then have different scales/offsets, and this needs to be handled by Glow at code generation (LLVM IR generation for the CPU backend).
In AveragePooling, I see that both input and output quantization parameters are used for the LLVM IR generation. 
But for MaxPooling, there is no specific code for the quantized version in `generateLLVMIRForInstr`. This implicitly means that the quantization parameters of the output tensor are assumed to be the same as the input tensor, what I think is not correct.

rdzhabarov:
As of now, the way Quantized MaxPooling is handled in Glow is the following. We assume that the output scale/offset is the same as the input, but add RescaleNode right after MaxPool to make sure that values are compatible with the profiled range (from the MaxPool) we have on the quantization profiling step.

We could allow MaxPool to have different {s,0} for the output param but that's going to be an optimization as in any case we perform conversion to the right range later.

tlepley-cadence:
@rdzhabarov Thanks for the explanation. Is it a general rule to add a `RescaleNode` for  handling the outputs, or there is a operator by operator decision made in Glow. And in this case, it it described somewhere ?

rdzhabarov:
> Is it a general rule to add a RescaleNode for handling the outputs, or is it an operator by operator decision made in Glow

Not really a general rule (and in some cases that could lead to accuracy loss), but for some nodes we've been historically doing that. For MaxPool this is OK, since we work on int8 input numbers and would need to rescale it to a different {s,o} either as part of the operator itself or as a next step anyway (currently processed in a next step in the Rescale node).

> is it described somewhere ?
I do not think there are docs covering this specific case (concrete steps for conversion of fp32 node to int8 node). But you can get a sense of how this is done by looking at https://github.com/pytorch/glow/blob/master/lib/Quantization/Quantization.cpp#L126.
When a node needs to consume the output of MaxPool it'll make sure that {s,o} are set according to profiled data and that procedure will introduce the Rescale node.

rdzhabarov:
@tlepley-cadence any objections to close this issue?

nadavrot:
@rdzhabarov Do we check in the verifier that the input and output integer type of MaxPool match?

rdzhabarov:
[...]

PR #1460, opened 2018-08-14
[verifier] Verify pooling nodes have the same element type and S,O for quantized types.

rdzhabarov:
As discussed in the https://github.com/pytorch/glow/issues/1452 we should check against the expectation of having the same quantization params for pooling nodes.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -99,20 +99,38 @@ std::string Variable::getDebugDesc() const {
 //                       Nodes verification
 //===----------------------------------------------------------------------===//
 
-/// Check that the type of the first operand matches the type of the second
-/// operand.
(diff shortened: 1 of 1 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -111,10 +111,27 @@ static void checkSameShape(NodeValue A, NodeValue B) {
   assert(A.dims() == B.dims() && "Invalid shape");
 }
 
+/// Check that the element type of the operand \p A matches expected type.
 static void checkType(NodeValue A, ElemKind expectedType) {
   assert(A.getElementType() == expectedType && "Invalid type");
 }
 
+/// Check that the type of the first operand matches the type of the second
```
qcolombet: \p A and \p B instead of first operand and second operand?

PR #1461, opened 2018-08-14
Made convolution support the filter as node output (not only as constant tensor)

tlepley-cadence:
We've got some ONNX models converted from TensorFlow where the convolution filter tensor is the output of a transpose (related to the conversion of depthwise convolution that is not supported by ONNX). This PR fixes Glow for this use-case.

I also removed the usage of the "kernel_shape" attribute in the ONNX loader since the kernel shape is given by the tensor itself.

### DIFF
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -234,7 +234,7 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
   }
 
   if (typeName == "Conv") {
-    // Load the inputs:
+    // Load the attributes
     std::vector<unsigned_t> strides(2, 1);
     if (dict.count("strides")) {
       strides = getShape<unsigned_t>(dict.at("strides"));
@@ -243,29 +243,37 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     // Pads : {pad_top, pad_left, pad_bottom, pad_right}
     std::vector<unsigned_t> pads = getPads(dict);
 
-    auto in = getNodeValueOrCreateVariableByName(op.input(0));
-    Tensor *w = getTensorByName(op.input(1));
+    // Load the inputs
+    NodeValue in = getNodeValueOrCreateVariableByName(op.input(0));
+    NodeValue filterValue = getNodeValueOrCreateVariableByName(op.input(1));
 
-    // Transpose the weights to the right format. Glow expects to read the
+    // Transpose the filter to the right format. Glow expects to read the
     // weights in the format CRSK. ONNX stores the operators as KCRS.
     // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
-    Tensor wtag;
-    w->transpose(&wtag, NCHW2NHWC);
+    TransposeNode *filterTransposeNode =
+        G_.createTranspose(opName, filterValue, NCHW2NHWC);
 
     // The structure of the conv weigts is: NHWC. We take the C, which is the
     // number of filters. We use this value to calculate the size of the bias
     // if it is not specified.
-    size_t depth = wtag.dims()[0];
+    const NodeValue filterTransposedValue = filterTransposeNode->getResult();
+    size_t depth = filterTransposedValue.dims()[0];
 
-    // Construct the Filter field.
-    auto *filter = G_.getParent()->createVariable("conv.filter", wtag);
+    // Get the kernel shape from the input.
+    std::vector<unsigned_t> kernelShape(2);
+    kernelShape[0] = filterTransposedValue.dims()[1];
+    kernelShape[1] = filterTransposedValue.dims()[2];
 
-    std::vector<unsigned_t> kernels(2);
+    // Extra check when the 'kernel_shape' attribute exists.
+    // The 'kernel_shape' attribute is redundant not mandatory.
     if (dict.count("kernel_shape")) {
-      kernels = getShape<unsigned_t>(dict.at("kernel_shape"));
-    } else {
-      kernels[0] = filter->dims()[1];
-      kernels[1] = filter->dims()[2];
+      std::vector<unsigned_t> kernelShapeAttribute =
+          getShape<unsigned_t>(dict.at("kernel_shape"));
+      assert(kernelShape[0] == kernelShapeAttribute[0] &&
+             kernelShape[1] == kernelShapeAttribute[1] &&
+             "The 'kernel_shape' attribute is not consistent with the actual "
+             "convolution kernel shape.");
+      (void)kernelShapeAttribute; // Avoids compilation warning in release mode.
     }
 
     // Construct the Bias field.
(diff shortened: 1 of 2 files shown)


lib/Importer/ONNX.cpp
```diff
@@ -243,29 +243,50 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
-    // number of filters. We use this value to calculate the size of the bias
-    // if it is not specified.
-    size_t depth = wtag.dims()[0];
-
-    // Construct the Filter field.
-    auto *filter = G_.getParent()->createVariable("conv.filter", wtag);
-
+    Node *filter_transpose_node = NULL;
     std::vector<unsigned> kernels(2);
-    if (dict.count("kernel_shape")) {
```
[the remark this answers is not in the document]
tlepley-cadence: Sure, I'll do that. I'm just wondering why ONNX has this `kernel_shape` that looks not really needed to me.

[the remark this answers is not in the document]
artemrakhov: I believe this is present only for historical reasons. And it's better for us to have more assertions.

lib/Importer/ONNX.cpp
```diff
@@ -243,29 +243,50 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
-      kernels = getShape<unsigned>(dict.at("kernel_shape"));
+    size_t depth;
+
+    if (hasTensorWithName(op.input(1))) {
```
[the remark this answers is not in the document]
tlepley-cadence: We could indeed use `getNodeValueOrCreateVariableByName`, but I think that currently, we still need to differentiate both cases.
1- If the kernel is a constant tensor, we statically transpose the tensor and create a new variable (we don't add any transpose node)
2- if it's not a constant tensor, we add a transpose node

If we had an optimization that optimize transpose of constants, we could get rid of the `if` condition and always create a transpose node when importing a convolution operator. But I assume that for now, it's preferable to keep this `if` condition. What do you think ?

lib/Importer/ONNX.cpp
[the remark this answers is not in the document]
tlepley-cadence: I added the extra check.

lib/Importer/ONNX.cpp
```diff
@@ -243,29 +243,58 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
 
-    std::vector<unsigned_t> kernels(2);
+    // Extra check when the 'kernel_shape' attribute exists.
+    // The 'kernel_shape' attribute is redundant not mandatory.
     if (dict.count("kernel_shape")) {
-      kernels = getShape<unsigned_t>(dict.at("kernel_shape"));
-    } else {
-      kernels[0] = filter->dims()[1];
-      kernels[1] = filter->dims()[2];
+      std::vector<unsigned_t> kernelShapeAttribute(2);
```
artemrakhov: nit: Effect of initialization with `(2)` gets immediately overridden by following assignment. You can do simply:
`auto kernelShapeAttribute = getShape<unsigned_t>(dict.at("kernel_shape"));`
Note, that when compiling in release, asserts are eliminated. So `kernelShapeAttribute` will be considered unused and we'll see compiler warning. To defeat this, we add `(void)kernelShapeAttribute;`

lib/Importer/ONNX.cpp
```diff
@@ -243,29 +243,58 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+      // The filter is generated by a node (example with depthwise
+      // convolution). Transpose it to meet the Glow expectation.
+      auto filter_node = getNodeValueOrCreateVariableByName(op.input(1));
+      filterTransposeNode = G_.createTranspose(opName, filter_node, NCHW2NHWC);
 
-    // The structure of the conv weigts is: NHWC. We take the C, which is the
-    // number of filters. We use this value to calculate the size of the bias
-    // if it is not specified.
-    size_t depth = wtag.dims()[0];
+      const NodeValue filterValue = filterTransposeNode->getNthResult(0);
```
rdzhabarov: nit: filterTransposeNode->getResult() should make it.

lib/Importer/ONNX.cpp
```diff
@@ -243,29 +243,58 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
-    // Transpose the weights to the right format. Glow expects to read the
-    // weights in the format CRSK. ONNX stores the operators as KCRS.
-    // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
-    Tensor wtag;
-    w->transpose(&wtag, NCHW2NHWC);
+    Node *filterTransposeNode = NULL;
+    std::vector<unsigned_t> kernelShape(2);
+    size_t depth;
+
+    if (hasTensorWithName(op.input(1))) {
```
rdzhabarov: maybe isStaticWeight or something like that?

rdzhabarov: I was just saying that we'd need a better name for this method which reflects its intent. If there is a way to avoid this call I'm all for it.

jfix71: @artemrakhov The fact that it's a static weight is used in other places already -- e.g. in the Conv Caffe2 importer, `getTensorByName()` is used to get the filter, and always transposes it immediately before creating a Variable for it. I agree though that we shouldn't do the transpose there, and instead should create a Transpose node, as the optimizer will do this static transposing anyway in `optimizeTranspose()`. And as far as I know (?) there's also no guarantee that the filter is statically known/loaded -- if so then it's theoretically possible we could load a proto that this would break, where the filter is calculated from another node.

However, it is important to understand if a tensor is static in other places. E.g. in `loadReshape()` we use `getTensorByName()` to load such a static tensor to use as the Reshape's shape, and so it's important that this tensor is static.

So I agree with @rdzhabarov on the renaming -- I think at least `getTensorByName()` should have its name changed, but whatever it ends up being named, it and `hasTensorWithName()` should match.

jfix71: Agreed on everything. That is what I meant -- we should add a transpose Node here and in the C2 loader as well, instead of doing the transpose directly on the static tensor. It would fix this potential issue.

lib/Importer/ONNX.cpp
tlepley-cadence: `filterTransposeNode` was declared as `Node` and only has `getNthResult`. Declaring it as `TransposeNode` or with `auto` enables `getResult`.

tlepley-cadence:
Since there is an optimization to optimize out Transpose nodes when it takes a constant as input, I followed the general consensus and used `getNodeValueOrCreateVariableByName`, what simplifies the importer code. 

Note, that this particular PR is about fixing the ONNX importer that fails with some of our models and my preference would be to handle Caffe2 in a separate PR.

lib/Importer/ONNX.cpp
tlepley-cadence: Thanks for the feedback. I preferred the explicit typing and used `std::vector<unsigned_t> kernelShapeAttribute = getShape<unsigned_t>(dict.at("kernel_shape"));`. I also added the suggested statement to avoid compilation warnings in release mode.

PR #1462, opened 2018-08-14
[Optimizer] Remove single identity transpose from the graph

tlepley-cadence:
We've got some ONNX models with an 'identity' transpose. It was not automatically removed by Glow. This PR removes the transpose node in such a case.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -108,6 +108,17 @@ static void DCE(Function *F) {
   }
 }
 
+/// \returns true if the \p shuffle corresponds to an identity operation, false
+/// otherwise.
+static bool isIdentityShuffle(llvm::ArrayRef<unsigned> shuffle) {
+  for (size_t i = 0, e = shuffle.size(); i < e; i++) {
+    if (shuffle[i] != i) {
+      return false;
+    }
+  }
+  return true;
+}
+
 /// \returns true if the masks \p shuffle1 and shuffle2 are
 /// the inverse of on another. Applying both masks should result in the identity
 /// shuffle.
@@ -289,6 +300,17 @@ static bool sinkCode(Function *F) {
       TN->getResult().replaceAllUsesOfWith(newTR);
       changed = true;
       continue;
+    }
+
+    // Remove 'identity' transpose operations.
+    if (auto *TR = dyn_cast<TransposeNode>(node)) {
+      auto mask = TR->getShuffle();
+
+      if (isIdentityShuffle(mask)) {
+        TR->getResult().replaceAllUsesOfWith(TR->getInput());
+        changed = true;
+        continue;
+      }
     }
 
     // Merge consecutive Transpose operations.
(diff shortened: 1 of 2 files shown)


tlepley-cadence:
@jfix71 Sure. Please note that tomorrow is public holidays in France, so I'll work on this Thursday.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -108,6 +108,18 @@ static void DCE(Function *F) {
   }
 }
 
+/// \returns true if the \p shuffle corresponds to an identity operation, false
+/// otherwise.
+static bool isIdentityShuffle(llvm::ArrayRef<unsigned> shuffle) {
+  // Check if the combined masks are the identity mask.
```
rdzhabarov: seems like stale comment ;)

tlepley-cadence: oops :) I just removed it.

tests/unittests/graphOptzTest.cpp
```diff
@@ -289,11 +289,25 @@ TEST_F(GraphOptz, cancelTwoTransposes) {
...
+                                VisibilityKind::Public, false);
+  Node *T = F_->createTranspose("transpose", A, {0, 1, 2, 3});
+  Node *K = F_->createRELU("relu", T);
+  F_->createSave("ret", K);
+
+  EXPECT_EQ(F_->getNodes().size(), 3);
+
+  ::glow::optimize(F_, CompilationMode::Infer);
+
+  EXPECT_EQ(F_->getNodes().size(), 2);
```
qcolombet: Could you check the actual shape of the resulting graph?
I.e., K's argument is A.

rdzhabarov: fwiw, we do not follow this good practice in the majority of the existing tests, unfortunately :(

tests/unittests/graphOptzTest.cpp
```diff
@@ -289,11 +289,25 @@ TEST_F(GraphOptz, cancelTwoTransposes) {
   EXPECT_EQ(F_->getNodes().size(), 2);
 }
 
+TEST_F(GraphOptz, sinkIdentityTranspose) {
```
rdzhabarov: nit: there is no actual sink happening here it's just elimination of identity transpose.

nadavrot:
Nice!

tests/unittests/graphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/graphOptzTest.cpp
tlepley-cadence: Renamed `remove`. For me `sink` was equivalent to `remove`. @rdzhabarov what is the actual meaning of  `sink` in the context of Glow ?

rdzhabarov: @tlepley-cadence sorry for delay, I'm OOO today.
sink means moving a node below the other node, check `Sink Transpose below batch RELU nodes.` in the graph optimizer for example.

nadavrot:
Congrats!

PR #1463, opened 2018-08-14
Add unsigned_t and use appropriately it in place of unsigned

jfix71:
Currently `unsigned_t` is aliased to `uint32_t`. I also tested with `uint8_t`, `uint16_t`, and `uint64_t` and everything still worked.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -36,6 +36,12 @@ using TypeRef = const Type *;
 
 constexpr unsigned max_tensor_dimensions = 6;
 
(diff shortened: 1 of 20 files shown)


include/glow/Base/Type.h
```diff
@@ -36,6 +36,8 @@ using TypeRef = const Type *;
 
 constexpr unsigned max_tensor_dimensions = 6;
 
+using unsigned_t = uint32_t;
```
rdzhabarov: can you comment here on when it would be appropriate to use this alias?

jfix71: Done.

PR #1464, opened 2018-08-15
[Operators] Support SigmoidCrossEntropyWithLogits

artemrakhov:


include/glow/Graph/Graph.h
```diff
@@ -268,6 +268,10 @@ class Function final : public Named {
   RegressionNode *createRegression(llvm::StringRef name, NodeValue input,
                                    NodeValue expected);
 
+  SigmoidCrossEntropyWithLogitsNode *
```
jfix71: Please add doxygen comments

artemrakhov: Done.

beicy:
I only found this op in Caffe2. Does ONNX have the corresponding op?

artemrakhov:
@beicy : no, there's no such op in ONNX

PR #1467, opened 2018-08-15
[Node] Allow axes equal to the rank of the tensor for Flatten

qcolombet:
Flatten takes a tensor of dimensions [d0, d1, ... dN] and an axis and
produces a 2D tensor of dimensions:
[d0 x d1 x ... x d_axis-1, d_axis x d_axis+1 x ... x dN]

Prior to this patch we would allow axes between [1; rank), where rank
is N + 1.
However, the caffe2 flatten operator allow to use axes up to rank.
In that case, the flattening consists in squashing all the dimensions
together and only have one row:
[d0 x d1 x ... x dN, 1]

This patch thus now allows axes between [1; rank] for the flatten operator.


qcolombet:
For reference, here is the caffe2 doc:
https://caffe2.ai/docs/operators-catalogue.html#flatten

nadavrot:
Nice!

PR #1470, opened 2018-08-16
[ModelRunner] Make leak sanitizer happy

qcolombet:
When running the model-runner, we don't explicitly delete the loader
that we dynamically allocate. Although the memory will be reclaimed
at the end of the process, this made leak sanitizer to complain.

NFC.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -243,6 +243,9 @@ public:
 
     if (size()) {
       size_t count = size() * type_.getElementSize();
+      // We are allocating memory specifically for this tensor,
+      // thus, it owns it.
+      isUnowned_ = false;
       data_ = reinterpret_cast<char *>(alignedAlloc(count, TensorAlignment));
       zero();
     }
--- a/tests/unittests/tensorsTest.cpp
+++ b/tests/unittests/tensorsTest.cpp
@@ -399,6 +399,34 @@ TEST(Tensor, nonOwnedTensor) {
   H1.dump();
 }
 
+/// Check that we properly take ownership of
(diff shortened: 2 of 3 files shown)


include/glow/Base/Tensor.h
```diff
@@ -243,6 +243,9 @@ class Tensor final {
 
     if (size()) {
       size_t count = size() * type_.getElementSize();
+      // We are allocating memory specifically for this tensor,
+      // thus, it owns it.
+      isUnowned_ = false;
```
rdzhabarov: can you reproduce this in a test so that we have sanitizer to complain about this issue?

qcolombet: Added a test case in the updated PR.

tests/unittests/tensorsTest.cpp
```diff
@@ -399,6 +399,34 @@ TEST(Tensor, nonOwnedTensor) {
...
 
+/// Check that we properly take ownership of
+/// the underlying memory when we reset the tensor
+/// shape. This test used to fail leak sanitizer.
+TEST(Tensor, nonOwnedTensorFollowedByReset) {
+  uint8_t raw_data[] = {0, 0, 0, 0};
+  Type F32Ty(ElemKind::FloatTy, {1});
+
+  // Create an unowned tensor.
+  Tensor T1(raw_data, &F32Ty);
```
jfix71: nit: Since this is `FloatTy`, why aren't you just creating the raw_data as `float raw_data[] = {0.};`?

qcolombet: Good point! I guess I was too attached as representing the memory byte by byte :P

PR #1471, opened 2018-08-16
[Type] Do not allow a dimension of zero

jfix71:
This caught a bug 😃


jfix71:
@qcolombet Yeah, I'll land this and then you should be good to remove it on your current PR, or we could open a new one.

PR #1473, opened 2018-08-17
[Importer] Support AveragedLoss and SigmoidCrossEntropyWithLogits

artemrakhov:


lib/Importer/Caffe2.cpp
```diff
@@ -637,7 +653,8 @@ void caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
       break;
     }
     case caffe2::TensorProto_DataType_INT32:
-    case caffe2::TensorProto_DataType_INT64: {
+    case caffe2::TensorProto_DataType_INT64:
+    case caffe2::TensorProto_DataType_BOOL: {
```
jfix71: nit: assuming you checked, but just wanted to make sure that the default value for bool is 0.

PR #1475, opened 2018-08-17
[LLVMIRGen] Make sure path delimiters don't result in empty entry names

qcolombet:
Prior to this patch setting a name ending with '/' would produce an
empty entry name. Combined with the bundle saver, it was possible to
end up with an empty name being appended to the files saved.

E.g.,
$ model-runner -model model -emit-bundle output
Would produce in output:
model.o
model.weights

But
$ model-runner -model model/ -emit-bundle output
Would produce in output:
.o
.weights

Thus, under unix those are hidden files and one could think the bundler
did not work.

This patch makes sure the entry name is never empty by walking back the '/'
until it finds a non-empty name or the beginning of the string.

### DIFF
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -103,10 +103,33 @@ void LLVMIRGen::initTargetMachine(llvm::StringRef T,
 std::string LLVMIRGen::getMainEntryName() const {
   llvm::StringRef name =
       mainEntryName_.empty() ? "main" : F_->getGraph()->getName();
-  auto delimPos = name.rfind('/');
-  if (delimPos != llvm::StringRef::npos)
-    name = name.substr(delimPos + 1);
-  return name;
+
+  // The name may finish with '/' and taking the last token
+  // after '/' will produce an empty name.
+  // If that's the case, move to the next '/' to
+  // try to find a non-empty token.
+  size_t delimPos = name.size();
+  llvm::StringRef entryName;
+  do {
+    auto slashPos = name.rfind('/', delimPos);
+    if (slashPos != llvm::StringRef::npos) {
+      entryName = name.substr(slashPos + 1,
+                              /*length = end - start*/ delimPos - slashPos - 1);
+      delimPos = slashPos;
+    } else {
+      // There is no more slashes, the resulting name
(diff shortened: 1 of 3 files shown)


qcolombet:
Hi @opti-mix,

While I was here I noticed that the behavior of the pair setMainEntryName and getMainEntryName is weird, because getMainEntryName always gives us the name of the underlying graph as soon as setMainEntryName is non-empty.

I wonder if we should have setMainEntryName at all or if we should not return the name of the underlying graph and use what mainEntryName has been set to. In the latter case, I guess we would still need to update the name of the graph given it is supposed to be the name that is being codegen'ed.

nadavrot:
Nice!

opti-mix:
@qcolombet Nice catch!

And yeah, the handling of the bundle names is lame at the moment ;-)
IIRC, the reason for this was to avoid an additional option to explicitly set the name of the entry function. Instead, we tried to derive it from the file name of the provided model. But may be having an explicit command-line option to set the name of the entry point is just the simplest solution?

qcolombet:
> Instead, we tried to derive it from the file name of the provided model. But may be having an explicit command-line option to set the name of the entry point is just the simplest solution?

I'm actually in favor of doing that :).

The thing that is strange to me is the setMainEntryName behavior, which we can address later.

qcolombet:
Is this one okay to land?

opti-mix:
@qcolombet I think this line:
```
 mainEntryName_.empty() ? "main" : F_->getGraph()->getName();
```
should be 
```
 mainEntryName_.empty() ? "main" : mainEntryName_;
```

qcolombet:
> @qcolombet I think this line:

> ` mainEntryName_.empty() ? "main" : F_->getGraph()->getName();`
> should be

> ` mainEntryName_.empty() ? "main" : mainEntryName_;`

Yes, that's what I meant when I was saying:
> The thing that is strange to me is the setMainEntryName behavior, which we can address later.

and

> While I was here I noticed that the behavior of the pair setMainEntryName and getMainEntryName is weird, because getMainEntryName always gives us the name of the underlying graph as soon as setMainEntryName is non-empty.

Thanks for confirming that wasn't intentional!

qcolombet:
Abandoning this one. @opti-mix convinced me that having a separate option for the filename was better.

PR #1476, opened 2018-08-17
[CPU] Add a relocation model option to be able to produce PIC object

qcolombet:
Not sure about the right way to plumb this up, but options works well.
Ideas for tests would be nice.


nadavrot:
LGTM!

PR #1477, opened 2018-08-17
[openCL] add group param to conv for OpenCL backend

beicy:
OpenCL fast kernel for group (group > 1) convolution has some issues which are need to be investigated further. Now for group conv, we only call the slow kernel.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -399,6 +399,14 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   PaddingTLBR pads(CC->getPads());
   ShapeHW kdim(CC->getKernels());
   ShapeHW sdim(CC->getStrides());
+  unsigned group = CC->getGroup();
+  // So far, we don't support fast convolution kernel if group > 1.
+  // For group convolution, the slow convolution kernel should be invoked.
+  // The following assertion should be removed once the group > 1 is supported
+  // in fast convolution kernel.
+  assert(group == 1 && "Group Convolution is not supported by OpenCL backend's "
+                       "fast convolution kernel.");
+
   // Create options for compiling the program.
   // Don't use names M, N, K as they are defined in precompiled headers.
 
@@ -406,7 +414,7 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   // Number of spacial axes.
   addIntOption(options, "v_nax", 2);
   // Number of groups.
-  addIntOption(options, "v_g", 1);
+  addIntOption(options, "v_g", group);
   // Parameters for kernel size, padding and stride
   addIntOption(options, "v_k_0", kdim.height);
   addIntOption(options, "v_k_1", kdim.width);
@@ -501,7 +509,6 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   }
 
   // Compute proper parameters for global work and workgroups.
-  int group = 1;
   auto fw_wgs0 = wg_size[0];
   auto fw_wgs1 = wg_size[1];
   int fw_wptn = WPTN;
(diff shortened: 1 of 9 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -525,9 +525,9 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
 
   // The global work for the N dimension (which covers all pixels) should be big
   // enough to cover all outputs for the correctness.
-  if (global[0] * fw_div_N < odim.h * odim.w) {
-    global[0] = (odim.h * odim.w - 1) / fw_div_N + 1;
-  }
+  //  if (global[0] * fw_div_N < odim.h * odim.w) {
+  //  global[0] = (odim.h * odim.w - 1) / fw_div_N + 1;
+  //}
```
opti-mix: Can you add this fix in a separate PR or at least in a separate commit? It is not related to group convolutions at all.

lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -36,6 +36,11 @@ bool OCLBackend::transformPostLowering(Function *F,
   bool changed = false;
   for (auto &node : F->getNodes()) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
+      // TODO: OpenCL fast kernel itself has some issue with group > 1,
```
opti-mix: fast kernel -> fast convolution kernel

lib/Backends/OpenCL/OpenCL.cpp
beicy: OK. Will revert it in this PR and create a new one.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -205,6 +205,13 @@ class OCLBackend final : public Backend {
     }
     return true;
   };
+
+  bool shouldLower(const Node *N) const override {
+    if (N->getKind() == Kinded::Kind::ConvolutionNodeKind)
```
opti-mix: Please add a comment like: "OpenCL backend can handle any kind of convolutions, including group convolutions" or something like that.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -399,14 +399,15 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
...
+  unsigned group = CC->getGroup();
   // Create options for compiling the program.
   // Don't use names M, N, K as they are defined in precompiled headers.
 
   std::vector<std::string> options;
   // Number of spacial axes.
   addIntOption(options, "v_nax", 2);
   // Number of groups.
-  addIntOption(options, "v_g", 1);
+  addIntOption(options, "v_g", group);
```
opti-mix: Add an assert that group is not > 1. And a comment that fast convolution kernels have issues with groups > 1 and that this assert should be removed once these issues are fixed.

beicy: This assertion has been added at the front of this function.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -755,11 +755,12 @@ __kernel void softmaxgradW(__global void *mem, cl_uint32_t origDest,
 __kernel void convolutionK(__global float *dest, __global float *src,
                            __global float *filter, __global float *bias,
                            ShapeHW filterSizes, ShapeHW strides,
-                           PaddingTLBR pads, ShapeNHWC odim, ShapeNHWC idim,
+                           PaddingTLBR pads, cl_uint32_t group, ShapeNHWC odim, ShapeNHWC idim,
                            ShapeNHWC filterDim) {
   size_t ax = get_global_id(0);
   size_t ay = get_global_id(1);
   size_t d = get_global_id(2);
+  size_t offset = d / (odim.c / group) * (idim.c / group);
```
artemrakhov: nit: offset is general name, which interferes a bit with scale/offset in Quantization case. How does `inChannelOffset` sound?
Also, let's make code more similar to Interpreter implementation by having `inCperG` and `outCperG`.

lib/Backends/OpenCL/kernels.cl
```diff
@@ -755,11 +755,14 @@ __kernel void softmaxgradW(__global void *mem, cl_uint32_t origDest,
 __kernel void convolutionK(__global float *dest, __global float *src,
                            __global float *filter, __global float *bias,
                            ShapeHW filterSizes, ShapeHW strides,
-                           PaddingTLBR pads, ShapeNHWC odim, ShapeNHWC idim,
+                           PaddingTLBR pads, cl_uint32_t group, ShapeNHWC odim, ShapeNHWC idim,
```
artemrakhov: nit: Line is longer than 80 chars. Apparently clang-format doesn't work for this file, because it has `.cl` extension. You can temporarily rename it to `.cpp` to run clang-format again. But feel free to do this in follow up PR. There may be other format changes, not related to your PR.

PR #1478, opened 2018-08-17
[OpenCL] A small change for fast convolution kernel.

beicy:


opti-mix:
Could you update the commit message. It is not a change. It is a bugfix. We should have had `int N_FW_ = odim.h * odim.w;` from the beginning, but due to copy&paste it was overlooked ;-)

opti-mix:
The kernel is not changed by this commit. It is the invocation of the fast convolution kernel that is changed here.  So, the message should be "A bugfix for the invocation parameters of the fast convolution kernel" ;-)

PR #1480, opened 2018-08-17
Make sure that MemoryAllocator uses wide-enough unsigned types for reresenting the address space being allocated

opti-mix:
The types should be able to represent any address in the address space of the device, whose memory is being allocated by means of the MemoryAllocator. This device may be different from the host running Glow. E.g. Glow may run on a host with a 32-bit address space, but use a  MemoryAllocator for a GPU or an accelerator with a 64-bit address space.

Therefore, to be on the safe side use `uint64_t` instead of `size_t` in MemoryAllocator APIs.

Update the user's of the API to use `uint64_t` instead of `size_t`.

### DIFF
--- a/include/glow/CodeGen/MemoryAllocator.h
+++ b/include/glow/CodeGen/MemoryAllocator.h
@@ -25,17 +25,17 @@ namespace glow {
 class Segment {
 public:
   /// The allocation starts at this address.
-  size_t begin_;
+  uint64_t begin_;
   /// The allocation ends before this address (half-open interval).
-  size_t end_;
+  uint64_t end_;
 
-  Segment(size_t begin, size_t end) : begin_(begin), end_(end) {}
+  Segment(uint64_t begin, uint64_t end) : begin_(begin), end_(end) {}
 
   /// \returns the size of the interval.
-  size_t size() const { return end_ - begin_; }
+  uint64_t size() const { return end_ - begin_; }
 
   /// \returns True if the value \p idx falls within this segment.
-  bool contains(size_t idx) const { return idx >= begin_ && idx < end_; }
+  bool contains(uint64_t idx) const { return idx >= begin_ && idx < end_; }
 };
 
 /// Allocates segments of memory.
@@ -43,15 +43,15 @@ class MemoryAllocator {
   /// A list of live buffers.
   std::list<Segment> allocations_;
   /// The size of the memory region that we can allocate segments into.
-  size_t poolSize_;
+  uint64_t poolSize_;
   /// This is the high water mark for the allocated memory.
-  size_t maxMemoryAllocated_{0};
+  uint64_t maxMemoryAllocated_{0};
 
 public:
   /// A reserved value to mark invalid allocation.
-  static const size_t npos;
+  static const uint64_t npos;
 
-  explicit MemoryAllocator(size_t poolSize) : poolSize_(poolSize) {}
+  explicit MemoryAllocator(uint64_t poolSize) : poolSize_(poolSize) {}
 
   void reset() {
     maxMemoryAllocated_ = 0;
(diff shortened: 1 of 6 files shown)


opti-mix:
If this PR is considered to make sense, more places need to be updated to use these types before it can be merged.

E.g. all places that use the `MemoryAllocator` and assign the returned addresses to `size_t` variables.

nadavrot:
I guess that the tradeoff in this patch is the reduced readability because of the move to the new type (memalloc_size_t) vs the chance that we may want to run on a 32-bit host and target a 64-bit accelerator. 

Do you know if LLVM supports cross compilation for huge programs from a 32-bit host to a 64-bit target? I don't think so. Also, even if we fix this data structure, what about std::vector that uses size_t? 

@opti-mix

opti-mix:
@nadavrot I think LLVM always used fixed-width types to avoid any possible confusion. 

The main point of this PR is to make sure that we use a type that is guaranteed to be at least 64bit. `size_t` does not provide any guarantee about its size. BTW, @stoklund also was very much in favor of using e.g. explicit `uint64_t` instead of `size_t` to avoid any possible host/target mismatches.

The secondary goal was to introduce the typedefs for `memalloc_ptr_t` and `memalloc_size_t` to express the the meaning of the types we need and make it easier to redefine them in the future. But as this is a secondary concern, I'm fine with using e.g. a `uint64_t`, which does not introduce any new types.

nadavrot:
@opti-mix About llvm, no.  I don't think that you checked. LLVM usually uses size_t. Grep /include/llvm and count the times they use size_t. Most of the uses are inherited from stl (size method), and the c++ spec (placement new). In some cases they use 'unsigned' (for example, the size of operand list). Whenever they deal with memory they use size_t (ArrayRef, StringView, ostream buffer sizes, StringTable offsets, MC stuff, debug info position, etc).
 
In very specific cases they use unit64_t explicitly.  For example: uint64_t seek(uint64_t off); 

I don't mind updating the interface from size_t to uint64_t, but I don't like memalloc_size_t because it's unreadable.

nadavrot:
@opti-mix  @jfix71 btw, I also don't like the new 'unsinged_t' type that was added. It's totally okay to use 'unsigned' for obvious things like the number of dimensions or kernel size. 'unsinged_t' does not improve readability.

opti-mix:
@nadavrot  

> LLVM usually uses size_t. Grep /include/llvm and count the times they use size_t. Most of the uses are inherited from stl (size method), and the c++ spec (placement new). In some cases they use 'unsigned' (for example, the size of operand list). Whenever they deal with memory they use size_t (ArrayRef, StringView, ostream buffer sizes, StringTable offsets, MC stuff, debug info position, etc).

As far as I can see most if not all of these uses are related to the e.g. indexing or addressing something on the host side, in which case `size_t` is pretty safe. 

But we use our `MemoryAllocator` also to represent memory allocation on the target device. And for the target device LLVM IR always uses fixed-sized types, right? There is no `size_t` in LLVM IR, IIRC.

opti-mix:
Based on our discussion, I updated the PR:
- Use `uint64_t` instead of `size_t` in the MemoryAllocator
- Update the users of the MemoryAllocator to use `uint64_t` instead of `size_t`

tests/unittests/memoryAllocatorTest.cpp
```diff
@@ -111,3 +111,17 @@ TEST(MemAlloc, allocateToTheMax) {
 
   EXPECT_EQ(MA.getMaxMemoryUsage(), 128);
 }
+
+/// Since MemoryAllocator is used for different purposes, including the
+/// modelling of a memory allocation process on a target device, the types used
+/// to represent the allocated addresses should be unsigned and at least 64-bit
+/// wide.
+TEST(MemAlloc, testMemoryAllocatorTypes) {
```
nadavrot: I think that this unit test can be replaced with a static assert.

opti-mix: Good idea! Added a static_assert instead of this unit test.

PR #1484, opened 2018-08-20
Extend MemoryAllocator to handle evictions and to support handles

opti-mix:
- There is a new version of the `allocate` API that supports eviction.

  If the allocator cannot find enough free space to satisfy the allocation request, it tries to find a set of allocated blocks whose eviction would result in freeing enough space to allocate the memory block of requested size. The client may specify which allocated blocks cannot be evicted.

  The current version of the allocator uses a first-fit approach when it looks for the buffers to be evicted.

- Also, each allocation now takes a `handle` parameter and associates this handle with the allocation. A handle is typically a client-specific object, representing what is being allocated, e.g. a `Value *` whose payload is being allocated by means of a MemoryAllocator. Clients can use handles to deallocate buffers, to describe buffers that should not be evicted, etc.

### DIFF
--- a/include/glow/CodeGen/MemoryAllocator.h
+++ b/include/glow/CodeGen/MemoryAllocator.h
@@ -18,6 +18,9 @@
 #include <cstddef>
 #include <cstdint>
 #include <list>
+#include <set>
+#include <unordered_map>
+#include <vector>
 
 namespace glow {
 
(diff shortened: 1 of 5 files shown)


opti-mix:
@nadavrot Please have a look. This is the initial implementation of what we discussed off-line.

nadavrot:
Beautiful!

Issue #1482, opened 2018-08-19
The quantization method should not delete the original function


nadavrot:
A recent change broke the behavior of the quantization function and made it unfit for certain usecases, such as comparing the quantized and unquantized methods. The solution is simple -- just remove the erase method that was added.  

https://github.com/pytorch/glow/blob/master/lib/Quantization/Quantization.cpp#L437

opti-mix:
@ZchiPitt I was pretty sure you fixed it already as we all agreed that this is the way forward. See https://github.com/pytorch/glow/pull/1458#issuecomment-413016807. Could you fix this trivial issue ASAP.

ZchiPitt:
thx @rdzhabarov for fixing it. Sorry i should have done it earlier

rdzhabarov:
no worries!

PR #1485, opened 2018-08-20
[quantization] Remove original function when needed after running quantization.

rdzhabarov:
closes https://github.com/pytorch/glow/issues/1482

image-classifier run looks good with dump/load profile.

### DIFF
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -164,7 +164,13 @@ struct Model {
       auto quantizationInfos = deserializeFromYaml(loadProfileFileOpt);
 
       // Quantize the graph based on the captured profile.
-      F_ = glow::quantization::quantizeFunction(EE_, quantizationInfos, F_);
+      auto *Q =
+          glow::quantization::quantizeFunction(EE_, quantizationInfos, F_);
+
+      // Erase the original function so that the redundant variables that are
+      // only referenced by the original function will be removed.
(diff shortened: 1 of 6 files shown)


rdzhabarov:
> Maybe we should add a really simple test here? E.g. create a graph, quantize it, verify that the number of users per variable is double, and then erase and verify the number went back down.

I think that should be tested as part of the eraseFunction test (This test is already in place) as it is not specific to quantization behavior.
The right place would be to have a loader test with the quantization flag on. Loader test seems to be a bit more problematic.

> Also, we call quantizeFunction() in the tests, e.g. in MLTest and quantizationTest. Should we add the erase there too?

We could but it's up to caller if he wants to actually erase function. Caller may want to keep function as well and it's a legit behavior.

examples/fr2en.cpp
```diff
@@ -164,7 +164,12 @@ struct Model {
       auto quantizationInfos = deserializeFromYaml(loadProfileFileOpt);
 
       // Quantize the graph based on the captured profile.
-      F_ = glow::quantization::quantizeFunction(EE_, quantizationInfos, F_);
+      auto *Q = glow::quantization::quantizeFunction(EE_, quantizationInfos, F_);
+      
+      // Erase the original function so that the redundant variables that are only
```
jfix71: nit: this looks a little too long, did you clang-format

PR #1486, opened 2018-08-20
[unnitest] A bugfix for nonSquarePaddingConvtTest in BackendCorrectnessTest.cpp

beicy:
The type of output node was inconsistent with the one in function "inferNonSquarePaddingConv" due to some incorrect "copy&paste".


beicy:
@rdzhabarov @jfix71 Yes, the tensor will be reset. Will file another PR for the clean up. Thanks!

PR #1491, opened 2018-08-21
[GraphOptimizer] Fix/simplify optimizeArithmeticNodes()

jfix71:
The previous version of `optimizeArithmeticNodes()` and `simplifyNode()`, which used a combination of a worklist and recursion, made this error prone, and hard to debug/understand. I simplified it by moving to just a worklist.

The previous version also had a bug. There were extra adds and maxes in Resnet50 because `simplifyNodes()` canonicalizes by creating a new node, but the old node was not being replaced by the new node. I think it was assumed it would be done recursively up the stack, but it was not the case. The added test case shows displays such a pattern. The adds and maxes were essentially being split into `n` new nodes given `n` users.

These extra nodes would have been recombined during CSE, however the `optimizeArithmeticNodes()` pass comes after CSE. Even still, it would be silly to rely on CSE to fix problems this pass creates.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -17,6 +17,7 @@
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Node.h"
 #include "glow/Graph/Nodes.h"
+#include "glow/Graph/Utils.h"
 #include "glow/Optimizer/Optimizer.h"
 #include "glow/Quantization/Base/Base.h"
 
@@ -153,32 +154,6 @@ bool isConstant(Node *N) { return isa<SplatNode>(N); }
 
 /// \returns the new simplified node or the original node.
 static Node *simplifyNode(Node *node, Function *F) {
-
-// Recursively simplify the operands of arithmetic nodes.
-#define SIMPLIFY_OPERANDS(NodeKind)                                            \
-  if (auto *NN = dyn_cast<NodeKind##Node>(node)) {                             \
-    Node *LHS = simplifyNode(NN->getLHS(), F);                                 \
-    if (LHS != NN->getLHS()) {                                                 \
-      return simplifyNode(                                                     \
-          F->create##NodeKind(NN->getName(), LHS, NN->getRHS()), F);           \
-    }                                                                          \
-    Node *RHS = simplifyNode(NN->getRHS(), F);                                 \
-    if (RHS != NN->getRHS()) {                                                 \
-      return simplifyNode(                                                     \
-          F->create##NodeKind(NN->getName(), NN->getLHS(), RHS), F);           \
-    }                                                                          \
-  }
-
-  SIMPLIFY_OPERANDS(Add)
-  SIMPLIFY_OPERANDS(Mul)
-  SIMPLIFY_OPERANDS(Div)
-  SIMPLIFY_OPERANDS(Sub)
-  SIMPLIFY_OPERANDS(Max)
-  SIMPLIFY_OPERANDS(Min)
-  SIMPLIFY_OPERANDS(CmpLTE)
-  SIMPLIFY_OPERANDS(CmpEQ)
-#undef SIMPLIFY_OPERANDS
-
 // Simplify commutative nodes by moving the constant operator to the right-hand
 // side.
 // Example:  C + X  =>  X + C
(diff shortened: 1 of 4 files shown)


opti-mix:
>  I don't think we have a way to iterate topologically over the graph at the node level.

Would `GraphPostOrderVisitor` or `PostOrderVisitor` help here?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1159,11 +1133,21 @@ static void optimizeArithmeticNodes(Function *F) {
...
+
+      // The simplified node's operands could be further simplified as well.
+      if (SN->isArithmetic()) {
+        auto *LHS = SN->getNthInput(0).getNode();
+        auto *RHS = SN->getNthInput(1).getNode();
+        if (LHS->isArithmetic())
+          worklist.push_back(LHS);
+        if (RHS->isArithmetic())
+          worklist.push_back(RHS);
+      }
```
beicy: Is that OK to push_back LHS and RHS first, then SN?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1159,11 +1133,21 @@ static void optimizeArithmeticNodes(Function *F) {
...
+    if (SN != node) {
+      node->getNthResult(0).replaceAllUsesOfWith(SN);
       // The simplified node could be further simplified.
-      worklist.push_back(sn);
+      worklist.push_back(SN);
+
+      // The simplified node's operands could be further simplified as well.
+      if (SN->isArithmetic()) {
+        auto *LHS = SN->getNthInput(0).getNode();
+        auto *RHS = SN->getNthInput(1).getNode();
```
qcolombet: Could we iterate over all the inputs instead of hardcoding the number of operands?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1159,11 +1133,21 @@ static void optimizeArithmeticNodes(Function *F) {
     Node *node = worklist.back();
     worklist.pop_back();
 
-    auto *sn = simplifyNode(node, F);
-    if (sn != node) {
-      node->getNthResult(0).replaceAllUsesOfWith(sn);
+    auto *SN = simplifyNode(node, F);
+    if (SN != node) {
+      node->getNthResult(0).replaceAllUsesOfWith(SN);
```
qcolombet: Maybe as a follow-up PR, could we feature a Node::replaceAllUsesOfWith that would replace each result of the node with the related results of the other nodes. Hardcoding the number of results makes me nervous.

We don't have arithmetic nodes with more than one result yet, but if we do at some point (e.g., something equivalent to LLVM UDIVREM that returns both the quotient and remainder), this is going to be wrong.

For now, we should assert that we only have one result. (We could leave it this way until the assert breaks too.)

tests/unittests/graphOptzTest.cpp
```diff
@@ -1169,3 +1169,42 @@ TEST_F(GraphOptz, VarsCSE) {
...
+  EXPECT_EQ(I1->getUsers().size(), 1);
+
+  // Simplify nodes will canonicalize add1, and should replace all its users.
+  ::glow::optimize(F_, CompilationMode::Infer);
+
+  // We should have the same five nodes: one splat, two adds, and two saves.
+  EXPECT_EQ(F_->getNodes().size(), 5);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::SplatNodeKind), 1);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::AddNodeKind), 2);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::SaveNodeKind), 2);
```
qcolombet: This doesn't check that the expected canonicalization happened.
Could you check that the splat is indeed on the left hand side?
We should also check that the shape of the graph is still what we want (i.e., save1 takes add1, etc.)

jfix71: This test isn't actually meant to verify canonicalization works -- it's meant to make sure that canonicalization does not accidentally duplicate a node for each user.  I will still check to make sure everything else is expected.

lib/Optimizer/GraphOptimizer.cpp
jfix71: I'm not sure what you mean here -- the operands are pushed on after the node itself. So the operands will be processed before the node.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1149,21 +1124,38 @@ static void optimizeArithmeticNodes(Function *F) {
...
   std::vector<Node *> worklist;
 
-  // Add all of the interesting nodes to the worklist.
-  for (auto &node : F->getNodes()) {
-    if (node.isArithmetic()) {
-      worklist.push_back(&node);
+  // Add all of the arithmetic nodes to the worklist, with a node's dependencies
+  // added after itself so they are processed before the node.
+  GraphPreOrderVisitor visitor(*F);
+  for (auto *N : visitor.getPreOrder()) {
```
opti-mix: Nit: You could use `worklist.reserve(visitor.getPreOrder().size());` to avoid multiple dynamic `resize` calls when you add elements to the worklist.

jfix71: Good idea, thanks! Done.

nadavrot:
@qcolombet @opti-mix Does LLVM use PreOrderVisitor in InstCombine? Does Swift use PreOrderVisitor in SilCombine? Why not?

qcolombet:
LLVM does not use pre order. LLVM uses a kind of fixed point algorithm (not really fixed point but that's the idea :P), where it processed last added first.

qcolombet:
Oops, I thought you were asking for DAGCombine!

qcolombet:
Yes, instcombine uses pre-order

nadavrot:
>> LLVM does not use pre order. 

Oh, yes. Scanning the basic block backwards is ... right. :) Thank you.

PR #1493, opened 2018-08-21
Prevent getInputName() from returning a dangling pointer

### LINKED ISSUE #1492: `ConcatNode::getInputName()` returns a reference to temporary string.
```
llvm::StringRef ConcatNode::getInputName(unsigned idx) const {
  idx -= 0;
  if (idx < Inputs_.size()) { return "Inputs" + std::to_string(idx); }
  idx -= Inputs_.size();
  llvm_unreachable("Invalid index");
}
```
This is also problematic for nodes which can accept the variable number of inputs.

boncheolgu:
fixes #1492

### DIFF
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -281,7 +281,7 @@ public:
 
   /// Getters/setters to access Node's inputs and outputs.
   unsigned getNumInputs() const;
-  llvm::StringRef getInputName(unsigned idx) const;
+  std::string getInputName(unsigned idx) const;
   NodeValue getNthInput(unsigned idx);
   const NodeValue getNthInput(unsigned idx) const;
   void setNthInput(unsigned idx, NodeValue val);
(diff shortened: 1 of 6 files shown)


rdzhabarov:
@boncheolgu what do you think if you remove that while loop and lazy initialization of names, but rather 
```
static std::vector<std::string> names;
if (names.empty()) {
   for(...) {
      names.push_back(...)
   }
}
```
This way you'll remove
```
while (idx >= names.size()) {
      names.push_back("Inputs" + std::to_string(names.size()));
    }
```
completely, which might be more readable.

Just initialize names right away and then return names[idx].

nadavrot:
I don't think that this PR fixes the problem.  We are still returning a StringRef that points to some internal storage std::vector<std::string>

nadavrot:
oops, I am sorry. @rdzhabarov and @opti-mix pointed out that the vector is static, which means that this PR is correct. 

I wonder, would it be better to just return std::string?

rdzhabarov:
+1 to std::string, that will simplify this method quite a bit (Performance does not seem to be a concern for this code path)

nadavrot:
👍

PR #1497, opened 2018-08-21
[opencl][quantization]: Support TopK for OpenCL backend, with also quantization support

ZchiPitt:
Support TopK for OpenCL backend, with also quantization support.
Also enabled end2endGRU test for OpenCL now.

For the time being the actual computation of TopK is performed on the host side and not by means of an OpenCL kernel. We may add an optimized kernel in the future.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -533,6 +533,39 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   enqueueKernel(commands_, kernel, deviceId_, global, local, kernelLaunches_);
 }
 
+/// This method is copied from InterpreterNodes.cpp. Please be aware that
+/// they should be in sync.
+template <typename T>
+static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
+  auto values = outW.getHandle<T>();
+  auto indices = indW.getHandle<size_t>();
+  auto in = inW.getHandle<T>();
+  size_t n = in.dims().back();
+
+  size_t in_p = 0, out_p = 0;
+  size_t tensor_end = in.size();
+  using pairType = std::pair<float, size_t>;
+  std::vector<pairType> buf(n);
+
+  while (in_p < tensor_end) {
+    for (size_t i = 0; i < n; i++) {
+      buf[i].first = in.raw(in_p++);
+      buf[i].second = i;
+    }
+    // NOTE: it's possible to do N + KlogK, while this version is NlogN
+    std::sort(buf.begin(), buf.end(), [](const pairType &a, const pairType &b) {
+      if (a.first != b.first)
+        return a.first > b.first;
+      return a.second < b.second;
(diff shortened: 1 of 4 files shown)


opti-mix:
@ZchiPitt Can you add to the commit message that for the time being the actual computation of TopK happens on the host side and not by means of an OpenCL kernel. We may add an optimized kernel in the future.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -533,6 +533,37 @@ void OpenCLFunction::executeConvolution(const OCLConvolutionInst *CC) {
   enqueueKernel(commands_, kernel, deviceId_, global, local, kernelLaunches_);
 }
 
+template <typename T>
```
opti-mix: Is it copied from e.g. interpreter backend? If so, please mention it in the comment. So that we know they should be in sync.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1218,6 +1249,30 @@ void OpenCLFunction::execute() {
       continue;
     }
 
+    if (auto *TK = dyn_cast<TopKInst>(&I)) {
+      clFinish(commands_);
```
opti-mix: Add a comment that you perform the computation on the host and to do that you copy the tensor from GPU memory to host memory, perform the operation and then copy the results back to GPU memory.

ZchiPitt:
@opti-mix added

lib/Backends/OpenCL/OpenCL.cpp
ZchiPitt: done

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1218,6 +1251,33 @@ void OpenCLFunction::execute() {
       continue;
     }
 
+    // For TopKInst, we perform the computation on the host side. We copy
+    // the tensor from GPU memory to host memory, performa the computation,
```
opti-mix: performa -> perform a

ZchiPitt: thx

tests/unittests/quantizationTest.cpp
```diff
@@ -332,7 +332,7 @@ static Function *createGRUForQuantization(Module *M, llvm::StringRef funcName) {
   return F;
 }
 
-TEST_P(Quantization, end2endGRU) {
+TEST_P(Operator, end2endGRU) {
```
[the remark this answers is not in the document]
ZchiPitt: yep, this is checked

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1218,6 +1251,33 @@ void OpenCLFunction::execute() {
       continue;
     }
 
+    // For TopKInst, we perform the computation on the host side. We copy
```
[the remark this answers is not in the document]
ZchiPitt: added

PR #1499, opened 2018-08-21
[Interpreter] Improve safety checks for Gather.

artemrakhov:


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -752,13 +752,17 @@ void InterpreterFunction::fwdGatherInst(const glow::GatherInst *I) {
   // Calculate the size of each sample in the batch.
   size_t numSamples = (dataT->size() * elementSize) / dataSampleSize;
 
+  // Calculate number of samples in the batch.
+  size_t batchSize = dataTy.dims()[batchDims];
```
qcolombet: I believe we would need a `(void)batchSize` to avoid release build to complain about an unused variable.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -752,13 +752,17 @@ void InterpreterFunction::fwdGatherInst(const glow::GatherInst *I) {
...
+
   // For each sample in the batch:
   for (size_t sample = 0; sample < numSamples; sample++) {
     size_t sampleStart = sample * dataSampleSize;
 
     // For each slice (small fragment) that we copy from the source memory:
     for (size_t i = 0, end = indicesT->size(); i < end; i++) {
       size_t slice = indicesT->getHandle<size_t>().raw(i);
+      assert(slice < batchSize && "Invalid index seen during Gather operation");
```
qcolombet: It looks to me like we could write something that would violate this assert.
Shouldn't we come up with a protection mechanism that works with release build as well?

qcolombet: Right, but here I believe we still want to complain loudly to avoid reading random memory.
In particular, it might be a security concern.

artemrakhov: I'm really not sure what is right here.
1. In release mode, `std::vector` doesn't check for index violations.
2. We have a lot of verifications, all use normal asserts and don't care about reading random memory. 
3. Do you suggest to also add this assert in CPU backend?
I always thought that in production mode it's fine to skip safety checks and let everything fail.

qcolombet: Maybe go with GLOW_ASSERT?

 I think there are two things to consider:
- What we run in the compiler
- What we code-gen

For what we run in the compiler, I would rather not read random memory, especially when we are run in JIT mode.
 
For what we code-gen, we basically generate what the developer asked for, therefore if they provide security vulnerable code, that's their problem. That said, we could imagine featuring a paranoid mode that would insert such checks.

Based on that, no, I wouldn't suggest adding this assert for the CPU nor for the interpreter. However, I'd say the interpreter is special because it is mainly a helper tool for debugging, thus we don't really care if it's slow(er). In particular, I don't expect we will run any production code with the interpreter.

artemrakhov: Imagine that someone called `createMatMul(name, outTy, lhs, rhs)`. `outTy` is 1000x1000 matrix, while both `lhs` and `rhs` are 1x1. We'll have out-of-bounds access with no error upon executing this in Release with Interpreter. I think it's conscious decision to have out-of-bounds checks only in Debug, but not in Release (`Tensor.h`):
https://github.com/pytorch/glow/blob/ba039c05881692dde6c79b6b8f293f166eb6812e/include/glow/Base/Tensor.h#L501-L507

There are other examples like this. You take a net, run it in Release, see incorrect result. Then you run it in Debug, and see that assert failed.

My point is that it's not only here where we do unsafe access. There are a lot more places. I just wanted to match the whole codebase's style.

qcolombet: I think there is also a trade-off of speed versus safety. I agree we wouldn't want to check all the tensors accesses as they are too many. In that case, it seemed relatively infrequent to be valuable.
The problem we have in general IMHO is that, with Debug mode it is too slow to be practical to run on large networks and with Release mode we get too few information to know what went wrong. It would be nice to have targeted (infrequent checks) that would quickly help pin-pointing what is going wrong.

Now, I agree we are not going to fix that in that one PR, so let us stick on what we did so far :).

Issue #1481, opened 2018-08-18
[RFC] Improve sinkRescaleQuantized


aturetsk:
When I was running some NN models with GLOW, I noticed that final (after all optimizations) graph contains the following sequence:

Convolution -> LRN -> RescaleQuantized -> MaxPool -> Convolution -> ...

To get rid of RescaleQuantized one can improve sinkRescaleQuantized pass the following way:
1) Allow Rescale to sink down with MaxPool/AvgPool
2) Combine Rescale down with Convolution

I made experimental change (4aa6449770d94be247175972c0f40ce6e6f42486) and it helped.
To me, both of these transformations seem legal, but I might be missing something. Can anyone confirm that they are OK?

If they are, I'll add some tests and split the changes in two pull request.

nadavrot:
Thanks for doing this work @aturetsk.  @rdzhabarov @qcolombet What do you think?

aturetsk:
> Another consideration would be to merge rescale up to the LRN node.

I think combining Rescale up with LRN will work as well. But...

I find this code a bit weird:
` while (sinkRescaleQuantizedNode(F)) {  // sink + combine down`
`       DCE(F);`
`       optimizeQuantization(F); // combine up + other optimizations`
`  }`

I wonder if it really necessary to have "combine **up**" rescale optimization at all, taking into account that we use rescale "sink" strategy. Because moving rescale down kind of contradicts our attempts to combine them up. If I'm not missing something, sinking rescale and combining it **down** with nodes should be enough to get rid of all unnecessary rescales. Do you have any use case where it's not so?

Another idea about rescale combination/sinking:
Do you think ClassGen can be used to generate sinkRescaleQuantizedNode/optimizeQuantization (add a special attribute to every node describing which strategy to chose to optimize rescales, e.g. sink, combine or custom)?
This would help to:
1) Avoid extending sinkRescaleQuantizedNode/optimizeQuantization manually for every node for trivial cases (sink and combine). Only complicated cases ('custom', e.g. for concat node, where we need to ensure all inputs have the same scale/offset) will be added manually.
2) Be able to cover backend-specific nodes in rescale optimizations as well. Backend may replace some nodes with backend specific versions and create some rescales (e.g. CPUMaxSplat). Right now, they won't be optimized, because sinkRescaleQuantizedNode/optimizeQuantization work only with generic nodes.

What do you think?

aturetsk:
[...]

PR #1500, opened 2018-08-21
[graph-optimizer] Sink Rescale down with AvgPool/MaxPool nodes

aturetsk:
The first part of the rescale optimization proposed in https://github.com/pytorch/glow/issues/1481: sink rescale through max/avg pooling node.


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1708,6 +1708,28 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       continue;
     }
 
+// Sink Rescale down with Pooling node.
+// PoolingNode(Rescale(X)) -> PoolingNode(Rescale(X)).
```
rdzhabarov: `PoolingNode(Rescale(X)) -> PoolingNode(Rescale(X)).` ?

PR #1501, opened 2018-08-21
[graph-optimizer] Combine Rescale down with Convolution node

aturetsk:
The first part of the rescale optimization proposed in #1481: combine rescale down with convolution node.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1725,6 +1725,28 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       continue;
     }
 
+    // Combine Rescale down with Convolution node.
+    // Convolution(Rescale(X), F, B) -> Convolution(X, F, B).
+    // Convolution(X, Rescale(F), B) -> Convolution(X, F, B).
+    // Convolution(X, F, Rescale(B)) -> Convolution(X, F, B).
+    // ... and different combinations.
+    if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
+      auto *rescaleX = dyn_cast<RescaleQuantizedNode>(CN->getInput());
+      auto *rescaleF = dyn_cast<RescaleQuantizedNode>(CN->getFilter());
+      auto *rescaleB = dyn_cast<RescaleQuantizedNode>(CN->getBias());
+      auto newX = rescaleX ? rescaleX->getInput() : CN->getInput();
+      auto newF = rescaleF ? rescaleF->getInput() : CN->getFilter();
+      auto newB = rescaleB ? rescaleB->getInput() : CN->getBias();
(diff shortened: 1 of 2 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1725,6 +1725,28 @@ static bool sinkRescaleQuantizedNode(Function *F) {
...
+    // Convolution(X, F, Rescale(B)) -> Convolution(X, F, B).
+    // ... and different combinations.
+    if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
+      auto *rescaleX = dyn_cast<RescaleQuantizedNode>(CN->getInput());
+      auto *rescaleF = dyn_cast<RescaleQuantizedNode>(CN->getFilter());
+      auto *rescaleB = dyn_cast<RescaleQuantizedNode>(CN->getBias());
+      auto newX = rescaleX ? rescaleX->getInput() : CN->getInput();
+      auto newF = rescaleF ? rescaleF->getInput() : CN->getFilter();
+      auto newB = rescaleB ? rescaleB->getInput() : CN->getBias();
+      if (rescaleX || rescaleF || rescaleB) {
```
rdzhabarov: why do you need this condition?

aturetsk: Without this condition convolution is going to be replaced with the same convolution every time sinkRescaleQuantizedNode is called (which is in a loop). While the transformation is correct, this will lead to node name exploding from "conv_name" to "conv_name11111111111...". Better to avoid that to make dumps more human-readable.

rdzhabarov: ah, I overlooked :) for some reason, I thought it was the check against newX, newF etc.

tests/unittests/graphOptzTest.cpp
```diff
@@ -905,6 +905,41 @@ TEST_F(GraphOptz, FuseRescaleIntoArithmetic) {
...
+TEST_F(GraphOptz, fuseRescaleIntoConv) {
+  // This test ensures the fact that fusing of rescale is done.
+  Variable *input =
+      mod_.createVariable(ElemKind::Int8QTy, {1, 10, 20, 3}, 0.5, 10, "input");
+  Variable *filter =
+      mod_.createVariable(ElemKind::Int8QTy, {16, 5, 5, 3}, 0.5, 10, "filter");
+  Variable *bias =
+      mod_.createVariable(ElemKind::Int8QTy, {16}, 0.5, 10, "bias");
+
+  input->getPayload().init(Tensor::InitKind::Broadcast, 1, mod_.getPRNG());
```
rdzhabarov: Init is not really necessary here

aturetsk:
Thanks for review! Fixed the test.

PR #1502, opened 2018-08-21
Fix bugs in our Caffe2 importer

qcolombet:
Each commit fixes a different bug:
- First one has to do with our modeling of concat which has a slightly different resulting shape between glow and caffe2
- Second one has to do with the indices used in the transpose for batched matmul.

Note: I stacked these commits in the same PR, because I use a newly introduced file (empty_init_net.pbtxt) in both of them.

### DIFF
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -270,6 +270,22 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     Node *node = G_.createConcat(opName, inputs, channel);
 
+    unsigned_t addAxis = dict.count("add_axis") ? loadInt(dict["add_axis"]) : 0;
+
+    if (addAxis) {
+      // When add axis is used, this means we have to add a new dimension before
+      // the axis, instead of merging on the axis.
+      std::vector<size_t> outputDims = inputs[0].dims();
+      for (const auto &input : inputs) {
+        GLOW_ASSERT(input.dims().size() > channel &&
+                    "input should have at least #axis dims");
+        GLOW_ASSERT(
+            outputDims[channel] == input.dims()[channel] &&
+            "inputs need all to have the same dims for concat with add_axis");
+      }
+      outputDims.insert(outputDims.begin() + channel, numInputs);
+      node = G_.createReshape(opName, node, outputDims);
+    }
     // Concat has multiple outputs in Caffe2, but I believe the other output
     // (split_info) is not used for inference.
     nodeValueByName_[op.output(0)] = NodeValue(node, 0);
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -277,8 +277,6 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       // the axis, instead of merging on the axis.
       std::vector<size_t> outputDims = inputs[0].dims();
       for (const auto &input : inputs) {
-        GLOW_ASSERT(input.dims().size() > channel &&
-                    "input should have at least #axis dims");
         GLOW_ASSERT(
             outputDims[channel] == input.dims()[channel] &&
             "inputs need all to have the same dims for concat with add_axis");
@@ -470,7 +468,18 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     assert(!transLHS && "Don't support transpose lhs for now.");
     bool transRHS = dict.count("trans_b") && (loadInt(dict["trans_b"]) == 1);
     if (transRHS) {
-      RHS = G_.createTranspose("RHS.transpose", RHS, {1, 0});
+      // The semantic of the transpose in that context is:
+      // swap the last two dimensions.
+      unsigned_t nbDims = RHS.dims().size();
+      GLOW_ASSERT(nbDims >= 2 && "C2 specs say rank of RHS must be >= 2");
+      std::vector<unsigned_t> shuffle;
+      unsigned_t i;
+      for (i = 0; i < nbDims - 2; ++i) {
+        shuffle.push_back(i);
+      }
+      shuffle.push_back(i + 1);
+      shuffle.push_back(i);
+      RHS = G_.createTranspose("RHS.transpose", RHS, shuffle);
     }
 
     Node *node = nullptr;
--- a/tests/models/caffe2Models/concat_predict_net.pbtxt
+++ b/tests/models/caffe2Models/concat_predict_net.pbtxt
@@ -1,0 +1,22 @@
+name: "concatx3"
+op {
+  input: "inputs_0"
+  input: "inputs_1"
+  input: "inputs_2"
+  output: "concat_result"
+  output: "split_info"
+  name: ""
+  type: "Concat"
+  arg {
+    name: "add_axis"
+    i: 0
+  }
+  arg {
+    name: "axis"
+    i: 1
+  }
+}
+external_input: "inputs_0"
+external_input: "inputs_1"
+external_input: "inputs_2"
+external_output: "concat_result"
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -53,3 +53,149 @@ TEST(caffe2, importConv) {
   for (size_t i = 0; i < 4 * 4; i++)
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
 }
+
+/// Test loading a concat node with add_axis.
+/// Concat nodes with add_axis have a different semantic
+/// than the plain glow concat.
+/// concat A(dim0, dim1), B(dim0, dim1), ... 1, add_axis = 1
+/// res = A, B...
+/// C2 shape: dim0, #input, dim1, i.e., three dimensions.
+/// Glow shape: dim0, #input x dim1, i.e., two dimensions.
+///
+/// To fill the gap between the two, glow issues a reshape
+/// right after its concat.
+TEST(caffe2, concatAddAxis) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename(
+      "tests/models/caffe2Models/concat_add_axis_predict_net.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  SaveNode *output;
+  Tensor inputs_0(ElemKind::FloatTy, {10, 7});
+  Tensor inputs_1(ElemKind::FloatTy, {10, 7});
+  Tensor inputs_2(ElemKind::FloatTy, {10, 7});
+  inputs_0.getHandle().randomize(-3.0, 3.0, mod.getPRNG());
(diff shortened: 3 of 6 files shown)


lib/Importer/Caffe2.cpp
```diff
@@ -454,7 +462,17 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     assert(!transLHS && "Don't support transpose lhs for now.");
     bool transRHS = dict.count("trans_b") && (loadInt(dict["trans_b"]) == 1);
     if (transRHS) {
-      RHS = G_.createTranspose("RHS.transpose", RHS, {1, 0});
+      // The semantic of the transpose in that context is:
+      // swap the last two dimensions.
+      size_t nbDims = RHS.dims().size();
+      std::vector<unsigned_t> shuffle;
+      unsigned_t i;
+      for (i = 0; i < (unsigned_t)nbDims - 2; ++i) {
```
artemrakhov: Bad things can happen if nbDims is less than 2.

lib/Importer/Caffe2.cpp
```diff
@@ -454,7 +462,17 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     assert(!transLHS && "Don't support transpose lhs for now.");
     bool transRHS = dict.count("trans_b") && (loadInt(dict["trans_b"]) == 1);
     if (transRHS) {
-      RHS = G_.createTranspose("RHS.transpose", RHS, {1, 0});
+      // The semantic of the transpose in that context is:
+      // swap the last two dimensions.
+      size_t nbDims = RHS.dims().size();
```
artemrakhov: nit: use `unsigned_t` here so you won't need to cast later.

lib/Importer/Caffe2.cpp
```diff
@@ -270,9 +270,17 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     Node *node = G_.createConcat(opName, inputs, channel);
 
+    // The semantic of the C2 concat operator is a bit different
```
artemrakhov: Concat in C2 has 2 modes. Do this new code path when `add_axis=1`, otherwise old implementation is correct.

tests/models/caffe2Models/concat_predict_net.pbtxt
```diff
@@ -0,0 +1,22 @@
...
+op {
+  input: "inputs_0"
+  input: "inputs_1"
+  input: "inputs_2"
+  output: "concat_result"
+  output: "split_info"
+  name: ""
+  type: "Concat"
+  arg {
+    name: "add_axis"
```
artemrakhov: Have you tried this net in Caffe2? If add_axis=0, it should just sum dimensions.

lib/Importer/Caffe2.cpp
qcolombet: Good point I haven’t!
And I managed to screw the writing of the test.

tests/models/caffe2Models/concat_predict_net.pbtxt
qcolombet: Good point I haven’t!
And I managed to screw the writing of the test.

lib/Importer/Caffe2.cpp
qcolombet: Added an assert on that. The C2 specs say that the rank of the inputs must be >= 2.

lib/Importer/Caffe2.cpp
qcolombet: Done

lib/Importer/Caffe2.cpp
qcolombet: Done and added a test for each case.

tests/models/caffe2Models/concat_predict_net.pbtxt
qcolombet: I concur this is what add_axis does. The documentation is really messed up it says it modifies the axis of the inputs!

lib/Importer/Caffe2.cpp
```diff
@@ -270,9 +270,20 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     Node *node = G_.createConcat(opName, inputs, channel);
 
+    unsigned_t addAxis = dict.count("add_axis") ? loadInt(dict["add_axis"]) : 0;
+
+    Node *concatOutput = node;
```
artemrakhov: nit: you can remove this line. Either rename above `Node *node=` to `Node *concatOutput=`, or use `node` instead of `concatOutput` below.

lib/Importer/Caffe2.cpp
```diff
@@ -270,9 +270,20 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+
+    if (addAxis) {
+      // When add axis is used, this means we have to add a new dimension before
+      // the axis, instead of merging on the axis.
+      std::vector<size_t> outputDims = inputs[0].dims();
```
artemrakhov: Should we also check that dims for inputs[1], inputs[2], ... match?

qcolombet:
@artemrakhov thanks for the review.
I believe I have addressed all the feedbacks, please take a look.

lib/Importer/Caffe2.cpp
```diff
@@ -454,7 +465,18 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
     bool transRHS = dict.count("trans_b") && (loadInt(dict["trans_b"]) == 1);
     if (transRHS) {
-      RHS = G_.createTranspose("RHS.transpose", RHS, {1, 0});
+      // The semantic of the transpose in that context is:
+      // swap the last two dimensions.
+      unsigned_t nbDims = RHS.dims().size();
+      assert(nbDims >= 2 && "C2 specs say rank of RHS must be >= 2");
+      std::vector<unsigned_t> shuffle;
+      unsigned_t i;
+      for (i = 0; i < nbDims - 2; ++i) {
```
artemrakhov: In release mode, if assertion fails, this loop may turn into time & memory eating thing. May be annoying to fight with on local machine.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -53,3 +53,209 @@ TEST(caffe2, importConv) {
...
+  // With have 2 inputs and one outputs.
+  EXPECT_EQ(mod.getVars().size(), 3);
+   // Check that the graph has the expected shape,
+  // starting from the output.
+   // Batched matmul with broadcasted RHS are lowered
+  // to a regular matmul, where LHS is reshaped from
+  // a 3D tensor to a flattened matrix.
+  Node *reshapeResult = output->getNthInput(0).getNode();
+  EXPECT_TRUE(llvm::isa<ReshapeNode>(reshapeResult));
+   Node *matmul = reshapeResult->getNthInput(0).getNode();
```
artemrakhov: nit: 3 spaces instead of 2 here and below. Please run clang-format.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -53,3 +53,209 @@ TEST(caffe2, importConv) {
...
+  EXPECT_TRUE(llvm::isa<MatMulNode>(matmul));
+  const size_t matmulDims[] = {30, 10};
+  EXPECT_EQ(matmul->dims(0), llvm::makeArrayRef(matmulDims));
+   NodeValue lhs = matmul->getNthInput(0);
+  EXPECT_TRUE(llvm::isa<ReshapeNode>(lhs.getNode()));
+  Node *lhsInput = lhs.getNode()->getNthInput(0).getNode();
+  EXPECT_TRUE(llvm::isa<Variable>(lhsInput));
+  EXPECT_TRUE(llvm::cast<Variable>(lhsInput)->getPayload().isEqual(inputs_0));
+   NodeValue rhs = matmul->getNthInput(1);
```
artemrakhov: Is it feasible to use `getRHS()` instead of indexed input? Here and everywhere else?

tests/unittests/caffe2ImporterTest.cpp
qcolombet: Strange I did run clang-format?!

lib/Importer/Caffe2.cpp
qcolombet: Good point!

lib/Importer/Caffe2.cpp
qcolombet: What's the proper way to report an error in the loader?
The loadOperator method does not return any status.

GLOW_ASSERT?

lib/Importer/Caffe2.cpp
qcolombet: Done

lib/Importer/Caffe2.cpp
qcolombet: Done.

Added a couple of GLOW_ASSERTs for that.

lib/Importer/Caffe2.cpp
qcolombet: Went with GLOW_ASSERT.

tests/unittests/caffe2ImporterTest.cpp
qcolombet: Done

tests/unittests/caffe2ImporterTest.cpp
qcolombet: Done

lib/Importer/Caffe2.cpp
```diff
@@ -270,6 +270,22 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+    unsigned_t addAxis = dict.count("add_axis") ? loadInt(dict["add_axis"]) : 0;
+
+    if (addAxis) {
+      // When add axis is used, this means we have to add a new dimension before
+      // the axis, instead of merging on the axis.
+      std::vector<size_t> outputDims = inputs[0].dims();
+      for (const auto &input : inputs) {
+        GLOW_ASSERT(input.dims().size() > channel &&
+                    "input should have at least #axis dims");
+        GLOW_ASSERT(
```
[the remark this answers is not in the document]
qcolombet: Admittedly I don't have a lot of experience in using our loader, but the thing that struck me is how little information it gives in release mode. Thus, I found those valuable in release mode as well.

I quickly had to use the debug version to realize that the input model had silly mistakes like this one.

What's our plan regarding error reporting?

In particular, I would rather have only sane networks hit the compiler. I guess we could run unconditionally the verifier in the compiler after loading the network, though today that wouldn't help since every checks in the verifier are done in asserts.

What do you think?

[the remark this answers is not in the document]
artemrakhov: This is orthogonal to the PR of course. LGTM on the PR.

What if you had a C++ program, accessing `std::vector` with incorrect indices, or something like that? There's a good chance that it would silently finish with wrong result.

I agree that yes, we need better error reporting. Right now asserts produce very short messages, don't inline numbers, can't print instruction number, etc. Often times after hitting assert I run `gdb` to see what exactly is going on.

In the other hand, I can imagine compiler being ran in production system, where performance is critical.
I understand that this particular check is lightweight, but when a lot of such checks add up, they may consume valuable CPU time.
I don't support running verifier in Release. What if user compiles the same net over and over again? They are absolutely sure that it's correct, and want to squeeze every single CPU cycle.

Probably this whole topic needs a discussion outside of PR :)

PR #1508, opened 2018-08-22
[quantization] Properly quantize ReLU.

rdzhabarov:
This is not yet ready for review.
Will clean this up a bit later. @jfix71 see if this patch helps with the problem you were seeing.

Essentially it was a bit misleading to have ReLU lowered to Max node with the negative floating point range. Now we create ReLU with the output type according to quantization parameters after profiling (Before that was a sequence of Max->Rescale and Rescale can go away etc).

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -247,8 +247,15 @@ static void verifyArithmetic(NodeValue LHS, NodeValue RHS, NodeValue res) {
   checkSameShape(LHS, RHS);
 }
 
-static void verifyRelu(NodeValue src, NodeValue dest) {
-  checkSameType(src, dest);
+static void verifyRelu(NodeValue result, NodeValue input) {
+  if (input.getType()->isQuantizedType()) {
+    assert(result.getType()->isQuantizedType());
+    checkSameShape(result, input);
+    assert(result.getType()->getOffset() == -128 &&
+           "Min fp32 value should be 0");
+  } else {
+    checkSameType(result, input);
+  }
 }
 
 static void verifyRegression(NodeValue src, NodeValue dest,
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -160,7 +160,8 @@ static Node *simplifyNode(Node *node, Function *F) {
 #define COMMUTE_CONST_TO_RHS(NodeKind)                                         \
   if (auto *NN = dyn_cast<NodeKind##Node>(node))                               \
     if (isConstant(NN->getLHS()) && !isConstant(NN->getRHS())) {               \
-      return F->create##NodeKind(NN->getName(), NN->getRHS(), NN->getLHS());   \
+      return F->create##NodeKind(NN->getName(), NN->getResult().getType(),     \
+                                 NN->getRHS(), NN->getLHS());                  \
     }
 
   COMMUTE_CONST_TO_RHS(Add)
@@ -240,7 +241,11 @@ static bool sinkCode(Function *F) {
         continue;
       }
 
-      auto *NRL = F->createRELU(RL->getName(), TR->getInput());
+      // Keep the same quantization parameters for ReLU output, but
+      // change the shape to appropriate value.
+      auto reluOutTy = F->getParent()->uniqueTypeWithNewShape(
+          RL->getResult().getType(), TR->getInput().dims());
+      auto *NRL = F->createRELU(RL->getName(), TR->getInput(), reluOutTy);
       auto *newTR = F->createTranspose(TR->getName(), NRL, TR->getShuffle());
       RL->getResult().replaceAllUsesOfWith(newTR);
       changed = true;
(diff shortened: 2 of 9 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -175,8 +179,8 @@ static Node *simplifyNode(Node *node, Function *F) {
   SIMPLIFY_OPERANDS(Sub)
   SIMPLIFY_OPERANDS(Max)
   SIMPLIFY_OPERANDS(Min)
-  SIMPLIFY_OPERANDS(CmpLTE)
-  SIMPLIFY_OPERANDS(CmpEQ)
+  // SIMPLIFY_OPERANDS(CmpLTE)
```
rdzhabarov: this will be brought back in the PR for review.

jfix71: It's actually already removed in master since I landed the `simplifyNode()` redesign, so if you rebase it will go away anyway after you fix the conflicts 😃

jfix71:
@rdzhabarov Yeah, this combined with the rescale convolution inputs optimization that recently landed should fix the problems. Thanks for the fix!

I am noticing now though that we have RELUs which may not be necessary. For example, I believe we could remove the RELU here and just set the output quantization parameters of the Conv to have `min = 0`, no?

screen shot 2018-08-22 at 2 48 18 pm

rdzhabarov:
@jfix71 I think it will go away if the type of "Result" for Splat is set correctly (e.g. needs to be the Relu result). I'll fix this in the PR once get back to it.

rdzhabarov:
> For example, I believe we could remove the RELU here and just set the output quantization parameters of the Conv to have min = 0, no?

Yup, can be done as a separate optimization.

lib/Graph/Nodes.cpp
```diff
@@ -248,7 +248,11 @@ static void verifyArithmetic(NodeValue LHS, NodeValue RHS, NodeValue res) {
 }
 
 static void verifyRelu(NodeValue src, NodeValue dest) {
-  checkSameType(src, dest);
+  if (src.getType()->isQuantizedType()) {
+    checkSameShape(src, dest);
```
jfix71: Maybe add a check here that the output quantization params min must be non-negative?

tests/unittests/quantizationTest.cpp
```diff
@@ -138,6 +138,35 @@ TEST(Quantization, quantizeGraph) {
   EE.run({}, {});
 }
 
+TEST(Quantization, quantizeReLU) {
```
jfix71: nit: add a comment describing the purpose of the test

tests/unittests/quantizationTest.cpp
```diff
@@ -138,6 +138,35 @@ TEST(Quantization, quantizeGraph) {
...
+      {NodeQuantizationInfo::generateNodeOutputName(input->getName()),
+       {0.2f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(relu->getName()),
+       {0.2f, -128}}};
+
+  F = quantization::quantizeFunction(EE, QI, F);
+
+  EE.compile(CompilationMode::Infer, F);
+
+  ASSERT_TRUE(llvm::isa<MaxNode>(F->getNodeByName("relu1")));
```
jfix71: Where does `"relu1"` come from? Is it hardcoded somewhere not in this test? In general I think it's better practice to access nodes in the compiled graph by going through the Variables, in this case either `"input"` or the SaveNode's Variable. I think that would be less fragile.

lib/Graph/Nodes.cpp
```diff
@@ -247,8 +247,14 @@ static void verifyArithmetic(NodeValue LHS, NodeValue RHS, NodeValue res) {
   checkSameShape(LHS, RHS);
 }
 
-static void verifyRelu(NodeValue src, NodeValue dest) {
-  checkSameType(src, dest);
+static void verifyRelu(NodeValue result, NodeValue input) {
+  if (input.getType()->isQuantizedType()) {
+    assert(result.getType()->isQuantizedType());
+    checkSameShape(result, input);
+    assert(result.getType()->getOffset() == -128 && "Min fp32 value should be 0");
```
jfix71: nit: clang-format

tests/unittests/quantizationTest.cpp
```diff
@@ -138,6 +138,37 @@ TEST(Quantization, quantizeGraph) {
...
 }
 
+/// Quantize ReLU node and make sure that quantized version
+/// has quantization parameters mapping to non-negative floating
+/// point range.
+TEST(Quantization, quantizeReLU) {
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  auto *input = mod.createVariable(ElemKind::FloatTy, {1, 3}, "input");
```
jfix71: nit: perhaps make this `VisibilityKind::Public`, because in theory we could add a new optimization later that folds a `max(splat, private Variable)` into the Variable (assuming the quantize node is also folded into the Variable), and so this test would break.

PR #1511, opened 2018-08-23
[WIP] Make softmax accept a different scale/offset for input & output

tlepley-cadence:
This is a WIP PR related to issue #1512 

- Add `Type::isSimilar` to enable less strict type checking for quantized types
- Enables `Softmax `node and instruction to have a different type for the output with regard to the input


lib/Graph/Nodes.cpp
```diff
@@ -228,7 +234,7 @@ static void verifyTanh(NodeValue src, NodeValue dest) {
 }
 
 static void verifySoftMax(NodeValue src, NodeValue dest) {
-  checkSameType(src, dest);
+  checkSimilarType(src, dest);
```
nadavrot: I think that isSimilar is not a very good name. Why not just compare the element type? It will work for integer and floats.

tlepley-cadence: I agree, the name is unclear :-) What I want to compare is not only the element type, but also the shape too. In other words, everything should be compared, except the quantization information (scale / offset)

jfix71: The shape is already checked in the next line (which could just be replaced with `checkSameShape()` to make that clear).

So you could just replace this `checkSimilarType()` with:
```
assert(src.getElementType() == dest.getElementType());
```

rdzhabarov:
The direction looks good overall, I assume you are going to fix the quantization procedure for SoftMax next?

rdzhabarov:
clicked the wrong button...

lib/Graph/Nodes.cpp
tlepley-cadence: @nadavrot @jfix71 OK, I removed 'isSimilar', and like for some other nodes, I just check the equality for the shape like for arithmetic operation.

tlepley-cadence:
@rdzhabarov I already changed `Quantization/Quantization.cpp` in this PR so that the quantization profile information is used for the output, what was not the case before.

include/glow/Graph/Graph.h
```diff
@@ -270,7 +270,7 @@ class Function final : public Named {
   LogNode *createLog(llvm::StringRef name, NodeValue input);
 
   SoftMaxNode *createSoftMax(llvm::StringRef name, NodeValue input,
```
nadavrot: @tlepley-cadence @rdzhabarov Most of our builders overload two versions of the builders: one with outTy and one without. Making outTy optional could be a good idea, but we need to be consistent. Not sure how to enforce the mandatory out type in a sane way.

tools/ClassGen/InstrGen.cpp
```diff
@@ -147,8 +147,7 @@ int main(int argc, char **argv) {
   BB.newInstr("SoftMax")
       .addOperand("Dest", OperandKind::Out)
       .addOperand("Src", OperandKind::In)
-      .inplaceOperand({"Dest", "Src"})
-      .autoVerify(VerifyKind::SameType, {"Dest", "Src"})
+      .autoVerify(VerifyKind::NoVerify)
```
jfix71: Can we do `VerifyKind::SameShape` here instead? I'd like to avoid `NoVerify` as much as possible.

include/glow/Graph/Graph.h
rdzhabarov: @nadavrot, I agree, let's do this consistently. I think there are two cases:
* Output shape calculations are based on the input shapes, and we do that inside the method which does not take outTy explicitly. The one which takes outTy just do the verification.
* We simply pass outTy as input.getType() to the other method

For the second case, we can eliminate one of the overloads and just check outTy for nullability. For the first case, we'd need to keep two overloads.

I think the way it's handled in the PR is the right way. Once this is merged I'll make sure we follow this practice elsewhere in the Graph.(h|cpp).

nadavrot: Ok.

tools/ClassGen/InstrGen.cpp
tlepley-cadence: Done. I added the shape checking to `InstrGen.cpp`.

lib/Graph/Nodes.cpp
```diff
@@ -228,8 +228,11 @@ static void verifyTanh(NodeValue src, NodeValue dest) {
 }
 
 static void verifySoftMax(NodeValue src, NodeValue dest) {
-  checkSameType(src, dest);
-  assert(src.dims() == dest.dims() && "Invalid shape");
+  if (src.getType()->isQuantizedType()) {
+    checkSameShape(src, dest);
```
jfix71: nit: Did you decide against also verifying src and dest have the same element type? Why not have it here?

tlepley-cadence: @jfix71 I actually did not add the element type equality check on purpose. For quantized networks, I think in general that we can't necessarily force the output element type to be the same as the input element type. I think this choice should be left to the quantization algorithm and it can depend on the target.

tlepley-cadence: Not testing the element type seems to be controversial and since I don't want to delay too much this PR that is primarily about solving a scale/offset issue, I reintroduced the element type checking for softmax. I'm currently doing some experiments about quantization and we can re-discuss more globally this element type problematic later in a different PR or issue.
As a side note, I've noticed that U8 is now handled in Glow not by introducing a new element type, but by acting on the S8 offset, so checking the element type equality between source and destination of nodes should not cause any issue in case the input and the output types of a node don't have the same signedness.

rdzhabarov:
Thank you @tlepley! Assuming we can close https://github.com/pytorch/glow/issues/1512

Note, you can specify "closes #Issue" in the description of the PR and then once it's merged, the issue will be closed automatically.

PR #1516, opened 2018-08-23
Move getRawDataPointer() to getUnsafePtr() where appropriate

jfix71:


jfix71:
In these specific instances it just made more sense to use the `getUnsafePtr()` instead -- we didn't care about the type of the pointer.

There are just a few more places (in two unittests) of `getRawDataPointer()` being used outside of `Tensor.h`. I could move them to casting `getUnsafePtr()`, and then move `getRawDataPointer()` to private. What do you think?

artemrakhov:
I think it's great idea. We don't want too many ways to do the same thing.

jfix71:
Done.

PR #1518, opened 2018-08-23
Fast pass for onnxGetBackendCompatibility

yinghai:
https://github.com/pytorch/glow/pull/1503 had some progress in creating the fast path for `onnxGetBackendCompatibility`, but it's not complete. This PR makes the pass faster and more complete. The reasoning is that preview one won't cover all the case aayway, why don't we just make a quick fact check as what type of ops we support. @rdzhabarov 

Test Plan
With this PR, C2-Glow integration can run through resnet50 correctly. 

$ pytest -s caffe2/python/onnx/test_onnxifi.py


yinghai:
Closing in favor of https://github.com/pytorch/glow/pull/1519

PR #1519, opened 2018-08-23
[onnxifi] Support more operations.

rdzhabarov:
Cover all operations from resnet50 in the `compatibility` call.
cc: @yinghai @Maratyszcza

### DIFF
--- a/lib/Importer/ONNXIFILoader.cpp
+++ b/lib/Importer/ONNXIFILoader.cpp
@@ -135,37 +135,61 @@ std::unique_ptr<ModelLoader> ModelLoader::parse(
   return loader;
 }
 
-std::unique_ptr<std::pair<Kinded::Kind, ElemKind>>
+std::vector<std::pair<Kinded::Kind, ElemKind>>
 ModelLoader::parseOperator(const void *onnxModel, size_t onnxModelSize) {
-
+  std::vector<std::pair<Kinded::Kind, ElemKind>> result;
   ONNX_NAMESPACE::ModelProto modelDef;
-  if (ONNXModelLoader::loadProto(modelDef, onnxModel, onnxModelSize)) {
-    return nullptr;
+  if (!ONNXModelLoader::loadProto(modelDef, onnxModel, onnxModelSize)) {
+    return result;
   }
 
   ONNX_NAMESPACE::GraphProto graph = modelDef.graph();
 
   // Only single operator is allowed to be in the onnxModel.
   if (graph.node_size() != 1) {
-    return nullptr;
+    return result;
   }
 
-  std::unique_ptr<std::pair<Kinded::Kind, ElemKind>> result;
   const std::string &operation = graph.node(0).op_type();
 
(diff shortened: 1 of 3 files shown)


yinghai:
@rdzhabarov  Check an alternative solution that covers all the ops that Glow supports now for ONNX. Neither solution is a complete check as when we convert one caffe2 op, it may result in multiple onnx nodes and one onnx node may translate to multiple Glow node. The translate logic is pretty specific. My solution is just check whether the type of op is supported.

rdzhabarov:
We also need to check whether specific backend supports the operation, that's what's done here.

We could support parsing of specific ONNX operation but it does not guarantee that the backend can execute it.

yinghai:
Then we need to address this: 
> Neither solution is a complete check as when we convert one caffe2 op, it may result in multiple onnx nodes and one onnx node may translate to multiple Glow node.

rdzhabarov:
> We do need to care about multiple ONNX nodes, because one C2 node will be translated into multiple ONNX nodes

Oh, I though compatibility check is always for one entity/node
cc: @Maratyszcza

yinghai:
Yeah, I think we might need to adjust the wording of it as it might not be the case.

lib/Importer/ONNXIFILoader.cpp
```diff
@@ -135,36 +135,71 @@ std::unique_ptr<ModelLoader> ModelLoader::parse(
...
-  std::unique_ptr<std::pair<Kinded::Kind, ElemKind>> result;
   const std::string &operation = graph.node(0).op_type();
 
   // Quantized and non-quantized operations are handled by
   // different ONNX operators, for now only handle fp32.
   // TODO: Add more operators.
-  if (operation == "conv") {
-    result.reset(new std::pair<Kinded::Kind, ElemKind>(
+  if (operation == "BatchNormalization") {
+    result.push_back(std::pair<Kinded::Kind, ElemKind>(
```
yinghai: nit: you can just do `result.emplace_back(Kinded::Kind::BatchNormalizationNodeKind, ElemKind::FloatTy)`

lib/Importer/ONNXIFILoader.cpp
```diff
@@ -135,36 +135,54 @@ std::unique_ptr<ModelLoader> ModelLoader::parse(
...
-  std::unique_ptr<std::pair<Kinded::Kind, ElemKind>> result;
   const std::string &operation = graph.node(0).op_type();
 
   // Quantized and non-quantized operations are handled by
   // different ONNX operators, for now only handle fp32.
   // TODO: Add more operators.
-  if (operation == "conv") {
-    result.reset(new std::pair<Kinded::Kind, ElemKind>(
-        Kinded::Kind::ConvolutionNodeKind, ElemKind::FloatTy));
+  if (operation == "BatchNormalization") {
```
rdzhabarov: talked to @opti-mix. Will change that to StringSwitch and add some macro for  lines like result.emplace_back(Kinded::Kind::BatchNormalizationNodeKind, ElemKind::FloatTy);

PR #1520, opened 2018-08-23
Move ElemKind::IndexTy to ElemKind::Int64ITy, and handles from size_t to int64_t

jfix71:
We often need to load weights from ONNX/Caffe2 that are typed explicitly as unsigned/signed and with a bitwidth. For example, ONNX has `ONNX_NAMESPACE::TensorProto::INT64`, and Caffe2 has `GivenTensorInt64Fill`.

This allows us to load such protos, instead of failing when seeing them. From here we could add new ElemKinds as well, e.g. `ElemKind::Int32ITy`, and then shrink the size of `ElemKind::Int64ITy` tensors to `ElemKind::Int32ITy` if we know they could fit in 32 bits instead.

### DIFF
--- a/lib/Importer/ONNXIFILoader.cpp
+++ b/lib/Importer/ONNXIFILoader.cpp
@@ -34,8 +34,7 @@ static void setTensorType(const ONNX_NAMESPACE::TypeProto &in, Tensor *T) {
     T->reset(ElemKind::FloatTy, dim);
   } else if (in.tensor_type().elem_type() ==
              ONNX_NAMESPACE::TensorProto::INT64) {
-    // TODO: either switch IndexTy to be 64 bit, or switch to another type here.
-    T->reset(ElemKind::IndexTy, dim);
+    T->reset(ElemKind::Int64ITy, dim);
   } else {
     assert(false && "Only float and index tensors are supported");
   }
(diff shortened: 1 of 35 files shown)


nadavrot:
Nice work! Thank you for cleaning things up.

lib/Importer/ONNXIFILoader.cpp
```diff
@@ -77,10 +76,9 @@ static bool loadWeight(const onnxTensorDescriptorV1 &in, Tensor *T) {
       TH.raw(i) = data[i];
     }
   } else if (in.dataType == ONNXIFI_DATATYPE_UINT64) {
```
yinghai:
```
} else if (in.dataType == ONNXIFI_DATATYPE_UINT64 || in.dataType == ONNXIFI_DATATYPE_INT64) {
```
?

lib/Importer/ONNXIFILoader.cpp
```diff
@@ -77,10 +76,9 @@ static bool loadWeight(const onnxTensorDescriptorV1 &in, Tensor *T) {
...
-    // TODO: either switch IndexTy to be 64 bit, or switch to another type here.
-    T->reset(ElemKind::IndexTy, dims);
+    T->reset(ElemKind::Int64ITy, dims);
 
-    auto TH = T->getHandle<size_t>();
+    auto TH = T->getHandle<int64_t>();
```
opti-mix: Question: `ONNXIFI` uses `ONNXIFI_DATATYPE_UINT64`, but we use a singed `int64_t` here. It is intentional? The value range of `int64_t` is smaller.

jfix71: Great catch! The next line here casts the data pointer into `int64_t`, and so I didn't notice it was based on `ONNXIFI_DATATYPE_UINT64`.

@yinghai @rdzhabarov Was this intentional? Perhaps this should be kept `uint64_t`, and then I can add an assert in the loop below here loading the tensor data to ensure we don't overflow `int64_t`? At least until we decide if we want to add `ElemKind::UInt64ITy`.

yinghai: Yes. I think uint64_t is fine for now. The only use case for int64_t so far is shape size which should be nonnegative.

lib/Importer/ONNXIFILoader.cpp
jfix71: Done.

lib/Importer/ONNXIFILoader.cpp
jfix71: Done -- I added an assert here in the loop.

PR #1521, opened 2018-08-24
[AOT] Add an option to specify the name of the network being bundled

qcolombet:
Hi @opti-mix,

This goes with the alternative we discussed for #1475.

One question for you, while updating the AOT docs, I started to wonder if, for the default network name, instead of defaulting to "net", we shouldn't use the directory name of the input model like we used to do.

I believe in that context fiddling with the input path to extract a name is acceptable.

What do you think?


opti-mix:
@qcolombet 

> One question for you, while updating the AOT docs, I started to wonder if, for the default network name, instead of defaulting to "net", we shouldn't use the directory name of the input model like we used to do.
I believe in that context fiddling with the input path to extract a name is acceptable.

Yeah, it sounds reasonable. Alternatively, we can just say that there is no good default value and the user has to explicitly provide the `-network-name` option with a reasonable value when `-emit-bundle` is used. This way we don't need to fiddle with the input path.

qcolombet:
I went with extracting the network name from model path by default because:
- We had the code around to do so from #1475 
- It felt redundant to have to specify the network name in the AOT.md examples since it was exactly the same as the name in `-model`

tools/loader/Loader.cpp
```diff
@@ -116,17 +116,85 @@ llvm::cl::opt<std::string>
...
+    llvm::cl::cat(loaderCat));
 } // namespace
 
 bool glow::emittingBundle() { return !emitBundle.empty(); }
 
+/// Given a string representing a \p path of the form
+/// "path/to/dir" returns the string representing the last
+/// directory (in this case "dir").
+/// \note This is similar to the basename(1) command line tool.
+llvm::StringRef getLastDirectoryName(llvm::StringRef path) {
```
opti-mix: @qcolombet LLVM provides convenience APIs for working with file systems. See e.g. `lib/Support/Path.h`. Eventually this whole function can be replaced by a couple of calls of those APIs.

qcolombet: Oh thanks!
I didn't know LLVM had that!!

qcolombet: Much better with the LLVM implementation, though you have to read the fine print :).
In particular, this part:
/// Path iterator.
[...]
/// * Dot, if one or more trailing non-root slash characters are present.

tests/unittests/LLVMIRGenTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "AllocationsInfo.h"
+
+#include "glow/IR/IR.h"
+
+#include "gtest/gtest.h"
+
+using namespace glow;
+
+#ifndef GLOW_WITH_CPU
+#error "This shouldn't be compiled without the CPU backend"
```
rdzhabarov: nit: avoiding to negative things :) "This should be compiled with the CPU backend"

tests/unittests/LLVMIRGenTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+#endif
+
+/// Check that we get/setMainEntryName behaves in a sane way.
```
rdzhabarov: nit: grammar

PR #1525, opened 2018-08-24
[Type.h] Add type Int16QTy

aturetsk:
Add int16 type to be able to use int16 tensors in back-end. Not used in the middle-end at the moment.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -181,6 +181,7 @@ inline bool operator==(const ShapeNCHW &LHS, const ShapeNCHW &RHS) {
 enum class ElemKind : unsigned char {
   FloatTy,  // 32-bit float type (float)
   Int8QTy,  // 8-bit quantized type (int8_t)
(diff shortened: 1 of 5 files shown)


include/glow/Base/Type.h
```diff
@@ -181,6 +181,7 @@ inline bool operator==(const ShapeNCHW &LHS, const ShapeNCHW &RHS) {
 enum class ElemKind : unsigned char {
   FloatTy,  // 32-bit float type (float)
   Int8QTy,  // 8-bit quantized type (int8_t)
+  Int16QTy, // 16-bit quantized type (int16_t)
```
rdzhabarov: can you explain the usage of this type? Is that for some intermediate results on int8 tensors?

aturetsk: Possible usages:
* intermediate results
* deal with our internal backend constraints (can't give details)
* in the long term we want to extend quantization to support 16 bits

PR #1527, opened 2018-08-25
Convert the Shapes from tuple<size_t> to tuple<uint64_t> so that cosistence with kernel.cl 's Shape definition.

lygstate:
Resolve a inssue sizeof(size_t) would be 4 on x86

platform(or armeabi) And in OpenCL we defines ShapeNHWC ShapeNCHW
PaddingTLBR ShapeHW to be tuple of uint64_t,
we need to make the parameter to be consistence.
Setting OCL Shapes properly to make sure clSetKernelArg
consistence with type struct defined in kernel.cl:

// The types of elements should be always matching the definitions of
```
// ShapeNHWC in Type.h
typedef struct {
  cl_uint64_t n; // Number of samples
  cl_uint64_t h; // Height
  cl_uint64_t w; // Width
  cl_uint64_t c; // Number of channels
} ShapeNHWC;

typedef struct {
  cl_uint64_t n; // Number of samples
  cl_uint64_t c; // Number of channels
  cl_uint64_t h; // Height
  cl_uint64_t w; // Width
} ShapeNCHW;

typedef struct {
  cl_uint64_t height;
  cl_uint64_t width;
} ShapeHW;

// The types of elements should be always matching the definitions of
// PaddingTLBR in Type.h
typedef struct {
  cl_uint64_t top;
  cl_uint64_t left;
  cl_uint64_t bottom;
  cl_uint64_t right;
} PaddingTLBR;
```

The type in Type.h
```
struct ShapeNHWC {
  size_t n; // Number of samples
  size_t h; // Height
  size_t w; // Width
  size_t c; // Number of Channels

struct ShapeNCHW {
  size_t n; // Number of samples
  size_t c; // Number of Channels
  size_t h; // Height
  size_t w; // Width


struct PaddingTLBR {
  size_t top;
  size_t left;
  size_t bottom;
  size_t right;

struct ShapeHW {
  size_t height;
  size_t width;
```

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -202,6 +202,41 @@ void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
   GLOW_ASSERT(err == CL_SUCCESS && "Unable to set parameter");
 }
 
+// sizeof(size_t) would be 4 on x86 platform or armeabi platform.
+// And in kernel.cl we defines ShapeNHWC ShapeNCHW PaddingTLBR ShapeHW to be tuple of uint64_t.
+// But the ShapeNHWC ShapeNCHW PaddingTLBR ShapeHW in Type.h are tuple of size_t.
+// this function are doing the convert work that convert tuple<size_t> to tuple<uint64_t>
+template <class T>
+void setKernelArgShape(cl_kernel kernel, unsigned argIdx, T shapeValue) {
+  const uint32_t shapeSize = sizeof(shapeValue) / sizeof(size_t);
+  uint64_t values[shapeSize];
+  for (int i = 0; i < shapeSize; ++i) {
+    values[i] = ((size_t *)&shapeValue)[i];
+  }
+  cl_int err = clSetKernelArg(kernel, argIdx, sizeof(values), &(values[0]));
+  GLOW_ASSERT(err == CL_SUCCESS && "Unable to set parameter");
+}
+
+template <>
+void setKernelArg(cl_kernel kernel, unsigned argIdx, ShapeNHWC value) {
+  setKernelArgShape(kernel, argIdx, value);
+}
+
+template <>
+void setKernelArg(cl_kernel kernel, unsigned argIdx, ShapeNCHW value) {
+  setKernelArgShape(kernel, argIdx, value);
+}
+
+template <>
+void setKernelArg(cl_kernel kernel, unsigned argIdx, PaddingTLBR value) {
+  setKernelArgShape(kernel, argIdx, value);
+}
+
+template <>
+void setKernelArg(cl_kernel kernel, unsigned argIdx, ShapeHW value) {
+  setKernelArgShape(kernel, argIdx, value);
+}
+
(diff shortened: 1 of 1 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -202,6 +202,37 @@ void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
   GLOW_ASSERT(err == CL_SUCCESS && "Unable to set parameter");
 }
 
+template<class T>
+void setKernelArgShape(cl_kernel kernel, unsigned argIdx, T shapeValue) {
```
opti-mix: Can you explain in the source code comment what is going on in this function and why? 

Are you trying to ensure a proper alignment?

opti-mix: If this is to ensure the alignment or padding, is it required by the OpenCL spec? It would be interesting to have a reference then.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -202,6 +202,37 @@ void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
...
+  const uint32_t shapeSize = sizeof(T) / sizeof(size_t);
```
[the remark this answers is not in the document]
lygstate: There is no differnces, the T is limit to ShapeNHWC ShapeNCHW PaddingTLBR ShapeHW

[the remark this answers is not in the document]
lygstate: What does sizeof(shapeSize)  mean? there is no such a usage.

lib/Backends/OpenCL/OpenCL.cpp
lygstate: Resolve a inssue sizeof(size_t) would be 4 on x86 platform(or armeabi)
And in OpenCL we defines ShapeNHWC ShapeNCHW PaddingTLBR ShapeHW to be tuple of
uint64_t, we need to make the parameter to be consistence.

lib/Backends/OpenCL/OpenCL.cpp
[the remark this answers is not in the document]
opti-mix: > What does sizeof(shapeSize) mean? there is no such a usage.

I meant `sizeof(shapeValue)`, i.e. `sizeof(T)`

lib/Backends/OpenCL/OpenCL.cpp
lygstate: It's not about OpenCL spec.
But the types defined  in kernel.cl

nadavrot:
@lygstate The message "Setting OCL Shapes properly" is not a descriptive enough. I can't understand from your commit message what problem this PR addresses and why this is a good solution to the problem. I am not saying that the commits are not good, just that I don't understand what's going on. Please update the commit and pull request with some text that explains:
1. What is the problem.
2. How is the commit fixing the problem.

opti-mix:
@lygstate Could you try #1547. I think it solves the issue in cleaner and more future-proof way.

PR #1528, opened 2018-08-25
Fixes for LLVM 7.

lygstate:


nadavrot:
@lygstate Please see my message on the other commits.  Your commit messages need to explain what problem is being solved and how. You left an empty description commit message. Please update the commit message with some text.

PR #1533, opened 2018-08-25
Fixes libjit under MSVC compiling.

lygstate:


nadavrot:
@lygstate I have a few problem with this commit. First, please don't add LIBJIT_VLA. We don't need it. You can change the few uses of variable-length array to fixed-size arrays. Please take a look at the code. 

Second, this commit bundles a bunch of changes that make it difficult for me to review this code. Please split this commit to small changes that people can review.

lygstate:
@nadavrot What's tie fixd-size length? clarify it clearly.

lygstate:
@nadavrot Are you people refuse to add MSVC support, if that the plan, then I would cancel the work. This pull request is small enough and can not be split.

lygstate:
@nadavrot   What's the maximal size of ywidth?

lygstate:
@nadavrot You've tolkd me these small changes a too big to review, for me, the code is even harder to read. tell clearly where the maximal possible size that doesn't need the VLA support.

nadavrot:
@lygstate For the max number of dimensions you can use the value 10. For the max number of registers you can use the value 32. Please add a comment that explains the max number of supported reg/dims.

lygstate:
how about ywidth and depthUnroll

nadavrot:
@lygstate 32 should be large enough.

PR #1535, opened 2018-08-26
Add float16 types and also define it properly in MSVC

lygstate:


nadavrot:
This code is compiled by clang to generate LLVM bitcode. Why do you need to make it compile with MSVC?

PR #1537, opened 2018-08-26
[msvc] Add float16 types and also define it properly for MSVC

lygstate:


nadavrot:
This PR does not make sense. Did you try to run it? Your float16 definition is not even a vector. 

Also, why do you need to compile libjit with MSVC?  The code needs to be compiled with clang to generate LLVM bytecode.

qcolombet:
I agree with @nadavrot, this PR doesn't make sense.
Are you trying to add float16 as a vector of 16 floats or half-precision floating point type?
Nonetheless, this code does neither!

lygstate:
Gocha, I've done the wrong things and have no idea how to deal with it

lygstate:
It's seems there is no equivalance for ext_vector_type in MSVC, don't know how to go.

qcolombet:
You can try to scalarize the uses of that type (that may not be possible, I haven’t checked how we use it)

nadavrot:
@lygstate @qcolombet Maybe we should just disable the gemmbench build on Windows. We have a strong dependency on LLVM in libjit, because we need to build bitcode. The only non-jit usage of libjit is in the gemm benchmark unit test. We can just disable this specific test on windows.

qcolombet:
Works for me too.

PR #1540, opened 2018-08-26
[msvc] Add posix_memalign and _aligned_free for MSVC.

lygstate:


nadavrot:
More info here: 

https://stackoverflow.com/questions/39625352/why-do-some-functions-in-c-have-an-underscore-prefix

lygstate:
uppercase or lowercase？

PR #1542, opened 2018-08-26
[msvc][libjit] Add VLA support in libjit for MSVC.

lygstate:


nadavrot:
@lygstate We had a few discussions on this VLA is not needed. You can just change the array to MAX_DIMS. Also, this code is only built with clang to build LLVM bitcode so we don't need MSVC support.

PR #1544, opened 2018-08-26
[msvc][opencl] uint64_t is not consistence with enqueueKernel's funct…

lygstate:
…ion parameter, change to size_t instead.


lygstate:
@qcolombet I've updated without modify the header.

PR #1545, opened 2018-08-26
[msvc] define ssize_t for MSVC.

lygstate:
Refer to https://stackoverflow.com/questions/22265610/why-ssize-t-in-visual-studio-2010-is-defined-as-unsigned

### DIFF
--- a/lib/Backends/CPU/libjit/libjit_defs.h
+++ b/lib/Backends/CPU/libjit/libjit_defs.h
@@ -19,6 +19,11 @@
 #include <stdint.h>
 #include <string.h>
 
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 typedef float float4 __attribute__((ext_vector_type(4)));
 typedef float float8 __attribute__((ext_vector_type(8)));
 


lygstate:
@nadavrot You make comment too early, do you really familer with libjit, if you don't, please do not review these code and close the pull request without thinking properly.
If I didn't face the problem, I won't create a pull request it.

This is the orignal CMake file, please look it carefully:
``

add_library(CPURuntime
            SHARED
              libjit/libjit.cpp
              libjit/libjit_conv.cpp
              libjit/libjit_matmul.cpp)
set_target_properties(CPURuntime
                      PROPERTIES
                        CXX_STANDARD 11)
target_compile_options(CPURuntime
                       PRIVATE
                         -ffast-math
                         -g0
                         -emit-llvm
                         -O0)
# NOTE(abdulras) explicitly override the compiler and linker invocations with
# custom rules.  The trailing `#` is the comment leader intended to nullify the
# compile and link commands.  Doing this allows us to override the compiler
# (driver) and the linker used for building the runtime which requires clang (to
# emit LLVM IR for the LTO'ed AOT JIT runtime) and uses the `llvm-link` as the
# linker to merge the object files.
set_target_properties(CPURuntime
                      PROPERTIES
                        RULE_LAUNCH_COMPILE
                          "${CLANG_BIN} <DEFINES> <INCLUDES> <FLAGS> -o <OBJECT> -c <SOURCE> # "
                        RULE_LAUNCH_LINK
                          "${LLVM_LINK_BIN} -o ${CMAKE_BINARY_DIR}/libjit.bc <OBJECTS> # "
                        OUTPUT_NAME
                          libjit.bc)

add_library(CPURuntimeNative
              libjit/libjit.cpp
              libjit/libjit_conv.cpp
              libjit/libjit_matmul.cpp)`
``

lygstate:
If that possible to remove the dependence to this from that unittest, that's OK for me. Keep libjit works with msvc is not easy.

PR #1547, opened 2018-08-26
[opencl] Define a type cl_host_size_t exactly matching the type size_t used on the host side

opti-mix:
This is required to e.g. properly pass struct parameters of types like ShapeNHWC, ShapeNCHW, etc. The definitions of these types on the host side use size_t for their members and they should be defined on the OpenCL's side using integer types of the same width.

Since the size of size_t on the host side is not guaranteed to be of a fixed-width known to OpenCL, it has to be provided to OpenCL dynamically. Thus, the size is passed down to the OpenCL code as a macro definition. This macro definition is then used to define in OpenCL the type cl_host_size_t matching the size_t on the host side.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -82,6 +82,19 @@ static void dumpCompileLog(cl_device_id dev, cl_program prog) {
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


lygstate:
@opti-mix I need have a test it for x86 platform.

lygstate:
Works for me.

PR #1562, opened 2018-08-29
Added the minimum needed to make the CPU backend inheritable

tlepley-cadence:
The CPU backend contains all the LLVM IR generation logic and we found very useful to have the possibility to create a backend derived from the CPU backend by simply enabling these 3 possibilities:
- add the possibility to inherit from `CPUBackend`
- add the possibility to call `CPUBackend::compile()` with an object derived from `LLVMIRGen`
- make `LLVMIRGen::generateLLVMIRForInstr` virtual

### DIFF
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -47,7 +47,8 @@ namespace {
 /// propagate them into relative addressing computations and the like and
 /// produce a very efficient code that uses absolute addressing whenever
 /// possible.
-static void emitJitMain(AllocationsInfo &allocationsInfo, LLVMIRGen &irgen) {
+static void emitJitMain(LLVMIRGen &irgen) {
+  AllocationsInfo &allocationsInfo = irgen.getAllocationsInfo();
   llvm::Type *voidTy = llvm::Type::getVoidTy(irgen.getLLVMContext());
   llvm::FunctionType *jitFuncTy = llvm::FunctionType::get(voidTy, {}, false);
   auto *func =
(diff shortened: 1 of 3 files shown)


lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -115,14 +115,13 @@ static void *allocateJITMemory(const IRFunction *F,
 } // end namespace
 
 std::unique_ptr<CompiledFunction>
-CPUBackend::compile(std::unique_ptr<IRFunction> IR) const {
-  AllocationsInfo allocationsInfo;
-  LLVMIRGen irgen(IR.get(), allocationsInfo, "");
+CPUBackend::compileWithIRGen(IRFunction *IR, LLVMIRGen &irgen,
```
qcolombet: Instead of having another compile method, could we add a virtual getIRGen method?

The reason I am saying that is because it feels strange to have CPUBackend being non-final will not having any virtual method.

tlepley-cadence: @qcolombet I don't have a strong opinion on this. It works for us in both ways. I implemented a solution that I believe implements your suggestion, by adding the virtual `createIRGen` method to `CPUBackend`.

qcolombet:
@tlepley-cadence, you'll need to rebase the patch and fix the conflicts for me to merge it.
Cheers.

tlepley-cadence:
@qcolombet sorry, I had not noticed there were a merge conflict. I just squashed the 2 commits and rebased. I should be fine now.

PR #1565, opened 2018-08-29
[GraphOptimizer] Remove useless assignment from DCE

qcolombet:
*Description*:
Another assignment of this boolean is done first thing at the beginning
of the next do;while loop.
This useless assignment triggers annoying warnings in our internal flow.

NFC.

*Testing*:
ninja check

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -70,9 +70,8 @@ static void DCE(Function *F) {
 
   // Remove unused nodes. Do not remove unused vars because they are the
   // interface to the user program.
-  bool changedLocally = true;
-  do {
-    changedLocally = false;
+  while (true) {
+    bool changedLocally = false;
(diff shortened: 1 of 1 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -70,9 +70,8 @@ static void DCE(Function *F) {
 
   // Remove unused nodes. Do not remove unused vars because they are the
   // interface to the user program.
-  bool changedLocally = true;
-  do {
-    changedLocally = false;
+  while(true) {
```
jfix71: nit: I think you forgot to `clang-format`, as I believe there should be a space after `while` 😃

qcolombet: I believe you're right :)

qcolombet: Done.

PR #1566, opened 2018-08-29
[CPU] Improve the 2D transpose implementation.

nadavrot:
*Description*:
This commit "tiles" the 2d transpose method. Processing the input matrix
in small blocks ensures that the source and destination will stay in
cache. Notice that scanning 1 row at a time and storing the scalar into
some column brings and invalidates many cache lines. The tiled scan
helps to reduce the number of used destination cache lines. I measured a
4x speedup on a small local benchmark. We could further improve this
procedure using vector instructions for the transpose of the single tile
and with recursive blocking of the 2d matrix.

*Testing*:
I ran a local benchmark (described above). Ran ninja check and the ./run.sh test to run inference on the ONNX models. 

*Documentation*:
I updated the comment in the implementation of the operation.

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -397,6 +397,11 @@ template <typename T>
 void libjit_transpose_generic(const T *inW, T *outW, const size_t *idim,
                               const size_t *odim, const size_t *shuffle,
                               size_t numDims) {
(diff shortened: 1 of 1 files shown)


nadavrot:
@qcolombet Okay. Yes, it would make sense to have such a benchmark. I like @bertmaher 's gemm benchmark and I want to have something similar for convolutions as well.

qcolombet:
Thanks!
Could be a different PR so that we get the nice speed up already :).

PR #1571, opened 2018-08-30
[cleanup] Update the comments and move a function. NFC.

nadavrot:
*Description*:

Before this commit the 'copy' method was listed under the 'convolution'
section. This commit moves the 'copy' method and adds a few labels. I
found this when I was reading the reference implementation for our
convolutions and decided to clean it up.

*Testing*: Ran the tests locally. 

*Documentation*: This commit just adds a comment.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -30,12 +30,6 @@ using namespace glow;
 //                       Convolution
 //===----------------------------------------------------------------------===//
 
-void InterpreterFunction::fwdCopyInst(const CopyInst *I) {
-  auto inT = getTensor(I->getSrc());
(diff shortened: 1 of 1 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1559,6 +1568,10 @@ static void fwdTopK(Tensor *outW, Tensor *indW, Tensor *inW, size_t k) {
   }
 }
 
+//===----------------------------------------------------------------------===//
+//                       Sorting operators
+//===----------------------------------------------------------------------===//
+
```
rdzhabarov: can you also fix debugging operations down below?

nadavrot: No problem.

PR #1577, opened 2018-09-04
Add option to disable quantizing specified node kinds

jfix71:
*Description*: We may prefer to not quantize every node kind in a model if it causes unacceptable accuracy loss. This commit adds the ability to disable quantization for node kinds that would otherwise be quantized, preventing accuracy loss.

*Testing*: Added a unit test.

*Documentation*: Added.

### DIFF
--- a/docs/Quantization.md
+++ b/docs/Quantization.md
@@ -102,6 +102,19 @@ the graph.
 ./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=resnet50 -load_profile="profile.yaml"
 ```
 
+By default, all nodes that can be quantized will be quantized. However, we may
+only want to quantize some parts of a model, e.g. if accuracy loss is too high
+when all node kinds are quantized. The Glow loader currently allows for
+disabling quantization of all nodes of a specific kind which are found in the
+graph. For example, if the loaded model sees high accuracy loss when
+element-wise Add is quantized, it can be left in floating point. This can be
+done by passing on the command line the node name via the option
+`-do_not_quantize_nodes`. Multiple node kinds can be specified to not be
+quantized. For example, to not quantize any Add or Div nodes when running the
+quantized text translator:
+
+```./bin/text-translator -m en2gr -load_profile=en2gr.yaml -do_not_quantize_nodes=Add,Div```
+
 ## Compiler Optimizations
 
 Glow features a number of compiler optimizations that transform the compute
--- a/include/glow/Quantization/Quantization.h
+++ b/include/glow/Quantization/Quantization.h
@@ -62,13 +62,16 @@ generateNodeQuantizationInfos(const Function *F,
 /// function based on \p quantizationInfos. This method converts to integer as
 /// many nodes as permitted by the backend \p EE. The new quantized function is
 /// called \p newFuncName. If no name is given the method will generate a name.
-/// This method clones original function \p F and caller is responsible
-/// for cleaning up/erasing original function \p F if needed.
+/// This method clones original function \p F and caller is responsible for
+/// cleaning up/erasing original function \p F if needed. Any nodes of kinds
+/// contained in \p doNotQuantizeKinds will not be quantized, even if a profile
+/// was gathered for them and the backend supports the quantized operation.
 /// \returns a new quantized function.
 Function *
 quantizeFunction(const ExecutionEngine &EE,
                  llvm::ArrayRef<NodeQuantizationInfo> quantizationInfos,
-                 Function *F, llvm::StringRef newFuncName = "");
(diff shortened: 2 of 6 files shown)


include/glow/Quantization/Quantization.h
```diff
@@ -62,13 +62,16 @@ generateNodeQuantizationInfos(const Function *F,
 /// function based on \p quantizationInfos. This method converts to integer as
 /// many nodes as permitted by the backend \p EE. The new quantized function is
 /// called \p newFuncName. If no name is given the method will generate a name.
-/// This method clones original function \p F and caller is responsible
-/// for cleaning up/erasing original function \p F if needed.
+/// This method clones original function \p F and caller is responsible for
+/// cleaning up/erasing original function \p F if needed. Any nodes of kinds
+/// contained in \p doNotQuantizeKinds will not be quantized, even if a profile
+/// was gathered for them.
```
rdzhabarov: `if a profile was gathered for them and backend supports quantized operation`?

docs/Quantization.md
```diff
@@ -102,6 +102,17 @@ the graph.
...
+By default, all nodes that can be quantized will be quantized. However, we may
+only want to quantize some parts of a model, e.g. if accuracy loss is too high
+when all node kinds are quantized. The Glow loader currently allows for
+disabling quantization of all nodes of a specific kind which are found in the
+graph. For example, if the loaded model sees high accuracy loss when
+element-wise Add is quantized, it can be left in floating point. This can be
+done by passing on the command line the node name via the option
+`-do_not_quantize_node`. For example, to not quantize the Add node when running
+the quantized text translator: ``` ./bin/text-translator -m en2gr
+-load_profile=en2gr.yaml -do_not_quantize_node=Add ```
```
rdzhabarov: can you mention that `do_not_quantize_node` could contain several node types? I got a sense that it's only a single node type can be blacklisted here.

tests/unittests/quantizationTest.cpp
```diff
@@ -659,6 +659,78 @@ TEST(Quantization, quantizeSoftmaxAndLRN) {
...
+
+  std::vector<NodeQuantizationInfo> QI{
+      {NodeQuantizationInfo::generateNodeOutputName(LHS->getName()), {0.3f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(RHS->getName()), {0.4f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(MMN->getName()), {0.6f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(TN->getName()), {0.5f, 0}},
+  };
+
+  // Do not quantize the tanh.
+  KindSet doNotQuantize;
```
rdzhabarov: does this not work: KindSet doNotQuantize = {Kinded::Kind::TanhNodeKind} ?

docs/Quantization.md
lygstate: Is there a more complete way to do that?
Such as an sub-group of nodes be disabled for quantize?

tests/unittests/quantizationTest.cpp
jfix71: Unfortunately not 😞

docs/Quantization.md
jfix71: Sure -- just to be clear I believe the way to add multiple nodes here is to specify each one individually, e.g. `-do_not_quantize_node=Add -do_not_quantize_node=Div` etc. Hence why I made it node and not nodes. Is that not the case, is there a way to do all under one flag?

jfix71: @lygstate Are you talking about specifying multiple Node kinds, as I replied above? Or by a sub-group of nodes do you mean that only some of the nodes with the Node kind would be quantized?

tests/unittests/quantizationTest.cpp
```diff
@@ -659,6 +659,78 @@ TEST(Quantization, quantizeSoftmaxAndLRN) {
...
+  doNotQuantize.insert(Kinded::Kind::TanhNodeKind);
```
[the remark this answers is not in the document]
jfix71: Done

include/glow/Quantization/Quantization.h
jfix71: Done

docs/Quantization.md
qcolombet: Using comma should work as well, IIRC the LLVM option parsing:
 `-do_not_quantize_node=Add,Div`

jfix71: Figured it out -- I needed to add `llvm::cl::CommaSeparated` to enable that. Thanks @qcolombet!

PR #1578, opened 2018-09-04
[ImageClassifier] Add command line option for the model's expected input image name

jfix71:
*Description*: Models may expect different names for the input image in ImageClassifier. We previously had a hack that allowed for two different names. This properly allows the model to have its image name specified.

*Testing*: updated and verified `run.sh` works

*Documentation*: Added.

### DIFF
--- a/docs/AOT.md
+++ b/docs/AOT.md
@@ -38,7 +38,7 @@ This document demonstrates how to produce a bundle for the host CPU using the
 directory.
 
 ```
-$image-classifier image.png -image_mode=0to1 -m resnet50 -cpu -emit-bundle build/
+$image-classifier image.png -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data -cpu -emit-bundle build/
 ```
 
 The command above would compile the neural network model described by the files
@@ -163,7 +163,7 @@ The makefile provides the following targets:
 * `download_weights`: it downloads the Resnet50 network model in the Caffe2 format.
 * `build/resnet50.o`: it generates the bundle files using the Glow image-classifier as described above.
   The concrete command line looks like this:
-  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -cpu -emit-bundle build`
+  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data -cpu -emit-bundle build`
   It reads the network model from `resnet50` and generates the `resnet50.o`
   and `resnet50.weights` files into the `build` directory.
 * `build/main.o`:  it compiles the `resnet50_standalone.cpp` file, which is the main file of the project.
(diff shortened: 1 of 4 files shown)


nadavrot:
LGTM. Thank you for cleaning this up. Would it make the code more readable to add quotes around the variable name?

-model_expected_input_name="gpu_0/data"

docs/AOT.md
```diff
@@ -163,7 +163,7 @@ The makefile provides the following targets:
 * `download_weights`: it downloads the Resnet50 network model in the Caffe2 format.
 * `build/resnet50.o`: it generates the bundle files using the Glow image-classifier as described above.
   The concrete command line looks like this:
-  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -cpu -emit-bundle build`
+  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -model_expected_input_name=gpu_0/data -cpu -emit-bundle build`
```
tlepley-cadence: `model_expected_input_name` loos a bit long. Also, I don't think it's an expected name, but this is the actual name.
I think that `input_name` or even `input` would be enough. Regarding the quotes around the name, my preference is to not have them, like I believe it's mode common to do. The only advantage of having the quote would be to enable the usage of space in names. It is a use case we want to cover ?

jfix71: Yeah I will remove `expected`. For `model` -- I wanted it to be very clear it has nothing to do with the file name of the image input being classified. I changed it to `--model_input_name` -- it's at least a little shorter.

docs/AOT.md
```diff
@@ -163,7 +163,7 @@ The makefile provides the following targets:
 * `download_weights`: it downloads the Resnet50 network model in the Caffe2 format.
 * `build/resnet50.o`: it generates the bundle files using the Glow image-classifier as described above.
   The concrete command line looks like this:
-  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -cpu -emit-bundle build`
+  `image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -model_input_name=gpu_0/data -cpu -emit-bundle build`
```
rdzhabarov: nit: -m resnet50 =? -m=resnet50 for consistency.

docs/AOT.md
```diff
@@ -186,9 +186,8 @@ To build and run the example, you just need to execute:
 
 This run performs almost the same steps as non-quantized Resnet50 version
 except it emits bundle based on the quantization profile:
-`image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50
--load_profile=profile.yml -cpu -emit-bundle build`
+`image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m resnet50 -model_input_name=gpu_0/data -load_profile=profile.yml -cpu -emit-bundle build`
```
rdzhabarov: ditto

docs/AOT.md
```diff
@@ -186,9 +186,8 @@ To build and run the example, you just need to execute:
...
 
 The `profile.yml` itself is captured at a prior step by executing image-classifier with the `dump_profile` option:
-`image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m resnet50 -dump_profile=profile.yml`.
+`image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m resnet50 -model_input_name=gpu_0/data -dump_profile=profile.yml`.
```
rdzhabarov: ditto

docs/AOT.md
jfix71: Done

jfix71:
For quotes vs. not, I don't have a strong feeling either way. It The other inputs mostly do not use quotes so I was trying to stay consistent. I doubt we would need to support spaces, but even if someone did want to they could just use quotes as is -- the underlying code implementing the option stays the same, so it's really just about how we display it in our docs and in `run.sh`.

jfix71:
@qcolombet I don't see any comments for the test -- and there is no test here other than updating `run.sh`, did you mean to comment something there?

jfix71:
Perhaps this comment was meant for https://github.com/pytorch/glow/pull/1577 ?

qcolombet:
Yes that was for #1577, don’t know how I ended up here :S

PR #1586, opened 2018-09-06
[Placeholder] Add support for Placeholders in the execution engine

nadavrot:
*Description*:

Implement the logic for executing Placeholders in the execution engine. This commit changes the backend interface and adds the ability to pass Placeholder variables in addition to Variables. 

This change breaks the out-of-tree backends.
    
This is the first step in adding support for placeholder variables in the execution engine. The commit adds support to the interpreter but not to the OpenCL and CPU backends. The support in the interpreter is single threaded at the moment.

In the next few PRs I'll complete the execution engine support in different backends for across all node kinds (SaveNode, for example).

*Testing*: Ran tests locally and added a new test.

*Documentation*: No functional changes to document. 

issue #1334

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -113,8 +113,15 @@ void ExecutionEngine::runBatch(size_t iterations,
 void ExecutionEngine::updateInputsAndRunNetwork(llvm::ArrayRef<Storage *> vars,
                                                 llvm::ArrayRef<Tensor *> inputs,
                                                 size_t sampleIdx) {
-
   llvm::SmallVector<Placeholder *, 8> placeholders;
+
+  // This container saves the tensor slices that were extracted from the inputs.
+  // The tensors are alive during the lifetime of this function. These tensors
+  // must not move, and we must pass them as ArrayRef, so we allocate them here
+  // manually.
+  llvm::SmallVector<Tensor, 8> tempTensorViews;
+  // We need to pass the tensors as array ref, so we reference the static
+  // storage above, that's frozen during the execution of the program.
   llvm::SmallVector<Tensor *, 8> tensors;
 
   // Update the input variables.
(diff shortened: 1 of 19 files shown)


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -93,16 +110,30 @@ void ExecutionEngine::runBatch(size_t iterations,
...
+    // removed because it's not threadsafe. See issue #1334.
+    if (Variable *V = llvm::dyn_cast<Variable>(vars[i])) {
+      loadValueFromTensorSlice(V, inputs[i], sampleIdx);
+      continue;
+    }
+
+    // This is a placeholder that we need to make concrete during the execution
+    // of the program.
+    placeholders.push_back(cast<Placeholder>(vars[i]));
+    tensors.push_back(inputs[i]);
```
jfix71: Don't we need to get slices for the placeholders, like we currently do for Variables?

lib/Graph/Graph.cpp
```diff
@@ -2106,6 +2106,24 @@ static void verifyNodeInput(const Node &N, size_t idx) {
       "Any node referencing another node N be in the use-list of the node N");
 }
 
+/// \returns True if \p n is a storage node (variable of placeholder) of the
```
jfix71: nit: variable or placeholder, not of

lib/Graph/Graph.cpp
```diff
@@ -2137,20 +2155,21 @@ void Function::verify() const {
...
       (void)input;
       // Verify each input of N.
       verifyNodeInput(N, idx);
-      assert((std::find(nodes_.begin(), nodes_.end(), *input) != nodes_.end() ||
-              std::find(vars.begin(), vars.end(), input) != vars.end()) &&
-             "Every node referenced by one of the graph"
-             " nodes should be part of the graph");
+      bool foundNode =
+          std::find(nodes_.begin(), nodes_.end(), *input) != nodes_.end();
+      bool foundStorage = isGraphStorageNode(input, this);
```
jfix71: nit: this isn't necessary to run if `foundNode = true`, maybe instead consider `if (!foundNode) foundNode = isGraphStorageNode()`.

lib/ExecutionEngine/ExecutionEngine.cpp
nadavrot: Yes. You are right. I missed that. I am not sure how to handle this right now. I think that the runBatch interface is not great. Maybe we should get rid of it and turn it into a utility that the unit tests use. 

I'll try to think of a way to fix this. Maybe I'll create a tensor view for the lifetime of the function.

nadavrot:
@jfix71 Thank you for the review. I added support for runBatch with the commit:

    Add support for the runBatch interface. This commit implements the parts
    of runBatch that copy parts of the tensors into the placeholder. The
    implementation is zero-copy because we use tensor views.  I hope that
    this is a temporary solution and that we'll get rid of runBatch soon
    (see #1587).

tests/unittests/BackendTest.cpp
```diff
@@ -173,6 +173,55 @@ TEST_P(BackendTest, decoupleCodegenFromGraph) {
   EXPECT_NEAR(HX.at({2}), 9, 1E-5);
 }
 
+TEST(Placeholder, simplePlaceholderValue) {
```
jfix71: nit: doxygen comments

tests/unittests/BackendTest.cpp
```diff
@@ -173,6 +173,55 @@ TEST_P(BackendTest, decoupleCodegenFromGraph) {
...
+
+  EE.compile(CompilationMode::Infer, F);
+
+  EE.run({input}, {&data});
+
+  auto &res = S->getVariable()->getPayload();
+  EXPECT_TRUE(res.isEqual(data));
+}
+
+TEST(Placeholder, runBatchTest) {
```
jfix71: nit: doxygen comments

nadavrot:
@jfix71 Thank you! I'll update the doxygen comments and wait for the tests to pass before landing.

PR #1594, opened 2018-09-07
[ExecutionEngine] Remove the logic that updates variables from the EE.

nadavrot:
*Description*:

This PR refactors the execution engine and removes all of the logic that is responsible for updating variables into external helper functions. It does not make sense for the execution engine, that managed the lifetime of the compiled function to also update variables, and once we migrate to placeholders we won't be able to rely on the execution engine to update the variables/placeholders for us, because they won't be tied to the run() method. 

This change removes the runBatch() method and keeps a simple run() method. All of the variable changes happen before the execution happens. This is the API change for the unit tests:

Before:

    EE.run({var1, var2}, {inputs1, inputs2});	 

After:

    updateVariables({var1, var2}, {inputs1, inputs2});
    EE.run();

*Testing*: I ran "ninja check", the mnist training test and the run.sh inference test.
*Documentation*: Updated doxygen.

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -89,19 +89,6 @@ public:
   void runBatch(size_t iterations, llvm::ArrayRef<Variable *> vars,
                 llvm::ArrayRef<Tensor *> inputs);
 
-private:
-  /// Update the inputs for all variables \p vars with data from the inputs \p
-  /// inputs at offset \p sampleIdx. Then perform a run of the network.
-  void updateInputsAndRunNetwork(llvm::ArrayRef<Variable *> vars,
-                                 llvm::ArrayRef<Tensor *> inputs,
-                                 size_t sampleIdx);
-
-  /// Update the content of the tensor \p v with some slices that from \p input.
-  /// The data starts at slice \p sampleIdx and wraps around until the
-  /// data in \p v is filled. All dimensions, except for the first (batch)
-  /// dimension must be identical.
-  void loadValueFromTensorSlice(Variable *v, Tensor *input, size_t sampleIdx);
-
   // Update the content of the tensor \p v with \p input.
   void loadValueFromTensor(Variable *v, Tensor *input);
 };
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -83,6 +83,14 @@ public:
   void updateVariables(llvm::ArrayRef<Variable *> vars,
                        llvm::ArrayRef<Tensor *> inputs);
 
+  /// Update the content of the tensors \p vars with some slices that from \p
+  /// inputs. The data starts at slice \p sampleIdx and wraps around until the
+  /// data in \p v is filled. All dimensions, except for the first (batch)
+  /// dimension must be identical.
+  void updateVariablesFromBatch(llvm::ArrayRef<Variable *> vars,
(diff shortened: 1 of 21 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -78,34 +78,34 @@ class ExecutionEngine final {
...
-  void loadValueFromTensor(Variable *v, Tensor *input);
+  /// Runs a single execution of the function.
+  void run();
 };
 
+//===----------------------------------------------------------------------===//
+//         Helper methods for running the execution engine.
+//===----------------------------------------------------------------------===//
+
+/// This method updates the variables in \p nodes with the tensor content
```
rdzhabarov: nit: \p vars with the tensor ...

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -55,77 +55,74 @@ void ExecutionEngine::setBackend(Backend *backend) {
...
+    // Extract the n'th slice, that must be a tensor.
+    size_t slc = sampleIdx % dim[0];
+    t.copyConsecutiveSlices(inputs[i], slc);
+  }
+}
+
+void glow::runBatch(ExecutionEngine &EE, size_t iterations,
+                    llvm::ArrayRef<Variable *> vars,
+                    llvm::ArrayRef<Tensor *> inputs) {
+  static size_t trainCounter = 0;
```
rdzhabarov: why `trainCounter` needs to be static? Can you please explain.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -78,34 +78,34 @@ class ExecutionEngine final {
...
+/// dimension must be identical.
+void updateVariablesFromBatch(llvm::ArrayRef<Variable *> vars,
+                              llvm::ArrayRef<Tensor *> inputs,
+                              size_t sampleIdx);
+
+/// Runs \p iterations iterations of the function. The method updates a local
+/// counter and future invocations of this method continue running iterations
+/// of the batch at the next available slice.
+/// The method updates the variables in \p vars with the tensors \p inputs.
+void runBatch(ExecutionEngine &EE, size_t iterations,
```
qcolombet: Is the part with the local counter still true?
Also, while you are here, the comment would benefit from more love :).
After reading it over and over, I am not sure how the iterations are applied to the batch size.

lib/ExecutionEngine/ExecutionEngine.cpp
nadavrot: I added a big comment that explains why this is a really bad idea to have this global variable. I think that I have a quick fix to get rid of the global variable. Basically, now that this is only a helper function we can change the call sites and make it a parameter.

include/glow/ExecutionEngine/ExecutionEngine.h
nadavrot: I updated the comment and got rid of the global variable. 

/// Runs \p iterations iterations of the compiled function. The method updates a
/// global counter and future invocations of this method continue running
/// iterations of the batch at the next available slice.
///
/// The method updates the variables in \p vars with the tensors \p inputs. The
/// shape of the slice has to be identical to the shape of slices in the batch.
/// All dimensions, except for the first (batch) dimension must be identical.
///
/// The variable \p sampleCounter is consumed and updated by the function. This
/// variable records the number of samples that were consumed by the network in
/// previous iterations. The next input to be loaded is
/// (sampleCounter % batchsize).

PR #1597, opened 2018-09-08
[Placeholder] Allow executing Placeholders in the execution engine.

nadavrot:
*Description*:
Implement the logic for executing Placeholders in the execution engine.
This commit changes the backend interface and adds the ability to pass
Placeholder variables in addition to Variables. This change breaks the
out-of-tree backends.

This is the first step in adding support for placeholder variables in
the execution engine. The commit adds support to the interpreter but not
to the OpenCL and CPU backends.

*Testing*: Ran the local tests (ninja check)
*Documentation*: updated doxygen comments.

Related to #1586 and #1334.

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -16,7 +16,15 @@
 #ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
 #define GLOW_BACKENDS_COMPILEDFUNCTION_H
 
+#include <unordered_map>
+
 namespace glow {
+
+class Placeholder;
+class Tensor;
+
+/// Maps placeholders to the tensors that back them.
+using PlaceholderMap = std::unordered_map<Placeholder *, Tensor *>;
 
 /// Interface for executing a compiled function.
 class CompiledFunction {
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -169,8 +169,19 @@ std::unique_ptr<IRFunction> ExecutionEngine::generateIR(CompilationMode mode,
   return IR;
 }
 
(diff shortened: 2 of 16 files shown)


include/glow/Backends/CompiledFunction.h
```diff
@@ -16,8 +16,16 @@
...
 
+#include <unordered_map>
+
 namespace glow {
 
+class Placeholder;
+class Tensor;
+
+/// Maps placeholders to the tensors that back them.
+using PlaceholderMap = std::unordered_map<Placeholder *, Tensor *>;
```
nadavrot: @qcolombet @hegemanjwh2 @artemrakhov @rdzhabarov I used the type "PlaceholderMap" which is an alias to some concrete type. If we ever want to change it to something else (vector, llvm::densemap, etc) it would be really easy to do.

nadavrot:
The second part of this PR is #1601, which adds CPU and OpenCL support for the changes to the execution engine.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -169,8 +169,19 @@ std::unique_ptr<IRFunction> ExecutionEngine::generateIR(CompilationMode mode,
...
-void ExecutionEngine::compile(CompilationMode mode, Function *F) {
-  function_ = backend_->compile(generateIR(mode, F));
+void ExecutionEngine::compile(CompilationMode mode, Function *F,
+                              llvm::ArrayRef<Placeholder *> placeholders,
+                              llvm::ArrayRef<Tensor *> inputs) {
+  PlaceholderMap pmap;
+  assert(placeholders.size() == inputs.size() &&
+         "Invalid number of placeholders");
+
+  for (int i = 0, e = placeholders.size(); i < e; i++) {
```
hegemanjwh2: prefer: std::size_t i = 0, ... but up to you.

PR #1600, opened 2018-09-09
[Placeholder] Implement the execution context.

nadavrot:
*Description*:

The Context provides a mapping between some graph nodes, which are a
symbolic representation of some computation, and concrete tensors that
represent the inputs and outputs to the graph. The context owns the
tensors and the graph uses these values as runtime. This is useful for
the multi-threaded execution of code, where each thread has a different
execution context. The difference between this class and a regular map
is that the Context owns the Tensors (not only the pointers) and manages
their lifetime.

The plan is to pass the context to the different graph builders (for
example, the LSTM builder) and have the builder initialize the weights
inside the context. Later, the context can be used for execution by the
unit tests. We would need to split the graph builders to APIs that
require context (for example, helper builders that create new variables)
and to low-level builders that don't. It should be easy to do.

Once the Context is in place we'll be able to get rid of the last
mutable variables.

*Testing*: Added a unit test.
*Documentation*: Added documentation. 

Related to #1334 and #1597.

### DIFF
--- a/include/glow/ExecutionEngine/Context.h
+++ b/include/glow/ExecutionEngine/Context.h
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_EXECUTIONENGINE_CONTEXT_H
+#define GLOW_EXECUTIONENGINE_CONTEXT_H
+
+#include <unordered_map>
+
+namespace glow {
+
+class Tensor;
+class Placeholder;
+
+/// This class provides a mapping between some graph nodes, which are a symbolic
+/// representation of some computation, and concrete tensors that represent the
+/// inputs and outputs to the graph. The context owns the tensors and the graph
+/// uses these values as runtime. This is useful for the multi-threaded
+/// execution of code, where each thread has a different execution context. The
+/// difference between this class and a regular map is that the Context owns the
+/// Tensors (not only the pointers) and manages their lifetime.
+class Context final {
+public:
+  /// Maps placeholders to the tensors that back them.
+  using PlaceholderMap = std::unordered_map<Placeholder *, Tensor *>;
+
+private:
+  /// Maps Placeholders to Tensors.
+  PlaceholderMap map_;
+
+public:
+  /// \returns the tensor that corresponds to Placeholder \p P or Null if the
+  /// tensor is not found.
+  Tensor *get(Placeholder *P);
+
+  /// Inserts the Placeholder-Tensor pair.
+  void insert(Placeholder *P, Tensor &&T);
+
+  /// \returns True if \p P is a registered Placeholder.
+  size_t count(Placeholder *P);
+
+  /// Deletes all tensors and clears the mapping between Placeholders and
+  /// tensors.
+  void clear();
+
+  /// \returns the mapping between placeholder to tensors.
+  PlaceholderMap &pairs() { return map_; }
+
+  Context() = default;
+
+  ~Context() { clear(); };
+
+  // Don't copy or move this class around.
+  Context(const Context &other) = delete;
+  Context(Context &&other) = delete;
+  Context &operator=(const Context &other) = delete;
+  Context &operator=(Context &&other) = delete;
+};
+
+} // namespace glow
+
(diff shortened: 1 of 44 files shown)


include/glow/ExecutionEngine/Context.h
```diff
@@ -0,0 +1,68 @@
...
+#define GLOW_EXECUTIONENGINE_CONTEXT_H
+
+#include "glow/Base/Tensor.h"
+
+#include <unordered_map>
+
+namespace glow {
+
+class Tensor;
+class Storage;
```
hegemanjwh2: Unused forward declaration...?

include/glow/ExecutionEngine/Context.h
```diff
@@ -0,0 +1,68 @@
...
+/// This class provides a mapping between some graph nodes, which are a symbolic
+/// representation of some computation, and concrete tensors that represent the
+/// inputs and outputs to the graph. The context owns the tensors and the graph
+/// uses these values as runtime. This is useful for the multi-threaded
+/// execution of code, where each thread has a different execution context. The
+/// difference between this class and a regular map is that the Context owns the
+/// Tensors (not only the pointers) and manages their lifetime.
+class Context final {
+  /// Maps Placeholders to Tensors.
+  std::unordered_map<Placeholder *, Tensor *> map_;
```
hegemanjwh2: Should this be a PlaceholderMap?

lib/ExecutionEngine/Context.cpp
```diff
@@ -0,0 +1,45 @@
...
+    return nullptr;
+  }
+
+  return it->second;
+}
+
+void Context::insert(Placeholder *P, Tensor &&T) {
+  assert(!map_.count(P) && "Placeholder already registered");
+  // Take ownership over the tensor.
+  map_[P] = new Tensor(std::move(T));
```
hegemanjwh2: T is already an rvalue reference here...  So std::move() is redundant.

include/glow/ExecutionEngine/Context.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+  /// \returns the tensor that corresponds to Placeholder \p P or Null if the
+  /// tensor is not found.
+  Tensor *get(Placeholder *P);
+
+  /// Inserts the Placeholder-Tensor pair.
+  void insert(Placeholder *P, Tensor &&T);
+
+  /// \returns True if \p P is a registered Placeholder.
+  bool count(Placeholder *P);
```
hegemanjwh2: Having a function called 'count()' return a bool is a little weird.  :-)

lib/ExecutionEngine/Context.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+bool Context::count(Placeholder *P) { return map_.count(P); }
```
hegemanjwh2: This feels a little weird, since std::unordered_map::count() returns a size_type rather than a bool.

include/glow/ExecutionEngine/Context.h
nadavrot: It seems like map::count() returns zero or one. I think that it's okay to return bool because this is a new wrapper API. But we way want to change the name?  What do you think? 


From the documents of unordered map:  
"
Because unordered_map containers do not allow for duplicate keys, this means that the function actually returns 1 if an element with that key exists in the container, and zero otherwise.
"

include/glow/ExecutionEngine/Context.h
hegemanjwh2: That seems very reasonable.  Passing a 'Context' to the various compile()'s seems a little "nicer".

include/glow/ExecutionEngine/Context.h
hegemanjwh2: Acknowledged.  I do not feel strongly about this kind of thing.  Just wanted to point it out.  :-)

lib/ExecutionEngine/Context.cpp
nadavrot: I am not sure why, but for some reason the code requires std::move. As you said, the parameter is properly marked as rvalue and I would expect things to work. However, when I remove the std::move I am getting the error below. I am not sure what's going on here. 

/Glow/lib/ExecutionEngine/Context.cpp:36:17: error: call to deleted constructor of 'glow::Tensor'
  map_[P] = new Tensor((T));
                ^      ~~~
/Glow/include/glow/Base/Tensor.h:152:3: note: 'Tensor' has been explicitly marked deleted here
  Tensor(const Tensor &other) = delete;
  ^
1 error generated.

hegemanjwh2: Oh, huh.  Interesting.  Okay.

hegemanjwh2: Ah, your original code is correct.  I was wrong to comment.  :-)

(Gleaned from the internet...) T has type rvalue reference but is an lvalue.  Therefore std::move() is necessary.

include/glow/ExecutionEngine/Context.h
nadavrot: Okay. I changed this to size_t.

lib/ExecutionEngine/Context.cpp
nadavrot: Ah, this is complicated.
🤒 I will never fully understand this. 🤕 
Thank you for explaining this to me.

PR #1607, opened 2018-09-10
More generic dotty graph generation for quantized types

tlepley-cadence:
Now displays correct information for 16 and 32 bits quantized types.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -205,18 +205,18 @@ struct Type final {
   /// Specifies the element type of the tensor.
   ElemKind elementType_{ElemKind::Int64ITy};
 
-  /// Initialize a new integer type with \p scale and \p offset.
+  /// Initialize a new quantized type with \p scale and \p offset.
   Type(ElemKind elemTy, llvm::ArrayRef<size_t> dims, float scale,
        int32_t offset)
       : scale_(scale), offset_(offset), elementType_(elemTy) {
-    assert(isQuantizedType() && "Only Integer types have a scale and offset");
+    assert(isQuantizedType() && "Only quantized types have a scale and offset");
     initDims(dims);
   }
 
-  /// Initialize a new float type.
+  /// Initialize a new non-quantized type.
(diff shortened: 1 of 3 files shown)


include/glow/Base/Type.h
```diff
@@ -352,6 +352,34 @@ struct Type final {
     GLOW_UNREACHABLE("Invalid type.");
   }
 
+  /// \return the minimum value of quantized elements.
+  int32_t getQMinValue() const {
```
rdzhabarov: should we just make this generic and return min value for any supported type, getMinValue()?
Could you also add a test to verify that min/max value returned is indeed what you expect?

Current implementation aborts when you call getQMinValue() seems potentially dangerous.

tlepley-cadence: The problem is that expressing the limit of different types need a different C type. For example the float limits can't be expressed as an int32_t or int64_t and reverselly, int32_t or int64_t limits can't be expressed accurately with a float. This is why I have decided to limit this functionality to quantized types and to pick int32_t to represent their limit.

tlepley-cadence: I changed a bit the code of the PR. In introduced `Type::getQuantizedValueRange()` instead, that gives the float value range represented by quantized type. It's really specific to quantized type because it's not the actual range of elements (that would be integers). I added an `assert(isQuantizedType())` consistently with `getOffset()` and `getRange()` and the code will similarly not abort anymore in case it's compiled in release mode.

tlepley-cadence: @rdzhabarov I added some basic tests for quantized types.

include/glow/Base/Type.h
```diff
@@ -244,6 +245,34 @@ struct Type final {
     return offset_;
   }
 
+  /// \returns the floating point value range that covers a quantized type (min
+  /// first, max second).
+  std::pair<float, float> getQuantizedValueRange() const {
+    assert(isQuantizedType() &&
+           "Can't get the quantized value range of a float type");
```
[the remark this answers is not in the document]
tlepley-cadence: @rdzhabarov Done. I cleaned a bit comments and error messages for the `Type` class.

PR #1609, opened 2018-09-10
Add a command line option for the image format (RGB/BGR).

tlepley-cadence:
Default stays BGR.

We've got ONNX models that expect a NHWC with a RGB channel ordering, not BGR. In order to avoid an extra image conversion, I preferred to add an 'image format' option to `loadImagesAndPreprocess` rather than adding an additional BGR2RGB conversion in `main`. 

_Side note_: `readPngImage` already reads the image file as a NHWC/RGB tensor. I then think that handling both dimension and channel ordering in `readPngImage` would save many image conversions/copies in case NHWC and/or RGB  is requested by the user.


tools/loader/ImageClassifier.cpp
```diff
@@ -109,9 +123,11 @@ llvm::cl::opt<std::string> modelInputName(
     llvm::cl::cat(imageLoaderCat));
 } // namespace
 
-/// Loads and normalizes all PNGs into a tensor in the NCHW 3x224x224 format.
+/// Loads and normalizes all PNGs into a tensor in the NHWC format with the
+/// requested channel ordering.
 void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
-                             Tensor *result, ImageNormalizationMode normMode) {
+                             Tensor *result, ImageNormalizationMode normMode,
+                             ImageFormat format) {
```
artemrakhov: nit: `imageMode` and `imageFormat` are global for this file, so there's no need to pass them to this function.

tools/loader/ImageClassifier.cpp
```diff
@@ -41,6 +41,11 @@ enum class ImageLayout {
   NHWC,
 };
 
+enum class ImageFormat {
```
artemrakhov: `Format` is pretty general word. Let's switch to `ChannelOrder`

tools/loader/ImageClassifier.cpp
```diff
@@ -89,6 +94,15 @@ llvm::cl::alias imageModeA("i", llvm::cl::desc("Alias for -image_mode"),
                            llvm::cl::aliasopt(imageMode),
                            llvm::cl::cat(imageLoaderCat));
 
+llvm::cl::opt<ImageFormat>
+    imageFormat("image_format",
+                llvm::cl::desc("Specify which image image to use"),
```
rdzhabarov: typo

tools/loader/ImageClassifier.cpp
tlepley-cadence: I think it's not necessary a good practice to use global variables extensively. My preference would be to limit the usage of global variables to the top level (in `main`) and pass them as function parameter. It is consistent with what was already already done.

tools/loader/ImageClassifier.cpp
```diff
@@ -159,7 +181,8 @@ int main(int argc, char **argv) {
 
   // Load and process the image data into the data Tensor.
   Tensor data;
-  loadImagesAndPreprocess(inputImageFilenames, &data, imageMode);
+  loadImagesAndPreprocess(inputImageFilenames, &data, imageNormMode,
```
rdzhabarov: I'd actually remove both` imageNormMode` and `imageChannelOrder` from the calling side. I imagine having those as a separate param would be beneficial for testing `loadImagesAndPreprocess` function, but there is no easy way to test that directly.

tlepley-cadence: @rdzhabarov I'm not sure to understand your request. Do you propose to remove `imageNormMode` and `imageChannelOrder` parameters and to get directly this information from global variables ?

tools/loader/ImageClassifier.cpp
```diff
@@ -140,12 +155,19 @@ void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
...
+            channelOrder == ImageChannelOrder::RGB) &&
+           "Invalid image format");
 
-    // Convert to BGR, as this is what imagenet models are expecting.
+    // Convert to NCHW with the requested channel ordering.
     for (unsigned z = 0; z < numChannels; z++) {
       for (unsigned y = 0; y < dims[1]; y++) {
         for (unsigned x = 0; x < dims[0]; x++) {
-          RH.at({n, numChannels - 1 - z, x, y}) = (imageH.at({x, y, z}));
+          if (channelOrder == ImageChannelOrder::BGR) {
```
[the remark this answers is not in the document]
tlepley-cadence: In my environment, the compiler doesn't fail with an error when a `case` statement is missing in the switch. It just give a warning, what I believe is not strong enough. Then, we would still need an assert for a sanity check like currently or in the switch `default` case if we add one. 
`glow/tools/loader/ImageClassifier.cpp:166:18: warning: enumeration value 'RGB' not handled in switch [-Wswitch]`

tools/loader/ImageClassifier.cpp
artemrakhov: Yes. There's no need to pass them, because they are globally accessible.

tlepley-cadence: OK, I did that change, even if I don't think it's a good practice to pass information *implicitly* to functions through globals. In this case, it's a constant global, so it's kind of OK.

artemrakhov: Well, it's not a regular global variable, it's a flag. AFAIK, there's no other way to define LLVM flag other than global. And if it is defined in the file anyway, why limit ourselves to only using these flags in `main()`?

tools/loader/ImageClassifier.cpp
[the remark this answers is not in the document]
artemrakhov: One can argue that it's better to have compilation warning rather then assert which is only active in debug mode.
We always treat warnings seriously and aim to eliminate them as soon as possible.

However, I don't have strong preference here.

PR #1611, opened 2018-09-10
Allow for backend to accept a Function instead of IRFunction

jfix71:
This PR allows for Backends to take a Function pointer for `compile()` and `save()`, instead of a IRFunction. The EE now does this. This allows for example for a backend to decide whether it wants to use our Instruction level IR or implement its own. The Interpreter, CPUBackend, and OpenCL backend all still use our Instruction level IR.

Note that I needed to keep the `compile()` that takes the IRFunction around, as it is used during unit tests. I couldn't think of good ways to change the unit tests to be able to remove it. Open to suggestions.

Also, the first commit here moves the IROptimizer's `optimize()` to take a bool for `shouldShareBuffers` so that Optimizer no longer depends on Backends. This prevents a cyclic dependence from the second commit.

(I also verified that bundling still works 😃)

### DIFF
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -40,12 +40,6 @@ public:
   /// Dtor.
   virtual ~Backend() = default;
 
-  /// Generate code for input IR function \param IR. \p ctx is the context that
-  /// maps the graph to the concrete execution environment for a specific
-  /// function.
-  virtual std::unique_ptr<CompiledFunction>
-  compile(std::unique_ptr<IRFunction> IR, const Context &ctx) const = 0;
-
   /// Generate code for input function \param F. \p ctx is the context that maps
   /// the graph to the concrete execution environment for a specific function.
   virtual std::unique_ptr<CompiledFunction>
@@ -90,6 +84,17 @@ public:
 
 /// Create a backend of kind \p kind.
 Backend *createBackend(BackendKind backendKind);
+
+// Backends that use Glow low-level IR should inherit from this class. It allows
+// for unit tests to create low-level IR to compile and run.
+class BackendUsingGlowIR : public Backend {
+public:
+  /// Generate code for input IR function \param IR. \p ctx is the context that
+  /// maps the graph to the concrete execution environment for a specific
+  /// function. This is used only for unit testing.
+  virtual std::unique_ptr<CompiledFunction>
+  compileIR(std::unique_ptr<IRFunction> IR, const Context &ctx) const = 0;
+};
 
 } // namespace glow
 
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -40,17 +40,22 @@ public:
   /// Dtor.
   virtual ~Backend() = default;
 
-  /// Generate code for input function \param IR. \p ctx is the context that
+  /// Generate code for input IR function \param IR. \p ctx is the context that
   /// maps the graph to the concrete execution environment for a specific
   /// function.
   virtual std::unique_ptr<CompiledFunction>
   compile(std::unique_ptr<IRFunction> IR, const Context &ctx) const = 0;
 
-  /// Save the bundle for \p IR for a later standalone execution
+  /// Generate code for input function \param F. \p ctx is the context that maps
+  /// the graph to the concrete execution environment for a specific function.
+  virtual std::unique_ptr<CompiledFunction>
+  compile(Function *F, const Context &ctx) const = 0;
+
+  /// Save the bundle for \p F for a later standalone execution
   /// in \p outputDir. Make \p networkName the function name for
   /// the entry point of the network and prepend all generated
   /// files with this name.
-  virtual void save(std::unique_ptr<IRFunction> IR, llvm::StringRef outputDir,
+  virtual void save(Function *F, llvm::StringRef outputDir,
                     llvm::StringRef networkName) const {
     GLOW_UNREACHABLE("Saving a bundle is not supported by the backend");
   }
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -31,10 +31,10 @@
 
 namespace glow {
 
-/// This is the ExecutionEngine. It owns the Graph, the IR, and the backends.
-/// The Graph, IR, etc in this class are defined as pointers, in order to
-/// erase the type and prevent the internal types from leaking out to the
-/// users of this class.
+/// This is the ExecutionEngine. It owns the Graph, the backend, and the
+/// compiled function.  The Graph, etc in this class are defined as pointers, in
+/// order to erase the type and prevent the internal types from leaking out to
+/// the users of this class.
 class ExecutionEngine final {
   /// The Module that represents the high-level program.
   Module M_;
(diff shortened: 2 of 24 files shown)


qcolombet:
> Also, the first commit here moves the IROptimizer's optimize() to take a bool for shouldShareBuffers so that Optimizer no longer depends on Backends. This prevents a cyclic dependence from the second commit.

Sounds reasonable. Going forward, we may what to introduce some kind of CompilerConfig if/when the number of options grow.

include/glow/Backends/Backend.h
```diff
@@ -40,21 +40,33 @@ class Backend {
...
   /// the entry point of the network and prepend all generated
   /// files with this name.
-  virtual void save(std::unique_ptr<IRFunction> IR, llvm::StringRef outputDir,
+  virtual void save(Function *F, llvm::StringRef outputDir,
                     llvm::StringRef networkName) const {
     GLOW_UNREACHABLE("Saving a bundle is not supported by the backend");
   }
 
+  /// Generates, optimizes, and \returns IR for the given Function \p F.
+  virtual std::unique_ptr<IRFunction> generateIR(Function *F) const;
```
qcolombet: Do we have users for this method?
I have mixed feeling about making this public as opposed to protected.

qcolombet:
> Your change does something else. Your change moves the low-level IR into the backend. This does not help targets that don't care about the low-level IR to remove the dependency on the low-level IR.

@nadavrot had a good point, I thought this was changing the ExecutionEngine interface, but after looking again at the patch I realize this changes both the EE and the backend and that's definitely wrong.

jfix71:
>This means that the IR is optional and instantiated inside the different backends. In other words, the CPU, OpenCL and Interpreter backends would need to instantiate their own IR, call the scheduler, IRGen, etc. during the compilation process.

I have pushed a new version that moves `Backend::compile()` to pure virtual. All backends must provide a concrete `compile()`, and can decide if they want to use our low-level IR/depend on the IR lib.  The Backends lib does not depend on the IR lib with this change.

Just to note, the previous version had `Backend::compile()` as virtual, which called into `generateIR()`. This meant that backends that didn't want to use our low-level IR could override it and skip using it as we discussed/as mentioned here. However it did also mean the the Backends lib depended on the IR lib, which shouldn't be necessary if the IR is optional.

As far as not changing the Backend, I don't see how it's possible for this change -- the idea is that we want a new interface between the Backend and the EE, so the Backend naturally will need to change in addition to the EE. Can you explain more what you are thinking here?

nadavrot:
@jfix71 The API proposed in this PR is having two virtual compile methods: compile(IR) and compile(Graph). I think that we need one compile method: compile(Graph). 

The way to implement this would be to move the IR-generation responsibility into the backends. Backends that want to use the IR would link with the IR library and instantiate IRFunction themselves before generating the backend-specific compiled function. In other words, the IR stuff is just a library that some backends can choose to use. This is similar to how the interpreter *can* use the memory allocator, because it's a library, but does not, because it's optional. We will need to change the CPU backend to instantiate the IR library during the compilation phase inside compile(G). What do you think?

jfix71:
>The API proposed in this PR is having two virtual compile methods: compile(IR) and compile(Graph). I think that we need one compile method: compile(Graph).

@nadavrot The only reason I kept `compile(IR)` is because some unit tests need it. I would like to remove it as well. We would need to modify the unit tests to remove it but it was not immediately apparent how to do this, as the tests depend on creating some low-level IR and running it on a backend.

In order to remove it from `Backend` I've added a `BackendUsingGlowIR` abstract class with this pure virtual `compile(IR)` for backends that do use our low-level IR. That way only backends which use our IR need to implement `compile(IR)`. Let me know what you think.

>We will need to change the CPU backend to instantiate the IR library during the compilation phase inside compile(G)

This is how it is currently implemented in the PR. There is a helper `generateAndOptimizeIR()` that is called by `CPUBackend::compile(G)` which creates the IRFunction from the provided Function via IRGen and then optimizes and returns it. It's this way for the Interpreter and OpenCL backend too. The Backends library does not depend on the IR library; the Interpreter, CPUBackend, and OpenCL libs do. So I think what you are describing is what is implemented in this PR, other than the existence of `compile(IR)`.

nadavrot:
@jfix71 I read the update PR again and it looks good. I have a few questions. It sounds like the complexity around BackendUsingGlowIR comes from the unit test that adds the debug-print instruction. Does this feature make sense in the context of the new design? Does it need a new API?  Also, I think that we should remove the dotty-printing of the low-level IR. I don't think that it adds much value. What do you think?

include/glow/Backends/Backend.h
```diff
@@ -86,6 +85,17 @@ class Backend {
 /// Create a backend of kind \p kind.
 Backend *createBackend(BackendKind backendKind);
 
+// Backends that use Glow low-level IR should inherit from this class. It allows
+// for unit tests to create low-level IR to compile and run.
+class BackendUsingGlowIR : public Backend {
+public:
```
qcolombet: I would have expected an override of the compile method for the Function here.

jfix71:
>It sounds like the complexity around BackendUsingGlowIR comes from the unit test that adds the debug-print instruction. Does this feature make sense in the context of the new design?

I think the DebugPrint instruction still useful for debugging backends that do use our IR. We have no equivalent Node for it, so perhaps it would make sense to add such a Node and IRGen it to the Inst like we do for all other instructions. This would solve the problem, but add some complexity in requiring a new DebugPrint Node.

However, the harder issue with the `dataParallelStackingTest` for the CPU backend. It creates a test case from Instructions. Perhaps this test could be reformulated using Nodes instead of Instructions. But it would be a bit fragile because even if we verify that the IRGen'd instructions currently generate the equivalent Instructions, this could change in the future and the test may no longer test what we want.

>Also, I think that we should remove the dotty-printing of the low-level IR. I don't think that it adds much value. What do you think?

Agreed. I think that should be done in a different PR though.

include/glow/Backends/Backend.h
jfix71: As in providing the `compile(Function)` implementation for the Interpreter, OpenCL, and CPU backends? This would require a call into `generateAndOptimizeIR()`. This means Backend would depend indirectly depend on IR (via depending on the Optimizer) which we do not want.

qcolombet: > As in providing the compile(Function) implementation for the Interpreter, OpenCL, and CPU backends?

Yes.

> This means Backend would depend indirectly depend on IR (via depending on the Optimizer) which we do not want.

Oh, but this is fine to have this dependency on the CPU and OpenCL backend, I see. The split seems artificial but okay. What I mean is we could have a backend lib and a backend with glow ir lib, with the dependency going like this:
```
libBackend -> libBackendWithGlowIR (shared code) -> libCPUbackend
                                        \
                                         +> libOCLbackend
```
Instead of
```
libBackend (with backendWithGlowIR) -> libCPUbackend (duplicate code)
         \
          +> libOCLbackend (duplicate code)
```

nadavrot:
@jfix71 Thank you for doing this work.

include/glow/Backends/Backend.h
jfix71: Ah yeah, that could work. I'm not sure it's worth the added complexity though since it's a pretty small amount of shared code, and `libBackendWithGlowIR` would be tiny, at least right now. Do you think it's worth it?

qcolombet: I agree, this is probably not worth the additional complexity, at least for now.
I guess if we end up with a bunch of backend using the IR, then we can rethink our strategy.

PR #1612, opened 2018-09-10
Allow the differentiation of Placeholder nodes.

nadavrot:
*Description*: Allow the differentiation of Placeholder nodes. This PR includes two main parts. First, we add a flag to Placeholder to match the Variable node and allow the differentiation of the selected nodes. The second part fixes a few places (scheduler, cloner, etc) that assume that all storage nodes are Variable. 
 
*Testing*: Added unit tests and ran ninja check.
*Documentation*: None.

This is related to #1334. This PR is a step in the direction of eliminating the Variables, and is required for porting our unit tests to Placeholders instead of training mutable variables.

### DIFF
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -31,8 +31,12 @@ namespace glow {
 // Storage is the base class for Variables, which are bound to tensors, and
 // Placeholder nodes which are unbound.
 class Storage : public Node {
+  /// Specifies if the variable or placeholder is trainable.
+  bool isTrainable_;
+
 public:
-  Storage(Kinded::Kind k, llvm::StringRef name) : Node(k, name) {}
+  Storage(Kinded::Kind k, llvm::StringRef name, bool isTrainable)
+      : Node(k, name), isTrainable_(isTrainable) {}
 
   /// \return the single output value of the node.
   NodeValue getOutput() { return getNthResult(0); }
@@ -50,6 +54,10 @@ public:
   Node *clone() const;
   /// @}
 
+  /// \returns True if the Variable or placeholder are trainable during
+  /// differentiation.
+  bool isTraining() const { return isTrainable_; }
+
   /// \returns result type of the variable.
   TypeRef getType() const { return Node::getType(0); }
 
@@ -66,8 +74,6 @@ public:
 };
 
 class Variable : public Storage {
-  /// Specifies if the variable is trainable.
-  bool isTrainable_;
   /// Specifies the visibility of the variable.
   VisibilityKind visibility_;
   /// The tensor payload that the variable holds.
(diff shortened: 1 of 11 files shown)


include/glow/Graph/Nodes.h
```diff
@@ -31,8 +31,12 @@ namespace glow {
 // Storage is the base class for Variables, which are bound to tensors, and
 // Placeholder nodes which are unbound.
 class Storage : public Node {
+  /// Specifies if the variable is trainable.
```
qcolombet: Instead of "variable" maybe use something that is not tie to a class name, like the "value in that storage".

include/glow/Graph/Nodes.h
```diff
@@ -50,6 +54,9 @@ class Storage : public Node {
   Node *clone() const;
   /// @}
 
+  /// \returns True if the Variable is initialized to be in training mode.
+  bool isTraining() const { return isTrainable_; }
```
qcolombet: Ditto.

lib/IR/GraphScheduler.cpp
```diff
@@ -144,8 +144,8 @@ class ChildMemSizeBasedScheduler : public Scheduler {
     // We don't model memory dependencies, but we still need to honor them.
     // Make sure the SaveNode happens after the last use of the output variable.
     if (auto *save = dyn_cast<SaveNode>(N)) {
-      Variable *output = save->getVariable();
```
qcolombet: Given getVariable may return something that is not a variable now, I would replace it into getStorage instead of going through getOutput.

(I understand this may impact a lot of unittests, feel free to do it in a follow-up PR or file an issue ;))

tests/unittests/BackendTest.cpp
```diff
@@ -207,9 +207,9 @@ TEST(Context, basicContextTest) {
 
   // Create a simple graph, just to have a few placeholders.
   Function *F = mod.createFunction("main");
-  auto *input1 = mod.createPlaceholder(ty, "input1");
-  auto *input2 = mod.createPlaceholder(ty, "input2");
-  auto *input3 = mod.createPlaceholder(ty, "input3");
+  auto *input1 = mod.createPlaceholder(ty, "input1", false);
```
qcolombet: Should we default isTrainable to false?

tests/unittests/graphGradTest.cpp
```diff
@@ -157,3 +157,32 @@ TEST(GraphAutoGrad, cloneAndDiff) {
...
+  // Expect a single user to the trainable input placeholder.
+  EXPECT_EQ(A->getNumUsers(), 1);
+
+  Function *TF = glow::differentiate(F, TC);
+  EE.compile(CompilationMode::Train, TF, ctx);
+  EE.compile(CompilationMode::Infer, F, ctx);
+
+  // Check that the Placeholder has multiple users, because at least one write
+  /// node will be added.
+  EXPECT_GE(A->getNumUsers(), 1);
```
qcolombet: Could we check that the gradient looks reasonable for the placeholder?
Like it is connected to a SGD that write back to it.

include/glow/Graph/Nodes.h
nadavrot: Good catch.

lib/IR/GraphScheduler.cpp
nadavrot: Thank you. Yes. I have this change locally.  I plan to change this to getOutput in a follow up PR.

tests/unittests/BackendTest.cpp
nadavrot: I considered 3 alternatives here. 1) copy the Variable interface which used a default parameter (the default was True). I was unhappy with the default and did not want to switch the default during the migration. 2) I considered using a different API: createTrainablePlaceholder. and 3) I considered using an enum. I decided to be conservative here because I'll need to update hundreds of unit tests soon, so I prefer to keep it simple an clean it up in stages.

tests/unittests/graphGradTest.cpp
nadavrot: Yes. I'll am not working on porting more tests to using placeholders.

lib/IR/GraphScheduler.cpp
nadavrot: Actually, this line changes getVariable to getOutput, which supports both Variable and Placeholder.

nadavrot: Yes. You are right. It should be getDestination().

tests/unittests/BackendTest.cpp
qcolombet: Sounds good.

nadavrot:
@artemrakhov Yes! That's a good point. Thank you.

PR #1614, opened 2018-09-11
[Context] Add a method to allocate a new backing tensor.

nadavrot:
*Description*:

While writing a unit test that replaces some Variable with a Placeholder
I realized that in most cases we mainly care about allocating new
tensors, and not moving an existing tensor into the context. To
implement the new 'allocate' method I had to move the Context into the
Graph library, because it depends on the Placeholder class to figure out
the type of the Placeholder for Tensor allocatio method I had to move
the Context into the Graph library, because it depends on the
Placeholder class to figure out the type of the Placeholder for Tensor
allocation.

*Testing*: Updated the unit test.
*Documentation*: None.

### DIFF
--- a/lib/Graph/Context.cpp
+++ b/lib/Graph/Context.cpp
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-#include "glow/Base/Context.h"
+#include "glow/Graph/Context.h"
 #include "glow/Base/Tensor.h"
+#include "glow/Graph/Nodes.h"
 
 using namespace glow;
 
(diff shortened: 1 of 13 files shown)


lib/Graph/Context.cpp
```diff
@@ -45,6 +46,12 @@ void Context::clear() {
   map_.clear();
 }
 
+Tensor *Context::allocate(Placeholder *P) {
+  Tensor *T = new Tensor(P->getType());
```
rdzhabarov: do we need to make sure that there are no existing tensors backing \p P?

nadavrot: Yes, you are right. I will fix this.

PR #1615, opened 2018-09-11
[GraphScheduler] Fix underflow bug

SplitInfinity:
**Description**
This commit fixes an underflow bug in
`ChildMemSizeBasedScheduler::orderChildNodesAndSchedule` that can
cause incorrect scheduling. There is code in this function that
orders the children of a node by the amount of memory they free up
by executing. For each node, this is computed as `maxSize - resultSize`,
where maxSize is the total memory needed to compute and store the node's
inputs, and resultSize is the total memory needed to store the node's
outputs. For some nodes, such as a `Concat` that takes `Variables` as input,
`resultSize > maxSize`, so this subtraction underflows, and the scheduler
thinks that the node frees up a lot of memory and schedules it before
some or all of its siblings that actually free up more memory.

This commit fixes this bug by setting the amount of memory freed by
the execution of a node to 0 if `maxSize < resultSize` for that node.

**Testing**
This commit includes a unit test that exposes this bug. This unit
tests constructs a graph in which a dummy node has two inputs:
a `Concat` node connected to `Variables`, and a `Slice` node connected to
a `Transpose` node. The Slice node should be scheduled before the
`Concat` node since it frees up memory, but without this patch, it is
scheduled after.

The remaining unit tests all pass.

### DIFF
--- a/include/glow/IR/GraphScheduler.h
+++ b/include/glow/IR/GraphScheduler.h
@@ -1,0 +1,97 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_IR_GRAPH_SCHEDULER_H
+#define GLOW_IR_GRAPH_SCHEDULER_H
+
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Nodes.h"
+#include "glow/Graph/Utils.h"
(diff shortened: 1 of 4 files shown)


opti-mix:
@SplitInfinity Nice catch! Thanks for working on this!

Could you split this PR into two PRs? One that fixes the bug and one that re-factors the code and moves the class definitions into a header file? 

Also, you moved the header file into the "include" sub-directory which is typically used if a header is public and can be used by different parts of the project. But our graph schedulers are used in only one specific place, i.e. they are an implementation detail currently. So, may be the header file should be placed in the same directory where it is used?

bertmaher:
I'm not super familiar with the scheduling algorithm so this might be unwise, but would it make sense to use negative numbers (changing the type to `ssize_t`, probably) when `maxSize < resultSize`?  It seems like it would be useful to know that the operator is actually going to increase the memory footprint.

SplitInfinity:
@opti-mix 

I moved the header to `/include` so that I could use it in the unit test I added (mostly following existing patterns). How can I keep it in `/lib/IR/` but still `#include` it in the unit test?

opti-mix:
@SplitInfinity 

> I moved the header to /include so that I could use it in the unit test I added (mostly following existing patterns). How can I keep it in /lib/IR/ but still #include it in the unit test?

Oh, if you need to use it in the unit test, it is fine to move the header to `/include` in this PR. But I'd still suggest having two commits in this PR. One that does the-refactoring. And one that fixes the actual bug and adds a test.

PR #1622, opened 2018-09-11
[DEMO] This is a preview of MNIST without mutable variables.

nadavrot:
*Do not review this PR!*

This PR is a demo that tests a port of the MNIST demo that does not use any variables, just placeholders. The intention here is to demonstrate the new APIs and to document the progress on the port. I did not touch all other unit tests that don't compile anymore because I modified the graph builder API.


nadavrot:
And the code runs... I have a bug in my port of the test but most of the results are somehow correct.

screen shot 2018-09-11 at 2 17 35 pm

PR #1625, opened 2018-09-12
[GraphScheduler] Fix underflow bug

SplitInfinity:
**Description**
This commit fixes an underflow bug in
`ChildMemSizeBasedScheduler::orderChildNodesAndSchedule` that can
cause incorrect scheduling. There is code in this function that
orders the children of a node by the amount of memory they free up
by executing. For each node, this is computed as `maxSize - resultSize`,
where `maxSize` is the total memory needed to compute and store the node's
`inputs`, and `resultSize` is the total memory needed to store the node's
outputs. For some nodes, such as a `Concat` that takes `Variables` as input,
`resultSize > maxSize`, so this subtraction underflows, and the scheduler
thinks that the node frees up a lot of memory and schedules it before
some or all of its siblings that actually free up more memory.

This commit fixes this bug by modifying the scheduler to use `int64_t`
instead of `size_t` to store `resultSize` and `maxSize` for all nodes.

**Testing**
This commit includes a unit test that exposes this bug. This unit
test constructs a graph in which a dummy node has two inputs:
a `Concat` node connected to `Variables`, and a `Slice` node connected to
a `Transpose` node. The `Slice` node should be scheduled before the
`Concat` node since it frees up memory, but without this patch, it is
scheduled after.

The remaining unit tests all pass.

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -124,6 +124,18 @@ target_link_libraries(graphOptzTest
                         testMain)
 add_glow_test(graphOptzTest ${GLOW_BINARY_DIR}/tests/graphOptzTest)
 
+add_executable(graphSchedulerTest
+	       graphSchedulerTest.cpp)
+target_link_libraries(graphSchedulerTest
+                      PRIVATE
+                        Graph
+                        IR
+                        gtest
+                        testMain)
+target_include_directories(graphSchedulerTest PRIVATE ${CMAKE_SOURCE_DIR}/lib/IR)
+
+add_glow_test(graphSchedulerTest ${GLOW_BINARY_DIR}/tests/graphSchedulerTest)
+
 add_executable(quantizationTest
                quantizationTest.cpp)
 target_link_libraries(quantizationTest
--- a/tests/unittests/graphSchedulerTest.cpp
+++ b/tests/unittests/graphSchedulerTest.cpp
@@ -1,0 +1,87 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "GraphScheduler.h"
+
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Node.h"
+#include "glow/Graph/Nodes.h"
+
+#include "gtest/gtest.h"
+
+using namespace glow;
+
+/// Tests a case in which the memory required to store a node's
+/// output is greater than the memory required to store its input.
+/// This node uses more memory after it executes, and should be
+/// scheduled after its siblings that free up memory after
+/// they execute.
+TEST(GraphScheduler, testMaxSizeLessThanResultSize) {
+  Module MD;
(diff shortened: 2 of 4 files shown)


tests/unittests/graphSchedulerTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (N == concatSmall) {
+      concatSmallIdx = i;
+    }
+
+    if (N == sliceBig) {
+      sliceBigIdx = i;
+    }
+
+    ++i;
+  }
```
qcolombet: Use two std::find instead and compare the iterators.
I reckon this loop is more efficient, but for testing compactness is better :).

tests/unittests/graphSchedulerTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  // For the reason given above, sliceBig should be scheduled
+  // before concatSmallIdx.
+  ASSERT_LT(sliceBigIdx, concatSmallIdx);
```
qcolombet: EXPECT instead of assert is fine here. There is nothing that depends on this being true in the remaining of the program (since this is the end :)).

tests/unittests/graphSchedulerTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+  F->createConcat("concat", {concatSmall, sliceBig}, 0);
+
+  // The graph created above looks like this:
+  //
+  //  bigTensor       smallTensorA       smallTensorB
+  // {100, 4, 4}        {1, 4, 4}         {1, 4, 4}
+  //     |                     \         /
+  //     v                      v       v
+  // transposeBig                 concatSmall
+  //  {0, 2, 1}                     {0}
```
qcolombet: The comments look great. Just one thing that I find confusing:
The syntax between the xxxTensor and the operation is ambiguous.
On one hand, the {} represent the shape of the tensor, on the other hand, they represent the input argument.

Could you put the argument next to the node name and have the output shape below the node name like xxxTensor?
E.g.,
transposeBig
{0, 2, 1}
=>
transposeBig {0, 2, 1}
{100, 4, 4}

Or put a @arg next to the related {} to clarify that these are arguments.
Or whatever you want that is unambiguous :).

tests/unittests/CMakeLists.txt
```diff
@@ -124,6 +124,18 @@ target_link_libraries(graphOptzTest
...
 
+add_executable(graphSchedulerTest
+	       graphSchedulerTest.cpp)
+target_link_libraries(graphSchedulerTest
+                      PRIVATE
+                        Graph
+                        IR
+                        gtest
+                        testMain)
+target_include_directories(graphSchedulerTest PUBLIC ${CMAKE_SOURCE_DIR}/lib/IR)
```
rdzhabarov: not sure if this needs to be `PUBLIC`. As those headers only used in the `graphSchedulerTest.cpp`, `PRIVATE` should be enough.

tests/unittests/graphSchedulerTest.cpp
SplitInfinity: Oh! I didn't know you could do comparisons with the iterators returned by `std::find`.

SplitInfinity: Just an FYI - I had to convert the schedule to a `std::vector` to get iterators I could compare because the schedule is a `std::list` and therefore doesn't have random access iterators.

tests/unittests/graphSchedulerTest.cpp
SplitInfinity: Done.

tests/unittests/CMakeLists.txt
SplitInfinity: Done.

tests/unittests/graphSchedulerTest.cpp
SplitInfinity: Done.

qcolombet:
Looks great, thanks!

tests/unittests/graphSchedulerTest.cpp
qcolombet: I believe you can avoid using a temporary vector by using std::distance.
E.g.,
ASSERT_LT(std::distance(sliceBigIt, end()), std::distance(concatSmallIt, end()))

PR #1627, opened 2018-09-12
[Placeholder] Fix a differentiation bug related to Placeholders.

nadavrot:
*Description*:

Fix a differentiation bug related to Placeholders. The line assumed that
the destination is a variable. 

*Testing*:

This bug was exposed in PR #1622. I did not add a specific test for this because as soon as we migrate more tests to using Placeholders this code will be tested.

*Documentation*: None

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -115,7 +115,7 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
                               cast<SaveNode>(N)->getInput().getType(), 0);
       toAppend.push_back(X);
       map.addGradient(cast<SaveNode>(N)->getInput(), X);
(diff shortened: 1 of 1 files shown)


nadavrot:
@artemrakhov Yes, but we can't get rid of getVariable until we get rid of Variables. :)   
I am taking small steps right now because soon we'll need to rewrite ALL of our unit tests to using Placeholder, instead of Variables, and this will be a huge change, and I am trying to prepare for this big change with small preparation changes.  In this PR: https://github.com/pytorch/glow/pull/1622 I tested the porting of MNIST. Things worked but I ran into a few things that I am now fixing. After migrating everything from Variable to Placeholder we'll be able to delete Storage and just keep Placeholder and Constant.

PR #1633, opened 2018-09-12
[WIP][GraphOptimizer] Sink Transpose below Slice

SplitInfinity:
**Description**
This commit modifies the `sinkCode` function in `GraphOptimizer.cpp`
to sink `TransposeNodes` below `SliceNodes`. In the best case, when
all of the children of a particular `TransposeNode` are `SliceNodes`,
this optimisation will help reduce the memory footprint by taking
slices directly out of the input to the original `TransposeNode`
and then transposing those slices (presumably) in-place. In the
worst case, when only one of the children of a `TransposeNode` is
a `SliceNode` and the rest are nodes that the optimizer does not
sink `TransposeNodes` into, this optimisation will keep the memory
footprint the same because the `TransposeNode` that it inserts
can presumably still be executed in-place.

**Testing**
This commit includes a unit test (`sinkTransposeBelowSlice`)
in `graphOptzTest.cpp` that tests this optimisation. All other unit
tests still pass.

**Example**
Below is an example of a graph that motivated this optimisation. It's hard to see, but the blue node is a `TransposeNode` whose children are all `SliceNodes` that feed into `ReshapeNodes` that in turn feed into `MatMulNodes`.

before

The memory footprint of this example can be reduced by sinking that `TransposeNode` into the `SliceNodes` and rewiring the `SliceNodes` to take slices out of the input to the TransposeNode, which is exactly what this optimisation does:

after

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -219,6 +219,43 @@ static bool sinkCode(Function *F) {
   // For each node:
   for (auto &N : nodes) {
     auto *node = &N;
+    // Sink Tranpose below slice nodes:
+    if (auto *S = dyn_cast<SliceNode>(node)) {
+      auto *TR = dyn_cast<TransposeNode>(S->getInput());
+
+      if (!TR) {
+        continue;
+      }
+
+      // Modify the indices of the SliceNode to make sure
+      // it takes the same slice from the input of the TransposeNode.
+      auto transposeShuffle = TR->getShuffle();
+      auto sliceStart = S->getStart();
+      auto sliceDims = S->getResult().dims();
+      auto numDims = transposeShuffle.size();
+
+      // "Undo" the transpose to figure out the correct start and
+      // output dimensions for the new SliceNode.
+      llvm::SmallVector<size_t, 8> newSliceStart(numDims);
+      llvm::SmallVector<size_t, 8> newSliceDims(numDims);
+
+      for (size_t i = 0; i < numDims; ++i) {
+        newSliceStart[transposeShuffle[i]] = sliceStart[i];
+        newSliceDims[transposeShuffle[i]] = sliceDims[i];
+      }
+
+      // Create a new SliceNode and TransposeNode and insert them in the
+      // appropriate place in the graph.
+      auto NSOutTy = F->getParent()->uniqueTypeWithNewShape(
+          TR->getResult().getType(), newSliceDims);
+      auto *NS =
+          F->createSlice(S->getName(), TR->getInput(), newSliceStart, NSOutTy);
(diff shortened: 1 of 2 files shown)


SplitInfinity:
Note: This is more of a work-in-progress/request-for-comments PR than one that should be seriously considered for merge.

qcolombet:
Nice findings!
Given this is going to scatter transposes around, I believe we want to do this optimization as late as possible. Also, it is possible that transposes smaller than a given size are going to be inefficient to execute on specific targets.
All in all, I think we should run this as a per target basis maybe as part of the pre lowering transformation.

cc: @jfix71, @opti-mix, @stoklund, @nadavrot, @bertmaher, @artemrakhov, @beicy, @hegemanjwh2.

nadavrot:
@qcolombet I agree that we need to start thinking about the design of our optimization pipeline. I want to throw in another optimization that we need to decide where/when to enable -- the optimization that merges small matrix multiplications to larger ones (that's enabled by default but does not speed things up AFAIK). Both of these optimizations might be profitable or not, depending on the context. 

Alright let's design the optimization pipeline.   :) :) :)

jfix71:
Cool! It's great to see removed Nodes from optimizations like these. I'm curious though -- have you looked at the IR before/after this optimization? I'm unsure if this would actually improve memory consumption, at least in the example you provided.

In the before case we would need an alloc of size `5x7x128` to write the transpose result (`RHS_transpose`) because of the other users (e.g. `matmul_result_sliceA_0`) of the reshape (`concat_result1`).

In the after case I believe all of the new sunken transposes (e.g. `RHS_transpose1`) would also still need an alloc to copy their result into (5 of them, each of size `1x7x128`), because they all have another user (e.g. `matmul_result_sliceA_0`) in addition to the sunken transpose.

In both cases, we could optimize this by making the transpose instruction scheduled last, after all of the other users, and then potentially do it in place. But that means we would save the memory in both cases.

opti-mix:
I think that overall this optimization makes sense.

But I also agree with the concerns raised by @jfix71 and @qcolombet regarding the benefits of applying this optimization. It seems like it is only beneficial in certain cases. But I'd say that the same concerns apply to many other optimizations that we have. It seems like we need a general way of deciding if it is profitable to apply any specific optimization.

artemrakhov:
Which network benefits from this optimization? Where is example graph coming from?

SplitInfinity:
@artemrakhov

Please do read my previous comment as well. After reading @jfix's comment, I realized the optimisation does not actually work on the example I provided, but would work on a graph with the same structure but different arguments for the `TransposeNode` and `SliceNodes`.

rdzhabarov:
@SplitInfinity what's the status here? good to close? any further follow-ups?

rdzhabarov:
This seems to be WIP for more than a month. I'm closing it now, @SplitInfinity feel free to reopen if you have followups.

PR #1636, opened 2018-09-13
Preserve the predicate across the sinkCode optimization

qcolombet:
*Description*:
This patch series makes sure the predicates are properly preserved through the sinkCode optimization.
The PR is quite lengthy, but the relevant changes (i.e., not test cases) are pretty small (maybe ~20 lines in GraphOptimizer.cpp).

The bulk of the PR is an hardening of the sinkCode tests + new tests for predication.

*Testing*:
ninja check
At this point, predicates don't change the behavior of the program anyway
*Documentation*: N/A

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -401,9 +401,11 @@ static bool sinkCode(Function *F) {
 #undef BOOLEAN_OP_CASE
 #undef ARITHMETIC_CASE
 
+      newAN->setPredicate(node->getPredicate());
       changed = true;
       auto *newTR =
           F->createTranspose(LTR->getName(), newAN, LTR->getShuffle());
+      newTR->setPredicate(node->getPredicate());
 #define GET_RESULT(NODE_) NODE_->getNthResult(0)
       GET_RESULT(node).replaceAllUsesOfWith(newTR);
 #undef GET_RESULT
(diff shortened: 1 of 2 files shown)


qcolombet:
> Notice that something is up with "thirdparty/onnx". You may need to rebase your branch.

Good catch!

> Do you think that we need to add so many unit tests?

I thought about that and unfortunately I didn't see a way around it. For instance, I discovered that sinking transpose through two different nodes was implemented by two different C++ paths, so unlike what I thought initially, I could not just test a few patterns and had to test all of them. Of course, the changes made to the different patterns are similar enough that one could think we don't need to test them as well. Although that's true, I don't trust myself for not making typos, so I test every single change :).

> How are we going to update all of these unit tests? What do you think is the right tradeoff?

I don't think we should compromise on adding tests because we may need to update them in the future. But again, it's because I don't trust myself. I reckon this is painful to update and that people may have different incentives on those matters!

Long term, I believe that we should make test cases simpler to write. One way to do that, is by getting away of pure C++ written tests. More specifically, I want to go toward writing IR in a text format and get the optimized IR in a text format, a la LLVM opt. The idea is checks become really easy to write, because the textual representation embeds all the constrains (e.g., dimensions, types, use-def chain), whereas in C++ you have to check them explicitly and independently.
The other advantage of text based tests is that it is easier to script the updates, since we don't have to deal with the syntactic sugar required by C++. We could even write some auto-upgrader program if scripting's too hard.

I've been working with bootcampers on bringing that textual IR up and will keep pushing and making the testing smoother.

nadavrot:
+1

PR #1639, opened 2018-09-13
Add TileNode, lowered to InsertTensorNode

jfix71:
*Description*: Backends may prefer to be passed a TileNode instead of implementing it as an InsertTensorNode/Instruction. Add the TileNode explicitly so that the backend can prevent lowering. I also added the creator for InsertTensorNode since it did not already exist.

*Testing*: Current tests for Tile should cover this already. I also added a test for InsertTensorNode which was not previously explicitly tested.

*Documentation*: N/A

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -820,8 +820,8 @@ ConcatNode *Function::createConcat(llvm::StringRef name,
   return addNode(new ConcatNode(name, OT, ops, dimension));
 }
 
-InsertTensorNode *Function::createTile(llvm::StringRef name, NodeValue input,
-                                       unsigned_t tiles, unsigned_t axis) {
+TileNode *Function::createTile(llvm::StringRef name, NodeValue input,
+                               unsigned_t tiles, unsigned_t axis) {
   assert(tiles > 0 && "Tiles must be non-zero.");
   assert(axis >= 0 && axis < input.dims().size() &&
          "Axis must fall in range of source dims.");
@@ -829,11 +829,15 @@ InsertTensorNode *Function::createTile(llvm::StringRef name, NodeValue input,
   ShapeVector outShape(input.dims().begin(), input.dims().end());
   outShape[axis] *= tiles;
   auto OT = getParent()->uniqueTypeWithNewShape(input.getType(), outShape);
+  return addNode(new TileNode(name, OT, input, tiles, axis));
+}
 
-  // Use a zero splat as the Big node input for InsertTensor.
-  auto *zero = createSplat(name, OT, 0);
-  auto start = std::vector<size_t>(input.dims().size(), 0);
-  return addNode(new InsertTensorNode(name, zero, input, start, tiles, axis));
+InsertTensorNode *Function::createInsertTensor(llvm::StringRef name,
+                                               NodeValue big, NodeValue small,
+                                               llvm::ArrayRef<size_t> start,
+                                               unsigned_t count,
+                                               unsigned_t axis) {
+  return addNode(new InsertTensorNode(name, big, small, start, count, axis));
 }
 
 SliceNode *Function::createSlice(llvm::StringRef name, NodeValue input,
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -423,6 +423,8 @@ void InsertTensorNode::verify() const {
   auto src = getSmall();
   auto offsets = getStart();
   unsigned numDims = dest.dims().size();
+  unsigned axis = getAxis();
+  unsigned count = getCount();
   (void)numDims;
   (void)dest;
   (void)src;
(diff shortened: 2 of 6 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -3314,6 +3314,46 @@ TEST_P(InterpAndCPU, SigmoidCrossEntropyWithLogits) {
...
 }
 
+/// Test the InsertTensor node works correctly.
+TEST_P(InterpAndCPU, insertTensorTest) {
+  auto SN0Ty = mod_.uniqueType(ElemKind::Int64ITy, {4, 6});
+  auto SN1Ty = mod_.uniqueType(ElemKind::Int64ITy, {2, 2});
+
+  // 1 1
+  // 1 1
+  Node *SN0 = F_->createSplat("zero", SN0Ty, 0.);
```
opti-mix: The values of SN0 are 0., but the comment states that it is 1.

tests/unittests/OperatorTest.cpp
```diff
@@ -3314,6 +3314,46 @@ TEST_P(InterpAndCPU, SigmoidCrossEntropyWithLogits) {
...
+
+  // 0 0 0 0 0 0
+  // 0 0 0 0 0 0
+  // 0 0 0 0 0 0
+  // 0 0 0 0 0 0
+  Node *SN1 = F_->createSplat("one", SN1Ty, 1.);
```
opti-mix: The values of SN1 are 1., but the comment states that it is 0.

tests/unittests/OperatorTest.cpp
```diff
@@ -3314,6 +3314,46 @@ TEST_P(InterpAndCPU, SigmoidCrossEntropyWithLogits) {
   EXPECT_TRUE(expected.isEqual(result->getPayload()));
 }
 
+/// Test the InsertTensor node works correctly.
+TEST_P(InterpAndCPU, insertTensorTest) {
```
opti-mix: Question: The PR introduces the new Tile node. Shouldn't we have a test checking that this node is lowered into insert_tensor nodes?

opti-mix: Oh, I see in the PR description that we re-use existing tile tests for it.

jfix71: Yeah before this PR there was already a `Graph::createTile()` which has unit tests for it. Before this PR it returned an `InsertTensorNode`, which I now moved to return a `TileNode` which is later lowered to an `InsertNode`.

lib/Graph/Nodes.cpp
```diff
@@ -452,6 +452,23 @@ void SliceNode::verify() const {
...
 }
 
+void TileNode::verify() const {
+  auto dest = getResult();
+  auto src = getInput();
+  unsigned axis = getAxis();
+  unsigned count = getCount();
+  (void)dest;
+  (void)src;
+  (void)axis;
```
beicy: dest, src and axis are used later. no no need to add (void)* here.

tools/ClassGen/NodeGen.cpp
```diff
@@ -395,6 +395,13 @@ int main(int argc, char **argv) {
                     "Data {{1,2},{3,4},{5,6}}, Slices {{-3,-4}}, and Indices "
                     "{1}, the result is {{1,2},{-3,-4},{5,6}}.");
 
+  BB.newNode("Tile")
+      .addInput("Input")
+      .addMember(MemberType::Unsigned, "Count")
+      .addMember(MemberType::Unsigned, "Axis")
+      .addResultFromCtorArg()
+      .setDocstring("Tile a an Input tensor Count times along Axis.");
```
artemrakhov: typo: `Tile an Input`

lib/Graph/Graph.cpp
```diff
@@ -820,20 +820,29 @@ ConcatNode *Function::createConcat(llvm::StringRef name,
...
   auto OT = getParent()->uniqueTypeWithNewShape(input.getType(), outShape);
+  return addNode(new TileNode(name, OT, input, tiles, axis));
+}
+
+InsertTensorNode *Function::createInsertTensor(llvm::StringRef name,
+                                               NodeValue big, NodeValue small,
+                                               llvm::ArrayRef<size_t> start,
+                                               unsigned_t count,
+                                               unsigned_t axis) {
+  assert(big.dims().size() == small.dims().size() &&
```
artemrakhov: I realized that we check this in `InsertTensorNode::verify()`. But we don't assert anything related to `count` and `axis` there.

Maybe makes sense to remove checks here and improve checks there?

lib/Graph/Nodes.cpp
jfix71: Fixed

tools/ClassGen/NodeGen.cpp
jfix71: Fixed

lib/Graph/Graph.cpp
jfix71: Yeah it seems this partially duplicates the checks from `verify()`. I removed them from here and added `axis`/`count` checks to InsertNode.

nadavrot:
@jfix71 We also need to implement the differentiation support for this new operator.

jfix71:
@nadavrot OK if I add it in a future PR? We didn't/don't have have differentiation support for InsertTensor either, so this PR doesn't degrade support at all.

nadavrot:
Okay. 

PR #1644, opened 2018-09-13
Refactor our optimization passes to be more modular/OO

xsunfeng:
*Description*: refactor the implementation of the optimizers so that all of them share a high level class Optimization/IROptimization.  The class features a run method (Optimization::run) running on a Module.
*Testing*: ninja check, 100% tests passed, 0 tests failed out of 20.

### DIFF
--- a/tests/unittests/MLTest.cpp
+++ b/tests/unittests/MLTest.cpp
@@ -58,7 +58,9 @@ TEST_P(MLTest, learnSqrt2Placeholder) {
 
   Node *M = F->createMul("Mult", A, A);
   M = F->createRegression("reg", M, E);
-  F->createSave("ret", M);
+  SaveNode *SN = F->createSave(ctx, "ret", M);
+
+  ctx.allocate(SN->getPlaceholder());
 
   Function *TF = glow::differentiate(F, TC);
   EE_.compile(CompilationMode::Train, TF, ctx);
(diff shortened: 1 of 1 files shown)


lib/Optimizer/IROptimizer.cpp
```diff
@@ -1292,204 +1294,214 @@ static bool allButFirstDimsEqual(llvm::ArrayRef<size_t> sourceDims,
...
-        name += instrName;
-        name += ".";
-        name += I->getKindName();
-        auto *dumpInstr = new DebugPrintInst(name, Op.first);
-        M.insertInstruction(&*next, dumpInstr);
-      }
-    }
-    it = next;
-  }
+class OptimizeInserts : public IROptimization {
```
artemrakhov: Should this become `OptimizeExtracts`?

nadavrot:
Hi @xsunfeng. What's the motivation for this change? Are you interested in preserving some state across the different optimization passes? Or is this change mainly about adding a namespace?

xsunfeng:
Hi @nadavrot 

Context:
Conceptually the optimizations have the same structure (though they do not share any implementation or high level concept): they
- Scan the function/module
- Transform the IR

Problems (Motivation):
When we perform certain actions on all optimization, now we have to do it one by one, examples are:
- debug each optimization one by one, then we have to set a breakpoint on all individual function
- make sure IR is valid after each optimization

nadavrot:
@xsunfeng Thank you for working on this. I am not sure why wrapping our optimization functions in classes would help the problems that you mentioned. I envision having a pass manager one day, but considering the fact that we have a small number of optimizations I am not sure what problem a pass manager would solve. Placing a breakpoint or commenting out an optimization is relatively easy to do when debugging a program.

qcolombet:
Hi @nadavrot,
I'm the one who asked @xsunfeng to look into that. (Haven't looked in the patch yet, though).

The motivation was that I wanted to be able to add pre-post hooks easily, like printing the IR after each optimization, calling the verifier and so on.
I.e., having a little bit more structure here would be beneficial.

qcolombet:
@xsunfeng, let us focus on one optimization first and lay down the ground work before changing everything.

Could you do only one Graph optimization for now.

nadavrot:
@xsunfeng @qcolombet Ah got it. I know that LLVM and GCC have similar utilities. The Swift compiler can even load YAML files that describe the optimization pipeline. I am wondering, is there a need to implement this in our compiler right now? Is this just something that came up by someone who worked on debugging some optimization because of a need to be able to debug the compiler or is this just a general idea of how to improve the current infrastructure?

artemrakhov:
@nadavrot : I was once debugging a net, which had a lot of nodes and took long time to generate code. I had to insert timestamp dumps between each two optimization.

qcolombet:
@nadavrot similar experience as @artemrakhov whereas I wanted to add some check in the verifier and see what optimization was breaking it.

lib/Optimizer/Optimization.h
```diff
@@ -0,0 +1,8 @@
+using namespace glow;
+
+class Optimization {
+public:
+  Optimization() {}
+  virtual ~Optimization(){};
+  virtual void run(Function *F) {}
```
qcolombet: Could you run a bool to indicate whether or not the function has been changed?

nadavrot:
Okay. Got it. Thank you for explaining the motivation. Please make sure to use the right terminology. Optimization -> Pass. 

@qcolombet @artemrakhov @xsunfeng

lib/Optimizer/Optimization.h
```diff
@@ -0,0 +1,8 @@
+using namespace glow;
+
+class Optimization {
+public:
+  Optimization() {}
+  virtual ~Optimization(){};
+  virtual bool run(Function *F) {}
```
opti-mix: I'd suggest adding a `getName()` method. It can be very useful for debug printing of pass information and for selecting via the command-line only specific passes to run.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -24,6 +24,8 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 
+#include "Optimization.h"
```
rdzhabarov: is this missing from the PR?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -24,6 +24,9 @@
...
+
+#include <iostream>
```
rdzhabarov: kill this

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -406,8 +427,10 @@ static bool sinkCode(Function *F) {
       if (L && R) {
         auto *newCN = F->createConcat(
             CN->getName(), {L->getInput(), R->getInput()}, CN->getDim());
+        newCN->setPredicate(node->getPredicate());
```
rdzhabarov: why all these changes?

nadavrot:
Please rebase and squash the PR.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -142,13 +145,13 @@ static bool isIdentityShuffle(llvm::ArrayRef<unsigned_t> shuffle1,
   return true;
 }
 
-/// \returns True if the node \p N always evaluates to zero.
-bool isZero(Node *N) {
+/// \returns True if the node \p N always evaluates to \p val.
+bool isSplatOfVal(Node *N, float val) {
```
rdzhabarov: this should be `static` if it's used in the scope of this CPP only.

lib/Optimizer/Optimization.h
```diff
@@ -0,0 +1,9 @@
+using namespace glow;
```
rdzhabarov: I do not think it's a good idea for ```using namespace glow;``` for anything which includes this file.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1110,23 +1130,42 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
-  for (auto &node : nodes) {
-    if (auto *CN = dyn_cast<ConcatNode>(&node)) {
-      NodeValue newCN = simplifyConcatNode(F, CN);
-      if (newCN.getNode()) {
-        CN->getResult().replaceAllUsesOfWith(newCN);
+class OptimizeConcatNodes : public Optimization {
+public:
+  OptimizeConcatNodes() : Optimization() {}
+  ~OptimizeConcatNodes(){};
+  bool run(Function *F) {
```
rdzhabarov: please, add `override` at the end for all of the relevant methods

xsunfeng:
@nadavrot I am trying to do rebase and squash, but commits by others are added as well. Sorry but I am not familiar with how this works.

rdzhabarov:
On your master make sure that you `git pull` the latest changes from Glow master.
On your working branch run: `git rebase -i master` where master is your local master.
Rebase changes, if needed use `f` to squash commits into a single one.

After rebase is completed, `git push -f origin HEAD`. That's the typical flow I use.

rdzhabarov:
Do not make changes on your local master as it's now: `xsunfeng:master`.
Create a separate feature branch and work on that instead.

xsunfeng:
@rdzhabarov Thanks for the instructions!

If I understand you correctly, xsunfeng/master (my copy) is used for keeping track of the latest changes on glow/master, while xsunfeng/feature is used as a working directory.

I reseted my master copy, created a local feature branch, and am working on the feature branch.  When I submit a pull request, a separate case is open: https://github.com/pytorch/glow/pull/1659.  Is there any way to merge that PR to this one?

rdzhabarov:
I think separate PR is fine (cannot unite PRs when different branches are used).

PR #1645, opened 2018-09-13
[Placeholder] Remove the use of variables from a few tests.

nadavrot:
*Description*: Remove the use of variables from a few tests.
*Testing*: This commit updates some tests. 
*Documentation*: None.


nadavrot:
After this patch we are down to 596 calls to createVariable().

PR #1647, opened 2018-09-14
[Placeholder] Migrate more tests to the new API.

nadavrot:
*Description*: Migrate more tests to the new API. Get rid of additional usages of getVariable().
*Testing*: This commit updates tests.
*Documentation*: None.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -587,8 +587,8 @@ AvgPoolNode *Function::createAvgPool(llvm::StringRef name, NodeValue input,
 }
 
 FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
-                                                   NodeValue input, Variable *W,
-                                                   Variable *B) {
+                                                   NodeValue input, Storage *W,
+                                                   Storage *B) {
   TypeRef T = input.getType();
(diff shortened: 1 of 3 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1998,6 +1998,379 @@ SaveNode *Function::createSave(Context &ctx, llvm::StringRef name,
   return addNode(new SaveNode(name, input, dest));
 }
 
+void Function::createGRU(Context &ctx, llvm::StringRef namePrefix,
```
bertmaher: Will we eventually remove the non-Context version of these?

nadavrot: Thanks for the review. Yes! We'll delete the old builders. We need to migrate all of our tests (~600 call sites to createVariable()) from variables (which are not thread safe) to Placeholders. It will be difficult to migrate everything in one big commit, so I am porting the tests one by one, duplicating the APIs for creating the graphs. The plan (#1334) is to delete the old variable APIs as soon as they become unused. Next, we'll refactor and clean the builder APIs. To east the migration I try to keep the builder APIs as close as possible to the old ones, but as soon as we finish the migration it would be a good idea to clean things up. @bertmaher

PR #1653, opened 2018-09-17
[Quantization] Support for Slice, Gather, Tile

jfix71:
Allow for a Tile to be quantized based on the profile that was gathered for it. I added Tile to the end2end quantization test.

Slice and Gather were already supported, but not yet added to the CPU Backend.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -828,15 +828,19 @@ ConcatNode *Function::createConcat(llvm::StringRef name,
 }
 
 TileNode *Function::createTile(llvm::StringRef name, NodeValue input,
-                               unsigned_t tiles, unsigned_t axis) {
+                               unsigned_t tiles, unsigned_t axis,
+                               TypeRef outTy) {
   assert(tiles > 0 && "Tiles must be non-zero.");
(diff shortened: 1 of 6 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -360,6 +360,19 @@ static Node *quantizeNode(Function *F, Node *node,
                                     quantizedInputs[1]);
     break;
   }
+  case Kinded::Kind::TileNodeKind: {
```
rdzhabarov: could you add Tile node to quantization e2e to verify that the conversion works properly?

lib/Graph/Graph.cpp
```diff
@@ -820,6 +820,12 @@ ConcatNode *Function::createConcat(llvm::StringRef name,
   return addNode(new ConcatNode(name, OT, ops, dimension));
 }
 
+TileNode *Function::createTile(llvm::StringRef name, NodeValue input,
```
rdzhabarov: similar `asserts` as in other createTile function, or better yet merge these two function in one with optional outTy param to reduce code dups.

jfix71:
All comments addressed.

PR #1654, opened 2018-09-17
Add support for quantizing Log

jfix71:
*Description*: Quantized `SigmoidNode` and `TanhNode` are implemented with a pre-calculated mapping that is input to a `IntLookupTable`. However implementing a quantized `LogNode` with a `IntLookupTable` requires calculating a new mapping depending on the input and output min/max. I added a helper `createMapping()` for creating such a mapping to Quantization/Base in the first commit. The second commit adds support for lowering a quantized `LogNode` into a `IntLookupTable` that has its mapping created via `createMapping()`.

*Testing*: Added a unit test for quantized log


jfix71:
Not positive if this is the direction we want to go here, since we would have a different mapping for each quantized `LogNode` with different quantization parameters. However, the mapping data is only 256 bytes, so it's not so bad.

If it's agreed that this is good direction, then I'm wondering if perhaps we should be using this for Sigmoid/Tanh instead of clipping their input ranges, for those nodes that have input ranges that are small/limited inside the input range we currently clip into. E.g. if we see a Sigmoid with input range `[-1.0, 1.0]` we are losing some precision since we always set the input range to `[-3.0, 3.0]`.

rdzhabarov:
> E.g. if we see a Sigmoid with input range [-1.0, 1.0] we are losing some precision since we always set the input range to [-6.0, 6.0]

I think the good thing is that if we clip to the certain range we could:
* reuse the same table at a later point if needed. From the practical standpoint, the difference in 0.5% seems to be negligible (that's what you'd get as a % difference).
* if it happens that there are outliers it's better to keep [min,max] within smaller range if possible (e.g. no significant accuracy drop).

With the Log, i'm not sure if we could clip similary.

include/glow/Quantization/Base/Base.h
```diff
@@ -99,6 +101,10 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
 TensorQuantizationParams chooseQuantizationParams(float min, float max,
                                                   Schema schema = Asymmetric);
 
+/// \return a mapping from the \p inTy to the \p outTy given the function \p f.
```
rdzhabarov: based on the return result this seems to imply that it works for quantized int8 only. can you reflect that in the comment?

lib/Graph/Graph.cpp
```diff
@@ -1120,6 +1120,11 @@ LogNode *Function::createLog(llvm::StringRef name, NodeValue input) {
   return addNode(new LogNode(name, input.getType(), input));
 }
 
+LogNode *Function::createLog(llvm::StringRef name, NodeValue input,
```
rdzhabarov: could you merge this and `createLog` above with optional TypeRef outTy arg? That makes Graph interface a bit leaner and does not bring a lot of complexity into createLog impl.

lib/Optimizer/Lower.cpp
```diff
@@ -187,6 +187,25 @@ void lowerReluNode(Function *F, ReluNode &R) {
   R.getResult().replaceAllUsesOfWith(relu);
 }
 
+/// Lower a quantized Log by creating an IntLookupTable with a new mapping given
+/// the input and output quantization parameters.
+void lowerQuantizedLogNode(Function *F, LogNode *LN) {
```
rdzhabarov: nit: static func and const LogNode *

tests/unittests/OperatorTest.cpp
```diff
@@ -2684,6 +2684,50 @@ TEST_P(InterpAndCPU, Int8Tanh) {
...
+TEST_P(InterpAndCPU, Int8Log) {
+  constexpr size_t size = 1000;
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {size}, "input");
+
+  const float min = 1.0;
+  const float max = 100.0;
+  input->getHandle().randomize(min, max, mod_.getPRNG());
+
+  // Input some random data into an fp log.
+  auto *fpTanh = F_->createLog("fpLog", input);
```
rdzhabarov: nit: var name

tests/unittests/OperatorTest.cpp
```diff
@@ -2684,6 +2684,50 @@ TEST_P(InterpAndCPU, Int8Tanh) {
   }
 }
 
+/// Verify that a quantized Log works correctly.
+TEST_P(InterpAndCPU, Int8Log) {
+  constexpr size_t size = 1000;
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {size}, "input");
```
rdzhabarov: maybe make this `var` public to avoid any potential optimizations on it.

lib/Optimizer/Lower.cpp
jfix71: I was matching the rest of them in here but I agree it should be the case. I will do a cleanup followup PR here to fix them all at once.

jfix71:
Addressed most issues. I also fixed a travis asan error -- the quantization range always must include 0.0, but `log(0.0)` is undefined of course, so I added a check if the input is 0.0 to use the minimum positive float value for `log()`.

include/glow/Quantization/Base/Base.h
```diff
@@ -99,6 +101,11 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
 TensorQuantizationParams chooseQuantizationParams(float min, float max,
                                                   Schema schema = Asymmetric);
 
+/// \return an int8 vector mapping from the \p inTy to the \p outTy given the
```
rdzhabarov: super nit: \returns

Maybe specify in the \pre that inTy and outTy should be int8qty types.

tests/unittests/OperatorTest.cpp
```diff
@@ -2684,6 +2684,51 @@ TEST_P(InterpAndCPU, Int8Tanh) {
   }
 }
 
+/// Verify that a quantized Log works correctly.
+TEST_P(InterpAndCPU, Int8Log) {
```
nadavrot: Would it be possible to add the log as a node to an existing network? This could minimize the number of tests that we need to update in the current and next refactoring.

jfix71: I think it's important to compare the float vs. int8 results to make sure they're relatively close. The only somewhat alternative to this (other than basically replicating what's in this test in another) would be to add it to the quantizationTest's end2end test. However I feel that test is more about making sure quantization via profiling works, not numerical comparison of results. Did you have something else in mind or a different opinion on this?

include/glow/Quantization/Base/Base.h
jfix71: Done

tests/unittests/OperatorTest.cpp
jfix71: @nadavrot Let me know if you would like me to change this test.

nadavrot: Sorry for the delay in response. Makes sense. Everything LGTM. Thank you for doing this work.

PR #1655, opened 2018-09-17
[Nodes] Do not verify offset of quantized ReluNode

jfix71:
This check only makes sense if we are using `quantization::Schema::Asymmetric`. Currently the schema is only used when dumping the profile, so it's not possible to know what schema is used when loading the profile/quantizing. Hence I am removing this verify.

This unbreaks models with relus when loading a profile that was dumped with `-quantization-schema=symmetric`. I verified this with Resnet50. I also added a unit test for Relu with symmetric parameters.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -265,8 +265,6 @@ static void verifyRelu(NodeValue result, NodeValue input) {
   if (input.getType()->isQuantizedType()) {
     assert(result.getType()->isQuantizedType());
     checkSameShape(result, input);
-    assert(result.getType()->getOffset() == -128 &&
-           "Min fp32 value should be 0");
   } else {
     checkSameType(result, input);
   }
--- a/tests/unittests/quantizationTest.cpp
+++ b/tests/unittests/quantizationTest.cpp
@@ -568,6 +568,52 @@ TEST(Quantization, chooseQuantizationSymmetric) {
   EXPECT_NEAR(symmetricParams.scale, 16.0 / 255, 0.001);
 }
 
+/// Check that Relu can use our symmetric quantization schema.
+TEST(Quantization, reluCanUseSymmetricSchema) {
+  Context ctx;
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  Placeholder *input =
+      mod.createPlaceholder(ElemKind::FloatTy, {10}, "input", false);
+  auto *inputTensor = ctx.allocate(input);
+  auto IH = inputTensor->getHandle<float>();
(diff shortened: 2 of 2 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -265,8 +265,6 @@ static void verifyRelu(NodeValue result, NodeValue input) {
   if (input.getType()->isQuantizedType()) {
     assert(result.getType()->isQuantizedType());
     checkSameShape(result, input);
-    assert(result.getType()->getOffset() == -128 &&
```
rdzhabarov: There are currently 3 quantization schemas: asymmetric, symmetric, and symmetric with uint8.
If you use regular symmetric schema (not the one with uint8) you'll end up wasting 1 bit for relu and hence acquire higher accuracy loss.
I'm not really sure if it's a good idea. Do you have a use case that would want this and not the one with uint8 or asymmetric?

jfix71: Yeah it's not a good idea just for RELU by itself. But we currently only allow a single quantization schema to be set for an entire network. So for example for Resnet50, we cannot use symmetric quantization because of this assert -- if you profile with symmetric and then load the profile to quantize with it then this assertion fails.

Additionally, different backends might only support symmetric quantization. Or it could be faster/more efficient to execute with symmetric quantization since it's simpler to implement than asymmetric. So I think we should allow for this even if there are obvious downsides, as we don't know what the backend supports or doesn't.

tlepley-cadence: @rdzhabarov I think that as soon as the `symmetric` schema is requested by the user, the offset should by 0 for all quantized tensors, even if it would  result in a lower accuracy. It's a user choice, potentially dictated by the targeted hardware. If Glow fails compiling with this `symmetric` schema as soon as a RELU is present in the model, I think it's an issue. `symmetric with uint8` is available for user who want to gain 1 bit at the output of RELU.

rdzhabarov: @jfix71 @tlepley makes sense.

tests/unittests/quantizationTest.cpp
```diff
@@ -568,6 +568,49 @@ TEST(Quantization, chooseQuantizationSymmetric) {
   EXPECT_NEAR(symmetricParams.scale, 16.0 / 255, 0.001);
 }
 
+/// Check that Relu can use our symmetric quantization schema.
+TEST(Quantization, reluCanUseSymmetricSchema) {
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  Variable *input = mod.createVariable(ElemKind::FloatTy, {10}, "input");
```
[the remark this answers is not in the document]
jfix71: Done

PR #1658, opened 2018-09-17
[Placeholder] Convert some graphOptz tests to placeholders

bertmaher:
*Description*: Start moving graphOptzTest to placeholders & context.

*Testing*: These tests.

*Documentation*: N/A

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1923,6 +1923,33 @@ void Function::createLSTM(llvm::StringRef namePrefix,
 //                   Placeholder-builder methods.
 //===----------------------------------------------------------------------===//
 
+BatchNormalizationNode *
+Function::createBatchNormalization(Context &ctx, llvm::StringRef name,
+                                   NodeValue input, unsigned_t channelIdx,
+                                   float epsilon, float momentum) {
+  // Figure out how many channels are in the tensor.
+  size_t channels = input.dims()[channelIdx];
+
+  // Allocate the learnable parameters beta and gamma.
+  auto *beta = getParent()->createPlaceholder(ElemKind::FloatTy, {channels},
+                                              "beta", true);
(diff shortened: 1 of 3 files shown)


bertmaher:
Haven't finished all these yet but wanted to make sure I'm on the right track.  I'm assuming that even though these tests aren't really "executable" that we still want to use placeholder/context, for API consistency, right?

tests/unittests/graphOptzTest.cpp
```diff
@@ -115,11 +117,12 @@ TEST_F(GraphOptz, liveCodeNotEliminated) {
 }
```
qcolombet: Note: the test is failing because getVars don't include the placeholders.

rdzhabarov:
`check format` seems to fail here :)

lib/Graph/Graph.cpp
```diff
@@ -1923,6 +1923,34 @@ void Function::createLSTM(llvm::StringRef namePrefix,
...
+  // Figure out how many channels are in the tensor.
+  size_t channels = input.dims()[channelIdx];
+
+  // Allocate the learnable parameters beta and gamma.
+  auto *beta = getParent()->createPlaceholder(
+      ElemKind::FloatTy, {channels}, "beta", true);
+  ctx.allocate(beta)->init(glow::Tensor::InitKind::Zero, 0, getPRNG());
+
+  auto *gamma = getParent()->createPlaceholder(
+      ElemKind::FloatTy, {channels}, "gamma", true);
```
rdzhabarov: curious, why this is always trainable?

bertmaher: Hm, I was just copying this from the existing (non-Context) createBatchNorm.  Should it be otherwise?

nadavrot: I think that this is correct. Instead of creating trainable variables we create trainable placeholders and initialize them.

PR #1659, opened 2018-09-18
Refactor optimization passes to be more modular/OO

xsunfeng:
*Description*: The motivation is to be able to add pre-post hooks easily, like printing the IR after each optimization, calling the verifier and so on.  This change aims to make it more structured by refactoring the implementation of the optimizers so that all of them share a high level class Optimization/IROptimization.  The class features a run() method (Optimization::run) running on a Module. A getName() method is also provided that allows debugger to know which optimizer is executed. Start with one optimizer method.

*Testing*: ninja check, 100% tests passed, 0 tests failed out of 20.

### DIFF
--- a/lib/Optimizer/FunctionPass.h
+++ b/lib/Optimizer/FunctionPass.h
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef OPTIMIZER_FUNCTION_PASS_H
+#define OPTIMIZER_FUNCTION_PASS_H
+
+#include "glow/Graph/Graph.h"
+
+class FunctionPass {
+public:
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+  virtual bool run(glow::Function *F) = 0;
+  virtual const llvm::StringRef getName() const = 0;
+};
+
+#endif // OPTIMIZER_FUNCTION_PASS_H
--- a/lib/Optimizer/FunctionPass.h
+++ b/lib/Optimizer/FunctionPass.h
@@ -16,14 +16,24 @@
 #ifndef OPTIMIZER_FUNCTION_PASS_H
 #define OPTIMIZER_FUNCTION_PASS_H
 
-#include "glow/Graph/Graph.h"
+namespace glow {
 
+class Function;
+
+/// A high level class (that will be) shared by all optimizers.
 class FunctionPass {
 public:
   FunctionPass() = default;
+
   virtual ~FunctionPass() = default;
-  virtual bool run(glow::Function *F) = 0;
+
+  /// \returns true if the optimizer changes something.
+  virtual bool run(Function *F) = 0;
+
+  /// \returns the name of the optimizer.
   virtual const llvm::StringRef getName() const = 0;
 };
+
+} // namespace glow
 
 #endif // OPTIMIZER_FUNCTION_PASS_H
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -24,6 +24,8 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 
+#include "FunctionPass.h"
+
 #include <unordered_map>
 #include <unordered_set>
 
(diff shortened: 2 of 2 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1137,19 +1139,30 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
-static void optimizeConcatNodes(Function *F) {
-  auto &nodes = F->getNodes();
-
-  // For each node:
-  for (auto &node : nodes) {
-    if (auto *CN = dyn_cast<ConcatNode>(&node)) {
-      NodeValue newCN = simplifyConcatNode(F, CN);
-      if (newCN.getNode()) {
-        CN->getResult().replaceAllUsesOfWith(newCN);
+/// Optimize Concat nodes.
```
rdzhabarov: nit: class name is self-explanatory, I'd remove the comment.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1137,19 +1139,30 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
+class OptimizeConcatNodes : public Optimization {
+public:
+  OptimizeConcatNodes() : Optimization() {}
```
rdzhabarov: if you are not planning to do anything here, just use `= default` syntax.
Same for .dtor

lib/Optimizer/Optimization.h
```diff
@@ -0,0 +1,9 @@
+#include "glow/Graph/Graph.h"
```
rdzhabarov: All header files need to have a guard against including it several times.

lib/Optimizer/Optimization.h
```diff
@@ -0,0 +1,9 @@
...
+
+class Optimization {
+public:
+  Optimization() {}
+  virtual ~Optimization(){};
+  virtual bool run(glow::Function *F) { return false; }
```
rdzhabarov: This feels like a good candidate for "interface". I'd make `run` and `getName` pure virtual.

lib/Optimizer/Optimization.h
rdzhabarov: The convention here is to use full path from lib, like OPTIMIZER_OPTIMIZATION_H.

lib/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,14 @@
...
+#define OPTIMIZER_FUNCTION_PASS_H
+
+#include "glow/Graph/Graph.h"
+
+class FunctionPass {
+public:
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+  virtual bool run(glow::Function *F) = 0;
+  virtual const char *getName() = 0;
```
nadavrot: get name should be const.

lib/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,14 @@
+#ifndef OPTIMIZER_FUNCTION_PASS_H
+#define OPTIMIZER_FUNCTION_PASS_H
+
+#include "glow/Graph/Graph.h"
+
+class FunctionPass {
```
nadavrot: Please add a doxygen comment.

lib/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,14 @@
+#ifndef OPTIMIZER_FUNCTION_PASS_H
+#define OPTIMIZER_FUNCTION_PASS_H
+
+#include "glow/Graph/Graph.h"
```
nadavrot: Why do we depend on the graph? Please forward declare what you need.

lib/Optimizer/FunctionPass.h
nadavrot: It should return llvm::stringref.

rdzhabarov:
@xsunfeng not all comments are addressed, just making sure you are aware of this.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -24,6 +24,8 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 
+#include "FunctionPass.h"
```
rdzhabarov: could you move this to the top?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1136,20 +1138,31 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
+public:
+  OptimizeConcatNodes(){};
+  ~OptimizeConcatNodes(){};
+  bool run(Function *F) override {
+    auto &nodes = F->getNodes();
+    bool changed = false;
+    // For each node:
+    for (auto &node : nodes) {
+      if (auto *CN = dyn_cast<ConcatNode>(&node)) {
+        NodeValue newCN = simplifyConcatNode(F, CN);
```
rdzhabarov: would it make sense to move `simplifyConcatNode` to `private` method of OptimizeConcatNodes for better encapsulation?

lib/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+  FunctionPass() = default;
+
+  virtual ~FunctionPass() = default;
+
+  /// \returns true if the optimizer changes something.
+  virtual bool run(Function *F) = 0;
+
+  /// \returns the name of the optimizer.
+  virtual const llvm::StringRef getName() const = 0;
```
rdzhabarov: missing headers for StringRef here.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1136,20 +1138,31 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
+          continue;
+        }
       }
     }
+    return changed;
   }
-}
+
+  const llvm::StringRef getName() const override {
+    return "OptimizeConcatNodes";
```
rdzhabarov: this is a bit worrisome, you return a reference to temporary here.

rdzhabarov:
@xsunfeng any updates? are you planning to work on this?

Issue #1329, opened 2018-07-25
Add support for FP16 and bfloat


stoklund:
Working with accelerators, we are going to need support for 16-bit floating point types.

- **FP16** is the *binary16* format defined in IEEE 754-2008.
- **bfloat** is the Tensorflow floating point type with the same exponent range as a float.

| Type          | binary16 | bfloat | binary32 |
|--------|--------:|-------:|-------:|
| *k*, storage width in bits |16|         16|         32|
| *w*, exponent field width in bits| 5|8|8|
| *t*, trailing significand field width in bits|10|7|23|

We should at a minimum add these types to `glow::ElemKind` so we can represent low-precision tensors in these types. Ideally, also add interpreter support.

nadavrot:
I did some research and tried to find a way to implement fp16 arithmetic in Glow. FP16 is not defined in the c++ language specification. There are a few libraries that implement software emulation for FP16. half.sf.net is the most popular library. 

Both clang and gcc support the __fp16 type extension. This allows us to implement the most basic operations, such as type conversion and basic arithmetic. The code below works with clang:

```
float foo(__fp16 *a, __fp16 *b) { return *a+*b; }
```

I couldn't find any mention of __fp16 in msvc. 
 
My recommendation would be to alias __fp16 to some other name (such as glow::float16) and use that type to implement fp16 support. If we ever need to support msvc we could replace the type alias with some platform specific implementation.

qcolombet:
Will look at this soon.
Assigning to myself.

qcolombet:
High level plan:
1. Add HalfTy type
2. Implement the support in the interpreter
3. Add a conversion mode FP32 -> FP16 for testing purposes
4. Add a FP16 quantization scheme
5. Add FP16 in the importer
6. Implement the support in the CPU

nadavrot:
@qcolombet Sounds like a great plan! 

I think that for item-4 the process should just be a conversion of the different nodes (1:1 conversion without profile guided stuff). For item-6 (CPU support) we could just "legalize" the graph from F16 to F32 (essentially reverse item-4). What do you think?

qcolombet:
> I think that for item-4 the process should just be a conversion of the different nodes (1:1 conversion without profile guided stuff)

Agree, let us stick to the basis first :).

> For item-6 (CPU support) we could just "legalize" the graph from F16 to F32 (essentially reverse item-4)

[...]

PR #1660, opened 2018-09-18
[Type] Add a half precision type

qcolombet:
Note: The choice of the library is debatable, see #1329. If we want to go without a library, we need to bump the llvm require to 6.0 (released last March).

*Description*:
This is the first step for adding the support of half precision
floating point numbers.

This patch adds the ElemKind HalfTy and teach the tensor class
how to deal with that type.

This also adds a dependency to a half precision library.

*Testing*:
ninja check
New test that creates a half typed tensor and randomizes it.

*Documentation*:

Added a section explaining how to get the library.

This is related to #1329


qcolombet:
FWIW, I feel adding this dependency is less risky than writing code. This library is well tested, has been around for years, is super stable, and compliant with IEEE 754. Moreover, we are talking about one header here.

On top of that, it has nice features like the ability to write half literal.

All in all, I believe this is a good compromise.

qcolombet:
Anyway, updated PR without the dependency on the library.
I'm implementing our own half class, this PR is blocked for now.

nadavrot:
@qcolombet This PR looks good. I think that you did not commit Half.h

qcolombet:
>  I think that you did not commit Half.h

Yes, I am writing it :).

include/glow/Base/Tensor.h
```diff
@@ -647,6 +649,19 @@ template <class ElemTy> class Handle final {
     }
   }
 
+  /// Fill the tensor with uniformly distributed values in the range
+  /// [low .. high].
```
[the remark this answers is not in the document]
qcolombet: Interesting, so all our comments are wrong :).
I'll update them all.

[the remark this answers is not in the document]
rdzhabarov: `int` is fine ;) (most or all comments should be correct) http://www.cplusplus.com/reference/random/uniform_int_distribution/

include/glow/Support/Half.h
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_SUPPORT_HALF_H
+#define GLOW_SUPPORT_HALF_H
+
+#include <iostream>
```
nadavrot: no iostream please.

include/glow/Support/Half.h
```diff
@@ -0,0 +1,53 @@
...
+  bool operator==(const half &b) const { return data_ == b.data_; }
+
+  /// Cast operators.
+  operator double() const { return double(data_); }
+  operator float() const { return float(data_); }
+}; // End class half.
+
+inline std::ostream &operator<<(std::ostream &os, const half &val) {
+  return os << float(val);
+}
```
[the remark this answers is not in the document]
nadavrot: I am not sure.

include/glow/Support/Half.h
qcolombet: Yeah that's because of EXPECT_EQ.
I'll dig to see how I can avoid it.

qcolombet: Found a way around. The default printer prints the raw bytes of the class and use a cast to long long to do that.
Added the cast and now we don't need iostream anymore. (Though that prints the raw bytes)

tests/unittests/halfTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+  half a = 3.5;
+  half b = 3.0;
+  EXPECT_EQ(a < b, float(a) < float(b));
+  EXPECT_FALSE(a < b);
+}
+
+TEST(Half, eq) {
+  half a = 3.5;
+  half b = 3.0;
+  EXPECT_EQ(a == b, float(a) == float(b));
```
nadavrot: All of the inputs (a, b) in all of the tests are the same. We could just combine all of them into one long test instead of having multiple tests.

qcolombet:
> Is it common to use "half" for 16-bit floating type? I think it's the first time I see it. Everybody in ML usually say "fp16". It could be confusing for new readers to see HalfTy everywhere.

It is common in GPU land + LLVM use "half" in its IR :).
I don't mind either way. I'll rename to FP16.

tests/unittests/halfTest.cpp
qcolombet: I'll change the values. I personally like the fact that the tests have different "names".

qcolombet:
We have a problem, our gcc on travis doesn't like __fp16:
../include/glow/Support/Half.h:24:17: error: ‘__fp16’ does not name a type
 using float16 = __fp16;

qcolombet:
Travis also fails at link time for clang.
It seems we miss some compiler-rt functions:
../include/glow/Support/Half.h:38: undefined reference to `__gnu_h2f_ieee'

artemrakhov:
I see that ONNX has `FLOAT`, `FLOAT16` and `DOUBLE`:
https://github.com/onnx/onnx/blob/a198b99bcc57c4e4b658d14d19e5c032721574a4/onnx/onnx.proto#L293

I'd prefer to follow their naming, because we are supporting their format after all. But maybe I'm biased because I haven't worked with LLVM.

qcolombet:
@Maratyszcza concurred that __fp16 is only supported on ARM for gcc.
I've checked _Float16 and clang supports it in 6+. For gcc, despite the type being standard, it is only supported for ARM and AArch64 again AFAICT, https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html#Floating-Types.

Given I ended up writing a wrapper, I believe now it is natural to use the same FP16 lib as C2. That way we don't add yet another dependency.

qcolombet:
- Reworked to use @Maratyszcza (a.k.a. the one used in C2) FP16 library.
- Renamed half into float16.

opti-mix:
@qcolombet Do you want to rename the PR to reflect the new naming of the half type?

qcolombet:
@opti-mix Sure!

include/glow/Support/Float16.h
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_SUPPORT_FLOAT16_H
+#define GLOW_SUPPORT_FLOAT16_H
+
+#include "fp16.h"
+
+#include <cstdint>
+#include <cstring>
```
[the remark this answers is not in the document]
qcolombet: Good catch! A relic of when I need to use memcpy

tests/unittests/tensorsTest.cpp
```diff
@@ -62,6 +62,18 @@ TEST(Tensor, randomizeInt) {
...
 
+TEST(Tensor, randomizeFloat16) {
+  PseudoRNG PRNG;
+  Tensor T(ElemKind::Float16Ty, {10, 10});
+  auto H = T.getHandle<float16_t>();
+  H.randomize(-50, 50, PRNG);
+
+  // Check that all of the numbers fall in the range -50 to 50.
+  for (size_t i = 0, e = H.size(); i < e; i++) {
+    EXPECT_NEAR(H.raw(0), 0, 50);
```
artemrakhov: `H.raw(i)`

qcolombet: 😱
Nice catch!

qcolombet: I did a copy paste from the previous test and that one is wrong too!!

artemrakhov:
So right now we cast numbers to float32, do compute and cast back to float16? Is there any way to operate directly in float16?

qcolombet:
> So right now we cast numbers to float32, do compute and cast back to float16? Is there any way to operate directly in float16?

No, unless we develop our own library or use another one. The one I originally proposed was supposedly better on that front...

That said, my understanding is that we are mostly fine accuracy wise [1], plus we already do crazy reassociations and whatnot in the backend (for vectorization for instance) that don't play nice with numerical stability anyway.
[1] https://github.com/apple/swift-llvm/blob/stable/lib/Transforms/InstCombine/InstCombineCasts.cpp#L1498
E.g., for add and sub:
        // Specifically, if OpWidth >= 2*DstWdith+1 and DstWidth is sufficient
        // to represent both sources, we can guarantee that the double
        // rounding is innocuous (See p50 of Figueroa's 2000 PhD thesis,
        // "A Rigorous Framework for Fully Supporting the IEEE Standard ..."
        // for proof of this fact).

include/glow/Base/Type.h
```diff
@@ -371,6 +378,9 @@ struct Type final {
     switch (Ty) {
     case ElemKind::FloatTy:
       return sizeof(float);
+    case ElemKind::Float16Ty:
+      static_assert(sizeof(float16_t) == 2, "Half precision should be 16-bit");
```
rdzhabarov: would it be better to move this assert close to the aliasing on top?

qcolombet: Works for me! Done.

Issue #1402, opened 2018-08-03
Need to implement channel-wise quantized fullyconnected node.


nadavrot:
One technique to improve the accuracy of quantized fullyconnected operations is channel-wise quantization. In this technique the activations remain regular quantized activations, but the FC weights are quantized with special per-channel information. PyTorch has support for this operator[1] 

To implement this you will need to do the following:
1. Add new target independent node to the Interpreter and CPU backends.
2. Implement the IR-lowering for the new target-independent FC. 
3. Implement a special kind of FC-quantization and lowering that lowers the FC into row-wise FC. This method will need to inspect the weights and extract the channel-specific information. 

[1] - https://github.com/caffe2/caffe2/blob/master/caffe2/operators/fused_rowwise_8bit_conversion_ops.cc

@rdzhabarov @jfix71

rdzhabarov:
The plan looks good to me, I think we'd need to have a way to control channel-wise quantization through some CLI flag as well. We could also extend this issue to cover Conv, but FC would have higher priority.

nadavrot:
@beicy FYI

rdzhabarov:
@beicy feel free to grab this and I'm happy to help with anything related to this issue. Otherwise, I'll tackle this some time in between ONNXIFI work.

beicy:
The design for this issue is described as follows:  

**1. Two new nodes are introduced for row-wise quantization** 

**- 1.1 FloatToFused8BitRowwiseQuantize**
The input of this op is a 2-D tensor (i.e. N * M) with float type for each element.  The output is a 2-D tensor with int8_t type. The shape of the output is (N * (M + 8)). For each row of the output, the stored data layout is : 

| ... quantized int8 data...|scale (float)|offset (int32_t)| 
| ...... M...bytes.................|.....4B..........|....4B................|

The **_scale_** and **_offset_** are chosen based on the min and max of each row of the input tensor. After the **_scale_** and the **_offset_** of each row is generated, the input data of this row are quantized and stored in the first M bytes of each output row.

**- 1.2 Fused8BitRowwiseQuantizedToFloat**
This is to de-quantize the 1.1 op. The input is a 2-D tensor (N * (M+8)) with int8_t type and the output is a 2-D tensor (N * M) with float type.
[...]

PR #1665, opened 2018-09-18
[Quantization] Add Row-Wise FC support for interpreter backend.

beicy:
*Description*:
This PR supports Row-wise quantized FC node for interpreter backend. In practice, this node can be  converted form FC during quantization time. It has 5 inputs :  

1. input (I):  regular quantized.  It is a 2D tensor with int8_t type.
2. bias(B)  : regular quantized. It is a 1D tensor with int8_t type.
3. weights(W): row-wise quantized.  It is quantized during compilation time. The scale and offset for each row are stored in the two tensor below. 
4. scales: the scale of each row of quantized weights. 
5. offsets: the offset of each row of quantized weights.

**Node**: Glow FC formula is : I * W + B, while the row-wise FC is : I * transpose(W) + B.
 
*Testing*: 
added rowWiseFCTest in operatorTest unittest.

*Documentation*:
[#1402]


tools/ClassGen/NodeGen.cpp
```diff
@@ -107,6 +107,18 @@ int main(int argc, char **argv) {
                     "Weights tensor are multiplied, and then the Bias tensor "
                     "is added to it, producing the Output.");
 
+  BB.newNode("RowWiseFullyConnected")
+      .addInput("Input")
+      .addInput("Weights")
+      .addInput("Bias")
+      .addResultFromCtorArg()
+      .setDocstring(
+          "Creates a RowWiseFullyConnected node where the Input 2-D tensor and "
```
artemrakhov:
```
          "Creates a RowWiseFullyConnected node where the Input matrix and "
          "the transpose of Weights matrix are multiplied, and then the Bias "
          "vector is broadcast-added to the result. Input, Bias and Result "
          "are regularly quantized, while Weights use row-wise quantization.");
```

include/glow/Quantization/Base/Base.h
```diff
@@ -99,6 +101,10 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
 TensorQuantizationParams chooseQuantizationParams(float min, float max,
                                                   Schema schema = Asymmetric);
 
+/// Row-Wise quantize the const tensor \p input. The param \p input is a /// 2-D
```
artemrakhov: `Row-wise quantize the tensor \p input. The \p input is a 2-D`

lib/Graph/Nodes.cpp
```diff
@@ -647,6 +647,37 @@ void TopKNode::verify() const {
...
+  auto dest = getResult();
+
+  (void)src;
+  (void)weights;
+  (void)scales;
+  (void)offsets;
+  (void)bias;
+  (void)dest;
+
+  assert(src.dims().size() == weights.dims().size() && src.dims().size() == 2 &&
```
artemrakhov: It's more readable to write
```
assert(src.dims().size() == 2 && weights.dims().size() == 2 && dest.dims().size() == 2 &&
       bias.dims().size() == 1 && scales.dims().size() == 1 && ... 
```

include/glow/Quantization/Base/Base.h
```diff
@@ -99,6 +101,13 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
 TensorQuantizationParams chooseQuantizationParams(float min, float max,
                                                   Schema schema = Asymmetric);
 
+/// Row-wise quantize the tensor \p input. The param \p input is a 2-D
+/// tensor (i.e. M * N), \p scales and \p offsets are generated by each row of
+/// \p input, \p qData is 2D tensor quantized from \p input using \p scales and
```
artemrakhov: qData -> result?

include/glow/Quantization/Base/Base.h
```diff
@@ -99,6 +101,13 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
...
+/// \p offsets for each row.
+void rowWiseQuantization(Tensor &input, Tensor *qData, Tensor *scales,
```
artemrakhov: Can we do `const Tensor &input` to denote that the function does not change it?

lib/Graph/Graph.cpp
```diff
@@ -629,6 +631,31 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
 
+RowwiseQuantizedFullyConnectedNode *
+Function::createRowwiseQuantizedFullyConnected(llvm::StringRef name, NodeValue input,
+                                      Node *W, Node *B, TypeRef outTy) {
+  // Since W is constant, quantize it in compilation time.
+  // The quantized data is in qWeights, the scale of each row is in scales,
+  // and the offset of each row is in offsets.
+  auto *weights = cast<Variable>(W);
+  auto &wT = weights->getPayload();
+  size_t row = wT.dims()[0];
```
artemrakhov: row -> numRows

lib/Graph/Graph.cpp
```diff
@@ -629,6 +631,31 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
 }
 
+RowwiseQuantizedFullyConnectedNode *
+Function::createRowwiseQuantizedFullyConnected(llvm::StringRef name, NodeValue input,
+                                      Node *W, Node *B, TypeRef outTy) {
+  // Since W is constant, quantize it in compilation time.
+  // The quantized data is in qWeights, the scale of each row is in scales,
+  // and the offset of each row is in offsets.
+  auto *weights = cast<Variable>(W);
+  auto &wT = weights->getPayload();
```
artemrakhov: wT -> wTensor

include/glow/Graph/Graph.h
```diff
@@ -268,6 +268,15 @@ class Function final : public Named {
   FullyConnectedNode *createFullyConnected(llvm::StringRef name,
                                            NodeValue input, size_t outDepth);
 
+  /// Create a row-wise fully connected node. This node is only used in
+  /// quantization. Parm \p input and \p B are quantized in regular way,
+  /// \p W is the constant weights and will be row-wise quantized during
+  /// compiler time.
+  RowwiseQuantizedFullyConnectedNode *createRowwiseQuantizedFullyConnected(llvm::StringRef name,
+                                                         NodeValue input,
+                                                         Node *W, Node *B,
```
artemrakhov: Why do we accept `Node*` here, and not `Storage` or `Variable`?

include/glow/Quantization/Base/Base.h
artemrakhov: rowWise -> rowwise, to match Caffe2

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+      0, 0.00001, mod_.getPRNG());
+  llvm::cast<Variable>(weights)->getPayload().getHandle().randomize(
+      -1.1, 1.1, mod_.getPRNG());
+
+  // The FC fomula is I * W + B, while the RWFC is I * transpose(W) + B.
+  // So get the tranpose of weights from the above FC.
+  auto *newWeights = mod_.createVariable(
+      ElemKind::FloatTy, {weights.dims()[1], weights.dims()[0]}, "newW");
+
+  genericTranspose(&(llvm::cast<Variable>(weights)->getPayload()),
```
artemrakhov: Could you use normal `transpose()` here?

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
   }
 }
 
+/// Test RowwiseQuantizedFullyConnected Node.
+TEST_P(InterpOnly, rowWiseFCTest) {
+  // In this test we subtract the outputs of a row-wise quantized FC and a
+  // floating-point FC and ensure that the error is below some low value.
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {1, 10, 10, 3}, "in");
```
artemrakhov: Let's use 2D input here, so you won't have to flatten later.

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  auto weights = fc->getWeights();
+  auto bias = fc->getBias();
+
+  input->getPayload().getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  llvm::cast<Variable>(bias)->getPayload().getHandle().randomize(
+      0, 0.00001, mod_.getPRNG());
+  llvm::cast<Variable>(weights)->getPayload().getHandle().randomize(
+      -1.1, 1.1, mod_.getPRNG());
+
+  // The FC fomula is I * W + B, while the RWFC is I * transpose(W) + B.
```
artemrakhov: RWFC -> RWQFC

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  auto H = res->getPayload().getHandle();
+  // Check that there aren't too many elements with a difference in the results
+  // of greater than 0.2.
+  int count = 0;
+  for (int i = 0, e = H.size(); i < e; i++) {
+    if (std::abs(H.raw(i)) > 0.2) {
+      count++;
+    }
+  }
+  EXPECT_LT(count, 2);
```
artemrakhov: What value do you actually observe? 0 or 1? I find `LT` a bit confusing here, because reader has to do this -1 in their head. Lets use `LE` or even `EQ`.

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  auto *fc = F_->createFullyConnected("FC", input, 30);
+  auto *res =
+      mod_.createVariable(ElemKind::FloatTy, fc->getResult().dims(), "res");
+
+  auto weights = fc->getWeights();
+  auto bias = fc->getBias();
+
+  input->getPayload().getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  llvm::cast<Variable>(bias)->getPayload().getHandle().randomize(
+      0, 0.00001, mod_.getPRNG());
```
artemrakhov: Why is bias so small? I have a feeling that such small value doesn't allow for good testing. Does the test still pass if you remove bias-related code (`sum += B`) from Interpreter implementation?

lib/Graph/Graph.cpp
```diff
@@ -629,6 +631,31 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
+  size_t row = wT.dims()[0];
+  auto *qWeights = getParent()->createVariable(
+      ElemKind::Int8QTy, wT.dims(), 0.0, 0, "weights.rwq",
```
artemrakhov: rwq -> rwqfc (here and below)

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,66 @@ TEST_P(InterpAndCPU, insertTensorTest) {
   }
 }
 
+/// Test RowwiseQuantizedFullyConnected Node.
+TEST_P(InterpOnly, rowWiseFCTest) {
```
artemrakhov: rowWise -> rowwiseQuantized

tests/unittests/OperatorTest.cpp
beicy: The value was 1.

tests/unittests/OperatorTest.cpp
beicy: I copied this settings from IntFC in OperatorTest.cpp. Will check this bias.

include/glow/Quantization/Base/Base.h
beicy: Here qData means quantized data. Since scales and offsets can be viewed as the results of this function, I think qData should be better.

include/glow/Quantization/Base/Base.h
beicy: I tried this one, but here was an error "note: candidate function not viable: 'this' argument has type 'const glow::Tensor', but method is not marked const
template <class ElemTy> Handle<ElemTy> Tensor::getHandle() {"

include/glow/Graph/Graph.h
beicy: Since previously, I added one node to do the row-wise quantization, and then W was the result of that RowWiseQuantization node. I forgot to update this interface... Now it is updated to "Variable".

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,63 @@ TEST_P(InterpAndCPU, insertTensorTest) {
   }
 }
 
+/// Test RowwiseQuantizedFullyConnected Node.
+TEST_P(InterpOnly, rowwiseQuantizedFCTest) {
+  // In this test we subtract the outputs of a row-wise quantized FC and a
+  // floating-point FC and ensure that the error is below some low value.
+  auto *input = mod_.createVariable(ElemKind::FloatTy, {1, 300}, "in");
```
rdzhabarov: can you use the new interface with "createPlaceholder"?

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,63 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+
+  // The FC fomula is I * W + B, while the RWQFC is I * transpose(W) + B.
+  // So get the tranpose of weights from the above FC.
+  auto *newWeights = mod_.createVariable(
+      ElemKind::FloatTy, {weights.dims()[1], weights.dims()[0]}, "newW");
+
+  newWeights->getPayload().transpose(
+      &(llvm::cast<Variable>(weights)->getPayload()), {1, 0});
+
+  TypeRef resTy = mod_.uniqueType(ElemKind::Int8QTy, res->dims(), 0.15, 4);
```
rdzhabarov: can you comment on how quantization params are chosen?

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,63 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  Context ctx;
+  EE_.compile(CompilationMode::Infer, F_, ctx);
+  EE_.run();
+
+  auto H = res->getPayload().getHandle();
+  // Check that there aren't too many elements with a difference in the results
+  // of greater than 0.2.
+  int count = 0;
+  for (int i = 0, e = H.size(); i < e; i++) {
+    if (std::abs(H.raw(i)) > 0.2) {
```
rdzhabarov: why not here just do, EXPECT_LT(std::abs(H.raw(i)), 0.2)?

rdzhabarov: it's hard to tell whether this works fine or not without knowing the actual numbers coming out of the FCs.
Have you looked at the results produced by rwFC and regular FC how big/small those are?

tests/unittests/OperatorTest.cpp
beicy: There is an unittest IntFC which tests the quantized FC. So I just copied the params from there...

rdzhabarov: I think it'd be useful to at least specify the floating point ranges represented by these {s,o}

include/glow/Graph/Graph.h
```diff
@@ -276,6 +276,14 @@ class Function final : public Named {
   FullyConnectedNode *createFullyConnected(llvm::StringRef name,
                                            NodeValue input, size_t outDepth);
 
+  /// Create a row-wise quantized fully connected node. This node is only used
```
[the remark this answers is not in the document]
beicy: For example, a Matmul node, we can say it is quantized or not by checking the input/output type. But this rwqfc is designed for better quantization only.

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,65 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  auto *weights = llvm::cast<Placeholder>(fc->getWeights());
+  auto *bias = llvm::cast<Placeholder>(fc->getBias());
+
+  auto *inputTensor = ctx.allocate(input);
+  auto *weightsTensor = ctx.get(weights);
+  auto *biasTensor = ctx.get(bias);
+  auto *resTensor = ctx.allocate(res);
+
+  inputTensor->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  biasTensor->getHandle().randomize(0, 0.00001, mod_.getPRNG());
```
artemrakhov: Why bias is so small?

So we have a code, doing 3 things: X, Y and Z. Ideally, you want to have a test, such that if you remove X-functionality from your code (so that it only does Y and Z), than test fails. Same for Y and Z. In other words, test must fail if code "forgets" to do X or Y or Z.

Right now if we remove "bias computation" (`sum += B`) code from Interpreter, the test will still pass. This is not good, because it means that we don't cover it. I have a feeling that if you increase bias magnitude, then we'll have coverage.

tools/ClassGen/NodeGen.cpp
```diff
@@ -107,6 +107,20 @@ int main(int argc, char **argv) {
...
+      .addInput("Input")
+      .addInput("Weights")
+      .addInput("Scales")
+      .addInput("Offsets")
+      .addInput("Bias")
+      .addResultFromCtorArg()
+      .setDocstring(
+          "Creates a RowwiseQuantizedFullyConnected node where the Input "
+          "matrix and the transpose of Weights matrix are multiplied, and "
+          "then the Bias vector is broadcast-added to the result. Inupt, "
```
artemrakhov: Inupt -> Input

tests/unittests/OperatorTest.cpp
```diff
@@ -3362,6 +3362,65 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  auto *sub = F_->createSub("compare", dequantRes, fc);
+
+  F_->createSave("save", sub, res);
+  EE_.compile(CompilationMode::Infer, F_, ctx);
+  EE_.run();
+
+  auto H = ctx.get(res)->getHandle();
+  // The difference in the results should be less than 0.2.
+  for (int i = 0, e = H.size(); i < e; i++) {
+    EXPECT_LT(std::abs(H.raw(i)), 0.2);
```
artemrakhov: Repeating comment by @rdzhabarov :

> it's hard to tell whether this works fine or not without knowing the actual numbers coming out of the FCs.
> Have you looked at the results produced by rwFC and regular FC how big/small those are?

lib/Quantization/Base/Base.cpp
```diff
@@ -238,5 +238,37 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
...
+  auto srcH = input->getHandle<float>();
+  auto destH = qData->getHandle<int8_t>();
+  auto scalesH = scales->getHandle<float>();
+  auto offsetsH = offsets->getHandle<int32_t>();
+  for (size_t i = 0; i < idim.height; i++) {
+    // auto rSrc = srcH.extractSlice(i).getHandle<float>();
+    // auto res = rSrc.minMaxArg();
+    // float min = rSrc.raw(res.first);
+    // float max = rSrc.raw(res.second);
+    float min = srcH.at({i, 0});
```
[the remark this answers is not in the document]
beicy: Using extractSlice had some stack issue in Travis CI build. In addition, last time I use this  extractSlice to abstract a row, and the last element in the row was incorrect. I just marked this function on my side and plan to debug more later.

lib/Quantization/Base/Base.cpp
```diff
@@ -263,5 +263,37 @@ std::vector<int8_t> createMapping(TypeRef inTy, TypeRef outTy,
...
+  auto srcH = input->getHandle<float>();
+  auto destH = qData->getHandle<int8_t>();
+  auto scalesH = scales->getHandle<float>();
+  auto offsetsH = offsets->getHandle<int32_t>();
+  for (size_t i = 0; i < idim.height; i++) {
+    auto rSrc = srcH.extractSlice(i).getHandle<float>();
+    auto res = rSrc.minMaxArg();
+    float min = rSrc.raw(res.first);
+    float max = rSrc.raw(res.second);
+    /*float min = srcH.at({i, 0});
```
artemrakhov: Why commented code?

beicy:
@artemrakhov @rdzhabarov : For the unnitest, the regular FC result is between (-5.0, 3), but around 3/4 are around (-1, 1). So I modified the test: let H1 is the abs diffs between RWQFC and FC, H2 is the abs diffs between QFC and FC. The test is EXPECT_LT(H1, H2).

include/glow/Graph/Graph.h
```diff
@@ -276,6 +276,15 @@ class Function final : public Named {
...
+  /// in quantization. Args \p input and \p B are quantized in regular way, \p W
+  /// is the constant weights and will be row-wise quantized during compile
```
rdzhabarov: maybe specify that it's "during node creation time" since "during compile time" sounds like we have some further optimizations etc.

include/glow/Quantization/Base/Base.h
```diff
@@ -107,6 +108,13 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
 std::vector<int8_t> createMapping(TypeRef inTy, TypeRef outTy,
                                   std::function<float(float)> f);
 
+/// Row-wise quantize the tensor \p input. The param \p input is a 2D
+/// tensor (i.e. M * N), \p scales and \p offsets are generated by each row of
+/// \p input, \p qData is 2D tensor quantized from \p input using \p scales and
+/// \p offsets for each row.
+void rowwiseQuantization(Tensor *input, Tensor *qData, Tensor *scales,
```
rdzhabarov: consider change the name to better reflect intent, like rowwiseTensorTransform or something.

lib/Graph/Graph.cpp
```diff
@@ -26,6 +26,8 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "glow/Quantization/Base/Base.h"
```
rdzhabarov: can you move that to the top to "glow" headers.

lib/Graph/Graph.cpp
```diff
@@ -639,6 +641,30 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
+Function::createRowwiseQuantizedFullyConnected(Context &ctx,
+                                               llvm::StringRef name,
+                                               NodeValue input, Storage *W,
+                                               Node *B, TypeRef outTy) {
+  // Since W is constant, quantize it in compilation time.
+  // The quantized data is in qWeights, the scale of each row is in scales,
+  // and the offset of each row is in offsets.
+  Placeholder *weights = llvm::cast<Placeholder>(W);
+  size_t numRows = W->getType()->dims()[0];
+  auto *qWeights = getParent()->createPlaceholder(
```
rdzhabarov: can you comment when scale and offset are set to 0.0, 0.

lib/Graph/Graph.cpp
```diff
@@ -639,6 +641,30 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
 
+RowwiseQuantizedFullyConnectedNode *
+Function::createRowwiseQuantizedFullyConnected(Context &ctx,
+                                               llvm::StringRef name,
+                                               NodeValue input, Storage *W,
+                                               Node *B, TypeRef outTy) {
+  // Since W is constant, quantize it in compilation time.
+  // The quantized data is in qWeights, the scale of each row is in scales,
+  // and the offset of each row is in offsets.
+  Placeholder *weights = llvm::cast<Placeholder>(W);
```
rdzhabarov: this would be a constant weight and we'd need to fix that later

include/glow/Quantization/Base/Base.h
rdzhabarov: as we expect all Tensors to be not null, I'd do Tensor & here (with const when needed).

beicy:
@artemrakhov : The following is the data after using "chooseQuantizationParams" to get the new scales and offsets( for input, output, weights( for regular quantized FC) , bias):
```
Data 1:
FC Result       RWQFC diff      QFC diffs
1.622382,	 0.027618, 	0.027618
1.911586,        0.038414, 	0.038414
-0.521064,	 0.021064, 	0.028936 * **
-8.882705,	 0.032704, 	0.032704
3.610024,	 0.039976, 	0.039976 **
-1.948685,	 0.001315, 	0.001315
-0.865643,	 0.034357, 	0.034357 **
-2.998229,	 0.001771, 	0.001771
-3.145861,	 0.004139, 	0.004139
3.915103,	 0.015103, 	0.015103
```
We can see that one number from RWQFC is better than QFC(marked as *). And both of them, there are three diffs which are greater than 2% of the expected result (marked as **).

The following is the data without adjusting the quantization params:
```
FC Result       RWQFC diff
1.622382,	 0.027618, 
1.911586,	 0.038414, 
-0.521064,	 0.071064, + 
-8.882705,	 0.032704, +
3.610024,	 0.010024,  -
-1.948685,	 0.001315, 
-0.865643,	 0.034357, 
-2.998229,	 0.001771, 
-3.145861,	 0.004139, 
3.915103,	 0.015103, 
```
The number marked as "+" means greater than "quantization params adjusting version" and "-" means smaller. Btw, there are still 3 diffs which is > 2%.

Actually, the size of input and output, the data range, the quantization params are working together to get the final result. The "Data 1" is the best I can get by adjusting the params (tensor size, data range, quantized params).  I use the result: the (number of accuracy drop > 2% ) <= 3 for this unittest so far.

include/glow/Quantization/Base/Base.h
beicy: @rdzhabarov The input tensor is const. However, if I use const here, "getHandle" can't be used since "'this' argument has type 'const glow::Tensor', but method is not marked const template <class ElemTy> Handle<ElemTy> Tensor::getHandle()" and the code will be complicated. So I didn't put "const" here.

tests/unittests/OperatorTest.cpp
```diff
@@ -3487,6 +3487,68 @@ TEST_P(InterpAndCPU, insertTensorTest) {
...
+  // floating-point FC and ensure that the error is below some low value.
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {2, 100}, "in", false);
+  auto *fc = F_->createFullyConnected(ctx_, "FC", input, 5);
+
+  auto *weights = llvm::cast<Placeholder>(fc->getWeights());
+  auto *bias = llvm::cast<Placeholder>(fc->getBias());
+
+  ctx_.allocate(input)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ctx_.get(bias)->getHandle().randomize(0, 0.1, mod_.getPRNG());
```
artemrakhov: Same question :) isn't 0.1 too low comparing to other tensors? If you remove `sum += B` from the implementation, does the test fail?

lib/Quantization/Base/Base.cpp
```diff
@@ -262,5 +262,30 @@ std::vector<int8_t> createMapping(TypeRef inTy, TypeRef outTy,
   return mapping;
 }
 
+void tensorRowwiseQuantization(Tensor &input, Tensor &qData, Tensor &scales,
```
bertmaher: Nits:
- `output` would probably be clearer than `qData`.
- Can `input` be `const Tensor&`?

lib/Graph/Graph.cpp
bertmaher: If I understand correctly here I think we want to use `Variable`, not `Placeholder`, since the intent is that the RowwiseQuantizedFC is given constant weights.  Right?

If it's the case that we need constant weights here, can we make the function signature take a `Variable *` instead of a `Storage *`, so that the API is type safe?

lib/Quantization/Base/Base.cpp
beicy: If we put input as const Tensor&, we can't use "getHandle()" function : "'this' argument has type 'const glow::Tensor', but method is not marked const
template <class ElemTy> Handle<ElemTy> Tensor::getHandle()" .

Although we can use "getUnsafePtr()" method for const Tensor, the code will be complicated...

tests/unittests/OperatorTest.cpp
beicy: I remember the accuracy drop should be less then 2%... @rdzhabarov Please correct me if I am wrong. For the bias 0.1, if "sum+=b" is removed, the test failed -- actual: 4 vs 3. Actually, I tried to increase it to 1 and 2, but the result was pretty bad -- over 50% .  I also tried 0.01, 0.001, and the result is almost as good as "0.1".   From the experiment, I can say the the result is affected by all the params, but I don't know how :P

artemrakhov: So if bias is generated in range [0, 2), then we have a lot of differences between normal FC and RWQFC? Isn't this suspicious? One way to verify that our actual implementation is correct is to replicate this in Caffe2. Do we also see big divergence between FC and RWQFC in Caffe2 when bias is bigger?

lib/Graph/Graph.cpp
beicy: "If I understand correctly here I think we want to use Variable, not Placeholder, since the intent is that the RowwiseQuantizedFC is given constant weights. Right?" 
--- Yes, the weights should be constant.
So in the future, "Variable" means "constant" ?

tests/unittests/OperatorTest.cpp
beicy: "So if bias is generated in range [0, 2), then we have a lot of differences between normal FC and RWQFC?"
--- Yes. For regular quantized FC, we has the similar bad result. So I think this is something related to params.

lib/Quantization/Base/Base.cpp
rdzhabarov: I think it would make sense to add one more method
```
template <class ElemTy = float> const Handle<ElemTy> getHandle() const;
```
this should resolve the issue you are seeing.

tests/unittests/OperatorTest.cpp
rdzhabarov: > I remember the accuracy drop should be less then 2%... @rdzhabarov Please correct me if I am wrong

There are many factors which impact the %, like whether scale and offset are selected correctly to represent the data (we do not profile here, no fancy {s,o} selection based on distribution), type of the operation we validate. We should make test to pass, but be conservative with the threshold. No hard rule for 2%. I'd look at the data and choose what makes sense (if the value is too high then see if there are any bugs).

lib/Quantization/Base/Base.cpp
beicy: @rdzhabarov In that case, I think the class Handle need to be improved as well. The current constructor is `explicit Handle(Tensor *tensor) : tensor_(tensor)` , and the private member `tensor` is not const. So I think if we decide to support getHandle() for const Tensor,  it is better to create a separate repo :)

include/glow/Quantization/Base/Base.h
```diff
@@ -17,6 +17,7 @@
 #ifndef GLOW_QUANTIZATION_BASE_BASE_H
 #define GLOW_QUANTIZATION_BASE_BASE_H
 
+#include "glow/Base/Tensor.h"
```
jfix71: nit: You can forward declare Tensor here instead of including it.

lib/Graph/Graph.cpp
```diff
@@ -639,6 +639,35 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
+  size_t numRows = W->getType()->dims()[0];
+
+  // Sor far, if we want to create a storage with Int8QTy/Int16QTy/Int32QTy,
```
jfix71: nit: typo "Sor"

tools/ClassGen/InstrGen.cpp
```diff
@@ -120,6 +120,16 @@ int main(int argc, char **argv) {
...
 
+  BB.newInstr("RowwiseQuantizedFullyConnected")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("Src", OperandKind::In)
+      .addOperand("Weights", OperandKind::In)
+      .addOperand("Bias", OperandKind::In)
+      .addOperand("Scales", OperandKind::In)
+      .addOperand("Offsets", OperandKind::In)
+      .autoIRGen()
+      .autoVerify(VerifyKind::NoVerify);
```
jfix71: Please add some verification here, e.g. checks for `SameElementType`.

nadavrot:
Nice!  🆒 🎉 🍨

rdzhabarov:
Great work @beicy!

PR #1666, opened 2018-09-18
[Placeholders] Finish migration of GraphOptz

bertmaher:
*Description*: In summary.
*Testing*: N/A
*Documentation*: N/A

### DIFF
--- a/tests/unittests/graphOptzTest.cpp
+++ b/tests/unittests/graphOptzTest.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "glow/Graph/Context.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Node.h"
 #include "glow/Graph/Nodes.h"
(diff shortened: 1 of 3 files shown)


tests/unittests/graphOptzTest.cpp
```diff
@@ -1818,17 +1776,17 @@ TEST_F(GraphOptz, VarsCSE) {
   auto *SN = F_->createSigmoid("sigmoid", input2);
   auto *RN = F_->createRELU("relu", input3);
   auto *CN = F_->createConcat("concat", {TN, SN, RN}, /* axis */ 0);
-  F_->createSave("ret", CN);
+  F_->createSave(ctx_, "ret", CN);
```
rdzhabarov: mod_.createVariable was not changed.

bertmaher: Right - there are several tests in graphOptz that rely on a variable having a compile-time constant value.  These will eventually be renamed from Variable to Constant.

PR #1680, opened 2018-09-19
[GraphOptimizer] Optimization to merge Transpose into MatMul

aturetsk:
*Description*:
GLOW uses NHWC format for convolutions and this leads to appearance of a transpose before matmuls in some classification networks if NCHW is used as input layout. This patch adds optimization to get rid of such transpose by merging it into matmul and reordering matmul's weights.
*Testing*:
Added a test in graphOptzTest.
Verified with internal classification net.
*Documentation*:
N/A

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -592,6 +592,88 @@ static void mergeMatMul(Function *F) {
   }
 }
 
+/// Merge Transpose into MatMul.
+/// MatMul(Reshape(Transpose(X)), Weights) ->
+/// -> MatMul(Reshape(X), reordered Weights)
+/// Common sequence while using NCHW as input layout, because GLOW convolution
+/// layout is NHWC:
+/// Transpose([N, H, W, C]) -> [N, C, H, W]
+/// Reshape([N, C, H, W]) -> [N, C * H * W]
+/// MatMul([N, C * H * W], [C * H * W, K]) -> [N, K]
+static bool mergeTransposeIntoMatMul(Function *F) {
+  bool changed = false;
+  for (auto &node : F->getNodes()) {
+    auto *MMN = dyn_cast<MatMulNode>(&node);
+    if (!MMN) {
+      continue;
+    }
+
+    // MatMul RHS is constant weights.
+    auto *W = dyn_cast<Variable>(MMN->getRHS());
+    if (!W || !W->isPrivate()) {
+      continue;
+    }
+
+    // Linearizing Reshape precedes MatMul.
+    // The first dimension must be kept unchanged, the others are linearized.
+    auto *RN = dyn_cast<ReshapeNode>(MMN->getLHS());
+    if (!RN || !RN->hasOneUse() ||
+        RN->getInput().dims()[0] != RN->getDims()[0]) {
+      continue;
+    }
+
+    // Transpose precedes Reshape.
+    // The first dimension must be kept unchanged, the others can be shuffled
+    // in any way.
+    auto *TN = dyn_cast<TransposeNode>(RN->getInput());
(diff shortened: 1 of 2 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -592,6 +592,78 @@ static void mergeMatMul(Function *F) {
   }
 }
 
+/// Merge Transpose into MatMul.
```
nadavrot: I wonder about the reshape in the optimization. Is the reshape coming from FC node? Would it make sense to catch this pattern before lowering the FC?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -592,6 +592,78 @@ static void mergeMatMul(Function *F) {
...
+      // in any way.
+      auto *TN = dyn_cast<TransposeNode>(RN->getInput());
+      if (!TN || !TN->hasOneUse() || TN->getShuffle()[0] != 0) {
+        continue;
+      }
+
+      // Reshaped weight types and shuffle to be able to simply use
+      // Tensor::transpose for weight reordering.
+      ShapeVector shape, newShape;
+      llvm::SmallVector<unsigned_t, max_tensor_dimensions> shuffle;
```
nadavrot: Could you please add a few comments here that describe the transformation?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1906,6 +1978,12 @@ void glow::optimize(Function *F, CompilationMode mode) {
 
   optimizeReshape(F);
 
+  // Merge Transpose into MatMul.
+  // Run after optimizeReshape to ensure a single Reshapes precedes MatMul.
+  // Run DCE to ensure correct number of node users.
```
nadavrot: Why do we need the DCE here?

aturetsk: Optimizations do not erase nodes, they just do 'ReplaceAllUsesWith' and relay on DCE to clean up. This corrupts number of users for nodes (after transformation we usually have a newly created optimized use and old 'dead' use), i.e. node->hasOneUse() may return false, while after DCE it'll be true.

lib/Optimizer/GraphOptimizer.cpp
nadavrot: Great!

lib/Optimizer/GraphOptimizer.cpp
nadavrot: Oh, wow this is a bug!  Could you please file an issue to track this bug?  Optimizations should not corrupt the graph.

lib/Optimizer/GraphOptimizer.cpp
aturetsk: I realized I was wrong. Reshape comes even before lowering from Caffe2 importer:
https://github.com/pytorch/glow/blob/master/lib/Importer/Caffe2.cpp line 300.
So there is no much sense to move the optimization to pre-lowering stage.

aturetsk:
Used early exit in the optimization and added more comments describing transformation.

lib/Optimizer/GraphOptimizer.cpp
aturetsk: Created https://github.com/pytorch/glow/issues/1688

nadavrot:
Thank you for doing this work Andrey! @aturetsk

nadavrot:
Andrey, can we land this PR?
@aturetsk

nadavrot:
I will land this PR to avoid merge conflict. I just realized that you don't have permissions to land pull requests. Next time please add a comment on the PR and someone will land it as soon as the PR is approved. Sorry about the delay.

PR #1681, opened 2018-09-20
Fp16 interpreter support: Convolution

qcolombet:
*Description*:
This PR adds fp16 support for the convolution node in the interpreter

*Testing*:
Added test cases
*Documentation*: N/A

This depends on #1660 and is related to #1329

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -31,15 +31,20 @@ using namespace glow;
 //===----------------------------------------------------------------------===//
 
 // This is the floating point implementation of Convolution.
+template <typename ElemTy>
 void InterpreterFunction::fwdConvolutionInst_FloatImpl(
     Value *inV, Value *outV, Value *filterV, Value *biasV,
     llvm::ArrayRef<unsigned_t> kernelSizes, llvm::ArrayRef<unsigned_t> strides,
     llvm::ArrayRef<unsigned_t> pads, size_t group) {
 
-  auto inW = getWeightHandle(inV);
-  auto outW = getWeightHandle(outV);
-  auto filterW = getWeightHandle(filterV);
-  auto biasW = getWeightHandle(biasV);
+  static_assert(
+      std::is_floating_point<ElemTy>::value ||
+          std::is_same<float16_t, typename std::remove_cv<ElemTy>::type>::value,
+      "This implementation is for floating-point values only");
+  auto inW = getWeightHandle<ElemTy>(inV);
+  auto outW = getWeightHandle<ElemTy>(outV);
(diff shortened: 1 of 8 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -31,15 +31,19 @@ using namespace glow;
 //===----------------------------------------------------------------------===//
 
 // This is the floating point implementation of Convolution.
-void InterpreterFunction::fwdConvolutionInst_FloatImpl(
+template <typename ElemTy>
+typename std::enable_if<
```
nadavrot: Why do we need the enable_if section? If we need to protect the usage of this method, could we insert a static_assert into the body of the function to keep the declaration simple?

tests/unittests/OperatorTest.cpp
```diff
@@ -1291,6 +1291,92 @@ TEST_P(InterpAndCPU, IntBatchedArith) {
...
+  // In this test we generate a Floating-point based convolution and an integer
+  // convolution. We pass the same values and then subtract the results. We
+  // check that the results are below some known delta.
+
+  // In this test the output of the convolution is in the range [-256 ... 256].
+  // The inputs (specified below) are in the range [-1 .. 1],
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  auto *input = mod.createVariable(ElemKind::FloatTy, {1, 10, 10, 3}, "in");
```
nadavrot: Please use the new createPlaceholder API.

tests/unittests/graphTest.cpp
```diff
@@ -64,6 +65,81 @@ TEST(Graph, simpleTestConv) {
   EXPECT_GT(M.getInstrs().size(), 0);
 }
 
+/// Check that we can create convolution with float16.
+TEST(Graph, float16Conv) {
+  Module MD;
+  Function *F = MD.createFunction("F");
+  Node *K = MD.createVariable(ElemKind::Float16Ty, {4, 320, 200, 3}, "input");
```
nadavrot: ditto. Please don't add new createVariable call sites.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "thirdparty/onnx"]
 	path = thirdparty/onnx
 	url = https://github.com/onnx/onnx.git
+[submodule "thirdparty/fp16"]
```
rdzhabarov: i thought this is merged :)

qcolombet: No yet. It will disappear when I rebase.

lib/Backends/Interpreter/InterpreterNodes.cpp
qcolombet: Good point.
Done!

PR #1692, opened 2018-09-21
[bug fix] When schema is symmetric with unit8 we need to allow 0 check in the assert.

rdzhabarov:
*Description*:
schema == SymmetricWithUInt8 and offset == 0 makes the

```
assert(result.offset == -128 || schema != quantization::Schema::SymmetricWithUInt8)
```
fail

*Testing*:
./tests/quantizationTest

*Documentation*:
N/A

cc: @opti-mix who discovered assert failure on dumping profile for some models.

### DIFF
--- a/lib/Quantization/Base/Base.cpp
+++ b/lib/Quantization/Base/Base.cpp
@@ -230,9 +230,7 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
          "Symmetric quantization should be centered on 0");
   // The only valid offsets for symmetric quantization with uint8 support are 0
   // and -128.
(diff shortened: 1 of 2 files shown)


tlepley-cadence:
I've got the same issue and had to implement a local fix.

@rdzhabarov Does this fix really need to be labelled 'work in process' ? Can't this fix be accepted as is now ? Currently, `SymmetricWithUInt8` will fail as soon as one tensor has got offset 0 (what I assume happens with all models).

rdzhabarov:
@tlepley the only reason that was under WIP is that I needed to add a test to expose the bug.
Change itself was good to be merged.

PR #1693, opened 2018-09-24
[Placeholders] Migrate partitioning and partitionTest

bertmaher:


lib/Optimizer/Partition.cpp
```diff
@@ -151,9 +153,10 @@ FunctionDAG doPartitioning(Function *F, NodeFunctionMap &mapping) {
         }
 
         // Create a new variable to represent this dependence.
-        auto *tmp = mod->createVariable(
+        auto *tmp = mod->createPlaceholder(
```
rdzhabarov: why not let `createSave` to create placeholder (less typing here :))

lib/Optimizer/Partition.cpp
```diff
@@ -114,10 +115,10 @@ NodeFunctionMap selectBasicBlockPartitions(Function *F) {
 
 /// Given a function \p F and partitioning \p mapping, \return a FunctionDAG
 /// that contains appropriately-partitioned functions and their dependences.
-FunctionDAG doPartitioning(Function *F, NodeFunctionMap &mapping) {
+FunctionDAG doPartitioning(Function *F, NodeFunctionMap &mapping,
```
nadavrot: Conceptually partitioning should be independent of a specific context. Can we initialize (add a backing tensor) outside of the partitioning method?

bertmaher: Yep, done.  (I think I'd originally plumbed Context through only to perform `createSave`, but I've realized there's no good reason to do that.)

PR #1708, opened 2018-09-24
[Optimizer] Optimization to convert Placeholders to Variables (Constants)

### LINKED ISSUE #1623: Implement a method to convert Placeholder to Constant
Once we eliminate all mutable variables we'll be left with Placeholders that could be bound to specific Tensors by the Context, and to Constants, that are owned by the graph. The optimizer is only allowed to optimize Constants, because the Placeholders are observable by the outside world and must not be mutated. 

To enable optimizations we need to implement a method that takes a Graph and some context and turn some Placeholders into constants. For example, consider the trained mnist graph. After training the program and before running inference it could be a good idea to convert all of the weights (but not the inputs and outputs) into constants. This would allow the optimizer to touch and mutate the constants and generate better code. 

One possible signature: 
    makePlaceholderConstant(Context &ctx, Function * F, llvm::ArrayRef<Placeholder*> vars);

bertmaher:
*Description*: Add an optimization pass that replaces Placeholders with constants, so that the values can be used in optimization.  Placeholders are bound to values passed in via a Context, minus a blacklist (to avoid binding, e.g., function inputs).  I haven't yet wired the new conversion pass into the ExecutionEngine optimization pipeline, because I think we'll want to experiment with where it's called with respect to training and inference passes before baking it in.
*Testing*: A new test in graphOptz, plus a second commit that converts many of the tests that require constants to use the new pass.
*Documentation*: None, yet.
Fixes #1623

### DIFF
--- a/lib/Graph/Node.cpp
+++ b/lib/Graph/Node.cpp
@@ -42,7 +42,7 @@ NodeValue::NodeValue(Node *N, unsigned resNo) {
   resNo_ = resNo;
 }
 
-void NodeValue::replaceAllUsesOfWith(NodeValue v) {
+void NodeValue::replaceAllUsesOfWith(NodeValue v, const Function *F) {
   if (v.getNode()) {
     assert(getType() == v.getType() && "Replacing value with the wrong type");
   }
@@ -53,6 +53,10 @@ void NodeValue::replaceAllUsesOfWith(NodeValue v) {
                                          nodeValueUsers.end());
   for (auto &U : usersVec) {
     NodeHandle *site = U.get();
+    auto *userF = U.getUser()->getParent();
+    // If the user is not in function F, don't touch it.
+    if (F && userF != F)
+      continue;
     assert(site->getNode() == node_ && "Invalid user");
     assert(site->getResNo() == getResNo() && "Invalid list of uses");
     site->setOperand(v.getNode(), v.getResNo());
--- a/tests/unittests/graphOptzTest.cpp
+++ b/tests/unittests/graphOptzTest.cpp
@@ -232,15 +232,16 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButVarReused) {
 }
 
 TEST_F(GraphOptz, transposePrivateVariable) {
-  Variable *A = mod_.createVariable(ElemKind::FloatTy, {1, 10, 20, 3}, "A",
-                                    VisibilityKind::Private, false);
-  A->getHandle().randomize(-7.0, 12.0, mod_.getPRNG());
+  auto *A =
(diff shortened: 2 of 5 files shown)


bertmaher:
@nadavrot Pointed out that running this optimization will make a training graph invalid (there will be SaveNodes writing into constants, which is bad).  Rather than simply hope this never happens, I'm going to modify the pass to only affect a specific Function, rather than the entire Module.

tests/unittests/graphOptzTest.cpp
```diff
@@ -1922,3 +1922,29 @@ TEST_F(GraphOptz, ReshapePrivateVarOneUse) {
...
+  auto *save3 = F_->createSave(ctx_, "save3", input3);
+
+  // No variables, six PHs (3 inputs, 3 saves).
+  EXPECT_EQ(mod_.getVars().size(), 0);
+  EXPECT_EQ(mod_.getPlaceholders().size(), 6);
+
+  // Allocate two of the three inputs, but mark input2 of them as non-constant.
+  ctx_.allocate(input1);
+  ctx_.allocate(input2);
+  ::glow::convertPlaceholdersToConstants(mod_, ctx_, {input2});
```
beicy: Why is "input3" not allocated here?

nadavrot: I think that the intension is to verify that unallocated placeholders don't crash the procedure. It would probably be a good idea to add a comment.

jfix71: The way I was reading this was that unallocated placeholders are just not considered to be constants, as they have no tensor backing them. 

Do we have some intended usage here for when placeholders should be allocated? It seems we have three options -- (1) they are all allocated before calling `compile()`, (2) they are all allocated after calling `compile()`, or (3) they can be allocated before or after `compile()`. The blacklist is necessary for (1) and (3). I was reading this test case as more like (3).

bertmaher: @jfix71 Yeah, the testcase is intended to exercise all the options; (1) allocated tensor/placeholder replaced by constant, (2) unallocated placeholder not replaced, and (3) allocated but blacklisted.

lib/Graph/Node.cpp
```diff
@@ -53,6 +53,9 @@ void NodeValue::replaceAllUsesOfWith(NodeValue v) {
                                          nodeValueUsers.end());
   for (auto &U : usersVec) {
     NodeHandle *site = U.get();
+    auto *userF = U.getUser()->getParent();
```
nadavrot: Could you please add a comment that says something like:  "If the user is not in function F then don't touch it".

nadavrot:
Nice!  This turned out to be very elegant.

PR #1712, opened 2018-09-25
[Placeholder] Update tests that were creating variables that were saved into.

nadavrot:
*Description*: [Placeholder] Update tests that were creating variables that were saved into.
*Testing*: This updates tests. I found this issue by adding assert false into the save node that created variables. 
*Documentation*: none.

#1334.


nadavrot:
Earlier today @rdzhabarov landed hundreds of changes that fixed all of the tests in operatorTest. I caught the remaining two issues by placing an assert in the SaveNode that created variables.

PR #1715, opened 2018-09-25
[fix] Fix double registration when loaded by onnxifi_load

rdzhabarov:
*Description*:
Fix libprotobuf.so issue when loading libonnxifi-glow through onnxifi_load.

*Testing*:
Running ./run_sh (in progress) and onnxifi_load by zrphercule (for now manually)
*Documentation*:
N/A

cc: @zrphercule @yinghai @bddppq


rdzhabarov:
That works, will update onnx submodule.

PR #1716, opened 2018-09-25
[examples] Convert PH->constants in mnist

bertmaher:
*Description*: After our recent placeholder conversions, our examples no longer use constant weights during inference.  Let's try out the new constant-conversion opt, starting with mnist.
*Testing*: `./bin/mnist`

### DIFF
--- a/examples/mnist.cpp
+++ b/examples/mnist.cpp
@@ -158,6 +158,8 @@ void testMNIST() {
     timer.stopTimer();
   }
   llvm::outs() << "Validating.\n";
+
+  ::glow::convertPlaceholdersToConstants(F, ctx, {A, result->getPlaceholder()});
   EE.compile(CompilationMode::Infer, F, ctx);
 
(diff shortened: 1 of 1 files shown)


bertmaher:
This is an example of how we can convert placeholders to constant weights after training.

What does everyone think?  It makes me a little uneasy to have to explicitly blacklist the input and output variables, but that kind of makes sense given the way a context is supposed to be used.

I'd also hoped to see an obvious way to move constant-promotion into the standard optimization pipeline, but given the need for a blacklist I feel like it might make the `::optimize()` API too clunky.

nadavrot:
@bertmaher This PR looks good to me. Our C++ API is not as nice as the PyTorch user interface, but I think that it's okay, because Glow is an execution engine, not a tool to be used by people. We could annotate the Placeholders with a special flag, like we do with the "trainable". I am not sure if it's a good idea.

nadavrot:
@bertmaher Thank you for doing this work and for presenting this example to everybody.

PR #1720, opened 2018-09-26
[Tensor] Disallow GetHandle() when Tensor is an rvalue object

artemrakhov:
*Description*:
Recently @beicy debugged an issue where doing `auto H = T.extractSlice(i).getHandle();` resulted in corrupted Handle. This happened because `T.extractSlice(i)` was creating a temporary Tensor object, which was destroyed on the same line. So Handle outlived Tensor it was pointing at. Such errors could be caught at compilation stage.

*Testing*:
Make sure that `auto TT = Tensor(424.).getHandle();` does not compile.

Let me know what you think. I don't see `&&` specifier used in this way in our codebase. But maybe we should add it everywhere else, e.g. `getUnsafePtr()`, `getUnowned()`.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -381,7 +381,10 @@ public:
   char *getUnsafePtr() const { return getData(); }
 
   /// \return a new handle that points and manages this tensor.
-  template <class ElemTy = float> Handle<ElemTy> getHandle();
+  template <class ElemTy = float> Handle<ElemTy> getHandle() &;
+
+  /// If Tensor is rvalue, it is an error to get its Handle.
+  template <class ElemTy = float> Handle<ElemTy> getHandle() && = delete;
 
 private:
   /// \returns a pointer to the raw data, of type \p ElemTy.
@@ -779,7 +782,7 @@ private:
   }
 };
 
-template <class ElemTy> Handle<ElemTy> Tensor::getHandle() {
+template <class ElemTy> Handle<ElemTy> Tensor::getHandle() & {
   assert(type_.isType<ElemTy>() && "Getting a handle to the wrong type.");
   return Handle<ElemTy>(this);
 }


bertmaher:
Wow, I've never seen ref qualifiers before.  Very interesting!

nadavrot:
@artemrakhov Thank you. This is great. I learned something new.

opti-mix:
Very cool!

But we have more edge cases, which are not covered by this fix. Consider the following snippet:

```c++
TEST(Tensor, getHandle) {
  Tensor longLivedTensor(ElemKind::FloatTy, {100});
  Handle<float> handle = longLivedTensor.getHandle(); 
  {
    Tensor shortLivedTensor(ElemKind::FloatTy, {10});
    // handle now refers to the local tensor object that will be deallocated at the end of this nested scope.
    handle = shortLivedTensor.getHandle();
  }
  // handle contains a dangling reference now.
  // If we try to dereference it, we get a crash. 
  handle.at({5}) = 10.0;
}
```

So, we need more work to avoid handles who outlive the tensors they refer to.

nadavrot:
@qcolombet Handles solve the problem of moving the dynamic dispatch parts when accessing the tensor and pre-calculate the address calculations for accessing the tensor. A naive x.at() access to tensor would require calculating a sequence of multiplications for the sizes of the tensor and the coordinates. Handles pre-calculate the size-integral, so accessing the tensor becomes a simpler operation. In practice, our handle-based convolutions in the interpreter are vectorized, because the handle moves all of the dynamic calculations to the beginning of the function. This is possible because the handle moves the address calculation to the stack, so the compiler knows that the sizes/dims don't alias with the tensor data or any other pointer or opaque call. Accessing x.at() would require that we load the dimensions, perform the address calculation and type check (a switch!) on each access to an element.

qcolombet:
@nadavrot We could encapsulate the "complex" accesses in tensor. I get the argument for efficientcy in the interpreter code. The question would then be, does any other component benefit from it? Put differently if we have this design to speed up the interpreter, maybe it should be in the interpreter.

Anyhow, I don't have a strong opinion either way, I am just pointing out that handles are tightly coupled with tensors.

nadavrot:
@qcolombet The optimizer (the optimizations that modify weights) and the unit tests also use handles.   How would you encapsulate the complex accesses in the tensor? If you don't extract the dimensions to the stack outside of the loop that iterate over tensors then you'll need to re-load them (and perform dynamic dispatch) before each access to each element in the tensor.

qcolombet:
> The optimizer (the optimizations that modify weights) and the unit tests also use handles. 

Sure. But they could equally use the tensor directly if we had a ::at method there.

> How would you encapsulate the complex accesses in the tensor?

With a handle underneath :).

>  If you don't extract the dimensions to the stack outside of the loop that iterate over tensors then you'll need to re-load them (and perform dynamic dispatch) before each access to each element in the tensor.

Two things; does this really matter aside from the interpreter? Couldn't we achieve the same thing without exposing the handle?

For the former I don't know, I would expect the relative time spent on optimizing weights is negligible but I could be wrong. For the later, it seems to me, modulo some memory overhead, that the handle could totally be kept within the tensor while providing fast accesses (assuming we need them generally).

Anyhow, let me be concrete about what I don't like about the handle thing:
```
Tensor someTensor(ElemKind::SomeType, dims);
Handle someHandle = someTensor.getHandle<underlyingType>(); // <-- the problem
```

The problem here is that:
1. We need to know what is the type of `someTensor` to get a proper handle to it
2. We need to know the underlying type of ElemKind::SomeType (e.g., FloatTy goes with float, Float16Ty goes with float16_t, and so on)

That means that every time we add a new type we have to make sure the right type dispatch (`switch(ElemKind) {case FloatTy: /*do something with float*/; case Float16Ty: /*do the same thing with Float16Ty*/; }`) happens in every places that works with handles. E.g., let say we want to add a double type, now we have to patch the optimization that merge BN(Conv), so that the handles use the proper types.

Ideally, we shouldn't care about the underlying type. E.g., tensors could deal with a high level data type that allows to do multiple, add, etc., whatever is used in the optimizations/compiler. Then adding a type, only touches the Type interface and doesn't leak everywhere. The down side is that the tensor would need to be lowered before being transmitted for execution on the device (e.g., from big array of a "class data", to whatever is the representation of that data on the device).

Again, I am not saying this should be fixed, it could be the right thing to do. In particular, we don't add types that often. Merely pointing out what annoys me :).

artemrakhov:
@bertmaher : good point about `T.extractSlice().getHandle().at({0})`. This particular expression is not used, because one can avoid doing `extractSlice()` and it's a heavy operation. But maybe there will be some examples where it makes sense to do one-liner. Let's remember that we disallowed this, so in the future we can allow it back if anything.

@opti-mix : agree, there are other possibilities of Tensor going out of scope. But what is proper solution then? We could disallow creation of Tensors on stack, and change pointer inside Handle to be smart pointer.

PR #1731, opened 2018-09-27
[importer]  add Clip operator for caffe2 model importer

### LINKED ISSUE #1706: [New operator] Clip
We need to support `Clip` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#clip

yzhao30:
*Description*: add new operator "Clip" in caffe2 model importer
also add unittest for the new operator
*Testing*: ninja all && ninja test
*Documentation*:
[Optional Fixes #issue] resolve #1706

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -558,6 +558,9 @@ public:
   Node *createWeightedSum(llvm::StringRef name, llvm::ArrayRef<NodeValue> data,
                           llvm::ArrayRef<NodeValue> weights);
 
+  /// Create a series of nodes for the Clip operator. It limits the given input
+  /// within an interval specified by the `min` and `max` arguments.
+  Node *createClip(llvm::StringRef name, NodeValue input, float min, float max);
   /// @}
 
   /// @name The builder functions below are identical to the builder functions
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1491,6 +1491,15 @@ Node *Function::createWeightedSum(llvm::StringRef name,
   return currAdd;
 }
 
+Node *Function::createClip(llvm::StringRef name, NodeValue input, float min,
+                           float max) {
+  auto *minSplat = createSplat(name.str() + ".minSplat", input.getType(), min);
+  auto *minClipped = createMax(name.str() + ".minClip", input, minSplat);
+  auto *maxSplat = createSplat(name.str() + ".maxSplat", input.getType(), max);
+  auto result = createMin(name.str(), minClipped, maxSplat);
+  return result;
+}
+
 //===----------------------------------------------------------------------===//
 //                   Placeholder-builder methods.
 //===----------------------------------------------------------------------===//
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -562,6 +562,20 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Clip") {
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    float cmin = (dict.count("min") && dict["min"]->has_f())
+                    ? loadFloat(dict["min"])
+                    : std::numeric_limits<float>::lowest();
+    float cmax = (dict.count("max") && dict["max"]->has_f())
+                    ? loadFloat(dict["max"])
+                    : std::numeric_limits<float>::max();
+
+    auto *node = G_.createClip(opName, in, cmin, cmax);
+    addNodeAsOutput(op, node);
+    return;
+  }
+
   unexpectedNodeError(op, "Unsupported operator.");
 }
 
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -348,3 +348,84 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
   // We don't actually check that the output is correct, because this
   // should be covered in the OperatorTest for MatMul already.
 }
+
+/// Test loading clip op from a Caffe2 model.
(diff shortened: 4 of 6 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1541,6 +1541,32 @@ Node *Function::createWeightedSum(llvm::StringRef name,
...
+Node *Function::createClip(llvm::StringRef name, NodeValue input, float min,
+                           float max) {
+  Node *minClipped, *result;
+  if (min != std::numeric_limits<float>::lowest()) {
+    auto *minSplat =
+        createSplat(name.str() + ".minSplat", input.getType(), min);
+    // minClipped = (input <= min) ? min : input
+    auto *LTEMinCond = createCmpLTE(name.str() + ".lteMin", input, minSplat);
+    minClipped =
+        createSelect(name.str() + ".clipMin", LTEMinCond, minSplat, input);
```
bertmaher: Is it possible to use `Min` rather than `CmpLTE+Select`?

include/glow/Graph/Graph.h
```diff
@@ -565,6 +565,12 @@ class Function final : public Named {
   Node *createWeightedSum(llvm::StringRef name, llvm::ArrayRef<NodeValue> data,
                           llvm::ArrayRef<NodeValue> weights);
 
+  /// Create a series of nodes for the Clip operator. It limits the given input 
+  /// within an interval specified by the `min` and `max` arguments. They default
```
jfix71: These default values are based on the Caffe2 operators. However this node creator is independent of Caffe2, i.e. it could be used by the ONNX loader or elsewhere. So I think either this comment should be removed, or the you should add the defaults for the function's arguments.

lib/Importer/Caffe2.cpp
```diff
@@ -562,6 +562,21 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Clip") {
+    // Load the inputs:
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    auto cmin = (dict.count("min") && dict["min"]->has_f())
```
jfix71: nit: I know we aren't consistent here, but I think we should make a push to use actual types when it's not too burdensome. So please use `float` here and below.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -348,3 +348,91 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
   // We don't actually check that the output is correct, because this
   // should be covered in the OperatorTest for MatMul already.
 }
+
+/// Test loading clip op from a Caffe2 model.
```
jfix71: nit: Can you specify either here or elsewhere what the min/max values of the Clip are, so we wouldn't need to go to look at the pbtxt to find out?

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -348,3 +348,91 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
...
+  Tensor inputs_0(ElemKind::FloatTy, {5, 5});
+  std::vector<float> testValues = {45.0, 16.0, 59.0, 99.0, 48.0, 12.0, 44.0,
+                                   46.0, 82.0, 28.0, 1.0,  91.0, 18.0, 9.0,
+                                   71.0, 24.0, 37.0, 61.0, 12.0, 81.0, 36.0,
+                                   38.0, 30.0, 84.0, 40.0};
+
+  for (size_t i = 0; i < 5 * 5; i++) {
+    inputs_0.getHandle().raw(i) = testValues[i];
+  }
+
```
beicy: Since `std::vector<float> testValues` is only used for initialization, it can be removed and write : inputs_0.getHandle<>() = {...};

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -348,3 +348,91 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
...
+                                       60.0, 24.0, 37.0, 60.0, 20.0, 60.0, 36.0,
+                                       38.0, 30.0, 60.0, 40.0};
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
+  for (size_t i = 0; i < 5 * 5; i++) {
+    EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
+  }
+}
+
+/// Test loading clip op from a Caffe2 model. When the args are not set, the
+/// default values are used.
```
beicy: Maybe explain a little bit about "default value" here, like  numeric_limits::lowest() and numeric_limits::max().

lib/Graph/Graph.cpp
yzhao30: Yes, it is possible. Code updated.

lib/Importer/Caffe2.cpp
yzhao30: Can you explain a little bit more? Not clear which part? You mean change `auto` to `float`?

jfix71: Yeah sorry, please use float instead of auto

include/glow/Graph/Graph.h
```diff
@@ -576,6 +576,11 @@ class Function final : public Named {
 
   /// @}
 
+  /// Create a series of nodes for the Clip operator. It limits the given input 
+  /// within an interval specified by the `min` and `max` arguments.  
+  Node *createClip(llvm::StringRef name, NodeValue input, float min,
```
rdzhabarov: please also move this up to be included into doxygen section of
```
/// @name High-level, operation-level IRBuilder.
   ///@{
```

rdzhabarov:
Oh, did not realize you could merge the PR.

For the future, please:
* make a better description for a commit message before merging
* do not rollback changes in submodules (unless needed). This commit reverted previous submodule update for onnx.

For (2) I'll push update again.

PR #1733, opened 2018-09-27
[Caffe2] Implement ReplaceNaN operator

### LINKED ISSUE #1704: [New operator] ReplaceNaN
We need to support `ReplaceNaN` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#replacenan

SplitInfinity:
**Description**
This commit fixes #1704 and adds support for the `ReplaceNaN` Caffe2 operator to the
Caffe2 model importer. This operator accepts one tensor as an input and
a value as an argument, and replaces all elements of the input that are
`NaN` with the given value. To implement this operator, this commit adds a new
Glow node, `IsNaN`, which accepts an input tensor and outputs a tensor in
which each element is a boolean indicating whether or not the
corresponding element in the input is `NaN`. With this new node in
hand, a `ReplaceNaN` operator is implemented as a `Select` node with the
`ReplaceNaN` input as one of the inputs, a `Splat` node created using the replacement value
as the other input, and the output of an `IsNaN` node as the comparator
input. This `IsNaN` node becomes a new instruction called `elementisnan`,
which is implemented using `std::isnan` from the `cmath` header and `isnan`
from the `math.h` header in the interpreter and CPU backends,
respectively.

**Testing**
This commit adds a new test case to `caffe2ImporterTest.cpp` which imports
and runs a Caffe2 model consisting of only the `ReplaceNaN` operator. This
test performs high level checks on the Glow node graph produced by the
importer, and checks that all instances of `NaN` in a test input tensor
are replaced with the replacement value provided in the model protobuf
and non-`NaN` elements have been left unchanged.

This commit also adds a new test case to `OperatorTest.cpp` for `createReplaceNaN` to test that
`NaN` values are replaced correctly by the interpreter and CPU backends.

All other unit tests still pass.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -413,6 +413,18 @@ public:
   ARITHMETIC_FUN_DECL(Pow);
 #undef ARITHMETIC_FUN_DECL
 
+  /// Create a node that produces an boolean output of the same shape as
+  /// \p input in which each element indicates whether or not the corresponding
+  /// element in \p input is NaN or not.
+  IsNaNNode *createIsNaN(llvm::StringRef name, NodeValue input);
+
+  /// Implements an operation that replaces all instances of NaN in \p input
+  /// with \p value. This operation is lowered to a Select node with \p input
+  /// as one of the inputs, a Splat node created using \p value as the other
+  /// input, and an IsNaN node as the comparator input.
+  /// \returns the Select node.
+  Node *createReplaceNaN(llvm::StringRef name, NodeValue input, float value);
+
   PowNode *createPow(llvm::StringRef name, NodeValue base, float exp);
 
   SelectNode *createSelect(llvm::StringRef name, NodeValue Cond, NodeValue LHS,
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1075,6 +1075,25 @@ CmpEQNode *Function::createCmpEQ(llvm::StringRef name, NodeValue LHS,
   return addNode(new CmpEQNode(name, OT, LHS, RHS));
 }
 
+IsNaNNode *Function::createIsNaN(llvm::StringRef name, NodeValue input) {
+  TypeRef OT = getResultTypeOfBooleanOp(*getParent(), input.getType());
+  return addNode(new IsNaNNode(name, OT, input));
+}
+
+Node *Function::createReplaceNaN(llvm::StringRef name, NodeValue input,
+                                 float value) {
+  // Create IsNaN node.
+  auto *INN = createIsNaN(name.str() + ".isNaN", input);
+
+  // Create Splat node.
+  auto *S = createSplat(name.str() + ".splat", input.getType(), value);
+
+  // Create Select node to pick between original and replacement values.
+  auto *SN = createSelect(name.str() + ".select", INN, S, input);
+
+  return SN;
+}
+
 PowNode *Function::createPow(llvm::StringRef name, NodeValue base, float exp) {
   auto *SP = createSplat(name, base.getType(), exp);
   return createPow(name, base, SP);
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -328,6 +328,19 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "ReplaceNaN") {
+    // Load the input and NaN replacement value:
+    auto input = getNodeValueOrCreateVariableByName(op.input(0));
+    auto valueIt = dict.find("value");
+    auto value = valueIt != dict.end() ? loadFloat(valueIt->second) : 0.0f;
+
+    auto *node = G_.createReplaceNaN(opName, input, value);
+
+    // Save the outputs:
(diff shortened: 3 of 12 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1155,6 +1155,30 @@ CmpEQNode *Function::createCmpEQ(llvm::StringRef name, NodeValue LHS,
...
+  return addNode(new IsNaNNode(name, OT, input));
+}
+
+Node *Function::createReplaceNaN(llvm::StringRef name, NodeValue input,
+                                 float value) {
+  // Create IsNaN node.
+  auto *INN = createIsNaN(name.str() + ".isNaN", input);
+
+  // Create tensor of the same size as input to pass to a Select node.
+  auto *valueVar = getParent()->createVariable(
```
SplitInfinity: What should I do here given that we're in the middle of migrating everything to placeholders? This is a constant.

include/glow/Graph/Graph.h
```diff
@@ -429,6 +429,10 @@ class Function final : public Named {
   ARITHMETIC_FUN_DECL(Pow);
 #undef ARITHMETIC_FUN_DECL
 
+  IsNaNNode *createIsNaN(llvm::StringRef name, NodeValue input);
```
jfix71: Doxygen comments please!

include/glow/Graph/Graph.h
```diff
@@ -429,6 +429,10 @@ class Function final : public Named {
...
+
+  Node *createReplaceNaN(llvm::StringRef name, NodeValue input, float value);
```
jfix71: Doxygen comments please!

lib/Graph/Graph.cpp
jfix71: I think this is fine for now -- we will eventually move all private Variables to Constants, and we don't yet have Constant.

tools/ClassGen/NodeGen.cpp
```diff
@@ -315,6 +315,9 @@ int main(int argc, char **argv) {
                     "Weights[0] * Slice(0) + Weights[1] * Slice(1) + ... "
                     "It implies that len(Weights) == len(Indices).");
 
+  BB.newNode("IsNaN").addInput("Input").addResultFromCtorArg().setDocstring(
```
jfix71: nit: please use `// clang-format off` and `// clang-format on` here -- see the `Log` node also in this file

lib/Importer/Caffe2.cpp
```diff
@@ -328,6 +328,18 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "ReplaceNaN") {
+    // Load the input and NaN replacement value:
+    auto input = getNodeValueOrCreateVariableByName(op.input(0));
+    auto value = loadFloat(dict["value"]);
```
rdzhabarov: "value" is an optional param. I'd explicitly check if it's present and if not assign 0 to value.

tools/ClassGen/InstrGen.cpp
```diff
@@ -300,6 +300,17 @@ int main(int argc, char **argv) {
...
 
+  BB.newInstr("ElementIsNaN")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("Src", OperandKind::In)
+      .inplaceOperand({
+          "Dest",
+          "Src",
+      })
+      .dataParallel()
+      .autoVerify(VerifyKind::SameType, {"Dest", "Src"})
```
rdzhabarov: dest is bool mask and input is tensor, why sameType? (maybe i'm missing something)

tests/models/caffe2Models/replace_nan_predict_net.pbtxt
```diff
@@ -0,0 +1,14 @@
+name: "replaceNaN"
+op {
+  input: "input"
+  output: "replace_nan_result"
+  name: ""
+  type: "ReplaceNaN"
+  arg {
+    name: "value"
+    i: 1
```
rdzhabarov: why both 'i' and 'f'?

lib/Importer/Caffe2.cpp
SplitInfinity: Won't `operator[]` do that for me?

rdzhabarov: it will if it's an std::map, we have typedef for type of `dict`. It's kinda fragile.

SplitInfinity: I'm still going to change it but just for everyone's info `dict` is a `std::unordered_map` and `operator[]` appears to have the same semantics: http://www.cplusplus.com/reference/unordered_map/unordered_map/operator[]/

lib/Graph/Graph.cpp
```diff
@@ -1075,6 +1075,30 @@ CmpEQNode *Function::createCmpEQ(llvm::StringRef name, NodeValue LHS,
...
+      input.getType(), name.str() + ".value", VisibilityKind::Private,
+      /* isTrainable = */ false);
+
+  // Fill tensor with 'value'.
```
nadavrot: @SplitInfinity @jfix71 Could we use a splat node here?

Issue #1595, opened 2018-09-07
Why Caffe2 Importer only supports inference currently?


wayneshawn:
Since we can construct NN with Glow C++ API like examples/mnist.cpp.
why cannot we just import Caffe2 Model and differentiate F and retrain? 
what's the main difference between Function built from model and from API ? (e.g. Tensor initialization, diferentiate).
thanks a lot~

jfix71:
There's no reason this couldn't be done in theory, importing the network structure from a proto to differentiate and train. The C++ API for building the graph is used for mnist, unittests/MLTest.cpp, the proto importers, etc. But currently our proto importers are built expecting pre-trained models, e.g. variables that are created during loading have the `trainable` property set to `false`. So you would need to modify the importer at least allowing for that, and I'm sure you would encounter other small obstacles. I don't expect it would be incredibly difficult overall, though.

wayneshawn:
@jfix71 thanks for your reply.
p.s. 
1.glow(NHWC) will add transpose node for Caffe2 input data(NCHW), and transpose back. there is something wrong when differentiate loaded graph directly, perhaps with some operator during lowering and optimization.
current initialization for model  tensor is GivenTensorFill, ConstantFill and UniformFill
but xaiver initializtion is used when creaeConv and createFullyConnected with C++ API.
that seems to be the main difference.

nadavrot:
@wayneshawn @jfix71 We plan to rewrite the Graph builder methods and make them more general (by separating the context). I hope that this rewrite will help to fix the generality problem that you've described (where the wrong initialization kind is used), but I suspect that we'll need to iterate over it a few times before we get it to work. Could you please open an issue with an accurate description of the problem so that we could track this and fix it?

PR #1739, opened 2018-09-28
[Placeholders] Migrate batchnorm

bertmaher:

### DIFF
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -247,13 +247,14 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     }
 
     auto channel = getChannel(dict);
-    auto *node = G_.createBatchNormalization(opName, in, channel, epsilon);
-
-    // Load the weights.
-    cast<Variable>(node->getScale())->assign(scale);
-    cast<Variable>(node->getBias())->assign(bias);
-    cast<Variable>(node->getMean())->assign(mean);
-    cast<Variable>(node->getVar())->assign(var);
+    auto *scaleV = G_.getParent()->createVariable("scale", *scale);
+    auto *biasV = G_.getParent()->createVariable("bias", *bias);
+    auto *meanV = G_.getParent()->createVariable(
+        "mean", *mean, VisibilityKind::Private, false);
+    auto *varV = G_.getParent()->createVariable("var", *var,
+                                                VisibilityKind::Private, false);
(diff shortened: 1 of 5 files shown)


lib/Importer/Caffe2.cpp
```diff
@@ -247,13 +247,14 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
 
     auto channel = getChannel(dict);
-    auto *node = G_.createBatchNormalization(opName, in, channel, epsilon);
-
-    // Load the weights.
-    cast<Variable>(node->getScale())->assign(scale);
-    cast<Variable>(node->getBias())->assign(bias);
-    cast<Variable>(node->getMean())->assign(mean);
-    cast<Variable>(node->getVar())->assign(var);
+    auto *scaleV = G_.getParent()->createVariable("scale", *scale);
```
rdzhabarov: what is the plan for this?
(I thought it's to make these placeholders and then covertPlaceholderToConstants when appropriate)

bertmaher: So I'm actually kind of unsure here.  We don't use Context at all in the loader right now, and plumbing it through to get at the placeholder versions of this (and FC and Conv) seems like kind of a lot of work.  So I'm conflicted.

rdzhabarov: ok, we can take care of this at later point.

rdzhabarov:
You need to rebase on the latest master to get rid of the failure (there was a PR merge with failed tests, which is fixed now)

lib/Importer/Caffe2.cpp
nadavrot: I think that it makes sense to create constants (variables) for all of the weights that we load, and Placeholder for all of the uninitialized buffers (inputs/outputs). right?

jfix71: @nadavrot I think that makes sense if we are only going to use the proto loaders for pre-trained networks. If we eventually want to differentiate a graph that we load from a proto then I believe we would want to use Placeholders. We've received a question about this before (https://github.com/pytorch/glow/issues/1595).

bertmaher: @jfix71 Yes, that's a good point, if we want to train from protos we'll need to create Placeholders here.  I think we can probably defer that for now, though.

nadavrot: @bertmaher @jfix71 Ah, yes, I forgot about that. You are right.

PR #1740, opened 2018-09-28
[onnxifi] Add support to more operators for compatibility call.

rdzhabarov:
*Description*:
Add more operators to be checked in compatibility call.

*Testing*:
Manual for now. We are working on test driver support.

*Documentation*:
N/A

cc: @yinghai @bertmaher


nadavrot:
Nice! @rdzhabarov

PR #1741, opened 2018-09-28
Fix some lint

bertmaher:
*Description*: FB's internal linter flagged a few things.  I don't think any of the shadowing was causing real problems though, it's just nicer not to have it.
*Testing*: ninja check
*Documentation*: N/A

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -136,7 +136,7 @@ void glow::runBatch(ExecutionEngine &EE, Context &ctx, size_t iterations,
          "The number of inputs does not match the number of variables");
 
   // For each iteration in the batch:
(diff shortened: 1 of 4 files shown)


rdzhabarov:
no submodule update :)

bertmaher:
Good catch on the submodules, thank you!

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -136,7 +136,7 @@ void glow::runBatch(ExecutionEngine &EE, Context &ctx, size_t iterations,
          "The number of inputs does not match the number of variables");
 
   // For each iteration in the batch:
-  for (size_t i = 0; i < iterations; i++) {
+  for (size_t j = 0; j < iterations; j++) {
```
nadavrot: Oh, wow!  Good linter!

PR #1743, opened 2018-09-28
[onnxifi] Wrap onnxifi function names into macro.

rdzhabarov:
*Description*:
Wrap onnxifi function names into macro. Now those can be controlled externally.

*Testing*:
build + manual test.

*Documentation*:
n/a

cc: @Maratyszcza @yinghai

### DIFF
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -18,6 +18,11 @@
 
 #include "glow/Importer/ONNXIFILoader.h"
 
+/// Allow defining names for onnxifi implementation.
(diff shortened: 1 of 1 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -18,6 +18,11 @@
 
 #include "glow/Importer/ONNXIFILoader.h"
 
+/// Allow defining names for onnxifi implementation.
+#ifndef ONNXIFI_LIBRARY_FUNCTION_WRAPPER
```
Maratyszcza: I'd suggest to use a Glow-specific name, e.g. `GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER` to avoid name conflicts

PR #1744, opened 2018-09-28
[Quantization] Add row-wise quantized FC for CPU backend

### LINKED ISSUE #1402: Need to implement channel-wise quantized fullyconnected node.

beicy:
*Description*:
Add row-wise quantized FC for CPU backend.  

*Testing*:
 Enable the unittest `rowwiseQuantizedFCTest` with CPU backend.
*Documentation*:
[Optional Fixes #1402 ]

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -274,11 +274,11 @@ public:
   /// Create a row-wise quantized fully connected node. This node is only used
   /// in quantization. Args \p input and \p B are quantized in regular way, \p W
   /// is the constant weights and will be row-wise quantized during node
-  /// creation time.
+  /// creation time. The output is quantized in the regular way, and its type
+  /// \p outTy is a quantized type.
   RowwiseQuantizedFullyConnectedNode *
-  createRowwiseQuantizedFullyConnected(Context &ctx, llvm::StringRef name,
-                                       NodeValue input, Variable *W, Node *B,
-                                       TypeRef outTy);
+  createRowwiseQuantizedFullyConnected(llvm::StringRef name, NodeValue input,
+                                       Variable *W, Node *B, TypeRef outTy);
 
   /// Implement an operation that computes the row-wise dot product of its
   /// inputs. Consequently, \p X and \p Y must be either 1D or 2D tensors. This
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -324,6 +324,9 @@ llvm::Value *LLVMIRGen::emitValueAddress(llvm::IRBuilder<> &builder,
   case ElemKind::Int8QTy:
     T = llvm::Type::getInt8PtrTy(ctx_);
     break;
+  case ElemKind::Int32QTy:
+    T = llvm::Type::getInt32PtrTy(ctx_);
+    break;
   case ElemKind::Int64ITy:
     T = llvm::Type::getInt64PtrTy(ctx_);
     break;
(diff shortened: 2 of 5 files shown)


lib/Backends/CPU/Transforms.cpp
```diff
@@ -79,6 +81,68 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
       CN->getBias(), CN->getKernels(), CN->getStrides(), CN->getPads(), group));
 }
 
+/// Convert a RWQFC node to its CPU version.
```
bertmaher: Could you add a bit more explanation to this doxygen comment?  It's not immediately clear why we need a CPU-specific RowwiseQuantizedFC.  (I think I get it, but a bit of explanation would help :) ).

lib/Backends/CPU/Transforms.cpp
```diff
@@ -79,6 +81,68 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
...
+  auto outputPreH = outputPreV->getHandle<int32_t>();
+  auto outputPostH = outputPostV->getHandle<int32_t>();
+  auto outputScaleH = outputScaleV->getHandle<int32_t>();
+
+  for (size_t i = 0; i < rowNum; i++) {
+    // Calculate the scale of the values that come out of the matrix
+    // multiplication part of the calculation.
+    float matMulScale = inputScale * scalesH.raw(i);
+
+    // Calculate the sacling parameters for the bias and output.
```
bertmaher: sp: 'sacling' -> 'scaling'

lib/Backends/CPU/Transforms.cpp
```diff
@@ -126,6 +190,15 @@ bool CPUBackend::transformPostLowering(Function *F,
       }
     }
 
+    // Replace rowwise quantized FC with cpu-specific version.
+    if (auto *RWQFC = dyn_cast<RowwiseQuantizedFullyConnectedNode>(&node)) {
```
nadavrot: Why do we need to replace the generic node with the CPU-specific node?

beicy: Yes, we don't need to do it. I was thinking adding a new CPU-specific node could clearly show the params differences between the CPU kernel and general rowwise quantized FC. But I will update it.  Thanks!

bertmaher: Is this case different from CPUConvDKKC8?  It seems like the same sort of transformation to me.

beicy: To my point of view, DKKC8 is not always applied. The "regular" conv should be in cpu backend as well. But for this rowwise quantized FC,  only "one" version is in CPU backend. So there is no need to add another node. Could you please let us know if there is any other reason for "DKKC8" cpu specific node @nadavrot ? Thanks!

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1293,6 +1293,35 @@ void libjit_insert_tensor_i8(int8_t *tensor, int8_t *slice, size_t *offset,
...
+    const int8_t *biasW, const int32_t *weightsOffsets, const int32_t *biasPre,
+    const int32_t *biasPost, const int32_t *biasScale, const int32_t *outPre,
+    const int32_t *outPost, const int32_t *outScale, const size_t *outWdims,
+    const size_t *inWdims, const size_t *weightsWdims, const size_t *biasWdims,
+    size_t rowNum, int32_t outOffset, int32_t inOffset, int32_t biasOffset) {
+  size_t in_w = inWdims[1];
+  size_t out_h = outWdims[0];
+  size_t out_w = outWdims[1];
+
+  for (size_t i = 0; i < out_h; i++) {
```
nadavrot: Please add a few comments here that describe what's going on.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1261,6 +1264,154 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
+
+    createCall(builder, F,
+               {destPtr, srcPtr, weightsPtr, biasPtr, weightsOffsetsPtr,
+                biasPrePtr, biasPostPtr, biasScalePtr, outputPrePtr,
+                outputPostPtr, outputScalePtr, destDims, srcDims, weightsDims,
+                biasDims, row, destOffset, srcOffset, biasOffset});
+    break;
+  }
+
+    /*  case Kinded::Kind::CPURowwiseQuantizedFCInstKind: {
```
nadavrot: What is this comment?

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1261,6 +1264,154 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     break;
   }
 
+  case Kinded::Kind::RowwiseQuantizedFullyConnectedInstKind: {
+    auto *RWQFC = cast<RowwiseQuantizedFullyConnectedInst>(I);
```
nadavrot: More comments here would be useful.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1293,6 +1293,35 @@ void libjit_insert_tensor_i8(int8_t *tensor, int8_t *slice, size_t *offset,
                        numDimsTensor, numDimsSlice, offsetDim, count, axis);
 }
 
+void libjit_rowwise_quantized_fc_i8(
```
bertmaher: Consider moving this to `libjit_matmul.cpp` since FC is basically a matmul.

lib/Backends/CPU/LLVMIRGen.cpp
bertmaher: Remember to delete this big commented-out block! :-D

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1270,6 +1273,110 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
+    std::vector<llvm::Constant *> outputPreV(rowNum);
+    std::vector<llvm::Constant *> outputPostV(rowNum);
+    std::vector<llvm::Constant *> outputScaleV(rowNum);
+
+    for (size_t i = 0; i < rowNum; i++) {
+      // Calculate the scale of the values that come out of the matrix
+      // multiplication part of the calculation.
+      float matMulScale = inputScale * scalesH.raw(i);
+
+      // Calculate the sacling parameters for the bias and output.
```
rdzhabarov: typo

include/glow/Graph/Graph.h
```diff
@@ -276,9 +276,8 @@ class Function final : public Named {
   /// is the constant weights and will be row-wise quantized during node
   /// creation time.
   RowwiseQuantizedFullyConnectedNode *
-  createRowwiseQuantizedFullyConnected(Context &ctx, llvm::StringRef name,
-                                       NodeValue input, Variable *W, Node *B,
-                                       TypeRef outTy);
+  createRowwiseQuantizedFullyConnected(llvm::StringRef name, NodeValue input,
```
rdzhabarov: can you also mention in the doxygen the outtype expectations. That's it's quantized in a regular way etc.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -1270,6 +1273,110 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
   }
 
+  case Kinded::Kind::RowwiseQuantizedFullyConnectedInstKind: {
+    auto *RWQFC = cast<RowwiseQuantizedFullyConnectedInst>(I);
+    // Since we can't get the variable from a glow::Value directly,
+    // we need to traverse the var list and find the one matching the given
+    // Value.
+    Tensor scalesT;
+    auto *F_ = getIRFunction();
+    for (auto &v : F_->getGraph()->getParent()->getVars()) {
```
rdzhabarov: would this be a useful utility for other cases?

beicy: So far, I don't think it is a common case to get the variable from a glow::Value like this.  Actually, we can add the scales and offsets as "Member" instead of "addInput" or "addOperand" , and they will be passed as ArrayRef as how we pass the strides vector in Convolution. But it requires extra work on LLVMIRGen::emit*** when passing the param to the kernels. But I think in the current way, the code itself looks more clean :)

bertmaher:
Oops, sorry, I didn't realize my request-changes was still blocking!  (GitHub needs to be more like Phabricator!!)

PR #1745, opened 2018-09-28
[Importer] Report error if model contains non-2D pool

### LINKED ISSUE #1736: Assertion failed: (Index < Length && "Invalid index!") when testing glow using onnx
When I use onnx test driver to glowm it crashed on AveragePool test because of:
Assertion failed: (Index < Length && "Invalid index!"), function operator[]
This is a llvm related error, so I need to find which command caused this manually.

Finally I found out the crash is caused by this command in ONNX.cpp::loadOperator when dealing with averagepool:
auto *tr = G_.createTranspose(opName, in, NCHW2NHWC);

I wonder what is this G_  for (since I did not see any relative definition in ONNX.cpp), and why we are using NCHW2NHWC here?

bertmaher:
*Description*: We were crashing on a 3D average pool, because we expect to do a 4D transpose for NCHW to NHWC (rather than a 5D transpose).  Return false instead.
*Testing*: Death test for onnxModelLoader.  (Note that ONNXIFI will not die; this is the standalone ONNX model loader that dies when a model can't be loaded, which seems proper.)
*Documentation*: N/A
Fixes #1736

### DIFF
--- a/tests/unittests/onnxImporterTest.cpp
+++ b/tests/unittests/onnxImporterTest.cpp
@@ -78,3 +78,18 @@ TEST(onnx, importConv) {
   for (size_t i = 0; i < 4 * 4; i++)
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
 }
+
+TEST(onnx, importAveragePool3D) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetFilename("tests/models/onnxModels/averagePool3D.onnxtxt");
+
+  // Destroy the loader after the graph is loaded since the following execution
+  // will not depend on anyting from the loader.
+  {
+    Tensor data(ElemKind::FloatTy, {1, 3, 32, 32, 32});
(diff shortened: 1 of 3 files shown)


tests/unittests/onnxImporterTest.cpp
```diff
@@ -78,3 +78,18 @@ TEST(onnx, importConv) {
   for (size_t i = 0; i < 4 * 4; i++)
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
 }
+
+TEST(onnx, importAveragePool3D) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetFilename("tests/models/onnxModels/averagePool3D.onnxtxt");
```
yinghai: We have this file already?

bertmaher: Sigh, forgot to git add, thanks.

bertmaher:
@zrphercule I'm not sure... I can't tell if onnxGetBackendCompatibility actually tries to parse the op, or if it just checks the op type.  If it's just the op type, then yeah, this is suboptimal b/c we'll get all the way to initGraph and then reject it.  We won't crash, but we won't use glow either :-).

nadavrot:
Nice!

PR #1746, opened 2018-09-28
[Caffe2] Implement DotProduct operator

### LINKED ISSUE #1701: [New operator] DotProduct
We need to support `DotProduct` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#dotproduct

SplitInfinity:
**Description**
This commit fixes #1701 and adds support for the `DotProduct` Caffe2
operator to the Caffe2 model importer. This operator accepts two tensors
of the same shape and with at most 2 dimensions, and computes their
row-wise dot product. This dot product operator is immediately lowered
to a `Mul` node if the inputs are 1D, and to a `Mul` node followed by a
`BatchedReduceAdd` on `axis=1` if the inputs are 2D.

**Testing**
This commit adds two test cases to `caffe2ImporterTest.cpp`, one for 1D
inputs and the other for 2D inputs.

All other unit tests pass.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1605,6 +1605,26 @@ SaveNode *Function::createSave(Context &ctx, llvm::StringRef name,
   return addNode(new SaveNode(name, input, dest));
 }
 
+Node *Function::createDotProduct(llvm::StringRef name, NodeValue X,
+                                 NodeValue Y) {
+  auto XDimsSize = X.dims().size();
+  (void)XDimsSize;
+
+  assert(X.dims() == Y.dims() && "X and Y must have the same shape");
+  assert((XDimsSize == 1) || (XDimsSize == 2) && "X and Y must be 1D or 2D");
+
+  // Create Mul node.
+  auto *MN = createMul(name.str() + ".mul", X, Y);
+
+  // If X and Y are 1D, the BatchedReduceAdd node is not needed.
+  if (XDimsSize == 1) {
+    return MN;
+  }
+
+  // Create and return BatchedReduceAdd node.
+  return createBatchedReduceAdd(name.str() + ".bra", MN, 1);
+}
+
 void Function::createGRU(Context &ctx, llvm::StringRef namePrefix,
                          llvm::ArrayRef<Node *> inputs, unsigned batchSize,
                          unsigned hiddenSize, unsigned outputSize,
--- a/lib/Importer/Caffe2.cpp
+++ b/lib/Importer/Caffe2.cpp
@@ -329,6 +329,19 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "DotProduct") {
+    // Load the inputs.
+    auto X = getNodeValueOrCreateVariableByName(op.input(0));
+    auto Y = getNodeValueOrCreateVariableByName(op.input(1));
+
+    // Create dot product node.
+    auto *node = G_.createDotProduct(opName, X, Y);
+
+    // Save the output.
+    addNodeAsOutput(op, node);
+    return;
+  }
+
   if (typeName == "ReplaceNaN") {
     // Load the input and NaN replacement value:
     auto input = getNodeValueOrCreateVariableByName(op.input(0));
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -490,3 +490,153 @@ TEST(caffe2, replaceNaN) {
     }
   }
 }
+
+/// Test loading a DotProduct operator with 1D inputs.
+TEST(caffe2, dotProduct1D) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename(
+      "tests/models/caffe2Models/dot_product_predict_net.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  Variable *output;
+
+  // Input tensors.
+  const size_t kDataSize = 10;
+  Tensor X(ElemKind::FloatTy, {kDataSize});
+  Tensor Y(ElemKind::FloatTy, {kDataSize});
+  auto XH = X.getHandle();
+  auto YH = Y.getHandle();
+
+  // Fill inputs with random values.
+  XH.randomize(-3.0, 3.0, mod.getPRNG());
+  YH.randomize(-3.0, 3.0, mod.getPRNG());
+
+  // Compute expected output.
+  Tensor O(ElemKind::FloatTy, {kDataSize});
+  auto OH = O.getHandle();
+
+  for (size_t i = 0; i < kDataSize; ++i) {
+    OH.at({i}) = XH.at({i}) * YH.at({i});
+  }
+
+  // Destroy the loader after the graph is loaded since the following execution
+  // will not depend on anyting from the loader.
+  {
+    caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
+                               {&X, &Y}, *F);
+    output = caffe2LD.getSingleOutput();
+  }
+
+  auto result = output->getHandle();
+
+  // Check that the shape of the output matches that of the expected output.
+  EXPECT_TRUE(result.dims().vec() == OH.dims().vec());
+
+  // High level checks on the content of the graph.
+  // We have 1 Mul and 1 Output.
+  EXPECT_EQ(F->getNodes().size(), 2);
+
+  // Check that the graph has the expected shape (Mul -> Save),
+  // starting from the output.
+  auto *saveNode = getSaveNodeFromVariable(output);
+  auto *MN = llvm::dyn_cast<MulNode>(saveNode->getInput());
+  ASSERT_TRUE(MN);
+
+  // With have two inputs and one output.
+  EXPECT_EQ(mod.getVars().size(), 3);
+
+  // Compile and run the model.
+  Context ctx;
+  EE.compile(CompilationMode::Infer, F, ctx);
+  EE.run();
+
+  // Check that the output tensor is the same as the expected output.
+  for (size_t i = 0; i < result.size(); ++i) {
+    EXPECT_NEAR(result.raw(i), OH.raw(i), 0.00001);
+  }
+}
+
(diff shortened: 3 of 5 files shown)


SplitInfinity:
These semantics for dot product seem weird, but that's what I understood after reading the comments in the Caffe2 source code: https://github.com/pytorch/pytorch/blob/master/caffe2/operators/distance_op.cc#L506

lib/Importer/Caffe2.cpp
```diff
@@ -328,6 +328,19 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "DotProduct") {
+    // Load the inputs.
+    auto X = getNodeValueOrCreateVariableByName(op.input(0));
+    auto Y = getNodeValueOrCreateVariableByName(op.input(1));
+
+    // Create dot product node.
+    auto *node = G_.createDotProduct(opName, X, Y);
```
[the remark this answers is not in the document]
qcolombet: cc @artemrakhov, he may have a different opinion and we should follow his recommendation. I thought we were doing as much expansion as possible in the importer, but now I am not sure seeing createReplaceNaN for instance.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -490,3 +490,156 @@ TEST(caffe2, replaceNaN) {
...
+  EXPECT_EQ(mod.getVars().size(), 3);
+
+  // Compile and run the model.
+  Context ctx;
+  EE.compile(CompilationMode::Infer, F, ctx);
+  EE.run();
+
+  // Check that the output tensor is the same as the expected output.
+  for (size_t i = 0; i < result.size(); ++i) {
+    EXPECT_NEAR(result.raw(i), OH.raw(i), 0.00001);
```
SplitInfinity: My first PR failed Travis here because I had `EXPECT_FLOAT_EQ` here and the two operands differed by about 0.000001. But I don't think this is the right thing to do because 0.00001 seems like too much error.

lib/Importer/Caffe2.cpp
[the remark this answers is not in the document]
SplitInfinity: I created the PR for `ReplaceNaN` and Artem was not the one who approved it, so that may not be the pattern he recommends. Similarly, the PR for `Clip` was not approved by Artem either. It would be good to decide the proper way to do this so that we can codify it in documentation about how to add C2 and ONNX operators.

nadavrot:
Why are we not generating the canonical representation which is a matrix multiplication? See #1701

SplitInfinity:
@nadavrot 

The comments in the Caffe2 code for the `DotProduct` op are hard to interpret unless you write out the vectors/matrices, so I typeset it and included it below for future reference.

For 1D inputs, `DotProduct` is defined like so:

<a href="https://www.codecogs.com/eqnedit.php?latex=$Given&space;two&space;vectors$\\&space;\begin{align*}&space;X&space;:=&space;\begin{bmatrix}&space;x_0&space;\\&space;\vdots&space;\\&space;x_n&space;\end{bmatrix},&space;Y&space;:=&space;\begin{bmatrix}&space;y_0&space;\\&space;\vdots&space;\\&space;y_n&space;\end{bmatrix}&space;\end{align*}&space;\\&space;$their&space;dot&space;product&space;is&space;defined&space;as&space;\\&space;\begin{align*}&space;\text{DotProduct}(X,&space;Y)&space;=&space;\begin{bmatrix}&space;x_0&space;\cdot&space;y_0&space;\\&space;\vdots&space;\\&space;x_n&space;\cdot&space;y_n&space;\end{bmatrix}&space;\end{align*}" target="_blank"></a>

and for 2D inputs:

<a href="https://www.codecogs.com/eqnedit.php?latex=\text{Given&space;two&space;matrices&space;}&space;\\&space;\begin{align*}&space;X&space;:=&space;\begin{bmatrix}&space;x_0^0&space;&&space;x_1^0&space;&&space;\ldots&space;&&space;x_k^0&space;\\&space;x_0^1&space;&&space;x_1^1&space;&&space;\ldots&space;&&space;x_k^1&space;\\&space;\vdots&space;&&space;\vdots&space;&&space;\ddots&space;&&space;\vdots&space;\\&space;x_0^n&space;&&space;x_1^n&space;&&space;\ldots&space;&&space;x_k^n&space;\end{bmatrix}\\&space;\\&space;Y&space;:=&space;\begin{bmatrix}&space;y_0^0&space;&&space;y_1^0&space;&&space;\ldots&space;&&space;y_k^0&space;\\&space;y_0^1&space;&&space;y_1^1&space;&&space;\ldots&space;&&space;y_k^1&space;\\&space;\vdots&space;&&space;\vdots&space;&&space;\ddots&space;&&space;\vdots&space;\\&space;y_0^n&space;&&space;y_1^n&space;&&space;\ldots&space;&&space;y_k^n&space;\end{bmatrix}&space;\end{align*}\\&space;$Their&space;dot&space;product&space;is&space;defined&space;as:$&space;\\&space;\begin{align*}&space;\text{DotProduct}(X,&space;Y)&space;=&space;\begin{bmatrix}&space;\sum_{i=0}^k&space;x_i^0&space;\cdot&space;y_i^0&space;\\&space;\vdots&space;\\&space;\sum_{i=0}^k&space;x_i^n&space;\cdot&space;y_i^n&space;\end{bmatrix}&space;\end{align*}" target="_blank"></a>

This is why it makes sense to me to lower these operations into elementwise `Mul` and `BatchReducedAdd`. Expressing these using matrix multiplication requires predication and `Diag` and `Trace` operators:

<a href="https://www.codecogs.com/eqnedit.php?latex=\begin{align*}&space;\text{DotProduct1D}(X,&space;Y)&space;=&space;\begin{bmatrix}&space;x_0&space;\cdot&space;y_0&space;\\&space;\vdots&space;\\&space;x_n&space;\cdot&space;y_n&space;\end{bmatrix}&space;=&space;\text{Trace(}XY^T)\\&space;\text{DotProduct2D}(X,&space;Y)&space;=&space;\begin{bmatrix}&space;\sum_{i=0}^k&space;x_i^0&space;\cdot&space;y_i^0&space;\\&space;\vdots&space;\\&space;\sum_{i=0}^k&space;x_i^n&space;\cdot&space;y_i^n&space;\end{bmatrix}&space;=&space;\text{Diag(}XY^T)&space;\end{align*}" target="_blank"></a>

nadavrot:
@SplitInfinity Thank you for taking the time to explain this to me. I did not think about non-1d tensors. Thank you. Your approach makes sense to me.

lib/Graph/Graph.cpp
```diff
@@ -1604,6 +1604,28 @@ SaveNode *Function::createSave(Context &ctx, llvm::StringRef name,
   return addNode(new SaveNode(name, input, dest));
 }
 
+Node *Function::createDotProduct(llvm::StringRef name, NodeValue X,
+                                 NodeValue Y) {
+  auto XDimsSize = X.dims().size();
+  auto YDimsSize = Y.dims().size();
+
+  // X and Y must have the same shape.
```
artemrakhov: This comment needs to be inside assert error message:
assert(XDimsSize == YDimsSize && "X and Y must have the same shape.");

lib/Graph/Graph.cpp
```diff
@@ -1604,6 +1604,28 @@ SaveNode *Function::createSave(Context &ctx, llvm::StringRef name,
...
+  assert(XDimsSize == YDimsSize);
```
artemrakhov: To have even better check, we should expect X.dims() == Y.dims()

lib/Importer/Caffe2.cpp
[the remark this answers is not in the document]
artemrakhov: I prefer to think about this in the following way: would you use this operator yourself if you were to create a Graph for toy example? I.e. is `createDotProduct()` alias useful to have in C++ API too, not only in loader(s)?

In my opinion, it is helpful. Probably I'd not use it for 1D inputs, but could use it for 2D tensors.

tests/unittests/caffe2ImporterTest.cpp
artemrakhov: Usually errors of 1E-4 are considered small enough and can come from fast-math optimizations. How big are numbers being compared? Around 100?

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -490,3 +490,156 @@ TEST(caffe2, replaceNaN) {
...
+  Tensor O(ElemKind::FloatTy, {kRows});
+  auto OH = O.getHandle();
+
+  for (size_t i = 0; i < kRows; ++i) {
+    auto dotProduct = 0.0f;
+
+    // Compute dot product of the i-th row of X and Y.
+    for (size_t j = 0; j < kCols; ++j) {
+      auto inputIdx = XH.getElementPtr({i, j});
+      dotProduct += (XH.raw(inputIdx) * YH.raw(inputIdx));
```
artemrakhov: could you use `.at({i, j})` here and in other cases?

tests/unittests/caffe2ImporterTest.cpp
SplitInfinity: Yes, I believe so, but it depends on the order and sign of the numbers generated to fill the input tensor.

lib/Graph/Graph.cpp
```diff
@@ -1605,6 +1605,30 @@ SaveNode *Function::createSave(Context &ctx, llvm::StringRef name,
...
+                                 NodeValue Y) {
+  auto XDimsSize = X.dims().size();
+
+  // Avoid unused variable warning in release builds.
+  (void)XDimsSize;
+
+  // X and Y must have the same shape.
+  assert(X.dims() == Y.dims() && "X and Y must have the same shape");
+  // X and Y must be 1D or 2D.
+  assert((XDimsSize == 1) || (XDimsSize == 2));
```
artemrakhov: Same here. Please move comment inside assert. And I think we can now delete comments, since their content is duplicated in the error message.

tests/unittests/caffe2ImporterTest.cpp
artemrakhov: Could you check magnitude of the numbers here in output? Also, if input size is increased to 1000x2000 will the error increase? If not, then it's probably normal precision error coming from C++ compiler optimizations, and it's not worth detailed investigations.

SplitInfinity: Just tried with 1000x2000 and error didn't increase.

SplitInfinity:
@qcolombet 

Have I missed any of the changes you requested?

nadavrot:
Nice!

PR #1747, opened 2018-09-29
[Loader] Add Fp16 conversion

qcolombet:
*Description*:
This lengthy PR (I would recommend to review the patch individually) adds the ability to convert an fp32 model into an fp16 model. This can be done by using `-use-fp16`.

The meat of the patch is actually not the change to the loader, but all the ground work to have type conversion being a thing at the node level, tensor level, function level.

*Testing*:
Added test cases.

*Documentation*:
N/A

Related to #1329.

### DIFF
--- a/include/glow/Graph/Utils.h
+++ b/include/glow/Graph/Utils.h
@@ -132,6 +132,21 @@ public:
   }
 };
 
+/// Change the type to \p newTy for all the results of \p node
+/// whose element type is equal to \p origTy.
+/// \note This API does not support converting to a quantized type.
+inline void mutateNodeResTypeThatMatch(Module &mod, Node &node, ElemKind origTy,
+                                       ElemKind newTy) {
+  for (unsigned idx = 0, end = node.getNumResults(); idx != end; ++idx) {
+    NodeValue res = node.getNthResult(idx);
+    if (res.getElementType() != origTy) {
+      continue;
+    }
+    TypeRef newTyRef = mod.uniqueType(newTy, res.dims());
+    res.setType(newTyRef);
+  }
+}
+
 } // namespace glow
 
 #endif // GLOW_GRAPH_UTILS_H
--- a/include/glow/Graph/Utils.h
+++ b/include/glow/Graph/Utils.h
@@ -147,6 +147,15 @@ inline void mutateNodeResTypeThatMatch(Module &mod, Node &node, ElemKind origTy,
   }
 }
 
+/// Replace with \p newTy, the type of all the nodes, placeholders and so on
+/// that use \p origTy.
+/// If a \p context is provided, this method makes sure the tensors
+/// backing the placeholders are properly updated. Otherwise,
+/// this method assumes the placeholders are not bound to anything.
+/// \pre The parent module has only one function, this one.
+void mutateNodesType(Function &F, ElemKind origTy, ElemKind newTy,
+                     Context *context = nullptr);
+
 } // namespace glow
 
 #endif // GLOW_GRAPH_UTILS_H
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -16,6 +16,7 @@
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Context.h"
 #include "glow/Graph/Nodes.h"
+#include "glow/Graph/Utils.h"
 #include "glow/Quantization/Base/Base.h"
 #include "glow/Support/Support.h"
 
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -161,6 +161,32 @@ TEST_P(InterpAndCPU, CmpEQ) {
   }
 }
 
+/// Check that the add operator works properly.
+TEST_P(Operator, add) {
+  PseudoRNG PRNG;
+
+  auto *inputA =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 3, 3, 1}, "A", false);
+  ctx_.allocate(inputA)->getHandle<float>().randomize(-3.0, 3.0, PRNG);
+  auto *inputB =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 3, 3, 1}, "B", false);
+  ctx_.allocate(inputB)->getHandle<float>().randomize(-3.0, 3.0, PRNG);
+  auto *Pool = F_->createAdd("pool", inputA, inputB);
+  auto *S = F_->createSave(ctx_, "save", Pool);
+  ctx_.allocate(S->getPlaceholder());
+
+  EE_.compile(CompilationMode::Infer, F_, ctx_);
+  EE_.run();
+
+  auto result = ctx_.get(S->getPlaceholder())->getHandle();
+  auto handleA = ctx_.get(inputA)->getHandle();
+  auto handleB = ctx_.get(inputB)->getHandle();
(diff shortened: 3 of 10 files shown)


include/glow/Base/Tensor.h
```diff
@@ -364,6 +364,24 @@ class Tensor final {
     }
   }
 
+  /// Cast each element of this tensor to \p newTy.
+  void cast(ElemKind newTy) {
```
nadavrot: Maybe a better name would be convertToType?

include/glow/Graph/Graph.h
```diff
@@ -679,6 +690,15 @@ class Function final : public Named {
...
   /// \returns pointer to the class member for the nodes list.
   static NodesList Function::*getNodesMemberPtr() { return &Function::nodes_; }
+
+  /// Replace with \p newTy, the type of all the nodes, placeholders and so on
+  /// that use \p origTy.
+  /// If a \p context is provided, this method makes sure the tensors
+  /// backing the placeholders are properly updated. Otherwise,
+  /// this method assumes the placeholders are not bound to anything.
+  /// \pre The parent module has only one function, this one.
+  void mutateNodesType(ElemKind origTy, ElemKind newTy,
```
nadavrot: Why is this a member of the module? The cloner and the quantizer are methods that operate outside the Module, not as members. This methods does not need to know the internals of the Module.

include/glow/Graph/Utils.h
```diff
@@ -132,6 +132,21 @@ class GraphPreOrderVisitor : public PreOrderVisitor {
   }
 };
 
+/// Change the type to \p newTy for all the results of \p node
```
nadavrot: What happens if the original type is Integer?  Is the only conversion possible is between Float and FP16? In that case, why is this a general API of the graph?

lib/Graph/Graph.cpp
```diff
@@ -2164,6 +2166,35 @@ Function *Function::clone(llvm::StringRef newName,
   return newF;
 }
 
+void Function::mutateNodesType(ElemKind origTy, ElemKind newTy,
```
nadavrot: Why is this a member of Function? The quantizer is not a member of Function.

tests/unittests/OperatorTest.cpp
```diff
@@ -1389,11 +1388,71 @@ void checkFloat16Convolution(ExecutionEngine &EE, unsigned convDepth) {
...
+
+  EE.compile(CompilationMode::Infer, F, ctx);
+
+  EE.run();
+
+  // Save the value of the fp16 run.
+  Tensor float16OutTensor = out->clone();
+
+  // Convert the graph to single precision.
+  F->mutateNodesType(ElemKind::Float16Ty, ElemKind::FloatTy, &ctx);
```
nadavrot: I don't think that this method makes sense to me. The quantizer clones the function and creates a new function that operates on in integers. The two methods should be compatible. There is no reason to have two different kinds of mechanisms, and especially not being a member of the function.

tests/unittests/graphTest.cpp
```diff
@@ -45,6 +45,57 @@ TEST(Graph, testVariableErasure) {
   EXPECT_EQ(std::distance(vars.begin(), vars.end()), vars.size());
 }
 
+/// Check that the clear method completely reset a module.
+TEST(Graph, clear) {
+  Module M;
+
+  // Check that the module is initially empty.
+  EXPECT_EQ(M.getVars().size(), 0);
```
nadavrot: Is it worth our time to check that the module starts with zero vars, zero placeholders, zero functions and then check it again and again and again and again 7 times in the body of the function?

tests/unittests/tensorsTest.cpp
```diff
@@ -324,6 +324,45 @@ TEST(Tensor, copyWithCast) {
...
+  }
+
+  // Cast back B from float16_t to float.
+  B.cast(ElemKind::FloatTy);
+  ASSERT_EQ(B.getElementType(), ElemKind::FloatTy);
+  {
+    auto BH = B.getHandle<>();
+
+    EXPECT_EQ(A.size(), B.size());
+    for (size_t idx = 0, end = A.size(); idx != end; ++idx) {
```
nadavrot: Tensors/Handles have equality methods.

qcolombet: Right, same type means I can use isEqual.

tests/unittests/graphTest.cpp
qcolombet: We need to define a baseline and check that we get back to that base line when we call clear. I don’t expect this to be measurable chunk of time of our testing.

What is your concern here?

tests/unittests/OperatorTest.cpp
qcolombet: I agree, and I’d argue the quantizer does it wrong. Indeed each type we add a new node, we have to go modify the quantization and create the node manually with the right type.

I agree we can do this on the site, but I equally think it could be part à the function API.

I don’t have a strong opinion either way. If you do I am happy to move the functionality to the side.

include/glow/Graph/Utils.h
qcolombet: The API would work equally for integer. Why do you think this is not general?

Note: I didn’t implemented the underlying casts from fp to int because I didn’t have any use case, but I don’t see any problem with extending it for that.

include/glow/Graph/Graph.h
qcolombet: Yes and no. We need to know what structures need to be updated. It happens that all what need to be updated in visible from the outside (node, placeholder and variable) but it could’ve not been the case. Moving it elsewhere exposes us to tighten the coupling between the internals of the graph and whatever does the conversion.

include/glow/Base/Tensor.h
qcolombet: Sounds gods!

include/glow/Graph/Graph.h
nadavrot: I don't understand this part. The optimizer is not a part of the graph. Right? 

You don't need to be a part of the internals of Function to be able to mutate it.

tests/unittests/OperatorTest.cpp
nadavrot: Regardless of which design is correct they should be consistent. At the moment your API handles the inputs/outputs and the processing of the network itself in a completely different way. 

 I like the design where we clone the function before modifying it because it allows to keep the original function and use it for testing the outputs. 

If you think that the implementation of the quantizer can be changed then let's do it. If you think that the API of the quantizer can be changed then let's also do that. But let's make them consistent. This includes the loader and the compiler.

tests/unittests/graphTest.cpp
nadavrot: This test is very long. I don't understand why we need to check that we didn't add new variables (and functions and placeholders) again and again.

qcolombet:
Split the PR in #1752, #1753, #1754.
This doesn't include the refactoring of quantization to use the same API as the conversion stuff. Working on this.

qcolombet:
The refactoring of quantization and conversion discussed here is done with the refactoring PR #1820, #1821, #1822, #1823.

qcolombet:
PR #1858 puts us just one step away from the functionality provided by this PR: running resnet50 in fp16. We miss only the conversion of the input and output and exposing the conversion mechanism in the loader.

qcolombet:
Conversion added to the loader in #1866.
The difference with that prototype is #1866 does not modify the input and output of the network.

PR #1754, opened 2018-10-01
[Tensor] Add a convertToType method

qcolombet:
*Description*
This patch adds a convertToType method that allows to convert the content
of a tensor from its current type to a new type.

The shape of the tensor remains unchanged, but the element type
gets updated to the new type.

For now, this method only supports conversion between float and
float16_t.

*Testing*
Added test case.

Related to #1329, this patch adds the functionality required to
change the content of a network while doing conversion from one
type to another.

*Description*:
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -364,6 +364,9 @@ public:
     }
   }
 
+  /// Convert each element of this tensor to \p newTy.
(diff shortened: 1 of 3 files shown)


qcolombet:
Sure thing.
Done!

PR #1756, opened 2018-10-01
[placeholders] Convert rowwise quantized FC

bertmaher:


lib/Graph/Graph.cpp
```diff
@@ -575,16 +575,15 @@ Function::createRowwiseQuantizedFullyConnected(Context &ctx,
   // it is assumed to be quantized data and the scale and offset should be
   // provided. But for rowwise quantization, the scales and offsets are stored
   // in vectors separately, we add the dummy scale and offset here.
-  auto *qWeights = getParent()->createVariable(ElemKind::Int8QTy, W->dims(),
-                                               0.0, 0, "weights.rwqfc",
-                                               VisibilityKind::Private, false);
+  auto *qWeights = getParent()->createPlaceholder(
+      ElemKind::Int8QTy, W->dims(), 0.0, 0, "weights.rwqfc", false);
```
beicy: Here qWeights is a constant. That is why I used "createVariable" for it.

bertmaher: Hm, I see your point.  Would this member ever be trained?  (i.e., could it be non-constant in the future?)

nadavrot: I don't think that we can train quantized operators.

beicy: No. The row-wise quantization is only for const weights so far.

PR #1758, opened 2018-10-01
[Optimizer] Fix the batchnorm(conv) opt for FP16

qcolombet:
[Optimizer] Teach the bacthNorm(Conv) optimization about FP16

*Description*
Prior to this patch the optimization that merges convolution followed
by batchNorm would assert on fp16 because the handles expect a regular
single precision floating point type.

This patch templatizes this part of the optimization so that it
works seamlessly with single or half precision floating point numbers.

*Testing*
Added test case.

Related to #1329

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -905,6 +905,100 @@ static Variable *getUniquelyUsedVariable(Node &node) {
   return dyn_cast<Variable>(&node);
 }
 
+/// Normalize the weight of \p CV with what \p BN is doing.
+/// \return Whether or not the normalization was possible.
+template <typename ElemTy>
+bool normalizeWeights(ConvolutionNode &CV, BatchNormalizationNode &BN) {
+
+  static_assert(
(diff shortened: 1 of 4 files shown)


nadavrot:
Nice!

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -905,6 +905,100 @@ static Variable *getUniquelyUsedVariable(Node &node) {
   return dyn_cast<Variable>(&node);
 }
 
+/// Normalize the weight of \p CV with what \p BN is doing.
+/// \return Whether or not the normalization was possible.
+template <typename ElemTy>
+bool normalizeWeights(ConvolutionNode &CV, BatchNormalizationNode &BN) {
+
+  static_assert(
```
rdzhabarov: how is it different from GLOW_ASSERT?

qcolombet: That’s a compile time assert. If you try to instantiate this function with a non-supported type, the compiler compiling glow will stop with an error.

PR #1761, opened 2018-10-02
[travis] Only `cat` detailed log on failure.

rdzhabarov:
*Description*:
Do not `cat` detailed logs on a successful run.

*Testing*:
Travis CI

*Documentation*:
N/A


jfix71:
```
 - CTEST_PARALLEL_LEVEL=2 ninja test || cat Testing/Temporary/LastTest.log
```

This line's exit status will always be zero even if ninja test fails and exits with non-zero status. Does Travis still report failure in this case?

rdzhabarov:
@jfix71 good point. I'll check.

PR #1768, opened 2018-10-02
[Backends.md] Add documentation on APIs for backends

jfix71:
This document should make it easier and more clear what is necessary for a backend to implement and what each API should be used for.


nadavrot:
Thanks for doing this work @jfix71 .

Issue #1755, opened 2018-10-01
[ONNXIFI] Crashed on 20 onnx test cases.


zrphercule:
This is a list of the crashed test cases. All crashes can be reproduced following the steps of #1736 .
It is still unclear if these are problems of onnxifi or glow.
Crashed onnx test cases:
  1 test_averagepool_2d_same_lower
  2 test_reshape_reordered_dims
  3 test_maxpool_with_argmax_2d_precomputed_strides
  4 test_softmax_axis_0
  5 test_maxpool_2d_precomputed_same_upper
  6 test_maxpool_2d_same_lower
  7 test_reshape_one_dim
  8 test_batchnorm_epsilon
  9 test_averagepool_2d_same_upper
 10 test_batchnorm_example
 11 test_reshape_extended_dims
 12 test_maxpool_2d_same_upper
 13 test_reshape_negative_dim
 14 test_maxpool_with_argmax_2d_precomputed_pads
 15 test_reshape_reduced_dims
 16 test_averagepool_2d_precomputed_same_upper
 17 test_sum_example
 18 test_flatten_axis0
 19 test_transpose_default
 20 test_sum_one_input

rdzhabarov:
That's a great data point! Thanks, @zrphercule for discovering those!

bertmaher:
I'll grab a block of these if nobody else has taken any.  I'll claim the first 3:
1 test_averagepool_2d_same_lower
2 test_reshape_reordered_dims
3 test_maxpool_with_argmax_2d_precomputed_strides

zrphercule:
@bertmaher I wonder how much work we need to do if we want to not use assert() in onnxifi-glow at all? Since I think it is very possible that the rest of these 20 cases are all due to these assert().
A better solution is to return exceptional onnxStatus and let test driver to handle this.

rdzhabarov:
@jackm321 another option, feel free to ping me offline for any details.

jackm321:
After #2143, the latest list of ONNXIFI node tests that are either failing or crashing can be found below. An important next step here is to dig into what is causing these tests to crash/fail and determine what can be done to prevent this in the loader or fix any underlying bugs in glow.

**failing:**
* test_gemm_broadcast
* test_averagepool_2d_pads
* test_gemm_nobroadcast
* test_averagepool_2d_precomputed_pads
* test_max_one_input
* test_min_example
* test_split_equal_parts_default_axis
* test_split_variable_parts_1d
[...]

PR #1771, opened 2018-10-02
[importer] Report error if reshape given with non-constant shape input

bertmaher:
*Description*: Glow requires static shapes, so yield an error if non-constant reshape is given.
*Testing*: onnx test case, test_reshape_reordered_dims
*Documentation*: N/A

See #1755

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -229,7 +229,7 @@ protected:
     }
   }
 
-  void loadReshape(const OpType &op, ArgumentDictionaryTy &dict) {
+  bool loadReshape(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in = getNodeValueOrCreateVariableByName(op.input(0));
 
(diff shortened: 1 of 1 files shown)


lib/Importer/ProtobufLoader.cpp
```diff
@@ -34,6 +34,10 @@ Tensor *ProtobufLoader::getTensorByName(llvm::StringRef name) {
   return tensors_[name];
```
artemrakhov: Maybe not critical, but now `tensors_` can contain nulls instead of valid Tensor pointers. So we'd need to add assertion to this method: not only check that `tensors_.count(name)`, but also `tensors_[name]`.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -104,6 +104,10 @@ class ProtobufLoader {
   /// \returns the tensor that was registered under the name \p name.
   Tensor *getTensorByName(llvm::StringRef name);
 
+  /// \returns the tensor that was registered under the name \p name, or null if
+  /// no such tensor exists.
+  Tensor *getTensorByNameOrNull(llvm::StringRef name);
```
artemrakhov: Before having this method, we normally did checks like `if (tensors_.count(name))`. Does it make sense to switch them to use the new method?

lib/Importer/ProtobufLoader.cpp
bertmaher: Ew.  Yeah, I don't think I want [] here.

include/glow/Importer/ProtobufLoader.h
bertmaher: I'm just going to use tensors_.count() to perform the check.  I don't think there's too much need to add more API, since that's a member anyways.

lib/Importer/ProtobufLoader.cpp
bertmaher: Great catch, btw!

PR #1773, opened 2018-10-03
[Caffe2] Support FCTransposed op.

### LINKED ISSUE #1697: [New operator] FCTransposed
We need to support `FCTransposed` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#fctransposed

beicy:
*Description*:
Support caffe2 FCTransposed op.
https://caffe2.ai/docs/operators-catalogue.html#fctransposed

*Testing*: 
Added unittest in caffe2ImporterTest.cpp
*Documentation*:
[Optional Fixes #1697 ]

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -294,7 +294,7 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
-  if (typeName == "FC") {
+  if (typeName == "FC" || typeName == "FCTransposed") {
     // Load the inputs:
     auto in = getNodeValueOrCreateVariableByName(op.input(0));
     if (in.getType()->dims().size() > 2) {
@@ -309,14 +309,19 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     // Caffe2 stores the transposed W matrix. In here we first coerce W to a 2D
     // matrix size if necessay and then transpose it back.
-    Tensor wtag;
+    Tensor tmp;
     if (w->dims().size() > 2) {
       auto wDims = flattenCdr(w->dims(), axis_w);
-      Tensor tmp(ElemKind::FloatTy, {wDims.first, wDims.second});
+      tmp.reset(ElemKind::FloatTy, {wDims.first, wDims.second});
       tmp.copyRawFrom(w);
-      tmp.transpose(&wtag, {1, 0});
-    } else
+      w = &tmp;
+    }
+    Tensor wtag;
+    if (typeName == "FC") {
       w->transpose(&wtag, {1, 0});
+    } else {
+      wtag.assign(w);
+    }
 
     auto W = G_.getParent()->addVar(
         new Variable("weights", VisibilityKind::Private, std::move(wtag)));
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -362,6 +362,193 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
   // should be covered in the OperatorTest for MatMul already.
 }
 
+/// Test loading a FC node : I * transpose(W) + B.
+TEST(caffe2, FC) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename("tests/models/caffe2Models/fc_predict_net.pbtxt");
+  std::string NetWeightFilename("tests/models/caffe2Models/fc_init_net.pbtxt");
+
+  Placeholder *output;
+  Context ctx;
+  // Destroy the loader after the graph is loaded since the following execution
+  // will not depend on anyting from the loader.
+  {
+    Tensor inputs(ElemKind::FloatTy, {2, 3});
+    inputs.getHandle() = {1, 2, 3, 4, 5, 6};
+
+    // Weights and bias are read from NetWeightFilename. And the values are:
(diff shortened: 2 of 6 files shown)


lib/Importer/Caffe2.cpp
```diff
@@ -314,9 +314,16 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       auto wDims = flattenCdr(w->dims(), axis_w);
       Tensor tmp(ElemKind::FloatTy, {wDims.first, wDims.second});
       tmp.copyRawFrom(w);
-      tmp.transpose(&wtag, {1, 0});
-    } else
+      if (typeName == "FC") {
```
jfix71: This feels a bit unnecessary, with two separate checks for `if (typeName == "FC")`. Is it possible to wrap all the logic already in here for transposing weights in `if (typeName == "FC")`, and else just assign `wtag`?

beicy: We need to flatten the w if its dim size is larger than 2.  So here is the logic:
1) if w's dim size > 2 and FC  then flatten and transpose.
2) if w's dim size <= 2 and FC then transpose.
3) if w's dim size > 2 and FCT then flatten and assign.
4) if w's dim size <= 2 and FCT then assign. 
Now to reused the "flatten" code, I use w'sdim size >2 as the outer condition. If we could like to use the (typename == "FC" )  as the outer condition, I think we have to duplicate the "flatten" code.

jfix71: Got it. Would it work to perform flatten if necessary, and then perform transpose if necessary? Something like:

```cpp
if (w->dims().size() > 2) {
  // flatten logic
}

if (typeName == "FC") {
  // transpose logic
}
```

beicy: Yes, we could do that. But in that case, no mater the flatten is needed, we need to copy w to a temp tensor(directly or flatten) and then do the transpose or direct copy.  I remember we added " w->dims().size() > 2" to avoid non-necessary copy. Otherwise, we can simply call "flatten" and then  run the tranpose logic.

jfix71: Ah ok, got it. In that case I wonder if we could use `std::move()` here instead of `assign()` to avoid any unnecessary copies? E.g. 

```cpp
if (w->dims().size() > 2) {
  // tmp = flattened w
  wtag = std::move(tmp);
} else {
  wtag = std::move(*w);
}
```

jfix71: True! If two FCs in the proto reference use the same weight then it would be an issue. I suppose such a network is technically valid, though unlikely in practice and not yet seen. BTW, we are currently using `std::move(*b)` for creating the bias (`Variable B`). So that should also be fixed for the same reason.

Anyway, this means no matter what we need to make an extra copy of `w`. So in the snippet I wrote above in the else case, if we replace `wtag = std::move(*w);` with `wtag.assign(w);` I believe it works fine, correct?

beicy: Yes. In my last version,  std::move is applied to Tensor tmp as you suggested, but keep using wtag.assign(w) in the else part.

beicy:
@nadavrot I think the execution code is necessary here. Since we do some extra work before create a FC code: 
1) For Caffe2 FC op: the weights need to be flatten(optional) and transposed. 
2) For Caffe2 FCTransposed op: the weights need to be flatten(optional).
For the above weights flatten, we don't create a node for it and just do the flatten on the fly if it is necessary. However, our in Operator test for FC the inputs are well prepared.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -314,9 +314,16 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       auto wDims = flattenCdr(w->dims(), axis_w);
       Tensor tmp(ElemKind::FloatTy, {wDims.first, wDims.second});
       tmp.copyRawFrom(w);
-      tmp.transpose(&wtag, {1, 0});
-    } else
+      if (typeName == "FC") {
```
nadavrot: Can you please add a comment that explains what this logic does?

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -349,6 +349,153 @@ TEST(caffe2, parallelBatchedMatmulRHS) {
...
+  for (size_t i = 0; i < 2 * 4; i++)
+    EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
+
+  // Test the case that the input needs to be flatten.
+  Function *F1 = mod.createFunction("main1");
+  {
+    Tensor inputs(ElemKind::FloatTy, {2, 1, 3});
+    inputs.getHandle() = {1, 2, 3, 4, 5, 6};
+
+    // weights and bias are read from NetWeightFilename. And the values are:
```
nadavrot: weights -> Weights

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -314,9 +314,16 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       auto wDims = flattenCdr(w->dims(), axis_w);
```
artemrakhov: @beicy : I think this piece of code can be simpler:
```
Tensor tmp;
if (w->dims().size() > 2) {
  auto wDims = flattenCdr(w->dims(), axis_w);
  tmp.reset(ElemKind::FloatTy, {wDims.first, wDims.second});
  tmp.copyRawFrom(w);
  w = &tmp;
}
Tensor wtag;
if (typeName == "FC") {
  w->transpose(&wtag, {1, 0});
} else {
  wtag.assign(w);
}
```

beicy: @artemrakhov Yes, it looks much better :) Thanks a lot!

PR #1774, opened 2018-10-03
[Caffe2] Implement BatchBoxCox Caffe2 operator

### LINKED ISSUE #1705: [New operator] BatchBoxCox
We need to support `BatchBoxCox` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#batchboxcox

SplitInfinity:
**Description**
This commit adds support for the `BatchBoxCox` Caffe2 operator. This
operator accepts a matrix `data` and two vectors `lambda1` and `lambda2`
as input and computes the two-parameter Box-Cox transform of each column
`c` of `data` with parameters `lambda1` and `lambda2`. The transform
`y_i` for each element `c_i` of the input column is given by

<a href="https://www.codecogs.com/eqnedit.php?latex=y&space;\left&space;(&space;x,&space;\lambda_{1},&space;\lambda_{2}&space;\right&space;)=&space;\begin{cases}&space;\ln&space;\left&space;(&space;\max(x&space;&plus;&space;\lambda_{2},&space;1&space;\times&space;10^{-6})\right&space;)&space;&&space;,\lambda_{1}&space;=&space;0&space;\\&space;\frac{(\max(x&space;&plus;&space;\lambda_{2},&space;1&space;\times&space;10^{-6}))^{\lambda_{1}}&space;-&space;1}{\lambda_{1}}&space;&&space;,\lambda_{1}&space;\neq&space;0&space;\\&space;\end{cases}" target="_blank"></a>

The `BatchBoxCox` operator is immediately lowered to a combination of
`Broadcast`, `Splat`, `Add`, `Pow`, `Max`, `Sub`, `Log`, `Div`, `Select`
and `CmpEq` nodes. After broadcasting `lambda1` and `lambda2` so that
they have the same shape as `data`, both cases shown above are computed
using basic elementwise arithmetic nodes, and a `Select` node is used at
the end to chose between them.

**Testing**
This commit adds a test case for this operator to `caffe2ImporterTest.cpp`.

**Fixes**
This commit fixes #1705.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1293,12 +1293,30 @@ void InterpreterFunction::fwdElementCmpLTEInst(const ElementCmpLTEInst *I) {
   }
 }
 
-void InterpreterFunction::fwdElementCmpEQInst(const ElementCmpEQInst *I) {
-  auto outW = getWeightHandle<int64_t>(I->getDest());
-  auto lhsW = getWeightHandle<int64_t>(I->getLHS());
-  auto rhsW = getWeightHandle<int64_t>(I->getRHS());
+template <typename ElemTy>
+void InterpreterFunction::fwdElementCmpEQInstImpl(const ElementCmpEQInst *I) {
+  auto outW = getWeightHandle<ElemTy>(I->getDest());
+  auto lhsW = getWeightHandle<ElemTy>(I->getLHS());
+  auto rhsW = getWeightHandle<ElemTy>(I->getRHS());
   for (size_t i = 0, e = outW.size(); i < e; i++) {
     outW.raw(i) = lhsW.raw(i) == rhsW.raw(i) ? 1 : 0;
+  }
+}
+
+void InterpreterFunction::fwdElementCmpEQInst(const ElementCmpEQInst *I) {
+  auto *T = getTensor(I->getDest());
+
+  switch (T->getElementType()) {
+  case ElemKind::FloatTy: {
+    fwdElementCmpEQInstImpl<float>(I);
+    break;
+  }
+  case ElemKind::Int64ITy: {
+    fwdElementCmpEQInstImpl<int64_t>(I);
+    break;
+  }
+  default:
+    llvm_unreachable("fwdElementCmpEQInst not implemented for this type!");
   }
 }
 
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1477,6 +1477,66 @@ Node *Function::createWeightedSum(llvm::StringRef name,
   return currAdd;
 }
 
+Node *Function::createBatchBoxCox(llvm::StringRef name, NodeValue data,
+                                  NodeValue lambda1, NodeValue lambda2) {
+  assert((lambda1.dims() == lambda2.dims()) &&
+         "lambda1 and lambda2 must have the same shape");
+  assert((lambda1.getType()->getElementType() == lambda2.getElementType()) &&
+         "lambda1 and lambda2 must have the same element type");
+  assert((lambda1.getType()->getElementType() == data.getElementType()) &&
+         "data and lambdas must have the same element type");
+  assert((lambda1.dims().size() == 1) && "lambda1 and lambda2 must be vectors");
+  assert((data.dims().size() == 2) && "data must be a matrix");
+  assert((data.dims()[1] == lambda1.dims()[0]) &&
+         "data, lambda1 and lambda2 must have the same number of rows");
+
+  // Broadcast lambda1 and lambda2 so that they are both the same size as the
+  // data.
+  auto *BL1 = createBroadcast(name.str() + ".broadcast", lambda1, data.dims(),
+                              /*axis=*/1);
+  auto *BL2 = createBroadcast(name.str() + ".broadcast", lambda2, data.dims(),
+                              /*axis=*/1);
+
+  // Add a small epsilon to lambda1 so that we can avoid dividing by zero later.
+  // It doesn't matter that this is technically incorrect because the final
+  // Select will discard the results of this computation.
+  auto *eps = createSplat(name.str() + ".eps", BL1->getNthResult(0).getType(),
+                          std::numeric_limits<float>::min());
+  auto *EBL1 = createAdd(name.str() + ".lambda1eps", BL1, eps);
+
+  // Compute data + BL2, which is needed regardless of whether
+  // lambda1 is 0 or not.
+  auto *AN = createAdd(name.str() + ".add", data, BL2);
+
+  // Take the max of data + BL2 and 1e-6 to void exponentiating or taking the
+  // logarithm of too small a number.
+  auto *minArg =
+      createSplat(name.str() + ".logpowmin", AN->getResult().getType(), 1e-6);
+  auto *MN = createMax(name.str() + ".max", AN, minArg);
+
+  // Compute the Box-Cox transform for the lambda1 == 0 case:
+  //    y = ln(max(x + lambda2, 1e-6))
+
+  auto *LN = createLog(name.str() + ".log", MN);
+
+  // Compute the Box-Cox transform for the lambda1 != 0 case:
+  //    y = (max(x + lambda2, 1e-6)^lambda1 - 1)/lambda1
+  auto *PN = createPow(name.str() + ".pow", MN, BL1);
+  auto *ones =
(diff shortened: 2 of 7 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1202,11 +1202,23 @@ void InterpreterFunction::fwdElementCmpLTEInst(const ElementCmpLTEInst *I) {
 }
 
 void InterpreterFunction::fwdElementCmpEQInst(const ElementCmpEQInst *I) {
-  auto outW = getWeightHandle<int64_t>(I->getDest());
```
SplitInfinity: I don't intend this to be the final solution for float tensor comparisons; I mainly wanted to get this PR out for discussion and feedback.

For the purposes of implementing this operator, I need to compare to a literal 0, so a fuzzy float comparison won't really be useful. How does everyone feel about a different instruction for float comparisons? Or an `IsZero` node, similar to the `IsNaN` node that was added recently?

lib/Graph/Graph.cpp
```diff
@@ -1482,6 +1482,58 @@ Node *Function::createWeightedSum(llvm::StringRef name,
...
+                              /*axis=*/1);
+
+  // Compute data + BL2, which is needed regardless of whether
+  // lambda1 is 0 or not.
+  auto *AN = createAdd(name.str() + ".add", data, BL2);
+
+  // Compute the Box-Cox transform for the lambda1 == 0 case:
+  //    y = ln(x + lambda2)
+
+  // Clip the data to the log below at 1e-6 to avoid taking the logarithm
```
SplitInfinity: This clip value is from Caffe2: https://github.com/pytorch/pytorch/blob/master/caffe2/operators/batch_box_cox_op.cc#L97

lib/Backends/Interpreter/InterpreterNodes.cpp
beicy: In that case, maybe you can put *WIP* in the title :)

SplitInfinity:
What's the best way to reproduce and investigate the ASAN test failure for this PR? I tried running the commands shown in the Travis log on my laptop but all tests consistently pass.

opti-mix:
@SplitInfinity Have you tried to build Glow with ASAN/sanitizers support enabled as described [here](https://github.com/pytorch/glow/blob/master/docs/Building.md) and see if you can reproduce the issue?

lib/Graph/Graph.cpp
```diff
@@ -1482,6 +1482,58 @@ Node *Function::createWeightedSum(llvm::StringRef name,
...
+  auto *ones =
+      createSplat(name.str() + ".ones", PN->getResult().getType(), 1.0f);
+  auto *SN = createSub(name.str() + ".sub", PN, ones);
+  auto *DN = createDiv(name.str() + ".div", SN, BL1);
+
+  // Compute predicates for selecting between the two cases above.
+  auto *zeroes =
+      createSplat(name.str() + ".floatmin", BL1->getNthResult(0).getType(),
+                  std::numeric_limits<float>::min());
+  auto *predicate = createCmpLTE(name.str() + ".cmplte", BL1, zeroes);
```
jfix71: Shouldn't this be CmpEQ?

lib/Graph/Graph.cpp
jfix71: This could make sense, but I would save it for a different PR, and leave it as is here.

lib/Graph/Graph.cpp
```diff
@@ -1482,6 +1482,58 @@ Node *Function::createWeightedSum(llvm::StringRef name,
...
+  auto *MN = createMax(name.str() + ".logmax", AN, minArg);
+  auto *LN = createLog(name.str() + ".log", MN);
+
+  // Compute the Box-Cox transform for the lambda1 != 0 case:
+  //    y = ((x + lambda2)^lambda1 - 1)/lambda1
+  auto *PN = createPow(name.str() + ".pow", AN, BL1);
+  auto *ones =
+      createSplat(name.str() + ".ones", PN->getResult().getType(), 1.0f);
+  auto *SN = createSub(name.str() + ".sub", PN, ones);
+  auto *DN = createDiv(name.str() + ".div", SN, BL1);
```
jfix71: Unfortunately this is going to execute for all elements regardless of if BL1 is zero. So it's perhaps the source of your UBSAN div by zero error. The way you have implemented this, both paths are executed and one is just selected in the end. This also means we are doing unnecessary work. It would be a great place for us to use predication, but of course it's not supported yet.

For now you could do something like CmpEQ(0) + Select a Splat close to zero as lambda1 to avoid div by zero, even though you know it will be thrown away in the final select regardless. It's not ideal though, as you need another Select... Anyone else have other ideas?

qcolombet: > It would be a great place for us to use predication, but of course it's not supported yet.

Predication as we envision it is not required for correctness. (Some hardwares may not support it.) Therefore we need to make sure we don't execute division by zero.

> For now you could do something like CmpEQ(0) + Select a Splat close to zero as lambda1 to avoid div by zero, even though you know it will be thrown away in the final select regardless. It's not ideal though, as you need another Select... Anyone else have other ideas?

If we are going to throw the result away, I'd rather divide by one or anything that is likely to be fast to execute. Could we do the computation on only what make sense instead?
(I haven't read the patch nor the semantic of the operator.)

qcolombet: Okay read the operator semantic and this "Could we do the computation on only what make sense instead?" doesn't seem to be practical.
So yeah, prepare the divider upfront sounds reasonable to me.

qcolombet: Assuming this doesn't affect the precision too much we could avoid the select with a splat and just add epsilon to every value.
I.e., instead of
```
predicate = input == 0
divider = select predicate, input, splat(1)
```
We could do
```
divider = input + splat(epsilon)
```
(This assumes that the input is not -epsilon though...)

lib/Graph/Graph.cpp
```diff
@@ -1482,6 +1482,65 @@ Node *Function::createWeightedSum(llvm::StringRef name,
...
+  // Clip the data to the log below at 1e-6 to avoid taking the logarithm
+  // of too small a number.
+  auto *minArg =
+      createSplat(name.str() + ".logminarg", AN->getResult().getType(), 1e-6);
+  auto *MN = createMax(name.str() + ".logmax", AN, minArg);
+  auto *LN = createLog(name.str() + ".log", MN);
+
+  // Compute the Box-Cox transform for the lambda1 != 0 case:
+  //    y = ((x + lambda2)^lambda1 - 1)/lambda1
+  auto *PN = createPow(name.str() + ".pow", AN, BL1);
```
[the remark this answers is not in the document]
artemrakhov: Yes, that's the best way!

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1260,11 +1260,23 @@ void InterpreterFunction::fwdElementCmpLTEInst(const ElementCmpLTEInst *I) {
...
+  ElemKind k = T->getElementType();
+
+  if (k == ElemKind::FloatTy) {
+    auto outW = getWeightHandle<float>(I->getDest());
+    auto lhsW = getWeightHandle<float>(I->getLHS());
+    auto rhsW = getWeightHandle<float>(I->getRHS());
+    for (size_t i = 0, e = outW.size(); i < e; i++) {
+      outW.raw(i) = lhsW.raw(i) == rhsW.raw(i) ? 1 : 0;
+    }
+  } else {
```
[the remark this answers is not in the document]
SplitInfinity: I've updated this code as per your suggestion. Let me know what you think.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1293,15 +1293,33 @@ void InterpreterFunction::fwdElementCmpLTEInst(const ElementCmpLTEInst *I) {
...
+  case ElemKind::FloatTy: {
+    fwdElementCmpEQInstImpl<float>(I);
+    break;
+  }
+  case ElemKind::Int64ITy: {
+    fwdElementCmpEQInstImpl<int64_t>(I);
+    break;
+  }
+  default:
+    assert("fwdElementCmpEQInst not implemented for this type!");
```
artemrakhov: please use `llvm_unreachable` instead of `assert`

rdzhabarov:
This needs ONNX loader and to make sure that CPU backend can run this fine.

PR #1775, opened 2018-10-03
[onnx] Glow does not support argmax output when pooling

bertmaher:
Issue #1755, test case 3 (test_maxpool_with_argmax_2d_precomputed_strides).

### DIFF
--- a/lib/Importer/ONNX.cpp
+++ b/lib/Importer/ONNX.cpp
@@ -321,6 +321,9 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
   }
 
   if (typeName == "MaxPool" || typeName == "AveragePool") {
+    // Glow doesn't support argmax output yet.
+    if (op.output_size() > 1)
+      return false;
     // Load the inputs:
     auto in = getNodeValueOrCreateVariableByName(op.input(0));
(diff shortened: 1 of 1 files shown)


lib/Importer/ONNX.cpp
```diff
@@ -321,6 +321,9 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
   }
 
   if (typeName == "MaxPool" || typeName == "AveragePool") {
+    // Glow doesn't support argmax output yet.
```
jfix71: Would it be useful to change the return type of `loadOperator()` to `llvm::Error` so these error messages can be sent upward in the stack? If so, not needed for this PR, but if it's something we want to do then better to do earlier than later.

rdzhabarov: Yup, I'm working on this to make error reporting better. We need to report this for ONNXIFI.

bertmaher: I think that's probably a good idea.  I guess we may not end up doing anything with the error messages (onnxifi will just reject the graph), but it would probably be better to have them than not.

rdzhabarov: in the ideal world we'd report unsupported param or whatnot back to pytorch/c2 during the compatibility call. ( which does not happen at the moment ).

PR #1776, opened 2018-10-03
[trivial] Fixing linter warnings

rdzhabarov:
*Description*:
Minor cleanups to make Lint happy.

*Testing*:
CI

*Documentation*:
N/A


rdzhabarov:
```
I like the always-braces style, but we omit braces lots of other places. Do we want to migrate? (Gradually, of course.)
```
Yes, currently there is a mix of both in the code base. Moving towards a single style over time makes sense.

PR #1778, opened 2018-10-03
[trivial][but contentious] Make ModelLoader naming consistent

bertmaher:
*Description*: OK, so I get that C++ isn't Java... but these classes have a non-trivial hierarchy, and when I started working on it, I kept getting confused by trivial things (like, is ModelLoader a base class?  Which file defines it?).  I'm sure some of this is a personal problem, but I also feel like it doesn't hurt to make things a bit more consistent in this area.  What do you all think?
*Testing*: `ninja check`
*Documentation*: N/A


jfix71:
I'm a big fan but I hesitate to approve until others have weighed in 😄

include/glow/Importer/ONNXIFIModelLoader.h
```diff
@@ -19,16 +19,16 @@
...
-#include "glow/Importer/ONNX.h"
+#include "glow/Importer/ONNXModelLoader.h"
 
 #include "llvm/ADT/StringMap.h"
 
 namespace glow {
 namespace onnxifi {
 
-class ModelLoader : public ONNXModelLoader {
+class ONNXIFIModelLoader : public ONNXModelLoader {
```
rdzhabarov: it's already in ```onnxifi``` namespace. I think it makes sense to keep a short name for the class.

bertmaher:
That sounds good to me.  I noticed the `onnxifi` namespace, but I'd still rather have a descriptive class name, since `ModelLoader` sounds like a base class.   (Sorry for the bikeshedding 🚲🏘️  :-D ).

include/glow/Importer/ONNXIFIModelLoader.h
```diff
@@ -19,16 +19,15 @@
 
 #include "onnx/onnxifi.h"
 
-#include "glow/Importer/ONNX.h"
+#include "glow/Importer/ONNXModelLoader.h"
```
rdzhabarov: can you fix include guards as well

PR #1781, opened 2018-10-03
[cpu] Compile libjit into CPUBackend

bertmaher:
*Description*: Glow will be easier to distribute if we can package libjit into the library, rather than keeping it as an external file dependency.  This PR implements a simple clone of `xxd -i`, which dumps a binary file as an unsigned char array, and uses it to compile libjit.bc into LLVMIRGen.
*Testing*:  `ninja check`
*Documentation*: N/A

### DIFF
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -136,6 +136,11 @@ void LLVMIRGen::loadBaseAddresses(llvm::IRBuilder<> &builder) {
   offsetsArray_ = F->args().begin() + 3;
 }
 
+static const unsigned char libjit_bc[] = {
+#include "glow/libjit_bc.inc"
+};
+static const size_t libjit_bc_size = sizeof(libjit_bc);
+
 // Search for the standard library bitcode file on disk and load it into an
 // LLVM module. We search for the standard library around the current executable
 // and also in the current directory.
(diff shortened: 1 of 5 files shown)


tools/IncludeBin/IncludeBin.cpp
```diff
@@ -0,0 +1,33 @@
+#include <cstdio>
```
jfix71: Missing copyright header?

tools/IncludeBin/IncludeBin.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <cstdlib>
+
+int main(int argc, char **argv) {
```
jfix71: Maybe add a comment explaining that this is similar to `xxd -i`?

opti-mix:
@bertmaher Can we split this PR into two parts: one for `IncludeBin` and one for using it for libjit?

And I think we would to use it for OpenCL kernels as well, right?

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -136,6 +136,11 @@ void LLVMIRGen::loadBaseAddresses(llvm::IRBuilder<> &builder) {
   offsetsArray_ = F->args().begin() + 3;
 }
 
+static const unsigned char libjit_bc[] = {
```
jfix71: Can you add comments here explaining what's going on?

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -145,37 +150,12 @@ loadStandardLibrary(llvm::LLVMContext *ctx, llvm::StringRef filename) {
...
-      }
-      return res;
-    }
-
-    // Go up the filesystem tree.
-    basePath = parent_path(basePath);
-  }
-
-  return llvm::parseIRFile(filename, error, *ctx);
+  return llvm::parseIR(
```
jfix71: Also comments, or just here if not where `libjit_bc` is declared

bertmaher:
Sure, splitting it up, will create new ones.

bertmaher:
I'll work on a 3rd PR to use this for OpenCL.

tools/IncludeBin/IncludeBin.cpp
rdzhabarov: just throwing it out there :D For utility things like that we might do python script as well.

PR #1782, opened 2018-10-03
[GraphOptimizer] Move transpose constant folding optimization before merging BatchNorm into Convolution

aturetsk:
*Description*:
An internal net model I'm working on has explicit transpose nodes for convolution weights. This prevents BatchNorm + Conv merge optimization from triggering.
The patch moves transpose constant folding before BatchNorm optimization in order to get rid of these transposes and make BatchNorm + Conv merge happen.

*Testing*:
Added a unit test.

*Documentation*:
N/A

### DIFF
--- a/tests/unittests/graphOptzTest.cpp
+++ b/tests/unittests/graphOptzTest.cpp
@@ -161,6 +161,33 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
   EXPECT_TRUE(llvm::isa<SaveNode>(save));
 }
 
+/// Check that transpose constant folding is done before BatchNorm optimization,
+/// which allows to merge BatchNorm into Convolution with transposed weights.
+TEST_F(GraphOptz, optimizeBatchNormAfterConvWithTransposedWeights) {
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 10, 20, 3}, "input", false);
+  auto *filter =
+      mod_.createPlaceholder(ElemKind::FloatTy, {16, 3, 5, 5}, "filter", false);
+  auto *bias = mod_.createPlaceholder(ElemKind::FloatTy, {16}, "bias", false);
+
+  auto *TN = F_->createTranspose("transpose", filter, NCHW2NHWC);
+  auto *CV = F_->createConv("conv", input, TN, bias,
+                            mod_.uniqueType(ElemKind::FloatTy, {1, 10, 20, 16}),
+                            5, 1, 2, 1);
+  auto *BN = F_->createBatchNormalization(ctx_, "batch", CV, 3, 0.0001, 0.9);
+  auto *ret = F_->createSave(ctx_, "ret", BN);
+
+  // Initialize to ensure that constant tensors are not optimized out.
+  ctx_.allocate(filter)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ctx_.allocate(bias)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ::glow::convertPlaceholdersToConstants(F_, ctx_, {input});
+
+  EXPECT_EQ(F_->getNodes().size(), 4);
+  ::glow::optimize(F_, CompilationMode::Infer);
+  EXPECT_EQ(F_->getNodes().size(), 2);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::BatchNormalizationNodeKind), 0);
+}
+
 /// Check that the batch normalization optimization is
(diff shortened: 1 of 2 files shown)


jfix71:
I tried running the unit test you added here without swapping the order of `optimizeTranspose` and `optimizeBatchNorm` and it still passed. Can you make sure the test exposes and tests the problem? I'm also a bit unsure what is preventing the Conv-BatchNorm merging from occurring. Can you explain a little more?

aturetsk:
@jfix71,
Fixed the test.

The problem is that optimizeBatchNorm expects convolution weights and bias to be Variables. If it's not so (for example, if there is a transpose node between convolution and weight variable), the optimization is not triggered. So if we run transpose constant folding before optimizeBatchNorm this issue can be avoided.

tests/unittests/graphOptzTest.cpp
```diff
@@ -161,6 +161,32 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  auto *ret = F_->createSave(ctx_, "ret", BN);
+
+  // Initialize to ensure that constant tensors are not optimized out.
+  ctx_.allocate(filter)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ctx_.allocate(bias)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ::glow::convertPlaceholdersToConstants(F_, ctx_, {});
+
+  EXPECT_EQ(F_->getNodes().size(), 4);
+  ::glow::optimize(F_, CompilationMode::Infer);
+  EXPECT_EQ(F_->getNodes().size(), 2);
```
jfix71: nit: Can you add a check that there are no BatchNorm nodes left? Can be simple using `countNodeKind`.

aturetsk:
@jfix71, thanks, done!

tests/unittests/graphOptzTest.cpp
```diff
@@ -161,6 +161,33 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
   EXPECT_TRUE(llvm::isa<SaveNode>(save));
 }
 
+/// Check that transpose constant folding is done before BatchNorm optimization,
+/// which allows to merge BatchNorm into Convolution with transposed weights.
+TEST_F(GraphOptz, optimizeBatchNormAfterConvWithTransposedWeights) {
+  Node *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 10, 20, 3}, "A", false);
+  auto *filter =
+      mod_.createPlaceholder(ElemKind::FloatTy, {16, 3, 5, 5}, "filter", true);
```
rdzhabarov: not sure why these are trainable :) any reason?

tests/unittests/graphOptzTest.cpp
```diff
@@ -161,6 +161,33 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  auto *CV = F_->createConv("conv", input, TN, bias,
+                            mod_.uniqueType(ElemKind::FloatTy, {1, 10, 20, 16}),
+                            5, 1, 2, 1);
+  auto *BN = F_->createBatchNormalization(ctx_, "batch", CV, 3, 0.0001, 0.9);
+  auto *ret = F_->createSave(ctx_, "ret", BN);
+
+  // Initialize to ensure that constant tensors are not optimized out.
+  ctx_.allocate(filter)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ctx_.allocate(bias)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  ::glow::convertPlaceholdersToConstants(F_, ctx_, {});
```
rdzhabarov: typically input is not converted to constant.

tests/unittests/graphOptzTest.cpp
aturetsk: No particular reason, the optimization should work with both.
Is there any reason to use un-trainable placeholder vs trainable in tests? Both kinds seem to be used randomly in tests...

rdzhabarov: it's just when I see that var is trainable I start to think whether there are any reasons behind that, or whether it has to be trainable.

aturetsk:
@rdzhabarov, made variables untrainable and included input in convertPlaceholdersToConstants exclusion list.

rdzhabarov:
it looks like "Merge branch 'master' into cadence/turetski/improve-batch-norm-opt" does not really add anything to commits here, so I'm going to merge this.

rdzhabarov:
thank you @aturetsk!

PR #1784, opened 2018-10-03
[include-bin] A tool to convert binary files to include-able char arrays

bertmaher:
*Description*: We need a tool to package binary data inside our libraries.  This tool, include-bin, is a stripped down implementation of `xxd -i`, since we can't depend on `xxd` existing on all our target platforms.
*Testing*: `ninja include-bin`
*Documentation*: N/A

### DIFF
--- a/tools/IncludeBin/IncludeBin.cpp
+++ b/tools/IncludeBin/IncludeBin.cpp
@@ -1,0 +1,59 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cstdio>
+#include <cstdlib>
+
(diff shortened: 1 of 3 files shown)


tools/IncludeBin/IncludeBin.cpp
```diff
@@ -0,0 +1,33 @@
+#include <cstdio>
+#include <cstdlib>
+
+int main(int argc, char **argv) {
+  if (argc != 3) {
+    fprintf(stderr, "Usage: bin-include input_file output_file\n");
+    exit(1);
+  }
```
qcolombet: If we wanted to simplify the handling of the argument, we could make this read from stdin and write to stdout.

tools/IncludeBin/IncludeBin.cpp
```diff
@@ -0,0 +1,33 @@
+#include <cstdio>
+#include <cstdlib>
+
+int main(int argc, char **argv) {
+  if (argc != 3) {
+    fprintf(stderr, "Usage: bin-include input_file output_file\n");
```
rdzhabarov: nit: since we are in c++, probably should switch to std::cout/std::cerr etc.

tools/IncludeBin/IncludeBin.cpp
```diff
@@ -0,0 +1,33 @@
+#include <cstdio>
+#include <cstdlib>
+
+int main(int argc, char **argv) {
```
rdzhabarov: can you put some doxygen comment here describing what's going on?

tools/IncludeBin/IncludeBin.cpp
bertmaher: Is there any real advantage to the std:: streams for something tiny like this?  I admit I <3 printf format strings and hate looking up the formatting functions for streams so I always printf if I think I can get away with it O:-).

PR #1785, opened 2018-10-03
[cpu] Compile libjit into CPU backend using include-bin

bertmaher:
Description: Glow will be easier to distribute if we can package libjit into the library, rather than keeping it as an external file dependency. This PR uses `include-bin` to compile libjit.bc into LLVMIRGen.
Testing: ninja check
Documentation: N/A

### DIFF
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -136,6 +136,15 @@ void LLVMIRGen::loadBaseAddresses(llvm::IRBuilder<> &builder) {
   offsetsArray_ = F->args().begin() + 3;
 }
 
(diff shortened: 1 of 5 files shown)


bertmaher:
This PR is logically stacked on #1784, please ignore the [include-bin] commit when reviewing.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -136,6 +136,11 @@ void LLVMIRGen::loadBaseAddresses(llvm::IRBuilder<> &builder) {
   offsetsArray_ = F->args().begin() + 3;
 }
 
+static const unsigned char libjit_bc[] = {
```
opti-mix: Add a comment here explaining where the content of libjit comes from or how it is generated.

PR #1787, opened 2018-10-03
[Caffe2] Add support for Copy, EnsureCPUOutput, EnsureDense operators

### LINKED ISSUE #1700: [New operator] No-ops: Copy, EnsureCPUOutput, EnsureDense
We need to support `Copy`, `EnsureCPUOutput`, `EnsureDense` operators from Caffe2. Most likely these operators won't do anything meaningful in Glow. Just "do nothing", similarly to how `Identity`, `Dropout`, `CopyCPUToMKL`, `CopyMKLToCPU` are implemented now.

Specs for these operators can be found here:
https://caffe2.ai/docs/operators-catalogue.html

SplitInfinity:
**Description**
This commit adds support for the `Copy`, `EnsureCPUOutput`, and `EnsureDense`
Caffe2 operators. These operators do not make much sense in the context
of Glow and have been implemented as no-ops.

**Testing**
All unit tests pass.

**Fixes**
This commit fixes #1700.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -453,8 +453,10 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
-  if (typeName == "CopyCPUToMKL" || typeName == "CopyMKLToCPU") {
-    // Glow does not support MKL now, just pass these two ops.
+  if (typeName == "CopyCPUToMKL" || typeName == "CopyMKLToCPU" ||
+      typeName == "Copy" || typeName == "EnsureCPUOutput" ||
+      typeName == "EnsureDense") {
+    // Glow does not support any of these ops now, so implement them as no-ops.
     auto in = getNodeValueOrCreateVariableByName(op.input(0));
     addNodeAsOutput(op, in);
     return;


lib/Importer/Caffe2.cpp
```diff
@@ -460,6 +460,27 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Copy") {
```
artemrakhov: Should we combine all these with MKL ops in a single if? There's zero chance that we will support any of these in the next 3 months.

SplitInfinity: Sure, that makes sense. Are `EnsureCPUOutput` and `EnsureDense` also MLK ops?

artemrakhov: No, they are not related to MKL.

lib/Importer/Caffe2.cpp
```diff
@@ -460,6 +460,27 @@ void caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+  }
+
+  if (typeName == "EnsureCPUOutput") {
+    // Glow does not support EnsureCPUOutput for now, implement this as a no-op.
+    auto in = getNodeValueOrCreateVariableByName(op.input(0));
+    addNodeAsOutput(op, in);
+    return;
+  }
+
+  if (typeName == "EnsureDense") {
```
nadavrot: typeName == "EnsureCPUOutput" || typeName == "EnsureDense" || typeName == ....

PR #1788, opened 2018-10-03
[Backend][Interpreter] Implement FP16 SoftMax

qcolombet:
*Description*
This commit templatizes the implementation of SoftMax and adds the
dispatch code to execute the single precision or half precision
variant for floating point types.

*Testing*
Added test case

Related to #1329

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -3692,6 +3692,27 @@ TEST_P(Operator, SoftMax) {
   EXPECT_TRUE(out.isEqual(*result, 0.001));
 }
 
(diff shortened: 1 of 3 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -3665,6 +3665,44 @@ TEST_P(InterpOnly, rowwiseQuantizedFCTest) {
   }
 }
 
+TEST_P(Operator, SoftMax) {
```
jfix71: nit: Comments here and for the other test?

qcolombet: Sure thing!

PR #1789, opened 2018-10-03
[Backend][Interpreter] Implement FP16 MatMul

qcolombet:
*Description*
This commit templatizes the implementation of MatMul and adds the
dispatch code to execute the single precision or half precision
variant for floating point types.

*Testing*
Added test case

Related to #1329

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -180,6 +180,27 @@ TEST_P(Operator, matmul) {
   EXPECT_NEAR(H.at({2, 0}), 95, 0.001);
 }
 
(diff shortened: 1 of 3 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -180,6 +180,26 @@ TEST_P(Operator, matmul) {
   EXPECT_NEAR(H.at({2, 0}), 95, 0.001);
 }
 
+TEST_P(InterpOnly, FP16Matmul) {
```
jfix71: nit: comment for what is being tested? It is kind of obvious from the test name...

qcolombet: Done!

PR #1790, opened 2018-10-04
CPU: change the way that we generate the JIT runtime

compnerd:
Use a custom linker language to change the linker invocation for the CPU
Runtime.  This allows us to properly track the dependencies and ensure that
Ninja does not rebuild the target unnecessarily.  As a small benefit, the linker
invocation is now also more portable.  The compiler rule still prevents the use
of the Makefile generator :-(.

*Description*:
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Backends/CPU/CMakeLists.txt
+++ b/lib/Backends/CPU/CMakeLists.txt
@@ -9,34 +9,39 @@ find_program(LLVM_LINK_BIN
                llvm-link-6.0
                llvm-link)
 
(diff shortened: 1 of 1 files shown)


compnerd:
lol, it names it `liblibjit.bc.a`.  I'll fix that.

bertmaher:
Still failing in CI b/c it can't find libjit :-(

compnerd:
Ah, its a location thing.  Checking the old build we used to write it to the root of the build tree.

PR #1797, opened 2018-10-04
[Placeholder] Remove the ability to save into variables.

nadavrot:
*Description*:

1.    [Placeholder] Remove the support for variables from SaveNode.

 2. [Placeholder] Remove the ability to create SaveNode that saves into a variable.
    
    This commit removes the SaveNode constructor that saves into a variable.
    It also fixes the lowering method and makes it create a placeholder for
    the GSUM storage.  It also renames createSaveNodePH into createSaveNode.
    
    At the moment we need to initialize the context after lowering because
    lowering creates new placeholders. When we migrate the context into the
    execution method we'll be able to move the initialization outside of the
    compilation code.

*Testing*: Ran ninja check and run.sh
*Documentation*: None.

#1334

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -203,9 +203,10 @@ void ExecutionEngine::optimizeFunction(CompilationMode mode, Function *F) {
   }
 }
 
-void ExecutionEngine::compile(CompilationMode mode, Function *F,
-                              const Context &ctx) {
+void ExecutionEngine::compile(CompilationMode mode, Function *F, Context &ctx) {
   optimizeFunction(mode, F);
+  // Make sure that the context has backing tensors for all placeholders.
+  ctx.allocate(M_.getPlaceholders());
   function_ = backend_->compile(F, ctx);
 }
 
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -70,7 +70,7 @@ public:
   /// target. This method should be invoked before the run method. The context
   /// \p ctx contains the mapping between symbolic values to concrete backing
   /// tensors.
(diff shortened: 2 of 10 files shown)


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -203,9 +203,10 @@ void ExecutionEngine::optimizeFunction(CompilationMode mode, Function *F) {
   }
 }
 
-void ExecutionEngine::compile(CompilationMode mode, Function *F,
-                              const Context &ctx) {
+void ExecutionEngine::compile(CompilationMode mode, Function *F, Context &ctx) {
   optimizeFunction(mode, F);
+  // Make sure that the context has backing tensors for all placeholders.
+  ctx.allocate(M_.getPlaceholders());
```
jfix71: Do we have some philosophy on correct usage for allocating Placeholders? Adding this call here means we could/will have inconsistent usage, where some users will allocate Placeholders prior to calling `compile()`, and others will just rely on this `allocate()` call here.

nadavrot: At the moment we are passing the context to the compile method. This means that the compile method has to remember the Placeholder buffers. The next step in this rewrite would be to move the context from the compile method to the execute method. This will allow multithreaded execution of compiled methods because the placeholders addresses won't be baked into the compiled function. So, initialization of the context has to happen in the test somewhere after lowering, but before execution. @jfix71

nadavrot: Moving the context parameter to the backends would require some work because at the moment the CPU backend assumes that it can use the pointers as constants. So, we'll need to adopt the AOT style compilation. To summarize, I'd like to move the line that I added here out side the compile method, but it will probably take a few weeks. @jfix71

rdzhabarov:
@nadavrot the check for "up to date" branch is gone, so it should be good to merge after squashing.

PR #1805, opened 2018-10-05
 Generate OpenCL kernel inc header files from original OpenCL kernel sources directly using the new include-bin utility

opti-mix:
*Description*:
It is now possible to treat the OpenCL sources as real C source files in the editors, formatting tools, etc.

It should also solve issues with compilers which could not handle raw string literals that are too long.

Also take the opportunity to make format.sh run on the OpenCL sources we have.

*Testing*:
ninja test

### DIFF
--- a/lib/Backends/OpenCL/CMakeLists.txt
+++ b/lib/Backends/OpenCL/CMakeLists.txt
@@ -1,4 +1,30 @@
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/glow/kernels.inc"
+  COMMAND include-bin 
+          "${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl"
+          "${CMAKE_BINARY_DIR}/glow/kernels.inc"
+  DEPENDS include-bin CPURuntime "${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl")
+
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/glow/kernels_fwd_conv.inc"
+  COMMAND include-bin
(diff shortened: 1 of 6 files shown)


lib/Backends/OpenCL/CMakeLists.txt
```diff
@@ -1,4 +1,30 @@
...
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/glow/kernels_fwd_conv.inc"
+  COMMAND include-bin
+          "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_conv.cl"
+          "${CMAKE_BINARY_DIR}/glow/kernels_fwd_conv.inc"
+  DEPENDS include-bin CPURuntime
+          "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_conv.cl")
+
+add_custom_command(
+  OUTPUT "${CMAKE_BINARY_DIR}/glow/kernels_fwd_quantized_conv.inc"
```
nadavrot: Would it make sense to let the xxd utility take multiple arguments and just concat them into a single array?

opti-mix: Yeah, I think it could be useful.

opti-mix: But for our OpenCL use of this utility we don't need it, because we compile those string arrays using the OpenCL driver at different points during the execution.

opti-mix:
@lygstate Could you check if this PR solves the issues you had with OpenCL sources and Visual Studio?

lib/Backends/OpenCL/CMakeLists.txt
nadavrot: Okay.

bertmaher:
Awesome :-D

PR #1809, opened 2018-10-08
[doc] Update the docs and mention constants and placeholders.

nadavrot:
*Description*:

[doc] Update the docs and mention constants and placeholders.

*Testing*: NFC.

*Documentation*: This is a doc change.

#1334

### DIFF
--- a/docs/IR.md
+++ b/docs/IR.md
@@ -18,32 +18,33 @@ to a graph that you may find inside Caffe or in ONNX format. When we load a
 neural network model from some file we construct this graph with a direct
 translation of one operator to one or more nodes. The high-level IR is a simple
 graph that allows basic transformations such as replacing all uses of some node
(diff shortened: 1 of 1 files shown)


docs/IR.md
```diff
@@ -85,42 +86,30 @@ lowered into the mid-level IR in a phase that's called "IRGen" (stands for IR
 generation). This is a one-to-many translation where each operator is translated
 into one or more instructions.
 
-### Variable Visibility
```
jfix71: We still have WeightVar Visibility -- should we keep a section about this? Actually, maybe we can remove Visibility from WeightVar too (and so remove it entirely)? I don't think there's any difference at this point between Mutability vs. Visibility  -- I believe Constant is always Private, and Mutable is always Public.

nadavrot: Good idea. I submitted #1811 to remove the visibility.

PR #1810, opened 2018-10-08
Support autoIRGen for nodes with multiple results in InstrGen

opti-mix:
*Description*:
Support autoIRGen for nodes with multiple results in InstrGen. 
This functionality may be needed for some of our backends.

*Testing*:
ninja test


opti-mix:
@jfix71 
> though it would be nice to have this tested somehow

Yes, I actually wanted to have it tested somehow. But it was not clear to me how to add such a test.

opti-mix:
@nadavrot Added more comments to the new and some old sections to make clear what is going on.

PR #1812, opened 2018-10-08
[docs | NFC] Update docs reflecting changes to Variables and Context

jfix71:


jfix71:
@nadavrot Yeah that might make more sense. The nice thing with this doc is it shows an end to end example with graph visualization, IR, etc. @bertmaher you added this, right? Did you take it from actual code somewhere in our unit tests that we could point to instead of including code here?

Issue #1517, opened 2018-08-23
[RFC] Error reporting, malformed graph, and general resilience


qcolombet:
### Context

Glow checks are for the most part bound to whether or not asserts are enabled in the build. 
The rationale is we expect the inputs (network, weights) to be correctly shaped and sized when we invoke the compiler/loaders and thus this design eliminates unnecessary overhead.

### Problems

Given most checks are removed in release mode, we end up in a situation where we cannot use this variant when bringing up new works.

For instance, silly mistakes like shaping incorrectly a transpose will produce random compiler errors or just invalid/inaccurate outputs with no way to tell what is going wrong.
Then, running the same network with a debug variant would trigger a verifier failure from the beginning, hinting directly at what the problem was. This process is far from being intuitive and is cumbersome.

In essence, we have the following problems:
- The debug mode gives us diagnostics but is slow to run.
- The release mode is fast to run but does not gives any information on what is going on.
- The error reporting through asserts has too little context to be actionable on its own.

To make a comparison, Glow behaves like a compiler that wouldn't issue a syntax error diagnostic if compiled with assert disabled. On top of that, when the assertions are enabled, errors like type mismatch would be reported bluntly: no location in the source file, no naming of the type that was causing the problem and so on.

In a nutshell, when working on a new model, the process may look like this:
- Run the network with the release variant
=> compiler crashes
- Run the network with the debug variant
=> assertion failure "invalid..."
- Run the network with the debug variant in a debugger
=> assertion failure "invalid...", then print variables, examine the memory, etc. to see what is wrong.

Finally, for large networks, it is possible that the debug mode is too slow to be practical, making debugging very hard.

### Ultimate Goal

Ideally, we would like to be sure that only valid IR hit the compiler. Then, in the compiler itself, it seems reasonable to rely on asserts for things related to the IR, like the invariants after such and such optimizations and so on. Indeed, at this point, any problem occurring in the compiler is the result of a wrong doing of the compiler itself and thus not actionable by the user.

Of course, we would still need to have some kind of non-assert based error reporting for anything that is not in the control of the compiler. E.g., failing to write to the filesystem, out-of-memory allocation, network too big for the targeted backend, etc.

### How Do We Get There?

We basically have two approaches to make sure only valid IR hit the compiler:
1. Have every "frontend" of Glow performs the checks (e.g., like the protobuf loaders).
2. Have Glow performs the checks before doing anything.

[...]

PR #1813, opened 2018-10-08
[Graph] Fix verify one writer

jfix71:
*Description*: This verify pass was broken during the Placeholder migration. I moved to a `dyn_cast` instead of `isa` followed by `cast` to make it less error prone.
*Testing*: Added a unit test so that it won't be accidentally broken again in the future.
*Documentation*: N/A

### DIFF
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -1068,3 +1068,25 @@ TEST(Graph, setType) {
   EXPECT_EQ(topK->getType(1), origTopKRes1);
   EXPECT_EQ(valRes1.getType(), origTopKRes1);
 }
+
+/// Tests that expect death from the verifier cannot currently run in Release
+/// mode as they would not die, since the verifier uses assertions for
+/// verification. Once the verifier moves to returning false instead of aborting
+/// (GH issue #1517), this can be removed and EXPECT_DEATH can be replaced by
+/// EXPECT_FALSE.
+#ifndef NDEBUG
+
+/// Check that verify doesn't allow for multiple writers to the same node.
+TEST(Graph, verifyOneWriter) {
+  Module M;
+  auto *F = M.createFunction("main");
+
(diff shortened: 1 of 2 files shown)


tests/unittests/graphTest.cpp
```diff
@@ -1068,3 +1068,16 @@ TEST(Graph, setType) {
...
+TEST(Graph, verifyOneWriter) {
+  Module M;
+  auto *F = M.createFunction("main");
+
+  auto *input = M.createPlaceholder(ElemKind::FloatTy, {5}, "input", false);
+  auto *output = M.createPlaceholder(ElemKind::FloatTy, {5}, "output", false);
+  F->createSave("Save1", input, output);
+  F->createSave("Save2", input, output);
+
+  EXPECT_DEATH(M.verify(), "");
```
qcolombet: This check won't work in release mode.
The verifier does nothing in that mode (and that's sad, and I'd like we change that see https://github.com/pytorch/glow/issues/1517#issuecomment-415549954, but we're not there yet).

jfix71: Ah, true. Any suggestion on how to proceed here? We could wrap this test with `NDEBUG` I suppose...

qcolombet: Sounds reasonable to me and unfortunately I don't have any other suggestions (aside what I said in #1517 :)).

jfix71: Done!

tests/unittests/graphTest.cpp
```diff
@@ -1068,3 +1068,25 @@ TEST(Graph, setType) {
   EXPECT_EQ(topK->getType(1), origTopKRes1);
   EXPECT_EQ(valRes1.getType(), origTopKRes1);
 }
+
+/// Tests that expect death from the verifier cannot currently run in Release
+/// mode as they would not die, since the verified uses assertions for
```
qcolombet: nit: verified -> verifier

PR #1814, opened 2018-10-08
[Graph] Verify that Constants are never written to

jfix71:
*Description*: Constants should never be written to, so I'm adding a verification pass checking this.
*Testing*: Added a test.
*Documentation*: N/A

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -2289,9 +2289,6 @@ void Function::verify() const {
     llvm_unreachable("Multiple nodes with the same name");
   }
 
-  const auto &vars = getParent()->getConstants();
-  (void)vars;
-
   // Any node referenced by one of the graph nodes should be part of the Graph.
   for (const auto &N : nodes_) {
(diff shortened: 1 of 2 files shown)


lib/Graph/Graph.cpp
```diff
@@ -2258,6 +2258,24 @@ static bool isGraphStorageNode(Node *n, const Function *F) {
...
+      if (node == N->getNthResult(i))
+        return true;
+    // See if node is used as an overwritten input anywhere.
+    for (unsigned i = 0; i < N->getNumInputs(); i++)
+      if (N->isOverwrittenNthInput(i))
+        if (node == N->getNthInput(i))
+          return true;
+  }
+  return false;
+}
```
qcolombet: Instead of adding this method, I believe we could modify the loop that check the placeholders and fail when we see the first use.
I am referring to the loop that start with:
 // Make sure all the placeholders are at most written once.

We could remove the:
      if (!isa<Constant>(input)) {
        continue;
      }

and adapt and the following assert:
 if (varToFirstDef != placeholderWrittenTo.end()) {

PR #1815, opened 2018-10-08
[third_party] Update googletest submodule again

artemrakhov:
My previous attempt to update was accidentally rolled back. Updating again.
I am not sure how can I prevent a downgrade from happening again.
Please don't approve a PR if it has changes to `tests/googletest`.
Maybe the downgrade happened because of GitHub's automatic rebase?

*Description*:
Update googletest submodule to newer version (branch 1.8.x latest).

*Testing*:
`ninja test`

*Documentation*:
Previous version generated a lot of warnings when built with g++. These are all fixed in the current version.
In order to update, I ran the following:
```
$ cd tests/googletest
$ git checkout origin/v1.8.x
```


bertmaher:
I find it really hard to notice when a submodule change has slipped into my PR.  It shows up as one line in the diff (just a commit hash).  I've solved the problem for myself by creating a bash alias that always does a submodule update when I rebase my diffs, but it's going to be an ongoing pain :-(

qcolombet:
Agree with @bertmaher, I've already been bitten by that same problem.
Luckily the reviewers of my PR caught it, but I am afraid it won't always be the case.

Maybe we could add in our `ninja check` a test that check that the submodules have a specific hash. If we update (or forget to update) the submodule this test will fail, if we import some changes in  the submodules we have to update this test explicitly.  (Basically let us automate Bert's solution.)

PR #1820, opened 2018-10-09
[Graph&IR] Add a ConvertTo node

qcolombet:
*Description*
This node/instruction represents a type conversion (a.k.a. a cast)
from the input type to the output type.
The shape of the input and output type must be identical, only the
element type is allowed to change.

This node will allow us to represent type conversions when converting
a graph from one type to another.

This patch adds the support for this node in the interpreter and
teaches the differentiation mechanism how to deal with ConvertTo
node.

Note: We could technically introduce different nodes for the different
conversion (e.g., fpext, fptrunc, inttofp, and so on) since this
node would have different interpretation based on the input and output
types. Right now, we decided not to do that because we expect this
node to not hit any backends. In particular, this node is only
accessible through the C++ API.

*Testing*
Added tests.

Related to #1312.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -299,6 +299,17 @@ void ConvolutionGradNode::verify() const {
                     Kernels_, Strides_, Pads_, Group_);
 }
 
+void ConvertToNode::verify() const {
+  assert(getInput().dims() == getResult().dims() && "Shape must be the same");
+  TypeRef srcTy = getInput().getType();
+  (void)srcTy;
+  TypeRef dstTy = getResult().getType();
+  (void)dstTy;
+  assert(srcTy != dstTy && "Nothing to convert");
+  assert(!srcTy->isQuantizedType() && !dstTy->isQuantizedType() &&
+         "Quantized conversion should use Dequantize, Quantize and Rescale");
+}
+
 void MaxPoolNode::verify() const {
   verifyPool(getInput(), getResult(), Kernels_, Strides_, Pads_);
 }
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -4124,6 +4124,50 @@ TEST_P(Operator, QuantizeSimple) {
   EXPECT_NEAR(RH.at({0, 0}), 21.0, 0.001);
 }
 
+/// Check that convertTo node works properly from float16_t to float.
+TEST_P(InterpOnly, ConvertFromFloat16ToFloat) {
+  auto *A = mod_.createPlaceholder(ElemKind::FloatTy, {20, 13}, "A", false);
+  auto inputHandle = ctx_.allocate(A)->getHandle<float>();
+  inputHandle.randomize(-3.0, 3.0, mod_.getPRNG());
+
+  TypeRef outTy = mod_.uniqueType(ElemKind::Float16Ty, A->dims());
+
+  auto *convertTo = F_->createConvertTo("convertTo", A, outTy);
+  auto *result = F_->createSave("save", convertTo);
+  ctx_.allocate(result->getPlaceholder());
+
+  EE_.compile(CompilationMode::Infer, F_, ctx_);
+  EE_.run();
+
+  auto *outputTensor = ctx_.get(result->getPlaceholder());
+  Tensor convertedInput = ctx_.get(A)->clone();
+  convertedInput.convertToType(ElemKind::Float16Ty);
+  ASSERT_EQ(outputTensor->size(), inputHandle.size());
+  EXPECT_TRUE(convertedInput.isEqual(*outputTensor));
+}
+
+/// Check that convertTo node works properly from float to float16_t.
+TEST_P(InterpOnly, ConvertFromFloatToFloat16) {
+  auto *A = mod_.createPlaceholder(ElemKind::Float16Ty, {20, 13}, "A", false);
+  auto inputHandle = ctx_.allocate(A)->getHandle<float16_t>();
+  inputHandle.randomize(-3.0, 3.0, mod_.getPRNG());
+
+  TypeRef outTy = mod_.uniqueType(ElemKind::FloatTy, A->dims());
+
+  auto *convertTo = F_->createConvertTo("convertTo", A, outTy);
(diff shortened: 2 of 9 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -4124,6 +4124,50 @@ TEST_P(Operator, QuantizeSimple) {
...
+
+  auto *outputTensor = ctx_.get(result->getPlaceholder());
+  Tensor convertedInput = ctx_.get(A)->clone();
+  convertedInput.convertToType(ElemKind::Float16Ty);
+  ASSERT_EQ(outputTensor->size(), inputHandle.size());
+  EXPECT_TRUE(convertedInput.isEqual(*outputTensor));
+}
+
+/// Check that convertTo node works properly from float to float16_t.
+TEST_P(InterpOnly, ConvertToFloatToFloat16) {
```
jfix71: nit: Same -- should this be `ConvertFromFloatToFloat16 `?

tests/unittests/graphGradTest.cpp
```diff
@@ -181,3 +181,30 @@ TEST(GraphAutoGrad, checkPlaceholderGradTest) {
...
+  Context ctx;
+
+  // Construct the network:
+  TC.learningRate = 0.001;
+
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  auto *A = mod.createPlaceholder(ElemKind::FloatTy, {20, 13}, "A", false);
+  auto inputHandle = ctx.allocate(A)->getHandle<float_t>();
```
jfix71: Why `float_t` instead of `float`?

tests/unittests/OperatorTest.cpp
```diff
@@ -4124,6 +4124,50 @@ TEST_P(Operator, QuantizeSimple) {
   EXPECT_NEAR(RH.at({0, 0}), 21.0, 0.001);
 }
 
+/// Check that convertTo node works properly from float16_t to float.
+TEST_P(InterpOnly, ConvertToFloat16ToFloat) {
+  auto *A = mod_.createPlaceholder(ElemKind::FloatTy, {20, 13}, "A", false);
+  auto inputHandle = ctx_.allocate(A)->getHandle<float_t>();
```
jfix71: Why `float_t` instead of `float`?

lib/Graph/Nodes.cpp
```diff
@@ -299,6 +299,17 @@ void ConvolutionGradNode::verify() const {
...
 
+void ConvertToNode::verify() const {
+  assert(getInput().dims() == getResult().dims() && "Shape must be the same");
+  TypeRef srcTy = getInput().getType();
+  (void)srcTy;
+  TypeRef dstTy = getResult().getType();
+  (void)dstTy;
+  assert(srcTy != dstTy && "Nothing to convert");
+  assert(!srcTy->isQuantizedType() && !dstTy->isQuantizedType() &&
+         "Quantized conversion should use Dequantize, Quantize and Rescale");
```
jfix71: Why don't we want to use this for quantization?

tests/unittests/OperatorTest.cpp
```diff
@@ -4124,6 +4124,50 @@ TEST_P(Operator, QuantizeSimple) {
   EXPECT_NEAR(RH.at({0, 0}), 21.0, 0.001);
 }
 
+/// Check that convertTo node works properly from float16_t to float.
+TEST_P(InterpOnly, ConvertToFloat16ToFloat) {
```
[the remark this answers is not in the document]
qcolombet: The idea behind the name of the test was NodeNameSrcTyToDstTy.
I didn't realize that NodeName (ConvertTo) would result in this poor test name x).

Anyhow, ConvertFromXXX sounds good.

tests/unittests/OperatorTest.cpp
qcolombet: Ditto

tests/unittests/graphGradTest.cpp
qcolombet: Good catch!
Editing issue :).
Switch `float16_t` into `float`, but miss the `_t`. I am actually surprised it compiled.

tests/unittests/OperatorTest.cpp
qcolombet: Ditto.

tests/unittests/graphGradTest.cpp
qcolombet: Done

tests/unittests/OperatorTest.cpp
qcolombet: Done

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
qcolombet: Done

tests/unittests/OperatorTest.cpp
qcolombet: Done

lib/Graph/Nodes.cpp
jfix71: Fair enough. One thing then is that `ConvertTo` is named somewhat generically, like it could apply to `Quantize`/`Dequantize`. Not a big deal but led me to believe it could be used on Quantized types. Perhaps the docstring can specify that at least?

Also, if we add other quantization bit widths, would we extend `Quantize`/`Dequantize` in the same way this node behaves (switching between types to determine what to do)? Would `Rescale` move between these different quantization widths? Maybe it's out of scope here, but seems related.

qcolombet: > Perhaps the docstring can specify that at least?

I'll post a patch in that direction.

> Also, if we add other quantization bit widths, would we extend Quantize/Dequantize in the same way this node behaves (switching between types to determine what to do)? 

I assume your talking about the implementation in the backends, if that's the case, I would expect so. Now, I am not a huge fan of that approach and I suggested ways to alleviate it in the compiler (optimization and interpreter, however, codegen would still need to go through that pain) see https://github.com/pytorch/glow/issues/1329#issuecomment-425284818, but that's the status we are now and given we don't add a type very often we might continue like this for quite some time :).

PR #1821, opened 2018-10-09
[Converter] Introduce a FunctionConverter helper class

qcolombet:
*Description*
Converting some operators of a function from one type to another,
for instance single precision floating point to half precision
floating point, single precision floating point to quantized type,
and son on, does not depend on the source and destination types.
Therefore, it is possible to share the same core framework for all
such conversion.

This patch introduces a new helper class FunctionConverter that
realizes this concept. In particular, it features the main conversion
procedure (FunctionConverter::convert) and relies on the specialization
of the class to implement the type specific bit like how to materialize
the conversion from the source type to the destination type, what type
should be used for a specific value and so on.

The main conversion procedure works like this:
for each node in function:
  check if the node need to be converted
  insert conversions for the inputs of node
  update the inputs of node to use the results of the conversions
  mutate the type of the outputs of the node
  insert conversions for the outputs of the node
  morph node
  postProcessing node
cleanUp

After each conversion step, we preserve the invariant that the input
and output types of "node" are untouched.
E.g.,
Before converting node:
res(outTy) = node arg1(in2Ty), arg2(in2Ty)

After converting "node", we will have something that looks like:
newArg1(convertedIn1Ty) = conversion arg1
newArg2(convertedIn2Ty) = conversion arg2
newRes(convertedOutTy) = node newArg1, newArg2
res(outTy) = conversion newRes

We will eventually optimize the useless intermediate conversions but
this is outside of the scope of this change.

*Testing*
N/A, this is an abstract class and this patch does not include any
user.

### DIFF
--- a/include/glow/Converter/FunctionConverter.h
+++ b/include/glow/Converter/FunctionConverter.h
@@ -1,0 +1,166 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


qcolombet:
This class can be seen in action with #1822 for quantization and #1823 for general type conversions.

PR #1822, opened 2018-10-09
[Quantization] Specialize the FunctionConverter to quantize a function

qcolombet:
Note: The diff will be much more approachable when #1821 lands, since this PR includes both #1821 and the change for this refactoring. Reviewing just the last commit of this PR would have the same effect as waiting for #1821 to be merged.

*Description*
This patch refactors the code of the function that quantizes a function
to reuse the framework set by the FunctionConverter.

This patch is not strictly speaking NFC because unlike the original
implementation the conversion process does not leave the original,
not use after conversion, node around. Other than that, it should
be NFC.

*Testing*
Ran the quantizationTest
Ran different quantized profiles (asymmetric, symmetric, symmetric
with uint8) with and without the patch for resnet50 and checked
that the output IR were identical (modulo some variable names).

Related to #1329, we provide a unified way for converting function
that fp16 is going to use as discussed in #1747.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -16,6 +16,7 @@
 
 #include "glow/Quantization/Quantization.h"
 
+#include "glow/Converter/FunctionConverter.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 
 #include <cmath>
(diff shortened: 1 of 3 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -27,6 +28,112 @@ using llvm::cast;
...
+
+    const TensorQuantizationParams &TQP = valTQPIt->second;
+    return mod_.uniqueType(ElemKind::Int8QTy, val.dims(), TQP.scale,
+                           TQP.offset);
+  }
+
+  /// \see FunctionConverter::canConvert.
+  /// Only convert nodes that use floating point types and that
+  /// weren't specifically marked as to-ignore with doNotQuantizeKinds_.
+  bool canConvert(const Node &node) const override;
```
bertmaher: Any particular reason this (and morphNode and postProcessing) are defined out-of-line?  It would be nice to be consistent with that.

lib/Quantization/Quantization.cpp
```diff
@@ -27,6 +28,112 @@ using llvm::cast;
...
+
+  /// Shortcut to the module of function_.
+  Module &mod_;
+  /// Execution engine used to check is a quantized operator is
+  /// supported.
+  const ExecutionEngine &EE_;
+  /// Set of node kinds that should not be quantized.
+  const KindSet &doNotQuantizeKinds_;
+  /// Map the (name of a node, idx) to its quantization parameters.
+  std::unordered_map<std::string, TensorQuantizationParams> nodeToTQP_;
```
bertmaher: Can these members be `private`?

lib/Quantization/Quantization.cpp
```diff
@@ -111,333 +192,146 @@ static bool canBeQuantized(const Node *node) {
...
-    auto QT =
-        F->getParent()->uniqueType(ElemKind::Int8QTy, FC->getResult().dims(),
-                                   qParams[0].scale, qParams[0].offset);
-    quantizedNode =
-        F->createFullyConnected(FC->getName(), quantizedInputs[0],
-                                quantizedInputs[1], quantizedInputs[2], QT);
-    break;
-  }
+Node &FunctionQuantizer::morphNode(Node &node) {
+  // FIXME: Right now, the TensorQuantizationParams only tracks one
```
bertmaher: Could you create a GH issue to track this FIXME, if you haven't already?

lib/Quantization/Quantization.cpp
```diff
@@ -111,333 +192,146 @@ static bool canBeQuantized(const Node *node) {
...
+  // The following switch fixes outTy for the few nodes where inTy and outTy
+  // can disagree.
+  // This happens for cases where for instance, the quantized parameters for the
+  // output have a different scale than the input, whereas the operation itself
+  // doesn't allow that.
+  //
+  // E.g., the constraints for `outTy = op inTy` are `outTy == inTy`, but the
+  // quantization profiling gave different types to outTy and inTy.
+  switch (node.getKind()) {
+  default:
```
bertmaher: Nit: `default` at the beginning feels kind of weird, since I think it's at the end of the switch everywhere else.

lib/Quantization/Quantization.cpp
qcolombet: Not really, it was more a judgement call like "hmm, those are going to be big". Turns out they are not really.

lib/Quantization/Quantization.cpp
qcolombet: Sure!

lib/Quantization/Quantization.cpp
qcolombet: Sure! I was wanting for @rdzhabarov to chime in to see what is his opinion for that, but we could and should definitely track that in an issue.

lib/Quantization/Quantization.cpp
qcolombet: Sure!

lib/Quantization/Quantization.cpp
qcolombet: Done

lib/Quantization/Quantization.cpp
qcolombet: Done

lib/Quantization/Quantization.cpp
qcolombet: Done

lib/Quantization/Quantization.cpp
qcolombet: That's #1832.

PR #1825, opened 2018-10-09
[unittests] New BackendCorrectnessTest API enabling easy quantization

jfix71:
*Description*: Our BackendCorrectnessTests are all float. We ideally would also test them quantized. It is preferable not to have to rewrite them all quantized. This PR intends to provide a better API for constructing a Function, optionally profiling/quantizing it, and then comparing its output to the Interpreter. For now I only did this for `BasicFCNet` as a trial -- if it's agreed this is a good direction I will convert over the other BackendCorrectnessTests in a future PR.

*Testing*: ninja test

*Documentation*: N/A

### DIFF
--- a/tests/unittests/BackendCorrectnessTest.cpp
+++ b/tests/unittests/BackendCorrectnessTest.cpp
@@ -22,6 +22,7 @@
 #include "glow/IR/IR.h"
 #include "glow/IR/IRBuilder.h"
 #include "glow/IR/Instrs.h"
+#include "glow/Quantization/Quantization.h"
 #include "glow/Support/Random.h"
 
 #include "gtest/gtest.h"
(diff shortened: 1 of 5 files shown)


tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -38,6 +39,68 @@ class BackendCorrectnessTest : public ::testing::TestWithParam<BackendKind> {
...
 class CPUOnly : public BackendCorrectnessTest {};
 
+namespace {
+
+/// Clone, profile, and run \p origF given the \p ctx and \p EE. \returns the
+/// quantization parameters from the profile.
+static std::vector<NodeQuantizationInfo>
+profileAndGetNodeQuantizationInfo(Context &ctx, ExecutionEngine &EE,
+                                  Function *origF) {
+  Function *profileF = origF->clone("profileF");
```
rdzhabarov: you do not really need to clone this F as it'll be taken care of as part of the ```profileQuantization```

jfix71: Thanks, done!

nadavrot:
Nice!

PR #1838, opened 2018-10-10
[unittests] Increase allowed error threshold while comparing floating point tensors in BackendCorrectnessTest

aturetsk:
*Description*:
I've got the following test failure while building Glow in Release mode (Debug is OK):

`[ RUN      ] CPU/BackendCorrectnessTest.basicFCNet/0`
`tests/unittests/BackendCorrectnessTest.cpp:499: Failure`
`Value of: compareAgainstInterpreter(GetParam(), createAndInitBasicFCNet, false)`
`  Actual: false`
`Expected: true`
`[  FAILED  ] CPU/BackendCorrectnessTest.basicFCNet/0, where GetParam() = 4-byte object <02-00 00-00> (3075 ms)`

It looks like accumulated floating point error exceeds allowed threshold. Increasing the threshold from 0.0001 to 0.0002 makes the test pass.
Fast-math is to blame, as just disabling it helped as well.

*Testing*:
BackendCorrectnessTest.

*Documentation*:
N/A


aturetsk:
@jfix71 , added an optional argument to `compareAgainstInterpreter()` for allowed error threshold.
Default value for this argument is equal to the default value in `isEqual` function.
Modified the allowed error only for the floating point test, the quantized test keeps using default value as before.

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -68,7 +68,7 @@ using CreateAndInitFunction =
 static bool
 compareAgainstInterpreter(BackendKind backendKind,
                           CreateAndInitFunction createAndInitFunction,
-                          bool quantize) {
+                          bool quantize, float allowerError = 0.0001) {
```
jfix71: nit: Assuming this was a typo -- please make this `allowedError`

aturetsk:
Ups, yes, this was a typo.
Thanks for review!

aturetsk:
Could please somebody land it?

PR #1841, opened 2018-10-11
[ONNXIFI] Don't mangle the name of ONNXIFI interface functions

yinghai:
*Description*:
We don't want to mangle the name of ONNXIFI interface or dlopen will fail. 

*Testing*:
Build onnxifi-glow.so and use `nm` to verify that names are not mangled. 
*Documentation*:
[Optional Fixes #issue]


nadavrot:
Nice!

PR #1842, opened 2018-10-11
[ONNXIFI] Fix resnet50 integration test

yinghai:
*Description*:
The assumption that ONNX model only contains one node doesn't hold because Caffe2 op might be exported into multiple ONNX nodes. 
*Testing*:
Run the ONNXIFI integration test from Pytorch repo:
pytest python/onnx/test_onnxifi.py
And it should pass. 

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Importer/ONNXIFIModelLoader.cpp
+++ b/lib/Importer/ONNXIFIModelLoader.cpp
@@ -146,55 +146,50 @@ ONNXIFIModelLoader::parseOperator(const void *onnxModel, size_t onnxModelSize) {
 
   ONNX_NAMESPACE::GraphProto graph = modelDef.graph();
 
-  // Only single operator is allowed to be in the onnxModel.
-  if (graph.node_size() != 1) {
-    return result;
-  }
-
-  const std::string &operation = graph.node(0).op_type();
-
 #define ADD_OP_MAPPING(NODE_KIND_, ELEM_KIND_)                                 \
   result.emplace_back(Kinded::Kind::NODE_KIND_, ElemKind::ELEM_KIND_);
-
-  // Single ONNX node can be represented by several Glow nodes,
-  // collect corresponding mapping in result vector.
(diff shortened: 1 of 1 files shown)


rdzhabarov:
Please, run clang-format on the changes and I'll merge after that.

yinghai:
What's wrong with my clang-format? It fails the clang format check....

lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -146,55 +146,51 @@ ONNXIFIModelLoader::parseOperator(const void *onnxModel, size_t onnxModelSize) {
 
   ONNX_NAMESPACE::GraphProto graph = modelDef.graph();
 
-  // Only single operator is allowed to be in the onnxModel.
-  if (graph.node_size() != 1) {
-    return result;
-  }
-
-  const std::string &operation = graph.node(0).op_type();
```
rdzhabarov: two empty lines here

rdzhabarov: you can run: ```./utils/format.sh fix```

yinghai:
> Why did it fail in the first place? I assume returning "non supported" operator should be properly handled in the graph-cutting.

I believe these lines were added later...

rdzhabarov:
> I believe these lines were added later...

I meant to say that if ONNXIFI backend reports that the operator (in this case several operators) is not supported it should not cause the failure we are seeing now in the integration test.
I worry if there is a bigger problem which is hidden now by returning "operator is supported".

yinghai:
> I meant to say that if ONNXIFI backend reports that the operator (in this case several operators) is not supported it should not cause the failure we are seeing now in the integration test.

This is because the issue of the test itself. It uses input(0) as tensor to feed but in this case it's not the tensor we want hence the failure.

PR #1847, opened 2018-10-11
Disable CPURuntimeNative for MSVC

lygstate:
*Description*:
Disable CPURuntimeNative for MSVC
because libjit can not compiled with cl.exe

Related to  https://github.com/pytorch/glow/issues/1144


bertmaher:
What's the problem with libjit?  Is it the vector extensions, or something else?

lygstate:
vector extensions and vla and other gcc attributes

bertmaher:
That makes sense.  I'm fine with this change then, it's just one test.

PR #1849, opened 2018-10-11
Fixes CPURuntime for MSVC.

lygstate:
*Description*:
Fixes CPURuntime for MSVC using custom cmake commands.

Related to https://github.com/pytorch/glow/issues/1144

### DIFF
--- a/lib/Backends/CPU/CMakeLists.txt
+++ b/lib/Backends/CPU/CMakeLists.txt
@@ -9,43 +9,34 @@ find_program(LLVM_LINK_BIN
                llvm-link-6.0
                llvm-link)
 
-set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
-set(CMAKE_LLIR_CREATE_SHARED_MODULE "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
-set(CMAKE_LLIR_CREATE_STATIC_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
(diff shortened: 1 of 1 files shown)


lib/Backends/CPU/CMakeLists.txt
```diff
@@ -9,43 +9,34 @@ find_program(LLVM_LINK_BIN
...
-                        PREFIX ""
-                        SUFFIX ""
-                        RULE_LAUNCH_COMPILE "${CLANG_BIN} <DEFINES> <INCLUDES> <FLAGS> -o <OBJECT> -c <SOURCE> #")
-target_compile_options(CPURuntime
-                       PRIVATE
-                         -ffast-math
-                         -g0
-                         -emit-llvm
-                         -O0)
+add_custom_target(CPURuntime
```
bertmaher: Using `add_custom_target` means that CPURuntime will always get rebuilt, even if its deps have not changed.  Please try to find a way around that.

ayermolo: I don't think it's quite correct. the CPURuntime will be evaluated, but it won't rebuild the .o files unless there is some change. Like they have been deleted., or the files that are compiles are modified. At least this is the behavior I am seeing on windows.

bertmaher:
This PR is over a month old and possibly abandoned, so I'm going to close it out for now.  Feel free to reopen or open a new one with reviewers' comments addressed.  Thanks!

lib/Backends/CPU/CMakeLists.txt
lygstate: @bertmaher Do you have a look at ayemolo's comment? There is no other way to building CPURuntime for Windows.

Issue #1727, opened 2018-09-27
Support directly loading a quantized model.


beicy:
Currently, the only way to run a quantized model is to do the profiling and reload it. We would like to support to directly load a quantized model. Since there is no quantized ONNX op support yet, we will start with Caffe2 loader. More details will be updated here later.

@qcolombet

yinghai:
@Maratyszcza what's the progress of ONNX support?

Maratyszcza:
no ONNX quantization spec yet

beicy:
Now we started it with supporting quantized ResNet50 : https://github.com/caffe2/models/tree/master/resnet50_quantized
To support this model, 
**1) the following ops need to be added:**
```
1. Int8Quantize
2. Int8Dequantize
3. Int8Conv
4. Int8ConvRelu
5. Int8MaxPool
6. Int8AveragePool
7. Int8FC
8. Int8SumRelu
9. Int8GivenIntTensorFill  -- This is int32 quantized data.
10. Int8GivenTensorFill     -- This is int8 quantized data.
```
The main difference of the above ops from the regular ops is two args added: 
```
arg {                                                                                                                                                               
      name: "Y_scale"                                                                                                                                                     
      f: 0.00044428                                                                                                                                                       
      }                                                                                                                                                                  
arg {                                                                                                                                                               
      name: "Y_zero_point"   --- this one is offset                                                                                                                                               
      i: 128                                                                                                                                                              
     } 
```    
**2) When a node is created with quantized value, we need to pass the scale and offset explicitly.**
  Based on the above new ops, createConstant, createMaxPool, createAvgPool, createFullyConnected need to be improved to support passing scale and offset.
**3) Nodes and Instrs verification function need to be updated based on quantized / non-quantized type.**

**### Now one thing needs to be discussed here:**
So far, we quantize weights and bias into int8. Both the interpreter and CPU backend support int8 bias. However, this quantized resnet50 model quantizes weights into int8, but bias into int32. According to Haixin,  it is because the partial sum of the matrix-matrix multiplication is accumulated into int32, so int32 bias can be added to the int32 partial sum for better accuracy (i.e. int8 bias caused accuracy drop). 
So, for our backend, do we need to support both int8 quantized bias and int32 quantize bias or only int32 quantized bias ? Or any other idea? Thanks!
@rdzhabarov @artemrakhov @opti-mix @nadavrot

rdzhabarov:
Thanks @beicy for going through the required op support.

[...]

PR #1853, opened 2018-10-11
[Importer] Support arg "order" in Caffe2 Conv, MaxPool and AveragePool ops

beicy:
*Description*:

Add arg "order" in Caffe2 Conv, MaxPool and AveragePool ops.
The "order" can be "NCHW" (by default) and "NHWC".
 
*Testing*: 
Unittests added.

*Documentation*:
N/A

Needed for https://github.com/pytorch/glow/issues/1727

### DIFF
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -61,6 +61,207 @@ TEST(caffe2, importConv) {
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
 }
 
+/// Test loading conv op from a Caffe2 model.
+/// The input is N*H*W*C (1*3*3*1), the kernel is 2,
+/// stride is 1, pad is 1, group is 1.
+TEST(caffe2, convNHWC) {
+  ExecutionEngine EE{BackendKind::Interpreter};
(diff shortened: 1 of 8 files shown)


tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -61,6 +61,131 @@ TEST(caffe2, importConv) {
...
+
+  // Destroy the loader after the graph is loaded since the following execution
+  // will not depend on anyting from the loader.
+  {
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"inputs"},
+                               {&inputs.getType()}, *F);
+    output = caffe2LD.getSingleOutput();
+
+    ctx.allocate(mod.getPlaceholders());
+    updateInputsByName(ctx, &mod, {"inputs"}, {&inputs});
```
nadavrot: Why do we need this part if we are not executing anything?

beicy: Good point! Will clean up this part.

beicy:
@rdzhabarov  The quantized resnet50.

qcolombet:
Could you add tests for the NCHW order, so that we can check the transposes are properly added?

Issue #1641, opened 2018-09-13
Optimization pipeline design with more backend control


jfix71:
Right now we have target-independent optimizations that may or may not benefit the backend executing the graph, and may even degrade performance. We need to allow the backend we are targeting have more control over what optimizations are performed.

For example, the GraphOptimizer's small matrix multiplication/batched add merging optimizations (`mergeMatMul()`/`mergeBatchedAdd()`) may not be beneficial (https://github.com/pytorch/glow/issues/1494).

Another example is the GraphOptimizer's `optimizePool()`, where we swap `RELU -> MaxPool` to `MaxPool -> RELU`. If the original subgraph is `Conv -> RELU -> MaxPool`, backends may prefer to vertically fuse `Conv -> RELU` into a single `ConvRELU` instead of  swapping the order of the `RELU` and `MaxPool`. So this optimization might be counterproductive/may prevent fusion from occurring.

Also related to https://github.com/pytorch/glow/pull/1633.

QiJune:
Maybe we need to construct a cost model for each backend to analyze the cost of potential optimization.

If you give some hardware information and some `Instruction`, the cost model can return a cost number. The interface can be like this:

```
size_t getCost(const Backend& b, std::vector<Instruction>& is)
```
For each step optimization, GraphOptimizer will calculate a cost to decide whether add it or not. With iterative query on the cost model, maybe the globally optimial solution.

I am not sure if we can construct a such accurate cost model for each backend. And what hardware information should be get to construct such cost model?

aturetsk:
Another thing we need to take into account is quantization.
At the moment we do `glow::optimize` before gathering profile and quantizing the model. Since it's not backend-specific, we end up with the same graph for every backend, so it's possible to gather quantization profile with one backend and run quantized model with another.
I think it'd be good to keep this property, as it makes sense that some backends may support only quantized models. In fact, last time I checked (a while ago) profile generation was not possible with CPU backend, only with Interpreter.

PR #1857, opened 2018-10-12
[GraphOptimizer] Fix the Conv-BatchNorm optimization

jfix71:
*Description*: On Resnet50, many BatchNormalizationNodes were no longer being merged with the Convolution node before it. This commit fixes the problem.

It occurred because [this commit](https://github.com/pytorch/glow/commit/bd18db777cb775672ab47d626c019e979ac9edc8) removed trainable from Variables. This means that constant deduplication, which did not apply to trainable variables, now does apply and is merging 7 biases which are the same. Instead of giving up when the filter/bias have multiple uses, instead duplicate the filter/bias and allow the optimization to proceed.

*Testing*: Added a unit test

*Documentation*: N/A

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -885,32 +885,55 @@ static void optimizePool(Function *F) {
   } // For all nodes in the graph.
 }
 
-/// \returns The uniquely used Constant from node or nullptr
-/// if node has more than one user or is not a Constant.
-static Constant *getUniquelyUsedConstant(Node &node) {
-  // If that node has more than one use, it may not
-  // be okay to modify the underlying Constant.
-  if (!node.hasOneUse()) {
+/// \returns a uniquely used Constant with the same contents as \p node. If \p
+/// node is not a Constant then \returns a nullptr. If \node is a Constant which
+/// has a single use, \p node is returned. If \node is a Constant which has
+/// multiple uses, then \returns a new duplicate Constant that has the same
+/// contents as \p node contained in \p M.
+static Constant *getUniquelyUsedConstant(Module *M, Node &node) {
+  Constant *constant = dyn_cast<Constant>(&node);
+  if (!constant) {
     return nullptr;
   }
-  return dyn_cast<Constant>(&node);
+
+  if (constant->hasOneUse()) {
+    return constant;
+  }
+
+  // If constant has more than one use, duplicate it and return the duplicate.
+  auto *NC = M->createConstant(constant->getType(), constant->getName());
+  NC->getPayload().assign(&constant->getPayload());
+  return NC;
 }
 
-/// Normalize the weight of \p CV with what \p BN is doing.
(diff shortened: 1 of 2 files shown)


qcolombet:
> However the optimization doesn't change the Conv's filter/bias, it just reads from them. So it's fine for the optimization to proceed.

Hmm, I am confused. Don't we modified them at line 974 and 988 respectively?

jfix71:
Whoops! Yeah, I totally misread it 🙁 . We need to duplicate the constant. I'll update the PR and messages.

jfix71:
Added a new version that duplicates Constants if necessary. This uncovered a problem in CSE for nodes with predicates, which I added a fix for in https://github.com/pytorch/glow/pull/1861. I will land this once that lands.

nadavrot:
Nice! Good catch!

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -885,34 +885,52 @@ static void optimizePool(Function *F) {
...
+  Constant *cbiasC = getUniquelyUsedConstant(F, *CV.getBias().getNode());
 
   if (!filterC || !cbiasC) {
-    return false;
+    return nullptr;
   }
 
+  // Create a new convolution node with the same parameters and input as CV, but
+  // with unique filter/bias which will be normalized below.
+  auto *newCV = F->createConv(CV.getName(), CV.getInput(), filterC, cbiasC,
```
qcolombet: Maybe add a comment that we need to create a new node because the constants may not be bound to that node.

Alternatively, we could always call setNThInput on the related arguments. BTW it may be better because it would avoid to rely on DCE to get rid of the now unconnected convolution node.

qcolombet:
BTW we might want to update the description of the PR in case we want to refer back to it.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -885,34 +885,58 @@ static void optimizePool(Function *F) {
...
-static Constant *getUniquelyUsedConstant(Node &node) {
-  // If that node has more than one use, it may not
-  // be okay to modify the underlying Constant.
-  if (!node.hasOneUse()) {
+/// \returns a uniquely used Constant with the same contents as \p node. If \p
+/// node is not a Constant then \returns a nullptr. If \node is a Constant which
+/// has a single use, \p node is returned. If \node is a Constant which has
+/// multiple uses, then \returns a new duplicate Constant that has the same
+/// contents as \p node.
+static Constant *getUniquelyUsedConstant(Function *F, Node &node) {
```
qcolombet: Nit: given we just use F->getParent(), maybe we should pass the module instead.

jfix71: Done

jfix71:
BTW, this optimization also applies to https://github.com/pytorch/glow/issues/1641, as it is possible (albeit unlikely) a backend would prefer to not duplicate the weights/bias Constants and instead just perform the batch norm.

PR #1858, opened 2018-10-12
[Converter] Optimize intermediate conversions away

qcolombet:
Note: After this PR lands, the only conversions left in ResNet50 converted to FP16 are the input and output of the networks. I'm working on adding a nice setup for this so that it will be easy to run a fully converted model with pre- and post- function for copying the input and output at the right place while running a model without conversions in the way.

*Description*
After converting a model we often end up with patterns like this:
```
conversion (conversion A to B) to C
```
Where A and C are actually the same type.

This PR adds a method to optimize away this kind of conversions as well as the conversions applied to constants. 

This optimization is not performed during the main conversion procedure (FunctionConverter::convert) because it potentially changes the semantic of the program when B would have incured some precision loss (e.g., float(int64_t(float (A))) != float(A)).

This works seamlessly for the FunctionQuantizer and the TypeAToTypeBFunctionConverter.

*Testing*
Added test case.

### DIFF
--- a/lib/Converter/FunctionConverter.cpp
+++ b/lib/Converter/FunctionConverter.cpp
@@ -15,11 +15,15 @@
  */
 #include "glow/Converter/FunctionConverter.h"
 
+#include "glow/Base/Tensor.h" // For Tensor.
 #include "glow/Graph/Graph.h" // For Function.
 #include "glow/Graph/Node.h"  // For Node.
 #include "glow/Graph/Nodes.h" // For Placeholder and Variable.
 
 using namespace glow;
+
+FunctionConverter::FunctionConverter(Function &F)
+    : mod_(*F.getParent()), function_(F) {}
 
 TypeRef
 FunctionConverter::getTargetTypeForOutput(const NodeValue &nodeVal) const {
@@ -159,6 +163,40 @@ void FunctionConverter::convert() {
   function_.verify();
 }
 
+bool FunctionConverter::canConvert(const Tensor &tensor, TypeRef dstTy) const {
+  // The default implementation of convertTensor uses Tensor::convertToType
+  // which does not support quantization for now.
+  return !dstTy->isQuantizedType() && !tensor.getType().isQuantizedType();
+}
+
+void FunctionConverter::convertTensor(Tensor &tensor, TypeRef dstTy) {
+  assert(canConvert(tensor, dstTy) && "We shouldn't call this method");
+  assert(tensor.dims() == dstTy->dims() &&
+         "Conversion should only change the values, not the shape");
+  tensor.convertToType(dstTy->getElementType());
+}
+
+NodeValue FunctionConverter::convertConstant(Constant &constant,
+                                             TypeRef dstTy) {
+  const Tensor &tensor = constant.getPayload();
+  if (!canConvert(tensor, dstTy)) {
+    return NodeValue();
+  }
+
+  Constant *constantToBeModified = &constant;
+  if (!constant.hasOneUse()) {
+    // Could be a simple clone, but storage classes don't support cloning for
+    // now.
+    constantToBeModified =
+        mod_.createConstant(constant.getType(), constant.getName());
+    constantToBeModified->getPayload().assign(&constant.getPayload());
+  }
+  Tensor &tensorToBeModified = constantToBeModified->getPayload();
+  constantToBeModified->setType(0, dstTy);
+  convertTensor(tensorToBeModified, dstTy);
+  return NodeValue(constantToBeModified, 0);
+}
+
 bool FunctionConverter::optimizeConversions() {
   // Traverse all the conversions introduced during the conversion step and
   // remove them.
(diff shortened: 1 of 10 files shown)


nadavrot:
@qcolombet Are you blocked on code review for any of the PRs?

nadavrot:
@qcolombet this implementation significantly complicates the converter and it would require that we duplicate some of the conversion logic. Have you considered implementing the following logic:

1. In the optimizer replace the sequence A->B->C into A->C, only if A, B, C are float types.
2. n the optimizer replace the sequence A->B->C into round(A)->C, if any of A, B, C are integer types.

Would it not be simpler?

lib/Converter/FunctionConverter.cpp
```diff
@@ -159,6 +163,39 @@ void FunctionConverter::convert() {
   function_.verify();
 }
 
+bool FunctionConverter::canConvert(const Tensor &tensor, TypeRef dstTy) const {
+  // The default implementation of convertTensor uses Tensor::converToType
```
jfix71: nit: typo "converToType"

lib/Converter/FunctionConverter.cpp
```diff
@@ -159,6 +163,39 @@ void FunctionConverter::convert() {
...
+  }
+
+  Constant *constantToBeModified = &constant;
+  if (!constant.hasOneUse()) {
+    // Could be a simple clone, but storage classes don't support cloning for
+    // now.
+    constantToBeModified =
+        mod_.createConstant(constant.getType(), constant.getName());
+  }
+  Tensor &tensorToBeModified = constantToBeModified->getPayload();
```
jfix71: Don't you need to assign the the new Constant's payload to the old one, or am I missing something? Also doesn't look like you're checking any values for this in the unit test added, so perhaps that should be added?

tests/unittests/typeAToTypeBFunctionConverterTest.cpp
```diff
@@ -439,6 +439,134 @@ TEST_P(AllBackends, int64IConversionFloatToFloat16) {
...
+  auto *convertedFCInput =
+      llvm::dyn_cast<ConvertToNode>(convertedFC->getInput());
+  ASSERT_NE(convertedFCInput, nullptr);
+  EXPECT_EQ(convertedFCInput->getElementType(0), ElemKind::Float16Ty);
+  EXPECT_TRUE(llvm::isa<Placeholder>(convertedFCInput->getInput()));
+  EXPECT_EQ(convertedFCInput->getInput().getElementType(), ElemKind::FloatTy);
+  EXPECT_EQ(convertedFCInput->getInput().getNode(), input);
+
+  // Check that the weights have been updated to float16.
+  auto *convertedFCWeights =
```
jfix71: As noted before, perhaps you can set the value of weights earlier in this test and then check that the value was actually converted?

qcolombet:
> Optimizing conversions inside the quantizer sounds like a shortcut to work around the need to figure out the rules. Let's figure out the legality rules before moving forward.

That’s not the idea. The idea is at the end of the quantization/conversion process, I wanted a clean model that can be exported. That is suboptimal to having to re-optimize the converted model each time we load it...
Now, it seems your concern is where the optimization code live. I am happy to move it in the optimizer and call the optimizer from outside the conversion process. It feels to me that every users of this feature would have to do that and moreover the patterns we are going to optimize don’t really happen in real networks. But if you prefer this way let’s go with it!

lib/Converter/FunctionConverter.cpp
qcolombet: My phone doesn’t show me the context but IIRC the following line convertTensor modifies the tensor given by reference.
Maybe I should make it more obvious.
Idea?

jfix71: `convertTensor()` modifies the tensor by reference, but the tensor's original contents (i.e. the contents from `constant`) don't appear to ever be assigned to `tensorToBeModified`. It looks to me like there are no references to the original `constant` after `constantToBeModified` is created, i.e. its backing payload tensor is probably left zero-initialized.

qcolombet: Got it!

qcolombet: I used clone previously before realizing it didn't work and stupidly just replaced it with createConstant which is not enough obviously.

lib/Converter/FunctionConverter.cpp
qcolombet: Done

lib/Converter/FunctionConverter.cpp
qcolombet: Done

tests/unittests/typeAToTypeBFunctionConverterTest.cpp
qcolombet: Done

nadavrot:
I think that it would be best to add a few peepholes in the Graph optimizer. This will place the conversion optimizations next to the quantization (quantize-unquantize) optimizations. This would also allow us to simplify the unit tests that won't have a dependency on the quantizer.

qcolombet:
Abandoned in favor of #1865.
In #1865, the conversion optimization is added to the graph optimizer and the converter is left untouched.

Issue #985, opened 2018-05-04
[unittest coverage] Improve test coverage for various aspects of glow


rdzhabarov:
Every master commit invokes coverage build and the results of coverage build can be found: https://fb-glow-assets.s3.amazonaws.com/coverage/coverage-master/index.html

It would be great to improve test coverage for certain low coverage places.

That's a good first issue to get some familiarity with the glow code base, contributions are appreciated.

rdzhabarov:
thanks @AnthonyOfSeattle for the interest.

It depends on a certain case, sometimes it makes sense to tweak an existing test to hit all the coverage points. Sometimes it's better to express that as a separate test, really on a case by case basis.

Here are few starters options:
* Missing coverage for getting quantization params based on the [min,max] range. See chooseQuantizationParams for details. There are some important cases.
* https://fb-glow-assets.s3.amazonaws.com/coverage/coverage-master/lib/Quantization/Quantization.cpp.gcov.html.
Majority of missing coverage in quantizeNode is coming from the quantization of some RNN/LSTM related operators. You could hit all those with the quantized fr2en test (--dump/load_profile options). Ideally, it would be nice to have e2e test similar to CNN e2e test (https://github.com/pytorch/glow/blob/master/tests/unittests/quantizationTest.cpp#L189).
* Coverage for Slice(Rescale(X)) -> Rescale(Slice(X)), see here https://fb-glow-assets.s3.amazonaws.com/coverage/coverage-master/lib/Optimizer/GraphOptimizer.cpp.gcov.html
* Make sure that performDebugInstrumentation does not fail, https://fb-glow-assets.s3.amazonaws.com/coverage/coverage-master/lib/Optimizer/IROptimizer.cpp.gcov.html


Let me know if i can help you anyhow.

AnthonyOfSeattle:
Thanks for the suggestions. It seems that it may not be too difficult to get an example like fr2en but with random initialization like the e2e CNN. Here is a initial idea for a simple graph:

``` 
/// Fills the tensor \p H with some stable random integers with the seed \p seed
/// and the range [0 .. scale].
static void fillStableRandomIndex(Handle<size_t> H, size_t seed,
                                  int scale = 10) {
  for (size_t i = 0, e = H.size(); i < e; i++) {
    H.raw(i) = int(i * 1921 + seed) % scale;
  }
}

static std::pair<Function *, SaveNode *>
createSimpleGRUForQuantization(Module *M) {
[...]

PR #1859, opened 2018-10-12
[caffe2ImporterTest] Add eq unittest to caffe2ImporterTest

gcatron:
*Description*: Added EQ node test to caffe2ImporterTest. This checks that the node is imported and the graph is the expected shape.
*Testing*: Run caffe2ImporterTest caffe2.EQ1D passes.
#985

### DIFF
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -1013,3 +1013,42 @@ TEST(caffe2, batchBoxCox) {
     EXPECT_FLOAT_EQ(result.raw(i), OH.raw(i));
   }
 }
+
+// Test loading a EQ operator with 1D inputs.
+TEST(caffe2, EQ1D) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename("tests/models/caffe2Models/eq_op_net.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  Placeholder *output;
+  Context ctx;
+
+  // Input tensors.
+  const size_t kDataSize = 10;
+  Tensor X(ElemKind::FloatTy, {kDataSize});
+  Tensor Y(ElemKind::FloatTy, {kDataSize});
+
+  // Destroy the loader after the graph is loaded
+  {
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
+                               {&X.getType(), &Y.getType()}, *F);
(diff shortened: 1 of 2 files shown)


tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1013,3 +1013,58 @@ TEST(caffe2, batchBoxCox) {
...
   }
 }
+
+// Test loading a EQ operator with 1D inputs.
+TEST(caffe2, EQ1D) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename("tests/models/caffe2Models/eq_op_net.pbtxt");
```
rdzhabarov: you would need to `git add` this file.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1013,3 +1013,58 @@ TEST(caffe2, batchBoxCox) {
...
+  Tensor Y(ElemKind::FloatTy, {kDataSize});
+  auto XH = X.getHandle();
+  auto YH = Y.getHandle();
+
+  // Fill inputs with same random values.
+  XH.randomize(-3.0, 3.0, mod.getPRNG());
+  for (size_t i = 0; i < kDataSize; ++i) {
+    YH.at({i}) = XH.at({i});
+  }
+
```
beicy: Since no real run is needed, don't need to assign value to the inputs.

rdzhabarov: @beicy can you outline the guideline on what needs and what does not need to be tested in this/and onnxLoader tests?
So it would be clear on what really should be in this test.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1013,3 +1013,44 @@ TEST(caffe2, batchBoxCox) {
...
+  Tensor Y(ElemKind::FloatTy, {kDataSize});
+
+  // Destroy the loader after the graph is loaded
+  {
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
+                               {&X.getType(), &Y.getType()}, *F);
+    output = caffe2LD.getSingleOutput();
+    ctx.allocate(mod.getPlaceholders());
+    updateInputsByName(ctx, &mod, {"X", "Y"}, {&X, &Y});
+  }
```
beicy: Also, line 1040 and line 1041 are not needed.

beicy:
Looks good:) Please squash the commits before merge. Btw, you don't have to make a new commit for every change, just amend it to your previous one.

rdzhabarov:
I think it's a good practice to not squash commits while addressing feedback, that makes reviewing addressed comments easier.

But before the merge, we should squash what needs to be a single commit.

PR #1862, opened 2018-10-12
[ONNXIFI] Return 2 supported backends + some cleanup.

rdzhabarov:
*Description*:
Return 2 supported backends + some cleanup.
Currenty C2 just uses backend[0] for execution.

*Testing*:
Int test

*Documentation*:
N/A

cc: @yinghai


yinghai:
Why do we want to do this?

rdzhabarov:
@yinghai regarding two backends.

That's good to exercise that higher level framework works fine in situations when there is more than one backend available through the ONNXIFI interface.
Also, make sure it does not crash onnxifi tests etc.

PR #1865, opened 2018-10-12
[GraphOptimizer] Optimize intermediate conversions away

qcolombet:
*Description*
After converting a model we often end up with patterns like this:
conversion (conversion A to B) to C
Where A and C are actually the same type.

This patch adds a method to optimize away this kind of conversions
as well as the conversions applied to constants.

*Testing*
Added test cases.


qcolombet:
Basically moving the code out of the FunctionConverter from #1858 to the graph optimizer per @nadavrot's request.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1609,6 +1609,126 @@ static void optimizeQuantizedMaxSplat(Function *F) {
...
+                                 TypeRef dstTy) {
+  // Sort out the easy case first.
+  if (constant.getType() == dstTy) {
+    return NodeValue(&constant, 0);
+  }
+  auto modifyConstantTyAndGet = [&]() -> Constant & {
+    Constant *oneUseCst = getUniquelyUsedConstant(&mod, constant);
+    assert(oneUseCst &&
+           "We should always be able to get a constant from a constant!");
+    oneUseCst->setType(0, dstTy);
```
nadavrot: I am not sure that I understand this part of the code. We are setting the type of the constant here. Does this convert the content of the buffer? Will it turn the Int8 backing tensor to Fp32?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2017,6 +2128,9 @@ void glow::optimize(Function *F, CompilationMode mode) {
   DCE(F);
   mergeTransposeIntoMatMul(F);
 
+  // Optimize away intermediate conversions.
```
nadavrot: // Optimize away intermediate type conversions.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: That's something I discovered when I wrote this code.
Basically we have the type of the constant in two places: the node and the tensor.
This part of the code changes the type of the constant and it doesn't have any effect on the tensor itself.
I should add a comment here and we should probably fix this dual type thing at some point!

qcolombet: Comment added.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: Done

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1609,6 +1609,126 @@ static void optimizeQuantizedMaxSplat(Function *F) {
...
+    switch (dstTy->getElementType()) {
+    case ElemKind::FloatTy:
+    case ElemKind::Float16Ty: {
+      // Plain conversion.
+      Constant &constantToBeModified = modifyConstantTyAndGet();
+      constantToBeModified.getPayload().convertToType(dstTy->getElementType());
+      return NodeValue(&constantToBeModified, 0);
+    }
+    case ElemKind::Int8QTy: {
+      // Quantization.
```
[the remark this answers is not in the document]
qcolombet: Done

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1609,6 +1609,135 @@ static void optimizeQuantizedMaxSplat(Function *F) {
...
+    // Dequantization: Int[8|16|32]QTy -> {FloatTy, Float16Ty, Int64I}.
+    // Rescale: Int[8|16|32]QTy -> Int[8|16|32]QTy.
+    // Plain conversion: Int64ITy -> {FloatTy, Float16Ty}.
+    // Quantization: Int64ITy -> Int[8|16|32]QTy.
+    return NodeValue();
+  }
+}
+
+/// Optimize away ConvertToNode.
+/// This basically turns conversion(conversion A to B) to C"
```
MattPD: Small typo: missing opening `"` in `conversion(conversion A to B) to C"`

lib/Optimizer/GraphOptimizer.cpp
nadavrot: And add something to the verifier.

qcolombet: Good point! Done in PR #1869.

lib/Optimizer/GraphOptimizer.cpp
qcolombet: Fixed the typo in PR #1869.

PR #1866, opened 2018-10-12
[Loader] Add an option to run a model in fp16

qcolombet:
*Description*
This patch adds an option `-convert-to-fp16` to convert all fp32 operators
into fp16 ones.

We may want to expose more conversion options in the future and change
the `convert-to` option to something taking an enum like fp32-to-fp16.

As part of the conversion mechanism it is possible to choose which nodes
are okay to convert (`-do-not-convert-nodes=<listOfNodes>`).

The conversion process does not alter the inputs and outputs of the
network. Thus, the converted graph will have ConvertToNode at least
at its start and end.

*Testing*
Able to convert and run resnet50 in fp16 with the interpreter.

*Documentation*
None so far, we want the interpreter to fully support fp16 before people
play with it.

Related to #1329

### DIFF
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -17,6 +17,7 @@
 #include "Loader.h"
 
 #include "glow/Base/Tensor.h"
+#include "glow/Converter/TypeAToTypeBFunctionConverter.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/IR/IR.h"
 #include "glow/Quantization/Serialization.h"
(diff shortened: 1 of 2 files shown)


tools/loader/Loader.cpp
```diff
@@ -123,6 +124,19 @@ llvm::cl::opt<bool> dumpGraphOpt("dumpGraph",
...
                                  llvm::cl::cat(modelExportCat));
 
+llvm::cl::list<std::string> doNotConvertNodesOpt(
+    "do-not-convert-nodes",
+    llvm::cl::desc(
+        "Use to specify the name of nodes (e.g. Add, Div, etc.) that should "
+        "not be converted."),
+    llvm::cl::value_desc("NodeNames (e.g. Add,Div)"), llvm::cl::ZeroOrMore,
+    llvm::cl::CommaSeparated, llvm::cl::cat(loaderCat));
+
```
beicy: The option "--do-not-convert-nodes" : Does it mean in most of cases, only a few of types of nodes should not be converted?  Just wondering why we don't use "do-convert-nodes" :)

nadavrot:
I am excited to see this feature landing. Go Go FP16 resnet50!

tools/loader/Loader.cpp
```diff
@@ -123,6 +124,19 @@ llvm::cl::opt<bool> dumpGraphOpt("dumpGraph",
                                  llvm::cl::desc("Prints Graph to stdout"),
                                  llvm::cl::cat(modelExportCat));
 
+llvm::cl::list<std::string> doNotConvertNodesOpt(
```
rdzhabarov: we already have `do-not-quantize-node` option, wondering if we could merge those to a single option.
As in principal those are the nodes we do not touch during any kind of processing.

tools/loader/Loader.cpp
qcolombet: I don't have a good answer to this. I reproduced what we do for quantization and I assumed it was the same kind of use case.
What do you think?

tools/loader/Loader.cpp
qcolombet: I was wondering about this too. The only reason I didn't do it, is because I couldn't come up with a good name :).
Any idea?

tools/loader/Loader.cpp
jfix71: Yeah, the reason I went with `do-not` instead of `do` for quantization is that I was expecting we would quantize as much as possible, and so it's likely the `do-not` list is much shorter than the `do` list. E.g. for en2gr I was quantizing everything but Add, so it made more sense to do like this.

tools/loader/Loader.cpp
jfix71: Maybe something like `keep-original-precision`?

qcolombet: Went with `keep-original-precision-for-nodes`.

PR #1867, opened 2018-10-13
[NodeGen] Update the docstring for ConvertToNode. NFC

qcolombet:
*Description*
Clarify that the ConvertToNode is not supposed to be used for quantized
types. Instead call out in the docstring that the related quantization
nodes should be used.

Sorry for the delay @jfix71!


nadavrot:
Do we verify this in the verifier?

qcolombet:
> Do we verify this in the verifier?

Thanks for pointing this out. Yes, it already does.

PR #1869, opened 2018-10-15
[Verifier] Check that the type of a constant and its tensor matches

qcolombet:
*Description*
Right now, the type of a constant and the type of its underlying tensor
are stored in two different places. As a result, it is possible to
modify one without modifying the other ending up in a situation
where the constant and its tensor have different types.

This patch makes sure that the verifier reports this problem.

*Testing*
Added test case

### DIFF
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -1155,4 +1155,33 @@ TEST(Graph, verifyConstantNoWriters) {
   EXPECT_DEATH(M.verify(), "");
 }
 
+/// Check that the verifier will complain if a constant and its
+/// underlying tensor have mismatching types.
+/// Here the constant is updated but not the tensor.
+TEST(Graph, verifyConstantTensorTypeMatchesConstantTypeChanged) {
+  Module M;
+
+  auto *input = M.createConstant(ElemKind::FloatTy, {5}, "input");
+  // Fresh constant should verify just fine.
+  input->verify();
+
+  input->setType(0, M.uniqueType(ElemKind::Float16Ty, {5}));
+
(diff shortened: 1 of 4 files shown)


jfix71:
This PR has made me realize that we never currently call verify on any Storage nodes during `compile()`.

`Module::verify()` doesn't call verify on the Storage nodes it owns (and it seems `Module::verify()` is only ever called directly in a couple recently added unit tests anyway).

`Function::verify()` only verifies the Nodes it owns, and not the Storage nodes accessed by its Nodes. 

And in fact Storage node verification is broken -- Constant (prior to this PR) and Placeholder do not implement their own `verify()`, so when `Node::verify()` is called on one, it gets stuck in infinite recursion calling itself and runs out of stack space/seg faults.

tests/unittests/graphTest.cpp
```diff
@@ -1155,4 +1155,33 @@ TEST(Graph, verifyConstantNoWriters) {
   EXPECT_DEATH(M.verify(), "");
 }
 
+/// Check that the verifier will complain if a constant and its
```
nadavrot: Do we get a good signal-to-noise ratio from this unit test?

Will the testing benefits of this unit test outweigh the cost of having to maintain it?

qcolombet: I have mixed feeling about this one.
Most of our verifier checks are untested but recently @jfix71 showed us a path how to check them and I liked it.
I don't expect this will need updating in the (near) future, but I don't have a strong opinion either way.
What do you think?

nadavrot: I don't have a strong opinion about this specific PR. I just wanted to start this discussion that can guide us in future work.

PR #1870, opened 2018-10-15
[ONNXifi] Make onnxRunGraph execute Glow graph asynchronously

### LINKED ISSUE #1570: [ONNXIFI] Execute onnxRunGraph async
ONNXIFI interface specifies that onnxRunGraph should be executed async, see [here](https://github.com/onnx/onnx/blob/master/onnx/onnxifi.h#L1704) for details.

```cpp
ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
onnxRunGraph(onnxGraph graph, const onnxMemoryFenceV1 *inputFence,
             onnxMemoryFenceV1 *outputFence)
```

Specifically, we should do the following sequence of actions:
* Do the checks against type of the syncronization event and tags (it's already handled on the current master)
* Init **output** event (this event is used by the caller to track whether current inference has completed or still in progress).
* Schedule run of this particular inference on a separate thread or thread pool (could be improved down the road)
* Return ONNXIFI_STATUS_SUCCESS from the onnxRunGraph
* On the async path of inference execution wait for the **input** event to fire (this signals that inputs are copied to the proper addresses, which are set on the prior SetGraphIO call)
* Run inference normally and copy outputs to the required addresses (addresses are specified in the prior SetGraphIO call)

cc: @Maratyszcza @yinghai

SplitInfinity:
**Description**
This commit modifies `onnxRunGraph` to make graph execution asynchronous
by submitting it as a lambda to an instance of `ThreadPool` (a new class
encapsulating a thread pool) stored in the `Backend` class. `ThreadPool`
consists of an adjustable number of threads that continuously pop and
execute an item from a common queue of `std::function<void(void)>` objects.

This commit fixes #1570.

**Testing**
This commit adds a unit test for the new `ThreadPool` class. It has also been tested on two known good ONNXifi tests, `test_concat_2d_axis_1`
and `test_concat_2d_axis_1`:

```
$ .setuptools-cmake-build/onnxifi_test_driver_gtests /tmp/onnx_tests
[==========] Running 2 tests from 1 test case.
[----------] 2 tests from ONNXCppAllTest/ONNXCppDriverTest
[ RUN      ] ONNXCppAllTest/ONNXCppDriverTest.ONNXCppDriverUnitTest/test_concat_2d_axis_1
[       OK ] ONNXCppAllTest/ONNXCppDriverTest.ONNXCppDriverUnitTest/test_concat_2d_axis_1 (89 ms)
[ RUN      ] ONNXCppAllTest/ONNXCppDriverTest.ONNXCppDriverUnitTest/test_concat_3d_axis_1
[       OK ] ONNXCppAllTest/ONNXCppDriverTest.ONNXCppDriverUnitTest/test_concat_3d_axis_1 (59 ms)
[----------] 2 tests from ONNXCppAllTest/ONNXCppDriverTest (148 ms total)

[==========] 2 tests from 1 test case ran. (148 ms total)
[  PASSED  ] 2 tests.
```
 as well as the resnet-50 C2/Glow integration test:
```
$ pytest -s caffe2/python/onnx/test_onnxifi.py 
========================================================================================================= test session starts =========================================================================================================
platform darwin -- Python 3.6.2+, pytest-3.8.2, py-1.7.0, pluggy-0.7.1
plugins: hypothesis-3.76.0

...                                                                                                                                                                                                                 

caffe2/python/onnx/test_onnxifi.py ssBatch size: 1, repeat inference 1 times

...
C2 runtime: 0.414517879486084s
Conversion time: 15.36s
Onnxifi runtime: 0.140455961227417s, improvement: 66.11582559441025%
.

================================================================================================ 1 passed, 2 skipped in 43.76 seconds =================================================================================================
```

Resnet-50 ONNX statistics on master, before this change:
```
C2 runtime: 0.411358118057251s
Conversion time: 15.47s
Onnxifi runtime: 0.1417860984802246s, improvement: 65.53219876883736%
```

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -26,6 +26,10 @@ bool BackendId::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) {
   return executionEngine_.isOpSupported(opKind, elementTy);
 }
 
+void Backend::runAsync(const std::function<void(void)> &fn) {
+  threadPool_.submit(fn);
+}
+
 bool Event::signal() {
   {
     std::lock_guard<std::mutex> guard(mutex_);
--- a/lib/Onnxifi/Base.h
+++ b/lib/Onnxifi/Base.h
@@ -16,6 +16,8 @@
 #ifndef GLOW_ONNXIFI_BASE_H
 #define GLOW_ONNXIFI_BASE_H
 
+#include "ThreadPool.h"
+
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/Importer/ONNXIFIModelLoader.h"
 
@@ -62,8 +64,13 @@ public:
   /// \returns Execution Engine associated with the Backend.
   glow::ExecutionEngine &getEE() { return backendIdPtr_->getEE(); }
 
+  /// Run async using backend thread pool.
+  void runAsync(const std::function<void(void)> &fn);
+
 private:
   BackendIdPtr backendIdPtr_;
+  // ThreadPool instance for the backend.
+  ThreadPool threadPool_;
 };
 
 typedef Backend *BackendPtr;
--- a/lib/Onnxifi/CMakeLists.txt
+++ b/lib/Onnxifi/CMakeLists.txt
@@ -5,7 +5,12 @@ add_library(onnxifi-glow
               Base.cpp
               onnxifiGlow.cpp)
 
+add_library(onnxifi-glow-thread-pool
+              ThreadPool.cpp)
+
 target_link_libraries(onnxifi-glow
                       PUBLIC
                         ExecutionEngine
-                        Importer)
+                        Importer
+                      PRIVATE
+                        onnxifi-glow-thread-pool)
--- a/lib/Onnxifi/ThreadPool.cpp
+++ b/lib/Onnxifi/ThreadPool.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "ThreadPool.h"
+
+#include <iostream>
+
+namespace glow {
+namespace onnxifi {
+
+ThreadPool::ThreadPool(unsigned numWorkers) : shouldStop_(false) {
+  // Intialize all workers and make each one run threadPoolWorkerMain.
+  for (unsigned i = 0; i < numWorkers; i++) {
+    std::thread th(std::bind(&ThreadPool::threadPoolWorkerMain, this));
+    workers_.push_back(std::move(th));
+  }
+}
+
+ThreadPool::~ThreadPool() {
+  // Lock mutex before signalling for threads to stop to make sure
+  // a thread can't wait on the condition variable after checking the
+  // *old* value of shouldStop_.
+  std::unique_lock<std::mutex> lock(workQueueMtx_);
+
+  // Signal to workers to stop.
+  shouldStop_ = true;
+
+  // Notify all worker threads in case any are waiting on the condition
+  // variable.
+  lock.unlock();
+  queueNotEmpty_.notify_all();
+
+  // Join all worker threads.
+  for (auto &w : workers_) {
+    w.join();
+  }
+  workers_.clear();
+}
+
+void ThreadPool::submit(const std::function<void(void)> &fn) {
+  // Add fn to the work queue.
+  std::unique_lock<std::mutex> lock(workQueueMtx_);
+  workQueue_.push(fn);
+  lock.unlock();
+  queueNotEmpty_.notify_one();
+}
+
+void ThreadPool::threadPoolWorkerMain() {
+  std::unique_lock<std::mutex> lock(workQueueMtx_, std::defer_lock);
+
+  while (!shouldStop_) {
+    // Lock the lock after processing a work item.
+    lock.lock();
+
+    // If work queue is empty, wait to be signalled when
+    // a work item is submitted.
+    while (workQueue_.empty() && !shouldStop_) {
+      queueNotEmpty_.wait(lock);
+    }
+
+    // If shouldStop_ was set to false while the thread
+    // was asleep, break out of the main loop.
+    if (shouldStop_) {
+      break;
+    }
+
+    // Pop a work item from the queue, and make sure to unlock
+    // the lock before processing it.
+    auto workItem = workQueue_.front();
+    workQueue_.pop();
+    lock.unlock();
+
+    // Process work item.
+    workItem();
+  }
+}
+} // namespace onnxifi
+} // namespace glow
--- a/lib/Onnxifi/ThreadPool.h
+++ b/lib/Onnxifi/ThreadPool.h
@@ -1,0 +1,70 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_ONNXIFI_THREAD_POOL_H
+#define GLOW_ONNXIFI_THREAD_POOL_H
+
+#include <atomic>
+#include <condition_variable>
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
+
+namespace glow {
+namespace onnxifi {
+
+/// Thread pool for asynchronous execution of generic functions.
+class ThreadPool final {
+public:
+  /// Constructor. Initializes a thread pool with \p numWorkers
+  /// threads and has them all run ThreadPool::threadPoolWorkerMain.
+  ThreadPool(unsigned numWorkers = kNumWorkers);
+
+  /// Destructor. Signals to all threads to stop and waits for all of them
+  /// to exit.
+  ~ThreadPool();
+
+  /// Submit \p fn as a work item for the thread pool.
+  void submit(const std::function<void(void)> &fn);
+
+private:
+  /// Main loop run by the workers in the thread pool.
+  void threadPoolWorkerMain();
+
+  /// The default number of workers in the thread pool (overridable).
+  constexpr static unsigned kNumWorkers = 10;
+
+  /// Flag checked by the workers in between work items to determine
+  /// whether they should stop and exit.
+  std::atomic<bool> shouldStop_;
+
+  /// Queue of work items.
+  std::queue<std::function<void(void)>> workQueue_;
+
+  /// Mutex to coordinate access to the work queue.
+  std::mutex workQueueMtx_;
+
+  /// Condition variable to signal to threads when work is added to
+  /// the work queue.
+  std::condition_variable queueNotEmpty_;
+
+  /// Vector of worker thread objects.
+  std::vector<std::thread> workers_;
+};
+} // namespace onnxifi
+} // namespace glow
+
+#endif // GLOW_ONNXIFI_THREAD_POOL_H
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -402,21 +402,22 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxRunGraph)(
     return initStatus;
   }
 
-  // Wait for all inputs to be ready.
-  auto waitStatus =
-      GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxWaitEvent)(inputFence->event);
-  if (waitStatus != ONNXIFI_STATUS_SUCCESS) {
-    return waitStatus;
-  }
+  // Submit graph for asynchronous execution.
+  glowGraph->backend()->runAsync([inputFence, outputFence, glowGraph]() {
+    // Wait for all inputs to be ready.
+    auto waitStatus =
+        GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxWaitEvent)(inputFence->event);
 
-  // TODO: Implement async graph run procedure. For now run that sync.
-  glowGraph->run();
+    // If all inputs are ready, run the graph.
+    if (waitStatus == ONNXIFI_STATUS_SUCCESS) {
+      // Run graph.
+      glowGraph->run();
+    }
 
-  auto signalStatus = GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxSignalEvent)(
-      outputFence->event);
-  if (signalStatus != ONNXIFI_STATUS_SUCCESS) {
-    return signalStatus;
-  }
+    // Signal that the output is ready.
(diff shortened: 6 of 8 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -402,21 +402,17 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxRunGraph)(
...
-  // TODO: Implement async graph run procedure. For now run that sync.
-  glowGraph->run();
-
-  auto signalStatus = GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxSignalEvent)(
-      outputFence->event);
-  if (signalStatus != ONNXIFI_STATUS_SUCCESS) {
-    return signalStatus;
-  }
+  // Submit graph for asynchronous execution.
+  glowGraph->runAsync([inputFence, outputFence, glowGraph]() {
```
SplitInfinity: This also looks weird. Maybe I should replace this with `Graph::runAsync(...)`?

lib/Onnxifi/Base.cpp
```diff
@@ -43,6 +43,51 @@ void Event::wait() {
...
+
+void ThreadPool::threadPoolWorkerMain() {
+  while (!shouldStop_) {
+    std::lock_guard<std::mutex> lock(workQueueMtx_);
+
+    if (!workQueue_.empty()) {
+      // Pop and process a work item.
+      auto workItem = std::move(workQueue_.front());
+      workQueue_.pop();
+      workItem();
```
bertmaher: You probably want to release the mutex before running `workItem()`, right?

lib/Onnxifi/Base.cpp
```diff
@@ -43,6 +43,51 @@ void Event::wait() {
...
+    } else {
+      // Wait for a while and then check the queue again.
+      std::this_thread::sleep_for(kThreadBreakDuration);
```
bertmaher: Also probably want to release the mutex before sleeping.

lib/Onnxifi/Base.h
```diff
@@ -112,9 +154,13 @@ class Graph {
   /// Run inference.
   onnxStatus run();
 
+  /// Run async using thread pool.
+  static void runAsync(std::function<void(void)> fn);
+
 private:
   BackendPtr backendPtr_;
   Function *function_;
+  static ThreadPool threadPool_;
```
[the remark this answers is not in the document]
bertmaher: Yeah, I think this is pretty reasonable.  But to avoid SIOF, I'd be inclined to do a Meyers singleton like, 
```
ThreadPool &getThreadPool() {
  static ThreadPool threadPool(kNumWorkers);
  return threadPool;
}
```

yinghai:
General question, is this thread pool targeting CPU backend only or something more general?

lib/Onnxifi/Base.h
rdzhabarov: I think the thread pool itself needs to live on the backend level. This way we can limit the number of threads for some of the accelerators and make bigger for CPU and Interpreter backend.

lib/Onnxifi/Base.cpp
```diff
@@ -26,6 +26,16 @@ bool BackendId::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) {
   return executionEngine_.isOpSupported(opKind, elementTy);
 }
 
+void Backend::runAsync(const std::function<void(void)> &fn) {
+  auto &threadPool = getThreadPool();
+  threadPool.submit(fn);
+}
+
+ThreadPool &Backend::getThreadPool() {
+  static ThreadPool threadPool;
```
rdzhabarov: we'd need thread pool per backend, not a single pool. Say you want to handle accelerator and CPU in the same time, for CPU you'll have a thread pool with hardware::concurrency number of threads and likely 1 thread for the accelerator.

rdzhabarov:
cc: @Maratyszcza @yinghai

lib/Onnxifi/Base.h
```diff
@@ -22,15 +22,58 @@
...
+#include <thread>
+#include <vector>
 
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/StringMap.h"
 
 namespace glow {
 namespace onnxifi {
 
+/// Thread pool for asynchronous execution of generic functions.
```
rdzhabarov: might be a good time to split this out of the Base.h and have separate threadpool header/impl. For now, i think we are ok with the custom implementation. Later if we need more performant one we could add a dependency.

lib/Onnxifi/Base.cpp
```diff
@@ -43,6 +47,53 @@ void Event::wait() {
...
+    if (!workQueue_.empty()) {
+      // Pop and process a work item.
+      auto workItem = std::move(workQueue_.front());
+      workQueue_.pop();
+      lock.unlock();
+      workItem();
+    } else {
+      // Wait for a while and then check the queue again.
+      lock.unlock();
+      std::this_thread::sleep_for(kThreadBreakDuration);
```
rdzhabarov: need proper signaling mechanism when an item is added to the queue.

once you extract this class into a separate file, please, add a test

rdzhabarov: I meant to use cond_var for example to signal when the queue is not empty.

SplitInfinity: Oh, I see.

yinghai: Here is an example thread pool: 
https://github.com/pytorch/pytorch/blob/dd4b9b06a434ca3b72c8a091c8739c33325ccead/caffe2/utils/thread_pool.h#L24-L180

Note that the overhead of synchronization is pretty high if we just use `mutex` and `conditional_variable`. High performance usually requires a better concurrent design but I think this might be a good first implementation.

rdzhabarov:
> Onnxifi runtime: 0.25087404251098633s
can you also dump what was the time before this change on master.

lib/Onnxifi/Base.h
```diff
@@ -22,15 +22,58 @@
...
+  /// The amount of time workers should wait after the work queue is
+  /// exhausted before checking again.
+  constexpr static auto kThreadBreakDuration = std::chrono::microseconds(10);
+
+  /// The default number of workers in the thread pool (overridable).
+  constexpr static unsigned kNumWorkers = 10;
+
+  /// Flag checked by the workers in between work items to determine
+  /// whether they should stop and exit.
+  bool shouldStop_;
```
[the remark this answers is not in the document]
SplitInfinity: I disagree. This variable has 1 writer and N readers, so there's no need for it to be atomic.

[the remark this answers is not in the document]
yinghai: With x86 strongly consistent memory model, you can get away from it yes. But I think it's generally safer to make it `std::atomic` as I don't see you guarding it with lock when changing it. Anyway, this probably will stay as is once you add those conditional variables.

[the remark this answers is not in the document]
bertmaher: I don't think it actually ends up mattering in practice here, but it's technically undefined behavior to have data races, so we should avoid it if possible.

lib/Onnxifi/Base.cpp
```diff
@@ -43,6 +47,53 @@ void Event::wait() {
...
+
+  while (!shouldStop_) {
+    lock.lock();
+
+    if (!workQueue_.empty()) {
+      // Pop and process a work item.
+      auto workItem = std::move(workQueue_.front());
+      workQueue_.pop();
+      lock.unlock();
+      workItem();
```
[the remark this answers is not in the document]
bertmaher: So, we're currently compiling glow with `-fno-exceptions`.  I'm not sure this is a good idea, since we're likely to be embedded in systems that *do* use exceptions.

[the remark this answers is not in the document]
bertmaher: I just realized what I wrote was ambiguous.  To clarify, I meant, I'm not sure compiling with `-fno-exceptions` is a good idea long-term.  It probably is a good idea to try/catch the `workItem()` call.

lib/Onnxifi/Base.h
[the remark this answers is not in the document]
yinghai: I remember I tried to do the same thing before and got caught by TSAN. :P

lib/Onnxifi/Base.cpp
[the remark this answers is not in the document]
yinghai: Hmm, actually it this is the case, it means that the workItem should will not throw whatsoever because itself is built with `-fno-exceptions`, so we should be good without `try/catch`.

cmake/modules/GlowDefaults.cmake
```diff
@@ -16,7 +16,7 @@ if(MSVC OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
   set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELEASE} /fp:fast")
   set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_RELEASE} /fp:fast")
 else()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wnon-virtual-dtor -fno-exceptions -fno-rtti")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wnon-virtual-dtor -fno-rtti")
```
bertmaher: Can you split this `-fno-exceptions` change into a separate PR?  It probably merits a bit of discussion on its own.

lib/Onnxifi/ThreadPool.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // Intialize all workers and make each one run threadPoolWorkerMain.
+  for (unsigned i = 0; i < numWorkers; i++) {
+    std::thread th(std::bind(&ThreadPool::threadPoolWorkerMain, this));
+    workers_.push_back(std::move(th));
+  }
+}
+
+ThreadPool::~ThreadPool() {
+  // Signal to workers to stop.
+  shouldStop_ = true;
```
bertmaher: I might have this wrong, but I think you need to lock the mutex when setting `shouldStop_ = true;`.  Otherwise the worker could get to the `cond_.wait()` after the `notify_all` and deadlock.

I'm thinking it's significant that the [example from cppreference](https://en.cppreference.com/w/cpp/thread/condition_variable) sets the predicate while holding the lock.

lib/Onnxifi/ThreadPool.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // Pop a work item from the queue, and make sure to unlock
+    // the lock before processing it.
+    auto workItem = workQueue_.front();
+    workQueue_.pop();
+    lock.unlock();
+
+    // Process work item.
+    try {
+      workItem();
+    } catch (const std::bad_function_call &e) {
```
bertmaher: Let's leave off the try/catch for now and discuss `-fno-exceptions` in a separate PR.  Sorry for the churn on this.

lib/Onnxifi/CMakeLists.txt
```diff
@@ -3,7 +3,11 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
 add_library(onnxifi-glow
             SHARED
               Base.cpp
-              onnxifiGlow.cpp)
+              onnxifiGlow.cpp
+              ThreadPool.cpp)
+
+add_library(onnxifi-glow-thread-pool
```
rdzhabarov: if you expose thread pool as a library then it's a bit weird that `ThreadPool.cpp` is also included in `onnxifi-glow` library.

I think it's OK to keep ThreadPool.cpp just in `onnxifi-glow`. If there will be other use cases for ThreadPool we could extract that into the library.

lib/Onnxifi/ThreadPool.cpp
```diff
@@ -0,0 +1,87 @@
...
+    w.join();
+  }
+  workers_.clear();
+}
+
+void ThreadPool::submit(const std::function<void(void)> &fn) {
+  // Add fn to the work queue.
+  std::lock_guard<std::mutex> lock(mtx_);
+  workQueue_.push(fn);
+  cond_.notify_one();
```
rdzhabarov: you could notify after unlocking the mutex.

lib/Onnxifi/ThreadPool.h
```diff
@@ -0,0 +1,70 @@
...
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
+
+namespace glow {
+namespace onnxifi {
+
+/// Thread pool for asynchronous execution of generic functions.
+class ThreadPool {
```
rdzhabarov: final

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -402,21 +402,21 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxRunGraph)(
...
+  glowGraph->backend()->runAsync([inputFence, outputFence, glowGraph]() {
+    // Wait for all inputs to be ready.
+    auto waitStatus =
+        GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxWaitEvent)(inputFence->event);
+
+    // If all inputs are ready, run the graph.
+    if (waitStatus == ONNXIFI_STATUS_SUCCESS) {
+      // Run graph.
+      glowGraph->run();
+      // Signal that the output is ready.
```
rdzhabarov: we'd need to signal anyway, in case of failure as well, otherwise C2 will stuck

lib/Onnxifi/ThreadPool.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Queue of work items.
+  std::queue<std::function<void(void)>> workQueue_;
+
+  /// Mutex to coordinate access to the work queue.
+  std::mutex mtx_;
+
+  /// Condition variable to signal to threads when work is added to
+  /// the work queue.
+  std::condition_variable cond_;
```
rdzhabarov: would be more readable if we name this, queueNotEmpty (and then notify and wait for that)

lib/Onnxifi/ThreadPool.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Flag checked by the workers in between work items to determine
+  /// whether they should stop and exit.
+  std::atomic<bool> shouldStop_;
+
+  /// Queue of work items.
+  std::queue<std::function<void(void)>> workQueue_;
+
+  /// Mutex to coordinate access to the work queue.
+  std::mutex mtx_;
```
rdzhabarov: queueMtx_ or something

rdzhabarov:
Looking good!

> Onnxifi runtime: 0.25087404251098633s
> can you also dump what was the time before this change on master.

Ok, now it looks better `Onnxifi runtime: 0.1417860984802246s`.

lib/Onnxifi/ThreadPool.cpp
SplitInfinity: Does it matter? The lock will be automatically unlocked when the `lock_guard` is destroyed.

cmake/modules/GlowDefaults.cmake
SplitInfinity: Sure, I only did this so that I could wrap the invocation of queue work items in a `try..catch`.

lib/Onnxifi/ThreadPool.cpp
rdzhabarov: it does not matter from the correctness perspective, but ... the other thread waiting on the condition var to be signalled, would not be able to make any progress with mtx_ locked anyway. I'm not too sure about exact performance penalties in this case.

lib/Onnxifi/ThreadPool.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // a thread can't wait on the condition variable after checking the
+  // *old* value of shouldStop_.
+  std::unique_lock<std::mutex> lock(workQueueMtx_);
+
+  // Signal to workers to stop.
+  shouldStop_ = true;
+
+  // Notify all worker threads in case any are waiting on the condition
+  // variable.
+  queueNotEmpty_.notify_all();
```
bertmaher: You can release the lock right before the notify_all.  That way the waiter can immediately start executing after it receives the signal.

lib/Onnxifi/Base.h
```diff
@@ -62,8 +64,13 @@ class Backend {
   /// \returns Execution Engine associated with the Backend.
   glow::ExecutionEngine &getEE() { return backendIdPtr_->getEE(); }
 
+  /// Run async using backend thread pool.
+  void runAsync(const std::function<void(void)> &fn);
+
 private:
   BackendIdPtr backendIdPtr_;
+  // ThreadPool instance for the backend.
+  ThreadPool threadPool_;
```
rdzhabarov: we'd need to pass #workers in a ctor of backend and initialize threadPool_ properly.

rdzhabarov:
Awesome! :)

PR #1871, opened 2018-10-15
[Bundle] Fix Resnet50 Bundle to support Placeholder Migration

### LINKED ISSUE #1800: [bundles] Bundles are broken
Looks like bundles are broken again -- this time due to 1057d66a603a401afbbd4dcfaaac8610c8112ab3 🙁 

```
Expected to find variable 'gpu_0_data'
Assertion failed: (mutableWeightVar && "Expected to find a mutable weight variable"), function getMutableWeightVar, file resnet50.cpp, line 267.
/bin/sh: line 1: 25925 Abort trap: 6           ./resnet50 $file
```

gcatron:
*Description*: Updated CPU AllocationInfo and CPU Bundle saver to support Placeholder migration. Memory is now properly allocated and the correct symbol table is emitted. The bundle works again.

*Testing*: Build and run resnet50, vgg19, and zfnet512. They all run without error. *Note vgg19 and zfnet incorrectly label 285 as 281. 

Fixes #1800

### DIFF
--- a/lib/Backends/CPU/AllocationsInfo.cpp
+++ b/lib/Backends/CPU/AllocationsInfo.cpp
@@ -56,18 +56,26 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
     }
   }
 
-  // Allocate addresses for the Placeholders.
-  for (auto PH : ctx.pairs()) {
-    assert(isa<WeightVar>(F->getWeightForNode(PH.first)));
-    auto *w = cast<WeightVar>(F->getWeightForNode(PH.first));
-    auto numBytes = w->getSizeInBytes();
-    size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-    if (!absoluteAddr) {
-      allocatedAddressed_[w] = addr;
-    } else {
+  if (absoluteAddr) {
+    // Allocate addresses for the Placeholders that have payloads defined at
+    // compile-time.
+    // TODO: Remove this branch once Context becomes a parameter of the
+    // CompiledFunction::execute method.
+    for (auto PH : ctx.pairs()) {
+      assert(isa<WeightVar>(F->getWeightForNode(PH.first)));
+      auto *w = cast<WeightVar>(F->getWeightForNode(PH.first));
       // Reuse the address used by the payload.
       allocatedAddressed_[w] =
           PH.second->getUnsafePtr() - static_cast<char *>(nullptr);
+    }
+  } else {
+    // Allocate based on size as reported by the formal type of Placeholders
+    for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+      assert(isa<WeightVar>(F->getWeightForNode(v)));
+      auto *w = cast<WeightVar>(F->getWeightForNode(v));
+      auto numBytes = w->getSizeInBytes();
+      size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
+      allocatedAddressed_[w] = addr;
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -56,19 +56,25 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
-  for (auto PH : ctx.pairs()) {
-    assert(isa<WeightVar>(F->getWeightForNode(PH.first)));
-    auto *w = cast<WeightVar>(F->getWeightForNode(PH.first));
-    auto numBytes = w->getSizeInBytes();
-    size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-    if (!absoluteAddr) {
-      allocatedAddressed_[w] = addr;
-    } else {
+  if (absoluteAddr) {
+
```
opti-mix: Why this empty line?

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -56,19 +56,25 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+    // Allocate addresses for the Placeholders.
+    for (auto PH : ctx.pairs()) {
+      assert(isa<WeightVar>(F->getWeightForNode(PH.first)));
+      auto *w = cast<WeightVar>(F->getWeightForNode(PH.first));
       // Reuse the address used by the payload.
       allocatedAddressed_[w] =
           PH.second->getUnsafePtr() - static_cast<char *>(nullptr);
     }
+  } else {
+    // Allocate based on size of Placeholder
```
opti-mix: // Allocate based on size as reported by the formal type of Placeholders

nadavrot:
Nice!

opti-mix:
@gcatron Very nice! Thanks for fixing this feature!

opti-mix:
@gcatron BTW, have you tested both quantized and non-quantized variants of bundles?

gcatron:
> @gcatron BTW, have you tested both quantized and non-quantized variants of bundles?

Good catch! I had only tested quantized however I just tested non-quantized and they are also functioning properly.

gcatron:
With the same observation regarding 285 predicted as 281 for zfnet and vgg19

lib/Backends/CPU/BundleSaver.cpp
```diff
@@ -95,7 +95,7 @@ void BundleSaver::emitSymbolTable() {
   llvm::SmallVector<llvm::Constant *, 128> entries;
   // Iterate over all weights and record information about their names, offset,
   // size and kind.
-  for (auto &v : F_->getGraph()->getParent()->getConstants()) {
+  for (auto &v : F_->getGraph()->getParent()->getPlaceholders()) {
```
rdzhabarov: I might be not up to date on this, but `Placeholders` were only for `inputs/outputs` of the network and
`Constants` are for weights.

Given that the change seems strange. Did we change the behavior for placeholders/constants?

bertmaher:
Re: the 285 vs. 281 issue: the classes are "egyptian cat" and "tabby cat", which are pretty dang hard to distinguish.  Personally I think cat_285.png is a tabby (281), but clearly I'm just failing the Turing test:-D

rdzhabarov:
> With the same observation regarding 285 predicted as 281 for zfnet and vgg19

iirc, that what was the case even in the past. You could run regular imageclassifier (tests/run.sh) and that should produce the same results.

lib/Backends/CPU/BundleSaver.cpp
opti-mix: @rdzhabarov  We need to dump here a "symbol table" for inputs/outputs. `getConstants` was a wrong change introduced during the Placeholders refactoring.

rdzhabarov: Yes, getConstants are not for inputs/outputs, but for static weights.

PR #1872, opened 2018-10-15
[GraphOptimizer] Do transpose and reshape optimizations first

aturetsk:
*Description*:
This is a cleanup rather than a functional change.
Transpose and Reshape nodes may prevent other optimizations from triggering (e.g. BatchNorm + Convolution merge, Transpose + MatMul merge). As optimizeTranspose and optimizeReshape do not depend on other optimizations, it makes sense to do them first, so that Transposes and Reshapes would be optimized out and let other optimizations work.
I did some optimization re-ordering related to reshape/transpose in my previous PRs (#1680, #1782). This is just a cleaner way to handle it. I think it's better to put the optimizations in the beginning and have an explicit comment explaining why we are doing this, rather than have implicit optimization ordering.

*Testing*:
No new tests, since it's just a clean up.

*Documentation*:
N/A

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1323,7 +1323,7 @@ static void optimizeArithmeticNodes(Function *F) {
 }
 
 /// Statically transpose Constants.
-static void optimizeTranspose(Function *F) {
+static void transposeConstants(Function *F) {
   auto &nodes = F->getNodes();
 
   for (auto &node : nodes) {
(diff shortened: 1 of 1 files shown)


nadavrot:
LGTM. Can you please mention the name of the networks that become faster? For example, "with this change TensorFlow's AlexNet is now optimized".

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2092,6 +2092,13 @@ void glow::optimize(Function *F, CompilationMode mode) {
     DCE(F);
   }
 
+  // Reshapes and transposes can prevent other optimizations from triggering, so
+  // try to optimize them out first.
+  optimizeReshape(F);
+  if (mode == CompilationMode::Infer) {
+    optimizeTranspose(F);
```
rdzhabarov: what is the reason transposes are not eliminated in training?

aturetsk:
@nadavrot, this is just a cleanup, so there is no performance improvement. I just tried to make it more obvious why some optimizations are done after other optimizations, by re-organizing code a bit.

@rdzhabarov, renamed optimizeTranspose to transposeConstants.

PR #1875, opened 2018-10-16
[NFC/cleanup] Remove more entries of variables.

rdzhabarov:
*Description*:
Removing references to variables when appropriate.

*Testing*:
unit tests

*Documentation*:
N/A

Related to https://github.com/pytorch/glow/issues/1334

### DIFF
--- a/lib/Optimizer/Partition.cpp
+++ b/lib/Optimizer/Partition.cpp
@@ -64,9 +64,9 @@ public:
   Function *operator[](Node *n) { return nodeToFunction_[n]; }
 };
 
-/// If \p node has a single input that is not a variable, return it.  Otherwise
+/// If \p node has a single input that is not a storage, return it.  Otherwise
 /// return nullptr.
-Node *singleNonVariableInput(Node *node) {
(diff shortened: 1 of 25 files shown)


yinghai:
Noob question: what's the reasoning of changing `variable` to `input`?

rdzhabarov:
That's a long discussion. See https://github.com/pytorch/glow/issues/1334 for details.
This PR is cleaning up what was left after migration.

lib/Optimizer/Partition.cpp
```diff
@@ -96,7 +96,7 @@ NodeFunctionMap selectBasicBlockPartitions(Function *F) {
 
     // If node has only one input, and that input has only one output, place it
     // in the same partition.
-    auto *in = singleNonVariableInput(node);
+    auto *in = singleNonStorageInput(node);
```
jfix71: Just to be sure -- was the intention here for any Storage node? Or was it really intended more for what are now Placeholder inputs? @bertmaher

rdzhabarov:
I like `updateInputPlaceholders`, more explicit to what's going on.

PR #1877, opened 2018-10-16
[WIP][bundles] Add lenet_mnist bundle example, and run in travis CI

jfix71:
*Description*: Add a lenet_mnist bundle example

*Testing*: I tested manually. I would like to add this to our CI to prevent future bundle breakage, since the weights are small and the model runs quickly.

*Documentation*: N/A


.travis.yml
```diff
@@ -46,6 +52,24 @@ jobs:
           -DCMAKE_BUILD_TYPE=Debug -DGLOW_WITH_OPENCL=OFF -DGLOW_WITH_CPU=ON
           -DCMAKE_PREFIX_PATH=/usr/lib/llvm-6.0/include/
           ../
+      script:
+        - function keep_alive() { while true; do echo -en "\a"; sleep 540; done }
```
nadavrot: Can we push all of this stuff into a script?  We should simplify the travis file.

opti-mix:
@jfix71 I like the idea! To be on the safe side we should check that it works both in quantized and non-quantized modes.

jfix71:
Now testing both quantized and unquantized bundles.

Also refactored travis to run the tests via `run_tests.sh`

.travis/run_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+    run_and_check_bundle NO
+    ;;
+
+*)
+    echo "Error, '${TEST_NAME}' not valid mode -- must be one of {ASAN, DEBUG}."
+    exit 1
+    ;;
+esac
+
+exit 0
```
rdzhabarov: do you need to spell that out explicitly?

.travis/run_tests.sh
```diff
@@ -0,0 +1,51 @@
...
+    CTEST_PARALLEL_LEVEL=2 ninja ${1} || ( cat Testing/Temporary/LastTest.log && exit 1 )
+}
+
+# Pass in YES/NO for QUANTIZE
+run_and_check_bundle() {
+    echo "Checking lenet_mnist bundle with QUANTIZE=${1}"
+    cd ${GLOW_SRC}/examples/bundles/lenet_mnist/
+    ( QUANTIZE=${1} make &> raw_results.txt ) || ( cat raw_results.txt && exit 1 )
+    ( tail -n72 raw_results.txt | grep -F "Result: " > results.txt ) || ( cat raw_results.txt && exit 1 )
+    diff results.txt ${GLOW_SRC}/.travis/lenet_mnist_expected_output.txt
```
rdzhabarov: if diff returns something then we'll return error code back to Travis, right? (no need to explicitly exit 1)

.travis/run_tests.sh
jfix71: I guess not!

.travis/run_tests.sh
jfix71: Yeah that was the intention, I believe it should work like that. I'll verify. I suppose I could also explicitly `exit 1` but it should work correctly like that.

jfix71: Yep, looks like it works as expected: https://travis-ci.org/pytorch/glow/jobs/442479011

PR #1880, opened 2018-10-17
[ONNXifi] Add support for ReplaceNaN operator

### LINKED ISSUE #1704: [New operator] ReplaceNaN

SplitInfinity:
**Description**
This commit adds support for the `ReplaceNaN` operator to the ONNX model
importer. Most of the work to support this operator was done in an
earlier commit that added support for the Caffe2 loader. Consequently,
this commit only includes code that parses inputs, outputs, and arguments
from an .onnxtxt and calls `Graph::createReplaceNaN` to implement the
operation.

**Testing**
This commit adds a unit test to `onnxImporterTest.cpp` to test loading
this operator from a .onnxtxt file.

**Fixes**
This commit fixes #1704.

### DIFF
--- a/tests/unittests/onnxImporterTest.cpp
+++ b/tests/unittests/onnxImporterTest.cpp
@@ -295,3 +295,37 @@ TEST(onnx, importLengthsToRanges) {
   ASSERT_TRUE(LTR);
   ASSERT_TRUE(llvm::isa<Placeholder>(LTR->getLengths()));
 }
+
+/// Test loading ReplaceNaN op from an ONNX model.
+/// Test with arg value = 1.0.
+TEST(onnx, importReplaceNaN) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string netFilename("tests/models/onnxModels/replaceNaN.onnxtxt");
+
+  Context ctx;
+  Placeholder *output;
+  Tensor x(ElemKind::FloatTy, {3, 3});
+
+  {
+    ONNXModelLoader onnxLD(netFilename, {"x"}, {&x.getType()}, *F);
+    output = onnxLD.getSingleOutput();
+    ctx.allocate(mod.getPlaceholders());
+    updateInputPlaceholdersByName(ctx, &mod, {"x"}, {&x});
+  }
+
+  // Verify structure: Input, IsNan, Splat -> Select -> Save.
+  ASSERT_EQ(mod.getPlaceholders().size(), 2);
+  ASSERT_EQ(F->getNodes().size(), 4);
+  auto *save = getSaveNodeFromDest(output);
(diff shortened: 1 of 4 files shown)


SplitInfinity:
@rdzhabarov 

Hmm, there is an ASAN failure that I'm pretty sure is not random. It looks like the operator isn't working at all during those failures (i.e. there are still NaNs in the result), but I am unable to reproduce this locally even after using the same cmake options. I also have operator tests for `ReplaceNaN` for both CPU and Interpreter backends which passed in this run, which I would have expected to fail if the operator wasn't working correctly.

tests/unittests/onnxImporterTest.cpp
```diff
@@ -132,3 +132,50 @@ TEST(onnx, importClip) {
...
+  // Output shape and input shape should match.
+  EXPECT_TRUE(result.dims().vec() == x.dims().vec());
+
+  // All NaNs in the input should be replaced with 1.0f and all other values
+  // should match.
+  for (size_t i = 0; i < result.size(); i++) {
+    auto resultValue = result.raw(i);
+    auto inputValue = XH.raw(i);
+
+    EXPECT_FLOAT_EQ(resultValue, std::isnan(inputValue) ? 1.0f : inputValue);
```
rdzhabarov: very interesting. Can you `std::cout` resultValue and `std::isnan(inputValue) ? 1.0f : inputValue` before expect eq.

SplitInfinity: Tried this locally and didn't find anything out of the ordinary:

```
...
Result value: 1
Expected value: 1
Result value: -2.18714
Expected value: -2.18714
Result value: 1
Expected value: 1
Result value: 2.01005
Expected value: 2.01005
Result value: 1
Expected value: 1
Result value: 2.81321
Expected value: 2.81321
Result value: 1
Expected value: 1
Result value: -1.6738
Expected value: -1.6738
Result value: 1
Expected value: 1
...
```

rdzhabarov: Can you do the same on CI, so we could look at the output.

tests/unittests/onnxImporterTest.cpp
```diff
@@ -206,3 +206,54 @@ TEST(onnx, importBatchBoxCox) {
     }
   }
 }
+
+/// Test loading ReplaceNaN op from an ONNX model.
+/// Test with arg value = 1.0.
+TEST(onnx, importReplaceNaN) {
```
rdzhabarov: we probably should understand what's happening with ASAN test. But for loader test it's sufficient to check graph structure after loading. Actual operator test is run in the operatorTest.

bertmaher:
Fwiw, it looks like we've stopped running ASAN on Travis.  I wonder if you rebase, if this will repro on CircleCI.

PR #1881, opened 2018-10-17
Add benchmark for convolution

### LINKED ISSUE #1616: Create a convolution benchmark
We have a benchmark harness for matrix multiplication; it would be nice to have one for convolution as well, so we can check for performance regressions (or improvements).  It'll need to have more parameters, of course.

hpwooten:
Description:
This is a benchmark for convolution in the same style as GemmBench. There are lots of parameters to convolution, and those parameters are subject to a set of requirements. I’ve tried to enforce those requirements in the creation of each benchmark instance.  Convolution also takes a while to run if you have large parameters, so the number of reps is smaller than for GemmBench. We may need to tweak the test cases. Another key difference is that I'm reporting the time in seconds as opposed to gflops.
Testing:
To test, run the ConvBench binary in the tests subdirectory of the build output.
[Fixes #1616 ]

### DIFF
--- a/tests/benchmark/ConvBench.cpp
+++ b/tests/benchmark/ConvBench.cpp
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <cstdlib>
+#include <random>
+
+#include "Bench.h"
+
+using namespace glow;
+
+extern "C" {
+// Forward declare functions from libjit.
+extern void libjit_convolution_f(float *outW, const float *inW, const float *filterW,
+                     const float *biasW, const size_t *outWdims,
+                     const size_t *inWdims, const size_t *filterWdims, 
+                     const size_t *biasWdims, const size_t *kernelSizes, 
+                     const size_t *strides, const size_t *pads, 
+                     size_t group, unsigned depthUnroll);
+}
+
+/// Benchmark a convolution with specified parameters on square inputs.
+class ConvBench : public Benchmark {
+  /// Matrices
+  std::vector<float> outW;
+  std::vector<float> inW;
+  std::vector<float> filterW;
+  std::vector<float> biasW;
+
+  /// Dimensions
+  // [batch, h, w, channels]
+  size_t outWdims[4];
+  size_t inWdims[4];
+  // [outputChannels, h, w, inputChannels]
+  size_t filterWdims[4];
+
+  /// Parameters
+  size_t kernelSizes[2];
+  size_t strides[2];
+  size_t pads[2];
+  size_t group;
+  unsigned depthUnroll;
+
+
+public:
+  ConvBench(size_t inputBatch, size_t inputEdgeSize, size_t inputChannels, size_t filterMultiplier, 
+            size_t kernelSize, size_t stride, size_t pad, size_t group)
+      : kernelSizes{kernelSize, kernelSize}, strides{stride, stride}, 
+      pads{pad, pad}, group(group) {
+
+        inWdims[0] = inputBatch;  
+        inWdims[1] = inputEdgeSize;
+        inWdims[2] = inputEdgeSize;
+        inWdims[3] = inputChannels;
+
+        filterWdims[0] = filterMultiplier * group;
+        filterWdims[1] = kernelSize;
+        filterWdims[2] = kernelSize;
+        filterWdims[3] = inWdims[3] / group;
+
+        size_t outEdgeSize = ((inputEdgeSize + (2 * pad) - kernelSize) / stride) + 1;
+        outWdims[0] = inWdims[0];
+        outWdims[1] = outEdgeSize;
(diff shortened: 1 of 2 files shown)


tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+int main() {
+  constexpr int reps = 10;
+  printf("inputBatch, inputEdgeSize, inputChannels, filterMultiplier, kernelSize, stride, pad, group, bestInSeconds\n");
+
+  for (size_t inputBatch : {1, 3}) {
+    for (size_t inputEdgeSize : {1, 224}) {
+      for (size_t inputChannels : {64, 128, 1024}) {
+        for (size_t filterMultiplier : {1, 8}) {
+          for (size_t kernelSize : {1, 5}) {
```
bertmaher: Mind adding a 3x3 kernel size test?  That's all over ResNet so I'd like to make sure it's benchmarked.

tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,155 @@
...
+            for (size_t stride : {1, 4}) {
+              for (size_t pad : {0, 2}) {
```
bertmaher: Along with the the 3x3 kernelSize, pad=1 is common.  If you want to cut down the # of iterations, it's probably fine to have no loop here and simply set `pad = kernelSize / 2`.

tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+};
+
+int main() {
+  constexpr int reps = 10;
+  printf("inputBatch, inputEdgeSize, inputChannels, filterMultiplier, kernelSize, stride, pad, group, bestInSeconds\n");
+
+  for (size_t inputBatch : {1, 3}) {
+    for (size_t inputEdgeSize : {1, 224}) {
```
bertmaher: 7, 14, 28, and 56 are all interesting, especially with the higher channel numbers.  Can we get one of those in this loop?

tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,155 @@
...
+/// Benchmark a convolution with specified parameters on square inputs.
+class ConvBench : public Benchmark {
+  /// Matrices
+  std::vector<float> outW;
+  std::vector<float> inW;
+  std::vector<float> filterW;
+  std::vector<float> biasW;
+
+  /// Dimensions
+  // [batch, x, y, channels]
```
rdzhabarov: nit: `h` and `w` would be common names for dims

tests/benchmark/ConvBench.cpp
hpwooten: Yes, I was concerned about number of iterations. Thanks for letting me know that it's okay to calculate pad based on kernelSize. I'll do that.

tests/benchmark/ConvBench.cpp
hpwooten: Do we also want 5x5 or can I just swap that out for 3?

hpwooten:
The runtime is currently between 3.5-4 minutes on my MacBook Pro.

rdzhabarov:
> The runtime is currently between 3.5-4 minutes on my MacBook Pro.

Seems very reasonable, given we do not run this on Travis CI.

tests/benchmark/ConvBench.cpp
bertmaher: Yeah, just 3x3 is fine by me, thanks.

tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,153 @@
...
+    size_t inSize = mapMult(inWdims, 4);
+    size_t filterSize = mapMult(filterWdims, 4);
+    size_t biasSize = filterWdims[0];
+
+    outW.resize(outSize);
+    inW.resize(inSize);
+    filterW.resize(filterSize);
+    biasW.resize(biasSize);
+
+    randomize(inSize, inW.data());
```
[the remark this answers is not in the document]
bertmaher: Probably copypasta from my implementation of GemmBench, which originally used malloc'ed arrays.  This STL thing is probably a fad, after all :-).

bertmaher:
Thanks very much for doing this!

tests/benchmark/ConvBench.cpp
```diff
@@ -0,0 +1,153 @@
...
+        a[i] = dis(gen);
+    }
+  }
+};
+
+int main() {
+  constexpr int reps = 10;
+  printf("inputBatch, inputEdgeSize, inputChannels, filterMultiplier, kernelSize, stride, pad, group, bestInSeconds\n");
+
+  for (size_t inputBatch : {1, 3}) {
```
nadavrot: @hpwooten @bertmaher @rdzhabarov 

Can we use realistic convolution values from known networks?

This is resnet50:

https://fb.quip.com/9MqKASIpqBSU#AaaACAaHTye

bertmaher: That's what these mostly are?  Granted some of them are weird like 224x224, 1024 channels, but the range of each parameter is drawn from the parameters that appear in ResNet.

bertmaher: Ugh, wait.  The `filterMultiplier` thing is kind of messed up.  That should just be output channels.  Willfix.

nadavrot: I wonder if instead of having a cartesian product of all possible combinations we should just have a list of filters from resnet, vgg, etc.

bertmaher: I'm not opposed to that, but it's not unreasonable to sweep through some parameters either.

jfix71:
Cool! It would be great if we could next extend this to also benchmark `libjit_convDKKC8_f()`. Perhaps we could extract and then call into the logic in `LLVMIRGen.cpp` for deciding `pixelScanFirst`, `numDepthRegs`, `sizeGroupY`, etc.

Issue #1885, opened 2018-10-18
[New operator] LengthsToRanges


artemrakhov:
We need to support `LengthsToRanges` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#lengthstoranges

One has to create new Graph-level Node (`LengthsToRangesNode`), create new low-level IR instruction, and implement it for Interpreter and CPU backends, also adding a unit test to `OperatorTest.cpp`.

PR #1891, opened 2018-10-19
[graph] Implement LengthsToRanges (interp only)

bertmaher:
*Description*: That's it.
*Testing*: New operatorTest.
*Documentation*: Docstrings
Partial fix for #1885

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -507,6 +507,12 @@ public:
                                  NodeValue weights, NodeValue indices,
                                  NodeValue lengths);
 
+  /// Given a vector of segment lengths, calculates offsets of each segment and
+  /// packs them next to the lengths. For the input vector of length N the
+  /// output is a Nx2 matrix with (offset, lengths) packaged for each segment.
+  LengthsToRangesNode *createLengthsToRanges(llvm::StringRef name,
+                                             NodeValue lengths);
+
   SaveNode *createSave(llvm::StringRef name, NodeValue input);
   SaveNode *createSave(llvm::StringRef name, NodeValue input,
                        Placeholder *output);
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -591,6 +591,18 @@ void SparseLengthsWeightedSumNode::verify() const {
          "Weights and Indices must have the same size");
 }
 
+void LengthsToRangesNode::verify() const {
+  assert(getResult().getElementType() == getLengths().getElementType() &&
+         "Mismatched element types");
+  assert(getLengths().getElementType() == ElemKind::Int64ITy &&
+         "Lengths must have index type");
+  assert(getLengths().dims().size() == 1 && "Lengths must be 1D vector");
+  assert(getResult().dims().size() == 2 && "Ranges must be 2D vector");
+  assert(getResult().dims()[0] == getLengths().dims()[0] &&
+         "Lengths and Ranges must have the same outer dimensions");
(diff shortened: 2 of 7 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -591,6 +591,18 @@ void SparseLengthsWeightedSumNode::verify() const {
          "Weights and Indices must have the same size");
 }
 
+void LengthsToRangesNode::verify() const {
+  assert(getResult().getElementType() == getLengths().getElementType() &&
+         "Mismatched element types");
+  assert(getLengths().getElementType() == ElemKind::Int64ITy &&
+         "Lengths must have index type");
```
beicy: According to the def of Caffe2 op , the input is `"lengths", "1D tensor of int32 segment lengths."`. Just wondering why we need to use "Int64ITy" here?

bertmaher: Fair question, I admit I did not really think about the bit width here :-).  IIUC these values are indices, and the only index type we have is Int64ITy (I think Int32QTy is intended for quantized data?).  So I think int64 is safe to use here, if a bit wasteful of memory.

bertmaher: Maybe @yinghai could tell us to what extent we should worry about preserving int32 vs int64 indices here.

yinghai: In terms of bit width it's fine. But input tensor from C2 is going to be int32 (https://github.com/pytorch/pytorch/blob/6476e4598ca55626b9d8a28074fa65e10f10f2bb/caffe2/operators/utility_ops.h#L737), hence after ONNX conversion, you are going to get a int32 tensor when you do `onnxSetGraphIO`, so this is an issue?

beicy: " Int32QTy is intended for quantized data ? " @rdzhabarov could you please verify it? I know if we create a type with "Int32QTy", the scale and offset should be provided. But I think we could just provide some dummy scale and offset for a regular int32_t :P If not, we might need to added some thing for the int32_t type.

yinghai: Hah, there is actually an issue because Glow importer doesn't support `int32` yet: https://github.com/pytorch/glow/blob/777f01c7f25fa88f77462b12904b0e89c29d84b9/lib/Importer/ONNXIFIModelLoader.cpp#L26-L40

Well, at C2 side, we didn't handle integer tensor either: https://github.com/pytorch/pytorch/blob/master/caffe2/opt/onnxifi_transformer.cc#L80-L81. So there is some work to do. :)

jfix71: @beicy Yeah we technically could do something hacky like that with `Int32QTy`, but if we want to use the data as indices we really should add `Int32ITy`.

include/glow/Graph/Graph.h
```diff
@@ -507,6 +507,9 @@ class Function final : public Named {
                                  NodeValue weights, NodeValue indices,
                                  NodeValue lengths);
```
jfix71: nit: Doxygen comments

lib/Graph/Nodes.cpp
rdzhabarov: > " Int32QTy is intended for quantized data ? "

True

I agree with Jordan.

beicy: In that case, I will add "Int32ITy" for this one and RowwiseQuantizedFC and "UInt8QTy" for quantized uint8_t activations.

bertmaher:
I created #1895 to introduce Int32ITy, but I don't think it should block this particular PR.  We can migrate to narrower types in the future.

PR #1892, opened 2018-10-19
Silence Apple's warning about the deprecation of OpenCL

opti-mix:
*Description*:
Starting with MacOS Mojave, using OpenCL results in a bunch of compiler warnings, because Apple is going to deprecate OpenCL. The warnings suggest defining CL_SILENCE_DEPRECATION to silence any such warnings.

*Testing*:
ninja test

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -15,6 +15,9 @@
  */
 #define DEBUG_TYPE "opencl"
 
+// Silence Apple's warning about the deprecation of OpenCL.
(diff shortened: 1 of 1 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -15,6 +15,9 @@
  */
 #define DEBUG_TYPE "opencl"
 
+// Silence Apple's warning about the deprecation of OpenCL.
+#define CL_SILENCE_DEPRECATION
```
nadavrot: ❤️

PR #1893, opened 2018-10-19
[trivial] Remove non-ASCII in docs/IR.md

bertmaher:


bertmaher:
For those interested, I found this because I was fiddling with using `xargs` to filter file lists:
```
git ls-tree --name-only -r HEAD | xargs -I% sh -c "file '%' | grep -q -v ASCII && echo '%'"
```

PR #1894, opened 2018-10-19
 [caffe2] Import LengthsToRanges

bertmaher:
*Description*: N/A
*Testing*: New caffe2ImporterTest
*Documentation*: N/A
Partial fix for #1885

Depends on #1891, please review only the top commit.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -594,6 +594,7 @@ void SparseLengthsWeightedSumNode::verify() const {
 void LengthsToRangesNode::verify() const {
   assert(getResult().getElementType() == getLengths().getElementType() &&
          "Mismatched element types");
+  // TODO: Lengths should be Int32ITy once that type is implemented.
   assert(getLengths().getElementType() == ElemKind::Int64ITy &&
          "Lengths must have index type");
   assert(getLengths().dims().size() == 1 && "Lengths must be 1D vector");
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -464,6 +464,13 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "LengthsToRanges") {
+    auto in = getNodeValueOrCreateConstantByName(op.input(0));
+    auto *node = G_.createLengthsToRanges(opName, in);
+    addNodeAsOutput(op, node);
+    return;
+  }
+
   if (typeName == "ExpandDims") {
     auto in = getNodeValueOrCreateConstantByName(op.input(0));
     auto dims = getShape(dict["dims"]);
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -1255,3 +1255,36 @@ TEST(caffe2, EQ1D) {
   // Graph has two inputs and one output.
   EXPECT_EQ(mod.getPlaceholders().size(), 3);
 }
+
+// Test loading a LengthsToRanges operator.
+TEST(caffe2, LengthsToRanges) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename(
+      "tests/models/caffe2Models/lengths_to_ranges.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/lengths_to_ranges_init_net.pbtxt");
+
+  Placeholder *output;
+
+  // Destroy the loader after the graph is loaded
+  {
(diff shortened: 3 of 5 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1747,6 +1747,18 @@ void InterpreterFunction::fwdSparseLengthsWeightedSumInst(
   }
 }
 
+void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {
```
rdzhabarov: can you add CPU support?

bertmaher:
Yep, going to write that diff tonight!

On Oct 19, 2018, at 5:52 PM, Roman Dzhabarov <notifications@github.com<mailto:notifications@github.com>> wrote:


@rdzhabarov commented on this pull request.

________________________________

In lib/Backends/Interpreter/InterpreterNodes.cpp<https://github.com/pytorch/glow/pull/1894#discussion_r226810470>:

> @@ -1747,6 +1747,18 @@ void InterpreterFunction::fwdSparseLengthsWeightedSumInst(
   }
 }

+void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {


can you add CPU support?

lib/Graph/Nodes.cpp
```diff
@@ -591,6 +591,18 @@ void SparseLengthsWeightedSumNode::verify() const {
          "Weights and Indices must have the same size");
 }
 
+void LengthsToRangesNode::verify() const {
+  assert(getResult().getElementType() == getLengths().getElementType() &&
+         "Mismatched element types");
+  assert(getLengths().getElementType() == ElemKind::Int64ITy &&
```
rdzhabarov: will it be int32ity eventually? (if so maybe add TODO, so that we could fix it real quick later).

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -464,6 +464,13 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "LengthsToRanges") {
```
rdzhabarov: In the ONNX loader it would be great to make all these
```
assert(getLengths().dims().size() == 1 && "Lengths must be 1D vector");
  assert(getResult().dims().size() == 2 && "Ranges must be 2D vector");
  assert(getResult().dims()[0] == getLengths().dims()[0] &&
         "Lengths and Ranges must have the same outer dimensions");
  assert(getResult().dims()[1] == 2 && "Inner dimension of Ranges must be 2");
```
checks in the loader and return false if anything fails.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1255,3 +1255,52 @@ TEST(caffe2, EQ1D) {
...
+  EXPECT_EQ(mod.getPlaceholders().size(), 1);
+
+  // Compile and run the model.
+  auto *res = ctx.get(output);
+  EE.compile(CompilationMode::Infer, F, ctx);
+  EE.run();
+
+  // Check that the output tensor is the same as the expected output.
+  // auto result = res->getHandle<int64_t>();
+  Tensor expected(ElemKind::Int64ITy, {4, 2});
```
rdzhabarov: that might be debatable, but I think the idea of loader tests is to test the loader, e.g., check that loaded graph contains certain Glow nodes.
this is not always possible and then we do some run checks, I think here it's fine just to check structure.

lib/Importer/Caffe2ModelLoader.cpp
bertmaher: Interesting point here.  All the verification stuff is done very late in the game, which is really problematic for ONNX.

tests/unittests/caffe2ImporterTest.cpp
bertmaher: Funny you should point this out :-).  I was in fact specifically interested in testing execution, because of the int32_t vs. int64_t issue; I was concerned that if we were somehow reading the init_net as int32, we could do the wrong computation here.  (I think that's impossible, given we don't even have the machinery to read int32 tensors :) ).

bertmaher: I'll go ahead and kill the execution stuff, anyways.  It probably doesn't make sense given how things work now.

lib/Importer/Caffe2ModelLoader.cpp
rdzhabarov: I do not think we'd need to care for C2 early stage verifications (not going to be used in prod anyway).
It's just ONNXModelLoader needs to handle this properly early on.

lib/Graph/Nodes.cpp
```diff
@@ -594,6 +594,7 @@ void SparseLengthsWeightedSumNode::verify() const {
 void LengthsToRangesNode::verify() const {
   assert(getResult().getElementType() == getLengths().getElementType() &&
          "Mismatched element types");
+  // TODO: Lengths should be Int32ITy once that type is implemented.
```
rdzhabarov: Int32ITy change is already in :)

PR #1897, opened 2018-10-19
[Caffe2] Add support for SparseToDense operator

### LINKED ISSUE #1702: [New operator] SparseToDense
We need to support `SparseToDense` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#sparsetodense

Caveat here is that the operator essentially does:
```
for i in 0..n:
  result[ids[i]] += values[i]
```
Size of `result` is defined as `max(ids[i])+1`. This leads to dynamic memory allocation: we won't know the size until runtime.

SplitInfinity:
**Description**
This commit adds support for the `SparseToDense` Caffe2 operator. This
operator takes a sparse representation of a tensor specified by an array
of indices and an array of tensor values and converts it into a dense
representation. In this dense representation, `values[i]` is stored at
`indices[i]`, and the rest of the dense representation is zero-filled.
Indices can appear twice in the sparse representation, and in this case, all
corresponding values are summed and stored in dense representation.

This operator is implemented as its own Glow node and instruction.

**Testing**
This commit adds tests for this operator to `caffe2ImporterTest.cpp` and
`OperatorTest.cpp`.

**Fixes**
This commit fixes #1702.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -513,6 +513,19 @@ public:
   LengthsToRangesNode *createLengthsToRanges(llvm::StringRef name,
                                              NodeValue lengths);
 
+  /// Implements an operation that converts the sparse representation given by
+  /// the pair of \p indices and \p values into a dense representation.
+  /// This representation contains each value of \p values at the corresponding
+  /// index given by \p indices. All indices that are not present in \p indices
+  /// are filled with zeroes. \p indices can contain duplicates, and in this
+  /// case, the corresponding values in \p values are added.
+  ///
+  /// \p dataToInferDim acts as a hint about the shape of the output. The first
+  /// dimension of the output is the first dimension of this tensor.
+  SparseToDenseNode *createSparseToDense(llvm::StringRef name,
+                                         NodeValue indices, NodeValue values,
+                                         NodeValue dataToInferDim);
+
   SaveNode *createSave(llvm::StringRef name, NodeValue input);
   SaveNode *createSave(llvm::StringRef name, NodeValue input,
                        Placeholder *output);
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1759,6 +1759,51 @@ void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {
   }
 }
 
+void InterpreterFunction::fwdSparseToDenseInst(const SparseToDenseInst *I) {
+  auto out = getTensor(I->getDest());
+  auto indices = getTensor(I->getIndices());
+  auto values = getTensor(I->getValues());
+
+  out->zero();
+
+  auto IH = indices->getHandle<int64_t>();
+
+  size_t numIndices = indices->dims()[0];
+  size_t numOutDims = out->dims().size();
+
+  // Convert sparse representation to dense representation by taking
+  // slices of output and values and accumulating the value slice into
+  // the output slice.
+
+  // Dimensions and offsets for the output and values slices. sliceDims
+  // will always be {1, [rest of output dimensions]} since the first dimension
+  // is the index in this operation. sliceOffsets will be {indices[j], 0, ...}
+  // for the output slice and {j, 0, ...} for the values slice so that the
+  // slice at index j gets mapped to index indices[j] in the dense
+  // representation.
+  ShapeVector sliceDims(out->dims().begin(), out->dims().end());
+  ShapeVector sliceOffsets(numOutDims, 0);
+  sliceDims[0] = 1;
+
+  for (size_t j = 0; j < numIndices; ++j) {
+    // Create values slice with offsets {j, 0, ...}.
+    sliceOffsets[0] = j;
+    auto VS = values->getUnowned(sliceDims, sliceOffsets);
+    auto VSH = VS.getHandle<float>();
+
+    // Create output slice with offsets {indices[j], 0, ...}.
+    sliceOffsets[0] = IH.at({j});
+    auto OS = out->getUnowned(sliceDims, sliceOffsets);
+    auto OSH = OS.getHandle<float>();
+
+    // Accumulate values slice into output slice.
+    size_t outputSliceSize = OS.size();
+    for (size_t k = 0; k < outputSliceSize; ++k) {
+      OSH.raw(k) += VSH.raw(k);
+    }
+  }
+}
+
 //===----------------------------------------------------------------------===//
 //                Instructions used by RNN
 //===----------------------------------------------------------------------===//
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1265,6 +1265,18 @@ LengthsToRangesNode *Function::createLengthsToRanges(llvm::StringRef name,
   return addNode(new LengthsToRangesNode(name, outTy, lengths));
 }
 
+SparseToDenseNode *Function::createSparseToDense(llvm::StringRef name,
+                                                 NodeValue indices,
+                                                 NodeValue values,
+                                                 NodeValue dataToInferDim) {
+  // The dimensions of the output are the same as the values tensor except for
+  // the first dimension, which should match that of dataToInferDim.
+  ShapeVector outDims(values.dims().begin(), values.dims().end());
+  outDims[0] = dataToInferDim.dims()[0];
+  auto outTy = getParent()->uniqueTypeWithNewShape(values.getType(), outDims);
+  return addNode(new SparseToDenseNode(name, outTy, indices, values));
+}
+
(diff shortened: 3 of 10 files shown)


SplitInfinity:
@artemrakhov 

Could you take a look before I merge?

rdzhabarov:
Thanks for implementing SparseToDense.

> Should I add a CPU backend implementation as well?

that's required for ONNXIFI, do not worry about OpenCL support though.

tools/ClassGen/NodeGen.cpp
```diff
@@ -318,6 +318,22 @@ int main(int argc, char **argv) {
...
 
+  // clang-format off
+  BB.newNode("SparseToDense")
+      .addInput("Indices")
+      .addInput("Values")
+      .addResultFromCtorArg()
+      .setDocstring(
+          "Converts the sparse representation specified by the pair "
+          "(Indices, Values) into a dense one. This dense "
+          "representation is as large as the maximum index in Indices "
```
artemrakhov: `As large as the maximum index` <- this is not true. We use 3rd input to figure out output size.

tools/ClassGen/NodeGen.cpp
```diff
@@ -318,6 +318,22 @@ int main(int argc, char **argv) {
...
+          "and contains each vector from Values at the corresponding "
```
artemrakhov: each vector -> each value

lib/Graph/Graph.cpp
```diff
@@ -1265,6 +1265,19 @@ LengthsToRangesNode *Function::createLengthsToRanges(llvm::StringRef name,
...
+SparseToDenseNode *Function::createSparseToDense(llvm::StringRef name,
+                                                 NodeValue indices,
+                                                 NodeValue values,
+                                                 NodeValue dataToInferDim) {
+  assert((indices.dims()[0] == values.dims()[0]) &&
+         "First dimension of indices and values must match");
+  assert((indices.dims().size() == 1) && "Indices must be a 1D vector");
+
+  auto outTy = getParent()->uniqueTypeWithNewShape(
+      values.getType(), {dataToInferDim.dims()[0], values.dims()[1]});
```
artemrakhov: What if `values` has 1 dimension or more than 2 dimensions?

lib/Graph/Graph.cpp
```diff
@@ -1265,6 +1265,19 @@ LengthsToRangesNode *Function::createLengthsToRanges(llvm::StringRef name,
   return addNode(new LengthsToRangesNode(name, outTy, lengths));
 }
 
+SparseToDenseNode *Function::createSparseToDense(llvm::StringRef name,
+                                                 NodeValue indices,
+                                                 NodeValue values,
+                                                 NodeValue dataToInferDim) {
+  assert((indices.dims()[0] == values.dims()[0]) &&
```
artemrakhov: Hm, we do check this later in the verifier. Maybe don't check here?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1759,6 +1759,56 @@ void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {
...
+  // the output slice.
+
+  // Dimensions and offsets for the output and values slices. sliceDims
+  // will always be {1, [rest of output dimensions]} since the first dimension
+  // is the index in this operation. sliceOffsets will be {indices[j], 0, ...}
+  // for the output slice and {j, 0, ...} for the values slice so that the
+  // slice at index j gets mapped to index indices[j] in the dense
+  // representation.
+  ShapeVector sliceDims(numOutDims);
+  ShapeVector sliceOffsets(numOutDims);
```
artemrakhov:
```suggestion
  ShapeVector sliceOffsets(numOutDims, 0);
```

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1759,6 +1759,56 @@ void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {
...
+  // slices of output and values and accumulating the value slice into
+  // the output slice.
+
+  // Dimensions and offsets for the output and values slices. sliceDims
+  // will always be {1, [rest of output dimensions]} since the first dimension
+  // is the index in this operation. sliceOffsets will be {indices[j], 0, ...}
+  // for the output slice and {j, 0, ...} for the values slice so that the
+  // slice at index j gets mapped to index indices[j] in the dense
+  // representation.
+  ShapeVector sliceDims(numOutDims);
```
artemrakhov:
```suggestion
  ShapeVector sliceDims(out->dims().begin(), out->dims().end());
```

bertmaher:
@artemrakhov Can you please request changes if these comments are blocking?  Since I already accepted the diff, it kind of leaves this in a weird ambiguous state have comments pending; better to be clear that action is required.  Thanks!

tools/ClassGen/NodeGen.cpp
```diff
@@ -318,6 +318,21 @@ int main(int argc, char **argv) {
                     "input vector of length N the output is a Nx2 matrix with "
                     "(offset, lengths) packaged for each segment.");
 
+  // clang-format off
```
jfix71: I don't think this is necessary here -- can you verify the formatting is OK without disabling `clang-format`?

SplitInfinity:
Updates:
 - removed `clang-format off/on` in `NodeGen`

jfix71:
@SplitInfinity Looks like you accidentally added `*.status` files here?

include/glow/Graph/Graph.h
```diff
@@ -513,6 +513,21 @@ class Function final : public Named {
...
                                              NodeValue lengths);
 
+  /// Implements an operation that converts the sparse representation given by
+  /// the pair of \p indices and \p values into a dense representation.
+  /// This representation contains each vector of \p values at the corresponding
+  /// index given by \p indices. All indices that are not present in \p indices
+  /// are filled with zeroes. \p indices can contain duplicates, and in this
+  /// case, the corresponding vectors in \p values are added.
+  ///
+  /// The first dimension of the output is the maximum index seen in
```
artemrakhov: Please change this comment the same way you changed doc for SparseToDenseNode. Let's not say "vector" here. And not mention maximum index, since we don't do it.

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1255,3 +1255,89 @@ TEST(caffe2, EQ1D) {
   // Graph has two inputs and one output.
   EXPECT_EQ(mod.getPlaceholders().size(), 3);
 }
+
+// Test loading a SparseToDense operator.
+TEST(caffe2, sparseToDense) {
```
artemrakhov: If I was implementing this, I would not add a test to `caffe2ImporterTest` at all. Of course, it's better to have more tests. But we are mixing tests here a bit.

This test runs the net and checks results. While the other new unit test in `OperatorTest` does the same, but with different input. Ideally all "run and check result" tests should be in the same place. And Importer tests should only check Importer's logic, which is complicated in some cases, but not always. For `SparseToDense`, the logic in Importer is trivial, so I'd not bother adding a file specifically for this.

What do you think? Maybe just check node's type in this test?

rdzhabarov:
We'd also need an ONNX version of this operator.

tests/unittests/caffe2ImporterTest.cpp
SplitInfinity: Yes, I agree. I will check only the structure of the graph here and move the "run and check result" test to the operator tests.

SplitInfinity:
@rdzhabarov 

Is it alright if I do that in a separate PR?

SplitInfinity:
Updates:
- removed "run and check result" test out of Caffe2 importer tests and into operator tests
- fixed comment in `Graph.h`

rdzhabarov:
> Is it alright if I do that in a separate PR?

Yup!

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1759,6 +1759,51 @@ void InterpreterFunction::fwdLengthsToRangesInst(const LengthsToRangesInst *I) {
   }
 }
 
+void InterpreterFunction::fwdSparseToDenseInst(const SparseToDenseInst *I) {
```
rdzhabarov: we need CPU support as well (separate PR would be great)

SplitInfinity:
I'm going to go ahead and merge this because the [details page of Travis CI shows that it is actually done](https://travis-ci.org/pytorch/glow/builds/444816004). [The commit that the build tested](https://github.com/pytorch/glow/commit/4633b339328f7deb114960b20f74e0c0098decbf) is f465a98, which is the same as the last commit I force pushed to this branch.

PR #1899, opened 2018-10-19
[Graph] Implement Logit operator

### LINKED ISSUE #1873: [New operator] Logit
We need to support `Logit` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#logit

It's Caffe2 implementation can be found here:
https://github.com/pytorch/pytorch/blob/master/caffe2/operators/logit_op.cc#L14
We'd like to follow it closely.

More detailed: need to add `createLogit()` method to our Function class, which does "immediate lowering". I.e. creates all necessary min, max, subtraction (1 - x), division, log nodes, and returns resulting sequence (syntactic tree). No changes to IR or Backends is required.

MattPD:
*Description*: This implements the Logit operator (Graph node and Caffe2 model loader support).

Domain asserts on the `eps` value correspond to the following ones in `caffe2::LogitFunctor`: `CAFFE_ENFORCE_GT(eps_, 0.0);`, `CAFFE_ENFORCE_LT(eps_, 0.5);`.
Cf. https://github.com/pytorch/pytorch/blob/master/caffe2/operators/logit_op.h

*Testing*:
- differential testing (oracle: comparison against C++ re-implementation using randomized inputs),
- property-based randomized testing (zero-sum property for the log-odds for complementary events probabilities; right-inverse property of the logit function with respect to the logistic function),
- Caffe2 model importer test (Output shape, high level checks on the content of the graph).

*Documentation*: Comments in the interface and the implementation (including the implementation and testing strategy)

[Fixes #1873]

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -298,6 +298,21 @@ public:
   LogNode *createLog(llvm::StringRef name, NodeValue input,
                      TypeRef outTy = nullptr);
 
+  /// Create a series of nodes with \p name that implements an element-wise
+  /// logit transform. For each element of the \p input x, this is
+  /// defined as:
+  ///
+  /// y = log(x / (1 - x))
+  ///
+  /// where the \p input is clamped in (\p eps, 1 - \p eps), and
+  /// the transform parameter \p eps is a positive value (< 0.5)
+  /// (needed to avoid degenerate probabilities of 0 or 1,
+  /// which would result in taking the logarithm of zero).
+  /// The transform itself is implemented using element-wise Clip, Sub,
+  /// Splat, Div, and Log nodes.
+  /// \returns the final node.
+  Node *createLogit(llvm::StringRef name, NodeValue input, float eps);
+
   SoftMaxNode *createSoftMax(llvm::StringRef name, NodeValue input,
                              NodeValue selected, TypeRef outTy = nullptr);
 
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1066,6 +1066,32 @@ LogNode *Function::createLog(llvm::StringRef name, NodeValue input,
   return addNode(new LogNode(name, outTy ? outTy : input.getType(), input));
 }
 
+Node *Function::createLogit(llvm::StringRef name, NodeValue input, float eps) {
+  assert(eps > 0.0f && "Clamping parameter eps must be strictly positive.");
+  assert(eps < 0.5f && "Clamping parameter eps must be less than 0.5.");
+
+  // Compute clamped x using clip(x, eps, 1 - eps).
+  auto epsComplement = 1.0f - eps;
+  auto *MaxN = createClip(name.str() + ".clip", input, eps, epsComplement);
+
+  // Compute the logit transform of clamped x,
+  // log(numerator / denominator),
+  // where numerator = clamped x = MaxN,
+  // and denominator = 1 - clamped x = 1 - MaxN.
+
+  // Compute denominator = 1 - clamped x.
+  auto *onesSplat =
+      createSplat(name.str() + ".onesSplat", input.getType(), 1.0f);
+
+  auto *SN = createSub(name.str() + ".sub", onesSplat, MaxN);
+
+  // Compute the quotient = numerator / denominator.
+  auto *DN = createDiv(name.str() + ".div", MaxN, SN);
+
+  // Compute and return the logit transform (the final node).
+  return createLog(name.str() + ".log", DN);
+}
+
 SelectNode *Function::createSelect(llvm::StringRef name, TypeRef outTy,
                                    NodeValue Cond, NodeValue LHS,
                                    NodeValue RHS) {
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -412,6 +412,18 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Logit") {
+    // Load the input and (optional) epsilon clamping value:
+    auto input = getNodeValueOrCreateConstantByName(op.input(0));
+    auto epsIt = dict.find("eps");
+    // default: 1e-6 (as in Caffe2)
+    auto eps = epsIt != dict.end() ? loadFloat(epsIt->second) : 1E-6f;
+    auto *node = G_.createLogit(opName, input, eps);
+    // Save the outputs:
+    addNodeAsOutput(op, node);
+    return;
+  }
+
   if (typeName == "EQ") {
     auto in0 = getNodeValueOrCreateConstantByName(op.input(0));
     auto in1 = getNodeValueOrCreateConstantByName(op.input(1));
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -135,6 +135,64 @@ TEST_P(InterpAndCPU, log) {
   }
 }
 
+TEST_P(Operator, logit) {
+  auto eps = 1E-6f;                // the default in Caffe2
+  constexpr std::size_t size = 10; // sample size for randomized tests
+
+  // differential test:
+  // ensure we match an oracle `logit_test` (a C++ reimplementation test)
+  auto clamp_test = [](float v, float lo, float hi) {
+    return std::max(std::min(v, hi), lo);
+  };
+  auto logit_test = [clamp_test](float x, float eps = 1E-6f) {
+    float p = clamp_test(x, eps, 1.0f - eps);
+    return std::log(p / (1.0f - p));
+  };
+
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {size}, "input", false);
+  // generate the input data in (0.0f, 1.0f) (probabilites including degenerate
+  // cases) and test that afterward the input data is clamped in
+  // (eps, 1 - eps) as in Caffe2.
+  ctx_.allocate(input)->getHandle().randomize(0.0f, 1.0f, mod_.getPRNG());
+
+  auto *logitDiff = F_->createLogit("logitDiff", input, eps);
+  auto *saveDiff = F_->createSave("saveDiff", logitDiff);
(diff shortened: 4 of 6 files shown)


include/glow/Graph/Graph.h
```diff
@@ -298,6 +298,19 @@ class Function final : public Named {
...
                      TypeRef outTy = nullptr);
 
+  /// Create a series of nodes with \p name that implements an element-wise
+  /// logit transform. For each element of the \p input x, this is
+  /// defined as:
+  ///
+  /// y = log(x / (1 - x))
+  ///
+  /// where the \p input is clamped in (\p eps, 1 - \p eps), and
+  /// the transform parameter \p eps is a small positive epsilon value.
```
SplitInfinity: `small positive epsilon value` --> `small positive value`

SplitInfinity: Suggestion: `small positive epsilon value` --> `small positive value`

Maybe it's a good idea to explain that `eps` is needed to avoid taking the logarithm of too small a value.

lib/Graph/Graph.cpp
```diff
@@ -1066,6 +1066,35 @@ LogNode *Function::createLog(llvm::StringRef name, NodeValue input,
   return addNode(new LogNode(name, outTy ? outTy : input.getType(), input));
 }
 
+Node *Function::createLogit(llvm::StringRef name, NodeValue data, float eps) {
```
SplitInfinity: `NodeValue data` --> `NodeValue input`

so that the parameter name here matches the parameter name in the header.

lib/Graph/Graph.cpp
```diff
@@ -1066,6 +1066,35 @@ LogNode *Function::createLog(llvm::StringRef name, NodeValue input,
...
+
+  // Compute clamped x = max(min(x, 1 - eps), eps).
+  // First, compute 1 - eps.
+  auto *ones =
+      createSplat(name.str() + ".ones", SE->getResult().getType(), 1.0f);
+  auto *minArg = createSub(name.str() + ".sub", ones, SE);
+  // Take the min of data and 1 - eps.
+  auto *MinN = createMin(name.str() + ".min", data, minArg);
+  // Take the max of the min and eps (thus obtaining the clamped x).
+  auto *MaxN = createMax(name.str() + ".max", MinN, SE);
```
SplitInfinity: There is an existing function `Function::createClip` that will clamp the input to a provided minimum and maximum. Consider using that instead, like so:

`auto CN = createClip(name.str() + ".clip", input, eps, 1-eps);`

`CN` will then replace `MaxN` as the input to the `SubNode` below.

tests/unittests/OperatorTest.cpp
```diff
@@ -135,6 +135,76 @@ TEST_P(InterpAndCPU, log) {
...
+  auto *savePlaceholderX = saveX->getPlaceholder();
+  ctx_.allocate(savePlaceholderX);
+
+  // differential test:
+  // ensure we match an oracle `logit_test` (a C++ reimplementation test)
+  auto clamp_test = [](float v, float lo, float hi) {
+    return std::max(std::min(v, hi), lo);
+  };
+  auto logit_test = [clamp_test](float x, float eps = 1E-6f) {
+    float p = clamp_test(x, eps, 1.0f - eps);
```
SplitInfinity: Consider changing this to `float p = std::max(std::min(x, 1.0f-eps), eps);`

`clamp_test` is only used here, so I don't see a reason to separate it out into its own lambda.

tests/unittests/OperatorTest.cpp
```diff
@@ -135,6 +135,76 @@ TEST_P(InterpAndCPU, log) {
...
+  ctx_.allocate(saveCompl->getPlaceholder());
+
+  // property: the logit function is the right-inverse of the logistic function
+  // i.e., logistic(logit(p)) == p
+  auto logistic_test = [](float x) { return 1.0f / (1.0f + std::exp(-x)); };
+
+  EE_.compile(CompilationMode::Infer, F_, ctx_);
+  EE_.run();
+
+  // results: known values
```
SplitInfinity: Consider removing this set of tests altogether. These tests are basically same as the tests against the oracle.

tests/unittests/OperatorTest.cpp
```diff
@@ -135,6 +135,76 @@ TEST_P(InterpAndCPU, log) {
...
+  };
+  auto logit_test = [clamp_test](float x, float eps = 1E-6f) {
+    float p = clamp_test(x, eps, 1.0f - eps);
+    return std::log(p / (1.0f - p));
+  };
+
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {size}, "input", false);
+  // ensure the input data is clamped in (eps, 1 - eps) (as in Caffe2)
+  ctx_.allocate(input)->getHandle().randomize(eps, 1.0f - eps, mod_.getPRNG());
```
SplitInfinity: I think it might be a good idea to set one of the values in this input to a value `< eps` and another to a value `> 1 - eps` to test that the implementation of the node clamps the input correctly.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -412,6 +412,17 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
+  if (typeName == "Logit") {
```
SplitInfinity: Consider adding a unit test to `caffe2ImporterTest.cpp` to make sure that this code parses the operator out of a Caffe2 protobuf correctly.

tests/unittests/OperatorTest.cpp
```diff
@@ -135,6 +135,76 @@ TEST_P(InterpAndCPU, log) {
...
+  // i.e., logit(p) + logit(1 - p) == 0
+  Node *complInput = input;
+  Node *const1 = F_->createSplat("const1", input->getType(), 1.0);
+  complInput = F_->createSub("sub", const1, complInput);
+  Node *logitCompl = F_->createLogit("logitCompl", complInput, eps);
+  auto *saveCompl = F_->createSave("saveCompl", logitCompl);
+  ctx_.allocate(saveCompl->getPlaceholder());
+
+  // property: the logit function is the right-inverse of the logistic function
+  // i.e., logistic(logit(p)) == p
```
SplitInfinity: Very cool!

tests/unittests/OperatorTest.cpp
```diff
@@ -135,6 +135,76 @@ TEST_P(InterpAndCPU, log) {
...
+      mod_.createPlaceholder(ElemKind::FloatTy, {size}, "input", false);
+  // ensure the input data is clamped in (eps, 1 - eps) (as in Caffe2)
+  ctx_.allocate(input)->getHandle().randomize(eps, 1.0f - eps, mod_.getPRNG());
+
+  auto *logitDiff = F_->createLogit("logitDiff", input, eps);
+  auto *saveDiff = F_->createSave("saveDiff", logitDiff);
+  ctx_.allocate(saveDiff->getPlaceholder());
+
+  // property: zero-sum for the log-odds for complementary events probabilities
+  // i.e., logit(p) + logit(1 - p) == 0
```
SplitInfinity: Very cool!

include/glow/Graph/Graph.h
MattPD: Thanks! Added a more refined explanation!

lib/Graph/Graph.cpp
MattPD: Thanks, done!

lib/Graph/Graph.cpp
MattPD: Nice! Clip operator would be much cleaner!

One aspect I'm wondering about is numerical reproducibility compared to Caffe2 (which also has Clip, but does not use it in its Logit implementation).
The difference is that Clip's implementation (in Caffe2 and in Glow) is a min-of-max (as in `std::min(std::max(v, lo), hi)`), whereas Logit's internal clamping implementation (in Caffe2) uses a max-of-min instead (as in `std::max(std::min(v, hi), lo)`).
In general, these are not commutative (https://2pi.dk/2016/05/ieee-min-max) although perhaps the inputs (like https://wandbox.org/permlink/iWTBzwmn4Y7y8Aka) are not reasonable enough to warrant special treatment (which a dedicated re-implementation may amount to). I'm curious, are there any reasonable values that would warrant it? If not I think Clip seems indeed preferable.

I'm wondering what's the best course of action here. Thoughts?

tests/unittests/OperatorTest.cpp
MattPD: Done!

tests/unittests/OperatorTest.cpp
MattPD: Thanks! :-)

lib/Graph/Graph.cpp
SplitInfinity: Oh, I didn't realize they weren't commutative. Thanks for the link to that blog post!

It seems to me that the two clamping implementations are commutative only if `lo <= hi` is true. I would argue that is not  a problem because if `lo > hi`, the caller is using the function incorrectly.

Since the domain of the logit function is [0, 1], I l think it makes sense to restrict values for `eps` to [0, 0.5] because this is the largest subset of [0, 1] for which `eps <= 1-eps` is true. Maybe it's worth adding an `assert` to make sure `eps` is in this range.

bertmaher: Hah, is it a coincidence that that blog post is written by @stoklund? :-D

bertmaher: Wow, that Alive-NJ stuff is really cool.  I've got to play with that sometime!

rdzhabarov:
> Also, we have a policy of one commit per pull request, so please squash all of your commits into one.
This is not always the case. Sometimes we'd like to preserve multiple commits in merged PR.

But anyway, I've added an option to "allow squash merge" in addition to "rebase merge".

lib/Importer/Caffe2ModelLoader.cpp
MattPD: Done! (Output shape, high level checks on the content of the graph).

MattPD:
Rebase & resolve: Done!
Thanks for all the feedback, by the way!

nadavrot:
Thank you Matt!  👍

PR #1900, opened 2018-10-20
[Import] Support Sum with 1 Input

yinghai:
*Description*:
We have seen cases where we have sum with 1 input. It's pretty trivial to support that. 

*Testing*:
unittest
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/tests/unittests/onnxImporterTest.cpp
+++ b/tests/unittests/onnxImporterTest.cpp
@@ -235,3 +235,42 @@ TEST(onnx, importDotProduct) {
   auto *batchedReduceAdd = llvm::cast<BatchedReduceAddNode>(saveInput);
   ASSERT_TRUE(llvm::isa<MulNode>(batchedReduceAdd->getBatch()));
 }
+
+/// Test loading Sum with one input and one output
+TEST(onnx, importSum1) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  std::string netFilename("tests/models/onnxModels/sum1.onnxtxt");
+
+  Context ctx;
+  Placeholder *output;
+  {
+    Tensor x(ElemKind::FloatTy, {3});
+    x.getHandle() = {1, 2, 3};
+    ONNXModelLoader onnxLD(netFilename, {"x"}, {&x.getType()}, *F);
+    output = onnxLD.getSingleOutput();
+
+    ctx.allocate(mod.getPlaceholders());
+    updateInputPlaceholdersByName(ctx, &mod, {"x"}, {&x});
+  }
+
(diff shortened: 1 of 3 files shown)


tests/unittests/onnxImporterTest.cpp
```diff
@@ -235,3 +235,36 @@ TEST(onnx, importDotProduct) {
...
+  Context ctx;
+  Placeholder *output;
+  {
+    Tensor x(ElemKind::FloatTy, {3});
+    x.getHandle() = {1, 2, 3};
+    ONNXModelLoader onnxLD(netFilename, {"x"}, {&x.getType()}, *F);
+    output = onnxLD.getSingleOutput();
+
+    ctx.allocate(mod.getPlaceholders());
+    updateInputPlaceholdersByName(ctx, &mod, {"x"}, {&x});
```
yinghai: @bertmaher I copy-pasted the unit test but what do above 2 lines do?

bertmaher: They let us provide the actual input values for the net.  When we load the net the inputs are represented by "placeholders" that aren't backed by any storage.  `ctx.allocate` creates storage for them, and `updateInputPlaceholdersByName` fills them with the named tensor values.

tests/unittests/onnxImporterTest.cpp
```diff
@@ -235,3 +235,36 @@ TEST(onnx, importDotProduct) {
...
+  }
+
```
bertmaher: It'd be nice to have some checks of the graph structure here since it's an interesting net.  Let me suggest something like
```
// Verify structure: input -> Save -> output
ASSERT_EQ(mod.getPlaceholders().size(), 2);
ASSERT_EQ(F->getNodes().size(), 1);
auto *save = getSaveNodeFromDest(output);
ASSERT_TRUE(llvm::isa<Placeholder>(save->getInput().getNode()));
```

tests/unittests/onnxImporterTest.cpp
```diff
@@ -235,3 +235,36 @@ TEST(onnx, importDotProduct) {
...
+  auto *res = ctx.get(output);
+  EE.compile(CompilationMode::Infer, F, ctx);
+  EE.run();
+
+  auto result = res->getHandle();
+  std::vector<size_t> expectedDims = {3};
+  std::vector<float> expectedValues = {1, 2, 3};
+
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
```
bddppq: Does glow wanna use gmock? :-) (which has matcher for comparing float arrays)

bertmaher: Heh :-).  Gmock can be useful, but I'd rather not add another external dependency unless it's really valuable.

PR #1901, opened 2018-10-20
[onnx importer] Import LengthsToRanges

bertmaher:
*Description*: that's it
*Testing*: new onnxImporterTest
*Documentation*: N/A
Another piece of #1885

### DIFF
--- a/lib/Importer/ONNXIFIModelLoader.cpp
+++ b/lib/Importer/ONNXIFIModelLoader.cpp
@@ -209,6 +209,8 @@ ONNXIFIModelLoader::parseOperators(const void *onnxModel,
     } else if (operation == "DotProduct") {
       ADD_OP_MAPPING(BatchedReduceAddNodeKind, FloatTy);
       ADD_OP_MAPPING(MulNodeKind, FloatTy);
+    } else if (operation == "LengthsToRanges") {
+      ADD_OP_MAPPING(LengthsToRangesNodeKind, Int64ITy);
     }
   }
 #undef ADD_OP_MAPPING
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -512,6 +512,13 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     return true;
   }
 
+  if (typeName == "LengthsToRanges") {
+    auto in = getNodeValueOrCreateConstantByName(op.input(0));
+    auto *node = G_.createLengthsToRanges(opName, in);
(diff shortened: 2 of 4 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -512,6 +512,13 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     return true;
   }
 
+  if (typeName == "LengthsToRanges") {
+    auto in = getNodeValueOrCreateConstantByName(op.input(0));
+    auto *node = G_.createLengthsToRanges(opName, in);
+    addNodeAsOutput(op, node);
+    return true;
+  }
+
```
beicy: This "LengthsToRanges" op can be added in CommonOperatorLoader.h. 
Btw, I can't find this op in ONNX op list : https://github.com/onnx/onnx/blob/master/docs/Operators.md.
Is it some customized op in ONNX?

yinghai: Yes.

https://github.com/pytorch/pytorch/blob/d4c58216d79f1e1b263e8eeb24cc7a466f3cbc55/caffe2/onnx/torch_ops/defs.cc#L8-L108

beicy: @yinghai I tried to search "LengthsToRanges" in the webpage but failed. Is there anything I missed? Thanks!

yinghai: We will add that.

lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -209,6 +209,8 @@ ONNXIFIModelLoader::parseOperators(const void *onnxModel,
     } else if (operation == "DotProduct") {
       ADD_OP_MAPPING(BatchedReduceAddNodeKind, FloatTy);
       ADD_OP_MAPPING(MulNodeKind, FloatTy);
+    } else if (operation == "LengthsToRanges") {
+      ADD_OP_MAPPING(LengthsToRangesNodeKind, Int64ITy);
```
rdzhabarov: we need to add this to `isOpSupported` of CPU backend and Interpreter as well.

bertmaher: Question here: it's already supported by default since it's not using Int8QTy.  Do you mean I should add quantized support?  (Which is basically a NOP, so maybe it already just works?)

rdzhabarov: I did not realize that it'll be supported by default. Ignore this comment.

PR #1902, opened 2018-10-20
[cpu] Implement LengthsToRanges

### LINKED ISSUE #1885: [New operator] LengthsToRanges

bertmaher:
*Description*: That's it
*Testing*: Enable CPU backend for operatorTest
*Documentation*: n/a
Fixes #1885

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -883,6 +883,17 @@ void libjit_scatterassign_i8(int8_t *data, const size_t *indices,
   libjit_scatterassign(data, indices, slices, numIndices, sliceSize);
 }
 
+void libjit_lengths_to_ranges_u(size_t *ranges, const size_t *lengths,
+                                size_t size) {
+  size_t offset = 0;
+  for (size_t i = 0; i < size; i++) {
+    auto length = lengths[i];
+    ranges[i * 2] = offset;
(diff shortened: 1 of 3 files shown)


lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -883,6 +883,17 @@ void libjit_scatterassign_i8(int8_t *data, const size_t *indices,
   libjit_scatterassign(data, indices, slices, numIndices, sliceSize);
 }
 
+void libjit_lengths_to_ranges_u(size_t *ranges, const size_t *lengths,
```
yinghai: What's the naming magic here? `libjit_function_type`?

beicy: "lenghths_to_ranges" is the name defined in LLVMIRGen.cpp(line 1676) , and then the function name in kernel is "libjit_" + name defined in LLVMIRGen.cpp(here is `lenghths_to_range`s)" + "_" + type of the output of this instr (here is `u`).

PR #1903, opened 2018-10-22
Allow -convert-inout-to-fp16 to be used in combination with -load_profile

opti-mix:
This PR builds on top #1905 and allows for taking any model, converting it into fp16 and then quantizing it. It is useful for running models on the backends without fp32 support.

You can now do the following:
```
bin/image-classifier tests/images/imagenet/cat_285.png -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data -convert-to-fp16  -convert-inout-to-fp16 -load_profile resnet50_symmetric_q8.yml
```

The following changes were required:
- updating a bunch of checks/assertions that assumed that only FP32 can be used
- extending the interpreter to support quantize/dequantize with fp16 operands

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1902,12 +1902,9 @@ void InterpreterFunction::fwdQuantizationProfileInst(
   quantization::generateTensorHistogram(inputTensor, currentHistogram, min,
                                         max);
 }
-/// Quantize floating point tensor. Scale and Offset are based on return type
-/// of the instruction \p I.
-void InterpreterFunction::fwdQuantizeInst(const glow::QuantizeInst *I) {
-  auto srcHandle = getWeightHandle(I->getSrc());
-  auto *destTensor = getTensor(I->getDest());
 
+template <typename ElemTy>
+static void fwdQuantize(Handle<ElemTy> srcHandle, Tensor *destTensor) {
   TensorQuantizationParams params{destTensor->getType().getScale(),
                                   destTensor->getType().getOffset()};
 
@@ -1916,18 +1913,55 @@ void InterpreterFunction::fwdQuantizeInst(const glow::QuantizeInst *I) {
     destHandle.raw(i) = quantization::quantize(srcHandle.raw(i), params);
   }
 }
-/// Dequantize integer tensor. Scale and Offset are based
-/// on the source tensor type.
-void InterpreterFunction::fwdDequantizeInst(const glow::DequantizeInst *I) {
-  auto *srcTensor = getTensor(I->getSrc());
-  auto destHandle = getWeightHandle(I->getDest());
 
+/// Quantize floating point tensor. Scale and Offset are based on return type
+/// of the instruction \p I.
+void InterpreterFunction::fwdQuantizeInst(const glow::QuantizeInst *I) {
+  auto *destTensor = getTensor(I->getDest());
+  switch (I->getSrc()->getElementType()) {
+  case ElemKind::FloatTy: {
+    auto srcHandle = getWeightHandle<float>(I->getSrc());
+    fwdQuantize(srcHandle, destTensor);
+    return;
+  }
+  case ElemKind::Float16Ty: {
+    auto srcHandle = getWeightHandle<float16>(I->getSrc());
+    fwdQuantize(srcHandle, destTensor);
+    return;
+  }
+  default:
+    llvm_unreachable("Type not supported");
+  }
+}
+
+template <typename ElemTy>
+static void fwdDequantize(Tensor *srcTensor, Handle<ElemTy> destHandle) {
   TensorQuantizationParams params{srcTensor->getType().getScale(),
                                   srcTensor->getType().getOffset()};
 
   auto srcHandle = srcTensor->getHandle<int8_t>();
   for (size_t i = 0, e = destHandle.size(); i < e; ++i) {
     destHandle.raw(i) = quantization::dequantize(srcHandle.raw(i), params);
+  }
+}
+
+/// Dequantize integer tensor. Scale and Offset are based
+/// on the source tensor type.
+void InterpreterFunction::fwdDequantizeInst(const glow::DequantizeInst *I) {
+  auto *srcTensor = getTensor(I->getSrc());
+  switch (I->getDest()->getElementType()) {
+  case ElemKind::FloatTy: {
+    auto destHandle = getWeightHandle<float>(I->getDest());
+    fwdDequantize(srcTensor, destHandle);
+    return;
+  }
+  case ElemKind::Float16Ty: {
+    auto destHandle = getWeightHandle<float16>(I->getDest());
+    fwdDequantize(srcTensor, destHandle);
+    return;
+  }
+  default:
+    llvm_unreachable("Type not supported");
   }
 }
 
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1414,8 +1414,7 @@ ScatterAssignNode *Function::createScatterAssign(llvm::StringRef name,
 
 QuantizeNode *Function::createQuantize(llvm::StringRef name, NodeValue input,
                                        TypeRef outTy) {
-  assert(input.getElementType() == ElemKind::FloatTy &&
-         "Input must be a floating type");
+  assert(input.getType()->isFPType() && "Input must be a floating type");
   assert(outTy->getElementType() == ElemKind::Int8QTy &&
          "Output must be a quantized type");
   assert(input.dims().equals(outTy->dims()) &&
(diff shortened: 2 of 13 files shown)


opti-mix:
@qcolombet Please have a look. I tried to implement it based on our off-line discussion. But may be a got some things wrong.

tools/loader/ImageClassifier.cpp
```diff
@@ -129,6 +130,11 @@ llvm::cl::opt<std::string> modelInputName(
     llvm::cl::desc("The name of the variable for the model's input image."),
     llvm::cl::value_desc("string_name"), llvm::cl::Required,
     llvm::cl::cat(imageLoaderCat));
+
+llvm::cl::opt<bool> runUsingFP16(
+    "run-using-fp16",
+    llvm::cl::desc("Run all floating-point computation using fp16."),
```
nadavrot: Maybe a better description would be "convert the inputs of the network to fp16"

tools/loader/ImageClassifier.cpp
```diff
@@ -238,6 +245,51 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
   }
 }
 
+/// Convert all placeholders with elements of type \p srcTy into type \p destTy.
+/// Also explicitly convert the input data and the output tensor into the type
+/// \p destTy.
+static void convertPlaceholders(Loader &loader, Context &ctx, Tensor &data,
+                                Placeholder *resultVar, ElemKind srcTy,
+                                ElemKind destTy) {
+  // Convert inputs and outputs into FP16.
```
nadavrot: The name of the method is generic, but it does something specific to fp16.

nadavrot: At least the comments say fp16.

opti-mix: Oops. Good catch. The method itself is not FP16 specific, but I forgot to update the comment. Will fix.

tools/ClassGen/InstrGen.cpp
```diff
@@ -455,14 +455,14 @@ int main(int argc, char **argv) {
...
-      .autoVerify(VerifyKind::SameElementType, {"Src", "ElemKind::FloatTy"})
+      .autoVerify(VerifyKind::TypePredicate, {"Src", "isFPType()"})
       .autoVerify(VerifyKind::SameShape, {"Dest", "Src"})
       .autoIRGen();
 
   BB.newInstr("Dequantize")
       .addOperand("Dest", OperandKind::Out)
       .addOperand("Src", OperandKind::In)
-      .autoVerify(VerifyKind::SameElementType, {"Dest", "ElemKind::FloatTy"})
+      .autoVerify(VerifyKind::TypePredicate, {"Dest", "isFPType()"})
```
nadavrot: I think that the name predicate is overloaded. Can we use a different name? Maybe type check?

opti-mix: Sure. Let's do it.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1648,10 +1648,20 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
...
-      tensorToBeModified =
-          quantization::quantizeTensor(tensorToBeModified, params);
+      switch (tensor.getElementType()) {
+      case ElemKind::FloatTy:
+        tensorToBeModified =
+            quantization::quantizeTensor<float>(tensorToBeModified, params);
+        break;
+      case ElemKind::Float16Ty:
+        tensorToBeModified =
+            quantization::quantizeTensor<float16>(tensorToBeModified, params);
```
qcolombet: Could we do the type dispatch within the quantizeTensor method?

tools/ClassGen/InstrGen.cpp
qcolombet: +1

tools/loader/ImageClassifier.cpp
```diff
@@ -238,6 +245,51 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
...
+    }
+  }
+  data.convertToType(destTy);
+  for (auto &user : resultVar->getUsers()) {
+    if (auto *SN = llvm::dyn_cast<SaveNode>(user.getUser())) {
+      SN->setNthInput(0, loader.getFunction()->createConvertTo(
+                             "cast", SN->getInput(), resultVar->getType()));
+    }
+  }
+}
```
qcolombet: This logic is essentially already in the function converter.
I'll post a PR with the required refactoring. Either wait for that PR and you won't have to cover that part, or I'll update if it lands before.

tools/loader/ImageClassifier.cpp
```diff
@@ -301,15 +359,10 @@ int main(int argc, char **argv) {
...
-      Tensor slice = H.extractSlice(i);
-      auto SH = slice.getHandle<>();
-      llvm::outs() << " File: " << inputImageFilenames[i];
-
-      auto topKPairs = getTopKPairs(SH);
-      printTopKPairs(topKPairs);
+    if (runUsingFP16) {
+      printResults<float16>(loader, SMVarT);
+    } else {
+      printResults<float>(loader, SMVarT);
```
qcolombet: Alternatively you could convert back the tensor to fp32 so you don't have to touch this part.

lib/Quantization/Base/Base.cpp
```diff
@@ -30,12 +30,27 @@ int8_t quantize(float input, const TensorQuantizationParams &TQP) {
...
   Tensor tmp(ElemKind::Int8QTy, tensor.dims(), TQP.scale, TQP.offset);
-  assert(tensor.getElementType() == ElemKind::FloatTy &&
-         "Type not supported yet");
-  auto srcHandle = tensor.getHandle();
   auto destHandle = tmp.getHandle<int8_t>();
-  for (size_t i = 0, e = destHandle.size(); i < e; ++i) {
-    destHandle.raw(i) = quantization::quantize(srcHandle.raw(i), TQP);
+  assert(tensor.getType().isFPType() && "Type not supported yet");
+  switch (tensor.getElementType()) {
+  case ElemKind::FloatTy: {
```
jfix71: nit: These cases could be made into a macro, since the only difference is the type of the Handle. Not a big deal though since there are only two cases and unlikely to grow much if at all.

tools/loader/ImageClassifier.cpp
qcolombet: That's in #1905

tools/ClassGen/InstrBuilder.cpp
```diff
@@ -239,6 +239,13 @@ void InstrBuilder::emitClass(std::ostream &os) const {
         }
         break;
       }
+      case VerifyKind::TypeCheck: {
```
nadavrot: We could use a comment here that explains what code is being generated. Same for the other cases in this switch.

qcolombet:
Ditto on the quantize/dequantize functionality from the quantization library?

opti-mix:
@qcolombet 

> Could you add tests for the added support for quantize/dequantize nodes in the interpreter as well?

Yes, I've written a new test in `OperatorTest.cpp`

> Ditto on the quantize/dequantize functionality from the quantization library?

What do you mean by this one? I'm not sure I understand.

qcolombet:
@opti-mix 

>> Ditto on the quantize/dequantize functionality from the quantization library?
> What do you mean by this one? I'm not sure I understand.

I mean, I want a test to go with this commit "Support quantization involving FP16". This only modifies the quantization library and could/should be tested separately from the interpreter.

opti-mix:
@qcolombet I added all the tests you requested.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1916,12 +1913,29 @@ void InterpreterFunction::fwdQuantizeInst(const glow::QuantizeInst *I) {
...
+    fwdQuantize(srcHandle, destTensor);
+    return;
+  }
+  default:
+    llvm_unreachable("Type not supported");
+  }
+}
+
+template <typename ElemTy>
+void static fwdDequantize(Tensor *srcTensor, Handle<ElemTy> destHandle) {
```
qcolombet: The static keyword is weirdly placed :).

lib/Graph/Graph.cpp
```diff
@@ -1434,6 +1433,14 @@ DequantizeNode *Function::createDequantize(llvm::StringRef name,
   return addNode(new DequantizeNode(name, outTy, input));
 }
 
+DequantizeNode *Function::createDequantize(llvm::StringRef name,
+                                           NodeValue input, TypeRef outTy) {
```
qcolombet: Could we have the other createDequantize call this one, so that we don't duplicate the assert?
Ditto for createQuantize.

PR #1905, opened 2018-10-22
[ImageClassifier] Add an option to convert the input and output to fp16

qcolombet:
[ImageClassifier] Add a convert-inout-to-fp16 option
    
    *Description*
    This teaches the image-classifier how to convert the input and output
    of a network to fp16.
    
    *Testing*
    Ran resnet50 with -convert-inout-to-fp16

### DIFF
--- a/lib/Converter/FunctionConverter.cpp
+++ b/lib/Converter/FunctionConverter.cpp
@@ -59,6 +59,60 @@ Node &FunctionConverter::morphNode(Node &node) { return node; }
 
 void FunctionConverter::postProcessing(Node &node) {}
 
+void FunctionConverter::convertOutputs(Node &node) {
+  for (unsigned idx = 0, end = node.getNumResults(); idx != end; ++idx) {
+    NodeValue val = node.getNthResult(idx);
+    TypeRef targetTy = getTargetTypeForOutput(val);
+    if (!targetTy || targetTy == val.getType()) {
+      continue;
+    }
+    // convert the node and create a conversion to keep the users happy.
+    assert(targetTy->dims() == val.getType()->dims() &&
+           "Conversion does not preserve shape");
+    TypeRef origTy = val.getType();
+    // Fake the morphing of the node so that the creation
+    // of the conversion works properly.
+    val.setType(targetTy);
+    // Create the conversion.
+    Node *conversion = createConversion(val, origTy);
+    // "conversion" uses val so after this call,
+    // we will get a use of conversion inside conversion.
+    NodeValue conversionVal = getConversionOutput(*conversion);
+    // Store the users in a temporary object because setOperand
+    // will invalidate the iterator.
+    llvm::SmallVector<NodeUse, 4> users(val.getUsers().begin(),
+                                        val.getUsers().end());
+    // We cannot use replaceAllUsesWith here because:
+    // 1. At this point, val and conversion don't have the same type
+    //    (one is converted the other is the original type), and that
+    //    would trigger an assertion.
+    // 2. We would end up replacing the use of val in "conversion" by
+    //   "conversion".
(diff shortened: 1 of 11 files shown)


qcolombet:
@opti-mix This patch should do what you want, modulo, you need you're commit with the optimization that removes the conversion between the same types.

This is WIP because I need to add the tests for the placeholder conversion. But the code itself should be done.

qcolombet:
Added test case.

lib/Converter/FunctionConverter.cpp
```diff
@@ -59,6 +62,99 @@ Node &FunctionConverter::morphNode(Node &node) { return node; }
...
+           "Conversion does not preserve shape");
+    TypeRef origTy = val.getType();
+    // Fake the morphing of the node so that the creation
+    // of the conversion works properly.
+    val.setType(targetTy);
+    // Store the users in a temporary object because setOperand
+    // will invalidate the iterator.
+    llvm::SmallVector<NodeUse, 4> users(val.getUsers().begin(),
+                                        val.getUsers().end());
+    // We cannot use replaceAllUsesWith here because:
```
beicy: I noticed this "4 users" before but forget to ask -- Just wondering how this "4" comes from :) Thanks!

lib/Converter/FunctionConverter.cpp
```diff
@@ -59,6 +62,99 @@ Node &FunctionConverter::morphNode(Node &node) { return node; }
...
+    //   "conversion".
+    // 3. Node may be a module level value and we need one conversion per
+    //    function.
+    for (auto use : users) {
+      Node *user = use.getUser();
+      Function *parent = user->getParent();
+      assert(parent && "User not in a function?!");
+
+      SaveNode *saveNode = llvm::dyn_cast<SaveNode>(user);
+      if (saveNode) {
```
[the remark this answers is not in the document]
qcolombet: Good catch, that's a relic of some specialization I removed from the code.

tools/loader/ImageClassifier.cpp
```diff
@@ -129,6 +130,12 @@ llvm::cl::opt<std::string> modelInputName(
     llvm::cl::desc("The name of the variable for the model's input image."),
     llvm::cl::value_desc("string_name"), llvm::cl::Required,
     llvm::cl::cat(imageLoaderCat));
+
+llvm::cl::opt<bool> convertInputAndOutToFp16(
```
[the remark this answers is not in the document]
qcolombet: Done.

lib/Converter/FunctionConverter.cpp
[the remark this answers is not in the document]
qcolombet: Done

tools/loader/ImageClassifier.cpp
```diff
@@ -299,6 +315,9 @@ int main(int argc, char **argv) {
 
     // Perform the inference execution.
     loader.runInference(ctx);
+    if (convertInAndOutToFp16) {
+      SMVarT->convertToType(ElemKind::FloatTy);
```
jfix71: Shouldn't this be `ElemKind::Float16Ty`?

opti-mix: @jfix71 If I understand the code correctly, then everything is fine. We convert the payload of the tensor containing final results from fp16 into fp32 and then use the usual fp32-based function afterwards to print those results. May be a comment before the `if` explaining what's going on would be useful.

qcolombet: That is correct.
I've added a comment to explain that. Let me know what you think.

jfix71: Looks good, thanks!

PR #1906, opened 2018-10-22
[cmake] Fix a bunch of dependencies so BUILD_SHARED_LIBS works

bertmaher:
*Description*: It seems we were getting away with lax dependencies, which were exposed when I tried building with `-DBUILD_SHARED_LIBS=ON`.  (Incrementally relinking tests is somewhat faster with shared libs, which is why I was trying this.)

*Testing*: `ninja all && ninja test`

*Documentation*: None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,8 +15,10 @@ set(CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+if (NOT BUILD_SHARED_LIBS)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+  set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+endif ()
 
 # Export a JSON file with the compilation commands that external tools can use
 # to analyze the source code of the project.
--- a/lib/CodeGen/CMakeLists.txt
+++ b/lib/CodeGen/CMakeLists.txt
@@ -1,1 +1,5 @@
 add_library(CodeGen MemoryAllocator.cpp)
+
+target_link_libraries(CodeGen
+                      PUBLIC
+                        Support)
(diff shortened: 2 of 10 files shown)


CMakeLists.txt
```diff
@@ -15,8 +15,10 @@ set(CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+if (NOT BUILD_SHARED_LIBS)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+  set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+endif ()
```
bertmaher: @compnerd Is there a "right" thing to do with respect to symbol visibility in shared libs that doesn't involve a ton of work annotating everything in our APIs?

lib/IR/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
+# Generated source files, defined in lib/Graph/CMakeLists.
+set(INSTR_SRC ${GLOW_BINARY_DIR}/glow/AutoGenInstr.cpp)
+set(INSTR_BLD_SRC ${GLOW_BINARY_DIR}/glow/AutoGenIRBuilder.cpp)
+
 add_library(IR
+              ${INSTR_SRC}
+              ${INSTR_BLD_SRC}
```
bertmaher: Hm, creating dependences like this doesn't seem to work the way I want it to.

CMakeLists.txt
compnerd: @bertmaher, no, annotations are the right thing to do unfortunately.

bertmaher:
Yay, this works!  Anybody want to review it?  It'll let you link tests oh so very much faster (3s vs 10s, yay developer efficiency)

lib/CodeGen/CMakeLists.txt
```diff
@@ -1 +1,5 @@
 add_library(CodeGen MemoryAllocator.cpp)
+
+target_link_libraries(CodeGen
+                      PUBLIC
```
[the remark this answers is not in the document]
bertmaher: Yup, this should be PRIVATE.  Good catch!

qcolombet:
Love it!

Issue #1904, opened 2018-10-22
Pass Context at runtime not compile time


gcatron:
In order to facilitate multi-threaded execution context passing needs to be moved from compile time to runtime.  **This will break the current API!**

At compile time, constant weight allocation will remain unchanged, however Placeholder memory will no longer be allocated at runtime. The expected sizes will be calculated from the the Placeholder types.

At runtime (ExecutionEngine::run) the context will be passed in and Placeholder memory will be assigned to point to the allocations in the context. 

When done context will no longer be a parameter of the compile function. ExecutionEngine::run and function::execute will expect a context parameter.
 
This change will be done in stages to minimize disruption. 

1. ExecutionEngine::run(Context &ctx) will be introduced. This is the new execution flow. When the bugs are worked out it will replace run. Tests will be ported to use run(ctx).

2. CompiledFunction::execute(Context &ctx) will be introduced, similar to run(ctx) this will replace execute() when the work is finished. Tests will be ported to execute(Context &ctx).

3. execute(Context &ctx) will be updated to handle memory allocation at runtime. 

4.  Once runctx and execute(ctx) are working and all tests pass run(ctx) will replace run and execute(ctx) will replace execute. At this point any external code will need to be updated to reflect these changes.

nadavrot:
@gcatron Thanks for doing this work. I have a question about the migration plan. Maybe I missed somethings. Would it make sense to start the process by adding the ctx parameter to run? I think that if you start by moving the ctx parameter to run you won't need to add the new executeCtx and compileCtx APIs.

gcatron:
We probably could get away without compileCtx for sure. My reasoning for adding executeCtx was to maintain an unbroken execution flow. I think the runtime allocation logic will need to reside in the execute method implementations as opposed to run. (would be happy to be wrong on that point)

bertmaher:
I think the end state is the right one.  But I'd second @nadavrot on avoiding adding the *Ctx versions if it's possible.  Since there are a ton of tests to port anyways, I worry that you'll find yourself spending a ton of time swizzling names :-p.  (Sometimes you can do these migrations quickly with `sed` but if not it's a lot of manual pain).

gcatron:
Perhaps an optional parameter passed to execute? Then when we're ready we just remove the optional part? This way there's only runctx added.

gcatron:
Documenting a conversation @opti-mix and I just had. A candidate API exposed by the backend,
[...]

PR #1907, opened 2018-10-22
[ExecutionEngine] Create runctx and execute(ctx) functions

gcatron:
*Description*: Create runctx and execute(ctx) functions for moving placeholder memory allocation from compile time to run time and update tests.
*Testing*: Ninja all
*Documentation*: N/A
Progress on #1904 This would bring us to completion of the second step.

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -20,14 +20,17 @@
 
 namespace glow {
 
+class Context;
+
 /// Interface for executing a compiled function.
 class CompiledFunction {
 public:
   /// Dtor.
   virtual ~CompiledFunction() = default;
 
-  /// Execute the network.
-  virtual void execute() = 0;
+  /// Execute the network and allocate Placeholder memory with given
+  /// \p ctx providing mapping between Placeholder and populated tensor.
+  virtual void execute(Context &ctx) = 0;
 };
 
 } // end namespace glow
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -81,8 +81,8 @@ public:
   void save(CompilationMode mode, Function *F, llvm::StringRef outputDir,
             llvm::StringRef networkName);
 
-  /// Runs a single execution of the function.
-  void run();
+  /// Context aware single execution of the function.
+  void run(Context &ctx);
 };
 
 //===----------------------------------------------------------------------===//
--- a/lib/Backends/Interpreter/InterpreterFunction.h
+++ b/lib/Backends/Interpreter/InterpreterFunction.h
@@ -55,7 +55,7 @@ public:
   ///@{
   ~InterpreterFunction() override;
 
-  void execute() override;
+  void execute(Context &ctx) override;
   ///@}
 
 private:
--- a/examples/char-rnn.cpp
+++ b/examples/char-rnn.cpp
@@ -276,7 +276,7 @@ int main(int argc, char **argv) {
     for (unsigned i = 0; i < generateChars; i++) {
       // Generate a char:
       updateInputPlaceholders(ctx, {X}, {&currCharInfer});
-      EE.run();
+      EE.run(ctx);
 
       // Pick a char at random from the softmax distribution.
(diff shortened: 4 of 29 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -83,6 +83,8 @@ class ExecutionEngine final {
 
   /// Runs a single execution of the function.
   void run();
+  /// Context aware single execution of the function
+  void runctx(Context &ctx);
```
rdzhabarov: why not call this simply `run`?

gcatron: no good reason.. and it won't be hard to fix let me correct that.

include/glow/Backends/CompiledFunction.h
```diff
@@ -28,6 +30,7 @@ class CompiledFunction {
 
   /// Execute the network.
   virtual void execute() = 0;
+  virtual void execute(Context &ctx) = 0;
```
rdzhabarov: can you flesh out comments on what ctx is etc in the Doxygen comment?

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,6 +16,8 @@
 #ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
 #define GLOW_BACKENDS_COMPILEDFUNCTION_H
 
+#include "glow/Graph/Context.h"
+
 #include <unordered_map>
```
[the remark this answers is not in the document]
gcatron: Can it? We're now passing context as a parameter of execute. It fails to compile when I comment it out.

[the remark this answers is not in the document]
nadavrot: You can “forward declare” the ctx.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -83,6 +83,8 @@ class ExecutionEngine final {
 
   /// Runs a single execution of the function.
   void run();
```
nadavrot: Can you remove the old run method?

lib/Backends/Interpreter/InterpreterFunction.h
```diff
@@ -55,6 +55,8 @@ class InterpreterFunction final : public CompiledFunction {
   ///@{
   ~InterpreterFunction() override;
 
+  void execute(Context &ctx) override;
+
   void execute() override;
```
nadavrot: Can you remove the old execute method?

include/glow/ExecutionEngine/ExecutionEngine.h
gcatron: I was planning to remove it once memory allocation is working in the new run/execute flow. This gives the community a chance to port this week while the allocation work is being done. If we remove it now it will break the API. If we go that route we should hold on the merge until after the email is sent out.

include/glow/Backends/CompiledFunction.h
qcolombet: Hmm, this doesn't necessarily need to hold this PR, but I am not sure I like the fact that the executer does the memory allocation. My concern is that it prevents us to reuse already allocated buffer from another run.
I.e., I would rather have the context have the location of the different placeholder at this point, instead of allocating them.
Anyhow, that's a more general design discussion, again no need to hold that PR, we would need the context to execute anyway.

include/glow/Backends/CompiledFunction.h
```diff
@@ -28,6 +30,9 @@ class CompiledFunction {
 
   /// Execute the network.
   virtual void execute() = 0;
+  /// Execute the network and allocate Placeholder memory with given
+  /// Context providing mapping between Placeholder and populated tensor.
```
qcolombet: nit: when using Doxygen, you can reference the actual parameter with \p.
E.g., [...] with the given \p ctx. [...]
That also encourages to use fully fleshed names (like context instead of ctx ;)).

nadavrot:
@gcatron Looks good to me. Please make sure to squash the commits.

rdzhabarov:
Please, make sure to resolve conflicts and good to go.

re squashing, you could also select an option to "squash and merge" on this PR, up to yous

PR #1910, opened 2018-10-23
[ImageClassifier] Add option for streaming input filenames

jfix71:
*Description*: Allow for streaming in filenames for inference, allowing for inference to be run multiple times after compiling once. This is done by passing `-` as the input image filenames command line parameter; then filenames are passed via stdin for each next batch of images to infer. The program is not compiled until the first set of filenames is provided. The first set of filenames passed is used to set the input shape for the compiled model, meaning that same batch size must be used for future inferences.

*Testing*: Manually tested using different input shapes

*Documentation*: Added/updated command line option description.

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -27,8 +27,10 @@
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <iostream>
 #include <memory>
 #include <queue>
+#include <sstream>
 
 using namespace glow;
 
@@ -78,9 +80,12 @@ namespace {
 
 /// Image loader options.
 llvm::cl::OptionCategory imageLoaderCat("Image Loader Options");
-llvm::cl::list<std::string> inputImageFilenames(llvm::cl::Positional,
-                                                llvm::cl::desc("<input files>"),
-                                                llvm::cl::OneOrMore);
+llvm::cl::list<std::string> inputImageFilenames(
+    llvm::cl::Positional,
+    llvm::cl::desc("<input files> (note: specifying '-' enables streaming "
+                   "mode, where the model is compiled once and then can be run "
+                   "many times with new input filenames passed via stdin)"),
+    llvm::cl::OneOrMore);
 llvm::cl::opt<ImageNormalizationMode> imageNormMode(
     "image_mode", llvm::cl::desc("Specify the image mode:"), llvm::cl::Required,
     llvm::cl::cat(imageLoaderCat),
@@ -141,7 +146,7 @@ llvm::cl::opt<bool> convertInAndOutToFp16(
 /// Loads and normalizes all PNGs into a tensor in the NHWC format with the
 /// requested channel ordering.
 void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
-                             Tensor *result) {
+                             Tensor *inputImageData) {
   assert(!filenames.empty() &&
          "There must be at least one filename in filenames.");
   auto range = normModeToRange(imageNormMode);
@@ -154,12 +159,12 @@ void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
   const size_t numChannels = isGray ? 1 : 3;
 
   // N x C x H x W
-  result->reset(ElemKind::FloatTy,
-                {numImages, numChannels, imgHeight, imgWidth});
-  auto RH = result->getHandle<>();
+  inputImageData->reset(ElemKind::FloatTy,
+                        {numImages, numChannels, imgHeight, imgWidth});
+  auto IIDH = inputImageData->getHandle<>();
 
-  // We iterate over all the png files, reading them all into our result tensor
-  // for processing
+  // We iterate over all the png files, reading them all into our inputImageData
+  // tensor for processing
   for (unsigned n = 0; n < filenames.size(); n++) {
     Tensor localCopy;
     // PNG images are loaded as NHWC & RGB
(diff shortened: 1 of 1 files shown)


tools/loader/ImageClassifier.cpp
```diff
@@ -291,25 +333,41 @@ int main(int argc, char **argv) {
...
-    // Print out the inferred image classification.
-    auto H = SMVarT->getHandle<>();
-    llvm::outs() << "Model: " << loader.getFunction()->getName() << "\n";
-    for (unsigned i = 0; i < inputImageFilenames.size(); i++) {
-      Tensor slice = H.extractSlice(i);
-      auto SH = slice.getHandle<>();
-      llvm::outs() << " File: " << inputImageFilenames[i];
+      // Load the next round of images to infer.
+      getNextImageFilenames(&inputImageFilenames);
+      loadImagesAndPreprocess(inputImageFilenames, &data);
```
qcolombet: Would it be possible to do that at the beginning of the loop so that we don't have to repeat it outside of the loop?

jfix71: Because we need to know what our input shape is before compilation, I think it needs to be done once prior to the loop.

The other option would be to add the input shape as a configurable command line parameter, then this could be done once, at the start of this loop. I didn't think that was a great path forward. Though we could set `{c,h,w}` to the imagenet defaults `{3, 224, 224}` and so batch would be the only one changed unless running on mnist or another dataset.

tools/loader/ImageClassifier.cpp
```diff
@@ -291,25 +334,41 @@ int main(int argc, char **argv) {
   // if requested from command line.
   loader.compile(ctx);
 
+  assert(!(streamInputFilenamesMode && emittingBundle()) &&
```
bertmaher: I'd `GLOW_ASSERT` this, since we'd want this check in Release mode too.

jfix71:
@bertmaher 

> In streaming mode what sets the batch size?

The first set of images that comes in sets the batch size -- the same as is currently. So if you pass in 3 images on the first inference then the batch size is 3, and all future sets of input filenames must also have batch size 3. (This first batch will be set either via stdin at runtime, or as a command-line param, depending on `stdin-filenames`).

>Are stdin-filenames and streaming-input orthogonal flags? Do they need to be? It seems like the logic could be simpler if, e.g., stdin input implies streaming.

They don't need to be orthogonal, I was just providing flexibility of different execution modes (e.g. `stdin-filenames && !streaming-input` executes one batch, but filenames for the single batch are provided at runtime via stdin). I suppose it's not super useful. I'll go ahead and combine them.

tools/loader/ImageClassifier.cpp
jfix71: Done

jfix71:
I removed `stdin-filenames` and merged it with `-`.

tools/loader/ImageClassifier.cpp
qcolombet: Alternatively we could move the logic doing the compilation inside the loop and guard it with a `isFirst` kind of thing.

jfix71:
@qcolombet I tried to do that, including a lot of refactoring to make it more readable... It still feels a bit messy to me but let me know if you think it's better this way.

tools/loader/ImageClassifier.cpp
```diff
@@ -189,6 +194,102 @@ void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
...
+  getline(std::cin, filenamesRaw);
+  std::istringstream iss(filenamesRaw);
+  std::string filename;
+  while (iss >> filename) {
+    filenames->push_back(filename);
+  }
+}
+
+/// Creates and \returns the protobuf loader given \p loader and the \p
+/// inputImageType. Note that this must come after loading images for inference
```
qcolombet: nit: `\p` and `inputImageType` should be on the same line.

tools/loader/ImageClassifier.cpp
```diff
@@ -245,96 +346,90 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
...
-  loader.compile(ctx);
+    // If this is the first run, then we need to build and compile the model.
+    if (isFirstRun) {
+      isFirstRun = false;
 
-  // If in bundle mode, do not run inference.
-  if (!emittingBundle()) {
+      // Build and compile the graph, and then get back the input Placeholder
+      // and output Softmax Tensor.
+      auto inputOutputPair = buildAndCompileAndGetInAndOutPair(
```
qcolombet: Could you use an explicit type here instead of auto?
This is not obvious from the function name what are the types held by this pair.

tools/loader/ImageClassifier.cpp
```diff
@@ -245,96 +346,90 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
...
-        *loader.getFunction(), ElemKind::FloatTy, ElemKind::Float16Ty);
-    for (auto *placeholder : loader.getModule()->getPlaceholders()) {
-      converter.convertPlaceholder(*placeholder, &ctx);
+  // These will be set during the first run.
+  Placeholder *inputImagePH = nullptr;
+  Tensor *SMVarT = nullptr;
+  do {
+    // If streaming filenames, then load the next set of them.
+    if (streamInputFilenamesMode) {
+      getNextImageFilenames(&inputImageFilenames);
```
qcolombet: Shouldn't we have something like `while(getNextImageFilenames)` instead of controlling the loop through `streamInputFilenamesMode`, which will always be true or always be false for a given run. I.e., we don't have a way of terminating the program.

tools/loader/ImageClassifier.cpp
jfix71: Done

tools/loader/ImageClassifier.cpp
jfix71: Done

tools/loader/ImageClassifier.cpp
jfix71: I updated it to include something like this (needs to be a little more complicated than `while(getNextImageFilenames)`)

jfix71:
The refactored version has grown on me a bit so I'm happy to go with it 😃

Fixed all your concerns.

PR #1912, opened 2018-10-23
[Caffe2 Importer] Add NCHW2NHWC Caffe2 op.

beicy:
*Description*:
https://caffe2.ai/docs/operators-catalogue.html#nchw2nhwc
*Testing*:
Test added.
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/tests/unittests/caffe2ImporterTest.cpp
+++ b/tests/unittests/caffe2ImporterTest.cpp
@@ -1341,3 +1341,47 @@ TEST(caffe2, sparseToDense) {
   // Graph has three inputs and one output.
   EXPECT_EQ(mod.getPlaceholders().size(), 4);
 }
+
+/// Test loading NCHW2NHWC op.
+TEST(caffe2, testNCHW2NHWC) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename(
+      "tests/models/caffe2Models/NCHW2NHWC_predict_net.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  Placeholder *output;
+  Context ctx;
+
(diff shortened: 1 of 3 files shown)


tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1341,3 +1341,41 @@ TEST(caffe2, sparseToDense) {
...
+
+  // Destroy the loader after the graph is loaded since the following execution
+  // will not depend on anyting from the loader.
+  {
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"inputs"},
+                               {&inputs.getType()}, *F);
+    output = caffe2LD.getSingleOutput();
+  }
+
+  // High level check on the content of the graph. We have 1 transpose and 1
```
jfix71: I believe checking the output shape in scope here for these tests, correct? If so can you add a check that the output shape is as expected (`{1,3,4,2}`)?

tests/unittests/caffe2ImporterTest.cpp
```diff
@@ -1341,3 +1341,41 @@ TEST(caffe2, sparseToDense) {
...
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetDescFilename(
+      "tests/models/caffe2Models/NCHW2NHWC_predict_net.pbtxt");
+  std::string NetWeightFilename(
+      "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  Placeholder *output;
+  Context ctx;
```
nadavrot: Why do we need the context?

tests/unittests/caffe2ImporterTest.cpp
beicy: Added the shape check.

tests/unittests/caffe2ImporterTest.cpp
beicy: It is actually for getting the output shape -- although the check is just added...

PR #1913, opened 2018-10-23
[Image] Fix writePngImage 90 degree rotation and black image output issues

qcolombet:
*Description*
The 90 degree rotation was simply a swapping of the column and row indices
when accessing the tensor.
The black image output was a misuse of the range input. The range represents
the range of values the image should have, not the range of values the input
tensor has.

In other words, we were doing:
- read: tensor = image * scale + bias
- write: image = tensor * scale + bias

Whereas we should be doing:
- read: tensor = image * scale + bias
- write: image = (tensor - bias ) / scale

Fixing both these issues makes the image being written properly.

*Testing*
The added test loads an existing image into a tensor, then writes it
into an image and finally load that image back.
The tensor obtained from the final image loading is expected to be
the same (modulo scale and bias rounding) as the one from the
original image.

### DIFF
--- a/tests/unittests/imageTest.cpp
+++ b/tests/unittests/imageTest.cpp
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Base/Image.h"
+
+#include "llvm/Support/FileSystem.h"
+
+#include "gtest/gtest.h"
+
+#include <cstdio>
+#include <utility>
+
+using namespace glow;
+
+TEST(Image, writePngImage) {
+  auto range = std::make_pair(0.f, 1.f);
+  Tensor localCopy;
+  bool loadSuccess =
+      !readPngImage(&localCopy, "tests/images/imagenet/cat_285.png", range);
+  ASSERT_TRUE(loadSuccess);
+
+  llvm::SmallVector<char, 10> resultPath;
+  llvm::sys::fs::createTemporaryFile("prefix", "suffix", resultPath);
+  std::string outfilename(resultPath.begin(), resultPath.end());
+
+  bool storeSuccess = !writePngImage(&localCopy, outfilename.c_str(), range);
+  ASSERT_TRUE(storeSuccess);
+
+  Tensor secondLocalCopy;
+  loadSuccess = !readPngImage(&secondLocalCopy, outfilename.c_str(), range);
+  ASSERT_TRUE(loadSuccess);
+  EXPECT_TRUE(secondLocalCopy.isEqual(localCopy, 0.01));
+
+  // Delete the temporary file.
+  std::remove(outfilename.c_str());
+}
--- a/lib/Base/Image.cpp
+++ b/lib/Base/Image.cpp
@@ -227,9 +227,9 @@ bool glow::writePngImage(Tensor *T, const char *filename,
     png_byte *row = row_pointers[y];
     for (size_t x = 0; x < width; x++) {
       png_byte *ptr = &(row[x * 4]);
-      ptr[0] = H.at({x, y, 0}) * scale + bias;
-      ptr[1] = H.at({x, y, 1}) * scale + bias;
-      ptr[2] = H.at({x, y, 2}) * scale + bias;
+      ptr[0] = (H.at({y, x, 0}) - bias) / scale;
+      ptr[1] = (H.at({y, x, 1}) - bias) / scale;
+      ptr[2] = (H.at({y, x, 2}) - bias) / scale;
       ptr[3] = 0xff;
     }
   }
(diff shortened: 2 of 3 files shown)


beicy:
Can we reuse the path "./tests/images/imagenet/" for the source image? I think it is OK to assume we run the unittest under proper directory.

qcolombet:
> Can we reuse the path "./tests/images/imagenet/" for the source image? I think it is OK to assume we run the unittest under proper directory.

Agreed.

qcolombet:
Hey @nadavrot, could you check if I am doing anything obviously wrong in that test?
I am trying to figure out if I am totally misusing writePngImage or if we have a bug (see #1914).

tests/unittests/imageTest.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <cstdio>
+#include <utility>
+
+using namespace glow;
+
+TEST(Image, writePngImage) {
+  auto range = std::make_pair(0.f, 1.f);
+  Tensor localCopy;
+  bool loadSuccess = !readPngImage(&localCopy, "cat_285.png", range);
+  ASSERT_TRUE(loadSuccess);
```
rdzhabarov: it looks like the test is failing here:
```
../tests/unittests/imageTest.cpp:30: Failure
Value of: loadSuccess
  Actual: false
Expected: true
```

it does not exercise read/write/read path yet...

qcolombet: Yeah, that's on purpose. I wanted to know if someone could spot something I was obviously doing wrong before I go all the way of adding the cmake plumbing for image copying.

rdzhabarov: haha, test looks good. One thing I noticed, although have not looked a lot:
`writePNG` uses PNG_COLOR_TYPE_RGB_ALPHA, but `readPNG` could use different mode.

jfix71: `readPNG` does strip out alpha if it exists, so that we always have 3 channels. So that does seem like a bug, should probably use `PNG_COLOR_TYPE_RGB` instead

jfix71: Ah, yeah you're right, it's added back in even though the input tensor has 3 channels.

RE: the image appearing black as mentioned in https://github.com/pytorch/glow/issues/1914, this is because of the range being 0to1. If you use 255.f instead of 1.f (so the scale/bias don't do anything) the cat image looks to be printed correctly, but it's rotated 90 degrees.

qcolombet: Ah thanks for the range. That's not obvious why we need to use something different for reading and writing. Regarding the 90 degrees rotation that sounds like there is still a bug.

qcolombet: Hi @rdzhabarov, @jfix71,

I have a quick look at the read and write functions and I was surprised that the range is used in the same way in both case. I would have expected one to be the inverse of the other.

Namely, we do:
- read: tensor = image * scale + bias
- write: image = tensor * scale + bias

I would have expected something like:
- read: tensor = image * scale + bias
- write: image = (tensor - bias )/ scale

What do you think?

qcolombet: The 90 degree rotation was easy. We used the index of the width for the row and the index of the height for the column whereas that should be the other way around.

With both the scale stuff and that fix, the images look similar.

qcolombet:
Alright now the patch has the right fixes, but I am not happy with the tmpnam thing as it triggers a deprecated warning.
I am going to remove the WIP, but I'd like you suggestion on how to fix that (I was thinking using mkstemp or std::tmpfile, but both return a file descriptor instead of a string, alternatively we could just write to a hard coded temporary file).

tests/unittests/imageTest.cpp
rdzhabarov: Nice catch @qcolombet!

rdzhabarov:
I used `llvm::sys::fs::createTemporaryFile("prefix", "suffix", resultPath);` in the past. Check quantizationTest.
I must admit that it's not cleaned though, which it should be after the test run :)

qcolombet:
Thanks for the suggestion @rdzhabarov, I'll take a look.

nadavrot:
:arrow_heading_down: :sunrise: :arrow_heading_up:

qcolombet:
@rdzhabarov is there a way to ssh into the CircleCI builder?
The test in this PR fails in libpng and I cannot reproduce it locally.

tests/unittests/imageTest.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <cstdio>
+#include <utility>
+
+using namespace glow;
+
+TEST(Image, writePngImage) {
+  auto range = std::make_pair(0.f, 1.f);
+  Tensor localCopy;
+  bool loadSuccess = !readPngImage(
+      &localCopy, ROOT_BINARY_DIR "/tests/images/imagenet/cat_285.png", range);
```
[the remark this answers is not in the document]
qcolombet: Up to us.
The problem with the onnxImporterTest is if you run it from outside of the build directory, it will fail.
E.g., ./tests/onnxImporterTest works, (cd tests; ./onnxImporterTest) doesn't.

I can remove the logic for the ROOT_BINARY_DIR stuff if that's fine for everybody.

qcolombet:
@rdzhabarov, the rerun and the ssh stuff are grayed out for me: "You need write permission to trigger builds".

rdzhabarov:
> the rerun and the ssh stuff are grayed out for me: "You need write permission to trigger builds".

have you signed in with your Github credentials?

qcolombet:
> have you signed in with your Github credentials?

That was the problem, thanks!

qcolombet:
The bug was funny to debug given the docker in circleci doesn't have access to a debugger.
Anyway, the bug was that we were using png_destroy_read_struct instead of png_destroy_write_struct and that doesn't work well on some libpng version!

Fixed that and removed the ROOT_BINARY_DIR thing as well.

qcolombet:
Hey @rdzhabarov!

I probably miss something obvious but CircleCI doesn't invoke the imageTest with the same path as what I get when running `ninja check` locally (The log shows `../tests/imageTest`, whereas locally I have `./tests/imageTest`) and given the path to the image is expected to be relative to the build directory, this fails.
Do you happen to know what I need to do to fix that?

Thanks!

qcolombet:
I was missing the WORKING_DIRECTORY option on the invocation of the imageTest.
Thanks @rdzhabarov for the hint.

PR #1917, opened 2018-10-23
[converter][easy] Assert destTy == dstKind_ when converting

bertmaher:
*Description*: Got an unused parameter lint warning and thought this was a good assert to add.
*Testing*: `ninja all && gtest-parallel tests/*Test`
*Documentation*: N/A


qcolombet:
Good catch!

PR #1925, opened 2018-10-25
[Caffe2/ONNX] Add support for LengthsSum operator

### LINKED ISSUE #1884: [New operator] LengthsSum
We need to support `LengthsSum` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#lengthssum

One has to create new Graph-level Node (`LengthsSumNode`), create new low-level IR instruction, and implement it for Interpreter and CPU backends, also adding a unit test to `OperatorTest.cpp`.

Note, that this operator is similar to `SparseLengthsSum`. Probably significant part of code can be taken from there.

SplitInfinity:
**Description**
This commit adds support for the `LengthsSum` operator, which
performs a `BatchedReduceAdd` operation on groups of consecutive
zero-axis slices of the `data` input. The sizes of these groups
are defined by the `lengths` input. This operator has implementations in
both the CPU and interpreter backends, and can be imported from a Caffe2
and ONNX model.

**Testing**
This commit adds an interpreter and CPU operator test as well as Caffe2
and ONNX importer tests.

**Fixes**
This commit fixes #1884.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1750,6 +1750,37 @@ void InterpreterFunction::fwdBatchedReduceAddInst(
   }
 }
 
+void InterpreterFunction::fwdLengthsSumInst(const LengthsSumInst *I) {
+  auto out = getTensor(I->getDest());
+  auto data = getTensor(I->getData());
+  auto lengths = getTensor(I->getLengths());
+
+  out->zero();
+
+  auto LH = lengths->getHandle<int64_t>();
+
+  size_t segments = lengths->dims()[0];
+  size_t sliceSize = data->size() / data->dims()[0];
+
+  auto DH = data->getHandle<float>();
+  auto OH = out->getHandle<float>();
+
+  size_t offsetIn = 0;
+  size_t offsetOut = 0;
+  for (size_t i = 0; i < segments; i++) {
+    for (size_t j = 0, e = LH.raw(i); j < e; j++) {
+      for (size_t k = 0; k < sliceSize; k++) {
+        OH.raw(offsetOut + k) += DH.raw(offsetIn + k);
+      }
+      offsetIn += sliceSize;
+    }
+    offsetOut += sliceSize;
+  }
+
+  assert(offsetIn == data->size() && "All values in Data should be consumed");
+  assert(offsetOut == out->size() && "All values in Dest should be written to");
+}
+
 void InterpreterFunction::fwdSparseLengthsWeightedSumInst(
     const SparseLengthsWeightedSumInst *I) {
   auto out = getTensor(I->getDest());
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -580,6 +580,10 @@ void BatchedReduceAddNode::verify() const {
   assert(getBatch().dims().size() > 0 && "Invalid shape");
 }
 
+void LengthsSumNode::verify() const {
+  assert(getLengths().dims().size() == 1 && "Lengths must be a 1D vector");
+}
+
 void SparseLengthsWeightedSumNode::verify() const {
   assert(getResult().getElementType() == getData().getElementType() &&
          "Mismatched element types");
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -504,6 +504,15 @@ public:
   BatchedAddNode *createBatchedAdd(llvm::StringRef name, TypeRef outTy,
                                    NodeValue batch, NodeValue sample);
 
+  /// Implements an operation that accumulates the values in \p data along the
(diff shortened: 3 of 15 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1749,6 +1749,37 @@ void InterpreterFunction::fwdBatchedReduceAddInst(
...
+  auto lengths = getTensor(I->getLengths());
+
+  out->zero();
+
+  auto LH = lengths->getHandle<int64_t>();
+
+  size_t segments = lengths->dims()[0];
+  size_t sliceSize = data->size() / data->dims()[0];
+
+  assert(!data->getType().isQuantizedType() &&
```
rdzhabarov: not sure how would we get quantized op here. it's not supported in c2/onnx

lib/Graph/Nodes.cpp
```diff
@@ -574,6 +574,10 @@ void BatchedReduceAddNode::verify() const {
   assert(getBatch().dims().size() > 0 && "Invalid shape");
 }
 
+void LengthsSumNode::verify() const {
+  assert(getLengths().dims().size() == 1 && "Lengths must be a 1D vector");
```
rdzhabarov: can you add this verification early in the loader?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -544,6 +544,20 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     return true;
   }
 
+  if (typeName == "LengthsSum") {
```
artemrakhov: Could you move this code and Caffe2 loader's code to `CommonOperatorLoader.h`?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1750,6 +1750,34 @@ void InterpreterFunction::fwdBatchedReduceAddInst(
...
+  for (size_t i = 0; i < segments; i++) {
+    for (size_t j = 0, e = LH.raw(i); j < e; j++) {
+      for (size_t k = 0; k < sliceSize; k++) {
+        OH.raw(offsetOut + k) += DH.raw(offsetIn + k);
+      }
+      offsetIn += sliceSize;
+    }
+    offsetOut += sliceSize;
+  }
+}
```
artemrakhov: please assert at the end that resulting offset is equal to corresponding tensor's size. That is, we used the whole tensor.

lib/Importer/ONNXModelLoader.cpp
SplitInfinity: I put it there initially, but then I moved it out to the two separate loaders because I added a check that the dimensions of the `lengths` input is 1 in the ONNX loader at Roman's request. The Caffe2 protobuf doesn't specify input sizes as far as I know so this is not applicable there.

artemrakhov: Caffe2 and ONNX protos are really very similar. In fact, ONNX format is re-branded Caffe2 proto :)

They both specify the same amount of information. We have static shapes in Glow, so `lengths` in your code is output of some other node (probably Variable). You always know its number of dims. So I don't see why this code can't be shared.

Now as for moving assertion here. I guess this question is bigger than this PR. `len(lengths) > 1` totally does not make sense, it means the input is broken. Should we do this validation or should this be done on higher level? And we'd just assert? I'm sure there are move such places, other than this one. @rdzhabarov

lib/Graph/Nodes.cpp
artemrakhov: Why?
It shouldn't be in the loader then, it should be in `createLengthsSum()` method. But I'm still curious to understand what goes to this verifier, and what should be checked right away.

rdzhabarov: Incorrect user input should be validated as soon as possible.
And in the case of ONNXIFI, we should report that we do not support this early on rather than failing on compilation/execution.

lib/Importer/ONNXModelLoader.cpp
rdzhabarov: > They both specify the same amount of information. We have static shapes in Glow, so lengths in your code is output of some other node (probably Variable). You always know its number of dims. So I don't see why this code can't be shared.

Code sharing is fine as long as we can report from the shared code that parsing of certain operator failed.

SplitInfinity: I think there is, but I'll just have to break convention and write a parser for this op that returns `bool` instead of `void` in `CommonOperatorLoader.h`.

artemrakhov: It's fine to break this convention. We will have to break it later anyway.

nadavrot:
Nice 🎊 🍔

PR #1926, opened 2018-10-25
Mark quantize, dequantize and rescale_quantized as data-parallel operations

opti-mix:
Adjust CPU and OpenCL backends accordingly. Allmost all of the required changes are mostly mechanical.

We forgot to mark these quantization-related operations as data-parallel and I noticed it just now.

*Description*:
*Testing*:
ninja test
run.sh for all of our NN models

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -632,9 +632,13 @@ void OpenCLFunction::execute(Context &ctx) {
                                      : ElemKind::FloatTy;
     std::string kernelName = getKernelName(I.getKindName(), elemTy);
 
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -712,63 +717,45 @@ void OpenCLFunction::execute(Context &ctx) {
             setKernelArg(kernel, ++numArgs, resultScaleParams);
           }
         }
+        // Quantize floating point tensor. Scale and Offset are based on return
+        // type
```
rdzhabarov: can you fix the layout here

opti-mix: Done!

PR #1928, opened 2018-10-25
[onnxifi] Throttle input concurrency with fixed number of threads on Glow side, configured per backend.

rdzhabarov:
*Description*:
This patch should allow handling of concurrent execution of inferences on the caller side.
On the Glow side, we limit concurrency by a number of threads in a thread pool per backend.
This PR should allow running multithreaded C2 execution on a single Glow thread.

*Testing*:
onnxifi integration test

*Documentation*:
N/A

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -60,8 +60,8 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
     return ONNXIFI_STATUS_INTERNAL_ERROR;
   }
 
-  onnxNameToInputVar_ = loader->getInputVarsMapping();
-  onnxNameToOutputNode_ = loader->getOutputVarsMapping();
+  onnxInputToPlaceholder_ = loader->getInputVarsMapping();
+  onnxOutputToPlaceholder_ = loader->getOutputVarsMapping();
 
   // Emit IR for the graph and compile it.
   backendPtr_->getEE().compile(CompilationMode::Infer, function_, ctx_);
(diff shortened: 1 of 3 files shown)


rdzhabarov:
@yinghai @SplitInfinity This is to fix concurrent execution through onnxifi.
This PR allows Glow backend to run on a single thread no matter of input (C2->onnxifi->glow) concurrency. Once @gcatron makes a progress on the concurrent execution of compiled model on Glow side, we could increase the number of workers to hardware::concurrency for CPU.

rdzhabarov:
@bertmaher mind taking a look?

Mostly it's limiting concurrency and performing setIO/run atomically.

lib/Onnxifi/Base.cpp
```diff
@@ -60,20 +60,46 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
...
+  llvm::DenseMap<Placeholder *, onnxPointer> outputPlaceholderToBuffer =
+      outputPlaceholderToBuffer_;
+
+  // Once inputs are copied we can allow processing concurrent requests.
+  inputRunMutex_.unlock();
+
+  // Submit graph for asynchronous execution.
+  // Concurrency for executing 'run' method is limited by number of threads in
+  // the backend specific thread pool.
+  backend()->runAsync([inputEvent, outputEvent, inputPlaceholderToBuffer,
```
yinghai: Is it good to capture these maps by value?

rdzhabarov:
@yinghai that should be tested through the onnxifi driver tests, setIO/run should go through.

yinghai:
Do we have a multithreaded test?

lib/Onnxifi/Base.cpp
rdzhabarov: That's copied intentionally so that those maps internal to graph class are not shared.

rdzhabarov:
Probably not, so some part is not going to be exercised (proper locking specifically).
One way is to run a concurrent test for resnet50 (not sure how hard would be to modify that) or benchmark one (easier to setup).

That's not going to be trivial to add test support right here.

yinghai:
> Sorry for the delayed review -- this looks pretty good so I'll accept to unblock you. I'm a bit nervous about holding a lock from onnxSetGraphIO to onnxRunGraph. But I'm also not sure I understand the onnxifi interface well enough to say whether it's a real problem. Happy to land this now and iterate later, though.

This a bit hacky but I think it gets the job done for now. We actually need `onnxRunGraph(inputDescriptors, outputDescripors)` To solve this problem properly. @Maratyszcza

PR #1931, opened 2018-10-25
[WIP]Full fp16 interpreter

qcolombet:
*Description*:
This PR implements the missing pieces for fp16 support in the interpreter to support all the operators used in inference.

*Testing*:
TODO, thus WIP.

### DIFF
--- a/lib/Backends/Interpreter/InterpreterFunction.h
+++ b/lib/Backends/Interpreter/InterpreterFunction.h
@@ -131,6 +131,61 @@ private:
 
   template <typename ElemTy>
   void fwdBatchOneHotImpl(const glow::BatchOneHotInst *I);
+
+  template <typename ElemTy>
+  void fwdSigmoidInst_FloatImpl(const SigmoidInst *I);
+
+  template <typename ElemTy> void fwdTanhInst_FloatImpl(const TanhInst *I);
+
+  template <typename ElemTy>
+  void fwdCrossEntropyLossInst_FloatImpl(const CrossEntropyLossInst *I);
+
+  template <typename ElemTy>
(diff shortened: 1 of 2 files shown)


qcolombet:
The first commit that extends float16 with the missing operators is track in #1930.

jfix71:
Oh just realized it's WIP -- ignore me 😄

qcolombet:
Thanks for taking a stab at it @jfix71.
I was mainly missing the tests so your feedback are already useful!

Good point for the static_assert, I forgot to add those this time :S. I am not sure I can push them in the macro though, since it has to be in the templated function itself.

jfix71:
Hmm, I was assuming the preprocessor would put them in the correct location before the static assert is checked.

nadavrot:
Nice!

qcolombet:
> Hmm, I was assuming the preprocessor would put them in the correct location before the static assert is checked.

Re-reading what you were saying, we were talking about something different :). You're right a new macro to insert the static_assert would do the trick (as opposed to extending the one that I added, which is what I thought you were suggesting).

nadavrot:
This PR looks really good. Is this blocked on anything?

qcolombet:
I want to double check what I was saying in my last comment: we can create fp16 variant of all the nodes.

qcolombet:
The only problematic create methods (they hardcode FloatTy) are:
GRU
SimpleRNN
LSTM
rowwiseQuantizedFC
quantizationProfile
Dequantize

We probably don't need quantization support on fp16, at least for now and the complex nodes (GRU, SimpleRNN, LSTM) can be expanded into simpler ones, so strictly speaking we are not block here.

We'll add those as we need. Therefore, I am going to coming this one as is.

PR #1934, opened 2018-10-26
Run a round of DCE before any other graph optimizations

opti-mix:
*Description*:

Optimizer may be called after backend specific transformations and some nodes may have become unused. It is a good idea to remove them, before proceeding with any further optimizations. Without it some optimizations do not kick in as they may check for the number of users of a given node and count the unused nodes as users as well.

I caught this issue when I was investigating why transpose(constant) is not optimized away at compile time after some backend-specific transformations. The reason was that the constant was used by the newly created transpose node and dead, but not yet removed node that initially used it.

*Testing*:
ninja test

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -2089,6 +2089,11 @@ void glow::convertPlaceholdersToConstants(Function *F, const Context &ctx,
 }
 
 void glow::optimize(Function *F, CompilationMode mode) {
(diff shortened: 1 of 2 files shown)


jfix71:
Is it possible to add a simple test showing the problem? E.g. in graphOptzTest create a graph that exhibits the problematic pattern you saw (with the dead Constant) and then call optimize.

opti-mix:
@jfix71 Done. Added a test.

PR #1943, opened 2018-10-31
Migrate CPU Backend to allocate memory at runtime.

gcatron:
*Description*: Migrate CPU Backend to allocate and initialize memory at runtime. Added a new runtimeInfo object that contains information needed for runtime memory allocation. 
*Testing*: ninja test passes, AOT bundles still work, testing with run.sh --iterations=20 -cpu -time showed minimal performance difference.
*Documentation*: N/A
Progress on #1904

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -16,23 +16,50 @@
 #ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
 #define GLOW_BACKENDS_COMPILEDFUNCTION_H
 
+#include "glow/Graph/Nodes.h"
 #include <unordered_map>
 
 namespace glow {
 
 class Context;
-
+namespace runtime {
+/// RuntimeSymbolInfo
+/// Contains information for initialization and handling of symbol at runtime.
+struct RuntimeSymbolInfo {
+  /// The size in bytes.
+  size_t size;
+  /// Offset in bytes from the base address.
+  size_t offset;
+};
+/// Runtime Bundle
+/// Contains the information needed to be passed forward from compile time to
+/// runtime. In order to allocate and initialize memory.
+struct RuntimeBundle {
+  /// Map from symbol name to a RuntimeSymbolInfo.
+  std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable;
+  /// Pointer to memory containing the weights for execution.
+  uint8_t *constants;
+  /// Amount of memory needed for weights.
+  const size_t constantWeightVarsMemSize;
+  /// Amount of memory needed for mutable vars.
+  const size_t mutableWeightVarsMemSize;
+  /// Amount of memory needed for activations.
+  const size_t activationsMemSize;
+  RuntimeBundle(size_t constWeight, size_t mutableWeight, size_t activations)
+      : constantWeightVarsMemSize(constWeight),
+        mutableWeightVarsMemSize(mutableWeight),
+        activationsMemSize(activations) {}
+};
+} // end namespace runtime
 /// Interface for executing a compiled function.
 class CompiledFunction {
 public:
   /// Dtor.
   virtual ~CompiledFunction() = default;
-
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
   virtual void execute(Context &ctx) = 0;
 };
-
 } // end namespace glow
 
 #endif // GLOW_BACKENDS_COMPILEDFUNCTION_H
--- a/lib/Backends/CPU/AllocationsInfo.cpp
+++ b/lib/Backends/CPU/AllocationsInfo.cpp
@@ -16,6 +16,7 @@
 #define DEBUG_TYPE "jit-allocations"
 
 #include "AllocationsInfo.h"
+#include "glow/Backends/CompiledFunction.h"
 #include "glow/CodeGen/MemoryAllocator.h"
 #include "glow/Graph/Context.h"
 #include "glow/Graph/Graph.h"
@@ -23,6 +24,7 @@
 #include "glow/IR/IRUtils.h"
 #include "glow/IR/Instrs.h"
 #include "glow/Support/Debug.h"
+#include "glow/Support/Memory.h"
 
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
@@ -32,9 +34,7 @@ using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
-void AllocationsInfo::allocateWeightVars(const IRFunction *F,
-                                         const Context &ctx,
-                                         bool absoluteAddr) {
+void AllocationsInfo::allocateWeightVars(const IRFunction *F) {
   // Use two different allocators, because constant weights and mutable weights
   // may use different memory blocks.
   MemoryAllocator constantWeightVarsAllocator("ConstantWeights", 0);
@@ -43,40 +43,21 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
   // Compute the new offsets for all the weights, do not reuse their current
   // addresses. Process all constant WeightVars first.
   for (auto &v : F->getGraph()->getParent()->getConstants()) {
-    assert(isa<WeightVar>(F->getWeightForNode(v)));
+    assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
     auto *w = cast<WeightVar>(F->getWeightForNode(v));
     auto numBytes = w->getSizeInBytes();
     size_t addr = constantWeightVarsAllocator.allocate(numBytes, w);
-    if (!absoluteAddr) {
-      allocatedAddressed_[w] = addr;
-    } else {
-      // Reuse the address used by the payload.
-      allocatedAddressed_[w] =
-          v->getPayload().getUnsafePtr() - static_cast<char *>(nullptr);
-    }
+    allocatedAddress_[w] = addr;
   }
 
-  if (absoluteAddr) {
-    // Allocate addresses for the Placeholders that have payloads defined at
-    // compile-time.
-    // TODO: Remove this branch once Context becomes a parameter of the
-    // CompiledFunction::execute method.
-    for (auto PH : ctx.pairs()) {
-      assert(isa<WeightVar>(F->getWeightForNode(PH.first)));
-      auto *w = cast<WeightVar>(F->getWeightForNode(PH.first));
-      // Reuse the address used by the payload.
-      allocatedAddressed_[w] =
-          PH.second->getUnsafePtr() - static_cast<char *>(nullptr);
-    }
-  } else {
-    // Allocate based on size as reported by the formal type of Placeholders
-    for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
-      assert(isa<WeightVar>(F->getWeightForNode(v)));
-      auto *w = cast<WeightVar>(F->getWeightForNode(v));
-      auto numBytes = w->getSizeInBytes();
-      size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-      allocatedAddressed_[w] = addr;
-    }
+  // Compute the offsets and total memory requirements for Placeholders.
+  for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+    // Get the WeightVar for each Placeholder to calculate offsets.
+    assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto numBytes = w->getSizeInBytes();
+    size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
+    allocatedAddress_[w] = addr;
   }
 
   // Remember that max required memory size for each kind of weights.
@@ -84,7 +65,7 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
   mutableWeightVarsMemSize_ = mutableWeightVarsAllocator.getMaxMemoryUsage();
 
   DEBUG_GLOW(for (auto &A
-                  : allocatedAddressed_) {
+                  : allocatedAddress_) {
     if (isa<AllocActivationInst>(A.first) || isa<TensorViewInst>(A.first))
       continue;
     assert(valueNumbers_.count(A.first) && "Unknown weight");
@@ -94,11 +75,45 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
             : "mutable weight";
     llvm::errs() << "Allocated " << kind << " " << A.first->getName()
                  << " size: " << A.first->getSizeInBytes()
-                 << "  address range:  [" << allocatedAddressed_[A.first]
-                 << ", "
-                 << allocatedAddressed_[A.first] + A.first->getSizeInBytes()
+                 << "  address range:  [" << allocatedAddress_[A.first] << ", "
+                 << allocatedAddress_[A.first] + A.first->getSizeInBytes()
                  << "]\n";
   });
+}
+
+void AllocationsInfo::collectConstants(const IRFunction *F) {
+
+  // At compile time condense constants to a single block of memory.
+  // This allows the graph to go away after compile time.
+  baseConstantWeightVarsStore_ =
+      (uint8_t *)alignedAlloc(constantWeightVarsMemSize_, TensorAlignment);
+  for (auto &v : F->getGraph()->getParent()->getConstants()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto payload = v->getPayload().getUnsafePtr();
+    auto numBytes = w->getSizeInBytes();
+    auto addr = allocatedAddress_[w];
+    // Copy weight to offset.
+    memcpy(baseConstantWeightVarsStore_ + addr, payload, numBytes);
+  }
+}
+
+runtime::RuntimeBundle
+AllocationsInfo::generateRuntimeBundle(const IRFunction *F) {
+  runtime::RuntimeBundle info(constantWeightVarsMemSize_,
+                              mutableWeightVarsMemSize_, activationsMemSize_);
+  std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
+  info.constants = baseConstantWeightVarsStore_;
+  for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    runtime::RuntimeSymbolInfo symbol;
+    symbol.offset = allocatedAddress_[w];
+    symbol.size = w->getSizeInBytes();
+    symbolTable.emplace(std::string(v->getName()), symbol);
+  }
+  info.symbolTable = std::move(symbolTable);
+  return info;
 }
 
 void AllocationsInfo::allocateActivations(const IRFunction *F) {
@@ -131,15 +146,14 @@ void AllocationsInfo::allocateActivations(const IRFunction *F) {
 
   // Register specific addresses within the heap to activations.
   for (auto &A : activationAddr) {
-    allocatedAddressed_[A.first] = A.second;
+    allocatedAddress_[A.first] = A.second;
   }
   DEBUG_GLOW(for (auto &A
-                  : allocatedAddressed_) {
+                  : allocatedAddress_) {
     llvm::errs() << "Allocated activation " << A.first->getName()
                  << " size: " << A.first->getSizeInBytes()
-                 << "  address range:  [" << allocatedAddressed_[A.first]
-                 << ", "
-                 << allocatedAddressed_[A.first] + A.first->getSizeInBytes()
+                 << "  address range:  [" << allocatedAddress_[A.first] << ", "
+                 << allocatedAddress_[A.first] + A.first->getSizeInBytes()
                  << "]\n";
   });
 }
@@ -174,18 +188,18 @@ void AllocationsInfo::allocateTensorViews(const IRFunction *F) {
   for (const auto &I : F->getInstrs()) {
     if (const auto *TVI = dyn_cast<TensorViewInst>(&I)) {
       auto *viewOrigin = getOrigin(TVI);
-      assert(allocatedAddressed_.count(viewOrigin) &&
+      assert(allocatedAddress_.count(viewOrigin) &&
              "Did not find original WeightVar or AllocActivation for a "
              "TensorView.");
-      size_t originAddr = allocatedAddressed_[viewOrigin];
+      size_t originAddr = allocatedAddress_[viewOrigin];
 
       // Calculate the offset into the underlying alloc activation.
       size_t offset = calculateTensorViewOffset(TVI);
 
       // Calculate the correct address using this offset into the alloc
       // activation and map from the original TVI to it.
-      assert(!allocatedAddressed_.count(TVI) && "Allocation already made!");
-      allocatedAddressed_[TVI] = originAddr + offset;
+      assert(!allocatedAddress_.count(TVI) && "Allocation already made!");
+      allocatedAddress_[TVI] = originAddr + offset;
       continue;
     }
   }
--- a/lib/Backends/CPU/AllocationsInfo.h
+++ b/lib/Backends/CPU/AllocationsInfo.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 #define GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 
+#include "glow/Graph/Nodes.h"
 #include "llvm/IR/Module.h"
 
 #include <functional>
@@ -27,6 +28,9 @@ class WeightVar;
 class Constant;
 class Context;
 
+namespace runtime {
+struct RuntimeBundle;
+}
 /// Information about allocations for activations, constant weight variables
 /// and mutable weight variables.
 struct AllocationsInfo {
@@ -39,42 +43,39 @@ struct AllocationsInfo {
   /// numberOffsets_[valueNumbers_[v]]
 
   /// Maps Values in the module to their offsets.
-  llvm::DenseMap<const Value *, uint64_t> allocatedAddressed_;
+  llvm::DenseMap<const Value *, uint64_t> allocatedAddress_;
   /// Amount of memory to be allocated for constant WeightVars.
   size_t constantWeightVarsMemSize_{0};
   /// Amount of memory to be allocated for mutable WeightVars.
   size_t mutableWeightVarsMemSize_{0};
   /// Amount of memory to be allocated for activations.
   size_t activationsMemSize_{0};
+  /// Base address of stored constant weights.
+  uint8_t *baseConstantWeightVarsStore_{nullptr};
   /// Base address of constant weights.
-  uint8_t *baseConstantWeightVarsAddress_{nullptr};
-  /// Base address of mutable WeightVars.
-  uint8_t *baseMutableWeightVarsAddress_{nullptr};
-  /// Base address of activations.
-  uint8_t *baseActivationsAddress_{nullptr};
 
   /// Assign offsets to all of the variables in the module \p M and to the
-  /// placeholders. \p ctx is the context that maps the graph to the concrete
-  /// execution environment for a specific function.
-  /// If the \p absoluteAddr is true, simply reuse the addresses already used
-  /// by the payloads of tensors corresponding to those WeightVars as offsets.
-  /// This is useful in a JIT setup. If \p absoluteAddr is false, then all the
-  /// WeightVars will get new offsets assigned.
-  void allocateWeightVars(const IRFunction *F, const Context &ctx,
-                          bool absoluteAddr);
+  /// placeholders.
+  void allocateWeightVars(const IRFunction *F);
   /// Assign offsets to all activations.
   /// No actual memory allocation is performed. All the allocations should be
   /// performed by the client based on the information provided by the
-  /// AllocationsInfo.
+  /// AllocationsInfo or RuntimeBundle.
   void allocateActivations(const IRFunction *F);
   /// Assign offsets to all tensorviews.
   /// No memory allocation is performed. Sets up all offsets into already
   /// defined offsets for WeightVars and AllocActivations. Assumes the weight
-  /// vars and alloc activations have already been added to allocatedAddressed_.
+  /// vars and alloc activations have already been added to allocatedAddress_.
   void allocateTensorViews(const IRFunction *F);
   /// Number all allocations and weight variables by assigning them unique
   /// numbers.
   void numberValues(const IRFunction *F);
+
+  /// Collect Constants into a single block of memory.
+  void collectConstants(const IRFunction *F);
+  /// Returns runtimeBundle object containing offsets and allocation sizes
+  /// needed for runtime.
+  runtime::RuntimeBundle generateRuntimeBundle(const IRFunction *F);
 };
 
 } // namespace glow
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -41,16 +41,17 @@ namespace {
 //                   Functions for executing code using JIT
 //===----------------------------------------------------------------------===//
 
-/// Emit the entry point for JIT called "jitmain". It simply calls the main
-/// entry of the module with the constant concrete addresses of all the memory
-/// areas. Since these addresses are constants, the LLVM optimizer will constant
-/// propagate them into relative addressing computations and the like and
-/// produce a very efficient code that uses absolute addressing whenever
-/// possible.
+/// Emit the entry point for JIT called "jitmain".
+/// Function has the following API:
+///   void jitmain(uint8_t *baseConstantWeightVars,
+///                uint8_t *baseInOutWeightVars,
+///                nuint8_t *baseActivations);
 static void emitJitMain(LLVMIRGen &irgen) {
   AllocationsInfo &allocationsInfo = irgen.getAllocationsInfo();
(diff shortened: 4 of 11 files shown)


include/glow/Backends/CompiledFunction.h
```diff
@@ -16,11 +16,25 @@
 #ifndef GLOW_BACKENDS_COMPILEDFUNCTION_H
 #define GLOW_BACKENDS_COMPILEDFUNCTION_H
 
+#include <glow/Graph/Nodes.h>
```
nadavrot: Why do we need nodes.h?  Can we use a forward declaration?

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,11 +16,25 @@
...
 #include <unordered_map>
 
 namespace glow {
 
 class Context;
+class Placeholder;
+
+struct runtimeSymbolInfo {
```
nadavrot: Doxygen please.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,11 +16,25 @@
...
+  size_t size;
+  size_t offset;
+};
+struct runtimeInfo {
```
nadavrot: doxygen for the class decl and all members.

include/glow/Backends/CompiledFunction.h
nadavrot: We use the "_" suffix for class members unless they are really simple pods like (point.x  point.y).

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -47,36 +46,16 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
-  } else {
-    // Allocate based on size as reported by the formal type of Placeholders
-    for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
-      assert(isa<WeightVar>(F->getWeightForNode(v)));
-      auto *w = cast<WeightVar>(F->getWeightForNode(v));
-      auto numBytes = w->getSizeInBytes();
-      size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-      allocatedAddressed_[w] = addr;
-    }
+  // Allocate based on size as reported by the formal type of Placeholders
```
nadavrot: Period at the end of the sentence.

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -47,36 +46,16 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+  for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
```
nadavrot: Could you add a comment here that explains what's going on?

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -101,6 +80,43 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
   });
 }
 
+void AllocationsInfo::collectConstants(const IRFunction *F) {
+
+  // At compile time condense constants to a single block of memory
+  // This allows the graph to go away after compile time
```
nadavrot: period.

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -101,6 +80,43 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+  baseConstantWeightVarsStore_ =
+      (uint8_t *)alignedAlloc(constantWeightVarsMemSize_, TensorAlignment);
+  for (auto &v : F->getGraph()->getParent()->getConstants()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto payload = v->getPayload().getUnsafePtr();
+    auto numBytes = w->getSizeInBytes();
+    auto addr = allocatedAddressed_[w];
+    // copy weight to offset.
```
nadavrot: Capital letter to start the sentence.

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -101,6 +80,43 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+    memcpy(baseConstantWeightVarsStore_ + addr, payload, numBytes);
+  }
+}
+
+std::unique_ptr<runtimeInfo>
+AllocationsInfo::generateRuntimeInfo(const IRFunction *F) {
+  std::unique_ptr<runtimeInfo> info(new runtimeInfo{});
```
nadavrot: Can you add a few comments here?

lib/Backends/CPU/AllocationsInfo.h
```diff
@@ -16,6 +16,7 @@
 #ifndef GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 #define GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 
+#include "glow/Backends/CompiledFunction.h"
```
nadavrot: Do we need this header or can we use a forward declaration?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -41,16 +41,16 @@ namespace {
...
 
-/// Emit the entry point for JIT called "jitmain". It simply calls the main
-/// entry of the module with the constant concrete addresses of all the memory
-/// areas. Since these addresses are constants, the LLVM optimizer will constant
-/// propagate them into relative addressing computations and the like and
-/// produce a very efficient code that uses absolute addressing whenever
-/// possible.
+/// Emit the entry point for JIT called "jitmain".
+/// Function has the following API:
+/// void jitmain(uint8_t *baseConstantWeightVars, uint8_t *baseInoutWeightVars,
```
nadavrot: Can you add two spaces between the comment marker and the "void" to make sure that the code looks like doc and not some commented out code? I am not sure if clang-format will allow this.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,87 @@
...
+CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
+                         std::unique_ptr<runtimeInfo> runtimeInfo)
+    : JIT_(std::move(JIT)), runtimeInfo_(std::move(runtimeInfo)) {}
+
+CPUFunction::~CPUFunction() { free(runtimeInfo_->constants); }
+
+void CPUFunction::loadFunctionOnDevice() {
+  baseConstantWeightVarsAddress_ = (uint8_t *)alignedAlloc(
+      runtimeInfo_->constantWeightVarsMemSize, TensorAlignment);
+  // Copy Constants to device
```
nadavrot: Period.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,12 +27,29 @@ class CPUFunction final : public CompiledFunction {
   /// The LLVM JIT engine. The jit must be initialized after the ctor
   /// initializes the LLVM backends.
   std::unique_ptr<llvm::orc::GlowJIT> JIT_;
-  /// This represents the heap, that stores the activations at runtime.
-  void *heap_;
+
+  /// runtimeInfo contains symbol offsets and allocation sizes
```
nadavrot: Period.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -76,6 +76,7 @@ void glow::updateInputPlaceholdersByName(Context &ctx, Module *mod,
 
 void ExecutionEngine::run(Context &ctx) {
   assert(function_ && "No function has been compiled");
+  // TODO call setup functions here not in execute.
```
nadavrot: Can you elaborate here?

For example: "We are working on migrating our API from A to B. In the process we will implement C. See task #123456 "

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,87 @@
...
     JitFuncType funcPtr = reinterpret_cast<JitFuncType>(address.get());
-    funcPtr();
+    funcPtr(baseConstantWeightVarsAddress_, baseMutableWeightVarsAddress_,
+            baseActivationsAddress_);
+    fetchFromDevice(ctx);
   } else {
     GLOW_ASSERT(false && "Error getting address.");
   }
-}
+  freeAllocations();
```
nadavrot: new line at the end of the file please.

include/glow/Backends/CompiledFunction.h
opti-mix: Type names in our sources typically start with an upper-case letter.

Please provide comments for the type and all of its fields/methods.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,11 +16,25 @@
...
+  std::unordered_map<std::string, runtimeSymbolInfo> symbolTable;
+  uint8_t *constants;
+  size_t constantWeightVarsMemSize;
```
opti-mix: Are all these `MemSize` mutable? If not, may be they should be `const`?

include/glow/Backends/CompiledFunction.h
opti-mix: I'm also not quite happy about the name of the type. It sounds too general to me. May be use something more specific? And/or may be it should be in its own namespace, e.g. `glow::runtime` or `glow::execution`?

lib/Backends/CPU/AllocationsInfo.h
```diff
@@ -75,6 +72,11 @@ struct AllocationsInfo {
   /// Number all allocations and weight variables by assigning them unique
   /// numbers.
   void numberValues(const IRFunction *F);
+
+  /// Collect Constants into a single block of memory
```
opti-mix: Add period at the end of this sentence.

lib/Backends/CPU/AllocationsInfo.h
```diff
@@ -75,6 +72,11 @@ struct AllocationsInfo {
...
+  void collectConstants(const IRFunction *F);
+
+  std::unique_ptr<runtimeInfo> generateRuntimeInfo(const IRFunction *F);
```
opti-mix: Add a doxygen comment.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,87 @@
...
+        runtimeInfo_->symbolTable.find(std::string(PH.first->getName()));
+    assert(symbolInfo != runtimeInfo_->symbolTable.end() && "Symbol not found");
+    auto addr = symbolInfo->second.offset;
+    auto numBytes = symbolInfo->second.size;
+    // copy PH to allocated memory.
+    memcpy(baseMutableWeightVarsAddress_ + addr, payload, numBytes);
+  }
+}
+
+void CPUFunction::fetchFromDevice(Context &ctx) {
```
opti-mix: Since you have a `loadMutablesOnDevice` method, the name for this one should be more symmetric, e.g. `fetchMutablesFromDevice`.

But in other backends, we usualy use names like 'copySomethingToDevice` and `copySomethingFromDevice`.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,12 +27,29 @@ class CPUFunction final : public CompiledFunction {
...
+  uint8_t *baseMutableWeightVarsAddress_;
 
 public:
   /// Ctor.
-  CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT, void *heap);
+  CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
+              std::unique_ptr<runtimeInfo> runtimeInfo);
+
+  /// Copy Function and Constant Memory to device.
+  void loadFunctionOnDevice();
```
opti-mix: I wonder if we always want to combine copying of a function (on a CPU no code needs to be copied, but on accelerators we may need to actually copy code) and copying of constant weights into a single function like `loadFunctionOnDevice`? May be it should be two functions? Or a single function that internally calls two different functions?

opti-mix: A more general question: Should the CompiledFunction class be responsible for copying to/from devices and for memory allocation? It seems to me that these are runtime/execution aspects and thus belong into a separate API. CompiledFunction objects themselves are passive objects that should be manipulated/executed by other subsystems IMHO.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,87 @@
...
   assert(sym && "Unable to JIT the code!");
-  using JitFuncType = void (*)(void);
+  using JitFuncType =
+      void (*)(uint8_t * constantWeightVars, uint8_t * mutableWeightVars,
+               uint8_t * activations);
   auto address = sym.getAddress();
   if (address) {
     JitFuncType funcPtr = reinterpret_cast<JitFuncType>(address.get());
-    funcPtr();
+    funcPtr(baseConstantWeightVarsAddress_, baseMutableWeightVarsAddress_,
```
opti-mix: This part related to the invocation of a JITed function looks good!

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -41,16 +41,16 @@ namespace {
...
+/// uint8_t *baseActivations);
 static void emitJitMain(LLVMIRGen &irgen) {
   AllocationsInfo &allocationsInfo = irgen.getAllocationsInfo();
   llvm::Type *voidTy = llvm::Type::getVoidTy(irgen.getLLVMContext());
-  llvm::FunctionType *jitFuncTy = llvm::FunctionType::get(voidTy, {}, false);
+  auto int8PtrTy = llvm::Type::getInt8PtrTy(irgen.getLLVMContext());
```
opti-mix: The changed logic for creation of the `jitmain` function looks good.

lib/Backends/CPU/CPUFunction.h
opti-mix: > Compiled function already has execute which is responsible for the actual execution, so I can see a case for keeping all runtime logic co-located. But maybe it's cleaner to separate that functionality to a separate API. Can we remove execute too? Then CompiledFunction would be just the code itself.

Yeah, I think mid term we most likely want the CompiledFunction to be just the code itself and have a separate API/logical system components to run/execute one or more CompiledFunctions. In case of a CPU execution, such a component may be e.g. responsible for managing threads, deciding which CompiledFunction should run in which thread, handle thread local memory @rdzhabarov mentioned, etc.

include/glow/Backends/CompiledFunction.h
gcatron: I may be going about it wrong, but it seems it needs to be there as part of the hash validity check on the unordered_map.

include/glow/Backends/CompiledFunction.h
gcatron: I like the glow::runtime namespace, I moved it in there and renamed it to RuntimeBundle and RuntimeSymbolInfo.

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -47,36 +46,16 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
-      size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-      allocatedAddressed_[w] = addr;
-    }
+  // Allocate based on size as reported by the formal type of Placeholders
+  for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto numBytes = w->getSizeInBytes();
+    size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
+    allocatedAddressed_[w] = addr;
```
rdzhabarov: nit: allocatedAddressed_ - typo here?

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -101,6 +80,43 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto payload = v->getPayload().getUnsafePtr();
+    auto numBytes = w->getSizeInBytes();
+    auto addr = allocatedAddressed_[w];
+    // copy weight to offset.
+    memcpy(baseConstantWeightVarsStore_ + addr, payload, numBytes);
+  }
+}
+
+std::unique_ptr<runtimeInfo>
```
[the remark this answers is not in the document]
gcatron: Good catch, the thought was it would be better not to copy around a large object but it's actually quite small. Now passing directly.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,12 +16,44 @@
...
+#include "glow/Graph/Nodes.h"
 #include <unordered_map>
 
 namespace glow {
 
 class Context;
+namespace runtime {
+/// Runtime Symbol Info
+/// Contains Information for initialization and handling of symbol at runtime.
+/// \p size is the size in bytes of the symbol.
```
opti-mix: The comments for members are put right before the members.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,12 +16,44 @@
...
+struct RuntimeSymbolInfo {
+  size_t size;
+  // namespace runtime
+  size_t offset;
+};
 
+/// Runtime Bundle
+/// Contains the information needed to be passed forward from compile time to
+/// runtime.
+/// \p symbolTable is a map from symbol name to a RuntimeSymbolObject.
```
opti-mix: Same here.

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -94,13 +75,50 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
+    memcpy(baseConstantWeightVarsStore_ + addr, payload, numBytes);
+  }
+}
+
+runtime::RuntimeBundle
+AllocationsInfo::generateRuntimeBundle(const IRFunction *F) {
+  runtime::RuntimeBundle info(constantWeightVarsMemSize_,
+                              mutableWeightVarsMemSize_, activationsMemSize_);
+  std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
+  // info->mutableWeightVarsMemSize = mutableWeightVarsMemSize_;
```
opti-mix: Remove the commented out code?

lib/Backends/CPU/AllocationsInfo.cpp
```diff
@@ -47,44 +47,25 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F,
...
-      assert(isa<WeightVar>(F->getWeightForNode(v)));
-      auto *w = cast<WeightVar>(F->getWeightForNode(v));
-      auto numBytes = w->getSizeInBytes();
-      size_t addr = mutableWeightVarsAllocator.allocate(numBytes, w);
-      allocatedAddressed_[w] = addr;
-    }
+  // Compute the offsets and total memory requirements for Placeholders.
+  for (auto &v : F->getGraph()->getParent()->getPlaceholders()) {
+    // Get the WeightVar for each Placeholder to calculate offsets.
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
```
opti-mix: asserts typically should have an error message.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,12 +27,31 @@ class CPUFunction final : public CompiledFunction {
   /// The LLVM JIT engine. The jit must be initialized after the ctor
   /// initializes the LLVM backends.
   std::unique_ptr<llvm::orc::GlowJIT> JIT_;
-  /// This represents the heap, that stores the activations at runtime.
-  void *heap_;
+
+  /// runtimeBundle contains symbol offsets and allocation sizes.
+  runtime::RuntimeBundle runtimeBundle_;
+
+  uint8_t *baseConstantWeightVarsAddress_;
```
opti-mix: Add doxygen comments for all of  these base addresses.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,12 +27,31 @@ class CPUFunction final : public CompiledFunction {
...
+  CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
+              runtime::RuntimeBundle runtimeBundle);
+
+  /// Copy Function to device, an empty function for CPU.
+  void copyFunctionToDevice(){};
+  /// Copy Constants to device.
+  void copyConstantsToDevice();
+  /// Allocate Mutable buffers on device, Activations and Placeholders.
+  void allocateMutableBuffersOnDevice();
+  /// Copy Mutable data to device
```
opti-mix: Dot at the end of each comment.

lib/Backends/CPU/AllocationsInfo.cpp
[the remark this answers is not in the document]
rdzhabarov: For return value it does not really matter because of RVO.
For passing it around you could use const reference.

include/glow/Backends/CompiledFunction.h
```diff
@@ -25,27 +25,30 @@ class Context;
 namespace runtime {
 /// Runtime Symbol Info
 /// Contains Information for initialization and handling of symbol at runtime.
-/// \p size is the size in bytes of the symbol.
-/// \p offset is the offset in bytes from the base address.
+
```
opti-mix: No empty lines between the doxygen comments and the code.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,12 +16,47 @@
...
 #define GLOW_BACKENDS_COMPILEDFUNCTION_H
 
+#include "glow/Graph/Nodes.h"
 #include <unordered_map>
 
 namespace glow {
 
 class Context;
+namespace runtime {
+/// Runtime Symbol Info
```
qcolombet: Nit: I fell this should be the class comment and need to be more explanatory.
http://llvm.org/docs/CodingStandards.html#commenting

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,12 +16,47 @@
...
+/// Contains Information for initialization and handling of symbol at runtime.
 
+struct RuntimeSymbolInfo {
+  /// \p size is the size in bytes of the symbol.
+  size_t size;
+
+  /// \p offset is the offset in bytes from the base address.
+  size_t offset;
```
qcolombet: More a remark than anything. With an offset we assume that the runtime has easily access to the base address and that it knows easily how to map a symbol to a base address.
For instance, let say I ask where is `foo`, we need to know somewhere that `foo` is a weight or activation to look at the right table.
More generally, again just future looking remarks, if we start to share weights between different graphs (as in not different instances of the same graph, but really different graphs[1]), we may want to create more fine tuned pool of symbols and we would need to somehow keep track of which symbol belongs to which pool.
The bottom line is moving forward, I wonder if this should be a full address instead of just an offset (or a base + offset in case we want to support relocating things)

[1] This could easily happen if we were to split a graph in the middle. E.g.,
Split
```
Weights
  |   \
  V    \
 op     \
  |     |
  V     |
 cpu Op |
  \     |
   \   /
    V V
    op
```
Into
```
Weights   [shared between G1 & G3]
--|---\---- vv G1 on accelerator A
  V    \
 op     \
--|-----|--- vv G2 on CPU
  V     |
 cpu Op |
  \     |
---\---/--- vv G3 on accelerator A
    V V
    op
```

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,23 +16,50 @@
...
+  /// The size in bytes.
+  size_t size;
+  /// Offset in bytes from the base address.
+  size_t offset;
+};
+/// Runtime Bundle
+/// Contains the information needed to be passed forward from compile time to
+/// runtime. In order to allocate and initialize memory.
+struct RuntimeBundle {
+  /// Map from symbol name to a RuntimeSymbolObject.
```
qcolombet: RuntimeSymbolObject -> RuntimeSymbolInfo

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,23 +16,50 @@
...
+  std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable;
+  /// Pointer to memory containing the weights for execution.
+  uint8_t *constants;
+  /// Amount of memory needed for weights.
+  const size_t constantWeightVarsMemSize;
+  /// Mmount of memory needed for mutable vars.
```
qcolombet: Mmount -> Amount

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,23 +16,50 @@
...
+  const size_t mutableWeightVarsMemSize;
+  /// Amount of memory needed for activations.
+  const size_t activationsMemSize;
```
qcolombet: Just checking, what is the reason we split mutableWeight and activations? Is to do something special for training?
Otherwise and I don't really see why we would need to track them differently.

include/glow/Backends/CompiledFunction.h
```diff
@@ -16,23 +16,50 @@
...
+  RuntimeBundle(size_t constWeight, size_t mutableWeight, size_t activations)
+      : constantWeightVarsMemSize(constWeight),
+        mutableWeightVarsMemSize(mutableWeight),
+        activationsMemSize(activations) {}
```
qcolombet: It feels weird that this API doesn't have anything to manage the symbolTable and the constants array.
It feels weird as well to have the size for the various section but the pointer only for the constants. Could you elaborate on how to you see this being used?

lib/Backends/CPU/AllocationsInfo.h
```diff
@@ -16,6 +16,8 @@
 #ifndef GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 #define GLOW_BACKENDS_CPU_ALLOCATIONSINFO_H
 
+// #include "glow/Backends/CompiledFunction.h"
```
qcolombet: Unused code ;)

lib/Backends/CPU/CPUBackend.h
```diff
@@ -47,6 +47,9 @@ class CPUBackend : public BackendUsingGlowIR {
 
   std::unique_ptr<CompiledFunction> compile(Function *F,
                                             const Context &ctx) const override;
+  /// Loads the function binary and constants onto device.
+  /// For CPU case this is primarily just allocating constant memory.
+  void loadFunctionOnDevice(const Context &ctx);
```
qcolombet: I believe this API should be in a different class. Like DeviceManager that each backend needs to implement.
Basically, a backend should come with different and separate APIs, at least two, compile and runtime.

include/glow/Backends/CompiledFunction.h
gcatron: These fields could be passed to the CompiledFunction directly but a struct gives us flexibility going forward to support different backend needs.

lib/Backends/CPU/CPUBackend.h
gcatron: So that particular function was meant to be deleted :P But! I agree I created #1951 to track designing and creating an API for runtime.

include/glow/Backends/CompiledFunction.h
qcolombet: What I am trying to get at is from a design perspective, one could mess with the constants pointer without doing the proper changes to the weight size member. That's why I am surprised it is not encapsulate behind some APIs. I.e., we expect the users of this API to do the right thing and that's when I get suspicious :).

Now, from a more high level point of view, the handling of constants and mutable allocations are the same. The only difference is that we know the contain of the constants at compile time. Therefore what I am questioning is whether or not this special handling of constants makes sense.

include/glow/Backends/CompiledFunction.h
gcatron: Summarizing the conversations between @opti-mix  and @qcolombet. I think adding a "kind" value to symbolInfo that allows interchange between offset and absolute memory addressing is a great idea, but not needed for this PR. We will revisit when we looks at expanding the runtime API perhaps in #1951 .

include/glow/Backends/CompiledFunction.h
gcatron: Due to the difference in the observable nature of Mutable vs activations. Mutable is user observable and may be placed in a different region of memory compared to activations.

qcolombet: Sounds good. We should clarify this in the comments.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,80 @@
...
+#include "glow/Graph/Context.h"
 #include "glow/Support/Compiler.h"
 #include "glow/Support/Memory.h"
 
 using namespace glow;
 
-CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT, void *heap)
-    : JIT_(std::move(JIT)), heap_(heap) {}
+CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
+                         runtime::RuntimeBundle runtimeBundle)
```
rdzhabarov: you can pass by const reference here.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,80 @@
...
+  copyMutablesToDevice(ctx);
+
   auto sym = JIT_->findSymbol("jitmain");
   assert(sym && "Unable to JIT the code!");
-  using JitFuncType = void (*)(void);
+  using JitFuncType =
+      void (*)(uint8_t * constantWeightVars, uint8_t * mutableWeightVars,
+               uint8_t * activations);
   auto address = sym.getAddress();
   if (address) {
```
rdzhabarov: I'm curious what is the situation when sym.getAddress() is null? Could it ever happen?

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,80 @@
...
+    auto payload = baseMutableWeightVarsAddress_ + symbolInfo->second.offset;
+    auto numBytes = symbolInfo->second.size;
+    auto addr = PH.second->getUnsafePtr();
+    // copy PH from allocated memory.
+    memcpy(addr, payload, numBytes);
+  }
+}
+
+void CPUFunction::freeAllocations() {
+  free(baseMutableWeightVarsAddress_);
```
rdzhabarov: must be alignedFree! :-)

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,80 @@
...
+  free(baseActivationsAddress_);
```
rdzhabarov: ditto

rdzhabarov:
Have you verified that AOT still works :)?

lib/Backends/CPU/CPUFunction.cpp
gcatron: > Have you verified that AOT still works :)?

Testing now, it has one core at 100% and has for the last 5 minutes....

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -41,16 +41,17 @@ namespace {
...
-/// Emit the entry point for JIT called "jitmain". It simply calls the main
-/// entry of the module with the constant concrete addresses of all the memory
-/// areas. Since these addresses are constants, the LLVM optimizer will constant
-/// propagate them into relative addressing computations and the like and
-/// produce a very efficient code that uses absolute addressing whenever
-/// possible.
+/// Emit the entry point for JIT called "jitmain".
+/// Function has the following API:
+///   void jitmain(uint8_t *baseConstantWeightVars, uint8_t
+///   *baseInoutWeightVars,
```
rdzhabarov: nit: baseIn"Out"

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -13,28 +13,80 @@
...
+  baseActivationsAddress_ = (uint8_t *)alignedAlloc(
+      runtimeBundle_.activationsMemSize, TensorAlignment);
+
+  baseMutableWeightVarsAddress_ = (uint8_t *)alignedAlloc(
+      runtimeBundle_.mutableWeightVarsMemSize, TensorAlignment);
+}
+
+void CPUFunction::copyMutablesToDevice(Context &ctx) {
+  // Copy Placeholders into allocated memory.
+  for (auto PH : ctx.pairs()) {
```
rdzhabarov: placeholders contain both inputs and outputs. Do we need to copy outputs? What's your plan here?

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,21 +27,35 @@ class CPUFunction final : public CompiledFunction {
...
-
+  CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
+              const runtime::RuntimeBundle &runtimeBundle);
+  /// Copy Function to device, an empty function for CPU.
+  void copyFunctionToDevice(){};
+  /// Copy Constants to device, an empty function for CPU.
+  void copyConstantsToDevice(){};
+  /// Allocate Mutable buffers on device, Activations and Placeholders.
+  void allocateMutableBuffersOnDevice();
+  /// Copy Mutable data to device.
```
rdzhabarov: what's actually "mutable data"?

It's clear what's constants, activations, placeholders.

gcatron: I don't know if we currently have a good way to filter this from the context. For this PR I suggest we copy all. I've created a separate issue #1953 to track how we'd like to pass inputs/outputs going forward.

lib/Backends/CPU/CPUFunction.cpp
gcatron: I don't know if we currently have a good way to filter this from the context. For this PR I suggest we copy all. I've created a separate issue #1953 to track how we'd like to pass inputs/outputs going forward.

lib/Backends/CPU/CPUFunction.cpp
opti-mix: It is very unlikely to happen, but we can add an assert here just for the case.

nadavrot:
Congrats! Nice work.

PR #1946, opened 2018-10-31
[docs] add some collaboration and compiler terminology

compnerd:
*Description*:

These terms can be confusing for those that do not have a background in
compilers.  Some of the collaboration terminology is derived from LLVM, and that
requires a specific background as well.

*Documentation*:
This PR covers docs.

### DIFF
--- a/docs/Lexicon.md
+++ b/docs/Lexicon.md
@@ -9,6 +9,18 @@
   A computational framework based on a synthetic construction of a the
   biological neural network in brains.
 
+* AOT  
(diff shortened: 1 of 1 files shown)


SplitInfinity:
These are just acronym expansions. Could you add some more detail? E.g.

WIP
Work In Progress
If your pull request is not ready to be considered for merging, label it with [WIP] and the work in progress tag. Common use cases of this are getting some initial feedback on a pull request and seeing if continuous integration tests pass.

PR #1955, opened 2018-11-05
[docs] Update docs/Lexicon.md

jackm321:
*Description*:
* define DMA
* fix header for "C" to be same size as other headers

*Testing*:
* look at markdown rendering in https://github.com/jackm321/glow/blob/update_lexicon/docs/Lexicon.md

*Documentation*:
* update docs/Lexicon.md

### DIFF
--- a/docs/Lexicon.md
+++ b/docs/Lexicon.md
@@ -21,7 +21,7 @@
   A representation of the instruction stream encoded to be efficiently
   interpreted by an executor.
 
-#### C
+### C
 
 * CNN  
   Convolutional Neural Network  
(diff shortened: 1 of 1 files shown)


docs/Lexicon.md
```diff
@@ -36,6 +36,10 @@
 * DAG  
   Directed Acylic Graph  
 
+* DMA  
+  Direct Memory Access
+  Copy data to/from memory without occupying CPU time
```
rdzhabarov: if you look at rendering that's on the same line as "Direct Memory Access", looks a bit messy. Can you fix this.

jackm321: fixed

PR #1956, opened 2018-11-06
[fix] Fix memory leak when allocate Tensor object.

rdzhabarov:
*Description*:
We allocate memory for the Tensor object but do not free it. Fix this.
Note, underlying tensor data is unowned by tensor so we should be fine to delete tensor.

From the internal tool:
```
==87496==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 1 object(s) allocated from:
     #0 libtools_build_sanitizers_asan-ubsan-py.so+0x29708 operator new(unsigned long)
     #1 glow/glow/lib/Onnxifi/Base.cpp:106       glow::onnxifi::Graph::run(llvm::DenseMap<...> const&, llvm::DenseMap<...> const&)
```

*Testing*:
onnxifi_driver tests.

*Documentation*:
N/A

cc: @yinghai

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -112,6 +112,12 @@ void Graph::run(
   updateInputPlaceholders(ctx_, phs, tensors);
   EE.run(ctx_);
 
+  // Tensors do not own underlying memory for input buffer,
+  // just delete memory allocated for the tensor object itself.
+  for (size_t i = 0; i < tensors.size(); ++i) {
+    delete tensors[i];
+  }
+
   // Copy outputs to the addresses specified in the outputPlaceholderToBuffer.
   for (auto outputVar : outputPlaceholderToBuffer) {
     void *outputAddress = reinterpret_cast<void *>(outputVar.second);


rdzhabarov:
cc: @zrphercule just putting on your radar. We'd probably found this bug if onnxifi_driver_test was executed in ASAN mode.

zrphercule:
@rdzhabarov wow great job! We didnt run onnxifi_driver_test in ASAN mode here is because of some strange memory leaking problem. Wonder if it is possible that this bug caused that? Thanks!

nadavrot:
Nice!

rdzhabarov:
> Wonder if it is possible that this bug caused that? Thanks!

Not sure. Would it be hard to rewire things for onnxifi_driver_test on asan build.

zrphercule:
@rdzhabarov @yinghai 
When we tried to enable onnxifi_driver_test on asan build in circle CI, some strange error occurred. We cannot found how to solve it, and then disabled onnxifi_driver_test on asan build:

https://github.com/pytorch/glow/blob/master/.circleci/test.sh#L59
https://github.com/pytorch/glow/blob/master/.circleci/build.sh#L54

Make a pr to enable asan test should be very easy. Should I make one to test if it works now?

yinghai:
Yeah, we can try.

PR #1957, opened 2018-11-06
[trivial] Add RAUW (Replace All Uses With) to the lexicon

bertmaher:

### DIFF
--- a/docs/Lexicon.md
+++ b/docs/Lexicon.md
@@ -102,6 +102,11 @@
 
 ### R
 
+* RAUW  
+  Replace All Uses With  
+  A term commonly used in LLVM referring to a method that replaces all the uses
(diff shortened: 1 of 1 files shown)


docs/Lexicon.md
```diff
@@ -102,6 +102,11 @@
 
 ### R
 
+* RAUW
+  Replace All Uses With
+  A term commonly used in LLVM referring to a method that replaces all the uses
```
rdzhabarov: check "view" for this change ;)

bertmaher: wtf?  what's different between this and all the other defs below it?

bertmaher: Omg.  Is it seriously the two spaces after "Replace All Uses With"?  Here I thought I liked Markdown.

PR #1958, opened 2018-11-06
Moved OpenCL memory allocation to runtime.

gcatron:
*Description*: Move OpenCL memory allocation to runtime. This is in support of #1904. 
*Testing*: Ran ninja test all pass. run.sh --iterations=10 -opencl -time. Accuracy is the same as master now. Time is almost identical for all except resnet 50. 
Master:
Wall time per iteration (s): 0.1335
   1.1925 (100.0%)   0.1420 (100.0%)   1.3345 (100.0%)   1.3350 (100.0%)  Infer
   1.1925 (100.0%)   0.1420 (100.0%)   1.3345 (100.0%)   1.3350 (100.0%)  Total
Wall time per iteration (s): 0.4013
Current Branch:
Wall time per iteration (s): 0.1628
   0.8187 (100.0%)   0.1483 (100.0%)   0.9670 (100.0%)   1.6282 (100.0%)  Infer
   0.8187 (100.0%)   0.1483 (100.0%)   0.9670 (100.0%)   1.6282 (100.0%)  Total
Wall time per iteration (s): 0.7370

*Documentation*: N/A
Progress on #1904

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_BACKENDUTILS_H
+#define GLOW_BACKENDS_BACKENDUTILS_H
+
+#include "glow/Backends/CompiledFunction.h"
+#include "glow/IR/IR.h"
+
+namespace glow {
+/// At compile time condense constants to a single block of memory.
+/// This allows the graph to go away after compile time.
+/// Allocates a block of memory of size \p constantMaxSize then walks the given
+/// function \p F and and copies weights to their address as specified by
+/// offsets contained in \p symbolTable.
+uint8_t *collectConstants(
+    const IRFunction *F, uint64_t constantMaxSize,
+    const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+        &symbolTable);
+/// Helper function to retrieve offset for Value: \p v from \p symbolTable.
+size_t
+getValueOffset(Value *v,
+               const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+                   &symbolTable);
+} // end namespace glow
+
+#endif // GLOW_BACKENDS_BACKENDUTILS_H
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "glow/Backends/BackendUtils.h"
+#include "glow/IR/Instrs.h"
+
+using namespace glow;
+using llvm::cast;
+using llvm::isa;
+
+uint8_t *glow::collectConstants(
+    const IRFunction *F, uint64_t constantMaxSize,
+    const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+        &symbolTable) {
+
+  // At compile time condense constants to a single block of memory.
+  // This allows the graph to go away after compile time.
+  uint8_t *baseConstantWeightVarsStore =
+      (uint8_t *)alignedAlloc(constantMaxSize, TensorAlignment);
+  for (auto &v : F->getGraph()->getParent()->getConstants()) {
+    assert(isa<WeightVar>(F->getWeightForNode(v)));
+    auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    auto payload = v->getPayload().getUnsafePtr();
+    auto numBytes = w->getSizeInBytes();
+    auto it = symbolTable.find(std::string(w->getName()));
+    assert(it != symbolTable.end() && "Symbol not found.");
+    auto addr = it->second.offset;
+    // Copy weight to offset.
+    memcpy(baseConstantWeightVarsStore + addr, payload, numBytes);
+  }
+  return baseConstantWeightVarsStore;
+}
+
+/// Helper function, gets offset of \p v from \p symbolTable.
+size_t glow::getValueOffset(
+    Value *v, const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+                  &symbolTable) {
+  auto it = symbolTable.find(std::string(v->getName()));
+  assert(it != symbolTable.end() && "Symbol not found.");
+  return it->second.offset;
+}
--- a/lib/Backends/CMakeLists.txt
+++ b/lib/Backends/CMakeLists.txt
@@ -1,5 +1,11 @@
 add_library(Backends Backends.cpp)
 
+add_library(BackendUtils BackendUtils.cpp)
+
+target_link_libraries(BackendUtils
+                      PRIVATE
+                      IR)
+
 add_subdirectory(Interpreter)
 
 if(GLOW_WITH_OPENCL)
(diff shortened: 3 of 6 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1452,141 +1441,181 @@ uint64_t OpenCLFunction::copyValueFromDevice(const Value *v, void *buf) {
...
-    if (!externalTensors_.count(it.first)) {
-      continue;
-    }
-    if (auto *W = dyn_cast<WeightVar>(it.first)) {
-      if (W->getMutability() == WeightVar::MutabilityKind::Constant)
-        continue;
-    }
-    copiedBytes += copyValueToDevice(it.first);
+void OpenCLFunction::copyConstantsToDevice() {
+  deviceBuffer_ = allocDeviceBuffer(bundle_.constantWeightVarsMemSize +
```
opti-mix: You allocate all of the required memory in copyConstantsToDevice, including constants, mutable weights and activations. It seems wrong. The buffer should be allocated before this function is called.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1452,141 +1441,181 @@ uint64_t OpenCLFunction::copyValueFromDevice(const Value *v, void *buf) {
...
+  // Do it!
   clFinish(commands_);
-  return copiedBytes;
 }
-
-void OpenCLFunction::allocateMemory(const Context &ctx) {
-  // The allocator assigns device memory addresses to the buffers.
+/// Collect Contstants from graph and store them for passing to execution.
+uint8_t *collectConstants(
+    const IRFunction *F, uint64_t constantMaxSize,
```
opti-mix: IIRC, you also have introduced a similar `collectConstants` in the CPU backend. May be you should create a utility function and call it from both backends?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1452,141 +1441,181 @@ uint64_t OpenCLFunction::copyValueFromDevice(const Value *v, void *buf) {
...
+    // Copy weight to offset.
+    memcpy(baseConstantWeightVarsStore + addr, payload, numBytes);
+  }
+  return baseConstantWeightVarsStore;
+}
+/// Computes offsets and total allocation for Constants, Placeholders, and
+/// Activations to build runtime symbol table. Returns
+/// RuntimeBundle.
+runtime::RuntimeBundle generateRuntimeBundle(const IRFunction *F) {
+  // Use a single allocator. The OpenCL backend uses a single buffer on the card
```
opti-mix: IIRC, you also have introduced a similar `generateRuntimeBundle` in the CPU backend. May be you should create a utility function and call it from both backends?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1055,12 +1055,15 @@ void OpenCLFunction::execute(Context &ctx) {
...
       setKernelArg(kernel, numArgs + 7, filterGradDim);
       // Zero memory for the output buffers.
-      fillBuffer(deviceBuffer_, tensors_[srcGrad], srcGrad->size(), 0,
-                 srcGrad->getElementType());
-      fillBuffer(deviceBuffer_, tensors_[filterGrad], filterGrad->size(), 0,
-                 filterGrad->getElementType());
-      fillBuffer(deviceBuffer_, tensors_[biasGrad], biasGrad->size(), 0,
-                 biasGrad->getElementType());
+      fillBuffer(deviceBuffer_,
+                 bundle_.symbolTable[std::string(srcGrad->getName())].offset,
```
opti-mix: Is it really necessary to do `std::string(srcGrad->getName())`, i.e. construct a string explicitly? Even if this is the case, could you create a helper function `getValueOffset(Value *v)` which would return `bundle_.symbolTable[std::string(v->getName())].offset`?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -619,11 +622,8 @@ static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
   }
 }
 void OpenCLFunction::execute(Context &ctx) {
-  auto copiedToDeviceBytes = copyMutableWeightsToDevice();
-  (void)copiedToDeviceBytes;
-  DEBUG_GLOW(llvm::dbgs() << "Copied " << copiedToDeviceBytes
-                          << " bytes to OpenCL device\n");
-
+  copyConstantsToDevice();
```
opti-mix: It seems like it would copy the constant to the device every time, which is not good. Also, IIRC, you cannot run simultaneously two instances of `OpenCLFunction::execute` for the same IRFunction. Thus, I'd suggest allocating the memory on first call of the `execute` and reusing it afterwards. 

With this approach, you can copy the constants only once, which improves the performance a lot.

lib/Backends/OpenCL/OpenCL.cpp
gcatron: That function differs a fair bit between backends. For now I recommend keeping them separate.

lib/Backends/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 add_library(Backends Backends.cpp)
 
+add_library(BackendUtils BackendUtils.cpp)
+
+target_link_libraries(BackendUtils
+                      PRIVATE
+                      IR)
```
rdzhabarov: can you make sure there is no circular dependency introduced here?

you can run `ninja dependency_graph` and check .dot file.

include/glow/Backends/BackendUtils.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef GLOW_BACKENDS_BACKENDUTILS_H
+#define GLOW_BACKENDS_BACKENDUTILS_H
+
+#include "glow/Backends/CompiledFunction.h"
+#include "glow/IR/IR.h"
+
+namespace glow {
+uint8_t *collectConstants(
+    const IRFunction *F, uint64_t constantMaxSize,
+    std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable);
```
rdzhabarov: do not copy unordered_map here, pass by const ref.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -120,8 +121,8 @@ static void addStringOption(std::vector<std::string> &options,
 }
 
 OpenCLFunction::OpenCLFunction(std::unique_ptr<IRFunction> F,
-                               const Context &ctx)
-    : F_(std::move(F)) {
+                               runtime::RuntimeBundle bundle)
+    : F_(std::move(F)), bundle_(std::move(bundle)) {
```
rdzhabarov: RuntimeBundle does not have efficient move copy ctor, I'd pass by const ref here as well.

lib/Backends/CMakeLists.txt
gcatron: Checked the Dot file, No circular dependency introduced. IR (introduced by BackendUtils) does not depend on anything upstream.

lib/Backends/OpenCL/OpenCL.cpp
gcatron: Currently I think it's necessary to construct the string. A helper function is a good idea though. I can do offset. And/Or I could write a helper function findSymbol(Value *v) which returns an iterator. (it->second.offset).

rdzhabarov:
> Time is almost identical for all except resnet 50.

I suspect it has something to do with copying weights per inference onto the device.
What about vgg19 inference time?

Do you track anywhere migration plan for changing this eventually to copying weights once per inference?

gcatron:
> > Time is almost identical for all except resnet 50.
> 
> I suspect it has something to do with copying weights per inference onto the device.
> What about vgg19 inference time?
> 
> Do you track anywhere migration plan for changing this eventually to copying weights once per inference?

vgg19 inference times were very similar between master and current branch.  The migration is called out as a TODO in execution engine which points back to #1904 .

lib/Backends/OpenCL/OpenCL.cpp
gcatron: This will be changed as part of the ongoing work for #1904. A new api is called out there which has more appropriate names and division of work.

lib/Backends/OpenCL/OpenCL.cpp
gcatron: Created issue #1967 to track making RuntimeBundle an object with a cleaner interface.

lib/Backends/BackendUtils.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/// Helper function, gets offset of \p v from \p symbolTable.
+size_t glow::getValueOffset(
+    Value *v, const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+                  &symbolTable) {
+  auto it = symbolTable.find(std::string(v->getName()));
+  assert(it != symbolTable.end() && "Symbol not found.");
+  return it->second.offset;
+}
```
opti-mix: All files should have a newline at the end. The red marker above shown by GH indicates that this is not the case.

include/glow/Backends/BackendUtils.h
```diff
@@ -0,0 +1,39 @@
...
+    const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+        &symbolTable);
+/// Helper function to retrieve offset for Value: \p v from \p symbolTable.
+size_t
+getValueOffset(Value *v,
+               const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
+                   &symbolTable);
+} // end namespace glow
+
+#endif // GLOW_BACKENDS_BACKENDUTILS_H
```
opti-mix: Add a newline

lib/Backends/OpenCL/OpenCL.cpp
opti-mix: Can you still add the comment here about the plans to move this allocation out of this function in the future?

opti-mix:
@gcatron And do not forget to squash the commits if needed.

PR #1959, opened 2018-11-06
[Quantization] Enabling quantization profiling in CPU bakend.

ayermolo:
*Description*:
Added quantization profiling to CPU backend. Basically ported from interpreter engine.

*Testing*:
Expanded Quantization test in unittests.

*Documentation*:
N/A

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1410,4 +1410,124 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     break;
   }
 }
+
+static void find_min_max_f(float *tensor, size_t size, float &min, float &max) {
+  min = tensor[0];
+  max = tensor[0];
+
+  for (size_t i = 1; i < size; ++i) {
+    float tensorVal = tensor[i];
+    if (tensorVal < min)
+      min = tensorVal;
+
+    if (tensorVal > max)
+      max = tensorVal;
+  }
+}
+
+static int check_all_zeros(float *arrayToCheck, size_t size) {
+  for (int i = 0; i < size; ++i) {
+    if (arrayToCheck[i] != 0) {
+      return 0;
+    }
+  }
+  return 1;
+}
+
+/// Gen a bin number to insert \p value into the histogram which has \p nBins
+/// with \p minValue and binWidth in histogram.
+static size_t get_bin(size_t nBins, float binWidth, float minValue,
+                      float value) {
+  size_t result =
+      binWidth == 0
+          ? 0
+          : MIN(static_cast<size_t>((value - minValue) / binWidth), nBins - 1);
+  return result;
+}
+
+// code ported from Profile.cpp: generateTensorHistogram
+__attribute__((noinline)) void
+libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
+                            size_t numDimsTensor, float *compInfo,
+                            float *existingHistogram, size_t *histDim) {
+  float minInput = inputTensor[0];
+  float maxInput = inputTensor[0];
+
+  size_t tensorSize = 1;
+  // calculating total size of the Tensor.
+  for (size_t i = 0; i < numDimsTensor; ++i) {
+    tensorSize *= tensorDim[i];
+  }
+
+  size_t nBins = histDim[0];
+
+  // finding min/max value for entire tensor
+  find_min_max_f(inputTensor, tensorSize, minInput, maxInput);
+
+  float min = compInfo[0];
+  float max = compInfo[1];
+  if (check_all_zeros(existingHistogram, nBins) == 1) {
+    compInfo[0] = minInput;
+    compInfo[1] = maxInput;
+
+    min = minInput;
+    max = maxInput;
+  }
+
+  // Check if we need to rescale histogram.
+  if (minInput < min || maxInput > max) {
+    float newMin = MIN(minInput, min);
+    float newMax = MAX(maxInput, max);
+
+    float destBinWidth = (newMax - newMin) / nBins;
+    float srcBinWidth = (max - min) / nBins;
+    float scaledHistogram[nBins];
+
+    for (size_t i = 0; i < nBins; ++i) {
+      if (existingHistogram[i] == 0)
+        continue;
+
+      float srcBinBegin = min + srcBinWidth * i;
+      size_t destBin = (srcBinBegin - newMin) / destBinWidth;
+      float destBinEnd = newMin + destBinWidth * (destBin + 1);
+
+      float srcBinEnd = srcBinBegin + srcBinWidth;
+      size_t destBinToVerify = (srcBinEnd - newMin) / destBinWidth;
+      // Make sure that destination bin is mapped at most to 2 final bins, based
+      // on that redistribute percentage is calculated.
+      assert(destBinToVerify <= destBin + 2);
+      (void)destBinToVerify;
+
+      // Calculate how much we need to redistribute.
+      uint64_t dstBinCnt = static_cast<uint64_t>(
+          MIN(static_cast<float>(round((destBinEnd - srcBinBegin) /
+                                       srcBinWidth * existingHistogram[i])),
+              existingHistogram[i]));
+
+      size_t newBin = get_bin(nBins, destBinWidth, newMin, srcBinBegin);
+      scaledHistogram[newBin] += dstBinCnt;
+
+      if (dstBinCnt < existingHistogram[i]) {
+        size_t newBin =
+            get_bin(nBins, destBinWidth, newMin, srcBinBegin + destBinWidth);
+        scaledHistogram[newBin] += existingHistogram[i] - dstBinCnt;
+      }
+    }
+
+    // Copy scaled histogram back to the existing histogram.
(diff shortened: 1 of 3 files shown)


rdzhabarov:
Meta question, do you plan to use histogram distribution for calculating scale/offset in better than naive way we are doing now? (note, currently we just use min/max for scale/offset calculations even though technically we capture the whole histogram)

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1392,4 +1392,123 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     break;
   }
 }
+
+void find_min_max_f(float *tensor, size_t size, float &min, float &max) {
+  for (size_t i = 1; i < size; ++i) {
```
jfix71: nit: I don't like how this function assumes min/max have already been set to `tensor[0]`. Can you do it here at the start of the function?

ayermolo:
Not at the moment.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1392,4 +1392,123 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     break;
   }
 }
+
+void find_min_max_f(float *tensor, size_t size, float &min, float &max) {
```
jfix71: nit: make this `static`

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1392,4 +1392,123 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
...
+    float tensorVal = tensor[i];
+    if (tensorVal < min)
+      min = tensorVal;
+
+    if (tensorVal > max)
+      max = tensorVal;
+  }
+}
+
+int check_all_zeros(float *arrayToCheck, size_t size) {
```
jfix71: nit: make this `static`

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1410,4 +1410,125 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     break;
   }
 }
+
+static void find_min_max_f(float *tensor, size_t size, float &min, float &max) {
```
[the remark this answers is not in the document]
ayermolo: yep, already did on AlexNet, images/imagenet-227/{zibra, dog}
Two nodes had difference, compared to interpreter, of .00001 and .0000001. Seems like difference was due to input values from Tensor being slightly different.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1410,4 +1410,125 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
...
+    if (tensorVal < min)
+      min = tensorVal;
+
+    if (tensorVal > max)
+      max = tensorVal;
+  }
+}
+
+static int check_all_zeros(float *arrayToCheck, size_t size) {
+  int are_all_zero = 1;
```
jfix71: nit: This is unused.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1410,4 +1410,125 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
...
+    float tensorVal = tensor[i];
+    if (tensorVal < min)
+      min = tensorVal;
+
+    if (tensorVal > max)
+      max = tensorVal;
+  }
+}
+
+static int check_all_zeros(float *arrayToCheck, size_t size) {
```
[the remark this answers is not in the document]
MattPD: BTW, you can avoid loads/stores in the loop (corresponding to reads/writes of arguments `min` and `max` passed by reference) by keeping the running min & max in local variables (and only writing at the end) -- as in the `VERSION_LOCAL` example in https://godbolt.org/z/JuzDsI (compare with `VERSION_BASELINE`; cf. also `VERSION_MINMAX` using `std::minmax_element` if using std. lib. algorithms is okay).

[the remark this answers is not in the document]
rdzhabarov: @ayermolo nvm, that's libjit which has different convention...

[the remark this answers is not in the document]
ayermolo: @MattPD Thank you for your feedback. Is such a low level optimization really necessary? Reason I say this is that glow/llvm will inline this min_max function. It will then specialize libjit_quantization_profile. Looking at LLVM IR it will also do unroll and vectorization.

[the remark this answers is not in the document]
MattPD: @ayermolo Good point, not necessarily! If it's inlined into `libjit_quantization_profile` I wouldn't expect aliasing to be a problem (there are distinct local variables `minInput` and `maxInput` there, so no such aliasing is possible). One can even obtain the same generated code -- depending on the compiler & optimization level combination: https://godbolt.org/z/tURXvl vs https://godbolt.org/z/0kzSLR (the approach to branching can yield different results, though: https://godbolt.org/z/h7ykqj). Perhaps `std::minmax_element` could be the simplest choice, then (if using `<algorithm>` is okay for this).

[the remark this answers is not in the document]
ayermolo: @rdzhabarov is using std libraries OK here?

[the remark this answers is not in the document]
opti-mix: The libjit is first compiled into a bitcode and then this bitcode is included into the LLVM modules produced from a Glow IR of your NN model. Then it is compiled by means of LLVM's ORC JIT and executed. 

Thus any references to external libraries need to be resolved by the ORC JIT "linking" layer and in general we try to limit the amount of references to external functions if possible. 

That being said, references to `libc` functions are usually fine, as `libc` is typically present on any system. References to C++ libs could be more problematic, but on the other hand libc++ is also present on almost all systems these days. And if the std functions are just template functions that are expanded into a self-contained code, it should not be a problem. 

To check if it is OK to use std C++ functions, just try to use them in libjit and then try to run the NN model using the `-cpu` backend and see if it works or if you get some errors at runtime, because ORC JIT cannot find the referenced symbol. It is also important to check that the AOY functionality is not broken by such a change. To check this, try to create a bundle from your NN model and then link your app against the produced `*.o` file. It should not result in any linking errors.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1410,4 +1410,125 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
...
+                     float value) {
+  size_t result =
+      binWidth == 0
+          ? 0
+          : MIN(static_cast<size_t>((value - minValue) / binWidth), nBins - 1);
+  return result;
+}
+
+// code ported from Profile.cpp: generateTensorHistogram
+__attribute__((noinline)) void
```
opti-mix: Nit: `__attribute__((noinline))` is not guaranteed to have any effect at the moment (see https://github.com/pytorch/glow/blob/master/lib/Backends/CPU/Pipeline.cpp#L79-L83 for more details)

ayermolo: Thanks for the link. I was following the definitions of other libjit functions in this file. Should I remove the attribute, or keep it for consistency with other functions?

opti-mix: @ayermolo You can keep it for consistency with other functions.

qcolombet:
This PR breaks the tests when the openCL backend is enabled.
Now the end2end tests don't work anymore.
```
[ RUN      ] OpenCL/Operator.end2end/0
Cannot select: quantizationprofile
../lib/Backends/OpenCL/OpenCL.cpp:1377: compilation failed
```

I believe it is useful to have a end2end test for opencl, but profiling is not that useful there. Thus, we probably want to duplicate the test one with profiling one w/o.
What do people think?

qcolombet:
BTW, it would be great if our CI would catch those problems!

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1410,4 +1410,124 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
...
+  }
+
+  // Check if we need to rescale histogram.
+  if (minInput < min || maxInput > max) {
+    float newMin = MIN(minInput, min);
+    float newMax = MAX(maxInput, max);
+
+    float destBinWidth = (newMax - newMin) / nBins;
+    float srcBinWidth = (max - min) / nBins;
+    float scaledHistogram[nBins];
```
yinghai: Can we initialize this with zeros?

rdzhabarov:
fixes and refinements here https://github.com/pytorch/glow/pull/1996

rdzhabarov:
OpenCl tests have been long on our radar, but given that we do not really invest in this backend it's been low priority.

ayermolo:
@rdzhabarov Thank you for fixes and improvements.

PR #1969, opened 2018-11-07
[fix] Do not allocate memory when size is 0 for CPU backend.

rdzhabarov:
*Description*:
In cases when we need not allocate memory we just bail.

```
grep -ir 'alignedAlloc' lib/
lib//Backends/CPU/AllocationsInfo.cpp:      (uint8_t *)alignedAlloc(constantWeightVarsMemSize_, TensorAlignment);
lib//Backends/CPU/CPUFunction.cpp:    baseActivationsAddress_ = (uint8_t *)alignedAlloc(
lib//Backends/CPU/CPUFunction.cpp:    baseMutableWeightVarsAddress_ = (uint8_t *)alignedAlloc(
```
This should be enough to get going.

*Testing*:
Unit tests + fbcode internal tests.

*Documentation*:
n/a

cc: @yinghai

### DIFF
--- a/lib/Backends/CPU/CPUFunction.cpp
+++ b/lib/Backends/CPU/CPUFunction.cpp
@@ -28,11 +28,15 @@ CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
 CPUFunction::~CPUFunction() { alignedFree(runtimeBundle_.constants); }
 
 void CPUFunction::allocateMutableBuffersOnDevice() {
-  baseActivationsAddress_ = (uint8_t *)alignedAlloc(
-      runtimeBundle_.activationsMemSize, TensorAlignment);
+  if (runtimeBundle_.activationsMemSize != 0) {
+    baseActivationsAddress_ = (uint8_t *)alignedAlloc(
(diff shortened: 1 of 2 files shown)


lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -64,8 +68,13 @@ void CPUFunction::copyOutputsFromDevice(Context &ctx) {
 }
 
 void CPUFunction::freeAllocations() {
-  alignedFree(baseMutableWeightVarsAddress_);
-  alignedFree(baseActivationsAddress_);
+  if (baseMutableWeightVarsAddress_) {
+    alignedFree(baseMutableWeightVarsAddress_);
```
bertmaher: It's safe to call `free` with null... but maybe alignedFree is different.  Not that it matters too much :)

rdzhabarov:
> Are them the only call sites?

See above grep in the description.

There are some not covered, but those would not matter if you search for all occurrences (it's mostly AOT bundles and what not).

lib/Backends/CPU/CPUFunction.cpp
rdzhabarov: I'd not risk, overhead seems pretty low here :)

PR #1974, opened 2018-11-07
[Tensor] Make dumpImpl(Tensor*) take a const ptr

### LINKED ISSUE #1968: Make dumpImpl(Tensor*) take a const Tensor*
Printing a Tensor shouldn't change its contents.  I think it currently takes a mutable pointer because it creates a Handle, and historically we couldn't handle const Handles.  I suspect the situation is better now so let's try to const-ify dumpImpl.

jackm321:
*Description*:
Replace several instances of `Tensor *` with `const Tensor *` in Tensor.h and Tensor.cpp where taking non-const Tensor * was unnecessary.
*Testing*:
compile and run `ninja check`, all tests passed
Fixes #1968

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -427,8 +427,8 @@ private:
 void dumpAsciiImpl(Tensor *T, llvm::raw_ostream &os);
 void dumpAsciiImpl(Tensor *T);
 
-void dumpImpl(Tensor *T, llvm::raw_ostream &os);
-void dumpImpl(Tensor *T);
+void dumpImpl(const Tensor *T, llvm::raw_ostream &os);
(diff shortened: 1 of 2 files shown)


jackm321:
The only testing I've done is compile, any other tests I should do?

rdzhabarov:
> The only testing I've done is compile, any other tests I should do?

should be enough.

rdzhabarov:
@jackm321 can you fix dumpAsciiImpl as well?

rdzhabarov:
Can you update description? (since now quite a few things were changed accordingly as well).

nadavrot:
@jackm321 Thanks for doing this work. 

*compile and run "ninja check"

Issue #1976, opened 2018-11-07
Build tools for "cosimulation" of networks


bertmaher:
This task is mostly for my own benefit ;-)

I've found it pretty useful to check intermediate results of networks against various backends to understand where things might be going off the rails.  There are a few things that I've built in a hacky local branch that I want to land in the main repo:
- A utility for converting png format images to raw binary Tensors.
- A utility for dumping Tensors to raw binary files and reading them back in from same.
- A utility function to cut a Graph at an arbitrary named Node, so as to easily read out the value at that point.
- (Something I haven't built) The ability to replace a named node with an input Variable.

Taken together these would allow you to very easily write verification scripts that slice up a network, run on different backends, and compare the results.  Conceptually they're similar to DebugPrint instrumentation, except "print" can be a tricky concept for ASICs/FPGAs/emulators depending on where each project is in its bring-up stage.

opti-mix:
Looking forward to see this functionality in action!

Two comments/suggestions:

> A utility for dumping Tensors to raw binary files and reading them back in from same.

We probably need to take care that these tensors are always written using the same byte-order, i.e. little-endian or big-endian, so that they can be read across different processors/HW accelerators.

> A utility function to cut a Graph at an arbitrary named Node, so as to easily read out the value at that point.

It could be also useful to be able to cut a Graph at an arbitrary kind of node (e.g. at any Conv node), so as to easily read out the value at that point.

bertmaher:
>  take care that these tensors are always written using the same byte-order

Just as you wrote this I was writing a comment about how the I/O functions I've created don't handle endianness.  */me looks sheepish*

PR #1978, opened 2018-11-08
[base] Utility functions for reading/writing binary tensor blobs

bertmaher:
*Description*: Some very rough functions for doing binary I/O on Tensors.
*Testing*: New UtilsTest
*Documentation*: Doxygen!
Part of #1976

### DIFF
--- a/include/glow/Base/IO.h
+++ b/include/glow/Base/IO.h
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BASE_IO_H
+#define GLOW_BASE_IO_H
+
+#include "glow/Base/Tensor.h"
+
+#include "llvm/ADT/StringRef.h"
+
+namespace glow {
+
+/// Write Tensor data \p T to a file named \p filename.
+///
+/// NB: This function is primarily a debugging aid, not a serialization format.
+/// It stores only the binary data of the tensor (not the dimensions), and
+/// writes tensors assuming host endianness.
+void writeToFile(const Tensor &T, llvm::StringRef filename);
+
+/// Read Tensor data \p T from a file named \p filename.
+///
+/// NB: This function is primarily a debugging aid, not a serialization format.
+/// It reads only the binary data of the tensor (the dimensions must be known
+/// through other means), and reads tensors assuming host endianness.
+void readFromFile(Tensor &T, llvm::StringRef filename);
+
+} // namespace glow
+
+#endif // GLOW_BASE_IO_H
--- a/lib/Base/IO.cpp
+++ b/lib/Base/IO.cpp
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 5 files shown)


include/glow/Base/IO.h
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+
+namespace glow {
+
+/// Write Tensor data \p T to a file named \p filename.
+///
+/// NB: This function is primarily a debugging aid, not a serialization format.
+/// It stores only the binary data of the tensor (not the dimensions), and
+/// writes tensors assuming host endianness.
+void writeToFile(const Tensor &T, std::string filename);
```
rdzhabarov: const ref for filename

include/glow/Base/IO.h
```diff
@@ -0,0 +1,40 @@
...
+
+/// Read Tensor data \p T from a file named \p filename.
+///
+/// NB: This function is primarily a debugging aid, not a serialization format.
+/// It reads only the binary data of the tensor (the dimensions must be known
+/// through other means), and reads tensors assuming host endianness.
+void readFromFile(Tensor &T, std::string filename);
```
rdzhabarov: same little thing here.

lib/Base/IO.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+} // namespace
+
+namespace glow {
+
+void writeToFile(const Tensor &T, std::string filename) {
+  FILE *fp = fopen(filename.c_str(), "wb");
+  GLOW_ASSERT(fp);
+  fwrite(T.getUnsafePtr(), T.getType().getElementSize(), T.size(), fp);
```
rdzhabarov: just a random thought, maybe return success/failure from this method depending on how many bytes written (whether it match the expected number of bytes written).

lib/Base/IO.cpp
```diff
@@ -0,0 +1,52 @@
...
+  fclose(fp);
+}
+
+void readFromFile(Tensor &T, std::string filename) {
+  GLOW_ASSERT(T.getType().getSizeInBytes() == filesize(filename));
+  FILE *fp = fopen(filename.c_str(), "rb");
+  GLOW_ASSERT(fp);
+  fread(T.getUnsafePtr(), T.getType().getElementSize(), T.size(), fp);
```
rdzhabarov: same thing here as for write.

tests/unittests/UtilsTest.cpp
```diff
@@ -49,3 +50,14 @@ TEST(Utils, deterministicPRNG) {
     EXPECT_EQ(dist(genA), dist(genB));
   }
 }
+
+TEST(Utils, readWriteTensor) {
+  const char *file = "readWriteTensor.bin";
```
rdzhabarov: can you delete the file afterwards?

lib/Base/IO.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "glow/Base/IO.h"
+
+#include <cstdio>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+namespace {
+
+static size_t filesize(std::string filename) {
```
rdzhabarov: const ref.

nadavrot: +1 on StringRef. 

I think that LLVM has portable utilities that implement this but also compile on windows.

lib/Base/IO.cpp
```diff
@@ -0,0 +1,52 @@
...
+namespace glow {
+
+void writeToFile(const Tensor &T, std::string filename) {
+  FILE *fp = fopen(filename.c_str(), "wb");
+  GLOW_ASSERT(fp);
+  fwrite(T.getUnsafePtr(), T.getType().getElementSize(), T.size(), fp);
+  fclose(fp);
+}
+
+void readFromFile(Tensor &T, std::string filename) {
```
nadavrot: I wonder if LLVM has any utils that we can use here.

lib/Base/IO.cpp
bertmaher: I didn't want to use StringRef because it doesn't guarantee a null terminator :-/

bertmaher:
I  think I'm going to abandon this.  LLVM probably has better utilities than anything I  can cook up without spending a lot of time on it.

lib/Base/IO.cpp
bertmaher: I started looking and it seems like `MemoryBuffer` is the way to read files into memory in LLVM?  But I got a bit worried about going down a rabbit hole here since I'm only so familiar with LLVMs APIs :-/.  Happy to have someone who knows them better land some improvements.  `fread` is kind of mid-90s,  I know :-p

nadavrot: I think that the 3-line implementation that uses fread is probably better because it won't break when we upgrade to a new LLVM. 

Everything looks great. ship it.

PR #1980, opened 2018-11-08
[runtime] Moved context use to runtime for Interpreter Backend

gcatron:
[runtime] Moved context use to runtime for Interpreter Backend

*Description*: Migrated use of passed in context from compile to execute for Interpreter. This is work on #1904.
*Testing*: Ninja test all pass. run.sh --iterations=10 -interpreter -time looks good with no significant time difference.
*Documentation*: N/A
Progress on #1904

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -30,6 +30,8 @@ struct RuntimeSymbolInfo {
   size_t size;
   /// Offset in bytes from the base address.
   size_t offset;
+  /// Type of symbol.
+  Type type;
 };
 /// Runtime Bundle
 /// Contains the information needed to be passed forward from compile time to
--- a/lib/Backends/Interpreter/Interpreter.cpp
+++ b/lib/Backends/Interpreter/Interpreter.cpp
@@ -17,11 +17,34 @@
 #include "Interpreter.h"
 #include "InterpreterFunction.h"
 
+#include "glow/Backends/BackendUtils.h"
+#include "glow/CodeGen/MemoryAllocator.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Nodes.h"
 #include "glow/IR/IR.h"
 
 using namespace glow;
+
+runtime::RuntimeBundle generateInterpreterRuntimeBundle(const IRFunction *F) {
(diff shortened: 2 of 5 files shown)


lib/Backends/Interpreter/InterpreterFunction.h
```diff
@@ -54,6 +54,17 @@ class InterpreterFunction final : public CompiledFunction {
   /// \name CompiledFunction interface
   ///@{
   ~InterpreterFunction() override;
+  /// Does any needed initialization work for the Backend. Adds references to
+  /// constants in externalTensors_.
+  void setupRuns();
```
rdzhabarov: `override` for all these methods.

gcatron: I'd need to add them to the base class first right? That's planned for the next task. I'll align all 3 backends to this api and add it to the base class.

gcatron:
Also after looking at it, I agree with @opti-mix generateRuntimeBundle should be refactored too. I've added #1988 to track that work.

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -17,12 +17,35 @@
 #include "Interpreter.h"
 #include "InterpreterFunction.h"
 
+#include "glow/Backends/BackendUtils.h"
```
rdzhabarov: you add include, but you have not add target_link_libraries for Interpreter. BackendUtils needs to be added.

Also, once you merge this can you update fbcode part, I'll show you how to do that.

include/glow/Backends/CompiledFunction.h
```diff
@@ -30,6 +30,10 @@ struct RuntimeSymbolInfo {
   size_t size;
   /// Offset in bytes from the base address.
   size_t offset;
+  /// Kind of constant
+  ElemKind kind;
+  /// Type of constant.
+  TypeRef type;
```
rdzhabarov: TypeRef has ElemKind, why do you need explicit ElemKind?

lib/Backends/Interpreter/InterpreterFunction.h
rdzhabarov: The reason I assumed that those are already defined in the CompiledFunction, because
these methods are in the section of 
```
/// \name CompiledFunction interface
  ///@{
```

It's fine as we are going to fix this very soon.

PR #1981, opened 2018-11-08
[quantization][trivial] Put FunctionQuantizer in anonymous namespace

bertmaher:
*Description*: Really minor thing, I just like having function-local stuff in anonymous namespaces so it's obviously not used elsewhere.  If anyone dislikes this I'll just drop the PR :)
*Testing*: compile
*Documentation*: N/A

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -25,8 +25,10 @@
 
 using llvm::cast;
 
-namespace glow {
-namespace quantization {
+namespace {
+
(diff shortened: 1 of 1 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -23,10 +23,12 @@
 #include <unordered_set>
 #include <vector>
 
+using namespace glow;
```
rdzhabarov: can you put this inside anon namespace?

PR #1984, opened 2018-11-08
[easy] Clean up temporary files

### LINKED ISSUE #1982: [quantization] Clean up temporary files used for serialization in tests
There are some temporary files created for serialization of tensor node quantization info.
Temporary files which are created during the tests need to be removed.
Please, see `quantizationTest` `testSerialization` method.

bertmaher:
*Description*: We use `llvm::sys::fs::createTemporaryFile` but don't always clean up afterwards.  Fix it!
*Testing*: `ninja check`
*Documentation*: N/A
Fixes #1982

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -250,6 +250,7 @@ void Module::dumpDAG() {
   llvm::SmallString<64> dotPath;
   llvm::sys::fs::createTemporaryFile("dotty_graph_dump", "dot", dotPath);
   dumpDAG(dotPath);
+  llvm::sys::fs::remove(dotPath);
 }
 
 void Module::dumpDAG(llvm::StringRef dotFilename) {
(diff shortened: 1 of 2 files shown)


zakk0610:
Hi @bertmaher 
Why do you want to clean the dag file? 
User cannot get the dag file when calling the dumpDAG.

rdzhabarov:
@zakk0610 good point, it makes sense for quantization test as we do not want to leave leftovers after the test, but we should not kill file in dumpDAG...

Do you want to send quick PR on fixing it?

bertmaher:
GraphTest leaves temporary files around since it calls dumpDAG().  If you remove the cleanup here please migrate those tests to properly remove their temp files.

bertmaher:
In general I expected developers to use dumpDAG("some_meaningful_name.dot") but I don't mind if we move tempfile deletion out of dumpDAG().

PR #1986, opened 2018-11-08
Fix node usedef

### LINKED ISSUE #1985: Type error with Node and UseDef
`Node` inherits from `UseDef<Node, NodeUse>` https://github.com/pytorch/glow/blob/master/include/glow/Graph/Node.h#L45.
`NodeUse::hasUser()` calls `.get()` on all stored instances of the second type parameter (`NodeUse` in this case) and compares against a given pointer to an instance of the first type parameter (`Node` in this case) https://github.com/pytorch/glow/blob/master/include/glow/Graph/UseDef.h#L51-L53.
However, `NodesUse::get()` returns a `NodeHandle *` not a `Node *` https://github.com/pytorch/glow/blob/master/include/glow/Graph/NodeValue.h#L214 so this doesn't work. The reason this ins't a compile time failure is because `NodeUse::hasUser()` is simply never called for `Node`s right now.

jackm321:
*Description*:
Swap type parameters on UseDef and make the second one optionally different from the first.

*Testing*:
build and run ninja test, all tests pass

*Documentation*:
no additional documentation

Fixes #1985


jackm321:
Actually, this doesn't really fix the inconsistency in types at all. I'm not sure what's the best fix here because the `hasUser` would have to take a `NodeHandle *` but `NodeUse` exists to wrap those. Maybe this doesn't matter and should be left as is. Thoughts?

jackm321:
added `getUser()` method as discussed in https://github.com/pytorch/glow/issues/1985

jackm321:
@rdzhabarov this would just cause a compilation error so I'm not sure if there is a good way to test that, is there?

Issue #1992, opened 2018-11-08
[ImageClassifier] Broken `-convert-inout-to-fp16` due to recent refactoring


qcolombet:
TL;DR We need to do the conversion from fp32 to fp16 of the data tensor (load image) after we compile not before.

When we landed #1910 we missed a subtle difference in the order we do things that leads to broken graphs when using `-convert-inout-to-fp16` alone for the image-classifier.

Basically, `-convert-inout-to-fp16` is responsible for changing the input and output placeholder from float32 to float16 and for inserting the proper convert nodes in the graph.
When this conversion sees a placeholder with already the right type, it does nothing as it should.
So far so good.

This refactoring however changed the type of the input placeholder before running the conversion has been done, thus we end up with the placeholder wrongly typed and the conversion thinks things are fine.

When you look at the code itself, it's hard to see at first glance what we did differently.
The old order look like:
1. Load the input data
2. Create the loader
3. Convert the input data from fp32 to fp16
4. Convert the input placeholder from fp32 to fp16
5. Compiling the network
6. Running

The new order swapped 2 <-> 3:
1. Load the input data
3. Convert the input data from fp32 to fp16
2. Create the loader
4. Convert the input placeholder from fp32 to fp16
5. Compiling the network
6. Running

This is problematic because the loader actually uses the type of the input tensor data to create the placeholder and thus it thinks the network is already fp16 ready and that's the bug.

The challenge with the refactoring is that 2, 4, 5 needs to happen only once, whereas 3 needs to happen at each run. We decided to move it right after 1, whereas we actually need to move it just before 6.

To reproduce, just build `image-classifier` with asserts enabled use resnet50 (for instance) and `-convert-inout-to-fp16` and the verifier should assert.

qcolombet:
Hey @jfix71!

Do you want to take a shot or should I?

jfix71:
Whoops, good catch! I'll fix it.

qcolombet:
Well I missed it during the review too :P.

Issue #1989, opened 2018-11-08
[TextTranslator] Quantized text translation is broken


jfix71:
Looks like a FunctionConverter refactor broke quantized text translation.

Before (https://github.com/pytorch/glow/commit/96377be74fde8678a50ebf43e5b6f1bc62b40a9a):
```
$ ./bin/text-translator -m=en2gr -load_profile=en2gr.yaml --do_not_quantize_nodes=Add
Enter a sentence in English to translate to German: My name is Jordan .
mein Name ist Jordan .
```

After (https://github.com/pytorch/glow/commit/2a8f7d52246fe1014d0b95e11ce2effa0968139b):
```
$ ./bin/text-translator -m=en2gr -load_profile=en2gr.yaml --do_not_quantize_nodes=Add
Enter a sentence in English to translate to German: My name is Jordan .
&quot; für , , der , . , .
```

jfix71:
@qcolombet Do you have any initial feeling/guess as to why this broke? The commit that appears to have broken it was intended to be NFC.

qcolombet:
No idea, I'll take a look.

qcolombet:
Oh, the one thing that could happen is the profile needs to be regenerated.

jfix71:
I regenerated it and unfortunately it's still broken 😕

qcolombet:
Thanks for double checking. Looking!

qcolombet:
For me 96377be is already acting strange (though not as strange as 2a8f7d5):
```
Enter a sentence in English to translate to German: My name is Jordan .
mein Name ist der Jordan @-@ Konzern , der sich mit der Verarbeitung
```

Anyway, investigating.

qcolombet:
I think I found the bug. I believe this is another instance of #1832 where essentially we decide new quantization parameters for the result of an operation and apply them, whereas the execution expects the output parameters to be the same as the input parameters. This instance of the bug is particularly nasty because usually the operators that have such constraints check it in the verifier, whereas this one was not checked :(.

The problematic node in that case is Slice.
We assign a new scale to its result, whereas the generated code doesn't perform any rescale internally, thus we should have the same scale and offset as the inputs.
The fix is trivial (will post a PR soon), but I will double check if we have more nodes in that situation and add the proper verifier checks.

jfix71:
@qcolombet Thanks for investigating this!

[...]

PR #1994, opened 2018-11-09
[ImageClassifier] Fix convert-inout-to-fp16

jfix71:
*Description*: Convert the input every time we get new inputs, and only after the graph is compiled. Do not convert the output back to fp32 anymore, as it forces making extra copies unnecessarily, and was incorrectly resetting the type of the SM tensor in the context.

*Testing*: Manually tested running in streaming mode with `-convert-inout-to-fp16`

*Documentation*: N/A

Fixes https://github.com/pytorch/glow/issues/1992

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -209,11 +209,6 @@ void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
     *inputImageData = std::move(dataNHWC);
     break;
   }
-
-  // Convert the raw input to fp16.
-  if (convertInAndOutToFp16) {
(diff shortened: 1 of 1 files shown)


jfix71:
@qcolombet Are you suggesting local tests for exercising the loaders? If so that would be great, instead of having to remember different combinations of options to try locally and assume it's probably fine. Would help with https://github.com/pytorch/glow/issues/1989 too.

qcolombet:
> @qcolombet Are you suggesting local tests for exercising the loaders? If so that would be great, instead of having to remember different combinations of options to try locally and assume it's probably fine. Would help with #1989 too.

Yes! I didn't look at it so far because I thought I was the only one annoy by this :P.

PR #1999, opened 2018-11-10
[FunctionQuantizer] Fix rescaling happening in the middle of nodes

qcolombet:
*Description*
Some nodes, at least in their interpreter implementation, expect that
their inputs and outputs are scaled with the exact same parameters.
These constraints were not enforced in the function quantizer for
the slice, transpose, and reshape nodes leading to a mismatch between
how the consumers of a value expect it to be quantized and how it
was actually quantized.

This patch teaches the function quantizer how to fix those constraints
for these additional nodes (the mechanism already existed) and adds
checks in the verifier to catch this kind of problem earlier.

*Testing*
Added test cases, but only for two nodes. Given they all use the same
code path I didn't bother adding them all.

Related to issue #1989.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -33,13 +33,15 @@ using namespace glow::quantization;
 /// This class produces a quantized function based on a provided profile.
 class FunctionQuantizer : public FunctionConverter {
 protected:
-  /// \see FunctionConverter::getTargetTypeForOutput.
-  /// Get the quantized type for \p out if any.
-  TypeRef getTargetTypeForOutput(const NodeValue &out) const override {
-    if (out.getElementType() != ElemKind::FloatTy) {
-      return nullptr;
-    }
-
+  /// Get the type that \p out should have at the end of the conversion
+  /// process regardless of what is its current type.
+  /// This is similar to ::getTargetTypeForOutput except that \p out
+  /// doesn't need to be a floating point type for this method to
+  /// return the target type.
+  /// The reason we need this method is because we may morph the type
+  /// of \p out to match some IR constraints, but the final type still
+  /// needs to be known to insert rescale nodes.
+  TypeRef getTargetTypeForOutputImpl(const NodeValue &out) const {
     const std::string nodeOutputName =
         NodeQuantizationInfo::generateNodeOutputName(out.getNode()->getName(),
                                                      out.getResNo());
@@ -50,6 +52,15 @@ protected:
     const TensorQuantizationParams &TQP = outTQPIt->second;
     return mod_.uniqueType(ElemKind::Int8QTy, out.dims(), TQP.scale,
                            TQP.offset);
+  }
+
+  /// \see FunctionConverter::getTargetTypeForOutput.
+  /// Get the quantized type for \p out if any.
+  TypeRef getTargetTypeForOutput(const NodeValue &out) const override {
+    if (out.getElementType() != ElemKind::FloatTy) {
+      return nullptr;
+    }
+    return getTargetTypeForOutputImpl(out);
   }
 
   /// \see FunctionConverter::getTargetTypeForInput.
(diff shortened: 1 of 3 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -33,13 +33,15 @@ using namespace glow::quantization;
...
-      return nullptr;
-    }
-
+  /// Get the type that \p out should have at the end of the conversion
+  /// process regardless of what is its current type.
+  /// This is similar to ::getTargetTypeForOutput except that \p out
+  /// doesn't need to be a floating point type for this method to
+  /// return the target type.
+  /// The reason we need this method is because we may morph the type
+  /// of \p out to match some IR constrains, but the final type still
```
jfix71: typo: constrains

lib/Quantization/Quantization.cpp
```diff
@@ -251,13 +276,18 @@ class FunctionQuantizer : public FunctionConverter {
 
       break;
     }
-    case Kinded::Kind::MaxPoolNodeKind:
-    case Kinded::Kind::AvgPoolNodeKind:
-    case Kinded::Kind::GatherNodeKind: {
+    casesForNodesWithIRConstraint : {
+      // Check that the main loop hands us the node in the order we expect:
+      // morph then postprocessing.
+      // If the assert breaks, that means that th morphNode and postprocessing
```
jfix71: typo: "th"

tests/unittests/quantizationTest.cpp
```diff
@@ -1111,6 +1111,154 @@ TEST(Quantization, quantizeFunctionConvertConstant) {
...
+
+  auto *input = mod.createPlaceholder(ElemKind::FloatTy, {4}, "input", true);
+  ctx.allocate(input)->init(Tensor::InitKind::Xavier, 3, mod.getPRNG());
+
+  auto *slice = F->createSlice("slice", input, {2}, {3});
+  auto *save = F->createSave("ret", slice);
+  auto *result = save->getPlaceholder();
+  ctx.allocate(result);
+
+  // Note that we are creating quantization info even for nodes that will not be
```
jfix71: I think this comment is not correct in this case?

tests/unittests/quantizationTest.cpp
```diff
@@ -1111,6 +1111,154 @@ TEST(Quantization, quantizeFunctionConvertConstant) {
...
+
+  auto *input = mod.createPlaceholder(ElemKind::FloatTy, {3, 3}, "input", true);
+  ctx.allocate(input)->init(Tensor::InitKind::Xavier, 3, mod.getPRNG());
+
+  auto *reshape = F->createReshape("reshape", input, {9});
+  auto *save = F->createSave("ret", reshape);
+  auto *result = save->getPlaceholder();
+  ctx.allocate(result);
+
+  // Note that we are creating quantization info even for nodes that will not be
```
jfix71: Same, don't think this comment applies.

tests/unittests/quantizationTest.cpp
```diff
@@ -1111,6 +1111,154 @@ TEST(Quantization, quantizeFunctionConvertConstant) {
...
+    ASSERT_EQ(result->getUsers().size(), 1);
+    auto *SN = llvm::dyn_cast<SaveNode>(result->getUsers().begin()->getUser());
+    ASSERT_TRUE(SN);
+    EXPECT_TRUE(!SN->getOutput().getType()->isQuantizedType());
+
+    // Verify that the input to save is a dequantize node.
+    auto *DN = llvm::dyn_cast<DequantizeNode>(SN->getInput());
+    ASSERT_TRUE(DN);
+
+    // Verify that the slice is rescaled after being quantized.
```
jfix71: reshape, not slice, correct?

tests/unittests/quantizationTest.cpp
qcolombet: 😱 have been caught doing copy/paste!

tests/unittests/quantizationTest.cpp
qcolombet: Indeed :)

lib/Quantization/Quantization.cpp
qcolombet: Done!

lib/Quantization/Quantization.cpp
qcolombet: Done!

tests/unittests/quantizationTest.cpp
qcolombet: Done!

tests/unittests/quantizationTest.cpp
qcolombet: Done!

tests/unittests/quantizationTest.cpp
qcolombet: Done!

qcolombet:
Thanks @jfix71 and @rdzhabarov for the quick review!

PR #2000, opened 2018-11-10
[ONNX] Support load of int32 Tensor

yinghai:
*Description*:
We need to support load of int32 tensor too because Caffe2 concat will output such tensors. In addition, at output, we need to consider outputing that tensor instead of the node. 

*Testing*:
Unit test

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -181,6 +181,8 @@ static void loadTensor(const ONNX_NAMESPACE::TensorProto &in, Tensor *T) {
       llvm_unreachable("Unsupported Tensor format.");
     }
   } else if (in.data_type() == ONNX_NAMESPACE::TensorProto::INT32) {
+    // There are few cases when we will have int32 tensors. For example, the
+    // second output of Concat from Caffe2 concat op is int32
     T->reset(ElemKind::Int32ITy, dim);
 
     if (in.int32_data_size() > 0) {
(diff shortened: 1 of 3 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -180,6 +180,21 @@ static void loadTensor(const ONNX_NAMESPACE::TensorProto &in, Tensor *T) {
     } else {
       llvm_unreachable("Unsupported Tensor format.");
     }
+  } else if (in.data_type() == ONNX_NAMESPACE::TensorProto::INT32) {
```
nadavrot: Can you please add a few lines of comments that explain what's going on?

yinghai: Ok.  Its just one extra possible data type that we need to support. The code here is highly similar.

nadavrot: Ok.

yinghai: I added the comments below this line on where int32 tensor comes from.

nadavrot: @artemrakhov The code around here in other if statements look undocumented. Can we go over and add a few comments that explain what we need the i-stream and other stuff that are not trivial?

yinghai: Ah, because the data we want can lie in either field like `int32_data`, `int64_data`, or it can be packed in the `raw_data` string.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -180,6 +180,23 @@ static void loadTensor(const ONNX_NAMESPACE::TensorProto &in, Tensor *T) {
...
+    // There are few cases when we will have int32 tensors. For example, the
+    // second output of Concat from Caffe2 concat op is int32
+    T->reset(ElemKind::Int32ITy, dim);
+
+    if (in.int32_data_size() > 0) {
+      auto TH = T->getHandle<>();
```
rdzhabarov: this needs to be T->getHandle<int32_t>

PR #2001, opened 2018-11-11
[Unittest] Improve the testing of convolutions. 

nadavrot:
*Description*:

I played with our implementation of convolutions in an attempt to figure out how to implement matrix multiplication using a convolution using @bertmaher 's method. During the investigation I discovered that when I break our convolutions the unit tests still pass. This PR improves the test coverage using two methods: 

    [Test] Add a new unit test to verify float convolutions. -- On my machine the test passes in 4.1 seconds in debug mode.

    [Test] Improve a few unit tests. - -This commit fixes a few places in our test that made the CPU backend test only the slow fallback code.

*Testing*: This is a test.
*Documentation*: Inline.

### DIFF
--- a/tests/unittests/OCLTest.cpp
+++ b/tests/unittests/OCLTest.cpp
@@ -34,8 +34,8 @@ TEST(OpenCLCorrectnessTest, convOps) {
   Tensor out1;
   Tensor out2;
 
-  inferBasicConvNet(&inputs, &out1, BackendKind::OpenCL, 64);
-  inferBasicConvNet(&inputs, &out2, BackendKind::Interpreter, 64);
+  inferBasicConvNet(&inputs, &out1, BackendKind::OpenCL, 8);
+  inferBasicConvNet(&inputs, &out2, BackendKind::Interpreter, 8);
 
   EXPECT_TRUE(out1.isEqual(out2));
 }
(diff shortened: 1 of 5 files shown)


tests/unittests/OCLTest.cpp
```diff
@@ -34,8 +34,8 @@ TEST(OpenCLCorrectnessTest, convOps) {
   Tensor out1;
   Tensor out2;
 
-  inferBasicConvNet(&inputs, &out1, BackendKind::OpenCL, 4);
-  inferBasicConvNet(&inputs, &out2, BackendKind::Interpreter, 4);
+  inferBasicConvNet(&inputs, &out1, BackendKind::OpenCL, 64);
```
rdzhabarov: we have all these tests in backend correctness tests. Maybe it's time to remove this OCLTest

tests/unittests/convTest.cpp
```diff
@@ -0,0 +1,88 @@
...
+        singleConvNet(&input, &out1, backendKind_, depth, kernel, 1, 0);
+        singleConvNet(&input, &out2, BackendKind::Interpreter, depth, kernel, 1,
+                      0);
+        EXPECT_TRUE(out1.isEqual(out2, 0.001));
+      }
+    }
+  }
+}
+
+#ifdef GLOW_WITH_CPU
```
rdzhabarov: no interpreter test?

PR #2005, opened 2018-11-12
[BackendUtils] Fix BackendUtils allocation

gcatron:
*Description*:Updated BackendUtils to not call alignedAlloc on size 0.
Fixed OpenCL backend to not copy size 0 constants to card.
*Testing*: ninja test and run.sh look good.
*Documentation*: NA

### DIFF
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -27,6 +27,10 @@ uint8_t *glow::collectConstants(
 
   // At compile time condense constants to a single block of memory.
   // This allows the graph to go away after compile time.
+  // If there are no constants return nullptr.
+  if (constantMaxSize == 0) {
+    return nullptr;
+  }
   uint8_t *baseConstantWeightVarsStore =
       (uint8_t *)alignedAlloc(constantMaxSize, TensorAlignment);
   for (auto &v : F->getGraph()->getParent()->getConstants()) {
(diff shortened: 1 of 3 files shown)


lib/Backends/BackendUtils.cpp
```diff
@@ -27,6 +27,10 @@ uint8_t *glow::collectConstants(
 
   // At compile time condense constants to a single block of memory.
   // This allows the graph to go away after compile time.
+  // If there are no constants return nullptr.
+  if (constantMaxSize == 0) {
+    return nullptr;
```
rdzhabarov: please also check the free part.

lib/Backends/BackendUtils.cpp
```diff
@@ -27,6 +27,10 @@ uint8_t *glow::collectConstants(
 
   // At compile time condense constants to a single block of memory.
   // This allows the graph to go away after compile time.
+  // If there are no constants return nullptr.
+  if (constantMaxSize == 0) {
```
rdzhabarov: please, check all the usages for constants_ from symbol table and make sure we handle `nullptr` properly :)

lib/Backends/BackendUtils.cpp
gcatron: Free is ok, alignedFree just aliases free which is a no action when called on nullptr.

PR #2007, opened 2018-11-12
[ImageClassifier] Add specialized imagenet normalization

jfix71:
*Description*: Models inferring on Imagenet can achieve better accuracy using specialized normalization for the loaded images ([as seen here](https://github.com/pytorch/examples/blob/master/imagenet/main.py#L138-L139))

*Testing*: Tested our default 3 images with Resnet50 and all looks good. I also calculated Resnet50 top-k and it has much better results (PR coming soon with script for running it), hence I am adding it to `run.sh`.


nadavrot:
Nice!

PR #2008, opened 2018-11-13
Remove context from Backend::compile

### LINKED ISSUE #1904: Pass Context at runtime not compile time

gcatron:
*Description*: Unifies backends to a single run time api and calls setup/teardown from the ExecutionEngine. Removes context from compile and compileIR as it is no longer needed at compile time.
*Testing*: ninja test, run.sh and bundles all look good.
*Documentation*: N/A
Fixes #1904

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -61,6 +61,17 @@ public:
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
   virtual void execute(Context &ctx) = 0;
+
+  /// Does any needed initialization work for the Backend.
+  /// This includes device init constant memory allocation and copying to
+  /// device.
+  virtual void setupRuns() = 0;
+  /// Per run setup. Copy inputs to device.
+  virtual void beforeRun(const Context &ctx) = 0;
+  /// Per run cleanup. Copy outputs from device.
+  virtual void afterRun(const Context &ctx) = 0;
+  /// Final cleanup. Release memory, reset device.
+  virtual void tearDownRuns() = 0;
 };
 } // end namespace glow
 
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -60,7 +60,7 @@ public:
   virtual ~CompiledFunction() = default;
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
-  virtual void execute(Context &ctx) = 0;
+  virtual void execute() = 0;
 
   /// Does any needed initialization work for the Backend.
   /// This includes device init constant memory allocation and copying to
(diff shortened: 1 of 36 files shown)


rdzhabarov:
you need to rebase on latest master to pick up the docker image changes.

include/glow/Backends/CompiledFunction.h
```diff
@@ -60,7 +60,16 @@ class CompiledFunction {
   virtual ~CompiledFunction() = default;
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
-  virtual void execute(Context &ctx) = 0;
+  virtual void execute() = 0;
+
+  /// Does any needed initialization work for the Backend.
```
rdzhabarov: can you add brief explanation of potential actions for each method. Copy constants, init device etc.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -157,8 +151,8 @@ void ExecutionEngine::optimizeFunction(CompilationMode mode, Function *F) {
 void ExecutionEngine::compile(CompilationMode mode, Function *F, Context &ctx) {
   optimizeFunction(mode, F);
   // Make sure that the context has backing tensors for all placeholders.
-  ctx.allocate(M_.getPlaceholders());
-  function_ = backend_->compile(F, ctx);
+  // ctx.allocate(M_.getPlaceholders());
```
rdzhabarov: remove commented code

rdzhabarov:
> Finishes #1904

Do we still copy weights every run?
Anything left before enabling real multithreading for handling ONNXIFI requests?

gcatron:
> > Finishes #1904
> 
> Do we still copy weights every run?
> Anything left before enabling real multithreading for handling ONNXIFI requests?

We do still copy weights every run. This PR moved the work up into the Execution engine layer. Next step is to design and build out the runtime engine that will handle concurrency, pipelining, and mutli-device issues. A short term solution would be to add logic to EE that only calls setupRuns on the first call of run.

rdzhabarov:
For sanity check, once this is merged can you update fbcode accordingly.

PR #2009, opened 2018-11-13
[utils] Add driver for determining imagenet topk accuracy

jfix71:
*Description*: This script is used as a driver to calculate Top-1 and Top-5 accuracy across Imagenet. It will resize+center-crop images prior to inference. This resize+center-crop can also be done to pre-process the images without performing inference, so that the images can be used many times repeatedly without having to resize+center-crop them over and over again.

On Resnet50, here is what I found (showing results with and without specialized imagenet normalization https://github.com/pytorch/glow/pull/2007):

- Float: 
  - Interpreter/CPU Backend (same results):
    - Standard normalization:
      - Top-1 accuracy: 0.6846
      - Top-5 accuracy: 0.8863
    - Specialized Imagenet normalization:
      - Top-1 accuracy: 0.7613
      - Top-5 accuracy: 0.9293

- Quantized:
  - Interpreter:
    - Standard normalization:
      - Top-1 accuracy: 0.6682
      - Top-5 accuracy: 0.8767
    - Specialized Imagenet normalization:
      - Top-1 accuracy: 0.7500
      - Top-5 accuracy: 0.9236
  - CPU Backend:
    - Standard normalization:
      - Top-1 accuracy: 0.6731
      - Top-5 accuracy: 0.8795
    - Specialized Imagenet normalization:
      - Top-1 accuracy: 0.7497
      - Top-5 accuracy: 0.9236

*Testing*: Tested it with many different inputs.

*Documentation*: Added to `Testing.md`

### DIFF
--- a/utils/imagenet_topk_accuracy_driver.py
+++ b/utils/imagenet_topk_accuracy_driver.py
@@ -1,0 +1,335 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import shutil
+import sys
+import tempfile
+
+import argparse
+import os
+import pexpect
+
+import PIL.Image as Image
+import torchvision
+
+parser = argparse.ArgumentParser(description="Glow image-classifier Driver for "
+                                 "TopK ImageNet Calculation")
+
+parser.add_argument("--validation-images-dir", metavar="DIR", required=True,
+                    help="Path to the directory containing the validation set "
+                    "of images. Subdirectories are expected to be organized "
+                    "such that when sorted their index corresponds to their "
+                    "label. For example, if the validation_images_dir contains "
+                    "{'abc/', 'def/', 'ghi/'}, then this should correspond to "
+                    "labels {0, 1, 2} respectively.")
+
+parser.add_argument("--batch-size", default=1, type=int, metavar="N",
+                    help="Batch size for use with the model. The total number "
+                    "of images in the validation_images_dir should be "
+                    "divisible by the batch size.")
+
+parser.add_argument("--only-resize-and-save", default=False,
+                    action="store_true", help="Use to pre-process images "
+                    "to 224x224. Saves the images to "
+                    "the validation_images_dir/processed/")
+
+parser.add_argument("--resize-input-images", default=False,
+                    action="store_true", help="Resize and center-crop images "
+                    "at runtime to 224x224.")
+
+parser.add_argument("--verbose", default=False,
+                    action="store_true", help="Verbose printing.")
+
+parser.add_argument("--image-classifier-cmd", default="",
+                    help="Command to use for running the image-classifier, "
+                    "including the binary and all of its command lime "
+                    "parameters.")
+
+## Opens and returns an image located at @param path using the PIL loader.
+def pil_loader(path):
+    # open path as file to avoid ResourceWarning
+    # (https://github.com/python-pillow/Pillow/issues/835)
+    with open(path, "rb") as img:
+        img = Image.open(img)
+        return img.convert("RGB")
+
+## Opens and returns an image located at @param path using the accimage loader.
+def accimage_loader(path):
+    import accimage
+    try:
+        return accimage.Image(path)
+    except IOError:
+        # Potentially a decoding problem, fall back to PIL.Image
+        return pil_loader(path)
+
+
+## Opens and returns an image located at @param path using either the accimage
+# loader or PIL loader.
+def default_image_loader(path):
+    if torchvision.get_image_backend() == "accimage":
+        return accimage_loader(path)
+    return pil_loader(path)
+
+def get_sorted_img_subdirs(validation_images_dir):
+    img_dir_paths = []
+    for img_dir in os.listdir(validation_images_dir):
+        dir_path = os.path.join(validation_images_dir, img_dir)
+        if os.path.isdir(dir_path):
+            img_dir_paths.append(img_dir)
+    img_dir_paths.sort()
+
+    return img_dir_paths
+
+
+## @returns two lists of the same length found in directory
+# @param validation_images_dir; the first list contains paths to all images
+# found, and the second list contains the corresponding labels of the image.
+def get_img_paths_and_labels(validation_images_dir):
+    img_subdirs = get_sorted_img_subdirs(validation_images_dir)
+
+    # Create lists holding paths to each image to be classified and the label
+    # for that image.
+    img_paths = []
+    img_labels = []
+    curr_label_idx = 0
+    for img_subdir in img_subdirs:
+        img_subdir_path = os.path.join(validation_images_dir, img_subdir)
+        for img in os.listdir(img_subdir_path):
+            full_img_path = os.path.join(img_subdir_path, img)
+            if os.path.isfile(full_img_path):
+                img_paths.append(full_img_path)
+                img_labels.append(curr_label_idx)
+        curr_label_idx = curr_label_idx + 1
+    return img_paths, img_labels
+
+## Given an image located at @param img_path, transform the image
+# and save it to the path @param path_to_new_img.
+def resize_and_save_image(img_path, path_to_new_img):
+    # Load the image.
+    img = default_image_loader(img_path)
+
+    # Use to Resize and CenterCrop the images to 224x224.
+    transform_resize = torchvision.transforms.Compose([
+        torchvision.transforms.Resize(256),
+        torchvision.transforms.CenterCrop(224),
+    ])
+
+    resized_img = transform_resize(img)
+    resized_img.save(path_to_new_img, format="png")
+
+## Used to pre-process an input set of images. Takes a string of a directory
+# @param validation_images_dir and saves the cropped subset of the images in a
+# subdirectory `processed/`, which must not yet exist.
+def save_centered_cropped_dataset(validation_images_dir):
+    processed_validation_images_dir = os.path.join(validation_images_dir,
+                                                   "processed")
+    print "Saving centered cropped input images: %s" % (
+          processed_validation_images_dir)
+
+    img_subdirs = get_sorted_img_subdirs(validation_images_dir)
+
+    try:
+        os.makedirs(processed_validation_images_dir)
+    except OSError:
+        sys.exit("New validation directory must not exist")
+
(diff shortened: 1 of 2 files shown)


jfix71:
I'm by no means a python expert, so happy to get suggestions here!

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,303 @@
...
+## top1_count, and @param top5_count.
+def print_topk_accuracy(total_image_count, top1_count, top5_count):
+    top1_accuracy = float(top1_count) / float(total_image_count)
+    top5_accuracy = float(top5_count) / float(total_image_count)
+    print("\tTop-1 accuracy: " + "{0:.4f}".format(top1_accuracy))
+    print("\tTop-5 accuracy: " + "{0:.4f}".format(top5_accuracy))
+
+
+def main():
+    global args
```
bertmaher: Can you avoid making `args` global, and pass the relevant bits down instead?

rdzhabarov:
This is amazing! Where do you get inputs (expected outputs) for the accuracy test?

utils/imagenet_topk_accuracy_driver.py
jfix71: Done!

jfix71:
@rdzhabarov External users can download labeled images from the [Imagenet dataset](http://www.image-net.org/)

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+import tempfile
+
+import PIL.Image as Image
+
+import torchvision.transforms as transforms
+
+parser = argparse.ArgumentParser(description='Glow image-classifier Driver for TopK ImageNet Calculation')
+
+parser.add_argument('--validation-images-dir', metavar='DIR', required=True,
+                    help=" Path to the directory containing the validation set "
```
artemrakhov: nit: " Path -> "Path

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+def accimage_loader(path):
+    import accimage
+    try:
+        return accimage.Image(path)
+    except IOError:
+        # Potentially a decoding problem, fall back to PIL.Image
+        return pil_loader(path)
+
+def default_image_loader(path):
+    from torchvision import get_image_backend
```
artemrakhov: please don't do import-not-at-top

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+    ])
+
+    resized_img = transform_resize(img)
+    resized_img.save(path_to_new_img, format="png")
+
+## Used to pre-process an input set of images. Takes a string of a directory
+# @param validation_images_dir and saves the cropped subset of the images in a
+# subdirectory `processed/`, which must not yet exist.
+def save_centered_cropped_dataset(validation_images_dir):
+    assert os.path.exists(validation_images_dir), "Validation directory does not exist: " + validation_images_dir
```
artemrakhov: This and some other lines are longer than 80 chars. Could you use pylint or something like this?

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+
+    try:
+        p = pexpect.spawn(args.image_classifier_cmd, logfile=sys.stdout,
+                          timeout=None)
+
+        # Create a temporary directory to store the transformed image we
+        # classify (if applicable) and the log of image-classifer output.
+        tmp_dir_name = tempfile.mkdtemp()
+        path_to_tmp_log = os.path.join(tmp_dir_name, "log.txt")
+        fout = file(path_to_tmp_log,'w')
```
artemrakhov: space between args please

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+                top5_labels = []
+                for x in xrange(5):
+                    label_and_prob = p.readline()
+                    # Get the label from the line.
+                    label = label_and_prob.split()[1]
+                    top5_labels.append(int(label))
+
+                expected_label = img_labels[img_index + batch_idx]
+                if expected_label == top5_labels[0]:
+                    top1_count = top1_count + 1
```
artemrakhov: you can do `+= 1`

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+
+    assert total_image_count % args.batch_size == 0, "Total number of images must be divisible by batch size"
+
+    if args.verbose:
+        print("Running image classifier with: " + args.image_classifier_cmd)
+
+    tmp_dir_name = ""
+
+    try:
+        p = pexpect.spawn(args.image_classifier_cmd, logfile=sys.stdout,
```
artemrakhov: can you move this line after `fout = ` and pass `logfile=fout`?

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+            if os.path.isfile(orig_img_path):
+                processed_img_path = os.path.join(processed_dir_path,
+                                                  orig_img_filename)
+                resize_and_save_image(orig_img_path, processed_img_path)
+
+## @returns a list of strings (of length equal to the batch size) which are
+# paths to images to do inference on. @param img_paths is the set of all image
+# paths, @param img_index is the next index to use in @param img_paths, and
+# @param tmp_dir_name is the location of where to save the images if necessary.
+def get_curr_img_paths(img_paths, img_index, tmp_dir_name, batch_size,
```
artemrakhov: I see there is a lot of duplicated logic between the function above and this one. Could you do "crop and center" for all images together before starting to run image-classifier? That way, all this logic can go into one function. It can even cache automatically by looking at whether "processed" dir already exists.

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,318 @@
...
+                      print("")
+
+    finally:
+        p.close(force=True)
+
+        # Remove the temp directory we used to save the images and log.
+        if tmp_dir_name:
+            shutil.rmtree(tmp_dir_name)
+
+    print("\nCompleted running; Final Top-1/5 accuracy across " +
```
artemrakhov: I see a mix of double-quotes and single quotes for strings. And also `print` sometimes with brackets and sometimes without. I believe pylint can detect this too.

utils/imagenet_topk_accuracy_driver.py
jfix71: I removed this one, however the accimage above is still there -- it seems like a conditional import based on the if below (I took this code snippet from a different run script). I'm not sure of a good way to get around this, do you have any suggestion?

artemrakhov: For accimage it's actually fine: it may not be present.

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,332 @@
...
+        curr_img_paths.append(img_path)
+
+    return curr_img_paths
+
+## Verifies that the @param image_classifier_cmd is well formatted via
+# assertions.
+def verify_spawn_cmd(image_classifier_cmd):
+    split_cmd = image_classifier_cmd.split()
+
+    assert "image-classifier" in split_cmd[0], \
```
artemrakhov: Using `\` for line breaks is discouraged by many style guides. You can use:
```
assert "image-classifier" in split_cmd[0], (
    "image-classifier binary must be provided first.")
```

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,332 @@
...
+        torchvision.transforms.CenterCrop(224),
+    ])
+
+    resized_img = transform_resize(img)
+    resized_img.save(path_to_new_img, format="png")
+
+## Used to pre-process an input set of images. Takes a string of a directory
+# @param validation_images_dir and saves the cropped subset of the images in a
+# subdirectory `processed/`, which must not yet exist.
+def save_centered_cropped_dataset(validation_images_dir):
```
artemrakhov: I think you can use this function for both "only-resize-and-save" and "resize-input-images" cases. Add a new argument "target_dir", and pass either `join(validation_images_dir, "processed")` or tmpdir (only run it if "resize-input-images"). 

So in case of "resize-input-images" you'll do all resizing in advance before starting classification. You'll reassign `img_paths` variable, and instead of calling `get_curr_img_paths` you'll just do: `img_paths[img_index, img_index + batch_size]`. Cleanup will happen automatically, when you remove tmpdir after `finally:`.

utils/imagenet_topk_accuracy_driver.py
jfix71: Done

utils/imagenet_topk_accuracy_driver.py
jfix71: >So in case of "resize-input-images" you'll do all resizing in advance before starting classification. 

Got it -- yeah that could work, but the downside here is the resized images must all exist at once. So for example the 50k images I was looking at would take up 4.2GB temporarily. I felt it was a waste to create that much temporary space on disk, but if you prefer that version in order to remove some duplicated code I can change it.

artemrakhov: Agree, that's a lot of space. But my understanding is that in current version we only clean up once, in the very end.

jfix71: Ah, actually the current version just reuses the same image name over and over again at [this line](https://github.com/pytorch/glow/pull/2009/files#diff-4ea927ddc0aaa413fc99b55398567f7aR182). So there are only ever `batch_size` temp images in the temp directory which is then cleaned up in the end. I can clarify that in the comments there if we proceed with this version.

artemrakhov: Oh, I totally missed this. Well, feel free to leave this logic then.

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,332 @@
...
+
+                expected_label = img_labels[img_index + batch_idx]
+                if expected_label == top5_labels[0]:
+                    top1_count += 1
+                if expected_label in top5_labels:
+                    top5_count += 1
+
+            curr_completed_count = img_index + batch_size
+            if curr_completed_count % 100 == 0:
+                print "Finished image index %s out of %s" % (
```
artemrakhov: can you use %d and not cast to `str`?

utils/imagenet_topk_accuracy_driver.py
```diff
@@ -0,0 +1,332 @@
...
+                else:
+                    print ""
+
+    finally:
+        classifier_proc.close(force=True)
+
+        # Remove the temp directory we used to save the images and log.
+        shutil.rmtree(tmp_dir_name)
+
+    print "\nCompleted running; Final Top-1/5 accuracy across %s images:" % (
```
artemrakhov: can you use %d and not cast to `str`?

jfix71:
Thanks for the reviews @artemrakhov!

Issue #1181, opened 2018-06-27
Consider enabling OpenCL in CI by using pocl.


lygstate:
And currently the CPU jit version is far slower than the Interpreter version, that's werid.

rdzhabarov:
> And currently the CPU jit version is far slower than the Interpreter version, that's werid.

where do you see this?

lygstate:
@rdzhabarov  from the ci， maybe because the compiling time is long

rdzhabarov:
it's because CPU version uses ASAN build.

rdzhabarov:
Some of the issues that could be caught in CI:
* https://github.com/pytorch/glow/pull/1185

opti-mix:
@lygstate Do you have more information about using POCL with Travis? 

What needs to be done to use POCL on Linux (and MacOS)? 
Are there pre-built binary packages for it? 
To use pocl, does the program just needs to link against the library and that's it?
If pocl is installed, would CMake be able to find and use it automatically when we look for OpenCL?

qcolombet:
+1 on enabling opencl testing in our CI.
It bit us again in #1959

opti-mix:
It seems like if pocl is installed on the system (by means of apt-get or any other package manager), we would need to add something like `GLOW_WITH_POCL` option to cmake and then in the CMake file in the OpenCL backend we'd need to replace this:

```
target_link_libraries(OpenCL
                      PRIVATE
                      OpenCL::OpenCL)
```

by something like:
```
if(NOT pocl_FOUND)
   set (OPENCL_LIBS OpenCL::OpenCL)
else()
   set (OPENCL_LIBS pocl)
endif()   

target_link_libraries(OpenCL
                      PRIVATE
                      ${OPENCL_LIBS})
```

This would prefer pocl to the native OpenCL libraries if GLOW_WITH_POCL was configured.

PR #2010, opened 2018-11-13
[opencl ci] Compile Glow with OpenCL and run some unit tests on CI.

rdzhabarov:
*Description*:
Add an option to compile with OpenCL support but do not run tests.
This is a first step to enable catching issues with OpenCL.

No major changes in build CI time for/after this PR.

See https://github.com/pytorch/glow/issues/1181 for details.

*Testing*:
CI

*Documentation*:
N/A

Follow up needs to enable tests with all the below fixed.
Failing tests:
```
[  FAILED  ] 11 tests, listed below:
[  FAILED  ] OpenCL/Operator.matMul/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.Transpose3Dims/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.ScatterAssign/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.TestQuantizedRescaleSequence/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.Squeeze/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.ExpandDims/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.Reshape/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.ReshapeInt/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.sliceReshape/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.Flatten/0, where GetParam() = 4-byte object <01-00 00-00>
[  FAILED  ] OpenCL/Operator.QuantizeSimple/0, where GetParam() = 4-byte object <01-00 00-00>


[  FAILED  ] OpenCL/BackendTest.simplePlaceholderValue/0, where GetParam() = 4-byte object <01-00 00-00> (103 ms)
```

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -6,6 +6,24 @@ set -ex
 
 export MAX_JOBS=8
 
+install_pocl() {
+   sudo apt-get install -y ocl-icd-opencl-dev clinfo libhwloc-dev
+   
+   wget https://github.com/pocl/pocl/archive/v1.2.tar.gz
+   tar xf v1.2.tar.gz
+   mkdir build_pocl
+   cd build_pocl
+   cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_ICD=ON ../pocl-1.2
+   make -j`nproc`
+   sudo make install
+
+   sudo mkdir -p /etc/OpenCL/vendors/
+   sudo cp /usr/local/etc/OpenCL/vendors/pocl.icd /etc/OpenCL/vendors/
+
+   clinfo
+   cd ../
+}
+
 # setup sccache wrappers
 if hash sccache 2>/dev/null; then
     SCCACHE_BIN_DIR="/tmp/sccache"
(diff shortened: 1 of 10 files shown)


opti-mix:
@rdzhabarov Thanks a lot for looking into this! 

As far as I understand, you managed to make the OpenCL backend and tests compile and link during the CI builds by making use of POCL. This is a pretty big step already! Great progress! This would allow us to catch any source-breaking changes.

The remaining problems seem to be related to the test execution failures. Fixing those would be nice as well, but this is a separate issue, IMHO.

BTW, when I was working on the OpenCL backend I've often used the debug-instrumentation to compare runs on different OpenCL devices and detect when they start to diverge from each other. So, may be we should just try to compare a run of a failing test on a local machine using `-opencl -device 0` (which is a CPU) with a run on the CI machine, which uses POCL to run on a CPU. May be we'll see something interesting.

rdzhabarov:
> I've often used the debug-instrumentation to compare runs on different OpenCL devices

I tried to enable it and saw core dump, need to investigate that.

rdzhabarov:
@opti-mix finally made this work, some tests are skipped and we can work on that in follow up prs.

opti-mix:
@rdzhabarov Very cool! Thanks a lot for finding a way to build it on CI!

rdzhabarov:
Yes, I'll split into two commits here

opti-mix:
@rdzhabarov I reproduced locally and found the reason for the failing tests. There is a bug in POCL! I created an issue: https://github.com/pocl/pocl/issues/677

opti-mix:
@rdzhabarov Apparently, they fixes this bug two weeks ago on their master branch. So, if you'd just build their master instead of the specific version you were using so far, all tests would pass. I checked it locally and they all pass.

rdzhabarov:
nice! let me update that.

PR #2012, opened 2018-11-13
sink transpose below ChannelShuffle

### LINKED ISSUE #1640: [GraphOptimizer] Remove unnecessary NHWC-NCWH transposes.
In Shufflenet we have unnecessary transposes between NCHW and NHWC. We transpose to NCHW, then do some shuffling of the C dimension (via Reshape-Transpose-Reshape), then transpose again back to NHWC. This is unnecessary -- we can remove the transposes between NCHW and NHWC, and then modify the Dims/Shuffle of the Reshape-Transpose-Reshape to shuffle the C dimension.

Such a subgraph from Shufflenet can be seen here:
shufflenet_subgraph

jackm321:
*Description*:
Sink Tranposes below nodes from ChannelShuffle (Reshape->Tranpose->Reshape) in GraphOptimizer. See more details in #1640.

*Testing*:
Added a new unit test, ran `ninja all && ninja test`
Also ran on `shufflenet/model.onnx` network before and after the optimization to see that transposes are correctly sunk and then eliminated by the identity transpose optimization.

DAGS for shufflenet before and after optimization
before: [shufflenetDAG_before.pdf]
after:(https://github.com/pytorch/glow/files/2577623/shufflenetDAG_before.pdf)
[shufflenetDAG_after.pdf](https://github.com/pytorch/glow/files/2577624/shufflenetDAG_after.pdf)

highlight of change in DAGS
before:  
screen shot 2018-11-13 at 10 59 49 am

after:  
screen shot 2018-11-13 at 11 00 04 am

*Documentation*:
comments in code
Fixes #1640


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,39 @@ static Node *simplifyNode(Node *node, Function *F) {
   return node;
 }
 
+// 2D vector of dimentions mapping a dimention that is reshaped into either 1
+// or 2 dimentions by the first reshape node in a ChannelShuffle operation
+using ChannelShuffleDimGroups =
+    llvm::SmallVector<llvm::SmallVector<unsigned_t, 2>, max_tensor_dimensions>;
+
+/// \returns the dims of \p node grouped by which input dimention they came from
```
jackm321: Oops I'll fix the description here to say returns success/failure and populates dimGroups with the output

tests/unittests/graphOptzTest.cpp
```diff
@@ -2152,3 +2152,40 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
   // transpose should be eliminated and replaced by the transposed constant.
   EXPECT_TRUE(llvm::isa<Constant>(save1->getInput()));
 }
+
+TEST_F(GraphOptz, sinkTransposeBelowChannelShuffleNodes) {
```
jfix71: nit: Doxygen comments on the purpose of the test.

tests/unittests/graphOptzTest.cpp
```diff
@@ -2152,3 +2152,40 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
...
+  const size_t inputDims[] = {3, 28, 28, 136};
+  Node *K;
+
+  K = mod_.createPlaceholder(ElemKind::FloatTy, inputDims, "input", false);
```
jfix71: You should be able to do `Node *K = ...` here and remove it above.

tests/unittests/graphOptzTest.cpp
```diff
@@ -2152,3 +2152,40 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
...
+  K = F_->createChannelShuffle("channel_shuffle", K, 4, 1);
+  K = F_->createTranspose("unnecessary_transpose_2", K, {0, 2, 3, 1});
+  auto *save = F_->createSave("ret", K);
+
+  EXPECT_EQ(F_->getNodes().size(), 6);
+
+  // Optimize away the unnecessary transposes
+  optimize(F_, CompilationMode::Infer);
+
+  // ensure the two unnecessary transposes a gone
```
jfix71: super nit: For comments everywhere, prefer capitalizing the first word and periods at the end of sentences :)

tests/unittests/graphOptzTest.cpp
```diff
@@ -2152,3 +2152,40 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
...
+  EXPECT_EQ(F_->getNodes().size(), 4);
+
+  // check that the channel shuffle nodes are still there
+  auto *RN1 = llvm::dyn_cast<ReshapeNode>(save->getInput().getNode());
+  EXPECT_NE(nullptr, RN1);
```
jfix71: These need to be `ASSERT_NE()`, as otherwise the test will segfault below. `EXPECT_NE()` doesn't exit the test on failure while `ASSERT_NE()` does.

tests/unittests/graphOptzTest.cpp
```diff
@@ -2152,3 +2152,40 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
...
+  auto *TR1 =  llvm::dyn_cast<TransposeNode>(RN1->getInput().getNode());
+  EXPECT_NE(nullptr, TR1);
+  auto *RN2 = llvm::dyn_cast<ReshapeNode>(TR1->getInput().getNode());
+  EXPECT_NE(nullptr, RN2);
+
+  // ensure last reshape has the same dimentions as the input
+  EXPECT_EQ(RN1->getDims(), llvm::makeArrayRef(inputDims));
```
jfix71: Huh, it was necessarily to make this and below array refs?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -326,6 +359,84 @@ static bool sinkCode(Function *F) {
       }
     }
 
+    // Sink Transpose below ChannelShuffle (Reshape->Transpose->Reshape)
+    if (auto *RN1 = dyn_cast<ReshapeNode>(node)) {
```
jfix71: Can you extract this to its own function? It's a bit long IMO to include along with all the others.

jfix71: nit: Can you use different names for the `RN1`/`RN2`/`TR1`/`TR2` here? It's a little hard to follow, especially because in the actual dataflow graph `RN2` comes before `RN1`, etc. given the current names.

E.g. something like `preShuffleRN`, `postShuffleRN`, `shuffleTR`, and `TR`.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -326,6 +359,84 @@ static bool sinkCode(Function *F) {
       }
     }
 
+    // Sink Transpose below ChannelShuffle (Reshape->Transpose->Reshape)
```
jfix71: Can you add a bit more of an explanation of what happens when we successfully sink? E.g. mention something about changing the shuffle dimensions of the Transpose

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,39 @@ static Node *simplifyNode(Node *node, Function *F) {
   return node;
 }
 
+// 2D vector of dimentions mapping a dimention that is reshaped into either 1
+// or 2 dimentions by the first reshape node in a ChannelShuffle operation
+using ChannelShuffleDimGroups =
+    llvm::SmallVector<llvm::SmallVector<unsigned_t, 2>, max_tensor_dimensions>;
+
+/// \returns the dims of \p node grouped by which input dimention they came from
+static bool groupDimentionsFromChannelShuffleReshape(
```
jfix71: nit: Dimensions is spelled incorrectly everywhere

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,39 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+  }
+  dimGroups.resize(inputDims.size());
+  bool foundReshapedDim = false;
+  int dimIdx, groupIdx;
+  for (dimIdx = 0, groupIdx = 0; dimIdx < dims.size(); ++dimIdx, ++groupIdx) {
+    auto &group = dimGroups[groupIdx];
+    group.push_back(dimIdx);
+    if (std::find(inputDims.begin(), inputDims.end(), dims[dimIdx]) ==
+        inputDims.end()) {
+      // make sure only one input dimention is changed and there's atleast one
```
jfix71: nit: atleast -> at least

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,39 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+    const ReshapeNode &node, ChannelShuffleDimGroups &dimGroups) {
+  const auto dims = node.getDims();
+  const auto inputDims = node.getInput().dims();
+  if (dims.size() != inputDims.size() + 1) {
+    return false;
+  }
+  dimGroups.resize(inputDims.size());
+  bool foundReshapedDim = false;
+  int dimIdx, groupIdx;
+  for (dimIdx = 0, groupIdx = 0; dimIdx < dims.size(); ++dimIdx, ++groupIdx) {
```
jfix71: I think you should be able to do 
```suggestion
  for (int dimIdx = 0, groupIdx = 0; dimIdx < dims.size(); ++dimIdx, ++groupIdx) {
```
And remove `int dimIdx, groupIdx;` above. And actually it should be `size_t`, not `int`

lib/Optimizer/GraphOptimizer.cpp
rdzhabarov: nit: dimensions :P everywhere

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,39 @@ static Node *simplifyNode(Node *node, Function *F) {
...
 }
 
+// 2D vector of dimentions mapping a dimention that is reshaped into either 1
+// or 2 dimentions by the first reshape node in a ChannelShuffle operation
+using ChannelShuffleDimGroups =
+    llvm::SmallVector<llvm::SmallVector<unsigned_t, 2>, max_tensor_dimensions>;
+
+/// \returns the dims of \p node grouped by which input dimention they came from
+static bool groupDimentionsFromChannelShuffleReshape(
+    const ReshapeNode &node, ChannelShuffleDimGroups &dimGroups) {
```
rdzhabarov: can you also update /docs for the optimization.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -326,6 +359,84 @@ static bool sinkCode(Function *F) {
...
+      if (!TR2) {
+        continue;
+      }
+
+      ChannelShuffleDimGroups rn2DimGroups;
+      if (!groupDimentionsFromChannelShuffleReshape(*RN2, rn2DimGroups)) {
+        continue;
+      }
+
+      // create newRN2
```
jfix71: Can you add more descriptive comments about what exactly is going on in each for loop here?

tests/unittests/graphOptzTest.cpp
jackm321: Yeah, I think something to do with these gtest macros not know what operator== to use otherwise

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -238,6 +242,118 @@ static bool groupDimentionsFromChannelShuffleReshape(
   return true;
 }
 
+// Sink Transpose below ChannelShuffle node sequence. For example
+// (Transpose_1->Reshape_1->Transpose_2->Reshape_2) becomes
+// (Reshape_1->Transpose_2->Reshape_2->Transpose_1) and Reshape_1,
+// Transpose_2, Reshape_2 are reconstructed so that the end result is the
+// ChannelShuffle operation on the same dimention as before Transpose_1 was
```
jackm321: dimention -> dimension
🤦‍♂️

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+    llvm::SmallVector<llvm::SmallVector<unsigned_t, 2>, max_tensor_dimensions>;
+
+/// fills \p dimGroups with the groups of dimensions of the result of \p node
+/// that get separated. This grouping is only possible if \p node is the first
+/// Reshape node in a ChannelShuffle operation (Reshape->Tranpose->Reshape)
+/// \returns true if successfully grouped output dimensions and false otherwise
+static bool
+groupDimensionsFromChannelShuffleReshape(const ReshapeNode &node,
+                                         ChannelShuffleDimGroups &dimGroups) {
+  const auto dims = node.getDims();
```
rdzhabarov: I think it's better to use concrete types(instead of auto) when the return type of the RHS is not clear.
This is of course arguable but it makes code easier to read and review.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+                                                preShuffleRNDimGroups)) {
+    return false;
+  }
+
+  // Get the inverse of the shuffle applied by the TranposeNode that is being
+  // sunk in order to apply this to the dimensions of preShuffleRN.
+  auto sinkingTRShuffle = sinkingTR->getShuffle();
+  llvm::SmallVector<unsigned_t, max_tensor_dimensions> inverseSinkingTRShuffle(
+      sinkingTRShuffle.size());
+  for (auto i = 0; i < sinkingTRShuffle.size(); i++) {
```
rdzhabarov: the convention we follow is to call size once, see loops here:
https://llvm.org/docs/CodingStandards.html#turn-predicate-loops-into-predicate-functions

tests/unittests/graphOptzTest.cpp
rdzhabarov: EXPECT_TRUE(RN1->getDims().equals(inputDims)) should make it

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+  dimGroups.resize(inputDims.size());
+  bool foundReshapedDim = false;
+  for (size_t dimIdx = 0, groupIdx = 0; dimIdx < dims.size();
+       ++dimIdx, ++groupIdx) {
+    auto &group = dimGroups[groupIdx];
+    group.push_back(dimIdx);
+    if (std::find(inputDims.begin(), inputDims.end(), dims[dimIdx]) ==
+        inputDims.end()) {
+      // make sure only one input dimension is changed and there's at least one
+      // more dim
```
nadavrot: Please capitalize your comments and end sentences with a period.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+groupDimensionsFromChannelShuffleReshape(const ReshapeNode &node,
+                                         ChannelShuffleDimGroups &dimGroups) {
+  const auto dims = node.getDims();
+  const auto inputDims = node.getInput().dims();
+  if (dims.size() != inputDims.size() + 1) {
+    return false;
+  }
+  dimGroups.resize(inputDims.size());
+  bool foundReshapedDim = false;
+  for (size_t dimIdx = 0, groupIdx = 0; dimIdx < dims.size();
```
nadavrot: Comments please.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
   return node;
 }
 
+// 2D vector of dimensions mapping a dimension that is reshaped into either 1
```
nadavrot: Period at the end of the sentence.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+// or 2 dimensions by the first reshape node in a ChannelShuffle operation
+using ChannelShuffleDimGroups =
+    llvm::SmallVector<llvm::SmallVector<unsigned_t, 2>, max_tensor_dimensions>;
+
+/// fills \p dimGroups with the groups of dimensions of the result of \p node
+/// that get separated. This grouping is only possible if \p node is the first
+/// Reshape node in a ChannelShuffle operation (Reshape->Tranpose->Reshape)
+/// \returns true if successfully grouped output dimensions and false otherwise
```
nadavrot: Period.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +205,155 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+  return true;
+}
+
+// Sink Transpose below ChannelShuffle node sequence. For example
+// (Transpose_1->Reshape_1->Transpose_2->Reshape_2) becomes
+// (Reshape_1->Transpose_2->Reshape_2->Transpose_1) and Reshape_1,
+// Transpose_2, Reshape_2 are reconstructed so that the end result is the
+// ChannelShuffle operation on the same dimention as before Transpose_1 was
+// sunk.
+static bool sinkTranposeBelowChannelShuffle(Function *F, Node *node) {
```
nadavrot: This function is long. Can we refactor some of the logic in the bottom of the function to new functions?

jackm321: I agree it's kinda long but I don't see too many opportunities for extracting something cleanly from this in its current form, do you have anything in particular that you'd like to see abstracted? Two thoughts I have on changing it though are
1. I could rewrite the bottom part using `Graph::createChannelShuffle` and then change the first portion of the code to deduce the correct parameters to the original ChannelShuffle and then and apply the inverse transpose to those to get the new parameters to feed to `createChannelShuffle`
2. we could go ahead and create a ChannelShuffle node as @jfix71 mentioned earlier that get's created at graph load time instead of immediately forming the reshape->transpose->reshape and do optimizations on that node however the shufflenet onnx file has this reshape->transpose->reshape already in it so in cases like this we'd have to first collapse this pattern to a ChannelShuffle when it's seen in a graph and then lower it back to reshape->transpose->reshape after the first round of graph optimization.
If we think that either of those is better than the current approach I can do that instead.

jackm321: I personally think that option 1 is a good middle ground here to go with

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +206,96 @@ static Node *simplifyNode(Node *node, Function *F) {
   return node;
 }
 
+// Parameters that are used to define ChannelShuffle operators
```
jackm321: I will add period.

lib/Optimizer/GraphOptimizer.cpp
jfix71: Yeah, I think we should not add a ChannelShuffleNode and just try to refactor the optimization here to make it as readable as possible.

jackm321: b5f6a7b does this refactoring and I think it ended up being much cleaner

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -205,6 +206,96 @@ static Node *simplifyNode(Node *node, Function *F) {
...
+
+  // Check that there is one more output dimension than input dimension.
+  if (resultDims.size() != inputDims.size() + 1) {
+    return resM;
+  }
+
+  // Find the first output dimension that doesn't match its corresponding input
+  // dimension.
+  ChannelShuffleParams params;
+  for (unsigned i = 0, e = resultDims.size(); i < e - 1; ++i) {
```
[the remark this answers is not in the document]
jackm321: You might be right about there being an issue here, this is a case I hadn't considered but I don't think I understand the trouble with specific example you give. In this case the loop will run until the end without finding any mismatch in input and output dims and the function will return the empty optional and no sinking will happen here and later trivial Reshape can be optimized away. Is this correct?

(Actually there is an issue that I think is unrelated to what your'e saying, `params` is uninitialized so the below access of it is invalid, I'll fix that)

lib/Optimizer/GraphOptimizer.cpp
rdzhabarov: doxygen comment :)

lib/Optimizer/GraphOptimizer.cpp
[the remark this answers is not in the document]
jfix71: @jackm321 Yeah, the uninitialized `ChannelShuffleParams` is what I was referring to -- sorry I wasn't clear.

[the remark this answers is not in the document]
jackm321: @jfix71 good catch, thanks!

lib/Optimizer/GraphOptimizer.cpp
jackm321: Something that's not good about having this in another function is that later on if someone add an if branch to `sinkCode()` for `ReshapeNode` in the same style as the other branches there (with a `continue` at the bottom then this code could potentially get skipped always.

jfix71: Agreed -- maybe you could change it so that you check if the node is a `ReshapeNode` in `sinkCode()`, and then pass the ReshapeNode into `sinkTranposeBelowChannelShuffle()`.

tests/unittests/graphOptzTest.cpp
rdzhabarov:
```

../tests/unittests/graphOptzTest.cpp:2220:78: error: call of overloaded ‘makeArrayRef(<brace-enclosed initializer list>)’ is ambiguous
   EXPECT_EQ(TR1->getShuffle(), llvm::makeArrayRef<unsigned_t>({0, 1, 2, 4, 3}));
```
this fails to compile. Did the getDims().equals not work?

lib/Optimizer/GraphOptimizer.cpp
jackm321: @jfix71 good suggestion

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -22,6 +22,7 @@
 #include "glow/Optimizer/Optimizer.h"
 #include "glow/Quantization/Base/Base.h"
 
+#include "llvm/ADT/APInt.h"
```
jfix71: What is this used for?

tests/unittests/graphOptzTest.cpp
jfix71: This description isn't fully accurate -- it's testing that the `Transpose-ChannelShuffle-Transpose` sequence can be optimized to remove both transposes, by sinking the first Transpose.

tests/unittests/graphOptzTest.cpp
```diff
@@ -2184,3 +2184,47 @@ TEST_F(GraphOptz, dceBeforeOptimizeTranpose) {
...
+  K = F_->createChannelShuffle("channel_shuffle", K, 4, 1);
+  K = F_->createTranspose("unnecessary_transpose_2", K, {0, 2, 3, 1});
+  auto *save = F_->createSave("ret", K);
+
+  EXPECT_EQ(F_->getNodes().size(), 6);
+
+  // Optimize away the unnecessary transposes.
+  optimize(F_, CompilationMode::Infer);
+
+  // Ensure the two unnecessary transposes a gone.
```
jfix71: nit: `are gone`

jfix71:
Also, don't forget to squash all your commits.

lib/Optimizer/GraphOptimizer.cpp
jackm321: llvm::Optional, however if I remove the include it seems fine so I'll get rid of it

jfix71: Ah got it, you should keep it then, was just curious.

jackm321:
@jfix71 
Here are the times from running on my macbook. This doesn't seem much faster on cpu backend.

```
before optimization:

build_Release ./bin/image-classifier tests/images/imagenet/zebra_340.png -image_mode=0to1 -model=shufflenet/model.onnx -model_input_name=gpu_0/data_0 --time --cpu
Wall time per iteration (s): 0.0365
===-------------------------------------------------------------------------===
                         Miscellaneous Ungrouped Timers
===-------------------------------------------------------------------------===

   0.0354 (100.0%)   0.0011 (100.0%)   0.0365 (100.0%)   0.0365 (100.0%)  Infer
   0.0354 (100.0%)   0.0011 (100.0%)   0.0365 (100.0%)   0.0365 (100.0%)  Total

after optimization:

build_Release ./bin/image-classifier tests/images/imagenet/zebra_340.png -image_mode=0to1 -model=shufflenet/model.onnx -model_input_name=gpu_0/data_0 --time --cpu
Wall time per iteration (s): 0.0357
===-------------------------------------------------------------------------===
                         Miscellaneous Ungrouped Timers
===-------------------------------------------------------------------------===

   0.0346 (100.0%)   0.0010 (100.0%)   0.0356 (100.0%)   0.0357 (100.0%)  Infer
   0.0346 (100.0%)   0.0010 (100.0%)   0.0356 (100.0%)   0.0357 (100.0%)  Total
```

jfix71:
Huh, good to know, thanks for measuring. I imagine this will have a larger win on accelerators.

PR #2014, opened 2018-11-13
[common loader] Move Clip and SparseToDense to common parsing.

rdzhabarov:
*Description*:
Move Clip and SparseToDense to common parsing

*Testing*:
existing unit tests

*Documentation*:
n/a

cc: @yinghai


rdzhabarov:
> Do we already have caffe2ImporterTests for these? Worth adding tests for onnxImporterTests too?

We have tests for both already :)

PR #2015, opened 2018-11-13
[image] Factor out PNG processing

bertmaher:
*Description*: I'd like to use correctly-processed PNGs in a standalone utility, so I thought factoring this functionality out of the ImageClassifier executable would make sense.
*Testing*: ./tests/image/run.sh
*Documentation*: N/A
Part of #1976

### DIFF
--- a/include/glow/Base/Image.h
+++ b/include/glow/Base/Image.h
@@ -21,6 +21,26 @@
 #include <tuple>
 
 namespace glow {
+
+enum class ImageNormalizationMode {
+  kneg1to1,     // Values are in the range: -1 and 1.
+  k0to1,        // Values are in the range: 0 and 1.
+  k0to255,      // Values are in the range: 0 and 255.
+  kneg128to127, // Values are in the range: -128 .. 127
+};
+
+enum class ImageLayout {
+  NCHW,
+  NHWC,
+};
+
+enum class ImageChannelOrder {
+  BGR,
+  RGB,
+};
+
+std::pair<float, float> normModeToRange(ImageNormalizationMode mode);
+
 /// Reads a png image header from png file \p filename and \returns a tuple
 /// containing height, width, and a bool if it is grayscale or not.
 std::tuple<size_t, size_t, bool> getPngInfo(const char *filename);
@@ -40,6 +60,11 @@ bool writePngImage(Tensor *T, const char *filename,
                    std::pair<float, float> range,
                    bool useImagenetNormalization = false);
 
+Tensor readPngImageAndPreprocess(const std::string &filename,
+                                 ImageNormalizationMode imageNormMode,
+                                 ImageChannelOrder imageChannelOrder,
+                                 ImageLayout imageLayer,
+                                 bool useImagenetNormalization);
 } // namespace glow
 
 #endif // GLOW_BASE_IMAGE_H
--- a/include/glow/Base/Image.h
+++ b/include/glow/Base/Image.h
@@ -22,6 +22,7 @@
 
 namespace glow {
 
+/// Pixel value ranges.
 enum class ImageNormalizationMode {
   kneg1to1,     // Values are in the range: -1 and 1.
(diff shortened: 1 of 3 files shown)


include/glow/Base/Image.h
```diff
@@ -40,6 +60,11 @@ bool writePngImage(Tensor *T, const char *filename,
                    std::pair<float, float> range,
                    bool useImagenetNormalization = false);
 
+Tensor readPngImageAndPreprocess(const std::string &filename,
```
jfix71: nit: Doxygen!

tools/loader/ImageClassifier.cpp
```diff
@@ -153,68 +120,46 @@ llvm::cl::opt<bool> useImagenetNormalization(
...
-  // Get first image's dimensions and check if grayscale or color.
-  size_t imgHeight, imgWidth;
-  bool isGray;
-  std::tie(imgHeight, imgWidth, isGray) = getPngInfo(filenames[0].c_str());
-  const size_t numChannels = isGray ? 1 : 3;
-
-  // N x C x H x W
-  inputImageData->reset(ElemKind::FloatTy,
-                        {numImages, numChannels, imgHeight, imgWidth});
+  Tensor initial =
```
jfix71: Curious, why did you decide to move loading the first image separately to set the dims of `inputImageData`? Before we just called `getPngInfo()` on the first image in order to set the dimensions of `inputImageData`, and then loaded all images in a loop. I guess it's one less `fopen()` this way? If you prefer this way then I think we could actually get rid of `getPngInfo()` entirely...

lib/Base/Image.cpp
```diff
@@ -270,6 +286,48 @@ bool glow::writePngImage(Tensor *T, const char *filename,
...
+Tensor glow::readPngImageAndPreprocess(const std::string &filename,
+                                       ImageNormalizationMode imageNormMode,
+                                       ImageChannelOrder imageChannelOrder,
+                                       ImageLayout imageLayout,
+                                       bool useImagenetNormalization) {
+  // Get image dimensions and check if grayscale or color.
+  size_t imgHeight;
+  size_t imgWidth;
+  bool isGray;
+  std::tie(imgHeight, imgWidth, isGray) = getPngInfo(filename.c_str());
```
jfix71: I don't think it's necessary to call `getPngInfo()` here -- you can get `imgHeight`, `imgWidth`, and `numChannels` from the shape of `imageData` below.

tools/loader/ImageClassifier.cpp
```diff
@@ -153,68 +120,46 @@ llvm::cl::opt<bool> useImagenetNormalization(
...
-    inputImageData->transpose(&dataNHWC, NCHW2NHWC);
-    *inputImageData = std::move(dataNHWC);
-    break;
+  insertImage(IIDH, IH, 0);
+
+  for (size_t n = 1; n < filenames.size(); n++) {
+    Tensor localCopy = readPngImageAndPreprocess(filenames[n], imageNormMode,
+                                                 imageChannelOrder, imageLayout,
+                                                 useImagenetNormalization);
+    assert(std::equal(localCopy.dims().begin(), localCopy.dims().end(),
```
jfix71: nit: Can you do `localCopy.dims() == initial.dims()` instead?

include/glow/Base/Image.h
nadavrot: llvm::StringRef

tools/loader/ImageClassifier.cpp
bertmaher: Hm, now that I'm using getPngInfo, I no longer have `initial`, so I'm using `std::equal` with `batch.dims().begin() + 1`.  What do you think of that?  It's kind of weird, but I don't have a better idea :-/

jfix71: You could create a const array above with the expected shape `{imgHeight, imgWidth, numChannels}` and then compare here. NBD either way though.

PR #2016, opened 2018-11-13
[Quantization] Load quantized resnet50 model

### LINKED ISSUE #1762: Add quantized ResNet50 to Glow's test suite
Last week @beicy mentioned that Caffe2 has recently added a quantized ResNet50 to the model zoo: https://github.com/caffe2/models/tree/master/resnet50_quantized.  Once we support loading quantized models (#1727), we should add quantized resnet to our caffe2 model downloader (`utils/download_caffe2_models.sh`) and add it to `run.sh` to test it with `image-classifier`.

beicy:
*Description*:
This PR finished the rest work of loading Caffe2 quantized Resnet50 model discussed in #1727.
The model can be found here : https://github.com/caffe2/models/tree/master/resnet50_quantized
For testing in Glow side, the model can be downloaded by running utils/download_caffe2_models.sh.

*Testing*:
added quantized ResNet50 to Glow's test suite (run.sh), tested for interpreter and cpu backend.
```
 File: tests/images/imagenet/cat_285.png	Label-K1: 281 (probability: 0.7541)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9583)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9952)
```

*Documentation*:
[Optional Fixes #1762 #1727]


lib/Graph/VerifierHelper.cpp
```diff
@@ -69,8 +69,12 @@ bool glow::checkType(NodeValue A, ElemKind expectedType, const Node *parent) {
 
 bool glow::checkSameIsQuantized(const TypeRef A, const TypeRef B,
                                 const Node *parent) {
-  return expectCompareTrue("Mismatching isQuantized", A->isQuantizedType(),
-                           B->isQuantizedType(), parent);
+  bool ret = expectCompareTrue("Mismatching isQuantized", A->isQuantizedType(),
+                               B->isQuantizedType(), parent);
+  if (!ret) {
+    printf("now quantization has issue\n");
```
jackm321: I see `llvm::errs()` is used more frequently in the code, maybe prefer that?
Also this error message doesn't provide much information

lib/Graph/Graph.cpp
```diff
@@ -493,6 +493,23 @@ MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
   return addNode(new MaxPoolNode(name, OT, input, kernels, strides, pads));
 }
 
+MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
+                                     TypeRef outTy,
+                                     llvm::ArrayRef<unsigned_t> kernels,
+                                     llvm::ArrayRef<unsigned_t> strides,
+                                     llvm::ArrayRef<unsigned_t> pads) {
+  ShapeNHWC idim = ShapeNHWC(input.dims());
+  (void)idim;
```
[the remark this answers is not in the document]
jackm321: ohh ok thank you!

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -39,6 +39,15 @@ using llvm::cast;
...
     std::unordered_map<std::string, const caffe2::Argument *>;
 
+/// For the quantized Caffe2 ops, the activations are quantized to uint_8.
+/// In Glow, the activations are quantized to int_8. Therefore, for the offset
+/// read from quantized caffe2 model, we need to subtract 128(i.e. INT8_MIN) to
+/// make the activations becomes int8_t.
+/// For Glow: -127 <= orig_fp32/scale_1 + offset_1 < 128
+/// For Caffe2: 0 <= orig_fp32/scale_2 + offset_2 < 255
+/// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128
+#define OFFSETSHIFT 128
```
rdzhabarov: can you make a constant instead of `define`

lib/Graph/VerifierHelper.cpp
beicy: Thanks! Actually, these lines are used for testing and should be removed :)

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -159,34 +169,18 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     Tensor *w = getTensorByName(op.input(1));
 
     // Transpose the weights to the right format. Glow expects to read the
-    // weights in the format CRSK. Caffe2 stores the operators as KCRS.
-    // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
+    // weights in the format NHWC.
```
nadavrot: I think that we should leave the comment that explains the weight layout.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -159,34 +169,18 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
     Tensor wtag;
-    w->transpose(&wtag, NCHW2NHWC);
+    if (typeName != "Conv" && order == "NHWC")
+      wtag.assign(w);
+    else
```
nadavrot: Can you explain in a comment what's going on here? Also, please add braces around the statements.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -159,34 +169,18 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      w->transpose(&wtag, NCHW2NHWC);
 
     // The structure of the conv weigts is: NHWC. We take the C, which is the
     // number of filters. We use this value to calculate the size of the bias
     // if it is not specified.
     size_t depth = wtag.dims()[0];
 
-    // Construct the Filter field.
```
nadavrot: Why did we remove this part of the code?

nadavrot:
Nice!

lib/Importer/Caffe2ModelLoader.cpp
beicy: Rewrite it later based on the filter's type -- could be int8 or float type

lib/Graph/Nodes.cpp
```diff
@@ -199,7 +199,8 @@ static bool verifyPool(NodeValue src, NodeValue dest,
   ShapeNHWC exp(idim.n, outSz.first, outSz.second, idim.c);
   isValid &=
       expectCompareTrue("Unexpected output dimensions", exp, odim, parent);
-  isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());
+  if (!src.getType()->isQuantizedType())
```
jfix71: The reason you added this is because pools may have different `scale`/`offset` for the `src` and `dest`, correct? Can you instead do `checkSameIsQuantized()` if `src` is quantized?

utils/download_caffe2_models.sh
```diff
@@ -26,11 +26,12 @@ vgg19
 zfnet512
 bvlc_alexnet
 en2gr
+quant_resnet50
 EOF
 )
 
 for model in $MODELS; do
-  for file in predict_net.pbtxt predict_net.pb init_net.pb; do
+  for file in predict_net.pbtxt predict_net.pb init_net.pb init_net.pbtxt; do
```
jfix71: not all models have `init_net.pbtxt` -- won't this cause cause `wget` to fail? And why do we need `init_net.pbtxt` instead of just using `init_net.pb`?

tests/images/run.sh
```diff
@@ -41,3 +41,5 @@ done
 for png_filename in tests/images/imagenet_299/*.png; do
   ./bin/image-classifier "$png_filename" -image_mode=0to1 -m=googlenet_v4_slim/googlenet_v4_slim.onnx -model_input_name=input:0 -image_layout=NHWC -label_offset=1 "$@"
 done
+#Quantized Resnet50 Caffe2 model test
+./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=quant_resnet50/predict_net.pbtxt -m=quant_resnet50/init_net.pbtxt  -model_input_name=gpu_0/data_0 -use-imagenet-normalization "$@"
```
jfix71: nit: add a new line at the end of the file

lib/Importer/Caffe2ModelLoader.cpp
beicy: @rdzhabarov In the quantized mode, for Int8MaxPool, the scale of input and output are the same. In Caffe2 implantation,  for INt8MaxPool, even if there is a pre-chosen quantization parameters for the output, it is ignored because maxpool output quantization should be same as the input. So I don't think we need to fix our backend implementation :)

rdzhabarov: Hm, i'm a bit confused. You modified the verifiers for pools
> isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());

This used to check that scale and offset is the same, not it does not.
Missing anything?

beicy: @rdzhabarov Actually, this one was triggered by Int8AvgPool, so  I modified it.

rdzhabarov: Both maxpool and avgpool would likely change input's floating point dynamic range.
I do not think it's correct to assume that inputs/outputs have the same scale.
However, these changes are not directly tight to this PR as it was the case we had this problem before.

Can you file issue to fix int8 int/max pool and fix related code (quantization procedure, loading of c2, etc).

rdzhabarov:
cc: @jspark1105

lib/Importer/Caffe2ModelLoader.cpp
beicy: For quantized avgpool, our current implantation does't require the input and output has the same scale and offset. So you mean for Int8MaxPool, we need to handle the input and output with different scale?

rdzhabarov: Yes, both operations need to handle diff {scale,offset} for input/output.

beicy: After discussed offline with @rdzhabarov, for maxpool, we will keep the quantized input and output have the same params.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -159,34 +169,22 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
 
     // Transpose the weights to the right format. Glow expects to read the
-    // weights in the format CRSK. Caffe2 stores the operators as KCRS.
+    // weights in the format CRSK.
     // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
+    // Caffe2 "Conv" op always stores the weight as KCRS, while for "Int8Conv",
+    // and "Int8ConvRelu", the weights always follows the "order" arg.
     Tensor wtag;
-    w->transpose(&wtag, NCHW2NHWC);
+    if (typeName != "Conv" && order == "NHWC") {
```
artemrakhov: Only NHWC order is supported by Int8Conv:
https://github.com/pytorch/pytorch/blob/9cb4bce847af688ca344caaa876d3c47a36f28bf/caffe2/operators/quantized/int8_conv_op.h#L25

Also, are you sure that "order" arg applies to weights too, not just input?
Reading the docs, it seems to me that weight always has to be transposed, because it's outC x inC x h x w
https://github.com/pytorch/pytorch/blob/9cb4bce847af688ca344caaa876d3c47a36f28bf/caffe2/operators/quantized/int8_conv_op.cc#L49

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -214,7 +279,49 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return;
   }
 
-  if (typeName == "MaxPool" || typeName == "AveragePool") {
+  if (typeName == "Int8SumRelu") {
```
artemrakhov: Can we make Caffe2 automatically unfuse this op for us? So we'd only support Int8Sum and Int8Relu?

artemrakhov: > Therefore, they don't really need "Int8Relu".

My understanding is that there is real range of numbers in the input activation. Let's say it is [-5.5, +5.5]. In Glow when we quantize it, we'll use the whole range of [-128, +127]. So our Relu will do `x' = max(x, 0)`. But in Caffe2 they use the whole range of [0, 255]. So their Relu will do `x' = max(x, 128)`. Essentially if input range is symmetric, then Relu needs to "cut" half of it. It's not possible to throw Relu away. Please correct me if I'm wrong.

> If we want the unfused op, the original fp32 model should be rewritten and re-quantized.

Why? You do unfuse here and keep the same quantization parameters.

> decomposing Int8SumRelu is not complicated

All decompositions are simple, but if we have too many of them, they could be maintenance burden.

lib/Graph/Graph.cpp
```diff
@@ -493,6 +493,23 @@ MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
   return addNode(new MaxPoolNode(name, OT, input, kernels, strides, pads));
 }
 
+MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
+                                     TypeRef outTy,
+                                     llvm::ArrayRef<unsigned_t> kernels,
+                                     llvm::ArrayRef<unsigned_t> strides,
+                                     llvm::ArrayRef<unsigned_t> pads) {
+  ShapeNHWC idim = ShapeNHWC(input.dims());
```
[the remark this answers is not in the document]
beicy: Created a new PR for this : [trivial] Rewrote some duplicated code. #2034

include/glow/Graph/Graph.h
```diff
@@ -260,6 +270,9 @@ class Function final : public Named {
                                            NodeValue input, Storage *W,
                                            Storage *B);
 
+  FullyConnectedNode *createFullyConnected(llvm::StringRef name, TypeRef outTy,
+                                           NodeValue input, Storage *W,
+                                           Storage *B);
   /// Create a fully connected node with the specified output type.
   /// Note, outputDepth is infered based on the output type.
   FullyConnectedNode *createFullyConnected(llvm::StringRef name,
```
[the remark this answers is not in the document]
beicy: Used the existing one.

lib/Graph/Graph.cpp
```diff
@@ -494,6 +494,16 @@ MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
   return addNode(new MaxPoolNode(name, OT, input, kernels, strides, pads));
 }
 
+MaxPoolNode *Function::createMaxPool(llvm::StringRef name, NodeValue input,
+                                     TypeRef outTy,
```
qcolombet: Something to double check but I am not sure the interpreter would work properly with an outTy that have different scale and offset as the input type.

beicy: We make sure that the output scale/offset is the same with the input.

include/glow/Graph/Graph.h
```diff
@@ -243,6 +243,11 @@ class Function final : public Named {
                              llvm::ArrayRef<unsigned_t> strides,
                              llvm::ArrayRef<unsigned_t> pads);
 
+  MaxPoolNode *createMaxPool(llvm::StringRef name, NodeValue input,
```
rdzhabarov: since we enforce output type to be the same as input type for maxPool, i think we can just use `createMaxPool` method above (no need to overload it).

beicy: updated. Thanks!

lib/Graph/Nodes.cpp
```diff
@@ -200,7 +200,11 @@ static bool verifyPool(NodeValue src, NodeValue dest,
   ShapeNHWC exp(idim.n, outSz.first, outSz.second, idim.c);
   isValid &=
       expectCompareTrue("Unexpected output dimensions", exp, odim, parent);
-  isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());
+  if (src.getType()->isQuantizedType()) {
+    isValid &= checkSameIsQuantized(src.getType(), dest.getType(), parent);
```
rdzhabarov: this check is a bit relaxed. For maxpool we'd like to check that type with ignored shape match as well (not only that the input/output is quantized).

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -201,10 +199,70 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+          Tensor *b = getTensorByName(biasTensorName);
+          biasTensor.assign(b);
+        }
+      }
+      outTy = G_.getParent()->uniqueType(ElemKind::FloatTy, outDims);
+      filter = G_.getParent()->createConstant("conv.filter", wtag);
+      bias = G_.getParent()->createConstant("conv.bias", biasTensor);
+    } else {
+      assert(dict.count("Y_zero_point") &&
+             "missing zero point for quantzied type");
```
rdzhabarov: can you be more explicit that the problem is with zero point for the `bias` param

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -201,10 +199,70 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      bias = G_.getParent()->createConstant(
+          ElemKind::Int32QTy, biasTensor.dims(),
+          biasTensor.getType().getScale(), biasTensor.getType().getOffset(),
+          "conv.bias");
+      bias->assign(&biasTensor);
+    }
 
     Node *node = G_.createConv(opName, tr, filter, bias, outTy, kernels,
                                strides, pads, group);
+    if (typeName == "Int8ConvRelu") {
```
rdzhabarov: similarly how you handle `Int8SumRelu` we should be able to handle Int8ConvRelu without additional Relu node.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -214,7 +272,45 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+    assert(dict.count("Y_zero_point") &&
+           "missing zero point for quantzied type");
+    assert(dict.count("Y_scale") && "missing Y_scale for quantzied type");
+    auto in0 = getNodeValueOrCreateConstantByName(op.input(0));
+    auto in1 = getNodeValueOrCreateConstantByName(op.input(1));
+    auto outDims = in0.getType()->dims();
+    auto outTy = G_.getParent()->uniqueType(
+        ElemKind::Int8QTy, outDims, loadFloat(dict["Y_scale"]),
+        loadInt(dict["Y_zero_point"]) - OFFSETSHIFT);
+    auto *node = G_.createAdd("tempadd", outTy, in0, in1);
```
rdzhabarov: `tempadd` ? why not op name?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,7 +334,28 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
 
     Node *node = nullptr;
-    if (typeName == "MaxPool") {
+
+    if (typeName == "Int8MaxPool" || typeName == "Int8AveragePool") {
+      // Create the node with quantized type.
+      assert(dict.count("Y_zero_point") &&
+             "missing zero point for quantzied type");
+      assert(dict.count("Y_scale") && "missing Y_scale for quantzied type");
+      ShapeNHWC idim = ShapeNHWC(tr->getType(0)->dims());
```
rdzhabarov: it would be less error prone to get result and type from the result and avoid explicit indexing?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,7 +334,28 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     }
 
     Node *node = nullptr;
-    if (typeName == "MaxPool") {
+
+    if (typeName == "Int8MaxPool" || typeName == "Int8AveragePool") {
```
rdzhabarov: do you need special casing for Int8MaxPool? I think `typeName == "MaxPool"` should handle both cases nicely. Anything i'm missing?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -602,6 +739,72 @@ void Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
...
+      // type of input and weights must be the same. Therefore, we need to
+      // convert it to int8 by subtracting 128.
+      T->reset(ElemKind::Int8QTy, dim, scale, offset - OFFSETSHIFT);
+      auto TH = T->getHandle<int8_t>();
+      std::string str = dict["values"]->s();
+      for (; i < str.size(); i++) {
+        TH.raw(i) = ((uint8_t)(str.c_str()[i]) - OFFSETSHIFT);
+      }
+    } else {
+      T->reset(ElemKind::Int32QTy, dim, scale, offset);
```
rdzhabarov: what is the `typeName` for which we set tensor type to be Int32?

tests/images/run.sh
```diff
@@ -41,3 +41,5 @@ done
 for png_filename in tests/images/imagenet_299/*.png; do
   ./bin/image-classifier "$png_filename" -image_mode=0to1 -m=googlenet_v4_slim/googlenet_v4_slim.onnx -model_input_name=input:0 -image_layout=NHWC -label_offset=1 "$@"
 done
+#Quantized Resnet50 Caffe2 model test
+./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=quant_resnet50 -model_input_name=gpu_0/data_0 -use-imagenet-normalization "$@"
```
rdzhabarov: where do we get model from? Likely we need to update utils script for downloading the model

lib/Importer/Caffe2ModelLoader.cpp
beicy: Yes. They are the same. However, for "Int8MaxPool", it has the ""Y_zero_point" and "Y_scale". I would like to combine it with Int8AveragePool to check if the fields are there (although for Int8MaxPool, those fields can be ignored).

tests/images/run.sh
beicy: Yes, the download_caffe2_models.sh is included in this PR.

lib/Importer/Caffe2ModelLoader.cpp
beicy: Yes.

tests/images/run.sh
rdzhabarov: missed that :(

lib/Importer/Caffe2ModelLoader.cpp
rdzhabarov: `Int8GivenIntTensorFill` this feels a bit strange as Int range is bigger than Int8 range. But since it's what's on C2 we probably cannot do anything.

lib/Importer/Caffe2ModelLoader.cpp
beicy: `tr` could either be the input node directly or a node with Transposed type. So I think it is easier to get the type directly without casting.

lib/Importer/Caffe2ModelLoader.cpp
beicy: It has been explained before defining OFFSETSHIFT if it is what you mean:).
```
/// For the quantized Caffe2 ops, the activations are quantized to uint_8.                                                                                                    
/// In Glow, the activations are quantized to int_8. Therefore, for the offset                                                                                                
/// read from quantized caffe2 model, we need to subtract 128(i.e. INT8_MIN) to                                                                                               
/// make the activations becomes int8_t.                                                                                                                                      
/// For Glow: -127 <= orig_fp32/scale_1 + offset_1 < 128                                                                                                                      
/// For Caffe2: 0 <= orig_fp32/scale_2 + offset_2 < 255                                                                                                                       
/// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128                                                                                                   
const int32_t OFFSETSHIFT = 128
```;

Comment (rdzhabarov) [reply lib/Importer/Caffe2ModelLoader.cpp]: yeah, i was referring to `"missing zero point for quantzied type"`
just to clarify that it's for the output type etc.

Comment (beicy) [reply lib/Importer/Caffe2ModelLoader.cpp]: Thanks! :)

File: lib/Graph/Nodes.cpp (hunk 16 above)
Comment (rdzhabarov) [reply lib/Graph/Nodes.cpp]: slightly simplified comparison.
both needs to be either quantized or not quantized and in case of max pool we'd need to check matching type except shape.

``` cpp
isValid &= checkSameIsQuantized(src.getType(), dest.getType(), parent);

if (!isAvgPool) {
  isValid &= checkTypeIgnoreShape(src, dest, parent);
}
```

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -201,7 +199,60 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      filter = G_.getParent()->createConstant("conv.filter", wtag);
+      bias = G_.getParent()->createConstant("conv.bias", biasTensor);
+    } else {
+      assert(dict.count("Y_zero_point") &&
+             "missing zero point for quantized output type");
+      assert(dict.count("Y_scale") &&
+             "missing Y_scale for quantized output type");
+      // Construct the Bias field.
+      Tensor biasTensor(ElemKind::Int32QTy, {depth}, 1.0, 0);
+      biasTensor.zero();
```
rdzhabarov: this is fine, as we have offset equal to 0, but isZero and zero does not properly work for quantized types currently.
Tracking this here: https://github.com/pytorch/glow/issues/2082

lib/Graph/Nodes.cpp
```diff
@@ -200,7 +200,14 @@ static bool verifyPool(NodeValue src, NodeValue dest,
   ShapeNHWC exp(idim.n, outSz.first, outSz.second, idim.c);
   isValid &=
       expectCompareTrue("Unexpected output dimensions", exp, odim, parent);
-  isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());
+  if (src.getType()->isQuantizedType() && isAvgPool) {
+    // For quantized AvgPool, the scale and offset of its input and output could
```
nadavrot: Is this documented anywhere? Like, in the doc string of the node?

beicy: No, it hasn't been documented yet. However, after this PR is merged, we need to update our quantization docs.

lib/Importer/Caffe2ModelLoader.cpp
beicy: Got it. Will improve it in the following PR.

beicy:
@rdzhabarov Is that OK to create a separate PR for the net testing? Since I think the unittest for each quantized op should be added as well.

rdzhabarov:
> Since I think the unittest for each quantized op should be added as well.

We already have tests for each quantized op (see operatorTest).


The only thing which is missing is loader tests and since we do not run resnet50 quantized.
This PR is getting bigger I'm ok to schedule followup to address these concerns.

lib/Importer/Caffe2ModelLoader.cpp
rdzhabarov: Nick was going to look at the issue above.

beicy:
@rdzhabarov I agree. Will test the Caffe2 loader for the now quantized caffe2 op.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -201,7 +199,60 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+    Constant *filter;
+    Constant *bias;
+    if (typeName == "Conv") {
+      // Construct the Bias field.
+      Tensor biasTensor(ElemKind::FloatTy, {depth});
+      biasTensor.zero();
+
+      // Check if we have a serialized bias vector.
+      if (op.input_size() > 2) {
+        auto &biasTensorName = op.input(2);
```
rdzhabarov: nit: const auto &

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -201,7 +199,60 @@ void Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      assert(dict.count("Y_zero_point") &&
+             "missing zero point for quantized output type");
+      assert(dict.count("Y_scale") &&
+             "missing Y_scale for quantized output type");
+      // Construct the Bias field.
+      Tensor biasTensor(ElemKind::Int32QTy, {depth}, 1.0, 0);
+      biasTensor.zero();
+      // Check if we have a serialized bias vector.
+      if (op.input_size() > 2) {
+        auto &biasTensorName = op.input(2);
```
rdzhabarov: ditto

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -602,6 +736,73 @@ void Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
     return;
   }
 
+  // Load tensors and quantize the values:
```
rdzhabarov: we do not really quantize values here, right. Just loading already quantized tensors.

utils/download_caffe2_models.sh
```diff
@@ -26,6 +26,7 @@ vgg19
 zfnet512
 bvlc_alexnet
 en2gr
+quant_resnet50
```
rdzhabarov: kinda unrelated question but why don't we download directly from the c2 zoo?

beicy: Their model has a problem: for their init_net.pb/init_net.pbtxt file, it has duplicated tensors which is unexpected. I removed the duplicated tensors in our version. According to the model owner, they will update the model on their side in the next release.

Issue #1979, opened 2018-11-08
[wishlist] Migrate bundle building from make to cmake


bertmaher:
We have makefiles for the bundle examples, but there's no reason (as far as I can see) we couldn't create real cmake targets for them, which would make the workflow way more accessible.  I've automated some slightly different workflows this way, so happy to give advice to anyone who wants to try this.

nadavrot:
The intent was to give people a small example that they can copy into their own programs. I think that @opti-mix first implemented this in a small shell script.

SplitInfinity:
I decided to do this since I need learn cmake anyway for one of my other big tasks. I can't figure out how to get `add_executable` to pick up and link against an object file built using `add_custom_command` (i.e. the bundle) and well as dynamically link against `libpng`. Once I have have that, I will be able to post a PR.

cc: `cmake` master @compnerd

SplitInfinity:
Never mind, figured everything out and posted a PR!

bertmaher:
@nadavrot I was thinking about your comment on this issue, and I think it might be valuable to have a Makefile example bundle and a CMake example.  The CMake stuff in particular is non-trivial, and since CMake is pretty popular I could see people appreciating having an example of how to seamlessly integrate Glow.

nadavrot:
@bertmaher @SplitInfinity I see the value of having tests that actually verify that a feature that we care about (AOT compilation) works. In that case, we absolutely need CMake file support.

bertmaher:
Probably want to update docs/AOT.md as well.

SplitInfinity:
I'll send another PR to update the AOT docs, but let's also use this issue instead of the PR to discuss what to do with the other bundles and Makefiles. 

I think we should add cmake support for all of the bundles and get rid of any that don't add additional value. If I understood @bertmaher 's comments on the PR correctly, the resnet50 bundle is sufficiently recent and complex to serve as a good example of how to make a standalone bundle, and the lenet-mnist bundle is small enough to run in CI. So maybe we should get rid of the other two, and add cmake support for lenet-mnist?

bertmaher:
I'm in favor of that approach (convert lenet_mnist to cmake and delete vgg19/zfnet512).

The more I've thought about it the less I like having a split between CMake and Make, and if I had to pick one for an example I'd choose CMake.  (My reasoning: getting this working in make is pretty easy, because make is very simple.  CMake is harder, so if someone's going to run into a wall with glow it'll be because they can't figure out the CMake part)

PR #2018, opened 2018-11-14
[examples] Add cmake for ResNet50 bundle

SplitInfinity:
**Description**
This commit adds the capability to build the ResNet50 standalone bundle
using `cmake` in addition to `make`. Quantization is controlled for all
bundles using the `GLOW_QUANTIZE_BUNDLES` `cmake` variable.

**Testing**
Configured fresh build and ran bundle running cmake target:

```
$ ninja RunResNet50Bundle -v
...
...
...
Allocated weights of size: 102121984
Expected weights of size: 102121984
Loaded weights of size: 102121984 from the file resnet50.weights
Allocated mutable weight variables of size: 606144
Loaded image: {...}/cat_285.png
Loaded image: {...}/dog_207.png
Loaded image: {...}/zebra_340.png
Loaded images size in bytes is: 1806336
Copying image data into mutable weight vars: 1806336 bytes
Result: 285
```

**Fixes**
Part of issue #1979.


SplitInfinity:
Travis fails because I'm using a feature from `cmake 3.11+`. Below that version, you're not allowed to use `add_executable` without providing any sources. I more or less translated the Makefile for the ResNet bundle, so I used this feature for the executable target that takes in the driver and bundle object files as input and produces the final executable.

examples/bundles/resnet50/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+    RunBundleCommand
+  COMMAND
+    ResNet50Bundle ${IMAGES}/*.png
+  DEPENDS
+    ResNet50Bundle
+)
+add_custom_target(RunResNet50Bundle DEPENDS RunBundleCommand ResNet50Bundle)
+
+# Final executable
+add_executable(ResNet50Bundle)
```
bertmaher: Would it work to declare this as
```
add_executable(ResNet50Bundle $<TARGET_OBJECTS:ResNet50BundleMain>)
```

bertmaher:
This is really nice work, thanks for digging into CMake and figuring this out.  It's pretty magical to have the bundle rebuild automatically when glow changes, and not have to manually supply GLOW_SRC and other environment variables.

bertmaher:
How would you feel about making separate targets for quantized and non-quantized bundles, rather than conditionally compiling based on a config parameter?  I could see wanting to build/test both without reconfiguring my build tree.

SplitInfinity:
Hmm, after separating out the quantized and non-quantized targets and taking @bertmaher 's suggestion on how to make the commit compatible with `cmake` 3.6, Travis fails with this:

```
[114/171] Generating resnet50.o
FAILED: cd /home/travis/build/pytorch/glow/build/examples/bundles/resnet50 && /home/travis/build/pytorch/glow/build/bin/image-classifier /home/travis/build/pytorch/glow/tests/images/imagenet/dog_207.png -g -image_mode=0to1 -m=/home/travis/build/pytorch/glow/build/examples/bundles/resnet50/resnet50 -model_input_name=gpu_0/data -cpu -emit-bundle /home/travis/build/pytorch/glow/build/examples/bundles/resnet50
image-classifier: Unknown command line argument '-g'.  Try: '/home/travis/build/pytorch/glow/build/bin/image-classifier -help'
image-classifier: Did you mean '-i'?
```

`image-classifier` definitely has a `-g` flag. Has anyone seen this before?

SplitInfinity:
It turns out `-g` is a `CPUBackend` flag and Travis compiles with that disabled. I've gated the bundle builds on `GLOW_WITH_CPU` to address this.

SplitInfinity:
Okay, I think this PR is ready for review. In particular, I'd like feedback on:
- CMakefile style and convention (if only `clang-format` took care of these too!)
- mode-specific build instructions? (e.g. do not allow quantization in DEBUG mode because it takes forever)

Once everything looks good, I'll migrate the other bundles either in this PR or a separate PR depending on what everyone prefers.

rdzhabarov:
Let's disable bundles by default it takes 40mins to run on CI: https://github.com/pytorch/glow/pull/2021

PR #2019, opened 2018-11-14
[GraphOptimizer] Fuse RescaleNodes down into MatMulNodes

jfix71:
*Description*: Found this in quantized en2gr, similar to other optimizations we already have.

*Testing*: Added unit test

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -2066,6 +2066,27 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       }
       continue;
     }
+
+    // Combine Rescale down with MatMul node.
+    //   MatMulNode(Rescale(LHS), RHS) -> MatMulNode(LHS', RHS).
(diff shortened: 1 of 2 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2066,6 +2066,27 @@ static bool sinkRescaleQuantizedNode(Function *F) {
...
+        MN = newMN;
+        changed = true;
+      }
+      if (auto *rescale = dyn_cast<RescaleQuantizedNode>(MN->getRHS())) {
+        auto *newMN = F->createMatMul(MN->getName(), MN->getResult().getType(),
+                                      MN->getLHS(), rescale->getInput());
+        MN->getResult().replaceAllUsesOfWith(newMN);
+        changed = true;
+      }
+      continue;
```
jackm321: this continue will get hit even if nothing changed, is that correct?

jackm321: oh nevermind I see it's fine :)

Issue #1599, opened 2018-09-08
AllocationsInfo allocates buffers for all module variables.


nadavrot:
AllocationsInfo allocates buffers for all variables, and not only for variables that are used by the function. I think that this is okay for the JIT, but not so great for the AOT compilation because the unrelated variables become a part of the interface. 

For more details see: AllocationsInfo::allocateWeightVars.

opti-mix:
Other backends (e.g. OpenCL) have this issue as well and we'd better fix it there as well.

May be we should have a helper method or a filtering iterator that yields only variables used by a specific Function? Then we'd need to allocate buffers only for the variables returned by such a helper method or iterator.

nickgg:
@opti-mix Is this still an issue?

opti-mix:
Yes, I think it is not fixed yet.

PR #2024, opened 2018-11-14
[BackendUtils]Moved generateRuntimeBundle to BackendUtils.

### LINKED ISSUE #1988: Move generateRuntimeBundle to BackendUtils.
GenerateRuntimeBundle has sufficient overlap between backends that it should be generalized and moved to BackendUtils. Generalize generateRuntimeBundles, and update Interpreter, CPU, and OpenCL backends to use new library function.

gcatron:
*Description*: Moved generateRuntimeBundle to BackendUtils and updated backend to use it.
*Testing*: ninja test and run.sh are good.
*Documentation*: NA
Fixes #1988

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -38,7 +38,7 @@ getValueOffset(Value *v,
 /// Computes offsets and total allocation for Constants, Placeholders, and
 /// Activations to build runtime symbol table. Returns RuntimeBundle.
 runtime::RuntimeBundle
-generateRuntimeBundle(const IRFunction *F, MemoryAllocator *constantAllocator,
+generateRuntimeBundle(const IRFunction &F, MemoryAllocator *constantAllocator,
                       MemoryAllocator *placeholderAllocator,
                       MemoryAllocator *activationsAllocator);
 } // end namespace glow
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -38,9 +38,9 @@ getValueOffset(Value *v,
 /// Computes offsets and total allocation for Constants, Placeholders, and
 /// Activations to build runtime symbol table. Returns RuntimeBundle.
 runtime::RuntimeBundle
-generateRuntimeBundle(const IRFunction &F, MemoryAllocator *constantAllocator,
-                      MemoryAllocator *placeholderAllocator,
-                      MemoryAllocator *activationsAllocator);
+generateRuntimeBundle(const IRFunction &F, MemoryAllocator &constantAllocator,
+                      MemoryAllocator &placeholderAllocator,
+                      MemoryAllocator &activationsAllocator);
 } // end namespace glow
 
 #endif // GLOW_BACKENDS_BACKENDUTILS_H
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_BACKENDUTILS_H
 
 #include "glow/Backends/CompiledFunction.h"
+#include "glow/CodeGen/MemoryAllocator.h"
 #include "glow/IR/IR.h"
 
 namespace glow {
@@ -34,6 +35,12 @@ size_t
 getValueOffset(Value *v,
                const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
                    &symbolTable);
+/// Computes offsets and total allocation for Constants, Placeholders, and
+/// Activations to build runtime symbol table. Returns RuntimeBundle.
+runtime::RuntimeBundle
+generateRuntimeBundle(const IRFunction *F, MemoryAllocator *constantAllocator,
+                      MemoryAllocator *placeholderAllocator,
+                      MemoryAllocator *activationsAllocator);
 } // end namespace glow
 
 #endif // GLOW_BACKENDS_BACKENDUTILS_H
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -57,12 +57,12 @@ size_t glow::getValueOffset(
 }
 
 runtime::RuntimeBundle
-glow::generateRuntimeBundle(const IRFunction *F,
+glow::generateRuntimeBundle(const IRFunction &F,
                             MemoryAllocator *constantAllocator,
                             MemoryAllocator *placeholderAllocator,
                             MemoryAllocator *activationsAllocator) {
   // Handle Constants, Placeholders, and Activations, in that order.
-  /// Symbol table mapping symbol name to offset for runtime.
+  // Symbol table mapping symbol name to offset for runtime.
   std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
   uint64_t constantMaxSize = 0;
   uint64_t placeholderMaxSize = 0;
@@ -70,9 +70,9 @@ glow::generateRuntimeBundle(const IRFunction *F,
 
   // Compute the offsets for Constants.
   if (constantAllocator) {
-    for (auto &v : F->getGraph()->getParent()->getConstants()) {
-      assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
-      auto *w = cast<WeightVar>(F->getWeightForNode(v));
+    for (auto &v : F.getGraph()->getParent()->getConstants()) {
+      assert(isa<WeightVar>(F.getWeightForNode(v)) && "Expected WeightVar");
+      auto *w = cast<WeightVar>(F.getWeightForNode(v));
       auto numBytes = w->getSizeInBytes();
       size_t addr = constantAllocator->allocate(numBytes, v);
       runtime::RuntimeSymbolInfo symbol;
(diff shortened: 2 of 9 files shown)


include/glow/Backends/BackendUtils.h
```diff
@@ -34,6 +35,13 @@ size_t
 getValueOffset(Value *v,
                const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
                    &symbolTable);
+/// Computes offsets and total allocation for Constants, Placeholders, and
+/// Activations to build runtime symbol table. Returns
```
rdzhabarov: nit: \returns ...

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
   return it->second.offset;
 }
+
+runtime::RuntimeBundle
+glow::generateRuntimeBundle(const IRFunction *F,
+                            MemoryAllocator *constantAllocator,
+                            MemoryAllocator *placeholderAllocator,
+                            MemoryAllocator *activationsAllocator) {
+  // Handle Constants, Placeholders, and Activations, in that order.
+  /// Symbol table mapping symbol name to offset for runtime.
```
rdzhabarov: nit: //

include/glow/Backends/BackendUtils.h
```diff
@@ -34,6 +35,12 @@ size_t
 getValueOffset(Value *v,
                const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
                    &symbolTable);
+/// Computes offsets and total allocation for Constants, Placeholders, and
+/// Activations to build runtime symbol table. Returns RuntimeBundle.
+runtime::RuntimeBundle
+generateRuntimeBundle(const IRFunction *F, MemoryAllocator *constantAllocator,
+                      MemoryAllocator *placeholderAllocator,
+                      MemoryAllocator *activationsAllocator);
```
qcolombet: I assume the function cannot be nullptr, so let us use a reference here.
Ditto for the memory allocator where it applies.

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+    }
+    activationsMaxSize = activationsAllocator->getMaxMemoryUsage();
+  }
+
+  runtime::RuntimeBundle info(constantMaxSize, placeholderMaxSize,
+                              activationsMaxSize);
+  info.symbolTable = std::move(symbolTable);
+  info.constants = collectConstants(F, constantMaxSize, info.symbolTable);
+  return info;
+}
```
qcolombet: Missing newline.

include/glow/Backends/BackendUtils.h
qcolombet: Does the memory allocator or bundle have a notion of out-of-memory?
I.e., where would we fail if we try to allocate a bundle that won't fit on the HW?

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+      auto numBytes = w->getSizeInBytes();
+      size_t addr = constantAllocator->allocate(numBytes, v);
+      runtime::RuntimeSymbolInfo symbol;
+      symbol.size = numBytes;
+      symbol.offset = addr;
+      symbol.type = *w->getType();
+      symbolTable.emplace(std::string(v->getName()), symbol);
+    }
+    constantMaxSize = constantAllocator->getMaxMemoryUsage();
+  }
```
qcolombet: What happen if the function has constants and we haven't provided a memory allocator?

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+      auto *w = cast<WeightVar>(F->getWeightForNode(v));
+      auto numBytes = w->getSizeInBytes();
+      size_t addr = constantAllocator->allocate(numBytes, v);
+      runtime::RuntimeSymbolInfo symbol;
+      symbol.size = numBytes;
+      symbol.offset = addr;
+      symbol.type = *w->getType();
+      symbolTable.emplace(std::string(v->getName()), symbol);
+    }
+    constantMaxSize = constantAllocator->getMaxMemoryUsage();
```
qcolombet: We should only allocate the constants that are used by a given function, not all the constants of the module.

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+  uint64_t constantMaxSize = 0;
+  uint64_t placeholderMaxSize = 0;
+  uint64_t activationsMaxSize = 0;
+
+  // Compute the offsets for Constants.
+  if (constantAllocator) {
+    for (auto &v : F->getGraph()->getParent()->getConstants()) {
+      assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
+      auto *w = cast<WeightVar>(F->getWeightForNode(v));
+      auto numBytes = w->getSizeInBytes();
```
qcolombet: I would expect this size to be backend specific (i.e., because of alignment constraints on the row and so on.)
Where do we account for that?

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+      size_t addr = placeholderAllocator->allocate(numBytes, w);
+      runtime::RuntimeSymbolInfo symbol;
+      symbol.offset = addr;
+      symbol.size = numBytes;
+      symbol.type = *w->getType();
+      symbolTable.emplace(std::string(v->getName()), symbol);
+    }
+    placeholderMaxSize = placeholderAllocator->getMaxMemoryUsage();
+  }
```
qcolombet: Same comments as the constants but for placeholder :).

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,105 @@ size_t glow::getValueOffset(
...
+        symbolTable.emplace(std::string(TV->getName()), symbol);
+        continue;
+      }
+
+      if (auto *D = llvm::dyn_cast<DeallocActivationInst>(&I)) {
+        auto *A = D->getAlloc();
+        assert(symbolTable.count(std::string(A->getName())) &&
+               "Invalid deallocation!");
+        activationsAllocator->deallocate(A);
+        continue;
```
qcolombet: I am confused by this. I don't know the details of the memory allocator class, but the comment reads "Frees the allocation associated with \p handle.". Therefore it seems strange that we would free something that we haven't run yet.
Could you add a comment explaining what this is doing?

include/glow/Backends/BackendUtils.h
gcatron: Currently we assume it will fit at compile time. We would fail at runtime allocation if there is insufficient memory.

lib/Backends/BackendUtils.cpp
gcatron: We don't do allocation for constants if a constant allocator is not provided.

qcolombet: At the very least we should report an error if we don’t provide a constant allocator whereas there are some constants in the function.

Another thing is do we need the flexibility of having those allocators optional?

gcatron: Probably not, @opti-mix and I were discussing this as well. I think they will be required references. It it doesn't get used that's fine.

qcolombet: Agree

lib/Backends/BackendUtils.cpp
gcatron: The allocator has some logic as far as alignment goes. For anything more involved we would likely want to introduce a more featured allocator.

qcolombet: Not something to fix here, more thoughts for food, why we split the different allocator per kind of data?
Basically I believe we would need something syncing the allocators to check that the whole thing fits, thus I wonder if we shouldn’t have just one allocator to start with.

lib/Backends/BackendUtils.cpp
gcatron: I agree I created #2040 and it would also need to address #1599. This PR is just to unify the existing functionality.

lib/Backends/BackendUtils.cpp
gcatron: Comment added, this deallocation is more of a planned deallocation. It's a way of making sure we only allocate the memory we need at runtime.

qcolombet: Hmm, that sounds weird.
I would have expected that we just deallocate all the activations and inputs/outputs at the end of running a network. Why do we need to register them?
Again not something to fix here. I see you're just moving things around.

lib/Backends/BackendUtils.cpp
```diff
@@ -55,3 +55,97 @@ size_t glow::getValueOffset(
...
+      runtime::RuntimeSymbolInfo symbol;
+      symbol.offset = getValueOffset(tvSource, symbolTable) +
+                      (offsetLength * TV->getType()->getElementSize());
+      symbol.size = TV->getSizeInBytes();
+      symbol.type = *TV->getType();
+      symbolTable.emplace(std::string(TV->getName()), symbol);
+      continue;
+    }
+    // Remove buffer from allocator if not needed at runtime.
+    if (auto *D = llvm::dyn_cast<DeallocActivationInst>(&I)) {
```
qcolombet: The comment makes the understanding even more confusing :P.
If we don't need a buffer at runtime, why does it show up at all?

I believe that's not what you meant.

PR #2030, opened 2018-11-15
Support int32 index for Gather

yinghai:
*Description*:
The indices input of `Gather` can be either `int32_t` or `int64_t`,  we need to support both. This PR does this and adds support for Interpreter and CPU backend. 

*Testing*:
Unit test
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Graph/VerifierHelper.h
+++ b/include/glow/Graph/VerifierHelper.h
@@ -135,6 +135,41 @@ bool expectCompareTrue(
   return false;
 }
 
+/// Check whether $V_{0,n}{comp(\p a, \p b_i)}$ is true.
+/// If that check fails, \p msg is printed out using glow::report
+/// and \p parent (if not nullptr), \p a, and \p b are printed out
+/// using glow::reportContext.
+/// \returns \p comp(\p a, \p b_0) v ... v comp(\p a, \p b_i).
+template <typename InputTy>
+bool expectCompareTrue(
+    const char *msg, const InputTy &a, llvm::ArrayRef<InputTy> b,
+    const Node *parent,
+    const CompareWithName<InputTy> &comp = CompareOperatorEqual<InputTy>()) {
+  bool result = false;
+  for (const auto &bi : b) {
+    result |= comp(a, bi);
+  }
+  if (result) {
+    return true;
+  }
+  if (parent) {
+    reportContext(parent);
+  }
+  report(msg);
+  report("\nFor comparison `LHS ");
+  report(comp.getCompareName());
+  report(" RHS` with:");
+  report("\nLHS: ");
+  reportContext(a);
+  report("\nRHS: ");
+  for (const auto &bi : b) {
+    reportContext(bi);
+    report(", ");
+  }
+  report("\n");
+  return false;
+}
+
 /// Check that the type of the first operand \p A matches the type of the second
 /// operand \p B. \p parent is used to print the context of that check
 /// in case the it fails.
@@ -148,10 +183,16 @@ bool checkSameType(NodeValue A, NodeValue B, const Node *parent);
 bool checkSameShape(NodeValue A, NodeValue B, const Node *parent);
 
 /// Check that the element type of the operand \p A matches expected type \p
-/// expected Type. \p parent is used to print the context of that check
+/// expectedType. \p parent is used to print the context of that check
 /// in case the it fails.
 /// \see expectCompareTrue for more details.
 bool checkType(NodeValue A, ElemKind expectedType, const Node *parent);
+
+/// Check that the element type of the operand \p A matches any of the expected
+/// types \p expectedTypes. \p parent is used to print the context of that
+/// check in case the it fails. \see expectCompareTrue for more details.
+bool checkType(NodeValue A, llvm::ArrayRef<ElemKind> expectedTypes,
+               const Node *parent);
 
 /// Check if \p A and \p B have the same value for isQuantized. \p parent is
 /// used to print the context of that check in case the it fails.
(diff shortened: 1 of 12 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -818,7 +818,8 @@ bool RowwiseQuantizedFullyConnectedNode::verify() const {
 
 bool GatherNode::verify() const {
   bool isValid = checkType(getResult(), getData().getElementType(), this);
-  isValid &= checkType(getIndices(), ElemKind::Int64ITy, this);
+  isValid &= (checkType(getIndices(), ElemKind::Int64ITy, this) ||
```
yinghai: @qcolombet I would like to check that if the type is either of the two I supplier. But it will print out some warnings if one of the check fails, which is undesirable. Do you have any suggestion to allow check for an array of kinds?

yinghai: Resolved.

include/glow/Graph/VerifierHelper.h
```diff
@@ -135,6 +135,41 @@ bool expectCompareTrue(
...
+/// If that check fails, \p msg is printed out using glow::report
+/// and \p parent (if not nullptr), \p a, and \p b are printed out
+/// using glow::reportContext.
+/// \returns \p comp(\p a, \p b_0) v ... v comp(\p a, \p b_i).
+template <typename InputTy>
+bool expectCompareTrue(
+    const char *msg, const InputTy &a, llvm::ArrayRef<InputTy> b,
+    const Node *parent,
+    const CompareWithName<InputTy> &comp = CompareOperatorEqual<InputTy>()) {
+  bool seed = false;
```
rdzhabarov: can you find a better naming result or something?

yinghai: How about `result`?

include/glow/Graph/VerifierHelper.h
```diff
@@ -153,6 +188,12 @@ bool checkSameShape(NodeValue A, NodeValue B, const Node *parent);
 /// \see expectCompareTrue for more details.
 bool checkType(NodeValue A, ElemKind expectedType, const Node *parent);
 
+/// Check that the element type of the operand \p A matches any of the expected
+/// types \p expected Types. \p parent is used to print the context of that
```
rdzhabarov: can you fix: 
\p expectedTypes
"... check in case it fails ..."

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -370,7 +370,7 @@ void libjit_gather(T *dest, const T *data, const size_t *indices,
     size_t sampleStart = sample * sampleSize;
 
     // For each slice that we fetch:
-    for (size_t i = 0; i < numIndices; i++) {
+    for (IDX i = 0; i < numIndices; i++) {
```
rdzhabarov: IDX is for indices[i] not for the loop iteration.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -2153,7 +2153,16 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     auto *numSamplesVal = emitConstSizeT(builder, numSamples);
     auto *sampleSizeVal = emitConstSizeT(builder, sampleSize);
 
-    auto *F = getFunction("gather", dest->getElementType());
+    llvm::Function *F = nullptr;
+    if (indices->getElementType() == ElemKind::Int64ITy) {
+      F = getFunction("gather64", dest->getElementType());
```
rdzhabarov: can you add comments explaining why we need gatherTYPE + another type.

lib/Graph/Nodes.cpp
```diff
@@ -818,7 +818,9 @@ bool RowwiseQuantizedFullyConnectedNode::verify() const {
 
 bool GatherNode::verify() const {
   bool isValid = checkType(getResult(), getData().getElementType(), this);
-  isValid &= checkType(getIndices(), ElemKind::Int64ITy, this);
+  isValid &= checkType(
+      getIndices(),
+      llvm::makeArrayRef({ElemKind::Int64ITy, ElemKind::Int32ITy}), this);
```
rdzhabarov: do you need explicitly "llvm::makeArrayRef" ?
What about just {ElemKind::Int64ITy, ElemKind::Int32ITy}?

yinghai: I just followed the convention in this file. I guess it doesn't matter too much here?

tests/unittests/onnxImporterTest.cpp
```diff
@@ -612,17 +612,18 @@ TEST(onnx, expandDims) {
...
   Tensor data(ElemKind::FloatTy, {3, 2});
-  Tensor indices(ElemKind::FloatTy, {2, 4});
+  Tensor indices(ElemKind::Int32ITy, {2, 4});
 
   {
     ONNXModelLoader onnxLD(netFilename, {"data", "indices"},
                            {&data.getType(), &indices.getType()}, *F);
     output = onnxLD.getSingleOutput();
   }
+  EE.compile(CompilationMode::Infer, F);
```
rdzhabarov: no need to compile since no computations are done (operatorTest runs through the compilation anyway)

lib/Graph/Nodes.cpp
rdzhabarov: Just harder to read with more things thrown here

tests/unittests/onnxImporterTest.cpp
yinghai: Actually, I added this explicitly because previous one didn't really go through the verification. That's why the test could even pass with Indices of type `float`.

lib/Graph/Nodes.cpp
qcolombet: Good :).

PR #2034, opened 2018-11-15
[trivial] Rewrote some duplicated code.

beicy:
*Description*:
Rewrote a few line to avoid code duplication.

*Testing*:
Current test cases.

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -421,6 +421,18 @@ Placeholder *Module::addPlaceholder(Placeholder *ph) {
   return ph;
 }
 
+/// Check the kernel size for Conv/Pooling ops.
+static void checkKernelSize(ShapeNHWC idim, llvm::ArrayRef<unsigned_t> kernels,
+                            llvm::ArrayRef<unsigned_t> pads) {
+  PaddingTLBR pdim(pads);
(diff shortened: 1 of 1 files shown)


lib/Graph/Graph.cpp
```diff
@@ -421,22 +421,29 @@ Placeholder *Module::addPlaceholder(Placeholder *ph) {
...
+static void checkKernelSize(ShapeNHWC idim, llvm::ArrayRef<unsigned_t> kernels,
+                            llvm::ArrayRef<unsigned_t> pads) {
+  PaddingTLBR pdim(pads);
+  (void)pdim;
+  ShapeHW kdim(kernels);
+  (void)kdim;
+  assert((idim.w + pdim.left + pdim.right) >= kdim.width &&
+         (idim.h + pdim.top + pdim.bottom) >= kdim.height &&
+         "Kernel size is too large");
+}
```
qcolombet: These checks already happen in Convolution::verify.
I suggest to call the verify (wrapped into an assert) instead of duplicating the logic.
One way to easily fix that would be to call ::verify on every node that we add to the graph and make sure all the checks we have here are in the verifier.
I'll file an issue in that direction.

PR #2035, opened 2018-11-15
[examples] Add cmake for LeNet MNIST bundle

SplitInfinity:
**Description**
This commit adds the capability to build the LeNet MNIST standalone bundle
using `cmake` in addition to `make`. There are separate targets for
quantized and non-quantized version of the same bundle so that both can
be built in the same build tree.

**Testing**
Configured fresh build and ran bundle running cmake targets:

```
$ ninja RunLeNetMnistBundle
...
Allocated weights of size: 1724672
Expected weights of size: 1724672
Loaded weights of size: 1724672 from the file lenet_mnist.weights
Allocated mutable weight variables of size: 3200
...
Loaded images size in bytes is: 31360
Copying image data into mutable weight vars: 31360 bytes
Result: 0
```

**Fixes**
This is the second of the four bundles that will be migrated as part of #1979.


examples/bundles/lenet_mnist/CMakeLists.txt
```diff
@@ -0,0 +1,106 @@
...
+set(RUNTIME_OUTPUT_DIRECTORY ${LENET_MNIST_BUNDLE_DIR})
+
+# Run Commands
+# ============
+# Regular
+add_custom_command(
+  OUTPUT
+    RunBundleCommand
+  COMMAND
+    LeNetMnistBundle ${IMAGES}/*.png
```
SplitInfinity: I just noticed only processes the first of the items produced by expanding `${IMAGES}/*.png`. I made this mistake with the Resnet bundle migration too.

The reason I wrote the command like this is that I couldn't get a bash for-loop to work. cmake didn't expand `LeNetMnistBundle` when I replaced the `COMMAND` with `for ...`. I think it has something to do with the reference to the cmake target being in the middle of the command instead of the beginning?

SplitInfinity:
@nadavrot 

Oh, I thought the whole point was to migrate all of them. I will gladly stop if setting up bundles for the others is not worth it.

SplitInfinity:
Discussed with @bertmaher and the resnet50 example should be enough. Closing this PR.

nadavrot:
@SplitInfinity Do we want to delete the Makefiles for the non-resnet examples?

bertmaher:
We should at least keep the lenet_mnist one because it's tiny and we can use it in CI.  But vgg19 and zfnet, I'm not sure those really add any value beyond what resnet50 provides...

SplitInfinity:
I think we should set up cmake for all of them so that everything just works out of the box but it's not something I feel too strongly about.

We could also get rid of the bundles that we don't plan on supplying Make or cmake files for. I think that's what @bertmaher is hinting at.

SplitInfinity:
Reopening PR since the new plan is this:
1) Convert lenet_mnist to cmake
2) Delete vgg19 and zfnet bundles
3) Update docs

examples/bundles/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
 if (GLOW_WITH_BUNDLES)
+  add_subdirectory(lenet_mnist)
   add_subdirectory(resnet50)
+  #add_subdirectory(vgg19)
```
bertmaher: Delete commented code (we can remove those dirs in a separate PR).

PR #2036, opened 2018-11-15
Add magic numbers to classes sent across onnixifi interface

### LINKED ISSUE #2026: [ONNXIFI] onnxifi interface implementation can't identify invalid IDs
The onnxifi interface implementation in its current form cannot distinguish between valid and invalid `onnxBackendID`s, `onnxBackend`s, `onnxEvent`s, `onnxGraph`s, etc. To do this it will need to actually track and update the set of IDs it has created and released.

jackm321:
*Description*:
Add magic numbers to classes sent across onnixifi interface in order to verify that they probably originated from glow.
*Testing*:
`ninja all && ninja test`
*Documentation*:
doxygen 
fixes #2026

### DIFF
--- a/include/glow/Base/Traits.h
+++ b/include/glow/Base/Traits.h
@@ -102,6 +102,27 @@ public:
 
 using KindSet = llvm::SmallSet<Kinded::Kind, 4>;
 
+/// Subclasses of this class begin with a uint32_t magic number that can be used
+/// to check if arbitrary chunks of memory are likely to be instances of that
+/// subclass. Note that this enforces that magic_ field has no offset from the
+/// beginning of the subclass in memory. This will only be true if subclasses
+/// inherit from this class first and do not contain a vtable pointer.
+template <typename CH, uint32_t MAGIC> class HasMagic {
+private:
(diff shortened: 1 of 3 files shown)


rdzhabarov:
Have you considered having static set for backendId, backend, etc to keep what we allocated and deallocated?

jackm321:
I did consider keeping a set of things that have been allocated but after reading the [onnx wiki](https://github.com/onnx/onnx/wiki/ONNX-Interface-for-Framework-Integration:-API-Proposal) it seems that this would be more than what the interface is expecting so I think it would be better to go with this more-simple solution. Thoughts?

rdzhabarov:
There are certain benefits as simplicity but unfortunately having magic embedding is fragile and does not allow to detect double deallocations.

bertmaher:
Do we actually expect caffe to hand us invalid graphs during normal operation, or is this a sort of debugging or defensive check?

jackm321:
@bertmaher my impression is that this is the second thing, a kind of defensive check that we are expected to do to prevent bad things from happening on our end if an onnxifi client hands us some thing bad accidentally.

PR #2052, opened 2018-11-19
[image] Fix build when libpng not found

bertmaher:
*Description*: I suspect no one ever builds glow without libpng because it's been broken for a while :-).
*Testing*: Actually pretty hard; I removed find_package(PNG) and got things to compile, but that's still not a guarantee that everything works because libpng is still in my system path, even if I haven't told CMake to go find it.  Anyways, at least it doesn't choke as quickly.
*Documentation*: n/a

### DIFF
--- a/lib/Base/Image.cpp
+++ b/lib/Base/Image.cpp
@@ -22,7 +22,6 @@
 
 using namespace glow;
 
-#if WITH_PNG
 #include <png.h>
 
 namespace glow {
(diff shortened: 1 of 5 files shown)


lib/Base/Image.cpp
```diff
@@ -18,11 +18,10 @@
 #include "glow/Base/Tensor.h"
 #include "glow/Support/Support.h"
 
-using namespace glow;
-
-#if WITH_PNG
 #include <png.h>
```
jfix71: Hmm, doesn't this still need to be wrapped by `WITH_PNG`? If it's not installed the header wouldn't be found.

bertmaher: I've changed the CMakeLists.txt for this library to only include Image.cpp if PNG_FOUND, so this entire file won't be included.  I think that makes sense, since there's probably nothing in here that's useful if you don't have libpng.

jfix71: Ah, got it. It seems like `WITH_PNG` isn't used at all after this PR then? Then what happens when you try to link Base to the `image-classifier`, wouldn't that throw an error since `readPngImageAndPreprocess()` won't be found? It seems like the `image-classifier` also shouldn't be built if `!PNG_FOUND`

bertmaher: Ah yes, I should remove `WITH_PNG`.  Good catch.  It looks like nothing under `tools/loader` is built unless`PNG_FOUND`... which is probably too conservative since we actually use that for `text-translator` and `model-runner`.  But that's a separate issue :-)

lib/Base/Image.cpp
```diff
@@ -22,7 +22,6 @@
 
 using namespace glow;
 
-#if WITH_PNG
```
qcolombet: Maybe add a #error in the header if WITH_PNG is false.
Actually, if you have other idea on how to signal that libpng is required for this file that works too.
The thing I am after is, say I use the API for readPng and such, I would like something complaining loudly other than a linker error. In other words, if I missed that I don't have libpng and uses it, I don't want to debug the build process to find that information :).

PR #2057, opened 2018-11-19
[tensor] Speed up convolution with unsafe_at

### LINKED ISSUE #2022: Debug mode interpreter is really slow
I don't know what we could do about this, but running resnet50 in Debug mode with the interpreter is just way too slow.  We could just say "don't do that", but since Debug mode is extremely useful for development I keep wishing it were faster.  I assume the problem is all the range checks on `.at()` during the convolutions but I don't know how to sidestep it while still providing some safety.

bertmaher:
*Description*:
We'd like a build that's decently fast (e.g., to run ResNet-50) but still enables assertions.  This diff does two things:
1. Enables asserts in RelWithDebInfo by stripping out -DNDEBUG.
2. Introduces `Handle::unsafe_at` and uses it in the interpreter's convolution implementation.  (Without this, even compiler optimizations aren't enough to make it fast).

*Testing*:
```
time ./bin/image-classifier tests/images/imagenet/*.png -use-imagenet-normalization -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data
```
goes from about 3m40s to 20s.  Release is 3s, which we could probably get by adding `unsafe_getRawDataPointer`, but 20 seconds is at least usable, and I don't really enjoy adding these `unsafe_` variants.

*Documentation*: n/a

Fixes #2022


jfix71:
RE: `unsafe_*`, this means we would never check the asserts inside convolutions even in Debug mode. That seems a bit scary -- I think Debug should be as safe as possible, and then RelWithDebInfo could strip out these asserts that hurt perf. Debug is already going to be slow due to lack of optimizations anyway, so I think keeping asserts there doesn't hurt much, right?

What do you think about instead wrapping the assert inside `at()` and `getElementPtr()` with our own macro check, so that we can keep it on for Debug mode but turn it off for RelWithDebInfo? This means it would be disabled for all uses during RelWithDebInfo, though that could also be solved.

nadavrot:
@bertmaher I wonder how much performance boost do we get from each one of thee commits. I tried to comment out the asserts in at() but I did not see any perf wins.

bertmaher:
> What do you think about instead wrapping the assert inside at() and getElementPtr() with our own macro check

@jfix71 Sure, I'm on board with that.  Debug is pretty much doomed for running real nets anyways, as you say :-).  I'll update this PR.

@nadavrot D'oh, I dropped the buffer I had the perf results in.  I'll re-run real quick, shouldn't take too long.

bertmaher:
Perf results.  It looks like each of the three assertions was contributing a significant amount of time:
```

time (s) | improvement (s) | function | assertion
-- | -- | -- | --
20 | 206 | all opts |  
134 | 92 | getElementPtr | assert(indices.size() <= numDims_ && "Invalid number of   indices");
184 | 42 | at | assert(index < size() && "Out of bounds");
163 | 63 | at | assert(tensor_->isInBounds(indices));
226 |   | baseline |  
```

jfix71:
Nice! And this was all with RelWithDebInfo?

bertmaher:
Oh, yes, RelWithDebInfo :)

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -198,15 +198,16 @@ void InterpreterFunction::fwdConvolutionInst_I8Impl(
             }
 
             // Scale the bias to match the scale of the matrix multiplication.
-            int32_t B = std::round(float(biasW.at({d}) - biasOffset) *
+            int32_t B = std::round(float(biasW.unsafe_at({d}) - biasOffset) *
```
qcolombet: Just a personal opinion: it feels strange to have unsafe_at here and not in the other places.

If we want to selectively enable faster access with a good granularity, I guess we could have something like:
::at(..., bool skipAssert = false) {
assert(skipAssert || (<regularAssertCode>))
}

Then we can instrument each access with macros, like USE_FAST_CONV.

Just an idea though. I expect this to be relatively ugly while not completely grasping the intent here x).

nadavrot:
@bertmaher I am not sure that I understand this PR. Is it our goal to run Resnet50 in debug mode?
Do I need to stop using Debug and start using RelWithDebInfo? Why only Conv and not ConvGrad? Our unit tests spend much more time in ConvGrad. Is the goal to make our unit tests run faster or run Renset50 in debug mode?

bertmaher:
My goal here is to have some compilation mode that is able to run a real network (resnet-50 just happens to be the one I use most often) while having most asserts enabled.  While debugging other backends I find that I frequently want to check the backend results against the interpreter, so a slow interpreter really gets in the way.  But I still want assertions because I'm deep in debugging and want them turned on to catch trivial mistakes.

bertmaher:
It's possible that I'm fighting against (a) working on a target that doesn't really support our unit tests, and (b) a lack of smaller tests that exercise more realistic behavior.  But I don't think I'm crazy for wanting less than a 70x slowdown from my Debug build :-).

As for other ops, I'm open to applying this technique to them too (MatMul comes to my mind).  I just haven't prioritized them because they're not as consistently painful as Conv.

lib/Backends/Interpreter/InterpreterNodes.cpp
bertmaher: Tbh, I kind of wanted this call to be jarring from an API perspective.  No one should be using this except for a well understood place like convolution, where we already have type checks in place to make sure e.g. your filter/input/output dimensions match.

nadavrot:
I thought that you want to reduce the execution time of our unit tests that run in Debug mode. 
Resnet50 has 7Gflops of calculations, so we can't perform bounds checks on all array accesses. Right? But we do want to have bounds checks when we run the unit tests because we do want to check that our 10x10 convolutions are correct. I usually work using two build trees (Debug and Release). 

One possible solution would be to add a cmake flag such as DISABLE_BOUNDS_CHECKS that will turn all at() to at_unsafe() by eliminating the assertions inside. This would allow you to run all of the assertions except for the bounds checks. Would this make sense?

bertmaher:
I think I'm the one most impacted by the low performance of debug builds, so I'm going to withdraw this PR but keep the branch around so anyone suffering like me can patch it in :-)

PR #2058, opened 2018-11-19
[BackendUtils]Convert RuntimeBundle to a class

### LINKED ISSUE #1967: Convert RuntimeBundle to object.
RuntimeBundle is currently a struct which contains the information passed forward from compile time to runtime. It contains max allocations, constants, and symbol information. This is a general backend tool so it should be moved to BackendUtils. It should also be made into a class so accessors like getSymbolOffset can be called. The backends will need to be updated as well.

gcatron:
*Description*: Converted RuntimeBundle to a class and added helper methods for getting symbolInfo. Moved RuntimeBundle to BackendUtils. Updated Backends to use new helper methods.
*Testing*: ninja test and ruh.sh both pass
*Documentation*: N/A
Fixes #1967

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -21,26 +21,64 @@
 #include "glow/IR/IR.h"
 
 namespace glow {
-/// At compile time condense constants to a single block of memory.
-/// This allows the graph to go away after compile time.
-/// Allocates a block of memory of size \p constantMaxSize then walks the given
-/// function \p F and and copies weights to their address as specified by
-/// offsets contained in \p symbolTable.
-uint8_t *collectConstants(
-    const IRFunction *F, uint64_t constantMaxSize,
-    const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
-        &symbolTable);
-/// Helper function to retrieve offset for Value: \p v from \p symbolTable.
-size_t
-getValueOffset(Value *v,
-               const std::unordered_map<std::string, runtime::RuntimeSymbolInfo>
-                   &symbolTable);
+namespace runtime {
+/// Contains information for initialization and handling of symbol at runtime.
+struct RuntimeSymbolInfo {
+  /// The size in bytes.
+  size_t size;
+  /// Offset in bytes from the base address.
+  size_t offset;
+  /// Type of symbol.
+  Type type;
+};
+
+/// Contains the information needed to be passed forward from compile time to
+/// runtime. In order to allocate and initialize memory.
+class RuntimeBundle {
+  /// Map from symbol name to a RuntimeSymbolInfo.
+  std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable_;
+  /// Pointer to memory containing the weights for execution.
+  uint8_t *constants_;
+  /// Amount of memory needed for weights.
+  const size_t constantWeightVarsMemSize_;
+  /// Amount of memory needed for mutable vars.
+  const size_t mutableWeightVarsMemSize_;
+  /// Amount of memory needed for activations.
+  const size_t activationsMemSize_;
+
+public:
+  /// Get Constant Weights memory size.
+  size_t getConstantWeightSize() const { return constantWeightVarsMemSize_; }
+  /// Get Mutable Weights memory size.
+  size_t getMutableWeightSize() const { return mutableWeightVarsMemSize_; }
+  /// Get Activations Weights memory size.
+  size_t getActivationsSize() const { return activationsMemSize_; }
+  /// Get pointer to memory block of constants.
+  uint8_t *getConstants() const { return constants_; }
+  /// Helper function, gets offset of \p v.
+  size_t getValueOffset(const Named *v) const;
+  /// Helper function, gets symbol info for \p v.
+  RuntimeSymbolInfo getSymbolInfo(const Named *v) const;
+  /// At compile time condense constants to a single block of memory.
+  /// This allows the graph to go away after compile time.
+  /// Allocates a block of memory of size \p constantMaxSize then walks the
+  /// given function \p F and and copies weights to their address as specified
+  /// by offsets contained in symbolTable_.
+  void collectConstants(const IRFunction *F);
+  RuntimeBundle(std::unordered_map<std::string, RuntimeSymbolInfo> &symbolTable,
+                size_t constWeight, size_t mutableWeight, size_t activations)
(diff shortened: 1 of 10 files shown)


lib/Backends/CPU/CPUFunction.h
```diff
@@ -18,6 +18,7 @@
 
 #include "GlowJIT.h"
 
+#include "glow/Backends/BackendUtils.h"
```
jfix71: Why is it necessary to include here? Can you just include in `CPUFunction.cpp` instead?

lib/Backends/Interpreter/InterpreterFunction.h
```diff
@@ -16,6 +16,7 @@
 #ifndef GLOW_BACKENDS_INTERPRETER_INTERPRETERFUNCTION_H
 #define GLOW_BACKENDS_INTERPRETER_INTERPRETERFUNCTION_H
 
+#include "glow/Backends/BackendUtils.h"
```
jfix71: Why is this necessary, does the interpreter use this at all? If so can it be included in the cpp where it's used?

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_OPENCL_OPENCL_H
 
 #include "glow/Backends/Backend.h"
+#include "glow/Backends/BackendUtils.h"
```
jfix71: Same as CPU -- can this be moved to `OpenCL.cpp` instead of including in the header?

include/glow/Backends/BackendUtils.h
```diff
@@ -21,6 +21,43 @@
 #include "glow/IR/IR.h"
 
 namespace glow {
+namespace runtime {
+/// RuntimeSymbolInfo
```
bertmaher: Nit: I'm not a fan of comments that just repeat the name of the thing being commented :-)

include/glow/Backends/BackendUtils.h
```diff
@@ -21,6 +21,43 @@
...
+/// Contains information for initialization and handling of symbol at runtime.
+struct RuntimeSymbolInfo {
+  /// The size in bytes.
+  size_t size;
+  /// Offset in bytes from the base address.
+  size_t offset;
+  /// Type of symbol.
+  Type type;
+};
+/// Runtime Bundle
```
bertmaher: Same nit (name of thing in comment).

lib/Backends/CPU/CPUFunction.h
jfix71: Ah whoops I totally missed what was going on here, ignore me 😀

include/glow/Backends/BackendUtils.h
bertmaher: I'd also prefer a blank line between top-level class/struct declarations.  (I dunno if this is in the official coding standards but we do it most other places.) (In general I hate blank lines but this is my one exception.) (I also love parenthetical comments.)

include/glow/Backends/BackendUtils.h
```diff
@@ -41,6 +78,7 @@ runtime::RuntimeBundle
 generateRuntimeBundle(const IRFunction &F, MemoryAllocator &constantAllocator,
                       MemoryAllocator &placeholderAllocator,
                       MemoryAllocator &activationsAllocator);
+
```
bertmaher: By contrast I could do without this additional blank line ;-).

include/glow/Backends/BackendUtils.h
```diff
@@ -21,6 +21,43 @@
...
+  /// Offset in bytes from the base address.
+  size_t offset;
+  /// Type of symbol.
+  Type type;
+};
+/// Runtime Bundle
+/// Contains the information needed to be passed forward from compile time to
+/// runtime. In order to allocate and initialize memory.
+class RuntimeBundle {
+public:
```
opti-mix: Typically, all members of structs are public by default and all members of classes are private by default. And if some internals of the class need to be exposed, it is done via getters and setter. Also the private members would have the "_" suffix in their names.

So, why is `RuntimeBundle` a class, if all its members are public?

include/glow/Backends/BackendUtils.h
```diff
@@ -21,26 +21,65 @@
...
+  /// Amount of memory needed for weights.
+  const size_t constantWeightVarsMemSize_;
+  /// Amount of memory needed for mutable vars.
+  const size_t mutableWeightVarsMemSize_;
+  /// Amount of memory needed for activations.
+  const size_t activationsMemSize_;
+
+public:
+  /// Get Constant Weights memory size.
+  size_t getConstantWeightSize() { return constantWeightVarsMemSize_; }
```
opti-mix: it should be a `const` method.

include/glow/Backends/BackendUtils.h
```diff
@@ -21,26 +21,65 @@
...
+  /// Get Mutable Weights memory size.
+  size_t getMutableWeightSize() { return mutableWeightVarsMemSize_; }
```
opti-mix: Same here.

include/glow/Backends/BackendUtils.h
```diff
@@ -21,26 +21,65 @@
...
+  /// Type of symbol.
+  Type type;
+};
+
+/// Contains the information needed to be passed forward from compile time to
+/// runtime. In order to allocate and initialize memory.
+class RuntimeBundle {
+  /// Map from symbol name to a RuntimeSymbolInfo.
+  std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable_;
+
```
opti-mix: Why the newline here?

include/glow/Backends/BackendUtils.h
```diff
@@ -21,26 +21,65 @@
...
+  size_t getValueOffset(const Named *v);
+  /// Helper function, gets symbol info for \p v.
+  RuntimeSymbolInfo getSymbolInfo(const Named *v);
+  /// At compile time condense constants to a single block of memory.
+  /// This allows the graph to go away after compile time.
+  /// Allocates a block of memory of size \p constantMaxSize then walks the
+  /// given function \p F and and copies weights to their address as specified
+  /// by offsets contained in symbolTable_.
+  void collectConstants(const IRFunction *F);
+  RuntimeBundle(std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable,
```
opti-mix: Should the `symbolTable` argument be passed by reference?

Issue #2059, opened 2018-11-20
[Bug] Overflow in FunctionSpecializer


yinghai:
I encountered this piece of code:
https://github.com/pytorch/glow/blob/fc98cd34b8e7374e4494ef0db82623725b86b826/lib/Backends/CPU/FunctionSpecializer.cpp#L362-L364

This is not safe as the number of operands of the function (`e`) can exceed `63` and this will induce overflow/undefined behavior at
https://github.com/pytorch/glow/blob/fc98cd34b8e7374e4494ef0db82623725b86b826/lib/Backends/CPU/FunctionSpecializer.cpp#L83

I actually did read the code enough to understand what functions in this file do but this doesn't look right.

yinghai:
I wonder if we can just truncate and use the first 64 arguments by replace
https://github.com/pytorch/glow/blob/fc98cd34b8e7374e4494ef0db82623725b86b826/lib/Backends/CPU/FunctionSpecializer.cpp#L81
with 
```
return (argIdx > 63) ? false : argsToBeSpecialized & (((uint64_t)1) << argIdx);
```

yinghai:
@bertmaher @opti-mix

opti-mix:
@yinghai Good catch! I agree that this will induce overflow/undefined if the number of operands of the function would exceed 63. 

But I don't think it is very likely that we ever get a function with >63 operands. Thus, as a simplest solution I'd suggest adding an `assert` checking that the number of operands is <=63. 

A more future proof solution would be to use e.g. a dynamic `bitset` (or any other data structure which can have a dynamically defined size) instead of a fixed-width 64bit unsigned value to hold the mask.

opti-mix:
> return (argIdx > 63) ? false : argsToBeSpecialized & (((uint64_t)1) << argIdx);

Yes, seems like this could work. And as I described above, it is very unlikely that we ever hit the case `argIdx > 63`.

yinghai:
> Yes, seems like this could work. And as I described above, it is very unlikely that we ever hit the case argIdx > 63.

It's very likely when we have a lot of parallel embedding looks as inputs. :)

opti-mix:
> It's very likely when we have a lot of parallel embedding looks as inputs. :)

You mean we would pass those inputs not as a tensors of size > 63, but as separate arguments of a function? Hmm. I haven't thought about this use-case ;-)

[...]

PR #2060, opened 2018-11-20
Fix overflow/undefiend behavior when #args exceeds 63

yinghai:
*Description*:
Check issues raised in https://github.com/pytorch/glow/issues/2059. 

If llvm provides some sort of bitset, I can see to a more generic solution. 

*Testing*:
Unit tests. 

*Documentation*:
Fixes https://github.com/pytorch/glow/issues/2059. 


yinghai:
@opti-mix Maybe I should just use `llvm::BitVector`?
http://llvm.org/doxygen/classllvm_1_1BitVector.html

opti-mix:
@yinghai Thanks for catching the issue and submitting this PR!

I'm not sure that your fix is quite correct.

It seems to me, that we should do something like this:
```c++
for (auto &arg : call->arg_operands()) {
      auto curArgIdx = argIdx++;

      // Replace the next line with : if (curArgIdx > 63 || !shouldSpecializeParameter(arg)) {
      // or even move this check inside `shouldSpecializeParameter`, which would get the additional idx argument.
      if (!shouldSpecializeParameter(arg)) {
        argsForSpecialized.push_back(arg);
        continue;
      }
      ...
}
```

And the functions for setting and checking the bit should assert on `idx > 63`.

opti-mix:
> Maybe I should just use llvm::BitVector?

Yes, I think it would work and we wouldn't have any limitation on the number of arguments.

Still, it seems like it doesn't make sense to specialize the function with >63 arguments in your case. The problem is that the logic in `shouldSpecializeParameter` is too general and too simple at the moment. It doesn't know anything about specific operators and thus it cannot do anything selectively.

But have a look at `LLVMIRGen::markArgAsUnspecialized`, which can be used to prohibit specialization of some arguments. Have a look how we use it for some specific operations. May be we should do the same for embedding tables lookups?

yinghai:
I start to think that this will create an issue on comparing the `SpecializationKey`, if we truncate at first 64 inputs, we might return `true` even if the 65th argments are different and this is incorrect.

yinghai:
@opti-mix I guess the fix works for now.

opti-mix:
@yinghai Let's have a look at your concrete example offline, before we decide how to proceed.

lib/Backends/CPU/FunctionSpecializer.cpp
```diff
@@ -267,7 +269,7 @@ class FunctionSpecializer {
     for (auto &arg : call->arg_operands()) {
       auto curArgIdx = argIdx++;
 
-      if (!shouldSpecializeParameter(arg)) {
+      if (curArgIdx > 63 || !shouldSpecializeParameter(arg)) {
```
opti-mix: Looking almost good. I'd just move this new check into `shouldSpecializeParameter`, because it would be cleaner.

Right now `shouldSpecializeParameter` takes only the `arg` as its arguments. But it seems like it should also take parameters called `callee` and `argIdx` with values `callee` and `curArgIdx` respectively. This way more complex conditions can be expressed inside `shouldSpecializeParameter` now and in the future.

lib/Backends/CPU/FunctionSpecializer.cpp
```diff
@@ -102,7 +104,15 @@ class FunctionSpecializer {
   /// \p arg. In the future, we may need to improve this logic by taking into
   /// account the semantics of the argument or even the specifics of the
   /// function call being specialized.
-  bool shouldSpecializeParameter(llvm::Value *arg) {
+  bool shouldSpecializeParameter(llvm::Value *arg, unsigned argIdx,
+                                 llvm::Function * /* unused */) {
```
opti-mix: Let's give this parameter a name `callee`. But you can leave the comment that it is unused for now.

opti-mix:
@yinghai Looking really good now!

opti-mix:
@yinghai Don't forget to squash before merging

yinghai:
> @yinghai Don't forget to squash before merging

Squash and merge can be done directly on github and is the default now. :)

PR #2062, opened 2018-11-20
README Updates for OpenCL on Ubuntu

hegemanjwh2:
*Description*: This commit adds instructions for using OpenCL with Glow on Ubuntu.

*Testing*: The instructions added were successfully tested on my Ubuntu 16.04.1 laptop with an i7-7600U (with Intel HD Graphics 620 onboard graphics).


rdzhabarov:
alternative way is to use device agnostic interface for the device: https://github.com/pocl/pocl

That's how we run it in CI. I'll update the doc with this.

rdzhabarov:
updated https://github.com/pytorch/glow/pull/2068. I think with that it's pretty straightforward to run OpenCL tests.

PR #2065, opened 2018-11-20
Limit the emitDataParallelKernel to generate function with no more th…

yinghai:
…an 64 arguments

*Description*:
As a followup of https://github.com/pytorch/glow/issues/2059. This will fix the root cause of why we even have functions with 64 arguments. 
*Testing*:
I have to test it through an internal case. @opti-mix, is there any easy way to build a test case for this case? 

*Documentation*:
Fixes https://github.com/pytorch/glow/issues/2059. 


lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -676,6 +655,50 @@ void LLVMIRGen::emitDataParallelKernel(
...
+    if (argTypes.size() + I->getOperands().size() > 64) {
+      emitDataParallelKernelImpl(builder, batchedBundle, argTypes,
+                                 bufferToArgNum, buffers);
+      batchedBundle.clear();
+      argTypes.clear();
+      bufferToArgNum.clear();
+      buffers.clear();
+    }
+    batchedBundle.push_back(I);
+    for (const auto &Op : I->getOperands()) {
```
opti-mix: In theory, to be very exact, we should look how many unique operands a given instruction has and then check if `argTypes.size() + numUniqueOperands > 64`. Your check above `argTypes.size() + I->getOperands().size() > 64` is correct, but more pessimistic.

lib/Backends/CPU/LLVMIRGen.h
```diff
@@ -140,8 +149,8 @@ class LLVMIRGen {
   /// the type specified by \p kind.
   llvm::Value *emitConst(llvm::IRBuilder<> &builder, float val,
                          glow::ElemKind kind);
-  /// Generates LLVM IR that materializes the constant array \p vals. Note that
-  /// it will cast non-size_t types T into size_t.
+  /// Generates LLVM IR that materializes the constant array \p vals. Note
+  /// that it will cast non-size_t types T into size_t.
```
opti-mix: All the formatting changes in comments here and below seem to be unrelated to your changes. Please remove them.

lib/Backends/CPU/LLVMIRGen.cpp
yinghai: Yes, it is perssimistic, I forgot to add this comments. But I think the trade-off is that the algorithm becomes simpler and it the downside isn't much. So algorithm wise, I think it is good.

lib/Backends/CPU/LLVMIRGen.h
yinghai: Clang format did this to me... I'll split to a different commit maybe.

opti-mix:
> I have to test it through an internal case. Is there any easy way to build a test case for this case?

It seems like this kind of LLVM code is produced from an IR function with a lot of `copy` instructions performing on different buffers of the same shape. So, just having a graph with a lot of copies should do the trick, I guess.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -676,6 +655,50 @@ void LLVMIRGen::emitDataParallelKernel(
...
+  // Map each buffer used by the kernel to the argument number of the kernel
+  // function. This ensures that same buffer is always mapped to the same
+  // argument.
+  llvm::DenseMap<Value *, int> bufferToArgNum;
+  // Buffers to be passed to the kernel function as arguments.
+  llvm::SmallVector<llvm::Value *, 32> buffers;
+  // Collect unique buffers used by the instructions of the kernel.
+  llvm::SmallVector<const Instruction *, 32> batchedBundle;
+  for (const auto I : bundle) {
+    if (argTypes.size() + I->getOperands().size() > 64) {
```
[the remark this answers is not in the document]
yinghai: I know. In this way, the algorithm here is pessimistic. But It makes things easier. Because if we count unique buffers, for instruction that will make the arg size over 64, we need to do a recount of unique buffers, which makes the algorithm less clean.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -62,6 +62,8 @@ llvm::cl::opt<llvm::Reloc::Model> relocModel(
         clEnumValN(llvm::Reloc::PIC_, "pic", "Position independent code")),
     llvm::cl::init(llvm::Reloc::Static), llvm::cl::cat(getCPUBackendCat()));
 
+constexpr size_t kArgLimit = 64;
```
opti-mix: Please add a comment for this variable and make it static.

opti-mix:
@yinghai Have you checked on your internal use-cases that the issue is solved now?

yinghai:
> Have you checked on your internal use-cases that the issue is solved now?
Yes. I actually coded this and tested internally and exported it as a git patch. :)

yinghai:
@opti-mix I have to add back the clang-format as it's failing the CI now. :(

PR #2064, opened 2018-11-20
[ci] Build with -Werror in CI

bertmaher:
*Description*: It'd be nice to keep our CI build clean of warnings, so let's enabled -Werror there.

*Testing*: CI

*Documentation*:

PR #2066, opened 2018-11-20
Prevent errors from use of deprecated OpenCL 1.2 function

hegemanjwh2:
*Description*: This commit allows for building with -Werror without incurring failures due to the usage of the deprecated clCreateCommandQueue() function in the OpenCL backend.  This commit should allow #2064  to succeed.

*Testing*: Tested build with -Werror and -DGLOW_WITH_OPENCL=ON on Ubuntu 16.04.1 laptop (with OCL runtime installed).


hegemanjwh2:
For now, we should keep clCreateCommandQueue() because I believe that the replacement function is only supported in OpenCL 2.x...

PR #2067, opened 2018-11-21
Fix readPngImage for non-square images

tlepley-cadence:
*Description*: Fix the tensor shape of loaded PNG images (was not working for non square images)
*Testing*: N/A
*Documentation*: N/A


tlepley-cadence:
@bertmaher I just added a test

PR #2072, opened 2018-11-23
[WIP] Fix all compilation warnings with GCC 8.2.1

tlepley-cadence:
*Description*: Fix all compilation warnings with GCC 8.2.1:
  - Signed/unsigned comparisons
  - Use of AVX vector without enabling it in the ABI

*Testing*: N/A

*Documentation*: N/A

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1449,7 +1449,7 @@ static void find_min_max_f(float *tensor, size_t size, float &min, float &max) {
 }
 
 static int check_all_zeros(float *arrayToCheck, size_t size) {
-  for (int i = 0; i < size; ++i) {
+  for (size_t i = 0; i < size; ++i) {
     if (arrayToCheck[i] != 0) {
       return 0;
     }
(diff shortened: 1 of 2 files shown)


lib/Backends/CPU/CMakeLists.txt
```diff
@@ -53,6 +53,9 @@ add_library(CPURuntimeNative
               libjit/libjit.cpp
               libjit/libjit_conv.cpp
               libjit/libjit_matmul.cpp)
+target_compile_options(CPURuntimeNative
```
nadavrot: Should this be AVX2?

tlepley-cadence:
For the moment, this PR creates many segfaults. The reason may be that the callee is compiled with the `-mavx` option and not the caller, but not the caller ? (not exactly the same ABI). I need to investigate further.

tlepley-cadence:
@bertmaher Is there a way for me to check the potential performance loss for my PRs in the CI ?

nadavrot:
@tlepley-cadence @bertmaher Why are we converting int to size_t and not ssize_t which is the signed size_t?

bertmaher:
Oh.  Yeah `size_t` isn't actually valid here, we do things like `m - mr + 1` that could become negative.

bertmaher:
Although we could cast those particular expressions.

rdzhabarov:
This seems like stale for a week. @tlepley do you plan to work on this?

tlepley-cadence:
@rdzhabarov yes, I'm going to continue workign on this. I was just busy with higher priority PRs lately.

compnerd:
@nadavrot ... `ssize_t` is non-portable.  It is not part of the C or C++ standard.

tlepley-cadence:
OK, I'm going to close this PR for now. I'll reactivate it when I'll be ready.

PR #2081, opened 2018-11-26
[RFC][WIP][tests] Pythonize tests/images/run.sh

bertmaher:
*Description*: I've been feeling a little cool towards run.sh as it's gotten bigger; I mostly end up copypasting bits of it to run, and thought it might be nice to make it a little more friendly.  In the process, I wanted to move it to python since big bash scripts are kind of hard to maintain.

There are still a fair number of things I'd like to improve about this version, so I mostly just want feedback on whether anybody else would like this (and therefore I should polish it up), or whether I should not invest too much more time in this :).

*Testing*: python3 ./tests/python/ImageClassifierTest.py

*Documentation*: TBA


tests/python/ImageClassifierTest.py
```diff
@@ -0,0 +1,230 @@
...
+    def execute(self, testcase):
+        if self.batching:
+            out = self.check_output(self.images())
+        else:
+            out = "".join(self.check_output([image]) for image in self.images())
+        predictions = [
+            line.split()[3] for line in out.split("\n") if line.startswith(" File:")
+        ]
+        for prediction, image in zip(predictions, images[self.dataset]):
+            testcase.assertIn(int(prediction), image.labels)
```
rdzhabarov: this is cool!

bertmaher: Yeah, I was a bit torn whether to make the predicted results specific to each model.  It kind of bugs me that our networks can't decide if cat_285 is a tabby cat or Egyptian cat :-p

bertmaher:
A few things I'd like to do (maybe in follow-up PRs?):
[ ] Make each model a separate test
[ ] Print out a nice command line for easy reproducibility upon failure
[ ] Bonus points: automatically download required models.  Maybe cache them somewhere.
[ ] Figure out a way to run tests in parallel

rdzhabarov:
:) i have little python experience, but it looks like once we have each model tested in a separate test the last item (parallel) is easy with https://pypi.org/project/pytest-parallel/

jfix71:
This looks great! Would also be great if it could also automatically test quantizing the model -- dump the profile, and then load it and run it quantized.

bertmaher:
Superseded by #2097.

PR #2084, opened 2018-11-27
 [quantization] fix Tensor zero() and isZero() to respect quantized offset

### LINKED ISSUE #2082: [Quantization/Tensor] 0 should be represented as an offset for quantized numbers
[zero](https://github.com/pytorch/glow/blob/master/include/glow/Base/Tensor.h#L95) and [isZero](https://github.com/pytorch/glow/blob/master/include/glow/Base/Tensor.h#L628) are not properly handled for quantized types.

Taking quantized int8 for example with {Scale, Offset} parameters.
Given that dequantization is calculated as Scale * (x_quantized - Offset), actual zero value should be represented as Offset and not just 0.

nickgg:
*Description*: Per #2082, Tensors zero() to 0 always which is not correct for quantized types which may have a zero offset. This changes zero() to set each item to the offset, and isZero() to check against the offset for quantized types.
*Testing*: All unittests, but especially tensor tests reset and the new test zeroQuantizedTensor.
*Documentation*:
fixes #2082.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -93,8 +93,26 @@ public:
 
   /// Set the content of the tensor to zero.
   void zero() {
-    std::fill(&getData()[0], &getData()[0] + size() * type_.getElementSize(),
-              0);
+    // Quantized tensors should go to their offset.
+    switch (type_.getElementType()) {
+    case ElemKind::Int8QTy: {
+      auto *data = reinterpret_cast<int8_t *>(getData());
+      std::fill(&data[0], &data[0] + size(), (int8_t)type_.getOffset());
+    } break;
+    case ElemKind::Int16QTy: {
+      auto *data = reinterpret_cast<int16_t *>(getData());
+      std::fill(&data[0], &data[0] + size(), (int16_t)type_.getOffset());
+    } break;
+    case ElemKind::Int32QTy: {
+      auto *data = reinterpret_cast<int32_t *>(getData());
+      std::fill(&data[0], &data[0] + size(), (int32_t)type_.getOffset());
+    } break;
+    default:
+      // Non-quantized tensors are set to 0.
+      std::fill(&getData()[0], &getData()[0] + size() * type_.getElementSize(),
+                0);
+      break;
+    }
   }
 
   /// \returns the shape of the tensor.
@@ -626,8 +644,9 @@ public:
 
   /// \returns true if tensor contains only elements equal to zero.
   bool isZero() const {
+    int32_t trueZero = getType().isQuantizedType() ? getType().getOffset() : 0;
     for (size_t i = 0, e = size(); i < e; i++) {
-      if (raw(i) != 0)
+      if (raw(i) != trueZero)
         return false;
     }
 
--- a/tests/unittests/tensorsTest.cpp
+++ b/tests/unittests/tensorsTest.cpp
@@ -385,7 +385,7 @@ TEST(Tensor, reset) {
     EXPECT_EQ(H.raw(i), 0.0);
   }
   for (size_t i = 0; i < 4 * 7 * 3 * 8; i++) {
-    EXPECT_EQ(QH.raw(i), 0);
+    EXPECT_EQ(QH.raw(i), QA.getType().getOffset());
   }
 }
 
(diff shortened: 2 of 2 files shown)


include/glow/Base/Tensor.h
```diff
@@ -93,8 +93,26 @@ class Tensor final {
 
   /// Set the content of the tensor to zero.
   void zero() {
-    std::fill(&getData()[0], &getData()[0] + size() * type_.getElementSize(),
-              0);
+    // quantized tensors should go to their offset
```
jfix71: nit: Capitalize sentence and end with period.

include/glow/Base/Tensor.h
```diff
@@ -93,8 +93,26 @@ class Tensor final {
...
+    case ElemKind::Int16QTy: {
+      auto *data = reinterpret_cast<int16_t *>(getData());
+      std::fill(&data[0], &data[0] + size(), (int16_t)type_.getOffset());
+    } break;
+    case ElemKind::Int32QTy: {
+      auto *data = reinterpret_cast<int32_t *>(getData());
+      std::fill(&data[0], &data[0] + size(), (int32_t)type_.getOffset());
+    } break;
+    default:
+      // not quanitized
```
jfix71: nit: Capitalize sentence and end with period.

tests/unittests/tensorsTest.cpp
```diff
@@ -652,6 +652,43 @@ TEST(Tensor, insertWithCountAndAxis) {
...
 }
 
+TEST(Tensor, zeroQuanitizedTensor) {
+  Tensor Q8T(ElemKind::Int8QTy, {3, 4, 5, 6}, 127, 0);
+  Tensor Q16T(ElemKind::Int16QTy, {3, 4, 5}, 1234.7, 223);
+  Tensor Q32T(ElemKind::Int32QTy, {3, 4}, 500.4, 53452);
+
+  auto Q8H = Q8T.getHandle<int8_t>();
+  EXPECT_TRUE(Q8H.isZero());
+  for (size_t i = 0; i < 3 * 4 * 5 * 6; i++) {
```
jfix71: Here and other loops below:

```suggestion
  for (size_t i = 0, e = Q8H.size(); i < e; i++) {
```

tests/unittests/tensorsTest.cpp
```diff
@@ -652,6 +652,43 @@ TEST(Tensor, insertWithCountAndAxis) {
   }
 }
```
jfix71: Can you add a doxygen comment here explaining what the test is doing?

tests/unittests/tensorsTest.cpp
```diff
@@ -652,6 +652,43 @@ TEST(Tensor, insertWithCountAndAxis) {
...
 
+TEST(Tensor, zeroQuanitizedTensor) {
+  Tensor Q8T(ElemKind::Int8QTy, {3, 4, 5, 6}, 127, 0);
+  Tensor Q16T(ElemKind::Int16QTy, {3, 4, 5}, 1234.7, 223);
+  Tensor Q32T(ElemKind::Int32QTy, {3, 4}, 500.4, 53452);
+
+  auto Q8H = Q8T.getHandle<int8_t>();
+  EXPECT_TRUE(Q8H.isZero());
+  for (size_t i = 0; i < 3 * 4 * 5 * 6; i++) {
+    EXPECT_EQ(Q8H.raw(i), 0);
```
jfix71: Perhaps you can get the offset from the Tensor's type here and below instead of explicitly writing out the offset? That way it's quite obvious what's going on.

jfix71: Huh -- what's the bug? Can you create an issue to track it?

An alternative could be `const int32_t offsetQ8 = 0;` and so on, and then you can use it everywhere (including in initializing each of the tensors).

nickgg: No bug currently. I just mean hypothetically if there was a bug in getOffset in the future, since we use it both to write the 'zero' into the buffer and to check that the buffer is 'zero', this test wouldn't give any signal that it was wrong. Using a constant gives us a little more coverage.

I like your alternative though, much clearer.

tests/unittests/tensorsTest.cpp
```diff
@@ -652,6 +652,50 @@ TEST(Tensor, insertWithCountAndAxis) {
   }
 }
 
+/// Verify that tensors that are quantized begin zero'd to their type's offset
+/// and are reset back to that offset.
+TEST(Tensor, zeroQuanitizedTensor) {
```
rdzhabarov: typo in `zeroQuanitizedTensor`

rdzhabarov: can you also add a test case where you set values to the tensor to be equal to the offset and isZero returns true?

PR #2086, opened 2018-11-27
[Quantization] Allow the input and output of a quantized AvgPool node have different quantization param.

beicy:
*Description*:
In our previous quantization procedure, we forced the quantized avgpool node output have the same quantization param as the input, which is not necessary. This PR removes this restriction.

*Testing*:
1. added unittest.
2. run 

 ```
>$./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data -dump_profile="profile.yaml"
Model: resnet50_profile
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5823)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9616)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9902)

>$ ./bin/image-classifier tests/images/imagenet/*.png -image_mode=0to1 -m=resnet50 -model_input_name=gpu_0/data -load_profile="profile.yaml"
Model: resnet50
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5964)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9521)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9887)
```
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -180,7 +180,7 @@ static bool verifyFullyConnected(NodeValue src, NodeValue weights,
 static bool verifyPool(NodeValue src, NodeValue dest,
                        llvm::ArrayRef<unsigned_t> kernels,
                        llvm::ArrayRef<unsigned_t> strides,
-                       llvm::ArrayRef<unsigned_t> pads) {
+                       llvm::ArrayRef<unsigned_t> pads, bool isAvgPool = true) {
   const Node *parent = dest.getNode();
   ShapeNHWC idim = ShapeNHWC(src.getType()->dims());
   ShapeNHWC odim = ShapeNHWC(dest.getType()->dims());
@@ -200,7 +200,14 @@ static bool verifyPool(NodeValue src, NodeValue dest,
   ShapeNHWC exp(idim.n, outSz.first, outSz.second, idim.c);
   isValid &=
       expectCompareTrue("Unexpected output dimensions", exp, odim, parent);
-  isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());
+
+  // For quantized AvgPool, the scale and offset of its input and output could
+  // be different. But for quantized MaxPool, the scale and ofset of its input
+  // and output should be the same.
+  isValid &= checkSameIsQuantized(src.getType(), dest.getType(), parent);
+  if (!isAvgPool) {
+    isValid &= checkTypeIgnoreShape(src, dest, parent);
+  }
   return isValid;
 }
 
@@ -327,7 +334,8 @@ bool ConvertToNode::verify() const {
 }
 
 bool MaxPoolNode::verify() const {
-  return verifyPool(getInput(), getResult(), Kernels_, Strides_, Pads_);
+  return verifyPool(getInput(), getResult(), Kernels_, Strides_, Pads_,
+                    /* isAvgPool */ false);
 }
 
 bool AvgPoolNode::verify() const {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tlepley-cadence** [comment]
As a general comment, I think that the choice of forcing the output scale/offset to be the same as the input should be a backend decision, not a fixed choice by the Glow framework.It is generally an accuracy vs speed tradeoff that IMO should be given to the backend designer. 

The Glow framework could simply authorize any scale/offset for inputs/outputs and it would be up to each backend to insert scale nodes in case it wants to force the same quantization params for the input and the output.

File: lib/Graph/Nodes.cpp [hunk 1]
```diff
@@ -327,7 +334,7 @@ bool ConvertToNode::verify() const {
 }
 
 bool MaxPoolNode::verify() const {
-  return verifyPool(getInput(), getResult(), Kernels_, Strides_, Pads_);
+  return verifyPool(getInput(), getResult(), Kernels_, Strides_, Pads_, false);
```
jfix71: nit: `/* isAvgPool /* false`

lib/Graph/Nodes.cpp
```diff
@@ -341,7 +348,7 @@ bool MaxPoolGradNode::verify() const {
       getOriginalOutputForResult(), getGradOfOriginalOutputNamedResult(), this);
   isValid &= verifyPool(getGradOfInputNamedInput(),
                         getGradOfOriginalOutputNamedResult(), Kernels_,
-                        Strides_, Pads_);
+                        Strides_, Pads_, false);
```
jfix71: nit: `/* isAvgPool /* false`

nadavrot:
@tlepley-cadence I agree.

lib/Graph/Nodes.cpp
```diff
@@ -180,7 +180,7 @@ static bool verifyFullyConnected(NodeValue src, NodeValue weights,
 static bool verifyPool(NodeValue src, NodeValue dest,
                        llvm::ArrayRef<unsigned_t> kernels,
                        llvm::ArrayRef<unsigned_t> strides,
-                       llvm::ArrayRef<unsigned_t> pads) {
+                       llvm::ArrayRef<unsigned_t> pads, bool isAvgPool = true) {
```
rdzhabarov: i have a dejavu that we've already changed this :D

lib/Graph/Nodes.cpp
```diff
@@ -200,7 +200,14 @@ static bool verifyPool(NodeValue src, NodeValue dest,
   ShapeNHWC exp(idim.n, outSz.first, outSz.second, idim.c);
   isValid &=
       expectCompareTrue("Unexpected output dimensions", exp, odim, parent);
-  isValid &= checkTypeIgnoreShape(src, dest, dest.getNode());
+  if (src.getType()->isQuantizedType() && isAvgPool) {
```
rdzhabarov: can you do it:

```
isValid &= checkSameIsQuantized(src.getType(), dest.getType(), parent);

if (!isAvgPool) {
  isValid &= checkTypeIgnoreShape(src, dest, parent);
}
```

beicy:
@jfix71 The reason we would like to force quantized maxpool to have the same scale and offset for input and output is related to performance :  If we force output scale and offset the same with the input, during the implantation, no dequantization/requantization is needed. We only need to compare the quantized data directly. In terms of the accuracy,  for maxpool the output range is included in the input range, so using the same scale and offset won't affect the accuracy.

lib/Graph/Nodes.cpp
beicy: Yes, in loading quantized model :P  Since this PR also need this part, I added it as well. Based on which one is merged first, the rest will be rebased :)

PR #2089, opened 2018-11-28
[image] clean up after failing to read a png image

### LINKED ISSUE #366: readPngImage() leaks memory when image fails to read
Also, we don't take into account what was returned from readPngImage(). E.g. must wrap readPngImage() call into assert: GLOW_ASSERT(!readPngImage() ... )

nickgg:
*Description*: Destroys allocated png structs if reading the PNG file fails, preventing a memory leak.
*Testing*: added a new unittest for reading corrupt and missing PNG files.
*Documentation*:
fixes #366.


rdzhabarov:
Change LGTM, can you pull latest master and then rebase? (I see some unrelated commits here)

Issue #2053, opened 2018-11-19
Support enum members on nodes


stoklund:
Currently, nodes defined by ClassGen can only have members that are scalars or vectors of scalars. When defining backend-specific nodes, it would be really helpful if we could add node members with `enum` type. At the moment, we do this by adding an `MemberType::Unsigned` member and casting to and from the real `enum` type. This is inviting type errors.

I suggest an interface like this:

```c++
auto frob = BB.newEnum("mybackend::Frob");

BB.newNode("MyPool")
    .addInput("Input")
    .addMember(frob, "FrobA")
    .addMember(frob, "FrobB")
    .addMember(MemberType::VectorUnsigned, "Kernels")
```

The `Builder::newEnum()` method should take enough arguments for ClassGen to generate a correct forward decl for the `enum class` type.

nadavrot:
We had this feature in ClassGen but we removed it last year. We had an arithmetic node that had an enum that specified if the node was Add, Sub, Mul, etc. It could be useful to search git log for the implementation.

jfix71:
@nadavrot We still have it ([see here](https://github.com/pytorch/glow/blob/master/tools/ClassGen/NodeBuilder.h#L102)), though it's currently unused. I believe it's only useful for adding a single enum with many different cases to a single node, e.g. as you mentioned for an Arithmetic node with Add, Sub, etc. modes, where those modes are enumerated in the node definition.

What @stoklund is suggesting uses enums that are externally defined. Then these enums can be reused across different nodes, and don't need to be enumerated on every node that uses them. Nodes could also have multiple enums.

stoklund:
> Then these enums can be reused across different nodes, and don't need to be enumerated on every node that uses them. Nodes could also have multiple enums.

I'm pretty sure we would need both of these features.

SplitInfinity:
I posted a WIP pull request for this. I have a few questions:
- do we have tests for `ClassGen`?
- do we need to support custom values? e.g.
```
enum class MyEnum {
   FIRST = 5,
   SECOND = 21,
}
```
- do we need to support customization of the underlying type of the enum? e.g.
```
enum class MyEnum: unsigned {
  ...
}
```

opti-mix:
> do we need to support custom values? e.g.

I'm pretty sure this would be useful.

> do we need to support customization of the underlying type of the enum?

I'd say it is nice to have, but not strictly needed.

opti-mix:
[...]

PR #2090, opened 2018-11-28
[NodeGen] Support enum members on nodes

SplitInfinity:
**Description**
This commit adds support for enum members on nodes. An enum type
can be defined within `NodeGen` itself, or defined externally.
In the latter case, it must be declared in `NodeGen` so
that it can be forward declared correctly when defining `Nodes` that
have members of that type.

The API looks something like this:

```
auto e = BB.newEnum("glow::greatBackend::tremendousEnum");
e.addValue("FIRST").addValue("SECOND");
BB.newNode(...).addMember(e, "EnumMember");
```

**Testing**
Tried out new features with a new autogenerated node class with four
different enum members, two of one type and two of another. The two enum
types were declared in different namespaces:

```
  auto &e = BB.newEnum("newbackend::namespace1::enum1");
  e.addEnumCase("FIRST", "first comment");
  e.addEnumCase("SECOND", "second comment");
  e.setDocstring("This is a test docstring. Lorem ipsum doloret something "
                 "something.");
  auto eType = e.getMemberTypeInfo();

  auto &f = BB.newEnum("newbackend::namespace1::namespace2::enum2");
  f.addEnumCaseWithValue("FIRST", 5);
  f.addEnumCaseWithValue("SECOND", 10);
  f.addEnumCaseWithValue("THIRD", 1);
  f.addEnumCase("FOURTH");
  f.setDocstring("This is a test docstring. Lorem ipsum doloret something "
                 "something.");
  auto fType = f.getMemberTypeInfo();

  BB.newNode("EnumMemberTest")
      .addInput("Input")
      .addResultFromCtorArg()
      .addMember(&eType, "EnumMember1")
      .addMember(&eType, "EnumMember2")
      .addMember(MemberType::Float, "someFloat")
      .addMember(&fType, "EnumMember3")
      .addMember(&fType, "EnumMember4");
```

Everything compiled and the generated code looked correct.

**Fixes**
Fixes issue #2053.

### DIFF
--- a/tools/ClassGen/MemberType.cpp
+++ b/tools/ClassGen/MemberType.cpp
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "MemberType.h"
+
+MemberTypeInfo kTypeRefTypeInfo{MemberType::TypeRef, "TypeRef", "TypeRef",
+                                "TypeRef"};
+MemberTypeInfo kFloatTypeInfo{MemberType::Float, "float", "float", "float"};
+MemberTypeInfo kUnsignedTypeInfo{MemberType::Unsigned, "unsigned_t",
+                                 "unsigned_t", "unsigned_t"};
+MemberTypeInfo kBooleanTypeInfo{MemberType::Boolean, "bool", "bool", "bool"};
+MemberTypeInfo kStringTypeInfo{MemberType::String, "std::string", "std::string",
+                               "std::string"};
+MemberTypeInfo kVectorFloatTypeInfo{MemberType::VectorFloat,
+                                    "llvm::ArrayRef<float>",
+                                    "std::vector<float>", "std::vector<float>"};
+MemberTypeInfo kVectorUnsignedTypeInfo{
+    MemberType::VectorUnsigned, "llvm::ArrayRef<unsigned_t>",
+    "std::vector<unsigned_t>", "std::vector<unsigned_t>"};
+MemberTypeInfo kVectorSizeTTypeInfo{
+    MemberType::VectorSizeT, "llvm::ArrayRef<size_t>", "std::vector<size_t>",
+    "std::vector<size_t>"};
+MemberTypeInfo kVectorNodeValueTypeInfo{
+    MemberType::VectorNodeValue, "NodeValueArrayRef", "std::vector<NodeHandle>",
+    "std::vector<NodeValue>"};
--- a/tools/ClassGen/NodeBuilder.cpp
+++ b/tools/ClassGen/NodeBuilder.cpp
@@ -16,6 +16,175 @@
 
 #include "NodeBuilder.h"
 
+EnumBuilder::EnumBuilder(std::ofstream &C, const std::string &name,
+                         bool isBackendSpecific)
+    : fullyQualifiedName_(name), cStream(C),
+      isBackendSpecific_(isBackendSpecific) {
+  // Parse out and store namespaces if any have been provided by repeatedly
+  // searching for tokens delimited by the scope resolution operator "::"".
+  const std::string kScopeOperator("::");
+
+  size_t namespaceStart = 0;
+  size_t namespaceEnd = name.find(kScopeOperator, namespaceStart);
+
+  while (namespaceEnd != std::string::npos) {
+    // Extract and store the current outermost namespace.
+    const size_t namespaceLength = namespaceEnd - namespaceStart;
+    namespaces_.emplace_back(name.substr(namespaceStart, namespaceLength));
+
+    // Look for the next namespace.
+    namespaceStart = namespaceEnd + kScopeOperator.size();
+    namespaceEnd = name.find(kScopeOperator, namespaceStart);
+  }
+
+  // Once the scope resolution operator is no longer found, that which was
+  // thought to be the beginning of the next namespace must be the beginning
+  // of the enum class name instead.
+  const size_t enumNameStart = namespaceStart;
+  name_ = name.substr(enumNameStart);
+
+  // Sanity check that there are no scope resolution operators in the parsed
+  // enum name.
+  assert(name_.find(kScopeOperator) == std::string::npos);
+}
+
+EnumBuilder &EnumBuilder::addEnumCase(const std::string &name,
+                                      const std::string &doc) {
+  enumCases_.emplace_back(EnumCase{name, unusedValue_++, doc, false});
+  return *this;
+}
+
+EnumBuilder &EnumBuilder::addEnumCaseWithValue(const std::string &name,
+                                               const int value,
+                                               const std::string &doc) {
+  // First, check if the value is already in use. If it is, replace the old
+  // use of it with unusedValue_.
+  for (auto &enumCase : enumCases_) {
+    if (!enumCase.definedWithValue && (enumCase.value == value)) {
+      enumCase.value = unusedValue_++;
+      break;
+    }
+  }
+
+  enumCases_.emplace_back(EnumCase{name, value, doc, true});
+
+  // Maintain the invariant that unusedValue_ is always a value that is NOT
+  // used by the enum.
+  unusedValue_ = std::max(unusedValue_, value + 1);
+  return *this;
+}
+
+MemberTypeInfo EnumBuilder::getMemberTypeInfo() const {
+  // Emit the forward declaration into a string so that it can be packed into
+  // a MemberTypeInfo.
+  std::stringstream ss;
+  emitForwardDecl(ss);
+  std::string forwardDecl = ss.str();
+
+  return MemberTypeInfo{MemberType::Enum, fullyQualifiedName_,
+                        fullyQualifiedName_, fullyQualifiedName_, forwardDecl};
+}
+
+EnumBuilder::~EnumBuilder() {
+  // If the enum is backend specific, do not print out a definition. It must
+  // be defined elsewhere in the same namespace it was declared with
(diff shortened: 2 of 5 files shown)


tools/ClassGen/NodeBuilder.h
```diff
@@ -229,6 +308,18 @@ class Builder {
     return NodeBuilder(hStream, cStream, dStream, name, isBackendSpecific);
   }
 
+  /// Declare a new enum and generate code for it.
+  EnumBuilder *newEnum(const std::string &name) {
```
SplitInfinity: I made this return a pointer because my understanding of r-values and move semantics is not strong enough for me to express this in those terms. If I make this return an `EnumBuilder`, the destructor gets called twice (once for for destroying the copy made by this function, another time for destroying the copy that resides in the caller's scope) which causes the enum definition to be emitted twice.

Some help expressing this without pointers would be awesome.

tools/ClassGen/NodeBuilder.h
```diff
@@ -229,6 +308,18 @@ class Builder {
...
+    const bool isBackendSpecific = false;
+    return new EnumBuilder(cStream, name, isBackendSpecific);
+  }
+
+  /// Declare a new backend specific enum. This will NOT generate a definition.
+  EnumBuilder *newBackendSpecificEnum(const std::string &name) {
```
SplitInfinity: Same here.

opti-mix:
@SplitInfinity Thanks for working on this feature! It is going to be very useful.

I've got a couple of requests for you.

The easy ones:
- Could you allow for setting doc-strings for enumerators as well?
  May be using an API like `addCase(std::string enumeratorName, std::string enumeratorDoc = "")`, so that the doc-string is optional.

- Could you allow for specifying a numeric value for the enumerators? 
  May be using an API like `addCaseWithValue(std::string enumeratorName, int enumeratorValue)` or something like this?

The more complex ones:
- Regarding using the enums declared externally: you say that they need to be forward declared. Does it need to be done using NodeGen APIs? Or do you need to write (or include) a usual C++ forward declaration? Do you need to/Can you include any backend-specific header files defining the actual type? Can your approach be used only for enums or could it be extended to support any other externally defined types? It could be very useful to be able to use externally defined types as types of the members.

- We may need to add the same features to the InstrGen, because e.g. when we lower a node into a backend-specific instruction, we most likely want to propagate the members of custom types into this instruction.

tools/ClassGen/NodeBuilder.h
```diff
@@ -45,6 +110,8 @@ class NodeBuilder {
   std::vector<std::pair<std::string, std::string>> nodeOutputs_;
   /// A list of node members. Format: (type, name).
   std::vector<std::pair<MemberType, std::string>> members_;
+  /// A list of external enum node members. Format: (EnumBuilder*, name).
+  std::vector<std::pair<const EnumBuilder *, std::string>> enumMembers_;
```
opti-mix: You add enum members to this separate list, different from `members_`. Therefore, the order of members in the generated class will be different from the order of members as it is provided in the NodeGen. May be such members should be added to the `members_`? E.g. `MemberType` would become something more complex than a simple enum describing a predefined set of types and  `EnumBuilder` should inherit from `MemberType`.

tools/ClassGen/NodeBuilder.h
jackm321: I'm not sure I understand your comment exactly but if you return an `EnumBuilder` from this function directly it should be no problem and then use an rvalue reference for addMember like `NodeBuilder &addMember(EnumBuilder &&eb, const std::string &name)` (and also move into `enumMembers_ ` inside of addMember) it should fix the problem.

jackm321: This will only allow an enum to be used in one node though so it may not be what you want.

nadavrot:
Nice!

tools/ClassGen/NodeBuilder.h
SplitInfinity: Yeah, one of the goals of this change is to allow multiple nodes to have a member of the same type of enum, so this solution is not going to work.

SplitInfinity:
Changes since last push:
- added support for enumeration documentation
- added support for custom enumeration values

tools/ClassGen/NodeBuilder.cpp
```diff
@@ -16,6 +16,125 @@
...
+  const size_t enumNameStart = namespaceStart;
+  name_ = name.substr(enumNameStart);
+
+  // Sanity check that there are no scope resolution operators in the parsed
+  // enum name.
+  assert(name_.find(kScopeOperator) == std::string::npos);
+}
+
+EnumBuilder &EnumBuilder::addEnumeration(const std::string &name,
+                                         const std::string &doc) {
```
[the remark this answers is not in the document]
SplitInfinity: It does have a default value and it is the empty string. It is specified in the declaration of the function in `NodeBuilder.h`.

tools/ClassGen/NodeBuilder.cpp
```diff
@@ -16,6 +16,125 @@
...
+  enumerations_.emplace_back(std::make_tuple(name, unusedValue_++, doc));
+  return *this;
+}
+
+EnumBuilder &EnumBuilder::addEnumerationWithValue(const std::string &name,
+                                                  const int value,
+                                                  const std::string &doc) {
```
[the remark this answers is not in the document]
SplitInfinity: It does have a default value and it is the empty string. It is specified in the declaration of the function in `NodeBuilder.h`.

tools/ClassGen/NodeBuilder.h
```diff
@@ -30,6 +30,73 @@
...
+              bool isBackendSpecific);
+
+  /// Set the documentation string. Each line will be prepended with "/// ".
+  EnumBuilder &setDocstring(const std::string &docstring) {
+    docstring_ = docstring;
+    return *this;
+  }
+
+  /// Add a new enumeration to the enumeration.
+  EnumBuilder &addEnumeration(const std::string &name,
```
opti-mix: I'd call it `addCase` or `addEnumerator`. Enumeration is usually used to refer to the whole enum.

SplitInfinity: Okay, I've fixed this locally and it will be included with my next push of major revisions.

SplitInfinity:
Changes since last push:
- renamed `addEnumeration...` to `addEnumCase...`
- modified `MemberType` so that the order of constructor arguments matches the declaration order in `NodeGen`. I'd like some feedback on the way in which I implemented this before I modify `InstrGen` to comply with these changes.

tools/ClassGen/MemberType.cpp
```diff
@@ -0,0 +1,38 @@
...
+DEFINE_BASIC_MEMBER_TYPE(String, "std::string", "std::string", "std::string")
+DEFINE_BASIC_MEMBER_TYPE(VectorFloat, "llvm::ArrayRef<float>",
+                         "std::vector<float>", "std::vector<float>")
+DEFINE_BASIC_MEMBER_TYPE(VectorUnsigned, "llvm::ArrayRef<unsigned_t>",
+                         "std::vector<unsigned_t>", "std::vector<unsigned_t>")
+DEFINE_BASIC_MEMBER_TYPE(VectorSizeT, "llvm::ArrayRef<size_t>",
+                         "std::vector<size_t>", "std::vector<size_t>")
+DEFINE_BASIC_MEMBER_TYPE(VectorNodeValue, "NodeValueArrayRef",
+                         "std::vector<NodeHandle>", "std::vector<NodeValue>")
+
```
opti-mix: Looks like you do not win a lot by using macro definitions here. You could just as well define these types directly.

tools/ClassGen/NodeBuilder.h
```diff
@@ -30,6 +30,73 @@
...
+
+class EnumBuilder {
+  /// The name of the enum.
+  std::string name_;
+  /// The fully qualified name of the enum (i.e. with all enclosing namespaces).
+  std::string fullyQualifiedName_;
+  /// The containing namespaces for the enum.
+  std::vector<std::string> namespaces_;
+  /// The enum case names, values, and their corresponding docstrings.
+  std::vector<std::tuple<std::string, int, std::string>> enumCases_;
```
bertmaher: I'd kinda prefer a struct with {name, value, docstring} over std::tuple.

tools/ClassGen/NodeBuilder.h
bertmaher: I kinda feel like this is a symptom of an overly-clever use of the destructor to emit code... I wouldn't mind an explicit `emit()` method.

tools/ClassGen/NodeBuilder.cpp
```diff
@@ -16,6 +16,169 @@
...
+  return *this;
+}
+
+EnumBuilder &EnumBuilder::addEnumCaseWithValue(const std::string &name,
+                                               const int value,
+                                               const std::string &doc) {
+  enumCases_.emplace_back(std::make_tuple(name, value, doc));
+  // Maintain the invariant that unusedValue_ is always a value that is NOT
+  // used by the enum.
+  unusedValue_ = std::max(unusedValue_, value + 1);
```
jackm321: This may not matter much but this doesn't prevent multiple cases with the same value if someone makes n cases with `addEnumCase()` then calls `addEnumCaseWithValue()` with value < n expecting that this will automatically avoid conflicts. Is it worth it to keep a set of values and assert that new cases aren't in that?

jackm321: Even just `assert(value >= unusedValue_)` would work then just make `unusedValue_ = value + 1`

tools/ClassGen/NodeBuilder.h
SplitInfinity: Fixed in latest push.

tools/ClassGen/NodeBuilder.cpp
SplitInfinity: I sort of fixed this by scanning for and fixing duplicate values in `addEnumCaseWithValue` if a user-defined value conflicts with a value assigned by the tool. It's a linear time method instead of a constant time method now, though.

SplitInfinity:
Changes since last push:
- replace `std::tuple` used to represent enum cases with an `EnumCase` struct
- handled case in which `addEnumCaseWithValue` is called with a value that has already been assigned by `EnumBuilder` during a previous call to `addEnumCase`. This way, mixing `addEnumCase` and `addEnumCaseWithValue` will "just work", but the tool will silently generate legal code if `addEnumCaseWithValue` is called twice with the same value and different names.

SplitInfinity:
@opti-mix 

Seems like a good candidate for LIT.

nadavrot:
Nice!

PR #2091, opened 2018-11-28
[ONNXModelLoader][Easy] Refactor ONNXModelLoader::loadOperator()

jackm321:
*Description*:
Pull the code for loading operators into individual functions like in `CommonOperatorLoader`. This will make adding additional checks before calling graph creation functions a bit cleaner
*Testing*:
`ninja check`
*Documentation*:
doxygen


jackm321:
This PR should only include 3171ce4 but this branch need to contain the changes from the branch from github.com/pytorch/glow/pull/2079. We can ignore this until that's landed. Does anyone know what the best way to represent stacked changes with pull requests is? I want to avoid adding even more code to 2079.

rdzhabarov:
@jackm321 can you rebased on the latest master with https://github.com/pytorch/glow/commit/4014bde87ac1e45aec7a3a039ece1473dceff94a in.

jackm321:
@rdzhabarov rebase done

nadavrot:
@jackm321 Some folks use the label (NFC) in the title. This means No Functional Change. This signals that the changes are just mechanical, like moving code around.

include/glow/Importer/ONNXModelLoader.h
```diff
@@ -53,6 +53,55 @@ class ONNXModelLoader
...
+
+  // Load Slice ONNX operator.
+  llvm::Error loadSlice(const ONNX_NAMESPACE::NodeProto &op,
+                        const ArgumentDictionaryTy &dict);
+
+  // Load Conv ONNX operator.
+  llvm::Error loadConv(const ONNX_NAMESPACE::NodeProto &op,
+                       const ArgumentDictionaryTy &dict);
+
+  // Load MaxPool or AveragePool ONNX operator.
```
rdzhabarov: can you add comments on the params which are not super obvious, like typeName here and in the rest of the methods where it's applicable.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -206,479 +206,545 @@ static llvm::Error loadTensor(const ONNX_NAMESPACE::TensorProto &in,
...
-    */
-
-    const auto &name = op.output(0);
-    // If the tensor is pre-populated by the user of this class then we don't
-    // need to allocate a new tensor.
-    if (tensors_.count(name)) {
-      return llvm::Error::success();
-    }
+  auto *T = new Tensor();
+  RETURN_IF_ERR(loadTensor(dict.at("value")->t(), T));
```
rdzhabarov: in case of returning here how do we free the memory for `auto *T`?

yinghai: What don't we use RAII instead of raw pointers?

jackm321: @rdzhabarov great catch thank you, I'll fix this here
@yinghai Yeah I've also been wondering this as well. I think we ought to be using unique_ptrs here and I don't see any reason so far not to. @rdzhabarov thoughts on this?

jackm321: #2115

yinghai:
LgTM

jackm321:
@rdzhabarov any more changes needed?

lib/Importer/ONNXModelLoader.cpp
rdzhabarov: Agree, it's better to have unique ptr to control the life time of that pointer. Can be done separately from this change.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -206,479 +206,548 @@ static llvm::Error loadTensor(const ONNX_NAMESPACE::TensorProto &in,
...
+  RETURN_ERR_IF_NOT(dict.count("ends"),
+                    "Slice: attribute 'ends' is mandatory.");
+  auto starts = getShape<ssize_t>(dict.at("starts"));
+  auto ends = getShape<ssize_t>(dict.at("ends"));
+  RETURN_ERR_IF_NOT(
+      (starts.size() == ends.size()),
+      "Slice: 'starts' and 'ends' arrays must have the same size.");
+
+  // Attribute 'axes' is optional.
+  std::vector<ssize_t> axes;
```
rdzhabarov: I'd prefer to stick to one DS when it makes sense (e.g., smallvector)

PR #2096, opened 2018-11-29
[Caffe2ModelLoader] Load GivenTensorIntFill using int32_t

jfix71:
*Description*: GivenTensorIntFill was using `int64_t`, but should be using `int32_t`.

*Testing*: Added a test for all the different supported TensorFills.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -579,23 +579,30 @@ void Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
     auto dim = getShape(dict["shape"]);
 
     size_t i = 0;
+#define LOAD_TENSOR_FILL(TYPE_NAME, NATIVE_TYPE, PROTO_TYPE_NAME)              \
+  T->reset(ElemKind::TYPE_NAME, dim);                                          \
+  auto TH = T->getHandle<NATIVE_TYPE>();                                       \
+  for (auto num : dict["values"]->PROTO_TYPE_NAME()) {                         \
+    TH.raw(i++) = num;                                                         \
+  }
+
     if (dict["values"]->floats_size()) {
       assert(typeName != "GivenTensorIntFill" &&
              typeName != "GivenTensorInt64Fill");
-      T->reset(ElemKind::FloatTy, dim);
-      auto TH = T->getHandle<>();
-      for (auto num : dict["values"]->floats()) {
-        TH.raw(i++) = num;
-      }
+      LOAD_TENSOR_FILL(FloatTy, float, floats);
     } else if (dict["values"]->ints_size()) {
-      T->reset(ElemKind::Int64ITy, dim);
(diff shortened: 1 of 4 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -579,28 +579,35 @@ void Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
     auto dim = getShape(dict["shape"]);
 
     size_t i = 0;
+#define LOAD_TENSOR_FILL(TYPE_NAME, NATIVE_TYPE, PROTO_TYPE_NAME)              \
+  T->reset(ElemKind::TYPE_NAME, dim);                                          \
+  auto TH = T->getHandle<NATIVE_TYPE>();                                       \
+  for (auto num : dict["values"]->PROTO_TYPE_NAME()) {                         \
```
bertmaher: If a `protobuf::RepeatedField` is convertible to `llvm::ArrayRef` this could be `TH = dict["values"]->PROTO_TYPE_NAME();` which looks cool.  But it's probably not :)

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -579,28 +579,35 @@ void Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
...
-      auto TH = T->getHandle<int64_t>();
-      for (auto num : dict["values"]->ints()) {
-        TH.raw(i++) = num;
+      if (typeName == "GivenTensorIntFill") {
+        LOAD_TENSOR_FILL(Int32ITy, int32_t, ints);
+      } else if (typeName == "GivenTensorInt64Fill" ||
+                 typeName == "GivenTensorFill") {
+        LOAD_TENSOR_FILL(Int64ITy, int64_t, ints);
+      } else {
+        unexpectedNodeError(op, "Unsupported data type for GivenTensor.*Fill.");
```
bertmaher: Since we have the opname can we use it here instead of "GivenTensor.*Fill"?

nadavrot:
This may conflict with the PR that @jackm321 is pushing. We may want to coordinate who lands first to minimize the merge pain.

lib/Importer/Caffe2ModelLoader.cpp
jfix71: Yup, not convertible :/

jfix71:
@jackm321 Do you mind if I land this before https://github.com/pytorch/glow/pull/2079? It looks like it already has a merge conflict with `Caffe2ModelLoader.cpp` like this PR will, and I think this PR adds just a few places where `RETURN_ERR`/`RETURN_ERR_IF_NOT`/`RETURN_SUCCESS` need to be used. Also can wait for that to land first it's an issue for you.

jackm321:
@jfix71 I just saw this now, glad you landed this and I'll take care of the merging from my end, thanks for the heads up :)

PR #2097, opened 2018-11-29
Use LIT to test our command line tools

qcolombet:
This PR is a proof of concept of how we could add tests for our command line tools using LLVM testing infrastructure (LIT, LLVM Integrated Tester).

`llvm-lit` is basically a tool that let you invoke arbitrary command by just typing what you want to run with `RUN` directive. Then using `CHECK` directive, it is possible to verify that something behave the way it should. `llvm-lit` also supports things like conditionally running thing (like this test runs only in debug mode) and takes advantages of running tests in parallel and whatnot.

Like the name of the tool suggests, this depends on LLVM and unfortunately this tool is not part of the release package, thus we add a new sort of dependency. Basically, what I did in this PR is the lit tests are enabled only if `llvm-lit` and `FileCheck` are in the path (FileCheck is part of the LLVM package that we already import).

Long story short, this PR would need some cleaning to be properly productized, but it illustrates how the things work. Basically, each time we would add a new command line test, we would need to create a `.test` file in `tests/unittests` (we can refine the directory structure after the fact) and add a couple of `RUN` and `CHECK` lines and done.

Then, the test is automatically added.

Rerun a test is relatively easy, using the `llvm-lit` command (that you can get from ninja -v litTests), one can add `-v` to get the actual command lines that were invoked by lit for the failing tests (or `-a` for all the tests).

@jfix71 I spent some time on this to address what we were discussing at some point in an issue or PR that it would be nice to make sure those things down break.
Could you give it a try?

Right now, the dependency checking and whatnot are missing, so to play with it you need to:
- Clone LLVM
- Do a quick build of LLVM (e.g., ninja llc), just to get `llvm-lit` created
- Add `FileCheck` and `llvm-lit` in your path
- Clone glow
- Configure glow in release mode (I run the lit test only in release since text-translator can be fairly slow otherwise)
- Extract the en2gr model in your build directory
- Run `ninja check` or `ninja text-translator && ninja litTests`

From that point, you should be able to play with the RUN and CHECK lines in `tests/unittests/text-translator-quantization.test` or add your own `.test` files if you want to add more stuff. (Right now I only added the proper path expansion for `text-translator`, if you want to play with say `image-classifier`, you'll have to use absolute path or add the expand in the lit.cfg.py file).

Bottom line to all is play with it and if you like it, I can productize it.

Note: I realized after I hacked this that it may overlap with what @bertmaher is proposing in #2081. I happy to drop that PR, no hard feelings ;).


bertmaher:
Well this is interesting!  I hadn't worked on that PR in a few days since it wasn't a burning need, but this showed up right as I was getting back to it :)

I kind of see advantages both ways.  The LIT tests are pretty explicit about the command line that's being tested, which is nice.  And the Python factors out the parameters for the models, which is also nice.  But maybe we could get some of the latter benefits by extending the templating harness used here?

I'm not strongly attached to my Python script so happy to discuss further!

qcolombet:
> And the Python factors out the parameters for the models, which is also nice. But maybe we could get some of the latter benefits by extending the templating harness used here?

That's definitely possible (assuming I understand what this means :P). LIT allows you to specify parameters that you want to apply to all commands or kind of commands.
E.g., in the cfg script we can decide that all the instances of `text-translator` would run with `-verify-module` (assuming we would have such an option).

Anyhow, we should probably discuss what are the needs in person :).

Note: We may not want to this extreme, but just to mention that in LLVM, LIT is actually used to also drive the googletests. Thus, if we were to go down that road, that means we would get the concurrent execution of our gtests for free!

tests/unittests/text-translator-quantization.test
```diff
@@ -0,0 +1,5 @@
+RUN: %text-translator -m %build_dir/en2gr  -keep-original-precision-for-nodes=Add -dump-profile=%t.yaml -quantization-schema=symmetric <<< "My name is Bob ." && \
+RUN: (%text-translator -m %build_dir/en2gr  -keep-original-precision-for-nodes=Add -load-profile=%t.yaml <<< "My name is Bob ." | FileCheck %s)
```
qcolombet: In case someone wonders, `%t` tells LIT to generate a temporary and unique name for this test.
Thus, it is perfect to store intermediate files like quantization profile.

qcolombet: `%s` means the source file, i.e., this file.
So `... | FileCheck %s` means run `FileCheck` with the output of `...` and with this file as input. `FileCheck` is the tool that looks for `CHECK` directives and do the pattern matching.

nadavrot:
This is awesome!

bertmaher:
Having slept on it, this is probably strictly better than a custom python script.  I'm going to close my PR in favor of this approach.

opti-mix:
Very cool!

jfix71:
This looks amazing! I will get all the dependencies and verify it works, but assuming it works as described/shown I think it would be really valuable.

bertmaher:
> the users have to make sure that llvm-lit and FileCheck are available.

It's kind of a bummer that llvm-lit doesn't get installed, but I guess that's because it's a testing utility rather than a real llvm tool?

qcolombet:
> It's kind of a bummer that llvm-lit doesn't get installed, but I guess that's because it's a testing utility rather than a real llvm tool?

Correct!

The crude reality is that this is (a pile of) python scripts that know a lot about llvm internals (like exactly which executables are involved in the llvm test suite) and unfortunately that would be a major effort (on the llvm side) to break those dependencies.

To illustrate that, checking again, I believe what I said with a quick llvm build in the initial post (the `ninja llc` part) is not accurate. You would need to run `ninja all` to make `llvm-lit` happy, but then, in our context it won't use any of the things that it expects to see.
The good news is that you have to do it only once.

qcolombet:
Update:
- Added some tool detections
- Use that mechanism to gate the related tests against the present of a given tool

After this update if a tool is not built, e.g., text-translator in that case, lit would report the test as unsupported and wouldn't run it.

jfix71:
Tried it out a little and it works great. Thanks for working on this @qcolombet.

It would be great to eventually replace the bundle test which currently runs on CI with this. It's currently pretty hacky -- I wrote a little bash script ([here](https://github.com/pytorch/glow/blob/master/.circleci/test.sh#L17-L25)) which takes the tail of the output, greps for just the expected output, and then compares it to a [txt file](https://github.com/pytorch/glow/blob/master/.ci/lenet_mnist_expected_output.txt).

qcolombet:
I spend some time improving the dependencies on different angles:
1. We depend on `lit.py` and not `llvm-lit`, thus we don't need to build anything in llvm anymore. We just need to clone LLVM sources and add `<llvm-path-to-src>/utils/lit` in the path.
2. I added some detection mechanism to work with the `REQUIRES` keyword of lit, so that we can gate some tests with specific requirements (e.g., `cpu`, `opencl`). I've added a test case to illustrate that.
3. I untied the litTests from release mode. Individual tests can still `REQUIRES` release mode, but we can now write tests for debug mode as well

@jfix71 Could you give it another try, without building llvm this time. Just to make sure I indeed removed the dependency.

Finally, the one thing missing here is how do we specify the dependencies on the models and how we honor them?
We have several options:
- We could assume they live in a specific directory (right now this is what the patch does, but it misses a check that the model are actually present)
- We could make this a cmake parameter (GLOW_PATH_TO_MODELS or something)
- We could download the model within the tests (that would be really slow and potentially involve some duplication

The cmake parameter is the easier, but involves rerunning cmake or thinking that through when you start a new build.

qcolombet:
For the record, thanks to depending on just lit.py, I could add llvm as a submodule and just depend on this script. Therefore, anybody pulling out glow would get the lit testing without the need to think about it. The downside is that pulling the submodule takes a while (llvm is huge!)

Anyway, just mentioning this. For CI, we could totally have the initial step cloning LLVM and adding `lit.py` in the path.

nadavrot:
Quick question: it seems like FileCheck is not built by the LLVM build scripts by default. FileCheck also does not come with the pre-built LLVM binaries. Moreover, downloading and building parts of LLVM as part of the CI will increase the build time of every single build. Moreover, I am not sure how to maintain this inside the internal build system.

What if we implemented FileCheck in 100 lines of Python or C ourselves?  It should be very easy to look for the "RUN" and "CHECK" lines and just run them. We could be compatible with LLVM's FileCheck or do our own thing. It's not a big deal either way. Would maintaining 100 lines of Python be more difficult than maintaining the extra configuration and settings for loading FileCheck from LLVM?

qcolombet:
> Quick question: it seems like FileCheck is not built by the LLVM build scripts by default. FileCheck also does not come with the pre-built LLVM binaries.

That's what I saw too. I was actually surprised to have FileCheck in the release package on my mac and thought that llvm changed what they put in there. Turns out they didn't thus, this additional FileCheck build step in that WIP. But that's just WIP :).

> Moreover, downloading and building parts of LLVM as part of the CI will increase the build time of every single build. Moreover, I am not sure how to maintain this inside the internal build system.

Agree. The downloading part we won't get away from it unless we don't use `lit` at all and in that case, we should revive @bertmaher's initial approach #2081 IMHO.

> What if we implemented FileCheck in 100 lines of Python or C ourselves? It should be very easy to look for the "RUN" and "CHECK" lines and just run them.

The RUN lines are actually handled by lit, FileCheck only looks at the check lines, which is much easier to reimplement.

> We could be compatible with LLVM's FileCheck or do our own thing. It's not a big deal either way. 

Agreed. I initially went with FileCheck because I had it in my package.

> Would maintaining 100 lines of Python be more difficult than maintaining the extra configuration and settings for loading FileCheck from LLVM?

For now, it is probably cheaper to have our own FileCheck thing. I could actually modify the current tests to just a bunch of grep and diff and we're good to go. In terms of difficulty, I cannot speak for the internal build system, but building FileCheck is pretty easy. We may already build it and could depend on it.
Also FWIW, regarding the build time for this kind of testing, we spend most of our time downloading the models themselves, thus whatever solution we use, it won't really help that.

Now, for the RUN lines and temporary files and whatnot that LIT brings, reimplementing that is a different story.

qcolombet:
Refactored the logic to set `lit` and `FileCheck` so that it can be called independently.
I am going to see how that plays in our internal build before cleaning up and deciding whether we should have our own FileCheck thing.

qcolombet:
Alright, I've cleaned the commit up so that if there is still an interest to merge this, this is ready (modulo review comments of course!)

To use LIT locally things got much easier since the beginning of that PR. Now, one need to:
```
# Add lit.py and FileCheck in the PATH if they don't exist
source utils/set_llvm_test_env.sh
# The lit tests we have only run in release for now and all requires to have the models
cmake <...> -DGLOW_MODELS_DIR=<pathToDownloadedC2Models> -DCMAKE_BUILD_TYPE=Release
ninja check
```
That's it.
Note the `source` thing needs to be done just once. Meaning as long as you don't reset your path every following cmake will just work.
Moreover, rerunning `source` will just update the PATH variable properly if call from the same dir.

I know there are interest for redoing stuff `lit` and or `FileCheck` on our own to have less dependencies. (I'd argue that we already depend on LLVM but that's a different story.) However, this PR gives us something we can use right now and can replace in the future, when we have time, while addressing a real need.

bertmaher:
This is really awesome.  I was impressed that you even thought to require release builds for running the tests, because I was afraid they would slow down my debug builds!  I also found it was really easy to add lit support on my mac by just extending my CMAKE_PREFIX_PATH a bit:
```
cmake -DCMAKE_PREFIX_PATH="$HOME/src/llvm_src/utils/lit;/usr/local/opt/llvm"
```

qcolombet:
Thanks @zakk0610!

I didn’t know there were official package for those! That would be simpler indeed!
I’ll give it a shot.

qcolombet:
@nadavrot, we already depend on llvm. I don’t see how this is different.
Moreover with what @zakk0610 is suggesting that’s going to be just a couple apt-get more.

The goal here is to improve testability but why would we reinvent the wheel when battle ready solutions exists?

qcolombet:
FWIW, on the productivity side, like @bertmaher mentioned, the dependency is optional, if you don't want it, you don't pay for it. As for the mac, the brew package of llvm comes with `FileCheck` so there is nothing to do...

qcolombet:
FWIW, on the productivity side, like @bertmaher mentioned, the dependency is optional, if you don't want it, you don't pay for it. As for the mac, the brew package of llvm comes with `FileCheck` so there is nothing to do...

qcolombet:
Patched the build llvm script to include FileCheck (we were already building it, just not installing it).

rdzhabarov:
we can revert that change and merge this one. so next time we'll make sure all changes are legit.

qcolombet:
Is it okay to merge or should I add the docs update in that PR too?
(I was planning to update the docs in a separate PR so that we already get the testing for the incoming PRs.)

PR #2100, opened 2018-11-29
[ONNX] Add support for the ONNX operator in the importer

tlepley-cadence:
*Description*: Add support for the ONNX operator in the importer. This operator maps well to the existing Slice node in the IR. 
ONNX spec: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Slice
Note that ONNX doesn't define clearly the semantics in some situations. I implemented the interpretation of the spec that I find the most relevant. This interpretation is also tested by the test suite. 

*Testing*: Added tests for various alternatives

*Documentation*: N/A


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -256,6 +256,96 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+           "Slice: 'starts' and 'end' arrays must have the same size.");
+
+    // Attribute 'axes' is optional.
+    std::vector<ssize_t> axes;
+    if (dict.count("axes")) {
+      // The ONNX spec is unclear so we consider that the 'axes' array may have
+      // any size. The constraints are:
+      // - the element value must be in range [0, numDims[,
+      // - 'starts' & 'ends' arrays must have the same size as the 'axes' array.
+      // In case an axe is specified multiple times, the later parameters will
```
bertmaher: sp: "axe" -> "axis"

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -256,6 +256,96 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+    auto ends = getShape<ssize_t>(dict["ends"]);
+    assert((starts.size() == ends.size()) &&
+           "Slice: 'starts' and 'end' arrays must have the same size.");
+
+    // Attribute 'axes' is optional.
+    std::vector<ssize_t> axes;
+    if (dict.count("axes")) {
+      // The ONNX spec is unclear so we consider that the 'axes' array may have
+      // any size. The constraints are:
+      // - the element value must be in range [0, numDims[,
```
bertmaher: closing ], not [

tests/unittests/onnxImporterTest.cpp
```diff
@@ -634,3 +634,85 @@ TEST(onnx, gather) {
...
+    }
+  }
+}
+
+TEST(onnx, importSliceAxesFull) {
+  importSliceTest("sliceAxesFull.onnxtxt", "data", {2, 3, 3, 3} /* input */,
+                  {0, 1, 1, 2} /* starts */, /* ends: {1, 2, 3, 3} */
+                  {1, 1, 2, 1} /* output */);
+}
+TEST(onnx, importSliceAxesAnyOrder) {
```
bertmaher: Please add a blank line between top-level definitions.

tests/unittests/onnxImporterTest.cpp
```diff
@@ -634,3 +634,85 @@ TEST(onnx, gather) {
...
+
+static void importSliceTest(std::string fileName, const char *inputName,
+                            const llvm::ArrayRef<size_t> inputShape,
+                            const llvm::ArrayRef<size_t> starts,
+                            const llvm::ArrayRef<size_t> outputShape) {
+  ExecutionEngine EE{BackendKind::Interpreter};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  std::string NetFilename("tests/models/onnxModels/" + fileName);
```
bertmaher: Does this way of concatenating strings actually work?  It looks like it's adding two char*'s together and then making a std::string from that, which is probably not the intent...

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -256,6 +256,96 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+      newStarts[i] = 0;
+      newEnds[i] = dims[i];
+    }
+
+    // Determine the coordinates of the sub-tensor to extract.
+    assert(starts.size() == ends.size());
+    for (size_t i = 0; i < axes.size(); i++) {
+      ssize_t newStart = starts[i];
+      ssize_t newEnd = ends[i];
+      ssize_t axesId = axes[i];
```
jackm321: nit maybe this should be `axisId` if it's just the one

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -256,6 +256,96 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+        axes.push_back(ssize_t(i));
+      }
+    }
+
+    // The ONNX description is unclear and doesn't describe what to do when a a
+    // sub-set of axes index are missing in the axes array. An interpretation
+    // is that in this case, the entire range is considered for the non
+    // specified axes. Then, initialize newStarts and newEnds with the full
+    // index range for all axes.
+    std::vector<size_t> newStarts(numDims), newEnds(numDims);
```
bertmaher: Declare these on separate lines.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -256,6 +256,96 @@ bool ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+    auto dims = data.dims();
+    auto numDims = dims.size();
+
+    // Attributes 'starts' and 'ends' are mandatory and must be consistent.
+    assert(dict.count("starts") && "Slice: attribute 'starts' is mandatory.");
+    assert(dict.count("ends") && "Slice: attribute 'ends' is mandatory.");
+    auto starts = getShape<ssize_t>(dict["starts"]);
+    auto ends = getShape<ssize_t>(dict["ends"]);
+    assert((starts.size() == ends.size()) &&
+           "Slice: 'starts' and 'end' arrays must have the same size.");
```
bertmaher: Do `starts` and `ends` have to be the same size as `numDims`?  I can't tell from the spec.

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: `[` means that  `numDims` is excluded from the range. Do you have another notation in the US ?

jfix71: `[0, numDims)`

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Well, the spec is very unclear and needs many clarifications. This is my interpretation that when there is a 'starts' integer, there *must* be a 'ends 'integer. Otherwise, for instance, if one 'ends' integer is missing, we don't know which axis has been omitted for 'ends' (it can be any).

tlepley-cadence: I think `starts` and `ends` must have the same size. But their size can be smaller or greater than `numDims` (it depends on `axes`).

tests/unittests/onnxImporterTest.cpp
tlepley-cadence: Well, it works in my environment, but I agree it's a bit unclean. I fixed this.

tests/unittests/onnxImporterTest.cpp
tlepley-cadence: Done

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Done.

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Done

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: @jfix71 thanks ! I changed the notation.

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Done.

nadavrot:
@jackm321 FYI.

tests/unittests/onnxImporterTest.cpp
MattPD: FWIW, this is legal in standard C++ -- there's an overload of `operator+` which takes `char*` (here resulting from `"tests/models/onnxModels/"`) and `std::string` (here: `fileName`) and returns `std::string`; cf. the docs in http://eel.is/c++draft/string.op.plus#itemdecl:5 / overload (4) in https://en.cppreference.com/w/cpp/string/basic_string/operator%2B. (For example: https://wandbox.org/permlink/birDVHjYKqblSC5t). Admittedly, things would be different for `llvm::StringRef` (although there's a lightweight approach to concatenation using `llvm::Twine`, e.g., https://github.com/llvm-mirror/llvm/blob/release_70/lib/Passes/PassPlugin.cpp#L21, http://llvm.org/doxygen/classllvm_1_1Twine.html#aba6d9549c0ddfa25919915044a81eaf3).

lib/Importer/ONNXModelLoader.cpp
bertmaher: Ha, wow, sorry, I thought it was just a typo.  That's my i18n lesson for the day!

PR #2101, opened 2018-11-29
[graph] Utilities for hooking intermediate net layers

### LINKED ISSUE #1976: Build tools for "cosimulation" of networks

bertmaher:
*Description*:
In bringing up backends, I found it useful to be able to truncate a network at some arbitrary intermediate layer and compare the output against the interpreter.  This is (more or less) the utility I wrote to do it, plus an example of going layer-by-layer through ResNet-50 (which is very slow, but that's OK).

If you're reviewing, the first commit is the hooking logic itself; the second is the example.

I'd like some feedback on whether people think this is generally useful.  If so, happy to do some cleanup and write docs.

*Testing*: `bin/resnet-verify`

*Documentation*: not yet

Fixes #1976

### DIFF
--- a/examples/resnet-verify.cpp
+++ b/examples/resnet-verify.cpp
@@ -1,0 +1,110 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Hook.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/Format.h"
+
+using namespace glow;
+
+const char inputName[] = "gpu_0/data";
+
+class Tester {
+  Context ctx;
+  ExecutionEngine EE;
+  Module &mod;
+  Function *F;
+  TypeRef inputType;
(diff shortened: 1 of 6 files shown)


examples/resnet-verify.cpp
```diff
@@ -0,0 +1,111 @@
...
+    // Allocate memory for input and bind it to the placeholders.
+    ctx.allocate(mod.getPlaceholders());
+    updateInputPlaceholders(ctx, {input}, {batch});
+  }
+
+  TypeRef getInputType() const { return inputType; }
+
+  Function *getFunction() const { return F; }
+
+  Tensor *hookAndRun(llvm::StringRef name) {
```
opti-mix: I have a small concern here. You try to hook by name, but names are not guaranteed to be unique in general. Should we also have a variant of `hookAndRun` that takes Node* as a parameter?

opti-mix: Though, it is just an example. So, we probably don't care too much.

bertmaher: Yes, this node-naming thing is a real problem.  You need to hook "the same" node in each network, but names aren't a strong guarantee.  It would be nice to be able to hook the function by node and then clone the function into the secondary backend, but cloning from one Module to another seemed kind of tricky.

bertmaher: The hooking is at least taking place before backend-specific transformation takes place.  I'm not certain that the model loader always behaves the same regardless of backend, but it's better than e.g. after compile().

nadavrot: I think that names are uniqe, except that they name may change due to compiler transformations.

bertmaher:
Ping!  Any change requests?  Can I get an approval for this?

nadavrot:
Sorry for the delay. Approved.

PR #2102, opened 2018-11-30
[Quantization] Enable RowwiseQuantizedFullyConnected under the option -enable-rowwise

beicy:
*Description*:
This PR adds a option to enable RowwiseQuantizedFullyConnected in Glow's quantization procedure:
When a FC node needs to be quantized, if both of the conditions are satisfied, we replace the quantized FC node by RowwiseQuantizedFullyConnected:
1. The weights of this FC is constant.
2. -enable-rowwise is set or set enableRowwise param in                                                                                                        
  quantization::quantizeFunction to true. In unittest, the later one is used.
 
*Testing*:
Added unittest.
Also manually checked -enable-rowwise opt.

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Quantization/Quantization.h
+++ b/include/glow/Quantization/Quantization.h
@@ -66,13 +66,16 @@ generateNodeQuantizationInfos(Context &ctx, const Function *F,
 /// This method clones original function \p F and caller is responsible for
 /// cleaning up/erasing original function \p F if needed. Any nodes of kinds
 /// contained in \p doNotQuantizeKinds will not be quantized, even if a profile
-/// was gathered for them and the backend supports the quantized operation.
-/// \returns a new quantized function.
+/// was gathered for them and the backend supports the quantized operation. If
+/// \p enableRowwise is true, during quantization, all quantized FullyConnected
+/// nodes will be converted to RowwiseQuantizedFullyConnected. \returns a new
+/// quantized function.
 Function *
 quantizeFunction(const ExecutionEngine &EE,
                  llvm::ArrayRef<NodeQuantizationInfo> quantizationInfos,
                  Function *F, llvm::StringRef newFuncName = "",
-                 const KindSet &doNotQuantizeKinds = {});
+                 const KindSet &doNotQuantizeKinds = {},
+                 bool enableRowwise = false);
 
 } // namespace quantization
 } // namespace glow
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -565,27 +565,42 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
 RowwiseQuantizedFullyConnectedNode *
 Function::createRowwiseQuantizedFullyConnected(llvm::StringRef name,
                                                NodeValue input, Constant *W,
-                                               Node *B, TypeRef outTy) {
+                                               Node *B, TypeRef outTy,
+                                               bool transposeWeight) {
   // Since W is constant, quantize it in compilation time.
   // The quantized data is in qWeights, the scale of each row is in scales,
   // and the offset of each row is in offsets.
   Constant *weights = llvm::cast<Constant>(W);
-  size_t numRows = W->getType()->dims()[0];
+  size_t numRows =
+      transposeWeight ? W->getType()->dims()[1] : W->getType()->dims()[0];
+  size_t numCols =
+      transposeWeight ? W->getType()->dims()[0] : W->getType()->dims()[1];
 
   // So far, if we want to create a storage with Int8QTy/Int16QTy,
   // it is assumed to be quantized data and the scale and offset should be
   // provided. But for rowwise quantization, the scales and offsets are stored
   // in vectors separately, we add the dummy scale and offset here.
-  auto *qWeights = getParent()->createConstant(ElemKind::Int8QTy, W->dims(),
-                                               0.0, 0, "weights.rwqfc");
+  auto *qWeights = getParent()->createConstant(
+      ElemKind::Int8QTy, {numRows, numCols}, 0.0, 0, "weights.rwqfc");
   auto *scales =
       getParent()->createConstant(ElemKind::FloatTy, {numRows}, "scales.rwqfc");
   auto *offsets = getParent()->createConstant(ElemKind::Int32ITy, {numRows},
                                               "offsets.rwqfc");
 
+  Tensor wt;
+  if (transposeWeight) {
+    // This happens when the RowwiseQuantizedFullyConnected node is converted
+    // from a quantized FullyConnected node in Glow's quantization procedure.
+    // Since in FC, the weights is stored as transposed (i.e. I * W + B), but in
+    // RowwiseQuantizedFullyConnected, the weights is stored as it is (i.e. I *
+    // W(T) + B).
+    weights->getPayload().transpose(&wt, {1, 0});
+  } else {
+    wt.assign(&(weights->getPayload()));
+  }
+
   quantization::tensorRowwiseQuantization(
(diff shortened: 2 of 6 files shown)


beicy:
I can manually add the -enable-rowwise option as the PR description mentioned. However, I have not idea how to put this option into "ninja check" to trigger the test automatically. Any suggestion? Thanks!

bertmaher:
I think you can add a test with custom arguments by adding something like this to `tests/unittests/CMakeLists.txt`:
```
add_glow_test(quantizationTestRowwise
  ${GLOW_BINARY_DIR}/tests/quantizationTest
  -enable-rowwise
  --gtest_output=xml:quantizationTest.xml)
```

nickgg:
Why a command line argument here? If you made `enableRowwise` an argument of `quantizationFunction` (with a default arg = false) then you could just have the new test invoke it manually.

With this change running the unittests manually will always fail in some tests, right?

beicy:
@nickgg We need a command line option to enable the rowwise. if we made enableRowwise an argument of quantizationFunction, I think in everyplace quantizeFunction is invoked, we still need to check the "-enable-rowwise" (Please correct me if I am wrong). 
"With this change running the unittests manually will always fail in some tests, right?"
--- With this change, running current unittests manually or "ninja run" won't fail.  The unittest I added needs "-enable-rowwise" to run the rowwise flow. Otherwise, it is just normal quantized FC as usual.

bertmaher:
But is it necessary to force every invocation of quantizeFunction everywhere to use rowwise quantization?  It would make more sense to me to make this decision at the call site, as @nickgg suggests, and pass the setting from above.

nickgg:
@beicy it depends I think if enabling rowwise is an option that makes sense to be globally on or off for a process, if there's ever a chance that you might want some quantization to be done with rowwise enabled and some without then this breaks down. I don't know a lot about the usage though, so that may be reasonable.

The one example I think definitely doesn't fit is the unit test binary, in which we'd want to be able to test both the enabled and disabled options. Ideally your new test `enableRowwiseQuantizedFullyConnected` should fail if enable-rowwise is false (or not working correctly), and some other test should fail if enable-rowwise is true - that way we get the best signal that everything works as expected. If you control it via a global command line option you can't have both those tests in the same binary.

beicy:
@nickgg : Thanks a lot!
 
" and some other test should fail if enable-rowwise is true " 
-- We can view RowwiseQuantizedFullyConnected as an optimization  compared with regular quantized FC. Ideally, for some cases, the accuracy drop should be decreased if we apply rowwise quantization. Therefore, other unittests shouldn't be failed if enable-rowwise is true. 

" Ideally your new test enableRowwiseQuantizedFullyConnected should fail if enable-rowwise is false (or not working correctly),"
-- Yes, I am going to added assert to checks if the conversion is enabled correctly by "-enable-rowwise" :)

lib/Graph/Graph.cpp
```diff
@@ -565,27 +565,42 @@ FullyConnectedNode *Function::createFullyConnected(llvm::StringRef name,
...
   auto *offsets = getParent()->createConstant(ElemKind::Int32ITy, {numRows},
                                               "offsets.rwqfc");
 
+  Tensor wt;
+  if (transposeWeight) {
+    // This happens when the RowwiseQuantizedFullyConnected node is converted
+    // from a quantized FullyConnected node in Glow's quantization procedure.
+    // Since in FC, the weights is stored as transposed (i.e. I * W + B), but in
+    // RowwiseQuantizedFullyConnected, the weights is stored as it is (i.e. I *
+    // W(T) + B).
```
bertmaher: I'm kind of confused about this transposition business.  I assume we're doing this because Caffe2's RowwiseFC gives us transposed weights?  Can we handle transposition in the loader like we do for NCHW convolutions?

beicy: So far, there is no `Caffe2 RowwiseFC op`. This op now is kind of internal and only for quantization.  In this quantization procedure, RowwiseQuantizedFullyConnected in created from FC, which has (I * W + B). But in the design of our RowwiseQuantizedFullyConnected node, it should be (I * W(T) + B) ( here the "rowwise" means the each row of "W").

lib/Quantization/Quantization.cpp
```diff
@@ -369,6 +378,45 @@ class FunctionQuantizer : public FunctionConverter {
...
+  void enableRowwise() {
+    // Traverse all node to find FullyConnected Node, and convert it to
+    // RowwiseQuantizedFullyConnected if the weights is constant.
+    auto nodeIt = function_.getNodes().end();
+    auto stopIt = function_.getNodes().begin();
+    do {
+      --nodeIt;
+      Node &node = *nodeIt;
+      if (auto *Q = llvm::dyn_cast<DequantizeNode>(&node)) {
+        if (auto *fcN = llvm::dyn_cast<FullyConnectedNode>(Q->getInput())) {
```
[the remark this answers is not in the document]
beicy: Added comments 2 lines above.

include/glow/Quantization/Quantization.h
```diff
@@ -72,7 +72,8 @@ Function *
 quantizeFunction(const ExecutionEngine &EE,
                  llvm::ArrayRef<NodeQuantizationInfo> quantizationInfos,
                  Function *F, llvm::StringRef newFuncName = "",
-                 const KindSet &doNotQuantizeKinds = {});
+                 const KindSet &doNotQuantizeKinds = {},
+                 bool enableRowwise = false);
```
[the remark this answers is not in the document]
beicy: Comments updated.

tools/loader/Loader.cpp
```diff
@@ -31,6 +31,13 @@
 
 using namespace glow;
 
+/// -enable-rowwise - Command line option to enable ENABLE_ROWWISE statements.
```
rdzhabarov: can you expand here `Command line option to enable ENABLE_ROWWISE statements`
with more details. `ENABLE_ROWWISE statements` does not tell much.

lib/Quantization/Quantization.cpp
```diff
@@ -369,6 +371,62 @@ class FunctionQuantizer : public FunctionConverter {
     // Use for debug purposes.
     lastMorphedNodeWithTypeChanges = nullptr;
   }
+
+  void enableRowwise() {
```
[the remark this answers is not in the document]
beicy: comments added. But enableRowwise() is a class member, and if it is declared as static, using the members from its base class will have errors.

tools/loader/Loader.cpp
```diff
@@ -31,6 +31,14 @@
 
 using namespace glow;
 
+/// -enable-rowwise : Command line option to enable rowwise quantized
+/// fullyconnected in quantization producure.
+bool enableRowwiseOpt;
+static llvm::cl::opt<bool, true>
```
rdzhabarov: have you run loader with this option succesfuly?

beicy: Yes.  The test is as follows. Also,  in that test, I added a printf to print out "now in rowwise" once a conversion is successful. There is only one `FC` in resnet50, so there is one "now in rowwise" printed out. 
```
wangm-mbp:build_Release1 wangm$ ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -load-profile="profile.yaml" -enable-rowwise
now in rowwise
Model: resnet50
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.6107)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9519)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9877)
```

Issue #2037, opened 2018-11-15
Issue with padding ?


tlepley-cadence:
I've got an ONNX model with standalone Pad operations. I'm currently adding the support in Glow, including an optimization to merge this standalone Pad with Convolution. 

Reading at ONNX and looking at the Glow code, I think that the Glow convolution doesn't correctly support ONNX padding (I didn't look yet at the Caffe2 padding). 

In particular, the Glow code assumes that padding in the model is just given for spatial dimensions (array of 4 numbers), while ONNX specifies that the padding should be given for all dimensions (array of 8 values for a 4D tensor):
> The value represent the number of pixels added to the beginning and end part of the corresponding axis. `pads` format should be as follow [x1_begin, x2_begin...x1_end, x2_end,...], where xi_begin the number of pixels added at the beginning of axis `i` and xi_end, the number of pixels added at the end of axis `i`.

I would like to make sure that we all agree on the interpretation of the ONNX spec, even if we decide for now that we just want to support spatial padding in Glow. For instance, if we want to grow the tensor spatially, the 'pads' attribute in the model would be (considering NCHW) :
[ 0, 0, 1, 1, 0, 0, 1, 1] , and not [1, 1, 1, 1]

Any opinion ?

tlepley-cadence:
I just looked at different models:
- ONNX models from Glow
- ONNX model generated my ft2onnx (one of the converters from the pytorch project)

Actually, while the description of the padding attribute in the ONNX documentation looks similar for the various operators, I get a padding vector of a different size depending on the situation for 4D tensors. There is IMO a problem and I think that we should always get a 8 element vector for 4D tensors.

**Glow zfnet512_onnx_model.txt:** 
- _Convolution_ op:  attribute = name: "pads" ints: 1 ints: 1 ints: 1 ints: 1 type: INTS 

**my_converted_onnx_model.txt:** 
- _Pad_ op: attribute = name: "pads" ints: 0 ints: 0 ints: 1 ints: 1 ints: 0 ints: 0 ints: 1 ints: 1 type: INTS 
- _Maxpool_ op: attribute = name: "pads" ints: 0 ints: 0 ints: 0 ints: 0 type: INTS

nadavrot:
CC @beicy

beicy:
@tlepley-cadence Thanks for posting this issue.

[...]

PR #2103, opened 2018-11-30
[ONNX] Add a partial support for the Pad operator

tlepley-cadence:
*Description*: 

This PR adds a partial support for the Pad operator: 1/ a Pad node, 2/ Support of Pad in the ONNX importer, 3/ Some Pad related optimizations. 
ONNX spec: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Pad

This PR is related to a use-case we got where a set of Pad nodes could **all** be merged into Convolution nodes. Due to the fact that Convolution introduces transpose nodes (conv doesn't follow the ONNX/Caffe2 convention and is NHWC), we had to add an extra transpose/pad sink optim. In our use case, all Pad nodes disappear at optimization, so we decided not to spend (much) more effort supporting Pad at the instruction level and in the various backends. This will need to be implemented at a time a real-use case appears with this need or when a 100% coverage of ONNX will be needed.

Notes:
- we discussed a bit padding with @beicy here: #2037
- that ClangFormat decided to change the formatting of some comments that are not part of the Pad code.

*Testing*: 3 families of tests: 1/ Pad ONNX import, 2/ Pad merge into convolution, 3/ Sink transpose over Pad

*Documentation*:
N/A

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -681,6 +681,59 @@ llvm::Error ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     return llvm::Error::success();
   }
 
+  if (typeName == "Pad") {
+    // Input
+    NodeValue input;
+    ASSIGN_VALUE_OR_RETURN_ERR(input,
+                               getNodeValueOrCreateConstantByName(op.input(0)));
+    auto inputDims = input.dims();
+    auto numDims = inputDims.size();
+
+    // Padding properties.
+    unsigned_t mode = PaddingMode::CONSTANT; // default is constant.
+    if (dict.count("mode")) {
+      std::string modeStr;
+      ASSIGN_VALUE_OR_RETURN_ERR(modeStr, loadStr(dict["mode"]));
+      if (modeStr == "constant") {
+        mode = PaddingMode::CONSTANT;
+      } else if (modeStr == "reflect") {
+        mode = PaddingMode::REFLECT;
+      } else if (modeStr == "edge") {
+        mode = PaddingMode::EDGE;
+      } else {
+        RETURN_ERR("Pad: Invalid mode");
+      }
+    }
+    float value = 0.f; // Default
+    if (dict.count("value")) {
+      ASSIGN_VALUE_OR_RETURN_ERR(value, loadFloat(dict["value"]));
+    }
+
+    // Pads are mandatory.
+    RETURN_ERR_IF_NOT(dict.count("pads"),
+                      "Pad: The 'pads' property is mandatory");
+    auto pads = getShape<int>(dict.at("pads"));
+    RETURN_ERR_IF_NOT(
+        (pads.size() == 2 * numDims),
+        "Pad: the 'pads' array must contain 2 values per dimensions");
+
+    // Compute the output type.
+    std::vector<size_t> outDims(numDims);
+    for (unsigned_t i = 0; i < numDims; i++) {
+      auto new_dim = inputDims[i] + pads[i] + pads[i + numDims];
+      RETURN_ERR_IF_NOT(new_dim > 0,
+                        "The padding can't remove all elements of a dimension");
+      outDims[i] = new_dim;
+    }
+    auto outTy = G_.getParent()->uniqueType(ElemKind::FloatTy, outDims);
+
+    // Create the IR node.
+    Node *N = G_.createPad(opName, input, outTy, mode, pads, value);
+    addNodeAsOutput(op, N);
+
+    return llvm::Error::success();
+  }
+
   RETURN_ERR("Failed to load operator.");
 }
 
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <unordered_map>
 #include <unordered_set>
+#include <vector>
 
 llvm::cl::OptionCategory graphOptCat("Graph Optimizations Options");
 llvm::cl::opt<unsigned> constDedupSizeOpt(
@@ -367,6 +368,47 @@ static bool sinkCode(Function *F) {
       continue;
     }
 
+    // Sink Transpose below Pad nodes.
+    if (auto *padNode = dyn_cast<PadNode>(node)) {
+      auto *transposeNode = dyn_cast<TransposeNode>(padNode->getInput());
+
+      if (!transposeNode) {
+        continue;
+      }
+
+      // The transpose shuffle specifies the source dimension.
+      // When sinking Transpose below Pad, shuffle describes the target
+      // dimension.
+      auto shuffle = transposeNode->getShuffle();
+
+      // Shuffle the Pad output type and the padding attribute.
+      auto outPadType = padNode->getResult().getType();
+      auto outPadShape = outPadType->dims();
+      auto pads = padNode->getPads();
+      size_t numDims = outPadShape.size();
+      std::vector<size_t> newOutPadShape(numDims);
+      std::vector<int> newPads(2 * numDims);
+      for (size_t i = 0; i < outPadShape.size(); i++) {
+        newOutPadShape[shuffle[i]] = outPadShape[i];
+        newPads[shuffle[i]] = pads[i];
+        newPads[shuffle[i] + numDims] = pads[i + numDims];
+      }
+
+      // New pad
+      auto newOutPadType =
+          F->getParent()->uniqueTypeWithNewShape(outPadType, newOutPadShape);
+      auto *NewPadNode = F->createPad(
+          padNode->getName(), transposeNode->getInput(), newOutPadType,
+          padNode->getMode(), newPads, padNode->getValue());
+      NewPadNode->setPredicate(node->getPredicate());
+      auto *newTransposeNode =
+          F->createTranspose(transposeNode->getName(), NewPadNode, shuffle);
+      newTransposeNode->setPredicate(node->getPredicate());
+      padNode->getResult().replaceAllUsesOfWith(newTransposeNode);
+      changed = true;
+      continue;
+    }
+
     // Sink Transpose below Tanh nodes.
     if (auto *TN = dyn_cast<TanhNode>(node)) {
       auto *TR = dyn_cast<TransposeNode>(TN->getInput());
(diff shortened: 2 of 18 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -683,6 +683,58 @@ llvm::Error ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+        RETURN_ERR("Pad: Invalid mode");
+      }
+    }
+    float value = 0.f; // Default
+    if (dict.count("value")) {
+      ASSIGN_VALUE_OR_RETURN_ERR(value, loadFloat(dict["value"]));
+    }
+
+    // Pads are mandatory.
+    assert(dict.count("pads") && "Pad: The 'pads' property is mandatory");
```
jackm321: Thank you for replacing the `assert`s with Errors! Please use `RETURN_ERR_IF_NOT` for this one also.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -683,6 +683,58 @@ llvm::Error ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
...
+    std::vector<size_t> outDims(numDims);
+    for (unsigned_t i = 0; i < numDims; i++) {
+      auto new_dim = inputDims[i] + pads[i] + pads[i + numDims];
+      RETURN_ERR_IF_NOT(new_dim > 0,
+                        "The padding can't remove all elements of a dimension");
+      outDims[i] = new_dim;
+    }
+    auto outTy = G_.getParent()->uniqueType(ElemKind::FloatTy, outDims);
+
+    // Create the IR node
```
nadavrot: Period.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -367,6 +368,47 @@ static bool sinkCode(Function *F) {
...
+    if (auto *padNode = dyn_cast<PadNode>(node)) {
+      auto *transposeNode = dyn_cast<TransposeNode>(padNode->getInput());
+
+      if (!transposeNode) {
+        continue;
+      }
+
+      // The transpose shuffle specifies the source dimension
+      // When sinking Transpose below Pad, shuffle describes the target
+      // dimension
```
nadavrot: Period.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -367,6 +368,47 @@ static bool sinkCode(Function *F) {
...
+      auto shuffle = transposeNode->getShuffle();
+
+      // Shuffle the Pad output type and he padding attribute
```
nadavrot: Period.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -684,6 +726,93 @@ static void mergeMatMul(Function *F) {
...
+    //       - the convolution takes NHWC input tensors.
+    bool canMerge = true;
+    auto pad_pads = PN->getPads();
+    auto cn_pads = CN->getPads();
+
+    // TODO: for now, there is a different interpretation of the ONNX spec for
+    // Pad and convolution the pads array won't have the same size because only
+    // spatial dimensions are specified for the convolution.
+
+    /*
```
nadavrot: Oops, this code is commented out.

nadavrot:
@tlepley-cadence Thank you for doing this work. I have a few questions/ideas/comments. 

1. Can we represent constant pad as a sequence of other primitive nodes (splat-insert)? Will it be more difficult to optimize the low-level IR?

2. Why do we need to add support for signed vectors? Can we have negative pad sizes?

3. Would it be possible to lower the PAD node to a sequence of splat-insert instructions, at least for the CONSTANT mode? 

4. Maybe we should only support the CONSTANT mode in the loader (and node) because we can't handle the other nodes and will just crash later.

tlepley-cadence:
@nadavrot very fast answers:
1, 3 & 4/ I need to think. I'm not too familiar with insert yet.
2/ yes, pads can be negative with the Pad operation

nadavrot:
@tlepley-cadence What happens when the pad is negative? The tensor is sliced? (reduced in size?) 🤕

lib/Optimizer/GraphOptimizer.cpp
tlepley-cadence: Fixed.

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Fixed.

lib/Optimizer/GraphOptimizer.cpp
tlepley-cadence: Fixed.

lib/Optimizer/GraphOptimizer.cpp
tlepley-cadence: Fixed.

lib/Importer/ONNXModelLoader.cpp
tlepley-cadence: Fixed.

tlepley-cadence:
In case of negative 'pads' value, the dimension is shrinked :)
ONNX: "pads: List of integers indicating the number of padding elements to add or remove (if negative) at the beginning and end of each axis".
I've not seen any use-case exploiting such feature. Then, for the time being, we can maybe decide not to support negative pads values too. By supporting CONSTANT with positive pads, I guess we would cover most of the realistic use-cases. On Monday, I'll look more in detail to the possibility to lower Pad (CONSTANT mode and positive pads) with splat&insert.

tlepley-cadence:
I think this PR is good to go. Here is the status:
- In the importer, Pad is fully supported, meaning that the node can carry negative pad and mode information. Nevertheless, we disallow creating a node with a non constant mode because this is not supported at all by Glow currently.
- There are optimizations to merge most common Pad cases with the Convolution node.
- Pad with negative padding are allowed at high level (can for instance potentially be optimized out if merged with a convolution node that has positive pads). Nevertheless, if such node still exists at lowering, it will raise an error because Pad with negative padding is not supported in backends.
- Pads with constant mode and positive padding are now lowered to Splat+Insert and then supported by CPU and Interpreter backends.
- There are multiple tests: ONNX import, optimizations, Pad execution.

lib/Optimizer/Lower.cpp
```diff
@@ -192,6 +191,30 @@ static void lowerReluNode(Function *F, const ReluNode &R) {
   R.getResult().replaceAllUsesOfWith(relu);
 }
 
+static void lowerPadNode(Function *F, const PadNode &P) {
```
nadavrot: Can you please assert that the Pad node is in the right mode (const)?

tests/unittests/graphOptzTest.cpp
```diff
@@ -1061,6 +1061,42 @@ TEST_F(GraphOptz, sinkTransposeBelowConcatWithPredicate) {
   EXPECT_EQ(F_->getNodes().size(), 3);
 }
 
+TEST_F(GraphOptz, sinkTransposeBelowPad) {
+  // Before optims
```
nadavrot: // The shape of the graph before the optimizations.

tests/unittests/graphOptzTest.cpp
```diff
@@ -1061,6 +1061,42 @@ TEST_F(GraphOptz, sinkTransposeBelowConcatWithPredicate) {
...
+  const size_t origDims[] = {1, 5, 10, 15};
+  const size_t outTransposeDims[] = {1, 10, 15, 5};
+  int pads[] = {0, 2, 3, 1, 4, 6, 7, 5};
+  const size_t outPadDims[] = {5, 18, 25, 11};
+
+  // After optims
```
nadavrot: // The shape of the graph after the optimizations.

tests/unittests/onnxImporterTest.cpp
```diff
@@ -720,3 +721,112 @@ TEST(onnx, importSliceNoAxes) {
...
+  auto *node = saveNode->getInput().getNode();
+  auto *padNode = llvm::dyn_cast<PadNode>(node);
+  EXPECT_NE(nullptr, padNode);
+
+  // Check node properties
+  assert(padNode->getMode() == mode);
+  if (mode == PaddingMode::CONSTANT) {
+    EXPECT_EQ(value, padNode->getValue());
+  }
+  // Check the node output shape
```
nadavrot: Period

tools/ClassGen/NodeGen.cpp
```diff
@@ -58,6 +58,14 @@ int main(int argc, char **argv) {
   //                   Convolution / Pool / FC
   //===--------------------------------------------------------------------===//
 
+  BB.newNode("Pad")
+      .addInput("Input")
+      .addMember(MemberType::Unsigned, "Mode")
+      .addMember(MemberType::VectorSigned, "Pads")
+      .addMember(MemberType::Float, "Value")
+      .addResultFromCtorArg()
+      .setDocstring("Performs padding");
```
nadavrot: Please add a more descriptive doc string. What is the semantics of the operator? What happens when negative values are used? What modes are supported? 

Also, can you please use ENUM instead of unsigned for the mode?

nadavrot:
@tlepley-cadence Thanks for doing this work. The approach looks good. I made a few comments in the code. In the future please try to run clang-format before submitting a PR. Most of the changes in this PR are unrelated to your change and it makes it difficult to review the PR.

tlepley-cadence:
@nadavrot My IDE is actually configured to to run clang-format automatically when saving files. The non related changes you have noticed in comments are due to clang-format. So either my clang-format is configured differently for comments, or the source file that I modified was initially not properly formatted with clang-format.

lib/Optimizer/Lower.cpp
tlepley-cadence: Done.

tests/unittests/graphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/graphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/onnxImporterTest.cpp
tlepley-cadence: Done.

tools/ClassGen/NodeGen.cpp
tlepley-cadence: I didn't use enum because `MemberType::Enum`  was not really supported by nodebuilder. I implemented a minimal support of `MemberType::Enum` in nodebuilder, that is still translated as `unsigned_t` currently in the generated API since as far as I know, there is no generic `enum` type in C++. The only thing that we know is:
> It is implementation-defined which integral type is used as the underlying type for an enumeration except that the underlying type shall not be larger than int unless the value of an enumerator cannot fit in an int or unsigned int.

tlepley-cadence: `Pad` doc string: done.

tlepley-cadence:
@nadavrot I addressed all your feedbacks.

PR #2105, opened 2018-11-30
[easy] Fix some lint

bertmaher:
Some issues flagged by FB linter.


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -313,6 +313,8 @@ llvm::Error ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
     }
 
     // Determine the coordinates of the sub-tensor to extract.
+    RETURN_ERR_IF_NOT(axes.size() == starts.size(),
```
tlepley-cadence: There is already a test earlier `axes.size() == starts.size()` in case the `axes` property is given. You may want to remove it since it's redundant.

PR #2106, opened 2018-12-01
[Error handling] Remove last asserts from loaders

jackm321:
*Description*:
This diff removes the last asserts and unhandled errors from the model loading layer. These are the asserts and errors that were in the model loader constructors which makes them uniquely unergonomic to deal with. The general pattern employed here for handling errors that may occur during construction is to have each constructor take an extra parameter that is a pointer to an error and if any errors are raised and if that pointer is not null then assign the error to the error pointer and return early otherwise if an error occurs but the pointer to the error parameter is null then abort. These error pointers get passed down the inheritance chain to each constructor.
This diff also handles any error that may have occurred during model loading in `onnxGetBackendCompatibility()` at the top level and returns a ONNXIFI_STATUS_INTERNAL_ERROR.

*Testing*:
`ninja all`, `ninja check`
ran caffe2 onnx tests with `pytest -s caffe2/python/onnx/test_onnxifi.py` 3/3 tests pass

*Documentation*:
doxygen

### DIFF
--- a/include/glow/Importer/Caffe2ModelLoader.h
+++ b/include/glow/Importer/Caffe2ModelLoader.h
@@ -70,10 +70,13 @@ public:
   /// \p netWeightFilename, and populates the network in \p F.
   /// The list \p types and \p names are used to initialized the inputs and
   /// outputs with specific names and types.
+  /// If \p errPtr is not null then if an error occurs it will get assigned
+  /// there otherwise if an error occurs it will abort.
   Caffe2ModelLoader(const std::string &netDescFilename,
                     const std::string &netWeightFilename,
                     llvm::ArrayRef<const char *> names,
-                    llvm::ArrayRef<TypeRef> types, Function &F);
+                    llvm::ArrayRef<TypeRef> types, Function &F,
+                    llvm::Error *errPtr = nullptr);
 };
 
 } // namespace glow
--- a/include/glow/Importer/ONNXIFIModelLoader.h
+++ b/include/glow/Importer/ONNXIFIModelLoader.h
@@ -27,7 +27,10 @@ namespace glow {
 
 class ONNXIFIModelLoader : public ONNXModelLoader {
 private:
-  ONNXIFIModelLoader(Function &F) : ONNXModelLoader(F) {}
+  /// If \p errPtr is not null then if an error occurs it will get assigned
+  /// there otherwise if an error occurs it will abort.
+  ONNXIFIModelLoader(Function &F, llvm::Error *errPtr = nullptr)
+      : ONNXModelLoader(F, errPtr) {}
 
   /// Load the inputs from the GraphProto. This is useful when the
   /// initializers are not available.
@@ -62,9 +65,10 @@ public:
   /// the \p onnxModel is not supported by the ONNX model parser.
   /// \returns std::vector of Glow operation kind and element kind otherwise.
   ///          It represents a mapping between ONNX nodes and Glow operations.
+  /// \returns llvm::Error if an error was encountered
   ///
   /// \param onnxModel contains a single ONNX operator.
-  static std::vector<std::pair<Kinded::Kind, ElemKind>>
+  static llvm::Expected<std::vector<std::pair<Kinded::Kind, ElemKind>>>
   parseOperators(const void *onnxModel, size_t onnxModelSize);
 };
 
--- a/include/glow/Importer/ONNXIFIModelLoader.h
+++ b/include/glow/Importer/ONNXIFIModelLoader.h
@@ -65,7 +65,7 @@ public:
   /// the \p onnxModel is not supported by the ONNX model parser.
   /// \returns std::vector of Glow operation kind and element kind otherwise.
   ///          It represents a mapping between ONNX nodes and Glow operations.
-  /// \returns llvm::Error if an error was encountered
+  /// \returns llvm::Error if an error was encountered.
   ///
   /// \param onnxModel contains a single ONNX operator.
   static llvm::Expected<std::vector<std::pair<Kinded::Kind, ElemKind>>>
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1088,8 +1088,8 @@ Caffe2ModelLoader::Caffe2ModelLoader(const std::string &netDescFilename,
                                      llvm::ArrayRef<TypeRef> types, Function &F,
                                      llvm::Error *errPtr)
     : CommonOperatorLoader(names, types, F, errPtr) {
-  // lambda to setup the Caffe2ModelLoader and return any llvm::Errors that were
-  // raised
+  // Lambda to setup the Caffe2ModelLoader and return any llvm::Errors that were
+  // raised.
   auto setup = [&]() -> llvm::Error {
     // The caffe2 network descriptor that we are deserializing.
     caffe2::NetDef networkDef;
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1088,6 +1088,11 @@ Caffe2ModelLoader::Caffe2ModelLoader(const std::string &netDescFilename,
                                      llvm::ArrayRef<TypeRef> types, Function &F,
                                      llvm::Error *errPtr)
     : CommonOperatorLoader(names, types, F, errPtr) {
+  // if errPtr already contains an error then don't continue with constructor
+  if (errPtr && *errPtr) {
+    return;
+  }
+
   // Lambda to setup the Caffe2ModelLoader and return any llvm::Errors that were
   // raised.
   auto setup = [&]() -> llvm::Error {
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1085,14 +1085,29 @@ llvm::Error Caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
 Caffe2ModelLoader::Caffe2ModelLoader(const std::string &netDescFilename,
                                      const std::string &netWeightFilename,
                                      llvm::ArrayRef<const char *> names,
-                                     llvm::ArrayRef<TypeRef> types, Function &F)
-    : CommonOperatorLoader(names, types, F) {
-  // The caffe2 network descriptor that we are deserializing.
-  caffe2::NetDef networkDef = EXIT_ON_ERR(loadProtoFile(netDescFilename));
+                                     llvm::ArrayRef<TypeRef> types, Function &F,
+                                     llvm::Error *errPtr)
+    : CommonOperatorLoader(names, types, F, errPtr) {
+  // lambda to setup the Caffe2ModelLoader and return any llvm::Errors that were
+  // raised
+  auto setup = [&]() -> llvm::Error {
+    // The caffe2 network descriptor that we are deserializing.
+    caffe2::NetDef networkDef;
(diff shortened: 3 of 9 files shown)


include/glow/Importer/ONNXIFIModelLoader.h
```diff
@@ -62,9 +65,10 @@ class ONNXIFIModelLoader : public ONNXModelLoader {
   /// the \p onnxModel is not supported by the ONNX model parser.
   /// \returns std::vector of Glow operation kind and element kind otherwise.
   ///          It represents a mapping between ONNX nodes and Glow operations.
+  /// \returns llvm::Error if an error was encountered
```
jackm321: forgot a period

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1078,14 +1078,29 @@ llvm::Error Caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
                                      llvm::ArrayRef<const char *> names,
-                                     llvm::ArrayRef<TypeRef> types, Function &F)
-    : CommonOperatorLoader(names, types, F) {
-  // The caffe2 network descriptor that we are deserializing.
-  caffe2::NetDef networkDef = EXIT_ON_ERR(loadProtoFile(netDescFilename));
+                                     llvm::ArrayRef<TypeRef> types, Function &F,
+                                     llvm::Error *errPtr)
+    : CommonOperatorLoader(names, types, F, errPtr) {
+  // lambda to setup the Caffe2ModelLoader and return any llvm::Errors that were
+  // raised
```
jackm321: period

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -760,19 +760,32 @@ ONNXModelLoader::checkInputs(ONNX_NAMESPACE::GraphProto &net,
...
-                                 llvm::ArrayRef<TypeRef> types, Function &F)
-    : CommonOperatorLoader(tensorNames, types, F) {
-  // The ONNX model that we are deserializing.
-  ONNX_NAMESPACE::ModelProto modelDef =
-      TEMP_EXIT_ON_ERR(loadProto(modelDescFilename));
+                                 llvm::ArrayRef<TypeRef> types, Function &F,
+                                 llvm::Error *errPtr)
+    : CommonOperatorLoader(tensorNames, types, F, errPtr) {
+  // lambda to setup the ONNXModelLoader and return any llvm::Errors that were
+  // raised
```
jackm321: period

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1078,14 +1078,29 @@ llvm::Error Caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
+    // The caffe2 weights that we are deserializing.
+    caffe2::NetDef weightsDef;
+    ASSIGN_VALUE_OR_RETURN_ERR(weightsDef, loadProtoFile(netWeightFilename));
+
+    RETURN_IF_ERR(loadWeights(weightsDef));
+    RETURN_IF_ERR(loadNetwork(networkDef));
 
-  // The caffe2 weights that we are deserializing.
-  caffe2::NetDef weightsDef = EXIT_ON_ERR(loadProtoFile(netWeightFilename));
+    return llvm::Error::success();
```
yinghai: I found `Error::success()` a bit counter-intuitive. Is this standard practice in LLVM?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1078,14 +1078,29 @@ llvm::Error Caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
+  };
 
-  TEMP_EXIT_ON_ERR(loadWeights(weightsDef));
-  TEMP_EXIT_ON_ERR(loadNetwork(networkDef));
+  if (errPtr) {
+    *errPtr = setup();
+  } else {
+    EXIT_ON_ERR(setup());
```
yinghai: Does this mean errors in importing onnx will terminate the whole program? 

https://github.com/pytorch/glow/blob/4014bde87ac1e45aec7a3a039ece1473dceff94a/include/glow/Support/Error.h#L48-L50

lib/Importer/ProtobufLoader.cpp
```diff
@@ -101,16 +100,34 @@ bool ProtobufLoader::hasNodeByName(llvm::StringRef name) const {
...
     : G_(F) {
   // Verify that the version of the library that we linked against is
   // compatible with the version of the headers we compiled against.
   GOOGLE_PROTOBUF_VERIFY_VERSION;
 
-  assert(tensorNames.size() == types.size() && "Invalid initialization list");
-  for (unsigned i = 0; i < tensorNames.size(); i++) {
-    assert(!hasNodeByName(tensorNames[i]) && "Input names have duplicate");
-    TEMP_EXIT_ON_ERR(createAndRegisterPlaceholder(tensorNames[i], types[i]));
+  // Lambda to setup the ProtobufLoader and return any llvm::Errors that were
```
nadavrot: Why do we need the Lambda here?

lib/Importer/Caffe2ModelLoader.cpp
jackm321: Yes this is the llvm standard practice as far as I can tell. It is just a special error value that signals no error has occurred. I also found this somewhat counterintuitive so I originally defined a RETURN_SUCCESS() macro to try to make it a bit more readable but this was removed during a previous code review.

lib/Importer/Caffe2ModelLoader.cpp
jackm321: Before having added this error handling stuff to onnx/c2 importing code any errors in the importer would have caused the whole program to terminate no matter what. Now most methods in the loader return an Error which can be handled and appropriately by the user. In this case, you can't return an Error from a constructor so you can either pass a pointer to an Error and any error that occurs during construction can be placed there or if the user chooses not to do that then importing will exit the program on error like before. This is strictly safer than before because now the user of the importer has the option to handle any Errors without killing the program whereas before the program would get killed on any error that occurred no matter what.

lib/Importer/ProtobufLoader.cpp
jackm321: This is here because we can't return an error from the constructor but we want to be able to isolate the fallible code in the constructor so that we can 1. reuse the same error handling mechanisms the rest of the loader uses such as `RETURN_IF_ERR`, `ASSIGN_VALUE_OR_RETURN_ERR`, etc and 2. not repeatedly check if errPtr==nullptr and handle each errors independently (we can just say here's the fallible logic in this lambda, if errPtr !=nullptr then assign any error that occurs there otherwise if errPtr==nullptr handle any errors by exiting). I chose to use a lambda and not a private method because this code should not get called outside of this constructor and if anyone wants to create another constructor in the future they can follow this pattern without any issues.

lib/Importer/Caffe2ModelLoader.cpp
yinghai: Ok.  I got it. Thanks for explanation. Let's make sure the caller do supply this error pointer. :) Exiting the program on import exceptions is not desirable.

jackm321:
Any changes needed here?

rdzhabarov:
sorry for the delay here, I'll look today.

yinghai:
Now that we cleaned things, does it make something to introduce some negative unittests and see if nothing aborts?

bertmaher:
I'm hesitating to accept because I don't love the pass-an-optional-error pointer for some reason, but I'm uncertain why, other than aesthetics.  Maybe it feels easier to ignore than a wrapped error return type?

What would you think of moving the "work" currently in the constructor to a `load()` method, which can return an error type?  To avoid the clunkiness of `ModelLoader().load("foo.pb"...);` we could make the default constructor private and have a static factory method like `ModelLoader::load("foo.pb"...)`

jackm321:
@yinghai good idea, I can add some negative unit tests, is it ok if I add those in a separate pr?

@bertmaher I thought about doing this (If I understand your suggestions correctly) but I think it has the following issues (maybe there is some nice way around these though):

1. There will have to be multiple `load()` methods if there are multiple constructors that do different things
2. Passing back an `Expected<ModelLoader>` from a factory method still incurs a copy unfortunately and this doesn't play well with the tensors `ModelLoader` owns (they get destroyed so they are still referenced from the model but are invalid so we'd have to do `Expected<std::unique_ptrModelLoader>>` to fix this I think.
3. This won't play well with the constructors that call base class constructors like `ONNXModelLoader::ONNXModelLoader` calls `CommonOperatorLoader::CommonOperatorLoader`, in order to capture any errors that may occur with the `load()` method, these constructors would have to also call the base class's `load()` or something funky like that in order to propagate those errors upwards (the factory methods can't call the base class factory methods).

I was thinking maybe we could have two classes of loaders, ones that are only used as base classes for other loaders not used directly and these can take an Error* in their constructors and then we have other loaders that are used directly and these can have a factory method returning `Expected<std::unique_ptrModelLoader>>` however this is kinda weird and also might not work cleanly because `ONNXModelLoader` is both used directly and inherited from `by ONNXIFIModelLoader`.
Sorry for the long response, let me know what you think.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -65,10 +65,13 @@ class Caffe2ModelLoader
...
   /// The list \p types and \p names are used to initialized the inputs and
   /// outputs with specific names and types.
+  /// If \p errPtr is not null then if an error occurs it will get assigned
+  /// there otherwise if an error occurs it will abort.
   Caffe2ModelLoader(const std::string &netDescFilename,
                     const std::string &netWeightFilename,
                     llvm::ArrayRef<const char *> names,
-                    llvm::ArrayRef<TypeRef> types, Function &F);
+                    llvm::ArrayRef<TypeRef> types, Function &F,
+                    llvm::Error *errPtr = nullptr);
```
rdzhabarov: this is the reason I like exceptions :)

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1078,14 +1078,34 @@ llvm::Error Caffe2ModelLoader::loadWeights(caffe2::NetDef &net) {
...
 
-  TEMP_EXIT_ON_ERR(loadWeights(weightsDef));
-  TEMP_EXIT_ON_ERR(loadNetwork(networkDef));
+    RETURN_IF_ERR(loadWeights(weightsDef));
+    RETURN_IF_ERR(loadNetwork(networkDef));
+
+    return llvm::Error::success();
+  };
+
+  if (errPtr) {
```
rdzhabarov: what's the case when this is not null but you still need to redefine it by *errPtr = setup()?

lib/Importer/ProtobufLoader.cpp
```diff
@@ -101,16 +100,39 @@ bool ProtobufLoader::hasNodeByName(llvm::StringRef name) const {
...
+  if (errPtr && *errPtr) {
+    return;
+  }
+
+  // Lambda to setup the ProtobufLoader and return any llvm::Errors that were
+  // raised.
+  auto setup = [&]() -> llvm::Error {
+    RETURN_ERR_IF_NOT(tensorNames.size() == types.size(),
+                      "Invalid initialization list");
+    for (unsigned i = 0; i < tensorNames.size(); i++) {
```
rdzhabarov: LLVM typically recommend to do this kind of loop
https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

rdzhabarov:
Can you also update fbcode once this is merged.

yinghai:
> @yinghai good idea, I can add some negative unit tests, is it ok if I add those in a separate pr?

Yeah, totally.

PR #2107, opened 2018-12-01
Fix OpenCL memleak and update ExecutionEngine

### LINKED ISSUE #2104: [GLOW] Memory leaking in opencl and build issues
I’m having memory issues running the mnist and cifar10 examples, and I had problems compiling which might be related.

The memory issue is that when I run e.g. the cifar10 training example with -opencl, the GPU memory usage grows at a rate of 200MB per second (looking at nvidia-smi) until the lack of memory crashes the program.

I tried updating my nvidia libraries and opencl libraries but the behaviour is the same.
Any ideas how to debug this problem? The mnist example actually makes it to the final predictions and they are right, so the program is running correctly in that respect.

The building issues I had might be related, firstly for some reason llvm-link-6.0 was found by cmake while the rest finds llvm-7.0, so I had to manually set that to llvm-link-7 (I’m running debian btw.)
Then I had to make sure the code within the FACEBOOK_INTERNAL && LLVM_VERSION_PATCH < 20181009 is run, and not the regular llvm-7 code. It seems that the LLVM_VERSION_PATCH variable is not set while I apparently need it.

Any help would be appreciated! The glow library seems otherwise perfect to use!

gcatron:
*Description*: Removed memleak from OpenCL backend and added logic to executionEngine to only call setupRuns and tearDownRuns once.
*Testing*: ninja test all pass
*Documentation*: NA
Partially fixes #2104

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -41,6 +41,10 @@ public:
   virtual void afterRun(const Context &ctx) = 0;
   /// Final cleanup. Release memory, reset device.
   virtual void tearDownRuns() = 0;
+
+protected:
+  /// Flag to ensure setupRuns is only called once.
+  bool runsSetup_{false};
 };
 } // end namespace glow
 
--- a/lib/Backends/CPU/CPUFunction.cpp
+++ b/lib/Backends/CPU/CPUFunction.cpp
@@ -25,17 +25,23 @@ CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
                          const runtime::RuntimeBundle &runtimeBundle)
     : JIT_(std::move(JIT)), runtimeBundle_(runtimeBundle) {}
 
-CPUFunction::~CPUFunction() { alignedFree(runtimeBundle_.getConstants()); }
+CPUFunction::~CPUFunction() {
+  alignedFree(runtimeBundle_.getConstants());
+  tearDownRuns();
+}
 
 void CPUFunction::setupRuns() {
-  if (runtimeBundle_.getActivationsSize() != 0) {
-    baseActivationsAddress_ = (uint8_t *)alignedAlloc(
-        runtimeBundle_.getActivationsSize(), TensorAlignment);
-  }
+  if (!runsSetup_) {
+    if (runtimeBundle_.getActivationsSize() != 0) {
+      baseActivationsAddress_ = (uint8_t *)alignedAlloc(
+          runtimeBundle_.getActivationsSize(), TensorAlignment);
+    }
 
-  if (runtimeBundle_.getMutableWeightSize() != 0) {
-    baseMutableWeightVarsAddress_ = (uint8_t *)alignedAlloc(
-        runtimeBundle_.getMutableWeightSize(), TensorAlignment);
+    if (runtimeBundle_.getMutableWeightSize() != 0) {
+      baseMutableWeightVarsAddress_ = (uint8_t *)alignedAlloc(
+          runtimeBundle_.getMutableWeightSize(), TensorAlignment);
+    }
+    runsSetup_ = true;
   }
 }
 
@@ -66,11 +72,14 @@ void CPUFunction::afterRun(const Context &ctx) {
 void CPUFunction::tearDownRuns() {
   if (baseMutableWeightVarsAddress_) {
     alignedFree(baseMutableWeightVarsAddress_);
+    baseMutableWeightVarsAddress_ = nullptr;
   }
 
   if (baseActivationsAddress_) {
     alignedFree(baseActivationsAddress_);
+    baseActivationsAddress_ = nullptr;
   }
+  runsSetup_ = false;
 }
 
 void CPUFunction::execute() {
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -82,7 +82,6 @@ void ExecutionEngine::run(Context &ctx) {
   function_->beforeRun(ctx);
   function_->execute();
   function_->afterRun(ctx);
-  function_->tearDownRuns();
 }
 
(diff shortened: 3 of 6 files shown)


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -25,21 +25,35 @@
...
 
 ExecutionEngine::ExecutionEngine(BackendKind backendKind)
-    : backend_(createBackend(backendKind)) {}
+    : backend_(createBackend(backendKind)) {
+  firstRun_ = true;
+}
 
 /// Set the code generator kind to \p backendKind.
 void ExecutionEngine::setBackend(BackendKind backendKind) {
+  if (function_) {
```
opti-mix: You repeat the following piece of code 3 times in your PR:

```c++
  if (function_) {
    function_->tearDownRuns();
  }
```
May be move it into a helper function?

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -42,6 +42,8 @@ class ExecutionEngine final {
   std::unique_ptr<Backend> backend_;
   /// A glow function compiled for this ExecutionEngine's backend.
   std::unique_ptr<CompiledFunction> function_;
+  /// Flag to ensure setupRuns is only called once
```
opti-mix: Period at the end of a comment.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Another thought: May be the destructor of `function_` should always call `tearDownRuns` to ensure that all resources are  always freed when the function is deleted?  It should be possible to call `tearDownRuns` multiple times on the same function, so that both the destructor and the user of the `CompiledFunction` can do it. The subsequent calls of `tearDownRuns` should see that there is nothing to release and simply return.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -42,6 +42,8 @@ class ExecutionEngine final {
...
+  bool firstRun_;
```
yinghai:
```
bool firstRun_{true};
```
And you don't need to do it in the ctor?

lib/ExecutionEngine/ExecutionEngine.cpp
yinghai: Maybe you can put it as customized deleter for the `unique_ptr`?

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -78,11 +92,13 @@ void ExecutionEngine::run(Context &ctx) {
   assert(function_ && "No function has been compiled");
   // Make sure that the context has backing tensors for all placeholders.
   ctx.allocate(M_.getPlaceholders());
-  function_->setupRuns();
+  if (firstRun_) {
+    function_->setupRuns();
```
yinghai: The introduction of `firstRun_` is to allow us to do later setup of the function? What is the benefit of this?

lib/ExecutionEngine/ExecutionEngine.cpp
gcatron: On thinking about this more, I think it makes more sense to add a setup flag to compiledFunction and check that flag when setupRuns is called. Similarly call teardownRuns from the function's dtor. This way all the logic is contained within the function. Also this would make setupRuns and tearDownRuns safe to call multiple times.

opti-mix: The small problem is: you can easily control the `ExecutionEngine`, but `CompiledFunction` is just an interface currently and thus you have to rely on the backends to correctly implement the logic you described above. And each backend would need to implement it correctly. May be some common logic for this could be somehow provided by the `CompiledFunction`, so that backends have less work to do?

Also, this setup flag is about managing runtime resources. And `CompiledFunction` is currently handling both the compilation and runtime aspects, but this logic will probably be separated in the future, or?

gcatron: This is probably a stop gap measure. Once it's implemented I expect the DeviceManager/DagExecutor to handle this sort of logic. Certain parts will flow forward I think. I like the idea of a requirement of the interface is that setupRuns and tearDownRuns must be capable of being called multiple times safely.

include/glow/Backends/CompiledFunction.h
```diff
@@ -41,6 +41,8 @@ class CompiledFunction {
   virtual void afterRun(const Context &ctx) = 0;
   /// Final cleanup. Release memory, reset device.
   virtual void tearDownRuns() = 0;
+  /// Flag to ensure setupRuns is only called once.
+  bool runsSetup_{false};
```
opti-mix: Should it be a private or protected member?

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -25,17 +25,22 @@ CPUFunction::CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
...
+  alignedFree(runtimeBundle_.getConstants());
+  tearDownRuns();
+}
 
 void CPUFunction::setupRuns() {
-  if (runtimeBundle_.getActivationsSize() != 0) {
-    baseActivationsAddress_ = (uint8_t *)alignedAlloc(
-        runtimeBundle_.getActivationsSize(), TensorAlignment);
-  }
+  if (!runsSetup_) {
```
opti-mix: Where do you set `setupRuns_` to true? Should it be at the end of this block?

yinghai:
I'm curious under what circumstances do we call `setupRun` and `tearDownRun` multiple times?

gcatron:
> I'm curious under what circumstances do we call `setupRun` and `tearDownRun` multiple times?

SetupRuns I'm not sure, tearDown I can see where we might want to free memory but keep the function around longer it would get called again in the dtor.

PR #2112, opened 2018-12-02
[ONNXIFI] Create a class to manage all ONNXIFI interface resources

### LINKED ISSUE #2026: [ONNXIFI] onnxifi interface implementation can't identify invalid IDs

jackm321:
*Description*:
Create a basic singleton class `GlowOnnxManager` to create and manage ONNX interface objects (`Graph`s, `Event`s, `Backend`s, etc) created by glow. `GlowOnnxManager` can be used to create or register objects, check if a given object was created by glow, and free a given object.
*Testing*:
`ninja check`
ran C2 onnxifi integration test with `caffe2/python/onnx/test_onnxifi.py` 3/3 pass
*Documentation*:
doxygen
fixes #2026


lib/Onnxifi/GlowOnnxManager.h
```diff
@@ -0,0 +1,96 @@
...
+namespace glow {
+namespace onnxifi {
+class BackendId;
+class Backend;
+class Event;
+class Graph;
+
+/// Singleton class for creating and destroying objects for the ONNX interface.
+/// GlowOnnxManager tracks objects it has created and can be used to check if
+/// if an ONNX interface object was created by glow or not.
```
rdzhabarov: typo "if if"

lib/Onnxifi/GlowOnnxManager.h
```diff
@@ -0,0 +1,96 @@
...
+#include <unordered_set>
+
+namespace glow {
+namespace onnxifi {
+class BackendId;
+class Backend;
+class Event;
+class Graph;
+
+/// Singleton class for creating and destroying objects for the ONNX interface.
```
rdzhabarov: s/ONNX/ONNXIFI

lib/Onnxifi/GlowOnnxManager.h
```diff
@@ -0,0 +1,96 @@
...
+/// if an ONNX interface object was created by glow or not.
+class GlowOnnxManager {
+public:
+  /// Get a reference to the GlowOnnxManager singleton. There should only ever
+  /// be one GlowOnnxManager.
+  static GlowOnnxManager &get();
+
+  // Disallow copying GlowOnnxManager to help enforce singleton pattern.
+  GlowOnnxManager(const GlowOnnxManager &) = delete;
+  GlowOnnxManager &operator=(const GlowOnnxManager &) = delete;
```
rdzhabarov: what about move copy ctor?
also, if we do not intend to inherit from the GlowOnnxManager we should make that `final` class.

lib/Onnxifi/GlowOnnxManager.h
```diff
@@ -0,0 +1,96 @@
...
+namespace onnxifi {
+class BackendId;
+class Backend;
+class Event;
+class Graph;
+
+/// Singleton class for creating and destroying objects for the ONNX interface.
+/// GlowOnnxManager tracks objects it has created and can be used to check if
+/// if an ONNX interface object was created by glow or not.
+class GlowOnnxManager {
```
rdzhabarov: I'd like to see a unit test for this class

lib/Onnxifi/GlowOnnxManager.h
```diff
@@ -0,0 +1,96 @@
...
+  /// Get a reference to the GlowOnnxManager singleton. There should only ever
+  /// be one GlowOnnxManager.
+  static GlowOnnxManager &get();
+
+  // Disallow copying GlowOnnxManager to help enforce singleton pattern.
+  GlowOnnxManager(const GlowOnnxManager &) = delete;
+  GlowOnnxManager &operator=(const GlowOnnxManager &) = delete;
+
+  /// Add a new glow \p backendId to the set of valid backendIds.
+  void addBackendId(BackendId *backendId);
```
rdzhabarov: we have typedefs for all those pointers, e.g., BackendIdPtr

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -60,13 +63,20 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
...
+  auto *cpuBackend =
+      new glow::onnxifi::BackendId(glow::BackendKind::CPU, /*id*/ 1,
+                                   /*concurrency*/ 1);
+  auto *interpreterBackend =
+      new glow::onnxifi::BackendId(glow::BackendKind::Interpreter,
+                                   /*id*/ 2, /*concurrency*/ 1);
+  manager.addBackendId(cpuBackend);
+  manager.addBackendId(interpreterBackend);
+
   // TODO: change concurrency level to std::thread::hardware_concurrency()
```
rdzhabarov: could you move this comment appropriately as well

lib/Onnxifi/GlowOnnxManager.cpp
```diff
@@ -0,0 +1,97 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "GlowOnnxManager.h"
+
+#include "Base.h"
+
+namespace glow {
+namespace onnxifi {
+// static
```
rdzhabarov: not sure if comment is useful :)

lib/Onnxifi/GlowOnnxManager.cpp
```diff
@@ -0,0 +1,97 @@
...
+  assert(!isValid(backendId));
+  backendIds_.insert(backendId);
+  assert(isValid(backendId));
+}
+
+BackendPtr GlowOnnxManager::createBackend(BackendIdPtr backendId) {
+  assert(isValid(backendId));
+  BackendPtr backend = new Backend(backendId);
+  backends_.insert(backend);
+  assert(isValid(backend));
```
yinghai: Here you can do
```
auto it = backends_.emplace(new Backend(BackendId))
assert(it.first);
return it.second;
```
Might be more natural.

yinghai:
I agree with @rdzhabarov that we need some unit tests for Glow/ONNXIFI runtime, especially multithreaded tests.

lib/Onnxifi/GlowOnnxManager.h
jackm321: Yeah I was trying to just forward declare `BackendId` etc and not import Base.h unnecessarily here but I would have to re typedef those (not sure if that' even legit) or import Base.h just for those typedefs. I'll just import it.

lib/Onnxifi/GlowOnnxManager.cpp
jackm321: Ok, I find these useful when reading code but if it's not in the style glow uses I'll remove it.

jackm321:
@rdzhabarov thank you for the comments :)

@yinghai thank you also for the comments and to address you points:

1. Yes good point, I was going to ask if anyone knows if the concurrency requirements of the ONNXIFI interface are specified anywhere? Can anything happen concurrently with anything else (within reason) or are somethings guaranteed to be single threaded etc. We can just put a single lock around everything for maximum safety and simplicity (or a lock per set but I'll need to check that we can't get into any kind of circular wait pattern with this) but I'd like to understand more about what the requirements are to see how we can optimize this.
2. What standards do we have to know if performance is sufficiently good or not? What can we measure to know if something (a particular locking scheme in this case) is too slow on the ONNXIFI interface?
3. I'm not sure I understand could you elaborate on this one?

jackm321:
@yinghai I added a lock over the whole thing, it's not ideal cause it's a global lock and get's locked multiple times in succession sometimes (because of the internal calls to `isValid()`) but I don't think it should cause any perf issues for now. We can always optimize it later if need be.

yinghai:
@jackm321 Yeah, definitely. As I mentioned, 2 and 3 and pluses and it makes sense to handle them separately. 

On the other hand, it would good to add a multithread test case.

lib/Onnxifi/GlowOnnxManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+  {
+    std::lock_guard<std::mutex> lock(m_);
+    graphs_.insert(graph);
+  }
+  assert(isValid(graph));
+  return graph;
+}
+
+bool GlowOnnxManager::isValid(BackendIdPtr backendId) {
+  std::lock_guard<std::mutex> lock(m_);
```
yinghai: This won't work. You will have a recursive lock?

jackm321: @yinghai why won't this work?

yinghai: OK. This works, as you don't have recursive locks. :) But this kind of things is a bit error prone...

lib/Onnxifi/GlowOnnxManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+GraphPtr GlowOnnxManager::createGraph(BackendPtr backend) {
+  assert(isValid(backend));
+  GraphPtr graph = new glow::onnxifi::Graph(backend);
+  {
+    std::lock_guard<std::mutex> lock(m_);
+    graphs_.insert(graph);
+  }
+  assert(isValid(graph));
```
yinghai: Like I pointed out above, this check is not necessary. The return value of `insert` should already tell you whether the graph is valid or not.

lib/Onnxifi/GlowOnnxManager.cpp
jackm321: Imo with `lock_guard` it's reasonably straightforward to see the scope of the locking and avoid errors. What would you view as being less prone to errors here?

lib/Onnxifi/GlowOnnxManager.cpp
jackm321: Yeah it's true, on the one hand I like how this encapsulates the logic around checking validity in one place but given that there's duplicated locking involved here I think it's worth it do the thing you're suggesting.

lib/Onnxifi/GlowOnnxManager.cpp
yinghai: When you have a function that grabs a lock and calls a function that grabs the same lock, it's kind of tricky to to explicitly manage the lock_guard to avoid recursive lock, like what you did here. It's also like three lock/unlock for a function. Ideally, one would be nice. BTW, in opt mode, none of the checks will generate actual code and we essentially will have only one lock, right?

jackm321:
@rdzhabarov this is version with the cmake issues

lib/Onnxifi/GlowOnnxManager.h
jackm321: @rdzhabarov declaring the copy constructor as deleted counts as declaring it I believe so the condition "there are no user-declared copy constructors" is false. This example seems to confirm this https://ideone.com/Clhvy5

lib/Onnxifi/GlowOnnxManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+namespace glow {
+namespace onnxifi {
+GlowOnnxManager &GlowOnnxManager::get() {
+  static GlowOnnxManager manager;
+  return manager;
+}
+
+void GlowOnnxManager::addBackendId(BackendIdPtr backendId) {
+  assert(!isValid(backendId) && backendId != nullptr);
```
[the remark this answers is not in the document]
jackm321: `isValid` already does this but in this case, we want to make sure that the reason why backendId is invalid is not because it's null but because it's actually not been added already so the extra check is needed (or this could be written without `isValid` which would probably be better actually)

lib/Onnxifi/GlowOnnxManager.h
jackm321: I'm focusing on the wrong details here, you're right that it's better to have move assignment and move construction explicitly deleted, I marked them deleted.

lib/Onnxifi/GlowOnnxManager.h
bertmaher: Orthogonal to the PR, but what's the rationale for those typedefs?  We don't generally do that for pointer types and I generally feel `XyzPtr` is less understandable than `Xyz*`.

lib/Onnxifi/GlowOnnxifiManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+  assert(isValid(backend) && "trying to release an invalid Backend");
+  {
+    std::lock_guard<std::mutex> lock(m_);
+    backends_.erase(backend);
+  }
+  delete backend;
+}
+
+void GlowOnnxifiManager::release(EventPtr event) {
+  assert(isValid(event) && "trying to release an invalid Event");
```
bertmaher: Are we concerned about double frees racing?  If so, don't you need to check `isValid` and do the `erase` atomically?  Otherwise two threads could see that the object is valid and thus both decide to erase it.

jackm321: @bertmaher yeah I'm not sure to be honest, that seems like an odd case to have multiple threads racing to free the same object (there is something bigger than concurrency wrong with that picture) but I could just check into the lock to make sure.

jackm321: fixed

jackm321:
Any further changes needed?

lib/Onnxifi/GlowOnnxifiManager.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+namespace glow {
+namespace onnxifi {
+GlowOnnxifiManager &GlowOnnxifiManager::get() {
+  static GlowOnnxifiManager manager;
+  return manager;
+}
+
+void GlowOnnxifiManager::addBackendId(BackendIdPtr backendId) {
+  {
```
rdzhabarov: no need for scoping

lib/Onnxifi/GlowOnnxifiManager.h
```diff
@@ -0,0 +1,115 @@
...
+
+  /// The set of all valid glow Events.
+  std::unordered_set<EventPtr> events_;
+
+  /// The set of all valid glow Graphs.
+  std::unordered_set<GraphPtr> graphs_;
+
+  /// Mutex that protects all members of GlowOnnxifiManager.
+  /// TODO: can use one mutex per set if performance becomes an issue.
+  mutable std::mutex m_;
```
rdzhabarov: why do you need this to be mutable?

tests/unittests/CMakeLists.txt
```diff
@@ -312,13 +322,26 @@ add_executable(onnxifiThreadPoolTest
 target_link_libraries(onnxifiThreadPoolTest
                       PRIVATE
                         onnxifi-glow-thread-pool
+                        Importer
```
rdzhabarov: why importer in glow thread pool test, seems unrelated

tests/unittests/CMakeLists.txt
```diff
@@ -312,13 +322,26 @@ add_executable(onnxifiThreadPoolTest
...
                         gtest
                         testMain)
 target_include_directories(onnxifiThreadPoolTest PRIVATE ${CMAKE_SOURCE_DIR}/lib/Onnxifi)
 add_glow_test(NAME onnxifiThreadPoolTest
               COMMAND ${GLOW_BINARY_DIR}/tests/onnxifiThreadPoolTest --gtest_output=xml:onnxifiThreadPoolTest.xml
               WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 
+add_executable(GlowOnnxifiManagerTest
```
rdzhabarov: can you fix the layout

lib/Onnxifi/GlowOnnxifiManager.h
jackm321: `mutable` for mutexes is usually good (I think) so that you can take a const ref to an object and still call methods that need to lock the mutex

tests/unittests/GlowOnnxifiManagerTest.cpp
```diff
@@ -0,0 +1,140 @@
+/**
```
rdzhabarov: remove ` tests/unittests/GlowOnnxManagerTest.cpp` file

tests/unittests/GlowOnnxifiManagerTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+  manager.release(event);
+  manager.release(graph);
+}
+
+TEST(GlowOnnxifiManagerTest, Concurrency) {
+  auto numThreads = 24;
+  auto numItersPerThread = 1000;
+
+  std::vector<std::thread> threads;
+  for (auto i = 0; i < numThreads; ++i) {
```
rdzhabarov: can you stick to the same style, e.g., `int` for loops here

lib/Onnxifi/GlowOnnxifiManager.h
rdzhabarov: we do not have any const methods in this class :)

jackm321: good point, valid() should be const, changed it

tests/unittests/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
+if(NOT TARGET onnx_proto)
```
rdzhabarov: we do not need this i think. the only thing we use is the onnxifi.h file.

jackm321:
@rdzhabarov yes I'll update fbcode, thank you for reviewing this!

PR #2120, opened 2018-12-05
[ci] Test with shared libraries

bertmaher:
*Description*: Since we've gone to the trouble of making shared libraries work let's make sure it stays that way.

A couple alternatives:
- Switch CI to shared (but most people are using static, so this seems bad)
- Switch our default to shared in root CMakeLists.  This could be OK.

I liked just adding test coverage so we can be sure all configs work, but I'm open to discussion.

*Testing*: new CI config
*Documentation*: n/a

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -57,32 +57,18 @@ cd ${GLOW_DIR}
 mkdir build && cd build
 CMAKE_ARGS=("-DCMAKE_CXX_FLAGS=-Werror")
 CMAKE_ARGS+=("-DGLOW_WITH_CPU=ON")
-
-case "${CIRCLE_JOB}" in
-    ASAN)
-        CMAKE_ARGS+=("-DGLOW_USE_SANITIZER='Address;Undefined'")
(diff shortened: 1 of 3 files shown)


.circleci/build.sh
```diff
@@ -57,17 +57,33 @@ cd ${GLOW_DIR}
...
+    ASAN)
+        CMAKE_ARGS+=("-DGLOW_USE_SANITIZER='Address;Undefined'")
+        CMAKE_ARGS+=("-DGLOW_WITH_OPENCL=OFF")
+        CMAKE_ARGS+=("-DCMAKE_BUILD_TYPE=Release")
+        ;;
+
+    DEBUG)
+        install_pocl
+        CMAKE_ARGS+=("-DCMAKE_BUILD_TYPE=Debug")
+        CMAKE_ARGS+=("-DGLOW_WITH_OPENCL=ON")
```
rdzhabarov: for brevity we could add to default options right after line 59 and only turn off for ASAN

bertmaher: done!

Issue #2115, opened 2018-12-04
Manage tensors in model loaders as unique_ptrs instead of raw pointers


jackm321:
Now that the model loaders have error handling, memory that they allocate must be handled with care and cleaned up when an error occurs or there could be a leak (there was already one). To make managing the memory allocated for tensors more robust in the face of errors, we should use `llvm::StringMap<std::unique_ptr<Tensor>> tensors_` instead of `llvm::StringMap<Tensor *> tensors_` in `ProtobufLoader`. Any loader inheriting from `ProtobufLoader` should allocate all tensors as unique_ptrs and move them to `tensors_` to avoid leaks.

PR #2124, opened 2018-12-05
[onnx loader] Replace raw pointers with unique ptr.

rdzhabarov:
*Description*:
Ensure there is no memory leaks associated with allocated tensors in various loaders
*Testing*:
Existing unit tests
*Documentation*:
N/A

Fixes https://github.com/pytorch/glow/issues/2115

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -845,14 +845,13 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
      floats: -0.028315347
      */
 
-    auto *T = new Tensor();
     for (auto &o : op.output()) {
-      tensors_[o] = T;
-    }
+      auto *T = new Tensor();
+      tensors_[o].reset(T);
 
-    auto dim = getShape(dict["shape"]);
+      auto dim = getShape(dict["shape"]);
 
-    size_t i = 0;
+      size_t i = 0;
 #define LOAD_TENSOR_FILL(TYPE_NAME, NATIVE_TYPE, PROTO_TYPE_NAME)              \
   T->reset(ElemKind::TYPE_NAME, dim);                                          \
   auto TH = T->getHandle<NATIVE_TYPE>();                                       \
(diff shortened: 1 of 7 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -845,46 +845,47 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
      floats: -0.028315347
      */
 
-    auto *T = new Tensor();
     for (auto &o : op.output()) {
-      tensors_[o] = T;
-    }
+      auto *T = new Tensor();
```
beicy: Not really related to this PR --- Just wondering why we use a loop for op.output() ?  There should be only 1 output for these ops.

rdzhabarov: the previous behavior was not correct as it looped through all outputs and only used the last one.
current code should handle both single/multiple outputs correctly if any.

bertmaher:
PR looks good, just a question: when we load the model, do we copy the weight tensors in to the module?  Or do they have the same lifetime as this loader object?

rdzhabarov:
> Or do they have the same lifetime as this loader object?

They will not. Load manager can be destroyed but loaded function `F` should be fine.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -915,47 +916,48 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
      i: 127
      }
      */
-    auto *T = new Tensor();
     for (auto &o : op.output()) {
+      auto *T = new Tensor();
```
jackm321: Part of the point of making these into unique pointers would be so that if we return early due to an error, the tensor won't be leaked if it wasn't first put into `tensors_`. To that end, I think it would be better if we used make_unique() instead of keeping raw pointers then putting these into unique_ptrs with `reset`

rdzhabarov: std::make_unique might not be available in all compilers we use.

rdzhabarov: but I see your point, I'll send PR

rdzhabarov: https://github.com/pytorch/glow/pull/2136

PR #2126, opened 2018-12-05
[fix] Fix build break

rdzhabarov:
*Description*:
Fixing build break. It was introduced in https://github.com/pytorch/glow/pull/2103 after merging with the latest master.

*Testing*:
CI

*Documentation*:
n/a


artemrakhov:
Proper fix would be to create a separate function `loadPad()`, right? Like what happened to all other ops?

jackm321:
@rdzhabarov would you mind either making a `loadPad()` function as @artemrakhov suggests or we can use #2127 instead

rdzhabarov:
closing in favor of https://github.com/pytorch/glow/pull/2127

PR #2130, opened 2018-12-06
[WIP]Low level IR loader

### LINKED ISSUE #2098: How to load low-level IR from a text file
LLVM provides a method `ParseIRFile` to load LLVM bitcode from a file.
If I write a text file of Glow low-level IR, how can I load it to Glow Compiler? This feature is very useful for debuging if I want to add some new instructions. I can try and verify it quickly.
Is there any suggestion? Thanks!

QiJune:
*Description*: 

This is just a trial implementation. I am not sure if it's the right direction. There are more works to do to finish this feature:

- return a `Instruction` in `parserInstruction` method
- InstructionParser related class can be auto-generated, such as `ConvolutionParser `, `AllocParser` 
- expose IRBuilder, and pass a low-level IR text file to the IRBuilder, IRBuilder will create corresponding InstructionParser to create instructions.

Before going on the work, I need some suggestions to avoid going wrong direction. 
Thanks!

*Testing*:
*Documentation*:
[Optional Fixes #issue]

Try to fix #2098


qcolombet:
cc: @opti-mix

cmake/modules/GlowDefaults.cmake
```diff
@@ -29,7 +29,7 @@ if(MSVC OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
   set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELEASE} /fp:fast")
   set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_RELEASE} /fp:fast")
 else()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wnon-virtual-dtor -fno-exceptions -fno-rtti")
```
jackm321: Do we really want to enable these?

bertmaher: Good catch @jackm321, we definitely don't want to enable exceptions without discussion on its own PR.

opti-mix:
@QiJune The overall direction seems reasonable.

Having dedicated parsers for each instruction e.g. `AllocParser` is probably an overkill, as a grammar for instructions is pretty regular, the only big difference is the final call of a constructor (or createXYZ method) to build the actual instruction.

> InstructionParser related class can be auto-generated,

Auto-generation sounds like a nice idea!

> expose IRBuilder, and pass a low-level IR text file to the IRBuilder, IRBuilder will create corresponding InstructionParser to create instructions.

I'm not sure that the utility function for reading the IR text file should be part of the IRBuilder. IMHO, it should be a separate helper class or function.

opti-mix:
@QiJune Having said that, I'd like to repeat that parsing is not the most difficult part here. 
At the moment, we have not formalized the textual serialization format for Glow IR (or Glow Graph IR). So, before we can parse it, we should agree on a formal grammar for it, make sure that we produce the actual textual IR representation according to this grammar and only then we can start parsing it.

Also, we should take the maintenance cost into account. While I agree that having the ability to parse a textual IR is nice, there was not burning need for it so far. If adding this feature would result e.g. in adding thousands lines of code and adding a number of 3rd-party library dependencies, then we should carefully evaluate if it is worth it.

Please see previous discussions on this topic to get more context and to see potential issues related to this.

bertmaher:
Hi @QiJune, thanks for trying this direction out, but I'm kind of against adding a textual IR parser.  It's a fair amount of complexity to add to the project, and the value added seems a bit dubious, since the C++ API can already be used to construct test IR.  (I know that LLVM provides this feature, but the case there is stronger, since bitcode can be used for product purposes, like cross-module compilation, multiple language front-ends, etc.)

I'm going to go ahead and close this PR.  If anyone thinks we should pursue this direction, though, please feel free to reopen and make your case :-)

PR #2132, opened 2018-12-06
Add a Testing library with a StrCheck class.

stoklund:
*Description*:
The Testing library holds utility classes that are only used for unit testing.

The StrCheck class provides similar functionality to LLVM's FileCheck utility, but in a unit test context.

*Testing*:
`ninja check`

*Documentation*:
In the StrCheck.h header file.


tlepley-cadence:
If it's only for testing, shouldn't it be put in the `tests` directory (next to gtest) rather than the `lib` directory that contains libs of the compiler ?

Issue #2045, opened 2018-11-17
[Glow Runtime] Top Level Task


qcolombet:
This is the top level issue to track all the work we plan to do to make the glow runtime supports concurrent execution, pipelining, batching and so on.

At a high level, the idea for the runtime is to be able to:
- Enqueue inputs: Run input0, then run input1 as soon as the previous run is done, etc.
- Slice the inputs into batch size and transparently run them: Take N input and sequentially run them in batches of M (where M is the size of the compiled model and N the actual run size.)
- Pipeline work across models: Run input1 on model M1, then run the result of M1 on M2 while running input2 on M1, etc.

Among other things, the glow runtime will have to:
- Manage input/output queues for each model (and communication with the devices)
- Manage incoming model
- Keep track of data dependencies and schedule next tasks to be done
- Split inputs
- Pad inputs
- Dispatch workload on device
- Keep track of the status of devices
Also, somewhat orthogonal to the runtime, but related, glow will need to:
- Determine what and where to run things (graph partitioning)

Right now, we started by splitting the compilation and runtime stages properly.
This work is tracked in:
#2040, #1967, #1953, #1951

gcatron:
Adding #2125 to the list. Work is being done on the HostManager. This is part of the new runtime design. The design has five major components: HostManager, Partitioner, Provisioner, DeviceManager, and Executor.
Partitioner:
This component is responsible for breaking up the provided network into subnetworks that can be run on multiple devices. It does its partitioning based on hardware constraints and heuristics to optimize execution time. It outputs a DAG to be used by the other components.
DeviceManager:
The DeviceManager handles interactions with the device. The manager handles initializing the device, compiling a network for the device and preparing the device for execution. It also handles unloading networks from the device.
Provisioner:
The Provisioner takes the output from the partitioner and assigns sub networks to specific devices updating the DAG with device assignments. The Provisioner calls into the device manager to compile and load the subnetworks to the device.
Executor:
The Executor handles the execution of the network. It walks the DAG calling execution of each sub network in accordance with their dependencies. 
HostManager:
The HostManager is the container for the other components. It serves as the interface externally, handling network init and run requests. The HostManager routes a request through the other components and holds the DAGs for each network.

PR #2133, opened 2018-12-06
[WIP][DO NOT ACCEPT][Runtime] Add runtime Executor class

SplitInfinity:
**Description**
graphs using a thread pool whose workers usher every received request to
completion through a series of state transitions.

Testing:
It compiles.

**Documentation**
The code has comments.


include/glow/Runtime/Executor/Executor.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include "glow/Graph/Context.h"
+#include "glow/Graph/Graph.h"
+
+#include <map>
+#include <set>
+#include <string>
+
+namespace glow {
+namespace runtime {
```
nickgg: Considered adding this also and think it's a good idea.

gcatron: I like the clear divide it makes. I think it would be the only sub namespace in the codebase. @opti-mix we've discussed namespaces in the past, what are your thoughts?

opti-mix: I think using separate namespaces for `runtime` and the like is fine.

opti-mix:
@SplitInfinity Question about `Functions`. The current APIs use `Functions`, which means that they are also responsible for compiling them. But may be they should actually use `CompiledFunctions` and let the compilation be controlled by the client of this API? WDYT?

I'm thinking e.g. about the use-case, where we have a bunch of pre-compiled functions and run some of them based on incoming requests.

opti-mix:
@SplitInfinity A couple more questions:

- What about cancellation of a running DAG? Should we support this scenario? If so, what would be the API? IMHO, the `run` method (and may be any other async method) should return a token associated with the action performed by such a method. This would allow us to refer to this action later, which can be very useful for different reasons.

- Your APIs seem to be callback-based. I'm not 100% convinced that it is a right approach. For example, we may run multiple DAGs at the same time and/or we may need to wait for completion (or any other events) from multiple DAGs/actions. This is rather difficult to achieve with callbacks-based APIs, but it probably much simpler using something like tokens described above and an API like `epoll` where you wait for events on a bunch of tokens...

opti-mix:
@SplitInfinity Do we have a design document with some of the envisioned typical use-cases and workflows described? I think it would be very useful for this kind of APIs. Otherwise it is too easy to miss/forget some important use-cases.

SplitInfinity:
@opti-mix 

Thanks for the quick feedback!

> Question about Functions. The current APIs use `Functions`, which means that they are also responsible for compiling them. But may be they should actually use `CompiledFunctions` and let the compilation be controlled by the client of this API? WDYT?

I used `Function` because I did not understand exactly the role of the `Executor` in the runtime. After discussing more with the team today, I realized that this is only used in the execution path and not in the compilation path, so using `CompiledFunction` instead makes sense.

> What about cancellation of a running DAG? Should we support this scenario? If so, what would be the API? IMHO, the run method (and may be any other async method) should return a token associated with the action performed by such a method. This would allow us to refer to this action later, which can be very useful for different reasons.

> Your APIs seem to be callback-based. I'm not 100% convinced that it is a right approach. For example, we may run multiple DAGs at the same time and/or we may need to wait for completion (or any other events) from multiple DAGs/actions. This is rather difficult to achieve with callbacks-based APIs, but it probably much simpler using something like tokens described above and an API like epoll where you wait for events on a bunch of tokens...

@nickgg and I talked about using `std::future` instead of passing around callbacks, but decided against it because the standard library futures are not as powerful as something like `folly::Future`. 

Maybe it is a good idea to make a design document instead of jumping right into coding.

nickgg:
@opti-mix @SplitInfinity - Both Function and CompiledFunction are needed in the DeviceManager (for compile and execute respectively), and if at all possible I think we should not share them with the Executor. Any object sharing in this context will require solving thread safety in the Graph.

The interface in #2134 uses identifiers for Module and Function (`DeviceNetworkID`and function name) which are used to reference the particular sub-graph-function to execute.

Personal opinion is that callbacks are fine here, at least for the prototype. We should make sure were getting a benefit before jumping in on something more complicated (and maybe not cross platform?).

jackm321:
@SplitInfinity @nickgg There seems to be a some commonality between this and #2134 in structure. Both have implemented a basic work queue structure but with some specific details like the dag structure in this case. Are these implementation specific details so great that we aren't able to use a generic executor in both cases and avoid all the extra code for synchronization and stuff? Can we just reenqueue the next piece of work here and can #2134's CPUDeviceManager be implemented with just a generic executor with 1 thread?

SplitInfinity:
Good point @jackm321 

I think we should move the ONNXIFI thread pool class into somewhere common and use that everywhere.

opti-mix:
@nickgg 

> Personal opinion is that callbacks are fine here, at least for the prototype. We should make sure were getting a benefit before jumping in on something more complicated (and maybe not cross platform?).

Callbacks should be fine for a prototype as long as most/all important use-cases can be expressed in their terms. I provided some exampled where the use of call-backs could be rather difficult. But it may very well be that we don't need to handle such use-cases. That's why I asked about a doc describing the high-level design, responsibilities of the entity (be it Executor or DeviceManager) and envisioned important use-cases.

> Both Function and CompiledFunction are needed in the DeviceManager (for compile and execute respectively)

What are the exact responsibilities of DeviceManager? Is it described anywhere, e.g. at least in the high-level class comments? Why does it need to control compilation?

nickgg:
@opti-mix - there's some design notes n the top level task #2045 and @gcatron can point you at the design doc. Can you give me an example of something specific that cannot be done, or is difficult to achieve with the callbacks pattern?

The DeviceManager piece is the interface to the physical (or in the case of CPU backends, logical) device. Specifically it holds the backend, which does the compilation, and so does the compilation.

opti-mix:
@nickgg 

> @opti-mix - there's some design notes n the top level task #2045 and @gcatron can point you at the design doc. 

Thanks for the links! Appreciated.

> Can you give me an example of something specific that cannot be done, or is difficult to achieve with the callbacks pattern?

For example, waiting for a set of multiple events to happen in a specific or any order (e.g. one of the actions/all actions are completed). Events may stem from the execution of different NN models running on the device. Or, e.g. you transfer data to the device using DMA (asynchronously) and you need to check that all/some of transfers have completed before you can proceed. Generally, IMHO, events (because they are data) can be composed better than callbacks (which is code).

BTW, you may want to have a look at e.g. OpenCL for an example of APIs which are very much event-based, rather than callback-based.

> The DeviceManager piece is the interface to the physical (or in the case of CPU backends, logical) device. Specifically it holds the backend, which does the compilation, and so does the compilation.

Oh, I see. The idea that DeviceManager holds the backend is a possible solution, but it could be argued that the backend could also live outside. And whatever it produces (e.g. CompiledFunctions) could be then run on different DeviceManagers "compatible" with this backend (e.g. you have multiple devices of the same kind available). It would be interesting to see the pros/cons of different approaches considered. But this is a longer discussion.

yinghai:
> Your APIs seem to be callback-based. I'm not 100% convinced that it is a right approach. For example, we may run multiple DAGs at the same time and/or we may need to wait for completion (or any other events) from multiple DAGs/actions. This is rather difficult to achieve with callbacks-based APIs, but it probably much simpler using something like tokens described above and an API like epoll where you wait for events on a bunch of tokens...

I agree with @opti-mix. It's really just IO here instead of actual CPU time because you are just sending the workload to accel and waiting. Event based async interface is more suitable here.

nickgg:
@opti-mix @yinghai I think you may be conflating the DeviceManager with accessing the real Device. For any given device we may want an event based interface, but the specifics of those interfaces will differ in syntax and semantics in ways we don't currently know. Maybe for a particular device they do use epoll/select, maybe the firmware does blocking calls, maybe they do something really strange, they could even use callbacks (?!). The safest and most flexible thing we can do here is to have as minimal and generic an interface to the device as is practical, and isolate the platform specific code to one place - in this design the accelerator specific DeviceManager implementations.

PR #2137, opened 2018-12-07
[opt] Make convertPlaceholdersToConstants work with quantized types

bertmaher:
*Description*: You should be able to create a `Constant` from a quantized tensor.
*Testing*:  New graphOptz test
*Documentation*: n/a


tests/unittests/graphOptzTest.cpp
```diff
@@ -2423,3 +2423,28 @@ TEST_F(GraphOptz, sinkTransposeBelowChannelShuffleNodesAndEliminate) {
...
+  auto *fc = F_->createFullyConnected("fc", input, weights, bias);
+  F_->createSave("save", fc, output);
+
+  ctx_.allocate(input);
+  ctx_.allocate(weights);
+  ctx_.allocate(bias);
+  ctx_.allocate(output);
+
+  glow::convertPlaceholdersToConstants(F_, ctx_, {input, output});
+  EXPECT_EQ(mod_.getConstants().size(), 2);
```
rdzhabarov: can you comment on what are constants (weight/bias)
and on placeholders (i assume it's input/output, what else?)

bertmaher: Yup, weights/bias are constants.  All 4 things are still placeholders in the module.  Even though weight/bias have been converted, they haven't been removed from the module.  I don't know how to make them go away :-).  I'll add comments to this effect.

tests/unittests/graphOptzTest.cpp
```diff
@@ -2423,3 +2423,31 @@ TEST_F(GraphOptz, sinkTransposeBelowChannelShuffleNodesAndEliminate) {
...
+
+  auto *fc = F_->createFullyConnected("fc", input, weights, bias);
+  F_->createSave("save", fc, output);
+
+  ctx_.allocate(input);
+  ctx_.allocate(weights);
+  ctx_.allocate(bias);
+  ctx_.allocate(output);
+
+  glow::convertPlaceholdersToConstants(F_, ctx_, {input, output});
```
beicy: I thought this stmt convert {input, output} to constant?

bertmaher: It's actually the other way around, but now that you point it out, it does seem like a confusing API.  The reason it's designed that way is because in a lot of places we want to convert a ton of things (every weight) to constant, but only leave a couple things (I/O) as Placeholders :-/

beicy: Sorry I read "!=" to "==" in the source code :P

PR #2138, opened 2018-12-07
[hotfix] fix build of lib/Testing

tlepley-cadence:
*Description*: 
Fix broken build due to #2132 .
Since `Testing` is in `lib/` and `gtest` in `tests`, cmake doesn't seem to properly 'connect' both. Declaring the `gtest` includes at top level like it is already done for `fp16` solves the build issue.
```
glow/lib/Testing/StrCheck.cpp:18:25: fatal error: gtest/gtest.h: No such file or directory
 #include "gtest/gtest.h"
                         ^
compilation terminated.
```

*Testing*: Builds correctly locally

*Documentation*: N/A


stoklund:
Would it be better to move the Testing directory into tests?

tlepley-cadence:
I cancel this fix PR that should be solved by #2141

PR #2141, opened 2018-12-07
Move Testing library into tests/

stoklund:
This way, the library only gets built when GLOW_BUILD_TESTS is set, and we avoid
problems with missing googletest include paths.

cc @tlepley-cadence #2138


tlepley-cadence:
@stoklund that looks good to me. I closed my fix PR.

stoklund:
> All of a sudden the lowercase-ness of our tests/ subdirectories bugs me :-).

If you fix it, everyone on a Mac will have big trouble. Git is case-sensitive and macOS is not, so that kind of rename does not end well ;-)

PR #2142, opened 2018-12-07
[cmake] Shared library dep for onnxifi-glow-lib

bertmaher:
*Description*: I don't understand why CI didn't catch this.
*Testing*: cmake -DBUILD_SHARED_LIBS=ON ...
*Documentation*: n/a

### DIFF
--- a/lib/Onnxifi/CMakeLists.txt
+++ b/lib/Onnxifi/CMakeLists.txt
@@ -18,6 +18,7 @@ add_library(onnxifi-glow-thread-pool
 target_link_libraries(onnxifi-glow-lib
                       PUBLIC
                         ExecutionEngine
+                        Graph
                         Importer
                       PRIVATE
                         onnxifi-glow-thread-pool)


lib/Onnxifi/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ add_library(onnxifi-glow-thread-pool
 target_link_libraries(onnxifi-glow-lib
                       PUBLIC
                         ExecutionEngine
+                        Graph
```
rdzhabarov: Transitive here, onnxifi-glow depends on EE and EE has Graph

rdzhabarov: Does that actually fail locally?

bertmaher: It does.  If I understand cmake correctly (which is questionable!), EE has Graph as a PRIVATE library dependence, so onnxifi-glow-lib won't see it.  I think we don't have this problem with static linking because libEE has the symbols from Graph linked in (maybe?  I didn't check that).

bertmaher: Hm, but that doesn't really make sense, because it's working on CI... I'd like to understand what's up, but I'm inclined to land this and move on :-/

Issue #1950, opened 2018-10-31
[ONNXIFI] Identify incompatibility of ONNX operators in a more reliable way


rdzhabarov:
Currently, Glow implements full [ONNXIFI interface](https://github.com/onnx/onnx/blob/master/onnx/onnxifi.h) with some relaxed checks in several places.

More specifically [onnxBackendCompatibility](https://github.com/pytorch/glow/blob/master/lib/Onnxifi/onnxifiGlow.cpp#L160) does not really parse ONNX operator nor it checks input/output dimensions.
Instead, the method has a mapping between ONNX opType and Glow nodes. Certain Glow nodes can or cannot be executed by a specific backend (which is also checked).
This approach is problematic and could lead to a failure at "onnxInitGraph" point rather than at "onnxBackendCompatibility" which is undesirable.

What we really need to do is to perform full ONNX parsing of the operator and validate whether all parsed Glow nodes could be handled by the specific backend.

rdzhabarov:
@jackm321 this could be a larger task to tackle.

rdzhabarov:
I think we already added support for this, thanks @jackm321

PR #2143, opened 2018-12-07
[onnxifi] Load subgraph to check onnxGetBackendCompatibility

jackm321:
*Description*:
Load and lower subgraphs given to glow through `onnxGetBackendCompatibility` in order to check if they are compatible with glow backend.

Previously Glow just listed out what glow operators we expected a given ONNX graph to be loaded as and lowered to if it were to successfully be loaded at all. This was problematic because it could be inaccurate and also loading may not succeed which will only be discovered during `onnxInitGraph`. After this PR, glow actually attempts to load the ONNX graph and checks the validity along the way, then it lowers the graph and checks each of the glow nodes that were produced to see if they are supported by the backend.

*Testing*:
* `ninja check`
* pytest -s caffe2/python/onnx/test_onnxifi.py` 3/3 tests pass
* added some temporary logging to print out returned statuses and all compatibility checks returned `ONNXIFI_STATUS_SUCCESS` as expected.

*Documentation*:
doxygen


jackm321:
@yinghai @rdzhabarov re: our perf discussion yesterday, I compiled with Release and ran the caffe2 ONNXIFI resnet50 test a few times on my laptop and found that this is on average 190ms slower than the switch statement it replaces for the whole graph.

lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -46,20 +46,27 @@ static llvm::Error setTensorType(const ONNX_NAMESPACE::TypeProto &in,
...
+    if (loadInputsAsPlaceholders) {
+      Tensor T;
+      RETURN_IF_ERR(setTensorType(in.type(), &T));
+
+      Placeholder *placeholder;
+      ASSIGN_VALUE_OR_RETURN_ERR(
+          placeholder, createAndRegisterPlaceholder(in.name(), &T.getType()));
+      onnxNameToInputVars_.try_emplace(in.name(), placeholder);
+    } else {
+      std::unique_ptr<Tensor> T(new Tensor());
```
yinghai: We don't have `make_unique` and stuff like that?

lib/Onnxifi/Base.cpp
```diff
@@ -17,13 +17,52 @@
 
 #include "glow/Importer/ONNXIFIModelLoader.h"
 
+#include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/SmallVector.h"
 
 namespace glow {
 namespace onnxifi {
+onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
```
yinghai: Do we compile or we just do verification?

lib/Onnxifi/Base.cpp
```diff
@@ -17,13 +17,52 @@
...
+    return ONNXIFI_STATUS_INTERNAL_ERROR;
+  }
+
+  glow::lower(function, *backend);
 
-bool BackendId::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) {
-  return executionEngine_.isOpSupported(opKind, elementTy);
+  const auto &nodes = function->getNodes();
+
+  for (const auto &node : nodes) {
```
yinghai: If we are able to lower the function, do we still need this check?

yinghai:
190ms is a bit high but might be OK. Considering it's one time job.

jackm321:
Some onnx tests are broken, I'll look into them

lib/Importer/ONNXIFIModelLoader.cpp
jackm321: Idk, @rdzhabarov mentioned on an earlier PR of his that we shouldn't assume we have it.

lib/Onnxifi/Base.cpp
jackm321: We only import onnx graph to glow graph then lower the glow graph, no optimization, no compiling of any kind. I'll update the comment in Base.h to say this.

lib/Importer/ONNXIFIModelLoader.cpp
bertmaher: Hm, I thought we used gcc-5 in Travis, and make_shared has been around since 4.9.  But I'm not complaining because as long as we don't use make shared I can keep limping along with gcc 4.8 ;-)

yinghai: `make_shared` is c++11 while `make_unique` is c++14 IIRC.....

bertmaher: Sorry, I meant make_unique.  I think make_shared is fine.

include/glow/Importer/ONNXIFIModelLoader.h
```diff
@@ -56,20 +59,14 @@ class ONNXIFIModelLoader : public ONNXModelLoader {
 
   /// \returns a unique_ptr<ONNXIFIModelLoader> if \p onnxModel can be
   /// parsed and static weights can be loaded from the \p wightDescriptors.
-  /// \returns Error otherwise.
+  /// \returns Error otherwise. \p loadInputsAsPlaceholders is passed to
+  /// loadInputs to determine whether inputs are loaded as Placeholders or
+  /// Tensors. Loading as Tensors is useful for when the graph being loaded is
+  /// actually a small slice of a larger graph.
   static llvm::Expected<std::unique_ptr<ONNXIFIModelLoader>>
```
rdzhabarov: btw, can we get rid of unique_ptr for ONNXIFIModelLoader? (the only use case was that we'd check whether it's null or not, but now we have Expected wrapper)

lib/Onnxifi/Base.h
```diff
@@ -41,9 +41,12 @@ class BackendId {
   explicit BackendId(glow::BackendKind kind, int id, int concurrency)
       : id_(id), concurrency_(concurrency), executionEngine_(kind) {}
 
-  /// Verify that given operation kind is supported by the backend.
```
rdzhabarov: why removed? If we can refine the comment - great :)

lib/Onnxifi/Base.h
```diff
@@ -41,9 +41,12 @@ class BackendId {
...
   bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy);
 
+  /// Verify that a given graph is supported by the backend.
```
rdzhabarov: can you outline what steps are taken in the method.

lib/Onnxifi/Base.cpp
rdzhabarov: how do we ensure thread safety here?

lib/Onnxifi/Base.cpp
```diff
@@ -17,13 +17,52 @@
...
+    // TODO: Make is isOpSupported more able to handle different ElemKinds.
+    bool opSupported = getEE().isOpSupported(node.getKind(), ElemKind::FloatTy);
+    if (!opSupported) {
+      // TODO: Use a more specific ONNXIFI error code here to denote what about
+      // this operator is not supporte (shape, type, etc).
```
rdzhabarov: typo

jackm321:
My original suspicion seems to be accurate that there are now fewer tests being skipped by the onnxifi test driver because now more operators are now supported by `onnxGetBackendCompatibility` and some of these newly enabled tests are failing. This can be confirmed by checking out master and forcing `onnxGetBackendCompatibility` to always return `ONNXIFI_STATUS_SUCCESS` and then run only `test_lrn` (one of the failing tests in this diff) and it fails in the same way. This isn't to say that there aren't potentially other issues introduced here but at least some of the issues are from exisiting problems being exposed by this diff because it has caused a larger set of tests to be run.

yinghai:
@jackm321 Sounds great! I suggest we skip newly flushed failed tests and move on with this PR. Meanwhile keep track of those not supposed to fail tests. :)

rdzhabarov:
Awesome progress! 

related issues: https://github.com/pytorch/glow/issues/1950, https://github.com/pytorch/glow/issues/1755

include/glow/Importer/ONNXIFIModelLoader.h
jackm321: We can't because returning a `llvm::Expected<T>` still causes a copy of the underlying `T`, which in most cases is fine and cheap because the expected value is a pointer, int, float, etc but here it would cause issues because it would be expensive and `ONNXIFIModelLoader` owns tensors

lib/Onnxifi/Base.cpp
jackm321: Ah good point, we'll need some locking

jackm321: No locking needed, just created a fresh Module

lib/Importer/ONNXIFIModelLoader.cpp
jackm321: There is also `llvm:: make_unique`

jackm321:
@rdzhabarov any more changes you'd like to see here?

rdzhabarov:
We need to make sure CI passes :)

rdzhabarov:
Once this lands we'd need to run through some of the tests in fbcode to make sure all works as expected. Could you try to run it prior merging this PR?

jackm321:
I wanted to rerun but it says "You need write permissions to trigger builds."

rdzhabarov:
Can you also update the task with failing ONNXIFI_DRIVER_TEST (if we fixed some and got new one failing)

PR #2144, opened 2018-12-07
[support] Move the Onnixifi thread pool into Support/ so it's easier to reuse.

nickgg:
*Description*: Per @jackm321's suggestion, we should reuse the Onnixifi ThreadPool in the runtime, so moving it into the common area.

I also split out the shutdown stage into a separate function to allow more flexible control of shutting down the pool.
*Testing*: ninja test
*Documentation*:

### DIFF
--- a/include/glow/Support/ThreadPool.h
+++ b/include/glow/Support/ThreadPool.h
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef GLOW_ONNXIFI_THREAD_POOL_H
-#define GLOW_ONNXIFI_THREAD_POOL_H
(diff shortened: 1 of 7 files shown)


include/glow/Support/ThreadPool.h
```diff
@@ -13,8 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef GLOW_ONNXIFI_THREAD_POOL_H
-#define GLOW_ONNXIFI_THREAD_POOL_H
+#ifndef GLOW_SUPPORT_THREAD_POOL_H
```
rdzhabarov: nit: THREADPOOL, _ for folder separation

rdzhabarov:
we have nice utility: `./utils/format.sh fix`

PR #2152, opened 2018-12-10
Add clone() method to Context, copying Placeholders with new backing Tensors

nickgg:
*Description*: Function creation builders, particularly in tests, sometimes take Context as a param and create and allocate new Placeholders. To run Functions in parallel we need multiple Context objects (at least for output Tensor memory). This PR adds a clone() function to Context which copies the Placeholders and allocates new backing Tensors for each.
*Testing*: ninja test in Debug & ASAN release.
*Documentation*:


include/glow/Graph/Context.h
```diff
@@ -71,6 +71,10 @@ class Context final {
   /// tensors.
   void clear();
 
+  /// Sets \p cloned to be a copy of the Context, with each placeholder mapped
+  /// to a new Tensor, with their own memory.
+  void clone(Context &cloned);
```
opti-mix: Should `clone` be a const method?

nadavrot:
Nice!

include/glow/Graph/Context.h
```diff
@@ -71,6 +71,10 @@ class Context final {
   /// tensors.
   void clear();
 
+  /// Sets \p cloned to be a copy of the Context, with each placeholder mapped
+  /// to a new Tensor, with their own memory.
+  void clone(Context &cloned) const;
```
rdzhabarov: typically clone() just returns cloned object. What's the benefit of passing it by ref?

lib/Graph/Context.cpp
```diff
@@ -46,6 +46,13 @@ void Context::clear() {
   map_.clear();
 }
 
+void Context::clone(Context &cloned) const {
+  cloned.clear();
+  for (auto PH : map_) {
+    cloned.insert(PH.first, PH.second->clone());
```
rdzhabarov: imo it's hard to read without knowing what's in PH.first/second.

tests/unittests/graphTest.cpp
```diff
@@ -1202,3 +1202,68 @@ TEST(Graph, hookTest) {
   ASSERT_TRUE(ph);
   ASSERT_EQ(ph, in);
 }
+
+TEST(Graph, cloneContext) {
+  ExecutionEngine EE;
```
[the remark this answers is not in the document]
nickgg: Yeah, I picked a Function that was complicated enough to have internal placeholders so the test would cover the broadest case. 

I can definitely do what you want here, but I think it's a worse test for the POV of this change. The reason to make this change is so we can run the same Function concurrently on different Contexts, even if it requires a single context at build time. Removing the Function from the test means makes it pretty artificial. Before I do that, do you think this could be fixed by comments?

lib/Graph/Context.cpp
nickgg: Does my change to pull out the types of PH make it easier to parse?

include/glow/Graph/Context.h
bertmaher: Can we make Context movable?  I don't see a good reason why not, and it makes the type more ergonomic.

lib/Graph/Context.cpp
```diff
@@ -46,6 +46,17 @@ void Context::clear() {
...
 
+Context Context::clone() const {
+  Context cloned;
+  for (auto PH : map_) {
+    Placeholder *P = PH.first;
+    Tensor *T = PH.second;
+    cloned.insert(P, T->clone());
+  }
+
+  return cloned;
```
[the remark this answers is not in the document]
nickgg: I believe it is moved, otherwise I'd expect it to work without an explicit move constructor.

lib/Graph/Context.cpp
```diff
@@ -101,3 +112,12 @@ Context::Context(llvm::ArrayRef<Placeholder *> placeholders,
     insert(placeholders[i], std::move(ptrT));
   }
 }
+
+Context::Context(Context &&other) {
```
bertmaher: Could this be: 
```
Context::Context(Context &&other) : map_(std::move(other.map_)) {}
```

tests/unittests/graphTest.cpp
```diff
@@ -1230,3 +1230,101 @@ TEST(Graph, contextSize) {
   EXPECT_EQ(ctx.get(PH)->size(), 4 * 320 * 200 * 3);
   EXPECT_EQ(ctx.getDataSize(), sizeof(float) * ctx.get(PH)->size());
 }
+
+/// Check that clones of the context are distinct and do not have shared 
+/// references to Tensors.
+TEST(Graph, cloneContext) {
+  ExecutionEngine EE;
```
rdzhabarov: nit: we do not need EE here, just Module.

tests/unittests/graphTest.cpp
```diff
@@ -1230,3 +1230,101 @@ TEST(Graph, contextSize) {
...
+}
+
+/// Check that running a function multiple times on cloned Contexts have
+/// distinct outputs.
+TEST(Graph, cloneContextRuns) {
+  ExecutionEngine EE;
+  PseudoRNG PRNG;
+
+  Tensor inputs(ElemKind::FloatTy, {1, 32, 32, 3});
+  // copied this network from nodesWithPredicates.
```
rdzhabarov: comment might be irrelevant

PR #2154, opened 2018-12-10
[Support] switch from std::function to std::packaged_task in ThreadPool

nickgg:
*Description*: The ThreadPool is storing tasks as std::functions which prevents using move only types in the lambda capture. std::packaged_task is a better abstraction for one time run lambdas and does not have this limitation. We'll have to monitor performance of the thread pool as we scale up.
*Testing*: ninja test, but especially two new ThreadPool tests - one which verifies move-only captures, and another which demonstrates we can use the packaged_tasks's included future to tell when the task has executed.
*Documentation*: n/a

### DIFF
--- a/include/glow/Support/ThreadPool.h
+++ b/include/glow/Support/ThreadPool.h
@@ -19,6 +19,7 @@
 #include <atomic>
 #include <condition_variable>
 #include <functional>
+#include <future>
 #include <mutex>
 #include <queue>
 #include <thread>
@@ -38,7 +39,20 @@ public:
   ~ThreadPool();
 
   /// Submit \p fn as a work item for the thread pool.
-  void submit(const std::function<void(void)> &fn);
+  /// \p fn must be a lambda with void return type and arguments.
+  template <typename F> std::future<void> submit(F &&fn) {
+    // Add fn to the work queue.
+    std::unique_lock<std::mutex> lock(workQueueMtx_);
+    std::packaged_task<void(void)> task(std::move(fn));
+    auto future = task.get_future();
+    workQueue_.push(std::move(task));
+    lock.unlock();
+    queueNotEmpty_.notify_one();
+    return future;
+  }
+
+  /// Submit \p task as a work item for the thread pool.
+  std::future<void> submit(std::packaged_task<void(void)> &&task);
 
   /// Stop all threads and optionally wait for them to join.
   void stop(bool block = false);
@@ -55,7 +69,7 @@ private:
   std::atomic<bool> shouldStop_;
 
   /// Queue of work items.
-  std::queue<std::function<void(void)>> workQueue_;
+  std::queue<std::packaged_task<void(void)>> workQueue_;
 
   /// Mutex to coordinate access to the work queue.
   std::mutex workQueueMtx_;
--- a/lib/Support/ThreadPool.cpp
+++ b/lib/Support/ThreadPool.cpp
@@ -27,12 +27,13 @@ ThreadPool::ThreadPool(unsigned numWorkers) : shouldStop_(false) {
 
 ThreadPool::~ThreadPool() { stop(true); }
 
-void ThreadPool::submit(const std::function<void(void)> &fn) {
-  // Add fn to the work queue.
+std::future<void> ThreadPool::submit(std::packaged_task<void(void)> &&task) {
   std::unique_lock<std::mutex> lock(workQueueMtx_);
-  workQueue_.push(fn);
+  auto future = task.get_future();
+  workQueue_.push(std::move(task));
   lock.unlock();
   queueNotEmpty_.notify_one();
+  return future;
 }
 
 void ThreadPool::stop(bool block) {
@@ -81,7 +82,7 @@ void ThreadPool::threadPoolWorkerMain() {
 
     // Pop a work item from the queue, and make sure to unlock
     // the lock before processing it.
-    auto workItem = workQueue_.front();
+    auto workItem = std::move(workQueue_.front());
     workQueue_.pop();
     lock.unlock();
 
--- a/tests/unittests/ThreadPoolTest.cpp
+++ b/tests/unittests/ThreadPoolTest.cpp
@@ -49,3 +49,29 @@ TEST(ThreadPool, BasicTest) {
     EXPECT_EQ(result, 2 * i);
   }
 }
+
+TEST(ThreadPool, moveCaptureTest) {
+  ThreadPool tp(1);
+
+  std::unique_ptr<int> input = std::make_unique<int>(42);
+  int output = 0;
+  auto func = [input = std::move(input), &output]() { output = (*input) * 2; };
+
+  auto done = tp.submit(std::move(func));
+
+  done.wait();
+  EXPECT_EQ(output, 84);
+}
+
+TEST(ThreadPool, completionFutureTest) {
+  ThreadPool tp(1);
+
+  int input = 42, output = 0;
+  std::packaged_task<void(void)> task(
+      [&input, &output]() { output = input * 3; });
+
+  auto done = tp.submit(std::move(task));
+
+  done.wait();
+  EXPECT_EQ(output, 126);
+}


nickgg:
One alternative interface is for submit() to return the future attached the packaged_task, which would be easier if we were using the second submit() overload a lot for that purpose.

nickgg:
Yeah I think that works better, interface wise.

lib/Support/ThreadPool.cpp
```diff
@@ -1,3 +1,4 @@
+
```
bertmaher: nit,  remove the blank line here at top-of-file

include/glow/Support/ThreadPool.h
```diff
@@ -38,7 +39,20 @@ class ThreadPool final {
   ~ThreadPool();
 
   /// Submit \p fn as a work item for the thread pool.
-  void submit(const std::function<void(void)> &fn);
+  /// \p fn must be a lambda with void return type and arguments.
+  template <typename F> std::future<void> submit(F &&fn) {
```
SplitInfinity: Did you mean to put the implementation of this in the header? Can you move it to the source file?

nickgg: It's a templated method, we can't implement it in the source file without binding the template parameter, which for lambdas is inconvenient.

ayermolo:
@nickgg  After this change I see failure with Visual Studio:

5>ThreadPoolTest.cpp
5>C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\include\functional(1195): error C2280: 'ThreadPool_moveCaptureTest_Test::TestBody::<lambda_73ed095b2309a32b00cb42e4033a417b>::<lambda_73ed095b2309a32b00cb42e4033a417b>(const ThreadPool_moveCaptureTest_Test::TestBody::<lambda_73ed095b2309a32b00cb42e4033a417b> &)': attempting to reference a deleted function
5>C:\projects\glow\tests\unittests\ThreadPoolTest.cpp(58): note: see declaration of 'ThreadPool_moveCaptureTest_Test::TestBody::<lambda_73ed095b2309a32b00cb42e4033a417b>::<lambda_73ed095b2309a32b00cb42e4033a417b>'
5>C:\projects\glow\tests\unittests\ThreadPoolTest.cpp(58): note: 'ThreadPool_moveCaptureTest_Test::TestBody::<lambda_73ed095b2309a32b00cb42e4033a417b>::<lambda_73ed095b2309a32b00cb42e4033a417b>(const ThreadPool_moveCaptureTest_Test::TestBody::<lambda_73ed095b2309a32b00cb42e4033a417b> &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'std::unique_ptr<int,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)'
5>        with
5>        [
5>            _Ty=int
5>        ]

nickgg:
I will take a look.

ayermolo:
Thanks. Really weird issue. I tried to reproduce it with small example, and it compiled fine. So generic lambda capture of unque_ptr works, passing it by move semantic works. Not clear to me what is going wrong in this case.

Toy example works fine:
class Test2 {
public:
  template <typename F> void doStuff(F &&fn) { fn(); };
};

class Test {
public:
  template <typename F> void doStuff(F &&fn) { Test2 t2; t2.doStuff(fn); };
};
int main()
{
    std::unique_ptr<int> input = std::make_unique<int>(42);
    int output = 0;
    auto func = [input = std::move(input), &output]() { output = (*input) * 2; };
    Test t;
    t.doStuff(std::move(func));
    std::cout << output << "Hello World!\n"; 
}

nickgg:
That is baffling.

My only thought is that maybe the error message is off and it's something to do with the std::packaged_task, googling around I see that MSVCs package_task implementation was wrong in the past and stored the lambda in a std::function which would cause the issue you're seeing: https://developercommunity.visualstudio.com/content/problem/108672/unable-to-move-stdpackaged-task-into-any-stl-conta.html

I'll try to track down a Windows machine to test on, but in the mean time could you try your example replacing the Test object with:
```
std::packaged_task<void(void)> task(std::move(func));
task();
```

ayermolo:
Yep that triggered it.
`int main()
{
    std::unique_ptr<int> input = std::make_unique<int>(42);
    int output = 0;
    auto func = [input = std::move(input), &output]() { output = (*input) * 2; };
    //Test t;
    //t.doStuff(std::move(func));
    std::packaged_task<void(void)> task(std::move(func));
    task();
    std::cout << output << "Hello World!\n"; 
}`

ayermolo:
Doesn't look like they are going to fix it anytime soon. heh

nickgg:
argh, that is frustrating.

It's possible that it is fixed in the Visual Studio 2019 Preview?

nadavrot:
@ayermolo I wonder if porting the interface to using llvm's function_ref could solve the problem.

https://llvm.org/docs/ProgrammersManual.html#the-function-ref-class-template

ayermolo:
@nadavrot Something like this? Yeah it works in VS.

```
  auto func = [input = std::move(input), &output]() { output = (*input) * 2; };
  auto done = tp.submit2(llvm::function_ref<void()>(func));

    template <typename F> std::future<void> submit2(llvm::function_ref<F> &fn) {
    // Add fn to the work queue.
    std::unique_lock<std::mutex> lock(workQueueMtx_);
    
    std::packaged_task<void(void)> task(std::move(fn));
    auto future = task.get_future();
    workQueue_.push(std::move(task));
    lock.unlock();
    queueNotEmpty_.notify_one();
    return future;
  }
```

ayermolo:
@nickgg I downloaded VS 19 preview, and still see the error. :/

ayermolo:
@nickgg @nadavrot What do you think is re-do interface using function_ref, or maybe overload submit function to use either, and just ifdef unit test related to original one? Original implementation is cleaner, and the problematic usage model is only exposed in unittest.

nadavrot:
@ayermolo:  @nickgg is on vacation. I think that they plan to change the implementation so that we don't pass smart pointers that are not copyable (just raw pointers) and this could allow a simpler design. Let's wait for Nick's return after the holidays and continue this discussion.

ayermolo:
Sounds good. Happy Holidays!

ayermolo:
@nickgg Any update on this? Thanks

nickgg:
Ah, sorry @ayermolo this slipped through the cracks, I'll take a look today.

nickgg:
So the FunctionRef implementation doesn't work because its a non owning reference. It relies on the caller of submit holding the function object (and everything in it's lambda capture) alive until the function executes. Not a usable restriction, I think.

E.g. this code crashes because the unique_ptr is destroyed by the time the function runs:

```
ThreadPool tp(5);

std::promise<void> done;
std::atomic<int> sum;
sum = 0;
for (int i = 0; i < 5; ++i) {
  std::unique_ptr<int> number = std::make_unique<int>(i);

  auto func = [n = std::move(number), &sum, &done] {
    sum += *n;
    if (sum == 10) {
      done.set_value();
    }
  };

  tp.submit(llvm::function_ref<void()>(func));
}

done.get_future().wait();
```

To support windows here we have to not use non-copyable types in ThreadPool tasks, which I think practically means we shouldn't use unique_ptr in the runtime at all. We have other non-movable types in the Graph but I guess just pass raw pointers?

ayermolo:
What about using shared_ptr?
Basically the approach in shared_function
https://stackoverflow.com/questions/25421346/how-to-create-an-stdfunction-from-a-move-capturing-lambda-expression

I tried and it does work with your example. Keeps lambda, and captured unique_ptr, around, and then frees it at tidy part.

nickgg:
Yes that would work and take care of the delete. The reason we went with unique_ptr was to avoid duplicate access to the object (and so avoid the need for synchronization) - we'll have to work out what to do about that.

ayermolo:
Something like this, or you had something else in mind?

```
  ThreadPool tp(5);

  std::promise<void> done;
  std::atomic<int> sum;
  sum = 0;

  for (int i = 0; i < 10; ++i) {
    auto func = [&sum, &done] {
      sum += 1;
      if (sum >= 7) {
        done.set_value();
      }
    };

    tp.submit(func);
  }

  done.get_future().wait();
```

nickgg:
That probably works for tests but has the same problem of the above for general usage: the caller needs to store the promise (not just the future) until the callback completes. I suppose we could use shared_ptr?

ayermolo:
I just wasn't sure which code sequence you wanted me to test in VS.

This is a local fix I did for us after latest rebase (shared_function from stackoverflow link). 

```
template <class F> struct shared_function {
  std::shared_ptr<F> f;
  shared_function() = delete; // = default works, but I don't use it
  shared_function(F &&f_) : f(std::make_shared<F>(std::move(f_))) {}
  shared_function(shared_function const &) = default;
  shared_function(shared_function &&) = default;
  shared_function &operator=(shared_function const &) = default;
  shared_function &operator=(shared_function &&) = default;
  template <class... As> auto operator()(As &&... as) const {
    return (*f)(std::forward<As>(as)...);
  }
};
template <class F>
shared_function<std::decay_t<F>> make_shared_function(F &&f) {
  return {std::forward<F>(f)};
}

RunIdentifierTy
QueueBackedDeviceManager::runFunction(std::string functionName,
                                      std::unique_ptr<Context> ctx,
                                      ResultCBTy callback) {

  RunIdentifierTy id = nextIdentifier_++;
  auto func = [this, id, functionName = std::move(functionName),
               ctx = std::move(ctx), callback = std::move(callback)]() mutable {
    runFunctionImpl(id, std::move(functionName), std::move(ctx),
                    std::move(callback));
  };

  std::function<void()> funcShared = make_shared_function(std::move(func));
  workThread_.submit(funcShared);
  return id;
}
```

Maybe throw it for general discussion in it's own issue?

nickgg:
Yes definitely that sounds good, thanks!

ayermolo:
@nickgg Can you file it? I think you are more familiar with this part of code base.

nickgg:
I've filed an issue, I'll need to borrow a windows machine to test it out though, so if you have time to make the change it'll likely be faster.

ayermolo:
Thanks. Posted a link to one solution.

PR #2157, opened 2018-12-11
[utils] Add script to enumerate opencl devices

MatzeB:
*Description*:
This helps to figure out possible values for the `-device` and
`-platform` flags of the opencl backend.


MatzeB:
Ah didn't know about `clinfo` before.

Issue #1948, opened 2018-10-31
Make the execution of the Glow compiler more deterministic between multiple runs


opti-mix:
There are certain parts of the compiler that expose some levels of non-determinism. And I don't mean by this that numeric results of running an NN model are non-deterministic.

Instead, I mean that some observable (via logs, traces, debug prints) decisions made by the compiler internally may be non-deterministic.

For example, many backends statically assign memory addresses to weights and activations. The order in which these addresses are assigned depends on the order of iteration over the set of weights/activations defined by a given graph/function. And these objects are currently often hold in e.g. unordered maps. As a result, the iteration order is non-deterministic and Glow may statically assign different addresses to the same weight/activation if you'd run it multiple times on the same inputs. 

This kind of non-determinism is very annoying, because it complicates the debugging and makes comparing the traces produced by different runs of the compiler (e.g. by means of -debug-glow-only=xyz) very difficult.

The aim of this issue is therefore:
- Identify the places where we have this kind of non-determinism
- Fix such issues by e.g. using data structures with a deterministic iteration order (map instead of unordered_map, etc). Depending on the case, we may want to either use such a deterministic data structure either for the whole duration of the compilation or only inside the code snippet that exposes the non-deterministic behavior.

tlepley-cadence:
Unordered data structures may actually be deterministic. It depends how they are implemented. Even (pseudo) random generators may be deterministic (for instance, same seed may result in the same list of random numbers). What is non deterministic is for instance to initialize a random number generator by reading an uninitialized memory location. What creates non determinism is often a bug (data race, memory related bugs).

bertmaher:
I think it’s not necessarily the unordered-ness that causes problems, but the use of pointers as keys.

nickgg:
@opti-mix is this still an issue?

opti-mix:
>  is this still an issue?

I think we have not expected all places where we iterate over unordered maps with pointer keys and the like, which are the typical places where the non-determinism comes from.

nadavrot:
@opti-mix What is actionable in this PR?

opti-mix:
> What is actionable in this PR?

We need to e.g. inspect the following:
- all places were we perform static memory allocation for different backends and check that we assign the same addresses when we compile the same inputs multiple times by different compiler runs. very likely we need to use maps instead of unordered_maps for iteration over e.g. placeholders, Values, etc.

[...]

PR #2158, opened 2018-12-11
Fix some nondeterminism

MatzeB:
*Description*:
Contains two patches to avoid two cases of nondeterminism in the compiler.

*Testing*:
see individal commits.

*Documentation*:
* #1948

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -25,6 +25,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Format.h"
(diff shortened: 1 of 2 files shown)


rdzhabarov:
running`./utils/format.sh fix` should fix travis failure

lib/Graph/Graph.cpp
```diff
@@ -177,9 +180,15 @@ class AbstractDottyPrinter {
   }
 
   std::string uniqueVertexName(void *N) {
+    VertexNumberMap::iterator i;
+    bool inserted;
+    std::tie(i, inserted) = vertex_numbers.insert(std::make_pair(N, 0u));
+    if (inserted)
```
opti-mix: We usually use `{}` even in such single-line cases. It is described somewhere in FB-style.

PR #2161, opened 2018-12-11
[WIP] Adding RuntimeTypes.h 

gcatron:
*Description*: Adds RuntimeTypes.h which contains data structures common to the Runtime Components.
*Testing*: Not yet
*Documentation*: NA
Related to #2045

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -30,6 +30,8 @@ struct RuntimeSymbolInfo {
   size_t offset;
   /// Type of symbol.
   Type type;
+  /// Is the symbol an input for the function.
+  bool input;
 };
 
 /// Contains the information needed to be passed forward from compile time to
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -32,6 +32,8 @@ struct RuntimeSymbolInfo {
   Type type;
   /// Is the symbol an input for the function.
   bool input;
+  /// Is the symbol an output for the function.
+  bool output;
 };
 
 /// Contains the information needed to be passed forward from compile time to
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -62,7 +62,7 @@ struct ExecutionDAG {
 
 /// Data structure containing the output from the Partitioner. It is consumed by
 /// the Provisioner and used to generate an executionDAG.
-struct DependancyDAG {
+struct DependencyDAG {
   /// Vector of unique pointers to modules containing sub-networks.
   std::vector<std::unique_ptr<Module>> modules;
   /// Vector of root nodes.
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -32,13 +32,11 @@ using DeviceIDty = size_t;
 /// and runtime.
 enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
 
-/// Data structure that contains memory information for each device. Used to
-/// communicate memory constraints to the Partitioner and Provisioner.
-struct DeviceMemoryInfo {
+/// Data structure that contains device constraint information for each device.
+/// Used to communicate memory constraints and later costs to the Partitioner.
+struct DeviceInfo {
   /// Available memory on device in bytes.
   uint64_t availableMemory;
-  /// Total useable memory on device in bytes.
-  uint64_t maximumUsableMemory;
 };
 
 /// Data structure that contains everything needed by the executor to execute a
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_RUNTIME_RUNTIMETYPES_H
 #define GLOW_RUNTIME_RUNTIMETYPES_H
 
+#include "glow/Backends/BackendUtils.h"
 #include "glow/Graph/Graph.h"
 
 #include <string>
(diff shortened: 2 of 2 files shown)


include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,84 @@
...
+/// the Provisioner and used to generate an executionDAG.
+struct DependancyDAG {
+  /// Vector of unique pointers to modules containing sub-networks.
+  std::vector<std::unique_ptr<Module>> modules;
+  /// Vector of root nodes.
+  std::vector<Module *> roots;
+  /// Mapping of Module * to a vector of it's dependancies.
+  std::unordered_map<Module *, std::vector<Module *>> dependencies;
+  /// Mapping of Module * to a vector of networks that depend on it.
+  std::unordered_map<Module *, std::vector<Module *>> dependents;
```
beicy: Is this a revert DAG ?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,84 @@
...
+  /// Mapping of Module * to a vector of input placeholder names for the
+  /// network.
+  std::unordered_map<Module *, std::vector<std::string>> inputs;
+  /// Mapping of Module * to a vector of output placeholder names for the
+  /// network.
+  std::unordered_map<Module *, std::vector<std::string>> outputs;
+};
```
beicy: Each module contains its placeholder and constants. So what the` input` and `output `for?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,84 @@
...
+/// and runtime.
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
+
+struct DeviceMemoryInfo {
+  /// Available memory on device in MB
+  double availableMemory;
+  /// Total useable memory on device in MB
+  double maximumUsableMemory;
+};
+
```
beicy: For Partitioner, we need the DeviceInfo contains: memory, number of devices, later the communication constraints.

include/glow/Runtime/RuntimeTypes.h
gcatron: yes, after talking with @SplitInfinity we decided it would be helpful to have both directions. If it turns out this can be removed we can pull it later.

include/glow/Runtime/RuntimeTypes.h
gcatron: Would the communication contraints be at a system level or device level? The thought here is the Partitioner would be given a vector of these DeviceInfo objects, one per device.

include/glow/Runtime/RuntimeTypes.h
gcatron: They are for runtime, at runtime the modules are no longer around but the executor will need to know how to pass inputs and outputs when executing the network.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,84 @@
...
+  /// network.
+  std::unordered_map<NetworkID, std::vector<std::string>> inputs;
+  /// Mapping of NetworkID to a vector of output placeholder names for the
+  /// network.
+  std::unordered_map<NetworkID, std::vector<std::string>> outputs;
+};
+
+/// Data structure containing the output from the Partitioner. It is consumed by
+/// the Provisioner and used to generate an executionDAG.
+struct DependancyDAG {
```
bertmaher: I prefer the spelling "dependency", but Wiktionary informs me that "dependancy" is an acceptable if archaic form.  So thou mayest leave the spelling alone if thou whilst :-D

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,84 @@
...
+using NetworkID = size_t;
+using DeviceID = size_t;
+
+/// Enum to communicate results when communicating with device at initialization
+/// and runtime.
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
+
+struct DeviceMemoryInfo {
+  /// Available memory on device in MB
+  double availableMemory;
```
bertmaher: I'd use `size_t` for memory, since it's not really a floating-point quantity.  I also have a slight preference for bytes (rather than a bigger unit like MB), but I also know it's kind of annoying to look at quantities like 536,870,912 instead of "512", so your discretion there.

include/glow/Runtime/RuntimeTypes.h
gcatron: Good catch I prefer dependency myself, consider it changed.

include/glow/Runtime/RuntimeTypes.h
bertmaher: Actually, `size_t` might be a little awkward since it's based on the host's address space.  In practice I don't think it would ever matter, but maybe a `uint64_t` would be better?

include/glow/Runtime/RuntimeTypes.h
gcatron: Removed from DependencyDAG, they will by generated by the HM and added to the executionDAG.

include/glow/Runtime/RuntimeTypes.h
gcatron: changed to uint64_t.

include/glow/Runtime/RuntimeTypes.h
beicy: "Would the communication contraints be at a system level or device level?"
-- device level. Should be device specific.  
It is OK to put it here, but for Partitioner, I don't think we need "double availableMemory;". Please correct me if I miss anything.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,78 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+namespace runtime {
+
+using NetworkID = size_t;
```
nadavrot: This is not the device id. It is the device id *type*.  Please call it NetworkIDTy

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,78 @@
...
+using DeviceID = size_t;
+
+/// Enum to communicate results when communicating with device at initialization
+/// and runtime.
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
+
+struct DeviceMemoryInfo {
```
nadavrot: oxygen.

include/glow/Runtime/RuntimeTypes.h
gcatron: You are correct, we don't need availableMemory right now for Partitioner, but it could be used for provisioner and potentially later in Partitioner.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,67 @@
...
+/// Used to communicate memory constraints and later costs to the Partitioner.
+struct DeviceInfo {
+  /// Available memory on device in bytes.
+  uint64_t availableMemory;
+};
+
+/// Individual Node in the DAG for a given network. This contains all the
+/// information needed to run the sub-network at inference time.
+struct DAGNode {
+  /// The children of this node, these are nodes that depend on the current
```
beicy: So, after graph partitioning, each function should have a DAGNode, right? Please correct me if I misunderstand something... If so, we should have some field in this struct to point the function ?  or the "name" in DAGNode is the function name?

gcatron: You are correct, name in the DAGNode is the function name.

beicy:
Looks like ci failed due to the head file is not used by any other file... Anyway to avoid this failure?

include/glow/Backends/BackendUtils.h
```diff
@@ -30,6 +30,10 @@ struct RuntimeSymbolInfo {
   size_t offset;
   /// Type of symbol.
   Type type;
+  /// Is the symbol an input for the function.
```
beicy: please run "format.sh fix"

gcatron: Format.sh fix didn't change anything? Is it the indentation?

beicy: Yes, I would like the indentation should be the consistent with the rest of file.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,67 @@
...
+  /// node.
+  std::vector<DAGNode> children;
+  /// Pointers to the parents of this node. This is used by the executor for
+  /// determining if a given node has all dependencies met.
+  std::vector<DAGNode *> parents;
+  /// ID of the deviceManager that this network is assigned to.
```
beicy: The "children" is with the type of   "std::vector<DAGNode>" and "parents"  is with the type of "std::vector<DAGNode*>". Actually, this would cause some complexity when creating the DAGNode. For each DAGNode of each sub-function, if we declare it as a local var, we have no idea how to push_back its pointer to its children's "parents" filed. If we new a DAGNode for each function, the "children" field is improper. After discussed with @gcatron , can we use "std::vector<DAGNode*> children;" instead of "std::vector<DAGNode> children;",  and add an Destructor of DAGNode? @nickgg

PR #2162, opened 2018-12-12
[GraphOptimizer] Fold Quantize into Splat

jfix71:
*Description*: Self-explanatory. Same idea as folding quantize into Constant.

*Testing*: Added a couple tests.


rdzhabarov:
curious what was the ASAN issue? anything interesting?

jfix71:
Not really 😃-- I was getting the value of the original Splat that was DCE'd away, so it was use-after-free. [Here's the change](https://github.com/pytorch/glow/compare/938fd5c2f2699c66583355d86c53c2952628c0d2..aa60680431a09e626f309ca73664e841411a2e76).

PR #2163, opened 2018-12-12
[graph] Add methods to Tensor, Module and Context to get the memory size of contained data.

nickgg:
*Description*: Helper methods for retrieving the size used by the data parts of Module and Context, for use in provisioning.
*Testing*: Ninja all + 2 new tests
*Documentation*: n/a

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -121,6 +121,9 @@ public:
   /// \returns the number of elements in the tensor.
   size_t size() const { return type_.size(); }
 
+  /// \returns the number of bytes required to store the tensor.
+  uint64_t getSizeInBytes() const { return type_.getSizeInBytes(); }
+
   /// Initialize an empty tensor.
   Tensor() = default;
 
--- a/include/glow/Graph/Context.h
+++ b/include/glow/Graph/Context.h
@@ -74,6 +74,9 @@ public:
   /// \returns the mapping between placeholder to tensors.
   const PlaceholderMap &pairs() const { return map_; }
 
+  /// \returns the size in bytes of allocated Tensors owned by Context.
+  uint64_t getDataSize() const;
+
   Context() = default;
 
   /// Construct the Context with an initial mapping between \p placeholders
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -181,6 +181,9 @@ public:
   /// Erase a function \p F from the module.
   void eraseFunction(Function *F);
 
+  /// \Returns the size in bytes of data used by constants.
+  uint64_t getConstantsSize();
+
   // Don't copy or move this class around.
   // The destructor will wipe the functions leaving
   // the original Module only dangling pointers.
--- a/lib/Graph/Context.cpp
+++ b/lib/Graph/Context.cpp
@@ -80,6 +80,15 @@ Placeholder *Context::getFirstUnallocated(std::list<Placeholder *> &lst) const {
   return nullptr;
 }
 
+uint64_t Context::getDataSize() const {
+  uint64_t size = 0;
+  for (const auto &PH : map_) {
+    Tensor *T = PH.second;
+    size += T->getSizeInBytes();
+  }
+  return size;
+}
+
 Context::Context(llvm::ArrayRef<Placeholder *> placeholders,
                  llvm::ArrayRef<Tensor *> inputs) {
   assert(placeholders.size() == inputs.size() &&
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -295,6 +295,14 @@ void Module::eraseFunction(Function *F) {
   delete F;
 }
 
+uint64_t Module::getConstantsSize() {
+  uint64_t size = 0;
+  for (auto *constant : constants_) {
+    size += constant->getPayload().getSizeInBytes();
+  }
+  return size;
+}
+
 Function::~Function() {
   // Delete all of the nodes.
   for (auto it = nodes_.begin(), e = nodes_.end(); it != e;) {
--- a/tests/unittests/graphTest.cpp
+++ b/tests/unittests/graphTest.cpp
@@ -1202,3 +1202,31 @@ TEST(Graph, hookTest) {
   ASSERT_TRUE(ph);
   ASSERT_EQ(ph, in);
 }
+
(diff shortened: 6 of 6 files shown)


include/glow/Graph/Context.h
```diff
@@ -74,6 +74,9 @@ class Context final {
   /// \returns the mapping between placeholder to tensors.
   const PlaceholderMap &pairs() const { return map_; }
 
+  /// \returns the size in bytes of all allocated Tensors
```
rdzhabarov: nit: '.' at the end

rdzhabarov: maybe "allocated Tensors owned by Context"

include/glow/Graph/Graph.h
```diff
@@ -181,6 +181,9 @@ class Module final {
   /// Erase a function \p F from the module.
   void eraseFunction(Function *F);
 
+  /// \Returns the size in bytes of data used by constants.
+  uint64_t dataSize();
```
rdzhabarov: would be clear to specify that it's for constants. maybe constantsSize() or something.

lib/Graph/Context.cpp
```diff
@@ -80,6 +80,15 @@ Placeholder *Context::getFirstUnallocated(std::list<Placeholder *> &lst) const {
   return nullptr;
 }
 
+uint64_t Context::dataSize() const {
+  size_t size = 0;
+  for (auto PH : map_) {
```
rdzhabarov: const ref to avoid copies?

lib/Graph/Context.cpp
```diff
@@ -80,6 +80,15 @@ Placeholder *Context::getFirstUnallocated(std::list<Placeholder *> &lst) const {
   return nullptr;
 }
 
+uint64_t Context::dataSize() const {
+  size_t size = 0;
```
jfix71: `uint64_t` instead of `size_t`? Since that's the return type, and also type returned by `sizeInBytes()`.

lib/Graph/Graph.cpp
```diff
@@ -295,6 +295,14 @@ void Module::eraseFunction(Function *F) {
   delete F;
 }
 
+uint64_t Module::dataSize() {
+  size_t size = 0;
```
jfix71: `uint64_t` again

tests/unittests/graphTest.cpp
```diff
@@ -1202,3 +1202,28 @@ TEST(Graph, hookTest) {
   ASSERT_TRUE(ph);
   ASSERT_EQ(ph, in);
 }
+
```
jfix71: Doxygen comments on both tests, please!

include/glow/Base/Tensor.h
```diff
@@ -121,6 +121,9 @@ class Tensor final {
   /// \returns the number of elements in the tensor.
   size_t size() const { return type_.size(); }
 
+  /// \returns the number of bytes required to store the tensor.
+  uint64_t sizeInBytes() const { return size() * type_.getElementSize(); }
```
opti-mix: Why not  just `type_.getSizeInBytes()` as a body of this function? Also, may be call it `getSizeInBytes` similar to how it is called for types?

lib/Graph/Context.cpp
opti-mix: `getDataSize` as we also have `getSizeInBytes` and `getElementSize` in Glow's source code?

lib/Graph/Graph.cpp
opti-mix: `getDataSize` as we also have `getSizeInBytes` and `getElementSize` in Glow's source code?

tests/unittests/graphTest.cpp
```diff
@@ -1202,3 +1202,28 @@ TEST(Graph, hookTest) {
...
+TEST(Graph, moduleSize) {
+  Module mod;
+
+  EXPECT_EQ(mod.dataSize(), 0);
+
+  mod.createConstant(ElemKind::FloatTy, {1}, "var");
+  EXPECT_EQ(mod.dataSize(), 4);
```
opti-mix: Use `sizeof(float)` instead of 4.

tests/unittests/graphTest.cpp
```diff
@@ -1202,3 +1202,28 @@ TEST(Graph, hookTest) {
...
+TEST(Graph, contextSize) {
+  Module mod;
+  Context ctx;
+
+  Placeholder *PH =
+      mod.createPlaceholder(ElemKind::FloatTy, {4, 320, 200, 3}, "input", true);
+
+  EXPECT_EQ(ctx.dataSize(), 0);
+  ctx.allocate(PH);
+  uint64_t expected = 4 * 4 * 320 * 200 * 3; // 4 byte floats * Tensor size.
```
opti-mix: Use `sizeof(float) * PH->getType().size()` instead.

tests/unittests/graphTest.cpp
```diff
@@ -1202,3 +1202,28 @@ TEST(Graph, hookTest) {
...
+TEST(Graph, moduleSize) {
+  Module mod;
+
+  EXPECT_EQ(mod.dataSize(), 0);
+
+  mod.createConstant(ElemKind::FloatTy, {1}, "var");
+  EXPECT_EQ(mod.dataSize(), 4);
+
+  mod.createConstant(ElemKind::FloatTy, {1, 32, 32, 16}, "var2");
+  EXPECT_EQ(mod.dataSize(), 4 + 4 * 32 * 32 * 16);
```
opti-mix: If you'd assign these constants to e.g. `const1` and `const2`, you could have a more future-proof check like:
`EXPECT_EQ(mod.dataSize(), const1->getType()->getSizeInBytes() + const2->getType()->getSizeInBytes());`

tests/unittests/graphTest.cpp
opti-mix: Or even `PH->getType().getSizeInBytes()`

tests/unittests/graphTest.cpp
nickgg: I don't think testing getSizeInBytes() would be very useful because that is the mechanism used to return mod.dataSize() - you're essentially expecting X == X. A check like that would pass even if the implementation is wrong, whereas if we check against a constant we're verifying the implementation is behaving as expected.

nickgg: Sure that works for me.

PR #2164, opened 2018-12-12
[Improvement] Building libjit in VS using clang

ayermolo:
*Description*: Modifying CPU backend cmake file to use clang to build libjit files within VS.
*Testing*: UnitTests linux/windows
*Documentation*: https://cmake.org/cmake/help/v3.2/command/add_custom_target.html
[Optional Fixes #issue]

I would like to revisit this. This approach is similar to another PR, #1849, that was close due to inactivity, but a more generic approach. It doesn't hardcode files, and plays around with add_custom_target, and add_custom_command to address some of the issues in another PR.

Behavior I see:
Visual Studio
After touch/modification of libjit:
1>Generating CPURuntime/libjit_obj_file_1.o
1>Generating ../../../libjit.bc
1>Generating ../../../glow/libjit_bc.inc
========== Build: 1 succeeded, 0 failed, 2 up-to-date, 0 skipped ==========

Doing Build again:
1>------ Build started: Project: CPURuntime, Configuration: Release x64 ------
========== Build: 1 succeeded, 0 failed, 2 up-to-date, 0 skipped ==========

LINUX
ayermolo:buildGlowOnly$ touch ../projectGit/glow/lib/Backends/CPU/libjit/libjit.cpp
ayermolo:buildGlowOnly$ $GLOW_REQ/bin/ninja all
[4/12] Building CXX object lib/Backends/CPU/CMakeFiles/CPURuntimeNative.dir/libjit/libjit.cpp.o

ayermolo:buildGlowOnly$ $GLOW_REQ/bin/ninja all
ninja: no work to do.

### DIFF
--- a/lib/Backends/CPU/CMakeLists.txt
+++ b/lib/Backends/CPU/CMakeLists.txt
@@ -13,52 +13,58 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 set(CMAKE_LLIR_CREATE_SHARED_MODULE "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 set(CMAKE_LLIR_CREATE_STATIC_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 
-add_library(CPURuntime
-              libjit/libjit.cpp
-              libjit/libjit_conv.cpp
-              libjit/libjit_matmul.cpp)
+set(CPURunttimeCompileOptions
+      -std=c++11
+      -ffast-math
+      -fno-finite-math-only
+      -g0
+      -emit-llvm 
+      -O0)
 
-# NOTE(abdulras) explicitly override the compiler invocations with a custom
-# rule.  The trailing `#` is the comment leader intended to nullify the compile
-# commands.  Doing this allows us to override the compiler (driver) used for
-# building the runtime which requires clang (to emit LLVM IR for the LTO'ed AOT
-# JIT runtime).  Use a custom linker language with the rule specified above to
-# invoke our custom linker to merge the bitcode files.
-set_target_properties(CPURuntime
-                      PROPERTIES
-                        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}
-                        CXX_STANDARD 11
-                        CXX_STANDARD_REQUIRED YES
-                        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}
-                        LINKER_LANGUAGE LLIR
-                        OUTPUT_NAME libjit.bc
-                        POSITION_INDEPENDENT_CODE YES
-                        PREFIX ""
-                        SUFFIX ""
-                        RULE_LAUNCH_COMPILE "${CLANG_BIN} <DEFINES> <INCLUDES> <FLAGS> -o <OBJECT> -c <SOURCE> #")
+set(libjit_files "libjit;libjit_conv;libjit_matmul")
 
-# Due to the usage for NaN in the runtime, '-fno-finite-math-only' is needed
-# to disable '-ffinite-math-only' activated by '-ffast-math', while benefiting
-# from math optimizations.
-target_compile_options(CPURuntime
-                       PRIVATE
-                         -ffast-math
-                         -fno-finite-math-only
-                         -g0
-                         -emit-llvm
-                         -O0)
+set(libjit_obj_file_path ${CMAKE_CURRENT_BINARY_DIR}/CPURuntime)
+file(MAKE_DIRECTORY ${libjit_obj_file_path})
+
+set(CPURuntime_OBJS)
+set(CPURuntime_SRCS)
+
+foreach(libjit_src_file ${libjit_files})
+  set(libjit_obj_file ${libjit_obj_file_path}/${libjit_src_file}${CMAKE_C_OUTPUT_EXTENSION})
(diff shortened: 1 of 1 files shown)


nadavrot:
@compnerd Can you please take a look?

lib/Backends/CPU/CMakeLists.txt
```diff
@@ -13,52 +13,58 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
...
-              libjit/libjit_conv.cpp
-              libjit/libjit_matmul.cpp)
+set(CPURunttimeCompileOptions 
+      -ffast-math
+      -fno-finite-math-only
+      -g0
+      -emit-llvm 
+      -O0)
+
+set(CPURuntime_val CPURuntime)
```
compnerd: Not sure I see the point of this variable.  Why not just inline it?

lib/Backends/CPU/CMakeLists.txt
```diff
@@ -13,52 +13,58 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
...
+  
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${CPURuntime_val})
+file(GLOB_RECURSE libjit_SOURCES LIST_DIRECTORIES false libjit/*.cpp)
```
compnerd: Please don't do the glob.  It causes issues with dependency tracking, and you don't know necessarily what files are going to get included.

lib/Backends/CPU/CMakeLists.txt
```diff
@@ -13,52 +13,58 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
...
+  set(libjit_obj_file_path ${CMAKE_CURRENT_BINARY_DIR}/${CPURuntime_val}/${libjit_obj_file})
+
+  add_custom_command(
+    OUTPUT ${libjit_obj_file_path}
+    COMMAND ${CLANG_BIN} -c ${libjit_src_file} ${CPURunttimeCompileOptions} -o ${libjit_obj_file_path}
+    DEPENDS ${libjit_src_file}
+    WORKING_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}"
+  )
+  list(APPEND CPURuntime_OBJS ${libjit_obj_file_path})
+endforeach()
```
compnerd: You should be able to use an object library here for this.  If you do that, you do not need to do the object file name computation and can use the `TARGET_OBJECTS` generator.

lib/Backends/CPU/CMakeLists.txt
```diff
@@ -13,52 +13,58 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
...
-target_compile_options(CPURuntime
-                       PRIVATE
-                         -ffast-math
-                         -fno-finite-math-only
-                         -g0
-                         -emit-llvm
-                         -O0)
+foreach(libjit_src_file ${libjit_SOURCES})
+  MATH(EXPR INCR_VAR "${INCR_VAR}+1")    
+  set(libjit_obj_file libjit_obj_file_${INCR_VAR}.o)
```
compnerd: Please don't use `.o` but rather `${CMAKE_C_OUTPUT_EXTENSION}`.

lib/Backends/CPU/CMakeLists.txt
compnerd: You can still add the sources (objects) expanding the target objects.

ayermolo: Downside of above approach is that it will always rebuild object files.

lib/Backends/CPU/CMakeLists.txt
bertmaher: Elsewhere we explicitly list the .cpp files included by a library.  As annoying as this is, as @compnerd mentions, using glob() causes issues with dependency tracking since the build system may not know when files are added.  (Some build systems like buck solve this by using a filesystem watcher - buck uses watchman - but cmake doesn't.)

ayermolo:
Updated change.
1) ${CMAKE_C_OUTPUT_EXTENSION} is now used
2) explicit use of libjit files.

ayermolo:
@compnerd  What do you think about latest revision?

lib/Backends/CPU/CMakeLists.txt
```diff
@@ -13,52 +13,57 @@ set(CMAKE_LLIR_CREATE_SHARED_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 set(CMAKE_LLIR_CREATE_SHARED_MODULE "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 set(CMAKE_LLIR_CREATE_STATIC_LIBRARY "${LLVM_LINK_BIN} -o <TARGET> <OBJECTS>")
 
-add_library(CPURuntime
-              libjit/libjit.cpp
-              libjit/libjit_conv.cpp
-              libjit/libjit_matmul.cpp)
+set(CPURunttimeCompileOptions
```
bertmaher: Please add `-std=c++11` to the compile options.

ayermolo: Updated.

ayermolo:
So just approval from @compnerd remains?

bertmaher:
I'll go ahead and merge this.  If it proves problematic, we can always revert.

PR #2167, opened 2018-12-12
Make sure we can quantize tensors to int8, int16, int32 with all schema

tlepley-cadence:
*Description*: 
Extend `chooseQuantizationParams()`, `quantize()`, `dequantize()` and `convertConstant()` functions to support any quantized type and any quantization scheme.

@nadavrot This is new step related to what we discussed in #2078 for having a more generic support in Glow for quantization data types. 

*Testing*:
Extended test `quantizeTensor` to test all combiantion: Qtype/Qscheme

*Documentation*: N/A


include/glow/Quantization/Base/Base.h
```diff
@@ -69,7 +69,7 @@ enum Schema {
   /// (i.e., offset == -128).
   /// An offset of -128 represents an unsigned int8 with an offset of zero:
   /// int8 is [-128; 127] - (-128) == uint8 [0; 255] - 0
-  SymmetricWithUInt8,
+  SymmetricWithUnsigned,
```
rdzhabarov: could you update doxygen comments for `SymmetricWithUnsigned`

include/glow/Quantization/Base/Base.h
```diff
@@ -91,9 +91,10 @@ inline DestTy quantize(float input, const TensorQuantizationParams &TQP) {
 
 /// Converts a quantized value (type eTy) to floating point based on the
 /// quantization parameters \p TQP.
+/// Note: use int64_t to cover the 'symmetric int32 with unsigned' case.
```
rdzhabarov: does int32 quantization make sense? it does not bring any memory savings and I'm not sure about perf considering extra operations needed for int32.

include/glow/Quantization/Base/Base.h
```diff
@@ -116,8 +117,9 @@ QuantizationTransform32To8 quantizeScaleOffset32To8(float scale,
 /// Calculate TensorQuantizationParams based on the clipped \p min and \p max
 /// floating point range and using the quantization method described
 /// by \p schema.
-TensorQuantizationParams chooseQuantizationParams(float min, float max,
-                                                  Schema schema = Asymmetric);
+TensorQuantizationParams
+chooseQuantizationParams(float min, float max, Schema schema = Asymmetric,
+                         ElemKind Ty = ElemKind::Int8QTy);
```
rdzhabarov: need to update comment.

include/glow/Quantization/Base/Base.h
tlepley-cadence: I think int32 or uint32 will only be fore specific cases, like scalars.
For instance already have a case for int32 with the convolution bias.  
Note that int64_t is only needed here for handling uint32, not int32.

lib/Quantization/Base/Base.cpp
```diff
@@ -309,10 +328,10 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
...
 
-  // The only valid offsets for symmetric quantization with uint8 support are 0
-  // and -128.
-  assert((result.offset == -128 || result.offset == 0 ||
-          schema != quantization::Schema::SymmetricWithUInt8) &&
+  // The only valid offsets for symmetric quantization with unsigned support are
+  // 0 and qmin.
+  assert((result.offset == qmin || result.offset == 0 ||
+          schema != quantization::Schema::SymmetricWithUnsigned) &&
          "Symmetric quantization should be centered on 0");
```
rdzhabarov: assert message needs to be fixed, it's not aligned to what we actually doing. (comment is correct)

include/glow/Quantization/Base/Base.h
tlepley-cadence: Done.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1861,8 +1861,9 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
       return NodeValue(&constantToBeModified, 0);
     }
     case ElemKind::Int32QTy:
+    case ElemKind::Int16QTy:
```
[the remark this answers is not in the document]
tlepley-cadence: Done.

lib/Quantization/Base/Base.cpp
tlepley-cadence: Done.

include/glow/Quantization/Base/Base.h
tlepley-cadence: Done.

PR #2171, opened 2018-12-14
[Instructions] Add memory-related verification for operands

jfix71:
*Description*: It seems that previously there was no verification to make sure Instructions are correctly using memory instructions. This adds more verification for these.
*Testing*: Added a couple tests.

### DIFF
--- a/lib/IR/IR.cpp
+++ b/lib/IR/IR.cpp
@@ -115,6 +115,17 @@ void Instruction::verifyUseList(
 }
 
 void Instruction::verify() const {
+  // All operands of instructions must be either memory (AllocActivation or
+  // WeightVar), or a TensorView of such.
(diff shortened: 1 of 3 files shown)


qcolombet:
Note: The title of the PR is truncated ;)

tools/ClassGen/InstrBuilder.cpp
```diff
@@ -214,6 +214,21 @@ void InstrBuilder::emitClass(std::ostream &os) const {
     os << "  void verify() const;\n";
   } else {
     os << "  void verify() const {\n";
+
+    // All non-memory instructions should have the property that all operands
```
nadavrot: What do you mean by non-memory instructions? Do quantization-profile instructions fall into this category?

PR #2180, opened 2018-12-17
Use isOpSupported() correctly for Interpreter with fp16

jfix71:
*Description*: Add all the nodes we support for fp16 to the Interpreter's `isOpSupported()`. We were previously returning false for all fp16 nodes. This is clearly incorrect; we should use the `isOpSupported()` API correctly. The FunctionQuantizer was already doing this. For this I added the EE to the FunctionConverter and removed it from FunctionQuantizer.

*Testing*: All tests still pass. Also checked some networks (e.g. Resnet50, shufflenet, etc.) to make sure the graph still appears the same.

### DIFF
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -174,7 +174,7 @@ bool CPUBackend::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
     }
   }
 
-  return true;
+  return elementTy == ElemKind::FloatTy;
 }
 
 bool CPUBackend::shouldLower(const Node *N) const {
--- a/include/glow/Converter/FunctionConverter.h
+++ b/include/glow/Converter/FunctionConverter.h
@@ -19,6 +19,7 @@
 #define GLOW_CONVERTER_FUNCTIONCONVERTER_H
 
 #include "glow/Base/Type.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
 
 #include <utility> // For std::pair.
 
@@ -37,6 +38,9 @@ class FunctionConverter {
 protected:
   /// The function to be converted.
   Function &function_;
+
+  /// Execution engine used to check if an operator is supported.
+  const ExecutionEngine &EE_;
 
   /// \return the type that \p out needs to have at the end of the conversion
   /// procedure. In other words, this is the type this value will have at the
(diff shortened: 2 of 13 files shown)


jfix71:
I also updated the CPU and OpenCL backends' `isOpSupported()`, and accordingly `typeAToTypeBFunctionConverterTest.cpp`. I don't think the way it was before made sense (the backend made no difference on the test so it was essentially just duplicating the test). I left one test on all the backends, where only the Placeholders are converted to fp16. The rest of the tests are for now only on the Interpreter, as we don't support fp16 on the OpenCL or CPU backends.

jfix71:
I'm having second thoughts on adding `isOpSupported()` to each FunctionConverter's `canConvert()`. Perhaps it/the EE should actually be removed entirely from all FunctionConverters, and we should iterate over the graph after conversion/before compilation to check if the converted graph with the desired precision is supported by the backend via `isOpSupported()`. I think this better fits where we're going with the Partitioner, where we would check the backend's `isOpSupported()` to determine how to partition the graph for what is supported on what backends.

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -174,7 +174,7 @@ bool CPUBackend::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
     }
   }
 
-  return true;
+  return elementTy == ElemKind::FloatTy;
```
beicy: This means CPUBackend only support either float or int8(quantized type) right? But
CPU backend support int32 quantization: `int32_t libjit_element_quantize_kernel_i32`, I think  `isOpSupported(Kinded::Kind::QuantizeNodeKind, ElemKind::Int32QTy)` should be true... Or this call is illegal?

jfix71: Ah, good catch -- yes, `isOpSupported(Kinded::Kind::QuantizeNodeKind, ElemKind::Int32QTy)` should return true.

rdzhabarov: strictly speaking not all fp32 operations are always implemented even if we have glow instr in place. Same goes to the interpreter.

bertmaher:
@jfix71 Are you still interested in landing this PR?  I can review it if so :-)

jfix71:
Yeah this one is obsolete, closing!

PR #2182, opened 2018-12-17
Remove deprecated partitioning approach

bertmaher:
*Description*: We're taking a fresh look at this problem, and this approach never got used outside of a unit test, so let's just drop the baggage.
*Testing*: n/a
*Documentation*: n/a


bertmaher:
Great, glad this code was useful while it lasted :-)

PR #2196, opened 2018-12-18
Fix Onnxifi lib build dependencies

nickgg:
*Description*: Onnxifi/Base.cpp uses glow::lower which needs `Optimizer` as a dependency. I'm not sure why this worked in the CI system.
*Testing*: ninja all in release.
*Documentation*: n/a


jackm321:
Cool thanks for fixing it

PR #2198, opened 2018-12-18
[WIP][ONNXIFI][Runtime] Create custom error types

jackm321:
*Description*:
This PR creates two custom error types for glow, a `ModelLoadingErr` and `RuntimeErr`. `ModelLoadingErr` can be used in model loading code to provide helpful error message when model loading fails as well as provide an error code that can be used to return an appropriate ONNXIFI status.
`RuntimeErr` only contains an error code and it too can also be used to return appropriate ONNXIFI statuses in the event that device or host memory is all consumed for example.

I think `ModelLoadingErr` is basically necessary however I'm not as convinced about creating a `RuntimeErr` as opposed to just using an enum to signify statuses in the runtime. The advantages of using an `RuntimeErr` over an enum are 

1. Guaranteed to be checked.
2. Unified error propagation throughout ONNXIFI and runtime, all code in the runtime and model loading just deals with llvm::Error (no matter what the underlying error type is) and it can either handle errors by type or propagate on anything that isn't `Error::Success` no matter what the type of error is, no need to check specific enum values. For example, if function `A` calls function `B` and function `B` returns an llvm::Error (be it a `ModelLoadingErr`, `RuntimeErr`, or some other error), function `A` can then just return the Error up the stack to the handlers at the ONNXIFI layer without worrying about what specifically went wrong).
3. Can use llvm::Expected syntax.
4. Can add more error information such as a string to explain the situation if necessary.

The cons of using `RuntimeErr` include more complexity than using an enum and higher cost of handling errors when they occur (need to allocate a  `RuntimeErr` if an error occurs).

Thoughts on this?

*Testing*:
n/a
*Documentation*:
not yet

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -45,6 +45,152 @@ template <typename> struct IsLLVMExpected : public std::false_type {};
 template <typename T>
 struct IsLLVMExpected<llvm::Expected<T>> : public std::true_type {};
 
+template <typename ThisErrT> class BaseErr : public llvm::ErrorInfoBase {
+public:
+  static const void *classID() { return &ThisErrT::ID; }
+
+  const void *dynamicClassID() const override { return &ThisErrT::ID; }
+
+  bool isA(const void *const ClassID) const override {
+    return ClassID == classID() || llvm::ErrorInfoBase::isA(ClassID);
+  }
+
+  virtual std::error_code convertToErrorCode() const override {
+    return llvm::inconvertibleErrorCode();
+  }
+
+  void log(llvm::raw_ostream &OS) const override {
+    OS << "file: " << fileName_ << " line: " << lineNumber_ << " error: \"";
+    static_cast<const ThisErrT *>(this)->logErr(OS);
+    OS << "\"";
+  }
+
+protected:
+  BaseErr(llvm::StringRef fileName, int32_t lineNumber) {
+    fileName_ = fileName;
+    lineNumber_ = lineNumber;
+  }
+
+private:
+  int32_t lineNumber_;
+  std::string fileName_;
+};
+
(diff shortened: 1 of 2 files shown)


jackm321:
cc @gcatron @beicy @nickgg

jackm321:
Also, this is a super rough cut, I mostly just wanted to get some thoughts on llvm::Errors vs an enum for the runtime

include/glow/Support/Error.h
```diff
@@ -45,6 +45,152 @@ template <typename> struct IsLLVMExpected : public std::false_type {};
...
+  static llvm::StringRef ecToString(const EC ec) {
+    switch (ec) {
+    case UNKNOWN:
+      return "UNKNOWN";
+    case UNSUPPORTED_SHAPE:
+      return "UNSUPPORTED_SHAPE";
+    case UNSUPPORTED_OPERATOR:
+      return "UNSUPPORTED_OPERATOR";
+    case UNSUPPORTED_ATTRIBUTE:
+      return "UNSUPPORTED_ATTRIBUTE";
```
rdzhabarov: does this work correctly?

in case if you have

const std::string& ret( ... ) {
   ...
   return "something";
}

this would return ref to the temp object.
StringRef might have similar problem

jackm321: Oh good point, I would not expect that this works actually but I could wrong. I'll change it

bertmaher: If I understand StringRef correctly, this pattern should be safe, because StringRef simply wraps the (statically allocated) const char* and length, without any stack/heap allocation: https://godbolt.org/z/jQT1j8

jackm321:
Going to clean this up and publish a more polished iteration of this with some use in model loading.

PR #2203, opened 2018-12-19
[onnxifi] Refactor onnxGetBackendIDs to make it easier to add optional backends

bertmaher:
*Description*: With this style, you should be able to add an arbitrary number of #ifdef blocks and have them work.
*Testing*: test_onnxifi.py from the pytorch repo.
*Documentation*: n/a


bertmaher:
Sure, I can do that.  Why does the existing code create an array though?

jackm321:
To run on different backends (cpu, interpreter, devices) different subclasses `DeviceManager` will be instantiated and registered with `HostManager` I believe

yinghai:
> Pytorch does not care about any ONNXIFI backend other than at the position [0].

We do care about other position. I think for compatibility, having an array makes sense.

PR #2204, opened 2018-12-20
[docs] Document how to build and use PyTorch with ONNXIFI

bertmaher:
*Description*: I've got this working entirely with open-source repos, so I thought the rest of the world would like to know how it's done.

*Testing*: Follow the doc!

*Documentation*: New doc: Onnxifi.md

### DIFF
--- a/docs/Onnxifi.md
+++ b/docs/Onnxifi.md
@@ -14,7 +14,7 @@ build, but I don't remember all of them.)**
 While not required, we recommend using a Python virtual environment, to avoid
 quirks associated with the system Python environment.
 
-```
+```bash
 cd $SRC
 mkdir virtualenv && cd virtualenv && virtualenv . && source bin/activate
 pip install pyyaml pytest future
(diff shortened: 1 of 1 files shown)


docs/Onnxifi.md
```diff
@@ -0,0 +1,73 @@
...
+```
+cd $SRC
+mkdir virtualenv && cd virtualenv && virtualenv . && source bin/activate
+pip install pyyaml pytest future
+```
+
+Get PyTorch, ONNX and Glow.
+
+```
+cd $SRC
```
rdzhabarov: if you say " ``` bash " it'll make some nice highlighting

docs/Onnxifi.md
```diff
@@ -0,0 +1,73 @@
...
+Build and install PyTorch.
+
+```
+cd $SRC/pytorch
+git submodule update --init --recursive
+python setup.py install
+```
+
+Remove the dummy libonnxifi.so that PyTorch installs. **(TODO: Needing to do
+this is weird.  Can we avoid it?)**
```
rdzhabarov: @yinghai where is it (libonnxifi) coming from?

yinghai: It's probably coming from onnx dummy backend. We shouldn't install that.

yinghai: @zrphercule Could you check?

PR #2205, opened 2018-12-20
[quantization] More flexible general quantization support (int8/int16)

rdzhabarov:
*Description*:
* Add macros for quantized ops and quantized ops with accumulation (conv, gemm, etc)
* Int16 support for conv, gemm, rescale quantized, max pool
* Relaxing verifiers to support wider range for quantization formats (including int16) for certain ops

More generic quantization handling to int8/int16, etc.

*Testing*:
* Existing tests should cover base scenarios of int8.
* Will add couple of tests for int16 (although), handling of int16 is not much different except wider accumulators.

*Documentation*:
n/a

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -300,8 +300,16 @@ public:
              "Offsets must match.");
       return isEqualImpl<int8_t>(other, allowedError);
     case ElemKind::Int16QTy:
+      assert(getType().getScale() == other.getType().getScale() &&
+             "Scales must match.");
+      assert(getType().getOffset() == other.getType().getOffset() &&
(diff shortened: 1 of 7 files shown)


include/glow/Base/Tensor.h
```diff
@@ -300,8 +300,16 @@ class Tensor final {
              "Offsets must match.");
       return isEqualImpl<int8_t>(other, allowedError);
     case ElemKind::Int16QTy:
+      assert(getType().getScale() == other.getType().getScale() &&
```
rdzhabarov: Comparison needs to be fixed properly for all quantized types, see https://github.com/pytorch/glow/issues/2206

Issue #2194, opened 2018-12-18
[RFC] isOpSupported() redesign?


jfix71:
I'm not sure the `isOpSupported()` API makes sense as currently specified -- I think backends need to decide if an op is supported based on the expected precision of all inputs and outputs. Perhaps the API should take in the Node itself and not the Kind+ElemKind. That way the backend can look at all of the node's inputs/outputs.

Additionally, for backends that use the Instruction-level IR, I think we should consider adding a pass over the final Instructions with another version of `isOpSupported()` for Instructions. It's possible for a bug to exist during the compilation phases (lowering to other Nodes, and then lowering via IRGen, and many graph/instruction optimization passes) that results in the final Instruction list not being supported by the backend. This should be caught as early as possible during compilation and return a compilation error up the stack.

jackm321:
Definitely agree with the first point, the current API doesn't fully make sense to me because I doubt that op compatibility can be decided from a single ElemKind in all cases. The ONNXIFI interface currently just passes `FloatTy` to `isOpSupported()` instead of inspecting each ElemKind which isn't great. Like you said, maybe the op should be passed, maybe just a set of all `ElemKind`s in the op could be passed and `isOpSupported()` could just check to see that it supports all of them?

To your second point, it sound like a good idea to catch issues that could occur during compilation but this raises a point that it will be pretty difficult to look at a graph (like one passed from onnxifi) and decide if the backend will support all ops in the graph without actually compiling to low level IR and checking with this new `isOpSupported()`. It would be ideal if we could do enough checks on the high level IR to be confident that the low level IR will run on the backend without issue but that might not be possible.

bertmaher:
An additional wrinkle here is that we also use this API to determine if a node can be quantized on this backend, so we might not have a physical node to pass in.  Maybe we actually need two distinct APIs here; and `isOpSupported`, and a `canBeQuantized`.  Of course that doesn't necessarily cover the (pretty common) case of a node that _must_ be quantized, like GEMMs compiled for an int8 matrix multiplier.

rdzhabarov:
Totally, original idea of `isOpSupported` was to invoke that during the quantization phase and make that slightly more generic rather than just canBeQuantized.
But it seems that it's not generic enough to cover all use cases, including `compatibility` mode for ONNXIFI.

I agree we should introduce `isOpSupported` with the node passed in which will be used in the ONNXIFI code path and `canBeQuantized` with the specified quantization precision needed for graph transformation/quantization.

[...]

PR #2207, opened 2018-12-20
[NodeBuilder] Add Input and Output Indices enums

jfix71:
As mentioned in https://github.com/pytorch/glow/issues/2194#issuecomment-449008434, there are places where we use indices explicitly to specify inputs or outputs of Nodes. It's a really bad practice as these indices could change if the Node definition or NodeBuilder change. This PR improves the situation by adding input and output indices enums to each node.

There's a commit here with a couple examples of places it can be used to clean things up. There are other places to update if we agree this approach makes sense. It could also be useful for use in the new design of `isOpSupported()` (https://github.com/pytorch/glow/issues/2194).

Here's what the enums look like for the `ConvolutionNode` in `AutoGenNodes.h` with this PR:

```cpp
class ConvolutionNode final : public Node {
  ...
  enum InputIndices {
    InputIdx = 0,
    FilterIdx = 1,
    BiasIdx = 2,
  };

  enum ResultIndices {
    ResultIdx = 0,
  };
```

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -433,7 +433,7 @@ public:
   /// empty, \p input is split to equal sized parts.
   void createSplit(llvm::StringRef name, NodeValue input, unsigned_t outputNum,
                    unsigned_t axis, llvm::ArrayRef<size_t> split,
-                   std::vector<Node *> &outputs);
+                   std::vector<SliceNode *> &outputs);
 
   BatchNormalizationNode *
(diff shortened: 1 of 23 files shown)


nadavrot:
Nice!

jfix71:
I removed all explicit indices I could find, mostly just replacing them with enums. These changes are basically all based around uses of `setType()`/`getType()`, `getElementType()`, `setNthResult()`/`getNthResult()`, and `setNthInput()`/`getNthInput()`.

PR #2209, opened 2018-12-20
Revert "[onnxifi] Refactor onnxGetBackendIDs to make it easier to add optional backends"

bertmaher:
Reverts pytorch/glow#2203
@yinghai depends on the backend array for some work in progress.


bertmaher:
Yeah, I'd originally done a bit of refactoring to make adding to the array easier, but there's no hurry on that.  I may resurrect it in a bit.

PR #2213, opened 2018-12-21
Refactor ONNXFIModelLoader

yinghai:
*Description*:
This PR refactors the ONNXIFIModelLoader to let it not inherit ONNXModelLoader. Instead, it uses it. This can separate the dependency of ONNXIFI on ONNX and allows us to load other models through ONNXIFI interface. 

*Testing*:
Unit tests

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -17,6 +17,8 @@
 #ifndef GLOW_IMPORTER_COMMONOPERATORLOADER_H
 #define GLOW_IMPORTER_COMMONOPERATORLOADER_H
 
+#include "onnx/onnxifi.h"
+
 #include "glow/Importer/ProtobufLoader.h"
 
 #include "glow/Base/Tensor.h"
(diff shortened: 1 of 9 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -17,6 +17,8 @@
 #ifndef GLOW_IMPORTER_COMMONOPERATORLOADER_H
 #define GLOW_IMPORTER_COMMONOPERATORLOADER_H
 
+#include "onnx/onnxifi.h"
```
bertmaher: It seems like CommonOperatorLoader shouldn't need to know about ONNX or ONNXIFI.  Is there a way to avoid that?

include/glow/Importer/ONNXIFIModelLoader.h
```diff
@@ -25,22 +25,14 @@
...
 namespace glow {
 
-class ONNXIFIModelLoader : public ONNXModelLoader {
+class ONNXIFIModelLoader {
 private:
   /// If \p errPtr is not null then if an error occurs it will get assigned
   /// there otherwise if an error occurs it will abort.
-  ONNXIFIModelLoader(Function &F, llvm::Error *errPtr = nullptr)
-      : ONNXModelLoader(F, errPtr) {}
+  ONNXIFIModelLoader(llvm::Error *errPtr = nullptr) {}
```
bertmaher: Make this `explicit`

include/glow/Importer/CommonOperatorLoader.h
yinghai: Hmm, one thing that is going to be in common is the ability to load weights from onnxTensorDescriptor. For that we need onnxifi.h.

bertmaher: Got it.  That seems fine then.

lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -20,147 +20,46 @@
...
-      new ONNXIFIModelLoader(F, &loaderConstructionErr));
+      new ONNXIFIModelLoader(&loaderConstructionErr));
   if (loaderConstructionErr) {
     return std::move(loaderConstructionErr);
   }
 
-  ONNX_NAMESPACE::ModelProto modelDef;
-  ASSIGN_VALUE_OR_RETURN_ERR(modelDef,
-                             loader->loadProto(onnxModel, onnxModelSize));
+  if (use_onnx) {
```
bertmaher: So right now if !use_onnx, this method kinda does nothing, right?  Should we assert(use_onnx) for the time being (just until the c2 implementation is in place)?

yinghai: C2 implementation is coming tomorrow, or tonight. :)

yinghai:
But yeah, I can add check here for now.

PR #2215, opened 2018-12-21
[importer] Handle GivenTensorFill like Caffe2

bertmaher:
*Description*: Caffe2 does not accept `ints` with `GivenTensorFill`, only `floats`.  Trying to use `ints` produces a check failure:
```
Check failed: output->numel() == values_.numel() output size: 314 given size: 0
```
I'd like Glow to match C2 semantics so we don't find ourselves executing something that C2 rejects.

There is apparently a `dtype` argument that would allow this, though I haven't added support for it.

*Testing*: Update existing test.

*Documentation*:

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -822,70 +822,67 @@ llvm::Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
   return llvm::Error::success();
 }
 
+template <typename ElemTy, typename RangeTy>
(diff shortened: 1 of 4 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -822,70 +822,67 @@ llvm::Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
   return llvm::Error::success();
 }
 
+template <typename ElemTy, typename RangeTy>
```
nadavrot: Doxygen comment please.

PR #2216, opened 2018-12-21
Replace lit with script

jackm321:
*Description*:
Replaces lit with a cheap knockoff script flit.py. Also replaces FileCheck with filechecker.py. It'll be easier to port a couple of python scripts than llvm testing infrastructure.

*Testing*:
Run existing lit tests. Created some dummy tests that are expected to PASS, SKIP, or FAIL and ensure that they do that.

*Documentation*:
Inline comments. Will also update the wiki.


tests/flit_cfg.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
nadavrot: I am not sure why we need this file. What features are implemented with this logic? Could we have a simpler mechanism?

tests/flit.py
```diff
@@ -0,0 +1,168 @@
...
+    config = Config()
+    numReceivedCLArgs = len(sys.argv)
+    for i in range(1, numReceivedCLArgs):
+        config_path = os.path.abspath(sys.argv[i])
+        config_source = imp.load_source('config_source', config_path)
+        import config_source
+        config_source.setup_config(config)
+    return config
+
+## Builds a Config from config files specfied in command line args then runs
```
nadavrot: Why do we need to support config files?

tests/flit_cfg.py
```diff
@@ -0,0 +1,69 @@
...
+    # name: The name of this test suite.
+    config.name = 'glow'
+
+    # suffixes: A list of file extensions to treat as test files.
+    config.suffixes = [".test"]
+
+    # Set the different string substitutions that we use in our test suite.
+    config.tool_to_path = {}
+
+    for tool_name in ('text-translator', 'image-classifier', 'model-runner'):
```
nadavrot: I am not sure why we need this logic. Can we just pick up the tools from the path?  Listing all of the tools manually just makes it difficult to add new tools. Can we get rid of this file?

tests/flit_cfg.py
```diff
@@ -0,0 +1,69 @@
...
+    if config.has_opencl:
+        config.available_features.add('opencl')
+    if config.has_cpu:
+        config.available_features.add('cpu')
+    if config.models_dir:
+        config.substitutions.insert(0, ('%models_dir', config.models_dir))
+        if os.path.isdir(config.models_dir):
+            config.available_features.add('models')
+
+    config.substitutions.insert(0, ('%FileCheck', config.filechecker_command))
```
nadavrot: I don't think that we need to follow the FileCheck convention. Why can't we have one tool that runs the RUN command and also validates the output with the CHECK lines?  We don't need to worry about temp files, piping, etc.

tests/filechecker.py
```diff
@@ -0,0 +1,103 @@
+import sys
+import os
```
nadavrot: I am not sure why this is an external tool. The script that runs and checks can be the same file. This will save a bunch of logic and simplify things.

tests/text-translator/en2gr-quantization.test
```diff
@@ -1,4 +1,4 @@
-RUN: %text-translator -m %models_dir/en2gr -keep-original-precision-for-nodes=Add -dump-profile=%t.yaml -quantization-schema=symmetric <<< "My name is Bob ." && \
+RUN: %text-translator -m %models_dir/en2gr -keep-original-precision-for-nodes=Add -dump-profile=%t.yaml -quantization-schema=symmetric <<< "My name is Bob ."
 RUN: (%text-translator -m %models_dir/en2gr -keep-original-precision-for-nodes=Add -load-profile=%t.yaml <<< "My name is Bob ." | %FileCheck %s)
```
nadavrot: I am not sure why we need to pipe stuff into FileCheck. Isn't this obvious that we need to verify the output and compare it to the CHECK lines in this file?

tests/flit_cfg.py
jackm321: Yeah that is actually what I started out with but I changed to the FileCheck convention in case we wanted to use FileCheck type utility for other testing later. I think it'll just be more simple to have one tool though so I'll switch back.

jackm321:
@nadavrot thanks for the review, I agree that the single script would be better as well, I'll switch back to that.

qcolombet:
I am going to repeat what I said in the PR that brought lit to glow: I still don’t understand why we want to reinvent the wheel with a proprietary script.
Regarding the portability side, lit is available as a standard python package and for FileCheck, it is available in apt packages and we already depend on llvm...

My 2c.

PR #2217, opened 2018-12-21
Add a Taggedlist and use it for InstrListTy in IRFunction

### LINKED ISSUE #1766: Replace llvm::iplist<Instruction> with TaggedList<Instruction> in IRFunction
I want to introduce a new linked list data structure, `TaggedList`. This is an intrusive doubly linked list like `llvm::iplist` with the following differences:

* Each node has a `uint32_t tag` member. These tags are maintained online in a strictly increasing order. This means that the relative position of elements in the list can be determined in (fast) constant time, even while inserting and removing elements from the list.
* The list has a constant time `size()` operation, maintained as a member in the `TaggedList` class. This is needed for the online tag maintenance algorithm.
* Splicing is a linear time operation because the new elements need to be tagged correctly. This is already the case for our specialization of `transferNodesFromList()`.
* The tag also makes it possible to detect when an `end()` iterator is dereferenced or incremented.

I have implemented a prototype of `TaggedList` based on the [Two simplified algorithms for maintaining order in a list](https://scholar.google.com/scholar?cluster=5225046542682967685&hl=en&as_sdt=0,5) paper in order to run benchmarks. List insertions are O(log N) with excellent constant factors.

We already have `class LiveIntervalsInstructionNumbering` in `IROptimizer.cpp`. This change makes that class redundant by providing the same functionality while also allowing online instruction insertions.

The online instruction ordering is needed to implement advanced memory allocation and scheduling in backends.

Preemptively Answered Questions:

* No, we can't extend `llvm::iplist` with this functionality because a) `iplist::size()` is linear time, and b) despite using 27 template classes for configuration and specialization, `iplist` doesn't offer the hooks needed.
* No, we can't implement this as a hack in `IRFunction` because a) we need the constant time `size()` operator, and b) `IRFunction::getInstrs()` exposes the list implementation, and we can't prevent users from modifying the list and invalidating our side tables.

stoklund:
*Description*:
A TaggedList is an intrusive linked list similar to `llvm::ilist`, except a
TaggedList maintains a node ordering so it is possible to compare relative
positions of two nodes in constant time.

*Testing*:
`ninja test`

*Documentation*:
Doxygen comments and unit tests.

Fixes #1766.

### DIFF
--- a/include/glow/Base/TaggedList.h
+++ b/include/glow/Base/TaggedList.h
@@ -1,0 +1,399 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BASE_TAGGEDLIST_H
+#define GLOW_BASE_TAGGEDLIST_H
+
+/// \file Intrusive doubly linked list with online node ordering.
+///
+/// This file implements a TaggedList<T> template which is very similar to
+/// llvm::ilist<T> with the addition of online node ordering. This means that
+/// iterators into a TaggedList<T> can be compared with the standard inequality
+/// operators (<, <=, >, >=), and the iterator relations are consistent with the
+/// list order.
+
+#include <cassert>
+#include <cinttypes>
+#include <cstddef>
+#include <iterator>
+#include <type_traits>
+
+namespace glow {
+
+template <typename T> class TaggedListNode;
+template <typename T, typename Traits> class TaggedList;
+
+/// Traits for a TaggedList<T>
+///
+/// Specialize these traits to customize list behavior.
+///
+/// The TaggedList will derive from the traits type.
+template <typename T> struct TaggedListTraits {
+  /// Delete a node.
+  ///
+  /// A TaggedList<T> never allocates nodes, but the erase() and clear() methods
+  /// will delete nodes.
+  static void deleteNode(T *node) { delete node; }
+
+  /// Called before node is added to this list.
+  void addNodeToList(T *) {}
+
+  /// Called before node is removed from this list.
+  void removeNodeFromList(T *) {}
+};
+
+/// Namespace of TaggedList implementation details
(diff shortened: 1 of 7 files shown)


include/glow/Base/TaggedList.h
```diff
@@ -0,0 +1,398 @@
...
+    this->addNodeToList(node);
+    impl_.insert(next.node_, node);
+    return iterator(node);
+  }
+
+  /// Remove `node` from this list without deleting it.
+  ///
+  /// Returns a pointer to the removed node.
+  T *remove(T *node) {
+    assert(node && node->inTaggedList());
```
opti-mix: Should it also check that the node belongs to the current list?

lib/Base/TaggedList.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // lo..hi.
+  uint32_t dy = tagCeil - tagFloor;
+  uint32_t dx = count - 1;
+
+  // Make an even larger hold at the insertion point than is strictly
+  // necessary. This anticipates future insertions near the same point.
+  dx += std::min(dx / 8, dy - dx);
+  uint32_t step = dy / dx;
+
+  for (; lo != insertionPoint; lo = lo->nextTaggedNode_, tagFloor += step)
```
opti-mix: Nit: We usually use `{}` in the code-base even for single-line line loop bodies or if-bodies.

opti-mix:
@stoklund Very nice job! 

It is impressive that changing the existing code to use your new `Taggedlist` required just a few changes. I expected more re-factoring ;-)

opti-mix:
@stoklund Have you also tested on bigger NN models using our `run.sh`? Since this change is so intrusive, we should get as much testing as possible.

nadavrot:
The code looks really good. It's clean, documented, simple and tested.

include/glow/Base/TaggedList.h
```diff
@@ -0,0 +1,398 @@
...
+
+  /// Post-decrement.
+  Iterator operator--(int) {
+    Iterator tmp = *this;
+    --*this;
+    return tmp;
+  }
+
+private:
+  // This is either pointing ot a list element of type T or the sentinel
```
bertmaher: typo: "pointing to"

include/glow/Base/TaggedList.h
stoklund: That would be nice, but the way the abstraction works, a list node doesn't know which list it belongs to.

We have parent pointers in the `Instruction` nodes, but they are not available to the list implementation itself. See the hacks around `getContainingFunction()`.

lib/Base/TaggedList.cpp
stoklund: Thanks, fixed.

stoklund:
> @stoklund Have you also tested on bigger NN models using our `run.sh`? Since this change is so intrusive, we should get as much testing as possible.

Well, it is an intrusive list! ;-)

I've tested the renumbering algorithm with lists up to 256M nodes in length. I'll run it through some real models too.

PR #2219, opened 2018-12-22
Enable ONNXIFI to import Caffe2 model proto

yinghai:
*Description*:
This PR enables ONNXIFI model loader to load a caffe2 model proto string. The function names and stuff can be more polished. And I think c2 and onnx loader can be more unified in terms of interface. But not sure I want to do it in this PR. 

*Testing*:
Internal integration test

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Importer/ONNXIFIModelLoader.cpp
+++ b/lib/Importer/ONNXIFIModelLoader.cpp
@@ -15,13 +15,15 @@
  */
 
 #include "glow/Importer/ONNXIFIModelLoader.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
 
+#include "caffe2/proto/caffe2.pb.h"
 #include "onnx/onnx_pb.h"
 
 namespace glow {
 
 llvm::Expected<std::unique_ptr<ONNXIFIModelLoader>> ONNXIFIModelLoader::parse(
-    const void *onnxModel, uint32_t onnxModelSize, uint32_t weightsCount,
+    const void *model, uint32_t modelSize, uint32_t weightsCount,
     const onnxTensorDescriptorV1 *weightDescriptors, Function &F,
     bool loadInputsAsPlaceholders, bool use_onnx) {
   llvm::Error loaderConstructionErr = llvm::Error::success();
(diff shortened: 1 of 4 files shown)


lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -59,6 +61,30 @@ llvm::Expected<std::unique_ptr<ONNXIFIModelLoader>> ONNXIFIModelLoader::parse(
 
     // Keep hold of the context
     loader->core_ = std::move(onnxLoader);
+  } else {
+    // Use Caffe2 Model loader
```
jackm321: This is starting to seem a bit tedious and specific to each model type even though the underlying pattern I think is pretty consistent across model types being loaded. It would be better if possible to move the logic of loading each model type (stuff like calling `loadProto` `loadInputs`, `loadInitializers`, etc) to a `load()` function or something like that in `ONNXModelLoader` and `Caffe2ModelLoader` so we can just call `loader->core_.load(model, modelSize, weightsCount, ...)`. Maybe it could even be in the constructor of those underlying loaders those seem to already have a lot of the logic to do this. Is there anything preventing using those here instead?

yinghai: Yeah, I've mentioned that in the summary. We probably want to do it in a refacotring PR.

lib/Importer/ONNXIFIModelLoader.cpp
```diff
@@ -59,6 +61,30 @@ llvm::Expected<std::unique_ptr<ONNXIFIModelLoader>> ONNXIFIModelLoader::parse(
...
+        new Caffe2ModelLoader(F, &loaderConstructionErr));
+    if (loaderConstructionErr) {
+      return std::move(loaderConstructionErr);
+    }
+
+    caffe2::NetDef networkDef;
+    ASSIGN_VALUE_OR_RETURN_ERR(networkDef,
+                               c2Loader->loadProto(model, modelSize));
+
+    RETURN_IF_ERR(c2Loader->loadWeights(weightsCount, weightDescriptors));
```
jackm321: I'm confused, shouldn't this be `c2Loader->loadWeightsFromNet` why does this compile?

yinghai: `loadWeightsFromNet` is a different function where we load the weights from the init net. Here we load weights from the TensorDescriptor.

yinghai:
Filed https://github.com/pytorch/glow/issues/2231 to track the clean up progress.

PR #2222, opened 2018-12-27
[CMake] add target AutoGen for explicit dependency check

ffk0716:
*Description*: 
If target does not link with Graph/IR.
It may failed with AutoGenXXX header missing during build with -j

AutoGen target is for explicit dependency check with target which using AutoGenXXX headers

*Testing*: no test
*Documentation*: no doc updated
[Optional Fixes #issue]

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,3 +1,5 @@
+add_custom_target(AutoGen)
+
 add_subdirectory(Backends)
 add_subdirectory(Base)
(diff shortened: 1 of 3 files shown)


ffk0716:
Currently all target using AutoGenXXX headers (directly/indirectly) are linked with Graph.
Therefore no target failed currently.

But I am porting a new backend in glow.
Some target use AutoGenXXX headers only, the original cmake cause problems.

PR #2227, opened 2019-01-02
Print stack trace on error signal

bertmaher:
*Description*: Enabling this signal handler makes it quicker to debug a crash since you get a stack trace without needing to run under a debugger.  (Unlike a debugger, this option doesn't seem to give line numbers, which is weird.  But still pretty useful.)
*Testing*: Add an error to one of the unittests and observe that a stack trace is printed.
*Documentation*: n/a

### DIFF
--- a/tests/unittests/TestMain.cpp
+++ b/tests/unittests/TestMain.cpp
@@ -15,10 +15,12 @@
  */
 
 #include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Signals.h"
 #include "gtest/gtest.h"
 
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
+  llvm::sys::PrintStackTraceOnErrorSignal(argv[0]);
   llvm::cl::ParseCommandLineOptions(argc, argv);
   return RUN_ALL_TESTS();
 }
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -26,6 +26,7 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/Path.h"
+#include "llvm/Support/Signals.h"
 #include "llvm/Support/Timer.h"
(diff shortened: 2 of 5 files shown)


tests/unittests/TestMain.cpp
```diff
@@ -15,10 +15,12 @@
  */
 
 #include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Signals.h"
 #include "gtest/gtest.h"
 
 int main(int argc, char **argv) {
   ::testing::InitGoogleTest(&argc, argv);
+  llvm::sys::PrintStackTraceOnErrorSignal(llvm::StringRef());
```
opti-mix: Should it pass `argv[0]` as the argument, because it is the name of the current binary?
See e.g. https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/ClangTidyMain.cpp for an example of calling `PrintStackTraceOnErrorSignal`

tools/loader/Loader.cpp
```diff
@@ -323,6 +324,7 @@ void Loader::generateAndSerializeQuantizationInfos(Context &ctx) {
 }
 
 Loader::Loader(int argc, char **argv) {
+  llvm::sys::PrintStackTraceOnErrorSignal(llvm::StringRef());
```
opti-mix: Same here.

tests/unittests/TestMain.cpp
bertmaher: I dunno.  The docs say it's used to find the symbolizer relative to the current binary (http://llvm.org/doxygen/namespacellvm_1_1sys.html#ab062fd190912d9ca714311df7cbe51d5), which seems useful for llvm tools but maybe not for Glow?  I guess it can't hurt.

bertmaher:
wtf, does PrintStackTraceOnErrorSignal leak memory??  Adding some debugging stuff, WIP until then.

bertmaher:
Apparently you have to have a symbolizer set up for suppressions to work, so I'm just going to roll #2229 into this PR.

PR #2228, opened 2019-01-02
[FIX] Fix linking against protobuff.

ayermolo:
*Description*:
Fix linking against protobuf built as dll.
*Testing*:
Unit tests.
*Documentation*:
NA
[Optional Fixes #issue]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ option(GLOW_WITH_OPENCL "Build the OpenCL backend" ON)
 option(GLOW_BUILD_EXAMPLES "Build the examples" ON)
 option(GLOW_BUILD_TESTS "Build the tests" ON)
 option(GLOW_WITH_BUNDLES "Build bundles" OFF)
+option(LINK_PROTOBUF_AS_DLL "Link against protobuf build as dynamic libray." OFF)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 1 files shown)


bertmaher:
Can you provide some more context here?  Under what conditions does an error occur (and what are the error messages)?  And why does this PR fix it?  Thanks!

CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ option(GLOW_WITH_OPENCL "Build the OpenCL backend" ON)
 option(GLOW_BUILD_EXAMPLES "Build the examples" ON)
 option(GLOW_BUILD_TESTS "Build the tests" ON)
 option(GLOW_WITH_BUNDLES "Build bundles" OFF)
+OPTION(LINK_PROTOBUF_AS_DLL "Link against protobuf build as dynamic libray." OFF)
```
nadavrot: All of the options before this one are spelled with lower case.

CMakeLists.txt
```diff
@@ -66,6 +67,18 @@ if (GLOW_WITH_OPENCL)
...
 
+# sets pre processor flag so some of glow projects
+# can link against protobuf libray that is build as DLL.
+if(LINK_PROTOBUF_AS_DLL)
+  # For protobuf warning when it is build as dll
+  # Supresses a warning that is treated as warning
+  # Basically one of the header files has interface class
+  # containing STL library. Which might cause issues
+  # if things are build with different compilers.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4251")
```
nadavrot: If we set LINK_PROTOBUF_AS_DLL to True on Linux things would not work. Can we wrap this in an if-def for windows?

CMakeLists.txt
ayermolo: Fixed

CMakeLists.txt
ayermolo: Added it as an AND.

nadavrot:
LGTM. Unless @bertmaher has any comments I am okay with landing this change as is.

PR #2229, opened 2019-01-03
[ci] Symbolize ASAN stack traces

bertmaher:
*Description*: So that we can actually debug ASAN failures that occur in CI but not locally
*Testing*: Used to debug #2227 
*Documentation*: n/a


bertmaher:
Rolled into #2227

PR #2233, opened 2019-01-03
[C2/ONNX] Add support for GatherRanges operator

### LINKED ISSUE #2210: [New operator] GatherRanges
We need to support `GatherRanges` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#gatherranges

It is not possible to lower this operation. It needs to be implemented for every backend. See `Gather` for guidance on how to implement, it is very similar op.

SplitInfinity:
**Description**
This commit adds support for the [`GatherRanges` operator (click me!)](https://caffe2.ai/docs/operators-catalogue.html#gatherranges). This operator
is not lowered and this commit includes interpreter and CPU backend
implementations for it.

**Testing**
This commit adds two unit tests (one with 32-bit ranges and another with
64-bit ranges) to `OperatorTest` that check operator correctness and one
unit test to `OnnxImporterTest` to check that the operator can be loaded
from a `.onnxtxt` correctly.

**Fixes**
This commit fixes #2210.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -857,6 +857,31 @@ protected:
     return true;
   }
 
+  llvm::Expected<bool> loadGatherRanges(const std::string &typeName,
+                                        const OpType &op,
+                                        const ArgumentDictionaryTy &dict) {
+    NodeValue data;
+    ASSIGN_VALUE_OR_RETURN_ERR(data,
+                               getNodeValueOrCreateConstantByName(op.input(0)));
+    RETURN_ERR_IF_NOT(data.dims().size() == 1, "Data must be a 1D vector.");
+
+    NodeValue ranges;
+    ASSIGN_VALUE_OR_RETURN_ERR(ranges,
+                               getNodeValueOrCreateConstantByName(op.input(1)));
+    RETURN_ERR_IF_NOT(ranges.dims().size() == 3, "Ranges must be a 3D vector.");
+    RETURN_ERR_IF_NOT(ranges.dims()[2] == 2,
+                      "Last dimension of ranges must be 2.");
+
+    unsigned_t maxOutputSize;
+    ASSIGN_VALUE_OR_RETURN_ERR(maxOutputSize,
+                               loadInt(dict.find("maxOutputSize")->second));
+
+    Node *GR = G_.createGatherRanges(loadOperatorName(op), data, ranges,
+                                     maxOutputSize);
+    addNodeAsOutput(op, GR);
+    return true;
+  }
+
   using ProtobufLoader::ProtobufLoader;
 
   /// If operator type is supported, returns Expected<true> and creates new
@@ -991,6 +1016,10 @@ protected:
 
     if (typeName == "Gather" || typeName == "BatchGather") {
       return loadGatherOps(typeName, op, dict);
+    }
+
+    if (typeName == "GatherRanges") {
+      return loadGatherRanges(typeName, op, dict);
     }
 
     return false;
--- a/lib/Backends/CPU/LLVMIRGen.cpp
+++ b/lib/Backends/CPU/LLVMIRGen.cpp
@@ -2209,6 +2209,45 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     break;
   }
 
+  case Kinded::Kind::GatherRangesInstKind: {
+    auto *GRI = llvm::cast<GatherRangesInst>(I);
+    auto *output = GRI->getOutput();
+    auto *lengths = GRI->getLengths();
+    auto *data = GRI->getData();
+    auto *ranges = GRI->getRanges();
+
+    auto *outputPtr = emitValueAddress(builder, output);
+    auto *lengthsPtr = emitValueAddress(builder, lengths);
+    auto *dataPtr = emitValueAddress(builder, data);
+    auto *rangesPtr = emitValueAddress(builder, ranges);
+
+    auto rangesType = ranges->getType();
+
+    // The number of examples in ranges.
+    size_t numExamples = rangesType->dims()[0];
+    // The number of range pairs in each example.
+    size_t exampleSize = rangesType->dims()[1];
+
+    auto *numExamplesVal = emitConstSizeT(builder, numExamples);
+    auto *exampleSizeVal = emitConstSizeT(builder, exampleSize);
+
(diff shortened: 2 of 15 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -876,6 +876,78 @@ void InterpreterFunction::fwdGatherInst(const glow::GatherInst *I) {
...
+      ElemTy start = rangesT->getHandle<ElemTy>().at({example, range, 0});
+      ElemTy len = rangesT->getHandle<ElemTy>().at({example, range, 1});
+
+      // Add the length of this current range to the example length counter.
+      totalLen += len;
+
+      // Copy the specified data to outT.
+      std::copy(&dataT->getUnsafePtr()[start * dataElementSize],
+                &dataT->getUnsafePtr()[(start + len) * dataElementSize],
+                &outT->getUnsafePtr()[outP]);
```
yinghai: Do we have some sort of check on whether this will go out of bound?

nadavrot: The TensorHandle API has bounds checks, but raw pointer access does not.

yinghai: Yeah, so here it looks like raw pinter access. Maybe add some checks?

yinghai:
Code looks good to me. My question is how we are going to lower this to hardware or is it doable?

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -857,6 +857,27 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+    ASSIGN_VALUE_OR_RETURN_ERR(data,
+                               getNodeValueOrCreateConstantByName(op.input(0)));
+
+    NodeValue ranges;
+    ASSIGN_VALUE_OR_RETURN_ERR(ranges,
+                               getNodeValueOrCreateConstantByName(op.input(1)));
+
+    unsigned_t maxOutputSize;
+    ASSIGN_VALUE_OR_RETURN_ERR(maxOutputSize,
+                               loadInt(dict.find("maxOutputSize")->second));
```
rdzhabarov: where is 'maxOutputSize' coming from? I cannot see it in https://caffe2.ai/docs/operators-catalogue.html#gatherranges

rdzhabarov:
@yinghai 
> My question is how we are going to lower this to hardware or is it doable?

Not directly on MM unit, but if there are general purpose CPUs on chip and depending on # cpus there, it might be possible to run with acceleration.

include/glow/Importer/CommonOperatorLoader.h
SplitInfinity: As @artemrakhov and I discussed in #2210, the shape of `output` can not be determined at compile time without a hint. `maxOutputSize` specifies the largest possible size of `output`, and any nodes that use it must use the value of `lengths` to determine how much of `output` is valid data (since `sum(lengths) == output.size()`).

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -807,6 +807,34 @@ TEST(onnx, gather) {
   EXPECT_TRUE(gather->getResult().dims().equals({2, 4, 2}));
 }
 
+/// Test loading GatherRanges from an ONNX model.
+TEST(onnx, gatherranges) {
```
rdzhabarov: nit: gatherRanges

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -807,6 +807,34 @@ TEST(onnx, gather) {
...
+  Placeholder *output;
+  Tensor data(ElemKind::FloatTy, {6});
+  Tensor ranges(ElemKind::Int32ITy, {2, 2, 2});
+
+  {
+    ONNXModelLoader onnxLD(netFilename, {"data", "ranges"},
+                           {&data.getType(), &ranges.getType()}, *F);
+    output = EXIT_ON_ERR(onnxLD.getOutputByName("output"));
+  }
+  EE.compile(CompilationMode::Infer, F);
```
rdzhabarov: why do we need to compile if not running?
Testing just loader part should be fine here and testing Op itself in the OperatorTest.

lib/Graph/Nodes.cpp
```diff
@@ -862,6 +862,32 @@ bool GatherNode::verify() const {
   return isValid;
 }
 
+bool GatherRangesNode::verify() const {
+  bool isValid = expectCompareTrue(
```
rdzhabarov: it would be good to reject op earlier in the pipeline, at the moment of actual loading in the loader.

SplitInfinity:
**Changes Since Last Revision**

@yinghai
- added bounds checks to interpreter implementation
- added unit test for Caffe2 importer to prove that it works for Caffe2 proto as well (overkill, maybe?)
 
@rdzhabarov 
- added code to reject op during model loading if inputs have incorrect dimensions
- fixed nits

@jfix71 
- added verification that `Data` and `Output` are the same type in `InstrGen`
- refactored unit test in `OperatorTest` using generic function templated on index type
- fixed nits

**Unresolved Issues**
- as @jfix71 pointed out, we use `size_t` for 64-bit width integer implementations instead of `int64_t`. I think this is best left to a separate PR to fix.

rdzhabarov:
> as @jfix71 pointed out, we use size_t for 64-bit width integer implementations instead of int64_t. I think this is best left to a separate PR to fix.

If you do not plan to fix it right away, can you file an issue for this?

SplitInfinity:
@jackm321 

I opened #2252 to address (most of) your comments.

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -2209,6 +2209,45 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
+
+    // Dispatching function depending on the input type of Ranges.
+    llvm::Function *F = nullptr;
+    if (ranges->getElementType() == ElemKind::Int64ITy) {
+      F = getFunction("gatherranges64", output->getElementType());
+    } else if (ranges->getElementType() == ElemKind::Int32ITy) {
+      F = getFunction("gatherranges32", output->getElementType());
+    }
+    if (!F) {
+      llvm_unreachable("Cannot get function for GatherRanges. "
```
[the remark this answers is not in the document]
SplitInfinity: I followed what was being done for the `gather` operator above

PR #2235, opened 2019-01-04
Redesign isOpSupported()

jfix71:
*Description*: This PR redesigns `isOpSupported()` to allow a backend to specify what ops it supports based on the input and output types of Nodes. It updates all of the open source backends we have.

*Testing*: All tests are passing. I've also done my best to make sure bigger tests (e.g. those in `tests/images/run.sh`) also still work.

*Documentation*: Updated.

Fixes https://github.com/pytorch/glow/issues/2194

### DIFF
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -73,9 +73,8 @@ public:
     return false;
   }
 
-  /// \returns true if backend supports given kind of operation with
-  /// the given \p elementTy element type.
-  virtual bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const = 0;
+  /// \returns whether the provided \p NI is supported by the backend.
+  virtual bool isOpSupported(const NodeInfo &NI) const = 0;
 
   /// \returns true if the supplied Node \N should be lowered. By default, all
   /// Nodes are candidates for lowering.
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -94,9 +94,10 @@ public:
   void insertCompiledFunction(llvm::StringRef name,
                               std::unique_ptr<CompiledFunction> func);
 
-  /// \returns whether operation is supported by the underlying backend.
-  bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
-    return backend_->isOpSupported(opKind, elementTy);
+  /// \returns whether a node with the provided \p NI is supported by the
+  /// underlying backend.
+  bool isOpSupported(const NodeInfo &NI) const {
+    return backend_->isOpSupported(NI);
   }
 
   /// Optimize the Function \p f and pass it to the backend to compile it for a
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -254,6 +254,10 @@ public:
     return outTypes_[idx]->getElementType();
   }
 
+  /// \returns whether all of the element types of inTypes_ and outTypes_ are
+  /// all the same and one of those found in \p allowedElemKinds. \p ignoreIn
+  /// and \p ignoreOut represent indices that can be skipped in inTypes_ and
+  /// outTypes_ respectively.
   bool
   allInputsAndOutputsHaveSameElemKind(llvm::ArrayRef<ElemKind> allowedElemKinds,
                                       const IndicesSet &ignoreIn = {},
@@ -267,6 +271,8 @@ public:
     return false;
   }
 
+  /// Helper for debugging which \returns a string representation for the
+  /// NodeInfo.
   std::string getDebugDesc() const {
     DescriptionBuilder db(getKindName());
     for (size_t i = 0; i < inTypes_.size(); ++i) {
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -26,8 +26,10 @@
 #include "glow/Base/Type.h"
 #include "glow/Graph/NodeValue.h"
 #include "glow/Graph/UseDef.h"
+#include "glow/Support/Support.h"
 
 #include <list>
+#include <unordered_set>
 
 namespace glow {
 
@@ -183,6 +185,98 @@ public:
 
   /// Dtor.
   virtual ~NodeWalker() = default;
+};
+
+using IndicesSet = std::unordered_set<unsigned>;
+
+/// Helper class to hold info of a Node, containing its \p opKind, \p inTypes,
+/// and \p outTypes
+class NodeInfo : public Kinded {
+private:
+  /// The input types of the NodeInfo.
+  std::vector<TypeRef> inTypes_;
+  /// The output types of the NodeInfo.
+  std::vector<TypeRef> outTypes_;
+
+  /// Helper function for checking if all of the ElemKinds contained in \p types
+  /// are equal to \p allowedElemKind. Indices in \p ignore are ignored when
+  /// checking from \p types.
+  bool allSameElemKind(const ElemKind allowedElemKind,
+                       llvm::ArrayRef<TypeRef> types,
+                       const IndicesSet &ignore) const {
+    for (size_t i = 0; i < types.size(); i++) {
+      if (ignore.count(i)) {
+        continue;
+      }
+      const TypeRef currType = types[i];
+      if (currType->getElementType() != allowedElemKind) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+public:
+  NodeInfo(Kinded::Kind kind, llvm::ArrayRef<TypeRef> inTypes,
+           llvm::ArrayRef<TypeRef> outTypes)
+      : Kinded(kind), inTypes_(inTypes), outTypes_(outTypes) {}
+
+  NodeInfo(const Node &N) : Kinded(N.getKind()) {
+    for (unsigned i = 0, e = N.getNumResults(); i < e; ++i) {
+      outTypes_.push_back(N.getType(i));
+    }
+    for (unsigned idx = 0, end = N.getNumInputs(); idx != end; ++idx) {
+      inTypes_.push_back(N.getNthInput(idx).getType());
+    }
+  }
+
+  /// \returns the input type located at \p idx.
+  const TypeRef getInTy(size_t idx) const {
+    assert(idx < inTypes_.size());
+    return inTypes_[idx];
+  }
+
+  /// \returns the output type located at \p idx.
+  const TypeRef getOutTy(size_t idx) const {
+    assert(idx < outTypes_.size());
+    return outTypes_[idx];
+  }
+
+  /// \returns the input type located at \p idx.
+  const ElemKind getInElemTy(size_t idx) const {
+    assert(idx < inTypes_.size());
+    return inTypes_[idx]->getElementType();
+  }
+
+  /// \returns the output type located at \p idx.
+  const ElemKind getOutElemTy(size_t idx) const {
+    assert(idx < outTypes_.size());
+    return outTypes_[idx]->getElementType();
+  }
+
+  bool
+  allInputsAndOutputsHaveSameElemKind(llvm::ArrayRef<ElemKind> allowedElemKinds,
+                                      const IndicesSet &ignoreIn = {},
+                                      const IndicesSet &ignoreOut = {}) const {
+    for (const ElemKind elemKind : allowedElemKinds) {
+      if (allSameElemKind(elemKind, inTypes_, ignoreIn) &&
+          allSameElemKind(elemKind, outTypes_, ignoreOut)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  std::string getDebugDesc() const {
+    DescriptionBuilder db(getKindName());
+    for (size_t i = 0; i < inTypes_.size(); ++i) {
+      db.addParam("inType" + std::to_string(i), *inTypes_[i]);
+    }
+    for (size_t i = 0; i < outTypes_.size(); ++i) {
+      db.addParam("outType" + std::to_string(i), *outTypes_[i]);
+    }
+    return db;
+  }
 };
 
 /// Helper class to walk through the specific uses of a NodeValue.
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -160,54 +160,258 @@ void CPUBackend::save(Function *F, llvm::StringRef outputDir,
   BundleSaver(IR.get()).save(tgt, outputDir, networkName);
 }
 
-bool CPUBackend::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
-  // Check for quantization support.
-  if (elementTy == ElemKind::Int8QTy) {
-    switch (opKind) {
-    case Kinded::Kind::AddNodeKind:
-    case Kinded::Kind::BatchedAddNodeKind:
-    case Kinded::Kind::BatchedReduceAddNodeKind:
-    case Kinded::Kind::CmpLTENodeKind:
-    case Kinded::Kind::ConcatNodeKind:
-    case Kinded::Kind::ConvolutionNodeKind:
-    case Kinded::Kind::DequantizeNodeKind:
-    case Kinded::Kind::DivNodeKind:
-    case Kinded::Kind::FullyConnectedNodeKind:
-    case Kinded::Kind::GatherNodeKind:
-    case Kinded::Kind::LogNodeKind:
-    case Kinded::Kind::MatMulNodeKind:
-    case Kinded::Kind::MaxNodeKind:
-    case Kinded::Kind::MinNodeKind:
-    case Kinded::Kind::MulNodeKind:
-    case Kinded::Kind::AvgPoolNodeKind:
-    case Kinded::Kind::MaxPoolNodeKind:
-    case Kinded::Kind::QuantizeNodeKind:
-    case Kinded::Kind::ReluNodeKind:
-    case Kinded::Kind::RescaleQuantizedNodeKind:
-    case Kinded::Kind::ReshapeNodeKind:
-    case Kinded::Kind::SelectNodeKind:
-    case Kinded::Kind::SliceNodeKind:
-    case Kinded::Kind::SigmoidNodeKind:
-    case Kinded::Kind::SubNodeKind:
-    case Kinded::Kind::TanhNodeKind:
-    case Kinded::Kind::TileNodeKind:
-    case Kinded::Kind::TopKNodeKind:
-    case Kinded::Kind::TransposeNodeKind:
-      return true;
-    default:
-      return false;
+bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
+  // Note: For brevity below, "X ==> Y, Z" signifes that Node X is IRGen'd into
+  // Instructions Y and Z.
+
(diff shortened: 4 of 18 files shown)


jackm321:
@jfix71 I'm glad to see this move in the direction of considering all inputs and outputs. I think the cost function is cool but I'm also a little unclear about how it would be used in practice. Is it imagined that the costs of various operators would just be compared within the same backend for different operators? This doesn't sound likely so then how will we ensure that these costs are defined so that they are comparable across backends when each backend defines it's own cost for each operator?

jfix71:
> Is it imagined that the costs of various operators would just be compared within the same backend for different operators? This doesn't sound likely so then how will we ensure that these costs are defined so that they are comparable across backends when each backend defines it's own cost for each operator?

Good questions! I was moving in this direction as suggested in https://github.com/pytorch/glow/issues/2194#issuecomment-449191002.

Overall yeah, I think it's more useful across different backends, both from the perspective of partitioning and from ONNXIFI determining where to execute a given function. If we wanted the cost to represent execution speed, I think we could just define these based on wall time, as we have an expectation of how long each operator would take to execute.

For the same backend, perhaps it could be useful to know for deciding where/when to partition a function across multiple of the same backend. As a really trivial (and perhaps unhelpful 😛) example, assume individually `Conv` and `Relu` cost "1", and a fused `ConvRelu` also costs "1" since `Relu` is essentially free on some accelerators when fused to `Conv`. So this logic lets you know it's a bad idea to partition in between them.

jackm321:
@jfix71 I don't know a lot about the hardware but estimating wallclock times for various inputs and outputs sounds like it would be difficult to get to be correct and approximations may miss things like the ConvRelu example if they're not accurate enough to discern that Conv and ConvRelu are the same cost.

Also for the ConvRelu fusion example, does the partitioner get an optimized graph or not? Because if it does it seems like it would be better just to have an optimization pass (or prelowering maybe?) that does fusions like this for sure on backends where it's good to do that and then the partitioner doesn't have to worry about whether it should or shouldn't make a cut there in the graph based on cost because it will just be a single fused node at that point.

Overall I guess my concern is that this cost function is nice because the idea behind `UnsupportedCostVal = std::numeric_limits<CostVal>::max()` but in practice I'm not sure if the values between `1` and `UnsupportedCostVal` can actually be reliably generated and used by the partitioner and/or onnxifi so this might be overly general. Not at all sure if this is true or not though 😛

jfix71:
> I don't know a lot about the hardware but estimating wallclock times for various inputs and outputs sounds like it would be difficult to get to be correct

@jackm321 I agree it would be difficult for an entire graph, but I don't think estimating a single operator's costs would be too difficult. It would be the responsibility of the higher level partitioner to understand the backend's architecture and the graph as a whole to make decisions on when to partition. Perhaps backends would need a `getSubgraphCost()` which would call into `getOpCost()`. Overall though I'll defer here to @gcatron and others working on the partitioner/runtime that have a better idea of where the partitioner is going/what info it needs/how it should get that info.

Also, another place this could be useful for a single backend is in the GraphOptimizer. Right now backends have no control over whether an optimization should take place (https://github.com/pytorch/glow/issues/1641). One solution here could be for the GraphOptimizer to query the backend to get the costs of the original subgraph and the new subgraph to decide whether to perform the opti. Similar to the partitioner, we could get each cost using something like `getSubgraphCost()`, calling into `getOpCost()`.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -67,11 +67,17 @@ class ExecutionEngine final {
...
+  /// \returns a cost associated with executing the provided \p opKind with \p
+  /// inputTypes and \p outputTypes on the underlying backend.
+  CostVal getOpCost(Kinded::Kind opKind, llvm::ArrayRef<TypeRef> inputTypes,
+                    llvm::ArrayRef<TypeRef> outputTypes) const {
+    return backend_->getOpCost(opKind, inputTypes, outputTypes);
   }
 
+  /// \returns whether the cost associated with executing Node \p N on the
+  /// underlying backend.
+  CostVal getOpCost(const Node &N) const { return backend_->getOpCost(N); }
```
zakk0610: I think maybe this function need to be `virtual` for backend to provide more accurate estimation.
because the cost is depend on operator's parameter., 
For example, if the input size is too small, the CPU maybe faster than the ASIC.

jfix71: The actual logic for computing an op's cost is contained in a virtual `getOpCost()`:

https://github.com/pytorch/glow/blob/0c01a667b8fb26f7752ae7ed63c4b6d134ed729c/include/glow/Backends/Backend.h#L78-L82

This function here in ExecutionEngine just a helper method. It forwards to the backend's helper `getOpCost(Node &N)`, which then calls into each backend's implementation of `getOpCost()`.

zakk0610: Thanks, I was missed the `TypeRef` can represents a type of tensor. 
Sorry to bother you.

jfix71: No bother at all, thanks for checking out the PR! We are happy to hear any questions or insights from everyone contributing to the project.

include/glow/Graph/Node.h
```diff
@@ -145,6 +145,9 @@ class Node : public Named,
 
   ~Node() { releaseUsers(); }
 
+  /// \returns an array ref of the types of the node.
```
rdzhabarov: based on the comment it's not clear whether it's input/output or both.

lib/Backends/Backends.cpp
```diff
@@ -62,3 +62,14 @@ Backend *glow::createBackend(BackendKind backendKind) {
   // always covers all possible values.
   llvm_unreachable("unreachable");
 }
+
+CostVal Backend::getOpCost(const Node &N) const {
+  // Gather the input types into a vector. Note that we do not need to do so for
+  // the result types because they are already contained in the node.
+  std::vector<TypeRef> inputTypeRefs;
```
rdzhabarov: nit: but when we know size in advance, just allocate it right away in vector.

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -168,13 +172,13 @@ bool CPUBackend::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
     case Kinded::Kind::TileNodeKind:
     case Kinded::Kind::TopKNodeKind:
     case Kinded::Kind::TransposeNodeKind:
-      return true;
+      return 1;
```
rdzhabarov: can you make this a named const (MinimalCostVal, there might be a better name)?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -135,7 +135,11 @@ void CPUBackend::save(Function *F, llvm::StringRef outputDir,
   BundleSaver(IR.get()).save(tgt, outputDir, networkName);
 }
 
-bool CPUBackend::isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const {
+CostVal CPUBackend::getOpCost(Kinded::Kind opKind,
+                              llvm::ArrayRef<TypeRef> inputTypes,
+                              llvm::ArrayRef<TypeRef> outputTypes) const {
+  // TEMPORARY: Will remove once not a WIP. This is just so things compile.
+  auto elementTy = inputTypes[0]->getElementType();
```
rdzhabarov: Do you plan on handling int16 and fp32 in the follow up?

lib/Backends/CPU/CPUBackend.cpp
jackm321: Yes I think this would help with clarity to have a name for this value. Not sure if minimal is right though because I guess somethings may be 0 if they are "free" like the Relu in ConvRelu example?

rdzhabarov: well it depends, on CPU if not fused `relu` is not "free".

lib/Backends/CPU/CPUBackend.cpp
jfix71: Yep! I will update the CPU and OpenCL backend's versions too in this PR. I first just wanted to make sure everyone agreed the Interpreter's `getOpCost()`/the general approach seemed reasonable.

include/glow/Graph/Node.h
jfix71: Done

lib/Backends/Backends.cpp
jfix71: Done

lib/Backends/CPU/CPUBackend.cpp
jfix71: I added `MinimumUnitCostVal` and replaced all `1`s with it. If we keep it I'm thinking backends would use this minimum unit cost to define other costs, i.e. as some multiple of that. It seems a little silly since it is just `1` and I'm not sure we would ever change it from `1`, but perhaps this preferable to just specifying `1`? Let me know what you think.

I also added `FreeCostVal = 0` to differentiate from the minimum, however it's not used anywhere, so not sure it's worth including for now.

rdzhabarov: > It seems a little silly since it is just 1 and I'm not sure we would ever change it from 1, but perhaps this preferable to just specifying 1?

hopefully name should give slightly more context than plain '1'.

> I also added FreeCostVal = 0 to differentiate from the minimum, however it's not used anywhere, so not sure it's worth including for now.

up to you, we can introduce that once there is a need.

jackm321: >  It seems a little silly since it is just 1 and I'm not sure we would ever change it from 1, but perhaps this preferable to just specifying 1?

@jfix71 I don't think it's too silly, I imagine it would give enough context to make it pretty easy to understand what's being expressed

include/glow/Backends/Backend.h
```diff
@@ -71,9 +71,14 @@ class Backend {
...
-  /// the given \p elementTy element type.
-  virtual bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const = 0;
+  /// \returns whether the provided \p NI is supported by the backend.
+  virtual bool isOpSupported(Function *F, const NodeInfo &NI) const = 0;
+
+  /// \returns whether the provided \p N is supported by the backend. Gathers
+  /// info from \p N calls into the virtual \p ref isOpSupported().
+  bool isOpSupported(Function *F, const Node &N) const;
+
+  bool isOpSupportedViaLowering(Function *F, const NodeInfo &NI) const;
```
jackm321: doxygen comment.

include/glow/Optimizer/Optimizer.h
```diff
@@ -43,6 +43,12 @@ void optimize(Function *F, CompilationMode mode);
 /// operators.
 void lower(Function *F, const Backend &B);
 
+/// \returns a pair of a bool and a vector of the component NodeInfo of nodes
+/// that would be lowered from NI. If no nodes would be lowered from this node,
+/// then the bool is set to false.
```
jackm321: Why is the bool needed? Why not just return the empty vector?

include/glow/Graph/Node.h
```diff
@@ -185,6 +191,35 @@ class NodeWalker {
...
+/// Helper struct to hold info of a Node, containing its \p opKind, \p inTypes,
+/// and \p outTypes
+struct NodeInfo {
+private:
+  Kinded::Kind opKind_;
+  llvm::ArrayRef<TypeRef> inTypes_;
+  llvm::ArrayRef<TypeRef> outTypes_;
+
+public:
+  NodeInfo(Kinded::Kind opKind, llvm::ArrayRef<TypeRef> inTypes,
```
jackm321: Maybe also add a constructor that takes const Node&?

lib/Backends/Backends.cpp
```diff
@@ -62,3 +62,27 @@ Backend *glow::createBackend(BackendKind backendKind) {
...
+bool Backend::isOpSupportedViaLowering(Function *F, const NodeInfo &NI) const {
+  std::pair<bool, std::vector<NodeInfo>> loweredNodeInfos =
+      getLoweredComponentInfo(F, NI);
+
+  // If the node cannot be lowered then return false.
+  if (!loweredNodeInfos.first) {
+    return false;
+  }
+
+  for (const NodeInfo &NI : loweredNodeInfos.second) {
```
jackm321: Could you use a fresh name here instead of aliasing the parameter?

lib/Optimizer/Lower.cpp
```diff
@@ -694,6 +694,141 @@ static void lowerTileNode(Function *F, const TileNode &TN) {
   TN.getResult().replaceAllUsesOfWith(IN);
 }
 
+std::pair<bool, std::vector<NodeInfo>>
+glow::getLoweredComponentInfo(Function *F, const NodeInfo &NI) {
```
jackm321: Add comments about what this does and to keep this in sync with actual lowering code.
Can we add tests to check that this is consistent with lowering code?

include/glow/Optimizer/Optimizer.h
jfix71: Some nodes are lowered by replacing them with their input, e.g. the `RegressionNode` and `AddGradNode`. So we want to say that the Node can be lowered and that it is lowered into nothing.

lib/Optimizer/Lower.cpp
jfix71: There are some comments in `Optimizer.h` -- I added a little more to it.

It shouldn't be too bad to create a `LowerTest.cpp`, which creates nodes and then calls into both `lower()` and `getLoweredComponentInfo()` and makes sure they match.

jfix71:
I've added a few tests in an initial `LowerTest`. It turned out pretty clean, and so should be pretty easy to add new tests for.

I'm worried about the size of this PR at this point. It's really big and this is before I've finished implementing `getLoweredComponentInfo()` or `OpenCL::/CPU::isOpSupported()` or `LowerTest`. Unfortunately I think they make sense/need to land together, though... 😞Any suggestions on how to split it would be welcome!

include/glow/Backends/Backend.h
```diff
@@ -71,9 +71,12 @@ class Backend {
...
   /// @}
 
-  /// \returns true if backend supports given kind of operation with
-  /// the given \p elementTy element type.
-  virtual bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const = 0;
+  /// \returns whether the provided \p NI is supported by the backend.
+  virtual bool isOpSupported(Function *F, const NodeInfo &NI) const = 0;
+
+  /// \returns whether the provided \p NI is supported by the backend via
+  /// lowering the \p NI into other NodeInfos, matching the lowering logic.
```
rdzhabarov: It would worthwhile to mention here that if node cannot be lowered than it's not `isOpSupportedViaLowering` even though the node itself could be supported otherwise.

lib/Graph/Nodes.cpp
```diff
@@ -296,8 +296,13 @@ static bool verifyRelu(NodeValue result, NodeValue input) {
 
 static bool verifyRegression(NodeValue src, NodeValue dest,
                              NodeValue expected) {
-  return checkSameType(src, dest, dest.getNode()) &&
-         checkSameType(dest, expected, dest.getNode());
+  const Node *parent = dest.getNode();
```
rdzhabarov: this seems unrelated to the PR.

jfix71: It was actually required for MLTest's `testFindPixelRegression` test to pass. It quantizes a regression node, which is eventually lowered into nothing during inference but is there for training. I could move it to another PR if desired.

jfix71:
Now that https://github.com/pytorch/glow/pull/2311 has landed I will remove all the `isOpSupportedViaLowering()`, `getLoweredComponentInfoI()`, etc. stuff. I think the only thing left to do here then will be to update the `isOpSupported()` for the OpenCL and CPU backends.

jackm321:
@jfix71 are you planning on using the details of NodeInfo for `CPUBackend::isOpSupported` and `OCLBackend::isOpSupported`? What's here looks good to me, just not sure if you're planning on adding stuff or not.

jfix71:
@jackm321 Yeah, still need to update `isOpSupported()` for the CPU and OpenCL backends, I just haven't had the cycles to make progress here. Additionally the Interpreter backend still needs some work too -- locally I've fixed a few bugs in what's already up here, but still need to add support for all the gradient Nodes that are missing. I've made some updates locally but haven't pushed them up yet.

jfix71:
Update: Interpreter `isOpSupported()` is now complete and passes all tests. Moving on to updating CPU and OpenCL backends next.

jfix71:
OK! This is finally ready.

FYI: I needed to fix a few tests. This is mostly no big deal. There is one test however which I disabled for the OpenCL backend -- the `end2endInt16` from QuantizationTest. OpenCL has no `Int16QTy` support -- the test previously passed because the quantizer basically just skipped over quantizing the whole thing, and so the results compared to the Interpreter passed. This isn't great -- I've filed an issue about it: https://github.com/pytorch/glow/issues/2464

tests/unittests/QuantizationTest.cpp
```diff
@@ -1708,33 +1711,37 @@ INSTANTIATE_TEST_CASE_P(Interpreter, Quantization,
...
 #ifdef GLOW_WITH_OPENCL
 INSTANTIATE_TEST_CASE_P(
-    OpenCL, Operator,
-    ::testing::Combine(::testing::Values(BackendKind::Interpreter,
-                                         BackendKind::Interpreter),
+    InterpProfOpenCLQuant, Operator,
+    ::testing::Combine(::testing::Values(BackendKind::Interpreter),
                        ::testing::Values(BackendKind::OpenCL)));
-#endif // GLOW_WITH_CPU
+#endif // GLOW_WITH_OPENCL
```
jackm321: nice

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -160,54 +160,348 @@ void CPUBackend::save(Function *F, llvm::StringRef outputDir,
...
+  // Note: For brevity below, "X ==> Y, Z" signifes that Node X is IRGen'd into
+  // Instructions Y and Z.
+
+  switch (NI.getKind()) {
+  case Kinded::Kind::AddNodeKind:
+  case Kinded::Kind::SubNodeKind:
+  case Kinded::Kind::MulNodeKind:
+  case Kinded::Kind::MaxNodeKind:
+  case Kinded::Kind::MinNodeKind:
+    // Note: Int8QTy Log and Tanh are IRGen'd into a lookup table.
```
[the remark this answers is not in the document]
jfix71: This is how `clang-format` indents comments inside switch statements 🤷‍♂️

include/glow/Graph/Node.h
```diff
@@ -185,6 +187,98 @@ class NodeWalker {
...
+    for (const ElemKind elemKind : allowedElemKinds) {
+      if (allSameElemKind(elemKind, inTypes_, ignoreIn) &&
+          allSameElemKind(elemKind, outTypes_, ignoreOut)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  std::string getDebugDesc() const {
```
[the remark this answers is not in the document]
jfix71: Not used but useful when debugging so I'd like to keep it.

jfix71:
@jackm321 Thanks for the review! Will merge once tests pass to avoid conflicts.

PR #2238, opened 2019-01-08
[Trivial] RuntimeBundle

beicy:
*Description*:
Removed "const" and added a default Constructor in RuntimeBundle Class.
 
*Testing*:
Ninja check

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -40,11 +40,11 @@ class RuntimeBundle {
   /// Pointer to memory containing the weights for execution.
   uint8_t *constants_;
   /// Amount of memory needed for weights.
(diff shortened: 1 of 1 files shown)


bertmaher:
Could you explain the rationale for this change?

bertmaher:
> const member variables means no default copy constructor or assignment operator

@gcatron After we talked about this I went digging a bit, because it sounded surprising to me, and I don't see evidence that const member variables inhibit the implicit default copy constructor (For example: https://gcc.godbolt.org/z/VYGtRH).

So anyways, I'm fine with this change because it's just not a big deal, but it might be worth checking if you're not hitting something subtly different that this change is fixing instead.

PR #2240, opened 2019-01-08
[cleanup] Trivial fixes

rdzhabarov:
*Description*:
Some trivial fixes in onnxifi.
*Testing*:
CI

cc: @yinghai


lib/Importer/ProtobufLoader.cpp
```diff
@@ -38,13 +38,9 @@ llvm::Expected<Placeholder *>
 ProtobufLoader::getOutputByName(llvm::StringRef name) const {
   RETURN_ERR_IF_NOT(
       outputVarsByName_.count(name),
```
yinghai: Can we use iterator to compare with end() to save one lookup?

rdzhabarov: yes, it does not hurt to make it a bit more performant.

PR #2241, opened 2019-01-08
[Caffe2ModelLoader] Fix an error-checking

beicy:
*Description*:
Just found quantized resnet50 failed due to improper error checking: 
For quantized ops, there is an arg "order" to indicate it is "NHWC" or "NCHW", and in the quantized resnet50 model, all orders are "NHWC". Therefore, the input of quantized conv and pool don't have to be only "Transposed" or "Storage" type.  So the current error checking caused the crash.

*Testing*: ninja check, run.sh

*Documentation*:
[Optional Fixes #issue]


beicy:
@rdzhabarov Yes, the quantized model is also tested in run.sh

rdzhabarov:
> Yes, the quantized model is also tested in run.sh

@beicy, This does not run on CI unfortunately.

PR #2243, opened 2019-01-08
[Windows Fix] Increasing stack size for windows.

ayermolo:
*Description*:
Default stack size on windows is less then in linux. One by product of this is BackendCorrectnessTest fails on windows. In debug mode a stack check fails within system dll call that does stack check allocation for frame. In release mode there is also run time exception. Set the size to 2^22. Next power of 2 to cover what BackendCorrectnessTest max uses.
*Testing*: Unit Tests
*Documentation*:
[Optional Fixes #issue]


nadavrot:
@ayermolo Should we be allocating so much memory on the stack? Can we update the test to allocate on the heap instead?

ayermolo:
The way I understand it looking at IR, and implementation of libjit_matmul_outer and libjit_matmul_f alloca comes from libjit_matmul_outer implementation:
float packedB[kc * nc] __attribute__((aligned(64)));.
So llvm just goes with it. I guess another fix be change implementation to use heap? Although it's still half the size of unix, 8mb.

rdzhabarov:
I'm fine with the increasing stack size. Making dynamic allocation in libjit_matmul_inner looks inefficient.

We have gemm bench, can you do benchmark gemm with the dynamic allocation?

bertmaher:
To provide context, those stack allocations are the packed arrays for gemm, and it's important for performance not to perform a dynamic allocation.  A potential alternative is to use thread-local statics (which could be allocated lazily if desired).  But bumping the Windows stack size is probably the easiest route :-).

ayermolo:
Thanks! :)

Issue #2226, opened 2019-01-02
[Docs] Need to update the Backends section to reflect the new APIs


nadavrot:
It looks like following section is not up to date with the recent changes done in compiledFunction class:

https://github.com/pytorch/glow/blob/master/docs/Backends.md#compiledfunction-abstract-class

PR #2246, opened 2019-01-08
Allow Compilation without Collecting Constants

gcatron:
*Description*: For the new runtime we need to be able to compile a function but collect the constants slightly later in the initialization process. This PR introduces a compileWithoutConstants method to the backend and updates existing backends to support it.
*Testing*: passes all unit tests.
*Documentation*: N/A
Fixes: #2254

### DIFF
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -43,6 +43,10 @@ public:
   /// Generate code for input function \param F.
   virtual std::unique_ptr<CompiledFunction> compile(Function *F) const = 0;
 
+  /// Generate code for input function \param F but do not collect constants.
+  virtual std::unique_ptr<CompiledFunction>
+  compileWithoutConstants(Function *F) const = 0;
+
   /// Save the bundle for \p F for a later standalone execution
   /// in \p outputDir. Make \p networkName the function name for
   /// the entry point of the network and prepend all generated
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -141,8 +141,6 @@ glow::generateRuntimeBundle(const IRFunction &F,
   }
   auto activationsMaxSize = activationsAllocator.getMaxMemoryUsage();
 
-  runtime::RuntimeBundle info(symbolTable, constantMaxSize, placeholderMaxSize,
-                              activationsMaxSize);
-  info.collectConstants(&F);
-  return info;
+  return runtime::RuntimeBundle(symbolTable, constantMaxSize,
+                                placeholderMaxSize, activationsMaxSize);
 }
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -99,14 +99,21 @@ CPUBackend::createIRGen(IRFunction *IR,
 
 std::unique_ptr<CompiledFunction>
 CPUBackend::compileIR(std::unique_ptr<IRFunction> IR) const {
+  auto function = compileIRWithoutConstants(IR.get());
+  static_cast<CPUFunction *>(function.get())->collectConstants(IR.get());
+  return function;
+}
+
+std::unique_ptr<CompiledFunction>
+CPUBackend::compileIRWithoutConstants(IRFunction *IR) const {
   AllocationsInfo allocationsInfo;
-  std::unique_ptr<LLVMIRGen> irgen = createIRGen(IR.get(), allocationsInfo);
+  std::unique_ptr<LLVMIRGen> irgen = createIRGen(IR, allocationsInfo);
   irgen->initTargetMachine(target.empty() ? "" : target.getValue(),
                            llvm::CodeModel::Model::Large);
   irgen->initCodeGen();
   // Perform the address assignment for activations and WeightVars.
 
-  allocateJITMemory(IR.get(), irgen->getAllocationsInfo());
+  allocateJITMemory(IR, irgen->getAllocationsInfo());
   // Create the jitmain function to be invoked by JIT.
   emitJitMain(*irgen);
   // Emit the code for the body of the entry function.
@@ -126,6 +133,12 @@ CPUBackend::compileIR(std::unique_ptr<IRFunction> IR) const {
 std::unique_ptr<CompiledFunction> CPUBackend::compile(Function *F) const {
   auto IR = generateAndOptimizeIR(F, shouldShareBuffers());
   return compileIR(std::move(IR));
+}
+
+std::unique_ptr<CompiledFunction>
+CPUBackend::compileWithoutConstants(Function *F) const {
+  auto IR = generateAndOptimizeIR(F, shouldShareBuffers());
+  return compileIRWithoutConstants(IR.get());
 }
 
 void CPUBackend::save(Function *F, llvm::StringRef outputDir,
--- a/lib/Backends/CPU/CPUBackend.h
+++ b/lib/Backends/CPU/CPUBackend.h
@@ -45,7 +45,13 @@ public:
   std::unique_ptr<CompiledFunction>
   compileIR(std::unique_ptr<IRFunction> IR) const override;
 
+  std::unique_ptr<CompiledFunction>
+  compileIRWithoutConstants(IRFunction *IR) const;
+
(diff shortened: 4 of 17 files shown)


nadavrot:
@gcatron Please make sure to update the docs (see #2226) and communicate the plan. We need to make sure that people who are developing Glow backends are not surprised by these changes. Please make sure to update the PR with the overall plan and send an email with link to the PR. 

https://github.com/pytorch/glow/issues/2226

include/glow/Backends/Backend.h
```diff
@@ -42,6 +42,9 @@ class Backend {
 
   /// Generate code for input function \param F.
   virtual std::unique_ptr<CompiledFunction> compile(Function *F) const = 0;
+  /// Generate code for input function \param F but do not collect constants.
```
bertmaher: vertical space b/w top level declarations.

lib/Backends/CPU/CPUBackend.h
```diff
@@ -45,7 +45,11 @@ class CPUBackend : public BackendUsingGlowIR {
   std::unique_ptr<CompiledFunction>
   compileIR(std::unique_ptr<IRFunction> IR) const override;
 
+  std::unique_ptr<CompiledFunction>
+  compileIRWithoutConstants(IRFunction *IR) const;
```
bertmaher: vertical space b/w top level declarations.

lib/Backends/CPU/CPUBackend.h
```diff
@@ -45,7 +45,11 @@ class CPUBackend : public BackendUsingGlowIR {
...
   std::unique_ptr<CompiledFunction> compile(Function *F) const override;
+  std::unique_ptr<CompiledFunction>
```
bertmaher: vertical space b/w top level declarations.

lib/Backends/CPU/CPUFunction.h
```diff
@@ -27,14 +27,14 @@ class CPUFunction final : public CompiledFunction {
...
-  /// runtimeBundle contains symbol offsets and allocation sizes.
-  runtime::RuntimeBundle runtimeBundle_;
   /// Base address for Activations memory block.
   uint8_t *baseActivationsAddress_{};
   /// Base address for Mutable weights memory block, Inputs and Outputs.
   uint8_t *baseMutableWeightVarsAddress_{};
 
 public:
+  /// runtimeBundle contains symbol offsets and allocation sizes.
+  runtime::RuntimeBundle runtimeBundle_;
```
bertmaher: I recommend keeping runtimeBundle private; having public data members has a way of bloating the API.  It seems like you should be able to just add a `collectConstants` method to `CPUFunction`.  (Or maybe even to the `CompiledFunction` interface, since it's used by every backend.)

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -29,8 +29,21 @@ std::unique_ptr<CompiledFunction> Interpreter::compile(Function *F) const {
   return compileIR(std::move(IR));
 }
 
+std::unique_ptr<CompiledFunction>
+Interpreter::compileWithoutConstants(Function *F) const {
+  auto IR = generateAndOptimizeIR(F, shouldShareBuffers());
+  return compileIRWithoutConstants(std::move(IR));
+}
```
bertmaher: Blank line between definitions.

lib/Backends/Interpreter/Interpreter.h
```diff
@@ -36,8 +36,12 @@ class Interpreter final : public BackendUsingGlowIR {
 
   std::unique_ptr<CompiledFunction>
   compileIR(std::unique_ptr<IRFunction> IR) const override;
+  std::unique_ptr<CompiledFunction>
+  compileIRWithoutConstants(std::unique_ptr<IRFunction> IR) const;
 
   std::unique_ptr<CompiledFunction> compile(Function *F) const override;
+  std::unique_ptr<CompiledFunction>
```
bertmaher: More blank lines plz :)

lib/Backends/Interpreter/InterpreterFunction.h
```diff
@@ -74,6 +74,8 @@ class InterpreterFunction final : public CompiledFunction {
   void tearDownRuns() override;
 
   void execute() override;
+  /// Get reference to IR function.
+  IRFunction *getIR() { return F_.get(); }
```
bertmaher: As above, I think adding a `collectConstants` would be better than adding `getIR` just to pass it back to runtime bundle.

lib/Backends/BackendUtils.cpp
```diff
@@ -143,6 +143,5 @@ glow::generateRuntimeBundle(const IRFunction &F,
 
   runtime::RuntimeBundle info(symbolTable, constantMaxSize, placeholderMaxSize,
```
rdzhabarov: nit: just return this right away

lib/Backends/CPU/CPUFunction.h
rdzhabarov: > Or maybe even to the CompiledFunction interface, since it's used by every backend.

this way we'd also avoid ugly static_casts on pointers to specific CompiledFunctions.

include/glow/Backends/Backend.h
```diff
@@ -42,6 +42,9 @@ class Backend {
...
+  virtual std::unique_ptr<CompiledFunction>
+  compileWithoutConstants(Function *F) const = 0;
```
rdzhabarov: where is it invoked?

gcatron: Currently nowhere but it will be called from the Provisioner.

lib/Backends/CPU/CPUFunction.h
gcatron: I think adding it as a method to each compiledFunction makes sense, I think eventually this will go away entirely, the new DeviceManager will have the collection logic. Unfortunately even if I added it to CompiledFunction OpenCL and Interpreter will still require a static_cast to get the IR. 

So how about I keep bundle private and add a new method to each CPUFunction, OpenCLFunction, and InterpreterFunction but not to compiledFunction, since ultimately we'd want to move this to the DeviceManager anyway?

lib/Backends/Interpreter/InterpreterFunction.h
gcatron: The challenge here is the CPUFunction, it doesn't keep the IR around so it would still need it passed in.

bertmaher: That is a good point.  Which makes me wonder, how could we even use `collectConstants` with the CPU backend?  Since `compileIR` takes ownership of the IR and then destroys it, how can we later pass it to collect?  (Right now we only collect inside the compile function, but that's going to change soon... right?

gcatron: Yes, the device manager will have a reference to the module so that is where it can collect constants from.

bertmaher: Looking at collectConstants, does it even need to take an IRFunction?  It looks like it pulls all the constants from the Module via `F->getGraph()->getParent()->getConstants()`.  It then uses `getWeightForNode` to turn those back into WeightVars but I don't see a reason for that, since all we do is call `getSizeInBytes`, which we can get without getting the WeightVar.  Am I missing something?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -128,6 +136,12 @@ std::unique_ptr<CompiledFunction> CPUBackend::compile(Function *F) const {
   return compileIR(std::move(IR));
 }
 
+std::unique_ptr<CompiledFunction>
+CPUBackend::compileWithoutConstants(Function *F) const {
+  auto IR = generateAndOptimizeIR(F, shouldShareBuffers());
+  return compileIRWithoutConstants(IR.get());
```
bertmaher: Did we come to a conclusion about how `collectConstants` is going to work for the CPU backend yesterday?  Since `compileWithoutConstants` throws away the IR, we don't have anything to pass to `collectConstants`...

gcatron: I think we decided to make expose collect constants as a method on each compiled function, and keep the bundle_ private. We call collectConstants in CompileIR where we still have the IR.

PR #2249, opened 2019-01-09
DeviceManager interface and CPUDeviceManager

nickgg:
*Description*: A second prototype of the DeviceManager, which is the interface to a single inference Device. The interface supports three operations:

** addNetwork() - load an existing compiled module (or sub module) onto the device ready to be executed.
** evictNetwork() - unload a module from the device and from memory.
** runFunction() - run a Function (subgraph) with the provided context and callback with results.

The DeviceManager operates in it's own thread, which for the CPU backend implemented here also does compilation and running of the function.

*Testing*: new tests, parallelized example in a follow up.
*Documentation*:
Part of the high level Runtime issue #2045.

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -1,0 +1,92 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_DEVICEMANAGER_H
+#define GLOW_BACKENDS_DEVICEMANAGER_H
+
+#include "glow/Backends/Backend.h"
+#include "glow/Backends/CompiledFunction.h"
+#include "glow/Graph/Context.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <functional>
+#include <map>
+#include <string>
+
+namespace glow {
+
+/// Callback signalling success/failure of loading a Module onto a device.
+using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
+/// Callback signalling the result of running a function.
+using ResultCBTy = std::function<void(
+    runtime::RunIdentifierTy, runtime::ResultCode, std::unique_ptr<Context>)>;
+/// Map of Function name -> CompiledFunction, used when loading a network onto a
+/// device.
+using FunctionMapTy = std::map<std::string, CompiledFunction *>;
+
+/// Interface managing a specific instance of a device.
+class DeviceManager {
+protected:
+  /// Type of Backend for this Device.
+  BackendKind backend_;
+
+public:
+  DeviceManager(BackendKind backend) : backend_(backend) {}
+  virtual ~DeviceManager() {}
+
+  /// Initialize the device.
+  virtual void init() {}
+
+  /// Load the provided module into the device, readyCB will be called when
+  /// ready to use.
+  /// \p functions contains the list of functions to load, keyed by their name
+  /// (as used in runFunction).
+  virtual void addNetwork(const Module *module, FunctionMapTy functions,
+                          ReadyCBTy readyCB) = 0;
+
+  /// Remove (and delete) the provided network and all it's functions, freeing
+  /// up space on the device.
+  virtual void evictNetwork(const Module *module) = 0;
+
+  /// Execute the named Function in an already provided network on the device.
+  /// functionName must match the name of a function already added.
+  /// Context should have all Placeholders allocated. resultCB will be called
+  /// with the Context results filled.
+  virtual runtime::RunIdentifierTy runFunction(std::string functionName,
+                                               std::unique_ptr<Context> ctx,
+                                               ResultCBTy resultCB) = 0;
+
+  /// Stops execution and shuts down the Device.
+  virtual void stop(bool block = true) {}
+
+  /// \returns the type of Backend that powers this Device.
+  BackendKind getBackendKind() { return backend_; }
+
+  /// \returns the maximum memory (in bytes) available on the device.
+  virtual uint64_t getMaximumMemory() = 0;
+
+  /// \returns the currently available memory (in bytes) available on the
+  /// device, for provisioning new networks.
+  virtual uint64_t getAvailableMemory() = 0;
+
+  /// \returns true if we expect a Module with the estimated constant size will
+  /// fit on the device.
+  virtual bool isMemoryAvailable(uint64_t estimate) = 0;
+};
+
+} // namespace glow
+
+#endif // GLOW_BACKENDS_DEVICEMANAGER_H
--- a/include/glow/Backends/QueueBackedDeviceManager.h
+++ b/include/glow/Backends/QueueBackedDeviceManager.h
@@ -1,0 +1,78 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_QUEUEBACKEDDEVICEMANAGER_H
+#define GLOW_BACKENDS_QUEUEBACKEDDEVICEMANAGER_H
+
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Support/ThreadPool.h"
+
+#include <atomic>
+
+namespace glow {
+
+class QueueBackedDeviceManager : public DeviceManager {
+protected:
+  /// Thread which interfaces with the device.
+  ThreadPool workThread_;
+
+  /// Identifier for next run.
+  std::atomic<runtime::RunIdentifierTy> nextIdentifier_{1};
+
+public:
+  QueueBackedDeviceManager(BackendKind backend);
+  virtual ~QueueBackedDeviceManager();
+
+  /// Initialize the device.
+  void init() override;
+
+  /// Load the provided module into the device, readyCB will be called when
+  /// ready to use
+  void addNetwork(const Module *module, FunctionMapTy functions,
+                  ReadyCBTy readyCB) override;
+
+  /// Remove (and delete) the provided network and all it's functions, freeing
+  /// up space on the device.
+  void evictNetwork(const Module *module) override;
+
+  /// Execute the named Function in an already provided network on the device.
+  /// functionName must match the name of a function already added.
+  /// Context should have all Placeholders allocated. resultCB will be called
+  /// with the Context results filled.
+  runtime::RunIdentifierTy runFunction(std::string functionName,
+                                       std::unique_ptr<Context> ctx,
+                                       ResultCBTy resultCB) override;
+
+  /// Stops execution and shuts down the Device.
+  void stop(bool block = true) override;
+
+protected:
+  /// Operator handling methods to be implemented in subclasses (i.e. per Device
+  /// type)
+
+  /// Load and compile the Module
+  virtual void addNetworkImpl(const Module *, FunctionMapTy, ReadyCBTy) = 0;
+
+  /// Remove the module and reclaim it's memory
+  virtual void evictNetworkImpl(const Module *) = 0;
+
+  /// Execute provided Function.
+  virtual void runFunctionImpl(runtime::RunIdentifierTy, std::string,
+                               std::unique_ptr<Context>, ResultCBTy) = 0;
+};
+
+} // namespace glow
+
+#endif // GLOW_BACKENDS_QUEUEBACKEDDEVICEMANAGER_H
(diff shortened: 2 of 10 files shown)


opti-mix:
@nickgg Quick question: What happens if someone would call `evictNetwork` while this network is running because it was launched by `runFunction`? Do we need an API to cancel the execution of the network before it finishes on its own?

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,105 @@
...
+
+  /// Remove (and delete) the provided network and all it's functions, freeing
+  /// up space on the device.
+  void evictNetwork(const Module *module);
+
+  /// Execute the named Function in an already provided network on the device.
+  /// functionName must match the name of a function already added.
+  /// Context should have all Placeholders allocated. resultCB will be called
+  /// with the Context results filled.
+  void runFunction(std::string functionName, std::unique_ptr<Context> ctx,
```
opti-mix: Is it possible that the same function is started with different contexts and we have multiple running instances of the function? Do we need to return some kind of unique id from `runFunction`, so that we can distinguish between multiple running instances? This could be useful e.g. for cancellation, profiling, monitoring the execution, etc.

nickgg: To a certain degree the Context pointer itself is that identifier, but if we want something more specific I figured we could put it into the lambda capture of the callback (e.g. how we use it in the tests).

How do you picture using it and the id?

nickgg:
@opti-mix I think we should keep the interface here as simple as possible, I'm not sure that all devices Glow supports will support being able to interrupt a running inference, so we'd be setting ourselves up for failure by supporting it in the DeviceManger layer. Do you know a use case that would require this?

The basic idea here is that we're serializing all calls into a single queue, and so the evict would take place after previous runs but prevent future runs.

include/glow/Backends/DeviceManager.h
opti-mix: My idea was that if we associate an id with the specific run, we could later query the device manager about the current state of the run by providing this id and ask how long this run took, when it was started, when it was finished, if it is still running, etc. We could also use it for cancellation, if we decide that we want to support cancellation.

And yes, I guess the Context pointer could be that id. But I'm not sure we want to guarantee it. Instead we could return an opaque id (which by default could be the Context pointer), which the client only uses to provide as input parameter to those query APIs I described above.

opti-mix: > Do you think it's useful to pass it in to the callback as well?

Yes, it could be useful if we pass it to the callback. May be it wants to query some information about this specific run.

nickgg:
That ASAN failure doesn't seem related, the alloc & overrun both happen within the call to compile(). Does anyone know about this, if not i'll look into it.

tests/unittests/CPUDeviceManagerTest.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+#include "CPUDeviceManager.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+
+#include "gtest/gtest.h"
+
+#include <chrono>
+#include <future>
+
+using namespace glow;
```
nadavrot: Great unit tests!

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,111 @@
...
+#include <map>
+#include <mutex>
+#include <thread>
+#include <unordered_map>
+
+namespace glow {
+
+using DeviceNetworkID = size_t;
+
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
```
nadavrot: ResultCode leaks into the Glow namespace. Is this Device-specific ResultCode or a generic glow ResultCode?

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,111 @@
...
+
+using ReadyCB = std::function<void(const Module *, ResultCode)>;
```
nadavrot: Would calling it ReadyCB ReadyCBTy make the code more readable?

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,111 @@
...
+#ifndef GLOW_BACKENDS_DEVICEMANAGER_H
+#define GLOW_BACKENDS_DEVICEMANAGER_H
+
+#include "glow/Backends/Backend.h"
+#include "glow/Backends/CompiledFunction.h"
+#include "glow/Graph/Context.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Support/ThreadPool.h"
+
+#include <atomic>
```
nadavrot: Can we move some of these includes to the cpp file?

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,111 @@
...
+    std::function<void(RunIdentifier, ResultCode, std::unique_ptr<Context>)>;
+using FunctionMap = std::map<std::string, CompiledFunction *>;
+
+class DeviceManager {
+protected:
+  // Type of Backend for this Device.
+  BackendKind backend_;
+
+  /// Thread which interfaces with the device.
+  ThreadPool workThread_;
```
opti-mix: It would be nice if we could hide the implementation details like `workThread` and `nextIdentifier`, so that only the implementation C++ file sees them. Then we could remove the includes for thread pools and atomics from this header file.

But C++ does not make it easy ;-)

include/glow/Backends/DeviceManager.h
nickgg: The plan, afaik, is to have one single ResultCode enum across the Runtime, will have to coordinate with @gcatron and @SplitInfinity about how this fits into the Provisioner and Executor pieces.

include/glow/Backends/DeviceManager.h
nickgg: We could do something similar to the backend where DeviceManager is an abstract interface and you must create them via a factory function. Hides the implementation details, but adds a new level of indirection. I'm guessing since thats what we do for backends thats fine though.

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -0,0 +1,101 @@
...
+uint64_t CPUDeviceManager::getAvailableMemory() {
+  return maxMemoryBytes - usedMemoryBytes;
+}
+
+bool CPUDeviceManager::isMemoryAvailable(uint64_t estimate) {
+  // No fuzz factor for the CPU device.
+  return maxMemoryBytes >= (usedMemoryBytes + estimate);
+}
+
+void CPUDeviceManager::addNetworkImpl(const Module *module,
```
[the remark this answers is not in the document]
nickgg: We're loading a set of functions (which we expect to share constants, e.g. the same graph with different batch sizes). The `FunctionMap` contains `CompiledFunction`s, but to collect the constants we need access back to the Placeholders which are owned by the module.

include/glow/Backends/DeviceManager.h
bertmaher: Might be nice to make it `enum class ResultCode` so we don't just have bare `READY` or `FAILED` etc. hanging out in the code.

nickgg: This belongs in the RuntimeTypes.h header from #2161, I'll include that in this PR.

include/glow/Backends/DeviceManager.h
nickgg: Actually I'm not sure we can and have those parts be inheritable by subclasses. I can move those members to the CPUDeviceManager which would clean up the DeviceManager header, but then we'd need to copy them if another device wants the ThreadPool, etc. What do you think?

nickgg:
Added RuntimeTypes, fixed some formatting. Didn't attack the header include part yet.

include/glow/Backends/DeviceManager.h
opti-mix: My main concern why I wanted these details to be hidden is: Are we sure that all device managers would always use <thread>-based implementation for their threading? Is it e.g. possible that in some cases a thread pool is provided by Caffe2 or any other outer system that uses Glow as a backend?

The second, less important concern is: Only the API is important. The clients of DeviceManager do not care much about how it functions internally and should not see those details.

> I can move those members to the CPUDeviceManager which would clean up the DeviceManager header, but then we'd need to copy them if another device wants the ThreadPool, etc. 

If you go along this lines, you could define a helper class with common data and logic that can be used by all interested implementations of device managers. And then those concrete DeviceManager implementations could include it as a member or inherit from it.

nickgg: Ah right, yeah thats easy enough. Will do.

nickgg:
Pulled out implementations to a `QueueBackedDeviceManger`helper class.

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,85 @@
...
+protected:
+  // Type of Backend for this Device.
+  BackendKind backend_;
+
+public:
+  DeviceManager(BackendKind backend) : backend_(backend) {}
+  virtual ~DeviceManager() {}
+
+  /// Initialize the device.
+  void init() {}
```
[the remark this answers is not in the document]
nickgg: I don't think we need it for the CPUBackend, but physical devices are likely to need some kind of "first-run" logic which makes sense to do explicitly. It should definitely be virtual though.

include/glow/Backends/QueueBackedDeviceManager.h
```diff
@@ -0,0 +1,81 @@
...
+  /// with the Context results filled.
+  runtime::RunIdentifierTy runFunction(std::string functionName,
+                                       std::unique_ptr<Context> ctx,
+                                       ResultCBTy resultCB);
+
+  /// Stops execution and shuts down the Device.
+  void stop(bool block = true);
+
+  /// \returns the type of Backend that powers this Device.
+  BackendKind getBackendKind() { return backend_; }
```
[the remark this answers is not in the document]
nickgg: It was an ask from @gcatron for use by the provisioner. I don't see the harm, regardless.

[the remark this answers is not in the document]
gcatron: This is used by the provisioner, we could store this information outside the device Manager alongside each instance but this seemed cleaner.

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,85 @@
...
+#include "glow/Graph/Graph.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <functional>
+#include <map>
+#include <string>
+
+namespace glow {
+
+using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
```
gcatron: I think we can drop Module * as a parameter to the CB, If it's purpose is to be an identifier that can be captured in the lambda the caller passes.

nickgg: There's two schools of thought on this PR about whether to rely entirely on the lambda capture or not. Since we have the RunIdentifier for runFunction I think we should have something for AddNetwork as well; perhaps we should bring back DeviceNetworkId?

gcatron: I think I'd prefer to keep the Module * over bringing back DeviceNetworkId, since the function name uniquely identifies the function to the Device Manager already a second identifier seems redundant. I'm good with keeping Module * so we do not have to be reliant on the lambda capture though.

include/glow/Backends/QueueBackedDeviceManager.h
```diff
@@ -0,0 +1,81 @@
...
+  void init();
+
+  /// Load the provided module into the device, readyCB will be called when
+  /// ready to use
+  void addNetwork(const Module *module, FunctionMap functions,
+                  ReadyCBTy readyCB);
+
+  /// Remove (and delete) the provided network and all it's functions, freeing
+  /// up space on the device.
+  void evictNetwork(const Module *module);
```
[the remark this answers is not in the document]
nickgg: We should keep it consistent with addNetwork and the ReadyCB, and since those are used to do comparisons I think better to keep it as pointer (ie. do a ptr compare, not an operator==).

Honestly I think the raw ptrs here should be shared_ptrs.

include/glow/Backends/QueueBackedDeviceManager.h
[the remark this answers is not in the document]
nickgg: Oh you meant that it's already in the DeviceManager and so we don't need it in the QueueBackedDeviceManager also? Good catch.

nickgg:
@rdzhabarov are you satisfied with those explanations or do you want me to change the interface?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+namespace runtime {
+
+using NetworkIDty = size_t;
+using DeviceIDty = size_t;
```
bertmaher: `Ty`, not `ty`

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,68 @@
...
+using RunIdentifierTy = size_t;
+
+/// Enum to communicate results when communicating with device at initialization
+/// and runtime.
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
```
bertmaher: Can you make this an `enum class`, and make them lower case for consistency with the rest of the code base.

nickgg: Ah yes that got lost when i pulled in RuntimeTypes, what do you mean by the second part? Is there a good example of the style you'd like me to match?

bertmaher: All of our other `enum classes` have UpperCamelCase values, e.g. `CompilationMode` has `Train` and `Infer` (rather than `TRAIN` and `INFER`).

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,92 @@
...
+#include <functional>
+#include <map>
+#include <string>
+
+namespace glow {
+
+/// Callback signalling success/failure of loading a Module onto a device.
+using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
+/// Callback signalling the result of running a function.
+using ResultCBTy = std::function<void(
```
rdzhabarov: super nit: why this is Ty and FunctionMap is not FunctionMapTy :-)?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -0,0 +1,68 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+namespace runtime {
+
+using NetworkIDty = size_t;
```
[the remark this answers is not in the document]
nickgg: NetworkID was used to identify the network in addNetwork, but we now use the Module* - i'll remove it.

nickgg:
Fixed remaining comments and will commit, build permitting.

ayermolo:
@nickgg u_int64_t is not c portable, and work with VS.
https://stackoverflow.com/questions/42979055/should-i-use-c-types-uint8-t-uint64-t-or-u-int8-t-u-int64-t

PR #2262, opened 2019-01-11
[quantization] Support int16 loader and custom precision in quantization procedure

rdzhabarov:
*Description*:
* Loader param for quantization precision
* Custom precision in quantization procedure

*Testing*:
* Unit tests (int16 quantization test).

*Documentation*:
Will update quantization doc.

cc: @bertmaher this should help with the automatic quantization you'll be doing for int16.


rdzhabarov:
@qcolombet hehe :) you are fast!

PR #2264, opened 2019-01-12
[RFC] Should we use glog?

bertmaher:
*Description*:
I imagine this change would be controversial, but thought it'd be worth discussing, esp if there's a clean solution already available in LLVM.  We have a lot of code like:
```
GLOW_ASSERT(!thing_failed && "A thing failed");
```
which doesn't tell you anything about the thing that failed (like, what was the unhandled instruction?)  You can get this information from the debugger by re-running, but that's a bit tedious, shouldn't really be necessary, and is not a great option for debugging things that happen in production.

So I often change those lines to:
```
if (thing_failed) {
  llvm::errs() << "A thing failed because: " << reasons;
  GLOW_UNREACHABLE("A thing failed");
}
```
but that is pretty clunky (multiple lines, redundant strings).

With glog, this is pretty clean:
```
CHECK(!thing_failed) << "A thing failed because: " << reasons;
```
 
The down side, of course, is that we add a dependency (albeit one that is practically standard), and the annoyance that glog uses std::ostream whereas llvm uses llvm::raw_ostream, which leads to eyebrow-raising things like the `operator<<` for llvm::StringRef in this sample PR.

Anyways, I *probably* won't spend a ton more time on this in the immediate future, but I'm curious to hear opinions.  Is this a total non-starter?  Is there an elegant way of handling these cases in LLVM?

*Testing*: n/a

*Documentation*: n/a

### DIFF
--- a/lib/Quantization/CMakeLists.txt
+++ b/lib/Quantization/CMakeLists.txt
@@ -1,5 +1,7 @@
 add_subdirectory(Base)
 
+find_package(glog REQUIRED)
+
 add_library(Quantization
               Serialization.cpp
               Quantization.cpp)
@@ -10,4 +12,5 @@ target_link_libraries(Quantization
                         Graph
                         ExecutionEngine
                         QuantizationBase
-                        LLVMSupport)
+                        LLVMSupport
+                        glog::glog)
--- a/lib/Quantization/Serialization.cpp
+++ b/lib/Quantization/Serialization.cpp
@@ -17,10 +17,18 @@
 #include "glow/Quantization/Serialization.h"
 #include "glow/Quantization/Quantization.h"
 
+#include "glog/logging.h"
+
 #include "llvm/Support/FileSystem.h"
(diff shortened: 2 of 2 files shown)


stoklund:
cc #1517 

I tend to use assertions even inside performance-sensitive loops because they're good at finding bugs in debug builds.

This kind of heavy-duty assertion seems closer to the error-reporting domain. It seems like this kind of check could change back and forth between returning an error and aborting the process.

So should we use the same framework for returning errors and heavy-duty assertions?

bertmaher:
@stoklund I'd forgotten about that discussion :-).  But yes, our code base is ambiguous about what `assert` means, since we use it both for bug-checking and for error-handling.  (And the example I give here falls pretty squarely in the error-handling category.)  I still suspect there's a place for better reporting when a fatal error occurs, although maybe it would be a better use of time to vastly reduce the category of things that currently cause fatal errors.

PR #2267, opened 2019-01-15
[WIP][Runtime] Add runtime Executor class

SplitInfinity:
**Description**
graphs using a thread pool. The  workers essentially do a
concurrent breadth-first search through the node graph to execute all
nodes, repeatedly copying partition components outputs to their descendants along
the way.

**Testing**
Tests WIP. I need to write some code to mock DAGs, symbol information and `DeviceManager`; let me know if you've done this already and can share the code.


SplitInfinity:
Overall, I think this is simpler and cleaner than v1, but there is still some iteration and cleanup required. There is still some complicated locking/unlocking going on, and there is probably a cleaner way to split up the state and responsibilities between `ThreadPoolExecutor` and `ExecutionState`. Right now, the latter is a simple contained for all the state related to a single run that allows me to avoid creating several instance variables in `ThreadPoolExecutor` of type `std::unordered_map<RuntimeIdTy, std::unordered_map<DAGNode*, T>>` which is annoying to read and reason about.

bertmaher:
Does this PR include the changes in #2249?  The device manager stuff looks similar...

nickgg:
@bertmaher it's got the interface and runtime types from that PR yes, ideally we land #2249 first and rebase onto it.

SplitInfinity:
Yep, @nickgg is correct. I will rebase on top of that PR after it lands.

include/glow/Backends/BackendUtils.h
```diff
@@ -59,6 +65,8 @@ class RuntimeBundle {
   size_t getValueOffset(const Named *v) const;
   /// Helper function, gets symbol info for \p v.
   RuntimeSymbolInfo getSymbolInfo(const Named *v) const;
+  /// Get a const reference to the symbol table.
```
gcatron: Channeling @bertmaher and since I'm a frequent transgressor, remember vertical whitespace between declarations.

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,106 @@
...
+#include <map>
+#include <mutex>
+#include <thread>
+#include <unordered_map>
+
+namespace glow {
+
+using DeviceNetworkID = size_t;
+
+enum ResultCode { READY, EXECUTED, FAILED, CANCELLED };
```
nadavrot: The ResultCode enum leaks into the global namespace.

include/glow/Backends/BackendUtils.h
```diff
@@ -30,13 +30,19 @@ struct RuntimeSymbolInfo {
   size_t offset;
   /// Type of symbol.
   Type type;
+  /// Is the symbol an input for the function.
+  bool input;
```
nadavrot: input -> isInput

include/glow/Backends/DeviceManager.h
```diff
@@ -0,0 +1,106 @@
...
+#ifndef GLOW_BACKENDS_DEVICEMANAGER_H
+#define GLOW_BACKENDS_DEVICEMANAGER_H
+
+#include "glow/Backends/Backend.h"
+#include "glow/Backends/CompiledFunction.h"
+#include "glow/Graph/Context.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Support/ThreadPool.h"
+
+#include <atomic>
```
nadavrot: Can you move some of the includes from the header file to the cpp file?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,249 @@
...
+ */
+
+#include "ThreadPoolExecutor.h"
+#include "glow/Backends/DeviceManager.h"
+
+namespace glow {
+namespace runtime {
+
+/// Forward declaration of getDeviceManager function.
+/// TODO: Talk to gcatron about the details of this.
```
nadavrot: Please don't leave usernames and todos in the codebase. You can file an issue.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,249 @@
...
+    return;
+  }
+
+  // Lock the lock for the runId.
+  std::unique_lock<std::mutex> executionStateLock(executionStateLockIt->second);
+
+  // Get the execution state for the run.
+  auto executionStateIt = executionStates_.find(runId);
+  if (executionStateIt == executionStates_.end()) {
+    // This should never happen. TODO: Log, assert, something.
```
nickgg: Feels safer to start with an assert than not.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,249 @@
...
+    return;
+  }
+
+  std::shared_ptr<ExecutionState> executionState = executionStateIt->second;
+
+  if (resultCode == ResultCode::CANCELLED || resultCode == ResultCode::FAILED) {
```
nickgg: nit: This seems clearer as `if (resultCode != ResultCode::EXECUTED)`

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,249 @@
...
+      return;
+    }
+
+    // Lock executionStateLock before accessing the shared execution state.
+    executionStateLock.lock();
+
+    if ((executionState->inflightNodes).empty()) {
+      // If there are no nodes inflight, that means all nodes are done. Call
+      // the callback and erase the state information.
+      executionState->cb(runId, ResultCode::EXECUTED,
```
nickgg: Holding a lock while calling the callback is dangerous, this could deadlock if the cb calls back into the Executor.

PR #2268, opened 2019-01-15
[Partitioner] First Graph Partitioning

beicy:
*Description*:
Graph partitioning. Use BFS order to get the initial partitions based on memory restrictions.  (working on the advanced algorithm, which is not included in this PR)

*Testing*:
Added an unittest.
The original function is: 
orig
And the partitions are: 
P1:
p1
P2:
p2
P3:
p3

*Documentation*:
[Optional Fixes #issue]


beicy:
Thanks a lot @bertmaher ! Borrowed some ideas and code from your previous  partition work :)

rdzhabarov:
> (working on the advanced algorithm, which is not included in this PR)

might be worthwhile outline that design in the Github issue.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+/// Adjust the logicalDevice ID for each DAGNode. This happens when \p num (i.e.
+/// the number of DAGNodes) is larger than the number of devices. E.g:
+/// node1(6GB) -> node2(14GB) -> node3(6GB). The memory limitation is 16GB, and
+/// there is only 2 devices.
+void Partitioner::adjustLogicalDevicdID(DAGNode *DAG, int num) {}
+
+/// Current only partition the representive function.
+DAGNode *Partitioner::doPartitioning(Function *F, NodeFunctionMap &mapping) {
+  // The dummy node.
+  DAGNode *DAG = new DAGNode();
```
rdzhabarov: can you return std::unique ptr to avoid any memory leak issues?

these days we should avoid naked `new` with all the support we have in std:: various pointers.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+  }
+}
+
+/// Get the minimal memory requirement (constant) for each op in the function.
+void Partitioner::initOpMemUsage() {
+  memUsage.clear();
+  for (auto &node : F_->getNodes()) {
+    int n = node.getNumInputs();
+    unsigned size = 0;
+    if (node.getKind() != Kinded::Kind::SaveNodeKind) {
```
jackm321: easier to read if we do `if (node.getKind() == Kinded::Kind::SaveNodeKind) {continue;}`

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+}
+
+static BFSLevel getBFSLevel(Function *F) {
+  // Visit graph nodes in BFS order. For each non-storage node, get its level.
+  // Use the preorder to get the roots. Now assume there is only one output op
+  // (i.e. root) now.
+  GraphPreOrderVisitor visitor(*F);
+  Node *node = nullptr;
+  for (auto &N : visitor.getPreOrder()) {
+    if (isa<Storage>(N))
```
jackm321: add brackets here

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+      continue;
+    node = N;
+    break;
```
jackm321: I'm confused, won't this just quit at the first non-storage node? Why can't there be multiple in the first level, is this the save node?

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  BFSLevel bfs;
+  int level = 0;
+  int current = 0;
+  auto newPair = std::make_pair(level, std::vector<Node *>());
+  newPair.second.push_back(node);
+  bfs.levels.push_back(newPair);
```
jackm321: this could just be `bfs.levels.push_back({0, {node}});`

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,134 @@
...
+
+#include <map>
+#include <set>
+#include <string>
+
+namespace glow {
+
+using namespace runtime;
+
+using ModuleList = std::unique_ptr<Module>;
```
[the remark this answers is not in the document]
beicy: Removed. It was from previous design and not used anymore.

lib/Partitioner/Partitioner.cpp
beicy: This is a save node.  The comment mentioned that now we assume there is only one root now. The current models I've met only have 1 output node. But it could be multiple roots as well later.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,134 @@
...
+  using Map = llvm::DenseMap<Node *, Function *>;
+
+  /// Newly-created partitions.
+  FunctionList functions_;
+
+  /// Map of nodes in the original function to their target partition.
+  Map nodeToFunction_;
+
+public:
+  /// Create a new partition \p F. Add it to the map with initial node \p N.
```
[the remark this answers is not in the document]
jackm321: I don't know, it says that in the comment on this line though

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+void Partitioner::selectRepFunc() {
+  auto funcList = module_->getFunctions();
+  for (Function *F : funcList) {
+    size_t size = memSize;
+
+    // The set to keep the placeholders (only for Inputs) whose size is already
+    // calcualted.
+    std::set<std::string> pSet;
+
+    for (auto &node : F->getNodes()) {
```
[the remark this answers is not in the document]
beicy: Actually, a few utility functions (the input one, the memory cost for one arbitrary sub-graph...) should be helpful. But I would prefer separate PRs.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,134 @@
...
+  /// The result of module partitioning.
+  DAGNodeList partitions_;
+
+  /// Total memory request by one module
+  size_t memSize;
+
+  /// The map of each operator and the accronsponding memory size
+  MemUsageMap memUsage;
+
+  bool isSameFunctionFamily() { return true; }
```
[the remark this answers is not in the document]
beicy: Yes. will be implemented later once I got the real example.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+  std::vector<std::pair<int, std::vector<Node *>>> levels;
+  /// A set of visited nodes.
+  std::unordered_set<const Node *> visited;
+};
+
+Partitioner::Partitioner(Module *parent, std::vector<DeviceInfo> &devices)
+    : deviceInfo_(devices) {
+  module_ = parent;
+  // Check if all the functions are in the same structure.
+  if (!isSameFunctionFamily()) {
```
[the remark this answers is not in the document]
beicy: In that case, we don't need this check.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,365 @@
...
+  }
+  memSize = module_->getConstantsSize();
+}
+
+/// Get the represent function : the one with the largest input, and update the
+/// memSize.
+void Partitioner::selectRepFunc() {
+  auto funcList = module_->getFunctions();
+  for (Function *F : funcList) {
+    size_t size = memSize;
```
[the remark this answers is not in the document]
beicy: memSize_ is initialed as all the constants size(which the API is provided already: module_->getConstantsSize();) . But for memory usage, we would like to include the input size as well.  memSize_ will be updated as well. In sum, after the representative function is selected, memSize should the the (input+constaints)

include/glow/Partitioner/Partitioner.h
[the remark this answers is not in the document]
beicy: I see. Forgot to update the comment :P

lib/Partitioner/Partitioner.cpp
beicy: Actually, there is a statement after this condition. I think it is OK to use "!=" here since I don't want to duplicate that statement .

lib/Partitioner/Partitioner.cpp
beicy: @rdzhabarov After discussing with @gcatron, we use shared_ptr since all DAGNodes are in HM.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,125 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_PARTITIONER_PARTITIONER_H
+#define GLOW_PARTITIONER_PARTITIONER_H
+
+#include "glow/Graph/Context.h"
```
nadavrot: Can we use forward declarations and move some of the includes to the cpp file?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,125 @@
...
+  const FunctionList &getFunctions() const { return functions_; }
+
+  /// Map API.
+  Map::iterator find(Node *N) { return nodeToFunction_.find(N); }
+  Map::iterator begin() { return nodeToFunction_.begin(); }
+  Map::iterator end() { return nodeToFunction_.end(); }
+  Function *operator[](Node *n) { return nodeToFunction_[n]; }
+};
+
+/// Partition
```
nadavrot: /// Partitions a function into multiple functions based on A... B... C...

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,125 @@
...
+  /// has the largest memory size.
+  Function *F_;
+
+  /// The cost model related to device.
+  std::vector<DeviceInfo> &deviceInfo_;
+
+  /// The result of module partitioning.
+  DAGNodeList partitions_;
+
+  /// Total memory requested by one module.
```
nadavrot: Size in bytes.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "glow/Partitioner/Partitioner.h"
+#include "glow/Graph/Utils.h"
+
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace glow;
+using llvm::isa;
+
+// Margin for code itself -- should be adjusted later.
```
nadavrot: Stale comment here.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,362 @@
...
+//#define MARGIN 1024 * 128
+
+struct BFSLevel {
```
nadavrot: Doxygen please.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,362 @@
...
+  // printf("the number of partition: %d\n", partitionMap.size());
+
+  std::shared_ptr<DAGNode> DAG = doPartitioning(F_, partitionMap);
+  partitions_.push_back(DAG);
+  // Remove the original function after partitioning.
+  module_->eraseFunction(F_);
+
+  auto funcList = module_->getFunctions();
+  for (Function *F : funcList) {
+    // printf("function name : %s\n", F->getName().data());
```
nadavrot: Printf

lib/Partitioner/Partitioner.cpp
rdzhabarov: Sounds good.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+
+public:
+  /// Create a new partition \p F.
+  void create(Function *F) { functions_.emplace_back(F); }
+
+  /// Add a new Node->Function mapping.
+  void add(Node *N, Function *F) { nodeToFunction_[N] = F; }
+
+  /// Get list of functions contained in this map.
+  const FunctionList &getFunctions() const { return functions_; }
```
rdzhabarov: getPartitions

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Map API.
+  Map::iterator find(Node *N) { return nodeToFunction_.find(N); }
+  Map::iterator begin() { return nodeToFunction_.begin(); }
+  Map::iterator end() { return nodeToFunction_.end(); }
+  Function *operator[](Node *n) { return nodeToFunction_[n]; }
+};
+
+/// Given a module, partitions each of the its functions into multiple ones
+/// based on memory constraints and minimizes the communication cost.
```
rdzhabarov: I'd describe how partitioning is done here as well.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+#include <set>
+#include <string>
+
+namespace glow {
+
+using namespace runtime;
+
+using MemUsageMap = std::unordered_map<Node *, unsigned>;
+
+using DAGNodeList = std::list<std::shared_ptr<runtime::DAGNode>>;
```
rdzhabarov: why list not vector here?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+
+public:
+  /// \p parent is the module which contains the functions need to be divided.
+  /// Here we assume that all the functions in one module belong to a same
+  /// "Function Family", that is, without considerting the "dynamic stuff" (i.e.
+  /// batch size, input/output shape of each op), all the functions are
+  /// identical. The required memory and computation cost for each op can be
+  /// found in Module. The \p devices provides the cost model related to
+  /// devices.
+  Partitioner(Module *parent, std::vector<DeviceInfo> &devices);
```
rdzhabarov: api question: why partitioner can modify devices?

lib/Partitioner/Partitioner.cpp
beicy: Comment added.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Decompose each function in a module and return a list of DAGNodes.
+  DAGNodeList &Partition();
```
rdzhabarov: why do we allow caller to modify internal state of partitions by returning reference to it?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+};
+} // namespace glow
+#endif
```
rdzhabarov: commend on what it ends after `#endif`

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -45,10 +45,10 @@ struct DeviceInfo {
 struct DAGNode {
   /// The children of this node, these are nodes that depend on the current
   /// node.
-  std::vector<DAGNode *> children;
+  std::vector<std::shared_ptr<DAGNode>> children;
```
rdzhabarov: can you rebase this on master.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,358 @@
...
+struct BFSLevel {
+  /// A list of <level, nodelist> with BFS order.
+  std::vector<std::pair<int, std::vector<Node *>>> levels;
+  /// A set of visited nodes.
+  std::unordered_set<const Node *> visited;
+};
+
+Partitioner::Partitioner(Module *parent, std::vector<DeviceInfo> &devices)
+    : deviceInfo_(devices) {
+  module_ = parent;
```
rdzhabarov: initialize in the initializer list.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,134 @@
...
+/// Dependency map type.
+using MapTy = llvm::DenseMap<Module *, ModuleList>;
+
+using DAGNodeList = std::list<runtime::DAGNode *>;
+
+/// Helper structure for building a partition. Records a mapping of nodes in the
+/// original function to destination partitions, along with a list of the
+/// newly-created functions.
+class NodeFunctionMap {
+  using Map = llvm::DenseMap<Node *, Function *>;
```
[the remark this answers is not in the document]
rdzhabarov: why DenseMap here but not in the visited list ` std::unordered_set<const Node *> visited;`.

lib/Partitioner/Partitioner.cpp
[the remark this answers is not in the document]
rdzhabarov: btw, for asserts like that you can do: assert(isSameFunctionFamily() && "not in the same function family")

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,358 @@
...
+  // Check if all the functions are in the same structure.
+  if (!isSameFunctionFamily()) {
+    assert("not in the same function family");
+  }
+  memSize_ = module_->getConstantsSize();
+}
+
+/// Get the representative function (the one with the largest input) and update
+/// the memSize_.
+void Partitioner::selectRepFunc() {
```
rdzhabarov: this should be static and have module as input, this way you can test it properly.

include/glow/Partitioner/Partitioner.h
beicy: Yes, should be vector.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,124 @@
...
+
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <map>
+#include <set>
+#include <string>
+
+namespace glow {
+
+using namespace runtime;
```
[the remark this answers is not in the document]
beicy: The DAGNode is defined in runtime namespace.

include/glow/Partitioner/Partitioner.h
beicy: The current one is just a simple algorithm, still evaluating the next step improvement. Will add the algorithm description later.

include/glow/Partitioner/Partitioner.h
beicy: No. Partitioner won't affect device.

beicy:
After more discussion with @opti-mix and @bertmaher, now inside of struct DAGNode, we use raw pointer. Then we use a DAGNodeList as a container to keep the unique_ptr of each DAGNode.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,128 @@
...
+  const FunctionList &getPartitions() const { return functions_; }
+
+  /// Map API.
+  Map::iterator find(Node *N) { return nodeToFunction_.find(N); }
+  Map::iterator begin() { return nodeToFunction_.begin(); }
+  Map::iterator end() { return nodeToFunction_.end(); }
+  Function *operator[](Node *n) { return nodeToFunction_[n]; }
+};
+
+/// The struct contains all the created DAGNodes.
```
opti-mix: Nice!

It could be a good idea to add some comments here and/or for DAGNode to explain the ownership model used here and who owns what. One of the consequences of this design is that DAGNode cannot outlive the DAGNodeList owning it. The second consequence is that DAGNodes can only refer to the DAGNodes from the same DAGNodeList and they don't need to use any smart pointers for that as all of DAGNodes are owned by the same DAGNodeList.

include/glow/Partitioner/Partitioner.h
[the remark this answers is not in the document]
rdzhabarov: I mean,
namespace glow {
namespace runtime {

... // rest of the code here.

}
}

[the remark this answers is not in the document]
beicy: We have include/glow/Runtime/.. and include/glow/Partitioner/... 
I think Partitioner is more "compiler" level, not "runtime". The namespace should be glow::Partitioner.

include/glow/Partitioner/Partitioner.h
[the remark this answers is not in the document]
nadavrot: I think that xxxxxTy would be a better name, because this is not a map, it is a map type.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+  /// update the memSize.
+  static Function *selectRepFunc(Module *parent, size_t &memSize) {
+    auto funcList = parent->getFunctions();
+    Function *ret = nullptr;
+    for (Function *F : funcList) {
+      size_t size = memSize;
+
+      // The set to keep the placeholders (only for Inputs) whose size is
+      // already calculated.
+      std::set<std::string> pSet;
```
nadavrot: I think that we can save stringRef here.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  return bfs;
+}
+
+/// Assign nodes to partitions and return the mapping.
+NodeToFunctionMap Partitioner::selectPartitions(Function *F,
+                                                unsigned availableMemory) {
+  NodeToFunctionMap mapping;
+  BFSLevel bfs = getBFSLevel(F);
+  unsigned level = bfs.levels.size();
+  std::vector<int>
```
nadavrot: This is formatted funny.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+class Partitioner {
+  /// The module that needs to be decomposed.
+  Module *module_;
+
+  /// The representative function used for partition. We choose the function who
+  /// has the largest memory size.
+  Function *F_;
+
+  /// The cost model related to device.
+  std::vector<DeviceInfo> &deviceInfo_;
```
nadavrot: Why is this a vector? Should the comment say that this is per-device info?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+  size_t memSize_;
+
+  /// The map of each operator and the corresponding memory size.
+  MemUsageMap memUsage_;
+
+  bool isSameFunctionFamily() { return true; }
+
+  /// Get the representative function (the one with the largest input) and
+  /// update the memSize.
+  static Function *selectRepFunc(Module *parent, size_t &memSize) {
```
nadavrot: We should move this to the CPP file.

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+    auto funcList = parent->getFunctions();
+    Function *ret = nullptr;
+    for (Function *F : funcList) {
+      size_t size = memSize;
+
+      // The set to keep the placeholders (only for Inputs) whose size is
+      // already calculated.
+      std::set<llvm::StringRef> pSet;
```
nadavrot: Do we want to save the names of the pointers to the placeholders?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+
+      for (auto &node : F->getNodes()) {
+        int n = node.getNumInputs();
+        if (node.getKind() == Kinded::Kind::SaveNodeKind) {
+          // Special node, the placeholder should be ignored?
```
nadavrot: Can you elaborate on this? Why should we ignore this node? What is special?

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+          continue;
+        }
+        for (int i = 0; i < n; i++) {
+          Placeholder *in =
+              llvm::dyn_cast<Placeholder>(node.getNthInput(i).getNode());
+          if (in && (pSet.find(in->getName()) == pSet.end())) {
```
nadavrot: find -> count()

include/glow/Partitioner/Partitioner.h
```diff
@@ -0,0 +1,164 @@
...
+            auto ty = in->getType();
+            size += ty->getSizeInBytes();
+            pSet.insert(in->getName());
+          }
+        }
+      }
+      if (size > memSize) {
```
nadavrot: Comments.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  /// A list of <level, nodelist> with BFS order.
+  std::vector<std::pair<int, std::vector<Node *>>> levels;
+  /// A set of visited nodes.
+  std::unordered_set<const Node *> visited;
+};
+
+Partitioner::Partitioner(Module *parent, std::vector<DeviceInfo> &devices)
+    : module_(parent), deviceInfo_(devices) {
+  // Check if all the functions are in the same structure.
+  assert(isSameFunctionFamily() && "not in the same function family");
```
nadavrot: I think that we decided not to use the concept of family functions.

lib/Partitioner/Partitioner.cpp
nadavrot: @beicy Use early exit here. 

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  BFSLevel bfs;
+  int level = 0;
+  int current = 0;
+  bfs.levels.push_back({level, {node}});
+  bfs.visited.insert(node);
+  level++;
+  while (current < level) {
+    std::vector<Node *> nList;
```
nadavrot: nList -> nodes?

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  // --- TODO
+
+  return mapping;
+}
+
+/// Adjust the logicalDevice ID for each DAGNode. This happens when \p num (i.e.
+/// the number of DAGNodes) is larger than the number of devices. E.g:
+/// node1(6GB) -> node2(14GB) -> node3(6GB). The memory limitation is 16GB, and
+/// there is only 2 devices.
+void Partitioner::adjustLogicalDeviceID(DAGNode *DAG, int num) {}
```
nadavrot: Should we add something like assert(false && "invalid .. .. ");

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+  llvm::DenseMap<Function *, DAGNode *> funcDAG;
+  for (auto *subF : mapping.getPartitions()) {
+    if (funcDAG.find(subF) == funcDAG.end()) {
+      std::unique_ptr<DAGNode> subDAG(new DAGNode());
+      subDAG->name = subF->getName();
+      subDAG->logicalDevice = logicalID++;
+      funcDAG[subF] = subDAG.get();
+      partitions_.nodes.push_back(std::move(subDAG));
+    }
+    for (auto &N : subF->getNodes()) {
```
nadavrot: comments.

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+      for (int inp = 0, e = N.getNumInputs(); inp < e; inp++) {
+        auto input = N.getNthInput(inp);
+        if (isa<Storage>(input.getNode()))
+          continue;
+
+        auto *inputF = mapping[input.getNode()];
+        if (subF == inputF)
+          continue;
+
+        if (funcDAG.find(inputF) == funcDAG.end()) {
```
nadavrot: comments

nadavrot:
@beicy This PR is looking good. I have a few small comments but I think that we are close.

lib/Partitioner/Partitioner.cpp
jackm321: @nadavrot We are not using function families for partitioning (this definitely makes sense to me) or at all (I'm not clear yet on how this can done)?

nadavrot: The semantics of "family" is not well defined. You can say that some functions are related if they share constants, or something, but these are all things that you can express in the Graph  (with functions that refer to constants).

lib/Partitioner/Partitioner.cpp
nadavrot: Generally speaking, and I am not sure if this applies to this case, I don't like shared_ptr because this is usually an indication that there is no clear ownership of some data structures because the lifetime of the objects are too complicated to analyze. Let's make sure that the design is simple enough for us to use regular pointers and containers with clear and simple lifetime.

beicy: @nadavrot we don't use shared_ptr now.  This is the current design: https://github.com/pytorch/glow/pull/2268#issuecomment-455456113

lib/Partitioner/Partitioner.cpp
```diff
@@ -0,0 +1,324 @@
...
+}
+
+DAGNodeList &Partitioner::Partition() {
+
+  // Find the representive function for running partitioning algrithm.
+  F_ = selectRepFunc(module_, memSize_);
+
+  // Possible minimal k devices for a succesful partitioning
+  // Note: here 2 is for testing;
+  unsigned k = 2; //(memSize_ + MARGIN) / devices[0].availableMemory;
```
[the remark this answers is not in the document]
beicy: For the partitioner, we will try to get the smallest k (which is the return value, not the param). Here 2 is just for testing.

include/glow/Partitioner/Partitioner.h
beicy: The reason we use vector here is maybe later, we are consider heterogeneous devices. Will updated the comment.

include/glow/Partitioner/Partitioner.h
beicy: For "Save" node, it is just create a placeholder for the output. Here, we only consider the input size.

lib/Partitioner/Partitioner.cpp
beicy: Partitioner will return the minimal k partitions. And the provisioner will double check if k fits the devices and give the invalid assert.

lib/Partitioner/Partitioner.cpp
beicy: Added

lib/Partitioner/Partitioner.cpp
beicy: Added.

PR #2269, opened 2019-01-15
[Refactor] Refactoring macro in to template code.

ayermolo:
*Description*: Refactored inlined macros, in to template code. I think it's easier to read, and at least in VS easier to step in to.
*Testing*: Unit Tests.
*Documentation*: NA
[Optional Fixes #issue]

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -2143,25 +2143,23 @@ using enable_if_same_t = std::enable_if_t<std::is_same<T, U>::value, U>;
 FUNCTION_ENABLE_IF_TEMPLATE(AvgPool) * createNode(Function &F, Args... args) {
   return F.createAvgPool(args...);
 }
-
 FUNCTION_ENABLE_IF_TEMPLATE(MaxPool) * createNode(Function &F, Args... args) {
   return F.createMaxPool(args...);
 }
-
 FUNCTION_ENABLE_IF_TEMPLATE(Add)
-*createNode(Function &F, Args... var2) { return F.createAdd(var2...); }
-
+*createNode(Function &F, Args... args) { return F.createAdd(args...); }
 FUNCTION_ENABLE_IF_TEMPLATE(Sub)
-*createNode(Function &F, Args... var2) { return F.createSub(var2...); }
-
+*createNode(Function &F, Args... args) { return F.createSub(args...); }
 FUNCTION_ENABLE_IF_TEMPLATE(Mul)
-*createNode(Function &F, Args... var2) { return F.createMul(var2...); }
+*createNode(Function &F, Args... args) { return F.createMul(args...); }
 FUNCTION_ENABLE_IF_TEMPLATE(Div)
-*createNode(Function &F, Args... var2) { return F.createDiv(var2...); }
+*createNode(Function &F, Args... args) { return F.createDiv(args...); }
 FUNCTION_ENABLE_IF_TEMPLATE(Min)
-*createNode(Function &F, Args... var2) { return F.createMin(var2...); }
+*createNode(Function &F, Args... args) { return F.createMin(args...); }
 FUNCTION_ENABLE_IF_TEMPLATE(Max)
-*createNode(Function &F, Args... var2) { return F.createMax(var2...); }
+*createNode(Function &F, Args... args) { return F.createMax(args...); }
+FUNCTION_ENABLE_IF_TEMPLATE(MatMul)
+*createNode(Function &F, Args... args) { return F.createMatMul(args...); }
 
 /// Sink Rescale down with Pooling node.
 /// PoolingNode(Rescale(X)) -> Rescale(PoolingNode(X)).
@@ -2358,24 +2356,8 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       continue;
     }
 
-    // Combine Rescale down with MatMul node.
-    //   MatMulNode(Rescale(LHS), RHS) -> MatMulNode(LHS', RHS).
-    //   MatMulNode(LHS, Rescale(RHS)) -> MatMulNode(LHS, RHS').
-    //   MatMulNode(Rescale(LHS), Rescale(RHS)) -> MatMulNode(LHS', RHS').
     if (auto *MN = dyn_cast<MatMulNode>(&node)) {
-      if (auto *rescale = dyn_cast<RescaleQuantizedNode>(MN->getLHS())) {
-        auto *newMN = F->createMatMul(MN->getName(), MN->getResult().getType(),
-                                      rescale->getInput(), MN->getRHS());
-        MN->getResult().replaceAllUsesOfWith(newMN);
-        MN = newMN;
-        changed = true;
-      }
-      if (auto *rescale = dyn_cast<RescaleQuantizedNode>(MN->getRHS())) {
-        auto *newMN = F->createMatMul(MN->getName(), MN->getResult().getType(),
-                                      MN->getLHS(), rescale->getInput());
-        MN->getResult().replaceAllUsesOfWith(newMN);
-        changed = true;
-      }
+      changed |= combineDownRescaleToArithmeticNode<MatMulNode>(*F, MN);
       continue;
     }
   }
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -2134,6 +2134,82 @@ static void optimizeQuantization(Function *F) {
   optimizeQuantizedMaxSplat(F);
 }
 
+template <class T, class U>
+using enable_if_same_t = std::enable_if_t<std::is_same<T, U>::value, U>;
+#define FUNCTION_ENABLE_IF_TEMPLATE(NODE_NAME_)                                \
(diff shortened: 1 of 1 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
   optimizeQuantizedMaxSplat(F);
 }
```
nadavrot: Docs please.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
...
 
+#define BOILER_PLATE_TEMPLATE(NODE_NAME_)                                      \
+  template <typename Q = T, typename... ArgsTypes>                             \
+  static boilerPlateResolve<NODE_NAME_##Node, Q> *
+
+template <typename T> struct NodeCreator {
```
nadavrot: @ayermolo The new code that you'e added is completely undocumented.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
...
+  create(Function &F, ArgsTypes... var2) { return F.createSub(var2...); }
+  BOILER_PLATE_TEMPLATE(Mul)
+  create(Function &F, ArgsTypes... var2) { return F.createMul(var2...); }
+  BOILER_PLATE_TEMPLATE(Div)
+  create(Function &F, ArgsTypes... var2) { return F.createDiv(var2...); }
+  BOILER_PLATE_TEMPLATE(Min)
+  create(Function &F, ArgsTypes... var2) { return F.createMin(var2...); }
+  BOILER_PLATE_TEMPLATE(Max)
+  create(Function &F, ArgsTypes... var2) { return F.createMax(var2...); }
+};
```
bertmaher: I don't think the NodeCreator class is necessary.  I'm pretty sure it will work to have a set of free functions:
```
template<typename T, typename ...Args>
enable_if_same_t<T, AvgPoolNode> *create(Function &F, Args... args) {
  return F.createAvgPool(args...);
}
```

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
   optimizeQuantizedMaxSplat(F);
 }
 
+#define BOILER_PLATE_TEMPLATE(NODE_NAME_)                                      \
+  template <typename Q = T, typename... ArgsTypes>                             \
+  static boilerPlateResolve<NODE_NAME_##Node, Q> *
+
+template <typename T> struct NodeCreator {
+  template <typename CmpType, typename InstType>
+  using boilerPlateResolve =
```
bertmaher: I'd like a more descriptive name than `boilerPlateResolve`.  May I suggest `enable_if_same_t`?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
...
+  BOILER_PLATE_TEMPLATE(Sub)
+  create(Function &F, ArgsTypes... var2) { return F.createSub(var2...); }
+  BOILER_PLATE_TEMPLATE(Mul)
+  create(Function &F, ArgsTypes... var2) { return F.createMul(var2...); }
+  BOILER_PLATE_TEMPLATE(Div)
+  create(Function &F, ArgsTypes... var2) { return F.createDiv(var2...); }
+  BOILER_PLATE_TEMPLATE(Min)
+  create(Function &F, ArgsTypes... var2) { return F.createMin(var2...); }
+  BOILER_PLATE_TEMPLATE(Max)
+  create(Function &F, ArgsTypes... var2) { return F.createMax(var2...); }
```
bertmaher: (I'm not sure this is a good idea, but...) If you wanted to really code golf this you could introduce an X-macro
```
#define CREATE(NAME) \
  template<typename T, typename ...Args> \
  enable_if_same_t<T, NAME ## Node> *create(Function &F, Args... args) { \
    return F.create ## NAME(args...); \
  }

CREATE(AvgPool)
...
#undef CREATE
```

(Yeah, ok maybe that's too baroque.)

bertmaher:
I'm also wondering if there's something we could add to ClassGen to cover these cases, to cut down on the template metaprogramming.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,78 @@ static void optimizeQuantization(Function *F) {
   optimizeQuantizedMaxSplat(F);
 }
 
+#define BOILER_PLATE_TEMPLATE(NODE_NAME_)                                      \
```
bertmaher: I think BOILER_PLATE_TEMPLATE is more confusing than the code it replaces.  I'd honestly rather see the substituted code (esp if `boilerPlateResolve` is turned into something descriptive).

lib/Optimizer/GraphOptimizer.cpp
ayermolo: This was added by @rdzhabarov 9 month ago.

lib/Optimizer/GraphOptimizer.cpp
ayermolo: @bertmaher Something like this?
enable_if_t is c++14, but I guess most people should be using compilers that support it by now.

```
template <class T, class U>
using enable_if_same_t = std::enable_if_t<std::is_same<T, U>::value, U>;

#define FUNCTION_ENABLE_IF_TEMPLATE(NODE_NAME_)                                \
  template <class T, typename... Args> enable_if_same_t<T, NODE_NAME_##Node>

FUNCTION_ENABLE_IF_TEMPLATE(AvgPool) * createNode(Function &F, Args... args) {
  return F.createAvgPool(args...);
}
```

lib/Optimizer/GraphOptimizer.cpp
rdzhabarov: hm :) what was added? (I thought `Docs please.` is for comment for the macro BOILER_PLATE_TEMPLATE)

lib/Optimizer/GraphOptimizer.cpp
ayermolo: I left FUNCTION_ENABLE_IF_TEMPLATE because I think it kind of hides away some of the template stuff, and if anyone cares it's just couple of lines above to look at. That being said don't have strong feelings about it. Can replace it with explicit template stuff. Just don't like repetitive code like that.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2257,37 +2321,30 @@ static bool sinkRescaleQuantizedNode(Function *F) {
       continue;
     }
 
-// Combine Rescale down with Arithmetic node.
-//   ArithmeticNode(Rescale(X), Rescale(Y)) -> ArithmeticNode(X, Y).
-//   ArithmeticNode(Rescale(X), Y) -> ArithmeticNode(X, Y).
-//   ArithmeticNode(X, Rescale(Y)) -> ArithmeticNode(X, Y).
-// Apply this optimization for Add, Sub, Mul, Div, Min, Max.
-#define COMBINE_DOWN_RESCALE_TO_ARITHMETIC_NODE(NODE_NAME_)                    \
```
rdzhabarov: what about MatMul node below, this can be handled the same way.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2134,6 +2134,82 @@ static void optimizeQuantization(Function *F) {
...
+FUNCTION_ENABLE_IF_TEMPLATE(AvgPool) * createNode(Function &F, Args... args) {
+  return F.createAvgPool(args...);
+}
+
+FUNCTION_ENABLE_IF_TEMPLATE(MaxPool) * createNode(Function &F, Args... args) {
+  return F.createMaxPool(args...);
+}
+
+FUNCTION_ENABLE_IF_TEMPLATE(Add)
+*createNode(Function &F, Args... var2) { return F.createAdd(var2...); }
```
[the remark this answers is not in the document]
ayermolo: Oops missed it. I think I was trying something. Changed

lib/Optimizer/GraphOptimizer.cpp
ayermolo: Ah yeah. Changed.

lib/Optimizer/GraphOptimizer.cpp
jfix71: @ayermolo I believe @nadavrot still wanted some comments here on `FUNCTION_ENABLE_IF_TEMPLATE`

rdzhabarov:
Great! Thank you @ayermolo working through all comments.

lib/Optimizer/GraphOptimizer.cpp
ayermolo: @nadavrot Ooops, missed it. Would  you like me to add comment in separate pull request?

bertmaher:
Nice work @ayermolo :-)

PR #2270, opened 2019-01-16
[quantization] Refine quantization condition.

rdzhabarov:
*Description*:

Given:
* Backends typically should support SaveNodes as it's just a copy instruction.
* On ONNXIFI level there is a call to all nodes whether they are supported or not. Save node will be present in all graphs/networks and should be supported

This PR adds special handling to the SaveNode as it should not be converted to the quantized version.

*Testing*:
Added test.

*Documentation*:
N/A

cc: @bertmaher that might help with the juggle with isOpSupported you mentioned.


lib/Quantization/Quantization.cpp
```diff
@@ -112,7 +112,13 @@ class FunctionQuantizer : public FunctionConverter {
...
   bool canConvert(const Node &node) const override {
     auto kind = node.getKind();
-    if (!EE_.isOpSupported(kind, quantizationPrecision_)) {
+
+    // Do not quantize node if target quantization precision is not supported
+    // by the backend or it's a SaveNode.
+    // SaveNode is typically a copy instruction, so it should be supported by
+    // the backend, but we do not want to quantize it.
+    if (!EE_.isOpSupported(kind, quantizationPrecision_) ||
+        kind == Kinded::Kind::SaveNodeKind) {
```
jfix71: `TypeAToTypeBFunctionConverter::canConvert()` calls into `FunctionConverter::canConvert()` which returns false for `SaveNode` and `Placeholder`. I wonder why we weren't doing that here already. Is that a preferable solution to this, at least for consistency's sake? Otherwise perhaps we should change `canConvert()` to pure virtual.

tests/unittests/QuantizationTest.cpp
```diff
@@ -839,7 +839,8 @@ class MockQuantBackend : public Backend {
   }
   bool isOpSupported(Kinded::Kind opKind, ElemKind elementTy) const override {
     if (opKind == Kinded::Kind::SoftMaxNodeKind ||
-        opKind == Kinded::Kind::LocalResponseNormalizationNodeKind) {
+        opKind == Kinded::Kind::LocalResponseNormalizationNodeKind ||
+        opKind == Kinded::Kind::SaveNodeKind) {
```
jfix71: Why was this added? Is it just to make sure that the check in `canConvert()` overrides the preferences of the backend's `isOpSupported()`? If so perhaps a check should be added in the `quantizeSoftmaxAndLRN` test that checks that the Save isn't quantized?

rdzhabarov: All backends need to support SaveNode, I wanted to make sure that quantization procedure handles Save node well.

rdzhabarov:
```
If so perhaps a check should be added in the quantizeSoftmaxAndLRN test that checks that the Save isn't quantized?
```

Additional checks will be useful.

jfix71: >All backends need to support SaveNode, I wanted to make sure that quantization procedure handles Save node well.

The `MockQuantBackend` calls into `Interpreter::isOpSupported()`, so I think it makes more sense to add `SaveNode` there to return true, no?

Though I guess this is kind of moot as https://github.com/pytorch/glow/pull/2235 redesigns `Interpreter::isOpSupported()` anyway. Note that once that lands `Interpreter::isOpSupported()` will return true for `SaveNode`. If you want to leave it this way it's fine and I'll remove it in https://github.com/pytorch/glow/pull/2235 later! 😃

rdzhabarov: Having check here will make sure that no matter what backend we use, SaveNode is not quantized, which what we want to check.

Yeah, I saw that PR and do not want to mess with it :)

jfix71: Well, the `MockQuantBackend` has hardcoded its "actual backend" as the Interpreter. But whatever it's NBD!

PR #2272, opened 2019-01-16
Fixes for small code and comment typos

jackm321:
*Description*:
Fix a couple of small typos in code and comments
*Testing*:
`ninja test`
*Documentation*:
n/a


lib/Quantization/Quantization.cpp
```diff
@@ -163,10 +163,11 @@ class FunctionQuantizer : public FunctionConverter {
   Node *createConversion(Function &function, NodeValue &val,
                          TypeRef destTy) override {
     if (destTy->isQuantizedType()) {
-      return function_.createQuantize("quantize", val, destTy);
+      return function.createQuantize("quantize", val, destTy);
```
jackm321: For this one, I'm guessing the `function` parameter was intended but I'm not sure.

lib/Quantization/Quantization.cpp
```diff
@@ -163,10 +163,11 @@ class FunctionQuantizer : public FunctionConverter {
...
     }
-    assert(destTy->getElementType() == ElemKind::FloatTy && "");
-    return function.createDequantize("quantize", val);
+    assert(destTy->getElementType() == ElemKind::FloatTy &&
+           "Can't dequantize to any type except float.");
```
rdzhabarov: Can you add explicitly that it's fp32, we also support fp16, but not for dequantize.

lib/Quantization/Quantization.cpp
rdzhabarov: hm, let me see :-) interesting find here.

jfix71: Nice catch!

Is it possible to add a test to check this?

rdzhabarov: Also, this seems to be a candidate to be a static method.

jackm321: @jfix71 I'm not actually sure how to test this, I'm not very familiar with this part of the code yet.
@rdzhabarov do you know if we could actually just remove the function parameter to this createConversion() and use `function_` instead? It seems like the whole functionConverter thing only works on one Function at a time so I don't know if it really makes sense to be able to pass in another Function ever. As for static method, I think you can't have a static method that also overridden.

rdzhabarov: You can easily test if you make that static ;)

You could remove `function` param, but it's not going to be as generic as it's now.
I'd vote to make that static, pass function and add a unit test.

nadavrot:
@jackm321 Please don't keep accepted PR open for a long time. Either land them or close them.

PR #2274, opened 2019-01-16
[RFC] Refactor CPUFunction and InterpreterFunction to remove per-run state

nickgg:
*Description*: To support running a compiled function multiple times, particularly concurrently on different devices, we must remove per-run state from the CompiledFunction. 

This is a suggested solution for the CPU and Interpreter backend:
* For the CPUFunction: we stored the contiguous buffers for activations and weights and so could only have one concurrent run. These buffers need live only as long as the execution, so have moved into the scope of the execute() method. This means also delaying filling of those buffers until execute() as well.
* For the InterpreterFunction: We use various Tensor objects rather than a single weights/activation buffer. These Tensors were stored in the InterpreterFunction itself, which meant concurrent runs would overwrite intermediate values. The sensible place for these is the Context, so I've added the ability to store Tensors keyed by name to the Context and removed the three Tensor maps in InterpreterFunction. This is a general interface, but I'm not sure if it will be useful outside of the interpreter.

The side effect of these changes is to make the non execute() members of CompiledFunction (setupRuns/tearDownRuns and beforeRun/afterRun) empty. The multi-stage execution flow is inherently stateful so I think we should remove them for all backends and move their logic to the various DeviceManagers.

*Testing*: Unit tests in debug, release & asan.
*Documentation*:  Will need to update, but interested in peoples thoughts.

### DIFF
--- a/include/glow/Backends/CompiledFunction.h
+++ b/include/glow/Backends/CompiledFunction.h
@@ -38,21 +38,21 @@ public:
   virtual ~CompiledFunction() = default;
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
-  virtual void execute() = 0;
+  virtual void execute(Context *ctx) = 0;
 
   /// Does any needed initialization work for the Backend.
   /// This includes device init constant memory allocation and copying to
-  /// device.
-  virtual void setupRuns() = 0;
+  /// device. \deprecated
+  virtual void setupRuns() { runsSetup_ = true; }
 
-  /// Per run setup. Copy inputs to device.
-  virtual void beforeRun(const Context &ctx) = 0;
+  /// Per run setup. Copy inputs to device. \deprecated
+  virtual void beforeRun(const Context &ctx) {}
 
-  /// Per run cleanup. Copy outputs from device.
-  virtual void afterRun(const Context &ctx) = 0;
+  /// Per run cleanup. Copy outputs from device. \deprecated
+  virtual void afterRun(const Context &ctx) {}
 
-  /// Final cleanup. Release memory, reset device.
-  virtual void tearDownRuns() = 0;
+  /// Final cleanup. Release memory, reset device. \deprecated
+  virtual void tearDownRuns() { runsSetup_ = false; }
 
   /// Getter for the runtimeBundle.
   const runtime::RuntimeBundle &getRuntimeBundle() const {
--- a/lib/Backends/CPU/CPUFunction.cpp
+++ b/lib/Backends/CPU/CPUFunction.cpp
@@ -30,42 +30,29 @@ CPUFunction::~CPUFunction() {
   tearDownRuns();
 }
 
-void CPUFunction::setupRuns() {
-  if (!runsSetup_) {
-    if (runtimeBundle_.getActivationsSize() != 0) {
-      baseActivationsAddress_ = (uint8_t *)alignedAlloc(
-          runtimeBundle_.getActivationsSize(), TensorAlignment);
-    }
-
-    if (runtimeBundle_.getMutableWeightSize() != 0) {
-      baseMutableWeightVarsAddress_ = (uint8_t *)alignedAlloc(
-          runtimeBundle_.getMutableWeightSize(), TensorAlignment);
-    }
-    runsSetup_ = true;
-  }
-}
-
 void CPUFunction::collectConstants(IRFunction *F) {
   runtimeBundle_.collectConstants(F);
 }
 
-void CPUFunction::beforeRun(const Context &ctx) {
+void CPUFunction::loadPlaceholders(Context *ctx,
+                                   uint8_t *baseMutableWeightVarsAddress) {
   // Copy Placeholders into allocated memory.
-  for (auto PH : ctx.pairs()) {
+  for (auto PH : ctx->pairs()) {
     auto payload = PH.second->getUnsafePtr();
     auto symbolInfo = runtimeBundle_.getSymbolInfo(PH.first);
     auto addr = symbolInfo.offset;
     auto numBytes = symbolInfo.size;
     // copy PH to allocated memory.
-    memcpy(baseMutableWeightVarsAddress_ + addr, payload, numBytes);
+    memcpy(baseMutableWeightVarsAddress + addr, payload, numBytes);
   }
 }
 
-void CPUFunction::afterRun(const Context &ctx) {
+void CPUFunction::updatePlaceholders(Context *ctx,
+                                     uint8_t *baseMutableWeightVarsAddress) {
   // Copy placeholders from device back into context.
-  for (auto PH : ctx.pairs()) {
+  for (auto PH : ctx->pairs()) {
     auto symbolInfo = runtimeBundle_.getSymbolInfo(PH.first);
-    auto payload = baseMutableWeightVarsAddress_ + symbolInfo.offset;
+    auto payload = baseMutableWeightVarsAddress + symbolInfo.offset;
     auto numBytes = symbolInfo.size;
     auto addr = PH.second->getUnsafePtr();
     // copy PH from allocated memory.
@@ -73,20 +60,25 @@ void CPUFunction::afterRun(const Context &ctx) {
   }
 }
 
-void CPUFunction::tearDownRuns() {
-  if (baseMutableWeightVarsAddress_) {
-    alignedFree(baseMutableWeightVarsAddress_);
-    baseMutableWeightVarsAddress_ = nullptr;
+void CPUFunction::execute(Context *ctx) {
+  /// Base address for Activations memory block.
+  uint8_t *baseActivationsAddress{nullptr};
+
+  /// Base address for Mutable weights memory block, Inputs and Outputs.
+  uint8_t *baseMutableWeightVarsAddress{nullptr};
+
+  if (runtimeBundle_.getActivationsSize() != 0) {
+    baseActivationsAddress = (uint8_t *)alignedAlloc(
+        runtimeBundle_.getActivationsSize(), TensorAlignment);
   }
 
-  if (baseActivationsAddress_) {
-    alignedFree(baseActivationsAddress_);
-    baseActivationsAddress_ = nullptr;
+  if (runtimeBundle_.getMutableWeightSize() != 0) {
+    baseMutableWeightVarsAddress = (uint8_t *)alignedAlloc(
+        runtimeBundle_.getMutableWeightSize(), TensorAlignment);
   }
-  runsSetup_ = false;
-}
 
-void CPUFunction::execute() {
+  loadPlaceholders(ctx, baseMutableWeightVarsAddress);
+
   auto sym = JIT_->findSymbol("jitmain");
   assert(sym && "Unable to JIT the code!");
   using JitFuncType =
@@ -95,9 +87,14 @@ void CPUFunction::execute() {
   auto address = sym.getAddress();
   if (address) {
     JitFuncType funcPtr = reinterpret_cast<JitFuncType>(address.get());
-    funcPtr(runtimeBundle_.getConstants(), baseMutableWeightVarsAddress_,
-            baseActivationsAddress_);
+    funcPtr(runtimeBundle_.getConstants(), baseMutableWeightVarsAddress,
+            baseActivationsAddress);
   } else {
     GLOW_ASSERT(false && "Error getting address.");
   }
+
+  updatePlaceholders(ctx, baseMutableWeightVarsAddress);
+
+  alignedFree(baseMutableWeightVarsAddress);
+  alignedFree(baseActivationsAddress);
 }
--- a/lib/Backends/CPU/CPUFunction.h
+++ b/lib/Backends/CPU/CPUFunction.h
@@ -28,12 +28,6 @@ class CPUFunction final : public CompiledFunction {
   /// initializes the LLVM backends.
   std::unique_ptr<llvm::orc::GlowJIT> JIT_;
 
-  /// Base address for Activations memory block.
-  uint8_t *baseActivationsAddress_{};
-
-  /// Base address for Mutable weights memory block, Inputs and Outputs.
-  uint8_t *baseMutableWeightVarsAddress_{};
-
 public:
   /// Ctor.
   CPUFunction(std::unique_ptr<llvm::orc::GlowJIT> JIT,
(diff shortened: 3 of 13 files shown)


lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -30,74 +30,72 @@ CPUFunction::~CPUFunction() {
...
 }
 
 void CPUFunction::collectConstants(IRFunction *F) {
   runtimeBundle_.collectConstants(F);
 }
 
-void CPUFunction::beforeRun(const Context &ctx) {
+void CPUFunction::loadPlaceholders(Context *ctx,
+                                   uint8_t *baseMutableWeightVarsAddress) {
+>>>>>>> 53d40a72... Refactor CompiledFunction to remove state
```
opti-mix: Stale diff comment.

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -43,115 +37,74 @@ void InterpreterFunction::collectConstants(IRFunction *F) {
...
+      }
 
-  auto *T = new Tensor();
-  *T = getTensor(src)->getUnowned(v->dims(), offsets);
-  tensors_[v] = T;
-  return T;
-}
+      auto symbolInfo = runtimeBundle_.getSymbolInfo(v);
+      auto addr = runtimeBundle_.getConstants() + symbolInfo.offset;
+      auto tensor = new Tensor(addr, &symbolInfo.type);
```
opti-mix: Maybe add a comment that this creates an unowned tensor.

include/glow/Backends/CompiledFunction.h
```diff
@@ -38,21 +38,21 @@ class CompiledFunction {
...
 
-  /// Per run cleanup. Copy outputs from device.
-  virtual void afterRun(const Context &ctx) = 0;
+  /// DEPRECATED. Per run cleanup. Copy outputs from device.
+  virtual void afterRun(const Context &ctx) {}
 
-  /// Final cleanup. Release memory, reset device.
-  virtual void tearDownRuns() = 0;
+  /// DEPRECATED> Final cleanup. Release memory, reset device.
+  virtual void tearDownRuns() { runsSetup_ = false; }
```
rdzhabarov: the comment below says: /// Flag to ensure setupRuns is only called once.

is it still the case, if so why we flip it back to false.

include/glow/Graph/Context.h
```diff
@@ -37,19 +38,43 @@ class Context final {
 public:
   /// Maps placeholders to the tensors that back them.
   using PlaceholderMap = std::unordered_map<Placeholder *, Tensor *>;
+  /// Additional Tensors used for this graph, and whether the memory is owned by
```
rdzhabarov: can you clarify what those "additional" tensors are.

include/glow/Graph/Context.h
```diff
@@ -37,19 +38,43 @@ class Context final {
...
 
 public:
   /// \returns the tensor that corresponds to Placeholder \p P or Null if the
   /// tensor is not found.
-  Tensor *get(Placeholder *P) const;
+  Tensor *get(const Placeholder *P) const;
+
+  /// \returns the tensor that corresponds to the name \name, or Null if the
+  /// tensor is not found.
+  Tensor *get(std::string name) const;
```
rdzhabarov: llvm::StringRef

include/glow/Graph/Context.h
```diff
@@ -37,19 +38,43 @@ class Context final {
...
 
   /// Inserts the Placeholder-Tensor pair.
   void insert(Placeholder *P, Tensor &&T);
 
+  /// Inserts the Placeholder-Tensor pair.
+  void insert(Placeholder *P, Tensor *T);
```
rdzhabarov: can you clarify who owns `Tensor *T` in this case. Not clear from API.

include/glow/Graph/Context.h
```diff
@@ -37,19 +38,43 @@ class Context final {
...
+  void registerTensor(const Named *N, Tensor *T, bool transferOwnership);
+
+  /// Attaches the tensor to the given name \p name.
+  void registerTensor(std::string name, Tensor *T, bool transferOwnership);
+
+  /// Detaches the Tensor of the given name and frees memory if owned.
+  void deleteTensor(const Named *N);
+
+  /// Detaches the Tensor of the given name and frees memory if owned.
+  void deleteTensor(std::string name);
```
rdzhabarov: ditto

lib/Graph/Context.cpp
```diff
@@ -20,47 +20,125 @@
...
 }
 
 size_t Context::count(Placeholder *P) const { return map_.count(P); }
 
 void Context::clear() {
   // Delete all of the tensors that are owned by the context.
-  for (auto PH : map_) {
-    delete PH.second;
+  for (auto TM : tensorLookup_) {
+    if (TM.second.second == true) {
```
rdzhabarov: TM.second.second is a bit hard to read

tests/unittests/CPUDeviceManagerTest.cpp
```diff
@@ -295,7 +295,7 @@ TEST(CPUDeviceManagerTest, MultiModule) {
   updateInputPlaceholders(*ctx1, {module1->getPlaceholderByName("input")},
                           {&inputs});
 
-  std::unique_ptr<Context> ctx2 = std::make_unique<Context>(ctx1->clone());
+  std::unique_ptr<Context> ctx2 = std::make_unique<Context>();
```
rdzhabarov: why change here?

tests/unittests/GraphTest.cpp
```diff
@@ -1324,6 +1324,58 @@ TEST(Graph, cloneContextRuns) {
   EXPECT_FALSE(saveBacking1->isEqual(*saveBacking2));
 }
 
+TEST(Graph, ContextDoubleAllocate) {
```
rdzhabarov: Graph test is pretty giant now, maybe extract context specific tests into a separate file.

include/glow/Backends/CompiledFunction.h
nadavrot: @nickgg You can use the deprecated attribute, this will issue a warning:

https://clang.llvm.org/docs/AttributeReference.html#deprecated

lib/Backends/Interpreter/InterpreterFunction.h
```diff
@@ -60,166 +54,164 @@ class InterpreterFunction final : public CompiledFunction {
...
+                                   Value *filterV, Value *biasV,
                                    llvm::ArrayRef<unsigned_t> kernelSizes,
                                    llvm::ArrayRef<unsigned_t> strides,
                                    llvm::ArrayRef<unsigned_t> pads,
                                    size_t group);
 
-  void fwdAvgPoolInstI8Impl(const AvgPoolInst *I);
-  template <typename ElemTy> void fwdAvgPoolInstFloatImpl(const AvgPoolInst *I);
-  template <typename ElemTy> void fwdSoftMaxInstImpl(const SoftMaxInst *I);
+  void fwdAvgPoolInstI8Impl(const AvgPoolInst *I, Context *ctx);
```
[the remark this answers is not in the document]
nickgg: Definitely agree that would be less invasive, but I think also it will be less clear about what's actually happening. I'll put up another commit with an implementation that looks like that and we can see.

include/glow/Backends/CompiledFunction.h
nickgg: @rdzhabarov This is how the existing code works, I suppose the idea is that the lifetime of a "Run" is different from the lifetime of the CompiledFunction. Ultimately the best solution is to remove these functions entirely once all device types are updated.

include/glow/Graph/Context.h
nickgg: I was trying to keep it generic at the Context level, as it could be different for different backends. For the Interpreter, the running of a network appears to require Tensors that aren't linked to existing Placeholders. Haven't looked into why exactly, but guessing it's intermediate tensors.

include/glow/Graph/Context.h
nickgg: ditto for which comment? the StringRef one or the ownership question?

include/glow/Backends/CompiledFunction.h
nickgg: @nadavrot Is a warning what we want now since we haven't updated the CompiledFunctions for all devices yet?

tests/unittests/CPUDeviceManagerTest.cpp
nickgg: So by keying the tensor with a string we have to deal with duplicate keys, in this case there are two models in the test but all their placeholders have the same name. Creating ctx2 as a clone of ctx1 copies across Tensors for the placeholders in the first model, then when we allocate the second model's tensors we hit this collision (which is being handled via assert). Cloning the context was never an important part of this test anyway, so removing it doesn't affect the goal of the test.

include/glow/Graph/Context.h
rdzhabarov: :P StringRef

nickgg:
Refactored based on @opti-mix's suggestion.

nickgg:
@bertmaher That device preparation stuff should happen in the DeviceManager, I think. E.g. for the case of moving constants to the device the DeviceManager should do it in the `addNetwork()` call.

nickgg:
Worth calling out that this PR changes the `CompiledFunction`interface, by adding the `Context *`argument to `execute(). We'll need to update all backends.

nickgg:
Can I get a review on this? I've got some follow ups piling up.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -95,9 +87,14 @@ void CPUFunction::execute() {
...
-            baseActivationsAddress_);
+    funcPtr(runtimeBundle_.getConstants(), baseMutableWeightVarsAddress,
+            baseActivationsAddress);
   } else {
     GLOW_ASSERT(false && "Error getting address.");
   }
+
+  updatePlaceholders(ctx, baseMutableWeightVarsAddress);
+
+  alignedFree(baseMutableWeightVarsAddress);
```
rdzhabarov: this should be fine for now (alloc and dealloc every inference request). But technically we could store this in thread local and reuse buffers.

include/glow/Backends/CompiledFunction.h
```diff
@@ -38,21 +38,21 @@ class CompiledFunction {
   virtual ~CompiledFunction() = default;
   /// Execute the network and allocate Placeholder memory with given
   /// \p ctx providing mapping between Placeholder and populated tensor.
```
rdzhabarov: magic :) comment about ctx was already in place

lib/Backends/CPU/CPUFunction.h
```diff
@@ -42,24 +36,19 @@ class CPUFunction final : public CompiledFunction {
...
-  void tearDownRuns() override;
-
   /// \name CompiledFunction interface
   ///@{
   ~CPUFunction() override;
-  void execute() override;
+  void execute(Context *ctx) override;
   ///@}
+private:
+  /// Load constant tensors from \p ctx into \p weightsAddress, as defined by
```
rdzhabarov: hm, I'm a bit confused. Placeholders are for inputs/outputs but not for constant tensors.

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -22,29 +22,26 @@
 
 #include "llvm/Support/Casting.h"
 
+#include "llvm/Support/raw_ostream.h"
```
rdzhabarov: not used.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -621,7 +621,9 @@ static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
   }
 }
 
-void OpenCLFunction::execute() {
+void OpenCLFunction::execute(Context *ctx) {
```
rdzhabarov: you can just do: execute (Context *) and remove (void)ctx.

lib/Backends/CPU/CPUFunction.cpp
```diff
@@ -30,63 +30,55 @@ CPUFunction::~CPUFunction() {
...
     memcpy(addr, payload, numBytes);
   }
 }
 
-void CPUFunction::tearDownRuns() {
-  if (baseMutableWeightVarsAddress_) {
-    alignedFree(baseMutableWeightVarsAddress_);
-    baseMutableWeightVarsAddress_ = nullptr;
+void CPUFunction::execute(Context *ctx) {
+  /// Base address for Activations memory block.
```
rdzhabarov: i'd remove comment, does not provide any additional info on top of the var name.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -96,7 +96,7 @@ class OpenCLFunction final : public CompiledFunction {
   ///@{
   ~OpenCLFunction() override;
 
-  void execute() override;
+  void execute(Context *ctx) override;
   ///@}
```
rdzhabarov: not related to PR but ///@? needs to be closed after `void tearDownRuns() override;`

lib/Backends/CPU/CPUFunction.cpp
nickgg: Yeah, if we run into CPU backend perf concerns we should add a memory pool here.

lib/Backends/CPU/CPUFunction.h
nickgg: I'm maintaining existing behaviour of CPUFunction in this diff, which is that all constants & placeholders have their space allocated in the RuntimeBundle and then we copy them into the per-run memory block for execution. The memory should be uninitialized so we don't need to memcpy it, but figured we could fix that when we get to it. It is a known issue with the RuntimeBundle.

lib/Backends/OpenCL/OpenCL.cpp
nickgg: heh, @opti-mix has a comment asking for the reverse above. Personally, not worried either way.

rdzhabarov: ok, does not matter indeed.

nickgg:
ah damn I pushed but didn't add the changes, i'll get em in the next one

rdzhabarov:
> ah damn I pushed but didn't add the changes, i'll get em in the next one

sounds good

PR #2276, opened 2019-01-17
Add provisioner for new Runtime

gcatron:
*Description*: This PR adds the Provisioner for the new runtime. This is an initial provisioner so it's rather straight forward. Later work can be done to improve the provisioning algorithm. This is almost ready, it needs to be updated to the new DAGNode when it get's merged from #2268 .
*Testing*: Added new unittest which uses the CPU backend and CPUDeviceManager.
*Documentation*: N/A
Related to #2045


include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -0,0 +1,48 @@
...
+namespace glow {
+namespace runtime {
+
+/// The Provisioner is responsible for assigning networks to an actual device.
+/// It is a stateless class, relying on information being passed in by the
+/// caller.
+class Provisioner final {
+public:
+  /// Walks \p networks and assigns each module to a DeviceManager in \p
+  /// devices. The Provisioner calls the addNetwork method for each
```
beicy: "assigns each module" -- assigns each function or DAGNode ?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -0,0 +1,108 @@
...
+    // will return an error.
+    // TODO Add ability to try against another device when currDevice has
+    // insufficient space.
+    auto currDevice = devices.begin();
+    // Set backend to match the device.
+    backend_.reset(createBackend(currDevice->second->getBackendKind()));
+    // Iterate over the nodes, compile them and add them to compiledFunctions.
+    for (auto node : nodes) {
+      node->deviceID = currDevice->first;
+      Function *function = module.getFunction(node->name);
```
beicy: In struct DAGNode, there is one field called "logicalDevice". The DAGNode with the same logicalDevice should be assigned to the same device according to our previous discussion. (e.g. n1(8GB)->n2(12GB) -> n3(5GB), if we only have 2 devices, n1 and n3 will be assigned the same logicalDevice) I don't see where Provisioner uses this item. Are you going to add this function in Provisioner?

include/glow/Runtime/Provisioner/Provisioner.h
gcatron: Good catch should be function.

lib/Runtime/Provisioner/Provisioner.cpp
gcatron: This current provisioner does not use that, but it will be added to a more robust provisioning algorithm.

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -0,0 +1,48 @@
...
+/// caller.
+class Provisioner final {
+public:
+  /// Walks \p networks and assigns each function to a DeviceManager in \p
+  /// devices. The Provisioner calls the addNetwork method for each
+  /// DeviceManager and uses the returned networkID to populate \p runDAG.
+  /// Returns a ResultCode indicating if the operation was a success.
+  ResultCode
+  provision(std::vector<DAGNode> &networks,
+            std::map<DeviceIDTy, std::unique_ptr<DeviceManager>> &devices,
```
opti-mix: Minor nit: Is it important that devices are a sorted map?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -0,0 +1,108 @@
...
+      auto compiled = backend_->compile(function);
+      node->runtimeBundle = compiled->getRuntimeBundle();
+      compiledFunctions.emplace(node->name, compiled.get());
+    }
+    // Check if sufficient space on device. Currently requiring a 10% buffer
+    // over the size of constants.
+    auto availableMemory = currDevice->second->getAvailableMemory();
+    if (availableMemory <
+        1.1 * nodes[0]->runtimeBundle.getConstantWeightSize()) {
```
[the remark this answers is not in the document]
gcatron: Agreed, I think we'll want something more robust than a percentage going forward, I'll make it a named variable for now.

include/glow/Runtime/Provisioner/Provisioner.h
gcatron: It doesn't strictly need to be, but the consistent ordering is useful for the round robin provisioning.

tests/unittests/ProvisionerTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "gtest/gtest.h"
+
+using namespace glow;
+using namespace glow::runtime;
+
+class ProvisionerTest : public ::testing::Test {};
+std::unique_ptr<Module> setupModule(uint functionCount) {
+  std::unique_ptr<Module> module = std::make_unique<Module>();
+  std::unique_ptr<Context> ctx = std::make_unique<Context>();
```
nadavrot: No one is using the context.

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -0,0 +1,48 @@
...
+            Module &module);
+
+private:
+  std::unique_ptr<Backend> backend_;
```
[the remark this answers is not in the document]
gcatron: Yes, since the compilation happens in the Provisioner.
Is there significant cost in initializing a backend? Currently the Provisioner will initialize the backend for each device. This could be optimized to check first, or have a map of backends for each device type.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <future>
+#include <queue>
+
+using namespace glow;
+using namespace runtime;
+using DeviceID = unsigned int;
+
+void addNodes(std::queue<std::vector<DAGNode *>> &nextNodes,
```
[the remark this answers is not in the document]
gcatron: It's being used as a work queue in the provision function, This stores the next set of functions to be compiled and added to a device. I don't think a separate function is still necessary though.

tests/unittests/ProvisionerTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+TEST_F(ProvisionerTest, provisionDag) {
+  auto mod = setupModule(6);
+  auto networks = setupDAG(mod.get(), 2, 3);
+  auto provisioner = Provisioner();
+  std::map<DeviceIDTy, std::unique_ptr<DeviceManager>> devices;
+  for (int i = 0; i < 6; i++) {
+    std::unique_ptr<DeviceManager> device(new CPUDeviceManager);
+    devices.emplace(i, std::move(device));
+  }
+  auto result = provisioner.provision(networks, devices, *mod.get());
```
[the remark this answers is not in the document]
gcatron: The provisioner doesn't own the module. The module is provided to the HostManager by ONNX-IFI as a pointer and that gets passed along to the provisioner. This unit test is making a custom module to test the provisioner handling multiple functions in one graph. The device map will be owned by the host manager, the provisioner takes in a reference.

lib/Runtime/Provisioner/Provisioner.cpp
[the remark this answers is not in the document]
gcatron: Named Constant....

gcatron:
ASAN is breaking because of the DAGNodes for the unit test, this will need to be updated after the Partitioner lands so I'll update and address the ASAN then.

gcatron:
hmm looks like I did some git mistakes, I'll move this to a clean branch. I rebased locally but then did a pull instead of a force push....

PR #2277, opened 2019-01-17
Add a small support function vformat to create std::strings using printf-like formatting

opti-mix:
*Description*:
STL is not very convenient if you want to produce a heavily-formatted string and `llvm::format` uses llvm streams to produce the output. The new `vformat` function is useful if you need a convenient method to create a formatted `std::string`.

An example of `vformat` use would be `std::string str = vformat("Happy %s Year %d", "New", 2019)`.

*Testing*:
ninja test


opti-mix:
This is a small utility function that I'd like to land before I can upstream some other changes that make use of it.

opti-mix:
BTW, I'm not quite sure about the name `vformat`. May be something like `strformat` makes more sense?

lib/Support/Support.cpp
```diff
@@ -77,4 +78,26 @@ std::string escapeDottyString(const std::string &str) {
 }
 
 void report(const char *msg) { errs() << msg; }
+
+const std::string vformat(const char *format, ...) {
```
rdzhabarov: can you add a test :)

opti-mix: Absolutely. The only reason I haven't added it yet is that we don't have any tests for functions defined in `Support.h`.

opti-mix: Done. Added a test.

include/glow/Support/Support.h
```diff
@@ -110,6 +110,10 @@ class DescriptionBuilder {
 void report(const char *msg);
 inline void report(const std::string &str) { report(str.c_str()); }
 inline void report(llvm::StringRef str) { report(str.data()); }
+
+/// Printf-like formatting for std::string.
+const std::string strFormat(const char *format, ...)
+    __attribute__((__format__(__printf__, 1, 2)));
```
rdzhabarov: what does this __attribute__ do?

opti-mix: It tells GCC/Clang to verify that passed varargs conform to the format string.

PR #2279, opened 2019-01-17
[Runtime] Use std::shared_ptr to replace raw pointer.

beicy:
*Description*:
To avoid memory leak, replace the raw pointer with std::shared_ptr.

*Testing*:
ninja check.

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -45,10 +45,10 @@ struct DeviceInfo {
 struct DAGNode {
   /// The children of this node, these are nodes that depend on the current
   /// node.
-  std::vector<DAGNode *> children;
+  std::vector<std::shared_ptr<DAGNode>> children;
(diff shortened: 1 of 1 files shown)


beicy:
@bertmaher  This is part of PR https://github.com/pytorch/glow/pull/2268. The Partitioner will create the DAGNodes and Provisioner will use them. So I filed a separate PR. We decided to use shared_ptr since all the DAGNodes will be in Host Manager, and we don't plan to transfer their ownership. If we are using the raw pointer, we need to write the destructor...

nadavrot:
@beicy Is there some entity that has a lifetime that spans the use of the objects? Maybe having some third entity own the nodes can allow the rest of the compiler refer to them by pointer?

beicy:
Close this PR. Plan to keep the raw nodes inside of DAGNode, and create another entity to own those nodes based on the suggestion from @opti-mix and @nadavrot .

Issue #1698, opened 2018-09-24
[New operator] SparseLengthsSumFused8BitRowwise


artemrakhov:
We need to support `SparseLengthsSumFused8BitRowwise` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#sparselengthssumfused8bitrowwise

Also there is weighted version: `SparseLengthsWeightedSumFused8BitRowwise`
We already support float version: `SparseLengthsSum` / `SparseLengthsWeightedSum`

rdzhabarov:
cc: @beicy ^

artemrakhov:
@beicy : any updates on this? It is the most important among our unimplemented ops. Feel free to unassign if its not on your radar.

beicy:
@artemrakhov Sorry,  forget that it is still assigned to me...  Unsigned it for now.

jfix71:
@artemrakhov Were you going to take this? Otherwise I can work on it.

artemrakhov:
No, I wasn't. Thanks for taking this, @jfix71

PR #2282, opened 2019-01-18
Implement RowwiseQuantizedSparseLengthsWeightedSum

jfix71:
*Description*: Implement RowwiseQuantizedSparseLengthsWeightedSum. Added support for both Interpreter and CPU backends.

*Testing*: Added unit tests for both RowwiseQuantizedSparseLengthsWeightedSum and RowwiseQuantizedSparseLengthsSum (implemented as the first but with weights as a float splat of 1.0).

*Documentation*: Added to Quantization.md

Related to https://github.com/pytorch/glow/issues/1698


lib/Quantization/Quantization.cpp
```diff
@@ -420,29 +420,32 @@ class FunctionQuantizer : public FunctionConverter {
...
+          result.getType()->isQuantizedType()) {
+        auto *wq = llvm::dyn_cast<QuantizeNode>(weights.getNode());
+        // For RowwiseQuantizedFullyConnected, the weights need to be constant.
+        if (Constant *wc = llvm::dyn_cast<Constant>(wq->getInput())) {
+          auto *fcq = function_.createRowwiseQuantizedFullyConnected(
+              "rowwiseqfc", input, wc, bias, result.getType(),
+              /* transposeWeight */ true);
+          // Replace the usages of quantized FC node to
+          // RowwiseQuantizedFullyConnected Node.
+          result.replaceAllUsesOfWith(fcq->getResult());
```
beicy: For the above diff,  it seems to me that only the code structure has been changed: 
`if (A) { if (B) {...} } => A =..; if (!A) continue; B=..; if (!B) continue; ...` ?  If so, I don't think it is related to this PR. Please let me know if I missed anything.

jfix71: Ok, I've removed it.

include/glow/Graph/Graph.h
```diff
@@ -565,6 +565,25 @@ class Function final : public Named {
...
                                  NodeValue indices, NodeValue lengths);
 
+  /// Create a node, performing SparseLengthsSum operation, using rowwise
+  /// quantization for the input data. Gathers slices of the outer-most
+  /// dimension of Data indexed by Indices vector, and then accumulates them
+  /// into len(Lengths) entries: first Lengths[0] slices are aggregated to
+  /// Result[0], next Lengths[1] slices are aggregated to Result[1],
+  /// etc. I.e. sum(Lengths) must be equal to len(Indices).
+  RowwiseQuantizedSparseLengthsWeightedSumNode *
+  createRowwiseQuantizedSparseLengthsSum(llvm::StringRef name, Tensor &data,
```
bertmaher: Passing a `Tensor` for `data` seems unusual vs. a `NodeValue` or a `Constant`.  I guess it makes sense given its role though, so I don't feel too strongly.

jfix71: Yeah, a `Constant`/`NodeValue` would essentially only be used as a container for `Tensor data`. I also thought it made sense to differentiate it from the other creator functions, to make it more clear that the data passed in is rowwise-quantized.

When we add in importer support for this Node which loads already-quantized `data` we can refactor/add a new creator method that takes a `NodeValue`/`Constant`.

rdzhabarov:
We'd need to have C2 loader in place in order to close https://github.com/pytorch/glow/issues/1698

jfix71:
I'm going to implement the C2 importer for this in a follow up PR.

PR #2283, opened 2019-01-18
Create GlowErr type with informational error code enum

jackm321:
*Description*:
Create an llvm::Error that holds a string and/or an enum that classifies the error so that this information can be used to send correct ONNXIFI statuses.
In followups, handle error codes to send ONNXIFI statuses. Use error codes in the rest of model loading stack.
*Testing*:
`ninja all`
*Documentation*:
doxygen

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -30,12 +30,6 @@ namespace glow {
 /// line numbers also.
 extern llvm::ExitOnError exitOnErr;
 
-/// Take a message \p str and prepend it with the given \p file and \p line
-/// number. This is useful for augmenting StringErrors with information about
-/// where they were generated.
-std::string addFileAndLineToError(llvm::StringRef str, llvm::StringRef file,
-                                  uint32_t line);
-
 /// Is true_type only if applied to llvm::Error or a descendant.
 template <typename T>
 struct IsLLVMError : public std::is_base_of<llvm::Error, T> {};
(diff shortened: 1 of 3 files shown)


include/glow/Support/Error.h
```diff
@@ -45,6 +39,78 @@ template <typename> struct IsLLVMExpected : public std::false_type {};
...
+    OS << "file: " << fileName_ << " line: " << lineNumber_;
+    if (ec_ != EC::UNKNOWN) {
+      OS << " error code: " << ec_;
+    }
+    if (!message_.empty()) {
+      OS << " message: " << message_;
+    }
+  }
+
+  GlowErr(llvm::StringRef fileName, int32_t lineNumber, llvm::StringRef message,
```
jfix71: `lineNumber` is `int32_t`, while `lineNumber_` is `size_t`.

lib/Support/Error.cpp
```diff
@@ -20,9 +20,5 @@
 namespace glow {
 llvm::ExitOnError exitOnErr("Encountered an error, exiting.\n");
 
-std::string addFileAndLineToError(llvm::StringRef str, llvm::StringRef file,
-                                  uint32_t line) {
-  return llvm::formatv("Error at file {0} line {1} \"{2}\"", file, line, str);
-}
-
+char GlowErr::ID = 0;
```
jfix71: Can you add a comment explaining what this is for?

include/glow/Support/Error.h
```diff
@@ -45,6 +39,78 @@ template <typename> struct IsLLVMExpected : public std::false_type {};
 template <typename T>
 struct IsLLVMExpected<llvm::Expected<T>> : public std::true_type {};
 
+/// Represents errors in Glow. GlowErr track the file name and line number of
+/// where they were created as well as a textual message and/or a error code to
+/// help identify the type of error the occurred programtically.
+class GlowErr final : public llvm::ErrorInfo<GlowErr> {
+public:
+  /// Used by ErrorInfo::classID.
+  static char ID;
```
rdzhabarov: can you make that uint8_t.

include/glow/Support/Error.h
```diff
@@ -45,6 +39,78 @@ template <typename> struct IsLLVMExpected : public std::false_type {};
...
+  /// An enumeration of error codes representing various possible errors that
+  /// could occur.
+  enum EC {
```
rdzhabarov: can this be enum class and also no need to explicit values.

rdzhabarov: ErrorCode would be more descriptive.

include/glow/Support/Error.h
jackm321: dang, thanks!

PR #2284, opened 2019-01-19
Add Host Manager

gcatron:
*Description*: This is a resurrected PR for the HostManager. It is almost ready for merging, once Executor, Partitioner, and Provisioner are merged this will be updated and be ready. 
The HostManager will serve as the entry point for execution calls and handle orchestrating the rest of the runtime components.
*Testing*: Will write Unit Test before merge.
*Documentation*: N/A
Related to #2045

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -1,0 +1,120 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
+#define GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
+#include "glow/Backends/Backend.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <atomic>
+#include <map>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+namespace runtime {
+
+/// Callback signalling the result of running a function.
+struct DeviceConfig {
+  std::string deviceName;
+  BackendKind backendKind;
+};
+
+class Executor;
+
+class Provisioner;
+
+/// The HostManager serves as an entry point into the Runtime environment. It
+/// provides an interface to add, run, and evict networks from the host. It
+/// handles DeviceManager initialization, houses the Executor, and calls into
+/// the Partitioner and Provisioner for network initialization.
+class HostManager final {
+  /// Count of current in-flight networks being run. Atomic to allow
+  /// concurrency in runNetwork.
+  std::atomic<size_t> activeRequestCount_{0};
+
+  /// Count of total requests, this is used as a run ID. Atomic to allow
+  /// concurrency in runNetwork.
+  std::atomic<size_t> totalRequestCount_{0};
+
+  /// Limit maximum count of networks run at once. Hardcoded for now this
+  /// should be a configurable value. Above this limite the HostManager will
+  /// refuse additional request and return Failed.
+  const unsigned int activeRequestLimit_ = 20;
+
+  /// A map from a networkName to the DAG that represents the root of a network.
+  std::unordered_map<std::string, std::unique_ptr<DAGNode>> roots_;
+
+  /// A map from a networkName to the DAG that represents the network.
+  std::unordered_map<std::string, std::unique_ptr<DAGNode>> networks_;
+
+  /// Mutex for networks_ and roots_ since runNetwork, addNetwork, and
+  /// removeNetwork can all be called concurrently, a guard is needed.
+  std::mutex networkLock_;
+
+  /// A map of DeviceManagers by deviceID. An ordered map is used here to allow
+  /// a stable iteration order over devices.
+  std::map<DeviceIDTy, std::shared_ptr<DeviceManager>> devices_;
+
+  /// Executor class, this handles dispatching execution requests to the
+  /// appropriate device managers for an inference request.
+  std::unique_ptr<Executor> executor_;
+
+  /// Backend pointer. This allows the HostManager to optimize functions before
+  /// they are passed to the Partitioner. It is just one since we are currently
+  /// assuming a homogenous set of devices. This may get moved into the
+  /// Partitioner at a later point.
+  std::unique_ptr<Backend> backend_;
+
+  /// The provisioner owns the compiledFunctions and handles loading functions
+  /// onto the devices.
+  std::unique_ptr<Provisioner> provisioner_;
+
+public:
+  /// Adds the network to the host and does the necessary setup work. This
+  /// includes partitioning, provisioning, compiling and initializing
+  /// backends. Additionally DAGs are created for each function and stored in
+  /// networks_. Returns a result code to indicate success.
+  ResultCode addNetwork(Module *M);
+
+  /// Given \p networkName removes that network from the host. This also
+  /// removes the network from any backends setup to execute it.
+  void removeNetwork(llvm::StringRef networkName);
+
+  /// Returns true if \p networkName is already added to the host.
+  bool networkAdded(llvm::StringRef networkName);
+
+  /// Removes all networks from the host, and stops execution on all devices.
+  void clearHost();
+
+  /// Runs the network specified by \p networkName using
+  /// the provided \p context, returns a runIdentifier which refers to the
+  /// specic inference request. Calls \p callback with the results when
+  /// inference is done.
+  /// Note: This method is intended to be thread-safe, it will be called
+  /// concurrently from mutliple threads.
+  /// Returns -1 if networkName not found or too many active requests.
+  RunIdentifierTy runNetwork(llvm::StringRef networkName,
+                             std::unique_ptr<Context> context,
+                             ResultCBTy callback);
+  HostManager(const std::vector<DeviceConfig> &configs);
+  ~HostManager();
+};
+} // namespace runtime
+} // namespace glow
+#endif // GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
--- a/lib/Runtime/HostManager/CMakeLists.txt
+++ b/lib/Runtime/HostManager/CMakeLists.txt
@@ -1,0 +1,10 @@
+add_library(HostManager
+              HostManager.cpp)
+
+target_link_libraries(HostManager
+                      PRIVATE
+                        Backends
+                        Provisioner
+                        Executor
+                        DeviceManager
+                        CPUDeviceManager)                       
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -1,0 +1,159 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Runtime/HostManager/HostManager.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Context.h"
+#include "glow/Partitioner/Partitioner.h"
+#include "glow/Runtime/Executor/Executor.h"
+#include "glow/Runtime/Provisioner/Provisioner.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <future>
+#include <queue>
+
+using namespace glow;
+using namespace runtime;
+
+HostManager::HostManager(const std::vector<DeviceConfig> &configs) {
+  DeviceIDTy deviceCount = 0;
+  provisioner_.reset(new Provisioner());
+  if (configs.size() > 0) {
+    backend_.reset(createBackend(configs[0].backendKind));
+  }
+  for (auto &config : configs) {
+    auto newDevice =
+        std::shared_ptr<DeviceManager>(DeviceManager::createDeviceManager(
+            config.backendKind, config.deviceName));
+    devices_.emplace(deviceCount, std::move(newDevice));
+    deviceCount++;
+  }
+
+  executor_.reset(createExecutor(devices_));
+}
+
+HostManager::~HostManager() { clearHost(); }
+
+ResultCode HostManager::addNetwork(Module *M) {
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto &device : devices_) {
+    DeviceInfo info = DeviceInfo();
+    info.availableMemory = device.second->getAvailableMemory();
+    deviceInfo.push_back(info);
+  }
+  // Optimize functions before passing to partitioner.
+  // Currently hardcoding inference.
+  if (backend_) {
+    for (auto F : M->getFunctions()) {
+      backend_->optimizeFunction(CompilationMode::Infer, F);
+    }
+  }
+  auto partitioner = Partitioner(M, deviceInfo);
+  auto nodeList = std::move(partitioner.Partition());
+  auto result = provisioner_->provision(nodeList.roots, devices_, *M);
+
+  for (auto &node : nodeList.roots) {
+    roots_.emplace(node->name, std::move(node));
+  }
+  for (auto &node : nodeList.nodes) {
+    networks_.emplace(node->name, std::move(node));
+  }
+  return result;
+}
+
+void HostManager::removeNetwork(llvm::StringRef networkName) {
+  // Walk the tree for the given function, calling evict function for each node
+  // before removing it from networks_ which frees the node.
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  if (roots_.find(networkName) == roots_.end()) {
+    return;
+  }
+  std::queue<std::string> nodes;
+  std::set<std::string> allNodes;
+  for (auto &child : roots_[networkName]->children) {
+    nodes.push(child->name);
+    allNodes.insert(child->name);
+  }
+
+  while (!nodes.empty()) {
+    auto nodeName = nodes.front();
+    nodes.pop();
+    auto it = networks_.find(nodeName);
+    if (it != networks_.end()) {
+      for (auto &node : it->second->children) {
+        auto name = node->name;
+        if (allNodes.find(node->name) == allNodes.end()) {
+          nodes.push(node->name);
+          allNodes.insert(node->name);
+        }
+      }
+    }
+  }
+  roots_.erase(roots_.find(networkName));
+  for (auto &networkName : allNodes) {
+    auto networkIterator = networks_.find(networkName);
+    if (networkIterator != networks_.end()) {
+      devices_[networkIterator->second->deviceID]->evictNetwork(networkName);
+      networks_.erase(networkIterator);
+    }
+  }
+}
+
+bool HostManager::networkAdded(llvm::StringRef networkName) {
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  return roots_.find(networkName) != roots_.end();
+}
+
+void HostManager::clearHost() {
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  for (auto &it : devices_) {
+    it.second->stop();
+  }
+  for (auto &network : networks_) {
+    devices_[network.second->deviceID]->evictNetwork(network.second->name);
+  }
+  networks_.clear();
+  roots_.clear();
+
+  activeRequestCount_ = 0;
+}
+
+RunIdentifierTy HostManager::runNetwork(llvm::StringRef networkName,
+                                        std::unique_ptr<Context> context,
+                                        ResultCBTy callback) {
+
+  auto currentRun = totalRequestCount_++;
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  if (roots_.find(networkName) == roots_.end()) {
+    callback(currentRun, ResultCode::Failed, std::move(context));
+    return currentRun;
+  }
+  if (activeRequestCount_ >= activeRequestLimit_) {
+    callback(currentRun, ResultCode::Cancelled, std::move(context));
+    return currentRun;
+  }
+  activeRequestCount_++;
+  executor_->run(roots_[networkName].get(), std::move(context), currentRun,
+                 [&activeRequest = this->activeRequestCount_,
+                  callback](RunIdentifierTy runID, ResultCode result,
+                            std::unique_ptr<Context> context) {
+                   --activeRequest;
+                   callback(runID, result, std::move(context));
+                 });
+  totalRequestCount_--;
+  return currentRun;
+}
--- a/tests/unittests/HostManagerTest.cpp
+++ b/tests/unittests/HostManagerTest.cpp
@@ -1,0 +1,146 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Runtime/HostManager/HostManager.h"
+#include "glow/Graph/Context.h"
+
+#include "gtest/gtest.h"
+
+#include <future>
+#include <thread>
+
+using namespace glow;
+using namespace glow::runtime;
+using DAGNodePairTy = std::pair<std::vector<std::unique_ptr<DAGNode>>,
+                                std::vector<std::unique_ptr<DAGNode>>>;
+
+class HostManagerTest : public ::testing::Test {};
+std::unique_ptr<Module> setupModule(uint functionCount) {
+  std::unique_ptr<Module> module = llvm::make_unique<Module>();
+  for (unsigned int i = 0; i < functionCount; i++) {
+    Function *F = module->createFunction("function" + std::to_string(i));
+    auto *X = module->createPlaceholder(ElemKind::FloatTy, {3},
+                                        "X" + std::to_string(i), false);
+    auto *pow = F->createPow("Pow" + std::to_string(i), X, 2.0);
+    F->createSave("save" + std::to_string(i), pow);
+  }
+  return module;
+}
+
+std::unique_ptr<HostManager> createHostManager(llvm::StringRef name,
+                                               BackendKind kind) {
+  std::vector<DeviceConfig> configs;
+  auto config = DeviceConfig();
+  config.deviceName = name;
+  config.backendKind = kind;
+  configs.push_back(config);
+  std::unique_ptr<HostManager> hostManager =
+      llvm::make_unique<HostManager>(configs);
+  return hostManager;
+}
+
+void addAndRemoveNetwork(HostManager *manager, unsigned int functionNumber) {
+  std::unique_ptr<Module> module = llvm::make_unique<Module>();
+  Function *F =
+      module->createFunction("function" + std::to_string(functionNumber));
+  auto *X = module->createPlaceholder(
+      ElemKind::FloatTy, {3}, "X" + std::to_string(functionNumber), false);
+  auto *pow = F->createPow("Pow" + std::to_string(functionNumber), X, 2.0);
+  F->createSave("save" + std::to_string(functionNumber), pow);
+
+  manager->addNetwork(module.get());
+  manager->removeNetwork("function" + std::to_string(functionNumber));
+}
+
+TEST_F(HostManagerTest, newHostManager) {
+  createHostManager("CPU0", BackendKind::CPU);
+}
+
+TEST_F(HostManagerTest, addNetwork) {
+  auto mod = setupModule(6);
+  auto hostManager = createHostManager("CPU0", BackendKind::CPU);
+  hostManager->addNetwork(mod.get());
+}
+
+TEST_F(HostManagerTest, runNetwork) {
+  Module mod;
+  std::unique_ptr<Context> ctx = llvm::make_unique<Context>();
+
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createPlaceholder(ElemKind::FloatTy, {3}, "X", false);
+  auto *XTensor = ctx->allocate(X);
+  XTensor->getHandle() = {1., 2., 3.};
+  auto *pow = F->createPow("Pow1", X, 2.0);
+  auto *save = F->createSave("save", pow);
+  auto *saveTensor = ctx->allocate(save->getPlaceholder());
+
+  auto hostManager = createHostManager("CPU0", BackendKind::CPU);
+  hostManager->addNetwork(&mod);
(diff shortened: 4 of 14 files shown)


lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Runtime/HostManager/HostManager.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Context.h"
+#include "glow/Runtime/RuntimeTypes.h"
+// include Partitioner
```
nadavrot: What is this?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+  for (auto node : nodeList.roots) {
+    networks_.emplace(node->name, std::move(node));
+  }
+  for (auto node : nodeList.nodes) {
+    networks_.emplace(node->name, std::move(node));
+  }
+  return result;
+}
+
+void HostManager::removeNetwork(std::string networkID) {
```
nadavrot: Please don't use std::string. Use llvm::StringRef for passing strings with clear ownership around.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  /// threshold.
+  RunIdentifierTy runNetwork(std::string networkID,
+                             std::unique_ptr<Context> context,
+                             ResultCBTy callback);
+  HostManager(std::vector<DeviceConfig> &configs);
+  ~HostManager();
+};
+} // namespace runtime.
+} // namespace glow.
+#endif // GLOW_RUNTIME_HOSTMANAGER_H
```
nadavrot: New line at the end of the file.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  std::atomic<size_t> activeRequestCount_{0};
+
+  /// Count of total requests, this is used as a run ID.
+  std::atomic<size_t> totalRequestCount_{0};
+
+  /// Limit maximum count of networks run at once. Hardcoded for now this should
+  /// be a configurable value.
+  const unsigned int activeRequestLimit_ = 20;
+  /// A map from a networkID to the DAG that represents the network.
+  std::unordered_map<std::string, DAGNode> networks_;
```
nadavrot: Is the NetworkID a string?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+    runtime::RunIdentifierTy, runtime::ResultCode, std::unique_ptr<Context>)>;
+
+class DeviceManager;
+class Context;
+class DeviceConfig;
+class HostManager final : std::enable_shared_from_this<HostManager> {
+  /// Count of current in-flight networks being run.
+  std::atomic<size_t> activeRequestCount_{0};
+
+  /// Count of total requests, this is used as a run ID.
```
nadavrot: Why is this atomic? Are you expecting multithreaded access to methods that mutate this field? I don't see any comments about this in the method that accesses the field.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+namespace runtime {
+
+/// Callback signalling the result of running a function.
+using ResultCBTy = std::function<void(
+    runtime::RunIdentifierTy, runtime::ResultCode, std::unique_ptr<Context>)>;
+
+class DeviceManager;
+class Context;
+class DeviceConfig;
+class HostManager final : std::enable_shared_from_this<HostManager> {
```
nadavrot: Why do you need enable_shared_from_this?   Are we planning to pass the HostManager around with no clear ownership?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  std::atomic<size_t> totalRequestCount_{0};
+
+  /// Limit maximum count of networks run at once. Hardcoded for now this should
+  /// be a configurable value.
+  const unsigned int activeRequestLimit_ = 20;
+  /// A map from a networkID to the DAG that represents the network.
+  std::unordered_map<std::string, DAGNode> networks_;
+  /// A map of DeviceManagers by deviceID.
+  std::map<DeviceIDTy, std::unique_ptr<DeviceManager>> devices_;
+  Executor executor_;
```
nadavrot: Doxygen comments please.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  Provisioner provisioner_;
+
+public:
+  /// Adds the network to the host and does the necessary setup work. This
+  /// includes partitioning, provisioning, compiling and initializing backends.
+  /// Returns the networkID of the network.
```
nadavrot: Is the comment about the return code correct?  It says ResultCode not network ID. Either there is a bug or the code is not readable.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  ResultCode addNetwork(Module *M);
+  /// Given \p networkID removes that network from the host. This also removes
+  /// the network from any backends setup to execute it.
+  void removeNetwork(std::string networkID);
```
nadavrot: Please do not use std::string unless you have a good reason, for ownership, for example.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,81 @@
...
+  /// Returns true if \p networkID is already added to the host.
+  bool networkAdded(std::string networkID);
+  /// Removes all networks from the host, and stops execution on all devices.
+  void clearHost();
+  /// Runs the network specified by \p networkID and \p functionName using the
+  /// provided \p context returns true when results are copied into the context
+  /// false if an error occurred or the count of current networks is above the
+  /// threshold.
+  RunIdentifierTy runNetwork(std::string networkID,
+                             std::unique_ptr<Context> context,
```
nadavrot: Are you moving the ownership of the context here? Why? Why not just pass a reference to the external owner?

include/glow/Runtime/HostManager/HostManager.h
nadavrot: Why do we need to leak the ONNX-IFI details here? Why do we need network ID?  At this part of the compiler we need to use Module*, Function* and Node*.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+bool HostManager::networkAdded(std::string networkID) {
+  return networks_.find(networkID) != networks_.end();
+}
+
+void HostManager::clearHost() {
+  for (auto it : networks_) {
+    removeNetwork(it->first);
+  }
+  for (auto it : devices_) {
```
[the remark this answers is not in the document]
nadavrot: Just in case the device uses the network data structures.

include/glow/Runtime/HostManager/HostManager.h
jackm321: Agree, I think Module* would be best here. In earlier versions of the HostManager interface, ownership of Module was lost during addNetwork(), this isn't the case any more so a separate ID shouldn't be necessary now.

jackm321: Also, what about concurrency here? I think this needs some locking, otherwise addNetwork, removeNetwork, and runNetwork can't be run concurrently.

include/glow/Runtime/HostManager/HostManager.h
nickgg: We're using ownership as synchronization, the Device is going to be modifying the memory owned by the context and this way we don't need to lock it since nowhere else should have a reference.

lib/Runtime/HostManager/HostManager.cpp
gcatron: Will update to actual include.

include/glow/Runtime/HostManager/HostManager.h
gcatron: You are correct, enable_shared_from_this is for the callback to be able to decrement the inflight counter. This is to handle the case where the ONNX-IFI interface releases the HostManager while there is an active request. This keeps the HostManager alive until outstanding requests finish.

include/glow/Runtime/HostManager/HostManager.h
gcatron: renaming NetworkId to networkName for clarity, Within a given module there are multiple functions. They each have a unique string name identifying them. Currently the plan is to use that string, unless a different unique identifier is determined by the Caffe2/ONNX-IFI to refer to a network once it's initialized. (This is flexible, we just need  a unique identifier so the HostManager can map that to its DAG which contains the information to execute.) 

I agree on the locking, it needs to be added to allow concurrency.

include/glow/Runtime/HostManager/HostManager.h
gcatron: Will add comments to indicate multi-thread expectations.

include/glow/Runtime/HostManager/HostManager.h
jackm321: Makes sense

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+                                        ResultCBTy callback) {
+  if (networks_.find(networkID) == networks_.end()) {
+    return Failed;
+  }
+  if (activeRequestCount_ >= activeRequestLimit_) {
+    return Cancelled;
+  }
+  auto currentRun = activeRequestCount_++;
+
+  ++totalRequestCount_;
```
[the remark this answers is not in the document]
gcatron: Like @jackm321  said, runNetwork can be called from multiple threads, the idea is to ensure a concurrent safe identifier per run and ensure we uniformly enforce the max request count.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+  auto shared_this = shared_from_this();
+  executor_.runNetwork(
+      networks_[networkID], context, currentRun,
+      [shared_this, callback](RunIdentifierTy runID, ResultCode result,
+                              std::unique_ptr<Context> context) {
+        callback(currentRun, result, std::move(context));
```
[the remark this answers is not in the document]
gcatron: We are returning ownership back to the caller. ONNX-IFI in our case. It is then free to delete the context.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+HostManager::~HostManager() { clearHost(); }
+ResultCode HostManager::addNetwork(Module *M) {
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto device : devices_) {
+    auto info = DeviceInfo();
+    info.availableMemory = device->first->getAvailableMemory();
+  }
+  auto partitioner = Partitioner(M, deviceInfo);
+  auto nodeList = partitioner.partition(M);
```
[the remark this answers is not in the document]
beicy: @jackm321 Partitioner will check the sub-functions' validity after partitioning.  For the placeholders, we only introduce placeholders and won't change the original placeholders so far.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,102 @@
...
+
+  /// The provisioner assigns networks to specifc devices and handles
+  /// compilation for that device.
+  Provisioner provisioner_;
+
+public:
+  /// Adds the network to the host and does the necessary setup work. This
+  /// includes partitioning, provisioning, compiling and initializing backends.
+  /// Returns a result code to indicate success.
+  ResultCode addNetwork(Module *M);
```
jackm321: What are the pros to not passing ownership of Module here?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+  for (auto config : configs) {
+    auto newDevice = DeviceManager.getDevice(config);
+    devices_.emplace(deviceCount, newDevice);
+    deviceCount++;
+  }
+  provisioner_ = Provisioner();
+}
+
+HostManager::~HostManager() { clearHost(); }
+ResultCode HostManager::addNetwork(Module *M) {
```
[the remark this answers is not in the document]
gcatron: Currently HostManager does not take ownership of the Module, this can be changed if we desire. networks_ contains the DAGNodes which contain the runtime information but not the module, constants, or the compiled function.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+  for (auto node : nodeList.nodes) {
+    networks_.emplace(node->name, std::move(node));
+  }
+  return result;
+}
+
+void HostManager::removeNetwork(std::string networkID) {
+  // Walk the tree for the given function, calling evict function for each node
+  // before removing it from networks_ which frees the node.
+  std::queue<std::string> nodes;
```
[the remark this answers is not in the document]
gcatron: Strings are being used because that's the key to the map that stores the nodes. I could use pointers instead and then pointer->name to get the iterator to remove. Using a string seemed cleaner.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+    deviceCount++;
+  }
+  provisioner_ = Provisioner();
+}
+
+HostManager::~HostManager() { clearHost(); }
+ResultCode HostManager::addNetwork(Module *M) {
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto device : devices_) {
+    auto info = DeviceInfo();
```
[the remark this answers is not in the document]
gcatron: Good catch, there was supposed to be a push_back to add info to deviceInfo. It's there now.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+  return result;
+}
+
+void HostManager::removeNetwork(std::string networkID) {
+  // Walk the tree for the given function, calling evict function for each node
+  // before removing it from networks_ which frees the node.
+  std::queue<std::string> nodes;
+  nodes.push_back(networkID);
+  while (!nodes.empty()) {
```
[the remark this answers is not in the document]
gcatron: forgot to pop..

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void HostManager::clearHost() {
+  for (auto it : networks_) {
+    removeNetwork(it->first);
+  }
+  for (auto it : devices_) {
+    it->second.stop();
+  }
+  activeRequestCount_ = 0;
```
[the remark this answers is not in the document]
gcatron: Agreed, created #2297 to track the change.

include/glow/Runtime/HostManager/HostManager.h
gcatron: Notionally the entity calling HostManager owns the module and persists, we can change this to pass ownership in and back out if desired.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto device : devices_) {
+    auto info = DeviceInfo();
+    info.availableMemory = device->first->getAvailableMemory();
+    deviceInfo.push_back(info);
+  }
+  auto partitioner = Partitioner(M, deviceInfo);
+  auto nodeList = partitioner.partition(M);
+  auto result = provisioner_.provision(nodeList.roots, devices_, M);
+  for (auto node : nodeList.roots) {
```
beicy: Both Provisioner and Partitioner are only used in this function. Why is provisioner_ is private? Can it be tmp var as Partitioner is?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <future>
+#include <queue>
+
+using namespace glow;
+using namespace runtime;
+
+HostManager::HostManager(std::vector<DeviceConfig> &configs) {
+  DeviceIDTy deviceCount = 0;
```
rdzhabarov: do we need this? regular loop with index over configs.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+  provisioner_ = Provisioner();
+  executor_ = Executor(devices_);
+}
+
+HostManager::~HostManager() { clearHost(); }
+
+ResultCode HostManager::addNetwork(Module *M) {
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto device : devices_) {
+    auto info = DeviceInfo();
```
rdzhabarov: nit: `DeviceInfo info;`

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+    info.availableMemory = device->first->getAvailableMemory();
+    deviceInfo.push_back(info);
+  }
+  auto partitioner = Partitioner(M, deviceInfo);
+  auto nodeList = partitioner.partition(M);
+  auto result = provisioner_.provision(nodeList.roots, devices_, M);
+  for (auto node : nodeList.roots) {
+    networks_.emplace(node->name, std::move(node));
```
rdzhabarov: why no synchronization on networks_ if addNetwork can be called concurrently

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+        nodes.push_back(node->name);
+      }
+      devices_[it->second->deviceID].evictNetwork(it->second->name);
+      networks_.erase(it);
+    }
+  }
+}
+
+bool HostManager::networkAdded(llvm::StringRef networkName) {
+  return networks_.find(networkName) != networks_.end();
```
rdzhabarov: this needs look as well

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  activeRequestCount_ = 0;
+}
+
+RunIdentifierTy HostManager::runNetwork(llvm::StringRef networkName,
+                                        std::unique_ptr<Context> context,
+                                        ResultCBTy callback) {
+
+  auto currentRun = totalRequestCount_++;
+  if (networks_.find(networkName) == networks_.end()) {
```
rdzhabarov: this find needs lock

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,102 @@
...
+#include <map>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+namespace runtime {
+
+/// Callback signalling the result of running a function.
+using ResultCBTy = std::function<void(
```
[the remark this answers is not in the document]
gcatron: Agreed, we can move it into RuntimeTypes.h

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,103 @@
...
+  unsigned int activeRequestLimit_ = 20;
+  /// A map from a networkName to the DAG that represents the network.
+  std::unordered_map<std::string, DAGNode> networks_;
+
+  /// Mutex for networks_ since runNetwork, addNetwork, and removeNetwork can
+  /// all be called concurrently a guard is needed.
+  std::mutex networkLock_;
+
+  /// A map of DeviceManagers by deviceID.
+  std::map<DeviceIDTy, std::unique_ptr<DeviceManager>> devices_;
```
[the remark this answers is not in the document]
gcatron: Unordered was considered, I decided to keep it ordered though. There won't be a large number of devices  and this allows the Provisioner to take advantage of a stable ordering.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,103 @@
...
+public:
+  /// Adds the network to the host and does the necessary setup work. This
+  /// includes partitioning, provisioning, compiling and initializing backends.
+  /// Additionally DAGs are created for each function and stored in networks_.
+  /// Returns a result code to indicate success.
+  ResultCode addNetwork(Module *M);
+
+  /// Given \p networkName removes that network from the host. This also
+  /// removes the network from any backends setup to execute it.
+  void removeNetwork(llvm::StringRef networkName);
```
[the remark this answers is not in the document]
gcatron: It is likely that when one function is removed all functions in that batch will be removed as well. The reason per function removal was used is it does not require an additional identifier to be kept around. Once a network is added we discard any notion of Module. To remove the batch that would require the caller and HostManager to keep a map of this batch. (That can be added if we think it's needed.)

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <future>
+#include <queue>
+
+using namespace glow;
+using namespace runtime;
+
+HostManager::HostManager(std::vector<DeviceConfig> &configs) {
+  DeviceIDTy deviceCount = 0;
+  for (auto &config : configs) {
+    auto newDevice = DeviceManager.getDevice(config);
```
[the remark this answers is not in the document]
gcatron: Added.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+#ifndef GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
+#define GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
+#include "glow/Backends/Backend.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include <atomic>
+#include <llvm/ADT/StringRef.h>
```
nadavrot: Please take a look at other places where we include LLVM. We usually place them as one chunk below the glow includes and above the system includes.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+ * limitations under the License.
+ */
+
+#include "glow/Runtime/HostManager/HostManager.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Context.h"
+#include "glow/Partitioner/Partitioner.h"
+#include "glow/Runtime/Provisioner/Provisioner.h"
+#include "glow/Runtime/RuntimeTypes.h"
+// #include "glow/Runtime/Executor.h"
```
nadavrot: This is commented out.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <future>
+#include <queue>
+
+using namespace glow;
+using namespace runtime;
+
+HostManager::HostManager(const std::vector<DeviceConfig> &configs) {
+  DeviceIDTy deviceCount = 0;
+  for (auto &config : configs) {
+    // auto newDevice = DeviceManager.getDevice(config);
```
nadavrot: This code is commented out.

lib/Runtime/HostManager/HostManager.cpp
gcatron: Yes, it's waiting for the executor to land.

include/glow/Runtime/HostManager/HostManager.h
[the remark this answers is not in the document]
rdzhabarov: can you add comment that you need stable iteration over devices.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,107 @@
...
+  /// Note: This method is intended to be thread-safe, it will be called
+  /// concurrently from mutliple threads.
+  /// Returns -1 if networkName not found or too many active requests.
+  RunIdentifierTy runNetwork(llvm::StringRef networkName,
+                             std::unique_ptr<Context> context,
+                             ResultCBTy callback);
+  HostManager(const std::vector<DeviceConfig> &configs);
+  ~HostManager();
+};
+} // namespace runtime.
```
rdzhabarov: super nit: no '.' in here after namespace

lib/Runtime/HostManager/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(HostManager
+              HostManager.cpp)
+
+target_link_libraries(HostManager
+                      PRIVATE
+                        Backends
+                        Provisioner
+                        DeviceManager)
```
rdzhabarov: kill spaces here.

tests/unittests/HostManagerTest.cpp
```diff
@@ -0,0 +1,120 @@
...
+  }
+  return module;
+}
+
+DAGNodePairTy setupDAG(uint rootCount, uint childCount) {
+  std::vector<std::unique_ptr<DAGNode>> networks;
+  std::vector<std::unique_ptr<DAGNode>> children;
+  uint currentFunction = 0;
+  for (unsigned int root = 0; root < rootCount; root++) {
+    auto rootNode = std::make_unique<DAGNode>();
```
rdzhabarov: can you fix all those std::make_unique to be llvm::make_unique

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+  // executor_ = Executor(devices_);
+}
+
+HostManager::~HostManager() { clearHost(); }
+
+ResultCode HostManager::addNetwork(Module *M) {
+  std::vector<DeviceInfo> deviceInfo;
+  for (auto &device : devices_) {
+    DeviceInfo info = DeviceInfo();
+    info.availableMemory = device.second->getAvailableMemory();
```
rdzhabarov: should not `getAvailableMemory` get synchronized?
when do we actually place network on a device?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // auto newDevice = DeviceManager.getDevice(config);
+    // devices_.emplace(deviceCount, newDevice);
+    deviceCount++;
+  }
+  // executor_ = Executor(devices_);
+}
+
+HostManager::~HostManager() { clearHost(); }
+
+ResultCode HostManager::addNetwork(Module *M) {
```
rdzhabarov: multithreaded tests for these things

lib/Runtime/HostManager/HostManager.cpp
gcatron: The HostManager doesn't keep track of device memory, it's likely the DeviceManager will keep it in sync, I don't envision this call to query the device each time.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+  for (auto &networkName : allNodes) {
+    auto networkIterator = networks_.find(networkName);
+    if (networkIterator != networks_.end()) {
+      devices_[networkIterator->second->deviceID]->evictNetwork(networkName);
+      networks_.erase(networkIterator);
+    }
+  }
+}
+
+bool HostManager::networkAdded(llvm::StringRef networkName) {
```
beicy: Can I ask where is this function called?

gcatron: The onnxifi interface has a similar function. The thought is it would be called from there.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,120 @@
...
+  const unsigned int activeRequestLimit_ = 20;
+
+  /// A map from a networkName to the DAG that represents the root of a network.
+  std::unordered_map<std::string, std::unique_ptr<DAGNode>> roots_;
+
+  /// A map from a networkName to the DAG that represents the network.
+  std::unordered_map<std::string, std::unique_ptr<DAGNode>> networks_;
+
+  /// Mutex for networks_ since runNetwork, addNetwork, and removeNetwork can
+  /// all be called concurrently a guard is needed.
```
jackm321: nit caps this

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  activeRequestCount_ = 0;
+}
+
+RunIdentifierTy HostManager::runNetwork(llvm::StringRef networkName,
+                                        std::unique_ptr<Context> context,
+                                        ResultCBTy callback) {
+
+  auto currentRun = totalRequestCount_++;
+  std::lock_guard<std::mutex> networkLock(networkLock_);
```
jackm321: does this lock need to be held while callback() is being called?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+}
+
+bool HostManager::networkAdded(llvm::StringRef networkName) {
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  return roots_.find(networkName) != roots_.end();
+}
+
+void HostManager::clearHost() {
+  std::lock_guard<std::mutex> networkLock(networkLock_);
```
jackm321: does lock need to be held while stopping devices?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,120 @@
...
+  /// concurrency in runNetwork.
+  std::atomic<size_t> totalRequestCount_{0};
+
+  /// Limit maximum count of networks run at once. Hardcoded for now this
+  /// should be a configurable value. Above this limite the HostManager will
+  /// refuse additional request and return Failed.
+  const unsigned int activeRequestLimit_ = 20;
+
+  /// A map from a networkName to the DAG that represents the root of a network.
+  std::unordered_map<std::string, std::unique_ptr<DAGNode>> roots_;
```
jackm321: Can we just store map of DAGNodeLists instead of this and networks_?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -0,0 +1,120 @@
...
+  /// backends. Additionally DAGs are created for each function and stored in
+  /// networks_. Returns a result code to indicate success.
+  ResultCode addNetwork(Module *M);
+
+  /// Given \p networkName removes that network from the host. This also
+  /// removes the network from any backends setup to execute it.
+  void removeNetwork(llvm::StringRef networkName);
+
+  /// Returns true if \p networkName is already added to the host.
+  bool networkAdded(llvm::StringRef networkName);
```
jackm321: where does the caller get networkName from?

include/glow/Runtime/HostManager/HostManager.h
jackm321: also protects roots_

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -0,0 +1,159 @@
...
+  for (auto &node : nodeList.roots) {
+    roots_.emplace(node->name, std::move(node));
+  }
+  for (auto &node : nodeList.nodes) {
+    networks_.emplace(node->name, std::move(node));
+  }
+  return result;
+}
+
+void HostManager::removeNetwork(llvm::StringRef networkName) {
```
jackm321: what if active requests are in flight for the network being removed?

lib/Runtime/HostManager/HostManager.cpp
gcatron: In the callback we are good since activeRequestCount_ is atomic. It's needed here since we are accessing roots_.

lib/Runtime/HostManager/HostManager.cpp
gcatron: For the device no, but this way we can't add a function to a device we are stopping.

include/glow/Runtime/HostManager/HostManager.h
gcatron: They can have name collision. That was the original design though....

include/glow/Runtime/HostManager/HostManager.h
gcatron: It's the name of the function in the provided Module.

lib/Runtime/HostManager/HostManager.cpp
gcatron: A great question. That's related to #2297 which will need to get tackled.

include/glow/Runtime/HostManager/HostManager.h
gcatron: I misunderstood, I think that's a great idea, I think it'll be better to land and do the refactor afterwards since to do it right we'd want to move DAGNodeList into runtimeTypes. I've created #2353 to track it.

gcatron:
I currently have the concurrent test disabled. This is because the compiled function is owned by the provisioner not the DeviceManager. Since there's no callback on evictNetwork The HostManager can free the compiledFunction out from under the DeviceManager. We either need to add a callback/make evict blocking, or move ownership of the CompiledFunction to the deviceManager.

nickgg:
@gcatron I think it makes sense to add a callback to the evict, we can probably use the same one from addNetwork.

gcatron:
> @gcatron I think it makes sense to add a callback to the evict, we can probably use the same one from addNetwork.

That sounds good to me. I created #2357 once we do that we should be able to re-enable the concurrent test. If we add a callback #2354 probably isn't strictly necessary.

PR #2287, opened 2019-01-20
[tests] Include private headers using relative paths

bertmaher:
*Description*: While it's best for tests to only use the "public" headers of a library, it sometimes makes sense to use the private/internal headers directly.  Instead of adding to the include path via cmake, let's refer to them using relative paths like `../../lib/Foo/Bar.h`.  This seems to be LLVM's approach, and I see two advantages:
1. It makes clear you're testing internal implementation, and
2. It plays nicely with FB's internal header layouts O:-)

*Testing*: ninja all

*Documentation*: n/a

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -141,7 +141,6 @@ target_link_libraries(GraphSchedulerTest
                         IR
                         gtest
                         TestMain)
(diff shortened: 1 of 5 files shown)


nadavrot:
> This seems to be LLVM's approach, and I see two advantages:

@bertmaher Where is LLVM using the relative path include? I could not find this patten using grep.

PR #2290, opened 2019-01-23
Remove IRFunction requirement from CollectConstants

nickgg:
*Description*: Pulling the collectConstants process out of compile meant that we must do it later in the DeviceManager. Practically this would mean storing the IR representation in the CompiledFunction so we can pull out the weight tensors. Instead, we can refactor to use the RuntimeBundle and the Module (both of which we know we have in the DM) and remove the extra state.
*Testing*: unittests
*Documentation*:

### DIFF
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -21,7 +21,10 @@ using llvm::cast;
 using llvm::isa;
 
 void glow::runtime::RuntimeBundle::collectConstants(const IRFunction *F) {
+  collectConstants(F->getGraph()->getParent());
+}
 
+void glow::runtime::RuntimeBundle::collectConstants(const Module *M) {
   // At compile time condense constants to a single block of memory.
   // This allows the graph to go away after compile time.
   // If there are no constants return nullptr.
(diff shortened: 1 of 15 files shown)


nadavrot:
@nickgg The IR is optional. Will your procedure work for backends that don't use the IR?

nickgg:
@nadavrot Yes, this _removes_ the requirement for the IR.

lib/Backends/BackendUtils.cpp
```diff
@@ -15,30 +15,42 @@
...
-    auto payload = v->getPayload().getUnsafePtr();
-    auto numBytes = w->getSizeInBytes();
-    auto addr = getValueOffset(v);
+
+  for (const auto &symbol : symbolTable_) {
+    llvm::StringRef name = symbol.first;
+    const RuntimeSymbolInfo &info = symbol.second;
+
+    Constant *c = M->getConstantByName(name);
+    if (!c) {
```
rdzhabarov: what is the situation when module does not have that constant?

nickgg: It's a consequence of the RuntimeBundle overlap issue you commented on in #2274, the SymbolTable contains both Constants and Placeholders. We only want to copy the contents of the Constants though, obviously.

PR #2291, opened 2019-01-23
[examples] Add resnet-concurrent example

nickgg:
*Description*: Adds a new example binary demonstrating concurrent execution of resnet50 using a configurable number of CPUDeviceManagers. We're able to keep cores mostly in use (some slight delays because file loading is inline):
image (run with 7 virtual devices).

*Testing*: run binary with 500 normalized images from imagenet, manually checked their identified class.
*Documentation*:

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -72,6 +72,19 @@ if(GLOW_WITH_CPU)
                           ExecutionEngine
                           Graph
                           Importer)
+
+  add_executable(resnet-concurrent
+                   resnet-concurrent.cpp)
+  target_link_libraries(resnet-concurrent
+                        PRIVATE
+                          Backends
+                          CPUDeviceManager
+                          DeviceManager
+                          ExecutionEngine
+                          Graph
+                          Importer
+                          Optimizer)
+  target_include_directories(resnet-concurrent PUBLIC ${CMAKE_SOURCE_DIR}/lib)
 endif()
 
 if(GLOW_WITH_BUNDLES)
--- a/examples/resnet-concurrent.cpp
+++ b/examples/resnet-concurrent.cpp
@@ -1,0 +1,216 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+// We have not written the piece in the HostManager that initializes and creates
+// DeviceManagers for individual devices, so for now just reach into lib/
+#include "Backends/CPU/CPUDeviceManager.h"
+
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+
+#include <chrono>
+#include <future>
+
+using namespace glow;
+using namespace glow::runtime;
+using namespace std::chrono_literals;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-concurrent Options");
+llvm::cl::opt<std::string>
+    inputDirectory(llvm::cl::desc("input directory for images, which must be "
+                                  "png's with standard imagenet normalization"),
+                   llvm::cl::init("../tests/images/imagenet/"),
+                   llvm::cl::Positional, llvm::cl::cat(category));
+llvm::cl::opt<unsigned> numDevices("numDevices",
+                                   llvm::cl::desc("Number of Devices to use"),
+                                   llvm::cl::init(7), llvm::cl::value_desc("N"),
+                                   llvm::cl::cat(category));
+llvm::cl::opt<unsigned>
+    maxImages("maxImages",
+              llvm::cl::desc("Maximum number of images to load and classify"),
+              llvm::cl::init(1000), llvm::cl::value_desc("N"),
+              llvm::cl::cat(category));
+} // namespace
+
+/// Loads the model into /p module and returns the input and output
+/// Placeholders.
+std::pair<Placeholder *, Placeholder *> loadResnet50Model(TypeRef inputType,
+                                                          Module &module) {
+  Function *F = module.createFunction("resnet50");
+
+  llvm::outs() << "Loading resnet50 model.\n";
+
+  const char inputName[] = "gpu_0/data";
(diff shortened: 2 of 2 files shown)


examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+  std::promise<void> finished;
+
+  size_t started = 0;
+  std::atomic<size_t> returned{0};
+
+  size_t max = 1000;
+
+  while (started++ < max) {
+    if (code.value() != 0 || dirIt == llvm::sys::fs::directory_iterator()) {
```
jfix71: `directory_iterator()` is the end iterator I'm assuming?

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,164 @@
...
+  Function *F = module.createFunction("resnet50");
+  TypeRef inputType(module.uniqueType(ElemKind::FloatTy, {1, 3, 224, 224}));
+
+  llvm::outs() << "Loading resnet50 model.\n";
+
+  const char inputName[] = "gpu_0/data";
+  Caffe2ModelLoader loader("resnet50/predict_net.pb", "resnet50/init_net.pb",
+                           {inputName}, {inputType}, *F);
+  Placeholder *input =
+      llvm::cast<Placeholder>(cantFail(loader.getNodeValueByName(inputName)));
```
rdzhabarov: did not know about `cantFail` - interesting.

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,164 @@
...
+  Placeholder *output = cantFail(loader.getSingleOutput());
+  auto *backend = createBackend(BackendKind::CPU);
```
rdzhabarov: why not through ExecutionEngine?

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,164 @@
...
+  llvm::outs() << "Starting compile.\n";
+  backend->optimizeFunction(CompilationMode::Infer, F);
+  auto compiledFunction = backend->compile(F);
+
+  FunctionMapTy functionMap;
+  functionMap.emplace("resnet50", compiledFunction.get());
+
+  std::promise<ResultCode> compilePromise;
+  size_t compilesLeft{numDevices};
+  for (int i = 0; i < numDevices; ++i) {
```
rdzhabarov: unsigned as well.

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+#include <future>
+
+using namespace glow;
+using namespace glow::runtime;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-concurrent Options");
+static llvm::cl::opt<std::string>
+    inputDirectory(llvm::cl::desc("input directory"),
```
rdzhabarov: can you specify what needs to be in the input directory

examples/CMakeLists.txt
```diff
@@ -72,6 +72,19 @@ if(GLOW_WITH_CPU)
...
+  target_link_libraries(resnet-concurrent
+                        PRIVATE
+                          Backends
+                          CPUDeviceManager
+                          DeviceManager
+                          ExecutionEngine
+                          Graph
+                          Importer
+                          Optimizer)
+  target_include_directories(resnet-concurrent PUBLIC ${CMAKE_SOURCE_DIR}/lib)
```
rdzhabarov: Can `CPUDeviceManager.h` be placed into glow/include instead of this? I assume other pieces would need that include as well later on.

nickgg: It would be weird if we did that for DeviceManagers but not for Backends though.

I think really we wouldn't expect to need to create a specific kind of device, instead we'd use the HostManager interface.

Don't really mind moving the header to include/ though I guess.

examples/resnet-concurrent.cpp
nickgg: I mean the whole point is that we're using the DeviceManager, which the ExecutionEngine doesn't use. I could still do the compile through the EE then pull out the Module and the compiledFunction. Doesn't really feel clearer to me but I'm closer to the implementation.

examples/resnet-concurrent.cpp
nickgg: Yeah directory_iterator doesn't have an end() it seems, it just uses the "empty iterator".

nickgg:
I'll write some better comments. I was kind of thinking this would be a temporary thing until we have the E2E Runtime equivalent.

examples/resnet-concurrent.cpp
rdzhabarov: got it, makes sense. do we need any includes, dependencies on EE?

examples/CMakeLists.txt
rdzhabarov: Yeah, I guess either way it's a bit weird. Not feeling strongly either way.

examples/resnet-concurrent.cpp
nickgg: Right now we do because `updateInputPlaceholders` is defined in the ExecutionEngine. I think we could move that func somewhere more generic, maybe to Context?

nickgg:
I've reorganised the example into some logical subfunctions.

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <chrono>
+#include <future>
+
+using namespace glow;
+using namespace glow::runtime;
+using namespace std::chrono_literals;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-concurrent Options");
+static llvm::cl::opt<std::string>
```
rdzhabarov: since it's already in the anon namespace we should not need `static`.

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,219 @@
...
+llvm::cl::opt<unsigned>
+    maxImages("maxImages",
+              llvm::cl::desc("Maximum number of images to load and classify"),
+              llvm::cl::init(1000), llvm::cl::value_desc("N"),
+              llvm::cl::cat(category));
+} // namespace
+
+/// Loads the model into /p module and returns the input and output
+/// Placeholders.
+std::pair<Placeholder *, Placeholder *> loadResnet50Model(TypeRef &inputType,
```
rdzhabarov: TypeRef is already a pointer, just TypeRef inputType

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,219 @@
...
+      argc, argv, "Run ResNet concurrently on a fixed number of CPU devices");
+
+  llvm::outs() << "Initializing " << numDevices << " CPU Devices.\n";
+  std::vector<std::unique_ptr<CPUDeviceManager>> devices;
+  for (unsigned int i = 0; i < numDevices; ++i) {
+    devices.emplace_back(std::make_unique<CPUDeviceManager>());
+    devices[i]->init();
+  }
+
+  /// Load and compile model.
```
rdzhabarov: no need doxygen style here.

examples/resnet-concurrent.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    std::string path = dirIt->path();
+
+    auto image =
+        readPngImageAndPreprocess(path, ImageNormalizationMode::k0to1,
+                                  ImageChannelOrder::BGR, ImageLayout::NCHW,
+                                  /* useImagenetNormalization */ true);
+
+    Tensor batch(inputType);
```
jfix71: Why can't you just pass `image` to `updateInputPlaceholders()`?

nickgg: I'm actually not sure, I took this pattern from the resnet-verify example. You're right that you don't need it though.

PR #2292, opened 2019-01-23
[Importer] Add C2 importer support for RWQ SLWS/SLS

jfix71:
*Description*: Add support in the Caffe2 model loader for `SparseLengthsWeightedSum8BitsRowwise` and `SparseLengthsSum8BitsRowwise`.

*Testing*: Added unit tests

*Documentation*: N/A

Related to https://github.com/pytorch/glow/issues/1698

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -870,6 +870,80 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return llvm::Error::success();
   }
 
+  if (typeName == "SparseLengthsWeightedSum8BitsRowwise" ||
+      typeName == "SparseLengthsSum8BitsRowwise") {
+    // If SparseLengthsWeightedSum8BitsRowwise, then the weights are the second
+    // input and so we need to shift indices/lengths/scalesBiases.
+    size_t indicesIdx = 1;
+    size_t lengthsIdx = 2;
+    size_t scalesBiasesIdx = 3;
+    if (typeName == "SparseLengthsWeightedSum8BitsRowwise") {
+      indicesIdx++;
+      lengthsIdx++;
+      scalesBiasesIdx++;
+    }
+
+    NodeValue data;
+    ASSIGN_VALUE_OR_RETURN_ERR(data,
+                               getNodeValueOrCreateConstantByName(op.input(0)));
+    NodeValue indices;
+    ASSIGN_VALUE_OR_RETURN_ERR(
+        indices, getNodeValueOrCreateConstantByName(op.input(indicesIdx)));
+    NodeValue lengths;
+    ASSIGN_VALUE_OR_RETURN_ERR(
+        lengths, getNodeValueOrCreateConstantByName(op.input(lengthsIdx)));
+    NodeValue scalesBiases;
+    ASSIGN_VALUE_OR_RETURN_ERR(scalesBiases, getNodeValueOrCreateConstantByName(
+                                                 op.input(scalesBiasesIdx)));
+
+    Constant *scalesBiasesC = llvm::dyn_cast<Constant>(scalesBiases);
+    RETURN_ERR_IF_NOT(scalesBiasesC, "scales_biases must be Constant.");
+    Constant *dataC = llvm::dyn_cast<Constant>(data);
+    RETURN_ERR_IF_NOT(dataC->getElementType() == ElemKind::Int8QTy,
+                      "Data must be Int8QTy.");
+
+    const size_t numRows = data.dims()[0];
+
+    // Make sure all the shapes make sense.
+    RETURN_ERR_IF_NOT(lengths.dims().size() == 1, "lengths must be a vector.");
+    RETURN_ERR_IF_NOT(indices.dims().size() == 1, "indices must be a vector.");
+    RETURN_ERR_IF_NOT(scalesBiases.dims().size() == 2,
+                      "scale_bias has to be a matrix.");
+    RETURN_ERR_IF_NOT(scalesBiases.dims()[0] == numRows,
+                      "scale_bias must have the same number of rows as data.");
+    RETURN_ERR_IF_NOT(scalesBiases.dims()[1] == 2,
+                      "Second dim of scale_bias has to be equal to 2.");
+
+    // Now strip out the scales and biases into their own tensors.
+    Constant *dataScales = G_.getParent()->createConstant(
+        ElemKind::FloatTy, {numRows}, "dataScales");
+    Constant *dataOffsets = G_.getParent()->createConstant(
+        ElemKind::Int32ITy, {numRows}, "dataOffsets");
+
+    auto dataScalesH = dataScales->getHandle<float>();
+    auto dataOffsetsH = dataOffsets->getHandle<int32_t>();
+    auto scalesBiasesH = scalesBiasesC->getHandle<float>();
+    for (size_t i = 0, e = numRows; i < e; i++) {
+      dataScalesH.at({i}) = scalesBiasesH.at({i, 0});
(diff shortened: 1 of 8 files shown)


jfix71:
Note that the way this and the SLWS Node are implemented requires duplicating the `Data` input to the SLWS node -- right now it strips out the fused scale/offset to create three separate Constant inputs to the node (unfused `Data`, `Scales`, and `Offsets`) when loading the model in order to create the SLWS Node.

This could be problematic, as `Data` could be very large and duplicating it could use too much host memory. We may want to use something like "offline Constants" here (deferring loading/transforming the fused `Data` Constant until copying the Constants to an accelerator) if it becomes an issue.

tests/models/caffe2Models/fused_rowwise_quantized_sparse_lengths_sum_predict_net.pbtxt
```diff
@@ -0,0 +1,15 @@
+name: "fused_rowwise_quantized_sparse_lengths_sum_predict_net_test"
+op {
+  input: "data"
+  input: "indices"
+  input: "lengths"
+  output: "result"
+  name: ""
+  type: "SparseLengthsSumFused8BitRowwise"
+  device_option {
```
rdzhabarov: nit: do not really need this.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -849,6 +849,98 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+    RETURN_ERR_IF_NOT(
+        fusedDataC,
+        "Input data must be Constant in order to retrieve Scales/Offsets.");
+    RETURN_ERR_IF_NOT(fusedDataC->getElementType() == ElemKind::Int8QTy,
+                      "Data must be Int8QTy.");
+    auto dataDims = fusedDataC->dims();
+    static_assert(sizeof(float) == 4, "float must be 4 bytes.");
+    RETURN_ERR_IF_NOT(dataDims.size() == 2 && dataDims[1] > 8,
+                      "Fused data must be 2 dimemsional, with 8 extra columns "
+                      "for holding the per row float scale and offset");
```
rdzhabarov: nit: float scale and int32 offset.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -849,6 +849,98 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+        ElemKind::Int32ITy, {dataDims[0]}, "dataOffsets");
+
+    // Now strip out all the data from the fused Constant and place them
+    // correctly in their respective Constant payloads. First set the rwqData
+    // from fusedData, skipping over the scales/offsets.
+    auto fusedDataH = fusedDataC->getHandle<int8_t>();
+    auto rwqDataH = rwqData->getHandle<int8_t>();
+    for (size_t i = 0; i < dataDims[0]; i++) {
+      for (size_t j = 0; j < secondDimDataLen; j++) {
+        rwqDataH.at({i, j}) = fusedDataH.at({i, j});
```
rdzhabarov: is fused data stored as signed int8?

lib/Importer/Caffe2ModelLoader.cpp
jfix71: They're actually both float in Caffe2 -- [e.g. see here](https://caffe2.ai/docs/operators-catalogue.html#floattofused8bitrowwisequantized):

>bias as a 32-bit float

lib/Importer/Caffe2ModelLoader.cpp
jfix71: At this point in Glow, yes -- Caffe2 uses uint8, but we convert it to int8 when we load from `Int8GivenTensorFill` (prior to this point).

lib/Importer/Caffe2ModelLoader.cpp
rdzhabarov: Got it, interesting. other C2 quantized ops assumes different.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -1549,3 +1549,197 @@ TEST(caffe2, tensorFillsTest) {
...
+    output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
+  }
+
+  ASSERT_TRUE(indices);
+  ASSERT_TRUE(lengths);
+
+  ctx.allocate(indices)->getHandle<int64_t>() = {
+      1, 0, 2, 0, 1, 2, 2, 0,
+  };
+  ctx.allocate(lengths)->getHandle<int32_t>() = {
```
nadavrot: The format here is funny.

jfix71: Yeah for some reason this is what `clang-format` prefers 😕... If there are more than 5 elements then it all goes into a single line like `indices` above, but this is only 4 elements.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -849,6 +849,98 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+        ElemKind::FloatTy, {dataDims[0]}, "dataScales");
+    Constant *dataOffsets = G_.getParent()->createConstant(
+        ElemKind::Int32ITy, {dataDims[0]}, "dataOffsets");
+
+    // Now strip out all the data from the fused Constant and place them
+    // correctly in their respective Constant payloads. First set the rwqData
+    // from fusedData, skipping over the scales/offsets.
+    auto fusedDataH = fusedDataC->getHandle<int8_t>();
+    auto rwqDataH = rwqData->getHandle<int8_t>();
+    for (size_t i = 0; i < dataDims[0]; i++) {
```
artemrakhov: How will this work if `data` is a huge tensor?

jfix71:
@artemrakhov My understanding from @stoklund was that fusing was for efficiency at execution for an accelerator, not compile time. Backends that need it fused could always re-fuse it later on.

I was thinking this could be made more efficient/possible with huge Constants using the "offline Constants" approach we previously discussed, which seemed to be something we may need for other uses in the future too. If so, this unfusing wouldn't happen in `Caffe2ModelLoader` anymore -- instead you would add an "unfuse" expression, and later on a backend could add a "fuse" expression, and they would cancel each other out, meaning the fused data stays fused the whole time.

For fused Tensors in Glow, like Caffe2 uses -- it feels pretty hacky to me, just modifying the shape to include per-row scales/offsets and assume all users of fused Tensors know what's going on. If we want to go down that path, I'd prefer creating a fused Tensor class which has the correct shape and implements less efficient `at()`, `raw()` etc. that take into account the scales/offsets data fused in.

artemrakhov:
I agree, fusing is done for efficiency at execution, not compile time. At execution time, we have to fetch a lot of rows from `Data` and we are limited by memory bandwidth. Having to fetch 2 more single numbers per each row will increase memory pressure. This is true for every inference platform.

I see your point about fusing later per Backend request. But in reality this may take a long time and consume extra memory, due to the size of these Tensors. We'll definitely need a concept of deferred (lazy) optimizations, which will cancel each other out in this case.

I don't think that fused op is hacky. Caffe2 actually implements both. I think it's fine for all backends to be aware of this fused format.

jfix71:
Update: After an offline discussion we decided we should implement both fused and unfused nodes. Ideally higher up in the stack we will ensure that Glow receives the weights either in fused or unfused format, depending on what the backend of the host prefers.

I will update this PR to have protos that are not fused + update the Caffe2Importer case, and then will put up future PRs to support the fused version of RWQ-SLWS/protos.

jfix71:
I've updated the PR to not used fused scale/offsets. In a future PR I will add in new nodes for fused, as well as Caffe2 importer support for fused.

PR #2296, opened 2019-01-24
[CPUDeviceManager] Relax restriction of addNetwork to allow reusing the Module*.

nickgg:
*Description*: We assumed that all functions in the module would be added to a given device at the same time, but this isn't a useful or necessary restriction. Also simplified the module map, we didn't use the CompiledFunction* here.
*Testing*: CPUDeviceManager tests + a new test.
*Documentation*: n/a

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -32,12 +32,12 @@ bool CPUDeviceManager::isMemoryAvailable(uint64_t estimate) {
 void CPUDeviceManager::addNetworkImpl(const Module *module,
                                       FunctionMapTy functions,
                                       ReadyCBTy readyCB) {
-  auto modIt = modules_.find(module);
-  if (modIt != modules_.end()) {
-    // Already have a module with this ID.
-    // TODO: should we replace it?
-    readyCB(module, ResultCode::Failed);
-    return;
+  // First check for uniqueness of the function name.
+  for (const auto &func : functions) {
+    if (functions_.count(func.first) != 0) {
+      readyCB(module, ResultCode::Failed);
+      return;
+    }
   }
 
   // TODO: we should update usedMemory but we don't currently have a nice way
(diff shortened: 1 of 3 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -32,12 +32,15 @@ bool CPUDeviceManager::isMemoryAvailable(uint64_t estimate) {
...
                                       FunctionMapTy functions,
                                       ReadyCBTy readyCB) {
-  auto modIt = modules_.find(module);
-  if (modIt != modules_.end()) {
-    // Already have a module with this ID.
-    // TODO: should we replace it?
-    readyCB(module, ResultCode::Failed);
-    return;
+
+  std::set<std::string> &moduleFuncs = modules_[module];
```
gcatron: Does this check global name uniqueness or just within a module? What if there are two functions with the same name from different modules? Or is the expectation that is handled higher up the stack?

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -70,9 +73,9 @@ void CPUDeviceManager::evictNetworkImpl(const Module *module) {
     return;
   }
 
-  FunctionMapTy moduleFuncs = std::move(modIt->second);
-  for (const auto &func : moduleFuncs) {
-    functions_.erase(func.first);
+  std::set<std::string> moduleFuncs = std::move(modIt->second);
```
rdzhabarov: nit: not really need to move construct, just take a ref `const auto & moduleFuncs = modIt->second;`

tests/unittests/CPUDeviceManagerTest.cpp
```diff
@@ -300,6 +300,83 @@ TEST(CPUDeviceManagerTest, MultiModule) {
...
+  future.wait_for(2s);
+  EXPECT_EQ(future.get(), module.get());
+
+  std::tie(promise, future) = getFutureHelper<const Module *>();
+  cpuCoreDevice.addNetwork(module.get(), std::move(functions2),
+                           [&promise](const Module *module, ResultCode result) {
+                             callbackHelper(promise, module, result,
+                                            ResultCode::Ready);
+                           });
+  future.wait_for(2s);
```
rdzhabarov: why do we need to wait for 2s and then future.get(). Can't it be just blocking future.get().

nickgg: It could, but in the event we have a bug in the test where the cb doesn't fire it'll hang. This way it just fails out in a couple seconds.

lib/Backends/CPU/CPUDeviceManager.cpp
nickgg: You're right it should be functions_.

PR #2300, opened 2019-01-25
Enable calling into runtime from onnxifi

jackm321:
*Description*:
Enable calling into the Glow runtime from ONNXIFI through the HostManager.
This diff enables running onnxifi inference through HostManager once HostManager is implemented.

This diff is intended to get end-to-end support for the runtime working but there are many things that we cannot yet do such as multiple runs with the same Graph, running multiple onnxifi Graphs on the same onnxifi Backend, etc. To enable these things, some more larger refactoring steps will be needed in followups including 1. Moving all resources necessary for running graphs into BackendId, 2. Abstracting BackendId to enable separate implementations for HostManager and ExecutionEngine backed BackendIds (module is still always owned by ExecutionEngine no matter what), 3 move all graph state into the graph itself to enable running multiple times with different contexts, 4 move components necessary for compatibility check into runtime and separate the ExecutionEngine from the HostManager implementations.

Depends on #2284.

*Testing*:
`ninja test`, CI

*Documentation*:
doxygen

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -23,8 +23,8 @@
 namespace glow {
 namespace onnxifi {
 namespace {
-const std::string inferenceFunctionName = "inference";
-const std::string compatibilityFunctionName = "check";
+const char *inferenceFunctionName = "inference";
+const char *compatibilityFunctionName = "check";
 } // namespace
 
 onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
@@ -156,7 +156,7 @@ void Graph::run(
     phs.push_back(var);
   }
 
-  auto ctx = std::make_unique<Context>();
+  auto ctx = llvm::make_unique<Context>();
 
   // Run inference.
   auto &EE = backendPtr_->getEE();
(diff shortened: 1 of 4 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -22,17 +22,22 @@
 
 namespace glow {
 namespace onnxifi {
+namespace {
+const std::string inferenceFunctionName = "inference";
```
nadavrot: Why is this std::string and not "const char*" ?

jackm321: No reason, I'll change this and the below std::string

lib/Onnxifi/Base.cpp
```diff
@@ -146,26 +156,42 @@ void Graph::run(
     phs.push_back(var);
   }
 
+  auto ctx = std::make_unique<Context>();
```
rdzhabarov: make this llvm::make_unique.

@bertmaher recently fixed bunch of places to make glow g++-4.8.5 compatible.

jackm321:
Landing this now so we can keep moving with refactoring the onnxifi interface to get ready for full HostManager integration once it's ready. Will add integration tests once HostManager is ready.

PR #2301, opened 2019-01-25
Fix g++-4.8.5 build

bertmaher:
*Description*: As long as CentOS 7 is a target platform we need to support gcc 4.8.5.  Sorry.  I'm going to see if we can use gcc 4.8.5 in CI to enforce this.
*Testing*: ninja test
*Documentation*: n/a


rdzhabarov:
Nice!

PR #2302, opened 2019-01-25
[ci] Run coverage test with gcc-4.8.5

bertmaher:
*Description*: For compatibility with CentOS 7 I want to keep Glow building with gcc 4.8.5.
*Testing*: this diff
*Documentation*: n/a


bertmaher:
Looks like it correctly failed before #2301 landed, now let's see if it passes!

Issue #2303, opened 2019-01-25
onnxtxt generation for tests


tlepley-cadence:
I'm fixing an issue in the ONNX importer and as usual, I'd like to create a test. 
For this, I need to create a onnxtxt file. 

Until now, I was doing it manually, because it was simple cases. This time, I would prefer to copy-paste a subset of an actual ONNX model. But I'm struggling for converting an onnx file into an onnxtxt file. Looking at the ONNX code, it doesn't seem they have an option to dump the text version and I didn't find much information on the web.

Any idea how to convert onnx models into onnxtxt files ?

bertmaher:
An onnx model is just a protobuf, so you should be able to use a simple python script like
```
import onnx
model = onnx.load("model.onnx")
open("model.onnxtxt", "w").write(str(model))
```

tlepley-cadence:
@bertmaher It works well this way. Thanks !

PR #2304, opened 2019-01-25
[ONNX] Fix a problem with batchnorm that can optionally have more tha…

tlepley-cadence:
*Description*:

Some ONNX models have a BatchNormalization node that declare the 4 optional outputs. The ONNX importer assumes there is only one output because the 4 optional outputs are not supported by Glow. The importer then fails in `addNodeAsOutput`.

The issue is that some models declare these optional outputs without to use them at all in the model. In this case, the importer should not fail, while it fails currently.

This PR proposes a simple fix to this issue. ~~I've put this PR as WIP because I'm waiting for an answer to issue #2303 before writing a test.~~

ONNX Batchnorm: https://github.com/onnx/onnx/blob/master/docs/Operators.md#BatchNormalization

*Testing*: Test "onnx.batchNormPR2304"

*Documentation*: N/A

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -107,13 +107,15 @@ protected:
   /// Associate the name of operation outputs to a NodeValues corresponding to
   /// node \p node. If \p numOutputs is lower than 0, then all outputs are
   /// associated. Otherwise, the first \p numOutputs outputs are associated.
-  void addNodeAsOutput(const OpType &op, Node *node, int numOutputs = -1) {
-    numOutputs = ((numOutputs < 0) || numOutputs > op.output_size())
-                     ? op.output_size()
-                     : numOutputs;
+  llvm::Error addNodeAsOutput(const OpType &op, Node *node,
+                              int numOutputs = -1) {
+    RETURN_ERR_IF_NOT(numOutputs <= op.output_size(),
+                      "Can't register more than outputs in the operation.");
+    numOutputs = (numOutputs < 0) ? op.output_size() : numOutputs;
     for (int i = 0; i < numOutputs; i++) {
       nodeValueByName_[op.output(i)] = NodeValue(node, i);
     }
+    return llvm::Error::success();
   }
 
   /// Load pre-trained weights from \p weightDescriptors.
@@ -135,7 +137,7 @@ protected:
     ASSIGN_VALUE_OR_RETURN_ERR(in,
                                getNodeValueOrCreateConstantByName(op.input(0)));
     auto *R = G_.createRELU(opName, in);
-    addNodeAsOutput(op, R);
+    RETURN_IF_ERR(addNodeAsOutput(op, R));
     return llvm::Error::success();
   }
 
@@ -145,7 +147,7 @@ protected:
     ASSIGN_VALUE_OR_RETURN_ERR(in,
                                getNodeValueOrCreateConstantByName(op.input(0)));
     auto *S = G_.createSigmoid(opName, in);
-    addNodeAsOutput(op, S);
+    RETURN_IF_ERR(addNodeAsOutput(op, S));
     return llvm::Error::success();
   }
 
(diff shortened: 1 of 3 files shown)


jackm321:
I'm trying to figure out why does the `addNodeAsOutput()` fail if there are some optional outputs that glow doesn't support? I see that there will be some extra NodeValues in nodeValueByName_ that won't get accessed but whats the error?
The solution seems fine to me though.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -104,9 +104,15 @@ class CommonOperatorLoader : public ProtobufLoader {
...
   virtual bool hasMultidirectionalBroadcast(const llvm::StringRef typeName) = 0;
 
-  void addNodeAsOutput(const OpType &op, Node *R) {
-    for (int i = 0, e = op.output_size(); i < e; i++) {
-      nodeValueByName_[op.output(i)] = NodeValue(R, i);
+  /// Associate the name of operation outputs to a NodeValues corresponding to
+  /// node \p node. if \p numOutputs is lower than 0, then all outputs are
+  /// associated. Otherwise, the first \p numOutputs outputs are associated.
+  void addNodeAsOutput(const OpType &op, Node *node, int numOutputs = -1) {
+    numOutputs = ((numOutputs < 0) || numOutputs > op.output_size())
```
jfix71: Why do you check `numOutputs > op.output_size()` here? This seems like it would represent a bug, so wouldn't be something we'd want to silently handle/ignore.

jfix71:
@jackm321 When constructing the `NodeValue` this would cause an assertion error -- it asserts `resNo < N->getNumResults()`.

include/glow/Importer/CommonOperatorLoader.h
tlepley-cadence: I was trying to avoid adding support of error handling in this function, what would require to change all calls to `addNodeAsOutput` (there are many). Would an `assert` be good enough for you ?

tlepley-cadence:
@jackm321 @jfix71 We indeed get the following error:
`glow/lib/Graph/NodeValue.cpp:30: glow::NodeValue::NodeValue(glow::Node*, unsigned int): Assertion `resNo < N->getNumResults() && "Invalid result number"' failed.`

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -647,7 +647,14 @@ ONNXModelLoader::loadBatchNormalization(const ONNX_NAMESPACE::NodeProto &op,
   auto *node = G_.createBatchNormalization(opName, in, biasV, scaleV, meanV,
                                            varV, 1, epsilon);
 
-  addNodeAsOutput(op, node);
+  // BatchNormalization has 4 optional outputs that are not supported by glow.
+  // Then: 1/ In case the optional outputs are present and used by other
+  // operations of the model, then the import should fail. 2/ In case the
+  // optional outputs are declared but not used, the import should succeed. By
+  // registing only the mandatory output, we make sure the import will fail if
```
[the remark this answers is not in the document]
tlepley-cadence: Fixed

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -104,9 +104,15 @@ class CommonOperatorLoader : public ProtobufLoader {
   /// for multidirectional broadcasting.
   virtual bool hasMultidirectionalBroadcast(const llvm::StringRef typeName) = 0;
 
-  void addNodeAsOutput(const OpType &op, Node *R) {
-    for (int i = 0, e = op.output_size(); i < e; i++) {
-      nodeValueByName_[op.output(i)] = NodeValue(R, i);
+  /// Associate the name of operation outputs to a NodeValues corresponding to
+  /// node \p node. if \p numOutputs is lower than 0, then all outputs are
```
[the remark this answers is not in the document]
tlepley-cadence: Fixed

tlepley-cadence:
Added a reproducer test.

include/glow/Importer/CommonOperatorLoader.h
jackm321: Agree this seems like a bug, please use proper error handling for the loaders, most likely `RETURN_ERR_IF_NOT` in this case

tlepley-cadence: Done.

PR #2306, opened 2019-01-25
[CPUDeviceManager] Factory for creating devices, additions for HostManager

nickgg:
*Description*: add a createDeviceManager static method for creating devices without needing to pull in backend specific code from lib. Also a couple of requests from @gcatron for use by the HostManager: adding a name parameter to DeviceManager constructor, and changing evictNetwork from taking the Module* to a function name.
*Testing*: unittests with asan.
*Documentation*:

### DIFF
--- a/include/glow/Backends/QueueBackedDeviceManager.h
+++ b/include/glow/Backends/QueueBackedDeviceManager.h
@@ -32,7 +32,7 @@ protected:
   std::atomic<runtime::RunIdentifierTy> nextIdentifier_{1};
 
 public:
-  QueueBackedDeviceManager(BackendKind backend);
+  QueueBackedDeviceManager(BackendKind backend, llvm::StringRef name);
   virtual ~QueueBackedDeviceManager();
 
   /// Initialize the device.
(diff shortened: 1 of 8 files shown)


include/glow/Backends/QueueBackedDeviceManager.h
```diff
@@ -32,7 +32,7 @@ class QueueBackedDeviceManager : public DeviceManager {
   std::atomic<runtime::RunIdentifierTy> nextIdentifier_{1};
 
 public:
-  QueueBackedDeviceManager(BackendKind backend);
+  QueueBackedDeviceManager(BackendKind backend, std::string name);
```
nadavrot: Should this be stringRef?

lib/Backends/DeviceManagers.cpp
```diff
@@ -0,0 +1,71 @@
+/**
```
nadavrot: @nickgg To add a new backend we now have to add code in two places. One way to simplify this and make it one place would be to make the device manager a member of the backend. Something like Backend::createDeviceManager. 

This is just an idea. I am not sure if this is a problem that we need to solve.

include/glow/Backends/QueueBackedDeviceManager.h
nickgg: At a certain point we want to store a copy, but we can do that at the last stage.

lib/Backends/DeviceManagers.cpp
nickgg: Yeah I definitely think that would be more convenient, my concern here would be coupling between the Backend and the DeviceManager.

In an ideal world I think the compile time and runtime parts of glow would be distinct, connected by the Provisioner and the CompiledFunction.

Issue #1465, opened 2018-08-15
[Quantization] Profile and Quantize components Nodes of lowered Nodes


jfix71:
Right now we do profiling on the graph prior to lowering. I will use `FullyConnected` nodes as a running example. They are lowered into a `MatMul` and `BatchedAdd` node. However the `FullyConnected` nodes have only their output profiled, meaning the output of the last node (`BatchedAdd`). This means we do not have profiling information about the `MatMul`, and we are forced to use the profile from the `BatchedAdd` for the `MatMul` when lowering the quantized `FullyConnected` node. This means we could lose some precision unnecessarily.

We could instead lower all nodes prior to profiling. We would then get profiles for both the `MatMul` and `BatchedAdd`. This way we do not lose any information about the components of the lowered `FullyConnected` node. After profiling we load this profile information back in for quantization, which now includes all component node's profiles.

At this point, one option to proceed is to apply quantization to the unlowered graph based on the lowered component profiles (e.g. still quantize the `FullyConnected` node using the `BatchedAdd` profile). This could be necessary because we currently quantize prior to compiling, which is where we do optimizations and lowering. We would then need to additionally keep track of each unlowered node's component node's quantization parameters, so that we can use these parameters when the nodes are lowered (e.g. make sure when we lower the quantized `FullyConnected` node we can find and use its component `MatMul` node's quantization parameters).

So it seems we would need to create a map from unlowered nodes to their lowered component profiles/quantization parameters, e.g. map from the `FullyConnected` Node to the `BatchedAdd` and `MatMul` profiles/quantization parameters. This map would need to be passed into `lower()`.

I believe other options would require a redesign of the order in which we apply quantization vs. optimizations and lowering.

qcolombet:
> I believe other options would require a redesign of the order in which we apply quantization vs. optimizations and lowering.

Should we shot for a redesign instead of patching the current one?

jfix71:
So I think the main redesign option would be to lower before quantizing. I believe this would involve lowering everything and quantizing it all, then pattern matching groups of nodes back to unlowered nodes if the backend requested so. This means we would need to do this "unlowering" before any optimizations occurred to ensure they do not prevent any unlowering that the backend wants.

[...]

PR #2311, opened 2019-01-29
Lower everything before profiling, allowing for lowered profile to be gathered

jfix71:
This is a somewhat controversial change, but something I've considered for a while and it seems it could solve a couple pain points we're running into right now.

Basically, when doing profiling, lower *everything*, but keep track of what was lowered from what. This allows us to get quantization parameters of lowered nodes, while still allowing for profiles to stay backend-agnostic.

E.g. if we lower an FC to MM + BA, we keep track of that via a StringStringMap, mapping between their quantization output names (e.g. `ba:0` maps to `fc:0`, as we did `FCResult.replaceAllUsesOfWith(BAResult)`).

This map is used to write both the unlowered and lowered names into the profile. So before this PR, the profile would contain just `fc:0`. With this PR, the profile would additionally contain `mm:0` and `ba:0`, where the quantization parameters for `fc:0` and `ba:0` are the same. This means `mm:0` is slightly better specified based on actually gathering a profile for it. A backend can then find the quantization parameters of any nodes regardless of whether it lowers a node or not.

This is much more important for more complex nodes we lower, for example `SigmoidCrossEntropyWithLogitsNode`, which really can't be accurately profiled/quantized right now because it is lowered. This could enable us to add actual nodes for others that are "directly lowered" in Graph.cpp, for example LSTM.

This would also enable us to get rid of `getLoweredComponentInfo()` from https://github.com/pytorch/glow/pull/2235, which I'm not a huge fan of.

Fixes https://github.com/pytorch/glow/issues/1465

### DIFF
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -240,6 +240,13 @@ void Loader::compile(Context &ctx) {
     // perform CSE, etc.
     ::optimize(F_, glow::CompilationMode::Infer);
 
+    // Lower everything, keeping track of what NodeValues were lowered to other
+    // NodeValues via the loweredMap_. This loweredMap_ is passed to
+    // generateNodeQuantizationInfos() when writing out the profile, allowing
+    // for both lowered and unlowered NodeValues to find their quantization
+    // parameters.
+    ::lower(F_, *EE_.getBackend(), &loweredMap_);
+
     // Instrument the graph to capture profiles for nodes' outputs.
     F_ = ::profileQuantization(ctx, F_);
   }
@@ -260,6 +267,12 @@ void Loader::compile(Context &ctx) {
     // part of the code we repeat the same transformation in order to create
     // the same graph structure.
     ::optimize(F_, glow::CompilationMode::Infer);
+
+    // Lower as the backend prefers. When generating the profile everything was
+    // lowered, however all lowered and unlowered components have a profile, and
+    // so the backend can lower however it prefers and always find all of its
+    // NodeValue's quantization parameters.
+    ::lower(F_, *EE_.getBackend());
 
     auto quantizationInfos = deserializeFromYaml(loadProfileFileOpt);
 
@@ -328,8 +341,8 @@ void Loader::generateAndSerializeQuantizationInfos(Context &ctx) {
   assert(!dumpProfileFileOpt.empty() &&
          "Filename to dump serialized profile to must not be empty.");
   std::vector<NodeQuantizationInfo> QI =
-      quantization::generateNodeQuantizationInfos(ctx, F_, quantizationSchema,
-                                                  quantizationPrecision);
+      quantization::generateNodeQuantizationInfos(
+          ctx, F_, loweredMap_, quantizationSchema, quantizationPrecision);
   serializeToYaml(dumpProfileFileOpt, QI);
 }
 
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -131,6 +131,7 @@ struct Model {
   Placeholder *seqLength_;
   Placeholder *output_;
   Context ctx;
+  LoweredNamesMap loweredMap_;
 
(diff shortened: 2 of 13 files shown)


jackm321:
@jfix71 I like this approach as compared to #2235. Comparing the two, this seems cleaner and easier to maintain.

jfix71:
@rdzhabarov Yeah good point on the new `optimize()` call after the new `lower()`. I had added the new `optimize()` call because I wanted DCE. Perhaps we could instead just expose DCE by itself so it could be called alone. This should ensure the graph stays the same when the lowering is different.

I verified it works on Resnet50 and VGG19 as well as a sparse net, profiling w/ the Interpreter and running with the CPU backend (I had to add some fixes to the FunctionQuantizer too for unhandled cases on IR constraints -- it looks like this is what is causing the `lenet_mnist` failure on CI). I verified the model outputs are still correct, and that the MM/BA lowered from FCs now have different parameters (whereas before they were the same).

rdzhabarov:
> I had added the new optimize() call because I wanted DCE. Perhaps we could instead just expose DCE by itself so it could be called alone

+1

jfix71:
@nadavrot I thought more about your suggestion of simply setting the newly created node's name to be the same as the old node we are lowering/replacing when we do `replaceAllUsesOfWith()` during lowering.

Unfortunately this won't work cleanly when we are replacing nodes that have multiple results. E.g. if we lower a single node that has two results into two new nodes that each have one result, the two new nodes would need different names. Additionally then when we generate their names for the profile the result numbers are appended to the end of the name. So before if we had a node with 2 results named `"node"`, their resulting names in the profile would be `"node:1"` and `"node:2"`.

We'd need to redesign naming of nodes/how we generate the output names for profiling, and I'm not sure there's a great workaround on first thought. E.g. we could do something like add a unique name for each node's result instead of just each node, I guess?

tools/loader/Loader.cpp
```diff
@@ -240,6 +240,14 @@ void Loader::compile(Context &ctx) {
     // perform CSE, etc.
     ::optimize(F_, glow::CompilationMode::Infer);
 
+    // Lower everything, keeping track of what NodeValues were lowered to other
+    // NodeValues via the loweredMap_. This loweredMap_ is passed to
+    // generateNodeQuantizationInfos() when writing out the profile, allowing
+    // for both lowered and unlowered NodeValues to find their quantization
+    // parameters.
+    ::lower(F_, *EE_.getBackend(), &loweredMap_);
+    ::DCE(F_);
```
jackm321: Why is DCE not just the last step of `glow::lower()`? Is there any time we don't want to DCE after lowering?

nadavrot:
@jfix71 Adding the optional stringmap parameter to the lowering phase complicates the design by adding a dependency and a data structure that is shared between two (already very complex) passes. Can we continue to iterate and think of ways to simplify this?

tools/loader/Loader.cpp
jackm321: We don't always optimize for example in the onnxifi check compatibility case we don't. If the overhead of DCE is small-ish I think we should just automatically DCE after lowering to avoid leaking calls to DCE from places outside of optimizers like here.

rdzhabarov: > We don't always optimize for example in the onnxifi check compatibility case we don't.

that's a bug :) You keep both lowered and prior lowering nodes for compatibility validation, which we should not do.

jackm321: @rdzhabarov yep it is, I think we should add DCE after lowering to fix it

jfix71: @jackm321 I think that adding `DCE` at the end of `lower()` could be reasonable. I think in practice DCE is pretty quick. Do you want to put up an issue or PR for that so we can discuss further?

jackm321: @jfix71 will do

PR #2312, opened 2019-01-29
[ONNXIFI] Move ExecutionEngine to Graph

jackm321:
*Description*:
Enable running multiple onnxifi graphs on the same onnxifi backend by moving the ExecutionEngine (and thus all state) to the onnxifi graph. This diff also enables ExecutionEngine to run with Backends that it does not own.
Next steps include:

- Make Backend::isOpSupported and Backend::shouldLower static so we can call them from HostManager as well without issue.
- Make onnxifi::Graph and onnxifi::Backend abstract classes and create a new type of graph and backend that integrates more closely with the HostManager interface.

*Testing*:
`ninja test`
sync to fbcode and run tests

*Documentation*:
doxygen

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -39,7 +39,11 @@ class ExecutionEngine final {
   /// The Module that represents the high-level program.
   Module M_;
   /// The network execution backend.
-  std::unique_ptr<Backend> backend_;
+  Backend *backend_ = nullptr;
+  /// Whether or not the ExecutionEngine owns the backend or is just using
+  /// a backend provided from elsewhere. If ownsBackend is true,
+  /// ~ExecutionEngine will delete the backend_.
+  bool ownsBackend_ = false;
   /// A glow function compiled for this ExecutionEngine's backend.
   std::unique_ptr<CompiledFunction> function_;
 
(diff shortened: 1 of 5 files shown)


rdzhabarov:
@jackm321 is it WIP? ready for review?

lib/Onnxifi/Base.h
```diff
@@ -16,6 +16,7 @@
 #ifndef GLOW_ONNXIFI_BASE_H
 #define GLOW_ONNXIFI_BASE_H
 
+#include "glow/Backends/Backend.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
```
gcatron: Do you need to include Backend.h? Can it be a forward declaration?

jackm321: This is needed for `glow::createBackend()`

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -55,6 +63,9 @@ class ExecutionEngine final {
   /// Set the code generator to a custom \p backend.
   void setBackend(Backend *backend);
 
+  /// Set the code generator to use a custom \p backend without owning it.
+  void useBackend(Backend *backend);
```
rdzhabarov: Can you keep the old version of setBackend and introduce a param on whether backend is owned or not?

jackm321: Sure

jackm321:
@rdzhabarov ready now

PR #2318, opened 2019-01-30
[ci] Enable CPU backend on Travis

bertmaher:
*Description*: Should work.  OpenCL could be added with pocl but I'm not sure it's worth fiddling with now.
*Testing*: CI
*Documentation*: n/a


rdzhabarov:
There were some issues with capturing gcov profile iirc, I digged a bit but was not able to fix it last time. If you could make it work let's merge this change :)

I'd not worry about OpenCL.

bertmaher:
Seems like it worked... It looks like it doesn't upload coverage from a fork, so let's see what happens if I actually land this.

PR #2319, opened 2019-01-30
[Runtime] Add runtime Executor class

SplitInfinity:
**Description**
graphs using a thread pool. The  workers essentially do a
concurrent breadth-first search through the node graph to execute all
nodes, repeatedly copying partition components outputs to their descendents along
the way.

**Testing**
This commit adds eight unit tests for the following scenarios:
- empty graph
- single node
- several instances of a single node graph at once
- duplicate runIds
- multi node graph
- multi node graph with a node that fails
- multi node graph with nodes running on multiple devices
- several instances of a multi node graph at once

**Documentation**
Comments.

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -29,13 +29,19 @@ struct RuntimeSymbolInfo {
   size_t offset;
   /// Type of symbol.
   Type type;
+  /// Is the symbol an input for the function.
+  bool input;
+  /// Is the symbol an output for the function.
+  bool output;
 };
+
+using SymbolTableTy = std::unordered_map<std::string, RuntimeSymbolInfo>;
 
 /// Contains the information needed to be passed forward from compile time to
 /// runtime. In order to allocate and initialize memory.
 class RuntimeBundle {
   /// Map from symbol name to a RuntimeSymbolInfo.
-  std::unordered_map<std::string, RuntimeSymbolInfo> symbolTable_;
+  SymbolTableTy symbolTable_;
   /// Pointer to memory containing the weights for execution.
   uint8_t *constants_;
   /// Amount of memory needed for weights.
@@ -60,6 +66,8 @@ public:
   size_t getValueOffset(const Named *v) const;
   /// Helper function, gets symbol info for \p v.
   RuntimeSymbolInfo getSymbolInfo(const Named *v) const;
+  /// Get a const reference to the symbol table.
+  const SymbolTableTy &getSymbolTable() const { return symbolTable_; }
   /// At compile time condense constants to a single block of memory.
   /// This allows the graph to go away after compile time.
   /// Allocates a block of memory of size \p constantMaxSize then walks the
@@ -69,8 +77,8 @@ public:
   void collectConstants(const Module *M);
 
   RuntimeBundle() = default;
-  RuntimeBundle(std::unordered_map<std::string, RuntimeSymbolInfo> &symbolTable,
-                size_t constWeight, size_t mutableWeight, size_t activations)
+  RuntimeBundle(SymbolTableTy &symbolTable, size_t constWeight,
+                size_t mutableWeight, size_t activations)
       : symbolTable_(std::move(symbolTable)), constants_(nullptr),
         constantWeightVarsMemSize_(constWeight),
         mutableWeightVarsMemSize_(mutableWeight),
--- a/lib/Runtime/Executor/ThreadPoolExecutor.cpp
+++ b/lib/Runtime/Executor/ThreadPoolExecutor.cpp
@@ -1,0 +1,357 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ThreadPoolExecutor.h"
+
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Graph/Context.h"
+
+#include <queue>
+#include <unordered_set>
+
+namespace glow {
+namespace runtime {
+
+ExecutionState::ExecutionState(RunIdentifierTy id, const DAGNode *root,
+                               std::unique_ptr<Context> resultContext,
+                               DoneCbTy doneCb)
+    : runId_(id), cb_(doneCb), resultCtx_(std::move(resultContext)),
+      inflightNodes_(0), resultCode_(ResultCode::Ready) {
+  // Create a queue for the breadth-first traversal through the graph.
+  std::queue<const DAGNode *> bfsQueue;
+
+  // Place the root nodes in the queue.
+  for (const auto &node : root->children) {
+    bfsQueue.push(node);
+  }
+
+  // Breadth-first search.
+  while (!bfsQueue.empty()) {
+    // Get the next node in the BFS queue.
+    const DAGNode *node = bfsQueue.front();
+    bfsQueue.pop();
+
+    // Make a counter for the number of node parents done.
+    nodeParentsDone_[node] = 0;
+
+    // Make an (empty) input Context for the node.
+    inputCtxs_.insert(std::make_pair(node, llvm::make_unique<Context>()));
+
+    // Push all unvisited children onto the BFS queue.
+    for (const auto &child : node->children) {
+      // Use nodeParentsDone_ as a set of nodes that have been visited already
+      // to avoid visiting a node more than once.
+      if (!nodeParentsDone_.count(child)) {
+        bfsQueue.push(child);
+      }
+    }
+  }
+}
+
+void ExecutionState::insertIntoNodeCtx(const DAGNode *node, Placeholder *P,
+                                       Tensor &&T) {
+  // Get a raw pointer to the input Context for the node. It should have
+  // been created in the constructor.
+  auto ctxIt = inputCtxs_.find(node);
+
+  if (ctxIt == inputCtxs_.end()) {
+    assert(!"Input context not found but should exist!");
+  }
+
+  Context *ctx = (ctxIt->second).get();
+  assert(ctx && "Input context for node is null");
+
+  // Insert the placeholder-tensor pair.
+  std::lock_guard<std::mutex> lock(contextMtx_);
+  ctx->insert(P, std::move(T));
+}
+
+std::unique_ptr<Context>
+ExecutionState::getUniqueNodeContextPtr(const DAGNode *node) {
+  // The input Context for the node should have been created in the constructor.
+  auto ctxIt = inputCtxs_.find(node);
+
+  if (ctxIt == inputCtxs_.end()) {
+    assert(!"Input context not found but should exist!");
+  }
+
+  return std::move(ctxIt->second);
+}
+
+void ExecutionState::incrementInflightNodes(unsigned increment) {
+  inflightNodes_ += increment;
+}
+
+bool ExecutionState::decrementInflightNodes(unsigned decrement) {
+  // fetch_sub must be used here so that the function returns true to only one
+  // caller.
+  unsigned previousValue = inflightNodes_.fetch_sub(decrement);
+
+  // The decrement should never be more than the value of the counter at the
+  // time of decrement.
+  if (previousValue < decrement) {
+    assert(!"More decrements than increments to inflight nodes!");
+  }
+
+  // Return true when the counter hits zero.
+  return (previousValue == decrement);
+}
+
+bool ExecutionState::incrementNodeParentsDone(const DAGNode *node,
+                                              unsigned increment) {
+  // Get the parents done counter for the node. It should have
+  // been created in the constructor.
+  auto it = nodeParentsDone_.find(node);
+
+  if (it == nodeParentsDone_.end()) {
+    assert(!"Node parents done counter should exist but not found!");
+  }
+
+  // fetch_add must be used here so that the function returns true to only
+  // one caller.
+  unsigned numParents = (node->parents).size();
+  unsigned previousValue = (it->second).fetch_add(increment);
+  unsigned newValue = previousValue + increment;
+
+  // The new value of the counter cannot exceed the number of parents that
+  // the node has.
+  if (newValue > numParents) {
+    assert(!"Node parents done counter incremented beyond limit!");
+  }
+
+  // Return true only when the counter hits the total numer of parents.
+  return (newValue == numParents);
+}
+
+void ExecutionState::insertIntoResultCtx(Placeholder *P, Tensor &&T) {
+  // The result Context should have been been created in the constructor
+  // and should not yet have been moved out if this function is being called.
+  assert(resultCtx_ && "Execution result context should exist!");
+  std::lock_guard<std::mutex> lock(contextMtx_);
+  Tensor *tensor = resultCtx_->get(P);
+
+  if (tensor) {
+    *tensor = std::move(T);
+  } else {
+    resultCtx_->insert(P, std::move(T));
+  }
+}
+
+std::unique_ptr<Context> ExecutionState::getUniqueResultContextPtr() {
+  // The result Context should have been been created in the constructor.
+  assert(resultCtx_ && "Execution result context should exist!");
+  return std::move(resultCtx_);
+}
+
+Context *ExecutionState::getRawResultContextPtr() const {
+  // The result Context should have been been created in the constructor
+  // and should not yet have been moved out if this function is being called.
+  assert(resultCtx_ && "Execution result context should exist!");
+  return resultCtx_.get();
+}
+
+void ExecutionState::setResultCode(const ResultCode resultCode) {
+  // If resultCode_ is ResultCode::Failed, that should "stick". In other words,
+  // once a run has failed due to one node, the fact that the other nodes
+  // executed sucessfully does not change the result code of the run.
+  if (resultCode_ != ResultCode::Failed) {
+    resultCode_ = resultCode;
+  }
+}
+
+ThreadPoolExecutor::~ThreadPoolExecutor() {
+  // Lock executionStatesMutex_ here to satisfy TSAN.
+  std::lock_guard<std::mutex> lock(executionStatesMutex_);
+  (void)lock;
+}
+
+void ThreadPoolExecutor::run(const DAGNode *root,
+                             std::unique_ptr<Context> context,
+                             RunIdentifierTy runId, DoneCbTy cb) {
+  // If list of roots is empty, there is nothing to do. Give back the
+  // context so the caller can reuse it.
+  if (!root) {
+    cb(runId, ResultCode::Executed, std::move(context));
+    return;
+  }
+
+  std::shared_ptr<ExecutionState> executionState = nullptr;
+  {
+    std::lock_guard<std::mutex> lock(executionStatesMutex_);
+
+    // If the given run ID corresponds to a run already in progress, there is
+    // also nothing to do, but return an error. Give back the context so the
+    // caller can reuse it.
+    if (executionStates_.find(runId) != executionStates_.end()) {
+      cb(runId, ResultCode::Failed, std::move(context));
+      return;
+    }
+
+    // Otherwise, create execution state tracker object for this run ID.
+    executionState = std::make_shared<ExecutionState>(
+        runId, root, std::move(context), std::move(cb));
+    executionStates_.insert(std::make_pair(runId, executionState));
+  }
+
+  // Execute all child nodes of root.
+
+  // Mark the child nodes as "inflight" (i.e. currently executing). This must be
+  // done here instead of inside executeDAGNode() so that a node can be
+  // executed while placeholders are being propagated for the next node without
+  // the callback for that node deleting the execution state.
+  executionState->incrementInflightNodes((root->children).size());
+
+  for (auto const &node : root->children) {
+    // Propagate placeholders from the given starter Context into the input
+    // Context for the current node being processed.
+    propagatePlaceholdersForNode(executionState, node,
+                                 executionState->getRawResultContextPtr());
+
+    // Execute the node.
+    executeDAGNode(executionState, node);
+  }
+}
+
+void ThreadPoolExecutor::propagatePlaceholdersForNode(
+    std::shared_ptr<ExecutionState> executionState, const DAGNode *node,
+    const Context *ctx) {
+  // Get the symbol table for the node.
+  const SymbolTableTy &symbolTable = (node->runtimeBundle).getSymbolTable();
+
+  for (const auto &symbolPair : symbolTable) {
+    const auto &symbolName = symbolPair.first;
+    const auto &symbolInfo = symbolPair.second;
+
+    if (symbolInfo.input) {
+      // If the symbol is an input, look for a Placeholder in ctx with
+      // the same name and copy the corresponding Tensor into the input Context
+      // being prepared for the node.
+      auto *placeholder = ctx->getPlaceholderByName(symbolName);
+
+      if (placeholder) {
+        const auto *tensor = ctx->get(placeholder);
+        executionState->insertIntoNodeCtx(node, placeholder, tensor->clone());
+      }
+    }
+  }
+}
+
+void ThreadPoolExecutor::executeDAGNode(
+    std::shared_ptr<ExecutionState> executionState, const DAGNode *node) {
+  // If execution has already failed due to another node, don't bother running
+  // this one.
+  if (executionState->getResultCode() == ResultCode::Failed) {
+    // Mark the node as no longer executing.
+    executionState->decrementInflightNodes();
+    return;
+  }
+
+  // Get the DeviceManager that can run the node.
+  auto deviceManagerIt = deviceManagers_.find(node->deviceID);
+
+  if (deviceManagerIt == deviceManagers_.end()) {
+    // Mark the node as no longer executing.
+    executionState->decrementInflightNodes();
+    executionState->setResultCode(ResultCode::Failed);
+    return;
+  }
+
+  std::shared_ptr<DeviceManager> deviceManager = deviceManagerIt->second;
+
+  // Get the Context containing all of the inputs for the node.
+  std::unique_ptr<Context> nodeCtx =
+      executionState->getUniqueNodeContextPtr(node);
+  // Run the node using the DeviceManager.
+  deviceManager->runFunction(
+      node->name, std::move(nodeCtx),
+      [this, executionState, node](RunIdentifierTy id, ResultCode resultCode,
+                                   std::unique_ptr<Context> resultCtx) {
+        // Immediately move the handling of the result onto threadPool_ to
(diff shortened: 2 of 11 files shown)


SplitInfinity:
I didn't realize I had to reopen the old PR before pushing again, so I had to make a new one. Sorry!

tests/unittests/ExecutorTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+              /*outputs=*/{deltaOut.get()});
+  dag.addNode("eps", testDeviceId, /*parents=*/{gamma},
+              /*inputs=*/{epsIn.get()},
+              /*outputs=*/{epsOut.get()});
+
+  // Create the input Context to pass to Executor::run().
+  auto inputContext = std::make_unique<Context>(
+      cloneContext(ctx_, /*whitelist=*/{alphaIn.get(), betaIn.get()}));
+
+  // Create input/output Contexts for all DAGNodes. The input Contexts are
```
SplitInfinity: The code here is super verbose. I thought about folding it into `ExecutorTestDAG::addNode`, but I think that `ExecutorTestDAG` should only create the DAG so that it can be reused in other places if needed.

tests/unittests/ExecutorTest.cpp
```diff
@@ -0,0 +1,501 @@
...
+  auto epsOutputContext = std::make_unique<Context>(
+      cloneContext(ctx_, /*whitelist=*/{epsOut.get()}));
+
+  // Create the Context that Executor::run() is expected to return.
+  auto refResultContext =
+      cloneContext(ctx_, /*whitelist=*/{deltaOut.get(), epsOut.get()});
+
+  // Register the expected input Contexts and desired output Contexts for
+  // all nodes with the TestDeviceManager.
+  deviceManager->registerResult("alpha", 0, ResultCode::Executed,
```
SplitInfinity: Similar issue here. I thought about folding this into `ExecutorTestDAG` but I think that class will be more useful if it is focused only on building valid DAGs.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,338 @@
...
+ */
+
+#include "ThreadPoolExecutor.h"
+
+#include <queue>
+
+namespace glow {
+namespace runtime {
+
+ExecutionState::ExecutionState(RunIdentifierTy id, const DAGNode *root,
```
SplitInfinity: Hm. I initially wrote this constructor in order to create all required `Contexts` beforehand and to insert the required `Placeholders` into them without backing `Tensors` to avoid synchronization and locking later. I quickly realized `Context` doesn't support that, but left this as is. Can we afford to take the time to do all this setup beforehand? Or should per-node state only be created when it is needed?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+  // Get a raw pointer to the input Context for the node. It should have
+  // been created in the constructor.
+  auto ctxIt = inputCtxs_.find(node);
+
+  if (ctxIt == inputCtxs_.end()) {
+    assert(false && "Input context not found but should exist!");
+  }
+
+  Context *ctx = (ctxIt->second).get();
+  assert(ctx);
```
gcatron: I'd add a comment here explaining the assert and it's motivation. These asserts *should* never happen so I'm good with them. If they do trip that means the entire host is invalidated.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+  assert(resultCtx_);
+  return std::move(resultCtx_);
+}
+
+void ThreadPoolExecutor::run(const DAGNode *root,
+                             std::unique_ptr<Context> context,
+                             RunIdentifierTy runId, DoneCbTy cb) {
+  // If list of roots is empty, there is nothing to do.
+  if (!root) {
+    cb(runId, ResultCode::Executed, nullptr);
```
gcatron: We should probably still move the context.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+    return;
+  }
+
+  // If the given run ID corresponds to a run already in progress, there is
+  // also nothing to do, but return an error.
+  std::unique_lock<std::mutex> lock(executionStatesMutex_);
+  if (executionStates_.find(runId) != executionStates_.end()) {
+    cb(runId, ResultCode::Failed, nullptr);
```
gcatron: Same for moving the context back.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+  auto executionState =
+      std::make_shared<ExecutionState>(runId, root, std::move(cb));
+  executionStates_.insert(std::make_pair(runId, executionState));
+  lock.unlock();
+
+  // Execute all child nodes of root.
+
+  // Mark the child nodes as "inflight" (i.e. currently executing). This must be
+  // done here instead of inside executeDAGNode() so that a node can be
+  // executed which placeholders are being propagated for the next node without
```
gcatron: which -> while ?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+                                                   const DAGNode *node) {
+  // Get the execution state for the run.
+  std::shared_ptr<ExecutionState> executionState = executionStates_[runId];
+
+  if (resultCode != ResultCode::Executed) {
+    // If the DeviceManager failed to execute the node, call the callback
+    // provided when ThreadPoolExecutor::run() was called, clean up the
+    // lock and execution state for the run and exit.
+    DoneCbTy cb = executionState->getCallback();
+    cb(runId, resultCode, nullptr);
```
gcatron: Pass the context back to the caller.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: But there is no `Context` to move? :S

Unless we want to give back the `Context` we were given as input? But that doesn't make sense to me.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: There is no "complete: `Context` to pass back here. The result `Context` being prepared in `ExecutionState` will not have all of the expected `Placeholders` in it.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Same here, there is no `Context` to hand back?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,339 @@
...
+#include <queue>
+#include <unordered_set>
+
+namespace glow {
+namespace runtime {
+
+ExecutionState::ExecutionState(RunIdentifierTy id, const DAGNode *root,
+                               DoneCbTy doneCb)
+    : cb_(doneCb), resultCtx_(std::make_unique<Context>()), inflightNodes_(0),
+      resultCode_(ResultCode::Executed) {
```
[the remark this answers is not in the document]
SplitInfinity: Yeah, this is set up before a run. My idea was to assume that execution is going well until we find out otherwise (when a thread sets it to `Failed`).

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+  if (ctxIt == inputCtxs_.end()) {
+    assert(false && "Input context not found but should exist!");
+  }
+
+  Context *ctx = (ctxIt->second).get();
+  assert(ctx && "Input context for node is null");
+
+  // Insert the placeholder-tensor pair.
+  std::lock_guard<std::mutex> lock(contextMtx_);
```
nickgg: Only one thread can write to any context at a time? Is there performance to be gained by having a lock per Context?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+std::unique_ptr<Context>
+ExecutionState::getUniqueNodeContextPtr(const DAGNode *node) {
+  auto ctxIt = inputCtxs_.find(node);
+
+  if (ctxIt == inputCtxs_.end()) {
+    assert(false && "Input context not found but should exist!");
+  }
+
+  return std::move(ctxIt->second);
```
nickgg: Whoa! This modifies the `inputCtxs_` map without the lock, this WILL crash `insertIntoNodeCtx` if called concurrently

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: Actually this feels dangerous, you look up an entry the map, get a non-owning pointer then try to lock the map. Even if the map is mutated under the lock this isn't safe because the Context* is pointing to an object that may be removed from the map before you get the lock.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+void ExecutionState::insertIntoResultCtx(Placeholder *P, Tensor &&T) {
+  assert(resultCtx_ && "Execution result context should exist!");
+  std::lock_guard<std::mutex> lock(contextMtx_);
+  resultCtx_->insert(P, std::move(T));
+}
+
+std::unique_ptr<Context> ExecutionState::getUniqueResultContextPtr() {
+  assert(resultCtx_ && "Execution result context should exist!");
+  return std::move(resultCtx_);
```
nickgg: You need to take the context lock here, because `insertIntoResultCtx` is using that to determine if it is safe to insert into it.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+void ExecutionState::setResultCode(const ResultCode resultCode) {
+  if (resultCode_ != ResultCode::Failed) {
+    resultCode_ = resultCode;
+  }
```
nickgg: Should you handle the `Failed` case here?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+}
+
+void ThreadPoolExecutor::run(const DAGNode *root,
+                             std::unique_ptr<Context> context,
+                             RunIdentifierTy runId, DoneCbTy cb) {
+  // If list of roots is empty, there is nothing to do. Give back the
+  // context so the caller can reuse it.
+  if (!root) {
+    cb(runId, ResultCode::Executed, std::move(context));
```
nickgg: This is kind of philosophical, but is an empty tree `Executed`? If a tree is executed in the forest and there's no one around to hear it does it make a sound? Or am I misreading and this is the termination case of the loop?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+  if (executionStates_.find(runId) != executionStates_.end()) {
+    cb(runId, ResultCode::Failed, std::move(context));
+    return;
+  }
+
+  // Create execution state tracker object for this run ID.
+  auto executionState =
+      std::make_shared<ExecutionState>(runId, root, std::move(cb));
+  executionStates_.insert(std::make_pair(runId, executionState));
+  lock.unlock();
```
nickgg: I really think it's safer to unlock via scope not a manual unlock() call, super easy to refactor this into a deadlock.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+    assert("Execution state should exist!");
+  }
+
+  std::shared_ptr<ExecutionState> executionState = it->second;
+
+  // Get the map of Placeholder -> Tensor mappings within ctx.
+  const Context::PlaceholderMap &phMap = ctx->pairs();
+
+  // Get the symbol table for the node.
+  const RuntimeBundle &bundle = node->runtimeBundle;
```
nickgg: We don't appear to use this variable, do we need it?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+    if (symbolInfo.input) {
+      // If the symbol is an input, look for a Placeholder in ctx with
+      // the same name and copy the corresponding Tensor into the input Context
+      // being prepared for the node.
+      for (const auto &phTensorPair : phMap) {
+        auto *placeholder = phTensorPair.first;
+        const auto *tensor = phTensorPair.second;
+
+        if (symbolName == placeholder->getName()) {
```
nickgg: Rather than this N*M comparison loop, why don't we add name->PH as state in the context and just do one lookup?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+          executionState->insertIntoNodeCtx(node, placeholder, tensor->clone());
```
nickgg: When is a clone necessary here, aren't we just moving outputs of one node to inputs of the next?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+  // Execute all child nodes of root.
+
+  // Mark the child nodes as "inflight" (i.e. currently executing). This must be
+  // done here instead of inside executeDAGNode() so that a node can be
+  // executed while placeholders are being propagated for the next node without
+  // the callback for that node deleting the execution state.
+  executionState->incrementInflightNodes((root->children).size());
+
+  // Convert the Context to shared_ptr.
+  std::shared_ptr<Context> sharedContext = std::move(context);
```
nickgg: Looks like we don't store sharedContext anywhere, are you just doing this to pass them into multiple functions? Feels safe to use a raw pointer here.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+    // Mark the node as no longer execution.
+    executionState->decrementInflightNodes();
+    return;
+  }
+
+  // Get the DeviceManager that can run the node.
+  auto deviceManagerIt = deviceManagers_.find(node->deviceID);
+
+  if (deviceManagerIt == deviceManagers_.end()) {
+    // Mark the node as no longer executing.
```
nickgg: Isn't this always a bug in our code? Could we log or maybe assert this here?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: Didn't they pass some context into us though? The one in `run()` - can we pass it back?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+  std::unique_ptr<Context> nodeCtx =
+      executionState->getUniqueNodeContextPtr(node);
+  // Run the node using the DeviceManager.
+  deviceManager->runFunction(
+      node->name, std::move(nodeCtx),
+      [this, runId, node](RunIdentifierTy id, ResultCode resultCode,
+                          std::unique_ptr<Context> resultCtx) {
+        // Immediately move the handling of the result onto threadPool_ to
+        // avoid doing work on the DeviceManager thread.
+        std::shared_ptr<Context> sharedResultCtx = std::move(resultCtx);
```
nickgg: Why do you convert to shared_ptr here? If you want to use shared_ptr instead of unique_ptr we should talk about making that the interface to the DeviceManager.

An advantage of using unique_ptr is that you can read/write the Context safely without locks because you know you don't have any other pointers to it. With shared_ptr I'm pretty suspicious of accesses to Context::pairs() for example.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  std::shared_ptr<ExecutionState> executionState = it->second;
+
+  // Copy all of the Placeholders in ctx into the result Context for the run.
+  for (const auto &phTensorPair : ctx->pairs()) {
+    auto *placeholder = phTensorPair.first;
+    const auto *tensor = phTensorPair.second;
+
+    executionState->insertIntoResultCtx(placeholder, tensor->clone());
```
nickgg: Do we have to copy here? Can we move?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+      // the callback if the entire execution succeeded and erase the state
+      // information. If some node failed, the callback was called in the
+      // invocation of handleDeviceManagerResult() for that node, so the
+      // callback should be skipped here.
+      if (executionState->getResultCode() == ResultCode::Executed) {
+        DoneCbTy cb = executionState->getCallback();
+        cb(runId, executionState->getResultCode(),
+           executionState->getUniqueResultContextPtr());
+      }
+      std::lock_guard<std::mutex> lock(executionStatesMutex_);
```
nickgg: It's weird that we grab the lock down here only? Don't we modify the executionState above? What if two nodes finish at the same time and both start modifying execution state?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -0,0 +1,377 @@
...
+      }
+    }
+  }
+}
+
+void ExecutionState::insertIntoNodeCtx(const DAGNode *node, Placeholder *P,
+                                       Tensor &&T) {
+  // Get a raw pointer to the input Context for the node. It should have
+  // been created in the constructor.
+  auto ctxIt = inputCtxs_.find(node);
```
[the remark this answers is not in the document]
SplitInfinity: After discussing with @gcatron, I believe it's because the map is set up in `ExecutionState()` and then new key-value pairs are never added after that. I'll add some comments.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Yes, only one thread can write to any `Context` at a time, since writing to it might involve inserting new key-value pairs into the underlying map inside the `Context`. This can happen if a node has multiple parent nodes that feed their outputs to it.

SplitInfinity: For the reasons mentioned above, it should not be possible for this to run after `getUniqueNodeContextPtr`.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: These two functions can't be concurrent due to the reasons mentioned above, but these are famous last works. I will still lock here so that I can add an assert before the `Context` is dereferenced in `insertNodeIntoCtx` to make that situation easier to debug if it happens.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I intended for the `Failed` case to be "sticky" so that even one failed node would make the whole DAG as failed. So if `resultCode_` is already `ResultCode::Failed`, it cannot become anything else.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I mean, `nullptr` is a valid linked list and binary tree 😝 

I'm fine with returning `ResultCode::Failed` here though.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I agree, but I did this because I don't think we should hold on to the lock as we call `propagatePlaceholdersForNode` and `executeDAGNode` below. Once the `runId` has been recorded in `executionStates_`, we don't need that lock anymore.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: We use it in the next line to get a reference to the symbol table, but I can combine them if it doesn't hurt readability.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Will do.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: It's necessary if you need to propagate an output from one node to multiple children.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: So this is just a downstream effect of having to convert the `unique_ptr<Context>` returned by the `DeviceManager` to a `shared_ptr<Context>` (see line 282) in order to move the handling of the result onto the `ThreadPoolExecutor` thread pool. Apparently you can't capture `unique_ptr` in a lambda?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I'll put in an assert for now.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I did this because I wasn't able to capture the `unique_ptr` in the lambda passed to `threadPool_` below :(. I read something online about how anything you want to capture in a lambda has to be moveable AND copyable, but this might have changed for newer C++ versions?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Yep, unlike the other case, I think we can! Unless we want to allow tapping off intermediate results as outputs?

include/glow/Backends/BackendUtils.h
```diff
@@ -29,13 +29,19 @@ struct RuntimeSymbolInfo {
   size_t offset;
   /// Type of symbol.
   Type type;
+  /// Is the symbol an input for the function.
+  bool input;
+  /// Is the symbol an output for the function.
+  bool output;
 };
 
+using SymbolTable = std::unordered_map<std::string, RuntimeSymbolInfo>;
```
nadavrot: Do we need to update or read the symbol table at inference time? Updating and accessing the symbol table can be expensive (malloc for update, hash for lookup).

tests/unittests/ExecutorTest.cpp
```diff
@@ -0,0 +1,527 @@
...
+#include <thread>
+
+using namespace glow;
+using namespace glow::runtime;
+
+namespace {
+/// Compare Contexts \p a and \p b. \returns true if
+/// they contain the same Placeholders mapped to
+/// equivalent Tensors.
+bool contextsMatch(const Context *a, const Context *b) {
```
nadavrot: We can add the clone and compare methods to Context.

tests/unittests/ExecutorTest.cpp
```diff
@@ -0,0 +1,527 @@
...
+  E_->run(root, std::move(inputContext), testRunId,
+          [&resultRunId, &resultCode,
+           &resultContext](RunIdentifierTy runId, ResultCode code,
+                           std::unique_ptr<Context> context) {
+            resultRunId = runId;
+            resultCode = code;
+            resultContext = std::move(context);
+          });
+
+  std::this_thread::sleep_for(std::chrono::seconds(2));
```
nadavrot: does this mean that our unittests sleep for 2 seconds?   Every unit test? I don't think that this scales with the number of tests. What happens when we get to 100 tests? We will wait 200 seconds?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: oh you need to do a move alias in the lambda capture, something like this:
```
[ctx = std::move(resultCtx), ...]
```

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: What I'm suggesting is that you add another scope block when you grab the lock, and close it here where you would unlock - letting the lock going out of scope unlock it for you. Doesn't look like you create any variables that are referenced outside of this zone.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: Sure, I guess as an optimization in the future we can track how many children there are are not do the copy on the last one. That way the common case where there is only one child has no allocs.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: You can, see above (or check out the DeviceManager implementations where we do it).

nickgg:
I'm a little concerned that implicit rules about when functions can and can't be called can be fragile long term. If you want to go that way I think we should be extremely explicit in the comments.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: That cannot happen - `noNodesInFlight` can only be true for one thread.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I misunderstood - turns out that the `Context` given in `run` contains both input and output `Placeholders`. So instead of creating a new result `Context` in the `ExecutionState` class, I should move this one into it and return it whenever the callback is called.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Oh, that's a cool trick!

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Also, `Context` doesn't have tracking for the number of node parents that are done, or these little intermediate input `Contexts`.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I think there might be some performance to be gained but I cannot say for certain without measurement. I'll keep it mind for when do start benchmarking and looking for optimization opportunities.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: I added a comment about this on top of the function declaration.

bertmaher:
Have you tried running this code under TSAN?

SplitInfinity:
@bertmaher 

Just tried, apparently there is a data race in the single threaded, single node test. I am a person of many talents, you know...

PR #2320, opened 2019-01-30
Fix warnings with gcc 4.8.5

bertmaher:
*Description*: These CPU backend warnings have been there for a while, time to squash them.
*Testing*: ninja all
*Documentation*: n/a


nadavrot:
It's good that we have an archaeologist on the team. @bertmaher  😺

bertmaher:
I'm pretty sure a team of mad scientists extracted Cretaceous mosquito DNA from a fossil and used it to create CentOS 7 :-p.

Issue #2211, opened 2018-12-21
Customized Node to Instruction Convertion


champyen:
Currently backend implementation has to provide one-to-one target-specific Node and Inst. 

And the conversion from Node to Inst is fully under glow's control. Is it possible to design & create an interface for backend to convert specific Node to Inst? 

It will be helpful for DLA which will be better or need to get some Inst sequences for some Node of operations.

champyen:
Yes, I'm interested on this feature. 

Recently, I'm thinking about another issue of Node/Graph manipulation. (Currently backend can only replace glow node with specific node by calling replaceAllUsesOfWith). 

For most NPUs/DLAs hardware engines have restrictions for some operators, it would be better to be able to split one node to multiple nodes. (eg: Hardware has sram size issue for convolution operator, it would required to partition one convolution to several convolutions with offset) Of course, it could be resolved by "Customized Node to Instruction Convertion". But it may be more intuitive and simpler to do this in Graph level. (Because for Low-level IR and target binary generation, the action is one-to-one translation)

I will try to write another issue when the idea is clear.

jfix71:
Ah ok, yeah I think that would make more sense at the Graph level. It seems that what you need to do is already supported via `replaceAllUsesOfWith()` -- you can use it to replace one Node with multiple Nodes.

For example, backends can decide whether to lower the `FullyConnectedNode` into a `MatMulNode` followed by a `BatchedAddNode`. When we do this with `replaceAllUsesOfWith()`, we replace the `FullyConnectedNode` by the `BatchedAddNode` ([see here](https://github.com/pytorch/glow/blob/master/lib/Optimizer/Lower.cpp#L99)). 

We also lower group Convolutions [here](https://github.com/pytorch/glow/blob/master/lib/Optimizer/Lower.cpp#L600), where we replace a `ConvolutionNode` with `group > 1`, with multiple `ConvolutionNodes` with `group = 1`, which are all concatenated together with a `ConcatNode`. The original `ConvolutionNode` with `group > 1` is replaced by this `ConcatNode` using `replaceAllUsesOfWith()`.

[...]

PR #2322, opened 2019-01-30
[Improvement] Custom translation from Node IR

ayermolo:
*Description*:  Extending framework to support custom translation from Node IR to Instruction IR. Right now there is auto generated lowering for custom nodes to custom instruction IR. This allows more flexibility. For example regularly supported Node that can be lowered to something custom BE adds for Instruction IR

Refactored `IRGenVisitor` to a header so that it can be used by a backend to generate custom Instructions during IRGen. Otherwise no real change to IRGen or the `IRGenVisitor`. A couple helper functions were added too. And then there's an extra call to `Backend::generateInst()` at the beginning of `post()` to check if the backend wants to IRGen the current node, otherwise our default IRGen logic takes place.

*Testing*: Unit tests
*Documentation*:
fixes https://github.com/pytorch/glow/issues/2211

### DIFF
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -27,6 +27,7 @@ namespace glow {
 class IRFunction;
 class Node;
 class Context;
+class IRGenVisitor;
 
 enum class BackendKind {
   Interpreter, // Execute the network with the built-in interpreter.
@@ -85,6 +86,12 @@ public:
 
   /// Optimize the Function \p F given compilation mode \p mode.
   void optimizeFunction(CompilationMode mode, Function *F);
+
+  /// \returns true if Backend generated Instruction for Node \p N,
+  /// using IRGenVisitor \p irgen.
+  virtual bool generateInst(Node *N, IRGenVisitor &irgen) const {
+    return false;
+  }
 };
 
 /// Create a backend of kind \p kind.
--- a/include/glow/IR/IR.h
+++ b/include/glow/IR/IR.h
@@ -232,6 +232,7 @@ private:
   void createNode(const Instruction &);
 };
 
+class Backend;
 class WeightVar;
 class Value;
 class Node;
@@ -276,7 +277,8 @@ public:
 
   /// Generate IR from the graph nodes. If the compilation mode is 'training'
   /// then this procedure will also generate the code for the backward pass.
-  void generateIR();
+  /// It allows Backend /p B to custom translate from a Node to Instruction IR.
+  void generateIR(const Backend &B);
 
   /// Wipe out the content of the function. This allows the function to be used
   /// again for another round of code generation.
--- a/include/glow/IR/IRGen.h
+++ b/include/glow/IR/IRGen.h
@@ -1,0 +1,82 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_IR_IRGEN_H
+#define GLOW_IR_IRGEN_H
+
+#include "glow/Backends/Backend.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Nodes.h"
+#include "glow/IR/IR.h"
+#include "glow/IR/IRBuilder.h"
+
+#include <unordered_set>
+
+//===----------------------------------------------------------------------===//
+//              IRGen visitor - the code that generates the IR.
+//===----------------------------------------------------------------------===//
+
+namespace glow {
+
+/// This class implements \p NodeWalker interface to translate from Node to
+/// Instruction IR.
+class IRGenVisitor : public NodeWalker {
+private:
+  using NodeValueToDestTy = std::unordered_map<NodeValue, Value *>;
+  using NodeToInstrTy = std::unordered_map<Node *, Instruction *>;
+
+  /// A set of visited nodes.
+  std::unordered_set<Node *> visited_;
+  /// Holds the mapping between graph nodes to the destination buffers.
+  NodeValueToDestTy generatedNodeDest_;
+  /// Holds the mapping between graph nodes and the lowered instructions. This
+  /// map is used by instructions that want to inspect the generated
+  /// instructions. For example, gradient instructions that look at operands
+  /// that do not exist at the graph level. Not all variables are representible.
+  NodeToInstrTy nodeToInstr_;
+
+  /// The function that we are building.
+  IRFunction *F_;
+  /// The builder that adds instructions into the function.
+  IRBuilder builder_;
+  /// The Backend /p B is used for custom lowering of Node to Instruction IR.
+  const Backend &B_;
+
+public:
+  bool shouldVisit(Node *parent, Node *N) override;
+
+  explicit IRGenVisitor(IRFunction *M, const Backend &B)
+      : F_(M), builder_(F_), B_(B) {}
+
+  /// \returns the generated instruction for the node \p N.
+  Value *valueForNode(NodeValue N);
+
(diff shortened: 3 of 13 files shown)


include/glow/IR/IRGen.h
```diff
@@ -0,0 +1,58 @@
+#ifndef GLOW_IR_IRGEN_H
```
jfix71: Please add the copyright header as seen in other `.h` files

include/glow/Optimizer/Optimizer.h
```diff
@@ -58,7 +58,7 @@ Function *profileQuantization(Context &ctx, Function *F,
 
 /// Helper to generate and optimize IR from given Function \p F. \p
 /// shouldShareBuffers signifies whether to use the share buffers optimization.
```
jfix71: Please update the doxygen here with `\p B`

include/glow/Backends/Backend.h
```diff
@@ -85,6 +86,8 @@ class Backend {
 
   /// Optimize the Function \p F given compilation mode \p mode.
   void optimizeFunction(CompilationMode mode, Function *F);
+  /// \returns true if Backend generated Instruction
```
jfix71: Please name the parameters and mention them here in the doxygen. Also end sentences with a period 😃.

include/glow/IR/IR.h
```diff
@@ -276,7 +277,7 @@ class IRFunction final : public Named {
 
   /// Generate IR from the graph nodes. If the compilation mode is 'training'
   /// then this procedure will also generate the code for the backward pass.
```
jfix71: Please update doxygen comments to mention `\p B`

lib/IR/IRGen.cpp
```diff
@@ -14,23 +14,17 @@
...
 #include <unordered_set>
 
-using namespace glow;
-
 //===----------------------------------------------------------------------===//
 //              IRGen visitor - the code that generates the IR.
 //===----------------------------------------------------------------------===//
 
-namespace {
+namespace glow {
```
jfix71: We prefer to use `using namespace glow;` in `cpp` files.

lib/IR/IRGen.cpp
```diff
@@ -49,365 +43,345 @@ static size_t getConsecutiveSameNodeCount(NodeValueArrayRef inputs,
...
+void IRGenVisitor::post(Node *parent, Node *N) {
+  visited_.insert(N);
 
-    switch (N->getKind()) {
-    default:
-      llvm_unreachable("Unhandled node; perhaps the node should have been "
-                       "lowered, or the backend should have specified an IRGen "
-                       "case for this node to a backend-specific Instr.");
-      break;
+  // allows backend to generate their custom instrution IR
```
jfix71: nit: Capitalize "allows" and end sentence with a period.

nadavrot:
@ayermolo Thanks for doing this work. I have a few comments and questions:

It is very difficult for me to review IRGen.cpp because so much code in the file changes.

Could you give an example where custom IR translation is useful?  

Could you explain why the approach taken (IR visitor) is the right way to implement this functionality and not something else?

ayermolo:
@nadavrot 
IRGen. Ah didn't notice. Hmm, looks like it's white spaces related. Locally beyondcompare ignores it. I'll fix on next commmit.

One usage model I envision where we have a common Node on Node IR level, and can take advantage of all the optimizations that apply to such nodes, but we want different Instruction IR then what is generated through common pass. One example is Concat. Right now it's lowered to bunch of insertTensors. Works for CPU, but might not for other custom back ends.

For this approach. One thing I tried to do is reduce the scope of changes. There was some refactoring or IRGen, but otherwise changes are minimal. Also in some sense it's similar to transform{Pre/Post}Lowering hooks in to backend. There is a common pass that goes through IRGEN, but now this allows Backend to intercept node being lowered in to instruction IR and do it's own thing.

I am certainly open to other approaches.

jfix71:
I think the proposed approach here is reasonable and not too intrusive, and as @ayermolo said it's somewhat in line with approaches we've used elsewhere in the codebase. We are already using the IRGenVisitor here.

I could envision custom IRGen being useful -- I believe @opti-mix had previously mentioned desiring this support.

include/glow/IR/IRGen.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <unordered_set>
+
+//===----------------------------------------------------------------------===//
+//              IRGen visitor - the code that generates the IR.
+//===----------------------------------------------------------------------===//
+
+namespace glow {
+
+/// A helper class for visiting and generating the dotty file from the graph.
```
jfix71: This comment is wrong -- I see it was copied from IRGen.cpp, but can you fix it while you're moving the class too?

ayermolo:
@nadavrot @jfix71 Thanks for input. I'll rework it and update this PR.

include/glow/IR/IR.h
```diff
@@ -276,7 +277,8 @@ class IRFunction final : public Named {
...
-  void generateIR();
+  /// It allows Backend /p B to custom transalte from a Node to Instruction IR.
```
jfix71: nit: typo `transalte`

include/glow/IR/IRGen.h
```diff
@@ -0,0 +1,81 @@
...
+  /// map is used by instructions that want to inspect the generated
+  /// instructions. For example, gradient instructions that look at operands
+  /// that do not exist at the graph level. Not all variables are representible.
+  NodeToInstrTy nodeToInstr_;
+
+  /// The function that we are building.
+  IRFunction *F_;
+  /// The builder that adds instructions into the function.
+  IRBuilder builder_;
+  const Backend &B_;
```
jfix71: Please add a doxygen comment for `B_`

tests/unittests/BackendTestUtils.h
```diff
@@ -38,6 +45,61 @@ class MockBackend : public Backend {
...
   }
+
+  bool generateInst(Node *N, IRGenVisitor &irgen) const override {
+    return false;
+  }
+};
+
+/// MockBackend2 used only for unit testing to test custom lowering from Node to
+/// Instruction IR.
+class MockBackend2 : public Backend {
```
jfix71: nit: Can you name `MockBackend2` a little more informatively? E.g. `MockBackendCustomIRGen`

include/glow/IR/IRGen.h
```diff
@@ -0,0 +1,82 @@
...
+
+  /// Saves the generated IR in \p v for the node \p N.
+  void registerIR(NodeValue N, Value *v);
+
+  /// Adds to Node \p N --> Instruction \p inst map.
+  void setNodeToIR(Node *N, Instruction *inst);
+
+  /// Return Instruction that is mapped to Node \p N.
+  /// If mapping doesn't exists returns nullptr.
+  Instruction *getNodeToIR(Node *N);
```
jfix71: This doesn't appear to be used anywhere.

opti-mix:
> My understanding is that it's mostly just refactoring `IRGenVisitor` to a header so that it can be used by a backend to generate custom Instructions during IRGen. This resulted in a whitespace change in `post()`, since the body of `post()` now has one less level of indentation.

> Otherwise I think it was no real change to IRGen or the `IRGenVisitor`. A couple helper functions were added too. And then there's an extra call to `Backend::generateInst()` at the beginning of `post()` to check if the backend wants to IRGen the current node, otherwise our default IRGen logic takes place.

This is also my understanding. I think it could be useful if the commit message and the PR description would clearly state, which part of the changes was just a re-factoring (e.g. moving `IRGenVisitor` to a header) and which part was the new functionality (e.g. helper functions and an extra call to `Backend::generateInst()`). This way it is easy for a reviewer to understand what's going on.

nadavrot:
@ayermolo @jfix71 @opti-mix Okay. Sounds good. I don't have any other comments. Thank you.

include/glow/IR/IRGen.h
ayermolo: Well I figured since I added the setter function, add getter too in case anyone needs to use it. In IRGen.cpp map is used directly : nodeToInstr_[N] = V; or auto *PI = cast<MaxPoolWithXYInst>(nodeToInstr_[poolOut.getNode()]);.

I can remove it, and if someone else needs it they can add it.

ayermolo:
@jfix71 @opti-mix Any other issues I need to resolve?
Want me to remove that un-used getter Jordan mentioned?

opti-mix:
I'm OK with merging it. 

@jfix71 Do we want the changes to be squashed into a single commit before merging?

@nadavrot Could you formally approve the PR, because otherwise GH would not allow to merge it.

ayermolo:
Great, thank  you very much.

PR #2325, opened 2019-01-30
[DeviceManagers] If there is no device manager for the new backend, just fail.

rdzhabarov:
Was working on some other backend and figured this was failing to compile.

### DIFF
--- a/lib/Backends/DeviceManagers.cpp
+++ b/lib/Backends/DeviceManagers.cpp
@@ -63,6 +63,8 @@ DeviceManager *DeviceManager::createDeviceManager(BackendKind backendKind,
     return createOCLDeviceManager(name);
   case BackendKind::CPU:
     return createCPUDeviceManager(name);
+  default:
+    GLOW_UNREACHABLE("not supported backend");
   }
 
   // This is to make compiler happy. It can never reach this point as switch


lib/Backends/DeviceManagers.cpp
```diff
@@ -63,6 +63,8 @@ DeviceManager *DeviceManager::createDeviceManager(BackendKind backendKind,
     return createOCLDeviceManager(name);
   case BackendKind::CPU:
     return createCPUDeviceManager(name);
+  default:
+    GLOW_ASSERT("not supported backend");
```
bertmaher: should be `GLOW_UNREACHABLE`

nadavrot:
@rdzhabarov @nickgg @bertmaher 

When adding a new backend we now need to update two places (device manager and the other place that registers the backend). Can we make createDeviceManager() a method of Backend? This will allow us to remove this method and prevent merge conflicts in this file?

rdzhabarov:
> This will allow us to remove this method and prevent merge conflicts in this file?

I can see potential upside here, but both Backends.cpp and DeviceManagers.cpp need to be updated once (and not touched after at all) which should not be too much hassle wrt conflicts and complexity of updates.

> Can we make createDeviceManager() a method of Backend?

This seems reasonable. @nickgg any concerns you had with this approach in the original PR?

In general, way we create those objects:
I think ideal scenario is to have a static registration factory for `backends` this way we'll remove switch logic all together for `backends`.

PR #2326, opened 2019-01-31
Always DCE after lowering

jackm321:
*Description*:
DCE should be performed after lowering like in other graph optimizations that can produce dead code. This may incur an extra call to DCE here or there but this should be relatively low overhead and the trade off is that the various places around the compiler that call lower() do not need to worry about manually doing DCE themselves.

*Testing*:
`ninja all`

*Documentation*:
doxygen

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -56,7 +56,6 @@ static bool shouldDeleteNode(Node *N) {
   return true;
 }
 
-/// Dead code elimination.
 void glow::DCE(Function *F) {
   auto &nodes = F->getNodes();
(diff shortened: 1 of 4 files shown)


jackm321:
See https://github.com/pytorch/glow/pull/2311#discussion_r252475706 for earlier discussion and a use case for this.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -57,7 +57,7 @@ static bool shouldDeleteNode(Node *N) {
 }
 
 /// Dead code elimination.
```
jfix71: nit: Don't need this anymore, doxygen is in the header.

PR #2328, opened 2019-01-31
[RFC] Profiling/Tracing prototype support in the Interpreter 

nickgg:
*Description*: @jackm321 and I have been chatting about what Profiling/Tracing might look like in glow, so I've hacked up machinery and support for it in the in the Interpreter backend. The new example will write a JSON file in the Google Chrome Trace Events format, which can be dropped into their profile UI to give us a visual representation of the time cost of each IR node.

My thinking about how this fits into the Runtime is that the Executor would keep one TraceLogger for the run, and create a new TraceThread for each DAGNode, which would let us group execution by accelerator.

This is just a proposal, very interested in comments on design choices, how this might suit different backends, etc.
*Testing*: unit tests under ASAN & release as normal.
*Documentation*:
Attached is the output of one run of resnet50 in the Interpreter, you can drop it into `chrome://tracing` and have a look around. (you may have to rename the file back to `.json`first).

[resnet-tracing.txt](https://github.com/pytorch/glow/files/2819179/resnet-tracing.txt)

And it should look like this: 

resnet-tracing

### DIFF
--- a/examples/resnet-tracing.cpp
+++ b/examples/resnet-tracing.cpp
@@ -1,0 +1,157 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "glow/Runtime/RuntimeTypes.h"
+#include "glow/Runtime/TraceLogger.h"
+
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+
+#include <chrono>
+#include <future>
+
+using namespace glow;
+using namespace glow::runtime;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-tracing Options");
+llvm::cl::opt<std::string>
+    inputImage(llvm::cl::desc("path to input image to classify, which must be "
+                              "a png with standard imagenet normalization"),
+               llvm::cl::init("../tests/images/imagenet/dog_207.png"),
+               llvm::cl::Positional, llvm::cl::cat(category));
+llvm::cl::opt<std::string>
+    outputJson(llvm::cl::desc("path to write output json trace events"),
+               llvm::cl::init("./resnet-tracing.json"), llvm::cl::Positional,
+               llvm::cl::cat(category));
+} // namespace
+
+/// Loads the model into /p module and returns the input and output
+/// Placeholders.
+std::pair<Placeholder *, Placeholder *> loadResnet50Model(TypeRef inputType,
+                                                          Module &module) {
+  Function *F = module.createFunction("resnet50");
+
+  llvm::outs() << "Loading resnet50 model.\n";
+
+  const char inputName[] = "gpu_0/data";
+  Caffe2ModelLoader loader("resnet50/predict_net.pb", "resnet50/init_net.pb",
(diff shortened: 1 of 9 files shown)


include/glow/Graph/Context.h
```diff
@@ -42,6 +45,9 @@ class Context final {
   /// Maps Placeholders to Tensors.
   PlaceholderMap map_;
 
+  /// Stores trace events for this run. May be nullptr.
```
nadavrot: Why is the tracer a part of the context? What is the motivation here?

nadavrot:
It's easy to add tracing to the interpreter because we can just write into some data structures. I suggest starting with OpenCL or CPU because this would allow us to develop more realistic interfaces. What do you say?

opti-mix:
> It's easy to add tracing to the interpreter because we can just write into some data structures. I suggest starting with OpenCL or CPU because this would allow us to develop more realistic interfaces. What do you say?

Does it make sense to introduce something similar to our existing `DebugPrintInst` and perform the instrumentation, so that each backend can either execute such an instruction directly or generate code to collect some profiling/tracing data?

include/glow/Graph/Context.h
nickgg: The Context is the per-run part of the state, and traces are also per run. Where else would you put it?

nickgg:
I picked Interpreter because it was the easiest. For any BackendKind will need to collate events in the DeviceManager and return them to the HostManager, the logical place to do this seems to be the Context.

nadavrot:
@nickgg Can we use an opaque tensor for writing the profiling info into? I envision something like a placeholder with a constant name ("profile_data"), where special instructions (as suggested by @opti-mix) write to specific offsets.

something like:

\@write_timestamp \@inout %profile_data, offset 34124 
\@convolution ..... 
\@write_timestamp \@inout %profile_data, offset 34125

nadavrot:
> We definitely could, and I imagine backends which we compile in the synchronization points would have to do something similar to that.
> 
> On some architectures we may get performance counters through a side channel, how would we integrate them into such a scheme?

Ah, good point. I don't know. 

> 
> How do we insert Events that are generated in the Runtime (such as queue sub-partition, etc) not inside a particular backend?

Yes, I agree. This is also a problem.  :( :(

nickgg:
So what I'd like is to agree on an intermediate format inside the Runtime, which we can push device specific events into during run in the DeviceManager, and aggregate in the Executor. Per device we may have different options, or different restrictions, but I think the best way to handle this is to try at the DeviceManager layer to translate it into some unified format.

examples/resnet-tracing.cpp
```diff
@@ -0,0 +1,157 @@
+/**
```
bertmaher: Can you just fold a tracing option into image-classifier instead of adding another resnet sample?

include/glow/Runtime/TraceLogger.h
```diff
@@ -0,0 +1,112 @@
...
+struct TraceEvent {
+  /// Human readable name for the item, will be used to match up begin and end.
+  std::string name;
+
+  /// Time of the event, in milliseconds since epoch.
+  uint64_t timestamp;
+
+  /// Type of the event, a one char code (see Event Descriptions in the Trace
+  /// Event Format spec). e.g. 'B' for begin event, 'E' for end event.
+  char type;
```
jackm321: Use `enum class EventType : char`

include/glow/Runtime/TraceLogger.h
```diff
@@ -0,0 +1,112 @@
...
+
+  /// \returns a new TraceThread for the given tid. That TraceThread does not
+  /// share state with the TraceLogger. Ia \p tid is not specified, will use the
+  /// next sequential Thread Id.
+  TraceThread getTraceThread(int tid = -1);
+
+  /// Collects TraceEvents from the provided \p traceThread and moves them into
+  /// the TraceLogger. The TraceThread can still be used, however it's event
+  /// vector is cleared.
+  void returnTraceThread(TraceThread &traceThread);
```
jackm321: Why is it necessary to return a traceThread as a ref? Maybe a different name would be better?

include/glow/Runtime/TraceLogger.h
```diff
@@ -0,0 +1,112 @@
...
+
+/// Aggregator for a single run's TraceEvents, i.e. for a single inference.
+// The usage pattern should be something like:
+//     1. getTraceThread() to get a logger per thread of execution.
+//     2. log events on the traceThread()
+//     3. when the execution is finished, returnTraceThread(traceThread) to
+//        aggregate TraceEvents.
+class TraceLogger {
+  /// TraceEvents for this pid. Can be unordered.
+  std::vector<TraceEvent> traceEvents_;
```
jackm321: Should this have a lock? We want this to run concurrently for multiple devices for example.

nickgg:
Closed in favor of #2375.

PR #2327, opened 2019-01-31
[Caffe2ImporterTest] Allow definition of test data path

bertmaher:
*Description*: To make it easier to integrate into build systems with different roots.
*Testing*: ninja Caffe2ImporterTest
*Documentation*: n/a

PR #2329, opened 2019-01-31
[new operators] Support Alias and EnforceFinite as no-ops

artemrakhov:
*Documentation*:
Adds support of `Alias` and `EnforceFinite` operators.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -960,7 +960,13 @@ protected:
       RETURN_IF_ERR(loadIdentity(op, dict));
       return true;
     }
-    if (typeName == "Identity") {
+    if (typeName == "EnforceFinite") {
+      // Load as Identity operation, because in Caffe2 this op does not change
+      // any data. It only raises error if some tensor element is Inf or NaN.
+      RETURN_IF_ERR(loadIdentity(op, dict));
+      return true;
+    }
+    if (typeName == "Identity" || typeName == "Alias") {
       RETURN_IF_ERR(loadIdentity(op, dict));
       return true;
     }
--- a/tests/models/caffe2Models/alias_op_net.pbtxt
+++ b/tests/models/caffe2Models/alias_op_net.pbtxt
@@ -1,0 +1,9 @@
+name: "alias"
+op {
+  input: "X"
+  output: "Y"
+  name: ""
+  type: "Alias"
+}
+external_input: "X"
+external_output: "Y"
--- a/tests/models/caffe2Models/enforcefinite_op_net.pbtxt
+++ b/tests/models/caffe2Models/enforcefinite_op_net.pbtxt
@@ -1,0 +1,9 @@
+name: "EnforceFinite"
+op {
+  input: "X"
+  output: "Y"
+  name: ""
+  type: "EnforceFinite"
+}
+external_input: "X"
+external_output: "Y"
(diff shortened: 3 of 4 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -960,7 +960,13 @@ class CommonOperatorLoader : public ProtobufLoader {
       RETURN_IF_ERR(loadIdentity(op, dict));
       return true;
     }
-    if (typeName == "Identity") {
+    if (typeName == "EnforceFinite") {
```
jackm321: why not just combine this with the case below?

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -960,7 +960,13 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+      // Load as Identity operation, because in Caffe2 this op does not change
+      // any data. It only raises error if some tensor element is Inf or NaN.
+      RETURN_IF_ERR(loadIdentity(op, dict));
+      return true;
+    }
+    if (typeName == "Identity" || "Alias") {
```
jackm321: Does this really work? You don't need `typeName == "Identity" || typeName ==  "Alias"`

jackm321:
Could you add some tests please.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -960,7 +960,13 @@ class CommonOperatorLoader : public ProtobufLoader {
       RETURN_IF_ERR(loadIdentity(op, dict));
       return true;
     }
-    if (typeName == "Identity") {
+    if (typeName == "EnforceFinite") {
+      // Load as Identity operation, because in Caffe2 this op does not change
+      // any data. It only raises error if some tensor element is Inf or NaN.
```
rdzhabarov: > It only raises error if some tensor element is Inf or NaN.

like execution fails? or some data will be logged?

bertmaher:
Can you add GLOW_DATA_PATH as in #2327 ?

include/glow/Importer/CommonOperatorLoader.h
artemrakhov: C++ exception is thrown.
There is also an option to crash with fatal error, but it's disabled by default.
Here is exactly what happens:
https://github.com/pytorch/pytorch/blob/a4475d529d8a80c81fb0c1e0bc0f02f40ac53282/c10/util/Logging.cpp#L36

bertmaher: It doesn't seem correct to emit an Identity op for EnforceFinite, since it drops the "enforce finite" part of the semantics.  I'd rather us support this op correctly or not at all, unless you have a compelling reason for doing it this way.

artemrakhov: OK. Is there a preferred way to raise an error from Backend layer? We've never done this before. We do have `assert`s, but they are only used in Debug for test purposes.

bertmaher: I'm not entirely sure we can...  I mean, obviously we could do something in the interpreter or cpu backend, but ASICs don't even necessarily provide control flow primitives that would allow us to implement it...

artemrakhov: In the past, we supported ops omitting some functionality. For example, there were `EnsureCPUOutput`, `CopyCPUToMKL`, `CopyMKLToCPU` and `Copy` found somewhere in the nets we wanted to run. Not to change the nets themselves, we added these ops as identity. Alternative would be to say that these actions don't make sense in Glow. Therefore won't be supported. And higher level framework (Caffe2) should take care of removing these ops.

But at that time we didn't have any integration, and goal was to load as much of the protobuf as possible (managed to load whole model as result).

artemrakhov: I created PR #2340 to add Alias op separately. I am going to close this PR, since not supporting `EnforceFinite` at all is probably the right thing to do.

PR #2333, opened 2019-02-01
Add support for multiple compiled functions in ExecutionEngine

jackm321:
*Description*:
A temporary home for Function families.
Enable multiple compiled functions in a single EE. This is useful for the ExecutionEngine onnxifi backends and probably for unit testing now.
*Testing*:
I'll write some unit tests for this but just wanted to get feedback on the approach first.

*Documentation*:
doxygen

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -149,16 +149,18 @@ void glow::runBatch(ExecutionEngine &EE, Context &ctx, size_t iterations,
   }
 }
 
-void ExecutionEngine::compile(CompilationMode mode, Function *F) {
-  backend_->optimizeFunction(mode, F);
-  compiledFunctions_.clear();
-  compiledFunctions_[F->getName()] = backend_->compile(F);
-}
-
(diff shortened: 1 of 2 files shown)


rdzhabarov:
Given that it's intermediate step, have you considered to handle this situation on onnxifiGlow side by having multiple EE handling each function?
Eventually we'll migrate to HostManager an Co, but seems like changing onnxifiGlow is less intrusive change to the rest of the compiler.

jackm321:
@rdzhabarov we can't have multiple EEs because we want to have multiple functions in the same module so that they can share constants.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -130,7 +151,15 @@ void glow::runBatch(ExecutionEngine &EE, Context &ctx, size_t iterations,
 
 void ExecutionEngine::compile(CompilationMode mode, Function *F) {
   backend_->optimizeFunction(mode, F);
-  function_ = backend_->compile(F);
+  compiledFunctions_.clear();
```
rdzhabarov: this behavior needs to be in the comments. `compile` essentially discards all previously compiled functions.

PR #2334, opened 2019-02-02
[New operator] Implement SparseToDenseMask

artemrakhov:
*Description*:
It is unfortunate that we now need to create a hash map inside an operator. Luckily, `mask` is known in advance, so we can precompute the hash map. I decided not to do it yet, because the goal of Interpreter backend is simplicity.

*Testing*:
`ninja test`

*Documentation*:
Introduce `SparseToDenseMask` operator, which matches Caffe2 semantics:
https://caffe2.ai/docs/operators-catalogue.html#sparsetodensemask

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1442,6 +1442,22 @@ SparseToDenseNode *Function::createSparseToDense(llvm::StringRef name,
   return addNode(new SparseToDenseNode(name, outTy, indices, values));
 }
 
+SparseToDenseMaskNode *Function::createSparseToDenseMask(
+    llvm::StringRef name, NodeValue indices, NodeValue values,
+    NodeValue defaultValue, NodeValue lengths, llvm::ArrayRef<int64_t> mask) {
+  auto lengthsDims = lengths.dims();
+  auto valueDims = defaultValue.dims();
+  ShapeVector outDims = {mask.size()};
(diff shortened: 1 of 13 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1442,6 +1442,21 @@ SparseToDenseNode *Function::createSparseToDense(llvm::StringRef name,
...
 }
 
+SparseToDenseMaskNode *Function::createSparseToDenseMask(
+    llvm::StringRef name, NodeValue indices, NodeValue values,
+    NodeValue defaultValue, NodeValue lengths, llvm::ArrayRef<int64_t> mask) {
+  auto lengthsDims = lengths.dims();
+  auto valueDims = defaultValue.dims();
+  ShapeVector outDims = {mask.size()};
+  // If lengths is 0-dimensional tensor, then there is no batch dimension.
+  if (lengthsDims.size() > 0)
```
bertmaher: nit: {} around single-line if

artemrakhov: Done.

artemrakhov:
@bertmaher I hope that documentation that I wrote makes sense and explains semantics. Implementation is similar to regular `SparseToDense`, and also `Gather`.

bertmaher:
The docs/comments seem good, it's just an inherently complicated op IMO :)

PR #2336, opened 2019-02-02
[quantization] Choose symmetric params even with infinities

bertmaher:
*Description*: This case is kind of pathological, but symmetric quantization was failing to find a proper zero point when min/max was infinite.  Work around it by forcing infinities to lowest/max as appropriate.
*Testing*: New QuantizationTest
*Documentation*: n/a

### DIFF
--- a/lib/Quantization/Base/Base.cpp
+++ b/lib/Quantization/Base/Base.cpp
@@ -277,7 +277,10 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
     }
   }
 
-  double scale = (max - min) / ((double)qmax - qmin);
+  min = std::max(min, std::numeric_limits<float>::lowest());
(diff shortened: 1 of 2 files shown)


lib/Quantization/Base/Base.cpp
```diff
@@ -277,7 +277,10 @@ TensorQuantizationParams chooseQuantizationParams(float min, float max,
     }
   }
 
-  double scale = (max - min) / ((double)qmax - qmin);
+  min = std::max(min, std::numeric_limits<float>::lowest());
```
rdzhabarov: @bertmaher I do not think this is a correct fix.

Having this will make scale values to be way overestimated (std::numeric_limits<float>::lowest() is too small number) and basically accuracy drop will be huge.
This needs to be handled on a caller side, min and max supplied here should not be infinities.
It will allow us to represent non infinity numbers correctly.

PR #2337, opened 2019-02-04
Use OutputCheck instead of LIT/FileCheck

jackm321:
*Description*:
Use [OutputCheck](https://github.com/stp/OutputCheck) (thanks @opti-mix for finding this) instead of FileCheck and use a shell script instead of lit.
This direction has the advantages of a. not having to install FileCheck and lit in different environments and b. not having to rewrite either one. If this seems like a good direction then we can go with it.
Any suggestions especially on how to improve the scripts would be awesome. Some things to improve on this include:
* Use find_program
* Check that models are available, CPU flag was passed, etc before running tests
* Clean up cmake variables
* Only run test in special mode similar to RELEASE_WITH_LIT

*Testing*:
Ran `cmake -G Ninja .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/usr/local/opt/llvm@7 && ninja outputCheckTests` locally.

*Documentation*:
comments


tests/text-translator/run_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
jfix71: I would include `set -euxo pipefail` in all bash scripts.

tests/text-translator/run_tests.sh
```diff
@@ -0,0 +1,12 @@
...
+# run a given script in the same directory as the current script
+run_and_check_output() {
```
jfix71: Assuming this is going to be used in other tests, you could move this helper to another generic helper file and source it to include it.

tests/text-translator/run_tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
jfix71: If this PR eventually lands, probably should add the copyright header to even these sort of trivial scripts I guess 🙁

qcolombet:
Just throwing that here, but why depending on `OutputCheck` is better than depending on `FileCheck`. Given we already depend on LLVM, I fell we are increasing our dependencies instead of decreasing them.

qcolombet:
Note: i’ve already said that a while back but we could stop installing llvm and just add it as a sub module...

jackm321:
After an offline discussion, I will do the following
* Move `| OutputCheck` to each script and not overcomplicate the build files
* Look for ways to run normal tests independently of the heavier OutputCheck tests without using a separate build flag `GLOW_BUILD_OUTPUTCHECK_TESTS`. Instead we should run a different set of tests normally verses in special cases when we want to run heavy output checks as well. Perhaps this can be done by using the ctest `--exclude-regex <regex>` flag to exclude heavy Filecheck tests during normal testing.

PR #2345, opened 2019-02-05
[New Operator] Implement Mod operator

### LINKED ISSUE #2330: [New operator] Mod
We need to support `Mod` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#mod

jackm321:
*Description*:
Implements caffe2 Mod operator.
I chose to call the IR node/instruction Modulo instead of Mod to avoid confusion with Modules which are referred to throughout the codebase as mod.

*Testing*:
`ninja test`

*Documentation*:
doxygen

Fixes #2330

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1669,8 +1669,8 @@ void BoundInterpreterFunction::fwdModuloInstImpl(glow::ModuloInst const *I) {
   auto srcH = getTensor(I->getSrc())->getHandle<ElemTy>();
   auto destH = getTensor(I->getDest())->getHandle<ElemTy>();
 
-  unsigned_t divisor = I->getDivisor();
-  bool signFollowDivisor = I->getSignFollowDivisor();
+  auto divisor = I->getDivisor();
+  auto signFollowDivisor = I->getSignFollowDivisor();
 
   for (size_t i = 0, e = srcH.size(); i < e; i++) {
     auto res = srcH.raw(i) % divisor;
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1652,6 +1652,21 @@ void BoundInterpreterFunction::fwdElementSelectInst(
                             I->getLHS()->getElementType(), I);
 }
 
+void BoundInterpreterFunction::fwdModuloInst(glow::ModuloInst const *I) {
+  auto srcH = getTensor(I->getSrc())->getHandle<int64_t>();
+  auto destH = getTensor(I->getDest())->getHandle<int64_t>();
+  int64_t divisor = I->getDivisor();
+  bool signFollowDivisor = I->getSignFollowDivisor();
+
+  for (size_t i = 0, e = srcH.size(); i < e; i++) {
+    auto res = srcH.raw(i) % divisor;
+    if (signFollowDivisor && res <= 0) {
+      res += divisor;
+    }
+    destH.raw(i) = res;
+  }
+}
+
 //===----------------------------------------------------------------------===//
 //                       Mat Mul
 //===----------------------------------------------------------------------===//
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -38,6 +38,18 @@ using namespace glow;
     llvm_unreachable("Type is not supported");                                 \
   }
 
+#define dispatchIndexTypeImpl(functionName, elemTy, ...)                       \
+  switch (elemTy) {                                                            \
+  case ElemKind::Int32ITy:                                                     \
+    functionName<int32_t>(__VA_ARGS__);                                        \
+    break;                                                                     \
+  case ElemKind::Int64ITy:                                                     \
+    functionName<int64_t>(__VA_ARGS__);                                        \
+    break;                                                                     \
+  default:                                                                     \
+    llvm_unreachable("Type is not supported");                                 \
+  }
+
 #define dispatchQuantizedImpl(functionName, elemTy, ...)                       \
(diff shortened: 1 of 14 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1652,6 +1652,21 @@ void BoundInterpreterFunction::fwdElementSelectInst(
...
 
+void BoundInterpreterFunction::fwdModuloInst(glow::ModuloInst const *I) {
+  auto srcH = getTensor(I->getSrc())->getHandle<int64_t>();
+  auto destH = getTensor(I->getDest())->getHandle<int64_t>();
+  int64_t divisor = I->getDivisor();
+  bool signFollowDivisor = I->getSignFollowDivisor();
+
+  for (size_t i = 0, e = srcH.size(); i < e; i++) {
+    auto res = srcH.raw(i) % divisor;
+    if (signFollowDivisor && res <= 0) {
```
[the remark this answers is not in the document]
jackm321: @artemrakhov this implementation is equivalent to the code you linked and skips the unnecessary `divisor > 0` check.

lib/Graph/Nodes.cpp
```diff
@@ -1104,6 +1104,11 @@ bool ConcatNode::verify() const {
   return isValid;
 }
 
+bool ModuloNode::verify() const {
+  // Divisor can't be 0.
```
rdzhabarov: I think comment brings more confusion than clarification.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1652,6 +1652,21 @@ void BoundInterpreterFunction::fwdElementSelectInst(
                             I->getLHS()->getElementType(), I);
 }
 
+void BoundInterpreterFunction::fwdModuloInst(glow::ModuloInst const *I) {
+  auto srcH = getTensor(I->getSrc())->getHandle<int64_t>();
```
rdzhabarov: from the doc: Input tensor with int32 or int64 data.
This will fail in debug mode if tensor of int32 data.

jackm321: Good point. Do we have an existing solution for this? Looks like `dispatchFloatingPointImpl`, `dispatchQuantizedImpl`, and `dispatchQuantizedWithAccumulationImpl` don't quite fit this.

rdzhabarov: Yes, Int32ITy and Int64ITy are indexed types, not quantized types. So, you'd need to add something similar.

lib/Backends/Interpreter/InterpreterNodes.cpp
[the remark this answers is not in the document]
artemrakhov: I don't agree that it is equivalent. They never touch 0. But this code adds `divisor` to 0, making result of 0 % 3 to be 3. And the unit test #2 that you've added confirms this.

nadavrot:
@jackm321 @artemrakhov @artemrakhov Is there a reason why we are not templating all of the binary operators? Is there a difference between Mul, Sub, Add except to the one character that specifies the op?

lib/Backends/Interpreter/InterpreterNodes.cpp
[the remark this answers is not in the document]
jackm321: Yeah you're right about the `output_ptr[i]` check, I'll add that to this too

artemrakhov:
@nadavrot: This op is different, because of `signFollowDivisor` option. @jackm321 also suggests that it's not a binary op, because divisor is constant.

We do have a lot of templates around Add, Mul, Sub, Div. Quantization formulas are different, that's why it's not one character difference. I admit we could use more templates for these.

rdzhabarov:
> is there any reason not to merge `dispatchFloatingPointImpl`, `dispatchIndexTypeImpl`, and `dispatchQuantizedImpl`? If not I think we should do that. I can create task for it.

For many cases (ops) there are no quantized impl for a certain op. I think it makes sense to keep them separately.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -849,6 +849,27 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
   }
 
+  if (typeName == "Mod") {
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in,
+                               getNodeValueOrCreateConstantByName(op.input(0)));
+    int64_t divisor;
+    ASSIGN_VALUE_OR_RETURN_ERR(divisor, loadInt(dict["divisor"]));
+
+    RETURN_ERR_IF_NOT(divisor >= 1, "Divisor must be positive.");
```
rdzhabarov: not less than 1?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -849,6 +849,27 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+
+    RETURN_ERR_IF_NOT(divisor >= 1, "Divisor must not be less than 1.");
+
+    bool signFollowDivisor = false;
+    if (dict.count("sign_follow_divisor")) {
+      ASSIGN_VALUE_OR_RETURN_ERR(signFollowDivisor,
+                                 loadInt(dict["sign_follow_divisor"]));
+    }
+
+    auto *node = G_.createModulo(opName, in, divisor, signFollowDivisor);
```
artemrakhov: At this point `divisor` is silently converted from 64-bit to 32-bit. Is it intentional? Caffe2 uses `int64_t`

jackm321: fixed

jackm321:
Thanks @rdzhabarov and @artemrakhov for reviewing!

PR #2350, opened 2019-02-05
Fix resnet-concurrent example

nickgg:
*Description*: The resnet50 model expects a batch, and we were giving it only a single image which caused a "Mismatch on Placeholder and Tenor types." assertion in the ExecutionEngine. This actually works fine in release, but fixing the debug build is easy.
*Testing*: Ran resnet-concurrent example in Debug.
*Documentation*:

### DIFF
--- a/examples/resnet-concurrent.cpp
+++ b/examples/resnet-concurrent.cpp
@@ -196,9 +196,11 @@ int main(int argc, char **argv) {
                                   ImageChannelOrder::BGR, ImageLayout::NCHW,
                                   /* useImagenetNormalization */ true);
 
+    Tensor batch = image.getUnowned(inputType->dims());
+
     auto ctx = llvm::make_unique<Context>();
     ctx->allocate(module.getPlaceholders());
-    updateInputPlaceholders(*ctx, {input}, {&image});
+    updateInputPlaceholders(*ctx, {input}, {&batch});
 
     dispatchClassify(started, devices[started % numDevices].get(),
                      std::move(path), output, std::move(ctx), returned,


examples/resnet-concurrent.cpp
```diff
@@ -196,9 +196,12 @@ int main(int argc, char **argv) {
                                   ImageChannelOrder::BGR, ImageLayout::NCHW,
                                   /* useImagenetNormalization */ true);
 
+    Tensor batch(inputType);
+    batch.getHandle<float>().insertSlice(image, 0);
```
jfix71: Can we instead get an unowned tensor of `image` with the correct dimensions? Using insertSlice will require copying the image data.

nickgg:
@rdzhabarov yeah I think llvm::outs() is not threadsafe between arguments. GLOG might be a better choice for the runtime.

@jfix71 Initially I thought not, since getUnowned is pretty serious about the tensor sizes matching, but we can just create a new unowned tensor and things work because the batch size is fixed to one. This wouldn't work for other batches so perhaps we should add a version of readPngImage which writes into an existing tensor?

examples/resnet-concurrent.cpp
```diff
@@ -196,9 +196,13 @@ int main(int argc, char **argv) {
                                   ImageChannelOrder::BGR, ImageLayout::NCHW,
                                   /* useImagenetNormalization */ true);
 
+    // The model expects a batch, but since we only want one item, the layout is
+    // the same and we can create an unowed tensor.
+    Tensor batch(image.getUnsafePtr(), inputType);
```
jfix71: I'd prefer `Tensor batch = image.getUnowned(inputType->dims());` -- I like to avoid calls to `getUnsafePtr()`, and `getUnowned()` includes asserts for the sizes to make sense.

nickgg: yeah thats much better, I was sure this would trip the asserts on dimensions since the new tensor is larger.

nickgg:
I made #2358 to track improving the readImage interface.

rdzhabarov:
> yeah I think llvm::outs() is not threadsafe between arguments. GLOG might be a better choice for the runtime.

Whether we use GLOG or not probably part of a bigger discussion, I think we should fix current issue. Does not need to be in this PR.

rdzhabarov:
> How do you mean fix the current issue? What are you suggesting?

e.g., lock around llvm::outputs()

nickgg:
You want to globally lock the log function? I **strongly** disagree with that approach. It's disastrous for performance in general, and worse it makes every new log line a performance hit which means we'll under log in fear of regressing perf.

rdzhabarov:
(1) Well, I do not propose locking in production code, that's just in this example.
e.g., https://github.com/pytorch/glow/pull/2350/files#diff-6b46aaa4bab6d284991bc0fa35d6a772R96.

(2) On the other hand we might use that logging facility inside DeviceManager or other cases where I agree we should not lock around those things.

Considering that there is (2), it's indeed problematic.

nickgg:
Ah right, don't have an issue locking that specific call to log, but it doesn't address the general issue.

The reason I suggested GLOG is that unlike llvm::outs() (et al) it only flushes on EOF, rather than on each argument - which makes it easier to avoid interleaving among threads.

PR #2352, opened 2019-02-06
[Runtime] Fix heap-use-after-free TSAN warning in ThreadPoolExecutor

SplitInfinity:
**Description**
This commit fixes a heap-use-after-free bug in `ThreadPoolExecutor`
detected by TSAN. The symptom of the bug is that on some invocations of
`ExecutorTest`, the `ThreadPoolExecutor` gets destroyed before
`ThreadPoolExecutor::handleDeviceManagerResult` finishes executing. This
happens because the callback provided in `ThreadPoolExecutor::run()` is
called *before* erasing the execution state for the run that just
finished. The invocation of the callback fills the `std::future` that the
test thread waits on before calling EXPECT* and ASSERT* on the values
returned by the `ThreadPoolExecutor`. Sometimes, the test thread checks
the return values and exists the scope of the test before the rest of
`handleDeviceManagerResult` can execute, which causes
`executionStateMap_` to be destroyed before the thread executing
`handleDeviceManagerResult` calls `erase` on it.

This commit fixes this bug by adding an `InflightBarrier` class  that
keeps track of calls to `DeviceManager::runFunction` that the executor
hasn't received callbacks for or has not finished processing the result of.
The barrier is incremented before calling `DeviceManager::runFunction`
and decremented after. `~ThreadPoolExecutor` waits for this barrier to
hit zero before allowed the destruction of the object.

**Testing**
Ran ExecutorTest with TSAN 100 times locally and with ASAN 50 times
locally.


SplitInfinity:
Hmm, now that I think about it, maybe it's better to just switch the order of the erase and the invocation of the callback in `ThreadPoolExecutor::handleDeviceManagerResult`?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -173,6 +205,11 @@ void ExecutionState::setResultCode(const ResultCode resultCode) {
 }
 
 ThreadPoolExecutor::~ThreadPoolExecutor() {
+  // Wait for all inflight DeviceManager::runFunction() calls to return and be
+  // processed before starting to destroy state that is used in
+  // handleDeviceManagerResult().
+  inflightBarrier_.wait();
+
   // Lock executionStatesMutex_ here to satisfy TSAN.
   std::lock_guard<std::mutex> lock(executionStatesMutex_);
```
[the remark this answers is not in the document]
SplitInfinity: From a practical standpoint, it doesn't make sense. But without locking here, TSAN reports a data race between the thread that calls `executionStates_.erase()` in `handleDeviceManagerResult` and the thread that calls `~ThreadPoolExecutor` (the destructor). This happens frequently in the executor tests when the `ThreadPoolExecutor` instance being tested is destroyed when the test finishes and the text fixture is destroyed while there is still a thread executing `handleDeviceManagerResult`.

lib/Runtime/Executor/ThreadPoolExecutor.h
```diff
@@ -25,6 +26,32 @@
 namespace glow {
 namespace runtime {
 
+/// This class implements a simple barrier with which to wait for all threads
```
nickgg: You've implemented a semaphore here. These kind of low level primitives are generally much more efficient when you use OS level primitives (e.g. mach semaphore, futex syscall). Not for now but important in the future.  I wonder if we should break it out into it's own class/file right away and make that work easier.

lib/Runtime/Executor/ThreadPoolExecutor.h
```diff
@@ -25,6 +26,32 @@
...
+  void decrement(unsigned decr = 1);
+
+  /// Increment the count of threads in the barrier by \p incr.
+  void increment(unsigned incr = 1);
+
+  /// \returns the current count of the barrier.
+  unsigned count();
+
+  /// Wait for the barrier count to hit zero before continuing. This is
+  /// potentially a blocking call.
```
nickgg: It's definitely a blocking call, in fact all methods in this class are because you grab a lock.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -173,9 +205,10 @@ void ExecutionState::setResultCode(const ResultCode resultCode) {
 }
 
 ThreadPoolExecutor::~ThreadPoolExecutor() {
-  // Lock executionStatesMutex_ here to satisfy TSAN.
-  std::lock_guard<std::mutex> lock(executionStatesMutex_);
-  (void)lock;
+  // Wait for all inflight DeviceManager::runFunction() calls to return and be
+  // processed before starting to destroy state that is used in
+  // handleDeviceManagerResult().
+  inflightBarrier_.wait();
```
nickgg: You can avoid a race here by adding a `stopped` flag to the class which prevents further enqueues, this can return while there is an active call to run() which hasn't yet incremented the barrier. You should check that flag under the barrier's lock.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -212,7 +245,9 @@ void ThreadPoolExecutor::run(const DAGNode *root,
   // done here instead of inside executeDAGNode() so that a node can be
   // executed while placeholders are being propagated for the next node without
   // the callback for that node deleting the execution state.
-  executionState->incrementInflightNodes((root->children).size());
+  auto numChildren = (root->children).size();
+  executionState->incrementInflightNodes(numChildren);
+  inflightBarrier_.increment(numChildren);
```
nickgg: You want to increment the inflight barrier first before doing anything else, because that is what's stopping the class from being deleted.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -264,8 +300,9 @@ void ThreadPoolExecutor::executeDAGNode(
 
   if (deviceManagerIt == deviceManagers_.end()) {
     // Mark the node as no longer executing.
-    executionState->decrementInflightNodes();
     executionState->setResultCode(ResultCode::Failed);
+    executionState->decrementInflightNodes();
```
nickgg: IIRC you made these executionState inflight counters atomic to avoid having to lock while changing them, but I think with the lock around the class level inflight counter it's actually less efficient for these to be atomic rather than locked since it will create multiple memory barriers.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -351,6 +389,14 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
     std::lock_guard<std::mutex> lock(executionStatesMutex_);
     executionStates_.erase(executionState->getRunId());
```
rdzhabarov: might be missing something, but should we unblock in dtor of ThreadPoolExecutor once executionState_ is empty?
In this case we could simply signal on some CV if executionStates_ is empty and do a wait in dtor. In this case change would be less intrusive (no need for additional primitives, etc).

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: The devil is in the details 😢

nickgg: Concurrency is hard ☹️

Issue #2339, opened 2019-02-04
[RFC] Can we remove transformPreLowering()?


jfix71:
The original intent of `transformPreLowering()` was to allow for a backend to transform the graph before lowering makes it potentially much harder to understand the higher-level structure of the graph and perform preferred transformations.

However, we later added `shouldLower()` to allow a backend to prevent such problematic lowering. This means that instead of using `transformPreLowering()`, the backend can prevent lowering of nodes via `shouldLower()`, and then do its preferred transformations during `transformPostLowering()`.

So, some motivations here to remove `transformPreLowering()`:

1. I'm not sure any of our backends are using this functionality given the ability to prevent lowering via `shouldLower()`.
1. Simplify the compilation pipeline if it's not clearly useful.
1. Once https://github.com/pytorch/glow/pull/2311 lands, when quantizing we will be lowering prior to `transformPreLowering()` anyway, so it's a misleading function and could cause bugs and/or confusion.

PR #2365, opened 2019-02-07
Remove transformPreLowering()

jfix71:
*Description*: As mentioned in https://github.com/pytorch/glow/issues/2339, I don't think we need `transformPreLowering()` anymore.

1. I don't think any of our backends are using this functionality given the ability to prevent lowering via `shouldLower()`, and then later transform the graph during `transformPostLowering()`.
1. Simplify the compilation pipeline since it doesn't seem useful/used.
1. Now that https://github.com/pytorch/glow/pull/2311 has landed, when quantizing we lower prior to `transformPreLowering()` anyway, so it's a misleading function and could cause bugs and/or confusion.

*Testing*: All unit tests still pass.

*Documentation*: Updated.

Closes https://github.com/pytorch/glow/issues/2339

### DIFF
--- a/docs/Backends.md
+++ b/docs/Backends.md
@@ -44,19 +44,6 @@ are two pure virtual functions all backends must implement:
 
 Additionally, there are virtual functions that backends can override:
 
-- `virtual bool transformPreLowering(Function *F, CompilationMode mode) const;`
-
-  - Allow the backend to transform the `Function *F` before [node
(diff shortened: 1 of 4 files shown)


jfix71:
Yeah we aren't explicitly testing this functionality anywhere. Perhaps we could add some tests with MockBackends for `shouldLower()`, `transformPostLowering()`, etc.

PR #2368, opened 2019-02-09
[New Operator] FusedRowwiseQuantizedSparseLengthsWeightedSumNode

jfix71:
*Description*: As noted in https://github.com/pytorch/glow/pull/2292, we decided to implement both fused and unfused versions of rowwise-quantized SLWS.

*Testing*: Added OperatorTests and Caffe2ImporterTests.

*Documentation*: Added.

Closes https://github.com/pytorch/glow/issues/1698

### DIFF
--- a/docs/Quantization.md
+++ b/docs/Quantization.md
@@ -212,3 +212,16 @@ Row-wise quantized SparseLengthsWeightedSum is also supported. Similar to the
 above, we compute scales and offsets per row, to be used with the `Data` input
 for the `RowwiseQuantizedSparseLengthsSumNode`. Scales and Offsets are inputs to
 the node. Output of this node is float, matching the Caffe2 implementation.
+
+### Fused Row-wise Quantization
+
+For some backends it may be beneficial to keep each row's scales and offsets
+fused inline with the data. Caffe2 implements nodes with fused storage, such as
+[SparseLengthsWeightedSum](https://caffe2.ai/docs/operators-catalogue.html#sparselengthsweightedsumfused8bitrowwise). Glow
+supports such fused Nodes/Instructions, for example
+`FusedRowwiseQuantizedSparseLengthsWeightedSum`. The `ElemKind` of fused tensors
+is `Int8FusedQTy`. Tensors with `Int8FusedQTy` are 2-dimensional, and have an
+extra 8 columns for each row. The first extra 4 bytes are the float scale of the
+row, and the second extra 4 bytes are the in32_t offset. Note that similar to
+normal row-wise quantized tensors, they use a dummy scale and offset in the
+Type.
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -288,6 +288,17 @@ public:
       return false;
     }
 
+    // For now, make sure that either both or neither of the tensors have
+    // Int8FusedQTy. While it is possible for an Int8QTy tensor to equal a
+    // Int8FusedQTy tensor if the Int8FusedQTy tensor has the same scale/offset
+    // on all of its rows, and that scale/offset match that of the Int8QTy, we
+    // do not support checking this for now.
+    assert(((getElementType() == ElemKind::Int8FusedQTy &&
+             other.getElementType() == ElemKind::Int8FusedQTy) ||
+            (getElementType() != ElemKind::Int8FusedQTy &&
+             other.getElementType() != ElemKind::Int8FusedQTy)) &&
+           "Int8FusedQTy only supports comparing against same ElemKind.");
+
     switch (getElementType()) {
     case ElemKind::FloatTy:
       return isEqualImpl<float>(other, allowedError);
@@ -315,6 +326,11 @@ public:
       return isEqualImpl<int32_t>(other, allowedError);
     case ElemKind::Int64ITy:
       return isEqualImpl<int64_t>(other, allowedError);
+      // Note: We can use isEqualImpl() here because the scales/offsets will be
+      // compared as if they were data, so we will return false if any rowwise
+      // scale/offset do not match.
+    case ElemKind::Int8FusedQTy:
+      return isEqualImpl<int8_t>(other, allowedError);
     }
 
     // This is to make compiler happy. It can never reach this point as switch
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -107,6 +107,18 @@ public:
       auto *data = reinterpret_cast<int32_t *>(getData());
       std::fill(&data[0], &data[0] + size(), (int32_t)type_.getOffset());
     } break;
+    case ElemKind::Int8FusedQTy: {
+      assert(dims().size() == 2 && "Fused tensor must be 2-dimensional.");
+      assert(dims()[1] > 8 && "Fused tensor must have more than 8 columns.");
+      const size_t width = dims()[1];
+      auto *data = reinterpret_cast<int8_t *>(getData());
+      for (size_t i = 0, e = dims()[0]; i < e; i++) {
+        int8_t *scaleOffsetPtr = &data[(i + 1) * width] - 8;
+        int32_t offset;
+        memcpy(&offset, scaleOffsetPtr + 4, 4);
+        std::fill(&data[i * width], scaleOffsetPtr, (int8_t)offset);
+      }
+    } break;
     default:
       // Non-quantized tensors are set to 0.
       std::fill(&getData()[0], &getData()[0] + size() * type_.getElementSize(),
@@ -174,8 +186,9 @@ public:
   Tensor &operator=(const Tensor &other) = delete;
 
   /// Initialize the content of the tensor using the \p init method. The value
-  /// \p val is the initialization parameter. \p PRNG is used to generate
-  /// random numbers.
+  /// \p val is the initialization parameter. \p PRNG is used to generate random
+  /// numbers. Note that if the tensor's kind is Int8FusedQTy, then the fused
+  /// scaled/offsets will not be modified.
   void init(InitKind init, float val, PseudoRNG &PRNG);
 
   /// \returns unowned tensor using the same data buffer as the current tensor
@@ -717,8 +730,23 @@ public:
     assert(filterSize > 0 && "invalid filter size");
     double scale = std::sqrt(3.0 / double(filterSize));
     std::uniform_real_distribution<> dist(-scale, scale);
-    for (auto &e : *this) {
-      e = dist(PRNG);
+    switch (getElementType()) {
+    default: {
+      for (auto &e : *this) {
+        e = dist(PRNG);
+      }
+      return;
+    }
+    case ElemKind::Int8FusedQTy: {
+      assert(dims().size() == 2 && "Fused tensor must be 2-dimensional.");
+      assert(dims()[1] > 8 && "Fused tensor must have more than 8 columns.");
+      for (size_t i = 0, e = dims()[0]; i < e; i++) {
+        for (size_t j = 0, f = dims()[1] - 8; j < f; j++) {
+          at({i, j}) = dist(PRNG);
+        }
+      }
+      return;
+    }
     }
   }
 
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -185,13 +185,14 @@ inline bool operator==(const ShapeNCHW &LHS, const ShapeNCHW &RHS) {
 /// An enum representing the type used by the elements of a tensor. The types of
 /// Handles for these tensors should match the element kind.
 enum class ElemKind : unsigned char {
-  FloatTy,   // 32-bit float type (float)
-  Float16Ty, // 16-bit float type (half, fp16)
-  Int8QTy,   // 8-bit quantized type (int8_t)
-  Int16QTy,  // 16-bit quantized type (int16_t)
-  Int32QTy,  // 32-bit quantized type (int32_t)
-  Int32ITy,  // 32-bit index type (int32_t)
-  Int64ITy,  // 64-bit index type (int64_t)
+  FloatTy,      // 32-bit float type (float)
+  Float16Ty,    // 16-bit float type (half, fp16)
+  Int8QTy,      // 8-bit quantized type (int8_t)
+  Int16QTy,     // 16-bit quantized type (int16_t)
+  Int32QTy,     // 32-bit quantized type (int32_t)
+  Int32ITy,     // 32-bit index type (int32_t)
+  Int64ITy,     // 64-bit index type (int64_t)
+  Int8FusedQTy, // 8-bit quantized type with fused scale/offset (int8_t)
 };
 
 /// A class that represents a type of a tensor.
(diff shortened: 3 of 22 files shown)


jfix71:
Note that Caffe2 does not have an explicit "fused Int8" fill/type, so when the Caffe2Importer loads the Constant `data` input it is loaded as normal `Int8QTy`. During loading a `FusedRowwiseQuantizedSparseLengthsWeightedSumNode`, it assumes that its `data` input was previously loaded as `Int8QTy`, and then changes the type to `Int8FusedQTy` instead.

If there are other users of the same Constant `data` that are expecting it to be `Int8QTy` then it will have an incorrect type. However I don't expect this to be the case, and it doesn't make sense that other users would be unaware that it's fused data.

yinghai:
What's the status of this one? Is it ready for review?

jfix71:
@yinghai Yep!

include/glow/Base/Tensor.h
```diff
@@ -315,6 +315,11 @@ class Tensor final {
       return isEqualImpl<int32_t>(other, allowedError);
     case ElemKind::Int64ITy:
       return isEqualImpl<int64_t>(other, allowedError);
+      // Note: We can use isEqualImpl() here because the scales/offsets will be
+      // compared as if they were data, so we will return false if any rowwise
+      // scale/offset do not match.
+    case ElemKind::Int8FusedQTy:
```
rdzhabarov: having this, `isEqual` became a little bit misleading since it's possible that Int8Fused and Int8QT tensors will be equal, but just written in different representation.
I'm ok if we address that later, just put some more details on the issue about comparing two quantized tensors.

lib/Base/Tensor.cpp
```diff
@@ -368,6 +372,12 @@ void glow::genericTranspose(const Tensor *src, Tensor *dest,
     transposeSelectImpl(srcH, destH, shuffle);
     return;
   }
+  case ElemKind::Int8FusedQTy: {
+    auto srcH = src->getHandle<int8_t>();
+    auto destH = dest->getHandle<int8_t>();
+    transposeSelectImpl(srcH, destH, shuffle);
```
rdzhabarov: how would this work when scale and offset is embedded into the row?
Say original tensor has two rows and 2 columns, so we have 2 rows and 4 columns to store scale and offset.
After we do transpose we mess up things. Should we disallow transpose on those tensors?

jfix71: I considered disallowing it, but thought it might make sense for a backend to prefer the data layout to be different for some weird reason (though I don't really expect it would ever make sense). Like you said you could transpose it and then have it "column-wise" fused quantized. We don't have a current use case for it though so we could always just disallow it for now.

include/glow/Graph/Graph.h
```diff
@@ -573,18 +573,19 @@ class Function final : public Named {
                                  NodeValue data, NodeValue weights,
                                  NodeValue indices, NodeValue lengths);
 
-  /// Create a node, performing SparseLengthsSum operation, using rowwise
-  /// quantization for the input data. Gathers slices of the outer-most
-  /// dimension of Data indexed by Indices vector, and then accumulates them
-  /// into len(Lengths) entries: first Lengths[0] slices are aggregated to
-  /// Result[0], next Lengths[1] slices are aggregated to Result[1],
-  /// etc. I.e. sum(Lengths) must be equal to len(Indices).
+  /// Create and \returns a node of \p name, performing the SparseLengthsSum
```
jackm321: nit "Creates and \returns" might sound better

include/glow/Graph/Graph.h
```diff
@@ -606,6 +607,45 @@ class Function final : public Named {
                                                  NodeValue indices,
                                                  NodeValue lengths);
 
+  /// Create and \returns a node of \p name, performing the SparseLengthsSum
+  /// operation, using fused rowwise quantization for the input \p data wherein
+  /// the scales and offsets are fused inline with each row of data. Gathers
```
jackm321: can you comment on how they are fused, maybe point to description in base.h

include/glow/Quantization/Base/Base.h
```diff
@@ -136,6 +136,14 @@ std::vector<int8_t> createMapping(TypeRef inTy, TypeRef outTy,
 void tensorRowwiseQuantization(const Tensor &input, Tensor &output,
                                Tensor &scales, Tensor &offsets);
 
+/// Fused-rowwise quantize the tensor \p input. Scales and offsets are generated
+/// from each row of \p input. \p output is tensor of the same shape as input
+/// but with 8 extra columns for storing fused scales and offset.
```
jackm321: maybe put a comment about the format here, something like data then 4 bytes for the scale then 4 bytes for offset

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -2318,6 +2325,29 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
     break;
   }
 
+  case Kinded::Kind::FusedRowwiseQuantizedSparseLengthsWeightedSumInstKind: {
+    auto *N = cast<FusedRowwiseQuantizedSparseLengthsWeightedSumInst>(I);
```
jackm321: > `FusedRowwiseQuantizedSparseLengthsWeightedSumInst`

wow these names are not brief

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1057,6 +1057,30 @@ void libjit_rowwise_quantized_sparse_lengths_weighted_sum_f(
   }
 }
 
+void libjit_fused_rowwise_quantized_sparse_lengths_weighted_sum_f(
+    float *dest, int8_t *data, float *weights, size_t *indices,
+    int32_t *lengths, size_t segments, size_t inLineSize, size_t outLineSize) {
+  memset(dest, 0, segments * outLineSize * sizeof(float));
+  size_t curIndex = 0;
+  for (size_t i = 0; i < segments; i++) {
+    for (int32_t j = 0; j < lengths[i]; j++) {
```
jackm321: ...`int32_t j = 0, end=lengths[i]; j < end`...

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1057,6 +1057,30 @@ void libjit_rowwise_quantized_sparse_lengths_weighted_sum_f(
...
+      const float weight = weights[curIndex];
+      const size_t line = indices[curIndex];
+      const int8_t *currRowScaleOffsetPtr =
+          data + ((line + 1) * inLineSize) - 8;
+      float scale;
+      int32_t offset;
+      memcpy(&scale, currRowScaleOffsetPtr, sizeof(float));
```
jackm321: is there a reason to prefer this over a reinterpret_cast?

lib/Base/Tensor.cpp
```diff
@@ -449,6 +463,10 @@ void Tensor::init(InitKind init, float val, PseudoRNG &PRNG) {
       getHandle<int64_t>().initXavier(val, PRNG);
       break;
     }
+    case ElemKind::Int8FusedQTy: {
+      getHandle<int8_t>().initXavier(val, PRNG);
```
jackm321: Do want to do anything special to not fill the last 8 bytes?

lib/Graph/Graph.cpp
```diff
@@ -1447,6 +1447,76 @@ Function::createRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
...
 }
 
+FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
+Function::createFusedRowwiseQuantizedSparseLengthsWeightedSum(
+    llvm::StringRef name, Constant *data, NodeValue weights, NodeValue indices,
+    NodeValue lengths) {
+  auto inDims = data->dims();
+  ShapeVector outDims(inDims.begin(), inDims.end());
+  outDims[0] = lengths.dims()[0];
+  outDims[1] -= 8;
```
jackm321: maybe add a comment here explaining this

lib/Base/Tensor.cpp
jfix71: I made this `llvm_unreachable()` for now.

include/glow/Base/Tensor.h
jfix71: I added added an assertion and a comment for this.

lib/Base/Tensor.cpp
jfix71: I added a commit to do this along with a test checking that it works. Not sure it'll ever really be used but better to at least handle it correctly.

lib/Backends/CPU/libjit/libjit.cpp
jfix71: Dereferencing a pointer doesn't work here for misaligned loads, so we can't just create a pointer to it and dereferencing it. We have to copy it to the stack.

lib/Backends/CPU/LLVMIRGen.cpp
jfix71: Yeah not really happy about it either but unless we want to start using acronyms, not sure how to shorten... 😞

include/glow/Graph/Graph.h
jfix71: I added a sentence saying data must have `ElemKind::Int8FusedQTy`. `ElemKind::Int8FusedQTy` isn't really specific to this node so I don't think it makes sense to explain the data layout there -- I expanded on the explanation a bit in `Quantization.md`.

lib/Backends/CPU/LLVMIRGen.cpp
rdzhabarov: FRQSLWS %) :D (good luck to someone reading this)

docs/Quantization.md
```diff
@@ -212,3 +212,16 @@ Row-wise quantized SparseLengthsWeightedSum is also supported. Similar to the
 above, we compute scales and offsets per row, to be used with the `Data` input
 for the `RowwiseQuantizedSparseLengthsSumNode`. Scales and Offsets are inputs to
 the node. Output of this node is float, matching the Caffe2 implementation.
+
+### Fused Row-wise Quantization
+
+For some backends it may be beneficial to keep each row's scales and offsets
+fused inline with the data. Caffe2 implements nodes with fused storage, such as
+[SparseLengthsWeightedSum](https://caffe2.ai/docs/operators-catalogue.html#sparselengthsweightedsumfused8bitrowwise). Glow
```
rdzhabarov: > uint8 tensor obtained with operator FloatToFused8BitRowwiseQuantized

How do we use the fact that it's uint8 stored, but not int8?

It seems like in C2 loaders we assume it's int8 type
> RETURN_ERR_IF_NOT(dataC->getElementType() == ElemKind::Int8QTy, "Data must be Int8QTy.");

include/glow/Quantization/Base/Base.h
```diff
@@ -136,6 +136,15 @@ std::vector<int8_t> createMapping(TypeRef inTy, TypeRef outTy,
 void tensorRowwiseQuantization(const Tensor &input, Tensor &output,
                                Tensor &scales, Tensor &offsets);
 
+/// Fused-rowwise quantize the tensor \p input. Scales and offsets are generated
+/// from each row of \p input. \p output is tensor of the same shape as input
+/// but with 8 extra columns for storing fused scales (4 bytes (columns) for
+/// float) and offset (4 bytes (columns) for int32_t.
```
rdzhabarov: nit: for int32_t)

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -1897,3 +1897,196 @@ TEST(caffe2, SparseLengthsSum8BitsRowwise) {
...
+/// RowwiseQuantizedSparseLengthsWeightedSumNode. The following inputs/outputs
+/// are used/expected for this test. Note that the DATA input is
+/// rowwise-quantized in the init_net proto.
+///    DATA  =   [[2.0, -0.5, 13]]
+///    WEIGHTS = [3, 1, 0, 0, 0, 0, 2, -0.5]
+///    INDICES = [1, 0, 2, 0, 1, 2, 2, 0]
+///    LENGTHS = [3, 0, 3, 2]
+///    OUTPUT =  [[0.5, 0, 0, 25]]
+TEST(caffe2, SparseLengthsWeightedSumFused8BitRowwise) {
+  ExecutionEngine EE{BackendKind::Interpreter};
```
rdzhabarov: can you do some on CPU backend as well, since we do verification against expected values.

jfix71: Well the OperatorTest already runs on the CPU backend. Why do you think it would be beneficial to test loading both on the CPU and Interpreter here? We don't do any of our Caffe2ImporterTests on the CPU backend.

rdzhabarov: this is a bit controversial, but C2Importer should test the import part not execution.
If it tests execution, why does that test only a single backend while there are impls in other backends as well.
I do not want to block on this anyway. Good to go.

include/glow/Base/Type.h
```diff
@@ -185,13 +185,14 @@ inline bool operator==(const ShapeNCHW &LHS, const ShapeNCHW &RHS) {
...
-  Int32ITy,  // 32-bit index type (int32_t)
-  Int64ITy,  // 64-bit index type (int64_t)
+  FloatTy,      // 32-bit float type (float)
+  Float16Ty,    // 16-bit float type (half, fp16)
+  Int8QTy,      // 8-bit quantized type (int8_t)
+  Int16QTy,     // 16-bit quantized type (int16_t)
+  Int32QTy,     // 32-bit quantized type (int32_t)
+  Int32ITy,     // 32-bit index type (int32_t)
+  Int64ITy,     // 64-bit index type (int64_t)
+  Int8FusedQTy, // 8-bit quantized type with fused scale/offset (int8_t)
```
yinghai: Why do we have such a type instead of plain `Int8ITy`?

docs/Quantization.md
```diff
@@ -212,3 +212,16 @@ Row-wise quantized SparseLengthsWeightedSum is also supported. Similar to the
...
+supports such fused Nodes/Instructions, for example
+`FusedRowwiseQuantizedSparseLengthsWeightedSum`. The `ElemKind` of fused tensors
+is `Int8FusedQTy`. Tensors with `Int8FusedQTy` are 2-dimensional, and have an
```
yinghai: It should be uint8?

include/glow/Base/Type.h
jfix71: The biggest reason I wanted to explicitly differentiate between normal `Int8QTy` and `Int8FusedQTy` is because the shape for `Int8FusedQTy` is very strange -- it appears to be 8 columns wider than the actual data. This prevents accidental interpretation of fused scales/offsets as instead data, or viewing the input data as an otherwise normal tensor. And I didn't think there was a large downside to adding the new type.

docs/Quantization.md
jfix71: We subtract `OFFSETSHIFT` to convert to `int8_t` in our importer:

https://github.com/pytorch/glow/blob/a64d64f2fdcce0819c75ae7a912d6f1ca6749c0a/lib/Importer/Caffe2ModelLoader.cpp#L1118-L1127

PR #2372, opened 2019-02-11
[Runtime][NFC] Runtime Cleanup

gcatron:
*Description*: Moved common types to RuntimeTypes, moved DeviceManager into runtime namespace.
*Testing*: Unit tests and resnet-runtime example pass.
*Documentation*: N/A

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -27,12 +27,11 @@
 #include <string>
 
 namespace glow {
+namespace runtime {
 
 /// Callback signalling success/failure of loading a Module onto a device.
 using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
-/// Callback signalling the result of running a function.
-using ResultCBTy = std::function<void(
-    runtime::RunIdentifierTy, runtime::ResultCode, std::unique_ptr<Context>)>;
+
 /// Map of Function name -> CompiledFunction, used when loading a network onto a
 /// device.
 using FunctionMapTy = std::map<std::string, CompiledFunction *>;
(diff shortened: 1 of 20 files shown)


include/glow/Backends/DeviceManager.h
```diff
@@ -27,12 +27,11 @@
 #include <string>
 
 namespace glow {
+namespace runtime {
 
 /// Callback signalling success/failure of loading a Module onto a device.
 using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
```
jackm321: why not move this to runtime types too?

gcatron: We can, it's only used in the deviceManagers (and the test deviceManager in executor test) So it decided to leave it in DeviceManger.h.

nadavrot:
@gcatron One way to communicate that this is just a refactoring change is to add the NFC label. NFC = No Functional Change. 

Example:
https://twitter.com/modocache/status/681671893963948032?lang=en

tests/unittests/ExecutorTest.cpp
```diff
@@ -35,7 +35,7 @@ using namespace glow::runtime;
 /// registerResult() should be called before calling Executor::run() in each
 /// test. The rest of the implementation of the DeviceManager interface exists
 /// to satisfy the compiler.
-class TestDeviceManager final : public DeviceManager {
+class TestDeviceManager final : public glow::runtime::DeviceManager {
```
jackm321: `glow::runtime` namespace is opened on line 30 why is this needed?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -19,16 +19,21 @@
...
 
 namespace glow {
 namespace runtime {
 
+class DeviceManager;
 using DeviceIDTy = size_t;
 using RunIdentifierTy = size_t;
 
+/// Map of DeviceIDTy -> DeviceManager.
+using DeviceManagerMapTy = std::map<DeviceIDTy, std::shared_ptr<DeviceManager>>;
```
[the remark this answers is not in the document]
jackm321: Perhaps a densemap would be best?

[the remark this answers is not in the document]
gcatron: Map uses less memory for small sizes and a stable ordering is useful for the provisioner. Shared pointers are for the executor, we may want to revisit that in the future though.

PR #2373, opened 2019-02-11
[testing] Move numerical FC tests to OperatorTest

gabagam:
Move the parts of `Caffe2ImporterTest` that check numerical correctness over to `OperatorTest`, to better separate FC operator loading tests from numerical correctness tests.

### DIFF
--- a/tests/unittests/Caffe2ImporterTest.cpp
+++ b/tests/unittests/Caffe2ImporterTest.cpp
@@ -612,6 +612,36 @@ TEST(caffe2, FC) {
       llvm::dyn_cast<FullyConnectedNode>(saveNode->getInput().getNode());
   ASSERT_TRUE(fcNode);
 
+  // Check the numerical values of the weights and biases.
+  {
+    const Constant* constant = mod.getConstantByName("weights");
+    ASSERT_TRUE(constant);
+    const Tensor& weights = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {3, 4};
+    const std::vector<float> expectedValues = {1.0f, 4.0f, 7.0f, 10.0f,   //
+                                               2.0f, 5.0f, 8.0f, 11.0f,   //
+                                               3.0f, 6.0f, 9.0f, 12.0f};
+    EXPECT_EQ(expectedDimensions, weights.dims().vec());
+    ASSERT_EQ(expectedValues.size(), weights.size());
+    const auto& elements = weights.getHandle();
+    for (size_t i = 0; i < expectedValues.size(); ++i) {
+      EXPECT_FLOAT_EQ(expectedValues.at(i), elements.raw(i)) << "Where i = " << i;
+    }
+  }
+  {
+    const Constant* constant = mod.getConstantByName("biases");
+    ASSERT_TRUE(constant);
+    const Tensor& bias = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {4};
+    const std::vector<float> expectedValues = {0.1f, 0.2f, 0.3f, 0.4f};
+    EXPECT_EQ(expectedDimensions, bias.dims().vec());
+    ASSERT_EQ(expectedValues.size(), bias.size());
+    const auto& elements = bias.getHandle();
+    for (size_t i = 0; i < expectedValues.size(); ++i) {
+      EXPECT_FLOAT_EQ(expectedValues.at(i), elements.raw(i)) << "Where i = " << i;
+    }
+  }
+
   // We don't actually check that the output is correct, because this is
   // already covered in the Operator.FC/* tests.
 }
@@ -654,6 +684,36 @@ TEST(caffe2, FCWithFlatten) {
   ASSERT_TRUE(fcNode);
   auto *reshape = llvm::dyn_cast<ReshapeNode>(fcNode->getInput());
   ASSERT_TRUE(reshape);
+
+  // Check the numerical values of the weights and biases.
+  {
+    const Constant* constant = mod.getConstantByName("weights");
+    ASSERT_TRUE(constant);
+    const Tensor& weights = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {3, 4};
+    const std::vector<float> expectedValues = {1.0f, 4.0f, 7.0f, 10.0f,   //
+                                               2.0f, 5.0f, 8.0f, 11.0f,   //
+                                               3.0f, 6.0f, 9.0f, 12.0f};
+    EXPECT_EQ(expectedDimensions, weights.dims().vec());
+    ASSERT_EQ(expectedValues.size(), weights.size());
+    const auto& elements = weights.getHandle();
+    for (size_t i = 0; i < expectedValues.size(); ++i) {
+      EXPECT_FLOAT_EQ(expectedValues.at(i), elements.raw(i)) << "Where i = " << i;
+    }
+  }
+  {
+    const Constant* constant = mod.getConstantByName("biases");
+    ASSERT_TRUE(constant);
+    const Tensor& bias = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {4};
+    const std::vector<float> expectedValues = {0.1f, 0.2f, 0.3f, 0.4f};
+    EXPECT_EQ(expectedDimensions, bias.dims().vec());
+    ASSERT_EQ(expectedValues.size(), bias.size());
+    const auto& elements = bias.getHandle();
+    for (size_t i = 0; i < expectedValues.size(); ++i) {
+      EXPECT_FLOAT_EQ(expectedValues.at(i), elements.raw(i)) << "Where i = " << i;
+    }
+  }
 
   // We don't actually check that the output is correct, because this is
   // already covered in the Operator.FCWithFlatten/* tests.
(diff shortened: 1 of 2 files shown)


rdzhabarov:
@beicy could you please take a look.

I remember you mentioned in the past the reason for having numerical correctness in some of the places for loaders.

SplitInfinity:
I will admit I did not look carefully enough when creating the issue for moving the tests of this particular op and so I did not realize this one also reads weights from the protobuf. 

Still, I would argue that it is sufficient to check that weights are loaded correctly and that the graph structure is as expected instead of actually carrying out the computation.

beicy:
When we load a FC, if the weights' dims is larger than 2, we do the flatten on the fly. That is, we don't really create a node to the flatten. Therefore, we need to test the correctness during loader to make sure we do flatten weights if necessary.

SplitInfinity:
@beicy 

What if we checked that the loading was done correctly (including any on-the-fly `Flatten` or `Transpose`) by checking that the `Constants` of the `Module` created by the loader for weights and bias have the expected values? I would argue that should be enough because `Flatten` and `Transpose` should only change the shape of the data, not the values.

And then we would have one FC numerical correctness test in `OperatorTest`.

SplitInfinity:
I read the code again and the cases with Flatten are more complicated, so I agree that we should not modify those.

But can we move the computation for the regular FC and transposed FC that don't have flatten to `OperatorTest`?

bertmaher:
Honestly I kind of like that the importer tests are lightweight e2e tests.  They take less than a second to run on my laptop.  What's the motivation for breaking them up?

SplitInfinity:
@bertmaher 

In my view, these are unit tests for the loader and running the nets duplicates the work done by `OperatorTest`. A unit test for the loader only needs to check whether the graph structure is right and whether `Constants` have been loaded properly if applicable.

Of the 33 tests, some of them run the net (Clip, BatchBoxCox) and others don't (MaxPool), so it's not like these were meant to be E2E tests in the first place. But maybe the task should actually be to add E2E tests for the loader test cases that don't have them?

I don't know how lightweight these could be, but based on the recent experience of a horse-racing enthusiast user, I would argue that true E2E tests would author a model in PyTorch/C2, export it to ONNX/Protobuf, and then load and run it using Glow.

bertmaher:
> I would argue that true E2E tests would author a model in PyTorch/C2

Yes, we should definitely have that.  I think we honestly have an abundance of unit tests, but hit tons of rough edges at the integration/e2e level, as you point out.

tests/unittests/OperatorTest.cpp
```diff
@@ -1842,6 +1842,62 @@ TEST_P(InterpAndCPU, IntConcat) {
   }
 }
 
+TEST_P(Operator, FC) {
+  auto *input =
```
beicy: We have IntFC in OperatorTest, which compares both float and quantized FC. Not sure if it is necessary to test this one again.

tests/unittests/OperatorTest.cpp
```diff
@@ -1842,6 +1842,62 @@ TEST_P(InterpAndCPU, IntConcat) {
...
+  std::vector<float> expectedValues = {14.1f, 32.2f, 50.3f,  68.4f,
+                                       32.1f, 77.2f, 122.3f, 167.4f};
+  EXPECT_TRUE(result.dims().vec() == expectedDimensions);
+  for (size_t i = 0; i < 2 * 4; i++)
+    EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
+}
+
+TEST_P(Operator, FCWithFlatten) {
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {2, 1, 3}, "input", false);
```
rdzhabarov: this is the only difference between two tests, please, remove code duplication.

tests/unittests/OperatorTest.cpp
```diff
@@ -1842,6 +1842,62 @@ TEST_P(InterpAndCPU, IntConcat) {
...
+  auto result = ctx_.get(S->getPlaceholder())->getHandle();
+  std::vector<size_t> expectedDimensions = {2, 4};
+  std::vector<float> expectedValues = {14.1f, 32.2f, 50.3f,  68.4f,
+                                       32.1f, 77.2f, 122.3f, 167.4f};
+  EXPECT_TRUE(result.dims().vec() == expectedDimensions);
+  for (size_t i = 0; i < 2 * 4; i++)
+    EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
+}
+
+TEST_P(Operator, FCWithFlatten) {
```
rdzhabarov: where does flattening happen?

SplitInfinity: Turns out it happens in `glow::lower`; FC is lowered to `Flatten`, `MatMul` and `BatchedAdd` for the CPU, Interpreter and OpenCL backends.

SplitInfinity: I would say it's a useful test to have even though it's not related to the issue that this PR is addressing. But it is only useful for backends that allow FC to be lowered...

rdzhabarov:
> horse-racing enthusiast user

lol

tests/unittests/OperatorTest.cpp
SplitInfinity: Agreed, let's remove this, then.

tests/unittests/OperatorTest.cpp
gabagam: Removed.

gabagam:
@SplitInfinity, @beicy and @rdzhabarov comments addressed.

tests/unittests/OperatorTest.cpp
```diff
@@ -1842,6 +1842,35 @@ TEST_P(InterpAndCPU, IntConcat) {
...
+
+  EE_.compile(CompilationMode::Infer, F_);
+  EE_.run(ctx_);
+
+  auto result = ctx_.get(S->getPlaceholder())->getHandle();
+  std::vector<size_t> expectedDimensions = {2, 4};
+  std::vector<float> expectedValues = {14.1f, 32.2f, 50.3f,  68.4f,
+                                       32.1f, 77.2f, 122.3f, 167.4f};
+  EXPECT_TRUE(result.dims().vec() == expectedDimensions);
+  for (size_t i = 0; i < 2 * 4; i++)
```
SplitInfinity: Could you add opening and closing braces (`{}`) for this loop? `clang-format` might take care of it too.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -605,16 +605,47 @@ TEST(caffe2, FC) {
...
+  }
+  {
+    const Constant *constant = mod.getConstantByName("biases");
+    ASSERT_TRUE(constant);
+    const Tensor &bias = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {4};
+    const std::vector<float> expectedValues = {0.1f, 0.2f, 0.3f, 0.4f};
+    EXPECT_EQ(expectedDimensions, bias.dims().vec());
+    ASSERT_EQ(expectedValues.size(), bias.size());
+    const auto &elements = bias.getHandle();
```
SplitInfinity: Same here.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -605,16 +605,47 @@ TEST(caffe2, FC) {
...
+    const Constant *constant = mod.getConstantByName("weights");
+    ASSERT_TRUE(constant);
+    const Tensor &weights = constant->getPayload();
+    const std::vector<size_t> expectedDimensions = {3, 4};
+    const std::vector<float> expectedValues = {1.0f, 4.0f, 7.0f, 10.0f, //
+                                               2.0f, 5.0f, 8.0f, 11.0f, //
+                                               3.0f, 6.0f, 9.0f, 12.0f};
+    EXPECT_EQ(expectedDimensions, weights.dims().vec());
+    ASSERT_EQ(expectedValues.size(), weights.size());
+    const auto &elements = weights.getHandle();
```
SplitInfinity: I don't think this is a reference type? `Constant::getHandle` returns a `Handle<ElemTy>`.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -605,16 +605,47 @@ TEST(caffe2, FC) {
...
   }
 
-  EE.compile(CompilationMode::Infer, F);
-  EE.run(ctx);
+  // High level check on the content of the graph. We have 1 FC node and 1 save.
+  EXPECT_EQ(F->getNodes().size(), 2);
+  auto *saveNode = getSaveNodeFromDest(output);
+  auto *fcNode =
+      llvm::dyn_cast<FullyConnectedNode>(saveNode->getInput().getNode());
+  ASSERT_TRUE(fcNode);
```
SplitInfinity: I *think* we can get away with an `EXPECT_TRUE` here since we don't dereference `fcNode` after this.

tests/unittests/OperatorTest.cpp
gabagam: Fixed.

Just note taht `clang-format` is not set up to enclose single-line `for` loops in braces. I'm running the `utils/format.sh` script before each commit, which runs `clang-format` under the hood.

tests/unittests/Caffe2ImporterTest.cpp
gabagam: Removed `&`.

tests/unittests/Caffe2ImporterTest.cpp
gabagam: You're right, it's not. Removed `&`.

tests/unittests/Caffe2ImporterTest.cpp
gabagam: Changed to `EXPECT_TRUE`.

PR #2374, opened 2019-02-11
[Conv3D] Types, plumbing, and verification

jackm321:
*Description*:
Create a new `Convolution3D` node and instruction and the associated verification and graph node creation methods. Interpreter implementation for `Convolution3DInst` and `Convolution3DGradInst` will be implemented next pr.
I considered using the existing `Convolution` node and gating everything on the dimensionality of the inputs because they share the same input and output structure however there wasn't much to be gained by doing this because much of the verification code can't be shared and neither will the backend implementation be able to be shared likely so it seemed better to just create the correct node up front during loading and not worry about which implementation to use after that.

*Testing*:
`ninja all`
Added a very basic test.

*Documentation*:
Doxygen

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -242,6 +242,20 @@ public:
                               unsigned_t kernel, unsigned_t stride,
                               unsigned_t pad, unsigned_t group);
 
+  Convolution3DNode *createConv3D(llvm::StringRef name, NodeValue input,
+                                  NodeValue filter, NodeValue bias,
+                                  TypeRef outTy,
+                                  llvm::ArrayRef<unsigned_t> kernels,
+                                  llvm::ArrayRef<unsigned_t> strides,
+                                  llvm::ArrayRef<unsigned_t> pads,
+                                  unsigned_t group);
+
+  Convolution3DNode *createConv3D(llvm::StringRef name, NodeValue input,
+                                  NodeValue filter, NodeValue bias,
+                                  TypeRef outTy, unsigned_t kernel,
+                                  unsigned_t stride, unsigned_t pad,
+                                  unsigned_t group);
+
   ConvertToNode *createConvertTo(llvm::StringRef name, NodeValue input,
                                  TypeRef outTy);
 
@@ -811,16 +825,28 @@ public:
                            float momentum = 0.9);
 
   ConvolutionNode *createConv(Context &ctx, llvm::StringRef name,
-                              NodeValue input, size_t depth,
+                              NodeValue input, size_t outChannels,
                               llvm::ArrayRef<unsigned_t> kernels,
                               llvm::ArrayRef<unsigned_t> strides,
                               llvm::ArrayRef<unsigned_t> pads,
                               unsigned_t group);
 
   ConvolutionNode *createConv(Context &ctx, llvm::StringRef name,
-                              NodeValue input, size_t depth, unsigned_t kernel,
-                              unsigned_t stride, unsigned_t pad,
-                              unsigned_t group);
+                              NodeValue input, size_t outChannels,
+                              unsigned_t kernel, unsigned_t stride,
+                              unsigned_t pad, unsigned_t group);
+
+  Convolution3DNode *createConv3D(Context &ctx, llvm::StringRef name,
+                                  NodeValue input, size_t outChannels,
+                                  llvm::ArrayRef<unsigned_t> kernels,
+                                  llvm::ArrayRef<unsigned_t> strides,
+                                  llvm::ArrayRef<unsigned_t> pads,
+                                  unsigned_t group);
+
+  Convolution3DNode *createConv3D(Context &ctx, llvm::StringRef name,
+                                  NodeValue input, size_t outChannels,
+                                  unsigned_t kernel, unsigned_t stride,
+                                  unsigned_t pad, unsigned_t group);
 
   /// Create a fully connected node with the given \p name, \p input and \p
   /// output depth. Trainable weight and bias variables are created implicitly.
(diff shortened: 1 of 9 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1873,13 +1940,73 @@ ConvolutionNode *Function::createConv(Context &ctx, llvm::StringRef name,
...
+                                          unsigned_t group) {
+  ShapeNHWCT idim(input.dims());
+  ShapeHWT kdim(kernels);
+  PaddingTLBRNF pdim(pads);
+  (void)pdim;
+  assert((idim.w + pdim.left + pdim.right) >= kdim.width &&
+         (idim.h + pdim.top + pdim.bottom) >= kdim.height &&
+         (idim.t + pdim.near + pdim.far) >= kdim.time &&
+         "buffer too small for selected stride");
+
```
beicy: Above there is "check3DKernelSize", do we still need to do the check twice?

nadavrot:
Nice!

tests/unittests/GraphTest.cpp
```diff
@@ -93,6 +93,27 @@ TEST(Graph, simpleTestConv) {
   EXPECT_GT(M.getInstrs().size(), 0);
 }
 
+TEST(Graph, simpleTestConv3D) {
```
jfix71: Doxygen!

lib/Graph/Graph.cpp
```diff
@@ -488,6 +502,32 @@ static void assertConvDims(NodeValue input, NodeValue filter, NodeValue bias,
...
+                             llvm::ArrayRef<unsigned_t> pads,
+                             unsigned_t group) {
+
+  ShapeNHWCT idim(input.dims());
+  ShapeHWT kdim(kernels);
+  (void)kdim;
+  check3DKernelSize(idim, kernels, pads);
+  assert(idim.c % group == 0 && "channels number must be divisible by groups");
+
+  // NOTE: here the N in NHWCT is abnormal because it is the number of filters
```
jfix71: I'm a little confused -- what is functionally different here for the logic in here? It seems to be the same as `assertConvDims()` except with the time dim, no?

include/glow/Graph/Graph.h
```diff
@@ -242,6 +242,20 @@ class Function final : public Named {
...
 
+  Convolution3DNode *createConv3D(llvm::StringRef name, NodeValue input,
+                                  NodeValue filter, NodeValue bias,
+                                  TypeRef outTy,
+                                  llvm::ArrayRef<unsigned_t> kernels,
+                                  llvm::ArrayRef<unsigned_t> strides,
+                                  llvm::ArrayRef<unsigned_t> pads,
+                                  unsigned_t group);
+
+  Convolution3DNode *createConv3D(llvm::StringRef name, NodeValue input,
```
jfix71: Doxygen!

include/glow/Graph/Graph.h
```diff
@@ -771,16 +785,28 @@ class Function final : public Named {
...
 
   ConvolutionNode *createConv(Context &ctx, llvm::StringRef name,
-                              NodeValue input, size_t depth, unsigned_t kernel,
-                              unsigned_t stride, unsigned_t pad,
-                              unsigned_t group);
+                              NodeValue input, size_t outChannels,
+                              unsigned_t kernel, unsigned_t stride,
+                              unsigned_t pad, unsigned_t group);
+
+  Convolution3DNode *createConv3D(Context &ctx, llvm::StringRef name,
```
jfix71: Doxygen!

include/glow/Graph/Graph.h
```diff
@@ -771,16 +785,28 @@ class Function final : public Named {
...
+                                  NodeValue input, size_t outChannels,
+                                  llvm::ArrayRef<unsigned_t> kernels,
+                                  llvm::ArrayRef<unsigned_t> strides,
+                                  llvm::ArrayRef<unsigned_t> pads,
+                                  unsigned_t group);
+
+  Convolution3DNode *createConv3D(Context &ctx, llvm::StringRef name,
```
jfix71: Doxygen!

lib/Graph/Graph.cpp
jackm321: The functionality here is the same except with the extra dimension. The note is referring to the fact that normally in things with shape NHWCT the N dimension is the number of examples but because this is being used on the filter, N is actually the number of output channels of the convolution which is different. The rest of the channels HWCT are all the same as what you'd expect them to be.

lib/Graph/Graph.cpp
```diff
@@ -488,6 +502,32 @@ static void assertConvDims(NodeValue input, NodeValue filter, NodeValue bias,
   assert(bias.getType()->size() == filterDims[0] && "Invalid bias size");
 }
 
+static void assertConv3DDims(NodeValue input, NodeValue filter, NodeValue bias,
```
jfix71: Doxygen missing here too. Maybe could note it's essentially the 2D case but with the extra dimension checking.

lib/Graph/Graph.cpp
jfix71: Ah ok, so then this note is applicable to 2D convs too -- this was a little confusing of a note since it's only in `assertConv3DDims()`, because I thought it was calling out a difference between 2D and 3D. Maybe makes sense to call out specifically that there's no difference, or include the note in the 2D case, or move the note to somewhere else?

jackm321: @jfix71 it doesn't apply to the 2D implementation because, although the fliters are laid out the same, the 2D implementation doesn't choose to use of `NHWC` and instead indexes directly into the shape vectors. I think that way is less readable so I used `NHWCT` for the 3D case even though it's technically not the correct usage of `NHWCT` beacuse the N dimension is different than normal which is what I noted here. I think I'll just changes the 2D implementation to use `NHWC` and add the same note there as well since it's just easier to read and understand what's going on because you can understand what each dimension of the filters tensor is.

jfix71: Ahh ok got it, understood, thanks!

jackm321:
Thanks @beicy and @jfix71 for reviewing!

PR #2375, opened 2019-02-11
Add tracing/profiling for the CPU and Interpreter backends

nickgg:
*Description*: Implement auto instrumentation and handling of TraceEvents in the CPU and Interpreter backends. If instrumentation is enabled then every Instruction in the IR will be timed. 

The included example (tracing-compare) runs resnet50 on both the CPU and Interpreter backends and outputs a json file in the chromium TraceEvents format, which can be loaded in the chrome about://tracing page. It looks like this:

tracing-compare

*Testing*: The example and manual testing. Will write unit tests in a follow up.
*Documentation*:

### DIFF
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -1,0 +1,178 @@
+/**
+ * Copyright(c) 2017 - present, Facebook Inc.
+ *
+ * Licensed under the Apache License, Version 2.0(the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "glow/Runtime/RuntimeTypes.h"
+
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+
+#include <chrono>
+#include <future>
+
+using namespace glow;
+using namespace glow::runtime;
+
+std::array<BackendKind, 2> supportedBackends({BackendKind::Interpreter,
+                                              BackendKind::CPU});
+
+namespace {
+llvm::cl::OptionCategory category("tracing-compare Options");
+llvm::cl::opt<std::string>
+    inputImage(llvm::cl::desc("path to input image to classify, which must be "
+                              "a png with standard imagenet normalization"),
+               llvm::cl::init("../tests/images/imagenet/dog_207.png"),
+               llvm::cl::Positional, llvm::cl::cat(category));
+llvm::cl::opt<std::string>
+    outputJson(llvm::cl::desc("path to write output json trace events"),
+               llvm::cl::init("./glow-trace.json"), llvm::cl::Positional,
+               llvm::cl::cat(category));
+
+} // namespace
+
+/// Loads the model into /p module and returns the input and output
+/// Placeholders.
+std::pair<Placeholder *, Placeholder *> loadResnet50Model(TypeRef inputType,
+                                                          Module &module) {
+  Function *F = module.createFunction("resnet50");
+
+  llvm::outs() << "Loading resnet50 model.\n";
+
+  const char inputName[] = "gpu_0/data";
+  Caffe2ModelLoader loader("resnet50/predict_net.pb", "resnet50/init_net.pb",
+                           {inputName}, {inputType}, *F);
+  Placeholder *input =
+      llvm::cast<Placeholder>(cantFail(loader.getNodeValueByName(inputName)));
+  Placeholder *output = cantFail(loader.getSingleOutput());
+
+  return std::make_pair(input, output);
+}
+
+/// Compiles the resnet50 function.
+std::unique_ptr<CompiledFunction> compileModel(Module &module,
+                                               BackendKind backendKind) {
+  auto *backend = createBackend(backendKind);
+  Function *F = module.getFunction("resnet50");
+
+  llvm::outs() << "Starting compile.\n";
+  backend->optimizeFunction(CompilationMode::Infer, F);
+  return backend->instrumentAndCompile(F);
+}
+
+std::future<ResultCode> addToDevice(unsigned int id, DeviceManager *device,
+                                    Module &module, FunctionMapTy functions) {
+  std::shared_ptr<std::promise<ResultCode>> compilePromise(
+      new std::promise<ResultCode>);
+  auto future = compilePromise->get_future();
+
+  device->addNetwork(&module, functions,
+                     [compilePromise, id](const Module *, ResultCode code) {
+                       if (code != ResultCode::Ready) {
+                         llvm::errs() << "Failed to compile model for device "
+                                      << id << ".\n";
+                       } else {
+                         llvm::outs()
+                             << "Successfully added to Device " << id << ".\n";
+                       }
+                       compilePromise->set_value(code);
+                     });
+
+  return future;
+}
+
+int main(int argc, char **argv) {
+  llvm::cl::ParseCommandLineOptions(
+      argc, argv, "Run resnet and export a json file containing trace events");
+
+  std::array<DeviceManager *, supportedBackends.size()> devices;
+  for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i) {
+    devices[i] = DeviceManager::createDeviceManager(supportedBackends[i],
+                                                    "tracing-compare");
+    devices[i]->init();
+  }
+
+  // Load and compile model.
+
+  Module module;
+  TypeRef inputType(module.uniqueType(ElemKind::FloatTy, {1, 3, 224, 224}));
+  Placeholder *input, *output;
+
+  std::tie(input, output) = loadResnet50Model(inputType, module);
+
+  std::array<std::unique_ptr<CompiledFunction>, supportedBackends.size()>
+      compiledFunctions;
+
+  for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i) {
+    compiledFunctions[i] = compileModel(module, supportedBackends[i]);
+
+    FunctionMapTy functions;
+    functions.emplace("resnet50", compiledFunctions[i].get());
+
+    auto f = addToDevice(i, devices[i], module, functions);
+    f.wait_for(/* timeout_duration */ std::chrono::seconds(30));
+    if (f.get() != ResultCode::Ready) {
+      return 1;
+    }
+  }
+
+  auto image =
+      readPngImageAndPreprocess(inputImage, ImageNormalizationMode::k0to1,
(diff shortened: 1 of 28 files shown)


include/glow/Backends/Backend.h
```diff
@@ -44,6 +44,12 @@ class Backend {
   /// Generate code for input function \param F.
   virtual std::unique_ptr<CompiledFunction> compile(Function *F) const = 0;
 
+  /// Generate instrumented (for tracing) code for input function \param F.
```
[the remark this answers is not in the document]
jfix71: Should this throw an error if you try to instrument a Function for a backend that does not support instrumentation?

include/glow/Backends/Backend.h
```diff
@@ -86,6 +92,13 @@ class Backend {
   virtual bool generateInst(Node *N, IRGenVisitor &irgen) const {
     return false;
   }
+
+protected:
+  /// Inserts a TraceEventInst between every instruction, the most basic form of
+  /// auto instrumentation. Necessary only if the Backend doesn't provide
+  /// profiling/tracing in another way.
+  /// Modifies \p IR and returns a new TraceInfo map.
```
jfix71: Mention `\p traceEventDataSize`

lib/Backends/Backend.cpp
```diff
@@ -40,3 +42,67 @@ void Backend::optimizeFunction(CompilationMode mode, Function *F) {
...
+    return TraceInfo(false, traceEventDataSize);
+  }
+
+  Function *F = IR->getGraph();
+  // Get all instructions in the IRFunction.
+  IRFunction::InstListTy &instructions = IR->getInstrs();
+
+  // Build a placeholder to a backing tensor with space to fit all timestamps.
+  auto *backingPH = F->getParent()->createPlaceholder(
+      ElemKind::Int64ITy, {instructions.size() + 1, traceEventDataSize},
```
jfix71: I'm confused by the dimensions here -- why is the second dimension here `traceEventDataSize`?

include/glow/Backends/TraceEvents.h
```diff
@@ -0,0 +1,87 @@
...
+/// Tracing / Profiling events map for a compiled function.
+struct TraceInfo {
+  TraceInfo() = default;
+  TraceInfo(bool e, size_t d) : enabled(e), dataSize(d) {}
+
+  /// Whether tracing is enabled for this run.
+  bool enabled{false};
+
+  /// The size of each item in the backing Tensor.
+  size_t dataSize{0};
```
jfix71: Isn't this always going to be a single value across all infos for a given backend? If so, would it be possible to just query this from the backend instead of storing it in each `TraceInfo`?

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1531,6 +1532,13 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
   }
 }
 
+void libjit_write_timestamp(uint64_t *tensor, size_t offset) {
+  uint64_t ts = std::chrono::duration_cast<std::chrono::milliseconds>(
```
opti-mix: In general, we try to avoid calling any C++ functions from the `libjit` to avoid any possible issues with C++ symbols resolution at run-time. If it is possible to replace it by an equivalent `libc`-call, it would be nice.

include/glow/Graph/Context.h
```diff
@@ -48,6 +49,9 @@ class Context final {
   /// Maps Placeholder names to Placeholders.
   PlaceholderNameMap nameMap_;
 
+  /// Trace Events recorded during this run.
+  std::vector<TraceEvent> traceEvents_;
```
opti-mix: As far as I can see, the intent here is to associate `traceEvents_` with each run.  And `Context` is de-facto a data structure associated with each run.

One problem though is that `Context` is not guaranteed to be associated with each/any run and/or only a single run. 

If we want to extend the semantics of `Context` to guarantee it is associated with a run, we should update the documentation for `Context` and make any such guarantees very explicit there.

Alternatively, we may need to introduce a new data structure to hold the information about runs. This data structure is likely to include `traceEvents_` and probably one or more `Contexts`. This is, of course, something for a separate PR.

examples/tracing-compare.cpp
```diff
@@ -0,0 +1,178 @@
...
+                                ImageChannelOrder::BGR, ImageLayout::NCHW,
+                                /* useImagenetNormalization */ true);
+
+  Tensor batch = image.getUnowned(inputType->dims());
+
+  llvm::outs() << "Starting Run.\n";
+  std::array<std::promise<std::unique_ptr<Context>>, supportedBackends.size()>
+      promises;
+
+  for (unsigned int i = 0; i < supportedBackends.size(); ++i) {
```
opti-mix: Minor stylistic nit: 
We usually use a `for`-style like this: `for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i){}`

include/glow/Graph/Context.h
nickgg: The context is guaranteed to the be the single per-run object in the Runtime because it must be provided (by unique_ptr) to the DeviceManager while scheduling a new inference - CompiledFunction::execute also requires a Context*. Agree we should update the docs.

It's come up a few times from different folk, do you think there should be a different per run object?

opti-mix: Historically, `Context` was introduced just as a simple mapping from placeholders to tensors. Over time, once we introduced the `Runtime` components, we started (mis)using it as a de-facto per run object.

My personal opinion is that our current `Context` is too simple for a per-run object in terms of APIs and information it contains (e.g. a per run object could contain collected information about performance, devices being used, different kinds of counters, mode of execution, etc). So, either we extend `Context` into a per-run object and add more APIs and members as needed and rename it into e.g. a `RunContext` or `ExecutionInfo`; or we should create a new per-run object with all that information.

examples/tracing-compare.cpp
```diff
@@ -0,0 +1,178 @@
+/**
```
nadavrot: Why are we adding another utility for running resnet?  What if I want to profile german-to-english or AlexNet?

include/glow/Backends/Backend.h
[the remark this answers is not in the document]
nadavrot: instrumentAndCompile looks strange. What if we want to add another action? Are we going to have verb1andCompile(), verb2andCompile(), verb2andCompileWinthoutConstants() ?

This needs to be designed.

include/glow/Backends/CompiledFunction.h
```diff
@@ -60,11 +61,25 @@ class CompiledFunction {
   /// Collects constants for runtime.
   virtual void collectConstants(Module *){};
 
+  /// Setter for TraceEvent lookup. Note: does not enable tracing automatically.
```
nadavrot: Why is this a property of the compiled function?

nadavrot:
I like the approach of writing into the tensor at runtime. I think that we need to iterate over the interfaces a bit and clean things up. Also, maybe we should add the profiling ability to the model runner?

examples/tracing-compare.cpp
opti-mix: Overall, half of the general functionality in this file seems to be very similar to ModelRunner/Loader/ImageClassifier. Just recently I've written something similar for one of the backends and also had to duplicate a lot of code.   

Ideally, we should re-factor  ModelRunner/Loader/ImageClassifier and make this functionality re-usable. This way we could easily write small tests/tools like this without duplicating a lot of boiler-plate code.

I filed https://github.com/pytorch/glow/issues/2381 to track this.

include/glow/Backends/Backend.h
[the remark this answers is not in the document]
nickgg: So the other options are either adding a new compilation mode (e.g INFER_PROF) or adding a new argument to the compile() call (flags?). We spoke about it in the office and on the quip and couldn't come up with a consensus so I implemented the simplest. Very interested in opinions on this.

lib/Backends/Backend.cpp
nickgg: Oh. The intention was for to be essentially an array of instructions+1 elements of traceEventDataSize bytes but then I changed the type to int64 and didn't update the size, so this is 8x what we need.

include/glow/Backends/TraceEvents.h
nickgg: That was my first thought as well, but it turned out to be very inconvenient to have it a static method of the backend. You need the dataSize when interpreting the metrics tensor after run, but e.g. in the DeviceManager we don't have a backend since it's otherwise only used during compile. Static methods aren't particularly useful on the backends either since they aren't defined in the include path, so we'd have to add the backend path manually. 

These are fixable problems, of course, but I didn't want to bikeshed too heavily on this diff.

examples/tracing-compare.cpp
nickgg: Don't object to adding this to the image classifier tool at all, but I was making a distinction here between tools (for running and getting results) and examples (code-based documentation). 

Definitely like the idea of pulling out some common code here.

lib/Backends/CPU/libjit/libjit.cpp
nickgg: OK I can use gettimeofday here instead.

include/glow/Backends/CompiledFunction.h
nickgg: Because it seems to me that which Trace Events are included are a property of the function. TraceInfo is the static per-function part of the profiling data, not the metrics themselves but how to interpret them (e.g. the first timestamp is the start of the function, the second timestamp is the end of Conv1 and the start of RELU, etc etc).

Storing them outside the compiled function means in the Runtime we'd need to keep a map in the Executor (?) of run -> TraceInfo and do interpretation there. It would add dependencies on the backend to the Executor, but would reduce latency.

include/glow/Backends/Backend.h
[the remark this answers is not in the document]
nickgg: @jfix71 via assert you mean?

lib/Backends/CPU/libjit/libjit.cpp
nickgg: I'm coming back on this, gettimeofday is super bad (e.g. https://blog.habets.se/2010/09/gettimeofday-should-never-be-used-to-measure-time.html) and there isn't a portable alternative that's not c++.

How serious is using C++ in the JIT? This could get very complicated.

include/glow/Backends/Backend.h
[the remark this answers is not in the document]
jfix71: Yeah I think `GLOW_UNREACHABLE()` would make sense here

lib/Backends/CPU/libjit/libjit.cpp
opti-mix: @nickgg :-( Hmm, if there is no portable way other than C++, we could use `std::chrono` as you suggest, provided that all tests pass including the ones that use JIT compilation and AOT compilation. But add a comment before this `chrono` that we fallback to C++ here because there is no portable C solution.

If people then have issues with this because e.g. C++ symbols cannot be found at runtime, they should let us know and we may reconsider what to do here.

nickgg:
Fixed the easier review comments pending the ones that need more conversation. SHARED build will still be broken until #2390 lands and I can work out whats going on.

lib/Backends/CPU/libjit/libjit.cpp
nickgg: For now I'll add that comment and come back to it.

We can use clock_gettime on Linux and OSX >= 12.12, and fall back to something else on other platforms. gettimeofday might be acceptable there? I filed #2397 for this as a follow up.

nadavrot:
We had some offline conversation. I think that this PR is a good start. We know that we need to change a few things (unify tools, maybe iterate over the compile API). I am okay with landing this patch and continue to iterate in future PRs.

PR #2379, opened 2019-02-12
[LLVMIRGen] Use TargetMachine for sizeof(size_t).

petecoup:
*Description*:
[LLVMIRGen] Use TargetMachine for sizeof(size_t).  Start moving LLVMIRGen to be more cross-compiler friendly.
*Testing*:
ninja check
*Documentation*:
None

### DIFF
--- a/lib/LLVMIRCodeGen/LLVMIRGen.cpp
+++ b/lib/LLVMIRCodeGen/LLVMIRGen.cpp
@@ -93,6 +93,15 @@ static llvm::StringRef getHostCpuName() {
   return cpu_name;
 }
 
+/// Query the TargetMachine to get the pointer size in bits
+static unsigned getPointerNumBits(const llvm::TargetMachine &TM) {
+#if LLVM_VERSION_MAJOR >= 7
+  return TM.getPointerSize(0) * 8;
+#else
+  return TM.getPointerSize() * 8;
+#endif
+}
+
 LLVMIRGen::LLVMIRGen(const IRFunction *F, AllocationsInfo &allocationsInfo,
                      std::string mainEntryName, llvm::StringRef libjitBC)
     : F_(F), allocationsInfo_(allocationsInfo), mainEntryName_(mainEntryName),
@@ -132,12 +141,12 @@ void LLVMIRGen::loadBaseAddresses(llvm::IRBuilder<> &builder) {
   // Load the base addresses at the beginning of the entry function once they
   // are set. They won't change after this point and all relative addressing
   // computations will simply use them.
-  baseActivationsAddr_ = builder.CreatePtrToInt(F->args().begin() + 2,
-                                                llvm::Type::getInt64Ty(ctx_));
+  auto sizeTTy = builder.getIntNTy(getTargetSizeTWidth());
+  baseActivationsAddr_ = builder.CreatePtrToInt(F->args().begin() + 2, sizeTTy);
   baseConstantWeightVarsAddr_ =
-      builder.CreatePtrToInt(F->args().begin(), llvm::Type::getInt64Ty(ctx_));
-  baseMutableWeightVarsAddr_ = builder.CreatePtrToInt(
-      F->args().begin() + 1, llvm::Type::getInt64Ty(ctx_));
+      builder.CreatePtrToInt(F->args().begin(), sizeTTy);
+  baseMutableWeightVarsAddr_ =
+      builder.CreatePtrToInt(F->args().begin() + 1, sizeTTy);
   offsetsArray_ = F->args().begin() + 3;
 }
 
@@ -194,7 +203,7 @@ void LLVMIRGen::initCodeGen() {
 
   // Create the entry function into the LLVM module.
   auto int8PtrTy = llvm::Type::getInt8PtrTy(ctx_);
-  auto sizeTPtrTy = llvm::Type::getIntNPtrTy(ctx_, sizeof(size_t) * 8);
+  auto sizeTPtrTy = llvm::Type::getIntNPtrTy(ctx_, getTargetSizeTWidth());
   // The entry point has the following API:
   // void entry(uint8_t *baseConstantWeightVars, uint8_t
   // *baseInoutWeightVars, uint8_t *baseActivations, size_t *offsets);
@@ -305,7 +314,7 @@ llvm::Value *LLVMIRGen::emitValueAddress(llvm::IRBuilder<> &builder,
                                          const glow::Value *val) {
   assert(allocationsInfo_.allocatedAddress_.count(val) &&
          "Value address was not allocated");
-  auto sizeTTy = builder.getIntNTy(sizeof(size_t) * 8);
+  auto sizeTTy = builder.getIntNTy(getTargetSizeTWidth());
   llvm::Type *T = nullptr;
 
   switch (val->getElementType()) {
@@ -365,7 +374,7 @@ llvm::Value *LLVMIRGen::emitValueAddress(llvm::IRBuilder<> &builder,
 llvm::Value *
 LLVMIRGen::emitConstOffsetsArray(llvm::IRBuilder<> &builder,
                                  const AllocationsInfo &allocationsInfo) {
-  auto sizeTType = builder.getIntNTy(sizeof(size_t) * 8);
+  auto sizeTType = builder.getIntNTy(getTargetSizeTWidth());
   std::vector<llvm::Constant *> elems(allocationsInfo.valueNumbers_.size());
   for (auto &I : allocationsInfo.valueNumbers_) {
     auto *V = I.first;
@@ -395,7 +404,7 @@ template <typename T>
 llvm::Value *LLVMIRGen::emitConstSizeTArray(llvm::IRBuilder<> &builder,
                                             llvm::ArrayRef<T> vals) {
   assert(std::is_integral<T>() && "Can only convert integral type to size_t.");
-  auto SizeTType = builder.getIntNTy(sizeof(size_t) * 8);
+  auto SizeTType = builder.getIntNTy(getTargetSizeTWidth());
   std::vector<llvm::Constant *> elems;
   for (auto I : vals) {
     assert(I >= 0 && "Only allow casting positive values into size_t.");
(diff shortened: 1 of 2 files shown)


petecoup:
Oops, this looks to need LLVM >= 7, let me re-work this...

nadavrot:
I wonder if we should just move Glow to using LLVM7.  

@compnerd @bertmaher @opti-mix @rdzhabarov what do you all think?

lib/Backends/CPU/LLVMIRGen.cpp
```diff
@@ -305,7 +306,7 @@ llvm::Value *LLVMIRGen::emitValueAddress(llvm::IRBuilder<> &builder,
                                          const glow::Value *val) {
   assert(allocationsInfo_.allocatedAddress_.count(val) &&
          "Value address was not allocated");
-  auto sizeTTy = builder.getIntNTy(sizeof(size_t) * 8);
+  auto sizeTTy = builder.getIntNTy(TM_->getPointerSizeInBits(0));
```
opti-mix: @petecoup Could you please add a short explanation in the commit message and PR header why it is OK to take the number of bits in a pointer as a number of bits in `size_t`.

bertmaher:
I've been using LLVM7 locally for a long time, so it's fine by me :-).  I think fbinternal llvm is also more-or-less 7, but if there's any rework needed for that, I can take it on.

jfix71:
>I wonder if we should just move Glow to using LLVM7.

I know I was recently annoyed at having to use a `std::set` recently instead of `llvm::SmallSet` because LLVM6 does not support iterating over `llvm::SmallSet` 😞

opti-mix:
👍 for LLVM7

lib/Backends/CPU/LLVMIRGen.cpp
petecoup: Sure.  In short: when trying to leverage LLVMIRGen in a cross-compiler, we want to emit i32 (i64) types to interface with the corresponding IR types from the cross-compiled libjit for size_t types. This means that these certain types should be "getIntNTy(target sizeof(size_t))", rather than "getIntNTy(host sizeof(size_t))".  So, yes, this change is making the assumption the only supported LLVMIRGen targets have "sizeof(size_t) == sizeof(void*)".  Is that OK (or already the case implicitly)?  If so, I can add something like this in the commit message.  If not, then I'll need to do something else :).

opti-mix: The motivation is clear. We definitely want to use the target "getIntNTy(target sizeof(size_t))". 

And most likely we implicitly assume that "sizeof(size_t) == sizeof(void*)", but we do not clearly state it anywhere.

That being said, I'm not sure we want to hard-code this assumption in the code. May be we should add a helper method to `LLVMIRGen` to return the LLVM type representing the target `size_t` integer type? The default implementation could use what you propose, i.e. `builder.getIntNTy(TM_->getPointerSizeInBits(0));` But if this is not the case for a given target, one could easily override this new method to return whatever is appropriate.

opti-mix: BTW, we had a lot of headaches in some other backends (e.g. OpenCL) due to the use of size_t. That's why we banned it from the OpenCL analog of `libjit` and use there explicit fixed-width integer types to avoid any mismatches. May be we should do the same for `libjit` to avoid issues with `size_t` size mismatches between the host and cross-target?

petecoup: Sure, makes sense.  I can do something along those lines.

tlepley-cadence:
If there is a requirement for LLVM 7+, I think there is the need to change cmake files too, in order to make sure that the build will fail cleanly in case someone usres LLVM < 7. 

Not that currently, I'm using LLVM 7.1, and I get lots of warnings, because it seems that Glow searches for only 7.0 and 6.0. So more globally, I assume the LLVM handling with cmake may need some rework.
```
CMake Warning at CMakeLists.txt:135 (find_package):
  Could not find a configuration file for package "LLVM" that is compatible
  with requested version "7".

  The following configuration files were considered but not accepted:

    /home/tlepley/pub/lib/cmake/llvm/LLVMConfig.cmake, version: 7.1.0

CMake Warning at CMakeLists.txt:140 (find_package):
  Could not find a configuration file for package "LLVM" that is compatible
  with requested version "6".

  The following configuration files were considered but not accepted:

    /home/tlepley/pub/lib/cmake/llvm/LLVMConfig.cmake, version: 7.1.0

-- Found LLVM 7.1.0
```

nadavrot:
@tlepley I agree. We need to migrate to LLVM7 in a separate PR. We need to update cmake, docs, build scripts, etc.

nadavrot:
@petecoup May I land this PR?

petecoup:
Yes, please land if we've met the pre-requisites:  Part of this request was to make sure we (Synopsys) had fleshed out all of the CLA wiring, and such.  Do you have a signature from them on file now?
Also, if @opti-mix would like this changed to add a hook to over-ride the calculated "target sizeof(size_t)", I can do that before it lands.

opti-mix:
> Also, if @opti-mix would like this changed to add a hook to over-ride the calculated "target sizeof(size_t)", I can do that before it lands.

@petecoup If you could add this hook before we land, it would be nice.

compnerd:
@nadavrot LLVM 7 is okay to move to.  LLVM 8 is really just around the corner, and there often are so many fixes that can be beneficial, I think that it is reasonable to require newer LLVMs.

nadavrot:
This is a great change and I'd like to land this, but @petecoup needs to sign the CLA.

opti-mix:
> This is a great change and I'd like to land this, but @petecoup needs to sign the CLA.

Yes, it would be nice to finalize and land this PR. 

@petecoup: Could you provide an update on the current status? Have you signed the CLA in the meantime? Do you plan to land this PR?

petecoup:
@opti-mix: First off, let me apologize for my misunderstanding the internal status of this at the outset.  I do plan to have this land, and we should be ready with that early next week.
Sorry again for jumping the gun here.

opti-mix:
@petecoup 

> First off, let me apologize for my misunderstanding the internal status of this at the outset. I do plan to have this land, and we should be ready with that early next week.
Sorry again for jumping the gun here.

No problem. Thanks for the update! BTW, I re-factored `LLVMIRGen` in the meantime, so you'd need to rebase and update your PR.

> Also, if @opti-mix would like this changed to add a hook to over-ride the calculated "target sizeof(size_t)", I can do that before it lands.

I'd like to suggest something like this:
-  Add the following protected method to the LLVMIRGen, so that backends can override it if needed.
``` c++
  /// \returns bit-width of the target size_t.
  virtual unsigned getTargetSizeTWidth();
```
- Provide a default implementation of this method in `LLVMIRGen.cpp`, using the code in this PR:

```c++
/// Query the TargetMachine to get the pointer size in bits
static unsigned getPointerNumBits(const llvm::TargetMachine &TM) {
#if LLVM_VERSION_MAJOR >= 7
  return TM.getPointerSize(0) * 8;
#else
  return TM.getPointerSize() * 8;
#endif
}

unsigned LLVMIRGen::getTargetSizeTWidth() { return getPointerNumBits(*TM_); }
```

- Update all the places where `sizeof(size_t) * 8` is used by `getTargetSizeTWidth()`

lib/LLVMIRCodeGen/FunctionSpecializer.cpp
```diff
@@ -27,6 +26,7 @@
 #include "llvm/Transforms/Utils/Cloning.h"
 #include "llvm/Transforms/Utils/ValueMapper.h"
 
+#define DEBUG_TYPE "ir-function-specializer"
```
petecoup: I had to move this.  When I remerged, I got build errors for undefined DEBUG_TYPE when expanding the STATISTIC macro.  Not sure if this is an LLVM 7 vs. 6 thing, or something else...but I do know some LLVM headers #undef DEBUG_TYPE.

petecoup:
Sorry again for the delay here, this should all be ready now.
Does this look OK to you @opti-mix?

lib/LLVMIRCodeGen/FunctionSpecializer.cpp
opti-mix: This is very strange.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
```diff
@@ -545,8 +554,8 @@ llvm::Function *LLVMIRGen::getFunction(const std::string &name,
 /// the second BB is the loop exit BB.
 static std::pair<llvm::BasicBlock *, llvm::BasicBlock *>
 createLoop(llvm::IRBuilder<> &builder, llvm::LLVMContext &ctx,
-           llvm::Value *numElements) {
-  auto sizeTTy = builder.getIntNTy(sizeof(size_t) * 8);
+           llvm::Value *numElements, unsigned sizeTNumBits) {
```
opti-mix: This `sizeTNumBits` parameter looks a bit ugly as it has very little to do with the semantics of `createLoop`. May be we should just make `createLoop` a method of LLVMIRGen? Then `createLoop` can call `getTargetSizeTWidth` when it needs to. WDYT?

opti-mix:
@petecoup Thanks for addressing my comments. Overall, it looks very good now. Please consider my minor comment about `createLoop`. 

Could you also move the ` #define DEBUG_TYPE "ir-function-specializer" ` change into a separate commit and squash two commits related to `size_t`changes into one?

And then we should be good and can merge.

opti-mix:
@petecoup Seems like the `DEBUG_TYPE` issue is going to be solved by #2495 , so no need to handle it in your PR.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
petecoup: Sure, this makes sense to me.

petecoup:
@opti-mix, I re-based after the DEBUG_TYPE move commit and made createLoop an LLVMIRGen method.

Issue #2168, opened 2018-12-12
Perform slice(constant1) -> constant2 transformation at compile time


opti-mix:
I've discovered a pattern `slice(constant)` in the AlexNet. 

In principle, we could implement `slice(constant1)->constant2` as a compile-time transformation in the GraphOptimizer, to save computations at run-time.

My only concern here is that it may potentially increase the amount of memory required for constants.

Opinions?

jfix71:
Seems like a good idea.

>My only concern here is that it may potentially increase the amount of memory required for constants.

Yeah, this is another one that might be dependent on the backend's preferences, so would be great to have it be controlled by the backend (https://github.com/pytorch/glow/issues/1641).

This is also is related to constant propagation (https://github.com/pytorch/glow/issues/1469) -- I wonder if we could design this in a generic way, e.g. generate the Tensor to back the new Constant by calling into the kernels in the Interpreter.

tlepley-cadence:
I also think it makes sense to have it backend controlled as it is a tradeoff between performance and memory with a decision that can vary across targets. Also even a single target could have different memory configuration depending on the customer. In this case, the decision will depend on the memory constraints.

nadavrot:
I agree that in theory there could be a tradeoff here. But I am not sure if the pattern that exposes this tradeoff between memory and compute is a common pattern. What do you say that we start with a simple peephole optimization (instead of target-specific hook) instead?

SumedhArani:
@nadavrot Could you elaborate how peephole optimizations would solve this tradeoff issue? 

If I understand correctly, we would still be bound by memory constraints depending on the hardware.

I agree with @jfix71 on how constant folding is more relevant here and the broader solution. 

I would like to contribute to these issues. I am starting new here. Would you be willing to give me starting pointers?

nadavrot:
@SumedhArani I don't have a solution to the tradeoff issue. I suggest implementing one solution (just do the peephole) and give it a try. If we figure out that we do have memory issues in real workloads then we can iterate and improve things. There is a difference between what is a possible workload, and what is likely. Maybe in practice one optimization strategy is always superior? I suggest that we implement the simple peephole optimization and give it a try and evaluate how it performs on the workloads that we are testing today. What do you think?

SumedhArani:
@nadavrot A feasibility analysis sounds good to me. 

I will try implementing a simple peephole optimization. We can then decide the further roadmap for this depending on the results. 

[...]

PR #2382, opened 2019-02-13
[optimization] Slice(constant1)->constant2 optimization

SumedhArani:
*Description*: Compile time graph optimization. In reference with #2168 
*Testing*: Need to write graphs to test the optimization exhaustively

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1727,6 +1727,22 @@ static void optimizeSliceOfSplat(Function *F) {
   }
 }
 
+/// Replace SliceNode when the input is Constant with another constant node
+/// Slice(constant1) -> constant2
+static void optimizeSliceOfConstant(Function *F) {
+  for (auto &node : F->getNodes()) {
+    auto *sliceNode = dyn_cast<SliceNode>(&node);
+    if (!sliceNode)
+      continue;
+    auto *constantNode = dyn_cast<Constant>(sliceNode->getInput());
+    if (!constantNode)
+      continue;
+    auto *newConstantNode = F->getParent()->createConstant(
+          sliceNode->getResult().getType(), constantNode->getName());
+    sliceNode->getResult().replaceAllUsesOfWith(newConstantNode);
+  }
+}
+
 /// Optimize reshape nodes.
 static void optimizeReshape(Function *F) {
   for (auto &node : F->getNodes()) {
@@ -2426,6 +2442,9 @@ void glow::optimize(Function *F, CompilationMode mode) {
 
   // Optimize Tensor shape transformations.
   optimizeSliceOfSplat(F);
+
+  // Optimize constants
+  optimizeSliceOfConstant(F);
 
   // Merge Transpose into MatMul/FC.
   // Run DCE to ensure correct number of node users.


opti-mix:
@SumedhArani Thanks for working on this!

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1727,6 +1727,22 @@ static void optimizeSliceOfSplat(Function *F) {
...
+/// Slice(constant1) -> constant2
+static void optimizeSliceOfConstant(Function *F) {
+  for (auto &node : F->getNodes()) {
+    auto *sliceNode = dyn_cast<SliceNode>(&node);
+    if (!sliceNode)
+      continue;
+    auto *constantNode = dyn_cast<Constant>(sliceNode->getInput());
+    if (!constantNode)
+      continue;
+    auto *newConstantNode = F->getParent()->createConstant(
```
opti-mix: What happens with the payload of this new constant node? It needs to be backed by a tensor that represents the slice of the original constant, or?

But I don't see here how you initialize this payload.

jfix71:
@SumedhArani Looks like a good start! Can you add a test to `GraphOptzTest` that checks that this optimization occurs correctly?

beicy:
Please run `./utils/format.sh fix ` after your change. This command will handle the format issue :)

lib/Optimizer/GraphOptimizer.cpp
SumedhArani: I missed that. Thank you for pointing that out. I will make the necessary changes and update.

rdzhabarov:
@SumedhArani do you plan to work on this PR?

SumedhArani:
@rdzhabarov I am genuinely interested to get this work done at the earliest. However, I was bogged done with prior commitments. I should have an update for you on this by the end of this week. 

I really apologize for the delay.

SumedhArani:
I know I have still not made any changes. I really appreciate your patience. The internship search has consumed a lot of time for me. I'll have made the required changes by the end of this week, 10 March.

Thank you once again for your patience.

lib/Optimizer/GraphOptimizer.cpp
SumedhArani: I have spent the last couple of days reading existing code and through the documentation. 
I have been however stuck at the point of initializing the payload of the newConstantNode to be the sliced version of the original constant.
Could you give me pointers regarding finding the sliced payload of a tensor?

opti-mix: @SumedhArani `Constant` has a `getPayload` method returning a tensor with a payload of this constant. Now you need to extract a proper slice from it. And `Tensor` has a number of methods for manipulating its content, e.g. `copySlice`, `copyConsecutiveSlices`, etc. You should try to use those method to form a new Tensor representing the sliced version of the original constant. Once you have, you can use `createConstant` that takes a tensor as one of the parameters or you can create a `Constant` using type, shape, etc and then set its payload by means of calling the `Constant::assign` method.

SumedhArani: Thanks @opti-mix !

It's of much help. Thanks for taking the time to respond back to me! 👍

PR #2390, opened 2019-02-13
Fix runtime build with BUILD_SHARED_LIBS

nickgg:
*Description*: This has been broken locally for me for awhile but seems to build fine in circleci so I never worried about it, but that time has passed.
*Testing*: Build with the following config:
`
cmake -GNinja -DCMAKE_CXX_FLAGS=-Werror -DGLOW_WITH_CPU=ON -DCMAKE_BUILD_TYPE=Debug -DGLOW_WITH_OPENCL=ON -DBUILD_SHARED_LIBS=ON ../
`
*Documentation*: I am assuming that circleci passes because it uses an older version of the compiler? Can't quite work out exactly what the difference is.

### DIFF
--- a/lib/Runtime/Executor/CMakeLists.txt
+++ b/lib/Runtime/Executor/CMakeLists.txt
@@ -3,7 +3,7 @@ add_library(Executor
               ThreadPoolExecutor.cpp)
 
 target_link_libraries(Executor
-            PUBLIC
-            ThreadPool
+                      PUBLIC
+                        ThreadPool
 	              PRIVATE
 		        Graph)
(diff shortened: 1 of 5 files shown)


lib/Runtime/Executor/CMakeLists.txt
```diff
@@ -3,5 +3,7 @@ add_library(Executor
               ThreadPoolExecutor.cpp)
 
 target_link_libraries(Executor
+            PUBLIC
+            ThreadPool
```
jackm321: Why is the indentation different here?

PR #2393, opened 2019-02-13
Make HostManager own Devices

### LINKED ISSUE #2297: HostManager should wait until there are no active requests before clearingHost
Make the new hostManager stop accepting requests once clearHost has been called and block until active requests are 0 before clearing host.

### LINKED ISSUE #2384: HostManager should fully own devices
`DeviceManager`s are stored as shared_ptrs but they should be unique_ptrs because they are really owned by `HostManager`. https://github.com/pytorch/glow/blob/b9752128b9af67fb7c1232b2834ef6485b0aa070/include/glow/Runtime/RuntimeTypes.h#L35
I think the blocker here is that `HostManager::clearHost()` won't block on in flight requests #2297 so the ownership of `DeviceManager` isn't as strong as it should be. `HostManager` can be deleted but the underlying devices can still be left around running in-flight requests which is why they need to be shared pointers now. Once that is addressed, the ownership model should be improved and these can be safely made `unique_ptrs`.

jackm321:
*Description*:
Make it so HostManager owns DeviceManagers and the rest of the runtime just uses them. Make the blocking on shutdown in Executor explicit to ensure that no inflight requests will be using DeviceManagers. Prevent processing new requests while shutting down.

*Testing*:
`ninja all`
cd to downloaded_models and `../bin/resnet-runtime ../../tests/images/imagenet`

*Documentation*:
Fixes #2384 
Fixes #2297

### DIFF
--- a/lib/Runtime/Executor/ThreadPoolExecutor.cpp
+++ b/lib/Runtime/Executor/ThreadPoolExecutor.cpp
@@ -48,13 +48,11 @@ unsigned InflightBarrier::count() {
 
 void InflightBarrier::wait() {
   std::unique_lock<std::mutex> lock(mtx_);
-  if (count_ != 0) {
-    // If count_ is not 0, wait until a signal is received that it is.
-    // The second argument below is a predicate that returns true when
-    // it is safe to wake up. It preserves correctness in the case of
-    // spurious wakeups.
-    cv_.wait(lock, [&] { return count_ == 0; });
-  }
+  // If count_ is not 0, wait until a signal is received that it is.
+  // The second argument below is a predicate that returns true when
+  // it is safe to wake up. It preserves correctness in the case of
+  // spurious wakeups.
+  cv_.wait(lock, [&] { return count_ == 0; });
 }
 
 ExecutionState::ExecutionState(RunIdentifierTy id, const DAGNode *root,
@@ -204,7 +202,10 @@ void ExecutionState::setResultCode(const ResultCode resultCode) {
   }
 }
 
-ThreadPoolExecutor::~ThreadPoolExecutor() {
+void ThreadPoolExecutor::shutdown() {
+  // Prevent more requests from being processed.
+  shuttingDown_ = true;
+
   // Wait for all inflight DeviceManager::runFunction() calls to return and be
   // processed before starting to destroy state that is used in
   // handleDeviceManagerResult().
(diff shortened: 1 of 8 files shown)


jackm321:
Also it's necessary to block until requests are done because otherwise this would be a use-after-free https://github.com/pytorch/glow/blob/d8cbb67ae4f60f19eba2aaca43633f62d8f84821/lib/Runtime/HostManager/HostManager.cpp#L154

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -214,6 +215,12 @@ ThreadPoolExecutor::~ThreadPoolExecutor() {
 void ThreadPoolExecutor::run(const DAGNode *root,
                              std::unique_ptr<Context> context,
                              RunIdentifierTy runId, ResultCBTy cb) {
+  // Don't process new requests if the executor is shutting down.
+  if (shuttingDown_) {
+    cb(runId, ResultCode::Canceled, std::move(context));
```
gcatron: nit, should it be a ResultCode::Failed? It never really started.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -119,6 +118,12 @@ bool HostManager::networkAdded(llvm::StringRef networkName) {
 }
 
 void HostManager::clearHost() {
+  // shutdown the executor, blocking on any current inflight and prevent new
+  // requests from being serviced.
+  executor_->shutdown();
+  assert(activeRequestCount_ == 0 &&
+         "All requests should be finished when shutting downt HostManager.");
```
gcatron: downt -> down

lib/Runtime/Executor/ThreadPoolExecutor.cpp
jackm321: @gcatron I don't know what these mean, it would be good if they had some comments saying what each of them is. Canceled sounds more correct though just from guessing because there was nothing wrong with the request, it just got there too late.

jackm321: If you tell me what each of these are, I can add a comment in this PR to the ResultCode enum

gcatron: Yeah, they are a bit vague, canceled sounds good.

jackm321:
@gcatron thanks for the speedy review!

tests/unittests/ExecutorTest.cpp
```diff
@@ -786,9 +785,9 @@ TEST_F(ThreadPoolExecutorTest, MultiNode) {
   // the ThreadPoolExecutor has a reference to) and the TestDeviceManager map
   // (which the ExecutorTestBuilder has a reference to).
   auto deviceManager =
-      std::make_shared<TestDeviceManager>(deviceManagerThreads);
-  deviceManagerMap_.insert(std::make_pair(testDeviceId, deviceManager));
-  testDeviceManagerMap_.insert(std::make_pair(testDeviceId, deviceManager));
+      llvm::make_unique<TestDeviceManager>(deviceManagerThreads);
+  deviceManagerMap_.insert(
```
rdzhabarov: nit: deviceManagerMap_.emplace(testDeviceId, std::move(deviceManager)) and in other places.

jackm321: @rdzhabarov I didn't write that but yeah I agree it's better I'll change it

PR #2394, opened 2019-02-14
[unittests] Move DotProduct numerical tests to OperatorTest.cpp

ltamasi:
*Description*: Caffe2ImporterTest currently tests both that a DotProduct
operator can be loaded and that it produces correct results. With this change,
the numerical checks are moved to OperatorTest.
*Testing*: Unit tests pass.

### DIFF
--- a/tests/unittests/Caffe2ImporterTest.cpp
+++ b/tests/unittests/Caffe2ImporterTest.cpp
@@ -937,13 +937,16 @@ TEST(caffe2, dotProduct1D) {
   constexpr std::size_t kDataSize = 10;
   auto type = mod.uniqueType(ElemKind::FloatTy, {kDataSize});
 
-  // Destroy the loader after the graph is loaded, since it's not used in
-  // the remainder of the test.
+  // Destroy the loader after the graph is loaded to ensure the function F
+  // does not depend on anything stored in it.
   {
     Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
                                {type, type}, *F);
     output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
   }
+
+  // Check that the shape of the output matches that of the expected output.
+  EXPECT_TRUE(output->dims().equals({kDataSize}));
 
   // High level checks on the content of the graph.
   // We have 1 Mul and 1 Output.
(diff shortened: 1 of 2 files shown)


tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -932,40 +932,19 @@ TEST(caffe2, dotProduct1D) {
...
     Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
-                               {&X.getType(), &Y.getType()}, *F);
+                               {type, type}, *F);
     output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
-    ctx.allocate(mod.getPlaceholders());
-    updateInputPlaceholdersByName(ctx, &mod, {"X", "Y"}, {&X, &Y});
   }
 
-  // Check that the shape of the output matches that of the expected output.
-  EXPECT_TRUE(output->dims().vec() == OH.dims().vec());
```
rdzhabarov: this seems useful to check output type against expectations.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -1003,55 +970,27 @@ TEST(caffe2, dotProduct2D) {
...
-    for (size_t j = 0; j < kCols; ++j) {
-      dotProduct += (XH.at({i, j}) * YH.at({i, j}));
-    }
-
-    OH.at({i}) = dotProduct;
-  }
-
-  // Destroy the loader after the graph is loaded since the following execution
-  // will not depend on anyting from the loader.
+  // Destroy the loader after the graph is loaded, since it's not used in
```
rdzhabarov: note, the real reason for early destruction is that we want to make sure that function `F` does not depend on anything stored inside the `Caffe2ModelLoader`

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -1003,55 +970,27 @@ TEST(caffe2, dotProduct2D) {
...
     Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"X", "Y"},
-                               {&X.getType(), &Y.getType()}, *F);
+                               {type, type}, *F);
     output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
-    ctx.allocate(mod.getPlaceholders());
-    updateInputPlaceholdersByName(ctx, &mod, {"X", "Y"}, {&X, &Y});
   }
 
-  // Check that the shape of the output matches that of the expected output.
-  EXPECT_TRUE(output->dims().vec() == OH.dims().vec());
```
rdzhabarov: I'd also keep this to check output dims.

tests/unittests/OperatorTest.cpp
```diff
@@ -5063,6 +5063,94 @@ TEST_P(InterpOnly, Modulo2) {
   }
 }
 
+/// Test a DotProduct operator with 1D inputs.
```
rdzhabarov: it's not for this PR, but this test is huge right now, we might need to split it to speed up execution of tests.

tests/unittests/OperatorTest.cpp
```diff
@@ -5063,6 +5063,94 @@ TEST_P(InterpOnly, Modulo2) {
...
+TEST_P(InterpAndCPU, dotProduct1D) {
```
rdzhabarov: That should work on OpenCL backend as well it's just a MulNode. Any reason it's not included here?

rdzhabarov:
Thank you @ltamasi for contribution.

PR #2398, opened 2019-02-15
[DeviceManagerTest] Disable CPU-only test when not building with CPU

jfix71:
When not building with CPU this was causing the build to break.

Perhaps would be nice to unit test even just building with just the Interpreter backend.

I tried wrapping this in its own gtest CPUOnly case/class but couldn't get it to build successfully. If someone has a better suggestion/fix on this let me know!

### DIFF
--- a/tests/unittests/DeviceManagerTest.cpp
+++ b/tests/unittests/DeviceManagerTest.cpp
@@ -392,6 +392,8 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_NE(ctx1, ctx2);
 }
 
+#ifdef GLOW_WITH_CPU
+
 TEST_P(DeviceManagerTest, AvailableMemory) {
   std::vector<std::unique_ptr<CompiledFunction>> backing;
   std::promise<const Module *> promise;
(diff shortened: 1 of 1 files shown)


tests/unittests/DeviceManagerTest.cpp
```diff
@@ -392,6 +392,8 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_NE(ctx1, ctx2);
 }
 
+#ifdef GLOW_WITH_CPU
+
```
rdzhabarov: I think checking backendKind on line 400 and instantiating right device manager is slightly better fix.

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -392,6 +392,8 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_NE(ctx1, ctx2);
 }
 
+#ifdef GLOW_WITH_CPU
```
rdzhabarov: slightly better fix is to call `createDeviceManager` with the backend kind on line 400

rdzhabarov:
:-( that was the test which I thought I fixed last time, apparently that part is slipped.

tests/unittests/DeviceManagerTest.cpp
jfix71: Well there is no `DeviceManager::maxMemoryBytes_` -- it's only in the Interpreter and CPUBackend. And it needs to be set to 1. So I don't think that would work.

Also the way the test is currently formulated I'm surprised passes. I think there needs to be better error handling here. It looks like we're explicitly creating a `CPUDeviceManager`, but then calling `addNetwork()` on it with a `InterpreterFunction`, and yet the test passes...?

rdzhabarov: If all device managers expect to have maxMemoryBytes supplied as part of the init we should extend createDeviceManager call with that param IMO.

SplitInfinity:
Thanks for doing this! I thought it was just me who was experiencing this.

PR #2399, opened 2019-02-15
[NFC] Apply modernize-use-override

bertmaher:
Use C++11’s override and remove virtual where applicable.


nadavrot:
Nice! Out of curiosity, did you use clang-tidy or some other tool?

bertmaher:
Heh, actually this was an internal diff from someone working in fbcode, that I offered to export for them :-).  I believe they did use an automated tool but I'm not sure which one.

PR #2401, opened 2019-02-15
[Runtime] Added OpenCL device manager

gcatron:
*Description*: Added OpenCLDeviceManager. Currently this does not support a compiled function on more than one device. That can be updated after executionInfo is added.
*Testing*: Added OCL deviceManager to deviceManager unit test. Passes. Also tested on resnet50 and it passes.
*Documentation*: N/A

### DIFF
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -1,0 +1,200 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#define DEBUG_TYPE "opencl"
+
+// Silence Apple's warning about the deprecation of OpenCL.
+#define CL_SILENCE_DEPRECATION
+
+// Silence warnings about using deprecated OpenCL 1.2 functions.
+#define CL_USE_DEPRECATED_OPENCL_1_2_APIS
+
+#include "OpenCLDeviceManager.h"
+
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace glow;
+using namespace glow::runtime;
+
+namespace glow {
+namespace runtime {
+DeviceManager *createOCLDeviceManager(llvm::StringRef name) {
+  return new OpenCLDeviceManager(name);
+}
+} // namespace runtime
+} // namespace glow
+
+cl_mem OpenCLDeviceManager::allocDeviceBuffer(uint64_t size) {
+  const uint64_t alignment = 128;
+  // Always allocate buffers properly aligned to hold values of any type.
+  size = alignedSize(size, alignment);
+  auto buf =
+      clCreateBuffer(context_, CL_MEM_READ_WRITE, size, nullptr, nullptr);
+  GLOW_ASSERT(buf && "Allocation failed!");
+  return buf;
+}
+OpenCLDeviceManager::OpenCLDeviceManager(llvm::StringRef name)
+    : QueueBackedDeviceManager(BackendKind::OpenCL, name) {
+  // For now we have a string, if the first digit is an int use it, otherwise
+  // use 0 as the default.
+  // There are flags in the OpenCLBackend to select devices. Once we refactor
+  // more functionality out of the OCLCompiledFunction we can move those flags
+  // here.
+  auto deviceId{0};
+  if (llvm::isDigit(name[0])) {
+    deviceId = std::stoi(name.substr(0, 1));
+  }
+  cl_uint numPlatforms{0};
+  cl_int err = clGetPlatformIDs(0, NULL, &numPlatforms);
+  GLOW_ASSERT(err == CL_SUCCESS && "clGetPlatformIDs Failed.");
+  std::vector<cl_platform_id> platform_ids(numPlatforms);
+  err = clGetPlatformIDs(numPlatforms, platform_ids.data(), NULL);
+  // Take the first platform.
+  cl_platform_id platform_id_used = platform_ids[0];
+  cl_uint num{0};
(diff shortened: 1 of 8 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  assert(usedMemoryBytes_ <= maxMemoryBytes_);
+
+  // Fire the ready CB.
+  readyCB(module, ResultCode::Ready);
+}
+
+void OpenCLDeviceManager::evictNetworkImpl(std::string functionName) {
+  if (functions_.erase(functionName)) {
+
```
opti-mix: Remove the empty line.

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -0,0 +1,109 @@
...
+namespace runtime {
+
+/// A class that contains an openCL device buffer. It frees the buffer when it
+/// is destroyed. Can be extended to store multiple buffers and rotate through
+/// them. Also tracks number of functions using this buffer.
+class OpenCLBuffer {
+  /// The OpenCL buffer being stored.
+  cl_mem buffer_;
+  /// Count of functions using this buffer.
+  unsigned int users_{0};
```
opti-mix: Should the operations on `users_` be atomic to avoid any possible race conditions?

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -465,3 +465,8 @@ INSTANTIATE_TEST_CASE_P(Interpreter, DeviceManagerTest,
 INSTANTIATE_TEST_CASE_P(CPU, DeviceManagerTest,
                         ::testing::Values(BackendKind::CPU));
 #endif // GLOW_WITH_CPU
+
+#ifdef GLOW_WITH_OPENCL
+INSTANTIATE_TEST_CASE_P(OpenCL, DeviceManagerTest,
+                        ::testing::Values(BackendKind::OpenCL));
+#endif
```
opti-mix: New line is missing.

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -0,0 +1,109 @@
...
+/// added, but only one inference is executed at a time.
+class OpenCLDeviceManager : public QueueBackedDeviceManager {
+  /// Compiled function list by name.
+  FunctionMapTy functions_;
+
+  /// Maximum available memory on the device.
+  uint64_t maxMemoryBytes_{0};
+
+  /// Amount of memory used by all models.
+  uint64_t usedMemoryBytes_{0};
```
opti-mix: Should the operations on `usedMemoryBytes_` be atomic to avoid any possible race conditions?

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // Always allocate buffers properly aligned to hold values of any type.
+  size = alignedSize(size, alignment);
+  auto buf =
+      clCreateBuffer(context_, CL_MEM_READ_WRITE, size, nullptr, nullptr);
+  GLOW_ASSERT(buf && "Allocation failed!");
+  return buf;
+}
+OpenCLDeviceManager::OpenCLDeviceManager(llvm::StringRef name)
+    : QueueBackedDeviceManager(BackendKind::OpenCL, name) {
+  // For now we have a string, if the first digit is an int use it, otherwise
```
opti-mix: We currently have some logic in the OpenCL.cpp, in `OpenCLFunction::OpenCLFunction` to select a proper platform and device. It can be configured via some command-line options. This code seems very similar. Should we add similar options here or may be reuse the existing ones?

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -0,0 +1,109 @@
...
+
+  /// CL compute device id.
+  cl_device_id deviceId_;
+  /// CL compute context.
+  cl_context context_;
+  /// CL compute command queue.
+  cl_command_queue commands_;
```
opti-mix: I guess you plan to remove the corresponding class members from `OpenCLFunction`, right? Do you plan to do it in a subsequent PR?

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -0,0 +1,197 @@
...
+    bundle.freeConstants();
+    readyCB(module, ResultCode::Failed);
+    return;
+  }
+
+  // Copy constants to device.
+  auto size = bundle.getConstantWeightSize() + bundle.getMutableWeightSize() +
+              bundle.getActivationsSize();
+  auto deviceBuffer = allocDeviceBuffer(size);
+  auto buffer = std::make_shared<OpenCLBuffer>(deviceBuffer, size);
```
opti-mix: We allocate the buffer rather early here, before the actual execution. And this reduces the amount of bytes available for the next network. If we add multiple networks this way, we may end-up with no memory available, even though none of them is being executed... Is it a concern? May be we should allocate on first use inside `runFunctionImpl`?

lib/Backends/OpenCL/OpenCLDeviceManager.h
gcatron: Since the OpenCLDeviceManager inherits from queuebackedDeviceManager, all calls are serialized so there shouldn't be any race conditions.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
gcatron: Once we have executionInfo there will be a migration of functionality from the compiledFunction to the deviceManager, we can move the options over then.

lib/Backends/OpenCL/OpenCLDeviceManager.h
gcatron: Yup!

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
gcatron: We do this so we can copy the constants over to the device and release to Module. The plan is for this to be similar to the general device use case. It is definitely possible to run out of memory though. Hopefully the DeviceManager returns a failure to add network for OOM situations.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+void OpenCLDeviceManager::evictNetworkImpl(std::string functionName) {
+  if (functions_.erase(functionName)) {
+    auto buffer = buffers_[functionName];
+    auto users = buffer->decrementUsers();
+    auto size = buffer->getSize();
+    buffers_.erase(functionName);
+    if (users == 0) {
+      usedMemoryBytes_ -= size;
```
nickgg: Can we assert that usedMemoryBytes is >= size before subtracting here?

PR #2414, opened 2019-02-20
Check type of CompiledFunction in DeviceManager before adding

### LINKED ISSUE #2410: [DeviceManager] Enforce correct type of CompiledFunction for addNetwork
While creating https://github.com/pytorch/glow/pull/2398, I realized that it seems the test currently is creating a `CPUDeviceManager`, but then calling `addNetwork()` on it with a `InterpreterFunction`. This seems like it shouldn't be possible or at least should fail somehow.

@rdzhabarov also mentioned that if all device managers are expected to have `maxMemoryBytes_` supplied as part of the init then perhaps we should extend the `createDeviceManager()` call with it. If we do expect all device managers to have it then it should probably also be moved to `DeviceManager`?

CC: @nickgg

nickgg:
*Description*: Adds a `getCompileBackendKind()`method to CompiledFunction which returns the kind of backend used to create it. In the DeviceManager we can use this to prevent CompiledFunctions of the wrong type from being added.
*Testing*: Ran unit test, found and fixed a bad DeviceManagerTest.
*Documentation*: fixes #2410 

Worth noting this adds a new pure virtual method to CompiledFunction, so new Backends will need to fill this out in their CompiledFunction-subclasses.

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "CPUDeviceManager.h"
+#include "CPUFunction.h"
 
 #include "llvm/Support/raw_ostream.h"
 
@@ -49,6 +50,12 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
                    << func.first << ".\n";
       readyCB(module, ResultCode::Failed);
       return;
+    }
+
+    if (func.second->getCompileBackendKind() != BackendKind::CPU) {
+      llvm::errs() << "Failed to add network: function " << func.first
+                   << " is not a CPUFunction.\n";
+      readyCB(module, ResultCode::Failed);
     }
   }
 
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -22,6 +22,7 @@
 #define CL_USE_DEPRECATED_OPENCL_1_2_APIS
 
 #include "OpenCLDeviceManager.h"
+#include "OpenCL.h"
 
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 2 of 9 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -114,6 +115,12 @@ void OpenCLDeviceManager::addNetworkImpl(const Module *module,
       readyCB(module, ResultCode::Failed);
       return;
     }
+
+    if (func.second->getCompileBackendKind() != BackendKind::OpenCL) {
+      llvm::errs() << "Failed to add network: function " << func.first
+                   << " is not an Function.\n";
```
beicy: Just want to double check: the error message means "not an function" or "not a correct backend function" ?

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -50,6 +51,12 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
       readyCB(module, ResultCode::Failed);
       return;
     }
+
+    if (func.second->getCompileBackendKind() != BackendKind::CPU) {
```
opti-mix: Minor nit: This check is almost identical for all device managers. May be it can be moved into a utility method or into a base class method?

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
nickgg: This should say "not an OpenCL function" but looks like i deleted the important word, thanks!

lib/Backends/CPU/CPUDeviceManager.cpp
nickgg: I could add a helper method `bool checkBackendKind(CompiledFunction*)` but it would replace just the check unless we wanted to pass in all the local state (cb etc). It would end up adding code not removing code, do you think it's clearer?

opti-mix: OK. In this case let's leave it as is.

Issue #1581, opened 2018-09-04
Currently using floats for Select, CmpEQ, CmpLTE


jfix71:
Right now we are using `FloatTy` for Variables that are used as the `Condition` input to `Select`, when it really should be some sort of boolean/predicate type. Similarly our Interpreter backend creates outputs from `CmpEQ`/`CmpLTE` as float. This likely needs to be redesigned with some sort of boolean or predicate ElemKind.

One option could be to do this with a generic (no width-specified) `ElemKind::PredicateTy`. The idea here is to allow the backend to implement it with whatever bit width it wants. One issue is that Variables have their ElemKind specified independently of a backend/prior to picking or setting the backend. So we would need to allocate the tensors backing these Variables with some specific size prior to knowing what size the backend we are running with would like to use. The backend would need to convert them to the size the backend prefers.

Another option could be to add something like `ElemKind::Predicate8Ty`, where the bit width is explicit. The width would stay static and we could allocate tensors backing Variables with this width as such. Backends could still change the size of the predicate if they wanted once the Instruction level IR reaches it, i.e. in `Backend::compile()`.

Either way it does seem we will need the backend to massage the type to fit its internal representation. So in some sense these are different flavors of the same thing; e.g. if we used 8 as the default width for for the `ElemKind::PredicateTy` option it seems almost identical to just using `ElemKind::Predicate8Ty`.

qcolombet:
I vote for using PredicateTy backed by a 8-bit width type for now.
Having the backend specify the size/internal layout of the tensor is something we would need in general in the future and this clean-up would fall in that category. (E.g., int32_t could be represented differently on different targets and we would need something to massage the bits to match what the target wants).

In other words, I don't thing we are ready nor it would get us anything for now to do something different here.

PR #2417, opened 2019-02-20
Add in new ElemKind::BoolTy for use with Compare, Select instructions

jfix71:
*Description*: Add in a new `ElemKind::BoolTy`. This removes the hacky implementation we had before for which just used the same `ElemKind` for the bool result/input. Note that I went with `BoolTy` instead of `PredicateTy`, but I can switch it if preferred.

While this is implemented with `bool` in the Interpreter, I used `int8_t` in the CPU/OpenCL backends.

*Testing*: Updated all impacted tests, and added a simple test for testing CmpLTE by itself.

Closes https://github.com/pytorch/glow/issues/1581

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -264,6 +264,7 @@ enum class ElemKind : unsigned char {
   Int32ITy,     // 32-bit index type (int32_t)
   Int64ITy,     // 64-bit index type (int64_t)
   Int8FusedQTy, // 8-bit quantized type with fused scale/offset (int8_t)
+  BoolTy,       // Bool type (bool)
 };
 
 /// A class that represents a type of a tensor.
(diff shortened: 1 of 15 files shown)


include/glow/Base/Type.h
```diff
@@ -490,7 +495,8 @@ struct Type final {
   /// \return the textual name of the element \p Ty.
   static llvm::StringRef getElementName(ElemKind Ty) {
     static const char *names[] = {
-        "float", "float16", "i8", "i16", "i32", "index32", "index64", "i8fused",
+        "float",   "float16", "i8",      "i16",  "i32",
```
jackm321: Is the spacing weird on this line? It looks like a bunch of extra spaces to me but I can't tell if that's just how it's showing up in gh

jfix71: The extra spaces are actually there -- `clang-format` aligns each element with the one below it, so it's aligning `"i16"` with `"bool"`

PR #2419, opened 2019-02-21
DCE script for caffe2 nets

jackm321:
*Description*:
Basic script that takes caffe2 predict_net and init_net protobuf files and removes all dead nodes (nodes that aren't used to calculate the external_output of the predict_net then outputs this back to protobuf files.

The motivation behind this script is that sometimes we get models that have a bunch of extra nodes in them that we don't need for computing the prediction output and some of them we don't support in Glow (for example the [Python op](https://caffe2.ai/docs/operators-catalogue.html#python)) so this can clean out these nodes so that we can then import them using the model loader.

*Testing*:
Added a test init_net and predict_net and manually checked the output of running the script on those files.
Ran the script on a large graph with lots of dead nodes and eyeballed the output in a graph visualizing tool.

*Documentation*:
Comments in code

### DIFF
--- a/tests/models/caffe2Models/dce_test_init_net.pbtxt
+++ b/tests/models/caffe2Models/dce_test_init_net.pbtxt
@@ -1,0 +1,74 @@
+name: "init"
+op {
+  output: "conv_w"
+  type: "GivenTensorFill"
+  arg {
+    name: "shape"
+    ints: 1
+    ints: 1
+    ints: 2
+    ints: 2
+  }
+  arg {
+    name: "values"
+    floats: 1.0
+    floats: 1.0
+    floats: 1.0
+    floats: 1.0
+  }
+}
+op {
+  output: "conv_b"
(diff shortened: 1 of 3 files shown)


beicy:
Thanks a lot! Btw, for the inin_net.pbtxt file, is that OK to add one more test : two tensors with the same name and type?   If this happens, just report an error/warning.  In our model loader, we assumes that the tensors have a unique name. However, I've met an case before that there were duplicated tensors (same name, but different values) which caused an issue in our side. (I think this should be viewed as some kind of dead code? )

jackm321:
@beicy that's an interesting case. In this case, I think our loader should be throwing an error because how can we know which is the correct value. The same would apply to this script, it's not clear how we would resolve which is the correct value and which should be eliminated. Also just to clarify, this script is not meant to do general error checking on c2 graphs, it's just to remove stuff that's not needed to compute the outputs of the predict_net

Issue #2395, opened 2019-02-14
Quantizing after lowering causes issues with FullyConnected


tlepley-cadence:
@jfix71 there are some side effects to your changes in #2311.  Some operations have specific quantized types. This is for instance the case for FullyConnected that has its bias param quantized as Int32QTy.

The problem is that this FullyConnected nodes gets lowered as MatMul and BatchAdd. Since the actual IR quantization now happens after Lowering,  the bias is now quantized as Int8QTy with BatchAdd, not anymore as Int32QTy.

I causes some issues with our backend.

jfix71:
@tlepley-cadence Thanks for raising this issue!

It seems the `FunctionQuantizer` needs to know each nodes' origin node that it was lowered from. After talking with @rdzhabarov, we think the best solution for now is to extend the `LoweredNamesMap` to include the origin node's `NodeKind` instead of just its output name. This way the `FunctionQuantizer` can check if e.g. a `BatchedAddNode` was lowered from a `FullyConnectedNode` in order to handle it correctly and use `Int32QTy` for the bias.

In the long run it may make sense to change the `LoweredNamesMap` to instead map between `NodeValues` instead of `string` -> `string` + `NodeKind`. This could be useful for other purposes, e.g. debugging to track where nodes came from, or to use during performance profiling. However one downside here is that the `NodeValue` includes a pointer to a `Node`, but the `Node` would already have been deleted by DCE by the time it's dereferenced. So we would need to modify DCE in this case to remove the node from the Function, but still keep it around in memory until it's no longer needed.

PR #2422, opened 2019-02-21
Refactor LoweredNamesMap to LoweredInfoMap and correctly quantized BAs lowered from FCs

jfix71:
*Description*: Refactor LoweredNamesMap to LoweredInfoMap. The LoweredInfoMap maps to a set of NodeNameAndKind structs containing the node's outputName and NodeKind. This is used to correctly quantized BAs lowered from FCs, so that during quantization we know if a batched add came from a fully connected. This also requires keeping track of the LoweredInfoMap prior to quantization, whereas before it was only prior to profiling.

*Testing*: Added to unit tests to verify the batched add is Int32QTy, and correct scale/offset.

Fixes https://github.com/pytorch/glow/issues/2395

### DIFF
--- a/include/glow/Optimizer/Optimizer.h
+++ b/include/glow/Optimizer/Optimizer.h
@@ -43,12 +43,18 @@ void optimize(Function *F, CompilationMode mode);
 
 /// Lower the high-level neural network nodes found in \p F into low-level
 /// linear algebra operators. \p B can prevent lowering of a node via \ref
-/// Backend::shouldLower(). If \p loweredMap is not a nullptr, then everything
-/// is lowered regardless of the preferences of \p B, and \p loweredMap will
-/// contain a mapping from output names of the nodes found and lowered in \p F
-/// to the output names of the nodes they were lowered from along with the
-/// NodeKind.
+/// Backend::shouldLower(). If \p loweredMap is not a nullptr, then \p
+/// loweredMap will contain a mapping from output names of the nodes found and
+/// lowered in \p F to the output names of the nodes they were lowered from
+/// along with the NodeKind.
 void lower(Function *F, const Backend &B, LoweredInfoMap *loweredMap = nullptr);
+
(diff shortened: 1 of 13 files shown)


rdzhabarov:
cc: @tlepley

include/glow/Optimizer/Optimizer.h
```diff
@@ -43,12 +43,18 @@ void optimize(Function *F, CompilationMode mode);
...
-/// is lowered regardless of the preferences of \p B, and \p loweredMap will
-/// contain a mapping from output names of the nodes found and lowered in \p F
-/// to the output names of the nodes they were lowered from from.
-void lower(Function *F, const Backend &B,
-           LoweredNamesMap *loweredMap = nullptr);
+/// Backend::shouldLower(). If \p loweredMap is not a nullptr, then \p
+/// loweredMap will contain a mapping from output names of the nodes found and
+/// lowered in \p F to the output names of the nodes they were lowered from
+/// along with the NodeKind.
+void lower(Function *F, const Backend &B, LoweredInfoMap *loweredMap = nullptr);
```
rdzhabarov: Do we need explicit `lowerEverything`? Could we have `Backend *` here optional and remove `lowerEverything` entirely?

jfix71: I removed `lowerEverything()`. I was thinking it would be annoying to have many places where we explicitly add a `nullptr` but I don't think it's too bad actually. LMK what you think.

include/glow/Quantization/Quantization.h
```diff
@@ -56,6 +50,32 @@ struct NodeQuantizationInfo {
   }
 };
 
+/// Struct containing the output name string and node kind for use in the
+/// LoweredInfoMap for keeping track of lowered node info.
+struct NodeNameAndKind : public Named, public Kinded {
+public:
+  NodeNameAndKind(NodeValue NV)
```
rdzhabarov: minor to pass by const ref here.

PR #2424, opened 2019-02-21
[Easy] Remove unnecessary Expected<bool> in loaders

jackm321:
*Description*:
Return an error instead of false because false indicates a common operator could not be loaded.

*Testing*:
`ninja test`

*Documentation*:
none added


jackm321:
@rdzhabarov I don't know. Are you talking about the node tests?

PR #2426, opened 2019-02-21
Fix onnxifi node tests

jackm321:
*Description*:
Make sure that onnxifi node tests only run on onnxifi backends that support onnx graphs.
This is accomplished by reducing the number of onnxifi backends being created to one interpreter backend plus another optional backend for CPU.

*Testing*:
`./tests/onnxifi/build_onnxifi_tests.sh && ./tests/onnxifi/test.sh`

*Documentation*:
Comments in code


jackm321:
Hmm yeah that's a good point. Talked briefly with @yinghai about this, I think we'll most likely go with changing the test framework to check that each backend supports onnx before trying to run onnx node tests on it instead of this.

Issue #2413, opened 2019-02-20
Quantizing/profiling after lowering and DepthWise convolution


ayermolo:
After #2311 now everything is lowered always. This becomes and issue with MobileNet which has a few depthwise convolutions. In ONNX/GLOW they are represented as regular convolutions with groupSize > 1. In at least one case group size is 1024.

In Lower the lowerGroupConvolutionNode breaks these convolutions in to multiple convolutions with group size 1. This leads to explosion of nodes.
This leads to
1) Drastic increase in compile time
2) Failure in IRGen in uniqueName because it fails to find unique name since it's limited by arbitrary 10k limit.

Is lowerGroupConvolutionNode needed? Seems like CPU backend supports groupSize > 1.

rdzhabarov:
Related issue: https://github.com/pytorch/glow/issues/2395.

We should explore supplying nodes which must not be lowered as part of the Loader. Currently it's up to backend whether `shouldLower` or not, but in this case it should be controlled at a higher level.

jfix71:
>Is lowerGroupConvolutionNode needed? Seems like CPU backend supports groupSize > 1.

It's currently not needed for any of our open source backends, however other backends might not support group convolutions. Perhaps we could remove the call into `lowerGroupConvolutionNode()` from `lower()`, and allow a backend to call it during `transformPostLowering()`. However this means we would not be able to get fine grained profiles for each convolution of the group, though this was already the case prior to https://github.com/pytorch/glow/pull/2311 anyway.

The benefit of @rdzhabarov's suggestion is that we could still allow the profile to be generated for each convolution of a group if desired. I'd prefer that here -- it'd be a simple CLI flag to specify a list of Node kinds, which would then be passed to `lower()`.

ayermolo:
Yeah seems like adding control that's not specifically tied to any backend implementation as @rdzhabarov proposed would be a good idea. This way profiling can still be done with generic CPU implementation, but what is lowered and what isn't would closer reflect whatever backend would use the profile.

PR #2427, opened 2019-02-22
[Lowering/Loader] Add CLI option and plumbing to disable lowering specified node kinds.

jfix71:
*Description*: Add option on the command line for loaders to prevent lowering during profiling. By default everything will be lowered for profiling. However this may cause performance issues for some models, e.g. if a model has group Convolutions which explode the size of the graph when lowered. Thus, this PR allows for disabling certain NodeKinds for profiling. This means that during quantization, these nodes should also not be lowered by the backend.

*Testing*: Added test. Also verified shufflenet works with the new command line option while making profiling much faster (what is was previously).

*Documentation*: Added documentation.

Fixes https://github.com/pytorch/glow/issues/2413

### DIFF
--- a/docs/Quantization.md
+++ b/docs/Quantization.md
@@ -85,6 +85,24 @@ For example, you can run the following command to capture profile for Resnet50.
 ```
 ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -dump-profile="profile.yaml"
 ```
+
+By default, everything will be lowered for profiling. This allows for the
+lowered components of nodes to be profiled, allowing for good precision of
+complex nodes. For example, the `SigmoidCrossEntropyWithLogitsNode` has many
+internal nodes that it is lowered to. Without profiling the internal nodes,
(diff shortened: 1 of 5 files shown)


jfix71:
CC: @ayermolo

ayermolo:
Awesome!

docs/Quantization.md
```diff
@@ -85,6 +85,23 @@ For example, you can run the following command to capture profile for Resnet50.
...
+internal nodes that it is lowered to. Without profiling the internal nodes,
+there would be no information on how best to quantize its internal nodes that it
+is lowered to.
+
+Lowering all nodes may cause performance issues for some models, e.g. if a model
+has group Convolutions which explode the size of the graph when lowered, leading
+to long compilation and run time during profiling. Thus, we allow for disabling
+certain NodeKinds for profiling. This means that during quantization, these
+nodes should also not be lowered by the backend. This can be done using the
+command line option `-do-not-lower-nodes-for-profiling`; for example:
```
rdzhabarov: can you also specify that multiple nodes should be comma separated

PR #2428, opened 2019-02-22
Avoid large stack allocation in libjit matmul

nickgg:
*Description*: As a performance optimization the libjit matmul has a branch where if a tensor dimension is >= 1024 it allocates about 2 mb on the stack and does the multiplication in that memory. 

Some systems (looking at you OSX) limit the stack size of created threads such that any graph with a large Tensor would crash if the compiled CPUFunction was run on the DeviceManager or on any new thread.

The fix is to allocate the scratchpad dynamically.
*Testing*: unit tests w/ ASAN.

Also, GemmBench to compare performance regression (3 runs each impl):

```
Alloc on Stack:
utX, outY, lhsX, lhsY, rhsX, rhsY, gflops/s,
1024, 1024,   1024, 1024,   1024,  1024,   108.48
1024, 1024,   1024, 1024,   1024,  1024,   109.21
1024, 1024,   1024, 1024,   1024,  1024,   116.51

  32, 32  ,     32, 1024,   1024,  32  ,   68.75
  32, 32  ,     32, 1024,   1024,  32  ,   74.80
  32, 32  ,     32, 1024,   1024,  32  ,   87.44

Alloc on Heap:
1024, 1024,   1024, 1024,   1024,  1024,   112.94
1024, 1024,   1024, 1024,   1024,  1024,   115.64
1024, 1024,   1024, 1024,   1024,  1024,   118.04

  32, 32  ,     32, 1024,   1024,  32  ,   74.69
  32, 32  ,     32, 1024,   1024,  32  ,   77.70
  32, 32  ,     32, 1024,   1024,  32  ,   63.44
```

Hard to see a clear drop off, but could just be 3 runs isn't enough.
*Documentation*: This was a fun one to find.

### DIFF
--- a/lib/Backends/CPU/libjit/libjit_defs.h
+++ b/lib/Backends/CPU/libjit/libjit_defs.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_BACKENDS_CPU_LIBJIT_LIBJIT_DEFS_H
 #define GLOW_BACKENDS_CPU_LIBJIT_LIBJIT_DEFS_H
 
+#include <cstdlib>
 #include <stdint.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


nadavrot:
@tlepley FYI

pshung:
The calling of posix_memalign is only for Linux toolchain. what if users use the newlib toolchain which lacks the API?

nickgg:
@pshung We could switch on an ifdef in libjit_defs for that situation. I don't have a local repro but I'd be very happy to review a PR. (hint hint!)

Issue #1255, opened 2018-07-11
OpenCL backend leaks memory


bertmaher:
I ran a build with LeakSanitizer enabled and it pointed to a few leaks.

```
cmake -DGLOW_USE_SANITIZER=Address ~/src/glow
export ASAN_OPTIONS=detect_leaks=1
ninja graphTest
./tests/graphTest
```

The leaks are of allocations inside OpenCL, but we should make sure we're not misusing the API.

e.g.,
```
Indirect leak of 36 byte(s) in 1 object(s) allocated from:
    #0 0x103024c33 in wrap_malloc (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x56c33)
    #1 0x7fff7e6f38b2 in _Block_copy (libsystem_blocks.dylib:x86_64+0x8b2)
    #2 0x7fff7e6f3b3d in _Block_object_assign (libsystem_blocks.dylib:x86_64+0xb3d)
    #3 0x7fff7e6f38ef in _Block_copy (libsystem_blocks.dylib:x86_64+0x8ef)
    #4 0x7fff7e66bfd3 in _dispatch_Block_copy (libdispatch.dylib:x86_64+0x1fd3)
    #5 0x7fff7e683a07 in _dispatch_source_set_cancel_handler (libdispatch.dylib:x86_64+0x19a07)
    #6 0x103023f9e in wrap_dispatch_source_set_cancel_handler (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x55f9e)
    #7 0x7fff7867db1a in (anonymous namespace)::RunElsewhere::instance() (SkyLight:x86_64+0x23b1a)
    #8 0x7fff787ba521 in __SLSInitialize_block_invoke (SkyLight:x86_64+0x160521)
    #9 0x7fff7e66bd4f in _dispatch_client_callout (libdispatch.dylib:x86_64+0x1d4f)
    #10 0x7fff7e66bd02 in dispatch_once_f (libdispatch.dylib:x86_64+0x1d02)
    #11 0x7fff78828d08 in SLSMainConnectionID (SkyLight:x86_64+0x1ced08)
    #12 0x7fff604e5c9a in gfxLoadPluginData (libGFXShared.dylib:x86_64+0xc9a)
    #13 0x7fff5f30308a  (OpenCL:x86_64+0x2008a)
    #14 0x7fff7e9309bd in __pthread_once_handler (libsystem_pthread.dylib:x86_64+0x39bd)
[...]

PR #2431, opened 2019-02-23
Build ExecutionEngine on DeviceManager

nickgg:
*Description*: There are currently two ways to execute a compiled function: using the ExecutionEngine, or using the DeviceManager. The ExecutionEngine has a good interface but it would be good to have a single path to maintain for execution. This is the first part of making that change: backing the ExecutionEngine by a DeviceManager.
*Testing*: unit tests
*Documentation*: We'll fall back to the DummyDeviceManager if the BackendKind does not have a DeviceManager of its own, but this should be a short term solution.

### DIFF
--- a/lib/Backends/CMakeLists.txt
+++ b/lib/Backends/CMakeLists.txt
@@ -17,10 +17,12 @@ endif()
 
 add_library(Backend
               Backend.cpp
+              BackendUtils.cpp
               TraceEvents.cpp)
 target_link_libraries(Backend
                       PRIVATE
                         Base
+                        CodeGen
                         Graph
                         IR
                         Optimizer)
@@ -28,19 +30,14 @@ target_link_libraries(Backend
 add_library(Backends
               Backends.cpp)
 target_link_libraries(Backends
+                      PUBLIC
+                        Backend
                       PRIVATE
-                        BackendUtils
                         Base
                         Graph
                         Interpreter
                         Optimizer
                         ${linked_backends})
-
-add_library(BackendUtils BackendUtils.cpp)
-target_link_libraries(BackendUtils
-                      PRIVATE
-                      CodeGen
-                      IR)
 
 add_library(DeviceManager
               DeviceManagers.cpp)
--- a/lib/Backends/DeviceManagers.cpp
+++ b/lib/Backends/DeviceManagers.cpp
@@ -15,8 +15,10 @@
  */
 
 #include "glow/Backends/DeviceManager.h"
+#include "glow/Backends/DummyDeviceManager.h"
 
 #include "llvm/Support/Casting.h"
+#include "llvm/Support/raw_ostream.h"
 
 using namespace glow;
 using namespace glow::runtime;
@@ -61,7 +63,12 @@ DeviceManager *DeviceManager::createDeviceManager(BackendKind backendKind,
   case BackendKind::CPU:
     return createCPUDeviceManager(name);
   default:
-    GLOW_UNREACHABLE("not supported backend");
+    // As a fallback to make developing new Backends easier we'll create a
+    // DummyDeviceManager here, but this is not threadsafe and very simplistic.
+    // Strongly recommended that you create a DeviceManager customized for your
+    // device.
+    llvm::errs() << "Warning: Creating a DummyDeviceManager.\n";
+    return new DummyDeviceManager(backendKind, name);
   }
 
   // This is to make compiler happy. It can never reach this point as switch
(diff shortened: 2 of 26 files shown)


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -31,6 +33,8 @@ ExecutionEngine::ExecutionEngine(BackendKind backendKind) {
 /// Set the code generator kind to \p backendKind.
 void ExecutionEngine::setBackend(BackendKind backendKind) {
   setBackend(createBackend(backendKind));
+  device_ = runtime::DeviceManager::createDeviceManager(backendKind,
```
opti-mix: The comment in the header says `Note: at the moment This DOES NOT recreate the DeviceManager`, but you seem to create a new device manager here. Is that comment outdated? Or may be I misunderstand the code here?

If the comment is correct, may be it is a good idea to add an assert checking that the new backendKind is the same as the old one?

nickgg: The comment is on the version of setBackend that takes a Backend* instead of BackendKind. There's currently no way to get the BackendKind of a Backend which means we can't do that check, and we can't just recreate the DeviceManager with the correct BackendKind. I avoided adding it to keep the diff simple, but it's a small enough thing it makes sense to add now I guess.

nickgg:
LSAN failures were caused by leaking the constant buffers due to inconsistent freeing of the RuntimeBundle. This change is already catching bugs! 😎

nickgg:
I'm seeing a bunch of LSAN failures in the OpenCLDeviceManager but I think its probably #1255, might need to update suppressions?

nickgg:
@opti-mix I added the example DeviceManager we spoke about, but then wondered why we didn't just make it the default. Now this PR should not break any Backends that don't yet have a DeviceManager.

opti-mix:
@nickgg 

> I added the example DeviceManager we spoke about, but then wondered why we didn't just make it the default. Now this PR should not break any Backends that don't yet have a DeviceManager.

I totally like the idea! This way it is no too intrusive and allows our users to gradually migrate to the DeviceManager APIs. 

Nit: I'd suggest renaming this default DeviceManager from `InlineDeviceManager` into `DummyDeviceManager` to stress that it should not be used in production or for anything serious.

opti-mix:
@nickgg 

You still have in the PR header (and may be in the commit message) this wording:

This will break any backend that doesn't have a DeviceManager, you can unblock it by adding a DeviceManager that does nothing except execute the CompiledFunction in the call to runFunction and everything should work.

You may want to update it as we have the `DummyDeviceManager` now. And also add a very explicit advice for our users to implement real DeviceManagers for their backends! :-)

lib/Backends/DeviceManagers.cpp
```diff
@@ -61,7 +62,7 @@ DeviceManager *DeviceManager::createDeviceManager(BackendKind backendKind,
   case BackendKind::CPU:
     return createCPUDeviceManager(name);
   default:
-    GLOW_UNREACHABLE("not supported backend");
+    return new InlineDeviceManager(backendKind, name);
```
opti-mix: Could you add a comment here saying that we create a dummy device manager here as a fallback solution, but people should better implement their own device manager with proper semantics.

opti-mix:
You still need to rebase it on top of #2437 and adjust the APIs, right?

nickgg:
Rebased, renamed and fixed up some comments.

lib/Backends/CMakeLists.txt
```diff
@@ -17,31 +17,28 @@ endif()
...
+                        Backend
                       PRIVATE
-                        BackendUtils
                         Base
                         Graph
                         Interpreter
                         Optimizer
                         ${linked_backends})
 
-add_library(BackendUtils BackendUtils.cpp)
```
jackm321: @nickgg what was the cause for this change? I'd really like to detangle all of the dependencies here in `Backends/` but for example `~CompiledFunction()` which is defined `CompiledFunction.h` calls `freeConstants` defined in `BackendUtils.cpp`. This is breaking shared libraries if I try to remove the unnecessary dependency on `Backend` by `Backends` but I can't fix it because `BackendUtils` library is gone now.

jackm321: Actually this is probably better and we can just depend on `Backends` and `Backend` wherever both are needed elsewhere (example onnxifi, unit tests, etc) and that way we can remove the dependency of `Backend` by `Backends` and make the dependencies cleaner here.

PR #2435, opened 2019-02-24
[Operator] Add ElementwiseLinear Caffe2 op

### LINKED ISSUE #2420: [New operator] ElementwiseLinear
The spec for the `ElementwiseLinear` operator from Caffe2 can be found here https://caffe2.ai/docs/operators-catalogue.html#elementwiselinear.

It should be possible to lower this operator to existing elementwise `Mul` and `Add` operations.

SplitInfinity:
**Description**
This commit adds support for the `ElementwiseLinear` Caffe2 operator. It
is lowered early in the `glow::Function::createElementwiseLinear`
implementation to `Broadcast`, `Mul` and `Add`.

**Testing**
This commit adds a test for this operator to `OperatorTest` (CPU,
Interpreter and OpenCL backends) and two tests to `Caffe2ImporterTest`;
one that has `axis` specified in the test protobuf and one that does
not. These new tests and all old unit tests all pass.

**Fixes**
This commit fixes #2420.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -339,6 +339,15 @@ public:
   /// \p Y are 2D. \returns either the Mul or BatchedReduceAdd node.
   Node *createDotProduct(llvm::StringRef name, NodeValue X, NodeValue Y);
 
+  /// Create a node that implements the elementwise linear operator. \p X is
+  /// 2D and \p w and \p b are 1D. \p w and \p b are broadcasted to match the
+  /// shape of \p X and then the output is computed by multiplying \p X and
+  /// broadcasted \p w and adding broadcasted \p b. \returns the
+  /// ElementwiseLinearNode. \p axis indicates the axis of the inputs (the other
+  /// axis of \p X is assumed to be the batch index).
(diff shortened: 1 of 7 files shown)


include/glow/Graph/Graph.h
```diff
@@ -339,6 +339,15 @@ class Function final : public Named {
   /// \p Y are 2D. \returns either the Mul or BatchedReduceAdd node.
   Node *createDotProduct(llvm::StringRef name, NodeValue X, NodeValue Y);
 
+  /// Create a node that implements the elementwise linear operator. \p X is
+  /// 2D and \p w and \p b are 2D. \p w and \p b are broadcasted to match the
```
jfix71: typo: w and b are 1D, right?

tests/unittests/OperatorTest.cpp
```diff
@@ -5181,6 +5181,63 @@ TEST_P(Operator, dotProduct1D) {
   }
 }
 
+// Test a DotProduct operator with 2D inputs.
```
jfix71: Stale comment

include/glow/Graph/Graph.h
SplitInfinity: Yes, good catch.

PR #2436, opened 2019-02-25
[Quantization] Correctly quantize Select

jfix71:
*Description*: After https://github.com/pytorch/glow/pull/2417, Select no longer was quantized correctly, as it no longer has all float inputs.

*Testing*: Added a new test.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -179,6 +179,11 @@ protected:
       return (SLS->getData().getElementType() == ElemKind::FloatTy) &&
              (SLS->getWeights().getElementType() == ElemKind::FloatTy);
     }
+    case Kinded::Kind::SelectNodeKind: {
+      auto *SN = cast<SelectNode>(&node);
+      return SN->getLHS().getElementType() == ElemKind::FloatTy &&
+             SN->getRHS().getElementType() == ElemKind::FloatTy;
+    }
     default:
       // Let the general procedure handle this node kind.
       break;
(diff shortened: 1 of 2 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -333,6 +338,28 @@ class FunctionQuantizer : public FunctionConverter {
 
       break;
     }
+
+      // For Select, need to set LHS and RHS inputs with same quantization
```
SplitInfinity: Should the comment block here be aligned with the `case` below? If so, it doesn't appear to be.

jfix71: This is how `clang-format` does it, unfortunately. I prefer it to be aligned with the case. I could always move the comment into the case so it looks more natural.

lib/Quantization/Quantization.cpp
```diff
@@ -179,6 +179,11 @@ class FunctionQuantizer : public FunctionConverter {
       return (SLS->getData().getElementType() == ElemKind::FloatTy) &&
              (SLS->getWeights().getElementType() == ElemKind::FloatTy);
     }
+    case Kinded::Kind::SelectNodeKind: {
+      auto *SN = cast<SelectNode>(&node);
+      return SN->getLHS().getElementType() == ElemKind::FloatTy &&
```
opti-mix: @jfix71 Do you really need `FloatTy` here or should any FP type be OK? If so, we have a predicate to check that it is one of FP types.

jfix71: It could be any FP type, however we assume it's FloatTy everywhere here in the FunctionQuantizer. Might be good for a separate PR to make the change globally in the FunctionQuantizer.

PR #2437, opened 2019-02-25
[DeviceManager] Update DeviceManager interface

### LINKED ISSUE #2357: Add callback to DeviceManager evictNetwork
Add a callback to evictNetwork so the caller can wait until it completes before freeing required resources.

SplitInfinity:
**Description**
This commit updates the `DeviceManager` interface; `init`/`stop` now
return a `glow::runtime::ResultCode` and `DeviceManager::evictNetwork`
now takes a callback.

**Testing**
This commit modifies `DeviceManager` test to check for the return codes
from `init`, `stop` and `evictNetwork` (via its callback). All tests
pass.

**Fixes**
This commit fixes #2357.

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -29,6 +29,10 @@
 namespace glow {
 namespace runtime {
 
+/// Callback signalling success/failure of evicting a function from a Device.
+using EvictFunctionCBTy =
+    std::function<void(std::string functionName, runtime::ResultCode)>;
+
 /// Callback signalling success/failure of loading a Module onto a device.
 using ReadyCBTy = std::function<void(const Module *, runtime::ResultCode)>;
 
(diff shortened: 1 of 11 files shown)


include/glow/Backends/DeviceManager.h
```diff
@@ -29,6 +29,10 @@
 namespace glow {
 namespace runtime {
 
+/// Callback signalling success/failure of evicting a function from a Device.
+using EvictCBTy =
```
opti-mix: Nit: Does the name`EvictCBTy` need to be more descriptive? E.g. if we could potentially evict something else besides functions, we should better call it `EvictFunctionCBTy`.

nickgg:
Looks good, we should remember to move the OpenCLDeviceManager initialization from the ctor to init so we can propagate errors back.

SplitInfinity:
@nickgg 

In this PR or later?

nickgg:
Later is fine, I made #2439 for it.

SplitInfinity:
Cool, just pushed an update to address @opti-mix 's comments and to (hopefully) fix the ASAN build. I'm curious how we got away with not explicitly `delete`ing all of the `DeviceManager` instances when `DeviceManagerTest` was landed...

PR #2440, opened 2019-02-25
Few fixes for windows build.

ayermolo:
*Description*: Minor fixes to get things working again with Windows.
*Testing*: Unit Tests
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/tests/unittests/OnnxImporterTest.cpp
+++ b/tests/unittests/OnnxImporterTest.cpp
@@ -966,7 +966,7 @@ static void importPad(std::string fileName, const char *inputName,
     if (expectLoadError) {
       llvm::Error err = llvm::Error::success();
       ONNXModelLoader(NetFilename, {inputName}, {&data.getType()}, *F, &err);
-      EXPECT_TRUE(errorToBool(std::move(err)));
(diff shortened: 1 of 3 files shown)


tests/unittests/OnnxImporterTest.cpp
```diff
@@ -966,7 +966,7 @@ static void importPad(std::string fileName, const char *inputName,
     if (expectLoadError) {
       llvm::Error err = llvm::Error::success();
       ONNXModelLoader(NetFilename, {inputName}, {&data.getType()}, *F, &err);
-      EXPECT_TRUE(errorToBool(std::move(err)));
+      EXPECT_TRUE(glow::errorToBool(std::move(err)));
```
jackm321: @ayermolo for my own edification, why is this needed given the `using namespace glow;` above?

ayermolo: There is also llvm::errorToBool. VS 2017 can't figure out which one. Frankly I wasn't sure, but since using namespace glow; was used and no name space was specified assumed it was glow one.

jfix71: Yeah I actually just hit this issue too -- the compiler complains the call is ambiguous.

Issue #2298, opened 2019-01-24
Graph Partitioning


beicy:
This issue is created to track Graph Partitioning work. It is also related to #2045.
Current Plan: 
1. Get the initial partitions based on BFS order. Each partition should satisfy device memory constraints. 
2. By adjusting the initial partitions (e.g. Kernighan–Lin algorithm), find out the optimized partitions based on proposed cost model ( Communication cost, computation cost, and workload balance).

PR #2441, opened 2019-02-25
[Partitioner] Add cost functions to partitioner

nrsatish:
*Description*: 
The PR adds cost functions in terms of compute and memory bandwidth costs so that later stages in partitioning can use them. 

*Testing*:
Add a test with manually computed bounds into PartitionerTest.

*Documentation*:
The PR adds a field to the Partitioner class: ComputeTimeMapTy computeTime_ in Partitioner.h. It adds a function to fill out these fields. 

The field is a map from each Node in the Function being partitioned to the corresponding roofline for the op. This roofline is computed as the max of compute time, and the SRAM/DRAM read+write times for the inputs and outputs of the node. In order to compute these rooflines, fields have been added to DeviceInfo struct in RuntimeTypes.h

The PR is related to Graph Partitioning #2298

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -48,6 +48,17 @@ using ResultCBTy = std::function<void(
 struct DeviceInfo {
   /// Available memory on device in bytes.
   size_t availableMemory;
+  /// Available SRAM capacity in bytes.
+  size_t sramCapacity;
+  /// Peak compute on device in ops/second. Assumes all ops are in int8.
+  /// TODO: distinguish between data types with different peak flops.
+  float peakCompute;
+  /// Peak memory bandwidth from DRAM on device in bytes/second.
+  float peakDramBw;
+  /// Peak memory bandwidth from SRAM on device in bytes/second.
+  float peakSramBw;
+  /// Peak ingress/egress PCI-E bandwidth from device in bytes/second.
+  float peakPCIeBw;
 };
 
 /// Individual Node in the DAG for a given network. This contains all the
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -81,6 +81,126 @@ void Partitioner::initOpMemUsage() {
   }
 }
 
+/// Get the minimal compute time for each op in the function.
+void Partitioner::initOpComputeTime() {
+  computeTime_.clear();
+
+  // This code assumes all ops are BW limited from SRAM; except
+  // if the input does not fit in SRAM -- then it is DRAM BW limited
+  float peakDramBw = deviceInfo_[0].peakDramBw;
+  float peakSramBw = deviceInfo_[0].peakSramBw;
+  size_t sramCapacity = deviceInfo_[0].sramCapacity;
+  float peakCompute = deviceInfo_[0].peakCompute;
+
+  for (auto &node : F_->getNodes()) {
+    /// compute memory side bytes for inputs from DRAM, SRAM.
+    /// TODO: think about whether this is better off computed inside a Node.
+
+    int n = node.getNumInputs();
+    uint64_t sizeDram = 0;
+    uint64_t sizeSram = 0;
+    if (node.getKind() == Kinded::Kind::SaveNodeKind) {
+      computeTime_[&node] = 0.0f;
+      continue;
+    }
+
+    /// The memory bytes for embedding table lookups is data dependent,
+    /// so it needs to be calculated as per the number of indices accessed.
+    if (node.getKind() == Kinded::Kind::SparseLengthsWeightedSumNodeKind) {
+      auto *SLWSN = llvm::dyn_cast<SparseLengthsWeightedSumNode>(&node);
+      /// compute how many entries of the embedding table we look up
+      auto numLookups = SLWSN->getIndices().getNode()->dims(0).front();
+      /// compute how many bytes we read per lookup
+      auto tableSize = SLWSN->getData().getNode()->getType(0)->getSizeInBytes();
+      auto numRows = SLWSN->getData().getNode()->dims(0).front();
+      auto sizePerLookup = tableSize / numRows;
+      /// compute total bytes read
+      uint64_t sizeInput = numLookups * sizePerLookup;
+
+      /// does the table fit in SRAM or DRAM
+      if (tableSize > sramCapacity) {
+        sizeDram += sizeInput;
+      } else {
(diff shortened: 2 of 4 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -81,6 +81,95 @@ void Partitioner::initOpMemUsage() {
...
+      if (size_output > sramCapacity) {
+        size_dram += size_output;
+      } else {
+        size_sram += size_output;
+      }
+    }
+
+    /// Assume actual compute is only for Matmuls.
+    /// TODO: fix this assumption later (FC, Convolutions, BatchMatmul)
+    uint64_t total_ops = 0;
```
beicy: Why can we do this assumption? The implementation of `(FC, Convolutions, BatchMatmul)` is an enhancement or "must be implemented" ?   If we want to test e.g. Resnet50, can we just use this function without considering `FC, Convolutions, BatchMatmul` ?

lib/Partitioner/Partitioner.cpp
```diff
@@ -81,6 +81,95 @@ void Partitioner::initOpMemUsage() {
...
+
+    /// Compute compute roofline as max of flops, DRAM, SRAM BW
+    computeTime_[&node] = std::max(
+        total_ops * 1.0f / peakCompute,
+        std::max(size_dram * 1.0f / peakDramBw, size_sram * 1.0f / peakSramBw));
+  }
+}
+
+/// Get the minimal communication time for each edge in the function.
+void Partitioner::initEdgeCommTime() {
```
beicy: I am not sure if we do need this initialization. The computation of the communication is pretty easy : output_size / peakPCIeBw and we can get it when we need it. Therefore, I don't think we need a map to store all the values.

beicy:
Thanks @nrsatish for working on it! If possible, could you please for add some open sourced info about roofline (like why the roofline is computed as the max of compute time (only for MatMul nodes right now) )?
In addition, I changed the title of this PR.

lib/Partitioner/Partitioner.cpp
nrsatish: @beicy : I think it is more in the required category. I will add in support for these ops.

lib/Partitioner/Partitioner.cpp
nrsatish: Ok, I think I agree with that.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -48,6 +48,17 @@ using ResultCBTy = std::function<void(
 struct DeviceInfo {
   /// Available memory on device in bytes.
   size_t availableMemory;
+  /// Available SRAM capacity in bytes.
+  size_t sramCapacity;
+  /// Peak compute on device in ops/second. Assumes all ops are in int8.
+  /// TODO: distinguish between data types with different peak flops.
+  float peakCompute;
```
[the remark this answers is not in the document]
nrsatish: Ideally yes. In general, it would be good to get the characteristics of the architecture from querying some vendor API. This current design is the first step towards that.

lib/Partitioner/Partitioner.cpp
```diff
@@ -81,6 +81,117 @@ void Partitioner::initOpMemUsage() {
...
+      auto inputDims = FCN->getInput().dims();
+      auto wtDims = FCN->getWeights().dims();
+      totalOps = 2 * inputDims[0] * inputDims[1] * wtDims[1];
+    } else if (node.getKind() == Kinded::Kind::ConvolutionNodeKind) {
+      auto *CN = llvm::dyn_cast<ConvolutionNode>(&node);
+      auto resultDims = CN->getResult().dims();
+      // Get the product of batch, output height, output dims, output channels
+      totalOps = resultDims[0];
+      for (int i = 1; i < resultDims.size(); i++) {
+        totalOps *= resultDims[i];
```
beicy: here "i" should be size_t, otherwise, the type check will fail.
Usually, it can be wrote as "for (size_t i = 1, e = resultDims.size(); i < e; i++)"

lib/Partitioner/Partitioner.cpp
```diff
@@ -427,7 +537,10 @@ DAGNodeList &Partitioner::Partition() {
   // Prepare 1: Get the min memory usage for each op.
   initOpMemUsage();
 
-  // Prepare 2: TODO: get the minimal comunication cost for any 2 ops (i.e. the
+  // Prepare 2: Get the roofline memory bandwidth estimate for each op.
+  initOpComputeTime();
+
+  // Prepare 3: TODO: get the minimal comunication cost for any 2 ops (i.e. the
```
beicy: the comment is not used.

lib/Partitioner/Partitioner.cpp
```diff
@@ -81,6 +81,117 @@ void Partitioner::initOpMemUsage() {
...
+      if (sizeOutput > sramCapacity) {
+        sizeDram += sizeOutput;
+      } else {
+        sizeSram += sizeOutput;
+      }
+    }
+
+    /// Calculate compute ops. Currently only computed for Matmul, Conv, FC
+    /// TODO: think about whether this is better off computed inside a Node.
+    uint64_t totalOps = 0;
```
beicy: Just want to double check again here: in the future, do we need to add the computation for each node?

lib/Partitioner/Partitioner.cpp
```diff
@@ -81,6 +81,117 @@ void Partitioner::initOpMemUsage() {
...
+    if (node.getKind() == Kinded::Kind::MatMulNodeKind) {
+      auto *MMN = llvm::dyn_cast<MatMulNode>(&node);
```
beicy: I prefer using "switch". If we need to add more node type here, "switch" looks better:)

lib/Partitioner/Partitioner.cpp
nrsatish: I can delete that comment.

lib/Partitioner/Partitioner.cpp
nrsatish: Yes we do. At least for  memory bytes if not flops.

lib/Partitioner/Partitioner.cpp
nrsatish: Makes sense.

lib/Partitioner/Partitioner.cpp
nrsatish: But for most ops, flops is less important. There are only a handful of ops here that will be at all compute bound.

PR #2443, opened 2019-02-25
[tensor] Make isEqual work correctly with NAN

bertmaher:
*Description*: We were erroneously reporting NAN as equal to everything instead of not-equal.
*Testing*: New unit tests.
*Documentation*: n/a


bertmaher:
This is still arguably incorrect for INF, since INF vs INF should be true (IEEE speaking), but this method will report false now (since INF - INF => NAN).  I wonder if there's a fix that isn't super complicated, but this PR is at least better than treating NANs as widcards :-/

bertmaher:
The compiler saw through my clever negation trick, and optimized it away in our Release build (because of -ffast-math, it assumes NaN never happens).  So let's go with an explicit std::isnan check, which seems to be OK in release mode.

(Oddly, if I write a standalone function that calls std::isnan and look at the disassembly, it always returns false.  So, I'm not really sure how much confidence I should have in std::isnan either)

bertmaher:
Compiler 2, Bert 0.  Going to revisit this problem tomorrow.

jackm321:
Lol I'm also going 0/2 with the accepts, I should have looked at the test output first.

jackm321:
Just messed around in godbolt for a bit but didn't find much, will be interesting if there's a way around this one.

bertmaher:
After some fiddling with godbolt I just don't think it's sane to try to handle NAN when -ffinite-math-only (implied by -ffast-math) is in effect.  Even if something happens to work, the compiler doesn't guarantee that it will continue working.  So, let's disable finite-math-only.

PR #2445, opened 2019-02-26
[Graph] keep the dotFile

zakk0610:
*Description*: I think we need to keep the dot file after calling dumpDAG(). The use scenario is calling dumpDAG() in somewhere but user can't got the file. 
*Testing*:
*Documentation*:
[Optional Fixes #issue]


bertmaher:
Good catch, thanks.

PR #2448, opened 2019-02-26
[Partitioner] Fix a BFS issue.

beicy:
*Description*:

When I run a testcase for several times, I found the final partitions of each run could be a little bit different. It was caused by the sequence of the nodes of each BFS levels (The key of the map we used is a pointer, which could be different value for each run. ) Therefore, to make sure the results are consistent for each run, I would like to add a sort for each level.

*Testing*:
unittest.
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Partitioner/PartitionerUtils.h
+++ b/include/glow/Partitioner/PartitionerUtils.h
@@ -35,8 +35,8 @@ struct GraphMemInfo {
   GraphMemInfo() : inMemSize(0), outMemSize(0), constMemSize(0){};
 };
 
-/// A list of <level, nodelist> with BFS order.
-using BFSLevel = std::vector<std::pair<int, std::vector<Node *>>>;
+/// A list of <nodelist> with BFS order.
+using BFSLevel = std::vector<std::vector<Node *>>;
 
 /// Visit nodes if Function \p F in BFS order and return the nodes by levels
 /// (the longest distance between one node and the root).
--- a/lib/Partitioner/PartitionerUtils.cpp
+++ b/lib/Partitioner/PartitionerUtils.cpp
@@ -15,16 +15,22 @@
  */
 
 #include "glow/Partitioner/PartitionerUtils.h"
+#include <unordered_set>
 
 using llvm::isa;
 
 namespace glow {
 
+/// Used to sort 2 Nodes by their name.
+static bool compFunc(Node *n1, Node *n2) {
+  return (n1->getName().compare(n2->getName()) > 0);
+}
+
 /// The nodes in function \p F which be grouped into levels based on how far
 /// (the longest distance) they are from the roots.
 BFSLevel getBFSLevel(Function *F) {
-  // The current list of nodes needs to be visited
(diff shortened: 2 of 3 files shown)


beicy:
Thanks @opti-mix for helping me debugging this issue.

include/glow/Partitioner/PartitionerUtils.h
```diff
@@ -33,10 +33,13 @@ struct GraphMemInfo {
   size_t constMemSize;
 
   GraphMemInfo() : inMemSize(0), outMemSize(0), constMemSize(0){};
+
+  // Get the total memory size of each partition.
+  size_t getTotalMemSize() { return inMemSize + outMemSize + constMemSize; }
 };
 
 /// A list of <level, nodelist> with BFS order.
```
jfix71: Comment needs to be updated?

include/glow/Partitioner/PartitionerUtils.h
```diff
@@ -33,10 +33,13 @@ struct GraphMemInfo {
   size_t constMemSize;
 
   GraphMemInfo() : inMemSize(0), outMemSize(0), constMemSize(0){};
+
+  // Get the total memory size of each partition.
```
jfix71: nit: Need three `/` for doxygen

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -20,6 +20,10 @@ using llvm::isa;
 
 namespace glow {
 
+bool compFunc(Node *n1, Node *n2) {
```
jfix71: Add doxygen please. Also can this be static?

include/glow/Partitioner/PartitionerUtils.h
```diff
@@ -33,10 +33,13 @@ struct GraphMemInfo {
...
+  size_t getTotalMemSize() { return inMemSize + outMemSize + constMemSize; }
```
jfix71: This seems to be unused/unrelated to the PR.

beicy:
@jfix71 You are right! The issue I found is from debugging an optimization, and I thought it should be better to split the BFS and create a PR first. But forgot to remove the unrelated one. Will remove them.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -41,24 +46,25 @@ BFSLevel getBFSLevel(Function *F) {
   BFSLevel bfs;
   int level = 0;
   int current = 0;
-  bfs.push_back({level, std::vector<Node *>()});
+  bfs.push_back(std::vector<Node *>());
   level++;
   while (current < level) {
     std::vector<Node *> nodes;
+    std::set<Node *> used;
```
jfix71: So `used` is there to make sure that the same node isn't inserted twice into `nodes`, right? Would it be possible to just move `nodes` and `cur` to be `std::unordered_set`, and then not introduce `used` at all? No big deal but maybe would make it a little simpler.

PR #2451, opened 2019-02-26
Refactor Backend::compile so it takes an options struct.

nickgg:
*Description*: We have 3 different methods in the Backend to invoke compile: `compileWithoutConstants()`, `instrumentAndCompile()` and the plain `compile()` (uh oh what if you want to compile without constants but with instrumentation?!). This PR collapses the three down to a single compile() function which takes a CompileOptions struct which indicates how the compile should be done. At the moment it's got only those options (collectConstants & autoInstrument).
*Testing*: unit tests.
*Documentation*: This changes the Backend interface and so all backends will need an update, but I kept the standard `compile()`to make migration easier.

### DIFF
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -73,7 +73,9 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
 
   llvm::outs() << "Starting compile.\n";
   backend->optimizeFunction(CompilationMode::Infer, F);
-  return backend->instrumentAndCompile(F);
+  CompileOptions opts;
+  opts.autoInstrument = true;
+  return backend->compile(F, opts);
 }
 
 std::future<ResultCode> addToDevice(unsigned int id, DeviceManager *device,
--- a/include/glow/Backends/Backend.h
+++ b/include/glow/Backends/Backend.h
@@ -35,6 +35,15 @@ enum class BackendKind {
   CPU,         // Compile and run the code on the host.
 };
 
+/// Configuration options for the compile() method on Backend.
+struct CompileOptions {
+  /// Allocate and collect constant Tensors in the RuntimeBundle.
+  bool collectConstants{true};
+
+  /// Insert TraceEvents between all instructions for profiling.
+  bool autoInstrument{false};
+};
+
 // This is the interface that glow backends need to implement.
 class Backend {
 public:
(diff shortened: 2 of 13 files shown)


include/glow/Backends/Backend.h
```diff
@@ -44,21 +53,14 @@ class Backend {
...
+  virtual std::unique_ptr<CompiledFunction> compile(Function *F) const {
+    CompileOptions opts;
+    return compile(F, opts);
   }
 
-  /// Generate code for input function \param F but do not collect constants.
+  /// Generate code for input function \param F.
   virtual std::unique_ptr<CompiledFunction>
-  compileWithoutConstants(Function *F) const = 0;
+  compile(Function *F, CompileOptions &opts) const = 0;
```
opti-mix: Should it be `const CompleOptions &opts` or do you intend to change them under certain circumstances?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -133,29 +133,22 @@ CPUBackend::compileIRWithoutConstants(IRFunction *IR) const {
...
-  auto compiledFunc = compileIRWithoutConstants(IR.get());
+  if (opts.autoInstrument) {
+    autoInstrument(traceInfo, IR.get());
+  }
+
+  std::unique_ptr<CompiledFunction> compiledFunc;
+  if (opts.collectConstants) {
+    compiledFunc = compileIR(std::move(IR));
+  } else {
+    compiledFunc = compileIRWithoutConstants(IR.get());
```
opti-mix: Do we really need different functions for `compileIR` and `compileIRWithoutConstants` or can we fold them together now and conditionally perform constants collection if needed?

examples/tracing-compare.cpp
```diff
@@ -73,7 +73,8 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
 
   llvm::outs() << "Starting compile.\n";
   backend->optimizeFunction(CompilationMode::Infer, F);
-  return backend->instrumentAndCompile(F);
+  CompileOptions opts{true, true};
```
opti-mix: I'm slightly annoyed by this form of the initialization. If you look at it, you don't really know which options were initialized. Could you use a more explicit syntax with field names?

nickgg: Sure thing, just trying to save on lines I guess but this isn't super readable (plus it doesn't work on g++ 4.8 as it turns out).

lib/Backends/CPU/CPUBackend.cpp
nickgg: So I think ultimately we don't need both - and I did try to remove it, but due to the type signatures it's a bit complicated: Currently takes a unique_ptr to the IR (and e.g. in the Interpreter stores this unique_ptr), but collecting constants requires the Module or the IR, so we can't move it in then collect constants afterwards since we wont have either.

opti-mix: Never mind then! Let's skip this for now.

opti-mix:
@nickgg Quick question: We also have functions that take `CompilationMode` arguments. Should we consider making them use `CompileOptions` instead, where `CompileOptions` would include a flag for the`CompilationMode`? WDYT?

nickgg:
@opti-mix I like the idea, it'll be a large interface change but I think it's sensible to rip the bandaid off all at once (we can keep some simple overrides to reduce churn). It would mean changing:

* `optimize / optimizeFunction` to take a CompilationOptions - this is good since we can support more granular optimization config than just Infer or Train.
* `ExecutionEngine::compile` - probably should do this anyway in this PR.
* `transformPostLowering` - unsure if we need more info here than just CompilationMode? probably, for the same reason as optimize.

I'm down, but anyone else have opinions on this?

nickgg:
I'll do it today but in a new PR just to keep things small and easy to review.

PR #2452, opened 2019-02-26
[Easy] Add transformPostLowering call to checkGraphCompatibility

### LINKED ISSUE #2190: [ONNXIFI] Various fixes for onnxGetBackendCompatibility
* onnxGetBackendCompatibility calls `glow::lower()` to lower graph nodes to what the backends would expect to see but doesn't delete the original nodes so DCE should be run.
* `verify()` should be run on the graphs created by `onnxGetBackendCompatibility` in order to check that they are indeed valid graphs
* `Backend::transformPreLowering()` should be called on the loaded graphs before lowering

jackm321:
*Description*:
Graphs should go through transformPostLowering in checkGraphCompatibility because this is what will happen in initNet

*Testing*:
ninja check

*Documentation*:
Fixes #2190

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -54,6 +54,8 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
 
   glow::lower(function, /* loweredMap */ nullptr, glowBackend_.get());
 
+  glowBackend_->transformPostLowering(function, CompilationMode::Infer);
+
   const auto &nodes = function->getNodes();
 
   for (const auto &node : nodes) {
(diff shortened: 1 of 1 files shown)


jackm321:
I'll need to add DCE to this.

lib/Onnxifi/Base.cpp
```diff
@@ -54,6 +54,10 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
 
   glow::lower(function, /* loweredMap */ nullptr, glowBackend_.get());
 
+  glowBackend_->transformPostLowering(function, CompilationMode::Infer);
```
jfix71:
```suggestion
  if (glowBackend_->transformPostLowering(function, CompilationMode::Infer)) {
    glow::DCE(function);
  }
```

jackm321:
@jfix71 is this ok now?

lib/Onnxifi/Base.cpp
```diff
@@ -54,6 +54,12 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
 
   glow::lower(function, /* loweredMap */ nullptr, glowBackend_.get());
 
+  // Call the backend's transformPostLowering to match the normal compilation
+  // pipeline then DCE any nodes that are no longer needed
```
jfix71: nit: period at end of sentence 😃

PR #2456, opened 2019-02-27
[ee] Don't leak DeviceManagers

bertmaher:
*Description*: This seems like a leak, since we assign device_ in setBackend (below), and LSAN reports it to me, but only when I'm using an experimental private backend.  I don't see any such warning on trunk, which makes me suspicious that I've done something wrong...

*Testing*: 

*Documentation*:


yinghai:
@nickgg could you follow up with unique_ptr?

nickgg:
@yinghai sure will do

PR #2459, opened 2019-02-27
Refactor optimize() functions to take a CompileOptions

nickgg:
*Description*: As a follow up #2451, push the CompileOptions struct down into the Optimize functions as well for future expansion of optimization configuration. I kept the existing CompilationMode function to ease transition.
*Testing*: unit tests, w/ ASAN.
*Documentation*:
This changes the Backend interface and will require updating any new backends.

### DIFF
--- a/examples/resnet-concurrent.cpp
+++ b/examples/resnet-concurrent.cpp
@@ -75,8 +75,11 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module) {
   Function *F = module.getFunction("resnet50");
 
   llvm::outs() << "Starting compile.\n";
-  backend->optimizeFunction(CompilationMode::Infer, F);
-  return backend->compile(F);
+
+  CompilationOptions opts;
+  opts.mode = CompilationMode::Infer;
+  backend->optimizeFunction(F, opts);
+  return backend->compile(F, opts);
 }
 
 /// Loads the CompliedFunction into device \p device.
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -72,9 +72,10 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
   Function *F = module.getFunction("resnet50");
 
   llvm::outs() << "Starting compile.\n";
-  backend->optimizeFunction(CompilationMode::Infer, F);
-  CompileOptions opts;
+  CompilationOptions opts;
+  opts.mode = CompilationMode::Infer;
   opts.autoInstrument = true;
+  backend->optimizeFunction(F, opts);
   return backend->compile(F, opts);
 }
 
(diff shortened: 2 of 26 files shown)


include/glow/Backends/CompileOptions.h
```diff
@@ -0,0 +1,40 @@
...
+namespace glow {
+
+enum class CompilationMode {
+  Train, /// Compile the graph in preperation for training.
+  Infer, /// Compile the graph for inference. Notice that this operation
+         /// changes the graph in a way that is not reversible.
+};
+
+/// Configuration options for the compile() method on Backend.
+struct CompileOptions {
```
opti-mix: We have `CompilationMode`, but `CompileOptions`. This looks odd. Should we use a more consistent naming?

opti-mix: Nit: Why do we have this header in the Backends subdirectory? Is it the most logical place to put it?

jfix71:
Maybe we should consider the same builder pattern as we use in NodeGen/InstrGen -- I think it would make things cleaner. Specifically adding something to the CompileOptions struct like:
```cpp
CompileOptions &setCompilationMode(CompilationMode m) {
  mode = m;
  return *this;
}
```

Then you can do something like:
```cpp
EE_.compile(F, CompileOptions()
                   .setCompilationMode(CompilationMode::Infer)
                   .setCollectConstants(true));
```

include/glow/Backends/CompileOptions.h
nickgg: What's your preferred alternative?

opti-mix: For naming, I'd go with `CompilationMode` and `CompilationOptions`.

For the subdirectory, I don't have any strong opinion. May be a new `include/glow/Options/` subdirectory?

nickgg:
I'm not totally convinced by the builder pattern here but also not really against it. Can we gather a few more opinions on it?

include/glow/Backends/CompileOptions.h
nickgg: CompilationOptions sounds fine to me but creating a new subdirectory for a single header seems overkill, what else would we need to put in there?

opti-mix:
> I'm not totally convinced by the builder pattern here but also not really against it.

+1.
I'm also not totally convinced if using the builder pattern for options is worth the effort.

include/glow/Backends/CompileOptions.h
opti-mix: > but creating a new subdirectory for a single header seems overkill, what else would we need to put in there?

Yeah, I agree. My initial (very small) concern was that the options are not for backends only. They are also used by optimizers, etc.

nickgg:
@jfix71 I'll land this now to avoid conflicts, and we can come back to the builder pattern question.

PR #2463, opened 2019-02-28
Added logicalDevice support to provisioner

gcatron:
*Description*: This updates the provisioner to honor logicalDevice in the DAGNode. 
*Testing*: unit tests pass, resnet-runtime runs successfully.
*Documentation*: NA

### DIFF
--- a/lib/Runtime/Provisioner/Provisioner.cpp
+++ b/lib/Runtime/Provisioner/Provisioner.cpp
@@ -20,6 +20,7 @@
 #include "glow/Graph/Graph.h"
 
 #include <future>
+#include <map>
 #include <queue>
 
 using namespace glow;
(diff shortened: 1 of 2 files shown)


lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -29,85 +30,96 @@ Provisioner::Provisioner(DeviceManagerMapTy &devices) {
   for (auto &device : devices) {
     devices_.push_back(device.second.get());
   }
+  auto backendKind = devices[0]->getBackendKind();
+  backend_.reset(createBackend(backendKind));
 }
 
+void walkNetwork(DAGNode *node,
+                 std::map<DeviceIDTy, std::vector<DAGNode *>> &logicalDevices) {
```
beicy: nit: is the consistent traversal order in logicalDevices necessary?  If not, maybe std::unordered_map is better :)

gcatron: I went with map because the number of elements is small enough it is more memory efficient than unordered_map.

beicy: I see. I didn't notice the size of the map :) So map should be better!

PR #2465, opened 2019-02-28
Lower ReduceMean in Lower or optimize with AvgPool when possible.

### LINKED ISSUE #2343: ReduceSum operation should support multiple axes
Running ResNet v1 50 model obtained from TF2ONNX and it fails with:
      `"Only supporting single reduction for now."`
This originates from `ReduceSum` operation processing in` CommonOperatorLoader.h` from `Importer`

I assume the support for multiple axis is to be added to `BatchedReduceAddNode`, however, I am curious why a simpler implementation has not being done to at least support ReduceSum w/ multiple axes until there is a better implementation:
For example, repeating current code in a for loop for each dimension, to emit serialized `BatchedReduceAddNode`  nodes (or serialized `BatchedReduceAddNode`, `Splat`, `Divide`, for `ReduceMean)` seems trivial yet enables more use cases.

vuzelac-cadence:
*Description*: 
`reduceMeanOrSum`  Importer function is split into into `reduceSum` and `reduceMean`. `reduceMean` is not lowered in Importer but in Optimizer. `reduceMean` is also optimized with `AvgPool` node if the reduce parameters are suitable:  input is 4D, last two dimensions are reduced, `keepdims=true`. 
*Testing*: 
OnnxImporterTest:  Add 4 reduce* testcases that run 4 newly added `reduce*.onnxtxt` models.
GraphOptzTest: Add a testcase that covers ReduceMean substitution with AvgPool.
OperatorTest: Add testcases that covered both optimized and lowered ReduceMean transformations.
*Documentation*: 
Updated Optimizations.md
Fixes #2343

### DIFF
--- a/docs/Optimizations.md
+++ b/docs/Optimizations.md
@@ -67,6 +67,12 @@ Below you can see the list of currently supported graph optimizations:
     This optimization performs a classic CSE with a goal of avoiding of any
     results that were computed already.
 
+  * Optimization of ReduceMean nodes
+
+    This optimization performs substitions of ReduceMean with AvgPool node if
+    the reduce parameters are suitable: input is 4D with last two dimensions
+    to be reduced. 
+
 #### Quantization specific optimizations
 
 Majority of the common optimizations above can be used on a quantized graph.
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -564,27 +564,31 @@ public:
   Node *createParallelBatchMatMul(llvm::StringRef name, NodeValue lhs,
                                   NodeValue rhs);
 
+  /// Create a node, performing BatchedReduceAdd operation. Output type is
+  /// based on the input \p batch type with dimensions specified with \p axes
+  /// removed.
   BatchedReduceAddNode *createBatchedReduceAdd(llvm::StringRef name,
                                                NodeValue batch,
-                                               unsigned_t axis);
+                                               llvm::ArrayRef<unsigned_t> axes);
 
+  /// Create a node, performing BatchedReduceAdd operation. Output type
+  /// matches input \p outTy type.
   BatchedReduceAddNode *createBatchedReduceAdd(llvm::StringRef name,
                                                TypeRef outTy, NodeValue batch,
-                                               unsigned_t axis);
+                                               llvm::ArrayRef<unsigned_t> axes);
 
-  /// Implements a batched reduce mean of the \p batch on the provided \p axis
-  /// with output type \p outTy with three nodes: a BatchedReduceAdd followed by
-  /// a DivNode with a SplatNode of the length of the \p axis
-  /// dimension. \returns the final DivNode.
-  DivNode *createBatchedReduceMean(llvm::StringRef name, TypeRef outTy,
-                                   NodeValue batch, unsigned_t axis);
+  /// Create a node, performing BatchedReduceMean operation. Output type
+  /// matches input \p outTy type.
+  BatchedReduceMeanNode *
+  createBatchedReduceMean(llvm::StringRef name, TypeRef outTy, NodeValue batch,
+                          llvm::ArrayRef<unsigned_t> axes);
 
-  /// Implements a batched reduce mean of the \p batch on the provided \p axis
-  /// with three nodes: a BatchedReduceAdd followed by a DivNode with a
-  /// SplatNode of the length of the \p axis dimension. \returns the final
-  /// DivNode.
-  DivNode *createBatchedReduceMean(llvm::StringRef name, NodeValue batch,
-                                   unsigned_t axis);
+  /// Create a node, performing BatchedReduceMean operation. Output type is
+  /// based on the input \p batch type with dimensions specified with \p axes
+  /// removed.
+  BatchedReduceMeanNode *
+  createBatchedReduceMean(llvm::StringRef name, NodeValue batch,
+                          llvm::ArrayRef<unsigned_t> axes);
 
   BatchedAddNode *createBatchedAdd(llvm::StringRef name, NodeValue batch,
                                    NodeValue sample);
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -625,17 +625,19 @@ protected:
     NodeValue in;
     ASSIGN_VALUE_OR_RETURN_ERR(in,
                                getNodeValueOrCreateConstantByName(op.input(0)));
-    auto axes = getShape(dict["axes"]);
-    RETURN_ERR_IF_NOT(axes.size() == 1,
-                      "Only supporting single reduction for now.");
-    auto axis = axes[0];
 
-    Node *node = nullptr;
+    auto shapeAxes = getShape<unsigned_t>(dict["axes"]);
+    std::sort(shapeAxes.begin(), shapeAxes.end());
 
-    if (typeName == "ReduceMean") {
-      node = G_.createBatchedReduceMean(opName, in, axis);
-    } else {
-      node = G_.createBatchedReduceAdd(opName, in, axis);
+    llvm::ArrayRef<unsigned_t> axes(shapeAxes);
+
+    // Check if axes elements are unique.
+    if (axes.size() > 1) {
+      auto it = std::unique(shapeAxes.begin(), shapeAxes.end());
+      if (it != shapeAxes.end()) {
+        RETURN_ERR("Axes values are not unique.",
+                   GlowErr::ErrorCode::MODEL_LOADER_UNSUPPORTED_SHAPE);
+      }
     }
 
     bool keepDims = true;
@@ -645,10 +647,23 @@ protected:
       keepDims = (bool)keepdims;
     }
 
-    // Our batched reduce add does not keep the dim; reshape if necessary.
+    Node *node = nullptr;
+
+    if (typeName == "ReduceMean") {
+      node = G_.createBatchedReduceMean(opName, in, axes);
+    } else {
+      node = G_.createBatchedReduceAdd(opName, in, axes);
+    }
+
+    // Our batched reduce add/mean does not keep the dim; reshape if necessary.
     if (keepDims) {
+
       std::vector<size_t> shape = node->dims(0);
-      shape.insert(shape.begin() + axis, 1);
+
+      // Add removed axes. Requires increasing order sort - done above.
+      for (const auto &axis : shapeAxes) {
+        shape.insert(shape.begin() + axis, 1);
+      }
       node = G_.createReshape(opName, node, shape);
     }
 
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -335,14 +335,23 @@ TypeRef Module::uniqueType(const Type &T) {
 
 TypeRef Module::getVoidTy() { return uniqueType(Type()); }
 
-/// \returns a ShapeVector of rank 1 less than the input \p dims, where the
-/// provided \p axis dimension is removed from the shape.
-static ShapeVector getNewShapeWithoutAxis(llvm::ArrayRef<size_t> dims,
-                                          size_t axis) {
-  assert(axis < dims.size() &&
-         "Axis to remove must fit inside dimensions of the provided dims.");
+/// \returns a ShapeVector of rank axes.size() less than the input \p dims,
+/// where the provided \p axes dimensions are removed from the shape.
+static ShapeVector getNewShapeWithoutAxes(llvm::ArrayRef<size_t> dims,
+                                          llvm::ArrayRef<unsigned_t> axes) {
+  assert(axes.size() < dims.size() &&
+         "Axes to remove must fit inside dimensions of the provided dims.");
   ShapeVector newDims(dims.begin(), dims.end());
-  newDims.erase(newDims.begin() + axis);
+  ShapeVector shapeAxes(axes.begin(), axes.end());
+
+  // Sort so that looping erase below doesn't fail.
+  std::sort(shapeAxes.rbegin(), shapeAxes.rend());
+
+  for (const auto &axis : shapeAxes) {
+    assert(axis <= dims.size() &&
+           "Axis to remove must fit inside dimensions of the provided dims.");
+    newDims.erase(newDims.begin() + axis);
+  }
   return newDims;
 }
 
@@ -1332,10 +1341,14 @@ Node *Function::createParallelBatchMatMul(llvm::StringRef name, NodeValue lhs,
                        {numBatches, N, P});
 }
 
-BatchedReduceAddNode *Function::createBatchedReduceAdd(llvm::StringRef name,
-                                                       TypeRef outTy,
-                                                       NodeValue batch,
-                                                       unsigned_t axis) {
+BatchedReduceAddNode *
+Function::createBatchedReduceAdd(llvm::StringRef name, TypeRef outTy,
+                                 NodeValue batch,
+                                 llvm::ArrayRef<unsigned_t> axes) {
+
+  assert(axes.size() == 1 && "Only supporting single reduction for now.");
+  auto axis = axes[0];
+
   // Calculate the expected total number of elements in the output tensor
   // based on the number of elements in the batch divided by the axis
   // dimension.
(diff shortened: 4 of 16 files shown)


vuzelac-cadence:
Hopefully somebody can spend a little time on the OpenCL issue we have here and provide us with some useful info. We don't use OpenCL but likely will have to, in order to solve this issue.
So, the issue is in OpenCL/Operator.batchedReduceMeanUsingAvgPool/0 and the results are mismatching the expected results. The test passes on CPU & Interpreter, while the quantized version of this test passes on all 3.

vuzelac-cadence:
[WIP] removed and testing problems fixed by removing OpenCL testing from newly added tests. This is similar to what some of the existing tests are doing. Issue #2469 is opened to track OpenCL issues.

include/glow/Graph/Graph.h
```diff
@@ -573,18 +573,30 @@ class Function final : public Named {
...
-  DivNode *createBatchedReduceMean(llvm::StringRef name, TypeRef outTy,
-                                   NodeValue batch, unsigned_t axis);
-
-  /// Implements a batched reduce mean of the \p batch on the provided \p axis
-  /// with three nodes: a BatchedReduceAdd followed by a DivNode with a
-  /// SplatNode of the length of the \p axis dimension. \returns the final
-  /// DivNode.
-  DivNode *createBatchedReduceMean(llvm::StringRef name, NodeValue batch,
-                                   unsigned_t axis);
+  /// with output type \p outTy with nodes: a) AveragePool, or b) three nodes:
```
jfix71: This comment now seems wrong. I would just say it creates and returns a `BatchedReduceMeanNode`. If you really want to you could mention that it may be lowered to three three nodes, or may be optimized into an AvgPool, but I don't think it's necessary.

docs/Optimizations.md
```diff
@@ -67,6 +67,12 @@ Below you can see the list of currently supported graph optimizations:
     This optimization performs a classic CSE with a goal of avoiding of any
     results that were computed already.
 
+  * Optimization of ReduceMean nodes
```
nadavrot: Why is this an optimization? Why would one form be better?

tests/unittests/OperatorTest.cpp
```diff
@@ -580,6 +580,114 @@ TEST_P(InterpAndCPU, batchedReduceMeanQuantizedWithAxis) {
...
+  auto *R = F_->createBatchedReduceMean("reduce.mean", batch, {2, 3}, true);
+
+  auto *save = F_->createSave("save", R);
+  auto *result = ctx_.allocate(save->getPlaceholder());
+  EE_.compile(CompilationMode::Infer, F_);
+
+  EE_.run(ctx_);
+  auto H = result->getHandle();
+
+  // 1st reduce has i*j*k elements for a 4D (i,j,k,l) input & l is reduced
```
nadavrot: Period.

tests/unittests/OperatorTest.cpp
```diff
@@ -580,6 +580,114 @@ TEST_P(InterpAndCPU, batchedReduceMeanQuantizedWithAxis) {
...
+        res += tmpRes[dims[2] * index + k];
+      }
+      results[index] = res / dims[2];
+
+      EXPECT_NEAR(H.at({i, j, 0, 0}), results[index], 0.001);
+    }
+  }
+}
+
+TEST_P(InterpAndCPU, batchedReduceMeanUsingAvgPoolQuantized) {
```
nadavrot: Please use doxygen comments to comment this method.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -277,6 +277,152 @@ TEST(onnx, importAveragePool3D) {
...
+      1.5, 3.5, 5.5, 7.5, 9.5, 11.5, 13.5, 15.5,
+  };
+
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
+  for (size_t i = 0; i < 8; i++) {
+    EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
+  }
+}
+
+TEST(onnx, reduceSum4D) {
```
nadavrot: Doxygen comments please.

docs/Optimizations.md
vuzelac-cadence: At least in our case, Reduce Mean is implemented using Average Pool. Originally I considered having it completely controlled by the backend, but based on @jfix71 's   comments in #2343, we took the recommended, shorter route and considered this as an always-to-do optimization.

nadavrot:
@vuzelac-cadence Thank you for doing this work. I was wondering, have you considered lowering the reduceMean node in the loader? What are the advantages of the late lowering? Why is lowering to AvgPool a good transformation? And, why can't we pattern match the batchedAdd-Splet-Div in the optimizer instead of adding a new node?

docs/Optimizations.md
jfix71: I consider it an optimization because otherwise we lower `BatchedReduceMean` to multiple nodes -- i.e. it's more efficient to execute an `AvgPool` than to lower it/execute it via a `BatchedReduceAdd` followed by a `Div`. 

But it really is dependent on what the backend supports -- if a backend has an efficient `BatchedReduceMean` implementation then they would likely prevent lowering on the node and add a backend-specific `BatchedReduceMeanInst`. At that point perhaps they wouldn't want this `ReduceMean -> AvgPool` optimization enabled. But this just another area where we'd want better backend control over graph optimizations (see https://github.com/pytorch/glow/issues/1641).

vuzelac-cadence:
@nadavrot - it was originally lowered in the Importer and that seemed wrong, based on the description  in #2343.  Simply, different backends will take differently on ReduceMean so they need to be aware of it.
Pattern match seemed cumbersome and unneeded. 
But also, fundamentally, it seemed wrong, to me at least :),  that the Importer combines things all the way from ONNX parsing to lowering... and doing pattern matching instead of received the operations/nodes.

As for the optimization, please see the comment above about this. It was just an option we considered, and accepted it finally, based on discussion in #2343.

nadavrot:
@vuzelac-cadence @jfix71  Sounds good.  This LGTM. 

About the ONNX-IFI loader, I agree that the loader should not be doing lowering, unless they are specific to the semantics of the ONNX-spec. Do you have examples of lowering that happens inside the loader that could be moved to the core of the compiler?

vuzelac-cadence:
I might be looking into Importer to be more or less like a parser, ignoring it's "Loader" functionality, which is from where I draw these, possibly wrong, conclusions. However, my general idea is that Parser and Loader could have been split, as loader likely has not input format related dependencies.

1. Here, I would have expected this to be a part of Graph, not Importer. So this one doesn't just combine the Importer/Loader but adds Graph specific things :
https://github.com/pytorch/glow/blob/63265f2cb458bce136015621981af1fefdcb8e51/include/glow/Importer/CommonOperatorLoader.h#L641-L646

2. Here, I would have expected `calculateConvPoolOutputDims` to be within the Graph, not Importer. Not sure why would Import/Loader know about this. Furthermore, loaders would have to keep replicating this logic. Then this comment:  `Int8Maxpool output quantization should be same as the input ` doesn't seem to be Caffe2 format related (?), in order to be in the Caffe2 Loader. So maybe something to be within Graph as well ? 
https://github.com/pytorch/glow/blob/63265f2cb458bce136015621981af1fefdcb8e51/lib/Importer/Caffe2ModelLoader.cpp#L442-L449

3.  `getNodeValueOrCreateConstantByName() `,  ubiquitously used across the loaders, could have been placed in Graph as well, or in a separate Loader (if Graph is not to build the connectivity between nodes). If there was a separate loader, the function above would not replicate in both ONNX and Caffe2 Importers. Similarly for `addNodeAsOutput()` ...

All in all, this is all nitpicking and please feel free to ignore this much of text. Especially since I likely just failed to correctly understand various architectural details...

tests/unittests/OperatorTest.cpp
vuzelac-cadence: Also fixed similar mistakes in other test files in this PR.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -619,8 +619,7 @@ class CommonOperatorLoader : public ProtobufLoader {
     return llvm::Error::success();
   }
 
-  llvm::Error loadReduceMeanOrSum(llvm::StringRef typeName, const OpType &op,
-                                  ArgumentDictionaryTy &dict) {
+  llvm::Error loadReduceSum(const OpType &op, ArgumentDictionaryTy &dict) {
```
jfix71: If we move `keepDims` into `createBatchedReduceAdd()` like you did for `createBatchedReduceMean()`, can't we keep these two cases merged together? I think the logic would again be the same and so it's better than duplicating the case.

vuzelac-cadence: That would require some more implementation changes, and testing that we didn't consider as they were not on our path. Also, we were trying focusing on additions, not modifications of the original code, thus the split, to keep ReduceSum as is. If you want that change as a part of this PR I can do that. Let me know.

tests/unittests/OperatorTest.cpp
```diff
@@ -580,6 +580,117 @@ TEST_P(InterpAndCPU, batchedReduceMeanQuantizedWithAxis) {
...
+
+  auto *save = F_->createSave("save", R);
+  auto *result = ctx_.allocate(save->getPlaceholder());
+  EE_.compile(CompilationMode::Infer, F_);
+
+  EE_.run(ctx_);
+  auto H = result->getHandle();
+
+  // 1st reduce has i*j*k elements for a 4D (i,j,k,l) input & l is reduced.
+  std::vector<float> tmpRes(sizeof(float) * dimAllMul);
```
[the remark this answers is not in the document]
vuzelac-cadence: Seems that I didn't cleanup & improve this test after getting it to work. Was giving less importance to test organization/effectiveness, and will fix that...

include/glow/Importer/CommonOperatorLoader.h
jfix71: Yes, I would say change it here in this PR. Our unit tests should make sure other things didn't break (I hope... 😃).

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
vuzelac-cadence: @jfix71, ok, my initial idea was to introduce randomized dimensions sizes, which would not fit array's const initializers requirements (and I don't believe we would want to allocate memory on heap here ?). While I don't have randomization included in this PR, this would allow adding it by changing `dims() `variable initialization only.  Is this a good enough explanation or you still want this simplified ?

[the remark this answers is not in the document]
jfix71: I'd rather it be simplified down to a 4D float array, thanks @vuzelac-cadence.

include/glow/Importer/CommonOperatorLoader.h
vuzelac-cadence: Done

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
vuzelac-cadence: Done

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1062,6 +1062,52 @@ static void mergeBatchedAdd(Function *F) {
...
+      std::vector<unsigned_t> kernels = {
+          static_cast<unsigned_t>(in->dims(0)[2]),
+          static_cast<unsigned_t>(in->dims(0)[3])};
+      std::vector<unsigned_t> strides = {1, 1};
+      std::vector<unsigned_t> pads = {0, 0, 0, 0};
+
+      // In Glow, AvgPool expects NHWC.
+      auto *TR1 =
+          F->createTranspose(RM->getName().str() + ".transposeNCHW2NHWC",
+                             RM->getNthInput(0), NCHW2NHWC);
```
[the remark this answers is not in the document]
jfix71: @vuzelac-cadence You missed this one -- can now replace it with `in`.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1062,6 +1062,51 @@ static void mergeBatchedAdd(Function *F) {
...
 }
 
+/// Optimize ReduceMean configuration with AvgPool if possible: last two axes
+/// in a 4D input must be reduced, and with the original shape kept.
+static void optimizeReduceMean(Function *F) {
+  auto &nodes = F->getNodes();
+
+  // For each node:
+  for (auto &node : nodes) {
+
```
jfix71: mega nit: remove new line here 😄

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2639,3 +2639,27 @@ TEST_F(GraphOptz, QuantizedFC) {
   // placeholders just aren't hooked up the the Graph F_.
   EXPECT_EQ(mod_.getPlaceholders().size(), 4);
 }
+
+/// Test batchedReduceMean optimization using AvgPool.
+TEST_F(GraphOptz, convertReduceMean2AvgPool) {
+  const size_t dims[] = {2, 2, 2, 2};
+
+  Node *A = mod_.createPlaceholder(ElemKind::FloatTy, dims, "input", false);
+  Node *R = F_->createBatchedReduceMean("reduce.mean", A, {2, 3}, true);
```
jfix71: nit: `/* keepDims */ true`

include/glow/Graph/Graph.h
```diff
@@ -566,25 +566,25 @@ class Function final : public Named {
...
-
-  /// Implements a batched reduce mean of the \p batch on the provided \p axis
-  /// with three nodes: a BatchedReduceAdd followed by a DivNode with a
-  /// SplatNode of the length of the \p axis dimension. \returns the final
-  /// DivNode.
-  DivNode *createBatchedReduceMean(llvm::StringRef name, NodeValue batch,
-                                   unsigned_t axis);
+                                               llvm::ArrayRef<unsigned_t> axes);
+
+  /// Create a node, performing BatchedReduceMean operation
```
jfix71: Can you make this doxygen style, references the parameters?

include/glow/Graph/Graph.h
```diff
@@ -566,25 +566,25 @@ class Function final : public Named {
 
   BatchedReduceAddNode *createBatchedReduceAdd(llvm::StringRef name,
                                                NodeValue batch,
-                                               unsigned_t axis);
+                                               llvm::ArrayRef<unsigned_t> axes,
+                                               bool keepDims = false);
```
jfix71: I see there weren't doxygen comments here before, but would you mind adding them here too?  and to the next `createBatchedReduceAdd()` too? I think it should be almost identical to those for `BatchedReduceMean`.

lib/Graph/Graph.cpp
```diff
@@ -1346,37 +1374,40 @@ BatchedReduceAddNode *Function::createBatchedReduceAdd(llvm::StringRef name,
...
-                                                       NodeValue batch,
-                                                       unsigned_t axis) {
-  auto outDims = getNewShapeWithoutAxis(batch.dims(), axis);
+BatchedReduceAddNode *
+Function::createBatchedReduceAdd(llvm::StringRef name, NodeValue batch,
+                                 llvm::ArrayRef<unsigned_t> axes,
+                                 bool keepDims) {
+
+  ShapeVector outDims(batch.dims().begin(), batch.dims().end());
+  outDims = keepDims ? getNewShapeReducedAxes(outDims, axes)
```
jfix71: I don't think this is the preferred way to handle `keepDims`. This encodes it into the output shape. This means the Interpreter's kernel, LLVM IRGen, etc. would need to inspect the output shape and behave differently based on it.

Instead I think we should create the output shape assuming `keepDims` is false, and then add a reshape to the output if `keepDims` is true. So you would always create the `BRA`/`BRM` with the shape without axes, and then if `keepDims == true` then add in a reshape here using the reduced axes shape.

Does this seem reasonable? The optimizer will get rid of any extra inserts/copies, so there's no performance cost, and it allows there to be a single canonical form of `BRA`/`BRM`.

docs/Optimizations.md
```diff
@@ -67,6 +67,12 @@ Below you can see the list of currently supported graph optimizations:
...
+
+    This optimization performs substitions of ReduceMean with AvgPool node if
+    the reduce parameters are suitable: input is 4D with last two dimensions
+    to be reduced, and keepdims parameter is true.
```
jfix71: Is it actually necessary for keepDims to be true? You can always perform the optimization and just add a reshape. It doesn't look like this is reflected in the optimization code itself either.

lib/Graph/Graph.cpp
vuzelac-cadence: > This means the Interpreter's kernel, LLVM IRGen, etc. would need to inspect the output shape and behave differently based on it.
I thought that has to be done anyway - A node is created with `TypeRef Result` argument which can take any value, as it can be set to just about anything that's compatible,  through that other `createBatchedReduceAdd` version that takes  `TypeRef outTy ` argument.

docs/Optimizations.md
vuzelac-cadence: Idea was indeed to do the optimization only if keepDims is true. That was ReduceMean we had problems with and were trying to change only what's required. But also, that was making code simpler - no keepDims passed to this node and no need to reshape and remove dimensions (since AvgPool keeps dimensions).
I can add `keepDims `to `BatchedReduceMeanNode` specification, and add Reshape that erases two dimensions, into this optimization.

vuzelac-cadence:
I have  accepted all the recommend changes. BRM/BRA now have a single canonical form where reduced shapes are always removed. For that to work, optimized ReduceMean  drops two axes and lowered ReduceMean does no reshape.  Reshape for both BRM/BRA is now back into Importer just that it works with multiple axes. This now looks closer to the original code. Let me know if I should squash all this commit mess now.

include/glow/Graph/Graph.h
```diff
@@ -564,27 +564,31 @@ class Function final : public Named {
   Node *createParallelBatchMatMul(llvm::StringRef name, NodeValue lhs,
                                   NodeValue rhs);
 
+  /// Create a node, performing BatchedReduceAdd operation. Output type
+  /// matches input \p outTy type.
```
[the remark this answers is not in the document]
vuzelac-cadence: Thank you for very insightful guidance here...

PR #2467, opened 2019-03-01
Make static methods for Backend

jackm321:
*Description*:
`Backend::transformPostLowering`, `Backend::isOpSupported`, and `Backend::shouldLower`, don't need a backend instance to operate so create a static version of each of these functions so that the ONNXIFI layer can call these functions for compatibility checking without needing an actual Backend. This is useful because in the near future HostManager will control all of the backends and they won't be cleanly accessible to ONNXIFI code.

*Testing*:
CI

*Documentation*:
Comments

### DIFF
--- a/include/glow/Optimizer/Optimizer.h
+++ b/include/glow/Optimizer/Optimizer.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_OPTIMIZER_OPTIMIZER_H
 #define GLOW_OPTIMIZER_OPTIMIZER_H
 
+#include "glow/Backends/Backend.h"
 #include "glow/Backends/CompilationOptions.h"
 #include "glow/Quantization/Quantization.h"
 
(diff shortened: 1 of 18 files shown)


beicy:
Btw, all CI failed with the error : "../lib/Optimizer/Lower.cpp:805: undefined reference to `glow::shouldLowerStatic(glow::BackendKind, glow::Node const*)'.  Is that expected at this stage?

jackm321:
@beicy thanks, I think it's due to a circular dependency that this introduces. I fixed one such issue earlier by removing `Backends` dependency on `Backend` but I need to also change this. I just added th work in progress tag and I'll try to fix this again soon.

tools/loader/Loader.cpp
```diff
@@ -267,7 +267,7 @@ void Loader::compile(Context &ctx) {
     // generateNodeQuantizationInfos() when writing out the profile, allowing
     // for both lowered and unlowered NodeValues to find their quantization
     // parameters.
-    ::lower(F_, &loweredMap_, /* backend */ nullptr,
+    ::lower(F_, &loweredMap_, /* backend */ nullptr, /* backendKind */ nullptr,
```
nadavrot: Why do you need both the backend and the backend kind? Can you get rid of one of the parameters?

include/glow/Optimizer/Optimizer.h
```diff
@@ -38,14 +39,18 @@ void optimize(Function *F, const CompilationOptions &opts);
...
+/// linear algebra operators. If \p backend is not a nullptr then it can prevent
+/// lowering of a node via \ref Backend::shouldLower(); likewise if \p
+/// backendKind is not null then Backend::shouldLowerStatic can prevent
+/// lowering; otherwise everything will be lowered. At most of one of backend
+/// and backendKind should be non-null. If \p loweredMap is not a nullptr, then
+/// \p loweredMap will contain a mapping from output names of the nodes found
+/// and lowered in \p F to the output names of the nodes they were lowered from
+/// along with the NodeKind. \p doNotLowerKinds is a set of NodeKinds which
+/// represents all Nodes that should not be lowered.
+void lower(Function *F, LoweredInfoMap *loweredMap,
```
nadavrot: The lower method now has 5 parameters. Do we need all of the parameters? Can we refactor and clean things up here?

jackm321: Yeah it would be good to refactor but I'm not sure what's the best thing to do.

It would be good if we could just use `BackendKind` in `lower` and then call the `shouldLowerStatic()` but we have some one off Backends in unit tests that don't have a BackendKind and implement their own `shouldLower()` methods for example `MockBackendUnloweredFC` so that won't work for all cases.

We could create two lower methods, one that uses backend for unit tests and one that uses backend kind for everything else but I'm not sure that's better. What do you think?

beicy: I prefer the first one. For `MockBackendUnloweredFC`, can we add a `shouldLower()` method simply return true or false?

jackm321:
I'm going to close this for now. I want to wait to see about two things: 1. how changes in lowering play out 2. how backend registry is implemented before picking this back up.

PR #2473, opened 2019-03-02
Make (Fused)RowwiseQuantizedSparseLengthsWeightedSum match Caffe2 using float offset/quantization scheme.

jfix71:
*Description*: Support (Fused)RowwiseQuantizedSparseLengthsWeightedSum with float offsets instead of int32, which matches the Caffe2 implementation ([see here](https://github.com/pytorch/pytorch/blob/db121375e7c07d43f48cfbc2a57e39e5f16e7d3b/caffe2/perfkernels/fused_8bit_rowwise_embedding_lookup.cc#L62-L68) for the C2 fused implementation)

I changed `Int8FusedQTy` to `UInt8FusedQTy`. It's assumed that `UInt8FusedQTy` uses float offsets.

Additionally I added a new fill type `GivenTensorByteStringToUInt8Fill` which is loaded and converted into `Int8QTy`. If a `FusedRowwiseQuantizedSparseLengthsWeightedSum` uses it then it converts it to `UInt8FusedQTy`. I left `RowwiseQuantizedSparseLengthsWeightedSum` using `Int8QTy`, meaning it still needs to convert to uint inside the kernel, since I didn't want to add a `UInt8QTy` right now.

*Testing*: Updated Caffe2ImporterTests and their protos. Added Caffe2Importer test for correctly loading `GivenTensorByteStringToUInt8Fill`. Updated `TensorsTest::initZeroFused` given move to `UInt8FusedQTy`.

*Documentation*: Added docs.


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -942,28 +942,22 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
-        // Must memcpy to the stack and back to avoid misaligned addresses.
-        char *currRowOffsetPtr = dataBasePtr + (i + 1) * width - 4;
-        float fOffset;
-        memcpy(&fOffset, currRowOffsetPtr, 4);
-        int32_t iOffset = static_cast<int32_t>(fOffset);
-        memcpy(currRowOffsetPtr, &iOffset, 4);
+        // We also need to update the data to be unsigned instead of signed.
+        auto dataCH = dataC->getHandle<int8_t>();
+        for (size_t i = 0, e = dataCH.size(); i < e; i++) {
+          dataCH.raw(i) = (uint8_t)dataCH.raw(i) + OFFSETSHIFT;
```
bertmaher: I'm a bit worried about potential UB here since dataCH.raw(i) is signed, and we might be trying to store an out-of-range uint8_t in it.

include/glow/Quantization/Base/Base.h
```diff
@@ -98,6 +97,30 @@ inline float dequantize(eTy input, const TensorQuantizationParams &TQP) {
...
 }
 
+/// Converts floating point value to DestTy (quantized type) based on the
+/// quantization parameters \p scale and \p offset. If the dest type is int8_t,
+/// then an offset of 128 is substracted to convert to int8_t.
+template <class DestTy>
+inline DestTy quantizeWithFloatOffset(float input, float scale, float offset) {
+  uint8_t d = static_cast<uint8_t>((input - offset) / scale);
+  if (std::is_same<int8_t, DestTy>::value) {
+    d -= 128;
```
bertmaher: Hmm... this could overflow the uint (which is ok), but then if the result is assigned to an int, it's also maybe UB.  Can you check this too?

yinghai:
BTW, does it pass our internal CI?

lib/Importer/Caffe2ModelLoader.cpp
jfix71: This was actually incorrect -- the Tensor should have been `uint8_t` (I changed the Constant but not the payload Tensor). And it wasn't an issue because when we generate the runtime bundle the type is just taken from the Constant and not the Tensor. Anyway, I fixed it and added in a verification check in `Function::verify()`.

include/glow/Quantization/Base/Base.h
jfix71: According to Integral conversions [in the spec](https://en.cppreference.com/w/cpp/language/implicit_conversion) this is "implementation-defined", but not undefined. It also passes all checks using UBSan locally/on CI.

jfix71:
> BTW, does it pass our internal CI?

I'll check tomorrow! For now I have verified that the new protos here for the `Caffe2ImporterTest` run correctly on both C2 and Glow.

bertmaher:
I don't know how I managed to trigger UB doing something similar recently.  Maybe I actually hit signed overflow (somehow?) and thought it was due to unsigned->signed conversion.  Oh well.  LGTM!

bertmaher:
Hah, okay so I dug up the UBSAN error I had gotten, and I'm pretty sure this diff is immune to it, but for posterity it was assigning an out-of-range float to an int, e.g. `int8_t xx = 240.0f;`  (that will fail to compile of course, but if it happens at runtime it's also a UBSAN error "value is outside the range of representable values of type 'int'".

jfix71:
Confirmed that it works on internal CI.

Issue #2396, opened 2019-02-14
[unittests] Cannot assume fp32 support on all backends


jfix71:
The OperatorTests below compare reduced precision results (fp16 or quantized) to fp32 results. These tests get the fp32 result from the backend that is being tested, but the backend may not support fp32. The fp32 result should instead be evaluated on the Interpreter.

We can probably refactor `compareAgainstInterpreter()` from `BackendCorrectnessTest.cpp` in order to automatically do this. This would also allow us to avoid explicitly picking quantization parameters which we do in some of these tests.

The flow would be (1) instrument profile the graph; (2) compile/run on the interpreter, gathering the fp32 result along with the quantization profile; (3) quantize the graph using that profile; (4) compile/run on the backend, gathering the reduced precision result to compare against the fp32 result from the interpreter.

Edit: Looks like we're doing something like this already in `testQuantizationEnd2End()`.

- QuantizeAndDequantizeFPtoI8
- QuantizedTranspose
- Int8Tanh
- Int8Sigmoid
- Int8ConvolutionDepth10
- Int8ConvolutionDepth8
- IntFC
- IntConcat
- Int8Log
- rowwiseQuantizedFCTest
- FP16ConvolutionDepth10
- FP16ConvolutionDepth8

Issue #2464, opened 2019-02-28
FunctionQuantizer silently skipping quantizing nodes


jfix71:
Right now the `FunctionQuantizer` silently skips quantizing a node if the target backend does not support the quantization precision being converted to (checked via `Backend::isOpSupported()`). Instead it leaves the node in fp32 and wraps it with dequantize/quantize nodes.

Due to this, I've noticed that some of our unit tests that are intended to test precision of quantized graphs are instead just wrapping many of the graph's nodes with dequantize/quantize. This kind of defeats the purpose of the test. A couple examples are QuantizationTest's `JIT/Operator.end2endInt16/0` and `OpenCL/Operator.end2endGRU/0`. We've also had occasional regressions where nodes previously quantized no longer were, requiring later fixes, e.g. https://github.com/pytorch/glow/pull/2436.

Instead, I think we should at least have an option for the `FunctionQuantizer` to report quantization failure up the stack if the target backend does not support a node it is trying to quantize. If a test or model expects a NodeKind to not be quantized, we already have `doNotQuantizeKinds` which can be used to prevent the `FunctionQuantizer` from returning failure in such cases and explicitly allow for specific NodeKinds to stay unquantized.

Another option could be to remove this `isOpSupported()` check in the `FunctionQuantizer` entirely, and then after quantization check if the graph is supported by the backend. However this is a bit more drastic. We could still use `doNotQuantizeKinds` in this case though, and this would probably mean we could remove `NodeInfo` from `isOpSupported()` (seen in https://github.com/pytorch/glow/pull/2235).

PR #2474, opened 2019-03-02
[unittests] Refactor reduced precision tests to prevent backends from needing to execute fp32

jfix71:
*Description*: Some of our unit tests were executing a graph that included both fp32 and reduced precision nodes on the backend to compare their execution. However, not all backends support fp32. 

This PR refactors `compareAgainstInterpreter()` to support executing a graph on both the Interpreter and some other backend, including profiling if we need to quantize, and comparing the results.

This should simplify adding tests for new operators into the future, as you only need to specify the fp32 graph, and then can call into `compareAgainstInterpreter()` to get comparisons of that graph in any precision desired and compared to any backend desired.

*Testing*: Updated all unittests that are listed in https://github.com/pytorch/glow/issues/2396. This included adjusting some allowed errors given we use new quantization parameters, and disabling some quantized tests on backends that were not being quantized at all in the first place (see https://github.com/pytorch/glow/issues/2464).

Fixes https://github.com/pytorch/glow/issues/2396

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -267,6 +267,12 @@ enum class ElemKind : unsigned char {
   BoolTy,        // Bool type (bool)
 };
 
+/// \returns whether \p e is a quantized ElemKind.
+inline bool isQuantizedElemKind(ElemKind e) {
+  return e == ElemKind::Int8QTy || e == ElemKind::Int16QTy ||
+         e == ElemKind::Int32QTy || e == ElemKind::UInt8FusedQTy;
+}
+
 /// A class that represents a type of a tensor.
 struct Type final {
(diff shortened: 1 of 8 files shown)


nickgg:
@jfix71 I think this might have broken the SHARED build, I'm seeing a bunch of undefined symbols for libBackendTestUtils

jfix71:
@nickgg It looks like everything is passing on master CI right now -- can you point me to where you're seeing issues?

nickgg:
@jfix71 on a local shared debug build. I wonder if the CI is passing because certain backends are disabled? Do you have a shared build locally?

jfix71:
@nickgg https://github.com/pytorch/glow/pull/2498 should fix it, sorry about that!

PR #2476, opened 2019-03-04
[unittests] Enable each test case for specific backends in OperatorTest.cpp.

### LINKED ISSUE #2170: Improve backend parameterization in OperatorTest
The parameterization of tests by backend in OperatorTest is kind of confusing.  To add a new backend (which of course passes only a subset of tests) I had to think really hard about how to create a new test subset.  Is there something better we could do here?

d1jang:
*Description*:

What we'd need is to specific backends that each test case is enabled for, however, GTEST is incapable of parameterizing each test case separately since GTEST's parameterization only works per harness.
We could create a new harness for a test case to achieve that, but it's too verbose. This diff proposes a compromise:
We put all test cases under a single harness instantiated for all possible backends -- as a result all test cases will be running for all backends. To whitelist specific backends for each specific test case, ENABLED_BACKENDS(backend1, backend2, ...) is used as a guard for each test to enable it only for specific backends.

*Testing*:

Ran OperationTest to see it passes.

*Documentation*:

Fixes #2170

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -34,7 +34,7 @@ public:
   ~Operator() override { mod_.clear(); }
 
 protected:
-  bool isEnabledBackend(const std::set<BackendKind>& enabledBackends) {
+  bool isEnabledBackend(const std::set<BackendKind> &enabledBackends) {
     return enabledBackends.find(GetParam()) != enabledBackends.end();
   }
 
@@ -59,7 +59,7 @@ INSTANTIATE_TEST_CASE_P(AllOperatorTest, Operator,
 #endif // GLOW_WITHOPENCL
                             BackendKind::Interpreter));
 
-// TODO(d1jang): Replace return for GTEST_SKIP() so that skipped tests are
+// TODO: Replace return for GTEST_SKIP() so that skipped tests are
 // correctly reported once the macro gets available.
 #define ENABLED_BACKENDS(...)                                                  \
   if (!isEnabledBackend({__VA_ARGS__}))                                        \
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -26,6 +26,7 @@
 #include "llvm/Support/raw_ostream.h"
 
 using namespace glow;
+
 class Operator : public ::testing::TestWithParam<BackendKind> {
 public:
   Operator() : mod_(EE_.getModule()) { F_ = mod_.createFunction("main"); }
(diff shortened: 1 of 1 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -26,22 +26,44 @@
...
+#ifdef GLOW_WITH_CPU
+                            BackendKind::CPU,
+#endif // GLOW_WITH_CPU
+#ifdef GLOW_WITH_OPENCL
+                            BackendKind::OpenCL,
+#endif // GLOW_WITHOPENCL
+                            BackendKind::Interpreter));
 
-class InterpOnly : public Operator {};
+// TODO(d1jang): Replace return for GTEST_SKIP() so that skipped tests are
```
jfix71: We usually don't add names to TODOs.

tests/unittests/OperatorTest.cpp
```diff
@@ -26,22 +26,44 @@
...
 using namespace glow;
+
 class Operator : public ::testing::TestWithParam<BackendKind> {
 public:
   Operator() : mod_(EE_.getModule()) { F_ = mod_.createFunction("main"); }
 
   ~Operator() override { mod_.clear(); }
 
 protected:
+  bool isEnabledBackend(const std::set<BackendKind>& enabledBackends) {
```
jfix71: CI is failing on this line because of clang-format

tests/unittests/OperatorTest.cpp
```diff
@@ -26,22 +26,44 @@
...
   Module &mod_;
   Function *F_;
   Context ctx_;
 };
 
-class InterpAndCPU : public Operator {};
+// The parameterized test suite is instantiated with all available backends.
+INSTANTIATE_TEST_CASE_P(AllOperatorTest, Operator,
+                        ::testing::Values(
+#ifdef GLOW_WITH_CPU
```
jfix71: If we move to using this `ENABLED_BACKENDS()` pattern in other similarly parameterized tests (e.g. `QuantizationTest`, `BackendCorrectnessTest`, etc.), then perhaps we should macro this declaration of all backends and place it in `BackendTestUtils.h`, so we don't need to ifdef etc. for every backend in every test that uses this. Also then maybe we'd move `isEnabledBackend()` and `ENABLED_BACKENDS()` there too.

tests/unittests/OperatorTest.cpp
d1jang: Removed.

tests/unittests/OperatorTest.cpp
d1jang: clang-format applied.

tests/unittests/OperatorTest.cpp
d1jang: Sounds like a good plan. Can we check in this change first and apply it in a follow-up PR? An incremental change would be easier to make.

Issue #2460, opened 2019-02-27
Stack corruption in child thread on windows.


ayermolo:
Seems like after recent changes by @nickgg there are now failures in multiple tests. First appeared in DeviceManagerTest, and after todays changes extended to GradCheckTest.
Full failure list. Not sure if it's same issue for others:
1 - BackendCorrectnessTest (OTHER_FAULT)
  	  2 - BackendTest (OTHER_FAULT)
  	  6 - DeviceManagerTest (OTHER_FAULT)
  	  9 - GradCheckTest (OTHER_FAULT)
  	 14 - HyphenTest (OTHER_FAULT)
  	 19 - MLTest (OTHER_FAULT)
  	 21 - OperatorTest (OTHER_FAULT)
  	 25 - QuantizationTest (OTHER_FAULT)
  	 29 - TraceEventsTest (OTHER_FAULT)

It's a very strange problem. In llvm core.cpp when function Expected<SymbolMap> ExecutionSessionBase::legacyLookup
is called. A promise is created, but not used. When function exits, destructor is called, and when destructor exits VS reports corruption of stack.

@nickgg  any ideas?

nickgg:
Interesting, the change is just that we're using the DeviceManager in all tests now so it's likely the same root cause.

You say a promise is created but not used, is that an error on Windows for some reason?

Is there a log of the stacktrace you can attach? What's the object being deleted, is it the promise (the one on the stack in EE::execute)?

ayermolo:
Stack trace below. Well I think stack check is inserted by VS. 
As for promise it's internal one that LLVM uses in ExecutionSessionBase::legacyLookup.

```
 	DeviceManagerTest.exe!failwithmessage(void * retaddr, int crttype, int errnum, const char * msg) Line 213	C++
 	DeviceManagerTest.exe!_RTC_StackFailure(void * retaddr, const char * varname) Line 263	C++
 	DeviceManagerTest.exe!_RTC_CheckStackVars(void * frame, _RTC_framedesc * v) Line 71	C++
>	DeviceManagerTest.exe!std::promise<void>::~promise<void>() Line 1588	C++
[...]

PR #2477, opened 2019-03-04
[Win32] Remove std::promise<llvm::Error>

nickgg:
*Description*: On Windows the LLVM JIT doesn't always work when executed on a different thread than the compile (see discussion in #2460). Until we work out how to fix that, just have windows build run unit tests in the DummyDeviceManager.
*Testing*: unit tests when forcing the ExecutionEngine through both branches.
*Documentation*:

cc @ayermolo

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -186,13 +186,16 @@ void ExecutionEngine::insertCompiledFunction(
   functionMap[name] = func.get();
   compiledFunctions_[name] = std::move(func);
 
-  std::promise<llvm::Error> addPromise;
+  std::promise<void> addPromise;
   auto fut = addPromise.get_future();
+  llvm::Error addErr = llvm::Error::success();
   device_->addNetwork(&M_, std::move(functionMap),
-                      [&addPromise](const Module *, llvm::Error err) {
-                        addPromise.set_value(std::move(err));
+                      [&addPromise, &addErr](const Module *, llvm::Error err) {
+                        addErr = std::move(err);
+                        addPromise.set_value();
                       });
-  EXIT_ON_ERR(fut.get());
+  fut.wait();
+  EXIT_ON_ERR(std::move(addErr));
 }
 
 void glow::runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
--- a/examples/resnet-concurrent.cpp
+++ b/examples/resnet-concurrent.cpp
@@ -84,9 +84,9 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module) {
 
 /// Loads the CompliedFunction into device \p device.
 /// Returns a future which is completed when the device is initialized.
-std::future<llvm::Error> addToDevice(unsigned int id, DeviceManager *device,
-                                     Module &module, FunctionMapTy functions) {
-  auto compilePromise = std::make_shared<std::promise<llvm::Error>>();
+std::future<void> addToDevice(unsigned int id, DeviceManager *device,
+                              Module &module, FunctionMapTy functions) {
+  auto compilePromise = std::make_shared<std::promise<void>>();
   auto future = compilePromise->get_future();
 
   device->addNetwork(&module, functions,
@@ -94,11 +94,12 @@ std::future<llvm::Error> addToDevice(unsigned int id, DeviceManager *device,
                        if (err) {
                          llvm::errs() << "Failed to compile model for device "
                                       << id << ".\n";
+                         EXIT_ON_ERR(std::move(err));
                        } else {
                          llvm::outs()
                              << "Successfully added to Device " << id << ".\n";
                        }
-                       compilePromise->set_value(std::move(err));
+                       compilePromise->set_value();
                      });
 
   return future;
@@ -156,7 +157,7 @@ int main(int argc, char **argv) {
   FunctionMapTy functions;
   functions.emplace("resnet50", compiledFunction.get());
 
-  std::vector<std::future<llvm::Error>> compiles;
+  std::vector<std::future<void>> compiles;
   compiles.reserve(numDevices);
 
   for (unsigned int i = 0; i < numDevices; ++i) {
@@ -165,7 +166,6 @@ int main(int argc, char **argv) {
 
   for (auto &f : compiles) {
     f.wait_for(/* timeout_duration */ std::chrono::seconds(30));
-    EXIT_ON_ERR(f.get());
   }
 
   llvm::outs() << "Loading files from " << inputDirectory << "\n";
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -79,9 +79,9 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
   return backend->compile(F, opts);
 }
 
-std::future<llvm::Error> addToDevice(unsigned int id, DeviceManager *device,
-                                     Module &module, FunctionMapTy functions) {
-  auto compilePromise = std::make_shared<std::promise<llvm::Error>>();
+std::future<void> addToDevice(unsigned int id, DeviceManager *device,
+                              Module &module, FunctionMapTy functions) {
+  auto compilePromise = std::make_shared<std::promise<void>>();
   auto future = compilePromise->get_future();
 
   device->addNetwork(&module, functions,
@@ -89,11 +89,12 @@ std::future<llvm::Error> addToDevice(unsigned int id, DeviceManager *device,
                        if (err) {
                          llvm::errs() << "Failed to compile model for device "
                                       << id << ".\n";
+                         EXIT_ON_ERR(std::move(err));
                        } else {
                          llvm::outs()
                              << "Successfully added to Device " << id << ".\n";
                        }
-                       compilePromise->set_value(std::move(err));
+                       compilePromise->set_value();
                      });
 
   return future;
@@ -128,7 +129,6 @@ int main(int argc, char **argv) {
 
     auto f = addToDevice(i, devices[i], module, functions);
     f.wait_for(/* timeout_duration */ std::chrono::seconds(30));
-    EXIT_ON_ERR(f.get());
   }
 
   auto image =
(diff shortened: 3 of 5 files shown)


ayermolo:
Hi Nick. Thanks for looking in to this.

Hmm. I ported to master branch, and still see runtime assertion.

Wrote crash dump file "C:\Users\ayermolo\AppData\Local\Temp\image-classifier.exe-60c6ba.dmp"
0x00007FF7514E13D4 (0x000000CC241FC600 0x0000000000000400 0x00007FF753A4CB98 0x00007FFF78017097), failwithmessage() + 0x234 bytes(s), d:\agent\_work\1\s\src\vctools\crt\vcstartup\src\rtc\error.cpp, line 213 + 0x13 byte(s)
0x00007FF7514E103A (0x000000CC241FCC28 0xCCCCCCCCCCCCCCCC 0x0000000000000000 0x00007FF7535970D0), _RTC_StackFailure() + 0xAA bytes(s), d:\agent\_work\1\s\src\vctools\crt\vcstartup\src\rtc\error.cpp, line 263
0x00007FF7514DF972 (0x000000CC241FCB10 0x0000000000000000 0x0000000000000000 0xCCCCCCCCCCCCCCCC), _RTC_CheckStackVars() + 0x62 bytes(s), d:\agent\_work\1\s\src\vctools\crt\vcstartup\src\rtc\stack.cpp, line 71
0x00007FF750503B29 (0x000000CC241FCC28 0x000000CC241FCDB8 0x0000000000000000 0x000000CC241FCE68), std::promise<void>::~promise<void>() + 0x139 bytes(s), c:\program files (x86)\microsoft visual studio\2017\professional\vc\tools\msvc\14.16.27023\include\future, line 1588 + 0x1F byte(s)

nickgg:
Interesting, maybe we misdiagnosed this. Are we sure it's connected to this commit https://github.com/pytorch/glow/commit/80d3092e60a7a4adbc35073f761a635392e3c10b ?

Could it have been a change to the jit that went in before that?

ayermolo:
Hmm. It's possible. I think it did start failing around that time.

ayermolo:
@nickgg Any update, do you need help narrowing it down, or anything to move this forward?

nickgg:
@ayermolo are you seeing this in a clean checkout of Glow, or only in your merge/rebase build? This diff should disable the multithreaded execution, can you verify for me that a unit test like OperatorTest is using the main thread only?

ayermolo:
@nickgg This is clean checkout of glow.
Yeah that's a weird part. Looking closer in to it, it happens in main thread.
Did a bit more random poking around. If I remove all the promise stuff in ExecutionEngine::insertCompiledFunction problem goes away.

ayermolo:
Also changing to using shared_ptr seems to solve the issue. Maybe there is issue with VS and capturing promise in lambda by reference?
  std::shared_ptr<std::promise<runtime::ResultCode>> addPromise =
      std::make_shared<std::promise<runtime::ResultCode>>();
  auto fut = addPromise->get_future();
  device_->addNetwork(&M_, std::move(functionMap),
                      [addPromise](const Module *, runtime::ResultCode code) {
                        addPromise->set_value(code);
                      });

nickgg:
@ayermolo The part that waits for runFunction() to complete? We need to do something there to block until it finishes - maybe can use a condition variable instead.

nickgg:
> Also changing to using shared_ptr seems to solve the issue. Maybe there is issue with VS and capturing promise in lambda by reference?
std::shared_ptr<std::promiseruntime::ResultCode> addPromise =
std::make_shared<std::promiseruntime::ResultCode>();
auto fut = addPromise->get_future();
device_->addNetwork(&M_, std::move(functionMap),
[addPromise](const Module *, runtime::ResultCode code) {
addPromise->set_value(code);
});

Interesting! That is a good clue. Don't have an issue with that change at all, but can't really work out what the problem would be.

ayermolo:
Sorry that was a false positive. After I did clean rebuild even with shared ptr problem returned. Seems like only way is to remove all the references to promise.

nickgg:
Ok, pushed a version that removes the std::promises in favor of a condition variable. This should let us know if its just windows & promises are weird.

ayermolo:
@nickgg  Will check it out.
On side note, what version of LLVM are you using?

nickgg:
I am using LLVM 7.0 on osx.

ayermolo:
OK, cool. I am using same one.
On side note VS has issues with std::promise<llvm::Error> addPromise;
It reports an error that constructor in llvm::Error is protected.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -185,13 +195,16 @@ void ExecutionEngine::insertCompiledFunction(
   functionMap[name] = func.get();
   compiledFunctions_[name] = std::move(func);
 
-  std::promise<llvm::Error> addPromise;
-  auto fut = addPromise.get_future();
+  llvm::Error err = MAKE_ERR("Function did not run");
```
jackm321: This will create an unchecked Error because it will be ovewritten in the lambda below. Use a boolean instead

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -185,13 +195,16 @@ void ExecutionEngine::insertCompiledFunction(
...
   device_->addNetwork(&M_, std::move(functionMap),
-                      [&addPromise](const Module *, llvm::Error err) {
-                        addPromise.set_value(std::move(err));
+                      [this, &err](const Module *, llvm::Error err_) {
+                        err = std::move(err_);
+                        cv_.notify_one();
                       });
-  EXIT_ON_ERR(fut.get());
+  std::unique_lock<std::mutex> l(m_);
+  cv_.wait_for(l, 2s);
```
jackm321: What is the point of the wait for 2s?

nickgg: Stops the test from deadlocking if the function doesn't get called.

jackm321:
@ayermolo I'm curious if the issue you raised previously is just related to `std::promise<llvm::Error>`, do you want to try making that an `std::promise<void>`, creating an err like `llvm::Error err = llvm::Error::success()` and then passing that `err` in by reference to the lambda and see what happens?

ayermolo:
@jackm321 Are you referring to windows build issue using VS? That is directly related to instantiating std::promise with llvm::Error

```
4>C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\include\future(226): error C2248: 'llvm::Error::Error': cannot access protected member declared in class 'llvm::Error'
4>C:\llvm_source\llvm-7.0.0.src\include\llvm/Support/Error.h(172): note: see declaration of 'llvm::Error::Error'
4>C:\llvm_source\llvm-7.0.0.src\include\llvm/Support/FormatVariadicDetails.h(20): note: see declaration of 'llvm::Error'
....
4>C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\include\future(1398): note: while compiling class template member function 'std::promise<llvm::Error>::promise(void)'
4>C:\projects\glowExternalCadence\lib\Runtime\Provisioner\Provisioner.cpp(103): note: see reference to function template instantiation 'std::promise<llvm::Error>::promise(void)' being compiled
```

jackm321:
@ayermolo yes, so if we create a std::promise<void> do all problems go away here or just that one build problem?

ayermolo:
Well before it was std::promise<runtime::ResultCode> and the issue with llvm and stack corruption was there. Changing to std::promise<llvm::Error> now introduced build error.
If you want me to try piece of code do MR, and I'll integrate it in to my environment.
VS2017, latest updates
LLVM 7.0.0

nickgg:
Does the issue exist if instead of the promise holding anything, they are std::promise<void> ?

ayermolo:
@nickgg  Sorry not quite following.

nickgg:
sorry the formatting ate part of my question: if you replace `std::promise<ErrorCode>` with `std::promise<void>` and ignore the code, do you still see stack corruption?

ayermolo:
Changed in insertCompiledFunction to:

```
  std::promise<void> addPromise;
  auto fut = addPromise.get_future();
  device_->addNetwork(&M_, std::move(functionMap),
                      [&addPromise](const Module *, llvm::Error err) {
                        addPromise.set_value();
                      });
  fut.get();
```
Is that what you had in mind?

Build, rebuild, just in case, Operator tests are running and it exits ExecutionSessionBase::legacyLookup without issues. This is in debug mode.
I am starting to think VS implementation of template version of std::promise has an issue.

nickgg:
Yeah looks like it, but if the void version works that's an easy fix. I'll put up that change and we can see how it goes.

ayermolo:
Great thanks. With change to void I didn't even had to make it one thread with that change in ExecutionEngine.

nickgg:
@ayermolo cool! so this change is enough to unblock you?

ayermolo:
There are other places where this is used:
`std::promise<llvm::Error>`

ayermolo:
To clarify with this change OperatorTest passes. So this seems to fix run time exception.
Other places where llvm::Error with std::promise is used need to be changed to fix compilation error.

nickgg:
@ayermolo modified all instances of `promise<Error>` that I found, does that address your issue or does any templated promise break?

ayermolo:
@nickgg thanks for working on this. Almost :) Two left resnet-concurrent.cpp and tracing-compare.cpp in Examples

nickgg:
Not sure how I missed those, will fix today.

nickgg:
@jackm321 @ayermolo are we good with this version?

ayermolo:
@nickgg Looks good, thanks.

ayermolo:
Thanks @nickgg, @jackm321  

On side note. There might be potential issue...
The llvm ErrorSuccess is just a thin wrapper around llvm::Error so payload is nullptr. When LLVM project is created with CMAKE with build type Debug it enables LLVM_ENABLE_ABI_BREAKING_CHECKS. When we do = std::move(someError) it invokes   Error &operator=(Error &&Other).

This in turn calls assertIsChecked() and results in exception being thrown when it checks payload.

jackm321:
@ayermolo thanks, we're aware of this now, there's a few places where this happens and I need to go through and replace them with a checked initial value.

PR #2482, opened 2019-03-05
[unittests] Fix GradCheckTest to exercise operations being tested.

### LINKED ISSUE #1559: GradCheckTest must be more strict
This is follow up issue for PR #1447.

We have a numeric gradient check tests. One test case is supposed to check that gradient for one operator is implemented correctly. That is, if there is a bug in gradient, a test case must fail. However, this is not always true. For most test cases, if you replace gradient's body with no-op ("do nothing code"), test is still passing. This could be happening for various reasons, different for each individual test. Sometimes delta and epsilon are not strict enough. Sometimes net architecture is too "smart", so if one grad is not working, other grads take its responsibility (see PR #1447).

Need to consider test cases one by one, and make sure that the test actually fails if the gradient being tested is removed (in case of lowered grad) or replaced with empty body (in case of grad is implemented by Interpreter).

d1jang:
*Description*:

Some test cases in GradCheckTest do not exercise operations being tested properly. These tests were proved to compute zero gradients, which means that input doesn't affect output at all. With the change at 165, 4 out of 17 tests fail.

This can happen either when 1) the test case has a bug in wring the graph or 2) the training process trains variables unrelated to input to emulate output.

This PR fixes those tests so that they can pass the check at line 165: 3. They were fixed to make the operation to the input matter to the output after training. gradientCheckFCConcatRELU is removed since hidden weights introduced by FC completely ignores the input. Also it's unclear what operation gradientCheckFCConcatRELU is actually testing.

Note that a fundamental way to improve this test is to convert these tests as part of OperationTest so that we can test an operation and they differentiation independently from other operations and training algorithms. I will follow up with a PR to prototype this idea if you think it's useful.

*Testing*:

Ran the updated test to observe that they are passing.

*Documentation*:

Partially fixes #1559; Still there can be tests that pass without properly testing the target operations.

### DIFF
--- a/tests/unittests/GradCheckTest.cpp
+++ b/tests/unittests/GradCheckTest.cpp
@@ -129,6 +129,7 @@ void performGradCheck(ExecutionEngine &EE, Context &ctx, SaveNode *result,
 
   auto analyticalGradsH = gradVarTensor->getHandle();
   auto inputsH = inputs->getHandle<>();
+  bool allZeroGradients = true;
   for (size_t i = 0; i < analyticalGradsH.size(); i++) {
     auto old = inputsH.raw(i);
     // Calculate f(x+e):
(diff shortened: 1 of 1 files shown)


tests/unittests/GradCheckTest.cpp
```diff
@@ -360,8 +331,10 @@ TEST_P(InterpreterGrad, gradientCheckArithmetic) {
   auto inputsH = inputs.getHandle<>();
   auto outputsH = outputs.getHandle<>();
 
-  inputsH.initXavier(1, mod.getPRNG());
-  outputsH.initXavier(1, mod.getPRNG());
+  inputsH.randomize(-1, 1, mod.getPRNG());
+  outputsH.randomize(-1, 1, mod.getPRNG());
+  // inputsH.initXavier(1, mod.getPRNG());
```
artemrakhov: Please delete commented code.

tests/unittests/GradCheckTest.cpp
```diff
@@ -427,20 +400,18 @@ TEST_P(InterpreterGrad, gradientCheckFC) {
...
   Function *F = mod.createFunction("main");
   auto *A =
       mod.createPlaceholder(ElemKind::FloatTy, {1, numInputElem}, "A", false);
   auto *Exp = mod.createPlaceholder(ElemKind::FloatTy, {1, numOutputElem},
                                     "Exp", false);
-  auto *result = F->createSave("ret", A);
-  ctx.allocate(result->getPlaceholder());
 
-  Node *FA = F->createSigmoid("sig", Exp);
+  Node *FA = F->createSigmoid("sig", A);
```
artemrakhov: Good catch here and below

tests/unittests/GradCheckTest.cpp
```diff
@@ -149,49 +150,19 @@ void performGradCheck(ExecutionEngine &EE, Context &ctx, SaveNode *result,
...
 
+    if (numericGrad != 0.0) {
+      allZeroGradients = false;
+    }
     // Make sure that the analytical and numerical gradients agree.
     EXPECT_LE(err, allowedError);
   }
-}
 
-TEST_P(InterpreterGrad, gradientCheckFCConcatRELU) {
```
artemrakhov: I don't agree with removal of this test. This is the only test which executes gradient of Concat. Maybe the name is unfortunate? Do you see 0 gradients here even when you replace `initXavier()` by `randomize()`?

FYI, we have multiple `createFullyConnected()` constructors. Some of which accept W and B as arguments, some create them as hidden (but still accessible) vars. If you prefer to have W and B as constants, feel free to create&init them in test body and call this method of `F`:
https://github.com/pytorch/glow/blob/259111daf2def53c6d74dbd2d2d5e988bc5a9d3d/lib/Graph/Graph.cpp#L647

d1jang: Thanks. I've added gradientCheckConcat() to address this concern. To me gradientCheckFCConcatRELU is more like an integration test that shouldn't be here and reasoning about it very hard. gradientCheckConcat is trying to use as little operations as possible in addition to concat to minimize its complexity. Let me know what you think!

artemrakhov: Looks good!

d1jang:
* Please squash 4 commits into one.

Done

* I don't think that the issue #1559 can be closed by this PR. We definitely improved state of things, but some tests are still no-ops (in a sense that gradient check passes even with empty implementation).

Changed the description to reflect this. I'll create a more concrete issue on that and close #1559 according to that.

PR #2483, opened 2019-03-05
New DeviceManager config object

gcatron:
*Description*: Introduced new DeviceConfig struct to hold device configuration information for device managers. Changed factory and constructors to accept new struct instead of the current name parameter.
*Testing*: ninja test and resnet-runtime
*Documentation*: NA

### DIFF
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -106,8 +106,7 @@ int main(int argc, char **argv) {
 
   std::array<DeviceManager *, supportedBackends.size()> devices;
   for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i) {
-    devices[i] = DeviceManager::createDeviceManager(supportedBackends[i],
-                                                    "tracing-compare");
+    devices[i] = DeviceManager::createDeviceManager(supportedBackends[i]);
     devices[i]->init();
   }
 
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_RUNTIME_RUNTIMETYPES_H
 #define GLOW_RUNTIME_RUNTIMETYPES_H
 
+#include "glow/Backends/Backend.h"
 #include "glow/Backends/BackendUtils.h"
 #include "glow/Graph/Graph.h"
 
@@ -79,6 +80,20 @@ struct DAGNodeList {
   std::vector<std::unique_ptr<DAGNode>> roots;
   /// The non-root DAGNode pointers.
   std::vector<std::unique_ptr<DAGNode>> nodes;
+};
+
+/// This is the base class for DeviceManager configurations. Any specific
+/// device can extend this class to contain information to identify
+/// and configure the device manager. Additionally it needs to set it's kind_
+/// member variable to it's correct BackendKind.
+class DeviceConfig {
+  const BackendKind backendKind_;
+
+protected:
+  DeviceConfig(BackendKind kind) : backendKind_(kind) {}
+
+public:
+  BackendKind getBackendKind() { return backendKind_; }
 };
 
 } // namespace runtime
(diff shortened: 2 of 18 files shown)


include/glow/Runtime/RuntimeTypes.h
```diff
@@ -81,6 +81,15 @@ struct DAGNodeList {
   std::vector<std::unique_ptr<DAGNode>> nodes;
 };
 
+/// This is the base struct for DeviceManager configurations. The base struct
+/// contains a string for the device name. Any specific device can extend this
+/// struct to contain additional information to identify and configure the
+/// device manager.
+struct DeviceConfig {
+  /// A string containing the name of the device the device manager controls.
+  std::string deviceName;
```
nickgg: You have deviceName as a parameter here but you've deleted every point in the code that set a device name as far as I can tell. Do we need this?

gcatron: Probably not, I'll go ahead and remove it.

lib/Backends/Interpreter/InterpreterDeviceManager.h
```diff
@@ -40,9 +40,9 @@ class InterpreterDeviceManager : public QueueBackedDeviceManager {
   const uint64_t functionCost_{1};
 
 public:
-  InterpreterDeviceManager(llvm::StringRef name = "unnamed",
+  InterpreterDeviceManager(std::unique_ptr<DeviceConfig> config = nullptr,
```
SplitInfinity: Since we know the type of backend at this point, I wonder if all of these constructors should take in subclass pointers instead. My goal is to help quickly catch mistakes like this that might be made while writing an example or unit test:

```
auto dm = std::make_unique<InterpreterDeviceManager>(cpuDeviceConfigPtr);
```

lib/Backends/DeviceManagers.cpp
```diff
@@ -29,46 +29,52 @@ namespace runtime {
 /// when you define a new DeviceManager.
 
 /// Create a new instance of the interpreter Device.
-DeviceManager *createInterpreterDeviceManager(llvm::StringRef name);
+DeviceManager *
+createInterpreterDeviceManager(std::unique_ptr<DeviceConfig> config = nullptr);
```
SplitInfinity: Why did you remove the name field? For some vendors, that might be used to differentiate between different devices (e.g. by PCIe bus name).

gcatron: That would go into the derived config object for that vendor. Since it's not global to all types I removed it.

examples/tracing-compare.cpp
```diff
@@ -106,8 +106,7 @@ int main(int argc, char **argv) {
 
   std::array<DeviceManager *, supportedBackends.size()> devices;
   for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i) {
-    devices[i] = DeviceManager::createDeviceManager(supportedBackends[i],
-                                                    "tracing-compare");
+    devices[i] = DeviceManager::createDeviceManager(supportedBackends[i]);
     devices[i]->init();
```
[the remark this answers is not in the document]
gcatron: created issue #2485

lib/Backends/Interpreter/InterpreterDeviceManager.h
gcatron: That still requires a check at runtime. (Ideally we'd want to dyn_cast and check). This can also be handled by the device manager that is using the config. Pushing the check down to the manager means we only need a config object defined for a manager that needs one.

nickgg: I think we will need a runtime check, but as discussed offline we don't have RTTI available.

Do you think it's worth adding a check for devices like Interpreter/CPU which probably wont use anything in the config?

gcatron: I don't think so, if we check in the device managers that actually use the config we only have to check when it is actually used. I'll add the mechanism to check kind to this PR.

jackm321:
Looks good.
I also have some concerns like what were raised already here about passing a generic config type to each backend without any RTTI we just have to cast and hope we got the right type of config for the backend. That being said, I'm not sure what a better alternative would really look like.

PR #2487, opened 2019-03-06
[function-specializer] Fix a bug related to the removal of non-specialized calls

opti-mix:
If a non-specialized call is replaced by a specialized call and the original call produces a value, all uses of this return value should be replaced by the return value of the specialized call.

Detected by building Glow with a debug version of LLVM.


opti-mix:
@bertmaher or @nadavrot Could you approve this pretty trivial bugfix?

PR #2488, opened 2019-03-06
Use HostManager for ONNXIFI

jackm321:
*Description*:
Replace `ExecutionEngine` with `HostManager` for Onnxifi.
Clean up ownership of tensors by `Context`
`GlowOnnxifiManager` manages HostManagers, creating at most one `HostManager` for each `BackendKind` and sharing this between `onnxifi::BackendId`s that use the same kind.

Next steps after this pr include removing `glow::Backend` from `onnxifi::BackendId`, this isn't yet possible though because this is needed for compatibility checks. Once the required functions are made static we can remove this.

*Testing*:
`ninja check`
Will run c2 tests later

*Documentation*:
Code comments

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -220,6 +220,9 @@ onnxStatus Graph::setIOAndRunAsync(
     ctx->insert(inPhPtr, std::move(t));
   }
 
+  std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+      phNameToOnnxTensorOutputs;
+
   // Create tensors for output placeholders
   for (unsigned i = 0; i < outputsCount; ++i) {
     const auto &outOnnxTensor = outputDescriptors[i];
@@ -232,7 +235,9 @@ onnxStatus Graph::setIOAndRunAsync(
 
     auto &outPhPtr = outPhIt->getValue();
 
-    Tensor t(outOnnxBuffer, outPhPtr->getType());
+    phNameToOnnxTensorOutputs[outPhPtr] = outOnnxTensor;
+
+    Tensor t(outPhPtr->getType());
 
     ctx->insert(outPhPtr, std::move(t));
   }
@@ -240,8 +245,23 @@ onnxStatus Graph::setIOAndRunAsync(
   // Run
   getHostManager().runNetwork(
       netName_, std::move(ctx),
-      [outputEvent](runtime::RunIdentifierTy runId, runtime::ResultCode result,
-                    std::unique_ptr<Context> ctx) { outputEvent->signal(); });
+      [phNameToOnnxTensorOutputs = std::move(phNameToOnnxTensorOutputs),
+       outputEvent](runtime::RunIdentifierTy runId, runtime::ResultCode result,
+                    std::unique_ptr<Context> ctx) {
+        for (auto &ph : ctx->pairs()) {
+          if (phNameToOnnxTensorOutputs.count(ph.first) == 0) {
+            continue;
+          }
+
+          auto &outOnnxTensor = phNameToOnnxTensorOutputs.at(ph.first);
+
+          void *outputAddress = reinterpret_cast<void *>(outOnnxTensor.buffer);
+          Tensor *res = ph.second;
+          memcpy(outputAddress, res->getUnsafePtr(),
+                 res->size() * res->getType().getElementSize());
+        }
+        outputEvent->signal();
+      });
 
   return ONNXIFI_STATUS_SUCCESS;
 }
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -105,8 +105,7 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
                             uint32_t weightCount,
                             const onnxTensorDescriptorV1 *weightDescriptors) {
 
-  auto id = makeUniqueGraphId();
-  netName_ = llvm::formatv("inference_function_{}", id).str();
+  netName_ = strFormat("onnxifi_function_%lu", makeUniqueGraphId());
 
   Function *function = m_.createFunction(netName_);
 
@@ -121,7 +120,7 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
 
   auto err = backendPtr_->getHostManager().addNetwork(&m_);
 
-  if (std::move(errToBool)) {
+  if (errToBool(std::move(err))) {
     return ONNXIFI_STATUS_INTERNAL_ERROR;
   }
 
@@ -192,8 +191,8 @@ onnxStatus Graph::setIOAndRunAsync(
       [phNameToOnnxTensorOutputs = std::move(phNameToOnnxTensorOutputs),
        outputEvent](runtime::RunIdentifierTy runId, llvm::Error err,
                     std::unique_ptr<ExecutionContext> ctx) {
-        // If an Error occurred then log it in errToBool and signal, output
-        // event, and quit.
+        // If an Error occurred then log it in errToBool and signal the output
+        // event
         if (errToBool(std::move(err))) {
           outputEvent->signal();
           return;
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -106,7 +106,7 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
                             const onnxTensorDescriptorV1 *weightDescriptors) {
 
   auto id = makeUniqueGraphId();
-  netName_ = llvm::formatv("inference_function_%d", id);
+  netName_ = llvm::formatv("inference_function_{}", id).str();
 
   Function *function = m_.createFunction(netName_);
 
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -193,7 +193,7 @@ Graph::Graph(BackendPtr backendPtr) : backendPtr_(backendPtr) {}
 
 Graph::~Graph() {
   // Remove network from hostmanager
-  // backendPtr_->getHostManager().removeNetwork(netName_);
+  backendPtr_->getHostManager().removeNetwork(netName_);
 }
 
 onnxStatus Graph::setIOAndRunAsync(
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -76,11 +76,11 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
 // static
 std::unique_ptr<runtime::HostManager>
 BackendId::createHostManager(glow::BackendKind kind) {
-  std::vector<runtime::DeviceConfig> configs;
-  auto config = runtime::DeviceConfig();
-  config.deviceName = "device";
+  std::vector<runtime::DeviceManagerConfig> configs;
+  auto config = runtime::DeviceManagerConfig();
+  config.deviceConfig = nullptr;
   config.backendKind = kind;
-  configs.push_back(config);
+  configs.push_back(std::move(config));
   return llvm::make_unique<runtime::HostManager>(configs);
 }
 
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -119,10 +119,9 @@ onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
   onnxInputToPlaceholder_ = loader->getInputVarsMapping();
   onnxOutputToPlaceholder_ = loader->getOutputVarsMapping();
 
-  auto res = backendPtr_->getHostManager().addNetwork(&m_);
+  auto err = backendPtr_->getHostManager().addNetwork(&m_);
 
-  // TODO: add higher resolution error reporting
-  if (res != runtime::ResultCode::Ready) {
+  if (std::move(errToBool)) {
     return ONNXIFI_STATUS_INTERNAL_ERROR;
   }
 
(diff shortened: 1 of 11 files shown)


lib/Graph/Context.cpp
```diff
@@ -70,46 +70,61 @@ Placeholder *Context::getPlaceholderByName(llvm::StringRef name) const {
...
-void Context::insert(Placeholder *P, Tensor &&T) {
-  assert(!map_.count(P) && "Placeholder already registered");
+void Context::insert(Placeholder *P, std::unique_ptr<Tensor> T,
+                     bool overwrite) {
+  assert((overwrite || !map_.count(P)) && "Placeholder already registered");
+  map_[P] = std::move(T);
+  nameMap_[P->getName()] = P;
+}
+
+void Context::insert(Placeholder *P, Tensor &&T, bool overwrite) {
```
SplitInfinity: Consider implementing this as

```
insert(P, llvm::make_unique<Tensor>(std::move(T)), overwrite);
```

if possible.

lib/Onnxifi/CMakeLists.txt
```diff
@@ -14,20 +14,20 @@ add_library(onnxifi-glow-lib
 target_link_libraries(onnxifi-glow-lib
                       PUBLIC
                         Backends
-                        ExecutionEngine
                         Graph
+                        HostManager
                         Importer
                         Optimizer
-                        ThreadPool)
+                      )
```
SplitInfinity: Please move this up to the end of the previous line.

lib/Onnxifi/CMakeLists.txt
```diff
@@ -14,20 +14,20 @@ add_library(onnxifi-glow-lib
...
 target_link_libraries(onnxifi-glow
                       PUBLIC
                         Backends
-                        ExecutionEngine
                         Graph
+                        HostManager
                         Importer
                         Optimizer
-                        ThreadPool)
+                        )
```
SplitInfinity: Please move this up to the end of the previous line.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -166,18 +166,13 @@ bool ExecutionState::incrementNodeParentsDone(const DAGNode *node,
...
   assert(resultCtx_ && "Execution result context should exist!");
   std::lock_guard<std::mutex> lock(contextMtx_);
-  Tensor *tensor = resultCtx_->get(P);
-
-  if (tensor) {
-    *tensor = std::move(T);
-  } else {
-    resultCtx_->insert(P, std::move(T));
-  }
+  resultCtx_->insert(P, std::move(T), /* overwrite */ true);
```
SplitInfinity: `/* overwrite */` --> `/*overwrite*/`

include/glow/Graph/Context.h
```diff
@@ -65,8 +63,17 @@ class Context final {
   /// found.
   Placeholder *getPlaceholderByName(llvm::StringRef name) const;
 
-  /// Inserts the Placeholder-Tensor pair.
-  void insert(Placeholder *P, Tensor &&T);
+  /// Inserts the Placeholder-Tensor pair. If \p overwrite is true then
+  /// deletes an existing tensors for this placeholder if it exists.
```
gcatron: I'd add what happens if overwrite is false and a tensor already exists to this comment.

include/glow/Graph/Context.h
```diff
@@ -65,8 +63,17 @@ class Context final {
...
+  void insert(Placeholder *P, Tensor &&T, bool overwrite = false);
+
+  /// Inserts the Placeholder-Tensor pair. If \p overwrite is true then
+  /// deletes an existing tensors for this placeholder if it exists.
+  void insert(Placeholder *P, std::unique_ptr<Tensor> T,
+              bool overwrite = false);
+
+  /// Remove the Placeholder and its mapped Tensor from the context.
```
gcatron: // and returns the removed tensor. Probably obvious from the signature but for completeness.

lib/Onnxifi/Base.cpp
```diff
@@ -75,8 +73,15 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
   return ONNXIFI_STATUS_SUCCESS;
 }
 
-void Backend::runAsync(std::function<void(void)> &&fn) {
-  threadPool_.submit(std::move(fn));
+// static
+std::unique_ptr<runtime::HostManager>
+BackendId::createHostManager(glow::BackendKind kind) {
+  std::vector<runtime::DeviceConfig> configs;
+  auto config = runtime::DeviceConfig();
```
gcatron: This will need to get updated once you rebase. It's now DeviceManagerConfig. Check #2483 
The new deviceconfig object should probably left blank unless you have device specific logic.

lib/Onnxifi/Base.cpp
```diff
@@ -99,141 +104,91 @@ void Event::wait() {
...
+    auto t = llvm::make_unique<Tensor>(outOnnxBuffer, outPhPtr->getType());
+
+    ctx->insert(outPhPtr, std::move(t));
   }
 
+  // Run
+  getHostManager().runNetwork(
+      netName_, std::move(ctx),
+      [outputEvent](runtime::RunIdentifierTy runId, runtime::ResultCode result,
+                    std::unique_ptr<Context> ctx) { outputEvent->signal(); });
```
gcatron: Are you doing anything with the returned context? Otherwise I think it will get freed.

lib/Onnxifi/Base.h
```diff
@@ -71,71 +56,39 @@ class BackendId {
...
-    // hostManager_->runNetwork(networkName, std::move(ctx), std::move(cb));
-  }
+  // \returns a unique_ptr to a new HostManager for the given BackendKind \p
+  // kind.
+  static std::unique_ptr<runtime::HostManager>
+  createHostManager(glow::BackendKind kind);
 
 private:
-  int id_;
+  runtime::HostManager *hostManager_;
```
gcatron: Does this need to be a raw pointer?

gcatron:
Great work!
It seems like this could have been two PRs though.
The general cleanup and refactoring, and then the switch to host manager.
Thoughts?

rdzhabarov:
+1 to @gcatron. Please, extract `Context` related changes and random fixes into separate PR.

lib/Graph/Context.cpp
jackm321: Yeah I thought about this but it's like 3 lines so I don't know that it's worth it. Especially because I think we should remove this method in the future and just only manage tensors in contexts as unique_ptrs, no reason to have this then.

lib/Onnxifi/Base.cpp
jackm321: Sounds good, I'll rebase and update it

lib/Onnxifi/Base.cpp
jackm321: I don't think there's anything that needs to be done with the ctx here right? I think it should be just freed.

lib/Onnxifi/Base.h
jackm321: We could use a reference, would you prefer that? It can't be a unique_ptr due to lack of ownership and I don't think a shared_ptr is necessary.

gcatron: I'm good regarding pointer/reference. Just wanted to make sure it's not possible that the pointer is invalidated out from under you.

jackm321: Right now, no, HostManager effectively live forever until the GlowOnnxifiManager singleton dies (which could be improved but will require some reference counting and stuff)

jackm321:
Addressed comments.
Removed `Context` related changes.

lib/Onnxifi/Base.cpp
```diff
@@ -75,8 +73,15 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
   return ONNXIFI_STATUS_SUCCESS;
 }
 
-void Backend::runAsync(std::function<void(void)> &&fn) {
-  threadPool_.submit(std::move(fn));
+// static
+std::unique_ptr<runtime::HostManager>
+BackendId::createHostManager(glow::BackendKind kind) {
+  std::vector<runtime::DeviceManagerConfig> configs;
+  auto config = runtime::DeviceManagerConfig();
```
rdzhabarov: this looks a bit weird: `runtime::DeviceManagerConfig config`.

lib/Onnxifi/Base.cpp
```diff
@@ -99,141 +104,111 @@ void Event::wait() {
...
+  backendPtr_->getHostManager().removeNetwork(netName_);
 }
 
-onnxStatus Graph::setIO(uint32_t inputsCount,
-                        const onnxTensorDescriptorV1 *inputDescriptors,
-                        uint32_t outputsCount,
-                        const onnxTensorDescriptorV1 *outputDescriptors) {
-  // Avoid race on setting inputs/outputs and graph run.
-  inputRunMutex_.lock();
+onnxStatus Graph::setIOAndRunAsync(
```
rdzhabarov: Could you check that tests are passing on FB code?

lib/Onnxifi/Base.cpp
```diff
@@ -99,141 +104,111 @@ void Event::wait() {
...
-  } else {
-    executionEngine_.run(*ctx);
-    afterRun(std::move(ctx));
-  }
+Graph::~Graph() {
+  // Remove network from hostmanager
+  backendPtr_->getHostManager().removeNetwork(netName_);
 }
 
-onnxStatus Graph::setIO(uint32_t inputsCount,
```
rdzhabarov: yay!! this is great! (very happy to see this going away)

lib/Onnxifi/GlowOnnxifiManager.h
```diff
@@ -92,6 +91,10 @@ class GlowOnnxifiManager final {
 private:
   GlowOnnxifiManager() = default;
 
+  /// Create a new HostManager managing backends of kind \p backendKind or get
+  /// an existing HostManager for the backendKind if one exists.
+  runtime::HostManager *getOrCreateHostManager(BackendKind backendKind);
```
rdzhabarov: can you mention that this method is not thread safe and must be synchronized

lib/Onnxifi/GlowOnnxifiManager.h
```diff
@@ -104,6 +107,9 @@ class GlowOnnxifiManager final {
   /// The set of all valid glow Graphs.
   std::unordered_set<GraphPtr> graphs_;
 
+  // Map from BackendKind to HostManager managing devices of that kind.
```
rdzhabarov: nit: ///

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -413,29 +399,7 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxSetGraphIO)(
...
-  }
-
-  auto outputStatus = verifyDescriptors(outputsCount, outputDescriptors);
-  if (outputStatus != ONNXIFI_STATUS_SUCCESS) {
-    return outputStatus;
-  }
-
-  return glowGraph->setIO(inputsCount, inputDescriptors, outputsCount,
-                          outputDescriptors);
+  llvm_unreachable("Use onnxSetIOAndRunGraph instead.");
```
rdzhabarov: we should return error code back to the caller in case `onnxSetGraphIO` is invoked.

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -444,35 +408,7 @@ EXTERNC ONNXIFI_PUBLIC ONNXIFI_CHECK_RESULT onnxStatus ONNXIFI_ABI
...
-  if (initStatus != ONNXIFI_STATUS_SUCCESS) {
-    return initStatus;
-  }
-
-  auto *inputEvent = static_cast<glow::onnxifi::EventPtr>(inputFence->event);
-  auto *outputEvent = static_cast<glow::onnxifi::EventPtr>(outputFence->event);
-
-  glowGraph->runAsync(inputEvent, outputEvent);
-  return ONNXIFI_STATUS_SUCCESS;
+  llvm_unreachable("Use onnxSetIOAndRunGraph instead.");
```
rdzhabarov: ditto.

jackm321:
This isn't stale, it's surfaced a couple of wrinkles in the runtime and I've been sorting those out one at a time, will update this soon.

jackm321:
Imported and ran internal tests, 100% are passing.

lib/Onnxifi/Base.cpp
```diff
@@ -99,142 +104,116 @@ void Event::wait() {
 onnxStatus Graph::initGraph(const void *onnxModel, size_t onnxModelSize,
                             uint32_t weightCount,
                             const onnxTensorDescriptorV1 *weightDescriptors) {
-  // TODO: support multiple functions here.
-  function_ =
-      executionEngine_.getModule().createFunction(inferenceFunctionName);
+
+  netName_ = strFormat("onnxifi_function_%lu", makeUniqueGraphId());
+
+  Function *function = m_.createFunction(netName_);
```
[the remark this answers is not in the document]
jackm321: ~Module does it but actually this is a good point, we should hand ownership of module over to HostManager which can consume it, I'll change it.

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -63,25 +63,16 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
...
-      glow::BackendKind::CPU, /*id*/ 3,
-      /*concurrency*/ 1, /*useOnnx*/ false, /*useHostManager*/ false);
-  auto *interpreterBackendC2 = new glow::onnxifi::BackendId(
-      glow::BackendKind::Interpreter,
-      /*id*/ 4, /*concurrency*/ 1, /*useOnnx*/ false, /*useHostManager*/ false);
-  manager.addBackendId(cpuBackendOnnx);
-  manager.addBackendId(interpreterBackendOnnx);
-  manager.addBackendId(cpuBackendC2);
-  manager.addBackendId(interpreterBackendC2);
+  auto *cpuBackendOnnx = manager.createBackendId(glow::BackendKind::CPU,
```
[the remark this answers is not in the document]
jackm321: There are no EE base backends now, you think we should distinguish from before when they were EE based?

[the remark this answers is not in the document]
rdzhabarov: I guess my confusion came from `interpreterBackendC2HostManager` below.
I'd just name it consistently one way or the other.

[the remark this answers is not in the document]
jackm321: Oh good point

lib/Onnxifi/Base.cpp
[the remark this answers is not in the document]
jackm321: It seems that trying to change HostManager to unique_ptr messes up HostManager test with segfaults in DeviceManager and I'm not sure exactly why so we can change this later.

PR #2489, opened 2019-03-06
[Runtime][OpenCL Device Manager]Added OpenCL Device Manager config object

### LINKED ISSUE #2479: OpenCL device selection broken with new ExecutionEngine/DeviceManager
The new OpenCL deviceManager does not look for command line parameters to select the device. With the executionEngine change to use the deviceManager that means openCL cannot be command line configured.

Going forward we want to extend the DeviceManager interface to accept it's configuration parameters and the ExecutionEngine/HostManager will create these config objects and pass them in when creating the device managers.

gcatron:
*Description*: Added OpenCLDeviceConfig object to OpenCL Device Manager. Also added command line parameter support for OCL device manager when a config object is not provided.
*Testing*: ninja test and resnet-runtime
*Documentation*: NA
Fixes #2479

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -73,20 +73,19 @@ static const unsigned char kernels_fwd_quantized_conv_cl_src[] = {
 static const size_t kernels_fwd_quantized_conv_cl_src_size =
     sizeof(kernels_fwd_quantized_conv_cl_src);
 
-namespace {
 llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
 
-static llvm::cl::opt<unsigned>
-    platformId("platform", llvm::cl::desc("OpenCL platform to be used"),
-               llvm::cl::init(0), llvm::cl::cat(OpenCLBackendCat));
-static llvm::cl::opt<unsigned>
-    deviceId("device", llvm::cl::desc("OpenCL device to be used"),
-             llvm::cl::init(0), llvm::cl::cat(OpenCLBackendCat));
-static llvm::cl::opt<bool> doProfile("opencl-profile",
-                                     llvm::cl::desc("Profile OpenCL kernels"),
-                                     llvm::cl::init(false),
-                                     llvm::cl::cat(OpenCLBackendCat));
-} // namespace
+llvm::cl::opt<unsigned>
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -119,6 +119,15 @@ class OpenCLDeviceManager : public QueueBackedDeviceManager {
   void runFunctionImpl(runtime::RunIdentifierTy id, std::string functionName,
                        std::unique_ptr<Context> ctx, ResultCBTy cb) override;
 };
+
+class OpenCLDeviceConfig : public DeviceConfig {
```
opti-mix: Add Doxygen comments.

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -119,6 +119,15 @@ class OpenCLDeviceManager : public QueueBackedDeviceManager {
...
+public:
+  OpenCLDeviceConfig() : DeviceConfig(BackendKind::OpenCL) {}
+  unsigned platformId;
```
opti-mix: This class has only public members. Make it a struct or make the data members private (and do not forget the underscores at the end) and provide accessors.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -73,20 +73,20 @@ static const unsigned char kernels_fwd_quantized_conv_cl_src[] = {
...
-               llvm::cl::init(0), llvm::cl::cat(OpenCLBackendCat));
-static llvm::cl::opt<unsigned>
-    deviceId("device", llvm::cl::desc("OpenCL device to be used"),
-             llvm::cl::init(0), llvm::cl::cat(OpenCLBackendCat));
-static llvm::cl::opt<bool> doProfile("opencl-profile",
-                                     llvm::cl::desc("Profile OpenCL kernels"),
-                                     llvm::cl::init(false),
-                                     llvm::cl::cat(OpenCLBackendCat));
-} // namespace
+llvm::cl::opt<unsigned> platformId("platform",
```
opti-mix: Since this options are global now, may be the names of the variables should be more OpenCL specific? `platformId` and `deviceId` seems too generic for a global variable.

gcatron: changed to clPlatformId, clDeviceId, and clDoProfile.

PR #2490, opened 2019-03-06
Rename Context to PlaceholderBindings

nickgg:
*Description*: As the first step towards a broader ExecutionContext object, this renames all instances of Context to the leader of the name poll: PlaceholderBindings. 
*Testing*: tests in shared debug, release and asan.
*Documentation*: This is a breaking change to the interfaces to the Graph interface and backends.


gcatron:
This looks in order, great job! Since it is such a sweeping change can you run some of the examples ti make sure they are good too? (I expect them to be but they aren't covered in our unittests.)

nickgg:
Good idea, I ran: char-rnn, resnet-concurrent, resnet-verify, tracing-verify, and the tests/images/run.sh script to verify the model loader / image-classifier is working right. Everything looked good.

jackm321:
We should make some `EXPENSIVE` tests to run those scripts with CI

PR #2491, opened 2019-03-07
Add GlowErrs to runtime

jackm321:
*Description*:
Return `GlowErr`s from methods in DeviceManager, Provisioner, HostManager, Executor, etc and handle them in onnxifi.

*Testing*:
`Ninja check`

*Documentation*:
Comments

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -185,16 +185,13 @@ void ExecutionEngine::insertCompiledFunction(
   functionMap[name] = func.get();
   compiledFunctions_[name] = std::move(func);
 
-  // TODO: make this a std::promise<llvm::Error>
-  std::promise<bool> addPromise;
+  std::promise<llvm::Error> addPromise;
   auto fut = addPromise.get_future();
   device_->addNetwork(&M_, std::move(functionMap),
                       [&addPromise](const Module *, llvm::Error err) {
-                        addPromise.set_value(
-                            !glow::errorToBool(std::move(err)));
+                        addPromise.set_value(std::move(err));
                       });
-  fut.wait();
-  assert(fut.get() && "Compiled function failed to be added to device");
+  EXIT_ON_ERR(fut.get());
 }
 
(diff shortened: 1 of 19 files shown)


jackm321:
This only 1% done but there will be quite a few places to change here so I'm putting this up now to get opinions on the general direction.

jackm321:
Actually it's about 50% done now, it's not too much churn

jackm321:
Ok that's all the changes I want to do right now. Changing `ResultCBTy` involves a huge ton of changes to executor that I'm going to hold off on.

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -76,29 +76,33 @@ std::pair<std::promise<ResultType>, std::future<ResultType>> getFutureHelper() {
...
 TEST_P(DeviceManagerTest, Basic) {
   auto module = makeBasicModule();
   std::vector<std::unique_ptr<CompiledFunction>> backing;
   FunctionMapTy functions =
       compileFunctions(backendKind, module.get(), backing);
 
   auto *device = DeviceManager::createDeviceManager(backendKind);
-  ResultCode initResult = device->init();
-  EXPECT_EQ(initResult, ResultCode::Executed);
+  EXIT_ON_ERR(device->init());
```
gcatron: Does exit_on_err end the entire bank of unit tests?

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -168,15 +166,16 @@ void ExecutionEngine::insertCompiledFunction(
   functionMap[name] = func.get();
   compiledFunctions_[name] = std::move(func);
 
-  std::promise<runtime::ResultCode> addPromise;
+  // TODO: make this a std::promise<llvm::Error>
+  std::promise<bool> addPromise;
```
rdzhabarov: qq: why not make it `std::promise<llvm::Error>` now?

jackm321: Oops good catch thanks! That's what the todo was for. I'll do it now

PR #2494, opened 2019-03-07
[Backends] Static registration for backends.

rdzhabarov:
*Description*:
* Introduce static registration factory for general purpose registration of Backends/DeviceManagers/etc
* Migrate backends to register via Factory

*Testing*:
CI

*Documentation*:
doxygen

### DIFF
--- a/cmake/modules/GlowDefaults.cmake
+++ b/cmake/modules/GlowDefaults.cmake
@@ -49,3 +49,30 @@ else()
     set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} -march=native ${FAST_MATH_FLAGS}")
   endif()
 endif()
+
+# Specify whole archive flag for SRC library.
+# This is useful when static registration is performed in the SRC library and we'd like to ensure
+# linker does not clean unused objects.
+function(make_whole_archive DST SRC)
+  get_target_property(__src_target_type ${SRC} TYPE)
+  
+  # Depending on the type of the source library, we will set up the
+  # link command for the specific SRC library.
+  if (${__src_target_type} STREQUAL "STATIC_LIBRARY")
+    if(APPLE)
+      target_link_libraries(
+          ${DST} INTERFACE -Wl,-force_load,$<TARGET_FILE:${SRC}>)
+    elseif(MSVC)
+      # In MSVC, we will add whole archive in default.
+      target_link_libraries(
+          ${DST} INTERFACE -WHOLEARCHIVE:$<TARGET_FILE:${SRC}>)
+    else()
+      # Assume everything else is like gcc
+      target_link_libraries(${DST} INTERFACE
+        "-Wl,--whole-archive $<TARGET_FILE:${SRC}> -Wl,--no-whole-archive")
+      set_target_properties(${DST} PROPERTIES LINK_FLAGS "-Wl,--exclude-libs,ALL")
+    endif()
+  else()
+    target_link_libraries(${DST} INTERFACE ${SRC})
+  endif()
+endfunction()
--- a/include/glow/Support/Register.h
+++ b/include/glow/Support/Register.h
@@ -1,0 +1,90 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_SUPPORT_REGISTER_H
+#define GLOW_SUPPORT_REGISTER_H
+
+#include <cassert>
+#include <map>
+
+namespace glow {
+
+/// Base factory interface which needs to be implemented
+/// for static registration of arbitrary classes.
+/// For example, CPUFactory : BaseFactory<BackendKind, Backend>
+/// would be responsible for creating CPU backends registred
+/// with BackendKind::CPU key.
+template <class Key, class Base> class BaseFactory {
+public:
+  virtual ~BaseFactory() = default;
+  /// Create an object of Base type.
+  virtual Base *create() = 0;
+  /// Key used for a registered factory.
+  virtual Key getRegistrationKey() const = 0;
+};
+
+/// General registry for implementation factories.
+/// The registry is templated by the Key class and Base class that a
+/// set of factories inherits from.
+template <class Key, class Base> class FactoryRegistry {
+public:
+  using FactoryMap = std::map<Key, BaseFactory<Key, Base> *>;
+
+  /// Register \p factory in a static map.
+  static void registerFactory(BaseFactory<Key, Base> &factory) {
+    Key registrationKey = factory.getRegistrationKey();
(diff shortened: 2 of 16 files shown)


opti-mix:
@rdzhabarov Are you aware of http://llvm.org/doxygen/Registry_8h_source.html? Could we use that instead?

rdzhabarov:
@opti-mix interesting! Nope, have not seen that, let me take a look

bertmaher:
You're almost certainly hitting the --whole-archive problem (in brief, if there's no reference from the executable to any of the translation units containing your static objects, the statics won't be constructed pre-main).  You can work around this by linking the library with --whole-archive, but CMake doesn't really give you a way to do this (and as far as I can tell there's no easy work around).  In buck it's easy, you just add link_whole=True to your cxx_library :).

I think Caffe2 does something with --whole-archive to achieve a similar effect so it might be worth looking at what they've done.  Or maybe llvm::Registry also has a way of solving this, I dunno.

rdzhabarov:
Thanks @bertmaher @opti-mix 
llvm::registry does not really fit as I would like to have customized keys for the map (like BackendKind or DeviceManager kind etc).

` --whole-archive` - that what I also found browsing through SO etc (I'll try to see how to make this work).

But one interesting aspect is that OpenCL registration does work, while others do not. @bertmaher do you have an idea why that?

include/glow/Support/Register.h
```diff
@@ -0,0 +1,71 @@
...
+    if (it == factories().end()) {
+      return nullptr;
+    }
+
+    return it->second->create();
+  }
+
+private:
+  static FactoryMap &factories() {
+    static FactoryMap *factories = new FactoryMap();
```
jackm321: Does asan complain about this?

rdzhabarov: that should be fine, see here: https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use

jackm321: Cool thanks, I also found this piece from there interesting and it also touches on what I was saying https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2.

rdzhabarov:
That all works locally on mac, now need to make it work on CI! :)

lib/Backends/CPU/CPUFactory.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "CPUBackend.h"
+
+namespace glow {
+
+/// Register CPU backend via CPUFactory.
+class CPUFactory : public BaseFactory<BackendKind, Backend> {
```
opti-mix: Should the Factory class definition be a macro as well? It seems like all these Factories are pretty much the same. Only their names and their BackendKind differ.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -73,7 +73,7 @@ static const unsigned char kernels_fwd_quantized_conv_cl_src[] = {
 static const size_t kernels_fwd_quantized_conv_cl_src_size =
     sizeof(kernels_fwd_quantized_conv_cl_src);
 
-llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
+static llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
```
opti-mix: Is this change related to the PR?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -303,9 +299,9 @@ void OpenCLFunction::fillBuffer(cl_mem buffer, uint64_t start, uint64_t len,
 
 /// \returns the max local workgroup size for each dimension, under the
 /// opencl constraints, with the global workgroup sizes of \p global;
-void getMaxLocalWorkgroupSize(cl_kernel kernel, cl_device_id device,
-                              llvm::ArrayRef<size_t> global,
-                              llvm::MutableArrayRef<size_t> local) {
+static void getMaxLocalWorkgroupSize(cl_kernel kernel, cl_device_id device,
```
opti-mix: Is this change related to the PR?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -24,13 +24,14 @@
 #include "glow/IR/Instrs.h"
 #include "glow/LLVMIRCodeGen/LLVMIRGen.h"
 #include "glow/Support/Debug.h"
+#include "glow/Support/Register.h"
```
opti-mix: Why do we need this include here? Where is it used?

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -24,13 +24,14 @@
...
 
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/LLVMContext.h"
 
-using namespace glow;
+namespace glow {
```
opti-mix: Is this change related to the PR?

lib/Backends/OpenCL/OpenCL.cpp
rdzhabarov: nope, can do separate commit in this PR

lib/Backends/CPU/CPUBackend.cpp
rdzhabarov: Should not be here, it was an artifact of having registration in this file.

lib/Backends/OpenCL/CMakeLists.txt
```diff
@@ -47,11 +47,16 @@ add_library(OpenCLDeviceManager
 target_link_libraries(OpenCLDeviceManager
                       PRIVATE
                         Backend
-                        Backends
+                        OpenCLBackend
```
opti-mix: @rdzhabarov Do we need the authors of backends and device managers take care of something or keep an invariant when they use these new factories? E.g. is there a rule like "Do not explicitly depend on Backends, but depend on the specific backend or specific factory?" If there are such rules now, we should state it somewhere.

rdzhabarov: Updated doc.

cmake/modules/GlowDefaults.cmake
```diff
@@ -49,3 +49,30 @@ else()
     set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} -march=native ${FAST_MATH_FLAGS}")
   endif()
 endif()
+
+# Specify whole archive flag for SRC library.
+# This is useful when static registration is performed in the SRC library and we'd like to ensure
+# linker does not clean unused objects.
+function(make_whole_archive DST SRC)
```
bertmaher: Amazing.  How did you figure this cmake wizardry out???

xerpi:
Hi, I'm having a problem where my backend depends on other `.a` libraries and I'm getting undefined references because it seems `make_whole_archive` breaks the dependence transitivity of `target_link_libraries`.
How should I proceed? Should I combine all the `.a` dependencies of my backend into a single `.a`? Should I also use `make_whole_archive` with my backend to include all its dependencies?

Thanks.

bertmaher:
@xerpi I think the right thing to do is to not depend on transitivity to pull in dependences of your backend, and instead add them as dependences where needed by other targets.

xerpi:
Thanks. I think I'll go by using dynamic libraries for my dependencies.

bertmaher:
@xerpi I think I may have misunderstood you.  Are your .a dependencies created by a separate build system?  If so, you need to teach cmake how to find those libraries, e.g., to find libexample.a
```
find_library(EXAMPLE example)
target_link_libraries(YourBackend "${EXAMPLE}" ...)
```
If libexample.a isn't in your standard sysroot (e.g., /usr/lib), you may need to set CMAKE_PREFIX_PATH when configuring, like: `cmake -DCMAKE_PREFIX_PATH=/path/to/my/libs`

PR #2495, opened 2019-03-07
Move DEBUG_TYPE after #includes

bertmaher:
*Description*: Some llvm headers #undef this macro, which leads to undefined symbol errors.  It's safer to #define it after all the other #includes.

*Testing*: build with an fb-internal platform where the issue repros.

*Documentation*: n/a


opti-mix:
This issue was also mentioned in #2379

PR #2499, opened 2019-03-08
replace runtime PlaceholderBindings with ExecutionContext

nickgg:
*Description*: Adds the new runtime ExecutionContext which includes PlaceholderBindings (previously: Context), DeviceBindings and TraceEvents storage, allowing us to (among other things) finish the OpenCL DeviceManager.
*Testing*: Unit tests debug, release and shared. Ran example code and verified.
*Documentation*: This is a breaking change to the Backend (due to change in execute() argument) but I broke that yesterday so pretty much a wash.

### DIFF
--- a/include/glow/Backends/ExecutionContext.h
+++ b/include/glow/Backends/ExecutionContext.h
@@ -1,0 +1,116 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_BACKENDS_EXECUTIONCONTEXT_H
+#define GLOW_BACKENDS_EXECUTIONCONTEXT_H
+
+#include "glow/Backends/TraceEvents.h"
+#include "glow/Graph/PlaceholderBindings.h"
+
+#include "llvm/ADT/STLExtras.h"
+
+namespace glow {
+
+enum class BackendKind;
+
+/// Sub-classed per backend, this holds Device specific per-function information
+/// if that is necessary on that particular backend.
+class DeviceBindings {
+  const glow::BackendKind backend_;
+
+public:
+  DeviceBindings(BackendKind kind) : backend_{kind} {}
+  virtual ~DeviceBindings() {}
+
+  virtual std::unique_ptr<DeviceBindings> clone() {
+    return llvm::make_unique<DeviceBindings>(backend_);
+  }
+
+  BackendKind getBackendKind() { return backend_; }
+};
+
+/// The runtime context for a single execution (Inferance or Training) in the
+/// the Glow Execution Engine or HostManager. This class includes the mapping
+/// between Input/Output Placeholders and the materialized Tensors used for this
+/// run, the set of Device specific details required to execute the function,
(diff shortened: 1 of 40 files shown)


include/glow/Backends/ExecutionContext.h
```diff
@@ -0,0 +1,93 @@
...
+#include "glow/Backends/TraceEvents.h"
+#include "glow/Graph/PlaceholderBindings.h"
+
+#include "llvm/ADT/STLExtras.h"
+
+namespace glow {
+
+/// Sub-classed per backend, this holds Device specific per-function information
+/// if that is necessary on that particular backend.
+class DeviceBindings {
```
gcatron: I think this will need the same solution we used for DeviceConfig, backendKind as a private enum set at construction so the CompiledFunction can check type before casting. (Or the compiledFunction can just trust it is always passed the right type.)

include/glow/Backends/ExecutionContext.h
```diff
@@ -0,0 +1,93 @@
...
+  std::unique_ptr<PlaceholderBindings> placeholderBindings_;
+  std::unique_ptr<DeviceBindings> deviceBindings_;
+
+  /// Trace Events recorded during this run.
+  std::vector<TraceEvent> traceEvents_;
+
+public:
+  ExecutionContext()
+      : placeholderBindings_(llvm::make_unique<PlaceholderBindings>()),
+        deviceBindings_(llvm::make_unique<DeviceBindings>()) {}
```
gcatron: I think deviceBindings can be a nullptr if one is not provided. The base class provides no value to a compiledFunction so it could just check for a nullptr.

include/glow/Backends/ExecutionContext.h
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /// Clones this ExecutionContext, but does not clone underlying Tensors.
+  ExecutionContext clone() {
+    return ExecutionContext(
+        llvm::make_unique<PlaceholderBindings>(placeholderBindings_->clone()),
+        deviceBindings_->clone());
+  }
+};
+
```
gcatron: Should we also have a getDeviceBindings?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -18,6 +18,7 @@
 
 #include "glow/Backends/Backend.h"
 #include "glow/Backends/BackendUtils.h"
+#include "glow/Backends/ExecutionContext.h"
```
gcatron: Can we use a forward declaration instead?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -30,7 +30,7 @@ void InflightBarrier::decrement(unsigned decr) {
   assert(count_ >= decr && "Barrier decrement cannot be less than count!");
   count_ -= decr;
 
-  // If count_ has hit zero, wake up all threads that are waiting.
+  // If count_ has hit zero, wake up all threads that are waiting.get();
```
gcatron: .get()?

include/glow/Backends/ExecutionContext.h
nickgg: Yeah, I think it makes sense to add BackendKind to the base class.

include/glow/Backends/ExecutionContext.h
nickgg: Wasn't sure how we'd use it so I was going to leave it up to you, nbd to add in something generic but sensible now.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: yank buffer strikes again

include/glow/Backends/ExecutionContext.h
```diff
@@ -0,0 +1,116 @@
...
+#include "llvm/ADT/STLExtras.h"
+
+namespace glow {
+
+enum class BackendKind;
+
+/// Sub-classed per backend, this holds Device specific per-function information
+/// if that is necessary on that particular backend.
+class DeviceBindings {
+  glow::BackendKind backend_;
```
gcatron: I think backend_ can be const.

PR #2516, opened 2019-03-11
Minor build fix for VS.

ayermolo:
*Description*: VS doesn't like explicit initialization of nodes, and reports build failure. This should accomplish the same thing.
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -93,6 +93,13 @@ struct DAGNode {
 using rootDAGNodeTy = std::unique_ptr<DAGNode>;
 using nodesDAGNodeTy = std::vector<std::unique_ptr<DAGNode>>;
 struct DAG {
+  DAG(){};
+  DAG(rootDAGNodeTy &&rootDAG, nodesDAGNodeTy &&nodesDAG)
+      : root(std::move(rootDAG)), nodes(std::move(nodesDAG)) {}
+
+  DAG(DAG &&node_)
+      : root(std::move(node_.root)), nodes(std::move(node_.nodes)) {}
+
   rootDAGNodeTy root;
   nodesDAGNodeTy nodes;
 };
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -491,9 +491,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   }
 
   // Push the DAG {root, nodes} into partitions_.
-  //  DAG newDag = {.root = std::move(DAGRoot), .nodes = std::move(nodes)};
-  partitions_.push_back(
-      {.root = std::move(DAGRoot), .nodes = std::move(nodes)});
+  partitions_.emplace_back(std::move(DAGRoot), std::move(nodes));
 
   // Update links between nodes in the cloned functions. Add placeholders (and
   // save nodes) where a link crosses a partition boundary.


lib/Partitioner/Partitioner.cpp
```diff
@@ -492,8 +492,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
 
   // Push the DAG {root, nodes} into partitions_.
   //  DAG newDag = {.root = std::move(DAGRoot), .nodes = std::move(nodes)};
```
jackm321: Would you mind also removing this, looks like dead code

lib/Partitioner/Partitioner.cpp
```diff
@@ -492,8 +492,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
-  partitions_.push_back(
-      {.root = std::move(DAGRoot), .nodes = std::move(nodes)});
+  partitions_.push_back({std::move(DAGRoot), std::move(nodes)});
```
jackm321: This seems a bit fragile to depend on the default constructor. Maybe create a new `DAG` and initialize each member or create a constructor for DAG?

ayermolo: @jackm321 Maybe just use emplace_back then?

jackm321: @ayermolo emplace_back using the same constructor would be the same issue. If you create a `DAG dag` and do emplace_back(std::move(dag)) that would be fine

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -93,6 +93,17 @@ struct DAGNode {
 using rootDAGNodeTy = std::unique_ptr<DAGNode>;
 using nodesDAGNodeTy = std::vector<std::unique_ptr<DAGNode>>;
 struct DAG {
+  DAG(){};
+  DAG(rootDAGNodeTy &&root_, nodesDAGNodeTy &&nodes_) {
```
rdzhabarov: nit: args should not have `_` and also use initializer list here instead of ctor body.

ayermolo: ok done

lib/Partitioner/Partitioner.cpp
ayermolo: Changed to emplace_back so only the explicit argument constructor is used. Added the move constructor for entire DAG in case push_back, or something similar, is used elsewhere.

rdzhabarov:
LGTM, thanks @ayermolo

ayermolo:
Great, thanks for feedback.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -93,6 +93,13 @@ struct DAGNode {
 using rootDAGNodeTy = std::unique_ptr<DAGNode>;
 using nodesDAGNodeTy = std::vector<std::unique_ptr<DAGNode>>;
 struct DAG {
+  DAG(){};
```
jackm321: This could be default

ayermolo:
Sorry, I guess I miss understood this part "or create a constructor for DAG"

jackm321:
@ayermolo my previous comment could be phrased better. I'm sorry, it was pretty abrupt. We probably don't need to create these constructors in this way but it works. Again sorry for the abruptness of my previous comment and thank you for fixing this :)

ayermolo:
@jackm321 No problem. Thank you for feedback. :)

PR #2517, opened 2019-03-11
Add images argument to resnet example tests

jackm321:
*Description*:
As title.
Also don't create text-translator variable because now we're passing in bin as variable so it's kind of redundant.

*Testing*:
`ninja check_expensive`
CI

*Documentation*:
none


jackm321:
If anyone who is less of a shell scripting noob than me can tell me a good way to output the results of the command being piped into OutputCheck if OutputCheck fails (or even no matter the result) that would be awesome. I was messing around with `tee` but it didn't seem to do what I wanted here.

nickgg:
Can you just redirect to a file in tmp/ then cat it?

jackm321:
@nickgg yeah I was thinking about doing that with tmpfile, that's probably the way to go

jackm321:
@jfix71 I'm going to land this again, hopefully this fixes the issue from this morning, if not it may be an environment issue on CI or something.

PR #2518, opened 2019-03-11
[Quantization] Add support for fused rowwise-quantizing SLWS

jfix71:
*Description*: Previously rowwise-quantized SLWS was only enabled by directly loading it as quantized. This allows for rowwise-quantizing it through our normal profile + quantization flow.

*Testing*: Added two unit tests; tested on other private models.

*Documentation*: Updated.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -476,14 +476,19 @@ public:
     lastMorphedNodeWithTypeChanges = nullptr;
   }
 
-  /// Traverse all nodes to find quantized FullyConnected nodes, and convert it
-  /// to RowwiseQuantizedFullyConnected if the weights is constant.
+  /// Traverse all nodes to find applicable quantized nodes, and convert them
+  /// to RowwiseQuantized versions if required inputs are Constant.
   void enableRowwise() {
     auto nodeIt = function_.getNodes().end();
     auto stopIt = function_.getNodes().begin();
     do {
       --nodeIt;
       Node &node = *nodeIt;
+      auto *Q = llvm::dyn_cast<DequantizeNode>(&node);
+      if (!Q) {
+        continue;
+      }
+
       // After function "convert()" is called, one FullyConnectedNode is
       // converted into:
       // [fp32 input] [fp32 weights] [fp32 bias]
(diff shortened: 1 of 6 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -476,14 +476,19 @@ class FunctionQuantizer : public FunctionConverter {
     lastMorphedNodeWithTypeChanges = nullptr;
   }
 
-  /// Traverse all nodes to find quantized FullyConnected nodes, and convert it
-  /// to RowwiseQuantizedFullyConnected if the weights is constant.
+  /// Traverse all nodes to find applicable quantized nodes, and convert them
+  /// to RowwiseQuantized versions if applicable is constant.
```
rdzhabarov: `if applicable is constant`? Did you mean something else here?

lib/Quantization/Quantization.cpp
```diff
@@ -502,31 +507,68 @@ class FunctionQuantizer : public FunctionConverter {
...
+            // Replace the usages of quantized FC node to
+            // RowwiseQuantizedFullyConnected Node.
+            result.replaceAllUsesOfWith(fcq->getResult());
           }
         }
       }
+
+      // Convert SLWS from normal quantized version to fused rowwise-quantized
+      // version if applicable. Data must be Constant for this to occur. We also
+      // will not quantized the weights as we do for the default normal
```
rdzhabarov: typo `quantized.

tests/unittests/QuantizationTest.cpp
```diff
@@ -333,6 +333,58 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
...
+
+  auto *data = mod.createPlaceholder(ElemKind::FloatTy, {3, 1}, "data", false);
+  auto *weights =
+      mod.createPlaceholder(ElemKind::FloatTy, {8}, "weights", false);
+  auto *indices =
+      mod.createPlaceholder(ElemKind::Int64ITy, {8}, "indices", false);
+  auto *lengths =
+      mod.createPlaceholder(ElemKind::Int32ITy, {4}, "lengths", false);
+
+  bindings.allocate(data)->init(Tensor::InitKind::Xavier, 3, mod.getPRNG());
```
rdzhabarov: nit: we do not necessarily need to `init` this data (we already kinda profile it by having `std::vector<NodeQuantizationInfo>`.

lib/Quantization/Quantization.cpp
```diff
@@ -502,31 +507,68 @@ class FunctionQuantizer : public FunctionConverter {
...
+      // quantized SLWS, as the rowwise version uses float weights.
+      if (auto *SLWS =
+              llvm::dyn_cast<SparseLengthsWeightedSumNode>(Q->getInput())) {
+        NodeValue data = SLWS->getData();
+        auto *dataQN = llvm::dyn_cast<QuantizeNode>(data.getNode());
+        assert(dataQN && "Data input should have been quantized");
```
rdzhabarov: any reason those checks are asserts vs for FC it's actually part of the precondition before conversion?

jfix71: Hmm, well I guess it's possible that the FC was not quantized due to `doNotQuantizeKinds`. I'll make it a precondition, thanks!

PR #2523, opened 2019-03-12
[Quantization] Rowwise-quantize already lowered FC

jfix71:
*Description*: This PR adds the ability to rowwise-quantize an FC that was already lowered to MM + BA. Because of lowering before quantization we were lowering FCs before looking for them to rowwise-quantize them, meaning lowered FCs weren't being rowwise-quantized.

*Testing*: Added unit tests.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -118,42 +118,28 @@ protected:
       // Check if this BatchedAdd was lowered from a FullyConnectedNode. If so
       // then we need to calculate the Int32QTy scale/offset and use that
       // instead of Int8QTy.
-      auto *baN = llvm::cast<BatchedAddNode>(&use);
-      NodeValue batch = baN->getBatch();
-
-      // Look for the set of NodeNameAndKinds corresponding to the
-      // BatchedAdd. If one exists, this means it was lowered.
-      auto it = loweredMap_.find(NodeQuantizationInfo::generateNodeOutputName(
-          baN->getName(), BatchedAddNode::ResultIdx));
-      if (it != loweredMap_.end()) {
-        // Look through the set looking to see if the BatchedAdd was lowered
-        // from a FullyConnectedNode.
-        auto &set = it->getValue();
-        for (auto i = set.begin(), e = set.end(); i != e; ++i) {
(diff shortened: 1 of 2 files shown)


beicy:
So for all FC, we do the row-wise quantization now?

jfix71:
@beicy Only if the `enableRowwise` flag is passed to `quantizeFunction()`, same as before. The problem here was, after I added lowering before quantization, the quantizer was no longer ever seeing FCs to rowwise quantize, because they were already lowered to MM + BA. So this PR allows for the quantizer to determine a MM + BA came from an FC and then replace the MM + BA with a rowwise-quantized FC.

lib/Quantization/Quantization.cpp
```diff
@@ -507,11 +515,37 @@ class FunctionQuantizer : public FunctionConverter {
...
-        NodeValue weights = fcN->getWeights();
-        NodeValue bias = fcN->getBias();
-        NodeValue result = fcN->getResult();
+        foundFC = true;
+        input = fcN->getInput();
+        weights = fcN->getWeights();
+        bias = fcN->getBias();
+        result = fcN->getResult();
+      } else if (const auto *baN =
+                     llvm::dyn_cast<BatchedAddNode>(Q->getInput())) {
```
beicy: Since now BatchedAddNode could be replaced as well, the comments in later replacement (line 561-562) need to be updated.

beicy: Also, one more question: if the batchedAdd and Matmul is not lowered from FC (that is, the module has 2 layers as matmal + bachedAdd), do we need to do this conversion?

jfix71: That means that the original model we load would have MatMul + BatchedAdd instead of FC. I'm not sure if we want to do that in this case -- we could, since it is equivalent after all. But that feels like outside of the scope of this PR. My initial intuition for supporting that would be that we'd want an earlier pass to replace MM + BA patterns with FC.

jfix71: >Since now BatchedAddNode could be replaced as well, the comments in later replacement (line 561-562) need to be updated.

Updated the comments.

PR #2524, opened 2019-03-12
[Training] Add support for Adagrad training algorithm

SplitInfinity:
**Description**
  This commit adds support for the Adagrad (adaptive gradient) training
  algorithm. This algorithm computes a unique learning rate for each
  trainable model parameters that is directly proportional to the sum of
  the squares of past gradients with respect to that weight. Weights with
  large gradients are paired with small learning rates, and vice versa.
    
  **Testing**
  This commit adds a test case that learns an exponential function using
  the Adagrad algorithm.

### DIFF
--- a/include/glow/Base/Train.h
+++ b/include/glow/Base/Train.h
@@ -27,7 +27,7 @@ namespace glow {
 class Tensor;
 
 /// These are all the supported training algorithms.
-enum class TrainingAlgorithm { None, StochasticGradientDescent };
+enum class TrainingAlgorithm { None, StochasticGradientDescent, Adagrad };
 
 /// This is a list of common parameters that all of the training algorithms use.
 struct TrainingParameters {
@@ -42,6 +42,11 @@ struct SGDParameters : public TrainingParameters {
   float momentum{0.0};
 };
 
+/// Additional training parameters used by Adagrad.
+struct AdagradParameters : public TrainingParameters {
+  float epsilon{1e-5};
+};
+
 /// Combination of training algorithm type and training parameters. Parameters
 /// should be casted to the appropriate type based on the algorithm type.
 struct TrainingConfig {
@@ -50,6 +55,8 @@ struct TrainingConfig {
       : algorithm(algo) {
     if (algorithm == TrainingAlgorithm::StochasticGradientDescent) {
       parameters = llvm::make_unique<SGDParameters>();
+    } else if (algorithm == TrainingAlgorithm::Adagrad) {
+      parameters = llvm::make_unique<AdagradParameters>();
     } else {
       llvm_unreachable("Invalid training algorithm.");
     }
--- a/include/glow/Base/Train.h
+++ b/include/glow/Base/Train.h
@@ -26,14 +26,42 @@ namespace glow {
 
 class Tensor;
 
-/// This is a list of parameters that the network trainers (such as sgd and
-/// adam) use for training the network.
-struct TrainingConfig {
+/// These are all the supported training algorithms.
+enum class TrainingAlgorithm { None, StochasticGradientDescent };
+
+/// This is a list of common parameters that all of the training algorithms use.
+struct TrainingParameters {
+  float learningRate{0.01f};
+  unsigned batchSize{1};
+};
+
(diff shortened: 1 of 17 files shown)


include/glow/Base/Train.h
```diff
@@ -26,14 +26,47 @@ namespace glow {
...
+};
+
+/// Combination of training algorithm type and training parameters. Parameters
+/// should be casted to the appropriate type based on the algorithm type.
+struct TrainingConfig {
+  TrainingConfig(
+      TrainingAlgorithm algo = TrainingAlgorithm::StochasticGradientDescent)
+      : algorithm(algo) {
+    if (algorithm == TrainingAlgorithm::StochasticGradientDescent) {
+      parameters = std::make_unique<SGDParameters>();
```
SplitInfinity: `llvm::make_unique`

include/glow/Base/Train.h
```diff
@@ -26,14 +26,47 @@ namespace glow {
...
+struct SGDParameters : public TrainingParameters {
   float L1Decay{0};
   float L2Decay{0};
-  float learningRate{0.01f};
   float momentum{0.0};
-  unsigned batchSize{1};
+};
+
+/// Additional training parameters used by Adagrad.
+struct AdagradParameters : public TrainingParameters {
```
SplitInfinity: Oops, my next commit is leaking...

lib/Graph/Grad.cpp
```diff
@@ -259,13 +260,42 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
     }
 
-    auto X = new SGDNode(PH->getName(), map.getGradient(PH), PH, conf.L1Decay,
-                         conf.L2Decay, conf.learningRate, conf.momentum,
-                         conf.batchSize);
-    toAppend.push_back(X);
-    // Now update the weight with the value computed by SGD.
-    auto *save = new SaveNode(PH->getName().str() + ".saveGrad", {X, 0}, PH);
-    toAppend.push_back(save);
+    if (conf.algorithm == TrainingAlgorithm::StochasticGradientDescent) {
```
rdzhabarov: I'm not in favor of this design. Adding new training algo should not change the `differentiate` procedure.

SplitInfinity: `differentiate` has to change, but the only changes required are those to insert different nodes that compute updates to trainable `Placeholders` based on the selected training algorithm. The current implementation is hardcoded to use an `SGDNode`, and that is only applicable to the stochastic gradient descent training algorithm.

In this specific case, `differentiate` needs to take in a reference to `PlaceholderBindings` because the algorithm needs to store a history of momentum values for each trainable parameter. I would prefer not to rely on the user to allocate extra `Placeholders` for this information and name them a specific way so that the code here can connect them correctly.

rdzhabarov: What if I'd like to add another training algorithm?

For one, we could pass ref/pointer to a base class down to a `differentiate` method. Base class would implement `instrument` logic and based on the specific implementation of `instrument` we'd do different things.
This way you abstract instrumentation away from the code of `differentiate` method.

SplitInfinity: What do you think of making `TrainingConfig` into a full-fledged class that stores training parameters (learning rate, decay, momentum, etc.) and has an `instrument(Function* F, PlaceholderBindings& bindings)` method? `differentiate already takes this as a parameter anyway.

Implementations of this method would iterate through the `Placeholders` and create the appropriate weight update nodes for the trainable ones.

rdzhabarov: I like this idea.

I'd also explore how other frameworks handle this like mxnet etc.

SplitInfinity: mxnet has a `Trainer` class that allows you to specify the training parameters and the algorithm type and has a method that updates your parameters when called.

I think this maps almost exactly to the Glow use case, except that this `Trainer` class will not update the weights. It will insert nodes into the graph that will update the weights automatically when `ExecutionEngine::run` or `CompiledFunction::execute` is called.

PR #2526, opened 2019-03-12
Remove OpenCLFunction dependence on old stateful execute scheme.

nickgg:
*Description*: As we did with the CPU and Interpreter backend, removing the requirement of calling setupRuns, beforeRuns, afterRuns and tearDownRuns in the OpenCL backend by moving their logic into `execute()`. This is the first step in making OpenCLFunction stateless, but we still need to move the buffers and kernelLaunch vector into DeviceBindings.
*Testing*: unit tests.
*Documentation*: Requires no change to calling code (although we could remove the unnecessary calls to the removed methods).

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -625,6 +625,13 @@ static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
 void OpenCLFunction::execute(ExecutionContext *context) {
   (void)context;
 
+  // TODO: deviceBuffer_ should go into DeviceBindings in context.
+  deviceBuffer_ = allocDeviceBuffer(runtimeBundle_.getConstantWeightSize() +
+                                    runtimeBundle_.getMutableWeightSize() +
+                                    runtimeBundle_.getActivationsSize());
+
+  loadPlaceholders(context->getPlaceholderBindings());
+
   for (const auto &I : F_->getInstrs()) {
     // Skip memory allocation instructions as they are NOPs.
     if (isa<AllocActivationInst>(I) || isa<DeallocActivationInst>(I) ||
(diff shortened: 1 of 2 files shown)


lib/Backends/OpenCL/OpenCL.h
```diff
@@ -157,6 +149,12 @@ class OpenCLFunction final : public CompiledFunction {
                      cl_device_id device, llvm::ArrayRef<size_t> global,
                      llvm::ArrayRef<size_t> local,
                      std::vector<KernelLaunch> &kernelLaunches);
+
+  /// Load inputs from \p bindings onto the device.
+  void loadPlaceholders(PlaceholderBindings *bindings);
+
+  // Load outputs from the device into \p bindings.
```
SplitInfinity: `//` --> `///`

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1440,35 +1462,28 @@ uint64_t OpenCLFunction::copyValueFromDevice(const Value *v, void *buf) {
...
     }
-    runsSetup_ = true;
+    // Do it!
+    clFinish(commands_);
   }
-}
 
-void OpenCLFunction::beforeRun(const PlaceholderBindings &bindings) {
-  for (auto PH : bindings.pairs()) {
+  // TODO: doesn't this copy constants twice?
```
SplitInfinity: Hm, once when the network is added, and a second time here?

opti-mix:
@nickgg Memory allocation/deallocation happens in `execute` now. Does it mean that we allocate memory and copy constants to the device on every run? If this is the case, it is rather bad in terms of performance. Ideally, we want to copy constants once and then use it by multiple runs of the same network. 

Of course, there are possible issues with it, e.g. if we have multiple network deployed on the device, they may compete for memory and you cannot deploy another network, because there is not enough memory available as it is reserved for networks that are deployed already and are ready to run... So, some kind of eviction scheme may be needed, to purge the memory of a network from the GPU memory if we run into "not enough memory" situation. This kind of optimization is of course something for a separate PR, if at all.

SplitInfinity:
@opti-mix 

I think that's what @nickgg 's TODO was about. The constants are already copied when the network is added to the device.

nickgg:
@SplitInfinity @opti-mix Exactly, this should and will be done by the OpenCLDeviceManager - and we should keep the buffers in the DeviceBindings which allows us to reuse them after reading outputs.

Moving it into execute is temporary, but I think not worse in the short term since we're calling setup/teardown every run anyway right now - but we're still storing per run state in the compiled function (which prevents reusing the function on multiple devices).

lib/Backends/OpenCL/OpenCL.cpp
nickgg: No actually I just thought Constants were in the PlaceholderBindings for a sec, but I'm wrong.

PR #2527, opened 2019-03-12
Resize output of FC with dims > 2

### LINKED ISSUE #2522: Need to output correct shape for C2 FC op
Today, if the input X or W for FC op has dim higher than 2, they will be flattened into 2D along `axis` or `axis_w`:
https://github.com/pytorch/glow/blob/143f862e7c5bd19b050c5865f3b465c7e5986185/lib/Importer/Caffe2ModelLoader.cpp#L569

And output will be 2D. This creates semantic difference between Glow and C2 ops, because c2's output shape will have the same dim size as input. The consequence is that if we have ` FC -> Transpose` sequence, with FC having 4D input and Transpose having 4D perms arguement, you will meet with a segfault, as Transpose is trying to access 4D dim of the input. 

The solution should be adding a `Reshape` node after FullyConnect layer when importing C2 FC op and when its input is more than 2D. The spec of the Reshape can be computed using:
https://github.com/pytorch/pytorch/blob/c10662962c5ea7da9762be65eacafee376d9c13c/caffe2/operators/fc_inference.cc#L4-L33

jackm321:
*Description*:
We need to match the behavior of Caffe2. When number of input dimensions to FC is greater than 2, Glow flattens down to 2 dimensions using axis but does not reshape the output back to original dimensions before the axis.

*Testing*:
Create a new C2 Graph and test that the output shape is the expected shape.

*Documentation*:
Fixes #2522

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -560,12 +560,16 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     NodeValue in;
     ASSIGN_VALUE_OR_RETURN_ERR(in,
                                getNodeValueOrCreateConstantByName(op.input(0)));
-    if (in.getType()->dims().size() > 2) {
-      size_t axis = 1;
-      if (dict.count("axis")) {
-        ASSIGN_VALUE_OR_RETURN_ERR(axis, loadInt(dict["axis"]));
-      }
 
+    auto originalInputDims = in.getType()->dims();
+
+    size_t axis = 1;
+    if (dict.count("axis")) {
+      ASSIGN_VALUE_OR_RETURN_ERR(axis, loadInt(dict["axis"]));
+    }
+
+    // If number of input dims is greater then 2 flatten on axis.
+    if (in.getType()->dims().size() > 2) {
       in = G_.createFlatten("fc.in", in, axis);
     }
 
@@ -580,10 +584,10 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     }
 
     // Caffe2 stores the transposed W matrix. In here we first coerce W to a 2D
-    // matrix size if necessay and then transpose it back.
+    // matrix size if necessary and then transpose it back.
     Tensor tmp;
+    auto wDims = flattenCdr(w->dims(), axis_w);
     if (w->dims().size() > 2) {
-      auto wDims = flattenCdr(w->dims(), axis_w);
       if (typeName == "FC" || typeName == "FCTransposed") {
         tmp.reset(ElemKind::FloatTy, {wDims.first, wDims.second});
       } else {
@@ -622,6 +626,30 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       node = G_.createFullyConnected(opName, in, W, B, outTy);
     } else {
       node = G_.createFullyConnected(opName, in, W, B);
+    }
+
+    // If number of original input dims is greater than 2, expand the output
+    // dims back with the same axis.
+    if (axis != 1) {
+      llvm::SmallVector<size_t, max_tensor_dimensions> reshapeDims;
+      size_t totalReshapeSize = 1;
+      for (size_t i = 0; i < axis; ++i) {
+        auto d = originalInputDims[i];
+        reshapeDims.push_back(d);
+        totalReshapeSize *= static_cast<size_t>(d);
+      }
+
+      size_t finalDim = typeName == "FCTransposed" ? wDims.second : wDims.first;
+
+      reshapeDims.push_back(finalDim);
+      totalReshapeSize *= finalDim;
+
+      size_t totalOriginalOutputSize = node->getNthResult(0).getType()->size();
+      RETURN_ERR_IF_NOT(totalReshapeSize == totalOriginalOutputSize,
+                        strFormat("Cannot reshape from size %lu to size %lu",
+                                  totalOriginalOutputSize, totalReshapeSize));
+
+      node = G_.createReshape("fc.out", node, reshapeDims);
     }
 
     // Save the outputs:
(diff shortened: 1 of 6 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -624,6 +628,18 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       node = G_.createFullyConnected(opName, in, W, B);
     }
 
+    // If number of original input dims is greater then expand the output dims
+    // back with the same axis.
+    if (originalInputDims.size() > 2) {
```
yinghai: You should check the whether the number of elements is the same?

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -798,79 +769,81 @@ TEST(caffe2, FCTransposed) {
...
-    for (size_t i = 0; i < expectedValues.size(); ++i) {
-      EXPECT_FLOAT_EQ(expectedValues.at(i), elements.raw(i))
-          << "Where i = " << i;
-    }
-  }
-
-  // We don't actually check that the output is correct, because this is
-  // already covered in the Operator.FCWithFlatten/* tests.
-}
+// TEST(caffe2, FCTransposedWithFlatten) {
```
hyuen: uncomment this?

lib/Importer/Caffe2ModelLoader.cpp
jackm321: If everything is correct it will be but this check is also done by createReshape

tests/unittests/Caffe2ImporterTest.cpp
jackm321: I need to rewrite this test, it's wrong now

lib/Importer/Caffe2ModelLoader.cpp
yinghai: And you get a coredump? We want to avoid coredump.

jackm321:
Still WIP, just putting this up now to unblock @yinghai, I still need to update tests and update the C++ api `Function::createFullyConnected(...)` to also do the right thing as well.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -580,10 +584,10 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     }
 
     // Caffe2 stores the transposed W matrix. In here we first coerce W to a 2D
-    // matrix size if necessay and then transpose it back.
+    // matrix size if necessary and then transpose it back.
     Tensor tmp;
+    auto wDims = flattenCdr(w->dims(), axis_w);
```
rdzhabarov: qq: why do we flatten no matter what (check `w->dims().size() > 2` seems reasonable).

jackm321: This doesn't flatten, it just finds the dims if we do flatten. I put this here so that it's accessible below.

lib/Importer/Caffe2ModelLoader.cpp
jackm321: Yeah right now we do but I have been pushing for putting all those check in to llvm::Errors as well so they will return back to c2. But sure for now I can add an additional check here

yinghai: Is this condition correct? I think it should be conditioned on axis?

jackm321: Do you have an example where this would not work?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -624,6 +628,31 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
     }
 
+    // If number of original input dims is greater than 2, expand the output
+    // dims back with the same axis.
+    if (originalInputDims.size() > 2) {
+
+      llvm::SmallVector<size_t, max_tensor_dimensions> reshapeDims;
+      size_t totalReshapeSize = 1;
+      for (size_t i = 0; i < axis; ++i) {
+        auto d = originalInputDims[i];
```
rdzhabarov: can you use concrete type here, it's not clear why there is static_cast below.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -560,12 +560,16 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     NodeValue in;
     ASSIGN_VALUE_OR_RETURN_ERR(in,
                                getNodeValueOrCreateConstantByName(op.input(0)));
-    if (in.getType()->dims().size() > 2) {
-      size_t axis = 1;
-      if (dict.count("axis")) {
-        ASSIGN_VALUE_OR_RETURN_ERR(axis, loadInt(dict["axis"]));
-      }
 
+    auto originalInputDims = in.getType()->dims();
```
rdzhabarov: Can you document the expected behavior when flattening happens and how the resulted shape after FC should look like.

lib/Importer/Caffe2ModelLoader.cpp
yinghai: It will work in terms of correctness but we don't necessarily need to add Reshape whenever input dim is greater than 2, right?

jackm321: We can do `axis != 1` and this should cover any case where output of conv should not be 2 dims right?

tests/models/caffe2Models/fc_4d_predict_net.pbtxt
```diff
@@ -0,0 +1,17 @@
...
+op {
+  input: "inputs"
+  input: "weights"
+  input: "bias"
+  output: "fc_result"
+  name: ""
+  type: "FC"
+  arg {
+    name: "axis"
+    i: 3
```
yinghai: We can have a case with `axis=1` for contrast.

PR #2530, opened 2019-03-13
Conv3d on interpreter

jackm321:
*Description*:
3D convolution implementation for interpreter backend (quantized and non-quantized).
Next steps include: CPU implementation, import conv3D in c2 model loader, implement backwards passes

*Testing*:
New unit tests added. Some reference outputs are computed in test when easy to do so, the others are from convolution computed by hand.

*Documentation*:
Comments in code

### DIFF
--- a/tests/unittests/GraphTest.cpp
+++ b/tests/unittests/GraphTest.cpp
@@ -101,7 +101,7 @@ TEST(Graph, simpleTestConv3D) {
   Function *F = MD.createFunction("F");
   IRFunction M(F);
   PlaceholderBindings bindings;
-  Node *K = MD.createPlaceholder(ElemKind::FloatTy, {4, 320, 200, 3, 100},
+  Node *K = MD.createPlaceholder(ElemKind::FloatTy, {4, 320, 200, 100, 3},
                                  "input", true);
   K = F->createConv3D(bindings, /* name */ "Conv3D", /* input */ K,
                       /* outChannels */ 16, /* kernel */ 3, /* stride */ 2,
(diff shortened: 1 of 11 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -3199,6 +3267,79 @@ TEST_P(OperatorTest, NonSquarePaddingConvolution) {
...
+  int nextVal = 1;
+  for (size_t i = 0; i < 4; i++) {
+    for (size_t j = 0; j < 4; j++) {
+      for (size_t k = 0; k < 4; k++) {
+        IH.at({0, i, j, k, 0}) = static_cast<float>(nextVal++);
+      } // D
+    }   // W
+  }     // H
+
+  auto filter = mod_.createPlaceholder(ElemKind::FloatTy, {2, 2, 2, 2, 1},
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -3199,6 +3267,79 @@ TEST_P(OperatorTest, NonSquarePaddingConvolution) {
...
+                                       "filter", false);
+  auto FH = bindings_.allocate(filter)->getHandle();
+  for (size_t i = 0; i < 2 * 2 * 2 * 2; i++) {
```
jfix71: nit: could just use `filter->getType()->size()`

tests/unittests/OperatorTest.cpp
```diff
@@ -3504,6 +3645,133 @@ TEST_P(OperatorTest, NonSquareKernelConvolution) {
...
+
+  auto qOutTy = mod_.uniqueType(ElemKind::Int16QTy, {1, 4, 3, 2, 1}, 0.1, 0);
+
+  Convolution3DNode *qCN =
+      F_->createConv3D("q_Conv3D", qInput, qFilter, qBias, qOutTy, {1, 2, 3},
+                       {1, 1, 1}, {0, 0, 0, 0, 0, 0}, 1);
+
+  SaveNode *S = F_->createSave("save", CN);
+
+  auto *deQ = F_->createDequantize("deQ_result", qCN);
```
jfix71: minor nit: Since you're using explicit class names for some of the other nodes in these tests, would be nice to be consistent and not use `auto` here.

tests/unittests/OperatorTest.cpp
```diff
@@ -3587,6 +3855,58 @@ TEST_P(OperatorTest, NonSquareStrideConvolution) {
...
+  int nextVal = 1;
+  for (size_t i = 0; i < 4; i++) {
+    for (size_t j = 0; j < 4; j++) {
+      for (size_t k = 0; k < 4; k++) {
+        IH.at({0, i, j, k, 0}) = static_cast<float>(nextVal++);
+      } // D
+    }   // W
+  }     // H
+
+  auto filter = mod_.createPlaceholder(ElemKind::FloatTy, {1, 2, 2, 2, 1},
```
jfix71: nit `auto *filter`

tests/unittests/GraphTest.cpp
```diff
@@ -183,6 +183,39 @@ TEST(Graph, float16Conv) {
   EXPECT_EQ(convInst->getBias()->getElementType(), ElemKind::Float16Ty);
 }
 
+/// Check that we can create conv3D with float16.
+TEST(Graph, float16Conv3D) {
```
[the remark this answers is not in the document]
jackm321: I was following the example of `float16Conv` but I think it does add some value because instantiating a conv3d with float16 inputs isn't tested elsewhere

PR #2531, opened 2019-03-13
Added new compileFunctions method

gcatron:
*Description*: Added new compileFunctions method which takes an ArrayRef of functions and returns a vector of compiledFunctions. This allows the compile to reason over shared constants between functions. Also added a new unittest and updated backends.md
*Testing*: ninja test
*Documentation*: Updated backends.md
Related to #2045


nickgg:
Hey @gcatron could we have used a Module here instead of an array of functions?

gcatron:
> Hey @gcatron could we have used a Module here instead of an array of functions?
I don't think so, we don't necessarily want to compile all functions in the module.

PR #2533, opened 2019-03-13
Added schema parameter to graph

ayermolo:
*Description*: Right now schema is hard coded in some of the glow function. Changing to use the command line option. Follow up step can be storing it and loading it from profile to avoid potential issue with profile being created under one quantization and used under another.
*Testing*: UnitTests
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -173,7 +173,8 @@ struct Model {
 
       // Quantize the graph based on the captured profile.
       auto *Q = glow::quantization::quantizeFunction(
-          EE_, quantizationInfos, ElemKind::Int8QTy, F_, loweredMap_);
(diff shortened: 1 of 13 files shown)


ayermolo:
@rdzhabarov Is this what you had in mind?

rdzhabarov:
@ayermolo I'm going to merge this. Please, file those issues.

ayermolo:
@rdzhabarov Great, thanks. Much appreciated.
For issue 1) https://github.com/pytorch/glow/issues/2599
For issue 2) https://github.com/pytorch/glow/issues/2600

Although probably just storing min/max would be enough? Then when it is used just use option specified by user to calculate scale/offset. Having both seems redundant. 

I have full plate at the moment, but if no one picks it up in short term would like to add it.

Issue #2537, opened 2019-03-14
Crashes with LLVM built with assertions


pjaaskel:
It seems there are runtime crashes when LLVM is built with assertions on.

This is one of them, but there was at least another trace which I saw:

```
[----------] 9 tests from CPU/BackendCorrectnessTest
[ RUN      ] CPU/BackendCorrectnessTest.convTest/0
Program aborted due to an unhandled Error:
Error value was Success. (Note: Success values must still be checked prior to being destroyed).
#0 0x00000000032d61eb llvm::sys::PrintStackTrace(llvm::raw_ostream&) /home/visit0r/src/tcemc/tce/llvm-build-Debug/release_70/lib/Support/Unix/Signals.inc:490:0
#1 0x00000000032d627e PrintStackTraceSignalHandler(void*) /home/visit0r/src/tcemc/tce/llvm-build-Debug/release_70/lib/Support/Unix/Signals.inc:554:0
#2 0x00000000032d3f8c llvm::sys::RunSignalHandlers() /home/visit0r/src/tcemc/tce/llvm-build-Debug/release_70/lib/Support/Signals.cpp:67:0
#3 0x00000000032d5be5 SignalHandler(int) /home/visit0r/src/tcemc/tce/llvm-build-Debug/release_70/lib/Support/Unix/Signals.inc:353:0
#4 0x00007f4ee9370390 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x11390)
#5 0x00007f4ee8729428 gsignal /build/glibc-LK5gWL/glibc-2.23/signal/../sysdeps/unix/sysv/linux/raise.c:54:0
#6 0x00007f4ee872b02a abort /build/glibc-LK5gWL/glibc-2.23/stdlib/abort.c:91:0
#7 0x0000000003213f0d /home/visit0r/src/tcemc/tce/llvm-build-Debug/release_70/lib/Support/Error.cpp:102:0
#8 0x0000000000560bd7 llvm::Error::assertIsChecked() /home/visit0r/local/stow/llvm-7.0-tcepatched/include/llvm/Support/Error.h:265:3
#9 0x0000000000562600 llvm::Error::operator=(llvm::Error&&) /home/visit0r/local/stow/llvm-7.0-tcepatched/include/llvm/Support/Error.h:209:12
[...]

PR #2539, opened 2019-03-14
[doc] Add a note of LLVM built with assertions.

pjaaskel:
*Description*: Glow does not currently work with assertion ON builds, but crashes at least with some of the test cases. I added a note of this to README.md to perhaps help others who struggle with it when starting with Glow.

*Testing*: Not all unit tests pass for me even with the assertions OFF, but many more do.

See Issue #2537.


jackm321:
Compiling and running with llvm in debug mode is something that should work but seems to be broken since earlier this week. I'd prefer that we fix it and add some CI tests to ensure that it's always working.

jackm321:
I am still planning on working on finding and fixing the existing errors and adding a periodic build with llvm debug CI test but haven't had time to do it yet

PR #2541, opened 2019-03-14
[Training] Add support for training SparseLengthsWeightedSum

SplitInfinity:
**Description**
This commit adds support for training the `SparseLengthsWeightedSum`
operator. The gradient for this node is computed by initializing the
`data` gradient to be zero, and then accumulating into `dataGradient[index]`
the gradient of the result slice that `data[index]` was added to during
the `SparseLengthsWeightedSum` operation multiplied by `weight[index]`
(the same weight that `data[index]` was multiplied by during the
`SparseLengthsWeightedSum` operator).
    
**Testing**
This commit adds a unit test to `MLTest` that trains the data input of a
`SparseLengthsWeightedSum` node so that the operation outputs a
specified `Tensor`. The indices, lengths, weights are fixed.

### DIFF
--- a/examples/char-rnn.cpp
+++ b/examples/char-rnn.cpp
@@ -231,7 +231,7 @@ int main(int argc, char **argv) {
   //// Train the network ////
   Function *F =
       createNetwork(mod, bindings, minibatchSize, numSteps, hiddenSize);
-  Function *TF = differentiate(F, TC);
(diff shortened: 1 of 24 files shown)


SplitInfinity:
Note that this PR is stacked on top of #2524.

PR #2542, opened 2019-03-14
Skip unknown symbols when loading Placeholders.

nickgg:
*Description*: The way we load placeholders right now assumes that all CompiledFunctions know about all symbols in the module (at least when we use `PlaceholderBindings.allocate(module.getPlaceholders())` which we do in a few different places). When loading placeholders we should just skip any that don't have an entry in the symbol table (since they won't be used anyway).
*Testing*: unit tests + the new unit test, which fails on all backends without this change.
*Documentation*: This won't break anything in any other backends, but they probably want to be doing something like this as well.

### DIFF
--- a/tests/unittests/BackendTest.cpp
+++ b/tests/unittests/BackendTest.cpp
@@ -254,6 +254,66 @@ TEST_P(BackendTest, simplePlaceholderValue) {
   EXPECT_TRUE(STensor->isEqual(data));
 }
 
+/// Add and compile a network, then add and compile another so that the first
+/// CompiledFunction does not know about every Placeholder in the module.
+TEST_P(BackendTest, compileThenAddNetwork) {
+  PlaceholderBindings bindings1, bindings2;
+
+  auto &mod = EE_.getModule();
+  Tensor inputs(ElemKind::FloatTy, {1, 10, 10, 3});
+  inputs.getHandle().randomize(-2, 2, mod.getPRNG());
+
+  // Create a simple graph that uses some placeholders.
+  Function *F = mod.createFunction("main");
+  auto *input =
+      mod.createPlaceholder(ElemKind::FloatTy, {1, 10, 10, 3}, "in", false);
+
+  auto *ex = mod.createPlaceholder(ElemKind::Int64ITy, {1, 1}, "exp", false);
+
+  auto *FC = F->createFullyConnected(bindings1, "FC", input, 30);
+  auto *RL = F->createRELU("RL2", FC);
+  auto *SM = F->createSoftMax("sm", RL, ex);
+  auto *S = F->createSave("ret", SM);
+
+  Placeholder *FC_weights =
+      llvm::dyn_cast<Placeholder>(FC->getWeights().getNode());
+
+  EE_.compile(CompilationMode::Infer, F);
+
+  // Recreate that graph in a different Function.
+  Function *F2 = mod.createFunction("other");
+  auto *input2 =
+      mod.createPlaceholder(ElemKind::FloatTy, {1, 10, 10, 3}, "in", false);
+
+  auto *ex2 = mod.createPlaceholder(ElemKind::Int64ITy, {1, 1}, "exp", false);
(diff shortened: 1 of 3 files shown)


tests/unittests/BackendTest.cpp
```diff
@@ -254,6 +254,55 @@ TEST_P(BackendTest, simplePlaceholderValue) {
...
+  auto *ex = mod.createPlaceholder(ElemKind::Int64ITy, {1, 1}, "exp", false);
+
+  auto *FC = F->createFullyConnected(bindings1, "FC", input, 30);
+  auto *RL = F->createRELU("RL2", FC);
+  auto *SM = F->createSoftMax("sm", RL, ex);
+  auto *S = F->createSave("ret", SM);
+
+  EE_.compile(CompilationMode::Infer, F);
+
+  // Recreate that graph in a different Function.
```
SplitInfinity: Why not use `Function::clone` here instead of explicitly building the exact same graph again? It also takes an optional `map<Node*, Node*>` that you can use to get references to the output of one `Placeholder` replaced with references to the output of another (so that you can make sure the cloned Function connects to different `Placeholders`).

You can get a pointer to the final `Save` node using `Function::getNodeByName` or `Function::getNodes`.

tests/unittests/BackendTest.cpp
```diff
@@ -254,6 +254,55 @@ TEST_P(BackendTest, simplePlaceholderValue) {
...
+
+  auto *ex2 = mod.createPlaceholder(ElemKind::Int64ITy, {1, 1}, "exp", false);
+  auto *FC2 = F2->createFullyConnected(bindings2, "FC", input2, 30);
+  auto *RL2 = F2->createRELU("RL2", FC2);
+  auto *SM2 = F2->createSoftMax("sm", RL2, ex2);
+  auto *S2 = F2->createSave("ret", SM2);
+
+  EE_.compile(CompilationMode::Infer, F2, /* clearOtherFunctions */ false);
+
+  // Allocate all placeholders!
```
SplitInfinity: `!` --> `.`

tests/unittests/BackendTest.cpp
```diff
@@ -254,6 +254,55 @@ TEST_P(BackendTest, simplePlaceholderValue) {
...
+  bindings1.allocate(mod.getPlaceholders());
+  bindings2.allocate(mod.getPlaceholders());
+  updateInputPlaceholders(bindings1, {input}, {&inputs});
+  updateInputPlaceholders(bindings2, {input2}, {&inputs});
+
+  EE_.run(bindings1, "main");
+  EE_.run(bindings2, "other");
+
+  // The graphs were the same so their outputs should be as well.
+  EXPECT_TRUE(bindings1.get(S->getPlaceholder())
```
SplitInfinity: I don't know how useful this is given that `inputs` and all tensors allocated by `PlaceholderBindings::allocate` are zero-initialized. Consider binding the `Placeholders` to randomly initialized tensors.

tests/unittests/BackendTest.cpp
SplitInfinity: But given that you are using the version of `Function::createFullyConnected` that accepts a reference to a `PlaceholderBindings` object as an argument and adds bindings for the weight/bias `Placeholders` it creates, doing what I suggested might be more trouble than it's worth unless you choose simpler models.

opti-mix:
@nickgg @gcatron  BTW, this PR made me think about something else:

`glow::generateRuntimeBundle` currently allocated memory even for weights and activations that are not used by the `IRFunction`.  This is a waste of memory. I believe we could skip such allocations.

opti-mix:
Created #2548 to track this memory allocation issue.

tests/unittests/BackendTest.cpp
nickgg: I think i'd prefer not to clone for this example for that reason. The goal here is to test that two totally different functions work in the same module.

nickgg:
Sorry about the delay, should have addressed those comments.

PR #2543, opened 2019-03-14
Make sure parent directory of build output files exist

bddppq:


bddppq:
@rdzhabarov I ran into the issue when trying to build glow in a llvm-7 docker image. This probably depends on different cmake versions. (But it took me a while to figure out the issue, the error message was simply "Could not open output file: No such file or directory").

rdzhabarov:
Got it. @bddppq cool fix!

PR #2544, opened 2019-03-14
Make cmake build more friendly for llvm-7

bddppq:


lib/Backends/CPU/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ endif()
 find_program(LLVM_LINK_BIN
              NAMES
                llvm-link-6.0
+               llvm-link-7
```
rdzhabarov: https://github.com/pytorch/glow/blob/master/CMakeLists.txt#L79

might want to switch order here.

bddppq:
We do have llvm-7 docker image that can be used to add llvm-7 builds on CircleCI. (Currently glow is using llvm-6 docker images https://github.com/pytorch/glow/blob/master/.circleci/config.yml#L71-L91)

PR #2546, opened 2019-03-14
Adding default values to Runtime Symbol struct

gcatron:
*Description*: This adds default values to the fields on RuntimeSymbolInfo.
*Testing*: ninja test
*Documentation*: NA

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -24,15 +24,15 @@ namespace runtime {
 /// Contains information for initialization and handling of symbol at runtime.
 struct RuntimeSymbolInfo {
   /// The size in bytes.
-  size_t size;
+  size_t size{0};
   /// Offset in bytes from the base address.
-  size_t offset;
+  size_t offset{0};
   /// Type of symbol.
   Type type;
   /// Is the symbol an input for the function.
-  bool input;
+  bool input{true};
   /// Is the symbol an output for the function.
-  bool output;
+  bool output{true};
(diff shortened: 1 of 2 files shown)


include/glow/Backends/BackendUtils.h
```diff
@@ -24,15 +24,15 @@ namespace runtime {
...
-  size_t size;
+  size_t size{0};
   /// Offset in bytes from the base address.
-  size_t offset;
+  size_t offset{0};
   /// Type of symbol.
   Type type;
   /// Is the symbol an input for the function.
-  bool input;
+  bool input{false};
```
opti-mix: Question: I wonder if `true` is a more natural value. E.g. if you'd have a logic somewhere that says "copy all symbols that are inputs to devices and then copy all symbols that are outputs from devices", the `true` value would result in copying all the required symbol and may be more. But with `false` value, nothing would be copied in either direction.

gcatron: I thought about that too, ultimately my thinking was default to false and if a symbol needs to be copied it should be set to true. I'm good either way though.

opti-mix: > if a symbol needs to be copied it should be set to true. 

Do we have the code that sets it to true if needed? Is this code invoked on every path leading to "execute" or the like?

gcatron: That's a fair point. It is in the runtime but not in the executionengine. I'll change it to true.

PR #2551, opened 2019-03-15
[OperatorTest] More missing tests for Gather/GatherRanges

jfix71:
Added a few more tests that we didn't have for `Gather`/`GatherRanges`:

- `Gather` now has Int8 and Float16 data versions

- `GatherRanges` now has Float and Float16 data tests.

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -1126,9 +1126,11 @@ TEST_P(OperatorTest, QuantizedTopK) {
   EXPECT_EQ(IH.at({2, 0, 2}), 4);
 }
 
-TEST_P(OperatorTest, Gather64) {
-  ENABLED_BACKENDS(Interpreter, CPU);
-
+/// Helper for testing Gather with different \p ITy / \p IndexType.
+template <typename IndexType>
+static void gatherFloatInputTest(glow::PlaceholderBindings &bindings_,
+                                 glow::Module &mod_, glow::Function *F_,
+                                 glow::ExecutionEngine &EE_, ElemKind ITy) {
   /*
     DATA  = [
         [1.0, 1.2],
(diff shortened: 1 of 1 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -1126,9 +1126,11 @@ TEST_P(OperatorTest, QuantizedTopK) {
   EXPECT_EQ(IH.at({2, 0, 2}), 4);
 }
 
-TEST_P(OperatorTest, Gather64) {
-  ENABLED_BACKENDS(Interpreter, CPU);
-
+/// Helper for testing Gather with different \p ITy / \p IndexType.
+template <typename IndexType>
+static void gatherFloatInputTest(glow::PlaceholderBindings &bindings_,
```
SplitInfinity: I think the convention is to user terminal underscores for private variables within the scope of a class, not for function parameter names. Consider removing them.

tests/unittests/OperatorTest.cpp
```diff
@@ -1194,42 +1195,56 @@ TEST_P(OperatorTest, Gather64) {
...
 
+/// Test that Gather works with Float data and Int64 indices.
+TEST_P(OperatorTest, GatherDataFloatIdxInt64) {
+  ENABLED_BACKENDS(Interpreter, CPU);
+  gatherFloatInputTest<int64_t>(bindings_, mod_, F_, EE_, ElemKind::Int64ITy);
+}
+
+/// Helper for testing Gather with different \p ITy / \p IndexType.
+template <typename IndexType>
+static void gatherInt8InputTest(glow::PlaceholderBindings &bindings_,
```
SplitInfinity: Consider removing terminal underscores in the parameter names.

tests/unittests/OperatorTest.cpp
```diff
@@ -1241,31 +1256,43 @@ TEST_P(OperatorTest, Gather32) {
...
 
-  EXPECT_FLOAT_EQ(H.at({0, 0, 0}), 1.0);
-  EXPECT_FLOAT_EQ(H.at({0, 0, 1}), 1.2);
-  EXPECT_FLOAT_EQ(H.at({0, 1, 0}), 2.3);
-  EXPECT_FLOAT_EQ(H.at({0, 1, 1}), 3.4);
-  EXPECT_FLOAT_EQ(H.at({0, 2, 0}), 1.0);
-  EXPECT_FLOAT_EQ(H.at({0, 2, 1}), 1.2);
-  EXPECT_FLOAT_EQ(H.at({0, 3, 0}), 2.3);
-  EXPECT_FLOAT_EQ(H.at({0, 3, 1}), 3.4);
+  EXPECT_EQ(H.at({0, 0, 0}), 1);
```
SplitInfinity: Consider creating an expected result `Tensor`, initializing it with the values in `OUTPUT` in the comment above, and then replacing this series of `EXPECT_EQ` statements with either a for-loop or a call to `Tensor::isEquals`:

jfix71:
@SplitInfinity addressed comments. Also added `Float16` tests that I realized were missing.

PR #2564, opened 2019-03-18
[LLVMIRCodeGen] A bunch of small improvements

opti-mix:
*Description*:

This is a set of small LLVMIRCodeGen improvements. 

The most important one is the ability to specify target, arch, cpu and cpu features for selecting the LLVM target machine. This is important for LLVM targets which have a lot of different configuration variants available.

List of changes:
* Move AllocationsInfo.h into the shared includes directory
    This makes it possible to use AllocationsInfo by various backends.

* Move command-line options for selecting LLVM target, arch, cpu and cpu features into a dedicated include file
   Use these options for a more fine-grained way to configure the TargetMachine.

* Make more methods in LLVMBackend and LLVMIRGen virtual, so that backends can provide custom implementations


lib/LLVMIRCodeGen/CommandLine.h
```diff
@@ -20,5 +20,13 @@
 #include "llvm/Support/CommandLine.h"
 
 llvm::cl::OptionCategory &getLLVMBackendCat();
+/// Target to be used by the LLVMBackend.
+extern llvm::cl::opt<std::string> llvmTarget;
+/// Architecture to be used by the LLVMBackend.
+extern llvm::cl::opt<std::string> llvmArch;
+/// CPU to be used by the LLVMBackend.
+extern llvm::cl::opt<std::string> llvmCPU;
```
rdzhabarov: can you put more comments here. I'm not sure I understand the difference between `arch` and `cpu`. What `cpu` exactly is?

opti-mix: Done.

PR #2566, opened 2019-03-19
[Backends] Autodiscovery of registered backends. Minimize changes in Backend CMake.

rdzhabarov:
*Description*:
Move backend and device manager additions to the corresponding backend dirs.
Autodiscovery of backend.

*Testing*:
CI

*Documentation*:
n/a

Addressing some @opti-mix offline comments.

### DIFF
--- a/lib/Backends/CMakeLists.txt
+++ b/lib/Backends/CMakeLists.txt
@@ -1,20 +1,16 @@
-add_subdirectory(Interpreter)
-
-LIST(APPEND linked_device_managers InterpreterDeviceManager)
-LIST(APPEND linked_factories InterpreterFactory)
-
-if(GLOW_WITH_OPENCL)
-  add_subdirectory(OpenCL)
-  LIST(APPEND linked_factories OpenCLFactory)
-  LIST(APPEND linked_device_managers OpenCLDeviceManager)
-endif()
-
-
-if(GLOW_WITH_CPU)
-  add_subdirectory(CPU)
-  LIST(APPEND linked_factories CPUFactory)
-  LIST(APPEND linked_device_managers CPUDeviceManager)
-endif()
+# Iterate all subdirectories and check whether
+# particular backend is enabled.
+FILE(GLOB subdirs RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/*)
+foreach(object ${subdirs})
+  if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${object})
+    set(backendEnabled GLOW_WITH_${object})
+    string(TOUPPER ${backendEnabled} backendEnabled)
+    if(NOT DEFINED ${backendEnabled} OR ${backendEnabled})
+      message("Adding ${object} backend.")
+      add_subdirectory(${object})
+    endif()
+  endif()
+endforeach()
 
 add_library(Backend
               Backend.cpp
--- a/lib/Backends/CMakeLists.txt
+++ b/lib/Backends/CMakeLists.txt
@@ -1,10 +1,14 @@
-# Iterate all subdirectories and check whether
-# particular backend is enabled.
+# Iterate all subdirectories and check whether particular backend is enabled.
 FILE(GLOB subdirs RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/*)
 foreach(object ${subdirs})
   if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${object})
     set(backendEnabled GLOW_WITH_${object})
     string(TOUPPER ${backendEnabled} backendEnabled)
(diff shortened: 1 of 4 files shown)


opti-mix:
@rdzhabarov This looks better. But can we get rid of 
```
if(GLOW_WITH_CPU)
  add_subdirectory(CPU)
endif ()
```
as well?

May be CMake can scan which subdirectories are available under `Backends` and if a subdirectory with name `X` is present and `GLOW_WITH_X` is defined, then cmake should perform `add_subdirectory(X)`. This way we don't need to change anything in the `Backends/CMakeLists.txt` if a new backend is added.

It would be also nice if we wouldn't need to explicitly add `GLOW_WITH_X` flags in the main `CMakeLists.txt` and define them in the backend-subdirectories instead.

But may be I'm going too far with my wishes ;-)

rdzhabarov:
> if(GLOW_WITH_CPU)
  add_subdirectory(CPU)
endif ()

Yeah, that would be ideal. I'll see if I can do some CMake magic.

rdzhabarov:
@opti-mix I'm pretty happy with the current state after these changes. good to go?

rdzhabarov:
> It would be also nice if we wouldn't need to explicitly add GLOW_WITH_X flags in the main CMakeLists.txt and define them in the backend-subdirectories instead.

I do not think it's possible. But what we could do is to add_subdirectory automatically if GLOW_WITH_X is not defined. If it's defined we only add_subdirectory if it's 'ON'. So you could 'OFF' when needed.

rdzhabarov:
@opti-mix I'm pretty happy with latest changes. By default, folder will be added if it's under /Backends/XXX. Otherwise we'd look at GLOW_WITH_XXX, whether it's enabled or not.

lib/Backends/CMakeLists.txt
```diff
@@ -1,20 +1,16 @@
...
-endif()
-
-
-if(GLOW_WITH_CPU)
-  add_subdirectory(CPU)
-  LIST(APPEND linked_factories CPUFactory)
-  LIST(APPEND linked_device_managers CPUDeviceManager)
-endif()
+# Iterate all subdirectories and check whether
+# particular backend is enabled.
```
opti-mix: Strange formatting. Why 2 lines instead of one?

lib/Backends/CMakeLists.txt
```diff
@@ -1,20 +1,16 @@
...
+FILE(GLOB subdirs RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/*)
+foreach(object ${subdirs})
+  if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/${object})
+    set(backendEnabled GLOW_WITH_${object})
+    string(TOUPPER ${backendEnabled} backendEnabled)
+    if(NOT DEFINED ${backendEnabled} OR ${backendEnabled})
```
opti-mix: This line deserves a comment. Just repeat what you stated in the PR comments about the logic here.

lib/Backends/CMakeLists.txt
rdzhabarov: no particular reason. can be made as a single line.

lib/Backends/CMakeLists.txt
```diff
@@ -1,20 +1,20 @@
...
-  LIST(APPEND linked_device_managers OpenCLDeviceManager)
-endif()
-
-
-if(GLOW_WITH_CPU)
-  add_subdirectory(CPU)
-  LIST(APPEND linked_factories CPUFactory)
-  LIST(APPEND linked_device_managers CPUDeviceManager)
-endif()
+# Iterate all subdirectories and check whether particular backend is enabled.
```
nadavrot: I am not sure what will be more painful. Merge conflicts on the left or porting the code on the right to BUCK and other build system. Why is the code on the right better?

bertmaher:
I don’t have strong feelings about the organization of the cmake build, since it isn’t used by the internal build system :-).  I probably wouldn’t bother porting this logic to our Buck build, since we can just explicitly list the targets we care about without stepping on any OSS code. 

If somebody is using CMake for their internal development this setup might be a smidge nicer since it avoids needing any local changes to the Open-sourced CMake files.

rdzhabarov:
> I am not sure what will be more painful. Merge conflicts on the left or porting the code on the right to BUCK and other build system. Why is the code on the right better?

I think this is logical follow up on having backend to be statically registered. This change allow just to drop custom backend under Backends/ folder and get it running. The benefit is simplicity and decoupling core from specific backend.
There are no implication on BUCK or other build systems. This PR does not have any code changes so it's build system specific.

rdzhabarov:
I agree with @bertmaher, we do not necessarily need to apply the same mechanism in fbcode, but it's nice to have this in CMake and open source part.

opti-mix:
BTW, @tlepley-cadence and @aturetsk told me off-line that they may have a couple more ideas how to make integrating new backends even easier. It would be nice if they could provide more details.

PR #2569, opened 2019-03-20
[test] Legalize a name when the placeholer is updated by name.

### LINKED ISSUE #2565: Placeholder unique name
For each Placeholder, Glow will register a unique name for it. However, it may cause the function `updateInputPlaceholdersByName` failed. 
E.g. In Resnet50, the input is `gpu_0/data_0`, but it will be changed into `gpu_0_data_0`. When we want to use the following code to fit the input,  we will get an error "There is no tensor registered with name". 
```
 Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"gpu_0/data_0"}, 
                                                      {&data.getType()}, *F);
...
 updateInputPlaceholdersByName(bindings, &mod, {"gpu_0/data_0"}, {&data});
```
It looks to me that `updateInputPlaceholdersByName` is used in Glow's unittests only.  But I think it is better to do the same conversion.

beicy:
*Description*:
Since one placeholder's name will be legalized before registration, when we update one placeholder by name, the name should be processed in the same way. Otherwise, finding the placeholder may be failed.

*Testing*:
unittest.

*Documentation*:
[Optional Fixes #2565]

### DIFF
--- a/include/glow/Support/Support.h
+++ b/include/glow/Support/Support.h
@@ -111,6 +111,11 @@ void report(const char *msg);
 inline void report(const std::string &str) { report(str.c_str()); }
 inline void report(llvm::StringRef str) { report(str.data()); }
 
+/// Legalize \p name used in Module. In GLow module, the name of placeholders
+/// and constants should look like valid C identifiers. Therefore, those symbols
+/// can be inspected under debugger.
+std::string legalizeName(llvm::StringRef name);
+
 /// Printf-like formatting for std::string.
 const std::string strFormat(const char *format, ...)
 #ifndef _MSC_VER
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -103,7 +103,7 @@ void glow::updateInputPlaceholdersByName(PlaceholderBindings &bindings,
          "The number of inputs does not match the number of Placeholders");
 
   for (int i = 0, e = ph.size(); i < e; i++) {
-    Placeholder *p = mod->getPlaceholderByName(ph[i]);
+    Placeholder *p = mod->getPlaceholderByName(legalizeName(ph[i]));
     Tensor *t = inputs[i];
     assert(t && "Invalid tensor.");
     assert(p && "Invalid placeholder.");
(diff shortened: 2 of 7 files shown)


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -103,7 +103,13 @@ void glow::updateInputPlaceholdersByName(PlaceholderBindings &bindings,
          "The number of inputs does not match the number of Placeholders");
 
   for (int i = 0, e = ph.size(); i < e; i++) {
-    Placeholder *p = mod->getPlaceholderByName(ph[i]);
+    // Legalize the name.
+    std::string legalName;
```
rdzhabarov: there are several places we do name legalization (see Graph.cpp), I'd suggest to place this in a Support lib and reuse in all those places.

include/glow/Support/Support.h
```diff
@@ -111,6 +111,9 @@ void report(const char *msg);
 inline void report(const std::string &str) { report(str.c_str()); }
 inline void report(llvm::StringRef str) { report(str.data()); }
 
+/// Legalize \p name used in Module.
```
rdzhabarov: could you add quick info here about what `legalize` would mean in this context.

lib/Support/Support.cpp
```diff
@@ -99,4 +99,22 @@ const std::string strFormat(const char *format, ...) {
   va_end(vaArgs);
   return std::string(str.data(), len);
 }
+
+std::string legalizeName(llvm::StringRef name) {
```
rdzhabarov: can you add quick test for this

beicy: I modified the Caffe2ImporterTest.cpp test to verify this case. Do you mean we need to write a separate test for the function only?

rdzhabarov: yup, to cover those cases when it does not start with alpha, when it's empty etc.

rdzhabarov:
to fix `expensive_tests` run you'd need to rebase on latest master.

lib/Support/Support.cpp
beicy: Tests added.

PR #2570, opened 2019-03-20
[OpenCL] Account for a case were device max WG size limits dims.

pjaaskel:
*Description*:

In OpenCL, the device can define maximum sizes separately for each dimension, but the total maximum WG size can be less than the multiple of all of the dimension maximums. In that case it simply means that not all dimensions can be set at their maximum at the same time. This patch accounts for that and reduces the second dimension to meet the maximum WG size limit of the device.

*Testing*: ninja check   results match the situation without the change, these fail/skip (LLVM 8, no asserts build) with both:

```
          1 - onnx_gtests (Not Run)
          2 - onnxifi_test_driver_gtests (Not Run)
         30 - QuantizationTest (Child aborted)
```


opti-mix:
@pjaaskel Quick question: Would it be possible to provide a small Glow test-case that demonstrates the issue? May be an operation on tensors of specific dimensions that would result in a problem without your fix?

pjaaskel:
It might be a little difficult as it's also OpenCL device-specific.

opti-mix:
> It might be a little difficult as it's also OpenCL device-specific.

OK. I'll merge the fix as is. If you come up with a test we can add in a subsequent PR.

PR #2571, opened 2019-03-20
[OperatorTests] Fill in many missing tests across various ElemKinds

jfix71:
We have many operators which are not being tested with different ElemKinds -- most gaps in fp16/int8. I tried to address many of the gaps here, reusing the existing logic for the test as much as possible.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1848,7 +1848,8 @@ Node *Function::createWeightedSum(llvm::StringRef name,
 }
 
 Node *Function::createBatchBoxCox(llvm::StringRef name, NodeValue data,
-                                  NodeValue lambda1, NodeValue lambda2) {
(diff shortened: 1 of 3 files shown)


lib/Graph/Graph.cpp
```diff
@@ -1364,7 +1364,11 @@ BatchedReduceAddNode *
 Function::createBatchedReduceAdd(llvm::StringRef name, NodeValue batch,
                                  llvm::ArrayRef<unsigned_t> axes) {
   auto outDims = getNewShapeWithoutAxes(batch.dims(), axes);
-  auto OT = getParent()->uniqueType(batch.getType()->getElementType(), outDims);
+  auto bTy = batch.getType();
+  auto OT = isQuantizedElemKind(bTy->getElementType())
```
rdzhabarov: could we use: Type::newShape(const Type &T, llvm::ArrayRef<size_t> dims), that takes care about quantization.

PR #2572, opened 2019-03-20
Pass ownership of Module to HostManager

### LINKED ISSUE #2567: [Runtime] HostManager::addNetwork() should take ownership of Module
`HostManager::addNetwork()` should be passed a `Module` as a unique_ptr to reflect the fact that it is actually taking ownership of and modifying the `Module`.

jackm321:
*Description*:
Make it so that HostManager and runtime owns the Module. HostManager will take ownership of Module in `addNetwork` and keeps the PlaceHolders around internally.

*Testing*:
`ninja check`
Run resnet-runtime example

*Documentation*:
comments
Fixes #2567

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -43,6 +43,15 @@ class Provisioner;
 /// handles DeviceManager initialization, houses the Executor, and calls into
 /// the Partitioner and Provisioner for network initialization.
 class HostManager final {
+  /// NetworkData contains data about each network in HostManager that is needed
+  /// by the runtime.
+  struct NetworkData {
+    DAG dag;
+    // Module that was used to create this network. Everything except
+    // placeholders and types have been removed from it.
+    std::shared_ptr<Module> module;
+  };
+
   /// Count of current in-flight networks being run. Atomic to allow
   /// concurrency in runNetwork.
   std::atomic<size_t> activeRequestCount_{0};
@@ -57,7 +66,7 @@ class HostManager final {
   const unsigned int activeRequestLimit_ = 100;
 
   /// A map from a networkName to a network, which is represented by struct DAG.
-  std::unordered_map<std::string, DAG> networks_;
+  std::unordered_map<std::string, NetworkData> networks_;
 
   /// Mutex for networks_ since runNetwork, addNetwork, and
   /// removeNetwork can all be called concurrently, a guard is needed.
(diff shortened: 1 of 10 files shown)


include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -43,6 +43,20 @@ class Provisioner;
 /// handles DeviceManager initialization, houses the Executor, and calls into
 /// the Partitioner and Provisioner for network initialization.
 class HostManager final {
+  /// NetworkData contains data about each network in HostManager that is needed
+  /// by the runtime.
+  struct NetworkData {
+    DAG dag;
+    // list of placeholders owned by the NetworkData.
+    PlaceholderList placeholders;
+
```
beicy: So why constants won't be needed in this struct?

jackm321: Collect constants will copy them
https://github.com/pytorch/glow/blob/master/include/glow/Backends/BackendUtils.h#L70-L77

gcatron:
Why do we need to keep the placeholders and module around after loading the network? It seems storing the placeholders is partially duplicating what the runtimeBundle does.

jackm321:
Module is not kept around it gets thrown out at the end of `addNetwork()` but the Placeholders were owned by the Module so those need to get kept or else they'll be freed with the Module as well. Where does RuntimeBundle keep Placeholders?

jackm321:
Oh I see it allocates symbols out of them. How would that actually work when the Context that's created for each run is created by code that only has original placeholders from compile time?

jackm321:
Discussed this offline with @gcatron, we will keep the Placeholders around during runtime for now as keys into the RuntimeBundle's allocations for the placeholders. In the longer term though, we should switch to replacing placeholders with just their names throughout the runtime since there's no reason to actually keep the placeholders around from compilation other than as a unique identifier.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -78,19 +78,23 @@ llvm::Error HostManager::addNetwork(Module *M) {
...
+  RETURN_IF_ERR(provisioner_->provision(nodeList, *module));
 
   for (auto &node : nodeList) {
-    networks_.emplace((node.root)->name, std::move(node));
+    auto &networkData = networks_[(node.root)->name];
+    networkData.dag = std::move(node);
+    networkData.placeholders = (module->takePlaceholders());
   }
 
+  modules_.push_back(std::move(module));
```
gcatron: Do we need to store the module here? If so when do we free it?

jackm321: Yes good catch, we shouldn't need this now.

PR #2575, opened 2019-03-20
[fix] Fix internal build.

rdzhabarov:
*Description*:
Fix fbcode build with the latest Glow.

*Testing*:
CI

*Documentation*:
n/a

### DIFF
--- a/include/glow/LLVMIRCodeGen/GlowJIT.h
+++ b/include/glow/LLVMIRCodeGen/GlowJIT.h
@@ -54,7 +54,7 @@ private:
   ExecutionSession ES_;
   std::shared_ptr<SymbolResolver> resolver_;
 #endif
-#if LLVM_VERSION_MAJOR < 8
+#if LLVM_VERSION_MAJOR < 8 || FACEBOOK_INTERNAL
   RTDyldObjectLinkingLayer objectLayer_;
   IRCompileLayer<decltype(objectLayer_), SimpleCompiler> compileLayer_;
 #else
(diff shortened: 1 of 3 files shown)


opti-mix:
@rdzhabarov `FACEBOOK_INTERNAL` is very fluffy. What will happen when FB updates the internal LLVM  to use e.g. LLVM >= 8?

Does `FACEBOOK_INTERNAL` has any kind of reasonable value, like 6 or 7 (reflecting which LLVM is used), which could be used for more precise checks?

jackm321:
fyi the `RELEASE_WITH_EXPENSIVE_TESTS` should be fixed for real (hopefully) this time with #2576

rdzhabarov:
> @rdzhabarov FACEBOOK_INTERNAL is very fluffy. What will happen when FB updates the internal LLVM to use e.g. LLVM >= 8?

> Does FACEBOOK_INTERNAL has any kind of reasonable value, like 6 or 7 (reflecting which LLVM is used), which could be used for more precise checks?

Agreed. It's not very transparent, but something that was used before (see other entries of that).
I'll see if we can eliminate that entirely.

bertmaher:
> What will happen when FB updates the internal LLVM to use e.g. LLVM >= 8?

Internally when the LLVM team wants to update, they run the update on CI, look at what fails, and then harass all the service owners to update their stuff (last time it happened to be me oncall :) ).  Unfortunately our internal build does not neatly correspond to any released LLVM, or at least it didn't several months ago.

> Does FACEBOOK_INTERNAL has any kind of reasonable value, like 6 or 7 (reflecting which LLVM is used), which could be used for more precise checks?

There's an LLVM_PATCH_VERSION that FB sets according to its own whims, but AFAIK it's not tied to anything meaningful in the outside world.

rdzhabarov:
There is a mix of LLVM versions used, I see both 7 and 8.0.20181009svn.

The fix above is required to run on both. Since updates are not really frequent and internal packages might be different I think keeping FB_INTERNAL seems a plausible solution.

@opti-mix let me know.

PR #2577, opened 2019-03-21
User can pass an input image list file containing list of images

suyash-cadence:
Image list file is used to specify the list of images that user wants to infer.

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -27,6 +27,7 @@
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <fstream>
 #include <iostream>
 #include <memory>
 #include <queue>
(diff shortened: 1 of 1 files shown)


tools/loader/ImageClassifier.cpp
```diff
@@ -332,12 +332,37 @@ static void processAndPrintResults(Tensor *SMT, llvm::StringRef functionName) {
...
+}
+
 int main(int argc, char **argv) {
   Context ctx;
   // The loader verifies/initializes command line parameters, and initializes
   // the ExecutionEngine and Function.
   Loader loader(argc, argv);
 
+  std::string ext(".txt");
+  if (inputImageFilenames[0].rfind(ext) ==
```
rdzhabarov: I'd just pass it as a separate arg (inputImagesListFile or something) and do not worry about specific extension etc.

rdzhabarov:
Please also sign the CLA above.

tools/loader/ImageClassifier.cpp
suyash-cadence: Since, inputImageFilenames has 'llvm::cl::OneOrMore'  attribute, I am obliged to reuse it. I am checking extension to differentiate between .png image files and list.txt file.

Just to clarify @rdzhabarov : Are you suggesting to create additional command line option 'inputImagesListFile'?

rdzhabarov: Yeah, additional one. Just make inputImageFilenames to comply with 0 entries.

tools/loader/ImageClassifier.cpp
```diff
@@ -319,12 +327,34 @@ static void processAndPrintResults(Tensor *SMT, llvm::StringRef functionName) {
...
+    }
+  }
+  inFile.close();
+}
+
 int main(int argc, char **argv) {
   PlaceholderBindings bindings;
   // The loader verifies/initializes command line parameters, and initializes
   // the ExecutionEngine and Function.
   Loader loader(argc, argv);
```
jfix71: Can you add a conditional here checking if `inputImageFilenames.size() == 0 && inputImageListFile.empty()`, and then print to `llvm::errs()` with an explanation that one of them must be used, and then `std::exit(1);`. We do something like this in `Loader.cpp` with `commandLineIsInvalid()`.

suyash-cadence: Sure, makes sense to add.

Issue #2581, opened 2019-03-23
[quantization] Quantized resnext results differ on Interpreter and CPU backends


artemrakhov:
This is what I see on Interpreter backend:
```
Model: /home/artem/glow/release_build/resnext_int8/predict_net.pbtxt
 File: tests/images/imagenet/cat_285.png	Label-K1: 281 (probability: 0.5911)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.8373)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9986)
```
On CPU backend:
```
Model: /home/artem/glow/release_build/resnext_int8/predict_net.pbtxt
 File: tests/images/imagenet/cat_285.png	Label-K1: 281 (probability: 0.7793)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9691)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9996)
```
Although labels are correct, probabilities are very different, which needs to be investigated.

jfix71:
I saw something similar a while back when looking at quantized Resnet50: https://github.com/pytorch/glow/issues/1837#issuecomment-429091696

The cause here for quantized Resnext could be something different, but at the time for quantized Resnet50 I assumed this was likely due to differences in quantization implementation. A good place to start would be to compare the first Conv on each backend. We should also run top-k on both backends to compare.

artemrakhov:
I debugged this and found that the difference is indeed in the fist quantized Convolution layer. We have very different Rescaling formulas, for converting accumulated int32 to int8 _(stored and passed further)_

1. Interpreter simply does: `round(int32 * fp)`
https://github.com/pytorch/glow/blob/0f61083a78ec49ee6389958613ffafe86290cf76/lib/Backends/Interpreter/InterpreterNodes.cpp#L242

2. CPU does: `((int32 >> preShift) * uint8) >> postShift`
https://github.com/pytorch/glow/blob/2a8f613aaa642e9cb6c03b10b1c2d56a85bd85ef/lib/Quantization/Base/Base.cpp#L197

3. Caffe2 does: `roundingShift(int32 * int, shift)`
https://github.com/pytorch/FBGEMM/blob/23aa5a35ece04fe31600cce31712214a133166ec/src/QuantUtils.cc#L101

[...]

PR #2582, opened 2019-03-23
[CPU backend] Fix dump_tensor in libjit.

artemrakhov:
Now CPU backend should be able to print results correctly on quantized resnext, which will help with issue #2581


lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1588,24 +1588,36 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
                    size_t elemKind, const char *name) {
   printf("%s\n", name);
   /// This definition should match the defintion in Glow.
```
rdzhabarov: good catch! this is extremely easy to miss as it's likely we'd update ElemKind in Type.h first.
Can you add a comment there (type.h) that we need to take care about this definition in libjit as well.

PR #2583, opened 2019-03-23
Command line option to pass mean, standard deviation for input normalization

suyash-cadence:
User has the control over the normalization

### DIFF
--- a/include/glow/Base/Image.h
+++ b/include/glow/Base/Image.h
@@ -61,6 +61,17 @@ extern ImageLayout imageLayout;
 /// -use-imagenet-normalization flag.
 extern bool useImagenetNormalization;
 
+/// These are standard normalization factors for imagenet, adjusted for
+/// normalizing values in the 0to255 range instead of 0to1, as seen at:
+/// https://github.com/pytorch/examples/blob/master/imagenet/main.py
+static const float imagenetNormMean[] = {0.485 * 255.0, 0.456 * 255.0,
+                                         0.406 * 255.0};
+static const float imagenetNormStd[] = {0.229, 0.224, 0.225};
+
+/// Default values for mean and stddev.
+static const std::vector<float> zeroMean(max_tensor_dimensions, 0.f);
+static const std::vector<float> oneStd(max_tensor_dimensions, 1.f);
+
 /// \returns the floating-point range corresponding to enum value \p mode.
 std::pair<float, float> normModeToRange(ImageNormalizationMode mode);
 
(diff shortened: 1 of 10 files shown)


include/glow/Base/Image.h
```diff
@@ -86,11 +87,15 @@ bool writePngImage(Tensor *T, const char *filename,
...
 /// \param useImagenetNormalization use special normalization for Imagenet.
+/// \param mean use special mean to normalize.
+/// \param stdev use special stddev to normalize.
 Tensor readPngImageAndPreprocess(llvm::StringRef filename,
                                  ImageNormalizationMode imageNormMode,
                                  ImageChannelOrder imageChannelOrder,
                                  ImageLayout imageLayout,
-                                 bool useImagenetNormalization);
+                                 bool useImagenetNormalization,
+                                 std::vector<float> &mean,
```
nadavrot: Please use ArrayRef

lib/Base/Image.cpp
```diff
@@ -143,7 +143,8 @@ static constexpr float imagenetNormMean[] = {0.485 * 255.0, 0.456 * 255.0,
 static constexpr float imagenetNormStd[] = {0.229, 0.224, 0.225};
 
 bool glow::readPngImage(Tensor *T, const char *filename,
-                        std::pair<float, float> range,
+                        std::pair<float, float> range, std::vector<float> &mean,
+                        std::vector<float> &stddev,
```
nadavrot: ArrayRef

lib/Base/Image.cpp
```diff
@@ -358,11 +362,13 @@ Tensor glow::readPngImageAndPreprocess(llvm::StringRef filename,
                                        ImageNormalizationMode imageNormMode,
                                        ImageChannelOrder imageChannelOrder,
                                        ImageLayout imageLayout,
-                                       bool useImagenetNormalization) {
+                                       bool useImagenetNormalization,
+                                       std::vector<float> &mean,
```
nadavrot: ArrayRef

suyash-cadence:
> Hi @suyash-cadence, thanks for the PR. I think we could simplify this instead of needing both `useImagenetNormalization` and `mean`/`std`.
> 
> What if we remove the `useImagenetNormalization` argument entirely, and instead pass two `ArrayRef<float>` for `mean` and `std`. We can make them default arguments, e.g. set `mean` to `std::vector<float>(max_tensor_dimensions, 1.f)` (and `0.f` for `std`). This way we don't need every caller to be updated, and by default no normalization would occur.
> 
> Then we could move `imagenetNormMean` and `imagenetNormStd` to `Image.h` and anyone who wants to use imagenet normalization would just provide those two in place of the default arguments.
> 
> What do you think?


I agree with those improvements. Let me do those changes and update.

include/glow/Base/Image.h
```diff
@@ -55,8 +55,16 @@ extern ImageChannelOrder imageChannelOrder;
...
-extern bool useImagenetNormalization;
+/// These are standard normalization factors for imagenet, adjusted for
+/// normalizing values in the 0to255 range instead of 0to1, as seen at:
+/// https://github.com/pytorch/examples/blob/master/imagenet/main.py
+static const float imagenetNormMean[] = {0.485 * 255.0, 0.456 * 255.0,
+                                         0.406 * 255.0};
+static const float imagenetNormStd[] = {0.229, 0.224, 0.225};
+
+/// default values for mean and stddev
+static const float zeroMean[] = {0.f, 0.f, 0.f};
```
jfix71: Why not `static const std::vector<float> zeroMean(max_tensor_dimensions, 0.f);` -- what if someone e.g. extends this with including the alpha channel?

tools/loader/ImageClassifier.cpp
```diff
@@ -100,11 +114,30 @@ void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
...
 
+  // assign mean and stddev for input normalization
+  llvm::ArrayRef<float> mean;
+  llvm::ArrayRef<float> stddev;
+  if (!meanValues.empty()) {
+    GLOW_ASSERT(meanValues.size() == numChannels &&
+                "Number of mean values != input channels");
+    mean = meanValues;
+  } else {
+    mean = std::vector<float>(numChannels, 0.f);
```
jfix71: This is going to cause a use after free. The vector here is constructed locally and will be destroyed after we leave the else scope, leaving the `mean` ArrayRef pointing to dealloc'd memory.

You can instead just reuse `meanValues`/`stddevValues` -- I'm not sure if you can set them directly (e.g. `meanValues = std::vector<float>(numChannels, 0.f);`), but at least `llvm::cl::list` does have a `addValue()` that you can use to append either the default values, or `imagenetNormMean`/`imagenetNormStd` if `useImagenetNormalization`.

lib/Base/Image.cpp
```diff
@@ -71,15 +71,6 @@ static llvm::cl::alias imageLayoutA("l",
                                     llvm::cl::desc("Alias for -image-layout"),
                                     llvm::cl::aliasopt(imageLayoutF),
                                     llvm::cl::cat(imageCat));
-
-bool useImagenetNormalization;
```
jfix71: Can you actually just move this to `ImageClassifier.cpp`, so the user doesn't have to specify the actual values of `imagenetNormMean`/`imagenetNormStd` on the command line? Then when setting `meanValues`/`stddevValues` inside `loadImagesAndPreprocess()` from `ImageClassifier.cpp` you can set the values correctly if `useImagenetNormalization`.

include/glow/Base/Image.h
```diff
@@ -66,31 +74,32 @@ std::pair<float, float> normModeToRange(ImageNormalizationMode mode);
...
 
 /// Reads a png image. \returns True if an error occurred. The values of the
-/// image are in the range \p range. If \p useImagenetNormalization then
-/// specialized normalization for Imagenet images is applied to the image.
-/// \pre !(useImagenetNormalization && numChannels != 3)
+/// image are in the range \p range.
 bool readPngImage(Tensor *T, const char *filename,
                   std::pair<float, float> range,
-                  bool useImagenetNormalization = false);
+                  const llvm::ArrayRef<float> &mean = zeroMean,
```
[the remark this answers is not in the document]
nadavrot: Agree.  ArrayRef is already immutable. A different class MutableArrayRef is mutable.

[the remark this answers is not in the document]
suyash-cadence: @nadavrot Thanks for clarifying.

tools/loader/ImageClassifier.cpp
```diff
@@ -70,6 +70,20 @@ llvm::cl::opt<bool> convertInAndOutToFp16(
         "Convert the input and output tensors of the network to fp16"),
     llvm::cl::cat(imageLoaderCat));
 
+llvm::cl::list<float> meanValues(
```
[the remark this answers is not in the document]
suyash-cadence: Okay, I have added "useImagenetNormalization" option too.

tools/loader/ImageClassifier.cpp
suyash-cadence: Now using `zeroMean` instead of `std::vector<float>(numChannels, 0.f);`.

PR #2584, opened 2019-03-23
Upstream onnxifi quantization

jackm321:
*Description*:
Create a new onnxifi backend that runs inline. Call quantization from that backend.

*Testing*:
Will test on internal tests

*Documentation*:


lib/Onnxifi/CMakeLists.txt
```diff
@@ -4,28 +4,36 @@ add_library(onnxifi-glow
...
                         HostManager
                         Importer
-                        Optimizer)
+                        Optimizer
+                        Quantization)
 
 target_link_libraries(onnxifi-glow
                       PUBLIC
                         Backends
+                        ExecutionEngine
```
rdzhabarov: why do we need ExecutionEngine here?

jackm321: The `QuantizationGraph` uses it

bertmaher:
Cool to see this getting upstreamed!

lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -0,0 +1,121 @@
...
+                 res->size() * res->getType().getElementSize());
+        }
+
+        outputEvent->signal();
+      });
+
+  return ONNXIFI_STATUS_SUCCESS;
+}
+
+// static
```
nadavrot: Static?

lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -0,0 +1,121 @@
...
+                      std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+                          phNameToOnnxTensorOutputs) {
+  // Run
+  backendPtr_->getBackendId()->runNetwork(
+      this, std::move(ctx),
+      [phNameToOnnxTensorOutputs = std::move(phNameToOnnxTensorOutputs),
+       outputEvent](runtime::RunIdentifierTy runId, llvm::Error err,
+                    std::unique_ptr<ExecutionContext> ctx) {
+        // If an Error occurred then log it in errToBool and signal the output
+        // event
```
nadavrot: Period.

lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "HostManagerOnnxifi.h"
+
+namespace glow {
+namespace onnxifi {
+
+// static
```
nadavrot: Static?

lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  return backendPtr_->getBackendId()->addNetwork(&m_);
+}
+
+onnxStatus
+HostManagerGraph::run(std::unique_ptr<ExecutionContext> ctx,
+                      EventPtr outputEvent,
+                      std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+                          phNameToOnnxTensorOutputs) {
+  // Run
```
nadavrot: Run?

lib/Onnxifi/HostManagerOnnxifi.h
```diff
@@ -0,0 +1,78 @@
...
+#include "Base.h"
+
+#include "glow/Runtime/HostManager/HostManager.h"
+
+namespace glow {
+namespace onnxifi {
+
+class HostManagerBackendId : public BackendId {
+public:
+  /// Create Glow ONNXIFI backend identifier using HostManager with the
```
nadavrot: Period.

lib/Onnxifi/HostManagerOnnxifi.h
```diff
@@ -0,0 +1,78 @@
...
+  initGraph(const void *onnxModel, size_t onnxModelSize, uint32_t weightCount,
+            const onnxTensorDescriptorV1 *weightDescriptors) override;
+
+  onnxStatus run(std::unique_ptr<ExecutionContext> ctx, EventPtr outputEvent,
+                 std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+                     phNameToOnnxTensorOutputs) override;
+
+  const std::string &getName() const { return netName_; }
+
+private:
```
nadavrot: Doxygen comments please.

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "InlineOnnxifi.h"
+
+#include "glow/Quantization/Quantization.h"
+#include "glow/Quantization/Serialization.h"
+#include "glow/Support/Support.h"
+
```
nadavrot: MD5 is really slow. Are you sure you don't need CRC?

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -0,0 +1,105 @@
...
+          onnxModel, onnxModelSize, weightCount, weightDescriptors, *function_,
+          true /*loadInputsAsPlaceholders*/, backendPtr_->getUseOnnx()));
+
+  onnxInputToPlaceholder_ = loader->getInputVarsMapping();
+  onnxOutputToPlaceholder_ = loader->getOutputVarsMapping();
+
+  computeModelHash(onnxModel, onnxModelSize, MD5_);
+  optimize(function_, CompilationMode::Infer);
+
+  // Profile
```
nadavrot: // --  Profile --

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -55,48 +55,74 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
...
+#else
+  constexpr bool withCPU = false;
+#endif
 
-  // In case backendIDs is nullptr or does not have enough capacity just return
-  // the total number of supported backends.
-  if (numBackendsCapacity < *numBackends || !backendIDs) {
-    return ONNXIFI_STATUS_FALLBACK;
-  }
+  // Only return quantization backend if GLOW_DUMP_PROFILE
```
nadavrot: Period.

lib/Onnxifi/InlineOnnxifi.cpp
jackm321: It's running on the model proto only so it I should be fast compared to other initGraph operations. I just timed `md5sum` on a couple of models and the most was 2ms

lib/Onnxifi/GlowOnnxifiManager.cpp
```diff
@@ -27,12 +25,17 @@ GlowOnnxifiManager &GlowOnnxifiManager::get() {
...
 
-  auto *hostManager = getOrCreateHostManager(kind);
-
-  BackendIdPtr backendId = new onnxifi::BackendId(hostManager, kind, useOnnx);
+  BackendIdPtr backendId;
+  if (forQuantization) {
+    backendId = new onnxifi::InlineBackendId(kind, useOnnx);
+  } else {
+    auto *hostManager = getOrCreateHostManager(kind);
+    backendId = new onnxifi::HostManagerBackendId(hostManager, kind, useOnnx);
```
yinghai: Do we need to care about the cost of virtual function?

jackm321: It's probably not a big deal but if it becomes a problem we can fix it

lib/Onnxifi/InlineOnnxifi.cpp
nadavrot: Okay

bertmaher:
Kind of a general comment on this PR, it's a little hard to review because it rolls together what seems like refactoring with functionality changes.  It would be nice to keep those separate (even if they're separate commits within a PR, I know GitHub is terrible since stacked PRs aren't a thing).

jackm321:
@bertmaher good to know. I know that idea but I'm not sure exactly what you mean by that with this PR. There was one onnxifi graph and backend ID before that calls host manager so I created a second one that runs inline and does quantization. What would you see as a helpful stacking of diffs if that were possible here

bertmaher:
> What would you see as a helpful stacking of diffs if that were possible here

So it's possible I'm just not trying hard enough to understand here :-), and GitHub makes it harder than Phabricator, which highlights code movement.  What I think might have helped me as a first diff that makes the onnxifi graph virtual and moves the hostmanager impl to its own file, followed by a second diff that adds the inline-quantization graph.  That way I can kind of mentally write off the first diff as "oh that's some code movement" and focus on the functionality changes in the second.

(And if it's the case that it would have been a giant pain to split things that way, I totally get it, and I'm happy to review a bigger patch.  I am just selfishly trying to make my reviewing life easier :-p ).

jackm321:
Yeah that sounds reasonable and would probably help me if I was reviewing it. Thanks!

rdzhabarov:
Another thing which would help reviewing this is to flesh out details in the description of the PR.
E.g., how do we make sure that C2 sends us the same graph for profile and load (since we use different backends for profiling and loading).

Also, since we have host manager now could we expose a single backend now and hide orchestration of underlying (cpu,interpreter, accelerator) by host manager?

jackm321:
> Another thing which would help reviewing this is to flesh out details in the description of the PR.
E.g., how do we make sure that C2 sends us the same graph for profile and load (since we use different backends for profiling and loading).

Sounds good, I'll add more details to description in the futre

>Also, since we have host manager now could we expose a single backend now and hide orchestration of underlying (cpu,interpreter, accelerator) by host manager?

We're getting close to being able to go from n HostManagers to 1 HostManager. We would need to add a bit more to the interface of HostManager to specify which backends to use and change some things about HostManager to enable heterogenius device managers but shouldn't be too much more work. So that would get us to 1 HostManager that is used by multiple onnxifi::backendids but on top of that, we would need to change stuff about onnxifi inteface to be able to have just 1 onnxifi::backendid that can be called with different parameters, I'm not sure if that's going to be worth doing or what it would look like exactly though

lib/Onnxifi/Base.h
```diff
@@ -53,20 +56,18 @@ class BackendId {
...
+  virtual void runNetwork(const Graph *graph,
+                          std::unique_ptr<ExecutionContext> context,
+                          runtime::ResultCBTy callback) {}
 
-  // \returns a unique_ptr to a new HostManager for the given BackendKind \p
-  // kind.
-  static std::unique_ptr<runtime::HostManager>
-  createHostManager(glow::BackendKind kind);
+  virtual onnxStatus addNetwork(Module *module) {
+    return ONNXIFI_STATUS_SUCCESS;
```
rdzhabarov: when this implementation is useful?

lib/Onnxifi/GlowOnnxifiManager.h
```diff
@@ -42,7 +44,8 @@ class GlowOnnxifiManager final {
   /// Create a new glow BackendId for BackendKind \p kind using onnx graphs if
   /// \p useOnnx and caffe2 graphs otherwise.
   /// Can be called safely by multiple threads concurrently.
-  BackendIdPtr createBackendId(glow::BackendKind kind, bool useOnnx);
+  BackendIdPtr createBackendId(glow::BackendKind kind, bool useOnnx,
```
rdzhabarov: can you update comments accordingly.

here and in the rest of the code.

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -358,7 +384,16 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxInitGraph)(
     return ONNXIFI_STATUS_INVALID_BACKEND;
   }
 
-  auto *glowGraph = manager.createGraph(glowBackend);
+  glow::onnxifi::OnnxifiQuantizationStep quantizationStep;
+  if (getenv("GLOW_DUMP_PROFILE")) {
```
rdzhabarov: that might be an effort not worth benefits, but `auxPropertiesList` could be a candidate to supply action instead of env vars.

jackm321:
Just tried dumping some profiles on internal tests and it works!

tests/unittests/GlowOnnxifiManagerTest.cpp
```diff
@@ -25,7 +25,7 @@ using namespace glow::onnxifi;
 TEST(GlowOnnxifiManagerTest, BackendIdTest) {
   auto &manager = GlowOnnxifiManager::get();
   auto *backendId = manager.createBackendId(glow::BackendKind::Interpreter,
-                                            /*use_onnx*/ true);
+                                            /* use_onnx */ true);
```
rdzhabarov: revert those changes.

lib/Onnxifi/InlineOnnxifi.h
```diff
@@ -0,0 +1,74 @@
...
+  Profile,  // Insert profiling nodes and dump quantization profiles
+  Quantize, // Load quantization profile and insert quantize/dequantize nodes
+};
+
+/// BackendId to be use with InlineGraphs. The implementation is basically empty
+/// because InlineGraph will do all of the heavy lifting for execution by
+/// itself.
+class InlineBackendId : public BackendId {
+public:
+  using BackendId::BackendId;
```
rdzhabarov: what is this using for?

lib/Onnxifi/InlineOnnxifi.h
```diff
@@ -0,0 +1,74 @@
...
+
+  /// Init Glow graph based on the ONNX model \p onnxModel and
+  /// static trained weights \p weightDescriptors.
+  onnxStatus
+  initGraph(const void *onnxModel, size_t onnxModelSize, uint32_t weightCount,
+            const onnxTensorDescriptorV1 *weightDescriptors) override;
+
+  onnxStatus run(std::unique_ptr<ExecutionContext> ctx, EventPtr outputEvent,
+                 std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+                     phNameToOnnxTensorOutputs) override;
```
rdzhabarov: do you intend to copy map here?

lib/Onnxifi/InlineOnnxifi.h
```diff
@@ -0,0 +1,74 @@
...
+  ExecutionEngine executionEngine_;
+  Function *function_;
+  OnnxifiQuantizationStep quantizationStep_;
+
+  /// A map between quantization profiling names of NodeValues that were lowered
+  /// from each other. Maps to a set of NodeValues that were replaced by the
+  /// NodeValue (key) that replaced them.
+  LoweredInfoMap loweredMap_;
+
+  /// Hash of the mode, used to find profiling data.
```
rdzhabarov: model. Can you name var appropriately as well, modelHash_ or whatever.

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -0,0 +1,107 @@
...
+                 std::unordered_map<Placeholder *, onnxTensorDescriptorV1>
+                     phNameToOnnxTensorOutputs) {
+
+  executionEngine_.run(*ctx);
+
+  // Dump profile if requested.
+  if (quantizationStep_ == OnnxifiQuantizationStep::Profile) {
+    auto QI = quantization::generateNodeQuantizationInfos(
+        *(ctx->getPlaceholderBindings()), function_, loweredMap_,
+        quantization::Schema::Symmetric, ElemKind::Int8QTy);
```
rdzhabarov: TODO here to be able to specify quantization schema, not everyone wants symmetric but it's kinda safe bet.

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "llvm/Support/MD5.h"
+
+#include <unistd.h>
+
+namespace glow {
+namespace onnxifi {
+
+namespace {
+std::string getProfileFile(llvm::StringRef hash) {
+  return strFormat("/tmp/glow-profile-%d-%s.yaml", getuid(),
```
rdzhabarov: who is going to clean this after temp file is created?

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -0,0 +1,107 @@
...
+    function_ = profileQuantization(dummyCtx, function_);
+  }
+
+  // -- Quantize --
+  if (quantizationStep_ == OnnxifiQuantizationStep::Quantize) {
+    auto QI = deserializeFromYaml(getProfileFile(MD5_));
+    std::string oldName = function_->getName();
+    function_->setName("old");
+    auto *Q = quantization::quantizeFunction(
+        executionEngine_, quantization::Schema::Asymmetric, QI,
```
rdzhabarov: symmetric here.

jackm321:
@rdzhabarov Another thing I'm wondering about, can we dump profiles in ~InlineGraph so that we can capture data from multiple runs? Does the profile dumping stuff support that?

rdzhabarov:
> Another thing I'm wondering about, can we dump profiles in ~InlineGraph so that we can capture data from multiple runs? Does the profile dumping stuff support that?

If you do not recompile graph you can potentially run as many inferences as you want on instrumented graph and dump at the end. Once you dumped the profile, it's done (scale and offset are calculated).

lib/Onnxifi/Base.h
jackm321: Not useful anymore, removed

lib/Onnxifi/InlineOnnxifi.cpp
jackm321: Nothing cleans this as is currently the case

lib/Onnxifi/InlineOnnxifi.h
jackm321: Removed

lib/Onnxifi/InlineOnnxifi.h
jackm321: This is intentional but it's not copied, `Graph::setIOAndRun` moves this map in here

lib/Onnxifi/onnxifiGlow.cpp
jackm321: Yeah I was thinking about how we can more closely integrate this into c2 side of onnxifi in the future and I think that would be one way to go

PR #2593, opened 2019-03-26
[Tensor] Add randomize() for UInt8FusedQTy

jfix71:
This was missing from when I added `UInt8FusedQTy`.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -813,14 +813,29 @@ public:
   }
 
   /// Fill the tensor with uniformly distributed values in the range
-  /// [low .. high].
+  /// [low .. high]. For quantized fused tensors leave scales/offsets unchanged.
   template <typename T = ElemTy>
   typename std::enable_if<std::is_integral<T>::value>::type
   randomize(int low, int high, PseudoRNG &PRNG) {
     assert(low < high && "invalid range");
     std::uniform_int_distribution<int> dist(low, high);
-    for (auto &elem : *this) {
-      elem = dist(PRNG);
+    switch (getElementType()) {
+    default: {
+      for (auto &elem : *this) {
+        elem = dist(PRNG);
+      }
+      return;
+    }
+    case ElemKind::UInt8FusedQTy: {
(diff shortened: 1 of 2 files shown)


tests/unittests/TensorsTest.cpp
```diff
@@ -898,3 +898,23 @@ TEST(Tensor, initBroadcastFused) {
...
+TEST(Tensor, randomizeFused) {
+  Tensor T(ElemKind::UInt8FusedQTy, {10, 10}, 1.0, 0);
+  auto TH = T.getHandle<uint8_t>();
+  for (size_t i = 0; i < 10; i++) {
+    for (size_t j = 0; j < 10; j++) {
+      TH.at({i, j}) = i * 10 + j;
+    }
+  }
+  PseudoRNG PRNG;
+  TH.randomize(0, 256, PRNG);
```
rdzhabarov: it's [low,high] so should be 0, 255.

include/glow/Base/Tensor.h
```diff
@@ -813,14 +813,30 @@ template <class ElemTy> class Handle final {
...
+      }
+      return;
+    }
+    case ElemKind::UInt8FusedQTy: {
+      assert(dims().size() == 2 && "Fused tensor must be 2-dimensional.");
+      assert(dims()[1] > 8 && "Fused tensor must have more than 8 columns.");
+      for (size_t i = 0, e = dims()[0]; i < e; i++) {
+        for (size_t j = 0, f = dims()[1] - 8; j < f; j++) {
+          auto v = dist(PRNG);
+          memcpy(&at({i, j}), &v, sizeof(uint8_t));
```
nadavrot: Why do we need the memcopy here? Why do we need the indexed access with X,Y instead of just using the raw API to access the flat buffer?

jfix71: @nadavrot Good call. I copied this over from the Xavier initialization. I think it was originally like that as a relic of debugging alignment issues. This led me to realize Xavier initialization doesn't even really make sense for non-float types: https://github.com/pytorch/glow/pull/2595

PR #2594, opened 2019-03-26
[interp] Loosen checking on SparseLengthsSum lengths

bertmaher:
*Description*: Surprisingly, Caffe2 doesn't check this rigorously, which causes some internal integration tests to fail.  Arguably Caffe2 should be stricter, but it doesn't really hurt anything as far as I can tell.
*Testing*: internal test
*Documentation*: n/a


bertmaher:
Looks like the CPU backend does not do this check :-)

PR #2601, opened 2019-03-27
[tensor] Allow specifying number of elements in dumpImpl

bertmaher:
*Description*: I use dumpImpl for debugging all the time, and it's really nice to be able to dump arbitrary amounts of the tensor without hunting down this constant in Tensor.cpp and recompiling.
*Testing*: tried it in model-runner
*Documentation*: n/a

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -507,10 +507,14 @@ private:
 //                    Tensor Handle
 //===----------------------------------------------------------------------===//
 
+constexpr unsigned MAX_DUMP_ELEMS = 100;
+
 void dumpAsciiImpl(const Tensor *T, llvm::raw_ostream &os);
 void dumpAsciiImpl(const Tensor *T);
 
-void dumpImpl(const Tensor *T, llvm::raw_ostream &os);
+void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
+              unsigned maxNumElem = MAX_DUMP_ELEMS);
+void dumpImpl(const Tensor *T, unsigned maxNumElem);
 void dumpImpl(const Tensor *T);
 
 /// A class that provides indexed access to a tensor. This class has value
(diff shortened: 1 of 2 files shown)


opti-mix:
@bertmaher Does it make sense to add the same parameter with a default value to `dump` methods as well?

opti-mix:
@bertmaher Another useful thing to have: printing which elements are different. E.g. printing their offsets (or indices) and their values. Obviously we don't want to print too many of those differences as there could be thousands of them. We should print only the first N differences. I don't have any strong opinion of this N should be configurable or not. If we make it non-configurable, may be a value of 10 could be a good default.

bertmaher:
@opti-mix Love both ideas.  I'll go ahead and modify Handle::dump in this PR, and I'll create an issue for printing diffs so we can get to it at some point.

include/glow/Base/Tensor.h
```diff
@@ -761,8 +765,11 @@ template <class ElemTy> class Handle final {
     return std::all_of(begin(), end(), [=](ElemTy e) { return e == trueZero; });
   }
 
-  void dump(llvm::raw_ostream &os) const { dumpImpl(tensor_, os); }
-  void dump() const { dumpImpl(tensor_); }
+  void dump(llvm::raw_ostream &os, unsigned maxNumElem = MAX_DUMP_ELEMS) const {
+    dumpImpl(tensor_, os, maxNumElem);
+  }
+  void dump(unsigned maxNumElem) const { dumpImpl(tensor_, maxNumElem); }
+  void dump() const { dumpImpl(tensor_, MAX_DUMP_ELEMS); }
```
opti-mix: Why do we need a separate `dump()`? Wouldn't it be easier to have `dump(unsigned maxNumElem = MAX_DUMP_ELEMS)`?

bertmaher: Yeah, so I did this because I'm never sure how visibility of default parameters works in the debugger and I know some people really like to dump tensors from the debugger.

opti-mix: Yeah, in the worst case one would need to call `mytensor.dump(100)` in the debugger. But we can also keep the version without arguments around for the sake of convenience. It is up to you.

PR #2602, opened 2019-03-27
[Tracing] Add DeviceManager tracing and the TraceContext

nickgg:
*Description*: Adds trace events in the three DeviceManager implementations (CPU, Interp, OpenCL) and machinery for collecting trace events inside the runtime. The new events cover enqueueing time vs time in `execute()` and the time to load placeholders onto the device. Not totally sure which events are going to end up being useful to us.
*Testing*: unit tests & the tracing-compare example.
*Documentation*:
In Chrome trace viewer the new events look like this:
image

Next step is adding equivalent TraceEvents to the Executor.

### DIFF
--- a/include/glow/Backends/TraceEvents.h
+++ b/include/glow/Backends/TraceEvents.h
@@ -56,9 +56,14 @@ struct TraceEvent {
 
   static void dumpTraceEvents(std::vector<TraceEvent> &events,
                               llvm::StringRef filename);
+
+  static uint64_t now();
 };
 
-/// Tracing / Profiling events map for a compiled function.
+/// Tracing / Profiling events map for a CompiledFunction.
+/// This class encodes information on how to read event metrics out of Tensors
+/// and into TraceEvents, and is used by
+/// CompiledFunction::translateTraceEvents().
 struct TraceInfo {
   TraceInfo() = default;
   TraceInfo(bool e, size_t d) : enabled(e), dataSize(d) {}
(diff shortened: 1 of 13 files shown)


lib/Backends/Backend.cpp
```diff
@@ -106,7 +106,8 @@ void Backend::autoInstrument(TraceInfo &traceInfo, IRFunction *IR) const {
       // This isn't ideal, the placeholder exists but we have no weight.
       // Probably indicates a bug in the graph, best we can do is create a new
       // placeholder and weight for the instrumentation.
-      assert(!"could not find weight for existing instrumentation placeholder");
+      // assert(!"could not find weight for existing instrumentation
```
nadavrot: Did you forget to uncomment this?

nickgg: Ah, no - leaving this check disabled until @gcatron finishes with the RuntimeBundle symboltable improvements.

nickgg:
My local build and the CI often disagree about success 😒

include/glow/Backends/TraceEvents.h
```diff
@@ -82,6 +87,52 @@ struct TraceInfo {
...
+  STANDARD, // Glow runtime events and backend operator events.
+  DEBUG     // Full debug events with extra information.
+};
+
+/// A context for storing TraceEvents throughout a run (ie. between partitioned
+/// CompiledFunctions).
+class TraceContext {
+  std::vector<TraceEvent> traceEvents_;
+  TraceLevel traceLevel_{TraceLevel::NONE};
+  int traceThread_{0};
```
jackm321: Whats this? Can you add a comment

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -108,8 +108,10 @@ void InterpreterDeviceManager::evictNetworkImpl(std::string functionName,
 void InterpreterDeviceManager::runFunctionImpl(
     RunIdentifierTy id, std::string function,
     std::unique_ptr<ExecutionContext> context, ResultCBTy resultCB) {
+  auto ev = context->scopedEvent("DM_run");
```
jackm321: One issue with the scoped events is they're going to include the time taken to call resultCBs if we don't always remember to always call the `end()` method before every callback which may not be easy to maintain the correctness of. If we don't have the scoped thing, it will be obvious if we forget to call the end() method because there will be no matching end() event but with scoped events it will be a bit inaccurate and we won't have any way to know it.

include/glow/Backends/TraceEvents.h
```diff
@@ -82,6 +87,52 @@ struct TraceInfo {
...
+class ScopedTraceBlock {
+  TraceContext *context_;
+  llvm::StringRef name_;
+  std::map<std::string, std::string> args_;
+  bool end_{false};
+
+public:
+  ScopedTraceBlock(TraceContext *context, llvm::StringRef name);
+  ~ScopedTraceBlock();
+  ScopedTraceBlock &addArg(llvm::StringRef key, llvm::StringRef value);
```
jackm321: Add some comments

include/glow/Backends/TraceEvents.h
```diff
@@ -82,6 +87,52 @@ struct TraceInfo {
...
+  void end();
```
jackm321: comments

include/glow/Backends/TraceEvents.h
nickgg: This is the tid in the output trace, ie the row in the final visualization. In the examples so far we have a tid per DeviceManager but we may want to use a different scheme once we get some experience working with traces.

I'll add a comment to this effect.

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
nickgg: I see what you're getting at but a) events will always be wrong if you don't terminate them at the right time, and b) this is just a helper interface, you can still add events manually (such as we do in the LLVMCompiledFunction in the lookupJitmain event.

Are you against using the Scoped version here, or having it at all?

nickgg:
Fixed up clang-format

include/glow/Backends/TraceEvents.h
```diff
@@ -56,9 +56,14 @@ struct TraceEvent {
 
   static void dumpTraceEvents(std::vector<TraceEvent> &events,
                               llvm::StringRef filename);
+
+  static uint64_t now();
 };
 
-/// Tracing / Profiling events map for a compiled function.
+/// Tracing / Profiling events map for a CompiledFunction.
+///   This class encodes information on how to read event metrics out of Tensors
```
jackm321: is this indentation intended?

nickgg: Probably not, been awhile since I wrote this. I'll remove it

PR #2603, opened 2019-03-27
[Build] Simplify the integration of external backends

tlepley-cadence:
*Description*:

Our backend is developed in a separate git repository. The simplest approach for an external backend integration is to place the entire backend at a single place. This is the purpose of this PR. An external backend placed in the `externalbackends` sub-directory will automatically be integrated to Glow.

In Glow, a backend is composed a 3 main components: core backend code, additional nodes/instructions, tests. The proposed external backend structure has been chosen to have a consistent naming with current Glow directory names:
```
<backend name>
+-- Backends
+-- ClassGen
+-- Tests
```
Note: an (nicer) option would be to name the main sub-directory of external backends as `Core` instead `Backends`. Any feedback ?

Extra information:
- Only the 'Backends' sub-directory is mandatory. `ClassGen` and `Tests` are optional
- The external backend can be deactivated with cmake variable:
  `GLOW_WITH_<name>=OFF`
- The backend can have an optional top level `CMakeLists.txt`
- Backend subdirectories must have their own `CMakeLists.txt`

*Testing*:
Under Linux and Windows with our backend.

@opti-mix follow-up of our discussion at the Glow summit


opti-mix:
@tlepley-cadence This looks awesome!

@rdzhabarov Could you have a look at this PR? WDYT?

opti-mix:
@bertmaher You may be interested in this as well.

bertmaher:
This is cool -- I've done something very similar with other projects that have closed-source components.  (Although I usually have the luxury of hiding the build-system stuff inside the Buck build definitions that are also closed source, so I haven't had to do as much CMake sorcery ;-) ).

rdzhabarov:
At high level, I like this approach, my only wish is to get rid of changes in `tools/ClassGen/InstrGen.cpp` and `NodeGen` so that one can just have a directory with Backend specific code and include glow as a submodule.

I'll review this in details today.

tlepley-cadence:
@rdzhabarov I'll work on some documentation tomorrow.

I thought also about using sub-modules, but I concluded it was not a convenient approach.
IMO, there are 2 main reasons:

- Having Glow as sub-module would mean that Glow is in a sub-directory of the backend, what I think would invert the actual hierarchy (the backend is just a plugin of Glow). Also, it would not work if we want to have multiple 'external' backends at the same time.
- I think that using sub-modules makes sense when 2 project are not developed in a strongly synchronous way and when the interfaces have some level of stability. In the case of a backend, at least in our case today, the backend is developed totally in sync with Glow on a day-to-day basis. Refactoring and new development in Glow impact the interface with the backend very often. In such context, I think that using sub-modules would complexify things more than it would simplify them.

CMakeLists.txt
```diff
@@ -75,6 +76,60 @@ if (GLOW_WITH_OPENCL)
   find_package(OpenCL REQUIRED)
 endif ()
 
+# Returns the name of the backend enabling
+MACRO(getBackendEnableVariable result backend_dir_name)
```
rdzhabarov: we try to keep this CMake lean, can you plase these macros to the cmake/modules/GlowDefaults

CMakeLists.txt
```diff
@@ -75,6 +76,60 @@ if (GLOW_WITH_OPENCL)
...
+  ENDFOREACH()
+  SET(${result} ${dirlist})
+ENDMACRO()
+
+# External backends
+set(EXTERNAL_BACKENDS_DIR ${GLOW_SOURCE_DIR}/externalbackends)
+include_directories(${EXTERNAL_BACKENDS_DIR})
+getSubDirList(SUBDIRS ${EXTERNAL_BACKENDS_DIR})
+
+FOREACH(child ${SUBDIRS})
```
rdzhabarov: can you place that to a separate file GlowExternalBackends.cmake in cmake/modules and include it here.

lib/Backends/CMakeLists.txt
```diff
@@ -16,6 +16,23 @@ foreach(object ${subdirs})
   endif()
 endforeach()
 
+# External backends
```
rdzhabarov: we need to add some docs around having external backends to add themselves to
linked_factories and linked_device_managers.

rdzhabarov:
> Having Glow as sub-module would mean that Glow is in a sub-directory of the backend, what I think would invert the actual hierarchy (the backend is just a plugin of Glow). Also, it would not work if we want to have multiple 'external' backends at the same time.

I meant the following:
my_glow_repo:
..  glow (submodule here)
..  externalbackends
This would allow pretty seamless updates.
Why this would not work with multiple backends?

> I think that using sub-modules makes sense when 2 project are not developed in a strongly synchronous way and when the interfaces have some level of stability. In the case of a backend, at least in our case today, the backend is developed totally in sync with Glow on a day-to-day basis. Refactoring and new development in Glow impact the interface with the backend very often. In such context, I think that using sub-modules would complexify things more than it would simplify them.

How do you keep that in sync?
(With submodule it's just, git pull + git checkout `sha` you need)

tools/ClassGen/NodeGen.cpp
```diff
@@ -708,6 +708,9 @@ int main(int argc, char **argv) {
 
 #include "Backends/CPU/CPUSpecificNodes.h"
 #include "Backends/OpenCL/OpenCLSpecificNodes.h"
+  // Add here external backend specific node headers.
+  // Example:
+  // #include "<Name>/ClassGen/<Name>SpecificNodes.h"
```
jfix71: Seems like even if a backend is disabled we will still include their backend-specific Nodes/Instrs, right? I suppose that's already true, but might be nice to disable that somehow.

CMakeLists.txt
tlepley-cadence: Done. I've moved the external backend related cmake code in cmake/modules

CMakeLists.txt
tlepley-cadence: Done. I've moved the external backend related cmake code in cmake/modules

lib/Backends/CMakeLists.txt
tlepley-cadence: I've added some docs about external backend. The purpose is not to detail all aspects that are common with standard backend, but more to focus on external backend specific aspects. Nevertheless, I've also summarized some aspects in common with standard backends.

tlepley-cadence:
@jfix71 @opti-mix @rdzhabarov @bertmaher I've reworked the PR to adress feedbacks and add documentation.

tools/ClassGen/InstrGen.cpp
```diff
@@ -624,6 +624,11 @@ int main(int argc, char **argv) {
 
 #include "Backends/CPU/CPUSpecificInstrs.h"
 #include "Backends/OpenCL/OpenCLSpecificInstrs.h"
+  // Add here external backend specific instructions headers.
```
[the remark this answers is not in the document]
tlepley-cadence: @opti-mix I like the idea, I think it would add complexity to this particular PR that needs to be handled carefully. I would then prefer to tackle such improvement in a separate PR after this one has landed.

The things that need to be taken care about with your proposal are the following:
- Today, the proposed external backend mechanism just knows about the fact that there is a ClassGen directory. It doesn't know whether there are backend specific nodes, specific instruction or both. Also, it doesn't know the name and location of the header files. Generating `ExternalBackendsSpecificInstrs.h` would mean that backends need to inform about their .h file and if they add specific nodes and/or specific instruction.
- We need to make sure that the dependencies are well managed. For instance, we don't want to recompile the glow library if there is no change in backend. It means that this `ExternalBackendsSpecificInstrs.h` should not be generated all the time and that the CMake build infrastructure should detect if there is any change in external backends before to touch this generated file.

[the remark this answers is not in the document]
opti-mix: @tlepley-cadence Sound good! It is fine to tackle this improvement in a separate PR.

opti-mix:
@tlepley-cadence Could you rebase your PR and resolve the conflicts? And then it should be ready to land!

docs/Backends.md
```diff
@@ -194,3 +194,14 @@ optimizations to take place.
 The `tools/ClassGen/Backends/CPU/CPUSpecificNodes.h` and
 `tools/ClassGen/Backends/CPU/CPUSpecificInstrs.h` files are included in
 `tools/ClassGen/NodeGen.cpp` and `tools/ClassGen/InstrGen.cpp`, respectively.
+
+
+#### External backends
+
+External backends can be added to Glow without to change the Glow build infrastructure.
```
rdzhabarov: nit: .. without changing ..

docs/ExternalBackend.md
```diff
@@ -0,0 +1,128 @@
+## External backends
+
+An external backend can be added to Glow without to change the Glow build infrastructure. An external backend must be provided as a single source directory, and then can be developped in a separate source management repository.
```
rdzhabarov: ditto

docs/ExternalBackend.md
```diff
@@ -0,0 +1,128 @@
...
+The core of the external backend is mandatory and located in `<Name>/Backends/`. It must contain a `CMakeLists.txt` file that will ensure the correct compilation of the backend and the good integration with Glow.
+
+The external backend must respect the same interface as standard backends.
+It can be summarized the following way:
+
+- Backend main class that inherits from the Glow `Backend` class. For example, class `<Name>Backend` in library `${EXTERNAL_BACKEND_NAME}`.
+- Backend factory registration in a backend specific factory library. For example, calling `REGISTER_GLOW_BACKEND_FACTORY(<Name>Factory, <Name>Backend, <Name>)` in library `${EXTERNAL_BACKEND_NAME}Factory`.
+
+
+The external backend is declared to GLow by setting the CMake `linked_factories` variable in the `<Name>/Backends/CMakeLists.txt`. Example:`set(linked_factories ${linked_factories} <Name>Factory PARENT_SCOPE)`
```
rdzhabarov: nit: Glow

docs/ExternalBackend.md
tlepley-cadence: done

docs/ExternalBackend.md
tlepley-cadence: Done

docs/Backends.md
tlepley-cadence: Done

tlepley-cadence:
@opti-mix Rebased.

opti-mix:
@tlepley-cadence Thanks again for adding this feature! I'm sure many users will find it useful!

Issue #2587, opened 2019-03-25
Shared libs build on mac is broken


rdzhabarov:
@nickgg mentioned that building shared libs are broken on macos (see also that linux CI build is healthy).
InterpreterFactory was introduced in this [PR](https://github.com/pytorch/glow/commit/c8d3956e301d74ab7c208f12ddd3d917c8d3f396)

Repro steps:

```
cmake -GNinja -DCMAKE_CXX_FLAGS=-Werror -DGLOW_WITH_CPU=ON -DCMAKE_BUILD_TYPE=Debug -DGLOW_WITH_OPENCL=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_PREFIX_PATH=/usr/local/opt/llvm ../glow
```

```
ninja InterpreterFactory
```

```
[125/241] Linking CXX shared library lib/Backends/Interpreter/libInterpreterFactory.dylib
FAILED: lib/Backends/Interpreter/libInterpreterFactory.dylib 
: && /usr/bin/c++ -Werror -Wall -Wnon-virtual-dtor -fno-exceptions -fno-rtti -g -fno-omit-frame-pointer -O0  -dynamiclib -Wl,-headerpad_max_install_names  -o lib/Backends/Interpreter/libInterpreterFactory.dylib -install_name @rpath/libInterpreterFactory.dylib lib/Backends/Interpreter/CMakeFiles/InterpreterFactory.dir/InterpreterFactory.cpp.o  -Wl,-rpath,/Users/rdzhabarov/src/release_build/lib/Backends/Interpreter lib/Backends/Interpreter/libInterpreter.dylib && :
Undefined symbols for architecture x86_64:
  "llvm::DisableABIBreakingChecks", referenced from:
      llvm::VerifyDisableABIBreakingChecks in InterpreterFactory.cpp.o
```

tlepley-cadence:
@ayermolo could you please take a look ? We had some issue under Windows related to `DisableABIBreakingChecks

ayermolo:
@tlepley-cadence different issues looks like, but hey for once it's not windows that is having some issue not reproducible on other platforms. :)

PR #2605, opened 2019-03-27
Fix local shared build

nickgg:
*Description*: @rdzhabarov's Backend Registry addition needs extra dependencies in local SHARED build - at least on my machine. The symbol that is missing for the factories is  llvm::DisableABIBreakingChecks but not sure exactly where that gets called.
*Testing*: build & test shared build locally.
*Documentation*:

### DIFF
--- a/lib/Backends/Interpreter/CMakeLists.txt
+++ b/lib/Backends/Interpreter/CMakeLists.txt
@@ -29,7 +29,8 @@ add_library(InterpreterFactory
               InterpreterFactory.cpp)
 target_link_libraries(InterpreterFactory
                       PRIVATE
-                        Interpreter)
+                        Interpreter
+                        LLVMCore)
 
 set(linked_factories ${linked_factories} InterpreterFactory PARENT_SCOPE)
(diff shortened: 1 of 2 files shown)


rdzhabarov:
https://github.com/pytorch/glow/issues/2587 needs to be closed after that.

opti-mix:
This is very strange. Could it be because we use some types from LLVM almost everywhere, e.g. `llvm::ArrayRef` and the like?

nickgg:
I would expect that to affect all libs though, not just the *Factory ones?

opti-mix:
`llvm/lib/Support/Error.cpp` in LLVM seems to refer to those symbols. May be we use some of LLVM error types or may be llvm assertions like LLVM_UNREACHABLE in the factory classes?

rdzhabarov:
If you look at InterpreterFactory it's plain simple
```
#include "Interpreter.h"

namespace glow {

REGISTER_GLOW_BACKEND_FACTORY(InterpreterFactory, Interpreter, Interpreter);

} // namespace glow
```

I'd assume it would not need anything on top of what Interpreter needs. But Interpreter does not need LLVMCore.

PR #2606, opened 2019-03-27
Require FC node to have 2D input

jfix71:
We were already flattening FC inputs when lowering the FC on open source backends, and when loading FCs in the C2 loader. I basically moved the flattening logic from `lowerFullyConnectedNode()`/the C2 loader for FCs into `createFullyConnected()`, along with allowing for specifying the axis. And then added verification for FC that is has 2D input.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -568,11 +568,6 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       ASSIGN_VALUE_OR_RETURN_ERR(axis, loadInt(dict["axis"]));
     }
 
-    // If number of input dims is greater then 2 flatten on axis.
-    if (in.getType()->dims().size() > 2) {
-      in = G_.createFlatten("fc.in", in, axis);
-    }
-
     // Load weights.
     Tensor *w;
     ASSIGN_VALUE_OR_RETURN_ERR(w, getTensorByName(op.input(1)));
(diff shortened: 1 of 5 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -623,9 +618,9 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
       auto outTy = G_.getParent()->uniqueType(
           ElemKind::Int8QTy, {in.getType()->dims()[0], B->getType()->dims()[0]},
           yScale, yZeroPoint - OFFSETSHIFT);
-      node = G_.createFullyConnected(opName, in, W, B, outTy);
+      node = G_.createFullyConnected(opName, in, W, B, outTy, axis);
     } else {
-      node = G_.createFullyConnected(opName, in, W, B);
+      node = G_.createFullyConnected(opName, in, W, B, axis);
```
jackm321: Should we move the stuff below starting at line 626 as well?

jfix71: Good question -- the reason I didn't was that I was thinking of the reshaping back after the FC as more specific to Caffe2's FC semantics. Meanwhile I think of requiring FC inputs to be 2D as independent of Caffe2/ONNX/etc., so that's why it made sense to move the reshape to the creator and add an IR constraint in the verification. What do you think?

jackm321: Hm yeah that kinda makes sense to me too

jfix71:
@beicy I think we should actually keep the `FCWithFlatten` test in `Caffe2ImporterTest` -- that test also checks that the output shape is correct based on Caffe2's expectations. This PR does not change the output shape -- the `FC` case in `Caffe2ModelImporter` still adds a reshape after the `FC` which is tested by `FCWithFlatten` test in `Caffe2ImporterTest`.

Also we already have an `FCWithFlatten` in `OperatorTest` which is testing that an FC with 3D input works correctly -- do you think we need another test here?

beicy:
> @beicy I think we should actually keep the `FCWithFlatten` test in `Caffe2ImporterTest` -- that test also checks that the output shape is correct based on Caffe2's expectations. This PR does not change the output shape -- the `FC` case in `Caffe2ModelImporter` still adds a reshape after the `FC` which is tested by `FCWithFlatten` test in `Caffe2ImporterTest`.
> 
> Also we already have an `FCWithFlatten` in `OperatorTest` which is testing that an FC with 3D input works correctly -- do you think we need another test here?

In that case, I think no more test is needed:)

PR #2609, opened 2019-03-27
[WIP][Benchmarks] Add ThreadPoolExecutor microbenchmark

SplitInfinity:
**Description**
   
This commit adds a microbenchmark for `ThreadPoolExecutor` that
 benchmarks the executor's performance on the following inputs:
   
- empty DAG
- single node DAG
- straight line DAG
- DAG with many nodes that can run in parallel
- large DAG

**Testing**
The benchmark compiles and runs. All unit tests pass.

Release mode benchmark run:
```
$ ./tests/ExecutorBench 
2019-03-28 12:32:35
Running ./tests/ExecutorBench
Run on (8 X 3100 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 8388K (x1)
Load Average: 2.11, 2.64, 3.65
---------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations
---------------------------------------------------------------------------------------
ThreadPoolExecutorBench/Empty                      5.12 us         5.10 us       125136
ThreadPoolExecutorBench/SingleNode                 32.4 us         21.6 us        32709
ThreadPoolExecutorBench/StraightLine/1             33.3 us         22.3 us        32010
ThreadPoolExecutorBench/StraightLine/2             48.6 us         24.6 us        28518
ThreadPoolExecutorBench/StraightLine/4             75.5 us         27.2 us        24108
ThreadPoolExecutorBench/StraightLine/8              140 us         34.9 us        19600
ThreadPoolExecutorBench/StraightLine/16             246 us         45.6 us        15437
ThreadPoolExecutorBench/StraightLine/32             477 us         66.9 us        10142
ThreadPoolExecutorBench/StraightLine/64             921 us          114 us         6204
ThreadPoolExecutorBench/ManyParallelNodes/1        32.9 us         22.1 us        31056
ThreadPoolExecutorBench/ManyParallelNodes/2        48.1 us         30.7 us        23145
ThreadPoolExecutorBench/ManyParallelNodes/4        85.3 us         62.9 us        11430
ThreadPoolExecutorBench/ManyParallelNodes/8         141 us          123 us         5600
ThreadPoolExecutorBench/ManyParallelNodes/16        262 us          235 us         3013
ThreadPoolExecutorBench/ManyParallelNodes/32        480 us          425 us         1619
ThreadPoolExecutorBench/ManyParallelNodes/64        800 us          550 us         1356
ThreadPoolExecutorBench/LargeDAG                    176 us         52.1 us        13312
```


SplitInfinity:
**`cmake`**
I am aware my `cmake` setup isn't that great. I have no strong opinions about where everything should live and am happy to take and act on suggestions.

**Naming Things**
Naming things is hard. `ExecutorTestBuilder`, `ExecutorUnitTestWrapper` and `ExecutorBenchmarkWrapper` are not the best of names. My main goal here was to reuse all the helper code I wrote for `ExecutorTest` that creates `RuntimeBundles` correctly, checks that intermediate `ExecutionContext` objects passed to `TestDeviceManager` are correct, etc. The benchmarks are very similar to the unit tests  except that they take in a `benchmark::State` object so that they can pause/resume timing while setting up for the actual benchmark. The first two commits in this stack deal with moving that code to a location accessible by both the unit tests and benchmarks and refactoring it so that it can be used to create both tests that check for correctness and benchmarks.

**Timing Control**
`benchmark::State::PauseTiming/ResumeTiming` have high overhead but I used them there for simplicity. I wonder if I should move all of the setup out of the `for (auto _ : state) {...}` loop so that I don't need to control the timer.

**`TestDeviceManager` Overhead**
The results shown in the PR summary include `TestDeviceManager` overhead of making sure the intermediate contexts are all correct. Do you think it is a good idea to use the `DeviceBindings` member of the `ExecutionContext` to pass through the `benchmark::State` so that `TestDeviceManager` can stop the clock while it does all those correctness checks?

bertmaher:
> The results shown in the PR summary include TestDeviceManager overhead of making sure the intermediate contexts are all correct. Do you think it is a good idea to use the DeviceBindings member of the ExecutionContext to pass through the benchmark::State so that TestDeviceManager can stop the clock while it does all those correctness checks?

So unfortunately I don't know this code well (at all?) but philosophically I feel like starting and stopping the timer in a bunch of places is kind of sketchy from a benchmarking perspective.  It'd be better to have a realistic DeviceManager component that doesn't do all the test-only checks (or have those checks compiled out in OPT builds).

SplitInfinity:
That's fair. I wanted to eliminate the correctness checks altogether, but @nickgg brought up the point that it's probably a good idea to leave them in to make sure the benchmark is measuring the right thing. There's no point measuring something that's incorrect.

But on the other hand, perhaps that's what unit tests are for.

tests/benchmark/ExecutorBench.cpp
```diff
@@ -0,0 +1,395 @@
...
+  ExecutorTestBuilder benchmarkBuilder_;
+  /// DeviceManager map for initializing executor_.
+  DeviceManagerMapTy deviceManagerMap_;
+};
+
+/// Benchmark for executing an empty DAG.
+BENCHMARK_DEFINE_F(ThreadPoolExecutorBench, Empty)(benchmark::State &state) {
+  constexpr RunIdentifierTy testRunId = 10;
+
+  for (auto _ : state) {
```
nickgg: I'm not familiar enough with gbench I guess, what are we iterating over here?

tests/benchmark/ExecutorBench.cpp
```diff
@@ -0,0 +1,395 @@
...
+    std::unique_ptr<ExecutionContext> executorOutputContext;
+
+    // Call Executor::run().
+    llvm::Error runErr = llvm::Error::success();
+    std::promise<void> promise;
+    std::future<void> future = promise.get_future();
+
+    // Start timing.
+    state.ResumeTiming();
+    executor_->run(nullptr, std::move(testContext), testRunId,
```
nickgg: Can you use the benchmark wrapper run here to avoid duplicating code?

tests/utils/ExecutorTestBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // Otherwise, make sure that the runId matches that of the previous nodes.
+  if (nodes_.empty()) {
+    runId_ = runId;
+  } else {
+    assert(runId == runId_ && "Node run ID does not match rest of graph!");
+  }
+
+  // If the result for this node is false, set the expected
+  // result for the entire test to false.
+  success_ &= success;
```
nickgg: What is the purpose of this bool? Is it to allow tests where running the sub graph fails?

tests/utils/ExecutorTestBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+      (newNode->parents).emplace_back(parentPtr);
+      (parentPtr->children).emplace_back(newNodeRawPtr);
+      leaves_.erase(parentPtr);
+    }
+  } else {
+    (newNode->parents).emplace_back(root_.get());
+    (root_->children).emplace_back(newNode.get());
+  }
+
+  // Iterate through inputs and outputs and:
```
nickgg: Is this normally done by the partitioner/provisioner? Can we reuse that code?

tests/benchmark/ExecutorBench.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+    // Check that the returned ID and context are as expected.
+    GLOW_ASSERT(!errToBool(std::move(runErr)));
+    GLOW_ASSERT(executorRunId == testRunId);
+    GLOW_ASSERT(PlaceholderBindings::compare(
+        refContext->getPlaceholderBindings(),
+        executorOutputContext->getPlaceholderBindings()));
+
+    // Resuming timing after checking correctness to preserve loop invariant.
+    state.ResumeTiming();
```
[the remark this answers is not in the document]
SplitInfinity: I tried this actually but no. You can't pause timing outside the body of `for (auto _ : state) { ... }`.

tests/utils/ExecutorTestBuilder.cpp
SplitInfinity: Yep.

tests/utils/ExecutorTestBuilder.cpp
SplitInfinity: @gcatron Can we?

tests/benchmark/ExecutorBench.cpp
SplitInfinity: The different iterations of the benchmark. If you pass a range of arguments, you can access the specific value the benchmark should run with through `state`.

tests/benchmark/ExecutorBench.cpp
SplitInfinity: Hmm, I'll try.

tests/utils/ExecutorTestBuilder.cpp
gcatron: I think we're short circuiting that logic here. The partitioner/provisioner reason over a network graph. This is creating the DAG directly.

PR #2612, opened 2019-03-28
[runtime] Force RuntimeBundle to be properly constructed

bertmaher:
*Description*: I was bitten hard today because the backend I was working on didn't build a RuntimeBundle (because it doesn't need one), and I discovered only through lots of debugging that the Provisioner requires the symbol table from a runtime bundle.  To help prevent others from making this mistake, let's make the illegal state (of a dummy RuntimeBundle hanging out in CompiledFunction) unrepresentable by deleting its default constructor.

I was going to force everything to use the static factory methods, but that got really messy in ExecutorTest so I gave up, and will allow people to shoot themselves in the foot as long as the type RuntimeBundle(SymbolTable(), 0, 0, 0) while doing so.

*Testing*: existing unit tests

*Documentation*: n/a

Stacked on #2611

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -81,7 +81,24 @@ public:
   /// Sets the input and output flags for each symbol in the symbolBundle.
   void setInputsandOutputs();
 
-  RuntimeBundle() = default;
+  /// Computes offsets and total allocation for Constants, Placeholders, and
+  /// Activations to build runtime symbol table. Returns RuntimeBundle.
+  static runtime::RuntimeBundle create(const IRFunction &F,
+                                       MemoryAllocator &constantAllocator,
+                                       MemoryAllocator &placeholderAllocator,
+                                       MemoryAllocator &activationsAllocator);
+
+  /// Build a runtime symbol table from a Function.  Computes Constant and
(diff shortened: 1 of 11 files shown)


include/glow/Backends/BackendUtils.h
```diff
@@ -81,7 +81,24 @@ class RuntimeBundle {
...
+  static runtime::RuntimeBundle create(const IRFunction &F,
+                                       MemoryAllocator &constantAllocator,
+                                       MemoryAllocator &placeholderAllocator,
+                                       MemoryAllocator &activationsAllocator);
+
+  /// Build a runtime symbol table from a Function.  Computes Constant and
+  /// Placeholder sizes, but not Activations, since Functions are unserialized.
+  /// Only use this method to generate bundles for backends that do not use
+  /// Glow's IR.
+  static runtime::RuntimeBundle create(const Function *F);
```
rdzhabarov: nit: why not ref here? (consistent with the other `create` method, does not allow to pass nullptr)

bertmaher: Sure, I can do that.  We use `Function *` way more often than `Function &` but I can't see any good reason for it...

PR #2614, opened 2019-03-28
[NFC] Move loadImagesAndPreprocess() to Image.h

artemrakhov:
*Description*:
Move this function so it could be reused in other files.

*Testing*:
`ninja all`

### DIFF
--- a/include/glow/Base/Image.h
+++ b/include/glow/Base/Image.h
@@ -22,6 +22,9 @@
 
 #include "glow/Base/Tensor.h"
 
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
+
 #include <tuple>
 
 namespace glow {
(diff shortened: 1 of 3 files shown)


include/glow/Base/Image.h
```diff
@@ -91,6 +93,21 @@ Tensor readPngImageAndPreprocess(llvm::StringRef filename,
...
+
+/// Loads and normalizes all PNGs into a tensor in the NHWC format with the
+/// requested channel ordering.
+/// \param filenames list of filenames to read.
+/// \param inputImageData Tensor to save the resulting output.
+/// \param imageNormMode normalize values to this range.
+/// \param imageChannelOrder the order of color channels.
+/// \param imageLayout the order of dimensions (channel, height, and width).
+/// \param useImagenetNormalization use special normalization for Imagenet.
+void loadImagesAndPreprocess(const llvm::cl::list<std::string> &filenames,
```
jfix71: You should be able to use `llvm::ArrayRef<std::string> filenames` instead here, and then you won't have to include `CommandLine.h` in `Image.h`.

PR #2616, opened 2019-03-28
Modify compilation to only use function symbols

### LINKED ISSUE #2548: Do not allocate memory for weights and activations that are not used by an IRFunction
`glow::generateRuntimeBundle` currently allocates memory even for weights and activations that are not used by the IRFunction being compiled. This is a waste of memory. We should skip such allocations.

gcatron:
*Description*: This PR modifies compilation to no longer check all placeholders/constants in the module. It now uses symbols that are associated with the function being compiled. 
*Testing*: ninja test and check_expensive pass
*Documentation*:  NA
Fixes #2548

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -200,6 +200,9 @@ class Function final : public Named {
   /// A list of nodes that the Function owns.
   NodesList nodes_;
 
+  /// A list of metadata PHs associated with the function.
+  std::vector<Placeholder *> metadataPlaceholders_;
+
   /// Stores a list of unique node names that were used by the module at some
   /// point.
   llvm::StringSet<> uniqueNodeNames_{};
@@ -213,7 +216,28 @@ public:
 
   ~Function();
 
+  /// Add placeholder for metadata such as profiling.
+  void addMetadataPlaceholder(Placeholder *PH) {
+    metadataPlaceholders_.push_back(PH);
+  }
+
+  /// Get list of metadata placeholders.
+  const std::vector<Placeholder *> &getMetadataPlaceholders() const {
+    return metadataPlaceholders_;
+  }
+
   Module *getParent() { return parent_; }
+
+  /// Search the Module containing the function to gather and return a list of
+  /// placeholders that are used by the Function.
+  PlaceholderList findPlaceholders();
+  PlaceholderList findPlaceholders() const;
+
+  /// Search the Module containing the function to gather and return a list of
+  /// constants that are used by the Function.
+  ConstList findConstants();
+  ConstList findConstants() const;
+
   const Module *getParent() const { return parent_; }
 
   /// Inserts the node \p N to the list of nodes, and returns the inserted node.
--- a/include/glow/IR/IR.h
+++ b/include/glow/IR/IR.h
@@ -320,6 +320,12 @@ public:
   /// \returns the variable map.
   VariableMap &getVariableMap() { return variableMap_; }
 
+  /// Returns a list of constants associated with function.
+  std::vector<const Constant *> findConstants() const;
+
+  /// Returns a list of placeholders associated with the function.
+  std::vector<const Placeholder *> findPlaceholders() const;
+
   /// \returns the weight that the variable \p v is lowered into, or null if the
   /// variable is unknown.
   Value *getWeightForNode(const Storage *V) const;
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -2712,6 +2712,58 @@ void Function::eraseNode(Node *N) {
   eraseNode(N->getIterator());
 }
 
+PlaceholderList Function::findPlaceholders() {
+  PlaceholderList list;
+  for (auto &PH : parent_->getPlaceholders()) {
+    for (auto &user : PH->getUsers()) {
+      if (user.getUser()->getParent() == this) {
+        list.push_back(PH);
+        break;
+      }
+    }
+  }
+  return list;
+}
+
+PlaceholderList Function::findPlaceholders() const {
+  PlaceholderList list;
+  for (auto &PH : parent_->getPlaceholders()) {
+    for (auto &user : PH->getUsers()) {
+      if (user.getUser()->getParent() == this) {
+        list.push_back(PH);
+        break;
+      }
+    }
+  }
+  return list;
+}
+
+ConstList Function::findConstants() {
+  ConstList list;
+  for (auto &constant : parent_->getConstants()) {
+    for (auto &user : constant->getUsers()) {
+      if (user.getUser()->getParent() == this) {
+        list.push_back(constant);
+        break;
+      }
+    }
+  }
+  return list;
+}
+
+ConstList Function::findConstants() const {
+  ConstList list;
+  for (auto &constant : parent_->getConstants()) {
+    for (auto &user : constant->getUsers()) {
+      if (user.getUser()->getParent() == this) {
+        list.push_back(constant);
+        break;
+      }
+    }
+  }
+  return list;
+}
+
 Function *Function::clone(llvm::StringRef newName,
                           llvm::DenseMap<Node *, Node *> *map) {
   Module *M = getParent();
--- a/lib/IR/GraphScheduler.cpp
+++ b/lib/IR/GraphScheduler.cpp
@@ -48,14 +48,20 @@ Scheduler *createScheduler(SchedulerKind schedulerKind, Function &G,
 
 void IRFunction::scheduleGraph(NodesPtrList &Schedule) {
   Schedule.clear();
-  for (auto &N : G_->getParent()->getConstants()) {
+  auto constants = G_->findConstants();
+  auto placeholders = G_->findPlaceholders();
+  for (auto &N : constants) {
     Schedule.push_back(N);
   }
-  for (auto &N : G_->getParent()->getPlaceholders()) {
+  for (auto &N : placeholders) {
     Schedule.push_back(N);
   }
-  auto numVars = G_->getParent()->getConstants().size();
-  auto numPlaceholders = G_->getParent()->getPlaceholders().size();
+  for (auto &N : G_->getMetadataPlaceholders()) {
+    Schedule.push_back(N);
+  }
+  auto numVars = constants.size();
+  auto numPlaceholders =
+      placeholders.size() + G_->getMetadataPlaceholders().size();
   (void)numVars;
   (void)numPlaceholders;
   std::unique_ptr<Scheduler> scheduler{
--- a/lib/IR/IR.cpp
+++ b/lib/IR/IR.cpp
@@ -428,6 +428,26 @@ static void dumpIRInContext(const Value *V, llvm::raw_ostream &out) {
   }
 }
 
+std::vector<const Constant *> IRFunction::findConstants() const {
+  std::vector<const Constant *> constants;
+  for (auto &node : variableMap_) {
+    if (const Constant *constant = dyn_cast<const Constant>(node.first)) {
+      constants.push_back(constant);
+    }
+  }
+  return constants;
+}
+
+std::vector<const Placeholder *> IRFunction::findPlaceholders() const {
+  std::vector<const Placeholder *> placeholders;
+  for (auto &node : variableMap_) {
+    if (const Placeholder *PH = dyn_cast<const Placeholder>(node.first)) {
+      placeholders.push_back(PH);
+    }
+  }
+  return placeholders;
+}
+
 /// Dump the instruction numbers of all users of \p V.
 static void dumpUsers(const Value *V, llvm::raw_ostream &out,
                       InstructionNumbering &InstrNumbering) {
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -53,7 +53,7 @@ void BundleSaver::saveWeights(llvm::StringRef weightsFileName) {
   // it should be configurable and set by the client.
   size_t pos = 0;
   size_t maxPos = 0;
-  for (auto &v : F_->getGraph()->getParent()->getConstants()) {
+  for (auto &v : F_->findConstants()) {
     auto *w = cast<WeightVar>(F_->getWeightForNode(v));
     auto numBytes = w->getSizeInBytes();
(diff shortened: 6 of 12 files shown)


nadavrot:
@bertmaher FYI

include/glow/Graph/Graph.h
```diff
@@ -212,7 +215,24 @@ class Function final : public Named {
 
   ~Function();
 
+  /// Add placeholder for metadata such as profiling.
+  void addMetaDataPlaceholder(Placeholder *PH) {
+    metaDataPlaceholders_.push_back(PH);
+  }
+
+  /// Get list of metaData placeholders.
+  const std::list<Placeholder *> &getMetadata() const {
```
jfix71: nit: For consistency can you capitalize "Data" here -- and perhaps also append Placeholders (e.g. `getMetaDataPlaceholders()`)?

lib/IR/GraphScheduler.cpp
```diff
@@ -48,14 +48,18 @@ Scheduler *createScheduler(SchedulerKind schedulerKind, Function &G,
 
 void IRFunction::scheduleGraph(NodesPtrList &Schedule) {
   Schedule.clear();
-  for (auto &N : G_->getParent()->getConstants()) {
+  for (auto &N : G_->getConstants()) {
```
jfix71: Since there can be duplicates in the returned list, aren't you adding duplicate Constants here if multiple Nodes refer to them? Maybe `getConstants()` should return an `unordered_set`?

tests/unittests/BackendTest.cpp
```diff
@@ -188,6 +190,40 @@ TEST_P(BackendTest, CompileWithoutConstants) {
...
+  CompilationOptions opts;
+  auto function = backend->compile(F, opts);
+  auto function2 = backend->compile(F2);
+  auto table1 = function->getRuntimeBundle().getSymbolTable();
+  auto table2 = function2->getRuntimeBundle().getSymbolTable();
+  /// Make sure no symbol in table1 is in table2.
+  for (auto sym : table1) {
+    auto it = table2.find(sym.first);
+    EXPECT_TRUE(it == table2.end());
+  }
```
jfix71: Can you also check that the number of symbols is correct/as expected? Maybe add multiple users to some of the Placeholders/Constants above too to make sure they're not accidentally somehow duplicated in the symbol table?

lib/IR/GraphScheduler.cpp
```diff
@@ -48,14 +48,18 @@ Scheduler *createScheduler(SchedulerKind schedulerKind, Function &G,
...
-  for (auto &N : G_->getParent()->getPlaceholders()) {
+  for (auto &N : G_->getPlaceholders()) {
     Schedule.push_back(N);
   }
-  auto numVars = G_->getParent()->getConstants().size();
-  auto numPlaceholders = G_->getParent()->getPlaceholders().size();
+  for (auto &N : G_->getMetadata()) {
+    Schedule.push_back(N);
+  }
+  auto numVars = G_->getConstants().size();
```
jfix71: You use `getConstants()` in this function twice, but it's no longer a simple getter -- probably should get it once at the top of the Function and reuse it to check its size. Same with `getPlaceholders()`.

lib/Graph/Graph.cpp
```diff
@@ -2689,6 +2689,34 @@ void Function::eraseNode(Node *N) {
   eraseNode(N->getIterator());
 }
 
+PlaceholderList Function::getPlaceholders() {
+  PlaceholderList list;
+  for (auto &node : nodes_) {
+    for (unsigned i = 0; i < node.getNumInputs(); i++) {
+      auto input = node.getNthInput(i).getNode();
+      if (input->getKind() == Kinded::Kind::PlaceholderKind) {
+        auto PH = static_cast<Placeholder *>(input);
```
jfix71: nit: `auto *PH`

lib/Graph/Graph.cpp
```diff
@@ -2689,6 +2689,34 @@ void Function::eraseNode(Node *N) {
...
+  return list;
+}
+
+ConstList Function::getConstants() {
+  ConstList list;
+  for (auto &node : nodes_) {
+    for (unsigned i = 0; i < node.getNumInputs(); i++) {
+      auto input = node.getNthInput(i).getNode();
+      if (input->getKind() == Kinded::Kind::ConstantKind) {
+        auto PH = static_cast<Constant *>(input);
```
jfix71: nit: `auto *C`

lib/IR/IR.cpp
```diff
@@ -428,6 +428,28 @@ static void dumpIRInContext(const Value *V, llvm::raw_ostream &out) {
   }
 }
 
+std::list<const Constant *> IRFunction::getConstants() const {
+  std::list<const Constant *> constants;
+  for (auto &node : variableMap_) {
+    if (node.first->getKind() == Kinded::Kind::ConstantKind) {
+      auto constant = static_cast<const Constant *>(node.first);
```
jfix71: nit: `auto *constant`

lib/IR/IR.cpp
```diff
@@ -428,6 +428,28 @@ static void dumpIRInContext(const Value *V, llvm::raw_ostream &out) {
...
+    }
+  }
+  return constants;
+}
+
+std::list<const Placeholder *> IRFunction::getPlaceholders() const {
+  std::list<const Placeholder *> placeholders;
+  for (auto &node : variableMap_) {
+    if (node.first->getKind() == Kinded::Kind::PlaceholderKind) {
+      auto constant = static_cast<const Placeholder *>(node.first);
```
jfix71: nit: `auto *PH`

include/glow/Graph/Graph.h
bertmaher: IMO it should be `Metadata` everywhere since it's one word, but I agree it should be consistent!

include/glow/Graph/Graph.h
```diff
@@ -199,6 +199,9 @@ class Function final : public Named {
   /// A list of nodes that the Function owns.
   NodesList nodes_;
 
+  /// A list of metadata PHs associated with the function.
+  std::list<Placeholder *> metaDataPlaceholders_;
```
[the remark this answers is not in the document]
bertmaher: +1, Also if it must be `list` we have a `PlaceholderList` typedef up above, might as well use it.

lib/Graph/Graph.cpp
```diff
@@ -2689,6 +2689,34 @@ void Function::eraseNode(Node *N) {
   eraseNode(N->getIterator());
 }
 
+PlaceholderList Function::getPlaceholders() {
+  PlaceholderList list;
+  for (auto &node : nodes_) {
+    for (unsigned i = 0; i < node.getNumInputs(); i++) {
+      auto input = node.getNthInput(i).getNode();
+      if (input->getKind() == Kinded::Kind::PlaceholderKind) {
```
bertmaher: how about:
```
if (auto *PH = llvm::dyn_cast<Placeholder>(input)) { ...
```
?

lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -238,6 +238,9 @@ void BundleSaver::emitBundleConfig() {
       llvm::StructType::get(irgen_->getLLVMContext(),
                             {SizeTType, SizeTType, SizeTType, SizeTType,
                              SizeTType, symbolTableEntryTy->getPointerTo()});
+  llvm::outs() << "mutable: "
```
bertmaher: Revert the debug prints

include/glow/Graph/Graph.h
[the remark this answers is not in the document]
gcatron: Good catch, I'll change it to a vector.

include/glow/Graph/Graph.h
gcatron: I'll make them all Metadata, and append Placeholders to getMetadata.

tests/unittests/BackendTest.cpp
gcatron: I originally was checking symbol count, but that changes if we run with optimizeIR on or off.
That's why I switch to checking that symbols are not shared when they shouldn't be.

tests/unittests/BackendTest.cpp
```diff
@@ -188,6 +190,40 @@ TEST_P(BackendTest, CompileWithoutConstants) {
   auto function = backend->compile(F, opts);
 }
 
+/// Test that the runtimeBundle includes only symbols from it's function and not
+/// the whole module.
+TEST_P(BackendTest, BundleFunctionSymbolsOnly) {
+  Module mod;
+  PlaceholderBindings ctx;
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createPlaceholder(ElemKind::FloatTy, {3}, "X", false);
```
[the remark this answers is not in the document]
gcatron: Good idea, I'll make the input a constant. And add a test where the two functions use the same constant as input.

lib/IR/GraphScheduler.cpp
gcatron: Since moving to walking the modules placeholders and constants I don't think we can add a duplicate anymore. So we should be good.

lib/Graph/Graph.cpp
```diff
@@ -2689,6 +2689,34 @@ void Function::eraseNode(Node *N) {
   eraseNode(N->getIterator());
 }
 
+PlaceholderList Function::getPlaceholders() {
+  PlaceholderList list;
+  for (auto &node : nodes_) {
```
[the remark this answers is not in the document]
gcatron: Good point. I'll make the change.

gcatron:
> I'm feeling uneasy about replacing a simple getter with an O(N) function that allocates a bunch of elements in a `std::list`. As @jfix71 points out there are places where this function is called redundantly, leading to multiple scans of the graph.
> 
> I don't have a great suggestion for improvement, other than maybe calling it `findConstants` and only using it where it actually matters (so like, maybe `scheduleGraph` can just use the `Module`).

I like the idea of find instead of get, I'll make that change. ScheduleGraph is currently the only callsite so I'll call once and store the result.

include/glow/IR/IR.h
```diff
@@ -320,6 +320,12 @@ class IRFunction final : public Named {
   /// \returns the variable map.
   VariableMap &getVariableMap() { return variableMap_; }
 
+  /// Returns a list of constants associated with function.
+  std::list<const Constant *> getConstants() const;
```
jfix71: Why not use `ConstList`?

include/glow/IR/IR.h
```diff
@@ -320,6 +320,12 @@ class IRFunction final : public Named {
...
+
+  /// Returns a list of placeholders associated with the function.
+  std::list<const Placeholder *> getPlaceholders() const;
```
jfix71: Why not use `PlaceholderList`?

lib/IR/IR.cpp
```diff
@@ -428,6 +428,28 @@ static void dumpIRInContext(const Value *V, llvm::raw_ostream &out) {
   }
 }
 
+std::list<const Constant *> IRFunction::getConstants() const {
+  std::list<const Constant *> constants;
+  for (auto &node : variableMap_) {
+    if (node.first->getKind() == Kinded::Kind::ConstantKind) {
```
jfix71: Can't we just do:

```
if (Constant *constant = dyn_cast<Constant>(node.first)) {
  constants.push_back(constant);
}
```

And same for Placeholders?

lib/IR/IR.cpp
```diff
@@ -428,6 +428,28 @@ static void dumpIRInContext(const Value *V, llvm::raw_ostream &out) {
   }
 }
 
+std::list<const Constant *> IRFunction::getConstants() const {
```
jfix71: Do we want to call this `findConstants()` too, since it's also not just a simpler getter? Same for `getPlaceholders()`.

tests/unittests/BackendTest.cpp
```diff
@@ -188,6 +190,67 @@ TEST_P(BackendTest, CompileWithoutConstants) {
   auto function = backend->compile(F, opts);
 }
 
+/// Test that the runtimeBundle includes only symbols from its function and not
+/// the whole module.
+TEST_P(BackendTest, BundleFunctionSymbolsOnly) {
+  Module mod;
+  PlaceholderBindings bindings;
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createConstant(ElemKind::FloatTy, {3}, "Y");
```
jfix71: minor nit: Called `"Y"` but var name is `X` 😛

tests/unittests/BackendTest.cpp
```diff
@@ -188,6 +190,67 @@ TEST_P(BackendTest, CompileWithoutConstants) {
...
+    EXPECT_TRUE(it == table2.end());
+  }
+}
+
+/// Test that a shared constant is in the bundle of both functions.
+TEST_P(BackendTest, BundleSharedConstant) {
+  Module mod;
+  PlaceholderBindings bindings;
+  Function *F = mod.createFunction("main");
+  auto *X = mod.createConstant(ElemKind::FloatTy, {3}, "Y");
```
jfix71: Same minor nit!

tests/unittests/BackendTest.cpp
```diff
@@ -188,6 +190,67 @@ TEST_P(BackendTest, CompileWithoutConstants) {
...
+  Function *F2 = mod.createFunction("main2");
+  auto *X2 = mod.createConstant(ElemKind::FloatTy, {3}, "X2");
+  X2->getHandle() = {1., 2., 3.};
+  auto *pow2 = F2->createPow("Pow2", X2, 2.0);
+  auto *save2 = F2->createSave("save2", pow2);
+  bindings2.allocate(save2->getPlaceholder());
+
+  std::unique_ptr<Backend> backend(createBackend(GetParam()));
+  CompilationOptions opts;
+  auto function = backend->compile(F, opts);
```
jfix71: nit: Using opts only on one `compile()` call -- can probably remove it and just use `compile(F)` on both. Same for other test below.

include/glow/IR/IR.h
gcatron: It's a list of const Constants, ConstList is just a list of Constants. :D

include/glow/IR/IR.h
gcatron: It's a list of const Constants, ConstList is a list of Constants

Issue #2501, opened 2019-03-08
[ModelZoo] support arcFace model


nickgg:
Add support for the arcFace Face Recognition model in glow (https://github.com/onnx/models/tree/master/models/face_recognition/ArcFace) and verify via a runnable example we can include in the ModelZoo regression test.

This model appears to be blocked by operator support in Glow, we are at least missing the PRelu operator.

This issue blocks #2500.

PR #2619, opened 2019-03-29
[New Operator] PRelu node creation, lowering and unittest

mortzur:
Implementing PRelu node creation and lowering (required for #2501).
PRelu is parameterized leaky Relu:
slope * x      if x < 0
x                   if x > =0
where slope is a parameter (not constant).
Lowering is implemented from the above description by:
select(cmpLTE(x, 0), mul(slope, x), x)

Unit tests pass.

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -353,6 +353,19 @@ static bool verifyRelu(NodeValue result, NodeValue input) {
   return checkSameType(result, input, parent);
 }
 
+static bool verifyPRelu(NodeValue result, NodeValue input, NodeValue slope) {
+  const Node *parent = result.getNode();
+  if (input.getType()->isQuantizedType()) {
+    return checkSameIsQuantized(input.getType(), result.getType(), parent) &&
+           checkSameIsQuantized(input.getType(), slope.getType(), parent) &&
+           checkSameShape(result, input, parent) &&
+           checkSameShape(slope, input, parent);
+  }
+  return checkSameType(result, input, parent) &&
+         checkSameType(slope, input, parent) &&
+         checkSameShape(slope, input, parent);
+}
+
 static bool verifyRegression(NodeValue src, NodeValue dest,
                              NodeValue expected) {
   return checkSameType(src, dest, dest.getNode()) &&
@@ -1159,6 +1172,10 @@ bool ReluGradNode::verify() const {
                                         getGradOfOriginalOutputNamedResult(),
                                         this) &&
          verifyRelu(getGradOfOriginalOutputNamedResult(), getInput());
+}
+
+bool PReluNode::verify() const {
+  return verifyPRelu(getResult(), getInput(), getSlope());
 }
 
 bool RegressionNode::verify() const {
--- a/lib/Optimizer/Lower.cpp
+++ b/lib/Optimizer/Lower.cpp
@@ -219,6 +219,20 @@ static void lowerReluNode(Function *F, LoweredInfoMap *loweredMap,
   replaceAllUsesOfWith(loweredMap, R.getResult(), relu);
 }
 
+static void lowerPReluNode(Function *F, LoweredInfoMap *loweredMap,
+                           const PReluNode &R) {
+  // PRelu is :
+  // slope * x    if x < 0
+  // x            if x >= 0
+  // where slope is an input from a different node.
+  auto *zeroSplat = F->createSplat("zeroSplat", R.getResult().getType(), 0.0);
+  auto *cmplgt = F->createCmpLTE("cmplgt", zeroSplat, R.getInput());
+  auto *mul = F->createMul("mul", R.getSlope(), R.getInput());
+  auto *prelu = F->createSelect("prelu", cmplgt, R.getInput(), mul);
+
+  replaceAllUsesOfWith(loweredMap, R.getResult(), prelu);
+}
+
 static void lowerPadNode(Function *F, LoweredInfoMap *loweredMap,
                          const PadNode &P) {
   auto *outputType = P.getResult().getType();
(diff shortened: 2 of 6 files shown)


tools/ClassGen/NodeGen.cpp
```diff
@@ -460,6 +460,14 @@ int main(int argc, char **argv) {
       .setDocstring(
           "Applies ReLU, max(0, x), to each element in the Input tensor.");
 
+  BB.newNode("PRelu")
+      .addInput("Input")
+      .addInput("Slope")
+      .addResultFromCtorArg()
+      .addGradient()
```
rdzhabarov: how do we handle PReLU gradient node? (i'd exclude it for now until support is not added).

tests/unittests/OperatorTest.cpp
```diff
@@ -1025,6 +1025,42 @@ TEST_P(OperatorTest, ReluSimple_Float16) {
...
+  }
+}
+
+/// Verify that the PRELU operator works correctly for Float.
+TEST_P(OperatorTest, PReluSimple_Float) {
+  testPReluSimple<float>(bindings_, mod_, F_, EE_, ElemKind::FloatTy);
+}
+
+/// Verify that the PRELU operator works correctly for Float16.
+TEST_P(OperatorTest, PReluSimple_Float16) {
```
rdzhabarov: I see ReLU follows this, but Name_Something is not the convention we follow (remove _)

tests/unittests/OperatorTest.cpp
```diff
@@ -1025,6 +1025,42 @@ TEST_P(OperatorTest, ReluSimple_Float16) {
   testReluSimple<float16_t>(bindings_, mod_, F_, EE_, ElemKind::Float16Ty);
 }
 
+/// Helper to test ReluSimple using \p DTy.
```
rdzhabarov: fix comment.

tests/unittests/OperatorTest.cpp
```diff
@@ -1025,6 +1025,42 @@ TEST_P(OperatorTest, ReluSimple_Float16) {
...
+  bindings.allocate(slope)->getHandle<DataType>() = {1, 1, 1, 1, 1, 1, 1};
+
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+
+  auto resultH = result->getHandle<DataType>();
+  auto inH = bindings.get(in)->getHandle<DataType>();
+
+  for (size_t i = 0; i < 7; i++) {
+    EXPECT_EQ(resultH.raw(i), inH.raw(i));
```
rdzhabarov: it may feel like it's always slope * input, can you change slope so that we better exercise that it's x for positive and slope*x for negative.

tests/unittests/OperatorTest.cpp
jfix71: I actually added underscores to some unit tests to specify types. We can remove them if desired -- but I thought it was a bit cleaner to separate operator from type being tested.

rdzhabarov: that's ok.

tools/ClassGen/NodeGen.cpp
mortzur: following our talk - removing PReluGrad node.

lib/Graph/Nodes.cpp
```diff
@@ -353,6 +353,17 @@ static bool verifyRelu(NodeValue result, NodeValue input) {
...
 }
 
+static bool verifyPRelu(NodeValue result, NodeValue input, NodeValue slope) {
+  const Node *parent = result.getNode();
+  if (input.getType()->isQuantizedType()) {
+    return checkSameIsQuantized(input.getType(), result.getType(), parent) &&
+           checkSameShape(result, input, parent);
+  }
+  return checkSameType(result, input, parent) &&
+         checkSameShape(result, input, slope) &&
```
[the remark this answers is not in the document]
mortzur: You are right. I misunderstood the interface, fixed this (amended to this PR).
Goal here is to verify that input, slope and result have the same type and shape.

lib/Optimizer/Lower.cpp
```diff
@@ -219,6 +219,20 @@ static void lowerReluNode(Function *F, LoweredInfoMap *loweredMap,
   replaceAllUsesOfWith(loweredMap, R.getResult(), relu);
 }
 
+static void lowerPReluNode(Function *F, LoweredInfoMap *loweredMap,
+                           const PReluNode &R) {
+  // PRelu is :
+  // slope * x    if x < 0
+  // x            if x >= 0
+  // where slope is a variable.
```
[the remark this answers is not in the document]
mortzur: Good point. I meant that this is not necessarily a const but an input of a different function (which is what the description says now). The documentation doesn't state anything about if the values should be identical (https://github.com/onnx/onnx/blob/fb1a80692c1ab0bd27b1072f2e7bffacba336777/docs/Operators.md#PRelu). 
I changed the unit-test accordingly to check for a general case.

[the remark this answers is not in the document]
jfix71: I'm not sure "function" makes sense for us either since we have `Functions` that are essentially a container for series of Nodes 😄-- do you just mean the `slope` input is an output from another `Node`?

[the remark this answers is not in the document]
mortzur: Yes... Not used to the terminology. Will get there. Changing to "node".

mortzur:
@rdzhabarov - yes , I'm going to add an ONNXModelLoader support in a following commit. Should I do this in this branch or in a new one (after this PR lands)?

rdzhabarov:
Yeah, I did not mean to block this PR. It's good to go.

nadavrot:
@mortzur Congrats.

PR #2621, opened 2019-03-29
Fix for windows onnxifi. 

ayermolo:
*Description*: Windows doesn't have getuid() and <unistd.h>.
*Testing*:
*Documentation*:
[Optional Fixes #issue] @jackm321 

### DIFF
--- a/lib/Onnxifi/InlineOnnxifi.cpp
+++ b/lib/Onnxifi/InlineOnnxifi.cpp
@@ -22,14 +22,28 @@
 
 #include "llvm/Support/MD5.h"
 
+#ifdef WIN32
+#include <Lmcons.h>
+#include <windows.h>
+#else
 #include <unistd.h>
+#endif // !
 
 namespace glow {
 namespace onnxifi {
 
 namespace {
 std::string getProfileFile(llvm::StringRef hash) {
-  return strFormat("/tmp/glow-profile-%d-%s.yaml", getuid(),
+  std::string sUID;
+#ifdef WIN32
+  char username[UNLEN + 1];
+  DWORD username_len = UNLEN + 1;
+  GetUserName(username, &username_len);
+  sUID = username;
+#else
+  sUID = getuid();
+#endif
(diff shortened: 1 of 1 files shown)


lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -22,14 +22,28 @@
...
+  std::string sUID;
+#ifdef WIN32
+  char username[UNLEN + 1];
+  DWORD username_len = UNLEN + 1;
+  GetUserName(username, &username_len);
+  sUID = username;
+#else
+  sUID = getuid();
+#endif
+  return strFormat("/tmp/glow-profile-%s-%s.yaml", sUID.c_str(),
```
rdzhabarov: question. Do we actually need a user id here?

jackm321: @rdzhabarov I moved this code without trying to change it as much as possible. I agree that this seems unnecessary and would rather see it go.

nadavrot: Okay. If we can get rid of getuid and GetUserName() that would be great. It would simplify the code significantly. Let's do it. 

Sorry for the delay @ayermolo .

rdzhabarov:
Yup, let's just remove userid thing and good to go.

lib/Onnxifi/InlineOnnxifi.cpp
ayermolo: No problem. Is there an ETA for this fix?

rdzhabarov: https://github.com/pytorch/glow/pull/2635

fixed here.

ayermolo: Awesome, thanks!

rdzhabarov:
should be fixed by https://github.com/pytorch/glow/pull/2635

PR #2626, opened 2019-03-30
[AOT] fix cross platform issue: using uint64_t to replace size_t.

### LINKED ISSUE #2592: [cross platform] 'size_t' in struct BundleConfig may cause platform mismatch
hi,
I am tring Ahead-of-time Compilation example : resnet50.

function 'emitBundleConfig' emit BundleConfig struct, and there are many 'size_t' member. it may cause some problem when cross platform :

while build a 32-bit binary on 64-bit, (for example build armv7 binary on x86_64 ubuntu). emit resnet50.o's size_t is 64-bit, but main.cpp.o's size_t is 32-bit (which is compiled by arm compiler).

so, is it OK to modify BundleConfig's size_t to be 'uint64_t'; Or modify function 'emitBundleConfig' to emit different 'size_t', according to the target's size_t.

xizhizhang:
Description: Ahead-of-time Compilation cross platform issue: build on 64-bit OS and run on 32-bit OS.
Testing: x86_64 and armv7 platform.
Documentation: modify AOT.md

Fixes #2592


examples/bundles/resnet50/main.cpp
```diff
@@ -17,6 +17,7 @@
 #include <png.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
```
jfix71: This isn't necessary is it?

xizhizhang: > This isn't necessary is it?

it is trying to fix compile error.

My environment is ubuntu 18.04, clang-6(which is from "apt install") and llvm-7(build from source code). 
my cmake command is:

> cmake .. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DGLOW_WITH_BUNDLES=ON -DGLOW_WITH_CPU=ON -DCMAKE_PREFIX_PATH=/home/zhangwm/git/glow/llvm_install/

> make ResNet50Bundle

then, error message is:

> 

> /home/zhangwm/git/glow/examples/bundles/resnet50/main.cpp:252:52: error: use of undeclared identifier 'strlen'
>     if (!strncmp(config.symbolTable[i].name, name, strlen(name))) {
>                                                    ^
> /home/zhangwm/git/glow/examples/bundles/resnet50/main.cpp:281:3: error: use of undeclared identifier 'memset'; did you mean
>       'wmemset'?
>   memset(ptr, 0, size);
>   ^~~~~~
>   wmemset
> /usr/include/wchar.h:271:17: note: 'wmemset' declared here
> extern wchar_t *wmemset (wchar_t *__s, wchar_t __c, size_t __n) __THROW;

opti-mix: Shouldn't memset be declared in `<cstring>`? See https://en.cppreference.com/w/cpp/string/byte/memset

xizhizhang: > Shouldn't memset be declared in `<cstring>`? See https://en.cppreference.com/w/cpp/string/byte/memset

yeah, seems csting is better than string.h in c++,  but it has same functions  with 
 string.h.

here is the cstring introduce page: https://en.cppreference.com/w/cpp/header/cstring

> This header was originally in the C standard library as <string.h>.
> This header is for C-style null-terminated byte strings.

 Since I am familiar c than c++, and there are already some head files like "stdlib.h" in main.cpp, then I used "string.h".

Should I recreate  PR  to modify to cstring, or this PR is just OK?

opti-mix: @xizhizhang 

> Should I recreate PR to modify to cstring, or this PR is just OK?

No need to modify the PR. `cstring` is not any better than `string.h`

xizhizhang: > @xizhizhang
> 
> > Should I recreate PR to modify to cstring, or this PR is just OK?
> 
> No need to modify the PR. `cstring` is not any better than `string.h`

OK，thanks

lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -82,13 +82,13 @@ void BundleSaver::emitSymbolTable() {
...
   // struct SymbolTableEntry {
   //  const char *name;
-  //  size_t offset;
-  //  size_t size;
+  //  uint64_t offset;
+  //  uint64_t size;
   //  char kind;
   // };
   auto *charTy = llvm::Type::getInt8Ty(irgen_->getLLVMContext());
   auto *sizeTTy =
```
opti-mix: The name `sizeTTy` doesn't make a lot of sense if we use `uint64_t` now

lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -219,20 +219,20 @@ void BundleSaver::emitBundleEntryFunction() {
...
 // };
 void BundleSaver::emitBundleConfig() {
   auto symbolTable = irgen_->getModule().getGlobalVariable(
       irgen_->getMainEntryName() + "SymbolTable", true);
   GLOW_ASSERT(symbolTable &&
               "Expected to find a symbol table for the AOT bundle");
-  // Get the integer type having the same size in bits as size_t.
-  auto *SizeTType = irgen_->getBuilder().getIntNTy(sizeof(size_t) * 8);
+  // Get the integer type having the same size in bits as uint64_t.
+  auto *SizeTType = irgen_->getBuilder().getIntNTy(sizeof(uint64_t) * 8);
```
opti-mix: The name `SizeTTy` doesn't make a lot of sense if we use uint64_t now. Also the name should start with a lower-case.

xizhizhang:
> @opti-mix I think I'd still prefer using `uint64_t`. And actually looking at the PR again, do we need to do so many `size_t` -> `uint64_t` changes? @xizhizhang How did you decide what to change to `uint64_t`, was it just a replace all `size_t` with `uint64_t`? E.g. You moved `imageDataSizeInBytes` to `uint64_t`, but it's used for `memcpy()` which takes `size_t`. Perhaps we should still use `size_t` in many places and just verify that the `uint64_t` in `BundleConfig` do not overflow `size_t`?

I am sorry I just replace all size_t with `uint64_t`. You are right, we need not so many `uint64_t`.

I replaced them because I did not want to see any `size_t`. As an embedded C programmer, I have meet a lot of similar cross platform problem because of `size_t / int / long` (something looks designed for cross platform, but actually confuse  and raise bugs).

@opti-mix Yeah, I prefer a target-specific size, but not `size_t` . Because, `uint64_t` in `BundleConfig` makes application running right, but, there is no need 64-bit member always(E.g. alignment / numSymbols). And 64-bit member always need several instructions to finish basic calculation (E.g. add/sub/mul/div) on 32-bit CPU.

In my opinion, "all `size_t`" -> "all `uint64_t`" -> "every value have it's best type" helps me develop application and debug.
on the other hand, "all size_t" -> "some size_t and some uint64_t" is another way, maybe it is more orthodox method, and better to keep code stable.

Any way, I would like to follow this project line of thinking, let me known how should I modify this pull requests.

lib/LLVMIRCodeGen/BundleSaver.cpp
petecoup: Does it make more sense to use the newly introduced `LLVMIRGen::getTargetSizeTWidth`?

opti-mix: @petecoup If we decide to go for target-specific `size_t` usage, we would definitely use `LLVMIRGen::getTargetSizeTWidth`. 

We are still trying to decide if it is better to use fixed-size fields instead, so that the `BundleConfig` has the same binary layout on all platforms and can be easily read e.g. by tools, helper functions, etc without thinking about any target-specific aspects.

petecoup: Sorry, I somehow missed the later comments...this makes sense!

xizhizhang:
> @xizhizhang I would say to basically change only `BundleConfig` to `uint64_t`, and keep as much otherwise in `size_t`, perhaps adding checks for overflowing `size_t` as necessary. And these are just examples for bundles, after all -- if you prefer all fixed width sizes instead then you could always create your own example.
> 
> Please address @opti-mix's comments as well as changing back as many `uint64_t` to `size_t` as possible, and this should be good to go. Thanks for your contributions!

Thanks, but I keep `uint64_t `in `SymbolTableEntry`, it looks have similar problem with `BundleConfig` .

@opti-mix I renamed `SizeTTy `to `uint64TTy`

opti-mix:
@xizhizhang Looks good!

But our CI complains about bad formatting:
```
diff --git a/lib/LLVMIRCodeGen/BundleSaver.cpp b/lib/LLVMIRCodeGen/BundleSaver.cpp
index d3d95da..098341d 100644
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -89,9 +89,9 @@ void BundleSaver::emitSymbolTable() {
   auto *charTy = llvm::Type::getInt8Ty(irgen_->getLLVMContext());
   auto *uint64TTy =
       llvm::Type::getIntNTy(irgen_->getLLVMContext(), sizeof(uint64_t) * 8);
-  auto symbolTableEntryTy =
-      llvm::StructType::get(irgen_->getLLVMContext(),
-                            {charTy->getPointerTo(), uint64TTy, uint64TTy, charTy});
+  auto symbolTableEntryTy = llvm::StructType::get(
+      irgen_->getLLVMContext(),
+      {charTy->getPointerTo(), uint64TTy, uint64TTy, charTy});
```

Could you run `utils/format.sh` on your sources?

xizhizhang:
> @xizhizhang Looks good!
> 
> But our CI complains about bad formatting:
> 
> ```
> diff --git a/lib/LLVMIRCodeGen/BundleSaver.cpp b/lib/LLVMIRCodeGen/BundleSaver.cpp
> index d3d95da..098341d 100644
> --- a/lib/LLVMIRCodeGen/BundleSaver.cpp
> +++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
> @@ -89,9 +89,9 @@ void BundleSaver::emitSymbolTable() {
>    auto *charTy = llvm::Type::getInt8Ty(irgen_->getLLVMContext());
>    auto *uint64TTy =
>        llvm::Type::getIntNTy(irgen_->getLLVMContext(), sizeof(uint64_t) * 8);
> -  auto symbolTableEntryTy =
> -      llvm::StructType::get(irgen_->getLLVMContext(),
> -                            {charTy->getPointerTo(), uint64TTy, uint64TTy, charTy});
> +  auto symbolTableEntryTy = llvm::StructType::get(
> +      irgen_->getLLVMContext(),
> +      {charTy->getPointerTo(), uint64TTy, uint64TTy, charTy});
> ```
> Could you run `utils/format.sh` on your sources?

Thanks. I have forgotten it.

opti-mix:
@jfix71 Could you put your stamp? And let's merge it!

PR #2627, opened 2019-03-31
Add glog

jackm321:
*Description*:
Add glog to Glow and add a couple example uses.

*Testing*:
Added a DCHECK(false) and compiled with debug mode on and crashes occur.

### DIFF
--- a/lib/Base/CMakeLists.txt
+++ b/lib/Base/CMakeLists.txt
@@ -6,7 +6,9 @@ add_library(Base
 
 target_link_libraries(Base
                       PUBLIC
-                        Support)
+                        Support
+                        glog)
+
 if(PNG_FOUND)
   target_sources(Base PRIVATE Image.cpp)
   target_include_directories(Base
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -42,7 +42,7 @@ GLOW_DIR=$PWD
 
 # Install Glow dependencies
 sudo apt-get update
-sudo apt-get install -y llvm-6.0 llvm-6.0-dev libpng-dev
+sudo apt-get install -y llvm-6.0 llvm-6.0-dev libpng-dev libgoogle-glog-dev
 
 # Redirect clang
 sudo ln -s /usr/bin/clang-6.0 /usr/bin/clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
         - TEST_NAME=COVERAGE
       before_install:
         - utils/install_protobuf.sh
-        - sudo apt-get install -y lcov
+        - sudo apt-get install -y lcov libgoogle-glog-dev
         - sudo pip install awscli --upgrade
         - sudo apt-get install -qq g++-4.8
         - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 90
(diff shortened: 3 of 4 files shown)


jackm321:
@nadavrot thanks for starting the discussion.
The main thing that I want to have is non-static log messages for when assertions are not met. Do you know of something in our existing dependencies that can do this? Would you prefer building something to meet this need?

Right now the overhead of having to add logs to print out additional information when an assertion is hit, recompile everything, rerun (hope to hit the same assertion again), delete the logs, and repeat can actually be pretty high especially when not just compiling Glow but compiling a larger service that is using Glow. Also this is wasted effort because the next person who hits the same assertion will need to do the same thing as well.

jackm321:
I checked llvm's coding standards and from what I could find [they support the use of assert](https://llvm.org/docs/CodingStandards.html#assert-liberally) but for the reasons mentioned above I think there are things that assert does not provide that would be helpful to have in Glow.

nadavrot:
LLVM uses assertions for checking for internal software bugs. Validation of inputs is used by other means. Their philosophy is that if we are debugging an issue we can always run the same input again with a debug build of the compiler with assertions enabled. Enabling assertions always has a huge runtime cost and we want to limit these checks to debug mode builds. 

In Glow we have the poorly named GLOW_ASSERT, which does not go away in debug builds. This is used to validate input. We may want to rename GLOW_ASSERT to GLOW_CHECK or something like that. 

Do you want to change the behavior of GLOW_ASSERT? or do you want to convert some asserts to GLOW_ASSERT?

jackm321:
It's probably true that we can nearly always rerun failing networks with debug on at least for the compile part. I think the issue is more that this will require time because to debug something we have to add printing, rebuild, rerun, etc which can take a long time for some services.
It's one way to work but I've just found myself a little frustrated recently with the amount of time it takes to do this.

GLOW_ASSERT does seem badly named but wouldn't help with this issue because it still just prints out the condition without any extra information about the specific values that caused the error.

bertmaher:
A few pseudo-random thoughts on glog/logging more generally:

1. There's kind of a philosophical difference between static compilers and server applications regarding logging.  Probably because you can easily ("easily") repro issues with static compilers in a debug setting, whereas server applications might be very involved to get a repro.
2. Glow is somewhere in between the two.  Especially with the Glow "runtime", we're leaning more into a server-application domain where it might be important to get logs live.
3. Not having good logs might make us code more defensively and thus ship fewer bugs in the first place ;-).  I'm not sure whether this is a pro or a con.
4. We could rewrite a bunch of our asserts to be something like:
```
  llvm::errs() << "Dying because " << xyz << " Is unsupported\n";
  assert(false && "Hey dude read the message above!");
```
That gets us part of the way there (although not entirely, since you don't have things like configurable logging levels).
5. Glog is a pretty stable dep as these things go.

lib/Base/CMakeLists.txt
```diff
@@ -6,7 +6,11 @@ add_library(Base
 
 target_link_libraries(Base
                       PUBLIC
-                        Support)
+                        Support
+                        glog)
+
+target_include_directories(Base PUBLIC ${GLOW_THIRDPARTY_DIR}/glog)
```
rdzhabarov: who is going to put it there (`${GLOW_THIRDPARTY_DIR}/glog`)?

jackm321: Ah I need to remove this. I originally was going to go this route but I realized depending on glog externally is maybe better since it's so readily available.

nadavrot:
Hey, stale bot, hold on. This is happening. Patience please.

PR #2630, opened 2019-04-01
Optimize TransposeNode into ReshapeNode when it moves no data.

### LINKED ISSUE #2629: TransposeNode should be optimized into ReshapeNode when it moves no data.
When `TransposeNode ` moves no data it makes sense to optimize it into `ReshapeNode ` which allows for further optimizations that already exist on reshape nodes.

vuzelac-cadence:
*Description*:
`TransposeNode` is optimized into `ReshapeNode` when it moves no data.  This allows for further optimizations. Currently, the algorithm only looks for cases when dimensions of size 1 are shuffled around. 
*Testing*:
GraphOptzTest: Add a testcase that covers `TransposeNode` substitution with `ReshapeNode`.
OperatorTest: Add a testcase that tests if the optimized model is still functionally correct. 
*Documentation*:
Updated Optimizations.md
Fixes #2629

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1807,6 +1807,37 @@ static void optimizeSliceOfSplat(Function *F) {
   }
 }
 
+/// Optimize TransposeNode into ReshapeNode when it actually moves no data.
+static void optimizeTransposeIntoReshape(Function *F) {
+  for (auto &node : F->getNodes()) {
+    auto *TR = dyn_cast<TransposeNode>(&node);
+    if (!TR)
+      continue;
+    auto inputNode = TR->getInput();
+    auto inputDims = inputNode.dims();
+    auto outputDims = TR->getResult().dims();
+    // Transpose moves no data if input/output dimensions match after they both
+    // drop dimensions of size 1. E.g. transposing [1 5 1 15] into [5 15 1 1]
+    // produces vectors (1, 3) for both dimensions so optimization is executed.
+    auto shuffle = TR->getShuffle();
+    ShapeVector inDims;
+    ShapeVector outDims;
+    for (size_t i = 0; i < inputDims.size(); i++) {
+      if (inputDims[i] != 1) {
+        inDims.push_back(i);
+      }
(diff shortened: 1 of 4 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -1807,6 +1807,37 @@ static void optimizeSliceOfSplat(Function *F) {
...
+static void optimizeTransposeIntoReshape(Function *F) {
+  for (auto &node : F->getNodes()) {
+    auto *TR = dyn_cast<TransposeNode>(&node);
+    if (!TR)
+      continue;
+    auto inputNode = TR->getInput();
+    auto inputDims = inputNode.dims();
+    auto outputDims = TR->getResult().dims();
+    // Transpose moves no data if input/output dimensions match after they both
+    // drop dimensions of size 1. E.g. transposing [1 5 1 15] into [5 15 1 1]
```
rdzhabarov: you do not really need additional memory here.

you could just do a pass over input and shuffle in a manner similar to merge sort.
you skip elem equals to 1 and then check if non 1 are equal from input and shuffle.

vuzelac-cadence: That was the original idea but this looked like a cleaner implementation (at least code wise). I could certainly finalize that other approach if you see this as having unacceptable performance impact.

rdzhabarov: Engineering excellence took over me. I do not think there is a significant perf impact here, considering that there are bunch of O(n*n) optimizations.

Either way is acceptable in this case, it's just I'm not sure if the other implementation is hard to follow.

vuzelac-cadence:
@opti-mix , @rdzhabarov , an issue showed up later in our bigger test setup; After this optimization, DCE is needed to remove multiple users from some nodes, which will prohibit optimizations such as Reshape(Constant). I have a fix for DCE, as well as the test that covers what should have failed. Please advice what's the next step - new Issue, reopen old issue, just Revert changes here or simply push more. Thx.

vuzelac-cadence:
#2666 opened.

PR #2632, opened 2019-04-01
[tracing] Add TraceEvent support for OpenCL

nickgg:
*Description*: Adds both auto instrumentation and manual TraceEvents for the OpenCL backend. This builds on the existing profiling support leaving it unchanged (ie. `-opencl-profile` still works).

OpenCL has a bit of concurrency in its execution, e.g. below you can see the enqueueKernels stage overlapping with the execution of early kernels in the run. We could model this as separate threads in the future but not worrying too much about how it looks yet.
*Testing*: unit tests (particularly TraceEventTests)+ the tracing-compare example.
*Documentation*:
looks like this"
image

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -120,7 +120,8 @@ static void addStringOption(std::vector<std::string> &options,
 }
 
 OpenCLFunction::OpenCLFunction(std::unique_ptr<IRFunction> F,
-                               const runtime::RuntimeBundle &bundle)
+                               const runtime::RuntimeBundle &bundle,
+                               TraceInfo traceInfo)
     : CompiledFunction(bundle), F_(std::move(F)) {
   cl_uint numPlatforms{0};
   cl_int err = clGetPlatformIDs(0, NULL, &numPlatforms);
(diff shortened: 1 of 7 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1557,36 +1578,129 @@ void OpenCLFunction::freeDeviceBuffer(cl_mem buf) { clReleaseMemObject(buf); }
...
+
+    clGetEventProfilingInfo(event, CL_PROFILING_COMMAND_START,
+                            sizeof(time_start), &time_start, NULL);
+    clGetEventProfilingInfo(event, CL_PROFILING_COMMAND_END, sizeof(time_end),
+                            &time_end, NULL);
+
+    if (type == "checkpoint") {
+      const auto &it = manualTraceEvents_.find(name);
+      if (it == manualTraceEvents_.end()) {
+        llvm::outs() << "warning: found manual trace event (" << name
```
opti-mix: Do we really want to print it each time to the standard output? Should be DEBUG_GLOW()? 

Or may be we should introduce something like a general-purpose way to produce messages in Glow? E.g. `glowProduceMessage(MessageKind, format, ...)`, where MessageKind is an enum defining INFO, ERROR, WARNING, etc?

nickgg: sounds suspiciously like glog...

I'll change it to DEBUG_GLOW for this PR.

PR #2639, opened 2019-04-02
Take NodeValue instead of Node * as arguments for createFullyConnected functions

opti-mix:
We generally take operands as `NodeValue`s in such node creation functions. These functions are pretty old and were never converted to use this style.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -350,8 +350,9 @@ public:
   /// \p W, bias \p B, and \p outTy. If \p input is not 2 dimensional then it is
   /// flattened along \p axis. Note, outputDepth is inferred based on \p outTy.
   FullyConnectedNode *createFullyConnected(llvm::StringRef name,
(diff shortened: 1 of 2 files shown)


include/glow/Graph/Graph.h
```diff
@@ -350,8 +350,9 @@ class Function final : public Named {
   /// \p W, bias \p B, and \p outTy. If \p input is not 2 dimensional then it is
   /// flattened along \p axis. Note, outputDepth is inferred based on \p outTy.
   FullyConnectedNode *createFullyConnected(llvm::StringRef name,
-                                           NodeValue input, Node *W, Node *B,
-                                           TypeRef outTy, unsigned_t axis = 1);
+                                           NodeValue input, NodeValue W,
```
rdzhabarov: what about the one above `line 345`

rdzhabarov: ok, let Storage be :D

PR #2642, opened 2019-04-02
[tracing] Add TraceEvents into the Runtime Execution Engine

nickgg:
*Description*: Adds some simple events in the Executor and the HostManager and adds the plumbing to enable tracing and retrieve TraceEvents (auto instrumentation is not yet supported).

Also fix a couple bugs, particularly we now export timestamps in the right domain. Next steps are to improve the process and thread labels and look at showing thread movement flow.

We treat the whole runtime as a monolithic "thread" even though it has several. Uncertain if we get much from modeling internal threads in the runtime here, open to opinions.

*Testing*: unittests with asan & the upgraded resnet-runtime example 
*Documentation*:
Here's the output of the resnet-runtime example with tracing enabled (6 CPU devices, 100 images):
image
Thread #50 is a placeholder for the Runtime threads in this example. You can see it is busy on and off in the first 600ms as images are loaded and added to the HostManager, then it does small amounts of work whenever a device finishes an inference. Zooming in a bit:
image
This shows a couple of interesting info about the runtime:
* We add about 400 microseconds latency from the point where we have "loaded" inference results to the point where the HostManager::runFunction callback is triggered. The majority of the time here is in freeing the weights & activation buffers, so pooling could pay off here. At least we could do the free after triggering the callback to the Executor. (OK so this is the CPU device and it's probably not worth optimizing this, but still, interesting).
* There is about a 20-30 microsecond gap where the DeviceManager is not working on an inference, this is the cost of submitting a task to the Executor's threadpool (the lowest green bar). This is actually time cost on the DeviceManager thread while moving the work to an Executor thread, but we're not using the kind of TraceEvents that can move between threads - will look into it.

### DIFF
--- a/examples/resnet-runtime.cpp
+++ b/examples/resnet-runtime.cpp
@@ -38,15 +38,23 @@ llvm::cl::opt<std::string>
                                   "png's with standard imagenet normalization"),
                    llvm::cl::init("../tests/images/imagenet/"),
                    llvm::cl::Positional, llvm::cl::cat(category));
-llvm::cl::opt<unsigned> numDevices("numDevices",
+llvm::cl::opt<unsigned> numDevices("num-devices",
                                    llvm::cl::desc("Number of Devices to use"),
                                    llvm::cl::init(5), llvm::cl::value_desc("N"),
                                    llvm::cl::cat(category));
 llvm::cl::opt<unsigned>
-    maxImages("maxImages",
+    maxImages("max-images",
               llvm::cl::desc("Maximum number of images to load and classify"),
-              llvm::cl::init(1000), llvm::cl::value_desc("N"),
+              llvm::cl::init(100), llvm::cl::value_desc("N"),
               llvm::cl::cat(category));
+llvm::cl::opt<std::string> tracePath("trace-path",
+                                     llvm::cl::desc("Write trace logs to disk"),
+                                     llvm::cl::init(""),
+                                     llvm::cl::cat(category));
+
+std::vector<TraceEvent> allEvents;
+std::mutex eventLock;
+
 } // namespace
 
 /// Loads the model into /p module and returns the input and output
@@ -87,6 +95,13 @@ void dispatchClassify(unsigned int id, HostManager *hostManager,
                 .minMaxArg()
                 .second;
         llvm::outs() << "(" << id << ") " << path << ": " << maxIdx << "\n";
+
+        if (!tracePath.empty()) {
+          std::lock_guard<std::mutex> l(eventLock);
+          auto &newEvents = context->getTraceContext()->getTraceEvents();
+          std::move(newEvents.begin(), newEvents.end(),
+                    std::back_inserter(allEvents));
+        }
 
         if (++returned == maxImages) {
           finished.set_value();
@@ -163,6 +178,8 @@ int main(int argc, char **argv) {
         ImageLayout::NCHW, imagenetNormMean, imagenetNormStd);
     std::unique_ptr<ExecutionContext> context =
         llvm::make_unique<ExecutionContext>();
+    context->setTraceContext(
+        llvm::make_unique<TraceContext>(TraceLevel::STANDARD, 50));
 
     context->getPlaceholderBindings()->allocate(phLists[index]);
     Tensor batch = image.getUnowned(inputShape);
@@ -179,6 +196,11 @@ int main(int argc, char **argv) {
   finished.get_future().wait();
 
   llvm::outs() << "Finished classifying " << started << " images.\n";
+
+  if (!tracePath.empty()) {
+    std::lock_guard<std::mutex> l(eventLock);
+    TraceEvent::dumpTraceEvents(allEvents, tracePath);
+  }
 
   return 0;
 }
--- a/include/glow/Backends/TraceEvents.h
+++ b/include/glow/Backends/TraceEvents.h
@@ -20,6 +20,7 @@
 #include "llvm/ADT/DenseMap.h"
 
 #include <map>
+#include <mutex>
 #include <vector>
 
 namespace glow {
@@ -28,8 +29,13 @@ class PlaceholderBindings;
 
 /// An individual tracing event, such as the begin or end of an instruction.
 /// Designed to match the Google Trace Event Format for Chrome:
-// https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU
+/// https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU
 struct TraceEvent {
+  /// Event Types.
+  static constexpr auto BeginType = "B";
+  static constexpr auto EndType = "E";
+  static constexpr auto InstantType = "I";
+
   /// Human readable name for the item, will be used to match up begin and end.
   std::string name;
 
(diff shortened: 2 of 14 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -113,10 +113,11 @@ void CPUDeviceManager::evictNetworkImpl(std::string functionName,
 void CPUDeviceManager::runFunctionImpl(
     RunIdentifierTy id, std::string function,
     std::unique_ptr<ExecutionContext> context, ResultCBTy resultCB) {
-  context->logTraceEvent("DM_run", "B");
+  std::string eventName = "run_" + function + "_" + std::to_string(id);
+  TRACE_EVENT_BEGIN(context->getTraceContext(), eventName);
   auto funcIt = functions_.find(function);
   if (funcIt == functions_.end()) {
-    context->logTraceEvent("DM_run", "E", {{"reason", "function not found"}});
+    context->logTraceEvent(eventName, "E", {{"reason", "function not found"}});
```
SplitInfinity: Consider making a `TRACE_EVENT_{BEGIN/END}_LOG` macro as well. I know what `"E"` represents because the definition of `TRACE_EVENT_END` is right above it in the list of changes but I don't think it would be as obvious if I looked at this file only.

Consider also a set of `constexpr` strings in the `TraceEvent` namespace for `"B", "E", and "I"` that you can reuse in the `TRACE_EVENT_XXX` macros. Something like this:

```
struct TraceEvent {
...
static constexpr auto BeginType = "B";
static constexpr auto InstantType = "I";
static constexpr auto EndType = "E";
...
};

...

#define TRACE_EVENT_BEGIN(ctx, name)                                           \
  if (ctx) {                                                                   \
    ctx->logTraceEvent(name, TraceEvent::BeginType);                           \
  }

...
```

Then you could write this as `context->logTraceEvent(eventName, TraceEvent::EndType, /*args=*/{{"reason", "function not found"}});`.

include/glow/Backends/TraceEvents.h
```diff
@@ -133,10 +138,34 @@ class TraceContext {
   TraceLevel getTraceLevel() { return traceLevel_; }
   void setTraceLevel(TraceLevel level) { traceLevel_ = level; }
 
-  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "i",
+  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "I",
+                     std::map<std::string, std::string> args = {});
```
SplitInfinity: Is `args` supposed to contain extra information in addition to name and type that should be added to the trace? Consider naming this `attributes`, `attrs`, `extraAttrs` or something like that.

lib/Backends/TraceEvents.cpp
```diff
@@ -64,8 +64,25 @@ void TraceContext::logTraceEvent(llvm::StringRef name, llvm::StringRef type,
   if (traceLevel_ == TraceLevel::NONE || traceLevel_ == TraceLevel::OPERATOR) {
```
SplitInfinity: Can this function be implemented as `logTraceEvent(name, type, TraceEvent::now(), std::move(args)`? Or is getting the timestamp so expensive that it shouldn't be done before making sure the trace level warrants the creation of a `TraceEvent`?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -373,7 +416,8 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
     if ((node->children).empty()) {
       // If the node has no children, propagate its outputs to the result
       // PlaceholderBindings for the run.
-      propagateOutputPlaceholders(executionState, std::move(ctx));
+      propagateOutputPlaceholders(executionState,
```
SplitInfinity: Can we trace this and/or `propagatePlaceholdersForNode`? I'm curious if all this tensor cloning is expensive.

include/glow/Backends/TraceEvents.h
```diff
@@ -133,10 +138,34 @@ class TraceContext {
   TraceLevel getTraceLevel() { return traceLevel_; }
   void setTraceLevel(TraceLevel level) { traceLevel_ = level; }
 
-  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "i",
+  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "I",
```
SplitInfinity: Consider adding a comment to explain what this does.

include/glow/Backends/TraceEvents.h
```diff
@@ -133,10 +138,34 @@ class TraceContext {
...
+                     std::map<std::string, std::string> args = {});
+
+  void logTraceEvent(llvm::StringRef name, llvm::StringRef type,
```
SplitInfinity: Consider adding a comment to explain what this does.

include/glow/Backends/TraceEvents.h
```diff
@@ -133,10 +138,34 @@ class TraceContext {
...
+                     uint64_t timestamp,
                      std::map<std::string, std::string> args = {});
 };
 
+#define TRACE_EVENT_BEGIN(ctx, name)                                           \
```
SplitInfinity: Consider adding comments to explain what these macros are for.

include/glow/Backends/TraceEvents.h
nickgg: Yes, it's custom metadata that is displayed in the trace. "args" is the name given in the spec, but agree it could be more descriptive here.

lib/Backends/TraceEvents.cpp
nickgg: Getting timestamps is relatively cheap, I suppose.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: @SplitInfinity We are tracing handleResult as a block (see the TRACE_EVENT_BEGIN above), do you mean it would be useful to have an additional breakdown for propagateOutputPlaceholders?

nickgg:
Thanks for the review @SplitInfinity, lots to consider :)

lib/Runtime/Executor/ThreadPoolExecutor.cpp
SplitInfinity: Yeah, that's what I mean.

nickgg: Cool I'll add it.

SplitInfinity: More so `propagatePlaceholdersForNode` because it calls `Tensor::clone` a lot.

nickgg:
I added events for propagating inputs and propagating outputs. I think in general we should be a little careful about how many events we add (since they have a small cost) but more info is better than less info while we're learning.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: I've only really examined resnet50 on CPU devices, but copying doesn't seem like a significant cost right now.

examples/resnet-runtime.cpp
```diff
@@ -88,6 +95,13 @@ void dispatchClassify(unsigned int id, HostManager *hostManager,
                 .second;
         llvm::outs() << "(" << id << ") " << path << ": " << maxIdx << "\n";
 
+        if (writeTrace != "") {
```
SplitInfinity: Nit: I personally prefer `writeTrace.empty()` to this.

examples/resnet-runtime.cpp
```diff
@@ -38,15 +38,22 @@ llvm::cl::opt<std::string>
...
                                    llvm::cl::cat(category));
 llvm::cl::opt<unsigned>
-    maxImages("maxImages",
+    maxImages("max-images",
               llvm::cl::desc("Maximum number of images to load and classify"),
-              llvm::cl::init(1000), llvm::cl::value_desc("N"),
+              llvm::cl::init(100), llvm::cl::value_desc("N"),
               llvm::cl::cat(category));
+llvm::cl::opt<std::string>
+    writeTrace("dump-trace", llvm::cl::desc("Write trace logs to disk"),
```
SplitInfinity: This sounds more like a name you'd give to a boolean option. Consider `tracefile-path` or something like that.

examples/resnet-runtime.cpp
```diff
@@ -182,5 +198,10 @@ int main(int argc, char **argv) {
 
   llvm::outs() << "Finished classifying " << started << " images.\n";
 
+  if (writeTrace != "") {
```
SplitInfinity: Same.

include/glow/Backends/TraceEvents.h
```diff
@@ -121,22 +127,68 @@ class TraceContext {
...
+                std::map<std::string, std::string> additionalAttributes = {});
+
+  // Logs a new TraceEvent at the provided \p timestamp, with the given \p name,
+  // \p type and optionally additional attributes.
+  void
+  logTraceEvent(llvm::StringRef name, llvm::StringRef type, uint64_t timestamp,
+                std::map<std::string, std::string> additionalAttributes = {});
 };
 
+/// Macros to make logging trace events easier.
```
SplitInfinity: Consider adding something like `These macros predicate the logging of a `TraceEvent` on a validity of the given `TracingContext`.

PR #2643, opened 2019-04-02
[Benchmarks] Add runtime benchmark framework

SplitInfinity:
**Description**
This commit adds a Glow runtime benchmark that runs the same model using
various components of the runtime stack in order to determine how much
overhead each one contributes to the overall latency.

The benchmark defines a template class named `RuntimeBenchmark` that extends
`benchmark::Fixture`. The template parameter is meant to be a specific
backend class (e.g. `CPUBackend`). This class provides a `virtual` `runBenchmark()` method for
subclasses to override in order to run the benchmarks they represent and
some common code to prepare `Backend`, `Module` and `ExecutionContext`
instances for use in the benchmarks.

There is a subclass of `RuntimeBenchmark` for each component of the inference runtime stack.
Each subclass maintains members for additional state it requires (e.g.
list of `DeviceManagers` in the case of `ExecutorBenchmark`) as well as code
to set up/tear down that state. That code sets everything up correctly
and needs to be overridden to benchmark special cases like device
failure.

There are two macros that are used to create the actual benchmarks.
`DECLARE_RUNTIME_BENCHMARK(name, moduleCreatorFn, dagCreatorFn)` declares
subclasses of `HostManagerBenchmark`, `ExecutorBenchmark` and
`DeviceManagerBenchmark` whose `setUpModule` and `setUpDAG` methods are
overriden using `moduleCreateFn` and `dagCreatorFn`.
`INSTANTIATE_RUNTIME_BENCHMARK(name, backend)` instantiates all three
kinds of benchmarks for the given backend. `moduleCreatorFn` and `dagCreatorFn`
are user-defined functions that create a `Module` and `Executor` `DAG` for
benchmarking, respectively.

**Testing**
All unit tests pass and new benchmarks run successfully (except for
Interpreter backend).

```
$ ./tests/RuntimeBench 
2019-04-02 15:59:22
Running ./tests/RuntimeBench
Run on (8 X 3100 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 8388K (x1)
Load Average: 3.01, 2.20, 2.01
***WARNING*** Library was built as DEBUG. Timings may be affected.
---------------------------------------------------------------------------------------------------------------
Benchmark                                                                     Time             CPU   Iterations
---------------------------------------------------------------------------------------------------------------
SingleNodeHostManagerBenchmark<CPUBackend>/HostManagerCPUBackend           51.6 us         32.7 us        21320
SingleNodeExecutorBenchmark<CPUBackend>/ExecutorCPUBackend                 52.0 us         32.0 us        22563
SingleNodeDeviceManagerBenchmark<CPUBackend>/DeviceManagerCPUBackend       14.8 us         5.29 us       132966
```
**Documentation**
Comments.


SplitInfinity:
I thought it might be useful to explain how to add a new benchmark and benchmark type to elucidate how this is supposed to work.

**Adding A New Benchmark**
1. Define a new module creator function: `std::unique_ptr<Module> myModuleCreator()`
2. Define a new DAG creator function: `std::unique_ptr<DAG> myDAGCreator()`
3. Declare the new benchmark: `DECLARE_RUNTIME_BENCHMARK(myNewBenchmark, myModuleCreator, myDAGCreator);
4. Instantiate the new benchmark for each target backend: `INSTANTIATE_RUNTIME_BENCHMARK(myNewBenchmark, CPUBackend)

If any of the default setup/tear down functions are not acceptable, you will have to create your own subclass of the relevant component benchmark fixture class (`DECLARE_RUNTIME_BENCHMARK` does this for you in the flow above).

**Adding A New Benchmark *Type***
1. Define a new subclass of `RuntimeBenchmark` for your component (e..g `ONNXIfiManager`). Define `SetUp` and `TearDown` appropriately.
2. Modify the `DECLARE...` and `INSTANTIATE...` macros to also declare and instantiate benchmarks for this new component.

nickgg:
This looks good! Will really simplify adding new benchmarks.

tests/benchmark/RuntimeBench.cpp
```diff
@@ -0,0 +1,612 @@
...
+void setUpDeviceManagerCommon(
+    benchmark::State &state, std::unique_ptr<Backend> &backend,
+    std::unique_ptr<Module> &mod, std::unique_ptr<DeviceManager> &deviceManager,
+    std::unordered_map<std::string, std::unique_ptr<CompiledFunction>>
+        &deviceManagerFunctions) {
+
+  // Check that the backend is valid.
+  if (!backend) {
+    state.SkipWithError(
+        "Unable to set up device manager - backend not set up!");
```
nickgg: nit: for this and other errors: we tend to use DeviceManager over device manager.

PR #2644, opened 2019-04-03
Added openCLDeviceBindings and moved OpenCLCompiledFunction to use them.

gcatron:
*Description*: Added support for DeviceBindings to the OpenCL implementation and migrated to use them for memory buffers and CL device information. This moves allocation out of the execute method. The compiledFunction is still not thread safe, it currently uses member variables to pass state it will require this to change #2651 .
*Testing*: ninja test
*Documentation*: NA

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -54,12 +54,6 @@ typedef uint32_t cl_size_t;
 /// compile these sources into our the OpenCL backend. During the execution
 /// these sources are compiled by the OpenCL driver.
 
-// This defines kernels for most operations.
-static const unsigned char kernels_cl_src[] = {
-#include "glow/kernels.inc"
-};
-static const size_t kernels_cl_src_size = sizeof(kernels_cl_src);
-
 // This defines kernels for optimized convolutions.
 static const unsigned char kernels_fwd_conv_cl_src[] = {
 #include "glow/kernels_fwd_conv.inc"
@@ -106,13 +100,6 @@ static void dumpCompileLog(cl_device_id dev, cl_program prog) {
 #endif
 }
 
-/// Add an macro definition with an integer value to the set of options.
-template <typename T>
-static void addIntOption(std::vector<std::string> &options,
-                         const std::string &name, const T value) {
-  options.push_back("-D" + name + "=" + std::to_string(value));
-}
-
 /// Add an macro definition with a string value to the set of options.
 static void addStringOption(std::vector<std::string> &options,
                             const std::string &name, const std::string &value) {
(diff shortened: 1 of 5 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -35,20 +35,39 @@ namespace runtime {
...
   cl_mem buffer_;
+
   /// Count of functions using this buffer.
   unsigned int users_{0};
+
   /// Size of the buffer in bytes.
   const size_t size_{0};
 
+  /// CL command queue for the network.
+  cl_command_queue commands_;
```
opti-mix: Why do you associate a `cl_command_queue` with `OpenCLBuffer`? It is kind of strange. In theory, the same buffer could be used by different command queues.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -176,21 +171,38 @@ void OpenCLDeviceManager::addNetworkImpl(const Module *module,
     return;
   }
 
+  // Create a shared command queue using traceInfo.
```
opti-mix: The comment is a bit misleading. The fact that is uses traceInfo is not very important, as it is optional.

It would be more interesting to describe who is sharing this command queue? Different runs of the same network? Different networks?

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -215,6 +215,23 @@ class OCLBackend final : public BackendUsingGlowIR {
   /// @}
 };
 
+namespace runtime {
+struct OpenCLDeviceBindings : DeviceBindings {
```
opti-mix: Doxygen comment, please!

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -215,6 +215,23 @@ class OCLBackend final : public BackendUsingGlowIR {
...
+  OpenCLDeviceBindings() : DeviceBindings(BackendKind::OpenCL) {}
+
+  /// CL memory buffer.
```
opti-mix: It could be interesting to describe which of these fields are run-specific and which are network specific.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -607,13 +555,25 @@ static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
...
-    auto ev = context->scopedEvent("alloc");
-    // TODO: deviceBuffer_ should go into DeviceBindings in context.
-    deviceBuffer_ = allocDeviceBuffer(runtimeBundle_.getConstantWeightSize() +
-                                      runtimeBundle_.getMutableWeightSize() +
-                                      runtimeBundle_.getActivationsSize());
-  }
+  auto clBindings = static_cast<runtime::OpenCLDeviceBindings *>(
+      context->getDeviceBindings());
+
+  deviceBuffer_ = clBindings->deviceBuffer;
```
opti-mix: Do you assign to the fields of `OpenCLFunction` here? This is not thread-safe, or? We may run the same `OpenCLFunction` on different inputs concurrently. 

Shouldn't these fields be removed from `OpenCLFunction` now that they are in the `OpenCLDeviceBindings`?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -607,13 +555,25 @@ static void topK(Tensor &outW, Tensor &indW, Tensor &inW, size_t k) {
...
+  // Configure the kernels by providing the size of size_t on the host size.
+  // This is required to e.g. properly pass struct parameters of types like
+  // ShapeNHWC, ShapeNCHW, etc. The definitions of these types on the host side
+  // use size_t for their members and they should be defined on the OpenCL's
+  // side using integer types of the same width.
+  addIntOption(options, "SIZEOF_HOST_SIZE_T", sizeof(size_t));
+  // Create the program from the source.
+  std::string source(reinterpret_cast<const char *>(kernels_cl_src),
+                     kernels_cl_src_size);
+  createProgram(source, options, commands_);
```
opti-mix: This code would "compile" the program using OpenCL compiler on the first call of `execute`. Shouldn't it be done when the network is added to the device manager?

lib/Backends/OpenCL/OpenCLDeviceManager.h
gcatron: I think a rename might be good here, the idea being there is one command queue for all functions that share a set of common weights since no more than one can run at a time. Alternatively I could create a command queue for each function.

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -35,20 +35,39 @@ namespace runtime {
...
+  OpenCLBuffer(cl_mem buffer, size_t size) : buffer_(buffer), size_(size) {}
+
   /// Returns the stored buffer.
   cl_mem getBuffer() { return buffer_; }
+
+  /// Returns the command queue.
+  cl_command_queue getCommandQueue() { return commands_; }
+
+  /// Sets the command queue.
+  void setCommandQueue(cl_command_queue command) { commands_ = command; }
```
bertmaher: I'm increasingly convinced setters are a bad idea.  Is it valid to have an OpenCLBuffer without a command queue?  If not, put it in the constructor.  If it is valid, why?

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -215,6 +215,23 @@ class OCLBackend final : public BackendUsingGlowIR {
   /// @}
 };
 
+namespace runtime {
+struct OpenCLDeviceBindings : DeviceBindings {
+  OpenCLDeviceBindings() : DeviceBindings(BackendKind::OpenCL) {}
```
bertmaher: An OpenCLDeviceBindings needs to have all the fields below initialized, right?  If so make them part of the ctor.

lib/Backends/OpenCL/OpenCL.cpp
gcatron: That is the next step. This PR introduces the bindings and moves allocation and setup out of the function. The next PR would be a refactor passing the bindings around to the functions that currently rely on the member variables.

PR #2646, opened 2019-04-03
[Partitioner] Fix a bug a remove duplicated children.

beicy:
*Description*:

*Testing*:
Current tests.

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -468,6 +468,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
     }
 
     // Link subF to its parents.
+    std::set<Function *> parents;
     for (auto &N : subF->getNodes()) {
       for (int inp = 0, e = N.getNumInputs(); inp < e; inp++) {
         auto input = N.getNthInput(inp);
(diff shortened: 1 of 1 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -468,6 +468,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
     }
 
     // Link subF to its parents.
+    std::set<Function *> parents;
```
jfix71: Assuming `parents` usually has a small number of Functions, you could consider using `llvm::SmallSet` here instead.

rdzhabarov: how hard would be to exercise buggy behavior in a unit test?

beicy: Actually, it won't show errors in the release mode. This may trigger the assertion that a placeholder is already registered.

rdzhabarov: Can we add a test for that? (debug mode failure would be enough)

beicy: I had one. Will added it later :)

nadavrot:
Man, could you please add testcases to pull requests that fix bugs?

beicy:
> Man, could you please add testcases to pull requests that fix bugs?

Yes. Will add it later.

PR #2649, opened 2019-04-03
Commonalize DeviceManagerConfig and DeviceConfig

nickgg:
*Description*: We currently have two structs for configuring devices: DeviceManagerConfig (used by the HostManager) and DeviceConfig (used by the DeviceManager). This is unnecessary and confusing, so removing DeviceManagerConfig and plumbing DeviceConfig all the way through.

Also, added a `getBackendName()` to get a string name for a device which can be used for logging and, in the next PR, used to generate the trace log.
*Testing*: unit tests, asan, etc
*Documentation*: This is a breaking change for Backends which need to implement the `getBackendName()` method.

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -103,12 +103,18 @@ using DAGListTy = std::vector<DAG>;
 /// member variable to it's correct BackendKind.
 class DeviceConfig {
   const BackendKind backendKind_;
-
-protected:
-  DeviceConfig(BackendKind kind) : backendKind_(kind) {}
+  std::string name_;
 
 public:
+  DeviceConfig(BackendKind kind) : backendKind_(kind) {}
+  DeviceConfig(BackendKind kind, std::string name)
+      : backendKind_(kind), name_(name) {}
+
   BackendKind getBackendKind() { return backendKind_; }
+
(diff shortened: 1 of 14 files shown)


include/glow/Runtime/RuntimeTypes.h
```diff
@@ -103,12 +103,18 @@ using DAGListTy = std::vector<DAG>;
...
   DeviceConfig(BackendKind kind) : backendKind_(kind) {}
+  DeviceConfig(BackendKind kind, std::string name)
+      : backendKind_(kind), name_(name) {}
 
-public:
   BackendKind getBackendKind() { return backendKind_; }
+
+  llvm::StringRef getName() { return name_; }
+  bool hasName() { return name_ != ""; }
+  void setName(llvm::StringRef name) { name_ = name; }
```
beicy: nit: maybe another function to compare the names?

nickgg: @beicy what do you suggest?

beicy: I thought we may need to compare if two DeviceConfig have the same backend name. But on second thought, we don't really need to do it, right?

nickgg: I don't think so, the BackendName is really just for logging anyway.

PR #2654, opened 2019-04-04
[habana] Add pipelining; improve linking

bertmaher:
*Description*:
We'd like to upstream a couple of significant improvements that we've made internally at FB:
1. Improvements to HabanaDeviceManager that enable multi-device execution and pipelining, by converting the synchronous execute/wait into an async operation;
2. Linking improvements; since the Habana libraries no longer expose dependences that conflict with Glow (e.g., libprotobuf), we can go back to normal linking rather than a round-about dlopen approach.

(Note: I'll try to upstream patches more gradually in the future.  We're just working out the whole open-source-backend process, and while that's being worked out I don't want to let our trees get hopelessly out of sync)

*Testing*: `ctest -R Habana.*Test`

*Documentation*: n/a


bertmaher:
By the way -- while the overall PR is sadly too large, the individual commits should be relatively reviewable.  I think everyone who's written these would welcome additional code review, so please feel free to give this a real look-through.

lib/Backends/Habana/Habana.h
```diff
@@ -23,16 +23,177 @@
...
+  /// HabanaIOBuffer instances, so the real copy is kept here and all HabanaIOBuffer
+  /// instances in the pool receive const references to this one.
+  std::unordered_map<const Placeholder *, off_t> offsets_;
+  /// The size of all input and output Placeholders provided during
+  /// construction. This is the effective size of one HabanaIOBuffer in this pool.
+  size_t size_;
+  /// The combined size of all HabanaIOBuffers in this pool (i.e. size_ *
+  /// numBuffers_).
+  size_t totalSize_;
+  /// Buffer that backs all of the HOmaIOBuffers in this pool. The first buffer
```
jfix71: nit: `HOmaIOBuffers` -> `HabanaIOBuffers`

bertmaher:
Yeah, so I think Mul in particular I literally copied directly from OperatorTest :-/.

But right, HabanaBackendTest has a lot of overlap with OperatorTest.  There are a few reasons why I created a new test suite:
1. Before `ENABLED_BACKENDS` landed it was kind of a pain to select the subset of tests for operators we cared about.  There's gtest_filter, but that gets old too.  It's quick and easy to just run HabanaBackendTest and not worry about it.
2. Since we were bringing operators up incrementally it was useful to add tests that hit each particular feature as we got to them.  There are a lot of operator tests and it's easier to write my own 32x32 FC than it is to find an existing one (that has the right types, and no extra ops, etc...)
3. While learning about how the HW and SW were expected to work it was useful to play around with small tests anyways.  E.g., I'd write something that does a dot-product that I can compute by hand to make sure I know how scale/bias work.  Some of the OperatorTests are like this, but for something like FC there are a bunch of more complicated tests and it's hard to know what about my implementation was wrong :-)

Whew, sorry, I didn't mean to write a dissertation here.  I do think there would be value in trying our "core" tests on Habana too.  And maybe either merging HabanaBackendTest into OperatorTest or killing it outright.

PR #2657, opened 2019-04-04
Image Classifier changes to improve misclassification reporting.

putivsky:
*Description*: Made changes in Image Classifier utility to provide an explicit expected index for matching ML category through optional command line flag
*Testing*:
Run shell file and check printed output, changing expected category index into invalid one, prints mismatch and exits shell script
*Documentation*:
New command line flag has been added, i.e. "idx" - expected matching top index for ML category
Modified script adding expected ML category indices


tools/loader/ImageClassifier.cpp
```diff
@@ -267,22 +291,24 @@ static void processAndPrintResultsImpl(Tensor *SMT,
...
     printTopKPairs(topKPairs);
+    retVal += matchIndices(topKPairs, fileName);
   }
+
+  return retVal;
 }
 
 /// Given the output Softmax Tensor \p SMT and \p functionName, switch between
 /// the correct element type to print the results of inference as contained in
 /// \p SMT.
```
jfix71: Please update doxygen with `\returns`.

tools/loader/ImageClassifier.cpp
```diff
@@ -232,11 +254,11 @@ template <typename ElemTy> static void applySoftmax(Handle<ElemTy> H) {
   }
 }
 
-/// Given the output Softmax Tensor \p SMT and \p functionName, print the
-/// results of inference.
+/// Given the output Softmax Tensor \p SMT and \p functionName, prints the
+/// results of inference and returns number of incorrect predictions
```
jfix71: nit: Add backslash to returns for Doxygen style (i.e. `\returns`).

tools/loader/ImageClassifier.cpp
```diff
@@ -77,6 +77,10 @@ llvm::cl::opt<bool> convertInAndOutToFp16(
     llvm::cl::desc(
         "Convert the input and output tensors of the network to fp16"),
     llvm::cl::cat(imageLoaderCat));
+llvm::cl::opt<unsigned> expectedCategoryIndex(
```
jfix71: This allows for only a single index to be specified, even though for the Caffe2 models we allow multiple inputs. Perhaps this could instead be a comma separated list of expected categories, in the same order as the input images order?

tools/loader/ImageClassifier.cpp
```diff
@@ -389,5 +416,5 @@ int main(int argc, char **argv) {
     loader.generateAndSerializeQuantizationInfos(bindings);
   }
 
-  return 0;
+  return numErrors;
```
jfix71: Specifying the correct index is optional, but if it's not specified then the program will exit with a non-zero return (assuming the input is not classified as `0`). If we move `expectedCategoryIndex` to a list then perhaps we can only return `numErrors` if the list is empty.

tools/loader/ImageClassifier.cpp
```diff
@@ -218,6 +222,24 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
   }
 }
 
+/// Checks if /p topKPairs have the index that matches the provided index
+static int matchIndices(const std::vector<FloatIndexPair> &topKPairs,
+                        const std::string &fileName) {
+  // loop through pairs and try to match the index from file name
+  for (const auto& p : topKPairs) {
+    if (p.second - labelOffset == expectedCategoryIndex) {
```
nickgg: So it looks like if you don't provide the expectedCategoryIndex argument this will mostly fail (since it defaults to index 0). Can we make it instead so that it always passes if you don't provide the arg, that way default behavior of image-classifier is unchanged.

tests/images/run.sh
```diff
@@ -13,47 +13,45 @@
...
+
+imagenet299Files=("cat_281_299" "dog_207_299" "zebra_340_299")
+googlenetV1IdxValue=(281 222 340)
+googlenetV4IdxValue=(281 207 340)
+
+mnistFiles=("0_1009" "1_1008" "2_1065" "3_1020" "4_1059" "5_1087" "6_1099" "7_1055" "8_1026" "9_1088")
+mnistIdxValue=(0 1 2 3 4 5 6 7 8 9)
+
+for i in ${!imagenetFiles[@]}; do
+	./bin/image-classifier tests/images/imagenet/${imagenetFiles[$i]}.png -idx=${imagenetIdxValue[$i]} -use-imagenet-normalization -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data "$@"
```
nickgg: So the image-classifier returns non-zero in the case of error, can we make this script also return non-zero if any run fails, that way we can use it as part of the CI system.

tools/loader/ImageClassifier.cpp
nickgg: Looks like this is what is breaking the CI runs for this PR.

tools/loader/ImageClassifier.cpp
```diff
@@ -218,6 +222,24 @@ static void printTopKPairs(const std::vector<FloatIndexPair> &topKPairs) {
   }
 }
 
+/// Checks if /p topKPairs have the index that matches the provided index
+static int matchIndices(const std::vector<FloatIndexPair> &topKPairs,
+                        const std::string &fileName) {
+  // loop through pairs and try to match the index from file name
```
nadavrot: Period at the end of the sentence.

tests/images/run.sh
putivsky: script has instruction:
set -e
it means whatever environment runs script will see script exits on error with the correspondent error code.

nickgg: Yes, but `set -e` early exits. Similar to the unit tests, it would be better to get the full list of failures in one run so they can be fixed in a single commit.

tools/loader/ImageClassifier.cpp
```diff
@@ -279,6 +284,25 @@ static int processAndPrintResultsImpl(Tensor *SMT,
   const size_t numLabels = SMT->dims()[1];
   std::vector<size_t> sliceOffset(SMT->dims().size(), 0);
 
+  // Parse provided, if any, category indices.
+  std::vector<signed> matchingIndices(inputImageFilenames.size(), -1);
+  if (!expectedMatchingIndices.empty()) {
+		llvm::SmallVector<llvm::StringRef, 1> parts;
+    llvm::StringRef(expectedMatchingIndices).split(parts, ',', -1, false);
```
nadavrot: Can you please add a comment that describes what this line does?

tools/loader/ImageClassifier.cpp
```diff
@@ -279,6 +284,25 @@ static int processAndPrintResultsImpl(Tensor *SMT,
...
+    if (parts.size() != inputImageFilenames.size()) {
+      llvm::outs() << "Number of matching indices: " << parts.size()
+                   << " doesn't match the number of files: "
+                   << inputImageFilenames.size() << "\n";
+    } else {
+      for (unsigned i = 0; i < parts.size(); i++) {
```
nadavrot: Comment please.

tools/loader/ImageClassifier.cpp
```diff
@@ -279,6 +284,25 @@ static int processAndPrintResultsImpl(Tensor *SMT,
...
+        if (parts[i].getAsInteger(10, matchingIndices[i])) { // true means error
```
nadavrot: Please move the comment above the line and punctuate it. 

https://releases.llvm.org/7.0.0/docs/CodingStandards.html#comment-formatting

tools/loader/ImageClassifier.cpp
```diff
@@ -77,6 +77,12 @@ llvm::cl::opt<bool> convertInAndOutToFp16(
     llvm::cl::desc(
         "Convert the input and output tensors of the network to fp16"),
     llvm::cl::cat(imageLoaderCat));
+
+llvm::cl::opt<std::string> expectedMatchingIndices(
```
jfix71: Instead of using a string, you could just make this a `llvm::cl::list<int>` and then specify it as `llvm::cl::CommaSeparated`. Then you wouldn't need to split the string or convert to int.

tools/loader/ImageClassifier.cpp
```diff
@@ -287,15 +287,20 @@ static int processAndPrintResultsImpl(Tensor *SMT,
   // Parse provided, if any, category indices.
   std::vector<signed> matchingIndices(inputImageFilenames.size(), -1);
   if (!expectedMatchingIndices.empty()) {
+    // If category indices list is not empty it must be in n,m,k,...,z format.
     llvm::SmallVector<llvm::StringRef, 1> parts;
```
nadavrot: Period at the end of the sentence.

tools/loader/ImageClassifier.cpp
putivsky: Thanks Jordan for great tip!

tools/loader/ImageClassifier.cpp
putivsky: Fixed in last commit

tests/images/run.sh
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

tools/loader/ImageClassifier.cpp
putivsky: Fixed in the last commit

putivsky:
Addressed all comments, please take a look and advice the next round of fixes.

tests/images/run.sh
```diff
@@ -14,46 +14,81 @@
...
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=neg128to127 -m=vgg19 -model-input-name=data "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=neg128to127 -m=squeezenet -model-input-name=data "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to255 -m=zfnet512 -model-input-name=gpu_0/data "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=densenet121 -model-input-name=data -compute-softmax "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=shufflenet -model-input-name=gpu_0/data "$@"
-./bin/image-classifier tests/images/mnist/*.png -image-mode=0to1 -m=lenet_mnist -model-input-name=data "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to255 -m=inception_v1 -model-input-name=data "$@"
-./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to255 -m=bvlc_alexnet -model-input-name=data "$@"
-for png_filename in tests/images/imagenet/*.png; do
-  ./bin/image-classifier "$png_filename" -use-imagenet-normalization -image-mode=0to1 -m=resnet50/model.onnx -model-input-name=gpu_0/data_0 "$@"
```
jfix71: It looks like all of these ONNX models were removed?

tools/loader/ImageClassifier.cpp
```diff
@@ -253,8 +284,25 @@ static void processAndPrintResultsImpl(Tensor *SMT,
   const size_t numLabels = SMT->dims()[1];
   std::vector<size_t> sliceOffset(SMT->dims().size(), 0);
 
+  // Parse provided, if any, category indices.
+  std::vector<int> matchingIndices(inputImageFilenames.size(), -1);
+  if (!expectedMatchingIndices.empty()) {
+    // The number of category indices must match the number of files.
+    if (expectedMatchingIndices.size() != inputImageFilenames.size()) {
```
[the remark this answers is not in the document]
putivsky: 1. no error/warning when compiled, but I can switch to unsigned.
2. agreed, if indices have been provided they must match the number of files.

tools/loader/ImageClassifier.cpp
```diff
@@ -77,6 +77,12 @@ llvm::cl::opt<bool> convertInAndOutToFp16(
     llvm::cl::desc(
         "Convert the input and output tensors of the network to fp16"),
     llvm::cl::cat(imageLoaderCat));
+
+llvm::cl::list<int> expectedMatchingIndices(
+    "idxs",
```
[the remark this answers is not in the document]
putivsky: ok

tests/images/run.sh
putivsky: checking...

putivsky:
Fixes upon Jordan's feedback.

putivsky:
> Looks good generally. Can you squash your commits down to a single one?

And how can I do it from UI or command line? Any link with examples would be helpful.

nickgg:
@putivsky: basically you want to do an interactive rebase (git rebase -i HEAD~<number of commits>) and squash down all but one. Here's a pretty decent guide: https://github.com/wprig/wprig/wiki/How-to-squash-commits.

nickgg:
You can auto fix the clang-formatting issues automatically by running:

`./utils/format.sh fix`

jfix71:
@putivsky [Gitup](https://gitup.co/) has a simple UI for managing multiple branches/commits. You can swap the order of commits, squash commits together, cherry pick across branches, etc.

PR #2658, opened 2019-04-04
[Partitioner] Create the intermediate ph based on the NodeValue in case that one node can has more than 1 output.

beicy:
*Description*:
When the Partitioner was created, I assumed one Node only has one output when creating the intermediate ph,  which is wrong.  Eg. TopK has 2 outputs and we need to know which one should be used when creating the intermediate ph during partitioning. This PR fixed this part.

*Testing*:
Existing tests. 
I wrote a runtime e2e test using En2gr model, which is used to test this PR as well. The testcase will be added later. 

* Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -456,7 +456,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   // For any dependency that crosses a partition, add a placeholder and save
   // node. Record the dependence in the function graph.
   int logicalID = 0;
-  llvm::DenseMap<Node *, Placeholder *> placeholders;
+  std::unordered_map<NodeValue, Placeholder *> placeholders;
   llvm::DenseMap<Function *, DAGNode *> funcDAG;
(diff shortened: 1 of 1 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -456,7 +456,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   // For any dependency that crosses a partition, add a placeholder and save
   // node. Record the dependence in the function graph.
   int logicalID = 0;
-  llvm::DenseMap<Node *, Placeholder *> placeholders;
+  std::unordered_map<NodeValue, Placeholder *> placeholders;
```
jfix71: Curious why the switch to `unordered_map` instead of `DenseMap`

beicy: llvm::DenseMap doesn't allow me to use NodeValue as a key.  The error message is like error: no member named 'getEmptyKey' in 'llvm::DenseMapInfo<glow::NodeValue>'. Is still looking for a proper explanation to this error message.

PR #2663, opened 2019-04-05
Share HostManagers in onnxifi BackendIds

jackm321:
*Description*:
This is important for avoiding static destruction order fiasco between DeviceManagers owned by HostManager and device api. It also ensures that HostManagers don't outlive their usefulness and are destroyed when onnxifi destroys all BackendIds.

*Testing*:
synced to internal, all tests pass

*Documentation*:
updated comments

### DIFF
--- a/lib/Onnxifi/GlowOnnxifiManager.cpp
+++ b/lib/Onnxifi/GlowOnnxifiManager.cpp
@@ -33,7 +33,7 @@ BackendIdPtr GlowOnnxifiManager::createBackendId(glow::BackendKind kind,
   if (forQuantization) {
     backendId = new onnxifi::BackendId(kind, useOnnx);
   } else {
-    auto *hostManager = getOrCreateHostManager(kind);
+    auto hostManager = getOrCreateHostManager(kind);
     backendId = new onnxifi::HostManagerBackendId(hostManager, kind, useOnnx);
   }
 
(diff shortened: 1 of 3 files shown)


lib/Onnxifi/GlowOnnxifiManager.h
```diff
@@ -116,8 +117,11 @@ class GlowOnnxifiManager final {
   /// The set of all valid glow Graphs.
   std::unordered_set<GraphPtr> graphs_;
 
-  /// Map from BackendKind to HostManager managing devices of that kind.
-  std::map<BackendKind, std::unique_ptr<runtime::HostManager>> hostManagers_;
+  /// Map from BackendKind to HostManager managing devices of that kind that is
+  /// shared by all BackendIds using that HostManager. HostManager is stored as
+  /// weak_ptr here so that it will be destructed when the last BackendId using
+  /// it is destroyed not when this singleton is destroyed.
```
nickgg: This only comes into effect if the GlowOnnxifiManager outlives all Backends, is that what we expect?

lib/Onnxifi/GlowOnnxifiManager.cpp
```diff
@@ -93,21 +93,23 @@ GlowOnnxifiManager::createGraph(BackendPtr backend,
...
-  // If a HostManager doesn't exist yet for this BackendKind then create one
-  if (it == hostManagers_.end()) {
-    auto newHostManager =
-        onnxifi::HostManagerBackendId::createHostManager(backendKind);
-    assert(newHostManager != nullptr);
-    auto ptr = newHostManager.get();
-    hostManagers_[backendKind] = std::move(newHostManager);
-    return ptr;
+  if (it != hostManagers_.end()) {
+    hostManager = it->second.lock();
```
nickgg: If this is null (because the object has been destroyed) we should probably remove it from the `hostManagers_` map.

lib/Onnxifi/GlowOnnxifiManager.h
jackm321: It always will, it's a static singleton

lib/Onnxifi/GlowOnnxifiManager.cpp
jackm321: There's not much point, if it's null it will be replaced in the map on line 109

jackm321:
@nickgg or anyone else have any comments?

PR #2664, opened 2019-04-05
Added symbolCategory to symbol info in the runtimeBundle.

### LINKED ISSUE #2597: RuntimeSymbolnfo should include the kind of a symbol, e.g. if it is a constant, an activation or a mutable weight
It would be very convenient if `RuntimeSymbolnfo` would include a kind of a symbol, e.g. if it is a constant, an activation or a mutable weight.

Among other things, this information would complement the `offset` field. `offset` describes the offset of the symbol in its respective memory area. But it is not obvious from the current `RuntimeSymbolnfo` which memory area the belongs to. And the kind field would provide exactly this information.

BTW, the name `kind` is a strawman attempt to name it. A more semantically relevant name would be better.

gcatron:
*Description*: Adds a new field: symbolCategory to the symbolInfo object. This denotes whether the symbol is a Placeholder, constant, or activation.
*Testing*: Added new unit test.
*Documentation*:
Fixes #2597

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -21,6 +21,11 @@
 
 namespace glow {
 namespace runtime {
+
+/// An enum to indicate what type each symbol in the bundle is. Current options
+/// are activations, constants, and placeholders.
+enum class SymbolCategory { Activation, Placeholder, Constant };
+
 /// Contains information for initialization and handling of symbol at runtime.
 struct RuntimeSymbolInfo {
   /// The size in bytes.
@@ -33,6 +38,8 @@ struct RuntimeSymbolInfo {
   bool input{true};
   /// Is the symbol an output for the function.
   bool output{true};
+  /// Indicates what category the symbol is.
+  SymbolCategory symbolCategory;
 };
 
 using SymbolTableTy = std::unordered_map<std::string, RuntimeSymbolInfo>;
--- a/tests/unittests/BackendTest.cpp
+++ b/tests/unittests/BackendTest.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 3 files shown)


include/glow/Backends/BackendUtils.h
```diff
@@ -21,6 +21,11 @@
 
 namespace glow {
 namespace runtime {
+
+/// An enum to indicate what type each symbol in the bundle is. Current options
+/// are activations, constants, tensor views, and placeholders.
+enum class SymbolCategory { Activation, Placeholder, Constant, TensorView };
```
opti-mix: I'm not sure we need a dedicated category for TensorViews. After all they are derivatives of their underlying symbols. It seems like tensorview symbols should get the same category as their parent symbols.

tests/unittests/BackendTest.cpp
```diff
@@ -246,6 +247,28 @@ TEST_P(BackendTest, BundleSharedConstant) {
...
+
+  std::unique_ptr<Backend> backend(createBackend(GetParam()));
+  auto function = backend->compile(F);
+  auto table1 = function->getRuntimeBundle().getSymbolTable();
+  /// Make sure X is in both tables.
+  auto it = table1.find(X->getName());
+  EXPECT_EQ(it->second.symbolCategory, glow::runtime::SymbolCategory::Constant);
+  it = table1.find(save->getName());
+  EXPECT_EQ(it->second.symbolCategory,
+            glow::runtime::SymbolCategory::Placeholder);
```
opti-mix: May be add some tests for activations and tensorviews as well? In particular, there should be tests for checking that tensorviews have the same category as their underlying tensors.

jfix71:
Nice! Does this change enable better checking for the tests from https://github.com/pytorch/glow/pull/2616, as we previously talked about?

PR #2665, opened 2019-04-05
[NFC] Use llvm-link from the given LLVM build directory

artemrakhov:
This helps to eliminate mismatch between given LLVM's linker and system's standard llvm-link


opti-mix:
@compnerd Any objections? I think you introduced this code originally. Do you see any reasons why this new code would be inferior to the old one?

compnerd:
I think it would be better to use the `LLVMConfig.cmake` for this (which we already load via the `find_package(LLVM CONFIG REQUIRED)`.

compnerd:
IIRC, we want to use `LLVM_OUTPUT_INTDIR` as a `HINT` to the `find_program`.  By preserving the `find_program` we can permit people to build glow against a system provided LLVM (particularly common on Linux).

PR #2667, opened 2019-04-05
Add ConvRelu shape/input info in Caffe2ModelLoader.cpp

zrphercule:
*Description*:
The ConvRelu is same to Conv or Int8Conv, so just use it directly.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -237,7 +237,7 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
   }
   const std::string &opName = loadOperatorName(op);
 
-  if (typeName == "Conv" || typeName == "Int8Conv" ||
+  if (typeName == "Conv" || typeName == "ConvRelu" || typeName == "Int8Conv" ||
       typeName == "Int8ConvRelu") {
     // Load the inputs:
     std::vector<unsigned_t> strides;
@@ -267,7 +267,7 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     // Caffe2 "Conv" op always stores the weight as CKRS, while for "Int8Conv",
     // and "Int8ConvRelu", the weights always follows the "order" arg.
     Tensor wtag;
-    if (typeName != "Conv" && order == "NHWC") {
+    if ((typeName != "ConvRelu" && typeName != "Conv") && order == "NHWC") {
       wtag.assign(w);
     } else {
       w->transpose(&wtag, NCHW2NHWC);
@@ -298,7 +298,7 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     TypeRef outTy;
     Constant *filter;
     Constant *bias;
-    if (typeName == "Conv") {
+    if (typeName == "Conv" || typeName == "ConvRelu") {
       // Construct the Bias field.
       Tensor biasTensor(ElemKind::FloatTy, {depth});
       biasTensor.zero();
@@ -355,7 +355,9 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     Node *node = G_.createConv(opName, finalIn, filter, bias, outTy, kernels,
                                strides, pads, group);
-
+    if (typeName == "ConvRelu") {
+      node = G_.createRELU(opName + ".relu", node);
+    }
     if (order == "NCHW") {
       // Transpose the output back.
       node = G_.createTranspose(opName, node, NHWC2NCHW);
--- a/tests/unittests/Caffe2ImporterTest.cpp
+++ b/tests/unittests/Caffe2ImporterTest.cpp
@@ -67,6 +67,43 @@ TEST(caffe2, importConv) {
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
 }
 
+TEST(caffe2, importConvRelu) {
+  ExecutionEngine EE{BackendKind::Interpreter};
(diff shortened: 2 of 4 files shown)


beicy:
In this case, ConvRelu will be treated as "Conv". I think we should split it into "conv" + "Rule" ?

zrphercule:
> In this case, ConvRelu will be treated as "Conv". I think we should split it into "conv" + "Rule" ?

But I noticed Int8ConvRelu is also here already, and it is correct

beicy:
> > In this case, ConvRelu will be treated as "Conv". I think we should split it into "conv" + "Rule" ?
> 
> But I noticed Int8ConvRelu is also here already, and it is correct

Right. But I mean under this switch, we have different flows to handle different op. There are a few branches " if (typeName == "Conv")" or " if (typeName != "Conv") ".

jfix71:
From what I can see we aren't actually creating a relu for `Int8ConvRelu `, no? This seems broken if so.

beicy:
> From what I can see we aren't actually creating a relu for `Int8ConvRelu `, no? This seems broken if so.

No. After some discussion, we found for us, adding "Int8Relu" here is kind of useless.

rdzhabarov:
As we've already have proper {scale,offset} from the the ConvRelu node (covering non negative range).
Adding additional ReLU seems to be redundant as proper range is already covered by setting proper quantization params for Conv.

rdzhabarov:
Please, run clang-format @zrphercule

jfix71:
>As we've already have proper {scale,offset} from the the ConvRelu node (covering non negative range). Adding additional ReLU seems to be redundant as proper range is already covered by setting proper quantization params for Conv.

Ok, that seems like something we should at least note in the importer logic here. I think adding an assert or check that the quantized range is only positive would be a good idea.

rdzhabarov:
> I think adding an assert or check that the quantized range is only positive would be a good idea.

Totally.

rdzhabarov:
@zrphercule good to go after addressing Jordan's concern.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -237,7 +237,7 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
   }
   const std::string &opName = loadOperatorName(op);
 
-  if (typeName == "Conv" || typeName == "Int8Conv" ||
+  if (typeName == "Conv" || typeName == "ConvRelu" || typeName == "Int8Conv" ||
```
nadavrot: @zrphercule Is this correct? Where are you constructing the Relu in the code below?

jfix71: Yeah, I think we need to create a ReLU for the float version. It only makes sense to skip the ReLU for the quantized version assuming that the Conv has non-negative quantization range.

zrphercule:
@rdzhabarov I wonder if this part is only for shape info, or it is for building the graph? If it includes build the graph, then I guess we cannot simply change it this way, since we cannot ignore the Relu in a normal ConvRelu.

zrphercule:
Hi all, I would like to add a Relu node here, so no need for more review for now. Will tag you guys once I finished hahah! Thanks!

rdzhabarov:
>  I wonder if this part is only for shape info, or it is for building the graph?

This is actual graph building.

I thought originally that this is for quantized ConvRelu, but appears to be it's actually the other way around. It's only for fp32. So, as we discussed let's verify that we operate on fp32 and do immediate lowering (Conv followed by Relu) in the loader.

Alternative solution would be to unfuse that operator on the ONNXIFI level which can be done at a later point. Loading ConvRelu by standalone Glow seems useful as well.

zrphercule:
@rdzhabarov Wonder if this is what we need?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -355,7 +355,10 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
 
     Node *node = G_.createConv(opName, finalIn, filter, bias, outTy, kernels,
                                strides, pads, group);
-
+    if (typeName == "ConvRelu") {
```
rdzhabarov: You need to add a test to Caffe2ImporterTest to verify ConvRelu.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -355,7 +355,10 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      NodeValue relu_in = node->getNthResult(0);
+      node = G_.createRELU(opName, relu_in);
```
rdzhabarov: node = G_.createRELU(opName, node) should be enough. (I'd add opName + ".relu")

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -67,6 +67,43 @@ TEST(caffe2, importConv) {
...
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename,
+                               {"gpu_0/data_0"}, {&data.getType()}, *F);
+    output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
+
+    bindings.allocate(mod.getPlaceholders());
+    updateInputPlaceholdersByName(bindings, &mod, {"gpu_0/data_0"}, {&data});
+  }
+
+  auto res = bindings.get(output);
+  EE.compile(CompilationMode::Infer, F);
```
rdzhabarov: for importer tests the goal is to verify structure, can you check that there are Conv, Relu nodes here.

rdzhabarov:
Test comment, the rest LGTM

zrphercule:
A reminder to myself: dont forget to add comments.

PR #2670, opened 2019-04-06
[Quantizer] Remove logic for updating nodeToTQP_

jfix71:
I'm not really sure why this logic is here. It doesn't really make sense to me -- given some `DequantizeNode`, it adds an entry to `nodeToTQP_` to map from the `DequantizeNode's` output to its \*input's\* quantization parameters. Meanwhile, normally this maps from a Node's output to that output's quantization parameters. So I don't get what's going on here. Without the logic, all unit tests still pass and I also tested `resnet50` and `en2gr`.

If anyone has a good idea why it's necessary please let me know and I'd be happy to kill this PR.

### DIFF
--- a/lib/Quantization/Quantization.cpp
+++ b/lib/Quantization/Quantization.cpp
@@ -451,11 +451,10 @@ protected:
     }
   }
 
-  /// Perform post processing for \p node. E.g., update the nodeToTQP_ for the
-  /// added conversions for \p node, or convert nodes into lookup tables if not
-  /// supported by the backend.
+  /// Perform post processing for \p node. Handles special cases, e.g.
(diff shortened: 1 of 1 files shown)


rdzhabarov:
It's been several evolutions of that code, I do not remember precisely off top of my head.
But I will check for potential use cases (happy to remove otherwise :D))

lib/Quantization/Quantization.cpp
```diff
@@ -329,7 +329,6 @@ class FunctionQuantizer : public FunctionConverter {
 
   /// Update the nodeToTQP_ for the added conversions for \p node.
```
rdzhabarov: I'm not sure comment really reflects what happens well.

PR #2671, opened 2019-04-07
Make GlowErr error strings formattable with strFormat

jackm321:
*Description*:
For error macros (MAKE_ERR, RETURN_ERR, RETURN_ERR_IF_NOT, etc) this PR makes it possible to instead of passing an error string, you can pass a printf-style format string followed by accompanying printf arguments and the error message will be generated from this.

See examples in in PR for more details.

*Testing*:
`ninja check`

*Documentation*:
comments

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -98,21 +98,40 @@ public:
     }
   }
 
-  GlowErr(llvm::StringRef fileName, size_t lineNumber, llvm::StringRef message,
-          ErrorCode ec)
-      : lineNumber_(lineNumber), fileName_(fileName), message_(message),
-        ec_(ec) {}
-
+/// Construct a GlowErr for an error at a given \p lineNumber in a given \p
+/// fileName with a given ErrorCode \p ec. The final arguments to this
+/// constructor are a printf-style format string and the accompanying printf
+/// arguments from which the error message will be constructed.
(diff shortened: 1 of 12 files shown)


include/glow/Support/Error.h
```diff
@@ -98,22 +98,41 @@ class GlowErr final : public llvm::ErrorInfo<GlowErr> {
...
-
+/// Construct a GlowErr for an error at a given \p lineNumber in a given \p
+/// fileName with a given ErrorCode \p ec. The final arguments to this
+/// constructor are a printf-style format string and the accompanying printf
+/// arguments from which the error message will be constructed.
+/// NOTE: This disables format-security warnings that enforce only string
+/// literals be passed to strFormat but GlowErr should only be constructed with
+/// formatStrings as string literals.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-security"
```
jackm321: This is a pretty big wtf, if anyone has any suggestions for improving it, I'm interested in hearing

nickgg: Yeah it's a weird one, I can only assume that one of the inputs is not const in some invocation. Can't see it in this diff though.

Do you still see the error if you use `const char *messageFormat` rather than a const ref to an array?

nickgg:
You'll need to ignore the warning in gcc also, probably.

PR #2673, opened 2019-04-08
[tracing] Add human readable trace thread names

nickgg:
*Description*: Allow setting the name of threads and the process in the tracing output. This is managed by a map of integer thread id to string name in the TraceContext.
*Testing*: unittests & resnet-runtime example.
*Documentation*:
image

### DIFF
--- a/examples/resnet-runtime.cpp
+++ b/examples/resnet-runtime.cpp
@@ -52,8 +52,8 @@ llvm::cl::opt<std::string> tracePath("trace-path",
                                      llvm::cl::init(""),
                                      llvm::cl::cat(category));
 
-std::vector<TraceEvent> allEvents;
 std::mutex eventLock;
+std::unique_ptr<TraceContext> traceContext;
 
 } // namespace
 
@@ -98,9 +98,8 @@ void dispatchClassify(unsigned int id, HostManager *hostManager,
 
         if (!tracePath.empty()) {
           std::lock_guard<std::mutex> l(eventLock);
-          auto &newEvents = context->getTraceContext()->getTraceEvents();
-          std::move(newEvents.begin(), newEvents.end(),
-                    std::back_inserter(allEvents));
+          // Merge this run's TraceEvents into the global TraceContext.
+          traceContext->merge(context->getTraceContext());
         }
 
         if (++returned == maxImages) {
@@ -126,6 +125,12 @@ int main(int argc, char **argv) {
 
   std::unique_ptr<HostManager> hostManager =
       llvm::make_unique<HostManager>(std::move(configs));
+
+  // If tracing is enabled, create a TraceContext to merge each runs events
+  // into.
+  if (!tracePath.empty()) {
+    traceContext = llvm::make_unique<TraceContext>(TraceLevel::STANDARD, 0);
+  }
 
   // Load model, create a context, and add to HostManager.
 
@@ -198,8 +203,7 @@ int main(int argc, char **argv) {
   llvm::outs() << "Finished classifying " << started << " images.\n";
 
   if (!tracePath.empty()) {
-    std::lock_guard<std::mutex> l(eventLock);
-    TraceEvent::dumpTraceEvents(allEvents, tracePath);
+    traceContext->dump(tracePath, "resnet-runtime");
   }
 
   return 0;
--- a/include/glow/Backends/TraceEvents.h
+++ b/include/glow/Backends/TraceEvents.h
@@ -60,8 +60,10 @@ struct TraceEvent {
              std::map<std::string, std::string> a)
       : name(n), timestamp(ts), type(c), tid(t), args(a) {}
 
-  static void dumpTraceEvents(std::vector<TraceEvent> &events,
-                              llvm::StringRef filename);
+  static void
+  dumpTraceEvents(std::vector<TraceEvent> &events, llvm::StringRef filename,
+                  const std::string &processName = "",
+                  const std::map<int, std::string> &threadNames = {});
 
   static uint64_t now();
 };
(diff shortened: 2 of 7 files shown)


include/glow/Backends/TraceEvents.h
```diff
@@ -161,6 +166,19 @@ class TraceContext {
...
+  /// Sets the human readable \p name for thread \tid.
+  void setThreadName(int tid, llvm::StringRef name);
+  /// \returns the list of human readable thread names.
+  std::map<int, std::string> &getThreadNames() { return threadNames_; }
+
+  /// Dumps all TraceEvents in json format to the given \filename, optionally
+  /// with a provided \p processName.
+  void dump(llvm::StringRef filename, std::string processName = "");
+
+  /// Moves all TraceEvents and thread names in \p other into this context.
```
gcatron: What is the use-case for the raw ptr version of merge? Is this to allow a TraceContext to be essentially emptied and then reused without needing to reallocate?
I'd extend the comment a bit to call out that it is destructive on `other`.

nickgg: Yes pretty much, allows reuse of TraceContext. Will add the comment.

nickgg:
In order to debug the coverage run failures i'm going hack up the script a bit and then revert the change.

examples/resnet-runtime.cpp
```diff
@@ -52,8 +52,8 @@ llvm::cl::opt<std::string> tracePath("trace-path",
                                      llvm::cl::init(""),
                                      llvm::cl::cat(category));
 
-std::vector<TraceEvent> allEvents;
 std::mutex eventLock;
+std::unique_ptr<TraceContext> allEvents;
```
rdzhabarov: s/allEvents/traceContext

lib/Backends/TraceEvents.cpp
```diff
@@ -23,18 +23,30 @@
 namespace glow {
 
 void TraceEvent::dumpTraceEvents(std::vector<TraceEvent> &events,
-                                 llvm::StringRef filename) {
+                                 llvm::StringRef filename,
+                                 std::string processName,
```
rdzhabarov: const ref for threadNames

include/glow/Backends/TraceEvents.h
```diff
@@ -161,6 +166,22 @@ class TraceContext {
   void
   logTraceEvent(llvm::StringRef name, llvm::StringRef type, uint64_t timestamp,
                 std::map<std::string, std::string> additionalAttributes = {});
+
+  /// Sets the human readable \p name for thread \tid.
+  void setThreadName(int tid, llvm::StringRef name);
+  /// \returns the list of human readable thread names.
+  std::map<int, std::string> &getThreadNames() { return threadNames_; }
```
rdzhabarov: why do we expose non const reference to the caller?

include/glow/Backends/TraceEvents.h
```diff
@@ -161,6 +166,22 @@ class TraceContext {
...
+
+  /// Dumps all TraceEvents in json format to the given \filename, optionally
+  /// with a provided \p processName.
+  void dump(llvm::StringRef filename, std::string processName = "");
```
rdzhabarov: const ref here for processName.

rdzhabarov: \p filename

gcatron:
> In order to debug the coverage run failures i'm going hack up the script a bit and then revert the change.

Looks like you fixed it, what was the issue?

nickgg:
@gcatron in the HostManager we were using the DeviceConfig in the same parameter list we were moving it into, which is fine in clang which resolves parameters left-to-right but will crash in gcc which resolves parameters right-to-left.

include/glow/Backends/TraceEvents.h
nickgg: because they may want to modify it

include/glow/Backends/TraceEvents.h
nickgg: Can't have a const ref default param.

lib/Backends/TraceEvents.cpp
nickgg: Can't have a const ref default param.

include/glow/Backends/TraceEvents.h
rdzhabarov: https://godbolt.org/z/NepmJQ

include/glow/Backends/TraceEvents.h
rdzhabarov: have not seen it modified

include/glow/Backends/TraceEvents.h
nickgg: huh. ok, guess I'm wrong let me take another look.

nickgg:
Thanks @rdzhabarov for the pointer.

examples/resnet-runtime.cpp
```diff
@@ -127,6 +125,10 @@ int main(int argc, char **argv) {
   std::unique_ptr<HostManager> hostManager =
       llvm::make_unique<HostManager>(std::move(configs));
 
+  if (!tracePath.empty()) {
```
nadavrot: Comments please.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -166,6 +166,11 @@ RunIdentifierTy
 HostManager::runNetwork(llvm::StringRef networkName,
                         std::unique_ptr<ExecutionContext> context,
                         ResultCBTy callback) {
+  auto *resultTraceContext = context->getTraceContext();
+  if (resultTraceContext) {
```
nadavrot: Comments please.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -386,6 +386,10 @@ void ThreadPoolExecutor::executeDAGNode(
   }
 
   auto &deviceManager = deviceManagerIt->second;
+  if (executionState->getRawResultContextPtr()->getTraceContext()) {
+    executionState->getRawResultContextPtr()->getTraceContext()->setThreadName(
```
nadavrot: Comments please.

PR #2674, opened 2019-04-08
Update createBundle to set symbol input and output

### LINKED ISSUE #1953: [runtime] Pass Input Output Placeholder labels
From the Context it's not easy to efficiently identify whether a placeholder is an input or output. This means all placeholders may be copied to and from the device. It would be good to pass this information to runtime to make device loading/unloading more efficient.

gcatron:
*Description*: This updates the create method on RuntimeBundle to properly set input and output status for symbols. 
*Testing*: ninja test: 
*Documentation*:
Fixes #1953

### DIFF
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -19,17 +19,11 @@
 using namespace glow;
 
 using llvm::cast;
+using llvm::dyn_cast;
 using llvm::isa;
 
 void glow::runtime::RuntimeBundle::collectConstants(const IRFunction *F) {
   collectConstants(F->getGraph()->getParent());
-}
-
-void glow::runtime::RuntimeBundle::setInputsandOutputs() {
-  for (auto &symbol : symbolTable_) {
-    symbol.second.input = true;
-    symbol.second.output = true;
-  }
 }
 
 void glow::runtime::RuntimeBundle::freeConstants() {
@@ -81,6 +75,45 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
   return it->second;
 }
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH) {
+  for (const auto &use : PH->getUsers()) {
+    // Look through the inputs of the PH's users. If an input is overwritten
+    // check if it's the PH, if it is return true.
+    auto *user = use.getUser();
+    for (unsigned i = 0, numInputs = user->getNumInputs(); i < numInputs; i++) {
+      // If the input is not overwritten we can continue.
+      if (!user->isOverwrittenNthInput(i)) {
+        continue;
+      }
+      auto input = use.getUser()->getNthInput(i);
+      if (input == PH) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+/// If \p PH is an input placeholder, \returns true.
+bool isInput(const Placeholder *PH) {
+  // Check that the PH is the input to a saveNode or is used by a non saveNode.
+  for (const auto &use : PH->getUsers()) {
+    // Check if PH is an input to a saveNode.
+    if (auto *save = dyn_cast<SaveNode>(use.getUser())) {
+      auto input = save->getInput();
+      // If the PH is not an input to the saveNode we keep looking.
+      if (input != PH) {
+        continue;
+      }
+    }
+    return true;
+  }
+  return false;
+}
+
 runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
   std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
 
@@ -95,6 +128,8 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
     symbol.offset = offset;
     symbol.size = size;
     symbol.type = *V->getType();
+    symbol.input = false;
+    symbol.output = false;
     symbol.symbolCategory = SymbolCategory::Constant;
     symbolTable.emplace(V->getName(), symbol);
   }
@@ -107,6 +142,8 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
     symbol.offset = offset;
     symbol.size = size;
     symbol.type = *V->getType();
+    symbol.output = isOutput(V);
+    symbol.input = isInput(V);
     symbol.symbolCategory = SymbolCategory::Placeholder;
     symbolTable.emplace(V->getName(), symbol);
   }
@@ -134,6 +171,8 @@ runtime::RuntimeBundle::create(const IRFunction &F,
     symbol.size = numBytes;
     symbol.offset = addr;
     symbol.type = *w->getType();
+    symbol.input = false;
+    symbol.output = false;
     symbol.symbolCategory = SymbolCategory::Constant;
     symbolTable.emplace(std::string(v->getName()), symbol);
   }
@@ -150,6 +189,8 @@ runtime::RuntimeBundle::create(const IRFunction &F,
     symbol.offset = addr;
     symbol.size = numBytes;
     symbol.type = *w->getType();
+    symbol.output = isOutput(v);
+    symbol.input = isInput(v);
     symbol.symbolCategory = SymbolCategory::Placeholder;
     symbolTable.emplace(std::string(v->getName()), symbol);
   }
@@ -158,7 +199,7 @@ runtime::RuntimeBundle::create(const IRFunction &F,
   // Compute the offsets for Activations.
 
   for (const auto &I : F.getInstrs()) {
-    if (auto *A = llvm::dyn_cast<AllocActivationInst>(&I)) {
+    if (auto *A = dyn_cast<AllocActivationInst>(&I)) {
       auto numBytes = I.getSizeInBytes();
       size_t addr = activationsAllocator.allocate(numBytes, A);
       assert(!symbolTable.count(std::string(A->getName())) &&
@@ -167,12 +208,14 @@ runtime::RuntimeBundle::create(const IRFunction &F,
       symbol.offset = addr;
       symbol.size = numBytes;
       symbol.type = *A->getType();
+      symbol.input = false;
+      symbol.output = false;
       symbol.symbolCategory = SymbolCategory::Activation;
       symbolTable.emplace(std::string(A->getName()), symbol);
       continue;
     }
 
-    if (auto *TV = llvm::dyn_cast<TensorViewInst>(&I)) {
+    if (auto *TV = dyn_cast<TensorViewInst>(&I)) {
       // Calculate and store the length of the offset into the base, using the
       // source of the tensorview.
       assert(!symbolTable.count(std::string(TV->getName())) &&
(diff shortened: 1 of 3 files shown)


lib/Backends/BackendUtils.cpp
```diff
@@ -150,6 +162,9 @@ runtime::RuntimeBundle::create(const IRFunction &F,
     symbol.offset = addr;
     symbol.size = numBytes;
     symbol.type = *w->getType();
+    bool isOutput = checkIfOutput(v);
+    symbol.output = isOutput;
+    symbol.input = !isOutput;
```
opti-mix: Question: Can it be that the same symbol is used as input and as output at the same time?

gcatron: I'm glad you asked, I'm not sure, I think for quantization it can be? But I believe for a single function a given symbol is one or the other. If there is a chance that input and output can both be true I can change this PR to properly set output, and leave input as true for now.

opti-mix: Should we add a verifier check for this properly that a symbol is always an input or an output in a function, but not both at the same time?

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,38 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
   return it->second;
 }
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool checkIfOutput(const Placeholder *PH) {
```
SplitInfinity: Consider renaming this to `isOutput`, `isOutputPH` or `isOutputPlaceholder`. That makes it more explicit that the return value is a boolean.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,38 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+          }
+        }
+      }
+    }
+  }
+  return false;
+}
+
+/// If \p PH is an input placeholder, \returns true.
+bool checkIfInput(const Placeholder *PH) {
```
SplitInfinity: Same here but :%s/output/input.

nadavrot: You have a big here. A ph can be the input to a save node. You need to check if this is a destination.

nadavrot:
Consider the two scenarios: 1. We have a bug in this code that the unit test catches. 2. We need to change the design of the function to make PH an argument list. 

Is this unit test carrying its weight? It will make it difficult for us to change the design of the compiler and at the same time won’t provide much safety. Maybe we can update an existing test instead of adding a new test that we will need to update later?

lib/Backends/BackendUtils.cpp
bertmaher: I think there’s an additional subtlety here, since a placeholder can be either the input or the overwritten-input of the SaveNode (which is kind of degenerate, but can happen in practice)

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,38 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool checkIfOutput(const Placeholder *PH) {
+  for (auto &use : PH->getUsers()) {
+    // Look through the inputs of the PH's users. If an input is overwritten
+    // check if it's the PH, if it is return true.
+    for (unsigned i = 0; i < use.getUser()->getNumInputs(); i++) {
+      if (use.getUser()->isOverwrittenNthInput(i)) {
```
bertmaher: IMO this is a good place to use `continue` instead of deeply nesting the control flow.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,38 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+        auto input = use.getUser()->getNthInput(i);
+        if (auto overwritten = llvm::dyn_cast<Placeholder>(input)) {
```
bertmaher: I think you can just check `input == PH`, there’s no need to cast to do a pointer equality check.

nadavrot:
There is a bug here. Save nodes accept placeholders as inputs. If this is a save node you need to check if this is the destination.

lib/Backends/BackendUtils.cpp
nadavrot: http://www.llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

lib/Backends/BackendUtils.cpp
gcatron: I think I'm confused. The assumption was if a PH has a single non-SaveNode user it's an input. Is this not the case? Or is it that it can still be considered an input if it's only used by a save node?

lib/Backends/BackendUtils.cpp
gcatron: I'm not sure I understand, Don't I need to examine all overwritten inputs to make sure the PH is not an output? I'm not seeing where I can short circuit the loop.

lib/Backends/BackendUtils.cpp
gcatron: So the method needs to check that the SaveNode uses the placeholder as a nonOverwritten input?

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,49 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+
+/// If \p PH is an input placeholder, \returns true.
+bool isInput(const Placeholder *PH) {
+  // Check if trainable, if it is return true.
+  if (PH->isTraining()) {
+    return true;
+  }
+  for (auto &use : PH->getUsers()) {
+    // Check if PH is an input to saveNode
+    if (auto save = llvm::dyn_cast<SaveNode>(use.getUser())) {
```
bertmaher: I think there’s still a bug here if a placeholder is the non-overwritten input of a SaveNode; it’ll be erroneously marked as an output.  It’s a degenerate case but it can happen.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,55 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+/// overwritten input.
+bool isOutput(const Placeholder *PH) {
+  // Check if trainable, if it is return true.
+  if (PH->isTraining()) {
+    return true;
+  }
+  for (auto &use : PH->getUsers()) {
+    // Look through the inputs of the PH's users. If an input is overwritten
+    // check if it's the PH, if it is return true.
+    for (unsigned i = 0; i < use.getUser()->getNumInputs(); i++) {
```
nadavrot: Don't evaluate getNumInputs on each iteration:

https://releases.llvm.org/7.0.0/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,55 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
   return it->second;
 }
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH) {
+  // Check if trainable, if it is return true.
+  if (PH->isTraining()) {
```
nadavrot: Why is this correct? The isTraining flag is only used by the differentiation method, not by the backend.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,55 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+
+/// If \p PH is an input placeholder, \returns true.
+bool isInput(const Placeholder *PH) {
+  // Check if trainable, if it is return true.
+  if (PH->isTraining()) {
+    return true;
+  }
+  for (auto &use : PH->getUsers()) {
+    // Check if PH is a non overwritten input to a saveNode.
+    if (auto save = llvm::dyn_cast<SaveNode>(use.getUser())) {
```
nadavrot: You can add 

using llvm::dyn_cast; 

at the end of the file so that you won't need to type the llvm namespace every time we use cast or dyn_cast. We do this in a few places in the compiler.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +74,55 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
+      auto input = save->getInput();
+      // Check if the PH is the input to the saveNode.
+      if (input != PH) {
+        continue;
+      }
+      // Check if input is overwritten.
```
nadavrot: What is operand zero?  Prefer to use the descriptive label of SaveNode. (getDestination I think)

lib/Backends/BackendUtils.cpp
gcatron: The thought was if a PH is marked as trainable, it would always be an input and output, but that assumption only holds for training, not inference. Check removed.

lib/Backends/BackendUtils.cpp
gcatron: Operand 0 is input and 1 is output for a saveNode, but after looking at isOverWrittenNthInput, this check is useless, since isOverWritten returns false for all values except 1.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +75,45 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
...
 }
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH) {
+  for (auto &use : PH->getUsers()) {
+    // Look through the inputs of the PH's users. If an input is overwritten
+    // check if it's the PH, if it is return true.
+    for (unsigned i = 0, numInputs = use.getUser()->getNumInputs();
```
jfix71: nit: you use `use.getUser()` multiple times, could just create a local variable for it.

lib/Backends/BackendUtils.cpp
```diff
@@ -81,6 +75,45 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
   return it->second;
 }
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH) {
+  for (auto &use : PH->getUsers()) {
```
jfix71: nit: `const auto &use`

nadavrot:
Nice!

tests/unittests/BackendTest.cpp
```diff
@@ -106,6 +106,64 @@ TEST(Interpreter, profileQuantizationForANetwork) {
...
+      mod.createPlaceholder(ElemKind::FloatTy, {1, 10, 10, 3}, "in", false);
+
+  auto *ex = mod.createConstant(ElemKind::Int64ITy, {1, 1}, "exp");
+
+  auto *FC = F->createFullyConnected(bindings, "FC", input, 30);
+  auto *RL = F->createRELU("RL2", FC);
+  auto *SM = F->createSoftMax("sm", RL, ex);
+  auto *S = F->createSave("ret", SM);
+  auto *qp = F->createQuantizationProfile(bindings, "qp", input);
+  qp->getHistogramPlaceholder()->getName();
```
jfix71: nit: Don't forget to remove this!

PR #2678, opened 2019-04-09
Transfer trace events from Glow across ONNXIFI interface

jackm321:
*Description*:
Transfer Glow trace events across onnxifi interface

*Testing*:
Synced and tested internally

*Documentation*:
doxygen

### DIFF
--- a/include/glow/Backends/TraceEvents.h
+++ b/include/glow/Backends/TraceEvents.h
@@ -142,6 +142,9 @@ public:
   /// \returns TraceEvents for the last run.
   std::vector<TraceEvent> &getTraceEvents() { return traceEvents_; }
 
+  /// \returns TraceEvents for the last run.
+  llvm::ArrayRef<TraceEvent> getTraceEvents() const { return traceEvents_; }
+
   /// \returns the integer thread id used for logged events in this context.
   int getTraceThread() const { return traceThread_; }
 
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -94,8 +94,15 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
                               const onnxTensorDescriptorV1 *inputDescriptors,
                               uint32_t outputsCount,
                               const onnxTensorDescriptorV1 *outputDescriptors,
-                              EventPtr outputEvent) {
+                              EventPtr outputEvent,
+                              onnxTraceEventList *traceEvents) {
   auto ctx = llvm::make_unique<ExecutionContext>();
+
+  if (traceEvents) {
+    // TODO: get thread ID
+    ctx->setTraceContext(
+        llvm::make_unique<TraceContext>(TraceLevel::STANDARD, 0));
+  }
 
   // Create tensors for input placeholders
   for (unsigned i = 0; i < inputsCount; ++i) {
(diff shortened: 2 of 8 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -135,7 +136,48 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+  std::vector<onnxTraceEvent *> traceEventsVec;
+  for (const auto &glowTraceEvent : traceContext.getTraceEvents()) {
+    auto *traceEvent = new onnxTraceEvent();
+    assert(
+        glowTraceEvent.type.size() == 1 &&
+        "Events with types longer than a single char not supported by onnxifi");
+    traceEvent->eventType = glowTraceEvent.type[0];
+    traceEvent->timestamp = glowTraceEvent.timestamp;
+    traceEvent->tid = glowTraceEvent.tid;
+    char *eventName = new char[glowTraceEvent.name.size() + 1];
```
nickgg: why use a temporary here rather than just allocating into `traceEvent->eventName`?

lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -104,6 +106,8 @@ HostManagerGraph::run(std::unique_ptr<ExecutionContext> ctx,
                  res->size() * res->getType().getElementSize());
         }
 
+        setTraceEvents(traceEvents, *ctx->getTraceContext());
```
nickgg: getTraceContext can return nullptr if the initial context does not set a TraceContext.

jackm321: yeah good point, will check that first

lib/Onnxifi/Base.cpp
jackm321: cause it's a const*

include/glow/Backends/TraceEvents.h
```diff
@@ -149,6 +149,9 @@ class TraceContext {
   /// Sets the level of verbosity for TraceEvents.
   void setTraceLevel(TraceLevel level) { traceLevel_ = level; }
 
+  /// Get a const reference to traceEvents_ for reading, doesn't bother locking.
```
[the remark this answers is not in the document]
jackm321: It's the second case. I don't really care if const or not but I'm not modifying it here which is why I took it as const. We could add a const and a non-const getTraceEvents. Likewise we could lock or not. Imo the best thing to do would be to make the lock mutable and then have a const and non const one of these that both take the lock since that's difficult to abuse in the future.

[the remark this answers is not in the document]
nickgg: What do you mean make the lock mutable? 

Another option is we can just move the events out under the lock and process them destructively

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -97,6 +97,8 @@ InlineGraph::run(std::unique_ptr<ExecutionContext> ctx, EventPtr outputEvent,
     serializeToYaml(getProfileFile(modelHash_), QI);
   }
 
+  setTraceEvents(traceEvents, *ctx->getTraceContext());
```
nickgg: Still looks like this could crash...

lib/Onnxifi/Base.cpp
```diff
@@ -157,7 +164,48 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+
+  std::vector<onnxTraceEvent *> traceEventsVec;
+  for (const auto &glowTraceEvent : traceContext.getTraceEvents()) {
+    auto *traceEvent = new onnxTraceEvent();
+    assert(
+        glowTraceEvent.type.size() == 1 &&
+        "Events with types longer than a single char not supported by onnxifi");
+    traceEvent->eventType = glowTraceEvent.type[0];
+    traceEvent->timestamp = glowTraceEvent.timestamp;
+    traceEvent->tid = glowTraceEvent.tid;
```
nickgg: It would be good to include `args` in the onnxTraceEvent since thats were we'll put per run metadata. Not sure if you can do maps

lib/Onnxifi/InlineOnnxifi.cpp
jackm321: Yeah missed one, thanks

lib/Onnxifi/Base.cpp
jackm321: Hmm yeah ok that will require changing the interface so let's do it in a followup

include/glow/Backends/TraceEvents.h
```diff
@@ -142,6 +142,9 @@ class TraceContext {
   /// \returns TraceEvents for the last run.
   std::vector<TraceEvent> &getTraceEvents() { return traceEvents_; }
 
+  /// \returns TraceEvents for the last run.
+  const std::vector<TraceEvent> &getTraceEvents() const { return traceEvents_; }
```
nadavrot: Would ArrayRef be a better interface here?

PR #2680, opened 2019-04-09
[Partitioner] If a proper partition is not found, return an error.

### LINKED ISSUE #2661: Unhelpful error message: "Provisioner found more logical devices than physical devices."
I think what this is trying to say is that I don't have enough memory to run my network?  Or I've configured my device manager with some incorrect amount of memory?  It's hard to know what to do with this string because "logical devices" isn't a terribly intuitive concept.

beicy:
*Description*:
If the number of devices is fewer than the number of partitions, we used to let Provisioner to report this error. Now we move the error detection into Partitioner.  If a proper partition can't be found, an error is reported.

Testing*:
Added unittest.

*Documentation*:
[Optional Fixes #2661 ]

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -19,6 +19,7 @@
 #include "glow/Graph/Graph.h"
 #include "glow/Partitioner/PartitionerUtils.h"
 #include "glow/Runtime/RuntimeTypes.h"
+#include "glow/Support/Error.h"
 
 namespace glow {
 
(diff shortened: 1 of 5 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -546,7 +546,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   }
 }
 
-DAGListTy &Partitioner::Partition() {
+llvm::Error Partitioner::Partition() {
```
opti-mix: Do we usually use `llvm::Error` or `llvm::Expected` in such cases, where a function is supposed to return a value?

beicy: We use llmv::Error. If there is no error, the function will return `llvm::Error::success();`

include/glow/Partitioner/Partitioner.h
```diff
@@ -157,7 +158,9 @@ class Partitioner {
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices);
 
   /// Decompose each function in a module and return a list of DAGNodes.
```
gcatron: Probably should update the comment since this no longer returns the list of DAGNodes.

include/glow/Partitioner/Partitioner.h
```diff
@@ -157,7 +158,9 @@ class Partitioner {
...
-  DAGListTy &Partition();
+  llvm::Error Partition();
+
+  DAGListTy &getPartitionResult() { return partitions_; }
```
gcatron: Need a comment for getPartitionResult.

PR #2683, opened 2019-04-09
Added TensorView enums to SymbolCategory

gcatron:
*Description*: This PR adds two new symbol Categories, TensorViewPlaceholder, and TensorViewConstant. This allows us to differentiate between the view and weight.
*Testing*: ninja test
*Documentation*: NA

### DIFF
--- a/include/glow/Backends/BackendUtils.h
+++ b/include/glow/Backends/BackendUtils.h
@@ -22,9 +22,14 @@
 namespace glow {
 namespace runtime {
 
-/// An enum to indicate what type each symbol in the bundle is. Current options
-/// are activations, constants, and placeholders.
-enum class SymbolCategory { Activation, Placeholder, Constant };
+/// An enum to indicate what type each symbol in the bundle is.
+enum class SymbolCategory {
+  Activation,
+  Placeholder,
+  Constant,
+  PlaceholderTensorView,
+  ConstantTensorView
+};
 
 /// Contains information for initialization and handling of symbol at runtime.
 struct RuntimeSymbolInfo {
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -191,8 +191,13 @@ runtime::RuntimeBundle::create(const IRFunction &F,
                       (offsetLength * TV->getType()->getElementSize());
       symbol.size = TV->getSizeInBytes();
       symbol.type = *TV->getType();
-      symbol.symbolCategory =
+      auto parentCategory =
           symbolTable.find(tvSource->getName())->second.symbolCategory;
+      if (parentCategory == SymbolCategory::Placeholder) {
+        symbol.symbolCategory = SymbolCategory::PlaceholderTensorView;
+      } else {
+        symbol.symbolCategory = SymbolCategory::ConstantTensorView;
+      }
       symbolTable.emplace(std::string(TV->getName()), symbol);
       continue;
     }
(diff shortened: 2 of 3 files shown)


nadavrot:
I don't understand this change. Why do we need to keep the symbols of tensor views? I expected these things to be internal to the backends and low-level IR.

gcatron:
> Looking good.
> 
> Nit: I wonder if it could be useful to have small helper methods like `isPlaceholder` and `isConstant` to check if something is a placeholder or a constant, but without distinguishing between the actual and tensorview-based ones. I dunno, maybe it is not worth the effort.

I'd be afraid that would introduce confusion. I'd probably say that can live as a helper method on a backend that needs it.

gcatron:
> I don't understand this change. Why do we need to keep the symbols of tensor views? I expected these things to be internal to the backends and low-level IR.

It's a good question. The runtimeBundle(probably a bad name...)  is used by both the runtime and our example backends (CPU,Interpreter, and OpenCL) the Activations and TensorViews in the bundle are used by the backends. Placeholders and Constants are used by both the runtime and backends.

include/glow/Backends/BackendUtils.h
```diff
@@ -24,7 +24,13 @@ namespace runtime {
 
 /// An enum to indicate what type each symbol in the bundle is. Current options
 /// are activations, constants, and placeholders.
```
jfix71: Comment here is stale now -- you could probably just remove the second sentence entirely.

nadavrot:
@gcatron I don't understand. There are many different kinds of instructions. What's special about TensorViews? Why do you need it in the SymbolCat?

gcatron:
@nadavrot It's added as a category in order to distinguish from actual Placeholders. @opti-mix would like to distinguish Placeholder vs Constant TVs because that could imply what region of memory the offset refers to. Activations and TVs are in the symbol table because they are the only instructions used to identify a memory buffer.

nadavrot:
@gcatron I don't understand this: "It's added as a category in order to distinguish from actual Placeholders".  @opti-mix do you understand this?

nadavrot:
@gcatron @opti-mix Is this the right way to communicate this information? What about an attribute at the IR level?

opti-mix:
@nadavrot `RuntimeBundle` currently contains `RuntimeSymbolInfo` for ALL `glow::Values` used by the low-level IR, which includes WeightVars representing constant and mutable weights as well as AllocActivations and TensorViews. 

This information is used by the backends and runtime to map a symbol to the actual address that it occupies at runtime. This address is not known at compile-time, as e.g. each run gets its own memory for mutable weights and activations. It is computed at runtime as (base_addr_of_memory_region + offset_of_the_symbol_provided_by_RuntimeSymbolInfo).

`RuntimeSymbolInfo` is also used to decide e.g. which symbols need to be copied to the device. Constant weights need to be copied once. Non-constant input weights need to be copied before each run, non-constant output weights need to be copied from a device after each run. We also want to copy only actual weights, but not their aliases (i.e. not any tensorviews based on those weights), to save useless copies. Thus tensorviews should be distinguished from actual weights.

As for the question of why we keep the information about tensorviews or activations in `RuntimeBundle`, we don't need them in most cases strictly speaking, as these values are an implementation detail of the network implementation. But it is convenient to have them for e.g. debugging purposes or e.g. to make the `DebugPrintInst` work, because this instruction may dump any tensor, i.e. not only weights, but also activations or tensorviews.

PR #2684, opened 2019-04-09
Make quantization profiling scalable

tlepley-cadence:
*Description*: 
If we want to perform the quantization profile generation on many images. The batch size (which corresponds to number of images passed as parameter) becomes too large and Glow requires a too huge amount of memory. This PR allows to (partially) decouple the image list and the batch size. In such a way, quantization profile generation can be performed on a very large number of images.

Content of this PR:
- new `-minibatch` option that specifies the batch size to be used for the compilation
- Bug fix in the implementation of the `QuantizationProfile` node in libjit. On the contrary to the Interpreter implementation, it was not 'accumulating' profile results across successive inference runs.
- Bug fix in the initialization of `QuantizationProfile` min/max values (was I guess initialized to  the default 0 value, what is incorrect)

*Testing*: Locally with our implementation

*Documentation*: N/A

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1630,14 +1630,13 @@ void libjit_write_timestamp(uint64_t *tensor, size_t offset) {
   memcpy(tensor + offset, &ts, sizeof(uint64_t));
 }
 
-// code ported from Profile.cpp: generateTensorHistogram
+/// Update min/max values \p compInfo and histogram \p existingHistogram with
+/// data collected from tensor \p inputTensor.
+/// Note: code ported from Profile.cpp: generateTensorHistogram
 __attribute__((noinline)) void
 libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
                             size_t numDimsTensor, float *compInfo,
                             float *existingHistogram, size_t *histDim) {
-  float minInput = inputTensor[0];
-  float maxInput = inputTensor[0];
-
   size_t tensorSize = 1;
   // calculating total size of the Tensor.
   for (size_t i = 0; i < numDimsTensor; ++i) {
@@ -1646,24 +1645,32 @@ libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
 
   size_t nBins = histDim[0];
 
-  // finding min/max value for entire tensor
-  find_min_max_f(inputTensor, tensorSize, minInput, maxInput);
-
+  // Min/max computed from previous runs. If this is the first run, compInfo is
+  // expected to be initialized as following:
+  // compInfo[0]: std::numeric_limits<float>::max()
+  // compInfo[1]: std::numeric_limits<float>::lowest()
   float min = compInfo[0];
   float max = compInfo[1];
-  if (check_all_zeros(existingHistogram, nBins) == 1) {
-    compInfo[0] = minInput;
-    compInfo[1] = maxInput;
 
+  // Min/max value for entire current input tensor.
+  float minInput;
+  float maxInput;
+  find_min_max_f(inputTensor, tensorSize, minInput, maxInput);
+
+  // Update the global min/max.
+  float newMin = MIN(minInput, min);
+  float newMax = MAX(maxInput, max);
+  compInfo[0] = newMin;
+  compInfo[1] = newMax;
+
+  // Initial profile.
+  if (check_all_zeros(existingHistogram, nBins) == 1) {
     min = minInput;
     max = maxInput;
   }
 
-  // Check if we need to rescale histogram.
-  if (minInput < min || maxInput > max) {
-    float newMin = MIN(minInput, min);
-    float newMax = MAX(maxInput, max);
-
+  // If the min/max range changes, there is the need to rescale the histogram.
+  if (newMin < min || newMax > max) {
     float destBinWidth = (newMax - newMin) / nBins;
     float srcBinWidth = (max - min) / nBins;
     float scaledHistogram[nBins];
@@ -1712,6 +1719,7 @@ libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
     max = newMax;
   }
 
+  // Update the histogram with the values of the current input tensor.
   float binWidth = (max - min) / nBins;
   for (size_t i = 0, e = tensorSize; i < e; ++i) {
     size_t newBin = get_bin(nBins, binWidth, min, inputTensor[i]);
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1634,15 +1634,9 @@ void libjit_write_timestamp(uint64_t *tensor, size_t offset) {
 /// data collected from tensor \p inputTensor.
 /// Note: code ported from Profile.cpp: generateTensorHistogram
 __attribute__((noinline)) void
-libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
-                            size_t numDimsTensor, float *compInfo,
-                            float *existingHistogram, size_t *histDim) {
-  size_t tensorSize = 1;
-  // calculating total size of the Tensor.
-  for (size_t i = 0; i < numDimsTensor; ++i) {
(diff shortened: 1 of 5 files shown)


rdzhabarov:
> If we want to perform the quantization profile generation on many images. The batch size (which corresponds to number of images passed as parameter) becomes too large and Glow requires a too huge amount of memory

This is awesome!

@tlepley could you please send bug fixes with tests covering those bugs in a separate PRs.

tlepley-cadence:
@rdzhabarov Well, I can't really decouple both. We can only see the CPU `QuantizationProfile` bug only when using the `-minibatch` option. This is why I've put them together in the PR.

tools/loader/ImageClassifier.cpp
```diff
@@ -321,26 +348,47 @@ int main(int argc, char **argv) {
...
+    inputImageBatchFilenames = inputImageFilenames;
+  }
+
+  // Print out the inferred image classification.
+  llvm::outs() << "Model: " << loader.getFunction()->getName() << "\n";
+
   while ((streamInputFilenamesMode &&
-          getNextImageFilenames(&inputImageFilenames)) ||
+          getNextImageFilenames(&inputImageBatchFilenames)) ||
+         (miniBatch &&
```
jfix71: nit: Use `miniBatchMode` here.

tools/loader/ImageClassifier.cpp
```diff
@@ -101,6 +112,24 @@ static bool getNextImageFilenames(std::vector<std::string> *filenames) {
   return !filenames->empty();
 }
 
+/// Generate in \p imageList the list of filenames corresponding to the next
+/// mini-batch of size \p miniBatchSize extracted from \p totalImageList at
+/// index \p minibatchIndex.
+static bool getNextMiniBatch(std::vector<std::string> &imageList,
```
jfix71: nit: Maybe could use `llvm::MutableArrayRef<std::string> imageList` here. Though not sure there's a `clear()` for it.

tools/loader/ImageClassifier.cpp
```diff
@@ -101,6 +112,24 @@ static bool getNextImageFilenames(std::vector<std::string> *filenames) {
...
+                             std::vector<std::string> &totalImageList,
```
jfix71: nit: could use `llvm::ArrayRef<std::string> totalImageList` here. Or at least make this `const`.

tools/loader/ImageClassifier.cpp
```diff
@@ -101,6 +112,24 @@ static bool getNextImageFilenames(std::vector<std::string> *filenames) {
   return !filenames->empty();
 }
 
+/// Generate in \p imageList the list of filenames corresponding to the next
+/// mini-batch of size \p miniBatchSize extracted from \p totalImageList at
+/// index \p minibatchIndex.
```
jfix71: Can you also update the doxygen here to include `\returns`, and also mention that `minibatchIndex` is updated?

tools/loader/ImageClassifier.cpp
```diff
@@ -53,6 +53,17 @@ llvm::cl::opt<std::string> inputImageListFile(
     llvm::cl::value_desc("string_name"), llvm::cl::Optional,
     llvm::cl::cat(imageLoaderCat));
 
+llvm::cl::opt<unsigned> miniBatch(
+    "minibatch",
+    llvm::cl::desc(
+        "Size of mini-batches. Split the input image list into a set a "
```
jackm321: should it be "set of"?

tlepley-cadence: Done

tools/loader/ImageClassifier.cpp
tlepley-cadence: Done

tools/loader/ImageClassifier.cpp
tlepley-cadence: Done

tools/loader/ImageClassifier.cpp
tlepley-cadence: `llvm::MutableArrayRef` is quite limited: no clear, not possibile to add an element too it seems. I then kept `std::vector<std::string> &`

tools/loader/ImageClassifier.cpp
tlepley-cadence: Done

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1630,14 +1630,13 @@ void libjit_write_timestamp(uint64_t *tensor, size_t offset) {
...
+/// data collected from tensor \p inputTensor.
+/// Note: code ported from Profile.cpp: generateTensorHistogram
 __attribute__((noinline)) void
 libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
                             size_t numDimsTensor, float *compInfo,
                             float *existingHistogram, size_t *histDim) {
-  float minInput = inputTensor[0];
-  float maxInput = inputTensor[0];
-
   size_t tensorSize = 1;
```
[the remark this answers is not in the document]
tlepley-cadence: Done. Good catch, it's indeed simpler and faster this way.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1646,24 +1645,32 @@ libjit_quantization_profile(float *inputTensor, size_t *tensorDim,
...
+  float maxInput;
+  find_min_max_f(inputTensor, tensorSize, minInput, maxInput);
+
+  // Update the global min/max.
+  float newMin = MIN(minInput, min);
+  float newMax = MAX(maxInput, max);
+  compInfo[0] = newMin;
+  compInfo[1] = newMax;
+
+  // Initial profile
```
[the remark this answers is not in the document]
tlepley-cadence: Done

tlepley-cadence:
@jfix71 @jackm321 all done.

tools/loader/ImageClassifier.cpp
```diff
@@ -321,26 +349,47 @@ int main(int argc, char **argv) {
...
   // Used to make sure we only compile once, and run only once if not streaming.
   bool isFirstRun = true;
 
   // These will be set during the first run.
   Placeholder *inputImagePH = nullptr;
   Tensor *SMT = nullptr;
-
+  size_t minibatchIndex = 0;
   Tensor inputImageData;
+  std::vector<std::string> inputImageBatchFilenames;
```
jfix71: nit: Realizing now that we could just re-use `inputImageFilenames` instead of adding this new `inputImageBatchFilenames`. We were already re-using `inputImageFilenames` inside of `getNextImageFilenames()`, which is pretty similar to `getNextMiniBatch()`. Not a huge deal, but would remove an extra `std::vector<std::string>`.

tools/loader/ImageClassifier.cpp
```diff
@@ -101,6 +112,25 @@ static bool getNextImageFilenames(std::vector<std::string> *filenames) {
   return !filenames->empty();
 }
 
+/// Generate in \p imageList the list of filenames corresponding to the next
+/// mini-batch of size \p miniBatchSize extracted from \p totalImageList at
+/// index \p minibatchIndex. /returns true if the index is valid, false
+/// otherwise.
```
jfix71: nit: Please update to also mention `minibatchIndex` is incremented by `miniBatchSize`, as that is not clear from the caller's perspective.

tools/loader/ImageClassifier.cpp
tlepley-cadence: We actually need both `inputImageBatchFilenames` and `inputImageFilenames`for mini(batching). The first contains the full list of images while the second contains only images of the current mini-batch. Its different from getting the image list from stdin because we have the full list of images at the beginning with mini-batching. 
`getNextMiniBatch(inputImageBatchFilenames, inputImageFilenames, minibatchIndex, miniBatch)`

jfix71: Ah right, got it!

tools/loader/ImageClassifier.cpp
tlepley-cadence: Done.

PR #2685, opened 2019-04-09
[habana] Support non 0 quantized offset

bertmaher:
Summary: Add support for non 0 offset for quantized ops.

Differential Revision: D14808564


bertmaher:
Hopefully these commits preserve authorship since this is actually yours, @rdzhabarov  :-D

PR #2688, opened 2019-04-09
[habana] Stub getMemInfo

bertmaher:
Summary:
I took this idea from @jackm321, we need to actually estimate
memory or the host manager doesn't work.

Differential Revision: D14816916


lib/Backends/Habana/HabanaDeviceManager.cpp
```diff
@@ -100,6 +96,26 @@ llvm::Error HabanaDeviceManager::init() {
   return llvm::Error::success();
 }
 
+llvm::Error HabanaDeviceManager::updateMemoryUsage() {
+  // TODO: use synGetMemInfo once implemented
```
nadavrot: Period. 

Period.

PR #2689, opened 2019-04-09
[glow-h] Optimize before compiling on H

bertmaher:
Summary:
The host manager doesn't do this, which is a bug on its part I guess,
but the Habana backend needs it because it expects specialized operator versions.


bertmaher:
As per linked issue, I'm going to close this PR in favor of a more principled fix.

PR #2692, opened 2019-04-09
[habana] Find SaveNode of output Placeholder w.r.t. specific Function

bertmaher:
Summary:
When we have partitioned models in which a Module has multiple
Functions, when we're looking for an output Placeholder we need to be specific
about which Function uses that Placeholder as an output (since it could be an
output for a *different* Function).

Reviewed By: rdzhabarov

Differential Revision: D14840737

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -93,10 +93,10 @@ static std::string getKernelName(llvm::StringRef kernelBase, ElemKind kind) {
 }
 
 /// If \p PH is an output placeholder, \returns the SaveNode.
-static SaveNode *getOutputSave(Placeholder *PH) {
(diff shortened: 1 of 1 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -93,10 +93,10 @@ static std::string getKernelName(llvm::StringRef kernelBase, ElemKind kind) {
 }
 
 /// If \p PH is an output placeholder, \returns the SaveNode.
-static SaveNode *getOutputSave(Placeholder *PH) {
```
nadavrot: LGTM.  We should probably move this method to the target independent part of the compiler.

bertmaher: Good idea, we've probably reimplemented this function at least a few other places.

PR #2693, opened 2019-04-09
[glow-h] Die loudly if !isOpSupported during compilation

bertmaher:
Summary:
To fully sanity check our compilation results, check isOpSupported for
each node while compiling.  Specifically, fp32 convolution is not supported by
H and we were silently miscompiling code as a result.

Differential Revision: D14841987


bertmaher:
As per linked issue, I'm going to close this PR in favor of a more principled fix.

Issue #2504, opened 2019-03-08
[ModelZoo] support emotions_ferplus model


nickgg:
Add support for the FER+ Emotion Recognition model in glow (https://github.com/onnx/models/tree/master/emotion_ferplus) and verify via a runnable example we can include in the ModelZoo regression test.

The image format for this model is 64x64 0to1.

This issue blocks #2500.

arunm-git:
@nickgg : is the image format really 0to1?
Looking at https://github.com/onnx/models/tree/master/emotion_ferplus#preprocessing, it seems like 0to255, right? At least, when I run that preprocessing on a bunch of images, I get values that appear to lie within 0 to 255.

nickgg:
Looks like it's not, sorry!

PR #2696, opened 2019-04-09
[OnnxModelLoader] Add SAME_UPPER/SAME_LOWER auto_pad modes.

### LINKED ISSUE #2503: [ModelZoo] Support DUC model
Add support for the Dense Upsampling Convolution (DUC) semantic segmentation model in glow (https://github.com/onnx/models/tree/master/models/semantic_segmentation/DUC) and verify via a runnable example we can include in the ModelZoo regression test.

This model requires a different image set than the imagenet images we include with glow (it was trained on the cityscapes dataset), but may be runnable with the ImageClassifier and the right configuration.

This issue blocks #2500.

arunm-git:
*Description*:

Add support for SAME_UPPER and SAME_LOWER auto_pad modes per the spec at
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Conv

Also applies to AveragePool.

*Testing*:
Added unit tests for VALID/SAME_UPPER/SAME_LOWER.
ninja check

*Issues*:
Related to #2504

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -198,7 +198,14 @@ namespace {
 using Pads = std::vector<unsigned_t>;
 } // namespace
 
-llvm::Expected<Pads> getPads(const ArgumentDictionaryTy &dict) {
+/// Get the Pads value based on setting for auto_pad.
+/// \p kdim : kernel sizes (HW)
+/// \p sdim: stride sizes (HW)
+/// \p idim: input sizes (HW)
+llvm::Expected<Pads> getPads(const ArgumentDictionaryTy &dict,
+                             llvm::ArrayRef<unsigned_t> kdim,
+                             llvm::ArrayRef<unsigned_t> sdim,
+                             llvm::ArrayRef<unsigned_t> idim) {
   if (dict.count("pads")) {
     return getShape<unsigned_t>(dict.at("pads"));
   }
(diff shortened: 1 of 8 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -208,8 +215,38 @@ llvm::Expected<Pads> getPads(const ArgumentDictionaryTy &dict) {
...
+      // o = floor((i + 2*p - k)/s) + 1
+      // Also, from https://github.com/onnx/onnx/blob/master/docs/Operators.md
+      // output_spatial_shape[i] =
+      //     ceil(input_spatial_shape[i] / strides_spatial_shape[i])
+      // pad_shape[i] =
+      //     (output_spatial_shape[i] - 1) * strides_spatial_shape[i]
+      //         + kernel_spatial_shape[i] - input_spatial_shape[i]
+      // Use the smallest padding possible out of the possible options.
+      std::vector<unsigned_t> pdim(2); // Total Paddding, HW.
+      for (size_t i = 0; i < pdim.size(); i++) {
```
jackm321: use something like `(size_t i = 0, size_t e = pdim.size(); i < e;...`

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -208,8 +215,38 @@ llvm::Expected<Pads> getPads(const ArgumentDictionaryTy &dict) {
...
+      // From https://arxiv.org/pdf/1603.07285.pdf 2.4,
+      // o = floor((i + 2*p - k)/s) + 1
+      // Also, from https://github.com/onnx/onnx/blob/master/docs/Operators.md
+      // output_spatial_shape[i] =
+      //     ceil(input_spatial_shape[i] / strides_spatial_shape[i])
+      // pad_shape[i] =
+      //     (output_spatial_shape[i] - 1) * strides_spatial_shape[i]
+      //         + kernel_spatial_shape[i] - input_spatial_shape[i]
+      // Use the smallest padding possible out of the possible options.
+      std::vector<unsigned_t> pdim(2); // Total Paddding, HW.
```
jackm321: This probably should be small vector size 2 but the other things (sdim, kdim, etc) aren't either so up to you

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -553,6 +592,14 @@ llvm::Error ONNXModelLoader::loadPool(const ONNX_NAMESPACE::NodeProto &op,
     kernels[1] = Ty->dims()[3];
   }
 
+  // NHWC
+  std::vector<unsigned_t> idim(2);
```
jackm321: The `idimHW` name above is move helpful

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -199,16 +199,19 @@ TEST(onnx, importUniBroadcastMultiOutput) {
...
-/// Test loading conv op from a ONNX model.
-/// The input is N*C*H*W (1*1*3*3), the kernels is {2, 2},
-/// strides is {1, 1}, pads is {1, 1, 1, 1}, group is 1.
-TEST(onnx, importConv) {
+/// Helper method to run the Conv operator test cases.
+/// \p filename contains the model .onnxtxt.
+/// \p expectedDims: output Tensor dimensions.
+/// \p expectedValues : output Tensor values expected.
+static void convTestHelper(std::string &filename,
+                           std::vector<size_t> &expectedDims,
```
jackm321: llvm::ArrayRefs

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -252,12 +255,53 @@ TEST(onnx, importConv) {
   EE.compile(CompilationMode::Infer, F);
   EE.run(bindings);
   auto result = bindings.get(graphOutputVar)->getHandle();
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
+  for (size_t i = 0; i < expectedDims.size(); i++)
```
jackm321: Use braces for for loop
You mean to loop over the expectedValues.size() right?

jackm321: Also same thing with `size_t e =  expectedValues.size()`

arunm-git: Good catch. Done.

arunm-git:
Added tests for AveragePool. I was mistaken when I mentioned GlobalAveragePool - doesn't apply there (the code invocation is the same and does the right thing, but there's no padding for that operator).

PR #2698, opened 2019-04-10
Schedule QuantizationProfile instructions right after the last node that updates their inputs to shorten lifetimes of buffers

### LINKED ISSUE #2697: Profile quantization consumes too much memory
The problem of scalability of quantization profile generation is addressed by: #2684 (minibatch)

Nevertheless, at equal batch size, performing quantization profile generation increases a lot the memory requirement, compared to standard float inference. This is a problem I discussed with @opti-mix at the Glow workshop. For instance for a model we tested, it increases from 8GB to 50GB. 

The insertion of QuantizeProfile nodes prevent the merging of elementwise nodes and QuantizeProfile need extra tensors (histogram), but the memory increase should not be as huge.

It turns out that the IR instruction scheduling (`sinkAllocas` ?) that targets reducing the live-range overlap of tensors doesn't work properly with `QuantizationProfile` nodes. When we dump the low-level Glow ir, we can see that `QuantizedProfile` nodes are all scheduled at the end. It has for effect to keep tensors of the model alive almost during the full model execution.

I didn't investigate further yet. Is it a side effect of the in-out param of `QuantizeProfile` nodes, what may disturb the memory optimization (in-out param are alive permanently) ?

opti-mix:
This handles QuantizationProfileNode scheduling in a more general way, which is not dependent on the kind of the node. It schedules QuantizationProfile instructions right after the last node that updates their inputs to shorten lifetimes of buffers.

Also, take the opportunity to generalize the "SaveNode hack" in the scheduler and make it independent of the Node's kind.

Fixes #2697

### DIFF
--- a/lib/IR/ChildMemSizeBasedScheduler.cpp
+++ b/lib/IR/ChildMemSizeBasedScheduler.cpp
@@ -98,24 +98,27 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
     orderedChildren.push_back(N->getPredicate());
   }
 
-  // SaveNode hack:
   // We don't model memory dependencies, but we still need to honor them.
-  // Make sure the SaveNode happens after the last use of the output
-  // placeholder.
-  if (auto *save = dyn_cast<SaveNode>(N)) {
-    auto *destination = save->getOutput().getNode();
+  // Make sure the a node mutating any of its inputs happens after the last
+  // non-mutating use of the operand being mutated. Some examples of such nodes
+  // would be SaveNode and QuantizationProfileNode.
+  for (unsigned idx = 0, e = N->getNumInputs(); idx < e; ++idx) {
+    // We don't care about inputs that are not mutated by the node.
+    if (!N->isOverwrittenNthInput(idx)) {
+      continue;
+    }
+    auto mutatedInput = N->getNthInput(idx);
+    auto *destination = mutatedInput.getNode();
     for (NodeUse &use : destination->getUsers()) {
       Node *user = use.getUser();
-      if (user == save) {
+      if (user == N) {
         continue;
       }
-      // Storage nodes may have users scattered across different functions.
+      // Nodes may have users scattered across different functions.
       // Only accounts for the ones in that function.
       if (&G_ != user->getParent()) {
         continue;
       }
-      assert(!isa<SaveNode>(user) &&
-             "Placeholder must be saved at most once in each function");
       orderedChildren.push_back(user);
     }
   }
@@ -148,9 +151,34 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
     orderChildNodesAndSchedule(child);
   }
 
-  // Schedule the node after all its children are scheduled.
-  DEBUG_GLOW(llvm::dbgs() << "Scheduled node: " << N->getName() << "\n");
+  // Schedule the node after all its children are scheduled. We need to perform
+  // an extra isScheduled check here, because the code below may have scheduled
+  // the current node while scheduling its children.
+  if (isScheduled(N)) {
+    return;
+  }
   scheduled_.push_back(N);
+  // If this node has a user which does not have any users and which does not
+  // require any additional memory, schedule it here, because we don't want to
+  // extend the lifetime of this value for no reason. We want to execute and get
+  // rid of this node as soon as possible to reduce the memory pressure.
+  for (NodeUse &use : N->getUsers()) {
+    Node *user = use.getUser();
+    // Users may be scattered across different functions.
+    // Only accounts for the ones in that function.
+    if (&G_ != user->getParent()) {
+      continue;
+    }
+    // Bail if a nodes has users, because nodes that have users can't be
+    // scheduled safely without violating dependencies.
+    if (user->getNumUsers()) {
+      continue;
+    }
+    // Schedule a node if it does not require any additional memory.
+    if (resultMemSize_[user] == 0) {
+      orderChildNodesAndSchedule(user);
+    }
+  }
 }
 
 void ChildMemSizeBasedScheduler::scheduleNodes() {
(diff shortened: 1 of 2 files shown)


nadavrot:
I wonder why the scheduler is not scheduling the profile nodes early and drags them all the way to the end of the functions.

opti-mix:
> I wonder why the scheduler is not scheduling the profile nodes early and drags them all the way to the end of the functions.

I looked at it. The thing is that profile nodes do not free any memory themselves. The have no outputs and all their inputs are `@in` or `@inout`. Since the scheduler tries to schedule first the nodes that free most memory and these nodes do not free anything, it postpones the scheduling of profile nodes for later.

nadavrot:
Maybe we should change the way we calculate what freeing memory means? Because scheduling these nodes early will reduce memory pressure. No?

opti-mix:
> Maybe we should change the way we calculate what freeing memory means? Because scheduling these nodes early will reduce memory pressure. No?

Maybe. I'll check if changing the scheduler would result in an easier solution.

opti-mix:
@nadavrot 

> Maybe we should change the way we calculate what freeing memory means? Because scheduling these nodes early will reduce memory pressure. No?

I added a second commit to this PR, which changes the scheduler. I leave both commits in this PR for now so that it is easier to compare both approaches.

tlepley-cadence:
I don't think I'll have time to give it a try today. I'll check and give it feedbacks tomorrow.
In general, I think the instruction scheduler is the right place to handle such memory pressure problematic. I also think that the code should be generic, not specific to the `QuantizeProfile` node: it's a about correctly handling in-out params.

tlepley-cadence:
I did some experiments with resnet50 and a batch size of 512.
I get the following memory requirement for the activation tensors for profile generation:
- original version:  45.6 GB
- scheduler and hoisting version: 2.5GB
This is a nice memory optimization ! :-)

Regarding the code itself, the scheduler version looks simpler and has my preference. Both versions nevertheless are specific to `QuantizeProfile` node. This is maybe a pragmatic approach, but I think this could certainly be handled as a node-kind independent optimization problem.

opti-mix:
> I did some experiments with resnet50 and a batch size of 512.
I get the following memory requirement for the activation tensors memory-pool at profile generation time:
>    original version: 45.6 GB
>    scheduler and hoisting version: 2.5GB
> This is a nice memory optimization ! :-)

Awesome! I'm glad it helps!

> Regarding the code itself, the scheduler version looks simpler and has my preference. Both versions nevertheless are specific to QuantizeProfile node. This is may be a pragmatic approach, but I think this could certainly be handled as a node-kind independent optimization problem.

I'm proudly presenting the alternative number 3!!! :-D :-D I just pushed it. It is simpler than both previous attempts and it is also node-kind independent. Please give it a try. You can comment out the invocations of two others to see the effect of this one in isolation.

lib/IR/ChildMemSizeBasedScheduler.cpp
```diff
@@ -98,24 +98,26 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
     orderedChildren.push_back(N->getPredicate());
   }
 
-  // SaveNode hack:
   // We don't model memory dependencies, but we still need to honor them.
```
nadavrot: I suggest mentioning the SaveNode and Profile* nodes as an example,

lib/IR/ChildMemSizeBasedScheduler.cpp
```diff
@@ -149,8 +151,28 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
   }
 
   // Schedule the node after all its children are scheduled.
```
nadavrot: Why do we need this here? We already checked that the node is not isSheduled(N);

lib/IR/ChildMemSizeBasedScheduler.cpp
```diff
@@ -149,8 +151,28 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
...
-  DEBUG_GLOW(llvm::dbgs() << "Scheduled node: " << N->getName() << "\n");
+  if (isScheduled(N)) {
+    return;
+  }
   scheduled_.push_back(N);
+  // If this node has a user which does not have any users and which does not
```
nadavrot: I would continue this commit with "because ...".  

... because we don't want to extend the lifetime of this value for no reason. We want to execute and get rid of this node as soon as possible to reduce the memory pressure.

lib/IR/ChildMemSizeBasedScheduler.cpp
```diff
@@ -149,8 +151,28 @@ void ChildMemSizeBasedScheduler::orderChildNodesAndSchedule(Node *N) {
...
+  // require any additional memory, schedule it here.
+  for (NodeUse &use : N->getUsers()) {
+    Node *user = use.getUser();
+    // Users may be scattered across different functions.
+    // Only accounts for the ones in that function.
+    if (&G_ != user->getParent()) {
+      continue;
+    }
+    // Bail if a nodes has users.
```
nadavrot: ... because nodes that have users can't be scheduled safely without violating dependencies.

lib/IR/ChildMemSizeBasedScheduler.cpp
opti-mix: Added a comment. We need to perform an extra isScheduled check here, because the code below may have scheduled the current node while scheduling its children.

opti-mix:
I personally like `#3` most, as it is the smallest, least intrusive and more general solution. Waiting for @tlepley-cadence to confirm if its works for him.

opti-mix:
@tlepley-cadence Could you check if my last version produces the same good results?

rdzhabarov:
do not close this

tlepley-cadence:
@opti-mix Sorry for the delay. I just come back from holiday. I'll look at this tomorrow.

tlepley-cadence:
I finally did the check with the latest solution :) 
It gives the same (good) results as the 2 earlier solution ! 
I also much prefer `#3` since it's generic, independent from node types.

opti-mix:
@tlepley-cadence 
> I finally did the check with the latest solution :)
It gives the same (good) results as the 2 earlier solution !

Yay! Finally! :-) Thanks for the confirmation!

> I also much prefer #3 since it's generic, independent from node types.

OK. I'll land the solution #3 in the final version of this PR.

opti-mix:
Will land soon. This week or next week.

Issue #2648, opened 2019-04-03
Currently calling lower twice during profiling/quantization


jfix71:
Right now we call `lower()` twice when doing profiling or quantization. E.g. for the Loader, we're calling `lower()` once prior to profiling or quantization inside `Loader::compile()`, and then again inside `Backend::optimizeFunction()`. We really only need to call `lower()` once.

One simple solution here is to add a new option to `CompilationOptions` for `skipLowering`, set when we know we've already called `lower()` prior to calling `compile()`.

Another more involved solution would be to move `quantizeFunction()`/`profileQuantization()` calls directly into `Backend::optimizeFunction()`, so that it could be done after `lower()`. This would require passing a lot of info through the `ExecutionEngine::compile()`, e.g. `quantizationInfos`, `quantizationSchema`, `bindings`, etc. (could be added to `CompilationOptions`).

nickgg:
it's a nit, but if you go with the first solution it's cleaner to have a `lower` option which defaults true rather than a `skipLowering` which defaults false.

jfix71:
Note that https://github.com/pytorch/glow/pull/2613/ currently implement `fold()` inside `optimizeFunction()`. If it stays there then the first solution above wouldn't work well, because the graph would be optimized and lowered before being folded, which may end up undoing lowering we want to perform, and also may make pattern matching inside folding more difficult.

One other option there is to move `fold()` to the end of importing (e.g. at the end of the `ONNXModelLoader`/`Caffe2ModelLoader`). This means we would be sure that no graph transformations have occurred prior to folding, which I think might make the most sense.

tlepley-cadence:
> One simple solution here is to add a new option to `CompilationOptions` for `skipLowering`, set when we know we've already called `lower()` prior to calling `compile()`.

I don't think this solution will work: the 2 calls to `lower()` serve different purposes and may do different things. The only purpose of the first lower is to control at which granularity the profiling is done. During the first lowering, the lowering for some operations may be disabled with the command line option `do-not-lower-nodes-for-profiling` (useful if the target backend is different from the backend that performs the profiling). Then, we still need to call `lower()` a second time to make sure the IR is compliant to what expects the backend that will perform the profiling.

jfix71:
Fixed via #2700

PR #2700, opened 2019-04-10
Move all precision transformation into optimizeFunction()

jfix71:
Summary: Move quantization/profiling/fp16 conversion into `optimizeFunction()`. This means we can call lower just once in a normal way, instead of forcing it to occur twice. Also cleans up callers to `compile()` so they do not need to worry about when/how to quantize/profile/convert to fp16.

Documentation: Added.

Test Plan: All tests pass.

Fixes https://github.com/pytorch/glow/issues/2648

### DIFF
--- a/docs/GraphOptimizationPipeline.md
+++ b/docs/GraphOptimizationPipeline.md
@@ -1,0 +1,151 @@
+## Glow Graph Optimization Pipeline
+
+This document describes the optimization pipeline and how to configure it for
+varying precision configurations. Note that the terms "Graph" and "Function" are
+often used interchangably.
+
+### Graph (Function) Optimization Stages and Order
+
+`glow::optimizeFunction()` is used to drive the optimization pipeline of a
+Function. Inside `optimizeFunction()`, a Function is optimized by the following
+optimization stages in order:
+
+- `fold()`
+- `optimize()`
+- `lower()`
+- `transformForPrecisionMode()`
+- `optimize()`
+- `Backend::transformPostLowering()`
+- `optimize()`
+- `checkAllNodesSupported()`
+
+Note that optimize is called many times, usually after every other stage. This
+is because the other stages may add or change the graph in some ways that allow
+for previously inapplicable optimizations to now occur, or new nodes may be able
+to take advantage of the optimizations.
+
+### Descriptions of Optimization Stages
+
+- `glow::fold()`: Fold low-level Nodes into higher-level Nodes. This is useful
+  when compiling an input model where some high-level operators have been
+  lowered (this can be for instance a side effect of model converters, like
+  converters from Tensorflow to ONNX). In this situation, such folding can then
+  enable more optimizations and also improve the performance of backends that
+  support natively such high-level operators. Folding is done first, as we want
+  to raise the graph to a higher level in order to take advantage of high-level
+  optimizations and allow for backends to prevent lowering on them as well if
+  desired.
+
+- `glow::lower()`: Lowers high-level Nodes into lower-level Nodes. This allows
+  backends to be agnostic to higher-level representations of Nodes. For example,
+  a backend may support an LSTM if it supports all of the sub-nodes found inside
+  an LSTM. However, it does not need to be responsible for understanding that it
+  supports an LSTM. This future-proofs the backend, allowing it to support
+  future high-level complex nodes without understanding their higher-level
+  semantics. Note that [Backends can prevent
+  lowering](Backends.md#backend-abstract-class) if preferred via
+  `Backend::shouldLower()`.
+
+- `transformForPrecisionMode()`: Transforms the graph depending on requested
+  precision configuration. There are a few different options here:
+
+  - `Profile`: Add special profiling nodes to the graph to gather a histogram of
+    values flowing through each Node in the graph. This can then be used to
+    automatically quantize the graph.
+
+  - `Quantize`: Quantize the graph, given a previously gathered profile as
+    mentioned above. This converts all Float inputs and outputs of each Node if
+    the backend supports the node as quantized.
+
+  - `ConvertToFP16`: Convert all Float inputs and outputs of Nodes to
+    Float16. Note that this does not require any sort of profile. It can also be
+    performed alongside Quantization to get a mixed precision graph.
+
+- `Backend::transformPostLowering()`: Allow the Backend to transform the
+  graph. This includes swapping in its own backend-specific nodes. More info can
+  be found [here](Backends.md#backend-abstract-class) and
+  [here](NewBackendSpecificNode.md#steps).
+
+- `glow::optimize()`: Performs a series of graph optimizations, as listed
+  [here](Optimizations.md#set-of-supported-graph-optimizations). Many of these
+  are common compiler optimizations, such as DCE and CSE. Others are more ML and
+  linear algebra related, such as fusing BatchNormalization into Convolutions in
+  inference mode, or combining a series of Transpose operations into a single
+  Transpose.
+
+- `checkAllNodesSupported()`: Given some Backend that we are optimizing the
+  Graph for, this verifies that the backend supports each of the nodes after the
+  entire optimization pipeline is complete, via
+  [Backend::isOpSupported()](Backends.md#backend-abstract-class). If any node is
+  not supported it will return an error, which `optimizeFunction()` will then
+  return up the stack.
+
+## How to call `glow::optimizeFunction()`
+
+Here we describe the API for `glow::optimizeFunction()` and how to use it in
+different modes.
+
+```
+llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
+                                   const CompilationContext &cctx);
+```
+
+An error is returned if something goes wrong during the optimization pipeline,
+for example if a Node is no longer supported by the backend after optimizations,
+or if the CompilationContext is not set up correctly.
+
+- `F`: The Function (graph) to optimize. It will be transformed; if the caller
+  would like to keep an original copy of the Function it is responsible for
+  cloning it before calling `optimizeFunction()`.
+
+- `B`: The Backend for which we are optimizing `F`. It is able to control what
+  is lowered (`Backend::shouldLower()`), to specify what should be quantized
+  (`Backend::isOpSupported()`); and to transform the graph however it wishes
+  (`Backend::transformPostLowering()`).
+
+- `cctx`: The CompilationContext, containing important data structures to keep
+  some state before/after compilation is complete, as well as configuration
+  information for what kinds of transformations to perform.
+
+  - `PlaceholderBindings *bindings`: Mapping between Placeholders and Tensors
+    for this compilation run. For example, this is used when instrumenting a
+    Function for quantizaiton profiling, to create Placeholders and their
+    Tensors for inserted `QuantizationProfileNodes`.
+
+  - `LoweredInfoMap *loweredInfoMap`: Mapping from each Node output name in a
+    Function (corresponding to unique NodeValues in a Function), to a set of
+    `NodeNameAndKind` (which is a pair of Node output name and Kind). This is
+    used to keep track of what Nodes are lowered into what other Nodes, for use
+    during Profiling and Quantization.
+
+  - `enum class CompilationMode compMode`: This is set to either `Train` or
+    `Infer`, depending on what the Function is being used for. Some
+    optimizations may only be performed during one of these modes.
+
+  - `struct BackendOptions backendOpts`: Different options for compilation after
+    graph optimizations have been completed. [See
+    here](Backends.md#backendoptions-helper-struct) for more info.
+
+  - `struct PrecisionConfiguration precisionConfig`: Configuration for different
+    precision modes, used during `transformForPrecisionMode()`. Contains the
+    following:
+
+    - `enum class QuantizationMode quantMode`: One of `None`, `Quantize`, or
+      `Profile`, as previously mentioned above.
+
+    - `quantization::QuantizationConfiguration quantConfig`: Configuration for
+      quantization, including the quantizaion precision `ElemKind`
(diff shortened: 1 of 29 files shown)


include/glow/Backends/CompilationOptions.h
```diff
@@ -33,6 +35,30 @@ struct CompilationOptions {
 
   /// Insert TraceEvents between all instructions for profiling.
   bool autoInstrument{false};
+
+  // Different precision modes. Note that convertToFP16 and quantize can be
```
jackm321: doxygen here and other places

include/glow/Backends/CompilationOptions.h
```diff
@@ -33,6 +35,30 @@ struct CompilationOptions {
...
+  // specified simultaneously.
+  bool convertToFP16{false};
+  bool quantize{false};
+  bool profileForQuantization{false};
```
jackm321: quantize and profileForQuantization should be mutually exclusive right

lib/Backends/Backend.cpp
```diff
@@ -15,23 +15,75 @@
...
-  ::glow::lower(F, /* loweredMap */ nullptr, this);
+  // Lower the graph into a sequence of low-level linear algebra operations. If
+  // profiling, pass a nullptr for the backend, as all nodes should be lowered.
+  ::glow::lower(F, &opts.loweredInfoMap,
+                opts.profileForQuantization ? nullptr : this,
+                opts.precisionModeKindSet);
+
+  // Transform given precision mode; may quantize, convert to fp16, or
+  // instrument with profiling nodes. This must be done after lowering.
+  transformForPrecisionMode(*this, F, opts);
```
jackm321: Should transformForPrecisionMode be moved to global scope like the rest of the function here, there's no real reason to have it be in Backend.cpp right?

include/glow/Backends/CompilationOptions.h
jfix71: Correct -- I suppose I could replace them with an enum for QuantizationMode. I wanted to still allow fp16 conversion + quantization.

lib/Backends/Backend.cpp
jfix71: I guess if we want to unit test it then making it global would make sense; otherwise it's basically just a helper that was extracted to make the logic in `optimizeFunction()` more readable.

Additionally, I think it along with `optimizeFunction()` should be moved to GraphOptimizer.cpp (see https://github.com/pytorch/glow/pull/2703).

include/glow/Backends/Backend.h
```diff
@@ -102,7 +102,7 @@ class Backend {
   virtual bool shouldShareBuffers() const { return true; }
 
   /// Optimize the Function \p F given compilation options \p opts.
-  void optimizeFunction(Function *F, const CompilationOptions &opts) const;
+  void optimizeFunction(Function *&F, CompilationOptions &opts) const;
```
nadavrot: I don't think that passing a pointer to a reference is a very good interface.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -118,8 +118,8 @@ class ExecutionEngine final {
   /// Make \p networkName the function name for
   /// the entry point of the network and prepend all generated
   /// files with this name.
-  void save(Function *F, const CompilationOptions &opts,
-            llvm::StringRef outputDir, llvm::StringRef networkName);
+  void save(Function *&F, CompilationOptions &opts, llvm::StringRef outputDir,
```
nadavrot: Why did you remove the const? Are we modifying the compilation options inside? This will make it really difficult to track things around.

lib/Backends/Backend.cpp
```diff
@@ -15,23 +15,75 @@
...
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/PlaceholderBindings.h"
 #include "glow/IR/Instrs.h"
 #include "glow/Optimizer/Optimizer.h"
 
 using namespace glow;
 
-void Backend::optimizeFunction(Function *F,
-                               const CompilationOptions &opts) const {
+static void transformForPrecisionMode(const Backend &B, Function *&F,
```
nadavrot: Why do we need to pass the backend around? This really complicates things. We should not continue to increase the number of parematers to all of our optimizations functions. This is a sign that we don't have clean APIs.

lib/Backends/CMakeLists.txt
```diff
@@ -38,7 +38,8 @@ target_link_libraries(Backend
                         CodeGen
                         Graph
                         IR
-                        Optimizer)
+                        Optimizer
```
nadavrot: Why would the backends, who's job is to generate code, depend on the quantizer? The quantizer is just a procedure that mutates the graph.

nadavrot:
Jordan, this PR adds new parameters to functions, and new dependencies to libraries. This is an indication that the design becomes more complex and the interfaces less clean. Can we simplify things somehow?

include/glow/ExecutionEngine/ExecutionEngine.h
jfix71: The only thing that's being modified is the `LoweredInfoMap` -- I could move that to be a pointer and keep `CompilationOptions` `const`. Then the `LoweredInfoMap` would need to be explicitly kept in the caller's scope or higher.

lib/Backends/Backend.cpp
jfix71: The Backend is required in `optimizeFunction()` (and then into here) because it reports on what the specific Backend supports + what should be lowered (via `isOpSupported()` + `shouldLower()`), and it also allows the Backend to transform the graph (via `transformPostLowering()`).

It's passed in here because the quantizer only quantizes Nodes given what the Backend supports (via `isOpSupported()`). It's just a helper method here for `optimizeFunction()`.

One way we could consider redesigning this would be to refactor Backend into multiple classes, like a BackendPreferences class. But I don't see a way to not pass at least something for the backend in here right now.

lib/Backends/CMakeLists.txt
jfix71: Agreed, this was because the `optimizeFunction()` lives in the Backend class, which I don't think makes sense anyway. I put up a PR to move `optimizeFunction()` to Optimizer in https://github.com/pytorch/glow/pull/2703.

jfix71:
I've cleaned this up a bit here and through other PRs:
- `CompilationOptions` is again `const` when passed into `compile()` and beyond.
- `optimizeFunction()` is now in `Optimizer`, which is what now depends on `Quantization`/`Converter`.
- `compile()` returns the transformed `Function` (i.e. transformed after all passes for folding, optimizations, lowering, and precision transformation), instead of passing `F` into `compile()` as a reference to a pointer.

Some other modifications that might be a good idea (whether for this PR or not), as discussed offline with @opti-mix and @jackm321:
- Perhaps `compile()` should not return the transformed `F` -- one alternative here is to keep a pointer to `F` in `CompilationOptions` and update it, though this means a big refactor to set `F` in `opts` everywhere instead of passing it into `compile()`. 
    - Note that the current returning of `F` mirrors `quantizeFunction()` and `profileQuantization()` which also return the transformed `F`.
- Rename `CompilationOptions` as it now includes things like `PlaceholderBindings`, `KindSet`, `LoweredInfoMap`, etc. (perhaps `CompilationContext` or `CompilationConfiguration`?).

If anyone has further/other thoughts/opinions here, happy to hear them.

jfix71:
I've given this an initial refactoring on the `Loader`, `fr2en`, and `compareAgainstInterpreter()`:
- `CompilationContext` replaces `CompilationOptions`. It's a struct containing:
    - `CompilationMode` enum (`Infer` or `Train`)
    - `PlaceholderBindings` pointer for this compilation
    - `LoweredInfoMap` pointer for this compilation
    - `PrecisionConfiguration`, a struct containing:
        - `convertToFP16` bool flag
        - `QuantizationMode` (enum of `Quantize`/`Profile`/`None`)
        - `QuantizationConfiguration` (from #2751), used if `QuantizationMode::Quantize`
    - `BackendOptions`, a struct passed into `Backend::compile()` after `optimizeFunction()` is called, containing:
        - `collectConstants` bool flag
        - `autoInstrument` bool flag

RE: `compile()` returning the transformed `Function` -- I think it's the best option for now. It mirrors the `quantizeFunction()` and `profileQuantization()` calls which also return the Function. Some other options which I find less desirable: 
- Pass `F` as a pointer to a pointer and update `F` inside `compile()`/its callees -- this is similar to the initial version of a reference to a pointer. Don't really love this as an interface.
- Keep a pointer to the final Function in the `CompilationContext` -- don't love this either. Would require removing `const` when passing `CompilationContext` around.

include/glow/Optimizer/Optimizer.h
```diff
@@ -75,9 +75,9 @@ Function *profileQuantization(PlaceholderBindings &bindings, Function *F,
 std::unique_ptr<IRFunction> generateAndOptimizeIR(Function *F, const Backend &B,
                                                   bool shouldShareBuffers);
 
-/// Optimize the Function \p F given compilation options \p opts for Backend \B.
-void optimizeFunction(Function *F, const Backend &B,
-                      const CompilationOptions &opts);
+/// Optimize the Function \p F given compilation context \p cctx for Backend \B.
```
nadavrot: //  \returns ... 

It is not clear from the signature if the function is cloned and optimized or just modified in place.

nadavrot:
@jfix71 Thanks for doing this work. The changes to introduce CompilationContext look great. 

It is not clear from the API if we return a cloned function (like the quantizer does) or the original function. I don't see why we would like to return the original function, because we had just passed that in. On the other hand, I am not seeing a call to clone(). I think that cloning the function may block some optimizations because constants will have users from multiple functions and we will run into a problem where we can't delete the original function and need to keep the new function. On the other hand it's kind of strange that the optimizer changes the function inside compile().

jfix71:
>It is not clear from the API if we return a cloned function (like the quantizer does) or the original function. I don't see why we would like to return the original function, because we had just passed that in. On the other hand, I am not seeing a call to clone(). I think that cloning the function may block some optimizations because constants will have users from multiple functions and we will run into a problem where we can't delete the original function and need to keep the new function. On the other hand it's kind of strange that the optimizer changes the function inside compile().

@nadavrot It returns a clone if we profile/quantize, and otherwise returns the original function. Our APIs are inconsistent here -- clone for profile/quantize, do not clone for optimize/convertToFP16. I actually think we should remove cloning from all of them -- I think if the caller wants to keep a clone of the original Function it should be responsible for doing so, and understand the Function passed in will be mutated. That way `compile()` would still return nothing, and all our interfaces would be consistent. WDYT?

> To avoid merge conflicts is there a way to start landing the changes to CompilationContext that are ready?

Yeah I can move `CompilationContext` to a separate PR. I was putting it in here because many of the changes for `CompilationContext` will touch the same places this PR will, but it would probably be too big a PR to reasonably review.

nadavrot:
@jfix71 Removing cloning makes sense to me. The overall plan looks good. Let me know if you want me to stamp this PR or if you want to land the pieces in multiple PRs.

jfix71:
@nadavrot I'll do it in a few PRs. First one is up in #2792. Will file another one later removing cloning from quantize/profile.

jfix71:
OK, I think this is finally ready! Would love for some reviews so I can land this and stop rebasing it 😃

docs/GraphOptimizationPipeline.md
```diff
@@ -0,0 +1,151 @@
...
+  - `Profile`: Add special profiling nodes to the graph to gather a histogram of
+    values flowing through each Node in the graph. This can then be used to
+    automatically quantize the graph.
+
+  - `Quantize`: Quantize the graph, given a previously gathered profile as
+    mentioned above. This converts all Float inputs and outputs of each Node if
+    the backend supports the node as quantized.
+
+  - `ConvertToFP16`: Convert all Float inputs and outputs of Nodes to
+    Float16. Note that this does require any sort of profile. It can also be
```
gcatron: this does *not* require

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2720,16 +2722,75 @@ static llvm::Error checkAllNodesSupported(const Function &F, const Backend &B) {
...
+
+  if (precConfig.convertToFP16) {
+    TypeAToTypeBFunctionConverter converter(*F, ElemKind::FloatTy,
+                                            ElemKind::Float16Ty,
+                                            &precConfig.precisionModeKindSet);
+    converter.convert();
+  }
+}
+
 llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
```
jackm321: Given that the documentation is very specific about these optimizations, maybe add a note here to update the documentation when this function changes

include/glow/Optimizer/CompilationContext.h
```diff
@@ -19,20 +19,90 @@
...
+    Profile,  /// Add profiling nodes for quantization statistics gathering.
+  } quantMode{QuantizationMode::None};
+
+  /// Configuration for Quantization.
+  quantization::QuantizationConfiguration quantConfig;
+
+  /// Whether to convert the FloatTy to Float16Ty in the Function.
+  bool convertToFP16{false};
+
+  /// Used during Quantization, Profiling, and convertToFP16.
```
jackm321: Since this comment could be more helpful, probably mentioning something about this being kinds to exclude or something like that

lib/Onnxifi/InlineOnnxifi.cpp
```diff
@@ -57,20 +57,16 @@ InlineGraph::initGraph(const void *onnxModel, size_t onnxModelSize,
   computeModelHash(onnxModel, onnxModelSize, modelHash_);
   optimize(function_, CompilationMode::Infer);
 
-  // -- Profile --
```
jackm321: woo, glad this is gone!

PR #2701, opened 2019-04-11
Allocate memory for input tensor

yinghai:
*Description*:
This diff is wrote by @rdzhabarov. It's a mid-step where Glow will take care of the max batch size input. Later this step can be optimized away. 

*Testing*:
unit test. 
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -270,6 +270,32 @@ public:
     reset(t);
   }
 
+  void resize(const Type &T) {
+    // Do nothing if it's the same shape.
+    if (type_.dims().equals(T.dims())) {
+      return;
+    }
+
+    // Allocate memory for new tensor.
+    size_t count = T.size() * T.getElementSize();
+    auto newBuffer =
+        reinterpret_cast<char *>(alignedAlloc(count, TensorAlignment));
+
+    // Copy data from previous tensor to a new one.
+    size_t copySize = std::min(size(), T.size()) * type_.getElementSize();
+    std::copy(data_, data_ + copySize, newBuffer);
+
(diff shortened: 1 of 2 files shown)


include/glow/Base/Tensor.h
```diff
@@ -270,6 +270,32 @@ class Tensor final {
     reset(t);
   }
```
nadavrot: Doxygen comment please.

include/glow/Base/Tensor.h
```diff
@@ -270,6 +270,32 @@ class Tensor final {
...
 
+  void resize(const Type &T) {
```
nadavrot: I don't think that T is a meaningful name.

include/glow/Base/Tensor.h
```diff
@@ -270,6 +270,32 @@ class Tensor final {
...
+    // Do nothing if it's the same shape.
+    if (type_.dims().equals(T.dims())) {
+      return;
```
nadavrot: You don't copy the content from the previous tensor if it's the same size?

include/glow/Base/Tensor.h
```diff
@@ -270,6 +270,32 @@ class Tensor final {
...
+    }
+
+    // Allocate memory for new tensor.
```
nadavrot: You can use reset()

rdzhabarov:
This was patched to the fbcode to unblock some internal testing scenarios.

I do not think it's ideal to merge it here in the current state, let's make it better there first and then republish.

rdzhabarov:
I'll re-publish once we sort this out internally.

PR #2704, opened 2019-04-11
[onnxifi] Pad input tensor if it's smaller than the size of input placeholder.

rdzhabarov:
*Description*:
* Add padding to input tensor passed from C2 if needed.
* Back tensor by memory allocated by C2 if padding is not needed (should be majority of cases)
* Allocate new chunk of memory in case padding needed

*Testing*:
Tested on internal build. No good tests on onnxifi yet.

*Documentation*:
n/a

@yinghai @jackm321

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -109,9 +109,31 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
 
     auto &inPhPtr = inPhIt->getValue();
 
-    Tensor t(inOnnxBuffer, inPhPtr->getType());
+    std::vector<size_t> inOnnxTensorDims(inOnnxTensor.dimensions);
+    size_t inOnnxTensorSize = 1;
+    for (unsigned j = 0; j < inOnnxTensor.dimensions; ++j) {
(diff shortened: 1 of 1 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -109,9 +109,31 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+    // Otherwise just back the tensor by memory provided by the caller.
+    Tensor inputTensor;
+    if (inPhPtr->dims().equals(inOnnxTensorDims)) {
+      inputTensor = Tensor(inOnnxBuffer, inPhPtr->getType());
+    } else {
+      inputTensor = Tensor(inPhPtr->getType());
+      unsigned elementSize = inPhPtr->getType()->getElementSize();
+      char *onnxBuffer = static_cast<char *>(inOnnxBuffer);
+      std::copy(onnxBuffer, onnxBuffer + inOnnxTensorSize * elementSize,
+                inputTensor.getUnsafePtr());
```
yinghai: Is the rest of the memory zero-padded?

jackm321: Yes Tensor zeros everything on initialization

PR #2707, opened 2019-04-11
Moving method getOutputSave to the common glow code.

putivsky:
*Description*:
Adding getOutputSave method into common glow code.
*Testing*:
GraphTest
*Documentation*:
getOutputSave method will be used across specific backends implementations
[Optional Fixes #issue]
#2692

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -1081,6 +1081,10 @@ Function *differentiate(Function *F, const TrainingConfig &config,
                         llvm::StringRef newFuncName = "",
                         VariableGradientsList *varGrads = nullptr);
 
+/// \returns the first SaveNode user of the placeholder \p PH or
+/// nullptr if none are found.
+SaveNode *getOutputSave(Function *F, Placeholder *PH);
+
 /// Helper vectors for common transpose shuffles.
 #define NCHW2NHWC                                                              \
   { 0u, 2u, 3u, 1u }
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -2993,3 +2993,19 @@ bool Function::verify() const {
   }
   return isValid;
 }
+
+SaveNode *glow::getOutputSave(Function *F, Placeholder *PH) {
+  // if parent is set for PH, check if it is the same as provided Function.
+  auto *PHP = PH->getParent();
+  if (PHP != nullptr && F != PHP) {
+    return nullptr;
+  }
+  for (auto &use : PH->getUsers()) {
+    if (auto *save = llvm::dyn_cast<SaveNode>(use.getUser())) {
+      if (save->getPlaceholder() == PH) {
+        return save;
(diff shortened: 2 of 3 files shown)


include/glow/Graph/Graph.h
```diff
@@ -31,6 +31,7 @@
 
 namespace glow {
 class PlaceholderBindings;
+class SaveNode;
```
SplitInfinity: I think this is not necessary. The full definition of this class is already included above thanks to `Nodes.h`.

lib/Graph/Graph.cpp
```diff
@@ -2993,3 +2993,15 @@ bool Function::verify() const {
   }
   return isValid;
 }
+
+/// If \p PH is an output placeholder, \returns the SaveNode.
```
nadavrot: Doxygen should only be at the header file.

include/glow/Graph/Graph.h
```diff
@@ -1081,6 +1082,9 @@ Function *differentiate(Function *F, const TrainingConfig &config,
                         llvm::StringRef newFuncName = "",
                         VariableGradientsList *varGrads = nullptr);
 
+/// If \p PH is an output placeholder, \returns the SaveNode.
```
nadavrot: A better message:
/// \returns the first SAveNode user of the placeholder \p PH or nullptr if none are found.

lib/Graph/Graph.cpp
```diff
@@ -2993,3 +2993,15 @@ bool Function::verify() const {
...
+SaveNode *glow::getOutputSave(Placeholder *PH) {
```
rdzhabarov: this needs test in GraphTest

lib/Graph/Graph.cpp
```diff
@@ -2993,3 +2993,15 @@ bool Function::verify() const {
...
+  for (auto &use : PH->getUsers()) {
+    if (auto *save = llvm::dyn_cast<SaveNode>(use.getUser())) {
+      if (save->getPlaceholder() == PH) {
```
rdzhabarov: can a placeholder be used by multiple Save nodes (after we clone function for example)?

jfix71: Yeah, this function likely needs to take the Function under from you want to find the `SaveNode`.

lib/Graph/Graph.cpp
```diff
@@ -2993,3 +2993,19 @@ bool Function::verify() const {
   }
   return isValid;
 }
+
+SaveNode *glow::getOutputSave(Function* F, Placeholder *PH) {
+  // if parent is set for PH, check if it is the same as provided Function.
+  auto* PHP = PH->getParent();
```
nadavrot: Clang format please. You can run the script 

./utils/format.h  or git-clang-format on the staged changes.

nadavrot:
LGTM with the exception of clang-format.

PR #2710, opened 2019-04-12
Add command line options to set interpreter max mem and cpu max mem.

gcatron:
*Description*: This PR is a quality of life change and adds two new command line options, interpreter-mem and cpu-mem, which set the max memory for the interpreter and cpu device managers respectively. 
*Testing*: ninja test, also ran resnet-runtime with different values and noted that the partitioner could not find a valid partitioning if memory was too low.
*Documentation*: Do we document all of our flags somewhere?

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -16,7 +16,13 @@
 #include "CPUDeviceManager.h"
 #include "CPUFunction.h"
 
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
+
+static llvm::cl::OptionCategory CPUBackendCat("Glow CPU Backend Options");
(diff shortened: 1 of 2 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -16,14 +16,24 @@
 #include "CPUDeviceManager.h"
 #include "CPUFunction.h"
 
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
+static llvm::cl::OptionCategory CPUBackendCat("Glow CPU Backend Options");
+llvm::cl::opt<unsigned>
+    cpuMaxMem("cpu-memory",
+              llvm::cl::desc("Interpreter DeviceManager maximum memory"),
```
jfix71: `Interpreter -> CPU`

PR #2712, opened 2019-04-12
build: only support LLVM 8

compnerd:
Adjust the build to use the `REQUIRED` parameter to `find_package` as we want to
drop support for LLVM 6, and LLVM 8 is the first version available built with
the C++11 ABI.

Make the llvm-link search more resilient to the future by using the
LLVM_VERSION_MAJOR variable to add the suffix.  Additionally, when building
against the build tree, fetch the exported target for llvm-link and use that.

*Description*:
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,22 +88,7 @@ endif ()
 # Top level setup for external backends
 ExternalBackendsInit()
 
-# Prefer LLVM 7.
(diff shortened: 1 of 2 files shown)


compnerd:
@mortzur is going to take the patch and put it into her PR, so I'm just gonna close this off.

Issue #2709, opened 2019-04-12
Giv eeverybody the possibility to trigger the CI manually


tlepley-cadence:
I noticed that the CI is sometimes unstable (network access problem of the server for instance), resulting in failure of the PR check while the PR is correct.

The only way I found to retrigger it is to change something in the PR branch and push it. I don't think it's a convenient solution.

I'm wondering if there could be a more elegant solution like a button in the PR page.

rdzhabarov:
In the top right corner there is a rerun button (https://circleci.com/gh/pytorch/glow/11151?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Is that option available for you when you sign in with your github credentials?

PR #2723, opened 2019-04-16
Add support for ChannelShuffle node.

### LINKED ISSUE #2722: Add ChannelShuffle node
Lowering in Lower has few advantages:
1. C2 Importer seems not to be a good place to do that.
2.  Avoid reconstructing ChannelShuffle parameters from ReshapeNode, in GraphOptimizer. In addition, GraphOptimizer just wants to swap Transpose and Channel Shuffle, so it should not have to know the Channel Shuffle operator details.
3. Optimizing Reshape->Transpose->Reshape into ChannelShuffle in the Backend is possible, but can be considered as dependent on a sequence that might change.

vuzelac-cadence:
*Description*:
Add ChannelShuffle node. It's lowered in Lower.
*Testing*:
Related GraphOptimizer test is updated.

Fixes #2722

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -186,92 +186,25 @@ static Node *simplifyNode(Node *node, Function *F) {
   return node;
 }
 
-/// Parameters that are used to define ChannelShuffle operators.
-struct ChannelShuffleParams {
-  size_t group;
-  size_t kernel;
-};
-
-/// Compute the original parameters to the ChannelShuffle operator (represented
-/// as ReshapeNode->TransposeNode->ReshapeNode) for which \p node is the leading
-/// ReshapeNode. \returns The original ChannelShuffle parameters if possible and
-/// empty Optional otherwise.
-static llvm::Optional<ChannelShuffleParams>
-getChannelShuffleParams(const ReshapeNode &node) {
-  auto resM = llvm::Optional<ChannelShuffleParams>();
-
-  llvm::ArrayRef<size_t> inputDims = node.getInput().dims();
-  llvm::ArrayRef<size_t> resultDims = node.getDims();
-
-  // Check that there is one more output dimension than input dimension.
-  if (resultDims.size() != inputDims.size() + 1) {
-    return resM;
-  }
-
-  // Find the first output dimension that doesn't match its corresponding input
-  // dimension.
-  ChannelShuffleParams params;
-  bool found = false;
-  for (size_t i = 0, e = resultDims.size(); i < e - 1; ++i) {
-    if (inputDims[i] != resultDims[i]) {
-      params.kernel = i;
-      params.group = resultDims[i];
-      found = true;
-      break;
-    }
-  }
-
-  // Double check the property that the mismatched output found dimension and
-  // its successor together evenly multiply to the input dimension they
-  // mismatched on.
-  if (found && resultDims[params.kernel] * resultDims[params.kernel + 1] ==
-                   inputDims[params.kernel]) {
-    resM = params;
-  }
-
-  return resM;
-}
-
-/// Sink Transpose below ChannelShuffle node sequence ending with \p
-/// postShuffleRN. For example (Transpose_1->Reshape_1->Transpose_2->Reshape_2)
-/// becomes (Reshape_1->Transpose_2->Reshape_2->Transpose_1). \returns true if
-/// tranpose was sunk below ChannelShuffle node sequence and false otherwise.
+/// Sink Transpose below ChannelShuffle node.
 static bool sinkTranposeBelowChannelShuffle(Function *F,
-                                            ReshapeNode *postShuffleRN) {
-  auto *shuffleTR = dyn_cast<TransposeNode>(postShuffleRN->getInput());
-  if (!shuffleTR) {
+                                            ChannelShuffleNode *CS) {
+  auto *TR = dyn_cast<TransposeNode>(CS->getInput());
+  if (!TR) {
     return false;
   }
 
-  auto *preShuffleRN = dyn_cast<ReshapeNode>(shuffleTR->getInput());
-  if (!preShuffleRN) {
-    return false;
-  }
-
-  auto *sinkingTR = dyn_cast<TransposeNode>(preShuffleRN->getInput());
-  if (!sinkingTR) {
-    return false;
-  }
-
-  // Compute the original parameters to ChannelShuffle.
-  auto paramsM = getChannelShuffleParams(*preShuffleRN);
-
-  if (!paramsM.hasValue()) {
-    return false;
-  }
-
-  // Create a new ChannelShuffle with kernel parameter tranposed by the
-  // sinkingTR's shuffle because that Transpose will now be moved below this
+  // Create a new ChannelShuffle with kernel parameter transposed by the
+  // sinking TR's shuffle because that Transpose will now be moved below this
   // ChannelShuffle operator.
-  auto *newChannelShuffle = F->createChannelShuffle(
-      "channel_shuffle", sinkingTR->getInput(), paramsM->group,
-      sinkingTR->getShuffle()[paramsM->kernel]);
+  auto *newCS =
+      F->createChannelShuffle(CS->getName(), TR->getInput(), CS->getGroup(),
+                              TR->getShuffle()[CS->getKernel()]);
 
   // Create a copy of sinkingTR and insert after newChannelShuffle.
-  auto *newSinkingTR = F->createTranspose(
-      sinkingTR->getName(), newChannelShuffle, sinkingTR->getShuffle());
+  auto *newTR = F->createTranspose(TR->getName(), newCS, TR->getShuffle());
 
-  postShuffleRN->getResult().replaceAllUsesOfWith(newSinkingTR);
+  CS->getResult().replaceAllUsesOfWith(newTR);
 
   return true;
 }
@@ -442,9 +375,9 @@ static bool sinkCode(Function *F) {
       continue;
     }
 
-    if (auto *RN = dyn_cast<ReshapeNode>(node)) {
+    if (auto *CS = dyn_cast<ChannelShuffleNode>(node)) {
       // Sink Transpose below ChannelShuffle.
-      if (sinkTranposeBelowChannelShuffle(F, RN)) {
+      if (sinkTranposeBelowChannelShuffle(F, CS)) {
         changed = true;
         continue;
       }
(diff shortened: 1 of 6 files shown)


rdzhabarov:
Even though this adds an additional node to the node gen, this makes some of the optimizations much cleaner.

@vuzelac-cadence what was the main motivation for the change? Is that because this could be directly supported efficiently on hardware?

vuzelac-cadence:
We have an efficient implementation of ChannelShuffle that's a bit different than Transpose. In addition, we just can't have Reshapes around as they are limiting our optimization efforts.
But, the change is done in Glow, rather than in our Backend, just because I though recognizing sequences likes Reshape->Transpose->Reshape is a bit too patchy. Not recognizing per se, but the need to reconstruct group/kernel, that we would have to do just like GraphOptimizer was doing.

vuzelac-cadence:
continuous-integration/travis-ci/pr failure:
```
CMake Error at CMakeLists.txt:97 (message):
  LLVM minimum required version is 7.  LLVM package version found: .
```

It would be good to have ability to manually trigger CI. See #2709

rdzhabarov:
Can you trigger CI from the circleci page? (there is rebuild option, i just triggered)

rdzhabarov:
> rather than in our Backend, just because I though recognizing sequences likes Reshape->Transpose->Reshape is a bit too patchy

That sequence appeared on the shufflenet, iirc.

I think having separate node would make sense and let backend decide whether lower or not.
I'd keep the previous optimization in place and add the new one (sinking transpose below channelshuffle node).

jfix71:
>I'd keep the previous optimization in place and add the new one (sinking transpose below channelshuffle node).

I think this actually would be a great fit for the `fold()` pass. We fold it back to the Node, and then can let the GraphOptimizer focus on optimizing it at a higher level, then later on we lower it again.

vuzelac-cadence:
> Can you trigger CI from the circleci page? (there is "rebuild/rerun workflow" option, i just triggered)

That button is shaded out, saying that I don't have write permissions. Above the button is my Github Avatar and on the left side says Pytorch (should be the organization). My github is authorized with CI.

vuzelac-cadence:
> > rather than in our Backend, just because I though recognizing sequences likes Reshape->Transpose->Reshape is a bit too patchy
> 
> That sequence appeared on the shufflenet, iirc.

Don't see in the original paper. But, since the channel shuffle is exactly that, my only problem with it was the reconstruction of the node in GraphOptimizer - just should not be there.

vuzelac-cadence:
> > I'd keep the previous optimization in place and add the new one (sinking transpose below channelshuffle node).
> 
> I think this actually would be a great fit for the `fold()` pass. We fold it back to the Node, and then can let the GraphOptimizer focus on optimizing it at a higher level, then later on we lower it again.

Want to make sure what exactly is to be done here: 
GraphOptimizer.cpp - understood - just add the original optimization along with the new one.
GraphOptzTest - original code can't be added as it was just calling `F->createChannelShuffle()` . Seems it needs a new sequence (RN->TR->RN) that would trigger the existing optimization ?

jfix71:
>GraphOptimizer.cpp - understood - just add the original optimization along with the new one.

Actually I think we just should keep the new optimization and drop the original. The old logic was basically to get around the issue of not having the unlowered version, and can be moved/refactored into `fold()`. Because we will fold it, I don't think we need to have both paths in the GraphOptimizer.

>GraphOptzTest - original code can't be added as it was just calling F->createChannelShuffle() . Seems it needs a new sequence (RN->TR->RN) that would trigger the existing optimization ?

I think the change you made that verifies that the new version of the optimization works is good. I think you should also add a `GraphFold` test (see the couple that already exist in `GraphOptzTest.cpp`) that checks that the sequence we are going to fold into the `ChannelShuffleNode` is in fact correctly folded.

vuzelac-cadence:
> > GraphOptimizer.cpp - understood - just add the original optimization along with the new one.
> 
> Actually I think we just should keep the new optimization and drop the original. The old logic was basically to get around the issue of not having the unlowered version, and can be moved/refactored into `fold()`. Because we will fold it, I don't think we need to have both paths in the GraphOptimizer.

Something is wrong here. I am not looking for more work but still want to avoid breaking anything - comment in GraphTest says you encountered Reshape->Transpose->Reshape in ShuffleNet. If so, removing this optimization would break your ShuffleNet, no ? (hopefully not :) ).  Curious where that ShuffleNet comes from ? TensorFlow one for sure will include RN->TR->RN as both TF and ONNX don't support ChannelShuffle. But that model should fail to compile in Glow due to other problems before it could even fail here.

vuzelac-cadence:
> I think the change you made that verifies that the new version of the optimization works is good. I think you should also add a `GraphFold` test (see the couple that already exist in `GraphOptzTest.cpp`) that checks that the sequence we are going to fold into the `ChannelShuffleNode` is in fact correctly folded.

I didn't get this one - that test calls `glow::fold() `which calls optimizeReshape() and foldLeakyRelu() only - it'll not fold `ChannelShuffle`. But also, I am not sure how the existing test is not sufficient: sinkTransposeBelowChannelShuffleNodesAndEliminate does seem to trigger the folding and checks if it's done correctly (?).

vuzelac-cadence:
Failing CI test (fails in OperatorTest) seems not to be related to my change:

`*** Error in `/var/lib/jenkins/workspace/build/tests/OperatorTest': double free or corruption (!prev): 0x0000000003f45ef0 ***`


`

jfix71:
@vuzelac-cadence Right now we have the following situation. Our Caffe2 Shufflenet sees it is importing a `ChannelShuffle` and calls `createChannelShuffle()` which creates Reshape-Transpose-Reshape (i.e. it's lowered directly at import time). This is why we later have the graph optimization pass `sinkTranposeBelowChannelShuffle()` which looks for this pattern along with a transpose above it and sinks the transpose below it. Changing `createChannelShuffle()` to create a `ChannelShuffleNode` means this optimization pass is no longer necessary, given the new version in this PR.

What I am suggesting for the new version with the newly added `ChannelShuffleNode` is to add a new pass to `fold()`, which is similar to the logic in `sinkTranposeBelowChannelShuffle()`, except it does no transpose sinking. It still looks for Reshape-Transpose-Reshape, but then simply and replaces it by the `ChannelShuffleNode` you are introducing. This fold pass was recently introduced by @tlepley-cadence. It occurs before any optimizations or transformations to the graph. Its purpose is to make analysis and optimizations of graphs much easier, for purposes exactly like this, where we might load graphs that have pre-lowered sequences and we'd prefer to see them unlowered.

>TensorFlow one for sure will include RN->TR->RN as both TF and ONNX don't support ChannelShuffle.

This should solve the issue for TF/ONNX, as the RN->TR->RN sequence will be folded into a `ChannelShuffleNode` before any optimizations/transformations. Then the `sinkTranposeBelowChannelShuffle()` pass will work correctly as implemented already in this PR. Note that we don't need this folding logic for our Caffe2 Shufflenet after this PR, as we will directly create the `ChannelShuffleNode` when importing -- it's only beneficial for the ONNX/TF version.

>I didn't get this one - that test calls `glow::fold() `which calls `optimizeReshape()` and `foldLeakyRelu() `only - it'll not fold `ChannelShuffle`. But also, I am not sure how the existing test is not sufficient: `sinkTransposeBelowChannelShuffleNodesAndEliminate `does seem to trigger the folding and checks if it's done correctly (?).

I'm suggesting you add a new `foldChannelShuffle()` pass to `fold()`, as I described above, looking for RN->TR->RN and replacing it with a `ChannelShuffleNode`. The `sinkTransposeBelowChannelShuffleNodesAndEliminate` test is totally find to keep as well. I'm saying when you add the new `fold()` logic you should add a test to verify it correctly folds a RN->TR->RN sequence into a `ChannelShuffleNode`. It should be pretty simple to create.

jackm321:
As the author of the original `sinkTranposeBelowChannelShuffle()` I 100% agree with @jfix71 that we should kill it and replace it by folding Reshape-Transpose-Reshape into a `ChannelShuffleNode` and using the logic from this PR instead.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2089,6 +2089,32 @@ TEST_F(GraphFold, foldLeakyReluFromConst) {
...
+  ::glow::fold(F_, CompilationMode::Infer);
+
+  ASSERT_EQ(F_->getNodes().size(), 2);
+
+  // Check for ChannelShuffle node.
+  auto *CS = llvm::dyn_cast<ChannelShuffleNode>(save->getInput().getNode());
+  ASSERT_NE(nullptr, CS);
+
+  // Ensure ChannelShuffle node has the same dimensions as the input.
+  EXPECT_EQ(CS->getType(0)->dims(), llvm::makeArrayRef(inputDims));
```
jfix71: nit: `CS->getResult().dims()` here

jfix71: Can you also check that group and kernel are set as expected?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2707,37 +2733,20 @@ TEST_F(GraphOptz, sinkTransposeBelowChannelShuffleNodesAndEliminate) {
...
-  EXPECT_EQ(preShuffleRN->getDims(),
-            llvm::makeArrayRef(expectedPreShuffleRNDims));
+  ASSERT_EQ(F_->getNodes().size(), 2);
+
+  // Check that the channel shuffle node is still there.
+  auto *CSN = llvm::dyn_cast<ChannelShuffleNode>(save->getInput().getNode());
+  ASSERT_NE(nullptr, CSN);
+
+  // Ensure ChannelShuffle node has the same dimensions as the input.
+  EXPECT_EQ(CSN->getType(0)->dims(), llvm::makeArrayRef(inputDims));
```
jfix71: Same as in other test: `CS->getResult().dims()` here, and check that group and kernel are set as expected?

lib/Optimizer/Lower.cpp
```diff
@@ -641,6 +641,36 @@ static void lowerTileNode(Function *F, LoweredInfoMap *loweredMap,
...
+  assert(D % group == 0);
+
+  dims.erase(dims.begin() + kernel);
+  // Reshape {D1, ... D_k, ... D_n} -> {D1, ... group, D_k / group, ... D_n}
+  dims.insert(dims.begin() + kernel, D / group);
+  dims.insert(dims.begin() + kernel, group);
+  auto *R1 = F->createReshape(CSN.getName().str() + ".reshape1", input, dims);
+
+  std::vector<unsigned_t> transpose(dims.size());
+  for (size_t i = 0; i < transpose.size(); i++)
```
jfix71: nit: I see this was copied from before, but we prefer curly brackets even on single line nests.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2512,6 +2445,87 @@ static void foldLeakyRelu(Function *F) {
...
+  return resM;
+}
+
+// Fold Reshape->Transpose->Reshape into ChannelShuffle when applicable.
+static void foldChannelShuffle(Function *F) {
+
+  auto &nodes = F->getNodes();
+  for (auto &node : nodes) {
+    auto *RN2 = dyn_cast<ReshapeNode>(&node);
+    if (!RN2 || RN2->getDims().size() != 4) {
```
jfix71: I don't think it's a requirement that the ChannelShuffle input has 4 dimensions, is it? I think we can drop the checks for the dims size and just let `getChannelShuffleParams()` handle the logic for checking if it's a valid ChannelShuffle.

lib/Optimizer/Lower.cpp
vuzelac-cadence: @jfix71, note that checking the kernel picked up a bug in setting kernel field in `createChannelShuffle` in  `foldChannelShuffle()`.

PR #2725, opened 2019-04-16
Allow zero sized tensor inputs

yinghai:
*Description*:
It is possible that input tensor descriptor has a null buffer with dim [0]. For example, we can look up 0 elements in embedding table in current batch. We need to allow this case. Also fix a bug where we return incorrect error code. 

*Testing*:
Internal test. 

*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -127,6 +127,8 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
       inputTensor = Tensor(inOnnxBuffer, inPhPtr->getType());
     } else {
       char *onnxBuffer = static_cast<char *>(inOnnxBuffer);
+      // If input onnxTensorDescriptor has a NULL buffer pointer, which is a
+      // valid case for empty tensor, skip copying
       if (inOnnxBuffer) {
         inputTensor = Tensor(inPhPtr->getType());
         unsigned elementSize = inPhPtr->getType()->getElementSize();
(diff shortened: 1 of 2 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -405,8 +405,8 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxInitGraph)(
   return ONNXIFI_STATUS_SUCCESS;
 }
 
-static bool verifyDescriptors(uint32_t count,
-                              const onnxTensorDescriptorV1 *descriptors) {
+static int verifyDescriptors(uint32_t count,
```
nadavrot: This is a good opportunity to add a doxygen comment.

lib/Onnxifi/Base.cpp
```diff
@@ -126,11 +126,13 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
     if (inPhPtr->dims().equals(inOnnxTensorDims)) {
       inputTensor = Tensor(inOnnxBuffer, inPhPtr->getType());
     } else {
-      inputTensor = Tensor(inPhPtr->getType());
-      unsigned elementSize = inPhPtr->getType()->getElementSize();
       char *onnxBuffer = static_cast<char *>(inOnnxBuffer);
-      std::copy(onnxBuffer, onnxBuffer + inOnnxTensorSize * elementSize,
-                inputTensor.getUnsafePtr());
+      if (inOnnxBuffer) {
```
nadavrot: Comment please.

PR #2699, opened 2019-04-10
removing support in llvm 6 and below

mortzur:
*Description*:
Removing code support in LLVM package versions <= 6.
*Edited*: CI scripts were modified to use LLVM version 8.
The reason for skipping LLVM-7 is a problematic llvm-7 distribution for Ubuntu-16.04 (xenial), which has been built with old C++ ABI and therefore is incompatible. Locally (MacOS) tests pass also on llvm-7.   

Should be followed up (in next commit) by removing all FACEBOOK_INTERNAL branches (with the specific patch number).

*Testing*:
* cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ../glow
   -- Found LLVM 7.0.1
    ...
   ninja test

* cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ../glow -DCMAKE_PREFIX_PATH=/usr/local/Cellar/llvm/8.0.0/
-- Found LLVM 8.0.0
   ...
   ninja test

* On a devserver (changes copied manually, llvm version 8):
   buck build //glow/...
   buck test //glow/...

*Documentation*:

Fixes #2560

PR #2726, opened 2019-04-16
[CMakeList] Allow for LLVM 7 with any minor version

jfix71:
I have LLVM 7.1 on my laptop. #2699 was requiring 7.0 specifically. This fix seems to work. I'm no CMake master so let me know if this makes sense or if there's a better way 😃

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,13 @@ if (NOT LLVM_FOUND)
   find_package(LLVM 7 CONFIG)
 
   if (NOT LLVM_FOUND)
-    message(FATAL_ERROR "LLVM minimum required version is 7. LLVM package version found: ${LLVM_PACKAGE_VERSION}.")
+    # Fallback to whatever is available.
+    find_package(LLVM CONFIG)
+
+    if (NOT LLVM_FOUND OR LLVM_PACKAGE_VERSION VERSION_LESS "7.0")
+      message(FATAL_ERROR "LLVM minimum required version is 7. LLVM package version found: ${LLVM_PACKAGE_VERSION}.")
+    endif()
+
   endif()
 endif()
 


CMakeLists.txt
```diff
@@ -90,10 +90,10 @@ ExternalBackendsInit()
...
 find_package(LLVM 8 CONFIG)
 if (NOT LLVM_FOUND)
-  # Fallback to LLVM 7
-  find_package(LLVM 7 CONFIG)
+  # Fallback to whatever is available.
+  find_package(LLVM CONFIG)
 
-  if (NOT LLVM_FOUND)
+  if (NOT LLVM_FOUND OR LLVM_VERSION VERSION_LESS "7.0")
     message(FATAL_ERROR "LLVM minimum required version is 7. LLVM package version found: ${LLVM_PACKAGE_VERSION}.")
```
arunm-git: Intentional to have LLVM_VERSION in the if(...) and LLVM_PACKAGE_VERSION in the message?

arunm-git:
Is it really 7.1 on your laptop (and not 7.0.1)? I ask because I didn't realize there was a 7.1 release (goes from 7.0.1 to 8.0).

From LLVMConfig.cmake, 
```
# LLVM is API-compatible only with matching major.minor versions
# and patch versions not less than that requested.
```
So, if you really have 7.1, the failure is expected since we are asking for 7 (which translates to 7.0).
And your fix seems fine to me to establish >= 7.0.

arunm-git:
Actually I might have gotten that wrong - the CMakeFile as written allows proceeding with a lower than 7 version and your change enforces >=7 (which is why the travis-ci is failing).
Now I'm puzzled why 7.1 (or 7.0.1) fails on your laptop...

jfix71:
>Is it really 7.1 on your laptop (and not 7.0.1)? I ask because I didn't realize there was a 7.1 release (goes from 7.0.1 to 8.0).

@arunm-git  It seems to be -- I was only basing this on the error message I'm getting when trying to build locally:

```
The following configuration files were considered but not accepted:
   /Users/jfix/glow/llvm/llvm_install/lib/cmake/llvm/LLVMConfig.cmake, version: 7.1.0
```

arunm-git:
> Actually I might have gotten that wrong - the CMakeFile as written allows proceeding with a lower than 7 version and your change enforces >=7 (which is why the travis-ci is failing).
> Now I'm puzzled why 7.1 (or 7.0.1) fails on your laptop...
Never mind : I wasn't reading the code carefully. Sorry for the noise.

Your updated commit looks fine to me.
CC: @mortzur who made a bunch of cmake changes recently.

CMakeLists.txt
```diff
@@ -94,7 +94,13 @@ if (NOT LLVM_FOUND)
   find_package(LLVM 7 CONFIG)
 
   if (NOT LLVM_FOUND)
-    message(FATAL_ERROR "LLVM minimum required version is 7. LLVM package version found: ${LLVM_PACKAGE_VERSION}.")
+    # Fallback to whatever is available.
+    find_package(LLVM CONFIG)
+
+    if (NOT LLVM_FOUND OR LLVM_PACKAGE_VERSION VERSION_LESS "7.0")
+      message(FATAL_ERROR "LLVM minimum required version is 7. LLVM package version found: ${LLVM_PACKAGE_VERSION}.")
```
rdzhabarov: can you close this by endif().

PR #2728, opened 2019-04-16
[Provisioner] Remove NETWORK_PADDING_FACTOR

beicy:
*Description*:
No need to use the factor. NETWORK_PADDING_FACTOR was added to reflect the memory use of the instruction. 

*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Runtime/Provisioner/Provisioner.cpp
+++ b/lib/Runtime/Provisioner/Provisioner.cpp
@@ -103,8 +103,7 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
 
   // Try to add functions to devices in order from largest to smallest.
   for (unsigned i = 0; i < logicalDeviceSize.size(); i++) {
-    RETURN_ERR_IF_NOT(logicalDeviceSize[i].second * NETWORK_PADDING_FACTOR <
-                          deviceMemory[i].second,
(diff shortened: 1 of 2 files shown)


lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -103,8 +103,7 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
 
   // Try to add functions to devices in order from largest to smallest.
   for (unsigned i = 0; i < logicalDeviceSize.size(); i++) {
-    RETURN_ERR_IF_NOT(logicalDeviceSize[i].second * NETWORK_PADDING_FACTOR <
-                          deviceMemory[i].second,
+    RETURN_ERR_IF_NOT(logicalDeviceSize[i].second < deviceMemory[i].second,
```
gcatron: It wasn't discussed earlier but I think we could change this line to use the deviceManager->isMemoryAvailable method. That way we still factor in the cost of instruction but can still remove the clunky padding factor.

gcatron: Actually this is good as is, since we getAvailableMemory from the DeviceManager we are good.

PR #2731, opened 2019-04-16
[Runtime] runtime e2e multi-cpu test.

beicy:
*Description*:
This PR uses en2gr to run runtime e2e multi-cpu test. 

*Testing*:
 ./bin/en2gr-runtime -cpu-memory=400000000 --- This we will get 3 partitions.
 ./bin/en2gr-runtime -cpu-memory=500000000 -- This we will get 2 partitions.
Use "I love music." as input, and the result is:
```
Initializing 5 CPU Devices on HostManager.
Loading en2gr model.
Adding to HostManager
ich liebe Musik .

Finished!
```
*Documentation*:
[Optional Fixes #issue]


nadavrot:
Very cool!  

I am worried that we now have two versions of the e2e test. Can we merge both examples into one?

beicy:
> Very cool!
> 
> I am worried that we now have two versions of the e2e test. Can we merge both examples into one?

To make it simple, in this test, I just hardcoded the input "I love music ."  I will modify the input to what text-translator does, and then I think we can remove en2gr_cpu_test.sh from our expensive test list.

nadavrot:
Why do we need two programs that translate between English and german? Can we delete the other program? I. M not worried about the test harness here. I am worried that this test is a copy-paste if the other two hundred line program.

beicy:
> Why do we need two programs that translate between English and german? Can we delete the other program? I am not worried about the test harness here. I am worried that this test is a copy-paste of the other two hundred line program.

I see your point now. OK, I will merge those 2 files.

nadavrot:
Keep alive.

examples/en2gr-runtime.cpp
```diff
@@ -0,0 +1,153 @@
...
+            bindings->get(outputTokenBeamList),
+            bindings->get(outputScoreBeamList),
+            bindings->get(outputPrevIndexBeamList));
+
+        finished.set_value();
+      });
+  (void)runid;
+}
+
+/// Run ResNet concurrently on the number CPU Devices provided by the user.
```
rdzhabarov: need to fix comment.

examples/en2gr-runtime.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+namespace {
+llvm::cl::OptionCategory category("en2gr-runtime Options");
+llvm::cl::opt<unsigned> numDevices("numDevices",
+                                   llvm::cl::desc("Number of Devices to use"),
+                                   llvm::cl::init(5), llvm::cl::value_desc("N"),
+                                   llvm::cl::cat(category));
+} // namespace
+
+/// Builds a simple graph, returns back input var and save node through refs.
```
rdzhabarov: it's hard to see how it returns back input var and save node ..

beicy: updated the comment.

nadavrot:
I am worried about having two versions of each test, one with the new runtime (example en2gr-runtime.cpp) and one with the legacy execution engine (text-translator ). How do we merge this tests back into one?

beicy:
@nadavrot I've abstract the common part of the test as a header file, and we don't have the duplicated implementation now. Since the original test also handle the profiling option, I don't think it will better to totally combine the test files.

nickgg:
@beicy can we just remove the text translator binary now? Is it obsolete?

jfix71:
Given my understanding of this PR and #2799 from @gcatron, I'm not sure we need `en2gr-runtime` at all -- we can modify the `text-translator` similar to the way #2799 modifies `image-classifier` so that it also uses the runtime. My understanding is we will be removing `resnet-runtime` once #2799 lands. If all this is correct I'd rather not refactor `text-translator` like this as I think we'd remove `en2gr-runtime` too after #2799.

beicy:
> @beicy can we just remove the text translator binary now? Is it obsolete?

@nickgg I don't think so. The text binary is used to test profiling as well.

beicy:
@jfix71 Thanks for pointing out this. I will close this PR. Once #2279 is landed, I will work on text-translator.

PR #2668, opened 2019-04-06
[Executor] Fix placeholder propagation bug

SplitInfinity:
**Description**
This commit fixes a bug in which intermediate outputs generated by partitions
are not propagated to their children. This is because the executor
assumes that DeviceManager::runFunction will return a set of PlaceholderBindings
containing the output Placeholders, but this is not true.
DeviceManager::runFunction needs to be invoked with PlaceholderBindings
that contain Tensors for both input and output Placeholders.

This commit fixes this bug by creating Placeholders for all intermediate
DAG node outputs, storing them in the ExecutionState corresponding
to the run that needs them, and mapping them to zero Tensors in the
input bindings for the DAG node that produces them.

The executor unit test has been modified accordingly.

**Testing**
All unit tests pass.

```
$ ninja all && ninja test
ninja: no work to do.
[0/1] Running tests...
Test project /Users/meghanl/fbsource/glow/build_DEBUG
      Start  1: BackendCorrectnessTest
 1/34 Test  #1: BackendCorrectnessTest ..............   Passed   18.96 sec
      Start  2: BackendTest
 2/34 Test  #2: BackendTest .........................   Passed    2.55 sec
      Start  3: BasicIRTest
 3/34 Test  #3: BasicIRTest .........................   Passed    0.04 sec
      Start  4: Caffe2ImporterTest
 4/34 Test  #4: Caffe2ImporterTest ..................   Passed    3.23 sec
      Start  5: ConvTest
 5/34 Test  #5: ConvTest ............................   Passed    5.47 sec
      Start  6: DeviceManagerTest
 6/34 Test  #6: DeviceManagerTest ...................   Passed    0.86 sec
      Start  7: Float16Test
 7/34 Test  #7: Float16Test .........................   Passed    0.01 sec
      Start  8: GemmTest
 8/34 Test  #8: GemmTest ............................   Passed    0.06 sec
      Start  9: GlowOnnxifiManagerTest
 9/34 Test  #9: GlowOnnxifiManagerTest ..............   Passed    0.38 sec
      Start 10: GradCheckTest
10/34 Test #10: GradCheckTest .......................   Passed    4.16 sec
      Start 11: OperatorGradTest
11/34 Test #11: OperatorGradTest ....................   Passed    0.04 sec
      Start 12: GraphGradTest
12/34 Test #12: GraphGradTest .......................   Passed    0.04 sec
      Start 13: GraphOptzTest
13/34 Test #13: GraphOptzTest .......................   Passed    0.02 sec
[...]

PR #2734, opened 2019-04-16
[habana] Fall back to looking up Placeholders by name

nickgg:
*Description*: (backport from glow-h) Since #2668 we may have copied placeholders with the same name but a different pointer. Sometimes we need to fall back to searching by name.
*Testing*: sanity unittests, but really tested in glow-h
*Documentation*: I don't really like this change and would love an alternative.


nadavrot:
> " Sometimes we need to fall back to searching by name."

I am worried about this part. Why do we have two ways of searching for placeholders? Why do we have multiple placeholders with the same name? They should be uniqued by the module.

nickgg:
@nadavrot I agree, consider this a temporary fix.

rdzhabarov:
This PR is the upstream from internal repo. If we want to fix this we'd need to fix there first to avoid divergence.

PR #2739, opened 2019-04-17
[Runtime Efficiency] Map placeholders by integer id instead of strings in PlaceholderBindings

jackm321:
*Description*:
Instead of storing Placeholders by their name in PlaceholderBindings, give each Placeholder a id and store them based on that mapping. We have seen in the past that the runtime cost of querying and updating placeholder bindings by string keys is huge, using integers should be a bit cheaper.

*Testing*:
* sync and test internally

*Documentation*:
doygen


include/glow/Backends/BackendUtils.h
```diff
@@ -37,6 +37,8 @@ struct RuntimeSymbolInfo {
   size_t size{0};
   /// Offset in bytes from the base address.
   size_t offset{0};
+  /// TODO: comment
```
nadavrot: Let's do it.

include/glow/Graph/Nodes.h
```diff
@@ -108,21 +108,35 @@ class Constant : public Storage {
   bool verify() const;
 };
 
+uint32_t getNewPlaceholderId();
```
nadavrot: Why is this a part of the Node? I would expect the mapping to be done by the Module/context or some other centralized place.

include/glow/Graph/Nodes.h
```diff
@@ -108,21 +108,35 @@ class Constant : public Storage {
...
+
 /// Placeholder nodes are unbound-storage. The content tensors are attached to
 /// this node at runtime. Placeholders are used as inputs and output nodes to
 /// the network.
 class Placeholder : public Storage {
   /// Specifies if the placeholder is trainable.
   bool isTrainable_;
+  // TODO: comment
```
nadavrot: I think that this should go in the module. The module will be able to unique and translate between the ID and the pointer.

lib/Graph/Node.cpp
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <atomic>
```
nadavrot: I don't think that a global variable is a good idea. I think that it should be a member of Module/Context.

jackm321: agreed, this is still WIP

jackm321:
@nadavrot creating unique placeholder ids in the module makes sense to me, I was originally planning on doing this but unfortunately ExecutorTest creates it's own Placeholders outside of a module so I need to figure out how to deal with that case. The only other case where this happens is in `ThreadpoolExecutor` but in that case Placeholder ids should be propagated from one ExecutionContext to another, no problem there.

As for where to store placeholder ids, I think it makes sense to store them in the placeholder itself because this can be though of as analogous to the name that's already in placeholder but for use in the runtime instead of the compiler. The alternatives I see to storing it in the Placeholder itself are
* keep some mapping in module, this would force this data to stick around in module during run time which isn't great and we'd have to propagate the module to the Executor and maybe other parts of the runtime too.
* only keep the mapping in PlaceholderBindings and somehow change the runtime design so that contexts aren't consumed as they are passed down to device manager for running. This one I don't think will be an easy change given how the runtime is currently setup. I also don't think it will work because we would still need a global mapping in the Module for the final copy back to output placeholders.
If other have ideas on how these could be made to work I'm interested in hearing though.

beicy:
If we use id, shall we allow phs have the same name?

jackm321:
@beicy I'm not sure if I understand the question but same name placeholders should still disallowed by the Module.

The way I see this is that basically have a unique id for placeholders in the compiler, their name, but in the run time we want integer unique id so that we don't have to spend so much time hashing string names. Also, for reasons that I don't fully understand, Placeholders are created and owned by the Executor for intermediate values outside of any Module and these placeholders correspond to existing Placeholders in the module thus their names and ids should match those and therefore are in fact not unique but they are meaningful (which prevents just using Placeholder* as unique placeholder id in the runtime).

nadavrot:
@jackm321 Placeholders are always owned by the module. You should never be allowed to create a placeholder without a module. How else would we unique their names? Name uniquing and id uniquing are in fact the same thing.  Maybe we can even use the pointer to the name in the string table as the ID.

jackm321:
@nadavrot this is not the case right now, `ExecutionState::createOrGetPlaceholder` is creating Placeholders outside of Modules. Maybe we should fix that issue and solution to this perf issue should be straightforward afterwards

beicy:
> @nadavrot this is not the case right now, `ExecutionState::createOrGetPlaceholder` is creating Placeholders outside of Modules. Maybe we should fix that issue and solution to this perf issue should be straightforward afterwards

After graph partition, new placeholders are created, but not registered in PlaceholderBindings, so Executor will do this work.  But I think that placeholders should be owned by the module?

nadavrot:
@beicy @jackm321 Thank you for explaining this to me. I still don't understand something. If the partitioner is creating the placeholders then that's good. But I am seeing that the ExecutionState is creating new placeholders. Why is that? 

I understand that before executing the program we need to insert some backing tensor to PlaceholderBindings but why do we need to create new placeholders? Perhaps we need new Tensors, not new placeholders. I must be missing something.

nickgg:
Agree with @beicy, I think. Why does it matter whether or not the Placeholders are in the original PlaceholderBindings if they are in the Module?

jackm321:
Thanks for the comments everyone. @nadavrot no I don't think you're missing something and @nickgg it sounds like everyone agrees, placeholders _should_ be created by the Partitioner and added to the Module and not created adhoc by the Executor. We just need to actually make that happen because that isn't what's going on today. This should be a relatively easy thing to fix.

jackm321:
Will rework this now that the issue above has been fixed by #2754

jackm321:
Abandoning in favor of using Placeholder* #2771

PR #2740, opened 2019-04-17
[LLVMIRCodeGen] Initialize all asm parsers

opti-mix:
This solves the issue with  "LLVM ERROR: Inline asm not supported by this streamer because we don't have an asm parser for this target" reported when compiling libjit by some backends.


opti-mix:
@nickgg I see that DEBUG config on CI fails because of a double free error in tests/TraceEventsTest. There is no way my change could have caused it as far as I can see. Could it be that our tests/TraceEventsTest is flaky or we have a real issue with tracing?

jfix71:
The DEBUG test became flaky when we moved to LLVM 8 on CI.

PR #2743, opened 2019-04-17
Adds network duplication support to the Provisioner

gcatron:
*Description*: This PR adds the logic to the provisioner to support loading a network on multiple devices. This is done by assigning more than one logicalDevice to the DAGNode for the function.
*Testing*: Extended the Provisioner test to duplicate networks. 
*Documentation*: NA

### DIFF
--- a/lib/Runtime/Provisioner/Provisioner.cpp
+++ b/lib/Runtime/Provisioner/Provisioner.cpp
@@ -55,15 +55,17 @@ Provisioner::Provisioner(DeviceManagerMapTy &devices) {
 llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
   // Walk the networks and group by logicalDeviceId.
   std::map<DeviceIDTy, std::vector<DAGNode *>> logicalDevices;
-
+  // For each network visit all the partitions (nodes) and add the node to each
+  // logical device it is assigned to.
   for (auto &network : networks) {
     for (auto &node : network.nodes) {
-      auto it = logicalDevices.find(node->logicalDevices[0]);
-      if (it != logicalDevices.end()) {
-        it->second.push_back(node.get());
-      } else {
-        logicalDevices.emplace(node->logicalDevices[0],
-                               std::vector<DAGNode *>{node.get()});
+      for (auto logical : node->logicalDevices) {
+        auto it = logicalDevices.find(logical);
+        if (it != logicalDevices.end()) {
+          it->second.push_back(node.get());
+        } else {
+          logicalDevices.emplace(logical, std::vector<DAGNode *>{node.get()});
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -58,12 +58,13 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
 
   for (auto &network : networks) {
     for (auto &node : network.nodes) {
```
nadavrot: Comments please.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -58,12 +58,13 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
...
-        it->second.push_back(node.get());
-      } else {
-        logicalDevices.emplace(node->logicalDevices[0],
-                               std::vector<DAGNode *>{node.get()});
+      for (auto logical : node->logicalDevices) {
+        auto it = logicalDevices.find(logical);
+        if (it != logicalDevices.end()) {
+          it->second.push_back(node.get());
+        } else {
+          logicalDevices.emplace(logical, std::vector<DAGNode *>{node.get()});
```
nadavrot: Are we creating an std::vector here? Why? Why can't we allocate a local static array and pass some ArrayRef or Begin-End pair? std::vector will cause a memory allocation.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -75,14 +76,19 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
...
-          llvm::make_unique<RuntimeBundle>(compiled->getRuntimeBundle());
-      functionMap.emplace(node->name, compiled.get());
-      functions_.emplace(node->name, std::move(compiled));
+      // Only compile if we haven't compiled before. If we have previously
+      // compiled the function reuse it.
+      auto it = functions_.find(node->name);
+      if (it == functions_.end()) {
+        Function *function = module.getFunction(node->name);
+        CompilationOptions compileOptions;
+        compileOptions.collectConstants = false;
```
nadavrot: This is not clear to me by reading the code what the provisioner is doing and why it needs to know something about collecting constants.

lib/Runtime/Provisioner/Provisioner.cpp
jackm321: The provisioner gives each DagNode a logical device id and this is grouping the DagNodes by those ids. Because these logical devices are created dynamically per graph it would be difficult to have a static allocation for them

lib/Runtime/Provisioner/Provisioner.cpp
jackm321: I agree. This is an existing fact of the provisioner rather than something introduced here though. I think the underlying issue is that runtime bundle isn't actually a thing that's needed by all backends, only backends running in the same memory space as the compiler (CPU, GPU, Interpreter) so while collect constants part of it has been moved to those devicemanagers, the runtime bundle still remains in the runtime but could probably be moved there as well. This is a conversation bigger than this pr though.

lib/Runtime/Provisioner/Provisioner.cpp
nadavrot: ah, sorry. I missed the fact that logicalDevices is a vector of vectors.

nadavrot:
LGTM.

PR #2745, opened 2019-04-17
[Runtime Efficiency] Avoid unnecessary tensor copies when propagating Placeholders

jackm321:
*Description*:
If a `DagNode` has only one child then there should be only one user of a given output placeholder from that `DagNode` and so copying the output placeholders' backing tensors is unnecessary, they can just be moved. If a `DagNode` has multiple children then the first N-1 children can get copies and we don't have to copy for the Nth `DagNode`.

*Testing*:
* Will sync internally and test
* Will update/add ExecutorTest unit tests

*Documentation*:
doxygen

### DIFF
--- a/lib/Runtime/Executor/ThreadPoolExecutor.cpp
+++ b/lib/Runtime/Executor/ThreadPoolExecutor.cpp
@@ -325,11 +325,14 @@ void ThreadPoolExecutor::run(const DAGNode *root,
   executionState->incrementInflightNodes(numChildren);
   inflightBarrier_.increment(numChildren);
 
-  for (auto const &node : root->children) {
+  for (size_t i = 0, e = root->children.size(); i < e; ++i) {
+    auto &node = root->children[i];
     // Propagate placeholders from the given starter PlaceholderBindings into
     // the input PlaceholderBindings for the current node being processed.
+    // TODO: do we always need to copy tensors here?
     propagatePlaceholdersForNode(executionState, node,
-                                 executionState->getRawResultContextPtr());
+                                 executionState->getRawResultContextPtr(),
+                                 /*cloneTensors*/ true);
 
     // Execute the node.
     executeDAGNode(executionState, node);
@@ -338,7 +341,7 @@ void ThreadPoolExecutor::run(const DAGNode *root,
 
 void ThreadPoolExecutor::propagatePlaceholdersForNode(
     std::shared_ptr<ExecutionState> executionState, const DAGNode *node,
-    const ExecutionContext *ctx) {
+    ExecutionContext *ctx, bool cloneTensors) {
   ScopedTraceBlock(executionState->getRawResultContextPtr()->getTraceContext(),
                    "EX_propagateInputs");
   // Get the symbol table for the node.
@@ -353,8 +356,13 @@ void ThreadPoolExecutor::propagatePlaceholdersForNode(
     // If ctx provides a mapping for the symbol, copy it into the context for
     // the node.
     if (placeholder) {
-      const auto *tensor = ctx->getPlaceholderBindings()->get(placeholder);
-      executionState->insertIntoNodeCtx(node, symbolName, tensor->clone());
+      if (cloneTensors) {
+        const auto *tensor = ctx->getPlaceholderBindings()->get(placeholder);
+        executionState->insertIntoNodeCtx(node, symbolName, tensor->clone());
+      } else {
+        auto *tensor = ctx->getPlaceholderBindings()->remove(placeholder);
+        executionState->insertIntoNodeCtx(node, symbolName, std::move(*tensor));
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


jackm321:
@nadavrot sounds good, I will write unit tests and I still need to find the best way to benchmark this. I mostly want to get feedback at this time, especially from @SplitInfinity, to see if this is this seems like a good direction for solving this issue.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -325,11 +325,14 @@ void ThreadPoolExecutor::run(const DAGNode *root,
   executionState->incrementInflightNodes(numChildren);
   inflightBarrier_.increment(numChildren);
 
-  for (auto const &node : root->children) {
+  for (size_t i = 0, e = root->children.size(); i < e; ++i) {
+    auto &node = root->children[i];
     // Propagate placeholders from the given starter PlaceholderBindings into
     // the input PlaceholderBindings for the current node being processed.
+    // TODO: do we always need to copy tensors here?
```
SplitInfinity: No, we can use the same logic as you did below (copy all but the last one).

nickgg:
Why do we ever need to copy in the Executor? Why not share the Tensor between different Placeholders in different subnode Functions?

SplitInfinity:
@nickgg 

I support this ideas but we'd need to make the `Tensors` in the `PlaceholderBindings` `shared_ptrs` or something semantically equivalent to make sure that `Tensors` are cleaned up only when the last reference to them is gone.

nadavrot:
Why do we need shared pointers? Do we have an unpredictable lifetime for the tensors?  In other parts of the compiler we rely on the clear ownership of things to manage lifetime. For example, the module owns functions and when it is destroyed the functions are deallocated. 

Are we expecting to have multiple owners of tensors that die at different times? If so, how are we going to manage the memory of the whole system? If not, I suggest using the simple ownership model of the phb owning the tensors.

nickgg:
@nadavrot In the runtime I think we might. The two cases where we might want a refcounting scheme for Tensors I can think of right now are:
* Activations between the different sub-partitions of a graph, where a Tensor might be used by two different partitions on different devices with unknown runtime. We can of course just keep all the Tensors around until the end of the high level run, but that might not be optimal.
* In the case where Tensor buffers are allocated by the Runtime and passed up (e.g. in the case where we allocate pinned buffers to avoid copies) we need to know when the outputs are no longer in use so we can reclaim the buffers. We likely will want object pools for buffers like these, as creating mapped hardware buffers is relatively expensive. Since the "free" happens outside of glow this would be unpredictable.

jackm321:
> Activations between the different sub-partitions of a graph, where a Tensor might be used by two different partitions on different devices with unknown runtime. We can of course just keep all the Tensors around until the end of the high level run, but that might not be optimal.

I think this is right but I don't know how suboptimal keeping intermediate Tensors around until the end of a request will end up being, we might even be able to use an arena for each request and put all the intermediate Tensors in that to make allocations super cheap.

nickgg:
@jackm321 I'm not familiar with that usage of "arena", what do you mean?

I do think we'll want some kind of object pooling for Tensors. The issue will be bursty load patterns though, I suspect.

jackm321:
@nickgg I'm talking about [arena allocation](https://en.wikipedia.org/wiki/Region-based_memory_management). It's used in latency sensitive applications where the lifetime of objects that depend on one another is well know so you can reduce the cost of allocation and deallocation

nickgg:
@jackm321 Gotcha, yes I think that makes definite sense for the Tensor buffers, possibly with or without the Tensor objects themselves.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -353,8 +356,13 @@ void ThreadPoolExecutor::propagatePlaceholdersForNode(
     // If ctx provides a mapping for the symbol, copy it into the context for
     // the node.
     if (placeholder) {
-      const auto *tensor = ctx->getPlaceholderBindings()->get(placeholder);
-      executionState->insertIntoNodeCtx(node, symbolName, tensor->clone());
+      if (cloneTensors) {
+        const auto *tensor = ctx->getPlaceholderBindings()->get(placeholder);
```
nadavrot: // Copy the tensor and ...

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -353,8 +356,13 @@ void ThreadPoolExecutor::propagatePlaceholdersForNode(
...
+        executionState->insertIntoNodeCtx(node, symbolName, tensor->clone());
+      } else {
+        auto *tensor = ctx->getPlaceholderBindings()->remove(placeholder);
```
nadavrot: // Remove the tensor and move the ownership to ...

jackm321:
Closing in favor of #2821

PR #2746, opened 2019-04-18
Add assertAllNodesQuantized option to quantizeFunction()

### LINKED ISSUE #2464: FunctionQuantizer silently skipping quantizing nodes

jfix71:
*Description*: This PR adds an `assertAllNodesQuantized` option to `quantizeFunction()`. If true, the Quantizer will die if it would have skipped quantizing a node because the backend does not support it as quantized. This is intended to be used by unit testing/testing bigger models via the Loader. For now it is disabled by default in the Loader, but by default true for `quantizeFunction()`. The motivation is explained in #2464.

*Testing*: Added tests for this functionality specifically. Also updated/removed tests where applicable.

*Documentation*: Added command line option with documentation on what it's for; added doxygen.

Fixes #2464

### DIFF
--- a/examples/bundles/lenet_mnist/Makefile
+++ b/examples/bundles/lenet_mnist/Makefile
@@ -35,7 +35,7 @@ ifeq ($(QUANTIZE),YES)
 build/lenet_mnist.o: profile.yml
 	mkdir -p build
 	# Create bundle with quantized weights and computation graph.
-	${LOADER} ${IMAGES}/3_1020.png -image-mode=0to1 -load-profile=profile.yml -m lenet_mnist -model-input-name=${MODEL_INPUT_NAME} -cpu -emit-bundle build -g
+	${LOADER} ${IMAGES}/3_1020.png -image-mode=0to1 -load-profile=profile.yml -m lenet_mnist -assert-all-nodes-quantized -keep-original-precision-for-nodes=SoftMax -model-input-name=${MODEL_INPUT_NAME} -cpu -emit-bundle build -g
 else
 build/lenet_mnist.o: download_weights
 	mkdir -p build
--- a/examples/bundles/resnet50/CMakeLists.txt
+++ b/examples/bundles/resnet50/CMakeLists.txt
@@ -50,8 +50,8 @@ add_custom_command(
   COMMAND
     image-classifier ${IMAGES}/dog_207.png -g -image-mode=0to1
     -m=${RESNET50_BUNDLE_DIR}/resnet50 -model-input-name=${MODEL_INPUT_NAME}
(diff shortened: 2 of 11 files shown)


beicy:
If I remember it correctly, we can define the set of nodes which are not quantized. So if we pass an empty set,  some nodes still may not be quantized due to that they are not supported, right?

jfix71:
>If I remember it correctly, we can define the set of nodes which are not quantized. So if we pass an empty set, some nodes still may not be quantized due to that they are not supported, right?

@beicy Yes -- so right now we have the `doNotQuantizeKinds` set, which is all nodes that should be skipped during quantization, regardless of whether `Backend::isOpSupported()` says it supports the node as quantized.

Currently if the `doNotQuantizeKinds` set is empty, and `Backend::isOpSupported()` says it does not support the node as quantized, then it does not quantize the node (meaning we may need to dequantize its inputs and/or quantize its outputs). The new option I'm adding, `enforceAllNodesQuantized`, will cause the quantizer to abort if it encounters this case.

The motivation here is many of our tests are not actually testing what we expect, in that they aren't fully quantizing the nodes in the test. And I've also seen regressions on our bigger models for quantization, where we make some change and the model is no longer fully quantized correctly.

PR #2751, opened 2019-04-18
[Quantization] Add QuantizationConfiguration struct for quantizeFunction()

jfix71:
*Description*: This PR changes `quantizeFunction()` by moving many of its current params into a simple aggregate struct `QuantizationConfiguration`. #2700 includes many of these parameters in `CompilationOptions`; I will rebase it on this PR to instead include `QuantizationConfiguration`. #2746 also adds another param which will be wrapped into `QuantizationConfiguration` too.

*Testing*: Updated all tests.

### DIFF
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -169,12 +169,12 @@ struct Model {
       // Lower however the backend prefers.
       ::lower(F_, &loweredMap_, EE_.getBackend());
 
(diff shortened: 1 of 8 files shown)


nadavrot:
LGTM.

jfix71:
I had to add constructors to keep Travis/gcc-4.8 happy, since it doesn't fully support C++14.

PR #2753, opened 2019-04-19
[Runtime] Added saturateHost flag and method to partitioner

gcatron:
*Description*: This PR adds a new saturateHost method to the Partitioner. If the saturateHost flag is set the Partitioner will duplicate networks to use all available devices on the host.
*Testing*: updated resnet-runtime to use new method, runs across multiple devices correctly.
*Documentation*: NA

### DIFF
--- a/examples/resnet-runtime.cpp
+++ b/examples/resnet-runtime.cpp
@@ -137,17 +137,14 @@ int main(int argc, char **argv) {
   std::vector<size_t> inputShape{1, 3, 224, 224};
 
   Placeholder *input;
-  std::vector<Placeholder *> inputs;
-  std::vector<PlaceholderList> phLists;
-  for (unsigned int i = 0; i < numDevices; i++) {
-    std::unique_ptr<Module> module = llvm::make_unique<Module>();
-    TypeRef inputType = module->uniqueType(ElemKind::FloatTy, inputShape);
-    input = loadResnet50Model(inputType, module.get(), i);
-    inputs.push_back(input);
-    llvm::outs() << "Adding to HostManager\n";
-    phLists.push_back(module->getPlaceholders());
-    EXIT_ON_ERR(hostManager->addNetwork(std::move(module)));
-  }
+  PlaceholderList phList;
+
+  std::unique_ptr<Module> module = llvm::make_unique<Module>();
+  TypeRef inputType = module->uniqueType(ElemKind::FloatTy, inputShape);
+  input = loadResnet50Model(inputType, module.get(), 0);
+  phList = module->getPlaceholders();
+  EXIT_ON_ERR(
+      hostManager->addNetwork(std::move(module), /*saturateHost*/ true));
 
   llvm::outs() << "Loading files from " << inputDirectory << "\n";
   std::error_code code;
@@ -175,7 +172,7 @@ int main(int argc, char **argv) {
       }
       break;
     }
-    int index = currDevice % numDevices;
+
     std::string path = dirIt->path();
 
     auto image = readPngImageAndPreprocess(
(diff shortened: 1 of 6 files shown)


gcatron:
> Rather than add a bool, should we think about a strict here? I can see perhaps wanting to provide the compile options.

I thought about it, I wasn't sure what else we'd add so I decided to use the bool for now and if we add something else make a struct then.

lib/Partitioner/Partitioner.cpp
```diff
@@ -532,10 +558,13 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   if (mapping.getPartitions().size() > deviceInfo_.size()) {
     adjustLogicalDeviceID(root, mapping.getPartitions().size());
   }
+  if (saturateHost_) {
```
beicy: nit:  this "if" can be the "else" branch of the "if" in line 558. like:
```
if (mapping.getPartitions().size() > deviceInfo_.size()) {..
} else if (saturateHost_) {
}
```

examples/resnet-runtime.cpp
```diff
@@ -137,17 +137,14 @@ int main(int argc, char **argv) {
...
-    llvm::outs() << "Adding to HostManager\n";
-    phLists.push_back(module->getPlaceholders());
-    EXIT_ON_ERR(hostManager->addNetwork(std::move(module)));
-  }
+  PlaceholderList phList;
+
+  std::unique_ptr<Module> module = llvm::make_unique<Module>();
+  TypeRef inputType = module->uniqueType(ElemKind::FloatTy, inputShape);
+  input = loadResnet50Model(inputType, module.get(), 0);
+  llvm::outs() << "Adding to HostManager\n";
```
jackm321: Do we want to keep this in? Probably not

examples/resnet-runtime.cpp
```diff
@@ -137,17 +137,14 @@ int main(int argc, char **argv) {
...
+  phList = module->getPlaceholders();
+  EXIT_ON_ERR(hostManager->addNetwork(std::move(module), true));
```
jackm321: add /*saturateHost*/ to the bool

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -92,7 +92,8 @@ class HostManager final {
   /// in networks_. Returns an llvm::Error containing the results of the
   /// operation. This function consumes the \p module so any pointers to data
   /// contained within the module should be considered invalid.
-  llvm::Error addNetwork(std::unique_ptr<Module> module);
+  llvm::Error addNetwork(std::unique_ptr<Module> module,
```
jackm321: add \p saturateHost

lib/Partitioner/Partitioner.cpp
```diff
@@ -531,11 +557,13 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   // Adjust the logicalDevice for each DAGNode.
   if (mapping.getPartitions().size() > deviceInfo_.size()) {
     adjustLogicalDeviceID(root, mapping.getPartitions().size());
+  } else if (saturateHost_) {
+    // Saturate the Host.
+    saturateHost(logicalID + 1);
```
jackm321: why +1? can you add a comment?

lib/Partitioner/Partitioner.cpp
```diff
@@ -423,6 +424,31 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+  if (duplications < 2) {
+    return;
+  }
+  // Add additional logical devices to each node.
+  for (auto &network : partitions_) {
+    for (auto &node : network.nodes) {
+      // Build list of new logical devices to add to node.
+      std::vector<unsigned> newDevices;
+      for (auto logical : node->logicalDevices) {
+        for (unsigned i = 1; i < duplications; i++) {
```
jackm321: can you add some comments on this logic, like why it starts at 1 and what logical + (i * logicalDeviceCount) is for

examples/resnet-runtime.cpp
gcatron: I kinda like it, it's a nice progress indicator.

nadavrot: We don’t just print to stdout whenever we like it. Glow is a library. You can use glog if you want to report progress.

Issue #2687, opened 2019-04-09
Support for a heteogeneous partitioning


opti-mix:
Glow's runtime currently supports partitioning of a graph between multiple devices of the same kind, i.e. controlled by the same kind of a backend.

But it could be very useful to support heterogeneous partitioning as well, so that a graph could be partitioned among devices of different kinds.

One specific use-case which would benefit from it is when you bring up new backends. Initially they do not support all of the ML operations. Thus, one could just partition the graph in such a way that all operations not supported by the new backend yet would be in a separate partition(s) executed by e.g. an existing, established backend (e.g Interpreter or CPU). All the operations supported by the new backend already would be in partitions executed by this backend.

Done this way, it becomes very easy to perform an end-2-end testing of the whole models with a new backend even though not all operations are supported yet. One can improve this support gradually, which makes the development of a new backend more manageable.

opti-mix:
@beicy

opti-mix:
cc @gcatron @jackm321 @nickgg

beicy:
Discussed with @opti-mix and @gcatron, our short term plan is to support 2-level partitions.
Level 1: do the partition based on backends type.
Level 2: for the type of backends, apply current partition algorithm. 
There are some changes :
1. Now HostManager should pass the BackendKind of each device to Partitioner through DeviceInfo.
2. The optimization should be done in Partitioner after level 1 partition.
3. In Level 2 partition, we just do the partition for current functions one by one.
4. In final DAG, each node should have the type "BackendKind" for later assignment.

PR #2755, opened 2019-04-19
[WIP][Partitioner] Add user-defined partition.

beicy:
*Description*:
We plan to add user-defined graph-partition. This PR defines the interfaces and a simple version of config file to provide users a quick way for model testing. Further discussion about the config file format is needed.  

*Testing*:
Add unit test.

*Documentation*:
#2687 

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -157,6 +157,10 @@ public:
   /// devices.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices);
 
+  /// Decompose a function based on a config.
+  llvm::Error PartitionFromConfig(const std::string &funcName,
+                                  const std::string &filename);
+
   /// Decompose each function in a module.
   llvm::Error Partition();
 
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -16,6 +16,8 @@
 
 #include "glow/Partitioner/Partitioner.h"
 
+#include <fstream>
+
 using namespace glow;
 using llvm::isa;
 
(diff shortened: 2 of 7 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -534,6 +536,56 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+  while (ff >> str) {
+    if (str.compare("#") == 0) {
+      // Start a new partition.
+      newF = F->getParent()->createFunction(std::string(F->getName()) +
+                                            "_part" + std::to_string(++color));
+      partitionMap.createPartition(newF);
+    } else {
+      auto *N = F->getNodeByName(str);
+      assert(N && "Invalid node name");
+      partitionMap.add(N, newF);
```
opti-mix: If a file has a wrong format, then this code would try to add to a partition that does not exist yet.
E.g.
```
node1 node2 node3
```

I'd suggest initializing `newF` with nullptr and asserting here if `newF` is nullptr.

lib/Partitioner/Partitioner.cpp
```diff
@@ -534,6 +536,56 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   }
 }
 
+/// Get the partitions of a function \p funcName based on a config file with \p
+/// filename. It is a text file with the format:
+///
+/// # node_name_1 node_name_2 ...node_name_n # node_name_x1 node_name_x2 ...
```
opti-mix: It is useful if one could also have comments in the config files as they allow for providing some explanations. Thus it would be nice if you could allow for some kind of comments. Usually comments start with either `//` or `#` in many languages. Therefore your choice of `#` for a partition indicator is probably not the best one ;-)

beicy: I see. We can start a partition with a number, since in Glow, a valid name should start with alphabet letters.

opti-mix: I'm not sure about using a number. May be something like braces? E.g. `[node1 node2 node3]` or `(node1 node2 node3])`. Or `-> node1 node2 node3`...

include/glow/Partitioner/Partitioner.h
```diff
@@ -157,6 +157,10 @@ class Partitioner {
   /// devices.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices);
 
+  /// Decompose a function based on a config.
+  llvm::Error PartitionFromConfig(const std::string &funcName,
```
nadavrot: Please pass StringRef here. std::string forces you to malloc twice, while string-ref gives you the option to pass a regular c-string.

tests/unittests/PartitionerTest.cpp
```diff
@@ -19,6 +19,10 @@
 
 #include "gtest/gtest.h"
 
+#ifndef GLOW_DATA_PATH
```
nadavrot: What is this?

nadavrot:
@beicy The partitioner should not include fstream and should not parse config files itself. It's a library. If you want to move forward with this design then please make the partitioner accept some data structure, such as a map of vectors, or a list of lists, or something. The parsing should happen outside of the partitioner. This library should not open files on disk, handle errors, etc. All of this should happen by the driver externally. Also, instead of inventing your own fragile format please use something like YAML.

tests/unittests/PartitionerTest.cpp
rdzhabarov: this is required for the fbcode setup. Man, can you add a comment on this.

lib/Partitioner/Partitioner.cpp
```diff
@@ -534,6 +536,56 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+  std::ifstream ff(filename, std::ios::in);
+  RETURN_ERR_IF_NOT(ff, "Can't find the configuration file.");
+
+  NodeToFunctionMap partitionMap;
+  std::string str;
+  int color = 0;
+  Function *newF;
+  int num = 0;
+  while (ff >> str) {
+    if (str.compare("#") == 0) {
```
rdzhabarov: looks a bit unnatural, str == "#"

lib/Partitioner/Partitioner.cpp
```diff
@@ -534,6 +536,56 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+      // Start a new partition.
+      newF = F->getParent()->createFunction(std::string(F->getName()) +
+                                            "_part" + std::to_string(++color));
+      partitionMap.createPartition(newF);
+    } else {
+      auto *N = F->getNodeByName(str);
+      assert(N && "Invalid node name");
```
rdzhabarov: asserts are good, but since we started using GLOG, maybe do CHECKS and provide better logging.
Like having node name here would be beneficial, same in other places.

PR #2757, opened 2019-04-20
[CMakeLists] Support 3.7 and prefer latest packages found via find_package

### LINKED ISSUE #2756: [CMake] Support CMake 3.7 to use sorting order feature of find_package?
CMake 3.7 has support for [sorting packages found from `find_package` via `CMAKE_FIND_PACKAGE_SORT_DIRECTION`](https://cmake.org/cmake/help/latest/release/3.7.html?highlight=cmake_find_package_sort_direction#id5), so that it prefers using the latest version found. This would help with annoyances e.g. in LLVM where right now we're searching for 8, then 7, then fallback, and that fallback is currently [failing on Travis CI since it's seemingly now installing 7.1 instead but instead finding 5 first](https://travis-ci.org/pytorch/glow/jobs/522379755) (and not sure why searching for 7 doesn't find 7.1, CC @arunm-git who also mentioned this in #2726).

@bertmaher moved to CMake 3.5 from 3.7 a few months ago in #2153. Not sure if this is still helpful or worth the tradeoff to drop this support in favor of using this sorting feature.

jfix71:
As described in / would fix #2756

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.7)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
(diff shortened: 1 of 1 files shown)


nadavrot:
I can see that this PR does not solve #2756. I am stamping this to unblock you, just in case you can figure out a way to change this PR and fix the build.

jfix71:
Closing in favor of https://github.com/pytorch/glow/pull/2761

PR #2759, opened 2019-04-22
Fix the GLOG CMake file.

nadavrot:
*Description*:

Before this change I was not able to build Glow on my home mac using the
brew glog. This change implements the recommended way to use glog with
CMake from the glog website.

*Testing*: Tested on my mac. The project now builds. 

*Documentation*: None.


nadavrot:
My trivial solution to use FIND_PACKAGE fails on linux. I am trying a workaround now.

rdzhabarov:
fixed by https://github.com/pytorch/glow/pull/2763

PR #2761, opened 2019-04-22
Update .travis.yml

compnerd:
Use `LLVM_DIR` rather than `CMAKE_PREFIX_PATH` to locate LLVM.  This is more modern, and avoids the problem of searching for multiple versions as we have provided an absolute path to the location of the CMake module for LLVM.

*Description*:
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         - mkdir build && cd build
         - CC=gcc-4.8 CXX=g++-4.8 cmake -G Ninja
           -DCMAKE_BUILD_TYPE=Debug -DGLOW_WITH_OPENCL=OFF -DGLOW_WITH_CPU=ON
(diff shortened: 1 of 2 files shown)


artemrakhov:
@compnerd: How does this work? Where does Glow's build system use `LLVM_DIR`? This broke build on my machine, because it now wants to use LLVM v5.

jfix71:
@artemrakhov When you run cmake you will point it to the path to `lib/cmake/llvm`, e.g. `cmake -G Ninja ../glow -DLLVM_DIR=/path/to/llvm_install/lib/cmake/llvm`

artemrakhov:
Got it. Apparently `LLVM_DIR` is special flag which tells CMake where to find LLVM. But if LLVM is not found exactly there, CMake can fall back to system's random LLVM version.

PR #2762, opened 2019-04-22
[CMake] Look for llvm 7.1

jfix71:
Trying to fix travis


CMakeLists.txt
```diff
@@ -107,17 +107,21 @@ ExternalBackendsInit()
 
 find_package(LLVM 8 CONFIG)
```
nadavrot: Could we refactor this code into a separate file under /cmake/modules/ ? 

We don't need to do this now, but after we fix the build it would be good to clean things up.

jfix71:
Closing in favor of https://github.com/pytorch/glow/pull/2761

PR #2768, opened 2019-04-22
[tracing] add automatic thread ids

nickgg:
*Description*: Replace Manual setting of thread id with automatic ids, which eliminates issues where multiple work was happening "concurrently". This disables thread names for now.
*Testing*: unit tests, tracing-compare and resnet-runtime
*Documentation*:

### DIFF
--- a/include/glow/Backends/TraceEvents.h
+++ b/include/glow/Backends/TraceEvents.h
@@ -65,7 +65,11 @@ struct TraceEvent {
                   const std::string &processName = "",
                   const std::map<int, std::string> &threadNames = {});
 
+  // Return the current time in microseconds in the timestamp domain.
   static uint64_t now();
+
+  // Returns a unique id associated with the current thread.
+  static size_t getThreadId();
 };
 
 /// Tracing / Profiling events map for a CompiledFunction.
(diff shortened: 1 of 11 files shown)


include/glow/Backends/TraceEvents.h
```diff
@@ -65,7 +65,11 @@ struct TraceEvent {
                   const std::string &processName = "",
                   const std::map<int, std::string> &threadNames = {});
 
+  // Return the current time in microseconds in the timestamp domain.
   static uint64_t now();
+
+  // Return the thread id of the current thread.
```
nadavrot: Is this correct? I think that this is more like a running counter that assigns a unique id for each thread.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -372,42 +371,33 @@ void ThreadPoolExecutor::executeDAGNode(
 
   // If tracing is enabled, set the thread name for TraceEvents for this node to
   // be the name of the Device.
-  if (executionState->getRawResultContextPtr()->getTraceContext()) {
-    executionState->getRawResultContextPtr()->getTraceContext()->setThreadName(
-        currentDevice, deviceManager->getDeviceConfig()->getName());
-  }
+  // if (executionState->getRawResultContextPtr()->getTraceContext()) {
```
nadavrot: Ooops.

nickgg: I do have the TODO for this below, but I could move it up so it's clearer. Intentionally commented out because I want to fix the thread names.

include/glow/Backends/TraceEvents.h
nickgg: Well, what is that but a thread id? It's not the OS thread idea though to be sure (std::this_thread::id() isn't easily convertable to something human readable).

 I'll adjust the comment.

PR #2769, opened 2019-04-23
[NFC] Use llvm-link from the given LLVM build directory

artemrakhov:
This helps to eliminate mismatch between given LLVM's linker and system's standard llvm-link

### DIFF
--- a/lib/Backends/CPU/CMakeLists.txt
+++ b/lib/Backends/CPU/CMakeLists.txt
@@ -4,16 +4,7 @@ else()
   find_program(CLANG_BIN clang++)
 endif()
 
-if(TARGET llvm-link)
-  set(LLVM_LINK_BIN $<TARGET_FILE:llvm-link>)
-else()
-  find_program(LLVM_LINK_BIN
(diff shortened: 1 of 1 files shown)


lib/Backends/CPU/CMakeLists.txt
```diff
@@ -4,16 +4,7 @@ else()
...
-  set(LLVM_LINK_BIN $<TARGET_FILE:llvm-link>)
-else()
-  find_program(LLVM_LINK_BIN
-               NAMES
-                 llvm-link-${LLVM_VERSION_MAJOR}
-                 llvm-link-8
-                 llvm-link-7
-                 llvm-link)
-endif()
+set(LLVM_LINK_BIN ${LLVM_BINARY_DIR}/bin/llvm-link)
```
compnerd: hmm, I *think* that this should be equivalent.

artemrakhov: Probably, but at least it does not try to use system's llvm-link now.

PR #2771, opened 2019-04-23
[Runtime Efficiency] Use placeholder pointers as ids in runtime

### LINKED ISSUE #2775: [Runtime efficiency] Don't use Placeholder names as ids in the runtime
After perf profiling, we identified hashing string names for placeholders as a big cost in the runtime during placeholder propagation. Avoid this by using integer ids or preferably Placeholder* in the runtime to uniquely identify placeholders instead for much faster map lookups.

jackm321:
*Description*:
Instead of storing Placeholders by their name in PlaceholderBindings, only store their pointers and map tensors in based on these pointers instead of on string names. We have seen from perf profiling in the past that the runtime cost of querying and updating placeholder bindings by string keys is huge, using pointers is cheaper.

TODO:
* test by syncing and test internally
* benchmark

*Testing*:
* resnet-runtime works
* sync and test internally

*Documentation*:
Doxygen updated

Fixes #2775


tests/unittests/CMakeLists.txt
```diff
@@ -421,17 +421,18 @@ target_link_libraries(QuantizationTest
...
-target_link_libraries(ExecutorTest
-                      PRIVATE
-                        DeviceManager
-                        Executor
-                        Graph
-                        gtest
-                        TestMain
-                        ThreadPool)
-add_glow_test(ExecutorTest ${GLOW_BINARY_DIR}/tests/ExecutorTest --gtest_output=xml:ExecutorTest.xml)
+# TODO:
```
nadavrot: ??

include/glow/Graph/PlaceholderBindings.h
```diff
@@ -38,18 +38,12 @@ class Placeholder;
 class PlaceholderBindings final {
 public:
   /// Maps placeholders to the tensors that back them.
-  using PlaceholderMap = std::unordered_map<Placeholder *, Tensor *>;
-
-  /// Maps Placeholder names to Placeholders.
-  using PlaceholderNameMap = std::unordered_map<std::string, Placeholder *>;
+  using PlaceholderMap = std::unordered_map<const Placeholder *, Tensor *>;
```
nadavrot: I think that this is fine for now. Keep in mind that LLVM may have more efficient data structures, such as densemap.

nadavrot:
Why do we need to change the code in examples/resnet-runtime.cpp ?  Are we not providing enough high-level APIs that the examples need to look into the internals of the binding? Maybe we should keep the high-level named PH API to make it easier to write unit tests? The named API can be slower but functional. I am not sure if it's a good idea but it's worth mentioning.

lib/Graph/PlaceholderBindings.cpp
```diff
@@ -64,26 +29,13 @@ Tensor *PlaceholderBindings::get(Placeholder *P) const {
   return it->second;
 }
 
-Placeholder *
-PlaceholderBindings::getPlaceholderByName(llvm::StringRef name) const {
```
nadavrot: Let's keep this API. It can be useful.

nickgg:
This is an efficiency improvement, right? Have you tested performance - what is the improvement?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -327,17 +328,19 @@ void ThreadPoolExecutor::propagatePlaceholdersForNode(
...
 
-    auto *placeholder =
-        ctx->getPlaceholderBindings()->getPlaceholderByName(symbolName);
-
-    // If ctx provides a mapping for the symbol, copy it into the context for
-    // the node.
-    if (placeholder) {
-      const auto *tensor = ctx->getPlaceholderBindings()->get(placeholder);
-      executionState->insertIntoNodeCtx(node, symbolName, tensor->clone());
+    if (symbolInfo.symbolCategory != SymbolCategory::Placeholder) {
```
nickgg: we don't need to check `symbolCategory` here, just if `placeholder` is set

lib/Graph/PlaceholderBindings.cpp
nickgg: If its just for unittests it doesn't need to be fast though, we can still drop the `nameMap_` and just iterate over all placeholders. It would be nice to obviously discourage use of this API in runtime code.

lib/Backends/BackendUtils.cpp
```diff
@@ -145,6 +145,7 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
     symbol.output = isOutput(V);
     symbol.input = isInput(V);
     symbol.symbolCategory = SymbolCategory::Placeholder;
+    symbol.placeholder = V;
```
nickgg: I like this change but we must ensure the Module outlives the RuntimeBundle for this pointer to be valid, is that guaranteed currently?

lib/Graph/PlaceholderBindings.cpp
```diff
@@ -20,42 +20,7 @@
 
 using namespace glow;
 
-bool PlaceholderBindings::compare(const PlaceholderBindings *A,
```
SplitInfinity: I think you'll need to keep this around to get the ExecutorTest to work. You should probably revert the name-based lookup (i.e. `const auto *tensorB = B->get(B->getPlaceholderByName(placeholder->getName()));`) but I think you'll need this function in general.

I'm also fine with moving it into the ExecutorTest to discourage people from using it all over the place.

jackm321: Yeah you're right I've added it back

lib/Backends/BackendUtils.cpp
jackm321: Yes good point, their lifetimes are tied together in HostManager

lib/Graph/PlaceholderBindings.cpp
jackm321: Sounds like a plan, I don't want to keep nameMap_ but iterating is fine

lib/Runtime/Executor/ThreadPoolExecutor.cpp
jackm321: It should always be `placeholder != nullptr` iff `symbolInfo.symbolCategory != SymbolCategory::Placeholder` so either way works, I can flip which is the condition and which is the assertion but I still want to check that the above is true. Do you think checking the placeholder* is better?

lib/Backends/BackendUtils.cpp
nickgg: We have usages of the CompiledFunction (+RuntimeBundle) outside of the runtime, can we add a comment somewhere about this restriction. Maybe even on `Backend::compile`?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
nickgg: I don't think flipping them is any better or worse, if you want to check both here then I think go with whatever you think is more readable. Another alternative is to add a getter for placeholder which does the assert but I don't have strong feelings either way.

examples/resnet-runtime.cpp
```diff
@@ -136,12 +134,11 @@ int main(int argc, char **argv) {
 
   std::vector<size_t> inputShape{1, 3, 224, 224};
 
-  Placeholder *input;
   PlaceholderList phList;
 
   std::unique_ptr<Module> module = llvm::make_unique<Module>();
   TypeRef inputType = module->uniqueType(ElemKind::FloatTy, inputShape);
-  input = loadResnet50Model(inputType, module.get(), 0);
+  auto inputAndOutput = loadResnet50Model(inputType, module.get(), 0);
```
nickgg: nit: might be clearer with std::tee

jackm321:
@nickgg no I haven't, it seems to be broken. @SplitInfinity Do you know why the RuntimeBench is broken?

PR #2772, opened 2019-04-23
fix cmake bug for OpenCL

ffk0716:
*Description*: cmake failed with OpenCL build
*Testing*:
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Backends/OpenCL/CMakeLists.txt
+++ b/lib/Backends/OpenCL/CMakeLists.txt
@@ -1,18 +1,24 @@
 set(OpenCL_DIR ${CMAKE_BINARY_DIR}/glow/OpenCL)
 
 add_custom_command(
+    OUTPUT ${OpenCL_DIR}
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${OpenCL_DIR})
+add_custom_target(make_opencl_dir
+    DEPENDS ${OpenCL_DIR})
+
+add_custom_command(
   OUTPUT "${OpenCL_DIR}/kernels.cl.inc"
   COMMAND include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl"
           "${OpenCL_DIR}/kernels.cl.inc"
-  DEPENDS include-bin "${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl")
+  DEPENDS make_opencl_dir include-bin "${CMAKE_CURRENT_SOURCE_DIR}/kernels.cl")
 
 add_custom_command(
   OUTPUT "${OpenCL_DIR}/kernels_fwd_conv.cl.inc"
   COMMAND include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_conv.cl"
           "${OpenCL_DIR}/kernels_fwd_conv.cl.inc"
-  DEPENDS include-bin
+  DEPENDS make_opencl_dir include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_conv.cl")
 
 add_custom_command(
@@ -20,7 +26,7 @@ add_custom_command(
   COMMAND include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_quantized_conv.cl"
           "${OpenCL_DIR}/kernels_fwd_quantized_conv.cl.inc"
-  DEPENDS include-bin
+  DEPENDS make_opencl_dir include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_quantized_conv.cl")
 
 add_library(OpenCLBackend
--- a/lib/Backends/OpenCL/CMakeLists.txt
+++ b/lib/Backends/OpenCL/CMakeLists.txt
@@ -1,30 +1,32 @@
+set(OpenCL_DIR ${CMAKE_BINARY_DIR}/glow/OpenCL)
+
 add_custom_command(
-  OUTPUT "${CMAKE_BINARY_DIR}/glow/OpenCL/kernels.cl.inc"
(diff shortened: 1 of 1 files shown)


nadavrot:
Thank you for doing this work @ffk0716 !

nadavrot:
@SplitInfinity

lib/Backends/OpenCL/CMakeLists.txt
```diff
@@ -1,30 +1,38 @@
...
 
 add_custom_command(
-  OUTPUT "${CMAKE_BINARY_DIR}/glow/OpenCL/kernels_fwd_quantized_conv.cl.inc"
+  OUTPUT "${OpenCL_DIR}/kernels_fwd_quantized_conv.cl.inc"
   COMMAND include-bin
           "${CMAKE_CURRENT_SOURCE_DIR}/kernels_fwd_quantized_conv.cl"
-          "${CMAKE_BINARY_DIR}/glow/OpenCL/kernels_fwd_quantized_conv.cl.inc"
-  DEPENDS include-bin CPURuntime
+          "${OpenCL_DIR}/kernels_fwd_quantized_conv.cl.inc"
+  DEPENDS make_opencl_dir include-bin CPURuntime
```
opti-mix: Not a problem in this PR, but why do we depend on `CPURuntime` in our `OpenCL` CMake file? Is it a copy&paste bug or something like this?

@rdzhabarov @SplitInfinity

arunm-git:
CC: @SplitInfinity since the original changes to the opencl cmakelists.txt file were made to handle internal builds.
If the location of the .h files is changing through this PR, that will need to be handled internally too.

SplitInfinity:
@arunm-git 

I made this change because handling it internally is harder (cc: @bertmaher). In #2721 I relied only on CI which was a mistake, so let me revisit that and figure out how to move the kernels properly without breaking the build.

SplitInfinity:
Oh, the location of the files is not changing, never mind.

lib/Backends/OpenCL/CMakeLists.txt
SplitInfinity: Hmm, strange, I'm not sure...

SplitInfinity:
Is anyone else able to reproduce this?

I just built from trunk and got:
```
meghanl-mbp:build_DEBUG meghanl$ ./tests/OCLTest 
[==========] Running 4 tests from 1 test case.
[----------] 4 tests from OpenCLCorrectnessTest
[ RUN      ] OpenCLCorrectnessTest.convOps
[       OK ] OpenCLCorrectnessTest.convOps (49 ms)
[ RUN      ] OpenCLCorrectnessTest.inferMixedNet
[       OK ] OpenCLCorrectnessTest.inferMixedNet (20 ms)
[ RUN      ] OpenCLCorrectnessTest.softmaxGradTest
[       OK ] OpenCLCorrectnessTest.softmaxGradTest (105 ms)
[ RUN      ] OpenCLCorrectnessTest.tanhConcatTest
[       OK ] OpenCLCorrectnessTest.tanhConcatTest (10 ms)
[----------] 4 tests from OpenCLCorrectnessTest (184 ms total)

[==========] 4 tests from 1 test case ran. (185 ms total)
[  PASSED  ] 4 tests.
```

@ffk0716 Which `cmake` version are you using? I'm using `3.12.1`.

ffk0716:
I am using cmake version 3.14.2

lib/Backends/OpenCL/CMakeLists.txt
rdzhabarov: hehe, no idea, but @opti-mix you introduced that in
https://github.com/pytorch/glow/commit/2b9abbc012437004e6b9766ded2b400903598a07

rdzhabarov:
Pure CMake changes should not affect any internal components anyhow (we should be good from that point of view).

lib/Backends/OpenCL/CMakeLists.txt
opti-mix: Ha-ha. So, it was me??? :-D I'm almost sure it is a copy&paste error.

ffk0716:
Hello?

SplitInfinity:
@ffk0716 

I am not able to reproduce this issue using `cmake v 3.12.1` on macOS Mojave 10.14.4. What are the details of your setup?

ffk0716:
Did you enable OpenCL backend?
Which command generates ${CMAKE_BINARY_DIR}/glow/OpenCL in your case?

In my case:
```
mkdir build
cd build
cmake .. -DGLOW_WITH_OPENCL=ON
ls glow/OpenCL
ls: glow/OpenCL: No such file or directory
make
[ 32%] Generating ../../../glow/OpenCL/kernels_fwd_quantized_conv.cl.inc
Could not open output file: No such file or directory

```

rdzhabarov:
@ffk0716 

Could you try ninja generator in cmake? (I can run cmake/ninja with cmake version 3.14.3 with no issues).
I do not think we were able to use GNU make for long time in Glow.

SplitInfinity:
I was just about to comment - I was also able to build just fine with `Ninja`.

ffk0716:
ninja works fine.
It seems gnu make does not mkdir automatically.
But ninja does.

rdzhabarov:
@ffk0716 
Thanks for digging into the problem!

I'd suggest to simplify a bit and just  run `file(MAKE_DIRECTORY ${OpenCL_DIR})` and remove custom target.

see: https://github.com/pytorch/glow/blob/e02b8732729bed843401398c1ab1822c9b623c09/lib/Backends/CPU/CMakeLists.txt#L27 for example.

ffk0716:
`file()` also works in this case.

But if we delete folder and remake.
comstom_target works, but `file()` not.

SplitInfinity:
No more outstanding questions or issues.

PR #2778, opened 2019-04-23
[tracing] [trivial] switch to std::system_clock for timestamps

nickgg:
*Description*: Internally we use system_clock for the timestamp domain, changing Glow trace events to match.
*Testing*: unittests
*Documentation*:


arunm-git:
We don't care about monotonicity in this use case?

nickgg:
We do, but not enough to change all the existing usage of system_clock, it seems.

In this case we shouldn't have big problems since we don't have events that cross threads.

PR #2780, opened 2019-04-23
[Training] Add support for training SparseLengthsWeightedSum

SplitInfinity:
**Description**
This commit adds support for training the `SparseLengthsWeightedSum`
operator. The gradient for this node is computed by initializing the
`data` gradient to be zero, and then accumulating into `dataGradient[index]`
the gradient of the result slice that `data[index]` was added to during
the `SparseLengthsWeightedSum` operation multiplied by `weight[index]`
(the same weight that `data[index]` was multiplied by during the
`SparseLengthsWeightedSum` operator).

**Testing**
This commit adds a unit test to `MLTest` that trains the data input of a
`SparseLengthsWeightedSum` node so that the operation outputs a
specified `Tensor`. The indices, lengths, weights are fixed.

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1089,6 +1089,30 @@ void libjit_sparse_lengths_weighted_sum_f(float *dest, float *data,
   }
 }
 
+void libjit_sparse_lengths_weighted_sum_grad_f(
+    const float *destGrad, float *dataGrad, const float *weights,
+    const size_t *indices, const int32_t *lengths, size_t segments,
+    size_t lineSize, size_t dataGradRawSize) {
+  // The data gradients not touched by this operation should
+  // be 0, so set the entire buffer to 0 to start with.
+  memset(dataGrad, 0, dataGradRawSize);
+  size_t curIndex = 0;
+  for (size_t i = 0; i < segments; i++) {
+    for (int32_t j = 0; j < lengths[i]; j++) {
+      // For each index in each segment, accumulate into the corresponding data
+      // gradient the product of the gradient of the result it was added to and
+      // the weight that it was multiplied by during the
+      // SparseLengthsWeightedSum operation.
+      float weight = weights[curIndex];
+      size_t line = indices[curIndex];
+      for (size_t k = 0; k < lineSize; k++) {
+        dataGrad[line * lineSize + k] += weight * destGrad[i * lineSize + k];
+      }
+      curIndex++;
+    }
+  }
+}
+
 void libjit_rowwise_quantized_sparse_lengths_weighted_sum_f(
     float *dest, int8_t *data, float *scales, float *offsets, float *weights,
     size_t *indices, int32_t *lengths, size_t segments, size_t lineSize) {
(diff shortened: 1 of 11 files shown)


SplitInfinity:
Note that this is the same as #2541 but it is not stacked on top of #2524 (adding support for Adagrad). OpenCL implementation coming soon.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1089,6 +1089,31 @@ void libjit_sparse_lengths_weighted_sum_f(float *dest, float *data,
...
 }
 
+void libjit_sparse_lengths_weighted_sum_grad_f(float *destGrad, float *dataGrad,
+                                               float *weights, size_t *indices,
+                                               int32_t *lengths,
+                                               size_t segments, size_t lineSize,
+                                               size_t numLines) {
+  // The data gradients not touched by this operation should
+  // be 0, so set the entire buffer to 0 to start with.
+  memset(dataGrad, 0, numLines * lineSize * sizeof(float));
```
jfix71: nit: Since `numLines` is only used here, perhaps calculate the size for `memset()` at compile time in LLVMIRGen and pass it instead of `numLines`.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1089,6 +1089,31 @@ void libjit_sparse_lengths_weighted_sum_f(float *dest, float *data,
   }
 }
 
+void libjit_sparse_lengths_weighted_sum_grad_f(float *destGrad, float *dataGrad,
```
jfix71: nit: You could all but `dataGrad` `const`.

tests/unittests/MLTest.cpp
```diff
@@ -1403,6 +1403,79 @@ TEST_P(MLTest, matrixRotationRecognition) {
...
+  PlaceholderBindings bindings;
+
+  Module &mod = EE_.getModule();
+  PseudoRNG &PRNG = mod.getPRNG();
+
+  // Create a model consisting of one SparseLengthsWeightedSum operator followed
+  // by a Regression node to get some non-zero gradients.
+  Function *F = mod.createFunction("SparseLengthsWeightedSum");
+  Placeholder *dataP =
+      mod.createPlaceholder(ElemKind::FloatTy, {10}, "dataP", true);
```
jfix71: nit: I know we don't use it super often, but I'd prefer adding `/* isTrainable */ true`, and so on especially since we are actually training here.

nadavrot:
Very cool!

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -187,6 +187,19 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
            (NI.getInElemTy(SparseLengthsWeightedSumNode::LengthsIdx) ==
             ElemKind::Int32ITy);
 
+  case Kinded::Kind::SparseLengthsWeightedSumGradNodeKind:
+    return NI.allInputsAndOutputsHaveSameElemKind(
+               {ElemKind::FloatTy},
+               {SparseLengthsWeightedSumGradNode::IndicesIdx,
+                SparseLengthsWeightedSumGradNode::LengthsIdx},
+               {SparseLengthsWeightedSumGradNode::GradOfInputNamedIndicesIdx,
```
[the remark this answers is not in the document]
SplitInfinity: I added this for the CPU backend too.

PR #2781, opened 2019-04-24
[utils] Add the imagenet model params to caffe2_pb_runner

enferex:
*Description*
Add the model parameters for imagenet data to the
caffe2 helper utility.  These values were discovered by comparing
matching the `tests/images/imagenet/{cat,dog,zebra}` images with
the image names listed in Table 8 here: https://arxiv.org/pdf/1311.2524.pdf

I'm not sure if calling this 'imagenet' is correct, but the dataset
associated to the aforementioned paper was provided by the image-net
project: http://www.image-net.org/challenges/LSVRC/2013/

*Testing*
```
python3 /glow/utils/caffe2_pb_runner.py \
    -i /glow/tests/images/imagenet/cat_285.png \
    -d bvlc_reference_rcnn_ilsvrc13/ \
    -m imagenet
```

The result should be something like, of importance is 'index 58' below,
which happens to match the 'cat' label in Table 8 of the paper referenced above.

```
Initial img shape: (224, 224, 4)
Shape of final_image: (1, 3, 224, 224)
The blobs in the workspace after FeedBlob: ['data']
Max index is 58
Predicted class at index 58 with probability 1.9546699523925781
Number of classes 200
Sum of probabilities is -398.8056676387787
```

### DIFF
--- a/utils/caffe2_pb_runner.py
+++ b/utils/caffe2_pb_runner.py
@@ -118,7 +118,7 @@ model_props = dict(
         224,
         3,
     ),
-    imagenet=Model(
+    rcnn_ilsvrc13=Model(
         "data",
         mode_0to256,
         224,
--- a/utils/caffe2_pb_runner.py
+++ b/utils/caffe2_pb_runner.py
@@ -118,6 +118,12 @@ model_props = dict(
         224,
         3,
     ),
+    imagenet=Model(
+        "data",
+        mode_0to256,
+        224,
+        3,
+    ),
 )
 
 MODEL = args.model_name


jfix71:
@enferex Thanks for updating this. I'm not sure that adding in an `imagenet` model makes sense here -- it looks like you tested it specifically with `bvlc_reference_rcnn_ilsvrc13`, and not all imagenet models will have these same parameters (e.g. we already have the other imagenet models listed there like resnet50, squeezenet, etc. with different `blob_name`/`image_mode_op`). I think it might make more sense to add this as a new Model with name something like `RCNN_ILSVC13`. What do you think?

enferex:
> RCNN_ILSVC13

Thanks for the info, and that is a superior suggestion!  Thanks!

enferex:
I was using this caffe2 model as a comparison test against  the onnx model with Glow.  I figured I'd add a model for it to the caffe2_pb_runner; however, I never modified the `download_caffe2_models.py`.  The caffe2 model was snagged from https://github.com/caffe2/models.  If you think it makes sense, we might want to add those .pb files to the fb-glow-assets s3 bucket and update the download script accordingly.

jfix71:
>If you think it makes sense, we might want to add those .pb files to the fb-glow-assets s3 bucket and update the download script accordingly.

Sure. It may also make sense to add the Caffe2 version to the `run.sh` script as well -- we already have duplicates for other model's Caffe2 vs. ONNX versions. Did you try running it on Glow as well? 

I've uploaded them at:

```
s3://fb-glow-assets/models/bvlc_reference_rcnn_ilsvrc13/predict_net.pb
s3://fb-glow-assets/models/bvlc_reference_rcnn_ilsvrc13/init_net.pb
```

enferex:
I'm not certain we want this patch.   It ran fine using caffe2, however; if the caffe2 model is run through Glow, it will fail to compile since this model uses legacy caffe padding (which is deprecated).  Glow should detect that case a prematurely terminate.  I'm not certain it's useful to add this to the caffe2_pb_runner, as it might just confuse users who expect the model to run in both caffe2 and Glow.

jfix71:
>if the caffe2 model is run through Glow, it will fail to compile since this model uses legacy caffe padding (which is deprecated)

Ah, I was unaware of this. Yeah I think we can skip this PR. Thanks for working on this @enferex!

PR #2791, opened 2019-04-25
[OpenCL] Remove the spurious dependency on CPURuntime

opti-mix:
This dependency was introduced by mistake as a result of a copy&paste.


SplitInfinity:
Great find!

PR #2793, opened 2019-04-25
Remove cloning of functions for quantization and profiling

jfix71:
*Description*: As discussed in #2700, this removes cloning the function for quantization and profiling. If the caller wants the original function after quantization/profiling it should clone the function prior to calling `quantizeFunction()`/`profileQuantization()`.

*Testing*: All tests still pass.


jackm321:
Nice

PR #2799, opened 2019-04-25
Changed loader to use HostManager rather than ExecutionEngine.

gcatron:
*Description*: This migrated loader (used in image-classifier) to use HostManager rather than ExecutionEngine. This means that image-classifier now executes through the new runtime.
*Testing*: Ninja test, tested image-classifier on resnet50 manually for cpu, interpreter, and OpenCL. Tested quantization. All tests in run.sh also pass.
*Documentation*: NA

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -91,9 +91,10 @@ public:
   /// operation. This function consumes the \p module so any pointers to data
   /// contained within the module should be considered invalid. If \p
   /// saturateHost is set to true the HostManager will try to use all available
-  /// devices on the host.
+  /// devices on the host. If \p profiling is true,the HostManager will check
+  /// that each function is not partitioned.
   llvm::Error addNetwork(std::unique_ptr<Module> module,
-                         bool saturateHost = false);
+                         bool saturateHost = false, bool profiling = false);
 
   /// Given \p networkName removes that network from the host. This also
   /// removes the network from any backends setup to execute it.
@@ -115,6 +116,13 @@ public:
   RunIdentifierTy runNetwork(llvm::StringRef networkName,
                              std::unique_ptr<ExecutionContext> context,
                              ResultCBTy callback);
+
+  /// A wrapper around runNetwork that provides a blocking interface for an
+  /// inference request. Runs the network provided in \p networkName using \p
+  /// bindings for placeholder bindings. \returns an llvm::Error indicating
+  /// success or failure.
+  llvm::Error runNetworkBlocking(llvm::StringRef networkName,
+                                 PlaceholderBindings &bindings);
   HostManager(std::vector<std::unique_ptr<DeviceConfig>> configs);
 
   /// Initialize the HostManager with the given \p configs creating one
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -60,7 +60,7 @@ HostManager::init(std::vector<std::unique_ptr<DeviceConfig>> configs) {
 HostManager::~HostManager() { llvm::toString(clearHost()); }
 
 llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
-                                    bool saturateHost) {
+                                    bool saturateHost, bool profiling) {
   std::lock_guard<std::mutex> networkLock(networkLock_);
   auto functions = module->getFunctions();
   for (auto &F : functions) {
@@ -79,14 +79,25 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
     info.backendKind = device.second->getBackendKind();
     deviceInfo.push_back(info);
   }
-
   auto partitioner = Partitioner(module.get(), deviceInfo, saturateHost);
   RETURN_IF_ERR(partitioner.Partition());
   auto nodeList = std::move(partitioner.getPartitionResult());
+  if (profiling) {
+    // Check that all functions were not partitioned.
+    for (auto &network : nodeList) {
+      if (network.nodes.size() > 1) {
+        return MAKE_ERR(
+            GlowErr::ErrorCode::RUNTIME_ERROR,
+            "Failed to add network for profiling: Network was "
+            "partitioned, this is likely because the network was "
+            "larger than the configured memory of a single device manager.");
+      }
+    }
+  }
 
   RETURN_IF_ERR(provisioner_->provision(nodeList, *module));
 
-  // Clear functions and constants contents from the module then put it a
+  // Clear constants contents from the module then put it in a
   // shared_ptr to be shared between all of the networks created from each
   // function in the module.
   module->strip();
@@ -161,6 +172,32 @@ llvm::Error HostManager::clearHost() {
   }
   networks_.clear();
   return errContainer.get();
+}
+
+llvm::Error HostManager::runNetworkBlocking(llvm::StringRef networkName,
+                                            PlaceholderBindings &bindings) {
+  std::unique_ptr<PlaceholderBindings> phBindings(&bindings);
+  std::unique_ptr<ExecutionContext> context =
+      llvm::make_unique<ExecutionContext>(std::move(phBindings));
+  std::promise<void> runPromise;
+  auto fut = runPromise.get_future();
+  llvm::Error runErr = llvm::Error::success();
+  runNetwork(
+      networkName, std::move(context),
+      [&runPromise, &runErr](runtime::RunIdentifierTy, llvm::Error err,
+                             std::unique_ptr<ExecutionContext> contextPtr) {
+        // Don't delete ph bindings since they were created from a passed in
+        // reference.
+        std::unique_ptr<PlaceholderBindings> phBind =
+            contextPtr->movePlaceholderBindings();
+        phBind.release();
+
+        runErr = std::move(err);
+        runPromise.set_value();
+      });
+
+  fut.wait();
+  return runErr;
 }
 
 RunIdentifierTy
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -18,10 +18,10 @@
 
 #include "glow/Base/Tensor.h"
 #include "glow/Converter/TypeAToTypeBFunctionConverter.h"
-#include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/IR/IR.h"
 #include "glow/Quantization/Quantization.h"
 #include "glow/Quantization/Serialization.h"
+#include "glow/Runtime/RuntimeTypes.h"
 
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
@@ -316,7 +316,7 @@ void Loader::compile(PlaceholderBindings &bindings) {
     // lowered, however all lowered and unlowered components have a profile, and
     // so the backend can lower however it prefers and always find all of its
     // NodeValue's quantization parameters.
-    ::lower(F_, &loweredMap_, EE_.getBackend());
+    ::lower(F_, &loweredMap_, backend_);
 
     quantization::QuantizationConfiguration quantConfig{
         deserializeFromYaml(loadProfileFileOpt)};
@@ -327,8 +327,8 @@ void Loader::compile(PlaceholderBindings &bindings) {
     quantConfig.enableRowwise = enableRowwiseOpt;
     quantConfig.assertAllNodesQuantized = assertAllNodesQuantizedOpt;
 
-    quantization::quantizeFunction(F_, quantConfig, *EE_.getBackend(),
-                                   loweredMap_, keepOriginalPrecisionForNodes);
+    quantization::quantizeFunction(F_, quantConfig, *backend_, loweredMap_,
+                                   keepOriginalPrecisionForNodes);
   }
 
   if (convertToFP16) {
@@ -341,19 +341,32 @@ void Loader::compile(PlaceholderBindings &bindings) {
 
   CompilationContext cctx;
   cctx.mode = CompilationMode::Infer;
+  // Store a raw pointer to the Module, we pass the unique_ptr to HostManager
+  // but the Module is stored by Hostmanager so the pointer will remain valid.
+  auto module = M_.get();
+
   if (emittingBundle()) {
     // Emit IR for the graph, compile it and save as a bundle.
-    EE_.save(F_, cctx, emitBundle, networkName);
+    auto error = ::glow::optimizeFunction(F_, *backend_, cctx);
+    EXIT_ON_ERR(std::move(error));
+    backend_->save(F_, emitBundle, networkName);
   } else {
     // Emit IR for the graph and compile it.
-    EE_.compile(F_, cctx);
+    auto error = hostManager_->addNetwork(std::move(M_), /*saturateHost*/ false,
+                                          profilingGraph());
+    EXIT_ON_ERR(std::move(error));
   }
-
   if (dumpGraphOpt) {
-    F_->dump();
+    for (auto function : module->getFunctions()) {
+      function->dump();
+    }
   }
   if (!dumpGraphDAGFileOpt.empty()) {
-    F_->dumpDAG(dumpGraphDAGFileOpt.c_str());
+    for (auto function : module->getFunctions()) {
+      std::string filename =
+          function->getName().str() + "_" + dumpGraphDAGFileOpt;
+      function->dumpDAG(filename.c_str());
+    }
   }
 }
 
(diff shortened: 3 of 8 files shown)


include/glow/Runtime/RuntimeTypes.h
```diff
@@ -138,6 +138,16 @@ class DeviceConfig {
   void setName(llvm::StringRef name) { name_ = name; }
 };
 
+/// OpenCL Device Manager config object. This contains the information needed to
```
nadavrot: Why do we have target-specific code in the target-independent part of the compiler? Do we need to add stuff like this for all of our targets?

tools/loader/Loader.cpp
```diff
@@ -379,7 +382,26 @@ void Loader::runInference(PlaceholderBindings &bindings, size_t batchSize) {
     timer.startTimer();
   }
   for (unsigned i = 0; i < iterationsOpt; i++) {
-    EE_.run(bindings);
```
nadavrot: whoa the single run() method turned into the code below?  We should introduce a convenience method to wrap all of this.

tools/loader/Loader.cpp
```diff
@@ -427,7 +449,19 @@ Loader::Loader(int argc, char **argv) {
     caffe2NetDescFilename_ = modelPathOpt[0];
     caffe2NetWeightFilename_ = modelPathOpt[1];
   }
+  M_.reset(new Module);
```
nadavrot: Why do we need target specific code and all of this logic here? This totally complicates the loader.

include/glow/Runtime/RuntimeTypes.h
gcatron: I moved it here so that RuntimeTypes.h had every data type needed to interface with the Runtime. I can move it back and also include OpenCLDeviceManager.h

nadavrot: Why would it need to know about the specifics of one target? All of this should be hidden behind virtual calls and interfaces.

tools/loader/Loader.cpp
```diff
@@ -427,7 +449,19 @@ Loader::Loader(int argc, char **argv) {
...
+  std::vector<std::unique_ptr<runtime::DeviceConfig>> configs;
+  if (ExecutionBackend == BackendKind::OpenCL) {
+
+    auto clConfig = new runtime::OpenCLDeviceConfig();
```
nadavrot: We need to have some method that accepts the device enum and creates the device config. We should not inline this code into every user (call site).

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -97,9 +97,9 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
 
   RETURN_IF_ERR(provisioner_->provision(nodeList, *module));
 
-  // Clear everything but placeholders from the module then put it a shared_ptr
-  // to be shared between all of the networks created from each function in the
-  // module.
+  // Clear everything but placeholders from the module then put it a
```
jfix71: NFC here? Maybe revert

tools/loader/Loader.cpp
```diff
@@ -366,7 +370,26 @@ void Loader::runInference(PlaceholderBindings &bindings, size_t batchSize) {
...
+    std::unique_ptr<ExecutionContext> context =
+        llvm::make_unique<ExecutionContext>(std::move(phBindings));
+    std::promise<void> runPromise;
+    auto fut = runPromise.get_future();
+    llvm::Error runErr = llvm::Error::success();
+    hostManager_->runNetwork(
+        modelPathOpt[0], std::move(context),
+        [&runPromise, &runErr](runtime::RunIdentifierTy, llvm::Error err,
+                               std::unique_ptr<ExecutionContext> contextPtr) {
+          // Don't delete ph bindings.
```
bertmaher: So, maybe this is out of scope for this precise diff, but the dance of creating a unique_ptr<ExecutionContext>, std::move-ing it in to a lambda, then std::move-ing out the unique_ptr<PlaceholderBindings> looks really sketchy.  Can that be avoided?

tools/loader/Loader.cpp
opti-mix: Or may be we could have a virtual method in DeviceManager and ask the DeviceManager to create an empty device config?  This way HM does not need to know the specifics...

tools/loader/Loader.h
```diff
@@ -37,10 +38,14 @@ class Loader {
   std::string caffe2NetWeightFilename_;
   /// ONNX model file name.
   std::string onnxModelFilename_;
-  /// Execution engine for compiling and running.
-  ExecutionEngine EE_{};
+  /// Host Manager for running the model.
+  std::unique_ptr<glow::runtime::HostManager> hostManager_;
+  /// Backend used for saving bundle and quantization.
+  glow::Backend *backend_;
```
opti-mix: We may need multiple backends here to support heterogeneous partitioning. But this can be done in a separate PR.

include/glow/Runtime/RuntimeTypes.h
opti-mix: See my comment about having a virtual method in DeviceManager to create device specific configs. We won't need to expose those backend-specific device configs if we'd use that approach or something similar.

tools/loader/Loader.cpp
gcatron: That would mean we include device discovery and assignment to the runtime. Is that preferred? For example if a user wants to target 3 specific devices out of a pool of 5, how do we let the user of the library select devices? Or is that something we are ok with not supporting?

tools/loader/Loader.cpp
nickgg: There are two things interacting here:
* The Runtime interface is a unique_ptr because explicit ownership transfers reduce the incidence of a class of production bugs. We can always make writing the code easier to write by switching back to raw pointer, and so long as someone else is going to do all the object lifetime bug debugging i'm fine with that :)
* This code creates the PlaceholderBindings on the stack, which requires the hack below (because despite being in the unique_ptr the code below doesn't really own the PlaceholderBindings and we have to make sure it doesn't get deleted). If we created the PlaceholderBindings on the heap, or always as part of the ExecutionContext then it's not so big a deal. Alternatively if we had a better continuation abstraction (such as folly::Future) we wouldn't need the callbacks, but in the version of standard C++ we support this was what we have to work with.

tools/loader/Loader.h
nickgg: This backend is just for saving and profiling, which we should make part of the runtime (since we'll need the Partitioner anyway).

tools/loader/Loader.cpp
nickgg: I think this is much more complicated because we have to create a promise, a future, an error to move the produced error into, etc. We could make this much simpler by splitting the `runInference` function into two parts:

```
void Loader::runInferences(std::unique_ptr<ExecutionContext> context) {
  assert(!emittingBundle() &&
         "No inference is performed in the bundle generation mode.");
  if (timeOpt) {
    timer_ = llvm::Timer timer("Infer", "Infer");
    timer_.startTimer();
  }

   auto iterations = iterationsOpt;
   runOneInference(std::move(context), --iterations);
}

void runOneInference(std::unique_ptr<ExecutionContext> context, size_t iterations) {
    hostManager->runNetwork(modelPathOpt[0], std::move(context), [this, iterations] 
        (runtime::RunIdentifierTy, llvm::Error err, std::unique_ptr<ExecutionContext> context) {
             EXIT_ON_ERR(std::move(runErr));
             if (iterations) {
                runOneInference(std::move(context), --iterations);
             } else {
                 if (timeOpt) {
                     timer_.stopTimer();
                     llvm::outs() << llvm::formatv("Wall time per item (s): {0:f4}\n",
                                               timer_.getTotalTime().getWallTime() /
                                               iterationsOpt / batchSize_);
                }
           }
      });
}
```

If you pull the timer code out into helper methods on the Loader class this is quite small.

nickgg: It's even simpler if you don't want to force the inferences to run sequentially and can just queue em all up in the HostManager and wait for `iterationsOpt` results.

tools/loader/Loader.cpp
bertmaher: I guess I'm fine with the API taking a unique_ptr since runNetwork is intended to "consume" the bindings, and my objection is really to the second part, which is the weird move-out-and-release pattern that prevents the consumption from happening.

Maybe we should be copying the bindings on each iteration?  That adds a bit of overhead but is possibly more realistic anyways since usually a network run will need to make new bindings anyways.

nadavrot: Why would we need new bindings? We could load new inputs into the existing input tensors by using DMA or RDMA, and in that case reuse the bindings forever. I think that it makes more sense to keep the bindings in a place that ensures the length of their lifetime and just pass them as pointer.

bertmaher: @nadavrot I agree with your conclusion that we should keep the bindings in a place that ensures the length of lifetime.  The current code that passes unique_ptrs around and strategically releases them appears to be both more verbose and more error-prone.

But to your point of why we need new bindings: on the input side, we currently bind tensors directly to whatever memory we're given (e.g., https://github.com/pytorch/glow/blob/master/lib/Onnxifi/Base.cpp#L132) and rely on the device-specific code to manage DMA buffer re-use (if needed/possible).  But that means a call to runNetwork needs updated bindings since the caller doesn't guarantee where the tensors are in host memory.

nadavrot: @bertmaher Thank you for explaining this to me. Makes sense.

nickgg: > The current code that passes unique_ptrs around and strategically releases them appears to be both more verbose and more error-prone.

Yes, agree this is bad - but it's not an intrinsic issue of the interface, its because the loader and the runtime disagree here. We should change the loader.

tools/loader/Loader.cpp
```diff
@@ -343,10 +345,12 @@ void Loader::compile(PlaceholderBindings &bindings) {
   cctx.mode = CompilationMode::Infer;
   if (emittingBundle()) {
     // Emit IR for the graph, compile it and save as a bundle.
-    EE_.save(F_, cctx, emitBundle, networkName);
+    ::glow::optimizeFunction(F_, *backend_, cctx);
+    backend_->save(F_, emitBundle, networkName);
   } else {
     // Emit IR for the graph and compile it.
-    EE_.compile(F_, cctx);
+    auto error = hostManager_->addNetwork(std::move(M_));
```
artemrakhov: Now after we `std::move` the module, `Function` object becomes unavailable. But it is still possible to call `loader.getFunction()`! This may or may not bring unexpected corruption.

For example, `ModelRunner.cpp` misbehaves when it's trying to `getFunction()->getName()`. Because this happens after compilation.

nickgg: Compilation is potentially destructive over the Function, and Partitioning definitely makes access to the original function make less sense. I think while it's convenient to keep using the Function after compilation it's a bad pattern we should remove.

artemrakhov: I agree that it's a bad pattern. But it's not an obvious pattern. How can we make it clear to everybody?

nickgg: @artemrakhov unique_ptr!

gcatron: I've removed unnecessary calls to getFunction that were just getting the name. I've also added comments to getFunction and getModule that they should not be called after compile.

gcatron:
Refactored quite a few things,
Added a helper to HostManager that provides a blocking interface for inference requests.
Added the plumbing to ensure Partitioner does not partition the function when we are profiling it.
Removed device specific logic from loader. A later PR can add command line support for multiple devices with the new string map configuration object.

tools/loader/Loader.h
```diff
@@ -37,20 +38,26 @@ class Loader {
...
+  std::unique_ptr<Module> M_;
   /// A map between quantization profiling names of NodeValues that were lowered
   /// from each other. Maps to a set of names of NodeValues and their NodeKinds
   /// that were replaced by the NodeValue (whose output name is the key) that
   /// replaced them.
   LoweredInfoMap loweredMap_;
 
 public:
-  /// Getter for the Function.
+  /// Getter for the Function. This should not be called after compile since the
```
jfix71: Is this still true since strip is now only stripping payloads off Constants?

tools/loader/Loader.h
```diff
@@ -37,20 +38,26 @@ class Loader {
...
+  /// compile process is destructive on the original function.
   Function *getFunction() { return F_; }
-  /// Getter for the Module.
+  /// Getter for the Module. This should not be called after compile since the
```
jfix71: Is this still true since strip is now only stripping payloads off Constants?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -91,13 +91,14 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
       RETURN_IF_ERR(::glow::optimizeFunction(F, *backend_, cctx));
     }
   }
-  auto partitioner = Partitioner(module.get(), deviceInfo, saturateHost);
+  auto partitioner =
```
jfix71: Can you put a comment here saying that if profiling then the Partitioner should only return a single partition? Perhaps could add an assert for this too?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -99,7 +99,7 @@ class HostManager final {
   /// saturateHost is set to true the HostManager will try to use all available
   /// devices on the host.
```
jfix71: Update doxygen

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -121,6 +121,11 @@ class HostManager final {
   RunIdentifierTy runNetwork(llvm::StringRef networkName,
                              std::unique_ptr<ExecutionContext> context,
                              ResultCBTy callback);
+
+  /// A wrapper around runNetwork that provides a blocking interface for an
+  /// inference request.
```
jfix71: Can you add references to the arguments and return?

include/glow/Partitioner/Partitioner.h
```diff
@@ -163,7 +167,7 @@ class Partitioner {
   /// found in Module. The \p devices provides the cost model related to
   /// devices.
```
jfix71: Update doxygen

tools/loader/Loader.cpp
```diff
@@ -341,20 +341,25 @@ void Loader::compile(PlaceholderBindings &bindings) {
...
-  } else {
-    // Emit IR for the graph and compile it.
-    EE_.compile(F_, cctx);
-  }
 
   if (dumpGraphOpt) {
     F_->dump();
   }
   if (!dumpGraphDAGFileOpt.empty()) {
     F_->dumpDAG(dumpGraphDAGFileOpt.c_str());
```
jfix71: This is now dumping the unoptimized graph instead of the optimized one. Since the Function should still be valid now, can you move this back below the calls to `addNetwork()`/`save()` below?

tools/loader/Loader.h
gcatron: The partitioner can still remove the original function. If profiling or no partitioning happens then it's safe.

lib/Runtime/HostManager/HostManager.cpp
gcatron: I'll add a comment, perhaps return an error rather than assert?

tools/loader/Loader.cpp
gcatron: I moved them back below addNetwork()/save() and now walk all functions in the module to dump them. This handles the case of partitioning.

nadavrot:
Nice!

wyessen:
@gcarton

I see that `ImageClassifier` is now using the `HostManager` directly. At the same time, I see that `Loader::runInference()` has been refactored to use the `HostManager` through the new `HostManager::runNetworkBlocking()`. 

This thread says that `ImageClassifier` has been tested (which, again, uses the `HostManager` directly). But has the new refactored `Loader::runInference()` been tested? It is giving me grief, have been debugging it for a few hours now (the implementation of `HostManager::runNetworkBlocking` is rather convoluted). Somehow it is segfaulting somewhere around the call to `fut.wait()` in `HostManager::runNetworkBlocking()`.

gcatron:
@wyessen 
Image Classifier has both execution paths, it calls into Hostmanager directly when the iterations flag is set, this allows for concurrent execution for benchmarking. When the flag isn't set it calls loader.runInference(). I just tested it locally and it is behaving for me but it's quite likely I missed something in the port. 
If you're segfaulting around the fut.wait() call that means it's happening while trying to execute on the device. Do you have a backtrace of the segfault? Not sure what's causing it but it would be good to catch it as an error rather than segfault.

wyessen:
@gcatron Thanks, I see it. Has this been tested on Linux?

gcatron:
> @gcatron Thanks, I see it. Has this been tested on Linux?

Yes, it's been tested on Centos 7.

wyessen:
@gcatron 

Alright, this was somewhat of a red herring, but I think I've finally caught it. See here: https://github.com/pytorch/glow/issues/3221.

PR #2800, opened 2019-04-25
Add HostManager check that all nodes are supported after optimization

jfix71:
*Description*: This is something we do in the ExecutionEngine. We should also check it somewhere in the HostManager, especially as PRs like #2799 move to using HostManager instead. This error represents some internal compilation error during optimizeFunction.

*Testing*: All unit tests pass.

### DIFF
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -82,7 +82,7 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
   CompilationContext cctx;
   cctx.mode = CompilationMode::Infer;
   cctx.backendOpts.autoInstrument = true;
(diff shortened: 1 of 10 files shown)


opti-mix:
@jfix71 I wonder if HostManager is the right place for this functionality? May be it should be a utility function called from HostManager and may be other places. Semantically it seems to be more like a verifier kind of functionality.

opti-mix:
Also, may be we should always call this check at the end of `optimizeFunction`?

jfix71:
@opti-mix @jackm321 I've changed this so that `optimizeFunction()` returns an `llvm::Error`. LMK what you think. All unsuccessful calls to `optimizeFunction()` will exit if there's an unsupported Node, except for the one in HostManager, which returns the error up the stack.

PR #2802, opened 2019-04-26
[importer] Detect legacy (Caffe1) padding and exit.

enferex:
*Description*:
Caffe1 legacy padding rounds the shape dimensions up for MaxPool
operators and Caffe2 rounds down.  According to Caffe2's
caffe2_legacy.proto definition, legacy padding is deprecated.

During graph generation, Glow might insert Caffe2 MaxPool nodes unaware
of the padding strategy.  If a MaxPool node is defined in the Caffe2
file, that follows the legacy strategy, we result in a model where some
MaxPool nodes will round their shapes up and others will round down.
Instead of introducing this confusion, we reject the deprecated logic in
favor of what we have always done, support non-legacy/Caffe2 padding.

See: https://github.com/pytorch/pytorch/blob/master/caffe2/proto/caffe2_legacy.proto

*Testing*:
`ninja test`  ... nothing should fail, in particular the OperatorTest should pass.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -48,6 +48,11 @@ using ArgumentDictionaryTy =
 /// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128
 const int32_t OFFSETSHIFT = 128;
 
+/// Legacy padding modes supported in caffe2.  These are used by MaxPool
+/// operators, and are defined in caffe2_legacy.proto in the caffe2 source
+/// tree.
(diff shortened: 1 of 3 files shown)


enferex:
> We need a test in Caffe2ImporterTest where legacy_pad is set.
> 
> (OperatorTests do not go this route).

Fair point, and my apologies for not supplying a corresponding test.  I have created one as the second patch associated to the PR.

nadavrot:
@enferex Please squash the commits in the PR before landing.

enferex:
Thanks for the reviews and feedback.  I cannot squash+merge this as I don't have permissions.

PR #2807, opened 2019-04-26
[readme] Update the required llvm version from 5 to 7. NFC.

enferex:
*Description*:
The CMakeLists.txt file requires llvm 7.0 or higher to build Glow.  This patch
updates the version number in the README documentation to reflect the configuration file
constraint.  We should update the instructions in the  building
on Ubuntu 16.04 section too, as that still refers to llvm 6.0.  I was
uncomfortable modifying that section as I cannot test it as written on
my setup at the moment.


nadavrot:
CC @mortzur

PR #2808, opened 2019-04-27
[cmake] Disable OpenCL support by default.

### LINKED ISSUE #2466: Default values for CPU and OpenCL flags contradict to the docs
Here we enable building with CPU and OpenCL by default:
https://github.com/pytorch/glow/blob/8e312ac3c3dd5a8aeb32ccdd6eeb4a521f7aa7a2/CMakeLists.txt#L8
This contradicts to our docs here:
https://github.com/pytorch/glow/blob/master/docs/Building.md

We need to either change defaults to OFF, or go through the docs verify them. I suggest the former, because not everyone has LLVM and OpenCL installed.

Moreover, we need to make sure that CI checks configurations without CPU. Right now we always build with CPU support, so it is possible to break building with CPU=OFF and not notice it.

After PR #2461 lands, we'll have one more flag to disable by default and also test ON/OFF on CI.

enferex:
*Description*:
This updated configuration matches the documentation at
https://github.com/pytorch/glow/blob/master/docs/Building.md
Notably, the line:
"By default Glow builds with only the interpreter backend enabled."

It was also mentioned that not everyone has OpenCL installed.  I think
it makes sense to not make OpenCL a requirement; however, I can also see
where it might be expected as a requirement.  I figured I'd at least put
this PR out for discussion.

*Testing*:
The unittests should still build and run correctly:
`ninja all && ninja test`

Fixes #2466

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ enable_testing()
 
 option(GLOW_WITH_CPU "Build the LLVM-based JIT CPU backend" ON)
 option(GLOW_WITH_LLVMIRCODEGEN "Build the LLVM-based code generation library" ON)
-option(GLOW_WITH_OPENCL "Build the OpenCL backend" ON)
+option(GLOW_WITH_OPENCL "Build the OpenCL backend" OFF)
 option(GLOW_WITH_HABANA "Build the Habana backend" OFF)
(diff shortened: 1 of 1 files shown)


arunm-git:
Seems reasonable to me. Without this change, I just hit an issue where enabling an opencl backend test failed due to no opencl platforms.

PR #2813, opened 2019-04-29
[LLVMIRCodeGen] Minor improvements making LLVMIRCodeGen more customizable by backends

opti-mix:
- Make it possible for backends to provide custom logicto decide which symbols need to be preserved during the LLVM's internalizeModule pass

  Some backends may have a custom logic for deciding which symbols need to be preserved.

- Provide APIs for setting target, arch and cpu properties of LLVMBackend

  Backends can now override these settings by setting these properties to appropriate, backend-specific values.

- Make it possible for backends to provide custom logic to decide which calls are eligible for specialization

### DIFF
--- a/include/glow/LLVMIRCodeGen/LLVMBackend.h
+++ b/include/glow/LLVMIRCodeGen/LLVMBackend.h
@@ -31,9 +31,27 @@ class PlaceholderBindings;
 class LLVMIRGen;
 
 class LLVMBackend : public BackendUsingGlowIR {
-public:
-  LLVMBackend() = default;
+  /// Target used by this backend.
+  std::string target_;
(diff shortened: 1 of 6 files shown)


include/glow/LLVMIRCodeGen/LLVMBackend.h
```diff
@@ -31,9 +31,27 @@ class PlaceholderBindings;
 class LLVMIRGen;
 
 class LLVMBackend : public BackendUsingGlowIR {
-public:
-  LLVMBackend() = default;
+  /// Target used by this backend.
```
nadavrot: Maybe this should be std::string in case the original string will go away.

PR #2817, opened 2019-04-29
Enable HostManagerTest for multiple backends

arunm-git:
*Description*:

Replace pow with tanh so the test can be enabled for different backends.

*Testing*:

ninja all && ninja test

(This will fail for opencl if you don't have that setup. But #2808 will address that)

### DIFF
--- a/tests/unittests/HostManagerTest.cpp
+++ b/tests/unittests/HostManagerTest.cpp
@@ -27,15 +27,20 @@ using namespace glow::runtime;
 using DAGNodePairTy = std::pair<std::vector<std::unique_ptr<DAGNode>>,
                                 std::vector<std::unique_ptr<DAGNode>>>;
 
-class HostManagerTest : public ::testing::Test {};
+class HostManagerTest : public ::testing::TestWithParam<BackendKind> {
+public:
+  void SetUp() override { backendKind = GetParam(); }
+  BackendKind backendKind;
+};
+
 std::unique_ptr<Module> setupModule(unsigned functionCount) {
   std::unique_ptr<Module> module = llvm::make_unique<Module>();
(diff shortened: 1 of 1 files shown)


nickgg:
I'm curious what we gain from this, we have tests of each backend already - this is testing the HostManager component which isn't BackendKind aware really. This will make the test 4x longer right? I wonder if we should actually test a Mock backend instead?

arunm-git:
In theory, it could expose if there are any backend-specific differences in the HostManager that lead to problems. The bigger context is that if we have unit tests that are specific to a particular backend, they should be run for all backends we care about. 

Agreed that it is of questionable marginal value in this case. The increase in cost is also small however - it's less than 4X and (4X of a very small number).

In any case, when I solicited feedback internally on the diff, I was told that it could be useful. Hence this PR.

nickgg:
HostManagerTest takes about 6s doesn't it? 6->24 doesn't seem negligible.

We don't have any backend specific differences in the HostManager right now, and I'm not aware of any planned.

arunm-git:
It doesn't appear to be linear. I suspect most of the costs are fixed and not a function of number of test cases:
CPU only: 6325ms
CPU + Interpreter: 6409ms
(on my laptop)

Another thing to note is that the test currently fails for opencl - might be a test bug though.

All this said, I don't have a strong preference on this PR and am happy to drop it.

CC: @bertmaher and @SplitInfinity who were sorta interested in seeing it added.

nickgg:
huh thats weird wouldn't all the costs be in compile

SplitInfinity:
If we have some other E2E test that checks all backends (that's not resnet-runtime, lol), I'm okay with abandoning this.

arunm-git:
Keeping this alive till I figure out what the opencl error is.

PR #2820, opened 2019-04-30
Add Gradient for MatMulNode.

putivsky:
*Description*:
Training some sparse features CNN requires Gradient for MatMulNode
*Testing*:
Run unit test: checkMatMulToGradTest
*Documentation*:

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -232,6 +232,31 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       continue;
     }
 
+    if (N->getKind() == Kind::MatMulNodeKind) {
+      MatMulNode *MMN = cast<MatMulNode>(N);
+      // Get gradient.
+      NodeValue OutputG = map.getGradient(MMN->getResult());
+
+      // Get LHS/RHS inputs and their transpose presentations.
+      NodeValue InputLHS = MMN->getLHS();
+      NodeValue InputRHS = MMN->getRHS();
+      auto *LT = G->createTranspose("lhs.T", InputLHS, {1, 0});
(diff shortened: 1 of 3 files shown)


lib/Graph/Grad.cpp
```diff
@@ -232,6 +232,31 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       continue;
     }
 
+    if (N->getKind() == Kind::MatMulNodeKind) {
+      MatMulNode *MMN = cast<MatMulNode>(N);
+      // Get gradient
+      NodeValue OutputG = map.getGradient(MMN->getResult());
+
+      // Get LHS/RHS inputs and their transpose presentations
```
nadavrot: Period.

lib/Graph/Grad.cpp
```diff
@@ -232,6 +232,31 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       continue;
     }
 
+    if (N->getKind() == Kind::MatMulNodeKind) {
+      MatMulNode *MMN = cast<MatMulNode>(N);
+      // Get gradient
```
nadavrot: Period.

tests/unittests/GraphGradTest.cpp
```diff
@@ -208,3 +208,32 @@ TEST(GraphAutoGrad, checkConvertToGradTest) {
   EE.compile(CompilationMode::Train, TF);
   EE.compile(CompilationMode::Infer, F);
 }
+
+/// Check that we can differentiate functions that use MatMulNode.
+TEST(GraphAutoGrad, checkMatMulToGradTest) {
```
SplitInfinity: How does this test show that the gradient for MatMul is correct? The graph gets differentiated and compiled, but it is never run.

Issue #2776, opened 2019-04-23
[Runtime efficiency] Don't copy intermediate tensors for DAGNodes with only one child


jackm321:
We have seen from perf profiling that propagating placeholder in the exeuctor is expensive in large part because of copying tensors from one DAGNode to another. One way to reduce the number of copies that occur is to only copy backing tensors for Placeholders that have multiple children and for DAGNodes with a single child just hand over ownership to that child node.

rdzhabarov:
This is fixed.

PR #2821, opened 2019-04-30
[runtime] Remove all tensor copies and almost all allocs from the executor

### LINKED ISSUE #2773: [Runtime efficiency] Don't zero tensors unnecessarily
Tensors are zeroed on initialization, investigate if/when this can be skipped

nickgg:
*Description*: The ThreadPoolExecutor is doing a lot of copies, which makes it pretty slow - when running realistic size traffic through the runtime we spent more than 3ms in the executor preparing Tensors. Fortunately the executor doesn't need to do any copies at all - since we serialize any nodes that would share an input/output pair, so I've cut them all out. Additionally, we were allocating new Tensor buffers for all sub-graphs which we don't need either.

This has a significant impact on Runtime performance - about a **15x reduction in introduced latency** in the test I ran (scales with input size though, obviously):

before:
image
after:
image

*Testing*: all tests under debug, release and ASAN/UBSAN.
*Documentation*: fixes #2773 and #2776 by removing affected code (I think).


nickgg:
By my measure this cuts a few seconds off the unittest run time as well, hooray

nadavrot:
Nice!

lib/Graph/PlaceholderBindings.cpp
```diff
@@ -97,6 +97,12 @@ void PlaceholderBindings::clear() {
   nameMap_.clear();
 }
 
+void PlaceholderBindings::erase(Placeholder *P) {
+  nameMap_.erase(P->getName());
```
nadavrot: Let's add an assert that P is in the map.

include/glow/Graph/PlaceholderBindings.h
```diff
@@ -88,6 +88,9 @@ class PlaceholderBindings final {
   /// tensors.
   void clear();
 
+  /// Removes the Tensor backing Placeholder \p P;
```
nadavrot: What happens if P is not in the map? 
/// If P is not in the map nothing happens. 

or 

/// P must be a valid placeholder in the map.

nrsatish:
This is great, thanks!

jackm321:
@nickgg this does what #2745 does but in a more complete way, let's close that one. In the longer term, we should discuss pooling placeholder bindings and not releasing this memory between runs.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -464,6 +407,9 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
   }
 
   if (noNodesInflight) {
+    // remove the intermediate placeholders so we don't leak them to the caller.
```
jackm321: Remove

lib/Runtime/Executor/ThreadPoolExecutor.h
```diff
@@ -127,8 +125,7 @@ class ExecutionState final {
       inputCtxs_;
   /// Placeholders for tensors generated by DAG nodes that aren't the final
   /// output (i.e. they have children). The set of currently executing nodes.
```
jackm321: Add a comment about lack of ownership of tensors here

nickgg:
found another unused propagate function to remove

lib/Runtime/Executor/ThreadPoolExecutor.h
nickgg: What do you mean here? The Placeholders are owned by `module_`. (maybe just write the comment you like here and i'll copy it in 😄)

nickgg:
change `intermediatePlaceholders_` to a vector since we never look up by key.

nadavrot:
@nickgg Congrats on landing this PR. This is great!

PR #2822, opened 2019-04-30
Add TraceEvents to Habana backend and simplify

nickgg:
*Description*: Add in some more detailed trace events into the Habana function and simplify it's threading logic. Multiple Runner and Waiter threads in the HabanaDeviceManager aren't providing any advantage since it's serialized in the H runtime, so put them both down to 1.

This reduces cold start time of the device quite a bit (see below) due to optimistically activating the topology on model load.

before:
image
after:
image

*Testing*: manual test with a device
*Documentation*:

### DIFF
--- a/lib/Backends/Habana/Habana.h
+++ b/lib/Backends/Habana/Habana.h
@@ -108,7 +108,7 @@ private:
   uint8_t *buffer_;
   /// The number of buffers in the pool.
   unsigned numBuffers_{kDefaultNumBuffers};
-  static constexpr unsigned kDefaultNumBuffers{3};
+  static constexpr unsigned kDefaultNumBuffers{10};
 
   /// Queue of HabanaIOBuffers and a mutex and condition variable for
   /// synchronized access.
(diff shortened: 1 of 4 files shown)


lib/Backends/Habana/Habana.h
```diff
@@ -108,7 +108,7 @@ class HabanaIOBufferPool {
   uint8_t *buffer_;
   /// The number of buffers in the pool.
   unsigned numBuffers_{kDefaultNumBuffers};
-  static constexpr unsigned kDefaultNumBuffers{3};
+  static constexpr unsigned kDefaultNumBuffers{10};
```
SplitInfinity: Why is this being increased? I have some suspicions but it would be nice to include the reason in the commit/PR message.

nickgg: This number determines the size of the pipeline (in batches) allowed on the Habana device: functionally rate limiting calls to synEnqueue. I noticed 3 was too small (was causing the runner threads to block consistently - see first trace above for getIOBuffer). 10 is as arbitrary a number as 3, but higher is better as it will result in less context switches (at the cost of memory).

PR #2824, opened 2019-04-30
[ir-optimizer] Add option -instrument-debug-only to instrument only instructions from the provided list

opti-mix:
Only instructions whose names contain a substring from the list will be debug-instrumented. This is very useful for debugging big NN models, where dumping inputs and outputs of all instructions usually produces huge logs, which are hard to navigate. Limiting the printing to specific instructions significantly reduces the size of the logs and shows only what the user wants to see.

### DIFF
--- a/lib/Optimizer/IROptimizer.cpp
+++ b/lib/Optimizer/IROptimizer.cpp
@@ -40,6 +40,13 @@ static llvm::cl::opt<bool>
     instrumentDebug("instrument-debug",
                     llvm::cl::desc("Instrument the IR for debugging"),
                     llvm::cl::init(false), llvm::cl::Hidden);
+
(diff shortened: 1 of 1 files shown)


lib/Optimizer/IROptimizer.cpp
```diff
@@ -40,6 +40,13 @@ static llvm::cl::opt<bool>
     instrumentDebug("instrument-debug",
                     llvm::cl::desc("Instrument the IR for debugging"),
                     llvm::cl::init(false), llvm::cl::Hidden);
+
+static llvm::cl::list<std::string> instrumentDebugOnly(
```
rdzhabarov: can you specify in the comment how this interacts with the "instrument-debug" option above.
What happens if instrument-debug == false and instrumentDebugOnly has something.

PR #2826, opened 2019-04-30
[DO NOT LAND][DO NOT REVIEW] debug opencl random fail

mortzur:
*Description*: debugging opencl random failure on CI

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -73,6 +73,7 @@ elif [[ "$CIRCLE_JOB" == RELEASE_WITH_EXPENSIVE_TESTS ]]; then
 else
     install_pocl
     CMAKE_ARGS+=("-DCMAKE_BUILD_TYPE=Debug")
+    CMAKE_ARGS+=("-DGLOW_USE_SANITIZER='Address;Undefined'")
     CMAKE_ARGS+=("-DGLOW_WITH_OPENCL=ON")
     if [[ "${CIRCLE_JOB}" == "SHARED" ]]; then
         CMAKE_ARGS+=("-DBUILD_SHARED_LIBS=ON")
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -13,7 +13,7 @@ install_pocl() {
    cd pocl && git checkout 368539f1b34ec84f94edd255961a39925b92066d && cd ../
    mkdir build_pocl
    cd build_pocl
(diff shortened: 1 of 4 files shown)


opti-mix:
@mortzur Thanks for looking into this issue! This bug is really annoying.

mortzur:
The last result of "revert pocl and hack its LLVM.cmake" is disappointing and interesting.
It uses the older version of pocl and only enabling 7.1 (where 7.0 was already supported) and we still see the flaky crash in convTest.
The version that worked for llvm 6.0 does not work for 7.1.

I'm not sure how to proceed. Any advice or idea on this?
@rdzhabarov @nadavrot

rdzhabarov:
One way to fix this is to resurrect https://github.com/pytorch/glow/pull/2833.
But we have to add proper driver installation if that all work I'd need to have another docker image with drivers embedded into the docker image.

I'll try to experiment with that today/tomorrow.

rdzhabarov:
Does that only happen in convTest?

@mortzur can you try to hop on the host and run all tests manually?

mortzur:
@rdzhabarov -
after re-running the Debug test with ssh, CI still report ConvTest failure.
However when I locally run ./tests/ConvTest it passes repeatedly:

`[       OK ] OpenCL/ConvCorrectnessTest.convTest/0 (3159 ms)
[----------] 1 test from OpenCL/ConvCorrectnessTest (3159 ms total)

[==========] 2 tests from 2 test cases ran. (6994 ms total)
[  PASSED  ] 2 tests.`

bertmaher:
Hey, just a quick heads up that I've disabled OpenCL tests on trunk (temporarily) while y'all are debugging this: #2866.  So if you rebase and want to test in CI, make sure to turn it back on :-D

rdzhabarov:
I'm fine disabling convTest on openCL CI if that's the only one which is flaky with your recent changes.

rdzhabarov:
@mortzur Did you have any luck with this change?

mortzur:
@rdzhabarov - The best result I got (all pass for several runs) is for the following case:
* installing and using llvm 7.1 , clang 7.1
* using the "old" version of pocl (meaning before my upgrade to llvm-8 on CI) + command-line hacking pocl's cmake file to accept 7.1 (not only 7.0).
* disabling convTest for OpenCL backend

Should I go ahead and make a new clean PR based on these?

rdzhabarov:
@mortzur 

This is great results!

> disabling convTest for OpenCL backend

Could you please only disable that on CI and not globally for OpenCL (and file an issue about convTest failing on CI)

PR #2830, opened 2019-04-30
[utils] Build clang alongside llvm

bertmaher:
*Description*: I don't know how common this is for people, but when trying to build
Glow on my increasingly-ancient CentOS 7 box I needed to build clang locally
too (I assume for compiling libjit to bitcode).

*Testing*: Used it

### DIFF
--- a/utils/build_llvm.sh
+++ b/utils/build_llvm.sh
@@ -17,8 +17,12 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


compnerd:
@nadavrot - I don't think so; this is a helper, the assumption is that the user already has llvm.

PR #2832, opened 2019-05-01
Merge glow-h changes in internal repo to github master

arunm-git:
*Description*:

This commit brings over changes that were made internally at FB for glow-h.
* Most of the changes are in lib/Backends/Habana/* and are just replaying changes made internally. Hopefully reviewing those should be easy enough for @bertmaher or @rdzhabarov who made most of those changes.
* DeviceManagerTest changes are to parameterize by backend kind
* HabanaBackendTest change adds one test case which is currently disabled.

*Testing*:
ninja all && ninja test
Habana specific testing will be done internally.

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -54,7 +54,7 @@ static synDataType getSynType(ElemKind kind) {
   case ElemKind::FloatTy:
     return syn_type_single;
   case ElemKind::Float16Ty:
-    return syn_type_half;
+    GLOW_UNREACHABLE("Unhandled ElemKind: Float16Ty");
   case ElemKind::Int8QTy:
     return syn_type_fixed;
   case ElemKind::Int16QTy:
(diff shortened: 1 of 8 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -307,6 +286,8 @@ class TensorHandle final {
   /// Get the underlying data buffer.
   void *getData() const { return buffer_; }
 
+  const std::string &getName() const { return name_; }
```
rdzhabarov: we should add doxygen, but probably in fbcode first.

lib/Base/Tensor.cpp
```diff
@@ -148,6 +148,8 @@ static void dumpGenericImpl(Handle<ElemTy> handle, llvm::raw_ostream &os,
   }
 
   os << "]\n";
+
+  os.flush();
```
rdzhabarov: looks legit, but wondering how that sneaked in :)

lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -59,6 +59,11 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
 #else
   constexpr bool withCPU = false;
 #endif
+#ifdef GLOW_WITH_HABANA
```
rdzhabarov: I think changes in this file are good to go, but we need to have a principled way of handling non open sourced backend in the onnxifi glow.

rdzhabarov:
can you check the test failure:
```
../lib/Backends/OpenCL/OpenCL.cpp:168: failed assertion `kernel && "clCreateKernel Failed."'
```

does not look like just a spurious failure.

lib/Base/Tensor.cpp
bertmaher: I approved this internally but I was a bit reluctant to push it to OSS.  I noticed that some interaction with buck's test runner as causing llvm::outs() to not be properly flushed when running via `buck test`.  Adding this flush makes the problem go away but it also seems like a workaround for the quirks of an FB-internal tool.  Anyways, if no one minds, I'm happy to have this land.  Just giving the context :)

lib/Backends/Habana/Habana.cpp
arunm-git: I can add it. This is the source of truth, so good to add it here.

PR #2838, opened 2019-05-01
[habana] Misc fixes for vision nets

bertmaher:
* Introduced LRN, fixed #144
* Added Concat node to the list of supported quantized nodes
* Added RescaleQuantized node, replaced quantized MatMul with GEMM

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -548,10 +548,10 @@ makeSynPoolParams(llvm::ArrayRef<unsigned_t> kernel,
   params->sW = stride[0];
   params->sH = stride[1];
   // Padding
-  params->pWbegin = pad[0];
-  params->pWend = pad[0];
-  params->pHbegin = pad[1];
-  params->pHend = pad[1];
+  params->pHbegin = pad[0];
+  params->pWbegin = pad[1];
+  params->pHend = pad[2];
+  params->pWend = pad[3];
   // Dilation
   params->dilW = 1;
   params->dilH = 1;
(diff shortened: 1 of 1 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -552,10 +552,10 @@ makeSynPoolParams(llvm::ArrayRef<unsigned_t> kernel,
   params->sW = stride[0];
   params->sH = stride[1];
   // Padding
-  params->pWbegin = pad[0];
-  params->pWend = pad[0];
-  params->pHbegin = pad[1];
-  params->pHend = pad[1];
+  params->pHbegin = pad[0];
```
SplitInfinity: Lol, whoops.

lib/Backends/Habana/Habana.cpp
```diff
@@ -1011,6 +1027,18 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
...
       break;
     }
+    case Kinded::Kind::LocalResponseNormalizationNodeKind: {
+      auto *NI = llvm::cast<LocalResponseNormalizationNode>(&I);
+      std::unique_ptr<ns_LrnKernel::Params> params = makeLrnParams(
+          NI->getAlpha(), NI->getBeta(), NI->getK(), NI->getHalfWindowSize());
+
+      chk(synCreateGenericNode(&tensors[NI->getInput()].get(),
+                               &tensors[NI].get(), 1, 1, (void *)params.get(),
+                               "lrn_f32", NI->getName().str().c_str(), nullptr,
```
SplitInfinity: Is the datatype intentionally hardcoded? Does this operator not exist for other datatypes? It makes sense that it wouldn't, but I just want to confirm.

SplitInfinity:
Also, it's interesting that we need LRN because I haven't heard of it being used for months now.

lib/Backends/Habana/Habana.cpp
bertmaher: Good eye - LRN does indeed only exist for f32.  We could assert that here, although it'll already be caught by isOpSupported.

bertmaher:
It's used in inception_v1 and bvlc_alexnet.  Unless it usually gets optimized away somehow?

SplitInfinity:
Ah, those two are 4-8 years old at this point. I thought some recent models use them.

sergeigofman:
LRN indeed aren't used in modern topologies. I introduced it more like an exe4cise and to enable inception_v1. Besides, Habana supports it so why not? Indeed it is only available for fp32.

PR #2844, opened 2019-05-02
[habana] Optimize away some conservative memcpys before concats

bertmaher:
*Description*: We only need to insert intervening memcpy between a concat and its src
if the src is used in multiple places.

*Testing*: unit tests


arunm-git:
D15169826

Issue #2836, opened 2019-05-01
Unable to compile for targets other than x86_64


wyessen:
I have tried emitting bundles for targets other than x86_64 to no avail. Even i386 is failing. It has something to do with tensor types. I suppose different targets expect/support different tensor types; however, I made sure that my network is using Float32. So this shouldn’t be a problem for i386. The network looks like this:

```
graph(%input_data : Float(1, 2)
      %1 : Float(8, 2)
      %2 : Float(8)
      %3 : Float(8, 8)
      %4 : Float(8)
      %5 : Float(8, 8)
      %6 : Float(8)
      %7 : Float(1, 8)
      %8 : Float(1)) {
  %9 : Float(1, 8) = onnx::Gemm[alpha=1, beta=1, transB=1](%input_data, %1, %2), scope: FeedForwardNN/Linear
  %10 : Float(1, 8) = onnx::Relu(%9), scope: FeedForwardNN/ReLU
  %11 : Float(1, 8) = onnx::Gemm[alpha=1, beta=1, transB=1](%10, %3, %4), scope: FeedForwardNN/ReLU
  %12 : Float(1, 8) = onnx::Relu(%11), scope: FeedForwardNN/ReLU
  %13 : Float(1, 8) = onnx::Gemm[alpha=1, beta=1, transB=1](%12, %5, %6), scope: FeedForwardNN/ReLU
  %14 : Float(1, 8) = onnx::Relu(%13), scope: FeedForwardNN/ReLU
  %15 : Float(1, 1) = onnx::Gemm[alpha=1, beta=1, transB=1](%14, %7, %8), scope: FeedForwardNN/ReLU
  %output : Float(1, 1) = onnx::Sigmoid(%15), scope: FeedForwardNN/Sigmoid
  return (%output);
}
```

The traceback that I get when trying to build is:

```
(base) wny@wny-Macbook ~/sandbox/nn/xgmr/build/debug/bin $ ./x-model-runner -model=/Users/wny/sandbox/nn/pytorch_tuts/ffnn.onnx -network-name=x -model-input-name=input_data -cpu -emit-bundle=../output -target i386
Assertion failed: ((i >= FTy->getNumParams() || FTy->getParamType(i) == args[i]->getType()) && "Calling a function with a bad signature: argument type mismatch."), function createCall, file ../../lib/LLVMIRCodeGen/LLVMIRGen.cpp, line 558.
[...]

PR #2847, opened 2019-05-02
[LLVMIRCodeGen] Make LLVMIRCodeGen more cross-compilation friendly

opti-mix:
Determine the size of size_t from the actual libjit bitcode, rather than from the TargetMachine. This seems to be more reliable as libjit kernels are the ones being invoked and the type used by the codegen should exactly match whatever size_t type was used to compile libjit into LLVM bitcode.

Related to #2836


opti-mix:
cc @petecoup Hopefully it would not break anything for you.

petecoup:
Should be fine, and this idea makes sense to me.
Does this make the previous `getTargetSizeTWidth` effectively dead?

opti-mix:
> Does this make the previous getTargetSizeTWidth effectively dead?

Kind of. But I think it could be still useful in some situations. E.g. if libjit was compiled for 32bit and you currently emit code for 64bit, you may want to know the native size_t of the target for different code generation purposes.

So I think we should keep it around.

PR #2848, opened 2019-05-02
Added Gradient for Gather Node (with some restrictions).

putivsky:
*Description*:
Added Gradient for GatherNode.
Indices shouldn't contain duplicate values (restriction).
*Testing*:
Run unit tests.
*Documentation*:

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -31,14 +31,12 @@ using llvm::cast;
 using llvm::isa;
 
 void GraphGradMapper::addGradient(NodeValue activation, NodeValue grad) {
-  if (map_.count(activation)) {
-    auto curr = map_[activation];
+  auto p = map_.insert({activation, grad});
+  if (!p.second) {
+    auto curr = p.first->second;
     auto *sum = F_->createAdd("updateGrad", curr, grad);
-    map_[activation] = sum;
-    return;
+    p.first->second = sum;
   }
-
-  map_[activation] = grad;
 }
 
 bool GraphGradMapper::hasGradient(NodeValue activation) {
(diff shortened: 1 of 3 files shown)


lib/Graph/Grad.cpp
```diff
@@ -275,6 +273,40 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
+      // Get gradient.
+      NodeValue Result = GN->getResult();
+      NodeValue OutputG = map.getGradient(Result);
+      // Get Data & Indices.
+      NodeValue Data = GN->getData();
+      NodeValue Indices = GN->getIndices();
+
+      // Reshape indices into a one-dimensional Tensor (Vector).
+      std::vector<size_t> IndicesDims{Indices.getType()->size()};
+      auto *RI = G->createReshape("reshape.indices.grad", Indices, IndicesDims);
```
SplitInfinity: Ah, so the organization of the indices doesn't matter to the gradient computation. Nice trick!

tests/unittests/GraphGradTest.cpp
```diff
@@ -260,3 +260,30 @@ TEST(GraphAutoGrad, checkBatchedReduceAddGradTest) {
   EE.compile(CompilationMode::Train, TF);
   EE.compile(CompilationMode::Infer, F);
 }
+
+/// Check that we can differentiate functions that use GatherNode.
+TEST(GraphAutoGrad, checkGatherGradTest) {
```
SplitInfinity: Can you add a case here with 2D indices (or 3D, etc.)? You don't have to create a new test, just add another node in the same `glow::Function`.

PR #2851, opened 2019-05-03
[cmake] Factor --gtest_output xml into add_glow_test

bertmaher:
*Description*:
Adding this to every test target is repetitive and error-prone (e.g. I noticed HostManagerTest was writing to ProvisionerTest.xml)

*Testing*: 
`ctest -VV`

### DIFF
--- a/cmake/modules/GlowTestSupport.cmake
+++ b/cmake/modules/GlowTestSupport.cmake
@@ -41,6 +41,7 @@ function(add_glow_test)
 
   list(GET ARG_COMMAND 0 TEST_EXEC)
   list(APPEND ARG_DEPENDS ${TEST_EXEC})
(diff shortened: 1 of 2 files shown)


bertmaher:
cc @tlepley since I think you all use these XML reports.

tlepley-cadence:
@bertmaher thanks for doing this ! Factorization is indeed much better

bertmaher:
Yeah Travis and the expensive_tests are unimpressed with my efforts.  Clearly I'm going to have to level up my CMake skills

bertmaher:
Hey @stalebot leave my masterwork alone.  You just don't understand true art takes time.

bertmaher:
anti-stalebot ping

PR #2852, opened 2019-05-03
Fixing Gradient for BatchedReduceAddNode.

putivsky:
*Description*:
Fixing Gradient for BatchedReduceAdd Node.
*Testing*:
Run unit tests for the generic case.
*Documentation*:

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -266,9 +266,14 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
 
       // Gradient for BatchedReduceAddNode is TileNode,
       // repeating OutputG batch times.
-      auto axis = BRA->getAxis();
-      auto Num = Input.dims()[axis] / OutputG.dims()[axis];
-      auto *TN = new TileNode("tile.grad", Input.getType(), OutputG, Num, axis);
+      auto Axis = BRA->getAxis();
+      // Copy input dimensions first.
+      std::vector<size_t> Dims{Input.dims()};
+      // Then set to 1 dimension size on axis.
+      Dims[Axis] = 1;
+      auto *RSN = G->createReshape("reshape.grad", OutputG, Dims);
+      auto *TN = new TileNode("tile.grad", Input.getType(), RSN->getResult(),
+                              Input.dims()[Axis], Axis);
 
       toAppend.push_back(TN);
       map.addGradient(Input, TN);
--- a/tests/unittests/GradCheckTest.cpp
+++ b/tests/unittests/GradCheckTest.cpp
@@ -234,27 +234,61 @@ TEST_P(InterpreterGrad, gradientCheckMatMul) {
                    0.01);
 }
 
-TEST_P(InterpreterGrad, gradientCheckBatchedReduceAdd) {
+TEST_P(InterpreterGrad, gradientCheckBatchedReduceAddAxis0) {
   PlaceholderBindings Bindings;
-  size_t BatchSize = 40;
-  size_t NumCols = 12;
+  size_t BatchSize = 4;
+  size_t NumRows = 3;
+  size_t NumCols = 5;
 
   auto &Mod = EE_.getModule();
   Function *F = Mod.createFunction("main");
 
-  auto *A = Mod.createPlaceholder(ElemKind::FloatTy, {BatchSize, NumCols}, "A",
-                                  false);
+  auto *A = Mod.createPlaceholder(ElemKind::FloatTy,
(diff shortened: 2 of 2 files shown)


lib/Graph/Grad.cpp
```diff
@@ -267,8 +267,12 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       // Gradient for BatchedReduceAddNode is TileNode,
       // repeating OutputG batch times.
       auto axis = BRA->getAxis();
-      auto Num = Input.dims()[axis] / OutputG.dims()[axis];
-      auto *TN = new TileNode("tile.grad", Input.getType(), OutputG, Num, axis);
+      auto Num = Input.dims()[axis];
+      std::vector<size_t> Dims{1};
+      Dims.insert(Dims.end(), OutputG.dims().begin(), OutputG.dims().end());
```
SplitInfinity: This will only work if `axis = 0`. You need to do something like this:
```
std::vector<size_t> Dims(Input.dims());   // might not work, you may need to use Input.dims().vec()
Dims[axis] = 1;
```
This will create a vector with the dimensions of `Input` but with 1 on the reduced dimension.

tests/unittests/GradCheckTest.cpp
```diff
@@ -236,25 +236,26 @@ TEST_P(InterpreterGrad, gradientCheckMatMul) {
...
+                                  {BatchSize, NumRows, NumCols}, "A", false);
 
-  auto *Exp =
-      Mod.createPlaceholder(ElemKind::FloatTy, {1, NumCols}, "exp", false);
+  auto *Exp = Mod.createPlaceholder(ElemKind::FloatTy, {NumRows, NumCols},
+                                    "exp", false);
 
-  TypeRef Ty = Mod.uniqueType(ElemKind::FloatTy, {1, NumCols});
+  TypeRef Ty = Mod.uniqueType(ElemKind::FloatTy, {NumRows, NumCols});
   Node *BRA = F->createBatchedReduceAdd("BRA", Ty, A, 0 /*axis*/);
```
SplitInfinity: Consider adding a test case with `axis` not equal to `0` to test the case I commented about above.

lib/Graph/Grad.cpp
```diff
@@ -267,8 +267,12 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       // Gradient for BatchedReduceAddNode is TileNode,
       // repeating OutputG batch times.
       auto axis = BRA->getAxis();
```
SplitInfinity: Please consistent between capitalized or non-capitalized variable names; use `Axis` here.

lib/Graph/Grad.cpp
```diff
@@ -267,8 +267,13 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
```
SplitInfinity: `axis` -> `Axis`.

tests/unittests/GradCheckTest.cpp
```diff
@@ -234,27 +234,61 @@ TEST_P(InterpreterGrad, gradientCheckMatMul) {
...
+  PlaceholderBindings Bindings;
+  size_t BatchSize = 4;
+  size_t NumRows = 3;
+  size_t NumCols = 5;
+
+  auto &Mod = EE_.getModule();
+  Function *F = Mod.createFunction("main");
+
+  auto *A = Mod.createPlaceholder(ElemKind::FloatTy,
+                                  {BatchSize, NumRows, NumCols}, "A", false);
```
SplitInfinity: Since you are reduce on axis = 1 here, I think it would be better for readability to make the 1st dimension `BatchSize`: `{NumRows, BatchSize, NumCols}`.

tests/unittests/GradCheckTest.cpp
```diff
@@ -234,27 +234,61 @@ TEST_P(InterpreterGrad, gradientCheckMatMul) {
...
+
+  auto *Exp = Mod.createPlaceholder(ElemKind::FloatTy, {BatchSize, NumCols},
```
SplitInfinity: Similarly, this should become `{NumRows, NumCols}`.

tests/unittests/GradCheckTest.cpp
```diff
@@ -234,27 +234,61 @@ TEST_P(InterpreterGrad, gradientCheckMatMul) {
...
+                                    "exp", false);
+
+  TypeRef Ty = Mod.uniqueType(ElemKind::FloatTy, {BatchSize, NumCols});
```
SplitInfinity: Same here.

PR #2853, opened 2019-05-04
[tracing] fix ScopedTraceBlock reordering bug

nickgg:
*Description*: I had noticed that some of the Scoped TraceEvents had incorrect timestamps, often with zero duration. There were two causes of this:
* In one case I forgot to name the object, which means it's destroyed immediately.
* Since the trace block object is never referenced the compiler was reordering the end timestamp to immediately after the creation of the object (rather than when the scope ended). Adding a memory fence prevents this reordering.
*Testing*: unit tests & a manual test inspecting trace events.
*Documentation*:

### DIFF
--- a/lib/Backends/TraceEvents.cpp
+++ b/lib/Backends/TraceEvents.cpp
@@ -121,6 +121,10 @@ ScopedTraceBlock::ScopedTraceBlock(TraceContext *context, llvm::StringRef name)
   if (context_) {
     context_->logTraceEvent(name_, TraceEvent::BeginType, std::move(args_));
   }
+
+  // A local memory fence to prevent the compiler reordering instructions to
+  // before taking the begin timestamp.
+  std::atomic_signal_fence(std::memory_order_seq_cst);
 }
 
 ScopedTraceBlock::~ScopedTraceBlock() { end(); }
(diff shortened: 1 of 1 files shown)


lib/Backends/TraceEvents.cpp
```diff
@@ -121,6 +121,8 @@ ScopedTraceBlock::ScopedTraceBlock(TraceContext *context, llvm::StringRef name)
   if (context_) {
     context_->logTraceEvent(name_, TraceEvent::BeginType, std::move(args_));
   }
+
```
nadavrot: I think that we need a comment here that explains what's going on and why do we have this fence.

lib/Backends/TraceEvents.cpp
```diff
@@ -132,6 +134,7 @@ ScopedTraceBlock &ScopedTraceBlock::addArg(llvm::StringRef key,
 }
 
 void ScopedTraceBlock::end() {
+  std::atomic_signal_fence(std::memory_order_seq_cst);
```
nadavrot: Also here

rdzhabarov:
> In one case I forgot to name the object, which means it's destroyed immediately.
was it merged in some other PR? (do not see it here currently)

nadavrot:
Perfect!

nickgg:
> was it merged in some other PR? (do not see it here currently)

@rdzhabarov yes, good catch - it was replaced by the TRACE_EVENT_SCOPE macro in #2854.

PR #2856, opened 2019-05-04
[runtime] Reduce multi device startup latency by minimizing lock time

nickgg:
*Description*: The HostManager and ThreadPoolExecutor hold their respective state locks for the full  duration of the work to set up the run (the HostManager also holds it's lock while the Executor is working) which essentially serializes all executor work into a single thread of execution.This diff changes each to hold their lock for the least possible time during `runFunction`: while accessing shared state.

The effect of this is easiest to see on a system with multiple devices:
(these traces are is based on top of #2854 as well for the nicer events, but this PR doesn't depend on it)

**before**:
image

You can see that while the calls to Onnxifi::setIOAndRun are concurrent on different threads, the HostManager and Executor calls are non overlapping. This means that the second device is waiting for the Executor to do the BFS for the first run before as well as its own run.

**after**:
image

Here the HostManager and ThreadPoolExecutor calls are overlapping, which greatly reduces the delay between starting to run on the first and second cards (here ~60us but that may be mostly the difference in timing on the Onnxifi call). 

*Testing*: unit tests in debug, release and with ASAN. Manual runs with multiple devices on H cards.
*Documentation*:
This does change HostManager behaviour slightly: since all API calls used to be serialized to one concurrent access it was always safe to remove a network as there couldn't be any other usages of it. Now that's not the case, so this enforces the requirement that no outstanding runs of a network exist when the user calls `removeNetwork`.

### DIFF
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -118,6 +118,12 @@ void HostManager::removeNetwork(llvm::StringRef networkName) {
   if (networkIterator == networks_.end()) {
     return;
   }
+
+  // TODO: should return an error here rather than asserting:
+  // https://github.com/pytorch/glow/issues/2855
+  assert(networkIterator->second.refcount == 0 &&
+         "Cannot remove a network since there are outstanding runs");
+
   auto &nodes = networkIterator->second.dag.nodes;
   for (auto &node : nodes) {
     for (auto device : node->deviceIDs) {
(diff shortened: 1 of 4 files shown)


nickgg:
fixed the std::bad_function_call crash. Very worried this passed the ASAN test locally and in CI.

nickgg:
@gcatron can you take a look at this one?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -199,10 +216,14 @@ HostManager::runNetwork(llvm::StringRef networkName,
...
   executor_->run(
       networks_[networkName].dag.root.get(), std::move(context), currentRun,
-      [&activeRequest = this->activeRequestCount_, callback,
+      [this, callback,
        name = networkName.str()](RunIdentifierTy runID, llvm::Error err,
                                  std::unique_ptr<ExecutionContext> context) {
-        --activeRequest;
+        --activeRequestCount_;
+        auto it = networks_.find(name);
+        if (it != networks_.end()) {
```
gcatron: Don't you need a lock around this networks_ access?

nickgg: Good catch! Yes, we do.

PR #2857, opened 2019-05-05
[runtime] add TensorPool and speed up adjust batch

nickgg:
*Description*: Add in an object pool for sharing Tensor allocations between runs. Nothing fancy just a lock, a map and a queue. The primary goal is to reduce the latency in the adjust batch stage of Onnxifi. Looks like about a 50% reduction in initial latency.

before:
image

after:
image

(* these numbers make sense but I'm a little suspicious, the copyInputs phase in the DeviceManager is noticeably longer (450us vs 300us - so still a win). Could be location in cache or maybe that we're not zeroing the tensors anymore?)

*Testing*: unit tests with all modes, but Onnxifi has no tests so take it with a grain of salt. New tests for TensorPool. Manual tests on an H-device.
*Documentation*:

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -35,6 +35,7 @@ namespace glow {
 template <class ElemTy> class Handle;
 
 class Tensor;
+class TensorPool;
 
 void genericTranspose(const Tensor *src, Tensor *dest,
                       llvm::ArrayRef<unsigned_t> shuffle);
@@ -64,6 +65,9 @@ private:
 
   /// If the tensor is unowned.
   bool isUnowned_{false};
+
+  /// The TensorPool that is managing this Tensor (if any).
+  TensorPool *tensorPool_{nullptr};
 
   template <class ElemTy> friend class Handle;
 
@@ -157,6 +161,10 @@ public:
   /// \returns the number of bytes required to store the tensor.
   uint64_t getSizeInBytes() const { return type_.getSizeInBytes(); }
 
+  /// \returns the TensorPool managing this object, or nullptr if it is
+  /// unmanaged.
+  TensorPool *getOwningPool() { return tensorPool_; }
+
   /// Initialize an empty tensor.
   Tensor() = default;
 
@@ -201,6 +209,12 @@ public:
          int32_t offset)
       : data_(nullptr), type_(elemTy, dims, scale, offset), isUnowned_{false} {
     reset(type_);
+  }
+
+  /// Allocate a new Tensor managed by the \p tensorPool.
+  explicit Tensor(TypeRef ty, TensorPool *tensorPool)
+      : data_(nullptr), type_(*ty), tensorPool_(tensorPool) {
+    reset(*ty);
   }
 
   Tensor(const Tensor &other) = delete;
@@ -312,6 +326,7 @@ public:
     std::swap(data_, other.data_);
     std::swap(type_, other.type_);
     std::swap(isUnowned_, other.isUnowned_);
+    std::swap(tensorPool_, other.tensorPool_);
   }
 
   /// Move assignment operator.
@@ -319,6 +334,7 @@ public:
     std::swap(data_, other.data_);
     std::swap(type_, other.type_);
     std::swap(isUnowned_, other.isUnowned_);
+    std::swap(tensorPool_, other.tensorPool_);
     return *this;
   }
 
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -390,6 +390,15 @@ struct Type final {
     return true;
   }
 
+  /// \returns a hash value for this Type. Hashes for Ty1 and Ty2 are equal if
+  /// Ty1.isEqual(Ty2).
+  llvm::hash_code equals_hash() const {
+    return llvm::hash_combine(
+        elementType_, dims(),
+        // hashing floats is tricky, fall back to std::hash
+        std::hash<float>{}(scale_), offset_);
+  }
+
   ElemKind getElementType() const { return elementType_; }
 
   /// \returns the shape of the tensor.
--- a/include/glow/Graph/PlaceholderBindings.h
+++ b/include/glow/Graph/PlaceholderBindings.h
@@ -67,6 +67,9 @@ public:
   /// Inserts the Placeholder-Tensor pair.
   void insert(Placeholder *P, Tensor &&T);
 
+  /// Inserts the Placeholder-Tensor pair. This takes ownership of the Tensor.
+  void insert(Placeholder *P, Tensor *T);
+
   /// Allocates a tensor to back the placeholder \p P. The new tensor has the
   /// type of P.
   Tensor *allocate(Placeholder *P);
--- a/include/glow/Support/TensorPool.h
+++ b/include/glow/Support/TensorPool.h
@@ -1,0 +1,95 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_TENSORPOOL_H
+#define GLOW_TENSORPOOL_H
+
+#include "glow/Base/Tensor.h"
+
+#include <atomic>
+#include <iostream>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+
+namespace glow {
+
+class TensorPool final {
+private:
+  struct TypeHash {
+    size_t operator()(const Type &t) const { return t.equals_hash(); }
+  };
+
+  struct TypeEquals {
+    bool operator()(const Type &a, const Type &b) const { return a.isEqual(b); }
+  };
+  /// A stack of available Tensors per Type.
+  std::unordered_map<Type, std::vector<Tensor *>, TypeHash, TypeEquals> pools_;
+
+  /// Mutex around pools_;
+  std::mutex lock_;
+
+  /// Whether or not to allow allocation of new buffers if the pool is empty.
+  const bool preventInlineAllocs_{false};
+
+public:
+  /// Statistics relating to the usage of the pool.
+  struct Stats {
+    /// The total number of Types that has ever been available in this pool.
+    std::atomic<uint64_t> totalTypes{0};
+    /// The number of Tensors currently allocated and available.
+    std::atomic<uint64_t> currentBuffers{0};
+    /// The number of Tensor allocations ever done by the pool.
+    std::atomic<uint64_t> totalAllocs{0};
+    /// The number of Tensor allocations that were done inline to get (as
+    /// opposed to reserve).
+    std::atomic<uint64_t> inlineAllocs{0};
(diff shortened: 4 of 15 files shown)


include/glow/Base/Type.h
```diff
@@ -390,6 +390,30 @@ struct Type final {
     return true;
   }
 
+  /// \returns a hash value for this Type. Hashes for Ty1 and Ty2 are equal iF
+  /// Ty1.isEqual(Ty2).
+  size_t hash() const {
```
nadavrot: You can use llvm::hash. This is what we do in other places in the code.

include/glow/Support/TensorPool.h
```diff
@@ -0,0 +1,83 @@
...
+  TensorPool(bool preventAllocs = false)
+      : preventInlineAllocs_{preventAllocs} {}
+
+  ~TensorPool() { clear(); }
+
+  /// Retrieve a Tensor with type \p ty from the pool - this type must have
+  /// previously been added by initialize. If the pool is empty this will
+  /// allocate a new Tensor unless preventAllocs was provided at construction
+  /// time.
+  Tensor get(TypeRef ty);
```
nadavrot: This should probably be "Tensor*". Which is a pointer to some tensor that is managed by the pool.

include/glow/Base/Tensor.h
```diff
@@ -157,6 +161,10 @@ class Tensor final {
   /// \returns the number of bytes required to store the tensor.
   uint64_t getSizeInBytes() const { return type_.getSizeInBytes(); }
 
+  /// \returns the TensorPool managing this object, or nullptr if it is
```
nadavrot: Why is this a part of Tensor?  I assumed that the tensor pool would manage some tensors. Just like the Placeholder binding is managing some other resources. The fact that the tensor is managed by some pool should not matter to the tensor, just like a car that is managed by valet should not have special features to support the valet - just a key.

nickgg: The PlaceholderBindings owns its Tensors - and will delete them on `clear()`, `erase()` and the destructor. This seemed the least disruptive option but there are some alternatives we could consider:
* Change Placeholder so it takes a `Tensor *` rather than `Tensor &&` doesn't own its placeholders and calling code must go through pairs() and delete or `reclaim()` as appropriate. Adds complexity to all usage of PlaceholderBindings and we're likely to forget.
* Change the TensorPool to return a new Tensor, which is deleted but reclaims the buffer in its destructor. We'd need to subclass Tensor and have the TensorPool return a new ManagedTensor* backed by a shared buffer, and have to change PlaceholderBindings to take a Tensor* rather than a Tensor&&. This will be slightly less efficient than what we have now, primarily mainly because we're no longer reusing the actual Tensor objects (just the backing memory) and we'll be hitting malloc regularly anyway. We'd also be putting all access to Tensor through a vtable.
* Change Tensor replacing isUnowned with a refcount over the backing buffer and have the TensorPool always hold a refount. Bit of juggling here to make sure that the TensorPool only returns buffers with a single refcount since it wont get a clear signal when the buffer isn't used anymore. A bit more complicated dealing with cases where the Tensor really doesn't own its buffer (such as the input buffers in Onnxifi).
* Change PlaceholderBindings so that it has another map of managed tensors (and which pool manages them). Basically the same as this except the additional code is in PlaceholderBindings instead of Tensor. The PlaceholderBindings would need to look up multiple maps to do anything though, potentially.

I prefer the small change to Tensor over any of these, but willing to be convinced.

include/glow/Support/TensorPool.h
nickgg: PlaceholderBindings takes Tensors by rvalue ref and takes ownership, so this doesn't work quite how you'd like, I think.

include/glow/Base/Tensor.h
nadavrot: Okay.

include/glow/Base/Type.h
nickgg: Unless I'm misreading we define the hash value of Tensor to be literally its size right now though...

https://github.com/pytorch/glow/blob/master/lib/Graph/Nodes.cpp#L1325

nickgg: @nadavrot I might be missing what you meant here, what do you suggest?

nickgg:
fix missing `<atomic>` header

nickgg:
Changed to store a pointer rather than Tensor object.

include/glow/Base/Type.h
nadavrot: LLVM has utilities for computing the hash value of things. It can merge hashes for you as well. This means that you don't need to implement the shifts and xors yourself. Just call into LLVM's hash method and it will digest things for you.

nadavrot:
Grep the code for 'llvm::hash'. There are a few places where we use LLVM's hash to combine things and digest names and values. 

Example:
llvm::hash_code Placeholder::getHash() const { return llvm::hash_combine(getName());

nickgg:
switch to llvm::hash

include/glow/Support/TensorPool.h
```diff
@@ -0,0 +1,85 @@
...
+private:
+  struct TypeHash {
+    size_t operator()(const Type &t) const { return t.equals_hash(); }
+  };
+
+  struct TypeEquals {
+    bool operator()(const Type &a, const Type &b) const { return a.isEqual(b); }
+  };
+  /// A stack of available Tensors per Type.
+  std::unordered_map<Type, std::vector<Tensor *>, TypeHash, TypeEquals> pools_;
```
SplitInfinity: I noticed that you did not use the random-access property of the `std::vector`. Could it be a `std::queue`? Or `std::stack`?

include/glow/Support/TensorPool.h
```diff
@@ -0,0 +1,85 @@
...
+
+  /// Mutex around pools_;
+  std::mutex lock_;
+
+  /// Whether or not to allow allocation of new buffers if the pool is empty.
+  const bool preventInlineAllocs_{false};
```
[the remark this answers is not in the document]
nickgg: Because the alloc happens inline to the `get()`function call as opposed to `initialize()`? Happy to change it what would you prefer?

lib/Support/TensorPool.cpp
```diff
@@ -0,0 +1,97 @@
...
+void TensorPool::reclaim(Tensor *t) {
+  std::lock_guard<std::mutex> l(lock_);
+  auto it = pools_.find(t->getType());
+  assert(it != pools_.end() && "Type has not been initialized");
+  stats_.totalReclaims++;
+  stats_.currentBuffers++;
+  it->second.push_back(t);
+}
+
+void TensorPool::initialize(TypeRef ty, size_t count) {
```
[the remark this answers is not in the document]
nickgg: I like `reserve`, will rename

tests/unittests/TensorPoolTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+/// Can get Tensor from the pool without allocation.
+TEST(TensorPool, BasicTest) {
+  TensorPool pool;
+  Type ty(ElemKind::FloatTy, {1, 2, 3});
+  pool.initialize(&ty, 1);
+
+  Tensor *T = pool.get(&ty);
+  EXPECT_TRUE(T->getType().isEqual(ty));
+  EXPECT_EQ(T->dims(), ty.dims());
```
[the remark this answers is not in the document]
nickgg: Yes, just being obvious about it :)

nickgg:
@SplitInfinity Yes we're relying on destruction order for safety. The alternative is to make the tensorPool_ field a shared_ptr, but given previous reactions to smart pointers I figured I'd save some time and stay with a raw pointer. 

In the Glow Runtime that should be fine because the PlaceholderBindings are scoped to a run and the TensorPools are scoped to process end.

include/glow/Base/Type.h
```diff
@@ -220,7 +220,6 @@ struct ShapeHWD {
     width = shape[1];
     depth = shape[2];
   }
-
```
bertmaher: Nit, revert the loss of vertical whitespace :-)
(It's like an old SAT analogy, Bert:Vertical whitespace::Nadav:Periods at end of sentences)

nickgg: Huh thats weird I didn't do that on purpose

include/glow/Graph/PlaceholderBindings.h
```diff
@@ -67,6 +67,9 @@ class PlaceholderBindings final {
   /// Inserts the Placeholder-Tensor pair.
   void insert(Placeholder *P, Tensor &&T);
 
+  /// Inserts the Placeholder-Tensor pair. This takes ownership of the Tensor.
```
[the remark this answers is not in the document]
nickgg: I don't understand what you mean here. Both versions of insert take ownership and return a pointer to the TensorPool. The difference here is that when you pass in a `Tensor *` it doesn't create a new tensor object.

lib/Support/TensorPool.cpp
```diff
@@ -0,0 +1,97 @@
...
+  std::vector<Tensor *> temp;
+  for (unsigned i = 0; i < count; ++i) {
+    stats_.totalAllocs++;
+    temp.push_back(new Tensor(ty, this));
```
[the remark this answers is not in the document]
nickgg: Yes, but really if we run out of memory to the point that `new` is failing we probably just want to crash anyway.

nickgg:
manual rebase

include/glow/Graph/PlaceholderBindings.h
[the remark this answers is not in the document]
SplitInfinity: Okay, fair. It just threw me off that the comment for the `Tensor&&` version doesn't say that it takes ownership.

nickgg:
I think I've addressed all outstanding comments, can I get a re-review up in here?

SplitInfinity:
@nickgg Looks like there are merge conflicts?

include/glow/Support/TensorPool.h
```diff
@@ -0,0 +1,85 @@
...
+
+public:
+  struct Stats {
```
SplitInfinity: Consider adding short one line comments for these statistics.

lib/Support/TensorPool.cpp
```diff
@@ -0,0 +1,99 @@
...
+      stats_.totalTypes++;
+    }
+
+    std::vector<Tensor *> &queue = pools_[*ty];
+    std::move(temp.begin(), temp.end(), std::back_inserter(queue));
+    stats_.currentBuffers += count;
+  }
+}
+
+void TensorPool::clear() {
```
SplitInfinity: Should `clear()` also clear the types within the pool? I think the easiest answer is no, because doing so would break `reclaim`. But I could see `clear()` being called and some types never being used after that. It's not enough memory that I'd lose sleep over it but I'd like to get your thoughts on this.

lib/Support/TensorPool.cpp
```diff
@@ -0,0 +1,99 @@
...
+  std::lock_guard<std::mutex> l(lock_);
+  for (auto &p : pools_) {
+    stats_.currentBuffers -= p.second.size();
+    for (auto *t : p.second) {
+      delete t;
```
SplitInfinity: I wonder if it would be useful to keep track of how much buffers were freed here so that we can get an idea of how much `clear` actually clears.

lib/Support/TensorPool.cpp
nickgg: Yeah I'm not sure, could see it going either way. Generally i guess we could provide an option for clearing or not but don't want to add unnecessary complexity.

lib/Support/TensorPool.cpp
nickgg: @SplitInfinity Via a new statistic or something?

lib/Support/TensorPool.cpp
nickgg: totalTypes statistic becomes more complicated if we remove a type, but that's not a strong reason either way

PR #2861, opened 2019-05-06
[glow] Add command line flags for # devices and Habana memory capacity

bertmaher:
*Description*: We need a way to set these externally.
*Testing*: Used them.
*Documentation*: Help strings?

### DIFF
--- a/lib/Backends/Habana/HabanaDeviceManager.cpp
+++ b/lib/Backends/Habana/HabanaDeviceManager.cpp
@@ -16,6 +16,7 @@
 
 #include "HabanaDeviceManager.h"
 
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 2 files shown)


lib/Backends/Habana/HabanaDeviceManager.cpp
```diff
@@ -26,12 +27,20 @@
 using namespace glow;
 using namespace glow::runtime;
 
+namespace glow {
+namespace runtime {
+
+uint64_t GlowHabanaMemory = std::numeric_limits<uint64_t>::max();
```
rdzhabarov: would it make sense to have more conservative value (7516192768; // 7GB)?

bertmaher: Probably for now.  I was thinking that once the actual get-mem-info API lands, we could use min(command line param, actual reported device memory)

PR #2865, opened 2019-05-07
Eliminate duplicate allocation for weights in model loaders

jackm321:
*Description*:
Instead of copying data from tensor that have already been loaded into constants, move the tensor memory instead. This should cut the amount of allocations for weights in the loaders roughly in half.

When loading weights in model loaders, load them directly into `Contant`s instead of into intermediary Tensors to be collected up later.

*Testing*:
CI
Ran on predictor service and dumped before and after heap profiles (see diff)

*Documentation*:
doxygen

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -227,8 +227,7 @@ protected:
   llvm::Error loadRelu(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
     auto *R = G_.createRELU(opName, in);
     RETURN_IF_ERR(addNodeAsOutput(op, R));
     return llvm::Error::success();
@@ -241,12 +240,10 @@ protected:
     const std::string &opName = loadOperatorName(op);
 
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
 
     NodeValue slope;
-    ASSIGN_VALUE_OR_RETURN_ERR(slope,
-                               getNodeValueOrCreateConstantByName(op.input(1)));
+    ASSIGN_VALUE_OR_RETURN_ERR(slope, getNodeValueByName(op.input(1)));
 
     // Do broadcasting.
     auto targetDim = in.dims();
@@ -262,8 +259,7 @@ protected:
   llvm::Error loadSigmoid(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
     auto *S = G_.createSigmoid(opName, in);
     RETURN_IF_ERR(addNodeAsOutput(op, S));
     return llvm::Error::success();
@@ -272,8 +268,7 @@ protected:
   llvm::Error loadTanh(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
     auto *T = G_.createTanh(opName, in);
     RETURN_IF_ERR(addNodeAsOutput(op, T));
     return llvm::Error::success();
@@ -282,8 +277,7 @@ protected:
   llvm::Error loadShape(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
 
     // This is statically known data, and so we create a Tensor for it.
     Tensor T(ElemKind::Int64ITy, {in.dims().size()});
@@ -302,8 +296,7 @@ protected:
   llvm::Error loadSqrt(const OpType &op, ArgumentDictionaryTy &dict) {
     const std::string &opName = loadOperatorName(op);
     NodeValue in;
-    ASSIGN_VALUE_OR_RETURN_ERR(in,
-                               getNodeValueOrCreateConstantByName(op.input(0)));
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
     auto *R = G_.createPow(opName, in, 0.5f);
     RETURN_IF_ERR(addNodeAsOutput(op, R));
     return llvm::Error::success();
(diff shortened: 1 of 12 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -276,14 +276,12 @@ class CommonOperatorLoader : public ProtobufLoader {
...
-    auto *T = new Tensor(ElemKind::Int64ITy, {in.dims().size()});
-    tensors_[opName].reset(T);
-    T->template getHandle<int64_t>() =
+    // This is statically known data, and so we create a Tensor for it.
+    Tensor T(ElemKind::Int64ITy, {in.dims().size()});
+    T.getHandle<int64_t>() =
         std::vector<int64_t>(in.dims().begin(), in.dims().end());
 
-    if (auto resultOrErr = createAndRegisterConstant(opName, *T)) {
+    if (auto resultOrErr = createAndRegisterConstant(opName, std::move(T))) {
```
putivsky: std::move(*T) ???

lib/Importer/ProtobufLoader.cpp
```diff
@@ -65,13 +81,13 @@ ProtobufLoader::getNodeValueByName(llvm::StringRef name) const {
...
 ProtobufLoader::createAndRegisterConstant(llvm::StringRef name,
-                                          const Tensor &tensor) {
+                                          Tensor &&tensor) {
   RETURN_ERR_IF_NOT(
       !hasNodeByName(name),
       llvm::Twine("Creating an already existing node ", name).str());
   // Note: We do not support training from models loaded from protos, so
   // trainable is always set to false here.
-  Constant *node = G_.getParent()->createConstant(name, tensor);
+  Constant *node = G_.getParent()->createConstant(name, std::move(tensor));
```
putivsky: More generic would be to forward tensor:
Constant *node = G_.getParent()->createConstant(name, std::forward<Tensor>(tensor));

lib/Graph/Graph.cpp
```diff
@@ -407,6 +407,10 @@ Constant *Module::createConstant(llvm::StringRef name, const Tensor &tensor) {
   return V;
 }
 
+Constant *Module::createConstant(llvm::StringRef name, Tensor &&tensor) {
+  return addConstant(new Constant(name, std::move(tensor)));
```
putivsky: consider forward vs move:
  return addConstant(new Constant(name, std::forward<Tensor>(tensor)));

lib/Importer/ProtobufLoader.cpp
```diff
@@ -34,6 +34,22 @@ llvm::Expected<Tensor *> ProtobufLoader::getTensorByName(llvm::StringRef name) {
...
+llvm::Expected<std::unique_ptr<Tensor>>
+ProtobufLoader::takeTensorByName(llvm::StringRef name) {
+  auto it = tensors_.find(name);
+  RETURN_ERR_IF_NOT(
+      it != tensors_.end(),
+      llvm::Twine("There is no tensor registered with name ", name).str());
+
+  std::unique_ptr<Tensor> t = std::move(it->second);
+
+  RETURN_ERR_IF_NOT(
```
rdzhabarov: getTensorByName now will return nullptr if we take it here first (probably not expected), maybe add similar condition to the getTensorByName.

lib/Graph/Graph.cpp
nadavrot: @putivsky Thank you for pointing this out. I am not sure that I understand the difference between the two. I was not able to find a good answer on google. Can you explain why you prefer std::forward over std::move?

bertmaher:
Nice memory win 🕶!  Something I'm wondering, will every Tensor created by `loadWeight` eventually be consumed by `createAndRegisterConstant`?  Should we consider making `loadWeight` directly create Constants and add them to the nodeValueByName_ map rather than passing through tensors_?

include/glow/Importer/CommonOperatorLoader.h
rdzhabarov: looks legit since now we use `Tensor T` instead of a pointer.

lib/Graph/Graph.cpp
SplitInfinity: I don't think `std::forward` is applicable here because `Tensor&& tensor` is not a universal reference. It's not possible for an lvalue reference to bind to it. For example: https://godbolt.org/z/yQwHhR

jackm321: @SplitInfinity nice explanation

jackm321:
> Should we consider making loadWeight directly create Constants and add them to the nodeValueByName_ map rather than passing through tensors_?

Yeah that's an interesting idea, let me look into doing that instead, I think it should work

lib/Importer/ProtobufLoader.cpp
SplitInfinity: I don't think `std::forward` is applicable here because `Tensor &&tensor` can only bind to rvalue references.

lib/Graph/Graph.cpp
SplitInfinity: In case people are curious, if this function was templated like so:
```
template <typename T>
Constant *Module::createConstant(llvm::StringRef name, T &&tensor) { ... }
```
then `std::forward` would be applicable here. I think you can still use it, but there's no point it will do the same thing that `std::move` does every time.

jackm321:
It appears that the TextTranslator tests fail because in the `en2gr` model the init_net is loading weights that have the same names as input placeholders ("encoder_inputs", "attn_weights" etc). Previously these weights were simply being thrown in `tensors_` and ignored by the rest of the model loading logic but this PR creates a Constant node directly from all weights so this leads to two NodeValues with the same name (one Constant and one Placeholder). @jfix71 do you know why this model does this?

jfix71:
@jackm321 I think this is an artifact of the export script that I used. The script adds "placeholder" (not our `Placeholder`) `GivenTensorFills` (`Constants`) for the external input blobs. I'm not sure the purpose of the placeholder `GivenTensorFills`. Then when you load in external input blobs in C2 I guess they just take precedence over any weights named the same.

BTW, the script that generated `en2gr` was very much in its early stages and I had to manually hack the protos to get them to successfully load into Glow (e.g. manually do constant propagation in the proto itself to make sure all shapes were statically known; hand deduplicate many `GivenTensorFills` so the proto wasn't insanely large; etc.). It might be worth finding out if "placeholder" thing is something C2 shouldn't actually support/something we can safely assume would never occur normally, and then we could manually update `en2gr` to no longer do this.

jackm321:
@bertmaher I followed your suggestion and now `tensors_` is gone. It kind of blew up the code size but I tried to make some distinct commits at least but I wish I could have stacked this thing.

lib/Graph/Graph.cpp
```diff
@@ -438,6 +442,7 @@ llvm::StringRef Module::uniqueName(llvm::StringRef name,
 
 Constant *Module::addConstant(Constant *V) {
   V->setName(uniqueName(V->getName(), uniqueVariableNames_));
+  V->setType(Constant::ResultIndices::OutputIdx, uniqueType(*V->getType()));
```
rdzhabarov: can you add a comment why we need to do this

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1104,7 +1064,7 @@ Caffe2ModelLoader::loadInputsWithTensorProtoType(const caffe2::NetDef &net,
   } else {
     std::unique_ptr<Tensor> T(new Tensor());
```
rdzhabarov: nit: since we move tensor anyway there is no need to allocate `T` on a heap.

lib/Importer/ProtobufLoader.cpp
```diff
@@ -63,17 +77,20 @@ ProtobufLoader::getNodeValueByName(llvm::StringRef name) const {
...
-                                          const Tensor &tensor) {
-  RETURN_ERR_IF_NOT(
-      !hasNodeByName(name),
-      llvm::Twine("Creating an already existing node ", name).str());
+llvm::Error ProtobufLoader::createAndRegisterConstant(llvm::StringRef name,
+                                                      Tensor &&tensor) {
+  auto it = nodeValueByName_.find(name);
+  if (it != nodeValueByName_.end()) {
+    if (llvm::dyn_cast<Placeholder>(it->second.getNode())) {
+      // Placeholders take precedents over Constants.
```
rdzhabarov: precedence?

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -161,8 +161,6 @@ TEST(caffe2, convNHWC) {
 
   // We have 2 placeholders:  1 input and 1 output.
   EXPECT_EQ(mod.getPlaceholders().size(), 2);
-  // We have 2 constants: Weights and bias.
-  EXPECT_EQ(mod.getConstants().size(), 2);
```
rdzhabarov: why this is removed?

PR #2866, opened 2019-05-07
Disable OpenCL tests temporarily due to sporadic double free in POCL

bertmaher:
*Description*: There's work ongoing to figure this out, but in the meantime we've sorta lost all signal from the Debug build config.  Let's just turn these off until it can get sorted.

*Testing*: This PR


rdzhabarov:
it's reasonable until we fix this.

PR #2871, opened 2019-05-08
Add Tensor release and Module strip.

gcatron:
*Description*: This fixes an issue with OpenCL in the runtime. The module was being cleared but OpenCL needs tensor type information. This adds a new release method to the tensor that frees the data and converts the tensor to an unowned tensor. This also introduces a strip method to the module that releases constants memory and frees functions.

*Testing*: ninja test, and resnet-runtime with openCL
*Documentation*:

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -176,9 +176,13 @@ public:
   /// Erase all of the functions from the module.
   void eraseFunctions();
 
-  /// Erase all the functions, variables, etc. If \p clearPlaceholders is false
-  /// then placeholders in the module will not be cleared out.
-  void clear(bool clearPlaceholders = true);
+  /// Erase all the functions, variables, etc.
+  void clear();
+
+  /// Strips payloads from constants and deletes functions. This is useful when
+  /// the Module will be kept around for metadata but we want to reduce memory
+  /// use. Unlike clear this leaves PHs and Constants in the module.
+  void strip();
 
   /// Erase a function \p F from the module.
   void eraseFunction(Function *F);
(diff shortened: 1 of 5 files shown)


include/glow/Graph/Graph.h
```diff
@@ -176,9 +176,11 @@ class Module final {
   /// Erase all of the functions from the module.
   void eraseFunctions();
 
-  /// Erase all the functions, variables, etc. If \p clearPlaceholders is false
-  /// then placeholders in the module will not be cleared out.
-  void clear(bool clearPlaceholders = true);
+  /// Erase all the functions, variables, etc.
+  void clear();
+
+  /// Strips payloads from constants and deletes functions.
```
opti-mix: Can you add more explanations here, e.g. that it keeps the metadata around and why it does it? And may be explain for `clear` that it removes everything and how both are different.

include/glow/Graph/Graph.h
```diff
@@ -176,9 +176,13 @@ class Module final {
...
   void eraseFunctions();
 
-  /// Erase all the functions, variables, etc. If \p clearPlaceholders is false
-  /// then placeholders in the module will not be cleared out.
-  void clear(bool clearPlaceholders = true);
+  /// Erase all the functions, variables, etc.
+  void clear();
+
+  /// Strips payloads from constants and deletes functions. This is useful when
+  /// the Module will be kept around for metaData but we want to reduce memory
```
opti-mix: `metaData` ?

gcatron: fixed

PR #2872, opened 2019-05-08
Added a command line flag to toggle activations memory allocation reuse

### LINKED ISSUE #2870: Add options to control how memory allocation is done in runtime::RuntimeBundle::create for IRFunctions
In particular it could be interesting to e.g. instruction the `activationsAllocator` to not reuse the memory, so that any specific memory location is ever used only by a single tensor. This would make debugging of memory allocation issues significantly easier.

gcatron:
*Description*: This PR adds a command line flag ( reuse-activation-memory-allocations) which allows toggling reuse of activation allocations. This is useful for debugging.
*Testing*: ninja test, ran with resnet-runtime. 
*Documentation*: NA
fixes #2870

### DIFF
--- a/lib/Backends/BackendUtils.cpp
+++ b/lib/Backends/BackendUtils.cpp
@@ -16,11 +16,20 @@
 #include "glow/Backends/BackendUtils.h"
 #include "glow/IR/Instrs.h"
 
+#include "llvm/Support/CommandLine.h"
+
 using namespace glow;
 
 using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
+
+static llvm::cl::OptionCategory BackendUtilsCat("Glow Backend Utils Options");
+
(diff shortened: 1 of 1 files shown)


lib/Backends/BackendUtils.cpp
```diff
@@ -16,12 +16,21 @@
...
+
 using namespace glow;
 
 using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
+static llvm::cl::OptionCategory BackendUtilsCat("Glow Backend Utils Options");
+
+llvm::cl::opt<bool> reuseActivations(
```
opti-mix: Should it be static as well?

opti-mix: And let's call it `reuseActivationsMemory`

PR #2874, opened 2019-05-08
Removed deviceConfig specialization. Added string map for parameters.

gcatron:
*Description*: An issue brought up in #2799 is that we want the Loader and any other caller to not require device specific code when calling into the runtime. This change adds a string map for holding parameters for Device Managers. The idea being these can be strings passed in from the command line or config files.
*Testing*: Ninja test, resnet-runtime with opencl
*Documentation*: NA

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -120,23 +120,25 @@ using DAGListTy = std::vector<DAG>;
 /// device can extend this class to contain information to identify
 /// and configure the device manager. Additionally it needs to set it's kind_
 /// member variable to it's correct BackendKind.
-class DeviceConfig {
+struct DeviceConfig {
   /// An enum indicating what kind of backend this config is for. It is used in
   /// checking the type of config before casting to a derived class.
-  const BackendKind backendKind_;
+  const BackendKind backendKind;
   /// A human readable name to identify the device.
-  std::string name_;
+  std::string name;
 
-public:
-  DeviceConfig(BackendKind kind) : backendKind_(kind) {}
+  /// A map of configuration parameters.
+  llvm::StringMap<std::string> parameters{};
+
+  DeviceConfig(BackendKind kind) : backendKind(kind) {}
   DeviceConfig(BackendKind kind, std::string name)
-      : backendKind_(kind), name_(name) {}
+      : backendKind(kind), name(name) {}
 
-  BackendKind getBackendKind() { return backendKind_; }
+  DeviceConfig(BackendKind kind, std::string name,
+               llvm::StringMap<std::string> parameters)
+      : backendKind(kind), name(name), parameters(parameters) {}
 
-  llvm::StringRef getName() const { return name_; }
-  bool hasName() const { return name_ != ""; }
-  void setName(llvm::StringRef name) { name_ = name; }
+  bool hasName() const { return name != ""; }
 };
 
 } // namespace runtime
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -50,6 +50,18 @@ DeviceManager *createOCLDeviceManager(std::unique_ptr<DeviceConfig> config) {
 } // namespace runtime
 } // namespace glow
 
+/// Helper method to parse a string parameter to an unsigned. \returns
+/// llvm::Expected with either the value or an error.
+static llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
+  char *end;
+  auto parsed = strtol(input.c_str(), &end, 10);
+  if (end == input.c_str() || *end != '\0') {
+    return MAKE_ERR(GlowErr::ErrorCode::RUNTIME_ERROR,
+                    "Invalid input expected integer got: " + input);
+  }
+  return parsed;
+}
+
 cl_mem OpenCLDeviceManager::allocDeviceBuffer(uint64_t size) {
   const uint64_t alignment = 128;
   // Always allocate buffers properly aligned to hold values of any type.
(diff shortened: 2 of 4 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -64,15 +64,23 @@ OpenCLDeviceManager::OpenCLDeviceManager(std::unique_ptr<DeviceConfig> config)
...
-    OpenCLDeviceConfig *config =
-        static_cast<OpenCLDeviceConfig *>(config_.get());
-    clDeviceId = config->deviceId;
-    clPlatformId = config->platformId;
-    clDoProfile = config->doProfile;
+  // Check if parameters are in map.
+  if (config_) {
+    auto it = config_->parameters.find("deviceId");
+    if (it != config_->parameters.end()) {
+      clDeviceId = std::stoi(it->second);
```
jfix71: What happens here if the input string is formatted incorrectly, and there's no int? Should we return an error?

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -64,15 +64,23 @@ OpenCLDeviceManager::OpenCLDeviceManager(std::unique_ptr<DeviceConfig> config)
...
+    }
+    it = config_->parameters.find("platformId");
+    if (it != config_->parameters.end()) {
+      clPlatformId = std::stoi(it->second);
+    }
+    it = config_->parameters.find("doProfile");
+    if (it != config_->parameters.end()) {
+      clDoProfile = it->second == "1";
```
jfix71: Maybe this should check if it's true/false and return an error otherwise?

gcatron:
> Nice! This should also go alongside a new command line param that takes these `string->string` configs, right? Shouldn't that be replacing OpenCL command line options (`clPlatformId`/`clDeviceId`/`clDoProfile`) from OpenCL.cpp now?

Yup that's the plan, there's a bit more work around configs for multiple devices but once that is done we can get rid of the OpenCL specific parameters.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -64,15 +64,23 @@ OpenCLDeviceManager::OpenCLDeviceManager(std::unique_ptr<DeviceConfig> config)
...
 
-  if (config_ && config_->getBackendKind() == BackendKind::OpenCL) {
-    OpenCLDeviceConfig *config =
-        static_cast<OpenCLDeviceConfig *>(config_.get());
-    clDeviceId = config->deviceId;
-    clPlatformId = config->platformId;
-    clDoProfile = config->doProfile;
+  // Check if parameters are in map.
+  if (config_) {
+    auto it = config_->parameters.find("deviceId");
```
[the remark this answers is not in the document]
gcatron: They are already declared as command line parameters, but once we remove the OpenCL parameters we'll probably want to do something like that.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -120,23 +120,25 @@ using DAGListTy = std::vector<DAG>;
...
 
-  BackendKind getBackendKind() { return backendKind_; }
+  DeviceConfig(BackendKind kind, std::string name,
+               llvm::StringMap<std::string> parameters)
+      : backendKind(kind), name(name), parameters(parameters) {}
 
-  llvm::StringRef getName() const { return name_; }
-  bool hasName() const { return name_ != ""; }
-  void setName(llvm::StringRef name) { name_ = name; }
+  bool hasName() const { return name != ""; }
```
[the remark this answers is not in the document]
gcatron: That's git saying I added a line because I moved it. So I'd prefer to leave it be.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -62,17 +62,52 @@ cl_mem OpenCLDeviceManager::allocDeviceBuffer(uint64_t size) {
 OpenCLDeviceManager::OpenCLDeviceManager(std::unique_ptr<DeviceConfig> config)
     : QueueBackedDeviceManager(BackendKind::OpenCL, std::move(config)) {}
 
+llvm::Error OpenCLDeviceManager::parseConfig() {
+  auto it = config_->parameters.find("deviceId");
+  if (it != config_->parameters.end()) {
+    char *end;
+    auto input = it->second.c_str();
+    auto parsed = strtol(input, &end, 10);
+    if (end == input) {
```
jfix71: This seems like an incorrect error condition check. E.g. you could pass in "1a2" and it wouldn't catch the error, right? I think you want to check that end got to the end of input, not that it's stuck at the beginning of the input.

jfix71: Also this seems like a good place to use a helper, since it's also used below for `platformId` and probably will be used across other DeviceManagers as well?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -120,23 +120,25 @@ using DAGListTy = std::vector<DAG>;
...
   /// An enum indicating what kind of backend this config is for. It is used in
   /// checking the type of config before casting to a derived class.
-  const BackendKind backendKind_;
+  const BackendKind backendKind;
   /// A human readable name to identify the device.
-  std::string name_;
+  std::string name;
+
+  /// A map of configuration parameters.
+  llvm::StringMap<std::string> parameters{};
```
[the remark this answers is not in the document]
gcatron: Probably not immediately, that would mean we need to make guarantees about the lifetime of that map.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -50,6 +50,17 @@ DeviceManager *createOCLDeviceManager(std::unique_ptr<DeviceConfig> config) {
 } // namespace runtime
 } // namespace glow
 
+llvm::Error parseInputAsUnsigned(std::string input, unsigned &output) {
```
jfix71: nit: I think this would be cleaner as an `llvm::Expected<unsigned>`, and then use `ASSIGN_VALUE_OR_RETURN_ERR()` at the callsite, instead of passing the reference to the output.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -50,6 +50,16 @@ DeviceManager *createOCLDeviceManager(std::unique_ptr<DeviceConfig> config) {
 } // namespace runtime
 } // namespace glow
 
+llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
```
jfix71: nit: `static`, and doxygen.

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -97,6 +97,8 @@ class OpenCLDeviceManager : public QueueBackedDeviceManager {
 
   llvm::Error init() override;
 
+  /// Parse config object provided at initialization.
```
jfix71: nit: Add `\returns` for doxygen.

PR #2877, opened 2019-05-09
Refactor Backends libraries

bertmaher:
*Description*: lib/Backends is a mess.  It creates four separate libraries, when the goal is to have one library per lib/ subdirectory.  Let's see if we can get back to that ideal state:
- Make "Backend" its own lib/ subdirectory.  The naming is suboptimal here, but that can be fixed too.
- Make "ExecutionContext" its own lib/ subdirectory.  I don't really know if that's a great idea, but if it merited its own cmake library...
- Roll DeviceManagers into Backends.  I just can't see any real reason to have a DeviceManager that can't actually run anything on those Devices.  I had to break a circular dependence with CPUDeviceManager, but that was actually trivial.  (Although we'll see what pops out on CI)

*Testing*: ninja all with and without shared libs


bertmaher:
You have an interesting point about AOT.  I guess I don't really understand how we want to factor AOT into our design.  Tbh it doesn't feel like it's well supported right now, but maybe it should be?

Maybe I'm moving the design in the wrong direction by combining these concepts (albeit only at the build level, but still).  To me it looks like running code happens via CompiledFunction::execute and we get a CompiledFunction via Backend::compile, which implies that we need a backend, no?

And since the DM interface for addNetwork takes a Module, I'm not sure how we're expected to use a DM without a Backend to do something with that Module.

nickgg:
Agree we should get together and work out the use cases for AOT and how we're going to support them. I think its quite important for non-datacenter use cases (w/ heterogeneous partitioning).

I think broadly thats correct, backend is what produces CompiledFunctions - but I guess I don't see why it would need to be in the same process, someone could want could serialize/deserialize the CompiledFunction and seems like it should work?

I believe the Module is used in the DeviceManager not because it's being compiled by the Backend, but rather because it owns the unique mapping for Types and Placeholders.

All that said, I'm fine with this new layout.

bertmaher:
Would you be even fine-er with separating DeviceManagers into their own libraries?  I could do that on top of this PR if it makes sense :)

nickgg:
I'd be happy to leave it for now and reassess later if we need to. We don't have any code that would benefit from the split right now.

PR #2879, opened 2019-05-09
[Interpreter] FullyConnectedNodeKind is not supported by Interpreter backend

beicy:
*Description*:

For non-quantized FC node, Interpreter::isOpSupported() should return false since this FC node is lowered before isOpSupported check.

*Testing*:
ninja check, run.sh
*Documentation*:
[Optional Fixes #issue]

### DIFF
--- a/lib/Backends/Interpreter/Interpreter.cpp
+++ b/lib/Backends/Interpreter/Interpreter.cpp
@@ -144,15 +144,6 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                {Convolution3DNode::BiasIdx}) &&
            (NI.getInElemTy(Convolution3DNode::BiasIdx) == ElemKind::Int32QTy);
 
-  case Kinded::Kind::FullyConnectedNodeKind:
-    if (!NI.getInTy(FullyConnectedNode::InputIdx)->isQuantizedType()) {
-      return NI.allInputsAndOutputsHaveSameElemKind(
-          {ElemKind::FloatTy, ElemKind::Float16Ty});
-    }
-    return NI.allInputsAndOutputsHaveSameElemKind(
-               {ElemKind::Int8QTy}, {FullyConnectedNode::BiasIdx}) &&
-           (NI.getInElemTy(FullyConnectedNode::BiasIdx) == ElemKind::Int32QTy);
-
   case Kinded::Kind::BatchedAddNodeKind:
     if (!NI.getInTy(BatchedAddNode::BatchIdx)->isQuantizedType()) {
       return NI.allInputsAndOutputsHaveSameElemKind(
--- a/tests/unittests/QuantizationTest.cpp
+++ b/tests/unittests/QuantizationTest.cpp
@@ -368,17 +368,37 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
   auto *S = F->createSave("ret", FC);
   bindings.allocate(S->getPlaceholder());
 
+  LoweredInfoMap loweredMapForQuant;
+  ::glow::lower(F, &loweredMapForQuant, EE.getBackend());
+
+  // Get the MatMul node and the Batched_Add node.
+  Node *matMul, *batchedAdd;
+  for (Node &N : F->getNodes()) {
+    if (N.getKind() == Kinded::Kind::MatMulNodeKind) {
+      matMul = &N;
+    }
+    if (N.getKind() == Kinded::Kind::BatchedAddNodeKind) {
+      batchedAdd = &N;
+    }
+  }
+  ASSERT_TRUE(matMul);
+  ASSERT_TRUE(batchedAdd);
+
   quantization::QuantizationConfiguration quantConfig{{
       {NodeQuantizationInfo::generateNodeOutputName(input->getName()),
        {0.2f, 0}},
       {NodeQuantizationInfo::generateNodeOutputName(WC->getName()), {0.3f, 0}},
       {NodeQuantizationInfo::generateNodeOutputName(B->getName()), {0.4f, 0}},
-      {NodeQuantizationInfo::generateNodeOutputName(FC->getName()), {0.6f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(matMul->getName()),
+       {0.6f, 0}},
+      {NodeQuantizationInfo::generateNodeOutputName(batchedAdd->getName()),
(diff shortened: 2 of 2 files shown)


lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -146,8 +146,7 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
 
   case Kinded::Kind::FullyConnectedNodeKind:
```
jfix71: I believe we should actually just remove this entire case for `Kinded::Kind::FullyConnectedNodeKind` -- it is not supported whether quantized or not. The only FullyConnected that is supported by the Interpreter is the RowwiseQuantized version, but it has its own case in here.

beicy: updated it.

tests/unittests/QuantizationTest.cpp
```diff
@@ -405,7 +423,7 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
   EE.compile(CompilationMode::Infer, F);
 
   EE.run(bindings);
-}
+} // namespace glow
```
jfix71: I think `// namespace glow` was accidentally added here by clang-format?

tests/unittests/QuantizationTest.cpp
```diff
@@ -368,17 +368,35 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
...
+  Node *MatMul, *BatchedAdd;
+  for (Node &N : F->getNodes()) {
+    if (N.getKind() == Kinded::Kind::MatMulNodeKind) {
+      MatMul = &N;
+    }
+    if (N.getKind() == Kinded::Kind::BatchedAddNodeKind) {
+      BatchedAdd = &N;
+    }
+  }
+
```
jfix71: You should probably `ASSERT_TRUE()` on the pointers here to make sure they're correctly found.

tests/unittests/QuantizationTest.cpp
```diff
@@ -368,17 +368,35 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
   auto *S = F->createSave("ret", FC);
   bindings.allocate(S->getPlaceholder());
 
+  LoweredInfoMap loweredMapForQuant;
+  ::glow::lower(F, &loweredMapForQuant, EE.getBackend());
+
+  // Get the MatMul node and the Bactched_Add node.
+  Node *MatMul, *BatchedAdd;
```
jfix71: nit: camelCase on these (`matMul`, `batchedAdd`)

tests/unittests/QuantizationTest.cpp
```diff
@@ -368,17 +368,35 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
   auto *S = F->createSave("ret", FC);
   bindings.allocate(S->getPlaceholder());
 
+  LoweredInfoMap loweredMapForQuant;
+  ::glow::lower(F, &loweredMapForQuant, EE.getBackend());
+
+  // Get the MatMul node and the Bactched_Add node.
```
jfix71: typo: Bactched_Add

tests/unittests/QuantizationTest.cpp
```diff
@@ -474,7 +513,8 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnectedSymmetric) {
   auto *offsetsNode = llvm::dyn_cast<Constant>(rwNode->getOffsets().getNode());
   ASSERT_TRUE(offsetsNode);
 
-  // Because we're using symmetric quantization, the offsets should all be zero.
+  // Because we're using symmetric quantization, the offsets should all be
```
jfix71: Looks like from here down it's a lot of accidental changes by clang-format, can you revert these?

tests/unittests/QuantizationTest.cpp
```diff
@@ -430,29 +448,50 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnectedSymmetric) {
...
   EXPECT_EQ(inputTQP.offset, 0);
-  EXPECT_EQ(fcTQP.offset, 0);
+  EXPECT_EQ(matmulTQP.offset, 0);
+  EXPECT_EQ(batchedaddTQP.offset, 0);
   EXPECT_EQ(biasTQP.offset, 0);
 
+  LoweredInfoMap loweredMapForQuant;
+  ::glow::lower(F, &loweredMapForQuant, EE.getBackend());
+
+  // Get the MatMul node and the Bactched_Add node.
```
jfix71: typo: Bactched_Add

tests/unittests/QuantizationTest.cpp
```diff
@@ -430,29 +448,50 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnectedSymmetric) {
...
+  Node *MatMul, *BatchedAdd;
```
jfix71: nit: camelCase on these (`matMul`, `batchedAdd`)

tests/unittests/QuantizationTest.cpp
```diff
@@ -430,29 +448,50 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnectedSymmetric) {
...
+  for (Node &N : F->getNodes()) {
+    if (N.getKind() == Kinded::Kind::MatMulNodeKind) {
+      MatMul = &N;
+    }
+    if (N.getKind() == Kinded::Kind::BatchedAddNodeKind) {
+      BatchedAdd = &N;
+    }
+  }
+
```
jfix71: You should probably `ASSERT_TRUE()` on the pointers here to make sure they're correctly found.

tests/unittests/QuantizationTest.cpp
beicy: Yes.

tests/unittests/QuantizationTest.cpp
beicy: Sure, I can revert it. Do you mean the current format is accepted and we don't need to run "./clang-format fix" ? Thanks!

jfix71: We do need to run clang-format still usually, but it looks like clang-format was run when there was an extra curly bracket somewhere, which made many comments reformat. I wanted to just keep the comments clean since it wasn't necessary. Looks good now I think, thanks!

PR #2880, opened 2019-05-09
[Partitioner] Heterogeneous backends partitioning 

beicy:
*Description*:

This PR adds the feature to support module partitioning for heterogeneous backends.  Please find the discussion in #2687 

*Testing*:
add unit test.

*Documentation*:
#2687 
[Optional Fixes #issue]

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -30,6 +30,17 @@ using ComputeTimeMapTy = std::unordered_map<Node *, float>;
 using NodesSetTy = std::set<Node *>;
 using PartitionCostMapTy = llvm::DenseMap<Function *, GraphMemInfo>;
 
+/// Data structure that contains the info for each type of backend used for
+/// partitioning.
+struct BackendInfo {
+  /// Num of the devices which has the same type of backend.
+  size_t num = 0;
+  /// The memory constraints for this backend kind.
+  uint64_t memSize;
+  /// Backend pointer.
+  Backend *backend = nullptr;
+};
+
 /// Helper structure for building a partition. Records mapping of nodes in
 /// the original function to destination partitions, along with a list of the
 /// newly-created functions;
@@ -47,6 +58,9 @@ struct FunctionNameComparator {
 using FunctionToNodesMapTy =
     std::map<Function *, NodesSetTy, FunctionNameComparator>;
 
+using FunctionToBackendKindMapTy =
+    std::map<Function *, BackendKind, FunctionNameComparator>;
+
 class NodeToFunctionMap {
 
   /// Newly-created partitions.
@@ -55,18 +69,44 @@ class NodeToFunctionMap {
   /// Map of nodes in the original function to their target partition.
   NodeToFunctionMapTy nodeToFunction_;
 
+  FunctionToBackendKindMapTy functionToBackendKind_;
+
   /// Map of sub-functions to their memory consumption.
   PartitionCostMapTy partitionCost_;
 
 public:
-  /// Create a new partition \p F.
-  void createPartition(Function *F) { functions_.emplace_back(F); }
+  /// Create a new partition \p F, and map it with \p backendKind.
+  void createPartition(Function *F, BackendKind backendKind) {
+    functions_.emplace_back(F);
+    functionToBackendKind_[F] = backendKind;
+  }
+
+  BackendKind getPartitionBackendKind(Function *F) {
+    assert(functionToBackendKind_.find(F) != functionToBackendKind_.end() &&
+           "Unknown partition");
+    return functionToBackendKind_.find(F)->second;
+  }
 
   /// Add a new Node->Function mapping.
   void add(Node *N, Function *F) { nodeToFunction_[N] = F; }
 
   /// Get list of functions contained in this map.
   const FunctionList &getPartitions() const { return functions_; }
+
+  /// attach \p map to current mapping.
+  void insert(NodeToFunctionMap &map) {
+    FunctionList flist = map.getPartitions();
+    for (auto it = flist.begin(); it != flist.end(); ++it) {
+      Function *func = *it;
+      auto backendKind = map.getPartitionBackendKind(func);
+      createPartition(func, backendKind);
+    }
+    for (auto it = map.begin(); it != map.end(); ++it) {
+      Node *n = it->first;
+      Function *f = it->second;
+      add(n, f);
+    }
+  }
 
   /// Map API.
   NodeToFunctionMapTy::iterator find(Node *N) {
@@ -100,6 +140,9 @@ class Partitioner {
   /// The cost model related to device.
   std::vector<DeviceInfo> deviceInfo_;
 
+  /// The backend pointers.
+  std::vector<Backend *> backends_;
+
   /// The result of module partitioning.
   DAGListTy partitions_;
 
@@ -116,16 +159,20 @@ class Partitioner {
   /// use all available devices.
   bool saturateHost_;
 
+  // Flag to set if the funcitons in the module are areadly optimized. By
+  // default, the optimization should be done in Partitioner due to
+  // heterogeneous partition.
+  bool optimized_;
+
   /// Get the representative function (the one with the largest input) and
   /// update the memSize.
   static Function *selectRepFunc(Module *parent, uint64_t &memSize);
 
-  /// Get the minimal memory requirement for each op in the representative
-  /// function.
-  void initOpMemUsage();
+  /// Get the minimal memory requirement for each op in the function \p F
+  void initOpMemUsage(Function *F);
 
-  /// Inititalize the minimal compute time for each op in the function.
-  void initOpComputeTime();
+  /// Inititalize the minimal compute time for each op in the function \p F.
+  void initOpComputeTime(Function *F);
 
   /// Combine the partitions if necessary : if all outside uses of the nodes in
   /// partition1 is in partition2, and the sum of memory consumption of
@@ -140,8 +187,9 @@ class Partitioner {
   void partitionsAdjust(NodeToFunctionMap &partitions,
                         uint64_t availableMemory);
 
-  /// Assign nodes to partitions and return the mapping.
-  NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory);
+  /// Assign nodes to partitions with \p backendKind and return the mapping.
+  NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory,
+                                     BackendKind backendKind);
 
   /// Adjust a logicalDevice ID to each DAGNode. It is possible that two
   /// sub-functions need to be assigned into 1 device due to the memory
@@ -151,8 +199,14 @@ class Partitioner {
   /// Duplicates all networks in the module order to saturate the Host.
   void saturateHost(unsigned logicalDeviceCount);
 
-  /// Given the node-function mapping, do the actual partitioning.
-  void doPartitioning(Function *F, NodeToFunctionMap &mapping);
+  FunctionToBackendKindMapTy
+  backendBasedPartition(Function *F, std::vector<Backend *> &backends);
+
+  /// Given the node-function mapping, do the actual partitioning. If \p saveDAG
+  /// is true, the DAG will be saved into partitions_, which is the final
+  /// partition result.
+  void doPartitioning(llvm::StringRef funcName, std::vector<Function *>,
+                      NodeToFunctionMap &mapping, bool saveDAG);
 
 public:
   /// \p parent is the module which contains the functions need to be divided.
@@ -163,9 +217,30 @@ public:
   /// found in Module. The \p devices provides the cost model related to
   /// devices.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-              bool saturateHost = false);
+              bool saturateHost = false, bool optimized = false);
 
-  /// Decompose each function in a module.
+  /// Users can create Mock Backends and pass their points to test Graph
+  /// Partitioning without actually register them in GLOW.
+  Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
+              const std::vector<Backend *> &backends, bool saturateHost = false,
+              bool optimized = false);
+
+  /// Get the map between the backendKind and the concrete backend info (e.g.
+  /// backend pointer, mem, number) used in this partiton. If there are backends
+  /// need to be created, we use \p backendsHolder to hold them for memory
+  /// purpose.
+  void getBackendMap(std::map<BackendKind, BackendInfo> &backendMap,
+                     std::vector<std::unique_ptr<Backend>> &backendsHolder,
+                     std::vector<Backend *> &backends);
+
+  /// If there is no need to do any partition, just generate the DAGNode based
+  /// on current functions in this module.
+  llvm::Error
+  createDAGWithoutPartition(BackendKind backendKind,
+                            std::map<BackendKind, BackendInfo> &backendMap);
+
+  /// Decompose each function in a module. Now we support partitioning a module
+  /// among different type of devices.
   llvm::Error Partition();
 
   /// Get the partitions.
--- a/include/glow/Runtime/Provisioner/Provisioner.h
+++ b/include/glow/Runtime/Provisioner/Provisioner.h
@@ -45,7 +45,7 @@ public:
 private:
   /// Pointer to backend used for compilation. This currently gets reset per
   /// device to ensure the correct backed per device.
-  std::unique_ptr<Backend> backend_;
+  std::vector<std::unique_ptr<Backend>> backends_;
 
   /// Map of compiledFunction unique pointers. This maintains ownership of the
   /// functions.
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -36,8 +36,17 @@ static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
 }
 
 Partitioner::Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-                         bool saturateHost)
-    : module_(parent), deviceInfo_(devices), saturateHost_(saturateHost) {
+                         const std::vector<Backend *> &backends,
+                         bool saturateHost, bool optimized)
+    : module_(parent), deviceInfo_(devices), backends_(backends),
+      saturateHost_(saturateHost), optimized_(optimized) {
+  memSize_ = module_->getConstantsSize();
+}
+
+Partitioner::Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
+                         bool saturateHost, bool optimized)
+    : module_(parent), deviceInfo_(devices), saturateHost_(saturateHost),
+      optimized_(optimized) {
   memSize_ = module_->getConstantsSize();
 }
 
@@ -80,9 +89,9 @@ Function *Partitioner::selectRepFunc(Module *parent, uint64_t &memSize) {
 }
 
 /// Get the minimal memory requirement (constant) for each op in the function.
-void Partitioner::initOpMemUsage() {
+void Partitioner::initOpMemUsage(Function *F) {
   memUsage_.clear();
-  for (auto &node : F_->getNodes()) {
+  for (auto &node : F->getNodes()) {
     int n = node.getNumInputs();
     uint64_t size = 0;
     if (node.getKind() == Kinded::Kind::SaveNodeKind) {
@@ -101,7 +110,7 @@ void Partitioner::initOpMemUsage() {
 }
 
 /// Get the minimal compute time for each op in the function.
-void Partitioner::initOpComputeTime() {
+void Partitioner::initOpComputeTime(Function *F) {
   computeTime_.clear();
 
   // This code assumes all ops are BW limited from SRAM; except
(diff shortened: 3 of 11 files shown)


include/glow/Partitioner/Partitioner.h
```diff
@@ -55,19 +69,45 @@ class NodeToFunctionMap {
...
+  }
 
   /// Add a new Node->Function mapping.
   void add(Node *N, Function *F) { nodeToFunction_[N] = F; }
 
   /// Get list of functions contained in this map.
   const FunctionList &getPartitions() const { return functions_; }
 
+  /// attach \p map1 to current mapping.
+  void insert(NodeToFunctionMap &map1) {
```
opti-mix: Why the name of this parameter is `map1` and not just `map`?

include/glow/Partitioner/Partitioner.h
```diff
@@ -163,9 +218,19 @@ class Partitioner {
...
+              std::vector<Backend *> backends, bool saturateHost = false,
+              bool optimized = false);
+
+  /// Get the map between the backendKind and the concrete backend info (e.g.
+  /// backend pointer, mem, number) used in this partiton.
+  void getBackendMap(std::map<BackendKind, BackendInfo> &backendMap,
+                     std::vector<Backend *> &backends);
 
-  /// Decompose each function in a module.
+  /// 2-level partition: 1. based on backend type; 2. partition.
```
opti-mix: Re-formulate the comment. `2. partition` looks strange

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -45,11 +45,17 @@ auto sortMostMemory = [](const std::pair<DeviceIDTy, uint64_t> &a,
 } // namespace
 
 Provisioner::Provisioner(DeviceManagerMapTy &devices) {
+  std::set<BackendKind> used;
```
opti-mix: You could use `llvm::SmallSet` here to avoid any heap allocations.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -45,11 +45,17 @@ auto sortMostMemory = [](const std::pair<DeviceIDTy, uint64_t> &a,
...
   for (auto &device : devices) {
     devices_.push_back(device.second.get());
+    auto backendKind = device.second->getBackendKind();
+    if (used.find(backendKind) == used.end()) {
+      std::unique_ptr<Backend> tmp;
+      tmp.reset(createBackend(backendKind));
+      backends_.push_back(std::move(tmp));
```
opti-mix: Can these 3 lines be one line? Why do we need `tmp`?

tests/unittests/PartitionerTest.cpp
```diff
@@ -370,3 +373,82 @@ TEST_F(PartitionerTest, Basic1Roofline) {
...
+      mod.createPlaceholder(ElemKind::FloatTy, {16}, "input2", false);
+  auto *sub = F->createSub("sub", input1, input2);
+  auto *mul = F->createMul("mul", input1, input2);
+  auto *sum = F->createMul("add", sub, mul);
+  auto *save = F->createSave("ret", sum);
+  (void)save;
+}
+
+TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
+  // REGISTER_GLOW_BACKEND_FACTORY(BackendWithoutSubFactory, BackendWithoutSub,
```
opti-mix: Should these comments be removed?

lib/Partitioner/Partitioner.cpp
```diff
@@ -36,14 +36,23 @@ static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
 }
 
 Partitioner::Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-                         bool saturateHost)
-    : module_(parent), deviceInfo_(devices), saturateHost_(saturateHost) {
+                         std::vector<Backend *> backends, bool saturateHost,
```
opti-mix: Why do you pass `backend` by value? It seems like it should be `const std::vector<Backend *> &backends`.

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+    }
+  }
+}
+
+FunctionToBackendKindMapTy
+Partitioner::backendBasedPartition(Function *F,
+                                   std::vector<Backend *> backends) {
+  FunctionToBackendKindMapTy ret;
+  NodeToFunctionMap mapping;
+  std::unordered_map<Node *, BackendKind> nodeToBackendKind;
```
opti-mix: Maybe use `llvm::SmallDenseMap` or something like this to avoid heap allocations.

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+    if (funcDAG[subF]->parents.size() == 0) {
+      funcDAG[subF]->parents.push_back(root);
+      root->children.push_back(funcDAG[subF]);
+    }
+  }
+}
+
+FunctionToBackendKindMapTy
+Partitioner::backendBasedPartition(Function *F,
+                                   std::vector<Backend *> backends) {
```
opti-mix: Why do you pass backend by value? It seems like it should be const std::vector<Backend *> &backends.

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+  FunctionToBackendKindMapTy ret;
+  NodeToFunctionMap mapping;
+  std::unordered_map<Node *, BackendKind> nodeToBackendKind;
+
+  // For each node find a backend that supports it.
+  for (auto &N : F->getNodes()) {
+    for (auto &backend : backends) {
+      // Find the first backend that supports this node. The order of backends
+      // is important.
+      if (backend->shouldLower(&N) || backend->isOpSupported(N)) {
```
opti-mix: Why this `shouldLower` check?

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+        newF = F->getParent()->createFunction(
+            std::string(F->getName()) + "_part" + std::to_string(++color));
+        mapping.createPartition(newF, backendKind);
+        ret[newF] = backendKind;
+      }
+      mapping.add(N, newF);
+    }
+  }
+
+  // Here we just need to split the function without generating DAG.
```
opti-mix: Could it be that this remaining piece of code until the end of the function is very similar to another partitioning function? If so, could it be moved into a helper function and used from both?

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
-  module_->eraseFunction(F_);
 
-  auto funcList = module_->getFunctions();
-  for (Function *F : funcList) {
-    (void)F;
-    assert(F->verify() && "Conversion led to invalid function");
+  // Step 1 : do the partition based on backends type.
+  FunctionToBackendKindMapTy funcToBackend;
+  std::string origName(F_->getName().data());
+  if (backends.size() == 1) {
```
opti-mix: Is it necessary to special case when there is only 1 backend? Shouldn't the general case of multiple backends cover this case already? Of course, it would create just a single backend type based partition in this case, which is a waste. But it would work. 

Of course, if it is just a performance optimization, then it is fine. But then explain it in the comment.

include/glow/Partitioner/Partitioner.h
```diff
@@ -100,6 +140,9 @@ class Partitioner {
   /// The cost model related to device.
   const std::vector<DeviceInfo> &deviceInfo_;
 
+  /// The backend pointers.
+  std::vector<Backend *> backends_;
```
opti-mix: Do you need a copy of the backends vector here? Or would `const std::vector<Backend*> &backends_` be enough?

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -45,7 +45,7 @@ class Provisioner final {
 private:
   /// Pointer to backend used for compilation. This currently gets reset per
   /// device to ensure the correct backed per device.
-  std::unique_ptr<Backend> backend_;
+  std::vector<std::unique_ptr<Backend>> backends_;
```
opti-mix: Do you need a copy of the backends vector here? Or would `std::vector<std::unique_ptr<Backend>> &backends_` be enough?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -111,31 +125,44 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
   std::sort(deviceMemory.begin(), deviceMemory.end(), sortMostMemory);
 
   // Try to add functions to devices in order from largest to smallest.
+  std::map<BackendKind, size_t> startPos;
   for (unsigned i = 0; i < logicalDeviceSize.size(); i++) {
```
opti-mix: I'm not sure I understand the logic here. But I remember that in the preliminary version of this PR that I tested there was a bug in this loop related to how deviceIds were looked up. So, please re-check the logic just in-case.

opti-mix:
@beicy I think this PR may also need to introduce a way (e.g. command-line options similar to the ones currently used to select the backend) to specify which backends should be used for running a specific model and in which order.

You could e.g. extend the current `ExecutionBackend` command-line option to take a comma separated list of backends instead of just a single one (and you should check that each backend is used only once in this list?).

And you may also need to introduce a way to pass backend-specific options to each of those backends. E.g. `-target`, `-march` and `-mcpu` flags are currently global and used by all LLVMBackend based backends, which would lead to wrong results if you have multiple such backends for heterogeneous partitioning. So, ideally, we may want a way to pass each of those backends its own list of options.

include/glow/Partitioner/Partitioner.h
```diff
@@ -30,6 +30,17 @@ using ComputeTimeMapTy = std::unordered_map<Node *, float>;
 using NodesSetTy = std::set<Node *>;
 using PartitionCostMapTy = llvm::DenseMap<Function *, GraphMemInfo>;
 
+/// Data structure that contains the info for each type of backend used for
+/// partitioning.
+struct BackendInfo {
+  // Num of the devices which has the same type of backend.
```
nadavrot: Members should be commented with three slashes '///'. This will allow Doxygen to keep the comment in the docs.

include/glow/Partitioner/Partitioner.h
```diff
@@ -116,16 +159,21 @@ class Partitioner {
...
+  bool optimized_;
+
   /// Get the representative function (the one with the largest input) and
   /// update the memSize.
   static Function *selectRepFunc(Module *parent, uint64_t &memSize);
 
   /// Get the minimal memory requirement for each op in the representative
   /// function.
-  void initOpMemUsage();
+  void initOpMemUsage(Function *F);
```
nadavrot: What is F?  Please document the code.

include/glow/Partitioner/Partitioner.h
```diff
@@ -141,7 +189,8 @@ class Partitioner {
                         uint64_t availableMemory);
 
   /// Assign nodes to partitions and return the mapping.
-  NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory);
+  NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory,
```
nadavrot: Update the comment please.

include/glow/Partitioner/Partitioner.h
```diff
@@ -163,9 +218,19 @@ class Partitioner {
   /// found in Module. The \p devices provides the cost model related to
   /// devices.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-              bool saturateHost = false);
+              bool saturateHost = false, bool optimized = false);
+
+  /// For mock backends. Mainly used in testing.
```
nadavrot: Please document the parameters.

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   }
 }
 
+void Partitioner::doFuncListPartitioning(llvm::StringRef funcName,
+                                         std::vector<Function *> funcs,
+                                         NodeToFunctionMap &mapping) {
+  // The dummy node.
```
nadavrot: What is a dummy node?  I don't understand the algorithm. It should be something like:

// Add a dummy node that will allow us to do XXX ...

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+    }
+
+    // Link subF to its parents.
+    std::set<Function *> parents;
+    for (auto &N : subF->getNodes()) {
+      for (int inp = 0, e = N.getNumInputs(); inp < e; inp++) {
+        auto input = N.getNthInput(inp);
+        if (isa<Storage>(input.getNode())) {
+          // It is possible that one input is the output of anther function.
+          Placeholder *ph = llvm::dyn_cast<Placeholder>(input.getNode());
```
nadavrot: This code is very nested and difficult to follow:

lib/Partitioner/Partitioner.cpp
beicy: Yes, it is due to the performance. We don't want to do a function copy (which is a waste) since no function split is necessary

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   }
 }
 
+void Partitioner::doFuncListPartitioning(llvm::StringRef funcName,
```
nadavrot: This function is very long. Is there a way to split it to small utils?

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+  for (auto *subF : mapping.getPartitions()) {
+    if (funcDAG[subF]->parents.size() == 0) {
+      funcDAG[subF]->parents.push_back(root);
+      root->children.push_back(funcDAG[subF]);
+    }
+  }
+}
+
+FunctionToBackendKindMapTy
+Partitioner::backendBasedPartition(Function *F,
```
nadavrot: This function is very long. Can we refactor it and extract some logic elsewhere?

lib/Partitioner/Partitioner.cpp
```diff
@@ -569,68 +581,390 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
...
+        nodes.push_back(std::move(DAG1));
+        partitions_.push_back({std::move(DAG0), std::move(nodes)});
+      }
+      if (saturateHost_) {
+        // Saturate the Host.
+        saturateHost(1);
+      }
+      return llvm::Error::success();
+    }
+    F_->dumpDAG("init.dot");
```
nadavrot: You forgot this.

include/glow/Partitioner/Partitioner.h
beicy: Actually, this one is only for test -- passing the "mock" backends.  If we have a better idea on how to create "mock" backends in unit test by using backendKind, no need to keep this field.

lib/Runtime/Provisioner/Provisioner.cpp
beicy: It is different from the version I sent to you before. The basic idea is, we still sort the logical devices and physical devices based on the available memory size from high to low. And then for each logical device, we select the physical device who has the same backendKind as the logical one and the largest available memory.

tests/unittests/PartitionerTest.cpp
```diff
@@ -123,12 +123,14 @@ TEST_F(PartitionerTest, Basic1) {
...
-  Partitioner myPartitioner(&mod_, devices);
+  std::vector<DeviceInfo> devices = {{3072, BackendKind::Interpreter},
+                                     {3072, BackendKind::Interpreter},
+                                     {3072, BackendKind::Interpreter}};
+  Partitioner myPartitioner(&mod_, devices, false, true);
   auto err = myPartitioner.Partition();
   EXPECT_FALSE(errToBool(std::move(err)));
   DAGListTy myList = std::move(myPartitioner.getPartitionResult());
-  ASSERT_EQ(mod_.getFunctions().size(), 3);
+  //  ASSERT_EQ(mod_.getFunctions().size(), 3);
```
gcatron: You can remove this commented out assert.

gcatron:
> also need to introduce a way (e.g. command-line options similar to the ones currently used to select the backend) to s

I agree we will need to take some time to extend our command line options and maybe go a config file route. I think that should be done in a follow up PR. We might want to make sure it aligns with the DeviceManager config work being done too.

lib/Partitioner/Partitioner.cpp
beicy: If an op(e.g. FC)  can be lowered in the backend, it means this op is supported. If not, we will check if this op can be supported directly by this Backend.

jfix71: Seeing this in action, I don't think it makes sense actually. Backends will return true for Nodes they do not know about. E.g. if a backend does not support `SparseLengthsWeightedSum`, it will return false for `isOpSupported()`, but true for `shouldLower()` even though `SparseLengthsWeightedSum` isn't lowered to anything.

One alternative here could be that when we encounter an unsupported node, clone it, and then try to lower it. If it's not lowered then it's unsupported. Else if it is lowered then recursively do the same check on all of the nodes added during lowering. This is going to require a non-trivial refactoring though, in lowering logic/figuring out how best to clone the Node temporarily and tracking the nodes created from lowering.

include/glow/Runtime/Provisioner/Provisioner.h
beicy: This one is not passed from HostManager, all the backends are created in Provisioner.

lib/Partitioner/Partitioner.cpp
beicy: > Seeing this in action, I don't think it makes sense actually. Backends will return true for Nodes they do not know about. E.g. if a backend does not support `SparseLengthsWeightedSum`, it will return false for `isOpSupported()`, but true for `shouldLower()` even though `SparseLengthsWeightedSum` isn't lowered to anything.
> 
> One alternative here could be that when we encounter an unsupported node, clone it, and then try to lower it. If it's not lowered then it's unsupported. Else if it is lowered then recursively do the same check on all of the nodes added during lowering. This is going to require a non-trivial refactoring though, in lowering logic/figuring out how best to clone the Node temporarily and tracking the nodes created from lowering.

Thanks for pointing it out!

include/glow/Partitioner/Partitioner.h
```diff
@@ -30,6 +30,17 @@ using ComputeTimeMapTy = std::unordered_map<Node *, float>;
...
 
+/// Data structure that contains the info for each type of backend used for
+/// partitioning.
+struct BackendInfo {
+  /// Num of the devices which has the same type of backend.
+  size_t num = 0;
+  /// The memory constraints for this backend kind.
+  uint64_t memSize;
+  /// Backend pointer.
+  Backend *backend;
```
opti-mix: Should the `backend` pointer be intialized with `nullptr` by default?

tests/unittests/PartitionerTest.cpp
```diff
@@ -385,3 +388,78 @@ TEST_F(PartitionerTest, SelectRepFunc) {
   auto err = myPartitioner.Partition();
   EXPECT_FALSE(errToBool(std::move(err)));
 }
+
+template <BackendKind kind, glow::Kinded::Kind unsupportedOpKind>
```
opti-mix: Could you add a comment about the purpose of the class and describe the meaning of its template parameters.

tests/unittests/PartitionerTest.cpp
```diff
@@ -385,3 +388,78 @@ TEST_F(PartitionerTest, SelectRepFunc) {
...
+    return true;
+  }
+
+  bool shouldLower(const Node *N) const override { return false; }
+
+  bool generateInst(Node *N, IRGenVisitor &irgen) const override {
+    return false;
+  }
+};
+
```
opti-mix: Add a comment explaining why these classes are needed and what they try to model.

lib/Partitioner/Partitioner.cpp
opti-mix: @beicy The comment about nesting is not addressed yet.

Use `if (!isa<Storage>(input.getNode())) {continue;}` and later `if (!ph) {continue;}` as early exits to reduce nesting.

beicy: Thanks! I will rewrite the logic as :  
```
if (isa<Storage>(input.getNode())) {
   if (isa<Constant>(input.getNode())) { continue;}
   //Handle Placeholder;
   continue;
}
```
The idea is, in one function, when we do the partition, we don't have to consider the Storage nodes. But if it possible that one Placeholder is the output of another function, then we need to handle this situation.

rdzhabarov:
@beicy please also do not forget to import to phabricator (just did it).

lib/Partitioner/Partitioner.cpp
opti-mix: Do you plan to address this part in this PR?

beicy: Sorry missed this part... Improved the code in this version. Thanks!

PR #2882, opened 2019-05-10
[Partitioner] Fix selectRepFunc() not to return null.

artemrakhov:
Also make sure that `&memSize` is updated correctly.


artemrakhov:
Also changed the class to own (copy) `deviceInfo_`

PR #2883, opened 2019-05-10
[ModelRunner] Allow Quantized output, save Function's name.

artemrakhov:


tools/loader/ModelRunner.cpp
```diff
@@ -53,10 +55,19 @@ int main(int argc, char **argv) {
   if (!emittingBundle()) {
     loader.runInference(bindings);
 
-    llvm::outs() << "Model: " << loader.getFunction()->getName() << "\n";
+    llvm::outs() << "Model: " << modelName << "\n";
 
     // Print out the result of output operator.
-    outputT->getHandle().dump();
+    switch (outputT->getElementType()) {
```
rdzhabarov: I think Tensor class already has `dump` method (which implements this switch), we should use that instead of doing switch here.

PR #2884, opened 2019-05-10
Use onnxifi output tensors directly

jackm321:
Summary: Instead of using separate Glow tensors and copying outputs, use outputs directly

Differential Revision: D15291353


nadavrot:
I tried to read the patch but there are almost no comments that explain what's going on. You had just touched the code, could you please add comments that explain what's going on in the area?

For example: 
 // In this section we are checking if the output tensor is in the right format and if it is then we reuse the original buffer ....

lib/Onnxifi/Base.cpp
```diff
@@ -168,16 +166,30 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
 
     auto &outPhPtr = outPhIt->getValue();
 
-    phNameToOnnxTensorOutputs[outPhPtr] = outOnnxTensor;
+    // Compute the total size of the onnxifi tensor.
+    std::vector<size_t> outOnnxTensorDims(outOnnxTensor.dimensions);
```
nickgg: Minor, but I believe we can skip building this vector by constructing an ArrayRef from `outOnnxTensor.shape, outOnnxTensor.dimensions`.

We can probably then get the total size via std::accumulate over that ArrayRef but that might not end up clearer.

jackm321:
> can you paste in the trace screenshots
They are in the diff's test plan

lib/Onnxifi/Base.cpp
jackm321: Eh ok I just tried this out and I think it looks more confusing, I'm just going to keep it like it is matching the inputs code

lib/Onnxifi/Base.cpp
```diff
@@ -168,16 +166,30 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+    size_t outOnnxTensorSize = 1;
+    for (unsigned j = 0; j < outOnnxTensor.dimensions; ++j) {
+      outOnnxTensorDims[j] = outOnnxTensor.shape[j];
+      outOnnxTensorSize *= outOnnxTensorDims[j];
+    }
 
-    Tensor t(outPhPtr->getType());
+    // Check that tensor provided by onnxifi is the correct size.
+    if (!outPhPtr->dims().equals(outOnnxTensorDims)) {
+      llvm::outs() << "Output tensor is the wrong shape: " << outOnnxTensorSize
```
rdzhabarov: nit: s/is/has

Issue #2554, opened 2019-03-15
[RFC] Support for multiple training algorithms


SplitInfinity:
**Preface**
In #2524, I added support for the Adagrad training algorithm in preparation for adding training support for `SparseLengthsWeightedSum` in #2541. The design I chose was not well-received, so I want to propose a new design based on the feedback in that PR and discuss it before I update the PR.

**Problem Statement**
`glow::differentiate` adds `SGDNodes` for every trainable `Placeholder` in a Graph that calculate the new updated values based on the stochastic gradient descent algorithm. It also adds `SaveNodes` to save these updated values back into their corresponding `Placeholders`. However, this approach does not allow Glow to support multiple training algorithms (i.e. multiple ways to compute new weights).

**Proposal**
There is a simple POD struct named `TrainingConfig` that includes a bunch of parameters used by stochastic gradient descent:

```
struct TrainingConfig {
  float L1Decay{0};
  float L2Decay{0};
  float learningRate{0.01f};
  float momentum{0.0};
  unsigned batchSize{1};
};
```

This struct is passed into `glow::differentiate` and used to construct `SGDNodes` with the proper learning rate, momentum, etc. I propose that we replace this with the following:

```
class Trainer {
public:
  Trainer(float learningRate, unsigned batchSize) : learningRate_(learningRate), batchSize_(batchSize) {}
  virtual bool instrument(Function* F, PlaceholderBindings& bindings) = 0;
protected:
  // Parameters used by multiple subclasses
  float learningRate_{0.01f};
  unsigned batchSize_{1};
};
```
 
New training algorithms are implemented as follows:

```
// Stochastic gradient descent
class SGDTrainer : final Trainer {
public:
  SGDTrainer(float learningRate, unsigned batchSize, float L1Decay, float L2Decay, float momentum) : Trainer(learningRate, batchSize), L1Decay_(L1Decay), L2Decay_(L2Decay), momentum_(momentum) {}
  bool instrument(Function* F, PlaceholderBindings& bindings) override {
    // code here to add SGDNode to F, save new weights into Placeholders, etc.

   // return true if instrumentation was successful
  }
private:
  float L1Decay{0};
  float L2Decay{0};
  float momentum{0.0};
};

// Adagrad
class AdagradTrainer : final Trainer {
public:
  AdagradTrainer(float learningRate, unsigned batchSize, float epsilon) : Trainer(learningRate, batchSize), epsilon_(epsilon) {}
[...]

PR #2890, opened 2019-05-11
ResNet training with Glow example.

putivsky:
*Description*:
Simple example, how to train ResNet50 with Glow
*Testing*:
Run unit test, training and evaluation ResNet50 on CIFAR images.
*Documentation*:

### DIFF
--- a/examples/training/resnet50/main.cpp
+++ b/examples/training/resnet50/main.cpp
@@ -1,0 +1,246 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Timer.h"
+
+#include <fstream>
+
+/// This is an example demonstrating how to load resnet50 model,
+/// randomize weights and perform training on a limited data set,
+/// and evaluate model.
+
+constexpr size_t IMAGE_COLORS = 3;
+constexpr size_t IMAGE_HEIGHT = 224;
+constexpr size_t IMAGE_WIDTH = 224;
+constexpr size_t LABELS_SIZE = 1000;
+constexpr size_t CIFAR_NUM_IMAGES = 20;
+
+using namespace glow;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-training Options");
+llvm::cl::opt<unsigned> numEpochs("epochs", llvm::cl::desc("Number of epochs."),
+                                  llvm::cl::init(32), llvm::cl::value_desc("N"),
+                                  llvm::cl::cat(category));
+llvm::cl::opt<unsigned> miniBatchSize("mini-batch",
+                                      llvm::cl::desc("Mini batch size."),
+                                      llvm::cl::init(8),
+                                      llvm::cl::value_desc("N"),
+                                      llvm::cl::cat(category));
+llvm::cl::opt<float> learningRate("learning-rate",
+                                  llvm::cl::desc("Learning rate parameter."),
+                                  llvm::cl::init(0.001),
+                                  llvm::cl::value_desc("N"),
+                                  llvm::cl::cat(category));
+llvm::cl::opt<float> momentum("momentum", llvm::cl::desc("Momentum parameter."),
+                              llvm::cl::init(0.9), llvm::cl::value_desc("N"),
+                              llvm::cl::cat(category));
+llvm::cl::opt<float> l2Decay("l2-decay", llvm::cl::desc("L2Decay parameter."),
+                             llvm::cl::init(0.0001), llvm::cl::value_desc("N"),
+                             llvm::cl::cat(category));
+llvm::cl::opt<BackendKind> executionBackendKind(
+    llvm::cl::desc("Backend to use:"), llvm::cl::Optional,
+    llvm::cl::values(clEnumValN(BackendKind::Interpreter, "Interpreter",
+                                "Use interpreter (default option)"),
+                     clEnumValN(BackendKind::CPU, "cpu", "Use CPU"),
+                     clEnumValN(BackendKind::OpenCL, "opencl", "Use OpenCL")),
+    llvm::cl::init(BackendKind::Interpreter), llvm::cl::cat(category));
+
+/// Finds the maximum value and its index within \p N range.
+size_t findMaxIndex(const Tensor &t, size_t range) {
+  auto handle = t.getHandle();
+  size_t ret{0};
+  float maxVal{handle.raw(0)};
+  for (size_t c = 1; c < range; ++c) {
+    auto val = handle.raw(c);
+    if (maxVal <= val) {
+      maxVal = val;
+      ret = c;
+    }
+  }
+  return ret;
+}
+
+void loadImagesAndLabels(Tensor &images, Tensor &labels) {
+  llvm::outs() << "Loading the CIFAR-10 database.\n";
+  /// ResNet expects ImageChannelOrder::BGR, ImageLayout::NCHW.
+  std::ifstream dbInput("cifar-10-batches-bin/data_batch_1.bin",
+                        std::ios::binary);
+
+  if (!dbInput.is_open()) {
+    llvm::errs() << "Failed to open cifar10 data file, probably missing.\n";
+    llvm::errs() << "Run 'python ../glow/utils/download_test_db.py'\n";
+    exit(1);
+  }
+
+  /// The CIFAR file format is structured as one byte label in the range 0..9.
+  /// The label is followed by an image: 32 x 32 pixels, in RGB format. Each
+  /// color is 1 byte. The first 1024 red bytes are followed by 1024 of green
+  /// and blue. Each 1024 byte color slice is organized in row-major format.
+  /// The database contains 10000 images.
+  /// Size: (1 + (32 * 32 * 3)) * 10000 = 30730000.
+  llvm::outs() << "Assigning images/labels bytes.\n";
+  auto labelsH = labels.getHandle();
+  auto imagesH = images.getHandle();
+  float scale = 1. / 255.0;
+  float bias = 0;
+
+  /// CIFAR image dimensions are 32x32, ResNet50 expects 224x224.
+  /// Simple scaling would be one CIFAR pixel copied into 7x7 area
+  constexpr unsigned expandImage = 7;
+  for (unsigned n = 0; n < CIFAR_NUM_IMAGES; ++n) {
+    unsigned long l = dbInput.get();
+    labelsH.at({n, l}) = 1;
+    // Even ResNet50 model got trained in NCHW format
+    for (unsigned c = 0; c < IMAGE_COLORS; ++c) {
+      auto bgrc = IMAGE_COLORS - 1 - c;
+      for (unsigned h = 0; h < 32; ++h) {
+        for (unsigned w = 0; w < 32; ++w) {
+          // ResNet BGR color space vs CIFAR RGB.
+          auto val = static_cast<float>(static_cast<uint8_t>(dbInput.get()));
+          // Normalize and scale image.
+          val = (val - imagenetNormMean[c]) * scale / imagenetNormStd[c] + bias;
+          for (unsigned sh = 0; sh < expandImage; ++sh) {
+            for (unsigned sw = 0; sw < expandImage; ++sw) {
+              imagesH.at(
+                  {n, bgrc, sh + h * expandImage, sw + w * expandImage}) = val;
+            }
+          }
+        }
+      }
+    }
+  }
+}
+} // namespace
+
+int main(int argc, char **argv) {
+  llvm::cl::ParseCommandLineOptions(argc, argv,
+                                    " ResNet50 Training Example\n\n");
+
+  // We expect the input to be NCHW.
+  llvm::ArrayRef<size_t> allImagesDims = {CIFAR_NUM_IMAGES, IMAGE_COLORS,
+                                          IMAGE_HEIGHT, IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> batchImagesDims = {miniBatchSize, IMAGE_COLORS,
+                                            IMAGE_HEIGHT, IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> initShapeDims = {1, IMAGE_COLORS, IMAGE_HEIGHT,
+                                          IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> allLabelsDims = {CIFAR_NUM_IMAGES, LABELS_SIZE};
+
+  Module mod;
+  PlaceholderBindings bindings;
+  const char inputName[] = "gpu_0/data";
+  TypeRef inputType = mod.uniqueType(ElemKind::FloatTy, initShapeDims);
+  // Load ResNet model.
+  Function *F = mod.createFunction("resnet50");
+  llvm::outs() << "Loading resnet50 model.\n";
+  // Loader has randomly initialized trainable weights.
+  Caffe2ModelLoader loader("resnet50/predict_net.pbtxt", "resnet50/init_net.pb",
+                           {inputName}, {inputType}, *F, nullptr, &bindings);
+
+  llvm::outs() << "Creating input/output placeholders.\n";
+  // Create the input layer.
+  auto *A = mod.createPlaceholder(ElemKind::FloatTy, batchImagesDims, "input",
+                                  /* isTraining */ false);
+  bindings.allocate(A);
+  // Create the output layer.
+  auto *E =
+      mod.createPlaceholder(ElemKind::FloatTy, {miniBatchSize, LABELS_SIZE},
+                            "expected", /* isTraining */ false);
+  bindings.allocate(E);
+
+  llvm::outs() << "Differentiating graph.\n";
+  TrainingConfig TC;
+  TC.learningRate = learningRate;
+  TC.momentum = momentum;
+  TC.L2Decay = l2Decay;
+  TC.batchSize = miniBatchSize;
+  Function *TF = glow::differentiate(F, TC);
+
+  llvm::outs() << "Compiling backend.\n";
+  BackendKind executionBackend{executionBackendKind};
+  ExecutionEngine EE(executionBackend);
+  EE.compile(CompilationMode::Train, TF);
+
+  // Allocate placeholders without tensors.
+  bindings.allocate(mod.getPlaceholders());
+
+  // Get output tensor.
+  auto *resultPH = mod.getPlaceholderByName("save_gpu_0_softmax");
+  auto *result = bindings.get(resultPH);
+
+  // These tensors allocate memory for all images and labels prepared for
+  // training.
+  Tensor images(ElemKind::FloatTy, allImagesDims);
+  Tensor labels(ElemKind::FloatTy, allLabelsDims);
+
+  loadImagesAndLabels(images, labels);
+
+  // From http://www.cs.toronto.edu/~kriz/cifar.html.
+  const char *textualLabels[LABELS_SIZE] = {
+      "airplane", "automobile", "bird",  "cat",  "deer",
+      "dog",      "frog",       "horse", "ship", "truck"};
+
+  size_t sampleCounter = 0;
+  int reportRate = 16;
+
+  llvm::Timer timer("Training", "Training");
+
+  llvm::outs() << "Training.\n";
+  auto labelsH = labels.getHandle();
+  for (size_t e = 0; e < numEpochs; ++e) {
+    llvm::outs() << "Training - epoch #" << e << " from total " << numEpochs
+                 << "\n";
+    // Train network on all examples.
+    timer.startTimer();
+    runBatch(EE, bindings, reportRate, sampleCounter, {A, E},
+             {&images, &labels});
+    timer.stopTimer();
+
+    llvm::outs() << "Evaluating.\n";
+    unsigned score = 0;
+    unsigned total = 0;
+
+    for (unsigned int i = 0; i < CIFAR_NUM_IMAGES / miniBatchSize; ++i) {
+      Tensor sample(ElemKind::FloatTy, batchImagesDims);
+      sample.copyConsecutiveSlices(&images, miniBatchSize * i);
+      updateInputPlaceholders(bindings, {A}, {&sample});
+      EE.run(bindings);
+
+      for (unsigned int example = 0; example < miniBatchSize; ++example) {
+        auto correct =
+            findMaxIndex(labelsH.extractSlice(miniBatchSize * i + example), 10);
+        auto guess = findMaxIndex(*result, 10);
+
+        score += guess == correct;
+        ++total;
+
+        if (i == 0) {
+          llvm::outs() << example << ") Expected: " << textualLabels[correct]
+                       << ", Got: " << textualLabels[guess] << "\n";
+        }
+      }
+    }
+
+    llvm::outs() << "Total accuracy: " << 100. * score / total << "\n";
+  }
+
+  return 0;
+}
--- a/include/glow/Importer/Caffe2ModelLoader.h
+++ b/include/glow/Importer/Caffe2ModelLoader.h
@@ -54,11 +54,17 @@ class Caffe2ModelLoader
   llvm::Error loadWeight(const caffe2::OperatorDef &op);
 
   /// Load the structure of the network from the 'net' file.
-  llvm::Error loadNetwork(caffe2::NetDef &net);
+  /// If Bindings are provided creates Placeholders vs Constants.
+  llvm::Error loadNetwork(caffe2::NetDef &net,
+                          PlaceholderBindings *bindings = nullptr);
 
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  llvm::Error loadOperator(const caffe2::OperatorDef &op);
+  /// \p bindings is an optional parameter. If it is provided,
+  /// the loaded weights will be discarded, replaced by trainable weights that
+  /// are initialized for training and added to the graph as Placeholders.
+  llvm::Error loadOperator(const caffe2::OperatorDef &op,
+                           PlaceholderBindings *bindings = nullptr);
 
   /// Reads a network (weights or structure) from the serialized protocol buffer
   /// file.
@@ -91,11 +97,15 @@ public:
   /// outputs with specific names and types.
   /// If \p errPtr is not null then if an error occurs it will get assigned
   /// there otherwise if an error occurs it will abort.
+  /// \p bindings is an optional parameter. If it is provided,
+  /// the loaded weights will be discarded, replaced by trainable weights that
+  /// are initialized for training and added to the graph as Placeholders.
   Caffe2ModelLoader(const std::string &netDescFilename,
                     const std::string &netWeightFilename,
                     llvm::ArrayRef<const char *> names,
                     llvm::ArrayRef<TypeRef> types, Function &F,
-                    llvm::Error *errPtr = nullptr);
+                    llvm::Error *errPtr = nullptr,
+                    PlaceholderBindings *bindings = nullptr);
 
   /// Creates a Caffe2 model loader to build \p F.
   /// If \p errPtr is not null then if an error occurs it will get assigned
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -71,11 +71,15 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
     // Concat ==> Splat + Insert. Both only support the following.
   case Kinded::Kind::ConcatNodeKind:
   case Kinded::Kind::SplatNodeKind:
-  case Kinded::Kind::TransposeNodeKind:
   case Kinded::Kind::SliceNodeKind:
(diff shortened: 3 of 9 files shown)


examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,252 @@
...
+          Val = (Val * Scale - NormMean[C]) / NormStd[C] + Bias;
+          for (unsigned SH = 0; SH < SCALE; ++SH) {
+            for (unsigned SW = 0; SW < SCALE; ++SW) {
+              ImagesH.at({N, BRGC, SH + H * SCALE, SW + W * SCALE}) = Val;
+            }
+          }
+        }
+      }
+    }
+    /*
```
nadavrot: oops

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -319,8 +320,28 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
         }
       }
       outTy = G_.getParent()->uniqueType(ElemKind::FloatTy, outDims);
-      filter = G_.getParent()->createConstant("conv.filter", wtag);
-      bias = G_.getParent()->createConstant("conv.bias", biasTensor);
+      if (bindings) {
```
nadavrot: I think that this section could use a few comments.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -532,10 +577,41 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+      auto *phScale =
+          G_.getParent()->createPlaceholder(&scale->getType(), "scale", true);
+      bindings->insert(phScale, scale->clone());
+      bindings->get(phScale)->init(glow::Tensor::InitKind::Zero, 0,
+                                   G_.getPRNG());
+      scaleV = phScale;
+      auto *phBias =
+          G_.getParent()->createPlaceholder(&bias->getType(), "bias", true);
+      bindings->insert(phBias, bias->clone());
+      // bindings->get(phBias)->init(glow::Tensor::InitKind::Broadcast, 1.0,
```
nadavrot: oops.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -532,10 +577,41 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
+          G_.getParent()->createPlaceholder(&mean->getType(), "mean", false);
+      bindings->insert(phMean, mean->clone());
+      bindings->get(phMean)->zero();
+      meanV = phMean;
+      auto *phVar =
+          G_.getParent()->createPlaceholder(&var->getType(), "var", false);
+      bindings->insert(phVar, var->clone());
+      bindings->get(phVar)->init(glow::Tensor::InitKind::Broadcast, 1.0,
+                                 G_.getPRNG());
+      // bindings->get(phVar)->zero();
```
nadavrot: oops

putivsky:
Tests are succeeded on dev box. MLTest is flaky or took too long to finish > 20 seconds.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+  /// The database contains 10000 images.
+  /// Size: (1 + (32 * 32 * 3)) * 10000 = 30730000.
+  static const float NormMean[] = {0.485, 0.456, 0.406};
+  static const float NormStd[] = {0.229, 0.224, 0.225};
+
+  static const std::pair<float, float> Range = {0., 1.};
+
+  llvm::outs() << "Assigned Images/Labels bytes.\n";
+  auto LabelsH = Labels.getHandle<float>();
+  auto ImagesH = Images.getHandle<>();
```
SplitInfinity: Same, I think you can get rid of the angle brackets.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+            for (unsigned SW = 0; SW < SCALE; ++SW) {
+              ImagesH.at({N, BRGC, SH + H * SCALE, SW + W * SCALE}) = Val;
+            }
+          }
+        }
+      }
+    }
+  }
+
+  const char *TextualLabels[LABELS_SIZE] = {
```
SplitInfinity: These don't seem like the CIFAR labels. What are they?

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+  int ReportRate = 16;
+
+  llvm::Timer timer("Training", "Training");
+
+  llvm::outs() << "Training.\n";
+  for (size_t e = 0; e < TrainingEpochs; ++e) {
+    llvm::outs() << "Training - epoch #" << e << " from total "
+                 << TrainingEpochs << "\n";
+    // Bind the images tensor to the input array A, and the labels tensor
+    // to the softmax node SM.
```
SplitInfinity: ??

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,238 @@
...
+      Sample.copyConsecutiveSlices(&Images, MiniBatchSize * i);
+      updateInputPlaceholders(Bindings, {A}, {&Sample});
+      EE.run(Bindings);
+
+      for (unsigned int Batch = 0; Batch < MiniBatchSize; ++Batch) {
+        auto CT = LabelsH.extractSlice(MiniBatchSize * i + Batch);
+        auto CH = CT.getHandle<>();
+        auto CorrectVal = CH.raw(0);
+        size_t CorrectIdx = 0;
+        for (size_t c = 0; c < 10; ++c) {
```
SplitInfinity: This code and the code below seem identical but operate on different tensors. Consider refactoring.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,18 @@ class Caffe2ModelLoader
   llvm::Error loadWeightsFromNet(caffe2::NetDef &net);
 
   /// Loads an individual weight \p op.
+  /// If bindings got provided creates placeholders vs constants.
```
SplitInfinity: `got` -> `are`
`placeholders` -> `Placeholders`
`constants` -> `Constants`

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,18 @@ class Caffe2ModelLoader
...
   llvm::Error loadWeight(const caffe2::OperatorDef &op);
 
   /// Load the structure of the network from the 'net' file.
-  llvm::Error loadNetwork(caffe2::NetDef &net);
+  /// If bindings got provided creates placeholders vs constants.
```
SplitInfinity: `got` -> `are`
`placeholders` -> `Placeholders`
`constants` -> `Constants`

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,18 @@ class Caffe2ModelLoader
...
+  llvm::Error loadNetwork(caffe2::NetDef &net,
+                          PlaceholderBindings *bindings = nullptr);
 
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  llvm::Error loadOperator(const caffe2::OperatorDef &op);
+  llvm::Error loadOperator(const caffe2::OperatorDef &op,
```
SplitInfinity: Add comment regarding `bindings` parameter.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -532,10 +586,39 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
     unsigned_t channel;
     ASSIGN_VALUE_OR_RETURN_ERR(channel, getChannel(dict));
-    auto *scaleV = G_.getParent()->createConstant("scale", *scale);
-    auto *biasV = G_.getParent()->createConstant("bias", *bias);
-    auto *meanV = G_.getParent()->createConstant("mean", *mean);
-    auto *varV = G_.getParent()->createConstant("var", *var);
+
+    Storage *scaleV, *biasV, *meanV, *varV;
+    if (bindings) {
+      auto *phScale =
```
SplitInfinity: Placeholder create code here seems very repetitive - consider refactoring.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+    timer.startTimer();
+    runBatch(EE, Bindings, ReportRate, SampleCounter, {A, E},
+             {&Images, &Labels});
+    timer.stopTimer();
+
+    llvm::outs() << "Evaluating.\n";
+    unsigned score = 0;
+    unsigned total = 0;
+
+    for (unsigned int i = 0; i < CIFAR_NUM_IMAGES / miniBatchSize; ++i) {
```
SplitInfinity: What happens if the dataset size is not a multiple of the minibatch size? I suggest validating the command-line parameters before starting training and reporting poor choice of minibatch to the user.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+      Tensor Sample(ElemKind::FloatTy, BatchImagesDims);
+      Sample.copyConsecutiveSlices(&Images, miniBatchSize * i);
+      updateInputPlaceholders(Bindings, {A}, {&Sample});
+      EE.run(Bindings);
+
+      for (unsigned int Batch = 0; Batch < miniBatchSize; ++Batch) {
```
SplitInfinity: I think `Batch` is a poor name. Consider `Ex`, `Example` or `MiniBatchPos` or something similar.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,20 @@ class Caffe2ModelLoader
   llvm::Error loadWeightsFromNet(caffe2::NetDef &net);
 
   /// Loads an individual weight \p op.
+  /// If Bindings are provided creates Placeholders vs Constants.
```
SplitInfinity: ???

This method does not have a bindings parameter.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,20 @@ class Caffe2ModelLoader
...
   llvm::Error loadWeight(const caffe2::OperatorDef &op);
 
   /// Load the structure of the network from the 'net' file.
-  llvm::Error loadNetwork(caffe2::NetDef &net);
+  /// If Bindings are provided creates Placeholders vs Constants.
```
SplitInfinity: If bindings are provided, weights are created as Placeholders. If not, they are created as Constants and their values are taken from the net weight file.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -51,14 +51,20 @@ class Caffe2ModelLoader
...
+  llvm::Error loadNetwork(caffe2::NetDef &net,
+                          PlaceholderBindings *bindings = nullptr);
 
   /// Load the operator \p op into the network. This creates one or more nodes
   /// in the network.
-  llvm::Error loadOperator(const caffe2::OperatorDef &op);
+  /// \p bindings is an optional parameters, if it is provided then training
```
SplitInfinity: `bindings is an optional parameters. If it is provided, then trainable weights will be added to the graph as Placeholders.`

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -95,7 +101,8 @@ class Caffe2ModelLoader
                     const std::string &netWeightFilename,
                     llvm::ArrayRef<const char *> names,
                     llvm::ArrayRef<TypeRef> types, Function &F,
-                    llvm::Error *errPtr = nullptr);
+                    llvm::Error *errPtr = nullptr,
+                    PlaceholderBindings *bindings = nullptr);
```
SplitInfinity: Add comment for `bindings` parameter here as well.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -101,6 +102,22 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
     RETURN_ERR("Only uint8 and int32 qtensors are supported");
   }
 }
+
+Storage *createPlaceHolderFromWeightTensor(
```
SplitInfinity: `createPlaceholderFromWeightTensor` (Placeholder not PlaceHolder)

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -101,6 +102,22 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
...
+    PlaceholderBindings *bindings, Function &f, Tensor *weights,
+    const std::string &name, bool moveOrCopy, bool isQuantized,
+    std::function<void(Tensor *)> &&initializer,
+    ElemKind type = ElemKind::FloatTy, bool isTrainable = true) {
+  auto *ph = isQuantized
```
SplitInfinity: Why do you need to create the Placeholder differently based on whether it's quantized or not? If you call `Placeholder *Module::createPlaceholder(TypeRef T, llvm::StringRef name, bool isTrainable)` with `&weights->getType()` as the parameter `T`, shouldn't the quantization information be encoded in the type?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -101,6 +102,22 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
     RETURN_ERR("Only uint8 and int32 qtensors are supported");
   }
 }
+
+Storage *createPlaceHolderFromWeightTensor(
+    PlaceholderBindings *bindings, Function &f, Tensor *weights,
+    const std::string &name, bool moveOrCopy, bool isQuantized,
```
SplitInfinity: Consider renaming `moveOrCopy` to `allowMove` or something like that. When `moveOrCopy` is `true`, it is not clear from the name what that means.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -300,54 +318,93 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
-        const auto &biasName = op.input(2);
-        biasConstant = getConstantByNameOrNull(biasName);
-      }
-
-      // If no serialized bias was found then create a zero bias.
-      if (!biasConstant) {
+      if (bindings) {
+        filter = createPlaceHolderFromWeightTensor(
+            bindings, G_, &W->getPayload(), "conv.filter", false /*move*/,
+            false /*quantized*/, [&kernels, &idim, this](Tensor *t) {
```
SplitInfinity: Do you need to capture `this`? Can you capture just the members you need (like `G_`)?

SplitInfinity: Use the exact same argument names in the inline comments as you did in the function declaration/definition (e.g. `isQuantized`).

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+#include <fstream>
+
+/// This is an example demonstrating how to load resnet50 model,
+/// randomize weights and perform training on a limited data set,
+/// and evaluate model.
+
+constexpr size_t IMAGE_COLORS = 3;
+constexpr size_t IMAGE_HEIGHT = 224;
+constexpr size_t IMAGE_WIDTH = 224;
+constexpr size_t LABELS_SIZE = 1000;
```
SplitInfinity: Changing the input size of ResNet is nontrivial so I think the scaling is a great idea. But is it possible to change the final FC layer so that it outputs only 10 classes?

lib/Importer/Caffe2ModelLoader.cpp
putivsky: Yes, createPlaceholder -> uniqueType(...) must receive scale and offset explicitly.

lib/Importer/Caffe2ModelLoader.cpp
putivsky: I can't capture directly G_ it's a member of class, but I can send a reference instead.

examples/training/resnet50/main.cpp
putivsky: This is just example (proof of concept), not a real program to train RestNet50, changing the last FC would looks like a hack.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,253 @@
...
+  // Allocate placeholders without tensors.
+  Bindings.allocate(Mod.getPlaceholders());
+
+  // Get output tensor.
+  auto *ResultPH = Mod.getPlaceholderByName("save_gpu_0_softmax");
+  auto *Result = Bindings.get(ResultPH);
+
+  llvm::outs() << "Loading the CIFAR-10 database.\n";
+  /// ResNet expects ImageChannelOrder::BGR, ImageLayout::NCHW
+  std::ifstream dbInput("cifar-10-batches-bin/data_batch_1.bin",
```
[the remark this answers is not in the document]
putivsky: There are plenty assumptions in the code about CIFAR images format and number of images in the batch, if caller will provide an arbitrary file it can easily breaks the internal logic.

examples/training/resnet50/main.cpp
putivsky: Rounding integer will prevent out of boundary problem.
If  CIFAR_NUM_IMAGES  % miniBatchSize != 0 then not all  CIFAR_NUM_IMAGES will be evaluated.
If  CIFAR_NUM_IMAGES  < miniBatchSize then no evaluation would occur.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -110,6 +111,22 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
     RETURN_ERR("Only uint8 and int32 qtensors are supported");
   }
 }
+
+Storage *createPlaceholderFromWeightTensor(
+    PlaceholderBindings *bindings, Function &f, Tensor &weights,
+    const std::string &name, bool allowMove, bool isQuantized,
```
jfix71: You should be able to remove `isQuantized` and instead just use the helper `isQuantizedElemKind(type)`

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -110,6 +111,22 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
     RETURN_ERR("Only uint8 and int32 qtensors are supported");
   }
 }
+
+Storage *createPlaceholderFromWeightTensor(
```
jfix71: This always returns a `Placeholder`, I'd change to that instead of `Storage`.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -91,11 +96,14 @@ class Caffe2ModelLoader
   /// outputs with specific names and types.
   /// If \p errPtr is not null then if an error occurs it will get assigned
   /// there otherwise if an error occurs it will abort.
+  /// \p bindings is an optional parameters. If it is provided, then trainable
+  /// weights will be added to the graph as Placeholders.
```
jfix71: Please change this comment to say that the loaded weights will be discarded, and are replaced by trainable weights that are initialized for training.

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -71,12 +71,16 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
...
-  case Kinded::Kind::TransposeNodeKind:
   case Kinded::Kind::SliceNodeKind:
   case Kinded::Kind::SpaceToDepthNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
         {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int64ITy});
 
+  case Kinded::Kind::TransposeNodeKind:
+    return NI.allInputsAndOutputsHaveSameElemKind(
+        {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int64ITy,
+         ElemKind::BoolTy});
```
jfix71: Can you add an OperatorTest unit test that checks this new Transpose with `BoolTy`?

lib/Importer/Caffe2ModelLoader.cpp
jfix71: How are you deciding when to use `allowMove`?  In what circumstances do we want to do that? It looks a bit arbitrary to me when you have decided to move vs. not, e.g. when loading a Conv you do it for bias but not filter.

putivsky: locally created tensors can be moved.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,251 @@
...
+                                  llvm::cl::init(0.001),
+                                  llvm::cl::value_desc("N"),
+                                  llvm::cl::cat(category));
+llvm::cl::opt<float> momentum("momentum", llvm::cl::desc("Momentum parameter."),
+                              llvm::cl::init(0.9), llvm::cl::value_desc("N"),
+                              llvm::cl::cat(category));
+llvm::cl::opt<float> l2Decay("l2-decay", llvm::cl::desc("L2Decay parameter."),
+                             llvm::cl::init(0.0001), llvm::cl::value_desc("N"),
+                             llvm::cl::cat(category));
+llvm::cl::opt<BackendKind> executionBackend(
```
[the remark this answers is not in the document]
putivsky: Logically Loader and Training are decoupled, it is cleaner to provide backend explicitly.

putivsky:
artemrakhov, training takes a while even for 100 images. Precision is not good as well. This PR is not for producing any meaningful results for resnet50 training but rather to demonstrate the training support by Glow for resnet50 in principal.

tests/unittests/OperatorTest.cpp
```diff
@@ -1864,6 +1864,25 @@ TEST_P(OperatorTest, FP16Transpose2Dims) {
   EXPECT_TRUE(bindings_.get(result->getPlaceholder())->isEqual(dest));
 }
 
+/// Check that transpose is supported for BoolTy.
+TEST_P(OperatorTest, BoolTranspose2Dims) {
+  ENABLED_BACKENDS(Interpreter);
```
jfix71: You added support for the CPU, so enable CPU here too.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+    auto val = handle.raw(c);
+    if (maxVal <= val) {
+      maxVal = val;
+      ret = c;
+    }
+  }
+  return ret;
+}
+
+void loadImagesAndLables(Tensor &images, Tensor &labels) {
```
jfix71: nit: Misspelled "Labels": `loadImagesAndLables -> loadImagesAndLabels`

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  // These tensors allocate memory for all images and labels prepared for
+  // training.
+  Tensor images(ElemKind::FloatTy, allImagesDims);
+  Tensor labels(ElemKind::FloatTy, allLabelsDims);
+
+  loadImagesAndLables(images, labels);
+
+  // From http://www.cs.toronto.edu/~kriz/cifar.html.
+  const char *textuallabels[LABELS_SIZE] = {
```
jfix71: nit: `textualLabels` (capitalize "labels")

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+  llvm::outs() << "Compiling backend.\n";
+  BackendKind executionBackend{executionBackendKind};
+  ExecutionEngine EE(executionBackend);
+  EE.compile(CompilationMode::Train, TF);
+
+  // Allocate placeholders without tensors.
+  bindings.allocate(mod.getPlaceholders());
+
+  // Get output tensor.
+  auto *ResultPH = mod.getPlaceholderByName("save_gpu_0_softmax");
```
jfix71: Lower camel case please.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+  auto *Result = bindings.get(ResultPH);
```
jfix71: Lower camel case please.

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+    runBatch(EE, bindings, reportRate, sampleCounter, {A, E},
+             {&images, &labels});
+    timer.stopTimer();
+
+    llvm::outs() << "Evaluating.\n";
+    unsigned score = 0;
+    unsigned total = 0;
+
+    for (unsigned int i = 0; i < CIFAR_NUM_IMAGES / miniBatchSize; ++i) {
+      Tensor Sample(ElemKind::FloatTy, batchImagesDims);
```
jfix71: Lower camel case please.

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -91,11 +97,15 @@ class Caffe2ModelLoader
   /// outputs with specific names and types.
   /// If \p errPtr is not null then if an error occurs it will get assigned
   /// there otherwise if an error occurs it will abort.
+  /// \p bindings is an optional parameters. If it is provided,
```
jfix71: nit: `parameters -> parameter` (remove "s")

putivsky:
jfix71, I have been thinking about replacing Constants with Placeholders from the beginning.
However there are cases when placeholders should be created as non-trainable. For instance, SpatialBN operator gradient creates an internal Placeholders for mean and variance and as a result if Loader create another ones placeholders end-up to have two references each and verification got failed. My point is it would be a situation when not all Constants would be replaceable by trainable Placeholders. It would be hard to come up with a generic solution having such exceptions.

artemrakhov:
I agree with Jordan on the design. We would have to change ONNX loader as well, and also have to think about it in every new loader that we'll have to support in the future. It would be great to have a separate optimization pass, to mark which constants are trainable.

What do you all think about the concept of *Training Config*? It could contain a list of node names that we are going to train, and also meta parameters such as learning rate and batch size. It is common for ML frameworks to have a separate config alongside with the model, which specifies how to train.

We can have a separate (standalone) util that would take a ResNet model protobuf and produce a training config for it.

As for accuracy: one problem that I see is scaling inputs from 32x32 to 224x224. It does not work, because all traits (e.g. animal eye, or human arm) become much bigger than what model was built to detect. You would have to increase kernel/filter size, at least. It is much more preferred to do padding, but I'm not sure how it would work when pad (empty space) is so big.

The PR is solving 2 problems:
1. Train ResNet
2. Introduce infrastructure to train protobuf

I think (1) is already very hard by itself. My understanding is that it doesn't work now, which is not good for testing purposes.
Is there a way to demonstrate (2) by training something simple? Maybe porting existing mnist test? @jfix71, did you train some simple mnist models using Caffe2?

putivsky:
Other thoughts I have in mind. Ideally we shouldn't have any weights provided for the loader at the first stage, just weights shapes.
Then I would create a model graph parsing operators and weights shapes creating Placeholders and allocating Tensors memory on a fly.
Then depending on the loader mode (inference vs training):
- in case of training mode, either randomize placeholder tensors values or fill out from provided weights values (pre-trained weights)
- in case of inference mode, convert Placeholders into Constants and copy weights into Constants payloads.
Let me know if this is something you consider as a plan, and I will abound this PR and work on refactoring Caffe2/ONNX loaders.

jfix71:
>Ideally we shouldn't have any weights provided for the loader at the first stage, just weights shapes.

We still need to load the proto entirely into memory, including the weights, so it doesn't seem much of a benefit to me, as all we would end up skipping is setting the pointer of the Constant's payload Tensor, right? I'm still unclear on the downside of my suggestion to do:

```
Caffe2ModelLoader (unchanged, from before this PR)
Constant->PH Conversion pass
Differentiate
```

>@jfix71, did you train some simple mnist models using Caffe2?

Yeah, I trained the `lenet_mnist` model on my laptop, IIRC it took just a couple hours. I used `utils/caffe2_train_and_dump_pb.py` to do it. The script dumps the trained `lenet_mnist` protos, which is what is downloaded via `utils/download_datasets_and_models.py`, as `lenet_mnist`, and run in our `run.sh` script. So using these protos could be a good starting point, as you can verify that training is working correctly in a reasonable amount of time.

artemrakhov:
> Ideally we shouldn't have any weights provided for the loader at the first stage, just weights shapes.

If we only know the shape, how will we init the weights? It is very important to init properly. I see that now initialization code is in the loader, meaning that it will be duplicated for ONNX loader and all future loaders. And this "xavier" algorithm is not set in stone. Probably there is already something better out there. Meaning that we should make this configurable by the user. TrainingConfig?

putivsky:
Loading already trained and serialized models (proto) doesn't make a lot of sense for training, because models has been trained already.
We need to answer the question: how will Glow users train new models. Writing C++ code in Glow API is one of the approaches. But most ML researches will use Caffe2/PyTorch to design models. We can plugin-in Glow C++ code inside those frameworks or use serialized models (not trained yet, without a meaningful weights). Let me schedule a meeting next week for a broader discussion.

SplitInfinity:
I think the `TrainingConfig` class is too simple. I created issue #2554 a few weeks ago to solicit feedback on how to generalize it to support more training algorithms.

As @artemrakhov mentioned, ML frameworks have configurable `Trainer` or `Optimizer` objects that are responsible for performing weight updates. If that is the functionality we need, `TrainingConfig` is too simple. As for the question of whether we need it, the answer depends on whether we envision Glow being capable of standalone training, or only as part of a backend for PyTorch. In the second case, we can leave all those details to PyTorch in the long term and do whatever we need to in the short term to plug those holes in order to work on the rest of the stack.

putivsky:
Taking another approach for refactoring.

artemrakhov:
@SplitInfinity: what I meant by TrainingConfig is creating another, much more general class which will describe how we are going to train. It should specify which weights are we going to train, and what are init parameters, and many other things.
I actually forgot that we already have `TrainingConfig` struct in Glow, which in my opinion needs to be renamed, as it is very algorithm specific. What if future training approach will use something other than SGD, and not even have L2Decay?

SplitInfinity:
Yes, I agree. But I would just like to point out that PyTorch has its own Optimizers and we should think carefully before we decide to redo all of the work to reimplement training algorithms (and add new ones as they are implemented) in Glow for this work.

artemrakhov:
This is good question. Can PyTorch differentiate the graph for us? Does it have a lot of Optimizers that play after we freeze the graph?

SplitInfinity:
PyTorch can differentiate for us. 

Looks like PyTorch has maybe [5-10 optimisers](https://pytorch.org/docs/stable/optim.html). They are not yet represented in the Pytorch JIT IR, but they are plans to add this.

PR #2894, opened 2019-05-13
Adding RecommendationSystemTest for SparseLengthSum based models

nickgg:
*Description*: This adds a small unit test which provides coverage and an example of a model implementing a simplified recommendation system. This RecSys model consists of two Multilayer Perceptrons (MLP), a set of sparse features (provided via SparseLengthsSum) and an interaction layer that combines the bottom MLP and the sparse features. All data, both inputs and weights, is randomized. We provide both floating point and quantized version of each component, as well as tests which exercise partitioning.
*Testing*: Ran the new RecommendationSystemTest code in various modes.
*Documentation*: **Warning** this test is marked `EXPENSIVE` but **will run** in `ninja test`. It does not run in `ninja check`.

### DIFF
--- a/tests/unittests/RecommendationSystemTest.cpp
+++ b/tests/unittests/RecommendationSystemTest.cpp
@@ -1,0 +1,667 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "BackendTestUtils.h"
+
+#include "glow/Converter/TypeAToTypeBFunctionConverter.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Partitioner/Partitioner.h"
+
+#include "gtest/gtest.h"
+
+using namespace glow;
+
+/// Fills the tensor \p H with some stable random data with the seed \p seed
+/// and the range [-scale .. scale].
+static void fillStableRandomData(Handle<float> H, size_t seed,
+                                 float scale = 1) {
+  for (size_t i = 0, e = H.size(); i < e; i++) {
+    H.raw(i) = scale * (float((int(i * 1921 + seed) % 100) - 50) / 50);
+  }
+}
+
+// Fills the tensor \p H with some stable random integers with the seed \p
+// seed
+/// and the range [0, scale).
+template <typename T>
+static void fillStableRandomIndex(Handle<T> H, size_t seed, size_t min = 0,
+                                  size_t max = 10) {
+  for (size_t i = 0, e = H.size(); i < e; i++) {
+    H.raw(i) = min + (int(i * 1921 + seed) % (max - min));
+  }
+}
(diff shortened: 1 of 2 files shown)


tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -0,0 +1,678 @@
...
+        endPt++;
+      }
+      curPt++;
+    }
+  }
+
+  /// Create partitions to run and compare results.
+  void runPartitionedGraph(size_t numDevices, size_t memSize,
+                           Tensor referenceResult,
+                           PlaceholderBindings &bindings) {
```
beicy: Can we apply HostManager to do Runtime End2End test directly?

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -0,0 +1,678 @@
...
+#include "BackendTestUtils.h"
+
+#include "glow/Converter/TypeAToTypeBFunctionConverter.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Partitioner/Partitioner.h"
+
+#include "gtest/gtest.h"
+
+// To enable production appropriate sizes, enable this:
```
bertmaher: Stale comment?

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -0,0 +1,667 @@
...
+///   * quantizeFC    enables Int8 Fused Rowwise Quantization for FC weights and
+///     activations inside the MLPs.
+///   * convertToFP16 walks the graph at the end of constructing the graph and
+///     converts all FP32 nodes & tensors to FP16, meaning the graph will use
+///     FP16 for internal weights, biases and activations (when not already Int8
+///     quantized). Inputs and outputs are still FP32 but are immediately
+///     dropped to FP16 precision at the beginning of the graph.
+
+TEST_P(RecommendationSystemTest, RecSys_FP32) {
+  ENABLED_BACKENDS(CPU);
```
rdzhabarov: Can you also enable tests on Habana (those that are passing now).

tests/unittests/RecommendationSystemTest.cpp
nickgg: Yes we should. Right now its blocked on #2874 and some additional plumbing so we can set device sizes and force the partitioning, I've created #2897 for this.

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+  deviceMemCapacity *= 2; // Double memory for this test
+
+  runPartitionedGraph(numDevices, deviceMemCapacity,
+                      bindings_.get(result)->clone(), bindings_);
+}
+
+TEST_P(RecommendationSystemTest, RecSys_Partitioned_RWQuantized_SLWS) {
+  ENABLED_BACKENDS(CPU);
+  ENABLED_BACKENDS(Habana);
```
rdzhabarov: nit: ENABLED_BACKENDS(CPU, Habana)

here and in other places.

tests/unittests/RecommendationSystemTest.cpp
rdzhabarov: this should be on Habana as well.

nickgg: It's literally the line below 😅

jfix71:
It looks like we're no longer checking the results of the non-partitioned versions -- is this intentional?

nickgg:
@jfix71 yes, I don't understand the results well enough to say what a good test is and I haven't got a strong answer yet. We should develop this.

tests/unittests/RecommendationSystemTest.cpp
rdzhabarov: Hm, ok, might miss it.

rdzhabarov:
@jfix71 previous check against `< 1` was originally introduced because of sigmoid op presence at the end.
It was not clear whether there is anything else forcing result to be < 1.

PR #2895, opened 2019-05-13
Change caffe2 loader to multi Qtensor proto version

zrphercule:
Reviewed By: yinghai

Differential Revision: D15115749

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -137,6 +137,14 @@ loadWeight(const onnxTensorDescriptorV1 &in) {
       for (size_t i = 0; i < TH.size(); ++i) {
         TH.raw(i) = data[i];
       }
+    } else if (in.dataType == ONNXIFI_DATATYPE_INT8) {
+      result.t->reset(ElemKind::Int8QTy, dims, in.scales[0], in.biases[0]);
+
+      auto TH = result.t->getHandle<int8_t>();
+      int8_t *data = (int8_t *)in.buffer;
+      for (size_t i = 0; i < TH.size(); ++i) {
+        TH.raw(i) = data[i];
+      }
     } else {
       RETURN_ERR("Only uint8 and int32 quantized tensors are supported.");
     }
--- a/lib/Importer/caffe2.proto
+++ b/lib/Importer/caffe2.proto
@@ -126,6 +126,17 @@ message QTensorProto {
   repeated int32 data = 6 [packed = true];
   optional string name = 7;
   optional TensorProto.DataType data_type = 8 [default = INT32];
+
+  // Multi-group quantization params
+  repeated double scales = 9;
+  repeated double biases = 10;
+
+  // Multi-group quantization needed, indicates in which dimension
+  // we do the "group wise quantization"
+  optional int32 axis = 11;
+
+  // It should be true if it is a multi-group quantization proto
+  optional bool is_multiparam = 12 [default = false];
 }
 
 // TensorProtos stores multiple TensorProto objects in one single proto. This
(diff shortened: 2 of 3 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -137,7 +137,16 @@ loadWeight(const onnxTensorDescriptorV1 &in) {
...
       }
-    } else {
+    } else if (in.dataType == ONNXIFI_DATATYPE_INT8) {
+      result.t->reset(ElemKind::Int8QTy, dims, in.scales[0],
+                      in.biases[0]);
+
+      auto TH = result.t->getHandle<int8_t>();
+      int8_t *data = (int8_t *)in.buffer;
+      for (size_t i = 0; i < TH.size(); ++i) {
+        TH.raw(i) = (int8_t)(data[i]);
```
beicy: Line 145, data is int8_t. Do we still need cast here?

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -137,7 +137,16 @@ loadWeight(const onnxTensorDescriptorV1 &in) {
       for (size_t i = 0; i < TH.size(); ++i) {
         TH.raw(i) = data[i];
       }
-    } else {
+    } else if (in.dataType == ONNXIFI_DATATYPE_INT8) {
```
rdzhabarov: need clang format.

rdzhabarov:
It does not compile ;)

lib/Importer/caffe2.proto
```diff
@@ -126,6 +126,10 @@ message QTensorProto {
   repeated int32 data = 6 [packed = true];
   optional string name = 7;
   optional TensorProto.DataType data_type = 8 [default = INT32];
+  repeated double scales = 9;
```
jackm321: Can we remove `scale` and `bias` now?

jackm321:
@rdzhabarov I'm going to put up the diff for importing `ChannelwiseQuantizedConv` later today (this PR was tested with it already) and that uses these

rdzhabarov:
> @rdzhabarov I'm going to put up the diff for importing ChannelwiseQuantizedConv later today (this PR was tested with it already) and that uses these

Sounds good! I wanted to see this added logic tested.

lib/Importer/caffe2.proto
zrphercule: @jackm321  No we cant... Remove things from proto is not a good idea since we dont know who  else is using it.

lib/Importer/caffe2.proto
```diff
@@ -126,6 +126,10 @@ message QTensorProto {
...
+  repeated double biases = 10;
```
rdzhabarov: Is there going to be a separate data type in C2 proto which would signify multiple scales/offsets?

lib/Importer/caffe2.proto
```diff
@@ -126,6 +126,10 @@ message QTensorProto {
...
+  optional int32 axis = 11;
+  optional bool is_multiparam = 12 [default = false];
```
rdzhabarov:
```
is_multiparam
```
I assume it refers to multiple scales/offsets? Can it have more explicit name)?

zrphercule: Yeah, any suggestions? since this is QTensorProto already, so multiparam cannot indicate anything else...

jackm321:
@zrphercule this will conflict with #2921. If that's accepted, I'll make sure not to land it until this one is in.

zrphercule:
> @zrphercule this will conflict with #2921. If that's accepted, I'll make sure not to land it until this one is in.

Thanks, no problem, target mergeing this today.

lib/Importer/caffe2.proto
zrphercule: Guess I will just use this name... This is in QTensorProto already, so multiparam cannot mean anything else other than the quantization param...

zrphercule: I think I can add some comments here in case someone forgot what it is

Issue #2358, opened 2019-02-06
Add version of readPngImage... that writes into an existing Tensor


nickgg:
The Image reading functions (lib/Base/Image.h) either take a Tensor* and reset it to the dimensions of the image, or in the case of readPngImageAndPreprocess create a new Tensor. This means that when loading a batch of images that should go into the same Tensor we do unnecessary copies.

Write a version of readImage (andPreprocess?) which takes a four dimensional tensor (w/ an index) and writes into it. The input tensor must have the same dimensions as the image.

PR #2896, opened 2019-05-13
[image]: Add a version for readPngImageAndPreprocess that can also take a tensor as input.

ZchiPitt:
*Description*:
Implement a version for readPngImageAndPreprocess that can also take a tensor as input.
*Testing*:
Add a test case in ImageTest: readPngImageAndPreprocessWithAndWithoutInputTensor() that verifies the new version works correctly.
[Optional Fixes #issue]
#2358

### DIFF
--- a/include/glow/Base/Image.h
+++ b/include/glow/Base/Image.h
@@ -93,7 +93,8 @@ bool writePngImage(Tensor *T, const char *filename,
                    llvm::ArrayRef<float> mean = zeroMean,
                    llvm::ArrayRef<float> stddev = oneStd);
 
-/// Read a png image and preprocess it according to several parameters.
+/// Read a png image and preprocess it according to several parameters. Create a
+/// tensor and store the preprocessed image data into this tensor.
 /// \param filename the png file to read.
 /// \param imageNormMode normalize values to this range.
 /// \param imageChannelOrder the order of color channels.
@@ -106,6 +107,24 @@ Tensor readPngImageAndPreprocess(llvm::StringRef filename,
                                  ImageLayout imageLayout,
                                  llvm::ArrayRef<float> mean = zeroMean,
                                  llvm::ArrayRef<float> stddev = oneStd);
+
+/// Read a png image and preprocess it according to several parameters. Take a
+/// tensor as a parameter and store the preprocessed image data into this
+/// tensor.
+/// \param imageData the tensor into which the preprocessed image data
+///  will be stored.
+/// \param filename the png file to read.
+/// \param imageNormMode normalize values to this range.
+/// \param imageChannelOrder the order of color channels.
+/// \param imageLayout the order of dimensions (channel, height, and width).
+/// \param mean use special mean to normalize.
+/// \param stdev use special stddev to normalize.
+void readPngImageAndPreprocess(Tensor &imageData, llvm::StringRef filename,
+                               ImageNormalizationMode imageNormMode,
+                               ImageChannelOrder imageChannelOrder,
+                               ImageLayout imageLayout,
+                               llvm::ArrayRef<float> mean = zeroMean,
+                               llvm::ArrayRef<float> stddev = oneStd);
 
 /// Loads and normalizes all PNGs into a tensor in the NHWC format with the
 /// requested channel ordering.
--- a/lib/Base/Image.cpp
+++ b/lib/Base/Image.cpp
@@ -362,6 +362,19 @@ Tensor glow::readPngImageAndPreprocess(llvm::StringRef filename,
                                        llvm::ArrayRef<float> mean,
                                        llvm::ArrayRef<float> stddev) {
   Tensor imageData;
+  readPngImageAndPreprocess(imageData, filename, imageNormMode,
+                            imageChannelOrder, imageLayout, mean, stddev);
+  return imageData;
+}
+
+void glow::readPngImageAndPreprocess(Tensor &imageData,
+                                     llvm::StringRef filename,
+                                     ImageNormalizationMode imageNormMode,
+                                     ImageChannelOrder imageChannelOrder,
+                                     ImageLayout imageLayout,
+                                     llvm::ArrayRef<float> mean,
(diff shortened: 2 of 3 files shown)


nickgg:
Added #2898 to cover porting the resnet example code to this new loader function.

lib/Base/Image.cpp
```diff
@@ -391,9 +391,32 @@ Tensor glow::readPngImageAndPreprocess(llvm::StringRef filename,
...
+                                       ImageLayout imageLayout,
+                                       llvm::ArrayRef<float> mean,
+                                       llvm::ArrayRef<float> stddev) {
+  Tensor imageData;
+  readPngImageAndPreprocessHelper(imageData, filename, imageNormMode,
+                                  imageChannelOrder, imageLayout, mean, stddev);
   return imageData;
 }
 
+void glow::readPngImageAndPreprocess(Tensor &imageData,
```
jfix71: Given that this version of `readPngImageAndPreprocess()` is just forwarding to `readPngImageAndPreprocessHelper()` without any extra logic, I think it's preferable to avoid adding `readPngImageAndPreprocessHelper()`. You can just call this version of `readPngImageAndPreprocess()` (which takes a Tensor reference) from the other version which returns the Tensor.

tests/unittests/ImageTest.cpp
```diff
@@ -55,6 +55,39 @@ TEST(Image, readBadImages) {
...
+
+  // Test if the preprocess actually happened.
+  size_t imgHeight = image1.dims()[0];
+  size_t imgWidth = image1.dims()[1];
+  size_t numChannels = image1.dims()[2];
+
+  Tensor swizzled(image1.getType());
+  auto IH = image1.getHandle();
+  auto SH = swizzled.getHandle();
+  for (unsigned z = 0; z < numChannels; z++) {
```
SplitInfinity: Nit: Counters are `unsigned` but bounds are `size_t`. Since the loop counter variables are used to index into Tensors they should probably be `size_t` as well.

include/glow/Base/Image.h
```diff
@@ -93,20 +93,51 @@ bool writePngImage(Tensor *T, const char *filename,
...
                    llvm::ArrayRef<float> stddev = oneStd);
 
-/// Read a png image and preprocess it according to several parameters.
-/// \param filename the png file to read.
-/// \param imageNormMode normalize values to this range.
-/// \param imageChannelOrder the order of color channels.
+/// A helper function called by readPngImageAndPreprocess that does all the
+/// reading file and preprocessing work.
+/// \param imageData the tensor into which the preprocessed image data will be
+/// stored. \param filename the png file to read. \param imageNormMode normalize
```
opti-mix: Bad formatting of the comment. Each `\param` should be at the beginning of a line.

include/glow/Base/Image.h
```diff
@@ -93,20 +93,51 @@ bool writePngImage(Tensor *T, const char *filename,
...
+void readPngImageAndPreprocessHelper(Tensor &imageData,
+                                     llvm::StringRef filename,
+                                     ImageNormalizationMode imageNormMode,
+                                     ImageChannelOrder imageChannelOrder,
+                                     ImageLayout imageLayout,
+                                     llvm::ArrayRef<float> mean = zeroMean,
+                                     llvm::ArrayRef<float> stddev = oneStd);
+
+/// Read a png image and preprocess it according to several parameters. Create a
+/// tensor and store the preprocessed image data into this tensor. \param
```
opti-mix: Bad formatting of the comment. Each `\param` should be at the beginning of a line.

include/glow/Base/Image.h
```diff
@@ -93,20 +93,51 @@ bool writePngImage(Tensor *T, const char *filename,
...
 Tensor readPngImageAndPreprocess(llvm::StringRef filename,
                                  ImageNormalizationMode imageNormMode,
                                  ImageChannelOrder imageChannelOrder,
                                  ImageLayout imageLayout,
                                  llvm::ArrayRef<float> mean = zeroMean,
                                  llvm::ArrayRef<float> stddev = oneStd);
 
+/// Read a png image and preprocess it according to several parameters. Take a
+/// tensor as a parameter and store the preprocessed image data into this
+/// tensor. \param imageData the tensor into which the preprocessed image data
```
opti-mix: Bad formatting of the comment. Each `\param` should be at the beginning of a line.

tests/unittests/ImageTest.cpp
```diff
@@ -55,6 +55,39 @@ TEST(Image, readBadImages) {
...
+                            ImageNormalizationMode::k0to1,
+                            ImageChannelOrder::BGR, ImageLayout::NCHW,
+                            imagenetNormMean, imagenetNormStd);
+
+  // Test if the preprocess actually happened.
+  size_t imgHeight = image1.dims()[0];
+  size_t imgWidth = image1.dims()[1];
+  size_t numChannels = image1.dims()[2];
+
+  Tensor swizzled(image1.getType());
```
SplitInfinity: This code code seems to be identical to the preprocessing code in `glow::readPngImageAndPreprocessHelper`. I'm not sure what the value of this is because a bug in both places won't get caught and a bug that is introduced into `readPngImageAndPreprocess` is likely to be hidden by copying that code here whenever it is updated.

lib/Base/Image.cpp
ZchiPitt: sounds good

ZchiPitt:
> Changes look good.
> 
> Nit: The doxygen comments are easier to read when each \param is on a new line. Can we change that?

sure. It must have been because of the auto format script lol

tests/unittests/ImageTest.cpp
ZchiPitt: sounds right. I modified the test case that I move the transpose operation on image2 instead to make image2 become NHWC and do the RBG change to image1, then i can check whether image1 equals image2. 

It should be same logic but the operations are not same as the `readPngImageAndPreprocess()` implementation

PR #2900, opened 2019-05-13
Make debug printing friendly to gcc 4.8.5

bertmaher:
Summary:
Use a unique_ptr to array rather than std::vector since older gcc
doesn't know how to handle a vector of arrays.

Reviewed By: rdzhabarov

Differential Revision: D15322072

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -420,22 +420,23 @@ void dumpTopologyInfo(uint32_t deviceId, uint64_t topologyId) {
                             numOfIntermediates));
 
   using TensorNames = char[ENQUEUE_TENSOR_NAME_MAX_SIZE];
-  std::vector<TensorNames> inputTensorNames(numOfInputs);
-  std::vector<TensorNames> outputTensorNames(numOfOutputs);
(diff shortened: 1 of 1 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -420,22 +420,23 @@ void dumpTopologyInfo(uint32_t deviceId, uint64_t topologyId) {
...
-  chk(synGetTensorsName(deviceId, topologyId, inputTensorNames.data(),
-                        numOfInputs, outputTensorNames.data(), numOfOutputs,
-                        intermediateTensorNames.data(), numOfIntermediates));
+  chk(synGetTensorsName(deviceId, topologyId, inputTensorNames.get(),
+                        numOfInputs, outputTensorNames.get(), numOfOutputs,
+                        intermediateTensorNames.get(), numOfIntermediates));
 
-  for (auto const &in : inputTensorNames) {
-    VLOG(1) << "Topology input: " << in;
+  for (size_t i = 0; i < numOfInputs; i++) {
```
rdzhabarov: nit: s/size_t/uint32_t

bertmaher:
I don't know what went wrong but this was merged as 0456859a86c32cdf03b4ca6c95d1042b30183578

PR #2908, opened 2019-05-14
[OpenCL] Add 8 and 16 element vectorized select kernel

### LINKED ISSUE #2835: [OpenCL] mnist test application fails (cannot find elementselectW16 kernel)
With pocl:
```
$ POCL_DEBUG=1 bin/mnist -platform=0 -device=0 -opencl
Loading the mnist database.
...
[2019-04-01 15:23:40.903036977]POCL: in fn POclCreateKernel at line 80:
  |     ERROR | CL_INVALID_KERNEL_NAME Can't find a the kernel with name elementselectW16 in this program
../lib/Backends/OpenCL/OpenCL.cpp:168: failed assertion `kernel && "clCreateKernel Failed."'
Aborted (core dumped)
```

Tested also with the Intel's CPU driver:
```
CL_ICD_VENDORS=/etc/OpenCL/vendors/intel_exp64.icd bin/mnist -platform=0 -device=0 -opencl
Loading the mnist database.
Loaded 50000 images.
Training.
Training - epoch #0
../lib/Backends/OpenCL/OpenCL.cpp:168: failed assertion `kernel && "clCreateKernel Failed."'
Aborted (core dumped)
```

SplitInfinity:
**Description**
This commit adds an 8- and 16-element vectorized elementselect kernel
for the OpenCL backend. It is implemented using the OpenCL `select(x, y,
z)` vector function that does the same thing.

**Testing**
Ran `mnist` example on the OpenCL backend and it is no longer broken:
```
$ ../build_DEBUG/bin/mnist -platform=0 -device=0 -opencl
Loading the mnist database.
...
-------------
Results: guessed/total:74/80
===-------------------------------------------------------------------------===
                         Miscellaneous Ungrouped Timers
===-------------------------------------------------------------------------===

  381.9755 (100.0%)   4.4327 (100.0%)  386.4082 (100.0%)  67.1256 (100.0%)  Training
  381.9755 (100.0%)   4.4327 (100.0%)  386.4082 (100.0%)  67.1256 (100.0%)  Total
```

**Fixes**
This PR addresses part of #2835.

### DIFF
--- a/lib/Backends/OpenCL/kernels.cl
+++ b/lib/Backends/OpenCL/kernels.cl
@@ -200,6 +200,14 @@ __kernel void dequantizeW(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
 /// \p name the name of the kernel
 /// \p type the type of the tensor elements and of the return value
 /// \p body the operation to be performed
+
+/// The K16 kernel uses 8-element vector primitives on two contiguous 8-element
+/// blocks of data (blocks i and i+1) to produce a 16-element vectorized
+/// kernel. Consequently, the global ID needs to be multiplied by 2 so that
+/// each kernel invocation will load/store blocks that are not loaded/stored by
+/// other kernels. If this is not done, the kernel with global ID 0 will
+/// load/store blocks 0 and 1, the kernel with global ID 1 will load/store on
+/// blocks 1 and 2, etc.
 #define DEFINE_OPENCL_TERNARY_DATA_PARALLEL_KERNEL(name, type, body)           \
   __kernel void name##K##16(__global type * dest, __global type * cond,        \
                             __global type * lhs, __global type * rhs) {        \
@@ -260,6 +268,14 @@ __kernel void dequantizeW(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
 /// \p name the name of the kernel
 /// \p type the type of the tensor elements and of the return value
 /// \p body the operation to be performed
+
+/// The K16 kernel uses 8-element vector primitives on two contiguous 8-element
+/// blocks of data (blocks i and i+1) to produce a 16-element vectorized
+/// kernel. Consequently, the global ID needs to be multiplied by 2 so that
+/// each kernel invocation will load/store blocks that are not loaded/stored by
+/// other kernels. If this is not done, the kernel with global ID 0 will
+/// load/store blocks 0 and 1, the kernel with global ID 1 will load/store on
+/// blocks 1 and 2, etc.
 #define DEFINE_OPENCL_BINARY_DATA_PARALLEL_KERNEL(name, type, body)            \
   __kernel void name##K##16(__global type * dest, __global type * lhs,         \
                             __global type * rhs) {                             \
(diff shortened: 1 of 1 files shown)


SplitInfinity:
I am not quite sure how to debug the 4 misclassified images. I tried to write naive, unvectorized versions of these kernels using the logic from the normal one that process 8/16 elements as appropriate and that actually lowered the accuracy.

I think the problem might be related to the semantics of the `select` OpenCL function:

```
For each component of a vector type,

result[i] = if MSB of c[i] is set ? b[i] : a[i].
```

Depending on how the predicate input of the select operation was produced, I think the LSB might be the one that indicates `true/false` not the MSB. Any suggestions on how to reverse the predicates to test this? Or to debug this in general?

SplitInfinity:
@opti-mix 

After changing the predicate to `isnotequal(convert_int8(vload8(i, cond), zeroVec)` the accuracy is unchanged. (72 images classified correctly).

opti-mix:
> Or to debug this in general?

OpenCL backend supports `--instrument-debug` option (or `--instrument-debug-only`) and would dump the values of all inputs and outputs of each instruction. This could help to compare it with e.g. the Interpreter running with the same `--instrument-debug`  option.

Also usual `printf` debugging should work.

SplitInfinity:
Hmm, I ran this with vectorization temporarily disabled for `elementselect` and that resulted in 73 images classified correctly (sometimes 74). So there is still a one image difference to explain, but it looks like there is another issue that can cause accuracy differences between OpenCL and CPU backends.

SplitInfinity:
I found a small bug in the 16-element vectorized implementation of `elementselect` and now the accuracy is the same as the OpenCL version that uses the non-vectorized version of `elementselect` by default. The global ID in `elementselectK16` has to be multiplied by 2 to avoid overlapping loads/stores. I think that is enough for this PR.

We separately need to investigate and address why this example sometimes fails on the OpenCL backend (73 correct answers instead of 74) but passes on the CPU backend (74 correct).

SplitInfinity:
Same vectorization bug exists for `elementcmplteK16`. I fixed that and the number of correctly classified images is consistently 74 now.

opti-mix:
> Same vectorization bug exists for elementcmplteK16. I fixed that and the number of correctly classified images is consistently 74 now.

Awesome! Great job finding the problem!

BTW, out of curiosity, could you check if it is possible to use the `DEFINE_OPENCL_TERNARY_DATA_PARALLEL_KERNEL` for these kernels? Doing so would make them more similar to other kernels and these macros also do not have the issue that you found.

jfix71:
Nice!

I'm noticing that we have very few comments in `kernels.cl`. For example, the fix (`size_t i = 2 * get_global_id(0);`) has no comments about what `i` is and why we need to multiply it by 2. Can you at least add in some comments on the parts of `kernels.cl` that you touch, so we slowly improve the situation?

SplitInfinity:
@opti-mix 

I didn't even see that macro! I don't think it can be used here as is because the type of the condition input is different from that of the LHS and RHS, and we had to do that little trick with comparing the condition to zero to make sure the MSB is set/not set as needed.

SplitInfinity:
Added comments explaining `2 * global_id` as requested by @jfix71.

opti-mix:
> Added comments explaining 2 * global_id

The macros you haven't even seen use pretty much the same approach. Since you are at it anyways, could you add the same comments to them as well?

SplitInfinity:
Addressed @opti-mix 's comment - added comment about 2 * global_id to all macros that produce 16-element vectorized kernels.

PR #2913, opened 2019-05-14
Print unsupported operator when verification fails

bertmaher:
Summary: I always end up adding this logging anyways

Differential Revision: D15345120


jfix71:
I added a local static helper that does something similar in GraphOptimizer (`checkAllNodesSupported()`) -- maybe it would make sense to pull it out and use it in both places.

PR #2915, opened 2019-05-15
[OpenCL] Make OpenCLFunction thread safe

### LINKED ISSUE #2651: OpenCLCompiledFunction should be thread safe
Currently the OpenCLCompiled function uses member variables for things like commandQueue, clContext, and deviceId. These should be parameters to the functions using them so they can be passed in at execute.

SplitInfinity:
**Description*
This commit makes `OpenCLFunction` thread-safe by deleting its member
variables for the OpenCL context, device ID, command queue, etc. and
using the instances of these variables passed to `OpenCLFunction::execute`
as part of the `DeviceBindings` inside the `ExecutionContext`.

**Testing**
All unit tests pass.

**Fixes**
This commit fixes #2651.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.h
+++ b/lib/Backends/OpenCL/OpenCL.h
@@ -82,21 +82,11 @@ class OpenCLFunction final : public CompiledFunction {
   };
   /// The IR to be executed.
   std::unique_ptr<IRFunction> F_;
-  /// CL compute device id.
-  cl_device_id deviceId_;
-  /// CL compute context.
-  cl_context context_;
-  /// CL compute command queue.
-  cl_command_queue commands_;
   /// Cache of compiled programs.
   /// The same source code can be compile with different options (e.g. with
   /// different set of macro definitions) and/or for a different device and
   /// would result in different programs.
   std::unordered_map<ProgramKey, cl_program, ProgramKeyHash> programsCache_;
-  /// A pointer to the on-device memory buffer.
-  cl_mem deviceBuffer_{0};
-  /// Information about kernel launches.
-  std::vector<KernelLaunch> kernelLaunches_;
   /// is kernel level profiling (autoInstrumentation) enabled.
   bool kernelProfiling_{false};
   /// Manual trace events:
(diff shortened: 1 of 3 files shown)


SplitInfinity:
Do we have any multithreaded OpenCL tests I can use to test this?

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -82,19 +82,11 @@ class OpenCLFunction final : public CompiledFunction {
...
-  cl_device_id deviceId_;
-  /// CL compute context.
-  cl_context context_;
-  /// CL compute command queue.
-  cl_command_queue commands_;
   /// Cache of compiled programs.
   /// The same source code can be compile with different options (e.g. with
   /// different set of macro definitions) and/or for a different device and
   /// would result in different programs.
   std::unordered_map<ProgramKey, cl_program, ProgramKeyHash> programsCache_;
```
opti-mix: May be not for this PR, but `kernelLaunches_`  should be moved out of `OpenCLFunction` as it collects a per-run information.

And may be `programsCache_` should be moved out of `OpenCLFunction` as well.

cc @gcatron

SplitInfinity:
@gcatron @nickgg 

Is `programsCache_` per-run state as well?

gcatron:
> @gcatron @nickgg
> 
> Is `programsCache_` per-run state as well?

I don't think programsCache_ is per-run state, @opti-mix can correct me on that if I'm mistaken. I think the thought is that it might be better to house it in the DeviceManager. It's caching different compilations of the programs, which I expect will be device specific. If that's the case it should be moved to DeviceManager but that could be a different PR.

opti-mix:
> Is programsCache_ per-run state as well?

They are per-device-type singletons, i.e. you'd keep one for AMD GPUs, one for NVidia-GPUs, etc. Or you could keep one global singleton for all devices around. I'd suggest moving it to the DeviceManager for now. But may be in a separate PR.

opti-mix:
What's the status of this PR? The stale bot is planning to close it ;-)

SplitInfinity:
Ah, I still need to test but for that I need to revive some old test that we don't have any more :(

PR #2917, opened 2019-05-15
Bugfix to counting partitions to saturate host

narayanan2004:
Summary: Fix off-by-one error in counting partitions.

Differential Revision: D15348342

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -566,8 +566,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
     adjustLogicalDeviceID(root, mapping.getPartitions().size());
   } else if (saturateHost_) {
     // Attempt to saturate the host. Passing in the count of logical devices.
-    // Since logicalId starts at 0 we add one.
-    saturateHost(logicalID + 1);
+    saturateHost(logicalID);
   }
 }
 


lib/Partitioner/Partitioner.cpp
```diff
@@ -567,7 +567,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   } else if (saturateHost_) {
     // Attempt to saturate the host. Passing in the count of logical devices.
     // Since logicalId starts at 0 we add one.
-    saturateHost(logicalID + 1);
+    saturateHost(logicalID);
```
rdzhabarov: Do we have any unit test covering this case?

lib/Partitioner/Partitioner.cpp
```diff
@@ -567,7 +567,7 @@ void Partitioner::doPartitioning(Function *F, NodeToFunctionMap &mapping) {
   } else if (saturateHost_) {
     // Attempt to saturate the host. Passing in the count of logical devices.
     // Since logicalId starts at 0 we add one.
```
rdzhabarov: comment is probably stale now, since we do not explicitly add 1 here.

rdzhabarov:
Nice catch!

lib/Partitioner/Partitioner.cpp
narayanan2004: Will fix the comment.

rdzhabarov:
@narayanan2004 @beicy please make sure that there is a unit test exposing this bug.

narayanan2004:
Change is already committed by PR #2930

PR #2918, opened 2019-05-15
Don't activate topos when executing a HabanaFunction

bertmaher:
Summary:
The device manager manages the active topo.  As a bonus, we can kill
the global synapseLock.

Reviewed By: narayanan2004

Differential Revision: D15346530


rdzhabarov:
here you go with the clang-format failure :D

PR #2919, opened 2019-05-15
Only optimistically activate topo if nothing else is loaded

bertmaher:
Summary:
Topology activation has to be synchronized with what's running on the
device, so it's only safe to optimistically activate a topo if nothing else is
active.

Differential Revision: D15350200


bertmaher:
Yeah, good idea.  Let me see about that.

bertmaher:
So interestingly enough this actually causes some of my more complex tests to crash.  I think there may be a couple layers of bugs here...

bertmaher:
This version seems clean :).  It's stacked on #2920 so just review the top commit.

bertmaher:
Yeah I agree, an explicit request queue would be better.

PR #2921, opened 2019-05-15
Avoid copying onnxifi weights where possible

jackm321:
Summary:
Make it so that Constants can have unowned tensors as their payload and will basically copy-on-write by making sure that when a mutable reference to the payload is requested, the payload will be copied to a new owned tensor if it was previously unowned.
The motivation behind this is to load weights from onnxifi using unowned tensors as often as possible only making a copy if the weight needs to be modified by Glow.
Because Glow currently lacks a UInt8QTy quantized type, all tensors of this type must still be copied and converted to Int8QTy but if a UInt8QTy type is added to Glow then these can be taken as unowned tensors as well.
One thing to consider here is that the lifetime of the memory backing Constants created at model loading time must live for at least as long as the Constants. For our current model loading this is true but it is something that will have to be maintained.

Documentation:
Doxygen

Test Plan:
CI
[WIP] Unit test for unowned Constants.

### DIFF
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -97,7 +97,7 @@ public:
     // If payload is unowned, make an owned copy of the payload for
     // modification.
     if (payload_.isUnowned()) {
-      return payload_ = payload_.clone();
+      payload_ = payload_.clone();
     }
     assert(!payload_.isUnowned() &&
            "Can only modify Constants with owned payloads");
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -91,8 +91,20 @@ public:
     return k->getKind() == Kinded::Kind::ConstantKind;
   }
 
-  Tensor &getPayload() { return payload_; }
+  /// \returns a mutable reference to the payload tensor. If the payload tensor
+  /// is unowned then it will be converted to an owned copy before returning.
+  Tensor &getPayloadMutable() {
+    // If payload is unowned, make an owned copy of the payload for
+    // modification.
+    if (payload_.isUnowned()) {
+      return payload_ = payload_.clone();
+    }
+    assert(!payload_.isUnowned() &&
+           "Can only modify Constants with owned payloads");
+    return payload_;
+  }
 
+  // Get an immutable reference to the payload tensor.
   const Tensor &getPayload() const { return payload_; }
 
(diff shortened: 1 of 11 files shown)


jackm321:
Once #2921 is done and Glow has UInt8QTy Tensors, all Constants from onnxifi should be able to be loaded without copies.

tests/unittests/GraphTest.cpp
```diff
@@ -1266,11 +1266,95 @@ TEST(Graph, verifyConstantTensorTypeMatchesTensorTypeChanged) {
...
+  auto originalH = originalT.getHandle();
+  auto unownedH = unownedT.getHandle();
+
+  for (size_t i = 0; i < originalT.size(); i++) {
+    originalH.raw(i) = i;
+  }
+
+  // Both Tensors should have the same values because unownedH has the same
+  // underlying memory as originalH.
+  for (size_t i = 0; i < unownedH.size(); i++) {
```
jfix71: nit: could just call `Tensor::isEqual()`.

jfix71: Checking that `Tensor::getUnsafePtr()` are equal would probably be a good check too, even if it's somewhat redundant. Could be done after getting the mutable payload too, to verify that the pointers are now different.

include/glow/Graph/Nodes.h
```diff
@@ -91,8 +91,20 @@ class Constant : public Storage {
...
-  Tensor &getPayload() { return payload_; }
+  /// \returns a mutable reference to the payload tensor. If the payload tensor
+  /// is unowned then it will be converted to an owned copy before returning.
+  Tensor &getPayloadMutable() {
+    // If payload is unowned, make an owned copy of the payload for
+    // modification.
+    if (payload_.isUnowned()) {
+      return payload_ = payload_.clone();
+    }
+    assert(!payload_.isUnowned() &&
```
jfix71: nit: This feels unnecessary, as the above if condition checks the not case and returns immediately.

jackm321: Oh I didn't mean to add the return there, copy paste mistake.
Do you think it's unnecessary to assert at all though? Maybe it is

jfix71: Ah ok. Yeah go ahead and leave it then.

bertmaher:
This is awesome :-D

nickgg:
This is cool, feels like it might have an impact on compile time?

PR #2924, opened 2019-05-15
Change ThreadPool to use one task queue per thread rather than a shared queue

nickgg:
Summary: This changes the ThreadPool implementation to an array of individual Executors rather than an array of threads all servicing the same queue. This will let us remove lookups and the lock in the ThreadPoolExecutor by assigning each request to a specific thread (next diff).

This shouldn't have much affect on performance: we lose the load balancing having all threads take from the same queue, but we gain through lower contention on the thread pool lock.

Test Plan: unit tests in various modes.

### DIFF
--- a/include/glow/Support/ThreadPool.h
+++ b/include/glow/Support/ThreadPool.h
@@ -27,8 +27,8 @@
 
 namespace glow {
 
-/// Thread pool for asynchronous execution of generic functions.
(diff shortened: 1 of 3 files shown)


nickgg:
deadlocks are the worst

nickgg:
deadlock fixed

PR #2925, opened 2019-05-15
[example] Added CLI option to configure the backend for resnet-runtime example

sri-desai:
Summary:
Added configurable backend feature to CLI for resnet-runtime example.

[Optional Fixes #issue]
2812

### DIFF
--- a/examples/resnet-runtime.cpp
+++ b/examples/resnet-runtime.cpp
@@ -51,6 +51,16 @@ llvm::cl::opt<std::string> tracePath("trace-path",
                                      llvm::cl::desc("Write trace logs to disk"),
                                      llvm::cl::init(""),
                                      llvm::cl::cat(category));
+llvm::cl::opt<BackendKind> backend(
+                                    llvm::cl::desc("Backend to use:"), llvm::cl::Optional,
+                                    llvm::cl::values(clEnumValN(BackendKind::Interpreter, "interpreter",
+                                                                    "Use interpreter (default option)"),
+                                                     clEnumValN(BackendKind::CPU, "cpu", "Use CPU"),
+                                                     clEnumValN(BackendKind::OpenCL, "opencl", "Use OpenCL"),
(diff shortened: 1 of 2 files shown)


examples/resnet-runtime.cpp
```diff
@@ -51,6 +51,14 @@ llvm::cl::opt<std::string> tracePath("trace-path",
                                      llvm::cl::desc("Write trace logs to disk"),
                                      llvm::cl::init(""),
                                      llvm::cl::cat(category));
+llvm::cl::opt<BackendKind> backend(
+    llvm::cl::desc("Backend to use:"), llvm::cl::Optional,
+    llvm::cl::values(clEnumValN(BackendKind::Interpreter, "interpreter",
+                                "Use interpreter (default option)"),
+                     clEnumValN(BackendKind::CPU, "cpu", "Use CPU"),
+                     clEnumValN(BackendKind::OpenCL, "opencl", "Use OpenCL"),
+                     clEnumValN(BackendKind::Habana, "Habana", "Use Habana")),
```
gcatron: nit: since every other option is lowercase should Habana also be lowercase?

examples/resnet-runtime.cpp
```diff
@@ -51,6 +51,14 @@ llvm::cl::opt<std::string> tracePath("trace-path",
...
+    llvm::cl::init(BackendKind::Interpreter), llvm::cl::cat(category));
```
rdzhabarov: nit: to preserve current semantics this needs to be cpu backend.

examples/resnet-runtime.cpp
sri-desai: Sure. Will modify that.  Thank you.

examples/resnet-runtime.cpp
sri-desai: Will change the default backendkind to cpu from interpreter.  Thank you.

PR #2926, opened 2019-05-16
[Type] Support of uint8.

rdzhabarov:
Summary:
Add basic support for uint8.

Documentation:
n/a

Test Plan:
CI

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -111,15 +111,23 @@ public:
     case ElemKind::Int8QTy: {
       auto *data = reinterpret_cast<int8_t *>(getData());
       std::fill(&data[0], &data[0] + size(), (int8_t)type_.getOffset());
-    } break;
+      break;
+    }
+    case ElemKind::UInt8QTy: {
+      auto *data = reinterpret_cast<uint8_t *>(getData());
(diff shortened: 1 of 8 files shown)


tlepley-cadence:
unsigned int8 has been until now models in Glow as int8 with offset -128. I have the feeling that introducing an actual uint8 type is going to be confusing. 

I think that when the target requires an actual uint8_t (what is our case by the way), we can still have the s8/-128 modelling during the full compiler flow and simply do the conversion from s8/-128 to u8 at the very end of the flow at code generation time.

rdzhabarov:
I'll split this into actual type addition and handling uint8 ops in the interpreter.

jackm321:
@rdzhabarov one approach to quickly unblock loading larger models without changing too much outside of the loaders would be to introduce a UInt8QTy, load Constants with this type if necessary, allow the "Rowwise" ops to transform their data into UInt8FusedQTy, then at the end of model loading transform all Constants with payloads still in UInt8QTy to Int8QTy. This would prevent any UInt8QTy outside of loaders and prevent copying large constants. Later we can add more full-fledged UInt8QTy support if needed.

rdzhabarov:
@jackm321 agree, that's what i'll do next. I think this PR is good to go though. I'll add support of the interpreter in the next PR.

PR #2928, opened 2019-05-16
Add preprocess-images flag to image-classifier

gcatron:
Summary:
This will preload all images before classifying them, this is useful for benchmarking.
Documentation: NA

Test Plan: Ninja test, and run image-classifier exercising the new option.

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -25,9 +25,12 @@
 #include "llvm/ADT/StringSwitch.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Format.h"
+#include "llvm/Support/Timer.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <atomic>
 #include <fstream>
+#include <future>
 #include <iostream>
 #include <memory>
 #include <queue>
@@ -35,6 +38,8 @@
 
 using namespace glow;
 
+/// timeOpt defined in the Loader.
+extern llvm::cl::opt<bool> timeOpt;
 namespace {
 
 /// Image loader options.
@@ -45,6 +50,11 @@ llvm::cl::list<std::string> inputImageFilenames(
                    "mode, where the model is compiled once and then can be run "
                    "many times with new input filenames passed via stdin)"),
     llvm::cl::ZeroOrMore);
+
+llvm::cl::opt<bool> preProcessImages(
+    "preprocess-images",
+    llvm::cl::desc("Perform Image processing outside of the inference loop."),
+    llvm::cl::Optional, llvm::cl::init(false), llvm::cl::cat(imageLoaderCat));
 
 llvm::cl::opt<std::string> inputImageListFile(
     "input-image-list-file",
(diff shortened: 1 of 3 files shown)


tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +479,83 @@ int main(int argc, char **argv) {
...
+      llvm::Error runErr = llvm::Error::success();
+      inFlight++;
+      hostManager->runNetwork(
+          name, std::move(batch.second),
+          [&runErr, output, names, &runPromise,
+           &inFlight](runtime::RunIdentifierTy, llvm::Error err,
+                      std::unique_ptr<ExecutionContext> contextPtr) {
+            runErr = std::move(err);
+            // Print the top-k results from the output Softmax tensor.
+            processAndPrintResults(output, names);
```
nickgg: The HostManager has several threads, will the results printing now possibly be interleaved on stdout?

tools/loader/ImageClassifier.cpp
```diff
@@ -198,9 +208,8 @@ buildAndCompileAndGetInAndOutPair(Loader &loader, PlaceholderBindings &bindings,
   // Get the Tensor from the Placeholder that the final expected Softmax writes
   // into at the end of image inference.
   Placeholder *SMPH = EXIT_ON_ERR(LD->getSingleOutput());
-  Tensor *SMT = bindings.get(SMPH);
```
nickgg: If we're not returning the Tensor do we still need to allocate it (above)?

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +479,83 @@ int main(int argc, char **argv) {
...
+            }
+          });
+
+      EXIT_ON_ERR(std::move(runErr));
+    }
+    // wait for all to finish.
+    fut.wait();
+    if (timeOpt) {
+      timer.stopTimer();
+      llvm::outs() << llvm::formatv("Wall time per item (s): {0:f4}\n",
```
nickgg: This PR also adds timing/benchmarking functionality to the ImageClassifier? Could you update the summary to reflect this (& how it works, what it looks like, etc)?

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +479,83 @@ int main(int argc, char **argv) {
...
+
+      preLoadedImages.push_back(std::move(inputImageData));
+      contexts.emplace(inputImageBatchFilenames, std::move(newContext));
+    } else {
+      loader.runInference(*bindings, batchSize);
+      // Print the top-k results from the output Softmax tensor.
+      numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
+    }
+  }
+  if (preProcessImages) {
```
nickgg: Doesn't this disable running the network if preProcessImages is false?

gcatron:
Had a discussion with @jfix71, Since the goal of this was really to enable benchmarking, and for that we don't really need to use real images I'm going to close this PR and refactor this to more of a pure benchmark use case.

tools/loader/ImageClassifier.cpp
gcatron: We still need the backing tensor, we just access it via the PH now.

PR #2929, opened 2019-05-16
[image-classifier]:Add a flag to dump DAG before compilation and optimization.

ZchiPitt:
Summary:
Add a flag -dump-graph-DAG-before-compile so that we can dump the DAG before all optimization and compilation takes place.

Test Plan:
Test with the commands below:
- ./bin/image-classifier tests/images/imagenet/cat_285.png -image-mode=0to1 -m=googlenet_v1_slim/googlenet_v1_slim.onnx -model-input-name=input:0 -image-layout=NHWC -cpu -dump-graph-DAG=2.dot -dump-graph-DAG-before-compile=1.dot
- dot -Tpdf 1.dot > before.pdf
- dot -Tpdf 2.dot > after.pdf

The generated DAGs look like below:

Before compilation and optimization
Screen Shot 2019-05-16 at 1 59 29 PM

====================================
After compilation and optimization
Screen Shot 2019-05-16 at 2 01 11 PM

### DIFF
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -148,6 +148,11 @@ llvm::cl::OptionCategory
                    "graphs by writing the IR/Graphs to "
                    "given files/stdout");
 
+llvm::cl::opt<std::string> dumpGraphDAGFileBeforeCompilationOpt(
+    "dump-graph-DAG-before-compile",
+    llvm::cl::desc("Specify the file to export the Graph in DOT format"),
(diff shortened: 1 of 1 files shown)


beicy:
Do we need to add similar code to ExecutionEngine::compile? Some of our models are created using c++.

nickgg:
Does that path have the post-compile dag option?

ZchiPitt:
I think it's easy to just call F->dump() before EE.compile(). It that's the case, we probably dont need to pass a parameter to do so.

PR #2930, opened 2019-05-16
[Partitioner] Saturating the host can only be enabled for single type of backend.

beicy:
Summary:
To make things simple, we won't consider saturating the host for heterogeneous partition now. If later users requests come, we can improve this part. 

Documentation:

Test Plan:
ninja check.

[Optional Fixes #issue]


rdzhabarov:
I'm a bit concerned that there are no tests here.

beicy:
> I'm a bit concerned that there are no tests here.



> I'm a bit concerned that there are no tests here.

Will add tests for saturating the host.

beicy:
Coded added to test "SaturatHost_" flag. @rdzhabarov @gcatron

PR #2931, opened 2019-05-16
Added benchmark ability to image-classifier using -iterations

gcatron:
Summary:
This flag will set image-classifier to run the specified number of inference requests against empty images and print the time taken per image to the terminal
Documentation:NA

Test Plan: tested new functionality and ensured it works and tested image-classifier original functionality is preserved. Also run.sh

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -72,6 +72,11 @@ llvm::cl::opt<unsigned> labelOffset(
     llvm::cl::desc("Label offset for TF ONNX models with 1001 classes"),
     llvm::cl::Optional, llvm::cl::init(0), llvm::cl::cat(imageLoaderCat));
 
+llvm::cl::opt<unsigned> poolSize(
+    "pool-size",
+    llvm::cl::desc("Size of context pool for the benchmark; default:10"),
+    llvm::cl::Optional, llvm::cl::init(10), llvm::cl::cat(imageLoaderCat));
+
 llvm::cl::opt<bool> computeSoftmax(
     "compute-softmax", llvm::cl::desc("Compute softmax of the network output"),
     llvm::cl::Optional, llvm::cl::init(false), llvm::cl::cat(imageLoaderCat));
@@ -366,12 +371,80 @@ static void parseInputImageList(const std::string &inputImageListFile) {
   inFile.close();
 }
 
+/// Run inference request on HostManager. This method builds a runNetwork
+/// request for the \p hostManager, this is a recursive call, in the callback
+/// provided to the HostManager this function can call itself if the desired
+/// number of requests has not yet been dispatched.
+static void runInference(runtime::HostManager *hostManager, std::string name,
+                         std::unique_ptr<ExecutionContext> batch,
+                         std::promise<void> &runPromise,
+                         std::atomic<unsigned> &inflight,
+                         std::atomic<int> &dispatched) {
+  hostManager->runNetwork(name, std::move(batch),
+                          [&runPromise, &inflight, &dispatched, hostManager,
+                           name](runtime::RunIdentifierTy, llvm::Error err,
+                                 std::unique_ptr<ExecutionContext> contextPtr) {
+                            EXIT_ON_ERR(std::move(err));
+                            // Kick off another run.
+                            if (dispatched.fetch_sub(1) > 0) {
+                              inflight++;
+                              runInference(hostManager, name,
+                                           std::move(contextPtr), runPromise,
+                                           inflight, dispatched);
+                            }
+                            if (--inflight == 0) {
+                              runPromise.set_value();
+                            }
+                          });
+}
+
+/// Run the requested number of benchmark requests \p requestCount through the
+/// HostManager from the \p loader using the provided context pool \p contexts
+/// and wait for all runs to complete.
+static void
+runBenchmark(std::string name, Loader &loader,
+             std::vector<std::unique_ptr<ExecutionContext>> contexts,
+             unsigned requestCount) {
+  runtime::HostManager *hostManager = loader.getHostManager();
+  std::atomic<unsigned> inflight(0);
+  std::atomic<int> dispatched(requestCount);
+  std::promise<void> runPromise;
+  auto fut = runPromise.get_future();
+
+  // Kick off initial pool of requests.
+  for (size_t i = 0, e = contexts.size(); i < e; i++) {
+    auto batch = std::move(contexts[i]);
+    inflight++;
+    dispatched--;
+    runInference(hostManager, name, std::move(batch), runPromise, inflight,
+                 dispatched);
+  }
+  // Wait for all to finish.
+  fut.wait();
+}
+
+/// Setup the pool of contexts needed for a benchmark run.
+static std::vector<std::unique_ptr<ExecutionContext>>
+setupContextPool(Placeholder *outputPH, Placeholder *inputImagePH,
+                 glow::Tensor &inputImageData) {
+  std::vector<std::unique_ptr<ExecutionContext>> contexts;
+  // Size of the pool, the smaller of poolSize or the actual number of
+  // requests.
+  unsigned iterations =
+      miniBatch ? std::min(int(poolSize), int(iterationsOpt / miniBatch)) : 1;
+  // Setup pool of inference requests to be run.
+  for (unsigned i = 0; i < iterations; i++) {
+    auto newContext = llvm::make_unique<ExecutionContext>();
+    auto ph = newContext->getPlaceholderBindings();
+    ph->insert(inputImagePH, Tensor(inputImageData.getType()));
+    ph->allocate(outputPH);
+    contexts.push_back(std::move(newContext));
+  }
+  return contexts;
+}
+
 int main(int argc, char **argv) {
-
-  auto context = llvm::make_unique<ExecutionContext>();
-  llvm::Timer timer("Infer", "Infer");
-
-  PlaceholderBindings *bindings = context->getPlaceholderBindings();
+  PlaceholderBindings bindings;
   // The loader verifies/initializes command line parameters, and initializes
   // the ExecutionEngine and Function.
   Loader loader(argc, argv);
@@ -403,14 +476,14 @@ int main(int argc, char **argv) {
   }
 
   // Stream input mode.
-  bool streamInputFilenamesMode =
+  const bool streamInputFilenamesMode =
       inputImageFilenames.size() == 1 && inputImageFilenames.front() == "-";
 
   GLOW_ASSERT(!(streamInputFilenamesMode && emittingBundle()) &&
               "Cannot emit a bundle and also stream inputs.");
 
   // Mini-batch mode.
-  bool miniBatchMode = miniBatch > 0;
+  const bool miniBatchMode = miniBatch > 0;
   GLOW_ASSERT(((!miniBatchMode) || (!streamInputFilenamesMode)) &&
               "The minibatch option is not compatible with the stream input "
               "image mode.");
@@ -418,7 +491,8 @@ int main(int argc, char **argv) {
       ((!miniBatchMode) || (inputImageFilenames.size() % miniBatch == 0)) &&
       "The number of input images must be a multiple of the mini-batch.");
 
-  GLOW_ASSERT(((!iterationsOpt) || (iterationsOpt % miniBatch == 0)) &&
+  GLOW_ASSERT(((!iterationsOpt) || (!miniBatchMode) ||
+               (iterationsOpt % miniBatch == 0)) &&
               "Benchmark count must be a multiple of the mini-batch.");
 
   // Used to make sure we only compile once, and run only once if not streaming.
@@ -427,7 +501,6 @@ int main(int argc, char **argv) {
   // These will be set during the first run.
   Placeholder *inputImagePH = nullptr;
   Placeholder *outputPH = nullptr;
-  Tensor *SMT = nullptr;
 
   size_t minibatchIndex = 0;
   Tensor inputImageData;
@@ -440,40 +513,27 @@ int main(int argc, char **argv) {
   llvm::outs() << "Model: " << loader.getFunction()->getName() << "\n";
 
   int numErrors = 0;
-
-  // Image tensors loaded up to be run at once for benchmark mode.
-  std::vector<std::unique_ptr<Tensor>> preLoadedImages;
-  std::vector<std::unique_ptr<ExecutionContext>> contexts;
-
   while ((streamInputFilenamesMode &&
           getNextImageFilenames(&inputImageBatchFilenames)) ||
          (miniBatchMode &&
           getNextMiniBatch(inputImageBatchFilenames, inputImageFilenames,
                            minibatchIndex, miniBatch)) ||
          isFirstRun) {
-    auto inputImageData = llvm::make_unique<Tensor>();
+    // Load and process the image data into the inputImageData Tensor.
+    loadImagesAndPreprocess(inputImageBatchFilenames, &inputImageData,
+                            imageNormMode, imageChannelOrder, imageLayout);
+
+    // It we are benchmarking reset the image data to the batch size we need.
     if (iterationsOpt) {
-      // Load one image so we know dimensions and can create the appropriate
-      // batch size.
-      streamInputFilenamesMode = false;
-      miniBatchMode = false;
-      std::vector<std::string> firstImage;
-      firstImage.push_back(inputImageBatchFilenames[0]);
-      loadImagesAndPreprocess(firstImage, inputImageData.get(), imageNormMode,
-                              imageChannelOrder, imageLayout);
-      ShapeVector imageSize(inputImageData->getType().dims().begin(),
-                            inputImageData->getType().dims().end());
+      ShapeVector imageSize(inputImageData.getType().dims().begin(),
+                            inputImageData.getType().dims().end());
       if (miniBatch) {
         imageSize[0] = miniBatch;
       } else {
         imageSize[0] = iterationsOpt;
       }
       // Resize the Tensor to the appropriate size.
-      inputImageData.get()->reset(ElemKind::FloatTy, imageSize);
-    } else {
-      // Load and process the image data into the inputImageData Tensor.
-      loadImagesAndPreprocess(inputImageBatchFilenames, inputImageData.get(),
-                              imageNormMode, imageChannelOrder, imageLayout);
+      inputImageData.reset(ElemKind::FloatTy, imageSize);
     }
     // If this is the first run, then we need to build and compile the model.
     if (isFirstRun) {
@@ -482,8 +542,8 @@ int main(int argc, char **argv) {
       // Build and compile the graph, and then get back the input Placeholder
       // and output Placeholder.
       std::pair<Placeholder *, Placeholder *> inputOutputPair =
-          buildAndCompileAndGetInAndOutPair(loader, *bindings,
-                                            &inputImageData->getType());
+          buildAndCompileAndGetInAndOutPair(loader, bindings,
+                                            &inputImageData.getType());
 
       // If in bundle mode, the bundle has been saved by the above call, so we
       // can safely return.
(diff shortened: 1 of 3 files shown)


bertmaher:
We have an -iterations flags that sounds like it does something similar...

gcatron:
> We have an -iterations flags that sounds like it does something similar...

Good point, I'll use that flag.

tools/loader/ImageClassifier.cpp
```diff
@@ -430,25 +446,50 @@ int main(int argc, char **argv) {
...
     // If this is the first run, then we need to build and compile the model.
     if (isFirstRun) {
       isFirstRun = false;
 
       // Build and compile the graph, and then get back the input Placeholder
       // and output Softmax Tensor.
-      std::pair<Placeholder *, Tensor *> inputOutputPair =
-          buildAndCompileAndGetInAndOutPair(loader, bindings,
-                                            &inputImageData.getType());
+      std::pair<Placeholder *, Placeholder *> inputOutputPair =
```
beicy: Comments need to be updated.

tools/loader/Loader.cpp
```diff
@@ -373,20 +375,20 @@ void Loader::compile(PlaceholderBindings &bindings) {
 void Loader::runInference(PlaceholderBindings &bindings, size_t batchSize) {
   assert(!emittingBundle() &&
          "No inference is performed in the bundle generation mode.");
-
+  unsigned iterations = iterationsOpt == 0 ? 1 : iterationsOpt;
```
beicy: Can we just initial `iterationsOpt` as 1?

gcatron: I made this change since I'm using the 0 state as a flag in Image-Classifier, I don't think there's a way to check that the iterationsOpt is 1 because it was provided or 1 because that's what it was defaulted to.

tools/loader/ImageClassifier.cpp
```diff
@@ -430,25 +446,50 @@ int main(int argc, char **argv) {
...
          (miniBatchMode &&
           getNextMiniBatch(inputImageBatchFilenames, inputImageFilenames,
                            minibatchIndex, miniBatch)) ||
          isFirstRun) {
-    // Load and process the image data into the inputImageData Tensor.
-    loadImagesAndPreprocess(inputImageBatchFilenames, &inputImageData,
-                            imageNormMode, imageChannelOrder, imageLayout);
-
+    auto inputImageData = llvm::make_unique<Tensor>();
+    if (iterationsOpt) {
```
bertmaher: I'm really bothered that it adds so much code to this already-complicated while loop just to support "iterations".  In my mind adding iterations should be a two line diff that just adds a loop :-).  Instead there's kind of a whole new way of doing everything and that feels brittle :-/

gcatron: A fair concern, I considered bypassing the while loop entirely but I think that would add more complexity, since it does seem to be almost a separate flow entirely maybe I do bypass the while loop entirely?

tools/loader/ImageClassifier.cpp
```diff
@@ -25,16 +25,24 @@
...
 #include <fstream>
+#include <future>
 #include <iostream>
 #include <memory>
 #include <queue>
 #include <sstream>
 
 using namespace glow;
 
+/// timeOpt defined in the Loader.
```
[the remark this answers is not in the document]
gcatron: Good point, I opted to keep internal storage but I moved the externs to Loader.h, which looks to the the same pattern used elsewhere in the codebase.

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +492,87 @@ int main(int argc, char **argv) {
...
+      unsigned requestCount = 1;
+      if (miniBatch) {
+        requestCount = iterationsOpt / miniBatch;
+      }
+      // Setup list of inference requests to be run.
+      for (unsigned i = 0; i < requestCount; i++) {
+        auto newImage = llvm::make_unique<Tensor>(inputImageData->getType());
+        auto newContext = llvm::make_unique<ExecutionContext>();
+        auto ph = newContext->getPlaceholderBindings();
+        ph->allocate(inputImagePH);
```
bertmaher: Am I reading it correctly that this loop is going to allocate `requestCount` batches of images ahead of time?  If I wanted to run 10,000 iterations of resnet50 (not unrealistic for the Habana chip) would I need to allocate 6 GB of memory?  That doesn't seem desirable for a benchmark...

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +492,87 @@ int main(int argc, char **argv) {
...
-    numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
+    auto batchSize = inputImageData->dims()[0];
+    if (iterationsOpt) {
+      unsigned requestCount = 1;
+      if (miniBatch) {
+        requestCount = iterationsOpt / miniBatch;
+      }
+      // Setup list of inference requests to be run.
+      for (unsigned i = 0; i < requestCount; i++) {
+        auto newImage = llvm::make_unique<Tensor>(inputImageData->getType());
```
bertmaher: Do we actually copy the image data, or do we just do inference on garbage data in benchmark mode?

tools/loader/ImageClassifier.cpp
```diff
@@ -430,25 +440,50 @@ int main(int argc, char **argv) {
...
           getNextImageFilenames(&inputImageBatchFilenames)) ||
          (miniBatchMode &&
           getNextMiniBatch(inputImageBatchFilenames, inputImageFilenames,
                            minibatchIndex, miniBatch)) ||
          isFirstRun) {
-    // Load and process the image data into the inputImageData Tensor.
-    loadImagesAndPreprocess(inputImageBatchFilenames, &inputImageData,
-                            imageNormMode, imageChannelOrder, imageLayout);
-
+    auto inputImageData = llvm::make_unique<Tensor>();
```
bertmaher: This is shadowing the `inputImageData` in the enclosing scope.  Is there any reason to move this inside the loop and unique-ptr it, rather than leaving it stack allocated outside the loop?  I strongly prefer the latter if possible.

tools/loader/ImageClassifier.cpp
```diff
@@ -430,25 +440,50 @@ int main(int argc, char **argv) {
...
+      std::vector<std::string> firstImage;
+      firstImage.push_back(inputImageBatchFilenames[0]);
+      loadImagesAndPreprocess(firstImage, inputImageData.get(), imageNormMode,
+                              imageChannelOrder, imageLayout);
+      ShapeVector imageSize(inputImageData->getType().dims().begin(),
+                            inputImageData->getType().dims().end());
+      if (miniBatch) {
+        imageSize[0] = miniBatch;
+      } else {
+        imageSize[0] = iterationsOpt;
```
bertmaher: This doesn't seem right, I don't think I should need to allocate `iterations` times as much memory to run a benchmark `iterations` times.  Coupled with the requests below I think we might be using something like `iterations`^2  as much memory as needed...

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +492,87 @@ int main(int argc, char **argv) {
...
+        contexts.push_back(std::move(newContext));
+      }
+    } else {
+      loader.runInference(*bindings, batchSize);
+      // Print the top-k results from the output Softmax tensor.
+      numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
+    }
+  }
+  if (iterationsOpt) {
+    runtime::HostManager *hostManager = loader.getHostManager();
```
bertmaher: Here's a thought.  Can this block be factored into a `run` function and used in place of `runInference` above?  I don't really see anything specific to `iterations` here other than that the batch is big.

tools/loader/ImageClassifier.cpp
bertmaher: I guess I wonder why we can't just `inputImageBatch` like we do in the non-iterations case.

tools/loader/ImageClassifier.cpp
gcatron: Garbage data we don't actually care about the results for benchmarking.

tools/loader/ImageClassifier.cpp
gcatron: That is correct, the thought here was we want each image to be a different location, we could reuse the same context N many times but wouldn't we see better results than expected due to caching?

tools/loader/ImageClassifier.cpp
gcatron: In the else branch we will only allocate iterationsOpt worth of memory because the enclosing loop will only iterate once.

tools/loader/ImageClassifier.cpp
gcatron: The unique pointer is because we plan on moving it into a vector and storing it, a single stack allocated image won't work for that.

tools/loader/ImageClassifier.cpp
gcatron: I'm not sure I follow, inputImageBatch is the number of images for a single request. We'd still need the logic to run multiple requests right?

nickgg: Maybe the branch should be above this loop then? It would be nice to refactor the loading logic into a helper to avoid this loop getting hard to follow. Generally I find code hard to read if a block's begin and end don't fit on the same screen.

tools/loader/ImageClassifier.cpp
nickgg: Might be garbage, might be zeros (?) we'd probably get a more interesting result by randomizing these tensors.

tools/loader/ImageClassifier.cpp
nickgg: We'd have a similar caching effect in the production system I think since that's more or less how we allocate there. I think maybe you can abstract that somewhat by using a TensorPool with a fixed size, with a condition variable that you can block waiting for old requests to finish.

tools/loader/Loader.cpp
nickgg: What does an iterationsOpt as 1 mean that's different from iterationsOpt == 0? You'd run one iteration either way, is there a side effect?

I assume yes, so I guess I'm really asking if we can separate `iterations` and it's side effect into different flags.

tools/loader/ImageClassifier.cpp
nickgg: I'm feeling a bit confused by the overlap between the number of iterations to run (& total number of images?) and the batchSize. It would be good to have the ability to run any number of iterations over any number of images with any batch size.

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +492,87 @@ int main(int argc, char **argv) {
...
+          [&runErr, &runPromise,
+           &inFlight](runtime::RunIdentifierTy, llvm::Error err,
+                      std::unique_ptr<ExecutionContext> contextPtr) {
+            runErr = std::move(err);
+            if (--inFlight == 0) {
+              runPromise.set_value();
+            }
+          });
+
+      EXIT_ON_ERR(std::move(runErr));
```
nickgg: `runErr` is likely not valid at this point since it's only set after the future is complete. I think for this design where the future is set after `inFlight` results you need to do the equivalent of logical OR on the runErr and check it at the end. Alternatively just exit (crash) in the runNetwork callback.

tools/loader/ImageClassifier.cpp
```diff
@@ -457,34 +492,87 @@ int main(int argc, char **argv) {
...
+    } else {
+      loader.runInference(*bindings, batchSize);
+      // Print the top-k results from the output Softmax tensor.
+      numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
+    }
+  }
+  if (iterationsOpt) {
+    runtime::HostManager *hostManager = loader.getHostManager();
+    std::string name = loader.getFunctionName();
+    std::atomic<unsigned> inFlight(0);
```
nickgg: nit: inflight is one word so this doesn't need camel case.

tools/loader/Loader.cpp
gcatron: In the loader iterations needs to be 1 or more or else the inference will not run. 
There was originally a separate benchmark flag, @bertmaher requested I combing into iterations as they were similar.

tools/loader/ImageClassifier.cpp
gcatron: mini-batch is the number of images per request, iterations is total images.
so if mini-batch is 10 and total images is 50 you will run 5 requests.

gcatron:
I ended up pulling the benchmarking out of the while loop now, and moved the actual run logic into its own function. Hopefully this helps with the added complexity. I'll probably add a TensorPool like @nickgg suggested and add an option to configure the size of the pool.

tools/loader/ImageClassifier.cpp
```diff
@@ -430,12 +464,15 @@ int main(int argc, char **argv) {
...
 
   int numErrors = 0;
-  while ((streamInputFilenamesMode &&
-          getNextImageFilenames(&inputImageBatchFilenames)) ||
-         (miniBatchMode &&
-          getNextMiniBatch(inputImageBatchFilenames, inputImageFilenames,
-                           minibatchIndex, miniBatch)) ||
-         isFirstRun) {
+
+  while (!iterationsOpt &&
```
jfix71: After looking through this it looks to me like there's a lot of shared logic between here inside the while loop, and in the `iterationsOpt` case after this while loop.

I'm wondering if either you could merge most of the `iterationsOpt` logic into the while loop here, or otherwise break out of the while loop early before we run inference so that many of the data structures are correctly set up? E.g. this loop sets up `inputImageBatchFilenames`, compiles the model, etc. which you also do below.

tools/loader/ImageClassifier.cpp
```diff
@@ -364,8 +366,36 @@ static void parseInputImageList(const std::string &inputImageListFile) {
...
+static void
+runBenchmark(std::string name, Loader &loader,
+             std::vector<std::unique_ptr<ExecutionContext>> contexts) {
+  runtime::HostManager *hostManager = loader.getHostManager();
+  std::atomic<unsigned> inflight(0);
+  std::promise<void> runPromise;
+  auto fut = runPromise.get_future();
+
+  for (auto &batch : contexts) {
+    llvm::Error runErr = llvm::Error::success();
```
jfix71: This appears to be unused?

tools/loader/ImageClassifier.cpp
```diff
@@ -475,12 +513,93 @@ int main(int argc, char **argv) {
...
     // Perform the inference execution, updating SMT.
     auto batchSize = inputImageData.dims()[0];
-    loader.runInference(bindings, batchSize);
 
+    loader.runInference(bindings, batchSize);
     // Print the top-k results from the output Softmax tensor.
     numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
   }
 
+  // Image tensors loaded up to be run at once for benchmark mode.
```
jfix71: `main()` is getting super long at this point. Can we extract all of the newly added logic here could probably go into a helper function?

tools/loader/ImageClassifier.cpp
gcatron: Yeah it's been a few iterations, I took another look at moved some logic in, but not all I think this is a pretty clean approach.

tools/loader/ImageClassifier.cpp
```diff
@@ -364,6 +372,76 @@ static void parseInputImageList(const std::string &inputImageListFile) {
...
+
+  // Kick off initial pool of requests.
+  for (unsigned i = 0; i < contexts.size(); i++) {
+    auto batch = std::move(contexts[i]);
+    inflight++;
+    dispatched--;
+    runInference(hostManager, name, std::move(batch), runPromise, inflight,
+                 dispatched);
+  }
+  // wait for all to finish.
```
jfix71: nit: Capitalize "wait"

tools/loader/ImageClassifier.cpp
```diff
@@ -469,18 +565,43 @@ int main(int argc, char **argv) {
...
     auto batchSize = inputImageData.dims()[0];
     loader.runInference(bindings, batchSize);
-
     // Print the top-k results from the output Softmax tensor.
     numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
   }
 
+  if (iterationsOpt) {
+    // Image tensors loaded up to be run at once for benchmark mode.
+    std::vector<std::unique_ptr<Tensor>> preLoadedImages;
```
jfix71: I think I might be missing something, but it seems like this isn't actually used anywhere? It's passed into `setupContextPool()` which adds tensors to it, but otherwise isn't used anywhere right? Once you call `updateInputPlaceholders()` the Placeholder has its own backing Tensor assigned from that Tensor. In fact inside `setupContextPool()` you could probably just insert `newImage` into the `PlaceholderBindings`, right? Either way not sure this is necessary.

tools/loader/ImageClassifier.cpp
```diff
@@ -364,6 +372,76 @@ static void parseInputImageList(const std::string &inputImageListFile) {
...
+  // Size of the pool, the smaller of poolSize or the actual number of
+  // requests.
+  unsigned iterations =
+      miniBatch ? std::min(int(poolSize), int(iterationsOpt / miniBatch)) : 1;
+  // Setup pool of inference requests to be run.
+  for (unsigned i = 0; i < iterations; i++) {
+    auto newImage = llvm::make_unique<Tensor>(inputImageData.getType());
+    auto newContext = llvm::make_unique<ExecutionContext>();
+    auto ph = newContext->getPlaceholderBindings();
+    ph->allocate(inputImagePH);
```
jfix71: Yeah so referring back to my other comment, I think instead of allocating here and then calling `updateInputPlaceholders()` you can insert use `PlaceholderBindings::insert()` to insert the `newImage` alongside `inputImagePH`. Then `preLoadedImages` can be removed. Right?

tools/loader/ImageClassifier.cpp
```diff
@@ -469,18 +565,43 @@ int main(int argc, char **argv) {
...
-
     // Print the top-k results from the output Softmax tensor.
     numErrors += processAndPrintResults(SMT, inputImageBatchFilenames);
   }
 
+  if (iterationsOpt) {
+    // Image tensors loaded up to be run at once for benchmark mode.
+    std::vector<std::unique_ptr<Tensor>> preLoadedImages;
+    std::vector<std::unique_ptr<ExecutionContext>> contexts;
+    setupContextPool(outputPH, inputImagePH, inputImageData, preLoadedImages,
```
jfix71: nit: You could probably just return the `std::vector<std::unique_ptr<ExecutionContext>>` from `setupContextPool()` instead of taking it as a reference.

tools/loader/ImageClassifier.cpp
```diff
@@ -364,6 +372,76 @@ static void parseInputImageList(const std::string &inputImageListFile) {
...
+             std::vector<std::unique_ptr<ExecutionContext>> contexts,
+             unsigned requestCount) {
+  runtime::HostManager *hostManager = loader.getHostManager();
+  std::atomic<unsigned> inflight(0);
+  std::atomic<int> dispatched(requestCount);
+  std::promise<void> runPromise;
+  auto fut = runPromise.get_future();
+
+  // Kick off initial pool of requests.
+  for (unsigned i = 0; i < contexts.size(); i++) {
```
jfix71: nit: `size_t` instead of `unsigned`. Also, follow for loop pattern from other places, e.g. `size_t i = 0, e = contexts.size(); i < e; i++`

tools/loader/ImageClassifier.cpp
```diff
@@ -364,6 +372,76 @@ static void parseInputImageList(const std::string &inputImageListFile) {
...
+                                           std::move(contextPtr), runPromise,
+                                           inflight, dispatched);
+                            }
+                            if (--inflight == 0) {
+                              runPromise.set_value();
+                            }
+                          });
+}
+
+/// Run benchmark and wait for all runs to complete.
```
jfix71: Can you expand on the doxygen here?

tools/loader/ImageClassifier.cpp
```diff
@@ -364,6 +372,76 @@ static void parseInputImageList(const std::string &inputImageListFile) {
   inFile.close();
 }
 
+/// Run inference request on HostManager.
```
jfix71: Expand on doxygen please.

PR #2932, opened 2019-05-16
Dilation support 

### LINKED ISSUE #2767: Support dilated convolution
We should support *dilated convolution*, described [here](https://arxiv.org/pdf/1603.07285.pdf) and animated [here](https://github.com/vdumoulin/conv_arithmetic/blob/master/README.md#dilated-convolution-animations).

Once we do this, we can optionally lower `ConvolutionGradNode` to a `ConvolutionNode` with dilation and optionally lower that further to multiple, non-dilated `ConvolutionNodes` and stitch their outputs together.

ksaurabh-cadence:
Summary:
1. Dilation support in loaders and interpreter
2. Add Dilation support in CPU backend kernel (float and int8)
3. Disable DKKC8 optimization for convolution when dilation!=1
4. Added operator test for dilation (Interp and CPU only)

Documentation:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Conv

Test Plan:
Unit Tests 

fixes #2767

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -277,7 +277,10 @@ public:
   /// of steps to take in the input for each output cell. \p pads defines how
   /// many zero padding cells should be added to the input during convolution.
   /// \p group defines the number of groups the input and output channels should
-  /// be divided into and convolved separately.
+  /// be divided into and convolved separately. \p dilation defines factor by
+  /// which gap between 2 neighboring kernel elements is expanded along each
+  /// axis.
+
   ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
                               NodeValue filter, NodeValue bias, TypeRef outTy,
                               llvm::ArrayRef<unsigned_t> kernels,
@@ -291,7 +294,10 @@ public:
   /// of steps to take in the input for each output cell. \p pad defines how
   /// many zero padding cells should be added to the input during convolution.
   /// \p group defines the number of groups the input and output channels should
-  /// be divided into and convolved separately.
+  /// be divided into and convolved separately. \p dilation defines factor by
+  /// which gap between 2 neighboring kernel elements is expanded along each
+  /// axis.
+
   ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
                               NodeValue filter, NodeValue bias, TypeRef outTy,
                               unsigned_t kernel, unsigned_t stride,
@@ -928,7 +934,9 @@ public:
   /// to take in the input for each output cell. \p pads defines how many zero
   /// padding cells should be added to the input during convolution. \p group
   /// defines the number of groups the input and output channels should be
-  /// divided into and convolved separately.
+  /// divided into and convolved separately. \p dilation defines factor by
+  /// which gap between 2 neighboring kernel elements is expanded along each
+  /// axis.
   ConvolutionNode *createConv(PlaceholderBindings &bindings,
                               llvm::StringRef name, NodeValue input,
                               size_t outChannels,
@@ -943,7 +951,9 @@ public:
   /// take in the input for each output cell. \p pad defines how many zero
   /// padding cells should be added to the input during convolution. \p group
   /// defines the number of groups the input and output channels should be
-  /// divided into and convolved separately.
+  /// divided into and convolved separately.\p dilation defines factor by
+  /// which gap between 2 neighboring kernel elements is expanded along each
+  /// axis.
   ConvolutionNode *createConv(PlaceholderBindings &bindings,
                               llvm::StringRef name, NodeValue input,
                               size_t outChannels, unsigned_t kernel,
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -282,8 +282,8 @@ public:
                               NodeValue filter, NodeValue bias, TypeRef outTy,
                               llvm::ArrayRef<unsigned_t> kernels,
                               llvm::ArrayRef<unsigned_t> strides,
-                              llvm::ArrayRef<unsigned_t> pads,
-                              unsigned_t group);
+                              llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
+                              unsigned_t dilation = 1);
 
   /// Creates a ConvolutionNode with the given \p name which convolves the 4D
   /// \p input with \p filter and \bias. \p kernel defines the size of the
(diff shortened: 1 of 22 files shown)


ksaurabh-cadence:
Why is the merging still blocked? Do I have to sign an individual CLA as well?  Thanks!

lib/Backends/CPU/Transforms.cpp
```diff
@@ -54,6 +54,11 @@ static Node *optimizeCPUConv(ConvolutionNode *CN, Function *F) {
     return nullptr;
   }
 
+  // This optimization is not supported with Dilation currently
```
jackm321: period

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -486,6 +486,11 @@ llvm::Error ONNXModelLoader::loadConv(const ONNX_NAMESPACE::NodeProto &op,
     ASSIGN_VALUE_OR_RETURN_ERR(group, loadInt(dict.at("group")));
   }
 
+  unsigned_t dilation = 1;
+  if (dict.count("dilation")) {
+    ASSIGN_VALUE_OR_RETURN_ERR(dilation, loadInt(dict.at("dilation")));
```
jackm321: This doesn't look like it's an onnx conv attribute https://github.com/onnx/onnx/blob/master/docs/Operators.md#conv. onnx uses per-axis dilation

ksaurabh-cadence: Fair point. I made this change primarily to address dilation in a network coming from Caffe2 domain. I will modify the code to read dilation as a vector instead of scalar. I'm curious if anybody has encountered a case of dilation being two different values for different axis (x and y).

lib/Backends/CPU/Transforms.cpp
ksaurabh-cadence: Are all the comments supposed to end with period?  It was not apparent from many other comments.

ksaurabh-cadence:
> Why is the implementation of gradient nodes not changed?

I am not sure I understand this point, and why is it needed for this checkin? Can you point me the location where you were expecting changes regarding this? 

> What about C2 loader? Can we move loading logic into CommonOperatorLoader.h ?

Caffe2ModelLoader.cpp has been modified.

lib/Importer/ONNXModelLoader.cpp
jackm321: You probably don't have to change dilation to vector everywhere, you could just read it as a vector in onnx model loader and return an Error if the dimensions are not the same. This may be exactly what you're saying, just making sure.

lib/Backends/CPU/Transforms.cpp
jackm321: Yeah comments should be properly punctuated https://llvm.org/docs/CodingStandards.html#commenting

lib/Importer/ONNXModelLoader.cpp
ksaurabh-cadence: Yeah that's what I plan to do.

lib/Backends/CPU/Transforms.cpp
ksaurabh-cadence: Thanks for the link, that makes it clear.

SplitInfinity:
@ksaurabh-cadence 

This PR will break training of conv nodes as it is because dilation needs to be accounted for during training as well. For example, it is not enough for `libjit_convolution_f` to account for dilation, `libjit_convolution_grad_f` needs to account for it also so that convolutions with `dilation > 1` can train correctly.

In addition, by convention, what value for dilation is considered no dilation? I see you have implemented it with `dilation = 1` as no dilation, but I wonder if it should be 0.

ksaurabh-cadence:
> @ksaurabh-cadence
> 
> This PR will break training of conv nodes as it is because dilation needs to be accounted for during training as well. For example, it is not enough for `libjit_convolution_f` to account for dilation, `libjit_convolution_grad_f` needs to account for it also so that convolutions with `dilation > 1` can train correctly.

I have updated the code to address this. 
 
> In addition, by convention, what value for dilation is considered no dilation? I see you have implemented it with `dilation = 1` as no dilation, but I wonder if it should be 0.

No it should be 1 (same interpretation on default across the frameworks).  https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvInteger
https://github.com/pytorch/pytorch/blob/master/caffe2/operators/conv_op.cc

rdzhabarov:
> @rdzhabarov
Why is the implementation of gradient nodes not changed?
I am not sure I understand this point, and why is it needed for this checkin? Can you point me to the location where you were expecting changes regarding this?

That's what @SplitInfinity pointed out in the next comment.

ksaurabh-cadence:
@rdzhabarov

Your comments have been addressed with updated files. Please review and approve. Thanks,

include/glow/Graph/Graph.h
```diff
@@ -282,8 +282,8 @@ class Function final : public Named {
                               NodeValue filter, NodeValue bias, TypeRef outTy,
                               llvm::ArrayRef<unsigned_t> kernels,
                               llvm::ArrayRef<unsigned_t> strides,
-                              llvm::ArrayRef<unsigned_t> pads,
-                              unsigned_t group);
+                              llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
+                              unsigned_t dilation = 1);
```
rdzhabarov: please update doxygen with the new param.

include/glow/Graph/Graph.h
```diff
@@ -295,7 +295,8 @@ class Function final : public Named {
   ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
                               NodeValue filter, NodeValue bias, TypeRef outTy,
                               unsigned_t kernel, unsigned_t stride,
-                              unsigned_t pad, unsigned_t group);
+                              unsigned_t pad, unsigned_t group,
+                              unsigned_t dilation = 1);
```
rdzhabarov: doxygen update.

include/glow/Graph/Graph.h
```diff
@@ -933,8 +934,8 @@ class Function final : public Named {
                               size_t outChannels,
                               llvm::ArrayRef<unsigned_t> kernels,
                               llvm::ArrayRef<unsigned_t> strides,
-                              llvm::ArrayRef<unsigned_t> pads,
-                              unsigned_t group);
+                              llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
+                              unsigned_t dilation = 1);
```
rdzhabarov: here and in other places doxygen as well.

include/glow/Graph/Graph.h
ksaurabh-cadence: Update doxygen coments.

ksaurabh-cadence:
> This all looks great, there are few small things to take care about.
> Please, add loader tests as well (Caffe2ImporterTest).
> Also, add some tests for verifying grad nodes, see GradCheckTest.

Have updated Caffe and ONNX test model files to include dilation parameter to ensure that parsing is fine. 

GradCheckTest has been added.

PR #2935, opened 2019-05-17
[IROpt] Improve copy elimination optimisation

SplitInfinity:
**Description**
This commit improves the copy instruction elimination optimization in
the case that the src and dest operands are `tensorview` instructions.
Prior to this commit, this optimization used to eliminate copy
instructions if the src an dest `tensorview` offsets were equal. This does
not work for all cases; for example, offsets {0} and {0, 0} are not
programmatically equal, but are semantically equal.
   
This commit modifies this optimization to check that the linearized
offsets into the underlying buffers after peeling away all tensorviews
are equal before erasing the copy instruction. For example, the previous
implementation incorrectly eliminated a copy between identical offsets
into two different intermediate tensorviews that provide a view into
different sections of the same buffer. This commit addresses this.
    
This commit also modifies `verify()` for `Insert/ExtractTensorInst` and
`TensorViewInst` to check that the number of offsets in the instruction is equal to the size of the
src/dest shape. This simple check precludes the case in which offsets `{0}` and `{0,
0}` are compared by this optimization and determined not to be equal.
Code that creates these types of instructions has also been updated
accordingly.

**Testing**
This commit adds the case mentioned above with a copy between identical
offsets into two separate `tensorviews` that have different offsets into
the same underlying buffer as a unit test in IROptTest. All unit tests pass.

### DIFF
--- a/lib/IR/IRUtils.cpp
+++ b/lib/IR/IRUtils.cpp
@@ -51,3 +51,36 @@ const Value *glow::getOrigin(const Value *V) {
   }
   return V;
 }
+
+size_t glow::getOriginOffset(Value *V) {
+  size_t off = 0;
+
+  // Since each TensorView can either maintain or decrease the size of a buffer,
+  // the linearized offset of a TensorView into its underlying buffer can be
+  // computed by adding together the linearized offsets of intermediate
+  // TensorViews relative to their sources.
+  while (true) {
+    auto *TVI = dyn_cast<TensorViewInst>(V);
+    if (!TVI) {
+      return off;
+    }
+
+    llvm::ArrayRef<size_t> offsets = TVI->getOffsets();
+    llvm::ArrayRef<size_t> srcDims = TVI->getSrc()->getType()->dims();
+
+    size_t numSrcDims = srcDims.size();
+
+    // Iterate backwards in order to figure out how big the slices corresponding
+    // to each offset element are.
+    for (size_t i = 0, j = numSrcDims - 1; i < numSrcDims; ++i, --j) {
+      // For each offset, add into the linearized offset the offset value
+      // multiplied by the slice size corresponding to that offset.
+      off += offsets[j] * TVI->getSrc()->getType()->getSliceSize(j);
+    }
+
+    // Move on to the src of the TensorView that was just processed.
+    V = TVI->getSrc();
+  }
+
+  return off;
+}
--- a/lib/Optimizer/IROptimizer.cpp
+++ b/lib/Optimizer/IROptimizer.cpp
@@ -1597,9 +1597,7 @@ void performPeepholeOptimizations(IRFunction &M) {
           continue;
         }
 
-        auto *srcTV = dyn_cast<TensorViewInst>(src);
-        auto *destTV = dyn_cast<TensorViewInst>(dest);
-        if (srcTV && destTV && (srcTV->getOffsets() != destTV->getOffsets())) {
+        if (getOriginOffset(src) != getOriginOffset(dest)) {
           continue;
         }
 
--- a/tests/unittests/IROptTest.cpp
+++ b/tests/unittests/IROptTest.cpp
@@ -84,7 +84,7 @@ TEST(Optimizer, dseDoNotRemloveLastWriteIntoWeightVar) {
   bb.createElementAddInst("elem_add", output, input1, input2);
   bb.createTensorViewInst(
       "cast", output, mod.uniqueType(Type(glow::ElemKind::FloatTy, {1, 1, 1})),
-      {0, 0, 0});
+      {0});
 
   optimize(M, MockBackend().shouldShareBuffers());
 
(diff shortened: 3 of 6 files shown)


lib/IR/IRUtils.cpp
```diff
@@ -51,3 +51,43 @@ const Value *glow::getOrigin(const Value *V) {
...
+    auto *TVI = dyn_cast<TensorViewInst>(V);
+    if (!TVI) {
+      return off;
+    }
+
+    llvm::ArrayRef<size_t> offsets = TVI->getOffsets();
+    llvm::ArrayRef<size_t> srcDims = TVI->getSrc()->getType()->dims();
+
+    size_t numOffsets = offsets.size(), numSrcDims = srcDims.size();
+    size_t sliceSize = 1;
```
opti-mix: `Type` has a `getSliceSize` method. May be it could be useful here to shorten the code?

opti-mix:
@jfix71 Could you have a look at this PR?

lib/Optimizer/IROptimizer.cpp
```diff
@@ -1599,7 +1599,8 @@ void performPeepholeOptimizations(IRFunction &M) {
 
         auto *srcTV = dyn_cast<TensorViewInst>(src);
         auto *destTV = dyn_cast<TensorViewInst>(dest);
-        if (srcTV && destTV && (srcTV->getOffsets() != destTV->getOffsets())) {
+        if (srcTV && destTV &&
```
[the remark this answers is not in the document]
SplitInfinity: My interpretation is that you would like to see a test case that a copy from a `TensorView` to a non-`TensorView` (perhaps its underlying buffer) is detected and eliminated. I tried to construct a test case like that by creating `view2` from a `view1` of a `WeightVar` and a copy from `view2` into the `WeightVar` but this is being detected already without my diff by some other optimization...

SplitInfinity:
I took @jfix71 's suggestion and updated `TensorViewInst::verify` to check that the number of offset dims matches the dims of the tensorview src and that fixed the original problem that motivated this PR by forcing me to also fix the optimization that replaces `inserttensor` with `tensorview` to follow this rule. It turns out that's what produces an offset of just '{0}' in one case.

lib/Optimizer/IROptimizer.cpp
```diff
@@ -1366,9 +1366,15 @@ void optimizeInserts(IRFunction &M) {
 
     // Create a new TensorView of the original dest of the InsertTensor,
     // with the same offset as the InsertTensor.
+    std::vector<size_t> tvOffsets(insertDest->dims().size(), 0);
```
jfix71: So it seems this change is because `InsertTensorInsts` do not have their offsets set to the correct number of dimensions as well -- is that correct? So while this change seems sane, it also feels like it might be a bandaid instead of fixing the underlying issue. I.e. would it make sense to also add the same `verify()` check to `InsertTensorInst::verify()` and `ExtractTensorInst::verify()`, to root out the origin of mismatch on offsets/starts?

lib/IR/IRUtils.cpp
SplitInfinity: Ah I still need to do this, do not accept yet.

lib/Optimizer/IROptimizer.cpp
SplitInfinity: I think so - done.

lib/Optimizer/IROptimizer.cpp
```diff
@@ -1364,8 +1364,6 @@ void optimizeInserts(IRFunction &M) {
     assert(origWriter &&
            "Did not find the original writer to the source alloc.");
 
-    // Create a new TensorView of the original dest of the InsertTensor,
```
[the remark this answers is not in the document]
SplitInfinity: Oh, I thought I added it, my bad.

SplitInfinity:
Added a test that checks that a copy is eliminated if the offsets of tensorviews are different but linearized offset into the underlying buffer is the same and another that checks that copies from a tensorview to its underlying buffer are optimised out (but this test passes without my diff for some reason???).

SplitInfinity:
Also, I think I added a test for when the src of copy is tensorview and the dest is the underlying buffer. Should I add another with the buffer as the src and the tensorview as the dest?

tests/unittests/IROptTest.cpp
```diff
@@ -715,3 +715,168 @@ TEST(Optimizer, bufferReuseWithoutDefsPlusCasts) {
...
+  // tensorview are not the same.
+  EXPECT_EQ(instrs.size(), 5);
+  EXPECT_FALSE(std::none_of(
+      instrs.begin(), instrs.end(),
+      [](const Instruction &I) -> bool { return isa<CopyInst>(&I); }));
+  EXPECT_FALSE(std::none_of(
+      instrs.begin(), instrs.end(),
+      [](const Instruction &I) -> bool { return isa<TensorViewInst>(&I); }));
+
+  // Reset state for next test.
```
jfix71: nit: Instead of resetting, maybe just create a new test?

PR #2936, opened 2019-05-17
[Partitioner] Add helper function to dump a DAG after partitioning.

beicy:
Summary:
This PR dump a dot file from the partition result. Since now all functions belong to a function family and they have the same partition, we only dump the one function's partition. For each node, so far we show the function name and the backendKind, which are the most important info in my mind.  
The following is the Heterogeneous partition example:
dag1
Documentation:

Test Plan:

[Optional Fixes #issue]

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -143,6 +143,10 @@ class Partitioner {
   /// The backend pointers.
   std::vector<Backend *> backends_;
 
+  /// The map between BackendKind and its name. It is used for generating debug
+  /// info.
+  std::map<BackendKind, std::string> backendName_;
+
   /// The result of module partitioning.
   DAGListTy partitions_;
 
@@ -245,6 +249,11 @@ public:
 
   /// Get the partitions.
   DAGListTy &getPartitionResult() { return partitions_; }
+
+  /// Dump the partition result to a dot file. Since now all functions belong to
+  /// a function family and they have the same partition, we only dump the one
+  /// function's partition.
+  void dumpDAG(llvm::StringRef dotFilename) const;
 
   /// Get function for computeTime_
   ComputeTimeMapTy getComputeTime() const { return computeTime_; }
--- a/include/glow/Support/Support.h
+++ b/include/glow/Support/Support.h
@@ -58,6 +58,9 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
 /// The char '\n' becomes '\'+'n' and quotes are handled correctly.
 std::string escapeDottyString(const std::string &str);
 
+/// \returns the node color based on \p index which is used in dot file.
+const char *getDotFileNodeColor(size_t index);
+
 /// Add quotes to the string \p in.
 inline std::string quote(const std::string &in) { return '"' + in + '"'; }
 
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -15,6 +15,13 @@
  */
 
 #include "glow/Partitioner/Partitioner.h"
+#include "glow/Support/Support.h"
+
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <fstream>
 
 using namespace glow;
 using llvm::isa;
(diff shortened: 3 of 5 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -35,6 +42,65 @@ static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
   return ret;
 }
 
+void Partitioner::dumpDAG(llvm::StringRef dotFilename) {
+  if (partitions_.size() == 0)
+    return;
+  std::map<BackendKind, const char *> backendName = {
```
opti-mix: It seems like a bad idea to encode the names of all backends in the partitioner, because the set of backends may change over time. I think backends have a `getName()` method that you could use.

beicy: At this stage, only "BackendKind" is kept, that is why I encoded the name. I think I will add another class member to keep the BackendName then.

lib/Partitioner/Partitioner.cpp
```diff
@@ -35,6 +42,65 @@ static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
...
+      {BackendKind::Interpreter, "Interpreter"},
+      {BackendKind::OpenCL, "OpenCL"},
+      {BackendKind::CPU, "CPU"},
+      {BackendKind::Habana, "Habana"}};
+  std::map<BackendKind, const char *> backendColor = {
```
jfix71: Instead of mapping explicitly each Backend to colors, which means we must update this when we add a new Backend, you could do what we do when assigning colors to Nodes in `dumpNode()` in `Graph.cpp` -- create a large array of `colorNames` and then use `llvm::hash_value` to get an index into it for the Backend. Actually, this code could probably be moved from `dumpNode()` to a helper that takes a string and returns a color string, and then shared in both places.

include/glow/Partitioner/Partitioner.h
```diff
@@ -143,6 +143,10 @@ class Partitioner {
   /// The backend pointers.
   std::vector<Backend *> backends_;
 
+  /// The map between BackendKind and its name. It is used for generating debug
+  /// info.
+  std::map<BackendKind, std::string> backendName_;
```
opti-mix: Does it need to be an ordered map?

beicy: No. The order doesn't matter. We just want to get the name from the kind.

nickgg: Does it save much to have this map? Don't you have a Backend ptr you can just use getName on?

include/glow/Support/Support.h
```diff
@@ -68,6 +68,19 @@ inline std::string tolower(const std::string &in) {
   return data;
 }
 
+/// \return the color based on \p index which is used in dot file.
+inline const char *getColor(size_t index) {
```
nickgg: This needs a more descriptive name (getColor for what? Why would we get color?)

lib/Partitioner/Partitioner.cpp
```diff
@@ -738,6 +796,12 @@ llvm::Error Partitioner::Partition() {
   std::vector<std::unique_ptr<Backend>> backendHolder;
   getBackendMap(backendMap, backendHolder, backends);
 
+  // Assign the backend name to its BackendKind.
+  for (size_t i = 0, e = backends.size(); i < e; i++) {
+    backendName_[backends[i]->getBackendKind()] = backends[i]->getBackendName();
+    llvm::outs() << backendName_[backends[i]->getBackendKind()] << " end\n";
```
nickgg: Left some debugging in?

include/glow/Partitioner/Partitioner.h
beicy: @nickgg No. The backend ptr is local and only used in backendBasedPartition now. (Btw, the "backends_" ptrs are only used for users passing the "Mock" backends which are not registered in Glow. )

include/glow/Support/Support.h
```diff
@@ -68,6 +68,19 @@ inline std::string tolower(const std::string &in) {
   return data;
 }
 
+/// \return the color based on \p index which is used in dot file.
+inline const char *getDotFileColor(size_t index) {
```
opti-mix: May be call it `getDotFileNodeColor`? Files do not have colors ;-)

include/glow/Partitioner/Partitioner.h
```diff
@@ -246,6 +250,11 @@ class Partitioner {
   /// Get the partitions.
   DAGListTy &getPartitionResult() { return partitions_; }
 
+  /// Dump the partition result to a dot file. Since now all functions belong to
+  /// a function family and they have the same partition, we only dump the one
+  /// function's partition.
+  void dumpDAG(llvm::StringRef dotFilename);
```
opti-mix: Does this method modify anything? If not, make it const.

include/glow/Support/Support.h
opti-mix: And I'd suggest moving the implementation into the cpp file.

beicy: Done.

include/glow/Support/Support.h
```diff
@@ -58,6 +58,9 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
 /// The char '\n' becomes '\'+'n' and quotes are handled correctly.
 std::string escapeDottyString(const std::string &str);
 
+/// \return the node color based on \p index which is used in dot file.
```
opti-mix: The name of the comment is `\returns`, not `\return`.

lib/Support/Support.cpp
```diff
@@ -117,4 +117,17 @@ std::string legalizeName(llvm::StringRef name) {
   return legalName;
 }
 
+/// \return the color based on \p index which is used in dot file.
```
opti-mix: The name of the comment is `\returns`, not `\return`.

PR #2938, opened 2019-05-17
[Partitioner] increase the memory of the DummyDeviceManager

### LINKED ISSUE #2937: Glow fails due to the partitioner
@beicy I've just rebased our work on top of the latest Glow master branch and it doesn't work anymore with our backend when calling `image-classifier` (we use the `-<backend>` command line option to specify our backend).

I get this strange error.
```
Encountered an error, exiting.
location: /home/tlepley/git/glow/lib/Partitioner/Partitioner.cpp:796 message: Partition failed: the number of given (CPU) devices (1) is fewer than the required minimal partitions (76).
```
Note that we use the  default `DummyDeviceManager`.

I'm currently investigating, but I didn't find yet the source of the problem.

tlepley-cadence:
Summary: Set the memory of `DummyDeviceManager` to the maximum value in order to avoid triggering any partitioning by default.

Documentation: N/A

Test Plan: N/A

Fixes #2937

### DIFF
--- a/include/glow/Backends/DummyDeviceManager.h
+++ b/include/glow/Backends/DummyDeviceManager.h
@@ -38,8 +38,12 @@ public:
 
   /// The DummyDeviceManager is a simple wrapper for testing, if you need
   /// memory guards you should implement a DeviceManager for your device.
-  uint64_t getMaximumMemory() const override { return 100; }
-  uint64_t getAvailableMemory() const override { return 100; }
+  uint64_t getMaximumMemory() const override {
+    return std::numeric_limits<uint64_t>::max();
+  }
+  uint64_t getAvailableMemory() const override {
+    return std::numeric_limits<uint64_t>::max();
+  }
   bool isMemoryAvailable(uint64_t) const override { return true; }
 
   /// Load the provided module into the device, readyCB will be called when


opti-mix:
cc @beicy

beicy:
The DummyDeviceManager is a simple wrapper for testing. I think maybe it is better to add some command option similar to this PR to control the memory limitation: https://github.com/pytorch/glow/pull/2710

cc @gcatron

tlepley-cadence:
@beicy Having some options to control memory can be interesting, but I would prefer this to be handle in a second step by someone who needs this functionality. The purpose of this PR is simply to fix `DummyDeviceManager` that is not usable anymore due to the too small declared memory. Note that we don't need the partitioner / device managers, so we are happy with a `DummyDeviceManager` that 'disables' the partitioner by default.

beicy:
> @beicy Having some options to control memory can be interesting, but I would prefer this to be handle in a second step by someone who needs this functionality. The purpose of this PR is simply to fix `DummyDeviceManager` that is not usable anymore due to the too small declared memory. Note that we don't need the partitioner / device managers, so we are happy with a `DummyDeviceManager` that 'disables' the partitioner by default.

 It is OK to change the memory now. I am not sure in the future users still need to change this size since it is possible that their backend  doesn't have so much memory.

Issue #1699, opened 2018-09-24
[New operator] LengthsRangeFill


artemrakhov:
We need to support `LengthsRangeFill` operator from Caffe2. Spec can be found here:
https://caffe2.ai/docs/operators-catalogue.html#lengthsrangefill

This operator requires dynamic memory allocation. Size of the output is the sum of all elements of the input.

qcolombet:
Two questions:
1. Is the input known at compile time or could it come from an activation?
2. What type of network uses this? Put differently, how important is that operator?

jfix71:
@jackm321 Still working/going to work on this, or can I take it from you?

jackm321:
@jfix71 all yours

PR #2941, opened 2019-05-18
[New Operator] Add support LengthsRangeFill operator

jfix71:
Summary: Add support LengthsRangeFill operator. Added Interpreter and CPU support. Note that this operator assumes a `maxOutputSize` is defined in the proto, as discussed in https://github.com/pytorch/glow/issues/1699.

Test Plan: Added tests in OperatorTest and Caffe2ImporterTest

Fixes https://github.com/pytorch/glow/issues/1699

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1049,6 +1049,22 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
     return llvm::Error::success();
   }
 
+  if (typeName == "LengthsRangeFill") {
+    NodeValue lengths;
+    ASSIGN_VALUE_OR_RETURN_ERR(lengths, getNodeValueByName(op.input(0)));
+    RETURN_ERR_IF_NOT(lengths.dims().size() == 1,
+                      "lengths must be a 1D vector.");
+
+    unsigned_t maxOutputSize;
+    ASSIGN_VALUE_OR_RETURN_ERR(maxOutputSize,
+                               loadInt(dict.find("maxOutputSize")->second));
+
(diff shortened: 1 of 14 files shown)


beicy:
If I understand it correctly, the "maxOutputSize" is not the exact size that the output should be.  But looks like here we use it as the size of the output. So once it is used as the input of the following ops, how do that op know the real size?

jfix71:
@beicy It will depend on the graph itself to understand it. For example there could be an external input representing the sum of `Lengths`, or there could be a `BatchedReduceAdd` that also loads `Lengths`. GatherRanges does something similar, specifying a max output size as well.

beicy:
I think the PR is good. But there are 2 test failures in Phabricator. Could you please double check it?

jfix71:
@beicy All good from Phabricator.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1049,6 +1049,22 @@ llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
...
 
+  if (typeName == "LengthsRangeFill") {
+    NodeValue lengths;
+    ASSIGN_VALUE_OR_RETURN_ERR(lengths, getNodeValueByName(op.input(0)));
+    RETURN_ERR_IF_NOT(lengths.dims().size() == 1,
+                      "lengths must be a 1D vector.");
+
+    unsigned_t maxOutputSize;
+    ASSIGN_VALUE_OR_RETURN_ERR(maxOutputSize,
+                               loadInt(dict.find("maxOutputSize")->second));
```
rdzhabarov: dict.find("maxOutputSize")->second if maxOutputSize does not exist would likely be a seg fault.

PR #2942, opened 2019-05-20
add a test that compares SLSW between c2 and H

hyuen:
Summary:
add a test that tries out SLS in the cpu as well as H and verifies it is fine

found a bug in Glow about input parameters, when there are none, it is ok for the input param to be none

Reviewed By: benjibc

Differential Revision: D15410717

### DIFF
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -18,6 +18,7 @@
 #include "GlowOnnxifiManager.h"
 
 #include "glow/Importer/ONNXIFIModelLoader.h"
+#include <glog/logging.h>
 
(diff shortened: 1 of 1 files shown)


rdzhabarov:
> found a bug in Glow about input parameters, when there are none, it is ok for the input param to be none

I think you basically neven will have onnxifi node with 0 inputs practically.

@yinghai is it possible to have onnxifi node with 0 inputs?

yinghai:
If we have `FillTensor` related ops, it might be possible. But in practical usage, I don't see much cases like this in predict net.

rdzhabarov:
closing this, as we always expect to have inputs/outputs. If those are not present we just bail from the onnxifi call which is fine.

PR #2943, opened 2019-05-20
Make stalebot less aggressive

bertmaher:
Summary: Five days is pretty aggressive, let's do 30.

Test Plan: n/a


nickgg:
Like the idea, 30 days seems like a lot though.

bertmaher:
15 it  is!

PR #2945, opened 2019-05-20
External storage for CPUDeviceManager memory

bertmaher:
Summary: Using external storage for this makes it easier to set... externally.

Differential Revision: D15406548

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -19,24 +19,22 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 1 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -19,24 +19,22 @@
...
-
-using namespace glow;
-using namespace glow::runtime;
-
 namespace glow {
 namespace runtime {
+
+uint64_t GlowCPUMemory = 0;
+
+static llvm::cl::opt<uint64_t, true>
```
jfix71: nit: `/* ExternalStorage */ true`

bertmaher:
> If we're going to do this for all command line options, can we do some macro magic to take care of the details for us?

Yeah, maybe.  I kind of didn't want to add too many of these because the approach we're using inside FB to route through gflags felt kind of squirrelly, but I haven't come up with anything better...

PR #2950, opened 2019-05-21
Refactoring Glow loaders to support training. Part I.

putivsky:
Summary: Glow loaders has been designed only for inference, changing internals to support training as well.

Reviewed By: artemrakhov

Differential Revision: D15418551


jackm321:
It would be nice to export diffs to GH before they are already landing so that more people can participate in review

PR #2956, opened 2019-05-22
[Backends] Change command line memory options to kilobytes

jfix71:
CI is broken because the `en2gr_cpu_partition_test` tries to set the CPU memory on the command line, but it seems that command line parameters don't work with `uint64_t` on some platforms. I've hit this issue in the past, and I'm not sure why it's the case.

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -22,16 +22,17 @@
 namespace glow {
 namespace runtime {
 
-uint64_t GlowCPUMemory = 0;
+unsigned GlowCPUMemory = 0;
 
-static llvm::cl::opt<uint64_t, /* ExternalStorage */ true>
-    GlowCPUMemoryOpt("cpu-memory",
-                     llvm::cl::desc("CPU DeviceManager maximum memory"),
-                     llvm::cl::location(GlowCPUMemory));
+static llvm::cl::opt<unsigned, /* ExternalStorage */ true> GlowCPUMemoryOpt(
+    "cpu-memory",
+    llvm::cl::desc("CPU DeviceManager maximum memory in kilobytes."),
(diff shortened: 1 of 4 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -22,16 +22,17 @@
...
+static llvm::cl::opt<unsigned, /* ExternalStorage */ true> GlowCPUMemoryOpt(
+    "cpu-memory",
+    llvm::cl::desc("CPU DeviceManager maximum memory in kilobytes."),
+    llvm::cl::location(GlowCPUMemory));
 
 DeviceManager *createCPUDeviceManager(std::unique_ptr<DeviceConfig> config) {
   if (GlowCPUMemory) {
-    return new CPUDeviceManager(std::move(config), GlowCPUMemory);
+    // Convert command line GlowCPUMemory to bytes from kilobytes.
+    return new CPUDeviceManager(std::move(config), GlowCPUMemory * 1000);
```
bertmaher: Should we use 1024 here, just to be precise?

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -32,7 +32,9 @@ namespace runtime {
 DeviceManager *
 createInterpreterDeviceManager(std::unique_ptr<DeviceConfig> config) {
   if (interpreterMaxMem) {
-    return new InterpreterDeviceManager(std::move(config), interpreterMaxMem);
+    // Convert command line interpreterMaxMem to bytes from kilobytes.
+    return new InterpreterDeviceManager(std::move(config),
+                                        interpreterMaxMem * 1000);
```
bertmaher: Same q about 1024.

jfix71: Yep updated both.

PR #2958, opened 2019-05-22
[OpenCL] Implement BatchedReduceAdd for arbitrary axes

SplitInfinity:
**Description**
This commit extends the implementation of `BatchedReduceAdd` in the OpenCL backend
so that it can handle any reduction axis, not just 1. This can be useful
when the first dimension is the batch dimension, and a reduction needs
to be performed within each example.

The existing implementation for axis = 0 computes each slice element in
parallel and linearizes the slice for simplicity (i.e. creates a 1D
global workspace as large as the number of elements in the output). This
implementation generalizes this concept by creating a global workspace
with rank equal to the number of dimensions of the output and computing
each one in parallel. The slice sizes of the input and output shapes are
are precomputed on the host and passed in as kernel arguments so that
the kernel can compute the correct offsets into the input and output
buffer by multiplying its set of global IDs with those slice sizes.

**Test Plan**
This commit enables the existing non-zero axis `BatchedReduceAdd` unit
test for OpenCL and modifies it test `axis=2`. All unit tests pass.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -913,18 +913,92 @@ llvm::Error OpenCLFunction::execute(ExecutionContext *context) {
     }
 
     if (auto *BRA = dyn_cast<BatchedReduceAddInst>(&I)) {
-      assert(BRA->getAxis() == 0 && "No current support for non-zero axis.");
+      auto axis = BRA->getAxis();
 
+      // Determine and store the slice sizes of each input dimension excluding
+      // the reduce axis into batchSliceSizes. These are used by the kernel to
+      // index correctly into the input buffer. If the input has one dimension
+      // (that is also the reduce axis), store one slice of size 1 into
+      // batchSliceSizes.
+      auto batchDims = BRA->getBatch()->getType()->dims();
+      auto numBatchDims = batchDims.size();
+      std::vector<size_t> batchSliceSizes(
+          numBatchDims > 1 ? numBatchDims - 1 : 1, 1);
+      size_t currentSliceSize = 1, axisSliceSize = 1;
+      for (size_t i = batchSliceSizes.size() - 1, j = i; i >= 0; ++i) {
+        if (i == axis) {
+          axisSliceSize = currentSliceSize;
+        } else {
+          batchSliceSizes[j--] = currentSliceSize;
+        }
+
+        currentSliceSize *= batchDims[i];
+      }
+
+      // Determine and store the slice sizes of each output dimension excluding
+      // the reduce axis into destSliceSizes. These are used by the kernel to
+      // index correctly into the output buffer. If the output has zero
+      // dimensions store one slice of size 1 into destSliceSizes.
+      auto destDims = BRA->getDest()->getType()->dims();
+      std::vector<size_t> destDimsVec(destDims.begin(), destDims.end());
+      if (destDims.empty()) {
+        destDimsVec.emplace_back(1);
+      }
+      auto numDestDims = destDimsVec.size();
+      std::vector<size_t> destSliceSizes(numDestDims > 0 ? numDestDims : 1, 1);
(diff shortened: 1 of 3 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -912,18 +912,92 @@ void OpenCLFunction::execute(ExecutionContext *context) {
...
+
+      // Allocate device buffers for batchSliceSizes and destSliceSizes.
+      size_t batchSlicesBufSize = batchSliceSizes.size() * sizeof(size_t);
+      size_t destSlicesBufSize = destSliceSizes.size() * sizeof(size_t);
+      cl_mem batchSlicesBuf = allocDeviceBuffer(batchSlicesBufSize);
+      cl_mem destSlicesBuf = allocDeviceBuffer(destSlicesBufSize);
+
+      // Copy batchSliceSizes and destSliceSizes from host to device.
+      cl_event writeBatchSlicesEvent{nullptr}, writeDestSlicesEvent{nullptr};
+      cl_int err = clEnqueueWriteBuffer(
```
SplitInfinity: I think this write can be nonblocking and waited on later.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -912,18 +912,92 @@ void OpenCLFunction::execute(ExecutionContext *context) {
...
+          /* event_list */ nullptr,
+          /* event */ kernelProfiling_ ? &writeBatchSlicesEvent : nullptr);
+      GLOW_ASSERT(err == CL_SUCCESS && "Unable to copy BRA data to the device");
+      if (kernelProfiling_) {
+        kernelLaunches_.emplace_back(KernelLaunch("batchedReduceAddSliceData",
+                                                  "batchedReduceAddSliceData",
+                                                  writeBatchSlicesEvent));
+      }
+
+      err = clEnqueueWriteBuffer(
```
SplitInfinity: Same here.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -913,18 +913,92 @@ llvm::Error OpenCLFunction::execute(ExecutionContext *context) {
...
+      // the reduce axis into batchSliceSizes. These are used by the kernel to
+      // index correctly into the input buffer. If the input has one dimension
+      // (that is also the reduce axis), store one slice of size 1 into
+      // batchSliceSizes.
+      auto batchDims = BRA->getBatch()->getType()->dims();
+      auto numBatchDims = batchDims.size();
+      std::vector<size_t> batchSliceSizes(
+          numBatchDims > 1 ? numBatchDims - 1 : 1, 1);
+      size_t currentSliceSize = 1, axisSliceSize = 1;
+      for (size_t i = 1, j = batchSliceSizes.size() - 1, e = batchDims.size();
```
nickgg: Variable names are a bit terse here, would prefer `end` and `index`.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -913,18 +913,92 @@ llvm::Error OpenCLFunction::execute(ExecutionContext *context) {
...
+           i <= e; ++i) {
+        size_t ind = e - i;
```
nickgg: Couldn't you just count down rather than count up than subtract from the end?

lib/Backends/OpenCL/OpenCL.cpp
nickgg: Are you planning to do this in this PR or a future one?

SplitInfinity: This PR

SplitInfinity:
`backedReduceAddWithAxis_Float` tests only `axis=1` so I modified to test `axis=2` (which in this case is the last axis, another thing that would be good to test). We have other tests for `axis=0` and to test the case that the result is zero-dimensional, so I'm reasonably confident now.

PR #2959, opened 2019-05-23
[DeviceConfig] add device memory size configuration to DeviceConfig

### LINKED ISSUE #2899: [runtime] add device memory size configuration to the DeviceConfig::parameters map on all backends
Since #2874 we have a mechanism for passing through specific options to DeviceManagers through the HostManager. One obvious configuration we may want across all devices is the ability to set the amount of available memory (useful for configuring partitioning).

Add support for Device size in the DeviceConfig which configures device size in all existing DeviceManagers (Interpreter, CPU, OpenCL & Habana).

superwizard2019:
Summary:

- Add deviceMemory member to DeviceConfig struct.
- Add  `setDeviceMemory()` and `getDeviceMemory()` to set (in bytes) and get the deviceMemory the `getDeviceMemory()` will return 0 by default.
- Add corresponding changes to backend (CPU, Interpreter, OpenCL and Habana). 
  - The CPU, Interpreter and Habana backends have options to use commandLine input as the deviceMemory. The commandLine should have higher priority than deviceConfig. 
  - The OpenCL backend used device info to get the maximum deviceMemory. Here the deviceConfig should have higher priority to limit the memory usage.

Test:
- Add test under the `DeviceManagerTest` to set then get device memory size from the
`DeviceConfig`. 
- Add test to DeviceManagerTest(CPU, Interpreter, OpenCL) to guarantee the priority logic mentioned above.
- Habana test is conducted on devgpu058.ash6.facebook.com. The test is included in the HabanaBackendTest.

Fixes:
This PR fixes #2899.

### DIFF
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -130,7 +130,8 @@ struct DeviceConfig {
   const BackendKind backendKind;
   /// A human readable name to identify the device.
   std::string name;
-
+  /// Device memory size in bytes.
+  uint64_t deviceMemory = 0;
   /// A map of configuration parameters.
   llvm::StringMap<std::string> parameters{};
 
@@ -143,6 +144,14 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceMemory(uint64_t memSize) { deviceMemory = memSize; }
+
+  uint64_t getDeviceMemory() const { return deviceMemory; }
+
+  uint64_t getDeviceMemory(uint64_t defaultMemory) const {
+    return deviceMemory == 0 ? defaultMemory : deviceMemory;
+  }
 };
 
 } // namespace runtime
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -32,7 +32,9 @@ static llvm::cl::opt<unsigned, /* ExternalStorage */ true> GlowCPUMemoryOpt(
 DeviceManager *createCPUDeviceManager(const DeviceConfig &config) {
   if (GlowCPUMemory) {
     // Convert command line GlowCPUMemory to bytes from kilobytes.
-    return new CPUDeviceManager(config, uint64_t{GlowCPUMemory} * 1024);
+    auto configNew = config;
+    configNew.setDeviceMemory(uint64_t{GlowCPUMemory} * 1024);
+    return new CPUDeviceManager(configNew);
   }
   return new CPUDeviceManager(config);
 }
--- a/lib/Backends/CPU/CPUDeviceManager.h
+++ b/lib/Backends/CPU/CPUDeviceManager.h
@@ -40,8 +40,9 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
   const uint64_t functionCost_{1};
 
 public:
-  CPUDeviceManager(const DeviceConfig &config, size_t maxMemory = 2000000000)
-      : QueueBackedDeviceManager(config), maxMemoryBytes_(maxMemory) {}
+  CPUDeviceManager(const DeviceConfig &config)
+      : QueueBackedDeviceManager(config),
+        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {}
 
   /// Returns the amount of memory in bytes available on the device when no
   /// models are loaded.
--- a/lib/Backends/Habana/HabanaDeviceManager.cpp
+++ b/lib/Backends/Habana/HabanaDeviceManager.cpp
@@ -114,8 +114,15 @@ llvm::Error HabanaDeviceManager::init() {
 llvm::Error HabanaDeviceManager::updateMemoryUsage() {
   // TODO: Use synGetMemInfo once implemented.
 
-  totalMemory_ = uint64_t{GlowHabanaMemory} * 1024;
-  freeMemory_ = uint64_t{GlowHabanaMemory} * 1024;
+  // Use GlowHabanaMemory if it is defined from GFLAGS or llvm params,
+  // otherwise, fall back to what config says.
+  uint64_t defaultMemory = 7 << 20;
+  if (GlowHabanaMemory == defaultMemory && config_.getDeviceMemory() != 0) {
+    totalMemory_ = config_.getDeviceMemory();
+  } else {
+    totalMemory_ = uint64_t{GlowHabanaMemory} * 1024;
+  }
+  freeMemory_ = totalMemory_;
 
   // Account for the size used by each function loaded on the card.
   for (const auto &pr : functions_) {
--- a/lib/Backends/Interpreter/InterpreterDeviceManager.h
+++ b/lib/Backends/Interpreter/InterpreterDeviceManager.h
@@ -40,9 +40,9 @@ class InterpreterDeviceManager : public QueueBackedDeviceManager {
   const uint64_t functionCost_{1};
 
 public:
-  InterpreterDeviceManager(const DeviceConfig &config,
-                           size_t maxMemory = 2000000000)
-      : QueueBackedDeviceManager(config), maxMemoryBytes_(maxMemory) {}
+  InterpreterDeviceManager(const DeviceConfig &config)
+      : QueueBackedDeviceManager(config),
+        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {}
 
   /// Returns the amount of memory in bytes available on the device when no
   /// models are loaded.
--- a/tests/unittests/DeviceManagerTest.cpp
+++ b/tests/unittests/DeviceManagerTest.cpp
@@ -18,8 +18,10 @@
 #include "glow/Backends/DummyDeviceManager.h"
 
 #include "../../lib/Backends/CPU/CPUDeviceManager.h"
+#include "../../lib/Backends/Interpreter/InterpreterDeviceManager.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/Optimizer/Optimizer.h"
+#include "glow/Runtime/RuntimeTypes.h"
 
 #include "gtest/gtest.h"
 
@@ -549,13 +551,30 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_TRUE(result2->isEqual(output2));
 }
 
+TEST(DeviceManagerTest, SetDeviceMemory) {
+  // Test Interpreter.
+  auto interpreterConfigEmpty = DeviceConfig(BackendKind::Interpreter);
+  auto interpreterConfigFull = DeviceConfig(BackendKind::Interpreter);
+  interpreterConfigFull.setDeviceMemory(32768);
+  // Only deviceConfig setting.
+  InterpreterDeviceManager interpreterDeviceSetByDeviceConfig(
+      interpreterConfigFull);
+  EXPECT_EQ(interpreterDeviceSetByDeviceConfig.getMaximumMemory(), 32768);
+  // No setting at all, default memory size.
+  InterpreterDeviceManager interpreterDeviceDefault(interpreterConfigEmpty);
+  EXPECT_EQ(interpreterDeviceDefault.getMaximumMemory(), 2000000000);
+}
+
 #ifdef GLOW_WITH_CPU
 
 TEST(DeviceManagerTest, AvailableMemory) {
   std::vector<std::unique_ptr<CompiledFunction>> backing;
   std::promise<const Module *> promise;
   std::future<const Module *> future;
-  CPUDeviceManager cpuCoreDevice(DeviceConfig(BackendKind::CPU), 1);
+
+  auto config = DeviceConfig(BackendKind::CPU);
+  config.setDeviceMemory(1);
+  CPUDeviceManager cpuCoreDevice(config);
   ASSERT_FALSE(errToBool(cpuCoreDevice.init()));
 
   uint64_t expectedBytes = 1;
@@ -624,6 +643,17 @@ TEST(DeviceManagerTest, AvailableMemory) {
   EXPECT_EQ(cpuCoreDevice.getAvailableMemory(), 0);
 
   EXPECT_FALSE(errToBool(cpuCoreDevice.stop()));
+
+  // Test CPU DeviceConfig.
+  auto cpuConfigEmpty = DeviceConfig(BackendKind::CPU);
+  auto cpuConfigFull = DeviceConfig(BackendKind::CPU);
+  cpuConfigFull.setDeviceMemory(32768);
+  // Only deviceConfig setting.
+  CPUDeviceManager cpuDeviceSetByDeviceConfig(cpuConfigFull);
+  EXPECT_EQ(cpuDeviceSetByDeviceConfig.getMaximumMemory(), 32768);
+  // No setting at all, default memory size.
+  CPUDeviceManager cpuDeviceDefault(cpuConfigEmpty);
+  EXPECT_EQ(cpuDeviceDefault.getMaximumMemory(), 2000000000);
 }
 
 TEST(DeviceManagerTest, DummyDeviceManager) {
--- a/tests/unittests/HabanaBackendTest.cpp
+++ b/tests/unittests/HabanaBackendTest.cpp
@@ -375,6 +375,21 @@ TEST_F(HabanaBackendTest, FuseConvAddRelu) {
   EXPECT_EQ(F_->getNodes().size(), 4);
 }
 
+TEST_F(HabanaBackendTest, SetDeviceMemory) {
+  uint64_t defaultMemory = (7 << 20);
+  auto configEmpty = glow::runtime::DeviceConfig(BackendKind::Habana);
+  auto configFull = glow::runtime::DeviceConfig(BackendKind::Habana);
+  configFull.setDeviceMemory(32768);
+  // With no commandline or deviceConfig, the memory should be default 7 <<20.
+  glow::runtime::HabanaDeviceManager device1(configEmpty, 1, 1);
+  llvm::Error err1 = device1.init();
+  EXPECT_EQ(defaultMemory * 1024, device1.getMaximumMemory());
+  // With only deviceConfig, the memory should be set by deviceConfig.
+  glow::runtime::HabanaDeviceManager device2(configFull, 1, 1);
+  llvm::Error err2 = device2.init();
+  EXPECT_EQ(32768, device2.getMaximumMemory());
+}
+
 TEST_F(HabanaBackendTest, ConvertFC) {
   HabanaBackend backend;
   auto *input =
--- a/lib/Backends/DeviceManagers.cpp
+++ b/lib/Backends/DeviceManagers.cpp
@@ -16,7 +16,6 @@
 
 #include "glow/Backends/DeviceManager.h"
 #include "glow/Backends/DummyDeviceManager.h"
-
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/raw_ostream.h"
 
--- a/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
+++ b/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
@@ -32,8 +32,9 @@ namespace runtime {
 DeviceManager *createInterpreterDeviceManager(const DeviceConfig &config) {
   if (interpreterMaxMem) {
     // Convert command line interpreterMaxMem to bytes from kilobytes.
-    return new InterpreterDeviceManager(config,
-                                        uint64_t{interpreterMaxMem} * 1024);
+    auto configNew = config;
+    configNew.setDeviceMemory(uint64_t{interpreterMaxMem} * 1024);
+    return new InterpreterDeviceManager(configNew);
   }
   return new InterpreterDeviceManager(config);
 }
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -149,7 +149,13 @@ llvm::Error OpenCLDeviceManager::init() {
   if (err != CL_SUCCESS) {
     RETURN_ERR("Error getting device memory limit");
   }
-  maxMemoryBytes_ = mem_size;
+
+  // If limited by deviceConfig, should allow less deviceMemory
+  if (config_.getDeviceMemory() != 0 && config_.getDeviceMemory() < mem_size) {
+    maxMemoryBytes_ = config_.getDeviceMemory();
+  } else {
+    maxMemoryBytes_ = mem_size;
+  }
 
   return llvm::Error::success();
 }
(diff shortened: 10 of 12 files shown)


include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,18 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceSize(uint64_t mem_size) {
+    parameters["device_size"] = std::to_string(mem_size);
+  }
+
+  uint64_t getDeviceSize() {
```
SplitInfinity: Getters are usually `const`:

`uint64_t getDeviceSize() const {...`

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,18 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceSize(uint64_t mem_size) {
```
SplitInfinity: We use camel case for variable names, so `mem_size` should be `memSize`.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,18 @@ struct DeviceConfig {
...
+    parameters["device_size"] = std::to_string(mem_size);
+  }
+
+  uint64_t getDeviceSize() {
+    uint64_t ret = 0;
+    if (parameters.find("device_size") != parameters.end()) {
+      ret = std::stoull(parameters["device_size"]);
```
SplitInfinity: Instead of performing another lookup if the key is in the map, save the iterator returned by `find` and use that in the body of the if statement:

```
auto it = parameters.find("device_size");
if (it != parameters.end()) {
      ret = std::stoull(it->second);
}
```

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -481,6 +482,18 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_TRUE(result2->isEqual(output2));
 }
 
+TEST(Runtimetest, SetDeviceSize) {
```
SplitInfinity: Variables in this test should be camel case as well.

bertmaher: Every other test case in this file is part of `DeviceManagerTest`, mind using that instead of `Runtimetest`?

include/glow/Runtime/RuntimeTypes.h
bertmaher: Let's call these `{set,get}DeviceMemory` instead of `Size`, since "size" is a bit ambiguous.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,18 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceSize(uint64_t mem_size) {
+    parameters["device_size"] = std::to_string(mem_size);
```
bertmaher: Same `size` -> `memory` for the parameter key.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,19 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceMemory(uint64_t memSize) {
+    parameters["deviceMemory"] = std::to_string(memSize);
```
rdzhabarov: do we expect some of the devices do not need to have deviceMemory?

Specifically why is it in the map and not the member of the DeviceConfig.

superwizard2019: @nickgg could you clarify this : )

superwizard2019: already moved the deviceMemory out to be a member of the deviceConfig.
Thanks :)

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +143,13 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  // Device memory size.
+  uint64_t deviceMemory = 0;
```
rdzhabarov: nit: can you group that to other members above
```
/// An enum indicating what kind of backend this config is for. It is used in
  /// checking the type of config before casting to a derived class.
  const BackendKind backendKind;
  /// A human readable name to identify the device.
  std::string name;
...
```

Also, /// not // for goxygen comment.

lib/Backends/Interpreter/InterpreterDeviceManager.h
```diff
@@ -42,9 +42,14 @@ class InterpreterDeviceManager : public QueueBackedDeviceManager {
 public:
   InterpreterDeviceManager(std::unique_ptr<DeviceConfig> config = nullptr,
                            size_t maxMemory = 2000000000)
-      : QueueBackedDeviceManager(BackendKind::Interpreter, std::move(config)),
-        maxMemoryBytes_(maxMemory) {}
-
+      : QueueBackedDeviceManager(BackendKind::Interpreter, std::move(config)) {
+    if (maxMemory == 2000000000 && config_ != nullptr &&
```
rdzhabarov: for brevity we could use, config_->getDeviceMemory(/*default*/maxMemory) and avoid all these checks across all device managers.

also, what are the scenarios when we create device manager with nullptr config?

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -481,6 +486,86 @@ TEST_P(DeviceManagerTest, ReuseModule) {
...
+
+  // Never set a value.
+  auto configEmpty = llvm::make_unique<DeviceConfig>(BackendKind::CPU);
+  EXPECT_EQ(configEmpty->getDeviceMemory(), 0);
+  // Has set a value.
+  auto configFull = llvm::make_unique<DeviceConfig>(BackendKind::CPU);
+  configFull->setDeviceMemory(32768);
+  EXPECT_EQ(configFull->getDeviceMemory(), 32768);
+
+#ifdef GLOW_WITH_CPU
```
rdzhabarov: Can this be done as TEST_P? look at other examples how that could help handling different backend kinds.
I'd prefer to limit usage of #ifdefs here and there.

lib/Backends/Interpreter/InterpreterDeviceManager.h
superwizard2019: > for brevity we could use, config_->getDeviceMemory(/_default_/maxMemory) and avoid all these checks across all device managers.
> 
> also, what are the scenarios when we create device manager with nullptr config?

I saw sometimes the `createDeviceManger` is called with only a backendkind parameter. So I thought the device config might be set to nullptr which is its default value.  Should I keep this or just ignore the nullptr?.  Also could you clarify the ` config_->getDeviceMemory(/default/maxMemory)` :)?

include/glow/Runtime/RuntimeTypes.h
superwizard2019: > nit: can you group that to other members above
> 
> ```
> /// An enum indicating what kind of backend this config is for. It is used in
>   /// checking the type of config before casting to a derived class.
>   const BackendKind backendKind;
>   /// A human readable name to identify the device.
>   std::string name;
> ...
> ```
> 
> Also, /// not // for goxygen comment.

OK. I will do that

tests/unittests/DeviceManagerTest.cpp
superwizard2019: > Can this be done as TEST_P? look at other examples how that could help handling different backend kinds.
> I'd prefer to limit usage of #ifdefs here and there.

I find that the test logic of those different backends are little bit different. To remove the `#ifdef`s, I moved them to different test files.  Should I merge the CPU and Interpreter test code into one TEST_P?

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -42,8 +42,14 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
 public:
   CPUDeviceManager(std::unique_ptr<DeviceConfig> config = nullptr,
                    size_t maxMemory = 2000000000)
-      : QueueBackedDeviceManager(BackendKind::CPU, std::move(config)),
-        maxMemoryBytes_(maxMemory) {}
+      : QueueBackedDeviceManager(BackendKind::CPU, std::move(config)) {
+    if (maxMemory == 2000000000 && config_ != nullptr &&
```
nickgg: If this is just a sentinel now, we should pick a more obvious value (e.g. zero).

lib/Backends/Interpreter/InterpreterDeviceManager.h
superwizard2019: > for brevity we could use, config_->getDeviceMemory(/_default_/maxMemory) and avoid all these checks across all device managers.
> 
> also, what are the scenarios when we create device manager with nullptr config?

OK,  I merged all the backend if/else branches into the config_->getDeviceMemory() function. 
As we discussed this morning, I still check if config_ is a nullptr.

tests/unittests/DeviceManagerTest.cpp
superwizard2019: > Can this be done as TEST_P? look at other examples how that could help handling different backend kinds.
> I'd prefer to limit usage of #ifdefs here and there.

I've looked through all backend tests. I think to merge those tests into one parameterized (TEST_P), there are a lot of extra functions required to implement. 
For example, the CPU and Interpreter backend will read the command line option as their default memory, which is not exposed to our test. The default createdevicemanager() does not accept a default memory size as its parameter. Also, those different backends have different default memory size, so the test should verify separately. The CPU and interpreter backend should prefer command line option rather than the deviceConfig. On the other side, the OpenCL backend will check the actual device memory size using OpenCL API.  If it is larger than deviceConfig memory size, we should prefer to use the size limited by deviceConfig. 
The habana default memory size is in KB and others are in Bytes.  Overall, to merge them into one Test_P, we still need to write may branches for different backend anyway without implementing a lot of extra helper functions. Could we just stay with current Test for now (I've moved those tests to different test file for different backends. So the #ifdefs have been removed) and maybe merge them later?

rdzhabarov:
> Please remember to squash your commits before landing.

not necessary since "land to fbsource" will squash all together.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -143,6 +144,55 @@ struct DeviceConfig {
       : backendKind(kind), name(name), parameters(parameters) {}
 
   bool hasName() const { return name != ""; }
+
+  void setDeviceMemory(uint64_t memSize) { deviceMemory = memSize; }
+
+  uint64_t getDeviceMemory() const { return deviceMemory; }
+
+  /// Handle memory setting priorites for all backends.
+  uint64_t getDeviceMemory(uint64_t defaultMemory) const {
```
rdzhabarov: I think cleaner implementation would be:

``` cpp
uint64_t getDeviceMemory(uint64_t defaultMemory) const {
  return deviceMemory == 0 ? defaultMemory : deviceMemory;
}

```

superwizard2019: > I think cleaner implementation would be:
> 
> ```c++
> uint64_t getDeviceMemory(uint64_t defaultMemory) const {
>   return deviceMemory == 0 ? defaultMemory : deviceMemory;
> }
> ```

OK. I've changed to this cleaner implementation.

rdzhabarov: Why do we need to have a switch on backendkind? Can caller provide the default value and we just return it in case deviceMemory  is not set (deviceMemory == 0).

superwizard2019: > Why do we need to have a switch on backendkind (this is not scalable, required updates every time we create a new backend)? Can caller provide the default value and we just return it in case deviceMemory is not set (deviceMemory == 0).

For CPU and interpreter backends, the default memory is 2GB. Basically it the following conditions.
1. we have both commandLine option(e.g. 1GB) and deviceConfigs(e.g.3GB), we should prefer commandline setting(1GB).
2. we have neither commandline nor deviceconfig, we need to stay with the default 2GB.
3. we have only commandline options(e.g. 1GB), we go with commandline option(1GB).
4. we only have deviceconfig(e.g. 3GB), we go with device config(3GB).

The habana has similar logic but different default memory size.

The OpenCL has a little bit different logic. The only difference is that the default memory size
might be changed by the device_mem retrieved from clGetDeviceInfo. If the deviceconfig is smaller than this value, we should go with deviceconfig. If the deviceConfig is larger, we should go with the clGetDeviceInfo because we do not have that much memory on device.

So that are a little bit different logics for different backends. So currently, I think we may need to implement them either in the devicemanager like previous commit or inside the deviceConfig. 
Let me know if you some other suggestions :)

rdzhabarov: Discussed this offline.

We need to move device memory entirely to the config and make all device managers accept non optional config. This is bigger change and could be done at later point.
To unblock this PR let's proceed with the original version.

rdzhabarov:
Let me do some cleanup in https://github.com/pytorch/glow/pull/2959 and then rebase/rework based on that.

superwizard2019:
> Let me do some cleanup in #3057 and then rebase/rework based on that.

OK. I made corresponding changes based on #3057.

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -41,7 +41,11 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
 
 public:
   CPUDeviceManager(const DeviceConfig &config, size_t maxMemory = 2000000000)
```
rdzhabarov: config already has memory param. Let's remove additional maxMemory here and make sure that `deviceMemory` is set properly in the config (make sure callers set it properly when needed, do not pass it as a separate param).

For the case of Habana where GlowHabanaMemory is defined from the GFLAGS or llvm params we should use it as long as it's set otherwise fallback to what config says.

lib/Backends/Habana/HabanaDeviceManager.h
```diff
@@ -31,6 +31,9 @@
 namespace glow {
 namespace runtime {
 
+unsigned GlowHabanaMemory = 7 << 20; // 7 GB.
```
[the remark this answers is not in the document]
superwizard2019: > it's very easy to break ODR when you define global var in the header file.

OK. I moved it back to cpp file.

lib/Backends/CPU/CPUDeviceManager.h
superwizard2019: > config already has memory param. Let's remove additional maxMemory here and make sure that `deviceMemory` is set properly in the config (make sure callers set it properly when needed, do not pass it as a separate param).
> 
> For the case of Habana where GlowHabanaMemory is defined from the GFLAGS or llvm params we should use it as long as it's set otherwise fallback to what config says.

OK. I created a new deviceConfig inside the `createCPUDeviceManager` or `createInterpreterDeviceManager` if it is set by command line. It seems that since it is a const that I could not set it directly. To keep the change smaller, I just created a new deviceConfig.

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -32,7 +32,9 @@ static llvm::cl::opt<unsigned, /* ExternalStorage */ true> GlowCPUMemoryOpt(
 DeviceManager *createCPUDeviceManager(const DeviceConfig &config) {
   if (GlowCPUMemory) {
     // Convert command line GlowCPUMemory to bytes from kilobytes.
-    return new CPUDeviceManager(config, uint64_t{GlowCPUMemory} * 1024);
+    auto configNew = DeviceConfig(BackendKind::CPU);
+    configNew.setDeviceMemory(uint64_t{GlowCPUMemory} * 1024);
```
rdzhabarov: we only need to do so if the 
```
const DeviceConfig &config)
```
does not have a device memory set, right?

rdzhabarov:
Thank you for doing this work. I think we are close to merging this.

lib/Backends/CPU/CPUDeviceManager.cpp
superwizard2019: > we only need to do so if the
> 
> ```
> const DeviceConfig &config)
> ```
> 
> does not have a device memory set, right?
> 
> It also not safe to create new device config this way. Say new param is added to the config, and now we start missing that param. We need to create new config based on the copy ctor and redefine device memory.

If I directly call the config->setDeviceMemory() , it will cause the error bellow:
 error: 'this' argument to member function 'setDeviceMemory' has type 'const glow::runtime::DeviceConfig', but function is not marked const
    config.setDeviceMemory(uint64_t{GlowCPUMemory} * 1024);

Is that because the argument is const and I should not change it?  Should I simply create new config based on the copy ctor?

rdzhabarov: > config->setDeviceMemory()

cannot be used because config is a const reference. This is ok.

> Should I simply create new config based on the copy ctor?

this seems reasonable.

superwizard2019: > > config->setDeviceMemory()
> 
> cannot be used because config is a const reference. This is ok.
> 
> > Should I simply create new config based on the copy ctor?
> 
> this seems reasonable.

OK. I switched to using copy ctor to create new config. Thanks :)

lib/Backends/Habana/HabanaDeviceManager.cpp
```diff
@@ -114,8 +114,15 @@ llvm::Error HabanaDeviceManager::init() {
 llvm::Error HabanaDeviceManager::updateMemoryUsage() {
   // TODO: Use synGetMemInfo once implemented.
 
-  totalMemory_ = uint64_t{GlowHabanaMemory} * 1024;
-  freeMemory_ = uint64_t{GlowHabanaMemory} * 1024;
+  // Use GlowHabanaMemory if it is defined from GFLAGS or llvm params,
+  // otherwise, fall back to what config says.
+  uint64_t defaultMemory = 7 << 20;
+  if (GlowHabanaMemory == defaultMemory && config_.getDeviceMemory() != 0) {
+    totalMemory_ = config_.getDeviceMemory();
```
rdzhabarov: how does it work if `getDeviceMemory()` method expects a param passed in.

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -549,13 +551,50 @@ TEST_P(DeviceManagerTest, ReuseModule) {
...
+  // Never set a value.
+  auto configEmpty = DeviceConfig(BackendKind::Interpreter);
+  EXPECT_EQ(configEmpty.getDeviceMemory(), 0);
+  // Has set a value.
+  auto configFull = DeviceConfig(BackendKind::Interpreter);
+  configFull.setDeviceMemory(32768);
+  EXPECT_EQ(configFull.getDeviceMemory(), 32768);
+
+  // Test CPU.
+  auto cpuConfigEmpty = DeviceConfig(BackendKind::CPU);
```
rdzhabarov: this is exactly the same as above. Probably can remove 564-566 lines.

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -549,13 +551,50 @@ TEST_P(DeviceManagerTest, ReuseModule) {
...
+  auto cpuConfigFull1 = DeviceConfig(BackendKind::CPU);
```
rdzhabarov: can you make better name? '1' seems unnecessary.

lib/Backends/Habana/HabanaDeviceManager.cpp
superwizard2019: > how does it work if `getDeviceMemory()` method expects a param passed in.

I defined two getDeviceMemory(). one with param(a default value ) and one without :)

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -549,13 +551,50 @@ TEST_P(DeviceManagerTest, ReuseModule) {
...
+  auto interpreterConfigFull1 = DeviceConfig(BackendKind::Interpreter);
+  interpreterConfigFull1.setDeviceMemory(32768);
+  // Only deviceConfig setting.
+  InterpreterDeviceManager interpreterDeviceSetByDeviceConfig(
+      interpreterConfigFull1);
+  EXPECT_EQ(interpreterDeviceSetByDeviceConfig.getMaximumMemory(), 32768);
+  // No setting at all, default memory size.
+  InterpreterDeviceManager interpreterDeviceDefault(interpreterConfigEmpty);
+  EXPECT_EQ(interpreterDeviceDefault.getMaximumMemory(), 2000000000);
+  // Will test OpenCL in OCLTest and test Habana in HabanaBackendTest
```
rdzhabarov: no need for this comment.

tests/unittests/DeviceManagerTest.cpp
superwizard2019: > can you make better name? '1' seems unnecessary.

Sure

tests/unittests/DeviceManagerTest.cpp
superwizard2019: > no need for this comment.

OK. I just removed that.

tests/unittests/HabanaBackendTest.cpp
```diff
@@ -375,6 +375,21 @@ TEST_F(HabanaBackendTest, FuseConvAddRelu) {
   EXPECT_EQ(F_->getNodes().size(), 4);
 }
 
+TEST_F(HabanaBackendTest, SetDeviceMemory) {
+  uint64_t defaultMemory = (7 << 20);
+  auto configEmpty = glow::runtime::DeviceConfig(BackendKind::Habana);
+  auto configFull1 = glow::runtime::DeviceConfig(BackendKind::Habana);
```
rdzhabarov: can you fix names

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -549,13 +551,41 @@ TEST_P(DeviceManagerTest, ReuseModule) {
   EXPECT_TRUE(result2->isEqual(output2));
 }
 
+TEST(DeviceManagerTest, SetDeviceMemory) {
+  // Test CPU.
+  auto cpuConfigEmpty = DeviceConfig(BackendKind::CPU);
```
rdzhabarov: can you move this cpu related stuff into `TEST(DeviceManagerTest, AvailableMemory) {` which is under ifdef GLOW_WITH_CPU.

tests/unittests/HabanaBackendTest.cpp
superwizard2019: > can you fix names

Sorry, forgot that one :)

tests/unittests/DeviceManagerTest.cpp
superwizard2019: > can you move this cpu related stuff into `TEST(DeviceManagerTest, AvailableMemory) {` which is under ifdef GLOW_WITH_CPU.

OK. Sure

superwizard2019:
> Some more nits and let's merge this.
> Make sure that sandcastle run is successful.

OK. I will make sure that

PR #2960, opened 2019-05-23
No longer use ffast-math

jfix71:
Summary: Regardless of whether we are compiling a model in Release or Debug mode, we should try to generate the same binary. Prior to this PR, for example the GraphOptimizer may have quantized a Constant Tensor payload slightly differently due to `ffast-math` when in Release vs. Debug.

Test Plan: All tests pass. 

Fixes https://github.com/pytorch/glow/issues/3856


cmake/modules/GlowDefaults.cmake
```diff
@@ -33,20 +33,19 @@ else()
...
   CHECK_CXX_COMPILER_FLAG("-Wno-psabi" HAS_W_NO_PSABI)
   if(HAS_W_NO_PSABI)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-psabi")
   endif()
   if((CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64 OR CMAKE_SYSTEM_PROCESSOR STREQUAL armv7)
      AND CMAKE_CXX_COMPILER_ID STREQUAL Clang)
-    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ${FAST_MATH_FLAGS}")
-    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} ${FAST_MATH_FLAGS}")
-    set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} ${FAST_MATH_FLAGS}")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
```
rdzhabarov: these all settings are noop.

jfix71:
> Do we have matching behavior in the fbcode?

I will make sure we do when landing this.

rdzhabarov:
@jfix71 do you want to keep this open? I see there is a fight with the stale bot going on ;)

bertmaher:
This seems fine let's go ahead and land it before it hits its 1st birthday and we have to take photos of it smashing a cake or something.

jfix71:
Yeah so the blocker for this landing was some issues in implementing it in fbcode correctly. And since then we also added constant folding (run by the Interpreter) which means that the Interpreter should also probably always be built without ffast-math, because otherwise compilation results will still depend on whether it's a Debug or Release build. So this needs to be sorted before landing.

Alternatively we could just disable fast-math entirely...

bertmaher:
I think we should disable -ffast-math entirely.  It's not enabled at all inside FB.  The amount of effort that goes into verifying numerical correctness is really big, and -ffast-math puts that entirely at the whims of the compiler in ways that are really hard to predict (e.g., looking at the recently linked issue, it's not even easy to understand that -ffast-math is going to mess that up).

PR #2961, opened 2019-05-23
[CompilationLog]: Add Compilation Scope class to track the current compilation phase.

ZchiPitt:
Summary:

- Implement a Compilation Scope class such that
  - Has a scope name which contains the function name and the compilation phase it's currently at. Example: `googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`
  - Whenever one object of CompilationScope is initialized, a message stating entering this compilation scope will be printed out to the compilation log
  - When the object destructor is called, a message stating leaving this compilation scope will be printed out to the compilation log
  - The CompilationScope will only be initialized in the beginning of one function and will be auto-collected when the function returns.

- Also implement a helper function dumpCompilationLog to dump the compilation log into a file.

Test Plan:
I intercepted five functions in the Glow codebase right now (there should be more):
- ExecutionEngine::compile()
- Loader::compile()
- glow::optimize()
- glow::optimizeFunction()
- Partitioner::Partition()

The scope at the each phase would be dumped by the helper function.

I tested by running image-classifer, with the following command `./bin/image-classifier tests/images/imagenet/cat_285.png -image-mode=0to1 -m=googlenet_v1_slim/googlenet_v1_slim.onnx -model-input-name=input:0 -image-layout=NHWC -cpu -dump-compilation-log=true`

Below is the dumped log:
`---Entering the compilation scope: Loader::compile`

`---Entering the compilation scope: Partitioner::Partition`

`---Entering the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimizeFunction`

`---Entering the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Leaving the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Entering the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Leaving the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Entering the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Leaving the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimize`

`---Leaving the compilation scope: googlenet_v1_slim/googlenet_v1_slim.onnx -> glow::optimizeFunction`

`---Leaving the compilation scope: Partitioner::Partition`

`---Leaving the compilation scope: Loader::compile`

### DIFF
--- a/include/glow/Support/CompilationLog.h
+++ b/include/glow/Support/CompilationLog.h
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_SUPPORT_COMPILATIONLOG_H
+#define GLOW_SUPPORT_COMPILATIONLOG_H
+
+#include "llvm/ADT/StringRef.h"
+
+#include <string>
+
+namespace glow {
+
+/// A class that defines the compilation scope.
+class CompilationScope {
+private:
+  /// Name of current scope.
+  /// Example: "ExecutionEngine::compile"
+  std::string funcName_;
+  std::string scopeName_;
+
+public:
+  /// When the constructor is called, current stack is pushed onto the scope
+  /// stack.
+  CompilationScope(llvm::StringRef funcName, llvm::StringRef scopeName);
+
+  /// When the destructor is called, current stack is popped out from the scope
+  /// stack.
+  ~CompilationScope();
+
+  llvm::StringRef getScopeName() const { return scopeName_; };
+};
+
+/// Dump the compilation log.
+/// \param logContent the content that will be written into the log file.
+void dumpCompilationLog(llvm::StringRef logContent);
+
+/// Dump the compilation log to a certain file.
+/// \param logFilename the file where the log will be written into.
+/// \param logContent the content that will be written into the log file.
+void dumpCompilationLog(llvm::StringRef logFilename,
+                        llvm::StringRef logContent);
+
+} // namespace glow
+
+#endif // GLOW_SUPPORT_COMPILATIONLOG_H
--- a/lib/ExecutionEngine/ExecutionEngine.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine.cpp
@@ -19,6 +19,7 @@
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/PlaceholderBindings.h"
 #include "glow/Optimizer/Optimizer.h"
+#include "glow/Support/CompilationLog.h"
 
 #include "llvm/ADT/STLExtras.h"
 
@@ -243,6 +244,9 @@ void ExecutionEngine::compile(CompilationMode mode, Function *F,
 
 void ExecutionEngine::compile(Function *F, const CompilationContext &cctx,
                               bool clearOtherFunctions) {
+  // Dump compilation log
+  CompilationScope cscope(F->getName(), "ExecutionEngine::compile");
+
   llvm::StringRef name = F->getName();
 
   if (clearOtherFunctions) {
--- a/lib/Support/CompilationLog.cpp
+++ b/lib/Support/CompilationLog.cpp
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Support/CompilationLog.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <fstream>
+
+namespace glow {
+auto constexpr defaultCompileLogFilename = "compile.log";
+static llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                          llvm::cl::desc("Dump compilation log"));
+
+CompilationScope::CompilationScope(llvm::StringRef funcName,
+                                   llvm::StringRef scopeName)
+    : funcName_(funcName), scopeName_(scopeName) {
+  if (dumpCompilationLogOpt) {
+    std::string startScopeMsg =
+        "\n---Entering the compilation scope: " + funcName_ + " -> " +
+        scopeName_;
+    dumpCompilationLog(startScopeMsg.append(scopeName_));
+  }
+}
+
+CompilationScope::~CompilationScope() {
+  if (dumpCompilationLogOpt) {
+    std::string endScopeMsg =
+        "\n---Leaving the compilation scope: " + funcName_ + " -> " +
+        scopeName_;
+    dumpCompilationLog(endScopeMsg);
+  }
+}
+
+void dumpCompilationLog(llvm::StringRef logContent) {
+  dumpCompilationLog(defaultCompileLogFilename, logContent);
+}
+void dumpCompilationLog(llvm::StringRef logFilename,
+                        llvm::StringRef logContent) {
+  llvm::outs() << "Writing compilation log file for Module to: " << logFilename
+               << '\n';
+  std::error_code EC;
+  llvm::raw_fd_ostream myfile(logFilename, EC, llvm::sys::fs::OF_Append);
+  myfile << logContent << "\n";
+}
+
+} // namespace glow
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -24,6 +24,7 @@
 #include "glow/Optimizer/Optimizer.h"
 #include "glow/Quantization/Base/Base.h"
 #include "glow/Quantization/Quantization.h"
+#include "glow/Support/CompilationLog.h"
 
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
@@ -2606,6 +2607,9 @@ void glow::fold(Function *F, CompilationMode mode) {
 }
 
 void glow::optimize(Function *F, const CompilationContext &cctx) {
+  // Dump compilation log
+  CompilationScope cscope(F->getName(), "glow::optimize");
+
   // Optimize may be called after backend specific transformations and some
   // nodes may have become unused. It is a good idea to remove them, before
   // proceeding with any further optimizations.
@@ -2764,6 +2768,9 @@ static void transformForPrecisionMode(const Backend &B, Function *F,
 // docs/GraphOptimizationPipeline.md
 llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
                                    const CompilationContext &cctx) {
+  // Dump compilation log
+  CompilationScope cscope(F->getName(), "glow::optimizeFunction");
+
   // Verify the function pre-optimization/lowering.
   assert(F->verify() && "Function must be valid");
 
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -21,6 +21,7 @@
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "glow/Support/CompilationLog.h"
 #include <fstream>
 
 using namespace glow;
@@ -782,6 +783,8 @@ llvm::Error Partitioner::createDAGWithoutPartition(
 }
 
 llvm::Error Partitioner::Partition(const CompilationContext &cctx) {
+  CompilationScope cscope("", "Partitioner::Partition");
+
   // Prepare the mapping between BackendKind and BackendInfo.
   std::map<BackendKind, BackendInfo> backendMap;
   std::vector<Backend *> backends;
--- a/lib/Support/CMakeLists.txt
+++ b/lib/Support/CMakeLists.txt
@@ -3,7 +3,8 @@ add_library(Support
               Error.cpp
               Random.cpp
               Support.cpp
-              ThreadPool.cpp)
+              ThreadPool.cpp
+              CompilationLog.cpp)
 target_link_libraries(Support
                       PUBLIC
                         LLVMSupport)
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -22,6 +22,7 @@
 #include "glow/Quantization/Quantization.h"
 #include "glow/Quantization/Serialization.h"
 #include "glow/Runtime/RuntimeTypes.h"
+#include "glow/Support/CompilationLog.h"
 
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
@@ -272,6 +273,9 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
 }
 
 void Loader::compile(PlaceholderBindings &bindings) {
+  // Dump compilation log
+  CompilationScope cscope("", "Loader::compile");
+
   // Dump the DAG before compilation if needed.
   if (!dumpGraphDAGFileBeforeCompilationOpt.empty()) {
     F_->dumpDAG(dumpGraphDAGFileBeforeCompilationOpt.c_str());


lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -243,6 +244,12 @@ void ExecutionEngine::compile(CompilationMode mode, Function *F,
 
 void ExecutionEngine::compile(Function *F, const CompilationContext &cctx,
                               bool clearOtherFunctions) {
+  // Dump compilation log
+  CompilationScope cscope("ExecutionEngine::compile");
+  if (dumpCompilationLogOpt) {
```
opti-mix: Idea: To avoid having this conditional check everywhere, maybe you could move the whole if statement into the constructor of `CompilationScope`?

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,62 @@
...
+
+public:
+    /// When the constructor is called, current stack is pushed in the scope stack.
+    CompilationScope(std::string scopeName);
+
+    /// When the destructor is called, current stack is poped out from the scope stack.
+    ~CompilationScope();
+
+    /// A function that returns the current compilation scope stack.
+    llvm::StringRef getCurrentCompilationScopeStack();
```
opti-mix: Given the fact that there is a `scopeStack` member in this class, the name of this method sounds a bit misleading. Maybe something like `getFullScopeName()` should be used instead?

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,62 @@
...
+
+extern  llvm::cl::opt<bool> dumpCompilationLogOpt;
+
+/// A class that defines the compilation scope. 
+///
+/// Example: 
+class CompilationScope {
+public:
+    /// A static stack that maintains the current scope stack.
+    static std::vector<std::string> scopeStack;
```
opti-mix: All class members should end with an underscore.

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,62 @@
...
+private:
+    std::string scopeName;
```
opti-mix: Please add a getter for this member.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+std::vector<std::string> CompilationScope::scopeStack = {};
+
+CompilationScope::CompilationScope(std::string scopeName) {
+    this->scopeName = scopeName;
+    scopeStack.push_back(scopeName);
+}
+
+CompilationScope::~CompilationScope() {
+    scopeStack.pop_back();
```
opti-mix: If you need to log the "end of scope" events, you could add it here.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+    for (auto scopeStr : scopeStack) {
+        scopes = scopes + "-->" + scopeStr;
+    }
+    return scopes;
+}
+
+void dumpCompilationLog (std::string logContent) {
+    dumpCompilationLog ("compile.log", logContent);
+}
+void dumpCompilationLog (llvm::StringRef logFilename, std::string logContent) {
```
SplitInfinity: Consider using a reference to a string here to avoid copying the log content (which could be quite long).

include/glow/Support/CompilationLog.h
SplitInfinity: `scopeName` -> `scopeName_`.

Also, when is this used? It is initialized but never used after that.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+    scopeStack.push_back(scopeName);
+}
+
+CompilationScope::~CompilationScope() {
+    scopeStack.pop_back();
+}
+
+llvm::StringRef CompilationScope::getCurrentCompilationScopeStack() {
+    std::string scopes = "";
+    for (auto scopeStr : scopeStack) {
```
SplitInfinity: You can optimize this a bit if you create a member variable for the full scope name and update it when a `CompilationScope` is created or  destroyed. If this is called multiple times within the same scope it will redo the same work each time to get the full scope name.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+        scopes = scopes + "-->" + scopeStr;
+    }
+    return scopes;
```
SplitInfinity: This function's return type is `StringRef` but the string to which the returned value refers is a local variable. For my own education, why does this work?

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace glow {
+llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                     llvm::cl::desc("Dump compilation log"));
+
+std::vector<std::string> CompilationScope::scopeStack = {};
+
+CompilationScope::CompilationScope(std::string scopeName) {
+    this->scopeName = scopeName;
```
SplitInfinity: Once you change the name of the member variable to `scopeName_`, initialize it in the constructor's initialization list:
```
CompilationScope::CompilationScope(std::string scopeName)  :  scopeName_(scopeName) {...
```

If you decide to keep the string as a member variable and a reference to it on the scope stack, you can move the provided `scopeName` into `scopeName_` instead.

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+#ifndef GLOW_SUPPORT_COMPILATIONLOG_H
+#define GLOW_SUPPORT_COMPILATIONLOG_H
+
+#include <string>
+#include "llvm/ADT/SmallString.h"
+#include "llvm/Support/CommandLine.h"
+
+
```
SplitInfinity: I think only one blank line is necessary here.

lib/ExecutionEngine/ExecutionEngine.cpp
SplitInfinity: We could also avoid `extern`ing `dumpCompilationLogOpt` if we did this.

include/glow/Support/CompilationLog.h
bertmaher: Must we actually store a stack?  That feels a bit allocation-intensive for what we're doing here, which is just logging begin/end of compilation phases.

lib/ExecutionEngine/ExecutionEngine.cpp
ZchiPitt: awesome!

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -243,6 +244,12 @@ void ExecutionEngine::compile(CompilationMode mode, Function *F,
 
 void ExecutionEngine::compile(Function *F, const CompilationContext &cctx,
                               bool clearOtherFunctions) {
+  // Dump compilation log
+  CompilationScope cscope("ExecutionEngine::compile");
```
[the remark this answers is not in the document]
ZchiPitt: good idea!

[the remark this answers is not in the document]
ZchiPitt: and the func didn't include the class name, which may cause confusion sometimes, especially we have Loader::compile and also glow::compile

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,58 @@
...
+    scopeStack.push_back(scopeName);
```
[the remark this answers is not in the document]
ZchiPitt: yea, it's a good idea. thx!

But probably i cant store references in the scopestack as i need concatenate them later to form a full stack name, `for (auto scopeStr : scopeStack_) {
    scopes = scopes + "-->" + scopeStr;
  }`.

lib/Support/CompilationLog.cpp
ZchiPitt: seems that ofstream cannot work with llvm::stringRef

SplitInfinity: [What about `llvm::raw_fd_ostream`?](https://llvm.org/doxygen/classllvm_1_1raw__ostream.html#a2d4592b138e1cf14fa0b8913d0747ac9)

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+namespace glow {
+
+/// A class that defines the compilation scope.
+///
+/// Example:
+class CompilationScope {
+public:
+  /// A static stack that maintains the current scope stack.
+  static std::vector<std::string> scopeStack_;
```
SplitInfinity: Why is this member public? It is only accessed within this class.

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+
+private:
+  std::string scopeName_;
```
SplitInfinity: Add a comment explaining what this member is.

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+
+public:
+  /// When the constructor is called, current stack is pushed in the scope
```
SplitInfinity: Did you mean `current scope is pushed onto the scope stack`?

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+  llvm::StringRef getCurrentCompilationScopeStack();
+};
+
+/// Dump the compilation log.
+/// \param logContent the content that will be written into the log file.
+void dumpCompilationLog(std::string logContent);
+
+/// Dump the compilation log to a certain file.
+/// \param logFilename the file where the log will be written into.
+/// \param logCotent the content that will be written into the log file.
```
SplitInfinity: `logCotent` -> `logContent`

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+namespace glow {
+static llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                          llvm::cl::desc("Dump compilation log"));
+
+std::vector<std::string> CompilationScope::scopeStack_ = {};
+
+CompilationScope::CompilationScope(std::string scopeName)
+    : scopeName_(scopeName) {
+  scopeStack_.push_back(scopeName);
```
SplitInfinity: See my earlier comment about keeping the actual data in `scopeName_` and references to it on `scopeStack_`.

Also consider `std::move`ing `scopeName` into either `scopeName_` or onto `scopeStack_` to avoid another copy constructor invocation.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+  if (dumpCompilationLogOpt) {
+    dumpCompilationLog(this->getCurrentCompilationScopeStack());
```
SplitInfinity: No need to use `this` within the scope of the class.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+CompilationScope::~CompilationScope() { scopeStack_.pop_back(); }
+
+llvm::StringRef CompilationScope::getCurrentCompilationScopeStack() {
+  std::string scopes = "";
+  for (auto scopeStr : scopeStack_) {
```
SplitInfinity: `auto` will never produce a reference and you don't need a local copy. Consider changing this to `const auto&`.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+    scopes = scopes + "-->" + scopeStr;
```
SplitInfinity: Consider using `std::string::append` or `std::string::operator+=` here instead of `std::string::operator+` to avoid a a few extra copies and assignment.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+    : scopeName_(scopeName) {
+  scopeStack_.push_back(scopeName);
+  if (dumpCompilationLogOpt) {
+    dumpCompilationLog(this->getCurrentCompilationScopeStack());
+  }
+}
+
+CompilationScope::~CompilationScope() { scopeStack_.pop_back(); }
+
+llvm::StringRef CompilationScope::getCurrentCompilationScopeStack() {
```
SplitInfinity: See my earlier comment. It looks like you can maintain the result of this function in a member variable, update in the constructor and destructor and return it here.

As is, this function will redo the same work each time it is called.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+void dumpCompilationLog(llvm::StringRef logFilename, std::string logContent) {
+  llvm::outs() << "Writing compilation log file for Module to: " << logFilename
+               << '\n';
+
+  std::ofstream myfile;
+  myfile.open(logFilename, std::ios_base::app);
+  myfile << logContent << "\n";
+}
+
+} // namespace glow
```
SplitInfinity: I think you're missing a newline at the end of this file.

include/glow/Support/CompilationLog.h
ZchiPitt: oops, thx for the grammar check, lol

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+  /// stack.
+  ~CompilationScope();
+
+  /// A function that returns the current compilation scope stack.
+  llvm::StringRef getCurrentCompilationScopeStack();
+};
+
+/// Dump the compilation log.
+/// \param logContent the content that will be written into the log file.
+void dumpCompilationLog(std::string logContent);
```
rdzhabarov: passing strings by value is not efficient. Can you go through the APIs and make them either StringRef or const std::string&

lib/Support/CompilationLog.cpp
ZchiPitt: works now! Thx a lot!

lib/Support/CompilationLog.cpp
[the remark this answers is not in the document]
ZchiPitt: change to append and it works now

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,60 @@
...
+    dumpCompilationLog(getFullScopeName());
+  }
+}
+
+CompilationScope::~CompilationScope() { scopeStack_.pop_back(); }
+
+llvm::StringRef CompilationScope::getFullScopeName() {
+  std::string scopes = "";
+  for (const auto scopeStr : scopeStack_) {
+    scopes.append(scopeStr);
```
SplitInfinity: Do you need to append `"--->"` as well?

lib/Support/CompilationLog.cpp
ZchiPitt: done

lib/Support/CompilationLog.cpp
ZchiPitt: i changed it to std::string to avoid any future problems

lib/Support/CompilationLog.cpp
ZchiPitt: added!

ZchiPitt:
The complaints of using a static stack totally make sense to me. Do we have any ideas other than that? Anything is welcomed!  

It is just to me, the easiest way to log in the start/end of compilation phases other than passing objects though the whole compilation process.

opti-mix:
> The complaints of using a static stack totally make sense to me. Do we have any ideas other than that? Anything is welcomed!
It is just to me, the easiest way to log in the start/end of compilation phases other than passing objects though the whole compilation process.

In theory, we may have multiple compilations happening simultaneously. Thus having a single static stack is not ideal. 

Maybe this compilation specific state can be associated with something/attached to something, e.g. to a Graph or to a Function/IRFunction object? Or may be we need to introduce some kind of a compilation context, so that you can always refer to it from a Graph/Function/IRFunction? This context could then contain the current stack or any other state related to the specific compilation.

bertmaher:
If the goal is to write a log, there's no need to store a stack at all; just log begin/end markers when each phase begins/ends.  Your log ends up being:
```
BeginCompile
BeginReticulateSplines
EndReticulateSplines
...
EndCompile
```
You can reconstruct the stack from that without the runtime cost of maintaining the stack in memory.

ZchiPitt:
@bertmaher if `In theory, we may have multiple compilations happening simultaneously`, then just output the beginning and ending of a scope may not be ideal in this case. 

So maybe add the Function name also to the output, which then associate the scope also with the function context? @opti-mix

ZchiPitt:
ok, so i removed the static stack right now. And log the begin/end markers of each compilation scope. Also i include the function name in the compilation scope name. The logic is much simpler right now.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <fstream>
+
+namespace glow {
+static llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                          llvm::cl::desc("Dump compilation log"));
+
+CompilationScope::CompilationScope(llvm::StringRef scopeName)
+    : scopeName_(std::move(scopeName)) {
```
SplitInfinity: I think `std::move` is not very useful here since `scopeName` is a `StringRef` now. I think I recommended `std::move` here when `scopeName`'s type was `std::string`.

lib/Support/CompilationLog.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+CompilationScope::~CompilationScope() {
+  if (dumpCompilationLogOpt) {
+    std::string endScopeMsg("\n---Leaving the compilation scope: ");
+    dumpCompilationLog(endScopeMsg.append(scopeName_));
+  }
+}
+
+void dumpCompilationLog(llvm::StringRef logContent) {
+  dumpCompilationLog("compile.log", logContent);
```
SplitInfinity: Consider either adding a comment to explain what `"compile.log"` is:
```
dumpCompilationLog(/*logFilename=*/"compile.log", logContent);
```
or define `"compile.log"` as a `constexpr` at the top of the file with a descriptive name:
```
namespace {
auto constexpr defaultCompileLogFilename = "compile.log";
} // namespace
...
dumpCompilationLog(defaultCompileLogFilename, logContent);
````

File: include/glow/Support/CompilationLog.h [hunk 27]
```diff
@@ -0,0 +1,60 @@
...
+
+/// Dump the compilation log to a certain file.
+/// \param logFilename the file where the log will be written into.
+/// \param logContent the content that will be written into the log file.
+void dumpCompilationLog(llvm::StringRef logFilename,
+                        llvm::StringRef logContent);
+
+} // namespace glow
+
+#endif // GLOW_SUPPORT_COMPILATIONLOG_H
```
SplitInfinity: I think you're missing a newline at the end of this file.

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+
+/// A class that defines the compilation scope.
+///
+/// Example:
+class CompilationScope {
+private:
+  /// Name of current scope.
+  /// Example: "ExecutionEngine::compile"
+  std::string scopeName_;
+  // Full scope name that reprensents the entire compilation phases till now.
```
SplitInfinity: `reprensents` -> `represents`

include/glow/Support/CompilationLog.h
```diff
@@ -0,0 +1,60 @@
...
+  std::string fullScopeName_;
```
SplitInfinity: Is this used anywhere?

ZchiPitt: Probably will be needed in the future. I'll remove it right now.

ZchiPitt:
Close and will resubmit with a new design later.

PR #2963, opened 2019-05-23
ONNXModelLoader can automatically create input placeholders.

putivsky:
Summary:
For inference mode ONNXModelLoader constructor can detect and create Placeholders for model inputs.
Caller may not know in advance inputs, auto-detection is a useful feature.

Differential Revision: D15475110

### DIFF
--- a/include/glow/Importer/ONNXModelLoader.h
+++ b/include/glow/Importer/ONNXModelLoader.h
@@ -187,6 +187,7 @@ public:
   /// serialized in \p modelDescFilename and populates the network into \p F.
   /// The types in \p types match the list of names \p tensorNames and used as
   /// inputs to the network.
+  /// If \p names and \p types are empty loader fills inputs automatically.
   /// If \p errPtr is not null then if an error occurs it will get assigned
   /// there otherwise if an error occurs it will abort.
   ONNXModelLoader(const std::string &modelDescFilename,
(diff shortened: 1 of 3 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1156,3 +1156,42 @@ ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
     EXIT_ON_ERR(setup());
   }
 }
+
+ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
```
jackm321: This is mostly copy pasted from previous constructor, Maybe just factor the `setup` lambda into a separate function with a bool input specifiying how to treat inputs and call it from each constructor or something like that?

putivsky: I can re-use constructor if names and types are empty then use loadInputs().

include/glow/Importer/ONNXModelLoader.h
```diff
@@ -187,6 +187,7 @@ class ONNXModelLoader
   /// serialized in \p modelDescFilename and populates the network into \p F.
   /// The types in \p types match the list of names \p tensorNames and used as
   /// inputs to the network.
+  /// If names and types are empty loader fills inputs automatically.
```
jackm321: nit use \p names and \p types

PR #2965, opened 2019-05-23
[Partitioner] Bug fix: remove duplicated generated "save" node in heterogeneous partition.

beicy:
Summary:
In Heterogeneous partition, after backendBasedPartition, we got a list of function. Then when we finally create the DAG, it is possible that one input in one function may use the output of another function. In that case, we don't need to generate an "save" node again. i

Documentation:

Test Plan:
ninja test. Check the generated partitions graph manually. 

[Optional Fixes #issue]


opti-mix:
@beicy Tested on one of the private test-cases and I can confirm it solves the issue I had! Thanks!

Question: Would it be possible to add a test-case for this edge-case?

opti-mix:
@beicy Thanks for adding the test!

Just checking: The test does not pass without this fix, right?

beicy:
> @beicy Thanks for adding the test!
> 
> Just checking: The test does not pass without this fix, right?

Yes. If I removed the fix, Heterogenous partition test failed.

PR #2966, opened 2019-05-23
Log data size increase from static sizing and PCI-e bw on Habana

bertmaher:
Summary:
Log the input tensor size from onnxifi vs GLOW static sizes.
Measure time for transferring input data to device on PCI-e and log achieved
PCI-e bandwidth.

Differential Revision: D15478267


bertmaher:
I apparently can't approve this because I exported it, but the PR is really from @narayanan2004.

PR #2967, opened 2019-05-23
[unittests] Add option to clone a FunctionTensorPair repeatedly inside itself

jfix71:
Summary: Take a Function with a single SaveNode and copy it repeatedly inside itself. This is to enable stress testing an architecture with many embarrassingly parallel graphs.

Test Plan: Added a couple example tests using this functionality.

For example, for the original matmul test:

Screen Shot 2019-05-23 at 11 12 21 PM

And after cloning it 3 times via `cloneFunInsideFun()`:

Screen Shot 2019-05-23 at 11 12 30 PM

Note that each Placeholder's corresponding backing tensor is copied in the PlaceholderBindings, so each parallel clone should have the same results.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -262,6 +262,14 @@ public:
     return N;
   }
 
+  /// Take ownership of \p N by removing it from its original Function, add it
+  /// to the current Function, and also unique its name.
+  void takeOwnershipOfNode(Node *N) {
+    N->getParent()->getNodes().remove(N);
+    N->setName(Module::uniqueName(N->getName(), uniqueNodeNames_));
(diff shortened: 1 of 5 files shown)


jfix71:
I've added a command line option for BackendTestUtils here called `-parallel-clone-count` which will clone Functions as described for a specific count. I added this option to all of the `compareAgainstInterpreter()` tests, and an example for the matmul test so the functionality is always tested.

I think it could be a good idea to enable this for as many tests as possible. By default they will test the normal graph as expected. We can then ramp up the parallel clone count to stress test the compiler across many of the OperatorTests. For now I'm not touching the other OperatorTests though.

PR #2968, opened 2019-05-24
Repro incorrect predictions with huge SLS embeddings

bertmaher:
Summary:
We're getting wrong results with big embeddings; this test repros it
as a pure Glow unit test.

Reviewed By: hyuen

Differential Revision: D15423338

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,10 +1,13 @@
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOW_BINARY_DIR}/tests/)
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/unittests)
+
 add_subdirectory(Testing)
 add_subdirectory(unittests)
 add_subdirectory(images)
 add_subdirectory(benchmark)
 add_subdirectory(models)
 add_subdirectory(runtime_test)
+add_subdirectory(stress)
 
 # External backends
 ExternalBackendsTest()
@@ -18,7 +21,7 @@ function(add_output_check_test)
   if (NOT ARG_NAME)
     message(FATAL_ERROR "Name mandatory")
   endif()
-  
+
   if (NOT ARG_COMMAND)
     message(FATAL_ERROR "Command mandatory")
   endif()
@@ -102,7 +105,7 @@ add_output_check_test(NAME resnet_runtime_test
                       COMMAND ${EXAMPLES_TESTS_DIR}/resnet-runtime-test.sh
                       REQUIRES MODELS)
 
-# Create target for OUTPUTCHECK_TESTS. 
+# Create target for OUTPUTCHECK_TESTS.
 LIST(APPEND OUTPUTCHECK_TESTS true)
 add_custom_target(output_check_tests
   COMMAND ${OUTPUTCHECK_TESTS}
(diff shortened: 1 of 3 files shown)


bertmaher:
A bit more about this diff:

I'm proposing a new category of test, which is a "stress" test (or you could say, an "acceptance" test).  These are rigorous and extensive but potentially very time-consuming tests.  We want to have them available in the repo, say, to validate new hardware or new backends, but we probably don't want to run them during development or even in CI.

For example, the SLS stress test added in this diff can take nearly an hour to run in debug mode -- mostly because our method of random-filling is super slow, but that's a side point -- and I don't want to massively slow down development so that we can re-check this on every diff.

jfix71:
Sounds like a great idea -- https://github.com/pytorch/glow/pull/2967 somewhat falls into this category too, cloning a graph inside the function many times to simultaneously stress many parallel compute engines of an accelerator.

tests/stress/SparseLengthsSumTest.cpp
```diff
@@ -0,0 +1,134 @@
...
+  }
+  for (auto *weight : weights) {
+    auto *W = bindings.allocate(weight);
+    W->getHandle<float>().randomize(-1.0, 1.0, mod_.getPRNG());
+  }
+  for (auto *result : results) {
+    bindings.allocate(result);
+  }
+
+  EE_.compile(CompilationMode::Infer, F_);
```
jfix71: This call to compile will transform `F_`. It might not be an issue currently, but it's possible the backend being tested could transform the Function in a way that breaks things for the Interpreter. So it would probably be a good idea to clone `F_` right before here so you have a clean, unoptimized version of `F_` for each call to compile.

tests/stress/SparseLengthsSumTest.cpp
```diff
@@ -0,0 +1,134 @@
...
+    T->getHandle<float>().clear(0);
+  }
+
+  ExecutionEngine interp{BackendKind::Interpreter};
+  interp.compile(CompilationMode::Infer, F_);
+  interp.run(bindings);
+  std::vector<Tensor> base;
+  for (auto *result : results) {
+    auto *T = bindings.get(result);
+    base.push_back(T->clone());
```
jfix71: Cloning + clearing the Tensors here seems unnecessary since no one else will use them after. Maybe just make `base` a vector of Tensor pointers and push each `T`.

tests/CMakeLists.txt
```diff
@@ -1,9 +1,12 @@
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOW_BINARY_DIR}/tests/)
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/unittests)
```
jfix71: Do we need this? Does `add_subdirectory(unittests)` below not handle this?

bertmaher: Yeah, the actually sets the `-I` include path so that it contains the headers in tests/unittests, whereas `add_subdirectory` just tells CMake to traverse that subdirectory.

tests/stress/SparseLengthsSumTest.cpp
bertmaher: Heh, so I actually initially tried cloning `F_` and apparently it caused some Habana-specific transformations to not kick in, in a way that caused the function to be invalid.  I guess I should probably figure those out since it's not great to have cloning change behavior.

PR #2969, opened 2019-05-24
[Dump] add dump functions for Glow data structure

### LINKED ISSUE #2831: Make all useful data structures in Glow easy to dump in a textual form
Being able to dump any interesting data structures is very useful for interactive debugging and print debugging.

Currently, many of interesting data structures (e.g. `glow::Value`) have `dump` methods. Some others, e.g. `glow::Node` have `getDebugDesc()` method, but no `dump` method. And many interesting data structures don't have any method to dump their content, e.g. `glow::Type`.

Ideally, it would be nice to have the same uniform way to dump any data structure. LLVM usually uses `dump` for it.

It is also very useful to be able to get a string representation of the dumped content of any data structure as it allows for including it into a bigger composite output consisting of many different parts.

So, in short it would be nice to have e.g. the following methods for all interesting data structures in Glow:
- `dump(llvm::raw_ostream &os)`
- `dump()` which just calls `dump(llvm::raw_ostream &os)` with a default output stream as a parameter
- `std::string toString()`, which dumps into string

superwizard2019:
Summary
- add missing `dump(llvm::raw_ostream &os)`, `dump()` , and `std::string toString()` for glow::Module, glow::Function, glow::IRFunction, glow::Value, glow::Instruction, glow::Tensor, glow::Type and glow::Node.
- add overloading operator `llvm::raw_ostream &operator<<`  for above data structures and their reference.

Test:
add tests in TensorsTest, BackendTest, GraphTest and IROptTest to verify the new dump functions work and the outputs match the expected results.

Fixes:
This PR fixes #2831

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -25,6 +25,7 @@
 #include "glow/Support/Random.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/Support/raw_ostream.h"
 
 namespace glow {
 
@@ -357,6 +358,27 @@ public:
     std::swap(tensorPool_, other.tensorPool_);
     return *this;
   }
+
+  /// Dump a textual representation of the Tensor into provided output stream.
+  void dump(llvm::raw_ostream &os) const;
+
+  /// Dump a textual representation of the Tensor into default output stream.
+  void dump() const;
+
+  /// Dump a textual representation of a specific number of elements in the
+  /// Tensor into provided output stream.
+  void dump(llvm::raw_ostream &os, unsigned maxNumElem) const;
+
+  /// Dump a textual representation of a specific number of elements in the
+  /// Tensor into default output stream.
+  void dump(unsigned maxNumElem) const;
+
+  /// Dump a textual representation of the Tensor to std::string.
+  std::string toString() const;
+
+  /// Dump a textual representation of a specific number of elements in the
+  /// Tensor to std::string.
+  std::string toString(unsigned maxNumElem) const;
 
   /// \returns true if the content of the other tensor \p other is identical to
   /// this one.
@@ -1003,6 +1025,9 @@ template <class ElemTy> const Handle<ElemTy> Tensor::getHandle() const & {
   return Handle<ElemTy>(const_cast<Tensor *>(this));
 }
 
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Tensor &t);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Tensor *t);
 } // namespace glow
 
 #endif // GLOW_BASE_TENSOR_H
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -532,6 +532,15 @@ struct Type final {
     return names[(int)Ty];
   }
 
+  /// Dump a textual representation of the Type into provided output stream.
+  void dump(llvm::raw_ostream &out) const;
+
+  /// Dump a textual representation of the Type into default output stream.
+  void dump() const;
+
+  /// Dump a textual representation of the Type to std::string.
+  std::string toString() const;
+
 private:
   /// Setup the internals of type that store the dimensions. This method is
   /// used by the constructor.
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -163,8 +163,14 @@ public:
   /// Get the pseudo-random number generator used by this module.
   PseudoRNG &getPRNG() { return PRNG_; }
 
-  /// Dumps the textual representation of the network.
+  /// Dump a textual representation of the Module into default output stream.
   void dump() const;
+
+  /// Dump a textual representation of the Module to std::string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the Module into provided output stream.
+  void dump(llvm::raw_ostream &os) const;
 
   /// Dump a dotty graph that depicts the Module.
   void dumpDAG();
@@ -1077,8 +1083,14 @@ public:
   /// \returns true when the function is valid. False otherwise.
   bool verify() const;
 
-  /// Dumps the textual representation of the network.
+  /// Dump a textual representation of the Function into provided output stream.
   void dump() const;
+
+  /// Dump a textual representation of the Function to std::string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the Function into default output stream.
+  void dump(llvm::raw_ostream &os) const;
 
   /// Dump a dotty graph that depicts the function into a file.
   /// \returns full path to the file.
@@ -1128,6 +1140,14 @@ SaveNode *getOutputSave(Function *F, Placeholder *PH);
   { 0u, 2u, 3u, 1u }
 #define NHWC2NCHW                                                              \
   { 0u, 3u, 1u, 2u }
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Module &mod);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Module *mod);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Function &F);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Function *F);
 
 } // namespace glow
 
--- a/include/glow/Graph/Node.h
+++ b/include/glow/Graph/Node.h
@@ -110,6 +110,15 @@ public:
   /// \returns a textual description of the node.
   std::string getDebugDesc() const;
 
+  /// Dump a textual representation of the Node into provided output stream.
+  void dump(llvm::raw_ostream &out) const;
+
+  /// Dump a textual representation of the Node into default output stream.
+  void dump() const;
+
+  /// Dump a textual representation of the Node to std::string.
+  std::string toString() const;
+
   /// \returns copy of the current node. Notice that the new node is not
   /// inserted into any DAG. The caller of this method should add it to some
   /// node-list.
@@ -380,6 +389,10 @@ constexpr unsigned LHSIdx = 0;
 constexpr unsigned RHSIdx = 1;
 constexpr unsigned ResultIdx = 0;
 } // namespace ArithmeticNode
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Node &node);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Node *node);
 
 } // namespace glow
 
--- a/include/glow/IR/IR.h
+++ b/include/glow/IR/IR.h
@@ -93,11 +93,14 @@ public:
   /// Verify the correctness of the instruction parameters.
   void verify(const IRFunction &M) const;
 
-  /// Print value.
+  /// Dump a textual representation of the Value into provided output stream.
   void dump(llvm::raw_ostream &out) const;
 
-  /// Print value using a default output stream.
+  /// Dump a textual representation of the Value into default output stream.
   void dump() const;
+
+  /// Dump a textual representation of the Value to std::string.
+  std::string toString() const;
 
   /// Print value in context.
   void dumpInContext(llvm::raw_ostream &out) const;
@@ -302,10 +305,15 @@ public:
   /// Verify the correctness of the function.
   void verify() const;
 
-  /// Dump a textual representation of the function.
+  /// Dump a textual representation of the IRFunction into default output
+  /// stream.
   void dump() const;
 
-  /// Dump a textual representation of the function into provided output stream.
+  /// Dump a textual representation of the IRFunction to std::string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the IRFunction into provided output
+  /// stream.
   void dump(llvm::raw_ostream &OS) const;
 
   /// Dump a dotty graph that depicts the function.
@@ -388,6 +396,14 @@ public:
   /// was assigned to this instruction.
   int64_t getInstrNumber(const Instruction *I) const;
 };
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Value &V);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Value *V);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const IRFunction &irf);
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const IRFunction *irf);
 
 } // namespace glow
 
--- a/lib/Base/Type.cpp
+++ b/lib/Base/Type.cpp
@@ -54,4 +54,16 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const TypeRef &type) {
   }
   return os << *type;
 }
+
+void Type::dump(llvm::raw_ostream &out) const { out << this; }
+
+void Type::dump() const { dump(llvm::outs()); }
+
+std::string Type::toString() const {
+  std::string storage;
+  llvm::raw_string_ostream os(storage);
+  os << this;
+  return os.str();
+}
+
 } // namespace glow
--- a/lib/Graph/Node.cpp
+++ b/lib/Graph/Node.cpp
@@ -258,6 +258,12 @@ std::string Node::getDebugDesc() const {
   }
 }
 
+void Node::dump(llvm::raw_ostream &out) const { out << this->getDebugDesc(); }
+
+void Node::dump() const { dump(llvm::outs()); }
+
+std::string Node::toString() const { return this->getDebugDesc(); }
+
 Node *Node::clone() const {
   switch (getKind()) {
 #define DEF_NODE(CLASS, NAME)                                                  \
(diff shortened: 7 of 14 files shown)


lib/Base/Type.cpp
```diff
@@ -54,4 +54,43 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const TypeRef &type) {
   }
   return os << *type;
 }
+
+void Type::dump(llvm::raw_ostream &out) const {
+  out << "Type structure :\n";
```
opti-mix: I think we typically do not print the prefix like "Type structure:".

lib/Base/Type.cpp
```diff
@@ -54,4 +54,43 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const TypeRef &type) {
...
+  out << this << "\n";
+}
+
+void Type::dump() const { dump(llvm::outs()); }
+
+std::string Type::toString() const {
```
opti-mix: You should probably use `llvm::raw_string_ostream` here. Then you can just do 
```c++
std::string storage;
llvm::raw_string_ostream os(storage);
os << this; 
return os.str();
```

This way you don't need to duplicate the printing logic.

lib/Graph/Node.cpp
```diff
@@ -258,6 +258,21 @@ std::string Node::getDebugDesc() const {
   }
 }
 
+void Node::dump(llvm::raw_ostream &out) const {
+  out << "Node structure " << this->getName() << ":\n";
```
opti-mix: We typically do not print the prefix like "Node structure:".

tests/unittests/BackendTest.cpp
```diff
@@ -497,3 +497,18 @@ INSTANTIATE_TEST_CASE_P(JIT, BackendTest, ::testing::Values(BackendKind::CPU));
 INSTANTIATE_TEST_CASE_P(OpenCL, BackendTest,
                         ::testing::Values(BackendKind::OpenCL));
 #endif // GLOW_WITH_OPENCL
+
+// check if the dump function works.
+TEST(BackendTest, dumpType) {
```
opti-mix: The test is good, because it checks if APIs are present at all. On top of it could also test the correctness of the output, e.g. of tyB->toString() matches the expected string literal.

include/glow/Base/Tensor.h
```diff
@@ -338,6 +339,23 @@ class Tensor final {
...
+  // Dump tensor information
+  void dump(llvm::raw_ostream &os) const;
+  void dump() const;
+  std::string toString() const;
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Tensor &t) {
+    t.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Tensor *const &t) {
```
SplitInfinity: Hmm, reference to a pointer? Why not pass just the pointer (i.e. `const Tensor *const t`)?

include/glow/Graph/Graph.h
```diff
@@ -1061,6 +1081,26 @@ class Function final : public Named {
...
   void dump() const;
 
+  /// Dump to string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the function into provided output stream.
+  void dump(llvm::raw_ostream &os) const;
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Function &fun) {
```
SplitInfinity: Typically we use `F` as a name for name standalone `Function` variables.

include/glow/Graph/Graph.h
```diff
@@ -1061,6 +1081,26 @@ class Function final : public Named {
...
+    fun.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Function *const &fun) {
```
SplitInfinity: Same here.

include/glow/IR/IR.h
```diff
@@ -99,6 +99,22 @@ class Value : public Named,
   /// Print value using a default output stream.
   void dump() const;
 
+  /// Print value to string.
+  std::string toString() const;
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Value &v) {
```
SplitInfinity: Typically we use `V` (capital) as a name for name standalone `Value` variables.

include/glow/IR/IR.h
```diff
@@ -99,6 +99,22 @@ class Value : public Named,
...
+    v.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Value *const &v) {
```
SplitInfinity: Same here.

include/glow/IR/IR.h
```diff
@@ -212,6 +228,29 @@ class Instruction : public Value, public TaggedListNode<Instruction> {
...
   static bool classof(const Instruction *I) { return true; }
 
+  /// Dump a textual representation of the Instruction.
+  void dump() const;
+
+  /// Dump to string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the function into provided output stream.
+  void dump(llvm::raw_ostream &OS) const;
```
SplitInfinity: Is there a reason this parameter name is capitalized but the others aren't?

include/glow/IR/IR.h
```diff
@@ -212,6 +228,29 @@ class Instruction : public Value, public TaggedListNode<Instruction> {
...
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Instruction &ins) {
```
SplitInfinity: Typically we use `I` as a name for name standalone `Instruction` variables.

include/glow/IR/IR.h
```diff
@@ -212,6 +228,29 @@ class Instruction : public Value, public TaggedListNode<Instruction> {
...
+    ins.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Instruction *const &ins) {
```
SplitInfinity: Same here.

lib/IR/IR.cpp
```diff
@@ -573,6 +580,27 @@ void IRFunction::dump(llvm::raw_ostream &OS) const {
...
+/// Dump to string.
+std::string Instruction::toString() const {
+  std::string storage;
+  llvm::raw_string_ostream os(storage);
+  dump(os);
+  return os.str();
+}
+
+/// Dump a textual representation of the function into provided output stream.
+void Instruction::dump(llvm::raw_ostream &OS) const { dumpOperands(OS); }
```
SplitInfinity: I think it would be useful to dump out the instruction name and type/kind as well.

include/glow/IR/IR.h
superwizard2019: I saw some existing dump functions in IR.h using the capitalized parameter (IRFunction). So I just wanted to keep them consistent. Should I keep this or change them all?

superwizard2019: > Is there a reason this parameter name is capitalized but the others aren't?

I saw some existing dump functions in IR.h using the capitalized parameter (IRFunction). So I just wanted to keep them consistent. Should I keep this or change them all?

include/glow/Base/Tensor.h
superwizard2019: > Hmm, reference to a pointer? Why not pass just the pointer (i.e. `const Tensor *const t`)?

I saw similar function under the Type.cpp using reference to a pointer of Type . Should I keep this or change them all?

include/glow/IR/IR.h
SplitInfinity: You used `os` everywhere else so I think you should change to `os` here.

include/glow/Base/Tensor.h
SplitInfinity: Interesting...let me ask around. I am curious as well.

opti-mix: Good catch! No, I think it is a copy&paste mistake. There is no reason to pass a const pointer by reference.

I believe it was initially written using this style because in many cases the object to be printed is passed by reference to avoid copying.

lib/IR/IR.cpp
opti-mix: Instructions are Values. And Values have a dump method already. No need to redefine it.

include/glow/IR/IR.h
```diff
@@ -305,9 +344,26 @@ class IRFunction final : public Named {
...
   void dump(llvm::raw_ostream &OS) const;
 
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const IRFunction &irf) {
+    irf.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const IRFunction *const &irf) {
+    if (!irf) {
```
opti-mix: I think it is better to assert here instead of printing "<none>"

include/glow/IR/IR.h
opti-mix: I think it is better to assert here instead of printing "none"

include/glow/IR/IR.h
```diff
@@ -99,6 +99,22 @@ class Value : public Named,
...
+  /// Print value to string.
+  std::string toString() const;
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Value &v) {
+    v.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Value *const &v) {
+    if (!v) {
```
opti-mix: I think it is better to assert here instead of printing "none"

include/glow/Graph/Node.h
```diff
@@ -110,6 +110,28 @@ class Node : public Named,
...
+  //// Print Node information to string;
+  std::string toString() const;
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Node &node) {
+    node.dump(os);
+    return os;
+  }
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Node *const &node) {
+    if (!node) {
```
opti-mix: I think it is better to assert here instead of printing "none"

include/glow/Graph/Graph.h
```diff
@@ -1061,6 +1081,26 @@ class Function final : public Named {
...
+    if (!fun) {
```
opti-mix: Same here.

include/glow/Base/Tensor.h
```diff
@@ -338,6 +339,23 @@ class Tensor final {
...
+    if (!t) {
```
opti-mix: I think it is better to assert here instead of printing "none"

include/glow/Base/Type.h
```diff
@@ -513,6 +513,15 @@ struct Type final {
     return names[(int)Ty];
   }
 
+  /// Print Type information.
+  void dump(llvm::raw_ostream &out) const;
+
+  /// Print Type information using a default output stream.
+  void dump() const;
+
+  //// Print Node information to std::string;
```
rdzhabarov: it's not a node.

tests/unittests/BackendTest.cpp
```diff
@@ -497,3 +497,13 @@ INSTANTIATE_TEST_CASE_P(JIT, BackendTest, ::testing::Values(BackendKind::CPU));
 INSTANTIATE_TEST_CASE_P(OpenCL, BackendTest,
                         ::testing::Values(BackendKind::OpenCL));
 #endif // GLOW_WITH_OPENCL
+
+// check if the dump function works.
```
rdzhabarov: doxygen comments everywhere.

superwizard2019: > doxygen comments everywhere.

OK. I added the doxygen comments to all the .h file and test functions to explain the defination and test goal.

include/glow/Base/Tensor.h
superwizard2019: OK. I will just simply use a pointer instead of a reference to a pointer

include/glow/Base/Tensor.h
```diff
@@ -347,6 +348,27 @@ class Tensor final {
...
+  /// Dump a textual representation of the Tensor to std::string.
+  std::string toString() const;
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Tensor &t) {
+    t.dump(os);
+    return os;
+  }
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                                       const Tensor *const t) {
```
opti-mix: Do we really need `const Tensor *const ` compared to `const Tensor *` we typically use everywhere? There are more places like this in your PR.

tests/unittests/BackendTest.cpp
```diff
@@ -499,3 +499,13 @@ INSTANTIATE_TEST_CASE_P(JIT, BackendTest, ::testing::Values(BackendKind::CPU));
 INSTANTIATE_TEST_CASE_P(OpenCL, BackendTest,
                         ::testing::Values(BackendKind::OpenCL));
 #endif // GLOW_WITH_OPENCL
+
+/// Check if the dump function works for Type.
+TEST(BackendTest, dumpType) {
+  Module mod;
+  TypeRef tyA = mod.uniqueType(ElemKind::FloatTy, {1, 32, 32, 3});
+  tyA->dump();
```
opti-mix: It is probably better to create a custom stream object and output into this stream, so that we do not pollute the real estate of the terminal during testing. You could even take this output stream and then compare it with the result of `toString`.

include/glow/Base/Tensor.h
superwizard2019: > Do we really need `const Tensor *const ` compared to `const Tensor *` we typically use everywhere? There are more places like this in your PR.

OK. I think they are the same. I removed the redundant const.

tests/unittests/BackendTest.cpp
superwizard2019: > It is probably better to create a custom stream object and output into this stream, so that we do not pollute the real estate of the terminal during testing. You could even take this output stream and then compare it with the result of `toString`.

OK, Thanks. I created a separate `llvm::raw_string_ostream` to verify the results. So the terminal looks clean.

tests/unittests/GraphTest.cpp
```diff
@@ -1672,3 +1672,59 @@ TEST(Graph, GetOutputSaveTest) {
...
+  IRFunction M(F);
+  PlaceholderBindings bindings;
+  Node *K = MD.createPlaceholder(ElemKind::FloatTy, {4, 320, 200, 100, 3},
+                                 "input", true);
+  // Test Node
+  std::string storageN1;
+  llvm::raw_string_ostream osN1(storageN1);
+  K->dump(osN1);
+  std::string mesN = K->toString();
+  std::string expectMes = "Placeholder\nname : \"input\"\noutput : float<4 x "
```
opti-mix: Nit: It may look nicer if you'd use raw string literals. See
https://en.cppreference.com/w/cpp/language/string_literal and https://stackoverflow.com/questions/20508534/c-multiline-string-raw-literal/20508617. Try it out and if it looks better - use it and update all such places in your tests!

include/glow/Base/Tensor.h
```diff
@@ -347,6 +348,26 @@ class Tensor final {
     return *this;
   }
 
+  /// Dump a textual representation of the Tensor into provided output stream.
+  void dump(llvm::raw_ostream &os) const;
```
opti-mix: For Tensors, we may want to additionally provide a couple more dump functions which are similar to `Handle::dump` versions that take an `unsigned maxNumElem` parameter. This makes it easier to dump as many elements of the tensor as you need.

include/glow/Graph/Graph.h
```diff
@@ -163,9 +163,27 @@ class Module final {
...
+  /// Dump a textual representation of the Module into default output stream.
   void dump() const;
 
+  /// Dump a textual representation of the Module to std::string.
+  std::string toString() const;
+
+  /// Dump a textual representation of the Module into provided output stream.
+  void dump(llvm::raw_ostream &os) const;
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
```
opti-mix: Is there a need to declare the `operator <<` inside the class as a friend? It seems like in Glow we declare it usually define such operators outside of the class. See e.g. `Type.h`.

More over, it probably makes sense to move the implementation into the cpp file.

tests/unittests/GraphTest.cpp
superwizard2019: > Nit: It may look nicer if you'd use raw string literals. See
> https://en.cppreference.com/w/cpp/language/string_literal and https://stackoverflow.com/questions/20508534/c-multiline-string-raw-literal/20508617. Try it out and if it looks better - use it and update all such places in your tests!

OK, thanks. Now it looks the same as what the terminal shows.

include/glow/Base/Tensor.h
superwizard2019: > For Tensors, we may want to additionally provide a couple more dump functions which are similar to `Handle::dump` versions that take an `unsigned maxNumElem` parameter. This makes it easier to dump as many elements of the tensor as you need.

OK.

include/glow/Graph/Graph.h
superwizard2019: > Is there a need to declare the `operator <<` inside the class as a friend? It seems like in Glow we declare it usually define such operators outside of the class. See e.g. `Type.h`.
> 
> More over, it probably makes sense to move the implementation into the cpp file.

OK. I switched them from friend function to outside and moved the implementations in the cpp file.

SplitInfinity:
Great job @superwizard2019!

Before you land, remember to squash everything into one commit. I can show you again if you need help.

Issue #2686, opened 2019-04-09
Teach Glow how to support custom alginment requirements for tensor dimensions


opti-mix:
Some backends may have specific requirements on the alignment of data inside tensors. E.g. each row of the innermost dimension to be cache line size aligned. In principle, any of the dimensions may have an alignment requirement.

Glow's `Tensor` class as of today is not able to represent this requirements, because they underlying `Type` cannot express it. `Type` only stores the actual dimensions, but not the aligned dimensions.

So, how could one use Glow Tensors to enforce any requirements on alignments?

Here are some possible alternatives:
1.  Glow Tensors could be extended to explicitly support custom strides. `Type` would have a dedicated `strides_` member in addition to `sizes_`. 
    - This change could be useful for different backends. 
    - Any code that accesses tensor elements only via logical indices and never through raw pointers should continue to work correctly.
    - But it may require a lot of changes all over the place, especially in libjit and some kernels and operations that iterate over the content of a tensor using raw pointers.
      - Everything that uses row pointers would start iterating over "padding" elements. These elementds may contain junk data and thus all such places should be updated to skip such elements.
      - libjit kernels should get additional parameters for strides, so that they can correctly compute linear indices into tensors.

 2. It is also possible to try tweaking the representation of tensors only in the specific backends that need it, without changing the `Tensor` and `Type` classes. E.g. one could introduce custom backend-specific Tensor classes. This is certainly doable, but it is very likely that this logic will be re-implemented in a number of different backends.
   - The disadvantages of this approach are:
     - it may require more memory to keep both the original and backend specific tensors.
     - it may involve a lot of copying between padded and non-padded tensors
     - none of the existing kernels can be re-used, because they work only with Glow's standard tensors.

3. It is also possible to use just the usual `Tensor` and `Type` classes. Every time when a custom-aligned tensor is needed, one would represent it e.g. as a Tensor with bigger dimensions that accommodate for the required alignments and contains some junk padding elements. Such a tensor would have a desired memory layout. But it is important to note that some of the existing operations on tensors would not produce correct results when applied to this tensor, because all those operations assume that all elements contain real data and need to be processed, which is not the case due to the added junk padding elements. The content of the original tensor would be copied into the "padded" tensor before the operation that makes use of custom alignments and would be copied back into the original tensor after that operation. 
   - The disadvantages of this approach are:
       - it may require more memory to keep both the original and the padded tensor.
       - it may involve a lot of copying between padded and non-padded tensors
[...]

PR #2970, opened 2019-05-24
[RFC] Added unpaddedSize_ to Tensor 

gcatron:
Summary:
This PR adds a new field to Tensor: unpaddedSize_.
Since we have a static batch size we copy the entire tensor to the device regardless of number of inputs. This would allow a DeviceManager to only copy the inputs if it supports it.
This means adding additional metaData to Tensor, this seemed cleaner than passing side channel data.

Documentation: NA

Test Plan: ninja test

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -69,6 +69,11 @@ private:
   /// The TensorPool that is managing this Tensor (if any).
   TensorPool *tensorPool_{nullptr};
 
+  /// Size in bytes of the unpadded region memory. This is useful  communicating
+  /// the actual size of the data, this allows for copying only inputs and not
+  /// padding to the device.
+  size_t unpaddedSize_{0};
+
   template <class ElemTy> friend class Handle;
 
   /// \returns a pointer to the tensor data buffer.
@@ -77,6 +82,10 @@ private:
 public:
   /// \returns true if it is an unowned tensor.
   bool isUnowned() const { return isUnowned_; }
+
+  /// \returns the size of the unpadded memory region. If unpaddedSize_ is not
+  /// set return the size of the entire payload.
+  size_t getUnpaddedSizeInBytes() const;
 
   /// \returns the type of the tensor.
   const Type &getType() const { return type_; }
@@ -206,9 +215,11 @@ public:
 
   /// Construct an unowned tensor provided an existing payload buffer.
   /// This constructor can be used when there is a need to work with
-  /// "externally" managed payload buffers using Tensor APIs.
-  Tensor(void *data, TypeRef ty)
-      : data_(reinterpret_cast<char *>(data)), type_(*ty), isUnowned_{false} {
+  /// "externally" managed payload buffers using Tensor APIs. Additionally \p
+  /// unpaddedSize can be set to indicate actual size of the inputs.
+  Tensor(void *data, TypeRef ty, size_t unpaddedSize = 0)
+      : data_(reinterpret_cast<char *>(data)),
+        type_(*ty), isUnowned_{false}, unpaddedSize_{unpaddedSize} {
     // Mark as unowned.
     isUnowned_ = true;
   }
--- a/lib/Base/Tensor.cpp
+++ b/lib/Base/Tensor.cpp
@@ -515,3 +515,11 @@ void Tensor::convertToType(ElemKind newTy) {
   }
   *this = std::move(tmp);
 }
+
+size_t Tensor::getUnpaddedSizeInBytes() const {
+  if (unpaddedSize_) {
+    return unpaddedSize_;
+  } else {
+    return type_.getSizeInBytes();
+  }
+}


bertmaher:
I had a pretty evil thought: do we ever check that the `Tensor` and bound `Placeholder` in `PlaceholderBindings` actually have the same type?  If we relax that restriction, we could simply bind smaller tensors.

I'm not sure that's a good idea, I'm just thinking about ways to avoid cluttering the main Tensor class.

bertmaher:
Here's a different thought that feels safer than binding wrong-sized tensors... we already create an ExecutionContext that gets passed all the way down to the execute method where we need the size; maybe we should keep a parallel map of physical sizes in that object.

bertmaher:
If we go with the design in this PR (passing the unpadded size in the Tensor) I'd like to make it a bit safer than simply adding a getter/setter.  Some suggestions:
1. This optimization only makes sense for unowned tensors; add a constructor for this case:
```
Tensor(void *data, TypeRef ty, size_t unpaddedSize);
```
2. Only provide a `getUnpaddedSize()`, and have it return `size()` if the tensor is not padded.
3. I guess I'm nervous about setting `unpaddedSize=0` to mean "the tensor is not padded".  I'd prefer to always set `unpaddedSize`, but I guess this is a bummer because you have to do some multiplications every time you allocate a Tensor.
4. Maybe add a convenience method `getUnpaddedSizeInBytes()`.

include/glow/Base/Tensor.h
```diff
@@ -63,6 +63,10 @@ class Tensor final {
   /// The type of the tensor.
   Type type_;
 
+  /// Size in bytes of the unpadded region memory. This is useful for only
+  /// copying valid inputs when the tensor is padded.
+  uint64_t unpaddedSize_{0};
```
bertmaher: `size_t` is more consistent with how we define sizes elsewhere.

lib/Base/Tensor.cpp
```diff
@@ -501,3 +501,11 @@ void Tensor::convertToType(ElemKind newTy) {
   }
   *this = std::move(tmp);
 }
+
+uint64_t Tensor::getUnpaddedSize() {
+  if (unpaddedSize_) {
+    return unpaddedSize_;
```
bertmaher: If I'm right that virtually-padded tensors are only valid if unowned, `assert` that the tensor is unowned here.

gcatron:
> Here's a different thought that feels safer than binding wrong-sized tensors... we already create an ExecutionContext that gets passed all the way down to the execute method where we need the size; maybe we should keep a parallel map of physical sizes in that object.

We can do that, I'm not sure that's cleaner though, it's trading one field of metadata on the tensor for a whole map in the executionContext.

gcatron:
> 1. This optimization only makes sense for unowned tensors; add a constructor for this case:
Unfortunately it's the opposite, in Onnxifi we only create a padded tensor when the size is smaller, if the sizes match we use an unowned tensor.

lib/Base/Tensor.cpp
gcatron: I don't think they're virtually padded they are real padded.

bertmaher:
OK, after sleeping on it I'm on board with the add-it-to-Tensor approach :-)

include/glow/Base/Tensor.h
```diff
@@ -78,6 +82,13 @@ class Tensor final {
   /// \returns true if it is an unowned tensor.
   bool isUnowned() const { return isUnowned_; }
 
+  /// \returns the size of the unpadded memory region. If unpaddedSize_ is not
+  /// set return the size of the entire payload.
+  size_t getUnpaddedSizeInBytes();
+
+  /// Set the size of unpaddedSize_ to \p size
+  void setUnpaddedSize(size_t size) { unpaddedSize_ = size; }
```
bertmaher: What do you think about killing the setter in favor of a new flavor of unowned constructor?  I always like to avoid exposing data members that can be arbitrarily twiddled.

jfix71:
Would it make sense to add this to Type instead of Tensor?

nadavrot:
CC @opti-mix who is looking into generalizing the padding support in the type system.

opti-mix:
> Would it make sense to add this to Type instead of Tensor?

I've implemented some support for providing custom strides/alignments for dimensions of tensors on my private branch ( related to #2686 ). E.g. you can say that a stride for a given dimension is a multiple of 64. If you only have 3 elements as the actual size of the dimension, the remaining 61 elements would be just "padding" elements filled with some garbage data.

I've been planning to upstream it, though not necessarily in the coming weeks. On this branch, I've added a new `strides_` member to `Type` and updated related places affected by this change.

If you think that the ability to specify strides on a per-dimension basis  would help with your issue, we could upstream it earlier than I initially planned.

But I'm under the impression that you are trying to do something different here.

bertmaher:
To provide context here, the problem we're trying to solve is that PCIe traffic sending inputs to devices is a bottleneck when we're using SparseLengthsSum operators.  The problem is that we're forced to provide a maximally-sized "indices" input, when it fact we often use fewer than the maximum number of indices.  (E.g., we might compile to a sequence length of 1000, but frequently use only 100).  So we end up transferring a bunch of zeros (or garbage) over PCIe, despite knowing (at runtime) the correct amount of data to transfer.

I think this feature is probably orthogonal to strided tensors; it's more like cheapo support for dynamic shapes ;-).

bertmaher:
> Would it make sense to add this to Type instead of Tensor?

That could be reasonable... although I'm not sure whether it's really a type-system property.  I've been thinking of it more as a storage property of the Tensor, like the type is still Int64[2000] but it's "virtually padded" as needed rather than being physically padded to the right size.

opti-mix:
I'd suggest not merging this PR yet. I've got some ideas I'd like to discuss next week.

nickgg:
Initially I preferred the Tensor approach, but I'm leaning towards adding it to Type now - what does it mean to e.g. MatMul two tensors with the same Type but different paddings?

bertmaher:
You could actually use this for matmul to express padding in the batch dimension.  There are maybe two concepts here, (1) can this dimension be padded, which could go in `Type`, and (2) how much padding is there actually, which probably has to go on `Tensor` (or some other runtime structure).

PR #2976, opened 2019-05-24
Teach Image Classifier to deal with different layouts

Max-K-cadence:
Summary:
Currently, Image Classifier expects output tensor to have shape
[Batch x Num Labels x 1 x 1 x ... x 1]. This patch allows Num Labels to be on an
arbitrary position, in particular, it enables glow on models with
output shape [Batch x 1 x ... x 1 x Num Labels]. The only requirement is that
the first dimension is Batch size and only one other dimension is not rqual to
one, it represents number of labels.

Testing: Ran several models with old and new layouts.

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -304,17 +304,26 @@ template <typename ElemTy> static void applySoftmax(Handle<ElemTy> H) {
 template <typename ElemTy>
 static int processAndPrintResultsImpl(Tensor *SMT,
                                       llvm::ArrayRef<std::string> imageList) {
-  // Softmax should have at least two dims: batchSize, numLabels, and then
-  // optionally trailing 1s.
+  // Softmax should have at least two dimensions: batchSize (first dimension),
+  // numLabels (any other dimension), and optionally - 1 in all other
+  // dimensions. The value of numLabels should be greater than 1.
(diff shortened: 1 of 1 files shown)


jfix71:
@Max-K-cadence Looks like CircleCI is having some issues here -- "fatal: refusing to merge unrelated histories". Perhaps you could try rebasing your commit to the latest master even though it says there are no conflicts? Looks like your branch for this PR is a bit old.

Max-K-cadence:
Hi @jfix71, I'm not sure why it is happening, will try to figure it out.

Max-K-cadence:
Rebased. The testing is still not working because of this:

Collecting cmake
  Downloading https://files.pythonhosted.org/packages/55/2a/021e5d8c03928ec7bad199d8a596ab5bc3811b116e40f6454891c3b5b821/cmake-3.14.4.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-ooQu04/cmake/setup.py", line 7, in <module>
        from skbuild import setup
    ImportError: No module named skbuild

jfix71:
@Max-K-cadence It was a CI issue that appears to be resolved. Thanks for rebasing, looks good to go now.

PR #2978, opened 2019-05-24
[Runtime] Add config file for Device Config used in Runtime.

beicy:
Summary:
Enable Loader to get device configs from a config yaml file. 
Now there are 2 ways in Loader to get the DeviceConfigs: 
1) If the config file is provided, use the file;
2) otherwise, use the "num-devices" which is 1 by default.

Documentation:

Test Plan:
The config file example is :
```
---
- name:     Device1
  kindName: CPU
  parameters: |
    "platformID":"1"
    "deviceID" : "0"
- name:     Device2
  kindName: CPU
  parameters: |
    "platformID":"1"
...
```
The command line: 
`./bin/text-translator -m "en2gr" -cpu -cpu-memory=500000 -load-device-configs="configs.yaml" <<< "I love music ."`

[Optional Fixes #issue]

### DIFF
--- a/include/glow/Support/Support.h
+++ b/include/glow/Support/Support.h
@@ -119,6 +119,31 @@ inline void report(llvm::StringRef str) { report(str.data()); }
 /// can be inspected under debugger.
 std::string legalizeName(llvm::StringRef name);
 
+/// Data structure for multi string format used in yaml file.
+struct MultiLineStr {
+  std::string str;
+};
+
+/// Data structure used to read the yaml file for Device Configs.
+struct DeviceConfigHelper {
+  /// Device Name.
+  std::string name_;
+  /// BackendKind name.
+  std::string kindName_;
+  /// A string with multi lines. Each line represents a param.
+  MultiLineStr parameters_;
+  DeviceConfigHelper() = default;
+  DeviceConfigHelper(std::string &name, std::string &kindName)
+      : name_(name), kindName_(kindName) {}
+  DeviceConfigHelper(std::string &kindName, std::string &name,
+                     MultiLineStr &parameters)
+      : name_(name), kindName_(kindName), parameters_(parameters) {}
+};
+
+/// Deserialize quantization infos from the file \p fileName.
+std::vector<DeviceConfigHelper>
+deserializeDeviceConfigFromYaml(llvm::StringRef fileName);
+
 /// Printf-like formatting for std::string.
 const std::string strFormat(const char *format, ...)
 #ifndef _MSC_VER
--- a/tests/unittests/SupportTest.cpp
+++ b/tests/unittests/SupportTest.cpp
@@ -18,6 +18,10 @@
 #include "glow/Testing/StrCheck.h"
 #include "gtest/gtest.h"
 
+#ifndef GLOW_DATA_PATH
+#define GLOW_DATA_PATH
+#endif
+
 using namespace glow;
 using glow::StrCheck;
 
@@ -45,4 +49,32 @@ TEST(Support, legalizeName) {
   // Check that a legal name won't be converted.
   std::string str3 = legalizeName("abc_1aBc");
   EXPECT_TRUE(str3.compare("abc_1aBc") == 0);
+}
+
+/// Check the reading Device config from a yaml file.
+TEST(Support, loadYamlFile) {
+  std::string yamlFilename(GLOW_DATA_PATH "tests/runtime_test/cpuConfigs.yaml");
+  std::vector<DeviceConfigHelper> lists;
+  lists = deserializeDeviceConfigFromYaml(yamlFilename);
+  EXPECT_EQ(lists.size(), 2);
+  // Check the loading items.
+  // The config file is:
+  //---
+  //- name:     Device1
+  //  kindName: CPU
+  //  parameters: |
+  //  "platformID":"1"
+  //    "deviceID" : "0"
+  //    - name:     Device2
+  //  kindName: CPU
+  //  parameters: |
+  //  "platformID":"1"
+  //...
+  EXPECT_EQ(lists[0].kindName_, "CPU");
+  EXPECT_EQ(lists[0].name_, "Device1");
+  EXPECT_EQ(lists[0].parameters_.str,
+            "\"platformID\":\"1\"\n\"deviceID\" : \"0\"\n");
+  EXPECT_EQ(lists[1].kindName_, "CPU");
+  EXPECT_EQ(lists[1].name_, "Device2");
+  EXPECT_EQ(lists[1].parameters_.str, "\"platformID\":\"1\"\n");
 }
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -31,6 +31,8 @@
 #include "llvm/Support/Timer.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <sstream>
+
 using namespace glow;
 
 /// -enable-rowwise : Command line option to enable rowwise quantized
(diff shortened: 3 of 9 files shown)


lib/Quantization/Serialization.cpp
```diff
@@ -35,11 +35,33 @@ template <> struct MappingTraits<glow::NodeQuantizationInfo> {
   }
 };
 
+template <> struct BlockScalarTraits<glow::MultiLineStr> {
```
rdzhabarov: that does not belong to quantization/serialization. Could you move that to a separate configs, lib/configs

rdzhabarov: also add a test for this.

tools/loader/Loader.cpp
```diff
@@ -269,6 +278,43 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
   GLOW_UNREACHABLE("Unknown node name.");
 }
 
+/// Helper to get the BackendKind type from a backend's \p name. Need to be
+/// updated if a new backend comes.
+static BackendKind getBackendKindFromName(std::string &name) {
+  if (name == "CPU")
```
opti-mix: LLVM has `llvm::StringMap` for it.

tools/loader/Loader.cpp
```diff
@@ -269,6 +278,43 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
...
+    return BackendKind::Interpreter;
+  if (name == "OpenCL")
+    return BackendKind::OpenCL;
+  if (name == "Habana")
+    return BackendKind::Habana;
+  GLOW_UNREACHABLE("Unknown backendKind name.");
+}
+
+/// Helper to get the parameters in DeviceConfig from \p str. The \p str has
+/// multiply lines, and each line with this format : "str1" : "str2".
```
opti-mix: mutiply->multiple.

beicy:
> * Do we really need this multi-line syntax? Or may be YAML has a nicer syntactic alternative for this?
> * It could be that you do more work than needed and YAML parser can do more for you. @rdzhabarov had some experience with it for quantization. May be he can suggest a better way to use the YAML parser?

Discussed with @rdzhabarov  about the yaml file. So far, we haven't found a better way...

beicy:
Added an unit test for checking yaml file format, and an expensive test for the whole flow.

tests/runtime/configs.yaml
```diff
@@ -0,0 +1,12 @@
+---
+- name:     Device1
+  kindName: CPU
```
gcatron: minor nit, if we are going to use a CPU maybe rename the parameters so it's clear they are test parameters, or put a comment that these parameters don't mean anything to the CPU backend.

include/glow/Support/Support.h
```diff
@@ -119,6 +119,26 @@ inline void report(llvm::StringRef str) { report(str.data()); }
 /// can be inspected under debugger.
 std::string legalizeName(llvm::StringRef name);
 
+struct MultiLineStr {
+  std::string str;
+};
+
+struct DeviceConfigHelper {
```
nadavrot: Should this be in "Support.h" or some other header file that is related to device and device configurations?

include/glow/Support/Support.h
```diff
@@ -119,6 +119,26 @@ inline void report(llvm::StringRef str) { report(str.data()); }
 /// can be inspected under debugger.
 std::string legalizeName(llvm::StringRef name);
 
+struct MultiLineStr {
```
nadavrot: Doxygen comments please.

include/glow/Support/Support.h
```diff
@@ -119,6 +119,26 @@ inline void report(llvm::StringRef str) { report(str.data()); }
...
+  std::string str;
+};
+
+struct DeviceConfigHelper {
+  std::string name_;
+  std::string kindName_;
+  MultiLineStr parameters_;
+  DeviceConfigHelper() = default;
```
nadavrot: Doxygen comment on all members and fields please.

tools/loader/Loader.cpp
```diff
@@ -271,6 +280,43 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
   GLOW_UNREACHABLE("Unknown node name.");
 }
 
+/// Helper to get the BackendKind type from a backend's \p name. Need to be
```
nadavrot: It looks like details of the hardware leak into the loader. This means that every time we add a new hardware backend we would need to update this file. Is there a way to avoid it?

@gcatron was working on something simiar.

tools/loader/Loader.cpp
```diff
@@ -401,11 +447,30 @@ Loader::Loader(int argc, char **argv) {
     caffe2NetWeightFilename_ = modelPathOpt[1];
   }
   M_.reset(new Module);
+
   std::vector<std::unique_ptr<runtime::DeviceConfig>> configs;
-  for (unsigned int i = 0; i < numDevices; ++i) {
-    auto config = llvm::make_unique<runtime::DeviceConfig>(ExecutionBackend);
-    configs.push_back(std::move(config));
+
```
nadavrot: This adds a lot of logic to a function that is already long. Is there a way to refactor things and make the code more readable?

tests/unittests/SupportTest.cpp
```diff
@@ -18,6 +18,10 @@
 #include "glow/Testing/StrCheck.h"
 #include "gtest/gtest.h"
 
+#ifndef GLOW_DATA_PATH
```
nadavrot: What is this?

tools/loader/Loader.cpp
```diff
@@ -271,6 +280,43 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
...
+/// Helper to get the parameters in DeviceConfig from \p str. The \p str has
+/// multiple lines, and each line with this format : "str1" : "str2".
+static llvm::StringMap<std::string> getBackendParams(std::string &str) {
+  llvm::StringMap<std::string> ret{};
+  std::string s;
+  std::istringstream f(str.c_str());
+  while (getline(f, s, '\n')) {
+    size_t pos1, pos2, pos3, pos4;
+
```
nadavrot: comments would be useful here.

tests/unittests/SupportTest.cpp
beicy: This is required for the fbcode setup.

include/glow/Support/Support.h
beicy: It is only used in Support.cpp for reading yaml file.

tools/loader/Loader.cpp
gcatron: I had a chat with @rdzhabarov last week related to this. When we move to dynamic backend registration there will be a string:backendKind map that can be used. I think this solution is a temporary one until we have that map.

jfix71: We already essentially have this map via `llvm::cl::opt<BackendKind> ExecutionBackend` here in the Loader. If we must have a map like this too, I think we should unify them somehow.

tests/unittests/SupportTest.cpp
nadavrot: Ok. Got it. Thank you.

tools/loader/Loader.cpp
beicy: I think we can have a follow up PR for this, since some tests command line maybe changed as well.

tools/loader/Loader.cpp
beicy: Get this piece of code into a function.

tools/loader/Loader.cpp
```diff
@@ -271,6 +280,73 @@ static Kinded::Kind getKindFromNodeName(llvm::StringRef nodeName) {
...
+        s.substr(pos3 + 1, pos4 - pos3 - 1);
+  }
+  return ret;
+}
+
+/// If the device config file \p loadDeviceDoncfigsFile available, load \p
+/// configs from the file. Otherwise, create \p numDevices number of devices
+/// based on \p backendKind.
+static void generateDeviceConfigs(
+    std::vector<std::unique_ptr<runtime::DeviceConfig>> &configs,
```
gcatron: can generateDeviceConfigs return the vector of configs? That way we don't have to pass in a reference.

beicy: Done! Thanks!

Issue #2481, opened 2019-03-04
More general implementation of Row/ChannelWiseWQuantization


ayermolo:
I would like to propose a generic approach for ‘RowWise’ quantization, that gives in addition the possibility for the backend to perform some customizations. I’ve created a WIP PR with a prototype implementation for review/discussion https://github.com/pytorch/glow/pull/2480.

Currently a specific ‘RowWise’ node type variant has been created for each operation concerned by rowwise quantization: RowwiseQuantizedFullyConnected/RowwiseQuantizedSparseLengthsWeightedSum. Going by this approach if we want to apply it to regular Convolution we would create ChannelWiseQuantizedConvolution, or something like that.

Our proposal is to add a generic ChannelWiseQuantization node instead of creating a specific node variant each time we want to support rowwise quantization. The input of ChannelWiseQuantization would be quantized weights, original floating point weight,  scales array, and offsets array. The latter two can either be in tensor or internal vector type to the node.

We keep the original floating point weight tensor since some backend may want to do a custom quantization. Keeping original weights around can be removed down the road if we change infrastracture of glow to allow backends hooks to quantize on node level.

I included Interpreter and CPU backend implementations for illustrative purposes.

During LLVM generation it would allow:
-	backends to use this information to select appropriate function for the target device, and pass relevant information in to it.
-	various optimizations to deal with only one node type, vs bunch of different node types of RowwiseQuantized variety

jfix71:
@ayermolo Thanks for the proposal and example PR. I agree we could do with a redesign on supporting channel/row-wise quantization.

The proposed approach feels a bit hacky to me. There's the `replaceAllUsesOfWith()` at IRGen time. And iterating over Instruction users at LLVM generation time, and during the execution of the Interpreter kernel. And then using the Instruction as basically a pointer to the weights/scales/offsets, which otherwise does nothing.

I wonder if instead of adding new a Node/Instruction, why can't we just handle row-wise quantization via another `ElemKind`? E.g. we would have a new `ElemKind::Int8RowwiseQTy`. Currently, Types with our other quantized ElemKinds (e.g. `Int8QTy`) have the scale and offset as scalars built into the Type. Here we could add `scales`/`offsets` vectors to the Type instead

Do you see some benefits to your proposal compared to adding the new `ElemKind`?

ayermolo:
Umm yeah on a hacky side for sure. Kind of tried to work around current limitations of GLOW.

[...]

PR #2980, opened 2019-05-24
[New Operator] [part 1] Import ConvolutionChannelwiseQuantizedNode

jackm321:
Summary:
* Import onnxTensors with multiple quantization parameters. If an onnxTensor has more than one quantization parameter, create separate `Constants` for the scales and offsets so that nodes that consume these can access them.
* Create a ConvolutionChannelwiseQuantizedNode which is a quantized convolution node that has independent quantization parameters for each filter or group of filters. These parameters are stored in separate scales and offsets tensors similar to `RowwiseQuantizedFullyConnectedNode`
* This PR only handles importing the `ConvolutionChannelwiseQuantizedNode` and it's weights and qparams, implementation of this op will be done in a follow up PR. I'm not sure if we will have an instruction for this node or simply default to lowering it to multiple quantized convolutions.

Documentation:
* Doyxgen

Test Plan:
Add operator test

### DIFF
--- a/include/glow/Importer/Caffe2ModelLoader.h
+++ b/include/glow/Importer/Caffe2ModelLoader.h
@@ -60,8 +60,16 @@ class Caffe2ModelLoader
   /// in the network.
   llvm::Error loadOperator(const caffe2::OperatorDef &op);
 
-  /// Reads a network (weights or structure) from the serialized protocol buffer
-  /// file.
+  /// Load the Conv or ConvRelu operators.
+  llvm::Error loadConv(const caffe2::OperatorDef &op,
+                       ArgumentDictionaryTy &dict);
+
+  /// Load the Int8Conv or Int8ConvRelu operators.
+  llvm::Error loadConvQuantized(const caffe2::OperatorDef &op,
+                                ArgumentDictionaryTy &dict);
+
+  /// Reads a network (weights or structure) from the serialized protocol
+  /// buffer file.
   llvm::Expected<caffe2::NetDef> loadProtoFile(const std::string &filename);
 
   /// loadInputs calls this function for each member in its target arguments.
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -109,17 +109,19 @@ static bool verifyConvolution(NodeValue src, NodeValue dest, NodeValue filter,
                               llvm::ArrayRef<unsigned_t> kernels,
                               llvm::ArrayRef<unsigned_t> strides,
                               llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
-                              unsigned_t dilation) {
+                              unsigned_t dilation, bool checkBiasType = true) {
   const Node *parent = dest.getNode();
   bool isValid = checkType(src, dest.getElementType(), parent);
   isValid &= checkType(src, filter.getElementType(), parent);
-  // Non quantization type check.
-  if (src.getElementType() == ElemKind::FloatTy) {
-    isValid &= checkType(bias, ElemKind::FloatTy, parent);
-  }
-  // Quantization type check.
-  if (src.getElementType() == ElemKind::Int8QTy) {
-    isValid &= checkType(bias, ElemKind::Int32QTy, parent);
+  if (checkBiasType) {
+    // Non quantization type check.
+    if (src.getElementType() == ElemKind::FloatTy) {
+      isValid &= checkType(bias, ElemKind::FloatTy, parent);
+    }
+    // Quantization type check.
+    if (src.getElementType() == ElemKind::Int8QTy) {
+      isValid &= checkType(bias, ElemKind::Int32QTy, parent);
+    }
   }
   ShapeNHWC idim(src.getType()->dims());
   ShapeNHWC odim(dest.getType()->dims());
@@ -407,6 +409,41 @@ bool PadNode::verify() const {
 bool ConvolutionNode::verify() const {
   return verifyConvolution(getInput(), getResult(), getFilter(), getBias(),
                            Kernels_, Strides_, Pads_, Group_, Dilation_);
+}
+
+bool ChannelwiseQuantizedConvolutionNode::verify() const {
+  bool isValid = expectCompareTrue("Only groupwise quantization is supported.",
+                                   getGroupwise(), true, this);
+
+  if (!isValid) {
+    return false;
+  }
+
+  isValid = verifyConvolution(getInput(), getResult(), getFilter(), getBias(),
+                              Kernels_, Strides_, Pads_, Group_,
+                              /* dilation */ 1, /* checkBiasType */ false);
+
+  isValid &= checkType(getBias(), ElemKind::FloatTy, this);
+  isValid &= checkType(getInput(), ElemKind::Int8QTy, this);
+
+  // check qparam types
+  isValid &= checkType(getOffsets(), ElemKind::Int32ITy, this);
+  isValid &= checkType(getScales(), ElemKind::FloatTy, this);
+
+  // check qparam dimensions
+  isValid &= expectCompareTrue("Offsets must be a 1D vector",
+                               getOffsets().dims().size(), size_t(1), this);
+  isValid &= expectCompareTrue("Scales must be a 1D vector",
+                               getScales().dims().size(), size_t(1), this);
+
+  // check qparam sizes
+  isValid &=
+      expectCompareTrue("There must be one filter offset qparam per group",
+                        getOffsets().dims()[0], size_t(getGroup()), this);
+  isValid &=
+      expectCompareTrue("There must be one filter scale qparam per group",
+                        getScales().dims()[0], size_t(getGroup()), this);
+  return isValid;
 }
 
 bool Convolution3DNode::verify() const {
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -57,7 +57,8 @@ namespace {
 /// Creates tensor \p T from the input \p in. Note, there is no data associated
 /// with the Tensor. This method makes sure that the tensor is created with the
 /// proper shape and element type.
-llvm::Error setTensorType(const caffe2::TensorProto &in, Tensor *T) {
+llvm::Expected<LoadWeightResult>
+createAndSetTensorType(const caffe2::TensorProto &in) {
   std::vector<size_t> dim;
   for (auto d : in.dims()) {
     if (d == 0) {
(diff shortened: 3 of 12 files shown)


ayermolo:
@jackm321 FYI I did this [WIP] PR: https://github.com/pytorch/glow/pull/2480
In case you find something there useful to re-use or whatever.

jackm321:
@ayermolo thanks I had not seen that. That's cool, adding a `ChannelWiseWeightQuantizationNode` instead of a new type of conv. I'm tempted to switch to that but I think at least for now I'm going to stick with adding a new node unless you strongly disagree with this. I would need to think how this would work with the convolution bias because right now Glow receives these unquantized for channelwise quantized convs. Another issues is support for groupwise quantization instead of channelwise quantization. In the longer run, it would be nice to add multiple quantization param support to glow's `Type` directly and remove this complexity.

ayermolo:
Yeah we had discussion on couple approaches: https://github.com/pytorch/glow/issues/2481

Adding to the type would probably be a cleaner approach, but it's a much more fundamental change.

In either case for node to be useful to different backends the quantization would needs to be delayed to/post legalization phase.
@tlepley-cadence 
Anyway thanks for working on this.

lib/Graph/Nodes.cpp
```diff
@@ -409,6 +411,40 @@ bool ConvolutionNode::verify() const {
...
 
+bool ConvolutionChannelwiseQuantizedNode::verify() const {
+  bool isValid = expectCompareTrue("Only groupwise quantization is supported.",
+                                   getGroupwise(), true, this);
+
+  if (!isValid) {
+    return false;
+  }
+
+  isValid = verifyConvolution(getInput(), getResult(), getFilter(), getBias(),
```
jfix71: This call just checks that the source and dest element types are the same. Do we expect that the `ConvolutionChannelwiseQuantizedNode` will ever operate with float inputs/output?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -95,21 +99,53 @@ llvm::Error setTensorType(const caffe2::QTensorProto &in, Tensor *T) {
...
-    return llvm::Error::success();
+    result.t->reset(ElemKind::Int32QTy, dim, scale, offset);
   } else {
-    RETURN_ERR("Only uint8 and int32 qtensors are supported");
+    RETURN_ERR("Only int8, uint8, and int32 qtensors are supported");
   }
-}
+
+  return llvm::Expected<LoadWeightResult>(std::move(result));
+} // namespace
```
jfix71: Accidental `// namespace` due to `clang-format`?

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -60,8 +60,16 @@ class Caffe2ModelLoader
   /// in the network.
   llvm::Error loadOperator(const caffe2::OperatorDef &op);
 
-  /// Reads a network (weights or structure) from the serialized protocol buffer
-  /// file.
+  // Load the Conv or ConvRelu operators.
+  llvm::Error loadConv(const caffe2::OperatorDef &op,
```
jfix71: Looks like this and `loadConvQuantized()` are essentially just helpers for the Caffe2ModelLoader -- perhaps we can make them just static inside Caffe2ModelLoader.cpp and not declare them here?

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -165,6 +165,49 @@ TEST(caffe2, convNHWC) {
...
+    Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename, {"input"},
+                               {&input.getType()}, *F);
+    output = EXIT_ON_ERR(caffe2LD.getSingleOutput());
+  }
+
+  // High level check on the content of the graph. We have 1
+  // ConvolutionChannelwiseQuantizedNode and 1 save.
+  EXPECT_EQ(F->getNodes().size(), 2);
+  auto *saveNode = getSaveNodeFromDest(output);
+  auto *groupwiseConv = llvm::dyn_cast<ConvolutionChannelwiseQuantizedNode>(
```
jfix71: Perhaps check that all the params of the conv are set correctly, e.g. kernels/strides/pads/group/order/etc.?

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -165,6 +165,49 @@ TEST(caffe2, convNHWC) {
...
+      saveNode->getInput().getNode());
+  ASSERT_TRUE(groupwiseConv);
+
+  // We have 2 placeholders: 1 input and 1 output.
+  EXPECT_EQ(mod.getPlaceholders().size(), 2);
+  // We have 4 constants: Bias, Weights, and Weights' separate scales and
+  // offsets.
+  EXPECT_EQ(mod.getConstants().size(), 4);
```
jfix71: Should we also check the shape and/or content of the Constants are correct?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
...
-      }
+  // Transpose the weights to the right format. Glow expects to read the
+  // weights in the format CRSK.
+  // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
+  // For Caffe2 "Int8Conv" and "Int8ConvRelu", the weights always follows the
+  // "order" arg.
+  if (order != "NHWC") {
+    Tensor wT;
+    w->getPayload().transpose(&wT, NCHW2NHWC);
+    w = G_.getParent()->createConstant(w->getName(), std::move(wT));
```
jfix71: nit: I see this is copied over from the previous impl, and it's probably done elsewhere in the loaders too -- but I don't think we should be doing these sorts of transformations/duplicate Constant creation here. We should just add a Transpose node to the weight and let the Graph optimizer handle merging the Transpose in.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
...
+  Constant *w;
+  ASSIGN_VALUE_OR_RETURN_ERR(w, getConstantByName(op.input(1)));
+
+  // Transpose the weights to the right format. Glow expects to read the
+  // weights in the format CRSK.
+  // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
+  // Caffe2 "Conv" op always stores the weight as CKRS.
+  Tensor wT;
+  w->getPayload().transpose(&wT, NCHW2NHWC);
+  w = G_.getParent()->createConstant(w->getName(), std::move(wT));
```
jfix71: Same as other comment, I'd rather not create this new Constant here, and instead defer to graph optimizations.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
   return false;
 }
 
-llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
-  ArgumentDictionaryTy dict = loadArgumentMap(op);
+llvm::Error Caffe2ModelLoader::loadConv(const caffe2::OperatorDef &op,
+                                        ArgumentDictionaryTy &dict) {
+  const std::string &opName = loadOperatorName(op);
   const std::string &typeName = op.type();
```
jfix71: nit: Given `typeName` is just used once now and much lower in this function, I'd move this down to just before its use, or just use `op.type()` directly at its use site to make clear it has no impact on any logic up here.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
...
+
+  Constant *w;
+  ASSIGN_VALUE_OR_RETURN_ERR(w, getConstantByName(op.input(1)));
+
+  // Transpose the weights to the right format. Glow expects to read the
+  // weights in the format CRSK.
+  // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
+  // Caffe2 "Conv" op always stores the weight as CKRS.
+  Tensor wT;
+  w->getPayload().transpose(&wT, NCHW2NHWC);
```
jfix71: So given this is not checking `order` at all, are we only supporting `NCHW` here? If so I think it would make sense to check that `order` is `NCHW` after assigning it above, and returning an error otherwise. Then you could get rid of the logic below checking `order`.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
...
+
+    node = G_.createChannelwiseQuantizedConv(opName, finalIn, w, bias, wScales,
+                                             wOffsets, outTy, kernels, strides,
+                                             pads, group);
+  } else {
+    // If the bias isn't quantized for a non group quantized conv, quantize it.
+    const Tensor &biasTensor = bias->getPayload();
+    if (biasTensor.getElementType() == ElemKind::FloatTy) {
+      TensorQuantizationParams tqp;
+      tqp.offset = 0;
```
jfix71: Can you add a comment explaining why we use `offset = 0`?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -238,145 +274,248 @@ bool Caffe2ModelLoader::hasMultidirectionalBroadcast(
...
+      // input scale * filter scale
```
jfix71: Can you expand a little here? As is the comment doesn't bring much value, since it's clear this is what's happening on the next line.

lib/Importer/Caffe2ModelLoader.cpp
jackm321: Yeah it's good point, I think this wasn't so straight forward before we loaded things directly into `Constant`s instead of into `Tensor`s but now that we do it seems kind of weird to do it this way.

lib/Graph/Nodes.cpp
jackm321: Is the check right below this what you mean?
`checkType(getInput(), ElemKind::Int8QTy, this);`

lib/Importer/Caffe2ModelLoader.cpp
jackm321: This should be matching the original code which had the following comment
```
// Caffe2 "Conv" op always stores the weight as CKRS, while for "Int8Conv",
// and "Int8ConvRelu", the weights always follows the "order" arg.
```
Since Glow expects CRSK and `Conv` and `ConvRelu` always use CKRS this should always transpose them.

lib/Importer/Caffe2ModelLoader.cpp
jackm321: Can we make this change in a follow up diff? I'd prefer not to increase the number of changes here more.

jackm321:
@jfix71 I think I've addressed or responded to all comments now

lib/Importer/Caffe2ModelLoader.cpp
jackm321: My understanding based on the previous code is that we the weights are always stored as CKRS regardless of order so we should not need to check the order here (though it's still needed for inputs/outputs)

PR #2981, opened 2019-05-24
[build] Fix dynamic backend support on Linux

artemrakhov:
This complicated `if` in CMake broke build on my Linux machine, and it only showed up after Bert's PR #2891. The problem was that I could no longer include CPUBackend into a unit test as a direct dependency.

I tested that I can now include CPUBackend into unit test, building with both clang and gcc on Linux.

### DIFF
--- a/cmake/modules/GlowDefaults.cmake
+++ b/cmake/modules/GlowDefaults.cmake
@@ -66,11 +66,6 @@ function(make_whole_archive DST SRC)
       # In MSVC, we will add whole archive in default.
       target_link_libraries(
           ${DST} INTERFACE -WHOLEARCHIVE:$<TARGET_FILE:${SRC}>)
-    else()
-      # Assume everything else is like gcc
-      target_link_libraries(${DST} INTERFACE
-        "-Wl,--whole-archive $<TARGET_FILE:${SRC}> -Wl,--no-whole-archive")
-      set_target_properties(${DST} PROPERTIES LINK_FLAGS "-Wl,--exclude-libs,ALL")
     endif()
   else()
     target_link_libraries(${DST} INTERFACE ${SRC})


bertmaher:
> The problem was that I could no longer include CPUBackend into a unit test as a direct dependency

Yeah I think that's fundamental to how the whole-archive thing works.  You probably want to just include Backends instead.

artemrakhov:
Hm, now I get compilation errors from CircleCI.

> > The problem was that I could no longer include CPUBackend into a unit test as a direct dependency
> 
> Yeah I think that's fundamental to how the whole-archive thing works. You probably want to just include Backends instead.

But what if your unit test only tests CPUBackend? I believe it's a good practice to directly have the backend as a direct dependency.
Also, we had different behavior on MacOS. @opti-mix had the unit test including CPUBackend to compile correctly on MacOS.

bertmaher:
> But what if your unit test only tests CPUBackend? I believe it's a good practice to directly have the backend as a direct dependency.

Is the error you're getting due to multiply defined symbols?  If so, you already have the CPU backend as a dependency; you don't need to add it again.

> Also, we had different behavior on MacOS. @opti-mix had the unit test including CPUBackend to compile correctly on MacOS.

Yes, --whole-archive and friends are very system-specific.  I'm not surprised it works differently across platforms.

bertmaher:
Can you put up a draft PR that shows the problem you're encountering?  I suspect this change will actually break dynamic registration...

rdzhabarov:
As we talked offline if linker sees a lib linked with --whole-archive and the same lib without whole-archive, order of those will matter for linker on linux (does not matter on macos).

I'd look into how you specify dependencies. Changes in this PR should not be valid.

artemrakhov:
Closing this PR since it didn't work. We must remember to never depend on any individual backend.

PR #2985, opened 2019-05-28
Back out "[glow][PR] ONNXModelLoader can automatically create input placeholders."

putivsky:
Summary: Original commit changeset: 8ea0dbf6837a

Differential Revision: D15521635


jackm321:
@putivsky context on why it's being backed out?

putivsky:
Originally I have thought that loadInputs would create Placeholders only for input tensor, which is not a case. It it were the case then it would be easily to load ONNX models and prepare bindings for inference. Now I have to wait till the caller provides input tensors map and use it for creating input placeholders and bindings.

PR #2993, opened 2019-05-29
Auto assign placeholders for inputs without weights.

putivsky:
Summary:
ONNX Loader can automatically create placeholders for the existed inputs without weights.
Usually it's only one input. If caller provides names and types for inputs code would work as before,
otherwise algorithm in constructor will create non-trainable placeholders for such inputs.

Differential Revision: D15538665

### DIFF
--- a/tests/unittests/OnnxImporterTest.cpp
+++ b/tests/unittests/OnnxImporterTest.cpp
@@ -1642,3 +1642,18 @@ TEST(onnx, batchNormPR2304) {
       llvm::dyn_cast<BatchNormalizationNode>(trNode->getInput().getNode());
   EXPECT_NE(nullptr, bnNode);
 }
+
+/// Test constructor for auto loading inputs case.
+TEST(onnx, autoLoadInputs) {
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+  std::string netFilename(GLOW_DATA_PATH
+                          "tests/models/onnxModels/batchNormPR2304.onnxtxt");
+  auto *F = mod.createFunction("main");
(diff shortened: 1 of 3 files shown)


tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1636,3 +1636,19 @@ TEST(onnx, batchNormPR2304) {
...
+  auto &mod = EE.getModule();
+  std::string netFilename(GLOW_DATA_PATH
+                          "tests/models/onnxModels/batchNormPR2304.onnxtxt");
+  auto *F = mod.createFunction("main");
+  Tensor inputTensor(ElemKind::FloatTy, {1, 2, 10, 10});
+  llvm::StringRef inputName = "input";
+  ONNXModelLoader onnxLD(netFilename, {}, {}, *F);
+  auto inputs = onnxLD.getInputVarsMapping();
+  EXPECT_EQ(inputs.size(), 1);
+  EXPECT_EQ(inputs["input"]->getName(), inputName);
```
jfix71: nit: Not sure we get much signal from this check.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1636,3 +1636,19 @@ TEST(onnx, batchNormPR2304) {
...
+  EXPECT_TRUE(inputTensor.getType().isEqual(inputs["input"]->getType()));
```
jfix71: nit: since we already have `inputName`, could just use `inputs[inputName]`. And actually, you probably should get `inputs["input"]` into its own variable before this and assert it's not null, then use the variable here.

jfix71:
@putivsky It looks like you are accidentally backing out a previously landed commit here.

PR #2996, opened 2019-05-29
Wire llvm::Errors into CompiledFunction

jackm321:
Summary:
* Return llvm::Error from CompiledFunction::execute() and handle results in DeviceManagers. Right now `execute()` methods mostly behave the same, will add more legit error handling internal to each in followups.
* Remove unused CompiledFunction methods `setupRuns`, `beforeRun`, `afterRun`, `tearDownRuns`.

Documentation:
doxygen

Test Plan:
CI

### DIFF
--- a/include/glow/Backend/CompiledFunction.h
+++ b/include/glow/Backend/CompiledFunction.h
@@ -19,8 +19,7 @@
 #include "glow/Backend/BackendUtils.h"
 #include "glow/ExecutionContext/ExecutionContext.h"
 #include "glow/Graph/Nodes.h"
-
-#include <unordered_map>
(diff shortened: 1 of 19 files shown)


jackm321:
@jfix71 I think compile _shouldn't_ fail if we have a legal graph, something must have gone wrong. In practice, I'm not sure if this is true for all backend's compilation processes.

jfix71:
Yeah so that's the contract we want to enforce for the backend, i.e. if `Backend::isOpSupported()` returns true then it shouldn't ever fail when handed a graph via `Backend::compile()`. But I'm not sure how realistic it is 🤷‍♂

PR #2999, opened 2019-05-29
[habana] Sink Transpose below Quantized Node.

rdzhabarov:
Summary:
This is direct porting from the private branch.

From the original PR.

```
Sink TransposeNode below QuantizedNode. Yields big improvement in vision topologies' performance.
Motivations:

This way, TransposeNode has to transpose fewer bytes.
In Habana GC's current version, QuantizeNode will have better performance when FCD is large. In typical case, this sequence is encountered at the beginning of vision topologies, where the FCD=W so it is large, and after the transpose FCD=C=3 so it is small.
```

cc: @sergeigofman

Documentation:
n/a

Test Plan:
CI/local run

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -1314,6 +1314,7 @@ bool HabanaBackend::isOpSupported(const NodeInfo &NI) const {
     case Kinded::Kind::SubNodeKind:
     case Kinded::Kind::TileNodeKind:
     case Kinded::Kind::ConcatNodeKind:
+    case Kinded::Kind::TransposeNodeKind:
       return true;
     case Kinded::Kind::RescaleQuantizedNodeKind:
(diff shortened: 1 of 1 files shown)


rdzhabarov:
will fix clang-format.

jfix71:
This makes sense, but why only implement it as a backend-specific transform? I would think all backends would want this...

bertmaher:
Can you give this PR a descriptive title?

bertmaher:
> This makes sense, but why only implement it as a backend-specific transform? I would think all backends would want this...

It's probably reasonable to add this everywhere, but we actually have perf evidence on Habana, so I'm kinda fine with landing it here right now.

rdzhabarov:
@jfix71 Sinking transpose below quantize makes sense for general case. I'd like to minimize back and forth between private branch and this change.

I'll create an issue and we can follow up on that.

PR #3001, opened 2019-05-30
Add some glog [1/n]

jackm321:
Summary:
I am slowly going through the code replacing `llvm::errs()`, `llvm::outs()`, some `assert`s (mostly in the runtime) and eventually all `GLOW_ASSERT`s with glog equivalents. Once I have done one or two of these, I'll finish writing up some guidelines for when and where to use glog, assert, llvm::Error, etc.

Documentation:
n/a

Test Plan:
CI

### DIFF
--- a/lib/Runtime/Executor/ThreadPoolExecutor.cpp
+++ b/lib/Runtime/Executor/ThreadPoolExecutor.cpp
@@ -29,7 +29,7 @@ namespace runtime {
 
 void InflightBarrier::decrement(unsigned decr) {
   std::unique_lock<std::mutex> lock(mtx_);
-  assert(count_ >= decr && "Barrier decrement cannot be less than count!");
+  DCHECK_GE(count_, decr) << "Barrier decrement cannot be less than count!";
   count_ -= decr;
 
   // If count_ has hit zero, wake up all threads that are waiting.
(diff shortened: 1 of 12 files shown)


lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -178,9 +176,8 @@ bool ExecutionState::incrementNodeParentsDone(const DAGNode *node,
   // been created in the constructor.
   auto it = nodeParentsDone_.find(node);
 
-  if (it == nodeParentsDone_.end()) {
-    assert(!"Node parents done counter should exist but not found!");
-  }
+  DCHECK(it != nodeParentsDone_.end())
```
jfix71: `DCHECK_NE()`?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -146,9 +146,8 @@ ExecutionState::getUniqueNodeContextPtr(const DAGNode *node) {
   // constructor.
   auto ctxIt = inputCtxs_.find(node);
 
-  if (ctxIt == inputCtxs_.end()) {
-    assert(!"Input bindings not found but should exist!");
-  }
+  DCHECK(ctxIt != inputCtxs_.end())
```
jfix71: `DCHECK_NE()`?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
jackm321: In order to do DCHECK_NE, the operands need to have operator== and also operator<< and hashmap iterator doesn't have this I guess. But also, I don't think _NE is that useful here, since the main thing we need to know is that it's not at end() not what the value of the iterator is.

PR #3002, opened 2019-05-30
[RuntimeBundle::create] change the allocation of placeholders to be contiguous

### LINKED ISSUE #2819: runtime::RuntimeBundle::create should allocate all outputs contiguously. Same for inputs.
Right now `runtime::RuntimeBundle::create` allocates memory for placeholders (i.e. inputs and outputs) in a random order.

It would be useful if it would allocate all outputs contiguously, i.e. would group all outputs together. The same approach should be taken for inputs. 

If done this way, it is easier to copy all inputs to the device as in principle it may require only a single DMA of a big contiguous block of memory. And reading results from the device would also need just a single transfer of a big contiguous block of memory.

superwizard2019:
Summary:
- modify the `runtime::RuntimeBundle::create()` to allocate the placeholders contiguously (Input|InputOutput|Output). So in principle, it may require only a single DMA of a big contiguous block of memory to move input or output.

- adjust the `AllocationsInfo::allocateWeightVars()` to match the order.

This PR fixes #2819 .

Test Plan:
Add test to BackendTest to make sure placeholders are allocated in the contiguous order.

### DIFF
--- a/include/glow/Backend/BackendUtils.h
+++ b/include/glow/Backend/BackendUtils.h
@@ -120,5 +120,49 @@ public:
 };
 } // namespace runtime
 
+/// If \p PH is an output placeholder in the Function \p F,
+/// \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH, const Function &F);
+
+/// If \p PH is an input placeholderin the Function \p F,
+/// \returns true.
+/// This is determined by checking if the PH is the input to a saveNode or is
+/// used by a non saveNode.
+bool isInput(const Placeholder *PH, const Function &F);
+
+/// If \p PH is an output placeholder in the IRFunction \p F,
+/// \returns true.
+/// This is determined by checking if the PH has weights which are referenced by
+/// other Instructions as OperandKind::InOut or OperandKind::Out.
+bool isOutput(const Placeholder *PH, const IRFunction &F);
+
+/// If \p PH is an input placeholder in the IRFunction \p F,
+/// \returns true.
+/// This is determined by checking if the PH is always used as an @in parameter
+/// by the current function.
+bool isInput(const Placeholder *PH, const IRFunction &F);
+
+/// Contains information for placeholder during allocation.
+struct PlaceholderInputOutputInfo {
+  /// The placeholder address.
+  const Placeholder *addr;
+  /// Is the placeholder an input for the function.
+  bool isInput;
+  /// Is the placeholder an onput for the function.
+  bool isOutput;
+};
+
+using ContiguousPlaceholders = std::vector<PlaceholderInputOutputInfo>;
+
+/// Convert placeholders to be ordered as input|inputOutput|output|neither.
+/// Packed into {Placeholder *, isInput, isOutput} as
+/// PlaceholderInputOutputInfo. FUN could be Function or IRFunction. ARR could
+/// be std::list<Placeholder *> or std::vector<const Placeholder *>
+template <typename FUN, typename ARR>
+ContiguousPlaceholders getContiguousPlaceHolder(const ARR &holders,
+                                                const FUN &F);
+
 } // end namespace glow
 #endif // GLOW_BACKENDS_BACKENDUTILS_H
--- a/lib/Backend/BackendUtils.cpp
+++ b/lib/Backend/BackendUtils.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "glow/Backend/BackendUtils.h"
+#include "glow/IR/IRUtils.h"
 #include "glow/IR/Instrs.h"
 
 #include "llvm/Support/CommandLine.h"
@@ -84,14 +85,17 @@ runtime::RuntimeBundle::getSymbolInfo(const Named *v) const {
   return it->second;
 }
 
-/// If \p PH is an output placeholder, \returns true.
-/// This is determined by checking if the PH has a user which uses the PH as an
-/// overwritten input.
-bool isOutput(const Placeholder *PH) {
+namespace glow {
+
+bool isOutput(const Placeholder *PH, const Function &F) {
   for (const auto &use : PH->getUsers()) {
     // Look through the inputs of the PH's users. If an input is overwritten
     // check if it's the PH, if it is return true.
     auto *user = use.getUser();
+    // Consider only users inside the same function.
+    if (user->getParent() != &F) {
+      continue;
+    }
     for (unsigned i = 0, numInputs = user->getNumInputs(); i < numInputs; i++) {
       // If the input is not overwritten we can continue.
       if (!user->isOverwrittenNthInput(i)) {
@@ -106,10 +110,13 @@ bool isOutput(const Placeholder *PH) {
   return false;
 }
 
-/// If \p PH is an input placeholder, \returns true.
-bool isInput(const Placeholder *PH) {
+bool isInput(const Placeholder *PH, const Function &F) {
   // Check that the PH is the input to a saveNode or is used by a non saveNode.
   for (const auto &use : PH->getUsers()) {
+    // Consider only users inside the same function.
+    if (use.getUser()->getParent() != &F) {
+      continue;
+    }
     // Check if PH is an input to a saveNode.
     if (auto *save = dyn_cast<SaveNode>(use.getUser())) {
       auto input = save->getInput();
@@ -122,6 +129,109 @@ bool isInput(const Placeholder *PH) {
   }
   return false;
 }
+
+bool isOutput(const Placeholder *PH, const IRFunction &F) {
+  auto *weight = F.getWeightForNode(PH);
+  assert(weight && "Weight for a node was not found");
+  for (const auto &use : ValueUses(weight)) {
+    Instruction *user = use.get();
+    // Ignore deallocs.
+    if (isa<DeallocActivationInst>(user)) {
+      continue;
+    }
+    OperandKind kind = use.getOperand().second;
+    if (kind == OperandKind::Out || kind == OperandKind::InOut) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool isInput(const Placeholder *PH, const IRFunction &F) {
+  // Check that the PH is always used as an @in parameter by the current
+  // function.
+  auto *weight = F.getWeightForNode(PH);
+  assert(weight && "Weight for a node was not found");
+  for (const auto &use : ValueUses(weight)) {
+    Instruction *user = use.get();
+    // Ignore deallocs.
+    if (isa<DeallocActivationInst>(user)) {
+      continue;
+    }
+    OperandKind kind = use.getOperand().second;
+    if (kind == OperandKind::In || kind == OperandKind::InOut) {
+      return true;
+    }
+  }
+  return false;
+}
+
+template <typename FUN, typename ARR>
+ContiguousPlaceholders getContiguousPlaceHolder(const ARR &holders,
+                                                const FUN &F) {
+  // Pure input placeholders.
+  std::vector<const Placeholder *> intputPlaceholders;
+  // Pure output placeholders.
+  std::vector<const Placeholder *> outputPlaceholders;
+  // Input&output placeholders.
+  std::vector<const Placeholder *> inputOutputPlaceholders;
+  // Neither input nor output placeholders.
+  std::vector<const Placeholder *> emptyPlaceholders;
+  // Return value.
+  ContiguousPlaceholders ret;
+
+  for (auto &v : holders) {
+    if (isInput(v, F)) {
+      if (!isOutput(v, F)) {
+        intputPlaceholders.push_back(v);
+      } else {
+        inputOutputPlaceholders.push_back(v);
+      }
+    } else {
+      if (isOutput(v, F)) {
+        outputPlaceholders.push_back(v);
+      } else {
+        emptyPlaceholders.push_back(v);
+      }
+    }
+  }
+
+  for (auto &v : intputPlaceholders) {
+    PlaceholderInputOutputInfo holder;
+    holder.addr = v;
+    holder.isInput = true;
+    holder.isOutput = false;
+    ret.push_back(holder);
+  }
+
+  for (auto &v : inputOutputPlaceholders) {
+    PlaceholderInputOutputInfo holder;
+    holder.addr = v;
+    holder.isInput = true;
+    holder.isOutput = true;
+    ret.push_back(holder);
+  }
+
+  for (auto &v : outputPlaceholders) {
+    PlaceholderInputOutputInfo holder;
+    holder.addr = v;
+    holder.isInput = false;
+    holder.isOutput = true;
+    ret.push_back(holder);
+  }
+
+  for (auto &v : emptyPlaceholders) {
+    PlaceholderInputOutputInfo holder;
+    holder.addr = v;
+    holder.isInput = false;
+    holder.isOutput = false;
+    ret.push_back(holder);
+  }
+
+  return ret;
+}
+
+} // namespace glow
 
 runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
   std::unordered_map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
@@ -144,15 +254,22 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
   }
 
   // Allocate placeholders.
-  for (auto const *V : F.findPlaceholders()) {
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  auto contiguousPlaceholders =
+      getContiguousPlaceHolder(F.findPlaceholders(), F);
+
+  // Compute the offsets for Placeholders.
+  for (auto it = contiguousPlaceholders.begin();
+       it != contiguousPlaceholders.end(); it++) {
+    auto &V = it->addr;
     auto size = V->getType()->getSizeInBytes();
     auto offset = placeholders.allocate(size, V);
     runtime::RuntimeSymbolInfo symbol;
     symbol.offset = offset;
     symbol.size = size;
     symbol.type = *V->getType();
-    symbol.output = isOutput(V);
-    symbol.input = isInput(V);
+    symbol.output = it->isOutput;
+    symbol.input = it->isInput;
     symbol.symbolCategory = SymbolCategory::Placeholder;
     symbolTable.emplace(V->getName(), symbol);
   }
@@ -187,9 +304,14 @@ runtime::RuntimeBundle::create(const IRFunction &F,
   }
   auto constantMaxSize = constantAllocator.getMaxMemoryUsage();
 
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  auto contiguousPlaceholders =
+      getContiguousPlaceHolder(F.findPlaceholders(), F);
+
   // Compute the offsets for Placeholders.
-  for (auto &v : F.findPlaceholders()) {
-    // Get the WeightVar for each Placeholder to calculate offsets.
+  for (auto it = contiguousPlaceholders.begin();
+       it != contiguousPlaceholders.end(); it++) {
+    auto &v = it->addr;
     assert(isa<WeightVar>(F.getWeightForNode(v)) && "Expected WeightVar");
     auto *w = cast<WeightVar>(F.getWeightForNode(v));
     auto numBytes = w->getSizeInBytes();
@@ -198,15 +320,14 @@ runtime::RuntimeBundle::create(const IRFunction &F,
     symbol.offset = addr;
     symbol.size = numBytes;
     symbol.type = *w->getType();
-    symbol.output = isOutput(v);
-    symbol.input = isInput(v);
+    symbol.output = it->isOutput;
+    symbol.input = it->isInput;
     symbol.symbolCategory = SymbolCategory::Placeholder;
     symbolTable.emplace(std::string(v->getName()), symbol);
   }
   auto placeholderMaxSize = placeholderAllocator.getMaxMemoryUsage();
 
   // Compute the offsets for Activations.
-
   for (const auto &I : F.getInstrs()) {
     if (auto *A = dyn_cast<AllocActivationInst>(&I)) {
       auto numBytes = I.getSizeInBytes();
--- a/lib/LLVMIRCodeGen/AllocationsInfo.cpp
+++ b/lib/LLVMIRCodeGen/AllocationsInfo.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "glow/LLVMIRCodeGen/AllocationsInfo.h"
+#include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
 #include "glow/CodeGen/MemoryAllocator.h"
 #include "glow/Graph/Graph.h"
@@ -51,8 +52,15 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F) {
     allocatedAddress_[w] = addr;
   }
 
+  // Placeholders should be allocated in a order of
+  // intput|inputOutput|output|neither.
+  auto contiguousPlaceholders =
+      getContiguousPlaceHolder(F->findPlaceholders(), *F);
+
   // Compute the offsets and total memory requirements for Placeholders.
-  for (auto &v : F->findPlaceholders()) {
+  for (auto it = contiguousPlaceholders.begin();
+       it != contiguousPlaceholders.end(); it++) {
+    auto &v = it->addr;
     // Get the WeightVar for each Placeholder to calculate offsets.
     assert(isa<WeightVar>(F->getWeightForNode(v)) && "Expected WeightVar");
     auto *w = cast<WeightVar>(F->getWeightForNode(v));
--- a/tests/unittests/BackendTest.cpp
+++ b/tests/unittests/BackendTest.cpp
@@ -27,6 +27,14 @@
 
 using namespace glow;
 
+/// An enum to indicate what type placholder it is.
+enum class PlaceholderType {
+  InputPlaceholder = 0,
+  InputOutputPlaceholder = 1,
+  OutputPlaceholder = 2,
+  NonePlaceholder = 3
+};
+
 class BackendTest : public ::testing::TestWithParam<BackendKind> {
 public:
   ExecutionEngine EE_{GetParam()};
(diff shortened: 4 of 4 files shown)


lib/LLVMIRCodeGen/AllocationsInfo.cpp
```diff
@@ -35,6 +35,45 @@ using llvm::cast;
...
 using llvm::isa;
 
+/// If \p PH is an output placeholder, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutputPlaceholder(const Placeholder *PH) {
+  for (const auto &use : PH->getUsers()) {
+    // Look through the inputs of the PH's users. If an input is overwritten
+    // check if it's the PH, if it is return true.
+    auto *user = use.getUser();
```
opti-mix: Your function operates at the graph level, but you use it from a function that operates at the `IRFunction` level. This is wrong.

Also, you only want to consider uses in the current function, not in the whole module.

lib/LLVMIRCodeGen/AllocationsInfo.cpp
```diff
@@ -35,6 +35,45 @@ using llvm::cast;
...
+      if (input.getNode() == PH) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+/// If \p PH is an input placeholder, \returns true.
+bool isInputPlaceholder(const Placeholder *PH) {
```
opti-mix: Same issue here.

lib/LLVMIRCodeGen/AllocationsInfo.cpp
```diff
@@ -35,6 +35,45 @@ using llvm::cast;
...
+      if (input.getNode() != PH) {
+        continue;
+      }
+    }
+    return true;
+  }
+  return false;
+}
+
 void AllocationsInfo::allocateWeightVars(const IRFunction *F) {
```
opti-mix: The fact that we still have `AllocationsInfo` which performs its own allocation worries me. It is a recipe for disaster. I thought we switched completely to `RuntimeBundle` and centralized the allocation logic there. But I see that our CPU backend currently invokes `allocateJITMemory` and `runtime::RuntimeBundle::create` from `LLVMBackend::compileIRWithoutConstants`! This seems very odd. 

@gcatron What was the reason for not switching completely to `RuntimeBundle` or at least implementing `AllocationsInfo` on top of `RuntimeBundle` in the CPU backend? Why does each of them has its own allocation logic?

tests/unittests/BackendTest.cpp
```diff
@@ -165,6 +165,77 @@ TEST(RuntimeBundle, BundleSymbolInfo) {
...
+  auto s = conv->getResult().getType()->size();
+  auto *fcFilter =
+      MD.createPlaceholder(ElemKind::Int8QTy, {s, 6}, 0.4, 2, "F", true);
+  auto *fcBias =
+      MD.createPlaceholder(ElemKind::Int32QTy, {6}, 0.4, 2, "B", true);
+  Node *O = F->createFullyConnected("fc1", conv, fcFilter, fcBias);
+  PlaceholderBindings bindings;
+  F->createSave("ret", O);
+  EE.compile(CompilationMode::Infer, F);
+  auto table = EE.getCompiledFunction().getRuntimeBundle().getSymbolTable();
```
opti-mix: It should be `auto &table`.

tests/unittests/BackendTest.cpp
```diff
@@ -165,6 +165,77 @@ TEST(RuntimeBundle, BundleSymbolInfo) {
...
 }
 
+// Test if the placeholders are allocated contiguously as
+// Input|InputOutput|Output.
+TEST(RuntimeBundle, ContiguousPlaceholder) {
+  ExecutionEngine EE;
+  auto &MD = EE.getModule();
+  auto *F = MD.createFunction("main");
+
+  unsigned depth = 16;
```
opti-mix: There is no need in a complex use-case involving convolutions and FCs for this PR. It just adds complexity. Use something simpler. Any operations (e.g. add, sub, etc) would do.

tests/unittests/BackendTest.cpp
```diff
@@ -165,6 +165,77 @@ TEST(RuntimeBundle, BundleSymbolInfo) {
...
+  }
+  // Sort the placeholders by offset.
+  sort(tableContainer.begin(), tableContainer.end(),
+       [](const glow::runtime::RuntimeSymbolInfo &a,
+          const glow::runtime::RuntimeSymbolInfo &b) {
+         return (a.offset < b.offset);
+       });
+  // Check if the placeholders are allocated in the contiguous order
+  // Input|Inputoutput|Output.
+  int prev = -2;
```
opti-mix: This is a test and the logic below is not very obvious. Please try to simplify it to make it obvious rather than fast or efficient.

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +187,29 @@ runtime::RuntimeBundle::create(const IRFunction &F,
...
   auto constantMaxSize = constantAllocator.getMaxMemoryUsage();
 
-  // Compute the offsets for Placeholders.
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  std::vector<Placeholder const *> inputOutputPlaceholders;
+  std::vector<Placeholder const *> outputPlaceholders;
+  std::vector<Placeholder const *> contiguousPlaceholders;
   for (auto &v : F.findPlaceholders()) {
-    // Get the WeightVar for each Placeholder to calculate offsets.
+    if (isInput(v)) {
```
opti-mix: I think that `isInput` and `isOutput` checks are not quite correct currently. They should consider only uses in the current function rather than in the whole module.

lib/LLVMIRCodeGen/AllocationsInfo.cpp
opti-mix: I filed #3011 to track this issue.

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +187,29 @@ runtime::RuntimeBundle::create(const IRFunction &F,
...
+      if (!isOutput(v)) {
+        contiguousPlaceholders.push_back(v);
+      } else {
+        inputOutputPlaceholders.push_back(v);
+      }
+    } else {
+      outputPlaceholders.push_back(v);
+    }
+  }
+  for (auto &v : inputOutputPlaceholders) {
```
nickgg: Rather than creating multiple vectors, it may be cleaner to move the body of the loop below (that does the allocation & insertion into the symbol table) into a helper function and then loop through your inputs, inputsOutputs and outputs containers calling that helper.

lib/Backend/BackendUtils.cpp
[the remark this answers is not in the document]
nickgg: Agree, I'd probably just have bools for the flags locally to this function.

nickgg: This could be simpler if you do the inputOutput check first:

```
if (input && output) {
  inputOutputPlacholders.push_back(v);
} else if (input) {
  contiguousPlaceholders.push_back(v);
} else { 
  outputPlaceholders.push_back(v);
}
```

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +187,29 @@ runtime::RuntimeBundle::create(const IRFunction &F,
...
   for (auto &v : F.findPlaceholders()) {
-    // Get the WeightVar for each Placeholder to calculate offsets.
+    if (isInput(v)) {
+      if (!isOutput(v)) {
+        contiguousPlaceholders.push_back(v);
+      } else {
+        inputOutputPlaceholders.push_back(v);
+      }
+    } else {
+      outputPlaceholders.push_back(v);
```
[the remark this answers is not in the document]
nickgg: What are you going to do for the case where it's neither an input or an output? We should at least log this.

lib/LLVMIRCodeGen/AllocationsInfo.cpp
```diff
@@ -51,8 +90,29 @@ void AllocationsInfo::allocateWeightVars(const IRFunction *F) {
     allocatedAddress_[w] = addr;
   }
 
-  // Compute the offsets and total memory requirements for Placeholders.
+  // Placeholders order should match the allocation order in BackendUtils.
+  std::vector<Placeholder const *> inputOutputPlaceholders;
+  std::vector<Placeholder const *> outputPlaceholders;
+  std::vector<Placeholder const *> contiguousPlaceholders;
   for (auto &v : F->findPlaceholders()) {
+    if (isInputPlaceholder(v)) {
```
nickgg: Isn't this a copy of the code above? Can we commonalize with a helper function instead?

lib/Backend/BackendUtils.cpp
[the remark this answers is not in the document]
superwizard2019: > It is probably better to call `isInput` and `isOutput` unconditionally for each placeholder and then conditionally add them to different sets based on the obtained flags.

OK. I've moved the existing `isInput` and` isOutput` into namespace glow as a helper function. And I created a temp tuple to store their value to make sure they are only called once.

lib/Backend/BackendUtils.cpp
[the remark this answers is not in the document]
superwizard2019: > It should check or assert that v is an output.

OK. previously, I thought there should be no placeholder that is neither an output nor an input. But it looks like I can not pass some existing test without allocating those. So currently I allocate them all. The order is input|inputoutput|output|neither.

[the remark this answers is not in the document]
superwizard2019: > What are you going to do for the case where it's neither an input or an output? We should at least log this.

yea. I should have allocated them all. Misunderstood the logic.

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +187,29 @@ runtime::RuntimeBundle::create(const IRFunction &F,
   }
   auto constantMaxSize = constantAllocator.getMaxMemoryUsage();
 
-  // Compute the offsets for Placeholders.
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  std::vector<Placeholder const *> inputOutputPlaceholders;
```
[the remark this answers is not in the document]
superwizard2019: > The three vectors do not correspond exactly to the comment above. Please update the comment to reflect the actual state of affairs here.

OK. I merged them into one helper function.

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +187,29 @@ runtime::RuntimeBundle::create(const IRFunction &F,
   }
   auto constantMaxSize = constantAllocator.getMaxMemoryUsage();
 
-  // Compute the offsets for Placeholders.
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
```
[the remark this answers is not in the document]
superwizard2019: > Do we need to add a similar logic to `runtime::RuntimeBundle::create(const Function &F)` in this file?

OK. I already added that.

lib/LLVMIRCodeGen/AllocationsInfo.cpp
superwizard2019: > Your function operates at the graph level, but you use it from a function that operates at the `IRFunction` level. This is wrong.
> 
> Also, you only want to consider uses in the current function, not in the whole module.

This was just a copy of existing `isInput` and `isOutput` in the `BackendUtils.cpp`. (I already merged them into a helper function currently) I noticed that these two functions are called during the existing allocations and saved inside the `runtime::RuntimeSymbolInfo`.  I was not sure if there will be an issue if I change the logic here. Do you have any suggestions about the logic of the implementations ? :)

tests/unittests/BackendTest.cpp
superwizard2019: > It should be `auto &table`.

OK. Already corrected it. Thanks :)

tests/unittests/BackendTest.cpp
superwizard2019: > There is no need in a complex use-case involving convolutions and FCs for this PR. It just adds complexity. Use something simpler. Any operations (e.g. add, sub, etc) would do.

OK. I just simplified the network.

lib/Backend/BackendUtils.cpp
superwizard2019: > Rather than creating multiple vectors, it may be cleaner to move the body of the loop below (that does the allocation & insertion into the symbol table) into a helper function and then loop through your inputs, inputsOutputs and outputs containers calling that helper.

OK. Moved them all into a helper function.

lib/Backend/BackendUtils.cpp
superwizard2019: > This could be simpler if you do the inputOutput check first:
> 
> ```
> if (input && output) {
>   inputOutputPlacholders.push_back(v);
> } else if (input) {
>   contiguousPlaceholders.push_back(v);
> } else { 
>   outputPlaceholders.push_back(v);
> }
> ```

Sure

lib/LLVMIRCodeGen/AllocationsInfo.cpp
superwizard2019: > Isn't this a copy of the code above? Can we commonalize with a helper function instead?

OK.

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,15 @@ class RuntimeBundle {
 };
 } // namespace runtime
 
+bool isOutput(const Placeholder *PH);
+
+bool isInput(const Placeholder *PH);
+
+std::vector<std::tuple<Placeholder const *, bool, bool>>
```
rdzhabarov: nit: typically it's const X * (and in other places)

rdzhabarov: Add doxygen comment and explain what those methods do.

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +125,108 @@ bool isInput(const Placeholder *PH) {
   return false;
 }
 
+/// Convert placeholders to be ordered as input|inputOutput|output|neither.
```
rdzhabarov: this needs to go to header file.

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +125,108 @@ bool isInput(const Placeholder *PH) {
...
+/// Packed into {Placeholder *, isInput, isOutput}.
+std::vector<std::tuple<Placeholder const *, bool, bool>>
+getContiguousPlaceHolder(std::list<Placeholder *> const &holders) {
+  // Pure input placeholders.
+  std::vector<Placeholder const *> intputPlaceholders;
+  // Pure output placeholders.
+  std::vector<Placeholder const *> outputPlaceholders;
+  // Input&output placeholders.
+  std::vector<Placeholder const *> inputOutputPlaceholders;
+  // Neither input nor output placeholders.
```
rdzhabarov: what kinda of placeholder is it? (it should not be possible)

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +125,108 @@ bool isInput(const Placeholder *PH) {
...
+  }
+
+  for (auto &v : inputOutputPlaceholders) {
+    ret.push_back(std::make_tuple(v, true, true));
+  }
+
+  for (auto &v : outputPlaceholders) {
+    ret.push_back(std::make_tuple(v, false, true));
+  }
+  // Currently, we have to allocate this to pass BackendCorrectnessTest.
```
rdzhabarov: > BackendCorrectnessTest

do you understand why test fails? (is the test correct?)

superwizard2019: > > BackendCorrectnessTest
> 
> do you understand why test fails? (is the test correct?)

It falls in CPU/CPUOnly.dataParallelStackingTest and report a segmentation fault:11 error. Haven't figured out why currently. 
Screen Shot 2019-06-07 at 1 50 04 PM

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,23 @@ class RuntimeBundle {
...
+/// overwritten input.
+bool isOutput(const Placeholder *PH);
+
+/// If \p PH is an input placeholder, \returns true.
+bool isInput(const Placeholder *PH);
+
+/// Convert placeholders to be ordered as input|inputOutput|output|neither.
+/// Packed into {Placeholder *, isInput, isOutput}.
+std::vector<std::tuple<const Placeholder *, bool, bool>>
+getContiguousPlaceHolder(std::vector<const Placeholder *> const &holders);
```
opti-mix: Try to have only one implementation of these two functions. Either call one from the other or may be make this function a template-function?

tests/unittests/BackendTest.cpp
```diff
@@ -165,6 +165,82 @@ TEST(RuntimeBundle, BundleSymbolInfo) {
...
+          const glow::runtime::RuntimeSymbolInfo &b) {
+         return (a.offset < b.offset);
+       });
+
+  // Define the order of placeholders.
+  auto order = [](glow::runtime::RuntimeSymbolInfo i) {
+    if (i.input) {
+      if (!i.output) {
+        // input only
+        return 0;
```
opti-mix: May be define an enum for these four cases? It would look cleaner. And add an explicit return type to the capture.

include/glow/Backend/BackendUtils.h
superwizard2019: > Try to have only one implementation of these two functions. Either call one from the other or may be make this function a template-function?

OK. I merged them into a template-function

tests/unittests/BackendTest.cpp
superwizard2019: > May be define an enum for these four cases? It would look cleaner. And add an explicit return type to the capture.

OK. I've defined a new enum PlaceholderType to make it clear

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,36 @@ class RuntimeBundle {
...
+
+/// Convert placeholders to be ordered as input|inputOutput|output|neither.
+/// Packed into {Placeholder *, isInput, isOutput}.
+template <typename FUN>
+std::vector<std::tuple<const Placeholder *, bool, bool>>
+getContiguousPlaceHolder(std::vector<const Placeholder *> const &holders,
+                         const FUN &F);
+
+/// An enum to indicate what type placholder it is.
+enum class PlaceholderType {
```
opti-mix: This enum is used only by tests, right? If this is true, just move it into the CPP file with tests.

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +131,96 @@ bool isInput(const Placeholder *PH) {
...
+    if (kind == OperandKind::In || kind == OperandKind::InOut) {
+      return true;
+    }
+  }
+  return false;
+}
+
+template <typename FUN>
+std::vector<std::tuple<const Placeholder *, bool, bool>>
+getContiguousPlaceHolder(std::vector<const Placeholder *> const &holders,
```
opti-mix: I'd suggest to make the type of holders a template type as well. Then you don't need to convert a list into vector in one of the callers.

lib/Backend/BackendUtils.cpp
```diff
@@ -144,15 +242,27 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
   }
 
   // Allocate placeholders.
-  for (auto const *V : F.findPlaceholders()) {
+  // Convert list of placeholders to vector
+  std::list<Placeholder *> listPlaceholders = F.findPlaceholders();
+  std::vector<const Placeholder *> vectorPlaceholders{
```
opti-mix: Per my comment above, you would not need to do it.

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +131,96 @@ bool isInput(const Placeholder *PH) {
...
+      } else {
+        inputOutputPlaceholders.push_back(v);
+      }
+    } else {
+      if (isOutput(v, F)) {
+        outputPlaceholders.push_back(v);
+      } else {
+        // Assert this empty may fail some test. Currently, we bypass the
+        // allocations. Need to log them somewhere else.
+        emptyPlaceholders.push_back(v);
```
opti-mix: It would be nice to understand what kind of placeholders are in this category. If they are not read or written, are they used at all in the current function?

May be it is fixed now, after you fixed isInput/isOutput functions?

include/glow/Backend/BackendUtils.h
superwizard2019: > This enum is used only by tests, right? If this is true, just move it into the CPP file with tests.

OK. I moved them to the test cpp file.

lib/Backend/BackendUtils.cpp
superwizard2019: > I'd suggest to make the type of holders a template type as well. Then you don't need to convert a list into vector in one of the callers.

Sure

lib/Backend/BackendUtils.cpp
superwizard2019: > It would be nice to understand what kind of placeholders are in this category. If they are not read or written, are they used at all in the current function?
> 
> May be it is fixed now, after you fixed isInput/isOutput functions?

If we remove the allocation of the emptyPlaceholders, it still pass the tests. But assert it being empty will not.  I tried to look into one of the failed test, 
in the BackendCorrectnessTest.convGradTest, I dumped out the emptyPlaceholders. It has only one nodes named var2  which is connected to the selected of SoftMax and SoftMaxGrad function. Below is the DAG of the function after differentiation.
conv1_F_after

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,27 @@ class RuntimeBundle {
...
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH, const IRFunction &F);
+
+/// If \p PH is an input placeholder in the function \p F, \returns true.
+bool isInput(const Placeholder *PH, const IRFunction &F);
+
+/// Convert placeholders to be ordered as input|inputOutput|output|neither.
+/// Packed into {Placeholder *, isInput, isOutput}.
+template <typename FUN, typename ARR>
```
opti-mix: Please add comments about `FUN` and `APR` parameters

opti-mix: BTW, how does it work for a call from `AllocationsInfo.cpp`? This header does not provide the implementation of this template function. How does `AllocationsInfo.cpp` manage to call it and find a proper implementation?

lib/Backend/BackendUtils.cpp
opti-mix: OK. Let's leave it for now then.

lib/Backend/BackendUtils.cpp
```diff
@@ -144,15 +241,24 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
   }
 
   // Allocate placeholders.
-  for (auto const *V : F.findPlaceholders()) {
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  std::vector<std::tuple<const Placeholder *, bool, bool>>
+      contiguousPlaceholders =
+          glow::getContiguousPlaceHolder<Function, std::list<Placeholder *>>(
```
opti-mix: Is it necessary to explicitly provide the type parameters for the template function? Is it because the compiler cannot deduce them automatically?

lib/Backend/BackendUtils.cpp
```diff
@@ -187,9 +293,17 @@ runtime::RuntimeBundle::create(const IRFunction &F,
   }
   auto constantMaxSize = constantAllocator.getMaxMemoryUsage();
 
+  // Placeholders should be allocated in a order of Input|InputOutput|Output.
+  std::vector<std::tuple<const Placeholder *, bool, bool>>
+      contiguousPlaceholders =
+          glow::getContiguousPlaceHolder<IRFunction,
```
opti-mix: Is it necessary to explicitly provide the type parameters for the template function?

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,27 @@ class RuntimeBundle {
...
+std::vector<std::tuple<const Placeholder *, bool, bool>>
```
opti-mix: Nit: I wonder of using a small struct with 3 fields as a helper would be a cleaner and more understandable solution?

superwizard2019: > Nit: I wonder of using a small struct with 3 fields as a helper would be a cleaner and more understandable solution?

OK. I grouped them into a struct with 3 fields.

lib/Backend/BackendUtils.cpp
superwizard2019: > Is it necessary to explicitly provide the type parameters for the template function?

I was thinking about making it more clear because of the std::tuple used. Since we switch to a struct, I just replace it with `auto`

include/glow/Backend/BackendUtils.h
superwizard2019: > BTW, how does it work for a call from `AllocationsInfo.cpp`? This header does not provide the implementation of this template function. How does `AllocationsInfo.cpp` manage to call it and find a proper implementation?

I tried to comment the declaration in the BackendUtils.h its a compile error and If I comment the implementation in the BackendUtils.cpp its a linking error. I searched 'BackendUtils.cpp' and it is not included in anywhere else. Still trying to figure out why it works.

include/glow/Backend/BackendUtils.h
superwizard2019: > Add doxygen comment and explain what those methods do.

OK sure

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,41 @@ class RuntimeBundle {
...
+/// If \p PH is an output placeholder in the function \p F, \returns true.
+/// This is determined by checking if the PH has a user which uses the PH as an
+/// overwritten input.
+bool isOutput(const Placeholder *PH, const IRFunction &F);
+
+/// If \p PH is an input placeholder in the function \p F, \returns true.
+bool isInput(const Placeholder *PH, const IRFunction &F);
+
+/// Contains information for placeholder during allocation.
+struct PlaceholderInfo {
```
opti-mix: The structure is OK. But its name is a bit too general. Let's call it something more specific, e.g. `PlaceholderInputOutputInfo` or something like that.

lib/Backend/BackendUtils.cpp
```diff
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 #include "glow/Backend/BackendUtils.h"
+#include "glow/IR/IRUtils.h"
 #include "glow/IR/Instrs.h"
 
 #include "llvm/Support/CommandLine.h"
+#include <tuple>
```
opti-mix: Not needed anymore?

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +131,109 @@ bool isInput(const Placeholder *PH) {
   return false;
 }
 
+bool isOutput(const Placeholder *PH, const IRFunction &F) {
```
opti-mix: Please add a Doxygen comment.

lib/Backend/BackendUtils.cpp
```diff
@@ -123,6 +131,109 @@ bool isInput(const Placeholder *PH) {
...
+    }
+    OperandKind kind = use.getOperand().second;
+    if (kind == OperandKind::Out || kind == OperandKind::InOut) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool isInput(const Placeholder *PH, const IRFunction &F) {
```
opti-mix: Please add a Doxygen comment.

lib/Backend/BackendUtils.cpp
opti-mix: I meant do we need to use `glow::getContiguousPlaceHolder<IRFunction, ...>(...)` vs just `glow::getContiguousPlaceHolder(...)` without specifying the types of template parameters? Also, why do you need to explicitly use the `glow` namespace prefix?

Your idea about using `auto` for `contiguousPlaceholders` also makes sense.

lib/Backend/BackendUtils.cpp
superwizard2019: > Please add a Doxygen comment.

I moved them to the `BackendUtils.h` . Should I add another one in the cpp file?

include/glow/Backend/BackendUtils.h
```diff
@@ -120,5 +120,41 @@ class RuntimeBundle {
 };
 } // namespace runtime
 
+/// If \p PH is an output placeholder, \returns true.
```
opti-mix: Describe the parameter F like you do for other functions. 

`\returns` should be usually on a new line, IIRC.

Update other Doxygen comments accordingly.

lib/Backend/BackendUtils.cpp
opti-mix: Oh, I missed it. No, it is fine then.

include/glow/Backend/BackendUtils.h
superwizard2019: > The structure is OK. But its name is a bit too general. Let's call it something more specific, e.g. `PlaceholderInputOutputInfo` or something like that.

SUre

lib/Backend/BackendUtils.cpp
superwizard2019: > I meant do we need to use `glow::getContiguousPlaceHolder<IRFunction, ...>(...)` vs just `glow::getContiguousPlaceHolder(...)` without specifying the types of template parameters? Also, why do you need to explicitly use the `glow` namespace prefix?
> 
> Your idea about using `auto` for `contiguousPlaceholders` also makes sense.

I just deleted the template parameters.

include/glow/Backend/BackendUtils.h
superwizard2019: > Describe the parameter F like you do for other functions.
> 
> `\returns` should be usually on a new line, IIRC.
> 
> Update other Doxygen comments accordingly.

OK. I made the updates and put the `\returns`  into new line.

superwizard2019:
> LGTM! Thanks for fixing this issue!
> 
> Before merging via Phabricator, please test locally with the CPU backend on big networks like ResNet50, etc to see that nothing is broken.

OK, Thank you! I tried the examples like mnist and cifar10 and it works fine

PR #3003, opened 2019-05-30
[Partitioner] Enable more than one partitions assigned to the same device

beicy:
Summary:
The PR allows more than one partitions to be assigned into one device.
This happens when the number of generated DAGNodes is larger than the number of devices for any backendKind . E.g:                                
node1(6GB) -> node2(14GB) -> node3(6GB). The memory limitation is 16GB, and there is only 2 devices. So we assign node1 and node3 into one device.  
Documentation:

Test Plan:
added unnitest. 

#2298 

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -37,6 +37,8 @@ struct BackendInfo {
   size_t num = 0;
   /// The memory constraints for this backend kind.
   uint64_t memSize;
+  /// Backend Kind name.
+  std::string name;
   /// Backend pointer.
   Backend *backend = nullptr;
 };
@@ -69,10 +71,16 @@ class NodeToFunctionMap {
   /// Map of nodes in the original function to their target partition.
   NodeToFunctionMapTy nodeToFunction_;
 
+  /// Map of the partitions to the backendKind which will be used for compiling
+  /// this partition.
   FunctionToBackendKindMapTy functionToBackendKind_;
 
   /// Map of sub-functions to their memory consumption.
   PartitionCostMapTy partitionCost_;
+
+  /// Map of partitions and the logicalDeviceID. The partitions with the same
+  /// logcialDeviceID will be assigned into the same physical device.
+  std::map<Function *, std::vector<DeviceIDTy>> logicalDeviceIDMap_;
 
 public:
   /// Create a new partition \p F, and map it with \p backendKind.
@@ -93,6 +101,19 @@ public:
   /// Get list of functions contained in this map.
   const FunctionList &getPartitions() const { return functions_; }
 
+  /// Get the list of logical device ID related to this function \p F.
+  std::vector<DeviceIDTy> getLogicalDeviceIDList(Function *F) {
+    return logicalDeviceIDMap_[F];
+  }
+
+  void appendLogicalDeviceID(Function *F, DeviceIDTy id) {
+    if (logicalDeviceIDMap_.find(F) == logicalDeviceIDMap_.end()) {
+      logicalDeviceIDMap_.emplace(
+          std::make_pair(F, std::vector<DeviceIDTy>{id}));
+    } else {
+      logicalDeviceIDMap_[F].push_back(id);
+    }
+  }
   /// attach \p map to current mapping.
   void insert(NodeToFunctionMap &map) {
     FunctionList flist = map.getPartitions();
@@ -100,6 +121,8 @@ public:
       Function *func = *it;
       auto backendKind = map.getPartitionBackendKind(func);
       createPartition(func, backendKind);
+      GraphMemInfo cost = map.getGraphMemInfo(func);
+      setGraphMemInfo(func, cost);
     }
     for (auto it = map.begin(); it != map.end(); ++it) {
       Node *n = it->first;
@@ -143,9 +166,16 @@ class Partitioner {
   /// The backend pointers.
   std::vector<Backend *> backends_;
 
-  /// The map between BackendKind and its name. It is used for generating debug
-  /// info.
-  std::map<BackendKind, std::string> backendName_;
+  /// The map between BackendKind and BackendInfo.
+  std::map<BackendKind, BackendInfo> backendMap_;
+
+  /// The map between partitions and the logicalDeviceID. The partitions with
+  /// the same logcialDeviceID will be assigned into the same physical device.
+  std::map<Function *, std::vector<DeviceIDTy>> logicalIDMap_;
+
+  /// The number of logicalDevice IDs, i.e. the number of physical devices
+  /// needed after partitions.
+  DeviceIDTy logicalDeviceID_;
 
   /// The result of module partitioning.
   DAGListTy partitions_;
@@ -195,10 +225,15 @@ class Partitioner {
   NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory,
                                      BackendKind backendKind);
 
-  /// Adjust a logicalDevice ID to each DAGNode. It is possible that two
-  /// sub-functions need to be assigned into 1 device due to the memory
-  /// constraits.
-  void adjustLogicalDeviceID();
+  /// Assign a logicalDeviceID to each partition. It is possible that two
+  /// partitions need to be assigned into 1 device due to the number of physical
+  /// devices.
+  DeviceIDTy assignLogicalDeviceID(NodeToFunctionMap &partitions);
+
+  /// Check if this partition way \p partitions satisfies number of physical
+  /// devices restriction. I.e. check if the number of logical devices is less
+  /// than the given physical devices.
+  llvm::Error logicalDevicesValidation(NodeToFunctionMap &partitions);
 
   /// Duplicates all networks in the module order to saturate the Host.
   void saturateHost(unsigned logicalDeviceCount);
@@ -209,8 +244,8 @@ class Partitioner {
   /// Given the node-function mapping, do the actual partitioning. If \p saveDAG
   /// is true, the DAG will be saved into partitions_, which is the final
   /// partition result.
-  DeviceIDTy doPartitioning(llvm::StringRef funcName, std::vector<Function *>,
-                            NodeToFunctionMap &mapping, bool saveDAG);
+  void doPartitioning(llvm::StringRef funcName, std::vector<Function *>,
+                      NodeToFunctionMap &mapping, bool saveDAG);
 
 public:
   /// \p parent is the module which contains the functions need to be divided.
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -26,6 +26,13 @@
 using namespace glow;
 using llvm::isa;
 
+// Sorted the std::pair<DAGNode *, uint64_t> based on the second from min to
+// max.
+auto sortMinMemory(const std::pair<Function *, uint64_t> &a,
+                   const std::pair<Function *, uint64_t> &b) -> bool {
+  return a.second <= b.second;
+}
+
 /// Check if the memory of \p node inputs is calculated already. \returns the
 /// total used memory after \p node is considered.
 static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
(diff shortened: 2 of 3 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -26,6 +26,15 @@
 using namespace glow;
 using llvm::isa;
 
+namespace {
+// Sorted the std::pair<DAGNode *, uint64_t> based on the second from min to
+// max.
+auto sortMinMemory = [](const std::pair<DAGNode *, uint64_t> &a,
```
opti-mix: Why is it a closure? It can be just a usual function, or?

lib/Partitioner/Partitioner.cpp
```diff
@@ -62,12 +71,12 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
     for (size_t i = 0; i < node->children.size(); i++) {
       auto child = node->children[i];
       DescriptionBuilder db(child->name.c_str());
-      db.addParam("BackendKind", backendName_.at(child->backendKind));
+      std::string backendName = backendMap_.at(child->backendKind).name;
```
opti-mix: `const std::string &backendName`?

lib/Partitioner/Partitioner.cpp
```diff
@@ -492,7 +501,78 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
 /// the number of DAGNodes) is larger than the number of devices. E.g:
 /// node1(6GB) -> node2(14GB) -> node3(6GB). The memory limitation is 16GB, and
 /// there is only 2 devices.
-void Partitioner::adjustLogicalDeviceID() {}
+llvm::Error Partitioner::adjustLogicalDeviceID(NodeToFunctionMap &mapping) {
+  for (auto &dag : partitions_) {
+    // Traver the DAGNodes, and get list of with each backendKind.
```
opti-mix: Traverse the DAGNodes and get list of ??? with each backendKind.

List of what? It seems like this sentence is not finished or misses some words.

lib/Partitioner/Partitioner.cpp
```diff
@@ -492,7 +501,78 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+    std::map<BackendKind, std::vector<DAGNode *>> backendKindNum;
```
opti-mix: The name `backendKindNum` is a bit odd as you do not map the backendKind to a number. If I understand correctly, you map the backendKind to a set of DAGNodes and then you check the number of elements in this set.

lib/Partitioner/Partitioner.cpp
```diff
@@ -492,7 +501,78 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+        Function *function = module_->getFunction(p.second[i]->name);
+        uint64_t totalMem = mapping.getGraphMemInfo(function).getTotalMemSize();
+        nodeSize.push_back(std::make_pair(p.second[i], totalMem));
+      }
+      std::sort(nodeSize.begin(), nodeSize.end(), sortMinMemory);
+
+      // Assume we have n devices and m partitions, where m > n. If these m
+      // partitions can be assigned to n devices, there must be 1 device have at
+      // least (m - n + 1) partitions. Therefore, we just need to check the
+      // minial (m - n + 1) partitions can be put into 1 device while the device
```
opti-mix: minial -> minimal

lib/Partitioner/Partitioner.cpp
```diff
@@ -492,7 +501,78 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+      for (size_t i = 0, e = p.second.size(); i < e; i++) {
+        Function *function = module_->getFunction(p.second[i]->name);
+        uint64_t totalMem = mapping.getGraphMemInfo(function).getTotalMemSize();
+        nodeSize.push_back(std::make_pair(p.second[i], totalMem));
+      }
+      std::sort(nodeSize.begin(), nodeSize.end(), sortMinMemory);
+
+      // Assume we have n devices and m partitions, where m > n. If these m
+      // partitions can be assigned to n devices, there must be 1 device have at
+      // least (m - n + 1) partitions. Therefore, we just need to check the
```
opti-mix: > If these m partitions can be assigned to n devices, there must be 1 device have at least (m - n + 1) partitions. 

Is that true? E.g. if we have 10 partitions and 5 devices, you say that at least one device should get (10 - 5 + 1) = 6 partitions. Why? Why can't each device e.g. get two partitions?

tests/unittests/PartitionerTest.cpp
```diff
@@ -494,29 +497,66 @@ TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
...
+        // heterogeneous partiton.
+        ASSERT_EQ(node->logicalDevices.size(), 1);
+      }
+    }
+    mod_.clear();
+  }
+}
+
+/// Test assigning more than one partitions in to one device.
+TEST_F(PartitionerTest, HeterogeneousPartitioning1) {
```
opti-mix: Please give this test a more meaningful name.

opti-mix: You just added/modified heterogeneous partitioning tests in this PR, but this new feature could be used also with a single type of backends, right? You should test this case as well.

lib/Partitioner/Partitioner.cpp
beicy: Good catch!  Thanks for pointing this! I thought the putting n+1 items to n boxes issue... I will rewrite this algorithm.

lib/Partitioner/Partitioner.cpp
beicy: No special reason, just followed some code sample :P  Now removed.

beicy:
@opti-mix @gcatron  The updates in this version:
1. renamed `adjustLogicalDeviceID` with `assignLogicalDeviceID`;
2. moved `assignLogicalDeviceID` before "doPartitioning". Now it is the last step of optimization.
3. Added unittest for heterogeneous partition(logicalIDTest1) and single backendKind (logicalIDTest0).
Thanks!

tests/unittests/PartitionerTest.cpp
```diff
@@ -530,6 +533,87 @@ TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
...
+      ASSERT_EQ(node->logicalDevices.size(), 1);
+      usedID.insert(node->logicalDevices[0]);
+    }
+    // Check there are 2 logical devices.
+    ASSERT_EQ(usedID.size(), 2);
+  }
+  mod_.clear();
+}
+
+/// Test assigning more than one partitions in to one device.
```
gcatron: This test has the same comment as the one above, can you make the comments more specific. I'm a little confused what the difference is.

beicy: Done. One test for single backendKind, one for heterogeneous partitioning.

lib/Partitioner/Partitioner.cpp
```diff
@@ -26,6 +26,13 @@
 using namespace glow;
 using llvm::isa;
 
+// Sorted the std::pair<DAGNode *, uint64_t> based on the second from min to
+// max.
+auto sortMinMemory = [](const std::pair<Function *, uint64_t> &a,
```
opti-mix: This is just a simple function. No need to use a capture here.

include/glow/Partitioner/Partitioner.h
```diff
@@ -195,10 +225,15 @@ class Partitioner {
...
-  /// sub-functions need to be assigned into 1 device due to the memory
-  /// constraits.
-  void adjustLogicalDeviceID();
+  /// Assign a logicalDeviceID to each partition. It is possible that two
+  /// partitions need to be assigned into 1 device due to the number of physical
+  /// devices.
+  DeviceIDTy assignLogicalDeviceID(NodeToFunctionMap &partitions);
+
+  /// Check if this partition way \p partitions satisfies number of physical
+  /// devices restriction. I.e. checkif the number of logical devices is less
```
opti-mix: checkif -> check if

lib/Partitioner/Partitioner.cpp
```diff
@@ -62,12 +69,12 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
     for (size_t i = 0; i < node->children.size(); i++) {
       auto child = node->children[i];
       DescriptionBuilder db(child->name.c_str());
-      db.addParam("BackendKind", backendName_.at(child->backendKind));
+      const std::string &backendName = backendMap_.at(child->backendKind).name;
+      db.addParam("BackendKind", backendName);
       myfile << child->name << " [ label = \"" << escapeDottyString(db) << "\"";
```
opti-mix: Please use escapeDottyString on all names/dynamic strings. I've seen some issues if we don't do it.

lib/Partitioner/Partitioner.cpp
```diff
@@ -485,14 +520,124 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+    for (size_t i = 0, e = p.second.size(); i < e; i++) {
+      Function *function = p.second[i];
+      uint64_t totalMem = mapping.getGraphMemInfo(function).getTotalMemSize();
+      nodeSize.push_back(std::make_pair(p.second[i], totalMem));
+    }
+    std::sort(nodeSize.begin(), nodeSize.end(), sortMinMemory);
+
+    // Assume we have n devices(NOTE: here the n devices have the same avaiable
+    // memory size, and the following algorithm can find the accurate result. If
+    // the momery size are differnt, this assignment issue will be a NP problem
```
gcatron: momery ->memory

lib/Partitioner/Partitioner.cpp
```diff
@@ -485,14 +520,124 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+      size_t numOfPartitionsWithSameID = (m - 1) / n + 1;
+      size_t start = p.second.size() - m;
+      size_t i;
+      for (i = start; i < p.second.size(); i++) {
+        if (usedMem + nodeSize[i].second > backendMap_[p.first].memSize) {
+          break;
+        }
+        usedMem += nodeSize[i].second;
+      }
+      // Step 4 : if k found in step 3 is larger than (m - 1) / n + 1, this
```
gcatron: larger -> smaller,
Also it might be helpful to call out in the comment that k = i- start.

lib/Partitioner/Partitioner.cpp
```diff
@@ -485,14 +520,124 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+        // Can't find a proper assignment. Assign each partition a unique
+        // logicalDevice ID and return;
+        logicalDeviceID = 0;
+        for (auto &func : mapping.getPartitions()) {
+          mapping.appendLogicalDeviceID(func, logicalDeviceID++);
+        }
+        return logicalDeviceID;
+      }
+
+      // Step 5 : Assign these partitios which are assigned to one device with
```
gcatron: partitios -> partitions

lib/Partitioner/Partitioner.cpp
```diff
@@ -26,6 +26,13 @@
 using namespace glow;
 using llvm::isa;
 
+// Sorted the std::pair<DAGNode *, uint64_t> based on the second from min to
+// max.
+auto sortMinMemory(const std::pair<Function *, uint64_t> &a,
+                   const std::pair<Function *, uint64_t> &b) -> bool {
+  return a.second <= b.second;
+};
```
opti-mix: No need for a semicolon.

lib/Partitioner/Partitioner.cpp
```diff
@@ -84,7 +91,8 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
     auto *root = nodes[i];
     for (size_t j = 0; j < root->children.size(); j++) {
       auto child = root->children[j];
-      myfile << root->name << " -> " << child->name << ";";
+      myfile << escapeDottyString(root->name) << " -> "
```
opti-mix: You probably need to add `\""` before and after each of these `escapeDottyString` calls, so that it produces `"Escaped string"`.

lib/Partitioner/Partitioner.cpp
```diff
@@ -93,19 +101,46 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
   return;
 }
 
+llvm::Error
+Partitioner::logicalDevicesValidation(NodeToFunctionMap &partitions) {
```
opti-mix: Do you have a Doxygen comment for this function anywhere? If not, please add it.

lib/Partitioner/Partitioner.cpp
```diff
@@ -485,14 +520,125 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
   // Step 2 : adjust the partition based on performance.
   partitionsAdjust(mapping, availableMemory);
 
+  // Step 3 : assign each partition with a logical device id. The partitions
+  // with the same logical device id will be assigned into the same physical
+  // device.
+  // logicalDeviceID_ = assignLogicalDeviceID(mapping, logicalDeviceID_);
```
opti-mix: Why is this line commented out? Does it need to be removed?

beicy: Yes, these lines (line 535-526) need to be removed.

lib/Partitioner/Partitioner.cpp
beicy: Yes. In Partitioner.h.

lib/Partitioner/Partitioner.cpp
```diff
@@ -26,6 +26,13 @@
 using namespace glow;
 using llvm::isa;
 
+// Sorted the std::pair<DAGNode *, uint64_t> based on the second from min to
+// max.
+auto sortMinMemory(const std::pair<Function *, uint64_t> &a,
```
bertmaher: Is there  a reason for using the `auto foo() -> bool` syntax instead of simply `bool foo()`?  It feels kind of distracting to use this style here as a one-off.

opti-mix: Good catch! Initially, I noticed the closure and asked @beicy to change it to a function. But I haven't noticed that she forgot to change the return type while changing it ;-)

beicy: Thanks! I didn't notice that :P Will add a follow up PR.

PR #3008, opened 2019-05-30
build: add a check for `llvm-link`, allow overriding

compnerd:
Summary:

Allow the user to specify `-DLLVM_LINK_BIN=` rather than assuming that it exists
as a particular location.  This already does not account for configuration types
properly except in the case of MSVC, and does not work well with all ways to
build LLVM.  We should either use the imported target for `llvm-link` as we were
or for now allow this override.

We now report a fatal error if `llvm-link` is not found and the CPU backend is
enabled as that is required to build the runtime.

Documentation:
n/a

Test Plan:
CI


jfix71:
It's still failing on CI though...

PR #3015, opened 2019-05-30
Prepare Glow Function for training, replacing constants with placeholders.

putivsky:
Summary:
Once graph has been loaded into Function weights got transformed to Constants (default), graph is ready for inference.
For training Constants must be converted into the correspoindent Placeholders with an optional initialization of the tensors.

Differential Revision: D15563530

### DIFF
--- a/examples/training/resnet50/main.cpp
+++ b/examples/training/resnet50/main.cpp
@@ -1,0 +1,272 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Base/Image.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/Graph/Nodes.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "glow/Optimizer/TrainingPreparation.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Timer.h"
+
+#include <fstream>
+
+/// This is an example demonstrating how to load resnet50 model,
+/// randomize weights and perform training on a limited data set,
+/// and evaluate model.
+
+constexpr size_t IMAGE_COLORS = 3;
+constexpr size_t IMAGE_HEIGHT = 224;
+constexpr size_t IMAGE_WIDTH = 224;
+constexpr size_t LABELS_SIZE = 1000;
+constexpr size_t CIFAR_NUM_IMAGES = 20;
+
+using namespace glow;
+
+namespace {
+llvm::cl::OptionCategory category("resnet-training Options");
+llvm::cl::opt<unsigned> numEpochs("epochs", llvm::cl::desc("Number of epochs."),
+                                  llvm::cl::init(32), llvm::cl::value_desc("N"),
+                                  llvm::cl::cat(category));
+llvm::cl::opt<unsigned> miniBatchSize("mini-batch",
+                                      llvm::cl::desc("Mini batch size."),
+                                      llvm::cl::init(8),
+                                      llvm::cl::value_desc("N"),
+                                      llvm::cl::cat(category));
+llvm::cl::opt<float> learningRate("learning-rate",
+                                  llvm::cl::desc("Learning rate parameter."),
+                                  llvm::cl::init(0.001),
+                                  llvm::cl::value_desc("N"),
+                                  llvm::cl::cat(category));
+llvm::cl::opt<float> momentum("momentum", llvm::cl::desc("Momentum parameter."),
+                              llvm::cl::init(0.9), llvm::cl::value_desc("N"),
+                              llvm::cl::cat(category));
+llvm::cl::opt<float> l2Decay("l2-decay", llvm::cl::desc("L2Decay parameter."),
+                             llvm::cl::init(0.0001), llvm::cl::value_desc("N"),
+                             llvm::cl::cat(category));
+llvm::cl::opt<std::string>
+    cifar10DbPath("cifar10", llvm::cl::desc("Path to the CIFAR-10 database."),
+                  llvm::cl::init("cifar-10-batches-bin/data_batch_1.bin"),
+                  llvm::cl::value_desc("db path"), llvm::cl::Optional,
+                  llvm::cl::cat(category));
+
+llvm::cl::opt<std::string>
+    resnet50Path("resnet50", llvm::cl::desc("Path to the ResNet50 model."),
+                 llvm::cl::init("resnet50"), llvm::cl::value_desc("model path"),
+                 llvm::cl::Optional, llvm::cl::cat(category));
+llvm::cl::opt<BackendKind> executionBackendKind(
+    llvm::cl::desc("Backend to use:"), llvm::cl::Optional,
+    llvm::cl::values(clEnumValN(BackendKind::Interpreter, "Interpreter",
+                                "Use interpreter (default option)"),
+                     clEnumValN(BackendKind::CPU, "cpu", "Use CPU"),
+                     clEnumValN(BackendKind::OpenCL, "opencl", "Use OpenCL")),
+    llvm::cl::init(BackendKind::Interpreter), llvm::cl::cat(category));
+
+/// Finds the maximum value and its index within \p N range.
+size_t findMaxIndex(const Tensor &t, size_t range) {
+  auto handle = t.getHandle();
+  size_t ret{0};
+  float maxVal{handle.raw(0)};
+  for (size_t c = 1; c < range; ++c) {
+    auto val = handle.raw(c);
+    if (maxVal <= val) {
+      maxVal = val;
+      ret = c;
+    }
+  }
+  return ret;
+}
+
+void loadImagesAndLabels(Tensor &images, Tensor &labels) {
+  llvm::outs() << "Loading the CIFAR-10 database.\n";
+  /// ResNet expects ImageChannelOrder::BGR, ImageLayout::NCHW.
+  std::ifstream dbInput(cifar10DbPath, std::ios::binary);
+
+  if (!dbInput.is_open()) {
+    llvm::errs() << "Failed to open cifar10 data file, probably missing.\n";
+    llvm::errs() << "Run 'python ../glow/utils/download_test_db.py'\n";
+    exit(1);
+  }
+
+  /// The CIFAR file format is structured as one byte label in the range 0..9.
+  /// The label is followed by an image: 32 x 32 pixels, in RGB format. Each
+  /// color is 1 byte. The first 1024 red bytes are followed by 1024 of green
+  /// and blue. Each 1024 byte color slice is organized in row-major format.
+  /// The database contains 10000 images.
+  /// Size: (1 + (32 * 32 * 3)) * 10000 = 30730000.
+  llvm::outs() << "Assigning images/labels bytes.\n";
+  auto labelsH = labels.getHandle();
+  auto imagesH = images.getHandle();
+  float scale = 1. / 255.0;
+  float bias = 0;
+
+  /// CIFAR image dimensions are 32x32, ResNet50 expects 224x224.
+  /// Simple scaling would be one CIFAR pixel copied into 7x7 area
+  constexpr unsigned expandImage = 7;
+  for (unsigned n = 0; n < CIFAR_NUM_IMAGES; ++n) {
+    unsigned long l = dbInput.get();
+    labelsH.at({n, l}) = 1;
+    // Even ResNet50 model got trained in NCHW format
+    for (unsigned c = 0; c < IMAGE_COLORS; ++c) {
+      auto bgrc = IMAGE_COLORS - 1 - c;
+      for (unsigned h = 0; h < 32; ++h) {
+        for (unsigned w = 0; w < 32; ++w) {
+          // ResNet BGR color space vs CIFAR RGB.
+          auto val = static_cast<float>(static_cast<uint8_t>(dbInput.get()));
+          // Normalize and scale image.
+          val = (val - imagenetNormMean[c]) * scale / imagenetNormStd[c] + bias;
+          for (unsigned sh = 0; sh < expandImage; ++sh) {
+            for (unsigned sw = 0; sw < expandImage; ++sw) {
+              imagesH.at(
+                  {n, bgrc, sh + h * expandImage, sw + w * expandImage}) = val;
+            }
+          }
+        }
+      }
+    }
+  }
+}
+} // namespace
+
+int main(int argc, char **argv) {
+  llvm::cl::ParseCommandLineOptions(argc, argv,
+                                    " ResNet50 Training Example\n\n");
+
+  // We expect the input to be NCHW.
+  llvm::ArrayRef<size_t> allImagesDims = {CIFAR_NUM_IMAGES, IMAGE_COLORS,
+                                          IMAGE_HEIGHT, IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> batchImagesDims = {miniBatchSize, IMAGE_COLORS,
+                                            IMAGE_HEIGHT, IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> initShapeDims = {1, IMAGE_COLORS, IMAGE_HEIGHT,
+                                          IMAGE_WIDTH};
+  llvm::ArrayRef<size_t> allLabelsDims = {CIFAR_NUM_IMAGES, LABELS_SIZE};
+
+  Module mod;
+  PlaceholderBindings bindings;
+  const char inputName[] = "gpu_0/data";
+  TypeRef inputType = mod.uniqueType(ElemKind::FloatTy, initShapeDims);
+  // Load ResNet model.
+  Function *F = mod.createFunction("resnet50");
+  llvm::outs() << "Loading resnet50 model.\n";
+  llvm::Error errPtr = llvm::Error::success();
+
+  // Loader has randomly initialized trainable weights.
+  Caffe2ModelLoader loader(resnet50Path + "/predict_net.pbtxt",
+                           resnet50Path + "/init_net.pb", {inputName},
+                           {inputType}, *F, &errPtr);
+
+  if (errPtr) {
+    llvm::errs() << "Loader failed to load resnet50 model from path: "
+                 << resnet50Path << "\n";
+    return -1;
+  }
+
+  llvm::outs() << "Creating input/output placeholders.\n";
+  // Create the input layer.
+  auto *A = mod.createPlaceholder(ElemKind::FloatTy, batchImagesDims, "input",
+                                  /* isTraining */ false);
+  bindings.allocate(A);
+  // Create the output layer.
+  auto *E =
+      mod.createPlaceholder(ElemKind::FloatTy, {miniBatchSize, LABELS_SIZE},
+                            "expected", /* isTraining */ false);
+  bindings.allocate(E);
+
+  llvm::outs() << "Preparing for training.\n";
+
+  if ((errPtr = prepareFunctionForTraining(F, bindings))) {
+    return -1;
+  }
+
+  llvm::outs() << "Differentiating graph.\n";
+  TrainingConfig TC;
+  TC.learningRate = learningRate;
+  TC.momentum = momentum;
+  TC.L2Decay = l2Decay;
+  TC.batchSize = miniBatchSize;
+  Function *TF = glow::differentiate(F, TC);
+
+  llvm::outs() << "Compiling backend.\n";
+  BackendKind executionBackend{executionBackendKind};
+  ExecutionEngine EE(executionBackend);
+  EE.compile(CompilationMode::Train, TF);
+
+  // Allocate placeholders without tensors.
+  bindings.allocate(mod.getPlaceholders());
+
+  // Get output tensor.
+  auto *resultPH = mod.getPlaceholderByName("save_gpu_0_softmax");
+  auto *result = bindings.get(resultPH);
+
+  // These tensors allocate memory for all images and labels prepared for
+  // training.
+  Tensor images(ElemKind::FloatTy, allImagesDims);
+  Tensor labels(ElemKind::FloatTy, allLabelsDims);
+
+  loadImagesAndLabels(images, labels);
+
+  // From http://www.cs.toronto.edu/~kriz/cifar.html.
+  const char *textualLabels[LABELS_SIZE] = {
+      "airplane", "automobile", "bird",  "cat",  "deer",
+      "dog",      "frog",       "horse", "ship", "truck"};
+
+  size_t sampleCounter = 0;
+  int reportRate = 16;
+
+  llvm::Timer timer("Training", "Training");
+
+  llvm::outs() << "Training.\n";
+  auto labelsH = labels.getHandle();
+  for (size_t e = 0; e < numEpochs; ++e) {
+    llvm::outs() << "Training - epoch #" << e << " from total " << numEpochs
+                 << "\n";
+    // Train network on all examples.
+    timer.startTimer();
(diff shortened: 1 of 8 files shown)


lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,113 @@
...
+ * limitations under the License.
+ */
+
+#include "glow/Optimizer/TrainingPreparation.h"
+#include "glow/Graph/PlaceholderBindings.h"
+
+namespace glow {
+
+TensorInitializer getDefaultTensorInitializer() {
+  return [](Function *F, Node *node, NodeValue &input, Tensor *tensor) {
```
opti-mix: Nit: This closure does not capture anything. It could be expressed as a static free-standing function before `getDefaultTensorInitializer ` and `getDefaultTensorInitializer` would just return it.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,113 @@
...
+    TensorInitializer &&initializer) {
+
+  auto &nodes = F->getNodes();
+  // Lookup all nodes, skip Storage types, enumerate inputs,
+  // replace Constant type with trainable Placeholders except special cases,
+  // like BatchNormalization inputs (mean and variance). In special cases
+  // replace Constant type with non-trainable Placeholders.
+  for (auto &node : nodes) {
+    // Skip storages.
+    if (Storage::classof(&node)) {
```
opti-mix: The more usual way how it is done in Glow and LLVM is `llvm::isa<Storage>(&node)`.

putivsky: Class method checks if node is Constant or Placeholder. I know at this moment there are only two derived classes but it could be changed in the future.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+        switch (node.getKind()) {
+          // Special case. Mean and Variance inputs are not trainable.
+        case Kinded::Kind::BatchNormalizationNodeKind: {
+          if (BatchNormalizationNode *BN =
+                  llvm::dyn_cast<BatchNormalizationNode>(&node)) {
+            if (INV == BN->getMean() || INV == BN->getVar()) {
+              isTrainable = false;
+            }
+          }
+        } break;
```
SplitInfinity: Nit: I think we usually write `break` inside the scope for the case. I didn't know you could do this actually...

lib/Optimizer/TrainingPreparation.cpp
SplitInfinity: I agree with @opti-mix. `llvm::isa<Storage>(&node)` will work even if more derived classes are added, right? What is the advantage of using `Storage::classof`?

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+        default:
+          break;
+        }
+        auto *PH = F->getParent()->createPlaceholder(C->getType(), C->getName(),
+                                                     isTrainable);
+        C->getOutput().replaceAllUsesOfWith(PH, F);
+        bindings.insert(PH, std::move(C->getPayloadMutable()));
+        F->getParent()->eraseConstant(C);
+        auto *tensor = bindings.get(PH);
```
SplitInfinity: I *think* you can avoid this extra lookup if you initialize the tensor before inserting it into `bindings`:
```
auto &tensor = C->getPayloadMutable();
initializer(F, &node, INV, tensor);
bindings.insert(PH, std::move(tensor));
F->getParent()->eraseConstant(C);
```

include/glow/Optimizer/TrainingPreparation.h
```diff
@@ -0,0 +1,52 @@
...
+#include "glow/Graph/Graph.h"
+
+#include <functional>
+
+namespace glow {
+
+/// Function for \p tensor initialization.
+/// Method gets following parameters Glow Function \p F, current \p node,
+/// \p input, and fills out (optionally) \p tensor.
+using TensorInitializer = void (*)(Function *F, Node *node, NodeValue &input,
```
SplitInfinity: Out of curiosity, why did you use a function pointer type here instead of something like `std::function<void(Function*, Node*, NodeValue&)>`?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1199,8 +1199,8 @@ ONNXModelLoader::ONNXModelLoader(const std::string &modelDescFilename,
     RETURN_IF_ERR(setVersion(modelDef));
 
     ONNX_NAMESPACE::GraphProto graphDef = modelDef.graph();
-    RETURN_IF_ERR(checkInputs(graphDef, tensorNames, types));
 
+    RETURN_IF_ERR(checkInputs(graphDef, tensorNames, types));
```
SplitInfinity: Looks like an extra space was inserted above this line. Was this a `clang-format` change? If not, please undo.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // like BatchNormalization inputs (mean and variance). In special cases
+  // replace Constant type with non-trainable Placeholders.
+  for (auto &node : nodes) {
+    // Skip storages.
+    if (Storage::classof(&node)) {
+      continue;
+    }
+
+    // Skip output node(s).
+    if (legalOutputNames.count(node.getName())) {
```
SplitInfinity: Do we need this information to be explicitly provided? Can we infer which nodes are outputs? I'll admit I didn't think about this too deeply but I think `Placeholders` that are written to only once by `SaveNodes` should be outputs, right?

examples/training/resnet50/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+  llvm::outs() << "Preparing for training.\n";
+
+  auto &outputs = loader.getOutputVarsMapping();
+  std::unordered_set<std::string> legalOutputNames;
+  for (auto &output : outputs) {
+    legalOutputNames.insert(legalizeName(output.getKey()));
+  }
+
+  prepareFunctionForTraining(F, bindings, legalOutputNames);
```
SplitInfinity: I would prefer if we could hide this inside `glow::differentiate` or `compile`. If we can infer function outputs as per my other comment, I think we can do this. If not, we'll have to live with it.

lib/Optimizer/CMakeLists.txt
```diff
@@ -2,13 +2,12 @@ add_library(Optimizer
...
               GraphOptimizer.cpp
               Lower.cpp
-              Quantization.cpp)
+              Quantization.cpp
+              TrainingPreparation.cpp)
 
 target_link_libraries(Optimizer
                       PRIVATE
                         Backend
-                        Converter
```
jfix71: Why did you remove dependencies here? This seems like an accidental change?

include/glow/Optimizer/TrainingPreparation.h
putivsky: I did, but Roman has suggested the opposite, i.e. function vs lambda. In my opinion, lambda is more flexible, cause can carry captured objects. I can reverted back to lambda if Roman is ok with that.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+            }
+          }
+        } break;
+        default:
+          break;
+        }
+        auto *PH = F->getParent()->createPlaceholder(C->getType(), C->getName(),
+                                                     isTrainable);
+        C->getOutput().replaceAllUsesOfWith(PH, F);
+        bindings.insert(PH, std::move(C->getPayloadMutable()));
```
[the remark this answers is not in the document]
putivsky: It would temporarily double the memory.

examples/training/resnet50/main.cpp
putivsky: We have a discussion and decision was - have a separate from loader API that user can use to prepare Function for training. Graph can be crated programmatically, in such a case there is no need to call API at all.

lib/Optimizer/CMakeLists.txt
putivsky: Quantization library includes Converter

lib/Optimizer/TrainingPreparation.cpp
putivsky: Loader has information about output nodes, but once we have decoupled API from Loader, it's kind of hard to do. I have tried and it looked like a hack with insights into Loader code.

include/glow/Optimizer/TrainingPreparation.h
opti-mix: I suggested to use the function, yes.

But the type itself can and probably should be std::function. This is a different thing. You can still use a usual function (i.e. non-closure) like you have today to assign to objects of this type. And you can assign closures to it as well if necessary.

lib/Optimizer/TrainingPreparation.cpp
putivsky: For instance, the output node is SoftMax in my testing example, not SaveNode and it can be other types besides SoftMax.

jfix71: Doesn't the SoftMax feed into a Save, though?

lib/Optimizer/CMakeLists.txt
jfix71: Sure, but Optimizer directly depends on Converter itself as well. If Quantization somehow no longer depends on Converter in the future then this will break. Removing this just makes the dependency graph less accurate. Is there some downside to keeping Converter etc. here?

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    // Skip output node(s).
+    if (legalOutputNames.count(node.getName())) {
+      continue;
+    }
+
+    for (unsigned idx = 0; idx < node.getNumInputs(); ++idx) {
+      auto INV = node.getNthInput(idx);
+      auto *IN = INV.getNode();
+      if (Constant *C = llvm::dyn_cast<Constant>(IN)) {
```
jfix71: We usually format these slightly differently to keep indentation levels lower, e.g.:

```cpp
Constant *C = llvm::dyn_cast<Constant>(IN);
if (!C) {
  continue;
}
...
```

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+        bool isTrainable = true;
+        switch (node.getKind()) {
+          // Special case. Mean and Variance inputs are not trainable.
+        case Kinded::Kind::BatchNormalizationNodeKind: {
+          if (BatchNormalizationNode *BN =
+                  llvm::dyn_cast<BatchNormalizationNode>(&node)) {
+            if (INV == BN->getMean() || INV == BN->getVar()) {
```
jfix71: You can simply check if `idx == BatchNormalizationNode::MeanIdx` etc. instead of all this logic of casting the BN and getting the actual inputs.

lib/Optimizer/TrainingPreparation.cpp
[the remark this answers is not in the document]
jfix71: Ok, sure. But are there still issues here if a Constant is used by more than one Node? What happens if we have an unrolled RNN, with LSTMs that all use the same Constant? They should all point to the same Trainable PH, right?

And actually, I think we currently only allow a single Save Node per Placeholder, so there may be some work to do here...

But regardless if we tried to run this e.g. on our en2gr example, it would probably have use after free issues as we'd be accessing an erased Constant. I think at the very least we should assert here that the Constant has no more users before erasing it.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (Storage::classof(&node)) {
+      continue;
+    }
+
+    // Skip output node(s).
+    if (legalOutputNames.count(node.getName())) {
+      continue;
+    }
+
+    for (unsigned idx = 0; idx < node.getNumInputs(); ++idx) {
```
jfix71: nit: Based on [LLVM coding standards](https://releases.llvm.org/7.0.0/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop), we prefer:

```cpp
for (unsigned idx = 0, e = node.getNumInputs(); idx < e; ++idx)
```

lib/Optimizer/TrainingPreparation.cpp
putivsky: Let me try, traverse nodes twice, one to find SaveNode and the correspondent input (whatever it is) and skip Placeholder replacement for that.

include/glow/Optimizer/TrainingPreparation.h
```diff
@@ -0,0 +1,44 @@
...
+#include "glow/Graph/Graph.h"
+#include "glow/Support/Error.h"
+
+#include <functional>
+
+namespace glow {
+
+/// Function for \p tensor initialization.
+/// Method gets following parameters Glow Function \p F, current \p node,
+/// \p input, and fills out (optionally) \p tensor.
```
jfix71: nit: Update Doxygen to reference `inputIdx`.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "glow/Graph/PlaceholderBindings.h"
+
+namespace glow {
+
+namespace {
+void defaultTensorInitializer(Function *F, Node *node, unsigned inputIdx,
+                              Tensor *tensor) {
+  switch (node->getKind()) {
+  case Kinded::Kind::ConvolutionNodeKind: {
+    ConvolutionNode *CN = llvm::dyn_cast<ConvolutionNode>(node);
```
jfix71: nit: Can be `llvm::cast` instead of `llvm::dyn_cast` since we're inside the case based on the Node's kind.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,147 @@
...
+      ShapeHW kdim(CN->getKernels());
+      size_t fanIn = kdim.height * kdim.width * idim.c;
+      tensor->init(Tensor::InitKind::Xavier, fanIn, F->getPRNG());
+    } else if (inputIdx == CN->BiasIdx) {
+      tensor->init(Tensor::InitKind::Broadcast, 0.1, F->getPRNG());
+    }
+    break;
+  }
+  case Kinded::Kind::BatchNormalizationNodeKind: {
+    BatchNormalizationNode *BN = llvm::dyn_cast<BatchNormalizationNode>(node);
```
jfix71: nit: Can be `llvm::cast` instead of `llvm::dyn_cast` since we're inside the case based on the Node's kind. And you probably don't even need to cast since you're only accessing the enums from the node -- you could avoid casting it entirely and below replace e.g. `BN->ScaleIdx` with `BatchNormalizationNode::ScaleIdx`.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,147 @@
...
+      tensor->init(glow::Tensor::InitKind::Broadcast, 0.1, F->getPRNG());
+    } else if (inputIdx == BN->MeanIdx) {
+      tensor->zero();
+    } else if (inputIdx == BN->VarIdx) {
+      tensor->init(glow::Tensor::InitKind::Broadcast, 1.0, F->getPRNG());
+    }
+    break;
+  }
+  case Kinded::Kind::FullyConnectedNodeKind: {
+    FullyConnectedNode *FCN = llvm::dyn_cast<FullyConnectedNode>(node);
```
jfix71: nit: Can be `llvm::cast` instead of `llvm::dyn_cast` since we're inside the case based on the Node's kind.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,147 @@
...
+  return defaultTensorInitializer;
+}
+
+llvm::Error prepareFunctionForTraining(Function *F,
+                                       PlaceholderBindings &bindings,
+                                       TensorInitializer &&initializer) {
+
+  auto &nodes = F->getNodes();
+
+  Node *outputNode{nullptr};
```
jfix71: Can you explain why we need to special case output nodes? When we have the loop below that actually does the Constant -> PH conversion, if we encounter a SaveNode we could skip it just like we do Storage Nodes, no? Or if its input is a Constant I suppose we could even convert its Placeholder to a Constant, though it's a bit of a silly case?

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,147 @@
...
+      if (!C) {
+        continue;
+      }
+      bool isTrainable = true;
+      switch (node.getKind()) {
+        // Special case. Mean and Variance inputs are not trainable.
+      case Kinded::Kind::BatchNormalizationNodeKind: {
+        BatchNormalizationNode *BN =
+            llvm::dyn_cast<BatchNormalizationNode>(&node);
+        if (BN && (idx = BN->MeanIdx || idx == BN->VarIdx)) {
```
jfix71: nit: Similar to earlier comments -- instead of casting in here and needing to check the node plus follow pointers to get the idx from the enum, the last few lines could be replaced with `if (idx == BatchNormalizationNode::MeanIdx && idx == BatchNormalizationNode::VarIdx)`.

jfix71: Also there's a typo here, missing an `=` in `idx = BN->MeanIdx`

lib/Optimizer/TrainingPreparation.cpp
putivsky: If output node (SoftMax) constant(s) got replaced with Placeholder then verification fails. Skipping SaveNode doesn't help, cause we need actually skip "SoftMax" of similar aggregation output node, which is in the chain: SaveNode -> Input (ReshapeNode) -> Input (SoftMax or similar).

putivsky: I'm still thinking this diff idea won't work in general case. Having one place where we load graph and making proper replacements (in a training mode) is a more reliable approach.

jfix71: I don't understand -- why is it a problem to convert the Softmax's `Selected` `Constant` input to a `Placeholder`? What verification fails?

putivsky: Here is the errors output:
Loading resnet50 model.
Creating input/output placeholders.
Preparing for training.
Differentiating graph.
Compiling backend.
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Add
name : gij_with_l2214
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 1
Result : index64<1 x 1>
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Add
name : dx_with_momentum214
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 2
Result : index64<1 x 1>
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Add
name : newW214
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 1
Result : index64<1 x 1>
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Mul
name : L2Grad2141
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 1
Result : index64<1 x 1>
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Mul
name : dx2141
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 1
Result : index64<1 x 1>
Unsupported node found while compiling Function resnet50_grad for backend Interpreter: Mul
name : GsumMult2141
LHS : index64<1 x 1>
RHS : index64<1 x 1>
users : 1
Result : index64<1 x 1>
Encountered an error, exiting.
location: /Users/putivsky/local/pytorch/glow/lib/Optimizer/GraphOptimizer.cpp:2725 error code: COMPILE_UNSUPPORTED_NODE_AFTER_OPTIMIZE message: Unsupported node(s) found after optimizing Function resnet50_grad for backend Interpreter
p

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,141 @@
...
+                                                   isTrainable);
+      C->getOutput().replaceAllUsesOfWith(PH, F);
+      RETURN_ERR_IF_NOT(!C->hasUsers(), "Constant is still in use.");
+      auto &tensor = C->getPayloadMutable();
+      initializer(F, &node, idx, &tensor);
+      bindings.insert(PH, std::move(tensor));
+    }
+  }
+
+  for (auto C : constantsToRemove) {
```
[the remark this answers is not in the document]
putivsky: C is a pointer, but I can keep the style.

lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,141 @@
...
+      if (!C) {
+        continue;
+      }
+      bool isTrainable =
+          !isBatchNormalization || (BatchNormalizationNode::MeanIdx != idx &&
+                                    BatchNormalizationNode::VarIdx != idx);
+      auto *PH = F->getParent()->createPlaceholder(C->getType(), C->getName(),
+                                                   isTrainable);
+      C->getOutput().replaceAllUsesOfWith(PH, F);
+      RETURN_ERR_IF_NOT(!C->hasUsers(), "Constant is still in use.");
```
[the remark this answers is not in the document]
putivsky: Erasing in a loop causes problem, potential iterators invalidation.

putivsky:
@jfix71, as I wrote, the purpose of this PR is not to get any meaningful training results, but rather to demonstrate the way of turning loaded model into the trainable one.
Regarding deleting Constants inside the loop, when I compiled with Buck and run it produces the verification errors.

jfix71:
> @jfix71, as I wrote, the purpose of this PR is not to get any meaningful training results, but rather to demonstrate the way of turning loaded model into the trainable one.

Ok I think that makes sense -- but then how have you tested `resnet-training`? What do you think about adding new unit tests that test `prepareFunctionForTraining()` more directly? It would create simple inference graphs (i.e. using Constants instead of Placeholders), and then run `prepareFunctionForTraining()` on it, and then verify the resulting Function is as expected. You could introduce unit tests for small pieces of a resnet that represent subgraphs from Resnet50, e.g. a unit test with Conv+BN+Relu+Pool, etc., and then the unit tests check that all Constants are moved to PHs with their trainable boolean correctly set.

> Regarding deleting Constants inside the loop, when I compiled with Buck and run it produces the verification errors.

Huh, weird, since we are never iterating over the Constants there. Iterating over the Nodes shouldn't cause any issues there, so not sure why that has occurred. Do you remember more info here?

putivsky:
<<  but then how have you tested resnet-training >>
Let me provide a more detailed answer, what is this PR for.
If I wanted to confirm that resnet-training really got training and model got converged, I should implement input training images augmentation, cropping, rotating, etc. and have millions such images to produce any meaningful model accuracy for 1000 categories. The size of such images 224x224x3x1M would be ~ 150GB. Moreover such model usually got training on multiple GPUs for weeks. For single-threaded CPU it would take forever. Goal of this PR is to confirm following things:
- Replacement of the most of the Constant (presumably weights) by Placeholders doesn't break any verification. 
- ResNet model can be successfully differentiated (no verification error) and all operators for ResNet have the correspondent Grad operators.
- Training algorithm runs without errors or warnings.
That's it, nothing else.
Next step would be to pick-up a simple, small model something similar to Cifar10 and apply prepareFunctionForTraining and confirm that training model got converged.

jfix71:
>- Replacement of the most of the Constant (presumably weights) by Placeholders doesn't break any verification.
>- ResNet model can be successfully differentiated (no verification error) and all operators for ResNet have the correspondent Grad operators.

Right now we aren't actually testing that `prepareFunctionForTraining()` correctly converted the Function for differentiation/training, because there is no check that the structure of the resulting Function is correct.

E.g. we don't check anywhere that Constants were replaced by PHs, nor that ResNet has the corresponding Grad operators after differentiation. Differentiation won't have any verification errors if you completely skip the call to `prepareFunctionForTraining()`. And as you said, we also can't check that the result of `resnet-training` is correct because it will take too long to do so.

This is why I am suggesting we at least add in a few basic unit tests that build a simple inference graph, call `prepareFunctionForTraining()`, and then check the structure of the Function is correct (i.e. Constants replaced by PHs with trainable set correctly). It would be similar to our `GraphOptz` tests, which run optimize and check that the graph is structured as expected.

>Next step would be to pick-up a simple, small model something similar to Cifar10 and apply prepareFunctionForTraining and confirm that training model got converged.

Yeah this is a great next step. It may make sense to use our `lenet-mnist` example since it's also loaded from a proto just like Resnet50, and I was able to train it on my laptop in a couple hours.

putivsky:
ok, let  me abound this PR and forget about resnet50, and switch to simple model first with a unit test verification.

putivsky:
@jfix71, FYI: https://github.com/pytorch/glow/pull/3075

PR #3017, opened 2019-05-31
build: try to goad CMake into understanding dependencies

compnerd:
Change the build rules for the JIT to use some more trickery to have CMake track
dependencies itself.  First, we swap out the compiler around the definition of
JIT module to ensure that we build with Clang.  Second, we build the library as
a module to control the link rule invocation.  Third, we change the linker
language to LLIR and provide a custom linker rule to invoke the LLVM linker.
This should hopefully provide more reliable null builds.  As a happy side
effect, the logic is easier to read.

Summary:

Documentation:

Test Plan:

[Optional Fixes #issue]

### DIFF
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -36,7 +36,7 @@ using namespace glow;
 /// The resulting file is included here to compile the bitcode image into our
 /// library.
 static const unsigned char libjit_bc[] = {
-#include "glow/CPU/libjit_bc.inc"
+#include "libjit.inc"
 };
 static const size_t libjit_bc_size = sizeof(libjit_bc);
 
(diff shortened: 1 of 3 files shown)


lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -36,7 +36,7 @@ using namespace glow;
 /// The resulting file is included here to compile the bitcode image into our
 /// library.
 static const unsigned char libjit_bc[] = {
-#include "glow/CPU/libjit_bc.inc"
+#include "glow/CPU/libjit.inc"
```
rdzhabarov: this most likely will break fbcode version (need corresponding changes)

compnerd: Yeah, I think that dealing with copying files is kinda silly and just complicates things, we should just fix the include paths instead.

PR #3018, opened 2019-05-31
[ConvTest] Use compareAgainstInterpreter(); add more precisions

jfix71:
Summary: Move ConvTest to be similar to all other tests comparing against backends. This also allows for easy testing of varying precisions; I added Int8 and FP16 tests.

Test Plan: New tests pass. Original test shouldn't have been functionally changed.

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -80,6 +80,7 @@ add_executable(ConvTest
                ConvTest.cpp)
 target_link_libraries(ConvTest
                       PRIVATE
+                        BackendTestUtils
                         Graph
                         IR
                         ExecutionEngine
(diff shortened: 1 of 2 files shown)


tests/unittests/CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ add_executable(ConvTest
                ConvTest.cpp)
 target_link_libraries(ConvTest
                       PRIVATE
+                        BackendTestUtils
```
rdzhabarov: no changes on fbcode side?

jfix71: There will be a simple buck change -- I haven't quite gotten a hang of the right work flow here. I tend to just work on GH directly then make the fbcode-specific changes internally. What's the preferred way, if something needs fbcode changes then do it internally and export? Either way the changes don't show up here, but I guess the benefit is you see them when looking on Phabricator right off the bat?

jfix71: Updated fbcode dependencies.

rdzhabarov: > I tend to just work on GH directly then make the fbcode-specific changes internally. 

I think this is fine approach. But we need to make sure that fbcode is not broken with the change. It's currently broken, so we'd need to fix TARGETS file before merging this PR.

jfix71: Yeah I've updated it, I guess it means I need another approval on the diff too in phabricator?

rdzhabarov: >  I need another approval on the diff too

done.

Issue #3012, opened 2019-05-30
RecommendationSystemTest should produce non-zero reference results


opti-mix:
RecommendationSystemTest produces zero tensors as reference results in many cases (see below). This is not very useful for comparing the results of a reference run with results produced e.g. by different backends or by partitioned execution. The test should be improved to produce more meaningful, non-zero results.

```
[==========] Running 39 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 39 tests from RecSys/RecommendationSystemTest
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_FP32/0
Number of embeddings concatenated: 11
Reference results:
shape: ( 16 1 )
[ Zero tensor ]
[       OK ] RecSys/RecommendationSystemTest.RecSys_FP32/0 (2097 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_FP32/1
[       OK ] RecSys/RecommendationSystemTest.RecSys_FP32/1 (117 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_FP32/2
[       OK ] RecSys/RecommendationSystemTest.RecSys_FP32/2 (0 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/0
Number of embeddings concatenated: 11
Reference results:
shape: ( 16 1 )
[ Zero tensor ]
[       OK ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/0 (4431 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/1
[       OK ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/1 (0 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/2
[       OK ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS_FC/2 (0 ms)
[ RUN      ] RecSys/RecommendationSystemTest.RecSys_RWQuantized_SLWS/0
Number of embeddings concatenated: 11
Reference results:
shape: ( 16 1 )
max: 0.002  min: 0.000
[[0.001],
[0.000],
[0.002],
[0.000],
[0.000],
[0.001],
[0.001],
[0.001],
[0.000],
[0.001],
[0.001],
[0.000],
[0.000],
[0.000],
[...]

PR #3020, opened 2019-05-31
[Unittest] Improve RecSys so that float MLP produces results other than 0.

artemrakhov:
Summary: the last FC layer of top-level MLP is essentially a dot product of non-negative vector with a lot of 0 by a vector with equal number of positives and negatives. Due to high number of 0s in LHS vector, it is very likely that we'll get a negative number after reduce, which turns into 0 because of the following Relu. I used skewed distribution instead, so that RHS vector has more positive values, and result is more likely to be positive.

Test Plan: `tests/RecommendationSystemTest`
Added debug prints to ensure the result is not 0 now.

Improves the situation with issue #3012

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -1698,6 +1698,7 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     FloatTy,       // 32-bit float type (float)
     Float16Ty,     // 16-bit float type (half, fp16)
     Int8QTy,       // 8-bit quantized type (int8_t)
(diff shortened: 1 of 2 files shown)


lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1698,6 +1698,7 @@ libjit_dump_tensor(uint8_t *tensor, size_t *tensorDim, size_t numDimsTensor,
     FloatTy,       // 32-bit float type (float)
     Float16Ty,     // 16-bit float type (half, fp16)
     Int8QTy,       // 8-bit quantized type (int8_t)
+    UInt8QTy,      // unsigned 8-bit quantized type (uint8_t)
```
rdzhabarov: good catch!

nickgg:
Awesome, thanks @artemrakhov

PR #3021, opened 2019-05-31
[unittests] Move ConvTest to ParameterSweepTest, and add MatMul to it

jfix71:
Summary: Rename the ConvTest to ParameterSweepTest, and then add MatMul sweep tests to it. I somewhat arbitrarily picked the dimensions for MatMul sweeping. FYI on my macbook it took ~80 seconds to run these tests sequentially, and ~26 seconds in parallel w/ gtest-parallel.

Test Plan: We can use these tests to exercise different configs of important ops on different backends. Note that this also uses the `parCloneCountOpt`, meaning we can use it to clone the Functions of these tests many times to try to test many different parallel execution units of an architecture.


jfix71:
I've updated fbcode too, so please review/accept the diff in phabricator as well 😃

PR #3022, opened 2019-05-31
Use pooling param structure from SDK headers

bertmaher:
Summary:
ns_SpatialReduction is the vendor-provided struct; no need to define
our own.

Differential Revision: D15575080

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -93,28 +93,6 @@ static std::string getKernelName(llvm::StringRef kernelBase, ElemKind kind) {
 }
 
 namespace {
-/// Parameters for pooling operation.
-struct synPoolParams {
-  // Padding
-  int pWbegin;
-  int pWend;
-  int pHbegin;
-  int pHend;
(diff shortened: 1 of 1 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -564,26 +541,26 @@ static std::unique_ptr<synConvolutionParams> makeSynConvolutionParams(
   return params;
 }
 
-static std::unique_ptr<synPoolParams>
+static std::unique_ptr<ns_SpatialReduction::Params>
```
rdzhabarov: we actually have dilation param now, but we can fix it later.

bertmaher:
> Was ns_SpatialReduction::Params just defined in recent 1.8.4?

It's been there since at least 0.1.7.1.  This struct is kind of a hangover from a very early resnet example that defined these parameter structs in the user code.

lib/Backends/Habana/Habana.cpp
SplitInfinity: I thought we only have dilation for convolution?

PR #3024, opened 2019-05-31
Remove Habana-specific instructions

bertmaher:
Summary: We switched to the Graph IR a long time ago and never cleaned this up.

Differential Revision: D15575254


rdzhabarov:
Can you rebase on the latest master: ```Temporary failure resolving 'apt.llvm.org'```
should not happen as we do not rely on it.

PR #3027, opened 2019-05-31
Fix AOT bundle CMake build and regression tests

### LINKED ISSUE #3013: AOT Bundles should be regression tested
Currently there is no unit tests or regression tests for Bundle saving and loading, we could probably just write a script that runs the four example bundles lenet_mnist and resnet50 (+ their quantized versions) and checks for expected outcomes.

nickgg:
Summary: The AOT bundle examples had multiple makefiles, of which the CMake version did not work properly. This fixes the CMakeLists for the two bundle examples (plus the two extra quantized versions) and deletes the extra Makefile. 

We did regression test the mnist bundle using the non-CMake makefile, but not resnet50 - I've modified the CI script to test both bundles using the CMakeList version.

Documentation: updated AOT.md to remove references to the Run*Bundle targets which never worked as described and I have removed.

fixes #3013

Test Plan: ran manually, waiting to see if it works in CI.


nickgg:
fixed portability of uint64_t printfs in bundle example code

nickgg:
We download what we're using for the bundles, not 100% sure why we need all models. My guess was because we ran run.sh but looks like we don't so maybe we can remove the model download?

PR #3030, opened 2019-05-31
[CI] only download resnet50 and en2gr for expensive tests

nickgg:
Summary: We're downloading all models for RELEASE_WITH_EXPENSIVE_TESTS CI build, but looks like we only use resnet and en2gr text translation models.

Documentation:

Test Plan: try to make CI work


nickgg:
Unfortunately looks like en2gr is the biggest model by a lot so unsure how much we're saving here.

PR #3032, opened 2019-05-31
[compilationContext]:Drop the const modifier for CompilationContext

ZchiPitt:
## Summary:
*Drop the const modifier for CompilationContext. Because I need to add a member variable that can be modified later.*

I propose to add a member variable (LogContext) into CompilationContext class in next PR.
The LogContext will be able to:
- Keep track the log scope (Together with another to-be-implemented class ScopedLogBlock)
- Store all the log strings
- Provide method to dump all the log strings

The coupling between LogContext and CompilationContext will make sure that:
- Logging will not be affected by multiple compilations that happen alternatively.
- Have a private stack for each LogContext instead of one static stack for the entire run of Glow, such that it's thread safe.

### DIFF
--- a/docs/Backends.md
+++ b/docs/Backends.md
@@ -52,7 +52,7 @@ are two pure virtual functions all backends must implement:
 
 Additionally, there are virtual functions that backends can override:
 
-- `virtual bool transformPostLowering(Function *F, const CompilationContext &cctx) const;`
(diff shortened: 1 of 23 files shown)


nickgg:
Having discussed with @ZchiPitt I think it makes sense to store the log in the CompilationContext, similar to how we store trace events in the ExecutionContext.

PR #3034, opened 2019-06-01
Return Expected from Backend::compile

jackm321:
Summary:

Documentation:

Test Plan:
CI
Test on H

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -52,23 +52,29 @@ public:
   /// Generate code for a vector of functions, \p functions. All compilations
   /// use the same settings provided by \p opts. This allows the compiler to
   /// support shared constants between functions.
-  virtual std::vector<std::unique_ptr<CompiledFunction>>
+  virtual llvm::Expected<std::vector<std::unique_ptr<CompiledFunction>>>
   compileFunctions(llvm::ArrayRef<Function *> functions,
                    BackendOptions &opts) const {
     std::vector<std::unique_ptr<CompiledFunction>> compiledFunctions;
     for (auto &function : functions) {
-      compiledFunctions.push_back(compile(function, opts));
+      if (auto resOrErr = compile(function, opts)) {
+        compiledFunctions.push_back(std::move(*resOrErr));
+      } else {
+        return resOrErr.takeError();
+      }
     }
-    return compiledFunctions;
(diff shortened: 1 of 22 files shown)


opti-mix:
@jackm321 Would it make sense to introduce a type alias for `llvm::Expected<std::unique_ptr<CompiledFunction>>` as it seems to be too long? E.g. `ExpectedCompiledFunction` or `ExpectedUniqueCompiledFunction`.

jackm321:
@opti-mix yeah idk I'm kind of indifferent I guess.
It would make the type name more brief but also obfuscate what's actually happening and each piece kind of matters. Also if we make
`using ExpectedCompiledFunction = llvm::Expected<std::unique_ptr<CompiledFunction>>;` this won't help with related types like `llvm::Expected<std::vector<std::unique_ptr<CompiledFunction>>>` for that we could create another type like `ExpectedCompiledFunctionVector` or we could instead just abstract the common `std::unique_ptr<CompiledFunction>` but this doesn't seem like it would be very helpful with reducing the verbosity. My vote is to just keep it as is but if you or anyone feel we should alias it I'm fine with it.

jackm321:
I'm going to split out the deeper habana changes in this PR to another PR

opti-mix:
@jackm321 

> My vote is to just keep it as is but if you or anyone feel we should alias it I'm fine with it.

OK. Let's keep it as is.

include/glow/Backend/Backend.h
```diff
@@ -52,23 +52,29 @@ class Backend {
   /// Generate code for a vector of functions, \p functions. All compilations
   /// use the same settings provided by \p opts. This allows the compiler to
   /// support shared constants between functions.
-  virtual std::vector<std::unique_ptr<CompiledFunction>>
+  virtual llvm::Expected<std::vector<std::unique_ptr<CompiledFunction>>>
```
jsubag: Not sure if it's worth the added error handling for the caller function, but you can allow partial success here by changing the return type to:
_std::vector<llvm::Expected_<std::unique_ptr<CompiledFunction>>>

jackm321: Yeah but imo if any function fails to compile then the whole thing should fail

jsubag: Fair enough :)

jackm321:
@opti-mix any additional comments here?

PR #3035, opened 2019-06-01
[ParameterSweepTest] Add FC sweep test

jfix71:


jfix71:
@beicy Would you mind reviewing the second commit I just added here? I needed to refactor things somewhat significantly into individual gtests, so that each gtest didn't take so long. This will improve testing speed as well. There's nothing new tested via this new commit, just reorganizing into multiple tests.

PR #3036, opened 2019-06-03
Updated Device Managers to support unpaddedSize field on Tensor.

gcatron:
Summary: This PR updates the current Device Managers to use unpaddedSize on the Tensor when copying to the device.

Documentation: NA

Test Plan: ninja test passes. Additional testing will follow when onnxifi is updated.

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -482,7 +482,7 @@ llvm::Error HabanaFunction::execute(ExecutionContext *context) {
     EnqueueTensorInfo eti;
     llvm::StringRef name = P->getName();
     eti.tensorName = name.data();
-    eti.tensorSize = T->getSizeInBytes();
+    eti.tensorSize = T->getUnpaddedSizeInBytes();
     eti.pTensorData = (char *)ioBuffer->get(P);
 
     copyInputsSizeInBytes += eti.tensorSize;
@@ -504,7 +504,7 @@ llvm::Error HabanaFunction::execute(ExecutionContext *context) {
     EnqueueTensorInfo eti;
     llvm::StringRef name = P->getName();
     eti.tensorName = name.data();
-    eti.tensorSize = T->getSizeInBytes();
+    eti.tensorSize = T->getUnpaddedSizeInBytes();
     eti.pTensorData = (char *)ioBuffer->get(P);
 
     outputInfo.push_back(eti);
--- a/lib/Backends/Interpreter/InterpreterFunction.cpp
+++ b/lib/Backends/Interpreter/InterpreterFunction.cpp
@@ -182,6 +182,23 @@ llvm::Error BoundInterpreterFunction::execute(IRFunction *F,
                                               ExecutionContext *context) {
   {
     auto ev = context->scopedEvent("registerTensors");
+
+    // Find all virtually padded tensors so they can be replaced.
+    std::vector<Placeholder *> virtualPadded;
+    for (auto &ph : context->getPlaceholderBindings()->pairs()) {
+      if (ph.second->getUnpaddedSizeInBytes() < ph.second->getSizeInBytes()) {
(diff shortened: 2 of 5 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -504,7 +504,7 @@ llvm::Error HabanaFunction::execute(ExecutionContext *context) {
     EnqueueTensorInfo eti;
     llvm::StringRef name = P->getName();
     eti.tensorName = name.data();
-    eti.tensorSize = T->getSizeInBytes();
+    eti.tensorSize = T->getUnpaddedSizeInBytes();
```
rdzhabarov: have you tested this :)?

gcatron: not yet, but since there are no virtually padded Tensors yet, unpaddedSizeInBytes = SizeInBytes.

nickgg: Can you test this in this PR via unit tests?

gcatron: good point, I'll add a unit test.

gcatron:
zero() is called in the Tensor creation so it's already initialized to zero.

nickgg:
Interesting, do we want to Zero all tensors by default? Seems wasteful if we're going to fill them with data anyway. Not something to worry for this PR anyway I guess.

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -182,6 +182,23 @@ llvm::Error BoundInterpreterFunction::execute(IRFunction *F,
...
+      }
+    }
+    // Replace all virtually padded tensors with real padding tensors.
+    for (auto &ph : virtualPadded) {
+      auto oldTensor = context->getPlaceholderBindings()->get(ph);
+      auto paddedTensor = new Tensor(oldTensor->getType());
+      memcpy(paddedTensor->getUnsafePtr(), oldTensor->getUnsafePtr(),
+             oldTensor->getUnpaddedSizeInBytes());
+      context->getPlaceholderBindings()->erase(ph);
+      context->getPlaceholderBindings()->insert(ph, paddedTensor);
```
rdzhabarov: can you use `void insert(Placeholder *P, Tensor &&T);` version here?

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -182,6 +182,23 @@ llvm::Error BoundInterpreterFunction::execute(IRFunction *F,
...
+    std::vector<Placeholder *> virtualPadded;
+    for (auto &ph : context->getPlaceholderBindings()->pairs()) {
+      if (ph.second->getUnpaddedSizeInBytes() < ph.second->size()) {
+        virtualPadded.push_back(ph.first);
+      }
+    }
+    // Replace all virtually padded tensors with real padding tensors.
+    for (auto &ph : virtualPadded) {
+      auto oldTensor = context->getPlaceholderBindings()->get(ph);
+      auto paddedTensor = new Tensor(oldTensor->getType());
```
rdzhabarov: as long as possible I'd like to avoid naked `new`.

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -141,6 +141,75 @@ TEST_P(DeviceManagerTest, Basic) {
...
+                     });
+
+  future.wait_for(std::chrono::seconds(2));
+  EXPECT_EQ(future.get(), module.get());
+
+  std::unique_ptr<ExecutionContext> context =
+      llvm::make_unique<ExecutionContext>();
+  context->getPlaceholderBindings()->allocate(output);
+
+  Tensor *input1 = new Tensor(ElemKind::FloatTy, {1});
```
rdzhabarov: nit: Tensor input;

PR #3039, opened 2019-06-04
Adding SLS node instead of reducing to SLWS

mortzur:
Summary:
Currently SLS is implemented by creating a splat node for constant 1.0  weights and reducing to SLWS. Better performance could be achieved by avoiding the unnecessary weights.
work-in-progress on this stack:
- add Habana impl
- add OpenCL impl
- do the same for the fused operator (SparseLengthsSumFused8BitRowwise) 

update: OpenCL was not implemented, lowering to SLWS was enables instead.  SparseLengthsSumFused8BitRowwise will be added on a separate PR.

Test Plan:

[Optional Fixes #issue]

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -2312,6 +2312,103 @@ void BoundInterpreterFunction::fwdLengthsSumInst(const LengthsSumInst *I) {
                             I->getData()->getElementType(), I)
 }
 
+void BoundInterpreterFunction::fwdSparseLengthsSumInstI8Impl(
+    const SparseLengthsSumInst *I) {
+
+  auto out = getTensor(I->getDest());
+  auto data = getTensor(I->getData());
+  auto indices = getTensor(I->getIndices());
+  auto lengths = getTensor(I->getLengths());
+
+  out->zero();
+
+  auto IH = indices->getHandle<int64_t>();
+  auto LH = lengths->getHandle<int32_t>();
+
+  size_t segments = lengths->dims()[0];
+  size_t totalLength = 0;
+  for (size_t i = 0; i < segments; i++) {
+    totalLength += LH.raw(i);
+  }
+  assert(totalLength <= indices->dims()[0] &&
+         "sum(Lengths) must be equal to len(Indices)");
+
+  size_t lineSize = data->size() / data->dims()[0];
+
+  auto DH = data->getHandle<int8_t>();
+  auto OH = out->getHandle<int8_t>();
+
+  auto TQP = [](Tensor *T) {
+    return TensorQuantizationParams{T->getType().getScale(),
+                                    T->getType().getOffset()};
+  };
+  using namespace quantization;
+
+  size_t curIdx = 0;
+  for (size_t i = 0; i < segments; i++) {
+    std::vector<float> accum(lineSize, 0.0f);
+    for (int32_t j = 0; j < LH.raw(i); j++) {
+      size_t offsetIn = IH.raw(curIdx) * lineSize;
+      for (size_t k = 0; k < lineSize; k++) {
+        accum[k] += dequantize(DH.raw(offsetIn++), TQP(data));
+      }
+      curIdx++;
+    }
+    size_t offsetOut = i * lineSize;
+    for (size_t k = 0; k < lineSize; k++) {
+      OH.raw(offsetOut++) = quantize(accum[k], TQP(out));
+    }
+  }
+}
+
+template <typename ElemTy>
+void BoundInterpreterFunction::fwdSparseLengthsSumInstFloatImpl(
+    const SparseLengthsSumInst *I) {
+  staticAssertFloatingPointType(ElemTy);
+
+  auto out = getTensor(I->getDest());
+  auto data = getTensor(I->getData());
+  auto indices = getTensor(I->getIndices());
+  auto lengths = getTensor(I->getLengths());
+
+  out->zero();
+
+  auto IH = indices->getHandle<int64_t>();
+  auto LH = lengths->getHandle<int32_t>();
+
+  size_t segments = lengths->dims()[0];
+  size_t totalLength = 0;
+  for (size_t i = 0; i < segments; i++) {
+    totalLength += LH.raw(i);
+  }
+  assert(totalLength <= indices->dims()[0] &&
+         "sum(Lengths) must be equal to len(Indices)");
+
+  size_t lineSize = data->size() / data->dims()[0];
+
+  auto DH = data->getHandle<ElemTy>();
+  auto OH = out->getHandle<ElemTy>();
+
+  size_t curIdx = 0;
+  for (size_t i = 0; i < segments; i++) {
+    for (size_t j = 0, e = LH.raw(i); j < e; j++) {
+      size_t offsetIn = IH.raw(curIdx++) * lineSize;
+      size_t offsetOut = i * lineSize;
+      for (size_t k = 0; k < lineSize; k++)
+        OH.raw(offsetOut++) += DH.raw(offsetIn++);
+    }
+  }
+}
+
+void BoundInterpreterFunction::fwdSparseLengthsSumInst(
+    const SparseLengthsSumInst *I) {
+  if (I->getDest()->getType()->isQuantizedType()) {
+    return fwdSparseLengthsSumInstI8Impl(I);
+  }
+  dispatchFloatingPointImpl(fwdSparseLengthsSumInstFloatImpl,
+                            I->getData()->getElementType(), I);
+}
+
 template <typename ElemTy>
 void BoundInterpreterFunction::fwdSparseLengthsWeightedSumInstFloatImpl(
     const SparseLengthsWeightedSumInst *I) {
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -2342,7 +2342,6 @@ void BoundInterpreterFunction::fwdSparseLengthsSumInstI8Impl(
     return TensorQuantizationParams{T->getType().getScale(),
                                     T->getType().getOffset()};
   };
-  using namespace quantization;
 
   size_t curIdx = 0;
   for (size_t i = 0; i < segments; i++) {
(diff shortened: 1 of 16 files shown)


rdzhabarov:
I agree with @jfix71 on not having explicit SLS node on graph level.

Note, for Habana we need to do this checking/matching before IR gen (we do not use IR gen for it).
For now I'd focus on optimizing Habana and then we could add general support as @jfix71 suggested.

bertmaher:
IMO it is simpler to add an SLS node than it is to try to pattern match it away in the IR (and also in the backends that don’t use IR)

jfix71:
> IMO it is simpler to add an SLS node than it is to try to pattern match it away in the IR (and also in the backends that don’t use IR)

I think the pattern matching logic is pretty small/simple here. Instead of adding an SLS case in a switch that mostly duplicates the SLWS logic, you'd add something like this to the SLWS case:

```cpp
std::string kernel;
SplatNode *SN = dyn_cast<SplatNode>(SLWS->getWeights());
if (SN && SN->getValue() == 1.0) {
  kernel = "sparse_lengths_sum_f32";
} else {
  kernel = "sparse_lengths_weighted_sum_f32";
  inputs.push_back(tensors[RI->getWeights()].get());
}
```

bertmaher:
Ok, sure, we can do the pattern match instead.

bertmaher:
Just so I don’t feel like an idiot, I’ve got to explain why I think it’s “simpler” to add the new node- it ends up adding a fair amount of boilerplate to add a new node, which I had forgotten about, but the boilerplate is all extremely regular. So there’s duplicated code but it’s identical to the surrounding code, so it’s pretty easy to write and reason about.

The custom optimization code is short, and I guess I prefer doing it here in the Habana backend just for the sake of diff size, but it’s a custom thing that has to be reasoned about (eg, will the Habana backend DCE away the splat?  It turns out it will, so this particular thing isn’t really an issue, but it could have been). 

I also have a larger issue with our “lowering” nodes that are present in the source only to then “raise” the generated pattern back to the original input.  But this isn’t the only place we do that, so, may as well not fight it.

opti-mix:
> I also have a larger issue with our “lowering” nodes that are present in the source only to then “raise” the generated pattern back to the original input. But this isn’t the only place we do that, so, may as well not fight it.

This is not the only place :-( I was coincidentally working on the same optimization for one of the backends. IMHO, it is not Habana-specific. Some other backends may benefit as well from being able to distinguish between SLWS and SLS. And as far as I can see some of them are graph IR based and some others are low-level IR based.

jfix71:
>will the Habana backend DCE away the splat? It turns out it will, so this particular thing isn’t really an issue, but it could have been

I don't follow here -- aren't you essentially lowering the graph to Habana IR? Why would Habana IR know about the Splat if you elide it when lowering?

>>I also have a larger issue with our “lowering” nodes that are present in the source only to then “raise” the generated pattern back to the original input. But this isn’t the only place we do that, so, may as well not fight it.

>This is not the only place :-( I was coincidentally working on the same optimization for one of the backends. IMHO, it is not Habana-specific. Some other backends may benefit as well from being able to distinguish between SLWS and SLS. And as far as I can see some of them are graph IR based and some others are low-level IR based.

Hmm, so I think we generally do the opposite, raising things as much as possible at the beginning (e.g. via `fold()`), then gradually lowering only as the backend desires. What other examples are you guys thinking of?

In general if it was at all difficult/fragile to pattern match the SLS from the SLWS then I'd understand the issue more, but it's super simple here. I agree this is definitely not Habana specific -- all backends would want to do this -- it's more a question of how to represent it in the IR. This same pattern matching logic would be used everywhere (i.e. check if the Weights input is a splat with Value 1.0). We could even make this a helper function if it'd make it easier.

nadavrot:
I am really enjoying reading this thread. There are always multiple ways to do things and it's great to have these technical discussions that help us to come up with a great design. But even beyond the scope of this specific PR, I love learning from discussion threads like this about different design tradeoffs.

mortzur:
@jfix71  - On one hand I would love to avoid changing the high level returned node type, because it would allow me to insert the modifications in a more modular way, without needing to change anything for backends which exist and work fine with SLWS . The motivation here is on Habana.

One the other hand, returning SLWS node which hides an implementation as SLS (if I understood right) sounds confusing for readability and debugging.

bertmaher:
> > will the Habana backend DCE away the splat? It turns out it will, so this particular thing isn’t really an issue, but it could have been
> 
> I don't follow here -- aren't you essentially lowering the graph to Habana IR? Why would Habana IR know about the Splat if you elide it when lowering?
> 

So the way the Habana codegen works is to iterate all the nodes, and call `synCreateGenericNode` to produce the corresponding Habana IR.  So if there's a splat node in the graph, we'll create a Habana node for it, even if it's not used by anything.  (I think the Habana GC is smart enough to DCE it later).   (Hopefully I'm not wrong about this description :) ).

Maybe there's some alternate design that would avoid visiting the splat at all?

opti-mix:
> So if there's a splat node in the graph, we'll create a Habana node for it, even if it's not used by anything. (I think the Habana GC is smart enough to DCE it later). (Hopefully I'm not wrong about this description :) ).

I totally agree that this is an issue. I see a similar problem with the IR-based backend I'm working on. I do the simple conditional check that @jfix71 proposed, but it is too late already, because the code for Splat was generated already for the weights operand of the SLWS instruction and this memory and compute to initialize it by the splat value is just wasted. Of course, I could have introduced a backend-specific SLS instruction and custom IRGen. Or a backend-specific SLS node and instruction. But it seems like we already have 2 or 3 backends that may need to do it or to use some other custom logic to understand that SLS could be used instead of SLWS. So, it seems like avoiding having a dedicated SLS node and instruction does not buy us much in terms of saving the development effort.

Another thing that we briefly discussed with @bertmaher and @mortzur is if backends should just say in `isOpSupported` if they want SLS to be lowered (into SLWS) or not. By default it should be lowered, i.e. we preserve the today's status quo. But backends could override this.

SplitInfinity:
> Better performance could be achieved by avoiding the unnecessary weights.

Could you share your measurements?

jfix71:
@bertmaher @opti-mix Ah, got it -- yeah I imagine you could walk the graph starting from outputs and then recursively generate IR for your inputs, so that when you generate an SLS from an SLWS you wouldn't generate the weights. But given the extra complexity that would bring I think it totally makes sense to just add the SLS node 🙂

>One the other hand, returning SLWS node which hides an implementation as SLS (if I understood right) sounds confusing for readability and debugging.

@mortzur Yeah I agree it is a little unexpected. However we do this already in many places in `Graph.cpp`, where we return some Node that implements some op even though we don't have a class for that op. E.g. see `createBroadcast()`, `createDotProduct()`, `createClip()`, `createLogit()`, etc. We do this for a similar reason to this PR -- to avoid adding a new Node to NodeGen. Though I agree this case is slightly different given that we want to IRGen it to something else.

artemrakhov:
Discussed offline. I still think that if we add unweighted versions of everything, we will end up with 2x more nodes:
* SparseLengthsWeightedSum
* RowwiseQuantizedSparseLengthsWeightedSum
* FusedRowwiseQuantizedSparseLengthsWeightedSum
* SparseLengthsSum
* RowwiseQuantizedSparseLengthsSum
* FusedRowwiseQuantizedSparseLengthsSum

bertmaher:
> Discussed offline. I still think that if we add unweighted versions of everything, we will end up with 2x more nodes:
> Which we probably don't want. It's a slippery slope, this is how we can quickly reach 500 ops.

Here's my take: I don't think we should focus on the number of operators for its own sake.  What we care about is performance of ML workloads.  In some cases having fewer ops helps you get performance, because you can find fusions or other "global" optimization opportunities that would be hard to find amongst a bunch of complicated operators.  (This is kind of like RISC vs CISC.)

If we had a nice fusion pass that would take in a graph containing SLWS+Splat(1.0) and produce code equivalent to SLS (and also DCE the splat), that would be preferable to adding an explicit SLS.  But we don't have that yet (and it's still an open question of whether we can get there since some backends don't make it easy to automatically create new fused kernels on the fly).

In the meantime we could use the performance boost today :-)

nrsatish:
Hi folks - it is becoming very difficult to partition SLWS ops which share the same splatted weights. Can we make this such that there is no "weights" input for the pure SLS ops?

bertmaher:
> Hi folks - it is becoming very difficult to partition due to these shared Splat nodes among all the ops. This PR would unblock partitioning progress.

This is a pretty good argument for having the new node.  An alternative we talked about was introducing a HabanaSparseLengthsSum and specializing it for that platform, but, given this PR already exists, I'm going to go ahead and review it as-is.  Code is mutable, we can always take a different approach later :-)

bertmaher:
Ok, so per offline convo we're going to introduce the notion of an optional input to nodes, which we'll use to magic away the weights input of SLWS.  So I'll wait for that instead.

lib/Graph/Nodes.cpp
```diff
@@ -376,6 +376,20 @@ static bool verifyRegression(NodeValue src, NodeValue dest,
          checkSameType(dest, expected, dest.getNode());
 }
 
+static bool verifySparseLengthsdSum(NodeValue dest, NodeValue data,
```
bertmaher: sp: `Lengthsd` -> `Lengths`

lib/Graph/Nodes.cpp
```diff
@@ -920,6 +934,11 @@ bool BatchedReduceMeanNode::verify() const {
   return isValid;
 }
 
+bool SparseLengthsSumNode::verify() const {
+  return verifySparseLengthsdSum(getResult(), getData(), getIndices(),
```
bertmaher: same spelling comment

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -2312,6 +2312,103 @@ void BoundInterpreterFunction::fwdLengthsSumInst(const LengthsSumInst *I) {
...
+  size_t lineSize = data->size() / data->dims()[0];
+
+  auto DH = data->getHandle<int8_t>();
+  auto OH = out->getHandle<int8_t>();
+
+  auto TQP = [](Tensor *T) {
+    return TensorQuantizationParams{T->getType().getScale(),
+                                    T->getType().getOffset()};
+  };
+  using namespace quantization;
```
rdzhabarov: this is a bit weird. I'd just use quantization::dequantize and quantization::quantize.

lib/Optimizer/Lower.cpp
```diff
@@ -794,6 +794,18 @@ static void lowerBatchMatMulNode(Function *F, CompilationContext &cctx,
   replaceAllUsesOfWith(cctx.loweredInfoMap, BMMN.getResult(), RN);
 }
 
+static void lowerSparseLengthsSumNode(Function *F, CompilationContext &cctx,
```
rdzhabarov: what's the reason for this given you added implementation of SLS to interpreter and CPU backend?

mortzur: It is meant for other backends that does not have SLS implementation, i.e. OpenCL.
I should have thought about it before implementing for interpreter and CPU.

rdzhabarov:
Please, check that internally this works on representative models before merging.

artemrakhov:
Just for the records, have we decided not to proceed with SLSNode? Closed without landing?

bertmaher:
No, this landed, the UI is just a bit confusing.  I don't think we really came to a decisive agreement about the Right Way forward, but this helps unblock some perf work in the short term.

PR #3040, opened 2019-06-04
[compilationContext][lower]: Change the second parameter of glow::lower() from LoweredInfoMap to CompilationContext

ZchiPitt:
Summary:
I modified the parameter of glow::lower() to take compilation context instead of loweredInfoMap.
- We need to be able to access compilation context through the glow::lower() method to all the lowering methods that lower() calls, where we want to produce the log using the functionality of CompilationContext::LogContext.

Test Plan:
Test with all Glow tests

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -53,12 +53,12 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
     return ONNXIFI_STATUS_INTERNAL_ERROR;
   }
 
-  glow::lower(function, /* loweredMap */ nullptr, glowBackend_.get());
+  CompilationContext cctx;
+  cctx.compMode = CompilationMode::Infer;
+  glow::lower(function, cctx, glowBackend_.get());
 
   // Call the backend's transformPostLowering to match the normal compilation
   // pipeline then DCE any nodes that are no longer needed.
-  CompilationContext cctx;
-  cctx.compMode = CompilationMode::Infer;
   if (glowBackend_->transformPostLowering(function, cctx)) {
     glow::DCE(function);
   }
--- a/lib/Optimizer/Lower.cpp
+++ b/lib/Optimizer/Lower.cpp
@@ -44,17 +44,19 @@ static void replaceAllUsesOfWith(LoweredInfoMap *loweredMap, NodeValue oldNV,
                       oldNV.getNode()->getKind()));
 }
 
-static void lowerAddGradNode(Function *F, LoweredInfoMap *loweredMap,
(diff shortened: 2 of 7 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -53,11 +53,11 @@ onnxStatus BackendId::checkGraphCompatibility(const void *onnxModel,
...
   }
 
-  glow::lower(function, /* loweredMap */ nullptr, glowBackend_.get());
+  CompilationContext cctx;
+  glow::lower(function, cctx, glowBackend_.get());
 
   // Call the backend's transformPostLowering to match the normal compilation
   // pipeline then DCE any nodes that are no longer needed.
-  CompilationContext cctx;
   cctx.compMode = CompilationMode::Infer;
```
nickgg: I would move this line setting the compilation mode up to where you moved creating the CompilationContext to. I don't think it's valid to have a context without the mode set.

lib/Optimizer/Lower.cpp
```diff
@@ -44,17 +44,19 @@ static void replaceAllUsesOfWith(LoweredInfoMap *loweredMap, NodeValue oldNV,
...
-static void lowerAddGradNode(Function *F, LoweredInfoMap *loweredMap,
+static void lowerAddGradNode(Function *F, CompilationContext &cctx,
                              const AddGradNode &node) {
   /// The chain rule for addition:
   /// delta(LHS) = dF/dLHS * delta(OUT) = 1 * delta(OUT)
   /// delta(RHS) = dF/dRHS * delta(OUT) = 1 * delta(OUT)
   auto outG = node.getGradOfOriginalOutputNamedResult();
-  replaceAllUsesOfWith(loweredMap, node.getGradOfInputNamedLHS(), outG);
-  replaceAllUsesOfWith(loweredMap, node.getGradOfInputNamedRHS(), outG);
+  replaceAllUsesOfWith(cctx.loweredInfoMap, node.getGradOfInputNamedLHS(),
```
nickgg: Before landing this I think you should think hard about whether or not you want to log replacement in each place it occurs, or whether you can log one time in `replaceAllUsesOfWith`. If you can do the second then we'll want to pass the context into that method.

lib/Onnxifi/Base.cpp
ZchiPitt: done. Btw, i found that by default CompilationContext set its mode as Infer, so probably we can also remove this line here

lib/Optimizer/Lower.cpp
ZchiPitt: Great idea. Working on intercepting just that method right now.

PR #3041, opened 2019-06-04
Remove most instances GLOW_UNREACHABLE

jackm321:
Summary:
Remove all of the easy to remove `GLOW_UNREACHABLE`. The remaining instances of `GLOW_UNREACHABLE` should be handled using llvm::Errors but aren't yet in a context where an llvm::Error can be returned

Documentation:
none

Test Plan:
CI

### DIFF
--- a/lib/Backend/Backend.cpp
+++ b/lib/Backend/Backend.cpp
@@ -40,7 +40,7 @@ TraceInfo Backend::buildManualTraceInfo(Function *F) const {
 
 void Backend::autoInstrument(TraceInfo &traceInfo, IRFunction *IR) const {
   if (getTraceEventDataSize() == 0) {
-    GLOW_UNREACHABLE("Auto instrumentation not supported on this backend");
+    LOG(ERROR) << "Auto instrumentation not supported on this backend";
     return;
   }
 
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1210,7 +1210,7 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
       RETURN_IF_ERR(
           fillTensor<int64_t>(T, ElemKind::Int64ITy, dim, values->ints()));
     } else {
-      GLOW_UNREACHABLE("Unhandled GivenTensorFill type");
+      RETURN_ERR(strFormat("Unhandled tensor fill type: %s", typeName.c_str()));
     }
     RETURN_IF_ERR(createAndRegisterConstant(op.output().Get(0), std::move(T)));
     return llvm::Error::success();
--- a/lib/LLVMIRCodeGen/LLVMIRGen.cpp
+++ b/lib/LLVMIRCodeGen/LLVMIRGen.cpp
@@ -533,7 +533,8 @@ llvm::Function *LLVMIRGen::getFunction(const std::string &name,
   case ElemKind::BoolTy:
     return get("libjit_" + name + "_b");
   default:
-    GLOW_UNREACHABLE("Unsupported element type");
+    LOG(FATAL) << "Unsupported element type: "
+               << Type::getElementName(elemTy).str();
   }
 }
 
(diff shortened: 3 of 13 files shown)


lib/LLVMIRCodeGen/LLVMIRGen.cpp
```diff
@@ -2382,12 +2382,10 @@ void LLVMIRGen::generateLLVMIRForInstr(llvm::IRBuilder<> &builder,
...
-#ifndef NDEBUG
-    llvm::errs() << "Cannot select the instruction:\n";
-    I->dump(llvm::errs());
-    llvm::errs() << "\n";
-#endif
-    GLOW_UNREACHABLE("ERROR: Cannot select the instruction.");
+    std::string sBuf;
+    llvm::raw_string_ostream s(sBuf);
+    I->dump(s);
+    LOG(DFATAL) << "Cannot select the instruction: " << s.str();
```
jfix71: Correct me if I'm wrong -- this isn't equivalent to the previous code, is it? Before only the printing was wrapped in NDEBUG, but it would always abort even if release. Does this also abort in release?

lib/LLVMIRCodeGen/LLVMIRGen.cpp
```diff
@@ -1325,12 +1326,10 @@ void LLVMIRGen::generateLLVMIRForDataParallelInstr(
...
```
jfix71: Same question as other.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1210,7 +1210,7 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
       RETURN_IF_ERR(
           fillTensor<int64_t>(T, ElemKind::Int64ITy, dim, values->ints()));
     } else {
-      GLOW_UNREACHABLE("Unhandled GivenTensorFill type");
+      RETURN_ERR("Unhandled GivenTensorFill type");
```
jfix71: nit: Possible to add the `typeName` here or no?

lib/Backend/Backend.cpp
```diff
@@ -40,7 +40,7 @@ TraceInfo Backend::buildManualTraceInfo(Function *F) const {
 
 void Backend::autoInstrument(TraceInfo &traceInfo, IRFunction *IR) const {
   if (getTraceEventDataSize() == 0) {
-    GLOW_UNREACHABLE("Auto instrumentation not supported on this backend");
+    LOG(ERROR) << "Auto instrumentation not supported on this backend";
     return;
```
jfix71: I see this return was here before, but is it still necessary?

lib/LLVMIRCodeGen/LLVMIRGen.cpp
jackm321: Yes it should be the same. DFATAL == ERROR if NDEBUG and FATAL otherwise

lib/LLVMIRCodeGen/LLVMIRGen.cpp
jackm321: Same answer

lib/Backend/Backend.cpp
jackm321: Will remove, also this should have been LOG(FATAL) oops

jackm321: Actually wait no, this is right. If autoInstrument isn't supported, we should just log that and keep going

jfix71: Ok, but that's different than previous behavior right? `GLOW_UNREACHABLE()` would always abort.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
jfix71: But doesn't `GLOW_UNREACHABLE()` abort even in release? This change means in release it instead will log an error but proceed with execution, no?

lib/Backend/Backend.cpp
jackm321: Yes this is different behavior we probably shouldn't kill the process due to this.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
jackm321: Oh yeah you're right. I misread the original code and thought there was an #else. I'll just LOG(FATAL) then.

PR #3042, opened 2019-06-04
[PlaceholderBindings] Add DCHECK to insert to check that types match

SplitInfinity:
**Description**
This commit adds a debug-mode check to `PlaceholderBindings::insert`
that checks that the types of the `Placeholder` and `Tensor` passed to
the method match. It is currently possible to bind a `Placeholder`
to any `Tensor` and that seems like a recipe for disaster. It will still
be possible after this commit, but will crash if the project is built in
Debug mode.

**Test Plan**
All unit tests pass.

### DIFF
--- a/lib/Graph/PlaceholderBindings.cpp
+++ b/lib/Graph/PlaceholderBindings.cpp
@@ -78,6 +78,10 @@ PlaceholderBindings::getPlaceholderByName(llvm::StringRef name) const {
 }
 
 void PlaceholderBindings::insert(Placeholder *P, Tensor &&T) {
+  DCHECK(T.getType().isEqual(*P->getType()))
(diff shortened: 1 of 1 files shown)


lib/Graph/PlaceholderBindings.cpp
```diff
@@ -78,6 +78,10 @@ PlaceholderBindings::getPlaceholderByName(llvm::StringRef name) const {
 }
 
 void PlaceholderBindings::insert(Placeholder *P, Tensor &&T) {
+  DCHECK(T.getType().isEqual(*P->getType()))
+      << "Placeholder " << P->getName().str() << " has type "
+      << P->getType()->toString() << " but Tensor has type "
+      << T.getType().toString() << "\n";
```
jackm321: don't need "\n" at the end

PR #3043, opened 2019-06-04
Loading ConstantOfShape ONNX operator as Glow Splat operator

putivsky:
Summary: Glow loaders load ConstantOfShape operator supported by Caffe2 and ONNX as Glow Splat operator with restrictions supporying only float type.

Differential Revision: D15614377

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -1050,6 +1050,41 @@ ONNXModelLoader::loadSpaceToDepth(const ONNX_NAMESPACE::NodeProto &op,
   return llvm::Error::success();
 }
 
+llvm::Error
+ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
+                                     const ArgumentDictionaryTy &dict) {
+  Tensor T(ElemKind::FloatTy, {1});
+  T.getHandle().raw(0) = 0.0;
+
+  if (dict.count("value")) {
+    RETURN_IF_ERR(loadTensor(dict.at("value")->t(), &T));
+    RETURN_ERR_IF_NOT(T.dims().size() == 1, "Value must be a 1D vector.");
+    RETURN_ERR_IF_NOT(T.getType().getElementType() == ElemKind::FloatTy,
+                      "Value must have float type");
+  }
+
+  TypeRef ty;
+  if (op.input_size() > 0) {
+    Constant *in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getConstantByName(op.input(0)));
+    // Must be 1D tensor of int64_t.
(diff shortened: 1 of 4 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -959,6 +959,34 @@ class CommonOperatorLoader : public ProtobufLoader {
     return llvm::Error::success();
   }
 
+  llvm::Error loadConstantOfShape(const std::string &typeName, const OpType &op,
```
jackm321: Should this be in OnnxModelLoader instead?

putivsky: Caffe2 does have the similar operator ConstantOfShape.

putivsky:
> looks good but we should add a test

I have thought about it, I have ONNX 50MB model that has ConstantOfShape operator to test, what is the best way to have model added to the fleet of the models?

rdzhabarov:
@putivsky you can just add a test for parsing this exact op. Check unit tests for loaders (onnxImporterTest etc)

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1050,6 +1050,41 @@ ONNXModelLoader::loadSpaceToDepth(const ONNX_NAMESPACE::NodeProto &op,
...
 
+llvm::Error
+ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
+                                     const ArgumentDictionaryTy &dict) {
+  Tensor T(ElemKind::FloatTy, {1});
+  T.getHandle().raw(0) = 0.0;
+
+  if (dict.count("value")) {
+    RETURN_IF_ERR(loadTensor(dict.at("value")->t(), &T));
+    RETURN_ERR_IF_NOT(T.dims().size() == 1, "Value must be a 1D vector.");
```
jfix71: 1D vector, makes sense, but doesn't it also need to be a single element? Perhaps this check should be `T.dims() == {1}`?

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1273,6 +1273,24 @@ TEST(onnx, constant) {
...
+                          "tests/models/onnxModels/constantOfShape.onnxtxt");
+  Placeholder *output;
+  {
+    ONNXModelLoader onnxLD(netFilename, {}, {}, *F);
+    output = EXIT_ON_ERR(onnxLD.getSingleOutput());
+    EXPECT_NE(output, nullptr);
+  }
+  // ConstantOfShape -> Save -> PH
+  ASSERT_EQ(mod.getPlaceholders().size(), 1);
+  ASSERT_EQ(F->getNodes().size(), 2);
```
jfix71: Can you also check the Splat is there with the expected Value and Shape?

PR #3047, opened 2019-06-05
[Partitioner] Handle more than one outputs for a single node when calculating the memory usage.

### LINKED ISSUE #2885: [PartitionerUtils] Multi-output nodes not correctly accounted for inMemSize?
In the snippet below, we're using `getType(0)` to update the `inMemSize`. This smells like it will incorrectly account for a multi-output node when calculating `inMemSize`. 

E.g. if we partition such that a `TopK` node's `Indices` output (which is its second output) as feeding the input of another partition, then I believe `inMemSize` will call `getSizeInBytes()` on the `Values` output instead of `Indices`.

FWIW it looks like it's correctly handled in `getOutMemPerNode()`.

@beicy does this appear to be an issue? 

https://github.com/pytorch/glow/blob/f2e37eb88cc2378622dc601afcbe28790cdbf61d/lib/Partitioner/PartitionerUtils.cpp#L189-L197

beicy:
Summary:
This PR fixed the issue described in #2885.  

Documentation:

Fixes #2885

Test Plan:
Added unittest. Ninja test. ./tests/image/run.sh

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -228,8 +228,8 @@ void Partitioner::initOpComputeTime(Function *F) {
     }
 
     // Repeat for outputs
-    if (node.getNumResults() > 0) {
-      auto myty = node.getType(0);
+    for (size_t i = 0, e = node.getNumResults(); i < e; i++) {
+      auto myty = node.getType(i);
       uint64_t sizeOutput = myty->getSizeInBytes();
       if (sizeOutput > sramCapacity) {
         sizeDram += sizeOutput;
--- a/tests/unittests/PartitionerTest.cpp
+++ b/tests/unittests/PartitionerTest.cpp
@@ -529,3 +529,41 @@ TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
     mod_.clear();
   }
 }
+
+/// Check the function getGraphMemInfo to handle more than one outputs of a
+/// single Node in PartitionerUtils.cpp
+TEST_F(PartitionerTest, getGraphMemInfo) {
+  auto *inp1 =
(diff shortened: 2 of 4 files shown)


tests/unittests/PartitionerTest.cpp
```diff
@@ -529,3 +529,43 @@ TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
...
+  std::set<Node *> nodes1, nodes2;
+  nodes1.insert(R1);
+  nodes1.insert(R2);
+  nodes2.insert(CV);
+  nodes2.insert(CI);
+  nodes2.insert(saveValues);
+  nodes2.insert(saveIndices);
+  GraphMemInfo res1 = getGraphMemInfo(nodes1);
+  GraphMemInfo res2 = getGraphMemInfo(nodes2);
+  ASSERT_EQ(res1.constMemSize, 0);
```
rdzhabarov: expect_eq should be preferable here as you'll get evaluation on all members in case something fails.

lib/Partitioner/Partitioner.cpp
```diff
@@ -228,8 +228,8 @@ void Partitioner::initOpComputeTime(Function *F) {
     }
 
     // Repeat for outputs
-    if (node.getNumResults() > 0) {
-      auto myty = node.getType(0);
+    for (size_t i = 0, e = node.getNumResults(); i < e; i++) {
```
rdzhabarov: random question: why only SparseLengthsWeightedSumNodeKind on line 195 and not fused versions etc.

beicy: This was implemented by @nrsatish  and will be improved to handle more node type.

PR #3050, opened 2019-06-05
[Backends/1] Minimal setup for registering backend based on its name.

rdzhabarov:
Summary:
First step in backend registration based on the names.

* Register backend based on the string name
* Minimal changes to use createBackend(llvm::StringRef backendName) on a caller side. There will be gradual migration from the BackendKind to backendName version in following PRs.

Documentation:
n/a

Test Plan:
CI

cc: @gcatron

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -160,11 +160,13 @@ public:
 };
 
 /// Perform Backend Factory registration.
-#define REGISTER_GLOW_BACKEND_FACTORY(FactoryName, BackendClass, BackendName)  \
+#define REGISTER_GLOW_BACKEND_FACTORY(FactoryName, BackendClass)               \
   class FactoryName : public BaseFactory<std::string, Backend> {               \
   public:                                                                      \
     Backend *create() override { return new BackendClass(); }                  \
-    std::string getRegistrationKey() const override { return BackendName; }    \
+    std::string getRegistrationKey() const override {                          \
+      return BackendClass::getName();                                          \
+    }                                                                          \
   };                                                                           \
   static RegisterFactory<std::string, FactoryName, Backend>                    \
       FactoryName##_REGISTERED;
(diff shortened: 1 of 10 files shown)


include/glow/Backend/Backend.h
```diff
@@ -128,8 +143,12 @@ class Backend {
 };
 
 /// Create a backend of kind \p kind.
+/// Deprecated. Moving gradualy to the createBackend based on the backend name.
```
beicy: So we won't keep this interface in the future. Is that due to code complexity or others? I thought we would support both : from Backend Name and Backend Kind

rdzhabarov: BackendKind will go away. It's not practical to keep that enum up to date with auto registration of backends.

In the following PRs each backend will have static a method `getName()` which will return name of that backend and essentially will be used as a key everywhere. There will be no need for the enum BackendKind.

include/glow/Backend/Backend.h
```diff
@@ -141,16 +160,13 @@ class BackendUsingGlowIR : public Backend {
 };
 
 /// Perform Backend Factory registration.
-#define REGISTER_GLOW_BACKEND_FACTORY(FactoryName, BackendClass,               \
-                                      SpecificBackendKind)                     \
-  class FactoryName : public BaseFactory<BackendKind, Backend> {               \
+#define REGISTER_GLOW_BACKEND_FACTORY(FactoryName, BackendClass, BackendName)  \
```
opti-mix: Do we need to provide `BackendName` as an argument? Could we just ask `BackendClass::getBackendName()` for a name, assuming each backend would provide such a static function?

Or can we have use-cases, where e.g. the same backend is registered under custom or even multiple names?

rdzhabarov: Yup `BackendClass::getBackendName()` this makes sense. I mentioned in the other comment that I plan to do so. But maybe I just add it here.

> Or can we have use-cases, where e.g. the same backend is registered under custom or even multiple names?

Not sure if we have this use case, but it could be easily done if needed.

PR #3052, opened 2019-06-06
[compilationLog]: Implement the functionalities for logging glow compilation/optimization and add logging statements at the places of nodes creation, replacement and deletion.

ZchiPitt:
Summary:
Implement the functionalities for logging glow compilation/optimization and add logging statements at the places of nodes creation, replacement and deletion.

- Implemented the LogContext class
   - Keep track the log scope (Together with another to-be-implemented class ScopedLogBlock)
   - Store all the log strings
   - Provide method to dump all the log strings into a file
   - Will be initialized as a member variable inside Function class
- Added logging functions to record the:
   - nodes creation
   - nodes replacement 
   - nodes deletion 

Test Plan:
ninja check

test with GraphOptzTest::optimizeBatchNormAfterConvWithTransposedWeights,
and the output compile log looks like below:
Screen Shot 2019-06-05 at 11 31 15 PM

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -17,6 +17,7 @@
 #define GLOW_GRAPH_GRAPH_H
 
 #include "glow/Base/Type.h"
+#include "glow/Graph/Log.h"
 #include "glow/Graph/Nodes.h"
 #include "glow/Quantization/Base/Base.h"
 
@@ -222,11 +223,17 @@ class Function final : public Named {
   /// A reference to the owner of the function.
   Module *parent_;
 
+  /// The log context associated with this function.
+  LogContext logCtx_;
+
 public:
   Function(Module *parent, llvm::StringRef Name = {})
       : Named(Name), parent_(parent) {}
 
   ~Function();
+
+  /// Return the log context.
+  LogContext &getLogContext() { return logCtx_; }
 
   /// Add placeholder for metadata such as profiling.
   void addMetadataPlaceholder(Placeholder *PH) {
@@ -265,6 +272,10 @@ public:
   template <class NodeTy> NodeTy *addNode(NodeTy *N) {
     N->setName(Module::uniqueName(N->getName(), uniqueNodeNames_));
     nodes_.push_back(N);
+
+    // Log the node creation.
+    logCtx_.logNodeCreation(N);
+
     return N;
   }
 
--- a/include/glow/Graph/Log.h
+++ b/include/glow/Graph/Log.h
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef GLOW_SUPPORT_LOG_H
+#define GLOW_SUPPORT_LOG_H
+
+#include "llvm/ADT/StringRef.h"
+
+#include <string>
+#include <vector>
+
+namespace glow {
+
+class Node;
+
+/// A class for logging all compilation related activities.
+class LogContext {
+private:
+  /// A vector that keeps track of all log contents.
+  std::vector<std::string> logContents_;
+
+  /// A vector that keeps track of current compilation scopes.
+  std::vector<std::string> logScopes_;
+
+  /// A string that represents the current full scope name.
+  std::string currentFullScope_;
+
+public:
+  /// Add content into the contents vector.
+  void addLogContent(llvm::StringRef logContent);
+
+  /// Add a scope into the scope vector.
+  void pushLogScope(llvm::StringRef scopeName);
+
+  /// Getter that returns the current full scope name.
+  llvm::StringRef getFullScopeName() { return currentFullScope_; }
+
+  /// Pops out the most recently added scope.
+  void popLogScope();
+
+  /// Dumps the log into a file named after the given \p funcName.
+  void dumpLog(llvm::StringRef funcName);
+
+  /// Logs the node creation with a list of input nodes.
+  void logNodeCreation(const Node *newNode);
+
+  /// Logs the node replacement.
+  void logNodeReplacement(const Node *oldNode, const Node *newNode);
+
+  /// Logs the node deletion.
+  void logNodeDeletion(const Node &deletedNode);
+};
+
+/// Logs a new log scope.
+#define LOG_SCOPE(ctx, name) ScopedLogBlock __scope__(ctx, name);
+
+/// Helper class which traces the start and end of a compilation log scope.
+class ScopedLogBlock {
+
+  /// Reference to the log context.
+  LogContext &ctx_;
+
+  /// The name of the log scope.
+  std::string name_;
+
+  /// Whether this log scope has already ended, avoiding logging
+  /// it twice.
+  bool end_{false};
+
+public:
+  ScopedLogBlock(LogContext &ctx, llvm::StringRef name);
+  ~ScopedLogBlock();
+
+private:
+  /// Triggers the endding operation of the current log scope.
+  void end();
+};
+
+} // namespace glow.
+
+#endif // GLOW_SUPPORT_LOG_H
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -327,6 +327,7 @@ Function::~Function() {
     auto cur = it++;
     eraseNode(&*cur);
   }
+  logCtx_.dumpLog(getName());
 }
 
 TypeRef Module::uniqueType(ElemKind elemTy, llvm::ArrayRef<size_t> dims) {
@@ -2688,7 +2689,12 @@ void Module::eraseConstant(ConstList::iterator I) {
   constants_.erase(I);
 }
 
-void Function::eraseNode(NodesList::iterator I) { nodes_.erase(I); }
+void Function::eraseNode(NodesList::iterator I) {
+  // Log node deletion.
+  logCtx_.logNodeDeletion(*I);
+
+  nodes_.erase(I);
+}
 
 Constant *Module::getConstantByName(llvm::StringRef name) const {
   for (auto *V : getConstants()) {
--- a/lib/Graph/Log.cpp
+++ b/lib/Graph/Log.cpp
@@ -1,0 +1,148 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Graph/Log.h"
+#include "glow/Graph/Node.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace glow {
+
+static llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log", llvm::cl::init(false),
+                          llvm::cl::desc("Dump compilation log"));
+
+void LogContext::addLogContent(llvm::StringRef logContent) {
+  logContents_.push_back(logContent);
+}
+
+void LogContext::pushLogScope(llvm::StringRef scopeName) {
+  logScopes_.push_back(scopeName);
+  currentFullScope_ += "->";
+  currentFullScope_ += scopeName;
+}
+
+void LogContext::popLogScope() {
+  if (logScopes_.size() > 0) {
+    currentFullScope_ = currentFullScope_.substr(
+        0, currentFullScope_.size() - logScopes_.back().size() - 2);
+    logScopes_.pop_back();
+  }
+}
+
+void LogContext::dumpLog(llvm::StringRef funcName) {
+  if (dumpCompilationLogOpt) {
+    std::string compileLogFilename = funcName;
+    compileLogFilename.append("_compile.log");
+
+    llvm::outs() << "Writing compilation log file for Module to: "
+                 << compileLogFilename << '\n';
+    std::error_code EC;
+    llvm::raw_fd_ostream myfile(compileLogFilename, EC);
+    for (const auto &s : logContents_) {
+      myfile << s;
+    }
+  }
+}
+
+/// Logs the node creation with a list of input nodes.
+void LogContext::logNodeCreation(const Node *newNode) {
+  if (dumpCompilationLogOpt) {
+    std::vector<Node *> inputs;
+    for (size_t idx = 0; idx < newNode->getNumInputs(); idx++) {
+      inputs.push_back(newNode->getNthInput(idx).getNode());
+    }
+    std::string logStr;
+    logStr.append("[FULL SCOPE: ");
+    logStr.append(getFullScopeName());
+    logStr.append("]");
+    logStr.append(" --- CREATE { (Kind: ");
+    logStr.append(newNode->getKindName());
+    logStr.append(", Name: ");
+    logStr.append(newNode->getName());
+    logStr.append(") <== ");
+    for (auto n : inputs) {
+      logStr.append(" (Kind: ");
+      logStr.append(n->getKindName());
+      logStr.append(", Name: ");
+      logStr.append(n->getName());
+      logStr.append(") ");
+    }
+    logStr.append(" }\n");
+    addLogContent(logStr);
+  }
+}
+
+/// Logs the node replacement.
+void LogContext::logNodeReplacement(const Node *oldNode, const Node *newNode) {
+  if (dumpCompilationLogOpt) {
+    std::string logStr;
+    logStr.append("[FULL SCOPE: ");
+    logStr.append(getFullScopeName());
+    logStr.append(" ]");
+    logStr.append(" --- REPLACE { (Kind: ");
+    logStr.append(oldNode->getKindName());
+    logStr.append(", Name: ");
+    logStr.append(oldNode->getName());
+    logStr.append(") <== (Kind: ");
+    logStr.append(newNode->getKindName());
+    logStr.append(", Name: ");
+    logStr.append(newNode->getName());
+    logStr.append(") }\n");
+    addLogContent(logStr);
+  }
+}
+
+/// Logs the node deletion.
+void LogContext::logNodeDeletion(const Node &deletedNode) {
+  if (dumpCompilationLogOpt) {
+    std::string logStr;
+    logStr.append("[FULL SCOPE: ");
+    logStr.append(getFullScopeName());
+    logStr.append(" ]");
+    logStr.append(" --- DELETE { (Kind: ");
+    logStr.append(deletedNode.getKindName());
+    logStr.append(", Name: ");
+    logStr.append(deletedNode.getName());
+    logStr.append(") }\n");
+    addLogContent(logStr);
+  }
+}
+
+ScopedLogBlock::ScopedLogBlock(LogContext &ctx, llvm::StringRef name)
+    : ctx_(ctx), name_(name) {
+  ctx_.pushLogScope(name_);
+  ctx_.addLogContent("============= ENTER SCOPE: ");
+  ctx_.addLogContent(ctx_.getFullScopeName());
+  ctx_.addLogContent(" ================================\n");
+};
+
+ScopedLogBlock::~ScopedLogBlock() {
+  ctx_.addLogContent("============= EXIT SCOPE: ");
+  ctx_.addLogContent(ctx_.getFullScopeName());
+  ctx_.addLogContent(" ================================\n");
+  end();
+};
+
+void ScopedLogBlock::end() {
+  if (!end_) {
+    ctx_.popLogScope();
+  }
+  end_ = true;
+}
+
+} // namespace glow
--- a/lib/Graph/NodeValue.cpp
+++ b/lib/Graph/NodeValue.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "glow/Graph/NodeValue.h"
+#include "glow/Graph/Graph.h"
 #include "glow/Graph/Node.h"
 
 using namespace glow;
@@ -50,6 +51,12 @@ void NodeValue::replaceAllUsesOfWith(NodeValue v, const Function *F) const {
     assert(site->getNode() == node_ && "Invalid user");
     assert(site->getResNo() == getResNo() && "Invalid list of uses");
     site->setOperand(v.getNode(), v.getResNo());
+  }
+
+  // Log all nodes replacement.
+  if (getNode()->getParent()) {
+    getNode()->getParent()->getLogContext().logNodeReplacement(getNode(),
+                                                               v.getNode());
   }
 }
 
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -17,6 +17,7 @@
 #include "glow/Backend/Backend.h"
 #include "glow/Converter/TypeAToTypeBFunctionConverter.h"
 #include "glow/Graph/Graph.h"
+#include "glow/Graph/Log.h"
 #include "glow/Graph/Node.h"
 #include "glow/Graph/Nodes.h"
 #include "glow/Graph/PlaceholderBindings.h"
@@ -2587,6 +2588,9 @@ static void foldChannelShuffle(Function *F) {
 }
 
 void glow::fold(Function *F, CompilationContext &cctx) {
+  // Log the start of current log scope.
+  LOG_SCOPE(F->getLogContext(), "glow::fold")
+
   (void)cctx;
   // Get Reshape nodes merged into constants to simplify folding.
   optimizeReshape(F);
@@ -2608,6 +2612,9 @@ void glow::fold(Function *F, CompilationMode mode) {
 }
 
 void glow::optimize(Function *F, CompilationContext &cctx) {
+  // Log the start of current log scope.
+  LOG_SCOPE(F->getLogContext(), "glow::optimize")
+
   // Optimize may be called after backend specific transformations and some
   // nodes may have become unused. It is a good idea to remove them, before
   // proceeding with any further optimizations.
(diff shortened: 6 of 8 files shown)


include/glow/Support/Log.h
```diff
@@ -0,0 +1,301 @@
...
+
+  /// Pops out the most recently added scope.
+  void popBackLogScope();
+
+  /// Dumps the log into a file named after the given \p function name.
+  void dumpLog(llvm::StringRef funcName);
+};
+
+/// Logs the node creation with no input node. (Example: Splat)
+#define LOG_NODE_CREATE(newNode)                                               \
```
opti-mix: Does it need to be a macro? Or can it be a function?

include/glow/Support/Log.h
```diff
@@ -0,0 +1,301 @@
...
+    logStr.append(" --- CREATE { (");                                          \
+    logStr.append(newNode->getKindName());                                     \
+    logStr.append(", ");                                                       \
+    logStr.append(newNode->getName());                                         \
+    logStr.append(") }\n");                                                    \
+    getLogContext().addLogContent(logStr);                                     \
+  }
+
+/// Logs the node creation with a list of input nodes. (Example: Concat)
+#define LOG_NODE_CREATE_WITH_LIST(newNode, list)                               \
```
opti-mix: Does it need to be a macro? Or can it be a function?

include/glow/Support/Log.h
```diff
@@ -0,0 +1,301 @@
...
+      logStr.append(", ");                                                     \
+      logStr.append(n.getNode()->getName());                                   \
+      logStr.append(" ) ");                                                    \
+    }                                                                          \
+    logStr.append(" }\n");                                                     \
+    getLogContext().addLogContent(logStr);                                     \
+  }
+
+/// Logs the node creation with one input node. (Example: SaveNode)
+#define LOG_NODE_CREATE_WITH_ONE_INPUT(newNode, src)                           \
```
opti-mix: Does it need to be a macro? Or can it be a function?

include/glow/Support/Log.h
```diff
@@ -0,0 +1,301 @@
...
+    logStr.append(" --- DELETE { (");                                          \
+    logStr.append(deletedNode->getKindName());                                 \
+    logStr.append(", ");                                                       \
+    logStr.append(deletedNode->getName());                                     \
+    logStr.append(") }\n");                                                    \
+    getLogContext().addLogContent(logStr);                                     \
+  }
+
+/// Logs a new log scope.
+#define LOG_SCOPE(ctx, name) ScopedLogBlock __scope__(ctx, name);
```
opti-mix: Since the name of the scope is fixed to `__scope__`, one can have only one `ScopedLogBlock ` object in any concrete scope, otherwise there will be name conflicts. This is probably intentional. So, I'm just checking.

include/glow/Support/Log.h
opti-mix: If it can be a function, it could be declared here, but implemented in a cpp file.

lib/Graph/Graph.cpp
```diff
@@ -474,7 +475,12 @@ PadNode *Function::createPad(llvm::StringRef name, NodeValue input,
                              llvm::ArrayRef<int> pads, float value) {
   assertPadsSize(input, pads);
   auto OT = getParent()->uniqueType(*outTy);
-  return addNode(new PadNode(name, OT, input, mode, pads, value));
+  PadNode *newNode = addNode(new PadNode(name, OT, input, mode, pads, value));
+
+  // Log creation of node
+  LOG_NODE_CREATE_WITH_ONE_INPUT(newNode, input.getNode())
```
opti-mix: Is is possible to avoid touching all those `create` functions? 

May be one could intercept e.g. `Node::Node` constructor and do something there? This constructor would have the `Node *` as `this` and it would know the `name` of the new Node.  But I don't know how to intercept `input` there...

Alternatively, you could try to change `NodeGen` so that it emits some code into constructors of auto-generated nodes to call  a matching `LOG_NODE_CREATE_WITH_XYZ`.

jackm321:
Looks cool!
Can you add some example outputs to the test plan please.

lib/Graph/NodeValue.cpp
```diff
@@ -51,6 +53,11 @@ void NodeValue::replaceAllUsesOfWith(NodeValue v, const Function *F) const {
     assert(site->getResNo() == getResNo() && "Invalid list of uses");
     site->setOperand(v.getNode(), v.getResNo());
   }
+
+  // Log all nodes replacement.
+  if (getNode()->getParent()) {
+    LOG_NODE_REPLACE(getNode(), v.getNode())
```
jackm321: Is this the right level to log at? replaceAllUsesOfWith replaces one NodeValue with another and that information would be lost here.

include/glow/Support/Log.h
ZchiPitt: put into a function now. But i can't put that in Log.h or Log.cpp as I can't include the Node.h in that scope, which may cause include errors. So i put the function definition in Graph.h

include/glow/Support/Log.h
ZchiPitt: Can have another macro with named scope

lib/Graph/Graph.cpp
ZchiPitt: Done by moving the logging inside F->addNode()

lib/Graph/NodeValue.cpp
ZchiPitt: Yes, we record the nodes replacement by the last node and record all newly created nodes before replaceAllUsesOfWith() by logging at where we create these nodes (F->addNode() now)

ZchiPitt:
After discussing with @opti-mix , I propose to move the logging of nodes creation inside Function::addNode().

This is called by all the createXXXNode() methods inside Graph.cpp, which are where I put the logging statements before.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      logCtx_.addLogContent(logStr);
+    }
+  }
+
   /// Inserts the node \p N to the list of nodes, and returns the inserted node.
   template <class NodeTy> NodeTy *addNode(NodeTy *N) {
     N->setName(Module::uniqueName(N->getName(), uniqueNodeNames_));
     nodes_.push_back(N);
+
+    // Log the node creation.
```
nadavrot: Why do you need the list of inputs? You can inspect the node N inside the logNodeCreation method.

include/glow/Graph/Graph.h
```diff
@@ -222,12 +223,20 @@ class Function final : public Named {
...
   Module *parent_;
 
+  /// The log context associated with this function.
+  LogContext logCtx_;
+
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {}
+      : Named(Name), parent_(parent) {
+    logCtx_ = LogContext();
```
nadavrot: Why do you need this line?

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
 
   const Module *getParent() const { return parent_; }
 
+  /// Logs the node creation with a list of input nodes.
+  void logNodeCreation(Node *newNode, llvm::ArrayRef<Node *> inputs) {
```
nadavrot: Why is this in Graph.h and not in Log.cpp?

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
 
   const Module *getParent() const { return parent_; }
 
+  /// Logs the node creation with a list of input nodes.
```
nadavrot: Notice that this file is included in all of our compilation unit. This means that the three long methods that you've added here will be parsed and compiled for each one of the files in the project and will slow down the compilation process.

include/glow/Support/Log.h
```diff
@@ -0,0 +1,92 @@
...
+  std::vector<std::string> logContents_;
+
+  /// A vector that keeps track of current compilation scopes.
+  std::vector<std::string> logScopes_;
+
+  /// A string that represents the current full scope name.
+  std::string currentFullScope_ = "";
+
+public:
+  LogContext(){};
```
nadavrot: No semicolon at the end of functions. 

LocContext() = default;

lib/Graph/NodeValue.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "glow/Graph/NodeValue.h"
+#include "glow/Graph/Graph.h"
```
nadavrot: Why do we need to add the graph.h here?  Can you wrap the logic in a cpp function and use forward decleration method to avoid the extra include?

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2586,6 +2587,9 @@ static void foldChannelShuffle(Function *F) {
 }
 
 void glow::fold(Function *F, CompilationContext &cctx) {
+  // Log the start of current log scope.
```
nadavrot: RAII would be a good pattern here.

include/glow/Graph/Graph.h
```diff
@@ -222,12 +223,20 @@ class Function final : public Named {
   /// A reference to the owner of the function.
   Module *parent_;
 
+  /// The log context associated with this function.
```
nadavrot: Why do functions have log contexts? Shouldn't this belong elsewhere?

include/glow/Graph/Graph.h
nickgg: You could move these functions into the logContext to avoid adding all this log specific code into graph.h

opti-mix:
@ZchiPitt Thanks for providing an example of how an output looks like!

Nit: While this output is kind of nice, have you thought how easy/difficult it would be to parse it later for the sake of analysis, etc? Ideally, the output should be easy enough to parse, search, etc. 

Also, it seems like the names of nodes are not shown there (e.g. for Transpose and BatchNorm)? How can one find an info about the node in this log if the names are not part of this log?

lib/Optimizer/GraphOptimizer.cpp
nickgg: He is using RAII-based macros, in this case we could just remove the LOG_SCOPE_END() define.

include/glow/Graph/Graph.h
SplitInfinity: Agreed, this will be constructed automatically.

include/glow/Graph/Graph.h
```diff
@@ -222,12 +223,20 @@ class Function final : public Named {
...
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {}
+      : Named(Name), parent_(parent) {
+    logCtx_ = LogContext();
+  }
 
   ~Function();
 
+  /// Return the log context
```
SplitInfinity: Period.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
 
   const Module *getParent() const { return parent_; }
 
+  /// Logs the node creation with a list of input nodes.
+  void logNodeCreation(Node *newNode, llvm::ArrayRef<Node *> inputs) {
+    if (dumpCompilationLogOpt) {
+      std::string logStr = "";
```
SplitInfinity: I think you can avoid explicit initialization to `""` here. The string will be empty by default.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      logStr.append("[FULL SCOPE: ");
+      logStr.append(logCtx_.getFullScopeName());
+      logStr.append("]");
+      logStr.append(" --- CREATE { (");
+      logStr.append(newNode->getKindName());
```
SplitInfinity: I think it might be useful to prefix these values with the name of the field (i.e. the output should be `Kind: Transpose` instead of just `Transpose`). In your sample output, some nodes are named the same as their kind name and it's hard to tell which is which.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      logStr.append(", ");
+      logStr.append(newNode->getName());
```
SplitInfinity: Same here, but the output should be (`Name: <nodeName>` instead of just `<nodeName>`.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      logStr.append(") <== ");
+      for (auto n : inputs) {
+        logStr.append(" ( ");
+        logStr.append(n->getKindName());
```
SplitInfinity: Same here and below.

include/glow/Graph/Graph.h
SplitInfinity: Now that I think about it, instead of building these strings piece by piece like this, why not use `llvm::formatv` or `glow::strFormat`. They will allow you to construct these strings using a style similar to Python3 format strings (`llvm::formatv`) or C printf format strings (`glow::strFormat`). I think you can reduce some lines of code here and make things slightly more readable.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      }
+      logStr.append(" }\n");
+      logCtx_.addLogContent(logStr);
+    }
+  }
+
+  /// Logs the node replacement.
+  void logNodeReplacement(Node *oldNode, Node *newNode) {
+    if (dumpCompilationLogOpt) {
+      std::string logStr = "";
```
SplitInfinity: Same comments here about using `glow::strFormat` or `llvm::formatv` and including field names.

include/glow/Graph/Graph.h
```diff
@@ -261,10 +270,78 @@ class Function final : public Named {
...
+      logStr.append(newNode->getName());
+      logStr.append(" ) }\n");
+      logCtx_.addLogContent(logStr);
+    }
+  }
+
+  /// Logs the node deletion.
+  void logNodeDeletion(NodesList::iterator deletedNode) {
+    if (dumpCompilationLogOpt) {
+      std::string logStr = "";
```
SplitInfinity: And here.

lib/Support/Log.cpp
```diff
@@ -0,0 +1,84 @@
...
+void LogContext::dumpLog(llvm::StringRef funcName) {
+  if (dumpCompilationLogOpt) {
+    std::string compileLogFilename = funcName;
+    compileLogFilename.append("_compile.log");
+
+    llvm::outs() << "Writing compilation log file for Module to: "
+                 << compileLogFilename << '\n';
+    std::error_code EC;
+    llvm::raw_fd_ostream myfile(compileLogFilename, EC);
+    for (auto s : logContents_) {
```
SplitInfinity: As far as I know, `auto` never resolves to a reference type so this implicitly makes a copy of each string in `logContents_`. Consider changing this to `const auto&` (`const` to make sure the scope name is not accidentally modified) since you don't need a local copy in this case.

lib/Support/Log.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+ScopedLogBlock::ScopedLogBlock(LogContext &ctx, llvm::StringRef name)
+    : ctx_(ctx), name_(name) {
+  ctx_.addLogScope(name_);
+  ctx_.addLogContent("============= ENTER SCOPE: ");
+  ctx_.addLogContent(ctx_.getFullScopeName());
+  ctx_.addLogContent(" ================================\n");
+};
```
SplitInfinity: Please add an empty line below this.

lib/Support/Log.cpp
```diff
@@ -0,0 +1,84 @@
...
+void LogContext::popBackLogScope() {
+  if (logScopes_.size() > 0) {
+    logScopes_.pop_back();
+  }
+  currentFullScope_ = "";
+  for (auto s : logScopes_) {
+    currentFullScope_ += "->";
+    currentFullScope_ += s;
+  }
+}
```
SplitInfinity: Please add an empty line below this.

lib/Support/Log.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                          llvm::cl::desc("Dump compilation log"));
+
+void LogContext::addLogContent(llvm::StringRef logContent) {
+  logContents_.push_back(logContent);
+}
+
+void LogContext::addLogScope(llvm::StringRef scopeName) {
```
SplitInfinity: Why do you need to rebuild `currentFullScope_` each time? Can't you append `"-->" + scopeName` to the end of the current value of `currentFullScope_`?

lib/Support/Log.cpp
```diff
@@ -0,0 +1,84 @@
...
+  logScopes_.push_back(scopeName);
+  currentFullScope_ = "";
+  for (auto s : logScopes_) {
+    currentFullScope_ += "->";
+    currentFullScope_ += s;
+  }
+}
+
+void LogContext::popBackLogScope() {
```
SplitInfinity: Similarly, why do you need to rebuild `currentFullScope_` each time? Can't you discard the last `logScopes_.back().size()` characters of `currentFullScope_` plus a few more for the `-->`?

include/glow/Graph/Graph.h
ZchiPitt: moved to Log.cpp now

include/glow/Graph/Graph.h
ZchiPitt: moved to Log.cpp

include/glow/Graph/Graph.h
ZchiPitt: moved to Log.cpp now

SplitInfinity:
Also, s I think ahead to the tools that will be written on top of this log, I wonder if it is best to store the log data in memory as string or collections of strings. Those tools will have to parse the log to convert it into a richer in-memory representation in order to extract useful insights, so maybe it's a good idea to design that in-memory representation now and store that in the `LogContext` instead of plain strings.

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2607,6 +2611,9 @@ void glow::fold(Function *F, CompilationMode mode) {
 }
 
 void glow::optimize(Function *F, CompilationContext &cctx) {
```
SplitInfinity: I think it would be useful to log the start/end of the scopes of individual optimization passes (e.g. `DCE`) also. Logging only the scopes for `glow::optimize`, `glow::lower` and `glow::fold` is too coarse in my opinion.

include/glow/Graph/Graph.h
jfix71: Yeah I'm a bit confused -- @ZchiPitt  I thought you removed const from `CompilationContext` everywhere to then add the `LogContext` there?

opti-mix:
> Also, as I think ahead to the tools that will be written on top of this log, I wonder if it is best to store the log data in memory as string or collections of strings. Those tools will have to parse the log to convert it into a richer in-memory representation in order to extract useful insights, so maybe it's a good idea to design that in-memory representation now and store that in the LogContext instead of plain strings.

Since functions for producing the actual output are well-localized inside a single file now, changing the output format could be easily done at any time. We probably don't want to overcomplicate this PR.

But coming back to your remark, do you suggest to avoid using strings altogether and use a more efficient in-memory representation and log this representation in a binary form into the output stream? If this is the case, I'm not sure it is a very good idea. The tools will be eventually written using a different implementation language (e.g. Python, Ruby, SQL) and thus it is rather unlikely that they can share the in-memory representation with the producer of the log. More over, they may need different representations depending on what they do. But I agree that the log should ideally make it easy to parse it and to extract (only) the information needed (e.g. they it could be a CSV format or something with a very structured, easy to parse format. JSON? ;-).

> I wonder if it is best to store the log data in memory as string or collections of strings. 

I'm a bit concerned that this collection can grow very big on big networks and occupy a lot of memory. May be output streams should be used instead so that they are automatically flushed every time the internal buffer is filled.

opti-mix:
BTW, is there an issue related to the overall task of producing logs and building tools on top of this? If not, may be we should create one. Many of these design discussions could be held there.

SplitInfinity:
> The tools will be eventually written using a different implementation language (e.g. Python, Ruby, SQL) and thus it is rather unlikely that they can share the in-memory representation with the producer of the log.

I didn't know this, but writing these analysis tools in C++ seems like an incredibly masochistic exercise, so it makes sense that it would be in another language. I support the idea of choosing an easily parsed representation (e.g. JSON, CSV) rather than what I was suggesting.

lib/Graph/NodeValue.cpp
ZchiPitt: Well, there's a statement getNode()->getParent()->getLogContext() that will actually dereference the Function class, which makes forward declaration unable to work here

include/glow/Graph/Graph.h
ZchiPitt: After consideration, we think it's better to move the LogContext into Function instead of CompilationContext, such that we can avoid touching many places to pass the the CompilationContext into these optimization phases.

include/glow/Graph/Graph.h
ZchiPitt: Great suggestion!

lib/Optimizer/GraphOptimizer.cpp
ZchiPitt: I think so too.  Will add in following prs

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,101 @@
...
+  std::vector<std::string> logContents_;
+
+  /// A vector that keeps track of current compilation scopes.
+  std::vector<std::string> logScopes_;
+
+  /// A string that represents the current full scope name.
+  std::string currentFullScope_ = "";
+
+public:
+  LogContext() = default;
```
rdzhabarov: no need.

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,101 @@
...
+class LogContext {
+private:
+  /// A vector that keeps track of all log contents.
+  std::vector<std::string> logContents_;
+
+  /// A vector that keeps track of current compilation scopes.
+  std::vector<std::string> logScopes_;
+
+  /// A string that represents the current full scope name.
+  std::string currentFullScope_ = "";
```
rdzhabarov: it'll be empty string anyway, no need for this initialization.

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,101 @@
...
+  /// Add a scope into the scope vector.
+  void addLogScope(llvm::StringRef scopeName);
+
+  /// Getter that returns the current full scope name.
+  llvm::StringRef getFullScopeName() { return currentFullScope_; }
+
+  /// Pops out the most recently added scope.
+  void popBackLogScope();
+
+  /// Dumps the log into a file named after the given \p function name.
```
rdzhabarov: \p funcName (it needs to match) or just make param to be functionName.

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,101 @@
...
+  void dumpLog(llvm::StringRef funcName);
+
+  /// Logs the node creation with a list of input nodes.
+  void logNodeCreation(Node *newNode);
```
rdzhabarov: make sure that what can be const is const. Like if you do not intend to modify newNode it needs to be const.
Check all places in this file.

lib/Graph/Log.cpp
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+#include "glow/Graph/Log.h"
+#include "glow/Graph/Node.h"
+
+namespace glow {
+
+llvm::cl::opt<bool>
+    dumpCompilationLogOpt("dump-compilation-log",
+                          llvm::cl::desc("Dump compilation log"));
```
rdzhabarov: what'd default value?

lib/Graph/Log.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+void LogContext::addLogContent(llvm::StringRef logContent) {
+  logContents_.push_back(logContent);
+}
+
+void LogContext::addLogScope(llvm::StringRef scopeName) {
+  logScopes_.push_back(scopeName);
+  currentFullScope_ = "";
+  for (auto s : logScopes_) {
```
rdzhabarov: avoid copies here, const auto& s

lib/Graph/Log.cpp
```diff
@@ -0,0 +1,149 @@
...
+  if (dumpCompilationLogOpt) {
+    std::string compileLogFilename = funcName;
+    compileLogFilename.append("_compile.log");
+
+    llvm::outs() << "Writing compilation log file for Module to: "
+                 << compileLogFilename << '\n';
+    std::error_code EC;
+    llvm::raw_fd_ostream myfile(compileLogFilename, EC);
+    for (auto s : logContents_) {
+      myfile << s;
```
rdzhabarov: what happens if file already exist?

lib/Graph/Log.cpp
```diff
@@ -0,0 +1,149 @@
...
+  ctx_.addLogContent(" ================================\n");
+};
+ScopedLogBlock::~ScopedLogBlock() {
+  ctx_.addLogContent("============= EXIT SCOPE: ");
+  ctx_.addLogContent(ctx_.getFullScopeName());
+  ctx_.addLogContent(" ================================\n");
+  end();
+};
+
+void ScopedLogBlock::end() {
```
rdzhabarov: does it need to be `public` method?

lib/Graph/Log.cpp
ZchiPitt: added

ZchiPitt: should be false

lib/Graph/Log.cpp
ZchiPitt: changed to private

lib/Graph/Log.cpp
ZchiPitt: it will be overwritten. But I guess we don't really care if there was a previous log file there. What we care is the content that is being written into the file right now.

ZchiPitt: It should also be the user's job to archive previous log files into some other directory.

ZchiPitt: But i guess i can also have another command line option to pass the specific name for the log file in the future.

ZchiPitt:
I've created an issue for the design discussion of producing logs and the tooling on top of that #3064

ZchiPitt:
@SplitInfinity it's a great idea for thinking ahead of how to format the log output so that the parsing can be efficient. Currently im thinking of implement tooling with Python, so CSV or any other text format would be something i wanna try right now.
@jfix71 I like your idea of adding verbosity to the log, it definitely should be something added in the future.
@opti-mix 's idea of creating an issue for all the discussion of producing logs and tooling is great, and I created an issue for that. Any idea is welcomed there.

Many thanks to everyone for dropping your valuable advice to help me improve the code quality of this commit.

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,100 @@
...
+#define LOG_SCOPE(ctx, name) ScopedLogBlock __scope__(ctx, name);
+
+/// Helper class which traces the start and end of a compilation log scope.
+class ScopedLogBlock {
+
+  /// Reference to the log context.
+  LogContext &ctx_;
+
+  /// The name of the log scope.
+  llvm::StringRef name_;
```
opti-mix: Is it a good idea that it is a StringRef? If names are always constant string literals, it would work. But if the name is ever created dynamically it may result in some issues I'm afraid.

include/glow/Graph/Log.h
```diff
@@ -0,0 +1,100 @@
...
+#include "llvm/Support/raw_ostream.h"
+
+#include <fstream>
+#include <string>
+
+namespace glow {
+
+class Node;
+
+extern llvm::cl::opt<bool> dumpCompilationLogOpt;
```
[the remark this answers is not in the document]
ZchiPitt: moved

include/glow/Graph/Log.h
ZchiPitt: make sense

PR #3054, opened 2019-06-06
[Tensor] Add verbose mode to isEqual check for easier debugging.

jfix71:
Summary: Add a verbose mode to isEqual so that when there is a mismatch during our unit tests we log info the error. I disabled this in the two places we aren't using this in unit tests and so we want to early exit when Tensors aren't equal, i.e. GraphOptimizer and PlaceholderBindings.

Test Plan: N/A

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -381,8 +381,11 @@ public:
   std::string toString(unsigned maxNumElem) const;
 
   /// \returns true if the content of the other tensor \p other is identical to
-  /// this one.
-  bool isEqual(const Tensor &other, float allowedError = 0.0001) const {
+  /// this one, given some \p allowedError. If \p verbose and the tensors are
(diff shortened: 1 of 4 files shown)


include/glow/Base/Tensor.h
```diff
@@ -559,18 +562,36 @@ class Tensor final {
...
+        }
+        numExceedingError += 1;
+        if (maxErr < delta) {
+          maxErr = delta;
+          maxErrIdx = i;
+        }
       }
     }
-    return true;
+    if (numExceedingError != 0) {
```
arunm-git: What happens if the only error is a NaN at index > 0? Seems like we'll increment numExceedingError for that, but the maxErr < delta check will fail and hence the maxErrIdx and maxErr won't be updated, right?

jfix71: Good call -- I should have taken a hint from the above NAN comment. Fixed!

PR #3056, opened 2019-06-06
Back out "[glow][PR] build: try to goad CMake into understanding dependencies"

rdzhabarov:
Summary: Original commit changeset: 9a5be1ba400e

Differential Revision: D15698342


rdzhabarov:
@compnerd It's certainly this change broke couple of tests.
I'm rolling back this while there is an ongoing investigation.

PR #3057, opened 2019-06-06
[DeviceConfig] Make device config non optional in device manager.

rdzhabarov:
Summary:
* Make deviceConfig non optional.
* @superwizard2019 is adding deviceMemory to the config and it's super confusing to pass deviceConfig which could be null, memory as part of the config and memory in a separate arg.
* Do not need to pass backendKind and config in the same time, get rid of backendKind.

Following PR for deviceMemory: https://github.com/pytorch/glow/pull/2959

Documentation:
n/a

Test Plan:
CI

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -48,19 +48,15 @@ protected:
   BackendKind backend_;
 
   /// Configuration object for the device.
-  std::unique_ptr<DeviceConfig> config_;
+  DeviceConfig config_;
 
 public:
-  DeviceManager(BackendKind backend,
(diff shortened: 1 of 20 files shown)


include/glow/Backends/DeviceManager.h
```diff
@@ -51,8 +51,7 @@ class DeviceManager {
   std::unique_ptr<DeviceConfig> config_;
 
 public:
-  DeviceManager(BackendKind backend,
-                std::unique_ptr<DeviceConfig> config = nullptr)
+  DeviceManager(BackendKind backend, std::unique_ptr<DeviceConfig> config)
```
rdzhabarov: remove backendkind it can be taken from the config.

PR #3060, opened 2019-06-07
Use "bool foo()" syntax instead of "auto foo() -> bool"

beicy:
Summary:

Documentation:
ninja test.

[Optional Fixes #issue]

Test Plan:


beicy:
This is a follow up for #3003. I removed the closure but forget to forgot to change the return type.
Thanks @bertmaher and @opti-mix :)

PR #3062, opened 2019-06-07
[NFC][partitioner] refactoring the way to calculate the memory usage of a node set

beicy:
Summary:
Refactoring the way to calculate the memory usage of a node set -- update the memory usage by adding node to the nodes set one by one.
Will polish the the usage in following PR.

Documentation:
#2298 
[Optional Fixes #issue]

Test Plan:
Added unnitest.

### DIFF
--- a/include/glow/Partitioner/PartitionerUtils.h
+++ b/include/glow/Partitioner/PartitionerUtils.h
@@ -20,6 +20,7 @@
 #include "llvm/ADT/DenseMap.h"
 
 namespace glow {
+using NodesSetTy = std::set<Node *>;
 
 /// The memory usage of a subgraph (i.e. a list of nodes of a function).
 struct GraphMemInfo {
@@ -59,19 +60,24 @@ using BFSLevel = std::vector<std::vector<Node *>>;
 BFSLevel getBFSLevel(Function *F);
 
 /// Given \p nodes, return a list of nodes who use any node in this set.
-std::vector<Node *> getOutUsers(const std::set<Node *> &nodes);
+std::vector<Node *> getOutUsers(const NodesSetTy &nodes);
 
 /// Given \p nodes, return a list of nodes who use only the nodes in this set or
 /// constant.
-std::vector<Node *>
-getOutUsersWithOnePredecessor(const std::set<Node *> &nodes);
+std::vector<Node *> getOutUsersWithOnePredecessor(const NodesSetTy &nodes);
 
-/// Return the memory usage of the output caused by \p node who has users not in
-/// the set \p nodes.
-uint64_t getOutMemPerNode(const std::set<Node *> &nodes, const Node *node);
+/// \returns the memory usage of the output caused by \p node who has users not
+/// in the set \p nodes.
+uint64_t getOutMemPerNode(const NodesSetTy &nodes, const Node *node);
+
+/// Given nodes set \p currNodes and its memory usage info \p info, \returns the
+/// new memory usage if \p newNode is added into \p currNodes.
+GraphMemInfo updateGraphMemInfoByAddingNode(const NodesSetTy &currNodes,
+                                            const GraphMemInfo &info,
+                                            Node *newNode);
 
 /// Return the memory usage of a given nodes set.
-GraphMemInfo getGraphMemInfo(const std::set<Node *> &nodes);
+GraphMemInfo getGraphMemInfo(const NodesSetTy &nodes);
 
 } // namespace glow
 #endif // GLOW_PARTITIONER_PARTITIONUTILS_H
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -50,22 +50,6 @@ bool sortMinMemory(const std::pair<Function *, uint64_t> &a,
   return a.second < b.second;
 }
 
-/// Check if the memory of \p node inputs is calculated already. \returns the
-/// total used memory after \p node is considered.
-static uint64_t updateUsedMem(const std::set<Storage *> &usedStorage,
-                              std::set<Storage *> &newStorage, Node *node,
-                              uint64_t mem) {
-  uint64_t ret = mem;
-  for (size_t i = 0, e = node->getNumInputs(); i < e; i++) {
-    Storage *in = llvm::dyn_cast<Storage>(node->getNthInput(i).getNode());
-    if (in && usedStorage.find(in) == usedStorage.end()) {
-      ret += in->getType()->getSizeInBytes();
-      newStorage.insert(in);
-    }
-  }
-  return ret;
-}
-
 void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
   if (partitions_.size() == 0)
     return;
@@ -408,14 +392,6 @@ void Partitioner::partitionsAdjust(NodeToFunctionMap &partitions,
     nodesSet[(*it).second].insert((*it).first);
   }
 
-  // Initial the memory cost for each partition. Now we use the output size to
-  // represent the communication cost.
-  for (FunctionToNodesMapTy::iterator it = nodesSet.begin();
-       it != nodesSet.end(); ++it) {
-    GraphMemInfo cost = getGraphMemInfo((*it).second);
-    partitions.setGraphMemInfo((*it).first, cost);
-  }
-
   // Move/Exchange nodes between any two connected partitions, until no gain is
   // get.
   // Step1 Move: Assume Partition1 -> Partition2, try to move nodes from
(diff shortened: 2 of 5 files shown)


lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
   return ret;
 }
 
+/// \returns the list of users who use \p node's \p idx - th output.
+std::vector<Node *> getNodeUsers(Node *node, size_t idx) {
```
jfix71: This seems like a useful helper. Do you know if we do it elsewhere? If so perhaps it should be located somewhere more central so it can be reused in those places? Maybe even inside NodeValue.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+  std::vector<Node *> ret;
+  NodeValue nodeVal = node->getNthResult(idx);
+  for (auto &U : node->getUsers()) {
+    Node *user = U.getUser();
+    for (size_t i = 0, e = user->getNumInputs(); i < e; i++) {
+      NodeValue val = user->getNthInput(i);
+      if (val == nodeVal) {
+        ret.push_back(user);
+        break;
```
jfix71: Should this be a set instead of a vector? This could behave weird/unexpected if a node takes another node as input multiple times. For example given some node `N`, you could do `createMatMul("M", N, N)`. Then if you pass `N` into this function, it would return a vector `{N, N}`.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+
+    if (nodes.count(N)) {
+      // This input is inside of the nodes set. Let node1 belongs to the nodes
+      // set and only node uses nodes1's output(i.e. node1 -> node, and if node1
+      // -> node2, node2 belongs to the current nodes set). Before node is
+      // added, the size of node1's ouput is added into outMemSize of this set's
+      // GraphMemInfo. But if node is added, node1's output size should be
+      // removed from outMemSize.
+      std::vector<Node *> users = getNodeUsers(N, nodeVal.getResNo());
+      bool flag = true;
```
jfix71: nit: Can you name `flag` a bit more descriptively?

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+      // a PlaceHolder for the data from outside.
+      ret.inMemSize += N->getType(nodeVal.getResNo())->getSizeInBytes();
+      usedNodeValue.insert(nodeVal);
+    }
+  }
+
+  // The memory usage changes due to node's output.
+  if (auto *SN = llvm::dyn_cast<SaveNode>(node)) {
+    // For SaveNode, add the output size.
+    Storage *out = llvm::dyn_cast<Storage>(SN->getOutput().getNode());
```
jfix71: nit: Use `SN->getPlaceholder()` instead.

beicy:
> I didn't read the old implementation, just the new one. Is there new functionality added here? If so can you update the description of the PR?

No new functionality. I've updated the PR. Thanks!

include/glow/Partitioner/PartitionerUtils.h
```diff
@@ -66,10 +66,19 @@ std::vector<Node *> getOutUsers(const std::set<Node *> &nodes);
...
+/// \returns the list of users who use \p node's \p idx - th output.
+std::vector<Node *> getNodeUsers(Node *node, size_t idx);
+
+/// \returns the memory usage of the output caused by \p node who has users not
+/// in the set \p nodes.
 uint64_t getOutMemPerNode(const std::set<Node *> &nodes, const Node *node);
 
+/// Given nodes set \p nodes and its memory usage info \p info, \returns the new
+/// memory usage if \p node is added into \p nodes.
+GraphMemInfo updateGraphMemInfoByAddingNode(const std::set<Node *> &nodes,
```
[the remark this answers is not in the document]
beicy: The function doesn't really add any node. It just get the updated memory *IF* a node is added. We will use this result to check a node should be added or not.

lib/Partitioner/PartitionerUtils.cpp
beicy: The function returns the "user". In your case, the result is {Matmul_Node}. Once we detect a user, the user will be put into the result and break. So a user is only be added once.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+  return ret;
+}
+
+GraphMemInfo getGraphMemInfo(const std::set<Node *> &nodes) {
+  GraphMemInfo ret;
+  std::set<Node *> nodeSet;
+  for (std::set<Node *>::iterator it = nodes.begin(); it != nodes.end(); ++it) {
+    Node *cur = *it;
+    ret = updateGraphMemInfoByAddingNode(nodeSet, ret, cur);
+    nodeSet.insert(cur);
```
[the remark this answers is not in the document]
beicy: No. In other optimization place, if after this node is added, the memory usage exceeds the device memory restriction, we won't add the node.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +140,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
-    // For Save onde, the only required memory is for output.
-    if (auto *SN = llvm::dyn_cast<SaveNode>(cur)) {
-      Storage *out = llvm::dyn_cast<Storage>(SN->getOutput().getNode());
-      ret.outMemSize += out->getType()->getSizeInBytes();
+/// Given nodes set \p nodes and its memory usage info \p info, \returns the new
+/// memory usage if \p node is added into \p nodes.
+GraphMemInfo updateGraphMemInfoByAddingNode(const std::set<Node *> &nodes,
+                                            const GraphMemInfo &info,
+                                            Node *node) {
+  GraphMemInfo ret = info;
```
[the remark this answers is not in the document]
beicy: I moved next line after the first loop. But for the "GraphMemInfo ret = info;" I would like to keep it as the first line, since it is the return value of this function :)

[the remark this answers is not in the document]
beicy: I moved next line after the first loop. But for the "GraphMemInfo ret = info;" I would like to keep it as the first line, since it is the return value of this function :)

lib/Partitioner/PartitionerUtils.cpp
beicy: I will check it. If it is necessary, I will create another PR for this function:)

lib/Partitioner/Partitioner.cpp
```diff
@@ -493,28 +477,27 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
...
+  GraphMemInfo graphMem;
+
   for (int i = level - 1; i >= 0; i--) {
     for (size_t j = 0, e = bfs[i].size(); j < e; j++) {
       Node *N = bfs[i][j];
-      std::set<Storage *> newStorage;
-      uint64_t newMem = updateUsedMem(usedStorage, newStorage, N, mem);
-      if (newMem > availableMemory) {
+      graphMem = updateGraphMemInfoByAddingNode(nodeSet, graphMem, N);
+      if (graphMem.getTotalMemSize() > availableMemory) {
```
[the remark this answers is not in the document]
beicy: Comments added.

lib/Partitioner/PartitionerUtils.cpp
jfix71: Whoops, sorry I had a typo. I think the result here would be `{Matmul_Node, Matmul_Node}` actually. I think `getUsers()` in general is confusing because in this case `N` would be seen as having two users instead of one, even though it's the same user.

And actually I don't think we actually need to iterate over the inputs of `user` here to find which input `node` is into `user`. We're not using that info anywhere, and it's guaranteed `user` is at least one input of `node`.

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +139,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+
+/// \returns the memory usage of the output caused by \p node who has users not
+/// in the set \p nodes.
+uint64_t getOutMemPerNode(const NodesSetTy &nodes, const Node *node) {
+  uint64_t ret = 0;
+  for (size_t i = 0, e = node->getNumResults(); i < e; i++) {
+    std::vector<Node *> users = getNodeUsers(const_cast<Node *>(node), i);
+    for (auto &user : users) {
+      if (nodes.find(const_cast<Node *>(user)) == nodes.end()) {
+        ret += node->getType(i)->getSizeInBytes();
```
jfix71: nit: If you create a `NodeValue NV(node, i)` to pass above to `getNodeUsers()`, then you can also just do `NV.getType()` here.

lib/Partitioner/PartitionerUtils.cpp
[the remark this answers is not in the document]
jfix71: Everywhere this function is called, after we call it with some `GraphMemInfo info` we never use the `info` that was passed in again, right? So it seems unnecessary to create a copy, no?

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +139,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+        }
       }
+      if (removeable) {
+        ret.outMemSize -= nodeVal.getType()->getSizeInBytes();
+      }
+    } else {
+      // In this case, this input is not a storage type node nor belongs
+      // to this subgraph. Therefore, when creating paritions, we need to add
+      // a PlaceHolder for the data from outside.
+      ret.inMemSize += N->getType(nodeVal.getResNo())->getSizeInBytes();
```
jfix71: nit: You can just do `nodeVal.getType()->getSizeInBytes();`

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -140,72 +139,154 @@ getOutUsersWithOnePredecessor(const std::set<Node *> &nodes) {
...
+    }
+
+    if (currNodes.count(N)) {
+      // This input is inside of the nodes set. Let node1 belongs to the nodes
+      // set and only node uses nodes1's output(i.e. node1 -> node, and if node1
+      // -> node2, node2 belongs to the current nodes set). Before node is
+      // added, the size of node1's ouput is added into outMemSize of this set's
+      // GraphMemInfo. But if node is added, node1's output size should be
+      // removed from outMemSize.
+      std::vector<Node *> users = getNodeUsers(N, nodeVal.getResNo());
```
jfix71: nit: If you change `getNodeUsers()` to take a NodeValue then you can just pass `nodeVal` here.

lib/Partitioner/PartitionerUtils.cpp
beicy: "Everywhere this function is called, after we call it with some GraphMemInfo info we never use the info that was passed in again, right? So it seems unnecessary to create a copy, no?"
-- This function won't change the current nodeset, it is just a "predication".

lib/Partitioner/PartitionerUtils.cpp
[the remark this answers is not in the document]
beicy: No. It is not actually "adding", just a prediction at this stage.

lib/Partitioner/PartitionerUtils.cpp
beicy: Now I just passed` node->getNthResult(i)` to getNodeUsers, so I think it is OK to use `node->getType(i)` without introducing another var:)

PR #3063, opened 2019-06-07
[DO NOT MERGE][WIP] GLOW: Introduce early constant folding on model load

Max-K-cadence:
Summary:
This patch creates infrastructure that allows to constant-fold nodes early of
load, if their inputs are all constant, rather than leave it up to backend to
deal with them later. By doing so, we reduce the size of graph and enable more
models, because some nodes require that some of their inputs are constants.

Theoretically, Constant Folding can be done for all nodes. But this API allows
to implement it for particular nodes only. Two reasons for that:
- We can implement constant-folding incrementally and not make huge changes in
  code;
- We might not want to evalutate some of the nodes on load because of
  performance reasons.

This patch introduces constant folding for Gather node.
We can then add the constant folding to the nodes one by one as we need it.

[Optional Fixes #issue]

Test Plan:
Added unit test.

### DIFF
--- a/include/glow/Graph/ConstantFolding.h
+++ b/include/glow/Graph/ConstantFolding.h
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) 2019-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


Max-K-cadence:
Guess it would be cleaner to return Tensor...

Max-K-cadence:
Actually, we can avoid moving code out of the interpreter, and just call the interpreter directly instead. Hold this one please, it should be a cleaner solution.

Max-K-cadence:
Closing this for now. We will come up with a more universal solution.

PR #3066, opened 2019-06-07
[runtime] propagate backendOpts to provisioner

### LINKED ISSUE #2992: [runtime] Support BackendOpts in the runtime
We should support setting compilation options in the runtime. Currently HostManager::addNetwork does take in a CompilationContext as a parameter, but doesn't pass it into the provisioner, so it's not used in the actual compilation.

An easy test of this functionality would be to extend the tracing-compare example so it enables auto instrumentation - and you can see per operator events in the event json.

nickgg:
Summary: The HostManager addNetwork call takes in a CompilationContext but does not use it. This fixes the HM to pass the context through to the provisioner and use it when compiling.

Documentation: Fixes #2992

Test Plan: Tested with new argument to resnet-runtime:

```
./bin/resnet-runtime --trace-path trace.json
(0) ../tests/images/imagenet/zebra_340.png: 340
(0) ../tests/images/imagenet/cat_285.png: 281
(0) ../tests/images/imagenet/dog_207.png: 207
I0607 14:55:35.043989 327415232 resnet-runtime.cpp:219] Finished classifying 3 images.
dumping 39 trace events to trace.json.
```

Enabling auto-instrument:

```
./bin/resnet-runtime --trace-path trace.json --auto-instrument
(0) ../tests/images/imagenet/cat_285.png: 281
(0)  ../tests/images/imagenet/zebra_340.png: 340
(0) ../tests/images/imagenet/dog_207.png: 207
I0607 14:55:22.365221 201782720 resnet-runtime.cpp:219] Finished classifying 3 images.
dumping 756 trace events to trace.json.
```

### DIFF
--- a/include/glow/Runtime/Provisioner/Provisioner.h
+++ b/include/glow/Runtime/Provisioner/Provisioner.h
@@ -33,11 +33,13 @@ class Provisioner final {
 public:
   Provisioner(DeviceManagerMapTy &devices);
 
-  /// Walks \p networks and assigns each function to a DeviceManager in \p
-  /// devices. The Provisioner calls the addNetwork method for each
-  /// DeviceManager. Returns a GlowErr indicating if the operation was a
-  /// success.
-  llvm::Error provision(DAGListTy &networks, Module &module);
+  /// Traverses the DAG \p networks and:
+  ///   1. Retrieves each node's Function from the provided \p module.
+  ///   2. Compiles it using the provided CompilationContext \p cctx.
+  ///   3. Assigns a device and calls addNetwork on the chosen device(s).
+  /// \returns a GlowErr indicating if the operation was a success.
+  llvm::Error provision(DAGListTy &networks, Module &module,
+                        CompilationContext &cctx);
 
   /// Remove stored compiledFunction.
   void removeFunction(llvm::StringRef name);
--- a/lib/Runtime/Provisioner/Provisioner.cpp
+++ b/lib/Runtime/Provisioner/Provisioner.cpp
@@ -18,6 +18,7 @@
 #include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
 #include "glow/Graph/Graph.h"
+#include "glow/Support/Debug.h"
 
 #include <future>
 #include <map>
@@ -56,7 +57,8 @@ Provisioner::Provisioner(DeviceManagerMapTy &devices) {
   }
 }
 
-llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
+llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
+                                   CompilationContext &cctx) {
   // Walk the networks and group by logicalDeviceId.
   std::map<DeviceIDTy, std::vector<DAGNode *>> logicalDevices;
   // For each network visit all the partitions (nodes) and add the node to each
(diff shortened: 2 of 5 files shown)


include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -37,7 +37,8 @@ class Provisioner final {
   /// devices. The Provisioner calls the addNetwork method for each
   /// DeviceManager. Returns a GlowErr indicating if the operation was a
   /// success.
```
jfix71: nit: update Doxygen. Also appears to be a little stale with no reference to `module`, and `\p devices` isn't correct as devices isn't a parameter for `provision()`.

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -37,7 +37,8 @@ class Provisioner final {
...
-  llvm::Error provision(DAGListTy &networks, Module &module);
+  llvm::Error provision(DAGListTy &networks, Module &module,
+                        CompilationContext &cctx);
```
jfix71: It seems like you only use `backendOpts` out of `cctx` inside `provision()`, right? Do you expect to use other ops from `cctx` in the future? Otherwise might make sense to just pass `backendOpts`.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -74,6 +75,11 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
     }
   }
 
+  // Set collectConstants to false, this is because the DeviceManager will
+  // handle moving constants to the device, this way we can eliminate one
+  // copy operation.
+  cctx.backendOpts.collectConstants = false;
```
jfix71: It seems weird that a caller to `addNetwork()` could have set `backendOpts.collectConstants = true` and we would silently ignore it. Would it make sense instead to check that the caller set `collectConstants = false`, i.e. return an error if `!collectConstants`?

include/glow/Runtime/Provisioner/Provisioner.h
nickgg: I think theres a reasonable chance we do (otherwise why not just fold backendOpts into the context?). E.g. in @ZchiPitt's log diff we discussed storing the compilation log in the context and one of the downsides was that it wasn't plumbed through in a bunch of places that passed in BackendOpts instead.

lib/Runtime/Provisioner/Provisioner.cpp
nickgg: I could see logging the difference, but don't like exiting with an error. collectConstants defaults to true but doesn't make sense in the runtime.

jfix71: Ok, yeah IMO it'd be a good idea to log something that says e.g. ignoring collectConstants.

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -33,11 +33,13 @@ class Provisioner final {
...
-  /// Walks \p networks and assigns each function to a DeviceManager in \p
-  /// devices. The Provisioner calls the addNetwork method for each
-  /// DeviceManager. Returns a GlowErr indicating if the operation was a
-  /// success.
-  llvm::Error provision(DAGListTy &networks, Module &module);
+  /// Traverses the DAG \p networks and:
+  ///   1. Retrieves each node's Function from the provided \p module.
+  ///   2. Compiles it using the provided CompilationContext \p cctx.
+  ///   3. Assigns a device and calls addNetwork on the chosen device(s).
+  /// \p returns a GlowErr indicating if the operation was a sucess.
```
jfix71: nit: `\p returns -> \returns`

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -73,6 +75,16 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module) {
       }
     }
   }
+  if (cctx.backendOpts.collectConstants) {
+    DEBUG_GLOW(llvm::dbgs()
```
jfix71: Could also use glog (e.g. `LOG(INFO)`) since we're moving toward it everywhere.

include/glow/Runtime/Provisioner/Provisioner.h
jfix71: Also, `sucess -> success`

nickgg: oops

lib/Runtime/Provisioner/Provisioner.cpp
nickgg: ah right yeah, that helps a lot. What i really wanted was LOG_EVERY_MS or LOG_EVERY_N to not flood the log in high traffic scenarios.

PR #3069, opened 2019-06-10
Basic PyTorch Integration

jackm321:
Summary:
Setup basics of model loader and import `mul` and `div` JIT IR nodes.
Still need to clean things up and figure out some things about the setup.py script.

Documentation:

Test Plan:
Python source:
Screen Shot 2019-06-10 at 8 26 09 AM

JIT IR:
Screen Shot 2019-06-10 at 8 25 14 AM

Glow graph:
Screen Shot 2019-06-10 at 8 24 46 AM

Results of running on JIT vs Glow match
Screen Shot 2019-06-10 at 8 29 28 AM

added a pytest test:
`cd torch_glow && python setup.py test`
Screen Shot 2019-06-21 at 11 55 00 AM

### DIFF
--- a/torch_glow/setup.py
+++ b/torch_glow/setup.py
@@ -1,0 +1,206 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from distutils.spawn import find_executable
+from distutils import sysconfig, log
+import setuptools
+import setuptools.command.build_py
+import setuptools.command.develop
+import setuptools.command.build_ext
+
+from collections import namedtuple
+from contextlib import contextmanager
+import glob
+import os
+import shlex
+import subprocess
+import sys
+import argparse
+from textwrap import dedent
+import multiprocessing
+
+try:
+    import torch
+except ImportError as e:
+    print('Unable to import torch. Error:')
+    print('\t', e)
+    print('You need to install pytorch first.')
+    sys.exit(1)
+
+
+FILE_DIR = os.path.realpath(os.path.dirname(__file__))
+TOP_DIR = os.path.realpath(os.path.dirname(FILE_DIR))
+CMAKE_BUILD_DIR = os.path.join(TOP_DIR, 'build')
+
+CMAKE = find_executable('cmake') or find_executable('cmake3')
+if not CMAKE:
+    print('Could not find "cmake". Make sure it is in your PATH.')
+    sys.exit(1)
+
+install_requires = []
+setup_requires = []
+tests_require = []
+extras_require = {}
+
+# ################################################################################
+# # Flags
+# ################################################################################
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--debug", type=bool, default=False, help="Compile with debug on")
+parser.add_argument("--cmake_prefix_path", type=str, help="Populates -DCMAKE_PREFIX_PATH")
+args = parser.parse_known_args()[0]
+
+
+# ################################################################################
+# # Utilities
+# ################################################################################
+
+
+@contextmanager
+def cd(path):
+    if not os.path.isabs(path):
+        raise RuntimeError('Can only cd to absolute path, got: {}'.format(path))
+    orig_path = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(orig_path)
+
+
+# ################################################################################
+# # Customized commands
+# ################################################################################
+
+
+class cmake_build(setuptools.Command):
+    """
+    Compiles everything when `python setup.py develop` is run using cmake.
+
+    Custom args can be passed to cmake by specifying the `CMAKE_ARGS`
+    environment variable.
+    """
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def _run_cmake(self):
+        with cd(CMAKE_BUILD_DIR):
+            cmake_args = [
+                CMAKE,
+                '-DGLOW_BUILD_PYTORCH_INTEGRATION=ON',
+                '-DBUILD_SHARED_LIBS=OFF',
+                '-DCMAKE_EXPORT_COMPILE_COMMANDS=ON',
+                '-DCMAKE_BUILD_TYPE={}'.format('Debug' if args.debug else 'Release'),
+                '-DPYTHON_EXECUTABLE={}'.format(sys.executable),
+                # PyTorch cmake args
+                '-DPYTORCH_DIR={}'.format(
+                    os.path.dirname(os.path.realpath(torch.__file__))),
+            ]
+
+            if args.cmake_prefix_path:
+                cmake_args.append('-DCMAKE_PREFIX_PATH={}'.format(args.cmake_prefix_path))
+
+            if 'CMAKE_ARGS' in os.environ:
+                extra_cmake_args = shlex.split(os.environ['CMAKE_ARGS'])
+                log.info('Extra cmake args: {}'.format(extra_cmake_args))
+                cmake_args.extend(extra_cmake_args)
+            cmake_args.append(TOP_DIR)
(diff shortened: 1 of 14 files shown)


yinghai:
Cool! I think figuring our the testing CI is probably a major part of this. :)

jackm321:
@yinghai agreed, good CI testing will be critical here. Also CI is one of the motivations for locating the integration point in within the Glow repo because, with good testing in CI, we can more easily keep this interface from falling out of sync with Glow and becoming broken.

zrphercule2:
Cool thing. Are we going to use circle ci as well?

jackm321:
@zrphercule2, yeah circleci should work for this

lib/Importer/pytorch/CachingGraphRunner.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "glow/Support/Support.h"
+
+GraphInfo::GraphInfo(const torch::jit::Node *node)
+    : subgraph(node->g(at::attr::Subgraph)) {}
+
+int32_t CachingGraphRunner::addGraph(const torch::jit::Node *node) {
+  // TODO: just use node* as key
+  GraphInfo graphInfo(node);
+  auto graphId = nextGraphId_++;
+  graphIdToInfoMap_.insert({graphId, std::move(graphInfo)});
```
zrphercule2: If we are going to use continous graph id, why dont we use a vector instead of an unordered_map since it is faster?

zrphercule2: do we ever delete from this map

lib/Importer/pytorch/PyTorchModelLoader.cpp
```diff
@@ -0,0 +1,115 @@
...
+  assert(inputs.size() == subgraphInputValues.size());
+
+  for (size_t i = 0; i < subgraphInputValues.size(); ++i) {
+    torch::jit::Value *inputValue = subgraphInputValues[i];
+    const c10::IValue inputIValue = inputs[i];
+    inputValue->inferTypeFrom(inputIValue.toTensor());
+    loadRes_->inputPlaceholders.push_back(loadValue(inputValue));
+  }
+
+  // Nodes are topologically sorted.
```
zrphercule2: Is this always guaranteed by pytorch?

zrphercule2:
btw wondering what is the status of this pr? Went through it carefully, and looks great as a start.

zrphercule2:
also, do we need another repo for this project?

jackm321:
@zrphercule2 
status: I need to fix the build setup a bit, I should be able to get to it in the next working day or so
another repo: no I think we should keep it in Glow repo for easier development and keep it in sync with Glow repo

lib/Importer/pytorch/PyTorchModelLoader.cpp
jackm321: As far as I know yes, but only know this from reading other code that depends on this property

lib/Importer/pytorch/CachingGraphRunner.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include "glow/Support/Support.h"
+
+GraphInfo::GraphInfo(const torch::jit::Node *node)
+    : subgraph(node->g(at::attr::Subgraph)) {}
+
+int32_t CachingGraphRunner::addGraph(const torch::jit::Node *node) {
+  // TODO: just use node* as key
+  GraphInfo graphInfo(node);
+  auto graphId = nextGraphId_++;
```
zrphercule: Since glow is maintain a service here, which might receive continous cmd from front end for years in theory, think nextGraphId_++ is not very safe right? Agreeing using *node is much better.

lib/Importer/pytorch/CachingGraphRunner.cpp
```diff
@@ -0,0 +1,87 @@
...
+  glow::CompilationContext cctx;
+  executionEngine_.compile(loaderResult->f, cctx);
+
+  assert(inputs.size() == loaderResult->inputPlaceholders.size());
+
+  glow::PlaceholderBindings bindings;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    glow::Placeholder *ph = loaderResult->inputPlaceholders[i];
+    glow::TypeRef ty = ph->getType();
+    glow::Tensor t(inputs[i].toTensor().data_ptr(), ty);
```
zrphercule: where did we get the shape inference?

lib/Importer/pytorch/PyTorchModelLoader.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return std::move(loadRes_);
+}
+
+PyTorchModelLoader::PyTorchModelLoader(glow::Module &mod,
+                                       torch::jit::Graph *subgraph)
+    : mod_(mod), subgraph_(subgraph) {}
+
+// static
```
zrphercule: shall we have a static list for this?

zrphercule:
Overall LGTM, didnt reviewed the setup.py yet since I think you might want to change it. Feel free to ping me once you finished, and I will review all and give a stamp lol

lib/Importer/pytorch/CachingGraphRunner.cpp
jackm321: It compiles on first run

lib/Importer/pytorch/PyTorchModelLoader.cpp
jackm321: I don't think so, eventually this should probably look more like onnxifi's checkGraphCompatibility

torch_glow/setup.py
```diff
@@ -0,0 +1,213 @@
...
+setup_requires = []
+tests_require = []
+extras_require = {}
+
+# ################################################################################
+# # Global flags and environment variables
+# ################################################################################
+
+# DEBUG = bool(os.getenv('DEBUG'))
+DEBUG = True # TODO
```
zrphercule: Always true?

zrphercule:
After this I think we can start to:
1. build a working CI with standard tests for this.
We would expect an onnx-test like test system, for each operator we support, we need to create unit test; also we need some e2e tests similar to test_basic.py. These tests should be run in a special "glow_pytorch" CI.
2. start to add more ops includes Conv.
Although using switch-case is acceptable for now since we only have very few ops supported/to-be-support, I think it will finally make our codebase dirty, since all op related works are in a few functions, and make other people confusing when they what to add new ops.
I think we can instead make a global list/registry like what we have done in pytorch, caffe2 and onnx. for each op we support, we have a unique file to record all its related functionalities.

lib/Importer/pytorch/PyTorchModelLoader.cpp
zrphercule: > I don't think so, eventually this should probably look more like onnxifi's checkGraphCompatibility

Check my comments lol

torch_glow/setup.py
jackm321: fixed, added real argument parser

jackm321:
> build a working CI with standard tests for this.
We would expect an onnx-test like test system, for each operator we support, we need to create unit test; also we need some e2e tests similar to test_basic.py. These tests should be run in a special "glow_pytorch" CI.

100% agree, we should basically make an analog of the `Caffe2ImporterTest` in glow (but maybe have different tests in different files since each one will require a python program) and I agree we should also have some bigger end-end tests like model zoo models.

> I think we can instead make a global list/registry like what we have done in pytorch, caffe2 and onnx. for each op we support, we have a unique file to record all its related functionalities.
This sounds different than how we import for example a c2 node as a subgraph in glow then try to lower it and run `isOpSupported`, is that the process you're talking about or something else. The tricky thing with PyTorch right now is that, unlike c2, we don't have shape information yet during this compatibility checking phase.

PR #3070, opened 2019-06-10
[Runtime] Remove the Executor factory

nickgg:
Summary: Currently we have a factory method that creates Executors in the runtime - similar to Backends and DeviceManagers. However we only have one type of Executor, and I don't see us on a trend to needing multiple types you can drop in different scenarios. The factory approach also makes it harder to have different configurations for different executors (we'd need a DeviceConfig equivalent for Executors, probably).

So proposing to remove it and the associated indirection until a future point where we feel like we need it.

Documentation: n/a
Test Plan: Unit tests w/ ASAN.


SplitInfinity:
Overall, I agree with this decision. Perhaps I should have taken your suggestion on my PR for the first draft of the `Executor` and not added this in the first place. Two things I want to bring up:

1. Should we rename `ExecutorTest.cpp` to `ThreadPoolExecutorTest.cpp` as well?
2. We might need another kind of executor for a backend that supports peer-to-peer communication because some implementations of it require the communication primitives to be invoked in a specific order. It could also be the case that this requirement is better addressed by creating an abstraction for the execution `strategy` and writing a different implementation of that for backends that support peer-to-peer communication rather than a new `Executor` subclass. cc: @gcatron

nickgg:
@SplitInfinity I think we should keep the Executor interface as it's possible that we find writing another kind of executor useful (e.g. an inline executor which doesn't use multiple devicemanagers, possibly for AOT). What I don't think we need at the moment is the ability to switch Executors in and out without code change. I think P2P should be a feature of the standard executor but maybe something we could support disabling via configuration.

I'll rename the test file.

PR #3073, opened 2019-06-11
Refactoring ONNIXIFI Loader

putivsky:
Summary: Encapsulate ONIXIFI specific into the correspondent constructors for ONNXModelLoader and Caffe2ModelLoader

Differential Revision: D15745436


jackm321:
Nice!
We should just kill OnnxifiModelLoader now since it's basically nothing.

PR #3075, opened 2019-06-11
Prepare Glow Function for training, replacing constants with placehol…

putivsky:
…ders.

Summary:
Created a comprehensive test for TrainingPreparation utility using MNIST model as an example
Documentation:
Test Plan:
Run ./bin/mnist for both tests.

### DIFF
--- a/lib/Optimizer/TrainingPreparation.cpp
+++ b/lib/Optimizer/TrainingPreparation.cpp
@@ -1,0 +1,133 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Optimizer/TrainingPreparation.h"
+#include "glow/Base/Tensor.h"
+#include "glow/Graph/PlaceholderBindings.h"
+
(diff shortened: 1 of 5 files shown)


lib/Optimizer/TrainingPreparation.cpp
```diff
@@ -0,0 +1,132 @@
...
+      const bool isTrainable =
+          !isSoftMax &&
+          (!isBatchNormalization || (BatchNormalizationNode::MeanIdx != idx &&
+                                     BatchNormalizationNode::VarIdx != idx));
+
+      auto *PH = F->getParent()->createPlaceholder(C->getType(), C->getName(),
+                                                   isTrainable);
+
+      if (isSoftMax) {
+        selected = PH;
```
jfix71: It's possible for there to be multiple SoftMax in a single Function, for example in the inception models where there are intermediate SoftMaxes. I think it makes more sense to remove this finding of the selected PH and just have the caller do it (just with a simple loop over the Function's Nodes) since it understands the design of the model better.

putivsky:
Yes, I have checked the results of training, see line 280 at mnist.cpp
About, BatchNorm - I'm planning (after landing this PR) go back to ResNet20 (includes BN) PR and check training.

lib/Optimizer/TrainingPreparation.cpp
putivsky: The general problem even with this PR is that we assuming caller will have enough information about model topology. In practice, caller needs a generic API, which works out of the box for any model. The proposed API can be part of the other framework and should work as it is without additional work. I'm leaning to keep it. Anyway of "selected" is not found it will be nullptr and caller can take actions.

jfix71: The caller has to be aware at least of the inputs it wants to use for the model and associated names. We could always programmatically search through the Function for the corresponding input PHs based on their name to set their associated input Tensors. This is essentially what's going on when you do `loader.getNodeValueByName(inputName)`. You're just doing it differently for the SoftMax here because you know the model has a single SoftMax.

I think it's fine for now in this PR, but I think we should at least add a check here that selected is a nullptr before setting it, and if that fails say the pass only currently supports models with single SoftMax.

PR #3076, opened 2019-06-11
Remove GLOW_ASSERTs from OpenCLDeviceManager

jackm321:
Summary:
Remove GLOW_ASSERTs from OpenCLDeviceManager and instead forward errors to the callback.
Also replace asserts and llvm::errs() without glog

Documentation:
none

### DIFF
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -62,13 +62,13 @@ static llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
   return parsed;
 }
 
-cl_mem OpenCLDeviceManager::allocDeviceBuffer(uint64_t size) {
+llvm::Expected<cl_mem> OpenCLDeviceManager::allocDeviceBuffer(uint64_t size) {
   const uint64_t alignment = 128;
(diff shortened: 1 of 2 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -293,7 +303,7 @@ void OpenCLDeviceManager::evictNetworkImpl(std::string functionName,
   if (evictCB) {
     evictCB(functionName, std::move(err));
   } else {
-    llvm::errs() << llvm::toString(std::move(err));
+    LOG(ERROR) << llvm::toString(std::move(err));
```
nickgg: Might just be me but std err and LOG(ERROR) feel like different severities. Either way this is kind of a weird log line: either it's invalid not to have an `evictCB` in which case this should probably be a DCHECK, or it's valid in which case we probably don't want to spam the log with these.

jackm321: Yeah good point, my preference would be to have it be invalid to not have an evictCB

jackm321: Actually we should just do this for most/all CBs in the runtime

PR #3078, opened 2019-06-11
[Runtime] Add setup configuration to the HostManager

nickgg:
Summary: The Runtime has various configuration parameters that there is no mechanism to set (such as the number of threads in the ThreadPoolExecutor). This PR adds a config structure to HostManager allowing passing these parameters in and implements both the number of threads and the limit of active requests.

Documentation: just comments

Test Plan: Unit tests + new HostManagerTest covering configuring the activeRequestLimit

### DIFF
--- a/tests/unittests/HostManagerTest.cpp
+++ b/tests/unittests/HostManagerTest.cpp
@@ -40,30 +40,34 @@ std::unique_ptr<Module> setupModule(unsigned functionCount) {
   return module;
 }
 
-std::unique_ptr<HostManager> createHostManager(BackendKind kind) {
+std::unique_ptr<HostManager>
+createHostManager(BackendKind kind, HostConfig hostConfig = HostConfig()) {
   std::vector<std::unique_ptr<DeviceConfig>> configs;
-  auto config = llvm::make_unique<DeviceConfig>(kind);
-  configs.push_back(std::move(config));
+  auto deviceConfig = llvm::make_unique<DeviceConfig>(kind);
+  configs.push_back(std::move(deviceConfig));
   std::unique_ptr<HostManager> hostManager =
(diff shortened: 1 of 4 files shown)


opti-mix:
@nickgg Seems like a useful functionality!

Quick question: We already support config files for providing DeviceConfigs. Should we add another section to it for HostConfig?

nickgg:
@opti-mix I didn't know we did, actually. I'll take a look but makes sense.

The broken build is a compiler bug, ooh! (this one http://open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#253) Easy to work around I think.

nickgg:
I made #3079 for the config file improvement since it felt a bit involved to add into this PR.

tests/unittests/HostManagerTest.cpp
```diff
@@ -175,3 +179,33 @@ TEST_F(HostManagerTest, ConcurrentAddRemoveDuplicate) {
     t.join();
   }
 }
+
+/// Test that the HostManager respects it's configuration parameters.
+TEST_F(HostManagerTest, ConfigureHostManager) {
```
rosierm: I'm seeing two new unhandled Errors with my Release+Asserts build.

```
[ RUN      ] HostManagerTest.ConfigureHostManager
Program aborted due to an unhandled Error:
Error value was Success. (Note: Success values must still be checked prior to being destroyed).
Program aborted due to an unhandled Error:
Error value was Success. (Note: Success values must still be checked prior to being destroyed).
Aborted
```

nickgg:
@rosierm sorry didn't see your comment before landing, will coordinate with @jackm321 and one of us will put up a fix. Short term looks like you can just (void)runErr to suppress this.

PR #3080, opened 2019-06-11
[Tracing] change TraceEvent type from fixed size string to char

nickgg:
Summary: TraceEvents have a 1 char type value which we were storing as a string internally. Change to a char to save some memory inside the Runtime.

Documentation: n/a we already export chars through Onnxifi.

Test Plan: unit tests, specifically TraceEventTests.

### DIFF
--- a/include/glow/ExecutionContext/ExecutionContext.h
+++ b/include/glow/ExecutionContext/ExecutionContext.h
@@ -129,7 +129,7 @@ public:
 
   /// A helper function to log a TraceEvent at the current time, if there is a
   /// TraceContext available.
-  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "i",
+  void logTraceEvent(llvm::StringRef name, char type = TraceEvent::InstantType,
                      std::map<std::string, std::string> args = {}) {
     TraceContext *traceContext = getTraceContext();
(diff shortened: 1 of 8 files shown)


include/glow/ExecutionContext/ExecutionContext.h
```diff
@@ -129,7 +129,7 @@ class ExecutionContext {
 
   /// A helper function to log a TraceEvent at the current time, if there is a
   /// TraceContext available.
-  void logTraceEvent(llvm::StringRef name, llvm::StringRef type = "i",
+  void logTraceEvent(llvm::StringRef name, char type = 'i',
```
jackm321: why is this the only lowercase?

nickgg: It should not be, good catch.

rdzhabarov: InstantType::InstantType might be a better here instead of 'i'

nickgg:
@jackm321 The thing with an enum is you have to specify all valid options, but we don't support all types in the Trace Event spec. However, it is useful to be able to insert events of a different type sometimes, e.g. as we do in the tracing-compare example to set thread names as Metadata events (type 'M').

PR #3083, opened 2019-06-11
[Tracing] Add support for thread and process name metadata.

nickgg:
Summary: Add support for setting thread names and including them efficiently in the trace using the Metadata event type. 

Output looks like this (not much change since we were hacking it together in tracing-compare):

image

json file looks like this:
```
[
{"cat": "__metadata", "ph":"M", "ts":0, "pid":0, "tid":0, "name":"process_name", "args": {"name":"tracing-compare"} },
{"cat": "__metadata", "ph":"M", "ts":0, "pid":0, "tid":0, "name":"thread_name", "args": {"name":"CPU"} },
{"cat": "__metadata", "ph":"M", "ts":0, "pid":0, "tid":1, "name":"thread_name", "args": {"name":"Interpreter"} },
{"cat": "__metadata", "ph":"M", "ts":0, "pid":0, "tid":2, "name":"thread_name", "args": {"name":"OpenCL"} },
{"name": "allocBuffers", "cat": "glow","ph": "X", "ts": 1560293092212516, "pid": 0, "tid": 0, "dur": 8},
{"name": "loadPlaceholders", "cat": "glow","ph": "X", "ts": 1560293092212529, "pid": 0, "tid": 0, "dur": 416},
...
```

Documentation: Doesn't change much since we already claimed to support thread names

Since our json files have a metadata "header" now, I added a script in utils to concatenate trace files together, since thats something we do regularly.

We don't support multiple PIDs yet, but they'll probably be useful.

Test Plan: unit tests & manual testing with tracing-compare example

### DIFF
--- a/examples/tracing-compare.cpp
+++ b/examples/tracing-compare.cpp
@@ -44,10 +44,11 @@ llvm::cl::opt<std::string>
                               "a png with standard imagenet normalization"),
                llvm::cl::init("../tests/images/imagenet/dog_207.png"),
                llvm::cl::Positional, llvm::cl::cat(category));
-llvm::cl::opt<std::string>
-    outputJson(llvm::cl::desc("path to write output json trace events"),
-               llvm::cl::init("./glow-trace.json"), llvm::cl::Positional,
-               llvm::cl::cat(category));
+
+llvm::cl::opt<std::string> tracePath("trace-path",
+                                     llvm::cl::desc("Write trace logs to disk"),
+                                     llvm::cl::init("./glow-trace.json"),
+                                     llvm::cl::cat(category));
 
 } // namespace
 
@@ -76,7 +77,7 @@ std::unique_ptr<CompiledFunction> compileModel(Module &module,
   Function *F = module.getFunction("resnet50");
   Function *F_ = F->clone("resnet50." + backendName.str());
 
-  llvm::outs() << "Starting compile on " << backendName << ".\n";
+  llvm::outs() << "Starting compile on " << backendName << " device.\n";
   CompilationContext cctx;
   cctx.compMode = CompilationMode::Infer;
   cctx.backendOpts.autoInstrument = true;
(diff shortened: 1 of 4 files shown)


nickgg:
test failure here should be fixed in #3089.

examples/tracing-compare.cpp
```diff
@@ -167,25 +169,20 @@ int main(int argc, char **argv) {
...
 
-  std::vector<TraceEvent> allEvents;
-
-  allEvents.push_back({"thread_name", 0, 'M', 0, {{"name", "CPU"}}});
-  allEvents.push_back({"thread_name", 0, 'M', 1, {{"name", "Interpreter"}}});
-#if (GLOW_WITH_OPENCL)
-  allEvents.push_back({"thread_name", 0, 'M', 2, {{"name", "OpenCL"}}});
-#endif
+  TraceContext allEvents(TraceLevel::STANDARD);
+  for (unsigned i = 0; i < supportedBackends.size(); ++i) {
```
SplitInfinity: Mega nit: `for (unsigned i = 0, e = supportedBackends.size(); i < e; ++i) {`

I doubt we'll ever have enough backends for this to be a problem but should `i` be a `size_t` or maybe a `vector<T>::size_type`? You could also use a range based for with a separate counter.

include/glow/Backend/Backend.h
```diff
@@ -39,7 +39,6 @@ enum class BackendKind {
 };
 
 /// Stop gap measure while migrating from BackendKind to backend names.
-/// TODO: Remove this after migration.
```
SplitInfinity: Is this migration no longer happening?

nickgg: No idea, but if we have BackendKind we should have a kind->name map.

nickgg: Ah I see what you mean since we're removing BackendKind (e.g. in #3087). Will rebase and update.

nickgg:
rebased onto master to catch @rdzhabarov s Backend changes

nickgg:
rebased

PR #3084, opened 2019-06-11
[Interpreter] Implement peer-to-peer communication primitives

SplitInfinity:
**Description**
This commit implements peer-to-peer communication for the Interpreter
backend. This feature can be used to send/receive Tensors from a
function on one device to/from a second function on another device
without using host memory as an intermediary (which is the status quo).

This commit adds Send and Receive nodes to the graph IR to explictly
represent these operations as well as corresponding send and receive
instructions in the instruction IR. A SendNode and ReceiveNode are
"paired" at execution time using the Tensor bound to their "address"
inputs. This is a Tensor of shape {1} that contains some sort of token
or address that the backend can interpret however it wants to transmit
data between the Send and Receive nodes to which that token is bound.
In addition to sending and receiving data using peer-to-peer
communication, the Send and Receive nodes also write the data they
send/receive to a Placeholder so that they can be inspected (e.g. for
debugging).

In the case of the Interpreter backend, the "token" mentioned above is
the address of the ReceiveNode's Placeholder, and the data is
transferred using a future-promise pair of type Tensor (i.e.
std::future<Tensor> and std::promise<Tensor>). Each tuple of (function
name, placeholder address) maps to a future-promise pair, and the
Interpreter implementation of send fulfills the promise, and the
implementation of recv waits on the future.

**Testing**
This commit adds a new unit test, PeerToPeerTest, which tests
sending/receiving data with one sender + one receiver, one sender with
multiple send nodes and multiple receivers, and multiple senders and
multiple receivers.
```
$ ./tests/PeerToPeerTest 
[==========] Running 3 tests from 1 test case.
[----------] 3 tests from Interpreter/PeerToPeerTest
[ RUN      ] Interpreter/PeerToPeerTest.SingleSenderSingleReceiver/0
[       OK ] Interpreter/PeerToPeerTest.SingleSenderSingleReceiver/0 (6 ms)
[ RUN      ] Interpreter/PeerToPeerTest.SingleSenderMultipleReceivers/0
[       OK ] Interpreter/PeerToPeerTest.SingleSenderMultipleReceivers/0 (1 ms)
[ RUN      ] Interpreter/PeerToPeerTest.MultipleSendersMultipleReceivers/0
[       OK ] Interpreter/PeerToPeerTest.MultipleSendersMultipleReceivers/0 (2 ms)
[----------] 3 tests from Interpreter/PeerToPeerTest (11 ms total)

[==========] 3 tests from 1 test case ran. (11 ms total)
[  PASSED  ] 3 tests.

```
**Documentation**
Once the design is finalized after a few rounds of code review,
documentation will be added to reflect the updated design.

### DIFF
--- a/include/glow/Base/Type.h
+++ b/include/glow/Base/Type.h
@@ -275,6 +275,7 @@ enum class ElemKind : unsigned char {
   Int64ITy,      // 64-bit index type (int64_t)
   UInt8FusedQTy, // 8-bit quantized type with fused scale/offset (uint8_t)
   BoolTy,        // Bool type (bool)
+  AddrTy,        // Address type (uintptr_t)
 };
 
 /// \returns whether \p e is a quantized ElemKind.
@@ -473,6 +474,8 @@ struct Type final {
       return std::is_same<ElemTy, uint8_t>::value;
     case ElemKind::BoolTy:
       return std::is_same<ElemTy, bool>::value;
+    case ElemKind::AddrTy:
+      return std::is_same<ElemTy, uintptr_t>::value;
     }
     LOG(FATAL) << "Invalid type: " << getElementName(Ty).str();
   }
@@ -516,6 +519,8 @@ struct Type final {
       return sizeof(uint8_t);
     case ElemKind::BoolTy:
       return sizeof(bool);
+    case ElemKind::AddrTy:
+      return sizeof(uintptr_t);
     }
     LOG(FATAL) << "Invalid type: " << getElementName(Ty).str();
   }
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -838,6 +838,18 @@ public:
   SaveNode *createSave(llvm::StringRef name, NodeValue input,
                        Placeholder *output);
 
+  /// Creates a SendNode named \p name that sends \p input to \p address. A
+  /// ReceiveNode that listens on the same address will receive the sent data.
+  /// The node also writes the input to a Placeholder.
+  SendNode *createSend(llvm::StringRef name, NodeValue input,
+                       Placeholder *address);
+
+  /// Creates a ReceiveNode named \p name that receives a Tensor of type \p
+  /// outTy by listening on \p address and writes the data to its output. The
+  /// node also writes the input to a Placeholder.
+  ReceiveNode *createReceive(llvm::StringRef name, TypeRef outTy,
+                             Placeholder *address);
+
   /// Create quantization profile node named \p name for the output tensor from
   /// \p input in PlaceholderBindings \p bindings. Capture observed node name in
   /// quantization profile node as original node can be replaced during lowering
--- a/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
+++ b/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
@@ -38,12 +38,61 @@ DeviceManager *createInterpreterDeviceManager(const DeviceConfig &config) {
   return new InterpreterDeviceManager(config);
 }
 
+InterpreterDeviceManager::~InterpreterDeviceManager() {
+  // Fulfil any outstanding tensor pipe promises because you can't destroy them
+  // otherwise for some reason.
+  for (auto &p : functionsToTensorPipes_) {
+    for (auto &q : p.second) {
+      auto &promise = q.second.second;
+      promise.set_value(Tensor());
+    }
+  }
+}
+
 uint64_t InterpreterDeviceManager::getMaximumMemory() const {
   return maxMemoryBytes_;
 }
 
 uint64_t InterpreterDeviceManager::getAvailableMemory() const {
   return maxMemoryBytes_ - usedMemoryBytes_;
+}
+
+llvm::Expected<uintptr_t>
+InterpreterDeviceManager::getPlaceholderAddress(std::string function,
+                                                const Placeholder *P) {
+  // Peer-to-peer communication is implemented in the Interpreter backend using
+  // future-promise pairs (refered to a as a tensor pipe). Each (functionName,
+  // placeholder) pair maps to a tensor pipe that can be used to send/receive a
+  // Tensor.
+
+  // Convert P to uintptr_t since that's what's stored in the map.
+  uintptr_t addr = reinterpret_cast<uintptr_t>(P);
+
+  // Try to get the map from uintptr_t -> pair<future, promise> for the
+  // function. This should have been created when the function was added to the
+  // device.
+  auto mapIt = functionsToTensorPipes_.find(function);
+
+  DCHECK(mapIt != functionsToTensorPipes_.end()) << "Function not added";
+
+  auto &tensorPipes = mapIt->second;
+
+  // Try to find the tensor pipe for the given Placeholder.
+  auto pipeIt = tensorPipes.find(addr);
+
+  if (pipeIt == tensorPipes.end()) {
+    // If the tensor pipe doesn't already exist, create one.
+    std::promise<Tensor> p;
+    tensorPipes.insert(
+        std::make_pair(addr, std::make_pair(p.get_future(), std::move(p))));
+    pipeIt = tensorPipes.find(addr);
+  }
+
+  // Return the *address* of the tensor pipe. After it is bound to the address
+  // Placeholder of a SendNode or ReceiveNode, the interpreter implementations
+  // of these operations will use the future (Receive) or promise (Send) part of
+  // the pipe as is appropriate.
+  return reinterpret_cast<uintptr_t>(&pipeIt->second);
 }
 
 bool InterpreterDeviceManager::isMemoryAvailable(uint64_t estimate) const {
@@ -88,6 +137,11 @@ void InterpreterDeviceManager::addNetworkImpl(const Module *module,
     }
     functions_.emplace(func.first, func.second);
     usedMemoryBytes_ += functionCost_; // TODO:: static moduleSize
+
+    // Create a map of tensor pipes for the added function to facilitate
+    // peer-to-peer communication of tensors.
+    functionsToTensorPipes_.insert(
+        std::make_pair(func.first, TensorPipeMapTy()));
   }
 
   assert(usedMemoryBytes_ <= maxMemoryBytes_);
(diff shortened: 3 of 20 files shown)


lib/Backends/Interpreter/InterpreterDeviceManager.h
```diff
@@ -17,14 +17,23 @@
 #define GLOW_BACKENDS_INTERPRETER_INTERPRETERDEVICEMANAGER_H
 
 #include "glow/Backends/QueueBackedDeviceManager.h"
+#include <future>
 
 namespace glow {
+
+using TensorPipeTy = std::pair<std::future<Tensor>, std::promise<Tensor>>;
```
putivsky: Consider replace std::pair with custom struct, it's very possible to add new field in the future.

lib/Backends/Interpreter/InterpreterDeviceManager.h
```diff
@@ -17,14 +17,23 @@
...
+using TensorPipeMapTy = std::map<uintptr_t, TensorPipeTy>;
```
putivsky: Because lower_bound/upper_bound algorithm are not required then using 
using TensorPipeMapTy = std::unordered_map<uintptr_t, TensorPipeTy>;
would be more efficient.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -985,6 +986,56 @@ void BoundInterpreterFunction::fwdCrossEntropyLossGradInst(
   }
 }
 
+//===----------------------------------------------------------------------===//
+//                       Peer-to-peer Communication
+//===----------------------------------------------------------------------===//
+
+void BoundInterpreterFunction::fwdSendInst(const SendInst *I) {
+  using TensorPipeTy = std::pair<std::future<Tensor>, std::promise<Tensor>>;
```
putivsky: Already defined in lib/Backends/Interpreter/InterpreterDeviceManager.h

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -985,6 +986,56 @@ void BoundInterpreterFunction::fwdCrossEntropyLossGradInst(
...
+  // The value in the address input should be the address of a TensorPipeTy.
+  auto pipe = reinterpret_cast<TensorPipeTy *>(AH.raw(0));
+
+  // Set the value of the promise part of the tensor pipe.
+  auto &promise = pipe->second;
+  promise.set_value(inT->clone());
+}
+
+void BoundInterpreterFunction::fwdRecvInst(const RecvInst *I) {
+  using TensorPipeTy = std::pair<std::future<Tensor>, std::promise<Tensor>>;
```
putivsky: Same here

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -985,6 +986,56 @@ void BoundInterpreterFunction::fwdCrossEntropyLossGradInst(
...
+
+  // The value in the address input should be the address of a TensorPipeTy.
+  auto pipe = reinterpret_cast<TensorPipeTy *>(AH.raw(0));
+
+  // Read the value written by some SendNode using the future part of the tensor
+  // pipe.
+  auto &future = pipe->first;
+
+  future.wait();
+  auto t = future.get();
```
putivsky: https://en.cppreference.com/w/cpp/thread/future/get

future.get(); calls wait() internally.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -985,6 +986,56 @@ void BoundInterpreterFunction::fwdCrossEntropyLossGradInst(
...
+
+  // Copy the received Tensor to the output buffer.
+  outT->copyRawFrom(&t);
+
+  // Refresh the tensor pipe so it can be used again.
+  std::promise<Tensor> promise;
+  *pipe = std::make_pair(promise.get_future(), std::move(promise));
```
putivsky: It can compiler re-ordering/optimization problem:

at the time of assigning the first argument promise.get_future(), the second argument can be already moved std::move(promise).

include/glow/Base/Type.h
```diff
@@ -275,6 +275,7 @@ enum class ElemKind : unsigned char {
   Int64ITy,      // 64-bit index type (int64_t)
   UInt8FusedQTy, // 8-bit quantized type with fused scale/offset (uint8_t)
   BoolTy,        // Bool type (bool)
+  AddrTy,        // Address type (uintptr_t)
```
opti-mix: Host-side `uintptr_t` may be not enough to hold the address of the device buffer. I'd suggest using the existing `Int64ITy` for it or define `AddrTy` as int64_t.

include/glow/Graph/Graph.h
```diff
@@ -838,6 +838,18 @@ class Function final : public Named {
   SaveNode *createSave(llvm::StringRef name, NodeValue input,
                        Placeholder *output);
 
+  /// Creates a SendNode named \p name that sends \p input to \p address. A
+  /// ReceiveNode that listens on the same address will receive the sent data.
+  /// The node also writes the input to a Placeholder.
+  SendNode *createSend(llvm::StringRef name, NodeValue input,
+                       Placeholder *address);
```
opti-mix: Do we really want to use addresses for this? At the time of execution on the device, placeholders may not even exist on the device side. 

May be some kind of a tag would be better? A tag could be e.g. uint64_t (or bigger) and it could be mapped by the backend/DeviceMgr or even by the runtime on device to specific buffers.

include/glow/Graph/Graph.h
```diff
@@ -838,6 +838,18 @@ class Function final : public Named {
...
+
+  /// Creates a ReceiveNode named \p name that receives a Tensor of type \p
+  /// outTy by listening on \p address and writes the data to its output. The
+  /// node also writes the input to a Placeholder.
+  ReceiveNode *createReceive(llvm::StringRef name, TypeRef outTy,
+                             Placeholder *address);
```
opti-mix: Same here.

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -38,6 +38,17 @@ DeviceManager *createInterpreterDeviceManager(const DeviceConfig &config) {
   return new InterpreterDeviceManager(config);
 }
 
+InterpreterDeviceManager::~InterpreterDeviceManager() {
+  // Fulfil any outstanding tensor pipe promises because you can't destroy them
```
SplitInfinity: I'm still not sure why tests crash without doing this...

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -46,6 +57,44 @@ uint64_t InterpreterDeviceManager::getAvailableMemory() const {
...
+llvm::Expected<uintptr_t>
+InterpreterDeviceManager::getPlaceholderAddress(std::string function,
+                                                const Placeholder *P) {
+  // Peer-to-peer communication is implemented in the Interpreter backend using
+  // future-promise pairs (refered to a as a tensor pipe). Each (functionName,
+  // placeholder) pair maps to a tensor pipe that can be used to send/receive a
+  // Tensor.
+
+  // Convert P to uintptr_t since that's what's stored in the map.
+  uintptr_t addr = reinterpret_cast<uintptr_t>(P);
```
putivsky: May be you can consider to use incremental atomic counter instead? Addressed can be reused by compiler. Once memory got released new allocation can grab it and in case of race conditions wrong entry can be referred.

lib/Backends/Interpreter/InterpreterNodes.cpp
SplitInfinity: I would prefer to avoid this. If I move the definitions of these types to `Interpreter.h`, they'll be available here but not in the DeviceManager which also needs them.

lib/Backends/Interpreter/InterpreterNodes.cpp
SplitInfinity: Same here.

include/glow/Graph/Graph.h
SplitInfinity: I should have chosen a better name here. This "`address`" is really a token that the backend interprets however it wants. The value itself is used only to indicate which `Send` and `Receive` nodes are paired at runtime.

SplitInfinity:
@putivsky 

I considered sockets but decided against them because I thought we would have to implement tensor serialization/deserialization in order to send a tensor over a socket and that seemed like a lot of work. But based on our meeting yesterday, we might implementation serialization/deserialization to/from a file anyway, so maybe the socket approach will be viable here.

tools/ClassGen/NodeGen.cpp
```diff
@@ -54,6 +54,31 @@ int main(int argc, char **argv) {
                     "this node and all of its ancestor nodes. Generally "
                     "intended to save the final result of a network.");
 
+  BB.newNode("Send")
+      .addInput("Input")
+      .addInput("Address")
```
opti-mix: Just an idea: May be address should be a string? It could encode e.g. a transport and an address, e.g. `"memory://buffer_address"`, `"tcp://1.2.3.4:8080"`, `"PCIe://..."`, etc. This way you could specify which kind of channel to use for the communication.

SplitInfinity:
Naming things is hard. T_T

I used the word `address` but in reality, it is just some token or tag that the backend interpreters however it wants in order to pair the right Send node with the right Receive node. In this case, the address of the destination Placeholder is used as a tag to find the correct `std::promise` and `std::future` that will connect the Send/Receive nodes as desired.

nickgg:
Just summarizing my position from the last meeting: I really dislike ReceiveNodes for three reasons:

1. You're essentially describing synchronization primitives, and leaving implementation up to hardware vendors. How do we deal with variability in semantics, performance and behaviour across devices?
2. Synchronization introduces the possibility of deadlock, how do we debug these kinds of problems in the running system? What observability do we have about the status of executions that are waiting on a receive?
3. It's more complicated, without much gain. Does it matter where the Tensor came from to the second stage? Would we never have a situation where an input could come from P2P or from the host at runtime?

Separate question: is there any thought about P2P with heterogenous devices?

PR #3085, opened 2019-06-12
[compilationLog][version]: Include version number and latest commit's info in the compilation log.

ZchiPitt:
Summary:
Include version number and latest commit's info (date, hash) in the beginning of the dumped log. With this provided, we can link a stored log to a particular build of glow.

Test Plan:
Test run with googlenet_v1_slim/googlenet_v1_slim.onnx on image-classifier, and the output of the version and commit's info is shown below:
`<!-- Log Version: v1.0.0 -->`
`<!-- Log commit sha1: 9afe27d9 -->`
`<!-- Log commit date: 2019-06-11 -->`

### DIFF
--- a/include/glow/Graph/Log.h
+++ b/include/glow/Graph/Log.h
@@ -39,6 +39,8 @@ private:
   std::string currentFullScope_;
 
 public:
+  LogContext() { addLogMetaData(); };
+
   /// Add content into the contents vector.
   void addLogContent(llvm::StringRef logContent);
 
@@ -62,6 +64,11 @@ public:
 
   /// Logs the node deletion.
   void logNodeDeletion(const Node &deletedNode);
+
+private:
+  /// Add log metadata which includes version number and latest commit's info
+  /// (hash, date).
+  void addLogMetaData();
 };
 
 /// Logs a new log scope.
--- a/lib/Graph/Log.cpp
+++ b/lib/Graph/Log.cpp
@@ -22,9 +22,29 @@
 
 namespace glow {
 
+/// Log version number.
+static constexpr auto logVersionNo_ = "v1.0.0";
+
 static llvm::cl::opt<bool>
     dumpCompilationLogOpt("dump-compilation-log", llvm::cl::init(false),
                           llvm::cl::desc("Dump compilation log"));
+
+void LogContext::addLogMetaData() {
+  addLogContent("<!-- Log Version: ");
+  addLogContent(logVersionNo_);
+  addLogContent(" -->\n");
+#ifdef GIT_SHA1
+  addLogContent("<!-- Log commit sha1: ");
+  addLogContent(GIT_SHA1);
+  addLogContent(" -->\n");
+#endif
+#ifdef GIT_DATE
+  addLogContent("<!-- Log commit date: ");
+  addLogContent(GIT_DATE);
+  addLogContent(" -->\n");
+#endif
+  addLogContent("\n");
+}
 
 void LogContext::addLogContent(llvm::StringRef logContent) {
   logContents_.push_back(logContent);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,22 @@ include_directories(${GLOW_SOURCE_DIR})
 file(GLOB_RECURSE header_files include/*.h tools/*.h lib/*.h)
 add_custom_target(CollectHeaders SOURCES ${header_files})
 
+find_package(Git)
+# Get the commit's short SHA1
+execute_process(COMMAND
+  "${GIT_EXECUTABLE}" log -1 --pretty=format:"%h"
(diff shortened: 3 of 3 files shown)


opti-mix:
@ZchiPitt I like the overall idea of including this information into the log. But I'm not quite convinced by the implementation.

CMake has some support for generation of header files, see e.g. here: http://xit0.org/2013/04/cmake-use-git-branch-and-commit-details-in-project/

Another thing: For the git commit string you probably want to have just its hash, but not the whole commit message and author.

ZchiPitt:
@opti-mix I actually explored that alternative way to use configure_file(), but for that option I need to add two files, which I'm not sure which directory to put into. 

If we are going to have a "version.cc.in" and "version.h file", where do you think I should put them in.

lib/Graph/CMakeLists.txt
```diff
@@ -17,6 +17,19 @@ add_custom_target(AutoGenNode
                      "${NODES_DEF}")
 add_dependencies(AutoGen AutoGenNode)
 
+find_package(Git)
+# Get the commit's SHA1, author name, date and subject
+execute_process(COMMAND
```
rdzhabarov: side note: make sure to incorporate this in fbcode with buck TARGETS before merging.

nickgg:
Can you just use the `add_definition()` command in CMAKE to set up preprocessor constants rather than const variables? That way you don't need the files at all.

lib/Graph/Log.cpp
```diff
@@ -22,10 +22,23 @@
 
 namespace glow {
 
+const std::string Version::LOG_VERSION = "v1.0.1";
```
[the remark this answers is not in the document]
nickgg: Why do you need two versions? Is the the one in the logContext just left over?

opti-mix:
@ZchiPitt 

> I actually explored that alternative way to use configure_file(), but for that option I need to add two files, which I'm not sure which directory to put into. If we are going to have a "version.cc.in" and "version.h" file, where do you think I should put them in. Can I put the version.h under include/glow/Graph/ ? This seems a bit weird.

I haven't thought about concrete details here. Maybe @nickgg's suggestion to set up preprocessor constants and avoid any additional files at all is the simplest one?

As for the git commit information and how to provide it: It seems to me that this is a generally useful information and it is no specific to the compilation log. I think it should be exposed in such a way that any part of Glow could get this information and use it in debug, error prints, etc. It could be e.g. just a macro definition or it could be exposed via a dedicated include file (something like `Version.h` somewhere in Glow includes, e.g. under `Support` or `Base` or its own parent directory)

lib/Graph/Log.cpp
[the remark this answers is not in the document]
ZchiPitt: Ah sorry that was a typo. Also the versionNo_ is something i had in my initial design. I'll delete it's not used anymore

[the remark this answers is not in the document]
ZchiPitt: Version should be like v1.0.0 right now. It's just a tentative number i had in my mind, probably v0.1 or anything is better?

lib/Graph/CMakeLists.txt
ZchiPitt: checked with @bertmaher about embedding build info/commit hashes in our fbcode binary, and it seems that it's already taken care of, and can be invoked with certain buck flags.

ZchiPitt:
Updated the pr with @nickgg 's suggestion of just passing into const constant instead of const variable. Also it's only dumping commit hash and date right now. 

Also, I explored with the way of adding files under glow/include and /lib. 
It seems that by using configrue_file(), it will always create new files into the glow repo, which will make too much hassle as it would mess up how Git keeps tracks the files and the user have to un-check the newly created file before git commits.

lib/Graph/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_custom_target(AutoGenNode
                      "${NODES_DEF}")
 add_dependencies(AutoGen AutoGenNode)
 
+
```
SplitInfinity: Was this blank line intentionally added?

ZchiPitt: removed now :P

include/glow/Graph/Log.h
```diff
@@ -62,6 +64,11 @@ class LogContext {
 
   /// Logs the node deletion.
   void logNodeDeletion(const Node &deletedNode);
+
+private:
+  /// Add log meta data which includes version number and latest commit's info
```
rdzhabarov: nit: metadata

ZchiPitt:
@jfix71 GIT_SHA1 is the short SHA. And as far as I know fbcode does have implemented ways of embedding build info into the binary. I'll try to figure out how to pass it into the log.

PR #3086, opened 2019-06-12
[RecSys] Move RELU into create*MLP(); Cleanup uses of explicit indices in dims()

jfix71:


rdzhabarov:
Test failures seem to be about to be fixed in https://github.com/pytorch/glow/pull/3089

PR #3087, opened 2019-06-12
[Backends/2] Migrate from backendkind to backend statically registered name part 2.

rdzhabarov:
Summary:
Migrate most of the places where BackendKind used to backend name.
Most of the changes are purely mechanical.

Documentation:
updated

Test Plan:
CI

cc: @gcatron

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -69,13 +69,13 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
     for (size_t i = 0; i < node->children.size(); i++) {
       auto child = node->children[i];
       DescriptionBuilder db(child->name.c_str());
-      const std::string &backendName = backendMap_.at(child->backendKind).name;
-      db.addParam("BackendKind", backendName);
+      const std::string &backendName = backendMap_.at(child->backendName).name;
+      db.addParam("BackendName", backendName);
       myfile << "\"" << escapeDottyString(child->name) << "\""
              << " [ label = \"" << escapeDottyString(db) << "\"";
       myfile << "\tshape = \"record\"\n";
       myfile << "\tstyle=\"filled,rounded\"\n";
-      auto colorIdx = llvm::hash_value(llvm::StringRef(backendName));
+      auto colorIdx = llvm::hash_value(backendName);
       myfile << "\tfillcolor=" << getDotFileNodeColor(colorIdx) << "\n";
       myfile << "penwidth = 2];\n";
       if (used.count(child) == 0) {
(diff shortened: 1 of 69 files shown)


tests/unittests/QuantizationTest.cpp
```diff
@@ -2221,7 +2215,7 @@ TEST(Quantization, CheckAssertQuantization) {
 /// Check that we can quantize nodes that have some quantized outputs as unused,
 /// e.g. a TopK node where values is unused but indices is.
 TEST(Quantization, QuantizationZeroUsersResult) {
-  ExecutionEngine EE{BackendKind::Interpreter};
+  ExecutionEngine EE{"Interpreter"};
```
opti-mix: Why do you leave `"Interpreter"` here, but removed it in many other places?

tests/unittests/QuantizationTest.cpp
```diff
@@ -2158,7 +2152,7 @@ TEST(Quantization, TestProfileQuantizationOfLoweredFCRowwise) {
 
 /// Check that asserting quantization for the quantizer works as expected.
 TEST(Quantization, CheckAssertQuantization) {
-  ExecutionEngine EE{BackendKind::Interpreter};
+  ExecutionEngine EE{"Interpreter"};
```
opti-mix: Why do you leave "Interpreter" here, but removed it in many other places?

tests/unittests/QuantizationTest.cpp
```diff
@@ -1962,7 +1956,7 @@ static NodeClass *findNodeKindOrReturnNull(Function *F) {
 template <class BackendClass>
 static void testProfileQuantizationOfFC(bool expectLoweredFC,
                                         bool rowwiseQuantizeFC) {
-  ExecutionEngine profileEE(BackendKind::Interpreter);
+  ExecutionEngine profileEE("Interpreter");
```
opti-mix: Why do you leave "Interpreter" here, but removed it in many other places?

tests/unittests/QuantizationTest.cpp
```diff
@@ -622,7 +616,7 @@ TEST(Quantization, quantizeReLU) {
 /// are implemented as IntLookupTables, because the Interpreter only supports
 /// them as such.
 TEST(Quantization, quantizeLookupTables) {
-  ExecutionEngine EE{BackendKind::Interpreter};
+  ExecutionEngine EE{"Interpreter"};
```
opti-mix: Why do you leave "Interpreter" here, but removed it in many other places?

tests/unittests/QuantizationTest.cpp
rdzhabarov: Yeah, I'll remove in all. No particular reason.

rdzhabarov:
> Question: Why do you use std::string for storing the backend name (e.g. in the partitioner), but StringRef in some other cases? How do you decide which type to use? Can we use e.g. StringRef almost everywhere?

StringRef for passing. I think storage is std::string everywhere.

lib/Partitioner/Partitioner.cpp
```diff
@@ -69,13 +69,13 @@ void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
     for (size_t i = 0; i < node->children.size(); i++) {
       auto child = node->children[i];
       DescriptionBuilder db(child->name.c_str());
-      const std::string &backendName = backendMap_.at(child->backendKind).name;
-      db.addParam("BackendKind", backendName);
+      auto backendName = "TEST"; // backendMap_[child->backendName].name;
```
rdzhabarov: TODO: FIX THIS.

rdzhabarov:
@opti-mix yeah, all those cases where std::string is passed in tests/unittests/BackendTestUtils.h is a mistake. I'll fix it!

opti-mix:
@rdzhabarov 

> yeah, all those cases where std::string is passed in tests/unittests/BackendTestUtils.h is a mistake. I'll fix it!

OK. Just be careful with the runtime-related things like partitioner/devicemeanager/etc. I remember seeing some cases, where one would e.g. hold a reference to a function before calling them and then use it after the call. The problem was that this function was gone already due to the changes introduced by the partitioner ;-) As a result you'd get a dangling reference and all the usual problems with it. I hope backends do not disappear in the same way ;-)

rdzhabarov:
doc update (if any) and comment update is pending before merging.

rdzhabarov:
Many files touched, almost every commit brings conflict or failure. I'd like to merge this sooner.

rdzhabarov:
Since there are many moving pieces, many potential conflicts and mechanical nature of the change
I'll land this once fbcode build passes.

Happy to address any further comments in the follow ups.

PR #3089, opened 2019-06-12
fix flaky HostManager test

nickgg:
Summary: The HostManager test I added with the configuration struct is flaky due to timing, this serializes the timing using a couple of promises.

Documentation: n/a

Test Plan: ran tests


rdzhabarov:
i'm fine merging after tsan is fixed.

nickgg:
I made it worse :/ I should probably use a lock to to synchronize this.

nickgg:
switched to a lock, and changed the behaviour of activeRequests slightly so it doesn't tick down until the cb is finished. I think this is correct though because the request is still using an Executor thread.

PR #3095, opened 2019-06-13
Add unpaddedSize to move ctor/assignment op

bertmaher:
Summary: Forgot to add this; without swapping the unpaddedSize, moving the Tensor loses the fact that it is, in fact, an unpadded tensor.

Reviewed By: rdzhabarov

Differential Revision: D15794380

### DIFF
--- a/tests/unittests/TensorsTest.cpp
+++ b/tests/unittests/TensorsTest.cpp
@@ -940,3 +940,14 @@ max: 1515.200  min: 1.200
   EXPECT_EQ(mes2, expectMes2);
   EXPECT_EQ(mes2, osT3.str());
 }
+
+/// Test unpadded size.
+TEST(Tensor, unpaddedSize) {
+  Tensor partial(ElemKind::FloatTy, {11});
+  auto paddedType = Type::newShape(partial.getType(), {256});
+  auto bytes = partial.getSizeInBytes();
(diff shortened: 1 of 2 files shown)


tests/unittests/TensorsTest.cpp
```diff
@@ -940,3 +940,14 @@ max: 1515.200  min: 1.200
...
+
+/// Test unpadded size.
+TEST(Tensor, unpaddedSize) {
+  Tensor partial(ElemKind::FloatTy, {11});
+  auto paddedType = Type::newShape(partial.getType(), {256});
+  auto bytes = partial.getSizeInBytes();
+  Tensor T(partial.getUnsafePtr(), &paddedType, bytes);
+  EXPECT_EQ(T.getUnpaddedSizeInBytes(), bytes);
+  auto moved = std::move(T);
+  EXPECT_EQ(moved.getUnpaddedSizeInBytes(), bytes);
```
rdzhabarov: btw, should we use unpaddedsize in the isEqual method?

rdzhabarov:
Could you also update summary a bit.

tests/unittests/TensorsTest.cpp
bertmaher: I'm unsure.  I think it's ok if the "normal" tensor methods expect to have a fully-backed tensor, and crash horribly otherwise.

PR #3096, opened 2019-06-13
[ParameterSweepTest] Add Concat sweep tests

jfix71:
Add in sweep tests across various Concat parameters. I tried to select various number of inputs/sizes/number of dimensions that were representative of a wide range of workloads. On my Macbook using 8 cores and gtest-parallel this takes about 110 seconds on Release. This test is expensive for a reason 🙂


rdzhabarov:
I like the idea, but testing time jumps by 10 mins.

Do we want to check correctness or performance here?
Could we reasonably reduce the execution time?

jfix71:
Yeah I'm thinking of moving these to the stress test and adding them to internal nightly builds or something.

jfix71:
Ok I've moved it to `stress/`. The point of these is to check correctness across tons of different possible configs. I don't think it's necessary to test this on every commit.

tests/stress/ParameterSweepTest.cpp
```diff
@@ -263,3 +271,101 @@ TEST_P(FCSweepTest, FCTest_Float16) {
...
+    SweepTest, ConcatSweepTest,
+    ::testing::Combine(/* numInputs */ ::testing::Values(1, 2, 4, 8, 16, 32, 64,
+                                                         128, 192, 256),
+                       /* numDims */ ::testing::Range(1, 4),
+                       /* maxLength */ ::testing::Values(16, 32, 64, 128),
+                       /* axis */ ::testing::Range(0, 3)));
+
+/// Compare backend against the interpreter in Float.
+TEST_P(ConcatSweepTest, ConcatTest_Float) {
+  ENABLED_BACKENDS(CPU, OpenCL);
```
rdzhabarov: anything blocking to enable this on Habana as well?

tests/stress/ParameterSweepTest.cpp
```diff
@@ -263,3 +271,101 @@ TEST_P(FCSweepTest, FCTest_Float16) {
...
+  return std::make_pair(F, resultTensor);
+}
+
+/// Helper to test sweeping across a variety of configurations of a Concat by
+/// comparing the results to the Interpreter given some \p allowedError.
+/// \p config contains the backend to compare the Interpreter against, plus the
+/// specific configuration to run for this test. \p interpK and \p backendK are
+/// the element kinds to use for the Interpreter and backend, respectively.
+static void testParamSweepConcat(FourIntTupleConfig config, ElemKind interpK,
+                                 ElemKind backendK, float allowedError) {
```
rdzhabarov: nit: s/backendK/elemKind ?

jfix71: There's `interpK` and `backendK` -- I can rename them to `interpElemKind`/`backendElemKind`.

tests/stress/ParameterSweepTest.cpp
jfix71: #3101

PR #3098, opened 2019-06-13
Where operator

ayermolo:
Summary: CPU + Interpreter + Importer Test + Operator Tests
Implementation of Where Operator with multi directional broadcast support.*

*Two restrictions on that. Rank of X, Y, Cond must be equal.
Cond tensor must much first K dimensions of X,Y rest that don't match must be 1.

To support multi directional broadcast Broadcast "nodes" are added if necessary for X,Y. There is a more generic code in loader, but I think it better handle it within the create function. This way if it's called from multiple places it is handled under the hood automatically.
Documentation:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Where
[Optional Fixes #issue]

Test Plan:
Unit Tests

### DIFF
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -1286,6 +1286,16 @@ bool SpaceToDepthNode::verify() const {
   return sameType && dimTransform;
 }
 
+bool WhereNode::verify() const {
+  NodeValue X = getX();
+  NodeValue Y = getY();
+
+  if (X.getType() != Y.getType()) {
+    return false;
+  }
+  return true;
+}
+
 bool SaveNode::verify() const {
   return checkSameType(getInput(), getOutput(), this);
 }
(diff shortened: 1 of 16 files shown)


lib/Graph/Nodes.cpp
```diff
@@ -1286,6 +1286,16 @@ bool SpaceToDepthNode::verify() const {
   return sameType && dimTransform;
 }
 
+bool WhereNode::verify() const {
```
nadavrot: return getX().getType() == getY().getType();

jfix71:
@ayermolo It seems like this is very similar to our Select Node/Instruction. Why is it necessary to add Where?

ayermolo:
Well that's embarrassing. Somehow I missed it. I'll just re-implement loading of Where with Less and add broadcasts nodes during load of onnx graph.

PR #3102, opened 2019-06-13
[TraceEvents][OpenCL] Fix OpenCL trace output

SplitInfinity:
**Description**
This commit ensures that traces generated by the OpenCL backend will
render correctly regardless of the default time duration unit returned
by `std::chrono::system_clock().now()`. This function is used in
`TraceEvents::now()` to get the current timestamp but it is cast to
`std::chrono::microseconds` before it is returned. This commit modifies
the call to this function in `OpenCLFunction::translateTraceEvents` to
also cast the result to `std::chrono::microseconds`. Without this, the
events corresponding to OpenCL kernels do not render currently as being
nested inside events that should encompass them (e.g.
`DeviceManager::runFunction`) if the return value of `std::chrono::system_clock().now()` is not in microseconds.

**Testing**
Trace from `tracing-compare` on my laptop after this PR:
Screen Shot 2019-06-13 at 1 44 33 PM

Trace from `tracing-compare` on my devserver before this PR:
Screen Shot 2019-06-13 at 2 40 13 PM

All of the events that are *not* related to ops are in the sliver near the beginning and all the op events are in the sliver on the right.

Trace from `tracing-compare` on my devserver after this PR:
Screen Shot 2019-06-13 at 1 56 38 PM

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -1586,8 +1586,9 @@ void OpenCLFunction::translateTraceEvents(ExecutionContext *context) const {
   // system_clock domain.
   // TODO: synchronize clocks better, this can be off the thread was yielded
   // since getting the timestamp in updatePlaceholders.
-  int64_t tsOffset =
-      std::chrono::system_clock().now().time_since_epoch().count();
+  int64_t tsOffset = std::chrono::duration_cast<std::chrono::microseconds>(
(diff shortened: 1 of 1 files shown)


SplitInfinity:
@nickgg and I debugged this offline. it turns out that the loop that enqueues all kernels takes a really long time to enqueue each op and that's why there are so many gaps between the ops. I'm going to debug this separately and land this PR.

PR #3103, opened 2019-06-13
SLS test requires custom kernel, moving this test to KernelTests.

rdzhabarov:
Summary: Move SLS test to KernelTests, otherwise it's failing.

Differential Revision: D15814744


rdzhabarov:
fixed in a different diff.

PR #3104, opened 2019-06-13
Fixed OpenclDeviceManager to properly collect constants for a multi function module.

gcatron:
Summary:
Now that the runtime bundle only has symbols for its function an assumption in the OpenCL deviceManager no longer holds. This PR changes the OpenCLDeviceManager to collect and copy over constants for each function being added.
Actual changes are lines 223, 207 and the wrapping for loop. Git doesn't like indentation changes...
updated multiFunction unit test to have different sized functions.

Documentation: NA

Test Plan: ninja test

### DIFF
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -203,67 +203,64 @@ void OpenCLDeviceManager::addNetworkImpl(const Module *module,
                   func.first)
                   .str()));
     }
-  }
-  // Collect constants once, since currently the bundle grabs everything in the
-  // module.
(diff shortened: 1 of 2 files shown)


nickgg:
Do we have a unit test that actually does that? Multiple functions in a module for OCL?

gcatron:
> Do we have a unit test that actually does that? Multiple functions in a module for OCL?

I'm not sure we currently do, but I encountered this issue while porting EE to use HostManager once that lands there will be.

nickgg:
> I'm not sure we currently do, but I encountered this issue while porting EE to use HostManager once that lands there will be.

What I was getting at is a test plan of `ninja test` only works if your change is covered by one of our tests.

PR #3107, opened 2019-06-14
create separate node for fused-quantized-sls and enable lowering to slws

mortzur:
Summary:
* Create separate node for fused-quantized-SLS
* Enable lowering by default to fused-quantized-SLWS
* For Habana, do not lower (-->kernel exists, import and test)

Phabricator diff: D15817338
(needed D15866346 to land before this for kernel tests to pass)

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -764,14 +764,14 @@ public:
   /// them into len(\p lengths) entries: first Lengths[0] slices are aggregated
   /// to Result[0], next Lengths[1] slices are aggregated to Result[1], etc.
   /// I.e. sum(Lengths) must be equal to len(Indices).
-  FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
+  FusedRowwiseQuantizedSparseLengthsSumNode *
   createFusedRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
                                               Constant *data, NodeValue indices,
                                               NodeValue lengths);
 
   /// Same as \ref createFusedRowwiseQuantizedSparseLengthsSum(), but expects
   /// float input \p data, which is rowwise-quantized and fused internally.
-  FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
+  FusedRowwiseQuantizedSparseLengthsSumNode *
   createFusedRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
                                               Tensor &data, NodeValue indices,
                                               NodeValue lengths);
@@ -780,7 +780,7 @@ public:
   /// is multiplied by weights[i]. len(weights) must be equal to len(indices).
   FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
   createFusedRowwiseQuantizedSparseLengthsWeightedSum(llvm::StringRef name,
-                                                      Tensor &data,
+                                                      NodeValue data,
                                                       NodeValue weights,
                                                       NodeValue indices,
                                                       NodeValue lengths);
@@ -790,7 +790,7 @@ public:
   /// internally.
   FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
   createFusedRowwiseQuantizedSparseLengthsWeightedSum(llvm::StringRef name,
-                                                      Constant *data,
+                                                      Tensor &data,
                                                       NodeValue weights,
                                                       NodeValue indices,
                                                       NodeValue lengths);
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -1170,6 +1170,19 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
       multiInputs.emplace_back(std::move(inputs));
       break;
     }
+    case Kinded::Kind::FusedRowwiseQuantizedSparseLengthsSumNodeKind: {
+      auto *RI = llvm::cast<FusedRowwiseQuantizedSparseLengthsSumNode>(&I);
+      std::vector<synTensor> inputs = {
+          tensors[RI->getData()].get(),
+          tensors[RI->getIndices()].get(),
+          tensors[RI->getLengths()].get(),
+      };
+      chk(synCreateGenericNode(inputs.data(), &tensors[RI].get(), inputs.size(),
+                               1, nullptr, "sparse_lengths_sum_u8_2D_f32_embed",
+                               RI->getName().data(), nullptr, nullptr));
+      multiInputs.emplace_back(std::move(inputs));
+      break;
+    }
     case Kinded::Kind::FusedRowwiseQuantizedSparseLengthsWeightedSumNodeKind: {
       auto *RI =
           llvm::cast<FusedRowwiseQuantizedSparseLengthsWeightedSumNode>(&I);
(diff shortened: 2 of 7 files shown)


include/glow/Graph/Graph.h
```diff
@@ -780,7 +780,7 @@ class Function final : public Named {
   /// is multiplied by weights[i]. len(weights) must be equal to len(indices).
   FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
   createFusedRowwiseQuantizedSparseLengthsWeightedSum(llvm::StringRef name,
-                                                      Tensor &data,
+                                                      NodeValue data,
```
bertmaher: Why'd you swap this parameter with the method below?  Is there some subtlety I'm missing?

bertmaher:
Oh, thanks for catching this @jfix71.  I think we should revert the interpreter changes and instruction and just leave this for Habana.

mortzur:
> I believe there's something a little off here, and I don't know what the intention is. So just to walk this PR through a bit:
> 
> * Adds `RWQSLS` Node and Instruction
> * `RWQSLS` is lowered to `RWQSLWS` by default
> * Updated `shouldLower()` for Habana to prevent lowering `RWQSLS -> RWQSLWS`, but not other backends' `shouldLower()`
> * Only added kernel for `RWQSLS` to Interpreter, but not CPU
> 
> So it seems the CPU/Interpreter backends will still lower `RWQSLS -> RWQSLWS`. This means the `RWQSLS` Instruction is never being used since Habana doesn't use Instructions and all other backends will only use the `RWQSLWS` Instruction. And also the new Interpreter kernel for `RWQSLS` is going unused.
> 
> Is that right? If so I think we want to clarify what we expect/want the Interpreter/CPU backends to be doing here. I think either we should not lower `RWQSLS -> RWQSLWS` for the Interpreter/CPU and add a CPU kernel for `RWQSLS`, or we should remove the `RWQSLS` Instruction and Interpreter kernel entirely.

yes, it's redundant. thanks @jfix71  , I'll remove that.

lib/Backends/Habana/Habana.cpp
```diff
@@ -1169,6 +1169,19 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
...
     }
+    case Kinded::Kind::FusedRowwiseQuantizedSparseLengthsSumNodeKind: {
+      auto *RI = llvm::cast<FusedRowwiseQuantizedSparseLengthsSumNode>(&I);
+      std::vector<synTensor> inputs = {
+          tensors[RI->getData()].get(),
+          tensors[RI->getIndices()].get(),
+          tensors[RI->getLengths()].get(),
+      };
+      chk(synCreateGenericNode(inputs.data(), &tensors[RI].get(), inputs.size(),
+                               1, nullptr, "sparse_lengths_sum_u8_2D_f32_embed",
```
rdzhabarov: have we already landed this kernel?

lib/Graph/Graph.cpp
```diff
@@ -1478,9 +1478,9 @@ Function::createRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
 
 FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
 Function::createFusedRowwiseQuantizedSparseLengthsWeightedSum(
-    llvm::StringRef name, Constant *data, NodeValue weights, NodeValue indices,
+    llvm::StringRef name, NodeValue data, NodeValue weights, NodeValue indices,
```
rdzhabarov: can you explain this change?

lib/Graph/Nodes.cpp
```diff
@@ -1022,6 +1022,32 @@ bool FusedRowwiseQuantizedSparseLengthsWeightedSumNode::verify() const {
   return isValid;
 }
 
+bool FusedRowwiseQuantizedSparseLengthsSumNode::verify() const {
+  bool isValid = checkType(getResult(), ElemKind::FloatTy, this);
```
jfix71: Since this is virtually all the same logic from `FusedRowwiseQuantizedSparseLengthsWeightedSumNode::verify()`, can you outline it and share it mostly? E.g. take a look at what we do with `verifyConvolution`. You could pass an empty `NodeValue` for `Weights` and only verify things related to it when not empty.

lib/Graph/Graph.cpp
```diff
@@ -1492,15 +1492,21 @@ Function::createFusedRowwiseQuantizedSparseLengthsWeightedSum(
...
+FusedRowwiseQuantizedSparseLengthsSumNode *
 Function::createFusedRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
                                                       Constant *data,
                                                       NodeValue indices,
                                                       NodeValue lengths) {
-  auto ty = getParent()->uniqueType(ElemKind::FloatTy, {indices.dims()[0]});
-  auto ones = createSplat(name.str() + ".ones", ty, 1.0);
-  return createFusedRowwiseQuantizedSparseLengthsWeightedSum(name, data, ones,
-                                                             indices, lengths);
+  auto inDims = data->dims();
```
jfix71: This logic is just copied from `createFusedRowwiseQuantizedSparseLengthsWeightedSum()` right? Can we just add a static helper here called by both that returns the `TypeRef`?

lib/Graph/Graph.cpp
```diff
@@ -1539,15 +1545,28 @@ Function::createFusedRowwiseQuantizedSparseLengthsWeightedSum(
...
+FusedRowwiseQuantizedSparseLengthsSumNode *
 Function::createFusedRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
                                                       Tensor &data,
                                                       NodeValue indices,
                                                       NodeValue lengths) {
-  auto ty = getParent()->uniqueType(ElemKind::FloatTy, {indices.dims()[0]});
-  auto ones = createSplat(name.str() + ".ones", ty, 1.0);
-  return quantizeDataAndCreateFusedRowwiseQuantizedSparseLengthsWeightedSum(
-      this, name, data, ones, indices, lengths);
+  // For fused rowwise quantization, we must have a two-dimensional input. If
```
jfix71: Same as other -- can we try to share as much of this logic as possible across RWQSLS/RWQSLWS, e.g. in static helpers here in Graph.cpp?

lib/Backends/Habana/Habana.cpp
mortzur: This kernel already exists but not used. testing right now.

lib/Graph/Graph.cpp
mortzur: lowerNode(... Node* n ...) receives a pointer to Node as an argument and passes a const SpecificNode& to lowerSpecificNode(..), which later calls createSpesific(NodeValue nv).

In this case "Specific" is "FRQSLWS" which is also used by ModelLoader that passes it a Constant* which is why the existing signature.
Now, when I want to use it also from lowerSpecificNode(..) I cannot cast/convert from a const SpecificNode& to NodeValue. However I can implicitly convert from Constant* to NodeValue (since Constant is derived from Node and NodeValue wraps Node*).  
That looked like a reasonable thing to do...

lib/Graph/Nodes.cpp
mortzur: Done.

lib/Graph/Graph.cpp
mortzur: Done.

lib/Graph/Graph.cpp
mortzur: Done with modifying quantizeData...(...) a little but. Let me know you think it's OK.

lib/Graph/Graph.cpp
```diff
@@ -1476,42 +1476,47 @@ Function::createRowwiseQuantizedSparseLengthsSum(llvm::StringRef name,
       this, name, data, ones, indices, lengths, schema);
 }
 
-FusedRowwiseQuantizedSparseLengthsWeightedSumNode *
-Function::createFusedRowwiseQuantizedSparseLengthsWeightedSum(
-    llvm::StringRef name, Constant *data, NodeValue weights, NodeValue indices,
-    NodeValue lengths) {
-  auto inDims = data->dims();
+static TypeRef getOutputTypeOfFusedRowwiseQuantizedSLS(
```
jfix71: nit: Doxygen!

PR #3108, opened 2019-06-14
[Tracing] work around a bug in the Chrome trace visualizer

nickgg:
Summary: Looks like chrome://tracing has a bug when two Complete events have the same start time but the longer of the two occurs later in the json file. Work around this by sorting the TraceEvents by time and duration before dumping.

Before:
image

After:
image

Documentation: n/a

Test Plan: tested manually with a device

### DIFF
--- a/lib/ExecutionContext/TraceEvents.cpp
+++ b/lib/ExecutionContext/TraceEvents.cpp
@@ -30,6 +30,17 @@ void TraceEvent::dumpTraceEvents(
   llvm::errs() << "dumping " << events.size() << " trace events to "
                << filename.str() << ".\n";
 
+  // Chrome trace UI has a bug with complete events which are ordered later in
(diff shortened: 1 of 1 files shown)


lib/ExecutionContext/TraceEvents.cpp
```diff
@@ -30,6 +30,16 @@ void TraceEvent::dumpTraceEvents(
   llvm::errs() << "dumping " << events.size() << " trace events to "
                << filename.str() << ".\n";
 
+  // Chrome trace UI has a bug with complete events which are ordered later in
+  // the json than an event they completely enclose, so sort the list of events
+  // by start time and duration.
+  std::sort(events.begin(), events.end(), [](TraceEvent &a, TraceEvent &b) {
```
SplitInfinity: Nit: `const TraceEvent &a, const TraceEvent &b`?

PR #3082, opened 2019-06-11
[Tracing] add docs for TraceEvents

nickgg:
Summary: Add docs for the Tracing and TraceEvent functionality.

Documentation: yes, it is.

Test Plan: doc review, so really its your fault...

PR #3109, opened 2019-06-14
[Tracing] Switch back to steady_clock and fix Onnxifi trace events

nickgg:
Summary:  `steady_clock` is better than `system_clock` for timestamps since it doesn't jump around across threads. Use the better timer in glow and convert into FB domain at the Onnxifi interface.

Documentation: will update Tracing docs in #3082.

Test Plan: unit tests, manual check with tracing-compare, manual check events on a run with Habana device.

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -219,15 +219,28 @@ void Graph::setTraceEvents(onnxTraceEventList *traceEvents,
     return;
   }
 
+  /// Internally we use steady_clock, but our interface is system_clock
+  /// timestamps. Do a simple conversion.
+  auto steadyTS = TraceEvent::now();
+  auto systemTS = std::chrono::duration_cast<std::chrono::microseconds>(
+                      std::chrono::system_clock::now().time_since_epoch())
+                      .count();
+
+  // Timestamps are uint64_t so branch rather than use abs(), we want to make
+  // sure we always subtract the smaller from the larger value to avoid
+  // underflowing the uint64_t. Then if the timestamp should be moved backwards
+  // negate the result.
+  int64_t offset =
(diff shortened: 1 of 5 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -219,15 +219,25 @@ void Graph::setTraceEvents(onnxTraceEventList *traceEvents,
...
+  /// Internally we use steady_clock, but our interface is system_clock
+  /// timestamps. Do a simple conversion.
+  auto steadyTS = TraceEvent::now();
+  auto systemTS = std::chrono::duration_cast<std::chrono::microseconds>(
+                      std::chrono::system_clock::now().time_since_epoch())
+                      .count();
+
+  // Timstamps are uint64_t so branch rather than use abs().
+  int64_t offset =
+      steadyTS > systemTS ? -(steadyTS - systemTS) : (systemTS - steadyTS);
```
SplitInfinity: Consider adding a comment here to explain that you need to make sure you always perform a subtraction that won't underflow, and then negate it if needed. It's not immediately obvious.

lib/Onnxifi/Base.cpp
```diff
@@ -219,15 +219,25 @@ void Graph::setTraceEvents(onnxTraceEventList *traceEvents,
...
   }
 
+  /// Internally we use steady_clock, but our interface is system_clock
+  /// timestamps. Do a simple conversion.
+  auto steadyTS = TraceEvent::now();
+  auto systemTS = std::chrono::duration_cast<std::chrono::microseconds>(
+                      std::chrono::system_clock::now().time_since_epoch())
+                      .count();
+
+  // Timstamps are uint64_t so branch rather than use abs().
```
SplitInfinity: `Timstamps` -> `Timestamps`

PR #3110, opened 2019-06-14
[Quantization] Perform quantization opts in a loop together with DCE

artemrakhov:
Problem:
sometimes we have the following graph to optimize:
`Conv1 -> Rescale1 -> Rescale2 -> Conv2`
We want to get the following after Quantization opts:
`Conv1' -> Conv2`

But this may not happen, because we do optimizations step-by-step, and always check that optimized Conv has one user. This is what can happen after one step:
```
Conv1 -> Rescale1 -> Rescale2     Conv2
     \                            ^
      --> Rescale12 -------------/
```
Now `Conv1` has 2 users (one just added and one dead), so that `Rescale12` can't be combined into it.

Solution is to run DCE in a loop together with Quantization opts, while something changes.

It's possible to speed up optimization process by highlighting maximal chains and dealing with them one by one starting from head. E.g. in example above find `Conv1 -> Rescale1 -> Rescale2`, and process it all at once by replacing with `Conv1'`. It is doable, but code would be more complex and more difficult to maintain.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -1847,7 +1847,9 @@ static void optimizeReshape(Function *F) {
 /// Splat and Max can be eliminated if Splat value cannot impact the result.
 /// For example, Max and Splat can be removed if splat value is smaller
 /// than quantization range [min, max].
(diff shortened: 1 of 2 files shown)


artemrakhov:
Addressed the comments and added a test. PTAL

PR #3111, opened 2019-06-14
[compilationLog]: Log the change to the input(operand) of a node.

ZchiPitt:
Summary:
Log the change to the input(operand) of a node. This is for the case when the input of one node changes after the creation, which may happen at:
- transformForPrecisionMode
  - quantizeFunction(): via invocations to setNthInput() 
  - convertToFP16: via invocations to setOperand() and setOperand()
- several other trivial cases like changing the bias of a Conv node
  - via invocation to setNthInput()

This PR intercepts the implementation of setNthInput() and the certain caller(inside FunctionConverter) to setOperand(), such that the changes to the input of a node is caught and logged.  

Test Plan:
test with Quantization.TestProfileQuantizationOfUnloweredFC, and the log output is attached below:
Screen Shot 2019-06-14 at 4 17 32 PM

### DIFF
--- a/include/glow/Graph/Log.h
+++ b/include/glow/Graph/Log.h
@@ -26,6 +26,8 @@ namespace glow {
 
 class Node;
 
+struct NodeValue;
+
 /// A class for logging all compilation related activities.
 class LogContext {
 private:
@@ -60,10 +62,18 @@ public:
   void logNodeCreation(const Node *newNode);
 
   /// Logs the node replacement.
-  void logNodeReplacement(const Node *oldNode, const Node *newNode);
+  void logNodeValueReplacement(const NodeValue &oldNodeVal,
+                               const NodeValue &newNodeVal);
 
   /// Logs the node deletion.
   void logNodeDeletion(const Node &deletedNode);
+
+  /// Logs node's input changes.
+  /// \p user is the user node of the operands
+  /// \p prevOpr previous operand
+  /// \p newOpr new operand
+  void logNodeInputChange(const Node *user, const NodeValue &prevOprVal,
+                          const NodeValue &newOprVal);
 
 private:
   /// Add log metadata which includes version number and latest commit's info
--- a/lib/Converter/FunctionConverter.cpp
+++ b/lib/Converter/FunctionConverter.cpp
@@ -126,6 +126,12 @@ void FunctionConverter::convertOutputs(Node &node) {
       if (user == conversionVal.getNode()) {
         continue;
       }
+      // Log the change of node input(operand).
+      if (Function *F = node.getParent()) {
+        F->getLogContext().logNodeInputChange(user, *(use.get()),
+                                              conversionVal);
+      }
+
       use.get()->setOperand(conversionVal.getNode(), conversionVal.getResNo());
     }
   }
--- a/lib/Graph/Log.cpp
+++ b/lib/Graph/Log.cpp
@@ -16,8 +16,10 @@
 
 #include "glow/Graph/Log.h"
 #include "glow/Graph/Node.h"
+#include "glow/Graph/NodeValue.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/FileSystem.h"
+#include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/raw_ostream.h"
 
 namespace glow {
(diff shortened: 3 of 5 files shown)


lib/Graph/Log.cpp
```diff
@@ -143,6 +143,41 @@ void LogContext::logNodeDeletion(const Node &deletedNode) {
   }
 }
 
+/// Logs node's input changes.
+void LogContext::logNodeInputChange(const Node *user, const Node *prevOpr,
+                                    const Node *newOpr) {
+  if (dumpCompilationLogOpt) {
+    std::string logStr;
+    logStr.append("[FULL SCOPE: ");
```
nickgg: These really would look clearer with a structured format, e.g. llvm::format

lib/Converter/FunctionConverter.cpp
```diff
@@ -126,6 +126,10 @@ void FunctionConverter::convertOutputs(Node &node) {
       if (user == conversionVal.getNode()) {
         continue;
       }
+      // Log the change of node input(operand).
+      node.getParent()->getLogContext().logNodeInputChange(
```
nickgg: If you're recording a change to the Nth input you might need to log which input changes so you can reconstruct graphs where nodes have multiple inputs.

opti-mix: Question: Would it make more sense to intercept inside `setOperand` instead of doing it here in `convertOutputs`?

jfix71: Yeah, so `replaceAllUsesOfWith()` also calls `setOperand()`. It may instead make sense to just log all of this via `setOperand()`.

The downside here is that we end up spamming the log a bit -- instead of simply logging that we're replacing all users of `X` (`A`, `B`, `C`) with `Y`, we would log that `A` had its input changed to `Y` and `B` had its input changed to `Y` and `C` had its input changed to `Y`. However, we wouldn't need to remember to add a log call any time we use `setOperand()` elsewhere.

If we do decide to leave it here in the FunctionConverter, I'd suggest moving the log outside of this loop over users, as it would be more consistent with `replaceAllUsersOfWith()`, where we don't log each new user's change and instead just log NV `X` was replaced by NV `Y`.

ZchiPitt: Yea, i actually tried with intercepting setOperand(), but it turns out it's also called inside the constructor of each node, which may produce too much repeated information. Also, setOperand() is only rarely called in the codebase as this case is a special one because replaceAllUsesOf is not ideal being called in FunctionConverter. 

But @jfix71 did remind me that the interceptions should be consistent with what happened in replaceAllUsesOfWith(). I have also added statements there to log the node input changes, such that we have a consistent output right now.

@nickgg For setNthInput(), I think it's probably better to log which input is replaced rather than its index, for the ease of any tools built upon the log. Such that we can avoid any potential problems if the index in the log is different with the index in actual code.

lib/Graph/Log.cpp
```diff
@@ -86,60 +87,67 @@ void LogContext::logNodeCreation(const Node *newNode) {
...
+
+/// Logs node's input changes.
+void LogContext::logNodeInputChange(const Node *user, const Node *prevOpr,
+                                    const Node *newOpr) {
+  if (dumpCompilationLogOpt) {
+    addLogContent(
+        llvm::formatv(
+            "[FULL SCOPE: {0} ] --- NODE INPUT CHANGE { User(Kind: {1}, "
+            "Name: {2}) :: ",
+            getFullScopeName(), user->getKindName(), user->getKindName())
```
SplitInfinity: I think you meant to get the name of the user for the last argument here, i.e. : `getFullScopeName(), user->getKindName(), user->getName())`

lib/Graph/Log.cpp
```diff
@@ -86,60 +87,67 @@ void LogContext::logNodeCreation(const Node *newNode) {
...
+            .str());
+    if (prevOpr) {
+      addLogContent(llvm::formatv("PrevOpr(Kind: {0}, Name: {1}) <- ",
+                                  prevOpr->getKindName(), prevOpr->getName())
+                        .str());
+    } else {
+      addLogContent("PrevOpr(null) <- ");
```
SplitInfinity: Should this arrow point in the other direction (i.e. `prev -> new`)? That seems more intuitive to me.

ZchiPitt: sounds pretty reasonable!

ZchiPitt:
@SplitInfinity 

> It's probably a bit late to ask but Is this feature going to be enabled by default? Will the compilation log always be produced? I was under the impression that there would be an option in ComplilationContext that would toggle this on/off as needed.

This is what i have in mind. Will submit a pr later to add an option in LogContext

lib/Graph/Log.cpp
```diff
@@ -86,60 +87,67 @@ void LogContext::logNodeCreation(const Node *newNode) {
...
+                                getFullScopeName(), deletedNode.getKindName(),
+                                deletedNode.getName())
+                      .str());
+  }
+}
+
+/// Logs node's input changes.
+void LogContext::logNodeInputChange(const Node *user, const size_t idx,
+                                    const Node *prevOpr, const Node *newOpr) {
+  if (dumpCompilationLogOpt) {
```
jfix71: nit: Can you change all of these to be early exits? I.e.

```cpp
if (!dumpCompilationLogOpt) {
  return;
}
```

include/glow/Graph/Log.h
```diff
@@ -65,6 +65,10 @@ class LogContext {
   /// Logs the node deletion.
   void logNodeDeletion(const Node &deletedNode);
 
+  /// Logs node's input changes.
```
jfix71: nit: Expand on the Doxygen here to account for all the parameters.

lib/Graph/Log.cpp
```diff
@@ -86,60 +87,67 @@ void LogContext::logNodeCreation(const Node *newNode) {
...
+    addLogContent(
+        llvm::formatv(
+            "[FULL SCOPE: {0} ] --- NODE_INPUT_CHANGE { User(Kind: {1}, "
+            "Name: {2}) :: ",
+            getFullScopeName(), user->getKindName(), user->getName())
+            .str());
+    // prevOpr should never be null for the places we intercept.
+    addLogContent(
+        llvm::formatv("PrevOpr(Kind: {0}, Name: {1}) at index({2}) -> ",
```
jfix71: What does `index` represent here?

lib/Converter/FunctionConverter.cpp
```diff
@@ -90,6 +90,7 @@ void FunctionConverter::convertOutputs(Node &node) {
     //   "conversion".
     // 3. Node may be a module level value and we need one conversion per
     //    function.
+    size_t index = 0;
```
jfix71: I'm a bit confused by what `index` is here. First off you're only incrementing it when you get past a couple continues in the body of the loop. Second even if it was incremented at the top of the loop, it represents which user number this is, is that right? I'm not sure why this is relevant?

lib/Converter/FunctionConverter.cpp
```diff
@@ -126,6 +127,12 @@ void FunctionConverter::convertOutputs(Node &node) {
       if (user == conversionVal.getNode()) {
         continue;
       }
+      // Log the change of node input(operand).
+      if (node.getParent()) {
+        node.getParent()->getLogContext().logNodeInputChange(
+            user, index++, use.get()->getNode(), conversionVal.getNode());
```
nickgg: is conversionVal.getResNo() the same as index always? Maybe don't need the new counter.

lib/Graph/NodeValue.cpp
```diff
@@ -51,6 +52,12 @@ void NodeValue::replaceAllUsesOfWith(NodeValue v, const Function *F) const {
     assert(site->getNode() == node_ && "Invalid user");
     assert(site->getResNo() == getResNo() && "Invalid list of uses");
     site->setOperand(v.getNode(), v.getResNo());
+
+    // Log the change of node input(operand).
+    if (getNode()->getParent()) {
```
nickgg: minor thing, but it's a bit strange that the log is written before calling setOperand above in the FunctionConverter but after here in replaceAllUsesWith. Lets pick one and be consistent.

lib/Converter/FunctionConverter.cpp
ZchiPitt: thx for bringing the comment, changed to use ResNo instead

ZchiPitt:
@jfix71  Thanks for bringing this to me. You are right. I just changed to logNodeReplacement() and logNodeInputChange() to all use NodeValue as parameter instead.

@nickgg Updated with the new output of log now

include/glow/Graph/Log.h
```diff
@@ -60,11 +62,19 @@ class LogContext {
   void logNodeCreation(const Node *newNode);
 
   /// Logs the node replacement.
-  void logNodeReplacement(const Node *oldNode, const Node *newNode);
+  void logNodeReplacement(const NodeValue &oldNodeVal,
```
jfix71: nit: Maybe this should be renamed `logNodeValueReplacement()`

lib/Converter/FunctionConverter.cpp
```diff
@@ -126,6 +126,12 @@ void FunctionConverter::convertOutputs(Node &node) {
...
         continue;
       }
+      // Log the change of node input(operand).
+      if (node.getParent()) {
+        node.getParent()->getLogContext().logNodeInputChange(user, *(use.get()),
+                                                             conversionVal);
+      }
+
       use.get()->setOperand(conversionVal.getNode(), conversionVal.getResNo());
     }
```
jfix71: I think it make sense to also `logNodeReplacement()` here too, WDYT?

This function loops over all of the `NodeValues` from the input `node` and replaces each one with `conversionVal`. So this is essentially a specialized `replaceAllUsesOfWith()`.

ZchiPitt: Yea, that's right. However i found that in this scope(convertOutputs()) the conversionVal actually may belong to different function, which means it's one nodeVal replaced by several nodeVal for different users.

ZchiPitt: Actually call nodeReplacement for each different conversionVal may produce some messy output that makes parsing more difficult

PR #3113, opened 2019-06-14
[Tracing] rework TraceEvent verbosity

nickgg:
Summary: Rather than being a weird upward scale with some exceptions, this PR changes the TraceLevel verbosity to be a set of flags which specify kinds of events you want to enable as a bitmask (including the new TraceLevel Request which is coming soon). Also prevent adding new TraceEvents without specifying verbosity so they don't fall through the cracks.

Documentation: updated Tracing.md
Test Plan: unit tests, plus new TraceEventTest

### DIFF
--- a/docs/Tracing.md
+++ b/docs/Tracing.md
@@ -40,10 +40,14 @@ Instrumenting or tracing execution can have overheads that may not be tolerated 
 * At the compile stage the **autoInstrument** flag in the CompilationOptions structure enables instrumenting the compile of a network so that it is able to produce per operator Trace Events. If this flag is not enabled, there is no expectation of events at that level of detail (coarser grained events may still be produced).
 * At the execution stage the **ExecutionContext** may contain a **TraceContext** member describing what level of Trace Event verbosity is requested for this run. At the time of writing there are five verbosity levels defined:
     * `NONE` - no Trace Events are required and none should be emitted.
+    * `REQUEST`- only high level events relating to requests should be emitted (e.g. request beginning and end).
     * `RUNTIME` - only Glow Runtime events from on the host side should be emitted.
     * `OPERATOR` - Backend operator events should be emitted only, i.e. the time taken for each operator in the graph on the accelerator device. This may depend on compiling the network with operator level events (ie. autoInstrumenting).
-    * `STANDARD` - The default setting for instrumented runs: currently equivalent to `RUNTIME | OPERATOR`. 
     * `DEBUG` - Additional events for in depth debugging. This is intended for engineers familiar with the device and with Glow to triage performance issues in particular models. Many more events are permitted here, but included events should focus on what is high value for diagnosis issues.
+
+TraceLevel is a bitmask and levels can be combined. The common default is:
+
+    * `STANDARD` - currently equivalent to `RUNTIME | OPERATOR`. 
 
 Often the device specific runtime components are capable of emitting a large amount of hardware events. Please keep Trace Event memory size reasonable: a good guideline is about 1-2 Kb per millisecond of execution time on the device.
 
@@ -163,7 +167,7 @@ This does not have any thread synchronization over the TraceContext, however, so
 They can be created via the TraceContext:
 
 ```
-ctx.getTraceContext()->logTraceEvent("operator", "B", TraceEvent::now());
+ctx.getTraceContext()->logTraceEvent("operator", TraceLevel::RUNTIME, "B", TraceEvent::now());
 ```
 
 The timestamp is not required here, it will default to the current time.
@@ -172,7 +176,7 @@ This method is thread safe via a lock around the TraceContext.
 If you prefer to use RAII to determine the beginning and end timestamps we provide the **ScopedTraceBlock**:
 
 ```
-ScopedTraceBlock ev(ctx.getTraceContext(), "operator");
+ScopedTraceBlock ev(ctx.getTraceContext(), TraceLevel::RUNTIME, "operator");
 // do work
 return;
 ```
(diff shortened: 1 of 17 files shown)


docs/Tracing.md
```diff
@@ -40,11 +40,15 @@ Instrumenting or tracing execution can have overheads that may not be tolerated
 * At the compile stage the **autoInstrument** flag in the CompilationOptions structure enables instrumenting the compile of a network so that it is able to produce per operator Trace Events. If this flag is not enabled, there is no expectation of events at that level of detail (coarser grained events may still be produced).
 * At the execution stage the **ExecutionContext** may contain a **TraceContext** member describing what level of Trace Event verbosity is requested for this run. At the time of writing there are five verbosity levels defined:
     * `NONE` - no Trace Events are required and none should be emitted.
+    * `REQUEST`- only high level events relating to requests should be emitted (e.g. request beginning and end().
```
rdzhabarov: typos: (e.g. ... )

include/glow/ExecutionContext/TraceEvents.h
```diff
@@ -265,7 +278,13 @@ class ScopedTraceBlock {
   bool end_{false};
 
 public:
-  ScopedTraceBlock(TraceContext *context, llvm::StringRef name);
+  /// TraceContext constructor.
+  ScopedTraceBlock(TraceContext *context, TraceLevel level,
+                   llvm::StringRef name);
+
+  /// ExecutionContext constructor.
```
rdzhabarov: typo

rdzhabarov: I'd just remove those comments, since it does not bring any value. nbd anyway.

opti-mix:
@nickgg Overall, it seems very reasonable.

Quick question: Would it looks nicer if you'd introduce e.g. `TRACE_RUNTIME_EVENT_SCOPE_NAMED` and would use it as `TRACE_RUNTIME_EVENT_SCOPE_NAMED(tc, "synEnqueue", seEvent);` instead of your current form `TRACE_EVENT_SCOPE_NAMED(tc, TraceLevel::RUNTIME, "synEnqueue", seEvent);`. Or would be more or less the same in terms of aesthetics and number of keystrokes to be typed?

include/glow/ExecutionContext/TraceEvents.h
```diff
@@ -129,12 +129,14 @@ struct TraceInfo {
 };
 
 /// The amount and type of TraceEvents that should appear in the trace.
-enum class TraceLevel {
-  NONE,     // No trace events.
-  RUNTIME,  // Glow runtime events only.
-  OPERATOR, // Backend operator instrumentation only.
-  STANDARD, // Glow runtime events and backend operator events.
-  DEBUG     // Full debug events with extra information.
+enum TraceLevel {
```
SplitInfinity: I'm guessing the motivation for converting this to a regular enum was to enable bitmasking behaviour but I *think* a consequence of this is loss of independent scope.

I wouldn't be worried if this was in a class, but this is out in the open (well, inside `namespace glow`). Can we see someone trying to define `NONE`, DEBUG`, or `STANDARD` somewhere else later?

nickgg: Good call, we should make it a member of TraceEvent probably.

tests/unittests/TraceEventsTest.cpp
```diff
@@ -701,6 +701,31 @@ TEST(TraceEventsTest, nestedScopedEventsTerm) {
...
+                                      TraceLevel::RUNTIME,
+                                      TraceLevel::OPERATOR};
+  for (auto L : levels) {
+    TraceContext context(L);
+    for (auto evl : levels) {
+      context.logTraceEvent("event", evl);
+    }
+
+    if (L == TraceLevel::NONE) {
+      ASSERT_EQ(context.getTraceEvents().size(), 0);
```
rdzhabarov: in this case expect_eq should be fine, this way we'll also get a chance to test more things down below (before aborting).

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -230,7 +229,8 @@ llvm::Error BoundInterpreterFunction::execute(IRFunction *F,
   }
 
   {
-    auto ev = context->scopedEvent("eraseTensors");
```
rdzhabarov: q: why this was replaced by the macro below?

nickgg:
@opti-mix Think it ends up about the same in terms of keystrokes, but I like verbosity-as-parameter because it reduces the number of macros we need to keep in sync. E.g. we have 4 event creation macros and 5 verbosity levels then we need to have 20 macros. If we change the signature of creating an event we have to update 20 places, if we add a new verbosity level (eg. DEBUG1) thats 4 more macros.

docs/Tracing.md
nickgg: What's the typo here?

include/glow/ExecutionContext/TraceEvents.h
nickgg: What's the typo here too? I don't see it.

I can remove, though.

lib/Backends/Interpreter/InterpreterFunction.cpp
nickgg: To match the style of other files.

docs/Tracing.md
rdzhabarov: (e.g. request beginning and end().

check the parentheses.

include/glow/ExecutionContext/TraceEvents.h
rdzhabarov: `ExecutionContext constructor.` like this is not ExecutionContext constructor.

nickgg: Ah! It's the constructor that *takes* an `ExecutionContext`, as opposed to the constructor that takes a `TraceContext`. I could make that clearer or I could just delete both, what do you think?

rdzhabarov: I'd just delete :)

PR #3116, opened 2019-06-14
[NFC] Separate the check from replaceAllUsesOfWith()

artemrakhov:


artemrakhov:
This is needed in case if we cannot do required transformation in one step. Sometimes we need to replace A with B, and then replace B with C, but intermediate B may not have correct type.

nickgg:
@ZchiPitt Can you check this change interacts well in with your log work?

PR #3117, opened 2019-06-14
[Determinism] Replace unordered_map with ordered one in few places.

artemrakhov:
This PR partially addresses issue #1948


artemrakhov:
With this PR, placeholders will get the same memory addresses between runs. Compiler will produce the same program from run to run.

PR #3120, opened 2019-06-15
Add log scope information inside optimize, lower, fold, transformForPrecisionMode and transformPostLowering

ZchiPitt:
Summary:
Add log scope information inside:
- optimize
- lower
- fold 
- transformForPrecisionMode 
- transformPostLowering

Test Plan:
ninja check
Screen Shot 2019-06-16 at 11 47 58 PM
An example of log output looks like this right now.

### DIFF
--- a/lib/Backends/CPU/Transforms.cpp
+++ b/lib/Backends/CPU/Transforms.cpp
@@ -121,6 +121,8 @@ static Node *optimizeCPUMaxSplat(MaxNode *MN, Function *F) {
 
 bool CPUBackend::transformPostLowering(Function *F,
                                        CompilationContext &) const {
+  LOG_SCOPE(F->getLogContext(), "CPUBackend::transformPostLowering")
+
   bool changed = false;
   for (auto &node : F->getNodes()) {
     // Try to replace generic convolution with cpu-optimized version.
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -61,6 +61,8 @@ static bool shouldDeleteNode(Node *N) {
 }
 
 void glow::DCE(Function *F) {
+  LOG_SCOPE(F->getLogContext(), "DCE")
+
   auto &nodes = F->getNodes();
   auto &consts = F->getParent()->getConstants();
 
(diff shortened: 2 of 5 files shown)


lib/Backends/CPU/Transforms.cpp
```diff
@@ -121,6 +121,9 @@ static Node *optimizeCPUMaxSplat(MaxNode *MN, Function *F) {
 
 bool CPUBackend::transformPostLowering(Function *F,
                                        CompilationContext &) const {
+  // Log the start of current log scope.
```
opti-mix: Nit: I'm not sure that this comment that you use everywhere in this PR adds a lot of value.

ZchiPitt: LOG_SCOPE seems not to be directly relevant with the what is going on inside each optimization function, so i thought it might be better to add a comment here for developers to read

lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2760,16 +2835,25 @@ static llvm::Error checkAllNodesSupported(const Function &F, const Backend &B) {
...
+  // Log scope info.
+  LOG_SCOPE(F->getLogContext(), "transformForPrecisionMode")
   const PrecisionConfiguration &precConfig = cctx.precisionConfig;
 
   switch (precConfig.quantMode) {
   case QuantizationMode::Profile: {
     assert(cctx.bindings);
+
+    // Log scope info.
+    LOG_SCOPE(F->getLogContext(), "glow::profileQuantization")
```
jfix71: Why did you decide to log scopes for each case here inside `transformForPrecisionMode()`, instead of logging them inside the function call itself (e.g. inside `profileQuantization()` below?

ZchiPitt: quantization::quantizeFunction calls FunctionConverter::convert(), which is same as for precConfig.convertToFP16.

So I moved the LOG_SCOPE outside to differentiate them

lib/Backends/CPU/Transforms.cpp
opti-mix: My point was that the SCOPE declaration code is so self-describing that the comment does not add much. But it is really a very minor comment.

ZchiPitt:
@nickgg glow::optimize is called several times, the part of log i provided is probably the last time of glow::optimize being called, i think it makes sense there're few transformations in it.

lib/Backends/CPU/Transforms.cpp
nickgg: I agree, the comment isn't adding anything that isn't in the macro name.

ZchiPitt: sounds reasonable, i'll just remove them

PR #3121, opened 2019-06-17
 [runtime] Perform a round of backend-independent optimizations before partitioning

opti-mix:
- Move the logic for backend-independent optimizations performed before lowering into a dedicated function

- Perform a round of backend-independent optimizations before partitioning

   Doing so reduces the number of nodes to be partitioned and partitioner has less work to do.

   When using a custom experimental backend and a heterogeneous partitioning, I've seen the number of partitions produced for Glow's ResNet50 model going down from 101 to 3 after running a round of backend-independent optimizations before partitioning.

### DIFF
--- a/lib/Optimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer.cpp
@@ -2845,11 +2845,9 @@ static void transformForPrecisionMode(const Backend &B, Function *F,
   }
 }
 
-// NOTE: When updating this function, please also update the documentation in
-// docs/GraphOptimizationPipeline.md
-llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
-                                   CompilationContext &cctx) {
-  LOG_SCOPE(F->getLogContext(), "glow::optimizeFunction")
+llvm::Error glow::optimizeFunctionBeforeLowering(Function *F,
+                                                 CompilationContext &cctx) {
+  LOG_SCOPE(F->getLogContext(), "glow::optimizeFunctionBeforeLowering")
 
   // Verify the function pre-optimization/lowering.
   assert(F->verify() && "Function must be valid");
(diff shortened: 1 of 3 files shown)


lib/Optimizer/GraphOptimizer.cpp
```diff
@@ -2812,7 +2808,16 @@ llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
...
+  return llvm::Error::success();
+}
 
+// NOTE: When updating this function, please also update the documentation in
+// docs/GraphOptimizationPipeline.md
+llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
+                                   CompilationContext &cctx) {
+  RETURN_IF_ERR(optimizeFunctionBeforeLowering(F, cctx));
+  // Log the start of current log scope.
+  LOG_SCOPE(F->getLogContext(), "glow::optimizeFunction");
```
jfix71: Should this come before the call to `optimizeFunctionBeforeLowering()`?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -92,6 +92,11 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
     info.backendName = device.second->getBackendName();
     deviceInfo.push_back(info);
   }
+  // Perform a round of target-independent graph optimizations to remove unused
```
jfix71: If it's just to remove unused nodes then can't you just call DCE directly? Otherwise I would rephrase this to make it clear we're not just removing unused Nodes.

opti-mix: Apparently, it is more than just DCE. DCE alone does not help that much. I'll rephrase.

lib/Optimizer/GraphOptimizer.cpp
opti-mix: No strong opinion here. If we put it before the call, we'll increase the nesting level of scopes. I can change it if you think it is cleaner.

jfix71: Feels to me like it's more in line with expected uses. @ZchiPitt might have a different feeling. NBD either way 🙂

ZchiPitt: I think it is fine here, as it might reflect the actual context here of the order of calling optimize

opti-mix:
@jfix71 
> FYI much of this will be changing after #1641 is complete as it will include a redesign of the graph optimization pipeline.

OK. Note taken.

> I'd be interested to know which optimizations in particular are helping here.

The particular optimizations that help most here are:
- `optimizeBatchNorm` (this is the most important one for ResNet50)
- `sinkCode`
- `DCE`

> If it's just to remove unused nodes then can't you just call DCE directly? Otherwise I would rephrase this to make it clear we're not just removing unused Nodes.

I'll update the comments.

PR #3126, opened 2019-06-17
Properly support Int8ConvRelu and Int8SumRelu

artemrakhov:
We do not add Relu in loader explicitly, because we rely on the Caffe2 protobuf always having scale and offset such, that Relu is not needed. However, it may break. Also, as we go with having more Loaders, we may want more general approach anyway.


artemrakhov:
Updated with a test.

PR #3127, opened 2019-06-18
[NFC] Add more debug prints and comments to BackendUtils.

artemrakhov:


jackm321:
@artemrakhov can we use glog for this?

artemrakhov:
@jackm321: do you propose to switch to `LOG(INFO)`? What is the motivation? I see that `DEBUG_GLOW` is always used together with `llvm::dbgs()`, and never with glog.
I think if we want to switch from one to the other, we should do it everywhere at once.

artemrakhov:
Talked offline, decided to switch to using `LOG(INFO)` here under DEBUG_GLOW macro.

PR #3128, opened 2019-06-18
Avoid zeroing Tensors unnecessarily

### LINKED ISSUE #3037: Don't zero tensors unnecessarily
Tensors are allocated and zeroed in their constructor, but usually we're going to copy/move data into them anyway without reading the initialized values. Remove this unnecessary step.

Probably requires going through the code for Tensor creation, working out which stuff relies on zeroing and adding a manual zero().

nickgg:
Summary: We zero tensors whenever they are `reset()`, and we reset Tensors a lot. Instead we should only zero buffers when we really need to. An example is if we're loading a constant from the protobuf into a Tensor it doesn't matter what the original value was.

Many tests broke when I did this, and most of the time I fixed it by intentionally calling `zero()` on the Tensor. Usually it was just because the test expected the result of working on an uninitialized Tensor to be the same between runs (e.g. when doing a comparison against the Interpreter).

A few cases were legitimate bugs hidden by the fact the Tensors happened to be zero, which i've fixed here:
* Quantization Profiling: The Tensor holding the profile is modified in place and needs to be manually zeroed. In this diff I've done it by marking it as "Trainable" which may be slight abuse of the term (we're training the quantizer to choose good params... ?).
* The Interpreters CrossEntropyLoss calculation for floats was modifying its output Tensor without clearing it first.
* Training: When we lowered the SGD node we weren't setting its PH to be trainable. If we guarantee all trainable Tensors are zeroed this works as intended. (Question: should training Tensors always start zero? Isn't it normal to start in a random state?).
* The `inferNonSquarePaddingConv` test helper used the wrong dimensions.
* ThreadPoolExecutor: This was the hardest to work out. The test was incorrectly verifying that output Tensors were consistent when passed into the DeviceManager but that is definitely not true now that we use a TensorPool, and was only passing because Tensors were initialized zero and were never modified (so all Tensors were the same).

Not fixed but weird:
* inferConvDKKC8 test is using the wrong indexes I think, will call out in a comment.

Documentation: Fixes #3037

Test Plan: unit tests under Release and ASAN, Tested again with the GLOW_DEBUG_TENSOR_INIT define set to test when the initial state is definitely not 0.

Manual testing with resnet-runtime shows a **90% reduction in zeroing buffers**:

before:
```
│I0617 21:46:29.172291 451777984 resnet-runtime.cpp:221] resets: 505
│I0617 21:46:29.172297 451777984 resnet-runtime.cpp:222] zeroes: 558

```
after:
```
│I0617 21:43:12.518196 208471488 resnet-runtime.cpp:221] resets: 505
│I0617 21:43:12.518205 208471488 resnet-runtime.cpp:222] zeroes: 53

```

(I determined this with a global atomic, counting code is not in the diff)

Will follow up later with some more prod like tests, but there are decisions here worth getting comment on early.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -322,7 +322,10 @@ public:
     // If the new size is identical to the allocated size then there is no need
     // to re-allocate the buffer.
     if (type_ == T && getData()) {
-      zero();
+#ifdef GLOW_DEBUG_TENSOR_INIT
+      PseudoRNG rng;
+      init(InitKind::Broadcast, GLOW_DEBUG_TENSOR_INIT, rng);
+#endif
       return;
     }
 
@@ -338,7 +341,11 @@ public:
     assert(size() > 0 && "Tensors must always have positive size.");
     size_t count = size() * type_.getElementSize();
     data_ = reinterpret_cast<char *>(alignedAlloc(count, TensorAlignment));
-    zero(getElementType() == ElemKind::UInt8FusedQTy);
+
+#ifdef GLOW_DEBUG_TENSOR_INIT
+    PseudoRNG rng;
+    init(InitKind::Broadcast, GLOW_DEBUG_TENSOR_INIT, rng);
+#endif
   }
   /// Releases the data buffer and sets the unOwned flag to true. This is useful
   /// for keeping metadata around but not the actual contents.
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1622,7 +1622,7 @@ Function::createQuantizationProfile(PlaceholderBindings &bindings,
   const size_t numberOfBuckets = 2000U;
   auto *histogram = getParent()->createPlaceholder(
       ElemKind::FloatTy, {numberOfBuckets}, "histogram_" + name.str(), false);
-  bindings.allocate(histogram);
+  bindings.allocate(histogram)->zero();
   // Intermediate data used for histogram calculations.
   // Min tensor value seen so far is kept on the first position.
   // Max tensor value seen so far is kept on the second position.
--- a/lib/Graph/PlaceholderBindings.cpp
+++ b/lib/Graph/PlaceholderBindings.cpp
@@ -150,6 +150,12 @@ Tensor *PlaceholderBindings::allocate(Placeholder *P) {
   DCHECK(!map_.count(P)) << "Placeholder with name \"" << P->getName().str()
                          << "\" already registered";
   Tensor *T = new Tensor(P->getType());
+
+  // If this Tensor needs to start zeroed, then zero it.
+  if (P->allocZero()) {
+    T->zero();
+  }
+
   map_[P] = T;
   nameMap_[P->getName()] = P;
   return T;
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -164,6 +164,11 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
       // pointer (which is a valid case if the tensor is empty).
       if (inOnnxBuffer) {
         memcpy(inputTensor->getUnsafePtr(), inOnnxBuffer, onnxBytes);
+        // Pad remaining space with zeroes.
+        memset(inputTensor->getUnsafePtr() + onnxBytes, 0,
+               inputTensor->getSizeInBytes() - onnxBytes);
+      } else {
+        inputTensor->zero();
       }
       ctx->getPlaceholderBindings()->insert(inPhPtr, inputTensor);
     }
--- a/lib/Optimizer/Lower.cpp
+++ b/lib/Optimizer/Lower.cpp
@@ -358,6 +358,7 @@ static void lowerSGDNode(Function *F, CompilationContext &cctx,
   if (momentum > 0.0) {
     Placeholder *Gsum =
         F->getParent()->createPlaceholder(W.getType(), "gsum", false);
+    Gsum->setAllocZero();
 
     auto *momentumSplat = F->createSplat("learningRateSplat", type, momentum);
     auto *GsumMult = F->createMul("GsumMult", momentumSplat, Gsum);
--- a/tests/unittests/BackendTestUtils.cpp
+++ b/tests/unittests/BackendTestUtils.cpp
@@ -583,7 +583,7 @@ void inferNonSquarePaddingConv(Tensor *out, llvm::StringRef kind) {
   auto *filterTensor = bindings.allocate(filter);
   auto FH = filterTensor->getHandle();
   for (size_t i = 0; i < 128; i++)
-    for (size_t j = 0; j < 16; j++) {
+    for (size_t j = 0; j < 32; j++) {
       FH.at({i, 0, 0, j}) = (i + j) / 100.0;
     }
   auto *zeroBias =
@@ -700,6 +700,7 @@ void inferConvDKKC8(Tensor *out, llvm::StringRef kind) {
   auto *filter = mod.createPlaceholder(ElemKind::FloatTy, {192, 3, 3, 32},
                                        "filter", false);
   auto *filterTensor = bindings.allocate(filter);
+  filterTensor->zero();
   auto FH = filterTensor->getHandle();
   for (size_t i = 0; i < 192; i++)
     for (size_t j = 0; j < 3; j++)
--- a/tests/unittests/ThreadPoolExecutorTest.cpp
+++ b/tests/unittests/ThreadPoolExecutorTest.cpp
@@ -82,9 +82,17 @@ public:
       std::unique_ptr<ExecutionContext> inputContext =
           std::move(registeredResult->inputContext);
 
-      if (PlaceholderBindings::compare(
-              context->getPlaceholderBindings(),
-              inputContext->getPlaceholderBindings())) {
+      bool equalInputs = true;
+      for (auto &p : inputContext->getPlaceholderBindings()->pairs()) {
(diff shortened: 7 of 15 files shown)


tests/unittests/BackendTestUtils.cpp
```diff
@@ -700,6 +700,7 @@ void inferConvDKKC8(Tensor *out, llvm::StringRef kind) {
   auto *filter = mod.createPlaceholder(ElemKind::FloatTy, {192, 3, 3, 32},
                                        "filter", false);
   auto *filterTensor = bindings.allocate(filter);
+  filterTensor->zero();
```
nickgg: So here the line below looks wrong:
`          FH.at({i, j, k, k}) = (i + j + k + l) / 200.0;
`
Should that last `k` be an `l` ?

If I remove the zeroing and change it to `l` the test fails, and it's not obvious why.

@nadavrot I think you wrote this one, do you remember the details?

bertmaher:
This is awesome!

What's the code path that triggers so many `reset()`s in the runtime?  Just curious.

nickgg:
I'm not 100% sure in all cases, but do plan to look into it. We call reset just to be safe in a few places I think.

Honestly I'm kind of stunned we're still doing 50+ zeros() running resnet50 - I think it might be because some weights are marked as trainable and I'm using that to determine whether to zero.

lib/Graph/Graph.cpp
```diff
@@ -1610,13 +1610,13 @@ Function::createQuantizationProfile(PlaceholderBindings &bindings,
   // TODO: this size is going to be refined. Just a placeholder now.
   const size_t numberOfBuckets = 2000U;
   auto *histogram = getParent()->createPlaceholder(
-      ElemKind::FloatTy, {numberOfBuckets}, "histogram_" + name.str(), false);
+      ElemKind::FloatTy, {numberOfBuckets}, "histogram_" + name.str(), true);
   bindings.allocate(histogram);
```
jfix71: Why not just zero here instead of setting trainable?

nickgg:
I cant repro these test fails locally so I'm just going to push a lot of commits to CI. Sorry in advance!

tests/unittests/ThreadPoolExecutorTest.cpp
```diff
@@ -155,13 +165,14 @@ class TestDeviceManager final : public runtime::DeviceManager {
   }
 
 private:
-  /// This struct wraps all of the data needed to reply to a runFunction() call.
-  /// It exists so that that all of these things can be stored in one map.
+  /// This struct wraps all of the data needed to reply to a runFunction()
```
jfix71: This/the rest of the comments didn't actually change, right? `clang-format` issue?

nickgg: Yeah just the auto clang-formatter, sorry for the noise.

nickgg:
* QuantizationProfiling was profiling output Tensors, which are uninitialized, and we're generating profiles with a bunch of NANs in them - tripping an assert. Fixed by excluding Placeholders who's user is a `SaveNode` from `profileQuantization`

nickgg:
* Fixed TextTranslator by manually zeroing inputs other than the tokens.

lib/Graph/PlaceholderBindings.cpp
```diff
@@ -150,6 +150,11 @@ Tensor *PlaceholderBindings::allocate(Placeholder *P) {
   DCHECK(!map_.count(P)) << "Placeholder with name \"" << P->getName().str()
                          << "\" already registered";
   Tensor *T = new Tensor(P->getType());
+
+  /// Trainable PH's need to start zeroed.
+  if (P->isTraining()) {
```
SplitInfinity: They shouldn't have to but I suspect you might have been forced to add this to make sure the `gsum` PH of `SGDNode` starts out zeroed.

lib/Optimizer/Lower.cpp
```diff
@@ -357,7 +357,7 @@ static void lowerSGDNode(Function *F, CompilationContext &cctx,
   // OptimizationStochasticGradientDescent/
   if (momentum > 0.0) {
     Placeholder *Gsum =
-        F->getParent()->createPlaceholder(W.getType(), "gsum", false);
+        F->getParent()->createPlaceholder(W.getType(), "gsum", true);
```
SplitInfinity: This needs to be initialized to zero but it is not trainable. It is unlikely someone would ever initialize it in another way because `gsum` is supposed to be a geometric series of previous weight updates with `momentum` as the ratio between successive terms.

@putivsky ran into something similar during his work on loading models for training with the mean/variance inputs of `SpatialBN`. He might have some ideas on the best fix.

nickgg: I mean the only other option is to add another flag to Placeholder. I started there but we only used it in one place so I figured I'd try to reduce extra code. Not super fussed though.

nickgg:
Added a new flag to placeholder rather than overload `isTraining()`

lib/Optimizer/Lower.cpp
SplitInfinity: We use `Placeholder::isTraining` during differentiation to determine which nodes should have SGD nodes and weight updates added. That didn't matter here because the fact that an `SGDNode` is being lowered means that the graph has already been differentiated.

tests/unittests/ThreadPoolExecutorTest.cpp
```diff
@@ -82,9 +82,13 @@ class TestDeviceManager final : public runtime::DeviceManager {
       std::unique_ptr<ExecutionContext> inputContext =
           std::move(registeredResult->inputContext);
 
-      if (PlaceholderBindings::compare(
-              context->getPlaceholderBindings(),
-              inputContext->getPlaceholderBindings())) {
+      bool equalInputs = true;
+      for (auto &p : inputContext->getPlaceholderBindings()->pairs()) {
+        Tensor *CT = context->getPlaceholderBindings()->get(p.first);
```
SplitInfinity: Nit: If `!CT` you can set `equalInputs` to `false` and `break`

tests/unittests/ThreadPoolExecutorTest.cpp
```diff
@@ -93,7 +97,13 @@ class TestDeviceManager final : public runtime::DeviceManager {
 
         for (auto p : registeredResult->resultContext->getPlaceholderBindings()
                           ->pairs()) {
-          context->getPlaceholderBindings()->get(p.first)->assign(p.second);
+          Tensor *T = context->getPlaceholderBindings()->get(p.first);
+          if (T) {
+            T->assign(p.second);
+          } else {
+            context->getPlaceholderBindings()->insert(p.first,
```
SplitInfinity: When is this branch taken? I was under the impression that all `Placeholders` used by a function have to be mapped in the `PlaceholderBindings` that it is executed with.

tests/unittests/ThreadPoolExecutorTest.cpp
```diff
@@ -355,7 +366,8 @@ class ExecutorTestBuilder final {
     }
 
     // Iterate through inputs and outputs and:
-    // 1) Create Placeholders and Tensors for inputs/output names that have not
+    // 1) Create Placeholders and Tensors for inputs/output names that have
+    // not
```
SplitInfinity: Weird line break here

tests/unittests/ThreadPoolExecutorTest.cpp
nickgg: You are right here, it's the input/result context that don't need to have all Placeholders in them.

tests/unittests/ThreadPoolExecutorTest.cpp
nickgg: Yeah my auto clang-formatter did it, but I'm wondering if I might have missed a `}` when working and it reformatted based on bad indentation? I'll look into it.

nickgg:
* Change caught a bug in Onnxifi during adjustInputs - we were copying partial input tensors but leaving the rest of the Tensor unchanged. Change Onnxifi to always zero the remaining space when using partial inputs.

nickgg:
Test fail looks like a POCL thing?

rdzhabarov:
Yes, ignore for now. If I'm not able to fix it today/tomorrow I'll disable it.

include/glow/Base/Tensor.h
```diff
@@ -322,7 +322,10 @@ class Tensor final {
     // If the new size is identical to the allocated size then there is no need
     // to re-allocate the buffer.
     if (type_ == T && getData()) {
-      zero();
+#ifdef GLOW_DEBUG_TENSOR_INIT
```
bertmaher: This is a great idea, do you think we should always do it in debug mode (like `#ifndef NDEBUG`)?  Your call either way.

lib/Onnxifi/Base.cpp
```diff
@@ -164,6 +164,11 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
       // pointer (which is a valid case if the tensor is empty).
       if (inOnnxBuffer) {
         memcpy(inputTensor->getUnsafePtr(), inOnnxBuffer, onnxBytes);
+        // Pad remaining space with zeroes.
+        memset(inputTensor->getUnsafePtr() + onnxBytes, 0,
+               inputTensor->getSizeInBytes() - onnxBytes);
```
bertmaher: Great catch here, thank you!

include/glow/Base/Tensor.h
nickgg: Good question. I actually started that way, but I think it covers some heap sentinels written by free in debug. No strong opinions either way, this is a bit more flexible but we don't really want to have to think about this.

PR #3130, opened 2019-06-18
Remove GLOW_ASSERTs from OpenCL backend

jackm321:
Summary:
Remove all of `GLOW_ASSERT`s in OpenCL backend and replace them with glog `CHECK*`s.
Also replace the asserts with `DCHECK*`s

Documentation:
n/a

Test Plan:
`ninja check`


opti-mix:
Nice!

PR #3131, opened 2019-06-18
Support ONNX Tile operation

Max-K-cadence:
Summary:
This patch adds support for ONNX Tile operation.

Test Plan:
Added unit tests

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -1085,6 +1085,37 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
   return llvm::Error::success();
 }
 
+llvm::Error ONNXModelLoader::loadTile(const ONNX_NAMESPACE::NodeProto &op,
+                                      const ArgumentDictionaryTy &dict) {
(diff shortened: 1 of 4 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1085,6 +1085,37 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
   return llvm::Error::success();
 }
 
+llvm::Error ONNXModelLoader::loadTile(const ONNX_NAMESPACE::NodeProto &op,
```
rdzhabarov: C2 loader already have a logic for loading `Tile` ops. Have you considered merging those in CommonOperatorLoader?

Max-K-cadence: C2 tile has a different API than ONNX's. It takes number of repeats and Axis as arguments and only expands across 1 dimension. ONNX Tile does not take axis and takes repeats along each dimension. So I don't think we can make it in common code given different inputs semantics. :)

rdzhabarov: fair enough

PR #3132, opened 2019-06-18
ChannelwiseQuantizedConvolutionNode impl

jackm321:
Summary:
* Create an interpreter implementation for ChannelwiseQuantizedConvolutionNode
* Add a lowering pass that converts ChannelwiseQuantizedConvolutionNode to a concatination of regular quantized convolutions

Differential Revision: D15838460

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -292,6 +292,16 @@ public:
     return unownedTensor;
   }
 
+  /// This is the same as \ref getUnowned() but it produces an owned tensor
+  /// instead. \returns owned tensor copied from the data buffer of the current
+  /// tensor but having different dimensions \p dims. \p offsets represents an
+  /// optional offset into the tensor representing the location of the first
+  /// element to start a subview from.
+  Tensor getOwnedSlice(llvm::ArrayRef<size_t> dims,
+                       llvm::ArrayRef<size_t> offsets = {}) const {
+    return getUnowned(dims, offsets).clone();
+  }
+
   /// Reset the shape and type of this tensor to match the shape and type of
   /// \p other.
   void reset(const Tensor *other) { reset(other->getType()); }
--- a/lib/Backends/Interpreter/Interpreter.cpp
+++ b/lib/Backends/Interpreter/Interpreter.cpp
@@ -136,6 +136,20 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                {ConvolutionNode::BiasIdx}) &&
            (NI.getInElemTy(ConvolutionNode::BiasIdx) == ElemKind::Int32QTy);
 
+  case Kinded::Kind::ChannelwiseQuantizedConvolutionNodeKind:
+    return (NI.getInElemTy(ChannelwiseQuantizedConvolutionNode::InputIdx) ==
+            ElemKind::Int8QTy) &&
+           (NI.getInElemTy(ChannelwiseQuantizedConvolutionNode::FilterIdx) ==
+            ElemKind::Int8QTy) &&
+           (NI.getInElemTy(ChannelwiseQuantizedConvolutionNode::BiasIdx) ==
+            ElemKind::FloatTy) &&
+           (NI.getInElemTy(ChannelwiseQuantizedConvolutionNode::ScalesIdx) ==
+            ElemKind::FloatTy) &&
+           (NI.getInElemTy(ChannelwiseQuantizedConvolutionNode::OffsetsIdx) ==
+            ElemKind::Int32ITy) &&
+           (NI.getOutElemTy(ChannelwiseQuantizedConvolutionNode::ResultIdx) ==
+            ElemKind::Int8QTy);
+
   case Kinded::Kind::Convolution3DNodeKind:
     if (!NI.getInTy(Convolution3DNode::InputIdx)->isQuantizedType()) {
       return NI.allInputsAndOutputsHaveSameElemKind(
@@ -395,6 +409,7 @@ bool Interpreter::shouldLower(const Node *N) const {
   switch (N->getKind()) {
   case Kinded::Kind::ConvolutionNodeKind:
   case Kinded::Kind::SparseLengthsSumNodeKind:
+  case Kinded::Kind::ChannelwiseQuantizedConvolutionNodeKind:
     return false;
   default:
     return true;
--- a/lib/Optimizer/Lower.cpp
+++ b/lib/Optimizer/Lower.cpp
@@ -647,6 +647,94 @@ static void lowerGroupConvolutionNode(Function *F, CompilationContext &cctx,
   replaceAllUsesOfWith(cctx.loweredInfoMap, BNG.getResult(), result);
 }
 
+static void lowerChannelwiseQuantizedConvolutionNode(
+    Function *F, CompilationContext &cctx,
+    const ChannelwiseQuantizedConvolutionNode &CQC) {
+  // ChannelwiseQuantizedConvolutionNode can be represented as a Concatenation
+  // of smaller dimension quantized Convolutions each with their own qparams.
+  // Input channels will be divided into equal groups of consecutive channels.
+  // These will be separately convolved each with its own filter and bias.
+  // This will result in 4 * Group + 1 nodes.
+
+  // Only lowering of groupwise, not channelwise is supported so far.
+  if (!CQC.getGroupwise()) {
+    return;
+  }
+
+  llvm::ArrayRef<unsigned_t> kernels = CQC.getKernels();
+  llvm::ArrayRef<unsigned_t> pads = CQC.getPads();
+  llvm::ArrayRef<unsigned_t> strides = CQC.getStrides();
+  unsigned_t group = CQC.getGroup();
+  auto in = CQC.getInput();
+
+  Constant *filter = llvm::cast<Constant>(CQC.getFilter());
+  Constant *bias = llvm::cast<Constant>(CQC.getBias());
+  Constant *scales = llvm::cast<Constant>(CQC.getScales());
+  Constant *offsets = llvm::cast<Constant>(CQC.getOffsets());
+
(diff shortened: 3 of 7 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -1257,6 +1257,7 @@ bool HabanaBackend::isOpSupported(const NodeInfo &NI) const {
     case Kinded::Kind::TileNodeKind:
     case Kinded::Kind::ConcatNodeKind:
     case Kinded::Kind::TransposeNodeKind:
+    case Kinded::Kind::ChannelwiseQuantizedConvolutionNodeKind:
```
beicy: Why Habana backend need to be updated in this PR?

include/glow/Base/Tensor.h
```diff
@@ -292,6 +292,16 @@ class Tensor final {
     return unownedTensor;
   }
 
+  /// This is the same as getOwnedSlice but it produces an owned tensor instead.
```
jfix71: nit: I think you meant `getUnowned()` here? Also could use `\ref`.

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -136,6 +136,9 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                {ConvolutionNode::BiasIdx}) &&
            (NI.getInElemTy(ConvolutionNode::BiasIdx) == ElemKind::Int32QTy);
 
+  case Kinded::Kind::ChannelwiseQuantizedConvolutionNodeKind:
+    return true;
```
jfix71: I think you should expand on this, no? For example it's possible we could create a `ChannelwiseQuantizedConvolutionNode` and then the FP16 converter could modify some input to FP16. See RWQ-SLWS -- it's a very specialized Node but we still check specific element kinds on the inputs/outputs.

lib/Optimizer/Lower.cpp
```diff
@@ -647,6 +647,93 @@ static void lowerGroupConvolutionNode(Function *F, CompilationContext &cctx,
...
+    slicedFilterTensor.setType(&quantizedFilterType);
+    Constant *slicedFilterConst = M->createConstant(
+        strFormat("%s_filter", CQC.getName().data()), slicedFilterTensor);
+
+    // Create bias sliced Constant. Bias's scale should always be inputScale *
+    // filterScale.
+    TensorQuantizationParams tqp;
+    tqp.offset = 0;
+    tqp.scale = inSlice->getInput().getType()->getScale() * filterScale;
+    Tensor slicedBiasTensor =
```
jfix71: This feels like something we should add as a graph optimization (i.e. `Slice(Constant) -> Constant`). Then you could just add Slice/Quantize nodes here.

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
...
+      /* pads */ {0, 0, 0, 0},
+      /* group */ groups);
+
+  SaveNode *saveNode = F->createSave("save", channelwiseQuantizedConvNode);
+
+  auto backend = MockBackend();
+  CompilationContext cctx;
+  lower(F, cctx, &backend);
+
+  // verify lowered graph
```
jfix71: nit: Capitalize and period.

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
   llvm::sys::fs::remove(filePath);
 }
 
+TEST(Graph, ConvChannelwiseQuantizedLower) {
+  // expect NHWC
```
jfix71: nit: Probably move this to just above `inputPH`? Also: Capitalize and period.

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
   llvm::sys::fs::remove(filePath);
 }
 
+TEST(Graph, ConvChannelwiseQuantizedLower) {
```
jfix71: Can you add doxygen?

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
...
+
+  constexpr size_t groups = 2;
+  constexpr size_t depth = 4;
+  constexpr size_t inputChannels = 4;
+
+  Node *inputPH = M.createPlaceholder(
+      ElemKind::Int8QTy, {1, 10, 10, inputChannels}, /* scale */ 1.0,
+      /* offset */ 0, "input", true);
+
+  // Filter
```
jfix71: Not sure these comments add value.

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
...
+  auto concatNode = llvm::dyn_cast<ConcatNode>(saveNode->getInput().getNode());
+  ASSERT_TRUE(concatNode != nullptr);
+
+  auto concatInputs = concatNode->getInputs();
+  EXPECT_EQ(concatInputs.size(), groups) << "Should have one conv per group";
+
+  for (size_t i = 0; i < concatInputs.size(); ++i) {
+    auto convNode = llvm::dyn_cast<ConvolutionNode>(concatInputs[i].getNode());
+    ASSERT_TRUE(concatNode != nullptr);
+    // check that each conv has the expected filter qparams
```
jfix71: nit: Capitalize and period.

tests/unittests/GraphTest.cpp
```diff
@@ -158,6 +158,85 @@ TEST(Graph, simpleTestConvCustomLower) {
...
+    auto *filterTy = convNode->getFilter().getType();
+    ASSERT_TRUE(filterTy != nullptr);
+    EXPECT_TRUE(filterTy->isQuantizedType());
+    EXPECT_EQ(filterTy->getScale(), float(1 + 2 * i));
+    EXPECT_EQ(filterTy->getOffset(), 2 + 2 * i);
+
+    // check that each conv has the expected output qparams
```
jfix71: nit: Capitalize and period.

tests/unittests/OperatorTest.cpp
```diff
@@ -3853,6 +3853,81 @@ TEST_P(OperatorTest, GroupConvolution) {
   EXPECT_FLOAT_EQ(result.at({0, 1, 0, 5}), (13 + 14 + 15 + 16) * 100000);
 }
 
+TEST_P(OperatorTest, GroupwiseQuantizedConvolution) {
```
jfix71: Doxygen

tests/unittests/OperatorTest.cpp
```diff
@@ -3853,6 +3853,81 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+  ENABLED_BACKENDS(Interpreter);
```
jfix71: Shouldn't CPU/OpenCL work here since it should be lowered correctly? What about Habana?

lib/Optimizer/Lower.cpp
jfix71: I don't think we have it yet -- I think we have it for Splat but not Constant. Don't need it for this PR necessarily, but perhaps file an issue for a new opt and mention there that we could then remove the slicing here in the lowering pass so we don't forget.

tests/unittests/OperatorTest.cpp
```diff
@@ -3853,6 +3853,83 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+  for (size_t i = 0; i < 6; i++) {
+    for (size_t j = 0; j < 4; j++) {
+      filterT.getHandle<int8_t>().at({i, 0, 0, j}) = (i % 2) + 1;
+    }
+  }
+  auto *filter = mod_.createConstant("filter", filterT);
+
+  auto biasT = Tensor(ElemKind::FloatTy, {6});
+  biasT.zero();
+  auto *bias = mod_.createConstant("bias", biasT);
```
jfix71: nit: when we create constants like this and below, we are unnecessarily creating two copies of the Tensor, because `createConstant()` here makes a copy of the passed in tensor (e.g. `biasT`). You could instead create the Constant first with a Type, and then get the Constant's payload and modify it.

lib/Optimizer/Lower.cpp
```diff
@@ -647,6 +647,93 @@ static void lowerGroupConvolutionNode(Function *F, CompilationContext &cctx,
...
+  for (unsigned_t groupId = 0; groupId < group; groupId++) {
+    float filterScale = scalesHandle.raw(groupId);
+    int32_t filterOffset = offsetsHandle.raw(groupId);
+
+    SliceNode *inSlice =
+        F->createSlice(CQC.getName(), in, {0, 0, 0, groupId * inCperG},
+                       {idim.n, idim.h, idim.w, (groupId + 1) * inCperG});
+
+    // Create quantized filter sliced Constant.
+    Tensor slicedFilterTensor = filter->getPayload().getOwnedSlice(
```
jfix71: I don't think we need an owned slice here or below since you're then copying over the slice anyway into the Constant via `createConstant()`, right?

tests/unittests/OperatorTest.cpp
jackm321: std::moved them

lib/Optimizer/Lower.cpp
jackm321: Ok I std::moved the tensors to the constants now so there's no additional copy. I tried using unowned tensors, put them into a constant, get mutable payload reference then modify that tensor and it looks more confusing than it's worth imo

lib/Optimizer/Lower.cpp
jackm321: Looks like it already exists https://github.com/pytorch/glow/issues/2168

PR #3135, opened 2019-06-18
Use partial unpadded tensors when supported by backend

bertmaher:
Summary:
This is take 2 on partial tensor support.  There are two components:

1. The backend needs to declare its support for partial tensors.  Since the
backend may have to do padding in some cases, partiality needs to be opt-in.

2. The Habana backend is our proof-of-concept here; it needs to check the types
of consuming operators to determine which are candidates for partial transfers.
Currently, only the indices and weights of SparseLengths(Weighted)Sum are
candidates.

Differential Revision: D15889467

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -114,8 +114,6 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
   TRACE_EVENT_SCOPE(traceContext, "Onnxifi::setIOAndRun");
   TRACE_EVENT_SCOPE_NAMED(traceContext, "adjustInputs", aiEvent);
 
-  size_t totalInputOnnxTensorSize = 0;
-  size_t totalInputGlowTensorSize = 0;
   // Create tensors for input placeholders
   for (unsigned i = 0; i < inputsCount; ++i) {
     const auto &inOnnxTensor = inputDescriptors[i];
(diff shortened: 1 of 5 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -141,36 +139,32 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+    } else if (backendPtr_->getBackendId()
+                   ->getBackend()
+                   .supportsPartialTensors() &&
+               inOnnxBuffer && inOnnxTensorSize > 0) {
+      // We have a partial input buffer.  Create a padded unownded tensor that
+      // remembers the actual size of the input.
+      ctx->getPlaceholderBindings()->insert(
+          inPhPtr, Tensor(inOnnxBuffer, inPhPtr->getType(), onnxBytes));
     } else {
+      // If input onnxTensorDescriptor has a NULL buffer pointer, which is a
```
rdzhabarov: this comment seems to be moved to the wrong position here.

lib/Onnxifi/Base.cpp
```diff
@@ -141,36 +139,32 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+      // valid case for empty tensor, skip copying.
       Tensor *inputTensor = tensorPool_.get(inPhPtr->getType());
       DCHECK(inputTensor) << "Tensorpool tensor not found for input "
```
rdzhabarov: since this is only debug check we'd need to handle nullptr in inputTensor gracefully, otherwise we'd get segfault on line 163.
We should return appropriate ONNXIFI return code in case of no available tensors in a pool.

bertmaher: I'm never sure whether we should check for programming errors at runtime in release builds.  Is it actually possible to get a nullptr back from `tensorPool_` in the absence of bugs?

nickgg: Theres a configuration option for the TensorPool where it wont allocate new tensors and will return null. We don't use it in Onnxifi though so this should always be valid.

PR #2947, opened 2019-05-20
[Partitioner] Add options to enable logging Partitioner.

beicy:
Summary:

Documentation:

Test Plan:
Ninja test.
```
wangm-mbp:buildR wangm$ ./bin/text-translator -m "en2gr" -backend=CPU -cpu-memory=488000 -num-devices=2 -log-partition -dump-partition<<< "I love music ."
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0618 13:02:46.204908 335373760 Partitioner.cpp:1034] The number of partitions is : 2, and the DAG is dumped into DAG.dot file.
I0618 13:02:46.205582 335373760 Partitioner.cpp:73] Writing dotty graph for DAG after graph partitioning: DAG.dot
I0618 13:02:46.205896 335373760 Partitioner.cpp:1043] 	 Partition 0:
		 Name :	en2gr_part1
		 BackendKind :	CPU
		 Memory :	499281972
		 LogicalDeviceIDs :	0
Writing dotty graph for Function to: partitionLogicalID0__en2gr_part1__CPU.dot
I0618 13:02:46.214095 335373760 Partitioner.cpp:1043] 	 Partition 1:
		 Name :	en2gr_part2
		 BackendKind :	CPU
		 Memory :	406380524
		 LogicalDeviceIDs :	1
Writing dotty graph for Function to: partitionLogicalID1__en2gr_part2__CPU.dot
Enter a sentence in English to translate to German: ich liebe Musik .
```

[Optional Fixes #issue]

Please see a detailed explanation of how to fill out the fields in the relevant sections in PULL_REQUEST.md.

PR #3136, opened 2019-06-18
[Partitioner] Add debug output to printout resulting partitions.

artemrakhov:
It is very helpful for debug to know what are resulting partitions. This PR adds code to dump partitions DAG, and every individual partition by itself.

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -17,11 +17,14 @@
 #include "glow/Partitioner/Partitioner.h"
 #include "glow/Support/Support.h"
 
+#include "glow/Support/Debug.h"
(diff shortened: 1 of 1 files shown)


opti-mix:
@artemrakhov The partition printing logic seems to be covered by #2947 already.

beicy:
Thanks for working on this. I think this PR https://github.com/pytorch/glow/pull/2947 has the same purpose. In this PR, the logs are enabled by default. The dumpDAG() belongs to partitioner log. But dumping the dot file for each partition is optional and controlled by  llvm options. The reason is, the DAG won't be too large and it is OK to dump it by default.

artemrakhov:
OK, I will drop the PR then.

PR #3137, opened 2019-06-19
[compilationLog]: Refine the locations of LOG_SCOPE

ZchiPitt:
Summary:
Refine the locations of LOG_SCOPE inside each optimize function.

- Previously, we log the scope at the beginning of each optimization function. But there are multiple iterations (iterate over all nodes) of this optimization going inside one scope, which makes the produced log not entirely showing the changes inside one iteration.
- I propose to move LOG_SCOPE into each iteration of every optimization, such that the produced scoped log blocks are in a finer granularity.

Test Plan:
test with ./bin/image-classifier tests/images/imagenet/cat_285.png -image-mode=0to1 -m=googlenet_v1_slim/googlenet_v1_slim.onnx -model-input-name=input:0 -image-layout=NHWC -backend=CPU -dump-compilation-log=true

The output log of optimizeReshape is shown below:
Before refining:
Screen Shot 2019-06-18 at 7 25 35 PM
After refining:
Screen Shot 2019-06-18 at 6 59 00 PM

We can see that the scope of optimizeReshape break into small chunks now.

### DIFF
--- a/lib/Backends/CPU/Transforms.cpp
+++ b/lib/Backends/CPU/Transforms.cpp
@@ -121,10 +121,10 @@ static Node *optimizeCPUMaxSplat(MaxNode *MN, Function *F) {
 
 bool CPUBackend::transformPostLowering(Function *F,
                                        CompilationContext &) const {
-  LOG_SCOPE(F->getLogContext(), "CPUBackend::transformPostLowering")
-
   bool changed = false;
   for (auto &node : F->getNodes()) {
+    LOG_SCOPE(F->getLogContext(), "CPUBackend::transformPostLowering")
+
     // Try to replace generic convolution with cpu-optimized version.
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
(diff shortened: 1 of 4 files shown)


nickgg:
@ZchiPitt what's the motivation for this change? The log will end up much larger, but we already log the scope on every NODE_INPUT_CHANGE. 

In the after log it looks like we write the scope 3 times for every input we change? Seems wasteful.

ZchiPitt:
@nickgg  Thanks for bringing this concern.
- As we know, at the end of most transformations, replaceAllUsesOfWith() is called, which invokes all users of caller to change their operand. In this case, if the replaced NodeValue has more than one user, the LOG_NODE_CHANGE will be produced more than one time. So we can't use LOG_NODE_CHANGE to decide if a transformation on one node is finished or not.
- There was no specific way to determine when we enter/exit one transformation. Previously the scope is for the entire function, which is too big. So I decide to narrow the scope to the transformation of each node.

ZchiPitt:
will submit later

PR #3138, opened 2019-06-19
Reduce PCI transfers of partial transfers

bertmaher:
Summary:
Synapse 0.1.9 has a secret feature to allow enqueues of partial
tensors.  Let's take advantage of this for embedding indices/weights.

Reviewed By: rdzhabarov

Differential Revision: D15895782


bertmaher:
This PR is stacked on #3135.

PR #3139, opened 2019-06-19
[NFC][RecSys] Make parameters tunable from the command line. Rename vars to lowerCamelCase.

jfix71:


tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -446,11 +504,13 @@ class RecommendationSystemTest : public BackendTest {
                          0.001);
     NodeValue bottom_MLP;
```
opti-mix: Nit: Not introduced in this PR, but why names with underscores, i.e. why `bottom_MLP` instead of `bottomMLP`?

jfix71: No idea 🙂 I can change them all here since it's NFC anyway.

PR #3140, opened 2019-06-19
[OpenCL] Add TraceEvent for compiling custom convolution kernels

SplitInfinity:
**Description**
This commit adds a TraceEvent that covers the time spent during
OpenCLFunction::execute() on compiling custom versions of
convolution kernels according to the parameters of the particular
convolutions present in the function. Depending on the OpenCL
implementation, this time can manifest as mysterious large gaps in the trace file.

**Test Plan**
Generated traces for `tracing-compare` example with AMD OpenCL.

Without this PR:
Screen Shot 2019-06-19 at 3 27 16 PM

With this PR:
Screen Shot 2019-06-19 at 3 21 26 PM

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -197,7 +197,8 @@ OpenCLFunction::createProgram(const std::string &source,
   if (program) {
     return program;
   }
-  // Create a new compiled program.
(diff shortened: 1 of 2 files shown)


SplitInfinity:
Please stand by while I get some screenshots on an OpenCL platform where this problem shows up much more clearly.

PR #3141, opened 2019-06-19
General Constant Folding for Onnx and Caffe2 Operators in Loader

ksaurabh-cadence:
**Summary**
- Add constant folding of operators to ONNXModelLoader:
- Add constant folding of operators to Caffe2Loader
- Made Constant folding of Operators an optional feature controlled by command line option const-fold-ops
- Added constant folding test based on gather and reshape to OnnxImporterTests and Caffe2ImporterTest
- Add constant folding test to OnnxImporterTests

**Documentation**
First check if the folding is enabled. If yes, check if current operator is foldable. If yes get the output of a graph where all the inputs are constant and current operator is the only operator in the graph. Run the graph and instead of loading the original opreator load the constant operator with value calculated. When done sequentially during the loader this can fold not just a single operator but a whole subgraph as long as all the inputs to the subgraphs are constants.

**Optional Fixes**
- Fix an issue with batchboxCox batchBoxCox.onnxtxt (needed for general onnx operator folding test)

**Test Plan**
(Test constant folding of ONNX operators with ONNXImporterTest)
First replace placeholders in the original onnx model with constants. After loading the graph get the graph output values without running the GraphProto. If this output is same as output of original graph after running constant folding, the test passes.

A simple network in caffe2 and onnx which requires constant fold to work:
The test gatherOpConstantFoldingAndReshape in ONNxImportertest and Caffe2ImporterTest verifies that Gather gets constant-folded, so that the argument of the reshape becomes constant.

Related to #2168

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -46,6 +46,15 @@ using PlaceholderList = std::list<Placeholder *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 /// Map from original Nodes to cloned Nodes.
 using NodeMap = llvm::DenseMap<Node *, Node *>;
+/// State of a function. This can be used to control optimizations which depend
+/// on the state of the Function. This is a temporary workaround until GH Issue
+/// #3213 is complete.
+enum class FunctionState {
+  /// Indicates that the function has been created but not completely loaded.
+  FuncCreated,
+  /// Indicates that the function has been completely loaded.
+  FuncLoaded,
+};
 
 class Module final {
   /// Stores the functions in the module.
@@ -239,15 +248,24 @@ class Function final : public Named {
   /// The log context associated with this function.
   std::shared_ptr<LogContext> logCtx_;
 
+  /// The state of this function.
+  FunctionState state_;
+
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {
+      : Named(Name), parent_(parent), state_(FunctionState::FuncCreated) {
     logCtx_ = std::make_shared<LogContext>();
     logCtx_->setParent(this);
     logCtx_->loadModuleLogContext();
   }
 
   ~Function();
+
+  /// Sets the state of the function.
+  void setState(FunctionState state) { state_ = state; }
+
+  /// Gets the state of the function.
+  FunctionState getState() { return state_; }
 
   std::string getFilename() { return getName().rsplit('/').second.str(); }
 
--- a/include/glow/Importer/Caffe2ModelLoader.h
+++ b/include/glow/Importer/Caffe2ModelLoader.h
@@ -60,6 +60,9 @@ class Caffe2ModelLoader
   /// in the network.
   llvm::Error loadOperator(const caffe2::OperatorDef &op);
 
+  /// \returns True if the operator \p op is successfully folded.
+  llvm::Expected<bool> foldOperator(const caffe2::OperatorDef &op);
+
   /// Load the Conv or ConvRelu operators.
   llvm::Error loadConv(const caffe2::OperatorDef &op,
                        ArgumentDictionaryTy &dict);
@@ -106,6 +109,13 @@ class Caffe2ModelLoader
                     llvm::Error *errPtr = nullptr);
 
   friend class ONNXIFIModelLoader;
+
+  /// \returns success if the folding of operator \p op in the loader
+  /// \p loader is successful. The folding utility uses temporary
+  /// loader \p tmpLoader, and associated temporary function \p F.
+  template <class LoaderType, class OpType>
+  friend llvm::Error constantFoldInLoader(Function *F, LoaderType &tmpLoader,
+                                          LoaderType *loader, const OpType &op);
 
 public:
   /// Loads the caffe2 model that's represented by a network description file,
--- a/include/glow/Importer/ONNXModelLoader.h
+++ b/include/glow/Importer/ONNXModelLoader.h
@@ -56,6 +56,9 @@ class ONNXModelLoader
   /// in the network. \returns Error if operator \p op cannot be loaded.
   llvm::Error loadOperator(const ONNX_NAMESPACE::NodeProto &op);
 
+  /// \returns True if the operator\ op is successfully folded.
+  llvm::Expected<bool> foldOperator(const ONNX_NAMESPACE::NodeProto &op);
+
   /// ONNX model ir_version;
   size_t irVersion_;
 
@@ -194,6 +197,13 @@ protected:
                   bool loadInputsAsPlaceholders, llvm::Error *errPtr = nullptr);
 
   friend class ONNXIFIModelLoader;
+
+  /// \returns success if the folding of operator \p op in the loader
+  /// \p loader is successful. The folding utility uses temporary
+  /// loader \p tmpLoader, and associated temporary function \p F.
+  template <class LoaderType, class OpType>
+  friend llvm::Error constantFoldInLoader(Function *F, LoaderType &tmpLoader,
+                                          LoaderType *loader, const OpType &op);
 
 public:
   /// Creates a ONNX model loader to build \p F.
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -18,7 +18,9 @@
 #define GLOW_IMPORTER_PROTOBUFLOADER_H
 
 #include "glow/Base/Tensor.h"
+#include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/Graph/Graph.h"
+#include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
 
 #include "glow/Support/Error.h"
 #include "llvm/ADT/ArrayRef.h"
@@ -37,6 +39,12 @@
 #define MAX_PROTO_SIZE 0x7FFFFFFF
 
 namespace glow {
+
+/// Enables or disables constant-folding of Loader Ops with \p flag.
+void setConstantFoldLoaderOpsFlag(bool flag);
+
+/// Returns true if constant-folding for loader Ops is enabled.
+bool getConstantFoldLoaderOpsFlag();
 
 /// Returns true iff all elements of \p a are the same.
 bool isArrayConstant(const llvm::ArrayRef<size_t> a);
@@ -181,7 +189,61 @@ public:
   /// \returns the Placeholder for the external output with \p name.
   /// \pre outputVarsByName_.find(name) != outputVarsByName_.end()
   llvm::Expected<Placeholder *> getOutputByName(llvm::StringRef name) const;
+
+  /// \returns True if the operator with name \p typeName having input node
+  /// list as \p inputs is constant foldable.
+  bool isConstantFoldable(llvm::ArrayRef<NodeValue> inputs,
+                          std::string typeName) const;
 };
+
+/// \returns success if the folding of operator \p op in the loader
+/// \p loader is successful. The folding utility uses temporary
+/// loader \p tmpLoader, and associated temporary function \p F.
+template <class LoaderType, class OpType>
+llvm::Error constantFoldInLoader(Function *F, LoaderType &tmpLoader,
+                                 LoaderType *loader, const OpType &op) {
+  PlaceholderBindings bindings;
+  std::vector<Tensor *> outTensors;
+  Module *mod = F->getParent();
+
+  // Register the constant inputs to the current op with the constant folding
+  // loader.
+  for (unsigned i = 0; i < op.input_size(); i++) {
+    Constant *tmpConst = mod->getConstantByName(op.input(i));
+    RETURN_ERR_IF_NOT(tmpConst, "No constant found");
+    tmpLoader.nodeValueByName_[op.input(i)] = tmpConst->getOutput();
+  }
+
+  // Using the loader to load the current operator.
+  RETURN_IF_ERR(tmpLoader.loadOperator(op));
+
+  // To collect the folded outputs allocate and add save nodes to the folding
+  // function.
+  for (int i = 0; i < op.output_size(); i++) {
+    const auto &outputName = op.output(i);
+    NodeValue r;
+    ASSIGN_VALUE_OR_RETURN_ERR(r, tmpLoader.getNodeValueByName(outputName));
+    SaveNode *SN = F->createSave("save_" + outputName, r);
+    auto *result = bindings.allocate(SN->getPlaceholder());
+    outTensors.push_back(result);
+  }
+
+  // Evaluate the constant outputs using interpreter backend.
+  std::unique_ptr<Backend> backend(createBackend("Interpreter"));
+  CompilationContext cctx;
+  cctx.compMode = CompilationMode::Infer;
+  cctx.optimizationOpts.enableConstantFolding = false;
+  cctx.backendOpts.collectConstants = true;
+  RETURN_IF_ERR(executeConstantFunction(*backend, *F, bindings, cctx));
+
+  // Using the graph output, place constant nodes in the original graph.
+  for (int i = 0; i < op.output_size(); i++) {
+    RETURN_IF_ERR(loader->createAndRegisterConstant(op.output(i),
+                                                    std::move(*outTensors[i])));
+  }
+
+  return llvm::Error::success();
+}
 
 } // namespace glow
 
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -528,6 +528,33 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
   return llvm::Error::success();
 }
 
+llvm::Expected<bool>
+Caffe2ModelLoader::foldOperator(const caffe2::OperatorDef &op) {
+  const unsigned numInputs = op.input_size();
+  const std::string &typeName = op.type();
+  llvm::SmallVector<NodeValue, 4> inputs;
+  inputs.reserve(numInputs);
+  for (unsigned i = 0; i < numInputs; i++) {
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(i)));
+    inputs.push_back(in);
+  }
+
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader to construct function representing
+  // current Op, and then constant fold the function using Interp backend.
+  Function *tmpF = G_.getParent()->createFunction("eval_const_fold__");
+  Caffe2ModelLoader tmpLoader(*tmpF, nullptr);
+  bool foldStatus =
+      !errToBool(constantFoldInLoader<Caffe2ModelLoader, caffe2::OperatorDef>(
+          tmpF, tmpLoader, this, op));
+  G_.getParent()->eraseFunction(tmpF);
+  return foldStatus;
+}
+
 llvm::Error Caffe2ModelLoader::loadOperator(const caffe2::OperatorDef &op) {
   ArgumentDictionaryTy dict = loadArgumentMap(op);
   const std::string &typeName = op.type();
@@ -1340,6 +1367,13 @@ llvm::Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
   /// Load the network operators:
   for (int i = 0; i < net.op_size(); i++) {
     auto &op = net.op(i);
+    if (getConstantFoldLoaderOpsFlag()) {
+      auto foldstatus = foldOperator(op);
+      if (foldstatus && foldstatus.get()) {
+        // Folded successfully.
+        continue;
+      }
+    }
     RETURN_IF_ERR(loadOperator(op));
   }
 
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -1152,6 +1152,34 @@ llvm::Error ONNXModelLoader::loadTile(const ONNX_NAMESPACE::NodeProto &op,
   return llvm::Error::success();
 }
 
+llvm::Expected<bool>
+ONNXModelLoader::foldOperator(const ONNX_NAMESPACE::NodeProto &op) {
+  const unsigned numInputs = op.input_size();
+  const std::string &typeName = op.op_type();
+  llvm::SmallVector<NodeValue, 4> inputs;
+  inputs.reserve(numInputs);
+  for (unsigned i = 0; i < numInputs; i++) {
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(i)));
+    inputs.push_back(in);
+  }
+
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader to construct function representing
+  // current Op, and then constant fold the function using Interp backend.
+  Function *tmpF = G_.getParent()->createFunction("eval_const_fold__");
+  ONNXModelLoader tmpLoader(*tmpF);
+  tmpLoader.opsetVersion_ = opsetVersion_;
+  bool foldStatus = !errToBool(
+      constantFoldInLoader<ONNXModelLoader, ONNX_NAMESPACE::NodeProto>(
+          tmpF, tmpLoader, this, op));
+  G_.getParent()->eraseFunction(tmpF);
+  return foldStatus;
+}
+
 llvm::Error ONNXModelLoader::loadOperator(const ONNX_NAMESPACE::NodeProto &op) {
   ArgumentDictionaryTy dict = loadArgumentMap(op);
   const std::string &typeName = op.op_type();
@@ -1256,6 +1284,13 @@ llvm::Error ONNXModelLoader::loadNetwork(ONNX_NAMESPACE::GraphProto &net) {
   /// Load the network operators:
   for (int i = 0; i < net.node_size(); i++) {
     auto &op = net.node(i);
+    if (getConstantFoldLoaderOpsFlag()) {
+      auto foldstatus = foldOperator(op);
+      if (foldstatus && foldstatus.get()) {
+        // Folded successfully.
+        continue;
+      }
+    }
     RETURN_IF_ERR(loadOperator(op));
   }
 
--- a/lib/Importer/ProtobufLoader.cpp
+++ b/lib/Importer/ProtobufLoader.cpp
@@ -15,16 +15,52 @@
  */
 
 #include "glow/Importer/ProtobufLoader.h"
-
+#include "llvm/Support/CommandLine.h"
 #include <string>
 
 namespace glow {
+
+llvm::cl::OptionCategory loaderOptCat("Model Loader Options");
+
+static llvm::cl::opt<bool> isConstFoldLoaderOps(
+    "const-fold-ops",
+    llvm::cl::desc(
+        "Performs constant folding on ONNX and Caffe Operators while loading."),
+    llvm::cl::init(false), llvm::cl::cat(loaderOptCat));
 
 bool isArrayConstant(llvm::ArrayRef<size_t> a) {
   for (size_t i = 1; i < a.size(); i++)
     if (a[0] != a[i])
       return false;
   return true;
+}
+
+void setConstantFoldLoaderOpsFlag(bool flag) { isConstFoldLoaderOps = flag; }
+
+bool getConstantFoldLoaderOpsFlag() { return isConstFoldLoaderOps; }
+
+bool ProtobufLoader::isConstantFoldable(llvm::ArrayRef<NodeValue> inputs,
+                                        std::string typeName) const {
+  int numInputs = inputs.size();
+  if (!getConstantFoldLoaderOpsFlag()) {
+    return false;
+  }
+  // foldUnsupportedTypes: List of typenames unsupported for folding.
+  std::string foldUnsupportedTypes[] = {"Constant"};
+  std::string *foo = std::find(std::begin(foldUnsupportedTypes),
+                               std::end(foldUnsupportedTypes), typeName);
+  // Early exit if folding is not supported for current operator.
+  if (foo != std::end(foldUnsupportedTypes)) {
+    return false;
+  }
+
+  // If all the inputs to the operator are constant this op can be folded.
+  bool isConstFoldCandidate = true;
+  for (int i = 0; (i < numInputs) && isConstFoldCandidate; i++) {
+    isConstFoldCandidate =
+        (inputs[i].getNode()->getKind() == Kinded::Kind::ConstantKind);
+  }
+  return isConstFoldCandidate;
 }
 
 Constant *ProtobufLoader::getConstantByNameOrNull(llvm::StringRef name) const {
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -62,6 +62,16 @@ static bool shouldDeleteNode(Node *N) {
   return true;
 }
 
+static bool shouldDeleteConstants(Function *F) {
+  Module *mod = F->getParent();
+  for (auto F : mod->getFunctions()) {
+    if (F->getState() < FunctionState::FuncLoaded) {
+      return false;
+    }
+  }
+  return true;
+}
+
 bool DCE::run(Function *F) {
   LOG_SCOPE(F->getLogContext(), getName());
 
@@ -98,6 +108,9 @@ bool DCE::run(Function *F) {
       break;
     }
   }
+
+  if (!shouldDeleteConstants(F))
+    return changed;
 
   // Delete unused Constants.
   for (auto it = consts.begin(), e = consts.end(); it != e;) {
@@ -2757,7 +2770,9 @@ void glow::fold(Function *F, CompilationMode mode) {
 
 void glow::optimize(Function *F, CompilationContext &cctx) {
   LOG_SCOPE(F->getLogContext(), "glow::optimize")
-
+  // Indicates if the given function is completely loaded. A temporary
+  // workaround until #3213 is complete.
+  F->setState(FunctionState::FuncLoaded);
   // Optimize may be called after backend specific transformations and some
   // nodes may have become unused. It is a good idea to remove them, before
   // proceeding with any further optimizations.
--- a/tests/unittests/Caffe2ImporterTest.cpp
+++ b/tests/unittests/Caffe2ImporterTest.cpp
@@ -1672,6 +1672,36 @@ TEST(caffe2, gatherRanges) {
   EXPECT_TRUE(gatherRanges->getLengths().dims().equals({2}));
 }
 
+/// Test loading Gather ops with constant folding from an Caffe2 model.
+TEST(caffe2, gatherConstantFoldingAndReshape) {
+  // This test verifies that Gather gets constant-folded, so that the argument
+  // of the reshape becomes constant.
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+
+  std::string netDescFilename(
+      GLOW_DATA_PATH "tests/models/caffe2Models/gather_const_fold.pbtxt");
+  std::string netWeightFilename(
+      GLOW_DATA_PATH "tests/models/caffe2Models/gather_const_fold_init.pbtxt");
+  PlaceholderBindings bindings;
+  auto *F = mod.createFunction("main");
+  Placeholder *output;
+  Tensor data(ElemKind::FloatTy, {1, 2, 4, 3});
+  setConstantFoldLoaderOpsFlag(true);
+  {
+    Caffe2ModelLoader caffe2LD(netDescFilename, netWeightFilename, {"data"},
+                               {&data.getType()}, *F);
+    output = EXIT_ON_ERR(caffe2LD.getOutputByName("result"));
+    bindings.allocate(mod.getPlaceholders());
+  }
+  setConstantFoldLoaderOpsFlag(false);
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+
+  auto result = bindings.get(output)->getHandle();
+  std::vector<size_t> expectedDims = {1, 4, 3, 2};
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
+}
 /// Test loading a LengthsRangeFill op.
 TEST(caffe2, LengthsRangeFill) {
   ExecutionEngine EE;
--- a/tests/unittests/OnnxImporterTest.cpp
+++ b/tests/unittests/OnnxImporterTest.cpp
@@ -19,6 +19,7 @@
 #include "glow/Graph/Nodes.h"
 #include "glow/Graph/PlaceholderBindings.h"
 #include "glow/Importer/ONNXModelLoader.h"
+#include "llvm/Support/FileSystem.h"
 #include "gtest/gtest.h"
 
 #ifndef GLOW_DATA_PATH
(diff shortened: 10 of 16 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -524,6 +525,65 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+  Module &mod_temp = ee_temp.getModule();
+  Function *pfunc_temp = mod_temp.createFunction("temp");
+  Caffe2ModelLoader loader_temp(*pfunc_temp, nullptr);
+
+  // Create a temporary graph with constant inputs, current operator and add
+  // save nodes. Recheck if even save nodes are necessary.
+  for (unsigned i = 0; i < numInputs; i++) {
+    Node *pn = inputs[i].getNode();
+    const Tensor &data = (llvm::dyn_cast<Constant>(pn))->getPayload();
+    Constant *pConst = mod_temp.createConstant(op.input(i), data);
```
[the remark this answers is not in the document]
ksaurabh-cadence: This copy of constant payload is temporary and should get freed upon exiting const fold. I chose to not move the original payload with an intention of not disturbing the original graph. Move would have required me to move back the constant to the original graph, and with RETURN_IF_ERR in the code, I didn't want to deal with that at that time. 

I toyed with the idea of writing a compile_light with optimizations turned off but apparently some operator depended on the optimization so I reverted back to the original compile. 

I am not sure why it is a problem if the constant is quantized, which anyway would have quantized in the later pass. I think during the later later pass it probably won't touch the already quantized constant. Pardon me, I haven't tried quantized networks hence my limited understanding on this example. Still I think since this is an optional feature, a user can always choose to not use it if some side-effects are seen. Also if there is a case encountered we can update the constantFoldable logic to indicate something as non-foldable.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -524,6 +525,65 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(i)));
+    inputs.push_back(in);
+  }
+
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader and execution engine.
```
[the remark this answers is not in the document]
ksaurabh-cadence: @opti-mix Are you suggesting to remove it out of Loader class? There are some loader specific things like setting up of opset version (specific only to onnx), **op** is also of different type for each loader hence the choice to to keep it separate.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -524,6 +525,65 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+  PlaceholderBindings binding_temp;
+  ExecutionEngine ee_temp;
+  Module &mod_temp = ee_temp.getModule();
```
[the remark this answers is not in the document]
ksaurabh-cadence: I will update that.

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: @jfix71  I actually tried to reuse the parent module by create a new temporary function let's call it "fold". My expectation was that this way I don't have to copy the constants and it could directly have access to the constants from the parent module. I view constants as a shared resource between functions created out of the same module. However, when the temporary function "fold" was optimized as part of compile it was started deleting the constants of the modules even though they were not part of the "fold" function or graph. Shouldn't it delete only the dead constants which are part of the current function, instead of deleting the modules of the parent module? Either this is a bug or my understanding is wrong.

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
opti-mix: > Are you suggesting to remove it out of Loader class?

Not removing it completely, but moving the common-logic into a reusable helper function if possible. Obviously, some part of the logic is pretty loader-specific and should stay in its respective loader.

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
jfix71: >However, when the temporary function "fold" was optimized as part of compile it started deleting the constants of the parent module even though they were not part of the "fold" function or graph. 

Your understanding is correct (or at least matches mine 🙂) -- this sounds wrong/broken. This is similar to what #3161 does -- create a new Function in the Module which references the same Constants and run it. How can you tell it's deleting the Constants, what kind of error is it throwing?

Also, I think #3161 is a more generalize implementation than this PR. I believe if we're able to get that one to work we can call into it from the Loaders like this one does. We would still need to add calls for it of course.

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: @opti-mix  I am addressing this by creating a template Friend function.

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: @jfix71 From constant folding at loader perspective I find this implementation to be much simpler and general enough.:-) I am hoping if loader can take care of this upfront, a lot less will percolate to the node level. I am not sure if it can be compared with #3161 as that one does not use the loader and probably will be called once the loading is finished. 

I am speculating that the deletion happens because I call the folding function while I am still loading the graph. All the uses of constants are still not currently marked in the original graph as it's only partially loaded. So it's possible when DCE is called for the temporary graph from the same module, it starts deleting the constants. I will be writing a compile_light, which I guess will solve the problem, unless somebody is already doing this.

ksaurabh-cadence:
@jfix71 @rdzhabarov 

Pushed a new version which 
a) avoids copying of constants
b) Introduce state for each Function. This helps avoid deleting constants when the parent graph/Function is partially loaded. 
c) Added a checking function shouldDeleteConstants which helps avoid pre-mature deletion of constants in DCE

Changes in previous update
a) common template folding function constantFoldInLoader introduce for folding via Loader

lib/Importer/Caffe2ModelLoader.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: @jfix71 Have added a FunctionState to the objects of type Function to indicate if the Function has been created/loaded/compiled/etc. In DCE introduced a check for deleteConstant to ensure that constants are deleted only if all the Functions in the parent modules  are already loaded. Now I don't have to copy the constants. I think in general it will be good to have state of Functions set to something meaningful.

opti-mix:
@ksaurabh-cadence Could you have a look at #3161? In particular, I think `glow::constantFold` in ConstantFolding.cpp could be useful for your purposes. And it completely avoids using any ExecutionEngines and other heavy-weight runtime components. If we need to adjust the APIs, please let me know. I intend to land this PR and it would be nice if it would also be useful for loaders.

ksaurabh-cadence:
@opti-mix I had a look at it. Calling constantFold for each Node doesn't make sense at loader level because  one op doesn't always map to one Node. What is being done in this MR is creation of a Function for an Op in "general" sense. Once the function is created I am using a execution engine to evaluate the constant, once a lighter version of compile/run is available even that can be used.

opti-mix:
@ksaurabh-cadence 

>  Calling constantFold for each Node doesn't make sense at loader level because one op doesn't always map to one Node. 

It can be a "root" node of the whole constant computation, i.e. it does not to be literally a single node. It just needs to be something recursively constant evaluatable at compile-time. 

If there is more than a single recursively constant evautable node (i.e. it is not a single tree, but e.g. a forest) my PR contains a ConstantFold pass that works on whole functions.

Or do you say it is still not enough for your purposes?

lib/Importer/ProtobufLoader.cpp
```diff
@@ -15,18 +15,53 @@
...
     if (a[0] != a[i])
       return false;
   return true;
 }
 
+void setConstantFoldLoaderOpsFlag(bool flag) { isConstFoldLoaderOps = flag; }
+
+bool getConstantFoldLoaderOpsFlag() { return isConstFoldLoaderOps; }
+
+bool ProtobufLoader::isConstantFoldable(llvm::SmallVector<NodeValue, 4> &inputs,
```
opti-mix: BTW, why does it check only for immediately constant foldable, but would not allow e.g. a recursively constant-foldable patterns like ((c1 + c2) - c3)*c4)

ksaurabh-cadence:
@opti-mix 

It's not clear to me how do I find the root node in the loader in a simple and generic way for an operator?  Secondly here the main effort is on creating the temporary function without copying anything. Once the function is created I am currently calling compile and Run. If there is a something lightweight I will be happy to call it, but so far my tests have been failing for anything lighter.

lib/Importer/ProtobufLoader.cpp
ksaurabh-cadence: The loading happens sequentially. You will always load the input nodes before the current node, so this greedy approach takes care of recursively foldable scenarios.

opti-mix:
>  If there is a something lightweight I will be happy to call it, but so far my tests have been failing for anything lighter.

Would e.g. `execute` from ConstantFolding.cpp work for you?

ksaurabh-cadence:
> > If there is a something lightweight I will be happy to call it, but so far my tests have been failing for anything lighter.
> 
> Would e.g. `execute` from ConstantFolding.cpp work for you?

How different is it from following lines in code? - 
 eeCF.compile(CompilationMode::Infer, pFuncCF);
 eeCF.run(bindingCF);

If they are same, it should.

opti-mix:
> How different is it from following lines in code? -
eeCF.compile(CompilationMode::Infer, pFuncCF);
eeCF.run(bindingCF);

Not much different in terms of compilation-related functionalities. But ExecutionEngine uses HostManager/DeviceManager and threads to perform the run. This is pretty heavy-weight if it is only used to perform compile-time constant folding. 

`execute` in my PR does not use any of these have weight components. But other than that it is the same.

ksaurabh-cadence:
> How different is it from following lines in code? -
> eeCF.compile(CompilationMode::Infer, pFuncCF);
> eeCF.run(bindingCF);
> 
> Not much different in terms of compilation-related functionalities. But ExecutionEngine uses HostManager/DeviceManager and threads to perform the run. This is pretty heavy-weight if it is only used to perform compile-time constant folding.
> execute in my PR does not use any of these have weight components. But other than that it is the same.

Then it should be good. Once your PR is in, I can try this to see if my tests are happy. :-)

opti-mix:
@ksaurabh-cadence 

> Then it should be good. Once your PR is in, I can try this to see if my tests are happy. :-)

My PR has landed ;-) Please have a look at `glow::executeConstantFunction` in ConstantFolding.cpp

ksaurabh-cadence:
@opti-mix OK I will try this and push an update if it works fine. I hope this PR can be closed soon now, I would like to merge few other MRs but would want this to close first.  :-)

ksaurabh-cadence:
@opti-mix @rdzhabarov The code is updated to use glow::executeConstantFunction. My local test for folding are passing.  However I am not sure why the ASAN test is failing here in the pipeline. Help?

ksaurabh-cadence:
@opti-mix @rdzhabarov Please review and help close this now. Thanks!

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,52 @@ class ProtobufLoader {
...
 
+/// \returns success if the folding of operator \p op in the loader
+/// \p pLoader is successful. The folding utility uses temporary
+/// loader \p loaderCF, and associated temporary function \p pFuncCF.
+template <class LoaderType, class OpType>
+llvm::Error constantFoldInLoader(Function *pFuncCF, LoaderType &loaderCF,
+                                 LoaderType *pLoader, const OpType &op) {
+  PlaceholderBindings bindingCF;
+  std::vector<Tensor *> pTensorVec;
+  Module *pModCF = pFuncCF->getParent();
```
opti-mix: Please add some comments explaining what the following two loops are doing.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
 
 using namespace glow;
 
+#include <fstream>
+using namespace std;
+
+llvm::Expected<ONNX_NAMESPACE::ModelProto>
```
opti-mix: Please add a Doxygen comment.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+    ONNX_NAMESPACE::ModelProto MP;
+    bool parseNet = google::protobuf::TextFormat::ParseFromString(str, &MP);
+    RETURN_ERR_IF_NOT(parseNet, "Failed to parse ModelProto",
+                      GlowErr::ErrorCode::MODEL_LOADER_INVALID_PROTOBUF);
+    return MP;
+  }
+  RETURN_ERR("Can't load proto file");
+}
+
+llvm::Expected<bool> saveProto(const std::string &filename,
```
opti-mix: Please add a Doxygen comment.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+    std::string onnx_message = model.DebugString();
+    ff << onnx_message;
+    ff.close();
+    return true;
+  }
+  ff.close();
+  return false;
+}
+
+llvm::Expected<bool>
```
opti-mix: Please add a Doxygen comment.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+        std::cout << "Unsupported datatype";
+        return false;
+      }
+    }
+  }
+  gp->clear_input();
+  return true;
+}
+
+bool checkConstFoldedOutput(std::string NetFilename,
```
opti-mix: Please add a Doxygen comment.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+                            llvm::ArrayRef<const char *> tensorNames,
+                            llvm::ArrayRef<Tensor *> tensors,
+                            llvm::ArrayRef<Tensor *> ref_tensors) {
```
opti-mix: Glow usually uses camelCase rather than underscores.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+  if (!lhsOrErrV) {
+    return false;
+  }
+  modelDef = std::move(lhsOrErrV.get());
+  if (!replacePlaceholderWithConstant(modelDef, tensorNames, tensors))
+    return false;
+  if (!saveProto(CFNetFilename, modelDef))
+    return false;
+  setConstantFoldLoaderOpsFlag(true);
+  PlaceholderBindings binding_temp;
```
opti-mix: Same here.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+  setConstantFoldLoaderOpsFlag(false);
+
+  const llvm::StringMap<Placeholder *> pOut = onnxLD.getOutputVarsMapping();
+  for (int i = 0; i < modelDef.graph().output_size(); i++) {
+    NodeValue nv;
+    auto lhsOrErrV =
+        onnxLD.getNodeValueByName(modelDef.graph().output(i).name());
+    static_assert(IsLLVMExpected<decltype(lhsOrErrV)>(),
+                  "Expected value to be a llvm::Expected");
+    if (lhsOrErrV) {
```
opti-mix: Try to use an early exit to reduce the indentation.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,174 @@
...
+      nv = std::move(lhsOrErrV.get());
+      switch (nv.getElementType()) {
+      case ElemKind::FloatTy:
+        EXPECT_TRUE(ref_tensors[i]->getHandle().dims() ==
```
opti-mix: Tiny nit: Each of the case blocks does the same, just the type differs. Looks like a lot of boilerplate code. May be a you could use a template helper and call it with different types?

include/glow/Graph/Graph.h
```diff
@@ -46,6 +46,14 @@ using PlaceholderList = std::list<Placeholder *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 /// Map from original Nodes to cloned Nodes.
 using NodeMap = llvm::DenseMap<Node *, Node *>;
+/// State of a function.
```
opti-mix: Please provide a bit more elaborative description for `FunctionState` and how/where it is supposed to be used.

opti-mix:
@jfix71 Do you have any comments in particular about the algorithmic part?

include/glow/Graph/Graph.h
```diff
@@ -46,6 +46,14 @@ using PlaceholderList = std::list<Placeholder *>;
...
+enum class FunctionState {
+  FuncCreated,
+  FuncLoaded,
+  FuncCompiled,
```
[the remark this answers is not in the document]
ksaurabh-cadence: done

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -107,6 +110,11 @@ class Caffe2ModelLoader
 
   friend class ONNXIFIModelLoader;
 
+  template <class LoaderType, class OpType>
```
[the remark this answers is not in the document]
ksaurabh-cadence: The comment is already present above function definition. Should it be copy pasted?

include/glow/Importer/ProtobufLoader.h
```diff
@@ -38,6 +40,12 @@
 
 namespace glow {
 
+// Enables or disables constant-folding of Loader Ops with \p flag.
```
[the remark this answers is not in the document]
ksaurabh-cadence: done

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,52 @@ class ProtobufLoader {
...
+  /// list as \p inputs is constant foldable.
+  bool isConstantFoldable(llvm::SmallVector<NodeValue, 4> &inputs,
+                          std::string typeName) const;
 };
 
+/// \returns success if the folding of operator \p op in the loader
+/// \p pLoader is successful. The folding utility uses temporary
+/// loader \p loaderCF, and associated temporary function \p pFuncCF.
+template <class LoaderType, class OpType>
+llvm::Error constantFoldInLoader(Function *pFuncCF, LoaderType &loaderCF,
```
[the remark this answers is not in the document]
ksaurabh-cadence: Made it just "F" like everywhere else.

include/glow/Importer/Caffe2ModelLoader.h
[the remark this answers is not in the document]
opti-mix: We usually have Doxygen comments for all function declarations/definitions inside header files. Definitions inside cpp files do sometimes omit them (but it is fine to have them here as well) if their declarations in the header files has it already.

include/glow/Importer/ProtobufLoader.h
ksaurabh-cadence: done.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,52 @@ class ProtobufLoader {
...
+                                 LoaderType *pLoader, const OpType &op) {
+  PlaceholderBindings bindingCF;
```
jfix71: I think everything in this function looks good, but it could use some comments in the body to discuss what's going on, so when I look at this code a month from now it's easier to consume. E.g. `// Find all Constant and place them in the new Loader ...`, `// Load operator in the new Loader and save each output from op`, etc.

lib/Importer/ProtobufLoader.cpp
jfix71: So in this case we would fold `c1 + c2 = c'`, then fold `c' - c3 = c''`, then fold `c'' * c4`. Which means there are multiple compile calls, building multiple temporary Functions/Loaders, etc. It'd be nice to somehow delay the folding to create a single temporary Function/Loader to use like we do with `recursiveClone()` in the graph optimizer, but at first thought this might be a bit complex to handle.

lib/Importer/ProtobufLoader.cpp
```diff
@@ -15,18 +15,53 @@
...
+                                        std::string typeName) const {
+  int numInputs = inputs.size();
+  if (!getConstantFoldLoaderOpsFlag()) {
+    return false;
+  }
+  // fold_unsupported_types: List of typenames unsupported for folding.
+  std::string fold_unsupported_types[] = {"Constant"};
```
jfix71: nit: `foldUnsupportedTypes`

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1337,6 +1364,13 @@ llvm::Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
   /// Load the network operators:
   for (int i = 0; i < net.op_size(); i++) {
     auto &op = net.op(i);
+    if (getConstantFoldLoaderOpsFlag()) {
+      auto foldstatus = foldOperator(op);
```
jfix71: nit: `foldStatus`

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1246,6 +1274,13 @@ llvm::Error ONNXModelLoader::loadNetwork(ONNX_NAMESPACE::GraphProto &net) {
   /// Load the network operators:
   for (int i = 0; i < net.node_size(); i++) {
     auto &op = net.node(i);
+    if (getConstantFoldLoaderOpsFlag()) {
+      auto foldstatus = foldOperator(op);
```
jfix71: nit: `foldStatus`

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1246,6 +1274,13 @@ llvm::Error ONNXModelLoader::loadNetwork(ONNX_NAMESPACE::GraphProto &net) {
...
+      if (foldstatus && foldstatus.get()) {
+        // folded successfully.
```
jfix71: nit: Capitalize sentence.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1337,6 +1364,13 @@ llvm::Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
...
+      if (foldstatus && foldstatus.get()) {
+        // folded successfully.
```
jfix71: nit: Capitalize sentence.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -527,6 +527,33 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader to construct function representing
+  // current Op, and then constant fold the function using Interp backend.
+  Function *pfuncCF = G_.getParent()->createFunction("eval_const_fold__");
+  Caffe2ModelLoader loaderCF(*pfuncCF, nullptr);
+  bool fold_status =
+      !errToBool(constantFoldInLoader<Caffe2ModelLoader, caffe2::OperatorDef>(
```
jfix71: It's a little funny that we convert this err to a bool even though we're inside a function that returns an `llvm::Expected`. I see we have `ASSIGN_VALUE_OR_RETURN_ERR()` above. I think it would make sense to either return the error here, or otherwise above return false if `getNodeValueByName()` returns an error above and change this function to just return a `bool`.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -62,6 +62,15 @@ static bool shouldDeleteNode(Node *N) {
   return true;
 }
 
+static bool shouldDeleteConstants(Function *F) {
+  Module *pMod = F->getParent();
+  bool state = true;
+  for (auto F : pMod->getFunctions()) {
+    state = state & (F->getState() >= FunctionState::FuncLoaded);
```
jfix71: nit: If we keep this, we could just do `if (F->getState() < FunctionState::FuncLoaded) return false` and then return true at the end of the func.

lib/Importer/ProtobufLoader.cpp
ksaurabh-cadence: I wanted it to be simple, generic enough and less error prone.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2754,7 +2766,7 @@ void glow::fold(Function *F, CompilationMode mode) {
 
 void glow::optimize(Function *F, CompilationContext &cctx) {
   LOG_SCOPE(F->getLogContext(), "glow::optimize")
-
+  F->setState(FunctionState::FuncLoaded);
```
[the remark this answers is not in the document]
ksaurabh-cadence: Folding or no folding, isn't the deletion of constants a bug, please correct me if I am wrong?  Let's say when there are two functions coming from a same module, sharing the constants, deletion of constants by DCE should be automatically protected unless all the functions are completely loaded even if we are not doing the folding. My intent was to somehow indicate that a function is completely loaded in backward compatible way.  I thought compile is a safe enough place to assume that the function is completely loaded, and probably there is no harm even if it gets set multiple times. 

I have no objection to adding enableUnusedConstantDeletion, but this puts onus of setting this flag on the caller to ensure that all sister Functions are completely loaded. With functionState this was taken care of by the framework. Probably a binary flag isFuncLoaded would have been less confusing. Let me know if your prefer the optimizationOptions way, I will rollback the FunctionState and use your suggestion.

lib/Importer/Caffe2ModelLoader.cpp
ksaurabh-cadence: Problem with ASSIGN_VALUE_OR_RETURN_ERR is that it doesn't give the opportunity to do cleanup. We have to erase the temporary function which we created just before constantFoldInLoader. Hence the use of errToBool. ASSIGN_VALUE_OR_RETURN_ERR used above doesn't require cleanup hence is fine.

tests/unittests/OnnxImporterTest.cpp
ksaurabh-cadence: I just discovered isEqual member function of Tensor, so this whole code will go.

include/glow/Graph/Graph.h
```diff
@@ -46,6 +46,14 @@ using PlaceholderList = std::list<Placeholder *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 /// Map from original Nodes to cloned Nodes.
 using NodeMap = llvm::DenseMap<Node *, Node *>;
+/// State of a function.
+enum class FunctionState {
+  FuncCreated,
```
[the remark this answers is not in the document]
ksaurabh-cadence: done.

tests/unittests/OnnxImporterTest.cpp
ksaurabh-cadence: done

> Looks pretty good -- some nits, and one main question about the necessity of `FunctionState`.
> 
> Also, can you also add a unit test for testing an op like TopK with multiple outputs?

@jfix71 I have a separate patch for TopK, which I am planning to submit as a separate MR to add support for k as an input instead of just attribute. With that folding test for TopK will be available.

lib/Importer/Caffe2ModelLoader.cpp
ksaurabh-cadence: done

ksaurabh-cadence:
@opti-mix @jfix71  Code has been updated to address most of your comments.
Please let me know if you are still not convinced about some things, I have an open mind :-)

include/glow/Importer/Caffe2ModelLoader.h
```diff
@@ -107,6 +110,14 @@ class Caffe2ModelLoader
 
   friend class ONNXIFIModelLoader;
 
+  /// \returns success if the folding of operator \p op in the loader
+  /// \p pLoader is successful. The folding utility uses temporary
+  /// loader \p loaderCF, and associated temporary function \p F.
+  template <class LoaderType, class OpType>
+  friend llvm::Error constantFoldInLoader(Function *F, LoaderType &loaderCF,
```
opti-mix: Can you also use just `tmpLoader` and `loader` instead of `loaderCF` and `pLoader`?

include/glow/Graph/Graph.h
```diff
@@ -46,6 +46,13 @@ using PlaceholderList = std::list<Placeholder *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 /// Map from original Nodes to cloned Nodes.
 using NodeMap = llvm::DenseMap<Node *, Node *>;
+/// State of a function. This can be used to control optimizations which depend
+/// on the state of the Function.
+enum class FunctionState {
+  FuncCreated, /// Indicates that the function has been created but not
```
opti-mix: We usually put comments before declarations, not in the same line.

include/glow/Importer/ONNXModelLoader.h
```diff
@@ -195,6 +198,14 @@ class ONNXModelLoader
 
   friend class ONNXIFIModelLoader;
 
+  /// \returns success if the folding of operator \p op in the loader
+  /// \p pLoader is successful. The folding utility uses temporary
+  /// loader \p loaderCF, and associated temporary function \p F.
+  template <class LoaderType, class OpType>
+  friend llvm::Error constantFoldInLoader(Function *F, LoaderType &loaderCF,
```
opti-mix: Same comments about names here.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
...
+  /// list as \p inputs is constant foldable.
+  bool isConstantFoldable(llvm::SmallVector<NodeValue, 4> &inputs,
+                          std::string typeName) const;
 };
 
+/// \returns success if the folding of operator \p op in the loader
+/// \p pLoader is successful. The folding utility uses temporary
+/// loader \p loaderCF, and associated temporary function \p F.
+template <class LoaderType, class OpType>
+llvm::Error constantFoldInLoader(Function *F, LoaderType &loaderCF,
```
opti-mix: Same comments about names here.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
...
+                                 LoaderType *pLoader, const OpType &op) {
+  PlaceholderBindings bindingCF;
+  std::vector<Tensor *> pTensorVec;
```
opti-mix: `pTensorVec` does not sounds like a Glow name ;-) Could you use something like `xyzTensors`, e.g. `resultTensors` or may be just `results`?

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
...
+
+  // Register the constant inputs to the current op with the constant folding
+  // loader.
+  for (unsigned i = 0; i < op.input_size(); i++) {
+    Constant *pConst = pModCF->getConstantByName(op.input(i));
+    RETURN_ERR_IF_NOT(pConst, "No constant found");
+    loaderCF.nodeValueByName_[op.input(i)] = pConst->getOutput();
+  }
+
+  // Using the loader load the current operator.
```
opti-mix: loader load -> loader to load

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
...
 
+/// \returns success if the folding of operator \p op in the loader
+/// \p pLoader is successful. The folding utility uses temporary
+/// loader \p loaderCF, and associated temporary function \p F.
+template <class LoaderType, class OpType>
+llvm::Error constantFoldInLoader(Function *F, LoaderType &loaderCF,
+                                 LoaderType *pLoader, const OpType &op) {
+  PlaceholderBindings bindingCF;
+  std::vector<Tensor *> pTensorVec;
+  Module *pModCF = F->getParent();
```
opti-mix: `pModCF` also sounds not like a typical Glow name.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -527,6 +527,33 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+    inputs.push_back(in);
+  }
+
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader to construct function representing
+  // current Op, and then constant fold the function using Interp backend.
+  Function *pfuncCF = G_.getParent()->createFunction("eval_const_fold__");
```
opti-mix: pFuncCF -> tmpF?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -527,6 +527,33 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
...
+  Caffe2ModelLoader loaderCF(*pfuncCF, nullptr);
```
opti-mix: loaderCF -> tmpLoader?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1142,6 +1142,34 @@ llvm::Error ONNXModelLoader::loadTile(const ONNX_NAMESPACE::NodeProto &op,
...
+    inputs.push_back(in);
+  }
+
+  if (!isConstantFoldable(inputs, typeName)) {
+    return false;
+  }
+
+  // Create a temporary lightweight loader to construct function representing
+  // current Op, and then constant fold the function using Interp backend.
+  Function *pfuncCF = G_.getParent()->createFunction("eval_const_fold__");
```
opti-mix: Similar comments about the naming.

lib/Importer/ProtobufLoader.cpp
```diff
@@ -15,18 +15,53 @@
...
+  int numInputs = inputs.size();
+  if (!getConstantFoldLoaderOpsFlag()) {
+    return false;
+  }
+  // foldUnsupportedTypes: List of typenames unsupported for folding.
+  std::string foldUnsupportedTypes[] = {"Constant"};
+  std::string *foo = std::find(std::begin(foldUnsupportedTypes),
+                               std::end(foldUnsupportedTypes), typeName);
+  // Early exit if folding is not supported for current operator.
+  if (foo == std::end(foldUnsupportedTypes)) {
```
opti-mix: May be reverse the condition and perform an early return. This avoid the big indented block which is currently the body of the `if`-statement.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -1672,6 +1672,36 @@ TEST(caffe2, gatherRanges) {
...
 }
 
+/// Test loading Gather ops with constant folding from an Caffe2 model.
+TEST(caffe2, gatherConstantFoldingAndReshape) {
+  // This test verifies that Gather gets constant-folded, so that the argument
+  // of the reshape becomes constant.
+  ExecutionEngine EE;
+  auto &mod = EE.getModule();
+
+  std::string NetDescFilename(
```
opti-mix: Glow uses camelCase, not CamelCase ;-)

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+replacePlaceholderWithConstant(ONNX_NAMESPACE::ModelProto &model,
+                               llvm::ArrayRef<const char *> tensorNames,
+                               llvm::ArrayRef<Tensor *> tensors) {
+  ONNX_NAMESPACE::NodeProto np;
+  ONNX_NAMESPACE::GraphProto *gp = model.mutable_graph();
+  RETURN_ERR_IF_NOT(gp, "Can't get mutable graph.",
+                    GlowErr::ErrorCode::RUNTIME_ERROR);
+  for (size_t i = 0; i < tensorNames.size(); i++) {
+    for (int j = 0; j < gp->input_size(); j++) {
+      ONNX_NAMESPACE::ValueInfoProto *pvalueInfo = gp->mutable_input(j);
```
opti-mix: pvalueInfo -> valueInfo or valueInfoProto?

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+      const std::string &inputName = pvalueInfo->name();
+      if (inputName != tensorNames[i]) {
+        continue;
+      }
+      std::string new_name = "dummy_input" + std::to_string(i);
```
opti-mix: new_name -> newName

include/glow/Graph/Graph.h
ksaurabh-cadence: I looked at enum class QuantizationMode and followed the same convention. Is this documented somewhere? Looking at the code its hard to figure out. It's different at multiple places. :-)

ksaurabh-cadence:
@opti-mix @jfix71 Code updated to address the review comments. Please take a look. Thanks.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+  if (!lhsOrErrV) {
+    return false;
+  }
+  modelDef = std::move(lhsOrErrV.get());
+  if (!replacePlaceholderWithConstant(modelDef, tensorNames, tensors))
+    return false;
+  if (!saveProto(CFNetFilename, modelDef))
+    return false;
+  setConstantFoldLoaderOpsFlag(true);
+  ExecutionEngine ee;
```
jfix71: nit: We usually use `EE`.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+  Module &mod = ee.getModule();
+  Function *F = mod.createFunction("temp");
+  ONNXModelLoader onnxLD(CFNetFilename, {}, {}, *F);
+  setConstantFoldLoaderOpsFlag(false);
+
+  const llvm::StringMap<Placeholder *> pOut = onnxLD.getOutputVarsMapping();
+  for (int i = 0; i < modelDef.graph().output_size(); i++) {
+    NodeValue nv;
```
jfix71: nit: We usually use `NV`.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+    auto lhsOrErrV =
+        onnxLD.getNodeValueByName(modelDef.graph().output(i).name());
+    static_assert(IsLLVMExpected<decltype(lhsOrErrV)>(),
+                  "Expected value to be a llvm::Expected");
+    if (!lhsOrErrV) {
+      return false;
+    }
+    nv = std::move(lhsOrErrV.get());
+    auto *constOut = llvm::dyn_cast<Constant>(nv.getNode());
+    EXPECT_TRUE(expectedTensors[i]->isEqual(constOut->getPayload()));
```
jfix71: You should `ASSERT_TRUE(constOut)` before this line to make sure it's actually a Constant.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+      }
+    }
+  }
+  gp->clear_input();
+  return true;
+}
+
+/// Performs constant folding test on the given model file \p NetFilename
+/// by replacing input tensors with name \p tensorNames, and values \p tensors
+/// and then checking against expected output expectedTensors. \returns true
```
jfix71: nit: `\p expectedTensors`

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -112,6 +249,9 @@ importArithMultiBroadcastTest(std::string fileName,
   for (size_t i = 0; i < result.getType().size(); i++) {
     EXPECT_FLOAT_EQ(result.raw(i), expectedValues[i]);
   }
+  // Constant Folding Test
```
jfix71: nit: Period at end of sentence here and everywhere comment was copied.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+  if (!saveProto(CFNetFilename, modelDef))
+    return false;
+  setConstantFoldLoaderOpsFlag(true);
+  ExecutionEngine ee;
+  Module &mod = ee.getModule();
+  Function *F = mod.createFunction("temp");
+  ONNXModelLoader onnxLD(CFNetFilename, {}, {}, *F);
+  setConstantFoldLoaderOpsFlag(false);
+
+  const llvm::StringMap<Placeholder *> pOut = onnxLD.getOutputVarsMapping();
```
jfix71: nit: Can you add some more comments in the body here? E.g. here we're getting Constants from the module that were created during Constant Folding inside Loader, which are named as the original outputs of the Function. Then we're comparing it to previously results calculated + passed in.

opti-mix:
@ksaurabh-cadence BTW, you marked many of the naming related comments as resolved, even though no renaming was performed. There are still a bunch of `pXYZ` or `xyzCF` names in your PR. Not very critical for landing it, but would be nice to follow the general naming conventions used by the codebase.

ksaurabh-cadence:
@opti-mix You are right! There was a glitch and the push didn't happen properly. I've pushed it again. Sorry about that.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
   /// \returns the Placeholder for the external output with \p name.
   /// \pre outputVarsByName_.find(name) != outputVarsByName_.end()
   llvm::Expected<Placeholder *> getOutputByName(llvm::StringRef name) const;
+
+  /// \returns True if the operator with name \p typeName having input node
+  /// list as \p inputs is constant foldable.
+  bool isConstantFoldable(llvm::SmallVector<NodeValue, 4> &inputs,
```
opti-mix: It is better to use `llvm::SmallVectorImpl<NodeValue> &` or even better `llvm::ArrayRef<NodeValue>` in such cases to support any size of the passed vector.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -181,8 +189,62 @@ class ProtobufLoader {
...
+llvm::Error constantFoldInLoader(Function *F, LoaderType &tmpLoader,
+                                 LoaderType *loader, const OpType &op) {
+  PlaceholderBindings bindings;
+  std::vector<Tensor *> outTensors;
+  Module *mod = F->getParent();
+
+  // Register the constant inputs to the current op with the constant folding
+  // loader.
+  for (unsigned i = 0; i < op.input_size(); i++) {
+    Constant *pConst = mod->getConstantByName(op.input(i));
```
opti-mix: pConst ->tmpConst

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: I am currently creating a temporary function from the same module, so that I can reuse the constants. With the current implementation delete constants will happen only when for last function in the module is loaded, not before that. Is your suggested change going to break this implementation, because then we won't be able to compile the temporary function ?  Did I miss something in your suggestion?

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -27,6 +27,144 @@
...
+/// by replacing input tensors with name \p tensorNames, and values \p tensors
+/// and then checking against expected output expectedTensors. \returns true
+/// if the test completes without error.
+bool checkConstFoldedOutput(std::string NetFilename,
+                            llvm::ArrayRef<const char *> tensorNames,
+                            llvm::ArrayRef<Tensor *> tensors,
+                            llvm::ArrayRef<Tensor *> expectedTensors) {
+  ONNX_NAMESPACE::ModelProto modelDef;
+  std::string CFNetFilename = "dummy.onnxtxt";
+  auto lhsOrErrV = loadProto(NetFilename);
```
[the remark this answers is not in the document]
ksaurabh-cadence: done. I was trying to avoid touching Error.h. :-)

ksaurabh-cadence:
@jfix71 @opti-mix Hopefully last revision. :-)

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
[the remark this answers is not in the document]
jfix71: I think we will end up with a temporary Module instead of a temporary Function. So yeah this implementation would break. But we'd fix it when we make that change.

include/glow/Graph/Graph.h
```diff
@@ -46,6 +46,14 @@ using PlaceholderList = std::list<Placeholder *>;
 using UnsignedArrayRef = llvm::ArrayRef<size_t>;
 /// Map from original Nodes to cloned Nodes.
 using NodeMap = llvm::DenseMap<Node *, Node *>;
+/// State of a function. This can be used to control optimizations which depend
+/// on the state of the Function.
```
jfix71: Can you add a disclaimer comment making it very clear this is a temporary workaround until GH Issue #3213 is complete? I don't want people to start relying on this if we will remove it in the future.

ksaurabh-cadence:
@jfix71 @opti-mix take care of the last piece as well. Please merge. Thanks.

opti-mix:
@ksaurabh-cadence 

It has finally landed! 🎆 🎊 

Great job! And thanks again for working on this useful feature!

PR #3142, opened 2019-06-19
[Partitioner] partition memory validation

beicy:
Summary:
Added partition memory usage validation function.  In current partitioner, we requires that the memory usage of each single node should be less than the device memory. After that , in each step , the algorithm will guarantee that the memory is satisfied. 
This validation can check the case that one single node is too large to fit into one device( e.g. the added unnittest).  The error message is like:
```
[ RUN      ] PartitionerTest.memoryUsageValidation1
E0619 14:53:38.690162 369878464 Error.h:244] Converting error to boolean: location: /Users/wangm/NewGLOW/glow/lib/Partitioner/Partitioner.cpp:141 message: Partition failed: the memory usage(848) of one partitoion exceeds the available memory(500) of given devices(Interpreter).
[       OK ] PartitionerTest.memoryUsageValidation1 (1 ms)
```
However, this validation will be more useful in user defined partitions (the feature will be added later) to check if their partition is valid or not.   

Documentation:

[Optional Fixes #issue]

Test Plan:
added unnittest, ninja test.

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -228,10 +228,14 @@ class Partitioner {
   /// devices.
   DeviceIDTy assignLogicalDeviceID(NodeToFunctionMap &partitions);
 
-  /// Check if this partition way \p partitions satisfies number of physical
-  /// devices restriction. I.e. check if the number of logical devices is less
-  /// than the given physical devices.
+  /// Check if \p partitions satisfies number of physical devices restriction.
+  /// I.e. check if the number of logical devices is less than the given
+  /// physical devices.
   llvm::Error logicalDevicesValidation(NodeToFunctionMap &partitions);
+
+  /// Check if the memory usage of each partition meets the physical device
+  /// memory restriction.
+  llvm::Error memoryUsageValidation(NodeToFunctionMap &partitions);
 
   /// Duplicates all networks in the module order to saturate the Host.
   void saturateHost(unsigned logicalDeviceCount);
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -128,6 +128,21 @@ Partitioner::logicalDevicesValidation(NodeToFunctionMap &partitions) {
   return llvm::Error::success();
 }
 
+llvm::Error Partitioner::memoryUsageValidation(NodeToFunctionMap &partitions) {
+  for (auto &func : partitions.getPartitions()) {
+    auto backendName = partitions.getPartitionBackendName(func);
+    auto usedMemSize = partitions.getGraphMemInfo(func).getTotalMemSize();
+    RETURN_ERR_IF_NOT(
+        usedMemSize <= backendMap_[backendName].memSize,
+        llvm::formatv(
+            "Partition failed: the memory usage({0}) of one partition exceeds "
+            "the available memory({1}) of given devices({2}).",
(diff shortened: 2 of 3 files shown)


include/glow/Partitioner/Partitioner.h
```diff
@@ -233,6 +233,10 @@ class Partitioner {
   /// than the given physical devices.
   llvm::Error logicalDevicesValidation(NodeToFunctionMap &partitions);
 
+  /// Check if the memory usage of each partition meets the physical device
+  /// memory restriction.
+  llvm::Error memoryUsageValidation(NodeToFunctionMap &partitions);
```
rdzhabarov: semantically this should not change partitions and any state of the class.

anything stops from this being:
llvm::Error memoryUsageValidation(```const``` NodeToFunctionMap &partitions) ```const```;

lib/Partitioner/Partitioner.cpp
```diff
@@ -128,6 +128,21 @@ Partitioner::logicalDevicesValidation(NodeToFunctionMap &partitions) {
...
 }
 
+llvm::Error Partitioner::memoryUsageValidation(NodeToFunctionMap &partitions) {
+  for (auto &func : partitions.getPartitions()) {
+    auto backendName = partitions.getPartitionBackendName(func);
+    auto usedMemSize = partitions.getGraphMemInfo(func).getTotalMemSize();
+    RETURN_ERR_IF_NOT(
+        usedMemSize <= backendMap_[backendName].memSize,
+        llvm::formatv(
+            "Partition failed: the memory usage({0}) of one partitoion exceeds "
```
rdzhabarov: typo

tests/unittests/PartitionerTest.cpp
```diff
@@ -790,3 +790,22 @@ TEST_F(PartitionerTest, graphMemInfoCalculation2) {
...
+
+/// This one test the memoryUsageValidation in Partitioner : the memory usage of
+/// one single node is larger than the given device memory.
+TEST_F(PartitionerTest, memoryUsageValidation1) {
+  auto *input1 =
+      mod_.createPlaceholder(ElemKind::FloatTy, {2, 10}, "input1", false);
+  auto *input2 =
+      mod_.createPlaceholder(ElemKind::FloatTy, {10, 16}, "input2", false);
+  auto *mul0 = F_->createMatMul("mul0", input1, input2);
+  auto *save = F_->createSave("ret", mul0);
```
rdzhabarov: do we need save node if there is no compilation here?

beicy: No, the "save" node is not necessary. Just want to contain more than one node in a function :)

rdzhabarov: it's fine, maybe just do F_->createSave("ret", mul) and remove (void)save.

beicy:
> Would it be possible to detect this sort of failure earlier and rebuild partition to fit on all devices?

Technically, this validation can be added at any stage after the initial mapping is created.  For the single node too large error,  we can add an separate check at the very beginning. Here I just use it as a test for the validation.

include/glow/Partitioner/Partitioner.h
beicy: Discussed offline. Since some of the members of class `NodeToFunctionMap` are not marked as `const`, adding `const` here would cause compile error.  Will have a separate PR  to clean up the 
`const` issue in Partitioner.

PR #3143, opened 2019-06-19
[Operator] Exponential operator support

ksaurabh-cadence:
**Summary**
Added support for exp operator 

**Documentation**
Interpreter supports: Float and FP16
CPUBackend supports: Float

**Test Plan**
added operator test 
added stateless test for exp float16

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -478,6 +478,10 @@ public:
   /// Result type will be implicitly set based on the \p input type.
   TanhNode *createTanh(llvm::StringRef name, NodeValue input);
 
+  /// Create an Exp  node with \p name, which calculates element-wise
+  /// exponential of \p input.
+  ExpNode *createExp(llvm::StringRef name, NodeValue input);
+
   /// Create a Log node with \p name, which calculates element-wise natural log
   /// of \p input, with output type \p outTy.
   LogNode *createLog(llvm::StringRef name, NodeValue input,
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -265,6 +265,15 @@ protected:
     return llvm::Error::success();
   }
 
+  llvm::Error loadExp(const OpType &op, ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
+    auto *E = G_.createExp(opName, in);
+    RETURN_IF_ERR(addNodeAsOutput(op, E));
+    return llvm::Error::success();
+  }
+
   llvm::Error loadShape(const OpType &op, ArgumentDictionaryTy &dict) {
     NodeValue in;
     ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
@@ -1003,6 +1012,10 @@ protected:
     }
     if (typeName == "Tanh") {
       RETURN_IF_ERR(loadTanh(op, dict));
+      return true;
+    }
+    if (typeName == "Exp") {
+      RETURN_IF_ERR(loadExp(op, dict));
       return true;
     }
     if (typeName == "Shape") {
--- a/lib/Graph/Nodes.cpp
+++ b/lib/Graph/Nodes.cpp
@@ -636,6 +636,18 @@ bool TanhGradNode::verify() const {
   return isValid;
 }
 
+bool ExpNode::verify() const {
+  const Node *parent = getResult().getNode();
+  bool isValid =
+      checkSameIsQuantized(getInput().getType(), getResult().getType(), parent);
+  if (getInput().getType()->isQuantizedType()) {
+    return false;
+  }
+  isValid &= checkSameType(getInput(), getResult(), parent);
+  isValid &= checkSameShape(getInput(), getResult(), parent);
+  return isValid;
+}
+
 bool SoftMaxNode::verify() const {
   return verifySoftMax(getInput(), getResult());
 }
--- a/tests/models/onnxModels/exp.onnxtxt
+++ b/tests/models/onnxModels/exp.onnxtxt
@@ -1,0 +1,58 @@
+ir_version: 5
+producer_name: "onnx-example"
+graph {
+  node {
+    input: "data"
+    output: "result"
+    op_type: "Exp"
+  }
+ name: "test-exp"
+  input {
+    name: "data"
+    type {
+      tensor_type {
+        elem_type: 1
+        shape {
+          dim {
+            dim_value: 1
+          }
+          dim {
+            dim_value: 2
+          }
+          dim {
+            dim_value: 4
+          }
+          dim {
+            dim_value: 3
+          }
+        }
+      }
+    }
+  }
+  output {
+    name: "result"
+    type {
+      tensor_type {
+        elem_type: 1
+        shape {
+          dim {
+            dim_value: 1
+          }
+          dim {
+            dim_value: 2
+          }
+          dim {
+            dim_value: 4
+          }
+          dim {
+            dim_value: 3
+          }
+        }
+      }
+    }
+  }
+}
+opset_import {
+  version: 10
+}
+
--- a/tests/unittests/Caffe2ImporterTest.cpp
+++ b/tests/unittests/Caffe2ImporterTest.cpp
@@ -24,6 +24,42 @@
 #endif
 
 using namespace glow;
+/// Test loading of Elementwise Unary Ops floating point.
+static void testEltwiseUnaryOpFloat(std::string fileName,
+                                    llvm::ArrayRef<size_t> inputShape,
+                                    std::string input_name, float delta,
+                                    const std::function<float(float)> &op) {
+  ExecutionEngine EE{};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+  std::string NetDescFilename =
+      std::string(GLOW_DATA_PATH "tests/models/caffe2Models/") + fileName;
+  std::string NetWeightFilename(
+      GLOW_DATA_PATH "tests/models/caffe2Models/empty_init_net.pbtxt");
+
+  PlaceholderBindings bindings;
+  Placeholder *graphOutputVar;
+  Type input_type(ElemKind::FloatTy, inputShape);
+  Caffe2ModelLoader caffe2LD(NetDescFilename, NetWeightFilename,
+                             {input_name.c_str()}, {&input_type}, *F);
+  graphOutputVar = EXIT_ON_ERR(caffe2LD.getSingleOutput());
+  auto PH = mod.getPlaceholderByName(input_name);
+  auto *inTensor = bindings.allocate(PH);
+  inTensor->getHandle().randomize(-10.0, 10.0, mod.getPRNG());
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+  auto result = bindings.get(graphOutputVar)->getHandle();
+  auto inHandle = inTensor->getHandle();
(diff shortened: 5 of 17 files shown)


tools/ClassGen/InstrGen.cpp
```diff
@@ -440,6 +440,17 @@ int main(int argc, char **argv) {
...
 
+  BB.newInstr("ElementExp")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("Src", OperandKind::In)
+      .inplaceOperand({
+          "Dest",
+          "Src",
+      })
+      .dataParallel()
+      .autoVerify(VerifyKind::SameType, {"Dest", "Src"})
```
rdzhabarov: a bit confusing, why do you have auto verify and also custom verification logic for this node.

lib/Graph/Nodes.cpp
```diff
@@ -636,6 +648,8 @@ bool TanhGradNode::verify() const {
   return isValid;
 }
 
+bool ExpNode::verify() const { return verifyExp(getInput(), getResult()); }
```
jfix71: nit: No need to use a helper function `verifyExp()` here, since it's only being used by this one Node.

tools/ClassGen/InstrGen.cpp
jfix71: This is for the Instr, the other verification is for the Node.

tools/ClassGen/NodeGen.cpp
```diff
@@ -311,6 +311,9 @@ int main(int argc, char **argv) {
       .setDocstring("Performs element-wise natural log to the Input.");
   // clang-format on
 
+  BB.newNode("Exp").addInput("Input").addResultFromCtorArg().setDocstring(
```
jfix71: Can you  disable clang-format for this Node since it gets formatted all weird? See just above for "Log".

tools/ClassGen/InstrGen.cpp
rdzhabarov: right, missed that

lib/Graph/Nodes.cpp
ksaurabh-cadence: Intention was to future proof this for Grad implementation, as was being done by the other nodes of similar kind. As a policy do you encourage future proofing?

jfix71: If all you intend to add grad support in the near future then it's fine to keep, NBD. Otherwise it's not very hard to modify in the future if/when we add grad support, and so I'd say to keep it inline since it only adds a level of indirection without benefit.

lib/Graph/Nodes.cpp
```diff
@@ -321,6 +321,18 @@ static bool verifyTanh(NodeValue src, NodeValue dest) {
   return isValid;
 }
 
+static bool verifyExp(NodeValue src, NodeValue dest) {
+  const Node *parent = dest.getNode();
+  bool isValid = checkSameIsQuantized(src.getType(), dest.getType(), parent);
+  if (src.getType()->isQuantizedType()) {
```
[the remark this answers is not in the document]
ksaurabh-cadence: The intention was future proofing. I chose to keep verifyExp very general. Individual backends can return unsupported errors for quantized inputs. So, if somebody adds a quant implementation later, he probably won't have to modify verifyExp. Is something wrong with my understanding here?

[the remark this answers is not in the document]
rdzhabarov: >  So, if somebody adds a quant implementation later, he probably won't have to modify verifyExp. Is something wrong with my understanding here?

I'm not sure that quantized exponent would make sense (huge accuracy loss). Do you have a use case where this is needed?

include/glow/Graph/Graph.h
```diff
@@ -478,6 +478,11 @@ class Function final : public Named {
   /// Result type will be implicitly set based on the \p input type.
   TanhNode *createTanh(llvm::StringRef name, NodeValue input);
 
+  /// Create an Exp  node with \p name, which calculates element-wise
+  /// exponential of \p input, with output type \p outTy.
+  ExpNode *createExp(llvm::StringRef name, NodeValue input,
```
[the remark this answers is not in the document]
ksaurabh-cadence: @rdzhabarov  Thanks for a quick review.  Looking at the current state of tests, I get an impression that the tests are added only for ops where there are attributes involved, or some special input/output scenarios. There is no test for log, tanh, sigmoid, which have similar template. So, I am not clear what's the criteria?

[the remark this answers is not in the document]
rdzhabarov: > Looking at the current state of tests, I get an impression that the tests are added only for ops where there are attributes involved

might be a coincidence. 

Output type is often used during the quantization of the graph so you can set different scale and offset for input/output.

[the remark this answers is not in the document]
ksaurabh-cadence: Regarding the outTy in function signature: 
I tried to model this upon Log which uses the same template.

Regarding the onnxImporterTest/caffe2ImporterTest:
I am not convinced if we really need a test for simple operators like log/exp/tanH etc. (none of them exist as of now), where there is nothing really to test in terms of parsing just the operator name. As far as functionality is concerned it's already being tested in the OperatorTest. Please let me know if there exists a simple test builder for such kind of tests.

lib/Graph/Nodes.cpp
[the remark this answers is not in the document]
ksaurabh-cadence: I modeled this on LogNode (though not completely). I won't rule out quantization of exp outright based on it's usage in detection networks for box decoding. If you are insist on YAGNI I will remove it from the merge.

lib/Graph/Nodes.cpp
ksaurabh-cadence: OK, makes sense.

tools/ClassGen/NodeGen.cpp
ksaurabh-cadence: will do that.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -265,6 +265,15 @@ class CommonOperatorLoader : public ProtobufLoader {
     return llvm::Error::success();
   }
 
+  llvm::Error loadExp(const OpType &op, ArgumentDictionaryTy &dict) {
+    const std::string &opName = loadOperatorName(op);
+    NodeValue in;
+    ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
+    auto *T = G_.createExp(opName, in);
```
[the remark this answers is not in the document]
ksaurabh-cadence: agree

include/glow/Graph/Graph.h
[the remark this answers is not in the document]
rdzhabarov: missing tests for log/exp/tanH is not a feature but a bug :). I think it makes sense to test loading logic otherwise how do you know that you load it properly ;).

lib/Graph/Nodes.cpp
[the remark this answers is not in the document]
rdzhabarov: If there is valid case when quantized op is used, I'd keep it here.

include/glow/Graph/Graph.h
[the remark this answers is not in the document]
rdzhabarov: > I tried to model this upon Log which uses the same template.

we can add this when it's needed. I'd remove now.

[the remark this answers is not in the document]
ksaurabh-cadence: I think we need a generic function for testing of loading Unary OPs. Strange it was never added when so many unary operators already exist. I will add that so that it benefits testing of other unary ops as well in future.

ksaurabh-cadence:
@rdzhabarov Comments have been addressed. Please also review the unit test for unary operator loader.

rdzhabarov:
@ksaurabh-cadence I'll review this today. Thanks for working on this!

tools/ClassGen/NodeGen.cpp
```diff
@@ -309,6 +309,11 @@ int main(int argc, char **argv) {
       .addInput("Input")
       .addResultFromCtorArg()
       .setDocstring("Performs element-wise natural log to the Input.");
+
+  BB.newNode("Exp")
+    .addInput("Input")
```
rdzhabarov: can you make formatting similar to the above Pow and Low.

tests/unittests/OperatorTest.cpp
```diff
@@ -4426,6 +4426,51 @@ TEST_P(OperatorTest, Tanh) {
   }
 }
 
+static FunctionTensorPair
+createAndInitBasicExpTest(glow::PlaceholderBindings &bindings,
```
rdzhabarov: can you merge this with createAndInitBasicTanhTest and make it more generic?

essentially it's all the same except F->createExp("Exp", input); call.

(reuse that for tanh and for exp).

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -255,6 +255,40 @@ TEST(onnx, importUniBroadcastMultiOutput) {
...
+  graphOutputVar = EXIT_ON_ERR(onnxLD.getSingleOutput());
+  auto PH = mod.getPlaceholderByName(input_name);
+  auto *inTensor = bindings.allocate(PH);
+  inTensor->getHandle().randomize(-10.0, 10.0, mod.getPRNG());
+  // Compile&run the graph, and check the output
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+  auto result = bindings.get(graphOutputVar)->getHandle();
+  auto inHandle = inTensor->getHandle();
+  EXPECT_TRUE(result.dims() == inputShape);
```
rdzhabarov: strictly speaking this should be ASSERT_TRUE, otherwise the loop below will segfault.

tests/unittests/Caffe2ImporterTest.cpp
```diff
@@ -24,6 +24,42 @@
...
+                             {input_name.c_str()}, {&input_type}, *F);
+  graphOutputVar = EXIT_ON_ERR(caffe2LD.getSingleOutput());
+  auto PH = mod.getPlaceholderByName(input_name);
+  auto *inTensor = bindings.allocate(PH);
+  inTensor->getHandle().randomize(-10.0, 10.0, mod.getPRNG());
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+  auto result = bindings.get(graphOutputVar)->getHandle();
+  auto inHandle = inTensor->getHandle();
+  EXPECT_TRUE(result.dims() == inputShape);
```
rdzhabarov: same here.

tests/models/onnxModels/exp.onnxtxt
```diff
@@ -0,0 +1,58 @@
...
+    output: "result"
+    op_type: "Exp"
+  }
+ name: "test-exp"
+  input {
+    name: "data"
+    type {
+      tensor_type {
+        elem_type: 1
+        shape {
```
rdzhabarov: it's up to you, but for simple test you could use simpler share (less chars to type ;))

ksaurabh-cadence:
@rdzhabarov Is it necessary to address all the comments?

rdzhabarov:
> @rdzhabarov Is it necessary to address all the comments?

those should be quick to fix. (some of those which were up to you - is up to you :))

tests/models/onnxModels/exp.onnxtxt
ksaurabh-cadence: not changing it would be even less chars to type ;-)

tests/unittests/OperatorTest.cpp
ksaurabh-cadence: It's not just Exp, in fact TanH, Log, Sigmoid all can be combined, and strange it was never observed before. In future I'd rather have it addressed as a separate MR for test cleanup.

ksaurabh-cadence:
@rdzhabarov Everything addressed. Please approve.

PR #3144, opened 2019-06-19
[Operator] Add int64, int32 datatype support for ConstantOfShape operator

ksaurabh-cadence:
Add int64, int32 datatype support for ConstantOfShape operator

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -1059,11 +1059,15 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
   if (dict.count("value")) {
     RETURN_IF_ERR(loadTensor(dict.at("value")->t(), &T));
     RETURN_ERR_IF_NOT(T.dims().size() == 1, "Value must be a 1D vector.");
-    RETURN_ERR_IF_NOT(T.getType().getElementType() == ElemKind::FloatTy,
-                      "Value must have float type");
+    RETURN_ERR_IF_NOT(T.getType().getElementType() == ElemKind::FloatTy ||
+                          T.getType().getElementType() == ElemKind::Int64ITy ||
+                          T.getType().getElementType() == ElemKind::Int32ITy,
+                      T.getType().getElementName().str() +
+                          " type Value is not supported.");
   }
 
   TypeRef ty;
+  Node *SN = nullptr;
   if (op.input_size() > 0) {
     Constant *in;
     ASSIGN_VALUE_OR_RETURN_ERR(in, getConstantByName(op.input(0)));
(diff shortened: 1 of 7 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1075,12 +1078,28 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
...
                                          (const size_t *)TH.end()};
-    ty = G_.getParent()->uniqueType(ElemKind::FloatTy, outputDims);
+    switch (T.getType().getElementType()) {
+    case ElemKind::FloatTy:
+      ty = G_.getParent()->uniqueType(ElemKind::FloatTy, outputDims);
+      SN = G_.createSplat(loadOperatorName(op), ty, T.getHandle().raw(0));
+      break;
+    case ElemKind::Int64ITy:
+      ty = G_.getParent()->uniqueType(ElemKind::Int64ITy, outputDims);
+      SN = G_.createSplat(loadOperatorName(op), ty,
```
rdzhabarov: This will not work.

createSplat takes floating point value of type `float`. Float cannot represent integer number bigger than 16 777 216 exactly (you'll lose precision).

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1075,12 +1078,28 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
...
     llvm::ArrayRef<size_t> outputDims = {(const size_t *)TH.begin(),
                                          (const size_t *)TH.end()};
-    ty = G_.getParent()->uniqueType(ElemKind::FloatTy, outputDims);
+    switch (T.getType().getElementType()) {
+    case ElemKind::FloatTy:
+      ty = G_.getParent()->uniqueType(ElemKind::FloatTy, outputDims);
+      SN = G_.createSplat(loadOperatorName(op), ty, T.getHandle().raw(0));
+      break;
+    case ElemKind::Int64ITy:
+      ty = G_.getParent()->uniqueType(ElemKind::Int64ITy, outputDims);
```
rdzhabarov: can you lift this up before switch (you have elem kind from T.getType().getElementType()).

rdzhabarov:
Once the issue is fixed, we'd need a unit test to make sure all works fine.

lib/Importer/ONNXModelLoader.cpp
ksaurabh-cadence: You are right. Actually this can happen with int32 as well.  I will put this check -

      int64_t v = T.getHandle<int64_t>().raw(0);
      RETURN_ERR_IF_NOT(v ==(static_cast<int64_t>(static_cast<float>v)),
                        "This ConstantOfShape implementation may cause losses for value " +
                            std::to_string(v)  + " .");

lib/Importer/ONNXModelLoader.cpp
ksaurabh-cadence: I didn't lift it up to take care of the the default scenario, which now kind of looks redundant because we already have a check in the beginning of the function. Will do that and remove the default scenario.

lib/Importer/ONNXModelLoader.cpp
rdzhabarov: > Actually this can happen with int32 as well.

totally, since it's just 16e6 which causes issues. That sounds reasonable short term.

ksaurabh-cadence:
@rdzhabarov  All comments have been addressed. Relevant tests have been added.

rdzhabarov:
Thanks @ksaurabh-cadence, will take a look today.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1271,22 +1271,46 @@ TEST(onnx, constant) {
...
+template <class ElemType>
+static void testConstantOfShape(std::string fileName, ElemType ref) {
   ExecutionEngine EE;
   auto &mod = EE.getModule();
   auto *F = mod.createFunction("main");
-  std::string netFilename(GLOW_DATA_PATH
-                          "tests/models/onnxModels/constantOfShape.onnxtxt");
+  PlaceholderBindings bindings;
+
+  std::string NetFilename =
```
rdzhabarov: nit: netFileName.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1271,22 +1271,46 @@ TEST(onnx, constant) {
   ASSERT_EQ(F->getNodes().size(), 1);
 }
 
-/// Test loading ConstantOfShape from an ONNX model.
-TEST(onnx, constantOfShape) {
+/// Test loading of Elementwise Unary Ops floating point.
```
rdzhabarov: fix comment.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1075,12 +1079,34 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
...
+          v == static_cast<int64_t>(static_cast<float>(v)),
+          "This ConstantOfShape implementation may cause losses for value" +
+              std::to_string(v) + " .");
+      SN = G_.createSplat(loadOperatorName(op), ty, v);
+      break;
+    }
+    case ElemKind::Int32ITy: {
+      int32_t v = T.getHandle<int32_t>().raw(0);
+      RETURN_ERR_IF_NOT(
+          v == static_cast<int32_t>(static_cast<float>(v)),
```
rdzhabarov: could you please add a test when this fails as well.

ksaurabh-cadence: added

ksaurabh-cadence:
@rdzhabarov All the comments have been addressed.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1287,6 +1290,46 @@ TEST(onnx, constantOfShape) {
...
+TEST(onnx, importConstantOfShapeInt64) {
+  testConstantOfShape<int64_t>("constantOfShapeInt64.onnxtxt", 16777216LL);
+}
+
+TEST(onnx, importConstantOfShapeInt64LossFailure) {
+  testConstantOfShapeFailure<int64_t>("constantOfShapeInt64Fail.onnxtxt");
+}
+
+TEST(onnx, importConstantOfShapeInt32LossFailure) {
+  testConstantOfShapeFailure<int64_t>("constantOfShapeInt32Fail.onnxtxt");
```
rdzhabarov: int32_t here in the template arg.

ksaurabh-cadence: fixed

PR #3146, opened 2019-06-20
[Operator] Support Max pool with Argmax output

### LINKED ISSUE #3053: Question: MaxPoolWithArgmax - new node or new output to existing node?
Hi, I'm working on #2974, and I have a design question to discuss before I come up with final version. In ONNX, MaxPool has an optional output that calculates ArgMax that we need to support. In GLOW, I did not find a mechanism to specify that an output is optional for some node. It means that if we have specified the output, we should at least initialize it correctly.

There are two possible ways to implement this optional output.

Approach 1 (which I currently have):
I've introduced a new node MaxPoolWithArgmax and did not touch the existing node MaxPool. When we parse ONNX file, we create one or another node, depending on whether the 2nd output is specified. Then I hacked the computation methods in CPU and Interpreter backends so that the same functions can be used to handle both kinds of nodes.
The upside of this approach is that it does not break any existing code.
The downside is that we have some code duplication in sense of node declaration.

Approach 2 (which we internally discuss):
We could instead tweak MaxPool node to have 2 outputs, add a Boolean field signaling whether or not the 2nd output is actually calculated. It means that we should always initialize this 2nd output with some tensor, even if we never use it.
The upside of that is less code duplication, and node's API will match those in ONNX.
The downsides are:
- We have this dummy type initialization whenever we create a node;
- Other (maybe downstream) backends may not compile if they create these nodes inside, so this may disrupt someone's work.
- If some backend is not aware that there can be this 2nd output, this may cause a more sneaky bug than just failure with unsupported node kind.

What do you guys think would be a better way to do it?

Max-K-cadence:
Summary:
This patch adds support for Argmax output of MaxPool node which is present in ONNX.
In order to do so, it replaces MaxPoolWithXY instructions with MaxPoolWithArgmax which
contains linearized NCHW indices of respective max elements. X and Y may be extracted
from this index using simple formula.

Fixes #3053

Test Plan:
Added node and ONNX unit tests.

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -1141,6 +1141,37 @@ protected:
     return false;
   }
 
+  /// Utility function which computes the resulting shape in case of
+  /// multidirectional broadcasting.
+  llvm::Expected<std::vector<size_t>>
+  computeMultidirectionalBroadcast(llvm::ArrayRef<size_t> shape0,
+                                   llvm::ArrayRef<size_t> shape1) {
+    size_t numDims0 = shape0.size();
+    size_t numDims1 = shape1.size();
+    size_t newNumDims = numDims0 > numDims1 ? numDims0 : numDims1;
+    std::vector<size_t> reshapeDims(newNumDims);
+
+    for (size_t i = 0; i < newNumDims; i++) {
+      reshapeDims[i] = 1;
+    }
+    RETURN_IF_ERR(mergeMultidirectionalBroadcast(reshapeDims, shape0));
+    RETURN_IF_ERR(mergeMultidirectionalBroadcast(reshapeDims, shape1));
+
+    return reshapeDims;
+  }
+
+  /// Associate all outputs of \p op with nodes in \p NVs. Number of outputs of
+  /// \p op should match the number of elements of \p NVs.
+  /// \returns error code in case of error.
+  llvm::Error assignNodeOutputs(const OpType &op,
+                                llvm::ArrayRef<NodeValue> NVs) {
+    RETURN_ERR_IF_NOT(NVs.size() == op.output_size(), "Output size mismatch.");
+    for (size_t i = 0; i < NVs.size(); i++) {
+      nodeValueByName_[op.output(i)] = NVs[i];
+    }
+    return llvm::Error::success();
+  }
+
   /// Load pre-trained weights from \p weightDescriptors.
   llvm::Error loadWeights(uint32_t weightsCount,
                           const onnxTensorDescriptorV1 *weightDescriptors) {
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -571,10 +571,10 @@ static void libjit_max_pool_generic(const T *inW, T *outW,
 }
 
 template <typename T>
-static void libjit_max_pool_xy_generic(const T *inW, T *outW, size_t *inXY,
-                                       const size_t *inWdims,
-                                       const size_t *outWdims, size_t *kernels,
-                                       size_t *strides, size_t *pads) {
+static void
+libjit_max_pool_argmax_generic(const T *inW, T *outW, int64_t *argmax,
+                               const size_t *inWdims, const size_t *outWdims,
+                               size_t *kernels, size_t *strides, size_t *pads) {
   size_t pad_t = pads[0];
   size_t pad_l = pads[1];
   size_t stride_h = strides[0];
@@ -592,8 +592,7 @@ static void libjit_max_pool_xy_generic(const T *inW, T *outW, size_t *inXY,
 
         // For each channel in the output tensor:
         for (size_t z = 0; z < outWdims[3]; z++) {
-          size_t maxX = x;
-          size_t maxY = y;
+          int64_t argmaxNHWC = 0;
           int first = 1;
           T max = 0;
 
@@ -606,24 +605,20 @@ static void libjit_max_pool_xy_generic(const T *inW, T *outW, size_t *inXY,
                   oy >= (ssize_t)inWdims[2]) {
                 continue;
               }
-
-              T val =
-                  inW[libjit_getXYZW(inWdims, n, (size_t)ox, (size_t)oy, z)];
+              const size_t flatIndex =
+                  libjit_getXYZW(inWdims, n, (size_t)ox, (size_t)oy, z);
+              T val = inW[flatIndex];
               if (first || (val >= max)) {
                 first = 0;
                 max = val;
-                maxX = ox;
-                maxY = oy;
+                argmaxNHWC = flatIndex;
               }
             }
           }
 
-          outW[libjit_getXYZW(outWdims, n, ax, ay, z)] = max;
-          // For the x and y argmax's, we use a 5-dimensional
-          // tensor whose fifth dimension has size 2:
-          size_t ix = 2 * libjit_getXYZW(outWdims, n, ax, ay, z);
-          inXY[ix] = maxX;
-          inXY[ix + 1] = maxY;
+          const size_t flatIndex = libjit_getXYZW(outWdims, n, ax, ay, z);
+          outW[flatIndex] = max;
+          argmax[flatIndex] = argmaxNHWC;
         } // C
       }   // W
     }     // H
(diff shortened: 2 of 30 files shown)


Max-K-cadence:
I'm not sure if failure of DEBUG is an infrastructure issue or a real failure, I can't reproduce it.

jfix71:
>I'm not sure if failure of DEBUG is an infrastructure issue or a real failure, I can't reproduce it.

This may just be due to flakiness. CC @rdzhabarov

Max-K-cadence:
I found one thing that was a real bug in OpenCL, but other tests are still failing. Going to investigate that. Unfortunately I don't have OpenCL on my side, so I'm kind of blindfolded.

Max-K-cadence:
I've squashed fix for isOpSupported in OpenCL; for other tests the failure messages don't give me any info to investigate. :(
@rdzhabarov could you please confirm that those are real failures and not infrastructure problem? I'd also appreciate any info about them, because I don't have OpenCL deployed on my side and it can be problematic to get it.

Max-K-cadence:
Failure looks similar to what is described here: https://devtalk.nvidia.com/default/topic/1027077/container-pytorch/-quot-unexpected-end-of-proc-mounts-line-overlay-quot-on-p3-8xlarge/

Max-K-cadence:
Put WIP. Likely something is wrong with kernel names in OpenCI. Cannot debug because of bug with 
../lib/Backends/OpenCL/OpenCL.cpp:462:58: error: too few arguments provided to function-like macro invocation

Will continue investigation next week.

Max-K-cadence:
Okay, I found what the problem was. OpenCL Kernel file did not recognize standard types. :)

Max-K-cadence:
Every round of testing shows different failures... Weird. Putting WIP back until it converges to something.

Max-K-cadence:
Yay, it's finally passing. :) We can remove WORK IN PROGRESS now.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -688,13 +683,29 @@ llvm::Error ONNXModelLoader::loadPool(const ONNX_NAMESPACE::NodeProto &op,
...
   Node *node = nullptr;
-  if (typeName == "MaxPool") {
+  if (op.output_size() > 1) {
+    if (typeName != "MaxPool") {
+      RETURN_ERR("Argmax output is only supported for MaxPool!",
+                 GlowErr::ErrorCode::MODEL_LOADER_UNSUPPORTED_OPERATOR);
+    }
+
     node = G_.createMaxPool(opName, tr, kernels, strides, pads);
+    auto *Res = G_.createTranspose(opName, NodeValue(node, 0), NHWC2NCHW);
```
jfix71: nit: We usually use `lowerCamelCase`, e.g. `res` here, `argmax` on next line, etc.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -1152,6 +1152,19 @@ class CommonOperatorLoader : public ProtobufLoader {
     return reshapeDims;
   }
 
+  /// Associate all outputs of \p node with consecutive set of outputs of \p op,
+  /// starting from op's output with index \p startingFrom.
+  llvm::Error appendNodeAsOutput(const OpType &op, Node *node,
```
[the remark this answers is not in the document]
Max-K-cadence: Well, we could make a manual setup of outputs, but I thought that having such helper might be useful in the future. Do you see use in making it manually?

lib/Importer/ONNXModelLoader.cpp
Max-K-cadence: Right, I have no idea why I named it like this here. :)

lib/Optimizer/IROptimizer.cpp
```diff
@@ -1518,13 +1518,13 @@ void performPeepholeOptimizations(IRFunction &M) {
     auto cur = it;
     auto *I = &*cur;
     it = std::next(it);
-    // MaxPoolWithXYInst -> MaxPoolInst.
-    if (auto *PMI = dyn_cast<MaxPoolWithXYInst>(I)) {
-      auto *SrcXY = PMI->getSrcXY();
+    // MaxPoolWithArgmaxInst -> MaxPoolInst.
+    if (auto *PMI = dyn_cast<MaxPoolWithArgmaxInst>(I)) {
+      auto *Argmax = PMI->getArgmax();
```
Max-K-cadence: Same here.

tools/ClassGen/NodeGen.cpp
```diff
@@ -125,10 +125,13 @@ int main(int argc, char **argv) {
...
-      .addResultFromCtorArg()
+      .addResultFromCtorArg("Result")
+      .addResultFromCtorArg("Argmax")
       .addGradient()
-      .setDocstring("Performs a Max Pool operation on the Input given provided "
-                    "Kernels, Strides, and Pads.");
+      .setDocstring(
+          "Performs a Max Pool with Argmax operation on the Input "
+          "given provided Kernels, Strides, and Pads. Argmax is a flattened "
+          "NCHW index corresponding to respective max element.");
```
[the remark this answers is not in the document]
Max-K-cadence: That's exactly how it works in ONNX format: https://github.com/onnx/onnx/blob/master/docs/Operators.md#MaxPool

If we need NHWC linearization for some reason, we can later add a flag switching between them. I didn't do it right now to not overcomplicate, but can do by request.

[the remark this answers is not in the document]
Max-K-cadence: The more I think about it, the less convinced I become. ONNX models should have their input in NCHW and then transposed, therefore NHWC index in transposed version should be the index we need after we transform it back to NCHW. I need to play a bit around it.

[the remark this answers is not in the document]
Max-K-cadence: You were right. I need to fix that.

Max-K-cadence:
Replaced flattening with NHWC, made sure that it is what works on network I am enabling.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -612,18 +611,14 @@ static void libjit_max_pool_xy_generic(const T *inW, T *outW, size_t *inXY,
               if (first || (val >= max)) {
                 first = 0;
                 max = val;
-                maxX = ox;
-                maxY = oy;
+                argmaxNHWC =
+                    ((n * inWdims[1] + ox) * inWdims[2] + oy) * inWdims[3] + z;
```
jfix71: nit: This is the same as `libjit_getXYZW(inWdims, n, ox, oy, z)`, you can just call that instead.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -1395,11 +1390,10 @@ void libjit_max_pool_xy_grad_f(float *inG, const float *outG,
 
       for (size_t ax = 0; ax < outWdims[1]; ax++) {
         for (size_t ay = 0; ay < outWdims[2]; ay++) {
-          // For the x and y argmax's, we use a 5-dimensional
-          // tensor whose fifth dimension has size 2:
-          size_t ix = 2 * libjit_getXYZW(outWdims, n, ax, ay, z);
-          size_t maxX = inXY[ix];
-          size_t maxY = inXY[ix + 1];
+          int64_t argmaxNHW =
```
jfix71: nit: Can you add comments about what's going on here?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -917,8 +921,9 @@ void BoundInterpreterFunction::fwdMaxPoolWithXYGradInst(
 
           float chainGrad = outG.at({n, ax, ay, z});
 
-          size_t maxX = SXY.at({n, ax, ay, z, 0});
-          size_t maxY = SXY.at({n, ax, ay, z, 1});
+          int64_t argmaxNHW = argmax.at({n, ax, ay, z}) / idim.c;
```
jfix71: nit: Can you add comments about what's going on here?

lib/Backends/OpenCL/kernels.cl
```diff
@@ -1307,11 +1308,10 @@ maxpoolwithxygradK(__global float *dest, __global cl_uint64_t *srcXY,
 
     for (size_t ax = 0; ax < destGradDim.h; ax++) {
       for (size_t ay = 0; ay < destGradDim.w; ay++) {
-        // For the x and y argmax's, we use a 5-dimensional
-        // tensor whose fifth dimension has size 2:
-        size_t ix = 2 * getNHWC(destGradDim, n, ax, ay, z);
-        size_t maxX = srcXY[ix];
-        size_t maxY = srcXY[ix + 1];
+        cl_uint64_t argmaxNHW =
```
jfix71: nit: Can you add comments about what's going on here?

lib/Backends/CPU/libjit/libjit.cpp
Max-K-cadence: I just realized that this whole thing makes no sense. We extract X and Y coordinates only to recompute flattened index (which is already stored in argmax). This can be done much simpler. :)

Max-K-cadence:
Thanks for thorough review. :) Now the logic in grad node looks much more transparent.

Max-K-cadence:
These failures don't look like something real. :)

PR #3147, opened 2019-06-20
[GraphOptimizer] Introduce FunctionPass

jfix71:
Summary: Introduce the `FunctionPass`. This allows for pre/post optimization pass functionality, e.g. easier debugging like dumping some info before/after every optimization via `doInitialization()`/`doFinalization()`. In future PRs I would like to integrate these into some pass manager, which could then query a backend to determine what passes to run/disable.

Related to #1641

Documentation: Will add once this gets closer to landing.

Test Plan: Passes all tests.

### DIFF
--- a/include/glow/Optimizer/FunctionPass.h
+++ b/include/glow/Optimizer/FunctionPass.h
@@ -23,7 +23,8 @@ class Function;
 /// Class used for all passes over Functions. All passes over Functions should
 /// derive from this class, implementing the pass logic and additionally can add
 /// logic for running before and after the pass runs.
-struct FunctionPass {
+class FunctionPass {
+public:
   FunctionPass() = default;
   virtual ~FunctionPass() = default;
 
--- a/include/glow/Optimizer/FunctionPass.h
+++ b/include/glow/Optimizer/FunctionPass.h
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_OPTIMIZER_FUNCTIONPASS_H
+#define GLOW_OPTIMIZER_FUNCTIONPASS_H
+
+namespace glow {
+
+/// Class used for all passes over Functions. All passes over Functions should
+/// derive from this class, implementing the pass logic and additionally can add
+/// logic for running before and after the pass runs.
+struct FunctionPass {
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+
+  // Run the pass on \p F. \returns whether the pass modifies \p F.
+  virtual bool run(Function *F) = 0;
+
+  /// \returns the name of the pass.
+  virtual llvm::StringRef getName() const = 0;
+};
+
+} // namespace glow
+
+#endif // GLOW_OPTIMIZER_FUNCTIONPASS_H
--- a/include/glow/Optimizer/FunctionPass.h
+++ b/include/glow/Optimizer/FunctionPass.h
@@ -18,6 +18,8 @@
 
 namespace glow {
 
+class Function;
+
 /// Class used for all passes over Functions. All passes over Functions should
 /// derive from this class, implementing the pass logic and additionally can add
 /// logic for running before and after the pass runs.
--- a/include/glow/Optimizer/FunctionPasses.h
+++ b/include/glow/Optimizer/FunctionPasses.h
@@ -21,7 +21,8 @@
 namespace glow {
 
 #define FUN_PASS(PASS_NAME)                                                    \
-  struct PASS_NAME : public FunctionPass {                                     \
+  class PASS_NAME : public FunctionPass {                                      \
+  public:                                                                      \
     bool run(Function *F) override;                                            \
     llvm::StringRef getName() const override { return #PASS_NAME; }            \
   };
--- a/include/glow/Optimizer/FunctionPasses.h
+++ b/include/glow/Optimizer/FunctionPasses.h
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_OPTIMIZER_FUNCTIONPASSES_H
+#define GLOW_OPTIMIZER_FUNCTIONPASSES_H
+
+#include "glow/Optimizer/FunctionPass.h"
+
+namespace glow {
+
+#define FUN_PASS(PASS_NAME)                                                    \
+  struct PASS_NAME : public FunctionPass {                                     \
+    bool run(Function *F) override;                                            \
+    llvm::StringRef getName() const override { return #PASS_NAME; }            \
+  };
+
+#include "FunctionPasses.def"
+
+} // namespace glow
+
+#endif // GLOW_OPTIMIZER_FUNCTIONPASSES_H
--- a/include/glow/Optimizer/Optimizer.h
+++ b/include/glow/Optimizer/Optimizer.h
@@ -17,6 +17,7 @@
 #define GLOW_OPTIMIZER_OPTIMIZER_H
 
 #include "glow/Optimizer/CompilationContext.h"
+#include "glow/Optimizer/FunctionPass.h"
 #include "glow/Support/Error.h"
 
 #include "llvm/ADT/ArrayRef.h"
@@ -51,7 +52,12 @@ void lower(Function *F, CompilationContext &cctx, const Backend *B = nullptr,
            const KindSet &doNotLowerKinds = {});
 
 /// Dead code elimination.
-void DCE(Function *F);
+struct DCE : public FunctionPass {
+  // Run DCE on \p F. \returns whether the pass modifies \p F.
+  bool run(Function *F) override;
+
+  llvm::StringRef getName() const override { return "DCE"; }
+};
 
 /// Convert placeholders in Module \p M to constants based on the values in \p
 /// bindings.  Do not convert any placeholders explicitly listed in \p vars.
--- a/include/glow/Optimizer/Optimizer.h
+++ b/include/glow/Optimizer/Optimizer.h
@@ -17,7 +17,6 @@
 #define GLOW_OPTIMIZER_OPTIMIZER_H
 
 #include "glow/Optimizer/CompilationContext.h"
-#include "glow/Optimizer/FunctionPass.h"
 #include "glow/Support/Error.h"
 
 #include "llvm/ADT/ArrayRef.h"
@@ -50,14 +49,6 @@ void fold(Function *F, CompilationMode mode);
 /// a set of NodeKinds which represents all Nodes that should not be lowered.
 void lower(Function *F, CompilationContext &cctx, const Backend *B = nullptr,
            const KindSet &doNotLowerKinds = {});
-
-/// Dead code elimination.
-struct DCE : public FunctionPass {
-  // Run DCE on \p F. \returns whether the pass modifies \p F.
-  bool run(Function *F) override;
-
-  llvm::StringRef getName() const override { return "DCE"; }
-};
 
 /// Convert placeholders in Module \p M to constants based on the values in \p
 /// bindings.  Do not convert any placeholders explicitly listed in \p vars.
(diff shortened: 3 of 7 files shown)


jfix71:
This is a WIP, would love some feedback. I will extend it to other GraphOptimization passes once a design is settled on.

beicy:
The name "FunctionPass" is a little bit confused to me. Maybe just "OptimizationPass" ?

include/glow/Optimizer/Optimizer.h
```diff
@@ -51,7 +52,13 @@ void lower(Function *F, CompilationContext &cctx, const Backend *B = nullptr,
            const KindSet &doNotLowerKinds = {});
 
 /// Dead code elimination.
-void DCE(Function *F);
+class DCE : public FunctionPass {
+private:
+  bool passImpl(Function *F);
```
beicy: I haven't seen where the return value is used.  Will it used in the future? If so, maybe some comments could be added.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,52 @@
...
+
+  /// Implementation of the pass. \returns true if the pass modifies \p F.
+  virtual bool passImpl(Function *F) = 0;
+
+public:
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+
+  // Run the pass, including initialization and finalization.
+  virtual bool runOnFunction(Function *F) {
```
beicy: Same here, it is better to add the comments about the return value.

include/glow/Optimizer/Optimizer.h
```diff
@@ -51,7 +52,13 @@ void lower(Function *F, CompilationContext &cctx, const Backend *B = nullptr,
...
+
+public:
+  const std::string getName() const { return "DCE"; }
```
rdzhabarov: nit: why not returning string ref or const std::string&. We probably do not want copying string around too much.

include/glow/Optimizer/Optimizer.h
rdzhabarov: nit: bool passImpl(...) override;

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,52 @@
...
+    doInitialization(F);
+    bool changed = passImpl(F);
+    doFinalization(F);
+    return changed;
+  };
+
+  /// \returns the name of the pass.
+  virtual const std::string getName() const = 0;
```
rdzhabarov: consider const std::string& to avoid copies.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+#ifndef GLOW_OPTIMIZER_FUNCTIONPASS_H
+#define GLOW_OPTIMIZER_FUNCTIONPASS_H
+
+namespace glow {
+
+class FunctionPass {
+protected:
+  /// Perform some tasks before the pass.
+  virtual void doInitialization(Function *F) const {
```
rdzhabarov: maybe, preProcessing, and postProcessing?

rdzhabarov:
How do you envision this to be be used to control optimizations on per backend basis?

jfix71:
RE: `FunctionPass`, `doInitialization()`, `doFinalization()`, I was [following LLVM here](https://llvm.org/docs/WritingAnLLVMPass.html#the-functionpass-class). Happy to go in a different direction if desired.

For FunctionPass though I kind of agree with LLVM here. The pass iterates over a Function. If we call it `OptimizationPass` then do we also have an `OptimizationPass` over Modules? Does that mean we have `FunctionOptimizationPass` vs. `ModuleOptimizationPass`? Or do we add `OptimizationPass::runOnModule()` which is undefined for a pass that is supposed to operate on a `Function`? Just seems cleaner for it to be `FunctionPass` to me.

include/glow/Optimizer/Optimizer.h
jfix71: This is used by the pass to signify if it changed the Function. Some of our passes do this, so I've set it up so all will.

beicy:
> RE: `FunctionPass`, `doInitialization()`, `doFinalization()`, I was [following LLVM here](https://llvm.org/docs/WritingAnLLVMPass.html#the-functionpass-class). Happy to go in a different direction if desired.
> 
> For FunctionPass though I kind of agree with LLVM here. The pass iterates over a Function. If we call it `OptimizationPass` then do we also have an `OptimizationPass` over Modules? Does that mean we have `FunctionOptimizationPass` vs. `ModuleOptimizationPass`? Or do we add `OptimizationPass::runOnModule()` which is undefined for a pass that is supposed to operate on a `Function`? Just seems cleaner for it to be `FunctionPass` to me.

Thanks for explaining it.  I just would like to add something in the name to emphasize "optimization" .  Anyway we don't have "ModulePass" so far.  But it is OK for me to keep the name.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,53 @@
...
+  /// Implementation of the pass. \returns whether the pass modifies \p F.
+  virtual bool passImpl(Function *F) = 0;
+
+public:
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+
+  // Run the pass, including initialization and finalization. \returns whether
+  // the pass modifies \p F.
+  virtual bool runOnFunction(Function *F) {
```
opti-mix: Tiny nit: Do we expect `runOnSomethingElse`? If not, maybe it should be called just `run` as it is obvious from the type of the single argument that it runs on a function.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,53 @@
...
+    doInitialization(F);
```
opti-mix: Do you expect each pass to perform pass-specific `doInitialization` and `doFinalization`? If the logic will be rather uniform for all passes, it should better be handled in the pass manager and then these two functions should not be a part of the API.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,53 @@
...
+  /// Perform some tasks before the pass.
+  virtual void doInitialization(Function *F) const {
+    LOG_SCOPE(F->getLogContext(), getName());
+  }
+
+  /// Perform some tasks after the pass.
+  virtual void doFinalization(Function *F) const {}
+
+  /// Implementation of the pass. \returns whether the pass modifies \p F.
+  virtual bool passImpl(Function *F) = 0;
```
opti-mix: Does it need to be the part of the API? `runFunction` should be enough for the clients. And a pass can decide for itself if it wants to put all of its logic into `runFunction` or if it needs to use a helper.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_OPTIMIZER_FUNCTIONPASS_H
+#define GLOW_OPTIMIZER_FUNCTIONPASS_H
+
+namespace glow {
+
+class FunctionPass {
```
opti-mix: Some designs of pass managers do not take `Function` as an explicit constructor argument of their passes. Instead they are injected by the pass manager via some APIs exposed by the pass. See e.g. https://github.com/apple/swift/blob/master/include/swift/SILOptimizer/PassManager/Transforms.h

I don't have a strong opinion about this.

opti-mix: Another interesting bit in the example above is the way how passes are mass-defined by means of a def file and macro meta-programming:
https://github.com/apple/swift/blob/master/include/swift/SILOptimizer/PassManager/Passes.def

include/glow/Optimizer/FunctionPass.h
jfix71: I was thinking that there would be uniform logic for all passes handled in the pass manager. The flow when we run a pass would be:

```cpp
PassManager::prePass();
FunctionPass::runOnFunction(); // this calls:
//  FunctionPass::doInitialization();
//  FunctionPass::passImpl();
//  FunctionPass::doFinalization();
PassManager::postPass();
```

For now I put `LOG_SCOPE()` inside `doInitialization()` because we don't have the pass manager yet, but it would be moved into `PassManager::prePass()` eventually.

include/glow/Optimizer/FunctionPass.h
jfix71: Makes sense, I guess the pass could be responsible for implementing any initialization/finalization logic in `runFunction()`. I can revert moving of `LOG_SCOPE()` and wait until the pass manager lands to put it there.

include/glow/Optimizer/FunctionPass.h
opti-mix: Based on this, it seems like you'd be able to move virtually all of the logic from doInitialization and doFinalization into the pass manager (it will call LOG_SCOPE, etc for all passes it runs), which means that these APIs are probably not strictly necessary. 

I'm not against having doInitialization and doFinalization. I just think that they will be NOPs most of the time.

jfix71: Yeah agreed, I will remove `doInitialization` and `doFinalization` and let the pass implement its logic however it wants.

jackm321:
Nice!

include/glow/Optimizer/Optimizer.h
```diff
@@ -51,7 +52,14 @@ void lower(Function *F, CompilationContext &cctx, const Backend *B = nullptr,
...
 
 /// Dead code elimination.
-void DCE(Function *F);
+class DCE : public FunctionPass {
+private:
+  /// Implementation of DCE. \returns whether the pass modifies \p F.
+  bool passImpl(Function *F) override;
+
+public:
+  llvm::StringRef getName() const override { return "DCE"; }
```
rdzhabarov: do we have any test which runs through the getName(), I suspect "DCE" might go out of scope and gets destroyed while StringRef still points to it. Maybe have static member passname inside the DCE class and return it here.

rdzhabarov: Thanks for double checking. "String literals have static storage duration, and thus exist in memory for the life of the program", so we should be safe here.

include/glow/Optimizer/FunctionPass.h
jfix71: I'm not sure I see the benefit of injection for now. I do like the macro meta-programming, I think it will be useful into the future when integrating into the pass manager and allowing backends to opt in/out of passes, or specify their own pass pipeline.

include/glow/Optimizer/FunctionPasses.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace glow {
+
+#define FUN_PASS(PASS_NAME)                                                    \
+  struct PASS_NAME : public FunctionPass {                                     \
+    bool run(Function *F) override;                                            \
+    llvm::StringRef getName() const override { return #PASS_NAME; }            \
+  };
+
+#include "FunctionPasses.def"
```
opti-mix: It is safer to add `#undef FUN_PASSS` afterwards.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,39 @@
...
+#define GLOW_OPTIMIZER_FUNCTIONPASS_H
+
+namespace glow {
+
+class Function;
+
+/// Class used for all passes over Functions. All passes over Functions should
+/// derive from this class, implementing the pass logic and additionally can add
+/// logic for running before and after the pass runs.
+struct FunctionPass {
```
opti-mix: Any reasons why it is a struct and not a class?

jfix71: Both functions are public, that's all 🙂 can make class if preferred.

include/glow/Optimizer/FunctionPasses.h
jfix71: I added it to the end of `FunctionPasses.def`, thought it was best there so it was never forgotten.

opti-mix: Oops. Never mind. I haven't noticed it there ;-)

include/glow/Optimizer/FunctionPass.h
opti-mix: Let's make them classes as we may add some private members over time (e.g. reference to a pass manager or a function) and expose them via some setter/getter APIs.

include/glow/Optimizer/FunctionPasses.h
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+#ifndef GLOW_OPTIMIZER_FUNCTIONPASSES_H
+#define GLOW_OPTIMIZER_FUNCTIONPASSES_H
+
+#include "glow/Optimizer/FunctionPass.h"
+
+namespace glow {
+
+#define FUN_PASS(PASS_NAME)                                                    \
```
opti-mix: BTW, do we want to define IR_FUN_PASS as well, which would be a pass over IRFunctions. This one could be useful for the IROptimizer. Doesn't need to happen in this PR though.

SplitInfinity:
cc: @ZchiPitt in case there is any significant impact on your work

ZchiPitt:
Looking good!

jfix71:
I've updated the rest of the optimizations. This included adding in `bool changed` to those optimizations missing it.

opti-mix:
@jfix71 It almost looks like introducing a small pass manager in this PR would save you some re-factoring in the future ;-) If you do it in subsequent PRs, you'll need to move those SCOPEs into the pass manager and you'd replace `glow::optimize` with a pipeline executed by the pass manager.

include/glow/Optimizer/FunctionPass.h
```diff
@@ -0,0 +1,40 @@
...
+
+/// Class used for all passes over Functions. All passes over Functions should
+/// derive from this class, implementing the pass logic and additionally can add
+/// logic for running before and after the pass runs.
+class FunctionPass {
+public:
+  FunctionPass() = default;
+  virtual ~FunctionPass() = default;
+
+  // Run the pass on \p F. \returns whether the pass modifies \p F.
```
rdzhabarov: ///

jfix71:
@opti-mix Going to just land as is, I don't want to make the PR too big and don't think it'll be too bad to refactor 🙂

PR #3148, opened 2019-06-20
Rename BackendId to Backend

bertmaher:
Summary:
Since the ID is just a fiction for onnxifi's benefit, IMO it's clearer
to just call this class `Backend`.

Reviewed By: rdzhabarov

Differential Revision: D15903690

### DIFF
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -77,9 +77,9 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
       return ONNXIFI_STATUS_FALLBACK;
     }
 
-    auto *quantizationBackendOnnx = manager.createBackendId(
+    auto *quantizationBackendOnnx = manager.createBackend(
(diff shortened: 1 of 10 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -77,9 +77,9 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
...
     }
 
-    auto *quantizationBackendOnnx = manager.createBackendId(
+    auto *quantizationBackendOnnx = manager.createBackend(
         "Interpreter", /*useOnnx*/ true, /*forQuantization*/ true);
-    auto *quantizationBackendC2 = manager.createBackendId(
+    auto *quantizationBackendC2 = manager.createBackend(
         "Interpreter", /*useOnnx*/ false, /*forQuantization*/ true);
 
     backendIDs[0] = quantizationBackendOnnx;
```
beicy: Should this one also be replaced with "backends" ?

bertmaher:
I thought about renaming the `backendIDs` in onnxifiGlow, but I decided to leave them alone since they're sort of the "onnxifi" side of things where backend-id is an actual concept (even if we're not really differentiating internally)

PR #3150, opened 2019-06-20
[clean up] Make a function to be const member function in partitioner if possible.

beicy:
Summary:
This PR is created from the following discussion.
https://github.com/pytorch/glow/pull/3142#discussion_r295558843

Basically,  in Partitioner, some functions were not const member function (although it won't modify any value in this class), and they can't be used in const class objects.  This PR marks those functions as const. 

Documentation:

[Optional Fixes #issue]

Test Plan:


beicy:
Thanks @rdzhabarov for pointing out this issue.

PR #3151, opened 2019-06-20
[BundleSaver] Separate bundle file name from main entry name.

artemrakhov:
I deleted `save()` from ExecutionEngine because it's being deprecated anyway. It was not used anywhere besides 1 unit test.


include/glow/LLVMIRCodeGen/LLVMIRGen.h
```diff
@@ -282,6 +284,10 @@ class LLVMIRGen {
   virtual void performSpecialization();
   /// \returns allocations info.
   virtual AllocationsInfo &getAllocationsInfo() { return allocationsInfo_; }
+  /// \returns the name of the bundle, to be used for filename when saving.
+  std::string getBundleName() const;
```
rdzhabarov: const std::string& or llvm::StringRef to avoid unnecessary copies.

PR #3152, opened 2019-06-20
[NFC][Partitioner] Code refactoring

beicy:
Summary:
This is a follow up code refactoring after #3062.  Basically, we use updateGraphMemInfoByAddingNode to update the memory usage of a node set to eliminate some duplicated calculation. 

Documentation:

[Optional Fixes #issue]

Test Plan:
ninja test

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -423,19 +423,21 @@ void Partitioner::partitionsAdjust(NodeToFunctionMap &partitions,
     gain = false;
     for (FunctionToNodesMapTy::iterator it = nodesSet.begin();
          it != nodesSet.end(); ++it) {
-      NodesSetTy nSet = (*it).second;
-      std::vector<Node *> outUsers = getOutUsersWithOnePredecessor(nSet);
+      NodesSetTy &curSet = (*it).second;
+      std::vector<Node *> outUsers = getOutUsersWithOnePredecessor(curSet);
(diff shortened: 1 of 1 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -423,19 +423,21 @@ void Partitioner::partitionsAdjust(NodeToFunctionMap &partitions,
     gain = false;
     for (FunctionToNodesMapTy::iterator it = nodesSet.begin();
          it != nodesSet.end(); ++it) {
-      NodesSetTy nSet = (*it).second;
-      std::vector<Node *> outUsers = getOutUsersWithOnePredecessor(nSet);
+      NodesSetTy *curSet = &((*it).second);
```
opti-mix: Should it rather be `NodesSetTy &curSet`? Or do you plan to change the value of curSet pointer?

beicy: Yes, the value maybe changed by inserting new node into this set.

opti-mix: I'm not asking about the content of `NodesSetTy`. I'm saying if you assign to the NodesSetTy pointer only once, like in this line, you can replace it with a reference. But if you need to assign to the pointer multiple times, you really need a pointer.

beicy: You are right!  No need to use the pointer now. Thanks!

PR #3154, opened 2019-06-20
[compilationLog]: Log the deletion of constants and placeholders.

ZchiPitt:
Summary:

Log the deletion of constants and placeholders

Test Plan:
ninja check

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -36,6 +36,20 @@ using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
+namespace {
+/// A helper function to log the deletion of constant/placeholder \p s of a
+/// module.
+void loggingStorageDeletionInUserFunc(Storage *s) {
+  std::unordered_set<Function *> userFuncs;
+  for (auto &use : s->getUsers()) {
+    userFuncs.insert(use.getUser()->getParent());
+  }
+  for (auto *F : userFuncs) {
+    F->getLogContext().logNodeDeletion(*s);
+  }
+}
+} // namespace
+
 bool Module::hasFunction(llvm::StringRef name) { return getFunction(name); }
 
 Function *Module::getFunction(llvm::StringRef name) {
@@ -62,10 +76,9 @@ void Module::strip() {
 }
 
 void Module::clear() {
-  eraseFunctions();
-
   for (auto it = constants_.begin(), e = constants_.end(); it != e; it++) {
     Constant *v = *it;
+    loggingStorageDeletionInUserFunc(v);
     delete v;
   }
 
@@ -74,8 +87,11 @@ void Module::clear() {
   for (auto it = placeholders_.begin(), e = placeholders_.end(); it != e;
        it++) {
     Placeholder *p = *it;
+    loggingStorageDeletionInUserFunc(p);
     delete p;
   }
+
+  eraseFunctions();
 
   placeholders_.clear();
 }
@@ -2698,6 +2714,7 @@ Node *Function::getNodeByName(llvm::StringRef name) {
 void Module::eraseConstant(ConstList::iterator I) {
   if (I == constants_.end())
     return;
+  loggingStorageDeletionInUserFunc(*I);
   delete *I;
   constants_.erase(I);
 }


lib/Graph/Graph.cpp
```diff
@@ -62,10 +62,13 @@ void Module::strip() {
 }
 
 void Module::clear() {
-  eraseFunctions();
-
   for (auto it = constants_.begin(), e = constants_.end(); it != e; it++) {
     Constant *v = *it;
+
+    for (auto F : functions_) {
+      F->getLogContext().logNodeDeletion(*v);
```
jfix71: What if `F` doesn't use `v`? This seems incorrect to me since we would have logged only deleting it but not adding it.

It feels to me like Constant modifications should be logged at the Module level. Even if we modify or add a Constant during a Function optimization, it should probably be logged somehow that the Module's Constant was modified/added, but noting that it was during optimizing `F`.

ZchiPitt: You are right, it will create confusion.  Maybe it's better to specify in the log that the constant is deleted from module instead of certain function.

I'm thinking of adding another logging method 'logConstantOrPlaceholderDeletion'. In this method, the log will be produced as "DELETE Constant XXX from Module XYZ", so that later when we processing the log the parser knows that the deleted constant may be used in another function.

ZchiPitt:
After discussing @jfix71 , i fixed the pr such that the constant deletion is only logged into the constant's user's parent function (instead of into all functions of the module). 
The logic right now is more correct and we avoid the risk of logging deleting constants in a function that does not even use it.

lib/Graph/Graph.cpp
```diff
@@ -62,10 +62,22 @@ void Module::strip() {
...
 void Module::clear() {
-  eraseFunctions();
-
   for (auto it = constants_.begin(), e = constants_.end(); it != e; it++) {
     Constant *v = *it;
+
+    // Log constant deletion in all its user functions.
+    FunctionList userFuncs;
+    for (auto use : v->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
```
SplitInfinity: Nit: Consider using `auto *` here to make it clearer that `userFunc` is a pointer and that you're not maintaining and searching through a list of entire objects.

lib/Graph/Graph.cpp
```diff
@@ -74,9 +86,26 @@ void Module::clear() {
   for (auto it = placeholders_.begin(), e = placeholders_.end(); it != e;
        it++) {
     Placeholder *p = *it;
+
+    // Log placeholder deletion in all its user functions.
+    FunctionList userFuncs;
+    for (auto use : p->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
```
SplitInfinity: Nit: Consider `auto *` here as well.

lib/Graph/Graph.cpp
```diff
@@ -62,10 +62,22 @@ void Module::strip() {
...
+    // Log constant deletion in all its user functions.
+    FunctionList userFuncs;
+    for (auto use : v->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
+      if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
+          userFuncs.end()) {
+        userFuncs.push_back(userFunc);
+      }
+    }
+    for (auto F : userFuncs) {
```
SplitInfinity: Nit: Consider `auto *` here as well.

lib/Graph/Graph.cpp
```diff
@@ -74,9 +86,26 @@ void Module::clear() {
...
+    FunctionList userFuncs;
+    for (auto use : p->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
+      if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
+          userFuncs.end()) {
+        userFuncs.push_back(userFunc);
+      }
+    }
+
+    for (auto F : userFuncs) {
```
SplitInfinity: Nit: Consider `auto *` here as well.

lib/Graph/Graph.cpp
```diff
@@ -2698,6 +2727,21 @@ Node *Function::getNodeByName(llvm::StringRef name) {
...
+  FunctionList userFuncs;
+  for (auto use : (*I)->getUsers()) {
+    auto userFunc = use.getUser()->getParent();
+    if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
+        userFuncs.end()) {
+      userFuncs.push_back(userFunc);
+    }
+  }
+
+  for (auto F : functions_) {
```
SplitInfinity: Nit: Consider `auto *` here as well.

lib/Graph/Graph.cpp
```diff
@@ -2698,6 +2727,21 @@ Node *Function::getNodeByName(llvm::StringRef name) {
 void Module::eraseConstant(ConstList::iterator I) {
   if (I == constants_.end())
     return;
+
+  // Log constant deletion in all its user functions.
+  FunctionList userFuncs;
+  for (auto use : (*I)->getUsers()) {
+    auto userFunc = use.getUser()->getParent();
```
SplitInfinity: Nit: Consider `auto *` here as well.

lib/Graph/Graph.cpp
```diff
@@ -62,10 +62,22 @@ void Module::strip() {
...
-  eraseFunctions();
-
   for (auto it = constants_.begin(), e = constants_.end(); it != e; it++) {
     Constant *v = *it;
+
+    // Log constant deletion in all its user functions.
+    FunctionList userFuncs;
+    for (auto use : v->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
+      if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
```
SplitInfinity: Consider performing the logging of node deletion in this loop itself. To avoid double deletion, you can keep a `std::unordered_set` or equivalent of `Function*` and check if a function is already there before logging. I think that will be faster than `std::find` in a list.

lib/Graph/Graph.cpp
```diff
@@ -74,9 +86,26 @@ void Module::clear() {
   for (auto it = placeholders_.begin(), e = placeholders_.end(); it != e;
        it++) {
     Placeholder *p = *it;
+
+    // Log placeholder deletion in all its user functions.
+    FunctionList userFuncs;
+    for (auto use : p->getUsers()) {
+      auto userFunc = use.getUser()->getParent();
+      if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
```
SplitInfinity: I think you can use a set here too.

lib/Graph/Graph.cpp
```diff
@@ -2698,6 +2727,21 @@ Node *Function::getNodeByName(llvm::StringRef name) {
 void Module::eraseConstant(ConstList::iterator I) {
   if (I == constants_.end())
     return;
+
+  // Log constant deletion in all its user functions.
+  FunctionList userFuncs;
+  for (auto use : (*I)->getUsers()) {
+    auto userFunc = use.getUser()->getParent();
+    if (std::find(userFuncs.begin(), userFuncs.end(), userFunc) ==
```
SplitInfinity: I think you can use a set here too.

SplitInfinity:
I see a bunch of `for (auto x : ...)` in this PR. In this case it's not a huge deal because `Use` contains a bunch of pointers and an integer but it's a good idea to get into the habit of using `for (const auto& : ...)` when you want to iterate through a container without modifying any of the contents.

ZchiPitt:
@jfix71 thx for the suggestion. Added a helper function now.
@SplitInfinity Thx for advice on the coding habit! Changed to auto* now. Cant add const for Function* as the function I call is not const yet. (Probably need to change a lot of places in order to make them all const. ) But thx for the advice, i'll keep that in mind in all future designs.

lib/Graph/Graph.cpp
```diff
@@ -61,11 +61,25 @@ void Module::strip() {
   }
 }
 
-void Module::clear() {
-  eraseFunctions();
+// A helper function to log the deletion of constants or placeholders of a
+// module.
+void loggingDeletionInUserFunc(const Node *n) {
+  std::unordered_set<Function *> userFuncs;
+  for (auto use : n->getUsers()) {
```
SplitInfinity: Consider making this `const auto&` or at least `auto&`.

lib/Graph/Graph.cpp
```diff
@@ -61,11 +61,25 @@ void Module::strip() {
   }
 }
 
-void Module::clear() {
-  eraseFunctions();
+// A helper function to log the deletion of constants or placeholders of a
```
jfix71: nit: `///`. Also reference `\p n`

lib/Graph/Graph.cpp
```diff
@@ -61,11 +61,25 @@ void Module::strip() {
...
+// module.
+void loggingDeletionInUserFunc(const Node *n) {
```
jfix71: nit: Make this `static`

jfix71: This is intended for Storage Nodes only, so would make more sense to pass `Storage *` instead of `Node *`

SplitInfinity: Regarding making this `static`: I think the more canonical way of doing this in C++ is to put it into an anonymous namespace, i.e.:

```
namespace {
void myFunc(...) { ... }
}
```

jfix71: Yeah, static is more C-like. We do a bit of both spread out across Glow, e.g. in `GraphOptimizer.cpp` it's all mostly static. I'm fine either way.

ZchiPitt: put it in a namespace now

lib/Graph/Graph.cpp
```diff
@@ -36,6 +36,23 @@ using llvm::cast;
...
 using llvm::isa;
 
+namespace {
+/// A helper function to log the deletion of constant/placeholder \p s of a
+/// module.
+void loggingStorageDeletionInUserFunc(Storage *s) {
+  std::unordered_set<Function *> userFuncs;
+  for (auto &use : s->getUsers()) {
+    auto *userFunc = use.getUser()->getParent();
+    if (userFuncs.find(userFunc) == userFuncs.end()) {
```
jfix71: nit: Don't need to search for `userFunc` before inserting into the set, since set will naturally handle multiple insertions of a duplicate `userFunc`.

PR #3155, opened 2019-06-21
Support int32 and Boolean types for Concat in Interpreter

Max-K-cadence:
Summary:
This patch adds support for Concat and related nodes for int32 and Boolean types
in interpreter. The main purpose of this is to leverage constant-folding to get rid of
such tensors, that's why no CPU support added so far (will be done if needed).

Test Plan:
Added unit tests

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -1120,6 +1120,10 @@ void BoundInterpreterFunction::fwdSplatInst(const glow::SplatInst *I) {
     return T->getHandle<int64_t>().clear(I->getValue());
   }
 
+  if (k == ElemKind::Int32ITy) {
+    return T->getHandle<int32_t>().clear(I->getValue());
+  }
+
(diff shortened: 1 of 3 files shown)


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1136,6 +1140,10 @@ void BoundInterpreterFunction::fwdSplatInst(const glow::SplatInst *I) {
     return T->getHandle<int8_t>().clear(quantization::quantize(val, destQ));
   }
 
+  if (k == ElemKind::BoolTy) {
+    return T->getHandle<bool>().clear(I->getValue());
```
jfix71: @Max-K-cadence Our internal linter isn't liking the implicit float -> bool cast here. Can you e.g. add an explicit `static_cast<bool>(I->getValue())`, or something similar to avoid the implicit cast?

Max-K-cadence: Done!

Issue #1469, opened 2018-08-16
Graph optimizer should perform constant folding


opti-mix:
If e.g. both operands of an arithmetic operation like `add` or `sub` are splats (i.e. constants) or more generally tensors with payloads statically known at compile-time , the result of the operation can be computed at compile-time. The process of evaluation the constant expressions at compile time is known as [`constant folding`](https://en.wikipedia.org/wiki/Constant_folding).

Such constant computations may arise e.g. after applying some other optimizations or transformations.

Currently, Glow's graph optimizer does not support constant folding. But it should be fairly easy to extend the graph optimizer to support constant folding. Most likely it should be performed during the execution of `simplifyNode` in GraphOptimizer.cpp

jfix71:
Related -- once we have constant folding implemented, it would be great to extend the optimizer to recognize commutative patterns like this (seen in en2gr) which could also use constant folding:

<img width="424" alt="screen shot 2018-11-13 at 12 04 10 pm" src="https://user-images.githubusercontent.com/1198212/48439884-48309e80-e73c-11e8-9bb9-a174b26b5a19.png">

jackm321:
This looks interesting to work on. I'm curious if anybody has a sense for what the prioritization is for this? Is it more of a wishlist item for now or would it be worth while to look into doing it?

jfix71:
>Is it more of a wishlist item for now or would it be worth while to look into doing it?

@jackm321 I'd lean a bit more towards wishlist for now. I think it will be important for models which are exported via ONNX, as the exporter is sometimes somewhat naive and outputs patterns like the one I posted above from en2gr. We do care about seq2seq networks like en2gr, and it would be helpful for other models exported from ONNX, but as far as I've seen it isn't prevalent  in other networks generally and so IMO isn't a high priority for now.

Also, I've seen related issues (as described here https://github.com/pytorch/glow/issues/1580#issuecomment-418606504) where the ONNX exporter decides to export multiple nodes to create a shape vector as an input tensor to a node instead of statically passing the shape in as member arguments (e.g. the shape for a Reshape was constructed via multiple nodes and passed in as a tensor). So it would be nice to somehow integrate this constant folding at graph building time too, allowing us to load such graphs.

nickgg:
@jfix71 how much is left to do of this?

jfix71:
[...]

PR #3161, opened 2019-06-22
[graph-optimizer] Implement general constant folding

opti-mix:
Any non-trivial node (i.e. not a constant or a splat) that can be computed at compile-time is going to be computed at compile-time! This is implemented by detecting nodes with constant inputs and evaluating their results at compile-time by means of using a temporary interpreter-based execution engine.

Related to #1469 and #3141.

### DIFF
--- a/include/glow/Optimizer/GraphOptimizer/CompilationContext.h
+++ b/include/glow/Optimizer/GraphOptimizer/CompilationContext.h
@@ -49,6 +49,12 @@ struct PrecisionConfiguration {
 
 using QuantizationMode = PrecisionConfiguration::QuantizationMode;
 
+/// Options relevant to optimizations during compilation.
+struct OptimizationOptions {
+  /// If true, perform compile-time computation of constant operations.
+  bool enableConstantFolding{true};
+};
+
 /// Context for compilation.
 struct CompilationContext {
   /// Used during Profiling.
@@ -66,6 +72,9 @@ struct CompilationContext {
 
   /// Options for the Backend to use.
   BackendOptions backendOpts;
+
+  /// Options for the optimizations to use.
+  OptimizationOptions optimizationOpts;
 
   /// Configuration for different precision modes.
   PrecisionConfiguration precisionConfig;
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -3058,6 +3058,26 @@ SaveNode *glow::getOutputSave(Function *F, Placeholder *PH) {
   return nullptr;
 }
 
+Node *glow::recursiveClone(Function *newF, Node *node, NodeMap &currToNew) {
+  Node *copy = node->clone();
+  currToNew[node] = copy;
+  newF->addNode(copy);
+  for (unsigned inp = 0, e = copy->getNumInputs(); inp < e; inp++) {
+    auto input = copy->getNthInput(inp);
+    auto it = currToNew.find(input.getNode());
+    Node *newInput;
+    if (it != currToNew.end()) {
+      newInput = it->second;
+    } else if (llvm::isa<Storage>(input.getNode())) {
+      continue;
+    } else {
+      newInput = recursiveClone(newF, input.getNode(), currToNew);
+    }
+    copy->setNthInput(inp, NodeValue(newInput, input.getResNo()));
+  }
+  return copy;
+}
+
 namespace glow {
 
 llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Module &mod) {
--- a/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
+++ b/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
@@ -1,0 +1,267 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
+
+#include "glow/Backend/Backend.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/Log.h"
+#include "glow/Graph/Node.h"
+#include "glow/Graph/Nodes.h"
+#include "glow/Graph/PlaceholderBindings.h"
+#include "glow/Graph/Utils.h"
+#include "glow/Optimizer/GraphOptimizer/FunctionPasses.h"
+
+#include "llvm/Support/Casting.h"
+
+using namespace glow;
+using llvm::dyn_cast;
+using llvm::isa;
+
+namespace {
+/// The name of the temporary function to be used to perform constant folding.
+constexpr const char *constEvaluationFunctionName =
+    "__constEvaluationFunction__";
+
+/// \returns true if a node \p N is a constant operation, i.e. it is a trivial
+/// constant like Constant or Splat or all of its inputs are recursively
+/// constant operations, and it has no side-effects and supported by the \p
+/// backend.
+bool isConstantOperation(const Node *N, const Backend &backend) {
+  // An operation with side-effects cannot be computed at compile-time.
+  if (N->hasSideEffects()) {
+    return false;
+  }
+  // Constant and splat nodes are trivially constant operations.
+  if (isa<Constant>(N) || isa<SplatNode>(N)) {
+    return true;
+  }
+  // If the operation is not supported by the backend, it cannot be computed at
+  // compile-time.
+  if (!backend.shouldLower(N) && !backend.isOpSupported(NodeInfo(*N))) {
+    return false;
+  }
+  if (isa<Placeholder>(N)) {
+    return false;
+  }
+  for (size_t idx = 0, e = N->getNumInputs(); idx < e; ++idx) {
+    auto input = N->getNthInput(idx);
+    if (!isConstantOperation(input.getNode(), backend)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/// \returns true if node \p N has at least one non-constant operation user.
+bool hasNonConstantOperationUser(const Node *N, const Backend &backend) {
+  assert(isConstantOperation(N, backend) && "Expected constant operation");
+  for (auto &use : N->getUsers()) {
+    auto *user = use.getUser();
+    // Only consider users in the current function.
+    if (user->getParent() != N->getParent()) {
+      continue;
+    }
+    if (!isConstantOperation(user, backend)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/// Compile the function \p F for the provided \p backend using the compilation
+/// context \p cctx.
+/// \returns compiled function.
+std::unique_ptr<CompiledFunction> compile(Backend &backend, Function &F,
+                                          CompilationContext &cctx) {
+  EXIT_ON_ERR(::glow::optimizeFunction(&F, backend, cctx));
+  for (const Node &N : F.getNodes()) {
+    CHECK(backend.isOpSupported(N))
+        << "Backend must support all nodes after high-level optimizations but "
+           "encountered unsupported operator: "
+        << N.getDebugDesc();
+  }
+  auto funcOrErr = backend.compile(&F, cctx.backendOpts);
+  EXIT_ON_ERR(funcOrErr.takeError());
+  return std::move(*funcOrErr);
+}
+
+/// Runs the compiled function \p compiledF on the \p backend using provided \p
+/// bindings.
+void run(Backend &backend, CompiledFunction &compiledF,
+         PlaceholderBindings &bindings) {
+  std::unique_ptr<PlaceholderBindings> bindingsPtr(&bindings);
+  ExecutionContext context(std::move(bindingsPtr));
+  // TODO: Add only constants used by F to the compiled function. This should
+  // reduce the amount of data that needs to be copied.
+  auto executeErr = compiledF.execute(&context);
+  EXIT_ON_ERR(std::move(executeErr));
+  // Don't delete bindings.
+  context.movePlaceholderBindings().release();
+}
+
+/// Evaluates a provided constant operation \p C using the provided \p backend
+/// and using the compilation context \p cctx.
+/// \returns constant results.
+std::vector<Constant *>
+evaluateConstantOperation(Backend &backend, CompilationContext &cctx, Node *C) {
+  PlaceholderBindings bindings;
+  assert(isConstantOperation(C, backend) && "Expected a constant expression");
+  // Constants and splats do not need to be constant evaluated.
+  if (isa<Constant>(C) || isa<SplatNode>(C)) {
+    return {};
+  }
+  Module &mod = *C->getParent()->getParent();
+  // Create a temporary function to perform the constant operation.
+  Function *constEvaluationF = mod.createFunction(constEvaluationFunctionName);
+  // Mapping from existing nodes to the new ones.
+  NodeMap currToNew;
+  // Clone the constant operation and some of its inputs if necessary.
+  auto *clonedC = recursiveClone(constEvaluationF, C, currToNew);
+  // Create save nodes for each of the results.
+  llvm::SmallVector<SaveNode *, 16> savedResults;
+  for (size_t idx = 0, e = clonedC->getNumResults(); idx < e; ++idx) {
+    auto *SN = constEvaluationF->createSave(clonedC->getName(),
+                                            clonedC->getNthResult(idx));
+    savedResults.emplace_back(SN);
+    bindings.allocate(SN->getPlaceholder());
+  }
+  // Run the temporary backend to perform this constant operation
+  // evaluation.
+  EXIT_ON_ERR(
+      executeConstantFunction(backend, *constEvaluationF, bindings, cctx));
+  // Get the results of the constant operation compile-time computation and
+  // create new constants from it.
+  std::vector<Constant *> constResults;
+  constResults.reserve(savedResults.size());
+  for (auto *SN : savedResults) {
+    Tensor *outputTensor = bindings.get(SN->getPlaceholder());
+    auto *constResult =
+        mod.createConstant(SN->getName(), std::move(*outputTensor));
+    constResults.emplace_back(constResult);
+  }
+  // Remove the temporary function.
+  mod.eraseFunction(constEvaluationF);
+  return constResults;
+}
+
+/// Check if function \p F consists of constant operations only.
+LLVM_ATTRIBUTE_USED
+llvm::Error verifyConstantFunction(Backend &backend, Function &F) {
+  // Perform the checks in DEBUG builds only.
+  for (auto &N : F.getNodes()) {
+    // Saving results is fine.
+    if (isa<SaveNode>(&N)) {
+      continue;
+    }
+    // Placeholders can be used just to save results.
+    if (!isa<Placeholder>(&N)) {
+      RETURN_ERR_IF_NOT(isConstantOperation(&N, backend),
+                        "Expected constant operation");
+      continue;
+    }
+    if (!N.hasOneUse()) {
+      RETURN_ERR("Expected constant operation");
+    }
+    auto *SN = dyn_cast<SaveNode>(N.getUsers().begin()->getUser());
+    if (SN && SN->getPlaceholder() == &N) {
+      continue;
+    }
+    RETURN_ERR("Expected constant operation");
+  }
+  return llvm::Error::success();
+}
+
+/// Perform a compile-time constant folding of the node \p N using the provided
+/// \p backend.
+/// \returns list of constants which are the result of the
+/// constant-folding. These constants correspond to results of the node. If no
+/// constant folding was possible an empty vector will be returned
+std::vector<Constant *> constantFoldNodeImpl(Backend &backend, Node *N) {
+  CompilationContext cctx;
+  // Do not recursively call constant folding.
+  cctx.optimizationOpts.enableConstantFolding = false;
+  cctx.backendOpts.collectConstants = true;
+  return evaluateConstantOperation(backend, cctx, N);
+}
+
+} // namespace
+
+llvm::Error glow::executeConstantFunction(Backend &backend, Function &F,
+                                          PlaceholderBindings &bindings,
+                                          CompilationContext &cctx) {
+// Perform the checks in DEBUG builds only.
+#ifndef NDEBUG
+  RETURN_IF_ERR(verifyConstantFunction(backend, F));
+#endif
+  auto compiledF = compile(backend, F, cctx);
+  run(backend, *compiledF, bindings);
+  return llvm::Error::success();
+}
+
+/// Perform constant folding in the function \p F . Any non-trivial node (i.e.
+/// not a constant or a splat) that can be computed at compile-time is going to
+/// be computed at compile-time. \returns true if any foldings were performed.
+bool glow::ConstantFold::run(Function *F) {
+  LOG_SCOPE(F->getLogContext(), "glow::constantFold")
+  bool changed = false;
+  // Backend to be used for compile-time computations.
+  std::unique_ptr<Backend> backend(createBackend("Interpreter"));
+  // Traverse nodes in post-order, so that children are seen before parents.
+  GraphPostOrderVisitor postOrderVisitor(*F);
+  auto nodes = postOrderVisitor.getPostOrder();
+  // Collect all non-trivial constant operations.
+  for (auto *N : nodes) {
+    // Skip trivial nodes/operations that do not require any constant
+    // computations.
+    if (isa<Storage>(N) || isa<Constant>(N) || isa<SplatNode>(N)) {
+      continue;
+    }
+    // Skip nodes that are not constant operations.
+    if (!isConstantOperation(N, *backend)) {
+      continue;
+    }
+    // Add only a constant operation node whose value is used by at least
+    // one non constant-operation node, because no other bigger constant
+    // operation containing the current node can completely replace the result
+    // of its computation. Doing this check allows for performing a smaller
+    // number of evaluateConstantOperation calls later and thus reduces the
+    // overhead.
+    if (!hasNonConstantOperationUser(N, *backend)) {
+      continue;
+    }
+    // Compute the constant value of the node.
+    std::vector<Constant *> constResults = constantFoldNodeImpl(*backend, N);
+    // Replace all results of the original operation by the computed
+    // compile-time results of this operation.
+    for (size_t idx = 0, e = constResults.size(); idx < e; ++idx) {
+      auto constResult = constResults[idx];
+      // Replace the old result by the new constant result.
+      N->getNthResult(idx).replaceAllUsesOfWith(constResult);
+    }
+    changed = true;
+  }
+  return changed;
+}
+
+std::vector<Constant *> glow::constantFold(Node *N) {
+  LOG_SCOPE(N->getParent()->getLogContext(), "glow::constantFold")
+  std::unique_ptr<Backend> backend(createBackend("Interpreter"));
+  if (!isConstantOperation(N, *backend)) {
+    return {};
+  }
+  return constantFoldNodeImpl(*backend, N);
+}
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -1450,7 +1450,11 @@ TEST(GraphOptzTest, SliceOfSplatNodeChain) {
 
     EXPECT_EQ(F->getNodes().size(), 4);
 
-    ::glow::optimize(F, CompilationMode::Train);
+    CompilationContext cctx;
+    cctx.compMode = CompilationMode::Train;
+    // Do not perform any compile-time constant folding.
+    cctx.optimizationOpts.enableConstantFolding = false;
+    ::glow::optimize(F, cctx);
 
     // This test illustrates some inconsistency in the optimization.
     // Chain splats are not guaranteed to be optimized.
@@ -1507,7 +1511,11 @@ TEST_F(GraphOptz, ReshapeAfterSplat) {
   // Before optimization, we have 9 nodes in the graph.
   EXPECT_EQ(F_->getNodes().size(), 9);
 
-  ::glow::optimize(F_, CompilationMode::Infer);
+  CompilationContext cctx;
+  cctx.compMode = CompilationMode::Infer;
+  // Do not perform any compile-time constant folding.
+  cctx.optimizationOpts.enableConstantFolding = false;
+  ::glow::optimize(F_, cctx);
 
   // After optimization, we expect to see only 8 nodes, as Z2,R2 would be
   // replace by a new splat node.
@@ -2444,7 +2452,11 @@ TEST_F(GraphOptz, VarsCSE) {
   // placeholder).
   EXPECT_EQ(mod_.getConstants().size(), 3);
 
-  ::glow::optimize(F_, CompilationMode::Infer);
+  CompilationContext cctx;
+  cctx.compMode = CompilationMode::Infer;
+  // Do not perform any compile-time constant folding.
+  cctx.optimizationOpts.enableConstantFolding = false;
+  ::glow::optimize(F_, cctx);
 
   // Now only two variables are left; input1 and input2 have been combined,
   // but input3 has not.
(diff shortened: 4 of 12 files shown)


opti-mix:
Seems like CMake doesn't like the circular dependency I introduced. The optimizer needs to use the ExecutionEngine now. But ExecutionEngine depends on the Optimizer currently.

Any good ideas how to solve this?

@bertmaher You've solved a couple of similar issues in the past. What's your opinion about this issue?

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,91 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
     EXIT_ON_ERR(funcOrErr.takeError());
   }
 }
+
+namespace {
+/// The name of the temporary function to be used to perform constant folding.
+constexpr const char *constEvaluationFunctionName =
+    "__constEvaluationFunction__";
+
+/// \returns true of a node is a constant expression, i.e. it is a constant or
```
jfix71: typo: \returns true *if* a node...

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
 }
+
+namespace {
+/// The name of the temporary function to be used to perform constant folding.
+constexpr const char *constEvaluationFunctionName =
+    "__constEvaluationFunction__";
+
+/// \returns true of a node is a constant operation, i.e. it is a constant or
+/// all of its inputs are constants.
+bool isConstantOperation(const Node *N) {
```
jackm321: Should check if the node has side effects

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+Node *evaluateConstantOperation(Function *F, Node *C) {
+  assert(isConstantOperation(C) && "Expected a constant expression");
+  // Create a temporary execution engine to compute only this specific
+  // operation.
+  PlaceholderBindings bindings;
+  ExecutionEngine EE("Interpreter");
+  Module &mod = *F->getParent();
+  // Create a temporary function to perform the constant operation.
+  Function *constEvaluationF = mod.createFunction(constEvaluationFunctionName);
+  unsigned constIdx = 0;
```
jackm321: is this supposed to be static or something?
Seems like it's not really necessary.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+  // Return the newly created constant, which is the result of a constant
+  // operation.
+  return mod.createConstant(strFormat("const_%u", constIdx++), *outputTensor);
+}
+} // namespace
+
+bool glow::constantFold(Function *F, CompilationContext &cctx) {
+  // Avoid recursively constant folding a function that is submitted for
+  // constant folding through a temporary execution engine.
+  if (F->getName().startswith(constEvaluationFunctionName)) {
```
jackm321: I see why this is needed but slightly fragile. I wonder if there is another way to avoid this. I can't think of one right now, but maybe we can add ability to disable certain optimization passes in CompilationContext or maybe we need a more simple way to run the interpreter without doing optimizations (would also probably solve the circular dep issue)

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+  // Mapping from existing nodes to the new ones.
+  NodeMap currToNew;
+  // Clone the constant operation and some of its inputs if necessary.
+  auto *clonedC = recursiveClone(constEvaluationF, C, currToNew);
+  SaveNode *SN = constEvaluationF->createSave("save", clonedC);
+  auto *outputPH = SN->getPlaceholder();
+  bindings.allocate(outputPH);
+  // Run the temporary execution engine to perform this constant operation
+  // evaluation.
+  EE.compile(CompilationMode::Infer, constEvaluationF);
```
jfix71: We'll need to check all ops are supported by the Interpreter (`isOpSupported()`) sometime before doing this compile call.

lib/ExecutionEngine/ExecutionEngine.cpp
jfix71: IMO we should/could just skip graph optimizations in this case. There's some weirdness here where the optimizer already also does Constant folding, e.g. our current `convertConstant()` optimizations that already do some of this Constant folding for `Quantize(Constant)` etc. In this case we may end up copying the results to a new buffer unnecessarily. This could be as easy as calling `backend->compile()` instead of `EE.compile()`. Not sure this helps with the circular dependency issue though.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+  ExecutionEngine EE("Interpreter");
+  Module &mod = *F->getParent();
+  // Create a temporary function to perform the constant operation.
+  Function *constEvaluationF = mod.createFunction(constEvaluationFunctionName);
+  unsigned constIdx = 0;
+  // Mapping from existing nodes to the new ones.
+  NodeMap currToNew;
+  // Clone the constant operation and some of its inputs if necessary.
+  auto *clonedC = recursiveClone(constEvaluationF, C, currToNew);
+  SaveNode *SN = constEvaluationF->createSave("save", clonedC);
```
jfix71: Instead of having this limitation of constant folding ops with single outputs, could iterate over the outputs and create save nodes for all of them.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+      return false;
+    }
+  }
+  return true;
+}
+
+/// Evaluates a provided constant operation \p C in the context of the function
+/// \p F.
+/// \returns a constant representing a result of computation.
+Node *evaluateConstantOperation(Function *F, Node *C) {
```
jfix71: nit: Could return a Constant instead of a Node.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -266,3 +266,100 @@ void ExecutionEngine::compile(Function *F, CompilationContext &cctx,
...
+      continue;
+    }
+    if (isConstantOperation(&N)) {
+      worklist.push_back(&N);
+    }
+  }
+  // Compute the result of each collected non-trivial constant operation.
+  while (!worklist.empty()) {
+    auto *C = worklist.pop_back_val();
+    assert(C->getNumResults() == 1 && "Expected a single result");
```
jfix71: Could remove this limitation of needing an op with a single result by returning a vector of Constants corresponding to each of the results of the op that is being constant folded.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Yes, this is exactly what we discussed with @jfix71 last week. The reason I haven't done it yet is that is seemed to be not so trivial as I expected ;-)

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: I was afraid of possible name conflicts when I produce named constants in the return statement of this function. But maybe I don't need to do anything special as uniquing will be performed anyways.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Good point!

opti-mix:
@jackm321 

> Another thing, our ExecutionEngine is getting more complicated and will soon be using HostManager which then includes the partitioner, executor, other things. I see the value in doing this for testing. But for things like exactly this PR, it would still be nice if we had a really simple way to call into the Interpreter.

Yes, I totally agree. Going every time through an EE based on a HostManager just to perform this kind of optimization is very heavy-weight today. Having a simpler and lighter way to use the interpreter (or may be even the CPU backend) would be very useful.

lib/ExecutionEngine/ExecutionEngine.cpp
ksaurabh-cadence: @jackm321  I tried having a compile light with optimizations disabled, but some of the ops actually depend on the lowering/optimization passes and tests started failing. One such example is reduceMean2AvgPool.

jfix71: Yeah this is true, we may need to at least lower.

lib/ExecutionEngine/ExecutionEngine.cpp
ksaurabh-cadence: or a vector of constants?

opti-mix:
Pushed a new version. Changes are:
- Handle nodes with any number of results
- Bail on nodes with side-effects
- Bail on nodes which are not supported by the backend used for compile-time computations

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

lib/ExecutionEngine/ExecutionEngine.cpp
opti-mix: Fixed.

opti-mix:
@jfix71 

> I think part of the circular dependence problem is that right now backends that use Glow Instructions depend on the optimizer for generateAndOptimizeIR(). Perhaps we could separate our IROptimizer and GraphOptimizer into two separate libraries. I believe the GraphOptimizer lib could then depend on the Interpreter.

This change can be done independently of this PR as a prerequisite, right? Could someone more skilled in CMake-fu try to submit it? @jfix: Could you sketch such a PR?

But would it really help? Backends do not have a `run` method anymore. Only ExecutionEngine or Host/DeviceManagers have it. How do we get around this? It seems like we still need to use an ExecutionEngine or something like that.

lib/Graph/Graph.cpp
```diff
@@ -3037,6 +3037,26 @@ SaveNode *glow::getOutputSave(Function *F, Placeholder *PH) {
   return nullptr;
 }
 
+Node *glow::recursiveClone(Function *newF, Node *node, NodeMap &currToNew) {
```
beicy: According to my understanding, a constant operation's inputs could be an "constant" or an "splat" node (An op can become a constant op when its all inputs are replaced with constant or splat nodes). So only if the input is an "splat" node, we need to copy the input. So maybe the "recursive" is not necessary? Just check if the input type is "splat", then copy the input as well.

opti-mix: @beicy `recursiveClone` does not copy `Storage` nodes. And `Constant` is one of such storage nodes. So, constant won't be copied.

Also, keep in mind that `recursiveClone` is a helper function that is used not only for constant operations evaluations, but in some other places as well. There is needs to be able to copy arbitrary nodes.

jfix71:
>This change can be done independently of this PR as a prerequisite, right? Could someone more skilled in CMake-fu try to submit it? @jfix71 : Could you sketch such a PR?

See #3165.

>But would it really help? Backends do not have a run method anymore. Only ExecutionEngine or Host/DeviceManagers have it. How do we get around this? It seems like we still need to use an ExecutionEngine or something like that.

I think the EE is somewhat just a helper function for running. I believe we should be able to use its logic directly, or perhaps move that logic elsewhere so the GraphOptimizer and EE can both call it.

opti-mix:
@jfix71 Pushed a new version. 

Main changes:
- The whole function constant folding is a pass now
- There is a separate glow::constandFold function for calling e.g. from loaders. It perform a constant folding of a specific node
- The functionality does not use ExecutionEngine/HostManager/DeviceManager directly anymore. It has its own simpler methods to execute a function using the Interpreter backend.

Still open:
- How to indicate that a given compilation should not use optimizations at all or just a limited subset of optimizations (e.g. via CompilationContext options?)
- I believe too many constants are being copied around because collectConstants works at the module level, not at the function level

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+         PlaceholderBindings &bindings) {
+  std::unique_ptr<PlaceholderBindings> bindingsPtr(&bindings);
+  ExecutionContext context(std::move(bindingsPtr));
+  // Allocate only placeholders used by F.
+  for (auto ph : F.findPlaceholders()) {
+    // Don't allocate tensors for placeholders that are already allocated.
+    if (context.getPlaceholderBindings()->count(ph)) {
+      continue;
+    }
+    context.getPlaceholderBindings()->allocate(ph);
```
jfix71: Why is this necessary? Didn't we already allocate all of the save nodes' PHs? And so there shouldn't be any more PHs otherwise right?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+             CompilationContext &cctx) {
+  auto compiledF = compile(backend, F, cctx);
+  run(backend, *compiledF, F, bindings);
+}
+
+/// Evaluates a provided constant operation \p C in the context of the function
+/// \p F using the provided \p backend. Constant results are stored in \p
+/// constResults.
+void evaluateConstantOperation(
+    Backend &backend, Function *F, CompilationContext &cctx, Node *C,
```
jfix71: nit: Can we avoid passing in `F` and just get the parent of `C`? It should never be a PH/Constant anyway right?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+using llvm::dyn_cast;
+using llvm::isa;
+
+namespace {
+/// The name of the temporary function to be used to perform constant folding.
+constexpr const char *constEvaluationFunctionName =
+    "__constEvaluationFunction__";
+
+/// \returns true if a node \p N is a constant operation, i.e. it is a constant
+/// or all of its inputs are constants, and it has no side-effects and supported
```
jfix71: nit: Feels slightly misleading to say "or all of its inputs are constants" -- since its inputs may be Nodes that have Constant inputs etc. Maybe say something about how it recursively walks inputs to check if all are Constant or something.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+  }
+  return true;
+}
+
+/// Compile the function \p F for the provided \p backend using the compilation
+/// context \p cctx. \returns compiled function
+std::unique_ptr<CompiledFunction> compile(Backend &backend, Function &F,
+                                          CompilationContext &cctx) {
+  EXIT_ON_ERR(::glow::optimizeFunction(&F, backend, cctx));
```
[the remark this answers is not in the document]
opti-mix: I just pushed an update, where I added some flags to cctx. Please have a look.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: I guess I'm a bit overprotective here. IIRC, I was thinking that eventually this "simplified" way to run functions on a backend may be used elsewhere, where it is not guaranteed that all placeholders are allocated already.

But I can remove this logic for now if you think it overcomplicates things.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: Removed F from the parameter list.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: Updated the comment.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+        << "Backend must support all nodes after high-level optimizations but "
+           "encountered unsupported operator: "
+        << N.getDebugDesc();
+  }
+  auto funcOrErr = backend.compile(&F, cctx.backendOpts);
+  EXIT_ON_ERR(funcOrErr.takeError());
+  return std::move(*funcOrErr);
+}
+
+/// Runs the compiled function \p compiledF from the module \p M on the \p
```
jfix71: nit: Doxygen is stale (mentioning `M` and not `F`).

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+    auto input = N->getNthInput(idx);
+    if (!isConstantOperation(input.getNode(), backend)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/// Compile the function \p F for the provided \p backend using the compilation
+/// context \p cctx. \returns compiled function
```
jfix71: nit: period.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+/// Execute function \p F by the \p backend using the provided \p bindings and
+/// the compilation context \p cctx.
+void execute(Backend &backend, Function &F, PlaceholderBindings &bindings,
+             CompilationContext &cctx) {
+  auto compiledF = compile(backend, F, cctx);
+  run(backend, *compiledF, F, bindings);
+}
+
+/// Evaluates a provided constant operation \p C in the context of the function
+/// \p F using the provided \p backend. Constant results are stored in \p
```
jfix71: nit: remove `F` from doxygen, mention `cctx`.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,219 @@
...
+/// constResults.
+void evaluateConstantOperation(
```
jfix71: nit: instead of `void` and passing `constResults`, could return `SmallVector`.

opti-mix: I'll return an std::vector as it is simpler for the client code.

ksaurabh-cadence:
@opti-mix Please refer to 3141 for the copy constant issue. I have been able to avoid copying of constants and pre-mature deletion of constants in the parent graph/module because of DCE. Hope it is useful.

opti-mix:
@ksaurabh-cadence 

> Please refer to 3141 for the copy constant issue. I have been able to avoid copying of constants and pre-mature deletion of constants in the parent graph/module because of DCE. Hope it is useful.

Could you elaborate which specific kinds of constant copies are avoided? From the loader into a module? From a module into the ExecutionEngine? Anything else? I'm trying to understand if I could use your trick for a regular constant folding, which may happen outside of loaders.

> pre-mature deletion of constants in the parent graph/module because of DCE.

I introduced in this PR optimization flags into CompilationContext. In principle we could just add a flag to disable DCE. It could be simpler than adding state to Functions.

ksaurabh-cadence:
@opti-mix 

Here is the flow in 3161-

In Loader thers is a main Function which is being constructed. As we encounter an op we check if it's foldable.  If it is foldable we create a temporary Function. This is where constant inputs to the nodes were being copied  basically function to function. This was done to avoid deletion of constants when DCE is called for temporary Function. Introduction of FunctionStates avoids constants from being deleted unless all the functions associated module are loaded. You should not be deleting any constants unless all the Functions are completely loaded, as constants are shared resource. 

Please look at my test cases, it is pretty exhaustive on constant folding. I realized disabling DCE has it's side effects, and ops like AddBroadcast require DCE without constant deletion.

opti-mix:
@jfix71 Pushed an updated version.

Changes:
- made `executeFunction` API available to clients. Checked that it works properly with #3141 if calls to ExecutionEngine are replaced by this call.
- Improved the logic for adding nodes to the worklist. It should results in fewer calls of `evaluateConstantOperation` and thus introduce less overhead.
- The cctx flags for the recursive compile calls only prohibit constant folding, but do not prohibit graph optimizations anymore. This could be useful for some cases, where e.g. lowering should happen first or something like this.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,277 @@
...
+  }
+  return llvm::Error::success();
+}
+
+} // namespace
+
+llvm::Error glow::executeFunction(Backend &backend, Function &F,
+                                  PlaceholderBindings &bindings,
+                                  CompilationContext &cctx, bool isConstant) {
+  if (isConstant) {
```
jfix71: So the motivation here of only returning this error if `isConstant` is if there's an external caller that doesn't expect to use this with a Constant expression? Do you have an expected use case for the future in this way?

If so I'd suggest moving it somewhere other than `ConstantFolding.cpp` as it's kind of confusing, maybe even to a different library so it can be reused without depending on the GraphOptimizer, e.g. somewhere in the runtime.

If not then I'd suggest removing `isConstant` and rename it to `executeConstantFunction()` and always return an error in this case.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+/// Perform constant folding in the function \p F . Any non-trivial node (i.e.
+/// not a constant or a splat) that can be computed at compile-time is going to
+/// be computed at compile-time. \returns true if any foldings were performed.
+bool glow::ConstantFold::run(Function *F) {
+  LOG_SCOPE(F->getLogContext(), "glow::constantFold")
+  CompilationContext cctx;
+  // Graph optimizations may be required to e.g. perform lowering of some nodes
+  // that are not supported natively.
+  cctx.optimizationOpts.enableGraphOptz = true;
```
jfix71: Do we need to explicitly set this to true?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,277 @@
...
+    auto input = N->getNthInput(idx);
+    if (!isConstantOperation(input.getNode(), backend)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/// \returns true if node \p N has at least one non-constant operation user.
+bool hasNonConsantOperationUser(const Node *N, const Backend &backend) {
```
jfix71: typo: "Consant"

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,277 @@
...
+    for (unsigned idx = 0, e = constResults.size(); idx < e; ++idx) {
+      auto constResult = constResults[idx];
+      // Replace the old result by the new constant result.
+      C->getNthResult(idx).replaceAllUsesOfWith(constResult);
+      // Add all users of constResult to the worklist if they became constant
+      // operations.
+      for (auto &use : constResult->getUsers()) {
+        auto *user = use.getUser();
+        if (isConstantOperation(user, *backend)) {
+          worklist.push_back(user);
```
jfix71: I'm not sure we need the worklist pattern at this point actually given the updated use of `hasNonConstantOperationUser()`. Is it possible for there to be ops that are newly constant-foldable after we do a round of constant folding of what was initially put in the worklist?

opti-mix: Yeah, I guess not. We can probably process the found constant nodes as we go without any worklist. The only scenario where it may fail is when new nodes are added to the function during constant folding (e.g. during lowering or something like that). But it probably highly unlikely.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: Removed `isConstant` for now.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: Removed the worklist.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,276 @@
...
+                        "Expected constant operation");
+      continue;
+    }
+    if (!N.hasOneUse()) {
+      RETURN_ERR("Expected constant operation");
+    }
+    auto *SN = dyn_cast<SaveNode>(N.getUsers().begin()->getUser());
+    if (SN && SN->getPlaceholder() == &N) {
+      continue;
+    }
```
jfix71: Don't you need to return an error at the end of the for loop here, since it's a PH that isn't used by a Save?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,276 @@
...
+  }
+  // Remove the temporary function.
+  mod.eraseFunction(constEvaluationF);
+  return constResults;
+}
+
+/// Check if function \p F consists of constant operations only.
+llvm::Error verifyConstantFunction(Backend &backend, Function &F) {
+  // Perform the checks in DEBUG builds only.
+#ifndef NDEBUG
```
jfix71: IMO this makes the function somewhat misleading -- perhaps we could instead wrap the actual call to `verifyConstantFunction()` in `NDEBUG`?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,276 @@
...
+    // overhead.
+    if (!hasNonConstantOperationUser(N, *backend)) {
+      continue;
+    }
+    // Compute the constant value of the node.
+    std::vector<Constant *> constResults =
+        evaluateConstantOperation(*backend, cctx, N);
+    // Replace all results of the original operation by the computed
+    // compile-time results of this operation.
+    for (unsigned idx = 0, e = constResults.size(); idx < e; ++idx) {
```
jfix71: nit: `size_t`, not `unsigned`

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2882,3 +2894,92 @@ TEST_F(GraphOptz, nopRelu) {
...
+  auto *splat4 = F_->createSplat(
+      "splat2", mod_.uniqueType(ElemKind::FloatTy, {2, 2}), 4.0f);
+
+  auto *add1 = F_->createAdd("add", const1, const2);
+  auto *mul1 = F_->createMul("mul1", add1, splat2);
+  auto *mul2 = F_->createMul("mul2", mul1, splat3);
+  auto *sub = F_->createSub("sub", mul2, const3);
+  auto *add2 = F_->createAdd("add2", sub, const4);
+  auto *mul3 = F_->createMul("mul3", add2, splat4);
+  auto *SN1 = F_->createSave("save", mul3);
```
jfix71: Can you add a TopK or something with multiple results that would be the final Node in the Constant tree, so we test correctness of ConstantProp with more than 1 result?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
opti-mix: Good catch! Thanks!

tests/unittests/GraphOptzTest.cpp
opti-mix: Fixed.

opti-mix:
@jfix71 

> In terms of placement of the ConstantFolding pass in the opt pipeline, do you think it makes sense to have at the end? I think it's not dependent on other optimizations and could enable other optimizations, so I think it makes sense to do early, perhaps after CSE/DCE?

Yeah, we could experiment with the placement of constant folding. So far I put it pretty late, because I seem to remember seeing a use-case where running it early would not catch enough opportunities. 

> So if we push up the pass we should probably remove all such Constant prop passes we already have, as they're redundant anyway.

Yes, in theory it makes some of those passes redundant. But removing them and experimenting with a better placement should be done in subsequent PRs, IMHO.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2882,3 +2894,96 @@ TEST_F(GraphOptz, nopRelu) {
...
+
+  auto *add1 = F_->createAdd("add", const1, const2);
+  auto *mul1 = F_->createMul("mul1", add1, splat2);
+  auto *mul2 = F_->createMul("mul2", mul1, splat3);
+  auto *sub = F_->createSub("sub", mul2, const3);
+  auto *add2 = F_->createAdd("add2", sub, const4);
+  auto *mul3 = F_->createMul("mul3", add2, splat4);
+  // Check compile-time constant folding for nodes with multiple results.
+  auto *topK = F_->createTopK("topK", mul3, 2);
+  auto *SN1_0 = F_->createSave("save", topK->getNthResult(0));
```
jfix71: nit: `topK->getValues()`

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2882,3 +2894,96 @@ TEST_F(GraphOptz, nopRelu) {
...
+  auto *SN1_1 = F_->createSave("save", topK->getNthResult(1));
```
jfix71: nit: `topK->getIndices()`

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
jfix71: You still have some of the logic here left, i.e. iterating over users and pushing to the worklist. I think you can remove all of it at this point.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
jfix71: Why do we set this to true here but false inside `glow::constantFold()`?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,279 @@
...
+    // operation containing the current node can completely replace the result
+    // of its computation. Doing this check allows for performing a smaller
+    // number of evaluateConstantOperation calls later and thus reduces the
+    // overhead.
+    if (!hasNonConstantOperationUser(N, *backend)) {
+      continue;
+    }
+    // Compute the constant value of the node.
+    std::vector<Constant *> constResults =
+        evaluateConstantOperation(*backend, cctx, N);
```
jfix71: NBD, just a thought: with the worklist removal, I'm wondering if we could just call into `glow::constantFold(Node *N)` instead of replicating some of this logic. I see we've already checked above `isConstantOperation()` so as is there'd be some redundant checks.

opti-mix:
@jfix71 Addressed you latest review comments.

include/glow/Optimizer/GraphOptimizer/CompilationContext.h
```diff
@@ -49,6 +49,16 @@ struct PrecisionConfiguration {
 
 using QuantizationMode = PrecisionConfiguration::QuantizationMode;
 
+/// Options relevant to optimizations during compilation.
+struct OptimizationOptions {
+  /// If true, perform graph optimizations.
+  bool enableGraphOptz{true};
```
jfix71: Is this ever set to false? Is it still necessary?

include/glow/Optimizer/GraphOptimizer/CompilationContext.h
```diff
@@ -49,6 +49,16 @@ struct PrecisionConfiguration {
...
+  /// If true, perform low-level IR optimizations.
+  bool enableIROptz{true};
```
jfix71: I don't think this is used either.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,271 @@
...
+#include "glow/Graph/PlaceholderBindings.h"
+#include "glow/Graph/Utils.h"
+#include "glow/Optimizer/GraphOptimizer/FunctionPasses.h"
+
+#include "llvm/Support/Casting.h"
+
+using namespace glow;
+using llvm::cast;
+using llvm::dyn_cast;
+using llvm::isa;
```
jfix71: nit: you added these but are still using the namespace with the commands below everywhere (e.g. `llvm::isa`).

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,271 @@
...
+/// bindings.
+void run(Backend &backend, CompiledFunction &compiledF,
+         PlaceholderBindings &bindings) {
+  std::unique_ptr<PlaceholderBindings> bindingsPtr(&bindings);
+  ExecutionContext context(std::move(bindingsPtr));
+  // TODO: Add only constants used by F to the compiled function. This should
+  // reduce the amount of data that needs to be copied.
+  auto executeErr = compiledF.execute(&context);
+  EXIT_ON_ERR(std::move(executeErr));
+  // don't delete bindings.
```
jfix71: nit: Capitalize sentence.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -0,0 +1,271 @@
...
+  // evaluation.
+  EXIT_ON_ERR(
+      executeConstantFunction(backend, *constEvaluationF, bindings, cctx));
+  // Get the results of the constant operation compile-time computation and
+  // create new constants from it.
+  std::vector<Constant *> constResults;
+  constResults.reserve(savedResults.size());
+  for (auto *SN : savedResults) {
+    Tensor *outputTensor = bindings.get(SN->getPlaceholder());
+    auto *constResult = mod.createConstant(SN->getName(), *outputTensor);
```
jfix71: nit: `std::move(*outputTensor)` here so we don't need to copy the tensor payload.

include/glow/Optimizer/GraphOptimizer/CompilationContext.h
opti-mix: Not for this PR. But could be useful in general. I'll remove it from this PR to keep it focused. These flags probably belong more into the PassManager anyways ;-)

PR #3162, opened 2019-06-22
Operator test for SLS with partial tensors

bertmaher:
Summary:
The Habana backend can transfer partial tensors (both during memcpy to
pinned host memory and across PCIe) as an optimization.  Test this with
variable-size embedding indices.

Differential Revision: D15956397


jfix71:
We aren't checking results are correct at all. Could we compare execution here to the same SLS but without partial? E.g. in this example create the same graph but with the length of `indices` as 20 instead of 20000, and run again on the Interpreter and compare results.

bertmaher:
With partial tensors I'm most worried about straight-up segfaults, but yeah I can add an output check too.

PR #3164, opened 2019-06-24
[CI] Make CI for glow-torch

zrphercule:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -127,6 +127,7 @@ elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     cd ..
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
+	git checkout 7fcfed19e7c4805405f3bec311fc056803ca7afb
     pip install -r requirements.txt
     sudo /usr/bin/python3.6 setup.py install
     cd ${GLOW_DIR}
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -123,14 +123,14 @@ elif [[ "$CIRCLE_JOB" == "CHECK_CLANG_FORMAT" ]]; then
 elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     # Build PyTorch
     cd /tmp
-	python3.6 -m virtualenv venv
-	source "venv/bin/activate"
+    python3.6 -m virtualenv venv
+    source "venv/bin/activate"
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
     git checkout 7fcfed19e7c4805405f3bec311fc056803ca7afb
     pip install -r requirements.txt
     python setup.py install
-	cd ${GLOW_DIR}
+    cd ${GLOW_DIR}
     cd build
 elif [[ "$CIRCLE_JOB" == "OPENCL" ]]; then
     install_pocl
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -6,6 +6,19 @@ set -ex
 
 export MAX_JOBS=8
 
+if hash sccache 2>/dev/null; then
+    SCCACHE_BIN_DIR="/tmp/sccache"
+    mkdir -p "$SCCACHE_BIN_DIR"
+    for compiler in cc c++ gcc g++ x86_64-linux-gnu-gcc; do
+        (
+            echo "#!/bin/sh"
+            echo "exec $(which sccache) $(which $compiler) \"\$@\""
+        ) > "$SCCACHE_BIN_DIR/$compiler"
+        chmod +x "$SCCACHE_BIN_DIR/$compiler"
+    done
+    export PATH="$SCCACHE_BIN_DIR:$PATH"
+fi
+
 install_pocl() {
    sudo apt-get install -y ocl-icd-opencl-dev clinfo libhwloc-dev opencl-headers
 
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -124,7 +124,7 @@ elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     # Build PyTorch
     cd /tmp
     python3.6 -m virtualenv venv
-    source "venv/bin/activate"
+    source venv/bin/activate
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
     git checkout 7fcfed19e7c4805405f3bec311fc056803ca7afb
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -122,15 +122,15 @@ elif [[ "$CIRCLE_JOB" == "CHECK_CLANG_FORMAT" ]]; then
     sudo apt-get install -y clang-format-7
 elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     # Build PyTorch
-    sudo pip install numpy
-    cd ${GLOW_DIR}
-    cd ..
+    cd /tmp
+	python3.6 -m virtualenv venv
+	source "venv/bin/activate"
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
-	git checkout 7fcfed19e7c4805405f3bec311fc056803ca7afb
+    git checkout 7fcfed19e7c4805405f3bec311fc056803ca7afb
     pip install -r requirements.txt
-    sudo /usr/bin/python3.6 setup.py install
-    cd ${GLOW_DIR}
+    python setup.py install
+	cd ${GLOW_DIR}
     cd build
 elif [[ "$CIRCLE_JOB" == "OPENCL" ]]; then
     install_pocl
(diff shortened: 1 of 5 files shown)


jackm321:
We probably don't want to build from source, it will take too long

jackm321:
Also I don't think that this should pull in new versions automatically (example pip install pytorch-nightly) because then something could change on the pytorch side and this will break the Glow CI for unrelated Glow diffs. It would be better if we could have control of when the pytorch version we test against is updated for example by installing a version into a docker image and updating that periodically.

zrphercule:
> Also I don't think that this should pull in new versions automatically (example pip install pytorch-nightly) because then something could change on the pytorch side and this will break the Glow CI for unrelated Glow diffs. It would be better if we could have control of when the pytorch version we test against is updated for example by installing a version into a docker image and updating that periodically.

Agreed. I am just dealing with the building here, that's the only reason I am building pytorch from src... It seems no matter how I try it is using C++ 17, which deprecated "register" in pybind11

jackm321:
> It seems no matter how I try it is using C++ 17, which deprecated "register" in pybind11

Weird, I get that error on my mac as well but its just a warning (i guess the difference is `-Werror`). I think upgrading to python 3 should fix this though, have you tried that?

zrphercule:
> > It seems no matter how I try it is using C++ 17, which deprecated "register" in pybind11
> 
> Weird, I get that error on my mac as well but its just a warning (i guess the difference is `-Werror`). I think upgrading to python 3 should fix this though, have you tried that?

Let me take a try, but, why does python 3 fix C++ standard error....

zrphercule:
> > > It seems no matter how I try it is using C++ 17, which deprecated "register" in pybind11
> > 
> > 
> > Weird, I get that error on my mac as well but its just a warning (i guess the difference is `-Werror`). I think upgrading to python 3 should fix this though, have you tried that?
> 
> Let me take a try, but, why does python 3 fix C++ standard error....

I see why now, pybind11 maybe better right....

jackm321:
Yeah supposedly the python3 version doesn't use this deprecated c++ feature anymore

.circleci/config.yml
```diff
@@ -76,7 +76,7 @@ jobs:
   OPENCL:
     environment:
       DOCKER_IMAGE: "308535385114.dkr.ecr.us-east-1.amazonaws.com/caffe2/py2-clang8-ubuntu16.04:283"
-    <<: *linux_default
```
jackm321: nit this change seems unnecessary

CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ option(GLOW_WITH_BUNDLES "Build bundles" OFF)
 option(LINK_PROTOBUF_AS_DLL "Link against protobuf build as dynamic libray." OFF)
 
 set(CMAKE_CXX_STANDARD 14)
+set(CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
jackm321: why are these changes needed?

.circleci/test.sh
```diff
@@ -46,6 +46,10 @@ run_and_check_resnet50_bundle() {
     done
 }
 
+run_pytorch_tests() {
+  python3.6 "${GLOW_SRC}/torch_glow/setup.py" test --run_cmake
```
jackm321: probably better to just cd to ${GLOW_SRC}/torch_glow and run from there

.circleci/build.sh
```diff
@@ -50,8 +61,15 @@ hash cmake ninja
 GLOW_DIR=$PWD
 cd ${GLOW_DIR}
 mkdir build && cd build
-CMAKE_ARGS=("-DCMAKE_CXX_COMPILER=/usr/bin/clang++-8")
-CMAKE_ARGS+=("-DCMAKE_C_COMPILER=/usr/bin/clang-8")
+
+if [[ "${CIRCLE_JOB}" == "PYTORCH" ]]; then
+    CMAKE_ARGS=("-DCMAKE_CXX_COMPILER=/usr/bin/clang++-7")
```
jackm321: I know you mentioned this but why do we need clang 7 for pytorch?

zrphercule:
> The test output says, do you know why this is?
> 
> ```
> Ran 0 tests in 0.000s
> 
> OK
> ```

Not really, python torch_glow/setup.py test shows the same thing locally & in my docker image. I will try to follow your suggestion by cd into torch_glow first

CMakeLists.txt
zrphercule: These are to be removed lol

.circleci/build.sh
zrphercule: Actually we dont. It is because we need a python3-clang7.0+ docker image, and we dont have python3-clang8 image......

.circleci/config.yml
zrphercule: hmmmmmmm not so sure what have been changed here...

zrphercule:
@jackm321 This pr is ready to be merged; however, currently CI is super slow because we are building it from src, like one hour waiting CI to be passed. 
I think so far 1 hour is tolerable, what is your opinion?

.circleci/build.sh
rdzhabarov: what stops from making one?

zrphercule:
> Current open source CI takes 10 mins to run which is nice and gives very quick signal (though ultimately we are still bottlenecked with sandcastle for merging)
> 
> @jackm321 @zrphercule any thoughts on how to reduce that (1 hr run)?
> Could we put more pre build stuff into docker container, etc?

I agree with that 1 hr is still too long for OSS building, at last we need to reduce the build time.
One solution, as you mentioned, is to build a docker image on our own, with a stable pytorch version and some environment we need (py3 & clang8). The only problem is we might need to sync pytorch periodically with this docker image to make sure that everything works fine (which I guess is not a problem maybe)

The discussion about this pr is whether we can afford 1 or 2 weeks slow build before we update the docker image (since I am not confident to say I can finish it in a day or two, although it looks easy but it is CI related, who knows...), since we might want to merge a lot of things including the support of resnet soon with this CI.

.circleci/build.sh
zrphercule: > what stops from making one?

nothing, will do this tomorrow.

yinghai:
Let's setup `sccache` and see how it helps.

rdzhabarov:
> The only problem is we might need to sync pytorch periodically with this docker image to make sure that everything works fine (which I guess is not a problem maybe)

This would require some manual work every time we do update of pytorch and likely we want to build against latest. I'd try to speed up existing approach.

Do you know rough breakdown where most of the time spend?

Is it just building pytorch (sccache as @yinghai mentioned, do we explore parallelism during the build)?
Setting up pytorch dependencies (those are easy to preinstall into docker)

jackm321:
I agree 1 hour is too long and we should look for an alternative
@rdzhabarov out of curiosity, how painful is it to update a docker image each time we want to update PyTorch version?
sccache sounds like it could be a better option

zrphercule:
@jackm321 Could you please merge this for me? I cant...

rdzhabarov:
@zrphercule perfect, thank you! Have you tried disabling compiling with CUDA as well? Does it speed up things?

zrphercule:
> @zrphercule perfect, thank you! Have you tried disabling compiling with CUDA as well? Does it speed up things?
Not really, since we dont have CUDA it is ignored actually. Anyway right now it is reduced to 15 min which I think is good enough. Cheers!

rdzhabarov:
Before merging please check why "coverage" now takes 20 mins instead of 10mins. Build time increased by a lot.

check:
https://circleci.com/gh/pytorch/glow/24822?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
vs
https://circleci.com/gh/pytorch/glow/24729?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
(or other previous coverage builds).

Pytorch test time looks great.

.circleci/config.yml
```diff
@@ -97,6 +97,11 @@ jobs:
     environment:
       DOCKER_IMAGE: "308535385114.dkr.ecr.us-east-1.amazonaws.com/caffe2/py2-clang8-ubuntu16.04:283"
     <<: *linux_default
+  PYTORCH:
+    resource_class: 2xlarge+
```
rdzhabarov: xlarge? I do not see this in the documentation: https://circleci.com/docs/2.0/configuration-reference/#resource_class

zrphercule:
Coverage time is reduced to 10 min after removed sccache from Coverage test.
Let's go!

.circleci/config.yml
zrphercule: hmmmm this is a recommendation of tvm. will double check.

rdzhabarov:
> hmmmm this is a recommendation of tvm. will double check.

Does not seem to work. Let's make that xlarge and merge this PR.

zrphercule:
> > hmmmm this is a recommendation of tvm. will double check.
> 
> Does not seem to work. Let's make that xlarge and merge this PR.

lol this diff is just landed internally. I will make another one to change this...

rdzhabarov:
Sounds good!

PR #3165, opened 2019-06-24
[Optimizer] Refactor Optimizer lib into {Graph,IR}Optimizer libs

jfix71:
Summary: Allow for more precise dependences on libs. Most places rely on either the Graph or IR Optimizer, not both. Related to #3161.

Test Plan: All tests pass.


rdzhabarov:
Is this going to be purely change in the open source only?

jfix71:
No, I'll make similar changes in fbcode. I tend to always work open-source first and make fbcode changes after the open source ones have been approved.

rdzhabarov:
Have not noticed moving cpps around and thought it's just CMake changes, yeah, sounds good.

opti-mix:
@jfix71 For some reason, DEBUG builds shows a lot of duplicate symbols errors.

rdzhabarov:
> @jfix71 For some reason, DEBUG builds shows a lot of duplicate symbols errors.

Those are the only one which uses open cl, seems like it's tests/OCLTest

jfix71:
Working on fixing. Not sure what's going on -- I need to include `OpenCLBackend` in `OCLTest` on my Mac for it to compile, but that causes duplicate symbols on Linux. But if I don't include `OpenCLBackend`  in `OCLTest` then it compiles on Linux but my Mac has missing symbols 🙁

opti-mix:
> Working on fixing. Not sure what's going on -- I need to include OpenCLBackend in OCLTest on my Mac for it to compile, but that causes duplicate symbols on Linux. But if I don't include OpenCLBackend in OCLTest then it compiles on Linux but my Mac has missing symbols 🙁

I think it is due to our dynamic backends registrations and whole-archive options, which work differently on Linux and MacOS. @rdzhabarov may have some war stories about it ;-)

tests/unittests/CMakeLists.txt
```diff
@@ -330,7 +325,7 @@ if(GLOW_WITH_OPENCL)
                           ExecutionEngine
                           Graph
                           IR
-                          Optimizer
+                          OpenCLBackend
```
rdzhabarov: this might be it.
As @opti-mix mentioned order of libs matters for linux linker. There might be two entries for oclbackend here, you could check with: ninja -v the exact command.

jfix71: It seems to work on both if I include Backends instead of OpenCLBackend. Does that seem sane?

jfix71:
OCLTest works now when I include Backends instead of OpenCLBackend. I also moved/updated all the headers.

tests/unittests/CMakeLists.txt
rdzhabarov: Yeah, that sounds good.

PR #3166, opened 2019-06-25
Name nodes by name, output(0), type

bertmaher:
Summary:
With D15859961 we lost the naming of nodes by output, leading to a lot
of things named by type, which makes debugging harder.

Differential Revision: D15981964

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -85,6 +85,9 @@ template <typename T> std::string loadOperatorName(const T &op) {
   if (op.name().length()) {
     return op.name();
   }
+  if (op.output_size() > 0) {
+    return op.output(0);
+  }
   return op.op_type();
 }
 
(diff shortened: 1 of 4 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -1344,7 +1344,7 @@ bool HabanaBackend::isOpSupported(const NodeInfo &NI) const {
   case Kinded::Kind::AddNodeKind:
   case Kinded::Kind::AvgPoolNodeKind:
   case Kinded::Kind::BatchedAddNodeKind:
-  case Kinded::Kind::BatchedReduceAddNodeKind:
+  //case Kinded::Kind::BatchedReduceAddNodeKind:
```
rdzhabarov: spurious?

hyuen: This was not intended to be part of the diff

bertmaher:
Sigh, the unrelated changes are due to a bad stack export.  Apparently ShipIt’s “preview” doesn’t actually show you the full set of things it’s going to push.

PR #3167, opened 2019-06-25
Glow model onnx exporters

putivsky:
Summary:
Glow exporter (ONNX) serializes the Glow Function graph from memory into the file according to ONNX standard (extended). 
Documentation:
Three layers of classes (similar to Glow loaders architecture). Protobuf writer, CommonOperatorWriter, and ONNXModelWriter.
[Optional Fixes #issue]

Test Plan:


lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+
+        // Handle placeholders cases.
+        if (kind == Kinded::Kind::PlaceholderKind) {
+          const auto *PH = llvm::dyn_cast<Placeholder>(N);
+          // Write global input, output only tensor shape.
+          auto *inputProto = graphProto->add_input();
+          tensorShapeFromPlaceholder(PH, inputProto);
+        } else {
+          // Write global initializer, output tensor bytes.
+          const auto *C = llvm::dyn_cast<Constant>(N);
```
opti-mix: Since you do not expect this cast to fail you should use `llvm::cast`, which will assert if the value cannot be casted.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+
+void ONNXModelWriter::writeTensor(const Tensor &T, TensorType *out) {
+  const auto &type = T.getType();
+  out->set_data_type(convertType(type));
+  const auto &dims = type.dims();
+  for (unsigned b = 0, e = dims.size(); b < e; ++b) {
+    out->add_dims(dims[b]);
+  }
+
+  out->set_raw_data(std::string(T.getUnsafePtr(), type.getSizeInBytes()));
```
opti-mix: Question: Do we need to care about the endianness of the tensor data? Does ONNX prescribe what needs to be used?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+
+llvm::Error ONNXModelWriter::writeArithmetic(const std::string &opName,
+                                             const Node *node, GraphType &graph,
+                                             ReporterType &reporter) {
+  auto *proto = graph.add_node();
+  proto->set_name(node->getName());
+  proto->set_op_type(opName);
+  outputsToProto(node, proto);
+
+  const auto *LHS = node->getNthInput(MulNode::LHSIdx).getNode();
```
opti-mix: You may want to explain why `MulNode::LHSIdx` is used here ;-) I guess because all arithmetic binary ops are structurally the same? May be introduce a helper var like: `unsigned lhsIdx = MulNode::LHSId;` and use it instead. Same for the RHS.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+  RETURN_ERR_IF_NOT(CDN, "Invalid node type, expected CPUConvDKKC8Node.");
+
+  auto *attr = proto->add_attribute();
+  attr->set_name("kernel_shape");
+  attr->set_type(AttrType::INTS);
+  for (auto value : CDN->getKernels()) {
+    attr->add_ints(value);
+  }
+
+  attr = proto->add_attribute();
```
opti-mix: Seems like this whole block is often repeated. Seems like a good candidate for introducing a helper method. There are many similar, but slightly different blocks that could also benefit from helpers.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+}
+
+// Exporting arithmetic operations.
+#define DEF_ARITHMETIC_WRITER_NODE(NAME)                                       \
+  llvm::Error ONNXModelWriter::write##NAME(const Node *node, GraphType &graph, \
+                                           ReporterType &reporter) {           \
+    return writeArithmetic(#NAME, node, graph, reporter);                      \
+  }
+
+DEF_ARITHMETIC_WRITER_NODE(Add)
```
opti-mix: Glow also has unary arithmetic operations. You needs to support them as well in `writeArithmetic`. Or you could have `writeBinaryArithmetic` and `writeUnaryArithmetic`.

tests/unittests/OnnxExporterTest.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace {
+void testLoadAndSaveONNXModel(const std::string &name) {
+  ExecutionEngine EE{};
+  auto &mod = EE.getModule();
+  Function *F = mod.createFunction("main");
+
+  llvm::Error err = llvm::Error::success();
+  ONNXModelLoader onnxLD(name, {}, {}, *F, &err);
+
+  if (!err) {
```
opti-mix: Use early exit instead? And maybe show a message that an ONNX model could not be loaded.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1367 @@
...
+  default:
+    RETURN_ERR_IF_NOT(false,
+                      "Can't find Splat/Reshape Node as part of PRelu Node.");
+  }
+
+  outputsToProto(node, proto);
+  return llvm::Error::success();
+}
+
+llvm::Error ONNXModelWriter::writeGather(const Node *node, GraphType &graph,
```
opti-mix: Many of these serialization methods seems to have a lot of boilerplate code. 

I wonder if it is possible to automate their serialization by means of auto-generating this code using NodeGen or something like it.

lib/Exporter/ONNXModelWriter.cpp
putivsky: I think default is little-endian:
https://github.com/Microsoft/onnxruntime/blob/master/onnxruntime/core/framework/tensorprotoutils.cc#L49

lib/Exporter/ONNXModelWriter.cpp
putivsky: Can you, please, provide an example of the unary arithmetic operation Glow nodes?

opti-mix: I was thinking about Log, Tanh, Sigmoid and the like.

lib/Exporter/ONNXModelWriter.cpp
opti-mix: OK. It means we need to take care of always producing little-endian here.

lib/Exporter/ONNXModelWriter.cpp
putivsky: Either  we need a new Glow loader then Glow exporter can be very much formalized and auto-generated, however if we want to use ONNX Loader (with some extension to support Glow specific nodes), it's not trivial to automate all special cases, for instance Conv requires Transposition of input and weights, whereas TransposeFC doesn't, etc.

lib/Exporter/ONNXModelWriter.cpp
putivsky: Log, Tanh, Sigmoid and similar operators got written by default, no special handling is required, see macro DEF_ALL_WRITER_NODE

putivsky:
@jfix71, I think we can have "extended" ONNX format for Glow operators for internal use only.
ONNX format has been designed very well, it has inputs, outputs, nodes, attributes, and initializers (tensors). All pieces for Glow serialization and deserialization are there. Later we can convert it into Glow format if needed and don't call it ONNX.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+
+llvm::Error ONNXModelWriter::writeLocalResponseNormalization(
+    const Node *node, GraphType &graph, ReporterType &reporter) {
+  auto *proto = graph.add_node();
+  proto->set_name(node->getName());
+  proto->set_op_type("LRN");
+  outputsToProto(node, proto);
+  // Find input from Transpose node
+  const Node *TN =
+      node->getNthInput(LocalResponseNormalizationNode::InputIdx).getNode();
```
jfix71: Same -- prefer using `llvm::dyn_cast` here and using the Node itself below. I.e. once `writeLocalResponseNormalization()` passes in an actual `LRNNode *` instead of a `Node *`, could simply be `const TransposeNode *TN = llvm::dyn_cast<TransposeNode>(LRN->getInput());`

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+                                         ReporterType &reporter) {
+  const bool isBatched = node->getKind() == Kinded::Kind::BatchMatMulNodeKind;
+  auto *proto = graph.add_node();
+  proto->set_name(node->getName());
+  proto->set_op_type("MatMul");
+
+  // Check if LHS/RHS are Transpose node
+  Node *LHS =
+      llvm::cast<TransposeNode>(node->getNthInput(BatchMatMulNode::LHSIdx));
+  if (LHS) {
```
jfix71: This is broken -- you need to use `llvm::dyn_cast<TransposeNode>` here, as `llvm::cast` does not return a nullptr if the type is incorrect.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+    proto->add_input(
+        LHS->getNthInput(TransposeNode::InputIdx).getNode()->getName());
```
jfix71: Since LHS is a `TransposeNode`, use `LHS->getInput()`.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+    proto->add_input(node->getNthInput(isBatched ? BatchMatMulNode::LHSIdx
+                                                 : MatMulNode::LHSIdx)
+                         .getNode()
+                         ->getName());
+  }
+  Node *RHS =
+      llvm::dyn_cast<TransposeNode>(node->getNthInput(BatchMatMulNode::RHSIdx));
+  if (RHS) {
+    proto->add_input(
+        RHS->getNthInput(TransposeNode::InputIdx).getNode()->getName());
```
jfix71: nit: `RHS->getInput()`.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+ * limitations under the License.
+ */
+
+#include "glow/Exporter/ONNXModelWriter.h"
+#include "glow/Graph/Utils.h"
+
+namespace glow {
+
+namespace {
+template <bool IsInteger, typename T> struct AttributeAssigner {
```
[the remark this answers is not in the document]
putivsky: It does, check the partial template specialization at lines 58 and invocation at line 71.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+  proto->set_name(node->getName());
+  proto->set_op_type(opName);
+  outputsToProto(node, proto);
+
+  const auto *LHS = node->getNthInput(lhsIdx).getNode();
+  if (LHS->getKind() == Kinded::Kind::TileNodeKind) {
+    LHS = unwindBroadcastInput(LHS, reporter);
+    RETURN_ERR_IF_NOT(LHS, "Can't unwind Tile node.");
+    proto->add_input(
+        LHS->getNthInput(ReshapeNode::InputIdx).getNode()->getName());
```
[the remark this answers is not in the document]
putivsky: unwindBroadcastInput assumes the last in the chain is ReshapeNode.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+  } else {
+    proto->add_input(LHS->getName());
+  }
+
+  const auto *RHS = node->getNthInput(rhsIdx).getNode();
+  if (RHS->getKind() == Kinded::Kind::TileNodeKind) {
+    RHS = unwindBroadcastInput(RHS, reporter);
+    RETURN_ERR_IF_NOT(RHS, "Can't unwind Tile node.");
+    proto->add_input(
+        RHS->getNthInput(ReshapeNode::InputIdx).getNode()->getName());
```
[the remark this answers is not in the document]
putivsky: unwindBroadcastInput assumes the last in the chain is ReshapeNode.

lib/Exporter/ONNXModelWriter.cpp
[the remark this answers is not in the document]
jfix71: Ok, but `Node::getNthInput()` simply returns the input of the current Node. `LHS->getNthInput(TileNode::InputIdx)` (or `LHS->getInput()` as I suggest) will return a `NodeValue` with that `ReshapeNode`, but you shouldn't use `ReshapeNode::InputIdx` as the index into the `TileNode`'s inputs. In general this is less confusing if you just use `TileNode::getInput()`, so I think we should use it instead.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1107 @@
...
+  case Kinded::Kind::TileNodeKind:
+    return true;
+  default:
+    return false;
+  }
+}
+
+const Node *ONNXModelWriter::unwindBroadcastInput(const Node *tile,
+                                                  ReporterType &reporter) {
+  while (tile && tile->getKind() == Kinded::Kind::TileNodeKind) {
```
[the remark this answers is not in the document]
putivsky: It is a little more complicate algorithm, i.e. it should return the last node, which is not a Tile, in your case it will return nullptr every time.

include/glow/Exporter/CommonOperatorWriter.h
```diff
@@ -0,0 +1,59 @@
...
+template <typename Traits> class CommonOperatorWriter : public ProtobufWriter {
+protected:
+  virtual ~CommonOperatorWriter() = default;
+
+  /// Declare pure virtual methods, one per each node kind.
+  /// Derived class must to implement all of it.
+#define DEF_NODE(CLASS, NAME)                                                  \
+  virtual llvm::Error write##NAME(                                             \
+      const Node *node, typename Traits::GraphProto &graph,                    \
+      typename Traits::NodeReporter &reporter) = 0;
```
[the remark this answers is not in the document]
putivsky: When rewinding happening, we don't want to visit or export those rewinded nodes again.

putivsky:
@jfix71, thanks a lot for the review, working on your comments...

lib/Exporter/ONNXModelWriter.cpp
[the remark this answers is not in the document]
putivsky: Changed upon request.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1031 @@
...
+DEF_UNSUPPORTED_NODE(CrossEntropyLoss)
+DEF_UNSUPPORTED_NODE(Convolution3DGrad)
+DEF_UNSUPPORTED_NODE(FullyConnectedGrad)
+DEF_UNSUPPORTED_NODE(CrossEntropyLossGrad)
+DEF_UNSUPPORTED_NODE(BatchNormalizationGrad)
+DEF_UNSUPPORTED_NODE(SparseLengthsWeightedSumGrad)
+DEF_UNSUPPORTED_NODE(SigmoidCrossEntropyWithLogits)
+DEF_UNSUPPORTED_NODE(LocalResponseNormalizationGrad)
+
+#if GLOW_WITH_OPENCL == 1
```
SplitInfinity: I added a new OpenCL-only node type yesterday called `OCLBatchedReduceAdd` so you'll have to rebase and add another small function below to write that as well.

putivsky: Yeah, have seen it, working on it.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+/// Writes all inputs from Node \p node to ptotobuf \p proto.
+void inputsToProto(const Node *node, ONNX_NAMESPACE::NodeProto *proto) {
+  for (unsigned b = 0, e = node->getNumInputs(); b < e; ++b) {
+    proto->add_input(node->getNthInput(b).getNode()->getName());
+  }
+}
+
+/// Writes MatMul operators from Node \p node into
+/// provided graph protobuf \p graph, optionally reports intermediate nodes as
+/// visited, signaling that such nodes must be ignored.
```
jfix71: nit: missing `\returns`

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    proto->add_input(RHS->getName());
+  }
+
+  outputsToProto(node, proto);
+  return llvm::Error::success();
+}
+
+/// Writes Arithmetic operators with name \p opName from Node \p node into
+/// provided graph protobuf \p graph, optionally reports intermediate nodes as
+/// visited, signaling that such nodes must be ignored.
```
jfix71: nit: missing `\returns`

tests/unittests/OnnxExporterTest.cpp
```diff
@@ -0,0 +1,65 @@
...
+  std::error_code code;
+  llvm::sys::fs::directory_iterator dirIt(inputDirectory, code);
+  for (llvm::sys::fs::directory_iterator dirIt(inputDirectory, code);
+       !code && dirIt != llvm::sys::fs::directory_iterator();
+       dirIt.increment(code)) {
+    auto name = dirIt->path();
+    if (name == "tests/models/onnxModels/preluInvalidBroadcastSlope.onnxtxt") {
+      continue;
+    }
+    testLoadAndSaveONNXModel(dirIt->path());
```
jfix71: Would it be possible to do something like load a proto, export it, load it back in, and run it? You could check the results are the same as if you just load/run it.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+template <typename T>
+llvm::Error writeMatMulKind(const T *node, ONNX_TRAITS::GraphProto &graph,
+                            ONNX_TRAITS::NodeReporter &reporter) {
+  auto *proto = graph.add_node();
+  proto->set_name(node->getName());
+  proto->set_op_type("MatMul");
+
+  // Check if LHS/RHS are Transpose node
```
jfix71: nit: Period.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+}
+
+} // namespace
+
+ONNXModelWriter::ONNXModelWriter(const std::string &modelFilename, Function &F,
+                                 size_t irVersion, size_t opsetVersion,
+                                 llvm::Error *errPtr, bool textMode)
+    : CommonOperatorWriter(modelFilename, F, errPtr),
+      opsetVersion_(opsetVersion) {
+  // if errPtr already contains an error then don't continue with constructor.
```
jfix71: nit: Capitalize sentence.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  proto->set_name(node->getName());
+  proto->set_op_type("PRelu");
+  proto->add_input(node->getInput().getNode()->getName());
+
+  const auto *slope = node->getSlope().getNode();
+  if (const auto *tile = llvm::dyn_cast<TileNode>(slope)) {
+    slope = unwindBroadcastInput(tile, reporter);
+  }
+  if (const SplatNode *SN = llvm::dyn_cast_or_null<SplatNode>(slope)) {
+    // Conversion a scalar to a tensor is required
```
jfix71: nit: Period.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+llvm::Error ONNXModelWriter::writeSplat(const SplatNode *node, GraphType &graph,
+                                        ReporterType &) {
+  auto *proto = graph.add_node();
+  // Find dictionary entires
+  addValueAttribute(proto, "value", node->getValue());
+
+  return writeAllWithNode(node->getName(), node, proto);
+}
+
+llvm::Error ONNXModelWriter::writeCPUMaxSplat(const CPUMaxSplatNode *node,
```
jfix71: Building with the CPU backend is also optional. Can we wrap the CPU specific nodes in `#ifdef GLOW_WITH_CPU` like we do for OpenCL?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  auto *tensorShape = tensorType->mutable_shape();
+  const auto &dims = glowType->dims();
+  for (unsigned b = 0, e = dims.size(); b < e; ++b) {
+    auto *tensorDims = tensorShape->add_dim();
+    tensorDims->set_dim_value(dims[b]);
+  }
+}
+
+bool ONNXModelWriter::isIgnoreNode(const Node *node) {
+  switch (node->getKind()) {
```
jfix71: Why do we ignore these? Is it because we expect to fold them into other Nodes, e.g. folding Broadcasted inputs into Arithmetic nodes? If yes, then won't we do this folding before we need to process the Tile or Splat Node, since we do post order traversal? It seems like we should be able to fall back to exporting these, no?

In case we do keep ignoring these, both of the Nodes here have logic defined for writing (e.g. `writeSplat()`/`writeTIle()`). If we are always going to ignore them then perhaps those functions should instead throw an Error, since we never expect to write them out. Also we should document exactly why we are ignoring these nodes.

putivsky: yes, you are right. Cause the graph navigation is done in post order all cases when Tile/Splat nodes had been added can be safely tracked, but explicit Tile/Splat nodes would be unaffected.

tests/unittests/OnnxExporterTest.cpp
putivsky: That was my plan for the next PR, because I would need to modify ONNXModelLoader to support Glow specific nodes first.

putivsky:
Made "reporter" as a class internal affairs.

PR #3169, opened 2019-06-25
[Partitioner] Add quantization profiling flow in Partitioner

### LINKED ISSUE #3112: Heterogeneous partitioning and quantization have issues when used together
I observe that an attempt to run an NN model on heterogeneous devices using a quantization profile produced before on a single device (e.g. CPU) does not always work as expected because after applying the heterogeneous partitioning some nodes are not found in the saved quantization profile and thus cannot be quantized.

beicy:
Summary:
This PR if for #3112  : if  "-dump-profile" is enabled, we won't do any partition. Instead, just generate the DAG (with only 1 node) and force the backend to be CPU. HostManager will force all DeviceManagers to be CPUDeviceManager and overwrite Provisioner and Executor.  Therefore, the network will be compiled and run under CPU backend. No change for "-load-profile" flow.

We run resnet50 for testing (see Test Plan)

Documentation:

[Optional Fixes #3112]
Test Plan:
Added test in ./tests/images/run.sh

The following example shows that dump/load quantization profile for resnet50. The profile is generated using CPU backend, while the quantized model in running with Interpreter backend. 
```
wangm-mbp:buildR wangm$  ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -interpreter-memory=20000 -num-devices=3 -dump-profile="partition_profile.yaml"
Model: resnet50
Running 1 thread(s).
I0711 12:45:56.583297 54956032 Partitioner.cpp:1221] Profiling a model to be partitioned cross different backends. Each sub-network will be optimized and run on cpu backend.
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5823)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9616)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9902)
wangm-mbp:buildR wangm$  ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -interpreter-memory=20000 -num-devices=3 -load-profile="partition_profile.yaml"
Model: resnet50
Running 1 thread(s).
I0711 12:46:11.744526 9793536 Partitioner.cpp:1346] The number of partitions is : 2, and the DAG is dumped into DAG.dot file.
I0711 12:46:11.745139 9793536 Partitioner.cpp:69] Writing dotty graph for DAG after graph partitioning: DAG.dot
I0711 12:46:11.745558 9793536 Partitioner.cpp:1355] 	 Partition 0:
		 Name :	resnet50_part1
		 BackendKind :	Interpreter
		 Memory :	18233280
		 LogicalDeviceIDs :	0
I0711 12:46:11.745569 9793536 Partitioner.cpp:1355] 	 Partition 1:
		 Name :	resnet50_part2
		 BackendKind :	Interpreter
		 Memory :	10703512
		 LogicalDeviceIDs :	1
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5565)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9551)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9890)
```

This is for heterogeneous partition testing (using this config file : tests/runtime_test/heterogeneousConfigs.yaml) :
```
wangm-mbp:buildR wangm$ ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -load-device-configs="tests/runtime_test/heterogeneousConfigs.yaml" -dump-profile="quantiP.yaml" 
Model: resnet50
Running 1 thread(s).
tests/runtime_test/heterogeneousConfigs.yaml
I0710 10:13:13.977905 142221312 Partitioner.cpp:1206] Profiling a model to be partitioned cross different backends. Each sub-network will be optimized and run on cpu backend.
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5823)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9616)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9902)
wangm-mbp:buildR wangm$ ./bin/image-classifier tests/images/imagenet/*.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -load-device-configs="tests/runtime_test/heterogeneousConfigs.yaml" -load-profile="quantiP.yaml" 
Model: resnet50
Running 1 thread(s).
tests/runtime_test/heterogeneousConfigs.yaml
I0710 10:13:32.141547 184852480 Partitioner.cpp:1330] The number of partitions is : 3, and the DAG is dumped into DAG.dot file.
I0710 10:13:32.142370 184852480 Partitioner.cpp:69] Writing dotty graph for DAG after graph partitioning: DAG.dot
I0710 10:13:32.142781 184852480 Partitioner.cpp:1339] 	 Partition 0:
		 Name :	resnet50_part1_part1
		 BackendKind :	CPU
		 Memory :	26571712
		 LogicalDeviceIDs :	0
I0710 10:13:32.142792 184852480 Partitioner.cpp:1339] 	 Partition 1:
		 Name :	resnet50_part2_part1
		 BackendKind :	Interpreter
		 Memory :	1228800
		 LogicalDeviceIDs :	1
I0710 10:13:32.142815 184852480 Partitioner.cpp:1339] 	 Partition 2:
		 Name :	resnet50_part3_part1
		 BackendKind :	CPU
		 Memory :	2088600
		 LogicalDeviceIDs :	0
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5676)
 File: tests/images/imagenet/dog_207.png	Label-K1: 207 (probability: 0.9563)
 File: tests/images/imagenet/zebra_340.png	Label-K1: 340 (probability: 0.9893)
```

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -147,6 +147,13 @@ public:
   static RegisterFactory<std::string, FactoryName, Backend>                    \
       FactoryName##_REGISTERED;
 
+/// The backend name used in Glow quantization profiling.
+#ifdef GLOW_WITH_CPU
+constexpr const char *profilingBackend = "CPU";
+#else
+constexpr const char *profilingBackend = "Interpreter";
+#endif
+
 } // namespace glow
 
 #endif // GLOW_BACKENDS_BACKEND_H
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -256,7 +256,8 @@ class Partitioner {
   void saturateHost(unsigned logicalDeviceCount);
 
   FunctionToBackendNameMap
-  backendBasedPartition(Function *F, std::vector<Backend *> &backends);
+  backendBasedPartition(Function *F, std::vector<Backend *> &backends,
+                        CompilationContext &cctx);
 
   /// Performs a load balancing optimization pass to optimize for load
   /// balance in addition to respecting memory constraints.
@@ -309,6 +310,14 @@ public:
   /// among different type of devices. \p cctx is used during optimization of
   /// the Function. \returns whether there was an error encountered.
   llvm::Error Partition(CompilationContext &cctx);
+
+  /// This partition approach is used in Glow Quantization Profiling flow. The
+  /// backendBasedPartition is applied first in case there are heterogeneous
+  /// backends. Then each sub-function will be compiled and run in CPU backend
+  /// for profiling.
+  llvm::Error QuantizationProfilingPartition(CompilationContext &cctx,
+                                             Function *F,
+                                             std::vector<Backend *> backends);
 
   /// Get the partitions.
   DAGListTy &getPartitionResult() { return partitions_; }
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -901,9 +901,8 @@ void Partitioner::doPartitioning(llvm::StringRef funcName,
   }
 }
 
-FunctionToBackendNameMap
-Partitioner::backendBasedPartition(Function *F,
-                                   std::vector<Backend *> &backends) {
+FunctionToBackendNameMap Partitioner::backendBasedPartition(
+    Function *F, std::vector<Backend *> &backends, CompilationContext &cctx) {
   FunctionToBackendNameMap ret;
   NodeToFunctionMap mapping;
   llvm::DenseMap<Node *, std::string> nodeToBackendName;
@@ -953,8 +952,15 @@ Partitioner::backendBasedPartition(Function *F,
   newF = F->getParent()->createFunction(std::string(F->getName()) + "_part" +
                                         std::to_string(++color));
   auto backendName = nodeToBackendName[bfs[level - 1][0]];
-  mapping.createPartition(newF, backendName);
-  ret[newF] = backendName;
+  if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
+    // When profiling, all the partition backend is assigned to
+    // profilingBackend.
+    mapping.createPartition(newF, profilingBackend);
+    ret[newF] = profilingBackend;
+  } else {
+    mapping.createPartition(newF, backendName);
+    ret[newF] = backendName;
+  }
   for (int i = level - 1; i >= 0; i--) {
     for (size_t j = 0, e = bfs[i].size(); j < e; j++) {
       Node *N = bfs[i][j];
@@ -963,17 +969,35 @@ Partitioner::backendBasedPartition(Function *F,
         backendName = bk;
         newF = F->getParent()->createFunction(
             std::string(F->getName()) + "_part" + std::to_string(++color));
-        mapping.createPartition(newF, backendName);
-        ret[newF] = backendName;
+        if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
+          // When profiling, all the partition backend is assigned to be
+          // profilingBackend.
+          mapping.createPartition(newF, profilingBackend);
+          ret[newF] = profilingBackend;
+        } else {
+          mapping.createPartition(newF, backendName);
+          ret[newF] = backendName;
+        }
       }
       mapping.add(N, newF);
     }
   }
 
-  // Here we just need to split the function without generating DAG.
   std::vector<Function *> funcs;
   funcs.push_back(F);
-  doPartitioning(F->getName(), funcs, mapping, false);
+  // When profiling, the partition flow will be stopped after
+  // backendBasedPartition. Therefore, the DAG needs to be generated. Otherwise,
+  // no need to generate DAG.
+  bool genDAG = cctx.precisionConfig.quantMode == QuantizationMode::Profile
+                    ? true
+                    : false;
+  if (genDAG) {
+    DeviceIDTy logicalDeviceID = 0;
+    for (auto &func : mapping.getPartitions()) {
+      mapping.appendLogicalDeviceID(func, logicalDeviceID++);
+    }
+  }
+  doPartitioning(F->getName(), funcs, mapping, genDAG);
 
   return ret;
 }
@@ -1177,6 +1201,30 @@ llvm::Error Partitioner::loadBalancedPartitioning(Function *F,
   return llvm::Error::success();
 }
 
+llvm::Error Partitioner::QuantizationProfilingPartition(
+    CompilationContext &cctx, Function *F, std::vector<Backend *> backends) {
+  // Quantization profiling flow is run under CPU backend, so we don't really
+  // need the concrete partition. The backendBasedPartition is necessary since
+  // we need the mapping between quantized tensor and original tensor.
+  FunctionToBackendNameMap funcToBackend;
+  funcToBackend = backendBasedPartition(F_, backends, cctx);
+  module_->eraseFunction(F_);
+  auto backend = createBackend(profilingBackend);
+  for (Function *subF : module_->getFunctions()) {
+    (void)subF;
+    assert(subF->verify() && "Conversion led to invalid function");
+    if (!optimized_) {
+      RETURN_IF_ERR(::glow::optimizeFunction(subF, *backend, cctx));
+    }
+  }
+  if (logPartition) {
+    LOG(INFO)
+        << "Profiling a model to be partitioned cross different backends. Each "
+           "sub-network will be optimized and run on cpu backend.\n";
+  }
+  return llvm::Error::success();
+}
+
 llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   // Prepare the mapping between BackendName and BackendInfo.
   std::vector<Backend *> backends;
@@ -1186,6 +1234,12 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   // Step 0: Find the representative function for running partitioning
   // algorithm.
   F_ = selectRepFunc(module_, memSize_);
+
+  if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
+    // Jump into profiling flow, and leave without generating partitions for the
+    // backends with same type..
+    return QuantizationProfilingPartition(cctx, F_, backends);
+  }
 
   // Step 1 : do the partition based on backends type.
   FunctionToBackendNameMap funcToBackend;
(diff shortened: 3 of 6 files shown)


beicy:
> What do you think about having two partitioning schemas (two partitioner classes) at high level: regular partitioner and profile capturing partitioner.
> 
> You'd instantiate HostManager with the right partitioner (pass required partitioner as a param in HostManager ctor) and remove all custom logic inside the partitioner. I have not thought about details, but in this case there supposed to be a cleaner implementation without too much if/elses inside the exact partitioner. I becomes increasingly harder to maintain all the custom logic inside this partitioner.

Thanks for your advice. Maybe not tow classes, but 2 partition methods? I will think about it.

lib/Partitioner/Partitioner.cpp
```diff
@@ -948,6 +958,24 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+      // For quantization profiling, no need to do partitioning at this time. No
+      // matter what type of backend it is, we use CPU backend for profiling.
+      auto backendName = "CPU";
+      funcToBackend[F_] = backendName;
+      if (backendMap_.find(backendName) == backendMap_.end()) {
+        // If "CPU" backend is not in current backendMap_, add it. This is used
+        // in createDAGWithoutPartition.
+        BackendInfo backendInfo;
+        backendInfo.backend = createBackend(backendName);
+        backendMap_[backendName] = backendInfo;
```
jfix71: So you are adding logic for if we're in profiling mode in the case where we have 1 backend (here) vs. not (below). I think if we're in profiling mode we could just always set the backend to the single backend we're profiling on (e.g. CPU like here), no? Then we wouldn't need to add different cases here?

E.g. above where we call `getBackendMap()`, instead we could check if we're in profiling mode -- if not then call `getBackendMap()` as is currently done, otherwise set up backends with e.g. the CPU backend like you do, and that's all that's needed, no?

beicy: You are right! Thanks :)

lib/Partitioner/Partitioner.cpp
```diff
@@ -935,6 +945,57 @@ llvm::Error Partitioner::createDAGWithoutPartition(
   return llvm::Error::success();
 }
 
+llvm::Error
+Partitioner::QuantizationProfilingPartition(CompilationContext &cctx) {
```
opti-mix: This function seems to be very similar to other functions in the partitioner. Is it possible to have some code-reuse?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -85,6 +85,22 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
                           name);
     }
   }
+
+  if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
```
opti-mix: Is it possible to tweak this outside of the host manager? May be in the client of the host-manager like image-classifier or something?

The reason I ask is that I'm a bit worried about making HostManager aware of quantization.

beicy: "Is it possible to tweak this outside of the host manager? May be in the client of the host-manager like image-classifier or something?" 
-- It is Host Manager calls the different partition method, so I don't think it is a good way to make this decision outside. 
But I think we can let the Partitioner to make this decision. Something like if we would like to skip the backendBasedPartition or not.

opti-mix: > But I think we can let the Partitioner to make this decision. Something like if we would like to skip the backendBasedPartition or not.

Sounds reasonable. This would be better than having both HostManager and Partitioner being aware of it.

opti-mix:
@beicy Ping! When do you plan to continue working on this PR?

beicy:
> @beicy Ping! When do you plan to continue working on this PR?

Updated. Now add testing for heterogenous partition.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1177,6 +1188,28 @@ llvm::Error Partitioner::loadBalancedPartitioning(Function *F,
   return llvm::Error::success();
 }
 
+llvm::Error Partitioner::QuantizationProfilingPartition(
+    CompilationContext &cctx, Function *F, std::vector<Backend *> backends) {
+  FunctionToBackendNameMap funcToBackend;
+  funcToBackend = backendBasedPartition(F_, backends, cctx);
+  module_->eraseFunction(F_);
+  auto backend = createBackend("CPU");
+  // Leave the partitioner early due to quantization profiling mode.
```
opti-mix: Is it still a correct comment? Where is the early exit? Is it this `RETURN_IF_ERR ` below?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -102,17 +103,20 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
...
-    // Check that all functions were not partitioned.
-    for (auto &network : nodeList) {
-      if (network.nodes.size() > 1) {
-        return MAKE_ERR(
-            GlowErr::ErrorCode::RUNTIME_ERROR,
-            "Failed to add network for profiling: Network was "
-            "partitioned, this is likely because the network was "
-            "larger than the configured memory of a single device manager.");
-      }
+    // For profiling, we use CPU backend. Overwrite Provisioner and Executor to
```
opti-mix: I think we discussed in previous review rounds that it would be nice to have all the quantization specific logic in one place. And you suggested to use `Partitioner` for it. Or do we need to do this in the host manager because there are no easy ways to hide it all in the `Partitioner`?

beicy: We leave the decision of which partition flow should be called to Partitioner. However, this is for Provisioner and Executor initialization, and this should be done in HM.

lib/Partitioner/Partitioner.cpp
beicy: Will move this comment to end of this function for better understanding.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -102,17 +103,20 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
...
+    // force the network is compiled and run in CPU backend.
+    size_t devicesNum = devices_.size();
+    for (size_t i = 0; i < devicesNum; i++) {
+      auto name = devices_[i]->getDeviceConfig().name;
+      auto config = llvm::make_unique<DeviceConfig>("CPU", name);
```
opti-mix: Nit: Can we define a named constexpr for the backend to be used for profiling and then use this named constexpr here and in the partitioner instead of hard-coding it?

lib/Partitioner/Partitioner.cpp
```diff
@@ -1187,6 +1220,11 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   // algorithm.
   F_ = selectRepFunc(module_, memSize_);
 
+  if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
+    // Jump into profiling flow.
```
opti-mix: You probably want the discussed comment about the early exit from the partitioner here.

beicy:
@opti-mix  Added test in tests/images/run.sh

include/glow/Backend/Backend.h
```diff
@@ -147,6 +147,9 @@ class BackendUsingGlowIR : public Backend {
   static RegisterFactory<std::string, FactoryName, Backend>                    \
       FactoryName##_REGISTERED;
 
+/// The backend name used in Glow quantization profiling.
+#define ProfilingBackend "CPU"
```
opti-mix: Could you rather use `constexpr char *profilngBackend = "CPU";`?

lib/Partitioner/Partitioner.cpp
```diff
@@ -963,17 +969,35 @@ Partitioner::backendBasedPartition(Function *F,
...
       mapping.add(N, newF);
     }
   }
 
-  // Here we just need to split the function without generating DAG.
   std::vector<Function *> funcs;
   funcs.push_back(F);
-  doPartitioning(F->getName(), funcs, mapping, false);
+  // When profiling, the partition flow will be stopped after
+  // backendBasedPartition. Therefore, the DAG need to be generated. Otherwise,
```
opti-mix: need -> needs

include/glow/Backend/Backend.h
jfix71: To ensure we don't break builds without the CPU backend enabled, can you make this something like:

```cpp
#ifdef GLOW_WITH_CPU
constexpr char *profilngBackend = "CPU";
#else
constexpr char *profilngBackend = "Interpreter";
#endif
```

include/glow/Partitioner/Partitioner.h
```diff
@@ -310,6 +311,14 @@ class Partitioner {
   /// the Function. \returns whether there was an error encountered.
   llvm::Error Partition(CompilationContext &cctx);
 
+  /// This partition approch is used in Glow Quantization Profiling flow. The
```
jfix71: typo: "approch" -> "approach"

tools/loader/Loader.cpp
```diff
@@ -406,6 +406,9 @@ void Loader::compile(PlaceholderBindings &bindings) {
     // Emit IR for the graph and compile it.
     auto error = hostManager_->addNetwork(std::move(M_), cctx);
     EXIT_ON_ERR(std::move(error));
+    // After partitioning, the original function may be removed. Need to update
+    // F_.
+    F_ = module->getFunctions().front();
```
jfix71: I'm curious, what do we use `F_` for after this point? It seems weird that we would just reset `F_` to the first function in the partition.

beicy: We will use F_ to get the module and then get all the functions in this module for generating quantization info in Loader::generateAndSerializeQuantizationInfos. When the fucntion Loader::generateAndSerializeQuantizationInfos is called, the M_ has been removed already. So we need to use a function in M_ to reach all functions in M_.

jfix71: Got it -- but we already have `M_` in the Loader, so could we just do `getModule()`?

beicy: getModule() here is { return F_->getParent();}  So we still need the updated F_  (the original F_ maybe erased from this module after partitioning) .  Now the code in  generateAndSerializeQuantizationInfos is
```
 for (auto F : getModule()->getFunctions()) {
    std::vector<NodeQuantizationInfo> tmp =
        quantization::generateNodeQuantizationInfos(bindings, F, loweredMap_,
                                                    quantizationSchema,
                                                    quantizationPrecision);
    QI.insert(QI.end(), tmp.begin(), tmp.end());
  }
```

PR #3145, opened 2019-06-19
[Operator] ONNX Where operator based on select glow node.

ayermolo:
Summary: Attempt 2.0. Re-did on top of select. Multidirectional numpy like broadcast is just with broadcast nodes.

Documentation:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Where
[Optional Fixes #issue]

Test Plan:
UnitTests
Please see a detailed explanation of how to fill out the fields in the relevant sections in PULL_REQUEST.md.

PR #3170, opened 2019-06-25
Refactor of broadcast support from Loaders.

ayermolo:
Summary:
Came out of https://github.com/pytorch/glow/pull/3145.
After this in, can update that PR and do another for Less operator.
Decided to go with this approach to have one method that can support multiple Node types. Instead of current approach for underlying base nodes of one method per node.

Documentation:

[Optional Fixes #issue]

Test Plan:
Current Importer tests capture Arithmetic nodes.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -627,6 +627,35 @@ public:
   ARITHMETIC_FUN_DECL(Pow);
 #undef ARITHMETIC_FUN_DECL
 
+  std::vector<NodeValue>
+  broadcastInputs(int axis, const llvm::ArrayRef<NodeValue> inputs);
+
+  template <class T, class U>
+  using enable_if_same_t = std::enable_if<std::is_same<T, U>::value, U>;
+
+#define DECLARE_BROADCAST_NODE(NODE_NAME, NUM_INPUTS)                          \
+  template <class T, class... Args>                                            \
+  typename enable_if_same_t<T, NODE_NAME##Node>::type *                        \
+  createNodeWithBroadcast(const std::string &name, int axis,                   \
+                          Args &&... inputArgs) {                              \
+    constexpr size_t numInputs = sizeof...(Args);                              \
+    static_assert(numInputs == NUM_INPUTS,                                     \
+                  "Invalid input passed in to createNodeWithBroadcast.");      \
+    std::vector<NodeValue> inputs = broadcastInputs(axis, {inputArgs...});     \
+    return create##NODE_NAME(name, inputs[0].getType(), inputs[0], inputs[1]); \
+  }
+
+  /// Template function that creates a node and normalizes its input shapes
+  /// with the use of BroadCast nodes. If axis is -1, it calculates it
+  /// automatically.
+  DECLARE_BROADCAST_NODE(Mul, /* NUM_INPUTS */ 2)
+  DECLARE_BROADCAST_NODE(Div, /* NUM_INPUTS */ 2)
+  DECLARE_BROADCAST_NODE(Add, /* NUM_INPUTS */ 2)
+  DECLARE_BROADCAST_NODE(Sub, /* NUM_INPUTS */ 2)
+
+#undef DECLARE_BROADCAST_NODE
+#undef BROADCAST_FUNC_COMMON_CODE
+
   /// Create a node that produces an boolean output of the same shape as
   /// \p input in which each element indicates whether or not the corresponding
   /// element in \p input is NaN or not.
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -475,58 +475,57 @@ protected:
     bool broadcast;
     ASSIGN_VALUE_OR_RETURN_ERR(broadcast, getBroadcast(dict));
 
-    NodeValue finalIn0 = in0;
-    NodeValue finalIn1 = in1;
+    int axis = -1;
 
-    if (broadcast) {
-      // Broadcasting can be:
-      // - multidirectional (ONNX opset 7+), or
-      // - unidirectional (ONNX opset 1->6,  Caffe2).
-      if (hasMultidirectionalBroadcast(typeName)) {
-        // Compute the target shape that is a combination of the operand shapes.
-        std::vector<size_t> targetDim;
-        ASSIGN_VALUE_OR_RETURN_ERR(targetDim, computeMultidirectionalBroadcast(
-                                                  in0.dims(), in1.dims()));
-        // Sets the axis of each inputs so that the trailing-most dimensions of
-        // input tensors and the target shape are aligned.
-        int axis0 = targetDim.size() - in0.dims().size();
-        int axis1 = targetDim.size() - in1.dims().size();
-        finalIn0 = G_.createBroadcast(opName, in0, targetDim, axis0);
-        finalIn1 = G_.createBroadcast(opName, in1, targetDim, axis1);
+    // Broadcasting can be:
+    // - multidirectional (ONNX opset 7+), or
+    // - unidirectional (ONNX opset 1->6,  Caffe2).
+
+    // Unidirectional broadcasting consists of broadcasting the right operand
+    // (in1) so that it matches the shape of the left operand (in0).
+    if (broadcast && !hasMultidirectionalBroadcast(typeName)) {
+      // With unidirectional broadcasting, the 'axis' attribute specifies
+      // from how much the right operand shape must be 'shifted' right.
+      // - In Caffe2, the 'axis' attribute is optional. If not specified, axis
+      // must be automatically computed so that the trailing-most dimensions
+      // of in1 is aligned to the trailing-most dimension of in0.
+      // - In ONNX, the 'axis' attribute is mandatory. axis == -1 is
+      // equivalent to no axis specified in Caffe2.
+
+      if (dict.count("axis")) {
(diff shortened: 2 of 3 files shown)


lib/Graph/Graph.cpp
```diff
@@ -50,6 +50,80 @@ void loggingStorageDeletionInUserFunc(Storage *s) {
...
 
+/// Merge shape \p shape into \p mergeShape, following multidirectional
+/// broadcasting rules.
+static void
+mergeMultidirectionalBroadcastHelper(std::vector<size_t> &mergeShape,
+                                     llvm::ArrayRef<size_t> shape) {
+  size_t shift = mergeShape.size() - shape.size();
+  for (size_t i = 0; i < shape.size(); i++) {
+    if (shape[i] == 1) {
+      // just leave mergeShape[i] as it is.
```
jfix71: nit: Capitalize sentence 🙂

lib/Graph/Graph.cpp
```diff
@@ -50,6 +50,80 @@ void loggingStorageDeletionInUserFunc(Storage *s) {
 }
 } // namespace
 
+/// Merge shape \p shape into \p mergeShape, following multidirectional
+/// broadcasting rules.
+static void
+mergeMultidirectionalBroadcastHelper(std::vector<size_t> &mergeShape,
+                                     llvm::ArrayRef<size_t> shape) {
+  size_t shift = mergeShape.size() - shape.size();
+  for (size_t i = 0; i < shape.size(); i++) {
```
jfix71: nit: `for (size_t i = 0, e = shape.size(); i < e; i++) {`

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -475,58 +475,59 @@ class CommonOperatorLoader : public ProtobufLoader {
...
     Node *node = nullptr;
-    if (typeName == "Mul") {
-      node = G_.createMul(opName, finalIn0, finalIn1);
-    } else if (typeName == "Add") {
-      node = G_.createAdd(opName, finalIn0, finalIn1);
-    } else if (typeName == "Sub") {
-      node = G_.createSub(opName, finalIn0, finalIn1);
-    } else if (typeName == "Div") {
-      node = G_.createDiv(opName, finalIn0, finalIn1);
+    if (broadcast) {
```
jfix71: If we're broadcasting, why can't we just call `broadcastInputs(axis, {in0, in1});` and then set `in0` and `in1` from the returned vector? Then we don't need all of the complicated macro/template stuff from `Graph.h`, and we can also call `node = G_.createMul(opName, in0, in1);` etc. without special casing it from broadcast vs not. E.g.:

```cpp
if (broadcast) {
  auto broadcastedInputs = broadcastInputs(axis, {in0, in1});
  assert(broadcastedInputs.size() == 2 && "...");
  in0 = broadcastedInputs[0];
  in1 = broadcastedInputs[1];
}
if (typeName == "Mul") {
  node = G_.createMul(opName, in0, in1);
} else if (typeName == "Add") {
  node = G_.createAdd(opName, in0, in1);
} else if (typeName == "Sub") {
  node = G_.createSub(opName, in0, in1);
} else if (typeName == "Div") {
  node = G_.createDiv(opName, in0, in1);
} else {
  RETURN_ERR("Unsupported arithmetic typeName");
}
```

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -475,58 +475,59 @@ class CommonOperatorLoader : public ProtobufLoader {
...
-        int axis0 = targetDim.size() - in0.dims().size();
-        int axis1 = targetDim.size() - in1.dims().size();
-        finalIn0 = G_.createBroadcast(opName, in0, targetDim, axis0);
-        finalIn1 = G_.createBroadcast(opName, in1, targetDim, axis1);
-      }
+
       // Unidirectional broadcasting consists of broadcasting the right operand
       // (in1) so that it matches the shape of the left operand (in0).
-      else {
+      if (!hasMultidirectionalBroadcast(typeName)) {
```
jfix71: Seems we can merge this if with the one above, right? I.e.

```cpp
if (broadcast && !hasMultidirectionalBroadcast(typeName)) {
```

include/glow/Importer/CommonOperatorLoader.h
ayermolo: We then have to replicate this logic here in OperatorTests, in Less, and Where. VS this approach it's all encapsulated in one place "under the hood" in one templated method.

include/glow/Graph/Graph.h
```diff
@@ -627,6 +627,55 @@ class Function final : public Named {
   ARITHMETIC_FUN_DECL(Pow);
 #undef ARITHMETIC_FUN_DECL
 
+  std::vector<NodeValue>
+  broadcastInputs(int axis, const llvm::ArrayRef<NodeValue> inputs);
+
+  template <class T, class U>
+  using enable_if_same_t = std::enable_if<std::is_same<T, U>::value, U>;
+#define BROADCAST_FUNC_T(NODE_NAME_)                                           \
```
bertmaher: I like the `enable_if` trick here (IIRC you did something similar in the optimizer), but I'd really prefer not to hide it in a macro.  It makes the definitions a bit shorter but relatively more confusing.  (When I looked at the optimizer it took me some time to mentally unwind the macro).

ayermolo: Yep. :)
So replace with full version?
i.e?
template <class T, class... Args>
typename enable_if_same_t<T, MulNode>::type

What about?
BROADCAST_FUNC_COMMON_CODE

bertmaher: Sgtm!  I'm cool with BROADCAST_FUNC_COMMON_CODE since it's kind of boilerplate-y asserts rather than tricksy template magic :-).

include/glow/Graph/Graph.h
```diff
@@ -627,6 +627,56 @@ class Function final : public Named {
...
+#define BROADCAT_FUNC_COMMON_CODE(NUM_INPUTS)                                  \
+  constexpr size_t numInputs = sizeof...(Args);                                \
+  static_assert(numInputs == NUM_INPUTS,                                       \
+                "Invalid input passed in to commonCreateBroadcast.");          \
+  std::vector<NodeValue> inputs = broadcastInputs(axis, {inputArgs...});
+
+  /// Template function that creates a node and normalizes it's input shapes
+  /// with the use of BroadCast nodes. If axis is -1, it calculates it
+  /// automatically.
+  template <class T, class... Args>
```
jfix71: nit: These are all boilerplate for their respective Nodes -- IMHO it's cleaner to just add a single macro for them and then do e.g. `DECLARE_BROADCAST_ARITHMETIC(Mul)` etc. This is also consistent with what we do with other cases for Arithmetic nodes.

include/glow/Graph/Graph.h
```diff
@@ -627,6 +627,56 @@ class Function final : public Named {
...
+  template <class T, class U>
+  using enable_if_same_t = std::enable_if<std::is_same<T, U>::value, U>;
+
+#define BROADCAT_FUNC_COMMON_CODE(NUM_INPUTS)                                  \
+  constexpr size_t numInputs = sizeof...(Args);                                \
+  static_assert(numInputs == NUM_INPUTS,                                       \
+                "Invalid input passed in to commonCreateBroadcast.");          \
+  std::vector<NodeValue> inputs = broadcastInputs(axis, {inputArgs...});
+
+  /// Template function that creates a node and normalizes it's input shapes
```
jfix71: Grammar nit 🙂: "it's" -> "its"

include/glow/Graph/Graph.h
ayermolo: @jfix71 I had something like this originally. Similar to what I did in GraphOptimizer.cpp @bertmaher mentioned it was not clear. Although I lean more towards macro I don't have a strong opinion one way or the other on this. Both approaches have their plus and minuses.

bertmaher: @jfix71 if you prefer the macro that's fine.  I guess opinions are different here :-)

jfix71: Huh, I thought you wanted the macro with just the extra template magic removed. I meant a macro for the entire Arithmetic Node, e.g. something like:

```cpp
#define DECLARE_BROADCAST_ARITHMETIC(NODE_NAME)              \
  template <class T, class... Args>                          \
  typename enable_if_same_t<T, NODE_NAME##Node>::type *      \
  createNodeWithBroadcast(const std::string &name, int axis, \
                          Args &&... inputArgs) {            \
    BROADCAST_FUNC_COMMON_CODE(/* NUM_INPUTS */ 2)           \
    return create##NODE_NAME(name, inputs[0].getType(),      \
                             inputs[0], inputs[1]);          \
  }
DECLARE_BROADCAST_ARITHMETIC(Mul)
DECLARE_BROADCAST_ARITHMETIC(Div)
DECLARE_BROADCAST_ARITHMETIC(Add)
DECLARE_BROADCAST_ARITHMETIC(Sub)
```
I'm fine either way though 🙂

ayermolo: Ah. I see. Yeah this looks good. I'll change to this. :)

ayermolo: Modified slightly to make it more generic to support Where/Select down the road. Probably Less/cmpLT.

ayermolo:
NP, thanks for feedback.

PR #3173, opened 2019-06-25
Improve error message when synEnqueue fails

bertmaher:
Summary:
The current error message simply gives a line number and the raw
status code; this diff decodes the status into a string and reports the failing
function.

There are probably other places that could use this treatment.  Will follow up
if this is accepted.

Differential Revision: D15997655

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "Habana.h"
+#include "HabanaUtils.h"
 
 #include "glow/Graph/PlaceholderBindings.h"
 #include "glow/IR/IR.h"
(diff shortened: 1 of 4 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -544,7 +545,7 @@ llvm::Error HabanaFunction::execute(ExecutionContext *context) {
       inputInfo.size(), outputInfo.data(), outputInfo.size(), &handle);
   seEvent.addArg("result", std::to_string(res));
   if (res != synSuccess) {
-    return chk_make_err(res);
+    return MAKE_ERR(strFormat("synEnqueueByName failed: %s", statusStr(res)));
```
rdzhabarov: maybe also include to the trace metadata above (seems useful)

lib/Backends/Habana/HabanaUtils.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "HabanaUtils.h"
+
+namespace glow {
+const char *statusStr(synStatus status) {
+  const char *strs[] = {
```
rdzhabarov: static storage here?

lib/Backends/Habana/HabanaUtils.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "HabanaUtils.h"
+
+namespace glow {
+const char *statusStr(synStatus status) {
```
rdzhabarov: for safety can we handle synStatus out of this range as well?

bertmaher:
Yeah, I'm pulling these from the enum definition in synapse_types.h (they'd need to be kept in sync, of course).  I didn't see a decoder function anywhere in the existing library although I'm sure one could be provided.

PR #3174, opened 2019-06-25
[compilationLog]: Log constant creation/deletion in module level.

ZchiPitt:
Summary:
Log constants creation/deletion in module level, such that:
- Implement a ModuleLogContext class that stores all module related changes(constants/placeholders changes) 
- When Module creates another function, all existing contents stored in ModuleLogContext will also be copied into Function's LogContext. Such that when this function uses a constant created from previous function, the log knows where this constant comes from.

Test Plan:
ninja check

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -61,6 +61,9 @@ class Module final {
   /// Deterministic PRNG used to initialize weights in this module.
   PseudoRNG PRNG_;
 
+  /// Module log context that stores all logs related to this module.
+  ModuleLogContext moduleLogCtx_;
+
 public:
   Module() = default;
 
@@ -199,6 +202,9 @@ public:
   /// \Returns the size in bytes of data used by constants.
   uint64_t getConstantsSize();
 
+  /// \Returns the module log context.
+  ModuleLogContext &getModuleLogContext() { return moduleLogCtx_; };
+
   // Don't copy or move this class around.
   // The destructor will wipe the functions leaving
   // the original Module only dangling pointers.
@@ -224,16 +230,19 @@ class Function final : public Named {
   Module *parent_;
 
   /// The log context associated with this function.
-  LogContext logCtx_;
+  std::shared_ptr<LogContext> logCtx_;
 
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {}
+      : Named(Name), parent_(parent) {
+    logCtx_ = std::make_shared<LogContext>();
+    logCtx_->setParent(this);
+  }
 
   ~Function();
 
   /// Return the log context.
-  LogContext &getLogContext() { return logCtx_; }
+  std::shared_ptr<LogContext> getLogContext() { return logCtx_; }
 
   /// Add placeholder for metadata such as profiling.
   void addMetadataPlaceholder(Placeholder *PH) {
@@ -274,7 +283,7 @@ public:
     nodes_.push_back(N);
 
     // Log the node creation.
-    logCtx_.logNodeCreation(N);
+    logCtx_->logNodeCreation(*N);
 
     return N;
   }
--- a/include/glow/Graph/Log.h
+++ b/include/glow/Graph/Log.h
@@ -24,14 +24,31 @@
 
 namespace glow {
 
+class Function;
+
 class Node;
 
 struct NodeValue;
 
-/// A class for logging all compilation related activities.
-class LogContext {
+/// A class for logging all module related logs.
+class ModuleLogContext {
 private:
-  /// A vector that keeps track of all log contents.
+  /// A vector that keeps track of all constant/placeholder changes of this
+  /// module.
+  std::vector<std::string> moduleLogContents_;
+
+public:
+  /// Add content into the contents vector.
+  void addModuleLogContent(llvm::StringRef logContent);
+
+  /// Getter that returns the module log contents.
+  std::vector<std::string> &getModuleLog() { return moduleLogContents_; }
+};
+
+/// A class for logging all compilation related activities.
+class LogContext final {
+private:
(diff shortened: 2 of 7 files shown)


include/glow/Graph/Log.h
```diff
@@ -84,7 +109,7 @@ class LogContext {
 class ScopedLogBlock {
 
   /// Reference to the log context.
-  LogContext &ctx_;
+  LogContext *ctx_;
```
opti-mix: & instead of *?

include/glow/Graph/Graph.h
```diff
@@ -224,16 +230,22 @@ class Function final : public Named {
   Module *parent_;
 
   /// The log context associated with this function.
-  LogContext logCtx_;
+  LogContext *logCtx_;
```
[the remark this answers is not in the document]
ZchiPitt: set to nullptr

include/glow/Graph/Graph.h
```diff
@@ -224,16 +230,22 @@ class Function final : public Named {
...
 
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {}
+      : Named(Name), parent_(parent) {
+    if (parent_) {
+      logCtx_ = new LogContext(this, parent_->getModuleLogContext());
```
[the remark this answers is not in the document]
ZchiPitt: add deletion in destructor

include/glow/Graph/Graph.h
```diff
@@ -224,16 +230,18 @@ class Function final : public Named {
   Module *parent_;
 
   /// The log context associated with this function.
-  LogContext logCtx_;
+  LogContext *logCtx_ = nullptr;
```
jfix71: What's the motivation of changing this to a pointer? Seems like the Function always owns it (new/delete contained in constructor/destructor).

include/glow/Graph/Log.h
```diff
@@ -40,8 +57,14 @@ class LogContext {
   /// A string that represents the current full scope name.
   std::string currentFullScope_;
 
+  /// A pointer to the function that has the LogContext.
```
jfix71: nit: reference, not pointer

include/glow/Graph/Log.h
```diff
@@ -40,8 +57,14 @@ class LogContext {
...
+  Function &parent_;
+
+  /// A pointer to the module's log context.
```
jfix71: nit: reference, not pointer

include/glow/Graph/Log.h
```diff
@@ -58,11 +81,13 @@ class LogContext {
   /// Dumps the log into a file named after the given \p funcName.
   void dumpLog(llvm::StringRef funcName);
 
-  /// Logs the node creation with a list of input nodes.
-  void logNodeCreation(const Node *newNode);
+  /// Logs the node creation. Also logs into Module log context if \p
+  /// logIntoModule set as true.
+  void logNodeCreation(const Node *newNode, bool logIntoModule = false);
```
jfix71: nit: You use a `Node *` here, but a `Node &` in `logNodeDeletion()`. Would be good to make consistent, probably making both a reference if possible.

lib/Graph/Graph.cpp
```diff
@@ -38,14 +38,18 @@ using llvm::isa;
 
 namespace {
 /// A helper function to log the deletion of constant/placeholder \p s of a
-/// module.
-void loggingStorageDeletionInUserFunc(Storage *s) {
-  std::unordered_set<Function *> userFuncs;
-  for (auto &use : s->getUsers()) {
-    userFuncs.insert(use.getUser()->getParent());
+/// module into the log context of given functions \p functions.
```
jfix71: Can you add a note explaining why we can't do it the way we were before, i.e. because all users have already been removed hence why we can remove it in the first place.

lib/Graph/Log.cpp
```diff
@@ -48,6 +53,21 @@ void LogContext::addLogMetaData() {
...
 }
 
+LogContext::LogContext(Function &parent, ModuleLogContext &moduleContext)
+    : parent_(parent), moduleContext_(moduleContext) {
+  addLogMetaData();
+
+  // Add the logs of module constants/placeholders into the log context of this
+  // function. Such that the Log Context has the information of any
+  // constants/placeholders created in previous functions.
+  for (auto &s : moduleContext.getModuleLog()) {
```
SplitInfinity: Can this be `const` since `addLogContext` accepts a `StringRef` argument?

lib/Graph/Graph.cpp
```diff
@@ -38,14 +38,18 @@ using llvm::isa;
...
+void loggingStorageDeletion(std::list<Function *> functions, Storage *s) {
+  for (auto *F : functions) {
+    F->getLogContext().logNodeDeletion(*s, /* logIntoModule */ true);
```
SplitInfinity: Won't this log the deletion into the module log once for every function? Is that intentional?

ZchiPitt: thx for reminding. This is not intentional. I just fixed that logging into module will only happen once.

include/glow/Graph/Graph.h
```diff
@@ -224,16 +230,19 @@ class Function final : public Named {
...
   /// The log context associated with this function.
-  LogContext logCtx_;
+  std::shared_ptr<LogContext> logCtx_;
 
 public:
   Function(Module *parent, llvm::StringRef Name = {})
-      : Named(Name), parent_(parent) {}
+      : Named(Name), parent_(parent) {
+    logCtx_ = std::make_shared<LogContext>();
+    logCtx_.get()->setParent(this);
```
nickgg: you dont need the call to get() here. You can just use the -> on a shared_ptr.

Similar comments on a lot of cals below.

PR #3176, opened 2019-06-26
[OpenCL] Add pool of reusable command queues

SplitInfinity:
**Description**
The creation of OpenCL command queues can take varying amounts of time
depending on the implementation. This commit adds
`OpenCLCommandQueuePool`, a class that manages a pool of command queues
separated by their queue properties and reuses queues when possible
instead of allocating a new one (which it does when all else fails).

**Test Plan**
Manual examination of traces, new unit tests.

DLRM training trace before this diff:
Screen Shot 2019-06-25 at 6 53 59 PM

DLRM training trace after this diff:
Screen Shot 2019-06-25 at 6 53 20 PM

Each minibatch (after the first one) takes around 6 ms instead of 17 ms (65% reduction).

### DIFF
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -64,6 +64,77 @@ static llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
   return parsed;
 }
 
+OpenCLCommandQueuePool::~OpenCLCommandQueuePool() {
+  // Make sure all queues have been returned to the pool.
+  DCHECK_EQ(queuesAllocated_, queuesAvailable_)
+      << "OpenCLCommandQueue destroyed before all queues returned!";
+  // For each properties -> vector of queues pair:
+  for (auto &kv : queues_) {
+    // For each queue in each vector:
+    for (auto &q : kv.second) {
+      // Release the backing queue.
+      cl_int err = clReleaseCommandQueue(q.backingQueue);
+      DCHECK_EQ(err, CL_SUCCESS)
+          << "clReleaseCommandQueue failed with error code " << err;
+    }
+  }
+}
+
+llvm::Expected<OpenCLCommandQueue> OpenCLCommandQueuePool::requestCommandQueue(
+    cl_command_queue_properties properties) {
+  OpenCLCommandQueue ret;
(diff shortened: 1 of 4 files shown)


SplitInfinity:
`TraceEventsTest` is failing, investigating now.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -64,6 +64,82 @@ static llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
...
+      << "OpenCLCommandQueue destroyed before all queues returned!";
+  // For each properties -> vector of queues pair:
+  for (auto &kv : queues_) {
+    // For each queue in each vector:
+    for (auto &q : kv.second) {
+      // Release the backing queue.
+      cl_int err = clReleaseCommandQueue(q.backingQueue);
+      DCHECK_EQ(err, CL_SUCCESS)
+          << "clReleaseCommandQueue failed with error code " << err;
+      // Adjust counters.
```
nickgg: The counters are destroyed at the exit of this function, do you need to adjust them?

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -64,6 +64,82 @@ static llvm::Expected<unsigned> parseInputAsUnsigned(std::string input) {
...
+    --queuesAvailable_;
+    --queuesAvailableByProps_[properties];
+  }
+
+  return ret;
+}
+
+void OpenCLCommandQueuePool::returnCommandQueue(OpenCLCommandQueue &queue) {
+  // Check that the number of available queues is less than the number of
+  // allocated queues for the properties of the given queue.
```
nickgg: Is this comment correct? Doesn't appear to use the properties.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -340,20 +414,30 @@ void OpenCLDeviceManager::runFunctionImpl(
 
   CompiledFunction *func = funcIt->second;
 
+  TRACE_EVENT_SCOPE(context->getTraceContext(), TraceEvent::STANDARD,
```
nickgg: Oh this is why your test is failing, this should be `TraceEvent::RUNTIME`

PR #3177, opened 2019-06-26
Bug fixes to partitioner; some restructuring

nrsatish:
Summary: Bug fixes to performance model in partitioning. Restructure to have DeviceInfo returned by DeviceManagers.

Details:
The changes made are: 
(1) Add compute costs for more kinds of nodes and Habana specific backend nodes
(2) Nits to remove doxygen style comments (from phabricator comments)
(3) Add a getDeviceInfo() function to enable onnxifi path for calling into the roofline.
Differential Revision: D15999942

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -50,6 +50,17 @@ bool CPUDeviceManager::isMemoryAvailable(uint64_t estimate) const {
   return maxMemoryBytes_ >= (usedMemoryBytes_ + estimate);
 }
 
+DeviceInfo CPUDeviceManager::getDeviceInfo() const {
+  // TODO: these may need to be tweaked depending on specific CPU.
+  DeviceInfo info = DeviceInfo();
+  info.sramCapacity = 256 * 1024 * 1024;
+  info.peakCompute = 2.2 * 1024 * 1024 * 1024 * 1024;
+  info.peakDramBw = 110.0 * 1024 * 1024 * 1024;
+  info.peakSramBw = 1024.0 * 1024 * 1024 * 1024;
+  info.peakPCIeBw = 16.0 * 1024 * 1024 * 1024;
+  return info;
+}
+
 void CPUDeviceManager::addNetworkImpl(const Module *module,
                                       FunctionMapTy functions,
                                       ReadyCBTy readyCB) {
--- a/lib/Backends/Habana/HabanaDeviceManager.h
+++ b/lib/Backends/Habana/HabanaDeviceManager.h
@@ -138,6 +138,10 @@ public:
   uint64_t getMaximumMemory() const override;
   uint64_t getAvailableMemory() const override;
   bool isMemoryAvailable(uint64_t estimate) const override;
+
+  /// Returns the DeviceInfo for this device containing peak limits for
+  /// compute and bandwidths (used in partitioning).
+  DeviceInfo getDeviceInfo() const override;
 };
 
 } // namespace runtime
--- a/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
+++ b/lib/Backends/Interpreter/InterpreterDeviceManager.cpp
@@ -51,6 +51,17 @@ bool InterpreterDeviceManager::isMemoryAvailable(uint64_t estimate) const {
   return maxMemoryBytes_ >= (usedMemoryBytes_ + estimate);
 }
 
+DeviceInfo InterpreterDeviceManager::getDeviceInfo() const {
+  // TODO: these may need to be tweaked depending on interpreter overheads.
+  DeviceInfo info = DeviceInfo();
+  info.sramCapacity = 256 * 1024 * 1024;
+  info.peakCompute = 2.2 * 1024 * 1024 * 1024 * 1024;
+  info.peakDramBw = 110.0 * 1024 * 1024 * 1024;
+  info.peakSramBw = 1024.0 * 1024 * 1024 * 1024;
+  info.peakPCIeBw = 16.0 * 1024 * 1024 * 1024;
+  return info;
+}
+
 void InterpreterDeviceManager::addNetworkImpl(const Module *module,
                                               FunctionMapTy functions,
                                               ReadyCBTy readyCB) {
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -235,8 +235,8 @@ void Partitioner::initOpComputeTime(Function *F) {
   float peakCompute = deviceInfo_[0].peakCompute;
 
   for (auto &node : F->getNodes()) {
-    /// compute memory side bytes for inputs from DRAM, SRAM.
-    /// TODO: think about whether this is better off computed inside a Node.
+    // compute memory side bytes for inputs from DRAM, SRAM.
+    // TODO: think about whether this is better off computed inside a Node.
 
     int n = node.getNumInputs();
     uint64_t sizeDram = 0;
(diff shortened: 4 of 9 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -328,6 +389,25 @@ void Partitioner::initOpComputeTime(Function *F) {
       totalOps *= (inputChannels * 1.0 / nGroups);
       break;
     }
+#ifdef GLOW_WITH_HABANA
+    case Kinded::Kind::HabanaConvolutionNodeKind: {
```
jackm321: This is copy pasted from above? Can it just share this logic?

lib/Partitioner/Partitioner.cpp
```diff
@@ -328,6 +389,25 @@ void Partitioner::initOpComputeTime(Function *F) {
       totalOps *= (inputChannels * 1.0 / nGroups);
       break;
     }
+#ifdef GLOW_WITH_HABANA
```
jackm321: Is this needed? HabanaConvolutionNodeKind will be defined regardless right?

lib/Partitioner/Partitioner.cpp
```diff
@@ -308,9 +360,18 @@ void Partitioner::initOpComputeTime(Function *F) {
       auto *FCN = llvm::dyn_cast<FullyConnectedNode>(&node);
       auto inputDims = FCN->getInput().dims();
       auto wtDims = FCN->getWeights().dims();
-      totalOps = 2 * inputDims[0] * inputDims[1] * wtDims[1];
+      totalOps = 2 * inputDims[0] * inputDims[1] * wtDims[0];
+      break;
+    }
+#ifdef GLOW_WITH_HABANA
+    case Kinded::Kind::HabanaFullyConnectedNodeKind: {
```
jackm321: Having backend-specific stuff in partitioner seems a bit entangled to me.
I see how this is backend-specific but it seems like we should factor this stuff out to maybe be part of the backend. We don't have have to do that in this pr but should look into it. Ideally most of these cases would be moved somewhere else I think.

include/glow/Backends/DummyDeviceManager.h
```diff
@@ -44,6 +44,9 @@ class DummyDeviceManager : public DeviceManager {
   }
   bool isMemoryAvailable(uint64_t) const override { return true; }
 
+  /// Returns uninitialized DeviceInfo
+  DeviceInfo getDeviceInfo() const override { return DeviceInfo(); }
```
jackm321: Would it be better to just make this the default implementation instead of having that be a pure virtual?

lib/Partitioner/Partitioner.cpp
nrsatish: Agreed in general. This does need to be refactored at a later point. But I would want that to be in a different PR.

lib/Partitioner/Partitioner.cpp
nrsatish: I’m not sure - I think I saw errors without this. But perhaps someone can tell if it’s possible to remove this in some other way?

lib/Partitioner/Partitioner.cpp
nrsatish: Yeah will look into that.

include/glow/Backends/DummyDeviceManager.h
nrsatish: Agreed, thanks for the suggestion.

lib/Partitioner/Partitioner.cpp
beicy: Agree with @jackm321 . In fact, in partitioner, the backend-specific info will comes from configure file (passed through device configs).  In addition, I think the "roofline" calculation can be moved from Partitioner.cpp and added in some support file or utility file.

beicy:
Could you please add more details about "Bug fixes to performance model in partitioning"?

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -50,6 +50,17 @@ bool CPUDeviceManager::isMemoryAvailable(uint64_t estimate) const {
   return maxMemoryBytes_ >= (usedMemoryBytes_ + estimate);
 }
 
+DeviceInfo CPUDeviceManager::getDeviceInfo() const {
+  // TODO: these may need to be tweaked depending on specific CPU.
+  DeviceInfo info = DeviceInfo();
+  info.sramCapacity = 256 * 1024 * 1024;
```
beicy: Should be hard coded? Can we get it from some config file?

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -51,6 +51,17 @@ bool InterpreterDeviceManager::isMemoryAvailable(uint64_t estimate) const {
   return maxMemoryBytes_ >= (usedMemoryBytes_ + estimate);
 }
 
+DeviceInfo InterpreterDeviceManager::getDeviceInfo() const {
+  // TODO: these may need to be tweaked depending on specific CPU
```
beicy: Could you please explain more about "specific CPU" here?  It is Interpreter backend.

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
```diff
@@ -51,6 +51,17 @@ bool InterpreterDeviceManager::isMemoryAvailable(uint64_t estimate) const {
...
+  DeviceInfo info = DeviceInfo();
+  info.sramCapacity = 256 * 1024 * 1024;
+  info.peakCompute = 2.2 * 1024 * 1024 * 1024;
```
beicy: Same here. If the number should be hard coded, could you please add some reference?

lib/Backends/Habana/HabanaDeviceManager.h
```diff
@@ -138,6 +138,10 @@ class HabanaDeviceManager : public DeviceManager {
   uint64_t getMaximumMemory() const override;
   uint64_t getAvailableMemory() const override;
   bool isMemoryAvailable(uint64_t estimate) const override;
+
+  /// Returns the DeviceInfo for this device containing peak limits for
+  /// compute and bandwidths (used in partitioning).
+  DeviceInfo getDeviceInfo() const override;
```
beicy: The DeviceConfigs is passed into HostManager for initialization. Basically, it contains all backend related information and then the DeviceManagers are created.  I don't think this function is necessary. The numbers used for roofline calculation should be added in to DeviceConfigs and passed to HostManager.

include/glow/Backends/DummyDeviceManager.h
beicy: The DeviceConfigs is passed into HostManager for initialization. Basically, it contains all backend related information and then the DeviceManagers are created. I don't think this function is necessary. The numbers used for roofline calculation should be added in to DeviceConfigs and passed to HostManager.

lib/Partitioner/Partitioner.cpp
bertmaher: I think HabanaConvolutionNode is only defined if `GLOW_WITH_HABANA` is enabled since the backend-specific parts of NodeGen are guarded by `ifdef`s and CMake variables.

lib/Partitioner/Partitioner.cpp
bertmaher: I agree that some refactoring is needed here eventually but I'd really like to move the partitioner diffs forward without a ton of redesign since they're important to understanding/optimizing perf on devices.  We can tidy this up as things become more stable.

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
bertmaher: I think this is a good idea but can we defer it?  We currently have no ability to pass configs via onnxifi anyways, which is where we most need these features.

beicy: Sure, it is OK to defer it.  Reading backend info from config file is one way, and another way is manually setting it before calling HostManager.  If the value here is kind of "fixed" for most of case, I think this function could be renamed as "getDefaultDeviceInfo()", or just set those values by default when the DeviceManager is created.

lib/Partitioner/Partitioner.cpp
jackm321: Sounds good

lib/Partitioner/Partitioner.cpp
```diff
@@ -258,17 +258,69 @@ void Partitioner::initOpComputeTime(Function *F) {
...
-        sizeSram += sizeInput;
-      }
-
+      /// tables are usually large and fit in DRAM
+      sizeDram += sizeInput;
       /// we also read the indices, weights and lengths arrays
       sizeSram += SLWSN->getIndices().getType()->getSizeInBytes();
       sizeSram += SLWSN->getWeights().getType()->getSizeInBytes();
       sizeSram += SLWSN->getLengths().getType()->getSizeInBytes();
+    } else if (node.getKind() == Kinded::Kind::SparseLengthsSumNodeKind) {
```
rdzhabarov: there is a lot of duplication in FusedRowwiseQuantizedSparseLengthsWeightedSumNodeKind and other weighted variations. Similar in non weighted. Can this be extracted into a template method to reuse the code.

rdzhabarov:
Totally agree with @jackm321 on refactoring of partitioner. That's something we also discussed with @nrsatish in internal diffs.

lib/Partitioner/Partitioner.cpp
nrsatish: I discussed offline; probably it makes most sense to couple this with the other restructuring changes in Partitioner.

include/glow/Backends/DummyDeviceManager.h
nrsatish: Addressed by removing pure virtual function.

lib/Backends/Interpreter/InterpreterDeviceManager.cpp
nrsatish: Added to summary

lib/Partitioner/Partitioner.cpp
nrsatish: Discussed offline and agreed to defer to the restructuring pass.

PR #3178, opened 2019-06-26
[Optimizer] Fix lint (references, curly braces, shadow variables, etc.)

jfix71:
Summary: Fix lint found in Optimizer. Mostly NFC.

Test Plan: Tests still pass.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -132,9 +132,9 @@ static bool isIdentityShuffle(llvm::ArrayRef<unsigned> shuffle) {
 /// \returns True if the node \p N always evaluates to \p val.
 bool isSplatOfVal(Node *N, float val) {
   SplatNode *Z = dyn_cast<SplatNode>(N);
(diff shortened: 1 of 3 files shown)


lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -1072,13 +1073,15 @@ bool MergeBatchedAdd::run(Function *F) {
     // We found a sequence of batched-add-slice that cover the input tensor.
     // We can transform the graph and create one big batched-add.
     std::vector<Node *> newSlices;
+    assert(order.size() > 1 && "order must contain at least 2 SliceNodes.");
```
rdzhabarov: nit: I think current preference is GLOG instead of old type asserts.

jfix71: Good call. We haven't updated GraphOptimizer yet though to use glog. Let's make all of the changes there at once.

rdzhabarov: sounds good.

PR #3179, opened 2019-06-26
[Operators] Support casts i32<->i64<->float<->float16 in Interpreter

Max-K-cadence:
The main purpose of that is constant-folding, that is why only the interpreter
backend is supported so far. Can be done for CPU backend separately if needed.

Test Plan:
Added unit tests.

### DIFF
--- a/lib/Backends/Interpreter/Interpreter.cpp
+++ b/lib/Backends/Interpreter/Interpreter.cpp
@@ -308,16 +308,21 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
   case Kinded::Kind::IntLookupTableNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind({ElemKind::Int8QTy});
 
-  case Kinded::Kind::ConvertToNodeKind:
-    return ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::FloatTy) &&
-            ((NI.getOutElemTy(ConvertToNode::ResultIdx) ==
-              ElemKind::Float16Ty) ||
-             (NI.getOutElemTy(ConvertToNode::ResultIdx) ==
-              ElemKind::Int64ITy))) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Float16Ty) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy)) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Int64ITy) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy));
+  case Kinded::Kind::ConvertToNodeKind: {
+    auto isConversionSupportedFor = [](ElemKind kind) {
+      switch (kind) {
+      case ElemKind::Float16Ty:
+      case ElemKind::FloatTy:
+      case ElemKind::Int32ITy:
+      case ElemKind::Int64ITy:
+        return true;
+      default:
+        return false;
+      }
+    };
+    return isConversionSupportedFor(NI.getInElemTy(ConvertToNode::InputIdx)) &&
+           isConversionSupportedFor(NI.getOutElemTy(ConvertToNode::ResultIdx));
+  }
 
   case Kinded::Kind::TopKNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -3117,35 +3117,31 @@ void BoundInterpreterFunction::fwdIntLookupTableInst(
 void BoundInterpreterFunction::fwdConvertToInst(const glow::ConvertToInst *I) {
   Tensor *source = getTensor(I->getInput());
   Tensor *dest = getTensor(I->getResult());
-  if (source->getType() == dest->getType()) {
+  auto srcElType = source->getType().getElementType();
+  auto destElType = dest->getType().getElementType();
+  if (srcElType == destElType) {
     // This is a noop conversion.
     dest->copyRawFrom(source);
     return;
   }
-  switch (source->getElementType()) {
-  case ElemKind::FloatTy:
-    switch (dest->getElementType()) {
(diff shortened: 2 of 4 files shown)


lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -309,15 +309,18 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
...
-             (NI.getOutElemTy(ConvertToNode::ResultIdx) ==
-              ElemKind::Int64ITy))) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Float16Ty) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy)) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Int64ITy) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy));
+    return (((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::FloatTy) ||
+             (NI.getInElemTy(ConvertToNode::InputIdx) ==
+              ElemKind::Float16Ty)) &&
+            ((NI.getInElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy) ||
```
rdzhabarov: Does not seem correct. ConvertToNode::ResultIdx - this is output of the node. NI.getInElemTy() checks for inputs.

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -309,15 +309,18 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
...
-    return ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::FloatTy) &&
-            ((NI.getOutElemTy(ConvertToNode::ResultIdx) ==
-              ElemKind::Float16Ty) ||
-             (NI.getOutElemTy(ConvertToNode::ResultIdx) ==
-              ElemKind::Int64ITy))) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Float16Ty) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy)) ||
-           ((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::Int64ITy) &&
-            (NI.getOutElemTy(ConvertToNode::ResultIdx) == ElemKind::FloatTy));
+    return (((NI.getInElemTy(ConvertToNode::InputIdx) == ElemKind::FloatTy) ||
```
rdzhabarov: I must admit this is a bit hard to read. Can you add a comment here explaining expectations, like what input types could be and what output types would be supported for those inputs.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1585,6 +1585,31 @@ TEST(onnx, importCastToInt64) {
...
+    bindings.allocate(mod.getPlaceholders());
+    updateInputPlaceholdersByName(bindings, &mod, {"input"}, {&data});
+  }
+
+  EE.compile(CompilationMode::Infer, F);
+  EE.run(bindings);
+  auto result = bindings.get(graphOutputVar)->getHandle();
+  std::vector<size_t> expectedDims = {1, 2, 4, 3};
+
+  EXPECT_TRUE(result.dims().vec() == expectedDims);
```
rdzhabarov: Might need to make testing better here (checking values, etc). Checking just dims seems insufficient.

lib/Backends/Interpreter/Interpreter.cpp
Max-K-cadence: Err, you're right. Good catch!

lib/Backends/Interpreter/Interpreter.cpp
Max-K-cadence: Actually I just realized that we also miss float16<-> conversions. The point here is to make all types `{int32, int64, float16, float32}` convertable to one another. I'll add support for float16, and this check will simplify a lot after that..

tests/unittests/OnnxImporterTest.cpp
Max-K-cadence: Actually this was supposed just to check that the model loads, but you're right, I'll add more checks.

Max-K-cadence: Surprisingly, this test isn't working. Conversion `float->int32->int64->int32->float` works as NOOP when ONNX loads it. Likely it's an underlying bug exposed. Investigating.

Max-K-cadence: The bug is in `OptimizeConversions`. It thinks it's a NOOP. Trying to fix it...

Max-K-cadence:
Putting WIP. I need to find a correct fix for graph optimizer.

Max-K-cadence:
Actually this ONNX test with fix for optimization can go separately.

tests/unittests/OnnxImporterTest.cpp
Max-K-cadence: Will go along with fix.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -3117,35 +3117,31 @@ void BoundInterpreterFunction::fwdIntLookupTableInst(
...
-  case ElemKind::Int64ITy:
-    assert(dest->getElementType() == ElemKind::FloatTy &&
-           "Conversion not supported");
-    dest->copyWithCast<float, int64_t>(source);
-    return;
-  default:
-    llvm_unreachable("Type not supported");
-  }
+
+#define CONVERT(_T_FROM_, _T_TO_, _DTY_FROM_, _DTY_TO_)                        \
```
rdzhabarov: nit: typically we do not have `_` params in the macro. Could you make it. T_FROM, T_TO, etc.

rdzhabarov:
@Max-K-cadence lgtm. Thank you! One nit and will merge after internal CI success.

lib/Backends/Interpreter/InterpreterNodes.cpp
Max-K-cadence: Well, all macros in OperatorTest.cpp use it for `_OP_NAME_` but not for other params. I guess that's the convention... I'll do the same then.

PR #3181, opened 2019-06-26
[PyTorch] Support more JIT IR nodes to PyTorchModelLoader

jackm321:
Summary:
* Add support for loading `Constant`, `add`, `sub`, `conv2d`, and `relu` PyTorch JIT IR Nodes
* Setup operator tests using PyTest and setuptools
* Added proper cl flags handling to setup.py

Documentation:
* doyxgen comments
* Add a short getting started guide in pytorch.md

Test Plan:
`python setup.py test`

Screen Shot 2019-06-26 at 2 04 15 PM


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -25,47 +28,354 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  glow::NodeValue rhs = getGlowNodeValue(inputs[0]);
+  glow::NodeValue lhs = getGlowNodeValue(inputs[1]);
+  glow::MulNode *glowNode = f_->createMul("mul", rhs, lhs);
+  addGlowNodeValue(outputs[0], glowNode->getResult());
+}
 
+void PyTorchModelLoader::loadDiv(const torch::jit::Node *ptNode) {
```
zrphercule: Great! I think this kind of design is much better than switch case!

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -25,47 +28,354 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+  // Glow expects conv weights to be in CRSK but PyTorch keeps them in CKRS
+  // so we tranpose them. C - output_depth, R - filter_height, S -
+  // filter_width, K - input_depth.
+  glow::NodeValue weights = getGlowNodeValue(inputs[1]);
+  weights = f_->createTranspose("conv_weights_transposed", weights, NCHW2NHWC);
+  glow::ShapeNHWC weightsShape(weights.dims());
+
+  // If a bias was provided then use it otherwise create a 0 bias.
+  glow::NodeValue bias;
+  if (hasGlowNodeValue(inputs[2])) {
```
zrphercule: One question: in jit if we dont have a param, when do we suppose to expect "hasGlowNodeValue" to be false, and when do we expect its value to be nonsense (like transposeHandle)?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -25,47 +28,354 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+  (void)transposedHandle;
+
+  glow::Handle<int32_t> groupsHandle =
+      getGlowConstantHandle<int32_t>(inputs[8]);
+  assert(groupsHandle.size() == 1);
+  uint32_t groups = groupsHandle.raw(0);
+
+  std::vector<uint32_t> kernels = {static_cast<uint32_t>(weightsShape.h),
+                                   static_cast<uint32_t>(weightsShape.w)};
+
```
zrphercule: btw where to see docs about other 3 inputs? (since we only deal with 9 here)

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -25,47 +28,354 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+    const auto &vals = c10::impl::toVector(valsList);
+    t.reset(glow::ElemKind::BoolTy, {vals.size()});
+    auto handle = t.getHandle<bool>();
+    for (size_t i = 0; i < vals.size(); ++i) {
+      handle.raw(i) = vals[i];
+    }
+  } else if (iVal.isTensor()) {
+    assert(false && "not yet implemented tensor constants");
+  } else if (iVal.isNone()) {
+    // Nothing to do for None
```
zrphercule: Do we need to initialize t to be none here?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -25,47 +28,354 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+  case at::aten::relu:
+    return loadRelu(ptNode);
   default:
     assert(false && "unrecognized node");
   }
 }
 
+// static
+bool PyTorchModelLoader::isNodeSupported(const torch::jit::Node *node) {
+  switch (node->kind()) {
```
zrphercule: constant?

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Agreed, a while ago I converted the Caffe2 model loader to this style, I think it's easier to work with.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: I referenced the code https://fburl.com/6m7fgaf9

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Basically we are handling None as an absence a of Glow::NodeValue so when we see one, we want to do nothing but it's still valid to see one

torch_glow/src/PyTorchModelLoader.cpp
jackm321: what do you mean?

jackm321: Oh Constant node? That's a good point, I meant to add a comment for this. So we don't want to actually add it here or else I think pt will hand over all constants to glow because glow says it can support them. If we leave it out then pt hands glow the constants that it needs because it has to express them somehow in the subgraph that gets handed to glow

torch_glow/src/PyTorchModelLoader.cpp
jackm321: conv bias is optional, it is specified by a constant with None output if there is no bias, we handle this by loading None as nothing. This gives basically the same behavior as the c2 model loader.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: oh okay, then the other 3 does not seem to be related.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Yeah, I mean t is not intialized here, is it okay?

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Agreed, a little different from XXXFill ops in c2.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Yes it's on purpose, Glow has no equivalent of python `None`

PR #3184, opened 2019-06-27
Use native batch matmul kernel

bertmaher:
Summary:
The vendor supports batched matrix multiply; let's use it instead of
emulating it with split/matmuls/concat/reshape.

Differential Revision: D16025030

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -1024,6 +1024,19 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
       }
       break;
     }
+    case Kinded::Kind::BatchMatMulNodeKind: {
+      auto *MI = llvm::cast<BatchMatMulNode>(&I);
+      std::vector<synTensor> inputs;
+      inputs.push_back(tensors[MI->getLHS()].get());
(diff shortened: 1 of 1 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -1024,6 +1024,19 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
       }
       break;
     }
+    case Kinded::Kind::BatchMatMulNodeKind: {
+      auto *MI = llvm::cast<BatchMatMulNode>(&I);
+      std::vector<synTensor> inputs;
+      inputs.push_back(tensors[MI->getLHS()].get());
+      inputs.push_back(tensors[MI->getRHS()].get());
+      chk(synCreateGenericNode(
+          inputs.data(), &tensors[MI].get(), 2, 1, nullptr,
```
jfix71: nit: Looks like this happens for all the calls to `synCreateGenericNode()`, but I have no idea what 2 or 1 represent here. Might be good at some point to go through and add comments to each. Is it input size/output size or something?

bertmaher: Yeah, it’s input/output size.  Should probably get it from the vector size...

PR #3185, opened 2019-06-27
[Optimizer] Add PassManager and FunctionPassPipelines

### LINKED ISSUE #1641: Optimization pipeline design with more backend control

jfix71:
Summary: Add PassManager and FunctionPassPipelines. I added OptimizationOptions from #3161 with slight modifications.

One open question right now is the best way to allow the backend to have at least some control over the passes. 

- Currently in the PR I added `OptimizationOptions::funPassesToSkip` which is a set of passes to skip during optimizations, and then pass `OptimizationOptions` into the Backend at the start of `optimizeFunction()` to allow it to modify it. This is a bit lacking in functionality, but it could be used during both calls to `optimize()` and `fold()`.
- Alternatively I considered adding the entire `FunctionPassPipeline` for optimizations to `OptimizationOptions`. This means it can be modified however the backend wants (adding in passes we may want to keep out of the default set; rearranging opts; etc.). But this feels a bit weird because there are different pipelines, e.g. the one for Fold, and the one for pre-lowering (though this one is supposed to be independent of the backend anyway and so the backend shouldn't really have any say here...). Maybe we could add in `foldPipeline`, `optPipeline`, etc. to the `OptimizationOptions` instead of having a single one.

Documentation: Added.

Fixes #1641

Test Plan: Tests still pass.


lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "glow/Optimizer/GraphOptimizer/PassManager.h"
+
+#include "glow/Optimizer/GraphOptimizer/FunctionPasses.h"
+
+#include <glog/logging.h>
+
+using namespace glow;
+
+bool FunctionPassManager::runPrePass(const FunctionPass &P) { return false; }
```
opti-mix: `runPrePass` and `runPostPass` could be good places to insert some dumping/logging before/after each pass.

opti-mix: Also, it could be a good idea to conditionally (if requested by the command line option or cctx) call a verifier before/after each pass.

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,93 @@
...
+using FunctionPassPipeline = llvm::SmallVector<FunctionPassConfig, 32>;
+
+/// Manager for running a series of FunctionPasses. Given some Function,
+/// CompilationContext, and provided Pipeline, it will run all passes on the
+/// Function. Enables easier debugging given runPrePass() and runPostPass()
+/// calls, which can be modified to run some code before or before every pass.
+class FunctionPassManager {
+private:
+  /// The Function to run the passes on.
+  Function &F_;
```
opti-mix: Please add at least a getter for it.

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,93 @@
...
+
+  /// The pipeline of passes to run.
+  const FunctionPassPipeline &pipeline_;
```
opti-mix: Do we want to have a const getter here?

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,93 @@
...
+  }
+};
+
+using FunctionPassPipeline = llvm::SmallVector<FunctionPassConfig, 32>;
+
+/// Manager for running a series of FunctionPasses. Given some Function,
+/// CompilationContext, and provided Pipeline, it will run all passes on the
+/// Function. Enables easier debugging given runPrePass() and runPostPass()
+/// calls, which can be modified to run some code before or before every pass.
+class FunctionPassManager {
```
opti-mix: BTW, it could useful to add a name to the FPM, similar to stageName in Swift PM (https://github.com/apple/swift/blob/master/include/swift/SILOptimizer/PassManager/PassManager.h). This is useful for debugging and logging.

beicy:
Great work! 
I have a question about optimization pipeline, does it mean the optimization MUST follow the sequence (that is, the later pass depends on the previous pass) ? E.g. I saw a few DCE() in the pipeline. If some DCE() are removed, does it mean we will loss some optimization abilities?

jfix71:
>I have a question about optimization pipeline, does it mean the optimization MUST follow the sequence (that is, the later pass depends on the previous pass) ? E.g. I saw a few DCE() in the pipeline. If some DCE() are removed, does it mean we will loss some optimization abilities?

@beicy The pipeline is just a sequential list of passes that the pass manager will execute in order. If a DCE is removed in the middle of the pipeline then yeah we may lose out on some optimizations if they depend on number of users and DCE would change number of users. This is no different than our current pipeline. We could consider adding dependencies for passes, i.e. `requireDCEPrePass` or something, to ensure this is not done. If so that'd be for a future PR.

include/glow/Optimizer/GraphOptimizer/FunctionPasses.h
```diff
@@ -20,14 +20,33 @@
...
+};
+
+/// Declare all FunctionPass classes.
 #define FUN_PASS(PASS_NAME)                                                    \
   class PASS_NAME : public FunctionPass {                                      \
   public:                                                                      \
     bool run(Function *F) override;                                            \
     llvm::StringRef getName() const override { return #PASS_NAME; }            \
+    FunctionPassID getID() const override {                                    \
+      return FunctionPassID::PASS_NAME;                                        \
```
rdzhabarov: is there any enforcement/check that pass names are unique?

include/glow/Backend/Backend.h
```diff
@@ -98,6 +98,10 @@ class Backend {
   /// performed.
   virtual bool shouldShareBuffers() const { return true; }
 
+  /// Modify the \p optOpts however desired.
```
beicy: nit: "optOpts"  -> "optimizationOpts"

include/glow/Optimizer/GraphOptimizer/FunctionPasses.h
jfix71: They're used inside an enum, so if there's a repeat name it should cause a compilation error.

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
```diff
@@ -29,12 +30,30 @@ class Module;
 class PlaceholderBindings;
 class Placeholder;
 
+/// \returns the default graph optimization pipeline.
+FunctionPassPipeline getDefaultGraphOptimizationPipeline();
```
beicy: If I understand it correctly, this "DefaultGraphOptimizationPipeline" is not target independent, and it contains all passes. But it is not clear to me in which case it should be called. "getTargetIndependentGraphOptimizationPipeline" should be the common opt pass. For users, I think they should add something in IndependentGraphOptimization pass?

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
```diff
@@ -57,18 +76,24 @@ void convertPlaceholdersToConstants(Function *F,
...
-/// Optimize the Function \p F given compilation options \p cctx for Backend \B.
-/// \returns success if all nodes in the final resulting optimized Function are
-/// supported by \p B; if not, this represents a compiler error.
+/// Optimize the Function \p F given compilation options \p cctx and pipeline
+/// \p pipeline for Backend \B. \returns success if all nodes in the final
+/// resulting optimized Function are supported by \p B; if not, this represents
+/// a compiler error.
 llvm::Error optimizeFunction(Function *F, const Backend &B,
-                             CompilationContext &cctx);
+                             CompilationContext &cctx,
```
beicy: I am not sure if an error should be returned. If an optimization is supported, then apply the optimization. If the optimization is not supported, can we just reports some warning and skip it? IMO, optimization is kind of "optional". Unless some opts are forced due to some execution correctness reason, we don't have to report a compiler error ?

lib/Optimizer/GraphOptimizer/OptimizationPipelines.cpp
```diff
@@ -0,0 +1,184 @@
...
+    // Optimize quantization related operators.
+    {FunctionPassID::OptimizeQuantization, PassConvergence::UntilFixedPoint},
+
+    // Optimize reshapes introduced during above optimizations.
+    {FunctionPassID::OptimizeReshape},
+
+    // Perform Dead Code Elimination.
+    {FunctionPassID::DCE},
+};
+FunctionPassPipeline glow::getDefaultGraphOptimizationPipeline() {
```
beicy: const?

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
```diff
@@ -29,12 +30,30 @@ class Module;
...
+/// \returns the default folding pipeline.
+FunctionPassPipeline getDefaultFoldPipeline();
+
 /// Perform optimizations on the graph representation.
-void optimize(Function *F, CompilationContext &cctx);
-void optimize(Function *F, CompilationMode mode);
-/// Fold nodes that were expressed lowered in the input model.
-void fold(Function *F, CompilationContext &cctx);
-void fold(Function *F, CompilationMode mode);
+void optimize(Function *F, CompilationContext &cctx,
```
rdzhabarov: > CompilationContext &cctx,

this reads as optimize would change the cctx. can that be `const`?

jfix71: It was dropped by @ZchiPitt in #3032, as they thought they were going to add the log to the cctx and modify it. But then that went in a different direction. So we could add it back.

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,93 @@
...
+  /// ID of the FunctionPass to run.
+  FunctionPassID passID{FunctionPassID::EmptyPass};
+
+  /// Convergence mode to inform the PassManager how to run the FunctionPass.
+  PassConvergence convergence{PassConvergence::OnePass};
+
+  /// Which modes the FunctionPass should be run in.
+  llvm::SmallSet<CompilationMode, 2> enabledModes;
+
+  FunctionPassConfig(FunctionPassID ID_,
```
rdzhabarov: nit: remove '_'

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2310,6 +2313,12 @@ bool OptimizeQuantization::run(Function *F) {
   }   // For each item in the worklist.
 
   changed |= optimizeQuantizedMaxSplat(F);
+
+  // If nothing has changed then run DCE and sink rescale quantization nodes.
```
rdzhabarov: Is this related to the PR?

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
rdzhabarov: Ok, I see backend might modify those params.

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
jfix71: Yeah I wasn't sure the best design here. I was thinking all backends would start with the pipeline with all passes ("DefaultGraphOptimizationPipeline") and then opt out of passes they didn't want via `OptimizationOptions ::funPassesToSkip`. But it could work in reverse, by adding just the TargetIndependent ones and then adding which additional passes they want.

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
jfix71: FYI returning an Error pre-dates this PR. I agree optimization passes are somewhat optional. Though one exception is lowering, since a backend might not support an op unless it gets lowered.

Either way, the motivation for returning an Error is essentially for reporting a compiler bug. For example, we return an error if the `cctx` is not set up correctly for a specific compilation setup (e.g. if `convertToFP16` is set when in profiling mode). We also return an error if after all of the optimization passes the backend reports it does not support one of the Nodes.

We could try to pull some of this error checking out into the callers if we decide it's problematic to return an Error here.

include/glow/Optimizer/GraphOptimizer/GraphOptimizer.h
beicy: IMO, adding passes may be more easier and debuggable.  Once a pass is added and tested, it is kind of reliable. Then another pass can be added, and so on.

jfix71:
Added a lot of updates here addressing most comments:
- Added FunctionPassSet implemented via a bitset
- Added class for FunctionPassConfig
- Made fold/optimize directly use their respective default pipelines for now
- Added class for FunctionPassPipeline
- Added in command line options for verification pre/post named passes or all passes
- Added name for FunctionPassManager
- Added getters to PassManager
- Fixed other random nits

Still TODO (future PR?):
- Global counter for #  passes run
- Insert some dumping/logging before/after each pass via run pre/post pass.

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include "llvm/Support/CommandLine.h"
+
+#include <glog/logging.h>
+
+using namespace glow;
+
+namespace {
+llvm::cl::OptionCategory passManagerCat("PassManager Options");
+
+llvm::cl::opt<bool> verifyBeforeAllPassesOpt(
```
opti-mix: Very nice! Could you also add similar flags for dumping the graph before/after some/all passes?

lib/Optimizer/GraphOptimizer/OptimizationPipelines.cpp
```diff
@@ -0,0 +1,139 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
+#include "glow/Optimizer/GraphOptimizer/PassManager.h"
+
+using namespace glow;
+
+FunctionPassPipeline glow::createDefaultGraphOptimizationPipeline() {
```
opti-mix: So, you decided to stay with the design where pipelines are very visible and exist outside of PassManagers? As opposed to hiding them inside PMs and using something like 
```c++
glow::addDefaultGraphOptimizationPasses(FunctionPassManager &PM) {
   PM.addPass(FunctionPassConfig(...DCE...));
   PM.addPass(FunctionPassConfig(...SinleCode...));
   ...
}
```

jfix71: I pushed a commit doing this. I was thinking we would want more direct access to the pipelines for a backend to manipulate, but that's not the case for now -- a backend can only opt out of passes from the default fold/opt. There's still a getter for a const ref to the pipeline from the PM, not sure it's useful or necessary for now.

lib/Optimizer/GraphOptimizer/PassManager.cpp
jfix71: Done.

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+  return false;
+}
+} // namespace
+
+bool FunctionPassManager::runPrePass(const FunctionPass &P) {
+  if (dumpDAGBeforeAllPassesOpt ||
+      listContainsString(dumpDAGBeforePassesOpt, P.getName())) {
+    F_->dumpDAG(F_->getName().str() + "_PrePass_" + P.getName().str() + ".dot");
```
opti-mix: If we run the same pass on the same function multiple times, this line would overwrite previous dumps and save only the last one. If would would maintain a pass number (inside PM or global) this could be easily solved by adding this pass number a suffix.

jfix71: Done.

opti-mix: Cool! Thanks!

But now that we have a pass counter let's go a step further ;-) ;-)
- Let's add an option to print all the passes being run, e.g. `Pass #N: CSE`.  May be also the pass manager name. This gives us a the log of which passes were run an in which order.
- Let's add an option to stop at the pass number N. This would allow for bisection of the pass pipeline in case when one of the passes introduces an error.

jfix71: Done 🙂 

- `-print-passes`
- `-stop-passes-after-num=N`

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -106,6 +118,10 @@ bool FunctionPassManager::runPrePass(const FunctionPass &P) {
       listContainsString(verifyBeforePassesOpt, P.getName())) {
     CHECK(F_->verify());
   }
+  if (printPassesOpt) {
+    LOG(INFO) << "Running Function Pass #" << globalPassCounter << ": "
```
opti-mix: We probably want to print the name of the function (and graph) as well.

jfix71: Not sure what you mean by function and graph -- I added the name of the Function for now.

include/glow/Backend/Backend.h
```diff
@@ -98,6 +98,10 @@ class Backend {
   /// performed.
   virtual bool shouldShareBuffers() const { return true; }
 
+  /// Modify the \p optimizationOpts however desired.
+  virtual void
+  setOptimizationOptions(OptimizationOptions &optimizationOpts) const {};
```
rdzhabarov: might be for future PRs, can we consolidate buffer sharing support in the OptimizationOptions

jfix71:
@opti-mix I've squashed all the old commits, and put up a new one to try to solve what we talked about offline, to allow the backend to return the FunctionPassPipeline. It's not exactly what we discussed -- I was having linking issues on shared builds trying to get the HostManager to be its own library (missing vtable entries for all of the implementations of the FunctionPasses which are in GraphOptimizer).

Instead I added a new `GraphOptimizerPipeline` library, which creates and returns the default graph optimization and fold pipelines. Additionally `FunctionPassID`, `FunctionPassConfig`, and `FunctionPassPipeline` are moved into `Pipeline.h`.

Let me know if you think this is a reasonable solution.

include/glow/Backend/Backend.h
```diff
@@ -98,6 +99,11 @@ class Backend {
   /// performed.
   virtual bool shouldShareBuffers() const { return true; }
 
+  /// Modify the \p optimizationOpts however desired.
+  virtual FunctionPassPipeline getOptimizationPipeline() const {
+    return createDefaultGraphOptimizationPasses();
```
opti-mix: Do we need to have this implementation inside the header file?

include/glow/Optimizer/GraphOptimizer/FunctionPasses.h
```diff
@@ -17,17 +17,32 @@
...
+    }                                                                          \
   };
+#include "FunctionPasses.def"
 
+/// Helper that creates and \returns a FunctionPass given a provided \p passID.
+inline std::unique_ptr<FunctionPass> createFunctionPass(FunctionPassID passID) {
```
opti-mix: Do we really need to have this implementation inside the header file?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2752,108 +2752,20 @@ void glow::fold(Function *F, CompilationMode mode) {
...
-  while (OptimizeQuantization().run(F) || SinkRescaleQuantizedNode().run(F)) {
-    DCE().run(F);
-    VLOG_IF_EVERY_N(0, google::COUNTER > 1, 100)
-        << "Warning: Quantization optimization applied another 100 iterations "
-           "without reaching fixed point";
-  }
-
-  // Optimize reshapes introduced during above optimizations.
-  OptimizeReshape().run(F);
+  FunctionPassManager FPM("TargetIndependentGraphOptzFPM",
```
opti-mix: This place looks funny: We ask a target backend to provide the optimization pipeline for target independent optimizations? ;-)

opti-mix: May be the name of the pass manager is a copy & paste error?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2954,6 +2866,9 @@ llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
                                    CompilationContext &cctx) {
   LOG_SCOPE(F->getLogContext(), "glow::optimizeFunction")
 
+  // Allow the backend to modify the optimization options to its preferences.
```
opti-mix: It is a TODO or you just forgot to remove it?

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    switch (passConfig.getConvergenceMode()) {
+    case FunctionPassConfig::ConvergenceMode::OnePass:
+      changed |= P->run(F);
+      break;
+
+    case FunctionPassConfig::ConvergenceMode::UntilFixedPoint:
+      while (P->run(F)) {
+        changed = true;
+        DCE().run(F);
```
opti-mix: I like the idea of running `DCE` after each graph-level optimization pass. But I'm not sure if this is the right place to do it. What if we soon decide to use `PassManager` not only for graph-level optimizations, but also for graph-level transformations or IR-level transformations? Maybe it is better to put this DCE call into a `standard` pre- or post-pass action that is executed for all graph FunctionPasses? And all graph function passes derive from some kind of a a GraphFunctionPass? Or we could have a base class for all pass managers and then some derived GraphOptimizationsPassManager, IROptimizationsPassManager, etc. And those derived classes would have these specialized pre/post actions? WDYT?

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+bool FunctionPassManager::runPostPass(Function *F,
+                                      const CompilationContext &cctx,
+                                      const FunctionPass &P) {
+  if (printPassesOpt) {
+    LOG(INFO) << "Finished Pass #" << globalPassCounter << ": "
+              << P.getName().str() << " on Function: \""
+              << F->getName().str() + "\"\n";
+  }
+  if (dumpDAGAfterAllPassesOpt ||
```
opti-mix: See my comments about moving this kind of logic into more specialized derived classes.

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include "llvm/ADT/SmallVector.h"
+
+namespace glow {
+
+/// Manager for running a series of FunctionPasses. Given some Function,
+/// CompilationContext, and provided Pipeline, it will run all passes on the
+/// Function. Enables easier debugging given runPrePass() and runPostPass()
+/// calls, which can be modified to run some code before or before every pass.
+class FunctionPassManager : public Named {
```
opti-mix: I wonder if we should move this header into a dedicated PassManager directory and prepare it to be a base class for graph-based and IR-based pass managers for optimizations and transformation passes? The reason I suggest to do it now is because it would affects paths and CMake files and thus it is probably easier to do it now, to avoid the updates of all these places in the future.  But this is just a suggestion. I don't have any strong opinions about it.

jfix71: I started to try this, but I'm not sure exactly how it will end up looking once we integrate other passes in. I imagine I'd need to update CMake stuff again either way. So I'd rather just do it on future PRs as the design is fleshed out.

lib/Optimizer/GraphOptimizer/FunctionPass.cpp
```diff
@@ -0,0 +1,156 @@
...
+bool FunctionPass::run(Function *F, const CompilationContext &cctx) {
+  const int count = ++FunctionPass::globalPassCounter;
+
+  // If we've exceeded the number of passes to run then early exit.
+  if (stopAfterPassNumOpt >= 0 && count > stopAfterPassNumOpt) {
+    return false;
+  }
+
+  bool changed = runPrePass(F, cctx, count);
+  changed |= passImpl(F, cctx);
```
opti-mix: I understand why you make clients override `passImpl`. But people who know e.g. LLVM are more used to overriding `run`. Maybe you can play with names and make clients override `run` and rename current `run`  into e.g. `runPass`, which would match nicely `runPrePass` and `runPostPass`? 

Nit: Another related question is: who should invoke pre and post functions? The pass itself or the PassManager? It can be done both ways. But if it is controlled by the pass, the PassManager has less insight into these three phases: pre, run, post and thus cannot e.g. log before/after each of these phases, etc. But this is probably not so important.

include/glow/Optimizer/GraphOptimizer/FunctionPasses.h
```diff
@@ -16,19 +16,28 @@
...
 #define GLOW_OPTIMIZER_GRAPHOPTIMIZER_FUNCTIONPASSES_H
 
+#include "glow/Optimizer/GraphOptimizer/CompilationContext.h"
 #include "glow/Optimizer/GraphOptimizer/FunctionPass.h"
+#include "glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h"
+#include "glow/Support/Support.h"
 
 namespace glow {
 
+/// Declare all FunctionPass classes.
```
opti-mix: Should we have a version of FUN_PASS that takes a base class as a parameter, so that one could say e.g. `FUN_PASS(myPassName, FunctionPassFollowedByDCE)` or something like that?

opti-mix: Also, if the property `needs to run DCE after the pass` is not inherent to a pass, but is rather a property defined by the pipeline (e.g. it uses iterate until fixed point), then it should be rather the pass manager who decides to call DCE. But it should probably be a Graph IR specific PassManager who does it as other pass managers (for low-level IR) won't even have the DCE pass.

jfix71:
@opti-mix I've updated this based on our offline conversation. 

- Removed previous commit, so much of the pass logic is back in PassManager
- Disallow passes from being used outside of a PassManager
- Add DCERequiredMode -- I thought this would be cleaner than adding virtual pre/post hooks to each FunctionPass to override, LMK what you think
- Move SinkRescaleQuantizedNode to a static helper, called from OptimizeQuantization, so that we don't have weirdness/a recursive optimization pipeline wherein we have a Pass creating a new PassManager, which felt a bit gross/unnecessary.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
```diff
@@ -0,0 +1,133 @@
...
+  UntilFixedPoint, // Run the pass over the Function until a fixed point is
+                   // reached.
+};
+
+/// Specifies a configuration for running a FunctionPass when used in a
+/// FunctionPassPipeline.
+class FunctionPassConfig {
+public:
+  /// Specifies whether the pass requires DCE.
+  enum class DCERequiredMode {
```
opti-mix: Maybe move it before `FunctionPassConfig`, right after `ConvergenceMode`? I just don't see any reason for one of them to be defined outside and another one inside `FunctionPassConfig`.

lib/Optimizer/GraphOptimizer/PassManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+bool FunctionPassManager::runPass(const FunctionPassConfig &passConfig,
+                                  Function *F, const CompilationContext &cctx) {
+  const FunctionPassID &passID = passConfig.getFunctionPassID();
+  assert(!(passID == FunctionPassID::DCE &&
+           passConfig.getDCERequiredMode() ==
+               FunctionPassConfig::DCERequiredMode::RequireDCEBefore) &&
+         "Cannot specify DCE requires DCE before it.");
+
+  // Special case for ConstantFold; skip it if specified.
```
opti-mix: Is it worth to special case `ConstanFold` here? Why not doing this check inside the pass itself using the provided `cctx`? We have a special-cased DCE already ;-)

Oh, this is probably because passes do not get `cctx` passed to them currently. Is it actually a good idea to not pass a CompilationContext down to passes? What if we have some options that control how a given pass should do its down? How would we inform a pass about it?

include/glow/Optimizer/GraphOptimizer/PassManager.h
```diff
@@ -0,0 +1,69 @@
...
```
opti-mix: Should most of the methods in `FunctionPassManager` be virtual?

lib/Optimizer/GraphOptimizer/Lower.cpp
```diff
@@ -1059,5 +1059,5 @@ void glow::lower(Function *F, CompilationContext &cctx, const Backend *B,
   }
 
   // Remove nodes that were lowered.
-  DCE().run(F);
+  FunctionPassManager("DCE_FPM", {getDCEPassConfig()}).run(F, cctx);
```
opti-mix: Minor nit: Not a big deal, but may be introduce a `runDCEPass(CompliationContext &cctx)` helper function? It seems like you use this line of code in multiple places.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
opti-mix: Nit: Overall, I'm still not a very big fan of making DCE so special. But it is probably OK. 

Maybe to make it less special we could have here something like `PassID prePass`, which is the pass to always run before the current one? It can be initialized to DCE pass ID by default, but could be something else if needed. `getDCERequiredMode` would become `getPrePass` then. And some checks will become less DCE centric. Don't know if it is an improvement though, as it the only actual use-case for now will be using DCE and my proposal could kind of make it less obvious ;-) WDYT?

include/glow/Optimizer/GraphOptimizer/PassManager.h
jfix71: In terms of allowing future pass managers to inherit from this?

I was thinking in a future PR I would create an abstract PassManager class with these methods as virtual, and then the FunctionPassManager/ModulePassManager/IRFunctionPassManager/etc. would implement them. But that requires some work on potentially adding some IRUnit class for Module/Function/Node/Instruction to inherit from, or templatizing it, and so thought we would do that in future PRs.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
jfix71: I thought DCERequiredMode was more specific to FunctionPass, while ConvergenceMode might be used by future Module/IRFunction passes. But NBD, can move it out.

jfix71: I thought about this, but I was a bit weary of it because I didn't want the pipeline to have the potential of accidentally infinite looping. E.g. you could imagine a circular dependence on passes via `prePass` which isn't trivial to detect. For now I made it DCE because you can always check that DCE does not require DCE before.

lib/Optimizer/GraphOptimizer/PassManager.cpp
jfix71: Sure, probably makes sense to pass the cctx everywhere.

include/glow/Optimizer/GraphOptimizer/PassManager.h
opti-mix: > In terms of allowing future pass managers to inherit from this?

Yes. It could happen even for Function-based PassManagers, e.g. if you want to redefine pre/post processing somewhat or something like that. But this is a very minor comment. Feel free to ignore for now.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
opti-mix: > E.g. you could imagine a circular dependence on passes via prePass which isn't trivial to detect.

Pass demanding itself as a prePass is very easy to detect (pretty much like you do for DCE now), but any non-obvious recursion would require more effort to detect it of course (e.g. one could check the list of passes executed so far and check for repeating never ending patterns ;-). 

As I said, I don't think it is a big issue. I jut don't like the fact that DCE is so special-cased at the moment.

opti-mix:
@jfix71 BTW, for debug purposes it could be a good idea to add some `dump` methods to PassManager and to Pipelines.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
jfix71: Yeah the non-obvious recursion is what I was thinking of. There aren't other passes which depend on others right now anyway other than DCE, so even if we added such an ability it's not currently useful, and like you said it makes the use-case less obvious. I'd prefer to just stick with DCE special cased for now and if we find its useful to expand it in the future we can.

jfix71:
@opti-mix All comments addressed. I added some `dump()` functions to Pipeline, PassManager, FunctionPassConfig. I also made the current index in the pipeline a member variable of the PassManager so when we dump it we know where we are in its pipeline.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
```diff
@@ -0,0 +1,142 @@
...
+/// Define an enum to identify all FunctionPasses, used to declare inside a
+/// FunctionPassConfig for a Pipeline.
+enum class FunctionPassID {
+#define FUN_PASS(PASS_NAME) PASS_NAME,
+#include "glow/Optimizer/GraphOptimizer/FunctionPasses.def"
+};
+
+/// Specifies convergence mode for a FunctionPass.
+enum class ConvergenceMode {
+  OnePass,         // Run a single pass over the Function.
```
opti-mix: Very minor nit: We usually put comments before declarations, i.e. on the previous line.

include/glow/Optimizer/GraphOptimizer/Pipeline/Pipeline.h
```diff
@@ -0,0 +1,142 @@
...
+  UntilFixedPoint, // Run the pass over the Function until a fixed point is
+                   // reached.
+};
+
+/// Specifies whether the pass requires DCE.
+enum class DCERequiredMode {
+  RequireDCEBefore, // Require that DCE is run before the pass.
```
opti-mix: Same here.

PR #3186, opened 2019-06-27
[compilationLog]: A script to generate DAG at any fixed point of Glow optimization process, with the input of compilation logs

ZchiPitt:
Summary:
Implemented a python script, such that it:
- Reads from compilation log.
- Reconstructs the node graph.
- Can dump the DAG before and after any certain phase(lower, fold ...etc) inside the log, with user using a command line flag to specify which phase to dump.

Test Plan:
Dump the log of MLTest/convNetForImageRecognition. Generate dot file with the command 
"python ../glow/utils/log_parser.py -f xxx.log --dump-phases lower"
And the DAG before *lower* looks like below:
Screen Shot 2019-06-27 at 1 14 30 AM

The DAG after *lower* looks like below:
Screen Shot 2019-06-27 at 1 13 14 AM

### DIFF
--- a/utils/log_parser.py
+++ b/utils/log_parser.py
@@ -1,0 +1,303 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017-present, Facebook, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import re
+
+# Maintaining all nodes
+NODES_MAP = {}
+
+# Maintaining all edges
+EDGES_MAP = {}
+
+# Scope stack
+SCOPE_STACK = []
+
+
+class Node:
+    def __init__(self, kindName, name):
+        self.kindName_ = kindName
+        self.name_ = name
+        self.inputs_ = []
+        self.users_ = {}
+
+    def get_kind_name(self):
+        return self.kindName_
+
+    def get_name(self):
+        return self.name_
+
+    def get_inputs(self):
+        return self.inputs_
+
+    def get_users(self):
+        return self.users_
+
+    def add_user(self, u):
+        if u not in self.users_:
+            self.users_[u] = 0
+        self.users_[u] += 1
+
+    def remove_user(self, u):
+        assert u in self.users_
+        self.users_[u] -= 1
+        if self.users_[u] == 0:
+            del self.users_[u]
+
+    def has_no_uses(self):
+        return len(self.users_) == 0
+
+    def set_input(self, node, resNo):
+        self.inputs_.append((node, resNo))
+
+    def replace_input(self, oldNode, oldResNo, newNode, newResNo):
+        assert (oldNode, oldResNo) in self.inputs_
+        tmp_inputs = [(newNode, newResNo)]
+        for n, rn in self.inputs_:
+            if n != oldNode or rn != oldResNo:
+                tmp_inputs.append((n, rn))
+        self.inputs_ = tmp_inputs
+
+    def set_scope_of_creation(self, creationScopeName):
+        self.creationScopeName_ = creationScopeName
+
+
+class DottyPrinter:
+    def __init__(self):
+        self.vertices_ = []
+        self.edges_ = []
+        self.uniqueVertexMap = {}
+        self.uniqueVertexNo = 0
+        self.colors = ["AliceBlue", "CadetBlue1", "Coral", "DarkOliveGreen1", "DarkSeaGreen1", "GhostWhite", "Khaki1", "LavenderBlush1", "LemonChiffon1",
+                       "LightSkyBlue", "MistyRose1", "MistyRose2", "PaleTurquoise2", "PeachPuff1", "PowderBlue", "Salmon", "Thistle1", "Thistle3", "Wheat1", "Yellow2"]
+
+    def get_unique_vertex_name(self, node):
+        if node not in self.uniqueVertexMap:
+            self.uniqueVertexMap[node] = self.uniqueVertexNo
+            self.uniqueVertexNo += 1
+        return "v" + str(self.uniqueVertexMap[node])
+
+    def dump_label(self, node):
+        labelStr = "{"
+        labelStr += " {<Inputs>Inputs}|"
+        labelStr += "{" + node.get_kind_name() + "\lname: " + \
+            node.get_name() + "}|"
+        labelStr += " {<Outputs>Outputs}"
+        labelStr += "}"
+        return labelStr
+
+    def get_color(self, node):
+        idx = hash(node.get_kind_name()) % len(self.colors)
+        return self.colors[idx]
+
+    def dump_node(self, node):
+        if not node:
+            return
+        nodeStr = self.get_unique_vertex_name(node) + "[\n"
+        nodeStr += "\tlabel = \""
+        nodeStr += self.dump_label(node)
+        nodeStr += "\"\n"
+        nodeStr += "\tshape = \"record\"\n"
+        nodeStr += "\tstyle=\"filled,rounded\"\n"
+        nodeStr += "\tfillcolor="
+        nodeStr += self.get_color(node)
+        nodeStr += "\npenwidth = 2];\n"
+        self.vertices_.append(nodeStr)
+
+    def visitNodes(self):
+        for k, n in NODES_MAP:
+            self.dump_node(NODES_MAP[(k, n)])
+
+    def visitEdges(self):
+        for k, n in NODES_MAP:
+            node = NODES_MAP[(k, n)]
+            for nodeInput in node.get_inputs():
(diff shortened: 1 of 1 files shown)


utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+    dotty = DottyPrinter()
+    dotty.dump_graph(dagName)
+
+
+def process(logLines, dumpPhases):
+    phaseNo = 0
+    for ln in logLines:
+        # Process CREATE statement
+        if re.search("CREATE", ln):
+            p = re.compile(
```
opti-mix: You probably want to hoist the compilation of a regex out of the loop. It will make things significantly faster.

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+def dump_dag(dagName):
+    dotty = DottyPrinter()
+    dotty.dump_graph(dagName)
+
+
+def process(logLines, dumpPhases):
+    phaseNo = 0
+    for ln in logLines:
+        # Process CREATE statement
+        if re.search("CREATE", ln):
```
opti-mix: Are you just looking for a constant substring? Why using `re` then?

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+                # Create the node with given kind and name
+                scopeName = g[0]
+                kindName = g[1].replace(" ", "")
+                nodeName = g[2].replace(" ", "")
+                createdNode = Node(kindName, nodeName)
+                createdNode.set_scope_of_creation(scopeName)
+                NODES_MAP[(kindName, nodeName)] = createdNode
+
+                # Set the inputs of the created node
+                inputs = re.findall('\([^\(\)]*\)', g[3])
```
opti-mix: Is the first argument of `findall` a regex? If so, may be it should be hoisted out of the loop as otherwise it is probably each time.

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+                if len(inputs) == 0:
+                    # there's no node input for Splat
+                    assert kindName in ("Splat", "Constant", "Placeholder")
+                for i in inputs:
+                    pi = re.compile('^\(Kind:(.*),.*Name:(.*),.*ResNo:(.*)\)$')
```
opti-mix: Hoist compile out of the loop.

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+                        inputNodeResno = gi[2].replace(" ", "")
+                        assert (inputNodeKind, inputNodeName) in NODES_MAP
+                        # (node, resno)
+                        inputNode = NODES_MAP[(inputNodeKind, inputNodeName)]
+                        createdNode.set_input(inputNode, inputNodeResno)
+                        inputNode.add_user(createdNode)
+
+        # Process NODE_INPUT_CHANGE statement
+        elif re.search("NODE_INPUT_CHANGE", ln):
+            p = re.compile(
```
opti-mix: hoist compile out of the loop

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+
+                # change the input of changedNode
+                changedNode.replace_input(
+                    prevNode, prevNodeResno, newNode, newNodeResno)
+                prevNode.remove_user(changedNode)
+                newNode.add_user(changedNode)
+
+        # Process DELETE statement
+        elif re.search("DELETE", ln):
+            p = re.compile(
```
opti-mix: hoist compile out of the loop

nickgg:
Nice work!

A couple initial questions:
* What's with the orphaned `gsum` placeholder? That looks like it might belong to an SGD node, but I don't see one in the graph. Can you try hacking up glow to dump the dag before/after lower for the same input and see if they match?
* It would be cool to colour the changed nodes in the "after" DAG something special so it's easy to see what's changed. E.g. Make the matul & batchedadd in the second graph stand out somehow.

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+NODES_MAP = {}
+
+# Maintaining all edges
+EDGES_MAP = {}
+
+# Scope stack
+SCOPE_STACK = []
+
+# Critical Phases
+CRITICAL_PHASES = ["lower", "optimize", "fold", "optimizeFunction",
```
nickgg: This list of phases needs to be kept in sync with the Glow compiler code (eg. if somebody adds a new phase they need to update this). Can we remove it somehow?

Perhaps if a list of phases is not provided on the command line we can build a list of phases that were encountered parsing the log?

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+        return self.name_
+
+    def get_inputs(self):
+        return self.inputs_
+
+    def get_users(self):
+        return self.users_
+
+    def add_user(self, u):
+        if u not in self.users_:
```
nickgg: You could simplify this using `DefaultDict`: https://docs.python.org/2/library/collections.html#collections.defaultdict

utils/log_parser.py
nickgg: Could compile this ahead of time with the `p` regex.

utils/log_parser.py
```diff
@@ -0,0 +1,311 @@
...
+                # in module destructor.
+                if "null" in g[4]:
+                    continue
+                pOpr = re.compile('^\(.*Kind:(.*),.*Name:(.*),.*ResNo:(.*)\)$')
+                mPrev = re.match(pOpr, g[3].replace(" ", ""))
+                mNew = re.match(pOpr, g[4].replace(" ", ""))
+
+                # Previous operand
+                assert mPrev
+                prevNodeKind = mPrev.groups()[0]
```
nickgg: nit: could do (prevNodeKind, prevNodeName, prevResNo) = mPrev.groups()[:3] here?

ZchiPitt:
Thanks for the reviews.
@opti-mix 

> Does this tool understand that there could be multiple rounds of each phase? E.g. there could be multiple rounds of optimize, etc. Will the tool produce multiple images for them or a single one?
There will be multiple dot files generated with a timestamp every time we enter/leave one phase.

@nickgg 

> What's with the orphaned gsum placeholder? That looks like it might belong to an SGD node, but I don't see one in the graph. Can you try hacking up glow to dump the dag before/after lower for the same input and see if they match?
The output is immediate after the lower phase, so there might be many orphaned nodes that didn't get DCE'ed. Actually there were many more orphaned constants in the dot files, I just didn't include them. To improve the DAG, I'll probably just dump the orphaned nodes in the future version.
> It would be cool to colour the changed nodes in the "after" DAG something special so it's easy to see what's changed. E.g. Make the matul & batchedadd in the second graph stand out somehow.
Prob will be able to do this after I include the filtering tool for transformation later.

utils/log_parser.py
```diff
@@ -0,0 +1,303 @@
...
+        nodeStr += "\"\n"
+        nodeStr += "\tshape = \"record\"\n"
+        nodeStr += "\tstyle=\"filled,rounded\"\n"
+        nodeStr += "\tfillcolor="
+        nodeStr += self.get_color(node)
+        nodeStr += "\npenwidth = 2];\n"
+        self.vertices_.append(nodeStr)
+
+    def visitNodes(self):
+        for k, n in NODES_MAP:
```
SplitInfinity: You can iterate through the values directly:
```
for node in NODES_MAP.values():
  self.dump_node(node)
```

utils/log_parser.py
```diff
@@ -0,0 +1,303 @@
...
+            self.dump_node(NODES_MAP[(k, n)])
+
+    def visitEdges(self):
+        for k, n in NODES_MAP:
```
SplitInfinity: Same here.

utils/log_parser.py
```diff
@@ -0,0 +1,303 @@
...
+                if (i.get_kind_name(), i.get_name()) not in NODES_MAP:
+                    print(i.get_kind_name(), i.get_name())
+                edgeStr = self.get_unique_vertex_name(i)+":Outputs -> "
+                edgeStr += self.get_unique_vertex_name(node)+":Inputs"
+                self.edges_.append(edgeStr)
+
+    def dump_graph(self, dagName):
+        self.visitNodes()
+        self.visitEdges()
+        f = open(dagName + "_dotty.dot", "w")
```
SplitInfinity: Consider using `with` instead of explicitly calling `open` and `close`:
```
with open(f"{dagName}_dotty.dot", "w") as f:
  f.write(...)
  ...
```

utils/log_parser.py
```diff
@@ -0,0 +1,303 @@
...
+    pEnter = re.compile('^ENTERSCOPE:(.*)$')
+    pExit = re.compile('^EXITSCOPE:(.*)$')
+
+    for ln in logLines:
+        # Process CREATE statement
+        if "CREATE" in ln:
+            m = re.match(pCreate, ln)
+            if m:
+                g = m.groups()
+                assert len(g) == 4
```
SplitInfinity: Consider adding a helpful message here in case this `assert` fails (and for all others in the file).

utils/log_parser.py
```diff
@@ -0,0 +1,303 @@
...
+        return labelStr
+
+    def get_color(self, node):
+        idx = hash(node.get_kind_name()) % len(self.colors)
+        return self.colors[idx]
+
+    def dump_node(self, node):
+        if not node:
+            return
+        nodeStr = self.get_unique_vertex_name(node) + "[\n"
```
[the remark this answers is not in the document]
SplitInfinity: In case you haven't see one before, they work like this:
```
a = 1
b = 2
c = 3

message = f"As easy as {a}, {b}, {c}" # As easy as 1, 2, 3
```

PR #3189, opened 2019-06-27
Truncate 64-bit indices to 32

bertmaher:
Summary:
We've found that PCIe transfers are a bottleneck for networks that do
a large number of embedding lookups.  Part of the problem is that these indices
are often unnecessarily specified as int64 due to ML framework defaults.
Ideally we'd fix the framework, but as a workaround we forcibly truncate these indices to 32 bits.

Differential Revision: D15609734

### DIFF
--- a/lib/Backends/Habana/Habana.cpp
+++ b/lib/Backends/Habana/Habana.cpp
@@ -125,9 +125,13 @@ public:
   /// Create an un-allocated handle.  Useful for moving objects.
   TensorHandle() = default;
 
-  /// Constructor. Create a tensor from Glow IR Value \p V.
-  TensorHandle(TypeRef V, llvm::StringRef name, void *buffer = nullptr,
-               IOType ioType = IOType::Intermediate)
+  /// Constructor. Create a Synapse tensor.
+  /// \param downcast64To32 If this tensor is a 64-bit type, forcible cast to
+  /// 32-bits. \param V Type of the tensor \param name a string identifying the
+  /// tensor \param buffer the static weights of the tensor \param ioType the
+  /// I/O class of the tensor
+  TensorHandle(bool downcast64To32, TypeRef V, llvm::StringRef name,
+               void *buffer = nullptr, IOType ioType = IOType::Intermediate)
       : buffer_(buffer), name_(name) {
     if (!buffer_) {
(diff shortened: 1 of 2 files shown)


lib/Backends/Habana/Habana.cpp
```diff
@@ -422,6 +435,26 @@ static bool allowsPartialInput(const Node *src, const Node *dst) {
   return false;
 }
 
+/// \returns true if \p dst allows \p src to be converted from 64 to 32 bits.
+static bool allows64To32Downcast(const Node *src, const Node *dst) {
```
jfix71: This would improve perf of SLS for all backends. Can this be a pass that we do higher in the stack instead of buried in/limited to one backend?

bertmaher: Right -- I think really this is really something that can be easily fixed at the model layer.  Doing it in the Habana backend is a short-term hack to help figure out the perf implications of 64-vs-32 bit indices.

I'm uncertain what the right way to provide this for all backends is... maybe insert some 64-to-32 casts and have that handled by the CPU backend?  (Of course, this particular transformation is a pessimization for CPU, but should be a win for ASICs and maybe GPUs)

lib/Backends/Habana/Habana.cpp
```diff
@@ -489,22 +541,41 @@ llvm::Error HabanaFunction::execute(ExecutionContext *context) {
     RETURN_ERR_IF_NOT(T, "Failed to get input tensor.");
 
     bool isPartial = partialInputs_.count(P);
+    bool isInt64 = int64Inputs_.count(P);
```
jfix71: nit: `isInt64` could be named better -- something can be int64 but return false for `allows64To32Downcast()`, right? Maybe `convertInt64To32`

lib/Backends/Habana/Habana.cpp
```diff
@@ -126,8 +126,8 @@ class TensorHandle final {
   TensorHandle() = default;
 
   /// Constructor. Create a tensor from Glow IR Value \p V.
```
jfix71: nit: Doxygen looks stale -- NBD but maybe good to include something about `downcast64To32` since it's not trivial.

Issue #1830, opened 2018-10-10
Optimization flags on nodes


bertmaher:
This is more of a wishlist-type idea, but here goes: other compilers I've seen have some "optimization properties" on a per-instruction basis that let the optimizer work in terms of those flags, rather than explicitly checking node types everywhere.

I've noticed one concrete use case (there may be others), which is sinking transposes.  Many different node types allow transpose to sink past them: BatchNorm, ReLU, Sigmoid, etc. basically just grep GraphOptimizer.cpp for "sink transpose".  We could have a node flag that says "can sink transpose".  (I haven't thought through details -- maybe each node type really is a special snowflake).

This would've been useful to me recently when adding a new backend-specific node type that would benefit from transpose sinking.

jfix71:
I've noticed this for Max and RELU. Most optimizations that apply to one apply to the other. Right now we essentially need to replicate all of those optimizations for both of them.

opti-mix:
@bertmaher Do you want to have such a flag per instruction kind or per instruction instance?

bertmaher:
I was thinking it would be per instruction kind (at least that's what makes sense for the transpose one).  There could be other properties that make more sense per instance, though, but I don't have examples.

nadavrot:
Could this be a good bootcamp task?

nadavrot:
I think that this could be a good task for people who are new to the project.

sergeigofman:
It's not only about transpose sinking but also about ReLU sinking below Concat. For the backends that enjoy fusing ReLU with preceding Convolution, sinking ReLU behind Concat is really bad.

PR #3190, opened 2019-06-27
[OpenCL] Optimize BatchedReduceAdd implementation

SplitInfinity:
**Description**
This commit reduces the time taken to execute a `batchedreduceadd`
instruction in the OpenCL backend by moving the transfer of the input
and output slice size data from execution time to `addNetwork` time.

The slice sizes of the input and output can be computed using static
shape information at compile time, so they don't have to be computed and
transferred once per operator invocation at runtime. This commit introduces
a post-lowering `OCLBackend` transformation that replaces `BatchedReduceAddNode`
with a semantically identical `OCLBatchedReduceAddNode` that has two additional
inputs for the slice sizes of the input and output nodes. The code to
compute these slice sizes has been moved from `OpenCLFunction::execute`
to `OCLBackend::transformPostLowering` and modified to write into the
payload `Tensors` of `Constants` that are used as the inputs to the
previously mentioned `OCLBatchedReduceAddNode`. These slice sizes are then
copied to the device with the rest of the constants needed by the
function.

**Test Plan**
All unit tests pass.

DLRM trace before this optimization:
Screen Shot 2019-06-27 at 2 57 36 PM

DLRM trace after this optimization:
Screen Shot 2019-06-27 at 2 57 18 PM

Time taken to process one minibatch has decreased by 20%.

### DIFF
--- a/lib/Backends/OpenCL/Transforms.cpp
+++ b/lib/Backends/OpenCL/Transforms.cpp
@@ -33,12 +33,13 @@ bool OCLBackend::transformPostLowering(Function *F,
 
   LOG_SCOPE(F->getLogContext(), "OCLBackend::transformPostLowering")
 
-  if (cctx.compMode == CompilationMode::Train)
-    return false;
-
   bool changed = false;
   for (auto &node : F->getNodes()) {
     if (auto *CN = dyn_cast<ConvolutionNode>(&node)) {
+      if (cctx.compMode == CompilationMode::Train) {
+        continue;
+      }
+
       // TODO: OpenCL fast convolution kernel itself has some issue with group >
       // 1, which will be investigated later. So far, if the group > 1, we just
       // call the slow convolution kernel.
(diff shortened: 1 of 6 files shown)


ksaurabh-cadence:
I am planning to upstream non-OpenCL reduceAdd support in a day or two. Should I expect a conflict?

SplitInfinity:
@ksaurabh-cadence 

No, I am not going to modify non-OpenCL `BatchedReduceAdd` as part of this or any PRs in the immediate future.

SplitInfinity:
I think this PR has the right idea, but I'm not so sure about the approach. `OCLBatchedReduceAdd`'s two `Constant` inputs contain what is essentially shape information for its inputs and outputs. If some optimization pass changes the input and output shapes without updating these `Constants`, the resultant graph will be incorrect. One way to defend against this is to add checks in `OCLBatchedReduceAdd::verify`, but that seems like a bandaid. The shape information shouldn't be duplicated in the first place.

jfix71:
>If some optimization pass changes the input and output shapes without updating these Constants, the resultant graph will be incorrect.

At least for now, I wouldn't be too worried about that -- currently backend-specific Nodes are complete black boxes*. They are never touched except for DCE/CSE. So I think only the backend itself would be able to make valid changes here that impact shapes, as otherwise they would also need to update the backend-specific Nodes which they don't understand.

*If we end up adding functionality mentioned in #1830 then we may need to revisit this.

lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -50,17 +50,92 @@ bool OCLBackend::transformPostLowering(Function *F,
...
+      for (ssize_t i = batchDims.size() - 1; i >= 0; --i) {
+        // If i is the reduce axis, currentSliceSize is the slice size at the
+        // reduce axis. Store it in axisSliceSize and not in batchSliceSizes. If
+        // not, do the opposite.
+        if (i == axis) {
+          axisSliceSize = currentSliceSize;
+        } else {
+          batchSliceSizesH.at({j--}) = currentSliceSize;
+        }
+
```
opti-mix: Remove the empty line.

lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -50,17 +50,92 @@ bool OCLBackend::transformPostLowering(Function *F,
...
     if (auto *PAN = dyn_cast<AvgPoolNode>(&node)) {
+      if (cctx.compMode == CompilationMode::Train)
+        continue;
+
       auto *NR = convertPoolToNCHWPool<AvgPoolNode, OCLAvgPoolNode>(PAN, F);
       PAN->getResult().replaceAllUsesOfWith(NR);
       changed = true;
       continue;
     }
+    if (auto *BRA = dyn_cast<BatchedReduceAddNode>(&node)) {
```
opti-mix: Could you add a comment before this `if` describing what it does, i.e. it replaces Glow's BatchedReduceAddNode by an OpenCL specific OCLBatchedReduceAddNode, because ...

opti-mix:
> I think this PR has the right idea, but I'm not so sure about the approach. OCLBatchedReduceAdd's two Constant inputs contain what is essentially shape information for its inputs and outputs. If some optimization pass changes the input and output shapes without updating these Constants, the resultant graph will be incorrect. One way to defend against this is to add checks in OCLBatchedReduceAdd::verify, but that seems like a bandaid. The shape information shouldn't be duplicated in the first place.

I think `OCLBatchedReduceAdd::verify` approach makes sense for the time being.

A totally different approach could be to run a backend specific pass (not post-lowering) that would perform the rewriting at the very end of  the graph processing pipeline or even as an IR pass after the usual IR generation. In this case nobody can overwrite its results and make them inconsistent.

SplitInfinity:
- Addressed comments
- Added `OCLBatchedReduceAddNode::verify()` implementation

SplitInfinity:
`How do you test it? Do we need to add any new tests?`

We had existing tests for this operator for reducing on axis 0 and axis 1 as well as producing a zero-dimensional result. I added another test in #2958 for axis 2 (which in that specific test case is also the last dimension), so I think we have enough tests.

tools/ClassGen/Backends/OpenCL/OpenCLSpecificNodesVerification.h
```diff
@@ -30,4 +30,65 @@ bool OCLConvolutionNode::verify() const {
 bool OCLAvgPoolNode::verify() const { return true; }
 
 bool OCLMaxPoolNode::verify() const { return true; }
+
+bool OCLBatchedReduceAddNode::verify() const {
```
opti-mix: Can we use `expectCompareTrue` and other helpers to make this verifier not only return false in case of failures, but produce some user-friendly error messages?

SplitInfinity:
- Updated `verify()` to use `expectCompareTrue`.

PR #3192, opened 2019-06-27
[compilationLog]: Load module log context into function log context after setting its parent.

ZchiPitt:
Summary:
Load module log context into function's LogContext as setting its parent.

Test Plan:

ninja check


nickgg:
Can we get a little more info on this (e.g why?). Does this fix a bug? What situation does doing this help?

ZchiPitt:
@nickgg 
It's a fix. Previously, this functionality is inside LogContext constructor, where the parent is not set (which makes this part of code never executed).

PR #3193, opened 2019-06-28
[Loader] Allow loading uint8 for fused sls and avoid weights duplication.

rdzhabarov:
Summary:
Data for fused sls could be huge (GBs), avoid sparse lookup table copies for fused slss.
Support for uint8 rowwise quantized non fused slss.

According to C2 docs:
* Non fused version of SLS has uint8 data with scales and offsets as a separate params.
* Fused version has uint8 data with 8 additional bytes per row (4 bytes scale, 4 bytes offset).

Documentation:
n/a

Test Plan:
CI/Internal tests for models.

### DIFF
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -91,8 +91,9 @@ createAndSetTensorType(const caffe2::TensorProto &in) {
     result.t->reset(ElemKind::Int32ITy, dim);
   } else if (in.data_type() == caffe2::TensorProto::INT64) {
     result.t->reset(ElemKind::Int64ITy, dim);
-  } else if (in.data_type() == caffe2::TensorProto::UINT8 ||
-             in.data_type() == caffe2::TensorProto::INT8) {
+  } else if (in.data_type() == caffe2::TensorProto::UINT8) {
+    result.t->reset(ElemKind::UInt8QTy, dim, 1.0, 0);
+  } else if (in.data_type() == caffe2::TensorProto::INT8) {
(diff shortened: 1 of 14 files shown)


lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1440,23 +1434,23 @@ llvm::Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
...
+      const std::string &str = dict["values"]->strings().Get(0);
 
-      // We're loading unsigned data into Int8QTy, so we use OFFSETSHIFT to
-      // convert to signed.
-      size_t i;
-      for (i = 0; i < str.size(); i++) {
-        TH.raw(i) = (uint8_t)str.c_str()[i] - OFFSETSHIFT;
+      size_t pos;
+      for (pos = 0; pos < str.size(); pos++) {
+        TH.raw(pos) = (uint8_t)str.c_str()[pos];
```
opti-mix: @rdzhabarov Is it guaranteed that str.c_str() does not allocate anything? I'd suggest hoisting `std.c_str()` out of the loop, just in case.

bertmaher: Why use `c_str` at all here?  It seems like `(uint8_t)str[pos]` is equivalent.

rdzhabarov: copy pasted from above. But yeah, @bertmaher suggestion should work.

rdzhabarov:
I did with loading prod size model.

PR #3195, opened 2019-06-28
Clone function for SparseLengthsSumTest stress test

bertmaher:
Summary:
Originally this test relied on compiling the same function for
different backends, which is a bit fragile.  But clone works now thanks to some
cleanup in the backend.

Differential Revision: D16046818


bertmaher:
Oh, damn, no, I meant to only do that for testing :-(

Issue #2814, opened 2019-04-29
Heterogeneous partitioning example code


nickgg:
Heterogeneous partitioning (where the runtime has devices of different types and operator support levels) is a common ask from users and vendors. We are adding support in #2687, and we should add a example code that demonstrates how to use it effectively.

In an existing example (resnet-runtime?) or as a new one demonstrate how to set up a system that contains one device with limited operator support and a CPU backend which handles any remaining operators.

beicy:
#3197 provides the test code. Will add doc as well.

PR #3197, opened 2019-06-28
[Partitioner] Heterogeneous partition testing

beicy:
Summary:
Provide an easier way for using heterogeneous partition : user can define which ops not running on which type of backends.
related to #2814 
Documentation:

[Optional Fixes #issue]

Test Plan:
added unittest;
```
wangm-mbp:buildR wangm$ ./bin/image-classifier tests/images/imagenet/cat_285.png -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data -cpu-memory=100000 -load-device-configs="tests/runtime_test/heterogeneousConfigs.yaml" 
Model: resnet50
Running 1 thread(s).
tests/runtime_test/heterogeneousConfigs.yaml
I0628 16:12:13.164857 43896832 Partitioner.cpp:1038] The number of partitions is : 3, and the DAG is dumped into DAG.dot file.
I0628 16:12:13.165649 43896832 Partitioner.cpp:58] Writing dotty graph for DAG after graph partitioning: DAG.dot
I0628 16:12:13.165989 43896832 Partitioner.cpp:1047] 	 Partition 0:
		 Name :	resnet50_part1_part1
		 BackendKind :	CPU
		 Memory :	94929408
		 LogicalDeviceIDs :	1
I0628 16:12:13.166002 43896832 Partitioner.cpp:1047] 	 Partition 1:
		 Name :	resnet50_part2_part1
		 BackendKind :	Interpreter
		 Memory :	409600
		 LogicalDeviceIDs :	2
I0628 16:12:13.166026 43896832 Partitioner.cpp:1047] 	 Partition 2:
		 Name :	resnet50_part3_part1
		 BackendKind :	CPU
		 Memory :	8208200
		 LogicalDeviceIDs :	0
 File: tests/images/imagenet/cat_285.png	Label-K1: 285 (probability: 0.5823)
```

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -89,6 +89,15 @@ public:
   /// \returns the name of backend that powers this Device.
   llvm::StringRef getBackendName() { return config_.backendName; }
 
+  /// \returns a string with \p name in parameters.
+  llvm::StringRef getParamByName(llvm::StringRef name) const {
+    auto it = config_.parameters.find(name);
+    if (it != config_.parameters.end()) {
+      return it->second;
+    }
+    return "";
+  }
+
   /// \returns the maximum memory (in bytes) available on the device.
   virtual uint64_t getMaximumMemory() const = 0;
 
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -34,6 +34,10 @@ struct BackendInfo {
   uint64_t memSize;
   /// Backend pointer.
   Backend *backend = nullptr;
+  /// The non-supported nodes kind.
+  std::set<Kinded::Kind> nonSupportedNodesKinds;
+  /// The supported nodes kind.
+  std::set<Kinded::Kind> supportedNodesKinds;
 };
 
 /// A mapping of newly-created functions along with a set of nodes sets. The
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -51,6 +51,14 @@ struct DeviceInfo {
   uint64_t availableMemory;
   /// Backend Type.
   std::string backendName;
+  /// A string contains the node names(e.g. Add, Div) which are separeted by
+  /// ",". E.g. "Div,Add". In Partitioner, those nodes won't be supported in
+  /// this backend.
+  std::string nonSupportedNodes;
+  /// A string contains the node names(e.g. Add, Div) which are separeted by
+  /// ",". E.g. "Div,Add". In Partitioner, the complementary set of those nodes
+  /// won't be supported in this backend.
+  std::string supportedNodes;
   /// Available SRAM capacity in bytes.
   uint64_t sramCapacity;
   /// Peak compute on device in ops/second. Assumes all ops are in int8.
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -912,7 +912,29 @@ Partitioner::backendBasedPartition(Function *F,
   for (auto &N : F->getNodes()) {
     for (auto &backend : backends) {
       // Find the first backend that supports this node. The order of backends
-      // is important.
+      // is important. The check flow is :
+
+      // Step 1: If a node is in pre-defined non-supported nodes set, it can not
+      // be assigned to this backend. Continue.
+      const auto &nonSupportedNodesKinds =
+          backendMap_[backend->getBackendName()].nonSupportedNodesKinds;
+      if (nonSupportedNodesKinds.count(N.getKind())) {
+        // This op is on the pre-defined non-supported op list:
+        continue;
+      }
+      // Step 2: If the pre-defined supported nodes set is empty, it means all
+      // nodes could be assigned to this backend. If the pre-defined supported
+      // nodes set is not empty, we check that if the node from Step 1 is in
+      // this set or not. If not, continue.
+      const auto &supportedNodesKinds =
+          backendMap_[backend->getBackendName()].supportedNodesKinds;
+      if (!supportedNodesKinds.empty() &&
+          !supportedNodesKinds.count(N.getKind())) {
+        // This op is not on the pre-definded supported op list:
+        continue;
+      }
+      // Step 3: Check if the node is actually supported in this backend, if so,
+      // assign it to this backend and break. Otherwise continue.
       // TODO: the logic here need to be improved.
       if (backend->shouldLower(&N) || backend->isOpSupported(N)) {
         // Put this node into a partition for this backend.
@@ -981,6 +1003,10 @@ void Partitioner::getBackendMap(
       // is the same.
       // TODO : will improve the algorithm for different memory size.
       backendInfo.memSize = deviceInfo_[i].availableMemory;
+      backendInfo.nonSupportedNodesKinds =
+          generateNodeKindsSet(deviceInfo_[i].nonSupportedNodes);
+      backendInfo.supportedNodesKinds =
+          generateNodeKindsSet(deviceInfo_[i].supportedNodes);
       if (hasBackends) {
         backendInfo.backend = backends_[i];
       } else {
--- a/lib/Partitioner/PartitionerUtils.cpp
+++ b/lib/Partitioner/PartitionerUtils.cpp
@@ -286,4 +286,17 @@ GraphMemInfo getGraphMemInfo(const NodesSet &nodes) {
   return ret;
 }
 
+std::set<Kinded::Kind> generateNodeKindsSet(llvm::StringRef names) {
+  std::set<Kinded::Kind> nodeKindsSet;
+  llvm::StringRef::size_type pos = names.find(',');
+  while (pos != llvm::StringRef::npos) {
+    nodeKindsSet.insert(getKindFromNodeName(names.substr(0, pos)));
+    names = names.substr(pos + 1);
+    pos = names.find(',');
+  }
+  if (!names.empty()) {
+    nodeKindsSet.insert(getKindFromNodeName(names));
+  }
+  return nodeKindsSet;
+}
 } // namespace glow
--- a/tests/runtime_test/heterogeneousConfigs.yaml
+++ b/tests/runtime_test/heterogeneousConfigs.yaml
@@ -1,0 +1,15 @@
+---
+- name:     Device1
+  backendName: CPU
+  parameters: |
+    "nonSupportedNodes":"AvgPool"
+    "deviceID" : "0"
+- name:     Device2
+  backendName: CPU
+  parameters: |
+    "nonSupportedNodes":"AvgPool"
+- name:     Device3
+  backendName: Interpreter
+  parameters: |
+    "supportedNodes":""
+...
--- a/tests/unittests/PartitionerTest.cpp
+++ b/tests/unittests/PartitionerTest.cpp
@@ -154,14 +154,14 @@ TEST_F(PartitionerTest, Basic1) {
   CompilationContext cctx;
   auto err = myPartitioner.Partition(cctx);
   EXPECT_FALSE(errToBool(std::move(err)));
-  DAGListTy myList = std::move(myPartitioner.getPartitionResult());
+  DAGListTy dagList = std::move(myPartitioner.getPartitionResult());
   ASSERT_EQ(mod_.getFunctions().size(), 3);
-  ASSERT_EQ(myList.size(), 1);
+  ASSERT_EQ(dagList.size(), 1);
   ASSERT_TRUE(checkSaveNode(mod_));
 
   // Run the paritioned graph and compare the results.
   bindings_.allocate(mod_.getPlaceholders());
-  for (auto it = myList.begin(); it != myList.end(); ++it) {
+  for (auto it = dagList.begin(); it != dagList.end(); ++it) {
     bindings_.allocate(mod_.getPlaceholders());
     executeDAG((*it).root.get(), mod_, bindings_, {input}, {&in});
     Tensor test = res.clone();
@@ -229,12 +229,12 @@ TEST_F(PartitionerTest, Basic2) {
   CompilationContext cctx;
   auto err = myPartitioner.Partition(cctx);
   EXPECT_FALSE(errToBool(std::move(err)));
-  DAGListTy myList = std::move(myPartitioner.getPartitionResult());
+  DAGListTy dagList = std::move(myPartitioner.getPartitionResult());
   ASSERT_EQ(mod_.getFunctions().size(), 2);
-  ASSERT_EQ(myList.size(), 1);
+  ASSERT_EQ(dagList.size(), 1);
   ASSERT_TRUE(checkSaveNode(mod_));
 
-  for (auto &dag : myList) {
+  for (auto &dag : dagList) {
     for (auto &node : dag.nodes) {
       // Since saturateHost is set true, in this case, there should be 2 copys
       // of the partitions.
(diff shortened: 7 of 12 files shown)


opti-mix:
@beicy  Nice! Could you also add an option to specify the opposite? 

The following should mean that only AvgPool is supported on the specific backend and everything else should be considered not supported by the backend.
```
parameters: |
    "supportedOp":"AvgPool"
```

I hope adding this options should be trivial.

include/glow/Backends/DeviceManager.h
```diff
@@ -89,6 +89,18 @@ class DeviceManager {
   /// \returns the name of backend that powers this Device.
   llvm::StringRef getBackendName() { return config_.backendName; }
 
+  /// \returns a string contains the node name (e.g Add, Sub) that won't be
+  /// supported by this DeviceManger. It doesn't mean that those nodes kind are
+  /// not supported by this device. It just means user don't want these kind of
+  /// nodes running on this device.
+  llvm::StringRef getNonSupprtedNodes() const {
```
opti-mix: It is probably better to implement `getParamByName(llvm::StringRef name)` to make it more reusable. And then you can just call it like `getParamByName("nonSupportedOp")`

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -272,4 +272,16 @@ GraphMemInfo getGraphMemInfo(const NodesSetTy &nodes) {
   return ret;
 }
 
+std::set<Kinded::Kind> generateNonSupportedNodes(llvm::StringRef names) {
+  std::set<Kinded::Kind> ret;
+  while (names.find(',') != std::string::npos) {
+    std::pair<llvm::StringRef, llvm::StringRef> newSplit = names.split(',');
```
opti-mix: Probably not very important, but could it be that split allocates new strings? Please have a look at the similar function for splitting comma-separated input here:

https://llvm.org/doxygen/CommandLine_8cpp_source.html#l00607

include/glow/Partitioner/Partitioner.h
```diff
@@ -38,6 +38,10 @@ struct BackendInfo {
   uint64_t memSize;
   /// Backend pointer.
   Backend *backend = nullptr;
+  /// The non-supported nodes kind.
+  std::set<Kinded::Kind> nonSupportedNodesKind;
```
opti-mix: You store multiple kinds in the set, right? So, should it be called `nonSupportedNodesKinds`? Same for `supportedNodesKinds`?

lib/Partitioner/Partitioner.cpp
```diff
@@ -914,7 +914,40 @@ Partitioner::backendBasedPartition(Function *F,
...
+      // Step 3: Check if the node is actually supported in this backend, if so,
+      // assign it to this backend and now we find the first backend that
+      // supports this node (The order of backends is important) and break.
+      // Otherwise goto Step 4.
+
+      // Step 4: Continue to check if the node can be
+      // assigned to the next backend.
+
+      // Step 1:
+      auto nonSupported =
```
opti-mix: `auto nonSupportedNodesKinds`?

lib/Partitioner/Partitioner.cpp
```diff
@@ -914,7 +914,40 @@ Partitioner::backendBasedPartition(Function *F,
...
+          backendMap_[backend->getBackendName()].nonSupportedNodesKind;
+      if (nonSupported.find(N.getKind()) != nonSupported.end()) {
+        // This op is on the pre-definded non-supported op list:
+        continue;
+      }
+      // Step 2:
+      auto supported =
```
opti-mix: `supportedNodesKinds`?

lib/Partitioner/Partitioner.cpp
```diff
@@ -914,7 +914,40 @@ Partitioner::backendBasedPartition(Function *F,
...
+          backendMap_[backend->getBackendName()].supportedNodesKind;
+      if (supported.size() > 0 &&
```
opti-mix: How about a simpler form?

```
if (!supprted.empty() && supported.count(N.getKind())) {
  // Continue.
}
```

lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -286,4 +286,16 @@ GraphMemInfo getGraphMemInfo(const NodesSetTy &nodes) {
   return ret;
 }
 
+std::set<Kinded::Kind> generateNodeKindsSet(llvm::StringRef names) {
+  std::set<Kinded::Kind> ret;
```
opti-mix: ret -> nodeKindsSet?

tests/runtime_test/heterogeneousConfigs.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- name:     Device1
+  backendName: CPU
+  parameters: |
+    "nonSupportedNodes":"AvgPool"
```
opti-mix: Can you add `supportedNodes` example as well, because it needs to be tested?

include/glow/Partitioner/Partitioner.h
bertmaher: Unless order matters here, prefer `llvm::DenseSet` or maybe `std::unordered_set`.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -51,6 +51,14 @@ struct DeviceInfo {
   uint64_t availableMemory;
   /// Backend Type.
   std::string backendName;
+  /// A string contains the node names(e.g. Add, Div) which are seperated by
```
bertmaher: sp: separated

lib/Partitioner/Partitioner.cpp
```diff
@@ -914,7 +914,40 @@ Partitioner::backendBasedPartition(Function *F,
   for (auto &N : F->getNodes()) {
     for (auto &backend : backends) {
       // Find the first backend that supports this node. The order of backends
-      // is important.
+      // is important. The check flow is :
+
+      // Step 1: If a node is in pre-defined non-supported nodes set, it can not
```
bertmaher: Maybe it's just my preference but here I would find it easier to read the "step N" comments inline with the code, rather than as a separate block that I need to scroll up to reference.  And that way we could get rid of the "step N" tags themselves, which have to be kept up to date over time.

lib/Partitioner/Partitioner.cpp
bertmaher: Probably want `const auto &nonSupported = ...` here, or this is going to invoke copy assignment.

lib/Partitioner/Partitioner.cpp
```diff
@@ -914,7 +914,40 @@ Partitioner::backendBasedPartition(Function *F,
...
+      // Otherwise goto Step 4.
+
+      // Step 4: Continue to check if the node can be
+      // assigned to the next backend.
+
+      // Step 1:
+      auto nonSupported =
+          backendMap_[backend->getBackendName()].nonSupportedNodesKind;
+      if (nonSupported.find(N.getKind()) != nonSupported.end()) {
+        // This op is on the pre-definded non-supported op list:
```
bertmaher: sp: `pre-defined`

lib/Partitioner/PartitionerUtils.cpp
bertmaher: /me wishes for `folly::split` (but of course we don't want to pull in folly, alas)

tests/runtime_test/heterogeneousConfigs.yaml
beicy: The "Device3" in this file has the "supportedNodes" params. Since there are many ops in resnet50, I just let Interpreter supports all. In unit test, we tested the supported nodes set.

lib/Partitioner/Partitioner.cpp
beicy: Done!

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
-    ASSERT_EQ(myList.size(), 1);
-    ASSERT_TRUE(checkSaveNode(mod_));
-
-    for (auto &dag : myList) {
-      for (auto &node : dag.nodes) {
-        // Although the saturateHost is set true, no saturating the host in
-        // heterogeneous partiton.
-        ASSERT_EQ(node->logicalDevices.size(), 1);
-      }
+  createSimpleModule(mod_);
```
opti-mix: I'm not sure what exactly you test here. You set up the backends, but you do not expect anything specific from partitions. Or do you?

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+  std::vector<DeviceInfo> devices = {{3072, "Interpreter"},
+                                     {3072, "Interpreter"},
+                                     {3072, "CPU"},
+                                     {3072, "CPU"}};
+  auto partitioner =
+      Partitioner(&mod_, devices, backends, /* saturateHost */ true);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy myList = std::move(partitioner.getPartitionResult());
```
opti-mix: Can you give this variable a more meaningful name than `myList`?  ;-)

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+  createSimpleModule(mod_);
+  std::vector<DeviceInfo> devices = {{3072, "Interpreter", "Mul"},
+                                     {3072, "Interpreter", "Mul"},
+                                     {3072, "CPU", "Sub"},
+                                     {3072, "CPU", "Sub"}};
+  auto partitioner = Partitioner(&mod_, devices);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy myList = std::move(partitioner.getPartitionResult());
```
opti-mix: Can you give this variable a more meaningful name than myList? ;-)

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+      // Although the saturateHost is set true, no saturating the host in
+      // heterogeneous partiton.
+      ASSERT_EQ(node->logicalDevices.size(), 1);
+    }
+  }
+  mod_.clear();
+}
+
+/// Test pre-defined non-supported ops used for choosing backend in
+/// Heterogeneous Partition.
```
opti-mix: It is not obvious to me from the code and comments below how the non-supported ops are tested.

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+TEST_F(PartitionerTest, ConfigHeterogeneousPartitioning1) {
```
opti-mix: Could you give this test a more meaningful name reflecting what it does?

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+      ASSERT_EQ(node->logicalDevices.size(), 1);
     }
-    mod_.clear();
   }
+  mod_.clear();
+}
+
+/// Test pre-defined supported ops used for choosing backend in
+/// Heterogeneous Partition.
+TEST_F(PartitionerTest, ConfigHeterogeneousPartitioning2) {
```
opti-mix: Could you give this test a more meaningful name reflecting what it does?

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+      // heterogeneous partiton.
+      ASSERT_EQ(node->logicalDevices.size(), 1);
     }
-    mod_.clear();
   }
+  mod_.clear();
+}
+
+/// Test pre-defined supported ops used for choosing backend in
+/// Heterogeneous Partition.
```
opti-mix: It is not obvious to me from the code and comments below how the supported ops are tested.

tests/unittests/PartitionerTest.cpp
```diff
@@ -526,40 +526,91 @@ static void createSimpleModule(Module &mod) {
...
+  std::vector<DeviceInfo> devices = {
+      {3072, "Interpreter", "Mul", "Sub,Add,Save"},
+      {3072, "Interpreter", "Mul", "Sub,Add,Save"},
+      {3072, "CPU", "Sub", "Mul,Add,Save"},
+      {3072, "CPU", "Sub", "Mul,Add,Save"}};
+  auto partitioner = Partitioner(&mod_, devices);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy myList = std::move(partitioner.getPartitionResult());
```
opti-mix: Can you give this variable a more meaningful name than myList? ;-)

tests/unittests/PartitionerTest.cpp
beicy: Those backends are mock backends which are subclass of "Interpreter" or "CPU".  They overload the isOpsupported() function to eliminate some nodes for Heterogeneous testing. Previously, we didn't have such "nonSupportedNodesKinds" nor "supportedNodesKinds" config. Therefore, we need to write mock backends for unittest test.

tests/unittests/PartitionerTest.cpp
beicy: The config yaml file is only loaded in "loader" flow (I added resnet50 test in tests/images/run.sh to test loading and parsing the yaml file). For Partitioner unnitest, we just pass the pre-defined nodes list to DeviceInfo used in Partitioner. Added comments in the tests.

lib/Partitioner/Partitioner.cpp
```diff
@@ -981,6 +1003,10 @@ void Partitioner::getBackendMap(
       // is the same.
       // TODO : will improve the algorithm for different memory size.
       backendInfo.memSize = deviceInfo_[i].availableMemory;
+      backendInfo.nonSupportedNodesKinds =
+          generateNodeKindsSet(deviceInfo_[i].nonSupportedNodes);
```
opti-mix: Maybe `std::move(generateNodeKindsSet(deviceInfo_[i].nonSupportedNodes))` to avoid copying the set?

tests/unittests/PartitionerTest.cpp
```diff
@@ -525,41 +525,124 @@ static void createSimpleModule(Module &mod) {
...
+      Partitioner(&mod_, devices, backends, /* saturateHost */ true);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy dagList = std::move(partitioner.getPartitionResult());
+  ASSERT_EQ(mod_.getFunctions().size(), 3);
+  ASSERT_EQ(dagList.size(), 1);
+  ASSERT_TRUE(checkSaveNode(mod_));
+
+  int numOfInterpreter = 0;
```
opti-mix: The names sounds odd. Maybe `numOfInterpreterNodes` and `numOfCPUNode`?

Could you also check that it is exactly the kinds of nodes (or the nodes) you expected to be on CPU and Interpreter respectively?

tests/unittests/PartitionerTest.cpp
```diff
@@ -525,41 +525,124 @@ static void createSimpleModule(Module &mod) {
...
+                                     {3072, "CPU", "Sub"}};
+  auto partitioner = Partitioner(&mod_, devices);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy dagList = std::move(partitioner.getPartitionResult());
+  ASSERT_EQ(mod_.getFunctions().size(), 3);
+  ASSERT_EQ(dagList.size(), 1);
+  ASSERT_TRUE(checkSaveNode(mod_));
+  int numOfInterpreter = 0;
```
opti-mix: Same comment here about the names.

PR #3199, opened 2019-06-29
Add load balance based partitioning to Glow

nrsatish:
Summary: Add a load balance based partitioner that uses the roofline performance to load balance nodes across partitions. At this point, the number of partitions created is equal to the number of devices.

Differential Revision: D16047615

### DIFF
--- a/lib/Partitioner/PartitionerUtils.cpp
+++ b/lib/Partitioner/PartitionerUtils.cpp
@@ -156,6 +156,20 @@ uint64_t getOutMemPerNode(const NodesSetTy &nodes, const Node *node) {
   return ret;
 }
 
+/// Given a node, \return the NodeSet of all nodes that create the results
+/// for any of the inputs of this node (i.e. input of inputs)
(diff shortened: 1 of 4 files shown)


lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -156,6 +156,20 @@ uint64_t getOutMemPerNode(const NodesSetTy &nodes, const Node *node) {
...
 }
 
+/// Given a node, \return the NodeSet of all nodes that create the results
+/// for any of the inputs of this node (i.e. input of inputs)
+NodesSetTy getInputs(const Node *node) {
+  NodesSetTy result;
+  for (size_t i = 0, e = node->getNumInputs(); i < e; i++) {
+    Node *input = node->getNthInput(i).getNode();
+    Storage *in = llvm::dyn_cast<Storage>(input);
+    if (!in) {
```
bertmaher: Could shorten this a bit as `if (!llvm::isa<Storage>(input))` but no big deal either way.

PR #3200, opened 2019-07-01
[NFC][Partitioner] Minor API clean-ups

bertmaher:
Summary:
Everywhere this is used we just do a map lookup, so better not to
expose a getter for the map.  (The exception is testing the map size in a unit
test, but I don't think that's a necessary test)

Reviewed By: beicy

Differential Revision: D16074968

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -25,9 +25,9 @@ namespace glow {
 
 using namespace runtime;
 
-using MemUsageMapTy = std::unordered_map<Node *, uint64_t>;
-using ComputeTimeMapTy = std::unordered_map<Node *, float>;
-using PartitionCostMapTy = llvm::DenseMap<Function *, GraphMemInfo>;
+using MemUsageMap = std::unordered_map<Node *, uint64_t>;
+using ComputeTimeMap = std::unordered_map<Node *, float>;
+using PartitionCostMap = llvm::DenseMap<Function *, GraphMemInfo>;
 
 /// Data structure that contains the info for each type of backend used for
 /// partitioning.
@@ -54,10 +54,10 @@ struct FunctionNameComparator {
     return strcmp(lhs->getName().data(), rhs->getName().data()) < 0;
   }
 };
-using FunctionToNodesMapTy =
-    std::map<Function *, NodesSetTy, FunctionNameComparator>;
+using FunctionToNodesMap =
+    std::map<Function *, NodesSet, FunctionNameComparator>;
 
-using FunctionToBackendNameMapTy =
+using FunctionToBackendNameMap =
     std::map<Function *, std::string, FunctionNameComparator>;
 
 class NodeToFunctionMap {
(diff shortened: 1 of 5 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -1056,14 +1054,14 @@ llvm::Error Partitioner::loadBalancedPartitioning(Function *F,
...
-  std::vector<NodesSetTy> nodesInPartitions(numDevices);
+  std::vector<NodesSet> nodesInPartitions(numDevices);
   std::vector<GraphMemInfo> graphMem(numDevices, GraphMemInfo{});
   std::vector<Function *> partitions(numDevices);
 
   // Compute total roofline time
   float totalRooflineTime = 0;
   for (auto &n : F->getNodes()) {
-    totalRooflineTime += getComputeTime()[&n];
+    totalRooflineTime += getComputeTime(&n);
```
rdzhabarov: It seems that there is a case where Node `n` is not present in the compute map and now it's crashing vs creating element in a map with default ctor and returning result here.
Wondering if this now exposes some hidden bug.

bertmaher: Thanks for raising this.  I did actually intend it, since I don't think it's desirable behavior to get `0` for looking up a missing node.  But thinking a bit more, I'm not actually certain what's desirable here.  I'm kind of leaning towards assert+default value, so it won't kill a release build, but would fail during development?

bertmaher:
The getComputeTime diff made me realize an unsettling property of the computeTime/memUsage maps: their keys are actually deleted nodes.  I think they refer to Nodes in the original Function, which is (I think?) destroyed by the partitioner.

lib/Partitioner/Partitioner.cpp
nrsatish: @rdzhabarov  - did you encounter a failure case here where the crash happens?

rdzhabarov: @nrsatish not under normal circumstances, but it's quite possible and would be hard crash.

include/glow/Partitioner/Partitioner.h
```diff
@@ -317,10 +315,10 @@ class Partitioner {
   void dumpDAG(llvm::StringRef dotFilename) const;
 
   /// Get function for computeTime_
-  ComputeTimeMapTy getComputeTime() const { return computeTime_; }
+  float getComputeTime(Node *N) const { return computeTime_.at(N); }
 
   /// Get function for memUsage_
-  MemUsageMapTy getMemUsage() const { return memUsage_; }
+  uint64_t getMemUsage(Node *N) const { return memUsage_.at(N); }
```
rdzhabarov: nit: const Node here and line 318.

bertmaher: It looks like I can't use a `const Node *` to look something up in a `map<Node *, ...>`.  Not without a `const_cast`, anyways.

PR #3201, opened 2019-07-02
[NFC][Runtime] Add DCHECKs to check that callbacks are not null

### LINKED ISSUE #3081: [runtime] Add checks for all callbacks in the runtime
Add DCHECKs throughout the runtime to check that all callbacks are present. There shouldn't be any cases when a callback function isn't provided.

SplitInfinity:
**Description**
This commit sprinkles `DHECK_NOTNULL` macros over the Glow runtime stack
to check whether any of the callbacks passed around within the runtime
are null. This should never happen, but these `DCHECKS` will help debug if
it ever does.

**Test Plan**
`ninja check` -> all tests pass.

**Fixes**
This pull request fixes #3081.

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -64,6 +64,8 @@ DeviceInfo CPUDeviceManager::getDeviceInfo() const {
 void CPUDeviceManager::addNetworkImpl(const Module *module,
                                       FunctionMapTy functions,
                                       ReadyCBTy readyCB) {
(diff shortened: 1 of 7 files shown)


SplitInfinity:
Build failures are from the results `DCHECK_NOTNULL` not being used. Should I `(void)` everything or switch to `DCHECK` [as suggested by glog](https://github.com/google/glog/blob/master/src/glog/logging.h.in#L1071)?

rdzhabarov:
@SplitInfinity consider

> DCHECK(readyCB != nullptr);

I'd imagine that should not produce any warnings/errors

PR #3202, opened 2019-07-02
[pytorch] Run resnet

jackm321:
Summary:
Run PyTorch Resnet/Resnext models on Glow except for AdaptiveAvgPool and FC

Documentation:
doxygen

Test Plan:
added operator tests

`python setup.py test`

to run example:
`python setup.py develop --run_cmake --release`
then
`python examples/resnet_example.py`

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -15,6 +15,9 @@
  */
 
 #include "PyTorchModelLoader.h"
+
+#include "glow/Support/Support.h"
+
 #include <torch/csrc/jit/ir.h>
 
 namespace {
@@ -127,7 +130,7 @@ PyTorchModelLoader::loadValue(const torch::jit::Value *value) {
   auto ptType = value->type()->cast<at::CompleteTensorType>();
   auto glowType = ptTypeToGlowType(*ptType.get());
   glow::Placeholder *ph =
-      mod_.createPlaceholder(&glowType, "input", /*isTrainable*/ false);
+      mod_->createPlaceholder(&glowType, "input", /*isTrainable*/ false);
   addGlowNodeValue(value, ph->getOutput());
   return ph;
 }
@@ -167,6 +170,7 @@ void PyTorchModelLoader::loadAdd(const torch::jit::Node *ptNode) {
   const auto scalarHandle = getGlowConstantHandle<int32_t>(inputs[2]);
   assert(scalarHandle.size() == 1);
   assert(scalarHandle.raw(0) == 1);
+  (void)scalarHandle;
 
   glow::NodeValue rhs = getGlowNodeValue(inputs[0]);
   glow::NodeValue lhs = getGlowNodeValue(inputs[1]);
@@ -185,6 +189,7 @@ void PyTorchModelLoader::loadSub(const torch::jit::Node *ptNode) {
   const auto scalarHandle = getGlowConstantHandle<int32_t>(inputs[2]);
   assert(scalarHandle.size() == 1);
   assert(scalarHandle.raw(0) == 1);
+  (void)scalarHandle;
 
   glow::NodeValue rhs = getGlowNodeValue(inputs[0]);
   glow::NodeValue lhs = getGlowNodeValue(inputs[1]);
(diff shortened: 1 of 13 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -407,9 +583,17 @@ void PyTorchModelLoader::load() {
     auto *save = f_->createSave("save", outputNodeValue);
     outputPlaceholders_.push_back(save->getPlaceholder());
   }
+
+  static int32_t dagNum = 0;
+  f_->dumpDAG(glow::strFormat("GlowFunction%d.dot", dagNum++));
```
jackm321: will remove this

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -46,12 +46,19 @@ class PyTorchModelLoader {
   /// Mapping from PyTorch Values to Glow NodeValues created during loading.
   std::unordered_map<const torch::jit::Value *, glow::NodeValue> valueMap_;
```
putivsky: What if the same input used more than one time, does it mean Value* will be the same? May by Value::uniqueName is a better key?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -339,60 +495,79 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+      [this](const torch::jit::Node *node) { return loadBatchNorm(node); };
+
+  nodeLoaderMapping_[at::Symbol::fromQualString("aten::max_pool2d")] =
+      [this](const torch::jit::Node *node) { return loadMaxPool2d(node); };
+}
+
+void PyTorchModelLoader::loadNode(const torch::jit::Node *node) {
+  auto kind = node->kind();
+  assert(nodeLoaderMapping_.count(kind));
+  return nodeLoaderMapping_.at(kind)(node);
```
putivsky: Double search:

```
  auto it = nodeLoaderMapping_.find(kind);
  assert(it != nodeLoaderMapping_.end());
  return it->second(node);
```

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -273,16 +292,144 @@ void PyTorchModelLoader::loadConvolution(const torch::jit::Node *ptNode) {
...
-      "conv_input_transposed", conv->getResult(), NHWC2NCHW);
+      "conv_output_transposed", conv->getResult(), NHWC2NCHW);
 
   addGlowNodeValue(outputs[0], output->getResult());
 }
 
+void PyTorchModelLoader::loadBatchNorm(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 9);
```
rdzhabarov: DCHECK everywhere?

torch_glow/src/PyTorchModelLoader.cpp
jackm321: The double search only occurs in an assert, I'd prefer to keep it this way if you don't mind.

torch_glow/src/PyTorchModelLoader.h
jackm321: > What if the same input used more than one time
Unless I misunderstand your question, it should be fine. glow::NodeValue can also be used multiple times.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: I will go through in a follow up and change things to DCHECK or llvm::Error now that things are more settled.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -167,6 +170,7 @@ void PyTorchModelLoader::loadAdd(const torch::jit::Node *ptNode) {
   const auto scalarHandle = getGlowConstantHandle<int32_t>(inputs[2]);
   assert(scalarHandle.size() == 1);
   assert(scalarHandle.raw(0) == 1);
+  (void)scalarHandle;
```
zrphercule: Wonder what is this for?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -298,6 +445,9 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
     t.getHandle<int32_t>().raw(0) = to32Bit(val);
   } else if (iVal.isIntList()) {
     const auto vals = iVal.toIntListRef();
+    if (vals.empty()) {
+      return;
```
zrphercule: shouldnt we have some msg indicate an empty list is being used here?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -339,60 +495,83 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
-  case at::aten::sub:
-  case at::aten::relu:
-  case at::aten::_convolution:
-    return true;
-  default:
-    return false;
-  }
-  return false;
+  static std::once_flag flag;
+  std::call_once(flag, [&]() {
```
zrphercule: this chunk of fancy code has already been put in "Rui's note: how to write graceful C++"

torch_glow/src/PyTorchModelLoader.cpp
jackm321: scalarHandle is only used in `assert()` so it will look as though it's unused when compiled in Release so this makes it so the compiler doesn't complain about that.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: > graceful C++
If it exists, I don't think this is it 😛

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -209,6 +214,20 @@ void PyTorchModelLoader::loadConvolution(const torch::jit::Node *ptNode) {
   assert(inputs.size() == 12);
   assert(outputs.size() == 1);
 
+  // Node inputs:
```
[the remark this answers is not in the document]
zrphercule: Then we will have different enum for each operators, wonder if it worth...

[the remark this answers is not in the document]
jfix71: If we're already going to list out the values in a comment for each operator anyway then IMO there's no real downside, since it'll be the same number of lines of code. And each enum would be contained to the scope of the operator's load function (e.g. `loadConvolution()`).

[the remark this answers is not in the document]
jackm321: Interesting idea, I had to use old school enum class (class with an anonymous enum in it) so that the values don't leak into the function and I don't need to static_cast everything. Kinda weird but I thinks it's still better.

PR #3207, opened 2019-07-03
Autodetect number of devices

bertmaher:
Summary:
I want to make it possible to build a single Glow library
(particularly, libOnnxifi) that can support multiple hardware backends.
Currently the target backend for Onnxifi is statically configured via the
GLOW_WITH_TARGET macros, which means that if you compile libOnnxifi for Habana,
you can't use it on a non-Habana machine.

This diff adds a really simple device detection API to DeviceManager that lets
you query how many devices of a specific type are available.  For CPU I've used
`hardware_concurrency()` and for Habana, I'm parsing `/proc/bus/pci/devices`.

Using that API, Onnxifi can ask at runtime whether Habana chips are available,
and fall back to CPU if not.

Differential Revision: D16110785

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -54,6 +54,9 @@ public:
   /// Create a device manager based on the device config \p config.
   static DeviceManager *createDeviceManager(const DeviceConfig &config);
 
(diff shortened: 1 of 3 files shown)


rdzhabarov:
Some of the comments were added in the internal diff. LGTM otherwise.

PR #3209, opened 2019-07-04
Fix optimizeConversions: consider cases where cast changes value

### LINKED ISSUE #3183: [Bug] optimizeConversions is too optimistic about what casts are NOOP
optimizeConversions unconditionally performs the following transform:
```
T1 X = ...;
T1 Y = T1(T2(X));

replace Y with X
```
It does not consider 2 important cases when this replacement is invalid:
- `T1` is a floating-point type, `T2` is not. In this case, the conversion may drop fractional part and the entire thing is not a NOOP.
- `T1` is a wider type than `T2` (e.g. `T1` is int64 and `T2` is int32). In this case, cast may also lose significant senior bits.

In both this cases, the said replacement should be disabled. 
The bug was fund while I was adding more checks to test of https://github.com/pytorch/glow/pull/3179/. Working on fix.

Max-K-cadence:
Summary:
The optimizing pass optimizeConversions assumes that the following operation
is always NOOP:

  T1 X = ...;
  T1 Y = T1(T2(X));

and replaces Y with X. In fact, there are at least two cases when such
replacement is invalid:
  - if T1 is a floating-point type and T2 is not, T2(X) may change the value by
    dropping the fractional part;
  - if mantissa of T2 has less significant bits than mantissa of T1, this cast
    may lead to loss of senior significant bits.

In both cases, the said replacement is non-equivalent. Justification that we can
can still perform this transform despite the sematic change because we eliminate
precision loss does not hold, because transform like float -> int -> float
can be done intentionally to strip away fractional part, and this transform
removes a semantically important piece.

This patch fixes that situation.

Fixes #3183

Test Plan:
Added unit test.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -2096,13 +2096,67 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
   }
 }
 
+/// Compute number of significant bits that are used to represent data of type
+/// \p kind. For FP, it is the number of bits in mantissa, for integers it's the
+/// number of bits except sign bit.
+/// \p returns number of significant bits of \p kind.
+/// TODO: Currently, for all supported types wider mantissa also means wider
+/// exponent. If we add a type for which this is not true, we should check both
+/// mantissa and exponent.
+static size_t numSignificantBits(ElemKind kind) {
+  switch (kind) {
+  case ElemKind::BoolTy:
+    return std::numeric_limits<bool>::digits;
+  case ElemKind::Int8QTy:
+    return std::numeric_limits<int8_t>::digits;
+  case ElemKind::UInt8QTy:
+  case ElemKind::UInt8FusedQTy:
+    return std::numeric_limits<uint8_t>::digits;
+  case ElemKind::Float16Ty:
+    // Custom type with layout 0 00000 0000000000.
+    return 10;
+  case ElemKind::Int16QTy:
+    return std::numeric_limits<int16_t>::digits;
+  case ElemKind::FloatTy:
+    return std::numeric_limits<float>::digits;
+  case ElemKind::Int32QTy:
+  case ElemKind::Int32ITy:
+    return std::numeric_limits<int32_t>::digits;
+  case ElemKind::Int64ITy:
+    return std::numeric_limits<int64_t>::digits;
+  default:
+    // Avoid compiler warning.
+    break;
+  }
+  llvm_unreachable("Unknown type!");
+}
+
+/// Returns true if casting value from \p srcTy to \p destTy may change it. As
+/// implication of this, casting value from \p srcTy to \p destTy and back may
+/// produce different value than before cast.
+static bool isValueChangingCast(TypeRef srcTy, TypeRef destTy) {
+  // FP-to-Int conversion may lead to loss of fraction, so it's not NOOP.
+  if (srcTy->isFPType() && !destTy->isFPType()) {
+    return true;
+  }
+  // Narrowing transform (e.g. int64 to int32) may lead to loss of
+  // significant senior bits, so it's not NOOP.
+  ElemKind srcElKind = srcTy->getElementType();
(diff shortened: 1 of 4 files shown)


opti-mix:
@Max-K-cadence Nice! The overall ideas seems pretty reasonable to me!

@rdzhabarov @jfix71 Could you have a look?

rdzhabarov:
Sorry for the delay @Max-K-cadence. Will do a pass over opened PRs today.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2055,9 +2055,42 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
...
+    return std::numeric_limits<int16_t>::digits;
+  case ElemKind::FloatTy:
+    return std::numeric_limits<float>::digits;
+  case ElemKind::Int32QTy:
+  case ElemKind::Int32ITy:
+    return std::numeric_limits<int32_t>::digits;
+  case ElemKind::Int64ITy:
+    return std::numeric_limits<int64_t>::digits;
+
+  default:
```
jfix71: nit: Let's leave off default case here. That way we will be sure to update this when we add new ElemKinds, as the switch won't cover the whole enum and cause compilation to fail.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2055,9 +2055,42 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
...
+    llvm_unreachable("Unknown type");
+  }
+  return 0;
```
jfix71: nit: Don't think we will need to return here, since the switch will cover all possible cases.

Max-K-cadence:
> I'm wondering about future ElemKinds -- let's say we add support for bfloat16. Compared to FP16, it has less mantissa bits but more exponent bits. I don't think it's safe to convert in either direction, as one direction loses precision but the other loses range.

The same problem already exists between int and float. I didn't accout for this type of problems, but we can take smth like float 10^10, turn it to int and back and also lose precision. :( I'm not sure what the valid behavior should be in this case. Let's add a TODO to think about it later.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
Max-K-cadence: We still need to have something here (at least llvm_unreachable), otherwise the compiler complains that control reaches the end of non-void function.

jfix71: Hmm ok, I think we do this in many other places but it's NBD.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2087,13 +2087,54 @@ static NodeValue convertConstant(Module &mod, Constant &constant,
   }
 }
 
+/// Compute number of significant bits that are used to represent data of type
+/// \p kind. For FP, it is the number of bits in mantissa, for integers it's the
+/// number of bits except sign bit.
+/// \p returns number of significant bits of \p kind.
+static size_t numSignificantBits(ElemKind kind) {
```
jfix71: Please add a TODO as discussed to more correctly handle this in the future as we add more ElemKinds.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -1814,6 +1814,46 @@ TEST(onnx, importCastToInt64) {
   importCast("castToInt64.onnxtxt", "data", {1, 2, 2, 2}, ElemKind::Int64ITy);
 }
 
+TEST(onnx, cast_32_64) {
```
jfix71: Is an ONNX importer test necessary here? I think a GraphOptzTest would be easier to add and upkeep, and more directly tests this PR. Also, this test does not test the float->int path, nor verify int->float works fine. Please add more coverage.

Max-K-cadence:
Added Operator tests to make sure we cover all cases.

Issue #3153, opened 2019-06-20
How to support TensorFlow operators that cannot be easily converted to ONNX?


Max-K-cadence:
Hi,

Sorry for long wall of text, I just want to give a comprehensive description of a problem I have and a solution I see.

I have a conceptual question about how we can support nodes that are not described in ONNX format, but are described in TensorFlow format and there is no easy way to convert one to another. Specifically, I am working with a model that has ScatterNd node which is described here:

https://www.tensorflow.org/api_docs/python/tf/scatter_nd

Conceptually, ScatterNd is pretty similar to Glow's ScatterAssign node, with two significant differences:

1) ScatterAssign only assigns `(N-1)`-dimensional slices to `N`-dimensional dest by `1`-dimensional indices. ScatterNd may assign `K`-dimensional slices to `N`-dimensional dest by `(N-K)`-dimensional indices;
2) ScatterNd allows duplication of indices. If some index is seen twice, their updates are summed.

Otherwise it does the same thing as ScatterAssign. Due to property 1), it is more generic. Along with property 2), ScatterNd is virtually a generalized version of "Scatter add" to a new destination.

There are three possible ways how we can bring support of ScatterNd in glow:

_Approach 1: support it in tf2onnx converter_
There is an ongoing discussion about it here: https://github.com/onnx/tensorflow-onnx/issues/455
Unfortunately, there is no simple mapping between ScatterNd and anything that ONNX currently supports. I've been studying it myself and came to same conclusion as @chinhuang007: implementing this conversion is going to be a lot of pain. I don't really believe it's going to happen soon.

_Approach 2: add this node to ONNX standard_
There is also a discussion about adding this node to ONNX standard: https://github.com/onnx/onnx/issues/2074#issuecomment-500511404. There is some idea about mapping it to ONNX Scatter node, however the complexity of index conversion that needs to be done before it is very high. I don't see much progress on this way either.

_Approach 3: support ScatterNd as if it was an ONNX node, using TensorFlow specification_
We may just pretend that Approach 2 has worked, and ScatterNd is described in ONNX standard, and its description is a perfect match with that in TensorFlow. Then we can just support it in Glow.

[...]

PR #3210, opened 2019-07-04
Support multi-dimensional Scatter Assign and Add

Max-K-cadence:
Summary:

This patch generalizes ScatterAssign node. It makes it capable of setting (N-K)
dimensional values by K-dimensional indices to N-dimensional destination.
Without this patch, only K=1 is supported.

It also adds "Cumulative" flag to ScatterAssign instruction. If it is set to
True, then the instruction will add values from Updates to Data by specified
indices instead of copying.

ScatterAssign node is renamed to ScatterData (with optionally set Cumulative
flag) to better reflect its new functionality after changes.

Added support for Interpreter and CPU backends only. OpenCL support can be
added separately by request.

This patch is a preparation step to support of ScatterNd node.

Based on discussion #3153

Test Plan:
Added unit tests. NFCI for existing nodes.

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -425,13 +425,56 @@ static void libjit_gatherranges(T *output, U *lengths, const T *data,
 }
 
 template <typename T>
-static void libjit_scatterassign(T *data, const size_t *indices,
-                                 const T *slices, size_t numIndices,
-                                 size_t sliceSize) {
+static void libjit_scatterdatacopy(T *data, const size_t *dataDims,
+                                   const size_t *indices, const T *slices,
+                                   size_t numIndices, size_t indexSize,
+                                   size_t sliceSize) {
   for (size_t i = 0; i < numIndices; i++) {
-    size_t destDataIdx = indices[i];
+    size_t destDataIdx = indices[i * indexSize];
+    for (size_t j = 1; j < indexSize; j++) {
+      destDataIdx *= dataDims[j];
+      destDataIdx += indices[i * indexSize + j];
+    }
     memcpy(data + destDataIdx * sliceSize, slices + i * sliceSize,
            sliceSize * sizeof(T));
+  }
+}
+
+template <typename T>
+static void libjit_scatterdataaddfloat(T *data, const size_t *dataDims,
+                                       const size_t *indices, const T *slices,
+                                       size_t numIndices, size_t indexSize,
+                                       size_t sliceSize) {
+  for (size_t i = 0; i < numIndices; i++) {
+    size_t destDataIdx = indices[i * indexSize];
+    for (size_t j = 1; j < indexSize; j++) {
+      destDataIdx *= dataDims[j];
+      destDataIdx += indices[i * indexSize + j];
+    }
+    for (size_t j = 0; j < sliceSize; j++) {
+      data[destDataIdx * sliceSize + j] += slices[i * sliceSize + j];
+    }
+  }
+}
+
+template <typename T>
+static void libjit_scatterdataaddquantized(
+    T *data, const size_t *dataDims, const size_t *indices, const T *slices,
+    size_t numIndices, size_t indexSize, size_t sliceSize, float dataScale,
+    int32_t dataOffset, float sliceScale, int32_t sliceOffset) {
+
+  for (size_t i = 0; i < numIndices; i++) {
+    size_t destDataIdx = indices[i * indexSize];
+    for (size_t j = 1; j < indexSize; j++) {
+      destDataIdx *= dataDims[j];
+      destDataIdx += indices[i * indexSize + j];
+    }
+    for (size_t j = 0; j < sliceSize; j++) {
+      float lhs = (data[destDataIdx * sliceSize + j] - dataOffset) * dataScale;
+      float rhs = (slices[i * sliceSize + j] - sliceOffset) * sliceScale;
+      T result = libjit_clip((lhs + rhs) / dataScale + dataOffset);
+      data[destDataIdx * sliceSize + j] = result;
+    }
   }
 }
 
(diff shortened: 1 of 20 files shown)


Max-K-cadence:
Looks like some existing models need replacement of ScatterAssign node to ScatterData. What is the location of this /var/lib/jenkins/workspace/downloaded_models/en2gr?

jfix71:
>Looks like some existing models need replacement of ScatterAssign node to ScatterData. What is the location of this /var/lib/jenkins/workspace/downloaded_models/en2gr?

@Max-K-cadence You can find the model in our normal model download script, it's called `en2gr` inside `download_datasets_and_models.py`

Max-K-cadence:
I still have no idea what's going on in this script. :D It has nothing but reference to http://fb-glow-assets.s3.amazonaws.com/models/en2gr/ to which I have no access. What should I do to fix this model?

jfix71:
>I still have no idea what's going on in this script. :D It has nothing but reference to http://fb-glow-assets.s3.amazonaws.com/models/en2gr/ to which I have no access. What should I do to fix this model? Is it a part of some different repository?

Not sure what you mean by you don't have access, it should be public. What happens when you run the following?
```
python utils/download_datasets_and_models.py --caffe2-models en2gr
```
It should download all of the files you need for this model to a `en2gr/` directory, and then you should be able to run it via [the instructions here](https://github.com/pytorch/glow/blob/master/docs/Testing.md#text-translation).

Max-K-cadence:
My bad, I didn't realize that not the model needed amendment, but the code in Caffe2 loader. Sorry for bothering. :)

lib/Graph/Grad.cpp
```diff
@@ -319,9 +319,9 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       // to the correspondent Data Tensors, where Vector value
       // points to Data Tensor.
       auto *SN = G->createSplat("splat.grad", Data.getType(), 0);
-      auto *SA = new ScatterAssignNode("scatter.assign.grad", SN->getResult(),
-                                       RI->getResult(),
-                                       RG ? RG->getResult() : OutputG);
+      auto *SA = new ScatterDataNode(
+          "scatter.assign.grad", SN->getResult(), RI->getResult(),
+          RG ? RG->getResult() : OutputG, /*Add*/ false);
```
jfix71: tiny nit: In most other places you're calling it `/*Cumulative*/`. Don't care either way, just prefer to be consistent.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1361,25 +1361,140 @@ void BoundInterpreterFunction::fwdGatherRangesInst(
...
+  const size_t numSlices = slicesT->size() / slicesT->dims()[0];
 
+  auto IH = indicesT->getHandle<int64_t>();
   // For each index, copy from the slice at that index into the location in data
   // given the offset from the indices tensor.
-  for (size_t i = 0, end = indicesT->size(); i < end; i++) {
-    size_t destDataIdx = indicesT->getHandle<int64_t>().raw(i);
+  for (size_t i = 0, end = indicesT->dims()[0]; i < end; i++) {
+    size_t destDataIdx = 0;
+    if (indicesT->dims().size() == 1) {
```
jfix71: Can you add some comments in here about what's going on?

Max-K-cadence:
Thanks Jordan! I'll address the comments and fix the build (now I need to add this node to ONNX writer too).

lib/Backends/Interpreter/InterpreterNodes.cpp
Max-K-cadence: Hm, I think we can simplify it by always turning indices to 2-dimensional.

lib/Graph/Grad.cpp
Max-K-cadence: Thanks for pointing out! Yeah, I renamed this during development, and it seems that vestige of old name is still in code. I'll review it carefully.

Max-K-cadence:
- Addressed nits;
- Indices are now only 2D for uniformity reasons;
- Added stub to ONNX Exporter. ScatterData has no direct analog in ONNX, but once this one is checked in, I plan to come up with ScatterNd operation that corresponds to it.

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -425,16 +425,58 @@ static void libjit_gatherranges(T *output, U *lengths, const T *data,
...
-static void libjit_scatterassign(T *data, const size_t *indices,
-                                 const T *slices, size_t numIndices,
-                                 size_t sliceSize) {
+static void libjit_scatterdatacopy(T *data, const size_t *dataDims,
+                                   const size_t *indices, const T *slices,
+                                   size_t numIndices, size_t indexSize,
+                                   size_t sliceSize) {
   for (size_t i = 0; i < numIndices; i++) {
-    size_t destDataIdx = indices[i];
+    size_t destDataIdx = indices[i * indexSize];
```
jfix71: This seems like it's still generalized past 2D right? Maybe simplify it given we only support 2D?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1473,25 +1473,136 @@ void BoundInterpreterFunction::fwdGatherRangesInst(
...
+
+  auto IH = indicesT->getHandle<int64_t>();
+  // For each index, copy from the slice at that index into the location in data
+  // given the offset from the indices tensor.
+  for (size_t i = 0, end = indicesT->dims()[0]; i < end; i++) {
+    size_t destDataIdx = 0;
+    if (indicesT->dims().size() == 1) {
+      destDataIdx = IH.raw(i);
+    } else {
+      assert(indicesT->dims().size() == 2 &&
```
jfix71: nit: Can move this outside of the loop.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1473,25 +1473,136 @@ void BoundInterpreterFunction::fwdGatherRangesInst(
...
+             "Multi-dimensional index should be stored in 2D tensor!");
+      for (size_t j = 0, e = indicesT->dims()[1]; j < e; j++) {
+        destDataIdx *= dataT->dims()[j];
+        destDataIdx += IH.at({i, j});
+      }
+    }
+    auto D = dataT->getHandle<ElemTy>(), S = slicesT->getHandle<ElemTy>();
```
jfix71: nit: Can move this outside of the loop.

lib/Backends/CPU/libjit/libjit.cpp
Max-K-cadence: This code was initially was done for 2D, and it also worked for 1D as `indexSize` as computed as `indices->size() / numIndices`, and for 1D case it was 1. Now I made change in LLVM IR gen to clearly take these values from indices->dims()[0], indices->dims()[1] to make this logic more transparent. But I don't see how we can simplify code here. What's your idea?

Max-K-cadence:
Fixed assert for OpenCL.

Max-K-cadence:
I'm not sure if it's a real failure or not. I don't think I did something that could cause this bug after fixing OpenMP assert, but I've seen it twice already. :(

lib/Graph/Nodes.cpp
```diff
@@ -1290,32 +1290,30 @@ bool GatherRangesNode::verify() const {
   return isValid;
 }
 
-bool ScatterAssignNode::verify() const {
+bool ScatterDataNode::verify() const {
```
jfix71: Can you add a check here that data and slices should have the same element kind?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -1473,25 +1473,128 @@ void BoundInterpreterFunction::fwdGatherRangesInst(
...
   Tensor *dataT = getTensor(I->getData());
   Tensor *indicesT = getTensor(I->getIndices());
   Tensor *slicesT = getTensor(I->getSlices());
 
-  size_t dataSliceSize =
-      dataT->size() / dataT->dims()[0] * dataT->getType().getElementSize();
+  assert(indicesT->dims().size() == 2 &&
+         "Index should be stored in 2D tensor!");
+  const size_t dataSliceSize =
+      slicesT->size() / slicesT->dims()[0] * dataT->getType().getElementSize();
```
jfix71: tiny nit: Even though they should be the same, I'd use `slicesT` element size instead since we're using its dims/size now just so it's not so confusing when reading this code.

tests/unittests/OperatorTest.cpp
```diff
@@ -2494,6 +2494,238 @@ TEST_P(OperatorTest, ScatterAssignQuantized) {
...
+
+  auto *data = mod_.createPlaceholder(ElemKind::FloatTy, {5, 2}, "data", false);
+  auto *indices =
+      mod_.createPlaceholder(ElemKind::Int64ITy, {2, 1}, "indices", false);
+  auto *slices =
+      mod_.createPlaceholder(ElemKind::FloatTy, {2, 2}, "slices", false);
+
+  bindings_.allocate(data)->getHandle() = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+  bindings_.allocate(indices)->getHandle<int64_t>() = {1, 3};
+  bindings_.allocate(slices)->getHandle() = {-3, -4, -7, -8};
```
jfix71: Can you make `-8` positive so we verify that the kernels correctly handle maxing out the quantized range?

lib/Backends/CPU/libjit/libjit.cpp
jfix71: Nevermind I think it's good.

tests/unittests/OperatorTest.cpp
Max-K-cadence: Hm, looks like they don't. Good catch! I'll fix this.

Max-K-cadence:
Thanks for looking through it so thoroughly. :)

Issue #2986, opened 2019-05-28
[New Operator] Add support for Bucketize


jfix71:
See Caffe2 implementation [here](https://github.com/pytorch/pytorch/blob/master/caffe2/operators/bucketize_op.cc).

shajrawi:
I'm Interested in grabbing this one!

PR #3224, opened 2019-07-11
[AutoGen] Fix a bug in VectorFloat hashing

shajrawi:
The `getHash()` automatically generated function for nodes containing `VectorFloat` member type did not compile:

We called `hash_combine_range` on a standard vector of floats. This is not supported.

This PR adds proper hashing for VectorFloat types.

I have encountered this bug while investigating this issue: https://github.com/pytorch/glow/issues/2986

Looking at the Caffe2 implementation, there's a `std::vector<float>` member as can be seen in this header file: https://github.com/pytorch/pytorch/blob/master/caffe2/operators/bucketize_op.h

We did not encounter this before, so auto generation for `VectorFloat` has not been trialled.

### DIFF
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -214,6 +214,11 @@ struct NodeValue;
 /// FIXME: This is a workaround, because defining the hash_code
 /// hash_value(float) does not work for some reason.
 size_t toBinary(float f);
+/// Convert a collection of floats into a vector of
+/// unsigned integer binary representation.
+/// FIXME: This is a workaround, because defining the hash_code
+/// hash_value(float) does not work for some reason.
+std::vector<size_t> toBinary(llvm::ArrayRef<float> vec);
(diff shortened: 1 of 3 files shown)


shajrawi:
@jfix71 can you please review?

jfix71:
Nice find @shajrawi! Thanks for working on this. Functionality LGTM.

For consistency with the existing `toBinary()` and to limit code duplication in case we add more VectorFloats, what do you think about moving the `vector<float> -> vector<size_t>` conversion into a function in `Nodes.cpp`? Its signature could be something like `std::vector<size_t> toBinary(llvm::ArrayRef<float> vec);`, and called from the lambda in `getHash()`. WDYT?

shajrawi:
I like it! Much cleaner. Thanks for the input @jfix71 , I amended the commit accordingly.

opti-mix:
@shajrawi Congrats! You first Glow PR has landed! 🎊 🎆

PR #3226, opened 2019-07-12
[ONNXIFI] Support passing onnxifi backend through the GFLAG/LLVM param.

rdzhabarov:
Summary:
Support manual override of onnxifi backend.
This would allow flexibility when there is a need to use onnxifi with private backend.

Reviewed By: yinghai

Differential Revision: D16215878

### DIFF
--- a/lib/Onnxifi/onnxifiGlow.cpp
+++ b/lib/Onnxifi/onnxifiGlow.cpp
@@ -16,6 +16,7 @@
 
 #include "Base.h"
 #include "GlowOnnxifiManager.h"
+#include "llvm/Support/CommandLine.h"
 
 #include "glow/Importer/ONNXIFIModelLoader.h"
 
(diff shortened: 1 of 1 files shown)


lib/Onnxifi/onnxifiGlow.cpp
```diff
@@ -57,67 +70,50 @@ GLOW_ONNXIFI_LIBRARY_FUNCTION_WRAPPER(onnxGetBackendIDs)(
...
-        "Interpreter", /*useOnnx*/ true, /*forQuantization*/ true);
     auto *quantizationBackendC2 = manager.createBackend(
-        "Interpreter", /*useOnnx*/ false, /*forQuantization*/ true);
-
-    backendIDs[0] = quantizationBackendOnnx;
-    backendIDs[1] = quantizationBackendC2;
-  } else if (withCPU || withHabana) {
-    *numBackends = 4;
+        GlowOnnxifiBackend.empty() ? "Interpreter" : GlowOnnxifiBackend,
+        /*useOnnx*/ false, /*forQuantization*/ true);
```
bertmaher: So are we never going to use onnx for quantization profiling via onnxifi?

rdzhabarov: yup, no more use of onnx for inference or profiling.

rdzhabarov:
This was merged. Something was messed up with fbcode/github hook.

Issue #3213, opened 2019-07-09
Compile should take a Module, not a Function


jfix71:
Currently, we have the concept of compiling a Function using the EE. This does not align with e.g. the HostManager, wherein we compile a Module instead of a Function. Additionally, when compiling a Function we do optimizations such as DCE on Constants. This is not always safe, as other Functions may still be being built (see #3141) and so Constants shouldn't be deleted so early on.

In general we should compile Modules instead of Functions. We should also move optimizations like `DCE` for Constants out of `optimizeFunction()` and into a new `optimizeModule()` that's performed after all Functions in a Module are optimized.

CC: @gcatron, @opti-mix, @ksaurabh-cadence

shajrawi:
Interested in working on this, will do so after EE2 replaces EE in the repo, this is also related to #3228 , once we are compiling at a module level we should probably follow suit and do LLVM IR generation on a module level instead of LLVM Function level.

shajrawi:
Putting this one up for grabs since I've been working on some other tasks since graduating bootcamp :)

PR #3227, opened 2019-07-12
Added ExecutionEngine2 as the new runtime backed Engine

gcatron:
Summary:
This PR adds a new ExecutionEngine2 which uses Hostmanager. It is currently alongside the original EE. This allows for an incremental migration of unit tests from EE to EE2. Once everything is ported to the new EE we can remove the original EE and rename EE2. This PR also ports BackendTest which verifies the new EE2 is working.
Documentation: EE docs will be updated once migration is complete.

Related to #3213

Test Plan: builds and BackendTest runs successfully

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine2.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine2.h
@@ -1,0 +1,147 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_EXECUTIONENGINE_EXECUTIONENGINE2_H
+#define GLOW_EXECUTIONENGINE_EXECUTIONENGINE2_H
+
+#include "glow/Backend/Backend.h"
+#include "glow/Backend/CompiledFunction.h"
+#include "glow/Backends/DeviceManager.h"
+#include "glow/Base/Train.h"
+#include "glow/Base/Traits.h"
+#include "glow/Graph/Graph.h"
+#include "glow/Graph/PlaceholderBindings.h"
+#include "glow/Runtime/HostManager/HostManager.h"
+
+#include "llvm/ADT/ArrayRef.h"
+
+#include <memory>
+#include <unordered_map>
+
+namespace glow {
+
+/// This is the ExecutionEngine. It owns the Graph, the backend, and the
+/// compiled function.  The Graph, etc in this class are defined as pointers, in
+/// order to erase the type and prevent the internal types from leaking out to
+/// the users of this class.
+class ExecutionEngine2 final {
+  /// Module containing the function and supporting information.
+  std::unique_ptr<Module> module_;
+
+  /// Raw pointer to module_ this is to support module access after the module
+  /// has been added to hostManager_.
+  Module *rawModule_;
+
+  /// Name of the backend being used for compilation and execution.
+  std::string backendName_ = "";
+
+  /// The HostManager for executing the compiled functions.
+  std::unique_ptr<runtime::HostManager> hostManager_;
+
+  /// Glow functions compiled for this ExecutionEngine's backend.
+  std::set<std::string> compiledFunctions_;
+
+  /// Single execution of the given \compiledFunction with the given context
+  /// \bindings.
+  void runInternal(ExecutionContext &context, llvm::StringRef name);
+
+public:
+  ExecutionEngine2(llvm::StringRef backend = "Interpreter");
+
+  ~ExecutionEngine2();
+
+  /// Set the code generator to \p backend. New code will be generated
+  /// using this backend. This clears all previously loaded functions and resets
+  /// the Module.
+  void setBackendName(llvm::StringRef backend);
+
+  /// Get a pointer to the backend.
+  llvm::StringRef getBackendName() const;
+
+  /// \returns the internal graph.
+  Module &getModule() { return *rawModule_; }
+
+  /// Clears the ExecutionEngine and all CompiledFunctions.
+  void clear();
+
+  /// \returns the DAG for the specified \p network.
+  llvm::Expected<runtime::DAG &> getDAG(llvm::StringRef network) {
+    return hostManager_->getNetworkDAG(network);
+  }
+
+  /// Compiles all functions in the Module with the given \p cctx.  This method
+  /// should be invoked before the run method and can only be called once
+  /// without resetting the backend.
+  void compile(CompilationContext &cctx);
+
+  /// A convenience function for the most common type of compile.
+  void compile(CompilationMode mode);
+
+  /// Context aware single execution of a function. If more than one
+  /// function has been compiled by this ExecutionEngine then a name must be
+  /// supplied to specify which function to run.
+  void run(ExecutionContext &context);
+
+  /// Context aware single execution of a function with the given \p
+  /// name.
+  void run(ExecutionContext &context, llvm::StringRef name);
+
+  /// Context aware single execution of a function. If more than one
+  /// function has been compiled by this ExecutionEngine then a name must be
+  /// supplied to specify which function to run.
+  void run(PlaceholderBindings &bindings);
+
+  /// Context aware single execution of a function with the given \p
+  /// name.
+  void run(PlaceholderBindings &bindings, llvm::StringRef name);
+};
+
+//===----------------------------------------------------------------------===//
+//         Helper methods for running the execution engine.
+//===----------------------------------------------------------------------===//
+
+/// This method updates the placeholders in \p ph with the tensor content
+/// values \p inputs, in \p bindings.
+void updateInputPlaceholders2(PlaceholderBindings &bindings,
+                              llvm::ArrayRef<Placeholder *> ph,
+                              llvm::ArrayRef<Tensor *> inputs);
+
+/// This method updates the placeholders in the module. The placeholders are
+/// found by name
+///  in \p ph with the tensor content values \p inputs.
+void updateInputPlaceholdersByName2(PlaceholderBindings &bindings, Module *mod,
+                                    llvm::ArrayRef<llvm::StringRef> ph,
+                                    llvm::ArrayRef<Tensor *> inputs);
+
+/// Runs \p iterations iterations of the compiled function. The method updates a
+/// global counter and future invocations of this method continue running
+/// iterations of the batch at the next available slice.
+///
+/// The method updates the placeholder in \p ph with the tensors \p inputs. The
+/// shape of the slice has to be identical to the shape of slices in the batch.
+/// All dimensions, except for the first (batch) dimension must be identical.
+///
+/// The variable \p sampleCounter is consumed and updated by the function. This
+/// variable records the number of samples that were consumed by the network in
+/// previous iterations. The next input to be loaded is
+/// (sampleCounter % batchsize).
+void runBatch2(ExecutionEngine2 &EE, PlaceholderBindings &bindings,
+               size_t iterations, size_t &sampleCounter,
+               llvm::ArrayRef<Placeholder *> ph,
+               llvm::ArrayRef<Tensor *> inputs, llvm::StringRef name = "");
+
+} // namespace glow
+
+#endif // GLOW_EXECUTIONENGINE_EXECUTIONENGINE2_H
--- a/lib/ExecutionEngine/ExecutionEngine2.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine2.cpp
@@ -1,0 +1,207 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


include/glow/ExecutionEngine/ExecutionEngine2.h
```diff
@@ -0,0 +1,163 @@
...
+#include <memory>
+#include <unordered_map>
+
+namespace glow {
+
+/// This is the ExecutionEngine. It owns the Graph, the backend, and the
+/// compiled function.  The Graph, etc in this class are defined as pointers, in
+/// order to erase the type and prevent the internal types from leaking out to
+/// the users of this class.
+class ExecutionEngine final {
```
nickgg: Using the same name for the ExecutionEngine2 means that this will have namespace collisions with the old one, effectively theres no way to have a test suite that uses either interchangeably. Consequences:
* Must migrate every test in a file at once, can't do piece by piece.
* New tests in a file that hasn't been migrated have to use the old EE, can't use any features of the new one.

Might be less total effort to actually name this `ExecutionEngine2` and code mod it back when removing the old one. Alternatively could move the old one to a new deprecated namespace so it's clear it's old?

include/glow/ExecutionEngine/ExecutionEngine2.h
```diff
@@ -0,0 +1,163 @@
...
+  /// Module containing the function and supporting information.
+  std::unique_ptr<Module> module_;
+
+  /// Raw pointer to module_ this is to support module access after the module
+  /// has been added to hostManager_.
+  Module *rawModule_;
+
+  /// The network execution backend.
+  Backend *backend_ = nullptr;
```
nickgg: This is misleading: this backend object is not used to do the compilation (or execution), or as far as I can see anything except retrieving the backend name to configure the HostManager. This is dangerous because if a future backend has modifiable state it will appear like you can modify then use it in the EE, but you actually can't.

An example would be if we have a backend that supports changing op support dynamically (e.g. cpu heterogenous testing), using getBackend() and setBackend() does not actually affect the compilation backend in the EE anymore.

Can you just remove this completely and replace with BackendName or DeviceConfig?

include/glow/ExecutionEngine/ExecutionEngine2.h
```diff
@@ -0,0 +1,163 @@
...
+
+  /// Whether or not the ExecutionEngine owns the backend or is just using
+  /// a backend provided from elsewhere. If ownsBackend is true,
+  /// ~ExecutionEngine will delete the backend_.
+  bool ownsBackend_ = false;
```
nickgg: again this is a bit misleading, wont the HostManager always create and destroy its own backend?

include/glow/ExecutionEngine/ExecutionEngine2.h
```diff
@@ -0,0 +1,163 @@
...
+
+  /// \returns the DAG for the specified \p network.
+  llvm::Expected<runtime::DAG &> getDAG(llvm::StringRef network) {
+    return hostManager_->getNetworkDAG(network);
+  }
+
+  /// \returns whether a node with the provided \p NI is supported by the
+  /// underlying backend.
+  bool isOpSupported(const NodeInfo &NI) const {
+    return backend_->isOpSupported(NI);
```
nickgg: E.g. this is an example of a call that can be wrong if this function is not static since a different Backend object is used for compile.

include/glow/ExecutionEngine/ExecutionEngine2.h
```diff
@@ -0,0 +1,163 @@
...
+/// compiled function.  The Graph, etc in this class are defined as pointers, in
+/// order to erase the type and prevent the internal types from leaking out to
+/// the users of this class.
+class ExecutionEngine final {
+  /// Module containing the function and supporting information.
+  std::unique_ptr<Module> module_;
+
+  /// Raw pointer to module_ this is to support module access after the module
+  /// has been added to hostManager_.
+  Module *rawModule_;
```
nickgg: Ick :( I know this is due to the existing interfaces and you don't have alternatives but this is a sign we don't want to use unique_ptr for module.

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -0,0 +1,216 @@
...
+    assert(t && "Invalid tensor.");
+    assert(p && "Invalid placeholder.");
+    updateInputPlaceholders(bindings, {p}, {t});
+  }
+}
+
+void ExecutionEngine::runInternal(ExecutionContext &context,
+                                  llvm::StringRef name) {
+  // Make sure that the bindings have backing tensors for all placeholders.
+  context.getPlaceholderBindings()->allocate(rawModule_->getPlaceholders());
```
nickgg: Should this be a call on HostManager rather than requiring we store a raw pointer to what we passed in? Alternatively, is it better to keep a copy of the full bindings, clone it and merge it?

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+  std::unique_ptr<ExecutionContext> contextPtr(&context);
```
nickgg: We should fix one of the interfaces so we don't have to pull tricks like this

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+void ExecutionEngine::compile(CompilationMode mode) {
+  CompilationContext cctx;
+  cctx.compMode = mode;
+  compile(cctx);
+}
+
+void ExecutionEngine::compile(CompilationContext &cctx) {
+  assert(module_.get() && "Compile has already been called.");
```
nickgg: So this call to compile() is terminal, in a way: once you call it you can't add any more functions or compile anything else until you setBackend() again? The old EE didn't work like that did it?

tests/unittests/BackendTest.cpp
```diff
@@ -53,23 +55,27 @@ TEST(Interpreter, profileQuantizationForANetwork) {
   Node *O = F->createFullyConnected(bindings, "fc", A, 4);
   O = F->createRELU("relu", O);
   O = F->createRegression("reg", O, Ex);
+  F->createSave("ret", O);
```
nickgg: just curious: why does the HM mean we need a save here? does the partitioner assume that all graphs have one?

tests/unittests/BackendTest.cpp
```diff
@@ -53,23 +55,27 @@ TEST(Interpreter, profileQuantizationForANetwork) {
...
-  EE.compile(F, cctx);
+  EE.compile(cctx);
+  bindings.allocate(mod.getPlaceholders());
 
   // TODO: Verify histogram itself, for now just verify min and max.
   // Run inference first time and capture tensor stats.
   updateInputPlaceholders(bindings, {A}, {&inputs});
   EE.run(bindings);
+  // Get the new function post compilation.
+  auto CF = mod.getFunctions().front();
```
nickgg: Why do we need to do this? What makes F invalid?

nit: name is confusing since despite being post compilation, this is not a CompiledFunction.

tests/unittests/BackendTest.cpp
```diff
@@ -122,8 +129,9 @@ TEST(RuntimeBundle, BundleSymbolInfo) {
   auto *S = F->createSave("ret", SM);
   auto *qp = F->createQuantizationProfile(bindings, "qp", input);
 
-  EE.compile(CompilationMode::Infer, F);
-  auto table = EE.getCompiledFunction().getRuntimeBundle().getSymbolTable();
+  EE.compile(CompilationMode::Infer);
+  auto dag = EE.getDAG("main");
+  auto table = dag->nodes[0]->runtimeBundle->getSymbolTable();
```
nickgg: nit: maybe assert nodes.size() > 0 first?

tests/unittests/BackendTest.cpp
```diff
@@ -295,8 +302,43 @@ TEST_P(BackendTest, debugPrint) {
   IRBuilder(IR.get()).createDebugPrintInst("print", *IR->getWeights().begin());
 
   auto function = backend->compileIR(std::move(IR));
-  EE_.insertCompiledFunction("main", std::move(function));
-  EE_.run(*ctx.get());
+
+  // Since we are compiling IR by hand we cannot go through the normal EE route.
```
nickgg: Not for this diff, but I could see an interface where we add a CompiledFunction to the HostManager being useful (use the single node dag builder in Partitioner, etc). That would have made this change really easy.

gcatron:
> I have questions!
> 
> * Why did you include the new EE2 alongside the EE rather than replacing it? Was it that the time/effort of replacing the old with the new in every test was too big for one diff?
> * If so, do you have something tracking the progress of the migration? (Maybe an issue with which test suites are done/not done?)
> * Do you need help from the team to do the migration or are you going to do it yourself? Either way, what's the time time for removing the deprecated ExecutionEngine?
> * I understand you're not updating the docs for this PR, but can you describe what the functional/interface difference is between EE and EE2 (ie. what we need to know to write tests now).

Great Questions! 

EE2 alongside is for an incremental change, I started out thinking they could all be done in one diff but I think that would be too large a diff and a pain to continually rebase. 

An issue tracking this sounds like a good idea, I'll post one and edit this comment with the link()

At this point I have ported all of the tests run with ninja check over, I still have to do recSys and a few others, with this approach the team is welcome to help but it may be more efficient for me to wrap up. 
Timeline should be a week or so, once this lands I'll be putting up PRs for most of the remaining tests.

Absolutely, I didn't have the doc written up and didn't want to delay getting feedback on the PR. 
High level difference is the the module is now owned by the EE and all functions in the EE are compiled at the same time with the same compilation options. A downside of this new EE is that in order to add new functions all previous functions must be cleared. In other tests this meant creating multiple EEs for training and inference. (Stay tuned for further documentation)

include/glow/ExecutionEngine/ExecutionEngine2.h
gcatron: Renaming to ExecutionEngine2 seems fine, I generally have been porting the whole file at a time but options are good.

tests/unittests/BackendTest.cpp
gcatron: So this could have been split into a separate PR, turns out without the Save node DCE wipes the whole function...

tests/unittests/BackendTest.cpp
gcatron: In this case F("main") no longer exists, partitioning created a new function "main_part1" can totally enhance the name and comments to explain this better.

tests/unittests/BackendTest.cpp
gcatron: That's a good idea, that way even if we aren't testing the compilation path, we would still be sharing a comment execute path. Wanna write up an issue to capture the thought?

include/glow/ExecutionEngine/ExecutionEngine2.h
gcatron: renamed

tests/unittests/BackendTest.cpp
nickgg: I cant imagine we're relying on that anywhere, should it be an error or something?

include/glow/ExecutionEngine/ExecutionEngine2.h
gcatron: replaced with backendName_

include/glow/ExecutionEngine/ExecutionEngine2.h
gcatron: Removed now

include/glow/ExecutionEngine/ExecutionEngine2.h
gcatron: also gone :)

lib/ExecutionEngine/ExecutionEngine2.cpp
gcatron: It did not, this is because the HostManager compilation is module wide and destructive.

lib/ExecutionEngine/ExecutionEngine2.cpp
gcatron: I think on the HostManager is a bit general, what do you think about just removing this call entirely and letting the caller ensure their bindings are properly allocated?

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -0,0 +1,209 @@
...
+  runInternal(context, name);
+}
+
+void ExecutionEngine2::run(PlaceholderBindings &bindings) {
+  assert(compiledFunctions_.size() == 1 &&
+         "Expected exactly one compiled function.");
+  std::unique_ptr<PlaceholderBindings> bindingsPtr(&bindings);
+  ExecutionContext context(std::move(bindingsPtr));
+  runInternal(context, *compiledFunctions_.begin());
+  // don't delete bindings
```
jfix71: nit: Capitalize and period.

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -0,0 +1,209 @@
...
+  context.movePlaceholderBindings().release();
+}
+
+void ExecutionEngine2::run(PlaceholderBindings &bindings,
+                           llvm::StringRef name) {
+  std::unique_ptr<PlaceholderBindings> bindingsPtr(&bindings);
+  ExecutionContext context(std::move(bindingsPtr));
+  runInternal(context, name);
+  // don't delete bindings
```
jfix71: nit: Capitalize and period.

tests/unittests/BackendTest.cpp
```diff
@@ -53,19 +55,24 @@ TEST(Interpreter, profileQuantizationForANetwork) {
...
+  EE.compile(cctx);
+  bindings.allocate(mod.getPlaceholders());
 
   // TODO: Verify histogram itself, for now just verify min and max.
   // Run inference first time and capture tensor stats.
-  updateInputPlaceholders(bindings, {A}, {&inputs});
+  updateInputPlaceholders2(bindings, {A}, {&inputs});
   EE.run(bindings);
+  // The partitioner deleted the original function and created a new one, get
+  // the new function.
```
beicy: very nit: I think the comment could be more accurate : " The partitioner **may** delete the original function " :)  Since it is possible that there is no partition and no function deletion.

tests/unittests/BackendTest.cpp
beicy: > I cant imagine we're relying on that anywhere, should it be an error or something?

If I understand your question correctly, in quantization profiling flow,  we need the updated function for generating quantization info.

PR #3231, opened 2019-07-12
Fixed a memory leak in Partitioner.

gcatron:
Summary: Fix a case where a Backend can be allocated and never freed. Wrapped the backend in a smart pointer.

Documentation: NA

Test Plan: Ninja check

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -1209,7 +1209,7 @@ llvm::Error Partitioner::QuantizationProfilingPartition(
   FunctionToBackendNameMap funcToBackend;
   funcToBackend = backendBasedPartition(F_, backends, cctx);
   module_->eraseFunction(F_);
-  auto backend = createBackend(profilingBackend);
+  std::unique_ptr<Backend> backend(createBackend(profilingBackend));
   for (Function *subF : module_->getFunctions()) {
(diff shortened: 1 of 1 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -1209,7 +1209,7 @@ llvm::Error Partitioner::QuantizationProfilingPartition(
   FunctionToBackendNameMap funcToBackend;
   funcToBackend = backendBasedPartition(F_, backends, cctx);
   module_->eraseFunction(F_);
-  auto backend = createBackend(profilingBackend);
+  std::unique_ptr<Backend> backend(createBackend(profilingBackend));
```
rdzhabarov: did we exercise this pass in tests? (it's very likely to be reported by asan)

gcatron: So I found it as a result of ASAN complaining when I went to land the BackendTest ported to executionEngine2, not sure how it slipped by before.

beicy: For quantization in Partitioner, the tests are added in ./tests/images/run.sh.  It is should be in CI, right?

nickgg: It is run in CI but not in ASAN.

PR #3232, opened 2019-07-12
[mnist] Fix uses of MaxPool without NodeValue

jfix71:
Summary: This was broken as a result of #3146

Test Plan: Ran mnist and it works again 🙂 

Reported in [PyTorch forums here](https://discuss.pytorch.org/t/glow-example-run-error/50461).

### DIFF
--- a/examples/mnist.cpp
+++ b/examples/mnist.cpp
@@ -95,11 +95,11 @@ void createModel(ExecutionEngine &EE, Function *F,
   auto *RL0 = F->createRELU("relu", CV0);
   auto *MP0 = F->createMaxPool("pool", RL0, 3, 3, 0);
 
-  auto *CV1 = F->createConv(bindings, "conv", MP0, 16, 5, 1, 2, 1);
+  auto *CV1 = F->createConv(bindings, "conv", MP0->getResult(), 16, 5, 1, 2, 1);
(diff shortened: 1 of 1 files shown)


examples/mnist.cpp
```diff
@@ -95,11 +95,11 @@ void createModel(ExecutionEngine &EE, Function *F,
   auto *RL0 = F->createRELU("relu", CV0);
   auto *MP0 = F->createMaxPool("pool", RL0, 3, 3, 0);
 
-  auto *CV1 = F->createConv(bindings, "conv", MP0, 16, 5, 1, 2, 1);
+  auto *CV1 = F->createConv(bindings, "conv", MP0->getResult(), 16, 5, 1, 2, 1);
```
rdzhabarov: qq: by default NodeValue is constructed with the output number 0 from the node. It seems like MaxPool has a result as input #0.

Did I miss something?

jfix71: Yeah so NodeValue will by default use output #0, but there's an assert to check that it's not a multi-output Node when doing so to make sure that we aren't accidentally using the wrong output. So it works correctly in release. #3146 added a new output to MaxPool, so all uses of it as a NodeValue needed to explicitly choose which NodeValue to use. The uses here were missed because this tested isn't on CI.

PR #3234, opened 2019-07-16
Added new BackendHints object 

gcatron:
Summary:

This PR adds a new BackendHints object and adds it to BackendOptions and DAGNode.
This PR also adds the plumbing for BackendHints created by the Partitioner to be piped through the Provisioner to the Backend. 
Currently there are two hints:
SRAM Pinning - An ordered list of Storage nodes to be pinned to SRAM. Order denotes priority.
Resource Reservation - An unsigned integer for the number of execution units to use for the function.

The plan is that new passes in the Partitioner will determine these hints and populate the backendHints field of the DAGNode for the function. This is then passed to the backend by the Provisioner. 

Documentation: Once the design is agreed on, Backends.md will need to be updated discussing the hints.

Related to #3235 

Test Plan: ninja test

### DIFF
--- a/include/glow/Backends/BackendOptions.h
+++ b/include/glow/Backends/BackendOptions.h
@@ -16,7 +16,22 @@
 #ifndef GLOW_BACKENDS_BACKENDOPTIONS_H
 #define GLOW_BACKENDS_BACKENDOPTIONS_H
 
+#include "llvm/ADT/SmallVector.h"
+#include <string>
+#include <vector>
+
 namespace glow {
+class Storage;
+
+/// Hints provided to the Backend, the backend is not required to honor them.
+struct BackendHints {
+  /// Number of execution units to reserve, these are the processing elements
+  /// like cores, 0 for unspecified.
+  unsigned executionUnits{0};
+
+  /// Storage nodes to be pinned to SRAM listed in order of priority.
+  std::vector<std::string> SRAMPrioritization;
+};
 
 /// Options relevant to Backends during compilation.
 struct BackendOptions {
@@ -25,6 +40,9 @@ struct BackendOptions {
 
   /// Insert TraceEvents between all instructions for profiling.
   bool autoInstrument{false};
+
+  /// Hints for the compiler for this compilation.
+  BackendHints backendHints;
 };
 
 }; // namespace glow
--- a/lib/Runtime/Provisioner/Provisioner.cpp
+++ b/lib/Runtime/Provisioner/Provisioner.cpp
@@ -82,6 +82,12 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
         << "Warning: collectConstants is set in a Runtime compile, "
            "ignoring it.";
   }
+  if (cctx.backendOpts.backendHints.SRAMPrioritization.size() != 0 ||
+      cctx.backendOpts.backendHints.executionUnits) {
+    VLOG_EVERY_N(1, 1000)
+        << "Warning: backendHints is set in a Runtime compile, "
+           "ignoring it.";
+  }
 
   // Set collectConstants to false, this is because the DeviceManager will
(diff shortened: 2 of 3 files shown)


include/glow/Backends/BackendOptions.h
```diff
@@ -16,7 +16,19 @@
...
 namespace glow {
+class Storage;
+
+/// Hints provided to the Backend, the backend is not required to honor them.
+struct BackendHints {
+  /// Number of execution units to reserve, 0 for unspecified.
+  unsigned executionUnits{0};
+
+  /// Storage nodes to be pinned to SRAM listed in order of priority.
+  llvm::SmallVector<Storage *, 0> SRAMPrioritization;
```
nrsatish: Can we design a way for this to be manually input to allow for experimentation before finalizing (just like with Partitioner?)

gcatron: Similar to user-defined partitioning? 
Would names make this easier to specify then?

bertmaher: Is there a reason for using an N=0 SmallVector?  Seems like that's not really different from a `std::vector`.

include/glow/Backends/BackendOptions.h
```diff
@@ -16,7 +16,19 @@
...
 #define GLOW_BACKENDS_BACKENDOPTIONS_H
 
+#include "llvm/ADT/SmallVector.h"
+
 namespace glow {
+class Storage;
+
+/// Hints provided to the Backend, the backend is not required to honor them.
+struct BackendHints {
+  /// Number of execution units to reserve, 0 for unspecified.
```
rdzhabarov: Can you add some details. What is "execution units to reserve"?
How is it used.

include/glow/Backends/BackendOptions.h
beicy: > Can we design a way for this to be manually input to allow for experimentation before finalizing (just like with Partitioner?)

That should be something in Partitioner.  After the Partitioner refactoring, I think it should be easy to add this feature in experimental flow.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -101,11 +101,14 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
       // compiled the function reuse it.
       auto it = functions_.find(node->name);
       if (it == functions_.end()) {
+        // Copy BackendOptions and add the compiler hints for this function.
+        auto options = cctx.backendOpts;
+        options.backendHints = node->backendHints;
```
nickgg: Putting this into the BackendOptions really makes me think I can configure it in the CompilationContext, but it is always overridden here. Can we be clearer about this? Maybe log an error if this is set already?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -101,11 +101,14 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
       // compiled the function reuse it.
       auto it = functions_.find(node->name);
       if (it == functions_.end()) {
+        // Copy BackendOptions and add the compiler hints for this function.
+        auto options = cctx.backendOpts;
```
nickgg: Why do we need a copy here? Is it to prevent the backendHints from leaking back out to the caller?

lib/Runtime/Provisioner/Provisioner.cpp
gcatron: Error log added.

include/glow/Backends/BackendOptions.h
gcatron: I updated the comment, it's a bit vague because each vendor may have a different name or way of subdividing a device. This would be like reserving cores on a CPU.

PR #3237, opened 2019-07-16
[Partitioner] User-defined partition

beicy:
Summary:
This PR added user-defined partition flow. Basically, a struct "PartitionConfig" containing the partition info is passed into Partitioner to enable this flow.  Now we let users have the full control of how to do the partitioning. 
To use this flow, users can write their helper function to generate  PartitionConfig, and call Partitioner directly.
In the following PR, we will add passing PartitionConfig through HostManager from a yaml file. 

Related to #2298
Documentation:

[Optional Fixes #issue] 

Test Plan:
Added unittest. ninja test.

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -204,10 +204,13 @@ class Partitioner {
   /// use all available devices.
   bool saturateHost_;
 
-  // Flag to set if the funcitons in the module are areadly optimized. By
-  // default, the optimization should be done in Partitioner due to
-  // heterogeneous partition.
+  /// Flag to set if the funcitons in the module are areadly optimized. By
+  /// default, the optimization should be done in Partitioner due to
+  /// heterogeneous partition.
   bool optimized_;
+
+  /// The struct contain user-defined partition info.
+  PartitionConfig partitionConfig_;
 
   /// Get the representative function (the one with the largest input) and
   /// update the memSize.
@@ -295,16 +298,25 @@ public:
   /// "Function Family", that is, without considerting the "dynamic stuff" (i.e.
   /// batch size, input/output shape of each op), all the functions are
   /// identical. The required memory and computation cost for each op can be
-  /// found in Module. The \p devices provides the cost model related to
-  /// devices.
+  /// found in Module.
+  /// The \p devices provides the cost model related to devices.
+  /// Saturating the host will be enabled if \p saturateHost is true.
+  /// \p optimized is false by default, which means the functions in this module
+  /// are not optimized. \p partitionConfig contains the user defined partition
+  /// info.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-              bool saturateHost = false, bool optimized = false);
+              bool saturateHost = false, bool optimized = false,
+              PartitionConfig partitionConfig = PartitionConfig());
 
   /// Users can create Mock Backends and pass their points to test Graph
   /// Partitioning without actually register them in GLOW.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
               const std::vector<Backend *> &backends, bool saturateHost = false,
               bool optimized = false);
+
+  /// Based on partitionConfig_ passed into Partitioner, do the user-defined
+  /// partition.
+  llvm::Error PartitionFromConfig();
 
   /// Decompose each function in a module. Now we support partitioning a module
   /// among different type of devices. \p cctx is used during optimization of
--- a/include/glow/Runtime/RuntimeTypes.h
+++ b/include/glow/Runtime/RuntimeTypes.h
@@ -171,6 +171,30 @@ struct HostConfig {
   size_t executorThreads{3};
 };
 
+/// This is struct for user defined partition.
+struct PartitionConfig {
+  /// The name of the function to be partitioned.
+  std::string funcName;
+  /// The number of user defined partitions.
+  /// The partition ids are between 0 and numOfPartitions - 1, inclusive.
+  size_t numOfPartitions;
+  /// The backend for each partition. backendNames.size() == numOfPartitions.
+  std::vector<std::string> backendNames;
+  /// The name for each partition. partitionNames.size() == numOfPartitions.
+  std::vector<std::string> partitionNames;
+  /// The mapping between nodes' name to Partition ids. Assume there are n nodes
+  /// and m partitions. We have 2 types of valid mapping: 1. all nodes are
+  /// mapped to a partition. 2. For i-th (0 <= i < m) partition, the nodes
+  /// mapped to this partition id are not in this map, and the nodes mapped to
+  /// other partitions ids must be in this map. The node's name should be the
+  /// name in Glow function and may be different from the original name from
+  /// models. Since Glow will mangle names to make them unique.
+  llvm::StringMap<size_t> nodeToPartition;
+
+  PartitionConfig() : numOfPartitions(0) {}
+  bool enabled() { return numOfPartitions > 0; }
+};
+
 } // namespace runtime
 } // namespace glow
 #endif // GLOW_RUNTIME_RUNTIMETYPES_H
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -22,11 +22,10 @@
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
-#include <fstream>
-
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <fstream>
 namespace glow {
 bool GlowEnableLoadBalancedPartitioning = false;
 static llvm::cl::opt<bool, /* ExternalStorage */ true>
@@ -60,6 +59,20 @@ using llvm::isa;
 bool sortMinMemory(const std::pair<Function *, uint64_t> &a,
                    const std::pair<Function *, uint64_t> &b) {
   return a.second < b.second;
+}
+
+static void dumpPartitionInfo(const NodeToFunctionMap &partitions) {
+  int i = 0;
+  for (Function *subF : partitions.getPartitions()) {
+    LOG(INFO) << "\t Partition " << i++ << ":\n"
+              << "\t\t Name :\t" << subF->getName().str() << "\n"
+              << "\t\t BackendKind :\t"
+              << partitions.getPartitionBackendName(subF) << "\n"
+              << "\t\t Memory :\t"
+              << partitions.getGraphMemInfo(subF).getTotalMemSize() << "\n"
+              << "\t\t LogicalDeviceIDs :\t"
+              << partitions.getLogicalDeviceIDList(subF)[0] << "\n";
+  }
 }
 
 void Partitioner::dumpDAG(llvm::StringRef dotFilename) const {
@@ -168,9 +181,10 @@ Partitioner::Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
 }
 
 Partitioner::Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-                         bool saturateHost, bool optimized)
+                         bool saturateHost, bool optimized,
+                         PartitionConfig partitionConfig)
     : module_(parent), deviceInfo_(devices), saturateHost_(saturateHost),
-      optimized_(optimized) {
+      optimized_(optimized), partitionConfig_(partitionConfig) {
   memSize_ = module_->getConstantsSize();
   logicalDeviceID_ = 0;
 }
@@ -1211,7 +1225,6 @@ llvm::Error Partitioner::QuantizationProfilingPartition(
   module_->eraseFunction(F_);
   std::unique_ptr<Backend> backend(createBackend(profilingBackend));
   for (Function *subF : module_->getFunctions()) {
-    (void)subF;
     assert(subF->verify() && "Conversion led to invalid function");
     if (!optimized_) {
       RETURN_IF_ERR(::glow::optimizeFunction(subF, *backend, cctx));
@@ -1230,6 +1243,11 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   std::vector<Backend *> backends;
   std::vector<std::unique_ptr<Backend>> backendHolder;
   getBackendMap(backendMap_, backendHolder, backends);
+
+  if (partitionConfig_.enabled()) {
+    // Jump into user-defined partition, and skip the following auto partition.
+    return PartitionFromConfig();
+  }
 
   // Step 0: Find the representative function for running partitioning
   // algorithm.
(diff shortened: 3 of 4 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1368,96 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+
+  // Do partition.
+  doPartitioning(F->getName(), {F}, partitionMap, true);
+  module_->eraseFunction(F);
+
+  // Do optimization based on backendName.
+  for (size_t i = 0; i < partitionConfig_.numOfPartitions; i++) {
+    auto func = funcList[i];
+    assert(func->verify() && "Conversion led to invalid function");
+    Backend *backend = createBackend(partitionConfig_.backendNames[i]);
```
rdzhabarov: backend is leaking

tests/unittests/PartitionerTest.cpp
```diff
@@ -886,3 +886,28 @@ TEST_F(PartitionerTest, memoryUsageValidation1) {
...
+  partitionConfig.numOfPartitions = 3;
+  partitionConfig.backendNames = {"Interpreter", "CPU", "Interpreter"};
+  partitionConfig.partitionNames = {"p1", "p2", "p3"};
+  partitionConfig.nodeToParitition = {{"sub", 0}, {"mul", 1}};
+  auto partitioner = Partitioner(&mod_, devices, false, false, partitionConfig);
+  CompilationContext cctx;
+  auto err = partitioner.Partition(cctx);
+  EXPECT_FALSE(errToBool(std::move(err)));
+  DAGListTy dagList = std::move(partitioner.getPartitionResult());
+  ASSERT_EQ(mod_.getFunctions().size(), 3);
```
rdzhabarov: those all can be expect_eq.

lib/Partitioner/Partitioner.cpp
```diff
@@ -52,6 +50,8 @@ static llvm::cl::opt<bool>
                   llvm::cl::desc("Enable dumping the graph of each partitions"),
                   llvm::cl::init(false), llvm::cl::cat(PartitionerCat));
 
+#include <fstream>
```
bertmaher: It doesn't look like `fstream` is used.  (If it is, put it with the other `#include`s above.

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,28 @@ struct HostConfig {
   size_t executorThreads{3};
 };
 
+/// This is struct is for user defined partition.
```
artemrakhov:
```suggestion
/// This is struct for user defined partition.
```

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,28 @@ struct HostConfig {
...
+struct PartitionConfig {
+  /// The name of the functiton to be partitioned.
```
artemrakhov:
```suggestion
  /// The name of the function to be partitioned.
```

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,28 @@ struct HostConfig {
...
+  std::string funcName;
+  /// The number of user defined partitions. The partition ids are [0,
```
artemrakhov:
```suggestion
  /// The number of user defined partitions.
  /// The partition ids are between 0 and numOfPartitions - 1, inclusive.
```

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,28 @@ struct HostConfig {
...
+  /// The backend for each partition. backendNames.size() == numOfPartitions.
+  std::vector<std::string> backendNames;
+  /// The name for each partition. partitionNames.size() == numOfPartitions.
+  std::vector<std::string> partitionNames;
+  /// The mapping between nodes and Partition ids. Assume there are n nodes and
+  /// m partitions. We have 2 types of valid mapping: 1. all nodes are mapped to
+  /// a partition. 2. For i-th (0 <= i < m) partition, the nodes mapped to this
+  /// partition id are not in this map, and the nodes mapped to other partitions
+  /// ids must be in this map.
+  llvm::StringMap<size_t> nodeToParitition;
```
artemrakhov:
```suggestion
  llvm::StringMap<size_t> nodeToPartition;
```

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,28 @@ struct HostConfig {
...
+  /// The name of the functiton to be partitioned.
+  std::string funcName;
+  /// The number of user defined partitions. The partition ids are [0,
+  /// numOfPartitions - 1];
+  size_t numOfPartitions;
+  /// The backend for each partition. backendNames.size() == numOfPartitions.
+  std::vector<std::string> backendNames;
+  /// The name for each partition. partitionNames.size() == numOfPartitions.
+  std::vector<std::string> partitionNames;
+  /// The mapping between nodes and Partition ids. Assume there are n nodes and
```
artemrakhov: Is this a map between node's name and unsigned partition number? It wasn't clear to me initially that name is used here. Also, need to clarify that Glow mangles names, meaning that Glow's node name is never the same as name from Caffe2 protobuf.

include/glow/Partitioner/Partitioner.h
```diff
@@ -295,17 +298,26 @@ class Partitioner {
...
+  /// found in Module.
+  /// The \p devices provides the cost model related to devices.
+  /// Saturating the host will be enabled if \p saturateHost is true.
+  /// \p optimized is false by default, which means the functions in this module
+  /// are not optimized. \p partitionConfig contains the user defined partition
+  /// info.
   Partitioner(Module *parent, const std::vector<DeviceInfo> &devices,
-              bool saturateHost = false, bool optimized = false);
+              bool saturateHost = false, bool optimized = false,
+              PartitionConfig partitionConfig = PartitionConfig());
```
artemrakhov: I have a question about API. Why do we have to change the constructor, this common API used by everyone? Why can't we leave it as is, but pass `PartitionConfig` into `PartitionFromConfig()` method? It is a bit obscure why we accept `PartitionConfig` here, store it, but delay using it until `PartitionFromConfig()` is called.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1235,6 +1250,11 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   // algorithm.
   F_ = selectRepFunc(module_, memSize_);
 
+  if (partitionConfig_.enabled()) {
```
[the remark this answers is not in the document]
beicy: will do it in front of selectRepFunc. However, we do need to run getBackendMap for later partition validation and logging.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1367,97 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+}
+
+llvm::Error Partitioner::PartitionFromConfig() {
+  Function *F = module_->getFunction(partitionConfig_.funcName);
+  RETURN_ERR_IF_NOT(F, "Can't find the function in current module.");
+
+  NodeToFunctionMap partitionMap;
+  std::vector<Function *> funcList;
+  std::unordered_set<size_t> unUsed;
+  std::vector<NodesSet> nodesSets;
```
artemrakhov:
```suggestion
  std::vector<NodesSet> nodesSets(partitionConfig_.numOfPartitions);
```
Then you won't need `nodesSets.push_back(NodesSet{});` later.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1367,97 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+      unMapped.push_back(&node);
+    } else {
+      size_t partitionID = iter->second;
+      partitionMap.add(&node, funcList[partitionID]);
+      unUsed.erase(partitionID);
+      nodesSets[partitionID].insert(&node);
+    }
+  }
+
+  // If there is un used partition and un-mapped nodes, map those nodes to the
```
artemrakhov:
```suggestion
  // If there is unused partition and unmapped nodes, map those nodes to the
```

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1367,97 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+  // Map the nodes the the partitions.
+  std::vector<Node *> unMapped;
+  for (auto &node : F->getNodes()) {
+    auto iter = partitionConfig_.nodeToParitition.find(node.getName());
+    if (iter == partitionConfig_.nodeToParitition.end()) {
+      // If a node in F is not in the node to partition mapping, put it into
+      // unMaped list.
+      unMapped.push_back(&node);
+    } else {
+      size_t partitionID = iter->second;
```
artemrakhov: Please check that partitionID is < partitionConfig_.numOfPartitions

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1367,97 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+      partitionMap.add(&node, funcList[partitionID]);
+      unUsed.erase(partitionID);
+      nodesSets[partitionID].insert(&node);
+    }
+  }
+
+  // If there is un used partition and un-mapped nodes, map those nodes to the
+  // un-used partition.
```
artemrakhov:
```suggestion
  // unused partition.
```

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1367,97 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+  if (unMapped.size()) {
+    assert(unUsed.size() && "Missing node to partition mapping.");
```
artemrakhov:
```suggestion
    assert(unUsed.size() == 1 && "There must be exactly 1 unused partition.");
```

include/glow/Partitioner/Partitioner.h
beicy: The reason is, we would like to leave the decision to Partitioner, and from all caller's the Partitioner interface is the same.  Later, we are going to refactor the Partitioner. After that, the Partitioner will be a facade that dispatches to the desired subclass/method.

lib/Partitioner/Partitioner.cpp
beicy: Done.

include/glow/Partitioner/Partitioner.h
artemrakhov: > from all caller's the Partitioner interface is the same

Why do we need this? This cannot be achieved. Imagine that one caller has user-defined partitions struct, the other caller doesn't. It means that caller #1 must pass this struct to partitioner, the second caller has nothing to pass. In caller Caller #1 calls `PartitionFromConfig()`, caller #2 calls something else. So interface is already different.

beicy: The interface for both caller 1 and 2 is Partition(). The partitionConfig is passed when the Partitioner is initialed. So in Partition(), it will check if "partitionConfig" is enabled, if so, PartitionFromConfig() will be called. Otherwise, just continue the auto-partition.

tests/unittests/PartitionerTest.cpp
```diff
@@ -886,3 +886,28 @@ TEST_F(PartitionerTest, memoryUsageValidation1) {
   auto err = myPartitioner.Partition(cctx);
   EXPECT_TRUE(errToBool(std::move(err)));
 }
+
+/// This one tests partition from a user-defined config.
+TEST_F(PartitionerTest, partitionFromConfig1) {
```
rdzhabarov: nit: partitionFromConfig

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,30 @@ struct HostConfig {
...
 };
 
+/// This is struct for user defined partition.
+struct PartitionConfig {
+  /// The name of the function to be partitioned.
+  std::string funcName;
+  /// The number of user defined partitions.
+  /// The partition ids are between 0 and numOfPartitions - 1, inclusive.
+  size_t numOfPartitions;
+  /// The backend for each partition. backendNames.size() == numOfPartitions.
```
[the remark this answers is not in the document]
beicy: This means backendNames.size() should be equal to numOfPartitions. Otherwise, there is an error.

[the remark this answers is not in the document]
rdzhabarov: Can't we just have backendNames alone and the size tells how many partitions.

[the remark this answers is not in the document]
beicy: It also works. The reason I added this separate field is: I think it is clear for people just use a number directly as a reference. The we compare the size of partition names and backends in case something is missing by accident.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1211,7 +1225,6 @@ llvm::Error Partitioner::QuantizationProfilingPartition(
   module_->eraseFunction(F_);
```
gcatron: Is it possible that the function can be renamed and will no longer match the name in partitionConfig?

include/glow/Runtime/RuntimeTypes.h
```diff
@@ -171,6 +171,30 @@ struct HostConfig {
...
+  std::vector<std::string> backendNames;
+  /// The name for each partition. partitionNames.size() == numOfPartitions.
+  std::vector<std::string> partitionNames;
+  /// The mapping between nodes' name to Partition ids. Assume there are n nodes
+  /// and m partitions. We have 2 types of valid mapping: 1. all nodes are
+  /// mapped to a partition. 2. For i-th (0 <= i < m) partition, the nodes
+  /// mapped to this partition id are not in this map, and the nodes mapped to
+  /// other partitions ids must be in this map. The node's name should be the
+  /// name in Glow function and may be different from the original name from
+  /// models. Since Glow will mangle names to make them unique.
```
artemrakhov: It's not only to make them unique. We also e.g. replace / with _. Maybe something else.
You may write: Since Glow will mangle names according to its own conventions.

include/glow/Partitioner/Partitioner.h
artemrakhov: You can see that it's already bad API. Let's say caller 1 has only `partitionConfig`. But now it has to pass `false, false` for some other 2 flags. Caller 1 has to figure out what is saturation, and what to pass for corresponding flag to disable.

It would be a good practice to make sure that incompatible flags are not passed together. E.g. saturation=false when user defined partition. But these checks can quickly explode if we have more flags/modes.

What if we later introduce another feature, third mode for partitioner? Say, PyTorch-defined partition? Then caller 3 will have to figure out what to pass for partitionConfig argument to disable user-defined partition feature.

This is why I think we should never accept a whole all-inclusive list of args in one function (in this case, partitioner constructor). I was going to propose create different method for user defined partition. But if you are going to refactor everything later into different classes, I'm also fine with that.

lib/Partitioner/Partitioner.cpp
```diff
@@ -1348,27 +1366,104 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
...
+  // If there is unused partition and unmapped nodes, map those nodes to the
+  // unused partition.
+  if (unMapped.size()) {
+    DCHECK(unused.size() == 1) << "There must be exactly 1 unused partition.";
+    auto partitionID = *(unused.begin());
+    for (auto &node : unMapped) {
+      partitionMap.add(node, funcList[partitionID]);
+      nodesSets[partitionID].insert(node);
+    }
+  }
```
artemrakhov: else {
what if there are unused partitions? Meaning that some partitions are empty. Is it valid?

lib/Partitioner/Partitioner.cpp
beicy: No. If partitionConfig is set, Partitioner will go to user-defined flow without considering other options like "-load-profiling".

lib/Partitioner/Partitioner.cpp
beicy: There can be at most 1 unused partition (there is the check). Otherwise, it is invalid. Since it means we have empty functions in the module and will cause problems later.

beicy:
@artemrakhov  Thanks a lot for the suggestion on the interface. Let's have more discussion about the interface offline.

lib/Partitioner/Partitioner.cpp
artemrakhov: > will cause problems later

It is better to nail these problems early. Here we still can report what exactly is wrong: no nodes were mapped to a partitions #i, #j, #k... Later it will be difficult to track down where the empty function is coming from. In fact, it may fail with obscure error, so first debugging step would be to figure out that the function is empty, and only after that find out why it's empty. I really suggest to report as many problems with the partition as we can find.

beicy: Yes, I agree. So far, if there is empty function, it will cause the assertion and stop. But you are right, the error message is not enough. Will improve the debugging ability. Thanks!

PR #3238, opened 2019-07-16
[New Operator][PyTorch] AdaptiveAvgPool

jackm321:
Summary:
Add the PyTorch AdaptiveAvgPool op and add loading from PyTorch [adaptiveavgpool2d](https://pytorch.org/docs/stable/nn.html#adaptiveavgpool2d) op.

Documentation:
doxygen

Test Plan:
Added unit tests
`python setup.py test`

### DIFF
--- a/lib/Exporter/ONNXModelWriter.cpp
+++ b/lib/Exporter/ONNXModelWriter.cpp
@@ -700,6 +700,18 @@ ONNXModelWriter::writeSparseToDenseMask(const SparseToDenseMaskNode *node,
   return writeAllWithNode(node->getName(), node, proto);
 }
 
+llvm::Error
+ONNXModelWriter::writeAdaptiveAvgPool(const AdaptiveAvgPoolNode *node,
+                                      GraphType &graph) {
+  auto *proto = graph.add_node();
+
+  const auto outShape = ShapeNHWC(node->getResult().dims());
+  std::vector<size_t> output_size{outShape.h, outShape.w};
+  addValueAttribute(proto, "output_size", llvm::makeArrayRef(output_size));
+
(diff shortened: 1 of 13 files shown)


lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -691,6 +691,18 @@ ONNXModelWriter::writeSparseToDenseMask(const SparseToDenseMaskNode *node,
   return writeAllWithNode(node->getName(), node, proto);
 }
 
+llvm::Error
+ONNXModelWriter::writeAdaptiveAvgPool(const AdaptiveAvgPoolNode *node,
```
yinghai: We still need this?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -445,6 +445,42 @@ void PyTorchModelLoader::loadMaxPool2d(const torch::jit::Node *ptNode) {
...
+
+  // Indexes of aten::adaptive_avg_pool2d inputs.
+  struct Inputs {
+    enum {
+      input = 0, // NCHW
+      output_size = 1,
+    };
+  };
+
+  // Glow expects inputs to be in NHWC but PyTorch keeps them in NCHW so we
```
yinghai: Glow works all in NHWC?

lib/Exporter/ONNXModelWriter.cpp
jackm321: Yeah actually this just landed yesterday #3167

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Yes

PR #3242, opened 2019-07-17
Fix bugs in JitVsGlow

zrphercule:
Two major problem now in jitVsGlow:
1. we should add no_grad flag, or else the generated graph may contain some differentiable ops which we dont accept.
2. in current jitVsGlow, if input "f" is already a ScriptModule, then everything about it will be cached in its first run, and it will never fuse any ops in its second run since it will only run the cached result.


zrphercule:
@jackm321 Could you please take a look? Since we are not using it in trainning, I guess we should add this?

jackm321:
Nice!
1. makes sense
2. can you explain this more, not sure I understand. Is this if we make a script module then pass it to jitVsGlow it will be a problem or if you call jitVsGlow twice on the same `f`, or something else?

zrphercule:
> Nice!
> 
> 1. makes sense
> 2. can you explain this more, not sure I understand. Is this if we make a script module then pass it to jitVsGlow it will be a problem or if you call jitVsGlow twice on the same `f`, or something else?

For example you have a script module call M = jit.load("a.pt"), and you want to test the lowering using the original script. What you will observe is, in the second run "glow_trace = torch.jit.trace(M, inputs)" will do nothing, not even fuse a single op. The reason is (M, inputs) is cached in the first run with no_fuse flag, and if M and inputs remains unchanged, this cache will always work and never fuse the nodes.
Therefore what I have done here is made a new API for this script. In the future when we do unit test with M, a script module or module loaded from .pt files, we should make a copy (or load twice as @wanchaol suggested) of M: M_copied, and call this function jitVsGlow_(M, M_copied, inputs) to make sure the first run didnt cache the second run.

Our current unit test can pass is because we only use function modules, and a function module will not be cached.

yinghai:
> can you explain this more, not sure I understand. Is this if we make a script module then pass it to jitVsGlow it will be a problem or if you call jitVsGlow twice on the same f, or something else?

https://github.com/pytorch/pytorch/blob/f81395b3e3d25368239e36541d820e51f5e7a3c2/torch/csrc/jit/graph_executor.cpp#L512-L521

PR #3244, opened 2019-07-17
Add bias for Habana Gemm node

bertmaher:
Summary:
Habana is allowing bias for gemm. This commit will fix resnet results
with asymmetric quantization

Differential Revision: D16348533


lib/Backends/Habana/Habana.cpp
```diff
@@ -1092,6 +1092,7 @@ HabanaBackend::compile(Function *F, const BackendOptions &opts) const {
         std::vector<synTensor> inputs;
         inputs.push_back(tensors[MI->getLHS()].get());
         inputs.push_back(tensors[MI->getRHS()].get());
+        inputs.push_back(nullptr);
```
jfix71: Can you add `/* bias */`? It's not really intuitive to think of a MatMul with a Bias.

rdzhabarov:
The failed test in coverage - PartitionerTest.Basic1 (I restarted it).
That should be fixed by https://github.com/pytorch/glow/pull/3241

PR #3258, opened 2019-07-18
QuantizationTest intermittent segfault fix.

gcatron:
Summary:
Moved collect constants in OCL backend to before creation of the compiled function. This fixes an intermittant segfault in quantizationTest.

Documentation:

Test Plan: ninja check, notice quantizationTest now passes reliably


jackm321:
What's the issue this fixes?

SplitInfinity:
Can you fix the PR title?

gcatron:
> What's the issue this fixes?

Quantization test was intermittently failing in the OpenCL end2endInt8 test. This fixes that bug, On a wider not this would fix any case for OpenCL tests that create a function in a module different from the member module of the OCL ExecutionEngine.

gcatron:
As follow on, I created #3259 to investigate a better way to handle moving the RuntimeBundle so this type of bug doesn't show up later.

Issue #3239, opened 2019-07-17
Port ExecutionEngine to use runtime


gcatron:
Port the ExecutionEngine to the new runtime interface through HostManager and update all Tests that use the EE.

The plan of action:
Port all tests and uses of EE and BackendTestUtils to EE2 BackendUtils2. Once all files have been migrated remove original ExecutionEngine and BackendTestUtils and rename ExecutionEngine2 to ExecutionEngine and BackendTestUtils2 to BackendTestUtils.

For several tests the port will be a no-op but moving it will make it easier to ensure nothing is missed.

PR #3260, opened 2019-07-19
Created transition BackendTestUtils2

gcatron:
Summary:
Created transition BackendTestUtils2 which is a copy of BackendTestUtils that has been ported to EE2. This is an intermediate step so tests can be ported one at a time. Once all users of BackendTestUtils have been ported BackendTestUtils2 will replace BackendTestUtils. Also ported HyphenTest which uses the new BackendUtils2.

Documentation: NA

Work on #3239 

Test Plan: ninja check, verify HyphenTest passes.

### DIFF
--- a/tests/unittests/BackendTestUtils2.cpp
+++ b/tests/unittests/BackendTestUtils2.cpp
@@ -1,0 +1,1129 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


jfix71:
@gcatron Ok I have a (hopefully not too annoying) request 😄. In theory this PR shouldn't be so hard to review because there probably aren't so many differences between BackendTestUtils2 and BackendTestUtils. But in practice because there's a single commit here with the new BackendTestUtils2 files and so it's hard to tell what's changed vs the original versions.

So, instead of directly copying + adding the new BackendTestUtils2 files, can you do something like:
```
git mv BackendTestUtils.cpp BackendTestUtils2.cpp
git commit BackendTestUtils.cpp -m "Just moved old version to new"
cp BackendTestUtils2.cpp BackendTestUtils.cpp
git add BackendTestUtils.cpp
git commit BackendTestUtils.cpp -m "Bring back the old version with the same name"
// Make changes to BackendTestUtils2.cpp
git commit BackendTestUtils2.cpp -m "New changes."
```
Then on GH it'd be possible to see the actual meaningful changes to `BackendTestUtils2.cpp` in the "New changes" commit.

I tried the above out myself on `EE->EE3` and it's pretty easy and works great. See [this branch](https://github.com/pytorch/glow/compare/master...jfix71:TEST) and [this commit](https://github.com/pytorch/glow/commit/99317cf8cfb89efabab20e8baf847e30633dc5b7). And it would make reviewing this PR and similar ones in the future much easier/more effective.

gcatron:
Closing this PR, 3263 was created for the same changes but a series of stacked commits to make review easier.

PR #3263, opened 2019-07-19
Created transition BackendTestUtils2

gcatron:
Summary:
Created transition BackendTestUtils2 which is a copy of BackendTestUtils that has been ported to EE2. This is an intermediate step so tests can be ported one at a time. Once all users of BackendTestUtils have been ported BackendTestUtils2 will replace BackendTestUtils. Also ported HyphenTest which uses the new BackendUtils2.
For ease of reviewing look at the last diff in the stack, it will show what actually changed in BackendTestUtils.

Documentation: NA

Work on #3239

Test Plan: ninja check, verify HyphenTest passes.


tests/unittests/BackendTestUtils2.cpp
```diff
@@ -340,47 +342,60 @@ void trainConvNet(Tensor *inputs, Tensor *kernel1, Tensor *bias1,
...
+    auto *conv2 = F->createConv(bindings, "conv2", reshape1, 2, 2, 2, 0, 1);
+    bindings.get(cast<Placeholder>(conv2->getFilter()))->assign(kernel2);
+    bindings.get(cast<Placeholder>(conv2->getBias()))->assign(bias2);
+    auto *reshape2 = F->createReshape("reshape2", conv2, shape2);
+    auto *softmax = F->createSoftMax("softmax", reshape2, var2);
+    F->createSave("ret", softmax);
+  }
+  trainingBindings.allocate(EET.getModule().getPlaceholders());
+  inferBindings.allocate(EEI.getModule().getPlaceholders());
+  bindings.copyTrainedWeightsTo(trainingBindings);
```
jfix71: Hm, so this is necessary because we need to copy in e.g. the trainable PHs' backing tensors created by `createConv()` etc., since they've been initialized in `bindings`?

I guess the name `copyTrainedWeightsTo()` feels misleading/wrong now, because the PHs aren't actually trained yet. Perhaps it would be better to rename it to `copyTrainableWeightsTo()` since that's what's actually happening anyway.

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -401,38 +418,52 @@ void trainLocalResponseNormalizationNet(Tensor *inputs, Tensor *weights,
...
+    auto *result = F->createSave("ret", softmax);
+    bindings.allocate(result->getPlaceholder());
+  }
+  trainingBindings.allocate(EET.getModule().getPlaceholders());
+  bindings.copyTrainedWeightsTo(trainingBindings);
+  bindings.clear();
+  bindings.allocate(EEI.getModule().getPlaceholders());
+  auto *TF = glow::differentiate(EET.getModule().getFunction(fName), TC);
+  auto tfName = TF->getName();
+  EET.compile(CompilationMode::Train);
```
jfix71: tiny nit: new line after here to try to separate the different phases. Also consistent with other places.

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -978,22 +1054,22 @@ void inferExtract3D(Tensor *input, Tensor *out, llvm::StringRef kind) {
...
+  ExecutionEngine2 EE(kind);
   auto &mod = EE.getModule();
   Function *F = mod.createFunction("main");
 
-  auto T = mod.uniqueType(ElemKind::Int8QTy, input->getType().dims(),
-                          2 * input->getType().getScale(),
-                          -input->getType().getOffset());
+  auto T =
+      mod.uniqueType(ElemKind::Int8QTy, input->getType().dims(),
+                     input->getType().getScale(), input->getType().getOffset());
```
jfix71: Why did you change these?

tests/unittests/BackendTestUtils2.h
```diff
@@ -298,6 +296,12 @@ void inferExtract3D(Tensor *input, Tensor *out, llvm::StringRef kind);
 
 void inferMaxSplat(Tensor *input, Tensor *out, llvm::StringRef kind);
 
+void insertCompiledFunction(llvm::StringRef name, CompiledFunction *func,
```
jfix71: Looks like this is unused. Is it going to be used by a future PR, if so maybe add it there?

tests/unittests/BackendTestUtils2.h
```diff
@@ -298,6 +296,12 @@ void inferExtract3D(Tensor *input, Tensor *out, llvm::StringRef kind);
...
+                            runtime::DeviceManager *device, Module *mod);
+
+void runOnDevice(ExecutionContext &context, llvm::StringRef name,
```
jfix71: Looks like this is unused. Is it going to be used by a future PR, if so maybe add it there?

tests/unittests/HyphenTest.cpp
```diff
@@ -323,8 +319,18 @@ struct HyphenNetwork {
 };
 } // namespace
 
+// Copy values from one PlaceholderBindings to another by name. This allows for
+// training a network on one EE then using the trained weights on another.
+void copyBetweenBindings(llvm::StringRef name, PlaceholderBindings &src,
```
jfix71: Isn't this exactly the same as `PlaceholderBindings::copyToTarget()`? Why is it necessary?

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -264,27 +258,29 @@ void inferIntLookupTableNet(Tensor *input, Tensor *out,
...
   PlaceholderBindings bindings;
-  ExecutionEngine EE(kind);
+  ExecutionEngine2 EE(kind);
   auto &mod = EE.getModule();
   Function *F = mod.createFunction("main");
-  auto outTy = mod.uniqueType(ElemKind::Int8QTy, {input->size()}, 3, 3);
-  auto var = createQuantizedPlaceholder(mod, bindings, input, outTy->getScale(),
-                                        outTy->getOffset(), "var");
-  auto *lookupTable = F->createIntLookupTable("lookuptable", var, table, outTy);
+  auto inTy =
```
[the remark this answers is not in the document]
gcatron: This is because I was running into issues with BackendCorrectness Test. I'll back out the type change for now.

tests/unittests/BackendTestUtils2.cpp
gcatron: It was related to BackendCorrectness, I'll back it out for this PR and revisit in the BackendCorrectnessTest PR.

tests/unittests/BackendTestUtils2.h
gcatron: These are both used in other PRs, I was porting BackendTestUtils as a whole, I can back these out until a later PR.

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -1007,11 +1083,47 @@ void inferMaxSplat(Tensor *input, Tensor *out, llvm::StringRef kind) {
...
-  EE.compile(CompilationMode::Infer, F);
+  EE.compile(CompilationMode::Infer);
 
-  updateInputPlaceholders(bindings, {var}, {input});
+  updateInputPlaceholders2(bindings, {var}, {input});
   EE.run(bindings);
   out->assign(resultTensor);
 }
 
+void insertCompiledFunction(llvm::StringRef name, CompiledFunction *func,
```
jfix71: Left behind the impl too?

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -1007,11 +1083,47 @@ void inferMaxSplat(Tensor *input, Tensor *out, llvm::StringRef kind) {
...
+  device->addNetwork(mod, std::move(functionMap),
+                     [&addPromise, &addErr](const Module *, llvm::Error err) {
+                       addErr = std::move(err);
+                       addPromise.set_value();
+                     });
+  fut.wait();
+  EXIT_ON_ERR(std::move(addErr));
+}
+
+void runOnDevice(ExecutionContext &context, llvm::StringRef name,
```
jfix71: Left behind the impl too?

tests/unittests/BackendTestUtils2.cpp
```diff
@@ -141,51 +143,44 @@ void compareAgainstInterpreter(llvm::StringRef backendName,
...
-  const bool alreadyProfiled = isQuantizedElemKind(interpElemKind) ||
-                               isQuantizedElemKind(backendElemKind);
-  if (!alreadyProfiled || interpElemKind != ElemKind::FloatTy) {
-    IEE.compile(IF, cctxI);
-    IEE.run(Ibindings);
-  }
+  BEE.compile(cctxB);
+  BEE.run(bBindings);
+
+  // Run the inference
```
jfix71: nit: Probably just remove this. If not add period.

tests/unittests/BackendTestUtils2.cpp
gcatron: Derp! removed now.

tests/unittests/BackendTestUtils2.cpp
gcatron: Also removed.

PR #3264, opened 2019-07-19
Extending ONNXModelLoader to support load of the specific Glow ops

putivsky:
Summary:
Glow uses ONNX protobufs for Graph serialization (writer) and deserialization (loader).
Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/lib/Exporter/ONNXModelWriter.cpp
+++ b/lib/Exporter/ONNXModelWriter.cpp
@@ -35,6 +35,15 @@ template <typename T> struct AttributeAssigner<false, llvm::ArrayRef<T>> {
   }
 };
 
+// Specialization for 1D, 1 element Tensor container types
+template <> struct AttributeAssigner<false, Tensor> {
+  static void assign(ONNX_NAMESPACE::AttributeProto *attr, const Tensor &T) {
+    auto *proto = attr->mutable_t();
+    ONNXModelWriter::writeTensor(T, proto);
+    attr->set_type(ONNX_NAMESPACE::AttributeProto::TENSOR);
+  }
+};
+
 // Specialization for string type
 template <> struct AttributeAssigner<false, std::string> {
   static void assign(ONNX_NAMESPACE::AttributeProto *attr,
@@ -70,23 +79,62 @@ void addValueAttribute(ONNX_NAMESPACE::NodeProto *proto,
                                                                    container);
 }
 
-/// Helper function to recursively get inputs for the broadcast node.
-/// Broadcast node get constructed as a chain of Reshape->Tile->...->Tile.
-const Node *unwindBroadcastInput(const TileNode *tile,
-                                 ReportedNodes &reporter) {
+/// Helper function to recursively rewind Tile \p node.
+/// Optionally, if provided fills out the repeats \p repeats.
+/// Returns the first Tile in a chain of Tiles.
+const TileNode *unwindTile(const TileNode *node, std::vector<size_t> *repeats,
+                           ReportedNodes &reporter) {
+  // unwind Tile
+  // Keep track of detected <axis, count> pairs.
+  std::vector<std::pair<unsigned_t, unsigned_t>> info;
+  const TileNode *tile = node;
   while (tile) {
-    reporter.insert(tile);
-    if (const ReshapeNode *RN =
-            llvm::dyn_cast<ReshapeNode>(tile->getInput().getNode())) {
-      return RN;
-    } else if (const TileNode *TN =
-                   llvm::dyn_cast<TileNode>(tile->getInput().getNode())) {
+    // Insert counts and axises in reverse order,
+    // cause rewind algorithm navigates from the bottom to the top.
+    info.insert(info.begin(), {tile->getAxis(), tile->getCount()});
+
+    if (const auto *TN = llvm::dyn_cast<TileNode>(tile->getInput().getNode())) {
+      reporter.insert(TN);
       tile = TN;
     } else {
-      return nullptr;
+      break;
+    }
+  }
+
+  if (repeats) {
+    unsigned_t numDims = tile->getInput().dims().size();
+    // axis is in a normal case will have [0, 1, ..., numDims - 1] values,
+    // in extreme case []. Find missing indices and insert count 1.
+    auto aB = info.begin();
+
+    for (unsigned_t i = 0; i < numDims; ++i, ++aB) {
+      if (aB == info.end() || aB->first != i) {
+        aB = info.insert(aB, {i, 1});
+      }
+    }
+
+    for (size_t b = 0, e = info.size(); b < e; ++b) {
+      repeats->push_back(info[b].second);
     }
   }
   return tile;
+}
+
+/// Helper function to recursively get inputs for the broadcast node.
+/// Broadcast node get constructed as a chain of Reshape->Tile->...->Tile.
+const Node *unwindBroadcastInput(const TileNode *tile,
+                                 std::vector<size_t> *repeats,
+                                 ReportedNodes &reporter) {
+  tile = unwindTile(tile, repeats, reporter);
+  DCHECK(tile);
+
+  reporter.insert(tile);
+  if (const ReshapeNode *RN =
+          llvm::dyn_cast<ReshapeNode>(tile->getInput().getNode())) {
+    return RN;
+  } else {
+    return nullptr;
+  }
 }
 
 /// Writes all outputs from Node \p node to protobuf \p proto.
@@ -106,6 +154,22 @@ void outputsToProto(const Node *node, ONNX_NAMESPACE::NodeProto *proto) {
       }
     }
   }
+}
+
+/// Write the output of the provided type only of node outputs.
+bool outputKindToProto(Kinded::Kind kind, const Node *node,
+                       ONNX_NAMESPACE::NodeProto *proto) {
+  bool found = false;
+  for (const auto &use : node->getUsers()) {
+    const auto *user = use.getUser();
+    if (user->getKind() == Kinded::Kind::SaveNodeKind) {
+      proto->add_output(user->getName());
+    } else if (user->getKind() == kind) {
+      found = true;
+      outputsToProto(user, proto);
+    }
+  }
+  return found;
 }
 
 /// Writes all inputs from Node \p node to protobuf \p proto.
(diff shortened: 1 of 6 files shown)


lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -166,28 +238,117 @@ llvm::Error writeArithmetic(const std::string &opName, const T *node,
...
+  auto *tensorShape = tensorType->mutable_shape();
+  const auto &dims = ty->dims();
+  for (unsigned b = 0, e = dims.size(); b < e; ++b) {
+    auto *tensorDims = tensorShape->add_dim();
+    tensorDims->set_dim_value(dims[b]);
+  }
+}
+
+/// A helper class for ordering Graph nodes in a truly post-order order.
+class GraphBackwardsPostOrderVisitor : public PostOrderVisitor {
```
jfix71: I'm confused by this -- how is this different than `GraphPreOrderVisitor`? Why is it necessary?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -211,51 +372,55 @@ ONNXModelWriter::ONNXModelWriter(const std::string &modelFilename, Function &F,
...
     graphProto->set_name("glow");
-
     // Use post order graph traversal.
     // If node is constant with uses, turned into "input" and create a tensor
     // If node is placeholder with uses, turned into "input" with tensor shape,
     // except the case when placeholder has use as SaveNode.
     // Otherwise call common operators method or special operators and write
     // protobuf inputs from node inputs and protobuf outputs from uses.
-    GraphPostOrderVisitor visitor(G_);
+    GraphBackwardsPostOrderVisitor visitor(G_);
```
jfix71: Why do we want pre order instead of post order?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -883,34 +1111,176 @@ DEF_ALL_WRITER_NODE(Log)
...
+  Tensor oneDimTensor(ElemKind::Int64ITy, {numDims});
+  auto handle = oneDimTensor.getHandle<int64_t>();
+
+  for (size_t b = 0, e = shape.size(); b < e; ++b) {
+    handle.raw(b) = shape[b];
+  }
+
+  // Add Tensor type attribute.
+  addValueAttribute(indices, "value", oneDimTensor);
+  // Add indices as input to the Title node
```
jfix71: typo: Tile

putivsky:
@jfix71, Broadcast Node is a good idea, but I would prefer to implement it later, cause many other files should be changed.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -70,25 +79,66 @@ void addValueAttribute(ONNX_NAMESPACE::NodeProto *proto,
...
+    if (const auto *TN = llvm::dyn_cast<TileNode>(tile->getInput().getNode())) {
+      reporter.insert(TN);
       tile = TN;
     } else {
-      return nullptr;
+      break;
+    }
+  }
+
+  if (shape) {
```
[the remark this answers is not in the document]
putivsky: It's more complicated than that. Not all Tile nodes got created if dim has size 1, however Tile has Axis information and it indicates if some Tile nodes got missed and helps to restore the correct dimensions.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -434,10 +577,13 @@ llvm::Error ONNXModelWriter::writeConvolution(const ConvolutionNode *node,
...
   }
 
+  proto->add_input(node->getBias().getNode()->getName());
+
   proto->set_name(node->getName());
   proto->set_op_type("Conv");
 
-  outputsToProto(node, proto);
+  // Use the output of transpose node.
+  outputKindToProto(Kinded::Kind::TransposeNodeKind, node, proto);
```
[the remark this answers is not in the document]
putivsky: Convolution has always a one following TransposeNode. We write the output of TransposeNode as proto output.

lib/Exporter/ONNXModelWriter.cpp
putivsky: There are GraphVisitor classes GraphPreOrderVisitor and GraphPortOrderVisitor, but both first visiting inputs and outputs and then only current node. In my case I need the current node got visited first and then inputs (at this time outputs already will be visited in most cases).

lib/Exporter/ONNXModelWriter.cpp
putivsky: We want post order but with a modified visiting order between current node and its inputs.

tests/unittests/OnnxExporterTest.cpp
```diff
@@ -38,17 +38,32 @@ void testLoadAndSaveONNXModel(const std::string &name) {
...
-  ONNXModelLoader onnxLD(name, {}, {}, *F, &err);
+  {
+    ONNXModelLoader onnxLD(name, {}, {}, *F, &err);
+    irVer = onnxLD.getIrVersion();
+    opsetVer = onnxLD.getOpSetVersion();
+  }
 
   if (err) {
     llvm::errs() << "ONNXModelLoader failed to load model: " << name << ".\n";
     return;
```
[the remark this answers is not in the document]
putivsky: No, there are a few files that ONNXModelLoader can't load (probably should be fixed in a first place), but I want test to continue, this is ONNXModelWrite test.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -108,6 +158,28 @@ void outputsToProto(const Node *node, ONNX_NAMESPACE::NodeProto *proto) {
...
+  bool found = false;
+  for (const auto &use : node->getUsers()) {
+    const auto *user = use.getUser();
+    if (user->getKind() == Kinded::Kind::SaveNodeKind) {
+      proto->add_output(user->getName());
+    } else if (user->getKind() == kind) {
+      // Find the user input that matches input node parameter.
+      for (unsigned b = 0, e = user->getNumInputs(); b < e; ++b) {
+        auto *UIN = user->getNthInput(b).getNode();
+        if (UIN == node) {
```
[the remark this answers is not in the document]
putivsky: I need the name of user's particular input that connects to the UIN. User can have more than one input.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1062,18 +1062,22 @@ ONNXModelLoader::loadSpaceToDepth(const ONNX_NAMESPACE::NodeProto &op,
...
     RETURN_IF_ERR(loadTensor(dict.at("value")->t(), &T));
-    RETURN_ERR_IF_NOT(T.dims().size() == 1, "Value must be a 1D vector.");
-    RETURN_ERR_IF_NOT(T.getType().getElementType() == ElemKind::FloatTy ||
-                          T.getType().getElementType() == ElemKind::Int64ITy ||
-                          T.getType().getElementType() == ElemKind::Int32ITy,
-                      T.getType().getElementName().str() +
-                          " type Value is not supported.");
+    if (!isSplat) {
+      // Validate tensor only for ConstantOfShape operator
+      RETURN_ERR_IF_NOT(T.dims().size() == 1, "Value must be a 1D vector.");
```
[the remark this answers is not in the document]
putivsky: If it's not Splat Node then Tensor can have more than 1 dimension.

putivsky:
@jfix71, thanks a lot for the review, suggestions and feedback.

lib/Exporter/ONNXModelWriter.cpp
jfix71: Hmm, I guess I don't see how it's different than `GraphPreOrderVisitor`. You're overriding pre and post to do the same thing as `GraphPreOrderVisitor`. And visit is the same thing, visiting all Constants and then all root nodes. Can you point out where they're different?

tests/unittests/OnnxExporterTest.cpp
```diff
@@ -60,9 +76,36 @@ TEST(exporter, onnxModels) {
...
+        name.find("/sparseLengthsSum.onnxtxt") != std::string::npos ||
+        name.find("/constantOfShapeInt32Fail.onnxtxt") != std::string::npos ||
+        name.find("/padEdge.onnxtxt") != std::string::npos ||
+        name.find("/castToFloat.onnxtxt") != std::string::npos ||
+        name.find("/castToFloat16.onnxtxt") != std::string::npos ||
+        name.find("/castToInt64.onnxtxt") != std::string::npos ||
+        name.find("/castToInt32.onnxtxt") != std::string::npos ||
+        name.find("/Where.onnxtxt") != std::string::npos ||
+        name.find("/constantOfShapeInt64Fail.onnxtxt") != std::string::npos) {
+      // Ignore invalid ONNX files and graphs without nodes.
```
jfix71: Curious, what makes these invalid?

lib/Exporter/ONNXModelWriter.cpp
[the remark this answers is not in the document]
putivsky: Detailed logic:
1. Loop through all users of the Parent Node.
  `for (const auto &use : node->getUsers())`
2. For each user, loop through its inputs.
  `for (unsigned b = 0, e = user->getNumInputs(); b < e; ++b)`
3. Check if the current input node is identical to the Parent Node. It can be input from other nodes.
  `if (UIN == node)`
4. Now once we found the "correct" input, get its name.
  `UIN->getName()`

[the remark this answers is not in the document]
putivsky: Ah, I have seen the second function, fixing it.

lib/Exporter/ONNXModelWriter.cpp
putivsky: Yes, you are right, I have started with PostOrder having a wrong perception that PreOrder will start with root.

tests/unittests/OnnxExporterTest.cpp
putivsky: Don't know, probably some changes over time.

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -211,51 +336,55 @@ ONNXModelWriter::ONNXModelWriter(const std::string &modelFilename, Function &F,
     opsetImportProto->set_version(opsetVersion);
     auto *graphProto = modelProto.mutable_graph();
     graphProto->set_name("glow");
-
     // Use post order graph traversal.
```
jfix71: nit: Please update this comment.

lib/Exporter/ONNXModelWriter.cpp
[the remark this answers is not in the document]
jfix71: Thanks for explaining this @putivsky. So my confusion is inside `outputKindToProto()`. It's guaranteed that `if (UIN == node)` will be true least once, because we already know `user` uses `node`, correct? And it seems like we don't care if we've found the "correct" input. So what is wrong with doing:

```cpp
  for (const auto &use : node->getUsers()) {
    const auto *user = use.getUser();
    if (user->getKind() == Kinded::Kind::SaveNodeKind) {
      proto->add_output(user->getName());
    } else if (user->getKind() == kind) {
      found = true;
      outputsToProto(user, proto);
    }
  }
```

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -502,12 +634,18 @@ ONNXModelWriter::writeMeanVarNormalization(const MeanVarNormalizationNode *node,
...
                                         GraphType &graph) {
   auto *proto = graph.add_node();
-  // Find dictionary entries.
+  // Add dictionary entries.
   auto starts = node->getStart();
   auto outs = node->getInput().dims();
   RETURN_ERR_IF_NOT(starts.size() == outs.size(),
                     "Mismatch starts and result dimensions.");
 
+  llvm::Error err = writeAllWithNode("Slice", node, proto);
```
jfix71: nit: Can we just do `RETURN_IF_ERR(writeAllWithNode("Slice", node, proto));`?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -573,10 +714,22 @@ llvm::Error ONNXModelWriter::writePow(const PowNode *node, GraphType &graph) {
...
-  return writeAllWithNode("topK", node, proto);
+  Tensor scalar(ElemKind::Int64ITy, {1});
+  auto handle = scalar.getHandle<int64_t>();
+  handle.raw(0) = node->getK();
+
+  auto *tensorProto = graph.add_initializer();
+  tensorProto->set_name("k");
+  writeTensor(scalar, tensorProto);
+
+  llvm::Error err = writeAllWithNode("TopK", node, proto);
```
jfix71: nit: Can we just do `RETURN_IF_ERR(writeAllWithNode("Slice", node, proto));`?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -883,34 +1087,176 @@ DEF_ALL_WRITER_NODE(Log)
...
+  proto->add_input(node->getBias().getNode()->getName());
+  outputsToProto(node, proto);
+  return llvm::Error::success();
+}
+
+llvm::Error ONNXModelWriter::writeSparseToDense(const SparseToDenseNode *node,
+                                                GraphType &graph) {
+  auto *proto = graph.add_node();
+
+  llvm::Error err = writeAllWithNode("SparseToDense", node, proto);
```
jfix71: nit: Can we just do `RETURN_IF_ERR(writeAllWithNode("Slice", node, proto));`?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -748,109 +904,157 @@ llvm::Error ONNXModelWriter::writeBatchBoxCox(const BatchBoxCoxNode *node,
...
-  return writeAllWithNode(node->getName(), node, proto);
+  return writeAllWithNode("ChannelwiseQuantizedConvolution", node, proto);
 }
 
 llvm::Error ONNXModelWriter::writeSplat(const SplatNode *node,
                                         GraphType &graph) {
   auto *proto = graph.add_node();
-  // Find dictionary entries.
-  addValueAttribute(proto, "value", node->getValue());
+  // Convert value to tensor with result shape
```
jfix71: So for Splat the proto will include `value` repeated `tensor.size()` times? Why can't we just use `ConstantOfShape` here? Or at least follow the same design as `ConstantOfShape`, where we provide a single `value` attribute and 1D `input` tensor of the shape? Just seems like it's an unnecessary waste of memory.

putivsky: The problem is Max operator, it needs not only the value but the shape as well.

PR #3268, opened 2019-07-19
Fix a bug: quantization profiling flow with partitioning can only be enabled when there is no previous registered network in HM.

beicy:
Summary:
When we enabled quantization profiling flow in Partitioner, we forced Provisioner to be reset. If there are registered networks in HM already, the status will be lost. Therefore, we restrict the quantization profiling flow to only 1 network. 

Documentation:

[Optional Fixes #issue]

Test Plan:
ninja test, ./tests/images/run.sh

### DIFF
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -113,6 +113,11 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
   auto nodeList = std::move(partitioner.getPartitionResult());
 
   if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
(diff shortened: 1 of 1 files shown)


lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -113,6 +113,9 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
   auto nodeList = std::move(partitioner.getPartitionResult());
 
   if (cctx.precisionConfig.quantMode == QuantizationMode::Profile) {
+    RETURN_ERR_IF_NOT(networks_.size() < 1,
```
gcatron: nit: I think RETURN_ERR_IF(networks_.size() != 0) might be a little easier to read.

PR #3269, opened 2019-07-19
[PyTorch] Add support for broadcasting arithmetic ops

jackm321:
Summary:
PyTorch arithmetic ops support multidirectional [broadcasting](https://github.com/onnx/onnx/blob/master/docs/Broadcasting.md) so add this to the loader

Documentation:
doxygen

Test Plan:
added tests


zrphercule:
One question: can I always use createNodeWithBroadcast, no matter what node it is?

jackm321:
@zrphercule no looks like it's just mul, div, add, and sub https://github.com/pytorch/glow/blob/f453b647fa8b8e60bec76af6202b3bf49a036e18/include/glow/Graph/Graph.h#L709-L712

PR #3271, opened 2019-07-20
Implement torch.min (two input elementwise version)

bertmaher:
Summary: As title

Test Plan: New test, `test_elementwise_min`.  Ran `python setup.py test`

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -497,7 +497,19 @@ void PyTorchModelLoader::loadTranspose(const torch::jit::Node *ptNode) {
   } else {
     assert(false && "Transpose requires input to have rank <= 2");
   }
+  addGlowNodeValue(outputs[0], output);
+}
 
+void PyTorchModelLoader::loadMin(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  auto lhs = getGlowNodeValue(inputs[0]);
+  auto rhs = getGlowNodeValue(inputs[1]);
+
+  auto output = f_->createMin("min", lhs, rhs);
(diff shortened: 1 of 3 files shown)


torch_glow/tests/nodes/min_test.py
```diff
@@ -0,0 +1,10 @@
+import torch
+
+from tests.utils import jitVsGlow
+
+
+def test_elementwise_min():
+    def test_f(a, b):
+        return torch.min(a + a, b + b)
```
bertmaher: @jackm321 I had to do these random pointless additions to get this function glow-ified... do you know why that would be the case?  I basically copied the idea from your transpose diff :)

jackm321: @bertmaher yeah so it's basically because we are making use of PTs graph fusion stuff to do the lowering from subgraphs of PT nodes to Glow and that thing needs at least two PT nodes to fuse together to create a `glow::CompilationGroup` node.
I was thinking it might be good to document this somewhere, maybe I'll throw it in pytorch.md

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -497,7 +497,19 @@ void PyTorchModelLoader::loadTranspose(const torch::jit::Node *ptNode) {
...
+  addGlowNodeValue(outputs[0], output);
+}
 
+void PyTorchModelLoader::loadMin(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  auto lhs = getGlowNodeValue(inputs[0]);
```
jackm321: Nice job not copying my backwards `lhs`/`rhs`s 😛

torch_glow/tests/nodes/min_test.py
bertmaher: Ahh, that makes sense, I half-wondered if that's what was going on.  Thanks!

PR #3276, opened 2019-07-22
Ported RecommendationSystemTest to EE2.

gcatron:
Summary: This PR ports the RecommendationSystemTest to the new EE2. Additionally this adds a new feature to EE2, the ability to set the memory of the device. This allows programmatically setting the EE device memory to support graphs larger than the default.

Documentation: NA

Work on #3239 

Test Plan: verify everything builds, and RecomendationSystemTest runs and passes.

### DIFF
--- a/lib/ExecutionEngine/ExecutionEngine2.cpp
+++ b/lib/ExecutionEngine/ExecutionEngine2.cpp
@@ -45,6 +45,9 @@ void ExecutionEngine2::setBackendName(llvm::StringRef backend) {
   if (backend != "") {
     std::vector<std::unique_ptr<runtime::DeviceConfig>> configs;
     auto config = llvm::make_unique<runtime::DeviceConfig>(backend);
+    if (deviceMemory_) {
+      config->setDeviceMemory(deviceMemory_);
+    }
     configs.push_back(std::move(config));
     hostManager_ = llvm::make_unique<runtime::HostManager>(std::move(configs));
   }
--- a/tests/unittests/RecommendationSystemTest.cpp
+++ b/tests/unittests/RecommendationSystemTest.cpp
@@ -13,10 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "BackendTestUtils.h"
+#include "BackendTestUtils2.h"
 
 #include "glow/Converter/TypeAToTypeBFunctionConverter.h"
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Partitioner/Partitioner.h"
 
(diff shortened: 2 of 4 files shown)


tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -567,94 +568,109 @@ class RecommendationSystemTest : public BackendTest {
...
+    // Create the tables.
+    std::vector<Placeholder *> lengths(tableSizes.size());
+    for (unsigned int i = 0; i < lengths.size(); i++) {
+      lengths[i] = mod.createPlaceholder(ElemKind::Int32ITy, {miniBatch},
+                                         "SL" + std::to_string(i), false);
+    }
+
+    auto *denseData = mod.createPlaceholder(ElemKind::FloatTy,
+                                            {miniBatch, denseDim}, "denseData",
+                                            false); // Dense dim can be anything
```
beicy: "Dense dim" means the param denseDim ? If so, may be just put /* ..*/ in front of "denseDim"

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -909,7 +934,8 @@ TEST_P(RecommendationSystemTest, RecSys_Partitioned_RWQuantized_SLWS_FP16) {
   testRecSys(/* quantizeSLWS */ true,
              /* quantizeFC */ false,
              /* convertToFP16 */ true,
-             /* gatherWeights */ false);
+             /* gatherWeights */ false,
+             /*setupPartitionTest */ true);
```
beicy: nit: a space before "setupPartitionTest"

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -80,7 +80,7 @@ llvm::cl::list<unsigned> tableCountsOpt(
 llvm::cl::opt<unsigned> deviceMemCapacityOpt(
     "device-mem-capacity",
     llvm::cl::desc("Device memory capacity. Default is 6 MB."),
```
nickgg: Need to update this comment.

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -745,40 +761,46 @@ class RecommendationSystemTest : public BackendTest {
   /// Create partitions to run and compare results.
   void runPartitionedGraph(size_t numDevices, size_t memSize,
```
nickgg: Do we still need this? Can we just add it to the EE and let it do its thing?

For context this was added originally because we didn't have a nice wrapper of the runtime to test partitioning with, and now we do.

gcatron: We still do, this function calls partitioner directly for finer grained testing. That's not something the EE exposes. An alternative would be to go straight to HostManager and bypass EE but that would require its own supporting code.

lib/ExecutionEngine/ExecutionEngine2.cpp
```diff
@@ -45,6 +45,9 @@ void ExecutionEngine2::setBackendName(llvm::StringRef backend) {
   if (backend != "") {
```
rdzhabarov: what's the use case when backend passed here is actually an empty string?

gcatron: You ask a good question.  It's currently called in  ~ExecutionEngine2, but the module without a HM doesn't make much sense. A follow up Diff to refactor and simplify seems like a good idea.

gcatron: Made #3280 to track it. I'll refactor when I have fewer branches to rebase :)

PR #3277, opened 2019-07-22
Ported GradCheckTest to EE2

gcatron:
Summary: This PR ports GradCheckTest to the new EE2

Documentation:

Work on #3239 

Test Plan: verify everything builds and GradCheckTest runs and passes.

### DIFF
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -156,7 +156,7 @@ add_executable(GradCheckTest
 target_link_libraries(GradCheckTest
                       PRIVATE
                         Base
-                        ExecutionEngine
+                        ExecutionEngine2
                         Graph
                         IR
                         gtest
--- a/tests/unittests/GradCheckTest.cpp
+++ b/tests/unittests/GradCheckTest.cpp
@@ -15,7 +15,7 @@
  */
 
 #include "glow/Base/Tensor.h"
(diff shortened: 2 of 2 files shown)


tests/unittests/CMakeLists.txt
```diff
@@ -156,7 +156,7 @@ add_executable(GradCheckTest
 target_link_libraries(GradCheckTest
                       PRIVATE
                         Base
-                        ExecutionEngine
+                        ExecutionEngine2
```
bertmaher: I should have noticed this sooner, but I don't think EE2 should be a separate library.  It doesn't follow our 1:1 library to lib/ subdirectory model.

gcatron: Are you ok with it being there only temporarily? Once everything is ported to EE2 there will be a mass rename to EE and EE2 will replace EE bringing the library count back to one.

tests/unittests/GradCheckTest.cpp
```diff
@@ -29,7 +29,14 @@ using llvm::cast;
 
 class GradCheckBase : public ::testing::TestWithParam<std::string> {
 public:
-  ExecutionEngine EE_{GetParam()};
+  ExecutionEngine2 EE_{GetParam()};
```
bertmaher: Looks like you only use `EE_` in one place, `gradientCheckCrossEntropyLoss`, so I'd be inclined to remove this and just use `EET_` there

PR #3278, opened 2019-07-23
[RuntimeBundle] Disable implicit copy for RuntimeBundle

### LINKED ISSUE #3259: Runtime Bundle should not be copiable.
Recently a bug was found where the runtime bundle was copied and then collectConstants was called on the copy that was thrown away, it might make sense for the Bundle to either not have a copy constructor or invalidate the old copy.

SplitInfinity:
**Summary**
This commit disables implicit copying of `RuntimeBundles`. They should be
moved whenever possible and explicitly copied if moving is not possible.
An example of this is in the `Provisioner`, where the `RuntimeBundle` is
copied from a `CompiledFunction` into its encompassing `DAGNode`. This commit
also adds an `isValid_` flag to the `RuntimeBundle` and sets it to false
after an instance has been moved out of in order to help debug problems
arising from manipulating a stale `RuntimeBundle` instance.

**Test Plan**
`ninja check`

**Fixes**
This PR fixes #3259.

### DIFF
--- a/include/glow/Backend/BackendUtils.h
+++ b/include/glow/Backend/BackendUtils.h
@@ -64,6 +64,8 @@ class RuntimeBundle {
   size_t mutableWeightVarsMemSize_{0};
   /// Amount of memory needed for activations.
   size_t activationsMemSize_{0};
+  /// True if the RuntimeBundle is valid, false if not.
+  bool isValid_{false};
 
 public:
   /// Get Constant Weights memory size.
@@ -118,7 +120,17 @@ public:
       : symbolTable_(std::move(symbolTable)), constants_(nullptr),
         constantWeightVarsMemSize_(constWeight),
         mutableWeightVarsMemSize_(mutableWeight),
-        activationsMemSize_(activations) {}
+        activationsMemSize_(activations), isValid_(true) {}
+
+  // Explicit copy constructor and deleted assignment operator. A RuntimeBundle
+  // should be moved. It should only be copied if absolutely necessary and never
+  // implicitly.
+  explicit RuntimeBundle(const RuntimeBundle &) = default;
+  RuntimeBundle &operator=(const RuntimeBundle &) = delete;
+
+  // Move constructor and assignment operator.
+  RuntimeBundle(RuntimeBundle &&rhs);
+  RuntimeBundle &operator=(RuntimeBundle &&rhs);
 };
 } // namespace runtime
 
--- a/lib/Backend/BackendUtils.cpp
+++ b/lib/Backend/BackendUtils.cpp
@@ -37,17 +37,45 @@ static llvm::cl::opt<bool> reuseActivationsMemory(
     llvm::cl::desc("Should activation memory allocations be reused"),
     llvm::cl::init(true), llvm::cl::cat(BackendUtilsCat));
 
+glow::runtime::RuntimeBundle::RuntimeBundle(
+    glow::runtime::RuntimeBundle &&rhs) {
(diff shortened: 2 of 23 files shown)


SplitInfinity:
Need to make these changes in Habana too.

lib/Backend/BackendUtils.cpp
```diff
@@ -37,17 +37,40 @@ static llvm::cl::opt<bool> reuseActivationsMemory(
     llvm::cl::desc("Should activation memory allocations be reused"),
     llvm::cl::init(true), llvm::cl::cat(BackendUtilsCat));
 
+glow::runtime::RuntimeBundle::RuntimeBundle(
+    glow::runtime::RuntimeBundle &&rhs) {
+  *this = std::move(rhs);
+}
+
+glow::runtime::RuntimeBundle &glow::runtime::RuntimeBundle::
+operator=(glow::runtime::RuntimeBundle &&rhs) {
```
rdzhabarov: early exit for self assignment, otherwise we made an object non valid `rhs.isValid_ = false;`.

lib/Backend/BackendUtils.cpp
```diff
@@ -37,17 +37,40 @@ static llvm::cl::opt<bool> reuseActivationsMemory(
...
+  std::swap(mutableWeightVarsMemSize_, rhs.mutableWeightVarsMemSize_);
+  std::swap(activationsMemSize_, rhs.activationsMemSize_);
+  std::swap(isValid_, rhs.isValid_);
+  // rhs is not valid now that all of its contents have been stolen.
+  rhs.isValid_ = false;
+  return *this;
+}
+
 void glow::runtime::RuntimeBundle::collectConstants(const IRFunction *F) {
+  assert(isValid_);
```
rdzhabarov: DCHECK, since we are moving to GLOG. Here and in the rest of the places.

SplitInfinity:
The main motivation is to prevent what caused the bug fixed by #3258. The problematic code was something like this:
```
auto bundle = glow::runtime::RuntimeBundle::create(...);
auto f = llvm::make_unique<OpenCLCompiledFunction>(bundle, ...);
bundle.someFunction();
return f;
```
`someFunction` was supposed to be called before constructing the `OpenCLCompiledFunction`, not after. The changes made by `someFunction` were supposed to be visible to `OpenCLCompiledFunction` but its constructor takes a `const RuntimeBundle&` and copies it into a private member inside the base class `CompiledFunction`. 

This suggests that the code should have been written like so:
```
auto bundle = glow::runtime::RuntimeBundle::create(...);
bundle.someFunction();
auto f = llvm::make_unique<OpenCLCompiledFunction>(std::move(bundle), ...);
return f;
```

With the changes in this diff, any attempts to modify a `RuntimeBundle` instance after moving it should be easily caught in debug mode.

SplitInfinity:
Addressed comments, added changes for Habana backend.

include/glow/Backend/BackendUtils.h
```diff
@@ -118,7 +120,17 @@ class RuntimeBundle {
       : symbolTable_(std::move(symbolTable)), constants_(nullptr),
         constantWeightVarsMemSize_(constWeight),
         mutableWeightVarsMemSize_(mutableWeight),
-        activationsMemSize_(activations) {}
+        activationsMemSize_(activations), isValid_(true) {}
+
+  // Explit copy constructor and deleted assignment operator. A RuntimeBundle
```
gcatron: nit: Explit -> Explicit

PR #3282, opened 2019-07-23
Ported PartitionerTest to EE2

gcatron:
Summary: The PR ports the PartitionerTest to the new EE2

Documentation:

Progress on #3239 

Test Plan: verify everything builds and that PartitionerTest runs and passes.

### DIFF
--- a/tests/unittests/PartitionerTest.cpp
+++ b/tests/unittests/PartitionerTest.cpp
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 #include "glow/Partitioner/Partitioner.h"
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
 #include "glow/Graph/Graph.h"
+#include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
 
 #include "gtest/gtest.h"
 
@@ -34,7 +35,7 @@ protected:
 /// Execute a graph of functions based on the given DAG.
 static void executeDAG(DAGNode *G, Module &mod, PlaceholderBindings &bindings,
                        llvm::ArrayRef<Placeholder *> vars,
-                       llvm::ArrayRef<Tensor *> inputs) {
+                       llvm::ArrayRef<Tensor *> inputs, ExecutionEngine2 *EE) {
   std::unordered_map<std::string, Function *> name2func;
 
   for (auto *F : mod.getFunctions()) {
@@ -51,11 +52,8 @@ static void executeDAG(DAGNode *G, Module &mod, PlaceholderBindings &bindings,
     DAGNode *dag = exeList.at(curPt);
     // The root in a G is always a dummy function.
     if (curPt > 0) {
-      ExecutionEngine EE;
-      Function *func = name2func[dag->name];
-      EE.compile(CompilationMode::Infer, func);
-      updateInputPlaceholders(bindings, vars, inputs);
-      EE.run(bindings);
+      updateInputPlaceholders2(bindings, vars, inputs);
+      EE->run(bindings, dag->name);
     }
     for (int i = 0, e = dag->children.size(); i < e; i++) {
       exeList.push_back(dag->children.at(i));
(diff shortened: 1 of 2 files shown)


gcatron:
@beicy I rebased this PR and had to merge it with newer commits, can you verify I didn't accidentally remove your latest changes in the merge?

nickgg:
Why are you doing every test on two EE's now?

gcatron:
> Why are you doing every test on two EE's now?

Many of the tests involve some compile time manipulation of the graph. Since compile is now a module wide event this means we need two modules->two EEs. In this specific case, one EE to run the function without partitioning (EER). The second EE has it's function partitioned and then runs the partitioned subfunctions(EEP).

nickgg:
Makes sense, can we add that in a comment?

tests/unittests/PartitionerTest.cpp
```diff
@@ -410,10 +449,11 @@ TEST_F(PartitionerTest, Basic1Roofline) {
...
   };
-  for (auto const &p : nodeNamesMap) {
-    auto *N = p.first;
-    EXPECT_EQ(myPartitioner.getComputeTime(N), expectedComputeTime[p.second]);
-  }
+  // for (auto const &p : nodeNamesMap) {
+  //   auto *N = p.first;
+  //   EXPECT_EQ(myPartitioner.getComputeTime(N),
+  //   expectedComputeTime[p.second]);
+  // }
```
beicy: I think this is part is testing the roofline. Why do you comment it out?

tests/unittests/PartitionerTest.cpp
```diff
@@ -435,13 +475,13 @@ TEST_F(PartitionerTest, Basic1Roofline) {
...
       {"fc_add_bias4", 32},
   };
-  for (auto const &p : nodeNamesMap) {
-    auto *N = p.first;
-    EXPECT_EQ(myPartitioner.getMemUsage(N), expectedMemUsage[p.second]);
-  }
+  // for (auto const &p : nodeNamesMap) {
+  //   auto *N = p.first;
+  //   EXPECT_EQ(myPartitioner.getMemUsage(N), expectedMemUsage[p.second]);
+  // }
```
beicy: Same. I think the checking should be here.

tests/unittests/PartitionerTest.cpp
```diff
@@ -175,67 +184,77 @@ TEST_F(PartitionerTest, Basic1) {
...
-  Partitioner myPartitioner(&mod_, devices, /* saturateHost */ true);
+  Partitioner myPartitioner(&EEP.getModule(), devices, /* saturateHost */ true);
   CompilationContext cctx;
   auto err = myPartitioner.Partition(cctx);
   EXPECT_FALSE(errToBool(std::move(err)));
   DAGListTy dagList = std::move(myPartitioner.getPartitionResult());
-  EXPECT_EQ(mod_.getFunctions().size(), 2);
-  EXPECT_EQ(dagList.size(), 1);
-  ASSERT_TRUE(checkSaveNode(mod_));
+  ASSERT_EQ(EEP.getModule().getFunctions().size(), 2);
```
beicy: I don't think this check is "fatal". Even it fails, we can still check the dagList.size() and then the saveNode. So I still prefer "EXPECT_EQ" here. I think the last check before running the DAG (i.e. checkSaveNode(EEP.getModule())) should be ASSERT_TRUE. But the checks before that (i.e. the ones checking the dag structure) could be EXPECT_TRUE.

tests/unittests/PartitionerTest.cpp
gcatron: fixed now

tests/unittests/PartitionerTest.cpp
gcatron: fixed too.

tests/unittests/PartitionerTest.cpp
```diff
@@ -246,145 +266,174 @@ TEST_F(PartitionerTest, Basic2) {
...
+  bindings_.clear();
+  bindings_.allocate(EER.getModule().getPlaceholders());
+  updateInputPlaceholders2(bindings_, {bindings_.getPlaceholderByName("input")},
+                           {&in});
+  EER.run(bindings_);
+
+  // Since the partitioner will look at all nodesin the function post
+  // optimization and lowering, we need to do so here for the same list of
+  // nodes.
+  auto backend = createBackend(EEP.getBackendName());
```
beicy: memory leak?

PR #3283, opened 2019-07-23
[PyTorch] Add support for operator reshape

zrphercule:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -445,6 +445,33 @@ void PyTorchModelLoader::loadMaxPool2d(const torch::jit::Node *ptNode) {
   addGlowNodeValue(outputs[0], output);
 }
 
+void PyTorchModelLoader::loadReshape(const torch::jit::Node *ptNode) {
```
rdzhabarov: General question.

Could any of the load methods fail, like some of the params are not supported etc?
In that case I'm wondering why it's all `void`.
cc: @jackm321

zrphercule: Good point, I think we need error checking mechanism.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -445,6 +445,33 @@ void PyTorchModelLoader::loadMaxPool2d(const torch::jit::Node *ptNode) {
...
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  // Indexes of aten::reshape inputs.
+  struct Inputs {
+	enum {
```
jackm321: formatting looks weird here?

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Yeah, we will add llvm::Errors like in c2 model loader. It's on the todo list :)

torch_glow/tests/utils.py
```diff
@@ -31,3 +31,31 @@ def jitVsGlow_(f_torch, f_glow, *inputs):
     assert num_glow_nodes == 1, "Expected exactly 1 Glow node, found {}".format(num_glow_nodes)
 
     assert torch.allclose(torch_res, glow_res, atol=01e-6)
+
+def jitScriptVsGlow(f_torch, f_glow, *inputs):
```
jackm321: Why is this needed? Can you add comments to it?

torch_glow/src/PyTorchModelLoader.cpp
jfix71: Hm, I guess we're not running the clang-format CI check in here...?

jackm321: Oh damn good catch, lets fix that

jackm321: https://github.com/pytorch/glow/pull/3304

PR #3287, opened 2019-07-24
[GraphOptimizer] Drop dependency on Backends.

artemrakhov:
Backend may want to call GraphOptimizer, that's why GraphOptimizer should not depend on Backends. Otherwise it may lead to cyclic dependency.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/CMakeLists.txt
+++ b/lib/Optimizer/GraphOptimizer/CMakeLists.txt
@@ -8,8 +8,6 @@ add_library(GraphOptimizer
 
 target_link_libraries(GraphOptimizer
                       PRIVATE
-                        Backend
-                        Backends
                         Converter
                         Graph
                         GraphOptimizerPipeline


jfix71:
In what circumstances does a Backend want to call the GraphOptimizer?

ConstantFolding depends on Backends, because it creates an Interpreter backend to do the constant folding on. So I'm not sure how this is even linking correctly...

artemrakhov:
> In what circumstances does a Backend want to call the GraphOptimizer?

Suppose that Backend operates on Graph level and inserts some transformation nodes. It may need to call Dead Code Elimination after; it may need to sink transposes below that nodes.

Maybe GraphOptimizer optimizer should depend on Interpreter explicitly, not all Backends?

rdzhabarov:
> GraphOptimizer should not depend on Backends. 

Currently there are legitimate use cases when Optimizer depends on Backends, e.g., check on isOpSupported, the case @jfix71 mentioned etc.

One idea is to extract optimizations not dependent on Backend into a separate library and get use of that.

jfix71:
Yeah the backend also has `transformPostLowering()` which is called by the GraphOptimizer, so not sure we can avoid this dependence easily.

I think ideally we would allow the backend to provide backend-specific FunctionPasses alongside the base FunctionPasses. Then we could replace the backend's `transformPostLowering()` by `getPostLoweringPipeline()` or something which would be hooked into a PassManager, and then the backend could mix its own FunctionPasses with the base FunctionPasses.

However, this approach would make things a bit more complicated, because then each backends' optimization passes would need to be conditionally included in the GraphOptimizer if the backend was enabled. Meaning there'd be a new place for backends to implement logic outside of `ClassGen/` and `Backends/`.

In general there is a difference between the "driver" of the GraphOptimizer (i.e. calling lowering/folding, precision transformation, calling into the backend's `transformPostLowering()`), and the actual FunctionPasses implementing the optimizations themselves. Perhaps we could try to separate the two. But in the end I have a feeling that ConstantFolding is going to cause complications here.

chadrosier:
> > In what circumstances does a Backend want to call the GraphOptimizer?
> 
> Suppose that Backend operates on Graph level and inserts some transformation nodes. It may need to call Dead Code Elimination after; it may need to sink transposes below that nodes.

FWIW, this is the exact scenario we're running into.

rdzhabarov:
Closing this one until proper fix is introduced.

PR #3298, opened 2019-07-25
[NFC] Add Python code formatting with autopep8

SplitInfinity:
**Summary**
This commit adds automatic formatting of Python code in `utils` and
`torch_glow` (just the tests) using `autopep8`, a Python autoformatting tool.
`README.md` has also been updated to include installation instructions
for `autopep8` via `brew`.
    
**Test Plan**
This commit also includes the files modified by running the updated
version of `utils/format.sh`.

**Documentation**
`README.md` has also been updated to include installation instructions for `autopep8` via brew.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -67,8 +67,8 @@ else
     sudo apt-get install -y libpng-dev libgoogle-glog-dev
 fi
 
-# Install ninja and (newest version of) cmake through pip
-sudo pip install ninja cmake
+# Install ninja, (newest version of) cmake and autopep8 through pip
+sudo pip install ninja cmake autopep8
 hash cmake ninja
 
 # Build glow
--- a/utils/format.sh
+++ b/utils/format.sh
@@ -36,6 +36,8 @@ fix_format() {
     -o -name \*.cpp -print0 \
     -o -name \*.cl -print0 \
   | xargs -0 -P8 -n1 $CLANG_COMMAND -i;
+
+  autopep8 -i -r -j -1 --exclude="*.eggs" torch_glow utils
 }
 
 check_format() {
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Install the required dependencies using either [Homebrew](https://brew.sh/) or
 [MacPorts](https://www.macports.org/). If using Homebrew, run:
 
   ```bash
-  brew install cmake graphviz libpng ninja protobuf wget glog
+  brew install cmake graphviz libpng ninja protobuf wget glog autopep8
   brew install llvm@7
   ```
 
--- a/torch_glow/examples/basic_example.py
+++ b/torch_glow/examples/basic_example.py
@@ -4,13 +4,15 @@ import torch_glow
 x = torch.randn(4)
 y = torch.randn(4)
 
+
 @torch.jit.script
 def foo(a, b):
-  c = a.mul(b)
-  a = c.mul(c)
-  a = c.mul(a)
-  d = c.div(a)
-  return d
+    c = a.mul(b)
+    a = c.mul(c)
+    a = c.mul(a)
+    d = c.div(a)
+    return d
+
 
 print("original jit ir")
 print(foo.graph_for(x, y))
@@ -19,9 +21,11 @@ jit_res = foo(x, y)
 
 torch_glow.enableFusionPass()
 
+
 @torch.jit.script
 def foo_glow(a, b):
     return foo(a, b)
+
 
 print("glow jit ir")
 print(foo_glow.graph_for(x, y))
--- a/torch_glow/examples/resnet_example.py
+++ b/torch_glow/examples/resnet_example.py
@@ -7,18 +7,25 @@ import utils.torchvision_fake.resnet as resnet
 
 import argparse
 
+
 def load_image(image_path):
     image = Image.open(image_path).convert('RGB')
     transformed_image = transform_image(image)
     return torch.reshape(transformed_image, (1, 3, 224, 224))
 
 # given a PIL image, transform it to a normalized tensor for classification.
+
+
 def transform_image(image):
     image = torchvisionTransforms.resize(image, 256)
     image = torchvisionTransforms.center_crop(image, 224)
     image = torchvisionTransforms.to_tensor(image)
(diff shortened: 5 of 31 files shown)


SplitInfinity:
The only changes that reviewers should focus on are in `utils/format.sh`. The rest are all changes to the Python files in the project made by running the new version of this script.

jackm321:
Amazing, looks so much better.
I hadn't heard of black but it does look pretty good and if it's used widely internally maybe it's best to use it.

SplitInfinity:
Switched to `black`. I also had to migrate the `print` functions in `utils/imagenet_topk_accuracy_driver.py` from Python 2 style syntax to Python 3 style syntax to satisfy `black`, so watch out for that during review.

`black` needs Python 3.6+ to run, but it looks like our Circle CI uses an older version and so it can't install `black` and that's why all of the CI jobs have failed.  @rdzhabarov, what is the right thing to do here?

SplitInfinity:
Added `-q` flag to `black` invocation so that it only prints messages when it encounters an error.

bertmaher:
It looks like .circleci/config.yml uses py3.6 for the pytorch tests.  Can we use that everywhere?

bertmaher:
Oh. clang8 vs clang7 might be a problem.  Although we should still be compatible with 7, I thought...

SplitInfinity:
I formatted #3290 with `black` and `flake8`/`autopep8` are not happy with the result. Phabricator doesn't like it either. Is it really the case that we use `black` internally?

bertmaher:
Supposedly, but who knows.  Let's just use autopep.

SplitInfinity:
`black --version` shows `black, version 19.3b0` on both my laptop and devserver, so that's not it. Okay, I'll convert back to `autopep8`.

SplitInfinity:
Back to `autopep8`. My local `pip` doesn't seem to have `autopep8` available in any version older than `pip3.7` though... :'(

SplitInfinity:
So for some reason there is a timeout on CI for the CHECK_CLANG_FORMAT job, but the same command takes a few seconds on my laptop:
```
$ time ./utils/format.sh check
/usr/local/bin/clang-format
Running check format.

real	0m7.084s
user	0m39.274s
sys	0m1.988s
```

@rdzhabarov Any advice?

.circleci/build.sh
```diff
@@ -68,7 +68,7 @@ else
 fi
 
 # Install ninja and (newest version of) cmake through pip
```
jackm321: update this comment (or remove it seems unhelpful)

utils/format.sh
```diff
@@ -36,6 +36,10 @@ fix_format() {
     -o -name \*.cpp -print0 \
     -o -name \*.cl -print0 \
   | xargs -0 -P8 -n1 $CLANG_COMMAND -i;
+
+  find torch_glow/tests utils \
+    -name \*.py -print0 \
+  | xargs -0 autopep8 -j 8 -i -a;
```
jackm321: for `-j`, let's just do -1 or something right? 
> "number of parallel jobs; match CPU count if value is less than 1"

the autopep8 readme uses two `-a` (extra aggressive), should we do the same?

utils/format.sh
```diff
@@ -36,6 +36,10 @@ fix_format() {
     -o -name \*.cpp -print0 \
     -o -name \*.cl -print0 \
   | xargs -0 -P8 -n1 $CLANG_COMMAND -i;
+
+  find torch_glow/tests utils \
```
jackm321: Why not everything in torch_glow? There's also examples, setup.py, utils, etc

SplitInfinity: That also formats things like `torch_glow//.eggs/py-1.8.0-py3.7.egg/py/_version.py`

jackm321: :(
those should are basically just build artifacts though

jackm321: autopep8 has an `exclude` configuration, maybe that can be used?

SplitInfinity: It's possible that I don't know how to write globs but I can't get it to work :(

Even `autopep8 -i --exclude *.py test.py` still results in `test.py` being formatted.

bertmaher: IMO formatting the eggs is fine.
Looks like `--exclude` thing doesn't work on explicitly listed files, but does work if you use `-r`.  Maybe instead of using `find|xargs` we should just do `autopep8 -r -j 8 -i -a .`

SplitInfinity: I modified the command in the formatting script to
```
autopep8 -i -r -a -a -j -1 --exclude="*.eggs" torch_glow utils
```

SplitInfinity:
That must be causing the CHECK_CLANG_FORMAT job timeout too :O

bertmaher:
It seems OK without the `-a`s, maybe let's just do the non-aggressive version?

SplitInfinity:
@bertmaher Out of curiosity, which version of `autopep8` do you have? It terminates on my machine.

```
$ autopep8 --version
autopep8 1.4.4 (pycodestyle: 2.5.0)
$ time autopep8 -i -a -a -j -1 utils/compilation_filter.py 

real	0m0.540s
user	0m0.448s
sys	0m0.072s
```

In any case, I'll remove the `-a` option and push again.

SplitInfinity:
Removed `-a` options in an attempt to prevent CHECK_CLANG_FORMAT job from timing out.

bertmaher:
Bizarre, I've the same thing:
```
autopep8 1.4.4 (pycodestyle: 2.5.0)
```

bertmaher:
Uh, this is interesting.  If I install with pip it hangs, with pip3.7 it's fast.  Is the py2 version somehow borked?

SplitInfinity:
That's a possibility - I got mine with `pip3.7`.

SplitInfinity:
At last! I guess we'll settle for regular mode for now and turn on aggressive later if we need it.

SplitInfinity:
Phabricator is *mostly* happy. I'll land this.

jackm321:
@SplitInfinity I don't know why but this doesn't work on my mac. I pip installed autopep8 (with python3) and ran `utils/format.sh fix` and it doesn't change anything (I don't think it's cause my Python is perfect already). When I feed files to autopep8 manually with the `-d` flag still it recommends no changes. Any ideas?

PR #3299, opened 2019-07-25
Ported QuantizationTest to EE2

gcatron:
Summary:Ported QuantizationTest to EE2

Documentation:

Progress on #3239 

Test Plan: verify everything builds, and QuantizationTest runs and passes.

### DIFF
--- a/tests/unittests/QuantizationTest.cpp
+++ b/tests/unittests/QuantizationTest.cpp
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-#include "BackendTestUtils.h"
+#include "BackendTestUtils2.h"
 
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
 #include "glow/Graph/Graph.h"
 #include "glow/IR/IR.h"
 #include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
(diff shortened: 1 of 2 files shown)


tests/unittests/QuantizationTest.cpp
```diff
@@ -393,7 +396,7 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnected) {
 
   quantConfig.enableRowwise = true;
   quantConfig.assertAllNodesQuantized = true;
-  quantization::quantizeFunction(F, quantConfig, *EE.getBackend(),
+  quantization::quantizeFunction(F, quantConfig, *backend.get(),
```
bertmaher: I think the `.get()` is redundant here, and you can just do `*backend`.

tests/unittests/QuantizationTest.cpp
```diff
@@ -491,7 +495,7 @@ TEST(Quantization, enableRowwiseQuantizedFullyConnectedSymmetric) {
   quantConfig.schema = quantization::Schema::Symmetric;
   quantConfig.enableRowwise = true;
   quantConfig.assertAllNodesQuantized = true;
-  quantization::quantizeFunction(F, quantConfig, *EE.getBackend(),
+  quantization::quantizeFunction(F, quantConfig, *backend.get(),
```
bertmaher: Same thing here and a few other places.

tests/unittests/QuantizationTest.cpp
```diff
@@ -974,15 +990,16 @@ static Function *createGRUForQuantization(Module *M,
 TEST_P(Operator, end2endGRU) {
   // STEP1 - Generate the first network to record the quantization parameters.
   auto *mod = &profileEE.getModule();
+  auto *modBackend = &backendSpecificEE.getModule();
   PlaceholderBindings bindings;
-  Function *F1 = createGRUForQuantization(mod, bindings, "main");
-  Function *F2 = F1->clone("main2");
```
bertmaher: Is `clone` totally useless now, since there's no way to "inject" a Function into an EE?

jfix71: It's still used elsewhere, e.g. in `cloneFunInsideFun()` where a function is replicated inside itself many times for parallel testing purposes.

Additionally we talked previously about adding an ability to clone a Module, I don't recall the exact reason we decided against it for now, but would be useful there too.

PR #3300, opened 2019-07-25
[OpenCL] Add dilation support to convolution operator

### LINKED ISSUE #3033: [OpenCl] Add dilation support to convolution operator for OpenCL
https://github.com/pytorch/glow/pull/2932 introduced dilation support in the Convolution node along with CPU and Interpreter implementation.

Dilation was not properly handled by the OpenCL backend.
We need:
* Add OpenCL implementation for conv with dilation
* Enable OpenCL OperatorTest back https://github.com/pytorch/glow/pull/3031

Some basic info on dilated convolution is [here](https://towardsdatascience.com/review-dilated-convolution-semantic-segmentation-9d5a5bd768f5)

cc: @ksaurabh-cadence

shajrawi:
Summary:
Aligns with CPU and Interpreter convolution node implementation.

Fixes #3033

Test Plan:
`ninja test`

### DIFF
--- a/include/glow/Backends/LayoutConverter.h
+++ b/include/glow/Backends/LayoutConverter.h
@@ -36,7 +36,7 @@ Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
 
   auto *NC = F->addNode(new NCHWConvNode(
       CN->getName(), outTy, NI, NF, CN->getBias(), CN->getKernels(),
-      CN->getStrides(), CN->getPads(), CN->getGroup()));
+      CN->getStrides(), CN->getPads(), CN->getGroup(), CN->getDilation()));
   auto *NR = F->createTranspose("conv.result", NC, NCHW2NHWC);
 
   return NR;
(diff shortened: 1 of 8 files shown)


rdzhabarov:
@shajrawi I'm very surprised that you do not have to change the kernel code for OpenCL conv :)

shajrawi:
I was surprised as well: apparently the kernel already took a dilation parameter that was hard-coded to be 1. changing that "just worked" even though it was never tested :)

SplitInfinity:
As @rdzhabarov said, this covers only the "fast" convolution kernel that is used when `group = 1`. It would be nice to also update the "slow" convolution kernels to handle dilation.

rdzhabarov:
@SplitInfinity good point.
https://github.com/pytorch/glow/blob/master/lib/Backends/OpenCL/kernels.cl#L952
This does not handle it correctly.

shajrawi:
Thanks for pointing that out @SplitInfinity !
@rdzhabarov I created a large new test case for that + updated the "slow" convolution kernels to handle dilation.

PR #3302, opened 2019-07-26
[PyTorch] Convert Placeholders to Constants

jackm321:
Summary:
Allows each input of PyTorch nodes to be marked as inputs that should be preserved and converts all other inputs to glow Constants. This solves the problem that things such as conv weights are loaded as Placeholders because in PyTorch they may vary from one run to another. Glow however needs these to be Constants so that they can be offloaded to devices and run efficiently.

This is somewhat of medium-term hack and should be replaced eventually. The main issues with this are how manual marking each input is and that it doesn't do the right thing if the weights do actually change from one run to another because they will not get updated in Glow. A follow up to this PR should be to add a flag to totally enable or disable this conversion step. 

Documentation:
Doxygen

Test Plan:
`python setup.py test`
run resnet example and see that weights are now Constants.
Before
Screen Shot 2019-07-25 at 3 51 23 PM

After
Screen Shot 2019-07-25 at 3 50 07 PM


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -575,86 +581,213 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+    return;
+  }
+
   addGlowNodeValue(outputs[0], glowConstant->getOutput());
 }
 
 void PyTorchModelLoader::populateNodeLoaderMapping() {
-  nodeLoaderMapping_[at::Symbol::fromQualString("prim::Constant")] =
-      [this](const torch::jit::Node *node) { return loadConstant(node); };
+  nodeLoaderMapping_.emplace(
```
jackm321: This is not the idea way to register these, it's getting a bit unreadable, but it works for now.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -32,6 +32,21 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+/// \returns a Glow tensor with the same type and underlying data as the given
+/// Pytorch tensor \p ptT.
+glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptT) {
+  // TODO: get correct ElemKind.
+  std::vector<size_t> dims;
+  for (const auto d : ptT.sizes()) {
+    dims.push_back(d);
+  }
+
+  glow::Type ty(glow::ElemKind::FloatTy, dims);
```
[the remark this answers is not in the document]
jackm321: For now, yeah

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -32,6 +32,21 @@ glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
   return glow::Type(glow::ElemKind::FloatTy, dims);
 }
 
+/// \returns a Glow tensor with the same type and underlying data as the given
+/// Pytorch tensor \p ptT.
+glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptT) {
+  // TODO: get correct ElemKind.
+  std::vector<size_t> dims;
+  for (const auto d : ptT.sizes()) {
```
SplitInfinity:
```suggestion
  for (const auto &d : ptT.sizes()) {
```

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -575,86 +581,213 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+  auto &loaderValue = nodeLoaderMapping_.at(kind);
+  ValueMap valueMapReplacements =
+      createConstantPhReplacements(node, loaderValue.preserveInputs);
 
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::div")] =
-      [this](const torch::jit::Node *node) { return loadDiv(node); };
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::div_")] =
-      [this](const torch::jit::Node *node) { return loadDiv(node); };
+  for (auto &kv : valueMapReplacements) {
+    std::swap(valueMapReplacements[kv.first], valueMap_[kv.first]);
```
SplitInfinity:
```suggestion
    std::swap(kv.second, valueMap_[kv.first]);
```

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -575,86 +581,213 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
-      [this](const torch::jit::Node *node) { return loadRelu(node); };
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::relu_")] =
-      [this](const torch::jit::Node *node) { return loadRelu(node); };
+// TODO: add a flag to disable this.
+PyTorchModelLoader::ValueMap PyTorchModelLoader::createConstantPhReplacements(
+    const torch::jit::Node *ptNode, const std::set<size_t> &preserveInputs) {
+  const auto inputs = ptNode->inputs();
+  PyTorchModelLoader::ValueMap valueMapReplacements;
+
+  for (size_t i = 0; i < inputs.size(); i++) {
```
SplitInfinity:
```suggestion
  for (size_t i = 0, e = inputs.size(); i < e; i++) {
```

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -575,86 +581,213 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+      }));
+
+  nodeLoaderMapping_.emplace(
+      at::Symbol::fromQualString("aten::mul"),
+      NodeLoaderInfo({0, 1}, [this](const torch::jit::Node *node) {
+        return loadMul(node);
+      }));
+  nodeLoaderMapping_.emplace(
+      at::Symbol::fromQualString("aten::mul_"),
+      NodeLoaderInfo({0, 1}, [this](const torch::jit::Node *node) {
```
SplitInfinity:
```suggestion
      NodeLoaderInfo({glow::AddNode::InputIndices::LHSIdx, glow::AddNode::InputIndices::RHSIdx}, [this](const torch::jit::Node *node) {
```
I doubt the mapping of indices to inputs will change, but in case it does, I think it would be better to refer to the indices using the enums from `AutogenNodes.h`.

torch_glow/src/PyTorchModelLoader.cpp
SplitInfinity: I think this was changed to a static map in another diff, so you'll need to rebase.

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -24,6 +24,27 @@
 
 /// Loads PyTorch JIT IR subgraphs as a Glow Function.
 class PyTorchModelLoader {
+
+  /// Contains information about each JIT IR node regarding how to load it,
+  /// which inputs should be converted to Constants, etc.
+  struct NodeLoaderInfo {
+    /// Indices on node inputs that should not get converted to Constants
+    /// becuase they are "real" inputs that vary across runs and not weights.
+    std::set<size_t> preserveInputs;
```
SplitInfinity:
```suggestion
    std::unordered_set<size_t> preserveInputs;
```

torch_glow/src/PyTorchModelLoader.cpp
putivsky: I have landed different registration, rebase is required.

torch_glow/src/PyTorchModelLoader.cpp
SplitInfinity: Also, why do you need to do this before *and* after loading the node? Shouldn't you do it only after loading the node?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -575,86 +581,213 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::add_")] =
-      [this](const torch::jit::Node *node) { return loadAdd(node); };
+  loaderValue.loaderFunction(node);
 
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::sub")] =
-      [this](const torch::jit::Node *node) { return loadSub(node); };
-  nodeLoaderMapping_[at::Symbol::fromQualString("aten::sub_")] =
-      [this](const torch::jit::Node *node) { return loadSub(node); };
+  for (auto &kv : valueMapReplacements) {
+    std::swap(valueMapReplacements[kv.first], valueMap_[kv.first]);
```
SplitInfinity:
```suggestion
    std::swap(kv.second, valueMap_[kv.first]);
```

torch_glow/src/PyTorchModelLoader.h
jfix71: Or `llvm::SmallSet`.

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -24,6 +24,27 @@
 
 /// Loads PyTorch JIT IR subgraphs as a Glow Function.
 class PyTorchModelLoader {
+
+  /// Contains information about each JIT IR node regarding how to load it,
+  /// which inputs should be converted to Constants, etc.
+  struct NodeLoaderInfo {
+    /// Indices on node inputs that should not get converted to Constants
+    /// becuase they are "real" inputs that vary across runs and not weights.
```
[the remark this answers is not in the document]
jfix71: Can you note here that these refer to the indices of the Glow node's input and not the PyTorch input?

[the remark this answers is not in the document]
jackm321: They're not refering to Glow node inputs, they are the PT inputs

torch_glow/src/PyTorchModelLoader.cpp
jackm321: why? ref to int always seems weird to me

torch_glow/src/PyTorchModelLoader.cpp
jackm321: These aren't Glow inputs though

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Technically we only know that an input should be constant-ified for a given node so to avoid contant-ifying other node's inputs I swap back after loading that node is done. In practice there isn't a difference probably because for example conv weights shouldn't get reused as inputs, it doesn't make sense but it's just to be correct.

torch_glow/src/PyTorchModelLoader.h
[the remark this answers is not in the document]
jfix71: Ah. Then I'm wondering if we could move the enums for the inputs that are currently in `load*()` to some header to reuse when setting `preserveInputs`. Or an anon namespace in `PyTorchModelLoader.cpp`. Not a huge deal I guess, but raw indices are my pet peeve, it's not readable and error prone.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Any suggestion on this are welcome. I could make each node loader a struct with an enum for inputs that get's used in loading and for this, but it seems like overkill

torch_glow/src/PyTorchModelLoader.cpp
SplitInfinity: Just personal preference. It's not a huge deal.

torch_glow/src/PyTorchModelLoader.cpp
SplitInfinity: I think it would be useful to add a comment about this, then.

torch_glow/src/PyTorchModelLoader.cpp
SplitInfinity: Oh okay hmm...we have structs like these
```
  // Indexes of aten::_convolution inputs.
  struct Inputs {
    enum {
      input = 0, // NCHW
      weights = 1,
      bias = 2,
      stride = 3,
      padding = 4,
      dilation = 5,
      transposed = 6,
      output_padding = 7,
      groups = 8,
      benchmark = 9,
      deterministic = 10,
      cudnn_enabled = 11
    };
  };
```
defined the scope of loader functions already. I think we should pull them all out into a separate header.

jackm321:
@jfix71 
>It seems like we will always mark specific inputs as Constant or not on the Node, right? So if we have e.g. a MatMul we aren't able to distinguish whether the LHS/RHS inputs come from a Constant or PH?

For now, yeah

torch_glow/src/PyTorchModelLoader.h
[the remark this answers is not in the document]
jackm321: fair enough, I'll create static enums for each loader function and use those

PR #3303, opened 2019-07-26
[PyTorch] Add support for elementwise max operator

SplitInfinity:
**Summary**
This commit adds support for the elementwise max operator to the
PyTorchModelLoader so that `torch.max(a, b)` can be executed on Glow.

**Test Plan**
This commit adds a unit test that runs a graph with two max nodes on Glow and
the PyTorch CPU backend and checks that the results are identical.

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -197,6 +197,18 @@ void PyTorchModelLoader::loadSub(const torch::jit::Node *ptNode) {
   addGlowNodeValue(outputs[0], glowNode->getResult());
 }
 
+void PyTorchModelLoader::loadMax(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  glow::NodeValue lhs = getGlowNodeValue(inputs[0]);
+  glow::NodeValue rhs = getGlowNodeValue(inputs[1]);
+  glow::MaxNode *glowNode = f_->createMax("max", lhs, rhs);
+  addGlowNodeValue(outputs[0], glowNode->getResult());
+}
+
 void PyTorchModelLoader::loadRelu(const torch::jit::Node *ptNode) {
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -197,6 +197,18 @@ void PyTorchModelLoader::loadSub(const torch::jit::Node *ptNode) {
   addGlowNodeValue(outputs[0], glowNode->getResult());
 }
 
+void PyTorchModelLoader::loadMax(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 2);
+  assert(outputs.size() == 1);
+
+  glow::NodeValue rhs = getGlowNodeValue(inputs[0]);
```
jackm321: This is lhs, I'm an idiot and messed this up sorry

torch_glow/tests/nodes/max_test.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+import torch_glow
+from tests.utils import jitVsGlow
+
+
+# Test of the PyTorch max Node on Glow.
+
```
jackm321: too much spacing here

SplitInfinity: Do you dare defy `autopep8`?

torch_glow/src/PyTorchModelLoader.cpp
jackm321: https://github.com/pytorch/glow/pull/3305/files
Not my proudest PR

torch_glow/tests/nodes/max_test.py
jackm321: Is that really what python comments are supposed to look like? Seems too weird. Are we supposed to use the """ inside the function style or something (https://www.python.org/dev/peps/pep-0008/#id33)?

SplitInfinity: Yep, docstrings in Python are usually written after the function prototype:
```
def max_test(...):
"""Tests max node"""
```

PR #3304, opened 2019-07-26
[PyTorch][Easy] Add torch_glow directory to format.sh

jackm321:
Summary:
as title

Documentation:
n/a

Test Plan:
mess up formatting then run `./utils/format.sh fix` and it gets fixed


jackm321:
Added in another pr

PR #3306, opened 2019-07-26
[PyTorch] Add support for sqrt operator

SplitInfinity:
**Summary**
This commit adds support for the elementwise sqrt operator to the
`PyTorchModelLoader` so that `torch.sqrt(a)` can be executed on Glow.

**Test Plan**
This commit adds a unit test that runs a graph with two sqrt nodes on Glow and
the PyTorch CPU backend and checks that the results are identical.

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -231,6 +231,17 @@ void PyTorchModelLoader::loadExp(const torch::jit::Node *ptNode) {
   addGlowNodeValue(outputs[0], glowNode->getResult());
 }
 
+void PyTorchModelLoader::loadSqrt(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  assert(inputs.size() == 1);
+  assert(outputs.size() == 1);
+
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -627,6 +638,9 @@ void PyTorchModelLoader::populateNodeLoaderMapping() {
   nodeLoaderMapping_[at::Symbol::fromQualString("aten::exp")] =
       [this](const torch::jit::Node *node) { return loadExp(node); };
 
+  nodeLoaderMapping_[at::Symbol::fromQualString("aten::sqrt")] =
```
jackm321: there's also sqrt_

SplitInfinity: What's the difference? How do I get that to show up in the PT IR instead of `sqrt`?

SplitInfinity:
Rebased

torch_glow/src/PyTorchModelLoader.cpp
jackm321: sqrt_ is inplace (all foo_ are in place version of foo). You can just call b.sqrt_().

SplitInfinity:
Added test for inplace version.

PR #3307, opened 2019-07-26
Update error messages to be more descriptive

### LINKED ISSUE #2647: Unhelpful error message: file: ../../lib/Importer/Caffe2ModelLoader.cpp line: 175 message: Can't find the model or network files.
Would be nice to print what model/network it is looking for.  Also, are "model" and "network" different things here?  Could we pick one word or the other?

SameerAsal:
Add the name of the network or model we are trying to load to the log
  messages we print at failures.

 Fixes #2647 

Test Plan:
 Ran ninja test.
  Checked that the network name gets printed when an error happens at loading.

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -186,7 +186,9 @@ ONNXModelLoader::loadProto(const void *onnxModel, size_t onnxModelSize) {
 llvm::Expected<ONNX_NAMESPACE::ModelProto>
 ONNXModelLoader::loadProto(const std::string &filename) {
   std::ifstream ff(filename, std::ios::in | std::ios::binary);
-  RETURN_ERR_IF_NOT(ff, "Can't find the model or network files.",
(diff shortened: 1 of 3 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -186,7 +186,9 @@ ONNXModelLoader::loadProto(const void *onnxModel, size_t onnxModelSize) {
 llvm::Expected<ONNX_NAMESPACE::ModelProto>
 ONNXModelLoader::loadProto(const std::string &filename) {
   std::ifstream ff(filename, std::ios::in | std::ios::binary);
-  RETURN_ERR_IF_NOT(ff, "Can't find the model or network files.",
+  RETURN_ERR_IF_NOT(ff,
+                    strFormat("Can't find the model or network files for .",
```
bertmaher: Looks like a missing `%s` here

SameerAsal:
Addressed comments.

PR #3309, opened 2019-07-26
Ported SparseLengthsSumTest to EE2

gcatron:
Summary:Ported SparseLengthsSumTest to EE2

Documentation:

Progress on #3239 

Test Plan: Verify everything builds, Verify SparseLengthsSumTest runs and passes. (I'm not sure this is actually tested in CI. Verified to run on my laptop.)

### DIFF
--- a/tests/stress/SparseLengthsSumTest.cpp
+++ b/tests/stress/SparseLengthsSumTest.cpp
@@ -14,94 +14,148 @@
  * limitations under the License.
  */
 
-#include "BackendTestUtils.h"
+#include "BackendTestUtils2.h"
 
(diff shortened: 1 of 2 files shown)


tests/stress/SparseLengthsSumTest.cpp
```diff
@@ -14,94 +14,148 @@
...
 
 #include <glog/logging.h>
 #include <gtest/gtest.h>
 
 using namespace glow;
 
-class SparseLengthsSum : public BackendTest {};
+class SparseLengthsSum : public BackendTest {
+public:
+  ~SparseLengthsSum() override{};
```
rdzhabarov: dtor in the base class needs to be virtual.

PR #3310, opened 2019-07-26
Fixing BatchNormalization initialization.

putivsky:
Summary: Variance cannot be initialized to zero.

Reviewed By: jfix71

Differential Revision: D16496669


lib/Graph/Graph.cpp
```diff
@@ -2060,23 +2060,22 @@ BatchNormalizationNode *Function::createBatchNormalization(
   // Allocate the learnable parameters beta and gamma.
```
rdzhabarov: can you fix gamma here.

PR #3312, opened 2019-07-26
[RecSys] Compare against Interpreter. Use Xaiver initialization based on tensor size.

jfix71:
Summary: A couple improvements to RecSys:
- If we're executing on a non-Interpreter backend, then run on the Interpreter and make sure the results are equal.
- Randomly initialize floating point tensors via Xaiver initialization, using twice the size of the tensor as the filter size. Hopefully this will reduce overflows as the test uses different sizes/configurations.

Test Plan: All tests pass.

### DIFF
--- a/tests/unittests/RecommendationSystemTest.cpp
+++ b/tests/unittests/RecommendationSystemTest.cpp
@@ -999,23 +999,3 @@ TEST_P(RecommendationSystemTest, RecSys_FP32_Gather_Weights) {
              /* convertToFP16 */ false,
              /* gatherWeights */ true);
 }
-
-/// Test gathering weights for SLWS.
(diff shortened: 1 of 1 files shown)


jackm321:
Nice!

jfix71:
I've decided to just remove the medium test `RecSys_FP32_Medium_Gather_Weights`, to reduce default time for testing. If anyone wants to run this same test they can get something basically the same via:

```
./tests/RecommendationSystemTest --gtest_filter=RecSys/RecommendationSystemTest.RecSys_FP32_Gather_Weights/* -embedding-table-sizes=800000,600000,700000,900000,1200000 -embedding-table-counts=3,3,3,3,3`
```

PR #3313, opened 2019-07-29
Added error handling to PyTorchModelLoader class.

putivsky:
Summary:
PyTorchModelLoader class should return error instead calling assert() and kill the process.
Caller will make a decision what to do on error.

Differential Revision: D16537764

### DIFF
--- a/torch_glow/src/PyTorchLoaderTest.cpp
+++ b/torch_glow/src/PyTorchLoaderTest.cpp
@@ -59,8 +59,11 @@ void convertJitGraphToGlowFunction(torch::jit::Stack &stack,
   // Load JIT Graph into Glow Function.
   std::vector<glow::Placeholder *> inputPlaceholders;
   std::vector<glow::Placeholder *> outputPlaceholders;
-  PyTorchModelLoader loader(f, graph, inputs, inputPlaceholders,
-                            outputPlaceholders);
+  llvm::Error errPtr = PyTorchModelLoader::loadJITGraph(
+      f, graph, inputs, inputPlaceholders, outputPlaceholders);
+
+  EXPECT_FALSE(errPtr);
+
   // Remove from stack input parameters.
   torch::jit::drop(stack, numInputs);
   // Lookup placeholders for the output shapes.
@@ -103,8 +106,9 @@ TEST(PyTorchModelLoader, Model) {
   vec.insert(vec.begin(), module->module_object());
 
   at::ArrayRef<torch::jit::IValue> inputs(vec);
-  PyTorchModelLoader loader(*f, *subgraph, inputs, inputPlaceholders,
-                            outputPlaceholders);
+  errPtr = PyTorchModelLoader::loadJITGraph(
+      *f, *subgraph, inputs, inputPlaceholders, outputPlaceholders);
+  EXPECT_FALSE(!errPtr); // TODO
 }
 
 TEST(PyTorchModelLoader, Fuser) {
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -16,6 +16,7 @@
 
 #include "PyTorchModelLoader.h"
 
+#include "glow/Support/Error.h"
 #include "glow/Support/Support.h"
 
 #include <torch/csrc/jit/ir.h>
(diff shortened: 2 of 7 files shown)


torch_glow/src/PyTorchLoaderTest.cpp
```diff
@@ -103,8 +106,9 @@ TEST(PyTorchModelLoader, Model) {
   vec.insert(vec.begin(), module->module_object());
 
   at::ArrayRef<torch::jit::IValue> inputs(vec);
-  PyTorchModelLoader loader(*f, *subgraph, inputs, inputPlaceholders,
-                            outputPlaceholders);
+  errPtr = PyTorchModelLoader::loadJITGraph(
+      *f, *subgraph, inputs, inputPlaceholders, outputPlaceholders);
+  EXPECT_FALSE(!errPtr); // TODO
```
jackm321: What is the TODO here?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -16,10 +16,13 @@
 
 #include "PyTorchModelLoader.h"
 
+#include "glow/Support/Error.h"
 #include "glow/Support/Support.h"
 
 #include <torch/csrc/jit/ir.h>
 
+using namespace glow;
```
jackm321: You're doing this because the things in Error.h are in `namespace glow` right?
I don't really want to do this because then types like `Tensor` become a bit harder to read if they aren't qualified (is it a PT or a Glow Tensor).
Maybe we could take the things in Error.h out of `namespace glow` or maybe we could just do a couple of `using foo = glow::foo` up top here for the things in Error.h. Thoughts?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
 }
 
 /// Downcast a double to a float.
-inline float to32Bit(double val) {
-  assert(val <= std::numeric_limits<float>::max());
-  assert(val >= std::numeric_limits<float>::lowest());
-  return static_cast<float>(val);
+inline llvm::Expected<float> to32Bit(double val) {
+  RETURN_ERR_IF_NOT(val <= std::numeric_limits<float>::max() ||
```
jackm321: Can we put the values in here

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+  return llvm::Expected<float>(static_cast<float>(val));
 }
 
 /// Downcast an int64_t to a int32_t.
-inline int32_t to32Bit(int64_t val) {
-  assert(val <= std::numeric_limits<int32_t>::max());
-  assert(val >= std::numeric_limits<int32_t>::lowest());
-  return static_cast<int32_t>(val);
+inline llvm::Expected<int32_t> to32Bit(int64_t val) {
+  RETURN_ERR_IF_NOT(val <= std::numeric_limits<int32_t>::max() ||
```
jackm321: Values

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
-  assert(handle.dims().size() == 1 && "Expected a 1d Glow Tensor");
-  assert(handle.size() == 1 ||
-         handle.size() == targetSize &&
-             "Expected input size to be either 1 or the target size");
-  std::vector<OutT> out;
+template <typename T, typename OutT>
+llvm::Error expandParamIfNeeded(const glow::Handle<T> &handle,
+                                size_t targetSize, std::vector<OutT> &out) {
+  RETURN_ERR_IF_NOT(handle.dims().size() == 1, "Expected a 1d Glow Tensor.");
+  RETURN_ERR_IF_NOT(handle.size() == 1 || handle.size() == targetSize,
```
jackm321: print targetSize value

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
-template <typename T, typename OutT = T>
-std::vector<OutT> expandParamIfNeeded(const glow::Handle<T> &handle,
-                                      size_t targetSize) {
-  assert(handle.dims().size() == 1 && "Expected a 1d Glow Tensor");
-  assert(handle.size() == 1 ||
-         handle.size() == targetSize &&
-             "Expected input size to be either 1 or the target size");
-  std::vector<OutT> out;
+template <typename T, typename OutT>
+llvm::Error expandParamIfNeeded(const glow::Handle<T> &handle,
```
jackm321: Any reason to not just use Expected<OutT>?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
 /// Given a Glow Handle \p Handle returns the first element in the Tensor
 /// of that Handle, asserting it is equivalent to all other elements in the
 /// Tensor.
-template <typename T> T contractParamIfNeeded(const glow::Handle<T> &handle) {
-  assert(handle.size() > 0);
-  const T firstElem = handle.raw(0);
+template <typename T, typename OutT>
+llvm::Error contractParamIfNeeded(const glow::Handle<T> &handle,
+                                  OutT &firstElem) {
+  RETURN_ERR_IF_NOT(handle.size() > 0, "Handle size must be > 0");
```
jackm321: add handle.size() to message

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -33,69 +36,100 @@ inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
...
+    RETURN_ERR_IF_NOT(handle.raw(i) == firstElem,
+                      "All elements must be equal.");
   }
-  return firstElem;
+  return llvm::Error::success();
+}
+
+/// Given Node inputs and outputs, check the expected sizes.
+template <typename T>
+llvm::Error checkInputAndOutputSizes(const T &inputs, size_t inputsSize,
```
jackm321: Awesome

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -140,165 +174,191 @@ bool PyTorchModelLoader::isNodeSupported(const torch::jit::Node *node) {
...
 
-void PyTorchModelLoader::loadMul(const torch::jit::Node *ptNode) {
+llvm::Error PyTorchModelLoader::loadMul(const torch::jit::Node *ptNode) {
   auto inputs = ptNode->inputs();
   auto outputs = ptNode->outputs();
-  assert(inputs.size() == 2);
-  assert(outputs.size() == 1);
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, 2, outputs, 1));
+
+  glow::NodeValue lhs;
```
jackm321: thanks for swapping these :)

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -411,20 +480,24 @@ void PyTorchModelLoader::loadBatchNorm(const torch::jit::Node *ptNode) {
...
+  auto trainingHandle = glow::Handle<bool>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      trainingHandle, getGlowConstantHandle<bool>(inputs[Inputs::training]));
+  RETURN_ERR_IF_NOT(trainingHandle.size() == 1, "Training must be 1D vector.");
+  RETURN_ERR_IF_NOT(trainingHandle.raw(0) == false,
+                    "Training must be scalar with false value.");
+
+  glow::NodeValue input;
+  ASSIGN_VALUE_OR_RETURN_ERR(input, getGlowNodeValue(inputs[Inputs::input]));
+  RETURN_ERR_IF_NOT(input.dims().size() == 4,
```
jackm321: Would be ideal to print values with these things

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -649,55 +740,88 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+        return;
+      }
+      inputPlaceholders.push_back(PH);
     }
   }
 
   // Nodes are topologically sorted.
   for (auto node : subgraph.nodes()) {
-    if (!loadNode(node)) {
+    if (!!(error = loadNode(node))) {
```
jackm321: should get cast to bool without the !! right?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -649,55 +740,88 @@ void PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
+  return addGlowNodeValue(outputs[0], glowConstant->getOutput());
+}
+
+/*static*/
+llvm::Error PyTorchModelLoader::loadJITGraph(
+    glow::Function &F, torch::jit::Graph &subgraph,
+    at::ArrayRef<torch::jit::IValue> &inputs,
+    std::vector<glow::Placeholder *> &inputPlaceholders,
+    std::vector<glow::Placeholder *> &outputPlaceholders) {
+  llvm::Error error = llvm::Error::success();
```
jackm321: `MARK_ERR_CHECKED`

jackm321:
Awesome, will be great to have actual error handling here!

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -33,99 +34,130 @@ class PyTorchModelLoader {
...
   /// will succeed because determining this requires more informations such as
   /// shape info that isn't yet available when this is run.
   static bool isNodeSupported(const torch::jit::Node *node);
 
+  /// Takes a glow::Function \p F, a jit::Graph \p subgraph to load, and a
+  /// stack of \p inputs for the subgraph to be loaded. Output parameters
+  /// \p inputPlaceholders and \p outputPlaceholders are filled out.
+  /// \returns error on failure
+  static llvm::Error
+  loadJITGraph(glow::Function &F, torch::jit::Graph &subgraph,
```
mortzur: Looking good. 
What are the different use cases for loading PT-JIT subgraph using the CTOR and using this static method?

jackm321: @mortzur the constructor does all of the loading which means it can fail but constructors can't return anything so we have to pass in an llvm::Error that gets assigned to if any error occurs. In order to avoid all of the unpleasantness that this causes, this makes the constructor private and just calls it from from loadJITGraph which aslo manages the error creation and returning. Best of both worlds (functional-style errors, stateful class for building graph)

torch_glow/src/PyTorchLoaderTest.cpp
putivsky: Currently not all operators are supported by Glow for resnet18, so the failure is expected, but once all operators are implemented, unit test will fail and TODO is:
EXPECT_FALSE(!errPtr); => EXPECT_FALSE(errPtr);

torch_glow/src/PyTorchModelLoader.cpp
putivsky: Yes, the problem with Error.h, which should be fixed first. I will take care of it in another diff, then we can go back to this code and remove "using namespace glow;"

jackm321: Ok sounds good

torch_glow/src/PyTorchModelLoader.cpp
putivsky: Template method usually deduces template types from input/output parameters, therefore sending output template parameter explicitly (not as a return value) makes compiler happy.

torch_glow/src/PyTorchModelLoader.cpp
putivsky: Can you explain how to use MARK_ERR_CHECKED?

jackm321: just add `MARK_ERR_CHECKED(error)` below `llvm::Error error = llvm::Error::success();`
`llvm::Error::success()` creates a new Error and assigns it to `error` but then when `error` is passed by ref into the PyTorchModelLoader constructor, it will be overwritten and the original Error created by `llvm::Error::success()` will be destroyed and all llvm::Errors have to be checked before they're destroyed so you can just check it up front (kinda a hack) using `MARK_ERR_CHECKED`

PR #3314, opened 2019-07-29
[PyTorch] Add support for reciprocal operator

SplitInfinity:
**Summary**
This commit adds support for the elementwise reciprocal operator to the
`PyTorchModelLoader` so that `torch.reciprocal(a)` and `a.reciprocal_()`
(inplace version) can be executed on Glow.

**Test Plan**
This commit adds two unit tests (one for the inplace version and one for
the out of place version) that runs a graph with two reciprocal nodes on
Glow and the PyTorch CPU backend and checks that the results are
identical.

```
tests/nodes/reciprocal_test.py::test_reciprocal PASSED                                                                                                                            [ 66%]
tests/nodes/reciprocal_test.py::test_inplace_reciprocal PASSED                                                                                                                    [ 70%]
```

### DIFF
--- a/torch_glow/tests/nodes/reciprocal_test.py
+++ b/torch_glow/tests/nodes/reciprocal_test.py
@@ -1,0 +1,28 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+import torch_glow
+from tests.utils import jitVsGlow
+
+
+def test_reciprocal():
+    """Test of the PyTorch reciprocal Node on Glow."""
+
+    def reciprocal(a):
+        return torch.reciprocal(a + a)
+
+    x = torch.randn(4)
+
+    jitVsGlow(reciprocal, x)
+
+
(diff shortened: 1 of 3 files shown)


torch_glow/tests/nodes/reciprocal_test.py
```diff
@@ -0,0 +1,28 @@
...
+import torch_glow
+from tests.utils import jitVsGlow
+
+
+def test_reciprocal():
+    """Test of the PyTorch reciprocal Node on Glow."""
+
+    def reciprocal(a):
+        c = torch.reciprocal(a)
+        return torch.reciprocal(c)
```
jackm321: It would be better to use another op in here like `return torch.reciprocal(a+a)` because test can't distinguish between reciprocal of reciprocal and noop.

SplitInfinity:
Rebased and addressed comments.

PR #3315, opened 2019-07-29
[PyTorch] Unbreak torch_glow

SplitInfinity:
**Summary**
`torch_glow` was accidentally broken by #3308 which removed
`PyTorchModelLoader::populateNodeLoaderMapping` and created a static map
in its place. Mappings for for `exp`, `sqrt` and `max` operators were added and
merged while PR #3308 was still a work in progress and they must have
gotten lost in the process of resolving merge conflicts.

**Test Plan**
The unit tests for `exp`, `sqrt` and `max` pass again.

### DIFF
--- a/torch_glow/src/CMakeLists.txt
+++ b/torch_glow/src/CMakeLists.txt
@@ -26,7 +26,7 @@ target_link_libraries(_torch_glow
                         ExecutionEngine
                         Graph
                         Importer
-                        Interpreter
+			 Support
                         Backends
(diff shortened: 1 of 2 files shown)


SplitInfinity:
@jackm321 

Hmm, looks like reverting the change to `target_link_libraries` for `torch_glow` fixed the PYTORCH CI job?

jackm321:
That seems like the only relevant change so makes sense.

PR #3316, opened 2019-07-29
[Do Not Merge] POCL debugging

gcatron:
Summary: Test branch for tracking down POCL double free in CI. This modifies OCLTest and resnet-runtime.

Documentation:

Test Plan:


gcatron:
Closed, looks like any call to OpenCL causes ASAN to report a double-free error. Will create a minimal example and file a bug on POCL

PR #3318, opened 2019-07-29
[Partitioner] Partitioner Refactoring 2

beicy:
Summary:
In this PR, the following methods are updated ( the way of how to calculate the memory usage/computation for each node is not changed)  :
1. In previous version, Partitioner had a map to store the node and its memory usage.  Now this one is removed, and a function `uint64_t getNodeMemUsage(const Node *node) ` is added in to PartitionerUtils.cpp. That is, instead of pre-calculating all nodes' memory usage and storing it in a map, we calculate it when it is necessary.  This function is now only called in `loadBalancedPartitioning`, and for each node, it is only called once. So no need to store the value in a map.
 2. In previous version, Partitioner had a map to store the node and its computation time. Similar as the above memory usage refactoring,  a function `float getNodeComputeTime(const Node *node, const BackendInfo backendInfo)` is added to return the computation time of a node when it is necessary.  Now the backendInfo is added as a param in this function for different type of backends.
  
Documentation:
#2298 
[Optional Fixes #issue]

Test Plan: ninja test.

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -16,7 +16,7 @@
 #ifndef GLOW_PARTITIONER_PARTITIONER_H
 #define GLOW_PARTITIONER_PARTITIONER_H
 
-#include "glow/Partitioner/PartitionerUtils.h"
+#include "glow/Partitioner/PartitionerTypes.h"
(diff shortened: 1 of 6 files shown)


nrsatish:
LGTM

PR #3321, opened 2019-07-29
[runtime] Fix memory usage when reusing allocator in compile

nickgg:
Summary: In the OpenCL backend we allocate all device buffers contiguously by reusing the same memory allocator for Constants, Placeholders and Activations. This diff fixes a bug in that logic which allocated the sum of the current stage and previous stages rather than just the memory needed for the current stage (e.g. we'd set sizes to be Constants, Placeholders + Constants, Activations + Placeholders + Constants).

This means we were allocating about 2-3x as much memory as we needed on the GPU. E.g. in OCLTest::ConvOps:

before:
```
  Constants: 2496
  Placeholders: 9216
  Activations: 12928
  Total alloced: 24640
```

after:
```
  Constants: 2496 
  Placeholders: 6720
  Activations: 3712
  Total alloced: 12928
```

Gets better the bigger the Constants block is, obviously.

Documentation: n/a

Test Plan: unit tests in ASAN and Release (OCLTest, BackendCorrectnessTest & OperatorTest being the important ones).

### DIFF
--- a/include/glow/Backend/BackendUtils.h
+++ b/include/glow/Backend/BackendUtils.h
@@ -104,6 +104,13 @@ public:
                                        MemoryAllocator &placeholderAllocator,
                                        MemoryAllocator &activationsAllocator);
 
+  /// Computes offsets and total allocations for Constants, Placeholders, and
+  /// Activations to build runtime symbol table. \returns RuntimeBundle. Uses a
+  /// single allocator \p allocator and allocates all buffers contiguously in
+  /// the same block.
+  static runtime::RuntimeBundle create(const IRFunction &F,
+                                       MemoryAllocator &allocator);
+
   /// Build a runtime symbol table from a Function.  Computes Constant and
   /// Placeholder sizes, but not Activations, since Functions are unserialized.
   /// Only use this method to generate bundles for backends that do not use
--- a/lib/Backend/BackendUtils.cpp
+++ b/lib/Backend/BackendUtils.cpp
@@ -327,6 +327,14 @@ runtime::RuntimeBundle::create(const IRFunction &F,
                                MemoryAllocator &constantAllocator,
                                MemoryAllocator &placeholderAllocator,
                                MemoryAllocator &activationsAllocator) {
+
+  // If all allocators refer to the same underlying allocator, Constants,
+  // Placeholders and activations will be allocated contiguously. The maximum
+  // memory usage reported by the allocator for each kind of storage will
+  // include the memory usage of all previously allocated types of storage and
+  // needs to be adjusted accordingly.
+  bool contiguous = (&constantAllocator == &placeholderAllocator &&
+                     &constantAllocator == &activationsAllocator);
   // Handle Constants, Placeholders, and Activations, in that order.
   // Symbol table mapping symbol name to offset for runtime.
   std::map<std::string, runtime::RuntimeSymbolInfo> symbolTable;
@@ -375,6 +383,9 @@ runtime::RuntimeBundle::create(const IRFunction &F,
                    w->getName().data(), symbol.offset, symbol.size));
   }
   auto placeholderMaxSize = placeholderAllocator.getMaxMemoryUsage();
+  if (contiguous) {
+    placeholderMaxSize -= constantMaxSize;
+  }
 
   // Compute the offsets for Activations.
   for (const auto &I : F.getInstrs()) {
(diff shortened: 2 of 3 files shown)


include/glow/Backend/BackendUtils.h
```diff
@@ -102,7 +102,15 @@ class RuntimeBundle {
   static runtime::RuntimeBundle create(const IRFunction &F,
                                        MemoryAllocator &constantAllocator,
                                        MemoryAllocator &placeholderAllocator,
-                                       MemoryAllocator &activationsAllocator);
+                                       MemoryAllocator &activationsAllocator,
+                                       bool contiguous = false);
+
+  /// Computers offsets and total allocations for Constants, Placeholders, and
```
SplitInfinity:
```suggestion
  /// Computes offsets and total allocations for Constants, Placeholders, and
```

include/glow/Backend/BackendUtils.h
```diff
@@ -102,7 +102,15 @@ class RuntimeBundle {
...
+  /// Activations to build runtime symbol table. Returns RuntimeBundle. Uses a
```
SplitInfinity:
```suggestion
  /// Activations to build runtime symbol table. \returns RuntimeBundle. Uses a
```

include/glow/Backend/BackendUtils.h
```diff
@@ -102,7 +102,15 @@ class RuntimeBundle {
   static runtime::RuntimeBundle create(const IRFunction &F,
```
SplitInfinity: Consider updating the comment for this function to include the new parameter `contiguous`.

lib/Backend/BackendUtils.cpp
```diff
@@ -443,7 +445,18 @@ runtime::RuntimeBundle::create(const IRFunction &F,
...
+  }
 
   return runtime::RuntimeBundle(symbolTable, constantMaxSize,
                                 placeholderMaxSize, activationsMaxSize);
 }
+
+runtime::RuntimeBundle
+runtime::RuntimeBundle::create(const IRFunction &F,
+                               MemoryAllocator &allocator) {
+  return create(F, allocator, allocator, allocator, true);
```
SplitInfinity:
```suggestion
  return create(F, allocator, allocator, allocator, /*contiguous-*/true);
```

lib/Backend/BackendUtils.cpp
```diff
@@ -443,7 +445,18 @@ runtime::RuntimeBundle::create(const IRFunction &F,
     }
   }
   auto activationsMaxSize = activationsAllocator.getMaxMemoryUsage();
+  if (contiguous) {
+    activationsMaxSize -= constantMaxSize + placeholderMaxSize;
+    assert(constantAllocator.getMaxMemoryUsage() ==
```
SplitInfinity:
```suggestion
    DCHECK_EQ(constantAllocator.getMaxMemoryUsage(),
```

lib/Backend/BackendUtils.cpp
```diff
@@ -322,11 +322,10 @@ runtime::RuntimeBundle runtime::RuntimeBundle::create(const Function &F) {
...
 
-runtime::RuntimeBundle
-runtime::RuntimeBundle::create(const IRFunction &F,
-                               MemoryAllocator &constantAllocator,
-                               MemoryAllocator &placeholderAllocator,
-                               MemoryAllocator &activationsAllocator) {
+runtime::RuntimeBundle runtime::RuntimeBundle::create(
+    const IRFunction &F, MemoryAllocator &constantAllocator,
+    MemoryAllocator &placeholderAllocator,
+    MemoryAllocator &activationsAllocator, bool contiguous) {
```
SplitInfinity: Can we infer the value of `contiguous` without depending on the caller to pass the correct value in? i.e.
```
bool contiguous = (&constantAllocator == &placeholderAllocator) && (&constantAllocator == &activationsAllocator);
```
or something like that?

nickgg: We could, do you think its clearer?

nickgg:
Kind of wish it amended code review suggestions, oh well I'll fix it up manually.

lib/Backend/BackendUtils.cpp
SplitInfinity: I don't think it's clearer by itself without any explanation, but I think a one or two line comment above the line I suggested would suffice.

My concern is that this overload of `RuntimeBundle::create` can still be called like so
```
auto r = runtime::RuntimeBundle::create(F, allocator, allocator, allocator);
```
and the memory usage will be 2-3x as much as needed.

nickgg: OK sure, that makes sense. There's a limit to how much we can guard against people using interfaces wrong though.

lib/Backend/BackendUtils.cpp
```diff
@@ -327,6 +327,9 @@ runtime::RuntimeBundle::create(const IRFunction &F,
                                MemoryAllocator &constantAllocator,
                                MemoryAllocator &placeholderAllocator,
                                MemoryAllocator &activationsAllocator) {
+
+  bool contiguous = (&constantAllocator == &placeholderAllocator &&
```
SplitInfinity:
```suggestion
  // If all allocators refer to the same underlying allocator, Constants, Placeholders and activations will be allocated contiguously. The maximum memory usage reported by the allocator for each kind of storage will include the memory usage of all previously allocated types of storage and needs to be adjusted accordingly.
  bool contiguous = (&constantAllocator == &placeholderAllocator &&
```

Or something like that.

PR #3327, opened 2019-07-30
Define a stats reporting interface

bertmaher:
Summary:
Production users of Glow may want to log statistics.  This diff
introduces an API that allows users to implement logging using custom
components that are loosely coupled to Glow's internals.  A unit test shows the
intended usage: a stats exporter is registered at startup, and Glow will log to
all such registered exporters.

Two statistics classes are defined:
- Counters, for simple numerical values,
- Quantiles, for aggregating statics into a frequency distribution

Currently no implementation of these loggers is provided (though one may be in
the future).  For server-side usage, consider a stats-querying library such as
[fb303](https://github.com/facebookincubator/fb303).

By way of example, this diff logs the number of each device type managed by
Glow.

Differential Revision: D16558211

### DIFF
--- a/include/glow/Runtime/StatsExporter.h
+++ b/include/glow/Runtime/StatsExporter.h
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_RUNTIME_STATSEXPORTER_H
+#define GLOW_RUNTIME_STATSEXPORTER_H
+
+#include "llvm/ADT/StringRef.h"
+
+#include <vector>
+
+namespace glow {
+
+/// Interface for exporting runtime statistics.  The base implementation
+/// delegates to any subclass registered via `registerStatsExporter`.
+class StatsExporter {
+public:
+  /// Dtor.
+  virtual ~StatsExporter() {}
+
+  /// Add value to a time series.  May be called concurrently.
+  virtual void addTimeSeriesValue(llvm::StringRef key, double value) = 0;
+
+  /// Increment a counter.  May be called concurrently.
+  virtual void incrementCounter(llvm::StringRef key, int64_t value = 1) = 0;
+
+  /// Set a counter.  May be called concurrently.
+  virtual void setCounter(llvm::StringRef key, int64_t value) = 0;
+};
+
+/// Registry of StatsExporters.
+class StatsExporterRegistry final {
+public:
+  /// Add value to a time series for all registered StatsExporters.
+  void addTimeSeriesValue(llvm::StringRef key, double value);
+
+  /// Increment a counter for all registered StatsExporters.
+  void incrementCounter(llvm::StringRef key, int64_t value = 1);
+
+  /// Set a counter for all registered StatsExporters.
+  void setCounter(llvm::StringRef key, int64_t value);
+
+  /// Register a StatsExporter.
+  void registerStatsExporter(StatsExporter *exporter);
+
+private:
+  /// Registered StatsExporters.
+  std::vector<StatsExporter *> exporters_;
+};
+
+/// Global singleton StatsExporter.
+StatsExporterRegistry *Stats();
+
+} // namespace glow
+
+#endif // GLOW_RUNTIME_STATSEXPORTER_H
--- a/lib/Backends/CPU/CPUDeviceManager.h
+++ b/lib/Backends/CPU/CPUDeviceManager.h
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_CPU_CPUDEVICEMANAGER_H
 
 #include "glow/Backends/QueueBackedDeviceManager.h"
+#include "glow/Runtime/StatsExporter.h"
 
 namespace glow {
 namespace runtime {
@@ -39,10 +40,19 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
   /// This is very arbitrary for the CPU backend.
   const uint64_t functionCost_{1};
 
+  /// String constant for logging number of in-use devices.
+  static constexpr const char *kDevicesUsedCPU = "glow.devices_used.cpu";
+
 public:
-  CPUDeviceManager(const DeviceConfig &config)
+  explicit CPUDeviceManager(const DeviceConfig &config)
       : QueueBackedDeviceManager(config),
-        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {}
+        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {
+    Stats()->incrementCounter(kDevicesUsedCPU);
+  }
+
+  ~CPUDeviceManager() override {
+    Stats()->incrementCounter(kDevicesUsedCPU, -1);
+  }
 
   /// Returns the amount of memory in bytes available on the device when no
   /// models are loaded.
--- a/lib/Runtime/StatsExporter.cpp
+++ b/lib/Runtime/StatsExporter.cpp
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Runtime/StatsExporter.h"
+
+#include <vector>
+
+namespace glow {
+
+void StatsExporterRegistry::registerStatsExporter(StatsExporter *exporter) {
+  exporters_.push_back(exporter);
+}
+
+void StatsExporterRegistry::addTimeSeriesValue(llvm::StringRef key,
+                                               double value) {
+  for (auto const &exporter : exporters_) {
+    exporter->addTimeSeriesValue(key, value);
+  }
+}
+
+void StatsExporterRegistry::setCounter(llvm::StringRef key, int64_t value) {
+  for (auto const &exporter : exporters_) {
+    exporter->setCounter(key, value);
+  }
+}
+
+void StatsExporterRegistry::incrementCounter(llvm::StringRef key,
+                                             int64_t value) {
+  for (auto const &exporter : exporters_) {
+    exporter->incrementCounter(key, value);
+  }
+}
+
+StatsExporterRegistry *Stats() {
+  static auto *stats = new StatsExporterRegistry();
+  return stats;
+}
+
+} // namespace glow
--- a/tests/unittests/StatsExporterTest.cpp
+++ b/tests/unittests/StatsExporterTest.cpp
@@ -1,0 +1,90 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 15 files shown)


include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,48 @@
...
+  static void registerStatsExporter(StatsExporter *exporter);
+
+  /// Dtor.
+  virtual ~StatsExporter() {}
+
+  /// Add value to a quantile stat.
+  virtual void addQuantileValue(llvm::StringRef key, double value);
+
+  /// Set a counter.
+  virtual void incrementCounter(llvm::StringRef key, int64_t value);
```
jackm321: Maybe make 1 the default for value for convenience?

jackm321: Maybe make a note that these could be called concurrently

lib/Runtime/StatsExporter.cpp
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+void StatsExporter::incrementCounter(llvm::StringRef key, int64_t value) {
+  for (auto const &exporter : getExporters()) {
+    exporter->incrementCounter(key, value);
+  }
+}
+
+StatsExporter Stats;
```
jackm321: Might cause some destruction order problems right?
What if some other things destructor calls `Stats.incrementCounter("foo" -1)` for example but this is already destroyed? Maybe make it a pointer?

tests/unittests/StatsExporterTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+using testing::ElementsAre;
+
+class MockStatsExporter : public StatsExporter {
+public:
+  MockStatsExporter() { StatsExporter::registerStatsExporter(this); }
+
+  ~MockStatsExporter() override {}
+
+  void addQuantileValue(llvm::StringRef key, double value) override {
```
jackm321: quantiles seems a bit niche and might not be supported by every exporter, maybe it should be an optional feature for an exporter?

include/glow/Runtime/StatsExporter.h
jfix71: What's our expectation on threading here? Do we need to use atomics/mutexes?

tests/unittests/StatsExporterTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+    quantiles[key].push_back(value);
+  }
+
+  void incrementCounter(llvm::StringRef key, int64_t value) override {
+    counters[key] += value;
```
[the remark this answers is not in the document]
jackm321: I am biased from using fb303 but I think this this is fine and better than enforcing the use of const char* or enums for keys because of the the flexibility it adds.

include/glow/Runtime/StatsExporter.h
bertmaher: Right, these can absolutely be called concurrently.  I'd expect whatever loggers are registered to handle concurrency properly (in FB's case, fb303 does).

tests/unittests/StatsExporterTest.cpp
bertmaher: Hm, interesting point.  What I expect this type of counter to be used for is something where an aggregated timeseries makes sense (so, like, p95 queue latency over a 60 second window), as opposed to the other counter, which is more of a static value.  This counter type doesn't necessarily have to specifically use quantiles (although I'm planning to hook up fb303's quantile stats to it).

Maybe it should be more generically named like `addTimeseriesValue`?

Hm, also it occurs to me this class design is kind of bad in that anything inheriting from the base StatsExporter without overriding is going to get the "iterate the registered list" behavior, which is unintuitive.  I'll fix it so these methods are actually optional.

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -40,9 +41,15 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
   const uint64_t functionCost_{1};
 
 public:
-  CPUDeviceManager(const DeviceConfig &config)
+  explicit CPUDeviceManager(const DeviceConfig &config)
       : QueueBackedDeviceManager(config),
-        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {}
+        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {
+    Stats.incrementCounter("accelerators_used.cpu", 1);
```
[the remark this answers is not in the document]
bertmaher: How about `glow.devices_used.xyz`?  I picked `accelerators` so that the string is a bit more descriptive, but the `glow.` prefix should at least help.

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -40,9 +41,15 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
...
+  }
+
+  ~CPUDeviceManager() override {
+    Stats.incrementCounter("accelerators_used.cpu", -1);
```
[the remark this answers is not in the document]
rdzhabarov: +1, maybe just add().

lib/Backends/CPU/CPUDeviceManager.h
rdzhabarov: how do we test this? (counters working properly in places how they supposed to work).

bertmaher: I'll extend the `StatsExporterTest`.  It's really easy since the global exporter supports registration, you can just register a mock exporter, create a device manager, and then check the stat value.

lib/Runtime/StatsExporter.cpp
bertmaher: Good catch, I only thought about SIOF and not SDOF :-p

include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef GLOW_RUNTIME_STATSEXPORTER_H
+#define GLOW_RUNTIME_STATSEXPORTER_H
+
+#include "llvm/ADT/StringRef.h"
+
+namespace glow {
+
+/// Interface for exporting runtime statistics.  The base implementation
+/// delegates to any subclass registered via `registerStatsExporter`.
+class StatsExporter {
```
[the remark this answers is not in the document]
nickgg: Yeah this matches internal terminology. This stub doesn't do any actual exporting (subclasses will) but I think thats the interface still, so I like it.

tests/unittests/StatsExporterTest.cpp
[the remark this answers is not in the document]
nickgg: Previously I've used codegen here to wrap these calls with static functions that can't get the key name wrong (ie moving the flexibility to compile time). That worked very well but it's quite a bit of extra code to write.

tests/unittests/StatsExporterTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  void setCounter(llvm::StringRef key, int64_t value) override {
+    counters[key] = value;
```
rdzhabarov: nit: Making the actual method MOCK_METHOD (but not implementation) gives us better granularity of checking. Like we may do EXPECT_CALL on the method with certain params and actually verify sequence of calls instead of just checking final value of a counter.

lib/Backends/CPU/CPUDeviceManager.h
rdzhabarov: Can we verify those counters in a corresponding device manager test?

tests/unittests/StatsExporterTest.cpp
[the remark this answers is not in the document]
jackm321: Say we want to export a memory usage counter per-device (for example), this wouldn't be possible with static strings or codegen unless the number of devices will be available at compile time.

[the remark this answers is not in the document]
bertmaher: Kind of on the fence here.  Since fb303 uses string keys I'm inclined to leave the interface as is.  I could see making `glow.devices_used.interpreter` etc into constants rather than literals though.

include/glow/Runtime/StatsExporter.h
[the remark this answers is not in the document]
bertmaher: Yeah on the naming (this and `incrementCounter`) I'm pretty much following fb303 terminology.  Obviously that's not strictly necessary, I just figured it would be more obvious to FB internal users.

include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,68 @@
...
+  /// Dtor.
+  ~StatsExporterRegistry() {}
+
+  /// Add value to a time series for all registered StatsExporters.
+  void addTimeSeriesValue(llvm::StringRef key, double value);
+
+  /// Set a counter for all registered StatsExporters.
+  void incrementCounter(llvm::StringRef key, int64_t value = 1);
+
+  /// Increment a counter for all registered StatsExporters.
```
rdzhabarov: comment seems incorrect

include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,68 @@
...
+/// Registry of StatsExporters.
+class StatsExporterRegistry final {
+public:
+  /// Dtor.
+  ~StatsExporterRegistry() {}
+
+  /// Add value to a time series for all registered StatsExporters.
+  void addTimeSeriesValue(llvm::StringRef key, double value);
+
+  /// Set a counter for all registered StatsExporters.
```
rdzhabarov: swap this and next comment.

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -40,9 +41,15 @@ class CPUDeviceManager : public QueueBackedDeviceManager {
   const uint64_t functionCost_{1};
 
 public:
-  CPUDeviceManager(const DeviceConfig &config)
+  explicit CPUDeviceManager(const DeviceConfig &config)
       : QueueBackedDeviceManager(config),
-        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {}
+        maxMemoryBytes_(config_.getDeviceMemory(2000000000)) {
+    Stats()->incrementCounter("glow.devices_used.cpu", 1);
```
rdzhabarov: nit: can you kill all '1' where it's not needed.

include/glow/Runtime/StatsExporter.h
bertmaher: See if y'all didn't make me write doxygen, I couldn't mess it up, lol

jackm321: https://i.imgflip.com/1iw40c.jpg

include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,70 @@
...
+  void addTimeSeriesValue(llvm::StringRef key, double value);
+
+  /// Increment a counter for all registered StatsExporters.
+  void incrementCounter(llvm::StringRef key, int64_t value = 1);
+
+  /// Set a counter for all registered StatsExporters.
+  void setCounter(llvm::StringRef key, int64_t value);
+
+  /// Register a StatsExporter.
+  void registerStatsExporter(StatsExporter *exporter);
```
rdzhabarov: once registered who is going to own StatsExporter?

tests/unittests/StatsExporterTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+  EXPECT_EQ(ts[0], 1);
+  EXPECT_EQ(ts[1], 3.14);
+  EXPECT_EQ(ts[2], 2.71);
+}
+
+TEST(StatsExporter, Device) {
+  using namespace glow::runtime;
+  std::unique_ptr<DeviceManager> DM(
+      DeviceManager::createDeviceManager(DeviceConfig("Interpreter")));
+  EXPECT_EQ(MockStats.counters["glow.devices_used.interpreter"], 1);
```
rdzhabarov: can you verify a stat corresponding to the dtor of device manager.

we'll have stats recorded in various part of the glow runtime, I think it's more scalable/clear to have these tests in corresponding test files like DeviceManagerTest for this specific one.

include/glow/Runtime/StatsExporter.h
bertmaher: Stats reporters should pretty much always be singletons. They’re owned by the process.

include/glow/Runtime/StatsExporter.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Set a counter.  May be called concurrently.
+  virtual void setCounter(llvm::StringRef key, int64_t value) = 0;
+};
+
+/// Registry of StatsExporters.
+class StatsExporterRegistry final {
+public:
+  /// Dtor.
+  ~StatsExporterRegistry() {}
```
jackm321: why does this need to be declared?

bertmaher: Oh.  It doesn't.  I'd originally derived this from StatsExporter, but after I did that I realized it was kind of useless.

bertmaher:
Oh, @jackm321 to come back your question about ExportType, I was kind of on the fence.  The fb303 interface (esp for quantile stats) has a fair number of options and I was a little bit wary of adding too much complexity to this interface up front before we figure out what's useful.  I'm totally open to argument on this, but maybe we can iterate after this initial API lands?

tests/unittests/StatsExporterTest.cpp
bertmaher: Added the dtor test (and a pre-construction test).  I figure we can add some more detailed stats in other places later.  I kind of wouldn't mind some e2e tests that mimic an actual use case and sanity check counters along the way.

PR #3331, opened 2019-07-30
Porting lenet-mnist training example to EE2

mortzur:
Summary:
Ported lenet-mnist training example to use EE2.

Test Plan:
output reproduced for the 2 test modes:
* Results: guessed/total:74/80 (model created programatically)
* Results: guessed/total:76/80 (model loaded from a file)

### DIFF
--- a/examples/mnist.cpp
+++ b/examples/mnist.cpp
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 #include "glow/Base/Image.h"
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Importer/Caffe2ModelLoader.h"
 #include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
@@ -82,7 +82,7 @@ unsigned loadMNIST(Tensor &imageInputs, Tensor &labelInputs) {
   return numImages;
 }
 
-void createModel(ExecutionEngine &EE, Function *F,
+void createModel(ExecutionEngine2 &EE, Function *F,
                  PlaceholderBindings &bindings, unsigned minibatchSize,
                  Placeholder *&inputPH, Placeholder *&outputPH,
                  Placeholder *&selectedPH) {
@@ -107,8 +107,8 @@ void createModel(ExecutionEngine &EE, Function *F,
   outputPH = result->getPlaceholder();
 }
 
-void trainModel(ExecutionEngine &EE, PlaceholderBindings &bindings, Function *F,
-                unsigned minibatchSize, unsigned numIterations,
+void trainModel(ExecutionEngine2 &EE, PlaceholderBindings &bindings,
+                Function *F, unsigned minibatchSize, unsigned numIterations,
                 Tensor &imageInputs, Tensor &labelInputs, Placeholder *inputPH,
                 Placeholder *selectedPH) {
   llvm::Timer timer("Training", "Training");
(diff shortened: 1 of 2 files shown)


examples/mnist.cpp
```diff
@@ -139,36 +141,39 @@ void trainModel(ExecutionEngine &EE, PlaceholderBindings &bindings, Function *F,
...
                    Function *F, unsigned minibatchSize, unsigned numIterations,
                    Tensor &imageInputs, Tensor &labelInputs,
                    Placeholder *inputPH, Placeholder *outputPH,
                    bool transpose) {
   LOG(INFO) << "Validating.";
 
   ::glow::convertPlaceholdersToConstants(F, bindings, {inputPH, outputPH});
-  EE.compile(CompilationMode::Infer, F);
+  EE.compile(CompilationMode::Infer);
+  std::cerr << "compiled infer" << std::endl;
```
gcatron: Do we need this debugging statement still?

examples/mnist.cpp
```diff
@@ -258,25 +272,50 @@ void testMNISTLoadAndTraining() {
...
-  if ((errPtr = glow::prepareFunctionForTraining(F, bindings, selected))) {
+  if ((errPtr =
+           glow::prepareFunctionForTraining(TF, trainingBindings, selected))) {
     return;
   }
 
   const int numIterations = 30;
+  // Get input placeholder.
+  auto *A = llvm::cast<glow::Placeholder>(
+      EXIT_ON_ERR(trainingLoader.getNodeValueByName(inputName)));
```
putivsky: better to request input placeholders and trainingLoader .getInputVarsMapping(), and obtain particular placeholder by name.

examples/mnist.cpp
```diff
@@ -139,36 +141,39 @@ void trainModel(ExecutionEngine &EE, PlaceholderBindings &bindings, Function *F,
...
 
   for (int iter = numIterations; iter < numIterations + 10; iter++) {
     bindings.get(inputPH)->copyConsecutiveSlices(&imageInputs,
                                                  minibatchSize * iter);
-    EE.run(bindings);
+    EE.run(bindings, fname);
 
     for (unsigned i = 0; i < minibatchSize; i++) {
       auto T = bindings.get(outputPH)->getHandle().extractSlice(i);
+
```
gcatron: Remove extra line.

examples/mnist.cpp
```diff
@@ -258,25 +272,50 @@ void testMNISTLoadAndTraining() {
...
+      EXIT_ON_ERR(trainingLoader.getNodeValueByName(inputName)));
+
+  trainModel(EET_, trainingBindings, TF, minibatchSize, numIterations,
+             imageInputs, labelInputs, A, selected);
 
-  trainModel(EE, bindings, F, minibatchSize, numIterations, imageInputs,
-             labelInputs, A, selected);
+  // Get input and output placeholders.
+  A = llvm::cast<glow::Placeholder>(
+      EXIT_ON_ERR(loader.getNodeValueByName(inputName)));
```
putivsky: Same here.

examples/mnist.cpp
mortzur: Do you mean to avoid the casting and EXIT_ON_ERR and use the mapping and CHECK instead ? (like the code below?) 

llvm::StringRef inputName = "input";
auto inputs = loader.getInputVarsMapping();
CHECK(inputs.size()>0 << "msg")
Placeholder *p = inputs[inputName];
CHECK(p << "msg")

putivsky: auto inputs = loader.getInputVarsMapping();
auto it = inputs.find(inputName);
// EXIT_ON_ERR (it == inputs.end()); - pseudo code.  
Placeholder *p = it->second;

mortzur: I agree with the point you made here.
Interestingly, the Caffe2ModelLoader constructor does not load the inputs into this map (another CTOR of this class does , but not this one) - it loads it implicitly via loadConv.
This is really weird and worth looking at, but outside the scope of this porting.
I'm leaving the existing code as is.

PR #3335, opened 2019-07-31
Ported char-rnn to EE2

gcatron:
Summary: Ported char-rnn to EE2

Documentation:

Progress on #3239 

Test Plan: Build and run char-rnn and check output.

Example run.. 
Screen Shot 2019-07-30 at 5 28 23 PM

### DIFF
--- a/examples/char-rnn.cpp
+++ b/examples/char-rnn.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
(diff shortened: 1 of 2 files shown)


examples/char-rnn.cpp
```diff
@@ -210,7 +210,7 @@ int main(int argc, char **argv) {
   auto mb = loadFile(inputFilename);
   auto text = mb.get()->getBuffer();
   LOG(INFO) << "Loaded " << text.size() << " chars.\n";
-  PlaceholderBindings bindings;
+  PlaceholderBindings bindings, inferBindings, trainingBindings;
```
mortzur: "bindings" seems to be not in use now. remove if that's the case.

mortzur:
LGTM!

PR #3336, opened 2019-07-31
Add string->string map backendSpecificOpts to BackendOptions

jfix71:
Summary: Allow for specifying options specific to a backend via a string to string map in the BackendOptions.

Test Plan: Added test for deserializing the string string map for the backend opts yaml.


gcatron:
Great work!
I agree with replacing hints entirely, I think the map covers what's needed, If we go that route it would also make sense to just but the map in options and get rid of hints entirely. 

I'm good with HostManager setting hints, @beicy do you have an opinion?

jfix71:
So thinking more on this, backend-specific options are set pre-partitioning. So it seems to me they're always per DAG. I.e. they're more similar to `collectConstants` and `autoInstrument` than to the current BackendHints, which is intended to be set by the partitioner per DAGNode.

So perhaps this backend-specific options string-string map should live in the BackendOptions, and the BackendHints should just stay as is? I also don't think we need the flexibility of the string-string map for BackendHints since they're set from the partitioner. So if we need a new field we can just add it to the struct. The only reason to use the string-string map is for flexibility across different backends with their own specific compilation flags. WDYT?

gcatron:
That seems fair, this map is really for backend level flags for configuration not really like hints which are partition specific?

jfix71:
Yeah exactly. The backend isn't even aware of the way the partitioner is operating. These are flags specified when loading the model.

gcatron:
Putting the map in options and leaving hints as is sounds good to me.

include/glow/Backends/BackendOptions.h
```diff
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_BACKENDOPTIONS_H
 
 #include "llvm/ADT/SmallVector.h"
+#include <map>
```
jackm321: any reason to not use unordered_map or llvm::StringMap?

jfix71:
@gcatron Updated to `backendSpecificOptions` in `BackendOptions`.

include/glow/Backends/BackendOptions.h
jfix71: `std::map` is what's used by `LLVM_YAML_IS_STRING_MAP`, and didn't think it'd make sense to convert from what's deserialized into `std::map` into one of those others.

PR #3337, opened 2019-07-31
[trivial] Fix OperatorTests::insertTensorTest

nickgg:
Summary: In OperatorTests::insertTensorTests the insertTensor is getting optimized out because all inputs are constant splats. Changing them to Placeholders not known at compile time fixes this.

Documentation: n/a

Test Plan: unit tests, added printfs to libjit_insert_tensor to verify that the InsertTensorInst is in the final product.

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -6986,18 +6986,17 @@ TEST_P(OperatorTest, SigmoidCrossEntropyWithLogits) {
 TEST_P(OperatorTest, insertTensorTest) {
   ENABLED_BACKENDS(Interpreter, CPU);
 
-  auto SN0Ty = mod_.uniqueType(ElemKind::Int64ITy, {4, 6});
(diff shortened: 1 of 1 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -6986,18 +6986,17 @@ TEST_P(OperatorTest, SigmoidCrossEntropyWithLogits) {
 TEST_P(OperatorTest, insertTensorTest) {
   ENABLED_BACKENDS(Interpreter, CPU);
```
rdzhabarov: nit: seems like OpenCL also has InsertTensor instr support, could just add that as well.

nickgg: OpenCL has insertTensor but only for FP32 and Int8 - this test covers Int64. So it's right to be disabled.

The concatTests optimize down to InsertTensors so the OCL code path is covered by OperatorTests.

PR #3339, opened 2019-07-31
Fixing return statement.

putivsky:
Summary:
Fixing PT-Glow integration code. Missing return statements and a couple incorrect checks.
Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -45,7 +45,8 @@ inline llvm::Expected<float> to32Bit(double val) {
 inline llvm::Expected<int32_t> to32Bit(int64_t val) {
   RETURN_ERR_IF_NOT(val <= std::numeric_limits<int32_t>::max() ||
                         val >= std::numeric_limits<int32_t>::lowest(),
-                    glow::strFormat("Value %ld is out of limit.", val));
+                    glow::strFormat("Value %lld is out of limit.",
+                                    static_cast<long long int>(val)));
   return llvm::Expected<int32_t>(static_cast<int32_t>(val));
 }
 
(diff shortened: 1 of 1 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -45,7 +45,7 @@ inline llvm::Expected<float> to32Bit(double val) {
 inline llvm::Expected<int32_t> to32Bit(int64_t val) {
   RETURN_ERR_IF_NOT(val <= std::numeric_limits<int32_t>::max() ||
                         val >= std::numeric_limits<int32_t>::lowest(),
-                    glow::strFormat("Value %ld is out of limit.", val));
+                    glow::strFormat("Value %lld is out of limit.", val));
```
jackm321: Doesn't like this, int64_t is a long I guess

jackm321: I don't know much about this, but interestingly int64_t seems to be a long on CI and long long on macos. Maybe we should just static cast this to long long for this print?

putivsky: Here is the problem:
cmake as a part of python build requires %lld:
<<< 
PyTorchModelLoader.cpp:48:67: warning: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
                    glow::strFormat("Value %ld is out of limit.", val));
                                           ~~~                    ^~~
                                           %lld
>>>

Whereas GitHub build requires %ld, I will make change for GitHub compliance.

jackm321: If we cast it then we should be able to make both happy

PR #3343, opened 2019-07-31
[QuantizationTest] Initialize filter and bias to prevent flakiness

jfix71:
Another flaky test fix.


jfix71:
Yeah I guess so... Maybe when combined with ASAN it was unlikely to get a chunk of memory with NANs in it, since ASAN uses a special allocator? I never saw any of these issues when running with sanitizers, it always came from that one assert from Constant dedup that fails due to NANs.

bertmaher:
Interesting, there’s a MemorySanitizer that will catch read from uninitialized memory: https://clang.llvm.org/docs/MemorySanitizer.html

PR #3344, opened 2019-07-31
ONNX Less implementation as GLOW CmpLT node.

ayermolo:
Summary: Implementation of Less ONNX node as CmpLT.

Documentation:

[Optional Fixes #issue]
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Less
Test Plan: Unit Tests.


tools/ClassGen/InstrGen.cpp
```diff
@@ -427,6 +427,16 @@ int main(int argc, char **argv) {
       .autoVerify(VerifyKind::SameElementType, {"Dest", "ElemKind::BoolTy"})
       .autoIRGen("CmpEQ");
 
+  BB.newInstr("ElementCmpLT")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("LHS", OperandKind::In)
+      .addOperand("RHS", OperandKind::In)
+      .dataParallel()
+      .autoVerify(VerifyKind::SameShape, {"Dest", "LHS", "RHS"})
+      .autoVerify(VerifyKind::SameElementType, {"LHS", "RHS"})
```
rdzhabarov: SameType here?

lib/Graph/Nodes.cpp
```diff
@@ -1487,6 +1487,14 @@ bool ResizeNearestNode::verify() const {
   return isValid;
 }
 
+bool CmpLTNode::verify() const {
```
jfix71: Can we just do `VERIFY_CMP(CmpLT)`?

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -2268,3 +2268,33 @@ TEST(onnx, importWhere) {
...
+                           {&X.getType(), &Y.getType()}, *F);
+    out = EXIT_ON_ERR(onnxLD.getOutputByName("Out"));
+  }
+
+  auto *save = getSaveNodeFromDest(out);
+
+  CmpLTNode *CMPLT = llvm::dyn_cast<CmpLTNode>(save->getInput().getNode());
+
+  ASSERT_TRUE(CMPLT);
+  EXPECT_EQ(CMPLT->getResult().dims()[0], 4);
```
jfix71: nit: `ASSERT_EQ(CMPLT->getResult().dims().size(), 3);` before here to prevent potential segfault for this and below.

tests/unittests/QuantizationTest.cpp
```diff
@@ -879,6 +879,7 @@ TEST_P(Operator, end2endInt8) {
   if (backendSpecificEE.getBackendName() == "OpenCL") {
     keepOriginalPrecisionForNodes.insert(Kinded::Kind::SelectNodeKind);
     keepOriginalPrecisionForNodes.insert(Kinded::Kind::CmpLTENodeKind);
+    keepOriginalPrecisionForNodes.insert(Kinded::Kind::CmpLTNodeKind);
```
jfix71: Why is this necessary? You didn't add OCL support at all, or add CMPLT to testQuantizationEnd2End.

tests/unittests/OperatorTest.cpp
```diff
@@ -69,6 +69,262 @@ static Tensor createTensorConditionallyQuantized(ElemKind T,
...
+  EE.compile(CompilationMode::Infer);
+  EE.run(bindings);
+
+  return saveAlloc->getHandle<bool>();
+}
+
+TEST_P(OperatorTest, less_int8) {
+  ENABLED_BACKENDS(Interpreter, CPU);
+
+  llvm::SmallVector<int8_t, 16> xValues = {
```
jfix71: Can we just use normal arrays here/in all the tests? I don't think we need a vector as we know the size statically.

tests/unittests/OperatorTest.cpp
```diff
@@ -69,6 +69,262 @@ static Tensor createTensorConditionallyQuantized(ElemKind T,
...
+      for (size_t k = 0; k < saveH.dims()[2]; ++k) {
+        for (size_t f = 0; f < saveH.dims()[3]; ++f) {
+          EXPECT_FLOAT_EQ(refResults[counter++], saveH.at({i, j, k, f}));
+        }
+      }
+    }
+  }
+}
+
+TEST_P(OperatorTest, less_floatCases) {
```
jfix71: What signal does this test provide that `less_float` doesn't?

tests/unittests/OperatorTest.cpp
```diff
@@ -69,6 +69,262 @@ static Tensor createTensorConditionallyQuantized(ElemKind T,
...
+      for (size_t k = 0; k < saveH.dims()[2]; ++k) {
+        for (size_t f = 0; f < saveH.dims()[3]; ++f) {
+          EXPECT_FLOAT_EQ(refResults[counter++], saveH.at({i, j, k, f}));
+        }
+      }
+    }
+  }
+}
+
+TEST_P(OperatorTest, less_float) {
```
jfix71: This and `less_int8` seem to be very similar tests. Can we combine them similar to how we do with e.g. `testBatchedReduceZeroDimResult` where we reuse the same data? Then we could also test `Int32ITy` and `Float16Ty` too without a lot of extra code.

include/glow/Graph/Graph.h
```diff
@@ -729,6 +730,17 @@ class Function final : public Named {
                         inputs[2]);
   }
 
+  /// Template function that creates a node and normalizes its input shapes
+  /// with the use of BroadCast nodes. If axis is -1, it calculates it
+  /// automatically for multi directional broadcast.
+  template <class T, class... Args>
+  typename enable_if_same_t<T, CmpLTNode>::type *
+  createNodeWithBroadcast(const std::string &name, int axis,
```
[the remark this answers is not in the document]
ayermolo: It doesn't quite work.  Macro assumes the output type is same as input type, and there is API that takes in type. For cmpLT the output type has ElemKind bool.

tests/unittests/QuantizationTest.cpp
ayermolo: I figured in case someone adds it later, but in retrospect they can modify this code also. Will remove.

tests/unittests/OperatorTest.cpp
ayermolo: Tests all three cases for cmp explicitly on a small dataset, without broadcast. I can remove.

tests/unittests/OperatorTest.cpp
ayermolo: This way there is some code duplication on data creation side, but it allows us to re-use the same function to test different shapes. I guess you can just have one data set and then come up with shapes that will map to it, but I think it will be limiting.

include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -998,6 +998,39 @@ class CommonOperatorLoader : public ProtobufLoader {
...
+    std::string opName = loadOperatorName(op);
+
+    auto *xNode = xNV.getNode();
+    auto *yNode = yNV.getNode();
+    auto xDims = xNV.dims();
+    auto yDims = yNV.dims();
+
+    std::vector<size_t> targetDim;
+    ASSIGN_VALUE_OR_RETURN_ERR(targetDim,
+                               computeMultidirectionalBroadcast(xDims, yDims));
```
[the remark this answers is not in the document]
ayermolo: ah yes. I think something got messed up during one of our rebases. Sorry didn't notice.

ayermolo:
@jackm321 what do you think, good to go, or you want me to make some other changes?

jfix71:
Sorry for the delay @ayermolo !

Mostly small nits. Just one main question on if we can reuse the logic for CmpLTE in the interpreter kernels. 

Also, I think there's a missing test for Float16 for the Interpreter, can you add just a basic one like the ones you have for "cases"? Thanks.

ayermolo:
@jfix71 No problem. Refactored InterpreterNodes so that cmpLT, cmpLTE, cmpEQ use the same helper function, and added fp16 for cmpLT.

tests/unittests/OperatorTest.cpp
```diff
@@ -69,6 +69,254 @@ static Tensor createTensorConditionallyQuantized(ElemKind T,
...
 }
 
+template <typename DataType>
+glow::Handle<bool>
+lessHelper(glow::PlaceholderBindings &bindings, glow::Module &mod,
+           glow::Function *F, glow::ExecutionEngine &EE, ElemKind DTy,
+           llvm::ArrayRef<DataType> xValues, llvm::ArrayRef<DataType> yValues,
+           llvm::ArrayRef<size_t> xDims, llvm::ArrayRef<size_t> yDims) {
+  auto *X = createPlaceholderConditionallyQuantized(mod, DTy, xDims, "X",
+                                                    DTy != ElemKind::FloatTy);
```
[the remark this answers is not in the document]
ayermolo: You are right, should be false.

ayermolo:
Awesome, thanks!

PR #3351, opened 2019-08-01
Ported fr2en to EE2

gcatron:
Summary: This PR ports fr2en to EE2.

Documentation:

Progress on #3239 

Test Plan: build, Run fr2en verify that it correctly translates the test phrases.

### DIFF
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "glow/ExecutionEngine/ExecutionEngine.h"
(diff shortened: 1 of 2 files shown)


examples/fr2en.cpp
```diff
@@ -164,7 +164,8 @@ struct Model {
       precConfig.quantConfig.assertAllNodesQuantized = true;
     }
 
-    EE_.compile(F_, cctx);
+    EE_.compile(cctx);
+    bindings.allocate(EE_.getModule().getPlaceholders());
```
jfix71: Huh, this wasn't needed before? Wondering why it's needed now with EE2.

gcatron: Good catch, it's only only needed when training, removed.

PR #3355, opened 2019-08-01
Test temporary files must have unique names, cause some testing frame…

putivsky:
…works execute the same tests in multile threads.

Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:


jackm321:
There's also std::tmpfile

jfix71:
There's also `llvm::sys::fs::createTemporaryFile()`, which we already use in other places, e.g. see the `writePngImage` test from `ImageTest.cpp`.

PR #3356, opened 2019-08-01
Add FP16 accumulation option to (Fused)-RWQ-SLWS/SLS

jfix71:
Summary: As above.

Test Plan: Added new FP16 accumulation versions of tests for the already FP16 versions that exist.


jfix71:
>Update RecSys with fp16 accumulation support

Yep this was next on my TODO list 🙂 

>I was thinking whether accumulation could be an enum, target accumulation type, but have not found a use cases other than fp16 and fp32. Let me know if you are aware of something where it would be useful.

I actually did this but then tore it back out haha -- I added a new member type `ElemKind` and then used it instead of UseFP16Accumulation. I tore it out because I couldn't think of use cases for now and figured we could add it if we decided in the future. Could always bring it back now though...

Issue #3328, opened 2019-07-30
deduplicateConstants assert will always fire if Constant contains NaN


geoffberry:
The assert referenced below will always assert for Constants that contain a NaN since such constants will always return false from isEqual.  Furthermore, Constants that contain NaN will not be de-duplicated for this same reason.  Perhaps it would be better to avoid using isEqual in ConstsEqDedup and instead do a bit-wise comparison.

This assert fires intermittently in the test GraphOptz.ReshapeConstantOneUse depending on whether the uninitialized input tensor contains a NaN.

https://github.com/pytorch/glow/blob/25ce2de7248d7be1e8e43ded9b36d5e3ce0e7cb2/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp#L1818

jfix71:
Thanks for reporting this issue @geoffberry! Are you interested in putting up a fix for this?

bertmaher:
Nice catch!  I kind of think that assert is problematic; iirc a vanilla std::map<float> will fail to find NAN keys too.  Idk if it’s worth adding special logic to dedup constants with NAN but I’m kind of ok either way.

rosierm:
I'm going to start working on a patch shortly.

chadrosier:
Fixed here: https://github.com/pytorch/glow/commit/f0c08f92477e64c4fb03bb80281e6afb2ba6ffd2

jfix71:
Thanks @chadrosier!

PR #3361, opened 2019-08-01
Enable deduplication of constants with NaNs. (#3328)

chadrosier:
Summary:
Constants that contain NaN will not be de-duplicated because isEqual always returns false for this case.  Instead, perform a bitwise comparison to achieve the expected behavior.

### DIFF
--- a/examples/training/resnet50/CMakeLists.txt
+++ b/examples/training/resnet50/CMakeLists.txt
@@ -2,7 +2,7 @@ add_executable(resnet-training
                  main.cpp)
 target_link_libraries(resnet-training
                       PRIVATE
-                        ExecutionEngine
+                        ExecutionEngine2
                         Graph
                         Importer
                         GraphOptimizer)
--- a/examples/training/resnet50/main.cpp
+++ b/examples/training/resnet50/main.cpp
@@ -15,7 +15,7 @@
  */
 
 #include "glow/Base/Image.h"
-#include "glow/ExecutionEngine/ExecutionEngine.h"
+#include "glow/ExecutionEngine/ExecutionEngine2.h"
 #include "glow/Graph/Nodes.h"
 #include "glow/Importer/Caffe2ModelLoader.h"
 #include "glow/Optimizer/GraphOptimizer/TrainingPreparation.h"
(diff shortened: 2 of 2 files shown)


tests/unittests/GraphOptzTest.cpp
```diff
@@ -2371,6 +2371,58 @@ TEST_F(GraphOptz, VarsCSE) {
...
 }
 
+TEST_F(GraphOptz, VarsCSENaN) {
+  // Create two variables that are Private, are not trainable, have no writers
+  // and include NaNs. The first two variables have the same data, and so should
+  // be combined via variable CSE.  In particular, the NaN constants should not
+  // prevent the variables from being combine.
+  auto *input1 = mod_.createConstant(ElemKind::FloatTy, {5}, "input1");
+  auto *input2 = mod_.createConstant(ElemKind::FloatTy, {5}, "input2");
+  auto nan = std::numeric_limits<float>::quiet_NaN();
```
rdzhabarov: interesting, is it any different from using NAN value?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2371,6 +2371,58 @@ TEST_F(GraphOptz, VarsCSE) {
...
+  input1->getHandle() = {nan, nan, nan, nan, nan};
+  input2->getHandle() = {nan, nan, nan, nan, nan};
```
rdzhabarov: I'd add some variety NAN + non NAN values in the constants to make sure it's handled properly.

rdzhabarov:
@chadrosier make sure to fix the test failures

include/glow/Base/Tensor.h
```diff
@@ -629,6 +643,16 @@ class Tensor final {
     }
     return numExceedingError == 0;
   }
+
+  bool isBitwiseEqualImpl(const Tensor &other) const {
+    auto const *myData = getRawDataPointer<uint8_t>();
```
rdzhabarov: I think you need: `getUnsafePtr`

tests/unittests/GraphOptzTest.cpp
chadrosier: This was an arbitrary choice.  I'll update the pull request to use the NAN value.

include/glow/Base/Tensor.h
chadrosier: Correct.  Thanks, Roman.  I missed this because I was testing on a release build (without asserts).

chadrosier:
Unfortunately, I'm not used to this development flow and my attempt to rebase my fork/branch to ToT went *way* off course.

Updated pull request here: https://github.com/pytorch/glow/pull/3376
This should address all of your comments, Roman.

PR #3363, opened 2019-08-01
Ported resnet-verify to EE2

gcatron:
Summary: Ported resnet-verify to EE2

Documentation:

Progress on #3239 

Test Plan: Verify everything builds, Run resnet-verify and ensure it completes successfully. I don't think this is run as part of CI. Tested locally on my machine.


nickgg:
Whats the purpose of EER, appears you only use EEI?

gcatron:
> Whats the purpose of EER, appears you only use EEI?

Good question, It was being used as Module storage, a unique_ptr can do the same. EER is removed.

PR #3206, opened 2019-07-03
[Operator] Add Support for ReduceMin Operator

ksaurabh-cadence:
Summary:
CPUBackend supports: Flt, Int32, Int64
Interpreter supports: Flt, FP16, Int32, Int64
Default Axis Support for ReduceOps

Documentation:
[Optional Fixes #issue]

Test Plan:
Added ReduceMin Operator Test
Added ReduceMin ONNXLoader Test

Please see a detailed explanation of how to fill out the fields in the relevant sections in PULL_REQUEST.md.

PR #3366, opened 2019-08-01
Adding support for Flatten Glow operator to PT/Glow.

putivsky:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/lib/Backends/CPU/libjit/libjit.cpp
+++ b/lib/Backends/CPU/libjit/libjit.cpp
@@ -16,6 +16,7 @@
 #include <algorithm>
 #include <assert.h>
 #include <chrono>
+#include <cmath>
 #include <math.h>
 #include <numeric>
 #include <stddef.h>
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -249,12 +249,20 @@ struct SoftMaxInputs {
     dtype = 2,
   };
 };
+
+// Indexes of aten::flatten inputs.
+struct FlattenInputs {
+  enum {
+    input = 0,
+    start_dim = 1,
+    end_dim = 2,
+  };
+};
 } // namespace
 
 // static
 const PyTorchModelLoader::MappingOfMemberFunctions &
 PyTorchModelLoader::getSymbolsMapping() {
-
   /// Static map of the set of PyTorch symbols to load, the PyTorchModelLoader
   /// for loading these symbols, and the set of inputs that should be considered
   /// immutable between inference invocations by Glow and loaded as Constants
@@ -330,6 +338,12 @@ PyTorchModelLoader::getSymbolsMapping() {
             AvgPoolInputs::divisor_override,
         }},
        {{"aten::mm"}, &PyTorchModelLoader::loadMatMul, {}},
+       {{"aten::flatten"},
+        &PyTorchModelLoader::loadFlatten,
+        {
+            FlattenInputs::start_dim,
+            FlattenInputs::end_dim,
+        }},
        {{"aten::prelu"},
         &PyTorchModelLoader::loadPRelu,
         {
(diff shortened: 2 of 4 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -734,6 +736,41 @@ llvm::Error PyTorchModelLoader::loadMin(const torch::jit::Node *ptNode) {
...
 
+llvm::Error PyTorchModelLoader::loadFlatten(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, 3, outputs, 1));
+
+  // Indexes of aten::flatten inputs.
+  struct Inputs {
+    enum {
+      input = 0, // NCHW
```
jackm321: don't need NCHW

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -734,6 +736,41 @@ llvm::Error PyTorchModelLoader::loadMin(const torch::jit::Node *ptNode) {
...
+  };
+
+  glow::NodeValue in;
+  ASSIGN_VALUE_OR_RETURN_ERR(in, getGlowNodeValue(inputs[Inputs::input]));
+
+  auto startHandle = glow::Handle<int32_t>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      startHandle, getGlowConstantHandle<int32_t>(inputs[Inputs::start_dim]));
+  uint32_t startDim;
+  RETURN_IF_ERR(contractParamIfNeeded(startHandle, startDim));
```
jackm321: why is `contractParamIfNeeded` need here? Doesn't seem like it should be

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -734,6 +736,41 @@ llvm::Error PyTorchModelLoader::loadMin(const torch::jit::Node *ptNode) {
...
+
+  auto endHandle = glow::Handle<int32_t>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      endHandle, getGlowConstantHandle<int32_t>(inputs[Inputs::end_dim]));
+  int32_t endDim;
+  RETURN_IF_ERR(contractParamIfNeeded(endHandle, endDim));
+  RETURN_ERR_IF_NOT(endDim == -1, "only -1 value for end_dim got supported.");
```
jackm321: it would be pretty trivial to support this as well, I think we should just do it, up to you though

putivsky: It's not clear for me now, how to do it right, leaving as TODO.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -242,12 +242,20 @@ struct PReluInputs {
     weight = 1,
   };
 };
+
+// Indexes of aten::flatten inputs.
+struct FlattenInputs {
+  enum {
+    input = 0, // NCHW
```
jackm321: still true that we dont need NCHW because that's not necessarily true

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1045,6 +1059,34 @@ llvm::Error PyTorchModelLoader::loadPRelu(const torch::jit::Node *ptNode) {
...
+      getGlowConstantHandle<int32_t>(inputs[FlattenInputs::start_dim]));
+  uint32_t startDim;
+  RETURN_IF_ERR(contractParamIfNeeded(startHandle, startDim));
+
+  auto endHandle = glow::Handle<int32_t>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(endHandle, getGlowConstantHandle<int32_t>(
+                                            inputs[FlattenInputs::end_dim]));
+  int32_t endDim;
+  RETURN_IF_ERR(contractParamIfNeeded(endHandle, endDim));
+  RETURN_ERR_IF_NOT(endDim == -1, "only -1 value for end_dim got supported.");
```
jackm321: probably "is supported" makes more sense

torch_glow/src/PyTorchModelLoader.cpp
jackm321: why is this marked as resolved, it's not resolved

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1045,6 +1059,34 @@ llvm::Error PyTorchModelLoader::loadPRelu(const torch::jit::Node *ptNode) {
...
+
+  glow::NodeValue in;
+  ASSIGN_VALUE_OR_RETURN_ERR(in,
+                             getGlowNodeValue(inputs[FlattenInputs::input]));
+
+  auto startHandle = glow::Handle<int32_t>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      startHandle,
+      getGlowConstantHandle<int32_t>(inputs[FlattenInputs::start_dim]));
+  uint32_t startDim;
```
jackm321: from the docs, this should default to 0 and endDim should default -1 https://pytorch.org/docs/stable/torch.html#torch.flatten

torch_glow/src/PyTorchModelLoader.cpp
putivsky: Yeah, sorry, multiple PR for the same file make things difficult to merge (I have to do it every time).

lib/Backends/CPU/libjit/libjit.cpp
```diff
@@ -16,6 +16,7 @@
 #include <algorithm>
 #include <assert.h>
 #include <chrono>
+#include <cmath>
```
jackm321: this isn't related right?

putivsky: yes, someone broke the build, just fix it at the same time.

putivsky: https://github.com/pytorch/glow/pull/3206

jackm321: Ah cool thanks for fixing it

PR #3367, opened 2019-08-01
Add Layout field to Conv and Pool nodes and remove OCL specific versions

nickgg:
Summary: For convolutions in glow we use NHWC layout, but it's more efficient to use NCHW on GPUS. To enable this the OCL backend adds some specific OCLConvolution, OCLMaxPool and OCLAvgPool nodes and transforms general Conv and Pool nodes into them by shuffling dimensions and adding TransposeNodes.

This PR adds a field `Layout` to ConvolutionNode, MaxPoolNode and AvgPoolNode which can be either NHWC or NCHW. The OCL backend still transforms these nodes but into the same node type with the layout set to NCHW (we still add transposes). This will allow us to reuse this logic in other backends with the same NCHW preference without needing multiple identical BackendConvolutionNodes (etc).

This **DOES NOT** add NCHW convolution support to any backends, or change any OCL kernels - it just removes OCL specific nodes in favour of more general ones.

This was a consensus decision, but worth thinking about whether or no the code is clearer after this change or not too.

Documentation: deleted the section in docs referencing these nodes.

Test Plan: ninja test in various modes (debug, release, asan). Ran resnet-runtime on OCL, ran image-classifier on OCL, tracing-compare across interp, cpu and ocl.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -341,14 +341,15 @@ public:
   /// \p group defines the number of groups the input and output channels should
   /// be divided into and convolved separately. \p dilation defines factor by
   /// which gap between 2 neighboring kernel elements is expanded along each
-  /// axis.
+  /// axis. \p layout defines the Tensor layout and must be either NHWC or NCHW.
 
-  ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
-                              NodeValue filter, NodeValue bias, TypeRef outTy,
-                              llvm::ArrayRef<unsigned_t> kernels,
-                              llvm::ArrayRef<unsigned_t> strides,
-                              llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
-                              unsigned_t dilation = 1);
+  ConvolutionNode *
+  createConv(llvm::StringRef name, NodeValue input, NodeValue filter,
+             NodeValue bias, TypeRef outTy, llvm::ArrayRef<unsigned_t> kernels,
+             llvm::ArrayRef<unsigned_t> strides,
+             llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
+             unsigned_t dilation = 1,
+             ConvolutionLayout layout = ConvolutionLayout::NHWC);
 
   /// Creates a ConvolutionNode with the given \p name which convolves the 4D
   /// \p input with \p filter and \bias. \p kernel defines the size of the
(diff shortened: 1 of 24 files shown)


include/glow/Graph/Graph.h
```diff
@@ -343,12 +343,13 @@ class Function final : public Named {
   /// which gap between 2 neighboring kernel elements is expanded along each
   /// axis.
 
-  ConvolutionNode *createConv(llvm::StringRef name, NodeValue input,
-                              NodeValue filter, NodeValue bias, TypeRef outTy,
-                              llvm::ArrayRef<unsigned_t> kernels,
-                              llvm::ArrayRef<unsigned_t> strides,
-                              llvm::ArrayRef<unsigned_t> pads, unsigned_t group,
-                              unsigned_t dilation = 1);
+  ConvolutionNode *
```
rdzhabarov: update doxygen with the new param everywhere.

include/glow/Graph/Nodes.h
```diff
@@ -203,6 +203,9 @@ inline ShapeHWD calculate3DConvPoolOutputDims(
 /// Modes of the padding operation.
 enum PaddingMode { CONSTANT = 0, REFLECT, EDGE };
 
+/// Convolution Layouts.
+enum ConvolutionLayout { NHWC = 0, NCHW };
```
rdzhabarov: since this is applicable to pools as well, i could see this as a TensorLayout.
Slight downside in that case is that TensorLayout might be too generic and overloaded term.

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -753,6 +755,7 @@ static void fwdMaxPool(Tensor *inW, Tensor *outW, Tensor *argmaxW,
 }
 
 void BoundInterpreterFunction::fwdMaxPoolInst(const MaxPoolInst *I) {
+  assert(I->getLayout() == NHWC && "Glow Interpreter supports only NHWC Pools");
```
rdzhabarov: same checks for gradient version?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1804,14 +1793,6 @@ bool OCLBackend::isOpSupported(const NodeInfo &NI) const {
                                                   {ConvolutionNode::BiasIdx}) &&
            (NI.getInElemTy(ConvolutionNode::BiasIdx) == ElemKind::Int32QTy);
 
-  case Kinded::Kind::OCLConvolutionNodeKind:
```
rdzhabarov: for convolution: we do not support NCHW with group > 1.

tools/ClassGen/NodeGen.cpp
```diff
@@ -125,23 +127,28 @@ int main(int argc, char **argv) {
...
       .addMember(MemberType::VectorUnsigned, "Pads")
+      .addMember(MemberType::Enum, "Layout")
       .addResultFromCtorArg()
       .addGradient()
-      .setDocstring("Performs an Average Pool operation on the Input given "
-                    "provided Kernels, Strides, and Pads.");
+      .setDocstring(
+          "Performs an Average Pool operation on the Input given "
+          "provided Kernels, Strides, and Pads. Supported layouts are defined "
+          "in the COnvolutionLayout enum: NHWC and NCHW.");
```
rdzhabarov: typo: COnvo../Convo...

tools/ClassGen/InstrGen.cpp
```diff
@@ -138,6 +140,7 @@ int main(int argc, char **argv) {
       .addMember(MemberType::VectorUnsigned, "Kernels")
       .addMember(MemberType::VectorUnsigned, "Strides")
       .addMember(MemberType::VectorUnsigned, "Pads")
+      .addMember(MemberType::Unsigned, "Layout")
```
rdzhabarov: why not enum as well?

nickgg: Enums are not supported for Instructions (enum is basically a typedef of Unsigned anyway in the Nodegen so no real difference).

lib/Backends/OpenCL/OpenCL.cpp
nickgg: True, but the node doesn't get Transformed if group is > 1. That's also a runtime param that can't be checked here in isOpSupported.

include/glow/Graph/Nodes.h
nickgg: I think TensorLayout is too general, but can't think of a better name :/

nickgg:
addressed comments

rdzhabarov:
@nickgg are you going to merge this?

PR #3368, opened 2019-08-02
[Tensor] Introduce Tensor alignments.

artemrakhov:
**Summary:**
In many architectures, data array is required (or recommended) to be aligned in memory. Meaning that total size of the array is multiple of X, and its starting memory address is a multiple of X. In case of multi-dimensional array (Tensor), each dimension should also be a multiple of X. It means that Tensor will have "holes" in it, i.e. useful data is padded by "garbage" values.

This PR introduces ability to create arbitrary-aligned Tensors, fixes various operations to work correctly.

**Documentation:**
- Class Type now has `strides_` member, which contains stride for each dimension.
- `Tensor::size()` still returns number of useful elements.
- `Tensor::actualSize()` is new method, returns number of allocated elements, including "holes" used for padding.
- `Tensor::size() <= Tensor::actualSize()` for every Tensor.
- We will not make any guarantees regarding actual value of pad-elements. They could be arbitrary numbers, all code should work correctly with any pads and not depend on their values.

**Test Plan:**
`ninja test`

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -16,6 +16,7 @@
 #ifndef GLOW_BASE_TENSOR_H
 #define GLOW_BASE_TENSOR_H
 
+#include <algorithm>
 #include <cassert>
 #include <vector>
 
@@ -188,15 +189,14 @@ public:
   /// \returns the shape of the tensor.
   llvm::ArrayRef<size_t> dims() const { return type_.dims(); }
 
-  /// \returns the number of elements in the tensor.
+  /// \returns the number of real meaningful elements in the tensor. Does not
+  /// take strides into account.
   size_t size() const { return type_.size(); }
 
   /// \returns the actual number of elements in the tensor taking striding into
   /// account. Since size() does not take striding into account, size() is
   /// always <= actualSize().
-  size_t actualSize() const {
-    return type_.getSizeInBytes() / type_.getElementSize();
-  }
+  size_t actualSize() const { return type_.actualSize(); }
 
   /// \returns the number of bytes required to store the tensor.
   uint64_t getSizeInBytes() const { return type_.getSizeInBytes(); }
@@ -287,10 +287,9 @@ public:
              "Number of dims of tensor must equal number of dims in offsets");
       // Find the index of the first element and use it to find the pointer to
       // the first element.
-      size_t index = 0, pi = 1;
-      for (int i = this->dims().size() - 1; i >= 0; i--) {
-        index += pi * offsets[i];
-        pi *= this->dims()[i];
+      size_t index = 0;
+      for (size_t i = 0; i < this->dims().size(); i++) {
+        index += type_.strides()[i] * offsets[i];
       }
       firstElemPtr = &firstElemPtr[index * type_.getElementSize()];
     }
@@ -682,6 +681,95 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
 void dumpImpl(const Tensor *T, unsigned maxNumElem);
 void dumpImpl(const Tensor *T);
 
+template <class ElemTy> class Handle;
+
+/// A class that provides ability to iterate over a Handle<ElemTy>. Since it's
+/// common to have both mutating and const iterators, this class has template
+/// parameter IsConst, which is true to create const_iterator and false
+/// otherwise.
+template <class ElemTy, bool IsConst>
+class HandleIterator
+    : public std::iterator<std::random_access_iterator_tag, ElemTy> {
+  using HandleTy = typename std::conditional_t<IsConst, const Handle<ElemTy> *,
+                                               Handle<ElemTy> *>;
+  using ElemTyRef =
+      typename std::conditional_t<IsConst, const ElemTy &, ElemTy &>;
+
+  /// At every given moment, the iterator maintains an index, which is used to
+  /// access the Handle. When moving the iterator forward, the index is
+  /// incremented. Only valid elements can be accessed.
+  /// 0 <= idx_ <= handle_->size()
+  HandleTy handle_;
+  llvm::ArrayRef<size_t> sizes_;
+  size_t idx_;
+  /// Holds true if the underlying tensor has non-trivial alignment (i.e. not 1)
+  bool isAligned_;
+
+  HandleIterator() = default;
+
+  HandleIterator(HandleTy handle) : handle_(handle) {
+    sizes_ = handle->dims();
+    isAligned_ = handle->size() < handle->actualSize();
+  }
+
+  static HandleIterator begin(HandleTy handle) {
+    auto res = HandleIterator(handle);
+    res.idx_ = 0;
+    return res;
+  }
+
+  static HandleIterator end(HandleTy handle) {
+    auto res = HandleIterator(handle);
+    res.idx_ = res.handle_->size();
+    return res;
+  }
+
+  friend class Handle<ElemTy>;
+
+public:
+  HandleIterator &operator++() {
+    if (*this != handle_->end()) {
+      idx_++;
+    }
+    return *this;
+  }
+  HandleIterator &operator--() {
+    if (idx_) {
+      idx_--;
+    }
+    return *this;
+  }
+  HandleIterator operator+(int n) const {
+    auto res = HandleIterator(handle_);
+    res.idx_ = std::max(static_cast<int>(idx_) + n, 0);
+    res.idx_ = std::min(res.idx_, res.handle_->size());
+    return res;
+  }
+  HandleIterator operator-(int n) const { return *this + (-n); }
+  operator int() const { return idx_; }
+
+  ElemTyRef operator*() {
+    if (!isAligned_) {
+      return handle_->raw(idx_);
+    }
+    std::vector<size_t> indices(sizes_.size(), 0);
+    size_t rem = idx_;
+    for (int i = static_cast<int>(sizes_.size()) - 1; i >= 0; i--) {
+      indices[i] = rem % sizes_[i];
+      rem /= sizes_[i];
+    }
+    return handle_->at(indices);
+  }
+
+  bool operator==(const HandleIterator<ElemTy, IsConst> &other) const {
+    return idx_ == other.idx_;
+  }
+
+  bool operator!=(const HandleIterator<ElemTy, IsConst> &other) const {
+    return !(*this == other);
+  }
+};
+
 /// A class that provides indexed access to a tensor. This class has value
 /// semantics and it's copied around. One of the reasons for making this class
 /// value semantics is to allow efficient index calculation that the compiler
@@ -703,28 +791,27 @@ template <class ElemTy> class Handle final {
   /// Saves the number of dimensions used in the tensor.
   uint8_t numDims_{0};
 
+  /// Remember end iterators. This is needed to speed up iterator increment,
+  /// which has to check that iterator hasn't reached the end yet.
+  HandleIterator<ElemTy, false> mutating_end_;
+  HandleIterator<ElemTy, true> const_end_;
+
   /// Create a new invalid handle. Notice that this method is private and may
   /// only be used by the static factory method below.
   Handle() = default;
 
 public:
-  /// Random access iterator to tensor elements.
-  using iterator = ElemTy *;
-
-  /// Constant random access iterator to tensor elements.
-  using const_iterator = const iterator;
-
   /// \returns an iterator to the first element of the tensor.
-  iterator begin() { return tensor_->getRawDataPointer<ElemTy>(); }
-  const_iterator begin() const { return tensor_->getRawDataPointer<ElemTy>(); }
+  HandleIterator<ElemTy, false> begin() {
+    return HandleIterator<ElemTy, false>::begin(this);
+  }
+  HandleIterator<ElemTy, true> begin() const {
+    return HandleIterator<ElemTy, true>::begin(this);
+  }
 
   /// \returns an iterator referring to the past-the-end element.
-  iterator end() {
-    return tensor_->getRawDataPointer<ElemTy>() + tensor_->actualSize();
-  }
-  const_iterator end() const {
-    return tensor_->getRawDataPointer<ElemTy>() + tensor_->actualSize();
-  }
+  HandleIterator<ElemTy, false> end() { return mutating_end_; }
+  HandleIterator<ElemTy, true> end() const { return const_end_; }
 
   /// Allocate a new invalid handle.
   static Handle createInvalidHandle() { return Handle(); }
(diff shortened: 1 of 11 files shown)


include/glow/Base/Tensor.h
```diff
@@ -682,9 +695,11 @@ template <class ElemTy> class Handle final {
   const_iterator begin() const { return tensor_->getRawDataPointer<ElemTy>(); }
 
   /// \returns an iterator referring to the past-the-end element.
-  iterator end() { return tensor_->getRawDataPointer<ElemTy>() + size(); }
+  iterator end() {
+    return tensor_->getRawDataPointer<ElemTy>() + tensor_->actualSize();
```
putivsky: Don't you think it will break iteration? Because iterator now can point to the "garbage" elements after incrementation beyond actual dimension size.

artemrakhov: Yes, it will break iteration. Now we need to be extra careful about where and why we use iteration.

It is worth to mention that I don't have intention to support custom-aligned Tensors for every backend. It is OK if Interpreter doesn't work correctly for non-1-aligned Tensors. The goal of this PR is to simply allow having such Tensors.

jackm321: > Yes, it will break iteration. Now we need to be extra careful about where and why we use iteration.

Maybe we should make a custom iterator?

artemrakhov: Implemented custom iterator in a separate commit

artemrakhov:
@putivsky @jackm321 : please either accept the PR or request changes

include/glow/Base/Tensor.h
putivsky: May be you can add assert() to the iterator begin(), end() methods, for non-zero strides? It will prevent the incorrect usage, till you will implement iterators with strides.

artemrakhov: Yesterday I added a commit to this PR, which implements `HandleIterator` class.

include/glow/Base/Tensor.h
```diff
@@ -658,6 +658,62 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+  }
+
+  static HandleIterator end(HandleTy handle) {
+    auto result = HandleIterator(handle);
+    result.indices_[0] = result.sizes_.size() ? result.sizes_[0] : 1;
+    return result;
+  }
+
+  HandleIterator &operator++() {
+    size_t i = sizes_.size() ? sizes_.size() - 1 : 0;
```
putivsky: Layover approach can be a problem, may be if iterator hits the end, ++ doesn't move it to the beginning?

include/glow/Base/Tensor.h
```diff
@@ -658,6 +658,62 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+      typename std::conditional_t<IsConst, const ElemTy &, ElemTy &>;
+
+  HandleTy handle_;
+  llvm::ArrayRef<size_t> sizes_;
+  size_t indices_[max_tensor_dimensions] = {0};
+
+  HandleIterator(HandleTy handle) : handle_(handle) { sizes_ = handle->dims(); }
+
+public:
+  static HandleIterator begin(HandleTy handle) {
```
putivsky: May be Handle class can be a friend class for HandleIterator and only Handle class can create an iterator instance?

include/glow/Base/Type.h
```diff
@@ -567,14 +617,30 @@ struct Type final {
...
   /// used by the constructor.
-  void initDims(llvm::ArrayRef<size_t> dims) {
+  /// \param dims of the tensor (in elements).
+  /// \param alignments of the tensor (in bytes).
+  void initDims(llvm::ArrayRef<size_t> dims,
+                llvm::ArrayRef<size_t> alignments) {
     assert(dims.size() <= max_tensor_dimensions && "Too many dimensions.");
+    assert(dims.size() == alignments.size() &&
+           "The number of dimensions and alignments should be the same");
     // Update the tensor sizes.
```
jackm321: extending this comment would be good now that this is more complicated

artemrakhov:
@putivsky : addressed your comments! Thanks!

include/glow/Base/Tensor.h
```diff
@@ -187,6 +191,13 @@ class Tensor final {
   /// \returns the number of elements in the tensor.
```
jackm321: this comment should be updated

include/glow/Base/Tensor.h
```diff
@@ -187,6 +191,13 @@ class Tensor final {
...
   size_t size() const { return type_.size(); }
 
+  /// \returns the actual number of elements in the tensor taking striding into
+  /// account. Since size() does not take striding into account, size() is
+  /// always <= actualSize().
+  size_t actualSize() const {
```
jackm321: This is pretty confusing unless you go look at what these do. If I used `size()` not knowing about `actualSize()` I would be confused, I'm not sure what a better naming scheme would be though. Is there anything similar we could look to for names>?

include/glow/Base/Tensor.h
```diff
@@ -187,6 +191,13 @@ class Tensor final {
...
+    return type_.getSizeInBytes() / type_.getElementSize();
```
jackm321: why not move this into type like size()

include/glow/Base/Tensor.h
```diff
@@ -290,14 +301,16 @@ class Tensor final {
     unownedTensor.unpaddedSize_ = unpaddedSize_;
 
     if (offsets.size() == 0) {
-      assert(size() == unownedTensor.size() && "The size of the unowned tensor "
-                                               "should the same as the size of "
-                                               "the original tensor");
+      assert(actualSize() == unownedTensor.actualSize() &&
```
jackm321: why is size() not sufficient here?

include/glow/Base/Tensor.h
```diff
@@ -645,6 +658,69 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+    return *this;
+  }
+
+  ElemTyRef operator*() { return handle_->at(indices_); }
+
+  bool operator==(const HandleIterator<ElemTy, IsConst> &other) const {
+    if (this->handle_ != other.handle_)
+      return false;
+    size_t e = sizes_.size();
+    if (!e)
```
jackm321: parens

include/glow/Base/Tensor.h
```diff
@@ -645,6 +658,69 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+template <class ElemTy, bool IsConst>
+class HandleIterator : public std::iterator<std::forward_iterator_tag, ElemTy> {
+  using HandleTy = typename std::conditional_t<IsConst, const Handle<ElemTy> *,
+                                               Handle<ElemTy> *>;
+  using ElemTyRef =
+      typename std::conditional_t<IsConst, const ElemTy &, ElemTy &>;
+
+  HandleTy handle_;
+  llvm::ArrayRef<size_t> sizes_;
+  size_t indices_[max_tensor_dimensions] = {0};
```
jackm321: comments

include/glow/Base/Tensor.h
```diff
@@ -645,6 +658,69 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+    result.indices_[0] = result.sizes_.size() ? result.sizes_[0] : 1;
+    return result;
+  }
+
+  friend class Handle<ElemTy>;
+
+public:
+  HandleIterator &operator++() {
+    // Check if the iterator already reached the end.
+    if (*this != end(handle_)) {
```
jackm321: Could probably cache this

include/glow/Base/Tensor.h
```diff
@@ -645,6 +658,69 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
 void dumpImpl(const Tensor *T, unsigned maxNumElem);
 void dumpImpl(const Tensor *T);
 
+template <class ElemTy> class Handle;
+
+template <class ElemTy, bool IsConst>
+class HandleIterator : public std::iterator<std::forward_iterator_tag, ElemTy> {
```
jackm321: @putivsky mentioned this earlier I think, you could have pre-check whether the tensor has any alignments and if not this can just fall back to a basically just a pointer with no additional logic

include/glow/Base/Tensor.h
```diff
@@ -645,6 +658,69 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+      while (i > 0 && indices_[i] >= sizes_[i]) {
+        indices_[i] = 0;
+        i--;
+        indices_[i]++;
+      }
+    }
+    return *this;
+  }
+
+  ElemTyRef operator*() { return handle_->at(indices_); }
```
jackm321: Seems also weird that things like `at()` will allow users to access garbage data

jackm321:
Another thing to consider is how this will interact with the partial tensors added by @bertmaher and @gcatron.
Also, can we add some TensorTests for this?

bertmaher:
It makes me a little sad that the probably-common case of unstrided tensor iterators is going to go from being a single pointer to a non-trivial data structure with a complex increment operator.  Maybe it doesn't matter, though, and I can't really think of any way around it unless we want iteration (or a variant of it) to ignore strides (and thus be kind of dangerous/unintuitive).

jfix71:
>It makes me a little sad that the probably-common case of unstrided tensor iterators is going to go from being a single pointer to a non-trivial data structure with a complex increment operator. Maybe it doesn't matter, though, and I can't really think of any way around it unless we want iteration (or a variant of it) to ignore strides (and thus be kind of dangerous/unintuitive).

I wonder if we could subclass Handle into strided and unstrided versions. Then when you call `Tensor::getHandle()` check it could check if there are non-zero strides to determine which to return.

include/glow/Base/Tensor.h
artemrakhov: Unowned tensor assumes that no data was copied. If actual size is different, then strides were changed, then probably we may address "garbage" element somewhere, which is unwanted.

In any way, how could this happen? When we create unowned tensor, we don't even pass new strides. It's just safer to verify more (that actual sizes match).

include/glow/Base/Tensor.h
artemrakhov: But what if we want to do raw copy of the Tensor?

artemrakhov:
> > It makes me a little sad that the probably-common case of unstrided tensor iterators is going to go from being a single pointer to a non-trivial data structure with a complex increment operator. Maybe it doesn't matter, though, and I can't really think of any way around it unless we want iteration (or a variant of it) to ignore strides (and thus be kind of dangerous/unintuitive).
> 
> I wonder if we could subclass Handle into strided and unstrided versions. Then when you call `Tensor::getHandle()` it could check if there are non-zero strides to determine which to return.

Or we can have Handle to conditionally return either one or the other iterator type, not to increase complexity.

artemrakhov:
@jackm321 : addressed most of your comments, please take a look

artemrakhov:
My motivation for leaving `Tensor::size()` is that we don't want to change whole codebase, possibly break any existing forks, and leave such a good name unused. Decision was between making `Tensor::size()` to return either padded or unpadded size. Number of real elements is definitely more important than "actual size". In my opinion, it's like `std::vector::size()` vs `std::vector::capacity`.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1091,8 +1091,10 @@ ONNXModelLoader::loadConstantOfShape(const ONNX_NAMESPACE::NodeProto &op,
                       "Input element type must be Int64ITy.");
     // Convert 1D tensor of int64_t into llvm::ArrayRef<size_t>.
     auto TH = in->getPayload().getHandle<int64_t>();
-    llvm::ArrayRef<size_t> outputDims = {(const size_t *)TH.begin(),
-                                         (const size_t *)TH.end()};
+    auto begin = &TH.raw(0);
```
jackm321: can use pointers to begin() and end() here

include/glow/Base/Tensor.h
jackm321: Wait so because we are operating with offsets from `firstElemPtr` here, getUnowned won't take strides into account and we'll end up with garbage values

artemrakhov:
Talked to @jackm321 offline, decided to change HandleIterator to support 2 modes. If Tensor has non-trivial alignments, then access through indices is used (`at`). Otherwise we use fast `at`.

Also, the usage pattern for Tensor is now as follows:
If someone is writing algorithmic optimization, they should not need to access garbage elements. They will use `Tensor::size()` and access Handle through Iterator or use `Handle::at()` method.
If someone is using raw memory API (`Handle::getRawDataPointer()` or `Handle::raw`), then they will need to account for padding. They will probably also need `Tensor::actualSize()`, and consider "holes" in the data, because the data might not be contigious anymore.

Most of the review comments are addressed, still need to add more tests.

include/glow/Base/Tensor.h
artemrakhov: Done.

include/glow/Base/Tensor.h
artemrakhov: The proposal is to have intuitive names for cases, when we don't want to worry about garbage data: `size()`, `at()`, iteration.
If user knows what they're doing, they can use `raw()`, `actualSize()`.

lib/Importer/ONNXModelLoader.cpp
artemrakhov: Can't. I changed the iterator to implement random access, but still not able to convert it to ArrayRef. It doesn't have constructor from `{TH.begin(), TH.end()}`

artemrakhov:
Fixed `getDimForPtr` and `getUnowned`.
Updated `HandleIterator` to implement random access iterator.

include/glow/Base/Tensor.h
artemrakhov: Good catch. Fixed now.

include/glow/Base/Tensor.h
```diff
@@ -669,6 +681,95 @@ void dumpImpl(const Tensor *T, llvm::raw_ostream &os,
...
+    auto res = HandleIterator(handle_);
+    res.idx_ = std::max(static_cast<int>(idx_) + n, 0);
+    res.idx_ = std::min(res.idx_, res.handle_->size());
+    return res;
+  }
+  HandleIterator operator-(int n) const { return *this + (-n); }
+  operator int() const { return idx_; }
+
+  ElemTyRef operator*() {
+    if (!isAligned_) {
```
jackm321: super nit but this could be moved into the handle then the size of the iterator can be reduced and indexed access on strided handles doesn't need an interator necessarily. Up to you though

artemrakhov: I believe that optimizing the Iterator is not a priority. And from code readability point of view I'd rather not add more fields to Handle.

artemrakhov:
Added 5 more tests

opti-mix:
@artemrakhov We discussed it before and it does not necessarily need to be part of this PR, but it could be useful to add some checks to the Node and/or IR verifiers used by different existing backends to make sure that by default no types with non-trivial alignments are used anywhere. E.g. our Interpreter and CPU backends currently implemented using this assumption (many interpreter/libjit functions use raw access to tensors memories). So let's check it that it holds. Those backends that need to support non-trivial alignments can skip this check or replace it by something more specific.

PR #3369, opened 2019-08-02
[PyTorch] Allow caller of jitVsGlow to specify expected fused ops

SplitInfinity:
**Description:**
This commit adds a parameter to `jitVsGlow` that the caller must use to
specify exactly which ops are expected to be placed into a Glow
compilation group. `jitVsGlow` checks that all ops outside of Glow
compilation groups are not in the list.

**Test Plan:**
All tests have been modified to use this new argument and they still
pass. After deleting a mapping in `symbolLoaderMapping`, the test
corresponding to that mapping fails.

Note that the transpose tests had to be disabled because the transpose
nodes don't get fused into the Glow fusion group for some reason.

### DIFF
--- a/torch_glow/tests/utils.py
+++ b/torch_glow/tests/utils.py
@@ -4,17 +4,23 @@ import torch
 import torch_glow
 
 GLOW_NODE_NAME = "glow::FusionGroup"
+SUBGRAPH_ATTR = "Subgraph"
 
 
-def jitVsGlow(f, *inputs):
+def jitVsGlow(f, *inputs, expected_fused_ops=None):
     """
-    Runs the given inputs *inputs on f both with and without lowering f to Glow
-    and compares the results.
+    Runs the given inputs *inputs on f both with and without lowering f to Glow,
+    compares the results, and checks that ops in expected_fused_ops were indeed
+    lowered to Glow.
     """
-    jitVsGlow_(f, f, *inputs)
+    jitVsGlow_(f, f, *inputs, expected_fused_ops=expected_fused_ops)
 
 
-def jitVsGlow_(f_torch, f_glow, *inputs):
+def jitVsGlow_(f_torch, f_glow, *inputs, expected_fused_ops):
+    assert (
+        expected_fused_ops is not None and len(expected_fused_ops) > 0
+    ), "Must pass non-empty list of ops that are expected to be fused"
+
     with torch.no_grad():
         torch_glow.disableFusionPass()
         torch_trace = torch.jit.trace(f_torch, inputs)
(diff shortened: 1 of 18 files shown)


SplitInfinity:
This is not done yet. I want to get some feedback on the approach before I go ahead and modify all of the tests to use this feature.

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,21 @@
+from argparse import Namespace
+
 import torch
 import torch_glow
 
 GLOW_NODE_NAME = "glow::CompilationGroup"
+SUBGRAPH_ATTR = "Subgraph"
+OP_KIND = Namespace(Add="aten::add", AddInplace="aten::add_")
```
SplitInfinity: I hope this already exists somewhere and someone can point me to it.

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,21 @@
+from argparse import Namespace
+
 import torch
 import torch_glow
 
 GLOW_NODE_NAME = "glow::CompilationGroup"
+SUBGRAPH_ATTR = "Subgraph"
```
SplitInfinity: I hope this already exists somewhere and someone can point me to it.

torch_glow/tests/utils.py
```diff
@@ -23,17 +27,28 @@ def jitVsGlow_(f_torch, f_glow, *inputs):
...
+                glow_group = node.g(SUBGRAPH_ATTR)
+
+                for op in glow_group.nodes():
+                    op_kind = op.kind()
+                    if "prim::" not in op_kind:
+                        assert op_kind in expected_fused_ops, "Did not expect {} to be fused".format(
+                            op_kind)
+            else:
+                # If the node is not a Glow compilation group, check that it is *not* in expected_fused_ops
+                if "prim::" not in kind:
```
jackm321: is this prim:: check needed?

torch_glow/tests/utils.py
jackm321: I don't think it exists for all nodes, the point is that these are supposed to be registerable. I don't think this is really needed, it would be fine to have `expected_fused_ops=["aten::add"]`

torch_glow/tests/utils.py
SplitInfinity: I noticed that some constants in the IR labelled as `Prim::Constants` get moved into the Glow compilation group and some don't based on what is and is not in the group. Maybe I should make this more specific?

torch_glow/tests/utils.py
SplitInfinity: I'm just worried that we'll repeat `aten::add` (and similar strings) in a bunch of places. The add tests themselves have regular, inplace and broadcasting tests, and add is used in other operator tests to make sure there is more than one op for PT to fuse.

jackm321: >I'm just worried that we'll repeat aten::add (and similar strings) in a bunch of places.

I still think having "aten::add" repeated a bunch of times is better than adding this random other Namespace stuff (imo).

>add is used in other operator tests to make sure there is more than one op for PT to fuse."

Thats a good point, I think we should check that all of things that should be in the subgraph are in the subgraph and that nothing that should be in the subgraph is not in the subgraph but not check that other things are not in the subgraph. what do you think?

torch_glow/tests/utils.py
jackm321: True, sounds good let's keep it

torch_glow/tests/utils.py
jackm321: Sorry meant to say that also other things can be in the subgraph even if not in expected_fused_ops

SplitInfinity: Okay, let's check that all ops in `expected_fused_ops` are found inside the Glow compilation group and not outside. The test should not fail if any ops that are *not* in that list can appear inside or outside of the group.

torch_glow/tests/utils.py
```diff
@@ -23,17 +31,19 @@ def jitVsGlow_(f_torch, f_glow, *inputs):
...
         # check that there are no Glow nodes in the torch graph
         torch_graph = torch_trace.graph_for(*inputs)
-        print("torch_graph,", torch_graph)
         num_glow_nodes = len(torch_graph.findAllNodes(GLOW_NODE_NAME))
         assert num_glow_nodes == 0, "Expected no Glow nodes, found {}".format(
             num_glow_nodes
         )
 
-        # check that there is exactly 1 Glow node in the glow graph
+        # Check that only ops in expected_fused_ops were fused
```
jackm321: the logic looks right but not the comment
It should be that no expected_fused_ops ops were not fused

I also think we should add a check to make sure that there is at least one of each op in expected_fused_ops in the graph because otherwise the test op may get lowered to something else that isn't fused and there is no warning of it

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,25 @@
+from argparse import Namespace
```
jackm321: delete

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,25 @@
...
+
 import torch
 import torch_glow
 
+
 GLOW_NODE_NAME = "glow::CompilationGroup"
+SUBGRAPH_ATTR = "Subgraph"
```
jackm321: delete

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,25 @@
+from argparse import Namespace
+
 import torch
 import torch_glow
 
+
 GLOW_NODE_NAME = "glow::CompilationGroup"
```
jackm321: this has been renamed to `glow::FusionGroup`

torch_glow/tests/utils.py
```diff
@@ -1,17 +1,25 @@
...
+SUBGRAPH_ATTR = "Subgraph"
 
 # Runs the given inputs \p *inputs on \p f both with and without lowering \p f
```
jackm321: update this comment with expected_fused_ops

torch_glow/tests/utils.py
jackm321: In fact here's an example where this could have helped
https://github.com/pytorch/glow/blob/30df28ebd82cd55ecc9957473558cf62b3244058/torch_glow/tests/nodes/addmm_test.py#L8-L12

torch_glow/tests/utils.py
SplitInfinity: Still need it now that we want to check that all ops in `expected_fused_ops` are present in the graph.

PR #3376, opened 2019-08-05
Enable deduplication of constants with NaNs. (#3328)

chadrosier:
Summary:
Constants that contain NaN will not be de-duplicated because isEqual always returns false for this case. Instead, perform a bitwise comparison to achieve the expected behavior.


rdzhabarov:
I'll merge once it passes internal testing.

PR #3378, opened 2019-08-05
PT differentiated graph example

mortzur:
Summary:
Adding a differentiated graph example.


rdzhabarov:
ping @jackm321 @zrphercule

jackm321:
@mortzur what would you like to do with this? Do you want to commit it as is, make a test out of it, do something else?

PR #3380, opened 2019-08-05
[PyTorch] Add sigmoid op

zrphercule:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -239,6 +239,7 @@ struct LinearInputs {
 // static
 const PyTorchModelLoader::MappingOfMemberFunctions &
 PyTorchModelLoader::getSymbolsMapping() {
+
   /// Static map of the set of PyTorch symbols to load, the PyTorchModelLoader
   /// for loading these symbols, and the set of inputs that should be considered
   /// immutable between inference invocations by Glow and loaded as Constants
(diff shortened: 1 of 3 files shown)


torch_glow/tests/nodes/sigmoid_test.py
```diff
@@ -0,0 +1,27 @@
...
+    jitVsGlow(sigmoid_basic, x)
+
+def test_sigmoid_inplace():
+    def sigmoid_inplace(a):
+        c = a.sigmoid_()
+        return c.sigmoid_()
+
+    x = torch.randn(6)
+
+    jitVsGlow(sigmoid_inplace, x)
```
yinghai: We support inplace because we copy Glow's output back to `x`?

zrphercule: In this case, I think the answer is yes, although I didnt see any using scenario

torch_glow/tests/nodes/sigmoid_test.py
```diff
@@ -0,0 +1,27 @@
...
+import torch
+import torch_glow
+
+from tests.utils import jitVsGlow
+
+# Basic test of the PyTorch sigmoid Node on Glow
+
+
+def test_sigmoid_basic():
+    def sigmoid_basic(a):
```
jackm321: I started just calling these `test_f` so I don't need to come up with a new name each time, you can do the same or not up to you

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -226,6 +226,7 @@ struct LinearInputs {
 // static
 const PyTorchModelLoader::MappingOfMemberFunctions &
 PyTorchModelLoader::getSymbolsMapping() {
+
```
jackm321: ?

PR #3384, opened 2019-08-05
[Partitioner] Partitioner refactoring 3

beicy:
Summary:
In this PR:
1. Added `class ParititionerBase` which contains interfaces and common functions used for partitioner.  and the `class Partitioner` is now derived from `class ParititionerBase` .
2. In `class Partitioner`, we have 3 partition approaches: 
```
llvm::Error HeterogeneousPartition(CompilationContext &cctx);
llvm::Error QuantizationProfilingPartition(CompilationContext &cctx);
llvm::Error PartitionFromConfig(PartitionConfig &partitionConfig);
````
The above 3 method can be called directly by users, or chosen by  
`llvm::Error Partition(CompilationContext &cctx)`

Note:  `loadBalancedPartitioning` is still included in `HeterogeneousPartition` since it uses the result of current partitioner approach. Will take it from `HeterogeneousPartition` in next PR and make it as the 4th partition approach. 

Documentation:

[Optional Fixes #issue]
#2298
Test Plan:
Ninja test, ./tests/image/run.sh

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -16,7 +16,7 @@
 #ifndef GLOW_PARTITIONER_PARTITIONER_H
 #define GLOW_PARTITIONER_PARTITIONER_H
 
-#include "glow/Partitioner/PartitionerTypes.h"
+#include "glow/Partitioner/PartitionerBase.h"
 #include "glow/Support/Error.h"
 
 namespace glow {
@@ -25,7 +25,7 @@ using namespace runtime;
 
 /// Given a module, partitions each of the its functions into multiple ones
 /// based on memory constraints and minimizes the communication cost.
-class Partitioner {
+class Partitioner final : public PartitionerBase {
   /// The module that needs to be decomposed.
   Module *module_;
 
@@ -36,7 +36,10 @@ class Partitioner {
   /// The cost model related to device.
   std::vector<DeviceInfo> deviceInfo_;
 
-  /// The backend pointers.
+  /// The backends created in Partitioner. Used for function optimization.
+  std::vector<std::unique_ptr<Backend>> backendHolder;
+
+  /// The raw backend pointers.
   std::vector<Backend *> backends_;
 
   /// The map between backend name and BackendInfo.
@@ -49,9 +52,6 @@ class Partitioner {
   /// The number of logicalDevice IDs, i.e. the number of physical devices
   /// needed after partitions.
   DeviceIDTy logicalDeviceID_;
-
-  /// The result of module partitioning.
-  DAGListTy partitions_;
 
   /// Total memory (bytes) requested by one module.
   uint64_t memSize_;
@@ -82,11 +82,19 @@ class Partitioner {
   NodeToFunctionMap selectPartitions(Function *F, uint64_t availableMemory,
                                      llvm::StringRef backendName);
 
-  /// Duplicates all networks in the module order to saturate the Host.
-  void saturateHost(unsigned logicalDeviceCount);
+  /// Duplicates \p partitions in the module order to saturate the Host. \p
+  /// logicalDeviceCount is the number of logical devices used by the current
+  /// partitions. For example: If a network is partitioned into two parts (\p
+  /// logicalDeviceCount) and there are six devices this would duplicate the
+  /// network three times.
+  void saturateHost(unsigned logicalDeviceCount, const DAGListTy &partitions);
 
-  FunctionToBackendNameMap
-  backendBasedPartition(Function *F, std::vector<Backend *> &backends,
+  /// Partition a function \p F based on backends \p backends. \returns the
+  /// final partition result(or an err) and a map between partitions and backend
+  /// names. \p cctx is used for functions optimization.
+  llvm::Expected<DAGListTy>
+  backendBasedPartition(FunctionToBackendNameMap &funcToBackend, Function *F,
+                        std::vector<Backend *> &backends,
                         CompilationContext &cctx);
 
   /// Performs a load balancing optimization pass to optimize for load
@@ -96,26 +104,20 @@ class Partitioner {
                                        llvm::StringRef backendName,
                                        NodeToFunctionMap &mapping);
 
-  /// Given the node-function mapping, do the actual partitioning. If \p saveDAG
-  /// is true, the DAG will be saved into partitions_, which is the final
-  /// partition result.
-  void doPartitioning(llvm::StringRef funcName, std::vector<Function *>,
-                      NodeToFunctionMap &mapping, bool saveDAG);
-
   /// If there is no need to do any partition, just generate the DAGNode based
   /// on current functions in this module for backend \p backendName found in \p
-  /// backendMap. \p cctx is used during optimization of the Function. \returns
-  /// whether there was an error encountered.
-  llvm::Error
+  /// backendMap. \p cctx is used for function optimization. \returns the
+  /// partition result or an error.
+  llvm::Expected<DAGListTy>
   createDAGWithoutPartition(llvm::StringRef backendName,
                             std::map<std::string, BackendInfo> &backendMap,
                             CompilationContext &cctx);
 
-  /// Get the map between the backend name and the concrete backend info (e.g.
-  /// backend pointer, mem, number) used in this partiton. If there are backends
-  /// need to be created, we use \p backendsHolder to hold them for memory
-  /// purpose.
-  void getBackendMap(std::map<std::string, BackendInfo> &backendMap,
+  /// Create the map between the backend name and the concrete backend info
+  /// (e.g. backend pointer, mem, number) used in this partiton. If there are
+  /// backends need to be created, we use \p backendsHolder to hold them for
+  /// memory purpose.
+  void genBackendMap(std::map<std::string, BackendInfo> &backendMap,
                      std::vector<std::unique_ptr<Backend>> &backendsHolder,
                      std::vector<Backend *> &backends);
 
@@ -141,30 +143,30 @@ public:
               const std::vector<Backend *> &backends, bool saturateHost = false,
               bool optimized = false);
 
-  /// Based on partitionConfig_ passed into Partitioner, do the user-defined
+  /// Based on \p partitionConfig passed into Partitioner, do user-defined
   /// partition.
-  llvm::Error PartitionFromConfig();
-
-  /// Decompose each function in a module. Now we support partitioning a module
-  /// among different type of devices. \p cctx is used during optimization of
-  /// the Function. \returns whether there was an error encountered.
-  llvm::Error Partition(CompilationContext &cctx);
+  llvm::Expected<DAGListTy>
+  partitionFromConfig(const PartitionConfig &partitionConfig);
 
   /// This partition approach is used in Glow Quantization Profiling flow. The
   /// backendBasedPartition is applied first in case there are heterogeneous
   /// backends. Then each sub-function will be compiled and run in CPU backend
-  /// for profiling.
-  llvm::Error QuantizationProfilingPartition(CompilationContext &cctx,
-                                             Function *F,
-                                             std::vector<Backend *> backends);
+  /// for profiling. \p cctx is used for function optimization. \returns the
+  /// partition result or an error.
+  llvm::Expected<DAGListTy>
+  quantizationProfilingPartition(CompilationContext &cctx);
 
-  /// Get the final partitions.
-  DAGListTy &getPartitionResult() { return partitions_; }
+  /// This partition approch first do the partition based on backend types, and
+  /// then based on cost models(memory usage and performance). \p cctx is used
+  /// for function optimization. \returns the partition result or an error.
+  llvm::Expected<DAGListTy> heterogeneousPartition(CompilationContext &cctx);
 
-  /// Dump the partition result to a dot file. Since now all functions belong to
-  /// a function family and they have the same partition, we only dump the one
-  /// function's partition.
-  void dumpDAG(llvm::StringRef dotFilename) const;
+  /// Decompose each function in a module. Given the parameters, this function
+  /// will choose different partition approches supported in this class:
+  /// heterogeneous partition, user-defined partition or quantization profiling.
+  /// \p cctx is used for function optimization. \returns the partition result
+  /// or an error.
+  llvm::Expected<DAGListTy> partition(CompilationContext &cctx) override;
 };
 } // namespace glow
 #endif // GLOW_PARTITIONER_PARTITIONER_H
--- a/include/glow/Partitioner/PartitionerBase.h
+++ b/include/glow/Partitioner/PartitionerBase.h
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_PARTITIONER_PARTITIONERBASE_H
+#define GLOW_PARTITIONER_PARTITIONERBASE_H
+
+#include "glow/Partitioner/PartitionerTypes.h"
+#include "glow/Support/Error.h"
+
+namespace glow {
+
+using namespace runtime;
+/// Given a module, partitions each of the its functions into multiple ones
+/// based on memory constraints and minimizes the communication cost.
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: include/glow/Partitioner/PartitionerBase.h [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+
+#include "glow/Partitioner/PartitionerTypes.h"
+#include "glow/Support/Error.h"
+
+namespace glow {
+
+using namespace runtime;
+/// Given a module, partitions each of the its functions into multiple ones
+/// based on memory constraints and minimizes the communication cost.
+class PartitionerBase {
```
rdzhabarov: let's clean up this interface a bit as discussed offline.

include/glow/Partitioner/PartitionerBase.h
```diff
@@ -0,0 +1,57 @@
...
+  DAGListTy partitions_;
+
+public:
+  PartitionerBase() {}
+
+  virtual ~PartitionerBase() = default;
+
+  /// Decompose each function in a module. \p cctx is used in function
+  /// optimization. \returns whether there was an error encountered.
+  virtual llvm::Error Partition(CompilationContext &cctx) {
```
bertmaher: Nit: should be lowercase `p` to match our coding conventions.

include/glow/Partitioner/Partitioner.h
```diff
@@ -143,28 +136,22 @@ class Partitioner {
 
   /// Based on partitionConfig_ passed into Partitioner, do the user-defined
   /// partition.
-  llvm::Error PartitionFromConfig();
-
-  /// Decompose each function in a module. Now we support partitioning a module
-  /// among different type of devices. \p cctx is used during optimization of
-  /// the Function. \returns whether there was an error encountered.
-  llvm::Error Partition(CompilationContext &cctx);
+  llvm::Error PartitionFromConfig(PartitionConfig &partitionConfig);
```
bertmaher: Same initial-caps nit, here and a few places below.

include/glow/Partitioner/PartitionerBase.h
beicy: Done.

beicy:
> BTW while you're refactoring, how do you feel about dropping the `Ty` suffix from the partitioner/runtime types? I think @bertmaher has mentioned this before. E.g. `DAGListTy`, `DeviceIDTy`, etc.

For "Ty", if we decide to move that, I think it is better to get a separate PR, since other parts of Runtime (e.g. Provisioner) use those date type as well.

include/glow/Partitioner/Partitioner.h
```diff
@@ -25,18 +25,23 @@ using namespace runtime;
...
+class Partitioner final : public PartitionerBase {
   /// The module that needs to be decomposed.
   Module *module_;
-
   /// The representative function used for partition. We choose the function who
   /// has the largest memory size.
   Function *F_;
 
+  /// The partition result.
+  DAGListTy partitions_;
```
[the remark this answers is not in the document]
beicy: "PartitionerBase::partition() will always return a reference to this DAGListTy, right?"
-- Yes.
"Would it make sense for this to be moved to PartitionerBase? "
-- After the offline discussion with @rdzhabarov, we would like to make the PartitionerBase "stateless".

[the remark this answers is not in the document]
rdzhabarov: Question: do we really need to keep  `DAGListTy partitions_;` in partitioner.

Role of the partitioner is to generate `DAGList` and return it back to the caller.
What do we gain by keeping partitions_ here?

include/glow/Partitioner/Partitioner.h
```diff
@@ -96,26 +98,20 @@ class Partitioner {
...
   createDAGWithoutPartition(llvm::StringRef backendName,
                             std::map<std::string, BackendInfo> &backendMap,
                             CompilationContext &cctx);
 
-  /// Get the map between the backend name and the concrete backend info (e.g.
-  /// backend pointer, mem, number) used in this partiton. If there are backends
-  /// need to be created, we use \p backendsHolder to hold them for memory
-  /// purpose.
-  void getBackendMap(std::map<std::string, BackendInfo> &backendMap,
+  /// Greate the map between the backend name and the concrete backend info
```
jfix71: typo: Greate -> Create

include/glow/Partitioner/Partitioner.h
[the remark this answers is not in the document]
beicy: Removed `DAGListTy partitions_;` in partitioner.

include/glow/Partitioner/Partitioner.h
```diff
@@ -83,10 +82,12 @@ class Partitioner {
                                      llvm::StringRef backendName);
 
   /// Duplicates all networks in the module order to saturate the Host.
-  void saturateHost(unsigned logicalDeviceCount);
+  void saturateHost(unsigned logicalDeviceCount, const DAGListTy &partitions);
 
-  FunctionToBackendNameMap
-  backendBasedPartition(Function *F, std::vector<Backend *> &backends,
+  /// Do the partition based on backend types.
```
rdzhabarov: could you add some description what F is and what's stored in functobackend map etc.

include/glow/Partitioner/Partitioner.h
```diff
@@ -143,28 +138,24 @@ class Partitioner {
...
   /// Based on partitionConfig_ passed into Partitioner, do the user-defined
   /// partition.
-  llvm::Error PartitionFromConfig();
-
-  /// Decompose each function in a module. Now we support partitioning a module
-  /// among different type of devices. \p cctx is used during optimization of
-  /// the Function. \returns whether there was an error encountered.
-  llvm::Error Partition(CompilationContext &cctx);
+  llvm::Expected<DAGListTy>
+  partitionFromConfig(PartitionConfig &partitionConfig);
```
rdzhabarov: why config is not a const here?

include/glow/Partitioner/Partitioner.h
```diff
@@ -143,28 +138,24 @@ class Partitioner {
...
 
   /// This partition approach is used in Glow Quantization Profiling flow. The
   /// backendBasedPartition is applied first in case there are heterogeneous
   /// backends. Then each sub-function will be compiled and run in CPU backend
   /// for profiling.
-  llvm::Error QuantizationProfilingPartition(CompilationContext &cctx,
-                                             Function *F,
-                                             std::vector<Backend *> backends);
+  llvm::Expected<DAGListTy>
+  quantizationProfilingPartition(CompilationContext &cctx);
```
rdzhabarov: why all these special methods are part of the public interface?

lib/Partitioner/Partitioner.cpp
```diff
@@ -723,30 +542,20 @@ llvm::Error Partitioner::QuantizationProfilingPartition(
         << "Profiling a model to be partitioned cross different backends. Each "
            "sub-network will be optimized and run on cpu backend.\n";
   }
-  return llvm::Error::success();
+  return partitions;
```
rdzhabarov: while refactoring can you replace asserts with DCHECK and add more detail to error message where applicable.

include/glow/Partitioner/Partitioner.h
beicy: Discussed with @artemrakhov before. It allows users to call those method directly. For example,  user-defined partitioner  can be called directly without setting params, e.g "optimized_" and "saturateHost_".  @artemrakhov could you please also help to double check the interface? Thanks!

include/glow/Partitioner/Partitioner.h
```diff
@@ -83,10 +83,13 @@ class Partitioner {
                                      llvm::StringRef backendName);
 
   /// Duplicates all networks in the module order to saturate the Host.
```
jfix71: nit: Update doxygen to mention the method parameters.

include/glow/Partitioner/Partitioner.h
```diff
@@ -83,10 +83,13 @@ class Partitioner {
...
-  void saturateHost(unsigned logicalDeviceCount);
+  void saturateHost(unsigned logicalDeviceCount, const DAGListTy &partitions);
 
-  FunctionToBackendNameMap
-  backendBasedPartition(Function *F, std::vector<Backend *> &backends,
+  /// Partition a function \p F based on backends \p backends. \returns the
+  /// final result and a map between partitions and backend names.
+  llvm::Expected<DAGListTy>
+  backendBasedPartition(FunctionToBackendNameMap &FuncToBackend, Function *F,
```
jfix71: nit: uncapitalize `FuncToBackend` (matches the cpp definition)

include/glow/Partitioner/Partitioner.h
```diff
@@ -83,10 +83,13 @@ class Partitioner {
                                      llvm::StringRef backendName);
 
   /// Duplicates all networks in the module order to saturate the Host.
-  void saturateHost(unsigned logicalDeviceCount);
+  void saturateHost(unsigned logicalDeviceCount, const DAGListTy &partitions);
 
-  FunctionToBackendNameMap
-  backendBasedPartition(Function *F, std::vector<Backend *> &backends,
+  /// Partition a function \p F based on backends \p backends. \returns the
+  /// final result and a map between partitions and backend names.
```
jfix71: nit: Update doxygen with all method params

include/glow/Partitioner/Partitioner.h
```diff
@@ -96,26 +99,20 @@ class Partitioner {
...
-  /// Given the node-function mapping, do the actual partitioning. If \p saveDAG
-  /// is true, the DAG will be saved into partitions_, which is the final
-  /// partition result.
-  void doPartitioning(llvm::StringRef funcName, std::vector<Function *>,
-                      NodeToFunctionMap &mapping, bool saveDAG);
-
   /// If there is no need to do any partition, just generate the DAGNode based
   /// on current functions in this module for backend \p backendName found in \p
   /// backendMap. \p cctx is used during optimization of the Function. \returns
   /// whether there was an error encountered.
```
jfix71: nit: This doesn't return if an error was encountered, it returns the DAG list or else an error.

include/glow/Partitioner/Partitioner.h
```diff
@@ -141,30 +138,26 @@ class Partitioner {
...
-  /// among different type of devices. \p cctx is used during optimization of
-  /// the Function. \returns whether there was an error encountered.
-  llvm::Error Partition(CompilationContext &cctx);
+  llvm::Expected<DAGListTy>
+  partitionFromConfig(const PartitionConfig &partitionConfig);
 
   /// This partition approach is used in Glow Quantization Profiling flow. The
   /// backendBasedPartition is applied first in case there are heterogeneous
   /// backends. Then each sub-function will be compiled and run in CPU backend
   /// for profiling.
```
jfix71: nit: `\returns ...` and mention cctx.

include/glow/Partitioner/Partitioner.h
```diff
@@ -141,30 +138,26 @@ class Partitioner {
...
-  llvm::Error QuantizationProfilingPartition(CompilationContext &cctx,
-                                             Function *F,
-                                             std::vector<Backend *> backends);
+  llvm::Expected<DAGListTy>
+  quantizationProfilingPartition(CompilationContext &cctx);
 
-  /// Get the final partitions.
-  DAGListTy &getPartitionResult() { return partitions_; }
+  /// This partition approch first do the partition based on backend types, and
+  /// then based on cost models(memory usage and performance).
```
jfix71: nit: `\returns ...` and mention cctx.

include/glow/Partitioner/Partitioner.h
```diff
@@ -141,30 +138,26 @@ class Partitioner {
...
   /// for profiling.
-  llvm::Error QuantizationProfilingPartition(CompilationContext &cctx,
-                                             Function *F,
-                                             std::vector<Backend *> backends);
+  llvm::Expected<DAGListTy>
+  quantizationProfilingPartition(CompilationContext &cctx);
 
-  /// Get the final partitions.
-  DAGListTy &getPartitionResult() { return partitions_; }
+  /// This partition approch first do the partition based on backend types, and
```
jfix71: nit: Couple typos: "approach" and "first does the partition"

include/glow/Partitioner/Partitioner.h
```diff
@@ -141,30 +138,26 @@ class Partitioner {
...
+  /// then based on cost models(memory usage and performance).
+  llvm::Expected<DAGListTy> heterogeneousPartition(CompilationContext &cctx);
 
-  /// Dump the partition result to a dot file. Since now all functions belong to
-  /// a function family and they have the same partition, we only dump the one
-  /// function's partition.
-  void dumpDAG(llvm::StringRef dotFilename) const;
+  /// Decompose each function in a module. Given the parameters, this function
+  /// will choose different partition approches supported in this class:
+  /// heterogeneous partition, user-defined partition or quantization profiling.
```
jfix71: nit: `\returns ...` and mention cctx.

include/glow/Partitioner/PartitionerBase.h
```diff
@@ -0,0 +1,49 @@
...
+/// based on memory constraints and minimizes the communication cost.
+class PartitionerBase {
+public:
+  virtual ~PartitionerBase() = default;
+
+  /// Decompose each function in a module. \p cctx is used in function
+  /// optimization. \returns the partition result.
+  virtual llvm::Expected<DAGListTy> partition(CompilationContext &cctx) = 0;
+
+  /// Dump the partition result to a dot file. Since now all functions belong to
```
jfix71: nit: mention parameters in doxygen

include/glow/Partitioner/PartitionerBase.h
```diff
@@ -0,0 +1,49 @@
...
+  /// a function family and they have the same partition, we only dump the one
+  /// function's partition.
+  void dumpDAG(llvm::StringRef dotFilename, const DAGListTy &partitions) const;
+
+protected:
+  /// Given the node-function mapping, do the actual partitioning. If \p saveDAG
```
jfix71: nit: `\p mapping` etc. and `\returns` instead of `partitions_`.

lib/Partitioner/Partitioner.cpp
```diff
@@ -237,13 +178,14 @@ NodeToFunctionMap Partitioner::selectPartitions(Function *F,
 /// Duplicate the network to saturate the number of devices. For example: If a
```
jfix71: Looks like this doxygen can be moved to the header?

lib/Partitioner/Partitioner.cpp
```diff
@@ -825,7 +641,7 @@ llvm::Error Partitioner::Partition(CompilationContext &cctx) {
   }
 
   // Step 5 : do the real partitioning for the function list.
-  doPartitioning(origName, funcs, mapping, true);
+  partitions = doPartitioning(origName, funcs, module_, mapping, true);
```
jfix71: nit: `/* saveDAG */ true`

lib/Partitioner/Partitioner.cpp
```diff
@@ -931,15 +751,15 @@ llvm::Error Partitioner::PartitionFromConfig() {
   // TODO : loop-free validation.
 
   // Do partition.
-  doPartitioning(F->getName(), {F}, partitionMap, true);
+  partitions = doPartitioning(F->getName(), {F}, module_, partitionMap, true);
```
jfix71: nit: `/* saveDAG */ true`

lib/Partitioner/PartitionerBase.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "glow/Partitioner/PartitionerBase.h"
+#include "glow/Optimizer/GraphOptimizer/GraphOptimizer.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/raw_ostream.h"
+#include <fstream>
+
+using namespace glow;
+using llvm::isa;
+
+/// Current only partition the representative function.
```
jfix71: nit: Don't need doxygen in cpp files.

include/glow/Partitioner/Partitioner.h
artemrakhov: Yes, this is the right direction! It is very useful to be able to call just one method.

But Roman wanted to bring a different question to your attention. Can we create more Partitioner-based (inherited) classes? I.e. we could further introduce `UserDefinedPartitioner`, `SaturatingPartitioner`, and so on. Whatever makes sense. That way, `Partitioner` class itself won't implement any "special" partitions. It will be a very general class.

beicy: > Yes, this is the right direction! It is very useful to be able to call just one method.
> 
> But Roman wanted to bring a different question to your attention. Can we create more Partitioner-based (inherited) classes? I.e. we could further introduce `UserDefinedPartitioner`, `SaturatingPartitioner`, and so on. Whatever makes sense. That way, `Partitioner` class itself won't implement any "special" partitions. It will be a very general class.

We now have "class PartitionerBase" and new partitioner class can be derived from that. The reason we have current partition flows in Partitioner because they shares lots of common data types, internal members.  We would like to keep "class PartitionerBase" stateless.

artemrakhov: We can do further inheritance. For example: PartitionerBase -> Partitioner -> UserDefinedPartitioner

PR #3387, opened 2019-08-06
[PyTorch] Fix torch tests missing expected_fused_ops

jackm321:
Summary:
Looks like some new tests were added while expected_fused_ops was landing.

Documentation:
n/a 

Test Plan:
`python setup.py test`

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -45,7 +45,7 @@ static llvm::cl::OptionCategory PartitionerCat("Glow Partitioner Options");
 static llvm::cl::opt<bool>
     logPartition("log-partition",
                  llvm::cl::desc("Enable logging partition info"),
-                 llvm::cl::init(true), llvm::cl::cat(PartitionerCat));
+                 llvm::cl::init(false), llvm::cl::cat(PartitionerCat));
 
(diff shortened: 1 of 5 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -45,7 +45,7 @@ static llvm::cl::OptionCategory PartitionerCat("Glow Partitioner Options");
 static llvm::cl::opt<bool>
     logPartition("log-partition",
                  llvm::cl::desc("Enable logging partition info"),
-                 llvm::cl::init(true), llvm::cl::cat(PartitionerCat));
+                 llvm::cl::init(false), llvm::cl::cat(PartitionerCat));
```
jackm321: @beicy this was making it pretty hard to see test outputs so I set it to false, let me know if you want me to change it back to true but it seems like it should default to false to me.

torch_glow/tests/utils.py
```diff
@@ -32,12 +32,16 @@ def jitVsGlow_(f_torch, f_glow, *inputs, expected_fused_ops):
 
         # check that there are no Glow nodes in the torch graph
         torch_graph = torch_trace.graph_for(*inputs)
+        print("torch_graph,", torch_graph)
```
jackm321: these are here intentionally, would like to keep them

PR #3391, opened 2019-08-06
Extended PyTorchFileLoader class.

putivsky:
Summary: Added Fusion JIT Graph loading into Glow Function.

Differential Revision: D16678267


torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,188 @@
...
+            return 0;
+          };
+        },
+        options)});
+
+    torch::jit::RegisterPass pass([&](std::shared_ptr<torch::jit::Graph> &g) {
+      // Trigger custom fusion pass only if local thread Glow Function is set.
+      if (localFusionInfo.function) {
+        // Fuse all linear operators.
+        FuseLinear(g);
```
bwasti: I guess backends would generally benefit from these canonicalization passes living in Pytorch's core, huh? @kimishpatel this is what we discussed in our meeting

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,188 @@
...
+/*static*/
+llvm::Error PyTorchFileLoader::loadPyTorchGraph(
+    const std::string &fileName, std::vector<torch::jit::IValue> &inputs,
+    glow::Function &F, std::vector<glow::Placeholder *> &inputPlaceholders,
+    std::vector<glow::Placeholder *> &outputPlaceholders, bool sanityCheck) {
+  // Register custom pass once.
+  static RegisterCustomFusionPass fuser;
+
+  // Convert PyTorch model into JIT Module.
+  std::shared_ptr<torch::jit::script::Module> module = {nullptr};
```
bwasti: nit: do you need to initialize a shared pointer like this? I think the default constructor is empty

torch_glow/src/PyTorchFileLoader.cpp
putivsky: Yeah, I have thought about moving pass into "common" file, need an agreement from code owner first.

putivsky: Moved.

torch_glow/src/FusingOptimizer.cpp
```diff
@@ -89,4 +92,16 @@ graph(%input, %weight):
   mm_to_linear.runOnGraph(graph);
 }
 
+void glowCustomFuse(std::shared_ptr<torch::jit::Graph> &g,
```
jackm321: We should eventually split fusions here into different files, can this go into PyTorchCommon.cpp?

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,182 @@
...
+        getFusionSymbol(),
+        [](const torch::jit::Node *node) {
+          return [node](torch::jit::Stack &stack) {
+            // Get JIT Graph.
+            auto graph = node->g(at::attr::Subgraph);
+            auto err = loadJitGraphToGlowFunction(
+                stack, *graph, *localFusionInfo.function,
+                *localFusionInfo.inputPlaceholders,
+                *localFusionInfo.outputPlaceholders);
+            if (static_cast<bool>(err)) {
```
jackm321: glow errToBool

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,182 @@
...
+  at::NoGradGuard guard;
+
+  // Set thread local pointer, not null values activate custom fusion pass.
+  localFusionInfo.set(&F, &inputPlaceholders, &outputPlaceholders);
+  auto err = evaluateModuleGraph(module, inputs);
+  localFusionInfo.reset();
+
+  RETURN_IF_ERR(err);
+
+  if (sanityCheck) {
```
jackm321: could honestly be in another function but not a big deal

torch_glow/src/PyTorchFileLoader.h
```diff
@@ -26,10 +27,20 @@ namespace glow {
...
-                    std::shared_ptr<torch::jit::script::Module> &module,
-                    llvm::Error *errPtr);
+  /// \returns error if any.
+  static llvm::Error
+  loadPyTorchModel(const std::string &fileName,
+                   std::shared_ptr<torch::jit::script::Module> &module);
+
+  /// Takes a model file \p fileName, loads optimized graph into Glow Function
+  /// \p F and input \p inputPlaceholders, output \p outputPlaceholders
+  /// placeholders, \returns error if any.
```
jackm321: Can you add some comments about the thread_local stuff here. How should this be used to take advantage of that? For example, "call this once per inference thread for a model" or something like that

torch_glow/src/PyTorchLoaderTest.cpp
```diff
@@ -16,149 +16,45 @@
...
+  llvm::Error err = glow::PyTorchFileLoader::loadPyTorchGraph(
+      fileName, vec, *F, inputPlaceholders, outputPlaceholders);
+
+  if (static_cast<bool>(err)) {
+    LOG(ERROR) << llvm::toString(std::move(err)) << "\n"; // size and reshape.
+  }
+}
+} // namespace
+
+int main(int /*argc*/, char ** /*argv*/) {
```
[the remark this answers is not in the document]
putivsky: I wan't able to build gtest under github/CMakefile.

torch_glow/src/PyTorchFileLoader.cpp
putivsky: I don't want to output error, instead just to throw exception.

torch_glow/src/PyTorchLoaderTest.cpp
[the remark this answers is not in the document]
jackm321: Hmm ok well all of our other c++ tests use gtest with cmake, I think we should continue using it here

torch_glow/src/PyTorchFileLoader.cpp
jackm321: I see ok

torch_glow/src/PyTorchLoaderTest.cpp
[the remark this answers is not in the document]
putivsky: I have meant:
python setup.py test --run_cmake
failed.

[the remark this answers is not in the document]
jackm321: What's the error?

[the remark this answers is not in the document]
putivsky: Something like, cannot find file: #include "gtest/gtest.h"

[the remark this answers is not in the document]
jackm321: And you included gtest in target_link_libraries?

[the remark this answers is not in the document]
putivsky: yes, I have added gtest to CMakefile.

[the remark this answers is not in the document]
jackm321: 😞 oh well let's move no as is then.
It would be great if this was like the rest of our tests. Integration with pytorch jit via C++  is important and we should add some more tests here eventaully, we can follow up on trying to get gtest working then

torch_glow/src/PyTorchLoaderTest.cpp
```diff
@@ -16,151 +16,44 @@
...
-    auto ptT = at::empty(sizes);
-    auto var = torch::autograd::make_variable(ptT);
-    // Add to stack output parameter.
-    stack.push_back(at::IValue(var));
-  }
-}
-} // namespace
-
-TEST(PyTorchModelLoader, Model) {
+void pyTorchModelLoaderModelTest() {
```
jackm321: Can you move this to the tests directory

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -53,7 +53,7 @@ class PyTorchModelLoader {
         llvm::Error (PyTorchModelLoader::*)(const torch::jit::Node *);
 
     /// Symbols (as strings) that this mapping value is applicable to.
-    const std::vector<const char *> symbols;
+    std::vector<const char *> symbols;
```
jackm321: why this change?

torch_glow/src/PyTorchLoaderTest.cpp
jackm321: doxygen

torch_glow/src/PyTorchFileLoader.h
```diff
@@ -17,19 +17,35 @@
...
+  /// Method is thread safe, internally it uses local thread structures for
+  /// executing custom fusion pass, registered globally. No other passes or
+  /// other treads calling this method will be affected.
+  static llvm::Error loadPyTorchGraph(
```
jackm321: doxygen like for example "\p sanityCheck ..."

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+  auto err = evaluateModuleGraph(module, inputs);
+  localFusionInfo.reset();
+
+  RETURN_IF_ERR(err);
+
+  return sanityCheck ? performSanityCheck() : llvm::Error::success();
+}
+
+/*static*/
+llvm::Error PyTorchFileLoader::performSanityCheck() {
```
jackm321: thanks for moving this

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+  // Sanity check, after "fusionSymbol" node, not other nodes should exist.
```
jackm321: move this comment above the function

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+/*static*/
+llvm::Error PyTorchFileLoader::loadPyTorchGraph(
+    const std::string &fileName, std::vector<torch::jit::IValue> &inputs,
+    glow::Function &F, std::vector<glow::Placeholder *> &inputPlaceholders,
+    std::vector<glow::Placeholder *> &outputPlaceholders,
+    const PyTorchLoaderSettings &settings, bool sanityCheck) {
+  // Register custom pass once.
+  static RegisterCustomFusionPass fuser;
+
+  // Convert PyTorch model into JIT Module.
```
jackm321: doxygen

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+  std::shared_ptr<torch::jit::script::Module> module;
+  RETURN_IF_ERR(PyTorchFileLoader::loadPyTorchModel(fileName, module));
+
+  // Disable gradient nodes generation.
+  at::NoGradGuard guard;
+
+  // Set thread local pointer, not null values activate custom fusion pass.
```
jackm321: doxygen

torch_glow/src/PyTorchCommon.h
```diff
@@ -40,6 +40,11 @@ PyTorchLoaderSettings &getPyTorchLoaderSettings();
 /// \returns the PyTorch symbol to be used for the PyTorch node which represents
 /// the subgraph that Glow will compile and run.
 const c10::Symbol &getGlowSymbol();
+
+/// Executes custom fuse pass for the given \p graph and \p fuseSymbol.
+void glowCustomFuse(std::shared_ptr<torch::jit::Graph> &graph,
```
jackm321: where is the thread_local stuff documented?

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+    inputPlaceholders = in;
+    outputPlaceholders = out;
+    settings = s;
+  }
+
+  void reset() { set(nullptr, nullptr, nullptr, nullptr); }
+};
+
+/// Meyer's singleton for static fusion symbol.
+static at::Symbol getFusionSymbol() {
```
jackm321: why do we need another symbol?

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -15,30 +15,189 @@
...
+llvm::Error PyTorchFileLoader::performSanityCheck() {
+  // Sanity check, after "fusionSymbol" node, not other nodes should exist.
+  std::shared_ptr<torch::jit::Graph> subgraph =
+      torch::jit::lastExecutedOptimizedGraph();
+  size_t fusedNodes = 0, missedNodes = 0;
+  const auto symbol = getFusionSymbol();
+  for (const auto &node : subgraph->nodes()) {
+    if (node->kind() == symbol) {
+      ++fusedNodes;
+    } else if (fusedNodes) {
```
jackm321: This will only work if the non-fused nodes are after the fusion group. Is this the intention? if so, why?

torch_glow/src/PyTorchModelLoader.h
putivsky: By design initialization of the members must be done at construction time, this is self-check making sure no one can change members after initialization.

torch_glow/src/PyTorchCommon.h
putivsky: Nowhere, this is hidden from public implementation details, caller shoudn't even know about all this local thread stuff. What caller needs to know is the ability to invoke public methods of PyTorchFileLoader class in multiple threads.

torch_glow/src/PyTorchFileLoader.cpp
putivsky: Yes, because we a registering different operators and correspondently different fuse passes.

torch_glow/src/PyTorchFileLoader.cpp
putivsky: Normally - ALL operators must be fused into one graph node. If you take a look at graph it contains from the beginning the original graph nodes and then fused node, even all nodes got fused.

torch_glow/src/PyTorchFileLoader.cpp
putivsky: can you explain, what should be changed to comply with doxygen style?.

torch_glow/src/PyTorchCommon.h
jackm321: discussed offline, makes sense (I think)

torch_glow/src/PyTorchFileLoader.cpp
jackm321: Oh whoa, that seems super weird. I think I see why that's happening, but why is that desirable instead of deleting the PT graph nodes once they've been fused into Glow?

torch_glow/setup.py
```diff
@@ -124,6 +125,7 @@ def _run_cmake(self):
                 # PyTorch cmake args
                 '-DPYTORCH_DIR={}'.format(
                     os.path.dirname(os.path.realpath(torch.__file__))),
+                '-DTORCH_GLOW={}'.format(FILE_DIR),
```
jackm321: What is this?

putivsky: Normally we should use include paths relative to "glow" folder, otherwise we would need (as in this case) path to the include file from different folder. Because I have moved test file into a separate folder, test makefile needs include path to other torch_glow files.

PR #3393, opened 2019-08-07
Update ExecutionEngine.h comments

gcatron:
Summary: Update ExecutionEngine.h comments

Documentation: This updates the EE's comments 

Progress on #3239 

Test Plan: NFC

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -32,19 +32,19 @@
 
 namespace glow {
 
-/// This is the ExecutionEngine. It owns the Graph, the backend, and the
-/// compiled function.  The Graph, etc in this class are defined as pointers, in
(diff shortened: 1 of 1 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -96,7 +96,8 @@ class ExecutionEngine final {
   /// without resetting the backend.
   void compile(CompilationContext &cctx);
 
-  /// A convenience function for the most common type of compile.
+  /// A convenience function for the most common type of compile. Can only be
```
rdzhabarov: can you also update getModule() comment. It's not safe to call it after graph has been compiled.

gcatron: Updated the comment to call out that elements of the module may no longer exist. The module itself is safe to access after compile though.

PR #3394, opened 2019-08-07
[DeviceManager] Make creation of device managers more generic.

rdzhabarov:
Summary:
One more step forward decoupling backend specific code from core.
* Migrate device manager to use backend auto discovery mechanism.
* Refactored Habana code to remove HabanaDeviceManager dependency on HabanaBackend.

Documentation:
doxygen

Test Plan:
CI


nickgg:
This is cool, but I think it's worth sanity checking that we're not making it too complicated by having a second registration system.

Is there any reason we can't just add a `getDeviceManager()` call to `Backend` and piggyback on Backend registration?

rdzhabarov:
> Is there any reason we can't just add a getDeviceManager() call to Backend and piggyback on Backend registration?

@nickgg This is a good question. I think it just boils down to whether there is a use case for device manager to exist without
a backend and whether device manager needs to know anything about the backend itself.
I do not think there is a compelling reason for these to exist, but let me know if it's possible.
In the meantime I'll give it a try to migrate device manager creation to the backend.

nickgg:
Yeah I dont think we have a use case for that right now, certainly inside the runtime we have both.

rdzhabarov:
@bertmaher if you can check this PR on monday. Majority of the habana changes (apart from creatingDeviceManager from the HabanaBackend) is moving code around from the Habana.cpp to HabanaDeviceManager and HabanaUtils to make sure there is no dependency of HabanaDeviceManager on Habana backend.

PR #3397, opened 2019-08-07
Fix addmm test & FusingOptimizer

zrphercule:
We did not have fusing of aten::mm in our previous version, which makes addmm test doesnt work in our newest version of pytorch. here we add it in and fixed the operator creating process.

### DIFF
--- a/torch_glow/src/FusingOptimizer.cpp
+++ b/torch_glow/src/FusingOptimizer.cpp
@@ -35,6 +35,12 @@ graph(%input, %weight, %bias, %4):
   %output = aten::matmul(%input, %weight_t)
   %res = aten::add_(%output, %bias, %4)
   return (%res))IR";
+  std::string mm_add_pattern = R"IR(
+graph(%input, %weight, %bias, %4):
+  %weight_t = aten::t(%weight)
+  %output = aten::mm(%input, %weight_t)
+  %res = aten::add_(%output, %bias, %4)
+  return (%res))IR";
   std::string fused_linear = R"IR(
 graph(%input, %weight, %bias, %4):
   %res = aten::linear(%input, %weight, %bias)
@@ -44,6 +50,11 @@ graph(%input, %weight, %bias, %4):
 graph(%input, %weight):
   %weight_t = aten::t(%weight)
   %output = aten::matmul(%input, %weight_t)
+  return (%output))IR";
+  std::string mm_pattern = R"IR(
+graph(%input, %weight):
+  %weight_t = aten::t(%weight)
+  %output = aten::mm(%input, %weight_t)
   return (%output))IR";
   std::string fused_linear_bias_none = R"IR(
 graph(%input, %weight):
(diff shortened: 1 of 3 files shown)


torch_glow/src/FusingOptimizer.cpp
```diff
@@ -35,6 +35,12 @@ graph(%input, %weight, %bias, %4):
   %output = aten::matmul(%input, %weight_t)
   %res = aten::add_(%output, %bias, %4)
   return (%res))IR";
+  std::string mm_add_pattern = R"IR(
+graph(%input, %weight, %bias, %4):
```
yinghai: weird indent?

zrphercule: Actually not, it is hardcoded IR pattern, and can only use this indent.

jackm321:
Thanks for fixing it!
I'm confused why PyTorch CI passed, there is an assert in jitVsGlow that requires the expected nodes to be lowered to be specified.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -684,23 +684,29 @@ llvm::Error PyTorchModelLoader::loadLinear(const torch::jit::Node *ptNode) {
...
+  auto *mul = F_.createMatMul("linear_mul", input, weights_t->getResult());
 
   if (hasGlowNodeValue(inputs[LinearInputs::bias])) {
     glow::NodeValue bias;
     ASSIGN_VALUE_OR_RETURN_ERR(bias,
                                getGlowNodeValue(inputs[LinearInputs::bias]));
-    return addGlowNodeValue(outputs[0], F_.createAdd("linear_add", mul, bias));
+    return addGlowNodeValue(
+        outputs[0],
+        (F_.createAdd("linear_add", mul->getResult(), bias))->getResult());
```
jackm321: might be cleaneer to just put the createAdd above

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -684,23 +684,29 @@ llvm::Error PyTorchModelLoader::loadLinear(const torch::jit::Node *ptNode) {
...
   glow::NodeValue weights;
   ASSIGN_VALUE_OR_RETURN_ERR(weights,
                              getGlowNodeValue(inputs[LinearInputs::weights]));
 
-  auto *mul = F_.createMatMul("linear_mul", input, weights);
+  // Currently in Pytorch => Glow, we only translate t()+addmm(), t()+mm() and
+  // t()+matmul() to linear, Which means we need a transpose anyway, and also
+  // since addmm and mm only works for 2-D, we dont need to consider other
+  // transpose implementation.
+  auto *weights_t = F_.createTranspose("weights_t", weights, {1, 0});
```
jackm321: can just put this in weights

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -684,23 +684,29 @@ llvm::Error PyTorchModelLoader::loadLinear(const torch::jit::Node *ptNode) {
...
-
   glow::NodeValue input;
   ASSIGN_VALUE_OR_RETURN_ERR(input,
                              getGlowNodeValue(inputs[LinearInputs::input]));
   glow::NodeValue weights;
   ASSIGN_VALUE_OR_RETURN_ERR(weights,
                              getGlowNodeValue(inputs[LinearInputs::weights]));
 
-  auto *mul = F_.createMatMul("linear_mul", input, weights);
+  // Currently in Pytorch => Glow, we only translate t()+addmm(), t()+mm() and
```
jackm321: This is some pretty baked-in logic, we'll have to revisit this later probably.

zrphercule: Exactly, that's why I left some comments here...

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: ok

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: what do you mean?

jackm321: no need to define a new weights_t

jackm321:
Please make sure to update jitVsGlow in addmm_test.py before landing

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: oh... I see, anyway no big deal hahaha

zrphercule:
> Please make sure to update jitVsGlow in addmm_test.py before landing

The reason why jitVsGlow passed at first is because when we first added the test, addmm is still "aten::addmm", but now somehow it is "aten::mm + aten::add" which caused this bug. Also you remember you fixed me a bug which makes this test doesnt really run the addmm op right? And that is why when we update pytorch version here, we didnt find out it was wrong...

Guess there is no need to change jitVsGlow?...

PR #3403, opened 2019-08-08
Migrate away from CustomFuseGraph

zrphercule:
This is basiclly the glow version of https://github.com/pytorch/tvm/pull/72
Will not use PyTorch's customFuseNode anymore.

Will add comment indicate the copied code and fix the lint once finished.
Please dont give detailed review until WIP is removed, but feel free to leave any big-scope opinion.


torch_glow/src/FusingOptimizer.cpp
```diff
@@ -68,4 +73,106 @@ graph(%input, %weight):
   matmul_to_linear.runOnGraph(graph);
 }
 
+const torch::jit::Symbol& getGlowSymbol() {
```
jackm321: use the symbol in PyTorchCommon

torch_glow/src/FusingOptimizer.cpp
```diff
@@ -68,4 +73,106 @@ graph(%input, %weight):
...
+  auto node_inputs = sortReverseTopological(node->inputs(), block);
+  for (auto input : node_inputs) {
+    if (auto group = tryMerge(node, input->node(), aliasDb, fn)) {
+	  return group.value()->reverseIterator();
+	}
+  }
+  return ++node->reverseIterator();
+}
+
+void GlowCustomFuse(std::shared_ptr<Graph> graph, isSupportFunc fn, Symbol kind, size_t arg_limit) {
```
jackm321: Let's put this new stuff in a different file for clarity

torch_glow/src/FusingOptimizer.cpp
zrphercule: yeah... Just realized there is one.

torch_glow/src/FusingOptimizer.cpp
zrphercule: Okay I will do it once it works in one file.

torch_glow/src/FusingOptimizer.cpp
```diff
@@ -68,4 +73,106 @@ graph(%input, %weight):
...
+  REQ(consumer->kind() == getGlowSymbol() || canMerge(consumer, aliasDb));
+
+  // Alias checks
+  // Requirement:
+  // - moveAfterTopologicallyValid(consumer, producer)
+  // - One of:
+  //   1) Both are in-place ops
+  //   2) Consumer is in-place, producer !hasInputWriters
+  //   3) Producer is in-place, consumer !hasOutputWriters
+  REQ(aliasDb.moveAfterTopologicallyValid(consumer, producer));
```
jackm321: it looks like this actually moves the consumer after producer, is there a reason we want to do this and not just see if it's possible with `couldMoveAfterTopologically`

jackm321: @bwasti

zrphercule: Guess there is not

zrphercule:
migrating to a seperate file now...

torch_glow/src/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ link_directories(${PYTORCH_DIR}/lib)
 
 add_library(PyTorchModelLoader
                      PyTorchCommon.cpp
-                     FusingOptimizer.cpp
+					 GlowFuser.cpp
```
yinghai: weird indent?

torch_glow/tests/utils.py
```diff
@@ -16,10 +16,12 @@ def jitVsGlow(f, *inputs, expected_fused_ops):
...
-def jitVsGlow_(f_torch, f_glow, *inputs, expected_fused_ops):
-    assert (
-        expected_fused_ops is not None and len(expected_fused_ops) > 0
-    ), "Must pass non-empty list of ops that are expected to be fused"
+def jitVsGlow_(f_torch, f_glow, *inputs, expected_fused_ops=None):
+
+    accept_all_ops = False
+    if (expected_fused_ops == None):
+        expected_fused_ops = []
+        accept_all_ops = True
```
jackm321: I don't think this is a good idea. The point of expected_fused_ops is to check that the things that should be getting fused are getting fused. Why do we need to have this wildcard?

torch_glow/src/GlowFuser.h
```diff
@@ -18,10 +18,17 @@
 #define GLOW_TORCH_GLOW_SRC_FUSINGOPTIMIZER_H
```
jackm321: please change this too since the file was moved

jackm321:
Looking great @zrphercule! Just a couple of comments

torch_glow/tests/utils.py
zrphercule: It is because we want to use jitVsGlow not only here in the unit test, but also in, like, testing xray (actually I did have a script testing xray model locally using jiVsGlow). we cannot indicate all ops in a big model like xray easily.
My thinking is, this functionality is like an extra. If we want to check it, then good we will have it to be checked. Else we just dont check it.

jackm321: Ok well I guess my two thoughts are that
1. I'd like to make sure all the operator tests use this so having operator checking be on by default (as opposed to this where it's basically off by default) would be preferred (maybe we can pass an extra bool flag to disable this?)
2. Shouldn't we be checking the ops in the xray model anyways? We want to make sure we're running what we expect to be running.

zrphercule: I agree with you for both two opinions. So our decision is:
1. A seperate by_default=False param to control if all ops should be accepted.
2. Once we have an official bigger model unit tests in our code base, we should also have a list of expected ops ( of course future operator unit tests should have this as well)

jackm321: That sounds good to me

PR #3404, opened 2019-08-08
[Partitioner] DAG validation

beicy:
Summary:
This PR checks if a partition is a DAG.   This validation is mainly used in user-defined partition so far. But it can be added to anyplace where a partition DAG is already created. 

Documentation:

[Optional Fixes #issue]

Test Plan:
unit test added.

### DIFF
--- a/lib/Partitioner/PartitionerValidation.cpp
+++ b/lib/Partitioner/PartitionerValidation.cpp
@@ -58,4 +58,46 @@ memoryUsageValidation(const NodeToFunctionMap &partitions,
   }
   return llvm::Error::success();
 }
+
+/// \returns true if \p node contains no cycles. \p path contains the nodes in a
+/// path, and \p visited contains the nodes checked before.
+static bool isDAG(DAGNode *node, llvm::SmallSet<DAGNode *, 10> &path,
+                  llvm::SmallSet<DAGNode *, 10> &visited) {
+  if (!visited.count(node)) {
+    path.insert(node);
+    visited.insert(node);
+    for (size_t i = 0; i < node->children.size(); i++) {
+      auto child = node->children[i];
+      if (path.count(child)) {
+        // Cycle found.
+        return false;
+      }
+      if (!isDAG(child, path, visited)) {
+        return false;
+      }
+    }
+  }
+  if (path.count(node)) {
+    path.erase(node);
+  }
+  return true;
+}
+
+llvm::Error dagValidation(const DAG &dag) {
+  auto *root = dag.root.get();
+  llvm::SmallSet<DAGNode *, 10> path;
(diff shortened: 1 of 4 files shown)


lib/Partitioner/PartitionerValidation.cpp
```diff
@@ -58,4 +58,36 @@ memoryUsageValidation(const NodeToFunctionMap &partitions,
   }
   return llvm::Error::success();
 }
+
+/// \returns true is \p node contains no cycles.
+static bool isDAG(DAGNode *node, llvm::SmallSet<DAGNode *, 10> &used) {
```
gcatron: I think this approach may lead to a isDAG being called on children nodes more than once. e.g. A->C B->C isDAG(C) will be called twice. I think this is fine since the number of partitions will always be small. Could you add a comment stating that?

lib/Partitioner/PartitionerValidation.cpp
```diff
@@ -58,4 +58,36 @@ memoryUsageValidation(const NodeToFunctionMap &partitions,
...
+llvm::Error dagValidation(const DAG &dag) {
+  auto *root = dag.root.get();
+  llvm::SmallSet<DAGNode *, 10> used;
+  used.insert(root);
+
+  // For the first condition: root->children.size() > 0 -- When a dag is
+  // created, its root is a dummy node and other DAGNode without parents will be
+  // linked to this root. Therefore, root without any child means that each of
+  // the rest of DAGNodes has at least one parent. That is, a cycle exists.
+  RETURN_ERR_IF_NOT((root->children.size() > 0 && isDAG(root, used)),
```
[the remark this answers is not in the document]
beicy: "If the root has no children it's an empty dag right?"
-- No.  As the comment says, if the root has no children,  there is a cycle (the dag won't be empty based on how we create it).  E.g : we have 2 partitions p1 and p2. Now p1->p2, and p2->p1.  When we create a dag, the final step is: we found the DAGNodes without parents, and link them to root. But in this case p1's parent is p2, and p2's parent is p1. Therefore, no node is linked to root. So, root's children is empty.

lib/Partitioner/PartitionerValidation.cpp
beicy: Improved the function to avoid visiting duplication.

lib/Partitioner/PartitionerValidation.cpp
```diff
@@ -58,4 +58,46 @@ memoryUsageValidation(const NodeToFunctionMap &partitions,
   }
   return llvm::Error::success();
 }
+
+/// \returns true is \p node contains no cycles. \p path contains the nodes in a
```
jfix71: typo: true if

PR #3406, opened 2019-08-09
OCLFunction multi-device support

gcatron:
Summary: This PR adds proper multi-device support to OCL function. It removes member variables that were being used for per run state and instead uses the values from CLDeviceBindings directly.

Documentation:

Test Plan: resnet-runtime with 3 devices and opencl now works. ninja check passes

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -153,22 +153,9 @@ static std::string getKernelName(const char *baseName, ElemKind elemTy) {
 cl_kernel OpenCLFunction::createKernel(const std::string &name,
                                        cl_program program) {
   cl_int err = CL_SUCCESS;
-  cl_kernel kernel = nullptr;
-  if (program) {
-    cl_kernel kernel = clCreateKernel(program, name.c_str(), &err);
-    CHECK(kernel) << "clCreateKernel Failed.";
-    CHECK_EQ(err, CL_SUCCESS) << "clCreateKernel Failed.";
-    return kernel;
-  }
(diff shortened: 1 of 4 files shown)


lib/Backends/OpenCL/OpenCL.h
```diff
@@ -134,28 +131,33 @@ class OpenCLFunction final : public CompiledFunction {
...
-  cl_mem allocDeviceBuffer(uint64_t size);
+  cl_mem allocDeviceBuffer(uint64_t size, cl_context clContext);
   /// Frees a device buffer.
   void freeDeviceBuffer(cl_mem buf);
 
   /// Create kernel with a given \p name from a \p program.
   /// If \p program is nullptr, try to find the kernel with a given \p name
   /// in any of compiled programs.
-  cl_kernel createKernel(const std::string &name, cl_program program = nullptr);
+  cl_kernel createKernel(const std::string &name, cl_program program);
```
nickgg: Now that this can no longer be null, should we check its not in the implementation?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -347,6 +336,8 @@ void OpenCLFunction::enqueueKernel(llvm::StringRef name,
 
 void OpenCLFunction::executeNCHWConvolution(
     const ConvolutionInst *CC, ExecutionContext *executionContext) {
+  auto devBindings = static_cast<runtime::OpenCLDeviceBindings *>(
+      executionContext->getDeviceBindings());
```
nickgg: Is deviceBindings guaranteed to be set? Should we DCHECK this?

PR #3408, opened 2019-08-09
[tests] add a simpler test of Convolutions

nickgg:
Summary: Was debugging a Convolution implementation and needed a minimal Convolutions test I could work out in my head for comparison, figured it might be a useful for future backends. It's a simple 3x3 input, output and filter.

Documentation: n/a

Test Plan: ran tests

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -3108,6 +3108,40 @@ TEST_P(OperatorTest, IntBatchedArith) {
   EXPECT_NEAR(H.at({0, 2, 2}), 9.3, allowedError);
 }
 
+TEST_P(OperatorTest, convTest) {
+  ENABLED_BACKENDS(Interpreter, CPU, OpenCL, Habana);
+  auto *input =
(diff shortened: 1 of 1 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -3108,6 +3108,42 @@ TEST_P(OperatorTest, IntBatchedArith) {
...
+  bindings_.allocate(zeroBias)->zero();
+
+  auto outTy = mod_.uniqueType(ElemKind::FloatTy, {1, 3, 3, 1});
+
+  ConvolutionNode *CN =
+      F_->createConv("Conv", input, filter, zeroBias, outTy, 3, 1, 1, 1);
+  SaveNode *S = F_->createSave("save", CN);
+  bindings_.allocate(S->getPlaceholder());
+
+  ::glow::convertPlaceholdersToConstants(F_, bindings_,
```
jackm321: What is the point of this?
Why not just create constants instead of placeholders?

nickgg: Honestly, it was in the test I adapted and I didn't think too much about it. I don't think its important so can just remove it.

PR #3409, opened 2019-08-12
[Partitioner] Partitioner refactoring - final

beicy:
Summary:
This is the last main parts of Partitioner refactoring:
1) Rewrote the load-balanced partition as a separate partition flow instead of an optimization pass since this one only uses the number of partitions generated by auto partition and overwrites the partition results.
2) Added unittest for load-balanced partition.
3) Abstract some common code as functions.

Documentation:

[Optional Fixes #issue]
#2298 
Test Plan:
added unnitest, ninja test, ./tests/images/run.sh

### DIFF
--- a/include/glow/Partitioner/Partitioner.h
+++ b/include/glow/Partitioner/Partitioner.h
@@ -33,6 +33,9 @@ class Partitioner final : public PartitionerBase {
   /// has the largest memory size.
   Function *F_;
 
+  /// True if there are more than 1 type of backends.
+  bool multiBackendNames_;
+
   /// The cost model related to device.
   std::vector<DeviceInfo> deviceInfo_;
 
@@ -72,6 +75,13 @@ class Partitioner final : public PartitionerBase {
   /// update the memSize.
   static Function *selectRepFunc(Module *parent, uint64_t &memSize);
 
+  /// Initialization. Called in class constructor.
+  void init();
+
+  /// Verify the generated functions in module, and dump partition logs from \p
+  /// partitions and \p mapping.
+  void finalize(const DAGListTy &partitions, const NodeToFunctionMap &mapping);
+
   /// After getting the initial partitions, adjust the partitions to minimize
   /// communication and computation cost.
   void partitionsAdjust(NodeToFunctionMap &partitions,
@@ -96,13 +106,6 @@ class Partitioner final : public PartitionerBase {
   backendBasedPartition(FunctionToBackendNameMap &funcToBackend, Function *F,
                         std::vector<Backend *> &backends,
                         CompilationContext &cctx);
-
-  /// Performs a load balancing optimization pass to optimize for load
-  /// balance in addition to respecting memory constraints.
-  llvm::Error loadBalancedPartitioning(Function *F, DeviceIDTy numDevices,
-                                       uint64_t availableMemory,
-                                       llvm::StringRef backendName,
-                                       NodeToFunctionMap &mapping);
 
   /// If there is no need to do any partition, just generate the DAGNode based
   /// on current functions in this module for backend \p backendName found in \p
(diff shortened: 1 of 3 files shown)


include/glow/Partitioner/Partitioner.h
```diff
@@ -161,6 +164,11 @@ class Partitioner final : public PartitionerBase {
   /// for function optimization. \returns the partition result or an error.
   llvm::Expected<DAGListTy> heterogeneousPartition(CompilationContext &cctx);
 
+  /// Performs a load balancing optimization pass to optimize for load ///
```
rdzhabarov: fix comment

rdzhabarov: can you add more details about "load balance", what exactly does that mean in the context of partitioner.

lib/Partitioner/Partitioner.cpp
```diff
@@ -378,11 +411,37 @@ llvm::Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
...
 
-llvm::Error Partitioner::loadBalancedPartitioning(Function *F,
-                                                  DeviceIDTy numDevices,
-                                                  uint64_t availableMemory,
-                                                  llvm::StringRef backendName,
-                                                  NodeToFunctionMap &mapping) {
+llvm::Expected<DAGListTy>
+Partitioner::loadBalancedPartition(CompilationContext &cctx,
+                                   size_t numDevices) {
+  DCHECK(module_->getFunctions().size() == 1)
```
rdzhabarov: as talked offline, similarly to the rest of the cases when we return error, we should return error here as well instead of debug level checking.

include/glow/Partitioner/Partitioner.h
```diff
@@ -161,6 +164,12 @@ class Partitioner final : public PartitionerBase {
   /// for function optimization. \returns the partition result or an error.
   llvm::Expected<DAGListTy> heterogeneousPartition(CompilationContext &cctx);
 
+  /// This partition approach is an experimental one. It tries to balance the
+  /// workloads of each accelerator/device in addition to respecting memory
+  /// constraints.
+  llvm::Expected<DAGListTy> loadBalancedPartition(CompilationContext &cctx,
+                                                  size_t numDevices = 0);
```
rdzhabarov: can you document numDevices. It seems that it gets overwritten under certain conditions inside the LBPartition.

lib/Partitioner/Partitioner.cpp
```diff
@@ -378,11 +411,38 @@ llvm::Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
...
-                                                  uint64_t availableMemory,
-                                                  llvm::StringRef backendName,
-                                                  NodeToFunctionMap &mapping) {
+llvm::Expected<DAGListTy>
+Partitioner::loadBalancedPartition(CompilationContext &cctx,
+                                   size_t numDevices) {
+  RETURN_ERR_IF_NOT(
+      module_->getFunctions().size() == 1,
+      "Invalid number of functions in a module. For load-balanced partition "
+      "flow, the module can only contain 1 function");
```
rdzhabarov: could you add actual number of function in a module to the error message as well.

lib/Partitioner/Partitioner.cpp
```diff
@@ -378,11 +411,38 @@ llvm::Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
...
+  if (multiBackendNames_) {
+    VLOG(1) << "For multi backend types, load-balanced partition can't be "
+               "applied. Call heterogeneous partition instead.";
+    return heterogeneousPartition(cctx);
+  }
+  F_ = selectRepFunc(module_, memSize_);
```
rdzhabarov: nit: since it's guaranteed to be a single function now, should we just get function from the module directly and remove selectRepFunc call?

beicy: "since it's guaranteed to be a single function now"
-- In fact, we still allow multi-functions in a module as long as none of the function needs partition. That is, for each function, we just create DAG for each function. 

"should we just get function from the module directly and remove selectRepFunc call?"
-- For heterogeneous partition, I think selectRepFunc call should be kept due to the above reason. But for other partition flows, "selectRepFunc" is not necessary.

PR #3411, opened 2019-08-13
Assorted memory-related fixes for HostManager, EE, RecSys

jfix71:
4 commits here:
- When calling `ExecutionEngine::clear()`, reset the Module
- Add device memory to the EE constructor so we don't need to reset
- Use `HostManager::removeNetwork()` from `HostManager::clearHost()`
- Refactor RecSys to make sure EEs are cleared once done and we save only the result tensor to compare against.

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -47,7 +47,7 @@ class ExecutionEngine final {
   /// this resets the ExecutionEngine.
   std::string backendName_ = "";
 
-  /// Size of device memory, if 0 device default is used.
+  /// Size of device memory in bytes, if 0 device default is used.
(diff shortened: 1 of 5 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -61,7 +61,8 @@ class ExecutionEngine final {
   void runInternal(ExecutionContext &context, llvm::StringRef name);
 
 public:
-  ExecutionEngine(llvm::StringRef backend = "Interpreter");
+  ExecutionEngine(llvm::StringRef backend = "Interpreter",
+                  uint64_t deviceMemory = 0);
```
rdzhabarov: could you update comment on device memory whether it's in KB or bytes.

PR #3413, opened 2019-08-13
Add check that IR instruction defs come before uses.

chadrosier:
Summary:
Nice check to have when modifying IR generation.


rdzhabarov:
Thanks @chadrosier, landing this.

PR #3414, opened 2019-08-13
[PyTorch] Cache Glow compilations

jackm321:
Summary:
Cache compiled Glow functions for each PyTorch subgraph.
Does not cache on input types yet.

Documentation:
Doxygen

Test Plan:
CI
Added prints to `loadGraphImpl` and checked that for each unit test, first time it is run it creates a new Glow function and caches it, second time it uses that cached function.

### DIFF
--- a/torch_glow/src/CachingGraphRunner.cpp
+++ b/torch_glow/src/CachingGraphRunner.cpp
@@ -20,41 +20,87 @@
 
 #include "glow/Support/Support.h"
 
+#include <torch/csrc/jit/argument_spec.h>
+#include <torch/csrc/utils/hash.h>
+
 namespace glow {
 
-llvm::Error CachingGraphRunner::runGraph(const torch::jit::Node *node,
-                                         torch::jit::Stack &stack) {
-  // If this is the first time this subgraph has been run then create a new
-  // GraphInfo object to store information about it.
-
+namespace {
+/// Returns a hash that represents a given PT subgraph represented by the PT
+/// node \p node and a set of tensor shape inputs to that subgraph from the \p
+/// stack.
+size_t getGraphHash(const torch::jit::Node *node,
+                    const torch::jit::Stack &stack) {
   const std::shared_ptr<torch::jit::Graph> graph = node->g(at::attr::Subgraph);
   const auto &graphInputs = graph->inputs();
-  const auto numInputs = graphInputs.size();
-  auto inputs = torch::jit::last(stack, numInputs);
-  const char *const functionName = "PyTorchFunction";
+  const auto inputs = torch::jit::last(stack, graphInputs.size());
+  torch::jit::CompleteArgumentSpec spec(/*with_grad*/ false, inputs);
 
-  glow::Function *f = nullptr;
-  // Discard compiled function.
-  executionEngine_.setBackendName(executionEngine_.getBackendName());
-  auto &mod = executionEngine_.getModule();
-  f = mod.createFunction(functionName);
-  std::vector<glow::Placeholder *> inputPlaceholders;
-  std::vector<glow::Placeholder *> outputPlaceholders;
+  return torch::hash_combine(reinterpret_cast<size_t>(node), spec.hashCode());
+}
+
+} // namespace
+
+llvm::Expected<CachingGraphRunner::PerGlowGraphInfo *>
+CachingGraphRunner::loadGraphImpl(const torch::jit::Node *node,
+                                  torch::jit::Stack &stack) {
+  const std::shared_ptr<torch::jit::Graph> graph = node->g(at::attr::Subgraph);
+  RETURN_ERR_IF_NOT(graph, "A subgraph couldn't be found for this node");
+
+  const auto &graphInputs = graph->inputs();
+  auto inputs = torch::jit::last(stack, graphInputs.size());
+
+  size_t hash = getGraphHash(node, stack);
+
+  // If we already have a Glow function compiled for this graph with and the
+  // given inputs then use that.
+  auto it = perGlowGraphInfoMap.find(hash);
+  if (it != perGlowGraphInfoMap.end()) {
+    return it->second.get();
+  }
+
+  auto info = llvm::make_unique<PerGlowGraphInfo>();
+  info->functionName = strFormat("PTFunction%lu", hash);
+  info->hash = hash;
+  info->node = node;
+
+  std::unique_ptr<Module> module = llvm::make_unique<Module>();
+  Function *f = module->createFunction(info->functionName);
 
   RETURN_IF_ERR(PyTorchModelLoader::loadJITGraph(
-      *f, *graph, inputs, inputPlaceholders, outputPlaceholders,
+      *f, *graph, inputs, info->inputPlaceholders, info->outputPlaceholders,
       getPyTorchLoaderSettings()));
 
-  glow::PlaceholderBindings bindings;
-  for (size_t i = 0; i < inputs.size(); ++i) {
-    glow::Placeholder *ph = inputPlaceholders[i];
+  glow::CompilationContext cctx;
+
+  RETURN_IF_ERR(hostManager_->addNetwork(std::move(module), cctx));
+
+  perGlowGraphInfoMap.insert({hash, std::move(info)});
+
+  return perGlowGraphInfoMap[hash].get();
+}
+
+llvm::Error CachingGraphRunner::runGraphImpl(const PerGlowGraphInfo &info,
+                                             torch::jit::Stack &stack) {
+  DCHECK_EQ(getGraphHash(info.node, stack), info.hash)
+      << "Tried to run a graph with incompatible input shapes.";
+
+  size_t numInputs = info.inputPlaceholders.size();
+
+  auto inputs = torch::jit::last(stack, numInputs);
+
+  std::unique_ptr<ExecutionContext> ctx = llvm::make_unique<ExecutionContext>();
+  auto *bindings = ctx->getPlaceholderBindings();
+
+  for (size_t i = 0; i < numInputs; ++i) {
+    glow::Placeholder *ph = info.inputPlaceholders[i];
     glow::TypeRef ty = ph->getType();
     glow::Tensor t(inputs[i].toTensor().data_ptr(), ty);
-    bindings.insert(ph, std::move(t));
+    bindings->insert(ph, std::move(t));
   }
 
   std::vector<at::IValue> outputs;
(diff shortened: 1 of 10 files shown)


jackm321:
@putivsky @yinghai @zrphercule

jackm321:
We can start out by using `CompleteArgumentSpec` but we will want to eventually use hashed `ProfiledTensorType`s plus we will need to do some stuff to hash on values to make sure that the cached glow functions truly represent the input being run.

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
+                                         torch::jit::Stack &stack) {
+  PerGlowGraphInfo *info;
+  ASSIGN_VALUE_OR_RETURN_ERR(info, loadGraphImpl(node, stack));
+  RETURN_IF_ERR(runGraphImpl(*DCHECK_NOTNULL(info), stack));
   return llvm::Error::success();
 }
+
+CachingGraphRunner::CachingGraphRunner() {
+  constexpr size_t numGlowDevices = 1;
+  constexpr const char *glowBackendName = "Interpreter";
```
yinghai: Why do we hardcode this?

torch_glow/src/CachingGraphRunner.h
```diff
@@ -20,18 +20,52 @@
...
+    /// PerGlowGraphInfo.
+    size_t hash;
+
+    /// The PyTorch node containing the subgraph that this PerGlowGraphInfo
+    /// represents.
+    const torch::jit::Node *node;
+  };
+
+  std::unique_ptr<runtime::HostManager> hostManager_;
+  std::unordered_map<size_t, std::unique_ptr<PerGlowGraphInfo>>
```
yinghai: When do we free up `PerGlowGraphInfo` elements?

jackm321: This is a good question, right now we never do (except at shutdown).
There are two possibilities:
1) find a hook for when graph get removed from JIT and remove all associated Glow function then (this would be ideal). In fact I think we could do this by keep a smart ptr to this structure in the Glow fusion op. I will try this as a follow up.
2) create max cache size here and remove the lru when it's hit (not as ideal)

torch_glow/src/CachingGraphRunner.cpp
jackm321: As opposed to a flag or what?

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
+
+  return torch::hash_combine(reinterpret_cast<size_t>(node), spec.hashCode());
+}
+
+} // namespace
 
+llvm::Expected<CachingGraphRunner::PerGlowGraphInfo *>
+CachingGraphRunner::loadGraphImpl(const torch::jit::Node *node,
+                                  torch::jit::Stack &stack) {
   const std::shared_ptr<torch::jit::Graph> graph = node->g(at::attr::Subgraph);
```
putivsky: Can you double check, there is no assert inside g() method?

torch_glow/src/CachingGraphRunner.h
yinghai: Yeah, smart pointer should be good.

torch_glow/src/CachingGraphRunner.cpp
yinghai: Yeah, I think we should be able to test on other backend (CPU, hardware). Flag can be more convenient.

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
+  RETURN_ERR_IF_NOT(graph, "A subgraph couldn't be found for this node");
+
   const auto &graphInputs = graph->inputs();
-  const auto numInputs = graphInputs.size();
-  auto inputs = torch::jit::last(stack, numInputs);
-  const char *const functionName = "PyTorchFunction";
+  auto inputs = torch::jit::last(stack, graphInputs.size());
+
+  size_t hash = getGraphHash(node, stack);
```
putivsky: getGraphHash repeats the same code as:
 
  `const std::shared_ptr<torch::jit::Graph> graph = node->g(at::attr::Subgraph);
   auto inputs = torch::jit::last(stack, graphInputs.size());
   `
  May you can provide **node** & **inputs** as input parameters into getGraphHash function?

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
-  executionEngine_.compile(cctx);
 
-  glow::PlaceholderBindings bindings;
-  for (size_t i = 0; i < inputs.size(); ++i) {
-    glow::Placeholder *ph = inputPlaceholders[i];
+  RETURN_IF_ERR(hostManager_->addNetwork(std::move(module), cctx));
+
+  perGlowGraphInfoMap.insert({hash, std::move(info)});
+
+  return perGlowGraphInfoMap[hash].get();
```
putivsky: Searching again, use the return iterator at line 78.

torch_glow/src/CachingGraphRunner.h
```diff
@@ -20,18 +20,52 @@
...
+  struct PerGlowGraphInfo {
+    std::vector<glow::Placeholder *> inputPlaceholders;
+    std::vector<glow::Placeholder *> outputPlaceholders;
+
+    /// Name of the Glow function maintained by HostManager for this subgraph.
+    std::string functionName;
+
+    /// The unique hash for the subgraph and input shapes represented by this
+    /// PerGlowGraphInfo.
+    size_t hash;
```
putivsky: it's a key in the map, why to store it again?

jackm321:
Actually, looks like `CompleteArgumentSpec` is getting wiped out so I'll change this to use ProfiledTensorType now
https://github.com/pytorch/pytorch/pull/24169
Thanks @Krovatkin for the heads up about this

torch_glow/src/CachingGraphRunner.cpp
putivsky: Even better you can insert dummy info:
  `auto p = perGlowGraphInfoMap.insert({hash, PerGlowGraphInfo()});
   if (!p.second) {
     return p.first->second;
   }

   populate p.first->second with correct values.
    

   `

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
+llvm::Error CachingGraphRunner::runGraphImpl(const PerGlowGraphInfo &info,
+                                             torch::jit::Stack &stack) {
+  DCHECK_EQ(getGraphHash(info.node, stack), info.hash)
+      << "Tried to run a graph with incompatible input shapes.";
+
+  size_t numInputs = info.inputPlaceholders.size();
+
+  auto inputs = torch::jit::last(stack, numInputs);
+
+  std::unique_ptr<ExecutionContext> ctx = llvm::make_unique<ExecutionContext>();
```
putivsky: It will create a bindings for every run, is it possible to cache bindings as well?

torch_glow/src/CachingGraphRunner.cpp
```diff
@@ -20,61 +20,128 @@
...
   }
 
+  return err;
+}
+
+llvm::Error CachingGraphRunner::runGraph(const torch::jit::Node *node,
+                                         torch::jit::Stack &stack) {
+  PerGlowGraphInfo *info;
+  ASSIGN_VALUE_OR_RETURN_ERR(info, loadGraphImpl(node, stack));
+  RETURN_IF_ERR(runGraphImpl(*DCHECK_NOTNULL(info), stack));
```
putivsky: will just
   `return runGraphImpl(*DCHECK_NOTNULL(info), stack);`
work?

torch_glow/src/CachingGraphRunner.cpp
jackm321: Good point, looks like it throws actually

torch_glow/src/CachingGraphRunner.cpp
jackm321: I did it this way because in case of failure there isn't a dummy node floating in perGlowGraphInfoMap that has to be removed, the extra lookup is fine imo since this only happens the first time when we're compiling the glow function which will be much slower anyways

torch_glow/src/CachingGraphRunner.cpp
jackm321: I'm not sure it'll be too much of a win, haven't noticed this being an issue in onnxifi

torch_glow/src/CachingGraphRunner.h
jackm321: To double check it against the argument spec in runGraphImpl, maybe it's not worth it

torch_glow/src/CachingGraphRunner.cpp
putivsky: Place it into PyTorchLoaderSettings struct.

torch_glow/src/CachingGraphRunner.cpp
putivsky: It can be done more elegantly, using llvm::scope_exit

   auto p = perGlowGraphInfoMap.insert({hash, PerGlowGraphInfo()});
   auto guard =  llvm::make_scope_exit([&]() {
     perGlowGraphInfoMap.erase(p.first);
   });
  
  // on failure
  return error;

   // on success
  guard.release();
  return success;

torch_glow/src/CachingGraphRunner.cpp
putivsky: Yeah, input/output tensor usually have reasonable sizes.

yinghai:
@jackm321 Anything blocking this?

jackm321:
@yinghai no, I will do it soon, have just been busy fixing stuff with profiled tensors, helping find issues with topk, working on fusing size + reshape because it's blocking @putivsky, etc

yinghai:
`ci/circleci: PYTORCH ` has been failing. Is it because of changes of PyTorch ProfilingTensor recently?

torch_glow/src/CachingGraphRunner.cpp
jackm321: I don't really prefer that way, I'd rather just not insert into the map until we are sure that it's good

torch_glow/src/CachingGraphRunner.h
jackm321: Added a TODO for this

PR #3416, opened 2019-08-14
Added new priority_queue for ordering queded requests

gcatron:
Summary: This PR adds an optional priority field to runNetwork on HostManager. Also it adds a priority queue. 
When a request is made of the HostManager the following happens:
The request is added to the priorityQueue if the max queue size limit has been reached return an error immediately. 
If there are less active requests than the activeRequestLimit dispatchNextRun() is called.
If the activeLimit has been reached return.

Documentation: Updated in code documentation for new data structure and priority parameter in runNetwork.

Test Plan: Added a new QueueTest to HostManagerTest.

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -23,6 +23,7 @@
 #include <atomic>
 #include <map>
 #include <mutex>
+#include <queue>
 #include <unordered_map>
 #include <vector>
 
@@ -50,6 +51,37 @@ class HostManager final {
     /// safety.
     std::atomic<size_t> refcount{0};
   };
+  /// Container for inference requests waiting in the queue.
+  struct InferRequest {
+    /// Name of the network the requested run is for.
+    std::string networkName;
+
+    /// The execution context for the request.
+    std::unique_ptr<ExecutionContext> context;
+
+    /// The user provided callback to run after execution finishes.
+    ResultCBTy callback;
+
+    /// The specified priority for the run.
+    uint64_t priority;
+
+    /// The runtime generated ID for the run request.
+    uint64_t requestID;
+
+    // Define greater than operator to allow sorting in priority_heap for queue
+    // reqests. If priority is the same fall back to order of submission.
+    bool operator>(const InferRequest &inferReq) const {
+      if (priority == inferReq.priority) {
+        return requestID > inferReq.requestID;
+      }
+      return priority > inferReq.priority;
+    }
+    InferRequest(std::string networkName,
+                 std::unique_ptr<ExecutionContext> context, ResultCBTy callback,
+                 uint64_t priority, uint64_t requestID)
+        : networkName{networkName}, context{std::move(context)},
+          callback{callback}, priority{priority}, requestID{requestID} {}
+  };
 
   /// Count of current in-flight networks being run. Atomic to allow
   /// concurrency in runNetwork.
@@ -58,6 +90,12 @@ class HostManager final {
   /// Count of total requests, this is used as a run ID. Atomic to allow
   /// concurrency in runNetwork.
   std::atomic<size_t> totalRequestCount_{0};
+
+  /// Priority queue for queued requests. This is a min-heap so lowest value is
+  /// popped first.
+  std::priority_queue<InferRequest, std::vector<InferRequest>,
+                      std::greater<InferRequest>>
+      inferQueue_;
 
   /// Configuration parameters for this Runtime Host.
   const HostConfig config_{};
@@ -87,6 +125,9 @@ class HostManager final {
 
   /// Set of networks in the process of being added.
   std::set<std::string> processingNetworks_;
+
+  /// Method to dispatch a new run to the executor.
+  void dispatchNextRun();
 
 public:
   /// Default constructor.
@@ -131,9 +172,12 @@ public:
   /// Note: This method is intended to be thread-safe, it will be called
   /// concurrently from multiple threads.
   /// Returns -1 if networkName not found or too many active requests.
+  /// The parameter \p priority is used to indicate queueing priority, priority
+  /// is lowest number first and in case of a tie the request that was submitted
+  /// first will go first.
   RunIdentifierTy runNetwork(llvm::StringRef networkName,
                              std::unique_ptr<ExecutionContext> context,
-                             ResultCBTy callback);
+                             ResultCBTy callback, uint64_t priority = 0);
 
   /// A wrapper around runNetwork that provides a blocking interface for an
   /// inference request. Runs the network provided in \p networkName using \p
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -316,10 +316,44 @@ llvm::Error HostManager::runNetworkBlocking(llvm::StringRef networkName,
   return std::move(*DCHECK_NOTNULL(runErr.get()));
 }
 
+void HostManager::dispatchNextRun() {
+
+  std::lock_guard<std::mutex> networkLock(networkLock_);
+  if (inferQueue_.size()) {
+    // Get the next request, unfortunately priority_queue only
+    // provides a const ref to the top element, since we need to move
+    // it we first cast it to remove the const.
+    auto request = std::move(const_cast<InferRequest &>(inferQueue_.top()));
+    inferQueue_.pop();
+    executor_->run(
+        networks_[request.networkName].dag.root.get(),
+        std::move(request.context), request.requestID,
+        [this, callback = request.callback, name = request.networkName](
+            RunIdentifierTy runID, llvm::Error err,
+            std::unique_ptr<ExecutionContext> context) {
+          {
+            std::lock_guard<std::mutex> networkLock(networkLock_);
+            auto it = networks_.find(name);
+            if (it != networks_.end()) {
+              it->second.refcount--;
+            }
+          }
+          TRACE_EVENT_INSTANT(context->getTraceContext(), TraceLevel::RUNTIME,
+                              "finish_" + name);
+          callback(runID, std::move(err), std::move(context));
+          dispatchNextRun();
+        });
+  } else {
+    // Decrement the activeRequest counter so new requests can
+    // launched.
+    --activeRequestCount_;
+  }
+}
+
 RunIdentifierTy
 HostManager::runNetwork(llvm::StringRef networkName,
                         std::unique_ptr<ExecutionContext> context,
-                        ResultCBTy callback) {
+                        ResultCBTy callback, uint64_t priority) {
   DCHECK(callback != nullptr);
 
   TRACE_EVENT_SCOPE(context->getTraceContext(), TraceLevel::RUNTIME,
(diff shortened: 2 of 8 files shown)


tests/unittests/HostManagerTest.cpp
```diff
@@ -217,3 +218,60 @@ TEST_F(HostManagerTest, ConfigureHostManager) {
...
+  hostManager->runNetwork(
+      "main", std::move(context2),
+      [&run2p, &counter](RunIdentifierTy runID, llvm::Error err,
+                         std::unique_ptr<ExecutionContext> context) {
+        EXIT_ON_ERR(std::move(err));
+        run2p.set_value(counter++);
+      },
+      1);
+  // Set the priority of the second to 0 so it should be first in the queue
+  // aftter run1.
```
beicy: nit : aftter->after

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -269,10 +269,48 @@ llvm::Error HostManager::runNetworkBlocking(llvm::StringRef networkName,
...
+        {
+          std::lock_guard<std::mutex> networkLock(networkLock_);
+          if (inferQueue_.size()) {
+            // Get the next request, unfortunately priority_queue only provides
+            // a const ref to the top element, since we need to move it we first
+            // cast it to remove the const.
+            auto request =
+                std::move(const_cast<InferRequest &>(inferQueue_.top()));
+            inferQueue_.pop();
+            dispatchRun(std::move(request));
```
beicy: When a request is pop from the queue, do we need to add 1 to  activeRequestCount_? Or it will be added somewhere?

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -50,15 +51,41 @@ class HostManager final {
     /// safety.
     std::atomic<size_t> refcount;
   };
+  /// Container for inference requests waiting in the queue.
+  struct InferRequest {
+    std::string networkName;
+    std::unique_ptr<ExecutionContext> context;
+    ResultCBTy callback;
+    uint64_t priority;
```
rdzhabarov: is priority going to be provided at onnxifi level?

lib/Runtime/HostManager/HostManager.cpp
gcatron: activeRequest gets decremented if there are no more requests in the queue, but queuedRequestCount_ definitely needs to be decremented.

include/glow/Runtime/HostManager/HostManager.h
gcatron: yes the plumbing for that still needs to be added. It should be straight forward so I'll add it to this PR.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -124,9 +154,11 @@ class HostManager final {
   /// Note: This method is intended to be thread-safe, it will be called
   /// concurrently from multiple threads.
   /// Returns -1 if networkName not found or too many active requests.
+  /// The parameter \p priority is used to indicate queuing priority, priotity
```
rdzhabarov: typo

nickgg:
@gcatron I thought this was a feature request to address a specific thrashing problem where we increase latency by interleaving partial parts of a network rather than running that network to the end, does this change fix that? Can we have a test that demonstrates it does?

gcatron:
@nickgg  This is addressing the thrash that sub-batching can introduce. where two requests get their sub-batches interleaved. Now if each request uses a different priority and assigned that priority to all of it's sub-batches they will be grouped together at execution. The QueueTest in HostManagerTest was added to test this functionality.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -344,37 +383,42 @@ HostManager::runNetwork(llvm::StringRef networkName,
...
   }
-
+  // Setup the request
+  InferRequest queuedRequest;
+  queuedRequest.networkName = networkName;
+  queuedRequest.context = std::move(context);
+  queuedRequest.priority = priority;
+  queuedRequest.callback = callback;
+  queuedRequest.requestID = currentRun;
+  // If we haven't reached maxActuveRequests kick off request.
```
bertmaher: typo: `maxActuveRequests`

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -344,37 +383,42 @@ HostManager::runNetwork(llvm::StringRef networkName,
         std::move(context));
     return currentRun;
   }
-
+  // Setup the request
+  InferRequest queuedRequest;
```
bertmaher: Why not have a proper constructor for `InferRequest`, instead of initializing fields one-by-one?

bertmaher:
I'll write up what I think I see in a moment, I just wanted to put a hold on this in case you were about to land it :-p

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -344,37 +383,42 @@ HostManager::runNetwork(llvm::StringRef networkName,
...
-                     }
-                   }
-                   TRACE_EVENT_INSTANT(context->getTraceContext(),
-                                       TraceLevel::RUNTIME, "finish_" + name);
-                   callback(runID, std::move(err), std::move(context));
-                   --activeRequestCount_;
-                 });
+  {
+    std::lock_guard<std::mutex> networkLock(networkLock_);
+    inferQueue_.push(std::move(queuedRequest));
```
bertmaher: So I think it might be possible to drop requests where we're close to the active request limit.  Here's what I suspect can happen:
Thread 1: request1 comes in, less than maxActive, dispatchRun is called, submits r1 to executor (T3)
Thread 2: request2 comes in, greater than maxActive, skips dispatchRun, but before it can be enqueued...
Thread 3 (executor): processes request1, sees an empty queue, returns
Thread 2: enqueues r2
... and nothing ever calls dispatchRun to process r2.

In fact it seems like the service could deadlock, if a bunch of requests pile up between the executor and the enqueue.

The control flow and synchronization here is pretty confusing so maybe this can't happen... but can you convince me it's not possible?

jackm321:
How do you envision priority being used?  I would think that a deadline might be more useful but I guess this could be used for that in a way

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -50,15 +51,51 @@ class HostManager final {
...
+    }
+  };
 
   /// Count of current in-flight networks being run. Atomic to allow
   /// concurrency in runNetwork.
   std::atomic<size_t> activeRequestCount_{0};
 
+  /// Count of current in-queue networks. Atomic to allow
+  /// concurrency in runNetwork.
+  std::atomic<size_t> queuedRequestCount_{0};
```
jackm321: why not use inferQueue_.size()?

lib/Runtime/HostManager/HostManager.cpp
gcatron: I think you're right, the second group of requests would be stuck until another request comes in, I'll fix it.

gcatron:
> How do you envision priority being used? I would think that a deadline might be more useful but I guess this could be used for that in a way

Notionally this allows for sub-batches to be grouped together, the thought would be a timestamp of submission time but a timestamp of a deadline would work too. Earliest deadline would be run first.

lib/Runtime/HostManager/HostManager.cpp
jackm321: I'm just skimming this, but that makes sense to me. Basically if thread2 gets descheduled at HostManager.cpp::399 while thread3 finishes the first request, nothing will pick up its request.
Also I agree that this is a bit confusing. The atomic state shared between the mutexed queues and the std::atomic counters describing the queues is a bit hard to follow.

gcatron: I think some of this complexity is I was trying to avoid enqueuing a request if it could be immediately dispatched. I think if I add the request to the queue first then there should be no race condition. 
T1 R1 comes in is enqueued, is dispatched
T2 R2, gets enqueued, if T1 finishes and grabs R2, then maxActive will still be at the limit, if T1 finishes early then maxActive will drop and R2 will get dispatched.
This means some amount of increased latency but I think that's better than more locking.

jackm321:
> Notionally this allows for sub-batches to be grouped together, the thought would be a timestamp of submission time but a timestamp of a deadline would work too. Earliest deadline would be run first.

Ah ok I see, thanks. FWIW from previous experience, deadline is often more useful.

tests/unittests/HostManagerTest.cpp
```diff
@@ -205,7 +206,7 @@ TEST_F(HostManagerTest, ConfigureHostManager) {
                           [lock](RunIdentifierTy runID, llvm::Error err,
                                  std::unique_ptr<ExecutionContext> context_) {
                             errToBool(std::move(err));
-                            std::unique_lock<std::mutex> guard(*lock);
+                            // std::unique_lock<std::mutex> guard(*lock);
```
bertmaher: commented code?

gcatron: removed now

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -316,10 +316,45 @@ llvm::Error HostManager::runNetworkBlocking(llvm::StringRef networkName,
   return std::move(*DCHECK_NOTNULL(runErr.get()));
 }
 
+void HostManager::dispatchNextRun() {
+  {
```
bertmaher: This scope looks redundant

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -316,10 +316,44 @@ llvm::Error HostManager::runNetworkBlocking(llvm::StringRef networkName,
...
+            std::lock_guard<std::mutex> networkLock(networkLock_);
+            auto it = networks_.find(name);
+            if (it != networks_.end()) {
+              it->second.refcount--;
+            }
+          }
+          TRACE_EVENT_INSTANT(context->getTraceContext(), TraceLevel::RUNTIME,
+                              "finish_" + name);
+          callback(runID, std::move(err), std::move(context));
+          dispatchNextRun();
```
bertmaher: The fact that we `dispatchNextRun` from inside a callback created by `dispatchNextRun` bugs me.  Is this, like, a common pattern in concurrent code?

It feels like what we really want is for the HostManager to push tasks into the priority queue, and to have a priority-aware ThreadPoolExecutor that knows how to pull the top pri item in its event loop.

cc @tracelogfb, just in case I'm out of my depth in concurrent program design...

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -334,47 +368,43 @@ HostManager::runNetwork(llvm::StringRef networkName,
...
+          MAKE_ERR(
+              GlowErr::ErrorCode::RUNTIME_REQUEST_REFUSED,
+              strFormat(
+                  "The number of allowed queued requests has been exceeded. "
+                  "queued requests: %lu allowed requests: %zu",
+                  queueSize, config_.maxQueueSize)),
+          std::move(context));
+      return currentRun;
+    }
+    inferQueue_.push(std::move(queuedRequest));
```
rdzhabarov: does not it race with inferQueue_.pop() https://github.com/pytorch/glow/pull/3416/files#diff-0ff368c87ec74b6b37e29804c4122427R327 ? (It seems that we do push without any lock here).

bertmaher: There's a lock on L365 it's just waaaay up the page (hopefully I'm counting braces correctly :-p )

rdzhabarov: ah, right '}' was removed and expand that lock to larger scope.

gcatron: hmm yeah the push should be under a lock.

rdzhabarov: one more dumb comment while we are here :)

Invoking callback under the lock might be a risky business ...

bertmaher: That's an interesting point, given we don't know what the callback does, it may not be totally safe to hold the lock through it.  @gcatron can you follow up?

PR #3418, opened 2019-08-14
[PyTorch] Add reshape op for new custom_fuse

zrphercule:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -233,6 +233,14 @@ struct LinearInputs {
   };
 };
 
+// Indexes of aten::reshape inputs.
+struct ReshapeInputs {
+  enum {
+    input = 0,
+	new_shape = 1,
+  }
+};
+
 // Indexes of aten::prelu inputs.
 struct PReluInputs {
   enum {
@@ -275,6 +283,8 @@ PyTorchModelLoader::getSymbolsMapping() {
        {{"aten::sigmoid", "aten::sigmoid_"},
         &PyTorchModelLoader::loadSigmoid,
         {}},
+	   {{"aten::reshape"}, &PyTorchModelLoader::loadReshape,
+	   {ReshapeInputs::input, ReshapeInputs::new_shape}},
        {{"aten::sub", "aten::sub_"}, &PyTorchModelLoader::loadSub, {}},
        {{"aten::relu", "aten::relu_"}, &PyTorchModelLoader::loadRelu, {}},
        {{"aten::t, aten::t_"}, &PyTorchModelLoader::loadTranspose, {}},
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -233,6 +233,14 @@ struct LinearInputs {
   };
 };
 
+// Indexes of aten::reshape inputs.
+struct ReshapeInputs {
+  enum {
+    input = 0,
+	new_shape = 1,
```
jackm321: vim is at again

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -275,6 +283,8 @@ PyTorchModelLoader::getSymbolsMapping() {
        {{"aten::sigmoid", "aten::sigmoid_"},
         &PyTorchModelLoader::loadSigmoid,
         {}},
+	   {{"aten::reshape"}, &PyTorchModelLoader::loadReshape,
+	   {ReshapeInputs::input, ReshapeInputs::new_shape}},
```
jackm321: input should not be frozen

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1050,6 +1060,25 @@ llvm::Error PyTorchModelLoader::loadMatMul(const torch::jit::Node *ptNode) {
...
+  auto outputs = ptNode->outputs();
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, 2, outputs, 1));
+
+  glow::NodeValue input;
+  ASSIGN_VALUE_OR_RETURN_ERR(input, getGlowNodeValue(inputs[ReshapeInputs::input]));
+
+  glow::Handle<uint32_t> shapeHandle;
+  ASSIGN_VALUE_OR_RETURN_ERR(shapeHandle, getGlowConstantHandle<uint32_t>inputs[ReshapeInputs::new_shape]);
+  std::vector<size_t> new_shape;
+  ASSIGN_VALUE_OR_RETURN_ERR(new_shape, expandParamIfNeeded<uint32_t, size_t>(shapeHandle, 4));
```
jackm321: this shouldn't be needed, we should take the shape as is and not expect a particular length

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -206,8 +206,13 @@ class PyTorchModelLoader {
   llvm::Error loadSqrt(const torch::jit::Node *ptNode);
 
   /// Load a PyTorch reciprocal node.
+  /// \returns error on failure.
   llvm::Error loadReciprocal(const torch::jit::Node *ptNode);
 
+  /// Load a PyTorch sqrt node.
```
jackm321: reshape node

torch_glow/tests/nodes/reshape_test.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+import torch_glow
+
+from tests.utils import jitVsGlow
+
+
+def jitScriptVsGlow(f_torch, f_glow, *inputs):
```
jackm321: Cool, no more tracing. Can we just use this for all tests?

torch_glow/tests/nodes/reshape_test.py
```diff
@@ -0,0 +1,48 @@
...
+        num_glow_nodes = len(glow_graph.findAllNodes(GLOW_NODE_NAME))
+        assert num_glow_nodes == 1, "Expected exactly 1 Glow node, found {}".format(num_glow_nodes)
+
+        assert torch.allclose(torch_res, glow_res, atol=01e-6)
+
+def test_reshape_basic():
+
+    @torch.jit.script
+    def reshape_basic(x, y, z):
+        # type: (Tensor, Tensor, Tuple[int, int]) -> Tensor
```
jackm321: I'm curious if it will work without this or no?

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: I wonder how can we do this? Given this shape is always a constant/ListConstruct

torch_glow/tests/nodes/reshape_test.py
zrphercule: I believe we can, since we have already migrated to our own customFuseNode

torch_glow/tests/nodes/reshape_test.py
zrphercule: If it is a torch script, then it wont work without this I believe.

zrphercule: But tracing does not have this problem.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Not sure what you mean? I'm probably missing something. We import IntLists as glow Constant int tensors, so can't we just use this here without expanding it to a specific size (4)?

PR #3420, opened 2019-08-15
Add FP16 converter helper, and then support UInt8FusedQTy to UInt8FusedFP16QTy conversion with it.

jfix71:
Summary: We want to support converting UInt8FusedQTy to UInt8FusedFP16QTy when we do FP16 conversion. This does not fit naturally into the TypeAToTypeBFunctionConverter. Instead of making major modifications/redesigning it, I added a helper to wrap the TypeAToTypeBFunctionConverter with Float -> Float16 with a secondary pass to convert UInt8FusedQTy to UInt8FusedFP16QTy.

I additionally added support for the graph optimization for ConvertTo to merge a Constant UInt8FusedQTy followed by a ConvertTo UInt8FusedFP16QTy into the Constant.

Test Plan: Added a test for the new graph optimization case that is supported. Added an OperatorTest and a conversion test to test that converting a FRWQ-SLWS to FP16 works correctly.


include/glow/Base/Tensor.h
```diff
@@ -573,6 +573,9 @@ class Tensor final {
   /// Convert each element of this tensor to \p newTy.
   void convertToType(ElemKind newTy);
```
rdzhabarov: could you update the comment on what's currently supported, in terms of typea -> typeb conversion

lib/Base/Tensor.cpp
```diff
@@ -543,20 +543,68 @@ void Tensor::init(InitKind init, float val, PseudoRNG &PRNG) {
...
-    llvm_unreachable("Type not supported");
+  *this = this->getCopyConvertedToType(newTy);
+}
+
+Tensor Tensor::getCopyConvertedToType(ElemKind newKind) const {
+  if (!isQuantizedElemKind(newKind)) {
+    Tensor tmp(newKind, dims());
+    switch (newKind) {
+    case ElemKind::Float16Ty:
+      assert(getElementType() == ElemKind::FloatTy && "Cast not implemented");
```
rdzhabarov: could we switch to DCHECK and add what exact cast is not implemented. Like Cast from X to Y is not implemented.

lib/Base/Tensor.cpp
```diff
@@ -543,20 +543,68 @@ void Tensor::init(InitKind init, float val, PseudoRNG &PRNG) {
...
+    return tmp;
+  }
+
+  // Handle Fused conversion. Currently only supports UInt8FusedQTy ->
+  // UInt8FusedFP16QTy.
+  assert((getElementType() == ElemKind::UInt8FusedQTy) &&
+         (newKind == ElemKind::UInt8FusedFP16QTy) &&
+         "Conversion not implemented.");
+  Tensor tmp(newKind,
+             {dims()[0], dims()[1] - 2 * (sizeof(float) - sizeof(float16_t))},
```
rdzhabarov: can we make an assumption here that dims().size() == 2?

lib/Graph/Nodes.cpp
```diff
@@ -553,10 +553,38 @@ bool Convolution3DGradNode::verify() const {
...
+
 bool ConvertToNode::verify() const {
-  bool isValid = checkSameShape(getInput(), getResult(), this);
   TypeRef srcTy = getInput().getType();
   TypeRef dstTy = getResult().getType();
+  const bool srcIsFused = isFusedQuantizedElemKind(srcTy->getElementType());
+  const bool dstIsFused = isFusedQuantizedElemKind(dstTy->getElementType());
+
+  bool isValid = expectCompareTrue(
+      "Must convert between Fused ElemKinds.",
```
rdzhabarov: a little bit incomplete, since it's ok for both to be unfused.

tests/unittests/BackendTestUtils.cpp
```diff
@@ -1131,4 +1131,81 @@ void runOnDevice(ExecutionContext &context, llvm::StringRef name,
...
+static void copyInScaleOffset(char *currRowScaleOffsetPtr) {
+  // Range (0, 255) -> (-0.1, 0.1)
+  T scale = 1.0f / 1275;
+  T offset = -0.1;
+
+  memcpy(currRowScaleOffsetPtr, &scale, sizeof(T));
+  memcpy(currRowScaleOffsetPtr + sizeof(T), &offset, sizeof(T));
+}
+
+Constant *createRandomizedConstant(Module &mod, TypeRef type,
```
rdzhabarov: wanted to put a comment to reuse this in recsys and noticed it's down below :) nice.

lib/Base/Tensor.cpp
jfix71: There are asserts in other places checking this for the Fused ElemKinds, but I can add a new check for it here as well.

jfix71:
@rdzhabarov All comments addressed!

PR #3423, opened 2019-08-15
[PyTorch] Add binding.h

zrphercule:
There are a few method in binding.cpp we would like to use.


zrphercule:
closed since there is another similar pr.

PR #3425, opened 2019-08-15
Refactor HostManager and Provisioner to lessen locking in addNetwork

### LINKED ISSUE #2842: [runtime] Compiles are done under the HostManager networkLock
Currently the HostManager holds the networkLock_ while doing optimization, compile and provisioning of networks. This has two unfortunate consequences:
* Only one compile can be done at a time (across all devices).
* No executions can be started while a compile is occurring.

Fix this by introducing a lock per backend (or better yet, a queue) and making the provisioner more granularly thread safe.

gcatron:
Summary:
This PR refactors HostManager::addNetwork and the Provisioner. Compiling functions no longer happens under lock.
If a network is in the process of being added and a new add or remove call is made targeting the same network name an error is returned that the network specified is busy.

Documentation:

Fixes #2842 

Test Plan: Ninja check and verify tests pass.

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -41,14 +41,14 @@ class HostManager final {
   /// NetworkData contains data about each network in HostManager that is needed
   /// by the runtime.
   struct NetworkData {
-    DAG dag;
+    DAG dag{};
     // Module that was used to create this network. Everything except
     // placeholders and types have been removed from it.
-    std::shared_ptr<Module> module;
+    std::shared_ptr<Module> module{nullptr};
 
     /// use an atomic refcount rather than just store a shared_ptr for thread
     /// safety.
-    std::atomic<size_t> refcount;
+    std::atomic<size_t> refcount{0};
   };
 
   /// Count of current in-flight networks being run. Atomic to allow
@@ -80,6 +80,13 @@ class HostManager final {
   /// The provisioner owns the compiledFunctions and handles loading functions
   /// onto the devices.
   std::unique_ptr<Provisioner> provisioner_;
+
+  /// Helper function to handle cleanup if an error occurs during addNetwork.
+  /// This must be called while holding the a lock on networkLock_.
+  void cleanupAddNetwork(llvm::ArrayRef<std::string> names);
+
+  /// Set of networks in the process of being added.
+  std::set<std::string> processingNetworks_;
 
 public:
   /// Default constructor.
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -92,21 +92,34 @@ HostManager::init(std::vector<std::unique_ptr<DeviceConfig>> configs) {
 
 HostManager::~HostManager() { llvm::toString(clearHost()); }
 
+void HostManager::cleanupAddNetwork(llvm::ArrayRef<std::string> names) {
+  for (auto &name : names) {
+    processingNetworks_.erase(name);
+  }
+}
+
 llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
                                     CompilationContext &cctx,
                                     bool saturateHost) {
-  std::lock_guard<std::mutex> networkLock(networkLock_);
-  auto functions = module->getFunctions();
-  for (auto &F : functions) {
-    std::string name = F->getName();
-    auto it = networks_.find(name);
-    if (it != networks_.end()) {
-      return MAKE_ERR(GlowErr::ErrorCode::RUNTIME_ERROR,
-                      "Failed to add network: already have a function called " +
-                          name);
+  std::vector<std::string> names;
+  {
+    std::lock_guard<std::mutex> networkLock(networkLock_);
+    auto functions = module->getFunctions();
+    for (auto &F : functions) {
+      std::string name = F->getName();
+      auto it = networks_.find(name);
+      if (it != networks_.end() ||
+          processingNetworks_.find(name) != processingNetworks_.end()) {
+        cleanupAddNetwork(names);
+        return MAKE_ERR(
+            GlowErr::ErrorCode::RUNTIME_ERROR,
+            "Failed to add network: already have a function called " + name);
+      }
+      // Add the network to processingNetworks_ so we know it's being worked on.
+      processingNetworks_.insert(name);
+      names.push_back(name);
     }
   }
-
   // Load backend-specific options if specified.
   if (!loadBackendSpecificOptionsOpt.empty()) {
     if (cctx.backendOpts.backendSpecificOpts.size() != 0) {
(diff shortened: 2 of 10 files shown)


include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -81,6 +81,9 @@ class HostManager final {
   /// onto the devices.
   std::unique_ptr<Provisioner> provisioner_;
 
+  /// Helper function to handle cleanup if an error occurs during addNetwork.
+  void cleanupAddNetwork(std::vector<std::string> names);
```
nadavrot: Please pass array_ref, or at least const-reference.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -155,21 +180,27 @@ llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
     provisioner_.reset(new Provisioner(devices_));
     executor_.reset(new ThreadPoolExecutor(devices_, config_.executorThreads));
   }
-
-  RETURN_IF_ERR(provisioner_->provision(nodeList, *module, cctx));
-
+  {
+    // std::lock_guard<std::mutex> networkLock(networkLock_);
```
nadavrot: What is this?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -116,17 +143,27 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
           if (backends_[i]->getBackendName() == nodeBackendName) {
             auto compiledOrErr = backends_[i]->compile(function, options);
             if (!compiledOrErr) {
+              {
```
nadavrot: This code is not documented. Could you add some comments that explains the logic?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -92,21 +92,36 @@ HostManager::init(std::vector<std::unique_ptr<DeviceConfig>> configs) {
...
+      std::string name = F->getName();
+      auto it = networks_.find(name);
+      if (it != networks_.end()) {
+        cleanupAddNetwork(names);
+        return MAKE_ERR(
+            GlowErr::ErrorCode::RUNTIME_ERROR,
+            "Failed to add network: already have a function called " + name);
+      }
+      // Put an entry into networks to denote this network is being added.
+      networks_[name];
```
rdzhabarov: a bit weird, networks_.insert ?

tests/unittests/ProvisionerTest.cpp
```diff
@@ -79,8 +79,8 @@ TEST_F(ProvisionerTest, provisionDag) {
   }
 
   CompilationContext cctx;
-  auto provisioner = Provisioner(devices);
```
rdzhabarov: curious why this change

gcatron: With the added mutex the default copy constructor is deleted, but a heap allocation isn't needed, changed back to stack allocation.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -92,21 +92,36 @@ HostManager::init(std::vector<std::unique_ptr<DeviceConfig>> configs) {
...
 HostManager::~HostManager() { llvm::toString(clearHost()); }
 
-llvm::Error HostManager::addNetwork(std::unique_ptr<Module> module,
-                                    CompilationContext &cctx,
-                                    bool saturateHost) {
-  std::lock_guard<std::mutex> networkLock(networkLock_);
-  auto functions = module->getFunctions();
-  for (auto &F : functions) {
-    std::string name = F->getName();
+void HostManager::cleanupAddNetwork(llvm::ArrayRef<std::string> names) {
```
rdzhabarov: I looked at the coverage for this PR: https://fb-glow-assets.s3.amazonaws.com/coverage/coverage-pull/3425/lib/Runtime/HostManager/HostManager.cpp.gcov.html

It looks like internals of the for loop below are not exercised. Can we make a test covering newly added code?

lib/Runtime/HostManager/HostManager.cpp
nickgg: I think you actually want a different map here, `pendingNetworks_` or something - that way you can store the key without needing a dummy value that might accidentally get used.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -61,16 +61,35 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
...
-  for (auto &network : networks) {
-    for (auto &node : network.nodes) {
-      for (auto logical : node->logicalDevices) {
-        auto it = logicalDevices.find(logical);
-        if (it != logicalDevices.end()) {
-          it->second.push_back(node.get());
-        } else {
-          logicalDevices.emplace(logical, std::vector<DAGNode *>{node.get()});
+  {
+    std::lock_guard<std::mutex> networkLock(functionsLock_);
```
nickgg: This makes me think the functionsLock_ protects `activeFunctions_` too since `functions_` is not touched in this block.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -115,18 +142,31 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
         for (size_t i = 0, e = backends_.size(); i < e; i++) {
           if (backends_[i]->getBackendName() == nodeBackendName) {
             auto compiledOrErr = backends_[i]->compile(function, options);
+            // Check to see if an error was encountered while compiling.
             if (!compiledOrErr) {
+              // If and error occured, clean up provisioning state and return
+              // the error.
+              {
+                std::lock_guard<std::mutex> functionLock(functionsLock_);
+                cleanupProvision(localActiveNames);
```
nickgg: It looks to me as though you always grab the functionsLock_ before calling cleanupProvision, consider moving the lock into the method call to avoid missing it.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -105,8 +124,16 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
     for (auto &node : device.second) {
       // Only compile if we haven't compiled before. If we have previously
       // compiled the function reuse it.
-      auto it = functions_.find(node->name);
-      if (it == functions_.end()) {
+      bool alreadyCompiled{false};
+      {
+        std::lock_guard<std::mutex> functionLock(functionsLock_);
```
nickgg: It's often more expensive to have a fine grained lock if you take and release it a lot, e.g. in this loop you take and release the lock 3 times per device?

I think you could refactor this to do the compile outside the lock then update all the state under the lock in one go.

lib/Runtime/Provisioner/Provisioner.cpp
gcatron: It does, updated the comment to call this out.

lib/Runtime/HostManager/HostManager.cpp
gcatron: @rdzhabarov  in that report blue means covered right? I think it's covered now.

lib/Runtime/Provisioner/Provisioner.cpp
gcatron: Good call, I refactored this to pull the lock out of the loop entirely. Now all compiled functions are added to functions_ at the end all in one lock.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -61,23 +61,44 @@ llvm::Error Provisioner::provision(DAGListTy &networks, Module &module,
...
+          }
         }
       }
     }
   }
   if (cctx.backendOpts.collectConstants) {
     // @todo(nickg) must use VLOG_EVERY_N temporarily here due to old glog
     // version.
+    // Locking here because glog log every n is not thread safe.
+    std::lock_guard<std::mutex> functionsLock(functionsLock_);
```
nickgg: What about the VLOG_EVERY_N calls below?

lib/Runtime/HostManager/HostManager.cpp
rdzhabarov: yup, blue is covered. red is not.

PR #3426, opened 2019-08-15
[PyTorch] Improve log quality

zrphercule:
Improve log qulity by ignore some "unsupported" ops


torch_glow/src/GlowFuser.cpp
```diff
@@ -131,16 +131,27 @@ bool canMerge(torch::jit::Block *block, isSupportFunc fn) {
     return c10::nullopt;                                                       \
   }
 
+
+bool isLogIgnoredOp(const string op_name) {
+  static string white_list[] = {"prim::GetAttr", "prim::Param"};
```
jackm321: std::string

torch_glow/src/GlowFuser.cpp
```diff
@@ -131,16 +131,27 @@ bool canMerge(torch::jit::Block *block, isSupportFunc fn) {
...
+  for (string st : white_list) {
```
jackm321: const std::string&

torch_glow/src/GlowFuser.cpp
jackm321: Maybe make this a unordered_set then check is O(1)

torch_glow/src/GlowFuser.cpp
```diff
@@ -131,16 +131,27 @@ bool canMerge(torch::jit::Block *block, isSupportFunc fn) {
     return c10::nullopt;                                                       \
   }
 
+
+bool isLogIgnoredOp(const string op_name) {
```
jackm321: doxygen

jackm321: const string&

torch_glow/src/GlowFuser.cpp
```diff
@@ -127,20 +127,28 @@ bool canMerge(torch::jit::Block *block, isSupportFunc fn) {
 
 #define REQ(cond, log_info)                                                    \
   if (!(cond)) {                                                               \
-    llvm::errs() << log_info;                                                  \
+    llvm::errs() << log_info << "\n";                                          \
     return c10::nullopt;                                                       \
   }
 
+// Check if a node is a known not-lowered operator, etc, a tensor generator.
+bool isLogIgnoredOp(const std::string &op_name) {
```
rdzhabarov: we do not use '_' in names, opName.

Same in other places.

rdzhabarov: also, make it static so that it's private to the translation unit.

zrphercule: oops, I used this convention for all the naming in torch_glow...

jackm321: We shouldn't use snake names, only camel case. The exception in torch_glow should be the stuff like the code in FuseLinear where it is copied basically straight from torch/tvm and I'm fine with no adapting it to glow because I would like to move it into a common place in pytorch. Other than that, we should be sticking with glow's regular style norms

rdzhabarov:
@zrphercule can you take a look at test failure. I remember I saw that before, maybe rebase on the latest master and push it again.

yinghai:
@zrphercule Can we land this?

zrphercule:
> @zrphercule Can we land this?

test seems flaky. let me rebase it.

rdzhabarov:
Ignore OpenCL failures, but Pytorch must be green

yinghai:
Yes, it definitely look fishy.

zrphercule:
Yeah I think I know the reason now. Fixing...

PR #3429, opened 2019-08-16
Fix ASAN stack usage after release detected bug.

putivsky:
Summary: Make a copy of the temporary object to avoid usage after release bug.

Differential Revision: D16865364

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -28,12 +28,14 @@ namespace {
 
 /// \returns a corresponding Glow Type for a given PyTorch CompleteTensorType \p
 /// ptType.
-inline glow::Type ptTypeToGlowType(const c10::ProfiledTensorType &ptType) {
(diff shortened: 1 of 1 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -28,12 +28,14 @@ namespace {
 
 /// \returns a corresponding Glow Type for a given PyTorch CompleteTensorType \p
 /// ptType.
-inline glow::Type ptTypeToGlowType(const c10::ProfiledTensorType &ptType) {
+inline glow::Type ptTypeToGlowType(const at::ProfiledTensorType &ptType) {
   // TODO: get correct ElemKind
   DCHECK_EQ(*ptType.scalarType(), at::kFloat)
       << "Only float type supported currently.";
   std::vector<size_t> dims;
-  for (auto &size : ptType.sizes().concrete_sizes().value()) {
```
rdzhabarov: would const ref work here?
for(const auto& size : ...) {
...
}

putivsky:
closing, Jack fixed it in another PR.

PR #3431, opened 2019-08-19
Fix fusion order bug

zrphercule:
Same problem with: https://github.com/pytorch/tvm/pull/88
When fusing, after a node is changed, we need to recursively re-merge all its neighbor nodes, instead of only searching the afterward nodes.
Here since we dont care about the time complexity, we just simply search all operators again (and again and again... O(N^2) worst case but still tolerable.)


torch_glow/src/GlowFuser.cpp
```diff
@@ -185,26 +185,33 @@ c10::optional<torch::jit::Node *> tryMerge(torch::jit::Node *consumer,
...
 void GlowCustomFuse(std::shared_ptr<torch::jit::Graph> graph, isSupportFunc fn,
                     at::Symbol kind) {
   torch::jit::AliasDb aliasDb(graph);
   auto block = graph->block();
 
-  for (auto it = block->nodes().rbegin(); it != block->nodes().rend();) {
-    it = getNewNode(*it, aliasDb, block, fn, kind);
-  }
+  bool is_changed;
+  do {
```
jackm321: why not just use a while loop like in https://github.com/pytorch/tvm/pull/88/files?

yinghai:
Yeah, I've brought up this several times. We should just have once piece of code.

jackm321:
@yinghai or @bwasti do you have any suggestions for this? Is there some place in pytorch we can put these things?

torch_glow/src/GlowFuser.cpp
zrphercule: o it is same thing...

zrphercule:
> @yinghai or @bwasti do you have any suggestions for this? Is there some place in pytorch we can put these things?

Think we can have this for now, and migrate it to pytorch at some time, or when we need to do some significant change

PR #3432, opened 2019-08-19
add an OperatorTest for Fp16 Convolutions

nickgg:
Summary: We don't have a simple test of Convolutions that uses FP16. Adding this one. Should it be enabled on Habana as well?

Documentation: n/a

Test Plan: run test


nickgg:
Yep, pretty much. Figured it didn't hurt to have a super simple convolution in the suite.

PR #3434, opened 2019-08-19
[Importer] Error out if bias specified but failed to load (#3216)

### LINKED ISSUE #3216: ONNX Importer loadConv doesn't return error when bias is not found but is expected.
When bias is expected, `getConstantByNameOrNull `allows for bias not to be found, which will result in the zero bias being created. Isn't returned NULL supposed to be an error ?
Actually, this was an error until PR #2865. The code before this change was:
`ASSIGN_VALUE_OR_RETURN_ERR(b, getTensorByName(biasTensorName));
`
https://github.com/pytorch/glow/blob/cb7c305923902f375522771379b77ac93210367e/lib/Importer/ONNXModelLoader.cpp#L604-L608

udo-git:
Summary
Return error if a bias was specified for the Conv operator but
it failed to load. Previous behavior was to (silently) use a
default bias. This change makes the failure explicit for better
handling downstream.

Documentation:
N/A

[Optional Fixes #issue]
Fixes #3216

Test Plan:
Unit tests 100% passing

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -605,7 +605,7 @@ llvm::Error ONNXModelLoader::loadConv(const ONNX_NAMESPACE::NodeProto &op,
   if (op.input_size() > 2) {
     auto &biasTensorName = op.input(2);
     // Load the serialized bias vector.
-    bias = getConstantByNameOrNull(biasTensorName);
+    ASSIGN_VALUE_OR_RETURN_ERR(bias, getConstantByName(biasTensorName));
   }
 
   // If a serialized bias wasn't found then create a zero bias.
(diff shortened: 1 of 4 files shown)


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -605,7 +605,7 @@ llvm::Error ONNXModelLoader::loadConv(const ONNX_NAMESPACE::NodeProto &op,
   if (op.input_size() > 2) {
     auto &biasTensorName = op.input(2);
     // Load the serialized bias vector.
-    bias = getConstantByNameOrNull(biasTensorName);
+    ASSIGN_VALUE_OR_RETURN_ERR(bias, getConstantByName(biasTensorName));
```
rdzhabarov: i do not think this fixes a problem.

getConstantByName does not return error in any case.


Also, we' need to have a unit test in ONNXModelLoader to make sure we cover the erroneous code path.

rdzhabarov: I actually overlooked `getConstantByName` and thought it's getConstantByNameOrNull.
This seems to be fine.

udo-git: Thanks! I added a test to make sure it does. The test passes as-is but fails when I change back to getConstantByNameOrNull.

PR #3435, opened 2019-08-19
[PyTorch] Use ProfiledTensorType

jackm321:
Summary:
Fixes glow build.
All the other tensor types went away so we'll use this one. Eventually we should enable (at least under a flag) the profiling executor but for now we just create ProfiledTensorTypes manually from graph inputs.

Documentation:
n/a

Test Plan:
python setup.py test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -28,15 +28,18 @@ namespace {
 
 /// \returns a corresponding Glow Type for a given PyTorch CompleteTensorType \p
 /// ptType.
-/// TODO: switch to ProfiledTensorType
-inline glow::Type ptTypeToGlowType(const at::CompleteTensorType &ptType) {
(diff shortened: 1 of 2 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -523,8 +526,8 @@ PyTorchModelLoader::loadValue(const torch::jit::Value *value) {
   RETURN_ERR_IF_NOT(value->isCompleteTensor(),
                     glow::strFormat("Value %s must have CompleteTensor type.",
                                     value->debugNameBase().c_str()));
-  auto ptType = value->type()->cast<at::CompleteTensorType>();
-  auto glowType = ptTypeToGlowType(*ptType.get());
+  auto glowType = ptTypeToGlowType(
+      *DCHECK_NOTNULL(value->type()->cast<at::ProfiledTensorType>()));
```
Krovatkin: fyi: you could also use `value->type()->expect<at::ProfiledTensorType>()` which also asserts that the cast is successful.

PR #3438, opened 2019-08-20
[PyTorch] Fuse reshape and its inputs to glow

jackm321:
Summary:
Import `aten::reshape` as well as `prim::ListConstruct` for int[] and `aten::size` in order to get that last couple of ops in resnet lowered.

Also this enable lowering of transpose an cleans up fusers.

This PR also introduces a bit of a problem. Before all outputs of Glow could be correctly mapped to PT tensors (even if not all inputs could). However now that we have `prim::ListConstruct` and `aten::size`, the glow subgraph can easily produce output IValues that are not tensors. We can handle this issue in a followup. We may need to add some metadata to each input/output placeholder to PyTorchModelLoader that explains how to map it to PyTorch values.

Documentation:
doxygen

Test Plan:
CI
run resnet_example.py

Screen Shot 2019-08-20 at 10 11 56 AM


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -634,6 +691,119 @@ llvm::Error PyTorchModelLoader::loadMax(const torch::jit::Node *ptNode) {
...
+      F_.getParent()->createConstant("size_output", std::move(sizeT));
+
+  return addGlowNodeValue(outputs[0], sizeConstant);
+}
+
+llvm::Error
+PyTorchModelLoader::loadListConstruct(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, -1, outputs, 1));
```
putivsky: Additional comment would be useful: -1 means no checks for inputs.

jackm321: I put this on checkInputAndOutputSizes definition, I can add a comment here too

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <torch/csrc/jit/passes/dead_code_elimination.h>
+#include <torch/csrc/jit/passes/subgraph_rewrite.h>
+
+namespace glow {
+
+/// This is mainly copied from pytorch/tvm
+/// This pass fuse the addmm or matmul + add generated by JIT back to linear
+/// to allow direct support with Glow integration with Glow IR
+/// This pass can be deleted once the JIT can emit the aten::linear in the
+/// future
```
putivsky: period at the end.

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+  %weight_t = aten::t(%weight)
+  %output = aten::mm(%input, %weight_t)
+  return (%output))IR";
+  std::string fused_linear_bias_none = R"IR(
+graph(%input, %weight):
+  %bias: Tensor? = prim::Constant()
+  %res = aten::linear(%input, %weight, %bias)
+  return (%res))IR";
+
+  // replace addmm pattern to linear
```
putivsky: ditto

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+  torch::jit::SubgraphRewriter addmm_to_linear;
+  addmm_to_linear.RegisterRewritePattern(addmm_pattern, fused_linear);
+  addmm_to_linear.runOnGraph(graph);
+
+  // replace matmul + add pattern to linear
```
putivsky: ditto

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+  torch::jit::SubgraphRewriter matmuladd_to_linear;
+  matmuladd_to_linear.RegisterRewritePattern(matmul_add_pattern, fused_linear);
+  matmuladd_to_linear.runOnGraph(graph);
+
+  // replace mm + add pattern to linear
```
putivsky: ditto

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+  torch::jit::SubgraphRewriter mmadd_to_linear;
+  mmadd_to_linear.RegisterRewritePattern(mm_add_pattern, fused_linear);
+  mmadd_to_linear.runOnGraph(graph);
+
+  // replace matmul with bias=None pattern to linear
```
putivsky: ditto

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,97 @@
...
+  torch::jit::SubgraphRewriter matmul_to_linear;
+  matmul_to_linear.RegisterRewritePattern(matmul_pattern,
+                                          fused_linear_bias_none);
+  matmul_to_linear.runOnGraph(graph);
+
+  // replace mm with bias=None pattern to linear
```
putivsky: ditto

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -151,16 +151,34 @@ llvm::Error contractParamIfNeeded(const glow::Handle<T> &handle,
...
 template <typename T>
-llvm::Error checkInputAndOutputSizes(const T &inputs, size_t inputsSize,
-                                     const T &outputs, size_t outputsSize) {
-  RETURN_ERR_IF_NOT(inputs.size() == inputsSize,
-                    glow::strFormat("Expected number inputs %lu, got %lu.",
-                                    inputsSize, inputs.size()));
-  RETURN_ERR_IF_NOT(outputs.size() == outputsSize,
-                    glow::strFormat("Expected number outputs %lu, got %lu.",
-                                    outputsSize, outputs.size()));
+llvm::Error checkInputAndOutputSizes(const T &inputs, int64_t inputsSize,
```
putivsky: Method can be extended to check not only sizes equality, == but also >=, may be negative size can be treated as >=.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -634,6 +691,119 @@ llvm::Error PyTorchModelLoader::loadMax(const torch::jit::Node *ptNode) {
...
+
+  // Index must be a Constant.
+  int32_t dim;
+  auto dimHandle = glow::Handle<int32_t>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      dimHandle, getGlowConstantHandle<int32_t>(inputs[SizeInputs::dim]));
+  RETURN_IF_ERR(getSingleValueFromConstant(dimHandle, dim));
+
+  if (dim == -1) {
+    dim = input.dims().size() - 1;
```
putivsky: what if the input dimensions size is zero?

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -101,10 +101,15 @@ class PyTorchModelLoader {
...
+  static glow::ElemKind convertScalarType(c10::ScalarType ty);
+
   /// Takes a glow::Function \p F, a jit::Graph \p subgraph to load, and a
-  /// stack of \p inputs for the subgraph to be loaded. Parameter \p settings
-  /// control the fusion details. Output parameters \p inputPlaceholders and
-  /// \p outputPlaceholders are filled out. \returns error on failure
+  /// stack of \p inputs for the subgraph to be loaded. Parameter \p
+  /// settings control the fusion details. Output parameters \p
+  /// inputPlaceholders and \p outputPlaceholders are filled out. \returns
+  /// error on failure
```
putivsky: period.

torch_glow/src/FuseLinear.cpp
jackm321: This is moved code but I will add periods (and also correct the snake-case format like matmul_to_linear)

torch_glow/src/FuseLinear.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <torch/csrc/jit/jit_log.h>
+#include <torch/csrc/jit/passes/alias_analysis.h>
+#include <torch/csrc/jit/passes/common_subexpression_elimination.h>
+#include <torch/csrc/jit/passes/dead_code_elimination.h>
+#include <torch/csrc/jit/passes/subgraph_rewrite.h>
+
+namespace glow {
+
+/// This is mainly copied from pytorch/tvm
+/// This pass fuse the addmm or matmul + add generated by JIT back to linear
```
jfix71: Maybe dumb question, but: if I understand correctly, we're essentially recreating the FC (linear) from the mm + add, and then importing it as linear, correct? Why can't we just import it as linear and then do this fusion in Glow?

jackm321: @jfix71 no it's not a dumb question, when glow's jit ir pass sees this, it's lowered to addmm, matmul, etc but would be nice to have it as linear and it's easy to fuse it here. Eventually the hope is this will move to pytorch directly then we won't have to worry about this, user will use a linear op and glow will see a linear op. Glow could introduce a folding pass that does this too.

jfix71: Ok got it. NBD to just go with this as is since it's already done, but my two cents are that we'll probably want to support importing addmm or matmul + add by themselves (if they aren't already supported). And that it's also useful to have this folding to FC logic in Glow too, since it would help all frontend importers. Whereas this seems like a temporary workaround that we'll throw away eventually and only helps PyTorch.

jackm321: 100% agree with all of that

PR #3440, opened 2019-08-20
Fix quantization when doing parallel clone testing; add check for bitwise accuracy

jfix71:
Summary: Previously we were only checking the results from parallel clones against the interpreter result within some threshold. With this change we will also check that all of the parallel clone results from the backend are bitwise accurate.

When implementing this I found that things were currently broken; the quantizer was not erroring out when it should have been (fixed in 1st commit). Additionally we were not copying over the quantization infos for the cloned nodes to find their correct quantization parameters, so without the check from the 1st commit those nodes were just not being quantized.

Test Plan: All tests still pass with the added check; also ran OperatorTests with `--parallel-clone-count=10`.

### DIFF
--- a/tests/unittests/BackendTestUtils.cpp
+++ b/tests/unittests/BackendTestUtils.cpp
@@ -159,12 +159,6 @@ void compareAgainstInterpreter(llvm::StringRef backendName,
   FunctionTensorPair IFT = createAndInitFunction(iBindings, IEE);
   FunctionTensorPair BFT = createAndInitFunction(bBindings, BEE);
 
-  // Clone the Function inside itself many times if desired.
-  std::unordered_set<Tensor *> resultTensors =
-      cloneFunInsideFun(BFT, &bBindings, count);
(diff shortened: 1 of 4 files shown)


tests/unittests/BackendTestUtils.cpp
```diff
@@ -188,10 +188,22 @@ void compareAgainstInterpreter(llvm::StringRef backendName,
   for (Tensor *T : resultTensors) {
     EXPECT_TRUE(IFT.second->isEqual(*T, allowedError));
   }
+
+  // Additionally check that each of the results from the parallel cloned
+  // Functions are bitwise equal.
+  Tensor *firstResult = nullptr;
+  for (Tensor *T : resultTensors) {
+    if (!firstResult) {
```
gcatron: Nit: I think this could be more readable if you move setting firstResult outside the loop.
`Tensor *firstResult = *resultTensors.begin()`

PR #3441, opened 2019-08-20
[PyTorch] Fix pytorch build

jackm321:
Summary:
See title

Documentation:
none

Test Plan:
python setup.py test
CI

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -128,7 +128,6 @@ elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     source venv/bin/activate
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
-    git checkout 7f86fb8995d9990fe1eac5a18335523cb3da7f09
     pip install -r requirements.txt
     python setup.py install
     cd ${GLOW_DIR}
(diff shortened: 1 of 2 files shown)


.circleci/build.sh
```diff
@@ -128,7 +128,6 @@ elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
     source venv/bin/activate
     git clone https://github.com/pytorch/pytorch.git --recursive
     cd pytorch
-    git checkout 7f86fb8995d9990fe1eac5a18335523cb3da7f09
```
rdzhabarov: I'm curious how did we get a build failure on a master if we have not updated pytorch sha before?

jackm321: Because the diff that caused this was a pytorch that changed tensor types in pytorch and in glow, there was a corresponding pytorch github pr but not a corresponding glow github pr so no way for failure to occur on master. I removed this sha here because basically the thought is that if glow is tracking pytorch master in fbcode it should be in open source also other wise this will happen. If anyone has alternative thoughts I'm open to it though.

rdzhabarov: Got it. I think this sounds good, it does not seem that build time increased a lot and anyway we track pytorch master in fbcode very close as you said.

PR #3443, opened 2019-08-21
Enable saturateHost in image-classifier and address a couple resulting bugs

gcatron:
Summary:
This PR enables saturate host in image-classifier, so now if multiple devices are created they will be used. 
As a result of this change bugs in OpenCL and CPUFunctions were exposed this PR fixes those too.

Documentation: 

Test Plan: run image-classifier with num-devices >0 and backends= CPU, Interpreter, and OpenCL.

### DIFF
--- a/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
+++ b/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
@@ -53,6 +53,10 @@ protected:
   /// The LLVM JIT engine. The jit must be initialized after the ctor
   /// initializes the LLVM backends.
   std::unique_ptr<llvm::orc::GlowJIT> JIT_;
+
+  /// The JIT can be accessed from multiple threads but is not thread safe,
+  /// JITLock_ protects it.
+  std::mutex JITLock_;
 };
 } // end namespace glow
 
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -237,14 +237,15 @@ static size_t setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
 
 void OpenCLFunction::fillBuffer(cl_mem buffer, uint64_t start, uint64_t len,
                                 float value, ElemKind elemKind,
-                                runtime::OpenCLDeviceBindings *devBindings) {
+                                runtime::OpenCLDeviceBindings *devBindings,
+                                std::vector<KernelLaunch> &kernelLaunches) {
   auto kernel =
       createKernel(getKernelName("splat", elemKind), devBindings->program);
   setKernelArg(kernel, 0, buffer);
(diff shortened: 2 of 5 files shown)


include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
```diff
@@ -53,6 +53,10 @@ class LLVMCompiledFunction : public CompiledFunction {
   /// The LLVM JIT engine. The jit must be initialized after the ctor
   /// initializes the LLVM backends.
   std::unique_ptr<llvm::orc::GlowJIT> JIT_;
+
+  /// The JIT can be accessed from multiple threads but is not thread safe
```
jfix71: nit: period.

tools/loader/Loader.cpp
```diff
@@ -419,7 +419,7 @@ void Loader::compile(CompilationContext &cctx) {
                    mainEntryName.empty() ? networkName : mainEntryName);
   } else {
     // Emit IR for the graph and compile it.
-    auto error = hostManager_->addNetwork(std::move(M_), cctx);
+    auto error = hostManager_->addNetwork(std::move(M_), cctx, true);
```
jfix71: Do we want to add a command line option for this, instead of hard coding true?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -237,14 +237,15 @@ static size_t setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
...
-                                runtime::OpenCLDeviceBindings *devBindings) {
+                                runtime::OpenCLDeviceBindings *devBindings,
+                                std::vector<KernelLaunch> &kernelLaunches) {
   auto kernel =
       createKernel(getKernelName("splat", elemKind), devBindings->program);
   setKernelArg(kernel, 0, buffer);
   setKernelArg<cl_uint>(kernel, 1, start);
   setKernelArg(kernel, 2, value);
   enqueueKernel("splat", devBindings->commandQueue, kernel,
-                devBindings->deviceId, {(size_t)len}, kernelLaunches_);
```
jfix71: Is `kernelLaunches_` no longer used? Should it be removed from `OpenCLFunction`?

tools/loader/Loader.cpp
gcatron: I think hardcoding here is fine, Since image-classifier only loads one network I think replication is the only real use when num-devices>1.

jfix71: What about partitioning? I guess if we need to use partitioning it just won't clone any of the partitions functions to the devices? Actually, curious what happens if we have 3 devices, 2 partitions, and saturate is true...?

gcatron: Then we don't duplicate, saturate only happens if there is enough space for a full copy of the network on available devices.

PR #3448, opened 2019-08-22
[NodeGen] Add the ability to use external user-defined types

opti-mix:
Summary: add the ability to use external user-defined types in NodeGen.

This is quick proof-of-concept based on discussions with @SplitInfinity and @nickgg about adding fields of custom types (including enums) to Nodes and Instructions.

Documentation:
Definitions of such user-defined types do not need to be known when compiling the NodeGen.cpp, but it should be known when compiling the generated files.  The types can be referenced in NodeGen by means of the MEMBER_TYPE_INFO macros as shown below. There is also MEMBER_TYPE_INFO_WITH_CMP for providing a custom comparator function, e.g. for pointer types.

The intended use in NodeGen is something like this:
```c++
  // Include the header providing the definitions of user-defined types into the generated header file.
  BB.includeHeader("glow/UserTypes/UserDefinedTypes.h");
  ...
  // Define a node that uses user-defined types.
  BB.newNode("CustomNode")
      .addMember(MEMBER_TYPE_INFO(UserDefinedExternalType), "F1")
      .addMember(MEMBER_TYPE_INFO_WITH_CMP(UserDefinedExternalType *, "isEqual"), "F2")
      .addMember(MEMBER_TYPE_INFO(UserDefinedExternalType &), "F3");
```
Each user-defined type used this way should satisfy the following requirements:
   * There should be a hash function with a signature like `llvm::hash_code hash_value(const T)` which takes T by value, by reference or as a pointer, depending on the intended use.
  * There should be a stream output operator with a signature like `llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const T);`, which takes T by value, by reference or as a pointer, depending on the intended use.
  * There should be a comparison operator `bool operator==(const T LHS, const T RHS)` (or a custom comparator function mentioned in MemberTypeInfo::cmpFn), which takes T by reference or by value depending on the intended use.

It is a responsibility of developer to define these functions for their custom types .

Test Plan:
ninja test


opti-mix:
I have locally added a node to the NodeGen and everything builds and runs just fine. I'm not sure though how to add this check as a test as we don't have any NodeGen tests yet and they can be a bit difficult to setup using our build system. Suggestions are welcome!

opti-mix:
@SplitInfinity Could you have a look? I think you tried to implement something very similar before and faced certain issues (e.g. use of types before they were defined, may be some namespacing related problems, etc). I'd like to check with you if this quick hack eventually has the same kind of problems.

SplitInfinity:
Passing a fully qualified name to `TYPE_INFO` should be enough to avoid namespace issues. The only potential issue I see is that the hash function, stream output operator and comparison operator for the user-defined type will have parameters of type `T&`, so the hasher, equator, and printer methods generated for a node may not work as intended if a member on the node has type `T*`:
- the printer method will print out the address of the object that the member points to, which isn't that useful
- the equator method will check that the addresses stored in the `T*` members of the LHS and RHS are equal, but perhaps a "deep comparison" that uses `bool operator==(const T &LHS, const T &RHS` is more useful in this case?
- the hasher will compute the hash based on the address stored in the `T*` member instead of invoking `llvm::hash_code hash_value(const T &)`

opti-mix:
> Passing a fully qualified name to TYPE_INFO should be enough to avoid namespace issues. 

Yep.

> The only potential issue I see is that the hash function, stream output operator and comparison operator for the user-defined type will have parameters of type T&, so the hasher, equator, and printer methods generated for a node may not work as intended if a member on the node has type T*:

Yes, this can be an issue. For the hasher and printer methods I relaxed the wording in the comments to make clear that they need to take the appropriate type. I also tested it locally and everything works just fine.

As it is impossible to define `operator ==` for pointer types, I added a possibility to provide a custom comparator function. Alternatively, one can use some kind of a smart pointer instead of a raw pointer and overload `operator ==` for this smart pointer type as needed.

SplitInfinity:
I think this makes sense now. Were you planning on landing this as is? Do you need this for something? I think it makes sense to land @nickgg's change to fuse convolution + activation and then refactor it afterwards using this change.

nickgg:
I like this idea

opti-mix:
> Were you planning on landing this as is? Do you need this for something? 

I wanted to make @nickgg's life easier ;-) And it could be useful in the future.

> I think it makes sense to land @nickgg's change to fuse convolution + activation and then refactor it afterwards using this change.

Yes. This is exactly the plan.

I'm trying to add some tests to the PR and then it should be good to ho.

SplitInfinity:
Should we get rid of `EnumBuilder` as well?

opti-mix:
Hmm. I seems like it is almost impossible to write a standalone test for ClassGen, i.e. a test that would define a node or instruction using user-defined types:
- The part that generates the headers and cpp files can be easily done without any issues.
- But when you try to use the new nodes/instructions by including those generated header files, you cannot do it, because Glow's graph layers includes `glow/AutoGenNodes.def` and `glow/AutoGenInstrs.def` in many places to dispatch different Nodes/Instrs methods. There is no clean way to tell it to include the custom headers generated for the tests. The only working way I can see is to define a new node in the NodeGen.cpp and a new instruction in the InstrGen.cpp, but we probably don't want to pollute these files just for the sake of testing. 

WDYT?

opti-mix:
@SplitInfinity 

> Should we get rid of EnumBuilder as well?

Yes, I seems it is just easier to define an external user-defined enum in C++ that can be used by both NodeGen and InstrGen rather than define some custom enums by means of EnumBuilder APIs.

opti-mix:
@SplitInfinity @nickgg Pushed some updates:
- Removed EnumBuilder
- Instead of adding a standalone test, which is almost impossible (see my remarks above), I actually changed the convolution node and instruction to use user-defined `ConvolutionLayout` and `FusedActivation` enums.

PR #3449, opened 2019-08-22
[glow][Executor] Move ExecutionState to private header

bertmaher:
Summary: Shrinking the public API of ThreadPoolExecutor by moving ExecutionState into lib/

Test Plan: ninja all


rdzhabarov:
This is NFC right?

bertmaher:
Yeah, NFC

PR #3454, opened 2019-08-23
Simplify worker thread loop

bertmaher:
Summary:
- Use the idiomatic predicate version of cv.wait while waiting on the work queue.
- Avoid checking shouldStop_ while not holding lock.
- Trim some comment verbosity.

Test Plan: ThreadPoolTest, ThreadPoolExecutorTest

### DIFF
--- a/lib/Support/ThreadPool.cpp
+++ b/lib/Support/ThreadPool.cpp
@@ -38,31 +38,16 @@ void ThreadExecutor::stop(bool block) {
 }
 
 void ThreadExecutor::threadPoolWorkerMain() {
-  std::unique_lock<std::mutex> lock(workQueueMtx_, std::defer_lock);
-
-  while (!shouldStop_) {
-    // Lock the lock after processing a work item.
-    lock.lock();
-
-    // If work queue is empty, wait to be signalled when
-    // a work item is submitted.
-    while (workQueue_.empty() && !shouldStop_) {
-      queueNotEmpty_.wait(lock);
-    }
-
-    // If shouldStop_ was set to false while the thread
-    // was asleep, break out of the main loop.
+  while (true) {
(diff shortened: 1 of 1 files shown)


rdzhabarov:
@bertmaher could you do "assignees". In that case it's clearly visible from https://github.com/pytorch/glow/pull page.

bertmaher:
Sure!  It's weird that there's a separate "request" and "assign", I thought they were the same concept!

nickgg:
Uh, I'm not following that problem @rdzhabarov:
* shouldStop_ is atomic.
* bool read/writes are sequentially consistent on x86.
* writing to shouldStop_ is done only under the lock anyway, and always triggers the cv event.

Don't really see what harm the comments were doing here as well.

rdzhabarov:
> shouldStop_ is atomic.

if it's atomic it's ok. There is no reason for that to be atomic as we do writes/reads under workQueueMtx_ locked.

bertmaher:
There wasn't actually a bug with `shouldStop_`, but it felt a bit fishy.  Regardless of atomicity it needs to be checked inside a critical section (like you said) so that it won't miss a wakeup.  Given that I don't see a point in doing an unlocked check of shouldStop - I guess it slightly optimizes the shutdown path?

bertmaher:
Re: the comments, I mean, maybe I should leave well enough alone, but things like
```
// Lock the lock.
lock.lock()
```
don't seem like they actually clarify anything, since they more-or-less repeat that code.  At a glance all the verbosity seemed like a signal that something tricky was going on, when in fact it's just a pretty standard work dispatch loop.

bertmaher:
Anyways, whatever.  This change isn't really necessary anyways.

PR #3455, opened 2019-08-23
[PyTorch] Cache graphs per node using input shapes

jackm321:
Summary:
* Use one CachingGraphRunner per glow::FusionGroup op (instead of globally)
* Use c10::TensorType shapes to cache glow functions for particular input shape

Documentation:
doxygen

Test Plan:
python setup.py test
add prints, see that things get reused

### DIFF
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -26,6 +26,28 @@
 
 namespace glow {
 
+namespace {
+/// Builds and \returns a HostManager instance.
+std::unique_ptr<runtime::HostManager> buildHostManager() {
+  constexpr size_t numGlowDevices = 1;
+
+  std::vector<std::unique_ptr<runtime::DeviceConfig>> deviceConfigs;
+  for (int i = 0; i < numGlowDevices; i++) {
+    deviceConfigs.push_back(llvm::make_unique<runtime::DeviceConfig>(
+        getPyTorchLoaderSettings().glowBackendName));
+  }
+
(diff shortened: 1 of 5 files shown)


torch_glow/src/PyTorchCommon.cpp
```diff
@@ -55,9 +77,13 @@ void registerGlowOp() {
   torch::jit::RegisterOperators op({torch::jit::Operator(
       getGlowSymbol(),
       [](const torch::jit::Node *node) {
-        return [node](torch::jit::Stack &stack) {
-          llvm::Error err =
-              CachingGraphRunner::getGraphRunner()->runGraph(node, stack);
+        std::shared_ptr<torch::jit::Graph> graph = node->g(at::attr::Subgraph);
```
yinghai: When is this shared_ptr released? Right after the op run is done?

jackm321: Once the OperationCreator is called, the Operation get's registered with the OperatorRegistry, but from reading that code I don't see any removal and [this comment ](https://github.com/pytorch/pytorch/blob/c6617b370b903b74a5647dd34e69fd0081713df8/torch/csrc/jit/operator.h#L35-L63)kind of suggests that things go in there forever so may be this does not go away when the graph is destroyed. If anyone knows of a better way to manage the lifetime of CachingGraphRunners let me know.

yinghai: Yeah, after reading the dispatching a bit I think this should work, as `std::shared_ptr<torch::jit::Graph> graph` created here will be persistent.

putivsky:
LGTM

PR #3456, opened 2019-08-23
[torch_glow] Implement clamp

bertmaher:
Test Plan: new node test, `test_clamp`

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -252,6 +252,15 @@ struct MaxPoolInputs {
   };
 };
 
+/// Indices of aten::clamp inputs.
+struct ClampInputs {
+  enum {
+    input = 0,
+    min = 1,
+    max = 2,
+  };
+};
+
 /// Indexes of aten::adaptive_avg_pool2d inputs.
 struct AdaptiveAvgPoolInputs {
   enum {
@@ -385,6 +394,12 @@ PyTorchModelLoader::getSymbolsMapping() {
        {{"aten::max"}, &PyTorchModelLoader::loadMax, {}},
        {{"aten::exp"}, &PyTorchModelLoader::loadExp, {}},
        {{"aten::sqrt", "aten::sqrt_"}, &PyTorchModelLoader::loadSqrt, {}},
+       {{"aten::clamp"},
+        &PyTorchModelLoader::loadClamp,
+        {
+            ClampInputs::min,
+            ClampInputs::max,
+        }},
        {{"aten::size"}, &PyTorchModelLoader::loadSize, {SizeInputs::dim}},
        // TODO: use -1 to freeze all inputs
        {{"prim::ListConstruct"}, &PyTorchModelLoader::loadListConstruct, {}},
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1209,6 +1224,35 @@ llvm::Error PyTorchModelLoader::loadAvgPool2d(const torch::jit::Node *ptNode) {
...
+
+  auto maxHandle = glow::Handle<float>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      maxHandle, getGlowConstantHandle<float>(inputs[ClampInputs::max]));
+  RETURN_ERR_IF_NOT(
+      maxHandle.size() == 1,
+      glow::strFormat("max size must be 1, got %lu", maxHandle.size()));
+  auto max = maxHandle.raw(0);
+
+  auto output = F_.createClip("clip", input, min, max);
```
bertmaher: Note I'm assuming min/max are constant, which PyTorch decidedly does not guarantee.  This seems like one more instance of the general caching problem so I just figured I'd not worry about it for the time being...

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1209,6 +1224,35 @@ llvm::Error PyTorchModelLoader::loadAvgPool2d(const torch::jit::Node *ptNode) {
...
+  ASSIGN_VALUE_OR_RETURN_ERR(input,
+                             getGlowNodeValue(inputs[ClampInputs::input]));
+
+  auto minHandle = glow::Handle<float>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      minHandle, getGlowConstantHandle<float>(inputs[ClampInputs::min]));
+  RETURN_ERR_IF_NOT(
+      minHandle.size() == 1,
+      glow::strFormat("min size must be 1, got %lu", minHandle.size()));
+  auto min = minHandle.raw(0);
```
jackm321: The amount of code that this takes for each op makes me kinda want to generalize the loader to include some kinda union type also to match the inputs (like something that could be a glow nodevalue or an int or int[] etc)

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Yeah, do not worry about it here.

torch_glow/tests/nodes/clamp_test.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_clamp():
+    """Test of the PyTorch min Node on Glow."""
+
+    def test_f(x):
+        return x * torch.clamp(x + 3.0, 0.0, 6.0) * (1.0 / 6.0)
+
+    jitVsGlow(test_f, torch.randn(7), expected_fused_ops={
+              "aten::add", "aten::mul", "aten::clamp"})
```
jackm321: why add all of these instead of just clamp?

torch_glow/tests/nodes/clamp_test.py
```diff
@@ -0,0 +1,15 @@
...
+import torch
+
+from tests.utils import jitVsGlow
+
+
+def test_clamp():
+    """Test of the PyTorch min Node on Glow."""
+
+    def test_f(x):
+        return x * torch.clamp(x + 3.0, 0.0, 6.0) * (1.0 / 6.0)
```
jackm321: what's the point of the ` * (1.0 / 6.0)`?

torch_glow/tests/nodes/clamp_test.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+
+from tests.utils import jitVsGlow
+
+
+def test_clamp():
+    """Test of the PyTorch min Node on Glow."""
```
jackm321: should be clamp Node

torch_glow/tests/nodes/clamp_test.py
bertmaher: Oh lol, sorry, I wanted this for an implementation of hswish I was working on in another place, and just copied that over as a unit test.  Let me change it.

PR #3458, opened 2019-08-23
[torch_glow] Add __init__.py to tests/ subdirs

bertmaher:
This is sort of a request-for-help more than a real PR.  Basically, I wanted to use `pytest` from inside `torch_glow` to run a subset of tests I care about, like `pytest -k test_clamp`.  (I could also do `python setup.py test --addopts "-k test_clamp"` but that's kind of a lot of typing.)

Buuuut, without this diff, doing `pytest` gives me the following inscrutable error:
```
====================================================== ERRORS =======================================================
___________________________________ ERROR collecting tests/training/init_test.py ____________________________________
import file mismatch:
imported module 'tests.training.init_test' has this __file__ attribute:
  /Users/bertrand/miniconda3/envs/pytorch/lib/python3.7/site-packages/torch_glow-0.0.0-py3.7-macosx-10.9-x86_64.egg/tests/training/init_test.py
which is not the same as the test file we want to collect:
  /Users/bertrand/src/glow/torch_glow/tests/training/init_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
============================================== 1 error in 0.89 seconds ==============================================
```

If I add these __init__.py files, the problem goes away.  (Also, if I remove tests/training/__init__.py, the problem goes away, but I assume it was put there for a reason?)

I don't really know how to Python properly, so hopefully someone can either (a) explain this to me, or (b) stamp this diff to appease the evil spirits haunting my dev environment.


putivsky:
Did you try to follow the advice in error description?
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules

bertmaher:
Yeah, that did not seem to help.  Apparently this has happened to other people around the internet, and I was pointed here, which iat least helps me understand the expected layout: http://doc.pytest.org/en/latest/goodpractices.html.  But it still didn't explain exactly what was going wrong in this particular case.

jackm321:
Lol going to choose b to exorcise our dev environments

Issue #2954, opened 2019-05-22
Need support for Deconvolution (Transposed convolution, ONNX ConvTranspose) node


vuzelac-cadence:
Want to discuss the approach for supporting Deconvolution, needed for upsampling in some segmentation networks.

I plan to add support for ONNX ConvTranspose  which nicely maps onto our backend Deconvolution. But not sure what that node should map to  in CPU/Interpreter

One approach is to reuse ConvolutionGrad node (and it's libjit implementation) and just skip unrelated things (like bias gradient) by adding a flags. However, I am not sure about the performance of this implementation - Convolution is supported in various libraries and leaving this one with only the default implementation seems slow. 

Alternatively, input can be prepared to perform regular convolution (slicing and scattering of input, transposing weights and what not), which again is not optimal just like the above.

We don't care much about any efficient implementation in CPU/Interpreter (other the the simulation speed during development) but wanted to hear if you have any plans here.

PR #3464, opened 2019-08-27
ConvTranspose ONNX/Caffe2 operator. CPU/Interpreter float support. No dilation support. Not all padding options supported.

vuzelac-cadence:
Summary:
This adds support for ConvTranspose ONNX and Caffe2 operators and the CPU/Interpreter float implementations.
Quantized/Fixed-point implementation is missing.
Dilation/Group is not supported.
Checking if kernel is too small/large is not done.
Extensive verification of the node inputs/arguments is not done.

Note:
We needed limited Deconvolution support in CPU float for our needs. It lacks a lot. Presented here in case in case such limited support is useful in Glow.

[Optional Fixes #issue]
#2954


vuzelac-cadence:
@jfix71 ,   ConvTranspose can been lowered into Convolution (for our backend we would skip that). I've added  OperatorTest  deconvConvCompare  function that does such conversion although not extensively (does filter transposing, pads conversion etc.).
First thing I want to ask if you are really targeting ConvTranspose lowering onto Convolution ? If so, this whole thing would have to be changed so might not be worth discussing the details of the current implementation.

jfix71:
>First thing I want to ask if you are really targeting ConvTranspose lowering onto Convolution ? 

@vuzelac-cadence Potentially, yes -- I'm curious what downsides you see in doing so. It seems like we can make all of the necessary transformations for the ConvTranspose to Conv at compile time, correct? If so then I don't see why not -- we could reuse all of the Conv infrastructure we have, e.g. optimizations + kernels for all backends.

vuzelac-cadence:
ConvTranspose can have large kernel which dedicated processors might not be supporting so special care is needed when implementing this op. Our backend would need to skip lowering the ConvTranspose operators, thus the use of existing optimizations is not feasible without further examining.  But this is the same as in the current approach - need dedicated optimizations.
ConvTranspose also has some special optimizations in case it comes from TF, where it's represented using ConvTransposeBackProp op followed by BiasAdd. 

Why I took this path - I don't know :(, seemed easier at some moment, looking from our backend perspective. Looking broader, ConvTranspose probably should be lowered onto Convolution for CPU and Interpreter backends. Maybe a flag in Conv node should indicate deconvolution. All in all, I'll close this PR is it needs work.

vuzelac-cadence:
@jfix71 , we would like to do a bit more upstream as our rebasing is getting harder and harder. Getting back to this old issue and the ability to represent ConvTranspose as Convolution. This transformation requires input to be padded and dilated. E.g. for input 2x2:
```
1 2 
3 4
```
Stride 2 ConvTranspose requires input to become this:
```
p p p p p p p
p p p p p p p
p p 1 0 2 p p
p p 0 0 0 p p
p p 3 0 4 p p
p p p p p p p
p p p p p p p
```

Number of pads (p) will depend on input, output, kernel, and deconv pads sizes but anyway, overall, Deconv to Conv folding requires ~ Padding + Scatter. We cannot afford inefficient CPU implementation as we use it heavily (our customers too).  Unless I missed to see something, would the original implementation work ? (I'll fix other TODOs).

jfix71:
@vuzelac-cadence I think the current implementation was mostly fine -- I was mostly curious about the choice of not lowering to Conv. Do you think it would make sense to lower to a Conv for the Interpreter backend at least? I suppose the only reason we should implement lowering logic here is if we expect any backend to want to also use the lowering logic -- otherwise it's a novelty that perhaps isn't so useful, as other backends would also probably benefit from seeing a reference implementation.

vuzelac-cadence:
@jfix71 , what would be the reason to lower for Interpreter, when basic deconvolution is just a simple loop. It is easier to implement - it's a few for loops and as I mentioned, should be faster vs dilating the input + software convolution.
And it's not that it saves us from introducing the Deconvolution node.

It is actually possible that a backend could use such lowered convolution, in theory. For example, we have even more efficient implementation using multiple interleaved convolutions, but ok, some backend would work with one. Such backend should be able support large kernels too. All possible, but still, I I think they all will tweak things to the point that this generic lowering won't work. That is, they all we want to see Deconvolution, as they'll have special handling for it.

jfix71:
I think the main potential benefit for some HW accelerators could be if the accelerator has a highly specialized PE for handling convolutions, but must fall back to less efficient HW (like CPU) for other ops, and the deconv may not be supported directly. I agree this may be too low level to expose outside the backend level, though, but no strong feeling.

vuzelac-cadence:
@jfix71,  I brought up this lowering example just to show how ConvTranspose maps onto a single Conv. But I really don't see what would be the lowering that could be called "generic". To begin with us - such lowering is not needed as we do the lowering differently.  Our lowering quite possibly won't be the best one for some other hardware.

lib/Graph/Graph.cpp
```diff
@@ -607,6 +607,17 @@ static void checkKernelSize(ShapeNHWC idim, llvm::ArrayRef<unsigned_t> kernels,
          "Kernel size is too large");
 }
 
+/// Check the kernel size for Deconvolution ops.
+static void checkDeconvKernelSize(ShapeNHWC idim,
```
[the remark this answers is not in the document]
vuzelac-cadence: More checks in verify(). This function is removed.

vuzelac-cadence:
@jfix71 , ready for review.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -3871,3 +3871,82 @@ TEST_F(GraphOptz, SinkClipBelowReshape) {
...
+  EXPECT_EQ(F->getNodes().size(), 2);
+
+  {
+    ConvTransposeNode *DN = llvm::dyn_cast<ConvTransposeNode>(S->getInput());
+    EXPECT_TRUE(DN);
+
+    Constant *CN = llvm::dyn_cast<Constant>(DN->getBias());
+    EXPECT_TRUE(CN);
+
+    auto BH = CN->getPayloadMutable().getHandle();
```
jfix71: nit: should be able to use `getPayload()` here instead of Mutable.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -3871,3 +3871,82 @@ TEST_F(GraphOptz, SinkClipBelowReshape) {
...
+    EXPECT_EQ(BH.raw(0), 1 + 1);
```
jfix71: Should we check all of the values instead of the first 5?

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+// 2-ch output.
+TEST_P(OperatorTest, sanityDeconvolution2OutCh) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
+
+  float biasVal[2] = {1.1, 2.2};
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 2, 2, 1}, "input", false);
+  bindings_.allocate(input)->getHandle() = {2., 3., 4., 5.};
+
+  auto filter =
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+      mod_.createPlaceholder(ElemKind::FloatTy, {2, 3, 3, 1}, "filter", false);
+  bindings_.allocate(filter)->getHandle() = {2., 3., 4.,  5., 6.,  7.,
+                                             8., 9., 10., 3., 4.,  5.,
+                                             6., 7., 8.,  9., 10., 11.};
+
+  auto bias = mod_.createPlaceholder(ElemKind::FloatTy, {2}, "bias", false);
```
jfix71: nit: `auto *bias`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+TEST_P(OperatorTest, sanityDeconvolution1OutCh) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
+
+  float biasVal = 2.4;
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 2, 2, 3}, "input", false);
+  bindings_.allocate(input)->getHandle() = {2., 3., 4., 3., 4., 5.,
+                                            4., 5., 6., 5., 6., 7.};
+
+  auto filter =
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 3, 3, 3}, "filter", false);
+  bindings_.allocate(filter)->getHandle() = {
+      2., 3., 4., 3., 4., 5., 4.,  5., 6.,  5.,  6.,  7.,  6., 7.,
+      8., 7., 8., 9., 8., 9., 10., 9., 10., 11., 10., 11., 12.};
+
+  auto bias = mod_.createPlaceholder(ElemKind::FloatTy, {1}, "bias", false);
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+// Deconvolution with easy to calculate output values. 2x2 input, 3-ch input,
+// 1-ch output.
+TEST_P(OperatorTest, sanityDeconvolutionStrided) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
+
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 2, 2, 1}, "input", false);
+  bindings_.allocate(input)->getHandle() = {2., 3., 4., 5.};
+
+  auto filter =
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 3, 3, 1}, "filter", false);
+  bindings_.allocate(filter)->getHandle() = {2., 3., 4., 5., 6.,
+                                             7., 8., 9., 10.};
+
+  auto bias = mod_.createPlaceholder(ElemKind::FloatTy, {1}, "bias", false);
```
jfix71: nit: `auto *bias`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+// Deconvolution with easy to calculate output values. 2x2 input, 3-ch input,
+// 1-ch output.
+TEST_P(OperatorTest, sanityDeconvolutionPads) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
+
+  auto *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 2, 2, 1}, "input", false);
+  bindings_.allocate(input)->getHandle() = {2., 3., 4., 5.};
+
+  auto filter =
```
jfix71: nit: `auto *filter`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 3, 3, 1}, "filter", false);
+  bindings_.allocate(filter)->getHandle() = {2., 3., 4., 5., 6.,
+                                             7., 8., 9., 10.};
+
+  auto bias = mod_.createPlaceholder(ElemKind::FloatTy, {1}, "bias", false);
```
jfix71: nit: `auto *bias`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+  llvm::SmallVector<unsigned_t, 4> pads = {pad, pad, pad, pad};
+  llvm::SmallVector<unsigned_t, 4> Cpads = {Cpad, Cpad, Cpad, Cpad};
+  llvm::SmallVector<unsigned_t, 2> kernels = {kernel, kernel};
+  llvm::SmallVector<unsigned_t, 2> strides = {stride, stride};
+
+  auto *input = mod.createPlaceholder(ElemKind::FloatTy, {1, idim, idim, 1},
+                                      "input", false);
+  bindings.allocate(input)->getHandle().randomize(-10.0, 10.0, mod.getPRNG());
+
+  auto filterConv = mod.createPlaceholder(
```
jfix71: nit: `auto *filterConv`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+      ElemKind::FloatTy, {1, kernel, kernel, 1}, "filterC", false);
+  bindings.allocate(filterConv)
+      ->getHandle()
+      .randomize(-10.0, 10.0, mod.getPRNG());
+  auto FCH = bindings.get(filterConv)->getHandle();
+
+  auto filterDeconv = mod.createPlaceholder(
```
jfix71: nit: `auto *filterDeconv`

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
   }
 }
 
+// Deconvolution with easy to calculate output values. 2x2 input, 1-ch input,
```
jfix71: Think we usually use `///` on these tests. Here and all others.

tests/unittests/OperatorTest.cpp
```diff
@@ -6057,13 +6057,335 @@ TEST_P(OperatorTest, Fp16Splat) {
...
+// 2-ch output.
+TEST_P(OperatorTest, sanityDeconvolution2OutCh) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71: Can we use `CHECK_IF_ENABLED()`? Here and all others. Requires updates to blacklists.

include/glow/Graph/Nodes.h
```diff
@@ -231,6 +231,27 @@ inline ShapeHWD calculate3DConvPoolOutputDims(
...
+  ShapeHW kdim(kernels);
+  ShapeHW sdim(strides);
+
+  assert((dilation == 1) &&
+         "Deconvolution output calculation doesn't support dilation");
+
+  size_t outsx = (sx - 1) * sdim.height + (kdim.height - 1) * (dilation - 1) -
+                 pdim.top - pdim.bottom + kdim.height;
+  size_t outsy = (sy - 1) * sdim.width + (kdim.width - 1) * (dilation - 1) -
+                 pdim.left - pdim.right + kdim.width;
```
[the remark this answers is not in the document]
vuzelac-cadence: I didn't want to drop it from calculation as they should be correct but are missing tests; I don't have and need for dilation thus didn't verify it. But if somebody is to add it later, maybe it'll be easier if the calculation is still there. Please confirm - I sure can delete.

vuzelac-cadence:
About the switch to ConvTranspose name - should I do it in the tests too ? (prefer not to.).  Also, didn't get why ONNX Importer should be ok with "Deconvolution" name ?

include/glow/Graph/Nodes.h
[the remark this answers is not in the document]
jfix71: Ok, no big deal, you can leave it.

jfix71:
>About the switch to ConvTranspose name - should I do it in the tests too ? (prefer not to.). Also, didn't get why ONNX Importer should be ok with "Deconvolution" name ?

@vuzelac-cadence I think we should name it consistently across the whole codebase. PyTorch, Caffe2, and ONNX all call this ConvTranspose, so I'd prefer to stay consistent with all of them. (I initially thought ONNX called it deconv but I was mistaken).

tests/unittests/GraphOptzTest.cpp
vuzelac-cadence: This is checking input bias only. For output I've added `checkNumericalEquivalence`

vuzelac-cadence:
@jfix71, `checkNumericalEquivalence ` sometimes fails for `FoldConvTransposeAddIntoBiasAddLHS`
This is one example:

```
I0207 13:09:18.175065 39247 Tensor.h:885] Tensors not equal: 1 out of 250 elements exceeded allowed error threshold 0.0001. Maximum error found was 16 at index 2: 2.25113e+08 vs. 2.25113e+08
I0207 13:09:18.175168 39247 Tensor.h:891] Maximum relative error found was: 3.55377e-08 at index: 2: 2.25113e+08 v.s. 2.25113e+08
/home/vuzelac/glow/github/glow/tests/unittests/BackendTestUtils.h:183: Failure
Value of: PlaceholderBindings::compare(&originalBindings, &optimizedBindings, allowedError)
```
Maximum relative error: **3.55377e-08**, but allowed error threshold is **0.0001**. Something is not right in the checker ?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -3871,3 +3871,90 @@ TEST_F(GraphOptz, SinkClipBelowReshape) {
...
+  EXPECT_EQ(BH.raw(4), 5 + 9);
+
+  bindings.allocate(mod.getPlaceholders());
+  bindings.get(input)->getHandle().randomize(-1.0, 1.0, mod.getPRNG());
+}
+
+/// Test that Add after ConvTranspose is folded into Bias add when the actual
+/// Add is is a broadcast of the bias.
+TEST_F(GraphOptz, FoldConvTransposeAddIntoBiasAddRHS) {
+  foldConvTransposeAddIntoBiasAdd(bindings_, mod_, F_, optimizedF_, 0);
```
jfix71: nit: `/* RnL */ false` instead of `0`

tests/unittests/GraphOptzTest.cpp
```diff
@@ -3871,3 +3871,90 @@ TEST_F(GraphOptz, SinkClipBelowReshape) {
...
+  checkNumericalEquivalence();
+}
+TEST_F(GraphOptz, FoldConvTransposeAddIntoBiasAddLHS) {
+  foldConvTransposeAddIntoBiasAdd(bindings_, mod_, F_, optimizedF_, 1);
```
jfix71: nit: `/* RnL */ true` instead of `1`

vuzelac-cadence:
Turns out filter was not initialized - was causing some really high float values in the output. Those values did differ with extremely small **relative** error but I failed to see that "relative" keyword. I think this is all fine.

PR #3470, opened 2019-08-28
Parsing PyTorch model and loading into Glow directly.

putivsky:
Summary:
Use JIT Module API to get lowered graph and load it directly into Glow without registered Fusion passes.
Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/torch_glow/src/PyTorchFileLoader.cpp
+++ b/torch_glow/src/PyTorchFileLoader.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "PyTorchFileLoader.h"
+#include "FuseLinear.h"
 #include "PyTorchModelLoader.h"
 #include "glow/Support/Error.h"
 #include "glow/Support/Support.h"
@@ -96,7 +97,7 @@ loadJitGraphToGlowFunction(torch::jit::Stack &stack, torch::jit::Graph &graph,
 /// Glow function is set.
 llvm::Error
 evaluateModuleGraph(std::shared_ptr<torch::jit::script::Module> &module,
-                    std::vector<torch::jit::IValue> &inputs) {
+                    const std::vector<torch::jit::IValue> &inputs) {
   try {
     module->forward(inputs);
   } catch (const std::exception &x) {
@@ -131,7 +132,8 @@ struct RegisterCustomFusionPass {
 
     torch::jit::RegisterPass pass([&](std::shared_ptr<torch::jit::Graph> &g) {
       // Trigger custom fusion pass only if local thread Glow Function is set.
-      if (localFusionInfo.function) {
+      if (localFusionInfo.settings &&
+          localFusionInfo.settings->fusionPassEnabled) {
         glowCustomFuse(g, getFusionSymbol());
       }
     });
@@ -155,12 +157,15 @@ llvm::Error PyTorchFileLoader::loadPyTorchModel(
 
 /*static*/
 llvm::Error PyTorchFileLoader::loadPyTorchGraph(
-    const std::string &fileName, std::vector<torch::jit::IValue> &inputs,
+    const std::string &fileName, const std::vector<torch::jit::IValue> &inputs,
     glow::Function &F, std::vector<glow::Placeholder *> &inputPlaceholders,
-    std::vector<glow::Placeholder *> &outputPlaceholders,
-    const PyTorchLoaderSettings &settings, bool sanityCheck) {
+    std::vector<glow::Placeholder *> &outputPlaceholders, bool sanityCheck) {
   // Register custom pass once.
   static RegisterCustomFusionPass fuser;
+
+  PyTorchLoaderSettings settings;
+  settings.fusionPassEnabled = true;
+  settings.weightFreezingEnabled = false;
 
   // Convert PyTorch model into JIT Module.
   std::shared_ptr<torch::jit::script::Module> module;
@@ -177,6 +182,27 @@ llvm::Error PyTorchFileLoader::loadPyTorchGraph(
   RETURN_IF_ERR(err);
 
   return sanityCheck ? performSanityCheck() : llvm::Error::success();
+}
+
+/*static*/
+llvm::Error PyTorchFileLoader::parsePyTorchGraph(
+    const std::string &fileName, const std::vector<torch::jit::IValue> &inputs,
+    glow::Function &F, std::vector<glow::Placeholder *> &inputPlaceholders,
+    std::vector<glow::Placeholder *> &outputPlaceholders) {
+
+  // Convert PyTorch model into JIT Module.
+  std::shared_ptr<torch::jit::script::Module> module;
+  RETURN_IF_ERR(PyTorchFileLoader::loadPyTorchModel(fileName, module));
+
+  auto method = module->get_method("forward");
+  auto graphAndTensors = method._lowered_graph();
+
+  FuseLinear(graphAndTensors.first);
+
+  // Parse JIT Graph and load into Glow Function.
+  return PyTorchModelLoader::parseJITGraph(
+      F, *graphAndTensors.first, inputs, graphAndTensors.second,
+      inputPlaceholders, outputPlaceholders);
 }
 
 // Sanity check, after "fusionSymbol" node, not other nodes should exist.
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -356,9 +356,10 @@ glow::Type PyTorchModelLoader::ptTypeToGlowType(const c10::TensorType &ptType) {
 /// \returns a Glow tensor with the same type and underlying data as the given
 /// PyTorch tensor \p ptT.
 llvm::Expected<glow::Tensor>
-PyTorchModelLoader::ptTensorToGlowTensor(const at::Tensor &ptT) {
+PyTorchModelLoader::ptTensorToGlowTensor(const at::Tensor &ptT, bool copy) {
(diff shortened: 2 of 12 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1613,4 +1643,93 @@ PyTorchModelLoader::PyTorchModelLoader(
...
+      }
+      inputPlaceholders.push_back(PH);
+    } else {
+      error =
+          MAKE_ERR(glow::strFormat("Only tensor type is supported for now."));
+      return;
+    }
+  }
+
+  for (size_t i = 0; i < parameters.size(); ++i, ++graphIdx) {
```
jackm321: We shouldn't have to duplicate this constructor, we could just have parameters be empty by default and everything else is exactly the same right?

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -132,6 +135,15 @@ class PyTorchModelLoader {
                std::vector<glow::Placeholder *> &outputPlaceholders,
                const PyTorchLoaderSettings &settings);
 
+  /// Takes a glow::Function \p F, a jit::Graph \p subgraph to load, and a
+  /// stack of the graph \p inputs. \returns error on failure.
+  static llvm::Error
```
jackm321: \p parameters?

torch_glow/src/training/TorchGlowTraining.cpp
```diff
@@ -166,17 +143,24 @@ llvm::Error TorchGlowTraining::train(const Tensor &samples,
...
 
   // Sanity check.
   const bool isSlice = isValidSamplesForSliceInput(*input, samples) &&
                        isValidSamplesForSliceInput(*label, labels);
 
   if (!isSlice && (!isValidSamplesForSingleInput(*input, samples) ||
                    !isValidSamplesForSingleInput(*label, labels))) {
-    RETURN_ERR("Invalid samples/labels dimensions.");
+    std::string O;
+    llvm::raw_string_ostream sink(O);
```
jackm321: use strFormat()

torch_glow/src/training/TorchGlowTraining.cpp
```diff
@@ -238,22 +222,22 @@ bool TorchGlowTrainingWrapper::init(const std::string &modelPath,
...
+      modelPath, ptInputs, backend, parameters_, config_,
       randomizeWeights ? TorchGlowTraining::RandomizeWeights::YES
                        : TorchGlowTraining::RandomizeWeights::NO));
 }
 
 bool TorchGlowTrainingWrapper::train(const at::Tensor &ptSamples,
                                      const at::Tensor &ptLabels) {
   llvm::Expected<glow::Tensor> glowSamples =
-      PyTorchModelLoader::ptTensorToGlowTensor(ptSamples);
+      PyTorchModelLoader::ptTensorToGlowTensor(ptSamples, false /*copy*/);
```
jackm321: I think this is usually written like `/*copy*/ false`

torch_glow/src/PyTorchFileLoader.cpp
```diff
@@ -157,11 +159,14 @@ llvm::Error PyTorchFileLoader::loadPyTorchModel(
...
     const std::string &fileName, std::vector<torch::jit::IValue> &inputs,
     glow::Function &F, std::vector<glow::Placeholder *> &inputPlaceholders,
-    std::vector<glow::Placeholder *> &outputPlaceholders,
-    const PyTorchLoaderSettings &settings, bool sanityCheck) {
+    std::vector<glow::Placeholder *> &outputPlaceholders, bool sanityCheck) {
   // Register custom pass once.
   static RegisterCustomFusionPass fuser;
 
+  PyTorchLoaderSettings settings;
+  settings.fusionPassEnabled = true;
```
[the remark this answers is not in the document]
putivsky: I would keep the fusion approach for a while, till the whole end-to-end training pipe-line will be built.

torch_glow/src/PyTorchModelLoader.cpp
putivsky: there is no settings as well.

torch_glow/src/training/TorchGlowTraining.cpp
putivsky: Does it work for glow Type objects?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1613,4 +1643,93 @@ PyTorchModelLoader::PyTorchModelLoader(
...
+    std::vector<glow::Placeholder *> &inputPlaceholders,
+    std::vector<glow::Placeholder *> &outputPlaceholders) {
+  llvm::Error error = llvm::Error::success();
+  MARK_ERR_CHECKED(error);
+  PyTorchModelLoader loader(F, graph, inputs, parameters, inputPlaceholders,
+                            outputPlaceholders, error);
+  return error;
+}
+
+PyTorchModelLoader::PyTorchModelLoader(
```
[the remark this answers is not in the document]
putivsky: it's cpp file, what kind doxygen comments should I add?

torch_glow/src/PyTorchModelLoader.cpp
putivsky: Refactored, I think having two constructors is ok, they have different parameters and do different things with some overlapped (I have moved common code into the separate methods).

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1613,4 +1643,93 @@ PyTorchModelLoader::PyTorchModelLoader(
   error = llvm::Error::success();
 }
 
+/// Takes a glow::Function \p F, a jit::Graph \p subgraph to load, detects
+/// inputs for the subgraph automatically. \returns error on failure.
+/*static*/
+llvm::Error PyTorchModelLoader::parseJITGraph(
```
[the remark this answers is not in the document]
putivsky: removed comments from cpp file.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1536,6 +1565,36 @@ PyTorchModelLoader::createGlowConstantFromIValue(
...
+  }
+  return llvm::Error::success();
+}
+
+llvm::Error PyTorchModelLoader::createSaveNodeAndOutputPlaceholders(
+    const torch::jit::Graph &graph,
+    std::vector<glow::Placeholder *> &outputPlaceholders) {
+  for (const torch::jit::Value *output : graph.outputs()) {
+    glow::NodeValue outputNodeValue;
+    if (auto resOrErr = getGlowNodeValue(output)) {
```
jackm321: ASSIGN_VALUE_OR_RETURN_ERR

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1572,12 +1632,9 @@ PyTorchModelLoader::PyTorchModelLoader(
     const torch::jit::Value *inputValue = graphInputValues[i];
     const c10::IValue inputIValue = inputs.at(i);
     if (inputIValue.isTensor()) {
-      const auto ptTensorType = c10::TensorType::create(inputIValue.toTensor());
       glow::Placeholder *PH;
-      if (auto resOrErr = loadValue(inputValue, *ptTensorType)) {
-        PH = std::move(*resOrErr);
-      } else {
-        error = resOrErr.takeError();
+      if ((error = createPlaceholderFromTensor(inputValue, PH,
```
jackm321: RETURN_IF_ERR

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1598,19 +1655,75 @@ PyTorchModelLoader::PyTorchModelLoader(
...
+    return;
+  }
+
+  size_t graphIdx = 0;
+  for (size_t i = 0; i < inputs.size(); ++i, ++graphIdx) {
+    const torch::jit::Value *inputValue = graphInputValues[graphIdx];
+    const c10::IValue inputIValue = inputs.at(i);
+    if (inputIValue.isTensor()) {
+      glow::Placeholder *PH;
+      if ((error = createPlaceholderFromTensor(inputValue, PH,
```
jackm321: RETURN_IF_ERR

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1536,6 +1565,36 @@ PyTorchModelLoader::createGlowConstantFromIValue(
   return F_.getParent()->createConstant("constant", std::move(t));
 }
 
+llvm::Error
+PyTorchModelLoader::createPlaceholderFromTensor(const torch::jit::Value *value,
+                                                glow::Placeholder *&PH,
+                                                const at::Tensor &ptTensor) {
```
jackm321: don't need tensor, just neet the tensor type. Why not just create the tensor type then call loadValue() like before? This is just adding a new method to do that

putivsky: because in each case we start with tensor and will call tensor -> tensor shape every time.

torch_glow/src/PyTorchModelLoader.cpp
putivsky: It's a constructor, return error is not allowed.

torch_glow/src/PyTorchModelLoader.cpp
putivsky: It's a constructor, return error is not allowed.

PR #3474, opened 2019-08-30
dump graph before compiling

hyuen:
Summary:
buck test //glow/fb/test:test_numerics_xxx -- limits

dot 1.dot -T png -o 1.png

Reviewed By: rdzhabarov

Differential Revision: D16978685


lib/Onnxifi/HostManagerOnnxifi.cpp
```diff
@@ -27,6 +27,7 @@ bool GlowDumpDebugTraces = false;
 bool GlowSaturateHost = false;
 bool GlowFP16 = false;
 bool GlowClipFP16 = false;
+bool GlowDumpGraph = false;
```
rdzhabarov: hm, I'm not sure we need to have it here now (once it's used by some open sourced backend we could add that).

PR #3479, opened 2019-08-30
[PyTorch] Create a union type GlowIValue for loading values from PyTorch

jackm321:
Summary:
* Add a type which maps closely to torch::jit::IValue but holds Glow tensors. This is a departure from the previous approach which is to just cast everything as Glow Constants.
* Separate values loaded from PyTorch into two categories, Tensors for which only shapes are known at load time, and all other values (ints, lists, etc) for which values can be reasoned about at load time. For each of these categories, the information known at load-time will eventually be guarded so that, for example if a tensor's shape changes or any other values's contents change, the graph will not be valid and reloading will happen
* This also adds most of the necessary ingredients to make loading quantized graphs more simple

Documentation:
doxygen
Will add some information to pytorch.md about how this works once it's settled

Test Plan:
CI


torch_glow/src/GlowIValue.cpp
```diff
@@ -0,0 +1,281 @@
...
+void GlowIValue::fromTuple(std::vector<GlowIValue> glowIValList) {
+  reset();
+  tag_ = Tag::Tuple;
+  payload_.asTuple = new std::vector<GlowIValue>;
+  std::swap(glowIValList, *payload_.asTuple);
+}
+
+llvm::Error GlowIValue::fromIValue(const at::IValue &ival) {
+  reset();
+  if (ival.isNone()) {
```
opti-mix: PyTorch does not provide something like `getTag()` so that one could write a switch statement?

torch_glow/src/GlowIValue.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+namespace glow {
+
+namespace {
+/// \returns a Glow tensor with the same type and underlying data as the given
+/// PyTorch tensor \p ptT.
+// TODO: don't redefine this
+llvm::Expected<glow::Tensor> ptTensorToGlowTensor2(const at::Tensor &ptT) {
+  // TODO: get correct ElemKind.
+  DCHECK_EQ(ptT.scalar_type(), at::kFloat)
```
putivsky: PyTorchModelLoader::convertScalarType method exists.

torch_glow/src/GlowIValue.cpp
```diff
@@ -0,0 +1,281 @@
...
+  }
+  tag_ = Tag::None;
+}
+
+GlowIValue::~GlowIValue() { reset(); }
+
+GlowIValue::GlowIValue(GlowIValue &&other) {
+  std::swap(tag_, other.tag_);
+  std::swap(payload_, other.payload_);
+}
```
bertmaher: Instead of swapping you could just set other.tag_=None.  It's kind of a pico-optimization though, I've just read [too many](https://www.quora.com/What-are-the-advantages-of-using-C-over-C-1/answer/Keith-Adams) [Scott Meyers articles](http://scottmeyers.blogspot.com/2014/06/the-drawbacks-of-implementing-move.html).

torch_glow/src/GlowIValue.cpp
```diff
@@ -0,0 +1,281 @@
...
+bool GlowIValue::isTensor() const { return Tag::Tensor == tag_; }
+bool GlowIValue::isDouble() const { return Tag::Double == tag_; }
+bool GlowIValue::isInt() const { return Tag::Int == tag_; }
+bool GlowIValue::isBool() const { return Tag::Bool == tag_; }
+bool GlowIValue::isIntList() const { return Tag::IntList == tag_; }
+bool GlowIValue::isDoubleList() const { return Tag::DoubleList == tag_; }
+bool GlowIValue::isBoolList() const { return Tag::BoolList == tag_; }
+bool GlowIValue::isTensorList() const { return Tag::TensorList == tag_; }
+
+#define ExpectTag(EXPECTED_TAG)                                                \
```
putivsky: Hard to debug macro, can be just inline function.

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -41,8 +79,12 @@ class PyTorchModelLoader {
   /// Mapping from PyTorch Values to Glow NodeValues created during loading.
   using ValueMap =
       std::unordered_map<const torch::jit::Value *, glow::NodeValue>;
+  // TODO: delete this once it's replace by the following 2 maps.
   ValueMap valueMap_;
 
+  std::unordered_map<const torch::jit::Value *, std::unique_ptr<ValueMapping>>
```
putivsky: If ValueMapping class is movable, not need to have a std::unique_ptr wrapper.

torch_glow/src/GlowIValue.cpp
jackm321: Unfortunately they do not, see this comment on why
https://github.com/pytorch/pytorch/blob/f8852c947bf527a994002aa04f8e77b78ac75f01/aten/src/ATen/core/ivalue.h#L435-L440

torch_glow/src/GlowIValue.cpp
jackm321: Yeah that actually makes a lot of sense

torch_glow/src/GlowIValue.cpp
jackm321: but this macro inserts a return

torch_glow/src/GlowIValue.cpp
```diff
@@ -0,0 +1,281 @@
...
+  } else if (ival.isBool()) {
+    fromBool(ival.toBool());
+  } else if (ival.isDoubleList()) {
+    const auto ivalDoubles = ival.toDoubleList();
+    std::vector<double> doubles(ivalDoubles.begin(), ivalDoubles.end());
+    fromDoubleList(doubles);
+  } else if (ival.isIntList()) {
+    const auto ivalInts = ival.toIntList();
+    std::vector<int32_t> ints(ivalInts.begin(), ivalInts.end());
+    fromIntList(ints);
```
jackm321: I forgot to move these

yinghai:
Cool! I think this is heading into the right direction.

jackm321:
Planning on wrapping this up today, there's a few more op loaders to convert and should be good to go

jackm321:
Everything works, need to cleanup then rebase then cleanup some more

torch_glow/src/PyTorchModelLoader.h
jackm321: I did this because ValueMapping hands out pointers to it's internals so those need to remain valid. I will move the unique_ptr to inside ValueMapping

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -61,6 +104,7 @@ const c10::Symbol &getGlowSymbol() {
 
 void glowCustomFuse(std::shared_ptr<torch::jit::Graph> &g,
                     at::Symbol fuseSymbol) {
+
```
yinghai: ?

Issue #3462, opened 2019-08-27
[Optimization] Splitting FC weights to facilitate distribution of compute across multiple computational units


rdzhabarov:
## Problem:
Given an FC operator with the input of [M;K], weights of [K;N] and bias of [N] dimensions.
For certain ASIC architectures it's beneficial to split weights into several chunks and distribute it over multiple
processing units. This way we could fit weights into SRAM, take advantage of parallel execution and reduce overall latency.

## Proposed changes:
Let's consider the scenario when weights are split into two chunks: [K;N/2], [K;N/2] and distributed over two computational units during the compilation time (with the help of partitioner). Bias with the [N/2] sub tensors correspondingly.
In this case:
* `input` activations of size [M;K] needs to be copied to those two processing units during the execution.
* Perform 2 FCs, input [M;K], weights [K;N/2], bias [N/2]
* Concat results of [M;N/2] into the final tensor of size [M;N]

From Glow perspective this could be expressed as a general graph level optimizations (followed by partitioner) which can be enabled/disabled by backends if needed.
Optimization needs to have a parameter specifying number of chunks to split weights and bias (mostly driven by number of computation units).
As well as size of N starting from which the optimization needs to be performed.
Both parameters are architecture dependent and might need to be tuned, for now we could make those as a constant and later figure out the way to set it on per backend basis.

Original idea is coming from @nrsatish, please, add any details/comments if any.
cc: @bertmaher @arunm-git @beicy

PR #3481, opened 2019-08-30
**WIP** [optimization] Perform FC optimization.

rdzhabarov:
There are still some missing pieces, mostly FIXFIXIFIX places in the code.

Summary:
Update this.

Documentation:
doxygen where applicable
will see if i need to update any docs wrt to Optimization passes.

[Optional Fixes #issue]
Addresses optimization part on the https://github.com/pytorch/glow/issues/3462

Test Plan:
Added unit test.


rdzhabarov:
@nrsatish

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -1586,6 +1586,78 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
+      auto *biasSlice =
+          F->createSlice("bias_slice." + bias.getNode()->getName().str(), bias,
+                         {sliceStart}, {sliceStart + elemPerChunk});
+      fcs.push_back(F->createFullyConnected("fc_slice." + FC->getName().str(),
+                                            input, weightSlice->getResult(),
+                                            biasSlice->getResult(), fcType));
+      sliceStart += elemPerChunk;
+    }
+
+    auto *concat = F->createConcat("concat." + FC->getName().str(), fcs, 1);
```
jfix71: nit: `/* dimension */ 1`

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2917,3 +2917,46 @@ TEST_F(GraphOptz, constantFoldWholeFunction) {
...
+
+  EXPECT_EQ(1, countNodeKind(F_, Kinded::Kind::ConcatNodeKind));
+  EXPECT_EQ(12, countNodeKind(F_, Kinded::Kind::FullyConnectedNodeKind));
+
+  auto *concatNode = llvm::dyn_cast<ConcatNode>(save->getInput());
+  ASSERT_TRUE(concatNode);
+  // 12 FCs are connected to the concat node.
+  EXPECT_EQ(12, concatNode->getInputs().size());
+
+  // Check all
```
jfix71: nit: Unfinished comment?

nadavrot:
I like this PR, but I think that we can improve things with management of complexity (considering support for multiple targetts). Let's split this PR into two sections: the policy and mechanism.  Consider having a function that's called 'splitFCs' that accepts some parameters like numOfChunks and minKToSplit instead of the constants.  Now you can call this function  splitFCs(12, 800) from the code. 

The next step would be, how do you decide on the parameters and how you manage the constants for the different backends. I suggest making this a target-specific transformation. The target should decide on the parameters to pass to the transformation function, which needs to be globally accessible in the optimizer. 

@jfix71 @rdzhabarov

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -69,6 +69,9 @@ FunctionPassPipeline glow::createDefaultGraphOptimizationPassPipeline() {
       // Perform Common Subexpression Elimination.
       {FunctionPassID::CSE},
 
+      // Perform vertical split for FC's weights and bias.
+      {FunctionPassID::VerticalFCWeightsSplit},
```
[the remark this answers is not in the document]
rdzhabarov: That sounds good.

tests/unittests/GraphOptzTest.cpp
rdzhabarov: :D correct

rdzhabarov:
> Consider having a function that's called 'splitFCs' that accepts some parameters like numOfChunks and minKToSplit instead of the constants. Now you can call this function splitFCs(12, 800) from the code.

That all makes sense and provides better flexibility.

We briefly talked about this in person I agree with this.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -1586,6 +1586,78 @@ static NodeValue simplifyConcatNode(Function *F, ConcatNode *CN) {
...
+      continue;
+    }
+
+    auto input = FC->getInput();
+    auto weights = FC->getWeights();
+    auto bias = FC->getBias();
+
+    size_t elemPerChunk = (bias.dims()[0] + numOfChunks - 1) / numOfChunks;
+    size_t sliceStart = 0;
+    llvm::SmallVector<NodeValue, numOfChunks> fcs;
```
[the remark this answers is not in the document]
rdzhabarov: standard array was not easy convertible to llvm::ArrayRef which is required in the createConcat call.
Replaced small vector with fixed preallocated std::vector (note, cannot use numOfChunks anymore in the small vector due to it not being constexpr).

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3011,3 +3011,71 @@ llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
...
+    // Input dimension is [M;K] and kept untouched.
+    // Bias dimension is [N], split into chunks.
+    // Weight dimension is [K;N], split into numOfChunks chunks,
+    // [K;N/numOfChunks] each.
+    // Last chunk might require special handling in case
+    // N is not divisible by numOfChunks.
+    auto *fcType = F->getParent()->uniqueTypeWithNewShape(
+        FC->getResult().getType(), {FC->getResult().dims()[0], elemPerChunk});
+
+    for (int i = 0; i < numOfChunks; ++i) {
```
jfix71: nit: `unsigned i` ?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3011,3 +3011,71 @@ llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
...
+                          << minKToSplit;
+
+  bool changed = false;
+  for (auto &node : F->getNodes()) {
+    auto *FC = dyn_cast<FullyConnectedNode>(&node);
+    if (!FC) {
+      continue;
+    }
+
+    int K = FC->getWeights().dims()[1];
```
jfix71: nit: `size_t`

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3011,3 +3011,71 @@ llvm::Error glow::optimizeFunction(Function *F, const Backend &B,
...
+
+bool glow::executeVerticalFCWeightsSplit(Function *F, unsigned numOfChunks,
+                                         unsigned minKToSplit) {
+  DCHECK(numOfChunks > 0) << "numOfChunks must be a positive number, given: "
+                          << numOfChunks;
+  DCHECK(minKToSplit > 0) << "minKToSplit must be a positive number, given: "
+                          << minKToSplit;
+
+  bool changed = false;
+  for (auto &node : F->getNodes()) {
```
jfix71: nit: Perhaps we should iterate over the Nodes backwards? Because we're iterating forward over all the Nodes in the Function, we will iterate over all of the newly added FCs from inside this again. This shouldn't cause infinite loop as is I believe because of the `K < minKToSplit` check. But things could change with this logic in the future. It's also slightly inefficient/unnecessary because we know we won't want to split those FCs again.

PR #3482, opened 2019-08-31
[CPU][Interpreter] Modify CPU and Interpreter DeviceManagers to use constant weight size for function cost

### LINKED ISSUE #3472: CPU and Interpreter DeviceManagers should use ConstantWeightSize for usedMemory.
Currently when a new network is added to the CPU or Interpreter device managers they use a fixed functionCost_ = 1 byte to update usedMemory.
This should be updated to used the size of the constants in the function. 
Once this is done the test added in https://github.com/pytorch/glow/pull/3465 will need to be updated to reflect the size of the constants instead of 1.

SplitInfinity:
**Description**
This commit modifies the CPU and Interpreter `DeviceManagers` to use the
total constant weight size of a function as its cost instead of a
hardcoded value of 1. `DeviceManagerTest` has been modified
accordingly.

**Testing**
All unit tests pass including `DeviceManagerTest`.

**Issues**
This PR fixes #3472.

### DIFF
--- a/lib/Backends/CPU/CPUDeviceManager.cpp
+++ b/lib/Backends/CPU/CPUDeviceManager.cpp
@@ -66,6 +66,8 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
                                       ReadyCBTy readyCB) {
   DCHECK(readyCB != nullptr);
 
+  uint64_t allFunctionsMemoryBytes{0};
+
   // First check for uniqueness of the function name.
   for (const auto &func : functions) {
     if (functions_.count(func.first) != 0) {
@@ -89,9 +91,12 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
                   .str()));
       return;
     }
+
+    allFunctionsMemoryBytes +=
+        func.second->getRuntimeBundle().getConstantWeightSize();
   }
 
-  if (usedMemoryBytes_ + functionCost_ > maxMemoryBytes_) {
+  if (usedMemoryBytes_ + allFunctionsMemoryBytes > maxMemoryBytes_) {
     readyCB(module, MAKE_ERR(GlowErr::ErrorCode::RUNTIME_OUT_OF_DEVICE_MEMORY,
                              "Failed to add network: not enough memory"));
     return;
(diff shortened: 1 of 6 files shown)


lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -89,9 +91,12 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
                   .str()));
       return;
     }
+
+    allFunctionsMemoryBytes +=
```
gcatron: Isn't the runtimeBundle per function? Why do we need the allFunctionsMemoryBytes?

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -116,8 +121,10 @@ void CPUDeviceManager::evictNetworkImpl(std::string functionName,
                                         EvictFunctionCBTy evictCB) {
   DCHECK(evictCB != nullptr);
 
-  if (functions_.erase(functionName)) {
-    usedMemoryBytes_ -= functionCost_; // TODO: static moduleSize
+  auto it = functions_.find(functionName);
+  if (it != functions_.end()) {
+    usedMemoryBytes_ -= it->second->getRuntimeBundle().getConstantWeightSize();
```
gcatron: If we use all functions above and subtract per single function here can we get out of sync?

lib/Backends/CPU/CPUDeviceManager.cpp
SplitInfinity: I may not be up to date on how `addNetwork` is supposed to work, but I thought every function in the provided map `functions` has to be added to the device, so you need to add up the sizes of the constants for each function.

lib/Backends/CPU/CPUDeviceManager.cpp
SplitInfinity: `evictNetwork` only allows you to specify the name of one function to evict. So I don't see any way to avoid getting "out of sync".

lib/Backends/CPU/CPUDeviceManager.cpp
gcatron: You're right, I thought it was outside the loop, looks good.

PR #3484, opened 2019-09-02
[FrontEnd] Add ModelCompiler tool

mciprian13:
**Summary**
ModelCompiler is a more generic front-end tool designed for AOT compilation. It is more generic than the ImageClassifier which (as the name suggests) is tied to image classification applications (assumes only 1 input, for AOT requires an input image to infer the input tensor shape).
The ModelCompiler can build models with any number of inputs:
- for ONNX models the information about the input placeholders is automatically inferred from the model (which contains the description of the input tensors)
- for Caffe2 models (which do not contain the information about the input tensors) the inputs are provided explicitly by the user to the tool as CLI arguments with the following format:
-model-input=name1,type1,shape1   -model-input=name2,type2,shape2    .........
For example:
-model-input=input_1_data,float,[1,1,20]   -model-input=data_bias,float,[10] etc

The model-compiler tool is intended to facilitate the compilation of a model as-is.

**Documentation**
Added section in AOT.md

**Test Plan**
Added unit test for ModelCompiler with ONNX/Caffe2 model with 2 input placeholders.


tools/loader/CMakeLists.txt
```diff
@@ -48,3 +48,20 @@ target_link_libraries(model-runner
...
+                        Backends
+                        Base
+                        Converter
+                        Graph
+                        HostManager
+                        Importer
+                        ExecutionEngine
+                        GraphOptimizer
+                        Quantization
+                        LLVMSupport)
```
opti-mix: Missing new line.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,153 @@
...
+        "the description of the inputs and hence must be provided  \n"
+        "explicitly using this option. One or more model inputs    \n"
+        "are provided using the following format:                  \n"
+        "   -model-input=<inputName1>,<inputType1>,<inputShape1>   \n"
+        "   -model-input=<inputName2>,<inputType2>,<inputShape2>   \n"
+        "   ....................................................   \n"
+        "For example we can can describe 2 inputs as:              \n"
+        "   -model-input=input_03_data,float,[1]                   \n"
+        "   -model-input=data_bias,int32,[1,32,32]                 \n"
+        "The supported types are: float, float16, int32, int64 and \n"
```
opti-mix: Why only this subset of types is supported? May be we should allow for (almost) all types supported by Glow? E.g. quantized types could be interesting if ones uses pre-quantized models.

docs/AOT.md
```diff
@@ -62,6 +62,29 @@ This option supports two modes:
 The second generated file is named `<network_name>.weights` and
 contains the weights required to run the compiled model.
 
+Another tool is the `model-compiler` which is used to compile a model into a bundle.
+This tool is more generic (is not tied just to image classification applications)
```
opti-mix: Question: May be `image-classifier` should be extended to support this new functionality to avoid having two different tools? We could also create a name alias called `model-compiler` for the existing `image-classifier` as it would support more than just image classification.

tools/loader/ModelCompiler.cpp
mciprian13: I chose only these types because they are standard. Currently the models are quantized with Glow only internally (in the middle) while keeping float placeholders (inputs/outputs).
I can also add the quantized types, this would mean adding a format of the type:
-model-input=<name><type>,<shape>                                for non-quantized types
-model-input=<name>,<type>,<scale>,<offset>,<shape>  for quantized types
where <type> is one of "int8q", "int16q", "int32q.
But I don`t know what this be used for.

docs/AOT.md
mciprian13: I think the final (cleanest) solution would be to have 2 tools:
- generic ModelCompiler - compiles any model
- generic ModelRunner - runs any model
I think the image-classifier should be heavily modified to be decoupled from image-classification and even with this modification will not be generic. E.g. a model with multiple inputs, in order to run, should be provided with extra stuff:
- data source for each input (file, stream, etc)
- data decoding method for each input (e.g. encode image/video -> raw, none)
- pre-processing method for each input (normalize [0,1], tokenize)
I think for now we should have them separately until the point when image-classifier will become the ModelRunner I mentioned.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,153 @@
...
+      << "Bundle output directory not provided. Use the -emit-bundle option!";
+
+  // Create the model based on the input model format.
+  std::unique_ptr<ProtobufLoader> LD;
+  if (!loader.getCaffe2NetDescFilename().empty()) {
+    // For Caffe2 format the input placeholder names/types
+    // must be provided explicitly.
+    std::vector<const char *> inputNameRefs;
+    std::vector<TypeRef> inputTypeRefs;
+    for (size_t idx = 0; idx < inputNames.size(); idx++) {
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,153 @@
...
+        "The supported types are: float, float16, int32, int64 and \n"
+        "bool. \n"),
+    llvm::cl::value_desc("int"), llvm::cl::cat(modelCompilerCat));
+
+} // namespace
+
+/// Parse the strings and get the model input names and types.
+static void getModelInputs(std::vector<std::string> &inputNames,
+                           std::vector<Type> &inputTypes) {
+  for (auto str : modelInputs) {
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,153 @@
...
+    // Split string
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,153 @@
...
+    } else if (type.equals("float16")) {
+      kind = ElemKind::Float16Ty;
+    } else if (type.equals("int32")) {
+      kind = ElemKind::Int32ITy;
+    } else if (type.equals("int64")) {
+      kind = ElemKind::Int64ITy;
+    } else if (type.equals("bool")) {
+      kind = ElemKind::BoolTy;
+    } else {
+      CHECK(false) << strFormat("Model input type %s not supported",
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/ModelCompiler.cpp
mciprian13: Per your suggestion I also added the quantized types with the format I mentioned:
-model-input=name,type,scale,offset,shape (type = int8q, int16q, int32q).
I also modified the Caffe2 unit test to use int8 quantized type.

tools/loader/CMakeLists.txt
mciprian13: Done.

docs/AOT.md
opti-mix: Yes, let's keep them separate for now.

docs/AOT.md
```diff
@@ -62,6 +62,29 @@ This option supports two modes:
...
+simply as: 
+  ```
+  $model-compiler -model=<onnx-model-path> -backend=CPU -emit-bundle=<bundle-dir>
+  ```
+- when using Caffe2 models the user must provide explicitly the description of the
+input tensors (which is not part of the model) using the following format:
+  ```
+  $model-compiler -model=<caffe2-model-path> -backend=CPU -emit-bundle=<bundle-dir> \
+      -model-input=<inputName1>,<inputType1>,<inputShape1> \
+      -model-input=<inputName2>,<inputType2>,<inputShape2> \
```
opti-mix: Please mention an example with quantized types to describe the syntax here

tools/loader/Loader.cpp
```diff
@@ -493,7 +493,7 @@ Loader::Loader() {
                             ExecutionBackend);
 
   hostManager_ = llvm::make_unique<runtime::HostManager>(std::move(configs));
-  backend_ = createBackend(ExecutionBackend);
+  backend_ = std::unique_ptr<Backend>(createBackend(ExecutionBackend));
```
opti-mix: I'd suggest submitting this bug-fix as a small separate PR as it is a totally orthogonal issue.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
```diff
@@ -66,8 +66,8 @@ llvm::cl::opt<llvm::Reloc::Model> relocModel(
 /// Limitation of number of arguments for `emitDataParallelKernel`.
 constexpr static size_t kArgLimit = 64;
 
-/// Generate the LLVM MAttr list of attributes.
-static llvm::SmallVector<std::string, 0> getMachineAttributes() {
+/// Generate the LLVM machine attribute list for the host.
+static llvm::SmallVector<std::string, 0> getHostMachineAttributes() {
```
opti-mix: I'd suggest submitting these LLVMIRGen.cpp changes (and the corresponding changes introducing floatABI options) as a small separate PR as it is a totally orthogonal issue.

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  // Verify/initialize command line parameters, and then loader initializes
+  // the ExecutionEngine and Function.
+  parseCommandLine(argc, argv);
+
+  // Initialize loader.
+  Loader loader;
+
+  // Get model input names and types.
+  std::vector<std::string> inputNames;
```
opti-mix: Should inputNames and inputTypes be moved into the `if (!loader.getCaffe2NetDescFilename().empty())` block as they are used only there?

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+        "   - float, float16 (floating point types)                \n"
+        "   - int32, int64 (integer types)                         \n"
+        "   - int8q, int16q, int32q (integer quantized types)      \n"
+        "   - bool (logic type)\n"),
+    llvm::cl::value_desc("name,type,shape"), llvm::cl::cat(modelCompilerCat));
+
+} // namespace
+
+/// Parse the strings and get the model input names and types.
+static void getModelInputs(std::vector<std::string> &inputNames,
```
opti-mix: I feel like this function for parsing input names and types could be useful as a re-usable utility function, so that new similar tools can be built on top of it. Please consider moving it somewhere where it could be re-used. Maybe somewhere in `Caffe2ModelLoader.cpp` (or `Loader.cpp`) as these flags seems to be relevant only to loaders for the time being?

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+        "The supported types are:                                  \n"
+        "   - float, float16 (floating point types)                \n"
+        "   - int32, int64 (integer types)                         \n"
+        "   - int8q, int16q, int32q (integer quantized types)      \n"
+        "   - bool (logic type)\n"),
+    llvm::cl::value_desc("name,type,shape"), llvm::cl::cat(modelCompilerCat));
+
+} // namespace
+
+/// Parse the strings and get the model input names and types.
```
opti-mix: Please add in the comment the description of the syntax being parsed by this function.

opti-mix:
@mciprian13 CLA is not signed yet, right?

tools/loader/ModelCompiler.cpp
```diff
@@ -0,0 +1,186 @@
...
+      auto splitRes = shape.split(',');
+      CHECK(!splitRes.first.getAsInteger(0, val))
+          << "Model input shape integer invalid";
+      dims.push_back(val);
+      shape = splitRes.second;
+    }
+    CHECK(!shape.getAsInteger(0, val)) << "Model input shape integer invalid";
+    dims.push_back(val);
+
+    // Push data.
```
opti-mix: Do we need to check if there are multiple inputs with the same name?

docs/AOT.md
mciprian13: Done.

tools/loader/Loader.cpp
mciprian13: I cannot do this since the model-compiler unit test fails (ASAN).
I reverted though the ABI option modification and created a separate PR.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
mciprian13: I reverted the changes and created a separate PR.

tools/loader/ModelCompiler.cpp
mciprian13: Done.

tools/loader/ModelCompiler.cpp
mciprian13: Done.

tools/loader/ModelCompiler.cpp
mciprian13: I`ll do it if you tell me explicitly where to put it :).
The function is currently written such that is packaged together with the CLI option.
My 2 cents: I wouldn`t put in Caffe2ModelLoader because it could be necessary later for ONNX models (don`t know). If I put in Loader.cpp then the image classifier will have both "-model-input-name" and "-model-input" (kind of confusing).

tools/loader/ModelCompiler.cpp
mciprian13: I`m not sure. It would be a weird (ill defined) graph in my opinion to have multiple placeholders with same name.

mciprian13:
CLA not signed (corporate version). Unfortunately I am not responsible with it: the legal dpt is "working" on it:))

tools/loader/ModelCompiler.cpp
mciprian13: Actually, for consistency I`m thinking about using this utility also in image-classifier. Then I`ll put it in Loader.cpp. The single thing I`m reluctant about is that I`ll replace the "-model-input-name" with "-model-input" and I`ll have to make changes in several places. What do you say?

tools/loader/ModelCompiler.cpp
opti-mix: Exactly! But a user could specify the same input multiple times on the command line. I'd say we should detect such cases here and bail with an error message.

tools/loader/ModelCompiler.cpp
opti-mix: > If I put in Loader.cpp then the image classifier will have both "-model-input-name" and "-model-input" (kind of confusing).

Yeah, but I don't think it is too bad. I'd say let's put it into Loader.cpp.

opti-mix: > I'll replace the "-model-input-name" with "-model-input" and I`ll have to make changes in several places. What do you say?

Overall, I kind of like this idea.

Can we also support syntax for `model-input`, where only name (without type and shape) is provided, so that e.g. we could declare `-model-input-name` as an alias for `-model-input` and use the old syntax... Then you don't need to update the existing uses of `model-input-name`. We may also check depending on the loader being used if the specified input has a type shape specified in the command-line or not.

tools/loader/ModelCompiler.cpp
mciprian13: I moved the utility in the Loader class and added a checker for unique name. Have a look.

tools/loader/ModelCompiler.cpp
mciprian13: I moved the utility function in the Loader class.
I also added the format with default type like this:
-model-input=<name>    with default type 'float' and default shape '[1]'

Now in principle this could be used to alias -model-input-name but I would keep the modifications to image-classifier as a separate PR because I envision there another modification:
- currently the input shape is derived from the input png images: I don`t like this because a user could provide images with the wrong size
- the input shape should be derived from the model itself for ONNX and from CLI for Caffe (just as for model-compiler) and then the image sizes should be checked against the model input size

mciprian13:
The CLA has been signed. Can we reopen this? Thanks!

opti-mix:
@mciprian13 Re-opened this PR per your request. Please resolve the merge conflicts and push a new version.

PR #3486, opened 2019-09-03
Add getters and setters for scale/offset from fused RWQ tensors

jfix71:
Summary: Make using fused tensors easier by adding a helper to Tensor to retrieve scale and offset from a specific row.

Test Plan: Added test. All other unit tests still pass.

### DIFF
--- a/lib/Base/Tensor.cpp
+++ b/lib/Base/Tensor.cpp
@@ -599,28 +599,15 @@ Tensor Tensor::getCopyConvertedToType(ElemKind newKind) const {
              {dims()[0], dims()[1] - 2 * (sizeof(float) - sizeof(float16_t))},
              1.0, 0);
 
-  const size_t srcWidth = dims()[1];
   const size_t dstWidth = tmp.dims()[1];
-  auto *srcData = reinterpret_cast<uint8_t *>(getUnsafePtr());
-  auto *dstData = reinterpret_cast<uint8_t *>(tmp.getUnsafePtr());
   auto srcH = getHandle<uint8_t>();
   auto dstH = tmp.getHandle<uint8_t>();
   for (size_t i = 0, e = dims()[0]; i < e; i++) {
-    // Get the src's scale/offset for the row.
-    uint8_t *srcScaleOffsetPtr =
-        &srcData[(i + 1) * srcWidth] - 2 * sizeof(float);
-    float srcScale, srcOffset;
-    memcpy(&srcScale, srcScaleOffsetPtr, sizeof(float));
-    memcpy(&srcOffset, srcScaleOffsetPtr + sizeof(float), sizeof(float));
-
-    // Convert scale/offset to FP16 and copy only those to dst.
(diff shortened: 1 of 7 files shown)


lib/Base/Tensor.cpp
```diff
@@ -599,28 +599,14 @@ Tensor Tensor::getCopyConvertedToType(ElemKind newKind) const {
...
-    float16_t dstOffset = static_cast<float16_t>(srcOffset);
-    uint8_t *dstScaleOffsetPtr =
-        &dstData[(i + 1) * dstWidth] - 2 * sizeof(float16_t);
-    memcpy(dstScaleOffsetPtr, &dstScale, sizeof(float16_t));
-    memcpy(dstScaleOffsetPtr + sizeof(float16_t), &dstOffset,
-           sizeof(float16_t));
+    // Copy the scale/offset from src to dst.
+    float scale, offset;
+    std::tie(scale, offset) = srcH.getFusedScaleOffsetFromRow<float>(i);
+    dstH.setFusedScaleOffsetInRow<float16_t>(i, scale, offset);
```
beicy: nit: is it better to use static_cast here?  `scale->static_cast<float16_t>scale`?

tests/unittests/TensorsTest.cpp
```diff
@@ -994,10 +994,41 @@ TEST(Tensor, randomizeFused_Float) {
...
+  }
+  for (size_t i = 0; i < 10; i++) {
+    ScaleOffsetT scale, offset;
+    std::tie(scale, offset) = TH.getFusedScaleOffsetFromRow<ScaleOffsetT>(i);
+    EXPECT_EQ(scale, (ScaleOffsetT)i);
+    EXPECT_EQ(offset, (ScaleOffsetT)i);
+  }
+}
+
+/// Test getting and setting fused offsets from UInt8FusedQTy.
```
beicy: "fused offsets" ->"fused offsets and scales".  Same for other 2 tests' comments.

lib/Base/Tensor.cpp
jfix71: Hm I think the implicit conversion will do the cast as expected, but I can add it in to be more explicit.

tests/unittests/TensorsTest.cpp
```diff
@@ -994,10 +994,41 @@ TEST(Tensor, randomizeFused_Float) {
...
+    EXPECT_EQ(offset, (ScaleOffsetT)i);
+  }
+}
+
+/// Test getting and setting fused scaled and offsets from UInt8FusedQTy.
+TEST(Tensor, GetFusedScaleOffset_UInt8FusedQTy) {
+  testGetSetFusedScaleOffset<float>(ElemKind::UInt8FusedQTy);
+}
+
+/// Test getting and setting fused scaled and offsets from UInt8FusedFP16QTy.
```
beicy: "scaled" -> "scales" ?

PR #3490, opened 2019-09-04
[LLVM] Add command line option for ABI type

mciprian13:
**Summary**
Add command line option for ABI type (soft/hard/default): useful to force the generated bundle to have a specific ABI such that it links properly with some MCU toolchains.
Add "-mattr" alias for "-target-feature" LLVM option (common usage from clang).


opti-mix:
@mciprian13 Looks good! Thanks for submitting it as a separate PR!

But we still need the CLA to be signed.

mciprian13:
The CLA has been finally signed. Can we have this landed?
Thanks!

PR #3491, opened 2019-09-05
Interpreter send-recv experiment == Not to land ==

mortzur:
===== NOT Intended To Land =====
Summary:
This experiment aims for finding a good representation for P2P primitives in Glow graph, suitable for P2P implementations in accelerator backends.
Here, I cheated my way in the interpreter backend to identify some of the issues need solving.
"Cheated", because of passing a pointer to the helper class instance which serves as a mapping between the sender and receiver (which made what was supposed to be the only API method "getRemotePeerToPeerAddress" redundant).

Possible lessons/conclusions:
* For a backend with MPI-like interface (i.e. send takes "rank" and "message" as args) we may use Glow nodes "Send" and "Recv" based on a channel-ID input (can be mapped to "rank") and the tensor-to-send ("message"). We may want to use different node types for these and direct-writes.
(This implementation misses the point because it mixes MPI-like helper with "direct-write" like nodes descriptions)

* In order to perform direct writes, "Send" should get a physical address in a remote device as an input (acquired using "RecvReady" executed on the remote device).

* The remote address may be represented as just another input placeholder in bindings. In this case, the Runtime is responsible to schedule and fully-run "RecvReady" before calling the function with the corresponding "Send". 

I will probably abandon this interpreter implementation and continue to develop this on the CPU backend.
I'd love to get your feedback and thoughts.

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -155,6 +155,20 @@ public:
   /// \returns the DeviceInfo for this device containing peak limits for
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
+
+  virtual bool isPeerToPeerSupported() { return false; }
+
+  virtual llvm::Error setupRemotePeerToPeer(int64_t channelId,
+                                            DeviceManager *remote) {
+    return MAKE_ERR(GlowErr::ErrorCode::RUNTIME_ERROR,
+                    "Backend does not support peer-to-peer communication.");
+  }
+
+  virtual llvm::Error getRemotePeerToPeerAddress(int64_t channelId,
+                                                 Placeholder *remoteAddress) {
+    return MAKE_ERR(GlowErr::ErrorCode::RUNTIME_ERROR,
+                    "Backend does not support peer-to-peer communication.");
+  }
 };
 
 } // namespace runtime
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -164,8 +164,8 @@ public:
                     "Backend does not support peer-to-peer communication.");
   }
 
-  virtual llvm::Error getRemotePeerToPeerAddress(int64_t channelId,
-                                                 Placeholder *remoteAddress) {
+  virtual llvm::Expected<int64_t>
+  getRemotePeerToPeerAddress(int64_t channelId, PlaceholderBindings *bindings) {
     return MAKE_ERR(GlowErr::ErrorCode::RUNTIME_ERROR,
                     "Backend does not support peer-to-peer communication.");
   }
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -134,6 +134,15 @@ class Placeholder : public Storage {
   /// Specifies if associated Tensors should be zeroed when allocated.
   bool allocZero_{false};
 
+  // Specifies the policy about loading the associated Tensor to the device,
+  // and reading it from the device:
+  // 0 - Input - always load, read on demand (i.e. for debug)
+  // 1 - Intermediate - do not load, read on demand
+  // 2 - IntermediateWithInit - load at init only, read on demand
+  // 3 - Output - do not load, always read
+  // 4 - InOut - always load, always read
+  int ioPolicy_{0};
+
 public:
   /// Create a new placeholder.
   Placeholder(llvm::StringRef name, TypeRef Ty, bool isTrainable)
@@ -148,6 +157,9 @@ public:
 
   /// \returns True if associated Tensors should be zeroed when allocated.
   bool allocZero() const { return allocZero_; }
+
+  int getIoPolicy() { return ioPolicy_; }
+  void setIoPolicy(int p) { ioPolicy_ = p; }
 
   /// Sets whether or not associated Tensors should be zeroed.
   void setAllocZero(bool on = true) { allocZero_ = on; }
--- a/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
+++ b/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
@@ -20,8 +20,25 @@
 
 #include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
+#include "glow/ExecutionContext/ExecutionContext.h"
 
 namespace glow {
+
+/// CPUDeviceBindings inherits from DeviceBindings, it contains per run
+/// device specific information used to run a compiled function on a specific
+/// device.
+struct CPUDeviceBindings : DeviceBindings {
+  CPUDeviceBindings(uint8_t *activationsBuffer, uint8_t *weightsBuffer)
+      : DeviceBindings("CPU"), deviceActivationsBuffer{activationsBuffer},
+        deviceWeightsBuffer{weightsBuffer} {}
+
+  ~CPUDeviceBindings() {}
+
+  /// TODO: determine if these are owned pointers! ==> should be freed in DTOR
+  uint8_t *deviceActivationsBuffer;
+  uint8_t *deviceWeightsBuffer;
+};
+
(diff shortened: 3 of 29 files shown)


include/glow/Graph/Nodes.h
```diff
@@ -149,6 +158,9 @@ class Placeholder : public Storage {
   /// \returns True if associated Tensors should be zeroed when allocated.
   bool allocZero() const { return allocZero_; }
 
+  int getIoPolicy() { return ioPolicy_; }
```
putivsky: const method.

include/glow/Backends/DeviceManager.h
```diff
@@ -155,6 +155,20 @@ class DeviceManager {
   /// \returns the DeviceInfo for this device containing peak limits for
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
+
+  virtual bool isPeerToPeerSupported() { return false; }
```
putivsky: const method

include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
```diff
@@ -20,8 +20,25 @@
...
+/// device.
+struct CPUDeviceBindings : DeviceBindings {
+  CPUDeviceBindings(uint8_t *activationsBuffer, uint8_t *weightsBuffer)
+      : DeviceBindings("CPU"), deviceActivationsBuffer{activationsBuffer},
+        deviceWeightsBuffer{weightsBuffer} {}
+
+  ~CPUDeviceBindings() {}
+
+  /// TODO: determine if these are owned pointers! ==> should be freed in DTOR
+  uint8_t *deviceActivationsBuffer;
```
putivsky: std::unique_ptr has a custom deleter, caller can provide something like std::unique_ptr<std::vector< uint8_t >> with custom deleter (control the ownership of unique_ptr memory).

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -163,5 +192,31 @@ void CPUDeviceManager::runFunctionImpl(
...
+  if (funcIt != functions_.end()) {
+    auto *baseAddress = buffers_[function]->getWeightsBuffer();
+    auto &bundle = funcIt->second->getRuntimeBundle();
+    auto &symbolTable = bundle.getSymbolTable();
+    auto *PH = bindings->getPlaceholderByName("recv_input");
+    auto it = symbolTable.find(PH->getName());
+    if (it == symbolTable.end()) {
+      llvm::outs() << "Did not find symbol\n";
+    }
+    auto symbolInfo = it->second;
```
putivsky: const auto &

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -24,6 +24,35 @@
...
 namespace runtime {
 
+/// A class that contains an openCL device buffer. It frees the buffer when it
+/// is destroyed.
+class CPUBuffer {
+  /// The OpenCL buffer being stored.
+  uint8_t* activationsBuffer_;
+  uint8_t* weightsBuffer_;
+  /// Size of the buffer in bytes.
+  const size_t activationsSize_{0};
```
putivsky: why is it defined as "const"?

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -24,6 +24,35 @@
...
+        activationsSize_(activationsSize), weightsSize_(weightsSize) {}
+
+  CPUBuffer* getBuffer() { return this; }
+
+  /// Returns the stored buffer.
+  uint8_t *getActivationsBuffer() { return activationsBuffer_; }
+  uint8_t *getWeightsBuffer() { return weightsBuffer_; }
+
+  /// Get size of buffer in bytes.
+  size_t getActivationsSize() { return activationsSize_; }
```
putivsky: const methods

lib/Backends/Interpreter/InterpreterDeviceManager.h
```diff
@@ -33,6 +36,8 @@ class InterpreterDeviceManager : public QueueBackedDeviceManager {
   /// This is very arbitrary for the Interpreter backend.
   const uint64_t functionCost_{1};
 
+  std::shared_ptr<InterpreterDeviceMemoryHelper> memoryHelperPtr_;
+
   /// String constant for logging number of in-use devices.
   static constexpr const char *kDevicesUsedInterpreter =
```
putivsky: no need for static - better to use anonymous namespace.

lib/Backends/Interpreter/InterpreterDeviceMemoryHelper.cpp
```diff
@@ -0,0 +1,85 @@
...
+InterpreterDeviceMemoryHelper::~InterpreterDeviceMemoryHelper() {
+  for (const auto &pair : tensors_) {
+    delete pair.second;
+  }
+  tensors_.clear();
+}
+
+void InterpreterDeviceMemoryHelper::allocateTensor(channel_id_type channelId,
+                                                   TypeRef tensorType) {
+  std::lock_guard<std::mutex> tGuard(tensorsGuard_);
```
putivsky: If anywhere else the order of lock will be reversed it will cause the dead-lock.
Better to lock only for the particular resource and release lock right away.

`std::unique_lock<std::mutex> tGuard(tensorsGuard_);

  assert(tensors_.find(channelId) == tensors_.end() &&
         "Tensor already allocated for channel ID");

  tensors_.insert(std::make_pair(channelId, new Tensor(tensorType)));
 
 tGuard.unlock();

 std::unique_lock<std::mutex> sGuard(transferStatusGuard_);

 transferStatus_.insert(std::make_pair(channelId, false));
 `

lib/Backends/Interpreter/InterpreterDeviceMemoryHelper.cpp
```diff
@@ -0,0 +1,85 @@
...
+Tensor *
+InterpreterDeviceMemoryHelper::getRemoteTensor(channel_id_type channelId) {
+  std::lock_guard<std::mutex> rGuard(remotesGuard_);
+  auto it = remoteReceiverDevices_.find(channelId);
+  assert(it != remoteReceiverDevices_.end() && "Unknown remote channel ID");
+  return it->second->getLocalTensor(channelId);
+}
+
+bool InterpreterDeviceMemoryHelper::getLocalStatus(channel_id_type channelId) {
+  const auto st = transferStatus_.find(channelId);
```
putivsky: it's already the copy of iterator, why to make it "const"?

lib/Backends/Interpreter/InterpreterDeviceMemoryHelper.cpp
```diff
@@ -0,0 +1,85 @@
...
+  assert(st != transferStatus_.end() && "Unknown channel ID");
+  return st->second;
+}
+
+void InterpreterDeviceMemoryHelper::addRemoteReceiver(
+    channel_id_type channelId, InterpreterDeviceMemoryHelper *receiver) {
+  std::lock_guard<std::mutex> rGuard(remotesGuard_);
+  remoteReceiverDevices_.insert(std::make_pair(channelId, receiver));
```
putivsky: what happen if the same channel Id will be inserted multiple times?

lib/Exporter/ONNXModelWriter.cpp
```diff
@@ -1140,6 +1140,9 @@ DEF_ALL_WRITER_NODE(RowwiseQuantizedFullyConnected)
 DEF_ALL_WRITER_NODE(RowwiseQuantizedSparseLengthsWeightedSum)
 DEF_ALL_WRITER_NODE(FusedRowwiseQuantizedSparseLengthsSum)
 DEF_ALL_WRITER_NODE(FusedRowwiseQuantizedSparseLengthsWeightedSum)
+DEF_ALL_WRITER_NODE(RecvReady)
```
putivsky: The correspondent ONNXLoader methods required.

PR #3492, opened 2019-09-05
[PyTorch] Add support for aten::quantize_linear and aten::dequantize

zrphercule:
This pr create basic support for aten::quantize_linear and aten::dequantize.
Todo: add unit test, the correctness of this pr is not tested yet.
Feel free to give some high level comments.

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -25,6 +25,15 @@
 namespace glow {
 
 namespace {
+/// For the quantized PyTorch ops, the activations are quantized to uint_8.
+/// In Glow, the activations are quantized to int_8. Therefore, for the offset
+/// read from quantized pytorch model, we need to subtract 128(i.e. INT8_MIN) to
+/// make the activations becomes int8_t.
+/// For Glow: -128 <= orig_fp32/scale_1 + offset_1 <= 127
+/// For PyTorch: 0 <= orig_fp32/scale_2 + offset_2 <= 255
+/// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128
+const int32_t OFFSETSHIFT = 128;
+
 /// Downcast a double to a float.
 Expected<float> to32Bit(double val) {
   RETURN_ERR_IF_NOT(val <= std::numeric_limits<float>::max() ||
@@ -347,6 +356,16 @@ struct AdaptiveAvgPoolInputs {
   };
 };
 
+/// Indexes of aten::quantize_linear inputs.
+struct QuantizeInputs {
+  enum {
+    input = 0,
+    scale = 1,
+    offset = 2,
+    dtype = 3,
+  };
+};
+
 /// Indexes of aten::linear inputs.
 struct LinearInputs {
   enum {
@@ -438,6 +457,10 @@ PyTorchModelLoader::getSymbolsMapping() {
             ClampInputs::min,
             ClampInputs::max,
         }},
+       {{"aten::quantize_linear"},
+        &PyTorchModelLoader::loadQuantize,
+        {QuantizeInputs::scale, QuantizeInputs::offset, QuantizeInputs::dtype}},
+       {{"aten::dequantize"}, &PyTorchModelLoader::loadDequantize, {}},
        {{"aten::size"}, &PyTorchModelLoader::loadSize, {SizeInputs::dim}},
        // TODO: use -1 to freeze all inputs
        {{"prim::ListConstruct"}, &PyTorchModelLoader::loadListConstruct, {}},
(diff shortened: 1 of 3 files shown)


nadavrot:
Cool @zrphercule .

torch_glow/tests/nodes/quantize_dequantize_test.py
```diff
@@ -0,0 +1,22 @@
...
+from tests.utils import jitVsGlow
+
+
+def test_quantization_basic():
+    """Basic test of the PyTorch quantize and dequantize Node on Glow."""
+
+    def test_f(a, b):
+        q = torch.nn.quantized.Quantize(255, 0, torch.quint8)
+        dq = torch.nn.quantized.DeQuantize()
+        return dq(q(a.add(b)))
```
jackm321: Why does this test add then do quant-dequant on empty network? Should it be dq(q(a)) or put the add in the middle of the quantized section?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1092,6 +1115,60 @@ llvm::Error PyTorchModelLoader::loadBatchNorm(const torch::jit::Node *ptNode) {
...
+                             getGlowNodeValue(inputs[QuantizeInputs::input]));
+
+  // scale
+  auto scaleHandle = glow::Handle<float>::createInvalidHandle();
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      scaleHandle, getGlowConstantHandle<float>(inputs[QuantizeInputs::scale]));
+  float outScale = scaleHandle.raw(0);
+
+  // offset
+  auto offsetHandle = glow::Handle<int8_t>::createInvalidHandle();
```
rdzhabarov: just checking on this. is offset int8? (typically it's int32)

zrphercule: yeah it is int32...

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1167,6 +1190,56 @@ llvm::Error PyTorchModelLoader::loadBatchNorm(const torch::jit::Node *ptNode) {
...
+  ASSIGN_VALUE_OR_RETURN_ERR(
+	  outOffset, iValToInt(getGlowIValueForValue(inputs[QuantizeInputs::offset])));
+
+  // dtype, we only support quantize to int8 for now
+  int32_t outDtype;
+  ASSIGN_VALUE_OR_RETURN_ERR(outDtype, iValToInt(getGlowIValueForValue(inputs[QuantizeInputs::dtype])));
+
+  // Right now pytorch only has quint8 quantization support, therefore we only
+  // support uint8 as well.
+  RETURN_ERR_IF_NOT(outDtype == 13,
```
jackm321: Can we use ScalarType::QUInt8 instead of 13 here?

zrphercule:
> I'm curious, unlike c2, we should be able to easily pass quantized tensors in and out of glow fusion group right? Like quantize and/or dequantize can happen in pytorch jit and just quantized ops will happen in glow. Is that correct?

I think you are right, the output of quantize & input of dequantize are both quantized tensors in pytorch, which means so long as we have socket for quantized tensor, it should work. Also quantize/dequantize in pytorch may help us evade some problem (but still we need to do requantize everytime we use some ops in glow right?).
While I also think we should keep these two ops working and leave the choice to users. Since sometimes quantize is not the start of the model and dequantize is not the end of the model, we dont want the graph to be separated by these two ops.

yinghai:
> I'm curious, unlike c2, we should be able to easily pass quantized tensors in and out of glow fusion group right? Like quantize and/or dequantize can happen in pytorch jit and just quantized ops will happen in glow. Is that correct?

The tricky part is how we are going to create a quantized pt tensor from glow quantized tensor and carry the quantization info. 
https://github.com/pytorch/glow/blob/aa4f6fe5ff380a0002567a8f05f8d0f76fc500ff/torch_glow/src/PyTorchCommon.cpp#L174-L175

torch_glow/tests/nodes/quantize_dequantize_test.py
zrphercule: That would need a translation of quantized add operator as well, which introduced more non-stable stuff. I prefer to let the test of quantize-dequantize be like this, at least for now.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1163,6 +1186,56 @@ Error PyTorchModelLoader::loadBatchNorm(const torch::jit::Node *ptNode) {
...
+                                            inputs[QuantizeInputs::offset])));
+
+  // dtype, we only support quantize to int8 for now
+  int32_t outDtype;
+  ASSIGN_VALUE_OR_RETURN_ERR(outDtype, iValToInt(getGlowIValueForValue(
+                                           inputs[QuantizeInputs::dtype])));
+
+  // Right now pytorch only has quint8 quantization support, therefore we only
+  // support uint8 as well.
+  RETURN_ERR_IF_NOT(outDtype == 13, "we only support to be quantized as uint8");
```
zrphercule: oh this equal to 13 stuff I need to change.

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -302,6 +302,14 @@ class PyTorchModelLoader {
   /// \returns error on failure.
   Error loadBatchNorm(const torch::jit::Node *ptNode);
 
+  /// Load a PyTorch quantize_linear node.
+  /// \returns error on failure.
+  Error loadQuantize(const torch::jit::Node *ptNode);
+
+  /// Load a PyTorch dequantize_linear node.
```
jackm321: dequantize not dequantize_linear

torch_glow/tests/nodes/quantize_dequantize_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_quantization_basic():
+    """Basic test of the PyTorch quantize and dequantize Node on Glow."""
+
+    # This only test if operator has been successfully projected
+    # Calculating process in glow will be optimized
+    def test_simple(a, b):
+        q = torch.nn.quantized.Quantize(1, 0, torch.quint8)
+        dq = torch.nn.quantized.DeQuantize()
+        return dq(q(a.add(b)))
```
jackm321: add a todo to move the add into the lowered graph?

zrphercule: I have added some comments before this test_simple already, but well add more wont bite ;)

PR #3495, opened 2019-09-10
[Graph] Make Module::addConstant private and delete Module::addPlaceholder

### LINKED ISSUE #3428: Verify module Storage node ownership/usage
Are Placeholders/Constants supposed to be owned/referenced by only one Module?  If so, it would be nice if this was verified by the verification functions.

We have run into a few test cases where Placeholders from one module were being used in a Function from a different module, which breaks one of our backend-specific optimizations that adds new Placeholders.

SplitInfinity:
**Summary**
`Storage` nodes (i.e. `Constants` and `Placeholders`) are supposed to be owned
by a `Module`. Accordingly, the preferred methods of creation for these
objects are `Module::createConstant` and `Module::createPlaceholder`.
Exposing public `addConstant` and `addPlaceholder` methods makes it possible
to create a `Storage` node and add it to multiple `Modules`. These methods
are not used frequently, so this commit makes one of them them private
and deletes the other in order to preclude this issue.

**Testing**
This commit adds a new unit test to `GraphTest` to make sure that
`Module::verify()` can catch the case in which a `Storage` node is owned by
one `Module` but used by a `Function` in another `Module`.

**Fixes**
This PR fixes #3428.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -75,6 +75,9 @@ class Module final {
   /// Module log context that stores all logs related to this module.
   ModuleLogContext moduleLogCtx_;
 
+  /// Inserts the constant \p V to the list of constants.
+  Constant *addConstant(Constant *V);
+
 public:
   Module() = default;
 
@@ -84,12 +87,6 @@ public:
   /// names are legal C identifiers in the form: "[a-zA-Z_][a-zA-Z0-9_]*".
   static llvm::StringRef uniqueName(llvm::StringRef name,
                                     llvm::StringSet<> &stringTable);
-
-  /// Inserts the constant \p V to the list of constants.
-  Constant *addConstant(Constant *V);
-
-  /// Inserts the placeholder node \p ph to the list of variables.
-  Placeholder *addPlaceholder(Placeholder *ph);
 
   /// Return a pointer to a uniqued type \p T.
   TypeRef uniqueType(const Type &T);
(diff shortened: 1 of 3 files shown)


jfix71:
Thanks for working on this @SplitInfinity.

I think this mostly makes sense, but I'm wondering if we should go a slightly different direction. Right now Node has a `Function *parent_` member, which points to the Node's containing Function. However, Storage is also a subclass of Node, meaning Storage has a Function parent as well, even though this concept doesn't really make sense anymore (did it ever?).

I'm wondering if perhaps we should consider adding some other subclass ComputeNode or something which has this `Function *parent_`. Then we could add a `Module *parent` to Storage. WDYT?

SplitInfinity:
Yeah, I think that makes sense. I think we should propose this to a broader audience because it affects a lot of code even though it is a simple change.

SplitInfinity:
Okay, this is ready for review. The new plan is to make `addPlaceholder/addConstant` private (this PR) to make it harder for one `Module` to own a `Storage` and for another to use it, and then to add a new subclass of `Node` called `Operator` that contains a parent `Function` pointer and to add a parent `Module` pointer to `Storage` (next PR).

Assigning @rdzhabarov since @jfix71 is not available this week.

jackm321:
Why not just get rid of them?

SplitInfinity:
@jackm321 

They are used inside some other methods in `Module`, e.g.:
```
Constant *Module::createConstant(TypeRef T, llvm::StringRef name) {
  auto FT = uniqueType(*T);
  return addConstant(new Constant(name, FT));
}
```

jackm321:
> They are used inside some other methods in Module, e.g.:

OK but addPlaceholder is only used once by createPlaceholder so might as well inline it and addConstant is used twice in the two createConstants so could go either way

rdzhabarov:
I'll be PTO tomorrow.

SplitInfinity:
Inlined `addPlaceholder`. I kept `addConstant` because it is used more than once.

PR #3501, opened 2019-09-11
Reintroduce HabanaConvolution because of weird filter layout

bertmaher:
Summary:
Even though we've introduced a clean mechanism for controlling the
input layout of convolutions (NHWC vs NCHW), we didn't account for a
potentially different *filter* layout.  It looks like Habana follows
[TensorFlow's
convention](https://www.tensorflow.org/api_docs/python/tf/nn/convolution) for
filter shape, which is (H, W, input_channels, output_channels), rather than
Glow's (OC, H, W, IC) or Caffe2's (OC, IC, H, W).

Without a way to express the differing filter layout, there's no way to check if the existing convolution needs to be transformed.  In the existing implementation, this manifests as the transformPostLowering step creating Transposes in an endless loop.

Since we don't have a great mechanism for configuring the filter layout, I'm
simply reverting to the Habana-specific convolution implementation, and feeling
a bit sad.

Differential Revision: D17311487

### DIFF
--- a/tests/unittests/HabanaGlowTest.cpp
+++ b/tests/unittests/HabanaGlowTest.cpp
@@ -149,13 +149,13 @@ TEST_F(HabanaBackendTest, FuseConvRelu) {
   //   Placeholder ------
   //       |            |
   //       v            v
-  //     Conv         Conv
+  //   HabanaConv       Conv
   //       |            |
   //       v            v
   //     Save         Relu
 
   // Check that HabanaConv feeds into the Save.
-  auto *HCA = llvm::dyn_cast<ConvolutionNode>(SN->getInput());
+  auto *HCA = llvm::dyn_cast<HabanaConvolutionNode>(SN->getInput());
   ASSERT_TRUE(HCA);
 
   // Check that the inputs to the HabanaConvAdd are the same as the Conv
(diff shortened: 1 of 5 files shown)


tests/unittests/HabanaGlowTest.cpp
```diff
@@ -403,6 +403,21 @@ TEST_F(HabanaBackendTest, ConvertFC) {
...
+  Placeholder *input =
+      mod_.createPlaceholder(ElemKind::FloatTy, {1, 10, 20, 3}, "input", false);
+  ConvolutionNode *conv = F_->createConv(ctx_, "conv", input, 3, 5, 1, 2, 1);
+  SaveNode *save = F_->createSave("save", conv);
+
+  CompilationContext cctx;
+  bool changed = backend.transformPostLowering(F_, cctx);
+  EXPECT_TRUE(changed);
+  ASSERT_TRUE(save);
+  ASSERT_TRUE(llvm::isa<HabanaConvolutionNode>(save->getInput()));
```
SplitInfinity: Consider checking that the `doRelu` property of the node is `false`.

bertmaher:
The motivation is that the existing code is wrong :-).  In particular, the transformation that transposes the filter to HWIO goes into an infinite loop (since the convolution never changes in any way that would make the check fail, it inserts transposes until your computer runs out of memory).

A solution would be to check the filter layout before performing the transformation, but we don't have any notion of the filter layout anywhere in Glow, so there's no way to check this.  We arguably should have such a notion, but in the meantime this diff is a straight-up revert of the HabanaConvolution removal that gets tests to pass.

SplitInfinity:
It would be helpful to mention that in the PR or commit message to preclude such silly questions in the future :). Another (less serious) problem is that if someone creates a `ConvolutionNode` like so
```
auto *CN = new ConvolutionNode( ..., FusedActivation::Relu);
F->addNode(CN);
```
instead of using `Function::createConv`, `Habana::transformPostLowering` won't insert the filter `Transpose` nodes.

Issue #3500, opened 2019-09-11
[GraphOptimizer] Fix bug in constant folding optimization.


shajrawi:
We're creating an `Interpreter` backend, but we are also running constant folding after lowering. This can be bad on backends that create transformations that break Glow's canonical tensor layout.
See `NonSquarePaddingConvolution` on the `OpenCL` backend for example. We have a `Reshape` node in `NCHW` layout. Constant folding creates the following function for it:
```
Graph structure __constEvaluationFunction__:
Reshape
name : conv_filter1
Input : float<2 x 2 x 2 x 1>
Dims : [2, 1, 2, 2]
Layout : NCHW
users : 1
Result : float<2 x 1 x 2 x 2>
Save
name : conv_filter11
Input : float<2 x 1 x 2 x 2>
Output : float<2 x 1 x 2 x 2>
users : 0
```
If we run a canonical layout verifier on said function, see https://github.com/pytorch/glow/issues/3452, it is gonna blow up: `Save` is getting `NCHW` layout but expects `NHWC` layout. The constant function that we created is bad.

shajrawi:
CC @opti-mix

Issue #3493, opened 2019-09-05
[GraphOptimizer] Fix bug in post-lowering optimizations for OpenCL


shajrawi:
As part of #3452 I created a layout verifier for the `OpenCL` backend. It caught a bug in our optimizations: see `smallConv` backend correctness test with the `OpenCL` backend for a tiny repro. We have a `NHWC` conv. which we turn into `NCHW` with a transpose of the inputs (which is fine). then the bug occurs when we optimize post lowering: `OptimizeTransposeIntoReshape` turns the transpose (of a placeholder) into a reshape and `DCE` removes the reshape. this is obviously incorrect because the placeholder is in `NHWC` format.

nadavrot:
Great catch @shajrawi !

shajrawi:
Found out why DCE deletes the Reshape node, at some point the conv. is no longer its user: it is done in the `OptimizeReshape` optimization, we do `Reshape(Constant) -> Constant'`. after the placeholder has been turned into a constant in placeholder->constant optimization.
This is fine, the verifier complained because it needs to know the layout of any constants / placeholders. The original constant had `NHWC` layout, and if we try to find out the layout of Constant->Conv. using Glow's canonical layout requirements then the new layout is `NHWC` too. `OptimizeRehsape` is not aware that it is running on `OpenCL` backend and as such it created a constant with `NHWC` instead.

Issue #3452, opened 2019-08-22
Teach Glow how to support layout requirements for nodes and tensors


shajrawi:
Taking the tensor layouts portion out of #2686 as alignment requirements have already been done and this is a different issue.

Quoting @opti-mix :
> Different operations (e.g. convolutions, gemms, etc) may require each of their inputs and outputs to use a specific memory layout, both in terms of alignments and logical ordering of dimensions (e.g. NHWC or NCHW). More over, these requirements may depend on the properties of the operation, e.g. dimensions of its operands, etc. E.g. a convolution with a small filter may need the input operands in a format different from a convolution with a big filter.
>
> Today, Glow core and custom backends implicitly hard-code this knowledge about the operations into (backend-specific) nodes and code that works with them, i.e. in the code that transforms Glow nodes, which are in the canonical `NHWC` format, into backend-specific nodes. This is pretty fragile and involves a lot of boiler plate code.
>
> What if Glow would be able to ask a backend for each node what are the requirements for each of node's inputs and results in terms of alignments and ordering of dimensions? 
>
> - The backend would return this information and Glow core could insert all the required layout transformations using a new node called `LayoutTransform` (may be we could extend/subsume Transpose for these purposes).
>
> - This new `LayoutTransform` node could be applied to Constant weights at compile-time. It can also be optimized in almost the same ways as Glow optimizes Transpose nodes, e.g. multiple `LayoutTransforms` can be combined into one, opposite `LayoutTransforms` can eliminate each other, etc.
>
> - This way all of the functionality to insert the required layout transforms is handled by the Glow core, which removes a lot of code duplication from backends. It also allows for better optimizations as the optimizer would understand the semantics of `LayoutTransform`. And it also allows for better verification, as Glow can now check that layouts of all inputs/outputs of each node satisfy the requirements of this operation as reported by backends.

To expand on this idea I'd like to share an idea by @artemrakhov 
> first start with a "raw" state of uncompliance, where we insert LayoutTransformation nodes while needed. Then we have a loop to sink and clamp layout transformations together, similarly to what we do to transposes. In fact, LayoutTransformation is generalization of Transpose, since we can not only change the order of dimensions, but also their alignments.

Basically, we need to port/expand transpose syncing / elimination optimization to support layouts.

Additionally, I think that if we pursue this direction, It should also include a verifier that makes sure that a node `A` that expects certain layout constraints for its input have said constraints satisfied after doing layout transformations.

[...]

Issue #3499, opened 2019-09-10
[GraphOptimizer] Fix wrong layout assumption in OptimizeReduceMean?


shajrawi:
I see the following code in the optimization:
```
      // In Glow, AvgPool expects NHWC.
      auto *TR1 = F->createTranspose(
          RM->getName().str() + ".transposeNCHW2NHWC", in, NCHW2NHWC);
      auto *AP = F->createAvgPool(RM->getName().str() + ".avgPool", TR1,
                                  kernels, strides, pads);
      auto *TR2 = F->createTranspose(
          RM->getName().str() + ".transposeNHWC2NCHW", AP, NHWC2NCHW);
```
This looks like a bug to me, the canonical representation in Glow is NHWC. Why do we expect the input to be in NCHW format and doing the two transposes?

shajrawi:
cc @vuzelac-cadence

dspmihai:
Will this code be ever executed? I see in lowerBatchedReduceMeanNode that ReduceMean can only have one axis. @jfix71, @mciprian13 , any idea how it's better to approach this again? I am thinking of lowering ReduceMean to AvgPool in Lower.cpp, if possible. Or maybe lower to multiple ReduceMean, then find the pattern in OptimizeReduceMean (possible error-prone after future optimizations)?

vuzelac-cadence:
@dspmihai, lowering is backend controlled. For example, we do use this optimization.

PR #3503, opened 2019-09-11
Teach Glow how to support layout requirements and fix uncovered bugs.

shajrawi:
Summary:

Note: I did not want to break the `enum ConvolutionLayout` introduced in 5074a7233b9eabbf2a7da1e82e776ff335a68fbf, As such, I used it in the verifier / did not change the creation of said nodes.
HOWEVER: We should use the more-generic string-based layout, which I introduce to Transpose node in this commit: it is basically an extendable enum that can be used in the backends without touching the generic code base. as a bonus, it makes differentiation easier: see how it is done for transpose now in `Function *glow::differentiate`.

Getting rid of said enum is a proposed TODO / follow-up.

Also note that some nodes *need* layout requirements, which have been added, namely we need to know the layout for placeholders and constants (obviously) and for reshapes (in case we optimized a transpose into a reshape).

An additional nice-to-have feature of the string-based layout is the wildcard / any-layout option. Some operations, such as data parallel nodes, might accept any layout.

A potential follow-up is to get create a "Solver" that automatically inserts transposes if the layouts do not match, this might greatly simplify the loader: we no longer need to insert transposes based on if we are importing NHWC or NCHW (for example). We just need to annotate the placeholder with the layout information we've get at load-time, and which we "forget" afterwards.

The verifier is useful even without creating said solver, it exposed a couple of bugs which are mentioned in this commit, as such any proposed solvers are not a must-have to demonstrate the usefulness of this commit.

Fixes https://github.com/pytorch/glow/issues/3452

Also Fixes https://github.com/pytorch/glow/issues/3493 and Fixes https://github.com/pytorch/glow/issues/3500 GraphOptimizer bugs which were found after adding the layout verifier.

Provides a workaround for the https://github.com/pytorch/glow/issues/3499 issue which was also found via the verifier.

Test Plan:
`ninja test`

### DIFF
--- a/docs/TensorLayout.md
+++ b/docs/TensorLayout.md
@@ -1,0 +1,286 @@
+## Tensor Layout
+
+This document describes the design of the tensor Layout requirements in Glow.
+
+Certain operations (e.g. convolutions, gemms, etc) need to know the semantic
+layout of their tensors, i.e. the logical ordering ordering of their dimensions
+(e.g. `NHWC`). Some backends enforce additional backend-specific requirements
+on said operations (e.g. tensor alignment).
+
+A theoretical clever backend, might even go a step further and have said
+layout requirements depend on the properties of the operation: a convolution
+with a small filter may need the input operands in a format different from a
+convolution with a big filter.
+
+Tensor layout is a property of the operation, some operations, such as
+element-wise operations, may not care about their input layout, we avoid adding
+a layout field for said operations to reduce the dynamic memory consumption of
+the compiler.
+
+For operations that do have layout requirements, Glow has an easily extendable
+string-based layout field. This allows backends to override Glow's default
+requirements without the hassle of creating a custom, backend-specific, node.
+
+The [class `TensorLayoutDescription`](https://github.com/pytorch/glow/blob/master/include/glow/Base/TensorLayoutUtils.h)
+represents the expected layout of tensors in Glow.
+It contains the following methods:
+
+- `bool isSameLayout(const TensorLayoutDescription &rhs) const`
+	- Returns true if both tensor layouts are the same.
+- `bool isSatisfiedBy(TypeRef ty, const TensorLayoutDescription *srcLayout) const`
+	- Returns true if the type `ty` satisfies the current layout.
+	- If `srcLayout` is provided, the verification is more complete,
+as it is taken into consideration, for example: the number of dimensions in
+`srcLayout` needs to match that of the current tensor.
+- `const TensorDimensionDescription &getNthDimDescription(size_t n) const`
+	- Given a dimension number `n`, Returns its description.
+- `const TensorDimensionDescription &getDimDescription(char name) const`
+	- Given a dimension name `name, Returns its description.
+- `llvm::ArrayRef<TensorDimensionDescription> getDims() const`
+	- Returns the description of all dimensions.
+- `size_t getNumDims()`
+	- Returns the number of dimensions.
+- `llvm::StringRef getLayoutName() const`
+	- Returns a string representing the name of the current tensor layout.
+- `bool isAnyLayout()`
+	- Returns true if the layout is "any" for all dimensions.
+
+It contains an array of [struct `TensorDimensionDescription`](https://github.com/pytorch/glow/blob/master/include/glow/Base/TensorLayoutUtils.h),
+One for each dimension of the tensor.
+The struct contains the following fields:
+
+- `size_t alignment`
+	- Expected alignment of the current dimension.
+- `uint8_t order`
+	- Expected order of the current dimension.
+- `char name`
+	- Expected name of the current dimension.
+
+As an example, lets assume we want to create the array-of-structs for Glow's
+backend-agnostic `NHWC` layout, it should look as follows:
+
+```
+static TensorDimensionDescription dimsNHWC[] = {
+    {.alignment = 1, .order = 0, .name = 'N'},
+    {.alignment = 1, .order = 1, .name = 'H'},
+    {.alignment = 1, .order = 2, .name = 'W'},
+    {.alignment = 1, .order = 3, .name = 'C'},
+};
+```
+
+The `alignment` field has been set to `1` for all dimensions because
+we do not have an alignment requirement pre-lowering.
+
+Given the `N` dimension, for example, we set the `order` value to `0`
+indicating that it is the first dimension of tensor.
+
+To help indicate its purpose, number of samples, we give it the name `N`.
+
+Since some operations can take any layout, giving we use the `*` character
+to annotate a dimension as wild.
+
+As such, the description of a four dimensional tensor for a data parallel
+operation would look as follows:
+
+```
+static TensorDimensionDescription dims4D[] = {
+    {.alignment = 1, .order = 0, .name = '*'},
+    {.alignment = 1, .order = 1, .name = '*'},
+    {.alignment = 1, .order = 2, .name = '*'},
+    {.alignment = 1, .order = 3, .name = '*'},
+};
+```
+
+It is *highly* recommended to explicitly pass the expected layout for any operation
+that is not data parallel.
+It is not, however, a strict requirement: if it is not provided the
+operation will be constructed with `*` layout.
+
+This is done for connivence and backward compatibility's sake. If the layouts
+are dropped the default verifier will assume the operation will produce an
+output is in the Glow's default format (i.e. `NHWC`).
+
+This does mean the verifier *might* fail if some operations have been annotated
+with `NCHW` while others are left as a wildcard. In a release build, if
+verifications are disabled, and assuming the graph is valid even if not
+correctly annotated, this will not cause a correctness issue on an
+Interpreter backend. This is not guaranteed on backends that use this (invalid)
+information for their backend-specific lowering and/or optimization.
+
+Another thing to note is that most optimization passes in Glow do not take
+a backend as an input, they might, however, be called post lowering. Layout
+verification will catch any (layout-requiring) transformation that breaks
+correctness due to implicitly assuming a canonical layout post-lowering.
+If a transformation requires Glow's canonical layout, it is its responsibility
+to verify that assumption.
+
+## Layout Requirements Interface
+
+Backends in Glow *may* derive from [base class `TensorLayoutCommon`](https://github.com/pytorch/glow/blob/master/include/glow/Graph/TensorLayout.h).
+Which includes the following virtual methods they can override:
+
+- `virtual TensorLayoutDescription getDefaultNDLayout(unsigned dims) const`
+
+  - This helper function takes a `unsigned dims` and returns the (current) default n-D layout.
+
+- `virtual TensorLayoutDescription getNthInputLayoutRequirements(const Node *node, size_t n) const`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth input `n`.
+
+- `virtual TensorLayoutDescription getNthResultLayoutRequirements(const Node *node, size_t n) const`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth result `n`.
+
+- `virtual std::array<TensorLayoutDescription, max_tensor_dimensions + 1> &getLayoutsForDims() const`
+
+  - This helper function returns an array of predefined layouts for all dimensions from `0-D` to Glow's max tensor layout dimension.
+
+An example of why backends may want to override such methods can be seen in the `OpenCL` backend:
+`OpenCL` Convolutions are more efficient in `NCHW` format, as such, we may lower a `ConvolutionNode`
+into a `NHWC` to `NCHW` transpose + convolution.
+The `OpenCL` verifier should expect `NCHW` for the input/output of the convolution instead of `NHWC`. 
+
+## Canonical Tensor Layout
+
+Before lowering a Glow graph into a specific, we introduce a "Canonical"
+representation that we expect for certain operations.
+This allows us to verify the graph after every transformation and may expose `GraphOptimizer` bugs [^tl0].
+[class `CanonicalTensorLayout`](https://github.com/pytorch/glow/blob/master/include/glow/Graph/TensorLayout.h)
+derives from `TensorLayoutCommon` and overrides the following functions:
+
+- `virtual TensorLayoutDescription getDefaultNDLayout(unsigned dims) const`
+
+  - Overrides the default `4-D` layout from "any" into `NHWC`
+
+- `virtual TensorLayoutDescription getNthInputLayoutRequirements(const Node *node, size_t n) const`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth input `n`.
+  - It returns Common layout constraints, for example, the input of `TransposeNode` is the same as layout of operation's result producing it.
+
+- `virtual TensorLayoutDescription getNthResultLayoutRequirements(const Node *node, size_t n) const`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth result `n`.
+  - It returns Common layout constraints, for example, `ConvolutionNode` should be in `NHWC` format.
+
+## Placeholders and Constants
+
+An important thing to note is that some operators may have a `Placeholder` or
+a `Constant` as their input. We may need to know a specific layout for said
+storage. For example, a Placeholder may need to be in `NHWC` format for a
+`ConvolutionNode`. However, we do not want to pollute the code by making
+this a hard requirement, especially since the canonical layout may accept
+anything for certain tensors (e.g. `1-D` tensor), as such, we introduce the
+notion of `ANY_LAYOUT` and initialize them with this wildcard by default.
+
+## Related Work
+
+Other machine learning frameworks introduced similar concepts, this is not a
+proposal unique to Glow, here are some notable mentions:
+
+### PlaidML
+
+Provides layout requirement information as a parameter to operations that need
+to know tensor layouts instead of setting a global layout that would apply to
+every operation. Allowing users to mix layouts throughout their network.
+
+PlaidML made the conscious decision to make the layout a property the operation
+instead of the tensor, making the implementation of certain operations more
+intuitive [^tl1].
+
+### TVM
+
+TOPI is the operator collection library for TVM [^tl2]. Certain TOPI operations
+include their layout requirements as a string. Here's layout section of
+`topi.nn.pool` taken from version 0.6 of the document:
+
+> layout (string) – Layout of the input data. The layout is supposed to be composed
+> of upper cases, lower cases and numbers, where upper case indicates a dimension
+> and the corresponding lower case with factor size indicates the split dimension.
+> For example, NCHW16c can describe a 5-D tensor of [batch_size, channel, height,
+> width, channel_block], in which channel_block=16 is a split of dimension channel.
+
+
+### XLA
+
+XLA adds backend specific layout constraints. Their CPU backend requires
+constant arrays to be column major when all of their users are dot operations [^tl3]. While
+Their GPU backend adds layout constraints on the cudnn custom-call instruction [^tl4].
+
+It is also worth taking a look at XLA's layout optimizer [^tl5], part of their
+effort to improve the out-of-the-box TensorFlow performance [^tl6].
+
+Another thing to note is that their alter layout pass [^tl7] is similar,
+in function, to the "Solver" we propose to automatically legalizes layouts in
+the future work section of this document.
+
+### MLIR
+
+Does not currently have such support, but there are ongoing discussions to add such
+support to MLIR Tensor Type [^tl8].
+
+## Future Work
+
+There are a few neat things we can, and probably should, do to expand this support:
+
+### Remove `enum ConvolutionLayout`
+
+Our string based representation is more generic and extendable as it is basically an
+extendable enum that can be used in the backends without touching the generic code base.
+
+### Remove shuffle arrays
+
+Some operations, such as `TransposeNode`, have a shuffle that tells them what to do.
+This can be deprecated and automatically deduced by specifying layout constraints.
+
+There is some discrepancy is the fact that with currently use both typed tensor, 
+with named dimensions, and explicitly indexed dimensions like we currently do
+everywhere in the code base, shuffle arrays being an example of that, This
+may lead to potential inconsistency in certain cases.
+We should gradually migrate towards typed tensors in the long run.
+
+### Introduce a "Solver" that automatically legalizes layouts
+
+Said solver will drastically reduce the complexity of loading models from other frameworks:
+We no longer need to insert transposes based on if we are importing `NHWC` or `NCHW`.
+We just need to annotate the `Placeholder` with the layout information we've get at load-time,
+and which we "forget" afterwards, and let the solver transpose said `Placeholder` to our
+canonical layout.
+
+First we will start with a "raw" state of non compliance, Then we have a loop to sink and
+clamp layout transformations together.
+
+### Remove backend specific nodes
+
+Today, Glow core and custom backends implicitly hard-code this knowledge about the operations
+into (backend-specific) nodes and code that works with them. This is pretty fragile and
+involves a lot of boiler plate code.
+
+Combining the proposed solver with the backend-specified layout constraints would improve
+this situation considerably:
+
+- The backend would return this information and Glow core could insert all the required layout transformations
+
+- The transformations can also be optimized "fore free": Glow currently optimizes `TransposeNode`:
+ - Multiple transposes can be combined into one
+ - Opposite transposes can eliminate each other
+
+- The functionality to insert the required layout transforms is handled by the Glow core,
+which removes a lot of code duplication from backends.
+
+[^tl0]: [Glow Issue: Fix bug in constant folding optimization](https://github.com/pytorch/glow/issues/3500)
+
+[^tl1]: [Tensor Layout Design Decision in PlaidML](https://github.com/plaidml/plaidml/blob/master/plaidml2/op/lib/design.md#tensor-layout)
+
+[^tl2]: [TVM Operator Inventory](https://docs.tvm.ai/api/python/topi.html)
+
+[^tl3]: [XLA CPU Layout Assignment](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/cpu/cpu_layout_assignment.cc)
+
+[^tl4]: [XLA GPU Layout Assignment](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/service/gpu/gpu_layout_assignment.cc)
+
+[^tl5]: [XLA Layout optimizer](https://github.com/tensorflow/tensorflow/blob/b6f7ce2b98b496886be4d900a6f88c24ae730f2c/tensorflow/core/grappler/optimizers/layout_optimizer.cc)
+
+[^tl6]: [TensorFlow Graph Optimizations](https://web.stanford.edu/class/cs245/slides/TFGraphOptimizationsStanford.pdf)
+
+[^tl7]: [XLA Alter Layout](https://github.com/dmlc/tvm/blob/025a6c8077cd1914bdd4132c6b86de007151344e/src/relay/pass/alter_op_layout.cc)
+
+[^tl8]: [Proposal to add layout attribute to MLIR Tensor Type](https://groups.google.com/a/tensorflow.org/forum/#!topic/mlir/sCaIEKm2RxA)
--- a/docs/TensorLayout.md
+++ b/docs/TensorLayout.md
@@ -25,7 +25,7 @@ Glow's string-based layout format is encoded as follows:
 
 1. A mandatory one character representing the current dimension. Either an  alphabetic letter or `*` (any layout).
 2. An optional token for the start of the current dimension's information: `[`.
-3. An optional namespace identifier for non-standard information, such as tiling, followed by `:`. Must have `[` from 2. in place. following said identifier, all subsequent data is considered as a "black box" until `]` is encountered.
+3. An optional namespace identifier for non-standard information, such as tiling, followed by `:`. Must have `[` from 2. in place. Following said identifier, all subsequent data is considered as a "black box" until `]` is encountered.
 4. Given that we have `[` from 2. in place, the closing bracket `]` for it.
 5. Optionally go back to 2.
 
@@ -70,9 +70,10 @@ Which includes the following virtual methods they can override:
 virtual bool isSatisfiedBy(TypeRef ty,
                                const TensorLayoutDescription &destLayout,
                                const TensorLayoutDescription *srcLayout) const
+                               ```
 	- This function checks if `ty` satisfies `destLayout` layout requirements, if `srcLayout` is provided for `ty`, take that into account.
 
-- `virtual std::array<TensorLayoutDescription, max_tensor_dimensions + 1> &getLayoutsForDims() const`
+- `virtual llvm::ArrayRef<TensorLayoutDescription> getLayoutsForDims() const`
 
   - This helper function returns an array of predefined layouts for all dimensions from `0-D` to Glow's max tensor layout dimension.
 
--- a/docs/TensorLayout.md
+++ b/docs/TensorLayout.md
@@ -21,124 +21,57 @@ For operations that do have layout requirements, Glow has an easily extendable
 string-based layout field. This allows backends to override Glow's default
 requirements without the hassle of creating a custom, backend-specific, node.
 
-The [class `TensorLayoutDescription`](https://github.com/pytorch/glow/blob/master/include/glow/Base/TensorLayoutUtils.h)
-represents the expected layout of tensors in Glow.
-It contains the following methods:
+Glow's string-based layout format is encoded as follows:
 
-- `bool isSameLayout(const TensorLayoutDescription &rhs) const`
-	- Returns true if both tensor layouts are the same.
-- `bool isSatisfiedBy(TypeRef ty, const TensorLayoutDescription *srcLayout) const`
-	- Returns true if the type `ty` satisfies the current layout.
-	- If `srcLayout` is provided, the verification is more complete,
-as it is taken into consideration, for example: the number of dimensions in
-`srcLayout` needs to match that of the current tensor.
-- `const TensorDimensionDescription &getNthDimDescription(size_t n) const`
-	- Given a dimension number `n`, Returns its description.
-- `const TensorDimensionDescription &getDimDescription(char name) const`
-	- Given a dimension name `name, Returns its description.
-- `llvm::ArrayRef<TensorDimensionDescription> getDims() const`
-	- Returns the description of all dimensions.
-- `size_t getNumDims()`
-	- Returns the number of dimensions.
-- `llvm::StringRef getLayoutName() const`
-	- Returns a string representing the name of the current tensor layout.
-- `bool isAnyLayout()`
-	- Returns true if the layout is "any" for all dimensions.
+1. A mandatory one character representing the current dimension. Either an  alphabetic letter or `*` (any layout).
+2. An optional token for the start of the current dimension's information: `[`.
+3. An optional namespace identifier for non-standard information, such as tiling, followed by `:`. Must have `[` from 2. in place. following said identifier, all subsequent data is considered as a "black box" until `]` is encountered.
+4. Given that we have `[` from 2. in place, the closing bracket `]` for it.
+5. Optionally go back to 2.
 
-It contains an array of [struct `TensorDimensionDescription`](https://github.com/pytorch/glow/blob/master/include/glow/Base/TensorLayoutUtils.h),
-One for each dimension of the tensor.
-The struct contains the following fields:
-
-- `size_t alignment`
-	- Expected alignment of the current dimension.
-- `uint8_t order`
-	- Expected order of the current dimension.
-- `char name`
-	- Expected name of the current dimension.
-
-As an example, lets assume we want to create the array-of-structs for Glow's
-backend-agnostic `NHWC` layout, it should look as follows:
-
-```
-static TensorDimensionDescription dimsNHWC[] = {
-    {.alignment = 1, .order = 0, .name = 'N'},
-    {.alignment = 1, .order = 1, .name = 'H'},
-    {.alignment = 1, .order = 2, .name = 'W'},
-    {.alignment = 1, .order = 3, .name = 'C'},
-};
-```
-
-The `alignment` field has been set to `1` for all dimensions because
-we do not have an alignment requirement pre-lowering.
-
-Given the `N` dimension, for example, we set the `order` value to `0`
-indicating that it is the first dimension of tensor.
-
-To help indicate its purpose, number of samples, we give it the name `N`.
-
-Since some operations can take any layout, giving we use the `*` character
-to annotate a dimension as wild.
-
-As such, the description of a four dimensional tensor for a data parallel
-operation would look as follows:
-
-```
-static TensorDimensionDescription dims4D[] = {
-    {.alignment = 1, .order = 0, .name = '*'},
-    {.alignment = 1, .order = 1, .name = '*'},
-    {.alignment = 1, .order = 2, .name = '*'},
-    {.alignment = 1, .order = 3, .name = '*'},
-};
-```
-
-It is *highly* recommended to explicitly pass the expected layout for any operation
-that is not data parallel.
-It is not, however, a strict requirement: if it is not provided the
-operation will be constructed with `*` layout.
-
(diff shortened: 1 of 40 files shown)


lib/Graph/Graph.cpp
```diff
@@ -3099,8 +3110,88 @@ insertAndReport(std::unordered_map<std::string, const Node *> &nameToNode,
...
+      isValid &= checkSameLayout(producerLayout, consumerLayout,
+                                 input.getType(), &N, inputName);
+    }
+  }
+  return isValid;
+}
+
+/// Verifies the correctness of tensor layouts in the function \p F based on
+/// Glow's "canonical" form.
+static bool verifyLayouts(const Function &F) {
```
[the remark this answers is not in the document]
shajrawi: Great idea! I am also moving the implementation somewhere else: getting linkage errors when building in shared libraries mode..

tests/unittests/GraphTest.cpp
```diff
@@ -1788,6 +1789,7 @@ TEST(Graph, testDumpStructure) {
   std::string mesN = K->toString();
   std::string expectMes = R"(Placeholder
 name : "input"
+layout : *
```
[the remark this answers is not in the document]
shajrawi: More complex. Currently we have `*` as the default value when constructing the placeholder.  We would need to change that and/or modify the creation method to count the number of dimensions. Same for constants. We can usually figure the the layout out from `*` so.. But I am open to making this change if we decide it is best.

tests/unittests/OperatorTest.cpp
```diff
@@ -7898,7 +7902,7 @@ static void testFlatten(glow::PlaceholderBindings &bindings, glow::Module &mod,
                         glow::Function *F, glow::ExecutionEngine &EE,
                         ElemKind DTy) {
   auto *tensor4D = createPlaceholderConditionallyQuantized(
-      mod, DTy, {3, 2, 4, 3}, "4D", false);
+      mod, DTy, {3, 2, 4, 3}, "4D", false, "NHWC");
```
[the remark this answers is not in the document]
shajrawi: Created `GLOW_DEFAULT_4D_LAYOUT` / replaced everywhere.

Re it being explicit: It is only the default value for 4D. placeholders can have dimensions != 4

tools/ClassGen/NodeGen.cpp
```diff
@@ -639,12 +639,14 @@ int main(int argc, char **argv) {
   BB.newNode("Reshape")
       .addInput("Input")
       .addMember(MemberType::VectorSizeT, "Dims")
+      .addMember(MemberType::String, "Layout")
       .addResultFromCtorArg()
       .setDocstring("Reshape the Input tensor to shape Dims.");
 
   BB.newNode("Transpose")
       .addInput("Input")
       .addMember(MemberType::VectorUnsigned, "Shuffle")
```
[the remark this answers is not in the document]
shajrawi: It would be clunky: manipulating strings in various portions of the code: we sometimes create shuffles by doing cleaver manipulations on the mask we want. just replacing the value of certain indices.

shajrawi:
Thanks for the initial review @opti-mix ! I addressed your comments and fixed the cmake / shared library issue.

examples/char-rnn.cpp
```diff
@@ -191,10 +191,10 @@ static Function *createNetwork(Module &mod, PlaceholderBindings &bindings,
 
   std::vector<NodeValue> resX;
   for (unsigned i = 0; i < numSteps; i++) {
-    auto *R =
-        F->createReshape("reshapeSelector", expectedX[i], {minibatchSize, 1});
+    auto *R = F->createReshape("reshapeSelector", expectedX[i],
+                               {minibatchSize, 1}, "*");
     auto *SM = F->createSoftMax("softmax", outputNodes[i], R);
-    auto *K = F->createReshape("reshapeSM", SM, {minibatchSize, 1, 128});
+    auto *K = F->createReshape("reshapeSM", SM, {minibatchSize, 1, 128}, "*");
```
opti-mix: Let's introduce something like a named constant called `ANY_LAYOUT` instead of explicitly using `"*"` everywhere.

opti-mix: Also, should it be the default value for this this parameter, so that we don't need to pass it explicitly everywhere?

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,277 @@
...
+/// Glow layouts for any specific number of dimensions.
+static std::array<TensorLayoutDescription, max_tensor_dimensions + 1>
+    layoutsForDims = {
+        layout0D, layout1D, layout2D, layout3D, layout4D, layout5D, layout6D,
+};
+
+static std::unordered_map<std::string, TensorLayoutDescription *>
+    layoutNameToLayoutDescription = {{
+        {"NCHW", &layoutNCHW},
+        {GLOW_DEFAULT_4D_LAYOUT, &layoutNHWC},
```
opti-mix: `GLOW_DEFAULT_4D_LAYOUT` should not be used here, because if we ever change the default, this line becomes incorrect. Please use a string here.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -212,7 +212,7 @@ static llvm::Expected<unsigned_t> getChannel(const ArgumentDictionaryTy &dict) {
   if (orderIt != dict.end()) {
     ASSIGN_VALUE_OR_RETURN_ERR(order, loadStr(orderIt->second));
   }
-  if (order == "NHWC") {
+  if (order == GLOW_DEFAULT_4D_LAYOUT) {
```
opti-mix: Please use a concrete string here, because any change to the definition of GLOW_DEFAULT_4D_LAYOUT would break this code.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -333,7 +333,8 @@ llvm::Error Caffe2ModelLoader::loadConv(const caffe2::OperatorDef &op,
   // Caffe2 "Conv" op always stores the weight as CKRS.
   Tensor wT;
   w->getPayload().transpose(&wT, NCHW2NHWC);
-  w = G_.getParent()->createConstant(w->getName(), std::move(wT));
+  w = G_.getParent()->createConstant(w->getName(), std::move(wT),
+                                     GLOW_DEFAULT_4D_LAYOUT);
```
opti-mix: Please use a concrete string here, because any change to the definition of GLOW_DEFAULT_4D_LAYOUT would break this code.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -343,7 +344,8 @@ llvm::Error Caffe2ModelLoader::loadConv(const caffe2::OperatorDef &op,
   // We expect the input to be NHWC.
   NodeValue finalIn;
   if (order == "NCHW") {
-    finalIn = G_.createTranspose(opName, in, NCHW2NHWC)->getResult();
+    finalIn = G_.createTranspose(opName, in, NCHW2NHWC, GLOW_DEFAULT_4D_LAYOUT)
```
opti-mix: Please use a concrete string here, because any change to the definition of GLOW_DEFAULT_4D_LAYOUT would break this code.

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -431,10 +433,11 @@ llvm::Error Caffe2ModelLoader::loadConvQuantized(const caffe2::OperatorDef &op,
   // C - output_depth, R - filter_height, S - filter_width, K - input_depth.
   // For Caffe2 "Int8Conv" and "Int8ConvRelu", the weights always follows the
   // "order" arg.
-  if (order != "NHWC") {
+  if (order != GLOW_DEFAULT_4D_LAYOUT) {
```
opti-mix: Please use a concrete string here, because any change to the definition of GLOW_DEFAULT_4D_LAYOUT would break this code.

lib/Backends/OpenCL/OpenCLTensorLayout.cpp
```diff
@@ -0,0 +1,122 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "OpenCLTensorLayout.h"
+#include "glow/Optimizer/GraphOptimizer/CompilationContext.h"
+
+using namespace glow;
+
+/// Definitions of different tensor layouts.
+static TensorDimensionDescription dimsNHWC[] = {
```
opti-mix: It is better to prefix the names of these static arrays with something, so that they do not look exactly the same as names in other CPP files. This would help when debugging. Otherwise a debugger will see to static variables with the same name and it would be almost impossible to refer to each of them.

tests/unittests/OperatorTest.cpp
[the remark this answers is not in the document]
opti-mix: Thanks for doing that. But it seems like it actually makes the code more error-prone in many cases. E.g. if the same line of code uses something like NHWC2NCHW and GLOW_DEFAULT_4D_LAYOUT in the same line, then changing GLOW_DEFAULT_4D_LAYOUT to something else than NHWC will break the logic. 

Overall, it seems like we really need to verify that the layout and the transformation/shuffle strings are compatible with each other. It is too easy to get them wrong currently.

tests/unittests/GraphTest.cpp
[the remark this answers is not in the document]
opti-mix: I don't have a strong opinion here. My main concern was that e.g. if two tensors have different number of dimensions and one of them has the `"*"` layout and another one may be a more concrete layout like `NCHW` we should be able to correctly handle this in our layout compatibility checking logic. As long as this is the case, I'm fine with this short way of designating the `any` layout.

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,277 @@
...
+    }
+    default:
+      return getLayoutsForDims()[dims.size()];
+    }
+  }
+  return getLayoutsForDims()[dims.size()];
+}
+
+/// \returns The index of node \p N input \p IN. NumInputs if not found.
+static unsigned getInputIdx(const Node *N, NodeValue IN) {
```
opti-mix: Why is IN upper-cased? I don't think we often use this kind of naming for NodeValues.

nadavrot:
@shajrawi Could you please add a section to /docs that describes the tensor layout proposal?

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,277 @@
...
+CanonicalTensorLayout::getDefaultNDLayout(unsigned dims) const {
+  if (dims == 4) {
+    return layoutNHWC;
+  }
+  return TensorLayoutCommon::getDefaultNDLayout(dims);
+}
+
+bool CanonicalTensorLayout::acceptsAnyLayout(const Node *node) const {
+  switch (node->getKind()) {
+  case Kinded::Kind::AddNodeKind:
```
nadavrot: I think that we have a flag that checks if an operation is an element-wise operation.

lib/IR/IRGen.cpp
```diff
@@ -439,7 +439,7 @@ void IRGenVisitor::post(Node *parent, Node *N) {
 }
 
 void IRFunction::generateIR(const Backend &B) {
-  assert(G_->verify() && "Invalid function");
+  // NOTE: Compilation context, which is not passed down to IR.
```
nadavrot: Why did you remove the assert?

lib/Graph/Node.cpp
```diff
@@ -299,6 +299,50 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Node *node) {
   node->dump(os);
   return os;
 }
+
+bool isDataParallel(const Node *node) {
```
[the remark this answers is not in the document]
shajrawi: Will do.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -690,8 +690,10 @@ llvm::Error ONNXModelLoader::loadPool(const ONNX_NAMESPACE::NodeProto &op,
     }
 
     node = G_.createMaxPool(opName, tr, kernels, strides, pads);
-    auto *res = G_.createTranspose(opName, NodeValue(node, 0), NHWC2NCHW);
-    auto *argmax = G_.createTranspose(opName, NodeValue(node, 1), NHWC2NCHW);
+    auto *res =
+        G_.createTranspose(opName, NodeValue(node, 0), NHWC2NCHW, "NCHW");
```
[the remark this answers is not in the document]
shajrawi: Yes. I do so in create transpose - see `Graph.cpp`.

lib/Backends/OpenCL/OpenCLTensorLayout.cpp
```diff
@@ -0,0 +1,122 @@
...
+  }
+  if (const auto APG = llvm::dyn_cast<AvgPoolGradNode>(node)) {
+    return getLayoutFromEnum(APG);
+  }
+
+  if (const auto *CN = llvm::dyn_cast<ConvolutionNode>(node)) {
+    switch (n) {
+    case ConvolutionNode::InputIndices::BiasIdx:
+      return &CanonicalTensorLayout::getInstance().getLayoutsForDims()[1];
+    default: { return getLayoutFromEnum(CN); }
```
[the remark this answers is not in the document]
shajrawi: Yup

include/glow/Graph/Graph.h
```diff
@@ -1340,6 +1356,9 @@ Node *recursiveClone(Function *newF, Node *node, NodeMap &currToNew);
 #define NHWC2NCHW                                                              \
   { 0u, 3u, 1u, 2u }
 
+/// Helper names for common tensor layouts.
+#define GLOW_DEFAULT_4D_LAYOUT "NHWC"
```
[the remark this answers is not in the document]
shajrawi: I removed the default layout `#define`, we can certainly talk about that but I feel that such a major change is separate than this PR.

shajrawi:
@nadavrot I amended the PR with a document that explains the motivation, interface, related work and future directions. See `docs/TensorLayout.md `.

@chadrosier, @pfierek, @tlepley-cadence, @mciprian13, and @vuzelac-cadence: I would appreciate any feedback you might have on this document and the changes it proposes to make the life of custom backends easier.

docs/TensorLayout.md
```diff
@@ -0,0 +1,169 @@
+## Tensor Layout
```
nadavrot: @shajrawi Thank you for the detailed doc!  I have a free questions. Could you give an example for the tensor layout and explain the semantics? Is the description optional (if we drop the layout hints will the program still be correct)? What does the optimizer need to know about the layout? Thank you.

shajrawi: I have expanded the `Tensor Layout` section of the document to answer these questions. I also added more references to the related work section and expanded the future work section.

docs/TensorLayout.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+It is *highly* recommended to explicitly pass the expected layout for any operation
+that is not data parallel.
+It is not, however, a strict requirement: if it is not provided the
+operation will be constructed with `*` layout.
+
+This is done for connivence and backward compatibility's sake. If the layouts
+are dropped the default verifier will assume the operation will produce an
+output is in `NHWC` format.
```
opti-mix: `is in the Glow's default format (i.e. NHWC)`.

docs/TensorLayout.md
```diff
@@ -0,0 +1,281 @@
...
+a backend as an input, they might, however, be called post lowering. Layout
+verification will catch any (layout-requiring) transformation that breaks
+correctness due to implicitly assuming a canonical layout post-lowering.
+If a transformation requires Glow's canonical layout, it is its responsibility
+to verify that assumption.
+
+## Layout Requirements Interface
+
+Backends in Glow *may* derive from [base class `TensorLayoutCommon`](https://github.com/pytorch/glow/blob/master/include/glow/Graph/TensorLayout.h).
+Which includes the following virtual methods they can override:
```
opti-mix: Could you add some wording about why backends may want to override such methods?

shajrawi: Added OpenCL convolutions as an example for why a backend may want to do that.

opti-mix:
@shajrawi Since this PR introduces a number of new data types for tensor layouts management, it would be nice to add a dedicated unit test file to check the basic functionalities of these data types. Could you please add such tests?

geoffberry:
@shajrawi I like the goal of being able to remove device-specific nodes, but this approach would need to be further extended to support tiled layouts before we could take advantage of it.

shajrawi:
@geoffberry Apologies if I am missing something, But a combination of `TensorDimensionDescription` and the string-based serialization / de-serialization already supports tiled layout? Here's the way I envision it:

Tiling should be representable by padding (which can be represented by the alignment field) + reshape + transpose. Reshape and transpose are already supported in Glow of course.

Given a layout-requiring node, the backend-specific lowering phase is allowed to augment the layout string, which is the layout metadata, however it likes. The string is just a serialized description of the layout, it should cover any backend-specific requirement that can be represented as an array of `TensorDimensionDescription`. There's nothing in the specification that prevents the backend from increasing/decreasing the number of dimensions. Something like TVM's `NCHW16c`, mentioned in related work, is representable (for example).

Since introducing a "solver" is not part of this commit, and is only a future-work proposal, it is the responsibility of the lowering phase to satisfy any constraints required. including a tiled layout requirement. i.e. adding the appropriate layout-transforming nodes (a combination of padding / reshape / transpose). I assume this is already being done by any backend that uses backend-specific nodes? It should continue to work as-is with a simple replacement of the backend specific convolution node (as an example) with Glow's node + a string representing the tiled layout. The `getNthInputLayoutRequirements` and `getNthResultLayoutRequirements` virtual methods need to be overridden, of course, to check / verify the tiled layout.

shajrawi:
We are waiting for some design discussions before proceeding with this PR, should be left open for now?

geoffberry:
@shajrawi Sorry, I've taken so long to reply.  Perhaps it would help clarify this proposal if you include an example of what NHWC -> NCHWc16 conversion would look like, spelling out the types, layouts, and transpose nodes, etc.

Another issue I'm not clear on is what the purpose of the `order` field is.  The description in the change doesn't include any cases where the `order` fields are anything other than `0...n-1`.  Perhaps the above example would clarify that as well?

shajrawi:
I've been thinking about the pros and cons of the current PR and having various discussions, here's what we propose:

First and foremost, the layout verification will be gated by a backend-specific flag that's 'off' by default. That way we can introduce this feature in a gradual manner without breaking anybody's workflow - especially for any closed-source backends out there. pre-lowering, the verifier will be set to 'on' to check up on the "canonical representation" transformations and catch any optimizer bugs (such as the ones mentioned earlier in this PR).

We do, however, reserve the right to change the default to 'on' in the future: the aim is to avoid any sudden disruptions and give everybody plenty of time to implement any backend-specific verifications and/or layout extensions such as splitting dimensions and tiling (more on that later).

Second, the canonical layout constraints representation will be the string-based version of what's in this (currently stale) PR. i.e. what's currently considered as the "serialized" representation. The `TensorLayoutDescription` struct will be considered as a convenience internal implementation detail for the open-source backends while exposed API will be dealing with strings:

This makes it easier for any closed-source backend with unique layout requirements to provide their own internal extensions to the string-based format without having to touch code out of the backend-specific folder. all while enjoying the same safety benefits as their open source counterparts. We do ask any such backend to append a special prefix to their serialized layout representation: that makes sure that any portion of glow that operates on the canonical layout, or any called layout interface method that has not been overridden yet, would treat the layouts as a "black box" and avoid making any decisions or assumption-breaking transformations (analogous to an unsafe basic block in a compiler).

As currently seen in the PR, everything will go through the `TensorLayoutCommon` interface, getting a pointer to said interface is the only extension to the backend base class, the interface will be extended with new overridable methods to support having everything going through it: including what it actually means for two layouts to be compatible (see the, currently stale, `isSatisfiedBy` function). The interface will, naturally, be modified to work on strings as its' output as rationalized above.

One thing I'd like to investigate further is if the layout string should be a property of the glow node, as it is currently the case and as is usually done (see the PlaidML discussion in the design document) or as a property of the type itself. I am currently considering all aspects of this design consideration.

I will update the PR with an updated implementation, design document and test cases within the next couple of weeks.

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,557 @@
...
+  }
+  // NHWC is the canonical default
+  return CanonicalTensorLayout::getInstance().getDefaultNDLayout(4);
+}
+
+std::string
+CanonicalTensorLayout::getNthInputLayoutRequirements(const Node *node,
+                                                     size_t n) {
+  auto baseReq = TensorLayoutCommon::getNthInputLayoutRequirements(node, n);
+  if (acceptsAnyLayout(node)) {
```
opti-mix: Should acceptsAnyLayout take n for the inputs/outputs number? Or do we say that all such nodes are always layout agnostic on all of their inputs/outputs?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -113,6 +114,43 @@ void run(Backend &backend, CompiledFunction &compiledF,
   context.movePlaceholderBindings().release();
 }
 
+static bool isCanonical(const NodeValue &RN, Backend &backend, Node *clonedC,
```
opti-mix: isCanonical -> isCanonicalLayout?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -113,6 +114,43 @@ void run(Backend &backend, CompiledFunction &compiledF,
...
+  auto default4D = TensorLayoutDescription(default4DStr);
+  if (resultLayout.getDims().size() == 4 &&
+      !canInstance.isSatisfiedBy(RN.getType(), default4D, &resultLayout)) {
+    return false;
+  }
+  return true;
+}
+
+static void
+bailOnNonCanLayout(Function *constEvaluationF, Module &mod,
```
opti-mix: bailOnNonCanLayout -> bailOnNonCanonicalLayout

And maybe move the comment before the function to explain what the function does.

docs/TensorLayout.md
```diff
@@ -0,0 +1,219 @@
...
+a layout field for said operations to reduce the dynamic memory consumption of
+the compiler.
+
+For operations that do have layout requirements, Glow has an easily extendable
+string-based layout field. This allows backends to override Glow's default
+requirements without the hassle of creating a custom, backend-specific, node.
+
+Glow's string-based layout format is encoded as follows:
+
+1. A mandatory one character representing the current dimension. Either an  alphabetic letter or `*` (any layout).
```
opti-mix: Please add a concrete example with `*` somewhere. It should be clear which layouts are compatible and what are the rules.

docs/TensorLayout.md
```diff
@@ -0,0 +1,219 @@
...
+
+Before lowering a Glow graph into a specific, we introduce a "Canonical"
+representation that we expect for certain operations.
+This allows us to verify the graph after every transformation and may expose `GraphOptimizer` bugs [^tl0].
+[class `CanonicalTensorLayout`](https://github.com/pytorch/glow/blob/master/include/glow/Graph/TensorLayout.h)
+derives from `TensorLayoutCommon` and overrides the following functions:
+
+- `std::string getDefaultNDLayout(unsigned dims) const`
+
+  - Overrides the default `4-D` layout from "any" into `NHWC`
```
opti-mix: Not only 4-D, but N-D layout, or?

docs/TensorLayout.md
```diff
@@ -0,0 +1,219 @@
...
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth input `n`.
+  - It returns Common layout constraints, for example, the input of `TransposeNode` is the same as layout of operation's result producing it.
+
+- `std::string getNthResultLayoutRequirements(const Node *node, size_t n)`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth result `n`.
+  - It returns Common layout constraints, for example, `ConvolutionNode` should be in `NHWC` format.
+
+## Placeholders and Constants
```
opti-mix: Can you add something about loaders ability to set layouts for constants and placeholders based on the network description.

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,557 @@
...
+    return false;
+  }
+  auto bracketPos = text.find(']');
+  assert(bracketPos != llvm::StringRef::npos && "Expected a closing bracket.");
+  return (bracketPos > nsPos);
+}
+
+// Serialization format -
+// The form for each dimension is as follows:
+// 1. (mandatory) one char representing the current dimension. Either an
```
opti-mix: Questions:
- Does it matter of it is upper case or lower case? E.g. in TVM they have different meanings
- Does it need to be a letter?
- Is one character enough for a dimension name?

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,557 @@
...
+  layoutNameToLayoutDescription.insert(std::make_pair(name, ret));
+  return ret;
+}
+
+std::string TensorLayoutCommon::getDefaultNDLayout(unsigned dims) const {
+  DCHECK_LE(dims, max_tensor_dimensions) << "Too many dimensions";
+  return getLayoutsForDims()[dims].getSerializedLayout();
+}
+
+std::string TensorLayoutCommon::getNthInputLayoutRequirements(const Node *node,
```
opti-mix: Could you add a doxygen comment about a general logic of such a function? Something like "Certain nodes are layout agnostic. Others expect their inputs/outputs to have a canonical format. For some layout agnostic nodes we need to look at the layout of their inputs to determine the layout of their outputs, e.g. ... "

docs/TensorLayout.md
```diff
@@ -0,0 +1,219 @@
...
+2. An optional token for the start of the current dimension's information: `[`.
+3. An optional namespace identifier for non-standard information, such as tiling, followed by `:`. Must have `[` from 2. in place. following said identifier, all subsequent data is considered as a "black box" until `]` is encountered.
+4. Given that we have `[` from 2. in place, the closing bracket `]` for it.
+5. Optionally go back to 2.
+
+As an example for this encoding, here's how we add alignment information,
+which is an officially supported extension, thus not requiring a namespace,
+followed by a backend-specific extension:
+`N[a=32][namespace_for_unsupported:<bla>]HWC` would represent 4-D tensor wherein
+`N` needs an alignment of 32 + some closed-backend requirements we don't know about.
```
opti-mix: "some closed-source backends" or "some private backends"

lib/Graph/TensorLayout.cpp
shajrawi: I refactored it: it accepts *any* for all input and output just for data parallel nodes, however that is not correct for nodes that can accept multiple layouts as input and propagate it to their output in canonical, but not all, backends. I did not add an `n` but I did make that clearer and split it into two functions.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
shajrawi: done

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
shajrawi: done - there's a large block of code inside that explains it, but I added a short description on top.

docs/TensorLayout.md
shajrawi: I added a new parsing test case for `*` AND I updated the document to explain that :)

docs/TensorLayout.md
shajrawi: good point - changed. in canonical we only care about 4-D, but, theoretically, just like the method's name implies, it could be any n-D

docs/TensorLayout.md
shajrawi: Done.

lib/Graph/TensorLayout.cpp
shajrawi: * No it can be either upper or lower, but we can change that if it matters?
* Either an ascii-table english-language letter or a `*` - no emojis :D 
* I assume so? makes it much easier to parse / smaller code, but should be easy to extend that if we change our mind.. I don't see why we'd need more than one.

lib/Graph/TensorLayout.cpp
shajrawi: Done - added that to the header file.

docs/TensorLayout.md
shajrawi: Done.

include/glow/Graph/Graph.h
```diff
@@ -173,26 +176,34 @@ class Module final {
   ///@{
 
   Placeholder *createPlaceholder(ElemKind T, llvm::ArrayRef<size_t> dims,
-                                 llvm::StringRef name, bool isTrainable);
+                                 llvm::StringRef name, bool isTrainable,
+                                 const std::string layout = ANY_LAYOUT);
```
opti-mix: IIRC, the linter suggested making all such `const std::string layout` arguments `const std::string &layout`.

include/glow/Graph/TensorLayout.h
```diff
@@ -0,0 +1,194 @@
...
+  std::string dims_[max_tensor_dimensions];
+  /// The serialization of the layout.
+  std::string serializedLayout_;
+  /// Expected number of dimensions.
+  size_t numDims_;
+
+public:
+  virtual ~TensorLayoutDescription() = default;
+  /// Constructs this helper class from a serialized string representation.
+  TensorLayoutDescription(std::string text);
```
opti-mix: Do you modify `text`? If not, should it be `const std::string &` or `StringRef`? Also, the name `text` is not very descriptive. Maybe `layout` or `layoutStr` or something like that?

include/glow/Graph/TensorLayout.h
```diff
@@ -0,0 +1,194 @@
...
+  /// \returns layout name.
+  llvm::StringRef getSerializedLayout() const { return serializedLayout_; }
+  /// \returns true if the layout is "*" in all dimensions.
+  bool isAnyLayout();
+  std::string getDebugDesc() const;
+
+protected:
+  /// parse helper: get the custom extensions information. the default, virtual,
+  /// implementation just ignores all the data until the end token.
+  virtual void parseCustomExtensions(llvm::StringRef &text, unsigned idx);
```
opti-mix: Seeing `llvm::StringRef &` is a bit unusual. But I see why I'd use it. Essentially it represents your parsing context. Hopefully it is enough for the future use-cases, otherwise we'd need to extend these APIs.

include/glow/Graph/TensorLayout.h
```diff
@@ -0,0 +1,194 @@
...
+                                                     size_t n);
+
+  /// \returns true if type \p ty satisfies the \p destLayout layout. If \p
+  /// srcLayout is provided, it is taken into account as well.
+  virtual bool isSatisfiedBy(TypeRef ty,
+                             const TensorLayoutDescription &destLayout,
+                             const TensorLayoutDescription *srcLayout) const;
+
+  /// \return layouts for all tensor dimensions.
+  virtual std::array<TensorLayoutDescription, max_tensor_dimensions + 1> &
```
opti-mix: Why `std::array` and not e.g. `llvm::ArrayRef`?

include/glow/Graph/Graph.h
shajrawi: solved, but the main linter issue i had is in switch-case statements where the internal linter disagreed with the external linter :)

include/glow/Graph/TensorLayout.h
shajrawi: The g++ 5 compiler does not like that, even though clang/llvm has no problem with it. complains that it cannot convert the c-style array (if we remove std::array) into llvm's ArrayRef

opti-mix: @shajrawi Could you use some kind of `llvm::makeArrayRef` here to convert `layoutsForDims` into an `ArrayRef`?

shajrawi: Done

examples/fr2en.cpp
```diff
@@ -339,14 +339,14 @@ void Model::loadDecoder() {
...
-        F_->createReshape("decoder.reshape", topK->getIndices(), {batchSize_});
+    lastWordIdx = F_->createReshape("decoder.reshape", topK->getIndices(),
+                                    {batchSize_}, "N");
     outputs.push_back(lastWordIdx);
   }
 
   Node *concat = F_->createConcat("decoder.output.concat", outputs, 0);
   Node *reshape = F_->createReshape("decoder.output.reshape", concat,
-                                    {MAX_LENGTH, batchSize_});
+                                    {MAX_LENGTH, batchSize_}, "*");
```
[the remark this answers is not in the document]
shajrawi: good catch - fixed

examples/fr2en.cpp
```diff
@@ -339,14 +339,14 @@ void Model::loadDecoder() {
     Node *FC = F_->createFullyConnected("decoder.outFC", hidden, outW, outB);
     auto *topK = F_->createTopK("decoder.topK", FC, 1);
 
-    lastWordIdx =
-        F_->createReshape("decoder.reshape", topK->getIndices(), {batchSize_});
+    lastWordIdx = F_->createReshape("decoder.reshape", topK->getIndices(),
+                                    {batchSize_}, "N");
```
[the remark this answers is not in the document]
shajrawi: Currently we don't make extensive use of named tensor, so `"N"` is probably not mandatory for the current backends and current verifications, but, lets say we add a different 1-D layout, for example a private backend adds it during lowering, and it is *NOT* `"N"`,  then the verifier should fail.

tools/ClassGen/NodeGen.cpp
```diff
@@ -684,12 +684,14 @@ int main(int argc, char **argv) {
   BB.newNode("Reshape")
       .addInput("Input")
       .addMember(MemberType::VectorSizeT, "Dims")
+      .addMember(MemberType::String, "Layout")
```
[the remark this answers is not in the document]
shajrawi: Our plan is to move the layout information from nodes to Glow types. this can be done in a follow-up commit, even as a first task for someone motivated, during an operator's construction it will propagate the layout string (a pointer to global hash of strings to save on size). so we will not add a member called layout.
said follow-up is straightforward / just a lot of manual labor if we are fine with the design - no point in something fancy in here :)

lib/Backends/OpenCL/OpenCLTensorLayout.cpp
```diff
@@ -0,0 +1,122 @@
...
+    return &oclLayoutNCHW;
+  }
+  return &oclLayoutNHWC;
+}
+
+/// \returns either NHWC or NCHW layout based on the instruction's layout enum
+/// if it has one. Else returns nullptr. This will be removed and refactored
+/// if/when we move to using strings for all layout specifications and get rid
+/// of the enum.
+static const TensorLayoutDescription *getLayouForTempEnumRep(size_t n,
```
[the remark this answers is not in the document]
shajrawi: done

lib/Optimizer/GraphOptimizer/Lower.cpp
```diff
@@ -171,7 +172,10 @@ static void lowerFullyConnectedGradNode(Function *F, CompilationContext &cctx,
   // dx = dout * w.T
   auto *wT = F->createTranspose("fcg.wT", FCG.getWeights(), {1, 0});
   auto *dx2 = F->createMatMul("fcg.dot", dout, wT);
-  auto *dx = F->createReshape("fcg.inG", dx2, FCG.getInput().getType()->dims());
+  auto *dx = F->createReshape(
+      "fcg.inG", dx2, FCG.getInput().getType()->dims(),
+      CanonicalTensorLayout::getInstance().getNthInputLayoutRequirements(&FCG,
+                                                                         0));
```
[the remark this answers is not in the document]
shajrawi: done

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -1594,8 +1607,11 @@ static NodeValue tryToOptimizeConcatOfRehapes(Function *F, ConcatNode *CN) {
     return NodeValue(nullptr);
   }
   auto *newCN = F->createConcat(CN->getName(), newConcatInputs, dim);
-  return F->createReshape(CN->getInputs().front().getNode()->getName(), newCN,
-                          CN->getResult().dims());
+  return F->createReshape(
+      CN->getInputs().front().getNode()->getName(), newCN,
+      CN->getResult().dims(),
+      CanonicalTensorLayout::getInstance().getNthResultLayoutRequirements(CN,
+                                                                          0));
```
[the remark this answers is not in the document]
shajrawi: done

include/glow/Graph/Graph.h
```diff
@@ -1376,6 +1397,8 @@ llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Function &F);
 
 llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const Function *F);
 
+#define NHWC2HWNC                                                              \
```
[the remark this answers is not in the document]
shajrawi: done

docs/TensorLayout.md
```diff
@@ -0,0 +1,235 @@
...
+
+- `virtual std::string getNthInputLayoutRequirements(const Node *node, size_t n)`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth input `n`.
+
+- `virtual std::string getNthResultLayoutRequirements(const Node *node, size_t n)`
+
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth result `n`.
+
+- ```
```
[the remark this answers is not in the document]
shajrawi: good catch!

docs/TensorLayout.md
```diff
@@ -0,0 +1,235 @@
...
+
+For operations that do have layout requirements, Glow has an easily extendable
+string-based layout field. This allows backends to override Glow's default
+requirements without the hassle of creating a custom, backend-specific, node.
+
+Glow's string-based layout format is encoded as follows:
+
+1. A mandatory one character representing the current dimension. Either an  alphabetic letter or `*` (any layout).
+2. An optional token for the start of the current dimension's information: `[`.
+3. An optional namespace identifier for non-standard information, such as tiling, followed by `:`. Must have `[` from 2. in place. following said identifier, all subsequent data is considered as a "black box" until `]` is encountered.
```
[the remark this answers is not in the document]
shajrawi: done

lib/Graph/TensorLayout.cpp
```diff
@@ -0,0 +1,581 @@
...
+  if (auto *TN = llvm::dyn_cast<TransposeNode>(node)) {
+    // If the result of Transpose is a concrete layout, try to use this specific
+    // layout.
+    if (auto *layout = getLayoutFromName(TN->getLayout(),
+                                         layoutNameToLayoutDescription_)) {
+      return layout->getSerializedLayout();
+    }
+    // Dynamically form the layout description for transposes.
+    auto input = TN->getInput();
+    auto inputLayout = getNthInputLayoutRequirements(node, 0);
```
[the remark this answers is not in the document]
shajrawi: because there's only one input for transpose?

shajrawi:
Thanks @jfix71 ! I added a new commit to address the nits

lib/Graph/TensorLayout.cpp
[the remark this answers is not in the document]
jfix71: Ah yeah misread it -- so then yeah I'd use TransposeNode::InputIdx here.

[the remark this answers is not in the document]
shajrawi: no worries - good suggestion: changed that for clarity :)

docs/TensorLayout.md
```diff
@@ -0,0 +1,236 @@
...
+  - This function takes an operator `Node *node` and returns the layout requirements of the Nth result `n`.
+
+- ```
+virtual bool isSatisfiedBy(TypeRef ty,
+                               const TensorLayoutDescription &destLayout,
+                               const TensorLayoutDescription *srcLayout) const
+                               ```
+	- This function checks if `ty` satisfies `destLayout` layout requirements, if `srcLayout` is provided for `ty`, take that into account.
+
+- `virtual std::array<TensorLayoutDescription, max_tensor_dimensions + 1> &getLayoutsForDims() const`
```
opti-mix: The return type needs to be updated to return ArrayRef, after you changed the signature.

shajrawi: good catch - done! (I already changed the signature and used make array ref).

opti-mix:
@shajrawi Looking good! I think it is ready to be merged.

opti-mix:
@jfix71 Do you have any further comments or are you OK with merging it?

PR #3507, opened 2019-09-12
Add option to IR optimize() to inhibit instrumentation

chadrosier:
Summary:
This is needed to avoid instrumenting the code multiple times if the
optimize() function is called from a backend.


chadrosier:
Thanks for the feedback @opti-mix and @jfix71; I agree with your suggestions.  Unfortunately, I don't have the bandwidth at the moment.  I'm going to close this pull request now, but I hope to get back around to this soon.

PR #3512, opened 2019-09-14
Added freeCompilationResources method to compiledFunction

gcatron:
Summary:
Added freeCompilationResources method to compiledFunction
 and added implementation of it for OpenCLFunction. 
This method allows for freeing resources in a compiledFunction that are no longer needed after the function has been loaded on a device.
Added pass in provisioner to call new method

Documentation: Updated Backends.md to include new method.

Test Plan: ninja check

### DIFF
--- a/include/glow/Runtime/Provisioner/Provisioner.h
+++ b/include/glow/Runtime/Provisioner/Provisioner.h
@@ -64,8 +64,10 @@ private:
   /// List of available DeviceManagers added during initialization.
   std::vector<DeviceManager *> devices_;
 
-  /// Helper function to cleanup a failed provision call.
-  void cleanupProvision(llvm::ArrayRef<std::string> names);
+  /// Helper function to cleanup a provision call. On a success free resources
+  /// that are no longer needed by the compiledFunctions. On failure free the
+  /// compiledFunctions that were created.
+  void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
 };
 } // namespace runtime
 } // namespace glow
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -124,6 +124,10 @@ OpenCLFunction::OpenCLFunction(std::unique_ptr<IRFunction> F,
   setTraceInfo(std::move(traceInfo));
 }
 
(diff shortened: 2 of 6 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -124,6 +124,10 @@ OpenCLFunction::OpenCLFunction(std::unique_ptr<IRFunction> F,
   setTraceInfo(std::move(traceInfo));
 }
 
+void OpenCLFunction::freeCompilationResources() {
+  runtimeBundle_.freeConstants();
```
beicy: Can I ask why only constants need to be freed?

gcatron: They are copied for initialization of the device after which they are no longer needed so they are freed. The symbolTable is still needed.

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -64,8 +64,8 @@ class Provisioner final {
   /// List of available DeviceManagers added during initialization.
   std::vector<DeviceManager *> devices_;
 
-  /// Helper function to cleanup a failed provision call.
-  void cleanupProvision(llvm::ArrayRef<std::string> names);
+  /// Helper function to cleanup a provision call.
+  void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
```
beicy: Please add a few words about \p failure.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -239,9 +239,17 @@ llvm::Error Provisioner::removeFunction(llvm::StringRef name) {
...
 
-void Provisioner::cleanupProvision(llvm::ArrayRef<std::string> names) {
+void Provisioner::cleanupProvision(llvm::ArrayRef<std::string> names,
+                                   bool failure) {
   std::lock_guard<std::mutex> functionLock(functionsLock_);
   for (auto &name : names) {
     activeFunctions_.erase(name);
+    if (failure) {
+      // Remove any functions added before the failure.
+      functions_.erase(name);
```
beicy: here do we need to free the resource as well?

Issue #3436, opened 2019-08-20
[Training][Accelerators] Increasing performance by `updateOutputPlaceholders` on demand


pfierek:
Copying of memory between host and device during training should be limited as much as possible.
In particular, there is no need to update the host-side weights every iteration, this can only be done after each epoch or even the whole learning process.
 
Currently in glow it is assumed that results (e.g. weights) are available after each `Function.execute()`.
Maybe `ExecutionEngine` should have the `updateOutputPlaceholders` helper function to update outputs on demand?
 
What do You think? I am open to discussion and other suggestions for resolving this issue.

jackm321:
@putivsky @mortzur

PR #3513, opened 2019-09-14
refactoring CPU device memory allocation out from function execution

mortzur:
Summary:

LLVMCompiledFunction::execute(...) allocates (and then frees) memory for activations and weights on each call.
This PR refactors the allocation part to CPUDeviceManager::addNetwork(..).
CPUBuffer is the struct owning the device memory. A different CPUBuffer is created for each function.
CPUBindings are used to pass the memory addresses to the compiled function.

This is all very similar to the OpenCL implementation, with some major distinctions:
LLVM is not a backend itself. CPU is the backend and CPUFunction is derived from LLVMCompiledFunction. This creates a small linkage mess.
I'd love to get feedback about how it may constructed better.

This is the CPU backend, so the change is not very impactful. It's a helper feature from experiments on training and P2P, enabling not to update (copy) output/intermediate placeholders on each iteration and handle special specific cases of input placeholders. 
Addressing issue #3436 .

This also improves the situation a bit for training which calls runBatch(...) with a large number of iterations.
For example, on resnet50 training using CIFAR10, it saves ~5% time on each epoch:
before:
  216.9277 (100.0%)  12.8826 (100.0%)  229.8103 (100.0%)  229.9044 (100.0%)  Training
after:
  209.2006 (100.0%)   2.2592 (100.0%)  211.4598 (100.0%)  211.4995 (100.0%)  Training

Test Plan:
ninja test

### DIFF
--- a/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
+++ b/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
@@ -20,8 +20,25 @@
 
 #include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
+#include "glow/ExecutionContext/ExecutionContext.h"
 
 namespace glow {
+
+/// CPUDeviceBindings inherits from DeviceBindings, it contains per run
+/// device specific information used to run a compiled function on a specific
+/// device.
+struct CPUDeviceBindings : DeviceBindings {
+  CPUDeviceBindings(uint8_t *activationsBuffer, uint8_t *weightsBuffer)
+      : DeviceBindings("CPU"), deviceActivationsBuffer{activationsBuffer},
+        deviceWeightsBuffer{weightsBuffer} {}
+
+  ~CPUDeviceBindings() {}
+
+  /// Non-owning pointers. Memory is owned by the device manager.
+  uint8_t *deviceActivationsBuffer;
+  uint8_t *deviceWeightsBuffer;
+};
+
 /// A Glow IR function compiled using LLVM.
 class LLVMCompiledFunction : public CompiledFunction {
 public:
--- a/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
+++ b/include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
@@ -24,16 +24,21 @@
 
 namespace glow {
 
-/// CPUDeviceBindings inherits from DeviceBindings, it contains per run
+/// LLVMDeviceBindings inherits from DeviceBindings, it contains per run
 /// device specific information used to run a compiled function on a specific
 /// device.
-struct CPUDeviceBindings : DeviceBindings {
-  CPUDeviceBindings(uint8_t *activationsBuffer, uint8_t *weightsBuffer)
-      : DeviceBindings("CPU"), deviceActivationsBuffer{activationsBuffer},
+class LLVMDeviceBindings : public DeviceBindings {
+public:
+  LLVMDeviceBindings(std::string backend, uint8_t *activationsBuffer,
+                     uint8_t *weightsBuffer)
+      : DeviceBindings(backend), deviceActivationsBuffer{activationsBuffer},
         deviceWeightsBuffer{weightsBuffer} {}
 
-  ~CPUDeviceBindings() {}
+  ~LLVMDeviceBindings() {}
 
+  friend class LLVMCompiledFunction;
+
+protected:
   /// Non-owning pointers. Memory is owned by the device manager.
   uint8_t *deviceActivationsBuffer;
   uint8_t *deviceWeightsBuffer;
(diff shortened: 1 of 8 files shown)


include/glow/LLVMIRCodeGen/LLVMCompiledFunction.h
```diff
@@ -20,8 +20,25 @@
 
 #include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
+#include "glow/ExecutionContext/ExecutionContext.h"
 
 namespace glow {
+
+/// CPUDeviceBindings inherits from DeviceBindings, it contains per run
```
gcatron: Seeing as how this is used in llvmCompiledFunction I'd just name it llvmDeviceBindings.

mortzur: Done. I also added CPUDeviceBindings just to keep clear naming and structure when used by CPUDeviceManager.

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -39,6 +39,18 @@ DeviceManager *createCPUDeviceManager(const DeviceConfig &config) {
   return new CPUDeviceManager(config);
 }
 
+CPUBuffer::CPUBuffer(size_t activationsSize, size_t activations_alignment,
+                     size_t weightsSize, size_t weights_alignment) {
+  activationsBuffer_ =
+      (uint8_t *)alignedAlloc(activationsSize, activations_alignment);
```
opti-mix: We usually use `static_cast` or similar for this purpose.

lib/Backends/CPU/CPUDeviceManager.h
```diff
@@ -24,13 +24,41 @@
 namespace glow {
 namespace runtime {
 
+/// A class that contains a CPU device buffer. It frees the buffer when it
+/// is destroyed.
+class CPUBuffer {
+  /// The OpenCL buffer being stored.
```
opti-mix: Remove OpenCL. It is a result of copy&paste.

lib/Backends/CPU/CPUDeviceManager.cpp
```diff
@@ -108,6 +120,12 @@ void CPUDeviceManager::addNetworkImpl(const Module *module,
       func.second->getRuntimeBundle().collectConstants(module);
     }
     functions_.emplace(func.first, func.second);
+
+    auto &bundle = func.second->getRuntimeBundle();
+    std::unique_ptr<CPUBuffer> deviceBuffer = llvm::make_unique<CPUBuffer>(
```
opti-mix: Does it mean that we allocate buffers for activations and mutable weights at the network deployment time? Shouldn't this happen per run of the deployed network? Or may be I miss something?

mortzur: Yes.
This is a point worth dwelling on because this is the point of this PR - it seems that the CPU does this allocation on execution while other backends (openCL, Habana) allocates on deployment. It's a time-memory tradeoff.
It is separated from the follow-up issue I was about to address - which is the actual loading and updating of the placeholders - which really doesn't have to happen for all placeholders each run. 

For example on Habana:
https://github.com/pytorch/glow/blob/63e16c654ee872c9b4539165cad71e20e5a66656/lib/Backends/Habana/HabanaDeviceManager.cpp#L186
https://github.com/pytorch/glow/blob/63e16c654ee872c9b4539165cad71e20e5a66656/lib/Backends/Habana/HabanaFunction.cpp#L61

opti-mix: How would you run multiple instances of the same network concurrently if there is only one buffer for activations and mutable weights? IMHO, you can share the buffer for constants among multiple running instances of the same network, but you cannot share the memory allocated for activations and mutable weights. Glow's CPU backend was able to run multiple instances of the same network concurrently without any issues using the existing buffer allocation scheme IIRC. But it wouldn't be possible anymore after your change, or?

mortzur: True. If the same DeviceManager instance is used, I wouldn't be able to run a few requests concurrently. Let me check this.

opti-mix: @gcatron Are DeviceManagers supposed to support execution of a few instances of the same NN model concurrently? If so, does it mean that OpenCL and Habana do not support it yet based on what @mortzur said above?

gcatron: For any of the DeviceManagers that inherit from QueueBacked, (OpenCL, Interpreter, and CPU) they use a threadpool of size 1 to run their requests which mean they don't support concurrent execution. Requests can be queued concurrently but only one is executed at a time.

mortzur: The solution for openCL and Habana device managers is to use buffer pools (circular buffer in both cases I think) to enable k-parallel requests. I can do a similar thing here.

PR #3514, opened 2019-09-14
[Onnxifi] Better error message

yinghai:
Summary:
Give better error message when input tensor is larger than the compiled spec. Adding shape will help on which dimension we got wrong. 

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -158,13 +158,13 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
 
     if (inOnnxTensorSize > inPhPtr->getType()->size()) {
       std::stringstream ss;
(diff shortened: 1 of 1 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -157,8 +157,18 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
     }
 
     if (inOnnxTensorSize > inPhPtr->getType()->size()) {
+      std::stringstream ss;
+      for (const auto i : inOnnxTensorDims) {
+        ss << i << ", ";
```
bertmaher: Would be nicer without the trailing comma, too bad we don't have folly::join :-)

yinghai: Python style. :)

Issue #2737, opened 2019-04-17
Enable LLVM and Glow assertions in Release mode


mssimpso:
It would be great if LLVM and Glow assertions were enabled by default in Release mode for continuous integration builds. We enable these internally and frequently hit them when merging in upstream changes.

nadavrot:
I like this idea. Do you have a patch that enables assertions already? Also, have you measured tthe impact on the test suite runtime? cc: @rdzhabarov

rosierm:
Can't provide a patch at the moment (waiting on corporate to sign CLA), but one can be easily derived from the LLVM project.  The impact on the test suite would be entirely compile-time; we've not measured it, but happy to do so.

rdzhabarov:
> Glow assertions were enabled by default in Release mode for continuous integration builds

That's already part of what we run in our "debug" circleci build.

As for LLVM it would be an interesting idea.

> would be entirely compile-time;
That would require building LLVM during the ci runs (unless it's prebuild in the docker image) which could be time consuming.
I'd also like to see how much overhead would that bring. Obviously, I'd like to keep our CI time within a sane limits.

geoffberry:
I understand not wanting to do an LLVM build in your ci runs.  We use a prebuilt LLVM release build with assertions enabled, and have workarounds for a handful of LLVM asserts in our Glow internal codebase currently.

rdzhabarov:
I think it would make sense to either install that into docker container or download from S3.

@geoffberry how do you run that on internal CI?

geoffberry:
@rdzhabarov I'm not sure what you're asking.  Do you mean how do we link with the prebuilt version of LLVM that we are using, or what tests are we running in CI, or something else?

rdzhabarov:
I'm curious about your CI setup if any

rosierm:
Our CI is configured with asserts enabled in release mode for both LLVM and Glow.  We use a prebuilt version of LLVM and protobuf and don't perform any type of containerization.  With each commit we run all of the unit tests along with the image-classifier application for all Caffe2 inputs and for all quantization schemes.

rosierm:
Attaching results from running the Glow integrated tests with asserts on and off for Glow/LLVM.
LLVM version: 7.0.1
Glow: 71f2ec208b1306f422da7fb1562254eb225d32ae

[int_tests_asserts_off.txt](https://github.com/pytorch/glow/files/3253442/int_tests_asserts_off.txt)
[...]

Issue #3505, opened 2019-09-12
Unchecked Expected introduced in 3706588


chadrosier:
The below commit introduced a number of unchecked llvm::Expected errors:
Partitioner refactoring 3 (#3384)

https://github.com/pytorch/glow/blob/d12d6d09b6ab994df271231f5c4a5bab44c3dae5/tests/unittests/PartitionerTest.cpp#L252
Needs something like EXPECT_TRUE((bool)dagList);

https://github.com/pytorch/glow/blob/d12d6d09b6ab994df271231f5c4a5bab44c3dae5/tests/unittests/PartitionerTest.cpp#L346
IIUC, this is a negative test with an Error.  The bool operator for Expected sets Unchecked = true if there's an actual Error.  Therefore, this expect remains unchecked.

https://github.com/pytorch/glow/blob/d12d6d09b6ab994df271231f5c4a5bab44c3dae5/tests/unittests/PartitionerTest.cpp#L893
Same as previous.

https://github.com/pytorch/glow/blob/d12d6d09b6ab994df271231f5c4a5bab44c3dae5/tests/unittests/PartitionerTest.cpp#L922
Same as previous.

https://github.com/pytorch/glow/blob/d12d6d09b6ab994df271231f5c4a5bab44c3dae5/tests/unittests/PartitionerTest.cpp#L954
Same as previous.

This requires a build of LLVM with asserts enabled.

Would you mind taking a look @beicy?

beicy:
Sure! Thanks! @chadrosier

chadrosier:
Fixed in 16dd2fd.

PR #3515, opened 2019-09-16
Create glow::Error and glow::Expected

jackm321:
_Note for review: Most real changes are in `Error.h`, `Error.cpp`, and `ErrorTest.cpp` and the rest is mostly codemod to switch from `llvm::Error/Expected` to `glow::Error/Expected`._

Summary:
Removes dependency on `llvm::Error` and ensures Errors are don't go unchecked by verifying this in debug build.

`glow::Error/Expected` are mostly drop-in replacements for `llvm::Error/Expected` but exclude some features not used in Glow such as multiple error value types, reference types contained in Expected, etc. They also have a couple of different features like `Error::empty()` which creates "pre-checked" `Error`s.

Documentation:
doxygen
fixes: #2737, #3505

Test Plan:
added `ErrorTest` unit tests
`ninja check`

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -16,40 +16,219 @@
 #ifndef GLOW_SUPPORT_ERROR_H
 #define GLOW_SUPPORT_ERROR_H
 
+#include <cassert>
+#include <memory>
 #include <mutex>
 #include <type_traits>
 
-#include "llvm/Support/Error.h"
-#include "llvm/Support/FormatVariadic.h"
-
 #include <glog/logging.h>
 
+/// NOTE: please only use code and macros that resides outside of the detail
+/// namespace in Error.h and Error.cpp so as to preserve a layer of
+/// abstraction between Error/Expected types and the specific classes that
+/// implement them.
+
 namespace glow {
-/// NOTE This should not be used directly, instead use EXIT_ON_ERR or
-/// TEMP_EXIT_ON_ERR. Callable that takes an llvm::Error or llvm::Expected<T>
-/// and exits the program if the Error is not equivalent llvm::Error::success()
-/// or the Expected<T> contains an error that is not equivalent
-/// llvm::Error::success()
-/// TODO: replace this with a function that will print file and
-/// line numbers also.
-extern llvm::ExitOnError exitOnErr;
+/// Consumes an Error \p err and \returns true iff the error contained an
+/// ErrorValue. Calls the log method on ErrorValue if the optional argument \p
+/// log is passed.
+#define ERR_TO_BOOL(...)                                                       \
+  (glow::detail::errorToBool(__FILE__, __LINE__, __VA_ARGS__))
 
-/// Is true_type only if applied to llvm::Error or a descendant.
+/// Consumes an Error \p err and \returns "success" if it does not contain an
+/// ErrorValue or the result of calling the log() if it does.
+#define ERR_TO_STRING(err) (glow::detail::errorToString((err)))
+
+/// Consumes an Error \p err. Calls the log method on the ErrorValue if the
+/// optional argument \p log is passed.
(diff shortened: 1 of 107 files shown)


lib/Support/Error.cpp
```diff
@@ -52,4 +49,109 @@ bool OneErrOnly::containsErr() {
...
 }
+
+namespace detail {
+std::string GlowErrorValue::logToString() const {
+  std::stringstream ss;
+  log(ss);
+  return ss.str();
+}
+
+// static
```
opti-mix: Do you plan to remove this `static`?

include/glow/Support/Error.h
```diff
@@ -16,40 +16,219 @@
...
+#include <memory>
 #include <mutex>
 #include <type_traits>
 
-#include "llvm/Support/Error.h"
-#include "llvm/Support/FormatVariadic.h"
-
 #include <glog/logging.h>
 
+/// NOTE: please only use code and macros that resides outside of the detail
```
opti-mix: Are most of the changes in this file "inspired" by LLVM's Error.h, i.e. is it pretty-much a copy of the LLVM's code with minor changes? Is there any possibility to change LLVM's behavior without literally re-defining all these types? Probably not. But it is annoying to have so much boilerplate code ;-(

lib/Support/Error.cpp
jackm321: I'll remove it

include/glow/Support/Error.h
jackm321: @opti-mix yeah it is overall pretty similar to LLVM's Error and Expected with a changes like abstracting out the state checking. The thing that it does is pretty dead simple so there isn't a lot of room for creative implementations.

As I mentioned in person, I would actually like to design this a bit differently, have only Expected<T> and model the Error use case with a template specialization for Expected<void> so we only have a single type everywhere. However for the sake of sanity, I think we should switch to a near drop-in replacement first and then consider later whether we want to go in a different direction like this instead of attempting to do everything at once.

Because llvm::Error's checking is disabled at compile time I don't think there is a clean and easy way to turn this back on by wrapping it externally and removing our dependency on llvm::Errors allows us the freedom to change things later as mentioned above.

jackm321:
@opti-mix Thank you for reviewing this monster, really appreciate it!

include/glow/Support/Error.h
```diff
@@ -16,40 +16,219 @@
...
+using ErrorEmpty = detail::GlowErrorEmpty;
+using ErrorValue = detail::GlowErrorValue;
+template <typename T> using Expected = detail::GlowExpected<T>;
+
+/// NOTE: detail namespace contains code that should not be used outside of
+/// Error.h and Error.cpp. Please instead use types and macros defined above.
+namespace detail {
+/// enableCheckingErrors is used to enable assertions that every Error and
+/// Expected has its status checked before it is destroyed. This should be
+/// enabled in debug builds but turned off otherwise.
```
opti-mix: BTW, a quick question: What if this could be controlled by a command-line flag instead? It would be disabled by default in release builds, but one could enable it via such a command-line option if needed? Would it bring any value?

jackm321: That would be worse because when it is disabled at build time
1. Error and Expected use less space because no need to store `CheckState:: checked_`
2. The code could be slightly faster because the code to do the checking is completely non-existent so no need even to check a flag to see if the state should be checked.
Also this is really a debugging mechanism so I think it makes sense to enable it in only debug mode. My view on it is that ideally, ensuring that Errors are always checked would be done statically but this is Rust so having some debug assertions is probably the next best thing, but we shouldn't be putting that check into Release. What do you think?

opti-mix: Yeah, makes sense. Thanks for the explanation!

PR #3518, opened 2019-09-16
[GraphOptimization] Replace Tile->Add with BatchedAdd

kumudb:
Summary:
Added compiler pass to detect and replace a Tile operation followed by an Add operation with a BatchedAdd to decrease memory footprint and memory reuse among other things.

Documentation:
See comments in the code.

[Optional Fixes #issue]
2208

Test Plan:
Added new unit test to GraphOptzTest

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -2861,6 +2861,67 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
   return changed;
 }
 
+// Fold Tile -> Add into BatchedAdd wherever applicable.
+bool FoldTileAddIntoBatchedAdd::run(Function *F,
+                                    const CompilationContext &cctx) {
+  LOG_SCOPE(F->getLogContext(), getName());
+
+  bool changed = false;
+  for (const auto &node : F->getNodes()) {
+    const auto *addNode = dyn_cast<AddNode>(&node);
+    if (!addNode) {
+      continue;
+    }
+
+    NodeValue batchNode, addedNode;
+    const auto &LHS = addNode->getLHS();
+    const auto &RHS = addNode->getRHS();
+    const TileNode *tileNode = nullptr;
+
+    // Check if LHS is a tile.
+    if ((tileNode = dyn_cast<TileNode>(LHS))) {
+      batchNode = RHS;
+      addedNode = tileNode->getInput();
+    }
+    // Check if RHS is a tile.
+    else if ((tileNode = dyn_cast<TileNode>(RHS))) {
+      batchNode = LHS;
+      addedNode = tileNode->getInput();
+    }
+    // If neither LHS or RHS is a tile, nothing to do.
+    else {
+      continue;
+    }
+
+    // If the tiling of the added node is not along the 0th axis,
+    // 'Add' cannot be replaced with 'BatchedAdd'.
+    if (tileNode->getAxis() != 0) {
+      continue;
+    }
+
+    auto oldDims = addedNode.dims();
+    // If the 0th dimension of the added node is not 1,
+    // then reducing dimension via reshaping is more complicated.
+    // Hence, Add will not be replaced with BatchedAdd.
+    if (oldDims.size() == 0 || oldDims[0] != 1) {
+      continue;
+    }
+
+    // Reshape the added node to create a slice for the batched add
+    // such that its dim size is one less than that of the batch.
+    const auto newDims = oldDims.take_back(oldDims.size() - 1);
+    auto *slice = F->createReshape(tileNode->getName().str() + "_reshape",
+                                   addedNode, newDims);
+
+    // Create a new batched add node to replace existing add node.
+    auto *newBA = F->createBatchedAdd(addNode->getName().str() + "_batched_add",
+                                      batchNode, slice);
+    addNode->getResult().replaceAllUsesOfWith(newBA);
+    changed = true;
+  }
+  return changed;
+}
+
 void glow::fold(Function *F, CompilationContext &cctx) {
   LOG_SCOPE(F->getLogContext(), "glow::fold")
 
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -2210,6 +2210,57 @@ TEST_F(GraphOptz, mergeBANodes) {
   EXPECT_EQ(countNodeKind(F_, Kinded::Kind::BatchedAddNodeKind), 1);
 }
 
+// Check that we are able to replace
+// Add(I, tile(B)) with -> BatchedAdd(I, B).
+TEST_F(GraphOptz, FoldTileAddIntoBatchedAdd) {
+  auto *batch =
+      mod_.createPlaceholder(ElemKind::FloatTy, {3, 1, 2}, "batch", false);
+  auto *added = mod_.createConstant(ElemKind::FloatTy, {1, 1, 2}, "added");
+  auto *addedTiled = F_->createTile("addedTiled", added, 3, 0);
+  auto *add = F_->createAdd("add", batch, addedTiled);
+  auto *save = F_->createSave("save", add);
+  auto *output = save->getPlaceholder();
+
+  bindings_.allocate(batch)->getHandle() = {2, 2, 3, 3, 4, 4};
+  added->getPayloadMutable().getHandle() = {1, 1};
+  bindings_.allocate(output);
+
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::TileNodeKind), 1);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::AddNodeKind), 1);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::BatchedAddNodeKind), 0);
+
+  ASSERT_TRUE(F_->verify());
+  ::glow::optimize(F_, CompilationMode::Infer);
+  ASSERT_TRUE(F_->verify());
+
+  // Check that the Tile node and the Add node is replaced by
+  // a BatchedAdd node.
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::TileNodeKind), 0);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::AddNodeKind), 0);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::BatchedAddNodeKind), 1);
+
+  // Verify the correctness of the input to BatchedAdd operator.
+  // The correctness of BatchedAdd operator itself is verified
+  // by operator's unit tests.
+  Tensor expectedBatch(ElemKind::FloatTy, {3, 1, 2});
+  expectedBatch.getHandle() = {2, 2, 3, 3, 4, 4};
+  Tensor expectedSlice(ElemKind::FloatTy, {1, 2});
+  expectedSlice.getHandle() = {1, 1};
+  for (auto &node : F_->getNodes()) {
+    auto *recvdBANode = llvm::dyn_cast<BatchedAddNode>(&node);
+    if (!recvdBANode) {
+      continue;
+    }
+    auto *recvdBatch = llvm::dyn_cast<Placeholder>(recvdBANode->getBatch());
+    auto *recvdSlice = llvm::dyn_cast<Constant>(recvdBANode->getSlice());
+    EXPECT_TRUE(recvdBatch->dims().equals({3, 1, 2}));
+    EXPECT_TRUE(recvdSlice->dims().equals({1, 2}));
+    EXPECT_TRUE(bindings_.get(recvdBatch)->isEqual(expectedBatch));
+    EXPECT_TRUE(recvdSlice->getPayload().isEqual(expectedSlice));
+    break;
+  }
+}
+
 // Check that we are able to eliminate concat nodes.
 TEST_F(GraphOptz, concatElim) {
   Node *input =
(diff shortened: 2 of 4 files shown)


tests/unittests/GraphOptzTest.cpp
```diff
@@ -2210,6 +2210,34 @@ TEST_F(GraphOptz, mergeBANodes) {
...
+  ASSERT_TRUE(F_ -> verify());
+  ::glow::optimize(F_, CompilationMode::Infer);
+  ASSERT_TRUE(F_ -> verify());
+
+  //Check that the Tile node and the Add node is replaced by
+  //a BatchedAdd node.
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::TileNodeKind), 0);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::AddNodeKind), 0);
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::BatchedAddNodeKind), 1);
+}
```
nickgg: You should verify that the BatchedAdd is correct here, do we usually compile & run the resultant function?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
 }
 
+// Fold Tile -> Add into BatchedAdd wherever applicable.
+bool FoldTileAddIntoBatchedAdd::run(Function *F,
+                                    const CompilationContext &cctx) {
+  LOG_SCOPE(F->getLogContext(), getName());
+
+  bool changed = false;
+  auto &nodes = F->getNodes();
+  for (auto &node : nodes) {
```
SplitInfinity:
```suggestion
  for (auto &node : F->getNodes()) {
```

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
   return changed;
 }
 
+// Fold Tile -> Add into BatchedAdd wherever applicable.
+bool FoldTileAddIntoBatchedAdd::run(Function *F,
+                                    const CompilationContext &cctx) {
+  LOG_SCOPE(F->getLogContext(), getName());
+
+  bool changed = false;
+  auto &nodes = F->getNodes();
```
SplitInfinity: It doesn't look like this is used anywhere else so I think you can plug the definition directly into the for loop.

tests/unittests/GraphOptzTest.cpp
kumudb: I fixed the issue you pointed above. Please take a look.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+      continue;
+    }
+
+    NodeValue batchNode, addedNode;
+    auto &LHS = addNode->getLHS();
+    auto &RHS = addNode->getRHS();
+    auto *tileNode = dyn_cast<TileNode>(LHS);
+
+    // Check if LHS is a tile.
+    if (tileNode) {
```
SplitInfinity:
```suggestion
    if (tileNode = dyn_cast<TileNode>(LHS)) {
```
to be consistent with the other cast below. I think this is more "LLVM-esque" anyway.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+      batchNode = RHS;
+      addedNode = tileNode->getInput();
+    }
+    // Check if RHS is a tile.
+    else if ((tileNode = dyn_cast<TileNode>(RHS))) {
```
SplitInfinity:
```suggestion
    else if (tileNode = dyn_cast<TileNode>(RHS)) {
```
I think there are extra braces here? I also remember getting some compiler error/warning when I tried something like this without the extra braces so I might be wrong here.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+
+    // If the tiling is not along the 0th axis,
+    // Add cannot be replaced with BatchedAdd.
+    if (tileNode->getAxis() != 0) {
+      continue;
+    }
+    auto oldDims = addedNode.dims();
+    // If the 0th dimension is not 1, then reducing dimension
+    // via reshaping is more complicated. Hence, Add will not be
+    // replaed with BatchedAdd.
```
SplitInfinity:
```suggestion
    // replaced with BatchedAdd.
```

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+    if (oldDims.size() < 2 || oldDims[0] != 1) {
```
SplitInfinity: Why do you skip the optimization if `oldDims.size() < 2`? You should still be able to do it without reshaping if `oldDims.size() == 1` and `oldDims[0] == 1`, right?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,62 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+      continue;
+    }
+    auto newDims = oldDims.take_back(oldDims.size() - 1);
+
+    auto *slice = F->createReshape("reshape", addedNode, newDims);
+    // Create a new batched node with batch node
```
SplitInfinity: Can you clarify this comment? Maybe something like: `Create a new BatchedAddNode with the input to the TileNode and the other input of the AddNode being replaced`.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -2210,6 +2210,58 @@ TEST_F(GraphOptz, mergeBANodes) {
...
+
+  // Verify the correctness of the input to BatchedAdd operator.
+  // The correctness of BatchedAdd operator itself is verified
+  // by operator's unit tests.
+  Tensor expectedBatch(ElemKind::FloatTy, {3, 1, 2});
+  expectedBatch.getHandle() = {2, 2, 3, 3, 4, 4};
+  Tensor expectedSlice(ElemKind::FloatTy, {1, 2});
+  expectedSlice.getHandle() = {1, 1};
+  auto &nodes = F_->getNodes();
+  for (auto &node : nodes) {
```
SplitInfinity:
```suggestion
  for (auto &node : F_->getNodes()) {
```

tests/unittests/GraphOptzTest.cpp
bertmaher: The GraphOptz tests usually just verify that the optimized Graph is the expected one.  I guess the theory is we should have another OperatorTest that verifies that the operators actually do what we expect.

nickgg: @bertmaher I'm fine with that only if we very that the expected one is correct - e.g. has the right inputs and outputs. It's not enough to just replace Tile+Add with BatchedAdd if they wouldn't produced the right output.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Extra braces are to make the compiler not show warnings like if it was meant to be ==

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Sure

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

tests/unittests/GraphOptzTest.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,63 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+    if ((tileNode = dyn_cast<TileNode>(LHS))) {
+      batchNode = RHS;
+      addedNode = tileNode->getInput();
+    }
+    // Check if RHS is a tile.
+    else if ((tileNode = dyn_cast<TileNode>(RHS))) {
+      batchNode = LHS;
+      addedNode = tileNode->getInput();
+      // If neither LHS or RHS is a tile, nothing to do.
+    } else {
```
SplitInfinity: Nit: the `else if` above is on a new line but this `else` is not.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,63 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
   return changed;
 }
 
+// Fold Tile -> Add into BatchedAdd wherever applicable.
+bool FoldTileAddIntoBatchedAdd::run(Function *F,
+                                    const CompilationContext &cctx) {
+  LOG_SCOPE(F->getLogContext(), getName());
+
+  bool changed = false;
+  for (auto &node : F->getNodes()) {
```
SplitInfinity:
```suggestion
  for (const auto &node : F->getNodes()) {
```
I like `const`-ing things wherever possible. `node` and `addNode` are not mutated so I think this should be possible but I could be wrong.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,63 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+      // If neither LHS or RHS is a tile, nothing to do.
+    } else {
+      continue;
+    }
+
+    // If the tiling of the added node is not along the 0th axis,
+    // Add cannot be replaced with BatchedAdd.
+    if (tileNode->getAxis() != 0) {
+      continue;
+    }
```
SplitInfinity: Nit: Add a new line below this.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,63 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+    // then reducing dimension via reshaping is more complicated.
+    // Hence, Add will not be replaced with BatchedAdd.
+    if (oldDims.size() == 0 || oldDims[0] != 1) {
+      continue;
+    }
+
+    // Reshape the added node to create a slice for the batched add
+    // such that its dim size is one less than that of the batch.
+    auto newDims = oldDims.take_back(oldDims.size() - 1);
+    auto *slice = F->createReshape("reshape", addedNode, newDims);
```
SplitInfinity: It's a good idea to prefix/suffix this node name with whatever it's replacing so that someone can make sense of it when examining the graph. If this optimization runs multiple times, all of the `Reshape` nodes will be named `reshape`, `reshape1`, `reshape11`, ..., which is not very useful.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2861,6 +2861,63 @@ bool FoldChannelShuffle::run(Function *F, const CompilationContext &cctx) {
...
+
+    // Create a new batched add node to replace existing add node.
+    auto *newBA = F->createBatchedAdd("batch_add", batchNode, slice);
```
SplitInfinity:
```suggestion
    auto *newBA = F->createBatchedAdd("batched_add", batchNode, slice);
```
Same here.

SplitInfinity:
Also, one of the tests in `RELEASE_WITH_EXPENSIVE_TESTS` failed. I don't think it's due to this PR, but maybe rebase and see if it goes away?

kumudb:
> Also, one of the tests in `RELEASE_WITH_EXPENSIVE_TESTS` failed. I don't think it's due to this PR, but maybe rebase and see if it goes away?
Yes, I agree. I will try to rebase it as long as I don't have conflicts.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
kumudb: Done!

SplitInfinity:
I think you made a mistake during the rebase. All of your commits should come after the ones authored by others that have already landed.

SplitInfinity:
Make sure to squash all of your commits into one before landing.

PR #3519, opened 2019-09-16
IO policy for placeholders

mortzur:
Summary:
On some scenarios, we want to indicate which placeholder/backing_tensor should be loaded to the device or read from the device on a certain run. For example:
* Inference when the model is partitioned between multiple cards and peer-to-peer communication is enabled.
* Training - weights tensors

This PR suggests to enable this by adding a member called IO Policy (with corresponding enum class) to each placeholder. I added a usage example for CPU/LLVM.
Feedback would be appreciated :)

### DIFF
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -130,6 +130,21 @@ public:
   bool verify() const;
 };
 
+/// An enum representing the IO policy of the tensor associated with the
+/// placeholder.
+enum class IoPolicy : unsigned char {
+  // Always load, always read
+  InOut,
+  // Always load, read on demand (i.e. for debug)
+  Input,
+  // Do not load, read on demand
+  Intermediate,
+  // Load at init only, read on demand
+  IntermediateWithInit,
+  // Do not load, always read
+  Output,
+};
+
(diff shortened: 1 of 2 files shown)


include/glow/Graph/Nodes.h
```diff
@@ -130,6 +130,21 @@ class Constant : public Storage {
   bool verify() const;
 };
 
+/// An enum representing the IO policy of the tensor associated with the
+/// placeholder.
+enum class IoPolicy : unsigned char {
+  // Always load, always read
```
opti-mix: Period missing at the end of sentence.

opti-mix: Also, could you elaborate in the comments what is meant by e.g. load, output and read or read on demand? Load from where (e.g. from host)? Read by whom? I suspect you mean something like:
- InOut is copied from a host to a device and later copied back to the host
- Input does not need to be copied from a host as it is in the device memory already
- Intermediate is probably for activations
- Output is for copying from a device to a host

You may want to make it more clear in the comment.

Also, I suspect that it may have some overlap with @nickgg's work on improving the support for device-specific tensors. Please check with him.

mortzur:
Discussing in context of DRT. Closing.

PR #3520, opened 2019-09-17
[Graph] Only tile RHS of a BatchMatMul if LHS batch size is > 1

SplitInfinity:
**Summary**
This commit modifies `Function::createBatchMatMul` to insert a `Tile`
operation to make the first dimension (the batch dimension) of the RHS
operand equal to that of the LHS operand only if the latter is greater
than 1. If it is equal to 1, the `Tile` is not needed.

Test Plan:
This commit adds a new unit test for `BatchMatMul` in which the batch
dimension of the LHS operand is 1. All unit tests pass.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -1440,7 +1440,7 @@ BatchMatMulNode *Function::createBatchMatMul(llvm::StringRef name,
     RHS = createExpandDims(name.str() + ".reshapeRHS", RHS, {0});
   }
   // If necessary, Tile the RHS input so it matches the numBatches of LHS.
-  if (RHS.dims()[0] == 1) {
+  if (RHS.dims()[0] == 1 && LHS.dims()[0] != 1) {
     RHS = createTile(name.str() + ".tileRHS", RHS, LHS.dims()[0], /*axis */ 0);
   }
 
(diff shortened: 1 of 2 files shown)


SplitInfinity:
Adding @jfix71 as reviewer based on GitHub's suggestion :)

rdzhabarov:
As an alternative solution.

What do you think about making an optimization which removes noop Tile ops.

Like Tile one time along some dimension can be completely removed.
This way we could simplify `createBatchMatMul` and potentially remove those useless Tiles ops.

rdzhabarov:
I'll take over since Jordan is on vacation this week.

SplitInfinity:
I think it's a good idea in general to have such a no-op Tile optimization to clean up any that result from successive rounds of optimization. In general, I am in favour of avoiding creating a situation that needs the optimization in the first place. 

As an aside, the code in `createBatchMatMul` doesn't handle the case in which both inputs are 3D but their batch dimensions don't match either. I think the ideal solution would be to tile both up to the lowest common multiple of their batch dimensions.

For context, I stumbled across this optimization when trying to differentiate a `Function` with a `BatchedMatMul`. `Tile` is currently not differentiable so the program crashed, and I noticed that the Tile was tiling only once. I don't mind using a no-op Tile optimisation to deal with this, but that means we may need to optimize the graph a bit before differentiating it? Eventually Tile differentiation support will be added, but a `Tile` no-op optimization pass may not work once that `TIle` node has been differentiated.

PR #3525, opened 2019-09-18
[OpenCL] speed up local size search

pjaaskel:
Summary:

Speed up the local size search. The starting point can be rather large especially for CPU devices, and decrementing it one unit at a time can take quite a long time.

Test Plan:

Passes ninja check for x86_64 pocl-hsa OpenCL target. I needed the 2nd patch in the PR since I had an OpenCL-only Glow build.

### DIFF
--- a/lib/Backends/Interpreter/tests/InterpreterBackendCorrectnessTest.cpp
+++ b/lib/Backends/Interpreter/tests/InterpreterBackendCorrectnessTest.cpp
@@ -17,4 +17,7 @@
 
 using namespace glow;
 
-std::set<std::string> glow::backendTestBlacklist = {};
+std::set<std::string> glow::backendTestBlacklist = {
+    // Interpreter does not support kernel stacking yet.
+    "dataParallelStackingTest/0",
+};
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -308,15 +308,15 @@ static void getMaxLocalWorkgroupSize(cl_kernel kernel, cl_device_id device,
   // constraints:
   size_t totalWorkPrevDims = 1;
   for (int i = 0, e = global.size(); i < e; i++) {
-    local[i] = L;
+    local[i] = std::min(L, WIS[i]);
+    local[i] = std::min(local[i], WGS / totalWorkPrevDims);
 
-    while (global[i] % local[i] || L % local[i] || local[i] > WIS[i] ||
-           local[i] * totalWorkPrevDims > WGS) {
+    while (global[i] % local[i] || L % local[i]) {
       local[i]--;
     }
 
     // Remember how much work we are doing in this dimension. Use it to make
-    // sure that the next dimenstions don't exceed the total allowed workgroup
+    // sure that the next dimensions don't exceed the total allowed workgroup
     // size.
     totalWorkPrevDims *= local[i];
   }
--- a/tests/unittests/BackendCorrectnessTest.cpp
+++ b/tests/unittests/BackendCorrectnessTest.cpp
@@ -64,7 +64,7 @@ TEST_P(BackendCorrectnessTest, extract3Dtest) {
   Tensor out1;
   Tensor out2;
 
-  inferExtract3D(&inputs, &out1, "CPU");
+  inferExtract3D(&inputs, &out1, backendName_);
   inferExtract3D(&inputs, &out2, "Interpreter");
 
   EXPECT_TRUE(out1.isEqual(out2));
@@ -373,7 +373,8 @@ TEST_P(BackendCorrectnessTest, nonSquarePaddingConvTest) {
   CHECK_IF_ENABLED();
   Tensor out1;
   Tensor out2;
-  inferNonSquarePaddingConv(&out1, "CPU");
+
+  inferNonSquarePaddingConv(&out1, backendName_);
   inferNonSquarePaddingConv(&out2, "Interpreter");
 
   EXPECT_TRUE(out1.isEqual(out2));
@@ -382,9 +383,11 @@ TEST_P(BackendCorrectnessTest, nonSquarePaddingConvTest) {
 /// This non-square kernel test targets the DKKC8 optimization.
 TEST_P(BackendCorrectnessTest, nonSquareKernelConvTest) {
   CHECK_IF_ENABLED();
+
   Tensor out1;
   Tensor out2;
-  inferNonSquareKernelConv(&out1, "CPU");
+
+  inferNonSquareKernelConv(&out1, backendName_);
   inferNonSquareKernelConv(&out2, "Interpreter");
 
   EXPECT_TRUE(out1.isEqual(out2));
@@ -395,7 +398,7 @@ TEST_P(BackendCorrectnessTest, nonSquareStrideConvTest) {
   CHECK_IF_ENABLED();
   Tensor out1;
   Tensor out2;
-  inferNonSquareStrideConv(&out1, "CPU");
+  inferNonSquareStrideConv(&out1, backendName_);
   inferNonSquareStrideConv(&out2, "Interpreter");
 
   EXPECT_TRUE(out1.isEqual(out2));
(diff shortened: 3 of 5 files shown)


tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -58,6 +58,9 @@ TEST_P(BackendCorrectnessTest, convTest) {
 
 TEST_P(BackendCorrectnessTest, extract3Dtest) {
   CHECK_IF_ENABLED();
+#ifndef GLOW_WITH_CPU
```
nickgg: What's the reason for this? CHECK_IF_ENABLED should handle this

rdzhabarov: it seems that the test actually is broken.

> inferExtract3D(&inputs, &out1, "CPU");
 inferExtract3D(&inputs, &out2, "Interpreter");

needs to be
inferExtract3D(&inputs, &out1, backendName_);
inferExtract3D(&inputs, &out2, "Interpreter");

rdzhabarov:
@pjaaskel could you properly fix tests where incorrect backend is supplied. See the comment above.

tests/unittests/PartitionerTest.cpp
```diff
@@ -615,6 +618,9 @@ TEST_F(PartitionerTest, SimpleHeterogeneousPartitioning) {
 /// Heterogeneous Partition. In this test, "Mul" is not supported in
 /// Interpreter backend, and "Sub" is not supported in CPU backend.
 TEST_F(PartitionerTest, heterogeneousPartitioningWithNonSupportedNodes) {
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: @beicy it seems that PartitionerTest implicitly assumes that Glow is built with the CPU support.
Is there any good way to make these tests more robust?

tests/unittests/BackendCorrectnessTest.cpp
pjaaskel: Done!

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -371,9 +375,13 @@ TEST_P(BackendCorrectnessTest, groupConvTest) {
 /// This test targets the DKKC8 optimization.
 TEST_P(BackendCorrectnessTest, nonSquarePaddingConvTest) {
   CHECK_IF_ENABLED();
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: I think this check can be removed

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -382,9 +390,14 @@ TEST_P(BackendCorrectnessTest, nonSquarePaddingConvTest) {
 /// This non-square kernel test targets the DKKC8 optimization.
 TEST_P(BackendCorrectnessTest, nonSquareKernelConvTest) {
   CHECK_IF_ENABLED();
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: ditto

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -393,9 +406,12 @@ TEST_P(BackendCorrectnessTest, nonSquareKernelConvTest) {
 /// This non-square stride test targets the DKKC8 optimization.
 TEST_P(BackendCorrectnessTest, nonSquareStrideConvTest) {
   CHECK_IF_ENABLED();
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: ditto

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -404,9 +420,12 @@ TEST_P(BackendCorrectnessTest, nonSquareStrideConvTest) {
 /// This test targets the DKKC8 opt correctionimization.
 TEST_P(BackendCorrectnessTest, convDKKC8Test) {
   CHECK_IF_ENABLED();
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: ditto

tests/unittests/BackendCorrectnessTest.cpp
```diff
@@ -256,6 +256,10 @@ TEST_P(BackendCorrectnessTest, dataParallelStackingTest) {
     bb.createDeallocActivationInst("dealloc", act);
   }
 
+#ifndef GLOW_WITH_CPU
```
rdzhabarov: this is fine due to usage of `MockCPUBackend`. Would be good to get rid of this as well.

pjaaskel:
Done.

tests/unittests/BackendCorrectnessTest.cpp
opti-mix: @pjaaskel Could you have a look at how `glow::backendTestBlacklist` is used in Glow. It is described in docs and it is also used by e.g. Habana backend. It would be much cleaner to just blacklist any tests that don't need to be executed by the OpenCL backend. This way there is no need for any preprocessor-based conditional compilation.

opti-mix: @pjaaskel Ping! Do you still plan to land this PR?

pjaaskel: Yes, it's in my task list. Hopefully I'll get into it soon.

pjaaskel:
I moved the one CPU-backend dependent check to the blacklist mechanism and also disabled the whole PartitionerTest in case CPU is not enabled since it seems a large chunk of the tests assume it is. Also I rebased the commits on top of the current master.

tests/unittests/CMakeLists.txt
```diff
@@ -335,7 +335,12 @@ target_link_libraries(PartitionerTest
                         Partitioner
                         gtest
                         TestMain)
-add_glow_test(PartitionerTest ${GLOW_BINARY_DIR}/tests/PartitionerTest --gtest_output=xml:PartitionerTest.xml)
+
+# There are multiple test cases in PartitionerTest that work only with CPU
```
opti-mix: @beicy Is it really necessary to for the PartitionerTest to use the CPU backend?

lib/Backends/Interpreter/tests/InterpreterBackendCorrectnessTest.cpp
```diff
@@ -17,4 +17,9 @@
 
 using namespace glow;
 
-std::set<std::string> glow::backendTestBlacklist = {};
+std::set<std::string> glow::backendTestBlacklist = {
+#ifndef GLOW_WITH_CPU
+    // Requires the CPU target due to the use of MockCPUBackend.
+    "dataParallelStackingTest/0",
```
opti-mix: Good catch! It seems like this test needs a backend supporting the kernel stacking functionality. Currently only the CPU backend supports this. IMHO, other backends should  unconditionally exclude it.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -308,11 +308,12 @@ static void getMaxLocalWorkgroupSize(cl_kernel kernel, cl_device_id device,
...
   for (int i = 0, e = global.size(); i < e; i++) {
-    local[i] = L;
+    local[i] = std::min(L, WIS[i]);
+    local[i] = std::min(local[i], WGS / totalWorkPrevDims);
 
-    while (global[i] % local[i] || L % local[i] || local[i] > WIS[i] ||
-           local[i] * totalWorkPrevDims > WGS) {
+    while (global[i] % local[i] || L % local[i]) {
       local[i]--;
+      assert(local[i] < L);
```
opti-mix: Could you use `DCHECK` with an error message instead.

tests/unittests/CMakeLists.txt
beicy: There are heterogeneous Partition test. So we use "CPU" and "Interpreter" for different type of backends.

opti-mix: @beicy Could it be two different Mock backends, both Interpreter based, but with different sets of supported operations? Or do they need to be based on different backends?

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: I think that assert was useless due to the min above. I removed it.

lib/Backends/Interpreter/tests/InterpreterBackendCorrectnessTest.cpp
pjaaskel: I made this blacklisting unconditional.

pjaaskel:
Good to go now?

opti-mix:
@pjaaskel Thanks for fixing this issue and addressing the review comments!

PR #3527, opened 2019-09-18
[Grad] Add support for differentiating TileNode

SplitInfinity:
**Summary**
This commit adds support for differentiating `TileNodes`. The gradient
with respect to the input of the `TileNode` is the sum of the slices of
the gradient with respect to the output of the `TileNode` that correspond
to the slices that were repeated and concatenated during the forward
pass of `TileNode`. This operation is performed by using a `ReshapeNode` to
modify the dimensions of the gradient with respect to the `TileNode` output
such that the tiling axis is replaced with `{numTiles, tileDim}` and then
adding all of the slices with a `BatchedReduceAdd` node on the axis
corresponding to `numTiles`.

**Testing**
This commit adds unit tests to `GraphGradTest` and `GradCheckTest` for
`TileNode`. The latter checks that the gradient is correct when tiling on
a nontrivial axis (i.e. not the first). All unit tests pass.

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -144,6 +144,38 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
       continue;
     }
 
+    if (N->getKind() == Kind::TileNodeKind) {
+      TileNode *TN = cast<TileNode>(N);
+      NodeValue outputG = map.getGradient(TN->getResult());
+
+      // To compute the gradient with respect to the input of the TileNode, all
+      // of the slices in outputG corresponding to the tiled slices in the
+      // forward pass need to be added together. This is achieved by reshaping
+      // outputG to replace the tiling axis with {numTiles, tileDim}, and then
+      // performing a BatchedReduceAdd on the axis with numTiles elements. For
+      // example, if the tile creates a {n,x,h,w} output with a {n,c,h,w}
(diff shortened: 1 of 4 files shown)


SplitInfinity:
I disabled `checkTileGradient` on OpenCL for now because that backend can only support `BatchedReduceAdd` with at most 4 dimensions.

lib/Graph/Grad.cpp
```diff
@@ -144,6 +144,33 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
...
     }
 
+    if (N->getKind() == Kind::TileNodeKind) {
+      TileNode *TN = cast<TileNode>(N);
+      NodeValue outputG = map.getGradient(TN->getResult());
+
+      // To compute the gradient with respect to the input of the TileNode, all
+      // of the slices in outputG corresponding to the tiled slices in the
+      // foward pass need to be added together. This is achieved by reshaping
+      // outputG to replace the tiling axis with {numTiles, tileDim}, and then
```
jackm321: I feel like this is a confusing description but I don't know how to say it better. Maybe an example like if the tile created {n,x,h,w} from {n,c,h,w} where x = c * num_tiles then we reshape {n,x,h,w} to {n, num_tiles, c, h, w} so that BatchedReduceAddNode eliminates the num_tiles getting back to {n,c,h,w}

PR #3528, opened 2019-09-18
[PyTorch] Update test all close tol to 1e-2

zrphercule:
As we implementing quantization, the result mismatch becomes common and tolerable. 
Therefore we would like to update our tol to 1e-2, to make sure all our tests not flaky.


yinghai:
Where is the `1e-2` change?

PR #3535, opened 2019-09-19
[GraphOptz] Add helper function to check numerical correctness of optimizations

### LINKED ISSUE #3523: [GraphOptzTest] Add a numerical correctness check helper to GraphOptzTest
These tests check that the graph is mutated as expected, but do not check that the optimized graph produces the same numerical output as the unoptimized graph. The goal in this issue is to add a helper function (e.g. `checkEqualOutputs(F1, F2, context)`) that does this check and then to add to all of the tests in this file.

SplitInfinity:
**Summary**
This commit adds a helper function to `GraphOptzTest` that helps check
whether a `Function` and its optimized counterpart produce the same
results when given the same inputs. At the moment, the tests in
`GraphOptzTest` make sure that a `Function` is mutated as expected, but
don't check that the optimization is correct.

**Testing**
This commit modifies one of the test cases in `GraphOptzTest`,
`cancelTwoTransposes` to use this new feature. All tests pass.

**Fixes**
This PR fixes #3523.

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -28,11 +28,38 @@ using namespace glow;
 
 class GraphOptz : public ::testing::Test {
 public:
-  GraphOptz() { F_ = mod_.createFunction("main"); }
+  GraphOptz() : mod_(EE_.getModule()) { F_ = mod_.createFunction("main"); }
 
 protected:
-  Module mod_;
-  Function *F_;
+  void checkNumericalEquivalence() {
+    // Check that the function and its optimized complement exist.
+    EXPECT_TRUE(F_);
+    EXPECT_TRUE(optimizedF_);
+
+    // Clone bindings to use for original and optimized functions.
+    PlaceholderBindings originalBindings = bindings_.clone();
+    PlaceholderBindings optimizedBindings = bindings_.clone();
+
+    // Compile and run functions.
+    EE_.compile(CompilationMode::Infer);
+    EE_.run(originalBindings, F_->getName());
+    EE_.run(optimizedBindings, optimizedF_->getName());
+
+    // Compare outputs.
+    EXPECT_TRUE(
+        PlaceholderBindings::compare(&originalBindings, &optimizedBindings));
+  }
+
+  /// ExecutionEngine instance for running functions to check numerical
+  /// equivalence.
+  ExecutionEngine EE_;
+  /// A reference to the Module inside EE_.
+  Module &mod_;
+  /// The original Function for the test case.
+  Function *F_{nullptr};
+  /// The optimized Function for the test case.
+  Function *optimizedF_{nullptr};
+  /// The bindings used to check numerical equivalence for the test case.
   PlaceholderBindings bindings_;
 };
 
@@ -48,6 +75,13 @@ static unsigned countNodeKind(Function *F, Kinded::Kind kind) {
   }
 
   return count;
+}
+
+/// Optimize the function \p F. \returns the optimized function.
+static Function *optimizeFunction(Function *F) {
+  auto *G = F->clone(F->getName().str() + "_optimized");
+  ::glow::optimize(G, CompilationMode::Infer);
+  return G;
 }
 
 TEST_F(GraphOptz, DCE) {
@@ -664,7 +698,8 @@ TEST_F(GraphOptz, sinkTransposeBelowRescale) {
 
 TEST_F(GraphOptz, cancelTwoTransposes) {
   const size_t origDims[] = {1, 5, 10, 15};
-  Node *A = mod_.createPlaceholder(ElemKind::FloatTy, origDims, "input", false);
+  Placeholder *A =
+      mod_.createPlaceholder(ElemKind::FloatTy, origDims, "input", false);
   Node *T1 = F_->createTranspose("transpose", A, NCHW2NHWC);
   Node *T2 = F_->createTranspose("transpose", T1, NHWC2NCHW);
   ReluNode *K = F_->createRELU("relu", T2);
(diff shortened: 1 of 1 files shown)


SplitInfinity:
This diff seems large but most of the changes are for changing `.` to `->` because members that used to be objects/references are pointers now. Reviewers should focus their attention on two things:
1) the newly added `checkNumericalEquivalence`
2) `cancelTwoTransposes`, the one test case that I have modified to use this new feature

I want to get feedback on how `checkNumericalEquivalence` should work before I spend the time to change all of the test cases.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -28,11 +28,34 @@ using namespace glow;
...
+    EE_.run(originalBindings, originalF_->getName());
+    EE_.run(optimizedBindings, optimizedF_->getName());
+
+    // Compare outputs.
+    EXPECT_TRUE(
+        PlaceholderBindings::compare(&originalBindings, &optimizedBindings));
+  }
+
+  ExecutionEngine EE_;
+  Module *mod_{nullptr};
```
opti-mix: Why do you need `mod_` to be a pointer rather than a reference?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -28,11 +28,34 @@ using namespace glow;
...
+  Function *originalF_{nullptr}, *optimizedF_{nullptr};
```
opti-mix: I'd keep the original function as just `F_`. This is more in line with other tests and also the amount of changes in your PR.

Also, please add Doxygen comments for `originalF_`(or `F_`) and `optimizedF_`.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -659,54 +689,69 @@ TEST_F(GraphOptz, sinkTransposeBelowRescale) {
...
+  Node *T1 = originalF_->createTranspose("transpose", A, NCHW2NHWC);
+  Node *T2 = originalF_->createTranspose("transpose", T1, NHWC2NCHW);
+  ReluNode *K = originalF_->createRELU("relu", T2);
+  SaveNode *save = originalF_->createSave("ret", K);
 
   EXPECT_EQ(K->getInput().dims(), llvm::makeArrayRef(origDims));
-  EXPECT_EQ(F_->getNodes().size(), 4);
+  EXPECT_EQ(originalF_->getNodes().size(), 4);
+
+  optimizedF_ = originalF_->clone(originalF_->getName().str() + "_optmized");
```
opti-mix: These two lines are going to be the same for all tests. They can be moved into a small helper function.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -659,54 +689,69 @@ TEST_F(GraphOptz, sinkTransposeBelowRescale) {
...
+    }
+  }
 
-  EXPECT_EQ(F_->getNodes().size(), 2);
   ReluNode *relu = llvm::dyn_cast<ReluNode>(save->getInput());
   ASSERT_TRUE(relu);
   EXPECT_EQ(relu->getResult().dims(), llvm::makeArrayRef(origDims));
   EXPECT_EQ(relu->getInput().getNode(), A);
+
+  bindings_.allocate(mod_->getPlaceholders());
```
opti-mix: Are these two lines going to be test-specific? If you mean here something generic like: "Allocate all placeholders and initialize them with random values" and then check for the numerical equivalence, then this logic could be moved into a test-independent helper function. 

BTW, using `randomize` makes this test non-deterministic, or? I.e. if we run it multiple times, it may eventually fail or not fail depending on the values of the placeholders. So, should we make it more deterministic by seeding the RNG in a special way? Also, may be we need to run the tests not only on a single combination of random placeholder values, but on a set of such combinations?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -28,11 +28,34 @@ using namespace glow;
...
-  GraphOptz() { F_ = mod_.createFunction("main"); }
+  GraphOptz() {
+    mod_ = &EE_.getModule();
+    originalF_ = mod_->createFunction("main");
+  }
 
 protected:
-  Module mod_;
-  Function *F_;
+  void checkNumericalEquivalence() {
```
opti-mix: Question: 
What if the unoptimized function was producing some kind of result that is e.g. the same for all inputs (e.g. all zeroes), because it was never run for real before and nobody noticed. May be there is a node there that makes all outputs zeros... In this case, the optimized function would eventually produce also all zeros, but not necessarily because the functions are equivalent, but rather because this node that zeros everything is still there. Are we OK with this? Can we improve our checks/tests somehow to catch such situations?

tests/unittests/GraphOptzTest.cpp
SplitInfinity: Now that there is an `ExecutionEngine` in this class, If we make `mod_`  a reference, it must be initialized in the initializer list of the constructor of this class, i.e.:
```
GraphOptz() : mod_(EE_.getModule()) { ... }
```
Is this defined behaviour? Will `EE_` be initialized before `mod_` even though it doesn't appear in the initializer list?

opti-mix: According to the standard:
```
The order of member initializers in the list is irrelevant: the actual order of initialization is as follows:

1) If the constructor is for the most-derived class, virtual base classes are initialized in the order in which they appear in depth-first left-to-right traversal of the base class declarations (left-to-right refers to the appearance in base-specifier lists)
2) Then, direct base classes are initialized in left-to-right order as they appear in this class's base-specifier list
3) Then, non-static data members are initialized in order of declaration in the class definition.
4) Finally, the body of the constructor is executed
```

So, I think you should be fine as long as `EE_` is declared before `mod_`.

tests/unittests/GraphOptzTest.cpp
SplitInfinity: > Are these two lines going to be test-specific? If you mean here something generic like: "Allocate all placeholders and initialize them with random values" and then check for the numerical equivalence, then this logic could be moved into a test-independent helper function.

I think some tests (such as this one) do not care about the specific values and can use random values. But if the optimisation is something like replacing `MaxPool -> Relu` with `Relu -> MaxPool`, I think the test writer may want to make sure that it is possible to specify that some of the input values should be negative to test the clamping behaviour of the relu operation.

> BTW, using randomize makes this test non-deterministic, or? I.e. if we run it multiple times, it may eventually fail or not fail depending on the values of the placeholders. So, should we make it more deterministic by seeding the RNG in a special way?

The PRNG inside the module is already seeded with some default seed, so I don't think we have to worry about non-determinism:
```
                     llvm::cl::init(glow::PseudoRNG::Engine::default_seed));
```

> Also, may be we need to run the tests not only on a single combination of random placeholder values, but on a set of such combinations?

What did you have in mind?

opti-mix: >> Also, may be we need to run the tests not only on a single combination of random placeholder values, but on a set of such combinations?

> What did you have in mind?

Like literally have a loop around `randomize inputs; checkNumericalEquivalence;` so that we check the numerical equivalence for a multiple rounds of random inputs. Would it improve our confidence in the equivalence of results?

SplitInfinity: I think it would, but we should prefer deliberate specification of input whenever possible. Consider the `MaxPool -> Relu` ==> `Relu -> MaxPool` mentioned earlier. If the input values were randomly generated and all came out positive (either due to bad luck or careless specification of the upper and lower bound of the RNG) and the optimization code was incorrect and omitted the Relu, a numerical equivalence check would pass.

tests/unittests/GraphOptzTest.cpp
nickgg: It feels hard to do this on a systematic basis since it really depends on the details of the Nodes.

SplitInfinity:
Addressed comments

opti-mix:
The diff looks much smaller now ;-)

opti-mix:
@SplitInfinity Do you plan to update the tests yourself or do you want to file it as a "good first issue"?

SplitInfinity:
@opti-mix I plan to file them as "good first issue" and bootcamp them.

PR #3536, opened 2019-09-20
Modified RecommendationSystemTest to use HostManager directly. 

### LINKED ISSUE #3521: RecommendationSystemTest should use HostManager instead of ExecutionEngine
RecommendationSystemTest currently uses ExecutionEngine, which is OK. But due to this fact the functionality of this test are a bit limited for the following scenarios:

- It is impossible to test with concurrent requests, because ExecutionEngine does not support concurrent execution

- It is impossible to test with heterogeneous partitioning, because ExecutionEngine supports only a single backend at a time. But heterogeneous partitioning use-cases are very important for bringing up new backends, which support only a subset of operations required by the full RecommendationSystemTest model.

Therefore, it would be better to use HostManager instead of ExecutionEngine to overcome the limitations mentioned above.

gcatron:
Summary:
Modified RecommendationSystemTest to use HostManager directly. 
This allows for testing concurrency and heterogeneous partitioning.
Added a second request to test concurrent requests.
Documentation:

Fixes #3521 

Test Plan: Verify RecommendationSystemTest still builds and runs successfully.

### DIFF
--- a/tests/unittests/RecommendationSystemTest.cpp
+++ b/tests/unittests/RecommendationSystemTest.cpp
@@ -22,11 +22,14 @@
 
 #include <algorithm>
 #include <cmath>
+#include <future>
 #include <random>
 
 #include "gtest/gtest.h"
 
 #include "llvm/Support/CommandLine.h"
+
+constexpr size_t MAX_MEMORY = 64e+9;
 
 using namespace glow;
 
@@ -36,6 +39,10 @@ llvm::cl::OptionCategory recSysTestCat("RecSys Category");
 llvm::cl::opt<unsigned> miniBatchOpt("mini-batch", llvm::cl::desc("Minibatch."),
                                      llvm::cl::Optional, llvm::cl::init(8),
                                      llvm::cl::cat(recSysTestCat));
+
+llvm::cl::opt<unsigned> concurrentReqestsOpt(
+    "concurrent-count", llvm::cl::desc("Number of concurrent requests."),
+    llvm::cl::Optional, llvm::cl::init(1), llvm::cl::cat(recSysTestCat));
 
 llvm::cl::opt<unsigned> embeddingDimOpt("embedding-dim",
                                         llvm::cl::desc("Embedding dim."),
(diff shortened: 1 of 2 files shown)


tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -892,10 +908,21 @@ class RecommendationSystemTest : public BackendTest {
     // Use the same precision transformation for compilation.
     CompilationContext cctx;
     cctx.precisionConfig = precConfig_;
-    EE_.compile(cctx);
+    std::vector<std::unique_ptr<DeviceConfig>> configs;
```
opti-mix: The new code here seems to have a lot of boiler-plate similar to some other parts you introduced. Could you try to move this common logic into a helper function?

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -733,13 +749,29 @@ class RecommendationSystemTest : public BackendTest {
...
+    hostManager->addNetwork(std::move(mod), cctx);
 
     // Run graph
-    EE_.run(context_);
+    // Create a second context and run it too to test concurrent requests.
+    ExecutionContext context2{};
+    std::promise<void> runNetwork, runNetwork2;
+    auto ready = runNetwork.get_future();
+    auto ready2 = runNetwork2.get_future();
+    dispatchInference(hostManager.get(), runNetwork, context_);
```
opti-mix: Does it mean we now unconditionally always run two concurrent requests? May be it should be done conditionally?

gcatron: Refactored it into a command-line arg concurrent-requests

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -769,60 +821,37 @@ class RecommendationSystemTest : public BackendTest {
...
     PlaceholderBindings *bindingsI = contextI.getPlaceholderBindings();
     Tensor *resultIT = createSimpleRecSysGraph(*modI, *bindingsI, IF,
                                                tableSizes, embeddingDim);
     bindingsI->allocate(modI->getPlaceholders());
 
+    std::vector<std::unique_ptr<DeviceConfig>> configs;
+    auto deviceConfig = llvm::make_unique<DeviceConfig>("Interpreter");
+    // Set device memory to 64GB to prevent partitioning. We are using the
+    // Interpreter's result just as a reference result to compare against.
+    deviceConfig->setDeviceMemory(64e+9);
```
opti-mix: Let's make `64e+9` a named constexpr.

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -769,60 +821,37 @@ class RecommendationSystemTest : public BackendTest {
...
-
   /// Run on the Interpreter and compare the result to previous result.
   void compareAgainstInterpreter() {
     ExecutionContext contextI;
-    // Set device memory to 64GB to prevent partitioning. We are using the
-    // Interpreter's result just as a reference result to compare against.
-    ExecutionEngine IEE("Interpreter", 64e+9);
-    auto *modI = &IEE.getModule();
+
+    std::unique_ptr<Module> modI(new Module);
```
opti-mix: Could you add some comments inside this function. Please explain some of the variables like  `modI` (interpreted module?), `pMod` (partitioned module?). 

Also please use a more uniform naming, e.g. use only (descriptive) prefixes or suffixes for indicating the meaning, but not both styles at the same time ;-)

gcatron:
> Please see below my initial comments.
> 
> I also have a small request:
> Could we use something similar to the following code from Loader.cpp?
> 
> ```c++
>  std::vector<std::unique_ptr<runtime::DeviceConfig>> configs =
>       generateDeviceConfigs(loadDeviceConfigsFileOpt, numDevices,
>                             ExecutionBackend);
> ```
> 
> The idea is to introduce an option to configure devices for a heterogeneous partitioning setup. If this option is not provided, a default setup should be used (which is hard-coded at currently).

I think I'd like to bump that to a follow up PR, Ideally I'd like to move the logic in Loader to a common file and then use it from both places. I'm also not sure what the ideal solution is here to support the partition test and normal test from a single config file. I can file a follow up issue.

opti-mix:
@gcatron 

> I think I'd like to bump that to a follow up PR, Ideally I'd like to move the logic in Loader to a common file and then use it from both places. I'm also not sure what the ideal solution is here to support the partition test and normal test from a single config file. I can file a follow up issue.

Sounds good! Please file such  a follow-up issue. 

And please address the remaining comments (comments, naming) so that we can merge this PR.

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -715,16 +721,69 @@ class RecommendationSystemTest : public BackendTest {
     }
   }
 
+  void dispatchInference(HostManager *hostManager, ExecutionContext &context) {
```
opti-mix: Please add a Doxygen comment.

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -715,16 +721,69 @@ class RecommendationSystemTest : public BackendTest {
...
+    }
+
+    for (auto &future : futures) {
+      future.wait();
+    }
+    // Release the original context passed in by reference so we don't free it.
+    contexts[0].release();
+  }
+
+  std::vector<std::unique_ptr<runtime::DeviceConfig>>
```
opti-mix: Please add a Doxygen comment. BTW, should this function (and may be a couple of other functions) be static or reside in an anonymous namespace?

tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -834,68 +871,72 @@ class RecommendationSystemTest : public BackendTest {
...
     // Use the same precision transformation for compilation.
     CompilationContext cctx;
     cctx.precisionConfig = precConfig_;
-    EE_.compile(cctx);
+    auto configs = generateDeviceConfigs(getBackendName(), 1, MAX_MEMORY);
+    std::unique_ptr<HostManager> hostManager(
+        new HostManager(std::move(configs)));
+    hostManager->addNetwork(std::move(mod), cctx);
 
     // Run graph
```
opti-mix: Period is missing at the end of the sentence.

PR #3537, opened 2019-09-20
[PyTorch] Fuse conv2d prepack into conv2d

zrphercule:
In this diff we would like to fuse conv2d prepack into conv2d it self.
Later we will only need to deal with two kinds of conv2d:
already packed conv2d, then first unpack them;
or fused conv2d, then deal with it directly.


torch_glow/src/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_library(PyTorchModelLoader
                      GlowIValue.cpp
                      PyTorchCommon.cpp
                      FuseLinear.cpp
+					 FusePrepack.cpp
```
zrphercule: lint...

jackm321:
@yinghai yeah it's a good question. I think if we can go ahead and skip some work like packing then unpacking weights by fusing we should do that. Also things that are expected to have just one user (also true in this case) are good candidates for fusing. For other things, it's fusing probably shouldn't be the go-to method

yinghai:
Yeah I think we have some basic mechanism such as glow ivalue in place, we can try to consume those small ops and do const propagation. One thing we eventually want to avoid is to write a lot of custom fusion passes.

zrphercule:
> Yeah I think we have some basic mechanism such as glow ivalue in place, we can try to consume those small ops and do const propagation. One thing we eventually want to avoid is to write a lot of custom fusion passes.

Agree. The only reason why we need this fuser is I didnt see conv_prepack related ops in glow, which means we need to rather fuse this op, or leave it outside of the fused node. Both solutions are valid, but anyway this provided a path.

PR #3541, opened 2019-09-21
[OpenCL] avoid vector datatypes if the device prefers so

pjaaskel:
Summary:

For some devices using intra vectors can be counter productive. This patch checks for the preference using the OpenCL device API and avoids using the intra vec ones in that case.

Test Plan:

Tested with multiple POCL OpenCL devices. clang-format'd (9.0).

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -239,6 +239,34 @@ static size_t setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
   return kernelArgIdx - 1;
 }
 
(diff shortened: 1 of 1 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -239,6 +239,31 @@ static size_t setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
   return kernelArgIdx - 1;
 }
 
+unsigned getPreferredVectorWidth(cl_device_id device, ElemKind elemTy) {
```
opti-mix: Could you make is static and add a Doxygen comment?

pjaaskel: Done.

PR #3542, opened 2019-09-21
Allow changing the tensor dimension type (to 32b)

pjaaskel:
Summary:

The previously used size_t is host-dependent and a 64b dimension width is clearly overkill for most of the networks. 32b data type adds more data level paralleism for computations with tensor dimension data as well as simplifies address computations.

The patch adds a new typedef glow::dim_t with a signed counterpart glow::sdim_t which are defined to a fixed width 32b type.

As a side effect, it simplified the OpenCL kernel interfacing since there is no need to pass host size_t width to it since the dimension type is now fixed width.

The 32b dimension type can be enabled via cmake -DTENSOR_DIMS_32_BITS=ON

Test Plan:

Tested with LLVM 7 (CPU backend) and OpenCL (pocl-phsa x86_64). Both with 32b and 64b index types. Passes the basic **ninja check** for both.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ option(GLOW_BUILD_PYTORCH_INTEGRATION "Build integration for PyTorch" OFF)
 option(GLOW_BUILD_TESTS "Build the tests" ON)
 option(GLOW_WITH_BUNDLES "Build bundles" OFF)
 option(LINK_PROTOBUF_AS_DLL "Link against protobuf build as dynamic libray." OFF)
+option(TENSOR_DIMS_32_BITS "Set the max bitwidth of the tensor dimension and related indices to 32b instead of 64b." OFF)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CXX_STANDARD_REQUIRED ON)
@@ -137,6 +138,14 @@ if (GLOW_WITH_HABANA)
     "${SYNAPSE_LIB}"
     "${TPCSIM_SHARED_LIB}"
     "${HL_THUNK_LIB}")
+endif ()
+
+if (TENSOR_DIMS_32_BITS)
+  add_definitions(-DDIM_T_32)
+  set(LLVMCPURuntimeExtraFlags "-DDIM_T_32")
+  message(STATUS "Using 32b tensor dimensions.")
+else()
+  message(STATUS "Using 64b tensor dimensions.")
 endif ()
 
 # Top level setup for external backends
--- a/examples/char-rnn.cpp
+++ b/examples/char-rnn.cpp
@@ -96,9 +96,9 @@ static void loadText(Tensor &inputText, Tensor &nextChar, llvm::StringRef text,
   //  |ello| |World
   //  |llo |W|orld
   //  |lo W|o|rld
-  for (size_t i = 0; i < B; i++) {
-    for (size_t j = 0; j < S; j++) {
-      size_t c = clipASCII(text[i + j]);
+  for (dim_t i = 0; i < B; i++) {
+    for (dim_t j = 0; j < S; j++) {
+      dim_t c = clipASCII(text[i + j]);
 
       IH.at({i, j, c}) = 1.0;
       if (train) {
@@ -125,14 +125,14 @@ PseudoRNG &getRNG() {
 /// to one. The algorithm that we use here picks a random number between zero
 /// and one. Then, we scan the tensor and accumulate the probabilities. We stop
 /// and pick the index when sum is greater than the selected random number.
-static char getPredictedChar(Tensor &inputText, size_t slice, size_t word) {
+static char getPredictedChar(Tensor &inputText, dim_t slice, dim_t word) {
   auto IH = inputText.getHandle();
 
   // Pick a random number between zero and one.
   double x = std::abs(getRNG().nextRand());
   double sum = 0;
   // Accumulate the probabilities into 'sum'.
-  for (size_t i = 0; i < 128; i++) {
+  for (dim_t i = 0; i < 128; i++) {
     sum += IH.at({slice, word, i});
     // As soon as we cross the threshold return the index.
     if (sum > x) {
@@ -159,8 +159,8 @@ static std::unique_ptr<llvm::MemoryBuffer> loadFile(llvm::StringRef filename) {
 /// Creates a new RNN network. The network answers the question, given N chars
 /// of input, what is the character following each one of these chars.
 static Function *createNetwork(Module &mod, PlaceholderBindings &bindings,
-                               size_t minibatchSize, size_t numSteps,
-                               size_t hiddenSize) {
+                               dim_t minibatchSize, dim_t numSteps,
+                               dim_t hiddenSize) {
   Function *F = mod.createFunction("main");
 
   auto *X = mod.createPlaceholder(
@@ -212,10 +212,10 @@ int main(int argc, char **argv) {
   LOG(INFO) << "Loaded " << text.size() << " chars.\n";
   PlaceholderBindings inferBindings, trainingBindings;
 
-  const size_t numSteps = 50;
-  const size_t minibatchSize = 32;
-  const size_t batchSize = text.size() - numSteps;
-  const size_t hiddenSize = 256;
+  const dim_t numSteps = 50;
+  const dim_t minibatchSize = 32;
+  const dim_t batchSize = text.size() - numSteps;
+  const dim_t hiddenSize = 256;
 
   CHECK_GT(text.size(), numSteps) << "Text is too short";
   TrainingConfig TC;
--- a/examples/fr2en.cpp
+++ b/examples/fr2en.cpp
@@ -175,7 +175,7 @@ private:
   Placeholder *embedding_fr_, *embedding_en_;
   Node *encoderHiddenOutput_;
 
-  Placeholder *loadEmbedding(llvm::StringRef langPrefix, size_t langSize) {
+  Placeholder *loadEmbedding(llvm::StringRef langPrefix, dim_t langSize) {
     auto &mod = EE_.getModule();
     auto *result =
         mod.createPlaceholder(ElemKind::FloatTy, {langSize, EMBEDDING_SIZE},
@@ -298,7 +298,7 @@ void Model::loadDecoder() {
   auto *input = mod.createPlaceholder(ElemKind::Int64ITy, {batchSize_},
                                       "decoder.input", false);
   auto *inputTensor = bindings.allocate(input);
-  for (size_t i = 0; i < batchSize_; i++) {
+  for (dim_t i = 0; i < batchSize_; i++) {
     inputTensor->getHandle<int64_t>().at({i}) = en_.word2index_["SOS"];
   }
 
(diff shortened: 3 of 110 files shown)


opti-mix:
@pjaaskel Thanks for this PR!

- Could you elaborate a bit if this PR is trying to address a concrete performance issue you've discovered or is it a hypothetical performance improvement?

- It does not change much for the OpenCL backend, or? This backend was using a device-specific size_t in all the kernels anyways.

pjaaskel:
> @pjaaskel Thanks for this PR!
>     * Could you elaborate a bit if this PR is trying to address a concrete performance issue you've discovered or is it a hypothetical performance improvement?

Yes, this had major (> 10%) performance benefit on a (private) target where 32b integer (SIMD) computation is faster than 64b. I didn't test other targets, but since I believe this is generally true usually (32b integer computations are either as fast or faster than 64b), it seemed a good idea to allow reducing address computation overheads this way.

I think this should improve the CPU (LLVM) backend perf. too especially in case there is vectorizable address computation based on indices that now allows to utilize 32b SIMD subwords. But I haven't measured this.

>     * It does not change much for the OpenCL backend, or? This backend was using a device-specific size_t in all the kernels anyways.

Unfortunately OpenCL has a fallacy here: size_t is tied to the address space width (unlike in C/C++), so if one wants to use 64b pointers/address space width (to global space), the device should declare size_t to 64b although wishing to strength reduce needless address computations to lower bit widths (and declare max _object size_ to be 32b), even though 32b would be much faster/higher throughput.

The current way of explicitly passing the dimension/index data type improves the situation as it communicates we need only 32b for the indices and we ensure the index width is in synch with the data from the host.  The previous didn't use size_t for indices, but uint64_t. Also the node validator assumed a 64b index type.

opti-mix:
@pjaaskel Thanks for the detailed explanations!

And thanks a lot for the PR! It is a significant amount of work to go through the whole source code and identify places where size_t/ssize_t could be semantically replaced by a custom type.

Overall, I like the idea! We've discussed it internally a couple of times in the past.

Let's collect some opinions and reactions from others. In particular, does it break any private backends eventually?

bertmaher:
I like the direction, for sure.  I'm in particular not a huge fan of `size_t`, since it's dependent on the host that's running Glow, which may not have any relation to the device running Glow's generated code.

Something I'm curious about: in the cases where you've observed a speedup from going 64->32 bits, was it actually because the tensor's dimensions were narrower, or because the data stored in the tensors became narrower (e.g., you can make `indices` tensors of embedding lookups 32 bits).  I've definitely seen the latter -- using `uint32_t` for indices of `SparseLengthsSum` is a substantial win since it halves the memory bandwidth spent on transferring the indices themselves.  But you can do that without forcing any tensor dimensions to physically be. 32 bit; you just have to create embedding lookups with the right dtype on their inputs.  If it's the former case (significant computations involving the actual tensor dimensions) I'd be curious to hear what it is.  (PM me if it's non-public, maybe we can work out NDAs to discuss in more detail, since I'd love to help move this forward).

I'm a bit concerned if we lose the ability to represent >32 bit dimensions.  There are at least semi-plausible domains where you could have a dimension > 2^32 (an embedding vector for every person, or something like that).  If I'm not mistaken, TensorFlow's XLA [uses 64 bits for shapes as well](https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/compiler/xla/shape.h#L216), which isn't totally dispositive, but might suggest that it's a useful property.

pjaaskel:
> Something I'm curious about: in the cases where you've observed a speedup from going 64->32 bits, was it actually because the tensor's dimensions were narrower, or because the data stored in the tensors became narrower (e.g., you can make `indices` tensors of embedding lookups 32 bits). I've definitely seen the latter -- using `uint32_t` for indices of `SparseLengthsSum` is a substantial win since it halves the memory bandwidth spent on transferring the indices themselves. But you can do that without forcing any tensor dimensions to physically be. 32 bit; you just have to create embedding lookups with the right dtype on their inputs. If it's the former case (significant computations involving the actual tensor dimensions)

I'm not sure what you mean by "embedding lookups 32bits" here.

> I'd be curious to hear what it is. (PM me if it's non-public, maybe we can work out NDAs to discuss in more detail, since I'd love to help move this forward).

In that private target all vectorized 64b integer computations and memory accesses are expensive since they fall back to scalarized execution instead of wide SIMD. So tensor index computations based on 64b shapes became a major bottleneck early on in our optimization assignment.
 
> I'm a bit concerned if we lose the ability to represent >32 bit dimensions. There are at least semi-plausible domains where you could have a dimension > 2^32 (an embedding vector for every person, or something like that). If I'm not mistaken, TensorFlow's XLA [uses 64 bits for shapes as well](https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/compiler/xla/shape.h#L216), which isn't totally dispositive, but might suggest that it's a useful property.

Yes, indeed this is my main concern of enabling it by default; are there feasible networks which need such huge dimensions? Can be, like you state, which at least hints that the index should be easily changeable (at build time?) back to 64b. I've tried to keep it such by means of the centralized dim_t/dims_t/IndexElementType typedefs, but I haven't re-tested setting dim_t to uint64_t after my modifications.

bertmaher:
> I'm not sure what you mean by "embedding lookups 32bits" here.

Usually the "indices" tensor of SparseLengthsSum.  But it sounds like it's the actual shapes that are slowing you down, not the data.

> > I'd be curious to hear what it is. (PM me if it's non-public, maybe we can work out NDAs to discuss in more detail, since I'd love to help move this forward).
> 
> In that private target all vectorized 64b integer computations and memory accesses are expensive since they fall back to scalarized execution instead of wide SIMD. So tensor index computations based on 64b shapes became a major bottleneck early on in our optimization assignment.

Very interesting.  I'll be honest I did not expect vectorizing indices to be a significant performance bottleneck.
 
> Yes, indeed this is my main concern of enabling it by default; are there feasible networks which need such huge dimensions? Can be, like you state, which at least hints that the index should be easily changeable (at build time?) back to 64b. I've tried to keep it such by means of the centralized dim_t/dims_t/IndexElementType typedefs, but I haven't re-tested setting dim_t to uint64_t after my modifications.

Cool, yeah, I think I like the `dim_t` refactoring regardless.  But I think I want to default to 64b and switch to 32b based on a build flag.

pjaaskel:
I made it a build-time option, clang format'd, fixed a bunch of warnings due to data type bitwidth downcasts etc. A huge diff, but pretty trivial changes, I think. What do you think now?

pjaaskel:
I rebased to the latest master and fixed a couple of new issues that came from new codes that used size_t.

pjaaskel:
Bling?

jfix71:
@pjaaskel sorry for the delay, @bertmaher has been out for the past few weeks and so this fell through the cracks. We will try to get this landed early next week.

bertmaher:
Hey @pjaaskel, sorry for the long latency, my wife and I had our 2nd child a few weeks ago so I've been changing diapers instead of reviewing diffs :-D.  Anyways, this change lgtm; can you rebase to fix the merge conflicts, and address the test failures, and then we can get this in?  Thanks!

pjaaskel:
Congratulations to your and your family (co-incidentally it's Father's day in Finland today). I've rebased the branch and fixed 'ninja check' to pass for 32b&64b LLVM 7 and OpenCL.

bertmaher:
Looks like the build is broken due to a change to the `foxi` submodule.  I think you need to revert foxi version in this PR to the version in master.  (I can never remember how to do this offhand, so hopefully you know :) ).

pjaaskel:
I rebased again (without conflicts). 

I'm not confortable with using submodules either, but I don't see any submodule revision changes in this PR and it builds for me. When I check the submodule stats in the two branches, they seem to match too:

```
Your branch is up to date with 'glow-origin/master'.
visit0r@x270:~/src/glow-upstreaming$ git submodule 
 eab62a5dd5129f6a4ebfbe4bbe41d35611f7c48d tests/OutputCheck (0.4.2-1-geab62a5)
 090faecb454fbd6e6e17a75ef8146acb037118d4 tests/googlebenchmark (v1.5.0)
 703bd9caab50b139428cea1aaff9974ebee5742e tests/googletest (release-1.8.0-2064-g703bd9ca)
 97fe555430a857581b9b826ecd955e4f0a3653f0 thirdparty/foxi (remotes/origin/HEAD)
 34d4bf01bbf7376f2baa71b8fa148b18524d45cf thirdparty/fp16 (34d4bf0)
 50dc186b50ea512d6888aa1f47414150fd782fa0 thirdparty/onnx (v1.3.0-275-g50dc186b)
 a1b71df137e015d44f7e31f7b6d4807253fb7871 thirdparty/pybind11 (v2.2.0-209-ga1b71df)
visit0r@x270:~/src/glow-upstreaming$ git checkout allow-changing-the-tensor-dimension-type
Switched to branch 'allow-changing-the-tensor-dimension-type'
Your branch is up to date with 'origin/allow-changing-the-tensor-dimension-type'.
visit0r@x270:~/src/glow-upstreaming$ git submodule 
 eab62a5dd5129f6a4ebfbe4bbe41d35611f7c48d tests/OutputCheck (0.4.2-1-geab62a5)
 090faecb454fbd6e6e17a75ef8146acb037118d4 tests/googlebenchmark (v1.5.0)
 703bd9caab50b139428cea1aaff9974ebee5742e tests/googletest (release-1.8.0-2064-g703bd9ca)
 97fe555430a857581b9b826ecd955e4f0a3653f0 thirdparty/foxi (remotes/origin/HEAD)
 34d4bf01bbf7376f2baa71b8fa148b18524d45cf thirdparty/fp16 (34d4bf0)
 50dc186b50ea512d6888aa1f47414150fd782fa0 thirdparty/onnx (v1.3.0-275-g50dc186b)
 a1b71df137e015d44f7e31f7b6d4807253fb7871 thirdparty/pybind11 (v2.2.0-209-ga1b71df)

```

bertmaher:
Weird, I thought I saw a change to thirdparty/foxi before you pushed the latest update.  Oh well.  Now CircleCI seems to be complaining about merging onto master; I don't understand why it's trying to do that though, it never did before.  @rdzhabarov do you know?

pjaaskel:
What's the status of this?

pjaaskel:
I rebased & fixed tests again. Any chance getting this in soon?

bertmaher:
Yeah, I’m not sure what was going wrong with CI, but I’ll try to get it through first thing on Monday

bertmaher:
I fixed the asan issues locally, and I need to make some changes to libjit for build system compatibility.  I’ll try to finish the merge in a few hours once I’m back to the office.

bertmaher:
Done!  Thanks for your patience, and for this great contribution!

pjaaskel:
Thanks! Would it be possible to get a builder bot that at least builds with the 32b dim_t on so it would catch some cases where 64b size_t is used for indices early on?

bertmaher:
It should be possible; in fact, I think you could actually add it yourself if you so desire.  Check out the SHARED build configuration defined in our .circleci setup.  We'll have to monitor whether it's slowing down our CI too much, but if it's just building it shouldn't be too onerous.

PR #3544, opened 2019-09-23
[PyTorch]Fix PYTORCH CI break

zrphercule:
Last week we have some API update in pytorch quantization jit ir, which caused our CI down.
This pr fixed this problem, and all tests should be passed.


zrphercule:
> @zrphercule it does not seem to be passing... Could you check the pytorch job.

skbuild is another issue... Let's first try to see what is wrong with skbuild before landing this.

zrphercule:
@rdzhabarov It seems there is something wrong with installing cmake using pip. Just try
pip install cmake
locally, and you will find the same problem.

PR #3545, opened 2019-09-23
[CI] Not to use pip to install cmake in PYTORCH and CHECK_FORMAT branch.

zrphercule:
Currently CI is down because of we are using pip install cmake which may cause some problem.
In this diff we will try to solve this problem by using apt-get or conda.

Since PYTORCH branch uses an older version of llvm, pip install cmake does not work for that. Therefore switch to use apt-get for that branch.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -72,7 +72,7 @@ sudo apt-get update
 sudo apt-get install cmake
 
 # Install ninja, (newest version of) autopep8 through pip
-sudo pip install ninja cmake autopep8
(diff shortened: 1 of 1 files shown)


zrphercule:
lol apt-get doesnt help, so how can I install cmake?

zrphercule:
PYTORCH branch is also down for another reason, we should ignore this test for now. see https://github.com/pytorch/glow/pull/3544

zrphercule:
@rdzhabarov Could you please give me a stamp? Thanks!

PR #3547, opened 2019-09-24
[GraphOptz] Add optimization to eliminate no-op TileNode

### LINKED ISSUE #3522: [GraphOptimization] Tile nodes that tile something only once should be optimized out
We should add a graph optimization that removes `TileNodes` that tile something only once because these are no-ops.

SplitInfinity:
**Summary**
This commit adds a new graph optimization that eliminates `TileNodes` that
tile their input only once because this operation is a no-op.

**Testing**
This commit adds a unit test to `GraphOptzTest` that checks that a graph
containing a `TileNode` with count=1 is eliminated.

**Fixes**
This PR fixes #3522.


tests/unittests/GraphOptzTest.cpp
```diff
@@ -2210,6 +2210,22 @@ TEST_F(GraphOptz, mergeBANodes) {
...
+                              /*axis=*/1);
+  auto *relu = F_->createRELU("relu", tile);
+  F_->createSave("save", relu);
+
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::TileNodeKind), 1);
+  ::glow::optimize(F_, CompilationMode::Infer);
+
+  // Check that the Tile node is eliminated.
+  EXPECT_EQ(countNodeKind(F_, Kinded::Kind::TileNodeKind), 0);
+}
```
nickgg: I think this is good, but similar to @kumudb's new optimization I think we should check that the functions outputs are identical before and after the change.

SplitInfinity: My change for that just landed, I'll rebase and push.

SplitInfinity:
Rebased and added numerical equivalence check

PR #3548, opened 2019-09-24
[PyTorch] Handle the common matrix multiply op variants and their broadcasted forms

jackm321:
Summary:
* Handle `aten::mm`, `aten::matmul`, and `aten::addmm`
* Remove broken linear fusion pass, not needed anymore.
* Add tests for all of these ops as well as `aten::linear` which is lowered to these ops

One consequence to note from this change is that we aren't able freeze weights for linear now because for matmul, it's not possible to tell from the inputs alone what should be frozen.

Documentation:
doxygen

Test Plan:
Added new unit tests
CI


include/glow/Support/Error.h
```diff
@@ -462,6 +462,7 @@ class GlowExpected final
   template <typename TT>
   friend TT detail::exitOnError(const char *fileName, size_t lineNumber,
```
putivsky: Once you have touched Error.h file, can you fix the Macros problem? Macros must be either outside glow namespace or any reference inside macro must have the fully decorated name "glow::", check marco: RETURN_IF_ERR and FAIL_TEST_IF_ERR

jackm321: @putivsky for the macros it looks like this is the case, can you point me to where it's not and I'll fix it.
GlowExpected is within the glow namespace so this is ok.

putivsky: Image you are calling Macro outside glow namespace. It will fail and you cannot call glow::Marco, the only way to use it it define "using namespace glow" - very inconvenient. Even Marco uses glow functions inside glow namespace it must call it will fully decorated name, i.e. glow::Error.
I have found two macros: RETURN_IF_ERR and FAIL_TEST_IF_ERR

In general it's a good practice to declare macros outside any namespaces.

PR #3550, opened 2019-09-25
[PyTorch] Add tests for quantized MaxPool and AvgPool

jackm321:
Summary:
No code changes since quantized pooling uses the same ops.

Documentation:
comments

Test Plan:
new tests
`python setup.py test`

### DIFF
--- a/torch_glow/tests/nodes/quantized_maxpool_test.py
+++ b/torch_glow/tests/nodes/quantized_maxpool_test.py
@@ -1,0 +1,22 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+
+from tests.utils import jitVsGlow
+
+
+def test_quantized_maxpool():
+    """Basic test of the PyTorch quantized::max_pool2d Node on Glow."""
+
+    def test_f(a):
+        q = torch.nn.quantized.Quantize(
+            scale=1.0/128, zero_point=3, dtype=torch.quint8)
+        dq = torch.nn.quantized.DeQuantize()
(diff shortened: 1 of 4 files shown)


torch_glow/tests/nodes/quantized_maxpool_test.py
```diff
@@ -0,0 +1,21 @@
...
+import torch
+
+from tests.utils import jitVsGlow
+
+
+def test_quantized_maxpool():
+    """Basic test of the PyTorch quantized::max_pool2d Node on Glow."""
+
+    def test_f(a):
+        q = torch.nn.quantized.Quantize(1/128, 3, torch.quint8)
```
rdzhabarov: 1/128 is 0, is that expected?

zrphercule:
lol, even harder than I expected...

zrphercule:
> LGTM. But param to the Quantize looks a bit fishy.

In fact 1/128 or 1/256 is what we supposed to give, since we are quantizating a 0~1 float to 0~255.

rdzhabarov:
NVM my comment on 1/128.

1/128 seems behave differently in python 2 vs python 3.

python 2 => 1/128 == 0
python 3 does fp division.

jackm321:
@rdzhabarov no it's a good catch, it should be explicit about it. I changed it. Thank you for catching it

rdzhabarov:
> scale=1.0/128, zero_point=5, dtype=torch.quint8

Thanks, it's much more readable!

Issue #3551, opened 2019-09-25
Migrate GlowJIT to LLVM's ORCv2 APIs


opti-mix:
LLVM 9.0 introduced new ORC JIT APIs. Old APIs are deprecated. Thus GlowJIT needs to be migrated to the new ORCv2 APIs.

mcurtis-cadence:
@SameerAsal @opti-mix 
I was looking at doing this unless migration is already underway by someone else.

opti-mix:
@mcurtis-cadence 

No, there are no migration efforts underway. So, if you could look into doing this it would be very useful and appreciated!

mcurtis-cadence:
@SameerAsal @opti-mix 
I've started working on this. Feel free to assign this issue to me if you think that makes sense.

Also, I'm a little concerned that I'll break code behind `FACEBOOK_INTERNAL`. It would help to know what LLVM versions are supported with `FACEBOOK_INTERNAL`.

opti-mix:
> Also, I'm a little concerned that I'll break code behind FACEBOOK_INTERNAL. It would help to know what LLVM versions are supported with FACEBOOK_INTERNAL.

Currently, FACEBOOK_INTERNAL means LLVM 9 in most cases.

PR #3552, opened 2019-09-25
Two minor changes for enabling the initial LLVM 9 support

opti-mix:
Summary:

This PR fixes a couple of issues were discovered while trying to build Glow with the recently released LLVM 9.0.

- Declares conversion methods as inline to avoid "duplicate symbol definition" linker errors
- Let backends set their code models
  Switching to LLVM 9 resulted in the code model related failures of some LLVMBackend-based backends when a hard-coded code model is used. These new APIs allow backends set a proper code model.

There are still some issues related to LLVM 9, which are not fixed by this PR. In particular, LLVM's ORC JIT APIs were changed once again. Glow's LLVM JIT related code compiles, but it uses deprecated ORCv1 APIs, which results in compiler warnings. I filed #3551 to track this issue.

Test Plan:

ninja test

### DIFF
--- a/include/glow/Backends/LayoutConverter.h
+++ b/include/glow/Backends/LayoutConverter.h
@@ -23,7 +23,7 @@ namespace glow {
 
 /// Convert regular convolution nodes (that use NHWC) into a backend-specific
 /// convolution nodes using NCHW.
-Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
+inline Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
   // Convert filter and input from NHWC (Glow's default) into NCHW.
   auto *NI = F->createTranspose("conv.input", CN->getInput(), NHWC2NCHW);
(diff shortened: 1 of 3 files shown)


include/glow/Backends/LayoutConverter.h
```diff
@@ -23,7 +23,7 @@ namespace glow {
 
 /// Convert regular convolution nodes (that use NHWC) into a backend-specific
 /// convolution nodes using NCHW.
-Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
+inline Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
```
rdzhabarov: unrelated to the PR? (did it break ODR somewhere?)

opti-mix: Yes, if two different backends include `LayoutConverter.h`, it breaks ODR and we get a linker error. Interestingly enough, this happens only with LLVM 9. When I compiled with LLVM 8, there were no such issues.

PR #3554, opened 2019-09-25
Clip elimination pass

yinghai:
Summary:
Add an optimization pass to eliminate back-to-back Clip ops and just keep one.

Next, we need to blacklist some ops in terms of clipping.

Differential Revision: D17566576


yinghai:
Sounds good to me.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2186,6 +2186,35 @@ static bool isValueChangingCast(TypeRef srcTy, TypeRef destTy) {
   return false;
 }
 
+/// Optimize away redundant ClipNodes.
+/// - First, we  basically turn "Clip(Clip(Clip(A)))" to "Clip(A)".
+/// - Then we remove the Clip at the output of transform/indexing nodes as they
```
jackm321: > Then we remove the Clip at the output of transform/indexing nodes as the

Is the logic that's described here implemented somewhere else? I don't see it in this function. If it's not part of this function we should remove this comment

tests/unittests/GraphOptzTest.cpp
```diff
@@ -84,6 +84,30 @@ static Function *optimizeFunction(Function *F) {
...
+    max -= 1.0;
+    K = F_->createClip("clip", K, min, max);
+  }
+  F_->createSave("ret", K);
+
+  EXPECT_EQ(F_->getNodes().size(), 11);
+
+  ::glow::optimize(F_, CompilationMode::Infer);
+  EXPECT_EQ(F_->getNodes().size(), 2);
+  Node *newClip = A->getUsers().begin()->getUser();
```
jackm321: nit: you could just do the cast llvm::dyn_cast<ClipNode> here then check that it's not nullptr

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3014,6 +3043,9 @@ static void transformForPrecisionMode(const Backend &B, Function *F,
   if (precConfig.convertToFP16) {
     LOG_SCOPE(F->getLogContext(), "glow::convertFunctionToFloat16")
     convertFunctionToFloat16(F, precConfig);
+    FunctionPassManager FPM("FP16GraphOptzFPM",
```
jfix71: Instead of adding this, I think I'd prefer adding another `::glow::optimize(F, cctx, B)` after `transformForPrecisionMode()`/before the second `lower()` [here](https://github.com/pytorch/glow/blob/d2faa645149cbe32126fbdf91e2cbaaf137dae4d/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp#L3063). It's possible we could be missing out on other optimization opportunities for the lower-precision model prior to lowering.

yinghai: Yeah, actually I don't have a very good sense on all the optimization passes here. We can try.

PR #3561, opened 2019-09-27
[PyTorch] Import quantized linear

jackm321:
Summary:
* Enable aten::quantize_linear loader to load as uint8 or int8
* Keep all values between nodes in the correct quantization scheme 
* Import quantized linear operator.
* Create dummy nodes for glow-specific fusion passes to use to eliminate packed intermediate values.

Documentation:
doxygen

Test Plan:
added unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -464,6 +464,17 @@ struct QuantizedAddInputs {
   };
 };
 
+/// Indexes of glow::unpacked_quantized_linear inputs.
+struct QuantizedLinearInputs {
+  enum {
+    input = 0,
+    weight = 1,
+    bias = 2,
+    scale = 3,
+    zero_point = 4,
+  };
+};
+
 /// Indexes of aten::quantize_per_tensor inputs.
 struct QuantizeInputs {
   enum {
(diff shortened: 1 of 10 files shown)


torch_glow/tests/nodes/quantized_linear_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_f(inputs, weights, bias=None):
+        q_int = torch.nn.quantized.Quantize(1.0/64, 2, torch.qint8)
+        q_uint = torch.nn.quantized.Quantize(1.0, 0, torch.quint8)
+
+        dq = torch.nn.quantized.DeQuantize()
+
+        q_inputs = q_uint(inputs)
+        q_weights = q_int(weights)
+
+        return dq(torch.nn.quantized.functional.linear(q_inputs, q_weights, bias))
```
yinghai: If we don't add `dq` here, what is going to happen?

jackm321: It won't work because `elemKindToScalarType()` doesn't know how to translate glow quantized types to pytorch quantized types yet but it should be pretty easy to add

yinghai: Yeah, I think one the things is to look into this "intermediate" blobs. @zrphercule, are yo covering this?

zrphercule:
Yes in order to do intermediate result checking we do need this, and like
Jack said it shouldn’t be hard. Will look into it next week ;)

Yinghai Lu <notifications@github.com>于2019年9月28日 周六下午5:59写道：

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -843,6 +862,34 @@ PyTorchModelLoader::getGlowIValueForValue(const torch::jit::Value *value) {
   return mappingValue.getMappedGlowIValue();
 }
 
+glow::NodeValue PyTorchModelLoader::rescaleUintToInt(glow::NodeValue input) {
```
yinghai: nit: UInt ?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -843,6 +862,34 @@ PyTorchModelLoader::getGlowIValueForValue(const torch::jit::Value *value) {
...
+}
+
+glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
+  auto *inputTy = input.getType();
+  if (inputTy->getElementType() == ElemKind::Int8QTy) {
+    auto dqInput = F_.createDequantize("dequantize", input);
+    auto *outputTy = F_.getParent()->uniqueType(
+        ElemKind::UInt8QTy, inputTy->dims(), inputTy->getScale(),
+        inputTy->getOffset() + OFFSETSHIFT);
+    auto *qOut = F_.createQuantize("quantizeale", dqInput, outputTy);
```
yinghai: Typo?

PR #3562, opened 2019-09-27
Allow for removing a pass from an optimization pipeline

jfix71:
Summary: This may be necessary if a backend wants to opt out of an optimization. This is cleaner than requiring the backend to recreate the default optimization pipeline manually without specific optimizations.

Differential Revision: D17620020

### DIFF
--- a/include/glow/Optimizer/GraphOptimizerPipeline/Pipeline.h
+++ b/include/glow/Optimizer/GraphOptimizerPipeline/Pipeline.h
@@ -98,6 +98,10 @@ class FunctionPassPipeline : private llvm::SmallVector<FunctionPassConfig, 64> {
 private:
   using ParentImpl = llvm::SmallVectorImpl<FunctionPassConfig>;
 
+  /// Removes the first instance of a pass with ID \p FPID. \returns whether an
+  /// instance of the pass was successfully found and removed.
+  bool removeFirstInstanceOfPass(FunctionPassID FPID);
+
 public:
   FunctionPassPipeline() = default;
 
@@ -122,6 +126,9 @@ public:
 
   /// Push a new \p FPC to the end of the pipeline.
   void pushBack(FunctionPassConfig FPC) { push_back(FPC); }
+
+  /// Removes all instances of a pass with ID \p FPID.
+  void removeAllInstancesOfPass(FunctionPassID FPID);
 
   /// Dump a textual representation of the pipeline to \p os.
   void dump(llvm::raw_ostream &os = llvm::outs()) const;
--- a/lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
+++ b/lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
@@ -197,3 +197,18 @@ void FunctionPassPipeline::dump(llvm::raw_ostream &os) const {
     os << "}\n";
   }
 }
+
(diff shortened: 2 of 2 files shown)


lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -197,3 +197,22 @@ void FunctionPassPipeline::dump(llvm::raw_ostream &os) const {
     os << "}\n";
   }
 }
+
+bool FunctionPassPipeline::removeFirstInstanceOfPass(FunctionPassID FPID) {
+  auto it = begin();
+  for (; it != end(); it++) {
+    if (it->getFunctionPassID() == FPID) {
+      break;
```
opti-mix: Why `break` and not just:
```c++
erase(it);
return true;
```

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -197,3 +197,22 @@ void FunctionPassPipeline::dump(llvm::raw_ostream &os) const {
...
+  if (it == end()) {
+    return false;
+  }
+  erase(it);
+  return true;
+}
+
+void FunctionPassPipeline::removeAllInstancesOfPass(FunctionPassID FPID) {
+  while (removeFirstInstanceOfPass(FPID))
+    ;
```
opti-mix: Let's use `{}` instead of `;`, because it looks a bit odd ;-)

include/glow/Optimizer/GraphOptimizerPipeline/Pipeline.h
```diff
@@ -123,6 +123,13 @@ class FunctionPassPipeline : private llvm::SmallVector<FunctionPassConfig, 64> {
   /// Push a new \p FPC to the end of the pipeline.
   void pushBack(FunctionPassConfig FPC) { push_back(FPC); }
 
+  /// Removes the first instance of a pass with ID \p FPID. \returns whether an
+  /// instance of the pass was successfully found and removed.
+  bool removeFirstInstanceOfPass(FunctionPassID FPID);
```
opti-mix: Do you have a use-case where you want to remove only the first instance if a pass, but not other instances of the same pass?

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
jfix71: Updated, but it looks a bit weird... `clang-format` has interesting ideas here 🙂

include/glow/Optimizer/GraphOptimizerPipeline/Pipeline.h
jfix71: Nope -- I've moved this to be private.

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -197,3 +197,19 @@ void FunctionPassPipeline::dump(llvm::raw_ostream &os) const {
     os << "}\n";
   }
 }
+
+bool FunctionPassPipeline::removeFirstInstanceOfPass(FunctionPassID FPID) {
+  auto it = begin();
```
opti-mix: Is there any reason to have `it` to be declared before the for-loop?

PR #3567, opened 2019-09-27
[PyTorch]Add PyTorch Quantized Convolution (unpacked)

zrphercule:
This pr is supposed to add support for pytorch quantized conv with unpakced weight&bias inputs.
With this solving packed w&b should also be easy.

This pr is still WIP and not runable so please dont review it in detail, but feel free to leave comments about the main idea.

### DIFF
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -207,35 +207,6 @@ void registerGlowFusionOpAndPass(std::function<bool()> enablePassFn) {
   registerGlowFusionPass(std::move(enablePassFn));
 }
 
-glow::Type ptTypeToGlowType(const c10::TensorType &ptType) {
-  DCHECK(ptType.scalarType().has_value())
-      << "TensorType has no associated scalar type.";
-  const auto concreteSizes = ptType.sizes().concrete_sizes().value();
-  std::vector<size_t> dims;
-  for (const auto &size : concreteSizes) {
-    dims.push_back(static_cast<size_t>(size));
-  }
-
-  auto scalarType = ptType.scalarType().value();
-  auto glowElemKind = scalarTypeToElemKind(scalarType);
-  return glow::Type(glowElemKind, dims);
-}
-
-glow::Type ptQTypeToGlowQType(const c10::TensorType &ptType, const float &scale,
-                              const int32_t &zero_point) {
-  DCHECK(ptType.scalarType().has_value())
-      << "TensorType has no associated scalar type.";
-  const auto concreteSizes = ptType.sizes().concrete_sizes().value();
-  std::vector<size_t> dims;
-  for (const auto &size : concreteSizes) {
-    dims.push_back(static_cast<size_t>(size));
-  }
-
-  auto scalarType = ptType.scalarType().value();
-  auto glowElemKind = scalarTypeToElemKind(scalarType);
-  return glow::Type(glowElemKind, dims, scale, zero_point);
-}
-
 at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
   std::vector<int64_t> sizes;
   for (const auto dim : glowType.dims()) {
@@ -248,14 +219,25 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
 
 glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor) {
   auto ptType = *c10::TensorType::create(ptTensor);
-  if (ptTensor.is_quantized()) {
-    auto glowType =
-        ptQTypeToGlowQType(ptType, ptTensor.q_scale(), ptTensor.q_zero_point());
-    return glow::Tensor(ptTensor.data_ptr(), &glowType);
-  } else {
-    auto glowType = ptTypeToGlowType(ptType);
-    return glow::Tensor(ptTensor.data_ptr(), &glowType);
+
+  DCHECK(ptType.scalarType().has_value())
+      << "TensorType has no associated scalar type.";
+  const auto concreteSizes = ptType.sizes().concrete_sizes().value();
+  std::vector<size_t> dims;
+  for (const auto &size : concreteSizes) {
+    dims.push_back(static_cast<size_t>(size));
   }
+
+  auto scalarType = ptType.scalarType().value();
+  auto glowElemKind = scalarTypeToElemKind(scalarType);
+
+  glow::Type glowType;
+  if (ptTensor.is_quantized()) {
+    glowType = glow::Type(glowElemKind, dims, ptTensor.q_scale(), ptTensor.q_zero_point());
+  } else {
+    glowType = glow::Type(glowElemKind, dims);
+  }
+  return glow::Tensor(ptTensor.data_ptr(), &glowType);
 }
 
 void fuseKnownPatterns(std::shared_ptr<torch::jit::Graph> &graph) {
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -233,7 +233,8 @@ glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor) {
 
   glow::Type glowType;
   if (ptTensor.is_quantized()) {
-    glowType = glow::Type(glowElemKind, dims, ptTensor.q_scale(), ptTensor.q_zero_point());
+    glowType = glow::Type(glowElemKind, dims, ptTensor.q_scale(),
+                          ptTensor.q_zero_point());
   } else {
     glowType = glow::Type(glowElemKind, dims);
   }
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -107,8 +107,10 @@ glow::ElemKind scalarTypeToElemKind(c10::ScalarType ty) {
     return ElemKind::Int64ITy;
   } else if (ty == at::kBool) {
     return ElemKind::BoolTy;
+  } else if (ty == at::kQInt8) {
+    return ElemKind::Int8QTy;
   } else {
-    LOG(DFATAL) << "Not supported yet.";
+    LOG(DFATAL) << ty << "Not supported yet.";
     return ElemKind::Int64ITy;
   }
 }
@@ -147,7 +149,9 @@ void glowCustomFuse(std::shared_ptr<torch::jit::Graph> &g,
 }
 
 void registerGlowOp(const c10::Symbol &symbol) {
+
   // Register dummy nodes used by custom fusers.
+  registerDummyOperator("glow::unpacked_quantized_conv2d");
   registerDummyOperator("glow::unpacked_quantized_linear");
 
   auto options = c10::OperatorOptions();
@@ -213,7 +217,23 @@ glow::Type ptTypeToGlowType(const c10::TensorType &ptType) {
   }
 
   auto scalarType = ptType.scalarType().value();
-  return glow::Type(scalarTypeToElemKind(scalarType), dims);
+  auto glowElemKind = scalarTypeToElemKind(scalarType);
+  return glow::Type(glowElemKind, dims);
+}
+
+glow::Type ptQTypeToGlowQType(const c10::TensorType &ptType, const float &scale,
+                              const int32_t &zero_point) {
+  DCHECK(ptType.scalarType().has_value())
+      << "TensorType has no associated scalar type.";
+  const auto concreteSizes = ptType.sizes().concrete_sizes().value();
+  std::vector<size_t> dims;
+  for (const auto &size : concreteSizes) {
+    dims.push_back(static_cast<size_t>(size));
+  }
+
+  auto scalarType = ptType.scalarType().value();
+  auto glowElemKind = scalarTypeToElemKind(scalarType);
+  return glow::Type(glowElemKind, dims, scale, zero_point);
 }
 
 at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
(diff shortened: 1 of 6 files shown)


torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def test_f(a):
+        q = torch.nn.quantized.Quantize(1/128, 3, torch.quint8)
+        dq = torch.nn.quantized.DeQuantize()
+        conv = torch.nn.quantized.Conv2d(3, 3, [2, 2])
+        return dq(conv(q(a)))
+
+    x = torch.randn([1, 3, 3, 3])
+
+    jitVsGlow(test_f, x, expected_fused_ops={"aten::relu",
```
zrphercule: expected fused ops list need to be updated.

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -213,7 +217,23 @@ glow::Type ptTypeToGlowType(const c10::TensorType &ptType) {
   }
 
   auto scalarType = ptType.scalarType().value();
-  return glow::Type(scalarTypeToElemKind(scalarType), dims);
+  auto glowElemKind = scalarTypeToElemKind(scalarType);
+  return glow::Type(glowElemKind, dims);
+}
+
+glow::Type ptQTypeToGlowQType(const c10::TensorType &ptType, const float &scale,
```
jackm321: Imo, instead of duplicating this, maybe let's just delete ptTypeToGlowType and ptQTypeToGlowQType and do this in ptTensorToGlowTensor since 1. these are only used there and 2. TensorType doesn't contain all the needed info to build a glow type since scale and zero point are missing

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -454,6 +454,21 @@ struct AdaptiveAvgPoolInputs {
...
+  enum {
+    input = 0, // NCHW
+    weights = 1,
+    bias = 2,
+    stride = 3,
+    padding = 4,
+    dilation = 5,
+    group = 6,
+    scale = 7,
+    offset = 8,
```
jackm321: zero_point?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1460,6 +1483,100 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
...
+        getGlowNodeValueForValue(inputs[QuantizedUnpackedConv2dInputs::bias]));
+  } else {
+    glow::Tensor biasT(glow::ElemKind::FloatTy, {weightsShape.n});
+    biasT.zero();
+    glow::Constant *biasConstant =
+        F_.getParent()->createConstant("qconv_bias", std::move(biasT));
+    bias = biasConstant->getOutput();
+  }
+  auto biasType =
+      F_.getParent()->uniqueType(glow::ElemKind::Int32QTy, bias.dims(), 1.0, 0);
```
jackm321: I think the bias scale needs to be input scale * weight scale. @rdzhabarov can you recall if that's correct?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1460,6 +1483,100 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
...
+  glow::TypeRef outTy = F_.getParent()->uniqueType(
+      glow::ElemKind::Int8QTy, outDims, outScale, outOffset);
+
+  glow::ConvolutionNode *qconv =
+      F_.createConv("qconv", input, weights, bias, outTy, kernels, strides,
+                    pads, groups, dilation);
+  glow::TransposeNode *output = F_.createTranspose(
+      "qconv_output_transposed", qconv->getResult(), NHWC2NCHW);
+
+  return addValueMapping(outputs[0], output->getResult());
```
jackm321: need to call rescaleIntToUInt

torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -0,0 +1,32 @@
...
+    # Due to the quantization error between
+    # PyTorch and Glow, we would like to use some
+    # determined test data instead of random ones
+    # x = torch.randn([3, 3, 3, 3])
+    # w = torch.randn([3, 3, 3, 3])
+    # b = torch.randn([3])
+
+    x = torch.tensor([[[[5., 6.], [7., 8.]]]])
+    w = torch.tensor([[[[1., 2.], [3., 4.]]]])
+    b = torch.zeros(1)
```
jackm321: Test with a non-zero bias?

jackm321: Also we should test it without a bias to make sure that's working

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -227,8 +247,15 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
 }
 
 glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor) {
-  auto glowType = ptTypeToGlowType(*c10::TensorType::create(ptTensor));
-  return glow::Tensor(ptTensor.data_ptr(), &glowType);
+  auto ptType = *c10::TensorType::create(ptTensor);
+  if (ptTensor.is_quantized()) {
```
jackm321: Is this path being tested anywhere? If not let's make sure we make a test for it.

torch_glow/tests/nodes/quantized_conv2d_test.py
zrphercule: Yeah agree

torch_glow/src/PyTorchCommon.cpp
zrphercule: I can do it by using torch.nn.quantized.Conv2d, however that might introduce the error of accuracy.

zrphercule:
@jackm321 Hi Jack, I added one more skipped testcase to test the ptTensorToGlowTensor workflow. Basicly in that workflow, weights are generated randomly and quantized using "aten::_empty_affine_quantized". Since we are not supporting this aten symbol yet, the quantized tensors will be excluded from glowFusionNode:

glow_graph, graph(%X : Float(*, *, *, *)):
  %51 : int[] = prim::Constant[value=[1]]()
  %50 : int[] = prim::Constant[value=[1, 1, 2, 2]]()
  %6 : int = prim::Constant[value=12]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %7 : int = prim::Constant[value=0]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %8 : Device = prim::Constant[value="cpu"]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %9 : bool = prim::Constant[value=0]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %10 : float = prim::Constant[value=1]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %12 : None = prim::Constant()
  %16 : int = prim::Constant[value=6]() # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:86:0
  %w : Tensor = aten::_empty_affine_quantized(%50, %6, %7, %8, %9, %10, %7, %12) # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:83:0
  %b : Float(*) = aten::zeros(%51, %16, %7, %8, %9) # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/conv.py:86:0
  %53 : Tensor = glow::FusionGroup_0(%w, %b, %X)
  return (%53)
with glow::FusionGroup_0 = graph(%3 : Tensor,
      %4 : Float(*),
      %18 : Float(*, *, *, *)):
  %19 : float = prim::Constant[value=0.5]()
  %20 : int = prim::Constant[value=0]()
  %21 : int = prim::Constant[value=13]()
  %input : Tensor = aten::quantize_per_tensor(%18, %19, %20, %21), scope: Quantize # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/linear.py:39:0
  %16 : int = prim::Constant[value=1]()
  %17 : int[] = prim::ListConstruct(%16, %16), scope: QuantizedConv2d
  %14 : int = prim::Constant[value=0]()
  %15 : int[] = prim::ListConstruct(%14, %14), scope: QuantizedConv2d
  %12 : int = prim::Constant[value=1]()
  %13 : int[] = prim::ListConstruct(%12, %12), scope: QuantizedConv2d
  %8 : int = prim::Constant[value=1]()
  %9 : float = prim::Constant[value=1]()
  %10 : int = prim::Constant[value=0]()
  %11 : Tensor = glow::unpacked_quantized_conv2d(%input, %3, %4, %17, %15, %13, %8, %9, %10)
  %1 : Tensor = aten::dequantize(%11), scope: DeQuantize # /Users/zrphercule/zfacebook/glow-torch/pytorch/torch/nn/quantized/modules/linear.py:69:0
  return (%1)

Therefore this workflow is tested. 

However since the accuracy problem still exist, this test only has like 20~30% chance to pass, and we cannot control it. Therefore I would like to add it with a skipped label.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Wonder what is our conclusion?

jackm321: Let's try changing it, I think this is correct

jackm321:
@zrphercule My point was that the changes in ptTensorToGlowTensor aren't tested and are independent of the rest of the code in this diff right? If so maybe we should save those changes for later and add some unit tests for them

jackm321:
@zrphercule it looks like the _empty_affine_quantized op is outside of the glow::FusionGroup and is just producing a quantized input which is exactly what we'd expect if testing quantized inputs to the glow graph, I could be missing something though?

zrphercule:
> @zrphercule it looks like the _empty_affine_quantized op is outside of the glow::FusionGroup and is just producing a quantized input which is exactly what we'd expect if testing quantized inputs to the glow graph, I could be missing something though?

You are right, and that is why I added this test to test ptTensorToGlowTensor, because we would have an outsider quantized tensor to feed into fused node.
Right now it seems the functionality works fine, however it only has ~25% chance of producing same result to pytorch, therefore I skipped it.

torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+
+from tests.utils import jitVsGlow
+import pytest
+
+
+def test_quantized_conv2d():
+    """Basic test of the PyTorch quantized::relu Node on Glow."""
```
jackm321: not quantized::relu

torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -0,0 +1,55 @@
...
+                                                   "aten::dequantize"})
+
+    jitVsGlow(test_f, x, w, b_zero, expected_fused_ops={"aten::quantize_per_tensor",
+                                                        "glow::unpacked_quantized_conv2d",
+                                                        "aten::dequantize"})
+
+
+@pytest.mark.skip(reason="accuracy between glow&ytorch")
+def test_quantized_conv2d_nonfunctional():
+    """Basic test of the PyTorch quantized::relu Node on Glow."""
```
jackm321: same

torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -0,0 +1,55 @@
...
+    jitVsGlow(test_f, x, w, b, expected_fused_ops={"aten::quantize_per_tensor",
+                                                   "glow::unpacked_quantized_conv2d",
+                                                   "aten::dequantize"})
+
+    jitVsGlow(test_f, x, w, b_zero, expected_fused_ops={"aten::quantize_per_tensor",
+                                                        "glow::unpacked_quantized_conv2d",
+                                                        "aten::dequantize"})
+
+
+@pytest.mark.skip(reason="accuracy between glow&ytorch")
```
jackm321: I'm not sure I understand, it looks the same but just has random weight initialization? Won't this always be different accross different runs?

zrphercule: Yes, but,
1. it has a empty_affine_quantize out of our fusion node, which gives a quantized tensor as a fusion node input, and invoke the code we have in https://github.com/pytorch/glow/pull/3635. Without 3635 it will crush.

2. Since it has random weight initialization, it generate error, and have a great chance to fail, although sometimes it can pass. This flakiness is similar to why we cannot have random float number test in regular quantized conv.

jackm321: 2. Yeah exactly, if it's random initializing weights, we can't run it twice and compare the results so it's probably not a good test

zrphercule: > 1. Yeah exactly, if it's random initializing weights, we can't run it twice and compare the results so it's probably not a good test

Agree, however that is the only way I know to exclude a quantized tensor from our fusion node, or else it will be fused in...

yinghai: Why can't we provide a hand written set of weights?

zrphercule: > Why can't we provide a hand written set of weights?

I can try and enbale this test in that new diff.

PR #3574, opened 2019-09-30
Add operator kind to OPERATOR level trace output

virtan:
Summary: Add operator kind to OPERATOR level trace output

Differential Revision: D17643033

### DIFF
--- a/tests/unittests/TraceEventsTest.cpp
+++ b/tests/unittests/TraceEventsTest.cpp
@@ -108,6 +108,63 @@ public:
     }
   }
 
+  // In the function below we're trying to compare auto instrumented trace
+  // output with optimized model graph. Since the code generation may make
+  // further changes to the graph representation we limit checking to a set of
+  // convolution / maxpool / cpumaxsplat kinds which are never changed.
+  std::set<std::string> checkedKinds = {"convolution", "maxpool",
+                                        "cpumaxsplat"};
+
+  template <class ExecutionEngineTy>
+  std::vector<std::pair<std::string, std::string>>
+  prepareKindsForComparison(ExecutionEngineTy &ExecutionEngine) {
+    std::vector<std::pair<std::string, std::string>> expectedKinds;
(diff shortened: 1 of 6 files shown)


tests/unittests/TraceEventsTest.cpp
```diff
@@ -108,6 +108,36 @@ class TraceEventsTest : public ::testing::TestWithParam<std::string> {
...
+  // In the function below we're trying to compare auto instrumented trace
+  // output with optimized model graph. Since the code generation may make
+  // further changes to the graph representation we limit checking to a set of
+  // convolution / maxpool / cpumaxsplat kinds which are never changed.
+  std::set<std::string> checkedKinds = {"convoluition", "maxpool",
+                                        "cpumaxsplat"};
+
+  void checkEventMetadata(
+      const std::vector<TraceEvent> &traceEvents,
+      std::vector<std::pair<std::string, std::string>> expected) {
```
rdzhabarov: it's test, does not matter to much, but can you make it `const vector&`

tests/unittests/TraceEventsTest.cpp
```diff
@@ -314,6 +360,21 @@ TEST_P(TraceEventsTest, manualAndAutomatic) {
   cctx.backendOpts.autoInstrument = true;
   EE_.compile(cctx);
 
+  std::vector<std::pair<std::string, std::string>> expectedKinds;
```
rdzhabarov: can you make this as a helper method which extracts node kinds from the given function and reuse in both tests

tests/unittests/TraceEventsTest.cpp
virtan: I actually repeated the interface of the same function above. But yeah, of course, no any reason to copy this vector here.

tests/unittests/TraceEventsTest.cpp
virtan: Ok

PR #3575, opened 2019-09-30
[LayoutConverter] Modify convertMaxPoolToNCHWPool to return argmax too

SplitInfinity:
**Summary**
`MaxPoolNode` has two outputs, `Result` and `Argmax`. The current
implementation of `convertMaxPoolToNCHWPool` returns one `Node`
corresponding to Result, but does not return any node corresponding to
`Argmax`. This commit fixes that by modifying the return type to
`std::pair<Node*, Node*>` and by returning both the `Result` and `Argmax` of
the new `MaxPoolNode` created inside the function (transposed back to the
original NHWC layout).

**Testing**
All unit tests pass.

### DIFF
--- a/include/glow/Backends/LayoutConverter.h
+++ b/include/glow/Backends/LayoutConverter.h
@@ -42,9 +42,11 @@ inline Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
   return NR;
 }
 
-/// Convert regular pool nodes (that use NHWC) into backend-specific nodes using
-/// NCHW.
-inline Node *convertMaxPoolToNCHWPool(MaxPoolNode *PN, Function *F) {
+/// Convert regular max pool nodes (that use NHWC) into backend-specific nodes
+/// using NCHW. \returns a pair containing the new MaxPool result and argmax
+/// that the result and argmax of the original MaxPool should be replaced with.
+inline std::pair<Node *, Node *> convertMaxPoolToNCHWPool(MaxPoolNode *PN,
+                                                          Function *F) {
   // Convert input from NHWC (Glow's default) into NCHW.
   auto *NI = F->createTranspose("maxpool.input", PN->getInput(), NHWC2NCHW);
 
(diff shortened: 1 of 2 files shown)


SplitInfinity:
You might be wondering why this pull request goes to the trouble of adding the `Argmax` output to the layout converter function but then ignores it in the part of the OpenCL backend that does layout conversion. This layout conversion has only applied to the input/output this whole time, and that the argmax output has been computed using the old, untransformed MaxPoolNode that does not get DCE'd when the argmax output has a use. OpenCL only has NHWC kernels to compute argmax, so replacing the NHWC MaxPool's Argmax output with the Argmax output of the NCHW MaxPool produces incorrect results. I need this for another backend, so adding those missing OpenCL kernels is not my highest priority. If everyone is okay with this, I'll open issues to track this follow up task of adding NCHW kernels that also compute argmax.

SplitInfinity:
You may be wondering why this PR goes to the trouble of adding the extra argmax output to the layout converter function but then doesn't even use it. The `MaxPoolWithArgmax` unit tests work at present because they use layout converted NCHW pooling kernels to compute the result and normal NHWC pooling kernels to compute argmax:
```
  0 %pool11 = maxpool @out %save_result, @in %input { Kernels: [4, 4], Strides: [4, 4], Pads: [0, 0, 0, 0], Layout: 1}
  1 %transposeInput_res = allocactivation  { Ty: float<1 x 4 x 4 x 3>} // size: 192 // Users: @out 6, @in 5, @out 2
  2 %transposeInput = transpose @out %transposeInput_res, @in %input { Shuffle: [0, 2, 3, 1]}
  3 %save_argmax2 = tensorview @in %save_argmax { Ty: index64<1 x 1 x 1 x 3>, Offsets: [0, 0, 0, 0]} // Users: @out 5
  4 %pool_res = allocactivation  { Ty: float<1 x 1 x 1 x 3>} // size: 12 // Users: @out 7, @out 5
  5 %pool = maxpoolwithargmax @out %pool_res, @in %transposeInput_res, @out %save_argmax2 { Kernels: [4, 4], Strides: [4, 4], Pads: [0, 0, 0, 0], Layout: 0}
  6 %dealloc_transposeInput_res = deallocactivation @out %transposeInput_res // size: 192
  7 %dealloc_pool_res = deallocactivation @out %pool_res // size: 12
```

The NHWC pooling kernels in the OpenCL backend compute argmax but the NCHW kernels do not. Therefore, when the argmax of the old, NHWC node is replaced with that of the NCHW node during layout conversion, these unit tests no longer pass. I need this for another backend so I if everyone is okay with it, I will create issues to track this problem.

include/glow/Backends/LayoutConverter.h
```diff
@@ -42,9 +42,10 @@ inline Node *convertConvToNCHWConv(ConvolutionNode *CN, Function *F) {
   return NR;
 }
 
-/// Convert regular pool nodes (that use NHWC) into backend-specific nodes using
-/// NCHW.
-inline Node *convertMaxPoolToNCHWPool(MaxPoolNode *PN, Function *F) {
+/// Convert regular max pool nodes (that use NHWC) into backend-specific nodes
+/// using NCHW.
+inline std::pair<Node *, Node *> convertMaxPoolToNCHWPool(MaxPoolNode *PN,
```
rdzhabarov: can you update, \returns goxygen

PR #3588, opened 2019-10-02
Refactored load device config functionality 

### LINKED ISSUE #3556: Add Config File support to RecommendationSystemTest.
Now that RecommendationSystemTest has been ported to HostManager directly, we need to add support for creating the DeviceConfigs from a config file. This is to allow heterogeneous device support.

gcatron:
Summary:
Refactored load device config functionality from loader into HostManager  this way it can be used elsewhere like in Recommendation System Test.
Documentation: 

[Fixes #3556 ]

Test Plan:
./image-classifier  ../images/imagenet/*.png -m=resnet50 -model-input-name=gpu_0/data -image-mode=0to1 -use-imagenet-normalization -load-profile=resnet50.yaml -load-device-configs="cpuConfigs.yaml"

./tests/CPURecommendationSystemTest -load-device-configs="cpuConfigs.yaml"

### DIFF
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -213,6 +213,14 @@ public:
 
   ~HostManager();
 };
+
+/// If the device config file specified in loadDeviceConfigsFileOpt is
+/// available, load \p configs from the file. Otherwise, create \p numDevices
+/// number of devices based on \p backendName.
+std::vector<std::unique_ptr<runtime::DeviceConfig>>
+generateDeviceConfigs(unsigned int numDevices, llvm::StringRef backendName,
+                      size_t memSize = 0);
+
 } // namespace runtime
 } // namespace glow
 #endif // GLOW_RUNTIME_HOSTMANAGERR_HOSTMANAGER_H
--- a/lib/Runtime/HostManager/HostManager.cpp
+++ b/lib/Runtime/HostManager/HostManager.cpp
@@ -46,6 +46,13 @@ llvm::cl::opt<std::string> loadBackendSpecificOptionsOpt(
     llvm::cl::cat(hostManagerCat));
 } // namespace
 
+/// The device configs file used for Runtime.
+llvm::cl::opt<std::string> loadDeviceConfigsFileOpt(
+    "load-device-configs",
+    llvm::cl::desc("Load device configs used in Runtime"),
+    llvm::cl::value_desc("configs.yaml"), llvm::cl::Optional,
+    llvm::cl::cat(hostManagerCat));
+
 HostManager::HostManager(const HostConfig &hostConfig) : config_(hostConfig) {}
 
 HostManager::HostManager(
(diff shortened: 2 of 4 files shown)


tests/unittests/RecommendationSystemTest.cpp
```diff
@@ -33,6 +33,8 @@ constexpr size_t MAX_MEMORY = 64e+9;
 
 using namespace glow;
 
+extern llvm::cl::opt<std::string> loadDeviceConfigsFileOpt;
```
opti-mix: May be this declaration should be in HostManager.h and then you wouldn't need it here?

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -46,6 +46,13 @@ llvm::cl::opt<std::string> loadBackendSpecificOptionsOpt(
     llvm::cl::cat(hostManagerCat));
 } // namespace
 
+/// The device configs file used for Runtime.
+llvm::cl::opt<std::string> loadDeviceConfigsFileOpt(
+    "load-device-configs",
+    llvm::cl::desc("Load device configs used in Runtime"),
+    llvm::cl::value_desc("profile.yaml"), llvm::cl::Optional,
```
opti-mix: `profile.yaml` is not the best value for a config file option. Maybe `config.yaml` is a better choice?

tools/loader/Loader.cpp
```diff
@@ -45,6 +45,9 @@ static llvm::cl::opt<bool, true>
                    llvm::cl::desc("Enable rowwise quantized fully connected."),
                    llvm::cl::location(enableRowwiseOpt), llvm::cl::init(false));
 
+/// The device configs file used for Runtime.
+extern llvm::cl::opt<std::string> loadDeviceConfigsFileOpt;
```
opti-mix: See my other comment.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -213,6 +213,19 @@ class HostManager final {
 
   ~HostManager();
 };
+
+/// Helper to get the parameters in DeviceConfig from \p str. The \p str has
+/// multiple lines, and each line with this format : "str1" : "str2".
+llvm::StringMap<std::string> getBackendParams(std::string &str);
```
opti-mix: Is it ever used independently of `generateDeviceConfigs` or is it just an implementation detail? If so, move the function definition as a `static` function into a cpp file where `generateDeviceConfigs` is defined.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -448,3 +455,59 @@ HostManager::runNetwork(llvm::StringRef networkName,
...
+}
+
+/// If the device config file \p loadDeviceDoncfigsFile available, load \p
+/// configs from the file. Otherwise, create \p numDevices number of devices
+/// based on \p backendName.
+std::vector<std::unique_ptr<runtime::DeviceConfig>>
+runtime::generateDeviceConfigs(unsigned int numDevices,
+                               llvm::StringRef backendName, size_t memSize) {
+  std::vector<std::unique_ptr<runtime::DeviceConfig>> configs;
+  if (loadDeviceConfigsFileOpt.empty()) {
```
opti-mix: I kind of like you now explicitly use loadDeviceConfigsFileOpt  in this function and do not take it as a parameter anymore. But it also makes it impossible to use this function if a profile-file-name comes from somewhere else rather then from this option. Is it a problem?

gcatron: I don't think so. It might need to be refactored down the road if we load all options from a config file, but for now I think using the option explicitly should be fine.

PR #3592, opened 2019-10-03
[utils] add imagenet-process script

nickgg:
Summary: Needed to normalize some images and had to look up how, here's a script so it's easier next time.

Documentation: ~self documenting~

Test Plan: ran the script on some un normalized JPEGs, then ran image-classifier on the processed pngs and verified the top label


nickgg:
autopep8

PR #3597, opened 2019-10-04
add ReluInst for backends that don't lower Relu

nickgg:
Summary: Add an Instruction equivalent to ReluNode so it can be used in backends if they wish to disabled Relu lowering.

Documentation: NFC

Test Plan: No backends use Relu Instructions yet, so no tests.

### DIFF
--- a/tools/ClassGen/InstrGen.cpp
+++ b/tools/ClassGen/InstrGen.cpp
@@ -529,6 +529,16 @@ int main(int argc, char **argv) {
   //===--------------------------------------------------------------------===//
   //                Non-linearities
   //===--------------------------------------------------------------------===//
+  BB.newInstr("Relu")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("Src", OperandKind::In)
+      .inplaceOperand({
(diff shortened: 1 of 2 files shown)


tools/ClassGen/InstrGen.cpp
```diff
@@ -529,6 +529,17 @@ int main(int argc, char **argv) {
...
+  BB.newInstr("Relu")
+      .addOperand("Dest", OperandKind::Out)
+      .addOperand("Src", OperandKind::In)
+      .inplaceOperand({
+          "Dest",
+          "Src",
+      })
+      .dataParallel()
+      .autoVerify(VerifyKind::SameElementType, {"Dest", "Src"})
+      .autoVerify(VerifyKind::SameShape, {"Dest", "Src"})
```
SplitInfinity: I *think* you can combine these into one check using `VerifyKind::SameType` but I'm not sure.

nickgg: Perhaps, this matches the style of the surrounding instructions though. Can we do that in a follow up?

PR #3599, opened 2019-10-04
[OpenCL] Add src operand to maxpoolgrad kernel

SplitInfinity:
**Summary**
This commit adds a src operand to the OpenCL maxpoolgrad kernel. This
operand was added to the `MaxPoolWithArgmaxGradInst` class in an earlier
commit, but was not added to the corresponding OpenCL kernel. This
broke the code in the OpenCL backend that sets kernel arguments using
operands from the corresponding instruction because it sets them in the order
of that they are declared in the instruction class.

**Testing**
All unit tests pass.

### DIFF
--- a/lib/Backends/OpenCL/kernels.cl
+++ b/lib/Backends/OpenCL/kernels.cl
@@ -1322,11 +1322,13 @@ __kernel void maxpoolwithargmaxgradK(__global float *dest,
 }
 
 __kernel void maxpoolwithargmaxgradW(__global void *mem, cl_uint32_t dest,
-                                     cl_uint32_t argmax, cl_uint32_t destGrad,
-                                     cl_uint32_t srcGrad,
+                                     cl_uint32_t src, cl_uint32_t argmax,
+                                     cl_uint32_t destGrad, cl_uint32_t srcGrad,
                                      cl_uint32_t kernelSize, cl_uint32_t stride,
                                      PaddingTLBR pads, ShapeNHWC srcGradDim,
                                      ShapeNHWC destDim) {
+  // src operand is present on the instruction but not needed by the OpenCL
+  // kernel.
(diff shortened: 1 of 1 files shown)


lib/Backends/OpenCL/kernels.cl
```diff
@@ -1292,7 +1292,7 @@ __kernel void maxpoolwithargmaxW(__global void *mem, cl_uint32_t dest,
                      pads, odim, idim);
 }
 
-__kernel void maxpoolwithargmaxgradK(__global float *dest,
+__kernel void maxpoolwithargmaxgradK(__global float *dest, __global float *src,
```
nickgg: If we don't use src here can we comment the name to make that clear?

SplitInfinity:
OpenCL doesn't need this extra operand, but I'm not sure there's a way to skip it at present. I can also add functionality to `setKernelArgsForBuffers` to skip operands if that seems like a cleaner solution.

nickgg:
I just meant ` __global float * \*src*\,`

SplitInfinity:
Yep, I understood, I typed my earlier comment and submitted it before I saw yours

SplitInfinity:
Hm, apparently that's not allowed in OpenCL:
```
<program source>:1287:53: error: parameter name omitted
                                 cl_uint32_t /*src*/, cl_uint32_t argmax,
```

SplitInfinity:
Hmm, apparently commenting out parameter name is not allowed in OpenCL:
```
<program source>:1287:53: error: parameter name omitted
                                 cl_uint32_t /*src*/, cl_uint32_t argmax,
```

SplitInfinity:
Look like commenting out parameter names is not allowed in OpenCL:
```
<program source>:1325:57: error: parameter name omitted
                                     cl_uint32_t /*src*/, cl_uint32_t argmax,
```

SplitInfinity:
I added a comment about this in the W kernel and removed the operand entirely from the K kernel.

PR #3606, opened 2019-10-07
Add numerical equivalence check

waz-fb:
Summary:
Add numerical equivalence check to GraphOptz.OptimizeClipFunnel test case

Documentation:

Test Plan:
ninja test
image


SplitInfinity:
This does not guarantee there there will be input values outside the clip range. I would set some input elements explicitly so that they are outside the range.

SplitInfinity:
This still does not *guarantee* that there will be input values outside the clip range. I would set some input elements explicitly so that they are outside the range, e.g. 
```
bindings_.get(A)->getHandle().raw(0) = -1000
```

PR #3607, opened 2019-10-07
evalBatch helper function in EE

mortzur:
Summary:
We have some evaluation code that gets duplicated each time we add a computer vision training example. Evaluation here means running inference and counting correct estimates.
I thought that a helper function may be useful here, so I added evalBatch in ExecutionEngine.
I modified mnist example to use this function while maintaining its original output.

Test Plan:
ninja test
./bin/mnist
Screen Shot 2019-10-07 at 2 59 23 PM

### DIFF
--- a/include/glow/ExecutionEngine/ExecutionEngine.h
+++ b/include/glow/ExecutionEngine/ExecutionEngine.h
@@ -164,6 +164,16 @@ void runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
               llvm::ArrayRef<Placeholder *> ph, llvm::ArrayRef<Tensor *> inputs,
               llvm::StringRef name = "");
 
+/// Runs \p numMinibatchRuns iterations of the compiled function called \p name.
+/// The method updates a global counter and future invocations of this method
+/// continue running iterations of the batch at the next available slice.
+/// The provided callback function \p cb is invoked on each sample.
+void evalBatch(
+    ExecutionEngine &EE, PlaceholderBindings &bindings, size_t numMinibatchRuns,
+    size_t &sampleCounter, Placeholder *inputPH, Placeholder *outputPH,
+    Tensor &samplesInput, Tensor &labelsInput, llvm::StringRef name,
+    std::function<void(const Tensor &sampleIn, const Tensor &sampleOut,
+                       const Tensor &label, size_t sampleIndex)> &&cb);
 } // namespace glow
 
 #endif // GLOW_EXECUTIONENGINE_EXECUTIONENGINE_H
--- a/examples/mnist.cpp
+++ b/examples/mnist.cpp
@@ -158,45 +158,35 @@ void validateModel(ExecutionEngine &EE, PlaceholderBindings &bindings,
   ::glow::convertPlaceholdersToConstants(F, bindings, {inputPH, outputPH});
   EE.compile(CompilationMode::Infer);
 
-  auto LIH = labelInputs.getHandle<int64_t>();
-
-  // Check how many examples out of eighty previously unseen digits we can
-  // classify correctly.
-  int rightAnswer = 0;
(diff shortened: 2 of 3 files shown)


include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -159,6 +159,17 @@ void runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
               llvm::ArrayRef<Placeholder *> ph, llvm::ArrayRef<Tensor *> inputs,
               llvm::StringRef name = "");
 
+/// Runs \p numMinibatchRuns iterations of the compiled function called \p name.
+/// The method updates a global counter and future invocations of this method
+/// continue running iterations of the batch at the next available slice.
+/// Each sample output is copmared VS ground-truth \p labelsInput, and
```
opti-mix: copmared -> compared

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -159,6 +159,17 @@ void runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
...
+/// \p numCorrectEstimates counter is updated correspondingly.
+void evalBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
```
opti-mix: Why do you put it into ExecutionEngine.h, which is a very generic header, unrelated to testing? Wouldn't some kind of TestUtils.h be a better place for it?

mortzur: This helper is very similar to runBatch (and is useful after using runBatch to train a model) so I thought it makes sense to place them together.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -159,6 +159,17 @@ void runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
...
+               size_t numMinibatchRuns, size_t &sampleCounter,
+               Placeholder *inputPH, Placeholder *outputPH,
+               Tensor &samplesInput, Tensor &labelsInput, llvm::StringRef name,
+               size_t &numCorrectEstimates,
+               std::vector<Tensor> *estimates = nullptr);
```
putivsky: How about two methods, one just returns back TWO tensors (output and labels) through callback, and second method given two tensors "visualizes" and calculating accuracy?

mortzur: Following an offline chat, modified evalBatch to take a callback (invoked on each sample) which allows the user to control results accumulation and visualization.

include/glow/ExecutionEngine/ExecutionEngine.h
```diff
@@ -159,6 +159,16 @@ void runBatch(ExecutionEngine &EE, PlaceholderBindings &bindings,
...
+/// Runs \p numMinibatchRuns iterations of the compiled function called \p name.
+/// The method updates a global counter and future invocations of this method
+/// continue running iterations of the batch at the next available slice.
+/// The provided callback function \p cb is invoked on each sample.
+void evalBatch(
+    ExecutionEngine &EE, PlaceholderBindings &bindings, size_t numMinibatchRuns,
+    size_t &sampleCounter, Placeholder *inputPH, Placeholder *outputPH,
+    Tensor &samplesInput, Tensor &labelsInput, llvm::StringRef name,
+    std::function<void(const Tensor &sampleIn, const Tensor &sampleOut,
+                       const Tensor &label, size_t sampleIndex)> &&cb);
```
opti-mix: Maybe introduce a nicely named type alias for this very long function type?

SplitInfinity:
@mortzur 

Do you plan to land this?

mortzur:
yes, will rebase and land

PR #3615, opened 2019-10-10
Refactored Provisioner to reduce complexity and memory consumption

gcatron:
Summary:
Refactored the provisioner, it now compiles and loads functions one logicalDevice at a time. This will decrease memory overhead. Also tried to split out different phases of provisioning into their own functions so it would be easier to reason over. 

Documentation:

Test Plan: ninja check

### DIFF
--- a/include/glow/Runtime/Provisioner/Provisioner.h
+++ b/include/glow/Runtime/Provisioner/Provisioner.h
@@ -45,12 +45,11 @@ public:
   Error removeFunction(llvm::StringRef name);
 
 private:
-  /// Pointer to backend used for compilation. This currently gets reset per
-  /// device to ensure the correct backed per device.
-  std::vector<std::unique_ptr<Backend>> backends_;
+  /// Map of backends for all devices, one backend per device type.
+  std::unordered_map<std::string, std::unique_ptr<Backend>> backends_;
 
-  /// Map of compiledFunction unique pointers. This maintains ownership of the
-  /// functions.
+  /// Map of compiledFunction unique pointers. This maintains
+  /// ownership of the functions.
   std::unordered_map<std::string, std::unique_ptr<CompiledFunction>> functions_;
 
   /// Set of active functions - these are functions that are currently being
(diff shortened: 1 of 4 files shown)


include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -64,10 +64,28 @@ class Provisioner final {
...
 
-  /// Helper function to cleanup a provision call. On a success free resources
-  /// that are no longer needed by the compiledFunctions. On failure free the
+  /// Helper function to cleanup a provision call. On failure free the
   /// compiledFunctions that were created.
   void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
+
+  /// Helper function to parse the DAG and generate logicalDevices.
+  std::map<DeviceIDTy, std::vector<DAGNode *>>
+  generateLogicalDevices(DAGListTy &networks);
```
opti-mix: Function argument could be `const &`, right?

Also please check all other functions you introduced and make arguments `const` where applicable.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -143,82 +316,59 @@ Error Provisioner::provision(DAGListTy &networks, Module &module,
...
-  std::map<std::string, size_t> startPos;
-  for (unsigned i = 0; i < logicalDeviceSize.size(); i++) {
-    std::string backendName =
-        logicalDeviceBackendName[logicalDeviceSize[i].first];
-    // Find the start point of each backendName device.
-    if (startPos.find(backendName) == startPos.end()) {
-      startPos[backendName] = 0;
+    // Free up memory no longer needed by the compiledFunction.
+    for (auto &func : functionMap) {
+      func.second->freeCompilationResources();
```
arunm-git: Wouldn't this free up memory used by a duplicated function too (before we are done with it)?

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -59,43 +59,163 @@ Provisioner::Provisioner(DeviceManagerMapTy &devices) {
...
+        logicalDevice.second) {
+      // There is enough space, assign the logical device to this physical
+      // device, increment the iterator and update the available memory.
+      deviceAssignment.emplace(
+          logicalDevice.first,
+          deviceMemoryMap[backendName][currentPosition].first);
+      deviceMemoryMap[backendName][currentPosition].second -=
+          logicalDevice.second;
+
+      // Check if we are at the end of the vector of devices.
```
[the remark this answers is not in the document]
gcatron: Yeah, this approach attempts to evenly spread networks to prevent a bunch of small networks clumping on one device.

lib/Runtime/Provisioner/Provisioner.cpp
gcatron: Good catch this was supposed to iterate over compiledFuncrtions.

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -48,54 +48,181 @@ auto sortMostMemory = [](const std::pair<DeviceIDTy, uint64_t> &a,
...
+      // current largest amount of available space is enough to fit.
+      std::sort(deviceMemoryMap[backendName].begin(),
+                deviceMemoryMap[backendName].end(), sortMostMemory);
+      if (deviceMemoryMap[backendName][0].second >= logicalDevice.second) {
+        // There's a device that still has room, assign the network here.
+        deviceAssignment.emplace(logicalDevice.first,
+                                 deviceMemoryMap[backendName][0].first);
+        deviceMemoryMap[backendName][0].second -= logicalDevice.second;
+
+        // Since after sorting we were abel to add to device 0  set the current
```
arunm-git: typo s/abel/able/

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -111,114 +238,138 @@ Error Provisioner::provision(DAGListTy &networks, Module &module,
...
-            });
-        ready.wait();
-        DCHECK_NOTNULL(addErr.get());
-        if (*addErr.get()) {
+
+        auto compiledOrErr =
+            backends_[deviceBackendName]->compile(function, options);
+        // Check to see if an error was encountered while compiling.
+        if (!compiledOrErr) {
+          // If and error occured, clean up provisioning state and return
```
arunm-git: an error

PR #3620, opened 2019-10-13
Initial step of exporting a glow Function before lowering

yinghai:
Summary:
Steps for a repro procedure for vendors
- Export Glow Function right before lowering
  - Export the net structure  <- this diff)
  - Export the weights and pack them together as a zip
- A `repro` binary to read in the zip file and execute the vendor backend
- Some randomization at net_runner.

Problem now, the ONNXWriter will mutate the input Function! This is not desirable. For the short term we need to clone a function. But exporter mutating a Function is a bit weird. Jordan, Roman

Differential Revision: D17897889

### DIFF
--- a/lib/Backends/NNPI/NNPI.cpp
+++ b/lib/Backends/NNPI/NNPI.cpp
@@ -16,6 +16,7 @@
 #include "NNPI.h"
 #include "NNPICompiledFunction.h"
 #include "NNPIDeviceManager.h"
+#include "glow/Exporter/ONNXModelWriter.h"
 #include "glow/Graph/Nodes.h"
 #include "glow/Optimizer/GraphOptimizerPipeline/Pipeline.h"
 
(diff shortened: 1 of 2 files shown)


lib/Backends/NNPI/NNPI.cpp
```diff
@@ -313,6 +315,16 @@ NNPIBackend::compile(Function *F, const BackendOptions &opts) const {
     LOG(INFO) << "Dumping net to " << fname;
     F->dumpDAG(fname);
   }
+  if (glow::onnxifi::GlowSaveModel) {
+    std::string fname = F->getName().str() + ".onnx";
+    LOG(INFO) << "Saving model to " << fname;
+    Error err = Error::empty();
+    { ONNXModelWriter onnxWR(fname, *F, 0, 0, &err, false); }
+    if (ERR_TO_BOOL(std::move(err))) {
+      llvm::errs() << "ONNXModelWriter failed to write model: " << fname
```
jackm321: LOG(ERROR)

PR #3621, opened 2019-10-14
[GraphOptimizer] Sink Transpose below Arithmetic nodes when one operand is a Constant

SplitInfinity:
**Summary**
This commit extends the optimization that sinks `Transpose` nodes below `Add`
nodes to handle the case in which one of the operands of the `Add` node is
a `Constant`.

**Testing**
This commit adds a unit test that tests that a `Transpose` node is sunk below
an `Add` node when one operand is a `Constant` and the other is a `Transpose` node.

```
$ ./tests/GraphOptzTest --gtest_filter=*sinkTransposeBelowArithmeticNodesWithConstantOperand*
Note: Google Test filter = *sinkTransposeBelowArithmeticNodesWithConstantOperand*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from GraphOptz
[ RUN      ] GraphOptz.sinkTransposeBelowArithmeticNodesWithConstantOperand
[       OK ] GraphOptz.sinkTransposeBelowArithmeticNodesWithConstantOperand (5 ms)
[----------] 1 test from GraphOptz (5 ms total)

[==========] 1 test from 1 test suite ran. (5 ms total)
[  PASSED  ] 1 test.
```

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -77,6 +77,56 @@ static bool shouldDeleteConstants(Function *F) {
   return true;
 }
 
+/// Helper that \returns the shuffle that inverts \p shuffle. For example, if
+/// \p shuffle is {3, 0, 1, 2}, then this function returns {1, 2, 3, 0}.
+static llvm::SmallVector<unsigned_t, max_tensor_dimensions>
+invertShuffle(llvm::ArrayRef<unsigned_t> shuffle) {
+  llvm::SmallVector<unsigned_t, max_tensor_dimensions> invertedShuffle;
+  invertedShuffle.resize(shuffle.size());
+
+  for (size_t i = 0; i < shuffle.size(); ++i) {
+    invertedShuffle[shuffle[i]] = i;
+  }
+
+  return invertedShuffle;
+}
+
+/// Add a TranposeNode after \p C in \p F that has the same shuffle as \p TR.
+/// This funcion assumes that the type of \p C and the output type of \p TR are
+/// the same. \returns the newly added TransposeNode.
+static TransposeNode *insertMatchingTransposeAfterConstant(Function *F,
+                                                           Constant *C,
+                                                           TransposeNode *TR) {
+  auto *CT = C->getOutput().getType();
+  auto *TRT = TR->getResult().getType();
+  DCHECK(CT->isEqual(TRT));
+
+  auto &T = C->getPayload();
+
+  // In order for a new Transpose node with the same shuffle as TR to
+  // be created at the output of the Constant, a new Constant should
+  // created that has the same type as the input to TR.
+  auto *NC = F->getParent()->createConstant(TR->getInput().getType(),
+                                            C->getName().str() + ".transposed");
+
+  // The payload of the original Constant C has the same type as the
+  // output of TR. In order to preserve correctness, this payload must
+  // be transposed using the inverse of the shuffle of TR and stored
+  // into the payload of the new Constant.
+  //
+  // Another way to think of this is that we are inserting two
+  // Transposes that are inverses of each other back to back after the original
+  // Constant. The shuffle of the second Transpose must match that of TR.
+  // In order to preserve correctness, the shuffle
+  // of the first Transpose must be the inverse of that shuffle of the
+  // second Transpose. The statement below statically computes this
+  // first Transpose.
+  T.transpose(&NC->getPayloadMutable(), invertShuffle(TR->getShuffle()));
+
+  // Create Transpose on the LHS that has the same shuffle as TR.
+  return F->createTranspose("transpose", NC, TR->getShuffle());
+}
+
 bool EmptyPass::run(Function *F, const CompilationContext &cctx) {
   return false;
 }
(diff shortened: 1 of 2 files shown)


lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -445,6 +460,70 @@ bool SinkCode::run(Function *F, const CompilationContext &cctx) {
                                     SN->getValue());
           RTR = F->createTranspose("transpose", NS, LTR->getShuffle());
           changed = true;
+        } else if (isa<Constant>(node->getNthInput(ArithmeticNode::LHSIdx)) &&
```
SplitInfinity: I need to check that the `Constant` has only one use as well, right?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -445,6 +460,70 @@ bool SinkCode::run(Function *F, const CompilationContext &cctx) {
...
+          // of the first Transpose must be the inverse of that shuffle of the
+          // second Transpose. The statement below statically computes this
+          // first Transpose.
+          T.transpose(&NC->getPayloadMutable(),
+                      invertShuffle(RTR->getShuffle()));
+
+          // Create Transpose on the LHS that has the same shuffle as RTR.
+          LTR = F->createTranspose("transpose", NC, RTR->getShuffle());
+          changed = true;
+        } else if (isa<Constant>(node->getNthInput(ArithmeticNode::RHSIdx)) &&
```
SplitInfinity: I need to check that the `Constant` has only one use as well, right?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -916,6 +916,57 @@ TEST_F(GraphOptz, sinkTransposeBelowArithmeticNodes) {
   EXPECT_EQ(F_->getNodes().size(), 3);
 }
 
+/// Check that Transpose node is sunk below arithmetic nodes when one of the
+/// operands is a Constant.
+TEST_F(GraphOptz, sinkTransposeBelowArithmeticNodesWithConstantOperand) {
```
SplitInfinity: Let's also test the case in which the `Constant` is on the RHS of the `Add`.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -445,6 +460,70 @@ bool SinkCode::run(Function *F, const CompilationContext &cctx) {
...
+          // Another way to think of this is that we are inserting two
+          // Transposes back to back between the old Constant and the Add node.
+          // In order to sink the second Transpose below the Add node, its
+          // shuffle must be equal to that of the Transpose that is the RHS
+          // operand of the Add. In order to preserve correctness, the shuffle
+          // of the first Transpose must be the inverse of that shuffle of the
+          // second Transpose. The statement below statically computes this
+          // first Transpose.
+          T.transpose(&NC->getPayloadMutable(),
+                      invertShuffle(RTR->getShuffle()));
```
opti-mix: This big snippet and a similar snippet below look almost identical. Move then into a helper function?

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
jfix71: Not checking isn't a correctness issue -- you're creating a new Constant and so other users should still have access to the un-transposed version. But there is a tradeoff here -- if it's a large Constant then we probably wouldn't want to make a duplicate. If it's a small Constant then maybe the compute tradeoff is worth the extra memory. For now (and if your use case only has one user) I'd say add the check.

SplitInfinity:
Addressed comments:
- moved the code that adds a `Transpose` after the `Constant` operand of an arithmetic node into a helper function so that it can be reused by the cases of the `if-statement` that handle the `Constant` as the RHS/LHS operand
- extended test to test that the optimization works when the `Constant` is LHS and RHS of an `Add` node

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
SplitInfinity: This is tricky thanks to the numerical correctness tests I added (hah!). The flow of a test is now like so:

1. Build graph.
2. Make pre-optimization assertions about the graph.
3. Clone the graph.
4. Optimize the cloned version of the graph.
5. Make post-optimization assertions about the optimized, cloned version of the graph.
6. Run both the original and optimized graphs with the same input and make sure the outputs are identical.

Because the graph is cloned and then its clone is optimized, these `Constants` actually have *two* uses: one in the original graph, and another in its clone.

jfix71: Hmm, yeah this is where it would be nice to have `Module::clone()` to use instead of just `Function::clone()`. Then we could copy the Constants in addition to the Function and they'd have the same number of users.

(Another place where this is a little annoying is unit tests for training, where we create two EEs/Modules: one for training and one for inference, and then we explicitly re-create the same Function twice on each. E.g. see MLTest `trainASimpleNetwork` and many others.)

nickgg:
@SplitInfinity did you push your update? I don't see that change.

SplitInfinity:
@nickgg 

Good call, I committed the changes locally but forgot to push.

PR #3622, opened 2019-10-14
[NFC] Run clang-format

SplitInfinity:
**Summary**
This commit was created by running `utils/format.sh` fix without making
any changes.

**Testing**
All tests pass.


SplitInfinity:
The `CLANG-FORMAT` CI job does not fail on master so this could be a local issue on my end. My `clang-format` version:
```
$ /tmp/clang-format --version
clang-format version 8.0.1 (tags/RELEASE_801/final)
```

SplitInfinity:
This is due to the version difference. I downgraded to `llvm7` and everything is fine.

PR #3625, opened 2019-10-15
[WIP][PyTorch] Compute output shapes for a given FunctionSchema and inputs for LazyTensor

jackm321:
Summary:
Compute output shapes for a given FunctionSchema and some inputs.
This is just a quick draft that makes a jit graph and creates a node from the FunctionSchema, probably we'll want to abstract PyTorchModelLoader a bit to avoid this.

Documentation:
comments

Test Plan:

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -2045,6 +2045,33 @@ Error PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
 }
 
 /*static*/
+Expected<std::vector<InputMeta>>
+computeOutputShapes(const c10::FunctionSchema &schema,
+                    const std::vector<c10::IValue> &inputs) {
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2044,6 +2044,33 @@ Error PyTorchModelLoader::loadConstant(const torch::jit::Node *ptNode) {
...
 }
 
+/*static*/
+Expected<std::vector<InputMeta>>
+computeOutputShapes(const c10::FunctionSchema &schema,
+                    const std::vector<c10::IValue> &inputs) {
+  torch::jit::Graph graph;
+  auto symbol = torch::jit::Symbol::fromQualString(schema.name());
+  auto node = graph.create(symbol, schema.returns().size());
+  node = graph.insertNode(node);
```
jackm321: may need to call graph.addInput() or something similar

PR #3627, opened 2019-10-16
refactor Compilation Log generation and usability

nickgg:
Summary: The Compilation Log option is very useful, but has a few rough edges. This PR attempts to improve usability by changing the log format.

The biggest change is switching the log itself to use JSON, it now looks like this (I did run the output through a json pretty printer):

image

I find this more readable personally, and it lets us remove all the regexs from the log_parser tool.

A few other QOL changes:

* Changed the command line option to `--compilation-log=filename` rather than a fixed filename based on Function name. This overwrites so currently will only log the last Function compiled in the process.
* By default, empty compilation stages are not logged, making it easier to see what was done. The `--verbose-compilation` flag  disables this and logs all compilation/optimization passes.
* Fixed a bunch of places where a Node's input was changed and we didn't log it, which broke all the log-based tooling.
* Simplified the log parser a bit.

Documentation: I still havent uploaded Compilation Log docs yet, so nothing to update.

Test Plan: ran tests & image-classifier, dumped log of compiling resnet and used the tools.

E.g.  here's the base of Resnet before and after lowering:
before:
image

after:
image

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -73,7 +73,7 @@ class Module final {
   PseudoRNG PRNG_;
 
   /// Module log context that stores all logs related to this module.
-  ModuleLogContext moduleLogCtx_;
+  LogContext moduleLogCtx_{nullptr};
 
   /// Inserts the constant \p V to the list of constants.
   Constant *addConstant(Constant *V);
@@ -221,7 +221,7 @@ public:
   uint64_t getConstantsSize();
 
   /// \Returns the module log context.
-  ModuleLogContext &getModuleLogContext() { return moduleLogCtx_; };
+  LogContext *getModuleLogContext() { return &moduleLogCtx_; };
 
   // Don't copy or move this class around.
   // The destructor will wipe the functions leaving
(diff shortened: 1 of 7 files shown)


include/glow/Graph/Graph.h
```diff
@@ -221,7 +221,7 @@ class Module final {
   uint64_t getConstantsSize();
 
   /// \Returns the module log context.
-  ModuleLogContext &getModuleLogContext() { return moduleLogCtx_; };
+  LogContext *getModuleLogContext() { return &moduleLogCtx_; };
```
opti-mix: Why is this a pointer instead of a reference now? It always points to moduleLogCtx_, right?

lib/Graph/Log.cpp
```diff
@@ -28,195 +28,273 @@ namespace glow {
...
 
   llvm::outs() << "Writing compilation log file for Module to: "
                << compileLogFilename << '\n';
   std::error_code EC;
-  llvm::raw_fd_ostream myfile(compileLogFilename, EC);
-  for (const auto &s : logContents_) {
-    myfile << s;
-  }
+  llvm::raw_fd_ostream myfile(compileLogFilename,
+                              EC); //, llvm::sys::fs::OF_Append);
```
opti-mix: Remove the comment?

include/glow/Graph/Graph.h
nickgg: The LogContext as owned by a Function is a pointer, and so it makes sense for interfaces that handle it to be a pointer as well, we could take a reference here but we'd have to convert to a ptr later anyway.

nickgg:
@jfix71 good call, totally forgot about that.

PR #3628, opened 2019-10-16
[ThirdParty] Add miniz-2.0.8 for zip operations

yinghai:
Summary:
`miniz` (https://github.com/richgel999/miniz) is a self-contained zip operation library. This specific version is taken from PyTorch and allows some fast CRC function internally. The reason to add this to pave the way to ship large models to vendors. Models with large weights cannot be exported in a single protobuf file and therefore need to be packed in a certain archive structure. I choose miniz here due to minimal dependency and potentially reuse of PyTorch code. 

Documentation:

[Optional Fixes #issue]

Test Plan:
It builds. 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,6 +177,10 @@ else()
   include_directories(${GLOW_THIRDPARTY_DIR}/fp16/include)
 endif()
 
+add_library(Miniz
+  ${GLOW_THIRDPARTY_DIR}/miniz-2.0.8/miniz.c)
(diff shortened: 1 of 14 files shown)


jfix71:
Hm, can we add this as a submodule instead of copying all the code directly into thirdparty?

yinghai:
Submodule is most straightforward way. But this is a specific version with some code changes to work with internal code. Pytorch's miniz has this weird setup too. In the end, we just use the 2 files. Maybe it's simple enough if we remove the examples?

yinghai:
That's a good idea, like hosting this in a repo somewhere. I can check with PyTorch folks and see.

PR #3630, opened 2019-10-16
[GraphOptz] Add numeric equivalence check to GraphOptz.optimizeBatchNormAfterConv

bmcdb:
Summary: Added numeric equivalence test to the GraphOptz.optimizeBatchNormAfterConv test case.

Test Plan: ninja test: 100% tests passed, 0 tests failed out of 47

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -193,7 +193,7 @@ TEST_F(GraphOptz, liveCodeNotEliminated) {
 }
 
 TEST_F(GraphOptz, optimizeBatchNormAfterConv) {
-  Node *A =
+  auto *A =
       mod_.createPlaceholder(ElemKind::FloatTy, {1, 10, 20, 3}, "A", false);
   Node *CV = F_->createConv(bindings_, "conv", A, 16, 5, 1, 2, 1);
   Node *BN =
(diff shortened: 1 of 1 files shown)


tests/unittests/GraphOptzTest.cpp
```diff
@@ -203,11 +203,15 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConv) {
...
+  EXPECT_EQ(optimizedF_->getNodes().size(), 2);
 
-  ASSERT_EQ(A->getNumUsers(), 1);
-  Node *newCV = A->getUsers().begin()->getUser();
+  bindings_.allocate(mod_.getPlaceholders());
+  bindings_.get(A)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  checkNumericalEquivalence();
+
+  ASSERT_EQ(A->getNumUsers(), 2);
+  Node *newCV = std::next(A->getUsers().begin(), 1)->getUser();
```
SplitInfinity: Is the iteration order guaranteed?  Maybe it would be more futureproof to loop through the nodes in `optimizedF_` and find the `ConvolutionNode` that is `!= CV`?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -203,11 +203,15 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConv) {
...
-  ::glow::optimize(F_, CompilationMode::Infer);
-  EXPECT_EQ(F_->getNodes().size(), 2);
+  optimizedF_ = optimizeFunction(F_);
+  EXPECT_EQ(optimizedF_->getNodes().size(), 2);
 
-  ASSERT_EQ(A->getNumUsers(), 1);
-  Node *newCV = A->getUsers().begin()->getUser();
+  bindings_.allocate(mod_.getPlaceholders());
+  bindings_.get(A)->getHandle().randomize(-1.0, 1.0, mod_.getPRNG());
+  checkNumericalEquivalence();
```
SplitInfinity: Nit: I think the rest of these tests make post-optimization asserts on the graph first and then perform this numerical equivalence check. Let's move this after the asserts below for the sake of consistency.

bmcdb: Done.

tests/unittests/GraphOptzTest.cpp
bmcdb: Done, but I'm not fully familiar with these objects, so there might be a cleaner way to do this with the relationship functions.

SplitInfinity:
Thanks for addressing all of my comments. It looks good now. Could you squash your commits into one with an appropriate title and message?

bmcdb:
Is this a flaky test? Should I try re-pushing to force another check?

PR #3635, opened 2019-10-16
[PyTorch]New ptTensorToGlowTensor

zrphercule:
Now we accept a quantized tensor as fused node input.
test is in test_quantized_conv2d_nonfunctional, however it is now not enbaled since it could be very flaky because of accuracy.

### DIFF
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -107,8 +107,10 @@ glow::ElemKind scalarTypeToElemKind(c10::ScalarType ty) {
     return ElemKind::Int64ITy;
   } else if (ty == at::kBool) {
     return ElemKind::BoolTy;
+  } else if (ty == at::kQInt8) {
+    return ElemKind::Int8QTy;
   } else {
-    LOG(DFATAL) << "Not supported yet.";
+    LOG(DFATAL) << ty << " Not supported yet.";
     return ElemKind::Int64ITy;
   }
 }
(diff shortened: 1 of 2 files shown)


torch_glow/src/PyTorchCommon.cpp
```diff
@@ -107,8 +107,10 @@ glow::ElemKind scalarTypeToElemKind(c10::ScalarType ty) {
     return ElemKind::Int64ITy;
   } else if (ty == at::kBool) {
     return ElemKind::BoolTy;
+  } else if (ty == at::kQInt8) {
```
jackm321: should we just add the uint8 and int32 while we're at it?

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -227,7 +216,26 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
...
+  std::vector<size_t> dims;
+  for (const auto &size : concreteSizes) {
+    dims.push_back(static_cast<size_t>(size));
+  }
+
+  auto scalarType = ptType.scalarType().value();
+  auto glowElemKind = scalarTypeToElemKind(scalarType);
+
+  glow::Type glowType;
+  if (ptTensor.is_quantized()) {
```
jackm321: We should probably check that it's not channel wise quantized right?

jackm321:
What is the flakiness of the test? Can we fix it?

torch_glow/src/PyTorchCommon.cpp
zrphercule: yeah agree.

torch_glow/src/PyTorchCommon.cpp
zrphercule: agree, let me take a look.

zrphercule:
> No tests?

tests are actually skipped here: https://github.com/pytorch/glow/pull/3567

jackm321:
@zrphercule let's create a test for this that doesn't use random weights so that we can have an enabled test for this

PR #3636, opened 2019-10-17
[PyTorch] Add support for aten::pow operator

dreamingleo:
Summary:
Add support for aten::pow operator

Test Plan:
New unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -437,6 +437,14 @@ struct MaxPoolInputs {
   };
 };
 
+/// Indices of aten::Pow inputs.
+struct PowInputs {
+  enum {
+    input = 0,
+    exponent = 1,
+  };
+};
+
 /// Indices of aten::clamp inputs.
 struct ClampInputs {
   enum {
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -571,6 +571,7 @@ PyTorchModelLoader::getSymbolsMapping() {
        {{"aten::min"}, &PyTorchModelLoader::loadMin, {}},
        {{"aten::max"}, &PyTorchModelLoader::loadMax, {}},
        {{"aten::exp"}, &PyTorchModelLoader::loadExp, {}},
+       {{"aten::pow"}, &PyTorchModelLoader::loadPow, {}},
```
jackm321: Add input 1 as a thing to be frozen

torch_glow/tests/nodes/pow_test.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+import torch_glow
```
jackm321: unused

PR #3642, opened 2019-10-18
[GraphOptz] Add numeric equivalence check to GraphOptz.optimizeBatchNormAfterConv…

bandkenfbcom:
Summary:
Added numeric equivalence test to the GraphOptz.optimizeBatchNormAfterConvFP16 test case.

Documentation:
none

Test Plan:
$ ninja all && ninja test

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -264,7 +264,7 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvMultiple) {
 }
 
 TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
-  Node *A =
+  auto *A =
       mod_.createPlaceholder(ElemKind::Float16Ty, {1, 10, 20, 3}, "A", false);
   Node *CV = F_->createConv(bindings_, "conv", A, 16, 5, 1, 2, 1);
   Node *BN =
@@ -273,16 +273,30 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
 
   EXPECT_EQ(F_->getNodes().size(), 3);
 
+  optimizedF_ = F_->clone(F_->getName().str() + "_optimized");
   ::glow::convertPlaceholdersToConstants(F_, bindings_, {});
-  ::glow::optimize(F_, CompilationMode::Infer);
-  EXPECT_EQ(F_->getNodes().size(), 2);
+  ::glow::convertPlaceholdersToConstants(optimizedF_, bindings_, {});
+  ::glow::optimize(optimizedF_, CompilationMode::Infer);
 
-  ASSERT_EQ(A->getNumUsers(), 1);
-  Node *newCV = A->getUsers().begin()->getUser();
-  EXPECT_TRUE(llvm::isa<ConvolutionNode>(newCV));
-  ASSERT_EQ(newCV->getNumUsers(), 1);
-  Node *save = newCV->getUsers().begin()->getUser();
-  EXPECT_TRUE(llvm::isa<SaveNode>(save));
+  EXPECT_EQ(optimizedF_->getNodes().size(), 2);
+
+  ASSERT_EQ(A->getNumUsers(), 2);
+
+  bool optimizedPathExists{false};
+  for (const auto &path : A->getUsers()) {
+    auto cv = path.getUser();
+    EXPECT_TRUE(llvm::isa<ConvolutionNode>(cv));
+    ASSERT_EQ(cv->getNumUsers(), 1);
+    auto next = cv->getUsers().begin()->getUser();
+    optimizedPathExists |= llvm::isa<SaveNode>(next);
+  }
+
+  EXPECT_TRUE(optimizedPathExists);
+
+  bindings_.allocate(A)->getHandle<float16_t>().randomize(-1.0, 1.0,
+                                                          mod_.getPRNG());
+
+  checkNumericalEquivalence();
 }
 
 /// Check that transpose constant folding is done before BatchNorm optimization,


tests/unittests/GraphOptzTest.cpp
```diff
@@ -283,6 +283,9 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
   ASSERT_EQ(newCV->getNumUsers(), 1);
   Node *save = newCV->getUsers().begin()->getUser();
   EXPECT_TRUE(llvm::isa<SaveNode>(save));
+
+  optimizedF_ = optimizeFunction(F_);
```
SplitInfinity: The numerical equivalence check feeds the same input to both `F_` and `optimizedF_`. `F_` at this point is already optimized due to line 277, so the numerical equivalence check is not useful. You should not optimize `F_` - this is where the `optimizeFunction` helper comes into play. It clones the function first, and then optimizes the clone.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -283,6 +283,9 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  checkNumericalEquivalence();
```
SplitInfinity: You need to provide inputs to the graph. before calling `checkNumericalEquivalence`. See `eliminateNoopTile` for an example.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -273,16 +273,35 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  EXPECT_TRUE(llvm::isa<SaveNode>(oldSave));
+
+  usersIt++;
+  Node *newCV = usersIt->getUser();
   EXPECT_TRUE(llvm::isa<ConvolutionNode>(newCV));
   ASSERT_EQ(newCV->getNumUsers(), 1);
-  Node *save = newCV->getUsers().begin()->getUser();
-  EXPECT_TRUE(llvm::isa<SaveNode>(save));
+  Node *newSave = newCV->getUsers().begin()->getUser();
+  EXPECT_TRUE(llvm::isa<SaveNode>(newSave));
```
bandkenfbcom: Is this part necessary?  This sort of shows that the BN node is folding into the CV node, but it is very dependent on the internal implementation.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -273,16 +273,35 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  ASSERT_EQ(A->getNumUsers(), 2);
+  auto usersIt = A->getUsers().begin();
+  Node *oldCV = usersIt->getUser();
+  EXPECT_TRUE(llvm::isa<ConvolutionNode>(oldCV));
+  ASSERT_EQ(oldCV->getNumUsers(), 1);
+  Node *oldBN = oldCV->getUsers().begin()->getUser();
+  EXPECT_TRUE(llvm::isa<BatchNormalizationNode>(oldBN));
+  ASSERT_EQ(oldBN->getNumUsers(), 1);
+  Node *oldSave = oldBN->getUsers().begin()->getUser();
+  EXPECT_TRUE(llvm::isa<SaveNode>(oldSave));
```
SplitInfinity: I think you can get rid of all this and skip to the second user of `A` directly.

tests/unittests/GraphOptzTest.cpp
SplitInfinity: Well, that is the whole point of the optimization that is being tested. :)

tests/unittests/GraphOptzTest.cpp
```diff
@@ -273,16 +273,27 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+  ::glow::optimize(optimizedF_, CompilationMode::Infer);
 
-  ASSERT_EQ(A->getNumUsers(), 1);
-  Node *newCV = A->getUsers().begin()->getUser();
+  EXPECT_EQ(optimizedF_->getNodes().size(), 2);
+
+  ASSERT_EQ(A->getNumUsers(), 2);
+
+  // the optimized path is the second one from A
+  auto usersIt = ++(A->getUsers().begin());
```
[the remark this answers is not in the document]
bandkenfbcom: I'm not sure what you mean.  It should check to make sure that it is constructed in an optimized form A->CV->save.  Do you mean in addition to the current check, like:

```
auto usersIt = A->getUsers().begin();
Node *origCV = usersIt->getUser();
EXPECT_TRUE(llvm::isa<ConvolutionNode>(origCV));
ASSERT_EQ(origCV->getNumUsers(), 1);
Node *origBN = origCV->getUsers().begin()->getUser();
EXPECT_TRUE(llvm::isa<BatchNormalizationNode>(origBN));
ASSERT_EQ(origBN->getNumUsers(), 1);
Node *origSave = origBN->getUsers().begin()->getUser();
EXPECT_TRUE(llvm::isa<SaveNode>(origSave));
```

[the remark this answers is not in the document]
SplitInfinity: No, I mean that you are assuming a certain iteration order for the users of A.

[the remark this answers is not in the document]
bandkenfbcom: OK, I think I get it.  Please look at the new review.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -273,16 +273,42 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvFP16) {
...
+      ASSERT_EQ(next->getNumUsers(), 1);
+      auto origSave = next->getUsers().begin()->getUser();
+      EXPECT_TRUE(llvm::isa<SaveNode>(origSave));
+      EXPECT_FALSE(originalPathExists);
+      originalPathExists = true;
+    }
+  }
+
+  EXPECT_TRUE(originalPathExists);
+  EXPECT_TRUE(optimizedPathExists);
```
SplitInfinity: You don't need to test the the original path exists. Just testing for the optimized path is fine.

bandkenfbcom: Got it!

PR #3644, opened 2019-10-18
[OpenCL] Refactor buffer copies into DeviceManager

nickgg:
Summary: As a preparation for supporting device resident Tensors, move Tensor copies out of the OCLCompiledFunction and into the DeviceManager. Mostly this is just a move of the relevant code into the new class, but I had to move the KernelLaunches to the OCLDeviceBindings.

Fixed a few small TraceEventTest issues as well.

Documentation: NFC

Test Plan: ran tests, image-classifier, and resnet-runtime with OCL

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.h
+++ b/lib/Backends/OpenCL/OpenCL.h
@@ -16,8 +16,6 @@
 #ifndef GLOW_BACKENDS_OPENCL_OPENCL_H
 #define GLOW_BACKENDS_OPENCL_OPENCL_H
 
-#include "OpenCLDeviceManager.h"
-
 #include "glow/Backend/Backend.h"
 #include "glow/Backend/BackendUtils.h"
 #include "glow/Backend/CompiledFunction.h"
@@ -73,6 +71,9 @@ static void addIntOption(std::vector<std::string> &options,
   options.push_back("-D" + name + "=" + std::to_string(value));
 }
 
+using ManualEventMap =
+    std::map<std::string, std::pair<Placeholder *, const TraceInfo::Event *>>;
+
 /// A Glow IR function compiled for OpenCL.
 class OpenCLFunction final : public CompiledFunction {
   /// A helper type representing a key for the program's cache.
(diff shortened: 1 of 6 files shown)


lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -412,6 +416,225 @@ void OpenCLDeviceManager::returnRunCommandQueue(OpenCLCommandQueue &queue) {
...
+void OpenCLDeviceManager::copyInputsToDevice(
+    const RuntimeBundle &runtimeBundle, ExecutionContext *context,
+    runtime::OpenCLDeviceBindings *devBindings) {
+  TRACE_EVENT_SCOPE(context->getTraceContext(), TraceLevel::RUNTIME,
+                    "copyInputsToDevice");
+  bool profilingEnabled =
+      context->getTraceContext() &&
+      (context->getTraceContext()->getTraceLevel() & TraceLevel::COPY);
+  size_t sizeInBytes = runtimeBundle.getConstantWeightSize();
+  if (runtimeBundle.getConstants()) {
```
gcatron: I think we can get rid of this section. The constants should be coped to the device during addNetwork. I would guess this branch is never entered...

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -16,8 +16,6 @@
 #ifndef GLOW_BACKENDS_OPENCL_OPENCL_H
 #define GLOW_BACKENDS_OPENCL_OPENCL_H
 
-#include "OpenCLDeviceManager.h"
```
mortzur: This change in "include" direction actually makes a lot of sense to me. Just wanted to note that CPU / Habana backends header files include the CPU / Habana device manager header (correspondingly, both cases only because of createDeviceManager(...) implemented in the header). This seems like the correct direction.

nickgg: Yeah we should just fix this everywhere I think. (not in this PR!)

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
nickgg: Good catch this is totally redundant.

nickgg:
The OpenCL test fail looks like the usual pocl issue, not related to this diff.

opti-mix:
Nice!

PR #3645, opened 2019-10-18
[Tracing] Add thread names for Interpreter, CPU and OCL

nickgg:
Summary: Adding thread names to glow Traces for the third time, hopefully this time it sticks. I've also modified the OpenCL DeviceManager so events that are generated on the device us a different thread than those on the host, which I think is easier to read.

Traces look like this (from image-classifier):
image
image

With multiple devices they look like this (from resnet-runtime):
image

Documentation: N/A

Test Plan: unit tests, image-classifier & resnet-runtime


nickgg:
rebased

PR #3646, opened 2019-10-20
[Glow] Add numerical equivalence check to GraphOptz.optimizeBatchNorm…

tongning-fb:
Summary: 
The task is to add a numerical equivalence check to this test case which checks that both the original and optimized graph return the same result for the same input.

Documentation:

[Optional Fixes #issue]

Test Plan:  
./tests/GraphOptzTest

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -345,7 +345,7 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvWithReshapeConst) {
 /// Check that the batch normalization optimization is
 /// not blocked by predicates and that it preserves them.
 TEST_F(GraphOptz, optimizeBatchNormAfterConvWithPred) {
-  Node *A =
+  auto *A =
       mod_.createPlaceholder(ElemKind::FloatTy, {1, 10, 20, 3}, "A", false);
   Node *pred1 =
       mod_.createPlaceholder(ElemKind::FloatTy, {1}, "predicate", false);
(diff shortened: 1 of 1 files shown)


tests/unittests/GraphOptzTest.cpp
```diff
@@ -372,6 +372,11 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvWithPred) {
   ASSERT_EQ(newCV->getNumUsers(), 1);
   Node *save = newCV->getUsers().begin()->getUser();
   EXPECT_TRUE(llvm::isa<SaveNode>(save));
+
+  optimizedF_ = optimizeFunction(F_);
```
SplitInfinity: The numerical equivalence check feeds the same input to both `F_` and `optimizedF_`. `F_` at this point is already optimized due to line 277, so the numerical equivalence check is not useful. You should not optimize `F_` - this is where the `optimizeFunction` helper comes into play. It clones the function first, and then optimizes the clone.

tongning-fb: Thank you for your clarification! I tried to change my code according to your advice, however, I still encounter one failure in line 374: ASSERT_TRUE(newCV->hasPredicate());
At your convenience, would you mind giving me some hint? The attached image is the modified version of my code.
Screen Shot 2019-10-22 at 9 57 51 PM

SplitInfinity: That's weird. I recommend printing out the graph using `Function::dumpDAG` and looking at it with `graphviz` to see what the cloned graph looks like.

PR #3647, opened 2019-10-21
[PyTorch] Add support for aten::slice operator

qizzzh:
Summary:
Add support for aten::slice operator

Test:
New unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -517,6 +517,17 @@ struct PReluInputs {
   };
 };
 
+/// Indexes of aten::slice inputs.
+struct SliceInputs {
+  enum {
+    input = 0,
+    dim = 1,
+    start = 2,
+    end = 3,
+    step = 4,
+  };
+};
+
 /// Indexes of aten::softmax inputs.
 struct SoftMaxInputs {
   enum {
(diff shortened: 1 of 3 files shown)


torch_glow/tests/nodes/slice_test.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+from tests.utils import jitVsGlow
+
+
+def test_slice_basic():
```
zrphercule: I wonder if you have the generated ir of this test? Since if operator of a (sub)graph is less than 3 ops, then nothing will be fused, fuser will just ignore it (to prevent a lot of 1-node lower graph in models)

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1793,6 +1812,43 @@ Error PyTorchModelLoader::loadPRelu(const torch::jit::Node *ptNode) {
...
+      dim, iValToInt(getGlowIValueForValue(inputs[SliceInputs::dim])));
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      start, iValToInt(getGlowIValueForValue(inputs[SliceInputs::start])));
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      end, iValToInt(getGlowIValueForValue(inputs[SliceInputs::end])));
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      step, iValToInt(getGlowIValueForValue(inputs[SliceInputs::step])));
+
+  RETURN_ERR_IF_NOT(step == 1, "loadSlice only supports step == 1");
+  int dims_size = input.dims().size();
```
jackm321: dimsSize style

torch_glow/tests/nodes/slice_test.py
qizzzh: I didn't check the generated ir. But jitVsGlow checks whether a glowFusion node exists and whether the expected op shows up in the fusion node, so I think it should work.

zrphercule: That is weird... Could you please take a look at the ir (before lower & after lower) ? you can output them in jitVsGlow by print(traced_model.graph())

PR #3651, opened 2019-10-21
[PyTorch] Add torch.mean support

zrphercule:
pytest tests/nodes/mean_test.py

This op is needed in resnet.


torch_glow/src/PyTorchModelLoader.h
```diff
@@ -359,6 +359,10 @@ class PyTorchModelLoader {
   /// \returns error on failure.
   Error loadTranspose(const torch::jit::Node *ptNode);
 
+  /// Load a PyTorch t (transpose) node.
```
jackm321: mean

PR #3653, opened 2019-10-21
[PyTorch] Add skipped tests for accuracy debuggings

zrphercule:
Add a few skipped tests which caused by glow&pytorch accuracy for debugging.
cc @hyuen


torch_glow/tests/nodes/quantized_conv2d_test.py
```diff
@@ -38,7 +38,7 @@ def test_f(a, w, b):
                                                         "aten::dequantize"})
 
 
-@pytest.mark.skip(reason="accuracy between glow&ytorch")
+@pytest.mark.skip(reason="accuracy between glow & pytorch")
```
jackm321: continuing the conversation from https://github.com/pytorch/glow/pull/3567#discussion_r336232328
This isn't necessarily an accuracy issue. It is initializing new conv parameter each time it runs so it's not going to be deterministic. It doesn't make sense as a test the way it is now.

zrphercule: Yes I agree. I think I can give a determined params for testing.

zrphercule:
@jackm321 How do you like this test now?

The thing in conv is more complex, only set the bias to 0 doesnt work perfectly (although increases the accuracy significantly...) Therefore I will leave conv's test untouched.

cc @hyuen

PR #3654, opened 2019-10-21
[OperatorTest] Unbreak build due to ambiguous call to floor

jfix71:
Build is breaking on Mac (seems fine on CI):

```
glow/tests/unittests/OperatorTest.cpp:6101:28: fatal error: call to 'floor' is ambiguous 
  uint32_t bin = (uint32_t)floor(bin_calc);
                           ^~~~~ 
```


hyuen:
thanks

PR #3659, opened 2019-10-22
[PyTorch] Load aten::bmm node

jackm321:
Summary:
Load batch matmul.

Documentation:
doxygen

Test Plan:
new unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -702,6 +702,7 @@ PyTorchModelLoader::getSymbolsMapping() {
         }},
        {{"aten::matmul"}, &PyTorchModelLoader::loadMatMul, {}},
        {{"aten::mm"}, &PyTorchModelLoader::loadMM, {}},
+       {{"aten::bmm"}, &PyTorchModelLoader::loadBmm, {}},
        {{"aten::addmm"},
         &PyTorchModelLoader::loadAddMM,
         {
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1976,6 +1977,29 @@ Error PyTorchModelLoader::loadMM(const torch::jit::Node *ptNode) {
...
+  auto outputs = ptNode->outputs();
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, 2, outputs, 1));
+
+  glow::NodeValue lhs;
+  ASSIGN_VALUE_OR_RETURN_ERR(lhs, getGlowNodeValueForValue(inputs[0]));
+  glow::NodeValue rhs;
+  ASSIGN_VALUE_OR_RETURN_ERR(rhs, getGlowNodeValueForValue(inputs[1]));
+
+  // Check dimensions of inputs
+  if (lhs.dims().size() != 3 || rhs.dims().size() != 3) {
```
qizzzh: I think RHS could be either 2 or 3 dimensional. createBatchMatMul works for both. Otherwise looks good to me!

jackm321: I'm going off the documentation here https://pytorch.org/docs/stable/torch.html#torch.bmm, does that seem right?

qizzzh: aten::bmm works for both. Looks like the pt op imposes an additional constraint. Since we are dealing with only pt models we should be good.

PR #3662, opened 2019-10-22
Load pt tanh node

jackm321:
Summary:
Load `aten::tanh` and inplace version `aten::tanh_`.

Documentation:
doxygen

Test Plan:
new unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -612,6 +612,7 @@ PyTorchModelLoader::getSymbolsMapping() {
         &PyTorchModelLoader::loadSigmoid,
         {}},
        {{"aten::relu", "aten::relu_"}, &PyTorchModelLoader::loadRelu, {}},
+       {{"aten::tanh", "aten::tanh_"}, &PyTorchModelLoader::loadTanh, {}},
        {{"aten::t", "aten::t_"}, &PyTorchModelLoader::loadTranspose, {}},
        {{"aten::min"}, &PyTorchModelLoader::loadMin, {}},
        {{"aten::max"}, &PyTorchModelLoader::loadMax, {}},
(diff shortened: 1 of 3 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1259,6 +1260,19 @@ Error PyTorchModelLoader::loadRelu(const torch::jit::Node *ptNode) {
...
 }
 
+Error PyTorchModelLoader::loadTanh(const torch::jit::Node *ptNode) {
+  auto inputs = ptNode->inputs();
+  auto outputs = ptNode->outputs();
+  RETURN_IF_ERR(checkInputAndOutputSizes(inputs, 1, outputs, 1));
+
+  glow::NodeValue input;
+  ASSIGN_VALUE_OR_RETURN_ERR(input, getGlowNodeValueForValue(inputs[0]));
+  input = rescaleUIntToInt(input);
```
yinghai: Why do we do this?

jackm321: Oh it's for quantized inputs, I copied it accidentally grabbed it from relu. I'll remove it and if someone needs it for quantized op they can add it later

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1259,6 +1260,19 @@ Error PyTorchModelLoader::loadRelu(const torch::jit::Node *ptNode) {
...
+
+  glow::TanhNode *glowNode = F_.createTanh("tanh", input);
+  return addValueMapping(outputs[0], rescaleIntToUint(glowNode->getResult()));
```
yinghai: Yeah, also here.

jackm321:
@yinghai what do you mean? If it's the uint8 requantization thing, it only runs on quantized inputs

yinghai:
@jackm321  land this?

PR #3668, opened 2019-10-23
add configurable alignment to MemoryAllocator

nickgg:
Summary: The MemoryAllocator supports alignment but is hard coded in the source, this adds a constructor argument setting the alignment so we can use different alignments in different backends of the same process.

Documentation: comments only

[Optional Fixes #issue]

Test Plan: unit tests, particularly new test


opti-mix:
Makes a lot of sense! LGTM

PR #3669, opened 2019-10-23
adjust the unit test for sigmoid sweep

hyuen:
Summary: Turns out the DSP device does handle denorms, so extending the sweep to -20,20 and removing the artificially zeroing of denorm values.

Differential Revision: D18087419


tests/unittests/OperatorTest.cpp
```diff
@@ -6037,11 +6037,8 @@ COMPARE_UNARY_OP_FUN(Sigmoid, 10, -10.0F, 10.0F)
 /// Reference ideal sigmoid implementation. Computes an fp32 sigmoid
 /// and casts the result to FP16, no denorms
```
jfix71: nit: Can you update the comment and add a period?

PR #3671, opened 2019-10-23
[DeviceResidentTensors] Implementing device resident tensors in OpenCL

mortzur:
Summary:
Device resident tensors API implementation with OpenCL backend.

* DeviceResidencyInfo - the class which stores residency info. 
  Tensor holds a shared_ptr to a default DRI object in order to be able to share the pointer on tensor "getUnowned" calls before any device transfers take place. Calls to "clone" produce an independent copy with a different DRI object. 

* DeviceTensorTransferManager - an interface of the transfer operations related to DRT. 

* Tensor: asserts were added on accessing host tensor data to ensure host manipulations and device residency are mutually exclusive.

* Functionality: OpenCL backend now supports DRT and leaves tensors on the device by default. This requires the user to transfer tensors to host to access data.
  - HostManager - implements the method "ensureOutputsAvailable()". Calling this after runNetworks ensures all tensors are on host.
  - Execution Engine - calling EE::run() uses hostManager and calls ensureOutputsAvailable().
  - DeviceManager based testing - "copyDeviceTensorsToHost" field was added to DeviceConfig to enable runFunction to transfer all tensors to host when complete. 

* OpenCL: in order to do device transfers a base address (cl_mem) and an offset are required. Both were wrapped in a struct OpenCLDeviceTransferContext. 
This should be eliminated if/when we change openCL compiled function to direct pointer calls (instead of base+offset pointer arithmetic, see below).    

Previous notes on openCL function:
OpenCL compiled function and kernels rely on pointer arithmetic in the contiguous memory section (defined by the runtime bundle). This doesn't fit with dynamic memory allocation for device tensor buffers. (Ongoing discussion with @nickgg ) This may be addressed multiple ways, for example:
    - Modify openCL function and kernels to support absolute addresses (instead of base+offset scheme). This enables dynamic memory allocation and also different memory management schemes (allocating per-tensor ahead-of time).
    - Blocking the entire contiguous memory section until all tensors are released (while doing the required ref-counting) 

Currently solved by "buffering" : copy to a separate device buffer when the function completes and free the function buffer.

Test Plan:
ninja test

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -152,6 +152,27 @@ public:
   /// \returns the DeviceInfo for this device containing peak limits for
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
+
+  /// Copies the contents of \p tensor from the host to the \p location address
+  /// on this device. Updates the tensor residency info.
+  virtual bool transferToDevice(Tensor &tensor, void *location) {
+    DCHECK("Not Implemented");
+    return false;
+  }
+
+  /// Copies the device buffer associated with \p tensor to the host.
+  /// The tensor must be resident on this device. If \p release is true, frees
+  /// the device memory. Updates the tensor residency info.
+  virtual bool transferFromDevice(Tensor &tensor, bool release = true) {
+    DCHECK("Not Implemented");
+    return false;
+  }
+
+  /// Releases the device buffer associated with \p tensor.
+  virtual bool releaseDeviceTensor(Tensor &tensor) {
+    DCHECK("Not Implemented");
+    return false;
+  }
 };
 
 } // namespace runtime
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_DEVICEMANAGER_H
 
 #include "glow/Backend/CompiledFunction.h"
+#include "glow/Base/DeviceTensorTransferManager.h"
 #include "glow/ExecutionContext/ExecutionContext.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Runtime/RuntimeTypes.h"
@@ -42,7 +43,7 @@ using ReadyCBTy = std::function<void(const Module *, Error)>;
 using FunctionMapTy = std::map<std::string, CompiledFunction *>;
 
 /// Interface managing a specific instance of a device.
-class DeviceManager {
+class DeviceManager : public DeviceTensorTransferManager {
 protected:
   /// Configuration object for the device.
   DeviceConfig config_;
@@ -153,23 +154,23 @@ public:
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
 
-  /// Copies the contents of \p tensor from the host to the \p location address
-  /// on this device. Updates the tensor residency info.
-  virtual bool transferToDevice(Tensor &tensor, void *location) {
+  /// Copies the contents of \p tensor from the host to the \p location
+  /// address on this device. Updates the tensor residency info.
+  virtual bool transferToDevice(Tensor &tensor, void *locationContext) {
     DCHECK("Not Implemented");
     return false;
   }
 
   /// Copies the device buffer associated with \p tensor to the host.
-  /// The tensor must be resident on this device. If \p release is true, frees
-  /// the device memory. Updates the tensor residency info.
+  /// The tensor must be resident on this device. If \p release is true,
+  /// frees the device memory. Updates the tensor residency info.
   virtual bool transferFromDevice(Tensor &tensor, bool release = true) {
     DCHECK("Not Implemented");
     return false;
   }
 
   /// Releases the device buffer associated with \p tensor.
-  virtual bool releaseDeviceTensor(Tensor &tensor) {
+  virtual bool releaseDeviceTensor(void *locationContext) {
     DCHECK("Not Implemented");
     return false;
   }
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -20,6 +20,7 @@
 #include <cassert>
 #include <vector>
 
+#include "glow/Base/DeviceTensorTransferManager.h"
 #include "glow/Base/Type.h"
 #include "glow/Support/Compiler.h"
 #include "glow/Support/Memory.h"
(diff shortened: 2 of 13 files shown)


include/glow/Base/Tensor.h
```diff
@@ -82,6 +123,33 @@ class Tensor final {
   char *getData() const { return data_; }
 
 public:
+  /// \returns true if tensor data is stored on a device
+  bool isDeviceResident() const { return residencyInfoP_->isDeviceResident(); }
+
+  // Update device residency info with new device manager and context
+  void moveToDevice(runtime::DeviceManager *deviceManager, void *context) {
```
nickgg: Question for a future diff: To support pinned Tensors, should we have an optional argument here to control whether tensorResidency_ is Device or Host with DeviceManager state?

include/glow/Base/Tensor.h
```diff
@@ -517,6 +587,7 @@ class Tensor final {
     reset(t);
     size_t bufferSize = type_.getSizeInBytes();
     std::copy(&t->getData()[0], &t->getData()[bufferSize], getData());
+    residencyInfoP_ = t->residencyInfoP_;
```
nickgg: Should assign work if the Tensor is device resident? It's essentially a clone.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -434,24 +434,51 @@ void OpenCLDeviceManager::copyInputsToDevice(
...
+          std::string name = ("(D2D) " + PH.first->getName()).str();
+          devBindings->kernelLaunches.emplace_back(name, "copy", event);
+        }
+        clFinish(devBindings->commandQueue);
+        // release temporary cl_mem buffer
+        releaseDeviceTensor(*tensor);
+        tensor->moveToDevice(
+            (DeviceManager *)this,
+            new OpenCLDeviceTransferContext(devBindings->deviceBuffer, addr));
+        delete ctx;
```
nickgg: This is a bit confusing, if it's already on the right device why do we have to do anything?

nickgg: Oh is this to handle the contiguous memory ranges?

include/glow/Base/Tensor.h
```diff
@@ -48,6 +48,43 @@ void genericTranspose(const Tensor *src, Tensor *dest,
...
+    deviceManager = nullptr;
+    context = nullptr;
+    tensorResidency_ = TensorResidency::Host;
+  }
+
+public:
+  DeviceResidencyInfo()
+      : tensorResidency_(TensorResidency::Host), deviceManager(nullptr),
+        context(nullptr) {}
+
```
nickgg: I think you should have a destructor here which calls deviceManager->releaseTensor if the tensor is resident.

include/glow/Backends/DeviceManager.h
```diff
@@ -152,6 +152,27 @@ class DeviceManager {
   /// \returns the DeviceInfo for this device containing peak limits for
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
+
+  /// Copies the contents of \p tensor from the host to the \p location address
+  /// on this device. Updates the tensor residency info.
+  virtual bool transferToDevice(Tensor &tensor, void *context) {
```
gcatron: This updates the tensor residency info to the information in void *context right? 
Should context be named residencyInfo or location?
Where does the location information come from? Should there be a getAddress() method as well?

nickgg: Personally, I think it should be called location and should definitely be optional.

Don't think getAddress() works because some devices may not use addresses.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
mortzur: Yes - this is the temporary buffering to workaround compiled function contiguous memory buffer.
Refactoring this "buffering" code (in both copyInputs, copyOutputs) to a helper function.

include/glow/Base/Tensor.h
mortzur: Yes!
What we want to do is actually **give DM an option to release / put-back-in-the-pool** the memory buffer. 

I think we should change the API to releaseDeviceTensor(void *context). 
What's needed in order to free the buffer is the context. Tensor may be used to get the context from DRI. If we want DRI DTOR to call "release" it can pass the stored context.
Is there another motivation to have Tensor& as the input arg?

include/glow/Backends/DeviceManager.h
mortzur: Device Residency Info contains pointers to:
1) DeviceManager
2) Opaque context
So it is not the residencyInfo itself.
In addition, I think there may be use cases where the context is useful to store additional information besides the actual memory address / location. 
For example, let's consider a backend with a memory pool for tensors. Its device manager should have a mapping between keys and buffers. If a tensor becomes device resident, it should get a buffer. In this case, one option is to store that key in the context.

Naming - how about locationContext?

GetAddress(...) : this goes to how a certain device manager creates a new locationContext.
The challenge here is to figure out if there is a single interface that fits most backends, i.e. if it makes sense to add createLocationContext() (taking no input arguments). Or does a certain backend needs the function name? or the placeholder name? (as we briefly started discussing :))

nickgg: I'd be concerned that if we find an approach that works well for all backends we know about now, we'll encounter a future backend that doesn't fit that scheme.

lib/ExecutionEngine/ExecutionEngine.cpp
```diff
@@ -114,20 +114,23 @@ void glow::updateInputPlaceholdersByName(PlaceholderBindings &bindings,
...
+                           [&runPromise, &runErr, &contextOut](
+                               runtime::RunIdentifierTy, Error err,
+                               std::unique_ptr<ExecutionContext> contextPtr) {
+                             contextOut = std::move(contextPtr);
+                             runErr = std::move(err);
+                             runPromise.set_value();
+                           });
 
   fut.wait();
+  hostManager_->ensureOutputsAvailable(*contextOut.get());
```
nickgg: It would be nice to be able to control this behaviour somehow so if we're using the EE but we want to leave Tensors of the device we could.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -480,6 +480,15 @@ void HostManager::updateExecutionStats(
                                 1000000 / duration);
 }
 
+void HostManager::ensureOutputsAvailable(ExecutionContext &context) {
```
nickgg: It's important to make sure we're calling this everywhere we use the HostManager until the user is able to deal with Tensors staying Device Resident. The big one I can think of is Onnxifi, but there may be others.

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -701,3 +731,105 @@ void OpenCLDeviceManager::runFunctionImpl(
...
+  if (releaseBuffer) {
+    // release temporary cl_mem buffer
+    releaseDeviceTensor((void *)sourceTransferContext);
+  }
+  tensor.moveToDevice((DeviceManager *)this, targetTransferContext);
+  return true;
+}
+
+/// Releases the device buffer associated with \p tensor.
+bool OpenCLDeviceManager::releaseDeviceTensor(void *locationContext) {
```
nickgg: I think we spoke about it in person, but I think it would reduce error potential if this took in a Tensor and cleared the DeviceResidencyInfo. This way the caller has to always clear the residency info themselves.

mortzur: This is doable but not as simple as it sounds, I think. 
Writing my thoughts to consider the pros and cons:

DRI is then required to hold a pointer to its owning tensor.
This is a results of DRI'a destructor calling releaseDeviceTensor(Tensor &t). 
This was put in place to prevent device memory leaks, but can also be error prone as it requires the user to be aware of double-free flows. For easiest use, "releaseDeviceTensor" must be implemented to check for double-free. 

Technical details: moving DRI member init from definition to each Tensor constructor (adds "static" complexity). 
In addition, a weird handling will then be required for Tensor's move-constructor and assignment operator , which swaps DRI pointers.

include/glow/Base/Tensor.h
mortzur: What is the use case for pinned tensors - 
Do want to be able to copy from pinned memory directly to device and vice-versa? (meaning: do we need 2 pointers (to opaque memory) in the tensor class for both pinned location and device location?)

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -701,3 +731,105 @@ void OpenCLDeviceManager::runFunctionImpl(
   // Fire the resultCB.
   resultCB(id, std::move(executeErr), std::move(context));
 }
+
+bool OpenCLDeviceManager::transferToDevice(Tensor &tensor, void *context) {
+  runtime::OpenCLDeviceTransferContext *ctx =
```
nickgg: DCHECK(context) ?

nickgg:
Are you planning to add any unit tests to this PR?

include/glow/Base/Tensor.h
nickgg: Yes we want to copy into pinned memory ahead of time and then move it to the device from that memory. I think we can store the extra pointer in the DeviceResidencyInfo for that backend? Unsure, but we dont need to solve it for this diff.

nickgg:
@mortzur is away for awhile and I'll take over this work. Since I'm going to have to put up a new PR to push to it anyway, I'm going to split out the OCL and Non-OCL parts.

PR #3673, opened 2019-10-23
[GraphOptz] Add numeric equivalence check to GraphOptz.optimizeBatchN…

mgraczyk:
…ormAfterConvButVarReused

Also simplify test and make it more similar to others in the file.

Summary:
Add numerical equivalence check to GraphOptz.optimizeBatchNormAfterConvButVarReused to ensure that the optimized graph gives the same numerical results as the unoptimized graph.

Test Plan:
Build and run tests a few times to make sure it's not flaky

```
$ for _ in {0..10}; do ninja test; done
```

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -84,6 +84,15 @@ static Function *optimizeFunction(Function *F) {
   return G;
 }
 
+/// \returns the first node in a function which has the specificied name.
+template <typename NodeT = Node>
+static const NodeT *findFunctionNodeByName(const Function *F,
+                                           const llvm::StringRef name) {
+  return llvm::dyn_cast<NodeT>(
+      std::find_if(F->getNodes().begin(), F->getNodes().end(),
+                   [=](auto &N) { return N.getName() == name; }));
+}
+
 TEST_F(GraphOptz, OptimizeClipFunnel) {
   auto *A =
       mod_.createPlaceholder(ElemKind::FloatTy, {100, 16}, "input", false);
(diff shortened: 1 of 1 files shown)


tests/unittests/GraphOptzTest.cpp
```diff
@@ -432,40 +440,43 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButConvReused) {
...
 
-  Node *BN = F_->createBatchNormalization("batch", CV, beta, gamma, mean, var,
-                                          3, 0.0001, 0.9);
-  SaveNode *ret = F_->createSave("ret", BN);
-  SaveNode *filterSave = F_->createSave("filterSave", CV->getFilter());
+  ConvolutionNode *CV = F_->createConv(bindings_, "conv", A, 16, 5, 1, 2, 1);
+  Node *BN =
+      F_->createBatchNormalization(bindings_, "batch", CV, 3, 0.0001, 0.9);
+  F_->createSave("ret", BN);
+  F_->createSave("filter", CV->getFilter());
```
SplitInfinity: Thanks for simplifying the test case construction code! Can you include screenshots of the original graph before and after your simplification to make sure they are the same? You can dump a dot graph using `Function::dumpDAG` and render it using `graphviz`.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -432,40 +440,43 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButConvReused) {
...
 
   EXPECT_EQ(F_->getNodes().size(), 4);
+  optimizedF_ = optimizeFunction(F_);
+  ASSERT_EQ(A->getNumUsers(), 2);
+
+  auto *optimizedF_bn =
+      findFunctionNodeByName<SaveNode>(optimizedF_, "ret_save");
```
SplitInfinity: Why is this suffixed with `_bn` if it is a `SaveNode`?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -432,40 +440,43 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButConvReused) {
...
-  EXPECT_TRUE(llvm::isa<BatchNormalizationNode>(ret->getInput()));
+  EXPECT_TRUE(llvm::isa<BatchNormalizationNode>(optimizedF_bn->getInput()));
+
   BatchNormalizationNode *batchNorm =
-      llvm::dyn_cast<BatchNormalizationNode>(ret->getInput());
-  EXPECT_EQ(batchNorm, BN);
-  EXPECT_TRUE(batchNorm && batchNorm->getInput() &&
-              batchNorm->getInput().getNode() == CV);
+      llvm::dyn_cast<BatchNormalizationNode>(optimizedF_bn->getInput());
+  EXPECT_TRUE(batchNorm && batchNorm->getInput());
```
SplitInfinity:
```suggestion
  ASSERT_TRUE(batchNorm);
```

tests/unittests/GraphOptzTest.cpp
```diff
@@ -432,40 +440,43 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButConvReused) {
...
+  auto *newCVNode =
+      llvm::dyn_cast<ConvolutionNode>(batchNorm->getInput().getNode());
+  EXPECT_EQ(newCVNode->getInput().getNode(), CV->getInput().getNode());
```
SplitInfinity:
```suggestion
  ASSERT_TRUE(newCVNode);
  EXPECT_EQ(newCVNode->getInput().getNode(), CV->getInput().getNode());
```

tests/unittests/GraphOptzTest.cpp
```diff
@@ -84,6 +84,14 @@ static Function *optimizeFunction(Function *F) {
   return G;
 }
 
+template <typename NodeT = Node>
+static const NodeT *findFunctionNodeByName(const Function *f,
+                                           const char *name) {
```
SplitInfinity: Why `const char*`? Why not `const std::string&` or even better yet, ` const llvm::StringRef`?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -84,6 +84,14 @@ static Function *optimizeFunction(Function *F) {
   return G;
 }
 
+template <typename NodeT = Node>
```
SplitInfinity: Please add documentation explaining what this function does. Look at other static helper functions for examples.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -84,6 +84,14 @@ static Function *optimizeFunction(Function *F) {
...
+static const NodeT *findFunctionNodeByName(const Function *f,
```
SplitInfinity:
```suggestion
static const NodeT *findFunctionNodeByName(const Function *F,
```

tests/unittests/GraphOptzTest.cpp
mgraczyk: fixed

tests/unittests/GraphOptzTest.cpp
```diff
@@ -432,40 +440,43 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvButConvReused) {
...
+  F_->createSave("filter", CV->getFilter());
 
   EXPECT_EQ(F_->getNodes().size(), 4);
+  optimizedF_ = optimizeFunction(F_);
+  ASSERT_EQ(A->getNumUsers(), 2);
+
+  auto *optimizedF_bn =
+      findFunctionNodeByName<SaveNode>(optimizedF_, "ret_save");
+  auto *optimizedF_filterSave =
+      findFunctionNodeByName<SaveNode>(optimizedF_, "filter_save");
```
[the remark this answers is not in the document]
mgraczyk: Thanks, fixed by using the node's name

tests/unittests/GraphOptzTest.cpp
mgraczyk: I'm not sure that this does preserve structure, but after the change the test is closer to testing what it actually purports to test. Also all of the other tests here setup their convs in this way.

tests/unittests/GraphOptzTest.cpp
mgraczyk: fixed

tests/unittests/GraphOptzTest.cpp
mgraczyk: thanks, added

tests/unittests/GraphOptzTest.cpp
mgraczyk: fixed

tests/unittests/GraphOptzTest.cpp
mgraczyk: fixed

tests/unittests/GraphOptzTest.cpp
SplitInfinity: I'm pretty sure it does. The only difference between the old and the new method is that all of the `Placeholders` for the `Convolution` and `BatchNormalization` nodes are created inside `Function::createXXX` instead of in the scope of the test. 

It's easy to verify - just print out the graph structure using `Function::dumpDAG` before and after your change and examine it using `graphviz`. It is a good idea to include it in the test plan too.

tests/unittests/GraphOptzTest.cpp
mgraczyk: fixed

mgraczyk:
Thanks @SplitInfinity , I've  made your suggested changes.

tests/unittests/GraphOptzTest.cpp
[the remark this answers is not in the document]
SplitInfinity: Ah yes that works too

mgraczyk:
Graph structure before and after changes:

### Before
before

### After
after


The name of the save node and some of the "users" count changed, but besides that it is the same structure.

tests/unittests/GraphOptzTest.cpp
mgraczyk: included below

SplitInfinity:
I'm a bit confused. In your "before" screenshot, I see a `SaveNode` for the output of the convolution and the output of the batch normalization. But the code your deleted creates a `SaveNode` to save the filter input of the convolution. Is there a bug in our graph construction code?

PR #3677, opened 2019-10-24
[BackendTestUtils] Add -run-disabled-tests option to ignore blacklist

jfix71:
Make it easier to test disabled tests instead of manually changing the blacklist and recompiling.

### DIFF
--- a/tests/unittests/BackendTestUtils.cpp
+++ b/tests/unittests/BackendTestUtils.cpp
@@ -33,7 +33,7 @@
 
 namespace glow {
 
-llvm::cl::OptionCategory operatorTestCat("OperatorTest Category");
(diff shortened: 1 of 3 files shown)


tests/unittests/BackendTestUtils.cpp
```diff
@@ -33,7 +33,7 @@
 
 namespace glow {
 
-llvm::cl::OptionCategory operatorTestCat("OperatorTest Category");
```
arunm-git: This was just a previously existing typo that you're fixing?

jfix71: Yep, think it was a copy paste typo.

PR #3678, opened 2019-10-24
Changing NNPIWalkConfig parameter to Constant

mjanderson09:
Summary:
Changed nnpiCustomDSPNode to accept NNPIWalkParams as a Constant instead of a pointer.

Reviewed By: jfix71

Differential Revision: D18093958


jfix71:
CC: @jsubag -- this is necessary because previously walk configs were allocated on the stack, meaning outside of a simple test scenario this was causing stack-use-after-return issues.

PR #3679, opened 2019-10-24
[PyTorch Lower prim::GetAttr to Glow

jackm321:
Summary:
Instead of bringing Module attributes like parameters into the Glow subgraph as inputs, pull the self class in as an input and then get the attributes from the Module hierarchy below that.

There are some follow ups to this that will be needed
* Change the Glow custom fuser to only fuse prim::GetAttrs that produce values consumed by Glow Subgraph
* Add checks to other node loaders to ensure that values that should come from attributes (like Conv weights for example) actually do
* Remove the other PyTorchModelLoader constructor used for onnxifi training and use this instead (and get is_parameter from the corresponding Module slot) for training use case

Documentation:
Doxygen

Test Plan:
Unit test

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -764,8 +764,15 @@ PyTorchModelLoader::getSymbolsMapping() {
 
 // static
 bool PyTorchModelLoader::isNodeSupported(const torch::jit::Node *ptNode) {
+  const auto kind = ptNode->kind();
+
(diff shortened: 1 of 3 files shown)


jackm321:
Thanks @qizzzh, I was trying to figure out why that didn't work and it makes sense now.
To me it seems like pulling the prim::GetAttrs into the Glow subgraph is the right way to go so I'll try to find a workaround for this, I'll take a look at the tracer code

Issue #3629, opened 2019-10-16
[RFC] Device Resident and Specific Tensors


nickgg:
# Device Resident and Specific Tensors

The Tensor class in Glow is used extensively as an abstraction for typed memory on the host and is used in the compiler, runtime interfaces and backend specific execution code. It’s able to serve a wide range of uses in the code by being fundamentally a thin wrapper around an underlying buffer and adding extra functionality through helper functions (such the `TensorHandle`).

As we get closer to production with Glow (and especially the runtime components) we have encountered some use cases that our Tensor abstraction does not handle well, or at all. This document addresses a few categories of data that we have no good abstraction for in Glow and need a solution for: Tensors with data that exist on the device but not the host (Device Resident) and Tensors which exist on the host in a device-specific format (e.g. aligned or pinned for a specific accelerator). Currently we handle these cases implicitly (e.g. we just know that this tensor must be aligned correctly), or we transform the Tensor before and after each usage (e.g. in DeviceManager::execute).

The solution we find to handle these issues may affect other open questions about how to transfer data between devices (e.g. P2P) and how to represent non-standard Tensors (e.g. large tensors we can’t fit in Host memory).

This is a large document, but I believe the implementation can be fairly uncomplicated (so feel free to skip to the end).

### Training (Device Resident)

When training weights we have a situation where there are mutable tensors which are modified on every execution, but are not inputs or outputs (until the run is complete). Our current DeviceManager implementations cannot handle this optimally, and must transfer the weights tensor from host to device at the start of each execution (minibatch) and back to the host at the end of each execution. This is a substantial overhead which can dominate training speed.

To support training we need to find a solution where trainable tensors are initialized before the first run, retrieved after the last epoch but on all runs in between only the inputs (for the forward pass) and outputs (for loss calculation) are transferred to/from the device.

![image](https://user-images.githubusercontent.com/701287/66937398-5ef70980-f00d-11e9-88cf-826f7a5b2058.png)

The above example demonstrates a typical training workflow where epochs are marked in green and each minibatch in purple. The weights are initialized at the beginning of the run, and are modified by the backward pass of each minibatch but reading and interpretation of the results is done only at the very end of the run. Every minibatch needs new inputs loaded from the training set, and the user may want to pull down the outputs of each minibatch to do loss calculation on the host. 

[...]

PR #3690, opened 2019-10-28
Added new interface for cacheable Placeholders

gcatron:
Summary:
This adds the notion of a cacheable Placeholder. This type of placeholder is used for values that do can be constant across more than one inference request. This is also useful for loading large constants after compilation by treating them as cacheable placeholders.
This is just the interface, implementations will follow.

Documentation:

Test Plan:

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -120,6 +120,16 @@ public:
               std::unique_ptr<ExecutionContext> context,
               runtime::ResultCBTy resultCB) = 0;
 
+  /// Copies the contents of Tensor \p T to the device resource allocated to
+  /// Placeholder \p PH. once finished calls \p resultCB with the result of the
+  /// operation.
+  virtual void
+  transferStaticPlaceholderToDevice(Placeholder *PH, Tensor *T,
+                                    std::function<void(Error)> resultCB) {
+    resultCB(MAKE_ERR(ErrorValue::ErrorCode::RUNTIME_ERROR,
+                      "Unsupported feature, cannot copy Placeholder."));
+  };
+
   /// Stops execution and shuts down the Device.
   virtual Error stop(bool block = true) { return Error::success(); };
 
--- a/include/glow/Graph/Nodes.h
+++ b/include/glow/Graph/Nodes.h
@@ -143,6 +143,10 @@ class Placeholder : public Storage {
   /// Specifies if associated Tensors should be zeroed when allocated.
   bool allocZero_{false};
 
+  /// Specifies if this is a static placeholder, this means it is set once
+  /// before the first network run and will be reused by following runs.
+  bool isStatic_{false};
+
 public:
   /// Create a new placeholder.
   Placeholder(llvm::StringRef name, TypeRef Ty, bool isTrainable)
@@ -157,6 +161,12 @@ public:
 
   /// \returns True if associated Tensors should be zeroed when allocated.
   bool allocZero() const { return allocZero_; }
+
+  /// Update the isStatic_ field.
+  void setStatic(bool isStatic) { isStatic_ = isStatic; }
+
+  /// Get the status of the isStatic_ flag.
+  bool isStatic() const { return isStatic_; }
 
   /// Sets whether or not associated Tensors should be zeroed.
   void setAllocZero(bool on = true) { allocZero_ = on; }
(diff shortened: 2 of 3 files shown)


gcatron:
@jsubag Here's the proposed interface.

include/glow/Backends/DeviceManager.h
```diff
@@ -120,6 +120,12 @@ class DeviceManager {
               std::unique_ptr<ExecutionContext> context,
               runtime::ResultCBTy resultCB) = 0;
 
+  /// Copies the contents of placeholder \p P to the device.
+  virtual Error copyTensorToDevice(Placeholder *P) {
```
jsubag: Shouldn't this also get the Function? 
For backends with placeholders that are bound to networks (instead of being global)

include/glow/Graph/Nodes.h
```diff
@@ -155,6 +164,18 @@ class Placeholder : public Storage {
   /// \returns True if associated Tensors should be zeroed when allocated.
   bool allocZero() const { return allocZero_; }
 
+  /// Update the isCacheable field.
+  void setCacheable(bool cacheable) { isCacheable_ = cacheable; }
+
+  /// Get the status of the cacheable flag.
+  bool isCacheable() { return isCacheable_; }
```
jsubag: Should probably be const

include/glow/Graph/Nodes.h
```diff
@@ -155,6 +164,18 @@ class Placeholder : public Storage {
...
+
+  /// Set the versionNumber.
+  void setVersionNumber(uint64_t version) { versionNumber_ = version; }
+
+  /// Returns the current version number.
+  uint64_t getVersion() { return versionNumber_; }
```
jsubag: Should probably be const

include/glow/Graph/Nodes.h
```diff
@@ -140,6 +140,15 @@ class Placeholder : public Storage {
...
   bool allocZero_{false};
 
+  /// Specifies if this is a cacheable placeholder, this means it should be
+  /// re-used between inferences.
+  bool isCacheable_{false};
+
+  /// Current version of the Placeholder, by default this is 0 but for a
+  /// cacheable placeholder the tensor can be updated, this is reflected by an
+  /// incrememted version number.
+  uint64_t versionNumber_{0};
```
jsubag: Not sure if this can be used from multiple threads - but if so, would suggest using an atomic.

nickgg:
@gcatron @jsubag This feels like an overlap with the Device Resident Tensors plan (https://github.com/pytorch/glow/issues/3629). I think we can solve this using that scheme.

jsubag:
@nickgg Wall of text hits me for 9000 :)
https://github.com/pytorch/glow/issues/3629 looks very detailed and it makes sense to handle this issue there too.
Will go over it.
CC: @gcatron

gcatron:
> @nickgg Wall of text hits me for 9000 :)
> #3629 looks very detailed and it makes sense to handle this issue there too.
> Will go over it.
> CC: @gcatron

@jsubag Here's a PR with an implementation on OpenCL #3671

include/glow/Backends/DeviceManager.h
gcatron: Updated this to take in a Placeholder and Tensor, I don't think network name is needed here. If the PH is used by multiple networks we want to update the value for all networks. Also added a cb function to make this an async interface

gcatron:
@jsubag, @nickgg and I had an offline conversation and determined that this is complementary to Device Resident Tensors. I've stripped this PR down to remove the overlap. The idea of versioning will be handled by DRT this will handle the deferred loading of constants.

nickgg:
Looks good to me, unsure of the `Cached` name. Offline maybe? Naming is hard.

include/glow/Backends/DeviceManager.h
jsubag: Looks good overall.
Please clarify how should a backend determine if the same PH is being used in multiple networks: name / PH address/ Tensor address?
Maybe this usage would benefit from something more explicit?

jsubag:
Just to throw my hat in the naming ring, how about persistent PH or static PH?

include/glow/Backends/DeviceManager.h
jsubag: And one more question, would the RT wait until pending inference have completed before updating such a PH or is that the backend's responsibility?

PR #3691, opened 2019-10-28
[OpenCL] relax allowedError rate. we are less than 0.0003 over it in …

shajrawi:
Test Plan:

`ninja test`


jfix71:
Can we do this on the failing test, instead of changing isEqual?

shajrawi:
@jfix71 done.

PR #3694, opened 2019-10-29
Add missing conversion to constant for DilatedConvolution test

jsubag:
Summary:
Adding a missing call to convertPlaceholdersToConstants in the DilatedConvolution test


jsubag:
@jfix71 @arunm-git , Please take a look - seems like this test is missing this conversion.

PR #3697, opened 2019-10-30
Adding a new SLWS node using offsets

qizzzh:
Summary: Offsets is the new lengths in PyTorch

Differential Revision: D18216798


include/glow/Graph/Graph.h
```diff
@@ -878,6 +878,12 @@ class Function final : public Named {
                                  NodeValue weights, NodeValue indices,
                                  NodeValue lengths);
 
+  /// Another version of SparseLengthsSum using offsets instead.
```
jackm321: Do we want to just call it EmbeddingBag since there's no lengths?

lib/Graph/Nodes.cpp
```diff
@@ -447,6 +447,31 @@ static bool verifySparseLengthsWeightedSum(NodeValue dest, NodeValue data,
...
+      expectCompareTrue("Indices must be a 1D vector", indices.dims().size(),
+                        size_t(1), dest.getNode());
+  isValid &=
+      expectCompareTrue("Offsets must be a 1D vector", offsets.dims().size(),
+                        size_t(1), dest.getNode());
+  isValid &=
+      expectCompareTrue("Weights must be a 1D vector", weights.dims().size(),
+                        size_t(1), dest.getNode());
+
+  isValid &=
```
jackm321: also verify dest vs offsets

tools/ClassGen/NodeGen.cpp
```diff
@@ -484,6 +484,23 @@ int main(int argc, char **argv) {
...
+      .addInput("Data")
+      .addInput("Weights")
+      .addInput("Indices")
+      .addInput("Offsets")
+      .addResultFromCtorArg()
+      .setDocstring(
+          "Gathers slices of the outer-most dimension of Data "
+          "indexed by Indices vector, and then accumulates them into "
+          "len(Offsets) entries: first slice between Offsets[0] and Offsets[1] "
+          "(or total length if there's only one elem in Offsets) are "
```
jackm321: @jfix71 and I were discussing this yesterday, when we compile for max length of all inputs including indices and weights then the end of the "real" part of this tensor if it's less than maximum will be larger than the legit inputs so the end indices won't be valid. I'm thinking it's ok because if we zero fill both tensors to the end then the sum of this section would be 0 due to the 0 weights. Depending on how it's implemented, this may be inefficient though. We could do something like add an extra offset at the end to mark the end of the last segment.

jackm321: Yeah actually nvm I think adding the extra offset in this glow op isn't feasible, I guess we have to rely on 0 weights

qizzzh: Currently we have a branch in this op since there's no mark of the end..

jackm321: Not sure I understand what you mean

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -3211,6 +3211,52 @@ void BoundInterpreterFunction::fwdSparseLengthsWeightedSumGradInst(
...
+  size_t lineSize = data->size() / data->dims()[0];
+
+  auto DH = data->getHandle<ElemTy>();
+  auto WH = weights->getHandle<ElemTy>();
+  auto OH = out->getHandle<ElemTy>();
+
+  size_t curIdx = 0;
+  for (size_t i = 0; i < segments; i++) {
+    size_t start = OFFH.raw(i);
+    size_t end = i == segments - 1 ? totalLength : OFFH.raw(i + 1);
```
qizzzh: @jackm321 I meant this line

jackm321: Ah ok thanks, yeah this is the thing that won't work. Because irl totalLength won't actually be the size of the "real" inputs.

qizzzh: Ok. I guess we'll just depend on weights padding for now.

PR #3700, opened 2019-10-30
v0.3.2H0 Update for the Intel NNPI Backend

jsubag:
Summary: v0.3.2H0 Update for the Intel NNPI Backend


lib/Backends/NNPI/ClassGen/NNPISpecificNodes.h
```diff
@@ -26,6 +26,6 @@ BB.newNode("NNPICustomDSP")
     .setDocstring("This is an experimental NNPI-specific node representing a "
                   "custom DSP op");
 
-BB.includeBackendSpecificVerification("glow/NNPISpecificNodesVerification.h");
+BB.includeBackendSpecificVerification("NNPISpecificNodesVerification.h");
```
arunm-git: I undid this change - it broke our internal builds and it looks like the glow/ prefix is what is consistent with the other backends. Maybe the NNPI CMakeLists.txt file needs a change?

arunm-git:
Hi @jsubag, this PR has been imported. 
* I've set supportsPartialTensors to return false for now, until we sort out the issues in it that we're discussing.
* Fixed a few minor lint issues and the include file change I mentioned.
Thanks for getting this up.

PR #3705, opened 2019-10-31
[PyTorch] Add profiling executor example

jackm321:
Summary:
Add a quick example of using the profiling executor to run scripted modules with Glow.

Documentation:
comments

Test Plan:
`python examples/profiling_executor_example.py`

### DIFF
--- a/torch_glow/examples/profiling_executor_example.py
+++ b/torch_glow/examples/profiling_executor_example.py
@@ -1,0 +1,36 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import torch
+import torch.nn as nn
+import torch_glow
+
+
+class Model(nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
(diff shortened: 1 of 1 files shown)


torch_glow/examples/profiling_executor_example.py
```diff
@@ -0,0 +1,36 @@
...
+class Model(nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
+        self.linear = nn.Linear(10, 2)
+
+    def forward(self, x):
+        return self.linear(x)
+
+
+torch._C._jit_set_profiling_mode(True)
```
yinghai: Using context manager might be a better example.

jackm321: Yeah, imo it's not worth pulling in jit_utils for it though

yinghai: Oh, yeah. Hmm, we need to move that stuff into `torch.jit`.

PR #3710, opened 2019-10-31
Update thresholds for ParameterSweepTest.

jsubag:
Summary: Updating thresholds for FC, BatchMatMul and Conv sweep tests.


jsubag:

jsubag:
CC: @omromano

jfix71:
@hyuen Can you take a look here?

hyuen:
@jfix71  and I met and discussed about this. It seems right now that we are comparing fp32 as a baseline vs a quantized approach. The error threshold is dependent on the dimensions of the inputs, and by changing these thresholds (extending them) we are just accepting the biggest error which comes from the biggest dimension.

The proposed fix is to compare with a baseline in the same precision, and expect bitwise accuracy for most of the cases, that way we can flush out any differences in the quantization or accumulator schemes.

jsubag:
@hyuen I'm all for comparing with a more accurate reference - but skeptical about _bitwise_ comparisons for non-integer arithmetic (especially across different HW).
Can you share more details about your reference model?

PR #3717, opened 2019-11-01
[PyTorch] quantize quantized::linear bias with input value

jackm321:
Summary:
Choose bias quantization params based on value in the bias. There are still mismatched between PyTorch and Glow quantization but they are much smaller now (see test plan).

Documentation:
n/a

Test Plan:
enable `test_quantized_linear_random_input()` `bias = torch.tensor([1, 2, 3, 4, 5, 6, 7], dtype=torch.float) * 0.1` and measure the diff between PyTorch and Glow results.

before
```
 tensor([[-0.0769, -0.1538, -0.3077, -0.4615, -0.5385,  0.3846,  0.2308],
        [-0.1538, -0.1538, -0.3077, -0.3846, -0.4615,  0.3846,  0.3077],
        [-0.0769, -0.1538, -0.3077, -0.3846, -0.4615,  0.3846,  0.3077],
        [-0.0769, -0.2308, -0.3077, -0.3846, -0.5385,  0.3846,  0.3077],
        [-0.1538, -0.1538, -0.3077, -0.3846, -0.5385,  0.4615,  0.3077],
        [-0.0769, -0.2308, -0.3077, -0.4615, -0.4615,  0.3846,  0.3077],
        [-0.1538, -0.1538, -0.3077, -0.3846, -0.4615,  0.3846,  0.3077]])
```
after
```
 tensor([[-0.0769,  0.0769,  0.0000, -0.0769,  0.0000,  0.0769,  0.0000],
        [ 0.0000,  0.0769,  0.0769,  0.0000,  0.0769,  0.0000,  0.0000],
        [-0.0769,  0.0769,  0.0000,  0.0000, -0.0769,  0.0000,  0.0000],
        [ 0.0000,  0.0000,  0.0000, -0.0769,  0.0000,  0.0000,  0.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000,  0.0000,  0.0000,  0.0000],
        [-0.0769,  0.0000,  0.0000,  0.0000,  0.0769,  0.0000,  0.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000,  0.0769,  0.0769,  0.0000]])
```

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -17,6 +17,7 @@
 #include "PyTorchModelLoader.h"
 #include "PyTorchCommon.h"
 
+#include "glow/Quantization/Base/Base.h"
 #include "glow/Support/Error.h"
 #include "glow/Support/Support.h"
 
@@ -876,6 +877,11 @@ Error PyTorchModelLoader::loadNodes(const torch::jit::Graph &graph) {
     RETURN_ERR_IF_NOT(it != mapping.end(),
                       glow::strFormat("Node kind %s is not supported by Glow",
                                       node->kind().toDisplayString()));
+
+    // TODO: once we have weight unpacking for quantized parameters we can
+    // totally remove this.
+    RETURN_IF_ERR(freezeWeights(node));
+
     RETURN_IF_ERR((this->*it->second.loadFn)(node));
   }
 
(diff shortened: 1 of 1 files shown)


jackm321:
For the remaining errors. The first thing I'm going to try next is creating an FC instruction in Glow and using that instead of Matmul + BatchedAdd for quantized FC because right now there is a precision loss when we convert matmul output to int8 then apply the bias instead of doing the bias addition in int32 then going to int8

jackm321:
@yinghai not yet, but one we should look into is adding a quantized fc op to the Glow interpreter so that the int32 bias is added to the accumulated sum before requantizing back to int8.

yinghai:
Sounds good. Thanks!

jackm321:
@zrphercule well from the above differences, we can see that the errors are much lower now but not zero so I think we need to keep working on this

PR #3718, opened 2019-11-01
[GraphOpt] Add optimization for folding Clip(Splat(args)) -> Splat(args')

jfix71:
Summary: Did not previously have this optimization. This is relevant when we are doing FP16 clipping -- without having this pass prior to the `ConstantFold` pass, we are stupidly materializing Splats into Constants.

Test Plan: Added unit test.


jfix71:
CC: @mjanderson09

jackm321:
Ah just saw the summary 😛  makes sense

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -80,7 +80,7 @@ FunctionPassPipeline glow::createDefaultGraphOptimizationPassPipeline() {
       {FunctionPassID::OptimizeArithmeticNodes},
 
       // Optimize Tensor shape transformations.
```
jackm321: Update comment?

Issue #3463, opened 2019-08-27
[Quantization] Fused 4-bit quantization support


beicy:
This issue is created to track fused 4-bit quantization support in Glow.  
The fused 4-bit quantization tensor is represented as follows:
```
|   .... int4 data ...       | scale | offset |                                                                                           
|num_of_input_columns * 0.5B |  2B   |   2B   | 
```
The plan is:
Step1 : implement an tensor fused 4-bit quantization/dequantization
Step2: Based on model requests, support fused 4-bit quantization/dequantization in Glow nodes.
Step3: implement  Glow fused4BitQuantizationNode (Optional).
Step4: 4-bits Quantized model testing.
cc @rdzhabarov @hyuen @bertmaher

opti-mix:
Can it be closed now or is there anything to be done yet?

beicy:
@opti-mix On my project side, the current support should be OK. If no more update from @jfix71 , I think it can be closed.

jfix71:
Yeah think this is done.

PR #3719, opened 2019-11-01
Add support for 4-bit fused-rowwise-quantized SLWS to Interpreter

jfix71:
Summary: This PR adds support for 4-bit rowwise quantization to FRWQ-SLWS.

Test Plan: Added tests to cover this. I also added a new set of tests (see second commit) that has two columns in addition to being weighted -- previously all of our weighted tests only had a single column of data.

Related to https://github.com/pytorch/glow/issues/3463

CC: @jsubag


lib/Graph/Graph.cpp
```diff
@@ -1680,28 +1680,39 @@ Function::createRowwiseQuantizedSparseLengthsSum(
...
 /// createRowwiseQuantizedSparseLengthsSum and
 /// createRowwiseQuantizedSparseLengthsWeightedSum.
-/// Function \p F is used to get the speficific type, using inputs \p inDims and
-/// \p lenghtsDims to compute output dimensions.
-static TypeRef getOutputTypeOfFusedRowwiseQuantizedSLS(
-    Function *F, const llvm::ArrayRef<size_t> &inDims,
-    const llvm::ArrayRef<size_t> &lengthsDims, ElemKind scaleOffsetKind) {
-  ShapeVector outDims(inDims.begin(), inDims.end());
+/// Function \p F is used to get the specific type, using inputs \p data and
+/// \p lengthsDims to compute output dimensions. \p dataKind represents the kind
```
beicy: Where is dataKind?

lib/Graph/Graph.cpp
```diff
@@ -1758,21 +1781,22 @@ static Constant *quantizeDataForFusedRowwiseQuantizedSparseLengthsWeightedSum(
...
+          this, data, fusedElemKind);
   return createFusedRowwiseQuantizedSparseLengthsWeightedSum(
       name, rwqData, weights, indices, lengths, precision, useFP16Accumulation);
 }
 
 FusedRowwiseQuantizedSparseLengthsSumNode *
 Function::createFusedRowwiseQuantizedSparseLengthsSum(
     llvm::StringRef name, Tensor &data, NodeValue indices, NodeValue lengths,
-    ElemKind precision, bool useFP16Accumulation) {
+    ElemKind precision, bool useFP16Accumulation, ElemKind fusedElemKind) {
```
beicy: since now there is fusedElemKind, do we still need precision?  I mean, for 4bit, it must be float16Ty.

jackm321:
LGTM!
The pytorch failure is unrelated and should be fixed by #3717

lib/Graph/Graph.cpp
jfix71: Good point @beicy -- I added a new commit where I went through and cleaned up all the APIs that don't need precision. Now we will pass the fused ElemKind and then infer the precision from the equivalent ElemKind used by scale/offset.

PR #3721, opened 2019-11-01
[Bundle](trivial) Change bundle weights file names to ".weights.bin" / ".weights.txt"

mciprian13:
**Summary**
The bundle generates the weights in 2 files with 2 formats:
- binary: the file name was "<bundle_name>.weights"
- text (include C buffer array format): the file name was "<bundle_name>.inc"

I changed the generated file names to have more clarity:
- <bundle_name>.weights.bin - for the binary format
- <bundle_name>.weights.txt - for the text format

**Documentation**
Small updates.

**Test Plan**
None


mciprian13:
I will refactor the AOT.md documentation as part of this PR.

mciprian13:
Question: can I also change the default option for the generated header file from dynamic to static? Personally I find it easier to use and create applications with it (the reason I made it in the first place).
Thanks!

opti-mix:
> Question: can I also change the default option for the generated header file from dynamic to static? Personally I find it easier to use and create applications with it (the reason I made it in the first place).
Thanks!

Sure! Let's go for static by default.

mciprian13:
Changed the default bundle API type to static and added extra sections in AOT.md: memory layout, static bundle API, dynamic bundle API, some extra useful infos.

PR #3730, opened 2019-11-04
[PyTorch] Add quantized packing weights support for fc & conv

zrphercule:
This pr added quantized conv, quantized conv_relu and quantized linear supported with packed weights & bias.
Also, conv & conv_relu has groupwise supported.
Unit test:

unit test 
quantized_conv2d_test.py 
quantized_conv2d_relu_test.py 
quantized_linear_test.py

### DIFF
--- a/torch_glow/src/GlowIValue.h
+++ b/torch_glow/src/GlowIValue.h
@@ -41,6 +41,7 @@ public:
     DoubleList,
     BoolList,
     Tuple,
+    PTTensor,
   };
 
 private:
@@ -56,6 +57,7 @@ private:
     std::vector<double> *asDoubleList;
     std::vector<bool> *asBoolList;
     std::vector<GlowIValue> *asTuple;
+    at::Tensor *asPTTensor;
   };
 
   Tag tag_ = Tag::None;
@@ -96,6 +98,7 @@ public:
   bool isDoubleList() const;
   bool isBoolList() const;
   bool isTuple() const;
+  bool isPTTensor() const;
 
   /// \returns Payload a glow Tensor or error if the tag is not Tensor.
   Expected<Tensor *> toTensor();
@@ -138,11 +141,22 @@ public:
   /// \returns Payload a vector of GlowIValues or error if the tag is not Tuple.
   Expected<const std::vector<GlowIValue> *> toTuple() const;
 
+  /// \returns Payload a PyTorch Tensor* or error if the tag is not a PyTorch
+  /// Tensor.
+  Expected<at::Tensor *> toPTTensor();
+
+  /// \returns Payload a const Pytorch Tensor* or error if the tag is not
+  /// Tensor.
+  Expected<const at::Tensor *> toPTTensor() const;
+
   /// Set the tag to None.
   void fromNone();
 
   /// Set the tag to Tensor.
   void fromTensor(Tensor tensor);
+
+  /// Set the tag to PyTorch Tensor.
+  void fromPTTensor(at::Tensor tensor);
 
   /// Set the tag to Double.
   void fromDouble(double d);
--- a/torch_glow/src/PyTorchCommon.cpp
+++ b/torch_glow/src/PyTorchCommon.cpp
@@ -132,6 +132,14 @@ glow::ElemKind scalarTypeToElemKind(c10::ScalarType ty) {
     return ElemKind::Int64ITy;
   } else if (ty == at::kBool) {
     return ElemKind::BoolTy;
+  } else if (ty == at::kByte) {
+    // We should have an 8-byte non-quantized integer type eventually
+    // Currently usage of Bool is fine
+    return ElemKind::BoolTy;
+  } else if (ty == at::kQInt8) {
+    return ElemKind::Int8QTy;
+  } else if (ty == at::kQUInt8) {
+    return ElemKind::UInt8QTy;
   } else {
     LOG(DFATAL) << "ScalarType " << static_cast<int>(ty)
                 << " not supported yet.";
@@ -177,6 +185,20 @@ glow::Type ptTypeToGlowType(const c10::TensorType &ptType) {
   return glow::Type(scalarTypeToElemKind(scalarType), dims);
 }
 
+glow::Type ptTypeToGlowType(const c10::TensorType &ptType, float scale,
+                            int32_t zero_point) {
+  DCHECK(ptType.scalarType().has_value())
+      << "TensorType has no associated scalar type.";
+  const auto concreteSizes = ptType.sizes().concrete_sizes().value();
+  std::vector<glow::dim_t> dims;
+  for (const auto &size : concreteSizes) {
+    dims.push_back(static_cast<glow::dim_t>(size));
+  }
+
+  auto scalarType = ptType.scalarType().value();
+  return glow::Type(scalarTypeToElemKind(scalarType), dims, scale, zero_point);
+}
+
 at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
   std::vector<int64_t> sizes;
   for (const auto dim : glowType.dims()) {
@@ -188,7 +210,41 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
 }
 
 glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor) {
-  auto glowType = ptTypeToGlowType(*c10::TensorType::create(ptTensor));
-  return glow::Tensor(ptTensor.data_ptr(), &glowType);
+  if (ptTensor.is_quantized()) {
+    float scale = 1.0;
+    int32_t offset = 0;
+    if (ptTensor.qscheme() == at::kPerChannelAffine) {
+      // If it is channel wise quantized, which means
+      // this tensor is the weight of quantized linear or conv
+      // Then we dont deal with the qparams here,
+      // and only set up soome dummy scale & offset by using the first
+      // elements's scale & offset.
+      scale = ptTensor.q_per_channel_scales()[0].item<float>();
+      offset = ptTensor.q_per_channel_zero_points()[0].item<int32_t>();
+    } else if (ptTensor.qscheme() == at::kPerTensorAffine) {
+      scale = static_cast<float>(ptTensor.q_scale());
+      offset = static_cast<int32_t>(ptTensor.q_zero_point());
+    } else {
+      LOG(DFATAL)
+          << "PyTorch tensor with unsupported quantization scheme detected.";
+    }
+    auto glowType =
+        ptTypeToGlowType(*c10::TensorType::create(ptTensor), scale, offset);
+    return glow::Tensor(ptTensor.data_ptr(), &glowType);
+  } else {
+    auto glowType = ptTypeToGlowType(*c10::TensorType::create(ptTensor));
+    return glow::Tensor(ptTensor.data_ptr(), &glowType);
+  }
 }
+
+at::Tensor glowTensorToPTTensor(const glow::Tensor &glowTensor,
+                                const at::ScalarType &torch_type) {
+  std::vector<int64_t> sizes;
+  for (const auto dim : glowTensor.dims()) {
+    sizes.push_back(dim);
+  }
+  return at::from_blob(glowTensor.getUnsafePtr(), sizes,
+                       at::device(at::kCPU).dtype(torch_type));
+}
+
 } // namespace glow
--- a/torch_glow/src/PyTorchCommon.h
+++ b/torch_glow/src/PyTorchCommon.h
@@ -26,6 +26,11 @@
 
 namespace glow {
 
+/// For Glow: -128 <= orig_fp32/scale_1 + offset_1 <= 127
+/// For PyTorch: 0 <= orig_fp32/scale_2 + offset_2 <= 255
+/// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128
+const int32_t OFFSETSHIFT = 128;
+
 extern bool GlowCompilePyTorchModule;
 /// Various settings to be used by code that loads PyTorch models. There should
 /// only be one of these and it should be obtained by calling
@@ -87,6 +92,11 @@ glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor);
 /// Given a Glow Type \p glowType, \returns an empty PyTorch Tensor with a
 /// matching type.
 at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType);
+
+/// Given a Glow Tensor \p glowTensor, \returns a PyTorch Tensor with the same
+/// type, shape and content.
+at::Tensor glowTensorToPTTensor(const glow::Tensor &glowTensor,
+                                const at::ScalarType &torch_type);
 
 } // namespace glow
 
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -22,6 +22,8 @@
 #include "glow/Support/Support.h"
 
 #include <ATen/ATen.h>
+#include <ATen/core/dispatch/Dispatcher.h>
+#include <ATen/native/c10_utils.h>
 #include <torch/csrc/jit/ir.h>
 
 namespace glow {
@@ -31,11 +33,6 @@ namespace {
 /// In Glow, the activations are quantized to int_8. Therefore, for the offset
 /// read from quantized pytorch model, we need to subtract 128(i.e. INT8_MIN) to
 /// make the activations becomes int8_t.
-
-/// For Glow: -128 <= orig_fp32/scale_1 + offset_1 <= 127
-/// For PyTorch: 0 <= orig_fp32/scale_2 + offset_2 <= 255
-/// Therefore, we can make scale_1 == scale_2, and offset_1 = offset2 - 128
-const int32_t OFFSETSHIFT = 128;
 
 /// Downcast a double to a float.
 Expected<float> to32Bit(double val) {
@@ -214,6 +211,16 @@ Expected<std::vector<int64_t> *>
 iValToIntList(Expected<GlowIValue *> expectedIVal) {
   if (expectedIVal) {
     return (*expectedIVal)->toIntList();
+  } else {
+    return expectedIVal.takeError();
+  }
+}
+
+/// Unwrap a Expected<GlowIValue *> \p expectedIVal and call toPTTensor,
+/// propogate any Errors.
+Expected<at::Tensor *> iValToPTTensor(Expected<GlowIValue *> expectedIVal) {
+  if (expectedIVal) {
+    return (*expectedIVal)->toPTTensor();
   } else {
     return expectedIVal.takeError();
   }
@@ -528,6 +535,20 @@ struct QuantizedUnpackedConv2dInputs {
   };
 };
 
+/// Indexes of quantized::conv2d and quantized::conv2d_relu inputs.
+struct QuantizedConv2dInputs {
+  enum {
+    input = 0, // NCHW
+    packed_weights = 1,
+    stride = 2,
+    padding = 3,
+    dilation = 4,
+    group = 5,
+    scale = 6,
+    zero_point = 7,
+  };
+};
+
 /// Indexes of quantized::add_relu inputs.
 struct QuantizedAddReluInputs {
   enum {
@@ -549,13 +570,23 @@ struct QuantizedAddInputs {
 };
 
 /// Indexes of glow::unpacked_quantized_linear inputs.
-struct QuantizedLinearInputs {
+struct QuantizedUnpackedLinearInputs {
   enum {
     input = 0,
     weight = 1,
     bias = 2,
     scale = 3,
     zero_point = 4,
+  };
+};
+
+/// Indexes of quantized::linear inputs.
+struct QuantizedLinearInputs {
+  enum {
+    input = 0,
+    packed_weights = 1,
+    scale = 2,
+    zero_point = 3,
   };
 };
 
@@ -761,13 +792,32 @@ PyTorchModelLoader::getSymbolsMapping() {
         QuantizedUnpackedConv2dInputs::scale,
         QuantizedUnpackedConv2dInputs::zero_point}},
       {{"glow::unpacked_quantized_linear"},
+       &PyTorchModelLoader::loadQuantizedLinearUnpacked,
+       {
+           QuantizedUnpackedLinearInputs::weight,
+           QuantizedUnpackedLinearInputs::bias,
+           QuantizedUnpackedLinearInputs::scale,
+           QuantizedUnpackedLinearInputs::zero_point,
+       }},
+      {{"quantized::linear"},
        &PyTorchModelLoader::loadQuantizedLinear,
        {
-           QuantizedLinearInputs::weight,
-           QuantizedLinearInputs::bias,
+           QuantizedLinearInputs::packed_weights,
            QuantizedLinearInputs::scale,
            QuantizedLinearInputs::zero_point,
        }},
+      {{"quantized::conv2d"},
+       &PyTorchModelLoader::loadQuantizedConv,
+       {QuantizedConv2dInputs::packed_weights, QuantizedConv2dInputs::stride,
+        QuantizedConv2dInputs::padding, QuantizedConv2dInputs::dilation,
+        QuantizedConv2dInputs::group, QuantizedConv2dInputs::scale,
+        QuantizedConv2dInputs::zero_point}},
+      {{"quantized::conv2d_relu"},
+       &PyTorchModelLoader::loadQuantizedConvRelu,
+       {QuantizedConv2dInputs::packed_weights, QuantizedConv2dInputs::stride,
+        QuantizedConv2dInputs::padding, QuantizedConv2dInputs::dilation,
+        QuantizedConv2dInputs::group, QuantizedConv2dInputs::scale,
+        QuantizedConv2dInputs::zero_point}},
       {{"aten::quantize_per_tensor"},
        &PyTorchModelLoader::loadQuantize,
        {QuantizeInputs::scale, QuantizeInputs::zero_point,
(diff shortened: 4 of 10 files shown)


zrphercule:
Also, the commit tree is a little mess now
will squash them into one at last.

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -188,7 +215,138 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
...
+}
+
+std::tuple<at::Tensor, c10::optional<at::Tensor>>
+fbgemm_conv_unpack(at::Tensor packed_weights, const int kSpatialDim) {
+
+  // TODO zrphercule:
+  // We should enbale the following code instead of using the pointer directly.
+  // auto& pack_ptr =
+  // at::cpp_custom_type_hack::cast<PackedConvWeight<2>>(packed_weights);
+  PackedConvWeight<2> &pack_ptr = *reinterpret_cast<PackedConvWeight<2> *>(
```
yinghai: What's 2 here?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1196,20 +1245,67 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
+      packedWeightConstant->getPayload().getHandle<bool>().clone();
+
+  auto ptTensor = glowTensorToPTTensor(packedWeightTensor);
+
+  auto unpackedParams = glow::fbgemm_linear_unpack(ptTensor);
+  const at::Tensor &ptWeightTensor = std::get<0>(unpackedParams);
+  const c10::optional<at::Tensor> &ptBiasTensor = std::get<1>(unpackedParams);
+
+  // unpacked weights
+  auto weightTensor = ptTensorToGlowTensor(ptWeightTensor).clone();
```
yinghai: Why do we clone here?

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -132,6 +137,14 @@ glow::ElemKind scalarTypeToElemKind(c10::ScalarType ty) {
     return ElemKind::Int64ITy;
   } else if (ty == at::kBool) {
     return ElemKind::BoolTy;
+  } else if (ty == at::kByte) {
```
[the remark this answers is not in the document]
zrphercule: The packed weights tensor is a tensor of byte. It is actually a pointer, pointer to another address, where they store real data

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -188,7 +215,138 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
...
+  return at::from_blob(glowTensor.getUnsafePtr(), sizes,
+                       at::device(at::kCPU).dtype(at::kByte));
 }
+
+std::tuple<at::Tensor, c10::optional<at::Tensor>>
+fbgemm_linear_unpack(at::Tensor packed_weight) {
+
+  // TODO zrphercule:
+  // We should enbale the following code instead of using the pointer directly.
+  // auto& pack_ptr =
```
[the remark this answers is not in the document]
zrphercule: the cpp_custom_hack:cast requires all the tensors to be created by cpp_custom_hack:create, and our tensor is not created by that. that cast was actually do two check and then use the pointer directly. 
See https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/cpp_custom_type_hack.h#L24

torch_glow/src/PyTorchCommon.cpp
zrphercule: 2 indicate 2d, and 3 indicate 3d. Since we did enbaled 3d yet.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: This is the bug that cause our result to be random. in ptTensorToGlowTensor, it uses the pytorch data pointer to create a glow tensor, however I think the address of the data is somehow not protected (or it is because of pytorch's tensor is not stored continously) Eventually if we dont clone a new one, move all the data to a safe place, they will be freed sometime when running.

zrphercule:
> Overall it looks good. I think we really need to start enabling internal tests. All these tensor operations make me worry about memory issue.

Agree, I maybe able to set up some e2e test when running resnet. Also we would need some unit test, I can do that later.

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
zrphercule: if we do call the cpp_custom_hack::cast, it will failed the second check, which claims we dont have the same deleter.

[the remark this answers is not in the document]
zrphercule: Eventually we would expect to not have these two functions in our codebase and call the function dispatcher of unpack. For now I would like to enbale the functionality first.

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
yinghai: Then why do we use BoolTy instead of UInt8Ty

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
yinghai: OK

torch_glow/src/PyTorchCommon.cpp
yinghai: Can you document this in the comments?

torch_glow/src/PyTorchModelLoader.cpp
yinghai: Document?

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
zrphercule: We dont have UInt8Ty... https://github.com/pytorch/glow/blob/d8ae137d59bf51a4a09449402159944b621738fb/include/glow/Base/Type.h#L267

We only have quantized 8-bits type, and it requires a lot of params and has some different API sometime, so not good to use. We can add a Byte later, in this diff we are using bool.

torch_glow/src/PyTorchCommon.cpp
zrphercule: Sure!

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
yinghai: OK. Yeah, we need to add that eventually.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Sure, I will address in all clone places.

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -25,6 +25,11 @@
 #include <torch/csrc/jit/pass_manager.h>
 #include <torch/csrc/jit/passes/subgraph_rewrite.h>
 
+#include <ATen/ATen.h>
```
jackm321: Move these to another file so we know what to delete later

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1196,20 +1245,67 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
-      weight, getGlowNodeValueForValue(inputs[QuantizedLinearInputs::weight]));
+      packedWeight,
+      getGlowNodeValueForValue(inputs[QuantizedLinearInputs::packed_weights]));
+
+  glow::Constant *packedWeightConstant =
+      llvm::dyn_cast<glow::Constant>(packedWeight.getNode());
+  RETURN_ERR_IF_NOT(packedWeightConstant,
+                    "quantized::linear packed weights must be constant");
+  const glow::Tensor &packedWeightTensor =
+      packedWeightConstant->getPayload().getHandle<bool>().clone();
```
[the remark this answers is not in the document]
jackm321: If my understanding of what's happening here is right, the later one makes sense but this one shouldn't be needed

torch_glow/src/PyTorchModelLoader.cpp
jackm321: I think instead of calling clone() here, you can/should just call ensureIsOwned() on the resulting glow constant because that's what that is for

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1196,20 +1245,67 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
+      "quantized_linear_weights", std::move(weightTensor));
+  auto weight = weightConstant->getOutput();
   weight = rescaleUIntToInt(weight);
 
-  RETURN_ERR_IF_NOT(weight.dims().size() == 2, "Expected 2d Linear weights");
+  // unpacked bias
+  glow::Tensor biasTensor;
+  glow::NodeValue bias;
+  if (ptBiasTensor.has_value()) {
+    biasTensor = ptTensorToGlowTensor(ptBiasTensor.value()).clone();
```
jackm321: same thing for this one

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1196,20 +1245,67 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
+  auto weightTensor = ptTensorToGlowTensor(ptWeightTensor).clone();
+  glow::Constant *weightConstant = F_.getParent()->createConstant(
+      "quantized_linear_weights", std::move(weightTensor));
+  auto weight = weightConstant->getOutput();
   weight = rescaleUIntToInt(weight);
 
-  RETURN_ERR_IF_NOT(weight.dims().size() == 2, "Expected 2d Linear weights");
+  // unpacked bias
+  glow::Tensor biasTensor;
+  glow::NodeValue bias;
```
jackm321: this declaration isn't needed up here, you can just get the output below like you did for weight above

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1196,20 +1245,67 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
+    biasTensor = ptTensorToGlowTensor(ptBiasTensor.value()).clone();
+  } else {
+    biasTensor = glow::Tensor(glow::ElemKind::FloatTy, {weight.dims()[1]});
+    biasTensor.zero();
+  }
 
+  // Choose bias quantization params and quantize it.
+  glow::Constant *biasConstant = F_.getParent()->createConstant(
+      "quantized_linear_bias", std::move(biasTensor));
+  RETURN_ERR_IF_NOT(biasConstant, "quantized::linear bias must be constant");
```
jackm321: You just created the constant 1 line above so this isnt' needed

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1226,9 +1322,50 @@ Error PyTorchModelLoader::loadQuantizedLinear(const torch::jit::Node *ptNode) {
...
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      input,
+      getGlowNodeValueForValue(inputs[QuantizedUnpackedLinearInputs::input]));
+  input = rescaleUIntToInt(input);
+
+  glow::NodeValue weight;
+  ASSIGN_VALUE_OR_RETURN_ERR(
+      weight,
+      getGlowNodeValueForValue(inputs[QuantizedUnpackedLinearInputs::weight]));
+  weight = rescaleUIntToInt(weight);
```
jackm321: is this needed? Aren't linear weights stored as int8?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1458,7 +1595,6 @@ Error PyTorchModelLoader::loadListConstruct(const torch::jit::Node *ptNode) {
   auto outputs = ptNode->outputs();
   // Requires -1 because this requires at least one input.
   RETURN_IF_ERR(checkInputAndOutputSizes(inputs, -1, outputs, 1));
-
```
jackm321: please don't delete these white spaces

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1960,6 +2095,237 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
   return addValueMapping(outputs[0], dn->getResult());
 }
 
+Error PyTorchModelLoader::loadQuantizedConvRelu(
```
jackm321: This is a lot of shared code with loadQuantizedConv, can you please factor this out into a helper function

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1960,6 +2095,237 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
...
+      packedWeightConstant->getPayload().getHandle<bool>().clone();
+
+  auto ptTensor = glowTensorToPTTensor(packedWeightTensor);
+
+  auto unpackedParams = glow::fbgemm_conv_unpack(ptTensor);
+  const at::Tensor &ptWeightTensor = std::get<0>(unpackedParams);
+  const c10::optional<at::Tensor> &ptBiasTensor = std::get<1>(unpackedParams);
+
+  // unpacked weights
+  auto weightTensor = ptTensorToGlowTensor(ptWeightTensor).clone();
```
jackm321: same comments about clone here

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1960,6 +2095,237 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
...
+  glow::Constant *weightConstant = F_.getParent()->createConstant(
+      "quantized_conv2d_weights", std::move(weightTensor));
+  auto weight = weightConstant->getOutput();
+  weight = rescaleUIntToInt(weight);
```
jackm321: is this actually a uint8?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1960,6 +2095,237 @@ Error PyTorchModelLoader::loadDequantize(const torch::jit::Node *ptNode) {
...
+    biasTensor = glow::Tensor(glow::ElemKind::FloatTy, {weightShape.n});
+    biasTensor.zero();
+  }
+  glow::Constant *biasConstant = F_.getParent()->createConstant(
+      "quantized_conv2d_bias", std::move(biasTensor));
+  bias = biasConstant->getOutput();
+
+  auto biasType = F_.getParent()->uniqueType(
+      glow::ElemKind::Int32QTy, bias.dims(),
+      input.getType()->getScale() * weight.getType()->getScale(), 0);
```
jackm321: is 0 the right zeropoint?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2198,7 +2564,7 @@ Error PyTorchModelLoader::loadAdaptiveAvgPool2d(
 
   size_t inputH = input.dims()[1];
   size_t inputW = input.dims()[2];
-
+  input = rescaleUIntToInt(input);
```
jackm321: thanks for adding this, can you put it right after the ASSIGN_VALUE_OR_RETURN_ERR so it doesn't get separate further, we had a bug earlier because of that

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2700,8 +3066,9 @@ Error PyTorchModelLoader::loadFlatten(const torch::jit::Node *ptNode) {
   RETURN_ERR_IF_NOT(endDim == -1, "only -1 value for end_dim is supported.");
 
   auto xDim = glow::flattenCdr(in.dims(), startDim);
+  in = rescaleUIntToInt(in);
```
jackm321: same

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2898,7 +3265,6 @@ Error PyTorchModelLoader::loadAttributes(
   std::unordered_map<const torch::jit::Value *,
                      std::pair<const c10::ivalue::Object *, std::string>>
       objectTree;
-
```
jackm321: please no

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2935,6 +3301,7 @@ Error PyTorchModelLoader::loadAttributes(
     const auto ival = parentObject->getAttr(attrName);
 
     // Concatenation of names of Objects and fields referenced in the Module
+    //
```
jackm321: formatting

torch_glow/tests/nodes/quantized_linear_test.py
```diff
@@ -5,6 +5,56 @@
...
+
+    linear.weight.data.fill_(1.2)
+    linear.bias.data.fill_(3.0)
+
+    model = torch.nn.Sequential(q, linear, dq)
+    model.qconfig = torch.quantization.get_default_qconfig('fbgemm')
+    torch.quantization.prepare(model, inplace=True)
+    torch.quantization.convert(model, inplace=True)
+
+    x = torch.tensor(range(5), dtype=torch.float)
```
jackm321: Unpacked linear had enough accuracy to use random inputs, does this not?

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
zrphercule: > Why not just add it now?

lol I tried to let it working for now; Will send another pr to do so.

torch_glow/src/PyTorchModelLoader.cpp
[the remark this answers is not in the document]
zrphercule: Yeah you are right

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -188,7 +215,138 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
 }
 
 glow::Tensor ptTensorToGlowTensor(const at::Tensor &ptTensor) {
-  auto glowType = ptTypeToGlowType(*c10::TensorType::create(ptTensor));
-  return glow::Tensor(ptTensor.data_ptr(), &glowType);
+  if (ptTensor.is_quantized()) {
+    float scale = 1.0;
+    int32_t offset = 0;
+    if (ptTensor.qscheme() == at::kPerChannelAffine) {
```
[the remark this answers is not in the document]
zrphercule: Yes. So for the packed weights loading as by default attributes in the model, it is channel wise quantized by default. Therefore, if we use the by default setting, we must have this degree of channel wise quantization support.

Currently I am only take the first qparams. The following work is to support real channel wise/group wise quantization.

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -188,7 +215,138 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
...
+      offset = ptTensor.q_per_channel_zero_points()[0].item<int32_t>();
+    } else if (ptTensor.qscheme() == at::kPerTensorAffine) {
+      scale = static_cast<float>(ptTensor.q_scale());
+      offset = static_cast<int32_t>(ptTensor.q_zero_point());
+    } else {
+      LOG(DFATAL)
+          << "PyTorch tensor with unsupported quantization scheme detected.";
+    }
+    if (ptTensor.scalar_type() == at::kQUInt8) {
+      offset -= OFFSETSHIFT;
```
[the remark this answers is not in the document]
zrphercule: Ah you are right, the shift stuff should only be done in the converting of type.

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -188,7 +215,138 @@ at::Tensor glowTypeToEmptyPTTensor(const glow::Type &glowType) {
...
+  }
+}
+
+at::Tensor glowTensorToPTTensor(const glow::Tensor &glowTensor) {
+  std::vector<int64_t> sizes;
+  for (const auto dim : glowTensor.dims()) {
+    sizes.push_back(dim);
+  }
+  return at::from_blob(glowTensor.getUnsafePtr(), sizes,
+                       at::device(at::kCPU).dtype(at::kByte));
```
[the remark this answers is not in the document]
zrphercule: I will add a template here. The reason I cannot let it using type of glowTensor is we dont have kByte in glow now... after I add that I think I can do the auto type detecting here.

torch_glow/src/PyTorchCommon.h
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#ifndef USE_FBGEMM
```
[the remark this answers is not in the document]
zrphercule: @jackm321 It is the flag indicate we are using FBGEMM kernel in pytorch, which is used when unpack calling fbgemm functions.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: changed all clone to ensureIsOwned, and it works.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Yes it is, but in case it is stored as uint8 somewhere in other models... This wont cause any trouble right

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: agree

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: no it is not in CV model.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Think we should do it same to linear?

torch_glow/tests/nodes/quantized_linear_test.py
zrphercule: Yes we already have a random unit test under this. Think I can delete this.

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
jackm321: Ok, so basically these just aren't used in the channelwise case so they could be any value

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+      {input.dims()[0], outSz.first, outSz.second, weightShape.n}};
+  glow::TypeRef outTy = F_.getParent()->uniqueType(
+      glow::ElemKind::Int8QTy, outDims, outScale, outOffset);
+
+  if (isGroupwiseQuantized) {
+    RETURN_ERR_IF_NOT(dilation <= 1,
+                      "Dilation not supported for group quantized convolution");
+
+    // Get groupwise quantized params if needed
+    glow::Tensor wScalesTensor, wOffsetsTensor;
```
jackm321: These don't need to be created up here and then reassigned below

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+    wOffsets = F_.getParent()->createConstant("channel_wised_offsets_of_qconv",
+                                              std::move(wOffsetsTensor));
+    wOffsets->ensureIsOwned();
+
+    auto qconv = F_.createChannelwiseQuantizedConv(
+        "qconv_channel_wised", input, weightConstant, biasConstant, wScales,
+        wOffsets, outTy, kernels, strides, pads, groups);
+    if (isRelu) {
+      glow::ReluNode *qrelu = F_.createRELU("qconv_relu", qconv);
+      output = F_.createTranspose("channel_wised_qconv_relu_output_transposed",
```
jackm321: nit channelwise

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+                                  qrelu->getResult(), NHWC2NCHW);
+    } else {
```
jackm321: you can just have transpose as last thing here and remove this else

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+                                  qconv->getResult(), NHWC2NCHW);
+    }
+  } else {
+    auto qconv = F_.createConv("qconv", input, weight, bias, outTy, kernels,
+                               strides, pads, groups, dilation);
+    if (isRelu) {
+      glow::ReluNode *qrelu = F_.createRELU("qconv_relu", qconv);
+      output = F_.createTranspose("qconv_relu_output_transposed",
+                                  qrelu->getResult(), NHWC2NCHW);
+    } else {
```
jackm321: same thing, don't need this else case

torch_glow/src/PyTorchModelLoader.cpp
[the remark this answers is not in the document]
jackm321: Did you try to remove it though?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1212,20 +1459,71 @@ Error PyTorchModelLoader::loadQuantizedAddRelu(const torch::jit::Node *ptNode) {
...
+  const glow::Tensor &packedWeightTensor =
+      packedWeightConstant->getPayload().getHandle<bool>().clone();
+
+  auto ptTensor = glowTensorToPTTensor(packedWeightTensor, at::kByte);
+
+  auto unpackedParams = glow::fbgemm_linear_unpack(ptTensor);
+  const at::Tensor &ptWeightTensor = std::get<0>(unpackedParams);
+  const c10::optional<at::Tensor> &ptBiasTensor = std::get<1>(unpackedParams);
+
+  bool isRowwiseQuantized = ptWeightTensor.is_quantized() &&
```
jackm321: same thing, are both checks necessary or should we assert that one implies the other?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1242,14 +1540,66 @@ Error PyTorchModelLoader::loadQuantizedLinear(const torch::jit::Node *ptNode) {
                                           {input.dims()[0], weight.dims()[1]},
                                           outScale, outZeroPoint - OFFSETSHIFT);
 
+  // TODO Rowwise quantization is not enabled and accuracy is not tested
+  // We should enable it later.
+  isRowwiseQuantized = false;
```
jackm321: Since this isn't getting tested or used at all, I don't really see the point in putting this in this pr

torch_glow/src/PyTorchCommon.cpp
```diff
@@ -25,6 +25,8 @@
 #include <torch/csrc/jit/pass_manager.h>
 #include <torch/csrc/jit/passes/subgraph_rewrite.h>
 
+#include <ATen/core/op_registration/op_registration.h>
```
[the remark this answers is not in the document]
zrphercule: seems to be some ancient relic

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+      packedWeightConstant->getPayload().getHandle<bool>().clone();
+
+  auto ptTensor = glowTensorToPTTensor(packedWeightTensor, at::kByte);
+
+  auto unpackedParams = glow::fbgemm_conv_unpack(ptTensor);
+  const at::Tensor &ptWeightTensor = std::get<0>(unpackedParams);
+  const c10::optional<at::Tensor> &ptBiasTensor = std::get<1>(unpackedParams);
+
+  // if groups == 1 it is regular conv
+  bool isGroupwiseQuantized = ptWeightTensor.is_quantized() &&
```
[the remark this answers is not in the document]
zrphercule: It is possible to have a pt scheme == kPerChannelAffine while group == 1. Although when group > 1, pt scheme is always kPerChannelAffine.

torch_glow/src/PyTorchCommon.cpp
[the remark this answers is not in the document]
jackm321: ?

jackm321:
Please update the description to match the glow pr template and can you add a summary of next steps and things to clean up in there as well please (things like deleting the copied pt fbgemm code, enabling rowwise linear, etc)

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+    RETURN_ERR_IF_NOT(dilation <= 1,
+                      "Dilation not supported for group quantized convolution");
+
+    // Get groupwise quantized params if needed
+    glow::Tensor wScalesTensor, wOffsetsTensor;
+    glow::Constant *wScales, *wOffsets;
+
+    // extract qparams from ptWeightTensor.
+    // Notice since the memory of qparams may not be continous
+    // we CANNOT use the data ptr of this chunk of memory and
```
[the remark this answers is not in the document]
zrphercule: I tried to call that at first, but it was crashed last week. I remember there are some known problem about this. Let me check if it is still there.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: In QuantizedLinear, we dont have group so we will only use this to indenty a channel wise quantized tensor.
Also, both of them are needed.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: ok I can remove it ;)

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1174,200 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
   }
 }
 
+Error PyTorchModelLoader::getQuantizedConvTransposeNode(
```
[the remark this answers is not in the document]
zrphercule: Not making many difference?... using ASSIGN_VALUE_OR_RETURN_ERR or RETURN_IF_ERROR

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: @jackm321 No I cannot actually, since one is qrelu and one is qconv, they have different type.

torch_glow/src/PyTorchModelLoader.cpp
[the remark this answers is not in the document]
jackm321: The reference to pointer thing is kinda weird though imho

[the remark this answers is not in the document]
zrphercule: Okay I will change it to Expected

torch_glow/src/PyTorchModelLoader.cpp
jackm321: You can just use their output NodeValue

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Please and thank you

torch_glow/src/PyTorchModelLoader.cpp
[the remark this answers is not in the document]
zrphercule: Wont dont this in this pr, will do it in a seperate one once I figure out why contiguous lead to crash.

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -3006,16 +3331,24 @@ Error PyTorchModelLoader::loadAttributes(
...
-      const auto &ptTensor = ival.toTensor();
-      auto glowTensor = ptTensorToGlowTensor(ptTensor);
-
-      glow::Constant *glowConstant = F_.getParent()->createConstant(
-          newNameHierarchy, std::move(glowTensor));
+      const auto ptTensor = ival.toTensor();
+      // PyTorch Tensor extracted type is kByte
+      // indicate it is the address of stored weights of quantized
+      // linear or conv.
+      if (ptTensor.scalar_type() == at::kByte) {
```
yinghai: You can check its meta data to make sure it's a specific hacked tensor by doing 
https://bddppq.github.io/codebrowser/pytorch/pytorch/aten/src/ATen/cpp_custom_type_hack.h.html#20

But it's fine for now.

zrphercule: yes, the meta type that custom_hack check is a type inside fbgemm's header, which require us to call fbgemm.h again. I think this should work for now, eventually I should thinking of having a registered metatype indicating this hacked type, or look its following node to make sure it is weight of conv/linear.

torch_glow/src/GlowIValue.h
```diff
@@ -138,12 +141,22 @@ class GlowIValue {
   /// \returns Payload a vector of GlowIValues or error if the tag is not Tuple.
   Expected<const std::vector<GlowIValue> *> toTuple() const;
 
+  /// \returns Payload a PyTorch Tensor* or error if the tag is not Tensor.
```
jackm321: nit: not a PyTorch tensor

torch_glow/src/PyTorchModelLoader.h
```diff
@@ -279,6 +279,13 @@ class PyTorchModelLoader {
   /// Rescale a int8 NodeValue \p input to the equivalent uint8 NodeValue.
   glow::NodeValue rescaleIntToUint(glow::NodeValue input);
 
+  /// Load a quantized conv node from ptNode to qconv.
+  /// a wrapper function of loadQuantizedConv and loadQuantizedConvRelu.
+  /// Returns error on failure.
+  Expected<NodeValue>
+  getQuantizedConvTransposeNode(const torch::jit::Node *ptNode,
```
jackm321: can we name this some thing like `loadQuantizedConvImpl` so it's clear what its doing

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1171,180 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+  auto unpackedParams = callOp(*op, *ptTensor);
+  const at::Tensor ptWeightTensor = unpackedParams[0].toTensor();
+  const c10::optional<at::Tensor> ptBiasTensor =
+      unpackedParams[1].toOptional<at::Tensor>();
+
+  // if groups == 1 it is regular conv
+  bool isGroupwiseQuantized = ptWeightTensor.is_quantized() &&
+                              ptWeightTensor.qscheme() == at::kPerChannelAffine;
+  if (isGroupwiseQuantized) {
+    isGroupwiseQuantized &= (groups > 1);
```
jackm321: isn't this the same without the if?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1121,6 +1171,180 @@ glow::NodeValue PyTorchModelLoader::rescaleIntToUint(glow::NodeValue input) {
...
+  }
+  // unpacked weights
+  auto weightTensor = ptTensorToGlowTensor(ptWeightTensor);
+  glow::Tensor weightTensorTransposed;
+  weightTensor.transpose(&weightTensorTransposed, NCHW2NHWC);
+  glow::Constant *weightConstant = F_.getParent()->createConstant(
+      "quantized_conv2d_weights", std::move(weightTensorTransposed));
+  weightConstant->ensureIsOwned();
```
jackm321: I'm curious if these are still needed now that pt is doing the unpacking?

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: you are right, this is some history stuff.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: You are right, now it is fine.

PR #3737, opened 2019-11-05
Quantized FC node

jackm321:
Summary:
Testing to see if matmul -> ba (difference in order of bias addition and requantization) can account for the small differences from pytorch quantized linear. It seem that when bias and scales are small, this is in fact much more accurate but is more subject to rounding differences when scales are large.

Documentation:
doxygen

Test Plan:
unit tests

### DIFF
--- a/torch_glow/tests/nodes/quantized_linear_test.py
+++ b/torch_glow/tests/nodes/quantized_linear_test.py
@@ -5,14 +5,14 @@ import torch
 from tests.utils import jitVsGlow
 
 
-def test_quantized_linear():
+def test_quantized_linear_random_input():
     """Basic test of the PyTorch quantized::linear Node on Glow."""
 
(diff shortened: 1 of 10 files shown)


torch_glow/tests/nodes/quantized_linear_test.py
```diff
@@ -5,39 +5,14 @@
 from tests.utils import jitVsGlow
 
 
-def test_quantized_linear():
```
jfix71: For my own curiosity, why remove this test?

jackm321: This was basically a placeholder test for linear with some constant inputs that gave accurate enough results. With the added FC node, the accuracy is good enough that we can switch to testing with a bunch of random inputs which is a better test.

PR #3739, opened 2019-11-05
Relax single-use constant restrictions

bertmaher:
Summary: In a few places we check that a constant must have one use before performing some optimization on it.  This means that we get unintuitive results if we try to clone a function and optimize the clone, because some optimizations won't kick in.  I think the intent is to avoid increasing the model's memory footprint, but it seems like a heavy hammer to never allow.

Test Plan: Modifications to some existing GraphOptzTests.  I noticed this behavior when reviewing #3731 and wanted to get to the bottom of it.


jackm321:
It seems like we should make tensors, or at least constants, ref counted and then we can have actual COW for situations like this.

PR #3740, opened 2019-11-06
[GraphOptimizer] Change pipeline so we only run the backend's preferred optimizations at the end of the pipeline

jfix71:
Summary: Prior to this PR, we ran the backend's preferred pipeline before and after backend transformation (via `Backend::transformPostLowering()`). However the backend's preferred pipeline may be pretty minimal. Let's use our default pipeline after the final round of lowering instead.


opti-mix:
LGTM

PR #3745, opened 2019-11-06
Device Resident Tensors - API & Framework

nickgg:
Summary: Taking over #3671, but spinning out the API and Glow-core level changes associated with the DRT plan in #3629. This does not implement DRT support on any device.

Documentation: See #3629.

Test Plan: Ran tests, added two simple new sanity checks to DeviceManagerTest. The first `DeviceResidentTensors` should run only for backends that support resident tensors (none currently). The second `CanHandleDeviceResidentTensors` should run on all devices.

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -17,6 +17,7 @@
 #define GLOW_BACKENDS_DEVICEMANAGER_H
 
 #include "glow/Backend/CompiledFunction.h"
+#include "glow/Base/DeviceTensorTransferManager.h"
 #include "glow/ExecutionContext/ExecutionContext.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Runtime/RuntimeTypes.h"
@@ -42,7 +43,7 @@ using ReadyCBTy = std::function<void(const Module *, Error)>;
 using FunctionMapTy = std::map<std::string, CompiledFunction *>;
 
 /// Interface managing a specific instance of a device.
-class DeviceManager {
+class DeviceManager : public DeviceTensorTransferManager {
 protected:
   /// Configuration object for the device.
   DeviceConfig config_;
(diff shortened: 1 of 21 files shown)


nickgg:
Thats annoying I ran the format.sh script and still get clang-format errors :(

include/glow/Base/Tensor.h
```diff
@@ -48,6 +49,66 @@ void genericTranspose(const Tensor *src, Tensor *dest,
...
+  DeviceResidencyInfo(DeviceResidencyInfo &&other) = delete;
+
+  /// Move assignment operator.
+  DeviceResidencyInfo &operator=(DeviceResidencyInfo &&other) = delete;
+
+  ~DeviceResidencyInfo() {
+    // If a tensor is device resident, let its device manager free the device
+    // buffer.
+    if (isDeviceResident()) {
+      ((DeviceTensorTransferManager *)deviceManager_)
```
opti-mix: Why do you need this cast? `deviceManager_` is declared as `DeviceTensorTransferManager *deviceManager_`.

include/glow/Base/Tensor.h
```diff
@@ -48,6 +49,66 @@ void genericTranspose(const Tensor *src, Tensor *dest,
...
+}
+
+class DeviceResidencyInfo final {
+  enum class TensorResidency {
+    Host,
+    Device,
+  };
+
+  /// The residency status of the tensor.
+  TensorResidency tensorResidency_{TensorResidency::Host};
```
opti-mix: Nit: You may want to move the declaration of `tensorResidency_` below `deviceManager_` and `deviceManager_` pointers. Due to the alignments rules for class/struct fields, this may reduce the size of `DeviceResidencyInfo` objects. I.e. `sizeof(DeviceResidencyInfo_with_suggested_changes)` will be very likely < `sizeof(DeviceResidencyInfo)`.

include/glow/Base/Tensor.h
```diff
@@ -48,6 +49,66 @@ void genericTranspose(const Tensor *src, Tensor *dest,
 /// returned dims. For example, input {2,1,4} would result in {2,1,4,1,1,1}.
 ShapeVector expandDimsToMax(llvm::ArrayRef<size_t> currDims);
 
+namespace runtime {
+class DeviceManager;
+}
+
+class DeviceResidencyInfo final {
```
opti-mix: Please add a Doxygen comment.

include/glow/Base/Tensor.h
```diff
@@ -48,6 +49,66 @@ void genericTranspose(const Tensor *src, Tensor *dest,
...
+  ~DeviceResidencyInfo() {
+    // If a tensor is device resident, let its device manager free the device
+    // buffer.
+    if (isDeviceResident()) {
+      ((DeviceTensorTransferManager *)deviceManager_)
+          ->releaseDeviceTensor(locationContext_);
+    }
+  }
+
+  void clear() {
```
opti-mix: Please add DoxyGen comments to this and all other new public methods.

include/glow/Base/Tensor.h
```diff
@@ -82,6 +147,48 @@ class Tensor final {
   char *getData() const { return data_; }
 
 public:
+  /// \returns true if tensor data is stored on a device
+  bool isDeviceResident() const { return residencyInfoP_->isDeviceResident(); }
+
+  // Update device residency info with new device manager and context
+  void moveToDevice(DeviceTensorTransferManager *deviceManager,
+                    void *locationContext) {
+    residencyInfoP_->deviceManager_ = deviceManager;
```
opti-mix: Nit: Is it performance critical to keep the implementations of this and many other device residency-related methods in the header file so that it can be inlined? If not, I'd suggest moving them into a cpp implementation file. `Tensor.h` is used virtually everywhere and changing anything in it results in recompiling pretty-much everything. Thus, anything that can be kept out of it without any negative performance implications, should be better kept out of it.

nickgg:
Thanks for the review @opti-mix - will fix those.

nickgg:
I'm going to switch the Transfer* api to be asynchronous now rather than in a follow up.

include/glow/Backends/DeviceManager.h
```diff
@@ -152,6 +153,35 @@ class DeviceManager {
   /// \returns the DeviceInfo for this device containing peak limits for
   /// compute and bandwidths (used in partitioning).
   virtual DeviceInfo getDeviceInfo() const { return DeviceInfo(); }
+
+  /// Copies the contents of \p tensor from the host to the \p location
+  /// address on this device. Updates the tensor residency info.
+  virtual bool transferToDevice(
```
gcatron: Do we need to return a bool here or can this be void? It seems with the callback returning the error the bool is not needed.

nickgg: void should be fine now, good point

PR #3746, opened 2019-11-07
Add error handling to benchmark tests

jsubag:
Summary: The benchmark tests don't check a potential error return during execution.
This asserts in debug on unchecked error object


jsubag:
@jfix71 Here's a small fix for the benchmarks.

PR #3747, opened 2019-11-07
[Dump DAG](trivial) Use the exact CLI file name for single function.

mciprian13:
**Summary**
When the module has a single function, even if the CLI parameter is given as (for example) 

> -dump-graph-DAG=graph.dot

the generated file name is "_graph.dot" which is not clean (pre-pending the function name makes sense when multiple functions are in the module).

Documentation/test:
Not required.

### DIFF
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -516,8 +516,9 @@ void Loader::compile(CompilationContext &cctx) {
     for (auto function : module->getFunctions()) {
       std::string filename =
           function->getFilename() + "_" + dumpGraphDAGFileOpt;
-      if (module->getFunctions().size() == 1)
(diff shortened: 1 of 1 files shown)


tools/loader/Loader.cpp
```diff
@@ -516,6 +516,8 @@ void Loader::compile(CompilationContext &cctx) {
     for (auto function : module->getFunctions()) {
       std::string filename =
           function->getFilename() + "_" + dumpGraphDAGFileOpt;
+      if (module->getFunctions().size() == 1)
```
jfix71: nit: Our style is generally to always use curly braces even on single line ifs.

mciprian13: Done.

PR #3749, opened 2019-11-07
Add path for only lowering a Function during optimizations, and use during GraphOptzTest numerical correctness

jfix71:
Summary: For the numerical correctness checks for GraphOptzTest, we compile both optimized and unoptimized Function. This meant that the unoptimized Function was actually being optimized still, meaning the numerical test was not really doing much. I've added an optimization option to only lower so we can still run the Function but skip all optimizations (either the Function has already been optimized, or we don't want it to be optimized).


SplitInfinity:
> Test Plan: All tests still pass.

PHEW.

jfix71:
@SplitInfinity Checked and it turns out we do call `optimizeFunctionBeforeLowering()` inside the HostManager, so I added an early exit there too. Manually verified for `optimizeBatchNormAfterConv` that it's working as expected.

jfix71:
Whoops closed early.

PR #3751, opened 2019-11-07
[LIBJIT][Do not commit][Debate] Remove heap usage in LIBJIT

mciprian13:
**Summary**

This PR should not be landed but it should initiate a discussion (a more elaborate solution). The problem is this:
- currently in LIBJIT the kernel **libjit_matmul_f()** uses dynamically memory allocation via the **libjit_aligned_malloc()** which for a Linux host maps to the **posix_memalign()** function

Allowing kernels in LIBJIT to dynamically allocate memory has a couple of problems when considering cross-compiling bundles **(AOT),** especially for MCU targets:
- The bundle would have **posix_memalign()** as an undefined symbol which will not be resolved by target toolchains
- Although the application code could define the function **posix_memalign()** with a custom implementation such that the bundle will work, allowing LIBJIT code to use in an uncontrolled manner dynamic memory allocation raises some issues, especially for low-end/embedded devices like MCUs, where the heap is commonly pre-configured (before run-time) to a fixed and relatively limited size. Using heap in an unpredictable manner calls for unwanted situations: debugging and figuring out that heap size was exceeded, reconfigure the linker to increase the heap size (by how much? we can`t set a size too big because we starve the rest of the RAM ... etc).

The great/elegant thing for the Glow bundles is (or should be) that, apart from the stack usage (which should be decent given a serial sequence of function calls), the memory usage is completely transparent. Without even running the bundle I can known right after I compile the bundle:
- the EXACT data memory requirement by inspecting the auto-generated header file
- the program memory requirement (or approximately) by looking at the size of the bundle object: this is commonly not significant but it is great that Glow has such a small program memory footprint given the AOT approach (since it contains only the necessary kernel code and not all the kernel code like the Tensor Flow Lite run-time does for example -> one of the reasons I love Glow)

This PR contains an example of how we can avoid using heap for the mentioned LIBJIT kernel (just an example, code is very dirty) by providing the required scratch size through the kernel interface, scratch size which becomes an operand of the Glow IR Instruction and allocated through "createAllocActivation" (I noticed that the **TopK** node does this similarly).

Also this has been a recurrent issue when specializing some LIBJIT kernels to some optimized ARM libraries: some kernels (mostly the big operators: Convolution, FullyConnected) required scratch so I provided in the ways described here. This means an elegant solution for this would benefit developers in general. 

As a conclusion, heap usage in LIBJIT should be discouraged.
Pros:
- more transparency & predictability for memory usage (essential for memory constrained devices)

Cons:
- slightly harder to implement than a simple **malloc**
- scratch size requirement is defined in a separate place in the code than the kernel implementation

**Side notes**:
- I am not familiar with the matmul implementation but I am not sure the allocated size is correct. Should not be like this:
   libjit_aligned_malloc((void **)&packedB, 64, kc * nc * **sizeof(float)**);
- Are there testcases for MatMul with big sizes such that the code branch with dynamic memory allocation ("packed") is tested?
- I think semantically the correct operand type for scratch would be **OperandKind::OutIn** (which I don`t think is currently defined in Glow IR)
- Maybe we can define a "Scratch" operand by default for all the instructions (through ClassGen) and make the kernel somehow specify the exact size ...

What do you think?

### DIFF
--- a/lib/Backends/CPU/libjit/libjit_defs.h
+++ b/lib/Backends/CPU/libjit/libjit_defs.h
@@ -125,13 +125,13 @@ inline int32_t libjit_scale_i32i8(int32_t input, int32_t pre, int32_t post,
   return ((((input >> pre) * scale) + rtn) >> post) + offset;
 }
 
-#ifdef _WIN32
-#define libjit_aligned_malloc(p, a, s)                                         \
-  (((*(p)) = _aligned_malloc((s), (a))), *(p) ? 0 : errno)
-#define libjit_aligned_free(p) _aligned_free(p)
-#else
-#define libjit_aligned_malloc(p, a, s) posix_memalign(p, a, s)
-#define libjit_aligned_free(p) free(p)
-#endif
+//#ifdef _WIN32
+//#define libjit_aligned_malloc(p, a, s)                                         \
+//  (((*(p)) = _aligned_malloc((s), (a))), *(p) ? 0 : errno)
+//#define libjit_aligned_free(p) _aligned_free(p)
+//#else
+//#define libjit_aligned_malloc(p, a, s) posix_memalign(p, a, s)
+//#define libjit_aligned_free(p) free(p)
+//#endif
 
 #endif // GLOW_BACKENDS_CPU_LIBJIT_LIBJIT_DEFS_H
--- a/lib/Backends/CPU/libjit/libjit_matmul.cpp
+++ b/lib/Backends/CPU/libjit/libjit_matmul.cpp
@@ -224,11 +224,12 @@ void libjit_matmul_inner(int m, int n, int k, const float *a, int lda,
 template <bool pack>
 void __attribute__((noinline))
 libjit_matmul_outer(size_t m, size_t n, size_t k, const float *a, size_t lda,
-                    const float *b, size_t ldb, float *c, size_t ldc) {
-  float *packedB = nullptr;
-  if (pack) {
-    libjit_aligned_malloc((void **)&packedB, 64, kc * nc);
-  }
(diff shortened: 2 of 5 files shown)


opti-mix:
@mciprian13 The overall idea of using scratch instead of using a dynamic memory allocation is nice. But I'm not sure that your proposed solution introduced scratch at the right level of abstraction. You introduce a new scratch field at the instruction level and you do it in a backend-independent way. It may be too generic as backends which are not libjit-based would not use it.

mciprian13:
I understand this proposed implementation cannot be enforced upon all instructions/functions using scratch.
Do you have an idea about how can we do it in a more generic way?
If not, at least can we use this implementation up to the point when we figure out how to do it in a smarter way?

bertmaher:
Maybe we should just kill the huge-matrix packing code?  It really only matters for extremely huge matrices (like 4K*4K elements), and frankly libjit_matmul is only a mediocre GEMM anyways.

mciprian13:
@opti-mix What do you say about the proposal of @bertmaher ?

opti-mix:
> What do you say about the proposal of @bertmaher ?

I think I agree with this proposal. Let's remove the complexity of dynamic heap allocations for now. Should we work in the future on optimizing the GEMM operations again, we know where to look for our previous attempts ;-)

mciprian13:
Ok so I will remove that branch of the Gemm. Should I do this as part of this PR or a new PR for cleanliness?
Thanks!

opti-mix:
> Should I do this as part of this PR or a new PR for cleanliness?

Yes, let's do a separate PR.

mciprian13:
Solved in #3857.

PR #3758, opened 2019-11-08
Fix/enable partial tensor support for NNPI

jfix71:
Summary: Enable partial tensors for NNPI. This included moving out some code from Habana into BackendUtils so it can be reused.

Differential Revision: D18403972

### DIFF
--- a/lib/Backends/NNPI/InferencePool.cpp
+++ b/lib/Backends/NNPI/InferencePool.cpp
@@ -68,8 +68,12 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
   PlaceholderBindings &bindings = *ctx->getPlaceholderBindings();
   ioTensors_.clear();
 
+  std::unordered_set<Tensor *> partialTensorInputs;
   for (auto &pht : bindings.pairs()) {
     ioTensors_.emplace(pht.first->getName(), pht.second);
+    if (partialInputs_->count(pht.first)) {
+      partialTensorInputs.insert(pht.second);
+    }
   }
 
   // Handle inputs & outputs (+convert).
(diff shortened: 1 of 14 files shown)


jfix71:
@jsubag @bertmaher Can you guys take a look here to make sure everything checks out?

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -79,22 +83,44 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
...
-      rawInputs_.push_back(tmp);
+      bufferPtr = reinterpret_cast<char *>(tmp);
+      tmpBuffers_.insert(bufferPtr);
+    }
+
+    // If this tensor cannot be treated as partial but it is partial, then we
+    // need to make a copy of the underlying buffer to the correct unpadded
+    // size, and zero out anything that's leftover in the extra padding, as it
+    // is still going to be used.
+    if (!partialTensorInputs_.count(t) && paddedSize != unpaddedSize) {
```
jsubag: Slightly more performant to compare the integers first then move on to searching the set.

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -79,22 +83,44 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
...
+      char *tmp = new char[paddedSize];
+      // Copy over original data.
+      memcpy(tmp, bufferPtr, unpaddedSize);
+      // Zero out rest of the new buffer.
+      memset(tmp + unpaddedSize, 0, paddedSize - unpaddedSize);
```
jsubag: Zero isn't always the idempotent value (e.g. 1 for ReduceMul).
You can extend the mechanism in the util function to also provide the correct value to pad with.
Not sure if there's value in that for your usage and either way can be added later.

lib/Backends/NNPI/NNPICompiledFunction.h
```diff
@@ -53,6 +53,11 @@ class NNPICompiledFunction final : public CompiledFunction {
   /// \returns the compilation config object.
   NNPICompilationConfig getCompilationConfig() const { return config_; }
 
+  /// \returns a reference to the set of Placeholders supporting partial inputs.
+  std::unordered_set<const Placeholder *> &getPartialInputs() {
```
jsubag: This can be a const function and return a const set&

lib/Backends/NNPI/InferencePool.cpp
jsubag: This code will still copy only the unpaddedSize to the device.
You should update the check in line 152|178  `if (bufferSize != fullBufferSize)` to also consider the partialTensorInput set.

jsubag:
@jfix71 I've added a few comments.
One condition that needs amending and a couple of suggestions.
Other that those - LGTM.

lib/Backends/NNPI/InferencePool.cpp
jsubag: Alternatively, if the desired behavior is to always use 0 for padding then the memset can happen above the backend (shared by all backends) and that tensor can get its unpaddedSize set to the full size at that point - thus removing the need for any special handling in the backends.

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -79,22 +83,44 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
...
+    // If this tensor cannot be treated as partial but it is partial, then we
+    // need to make a copy of the underlying buffer to the correct unpadded
+    // size, and zero out anything that's leftover in the extra padding, as it
+    // is still going to be used.
+    if (paddedSize != unpaddedSize && !partialTensorInputs.count(t)) {
+      char *tmp = new char[paddedSize];
+      // Copy over original data.
+      memcpy(tmp, bufferPtr, unpaddedSize);
+      // Zero out rest of the new buffer.
+      memset(tmp + unpaddedSize, 0, paddedSize - unpaddedSize);
```
arunm-git: Given that there are two separate/independent checks (int64 -> int32? and 'treat as partial'), there's probably a way to code this up where we only need to do 1 memory allocation for the case when both of those checks are true. Not clear to me if the savings is worth the effort/code complexity.

jfix71: Good idea, done -- don't think the extra logic is so bad.

jsubag:
LGTM.

PR #3761, opened 2019-11-08
use legacy executor for glow [WIP]

Krovatkin:
Summary:

Forces Glow to use the legacy GraphExecutor as pytorch jit is enabling to the new infrastructure by default.

Documentation:

https://github.com/pytorch/pytorch/pull/29230

Test Plan:


Krovatkin:
@jackm321 looks like it's passing except for the clang-format check, which should be fixed now. Let me know if you would want me to put this logic elsewhere.

PR #3763, opened 2019-11-08
[Operator] Add SparseLengthSumGradNode and Inst

SplitInfinity:
**Summary**
This commit adds `SparseLengthsSumGradNode` and `Inst`. `SparseLengthsSum` is currently
lowered to `SparseLengthsWeightedSum` with all weights set to 1, but some backends
might want to opt out of this lowering. Those backends will then need
`SparseLengthsSumGradNode` and the corresponding `Inst`.

**Test Plan**
Not much to test in this commit but all unit tests pass.


navidR:
I haven't abled to track down the issue. But this pr completely breaks my build. Does anybody have the same issue?

SplitInfinity:
@navidR 

Can you include any useful excerpts of the error messages you see?

PR #3764, opened 2019-11-09
[PyTorch] load aten::stack

jackm321:
Summary:
See title.
Refactor aten::PrimConcat fusion code for use with aten::stack.

Documentation:
doxygen

Test Plan:
new unit test

### DIFF
--- a/torch_glow/src/FuseKnownPatterns.cpp
+++ b/torch_glow/src/FuseKnownPatterns.cpp
@@ -45,44 +45,6 @@ void registerDummyOperator(const char *opName) {
       options)});
 }
 
-/// To judge if we can fuse the current node into a prim::FusedConcat.
-bool isFusableConcatNode(torch::jit::Node *node) {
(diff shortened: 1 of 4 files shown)


torch_glow/src/FuseKnownPatterns.cpp
```diff
@@ -337,6 +326,7 @@ void fuseKnownPatterns(std::shared_ptr<torch::jit::Graph> &graph) {
   std::call_once(onceFlag, []() {
     registerDummyOperator("glow::unpacked_quantized_linear");
     registerDummyOperator("glow::unpacked_quantized_conv2d");
+    registerDummyOperator("glow::FusedStack");
```
yinghai: Inconsistent naming

jackm321: Yeah, I was trying to name it like prim::FusedConcat but I'll change it

PR #3765, opened 2019-11-09
Add dynamic buffer support to OCL Backend

nickgg:
Summary: The OpenCL Backend uses a static memory allocation strategy of allocating a single large buffer and then using offsets into it, which is good for the general case, but doesn't allow us to get the most benefit out of Device Resident Tensors (when we'd like to leave an output on the device to be used as the input to another network). This PR adds a more dynamic mapping of device buffers to the OCL backend via OpenCL SubBuffers, which are similar to Glow TensorViews in that they provide access to a region without additional allocations.

There is no behavioural change in this PR, but it provides infrastructure to reference buffers outside of the range of the DeviceBuffer in the future, which we need to get DRT perf wins.

The immediate benefit is that I was able to simplify the OCL kernel code, deleting about 25% of kernels.cl.

Documentation:  NFC

Test Plan: tests in release and ASAN

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -108,6 +108,10 @@ static void addStringOption(std::vector<std::string> &options,
   options.push_back("-D" + name + "=" + value);
 }
 
+cl_mem glow::runtime::OpenCLDeviceBindings::getBuffer(glow::Value *v) {
+  return weightBuffers[v->getName()];
+}
+
 OpenCLFunction::OpenCLFunction(std::unique_ptr<IRFunction> F,
                                runtime::RuntimeBundle &&bundle,
                                TraceInfo traceInfo)
@@ -140,15 +144,15 @@ static std::string getKernelName(const char *baseName, ElemKind elemTy) {
   std::string name = baseName;
   switch (elemTy) {
   case ElemKind::FloatTy:
-    return name + "W";
+    return name + "K";
   case ElemKind::Int8QTy:
-    return name + "_i8W";
+    return name + "_i8K";
   case ElemKind::Int32QTy:
-    return name + "_i32W";
+    return name + "_i32K";
   case ElemKind::Int64ITy:
-    return name + "_uW";
+    return name + "_uK";
   case ElemKind::BoolTy:
-    return name + "_bW";
+    return name + "_bK";
   default:
     LOG(FATAL) << "Unsupported data type: "
                << Type::getElementName(elemTy).str();
@@ -218,14 +222,14 @@ static void setKernelArg(cl_kernel kernel, unsigned argIdx, T value) {
 ///
 /// \returns the index of the last set OpenCL kernel argument.
 static size_t setKernelArgsForBuffers(cl_kernel kernel, const Instruction &I,
-                                      size_t nextKernelArgIdx,
+                                      runtime::OpenCLDeviceBindings *clBindings,
                                       runtime::RuntimeBundle &bundle) {
   // Number of instruction operands.
   auto numArgs = I.getNumOperands();
   // The predicate of the instruction if available.
   Value *predicate = I.hasPredicate() ? I.getPredicate() : nullptr;
   // The index of the kernel argument to be set.
-  unsigned kernelArgIdx = nextKernelArgIdx;
+  unsigned kernelArgIdx = 0;
   // Go over all operands and pass buffer operands to the kernel.
   for (unsigned arg = 0; arg < numArgs; arg++) {
     auto *value = I.getOperand(arg).first;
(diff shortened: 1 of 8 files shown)


nickgg:
Perf should be neutral, but I'll run some tests with image-classifier and attach the results.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -367,6 +375,7 @@ void OpenCLFunction::enqueueKernel(llvm::StringRef name,
                                profile ? &event : nullptr);
   CHECK_EQ(err, CL_SUCCESS) << "Error in clEnqueueNDRangeKernel.";
   kernelLaunches.push_back(KernelLaunch(kernel, name, kernelType, event));
+  CL_INVALID_ARG_INDEX;
```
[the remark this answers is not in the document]
nickgg: Yes! It took me a long time to debug the last issue and i must have undone past where I deleted this.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -992,8 +1008,17 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
       // This is a naive implementation that parallelizes using three dims:
       // the X and the Y in the output filter.
       cl_kernel kernel = createKernel(kernelName, program);
-      setKernelArg(kernel, 0, deviceBuffer);
-      auto numArgs = setKernelArgsForBuffers(kernel, I, 1, runtimeBundle_);
+      unsigned numArgs = 0;
+      setKernelArg(kernel, numArgs++, deviceBuffer);
```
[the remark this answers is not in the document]
nickgg: In this case it looks like it's because I made the deviceBuffer the first argument again. I actually had a lot of trouble making this kernel work correctly and vaguely remember this being the only thing that worked. Will look into it.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -495,26 +504,25 @@ void OpenCLFunction::executeNCHWConvolution(
 
   auto kernelName = isQuantized ? "conv_forward_mem_i8" : "conv_forward_mem";
   auto kernel = createKernel(kernelName, prog);
-  setKernelArg(kernel, 0, devBindings->deviceBuffer);
-  setKernelArg<cl_uint>(kernel, 1, runtimeBundle_.getValueOffset(input));
-  setKernelArg<cl_uint>(kernel, 2, runtimeBundle_.getValueOffset(weights));
-  setKernelArg<cl_uint>(kernel, 3, runtimeBundle_.getValueOffset(bias));
-  setKernelArg<cl_uint>(kernel, 4, runtimeBundle_.getValueOffset(output));
+  setKernelArg(kernel, 0, devBindings->weightBuffers[input->getName()]);
```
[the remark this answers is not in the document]
nickgg: Yeah good idea, will fix.

nickgg:
Perf comparison:

before:
image

after:
image


Looks neutral to me.

opti-mix:
> For #2 I agree, was thinking the same thing while doing the work. Figured this diff was big enough as it is, follow up?

Yes, at least file an issue about it, so that we do not forget.

opti-mix:
@nickgg Thanks for checking the performance. Looks like the change is neutral, which is very good.

lib/Backends/OpenCL/OpenCL.cpp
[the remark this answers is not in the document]
nickgg: I've added a comment about this but basically if you remove the first `void*` arg from this kernel it doesn't compile. Why? No idea it should be fine, and all the other kernels were. This is a compromise.

nickgg:
Both test suite fails here look spurious, but it seems like I can't rerun.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1244,13 +1275,22 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
...
+      cl_mem srcBuf = clBindings->getBuffer(src);
       size_t sizeInBytes = dest->getSizeInBytes();
       cl_event event{nullptr};
-      cl_int err = clEnqueueCopyBuffer(commands, deviceBuffer, deviceBuffer,
-                                       srcOff, destOff, sizeInBytes, 0, nullptr,
-                                       kernelProfiling_ ? &event : nullptr);
+      cl_int err =
+          clEnqueueCopyBuffer(commands, srcBuf, destBuf, 0, 0, sizeInBytes, 0,
+                              nullptr, kernelProfiling_ ? &event : nullptr);
+      llvm::outs() << "COPY\n";
```
opti-mix: Please remove debug prints.

nickgg: just temporary I'm trying to printf debug the POCL build. will fix before landing

tests/unittests/MLTest.cpp
```diff
@@ -1376,6 +1376,7 @@ TEST_P(MLTest, testFindPixelRegression) {
     auto dx = LH.at({i, 0}) - RH.at({i, 0});
     auto dy = LH.at({i, 1}) - RH.at({i, 1});
     auto distance = std::sqrt(std::pow(dx, 2) + std::pow(dy, 2));
+    llvm::outs() << distance << "\n";
```
gcatron: Was this a debugging print statement?

nickgg: yup will remove as well

nickgg:
OK! Lint problems are from https://github.com/pytorch/glow/commit/fc64547d2061f0faf2d959868a935fb306ec2396 not this diff, OpenCL build is just the normal POCL issues, Pytorch broken in master. I'm going to land this if it kills me.

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: Did you notice POCL_DEBUG=1 which might be useful?

pjaaskel:
> ... OpenCL build is just the normal POCL issues, Pytorch broken in master. I'm going to land this if it kills me.

Just curious: does github.com/pocl/pocl/issues know about "the normal pocl issues" you refer to here?

For me Glow now works quite well with pocl, but I've one single liner patch I need to upstream to pocl due to the way Glow checks for platform existence with 0 device query which currently fails.
Are the other remaining issues?

PR #3766, opened 2019-11-09
[ImageClassifier](trivial) Allow profiling SSD models.

mciprian13:
**Summary**
Make the **image-classifier** tool to be capable to profile models which have an input image and multiple output placeholders, like the SSD models which commonly have two output placeholders: scores and boxes. One such SSD example can be found in ONNX format [here](https://github.com/qfgaohao/pytorch-ssd).

Further, using the model-compiler we can compile quantized SSD models (this tool already accepts any number of inputs or outputs).

Documentation:
None

Test Plan:
None


mciprian13:
Can we have this reviewed (given that it's relatively short)?

PR #3770, opened 2019-11-10
[Quantization](trivial) Update logic for computing scale of int8/int16 bias scale

mciprian13:
**Summary**
Use the same logic for computing the scale of int8/int16 as for int32: try to force the bias_scale to product input_scale * weights_scale if no saturation occurs. This has run-time benefits since the effective scale applied at run-time (which is bias_scale / (input_scale * weights_scale)) will be 1.0 resulting in biasPre = 0, biasPost = 0, biasScale = 1.

**Documentation**
None

**Test Plan**
None


mciprian13:
Can we have this landed?

PR #3772, opened 2019-11-11
Remove embedding save nodes in RecommendationSystemTest

jsubag:
Summary: Removing extra save nodes created for SLWS nodes for debugging.


jsubag:
@jfix71 Please take a look.

PR #3775, opened 2019-11-11
Add Backend::saveFunctions to store multiple Glow functions into a single bundle

### LINKED ISSUE #3228: Extend LLVMIRCodeGen to support compiling multiple Glow functions into a single LLVM Module or object file
A single `glow::Module` may contain multiple `glow::Functions`. It could happen e.g. when Glow uses partitioning or if we would introduce scheduling of subgraphs for parallel execution, etc.

It may be very useful to compile these multiple functions together into the same LLVM module or into the same object file. 

To achieve this, some changes in the LLVMIRCodeGen may be required.

Here are some possible LLVMIRCodeGen changes:
- When compiling an IRFunction, allow the LLVMBackend to re-use an already existing LLVM module if it is present. This would allow to add more functions to the same module.
- Or continue compiling each IRFunction into its own module, but use LLVM's [Linker APIs](https://llvm.org/doxygen/classllvm_1_1Linker.html) to link those modules together before starting any LLVM optimizations.
- BTW, the libjit bitcode is currently read/included into each LLVM module created for each IRFunction. This could be changed to create a separate LLVM module for the libjit and then link the LLVM module(s) produced from an IRFunction (or multiple IRFunctions) with it.
- And of course, one should take care of function names to avoid possible name conflicts. Each IRFunction should produce an LLVM::Function with a unique name. All other generated helper LLVM functions should be marked as local to the module, unless they are supposed to be shared by multiple modules (e.g. libjit functions are likely to have a linkage which allows to use them from other modules).

opti-mix:
Summary: Extend LLVMBackend and BundleSaver to support saving multiple Glow functions into a single bundle.

Detailed changes are:
- [LLVMIRCodeGen] Preserve the state in AllocationsInfo to enable allocation of multiple functions using the same AllocationsInfo object

- [LLVMIRCodeGen] Split finishCodeGen from performCodeGen to support compilation of multiple Glow IR functions into a single LLVM module

- [LLVMIRCodeGen] Extend FunctionSpecializer to support LLVM modules with multiple main entry points

- [LLVMIRCodeGen] Fix debug info emission for multiple functions inside the same bundle

- Add an example of an AOT bundle with multiple entry points

Fixes #3228

Test Plan:
ninja test

### DIFF
--- a/examples/bundles/bundle_with_multiple_entries/CMakeLists.txt
+++ b/examples/bundles/bundle_with_multiple_entries/CMakeLists.txt
@@ -1,0 +1,44 @@
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOW_BINARY_DIR}/bundles)
+
+# Output directories for regular and quantized outputs.
+set(BUNDLE_OUTPUT_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/bundle_with_multiple_entries)
+
+add_custom_target(BundleWithMultipleEntriesDir ALL
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${BUNDLE_OUTPUT_DIRECTORY}
+)
+
+# Final Executables.
+# =================
+# Regular.
+add_executable(bundle_with_multiple_entries $<TARGET_OBJECTS:bundle_with_multiple_entriesMain>)
+set_target_properties(bundle_with_multiple_entries PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${BUNDLE_OUTPUT_DIRECTORY})
+target_link_libraries(bundle_with_multiple_entries ${BUNDLE_OUTPUT_DIRECTORY}/testBundle.o png)
+add_dependencies(bundle_with_multiple_entries bundle_with_multiple_entriesMain bundle_with_multiple_entriesNet)
+
+add_executable(bundle_with_multiple_entriesBundleSaver BundleSaver.cpp)
+set_target_properties(bundle_with_multiple_entriesBundleSaver PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${BUNDLE_OUTPUT_DIRECTORY})
+target_link_libraries(bundle_with_multiple_entriesBundleSaver
+                        PRIVATE
+                          Backends
+                          Graph
+                          Support)
+# Glow Bundles.
+# ============
+# Regular Bundle.
+add_custom_command(
+  OUTPUT
+    ${BUNDLE_OUTPUT_DIRECTORY}/testBundle.o
+  COMMAND
+    ${BUNDLE_OUTPUT_DIRECTORY}/bundle_with_multiple_entriesBundleSaver -emit-bundle ${BUNDLE_OUTPUT_DIRECTORY} -bundle-api=dynamic -g
+  DEPENDS
+    bundle_with_multiple_entriesBundleSaver BundleWithMultipleEntriesDir
+)
+add_custom_target(bundle_with_multiple_entriesNet DEPENDS ${BUNDLE_OUTPUT_DIRECTORY}/testBundle.o)
+
+# Other.
+# =====
+# Driver program with main function for regular bundle.
+add_library(bundle_with_multiple_entriesMain OBJECT main.cpp)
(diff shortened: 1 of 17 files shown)


examples/bundles/bundle_with_multiple_entries/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOW_BINARY_DIR}/bundles)
+#set(LENET_MNIST_BUNDLE_DIR ${GLOW_BINARY_DIR}/examples/bundles/bundle_with_multiple_entries)
+
+# Output directories for regular and quantized outputs
```
shajrawi: . at end of comment ;-) - same thing throughout this cmake file

examples/bundles/bundle_with_multiple_entries/main.cpp
```diff
@@ -0,0 +1,177 @@
...
+  (void)res;
+  return ptr;
+}
+
+/// Initialize the constant weights memory block by loading the weights from the
+/// weights file.
+static uint8_t *initConstantWeights(const char *weightsFileName,
+                                    const BundleConfig &config) {
+  // Load weights.
+  FILE *weightsFile = fopen(weightsFileName, "rb");
```
shajrawi: Why not use LLVM's file open/write wrapper? i.e. raw_fd_ostream

lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -190,7 +190,73 @@ static void serializeBinaryToText(llvm::StringRef binFileName,
...
+bool BundleSaver::WeightAddrComparator::
+operator()(const WeightInfo &LHS, const WeightInfo &RHS) const {
+  auto lhsAddr =
+      bundleSaver_.allocationsInfo_.allocatedAddress_.lookup(LHS.first);
+  auto rhsAddr =
+      bundleSaver_.allocationsInfo_.allocatedAddress_.lookup(RHS.first);
+  return lhsAddr < rhsAddr;
+}
+
+std::set<BundleSaver::WeightInfo, BundleSaver::WeightAddrComparator>
```
shajrawi: not llvm's SetVector? :)

lib/LLVMIRCodeGen/FunctionSpecializer.cpp
```diff
@@ -417,8 +420,8 @@ class FunctionSpecializer {
     }
   };
 
-  /// The entry function of the module.
-  llvm::Function *entryF_;
+  /// The entry functions of the module.
+  std::vector<llvm::Function *> &entryFunctions_;
```
shajrawi: llvm smallvector instead?

examples/bundles/bundle_with_multiple_entries/main.cpp
opti-mix: IIRC, all of our bundles examples do not use any of LLVM APIs as they are examples of "standalone" apps.

lib/LLVMIRCodeGen/BundleSaver.cpp
opti-mix: Does `llvm::SetVector` support custom comparators?

shajrawi: errm good point: it is clunkier with LLVM: would probably need to do that via explicitly via overriding the default hash value / implement their hashing API. forget about this nit - i withdraw my comment :)

examples/bundles/bundle_with_multiple_entries/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${GLOW_BINARY_DIR}/bundles)
+#set(LENET_MNIST_BUNDLE_DIR ${GLOW_BINARY_DIR}/examples/bundles/bundle_with_multiple_entries)
```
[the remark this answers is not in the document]
opti-mix: Done.

lib/LLVMIRCodeGen/FunctionSpecializer.cpp
opti-mix: Done.

PR #3778, opened 2019-11-12
Add missing accumulator type parameter propagation for SLWS

jsubag:
Summary:
Fixed a missing layer parameter propagation for SLWS accumulation type during lowering.


jsubag:

PR #3779, opened 2019-11-12
v0.4.0.0 Update for the Intel NNPI Backend

jsubag:
Summary: v0.4.0.0 Update for the Intel NNPI Backend

### DIFF
--- a/lib/Backends/NNPI/BlockStream.cpp
+++ b/lib/Backends/NNPI/BlockStream.cpp
@@ -20,7 +20,7 @@
 namespace glow {
 
 BlockStream::BlockStream(size_t preallocSize, uint32_t pageBlockSize)
-    : readOffest_(0), writeOffest_(0), blockSize_(pageBlockSize) {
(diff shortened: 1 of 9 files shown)


jsubag:

arunm-git:
Hi Jacob,
Changes here look fine, but some general questions:
* I'm guessing this PR requires the 0.4.0 drop  to work? (i.e. - we can't take this PR in isolation without moving to the 0.4.0 drop)? If so, we'll wait on this PR till we update nnpi-sw internally.
* BlockStream.cpp: the changes in this PR are fine. While reading the code in that file, I noticed that there may be a bug in allocateBlocks() where we update 'available' and return that new value even when the allocation fails. Mentioning it so we can address that (in a different PR is fine), and also fix the typos (*Offest instead of *Offset).

jsubag:
CC: @omromano

jsubag:
@arunm-git You're correct on both points - good catch.
I'll update the PR shortly

PR #3780, opened 2019-11-12
[PyTorch] update Glow fuser

jackm321:
Summary:
* Add support for minimum fusion group size
* Only fuse nodes with non-tensor outputs if those outputs are only consumed by nodes that have already been fused
* Only fuse getattrs that have uses within the fusion group
* Add a verifier step after fusion to ensure that fusion requirements are met
* Added a debugging mechanism to call fuser on graphs directly from Python

Documentation:
doxygen

Test Plan:
Added unit tests
try fusion on resnet and bert

### DIFF
--- a/torch_glow/src/GlowFuser.cpp
+++ b/torch_glow/src/GlowFuser.cpp
@@ -32,7 +32,7 @@
 
 namespace glow {
 namespace {
-using isSupportFunc = std::function<bool(torch::jit::Node *)>;
+using IsSupportFunc = std::function<bool(const torch::jit::Node *)>;
 
 torch::jit::value_list
 sortReverseTopological(at::ArrayRef<torch::jit::Value *> inputs,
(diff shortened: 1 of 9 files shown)


torch_glow/src/GlowFuser.cpp
```diff
@@ -51,140 +49,274 @@ sortReverseTopological(at::ArrayRef<torch::jit::Value *> inputs,
...
+  }
 
-#define REQ(cond, ignore, log_info)                                            \
-  if (!(cond)) {                                                               \
-    if (!(ignore)) {                                                           \
-      DLOG(ERROR) << (log_info);                                               \
-    }                                                                          \
-    return c10::nullopt;                                                       \
+  // Check that the node is supported
+  if (!(fn(node) || node->kind() == torch::jit::prim::Constant ||
```
yinghai: We should merge the kind() to the first if?

torch_glow/src/GlowFuser.cpp
```diff
@@ -51,140 +49,274 @@ sortReverseTopological(at::ArrayRef<torch::jit::Value *> inputs,
...
 
-c10::optional<torch::jit::Node *> tryMerge(torch::jit::Node *consumer,
-                                           torch::jit::Node *producer,
-                                           torch::jit::AliasDb &aliasDb,
-                                           isSupportFunc fn, at::Symbol kind) {
+    // Producers can have non-tensor outputs as long as they are only consumed
+    // by consumer. Consumers cannot have non-tensor outputs.
+    if (consumer) {
+      for (auto use : output->uses()) {
+        if (use.user != consumer) {
```
yinghai: Why is this inequality?

torch_glow/src/GlowFuser.cpp
```diff
@@ -51,140 +49,274 @@ sortReverseTopological(at::ArrayRef<torch::jit::Value *> inputs,
...
-      "Detected unknown node: " + symbol_name_producer + ".\n");
-  REQ(consumer->kind() == kind || canMerge(consumer, fn),
-      isLogIgnoredOp(symbol_name_consumer),
-      "Detected unknown node: " + symbol_name_consumer + ".\n");
+  return true;
+}
 
+bool aliasChecks(torch::jit::Node *consumer, torch::jit::Node *producer,
+                 torch::jit::AliasDb &aliasDb) {
   // Alias checks
```
yinghai: Move this up to make it doxygen string?

torch_glow/src/GlowFuser.cpp
```diff
@@ -51,140 +49,274 @@ sortReverseTopological(at::ArrayRef<torch::jit::Value *> inputs,
...
+                        at::Symbol kind, IsSupportFunc fn) {
+  const auto &settings = getPyTorchLoaderSettings();
+  const auto blacklist = settings.opBlacklist;
+  const auto minFusionGroupSize = settings.minFusionGroupSize;
+
+  // Wrap fn in function that first checks the blacklist.
+  IsSupportFunc nodeSupportedFn = [bl = std::move(blacklist),
+                                   fn](const torch::jit::Node *ptNode) {
+    if (bl.count(ptNode->kind())) {
+      std::cout << "Skipping black list op kind: "
```
yinghai: Why cout?

jackm321:
> We should also start to enable all the tests internally.
I agree, do you have any cycles to do this?

yinghai:
nope.

torch_glow/src/GlowFuser.cpp
jackm321: I prefer this way because the first if is a special case we explicitly don't support, this if is the opposite, these are nodes we do support including some special cases

torch_glow/src/GlowFuser.cpp
jackm321: This is checking if the node is used anywhere outside of the fusion group

PR #3782, opened 2019-11-12
Add stack of return points for ErrorValues

jackm321:
Summary:
For easier debugging, instead of just reporting the place where an Error was created and the place where it was destroyed, report all of the places it was forwarded up the stack as well.

**before**
```
location: ../torch_glow/src/PyTorchModelLoader.cpp:1003 message: Did not find a NodeValue mapping for Value 1
```

**after**
```
Error message: Did not find a NodeValue mapping for Value 1
Error return stack:
../torch_glow/src/PyTorchModelLoader.cpp:1003
../torch_glow/src/PyTorchModelLoader.cpp:2892
../torch_glow/src/CachingGraphRunner.cpp:67
../torch_glow/src/CachingGraphRunner.cpp:133
```

Documentation:
doxygen

Test Plan:
create an error and try it

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -28,7 +28,6 @@
 /// abstraction between Error/Expected types and the specific classes that
 /// implement them.
 
-namespace glow {
 /// Consumes an Error \p err and \returns true iff the error contained an
 /// ErrorValue. Calls the log method on ErrorValue if the optional argument \p
 /// log is passed.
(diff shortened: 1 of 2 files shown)


include/glow/Support/Error.h
```diff
@@ -280,43 +284,44 @@ class GlowErrorValue final {
...
 
-  /// The line number the error was generated on.
-  size_t lineNumber_;
-  /// The name of the file the error was generated in.
-  std::string fileName_;
   /// Optional message associated with the error.
   std::string message_;
+  /// The filename and line number of all places that created, forwarded, and
+  /// destroyed the ErrorValue.
+  std::vector<std::pair<std::string, size_t>> stack_;
```
opti-mix: Quick question: Do we need to be concerned about the overhead of growing this vector and performing dynamic memory allocations? Or is it very low and we don't care?

jackm321: It's a good question, especially given that is code executing on error path.
I think it should be sufficiently small that this doesn't matter but if you know of a case where this could be an issue but it should actually be possible to avoid allocations here if it becomes a problem.

jackm321: As to the overheads specifically, allocations only happens when an Error actually occurs

PR #3786, opened 2019-11-15
[PyTorch] allow non-tensor inputs to arithmetic ops

jackm321:
Summary:
Allow float and int inputs to aten::add/sub/mul/div

Documentation:
doxygen

Test Plan:
added unit tests

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -1254,20 +1254,65 @@ Error PyTorchModelLoader::loadGlowFusedLinear(const torch::jit::Node *ptNode) {
   return addValueMapping(outputs[0], output);
 }
 
+Expected<NodeValue> PyTorchModelLoader::loadNodeValueOrBroadcastedConstant(
+    const torch::jit::Value *value, llvm::ArrayRef<size_t> dims) {
+  if (hasGlowNodeValueForValue(value)) {
+    return getGlowNodeValueForValue(value);
+  } else {
+    GlowIValue *ival;
+    ASSIGN_VALUE_OR_RETURN_ERR(ival, getGlowIValueForValue(value));
+    float constVal;
+    if (ival->isInt()) {
+      ASSIGN_VALUE_OR_RETURN_ERR(constVal,
+                                 static_cast_expected<float>(ival->toInt()));
+    } else {
+      ASSIGN_VALUE_OR_RETURN_ERR(constVal,
+                                 static_cast_expected<float>(ival->toDouble()));
+    }
+    glow::Tensor t(glow::ElemKind::FloatTy, dims);
+    t.init(glow::Tensor::InitKind::Broadcast, constVal,
+           F_.getParent()->getPRNG());
+    return F_.getParent()
+        ->createConstant("constant", std::move(t))
+        ->getOutput();
+  }
+}
+
+template <typename GlowNode>
+Expected<NodeValue>
+PyTorchModelLoader::loadArithmeticNode(llvm::StringRef name,
+                                       const torch::jit::Value *lhs,
+                                       const torch::jit::Value *rhs) {
+  glow::NodeValue lhsInput;
+  glow::NodeValue rhsInput;
+
+  if (hasGlowNodeValueForValue(lhs)) {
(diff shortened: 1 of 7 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1252,15 +1252,49 @@ Error PyTorchModelLoader::loadGlowFusedLinear(const torch::jit::Node *ptNode) {
...
+    if (ival->isInt()) {
+      ASSIGN_VALUE_OR_RETURN_ERR(constVal,
+                                 static_cast_expected<float>(ival->toInt()));
+    } else {
+      ASSIGN_VALUE_OR_RETURN_ERR(constVal,
+                                 static_cast_expected<float>(ival->toDouble()));
+    }
+    glow::Tensor t(glow::ElemKind::FloatTy, dims);
+    t.init(glow::Tensor::InitKind::Broadcast, constVal,
+           F_.getParent()->getPRNG());
```
yinghai: What does `getPRNG` do?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1252,15 +1252,49 @@ Error PyTorchModelLoader::loadGlowFusedLinear(const torch::jit::Node *ptNode) {
...
 
+Expected<NodeValue> PyTorchModelLoader::loadNodeValueOrBroadcastedConstant(
+    const torch::jit::Value *value, llvm::ArrayRef<size_t> dims) {
+  if (hasGlowNodeValueForValue(value)) {
+    return getGlowNodeValueForValue(value);
+  } else {
+    GlowIValue *ival;
+    ASSIGN_VALUE_OR_RETURN_ERR(ival, getGlowIValueForValue(value));
+    float constVal;
+    if (ival->isInt()) {
```
yinghai: Basically we cast them to float in Glow?

torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1319,9 +1373,19 @@ Error PyTorchModelLoader::loadSub(const torch::jit::Node *ptNode) {
...
+    ASSIGN_VALUE_OR_RETURN_ERR(lhs, getGlowNodeValueForValue(inputs[0]));
+    ASSIGN_VALUE_OR_RETURN_ERR(
+        rhs, loadNodeValueOrBroadcastedConstant(inputs[1], lhs.dims()));
+  } else if (hasGlowNodeValueForValue(inputs[1])) {
+    ASSIGN_VALUE_OR_RETURN_ERR(rhs, getGlowNodeValueForValue(inputs[1]));
+    ASSIGN_VALUE_OR_RETURN_ERR(
+        lhs, loadNodeValueOrBroadcastedConstant(inputs[0], rhs.dims()));
+  } else {
+    return MAKE_ERR("Either lhs or rhs of arithmetic node must be a tensor");
+  }
```
yinghai: Are these code all replicated?

torch_glow/src/PyTorchModelLoader.cpp
jackm321: gets a PRNG 😛 
init takes a PRNG because if the InitKind is Xavier it would need it for random filling.

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Turns int into float tensor because we can't add ints in Glow

torch_glow/src/PyTorchModelLoader.cpp
jackm321: Yeah, idk I could factor this part out. Probably we should just factor everything out of the arithmetic ops since they're very similar

torch_glow/src/PyTorchModelLoader.cpp
yinghai: Yeah, just to make sure that it won't actually random fill the tensor, right?

jackm321: That's right, it will not.

jackm321:
I'm willing to take a stab at cleaning it up here, if it's annoying I'll move it to another diff.

PR #3788, opened 2019-11-16
[TensorLayout] Add setAlignment helper method to TensorLayoutDescription

shajrawi:
The optional `TensorLayoutDescription` helper class only supported a getter method for the official alignment extension. This PR adds a setter method for alignment + test case for it.

Test Plan:
`ninja test`


opti-mix:
@shajrawi Looks pretty good. But can we generalize it, i.e. support setting any attribute by providing the attribute name and attribute value? Also getting the value of any attribute by providing the attribute name?

shajrawi:
@opti-mix Done. added a new commit.

PR #3789, opened 2019-11-17
[PyTorch] aten::layer_norm

jackm321:
Summary:
Load aten::layer_norm

Documentation:
doxygen

Test Plan:
unit tests


lib/Optimizer/GraphOptimizer/Lower.cpp
```diff
@@ -422,6 +422,92 @@ static void lowerBatchNormalizationNode(Function *F, CompilationContext &cctx,
   replaceAllUsesOfWith(cctx.loweredInfoMap, BN.getResult(), newResult);
 }
 
+static void lowerLayerNormalizationNode(Function *F, CompilationContext &cctx,
```
yinghai: Maybe put a link to the math as a docstring.

PR #3792, opened 2019-11-18
[PyTorch] Add fb::embedding_bag_byte_rowwise_offsets

jackm321:
Summary:
* Add loader for `fb::embedding_bag_byte_rowwise_offsets`
* Add `EmbeddingBagByteRowwiseOffsetsNode` similar to `FusedRowwiseQuantizedSparseLengthsWeightedSum` but with offsets instead of lengths

Documentation:
doxygen

Test Plan:
Added operator test
Not sure yet how to add a test for the loader


lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -321,6 +321,7 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                 RowwiseQuantizedSparseLengthsWeightedSumNode::LengthsIdx) ==
             ElemKind::Int32ITy);
 
+  case Kinded::Kind::EmbeddingBagByteRowwiseOffsetsNodeKind:
```
jfix71: Do the two here support all the same data types? Even if they do it's at best confusing to read this code seeing that the EmbeddingBag will check inputs and outputs types based on RWQSLWS's indices for inputs and outputs. It's fragile in case ordering of inputs and outputs ever deviates, even if that is pretty unlikely.

jackm321: Yep this was a bad idea, I'll fix it

PR #3794, opened 2019-11-18
[ClassGen] add support for fixed IRGen

nickgg:
Summary: Adds a method to the InstrBuilder which allows specifying a fixed string to be used for IRGen. We usually specify IRGen manually in lib/IR/IRGen.cpp, but this doesn't work for Backend Specific Operators (now: BSOs). So far all BSOs have used autoIRGen, but I'd like to add one that doesn't. Does nothing if not set and don't think it's worth modifying an existing operator as a test.

Documentation: N/A

Test Plan: build only

### DIFF
--- a/tools/ClassGen/InstrBuilder.cpp
+++ b/tools/ClassGen/InstrBuilder.cpp
@@ -338,6 +338,11 @@ void InstrBuilder::addGradientInstr(
 }
 
 void InstrBuilder::emitAutoIRGen(std::ostream &os) const {
+  if (!IRGenBody.empty()) {
+    os << IRGenBody << "\n";
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


jfix71:
Just to make sure -- there already exists [`Backend::generateInst()`](https://github.com/pytorch/glow/blob/master/include/glow/Backend/Backend.h#L166) to allow a backend to customize IRGen for a backend specific Node or override IRGen for another Node. Is this intended to make the interface cleaner/replace that?

nickgg:
Oh I actually just didn't know about that. Let me investigate.

nickgg:
That interface seems fine to me, i'll use that. Thanks for the heads up @jfix71

PR #3797, opened 2019-11-19
[partitioner] Add logging for single partition

gaurav-work:
**Summary:**
Partition Info was logged only when more than one partitions were created.
Added code to log key partition info for a single partition case as well.

**Test Plan:**
Ran image_classifier on few different models (resnet50, vgg19, inception_v1)
with enough Interpreter backend memory to use a single partition and then
with less memory and multiple devices to create partitions. Compared the
size of input memory, output memory and constants as well as the logical
device ID.

**Testing:**
```
$ ninja all && ninja test
$ ./bin/image-classifier tests/images/imagenet/zebra_340.png -expected-labels=340 -use-imagenet-normalization -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data  --log-partition --interpreter-memory=400000
$ ./bin/image-classifier tests/images/imagenet/zebra_340.png -expected-labels=340 -use-imagenet-normalization -image-mode=0to1 -m=resnet50 -model-input-name=gpu_0/data  --log-partition --num-devices=4 --interpreter-memory=40000
```

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -423,33 +423,8 @@ Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
   NodeToFunctionMap mapping;
   for (auto func : module_->getFunctions()) {
     mapping.createPartition(func, backendName);
+    mapping.setGraphMemInfo(func, getFunctionMemory(func));
 
-    GraphMemInfo graphMem;
-    graphMem.constMemSize = func->getParent()->getConstantsSize();
-
-    // Walk thru all the nodes to compute input and output mem size.
-    for (auto &node : func->getNodes()) {
-      // If input to a node is a storage node but not constant then
-      // accummulate it as placeholder size.
-      for (size_t i = 0, e = node.getNumInputs(); i < e; i++) {
-        if (auto *in = llvm::dyn_cast<Storage>(node.getNthInput(i).getNode())) {
-          if (in->getKind() != Kinded::Kind::ConstantKind) {
-            graphMem.inMemSize += in->getType()->getSizeInBytes();
-          }
-        }
-      }
-
-      // If a node is SaveNode then accumulate size for output memory.
-      if (auto *SN = llvm::dyn_cast<SaveNode>(&node)) {
-        Storage *out = llvm::dyn_cast<Storage>(SN->getPlaceholder());
-        graphMem.outMemSize += out->getType()->getSizeInBytes();
-      }
-    }
-    // Output memory got double counted as input as it is storage node
-    // so subtract from the input memory size.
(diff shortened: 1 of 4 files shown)


lib/Partitioner/Partitioner.cpp
```diff
@@ -425,6 +426,38 @@ Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
   }
 
   NodeToFunctionMap mapping;
+  for (auto func : module_->getFunctions()) {
+    mapping.createPartition(func, backendName);
+
+    GraphMemInfo graphMem;
+    graphMem.constMemSize = func->getParent()->getConstantsSize();
+
+    // walk thru all the nodes to compute input and output mem size
```
beicy: nit: add a full stop (.) after each comment :)

gaurav-work: Sure thing. I read the guidelines earlier but forgot. Fixed now. Thanks.

lib/Partitioner/Partitioner.cpp
```diff
@@ -425,6 +426,39 @@ Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
...
 
   NodeToFunctionMap mapping;
+  for (auto func : module_->getFunctions()) {
+    mapping.createPartition(func, backendName);
+
+    GraphMemInfo graphMem;
+    graphMem.constMemSize = func->getParent()->getConstantsSize();
+
+    // Walk thru all the nodes to compute input and output mem size.
+    for (auto &node : func->getNodes()) {
```
beicy: For calculating the graphMem of a single function, I think it is better write it as a function in PartitionerUtils.cpp. Then we can write a unittest in similar to https://github.com/pytorch/glow/blob/master/tests/unittests/PartitionerTest.cpp#L727
.

gaurav-work:
@beicy - Great feedback. I will make appropriate changes and submit new code soon.

lib/Partitioner/Partitioner.cpp
```diff
@@ -425,6 +426,39 @@ Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
   }
 
   NodeToFunctionMap mapping;
+  for (auto func : module_->getFunctions()) {
+    mapping.createPartition(func, backendName);
+
+    GraphMemInfo graphMem;
+    graphMem.constMemSize = func->getParent()->getConstantsSize();
```
gcatron: The parent of function is the Module, this will report the constant size for the whole module,
To get just the constants used by the function you can use func->findConstants which can then be summed.

gaurav-work: Good catch. Made the appropriate change in latest push.

PR #3798, opened 2019-11-20
[Importer] Preserve output PH names in C2 loader

### LINKED ISSUE #3774: [Caffe2] Output graph name is not preserved after compilation.
For ONNX models all looks clean:
- the input/output placholder names are preserved after compilation (previously the output placeholders where prepended with "save_" which was ugly but now all is good).

For Caffe2 models the output placeholder name after compilation is not preserved:
- for example, for the LenetMnist bundle example, the original output placeholder name is "softmax"
whereas in the compiled model the name is "softmax__1" which is weird

In terms of usability and clarity, the output placeholder names for Caffe2 models should be preserved as well.

nickgg:
Summary: It's important to preserve the name of each external output in outside models, but since they must be unique with Node names we'll need to claim them before naming any nodes. This fixes the C2 Model Loader to do that.

Documentation: Fixes #3774 

Test Plan: ninja test, image-classifier with -dump-ir to make sure weights were named right.

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -109,6 +109,12 @@ public:
     usedNodeNames_.insert(name);
   }
 
+  /// Registers a name as used by a Storage node (Constant or Placeholder) in
+  /// this module.
+  void registerStorageName(llvm::StringRef name) {
+    usedStorageNames_.insert(name);
+  }
+
   /// Return a pointer to a uniqued type \p T.
   TypeRef uniqueType(const Type &T);
 
--- a/lib/Importer/Caffe2ModelLoader.cpp
+++ b/lib/Importer/Caffe2ModelLoader.cpp
@@ -1444,6 +1444,12 @@ Error Caffe2ModelLoader::loadInputs(const caffe2::NetDef &net,
 }
 
 Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
+  // Make a claim on the unique name of all output Placeholders.
+  for (int i = 0; i < net.external_output_size(); i++) {
+    auto &outputName = net.external_output(i);
+    G_.getParent()->registerStorageName(legalizeName(outputName));
+  }
+
   /// Load the network operators:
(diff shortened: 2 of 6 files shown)


mciprian13:
I assume also for the Resnet50 example in main.cpp (line 303) the symbol "gpu_0_softmax__1" should be changed to "gpu_0_softmax".

mciprian13:
I was wondering whether we should add unit tests to run the generated bundles (now -I think - the CI only verifies they build/compile properly).

include/glow/Graph/Graph.h
```diff
@@ -109,6 +109,10 @@ class Module final {
     usedNodeNames_.insert(name);
   }
 
+  void registerStorageName(llvm::StringRef name) {
```
bertmaher: Doxygen

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1464,8 +1471,13 @@ Error Caffe2ModelLoader::loadNetwork(caffe2::NetDef &net) {
     auto &outputName = net.external_output(i);
     NodeValue r;
     ASSIGN_VALUE_OR_RETURN_ERR(r, getNodeValueByName(outputName));
```
bertmaher: I don't understand something here: if nodes and placeholders share a namespace, wouldn't this `getNodeValueByName` fail to find the desired node, since it's probably been re-named something like `output__1`?

nickgg: Good question. It clearly works for e.g resnet, but your explanation sounds right. I'll look into it.

nickgg: OK, so the way this works is `getNodeValueByName` has its own map (ie not the Function/Module) which claiming a name in the Module doesn't affect. This is good because we're looking up illegal names directly, and double good because we write into that name without checking if the name is already taken. TLDR: nodeValueByName uses different names than the Nodes themselves.

PR #3805, opened 2019-11-21
[Partitioner] Add logging for a single partition

gaurav-work:
**Summary:**
Partition Info was logged only when more than one partitions were created.
Added code to log key partition info for a single partition case as well.
Added utility function to compute input, output and constant memory
used by a function.

**Test plan:**
Ran image_classifier on few different models (resnet50, vgg19, inception_v1)
with enough Interpreter backend memory to use a single partition and then
with less memory and multiple devices to create partitions. Compared the
size of input memory, output memory and constants as well as the logical
device ID.

Also added Partitioner Unittest with different sample functions to check
the graph memory computations.

**Testing:**
```
$ ninja all
$ tests/PartitionerTest --gtest_filter=PartitionerTest.funcMemInfoCalculation*
$ ninja test
```


lib/Partitioner/PartitionerUtils.cpp
```diff
@@ -493,6 +493,31 @@ GraphMemInfo getGraphMemInfo(const NodesSet &nodes) {
...
+
+  for (auto &node : func->getNodes()) {
+    // If a node is SaveNode then accumulate size for output memory.
+    if (auto *SN = llvm::dyn_cast<SaveNode>(&node)) {
+      graphMem.outMemSize += SN->getOutput().getType()->getSizeInBytes();
+    }
+  }
+  // Output memory got double counted as inputs during Placeholder list scan
+  // so subtract from the input memory size.
+  graphMem.inMemSize -= graphMem.outMemSize;
```
gcatron: This assumes a PH can never be an input and output if a PH is both then we would have the correct output size but input size would be smaller than it should be.
I think with training it's not safe to assume input/output is exclusive since a trainable weight is both an input and output. @SplitInfinity is that correct?

SplitInfinity: A trainable `Placeholder` is considered both an input and an output, yes.

gaurav-work: Good point. I didn't realize this at all. I modified the code just now to use pre-existing functions to determine Input, Output by determining users for the PH.

gaurav-work:
@gcatron - Who can land the changes?  We need to squash all commits into a single commit as well, correct?

beicy:
@gaurav-work I think so. After that, I think you can press "Import to Phabricator" , and then "Land to fbsource". If you can't see those 2 buttons, @rdzhabarov should be able to add you.

PR #3806, opened 2019-11-22
[LLVMIRCodeGen] Minor improvements to make backends more configurable

opti-mix:
Summary:

-  Allow backends to set/override the bundle API type

- Allow backends to override the logic for setting inline attributes on LLVM-functions

### DIFF
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -43,37 +43,6 @@ using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
-enum BundleApiType {
-  /// Dynamic bundle API (default) with the following features:
-  /// - the weights are exported in a binary file which are assumed
-  ///   to be loaded dynamically at run-time.
-  /// - the memory layout information (bundle configuration) is only
-  ///   available at run-time and therefore allows ONLY dynamic memory
-  ///   allocaton.
-  Dynamic,
-  /// Static bundle API with the following features:
-  /// - the weights are exported in a binary file but and also in a
-  ///   text file (C array format) suitable to include at compile-time.
-  /// - the memory layout information (bundle configuration) is available
-  ///   at compile-time through macros printed in the header file and thus
-  ///   allows also static memory allocation.
-  /// - this API is suitable for low end devices with no file system or OS
-  ///   (bare-metal).
-  Static,
-};
-
-namespace {
-llvm::cl::OptionCategory bundleSaverCat("Bundle Options");
-
-llvm::cl::opt<BundleApiType> bundleApi(
-    "bundle-api", llvm::cl::desc("Specify which bundle API to use."),
-    llvm::cl::Optional,
-    llvm::cl::values(clEnumValN(BundleApiType::Dynamic, "dynamic",
-                                "Dynamic API"),
-                     clEnumValN(BundleApiType::Static, "static", "Static API")),
-    llvm::cl::init(BundleApiType::Static), llvm::cl::cat(bundleSaverCat));
-} // namespace
-
 /// Header file string template.
 static const char *headerFileTemplate =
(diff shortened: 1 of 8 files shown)


lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -190,14 +159,16 @@ static void serializeBinaryToText(llvm::StringRef binFileName,
 }
 
 BundleSaver::BundleSaver(const IRFunction *F, const LLVMBackend &llvmBackend)
-    : irgen_(llvmBackend.createIRGen(F, allocationsInfo_)) {
+    : irgen_(llvmBackend.createIRGen(F, allocationsInfo_)),
+      bundleAPI_(BundleApiType::Static) {
```
shajrawi: According to the BundleApiType documentation, the default is dynamic. can you please either update the documentation comment or change the default construction to dynamic?

lib/LLVMIRCodeGen/CommandLine.cpp
```diff
@@ -88,3 +88,15 @@ llvm::cl::opt<llvm::FloatABI::ABIType>
...
+static llvm::cl::OptionCategory bundleSaverCat("Bundle Options");
+
+llvm::cl::opt<glow::BundleApiType>
+    bundleAPI("bundle-api", llvm::cl::desc("Specify which bundle API to use."),
+              llvm::cl::Optional,
+              llvm::cl::values(clEnumValN(glow::BundleApiType::Dynamic,
+                                          "dynamic", "Dynamic API"),
+                               clEnumValN(glow::BundleApiType::Static, "static",
+                                          "Static API")),
+              llvm::cl::init(glow::BundleApiType::Static),
```
shajrawi: I see it is initialized to static here as well - same comment as above.

lib/LLVMIRCodeGen/Pipeline.cpp
```diff
@@ -63,6 +63,42 @@ bool LLVMIRGen::preserveSymbol(const llvm::GlobalValue &GV) {
...
+        FF.hasFnAttribute(llvm::Attribute::AttrKind::AlwaysInline);
+    auto inlineAttr = getInlinineAttr(&FF);
+    if (inlineAttr != llvm::Attribute::AttrKind::None) {
+      DCHECK(inlineAttr == llvm::Attribute::AttrKind::AlwaysInline ||
+             inlineAttr == llvm::Attribute::AttrKind::NoInline)
+          << "Unknown inlining attribute returned by getInlinineAttr";
+      dontInline = (inlineAttr == llvm::Attribute::AttrKind::NoInline);
+    }
+    // Clear all attributes.
+    FF.setAttributes(AL);
```
shajrawi: I am not sure about this: we are clearing *all* attributes, not just inline, and not restoring the others for now.
This is fine, for now, but what If someone adds another llvm attribute in the future? could be a recipe for future bugs..

SameerAsal: Good catch, this however seems to be the old behaviour (this code was moved).

lib/LLVMIRCodeGen/BundleSaver.cpp
opti-mix: Updated the documentation.

lib/LLVMIRCodeGen/Pipeline.cpp
opti-mix: @shajrawi Sameer is right. We had this logic since ever. But I agree that it is currently not very easy to compose with other attributes.
I just tried to limit the scope and to remove just inlining related attributes, i.e. NoInline and InlineAlways. That in turn required removing OptimizeNone attribute as well, as it cannot exist without NoInline. Once I've done that, some of the CPU tests started failing because they now produce wrong results. I'm not quite sure why this happens, but it seems like it is better to leave the old logic in place for now.

PR #3807, opened 2019-11-22
Allow for folding of ElemKind conversion into IO where applicable for better bounds checking/testing

jfix71:
Summary: This PR introduces a `FoldElemKindConversionIntoIO` pass which looks for single-use Placeholders that have ConvertTo/Quantize/Dequantize right after/before them and folds them into the Placeholder (and SaveNode if an output). It then adds usage of this pass to `compareAgainstInterpreter()` tests. By doing this we allow the backend to more directly test the operator in its desired precision instead of also allowing the results of the test to be impacted by the specific logic used for type conversion (see bottom for before/after Functions).

This is opt-in, and at least initially meant for better bounds testing on tests which use `compareAgainstInterpreter()`. This is a dangerous pass, as it requires converting the associated Tensors for these Placeholders in PlaceholderBindings, and correctly getting handles to these Tensors based on these types. I've made this opt-in based on the optimization options in the compilation context.

Test Plan: All tests still pass. I modified the one test (BackendCorrectnessTest's `basicFCNetQuantized`) which currently does Int8 to Int8 comparison to expect bitwise accuracy.

Related to https://github.com/pytorch/glow/pull/3710

Before:
before

After:
after

### DIFF
--- a/tests/unittests/BackendTestUtils.cpp
+++ b/tests/unittests/BackendTestUtils.cpp
@@ -220,6 +220,40 @@ void dispatchInference(const std::string &fname,
   contexts[0].release();
 }
 
+/// Helper that iterates over all of the Placeholders in \p PHs and converts the
+/// Tensor pair found in \p bindings to the same type as the Placeholder if
+/// necessary.
+static void convertBindingsToCorrectType(PlaceholderBindings &bindings,
+                                         PlaceholderList PHs) {
+  for (Placeholder *PH : PHs) {
+    Tensor *T = bindings.get(PH);
+    TypeRef newTy = PH->getType();
+    if (T->getType().isEqual(newTy)) {
+      continue;
+    }
+    ElemKind newK = newTy->getElementType();
+    if (isQuantizedElemKind(newK)) {
+      Tensor QT = quantization::quantizeTensor(
+          *T, {newTy->getScale(), newTy->getOffset()}, newK);
+      T->assign(&QT);
+    } else {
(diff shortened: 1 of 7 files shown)


tests/unittests/BackendTestUtils.cpp
```diff
@@ -220,6 +220,40 @@ void dispatchInference(const std::string &fname,
   contexts[0].release();
 }
 
+/// Helper that iterates over all of the Placeholders in \p PHs and converts the
+/// Tensor pair found in \p bindings to the same type as the Placeholder if
+/// necessary.
+static void convertBindingsToCorrectType(PlaceholderBindings &bindings,
```
gcatron: Does a seperate PH list need to be passed in here?  We could iterate over all PHs in the bindings.

jfix71: @gcatron Hm so after reading it again, I think I did it like this because I wanted to be positive I only changed the placeholders that came from the specific Function I was compiling (so I pass in `Function::findPlaceholders()` to this function). I can't think of a specific case where we wouldn't want to fix any other mismatches but I guess since it felt like an aggressive thing to do to convert Tensors I was being conservative. I think I'm fine with changing it to convert all of the bindings but no strong feelings, WDYT?

gcatron: I think it's fine as is then. Being more specific also allows for more specific testing too.

PR #3810, opened 2019-11-22
Add large model support to glow

gcatron:
Hindsight being 20/20 there may have been a more incremental path here...

Summary:
Added new deferredWeightLoader that can be used to load large weight after compilation
Updated Provisioner to support new use case
Added Support for static placeholders to OpenCL device manager.
Added new test to deviceManagerTest for staticPlaceholder
Added new StaticPlaceholderTest

Documentation:

Test Plan: ninja check to make sure I didn't break anything and added StaticPlaceholderTest to test end2 end flow and a DeviceManager test for DM level test.

### DIFF
--- a/include/glow/Runtime/DeferredWeightLoader.h
+++ b/include/glow/Runtime/DeferredWeightLoader.h
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) Glow Contributors. See CONTRIBUTORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_RUNTIME_DEFERREDWEIGHTLOADER_H
+#define GLOW_RUNTIME_DEFERREDWEIGHTLOADER_H
+
+#include "glog/logging.h"
+
+#include "glow/Base/Tensor.h"
+#include "glow/Support/Error.h"
+#include "glow/Support/Register.h"
+#include "glow/Support/Support.h"
+
+namespace glow {
+namespace runtime {
+
+/// A base class for deferred weight loaders. This allows for large weights to
+/// be skipped during compilation and loaded after compilation one at a time.
+class DeferredWeightLoader {
+public:
+  /// Loads the next weight, returns an Error indicating success/failure. Frees
+  /// any resources used by the current deferred weight.
+  virtual Error loadNextWeight() = 0;
+
+  /// Accepts a void * \p loaderObject with is passed in from the interface
+  /// library, e.g. deferredBlobReader in onnxifi.
+  virtual Error setSrc(void *loaderObject) = 0;
+
+  /// Gets the name of the currently loaded weight.
+  virtual std::string getName() = 0;
+
+  /// Gets the Tensor for the current weight.
+  virtual Tensor *getTensor() = 0;
+
+  virtual ~DeferredWeightLoader() = default;
+};
+
+class DeferredWeightLoaderRegistry final {
+public:
+  void registerLoader(DeferredWeightLoader *loader);
+  DeferredWeightLoader *getLoader();
+
+private:
+  DeferredWeightLoader *loader_{nullptr};
+};
+
+/// Global singleton.
+DeferredWeightLoaderRegistry *DeferredLoader();
+
+} // namespace runtime
+} // namespace glow
+
+#endif // GLOW_RUNTIME_DEFERREDWEIGHTLOADER_H
--- a/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
+++ b/lib/Backends/OpenCL/OpenCLDeviceManager.cpp
@@ -378,6 +378,13 @@ void OpenCLDeviceManager::addNetworkImpl(const Module *module,
     buffers_.emplace(func.first, buffer);
     buffer->incrementUsers();
 
+    // Add function name to map for static placeholders.
+    for (auto PH : function->getIR()->getGraph()->findPlaceholders()) {
+      if (PH->isStatic()) {
+        staticPlaceholderToFunctions_[PH].push_back(func.first);
+      }
+    }
+
     DCHECK_LE(usedMemoryBytes_, maxMemoryBytes_);
     clReleaseCommandQueue(commands);
   }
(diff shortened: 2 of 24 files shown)


gcatron:
@jsubag Sorry it took a bit longer to get this up, there's a TransferStaticPlaceholderTest in DeviceManagerTest that should be useful. The new deferredWeightLoaderTest is higher level and more of an end2end test.

jsubag:
@gcatron Thanks!

tests/unittests/DeviceManagerTest.cpp
```diff
@@ -237,6 +237,85 @@ TEST_P(DeviceManagerTest, PartialTensorCopy) {
...
+  context->getPlaceholderBindings()->ensureOnHost();
+
+  Tensor *result = context->getPlaceholderBindings()->get(output);
+
+  ASSERT_TRUE(result);
+  EXPECT_NEAR(result->getHandle().at({0}), 8.0, 1E-5);
+}
+
 TEST_P(DeviceManagerTest, MultiRun) {
   auto module = makeBasicModule();
```
jsubag: Unrelated to this commit, but looks like this test is missing   CHECK_IF_ENABLED();

tests/unittests/DeferredWeightLoaderTest.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<std::unique_ptr<DeviceConfig>> configs;
+  auto deviceConfig = glow::make_unique<DeviceConfig>(backendName);
+  configs.push_back(std::move(deviceConfig));
+  std::unique_ptr<HostManager> hostManager =
+      glow::make_unique<HostManager>(std::move(configs), hostConfig);
+  return hostManager;
+}
+
+TEST_F(DeferredWeightLoaderTest, staticPlaceholderInference) {
+
```
jsubag: Looks like this test is missing   CHECK_IF_ENABLED(); here.

tests/unittests/DeviceManagerTest.cpp
gcatron: I think that's because multirun is considered a core feature the DeviceManager is incomplete if it doesn't support multiple requests coming in at once.

tests/unittests/DeferredWeightLoaderTest.cpp
gcatron: Good catch, it's currently hardcoded to OpenCL, I'll generalize it and add CHECK_IF_ENABLED().

lib/Runtime/DeferredWeightLoader.cpp
```diff
@@ -0,0 +1,49 @@
...
+  return loader_;
+}
+
+void DeferredWeightLoaderRegistry::registerLoader(
+    DeferredWeightLoader *loader) {
+  loader_ = loader;
+}
+DeferredWeightLoaderRegistry *DeferredLoader() {
+  static auto *deferredLoader = new DeferredWeightLoaderRegistry();
+  deferredLoader->registerLoader(new DeferredWeightLoader());
```
bertmaher: I don't think you want to instantiate a  `DeferredWeightLoader` inside this method, just a `DeferredWeightLoaderRegistry`.  You want to call `registerLoader` from a static instantiation of a subclass of `DeferredWeightLoader`.  See how this is done for StatsExporter: https://github.com/pytorch/glow/blob/master/tests/unittests/StatsExporterTest.cpp

lib/Backends/OpenCL/OpenCLDeviceManager.cpp
```diff
@@ -425,6 +432,70 @@ void OpenCLDeviceManager::returnRunCommandQueue(OpenCLCommandQueue &queue) {
   commandQueuePool_.returnCommandQueue(queue);
 }
 
+void OpenCLDeviceManager::transferStaticPlaceholderToDevice(
+    Placeholder *PH, Tensor *T, std::function<void(Error)> resultCB) {
+  // Need kernelLaunches commandQueue, buffer, symbol offset, Function.
```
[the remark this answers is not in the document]
gcatron: That's because it was a note to myself that was meant to be deleted... :)

lib/Backends/OpenCL/OpenCLDeviceManager.h
```diff
@@ -164,6 +164,10 @@ class OpenCLDeviceManager : public QueueBackedDeviceManager {
   /// The TID of the device (for TraceEvents);
   int deviceTid_{-1};
 
+  /// Map from PH to functionName for static placeholders.
```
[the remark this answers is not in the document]
gcatron: Most of the OpenCL DeviceManager keys it's structures by functionName so it's more convenient. Since the OpenCLCompiledFunction keeps a reference to Function* we could use that instead but this seems cleaner.

lib/Runtime/DeferredWeightLoader.cpp
gcatron: Good catch, the testDeferredWeightLoader is registered in the test and used for the test. Removed the instantiation of DeferredWeightLoader here.

include/glow/Runtime/DeferredWeightLoader.h
```diff
@@ -0,0 +1,65 @@
...
+namespace glow {
+namespace runtime {
+
+/// A base class for deferred weight loaders. This allows for large weights to
+/// be skipped during compilation and loaded after compilation one at a time.
+class DeferredWeightLoader {
+public:
+  /// Loads the next weight, returns an Error indicating success/failure. Frees
+  /// any resources used by the current deferred weight.
+  virtual Error loadNextWeight();
```
bertmaher: I kind of think these methods should all be pure virtual instead of having implementations that do nothing but throw or return dummy values.  You'd never actually want to instantiate the base class.

lib/Runtime/DeferredWeightLoader.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+DeferredWeightLoader *DeferredWeightLoaderRegistry::getLoader() {
+  return loader_;
+}
+
+void DeferredWeightLoaderRegistry::registerLoader(
+    DeferredWeightLoader *loader) {
+  loader_ = loader;
+}
+DeferredWeightLoaderRegistry *DeferredLoader() {
```
bertmaher: Vertical whitespace here please!

bertmaher:
Oh, you already pure-virtualized things.  Sorry, my review page was stale.  Thanks!

gcatron:
> Looks ready to go. Do consider the suggestion of making the DeferredWeightLoaderInterface pure virtual.

Already done!

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -396,6 +428,51 @@ Error Provisioner::provision(DAGListTy &networks, Module &module,
...
+        return MAKE_ERR(
+            ErrorValue::ErrorCode::RUNTIME_ERROR,
+            llvm::formatv(
+                "Error loading deferred weight. Name: {0} not found in module.",
+                weightName)
+                .str());
+      }
+      // Transfer weight to all devices needed.
+      for (const auto &device : placeholderToDevicManager[PH]) {
+        std::promise<Error> transferPromise;
```
ayermolo: This breaks Visual Studio build.
Not sure what's the best way to fix it. 
Adding std::promise as a friend doesn't seem to work. It produces other errors on the friend line.
Removing protected, just to try, leads to other errors. I think because constructor is explicit. Both of which were done for a reason.

include\future(226): error C2248: 'glow::detail::GlowError::GlowError': cannot access protected member declared in class 'glow::detail::GlowError'
25>C:\projects\glowExternal\glow\include\glow/Support/Error.h(384): note: see declaration of 'glow::detail::GlowError::GlowError'
25>C:\projects\glowExternal\glow\include\glow/Support/Error.h(341): note: see declaration of 'glow::detail::GlowError'

@jackm321 @gcatron  any ideas?

gcatron: @ayermolo what version of visual studio are you building with?

ayermolo: @gcatron Visual Studio 2017 15.9.17.

gcatron: @ayermolo sorry for the delay getting back to you, can you try changing the promise to void and capturing a reference to the promise in the lambda function? Like we do in ExecutionEngine::runInternal? If that compiles correctly with VisualStudio we can change to that.

ayermolo: @gcatron No problem, we all busy with internal stuff. :)
Something like this?
```
std::promise<void> transferPromise;
        Error transferError = Error::empty();
        auto done = transferPromise.get_future();
        devices_[device]->transferStaticPlaceholderToDevice(
            PH, weight, [&transferPromise, &transferError](Error err) {
              transferError = std::move(err);
              transferPromise.set_value();
            });
        RETURN_IF_ERR(std::move(transferError));
```

It passes. I also made change in DeviceManagerTest, and it builds and passes.
Alternative is this MR I did:
https://github.com/pytorch/glow/pull/3926
On plus side it solves the issue at the "root" so if down the road someone uses current way it is done it won't break build again. On downside the affect of the change has higher scope.
Right now it's showing OpenCL failing, but I think it's unrelated. I have seen QuantizationTest fail randomly locally for a while now.

gcatron: Exactly like that, that being said your changes in #3926 look fine to me too, I'd like to get feedback from @jackm321 before landing though in case there was a specific reason it's that way.

bertmaher:
A semi random idea here, if you declare the promise as `std::promise<GlowErrorEmpty>`, do things work?  It feels kind of weird but that should default-construct an empty error.

ayermolo:
Sorry was on holiday vacation.
@bertmaher This seems weird to me. Rest of the framework deals with GlowError. So in lambda function we will be taking in GlowError and then up casting it to set_value in our promise. Or am I missing something?

bertmaher:
Yeah, nm, I originally thought that would work without casting in set_value but it won’t, the converting ctors are in the wrong direction.  Anyways it’s kind of an abuse of the type system.

Issue #3791, opened 2019-11-18
[ONNX Importer] Replace symbolic dimension sizes with 1


mciprian13:
A lot of ONNX models obtained through conversion from TensorFlow models contain symbolic tensor sizes (allowed by the ONNX format through the "dim_param" specifier). This mostly occur for the batch size of the model like the example below:
![image](https://user-images.githubusercontent.com/29785908/69080470-bc370e00-0a44-11ea-8438-1ae01f600509.png)
Glow currently allows only statically typed tensor sizes but it should be capable to import models with symbolic sizes by forcing their value to "1" (I think it is a safe approach).

The tensor description in the proto looks like this:
```
  input {
    name: "data"
    type {
      tensor_type {
        elem_type: 1
        shape {
          dim {
            dim_param: "None"
          }
          dim {
            dim_value: 3
          }
          dim {
            dim_value: 416
          }
          dim {
            dim_value: 416
          }
        }
      }
    }
  }
```

tlepley-cadence:
When converting Tensorflow models to ONNX models (with TF2ONNX), I had noticed that the batch dimension was always replaced by 1 in the ONNX model. 

Which TF->ONNX converter did you use to generate this ONNX model with a "None" batch dimensions ?

mciprian13:
I think it was the MMDNN tool:
https://github.com/microsoft/MMdnn

mciprian13:
This issue was solved in #3817.

PR #3817, opened 2019-11-25
[ONNX Importer] Allow user to resolve undefined model symbols ("dim_param").

mciprian13:
**Summary**
Allow symbols in ONNX models (expressed by "dim_param") and force them to 1.
Related to #3791.

**Documentation**
None.

**Test Plan**
Two unit tests in ONNX Importer.


lib/Importer/ONNXModelLoader.cpp
```diff
@@ -36,14 +36,29 @@ using namespace glow;
 using llvm::cast;
 
 namespace {
+
+/// Get the shape of a TensorShapeProto given by \p shapeProto and return the
+/// dimensions in the vector \p dim passed by reference.
+Error getProtoShape(const ONNX_NAMESPACE::TensorShapeProto &shapeProto,
```
jfix71: tiny nit: Could return `Expected<std::vector<size_t>>` here instead of `Error`.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -36,14 +36,29 @@ using namespace glow;
...
+                    std::vector<size_t> &dim) {
+  for (auto d : shapeProto.dim()) {
+    // If the proto has a symbolic dimension "dim_param" we will force it to 1.
```
jfix71: This seems kind of like a hack. I'm wondering if we should make this opt-in, maybe through a cl opt, or cmake flag to configure Glow with if cl opt doesn't work for you. WDYT?

mciprian13: It looks like a hack but I wouldn`t treat like one. A lot of the models obtained through conversion from TensorFlow to ONNX exhibit placeholders with the batch size defined as a "dim_param" with some funny value (like "None" or "unk_symbol" etc). For these kind of models Glow currently retrieves an error because Glow does not support the "dim_param" field which is intended for symbolic and implicitly dynamic sizes which Glow does not support (and most probably never will).

lib/Importer/ONNXModelLoader.cpp
mciprian13: Done.

lib/Importer/ONNXModelLoader.cpp
mciprian13: Again you can take a look at the example #3791.
If it feels hackish for you then you what should be the elegant solution?

mciprian13: So I pushed another commit in which I added a command line option called `-onnx-define-symbol` which allows the user to define symbols in order to resolve undefined symbols from the model. In order to make it less hackish (as you suggested) I made it mandatory, in the sense that if there`s an unresolved symbol in the model I don`t force it by default to a value of 1 but instead I expect the user to define it (otherwise I throw an error).

The syntax of this command line options is: `-onnx-define-symbol=name,value`. For example in my case I use it as: `-onnx-define-symbol=None,1`. Multiple symbols can be defined by using this option multiple times.

What do you say?

mciprian13:
Can we land this? This is already reviewed.
Thanks!

jfix71:
@mciprian13 Sorry for the delay -- I needed to make some adjustments for our internal build/test system. I changed this slightly for testing purposes, with a `setOnnxDefineSymbol()` function you can use to set/override what would otherwise be provided on the command line. This way we can unit test in a more targeted way. It should be landing today.

PR #3818, opened 2019-11-25
Added calls to logPartitionInfo to partitioning when an error is encountered

gcatron:
Summary:
Added calls to logPartitionInfo to partitioning when an error is encountered to improve debugging.

Documentation:

Test Plan:
Ninja check to ensure nothing broke
Notice that PartitionerTest.memoryUsageValidation1 now prints PartitionInfo.

### DIFF
--- a/lib/Partitioner/PartitionerValidation.cpp
+++ b/lib/Partitioner/PartitionerValidation.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "glow/Partitioner/PartitionerValidation.h"
+#include "glow/Partitioner/PartitionerUtils.h"
 
 #include "llvm/Support/FormatVariadic.h"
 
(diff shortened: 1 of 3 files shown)


lib/Partitioner/PartitionerValidation.cpp
```diff
@@ -32,6 +33,9 @@ Error logicalDevicesValidation(
       partitionsNum[backendName].insert(logicalIDList[i]);
     }
     auto backendNum = backendMap.at(backendName).num;
+    if (partitionsNum[backendName].size() > backendNum) {
+      logPartitionInfo(partitions);
+    }
```
beicy: nit: Since the error has been detected already, next line "RETURN_ERR_IF_NOT" could be an else branch and just return the error.

lib/Partitioner/PartitionerValidation.cpp
```diff
@@ -50,6 +54,9 @@ Error memoryUsageValidation(
     auto backendName = partitions.getPartitionBackendName(func);
     auto usedMemSize = partitions.getGraphMemInfo(func).getTotalMemSize();
     auto availableMemSize = backendMap.at(backendName).memSize;
+    if (usedMemSize > availableMemSize) {
+      logPartitionInfo(partitions);
+    }
```
beicy: nit: same as above.

PR #3823, opened 2019-11-27
[BundleSaver](trivial) Minimize header file content.

mciprian13:
**Summary**
In one of the latest commits the BundleSaver prints in the auto-generated header file also the model constants. For large models the generated API header file becomes more of a log file and less of a clean API. For example:
[bundle.txt](https://github.com/pytorch/glow/files/3898153/bundle.txt)
This feature (verbosity) should be made optional since a clean approach should mandate that the API includes only what`s necessary and sufficient (in this case placeholder information).

Documentation
None

[Optional Fixes #issue]

Test Plan
None

### DIFF
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -284,9 +284,10 @@ void BundleSaver::saveHeader(llvm::StringRef headerFileName) {
 
   // Format model description.
   std::string modelInfo = strFormat("// Model name: \"%s\"\n"
-                                    "// Total data size: %lu (bytes)\n"
-                                    "// Placeholders:\n",
+                                    "// Total data size: %lu (bytes)\n",
                                     bundleName.data(), totMemSize);
+  // Print placeholders (mandatory).
+  modelInfo += "// Placeholders:\n";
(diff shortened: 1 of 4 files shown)


opti-mix:
@mciprian13 Yes, makes sense.

Two minor comments:
- Minimalize should be "minimize". Please update the commit message and the PR header
- Could you add an option to enable/disable this flag via a command-line?

mciprian13:
Done. Have a look.

lib/LLVMIRCodeGen/BundleSaver.cpp
```diff
@@ -43,6 +43,15 @@ using llvm::cast;
 using llvm::dyn_cast;
 using llvm::isa;
 
+namespace {
+llvm::cl::OptionCategory bundleSaverCat("Bundle Options");
```
opti-mix: @mciprian13  Could you move these definitions into `LLVMIRCodeGen/CommandLine.cpp`? We have `bundleSaverCat` and one of the bundle options defined there already. Let's keep it in one place.

mciprian13: Sorry ... it should have been obvious. I solved it.

PR #3830, opened 2019-12-02
Add method to list supported options for device managers and compilation functions.

omromano:
Summary:
Added new methods in the Backend to expose the supported options for device manages and compiled functions.
Compiled function options can be specified in BackendOptions::backendSpecificOpts (in Backed::compile).
Device manager options can be specified in DeviceConfig::parameters (in Backed::createDeviceManager).

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -174,6 +174,18 @@ public:
   virtual runtime::DeviceManager *
   createDeviceManager(const runtime::DeviceConfig &deviceConfig);
 
+  /// \returns the supported options for compiled functions (name=>description).
+  virtual llvm::StringMap<std::string>
+  getSupportedCompiledFunctionOptions() const {
+    return llvm::StringMap<std::string>();
+  };
+
+  /// \returns the supported options for device managers (name=>description).
+  virtual llvm::StringMap<std::string>
+  getSupportedDeviceManagerOptions() const {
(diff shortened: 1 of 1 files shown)


omromano:
Hi @jfix71, please take a look.
Thanks,
Ophir.
CC: @jsubag

jsubag:
Hi @gcatron,
Isn't the current design more inline with having the application go through the Backend when creating the DeviceManager?
So the application doesn't need to know the actual DeviceManager type (or the backend type for that matter) and since the options are used when creating the device manager it's not viable to first create then query for the supported options (with some virtual function).

gcatron:
Hey @jsubag,
I agree the application doesn't really need to know about the DeviceManager, my thinking was the application would query the backend for supported features but the runtime would be the one querying the deviceManager for supported features. The thought was this would allow for two deviceManagers of the same type to support different features based on the reality of the hardware. But that may be overkill, if it's a static thing then both methods on backend should be fine.

jsubag:
Hi @gcatron,
I haven't thought of that usage tbh - but as you said this is static for now.
However, if you're considering different capabilities across device managers of the same Backend, then this implies either a mechanism to query their capabilities by ID (from the Backend) prior to creation or create first and then have a secondary set of options/caps you can toggle afterwards - at which point the DeviceManager method would make more sense. 
Either of those will need a larger change in the Backend/DeviceManager interface.

PR #3832, opened 2019-12-02
[TensorLayout] Mark MatMul and Pad as nodes that can operate on any layout

### LINKED ISSUE #3804: [Layout] Layout verification failed
Now I don`t know why this happens (a previous version of Glow did not exhibit this) but I get an error during Layout verification in this portion of graph:

Untitled

I get the following error:
```
In 'dense_1' From 'model.onnx'
input 0
MatMul
name : dense_1
LHS : float<10 x 1280 x 1 x 1>
RHS : float<1280 x 50>
users : 1
Result : float<10 x 50>

Mismatching layouts:
Provided layout
Layout: NCHW [name = N : alignment = 1 : index = 0, name = C : alignment = 1 : index = 1, name = H : alignment = 1 : index = 2, name = W : alignment = 1 : index = 3]
Expected layout
Layout: NHWC [name = N : alignment = 1 : index = 0, name = H : alignment = 1 : index = 1, name = W : alignment = 1 : index = 2, name = C : alignment = 1 : index = 3]
From 'model/keurig.onnx'
Expected correct Glow canonical layouts for the graph
For comparison `LHS Equal RHS` with:
LHS: 0
RHS: 1
```
I`m new to this Layout verification thing. I don`t understand the problem here and the solution.
Can you please hint the problem/solutin? (I think @shajrawi is the author of this).

Thanks!

shajrawi:
Summary:
While they are not data parallel, they can operate on any order of dimensions.

Fixes #3804

Test Plan:
```
model-compiler -backend=CPU -model=model1.onnx -emit-bundle=build
model-compiler -backend=CPU -model=model2.onnx -emit-bundle=build
```

### DIFF
--- a/lib/Graph/TensorLayout.cpp
+++ b/lib/Graph/TensorLayout.cpp
@@ -623,8 +623,10 @@ static bool acceptsAnyInputLayout(const glow::Node *node) {
   case Kinded::Kind::BatchedReduceMinNodeKind:
   case Kinded::Kind::BatchNormalizationNodeKind:
   case Kinded::Kind::BatchNormalizationGradNodeKind:
+  case Kinded::Kind::PadNodeKind:
   case Kinded::Kind::ReshapeNodeKind:
   case Kinded::Kind::MeanVarNormalizationNodeKind:
+  case Kinded::Kind::MatMulNodeKind:
   case Kinded::Kind::SGDNodeKind: {
     return true;
   }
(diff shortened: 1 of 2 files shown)


mciprian13:
If I'm not too tiresome :) can you take a look over all the operators currently supported by Glow and verify if there is one which might have slipped from being marked as 'any layout'? Thanks for fixing this!

PR #3837, opened 2019-12-02
repro tool: add support for partial tensor

tracelogfb:
Summary:
The old save/load mechanism in the repro tool didn't take partial tensors into consideration. This adds the support.

If we detect an input is a partial tensor, we will save only the slice that contains the data. On the loading side, we'll compare what we load with the shape specified in the model. If the size doesn't match, we'll either enable parital tensor or pad it with 0.

Differential Revision: D18771353

### DIFF
--- a/lib/Onnxifi/Base.cpp
+++ b/lib/Onnxifi/Base.cpp
@@ -232,7 +232,24 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
       ONNX_NAMESPACE::GraphProto inputG;
       for (const auto &p : ctx->getPlaceholderBindings()->pairs()) {
         auto *t = inputG.add_initializer();
-        ONNXModelWriter::writeTensor(*p.second, t);
+        const auto &inputTensor = *p.second;
+        size_t unpaddedSize = inputTensor.getUnpaddedSizeInBytes();
+        size_t tensorSize = inputTensor.getSizeInBytes();
+        if (unpaddedSize == tensorSize) {
+          ONNXModelWriter::writeTensor(inputTensor, t);
+        } else {
+          // If the input is a partial tensor, then save only the part that has
+          // data.
+          auto ty = inputTensor.getType();
+          auto dims = ty.dims().vec();
+          dims[0] = dims[0] / (tensorSize / unpaddedSize);
+          const auto &resized = inputTensor.getUnowned(dims);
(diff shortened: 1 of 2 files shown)


lib/Onnxifi/Base.cpp
```diff
@@ -232,7 +232,24 @@ onnxStatus Graph::setIOAndRun(uint32_t inputsCount,
...
+        size_t unpaddedSize = inputTensor.getUnpaddedSizeInBytes();
+        size_t tensorSize = inputTensor.getSizeInBytes();
+        if (unpaddedSize == tensorSize) {
+          ONNXModelWriter::writeTensor(inputTensor, t);
+        } else {
+          // If the input is a partial tensor, then save only the part that has
+          // data.
+          auto ty = inputTensor.getType();
+          auto dims = ty.dims().vec();
+          dims[0] = dims[0] / (tensorSize / unpaddedSize);
```
jfix71: I don't understand the logic here -- I think we want to do something like `dims[0] = unpaddedSize / ty.getElementSize()`, no?

tests/unittests/Repro.cpp
```diff
@@ -141,14 +146,50 @@ void run() {
...
+      if (loadedSize != fullSize) {
+        if (enablePartialTensor) {
+          LOG(INFO) << "Loading " << tp.name()
+                    << " as a partial tensor: partial size="
+                    << tensor->getType().toString()
+                    << " full size=" << fullType.toString();
+          Tensor fullTensor(tensor->getUnsafePtr(), &fullType,
+                            tensor->getSizeInBytes());
+          // 'fullTensor' doesn't own the underlying data. 'tensor' does. So we
+          // want to keep the origianl tensor object around until inference is
```
jfix71: typo: `origianl`

lib/Onnxifi/Base.cpp
tracelogfb: dims[0] = unpaddedSize / ty.getElementSize() would only work if the tensor is 1D. From my testing on real models, partial tensor doesn't have to be 1D. So say we have a 8x10 tensor and it's partially populated with 2 entries in it, then we just want to take a slice of 2. So dim 0 needs to be (8 / 4). We get 4 from (tensorSize / unpaddedSize) = (8*10*element_size) / (2*10*element_size) = 4. 

So altogether it's dims[0] / (tensorSize / unpaddedSize).

PR #3844, opened 2019-12-04
32b dim_t fixes + a new builder for 32b dim t

pjaaskel:
Summary:

There were a few remaining 32b dim_t issues in recent code. E.g. one case of wrong vector initialization in the onnximporter/exporter.

Also adds a new circleci test for 32B_DIM_T that also builds the OpenCL backend to catch most of the size_t/64b assumed index code additions first hand. 

Test Plan:

Tested with LLVM7 64b/32b and pocl 64b/32b.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -141,6 +141,9 @@ elif [[ "$CIRCLE_JOB" == "PYTORCH" ]]; then
 elif [[ "$CIRCLE_JOB" == "OPENCL" ]]; then
     install_pocl
     CMAKE_ARGS+=("-DGLOW_WITH_OPENCL=ON")
+elif [[ "$CIRCLE_JOB" == "32B_DIM_T" ]]; then
+    install_pocl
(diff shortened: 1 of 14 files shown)


bertmaher:
What is dim32/sparseToDense.onnxtxt?  It causes onnxExporterTest to fail, because it tries to load the dim32 directory as an onnx protobuf...

pjaaskel:
> What is dim32/sparseToDense.onnxtxt? It causes onnxExporterTest to fail, because it tries to load the dim32 directory as an onnx protobuf...

It's a 32b index version of that model. How can I reproduce the fail? It doesn't fail for me by default.

bertmaher:
Weird, the test fails using the FB-internal build system, but not with cmake.  This is probably something I just need to solve on our end.  Sorry for the confusion.

bertmaher:
I did the merge internally to fix the OnnxExporter issues; landed as d09a273

PR #3851, opened 2019-12-04
fix partial tensor handling in NNPI backend

tracelogfb:
Summary: Previously for tensors that don't have the full size populated, but not a SLS length/index, we didn't copy over the whole tensor (omitting the parts that were zeroed out).

Differential Revision: D18814312


tracelogfb:
cc: @jsubag

jsubag:
LGTM

PR #3858, opened 2019-12-05
Add Quantize and convertToFP16 support to large model loading

gcatron:
Summary:
This adds support for loading deferred weights when they need to be quantized or converted to FP16. 
This PR splits foldElemKindConversionIntoIO into Input and Output and modified Input pass to support selectively folding static placeholders only and updated the Provisioner to check if a conversion is needed before transferring the weight to the device.

Documentation:

Test Plan: added new FP16StaticPlaceholderInference test to test conversion flow.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -3142,21 +3142,21 @@ bool FoldTileAddIntoBatchedAdd::run(Function *F,
   return changed;
 }
 
-/// Fold ElemKind conversion nodes (ConvertTo, Quantize, Dequantize) into
-/// single-user Placeholders and SaveNodes. Note that this changes the semantics
+/// Fold ElemKind conversion nodes (ConvertTo, Quantize) into
+/// single-user Placeholders. Note that this changes the semantics
 /// of the IO of the Function and so must be done carefully, i.e. should always
 /// be opt-in and done alongside conversion of corresponding Tensors in
-/// PlaceholderBindings.
-bool FoldElemKindConversionIntoIO::run(Function *F,
-                                       const CompilationContext &cctx) {
+/// PlaceholderBindings. If
+/// cctx.optimizationOpts.foldStaticPlaceholderConversions is set this will
+/// only change Placeholders marked as static.
+bool FoldElemKindConversionIntoInputs::run(Function *F,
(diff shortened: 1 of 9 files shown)


lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -448,18 +448,32 @@ Error Provisioner::provision(DAGListTy &networks, Module &module,
                 weightName)
                 .str());
       }
+      // Convert the weight if needed.
+      auto newTy = PH->getType();
+      auto weight = loader->getTensor();
+      if (!weight->getType().isEqual(newTy)) {
```
jfix71: This should only be the case if it's a static PH right? Maybe add an assert here?

gcatron:
@jfix71  Good point, I went ahead and renamed `onlyFoldStaticPlaceholderConversions` to `foldStaticPlaceholderConversions` and modified the fold inputs pass to skip non-static placeholders if `foldElemKindConversionIntoIO` is not set. Both options should work together now.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3374,11 +3399,22 @@ Error glow::optimizeFunction(Function *F, const Backend &B,
...
+  }
+
   // If requested, fold ElemKind conversion Nodes into inputs and outputs
   // (Placeholders and SaveNodes).
   if (cctx.optimizationOpts.foldElemKindConversionIntoIO) {
-    FunctionPassManager FPM("FoldElemKindConversionIntoIO",
-                            {FunctionPassID::FoldElemKindConversionIntoIO});
+    FunctionPassManager FPM(
+        "FoldElemKindConversionIntoIO",
+        {{FunctionPassID::FoldElemKindConversionIntoInputs},
```
jfix71: small nit: Looks a little funny/inefficient to me to back to back potentially create pipelines with `FoldElemKindConversionIntoInputs`, especially since they could both be run.

One thing you could do is merge this into a single `FunctionPassManager` -- check if `foldStaticPlaceholderConversions || foldElemKindConversionIntoIO`, create a `FunctionPassPipeline` with `FoldElemKindConversionIntoInputs` since it'll always be run, and then conditionally add `FoldElemKindConversionIntoOutputs` if `foldElemKindConversionIntoIO`. Then feed that `FunctionPassPipeline` into an `FPM`.

(whew that's a mouthful)

PR #3859, opened 2019-12-05
[GraphOptimizer] Do transformPostLowering in a loop

shajrawi:
Summary:

Iterate to a fixed point instead of doing the post-lowering once. anywhere we iterate to a fixed point we need to dce in between passes or else we keep optimizing dead code.

### DIFF
--- a/lib/Backends/OpenCL/Transforms.cpp
+++ b/lib/Backends/OpenCL/Transforms.cpp
@@ -65,6 +65,14 @@ bool OCLBackend::transformPostLowering(Function *F,
         continue;
       }
 
+      // We need to replace both of the MaxPool results with their NCHW
+      // counterpart in order to get rid of the old node. There appears to be a
+      // bug in the implementation of the result(1) portion of the optimized
+      // kernel, if getArgmax got users - bail.
+      if (PMN->getArgmax().getNumUsers() > 0) {
+        continue;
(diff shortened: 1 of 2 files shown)


lib/Backends/OpenCL/Transforms.cpp
```diff
@@ -65,6 +65,14 @@ bool OCLBackend::transformPostLowering(Function *F,
         continue;
       }
 
+      // We need to replace both of the MaxPool results with their NCHW
+      // counterpart in order to get rid of the old node. There appears to be a
+      // bug in the implementation of the result(1) portion of the optimized
+      // kernel, if getArgmax got users - bail.
+      if (PMN->getArgmax().getNumUsers() > 0) {
+        continue;
```
opti-mix: If we do not call convertMaxPoolToNCHWPool here, what happens at runtime? Can this operation be executed by the OpenCL backend at all? I.e. is using a non-NCHW layout an option at all?

shajrawi: Just like other OpenCL kernels, we have multiple implementations. an 'optimized' one (NCHW) and an unoptimized one (NHWC) in this case. so, if we do not convert, then in this case we will have the reference implementation that'll give lower performance but maintain correctness.

PR #3865, opened 2019-12-06
Add SLWS in the interpreter for fp16

SameerAsal:
Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/lib/Backends/Interpreter/Interpreter.cpp
+++ b/lib/Backends/Interpreter/Interpreter.cpp
@@ -365,11 +365,25 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                   FusedRowwiseQuantizedSparseLengthsWeightedSumNode::
                       ResultIdx) == ElemKind::Float16Ty);
     case ElemKind::UInt8FusedQTy:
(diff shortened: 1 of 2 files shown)


lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -365,11 +365,25 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
                   FusedRowwiseQuantizedSparseLengthsWeightedSumNode::
                       ResultIdx) == ElemKind::Float16Ty);
     case ElemKind::UInt8FusedQTy:
-      return (NI.getInElemTy(FusedRowwiseQuantizedSparseLengthsWeightedSumNode::
-                                 WeightsIdx) == ElemKind::FloatTy) &&
-             (NI.getOutElemTy(
-                  FusedRowwiseQuantizedSparseLengthsWeightedSumNode::
-                      ResultIdx) == ElemKind::FloatTy);
+      return ((NI.getInElemTy(
```
opti-mix: @SameerAsal  Could you split this huge expression into multiple parts? It is a bit hard to read it.

May be something like:
if (part1) {
  return true;
}
return part2;

or something like that.

PR #3866, opened 2019-12-06
Use fused activations in NNPI

jfix71:
Summary: Now that we have fused activations supported generically, start using them in NNPI. This should make optimizations more straightforward and match HW better.

Differential Revision: D18847351

### DIFF
--- a/include/glow/Graph/Graph.h
+++ b/include/glow/Graph/Graph.h
@@ -345,10 +345,17 @@ public:
 
   const Module *getParent() const { return parent_; }
 
+  /// Calls into \ref Module::uniqueName() and \returns a unique name from
+  /// \p name based on the Module's used storage names and the unique node names
+  /// used by this Function.
+  llvm::StringRef uniqueName(llvm::StringRef name) {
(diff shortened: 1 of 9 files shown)


jfix71:
@jsubag Please take a look here and verify this change.

jfix71:
@jsubag Note that this is not currently working/seems to be failing inside `nnpiNetworkBuild()`:

```
I1205 23:54:42.204399 4172633 Importer.cpp:507] Failed to build network
E1205 23:54:42.205538 4172633 NNPICompiledFunction.cpp:28] Failed to import function
```

If you have any quick tips here on what's wrong please let me know, thanks!

jsubag:
Hi @jfix71, 
Off the top of my head you can check the following:
* Are you using the same name for both op and activation? (or same names among tensors)
* Is the intermediate tensor between them no longer represented in the graph?
You can also set the env variable NNPI_LOG_LEVEL=0 or build in debug and see where it fails a condition in the libnnpi_transformer.so

jfix71:
Going to shelve this for now and come back to it when I have more cycles.

PR #3871, opened 2019-12-07
Add torch_glow unittests

yinghai:
Summary:
Enable torch_glow unittests in fbcode

Note that we need to convert all the tests to using unittest instead pytest.

Reviewed By: qizzzh

Differential Revision: D18863278

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -2645,7 +2645,8 @@ Error PyTorchModelLoader::loadMean(const torch::jit::Node *ptNode) {
     }
   }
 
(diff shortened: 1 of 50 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -2645,7 +2645,8 @@ Error PyTorchModelLoader::loadMean(const torch::jit::Node *ptNode) {
     }
   }
 
-  if (hasGlowIValueForValue(inputs[MeanInputs::keepdims])) {
+  if (inputs.size() > 2 &&
```
zrphercule: What is this for?

yinghai: It's the good thing about having asan enabled by default internally. It flushed out this read overflow issue.

PR #3873, opened 2019-12-07
[PyTorch] Add rowwise quantized linear support

zrphercule:
Summary:
Add support for rowwise quantized linear in pytorch-glow
Documentation:

[Optional Fixes #issue]

Test Plan:
pytest torch_glow/tests/nodes/quantized_linear_test.py::test_quantized_linear_packed_rowwise

### DIFF
--- a/torch_glow/src/PyTorchModelLoader.cpp
+++ b/torch_glow/src/PyTorchModelLoader.cpp
@@ -1455,6 +1455,9 @@ Error PyTorchModelLoader::loadQuantizedLinear(const torch::jit::Node *ptNode) {
   const c10::optional<at::Tensor> ptBiasTensor =
       unpackedParams[1].toOptional<at::Tensor>();
 
+  bool isRowwiseQuantized = ptWeightTensor.is_quantized() &&
+                            ptWeightTensor.qscheme() == at::kPerChannelAffine;
+
   // unpacked weights
   auto weightTensor = ptTensorToGlowTensor(ptWeightTensor);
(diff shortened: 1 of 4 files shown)


torch_glow/src/PyTorchModelLoader.cpp
```diff
@@ -1455,6 +1455,9 @@ Error PyTorchModelLoader::loadQuantizedLinear(const torch::jit::Node *ptNode) {
   const c10::optional<at::Tensor> ptBiasTensor =
       unpackedParams[1].toOptional<at::Tensor>();
 
+  bool isGroupwiseQuantized = ptWeightTensor.is_quantized() &&
```
yinghai: Why not isChannelwiseQuantized()?

torch_glow/tests/nodes/quantized_linear_test.py
```diff
@@ -13,6 +13,8 @@ def test_quantized_linear_packed():
     dq = torch.nn.quantized.DeQuantize()
     linear = torch.nn.Linear(5, 5)
 
+    # all weights are the same
+    # therefore is not rowwise quantized
```
yinghai: What's the significance of this?

zrphercule: To tell the different between this test and the new added test, in case we want to modify the test in the future.

torch_glow/src/PyTorchModelLoader.cpp
zrphercule: Yeah... Actually here it should be isRowwiseQuantized

PR #3875, opened 2019-12-07
[PyTorch]Quantized resnet unittests for fbcode/glow/torch_glow

zrphercule:
Differential Revision: D18875664


torch_glow/tests/utils.py
```diff
@@ -7,6 +7,35 @@
 SUBGRAPH_ATTR = "Subgraph"
 
 
+def get_image(filename):
```
zrphercule: For this chunk of code, I have already checked with Adnan, and he confirmed it should be Okay to publish this to OSS.
In order to repro this with real-life img, this function is necessary.

torch_glow/tests/models/resnet_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+        torch_answer = model(x).topk(3)
+
+        with torch.no_grad():
+            torch_glow.enableFusionPass()
+
+            traced_model = torch.jit.trace(model, x)
+            traced_model.eval()
+
+            glow_answer = traced_model(x).topk(3)
```
raghuramank100: Great to see this end to end example, a couple of comments about the glow API:
The current API is non-intuitive. It is not clear why enabling fusion pass in torch_glow affects the traced model.
It will be more intuitive to do something like:
```python
traced_model = torch.jit.trace(model,x)
glow_model = torch_glow.convert(model, glow_options)
torch_answer = traced_model(x)
glow_answer = glow_model(x)
```
Is this possible? (cc: @yinghai , @jackm321 )

yinghai: Maybe but python interface isn't really what we focus on. This is for convenience of testing.

yinghai:
Checked in internally.

PR #3879, opened 2019-12-09
Update NNPI Backend to v0.4.1.2

omromano:
Summary: Updating NNPI Backend to version 0.4.1.2


omromano:
@jfix71, @arunm-git, please take a look.
Thanks.

lib/Backends/NNPI/NNPIOptions.h
```diff
@@ -0,0 +1,183 @@
...
+  /// Compilation paramters used as defaults (if exists) that may be overriden
+  /// by environment variables.
+  NNPICompilationOptions(const std::map<std::string, std::string> *parameters);
+  /// Dump the status of all variables.
+  virtual std::string dumpStatus() const override;
+  /// Lists the supported options (name=>description).
+  static llvm::StringMap<std::string> getSupportedOptions();
+  /// Compile to ICE-T if true, ICE-Ref otherwise - this variable cab be
+  /// verridden by inferOnDevice_.
+  bool useIceT_;
```
jfix71: nit: Our style is for public variables of classes to not have underscore suffixes -- only private/protected members do.

lib/Backends/NNPI/NNPIOptions.cpp
```diff
@@ -0,0 +1,437 @@
...
+          CompilationDebugConfigFileOption, parameters);
+  customDspKernelsFile_ =
+      NNPIOptions::getStringVal<std::map<std::string, std::string>>(
+          CustomDSPLibOption, parameters);
+
+  setLogLevel(compilationLogLevel_);
+}
+
+void NNPICompilationOptions::setLogLevel(int logLevel) {
+  // We have only one log level for NNPI compilation. Setting it will be change
```
jfix71: tiny grammar nit: "Setting it will be change"

lib/Backends/NNPI/NNPIOptions.cpp
```diff
@@ -0,0 +1,437 @@
...
+  options[IceCoresOption] = {
+      "IceCores", "NNPI_ICE_CORES",
+      "Force compilation with specified amount of ice cores (1-12).", "-1"};
+  options[DeviceTraceOption] = {
+      "DeviceTracing", "NNPI_DEVICE_TRACING",
+      "Enabled device tracing (host2device, device2host copy infer etc.).",
+      "0"};
+  options[CompilationLogLevelOption] = {
+      "CompilationLogLevel", "NNPI_LOG_LEVEL",
+      "Sets the compilation logging level (0 most detailed).", "0"};
```
jfix71: nit: Perhaps mention what the max value accepted here is? Or even mention names of each level.

lib/Backends/NNPI/NNPIOptions.h
```diff
@@ -0,0 +1,183 @@
...
+  InferOnDeviceOption,
+  CompiledFileOption,
+  UseSymlowpOption,
+  DeviceVersionOption,
+  NumOfWorkersOption,
+  DeviceIDOption,
+  IceCoresOption,
+  DeviceTraceOption,
+  CompilationLogLevelOption,
+  OverideNNPIMemoryOption,
```
jfix71: nit: Should be spelled "Override"

lib/Backends/NNPI/NNPIOptions.cpp
```diff
@@ -0,0 +1,437 @@
...
+  numWorkers_ = NNPIOptions::getIntVal<llvm::StringMap<std::string>>(
+      NumOfWorkersOption, parameters);
+  enabledDeviceTraceing_ =
+      NNPIOptions::getBoolVal<llvm::StringMap<std::string>>(DeviceTraceOption,
+                                                            parameters);
+// Todo: remove if 1 section once nnpiDeviceGetInfo changes are implemented.
+#if 1
+  deviceMemory_ = 16000000;
+#else
+  deviceMemory_ = NNPIOptions::getUnsignedVal<llvm::StringMap<std::string>>(
```
jfix71: Why can't this just be enabled now, and override `deviceMemory_ ` if specified? I.e. just check if the specified value is the default 0 and if not then use it.

lib/Backends/NNPI/tests/NNPIDeviceManagerTest.cpp
```diff
@@ -18,7 +18,20 @@
 using namespace glow;
 
 std::set<std::string> glow::backendTestBlacklist = {
-    "MultiFunction/0",
-    "DeviceResidentTensors/0",
-    "TransferStaticPlaceholderTest/0",
-};
+    "MultiFunction/0", "DeviceResidentTensors/0",
+    "TransferStaticPlaceholderTest/0", "CanHandleDeviceResidentTensors/0"};
```
jfix71: nit: Please add a comma after the last item here so they are clang-formatted to one per line.

lib/Backends/NNPI/NNPIDeviceManager.h
```diff
@@ -37,6 +38,43 @@ namespace runtime {
 /// inference is executed at a time.
 class InferenceThreadEnv;
 
+class NNPIStaticPlaceHolderContainer {
```
jfix71: nit: Can we make this "Placeholder" everywhere, i.e. using a lowercase "Holder"? It will match our capitalization everywhere else.

lib/Backends/NNPI/NNPI.h
```diff
@@ -44,25 +42,34 @@ class NNPIBackend final : public Backend {
   bool shouldLower(const Node *N) const override;
   bool shouldShareBuffers() const override { return false; }
   bool supportsPartialTensors() const override { return true; }
+  bool supportsStaticPlaceholders() const { return true; }
```
jfix71: `override`

omromano:
Hi @jfix71, @arunm-git, made changes based on your comments. Please take a look.
Thanks.

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -308,6 +330,23 @@ bool NNPIBackend::shouldLower(const Node *N) const {
...
+        llvm::cast<FusedRowwiseQuantizedSparseLengthsSumNode>(N);
+    if ((backendOptions_.useIceT || backendOptions_.inferOnDevice) &&
+        (SLSN->getResult().getElementType() == ElemKind::Float16Ty)) {
+      return false; // Don't lower == keep without weights
+    } else {
+      return true;
+    }
+  }
+  case Kinded::Kind::SparseLengthsSumNodeKind:
+    // WA - lower until ICE-T implements it.
```
jfix71: So this is temporary because SLS is implemented in ICEREF but still waiting on the implementation to be ready for device, right?

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -308,6 +330,23 @@ bool NNPIBackend::shouldLower(const Node *N) const {
...
   case Kinded::Kind::BatchedReduceMinNodeKind:
     return false;
+  case Kinded::Kind::FusedRowwiseQuantizedSparseLengthsSumNodeKind: {
+    const FusedRowwiseQuantizedSparseLengthsSumNode *SLSN =
+        llvm::cast<FusedRowwiseQuantizedSparseLengthsSumNode>(N);
+    if ((backendOptions_.useIceT || backendOptions_.inferOnDevice) &&
+        (SLSN->getResult().getElementType() == ElemKind::Float16Ty)) {
+      return false; // Don't lower == keep without weights
+    } else {
+      return true;
```
jfix71: So we're lowering (to include weights of 1) only if on ICEREF? I expected there to always be an ICEREF version of anything supported on device, so this seems strange to me.

lib/Backends/NNPI/NNPI.cpp
omromano: @jfix71 You are correct.

lib/Backends/NNPI/NNPI.cpp
omromano: @jfix71 You are correct, ICE-Ref does not support this yet while the HW does.

PR #3880, opened 2019-12-09
Update CommonOperatorLoader to support offline weights 

gcatron:
Summary:
Add support for offline weights in CommonOperatorLoader, it now creates a static placeholder for each offline weight.
Documentation:

Test Plan: ninja check

### DIFF
--- a/include/glow/Importer/CommonOperatorLoader.h
+++ b/include/glow/Importer/CommonOperatorLoader.h
@@ -46,6 +46,8 @@ struct LoadWeightResult {
   /// Glow tensor containing quantization scales. This should only be non-null
   /// if there is more than 1 quantization parameter found.
   std::unique_ptr<Tensor> scales;
+  /// Type info of the weight, this is used for offline weights.
+  Type type;
 };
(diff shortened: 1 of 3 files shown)


include/glow/Importer/CommonOperatorLoader.h
```diff
@@ -150,19 +143,26 @@ class CommonOperatorLoader : public ProtobufLoader {
     if (in.dataType == ONNXIFI_DATATYPE_UINT8) {
```
jfix71: nit: Might make sense to move this to a switch statement. Then you can have a single check for `isOffline` like you did for the unquantized case.

include/glow/Importer/ProtobufLoader.h
```diff
@@ -134,7 +134,8 @@ class ProtobufLoader {
   /// Create a new Placeholder of type \p T, and register it
   /// under the name \p name. \returns The newly created placeholder.
```
jfix71: nit: Update doxygen

PR #3883, opened 2019-12-10
Pass more device data to backend's post lowering graph transformation

pjaaskel:
Summary:

Pass more device data to backend's post lowering graph transformation

First use case is the OpenCL backend of which actual device properties
can vary greatly. In the first use case, we check if there is actual
explicit scratchpad memory in the device which impacts the optimal
choice of kernels; if there is none, using the OpenCL "local" mem
as a temporary storage is often just an extra copy that should be
avoided if possible.

This gives around 80x speedup with pocl 1.5-pre (i7-7500U, pthread, LLVM 8)
and around 3x speedup with Intel OpenCL SDK for the same CPU. This is
measured with squeezenet (-time switch, wall clock time).

Test Plan:

Tested with clang8 build, g++ builds, 32b and 64b dim_t, OpenCL (pocl and Intel SDK).

For reference, the OpenCL squeezenet wall clock -time for this laptop's CPU+GPU:
```With the patch
--------------

* Intel GPU (HD620) (has local mem)

   5.3613 (100.0%)   0.0156 (100.0%)   5.3770 (100.0%)   5.3932 (100.0%)  Infer
   5.3613 (100.0%)   0.0156 (100.0%)   5.3770 (100.0%)   5.3932 (100.0%)  Total
   
* Intel CPU (i7-7500U)

   1.2382 (100.0%)   0.0277 (100.0%)   1.2658 (100.0%)   0.5523 (100.0%)  Infer
   1.2382 (100.0%)   0.0277 (100.0%)   1.2658 (100.0%)   0.5523 (100.0%)  Total

* pocl 1.5-pre (i7-7500U, LLVM 8.0, pthread)

   0.5970 (100.0%)   0.0032 (100.0%)   0.6002 (100.0%)   0.1540 (100.0%)  Infer
   0.5970 (100.0%)   0.0032 (100.0%)   0.6002 (100.0%)   0.1540 (100.0%)  Total

   
Without the patch
-----------------

* Intel GPU (HD620) (has local mem)

No significant difference (as expected).

* Intel CPU (i7-7500U)

   2.0702 (100.0%)   0.0273 (100.0%)   2.0974 (100.0%)   1.5960 (100.0%)  Infer
   2.0702 (100.0%)   0.0273 (100.0%)   2.0974 (100.0%)   1.5960 (100.0%)  Total

* pocl 1.5-pre (i7-7500U, LLVM 8.0, pthread)

  12.0850 (100.0%)   0.2214 (100.0%)  12.3064 (100.0%)  12.6355 (100.0%)  Infer
  12.0850 (100.0%)   0.2214 (100.0%)  12.3064 (100.0%)  12.6355 (100.0%)  Total

```

### DIFF
--- a/lib/Partitioner/Partitioner.cpp
+++ b/lib/Partitioner/Partitioner.cpp
@@ -394,6 +394,15 @@ void Partitioner::genBackendMap(
   }
 }
 
+const DeviceInfo &
+Partitioner::getDeviceInfoForBackend(llvm::StringRef backendName) {
+  for (DeviceInfo &devInfo : deviceInfo_) {
+    if (devInfo.backendName == backendName)
+      return devInfo;
+  }
+  llvm_unreachable("Each backend should have at least one device");
+}
+
 Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
     llvm::StringRef backendName, std::map<std::string, BackendInfo> &backendMap,
     CompilationContext &cctx) {
@@ -402,7 +411,8 @@ Expected<DAGListTy> Partitioner::createDAGWithoutPartition(
   for (auto F : module_->getFunctions()) {
     if (!optimized_) {
       auto backend = backendMap[backendName].backend;
-      RETURN_IF_ERR(::glow::optimizeFunction(F, *backend, cctx));
+      RETURN_IF_ERR(::glow::optimizeFunction(
+          F, *backend, cctx, &getDeviceInfoForBackend(backendName)));
     }
     std::unique_ptr<DAGNode> DAG0 = glow::make_unique<DAGNode>();
     DAG0->logicalDevices = {logDevice};
@@ -722,7 +732,9 @@ Partitioner::heterogeneousPartition(CompilationContext &cctx) {
     DCHECK(func->verify()) << "Conversion led to invalid function";
     // Step 2.1 : optimize a function if it has not been optimized yet.
     if (!optimized_) {
-      RETURN_IF_ERR(::glow::optimizeFunction(func, *backend, cctx));
+      RETURN_IF_ERR(::glow::optimizeFunction(
+          func, *backend, cctx,
+          &getDeviceInfoForBackend(backend->getBackendName())));
     }
 
     // Step 2.2 : apply graph partitioning algrithm to find out the partition.
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -44,6 +44,7 @@ struct BundleEntry {
 namespace runtime {
 
 class DeviceManager;
+struct DeviceInfo;
 struct DeviceConfig;
 
 } // namespace runtime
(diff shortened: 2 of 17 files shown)


opti-mix:
cc @gcatron @nickgg @jfix71 What do you think about passing more device data to backend's post lowering graph transformation?

lib/Partitioner/Partitioner.cpp
```diff
@@ -394,6 +394,15 @@ void Partitioner::genBackendMap(
   }
 }
 
+const DeviceInfo &
+Partitioner::getDeviceInfoForBackend(llvm::StringRef backendName) {
+  for (DeviceInfo &devInfo : deviceInfo_) {
```
gcatron: I think I'm missing something, where is deviceInfo_ getting set?

gcatron:
I like the idea of providing additional device info to the backend, It makes particular sense for the OpenCL backend where it targets a wide variance of hardware.  I'm not sure if plumbing it through the runtime vs a seperate config file makes more sense though? Unless the expectation is that the backend will compile differently based on the changing state of the hardware.

opti-mix:
One thing I'm a bit concerned about is that the partitioner provides a device information to the optimizer. But does partitioner really know which actual device will be used later? IIRC, it is decided later by the provisioner, or? Is it a problem?

cc @beicy

lib/Partitioner/Partitioner.cpp
beicy: @gcatron deviceInfo_ is passed to Partitioner from HostManager when Partitioner is created (i.e. Partitioner constructor)

beicy:
> One thing I'm a bit concerned about is that the partitioner provides a device information to the optimizer. But does partitioner really know which actual device will be used later? IIRC, it is decided later by the provisioner, or? Is it a problem?
> 
> cc @beicy

For our current Partitioner, we assume that if tow devices, e.g. card_1 and card_2,  have the same backend type e.g. T1, they should be identical. So if a partition is optimized for backend T1, Provisioner will use some algorithm to choose either card_1 or card_2 for this Partition.   In this PR, a new filed "availableLocalMemory" is introduced. If for the same type of backends, this value is fixed, our current Partitioner and Provisioner shouldn't be affect. However, if for same type of backends, this value could be different, our Partitioner and Provisioner should be improved, since now it is possible that one partition,  which is optimized use card_1's deviceInfo_,  won't be finally assigned to card_1.

pjaaskel:
I got the impression that for OpenCL in Glow, multiple device support is not implemented (at least not complete) in practice, so in essence the device will be always the same if targeting OpenCL. But of course in case of OpenCL, the actual device properties can vary greatly - it's about heterogeneous computing after all. 

Here the property of having a local memory or not (e.g. CPU device vs. a typical GPU device) affects the selection of a graph node, which lowers to a non-local mem using kernel in case there is no explicit local memory in the target device which has dramatic impact for CPU devices (and for other devices which don't have a scratchpad).

pjaaskel:
I rebased and fixed some 32b dim_t issues also. Any feedback to this one on how to get this in master?

pjaaskel:
Ping? Anything I can improve here?

gcatron:
> Ping? Anything I can improve here?

Hey @pjaaskel sorry for the delay, I think this looks good, can you rebase to get rid of the merge conflict and I'll see about getting it merged.

pjaaskel:
Done. I cannot reproduce the QuantizationTest fail that happens in the CI check. It doesn't seem related to my changes.

gcatron:
pretty sure 

> Done. I cannot reproduce the QuantizationTest fail that happens in the CI check. It doesn't seem related to my changes.

I think that's a flaky test, rerunning the test seems to pass.

pjaaskel:
What are the failures about? Anything I can do here?

gcatron:
> What are the failures about? Anything I can do here?

Sorry for the delay was out last week, looks like transformPostLowering wasn't updated in the habana backend.

pjaaskel:
@gcatron is it possible to build the Habana without having the HW, or are the headers/libs proprietary?

gcatron:
> @gcatron is it possible to build the Habana without having the HW, or are the headers/libs proprietary?

You do need additional libraries, I went ahead and fix it, should be merged soon. Thanks for you patience getting this landed.

PR #3892, opened 2019-12-11
[GraphOptz] Add numeric equivalence test for DCE

gaurav-work:
Summary:
Add a numeric equivalence test to the DCE sub-test in GraphOptzTest.
The test creates a network with some dead code and then calls
optimization function that would eliminate all of the dead code.
Confirm that all nodes are removed and that there is numeric
equivalence on the original and optimized functions (networks) on
random inputs.

Testing:
Both in Debug and Release mode

Test Plan:
$ ninja all && tests/GraphOptzTest --gtest_filter=GraphOptz.DCE
$ ninja test


gaurav-work:
Waiting for some CircleCI love.

PR #3893, opened 2019-12-11
improve names of grad nodes

nickgg:
Summary:
Improves readability of DAGs, Traces, logs, etc by adjusting names of GradNodes:
1. Don't use the name of the forward pass node directly.
2. Don't use a fixed name for all examples of a backwards pass node.
3. Include the name of the original node in all nodes generated during GradNode conversion.

Differential Revision: D18937144

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "glow/Graph/Grad.h"
+
 #include "glow/Base/Train.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/Nodes.h"
@@ -111,7 +112,7 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
 
     if (N->getKind() == Kind::SaveNodeKind) {
       // Swap the src and dest. Send the Zero value as gradient for both sides.
-      auto *X = new SplatNode(N->getName(),
+      auto *X = new SplatNode(N->getName().str() + "_grad",
                               cast<SaveNode>(N)->getInput().getType(), 0);
       toAppend.push_back(X);
       map.addGradient(cast<SaveNode>(N)->getInput(), X);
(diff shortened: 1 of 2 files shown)


lib/Graph/Grad.cpp
```diff
@@ -111,7 +112,7 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
 
     if (N->getKind() == Kind::SaveNodeKind) {
       // Swap the src and dest. Send the Zero value as gradient for both sides.
-      auto *X = new SplatNode(N->getName(),
+      auto *X = new SplatNode(N->getName().str() + "_grad",
```
opti-mix: Since this pattern: `x.str() + "_grad"` is used many times in your PR, I'd suggest creating a helper function for it. Something like `getGradName`.

lib/Graph/Grad.cpp
```diff
@@ -166,10 +168,10 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
           F->getParent()->uniqueTypeWithNewShape(TNInputType, BRAInputDims);
 
       auto *RN =
-          new ReshapeNode(TN->getName().str() + ".grad.reshape", BRAInputType,
+          new ReshapeNode(TN->getName().str() + "_grad_reshape", BRAInputType,
```
opti-mix: It seems like it would be also beneficial to have another helper(s) to create a compound name from name segments delimited by `_`, e.g. `getName(x, "grad_reshape")` or even  `getName(x, "grad", "reshape")`. It avoids an explicit string manipulation and hides the implementation details like the use of `_` as a delimiter.

lib/Graph/Grad.cpp
```diff
@@ -429,7 +442,7 @@ Function *glow::differentiate(Function *F, const TrainingConfig &conf,
                          conf.batchSize);
     toAppend.push_back(X);
     // Now update the weight with the value computed by SGD.
-    auto *save = new SaveNode(PH->getName().str() + ".saveGrad", {X, 0}, PH);
+    auto *save = new SaveNode(PH->getName().str() + "_saveGrad", {X, 0}, PH);
```
opti-mix: Why `saveGrad` and not `_save_grad` as in most other places in this PR?

SplitInfinity:
Also, I would prefer that we use [`llvm::formatv`](http://llvm.org/docs/ProgrammersManual.html#formatting-strings-the-formatv-function) instead of converting `llvm::StringRefs` to `std::string` and then concatenating string literals to them.

With `formatv`, `N->getName().str() + "_grad"` becomes `llvm::formatv("{0}_grad", N->getName())`.

nickgg:
cosmetic changes to every line plus #3890 moving all the code means i'll have to rewrite this, I'll see if i have time later.

PR #3896, opened 2019-12-12
Add SLWS in the interpreter for fp16

SameerAsal:
* Update the Interpreter to support RWQ-SLWS with data input
    of UInt8FusedQTy with the output tensor if type fp16. It
    will support a node that looks like:
      name : RQSLWS
      Data : ui8fused[S:0.0000 O:0][0.000,0.000]<3 x 9>
      Weights : float16<8>
      Indices : index64<8>
      Lengths : index32<4>
      UseFP16Accumulation : 0

  * Update the interpreter to decouple the output tensor type from
    the type for the fused Scale and Offset. The type of scale and
    offset for a fused quantized type should depend on the input.

Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:
  Added a new test
  ninja test passes

### DIFF
--- a/lib/Backends/Interpreter/InterpreterNodes.cpp
+++ b/lib/Backends/Interpreter/InterpreterNodes.cpp
@@ -3479,8 +3479,23 @@ void BoundInterpreterFunction::
     for (dim_t j = 0, e = LH.raw(i); j < e; j++) {
       const float weight = static_cast<float>(WH.raw(curIdx));
       const dim_t rowIdx = IH.raw(curIdx++);
-      T scale, offset;
-      std::tie(scale, offset) = DH.getFusedScaleOffsetFromRow<T>(rowIdx);
+      // Data type for the Scale and Offset for fused types need not follow
+      // the type for the output Tensor passed in T.
+      float scale, offset;
+      switch (
+          getScaleOffsetElemKindFromFused(data->getType().getElementType())) {
+      case ElemKind::FloatTy:
+        std::tie(scale, offset) = DH.getFusedScaleOffsetFromRow<float>(rowIdx);
+        break;
+      case ElemKind::Float16Ty:
+        std::tie(scale, offset) =
+            DH.getFusedScaleOffsetFromRow<float16_t>(rowIdx);
+        break;
+      default:
+        llvm_unreachable("Type is not supported");
+        break;
+      }
+
       for (dim_t k = 0; k < outLineSize; k++) {
         float d = 0.0f;
         if (!using4BitQuantization) {
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -8029,8 +8029,10 @@ TEST_P(OperatorTest,
       /* useFP16Accumulation */ true);
 }
 
-TEST_P(OperatorTest,
-       FusedRowwiseQuantizedSparseLengthsWeightedSum_ConvertedFloat16) {
+static void testRowwiseQuantizedSparseLengthsSum_ConvertedFloat16(
+    glow::PlaceholderBindings &bindings, glow::Module &mod, glow::Function *F,
+    glow::ExecutionEngine &EE, float allowedError,
+    bool convertFusedToFP16 = true) {
   CHECK_IF_ENABLED();
   /*
     DATA  =   [[2.0, -0.5, 13]]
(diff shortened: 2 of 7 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -7976,6 +7976,71 @@ TEST_P(OperatorTest, FusedRowwiseQuantizedSparseLengthsWeightedSum_Float) {
       bindings_, mod_, F_, EE_, ElemKind::UInt8FusedQTy, 0.0001);
 }
 
+/// Test Fused-RWQ-SLWS in where the weights are in Fp16, inputs are
+/// UInt8FusedQTy.
+TEST_P(OperatorTest,
+       FusedRowwiseQuantizedSparseLengthsWeightedSum_float16_weights) {
```
jfix71: nit: Can you capitalize `float16` so it's consistent with the rest of the tests?

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -3480,7 +3480,23 @@ void BoundInterpreterFunction::
...
       const dim_t rowIdx = IH.raw(curIdx++);
       T scale, offset;
-      std::tie(scale, offset) = DH.getFusedScaleOffsetFromRow<T>(rowIdx);
+
+      // Data type for the Scale and Offset for fused types need not follow
+      // the type for the output Tensor passed in T.
+      switch (
+          getScaleOffsetElemKindFromFused(data->getType().getElementType())) {
+      case ElemKind::FloatTy:
+        std::tie(scale, offset) = DH.getFusedScaleOffsetFromRow<float>(rowIdx);
```
jfix71: Hm, can you not just do `DH.getFusedScaleOffsetFromRow<T>` and skip the switch?

tests/unittests/OperatorTest.cpp
```diff
@@ -7976,6 +7976,71 @@ TEST_P(OperatorTest, FusedRowwiseQuantizedSparseLengthsWeightedSum_Float) {
       bindings_, mod_, F_, EE_, ElemKind::UInt8FusedQTy, 0.0001);
 }
 
+/// Test Fused-RWQ-SLWS in where the weights are in Fp16, inputs are
+/// UInt8FusedQTy.
+TEST_P(OperatorTest,
```
jfix71: This looks like an almost exact copy of `FusedRowwiseQuantizedSparseLengthsWeightedSum_ConvertedFloat16` -- can we create a helper out of the logic from test, that takes a bool for setting `cctx.precisionConfig.convertFusedToFP16`? E.g. something similar to `testFusedRowwiseQuantizedSparseLengthsWeightedSum`. Then we can reuse the logic for both tests.

tests/unittests/OperatorTest.cpp
```diff
@@ -8046,39 +8048,40 @@ TEST_P(OperatorTest,
...
+  bindings.allocate(S->getPlaceholder());
 
   CompilationContext cctx;
   cctx.precisionConfig.convertToFP16 = true;
-  cctx.precisionConfig.convertFusedToFP16 = true;
-  EE_.compile(cctx);
-  EE_.run(bindings_);
+  cctx.precisionConfig.convertFusedToFP16 = convertFusedToFP16;
+  EE.compile(cctx);
+  F->dump();
```
jfix71: nit: remove this?

tests/unittests/OperatorTest.cpp
```diff
@@ -8087,7 +8090,28 @@ TEST_P(OperatorTest,
...
+
+/// Test Fused-RWQ-SLWS in where the weights are in Fp16, data
+/// inputs are UInt8FusedQTy.
+TEST_P(
+    OperatorTest,
+    FusedRowwiseQuantizedSparseLengthsWeightedSum_ConvertedFloat16_NoFusedConvert) {
+  CHECK_IF_ENABLED();
+  return testRowwiseQuantizedSparseLengthsSum_ConvertedFloat16(
+      bindings_, mod_, F_, EE_, 0.02,
+      /* useFP16Accumulation */ false,
```
jfix71: This is false for both current calls, and it's not even being used inside `testRowwiseQuantizedSparseLengthsSum_ConvertedFloat16`, so let's remove it.

SameerAsal: I actually intentionally kept this to make the test helper more "configurable". I'll remove it  for now :)

PR #3900, opened 2019-12-12
Unsilo Interpreter Backend, use directly by GraphOptimizer to allow other backends to call passes directly, and wrap transformPostLowering in if again

jfix71:
Summary: We've had a lot of annoyances caused by requiring GraphOptimizer to depend on Backends, which prevents Backends from calling optimizations in their `transformPostLowering()` call. This PR makes the following changes to resolve this:

- Unsilo the Interpreter backend from `lib/` (move all headers to `include/`)
- Directly create an Interpreter backend for use during ConstantFolding instead of going through the factory in Backends
- Remove GraphOptimizer's dependence on Backends, and instead depend on the Interpreter
- Move back to if (transformPostLowering) in the graph optimization pipeline, as the backend can now call optimizations directly (essentially backs out https://github.com/pytorch/glow/pull/3859)


shajrawi:
cc @opti-mix

jfix71:
>Please check if backends can really invoke the optimizer inside `transformPostLowering` .

@opti-mix Verified that I could run a DCE pass from the CPU's `transformPostLowering()` via CMake with Mac and Linux VM, and via buck in fbcode.

PR #3726, opened 2019-11-01
bundleName_ in LLVMIRGen should get legalized.

navidR:
#3725

PR #3901, opened 2019-12-13
[AOT] Legalize bundle name. Change Loader default backend to CPU.

mciprian13:
**Summary**

Duplicates #3726 (mostly) but I need this landed kind of ASAP.
- Legalize bundle name.
- Legalize bundle main entry point.
- Removed **bundleName** and **outputDir** attributes from the BundleSaver class since they are redundant (duplicate the attributes of the LLVMIRGen class) and create even more confusion in the context of legalization.
- When building bundle, create directory if does not exist.
- Remove ".onnx" extension when building bundles for ONNX models without the "network-name" parameter. Otherwise the bundle artifacts will be named like this: **model.onnx.o** , **model.onnx.h** etc (and the inference function itself will be named **model_onnx**) which is kind of ugly IMHO.
- Change Loader default backend to CPU: currently the default backend is Interpreter so when building bundles without the "backend" parameter (which is not mandatory), the build crashes since the Interpreter cannot save a bundle.

**Documentation**
None

**Test Plan**
None

### DIFF
--- a/tools/loader/Loader.cpp
+++ b/tools/loader/Loader.cpp
@@ -400,6 +400,11 @@ static bool commandLineIsInvalid() {
                   end = llvm::sys::path::rend(modelPathOpt[0]);
              it != end; ++it) {
           networkName = *it;
+          // Strip extension (if any).
+          size_t lastDotPos = networkName.find_last_of(".");
+          if (lastDotPos != std::string::npos) {
+            networkName = networkName.substr(0, lastDotPos);
+          }
           // Empty names are replaced by '.' (see Path.h in LLVM).
           if (!networkName.empty() && networkName != ".") {
             break;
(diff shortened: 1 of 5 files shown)


tools/loader/Loader.cpp
```diff
@@ -173,7 +173,7 @@ llvm::cl::list<std::string> doNotLowerNodesForProfilingOpt(
 llvm::cl::opt<std::string> ExecutionBackend(
     "backend",
     llvm::cl::desc("Backend to use, e.g., Interpreter, CPU, OpenCL:"),
-    llvm::cl::init("Interpreter"), llvm::cl::cat(loaderCat));
+    llvm::cl::init("CPU"), llvm::cl::cat(loaderCat));
```
opti-mix: I'm a bit concerned about this specific change. Glow's CPU backend does not support all of the features provided by the Interpreter backend (e.g. fp16 support and some other features). Thus changing the default may have some unintended consequences.

mciprian13: I can change it back if you want (not very important for me).

mciprian13: Done. Changed the default backend from CPU back to Interpreter.

mciprian13:
Just a reminder ... when you land this you can also close #3726.

opti-mix:
@mciprian13 Could you squash your PR it into a single commit, which does not touch the default backend setting? It would be silly to have two commits here.

mciprian13:
I squashed the commits as suggested.

PR #3914, opened 2019-12-16
Add Backend option to Backend creation

omromano:
Summary: 
Add am option to pass  backend options on backend creation to allow multiple backend instances with different configurations.


omromano:
CC: @jsubag

opti-mix:
cc @rdzhabarov

PR #3918, opened 2019-12-17
Improve some lowered and auto differentiated Node names

nickgg:
Summary: preprends the name of the node causing creation of lowered or differentiated nodes, making it easier to debug.

Documentation: N/A

[Optional Fixes #issue]

Test Plan: ninja check

### DIFF
--- a/lib/Graph/Grad.cpp
+++ b/lib/Graph/Grad.cpp
@@ -22,6 +22,7 @@
 #include "glow/Support/Support.h"
 
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/raw_ostream.h"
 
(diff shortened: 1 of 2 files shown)


lib/Graph/Grad.cpp
```diff
@@ -30,6 +31,9 @@ using namespace glow;
 using llvm::cast;
 using llvm::isa;
 
+#define DECORATE_NODE_NAME(Node, ...)                                          \
```
opti-mix: Nit: Why is this macro defined in two places? Should this definition moved into one place, e.g. into a header? Also, does it need to be a macro? It seems like LLVM defines join_items as a template function with variadic arguments.

nickgg: It has slightly different semantics in both places (ptr vs ref). Also figured it was maybe better not to pollute the glow namespace with this helper.

Macro instead of just calling llvm::join_items to avoid replicating the `Node->getName()` everywhere.

PR #3920, opened 2019-12-17
Update NNPI Backend to v0.4.1.7

omromano:
Summary: Updating NNPI Backend to version 0.4.1.7

### DIFF
--- a/lib/Backends/NNPI/CMakeLists.txt
+++ b/lib/Backends/NNPI/CMakeLists.txt
@@ -124,6 +124,7 @@ target_link_libraries(NNPI
                       Backend
                       Base
                       Graph
+                      GraphOptimizer
                       CodeGen
                       IR
                       Support
--- a/lib/Backends/NNPI/NNPIOptions.h
+++ b/lib/Backends/NNPI/NNPIOptions.h
@@ -18,6 +18,8 @@
 
 #include "nnpi_transformer_types.h"
 #include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/raw_ostream.h"
 #include <map>
 #include <string>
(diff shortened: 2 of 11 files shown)


omromano:
Hi @jfix71, @arunm-git, 
Please take a look.
Thanks!
CC: @jsubag

lib/Backends/NNPI/NNPIOptions.h
```diff
@@ -18,166 +18,306 @@
...
-  virtual std::string dumpStatus() const override;
-  /// Lists the supported options (name=>description).
-  static llvm::StringMap<std::string> getSupportedOptions();
   /// Compile to ICE-T if true, ICE-Ref otherwise.
-  bool useIceT;
-  /// Enable execution on device (useIceT_ and !inferOnDevice_ will compile but
+  DECLARE_NNPI_OPTION(useIceT, bool, "UseIceT", "Use inference api",
+                      "USE_ICE_T", "0");
+  /// Enable execution on device (useIceT and !inferOnDevice will compile but
   /// not execute inference).
```
jfix71: `inferOnDevice` implies `useIceT`, correct? Can we expand on some descriptions of what happens here?

lib/Backends/NNPI/NNPIOptions.h
```diff
@@ -18,166 +18,306 @@
...
-  /// Lists the supported options (name=>description).
-  static llvm::StringMap<std::string> getSupportedOptions();
-  /// Compile to ICE-T if true, ICE-Ref otherwise - this variable cab be
-  /// verridden by inferOnDevice_.
-  bool useIceT;
-  /// Enable execution on device (useIceT_ and !inferOnDevice_ will compile but
+  /// Compile to ICE-T if true, ICE-Ref otherwise.
+  DECLARE_NNPI_OPTION(useIceT, bool, "UseIceT", "Use inference api",
+                      "USE_ICE_T", "0");
+  /// Enable execution on device (useIceT and !inferOnDevice will compile but
```
jfix71: Same as other comment -- can we expand on descriptions of interactions between `useIceT` and `inferOnDevice`?

arunm-git:
@omromano : did this PR build cleanly for you? Note that the NNPI backend is disabled by default in pytorch/glow and to actually get it to build you have to set GLOW_WITH_NNPI.
There are 3 build breaks:
* Missing #include <sstream> in NNPIOption.cpp
* NNPIDeviceManager.cpp: the change to variable name from  result to res breaks since the LOG_NNPI_INF_ERROR uses res inside the macro.
* InferencePool.cpp: same error as above with the result -> res change.
(The last two breaks were present in the 4.1.2 PR and that's why I renamed to result)

lib/Backends/NNPI/CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ target_link_libraries(NNPI
                       CodeGen
                       IR
                       Support
+                      GraphOptimizer
```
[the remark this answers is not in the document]
omromano: Compilation failed for us without it after this commit: a0f8e3beb1e398b5e279e2bbf22b98c11f3e3e5a
(lowerNode not defined)

omromano:
Hi @arunm-git, 
Compilation did not fail for us, but based on your comment we fixed the issues you have described.
Thanks!

omromano:
Hi @jfix71, 
I did not change the option definitions (compiledFile, customDspKernelsFile etc.) to llvm::StringRef since these are not constants, these are loaded from the environment variables and parameter maps (reverted back to std::string).

lib/Backends/NNPI/NNPIOptions.h
```diff
@@ -18,166 +18,309 @@
...
   /// specified.
-  std::string compiledFile;
-  /// Setting this variable will force compilation to the to use no more than
-  /// the amount of ice cores (1-12) - default 0.
-  int iceCores;
+  DECLARE_NNPI_OPTION(compiledFile, std::string, "CompiledFile",
+                      "Sets a file name to save the compilation output to the "
+                      "filename specified.",
+                      "ICE_T_FILE", "");
+  /// Force compilation with maximum amount of ice cores, -1 for unlimited.
```
jfix71: Curious -- what does unlimited mean here, as opposed to 12 as the actual max?

omromano: We did not want to define a number limit since future version may support a different number of cores.

Issue #3915, opened 2019-12-16
calculateTensorViewOffset should take strides/padding into account


opti-mix:
calculateTensorViewOffset currently does not take strides/padding into account and thus produces wrong offsets for tensors which are not densely packed and have non-trivial alignment requirements. This in turn leads to wrong results during the computation.

SameerAsal:
@opti-mix  You mentioned there is a file to look at for this. can you please point it out.

opti-mix:
@SameerAsal `calculateTensorViewOffset` is defined in `lib/LLVMIRCodeGen/AllocationsInfo.cpp`

PR #3922, opened 2019-12-17
Fix TensorView offsets to account for strides

jfix71:
Summary: See title. This was not updated when we added strides in https://github.com/pytorch/glow/pull/3368.

Fixes https://github.com/pytorch/glow/issues/3915

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -956,6 +956,23 @@ public:
   }
 };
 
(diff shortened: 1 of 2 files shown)


opti-mix:
@jfix71 Thanks a lot! We checked that it fixed the issue we were observing!

opti-mix:
@kunmingho will try later to create a unit test for it, but in a separate PR.

PR #3923, opened 2019-12-17
[DRT] add method for updating a device Tensor without copying it to host

nickgg:
Summary: Adds a new method to Tensor: `copyRawToDevice` which is an equivalent to `copyRawFrom` when the target is device resident. This is particularly useful for Inputs which are write only on the device.

Documentation: N/A

Test Plan: new test in DeviceManagerTests, since you need a DeviceManager to make a Tensor device Resident.

### DIFF
--- a/include/glow/Base/Tensor.h
+++ b/include/glow/Base/Tensor.h
@@ -747,6 +747,10 @@ public:
   /// associated device memory.
   void ensureOnHost();
 
+  /// Updates contents of a device resident Tensor with the data from \p t
+  /// without copying its contents to host.
+  void copyRawToDevice(const Tensor *t);
+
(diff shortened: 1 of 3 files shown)


include/glow/Base/Tensor.h
```diff
@@ -747,6 +747,10 @@ class Tensor final {
   /// associated device memory.
   void ensureOnHost();
 
+  /// Updates contents of a device resident Tensor with the data from \p t
+  /// without copying it's contents to host.
```
SplitInfinity: `it's` -> `its`

nickgg: it is.

PR #3924, opened 2019-12-17
Use Node names for vertices in dotty DAG dump

nickgg:
Summary: Now that all nodes in a Function, and all Storage names are unique, we can use the node name as the vertex name in the DAG output rather than a random string of numbers. This has the advantage of making the same node from different dot files use the name Vertex name, which i need for a _\~secret project\~_.

Documentation: N/A

Test Plan: ran unit tests, dumped a DAG and visualized it.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -276,14 +276,19 @@ protected:
     os << "}";
   }
 
-  void dumpNode(Node *N) {
+  void dumpNode(Node *N, bool uniqueNames) {
     if (!N) {
       return;
     }
     std::ostringstream os;
     // Print a node descriptor that looks like this:
-    // vNNNN [ shape = "record" label = "{...}" ];
-    os << uniqueVertexName(N) << "[\n";
+    if (uniqueNames) {
+      // vNNNN [ shape = "record" label = "{...}" ];
+      os << uniqueVertexName(N) << "[\n";
+    } else {
+      // <name> [ shape = "record" label = "{...}" ];
+      os << N->getName().str() << "[\n";
+    }
     os << "\tlabel = \"";
     dumpLabel(N, os);
     os << "\"\n";
@@ -381,7 +386,7 @@ class ModuleDottyPrinter : public AbstractDottyPrinter {
 public:
   void visitModule(Module *M) {
     for (auto N : M->getConstants()) {
-      dumpNode(N);
+      dumpNode(N, true);
     }
 
     for (auto F : M->getFunctions()) {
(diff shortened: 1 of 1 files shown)


lib/Graph/Graph.cpp
```diff
@@ -276,14 +276,18 @@ class AbstractDottyPrinter {
...
+  void dumpNode(Node *N, bool uniqueNames) {
     if (!N) {
       return;
     }
     std::ostringstream os;
-    // Print a node descriptor that looks like this:
-    // vNNNN [ shape = "record" label = "{...}" ];
-    os << uniqueVertexName(N) << "[\n";
+    if (uniqueNames) {
+      // Print a node descriptor that looks like this:
```
jfix71: Think this comment should be bumped back above the if block here.

lib/Graph/Graph.cpp
```diff
@@ -381,7 +385,7 @@ class ModuleDottyPrinter : public AbstractDottyPrinter {
 public:
   void visitModule(Module *M) {
     for (auto N : M->getConstants()) {
-      dumpNode(N);
+      dumpNode(N, true);
```
jfix71: Why do we need to still use the random string of numbers in this case?

nickgg: Because a module may have multiple functions with the same nodes, however they should not be the same vertices in the dot output.

lib/Graph/Graph.cpp
nickgg: The node descriptor doesn't look like that though, would you like me to add a comment with how the other node descriptor looks?

jfix71: Oh -- well I think the original comment was referring to everything being printed, not just the initial vertex name. NBD.

lib/Graph/Graph.cpp
jfix71: Isn't this handling Constants though? Clearly I'm confused 🤷‍♂ maybe add a comment explaining what's going on?

nickgg: It's a bit intricate, yeah. This block is handling constants, and is setting the Vertex name of the constant so we can add edges to it later. In this path where we're dumping the Module we use a version of the Function that does node uniquing since it might have multiple vertices for the same Node name. Since we're doing uniquing all vertices need to use the same uniquing rule in order for us to create the right edges.

TLDR: use node names for single function dumping, use uniqued names for Module dumping.

lib/Graph/Graph.cpp
nickgg: I'll clean this up.

PR #3925, opened 2019-12-18
[PyTorch] Remove flaky redundent unit test

zrphercule:
Given we have test_quantized_linear_packed_rowwise already, this test is weak enough to be removed, but strong enough to cause the off-by-one problem because the input is totally random floating number.


yinghai:
@zrphercule Is this still valid?

zrphercule:
> @zrphercule Is this still valid?

Yes, actually this could occationally break internal CI, since it has some randomness.

yinghai:
Shall we mark it as skipped?

zrphercule:
> Shall we mark it as skipped?

Sure I can do that.

PR #3926, opened 2019-12-18
Fix for windows break build. 

ayermolo:
Summary:
The VS implementation of promise was trying to access protected constructor.
Honestly not 100% why default constructor was made explicit, or protected.
Documentation:

[Optional Fixes #issue]

Test Plan:
Unit Tests

### DIFF
--- a/include/glow/Support/Error.h
+++ b/include/glow/Support/Error.h
@@ -379,9 +379,9 @@ class GlowError : protected detail::CheckState<detail::enableCheckingErrors> {
     return std::move(errorValue_);
   }
 
-protected:
+public:
   /// Construct a new empty Error.
-  explicit GlowError() { setErrorValue(nullptr, /*skipCheck*/ true); }
+  GlowError() { setErrorValue(nullptr, /*skipCheck*/ true); }
 
 public:
   /// Construct an Error from an ErrorValue \p errorValue.


gcatron:
This looks acceptable to me, @jackm321 what do you think?

ayermolo:
@gcatron @jackm321  Can this land, or some changes required? Seems like internal tests are failing?

jackm321:
Hey @ayermolo, thanks for working on this and sorry that it's continued to cause issues for you with VS. I'd like to say though, that I'm personally not in favor of this particular change.

The reason this constructor is protected and explicit is so that empty Errors cannot be created except by intensionally calling `GlowError::success()` (and now `GlowError::empty()` for passing into constructors since that ended up being a popular pattern). To disable this will enable implicit construction of empty Errors in many situations (including promises) without any idea of what their meaning is. This pattern comes from the original llvm::Errors that we used to be using here.

It seems like VS is probably counting on default constructing objects in promises before they're filled? It seems weird to me but to avoid the issue, I'd prefer to just continue to try to not use Errors in promises than change Errors themselves to make them less functional. Thoughts?

ayermolo:
@jackm321 
Yeah that makes sense. Thanks for explanation. I'll push another MR, that doesn't uses promise.

As for why VS has this issue. Short version it's an issue in latest MSVC 2017 and 2019, with implementation relying on class it's instantiated with having default constructor.
It was brought up and fixed in some branch, but apparently never propagated to release versions.

https://developercommunity.visualstudio.com/content/problem/60897/c-shared-state-futuresstate-default-constructs-the.html

ayermolo:
Closing this one.
Please see: https://github.com/pytorch/glow/pull/3957/

PR #3938, opened 2019-12-21
Fix memory leak

mjanderson09:
Summary: commands array is never deleted. Changing it to local variable.

Differential Revision: D19204847


mjanderson09:
@jsubag Can you take a look?

@jfix71

jsubag:
LGTM - nice catch.

PR #3951, opened 2019-12-30
[OpenCL] Fix a bug when running an external model with the image-loader transposing the input to NCHW

### LINKED ISSUE #3815: [OpenCL] Convolution NHWC vs. NCHW mismatch
After fixing Issue #3802, inception_v1 now fails due to a mismatching layouts error. Should there be automatic conversion operation insertion in such a case?

```
./bin/image-classifier -backend=OpenCL tests/images/imagenet/dog_207.png -expected-labels=207 -image-mode=0to255 -m=../build-llvm-7/inception_v1 -model-input-name=data 
Model: ../build-llvm-7/inception_v1
Running 1 thread(s).



In 'conv1_7x7_s2__6' From '../build-llvm-7/inception_v1'
input 0
Convolution
name : conv1_7x7_s2__6
Input : float<1 x 3 x 224 x 224>
Filter : float<64 x 3 x 7 x 7>
Bias : float<64>
Kernels : [7, 7]
Strides : [2, 2]
Pads : [3, 3, 3, 3]
Group : 1
Dilation : 1
Layout : NCHW
FusedActivation : 
users : 1
Result : float<1 x 64 x 112 x 112>

Mismatching layouts:
Provided layout
Layout: NHWC [name = N : alignment = 1 : index = 0, name = H : alignment = 1 : index = 1, name = W : alignment = 1 : index = 2, name = C : alignment = 1 : index = 3]
Expected layout
Layout: NCHW [name = N : alignment = 1 : index = 0, name = C : alignment = 1 : index = 1, name = H : alignment = 1 : index = 2, name = W : alignment = 1 : index = 3]
From '../build-llvm-7/inception_v1'
Expected correct backend-specific layouts for the graph
For comparison `LHS Equal RHS` with:
LHS: 0
RHS: 1
F1123 11:26:22.983803 15371 Error.cpp:119] exitOnError(Error) got an unexpected ErrorValue: 
Error code: COMPILE_UNSUPPORTED_NODE_AFTER_OPTIMIZE
Error message: Unsupported node(s) found after optimizing Function ../build-llvm-7/inception_v1 for backend OpenCL
Error return stack:
../lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp:3293
../lib/Partitioner/Partitioner.cpp:401
../tools/loader/Loader.cpp:505
*** Check failure stack trace: ***
Aborted (core dumped)
```

Any pointers? We can take a look if someone pokes us to the right direction.

shajrawi:
Summary:
See the in-source comment for workaround information, but: We have a model we load from the outside with no-way of knowing the constant/placeholder input layout, the default assumption for 4-D tensors (images) is NHWC format which is the canonical Glow format, PNG files are in NHWC format.
Our image loader, when using the `image-layout` flag, transposes the image outside the Glow graph, since there's no easy way to propagate that information, weaken the OpenCL verifier, not the canonical verifier: for placeholders and constants, assume that the loader knows what it is doing and they are in the right format.

Fixes #3815

Test Plan:
`ninja test`

### DIFF
--- a/lib/Backends/OpenCL/OpenCLTensorLayout.cpp
+++ b/lib/Backends/OpenCL/OpenCLTensorLayout.cpp
@@ -44,6 +44,14 @@ static std::string returnBaseReqOrNHWC(TensorLayoutDescription &baseReq,
     // These nodes accept any 4-D layout.
     return baseReq.getSerializedLayout();
   }
+  // For Placeholders and Constants that were loaded from another tool, we don't
+  // have the layout information in during load time. This makes us assume they
+  // are in Glow's Canonical NHWC format, which is not correct if we are loading
+  // an image with NCHW such as in resent loader. Weaken the verifier to avoid a
+  // runtime crash: if this is a storage location, return the baseReq.
+  if (llvm::dyn_cast<Storage>(node)) {
+    return baseReq.getSerializedLayout();
+  }
 
   return CanonicalTensorLayout::getInstance().getDefaultNDLayout(4);
 }


SameerAsal:
Would it be useful to propagate the layout from the loader into Glow placeholder? We can track that t in an issue?

shajrawi:
The ideal solution would have the loader adding a transpose node to the graph instead of the way `readPngImageAndPreprocess` currently does things, but that's a huge architectural change

PR #3953, opened 2020-01-02
[GraphOptz] Add helper to check whether a Node is in a Function

SplitInfinity:
**Summary**
The tests in `GraphOptzTest` use the helper predicate `IsSameNodeAddress` in
conjunction with `std::find_if` to check whether a `Node` is still in a
`Function` after it has been optimized. This commit introduces a helper
function `functionContainsNode` that contains this logic and replaces all of
the calls to `std::find_if` with `IsSameNodeAddress` with this new function
for improved readability. This function will also be used frequently
while adding numerical equivalence checks to more tests.

**Test Plan**
All unit tests pass.

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -28,6 +28,27 @@ using namespace glow;
 
 class GraphFold : public GraphOptz {};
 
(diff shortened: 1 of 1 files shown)


SplitInfinity:
This pull request is a refresh of #3812 which was closed by stalebot before I could land it.

PR #3955, opened 2020-01-03
[ONNX Importer] Export all the RNN, GRU and LSTM outputs.

mciprian13:
**Summary**
Up until now the the RNN, GRU and LSTM defined only 1 output.
When I implemented these I did not have an example of a node with multiple outputs.
The change is pretty trivial.
This fixes the issue #3939.

Test Plan:
None


mciprian13:
I modified the existing unit tests to use (export) all the outputs of the ONNX node definition.

PR #3959, opened 2020-01-03
Fix for Function::clone() not cloning Predicates

UncleGene:
Summary: Function::clone() behavior fixed to preserve predicates from the original function nodes.

Test Plan: New unit test, fails without the fix and succeeds with it. All tests pass.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -4018,6 +4018,9 @@ Function *Function::clone(llvm::StringRef newName,
     // Record the copy relationship between the graphs.
     currToNew[&N] = copy;
     newF->addNode(copy);
+    if (N.hasPredicate()) {
+      copy->setPredicate(N.getPredicate());
+    }
   }
 
   // At this point we have a new invalid function that points into nodes in
@@ -4037,6 +4040,13 @@ Function *Function::clone(llvm::StringRef newName,
 
       // Update the node with the edge to the current graph.
       N.setNthInput(inp, NodeValue(it->second, input.getResNo()));
+    }
+
+    if (N.hasPredicate()) {
+      auto it = currToNew.find(N.getPredicate().getNode());
+      if (it != currToNew.end()) {
+        N.setPredicate(NodeValue(it->second, N.getPredicate().getResNo()));
+      }
     }
   }
 
--- a/tests/unittests/GraphTest.cpp
+++ b/tests/unittests/GraphTest.cpp
@@ -32,6 +32,16 @@
 
 using namespace glow;
 
+// Helper to find a node in the Function by name
+static const Node *nodeByName(const Function *F, const std::string &name) {
+  for (auto &n : F->getNodes()) {
+    if (n.getName().str() == name) {
+      return &n;
+    }
+  }
+  return nullptr;
+}
+
 TEST(Graph, testVariableErasure) {
   Module MD;
   auto &vars = MD.getConstants();
@@ -642,7 +652,7 @@ TEST(Graph, functionDependenciesTest) {
   M.dumpDAG();
 }
 
-TEST(Graph, cloneTest2) {
+TEST(Graph, functionCloneTest) {
   Module M;
   PlaceholderBindings bindings;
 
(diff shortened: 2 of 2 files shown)


lib/Graph/Graph.cpp
```diff
@@ -4018,6 +4018,9 @@ Function *Function::clone(llvm::StringRef newName,
     // Record the copy relationship between the graphs.
     currToNew[&N] = copy;
     newF->addNode(copy);
+    if (N.hasPredicate()) {
+      copy->setPredicate(N.getPredicate());
```
opti-mix: Shouldn't setPredicate argument refer to a clone of the original predicate? Currently it seems to use the original predicate instead of the copy.

UncleGene: Predicate is NodeValue, and (AFAIK) is transferred by value

opti-mix: > Predicate is NodeValue, and (AFAIK) is transferred by value

Yes, it is a NodeValue. But it should refer to the node in the new cloned graph, right? Could you check in your test that the predicate in the cloned graph actually refers to the node in the cloned graph?

UncleGene: Yep, good idea. But I also have a design question: should cloning predicate be a part of function clone, or should Node own it (though implementation would be trickier, as node clone is generated method for each kind)?

UncleGene: Actually, predicate refers to placeholder "pred". Function::clone does not copy any placeholders.

opti-mix: > Actually, predicate refers to placeholder "pred". Function::clone does not copy any placeholders.

But can a predicate be a non-placeholder, but a result of a node being computed during during the run of a model? If so, can we have a test to check that that the predicate in the cloned graph actually refers to the node in the cloned graph?

UncleGene: Perhaps I do not understand something, but why should it refer to the node in the same graph?  I  believe the same predicate can be shared by multiple nodes (at least by implementation), not necessarily from the same graph.
If this is misunderstanding, can you please provide a code snippet for the tree you have in mind.

opti-mix: > Perhaps I do not understand something, but why should it refer to the node in the same graph? I believe the same predicate can be shared by multiple nodes (at least by implementation), not necessarily from the same graph.

I'm not saying a node with a predicate should always refer to the predicate node in the same graph. But it could refer to a predicate in the same graph in some cases. Do you agree with this? 

I'm just saying it would be nice to have a test for such a situation. Such a test should check that if a node with a predicate which was referring to a predicate node in the original graph, then the cloned node with a predicate in the cloned graph should refer to the predicate in the cloned graph, because if it does not hold and it refers to the predicate in the original graph and the original graph is removed after cloning, the reference to the predicate in the cloned graph would refer to a removed predicate from the original graph.

Another test could be for testing the situation where the same predicate is shared by multiple predicated nodes in possible multiple graphs. This is the case you are talking about. In such a case, you probably want to test that after cloning there is still only one shared version of a predicate and this predicate node does not belong to any concrete graph (e.g. it is a placeholder).

UncleGene: I see what you mean. Can you still help with a more-or-less realistic code example of adding a predicate node to the graph?

opti-mix: Please have a look e.g. at `TEST_P(OperatorTest, simplePredication)`. It has a predicate defined like this:
` Node *pred = F_->createCmpLTE("cmp", C5, counters);`

UncleGene: Thank you! I'll add tests (and fix the code, you are right, this case is not covered)

tests/unittests/GraphTest.cpp
```diff
@@ -642,30 +652,63 @@ TEST(Graph, functionDependenciesTest) {
...
+
+  auto *newF = F->clone("new_main");
+
+  EXPECT_TRUE(newF->verify());
+  EXPECT_EQ(newF->getNodes().size(), F->getNodes().size());
+  EXPECT_EQ(newF->getParent(), F->getParent());
+
+  EXPECT_EQ(reluExt->getPredicate().getNode(), externalPredicate);
+  EXPECT_EQ(reluInt->getPredicate().getNode(), internalPredicate);
+  EXPECT_EQ(
```
opti-mix: Could you add comments for this check and for the subsequent check describing what exactly they try to ensure?

PR #3962, opened 2020-01-04
[OpenCL] add an option to cache programs on disk

pjaaskel:
Summary:

When a directory is given with -opencl-program-cache-dir, the OpenCL backend looks in it for cached OpenCL programs for the targeted device and tries to load from the found binary instead of building from the sources.

If it fails to load the binary, it assumes the file is corrupted or for a wrong device (or driver version) and rebuilds it.

When hitting the disk cache, I measured the following speedups with resnet50 (best-of-ten):

* Intel HD 620 GPU: From 7.0 s (cold cache) down to 0.3 s (hot cache).

* Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz (Intel's OpenCL SDK): From 2.7 s (cold cache) down to 1.3 s (hot cache).

* pocl 1.5-pre (LLVM 9) for the same CPU: From 1.5 s (pocl's internal kcache helps) down to 1.0 (hot cache).

Test Plan:

ninja check with CPU/LLVM7, OpenCL/pocl, OpenCL/IntelCPU, OpenCL/IntelGPU

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -38,8 +38,11 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
+
+#include <fstream>
 
 #define DEBUG_TYPE "opencl"
 
@@ -84,6 +87,13 @@ llvm::cl::opt<bool> clDoProfile("opencl-profile",
                                 llvm::cl::desc("Profile OpenCL kernels"),
                                 llvm::cl::init(false),
                                 llvm::cl::cat(OpenCLBackendCat));
+
+llvm::cl::opt<std::string> clDeviceProgramCacheDir(
+    "opencl-program-cache-dir",
+    llvm::cl::desc("The program disk cache directory for the "
+                   "used device. If empty, disk caching "
+                   "disabled."),
+    llvm::cl::init(""), llvm::cl::cat(OpenCLBackendCat));
 
 static void dumpCompileLog(cl_device_id dev, cl_program prog) {
 #ifndef NDEBUG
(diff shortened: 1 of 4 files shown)


utils/scripts/gen_onnx_rnn_model.py
```diff
@@ -173,6 +173,7 @@ def compute_rnn(forward):
         Wi, Ri, bWi, bRi = get_weights(dir_idx)
 
         def f(x): return np.tanh(x)
+
```
opti-mix: Unrelated change?

utils/scripts/gen_onnx_gru_model.py
```diff
@@ -190,7 +190,9 @@ def compute_gru(forward):
             dir_idx)
 
         def f(x): return (1 / (1 + np.exp(-x)))
+
```
opti-mix: Unrelated changes in this file?

tests/unittests/PartitionerTest.cpp
```diff
@@ -541,7 +541,7 @@ static void createSimpleSparseNNModule(Module &mod) {
   for (int table = 0; table < 5; table++) {
     Tensor data(ElemKind::FloatTy, {tableEntries, tableWidth});
     auto indices = mod.createPlaceholder(
-        ElemKind::Int64ITy, {numIndices * batchSize}, "indices", false);
+        IndexElemKind, {numIndices * batchSize}, "indices", false);
```
opti-mix: Unrelated change?

tests/unittests/OperatorTest.cpp
```diff
@@ -7756,14 +7756,13 @@ createAndInitRWQSLWSAllSame(glow::PlaceholderBindings &bindings,
       0.2244578,  0.44881952, 0.42696562, 0.33007848, 0.4511249,  0.11568925,
       0.02629679, 0.33864713, 0.42614424};
 
-  Placeholder *indices =
-      mod.createPlaceholder(ElemKind::Int64ITy, {21}, "indices",
-                            /* isTrainable */ false);
+  Placeholder *indices = mod.createPlaceholder(IndexElemKind, {21}, "indices",
```
opti-mix: Unrelated changes in this file?

lib/Graph/Nodes.cpp
```diff
@@ -1291,7 +1291,7 @@ static bool verifyFusedRowwiseQuantizedSparseLengthsSum(
   // For EmbeddingBagByteRowwiseOffsets lengths are really offsets and should be
   // Int64ITy.
   if (isEmbeddingBagByteRowwiseOffsets) {
-    isValid &= checkType(lengths, ElemKind::Int64ITy, parent);
+    isValid &= checkType(lengths, IndexElemKind, parent);
```
opti-mix: Unrelated change?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -201,17 +301,42 @@ OpenCLFunction::createProgram(const std::string &source,
...
+    CHECK(program) << "clCreateProgramWithSource failed.";
+  }
   err = clBuildProgram(program, 0, nullptr, combinedOptions.c_str(), nullptr,
                        nullptr);
   if (err) {
     dumpCompileLog(deviceId, program);
   }
   CHECK_EQ(err, CL_SUCCESS) << "clBuildProgram Failed.";
 
+  if (useDiskCache)
```
opti-mix: Please use {} around the single-line then-statement.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +178,100 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+  CHECK_EQ(errC, CL_SUCCESS)
+      << "clGetProgramInfo for CL_PROGRAM_BINARIES failed.";
+
+  int fileDesc = open(programPath.c_str(), O_WRONLY | O_CREAT | O_TRUNC,
+                      S_IRUSR | S_IWUSR);
+
+  CHECK_GT(fileDesc, -1) << "Error opening " << programPath << " for writing.";
+
+  ssize_t numWritten = write(fileDesc, bin, binSize);
+  if (numWritten == -1)
```
opti-mix: Please use {} around the single-line then-statement.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +178,100 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+    perror("Could not write binary:");
```
opti-mix: Why perror instead of CHECK_* or some other mechanisms used in Glow?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +178,100 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+      << "clGetProgramInfo for CL_PROGRAM_NUM_DEVICES failed";
+  CHECK_EQ(numDevices, 1) << "Only one OpenCL device supported";
+
+  size_t binSize = 0;
+  errC = clGetProgramInfo(program, CL_PROGRAM_BINARY_SIZES, sizeof(size_t *),
+                          &binSize, nullptr);
+  CHECK_EQ(errC, CL_SUCCESS)
+      << "clGetProgramInfo for CL_PROGRAM_BINARY_SIZES failed";
+
+  unsigned char *bin = new unsigned char[binSize];
```
opti-mix: Maybe use unique_ptr?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -38,9 +38,14 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <fcntl.h>
```
opti-mix: Are all of these headers supported on Windows as well?

utils/scripts/gen_onnx_rnn_model.py
pjaaskel: Yes, not directly related, but these cleanups and 32b dim_t fixes which were needed to make ninja check pass and 'format.sh fix' pass without changes made to the tree. There is a separate commit for these fixes, the new feature is in a commit after it.

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: That was an accidental leftover. Removed (in the next update).

opti-mix:
@pjaaskel I'd appreciate if you could submit your `32b dim_t fixes` commit as a separate PR if it is not too much work.

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: Good point. Might not be without some sort of POSIX wrapper (I haven't developed in Windows in ages). I better rewrite these using C++ streams or the LLVM wrappers.

pjaaskel:
Done!

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +176,96 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+    return nullptr;
+
+  uint64_t binarySize;
+  std::error_code errc = llvm::sys::fs::file_size(programPath, binarySize);
+  CHECK(!errc) << "Error getting the file size of " << programPath << ".";
+
+  std::ifstream binFile(programPath.c_str(), std::ios::binary);
+  CHECK(binFile) << "Error opening " << programPath << " for reading.";
+
+  auto binary = std::make_unique<unsigned char[]>(binarySize);
```
opti-mix: We use `glow::make_unique` in Glow.

opti-mix:
@ayermolo Would it be possible for you to test this PR on Windows to see if you can build Glow with OpenCL backend enabled? If you could run OpenCL tests it would be even better. We want to be sure this PR does not break Windows builds.

ayermolo:
@opti-mix Unfortunately I, or anyone here, don't have OCL environment setup on windows. We build and run with OCL backend turned off.
I could test other big commits, but not for OCL.

opti-mix:
@ayermolo 

> Unfortunately I, or anyone here, don't have OCL environment setup on windows. We build and run with OCL backend turned off. I could test other big commits, but not for OCL.

I see. Thanks anyways!

pjaaskel:
I changed the std::make_uniques to glow::make_unique. Unfortunately I also have no access to proper Windows development envs at hand to test it with, but I'm pretty confident it works since it now uses C++ streams for the file I/O instead of the POSIX C APIs.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +176,96 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+
+  auto binary = glow::make_unique<unsigned char[]>(binarySize);
+  binFile.read((char *)binary.get(), binarySize);
+  CHECK(binFile) << "Could not read the binary.";
+  binFile.close();
+
+  cl_int status;
+  cl_int err;
+  cl_program prog =
+      clCreateProgramWithBinary(ctx, 1, &device, &binarySize,
```
opti-mix: @pjaaskel OpenCL standard requires the fourth parameter (i.e. binarySize) go be `size_t`. This code does not compile on MacOS because of this. But `llvm::sys::fs::file_size(programPath, binarySize);` requires its last parameter to be `uint64_t`. Please fix it.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -166,6 +176,96 @@ cl_kernel OpenCLFunction::createKernel(const std::string &name,
...
+                                (const unsigned char **)&binary, &status, &err);
```
opti-mix: @pjaaskel Is it OK to cast a unique pointer `binary` like this `(const unsigned char **)&binary`?

pjaaskel:
Indeed, it was badly broken and went into the branch due to problems in the disk cache test which I relied on after changing the basic pointer to a unique_ptr. 

Here's a fix: https://github.com/pytorch/glow/pull/3993

Sorry!

PR #3965, opened 2020-01-06
Fix StringRef issue in ExecutionEngine

talc23:
ExecutionEngine::setBackendName() is getting a llvm::StringRef as input. in some cases, the reference is invalid after calling hostManager_.reset(), and when trying to use it in the next few lines you'll get an unexpected behavior (probably failing on line 56 check).
suggesting to use the member string that gets updated in the beginning of the function instead of using the input itself.

Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:


talc23:
@jsubag

jsubag:
@jfix71 Please take a look

PR #3968, opened 2020-01-06
Update NNPI Backend to v0.4.2.5

omromano:
Summary: Updating NNPI Backend to version 0.4.2.5.

### DIFF
--- a/lib/Backends/NNPI/Importer.cpp
+++ b/lib/Backends/NNPI/Importer.cpp
@@ -941,7 +941,7 @@ public:
     auto numInputs = glowConcat->getNumInputs();
     NNPIObjectName *inputs = new NNPIObjectName[numInputs];
     LOG_AND_RETURN_IF_NOT(ERROR, inputs, "No inputs", NNPI_INVALID_PARAM);
-    std::set<std::string> inputTensors;
+    std::unordered_set<std::string> inputTensors;
 
     for (unsigned i = 0; i < numInputs; i++) {
       auto nvName = nodeValueName(glowConcat->getNthInput(i));
@@ -969,11 +969,29 @@ public:
     importer.setUsedTensors({nodeValueName(glowTile->getInput())},
                             {nodeValueName(glowTile->getResult())});
 
-    return nnpiNetworkAddTileOp(importer.getNetwork(),
-                                glowTile->getName().begin(),
-                                nodeValueName(glowTile->getInput()).c_str(),
-                                nodeValueName(glowTile->getResult()).c_str(),
-                                glowTile->getCount(), glowTile->getAxis());
+    auto numDims = glowTile->getInput().getType()->dims().size();
+    std::vector<int32_t> repeats(numDims, 1);
+    auto axis = glowTile->getAxis();
+    LOG_AND_RETURN_IF_NOT(ERROR, axis >= 0 && axis < numDims,
+                          "tile axis is invalid", NNPI_INVALID_PARAM);
+    repeats[axis] = glowTile->getCount();
+    NNPITensorDesc desc;
+    desc.attributes.value = 0;
+    desc.attributes.constant = 1;
+    desc.numDims = 1;
+    desc.dims[0] = numDims;
+    desc.quantParams.precision = NNPI_PRECISION_INT32;
+    desc.quantParams.type = NNPI_QUANTIZATION_NONE;
+    desc.layout = NNPI_LAYOUT_ANY;
+
+    auto repeatsTensorName = glowTile->getName().str() + "_repeats";
+
+    importer.addTensor(repeatsTensorName, desc, repeats.data());
+
+    return nnpiNetworkAddTileOp(
+        importer.getNetwork(), glowTile->getName().begin(),
+        nodeValueName(glowTile->getInput()).c_str(), repeatsTensorName.c_str(),
+        nodeValueName(glowTile->getResult()).c_str());
   }
 };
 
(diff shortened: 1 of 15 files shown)


omromano:
Hi @jfix71  and @arunm-git , 
Please take a look.
CC @jsubag

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -73,7 +130,8 @@ InferenceThreadEnv::~InferenceThreadEnv() {
 void InferenceThreadEnv::execute(RunIdentifierTy runId,
```
jfix71: Not for now but for future releases -- the length of these functions in InferencePool are pretty high -- e.g. this one is 430 lines. I'd like for this to be refactored at some point -- we should try to carve out big chunks here to other functions, e.g. add a `prepareInputsForDevice()` and a `prepareInputsForICEREF()` so we can make the flow here clearer and less error prone.

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -159,267 +227,304 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
...
+          commandList_, UINT32_MAX, commandErrors_, numCommands_, &numErrors);
+      TRACE_EVENT_END(ctx->getTraceContext(), TraceLevel::OPERATOR,
+                      TRACING_INFERENCE);
+      TRACE_EVENT_BEGIN(ctx->getTraceContext(), TraceLevel::COPY,
+                        TRACING_POST_PROCESS);
+      if (res != NNPI_INF_NO_ERROR) {
+        LOG_NNPI_INF_ERROR(res, "Failed to wait on command list");
+      } else {
+        for (uint32_t i = 0; i < numErrors; i++) {
+          LOG(ERROR) << NNPI_INF_ERROR_MSG(commandErrors_[i].err,
```
jfix71: This looks weird to me -- so it's possible that `res == NNPI_INF_NO_ERROR && numErrors > 0`?

lib/Backends/NNPI/InferencePool.cpp
```diff
@@ -159,267 +227,304 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
...
-          LOG_AND_CALLBACK_EXECUTE_NNPI_INF_ERROR(
-              nnpiHostResourceLock(hostOutputs_[i].handle, NNPI_LOCK_FOR_READ,
-                                   UINT32_MAX, &pHostOutput),
-              "Failed to lock host resource", runId, ctx, resultCB);
-          LOG_AND_FAIL_EXECUTE_CALLBACK_IF_NOT(ERROR, pHostOutput, "Bad output",
-                                               runId, ctx, resultCB);
-        }
-
-        // Copy data to output tensor.
+      if (res != NNPI_INF_NO_ERROR || numErrors > 0) {
```
jfix71: Yeah same as above -- so it's possible that we have `NNPI_INF_NO_ERROR` and also have errors which seems weird. Perhaps we can add some comments here to explain a bit.

lib/Backends/NNPI/NNPIDeviceManager.cpp
```diff
@@ -356,19 +359,19 @@ void NNPIDeviceManager::transferStaticPlaceholderToDevice(
     uint32_t numErrors(0);
     NNPICommandListError singleError;
     memset(&singleError, 0, sizeof(NNPICommandListError));
-    NNPIInferenceErrorCode result =
+    NNPIInferenceErrorCode res =
         nnpiCommandListWait(cmdList, UINT32_MAX, &singleError, 1, &numErrors);
```
jfix71: Same notes as applied to `nnpiCommandListWait()` in the other place. I see this was not introduced in this PR but still would like better comments at least.

lib/Backends/NNPI/NNPIDeviceManager.cpp
```diff
@@ -130,20 +130,24 @@ Error NNPIDeviceManager::init() {
     if (deviceOptions_.enabledDeviceTracing) {
       deviceTracing_ = NNPIDeviceTracing::getForDevice(deviceId_);
     }
+    NNPIDeviceInfo deviceInfo;
+    LOG_NNPI_INF_ERROR_RETURN_LLVMERROR(
+        nnpiDeviceGetInfo(deviceId_, &deviceInfo),
+        "Failed to get NNPI Device Info");
+    maxMemoryBytes_ =
+        static_cast<uint64_t>(deviceInfo.totalUnprotectedMemory) * KB;
```
jfix71: Is `totalUnprotectedMemory` the size of DRAM? If so, perhaps we should set `maxMemoryBytes_` to what's actually usable?

lib/Backends/NNPI/InferencePool.cpp
jsubag: That's already part of an upcoming refactor - great minds... :)

lib/Backends/NNPI/InferencePool.cpp
jsubag: yes the call to query for error succeeds by fetching other errors

lib/Backends/NNPI/NNPIDeviceManager.cpp
jsubag: That's the usable portion of the DRAM for deployed networks without going into special directives for protected memory

lib/Backends/NNPI/Importer.cpp
```diff
@@ -969,11 +969,29 @@ class TileNodeImporter : public INNPINodeImporter {
...
-                                nodeValueName(glowTile->getInput()).c_str(),
-                                nodeValueName(glowTile->getResult()).c_str(),
-                                glowTile->getCount(), glowTile->getAxis());
+    auto numDims = glowTile->getInput().getType()->dims().size();
+    std::vector<int32_t> repeats(numDims, 1);
+    auto axis = glowTile->getAxis();
+    LOG_AND_RETURN_IF_NOT(ERROR, axis >= 0 && axis < numDims,
+                          "tile axis is invalid", NNPI_INVALID_PARAM);
+    repeats[axis] = glowTile->getCount();
+    NNPITensorDesc desc;
```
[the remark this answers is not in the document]
omromano: It gets optimised in the compiler but the way to express a constant is with a constant tensor.

PR #3971, opened 2020-01-06
Add flag for enabling extra offset in EmbeddingBag ops

jackm321:
Summary:
Add a flag to enable Glow's `EmbeddingBag` and `EmbeddingBagByteRowwiseOffsets` nodes to have the same semantics as is proposed for PyTorch's EmbeddingBag operator here https://github.com/pytorch/pytorch/issues/29019.

Documentation:
doxygen

Test Plan:
new unit tests


jackm321:
@yinghai let's check this in now and we can connect it up with PyTorch when that's committed or modify it at that point if necessary. I'd prefer not to leave this uncommitted for a long while.

PR #3975, opened 2020-01-07
[GraphOptz] Add numeric equivalence test for cseRespectsPredicates

UncleGene:
Summary: Added numeric equivalence check for the test that assures that common subexpressions with different predicates are not combined.

Testi Plan:
 ```
$ninja all && tests/GraphOptzTest --gtest_filter=GraphOptz.cseRespectsPredicates
$ninja test
```

### DIFF
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -392,9 +392,11 @@ TEST_F(GraphOptz, optimizeBatchNormAfterConvWithPred) {
 /// Check CSE will not merge two nodes that have all the same inputs but
 /// different predicates.
 TEST_F(GraphOptz, cseRespectsPredicates) {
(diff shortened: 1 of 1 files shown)


SplitInfinity:
Looks good! Thanks for working on this. As a sanity check, could you dump out and include the dot graphs for the unoptimized and optimized graphs in your test plan? Just to make sure we are testing two different graphs here.

UncleGene:
The graphs are identical, so not sure how dot files can help. checkNumericalEquivalence compares F_ and optimizedF_, and optimizeFunction produces the clone as a separate graph (otherwise numerical equivalence check would not make sense for any of the tests)

SplitInfinity:
That makes sense. This test case is testing that an optimization did *not* run, so it makes sense that at the graphs would be the same.

PR #3976, opened 2020-01-07
Enable blacklist for stress tests

talc23:
Summary:
Enables stress tests(ParameterSweep and SparseLengthsSumTest) to run with blacklist

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/lib/Backends/NNPI/tests/NNPIParameterSweepTest.cpp
+++ b/lib/Backends/NNPI/tests/NNPIParameterSweepTest.cpp
@@ -18,3 +18,114 @@
 using namespace glow;
 
 std::set<std::string> glow::backendTestBlacklist = {};
+
+struct BlacklistInitializer {
+  BlacklistInitializer() {
+    backendTestBlacklist = {
+        "BatchMatMulTest_Int8/39", "BatchMatMulTest_Int8/43",
+        "BatchMatMulTest_Int8/46", "BatchMatMulTest_Int8/51",
+        "BatchMatMulTest_Int8/54", "BatchMatMulTest_Int8/59",
+        "BatchMatMulTest_Int8/63", "BatchMatMulTest_Int8/67",
+        "BatchMatMulTest_Int8/71", "BatchMatMulTest_Int8/75",
+        "BatchMatMulTest_Int8/83", "BatchMatMulTest_Int8/87",
+        "BatchMatMulTest_Int8/90", "BatchMatMulTest_Int8/91",
+        "BatchMatMulTest_Int8/95", "FCTest_Int8/139",
+    };
+
+    for (int i = 0; i < 80; i++) {
+      backendTestBlacklist.insert("RWQSLWS_Float16_AccumFloat16/" +
+                                  std::to_string(i));
+    }
+    for (int i = 0; i < 80; i++) {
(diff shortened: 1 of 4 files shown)


tests/unittests/BackendTestUtils.h
```diff
@@ -97,20 +97,29 @@ class BackendTest : public BackendStatelessTest {
...
-#endif // GLOW_WITH_CPU
-#ifdef GLOW_WITH_OPENCL
-    "OpenCL",
-#endif // GLOW_WITH_OPENCL
-#ifdef GLOW_WITH_HABANA
-    "Habana",
-#endif // GLOW_WITH_HABANA
-    "Interpreter");
+
+#ifdef GLOW_TEST_BACKEND
```
talc23: this was done so that the created per-backend test binary won't have Interpreter tests in it (doubles the number of the tests for each backend)

talc23:
@jsubag

jsubag:
@jfix71 Take a look please.

jsubag:
@jfix, @arunm-git , Any update for this PR?

jfix71:
@jsubag I'll get it landed by the end of this week, sorry for the delay

jsubag:
Thank you!

lib/Backends/NNPI/tests/NNPIParameterSweepTest.cpp
```diff
@@ -18,3 +18,91 @@
 using namespace glow;
 
 std::set<std::string> glow::backendTestBlacklist = {};
+
+struct SoftwareOnlyTests {
```
jfix71: The design here is a bit strange. It looks like we're setting up a blacklist for ICEREF here, but then later on in `EmulatorOnlyTests` we're overriding that with different tests?

Is ICEREF's blacklist not a subset of HW's blacklist? If yes, why isn't it a subset?

If no, the design here seems kind of weird/misleading, would prefer to check `USE_INF_API` in here to make things more clear. And I'd also like to cleanup the names of the structs because they're confusing/misleading.

lib/Backends/NNPI/tests/NNPIParameterSweepTest.cpp
```diff
@@ -18,3 +18,91 @@
...
+  SoftwareOnlyTests() {
+    // If USE_NNPI_THRESHOLD is set, we should have empty blacklist
+    auto useNnpiThreshold = getenv("USE_NNPI_THRESHOLD");
```
jfix71: What threshold is supposed to be used if `USE_NNPI_THRESHOLD`? Is part of the PR missing?

talc23: it's part of our internal testing, removing it

lib/Backends/NNPI/tests/NNPIParameterSweepTest.cpp
```diff
@@ -18,3 +18,114 @@
 using namespace glow;
 
 std::set<std::string> glow::backendTestBlacklist = {};
+
+struct BlacklistInitializer {
```
talc23: BlacklistInitializer is is a dummy struct just so initialzation would happen.
as I want to add tests through a loop(makes the file more readable), I can't initialize backendTestBlacklist outside a function, thus I do it inside the BlacklistInitializer c'tor.

PR #3977, opened 2020-01-07
Changed the default tracing level in image-classifier to NONE

gcatron:
Summary:
As in title, changed the trace-level default to NONE, tracing can be opted into.

Documentation:

Test Plan:
builds locally.


gcatron:
cc @jsubag

PR #3982, opened 2020-01-07
[Base] Support padding in Tensor::isEqual()

### LINKED ISSUE #3898: Tensor::isEqual() not aware of padding
Current implementation of Tensor::isEqual() does not support tensors with padding. It just performs a plain traverse https://github.com/pytorch/glow/blob/master/include/glow/Base/Tensor.h#L817 without skipping padding positions. Also, it uses size() instead of actualSize() so it misses last elements when padding exists. An update is required to iterate over all valid elements while skipping pad-elements.

shajrawi:
Summary:

It previously implemented its own iteration on the raw data, which did not take into account padding information.

This PR update the implementation to use our Handle Iterators to inspect the actual value of the elements, which, in turn, supports padding.

Fixes #3898 

Test Plan:
`ninja test`

### DIFF
--- a/tests/unittests/TensorsTest.cpp
+++ b/tests/unittests/TensorsTest.cpp
@@ -1238,3 +1238,40 @@ TEST(CustomAlignedTensor, getDimForPtr) {
   EXPECT_EQ(H.getDimForPtr(1, 3), 1);
   EXPECT_EQ(H.getDimForPtr(2, 3), 0);
 }
+
+// Check that we iterate over tensors correctly: unit test for a bug wherein we
+// used size() instead of actualSize() when treating the data as a raw pointer.
+TEST(Tensor, sameAlignment) {
+  Type Ty1(ElemKind::Float16Ty, {2, 1}, {4, 1});
+  Type Ty2(ElemKind::Float16Ty, {2, 1}, {4, 1});
+  Tensor T1(Ty1);
+  Tensor T2(Ty2);
+  auto T1H = T1.getHandle<float16_t>();
+  auto T2H = T2.getHandle<float16_t>();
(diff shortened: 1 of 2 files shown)


tests/unittests/TensorsTest.cpp
```diff
@@ -1238,3 +1238,31 @@ TEST(CustomAlignedTensor, getDimForPtr) {
...
+  Tensor T1(Ty1);
+  Tensor T2(Ty2);
+  auto T1H = T1.getHandle<float16_t>();
+  auto T2H = T2.getHandle<float16_t>();
+  T1H.clear(0);
+  T2H.clear(1);
+  T1H.at({0, 0}) = T2H.at({0, 0}) = 1;
+  T1H.at({1, 0}) = T2H.at({1, 0}) = 2;
+
+  EXPECT_TRUE(T1.isEqual(T2));
```
nickgg: Can you add a case where `isEqual()` is false? Maybe just by changing on of the handles here and rechecking. (and in the differentAlignment test below).

shajrawi: Done. added it to both tests.

tests/unittests/TensorsTest.cpp
```diff
@@ -1238,3 +1238,31 @@ TEST(CustomAlignedTensor, getDimForPtr) {
...
+}
+
+TEST(Tensor, differentAlignment) {
```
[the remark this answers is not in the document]
shajrawi: I added comments for the test-cases: It actually should stay with the same name after adding the CHECK_FALSE Nick suggested: I modified the name of the other test-case to bring it inline with this one.

PR #3986, opened 2020-01-08
Updated tensor.assign to only copy unpadded size in the case of an unpadded tensor. (Quick Fix for Repro)

gcatron:
Summary:
This fixes an issue with the Repro tool where only the first Tensor has it's unpaddedSize set.
This is a quick fix and won't be landed, a more proper fix to Tensor will follow.
Documentation:

Test Plan:
ninja check


gcatron:
cc @jsubag

jfix71:
Closing in favor of #4026

PR #3991, opened 2020-01-08
Simple example backend that performs individual SLS ops

bertmaher:
Summary: We don't have code for an example backend, and there's enough stuff involved (Backend, CompilerFunction, maybe DeviceManager, lots of optional virtual methods, ...) that it's pretty intimidating to get started for someone who wants to use Glow but isn't an expert.  I think it might be worth having an executable example like this that shows someone how they might start building a backend.

This example might be kind of bad in a lot of ways, like, it falls over nastily if you give it any graph that isn't SLS->Save.  And I'm not sure what the best practices are around binding inputs are anymore.  And it doesn't create a DeviceManager.  And it doesn't have any comments.  And it kind of sucks to create a giant blacklist just to run a single OperatorTest.  But I'm hoping we can discuss whether this is worthwhile and what would need to be added to make it useful.

Documentation: code is executable documentation!


opti-mix:
@bertmaher I like the idea a lot! It would be very useful to have an example backend. And your PR is a good starting point!

That being said, I think ideally we would like to have a minimal example that covers most/all major topics described in the docs (e.g. in the `Backends.md`). It should probably include the following:
- Defining backend-specific `isOpSupported` and `shouldLower`
- Defining a backend-specific node(s)
- Defining a backend-specific instruction(s)
- Describing how to IRGen the new backend-specific nodes into new instructions (automatically or manually)
- Defining a backend-specific `transformPostLowering` with some simple graph transformations
- Defining a backend-specific graph verifier
- Defining backend-specific tensor layouts and their verification
- Defining a backend-specific optimization pipeline 
- (Maybe) Defining a backend-specific DeviceManager
- etc

It looks like a lot of things, but we need just skeletons of those, without any interesting functionality. And we have most of the real code examples scattered among different existing backends.

bertmaher:
@opti-mix I think that's a great list of topics.  Where I'm a bit conflicted is that adding all of those facets to a single example significantly increases the complexity.  Like, maybe a backend doesn't need IR, or doesn't need custom nodes, etc.  It might be cool to have some staged examples (thinking like LLVM's Kaleidoscope tutorial), where it's like:
- Chapter 1: here's the bare minimum backend you could possibly want, 
- Chapter 2: here's how you lower to IR
- Chapter 3: here's how you add custom nodes
- Chapter 4: custom optimizations,
- etc...

But maybe failing that it would be better to have a single, more complete example than this one?  I guess I'm trying to hit the sweet spot of complexity vs utility vs time to create.

bertmaher:
Something that does make me a little uncomfortable about this example as-is, is it's basically just a system for offloading (well, pretending to offload) individual sparse-lengths-sum operators, which isn't really in the spirit of a "graph compiler" :-p.

PR #3993, opened 2020-01-10
[OpenCL] fixes to the disk caching support

pjaaskel:
* Use size_t for the binary size parameter.
* Fix incorrect reference to the unique_ptr object.
* Fix the flaky test.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -203,6 +203,8 @@ cl_program OpenCLFunction::loadProgramFromDiskCache(std::string cacheDirectory,
   std::error_code errc = llvm::sys::fs::file_size(programPath, binarySize);
   CHECK(!errc) << "Error getting the file size of " << programPath << ".";
 
+  size_t binarySizeST = binarySize;
+
   std::ifstream binFile(programPath.c_str(), std::ios::binary);
   CHECK(binFile) << "Error opening " << programPath << " for reading.";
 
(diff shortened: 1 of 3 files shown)


lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -199,7 +199,7 @@ cl_program OpenCLFunction::loadProgramFromDiskCache(std::string cacheDirectory,
   if (!llvm::sys::fs::exists(programPath))
     return nullptr;
 
-  uint64_t binarySize;
+  size_t binarySize;
```
opti-mix: `size_t` on MacOS is not the same as uint64_t and thus your fix breaks `llvm::sys::fs::file_size(programPath, binarySize);`, because it expects binarySize to be `uint64_t`.

opti-mix: @pjaaskel So, you may need to have two different variables as you need two different types at different places.

pjaaskel: Of course :/ I pushed a fix. Does it work finally for you? (I do not have a MacOS to test with).

opti-mix: Yes, it works now!

opti-mix:
@pjaaskel The `size_t` issue is fixed now. Thanks!

But could you explain how to run your new test? It seems like it is not built on MacOS. Do I need to use any special CMake command-line options or something like this? Or maybe I need to run a special command to run the test?

pjaaskel:
It should get added to the test suite if OpenCL is enabled, so simply ninja check should suffice. I do not know what could be Apple-special in this one.

opti-mix:
@pjaaskel Works now. Don't know why it didn't work before.

PR #3994, opened 2020-01-10
[GraphOptz] Stop collapsing Dequantize(Quantize)

UncleGene:
Summary: Implementation optimizes Dequantize(Quantize(X)) -> X. This drops the rescaling aspect of Quantize.
Combination should be pretty rare and does not deserve introduction of new node type,
unless non-quantized rescaling can be useful by itself

Testing: New test fails without change, all tests succeed with it.


UncleGene:
Closing and submitting an issue, blast radius is larger than expected

PR #3998, opened 2020-01-11
Fix #3929 by force defining __STDC_FORMAT_MACROS

### LINKED ISSUE #3929: Compilation problem due to type macro usage in DimType.h
Hi,

I'm having a build problem with the following commit

https://github.com/pytorch/glow/commit/d2794f48c2eeeab3fb279a55f29a43257059ef4a

Uses some type-qualifying macros like `PRId64` to define its own custom macos:

```
#define PRIdDIM PRId64
#define PRIuDIM PRIu64
```

These macros (`PRId64` and such) are defined in https://en.cppreference.com/w/cpp/header/cinttypes only if __STDC_FORMAT_MACROS is defined. I'm getting the following compilation problem:

```
glow/lib/LLVMIRCodeGen/BundleSaver.cpp: In member function â€˜void glow::BundleSaver::saveHeader(llvm::StringRef)â€™:
glow/include/glow/Base/DimType.h:41:17: error: expected â€˜)â€™ before â€˜PRIu64â€™
 #define PRIuDIM PRIu64
                 ^
glow/lib/LLVMIRCodeGen/BundleSaver.cpp:293:35: note: in expansion of macro â€˜PRIuDIMâ€™
         shapeStr += strFormat("%" PRIuDIM ", ", dims[idx]);
```

The reason of this is that at the point where `<cinttypes>` was *first* included to this file, __STDC_FORMAT_MACROS was not defined. As result, PRIu64 is not defined at this point.

Note than there is a precedent where __STDC_FORMAT_MACROS was defined explicitly:
```
./tests/googlebenchmark/src/benchmark_register.cc:37:#define __STDC_FORMAT_MACROS
```

Note: defining __STDC_FORMAT_MACROS  in DimType.h does not work, because there are multiple possible ways how `<cinttypes>` may come to the destination file. If `<cinttypes>` was first included before DimTypes.h was included, the macros like `PRIu64` will not be defined.

I can work around it by defining __STDC_FORMAT_MACROS in *above all includes* to BundleSaver.cpp, but it does not sound like the correct solution because we'll need to do it in each file that uses these macros.

So it seems that this commit messes with macros in a way it shouldn't. I'm not sure what is the correct solution for that.

pjaaskel:
This definition is due to an older buggy C++11 cstdint implementation
which does not import the printf format macros (used at least by DimType.h)
without it in place. Since the definition doesn't harm the correct
implementations, we'll define it explicitly here without a preprocessor
test.

See: https://github.com/pytorch/glow/issues/3929 and
     https://sourceware.org/bugzilla/show_bug.cgi?id=15366

Hopefully Fixes #3929 

Test Plan: 

Tested only on Ubuntu 18.04 LTS where this problem was not happening the first place.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,3 +1,13 @@
+
+# This definition is due to older buggy C++11 cstdint implementations which do
+# not import the printf format macros (used at least by DimType.h) without it
+# in place. Since the definition doesn't harm the correct implementations,
+# we'll define it explicitly here without a preprocessor test.
+#
+# See: https://github.com/pytorch/glow/issues/3929 and
+#      https://sourceware.org/bugzilla/show_bug.cgi?id=15366
+add_definitions("-D__STDC_FORMAT_MACROS")
+
 add_custom_target(AutoGen)
 
 add_subdirectory(Backend)


lib/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
+
+# This definition is due to older buggy C++11 cstdint implementations which do
```
Max-K-cadence: I would not say it was buggy. It works exactly as it was described. That it might look unreasonable is another question. :)

PR #4000, opened 2020-01-13
add TouchInst for buffers that don't need to be initialized

### LINKED ISSUE #3903: Unneeded Splat when lowering ConcatNode
A Splat instruction is being added to initialize dest Tensor when lowering Splat node (https://github.com/pytorch/glow/blob/master/lib/IR/IRGen.cpp#L272). It seems unneeded because dest Tensor gets eventually written by subsequent InsertTensor instructions. Can we remove this Splat?

nickgg:
Summary: Per #3903 when IRGening ConcatNode we insert a `SplatInst` so the output buffer is initialized. In that case since we overwrite the entire buffer we don't need that splat and it's wasted work. This PR solves the problem by doing nothing, or rather by adding a new instruction `TouchInst` which does nothing except mark its output as initialized. I've only updated the case of Concat, but there may be other situations we use `splat` as an initializer for memory that is overwritten.

Documentation: N/A

Test Plan: Ran tests in debug and release. This code is covered by the Concat operator tests for backends that use IR - Interpreter, CPU and OpenCL (Habana and NNPI use Nodes instead).

Fixes #3903.

### DIFF
--- a/lib/IR/IRGen.cpp
+++ b/lib/IR/IRGen.cpp
@@ -269,7 +269,9 @@ void IRGenVisitor::post(Node *parent, Node *N) {
 
     auto *dest = builder_.createAllocActivationInst(CC->getName(),
                                                     CC->getResult().getType());
-    builder_.createSplatInst(CC->getName(), dest, 0);
+    // Mark the buffer as initialized, this is safe since the InsertTensors
+    // below will fully overwrite the buffer.
+    builder_.createTouchInst(CC->getName(), dest);
     auto inputs = CC->getInputs();
 
     // We start inserting to the shape at (0,0, ... ).
--- a/tools/ClassGen/InstrGen.cpp
+++ b/tools/ClassGen/InstrGen.cpp
@@ -621,6 +621,11 @@ int main(int argc, char **argv) {
       .autoVerify(VerifyKind::NoVerify)
       .autoIRGen();
 
+  BB.newInstr("Touch")
+      .addOperand("Dest", OperandKind::Out)
+      .dataParallel()
+      .autoVerify(VerifyKind::NoVerify);
+
   BB.newInstr("InsertTensor")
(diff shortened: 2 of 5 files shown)


tools/ClassGen/InstrGen.cpp
```diff
@@ -621,6 +621,11 @@ int main(int argc, char **argv) {
       .autoVerify(VerifyKind::NoVerify)
       .autoIRGen();
 
+  BB.newInstr("Touch")
+      .addOperand("Dest", OperandKind::Out)
+      .dataParallel()
```
jfix71: We can skip this, right? Seems a little misleading/weird to mark essentially a nop as data parallel.

nickgg:
This instruction must be implemented on all backends that use Glow IR (although it does nothing). CC: @mciprian13, @et-nivard.

lib/IR/IRGen.cpp
```diff
@@ -269,7 +269,8 @@ void IRGenVisitor::post(Node *parent, Node *N) {
 
     auto *dest = builder_.createAllocActivationInst(CC->getName(),
                                                     CC->getResult().getType());
-    builder_.createSplatInst(CC->getName(), dest, 0);
+    // Mark the buffer as initialized.
```
jfix71: nit: Perhaps add a little more info, e.g. "This is safe because the insert insts below which implement the concat must cover the entire buffer" or something?

tools/ClassGen/InstrGen.cpp
nickgg: We cannot skip this. I'm not 100% sure why but Instructions that only have an output (e.g. splat) must be data parallel or they don't work.

lib/IR/IRGen.cpp
nickgg: sure, I'll add some more colour.

PR #4001, opened 2020-01-13
Fix for sleep on windows.

ayermolo:
Summary:
Windows doesn't have sleep, but does have Sleep which takes in millisconds vs seconds.
Documentation:

[Optional Fixes #issue]

Test Plan: UnitTests

### DIFF
--- a/tests/unittests/SharedMutexTest.cpp
+++ b/tests/unittests/SharedMutexTest.cpp
@@ -54,7 +54,7 @@ TEST(SharedMutex, Performance) {
   }
 
   /* sleep override */
-  sleep(1);
+  std::this_thread::sleep_for(std::chrono::seconds(1));
 
   stopThread = true;
 


ayermolo:
@bertmaher I was just following original implementation with primary goal to fix windows. The sleep_for should work also I think.
@putivsky Any objections, since this is your original code.

ayermolo:
On side note pytorch failure seems unrelated. Is it possible to re-trigger the CI manually?

Unable to import torch. Error:
	 /tmp/venv/lib/python3.6/site-packages/torch/lib/libtorch_python.so: undefined symbol: _ZN5torch3jit14runJITCPPTestsEb

putivsky:
> @bertmaher I was just following original implementation with primary goal to fix windows. The sleep_for should work also I think.
> @putivsky Any objections, since this is your original code.

std::this_thread::sleep_for is the way to go.

ayermolo:
@bertmaher seems like pipeline failure is not related.

bertmaher:
Thanks!

PR #4003, opened 2020-01-14
Fix NNPI lowering for 4Bit SLS

jsubag:
Summary: Lowering SLS nodes to SLWS when using 4Bit data for the NNPI  backend


jsubag:
@arunm-git , @jfix71  Please take a look.

PR #4013, opened 2020-01-14
Allow backends to manually set the default parallel clone count, and set to 12 for NNPI

jfix71:
Summary:
By default let's test all 12 ICE cores if the test supports it by setting parallelCloneCount. Unit tests run pretty quickly so this doesn't cost much in runtime overhead and gets us better coverage on the NNPI HW and stack.

cl opt `-parallel-clone-count` will still work/override whatever the backend sets as its own default.

This included a fix to avoid std::moving a Tensor, which later led to a nullptr deref.

Differential Revision: D19401176


jfix71:
CC: @jsubag

jfix71:
Going to do this separately/in a different way.

PR #4015, opened 2020-01-15
Implement the Module cloning functionality

opti-mix:
Follow the examples of the existing cloning APIs for nodes and functions and extend this functionality to work with modules as well.

While doing that, also tighten the graph verifier checks when it comes to types used inside a module.

### DIFF
--- a/lib/Graph/Graph.cpp
+++ b/lib/Graph/Graph.cpp
@@ -4037,12 +4037,23 @@ ConstList Function::findConstants() const {
 }
 
 Function *Function::clone(llvm::StringRef newName,
-                          llvm::DenseMap<Node *, Node *> *map) {
+                          llvm::DenseMap<const Node *, Node *> *map,
+                          llvm::DenseMap<const Node *, Node *> *currToNewMap) {
   Module *M = getParent();
   auto *newF = M->createFunction(newName);
+  return clone(newF, map, currToNewMap);
+}
 
+Function *
+Function::clone(Function *newF, llvm::DenseMap<const Node *, Node *> *map,
+                llvm::DenseMap<const Node *, Node *> *currToNewMap) const {
   // Maps current nodes to new nodes.
-  llvm::DenseMap<Node *, Node *> currToNew;
+  llvm::DenseMap<const Node *, Node *> currToNew;
+
+  // Initialize the map from a user-provided map.
+  if (currToNewMap) {
+    currToNew.insert(currToNewMap->begin(), currToNewMap->end());
+  }
 
   // Clone all of the nodes in the function.
   for (auto &N : getNodes()) {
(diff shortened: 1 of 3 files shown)


tests/unittests/GraphTest.cpp
```diff
@@ -674,6 +674,89 @@ TEST(Graph, functionCloneTest) {
...
+    auto *F = originalM.createFunction("main");
+    Node *input1 = originalM.createPlaceholder(ElemKind::FloatTy,
+                                               {4, 10, 10, 3}, "input", true);
+
+    Node *add = F->createAdd("add", input1, input1);
+    Node *relu = F->createRELU("Relu", add);
+    auto *concat = F->createConcat("concat", {relu, relu, relu}, 0);
+    auto *C = originalM.createConstant(concat->getResult().getType(), "C");
+    C->getPayloadMutable().getHandle().clear(1.0f);
+    Node *SM = F->createAdd("sub", concat, C);
```
jackm321: nit "add"

lib/Graph/Graph.cpp
```diff
@@ -4093,6 +4123,43 @@ static bool verifyNodeInput(const Node &N, size_t idx) {
...
+  for (auto &C : getConstants()) {
+    auto *copyC = M->createConstant(C->getType(), C->getName(), C->getLayout());
+    copyC->assign(&C->getPayload());
+    currToNew[C] = copyC;
+  }
+  // Clone all functions.
+  for (auto &F : getFunctions()) {
+    // Create an empty clone function in the new module.
+    auto copyF = M->createFunction(F->getName());
+    // Clone function's body into the newy created cloned function. Use the
```
shajrawi: nit: newly

lib/Graph/Graph.cpp
```diff
@@ -4093,6 +4123,43 @@ static bool verifyNodeInput(const Node &N, size_t idx) {
...
+  // Clone constants.
+  for (auto &C : getConstants()) {
+    auto *copyC = M->createConstant(C->getType(), C->getName(), C->getLayout());
+    copyC->assign(&C->getPayload());
+    currToNew[C] = copyC;
+  }
+  // Clone all functions.
+  for (auto &F : getFunctions()) {
+    // Create an empty clone function in the new module.
+    auto copyF = M->createFunction(F->getName());
```
jfix71: nit: `auto *copyF`

lib/Graph/Graph.cpp
```diff
@@ -4093,6 +4123,43 @@ static bool verifyNodeInput(const Node &N, size_t idx) {
...
+    currToNew[PH] = copyPH;
+  }
+  // Clone constants.
+  for (auto &C : getConstants()) {
+    auto *copyC = M->createConstant(C->getType(), C->getName(), C->getLayout());
+    copyC->assign(&C->getPayload());
+    currToNew[C] = copyC;
+  }
+  // Clone all functions.
+  for (auto &F : getFunctions()) {
```
jfix71: nit: `auto *F` instead? Right now it's a reference to a pointer.

opti-mix:
@jackm321 

> What is the expectation for modules that have been stripped? I think this will reallocate memory for the Constants but that might not be true. In any case, it would be good to document and test what the expected behavior is.

This is a great catch! I added a DCHECK for now to catch such cases where we have constants with unowned payloads and bail on it. These situations cannot be resolved by the cloner on its own without some input from the client.

tests/unittests/GraphTest.cpp
```diff
@@ -674,6 +674,89 @@ TEST(Graph, functionCloneTest) {
...
+  // module was deleted.
+  EXPECT_TRUE(clonedM.verify());
+  // Compile and run the cloned model.
+  compileAndRun(clonedEE, clonedBindings, clonedM, "input", resultName);
+  // Store the result of running the cloned module.
+  clonedResult.assign(
+      clonedBindings.get(clonedBindings.getPlaceholderByName(resultName)));
+  // The results of execution should be exactly the same in both cases.
+  EXPECT_TRUE(originalResult.isEqual(clonedResult, 0));
+  clonedEE.setBackendName("Interpreter");
```
jfix71: Why is this necessary?

lib/Graph/Graph.cpp
```diff
@@ -4093,6 +4123,49 @@ static bool verifyNodeInput(const Node &N, size_t idx) {
...
+  for (auto &PH : getPlaceholders()) {
+    auto *copyPH = M->createPlaceholder(PH->getType(), PH->getName(),
+                                        PH->isTraining(), PH->getLayout());
+    currToNew[PH] = copyPH;
+  }
+  // Clone constants.
+  for (auto &C : getConstants()) {
+    // Cloner cannot decide on its own what to do with constants having unowned
+    // payloads. Some kind of policy/hook maybe needed in the future for
+    // decidiing what needs to be done in such cases.
```
jfix71: typo: decidiing

tests/unittests/GraphTest.cpp
opti-mix: Oops. Good catch! This should not be there. It was from an experiment ;-) Removed!

PR #4022, opened 2020-01-16
Don't attempt constant folding of Backend specific Nodes

nickgg:
Summary: Backend Specific nodes tend not to be implemented in the Interpreter, and thus can't be constant folded. Prevent attempts to optimize out these nodes because they will assert.

Documentation: N/A

Test Plan: ran tests in debug and release

### DIFF
--- a/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
+++ b/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
@@ -289,10 +289,18 @@ bool glow::ConstantFold::run(Function *F, const CompilationContext &cctx) {
     if (isa<Storage>(N) || isa<Constant>(N) || isa<SplatNode>(N)) {
       continue;
     }
+
+    // Don't try to constant fold Nodes that are not supported by the
+    // Interpreter. These are usually backend-specific.
+    if (!backend->isOpSupported(*N)) {
+      continue;
+    }
+
     // Skip nodes that are not constant operations.
     if (!isConstantOperation(N, *backend)) {
       continue;
     }
+
     // Add only a constant operation node whose value is used by at least
     // one non constant-operation node, because no other bigger constant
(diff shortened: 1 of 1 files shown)


lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -302,6 +302,12 @@ bool glow::ConstantFold::run(Function *F, const CompilationContext &cctx) {
     if (!hasNonConstantOperationUser(N, *backend)) {
       continue;
     }
+    // Don't try to constant fold Nodes that are not supported by the
+    // Interpreter. These are usually Backend specific.
```
SplitInfinity:
```suggestion
    // Interpreter. These are usually backend-specific.
```

SplitInfinity:
Seems legit but I will defer to @jfix71 on this one because I don't know the details of this pass that well.

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
nickgg: I dunno, they're specific to a `Backend` object aren't they?

lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -302,6 +302,12 @@ bool glow::ConstantFold::run(Function *F, const CompilationContext &cctx) {
     if (!hasNonConstantOperationUser(N, *backend)) {
       continue;
     }
+    // Don't try to constant fold Nodes that are not supported by the
+    // Interpreter. These are usually backend-specific.
+    if (!backend->isOpSupported(*N)) {
```
jfix71: nit: Could probably move this up in the loop, as the above two calls are a bit more complex.

Issue #3827, opened 2019-11-28
[Reshape] Allow the shape to be computed from constants


mciprian13:
Assume we have the following case:
![image](https://user-images.githubusercontent.com/29785908/69807139-3c9c0280-11ed-11ea-860f-01c77195e2dd.png)

Assume also that even though the graph has this form, all the tensor sizes are statically known (that is the size for the output of conv and transpose are known). Right now this model will not work because the Reshape node complains that the "shape" operator is not explicitly a constant even though conceptually the shape is a constant.

Can we do something to support this case?
Maybe something like moving the restriction of "shape" operand being a constant right after all the graph optimizations took place.

ayermolo:
Have you tried -const-fold-ops? It tries to constant fold nodes during loading.

mciprian13:
Without the flag I get one error:
`Error message: Non-constant shape tensors are unsupported by Glow.`

With the -const-fold-ops flag I get multiple errors:
```
Error message: No constant found
Error message: No constant found
Error message: No constant found
Error message: Non-constant shape tensors are unsupported by Glow.
```

ayermolo:
Maybe I miss understood the issue. Are you saying output of Conv is constant, as in can be deduced at compile time of the graph?

mciprian13:
The tensor size (shape) is constant not the tensor values. The reshape operator has 2 operands: a tensor and a shape. The problem is about the shape operand.

ayermolo:
Sorry derp moment. Are you sure problem is not elsewhere in the graph?
Yeah when shape is loaded it will create a constant, and -constant-fold-ops should fold gather, concat, and by the time reshape is loaded that input should be constant.

I just double checked on onnx ssd which has similar pattern and it worked. Although granted it's on internal branch...

mciprian13:
@jfix71, @ayermolo I understand the flag `-const-fold-ops` might do some hocus-pocus with nodes which involve constants. My questions:
1. Why isn`t this flag enabled by default?
2. Why isn`t this flag solving the problem? ayermolo mentions that this does the trick on an internal branch ...

jfix71:
@mciprian13 

>Why isn`t this flag enabled by default?

Not sure 🙂I don't think there's any real downside to enabling it by default.

> Why isn`t this flag solving the problem? ayermolo mentions that this does the trick on an internal branch ...

[...]

PR #4027, opened 2020-01-17
[Importer] Enable constant folding by default and extra fix.

mciprian13:
**Summary**
- Fix for #3827.
- Enable the constant folding by default.

**Documentation**
None

**Test Plan**
None

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -227,7 +227,7 @@ Error constantFoldInLoader(Function *F, LoaderType &tmpLoader,
   // Register the constant inputs to the current op with the constant folding
   // loader.
   for (unsigned i = 0; i < (dim_t)op.input_size(); i++) {
-    Constant *tmpConst = mod->getConstantByName(op.input(i));
+    Constant *tmpConst = loader->getConstantByNameOrNull(op.input(i));
     RETURN_ERR_IF_NOT(tmpConst, "No constant found");
     tmpLoader.nodeValueByName_[op.input(i)] = tmpConst->getOutput();
   }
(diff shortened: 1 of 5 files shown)


mciprian13:
@jfix71 
It seems the text translator model fails if the constant folding is enabled with the attached error:
[build_54728_step_106_container_0.txt](https://github.com/pytorch/glow/files/4076404/build_54728_step_106_container_0.txt)
Namely:
`Unsupported node found while compiling Function eval_const_fold__ for backend Interpreter: ScatterData
`
If my understanding is correct, the constant folding fails with error if the node cannot be folded (not supported by Interpreter).
In my opinion, the constant folding should be interpreted as a best-effort optimization and therefore the model compilation should **NOT fail if a node cannot be folded** (at most a notification should be printed in the console) but the compilation should continue without folding that node.
What do you think?

jfix71:
>In my opinion, the constant folding should be interpreted as a best-effort optimization and therefore the model compilation should NOT fail if a node cannot be folded (at most a notification should be printed in the console) but the compilation should continue without folding that node.

@mciprian13 Yeah I agree. I'm a bit confused as to how this has occurred, as I believe the Interpreter supports all the ops in this model. Regardless, please file an issue for this, it shouldn't be so hard to create a fix.

mciprian13:
@jfix71 I created an issue for the point discussed above.

jfix71:
I'm merging https://github.com/pytorch/glow/pull/4050 and https://github.com/pytorch/glow/pull/4061 into this PR when it lands in a bit, as it needed some fixes based on internal CI.

Verma-Rajat:
quantization flows are broken due to this commit.
https://github.com/pytorch/glow/issues/4188

PR #4028, opened 2020-01-17
[Loader] Remove assert which prevents users to load ONNX network from current directory w/o `./` prefix

aturetsk:
Summary:
Currently, due to assert in Loader::getModelOptPath, one cannot use `-m=model.onnx`, but have to do `-m=./model.onnx` instead. The patch removes the assert and makes the function return "." as the directory.

Documentation: N/A

Test Plan: manually tested


aturetsk:
@jfix71, it looks like "Facebook Internal - Builds & Tests" failed due to some warning I'm not able to get to. Also, DEBUG and PYTORCH CIs failed due to some python issues. How should I proceed with the patch?

jfix71:
@aturetsk Looks like it's transient stuff, I'll get this landed today.

PR #4029, opened 2020-01-17
[NFC] Fix a bunch of compilation warnings (GCC).

aturetsk:
Summary:
[NFC] Fix a bunch of compilation warnings (GCC)

Documentation: N/A

Test Plan: N/A


aturetsk:
@jfix71, it does. I believe `llvm_unreachable()` gets expanded into an expression which hints compiler that this place in the code is never reached, thus compiler does not generate a warning.

jfix71:
@aturetsk Hm ok, if it's all the same to you I'd prefer `llvm_unreachable` then, allows the compiler to optimize slightly better 🙂

aturetsk:
@jfix71, updated with `llvm_unreachable`

PR #4046, opened 2020-01-23
Updated cleanupProvision to evict partitions for an unsuccessful provisioning.

gcatron:
Summary:
Currently if the provisioner fails while loading a network part way through any partitions loaded are left on device. This PR adds the cleanup logic to properly evict the partial load. Also added a unit test to exercise the code flow.

Documentation:

Test Plan:
Added new unit test.

### DIFF
--- a/include/glow/Runtime/Provisioner/Provisioner.h
+++ b/include/glow/Runtime/Provisioner/Provisioner.h
@@ -44,6 +44,9 @@ public:
   /// Remove stored compiledFunction.
   Error removeFunction(llvm::StringRef name);
 
+  /// Evict function from device.
+  Error evictFunction(llvm::StringRef name, DeviceIDTy device);
+
   /// \returns a reference to the backend with name \p backendName.
   Backend &getBackend(llvm::StringRef backendName) const;
 
@@ -66,9 +69,13 @@ private:
   /// List of available DeviceManagers added during initialization.
   std::vector<DeviceManager *> devices_;
 
-  /// Helper function to cleanup a provision call. On failure free the
-  /// compiledFunctions that were created.
-  void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
+  /// Helper function to cleanup a provision call. On \p failure free the
+  /// compiledFunctions that were created, \p names , and remove networks
+  /// already added to devices, \p currentNetworkResidency .
+  void cleanupProvision(llvm::ArrayRef<std::string> names,
+                        std::map<DeviceIDTy, std::vector<std::string>> const
+                            &currentNetworkResidency,
+                        bool failure = true);
 
   /// Helper function to parse the DAG and generate logicalDevices.
   std::map<DeviceIDTy, std::vector<DAGNode *>>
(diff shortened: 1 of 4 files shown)


lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -530,4 +538,24 @@ void Provisioner::cleanupProvision(llvm::ArrayRef<std::string> names,
       functions_.erase(name);
     }
   }
+  if (failure) {
+    // Remove any partitions added to devices.
+    for (auto &device : deviceCleanup) {
+      for (auto network : device.second) {
```
jfix71: nit: `auto &network`

lib/Runtime/Provisioner/Provisioner.cpp
```diff
@@ -530,4 +538,24 @@ void Provisioner::cleanupProvision(llvm::ArrayRef<std::string> names,
...
+        std::promise<void> evictPromise;
+        auto done = evictPromise.get_future();
+        std::unique_ptr<Error> evictErr;
+        devices_[device.first]->evictNetwork(
+            network, [&evictErr, &evictPromise](std::string, Error err) {
+              evictErr = glow::make_unique<Error>(std::move(err));
+              evictPromise.set_value();
+            });
+        done.wait();
```
jfix71: This seems somewhat boilerplate -- and I think we have pretty similar code in HostManager's `removeNetwork()`. Worth colocating this logic in a helper or something?

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -67,8 +67,12 @@ class Provisioner final {
   std::vector<DeviceManager *> devices_;
 
   /// Helper function to cleanup a provision call. On failure free the
-  /// compiledFunctions that were created.
-  void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
+  /// compiledFunctions that were created and remove and networks already added
+  /// to devices.
+  void
+  cleanupProvision(llvm::ArrayRef<std::string> names,
+                   std::map<DeviceIDTy, std::vector<std::string>> deviceCleanup,
```
jfix71: const ref for the map?

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -67,8 +67,12 @@ class Provisioner final {
   std::vector<DeviceManager *> devices_;
 
   /// Helper function to cleanup a provision call. On failure free the
-  /// compiledFunctions that were created.
-  void cleanupProvision(llvm::ArrayRef<std::string> names, bool failure = true);
+  /// compiledFunctions that were created and remove and networks already added
```
jfix71: nit "and remove and networks"

include/glow/Runtime/Provisioner/Provisioner.h
```diff
@@ -67,8 +67,12 @@ class Provisioner final {
...
+  /// to devices.
```
jfix71: nit: Reference the parameters in doxygen? E.g. you we remove all networks already added to devices, but this is actually based on whatever is found in `\p deviceCleanup`.

include/glow/Runtime/Provisioner/Provisioner.h
jfix71: Also, tiny nit, I'd suggest different naming for `deviceCleanup` -- device cleanup is more what it's used for, right? It's more like the `currentNetworkResidency` or something.

Issue #3785, opened 2019-11-15
Make backends more self-contained in their sub-directories


opti-mix:
With recent re-factoring, new backends can be mostly self-contained inside their respective `backends/BackendName` sub-directories.

But there are a couple of places outside of those directories, which still need to be updated when a new backend is added:
- `DeviceManager::numDevices` contains a switch statement based on backend names
- `ONNXModelWriter.cpp` needs to be aware of backend-specific nodes
- `tests/unittests/BackendTestUtils.h` contains explicit names of all backends

Overall, these are just a couple of places, but they are preventing backends from being completely self-contained.

It would be nice to find a way to resolve those last remaining hard-coded dependencies and make backends completely self-contained.

PR #4048, opened 2020-01-23
Change DeviceManager::numDevices() to make backends more self-contained

soumyadeepghosh:
Summary:

- Introduce a virtual method `numDevices()` in every Backend Factory
  class that must be implemented by every new backend. This method
  must return the number of devices of the specified kind.
- Move the implementation of each individual kind of `numDevices()` from
  the `DeviceManager` class to each individual backend class.
- Modify the `numDevices()` method in `DeviceManager` to return values
  based on all the backends already registered.

Documentation:

[Fixes part 1 of #3785 ]

Test Plan: ninja test

### DIFF
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -409,6 +409,10 @@ bool CPUBackend::shouldLower(const Node *N) const {
   }
 }
 
+unsigned CPUBackend::numDevices() {
+  return std::thread::hardware_concurrency();
+}
+
 std::unique_ptr<CompiledFunction> CPUBackend::createCompiledFunction(
     std::unique_ptr<llvm::orc::GlowJIT> JIT,
     runtime::RuntimeBundle &&runtimeBundle) const {
--- a/lib/Backends/NNPI/NNPI.cpp
+++ b/lib/Backends/NNPI/NNPI.cpp
@@ -21,6 +21,8 @@
 #include "glow/Optimizer/Lower/Lower.h"
 #include "llvm/Support/CommandLine.h"
 
+#include <fstream>
+
 using namespace glow;
 
 namespace glow {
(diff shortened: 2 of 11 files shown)


lib/Backends/NNPI/NNPI.h
```diff
@@ -34,6 +34,27 @@ class NNPIBackend final : public Backend {
 
   std::string getBackendName() const override { return getName(); }
   static std::string getName() { return "NNPI"; }
+  static unsigned numDevices() {
+#if defined(GLOW_WITH_NNPI)
+    // TODO: unify with numHabanaDevices. copy-pasta with a different device
```
SameerAsal: typo: pasta :) 🍝

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -409,6 +409,14 @@ bool CPUBackend::shouldLower(const Node *N) const {
   }
 }
 
+unsigned CPUBackend::numDevices() {
+#if defined(GLOW_WITH_CPU)
+  return std::thread::hardware_concurrency();
+#else
+  return 0;
```
jfix71: Don't think this is necessary anymore. If we don't build with `GLOW_WITH_CPU` then the CPU would never be registered with the factory anyway. Same on the other backends too.

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -68,6 +68,24 @@ int32_t GlowNNPINumParallelChunks = 1;
 
 NNPIBackendOptions NNPIBackend::backendOptions_;
 
+static unsigned NNPIBackend::numDevices() {
+  // TODO: unify with numHabanaDevices. copy-paste with a different device
```
opti-mix: We usually end the comment with a period.

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -68,6 +68,24 @@ int32_t GlowNNPINumParallelChunks = 1;
...
+  while (std::getline(devices, device)) {
+    if (device.find("sph_pcie") != std::string::npos) {
+      count++;
+    }
+  }
+  if (count > 0) {
+    return count;
+  }
+  // Todo
+  return 1; // Fall back to emulator since GLOW_NNPI is set. This feels hacky.
```
opti-mix: Glow usually puts comment before the line, not on the same line.

Issue #4030, opened 2020-01-20
[Importer] Constant folding raises error if node not supported in Interpreter


mciprian13:
In #4027 the constant folding has been enabled by default.
During the import procedure, if a node is not supported by the Interpreter, an error is raised.
The constant folding should be a best-effort optimization and therefore an error should NOT be raised if a node cannot be folded. The import procedure should continue without folding that node.
Such an example is the TextTranslator application which raises an error when using constant folding:
`Unsupported node found while compiling Function eval_const_fold__ for backend Interpreter: ScatterData`.

opti-mix:
Yes, this sounds very reasonable.

tlepley-cadence:
There are 2 constant folding optims:

1. one done at the very beginning of the flow when the model is loaded
2. one performed in the optimize() function later in the flow.

@mciprian13 I understand that you speak about a problem with 1.

On my side, I have in the past encountered this same kind of issue in case 2, when calling optimize() after backend post-lowering. In case backend specific nodes were introduced, optimize() was failling at constant folding time because it was not detecting that the Interpreter wasn't supporting  these nodes. Our local fix has been to disable constant folding when running optimize() after calling backend specific methods.

This problem may not exist anymore today (I've not checked yet), but it looks very similar to the one discussed here.

jfix71:
Fixed as part of https://github.com/pytorch/glow/pull/4050 (merged in https://github.com/pytorch/glow/commit/855775629b8be67116012770bc044e90b23646e7)

PR #4050, opened 2020-01-23
Fix ConstantFolding to not fail compilation

jfix71:
Summary:
Fixes https://github.com/pytorch/glow/issues/4030. Comes with additional fixes on top of https://github.com/pytorch/glow/pull/4027.

Includes a new verboseCompile bool in the CompilationContext and set it to false for constant folding (true by default) so if there are constant folding errors we don't spam the log.

CC: @mciprian13 @tlepley-cadence

Differential Revision: D19543559


jfix71:
Merged into https://github.com/pytorch/glow/pull/4027.

PR #4051, opened 2020-01-24
Add folly dependency

bertmaher:


CMakeLists.txt
```diff
@@ -186,6 +186,27 @@ else()
...
+    set(FOLLY_HAVE_WEAK_SYMBOLS ON CACHE BOOL "Compiler supports weak symbols")
+    set(OPENSSL_ROOT_DIR /usr/local/opt/openssl)
+    find_package(OpenSSL REQUIRED)
+  endif()
+
+  # Remove Glow-specific CMAKE_CXX_FLAGS to build folly.
+  set(SAVE_CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+  set(CMAKE_CXX_FLAGS "")
+  add_subdirectory("${GLOW_THIRDPARTY_DIR}/folly" EXCLUDE_FROM_ALL)
+  target_include_directories(folly PUBLIC "${OPENSSL_INCLUDE_DIR}")
```
jackm321: OPENSSL_INCLUDE_DIR?

jackm321: Oh nvm, needed by folly I guess

bertmaher: Yeah I dunno why folly’s internal build system doesn’t figure this out properly but there you have it.

bertmaher:
@ayermolo Here's my second take at importing folly.  This PR only requires boost, double-conversion, and jemalloc as external dependencies.  I've taken a *very* aggressive subset of folly.  Possibly *too* aggressive, since people who want to use folly will need to wade through a few pages of missing symbol errors until they link all the sources they want.  But, it ensures we don't pull in too much stuff.

ayermolo:
@bertmaher Really appreciate it. I am trying to wrap up something internally, will try to pull this branch in by EOW to try to build on windows.

bertmaher:
Unfortunately this approach is not working on Linux.  I'm going to revert to the original approach.

ayermolo:
That is unfortunate. What was the original approach bring in all the dependencies?

bertmaher:
Yeah, bringing in the deps.  Tbh, the only one I was able to avoid was openssl with the other approach.

ayermolo:
I see. Thank you for trying.

Issue #3845, opened 2019-12-04
[Debug] Option to dump all graph intermediate computations during inference as binaries


mciprian13:
It would be nice to have this feature while using (for example) image-classifier.
For debugging purposes of course it would be nice for the tool to dump all the intermediate tensors during inference in separate named binaries.

opti-mix:
@mciprian13 Could it be something like the current `-instrument-debug` option, which dumps all the intermediate tensors in the textual format?

mciprian13:
OK I didn`t know about that.
I tried it, but I think it would be more useful with a couple of features:
- print every tensor in a separate file (and print the entire tensor because currently the large ones are not printed entirely)
- flexibility in the format (binary or text)

opti-mix:
> I tried it, but I think it would be more useful with a couple of features:

PRs are welcome! :-)

> print every tensor in a separate file (and print the entire tensor because currently the large ones are not printed entirely)

Should be trivial to add.

> flexibility in the format (binary or text)

Should not be too hard too.

mciprian13:
Solved in #4392. Closing this issue.

PR #4053, opened 2020-01-24
Create network debugger tool

SameerAsal:
This PR introduces a tool called networkDebugger, this tool allows us
to compare runs of the network, layer by layer, on a reference backend
and a test backend.

The tool itself is an example of how to use the underlying functionality
implemented in Network Comparator and it could be used in any other
fashion.

Documentation:
 Added an md file to describe the work.
 Partially addresses #3845

Test Plan:
  ran the tool to detect errors on an internal network with a new backend.

### DIFF
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -223,9 +223,7 @@ void glow::setOnnxDefineSymbol(const std::vector<std::string> &strs) {
   onnxDefineSymbol = strs;
 }
 
-/// Parse as input file name \p fileName which is an ONNX file
-/// and \returns a parsed GraphProto.
-::ONNX_NAMESPACE::GraphProto glow::parseOnnxFile(const std::string &fileName) {
+ONNX_NAMESPACE::GraphProto glow::parseOnnxFile(const std::string &fileName) {
   ::ONNX_NAMESPACE::GraphProto graphProto;
   std::ifstream inputFileStream(fileName, std::ios::in | std::ios::binary);
   CHECK(inputFileStream) << "Can't find the input file for " << fileName;
@@ -237,11 +235,9 @@ void glow::setOnnxDefineSymbol(const std::vector<std::string> &strs) {
   return graphProto;
 }
 
-/// Taken an ONNX file name in \p fileName and loads the tensors
-/// in \p bindings.
-void glow::fillPlaceholders(const std::string &fileName,
-                            PlaceholderBindings *bindings) {
-  auto inputGroup = parseOnnxFile(fileName);
+void glow::fillPlaceholders(const ONNX_NAMESPACE::GraphProto &inputGroup,
+                            PlaceholderBindings *bindings,
+                            std::vector<Tensor> *partialTensorPayloads) {
   for (const auto &tensorProto : inputGroup.initializer()) {
     auto *tensor =
         bindings->get(bindings->getPlaceholderByName(tensorProto.name()));
@@ -253,19 +249,40 @@ void glow::fillPlaceholders(const std::string &fileName,
     CHECK(!hasError) << "Cannot load input tensor";
     size_t loadedSize = tensor->getSizeInBytes();
     if (loadedSize != fullSize) {
-      // pad with 0
-      LOG(INFO) << "Loading and padding " << tensorProto.name()
-                << " as a partial tensor: partial size="
-                << tensor->getType().toString()
-                << " full size=" << fullType.toString();
-      Tensor fullTensor(&fullType);
-      std::memcpy(fullTensor.getUnsafePtr(), tensor->getUnsafePtr(),
-                  tensor->getSizeInBytes());
-      std::memset(fullTensor.getUnsafePtr() + tensor->getSizeInBytes(), 0,
-                  fullTensor.getSizeInBytes() - tensor->getSizeInBytes());
-      *tensor = std::move(fullTensor);
+      if (partialTensorPayloads) {
+        LOG(INFO) << "Loading " << tensorProto.name()
+                  << " as a partial tensor: partial size="
+                  << tensor->getType().toString()
+                  << " full size=" << fullType.toString();
+        Tensor fullTensor(tensor->getUnsafePtr(), &fullType,
+                          tensor->getSizeInBytes());
+        // 'fullTensor' doesn't own the underlying data. 'tensor' does. So
+        // we want to keep the original tensor object around until inference
+        // is finished.
+        partialTensorPayloads->emplace_back(std::move(*tensor));
+        *tensor = std::move(fullTensor);
+      } else {
+        // pad with 0
+        LOG(INFO) << "Loading and padding " << tensorProto.name()
+                  << " as a partial tensor: partial size="
+                  << tensor->getType().toString()
+                  << " full size=" << fullType.toString();
+        Tensor fullTensor(&fullType);
+        std::memcpy(fullTensor.getUnsafePtr(), tensor->getUnsafePtr(),
+                    tensor->getSizeInBytes());
+        std::memset(fullTensor.getUnsafePtr() + tensor->getSizeInBytes(), 0,
+                    fullTensor.getSizeInBytes() - tensor->getSizeInBytes());
+        *tensor = std::move(fullTensor);
+      }
     }
   }
+}
+
+void glow::fillPlaceholders(const std::string &fileName,
+                            PlaceholderBindings *bindings,
+                            std::vector<Tensor> *partialTensorPayloads) {
+  const ONNX_NAMESPACE::GraphProto &inputGroup = parseOnnxFile(fileName);
+  fillPlaceholders(inputGroup, bindings, partialTensorPayloads);
 }
 
 /// Loads tensor \p T from the input \p in.
@@ -2711,8 +2728,8 @@ Error ONNXModelLoader::loadNetwork(ONNX_NAMESPACE::GraphProto &net) {
       if (!tryFold) {
         // Error during constant folding; load the op normally below.
         const std::string errStr = ERR_TO_STRING(tryFold.takeError());
-        VLOG(1) << "Error while trying to ConstantFold " << loadOperatorName(op)
-                << ": " << errStr;
+        LOG(INFO) << "Error while trying to ConstantFold "
+                  << loadOperatorName(op) << ": " << errStr;
       } else if (tryFold.get()) {
         // Folded successfully, so skip loading the op below.
         continue;
--- a/lib/Importer/ONNXModelLoader.cpp
+++ b/lib/Importer/ONNXModelLoader.cpp
@@ -223,6 +223,51 @@ void glow::setOnnxDefineSymbol(const std::vector<std::string> &strs) {
   onnxDefineSymbol = strs;
 }
 
+/// Parse as input file name \p fileName which is an ONNX file
+/// and \returns a parsed GraphProto.
+::ONNX_NAMESPACE::GraphProto glow::parseOnnxFile(const std::string &fileName) {
+  ::ONNX_NAMESPACE::GraphProto graphProto;
+  std::ifstream inputFileStream(fileName, std::ios::in | std::ios::binary);
+  CHECK(inputFileStream) << "Can't find the input file for " << fileName;
+  google::protobuf::io::IstreamInputStream protobufFileStream(&inputFileStream);
+  google::protobuf::io::CodedInputStream codedStream(&protobufFileStream);
+  codedStream.SetTotalBytesLimit(MAX_PROTO_SIZE, MAX_PROTO_SIZE);
+  bool parsedSuccessfully = graphProto.ParseFromCodedStream(&codedStream);
+  CHECK(parsedSuccessfully) << "Failed to parse GraphProto";
+  return graphProto;
+}
+
+/// Taken an ONNX file name in \p fileName and loads the tensors
+/// in \p bindings.
+void glow::fillPlaceholders(const std::string &fileName,
+                            PlaceholderBindings *bindings) {
+  auto inputGroup = parseOnnxFile(fileName);
+  for (const auto &tensorProto : inputGroup.initializer()) {
+    auto *tensor =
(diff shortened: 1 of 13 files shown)


lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+    while (itRef != endRef && itCheck != endCheck) {
+      if (!itRef->second->isEqual(*(itCheck->second), numericCmpThreshold_,
+                                  true)) {
+        LOG(INFO) << "Error: " << itRef->first << "\n";
+        return false;
+      }
+      itRef++;
+      itCheck++;
+    }
+  } else {
```
jfix71: nit: Could check this first and return false early. Allows for the above logic in the current if case to be one nested level down.

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+    std::unordered_map<std::string, Tensor *> &refOuts,
+    std::unordered_map<std::string, Tensor *> &checkOuts) {
+  if (refOuts.size() == checkOuts.size()) {
+    auto itRef = refOuts.begin();
+    auto itCheck = checkOuts.begin();
+    auto endRef = refOuts.end();
+    auto endCheck = checkOuts.end();
+    while (itRef != endRef && itCheck != endCheck) {
+      if (!itRef->second->isEqual(*(itCheck->second), numericCmpThreshold_,
+                                  true)) {
```
jfix71: nit: `/* verbose */ true`

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+NetworkComparatorBase::NetworkComparatorBase(Module &mod,
+                                             std::string referenceBackend,
+                                             std::string testBackend,
+                                             float numericCmpThreshold,
+                                             bool dumpTensorsForBadLayer)
+    : numericCmpThreshold_(numericCmpThreshold),
+      dumpTensorsForBadLayer_(dumpTensorsForBadLayer) {
+  DCHECK(mod.getFunctions().size() == 1)
+      << "Module must have exactly one functions";
+  inputModule_ = mod.clone();
```
jfix71: Do we need to clone here? Not sure it's necessary.

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+}
+
+NetworkComparatorBase::InOutTensors RecursiveLayerComparator::hookAndRun(
+    llvm::StringRef layerName, PlaceholderBindings *bindings, bool isRef) {
+  ExecutionEngine execEngine;
+  execEngine.setBackendName(isRef ? EERefNet_.getBackendName()
+                                  : EETestNet_.getBackendName());
+  inputModule_->clone(&execEngine.getModule());
+  HookedFunction hook =
+      hookNode(execEngine.getSingleFunctionFromModule(), layerName, true);
```
jfix71: nit: `/* hookInputs */ true`

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+  PlaceholderBindings &inferBindings =
+      isRef ? inferBindingsRef : inferBindingsCheck;
+  inferBindings.allocate(execEngine.getModule().getPlaceholders());
+  // Copy the tensors from the bindings to the inference bindings
+  // we feed to the Temp network.
+  for (auto PH : bindings->pairs()) {
```
jfix71: nit: `auto &PH`

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+bool IntermediateLayerComparator::testSingleLayer(const Node *node) {
+  bool pass = true;
+  std::unordered_map<std::string, Tensor *> singleLayerInputMap;
+  ExecutionEngine singleLayerExecEng(EETestNet_.getBackendName());
+  PlaceholderBindings singleLayerBindings;
+  Module &singleLayerMod = singleLayerExecEng.getModule();
+  Function *singleLayerFunc = singleLayerMod.createFunction(node->getName());
+  Node *singleLayerNode = node->clone();
+  std::string layerName = node->getName();
```
jfix71: `llvm::StringRef`?

lib/Graph/NetworkComparator.cpp
```diff
@@ -0,0 +1,365 @@
...
+  std::string kindName = node->getKindName();
```
jfix71: `llvm::StringRef`?

lib/Graph/CMakeLists.txt
```diff
@@ -34,6 +35,7 @@ add_library(Graph
 
 target_link_libraries(Graph
                       PUBLIC
+                   #     ExecutionEngine
```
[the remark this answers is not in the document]
SameerAsal: this as just a temp trying to figure out a linking problem I had with SHARED build. Gone now.

tools/Debugger/network-debugger.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+void parseCommandLine(int argc, char **argv) {
+  llvm::sys::PrintStackTraceOnErrorSignal(argv[0]);
+  llvm::cl::ParseCommandLineOptions(
+      argc, argv,
+      " Network Debugger tool, part of the Glow compiler\n\n"
+      "Glow is a compiler for neural network accelerators.\n");
+}
+
+::ONNX_NAMESPACE::GraphProto parseIO(const std::string &filename) {
```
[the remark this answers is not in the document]
SameerAsal: What support lib are you referring to? I agree this should be shared with Repro

tools/Debugger/network-debugger.cpp
```diff
@@ -0,0 +1,242 @@
...
+    if (verbose)
+      llvm::outs() << "Clipping to fp16 enabled\n";
+  }
+  if (forceFP16AccumSLSOpt) {
+    precConfig.forceFP16AccumSLS = true;
+    if (verbose)
+      llvm::outs() << "Forcing fp16 accumulation for SLS ops enabled\n";
+  }
+  // Convert Network to fp16.
+  glow::convertFunctionToFloat16(F, precConfig);
```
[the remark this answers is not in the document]
SameerAsal: The idea here was to not compile the module, just optimize it to remove convertTo nodes that get add after we convert it to fp16.


> - are they going to be dumped from Glow? Are they going to have gone through Glow's optimization pipeline already?
If I understood your question correctly, no. they will just be loaded from the protobuf file. The optimizer is just to get rid of extra nodes. I didn't want the module ot be compiled because the  Comparator classes ar going to further modify it (by inserting hooks) then compile. You can't compile a module twice

tools/Debugger/network-debugger.cpp
```diff
@@ -0,0 +1,242 @@
...
+void loadModelIntoFunc(Function *F_, bool verbose = true) {
+  Error err = Error::empty();
+  { ONNXModelLoader onnxLD(modelPathOpt, {}, {}, *F_, &err, /*zipMode*/ true); }
+  CHECK(!ERR_TO_BOOL(std::move(err)))
+      << "ONNXModelLoader failed to load model: " << modelPathOpt;
+  convertNetwork(F_, false);
+}
+void fillInputPlaceholders(size_t inputIdx, PlaceholderBindings *bindings) {
+  auto inputGroup = parseIO(inputsOpt[inputIdx]);
+  for (const auto &tp : inputGroup.initializer()) {
```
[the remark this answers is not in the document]
SameerAsal: This function should be factored out as well. 

@yinghai  suggested that he prefers to keep the comparator logic not part of Repro.

lib/Graph/NetworkComparator.cpp
SameerAsal: you mean something like:

```
  if (refOuts.size() != checkOuts.size()) {
    LOG(ERROR) << "Backends produced different number of results\n";
    return false;
  }

  auto itRef = refOuts.begin();
  auto itCheck = checkOuts.begin();
  auto endRef = refOuts.end();
  auto endCheck = checkOuts.end();
  while (itRef != endRef && itCheck != endCheck) {
    if (!itRef->second->isEqual(*(itCheck->second), numericCmpThreshold_,
                                /* verbose */ true)) {
      LOG(INFO) << "Error: " << itRef->first << "\n";
      return false;
    }
    itRef++;
    itCheck++;
  }
```

So the rest of the code is not indented in, right?

lib/Graph/NetworkComparator.cpp
SameerAsal: Good catch, I only take a pointer now.

lib/Graph/NetworkComparator.cpp
SameerAsal: llvm::stringReg doesn't work with LOG() output streams .

lib/Graph/NetworkComparator.cpp
SameerAsal: StringRef only  has overrides with the llvm::outs output streams :(

tools/Debugger/network-debugger.cpp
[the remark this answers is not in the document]
SameerAsal: Should I move this to ONNXModelLoderr.h\cpp ?

tools/Debugger/NetworkComparator.cpp
```diff
@@ -0,0 +1,362 @@
...
+  for (auto const *nodePtr : order) {
+    auto &node = *nodePtr;
+    DCHECK(nodePtr) << "Node is empty!";
+    if (llvm::isa<SaveNode>(&node) || llvm::isa<Constant>(&node) ||
+        llvm::isa<Placeholder>(&node)) {
+      continue;
+    }
+    std::string layerName = node.getName();
+    std::string kName = node.getKindName();
+    LOG(INFO) << "Verifying layer: " << layerName << "\tType: " << kName
```
jfix71: nit: Not a huge deal, but if you use `llvm::StringRef` above you can use `layerName.data()` here, and we'd never need to copy the string.

tools/Debugger/NetworkComparator.cpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+NetworkComparatorBase::InOutTensors RecursiveLayerComparator::hookAndRun(
+    llvm::StringRef layerName, PlaceholderBindings *bindings, bool isRef) {
+  ExecutionEngine execEngine;
+  execEngine.setBackendName(isRef ? EERefNet_.getBackendName()
+                                  : EETestNet_.getBackendName());
+  inputModule_->clone(&execEngine.getModule());
+  HookedFunction hook =
+      hookNode(execEngine.getSingleFunctionFromModule(), layerName, true);
```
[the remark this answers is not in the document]
SameerAsal: Actually, I looked around and gound out that glow::loadTensor() is in OnnxLoader and it seemed more natural to move the APIs for parseIO and fill place holders o the same files. Thanks for the suggestion.

SameerAsal:
> So, if the motivation of this tool is to debug a model we have that's misbehaving, we probably want to use the actual optimization pipeline, right? That is, right now you're converting to FP16 explicitly before the compilation pipeline, whereas in the normal flow for a model we'd do that during the compilation pipeline.
That is correct. More specifically, the goal from this tool is to find a layer in a network that is producing wrong results (think a bad kernel). 

The reason I'd like to convert and optimize the module before I pass it to make sure the Comparator goes over the layers that actually gets executed. Since optimization can delete \ change some layers it made sense to me to convert and optimize the layer so the convertor can loop and find the broken layer. Since the comparator will basically "chunk up" the network and start testing we might end up with different eventual node types for a layer being execution . 

I agree with you that having it here seems out of place, given what I had in mind I am open to suggestions on how else you think we could do this!

tools/Debugger/NetworkComparator.cpp
```diff
@@ -0,0 +1,362 @@
...
+    DCHECK(nodePtr) << "Node is empty!";
+    if (llvm::isa<SaveNode>(&node) || llvm::isa<Constant>(&node) ||
+        llvm::isa<Placeholder>(&node)) {
+      continue;
+    }
+    llvm::StringRef layerName = node.getName();
+    llvm::StringRef kName = node.getKindName();
+    LOG(INFO) << "Verifying layer: " << layerName.data()
+              << "\tType: " << kName.data() << "\n";
+    auto refOuts = hookAndRun(layerName, bindings, true);
```
jfix71: nit: `/* isRef */ true`

tools/Debugger/NetworkComparator.cpp
```diff
@@ -0,0 +1,362 @@
...
+    auto checkOuts = hookAndRun(layerName, bindings, false);
```
jfix71: nit: `/* isRef */ false`

SameerAsal:
Using `transformForPrecisionMode()` does see cleaner. I filed an issue to myself to do that once #4112 is resolved.

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -223,6 +223,73 @@ void glow::setOnnxDefineSymbol(const std::vector<std::string> &strs) {
...
+        bindings->get(bindings->getPlaceholderByName(tensorProto.name()));
+    CHECK(tensor);
+    size_t fullSize = tensor->getSizeInBytes();
+    const auto fullType = tensor->getType();
+    auto error = loadTensor(tensorProto, tensor);
+    bool hasError = ERR_TO_BOOL(std::move(error));
+    CHECK(!hasError) << "Cannot load input tensor";
+    size_t loadedSize = tensor->getSizeInBytes();
+    if (loadedSize != fullSize) {
+      if (enablePartialTensor) {
```
jfix71: small nit, no big deal: seems like we could simply check `if (partialTensorPayloads)` instead here, and get rid of `enablePartialTensor`? In repro we'd just pass `enablePartialTensor ? &partialTensorPayloads : nullptr`.

SameerAsal:
* Tested that repro still works by running an internal test network.
* addressed comments.

jfix71:
FYI you're going to hit conflicts due to https://github.com/pytorch/glow/pull/4123

Issue #2165, opened 2018-12-12
Delay the 'Quantize(Constant)->Constant' optim until after post-lowering.


tlepley-cadence:
Currently, Constant are statically quantized early in the compilation flow. Basically the `Quantize(Constant)->Constant)` optim happens in the `glow::optimize()` function.

There are multiple issues with this, potentially resulting into accuracy loss because the scale/offset of `Constant` may still have to change during the Glow compilation flow. By keeping `Quantize` nodes longer in the IR, scale/offset can be changed freely without resulting into any accuracy loss for constants.

In particular, the backend may at post-lowering:
- introduce rescale nodes to handle specific constrains between node operands for the target, resulting to the need to requantize some constants after rescale propagation / optims.
- do extra target specific optimizations resulting in the need to requantize some constants.

Just to illustrate with an example, the convolution bias is currently quantized to have a scale equal to (scale of input) * (scale of weights). We noticed that after post-lowering this equality was not necessary true anymore, due to some lowering & optimization combinations. the bias then needs to be requantized, what can be done without loss only if the `Quantize` node is still in the IR.

beicy:
If we delay this "Quantize(Constant)->Constant" utile after post-lowering, some backend may be affected. For example, after double checked with @opti-mix , we found one of the backends requires the weights/bias of a conv node should be constant. But if we leave this "Quantize(Constant)->Constant" opt after that, the weights/bias would be "QuantizeNode" which would cause verification issue later.

To address the accuracy loss issue, @opti-mix proposes the following approach: 
```
void ExecutionEngine::optimizeFunction(CompilationMode mode, Function *F) {
 
  // Verify the function pre-optimization/lowering.                                                                                                                        
  assert(F->verify() && "Function must be valid");

  //***TODO*** add a function hooked to backend to allow the backend to do some 
  // special optimization based on the original graph. In this case, 
  //we can allow backends to adjust the scale/offset if they want.
  backend_->preOptimization(F, mode)  --- Added

  // Optimize the graph.                                                                                                                                                   
  ::glow::optimize(F, mode);

...
}
```
So, we don't have to separate  the quantize opt from our optimization function, and our current backends won't be affected.

jfix71:
The source of the issue is that when we quantize a float Constant, we quantize the underlying data immediately and also change the output type to the Quantize node we are mering in.

[...]

PR #4055, opened 2020-01-24
Add an opt-in backend option to delay the actual quantization of constant 

tlepley-cadence:
**Summary**: Add an opt-in backend option to delay the actual quantization of constant tensors after post-lowering. For us, letting the backend handle itself some aspects of the quantization is fundamental. For doing this in a lossless manner, we need to delay the actual quantization of constant (Quantize(Constant) -> Constant)

**Documentation**: N/A

**Fixes** #2165 #2481

**Test Plan**: Added a mock backend tests to check at what time the actual constant quantization is done, depending on the backend option. Changed existing related tests to reflect the code change.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -3137,21 +3137,6 @@ bool OptimizeQuantization::run(Function *F, const CompilationContext &cctx) {
         continue;
       }
 
-      if (auto *C = dyn_cast<Constant>(Q->getInput())) {
-        // Quantize(Constant) -> Constant
-        // Note, it does not really matter how many usages this Constant has.
-        // Quantized graph will use optimized Constant and other functions will
-        // refer to the floating point original Constant.
-        NodeValue NC =
-            convertConstant(*F->getParent(), *C, Q->getResult().getType());
-        if (NC == NodeValue()) {
-          continue;
-        }
-        changed = true;
-        Q->getResult().replaceAllUsesOfWith(NC);
-        continue;
-      }
-
       if (auto *SN = dyn_cast<SplatNode>(Q->getInput())) {
         // Quantize(Splat) -> Splat'
         changed = true;
@@ -3264,6 +3249,33 @@ bool OptimizeQuantization::run(Function *F, const CompilationContext &cctx) {
     changed = sinkRescaleQuantizedNode(F);
   }
   return changed;
+}
+
+void glow::convertQuantizedConstants(Function *F, CompilationContext &cctx) {
+  for (auto &node : F->getNodes()) {
+    auto *Q = dyn_cast<QuantizeNode>(&node);
+    if (!Q) {
+      continue;
+    }
+    auto *C = dyn_cast<Constant>(Q->getInput());
+    if (!C) {
+      continue;
+    }
+
+    // Quantize(Constant) -> Constant
+    // Note, it does not really matter how many usages this Constant has.
+    // Quantized graph will use optimized Constant and other functions will
+    // refer to the floating point original Constant.
+    NodeValue NC =
+        convertConstant(*F->getParent(), *C, Q->getResult().getType());
+    if (NC == NodeValue()) {
+      continue;
+    }
+    Q->getResult().replaceAllUsesOfWith(NC);
+  }
+
+  // Perform Dead Code Elimination.
+  runDCEPass(F, cctx);
 }
 
 void glow::convertPlaceholdersToConstants(Function *F,
@@ -3873,13 +3885,24 @@ Error glow::optimizeFunction(Function *F, const Backend &B,
     }
   }
 
-  // Allow the backend to transform the graph after lowering.
-  if (B.transformPostLowering(F, cctx, devInfo)) {
-    // If the backend made changes, optimize the graph again. Perform only
-    // passes that the Backend has requested so we do not interfere with the
-    // backend's transformations.
-    ::glow::optimize(F, cctx, B);
+  if (B.shouldPreQuantizeConstants()) {
+    // Do the actual float ->fix-point conversion of constant tensors before
+    // Post-lowering.
+    ::glow::convertQuantizedConstants(F, cctx);
   }
+
+  // Allow the backend to transform the graph after lowering.
+  B.transformPostLowering(F, cctx, devInfo);
+
+  if (!B.shouldPreQuantizeConstants()) {
+    // Do the actual float ->fix-point conversion of constant tensors after
+    // Post-lowering.
+    ::glow::convertQuantizedConstants(F, cctx);
+  }
+
+  // Optimize the graph again after the backend transformation.
+  // In particular, DCE is very likely to be useful.
+  ::glow::optimize(F, cctx, B);
 
   // We already started using backend specific verification when the function
   // state became lowered. Do one more verification pass to make sure everything
--- a/tests/unittests/GraphOptzTest.cpp
+++ b/tests/unittests/GraphOptzTest.cpp
@@ -1961,7 +1961,7 @@ TEST_F(GraphOptz, DCEPublicVars) {
   EXPECT_EQ(mod_.getPlaceholders().size(), 1);
 }
 
-TEST_F(GraphOptz, foldQuantizeIntoVar) {
+TEST_F(GraphOptz, foldQuantizeIntoConstant) {
   auto *input = mod_.createPlaceholder(ElemKind::FloatTy, {4}, "input", true);
   *bindings_.allocate(input) = {10, 10, 10, 10};
   auto qType = mod_.uniqueType(ElemKind::Int8QTy, {4}, 2, 0);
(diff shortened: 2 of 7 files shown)


tlepley-cadence:
@jfix71Could you please take a look to this PR ? I'd like to rapidly land it if possible.
This PR should not affect existing backends since it's an opt-in.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2922,6 +2907,29 @@ bool OptimizeQuantization::run(Function *F, const CompilationContext &cctx) {
   return changed;
 }
 
+void glow::convertQuantizedConstants(Function *F, CompilationContext &cctx) {
+  for (auto &node : F->getNodes()) {
+    if (auto *Q = dyn_cast<QuantizeNode>(&node)) {
```
jfix71: nit: could early continue to reduce amount of indentation, e.g.

```cpp
auto *Q = dyn_cast<QuantizeNode>(&node);
if (!Q) {
  continue;
}
```

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2922,6 +2907,29 @@ bool OptimizeQuantization::run(Function *F, const CompilationContext &cctx) {
...
+      if (auto *C = dyn_cast<Constant>(Q->getInput())) {
```
jfix71: Same as above, could early continue here.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2922,6 +2907,29 @@ bool OptimizeQuantization::run(Function *F, const CompilationContext &cctx) {
...
+          continue;
+        }
+        Q->getResult().replaceAllUsesOfWith(NC);
+        continue;
+      }
+    }
+  }
+
+  // Perform Dead Code Elimination.
+  runDCEPass(F, cctx);
```
jfix71: The optimization pipeline runs DCE before every pass (unless the pass opts out via `DCERequiredMode`, and none currently do). So not sure this is necessary.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
   }
 }
 
+static SaveNode *getUniqueSaveNode(Function *F) {
```
jfix71: nit: doxygen. Prefer to note here this call does some checking itself/may fail the test.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+  SaveNode *foundSaveNode = nullptr;
+  for (auto &node : F->getNodes()) {
+    if (auto *s = llvm::dyn_cast<SaveNode>(&node)) {
+      EXPECT_EQ(foundSaveNode, nullptr);
+      foundSaveNode = s;
+    }
+  }
+  EXPECT_NE(foundSaveNode, nullptr);
```
jfix71: nit: `ASSERT_NE` instead? The caller is expecting this to be non-null, so could cause a segfault in the test otherwise.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+    EXPECT_TRUE(quant);
+    EXPECT_TRUE(llvm::isa<Constant>(quant->getInput()));
+
+    return false;
+  }
+};
+
+/// Check the backend actual constant quantization is done before post-lowering.
+TEST_F(GraphOptz, foldQuantizeIntoConstantBeforePostLowering) {
+  // auto *input = mod_.createPlaceholder(ElemKind::FloatTy, {4}, "input",
```
jfix71: Delete commented out code.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+  // true); *bindings_.allocate(input) = {10, 10, 10, 10};
+
+  auto *input = mod_.createConstant(ElemKind::FloatTy, {4}, "input");
+  input->getHandle<float>() = {10, 10, 10, 10};
+  auto qType = mod_.uniqueType(ElemKind::Int8QTy, {4}, 2, 0);
+  auto *Q = F_->createQuantize("quantize", input, qType);
+  F_->createSave("save", Q);
+
+  CompilationContext cctx;
+  BackendOptions opts;
```
jfix71: nit: This is not used, remove it.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+/// Check the backend actual constant quantization is done after post-lowering.
+TEST_F(GraphOptz, foldQuantizeIntoConstantAfterPostLowering) {
+  auto *input = mod_.createConstant(ElemKind::FloatTy, {4}, "input");
+  input->getHandle<float>() = {10, 10, 10, 10};
+  auto qType = mod_.uniqueType(ElemKind::Int8QTy, {4}, 2, 0);
+  auto *Q = F_->createQuantize("quantize", input, qType);
+  auto *save = F_->createSave("save", Q);
+
+  CompilationContext cctx;
+  BackendOptions opts;
```
jfix71: nit: This is not used, remove it.

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+  input->getHandle<float>() = {10, 10, 10, 10};
+  auto qType = mod_.uniqueType(ElemKind::Int8QTy, {4}, 2, 0);
+  auto *Q = F_->createQuantize("quantize", input, qType);
+  F_->createSave("save", Q);
+
+  CompilationContext cctx;
+  BackendOptions opts;
+  auto B = MockBackendPrequantizeConst();
+  EXIT_ON_ERR(::glow::optimizeFunction(F_, B, cctx));
+}
```
jfix71: Can you do checking here that the constant was quantized as well (same logic as below)?

tests/unittests/GraphOptzTest.cpp
```diff
@@ -1817,6 +1825,86 @@ TEST_F(GraphOptz, foldQuantizeIntoVarMultipleUsages) {
...
+
+/// Check the backend actual constant quantization is done after post-lowering.
+TEST_F(GraphOptz, foldQuantizeIntoConstantAfterPostLowering) {
```
jfix71: Actually, this test should be identical to `foldQuantizeIntoConstantBeforePostLowering` except for the different backend, right? Can we combine these into a single helper function templatized on which mock backend to use?

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: `ASSERT` only work in functions returning void or in the main test functions. If I use it, it won't compile.
```
error: void value not ignored as it ought to be
 2034 |   ASSERT_NE(foundSaveNode, nullptr);
```

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: The check is check that Quantize is folded into Constant is before PostLowering is done in
`MockBackendPrequantizeConst::transformPostLowering` . Then, I don't think it's needed after calling `::glow::optimizeFunction`. 
I've added a comment in the code to mention this internal backend check.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: OK, it means that I then need to add the test you were requested in the previous feedback :)
There is then a good reason to do it now.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: OK, done to create a common template function.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: Done.

tests/unittests/GraphOptzTest.cpp
tlepley-cadence: Done.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
tlepley-cadence: Done

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
tlepley-cadence: Done

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
tlepley-cadence: Done

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
tlepley-cadence: `convertQuantizedConstants` is not an optimization pass, but just a function called by `glow::optimizeFunction`. Then, DCE needs to be called manually.

tlepley-cadence:
@jfix71 All done. Please go ahead.

Issue #4020, opened 2020-01-16
Refactor Image Classifier.


ayermolo:
Right now Image-classifier primary works with well image classification networks. There are other networks such as Segementation, Object Detection. The majority of the code to execute them is the same. Main differences are how some pre-processing stuff and outputs are handled.  For example for object detection not only they have multiple outputs, but also usually MAAP score is calculated by python scripts.

Would people be open to idea of refactoring common code in to a seperate file with some hooks for pre-processing and post processing. So that image-classifier, and maybe future apps like object-detection be a wrapper around it?

I am still thinking full implementation through, but wanted to open this to conversation to judge if there is an interest or none starter. Before I invest too much time in to it.

mciprian13:
I was thinking about this for some while now. I even put some ideas in a document I would like to share:
[Glow_Front_End_Tools.zip](https://github.com/pytorch/glow/files/4069860/Glow_Front_End_Tools.zip)

I would like to contribute to this. Maybe others would like as well.
Let`s first exchange some ideas: for example you can write your own ideas in the document and maybe send it back.

Cheers,!

ayermolo:
Thanks for sharing your doc. It looks like a more high level approach.  I was thinking more low level to provide building blocks for others to build on top of.
More specifically refactor processImageRange, and major functions it's using like loadImagesAndPreprocess, buildAndCompileAndGetInAndOutPair in to a class/seperate file. and provide mechanism to register hooks to 

-  run on images as pre-process step to be executed on Tensor containing input images.
- run after model is loaded. This can be used to append/prepend graph.
- run custom "processAndPrintResults". So for image classifiers it can be whatever image-classifier does now. For object detection handle multiple outputs to dump if network already has post-processing, or if one desires create it programatically, whatever.

The mechanism I was thinking is having Interface classes that someone can implement and then register with the executor.

I think this should provide flexible enough mechanism to build what you described in your doc.

Just my general idea. Was going to flash it out more next week and write up something.

mciprian13:
Understood. Your approach is inclined more towards enabling the developers. I was thinking more towards enabling end users directly with an out-of-the-box solution without the need for further development.
We still should see if there are common things we want to do and collaborate on that.

jfix71:
Seems reasonable to me, but not sure we have cycles to contribute on our end. Happy to review/help land PRs.

ayermolo:
[...]

PR #4056, opened 2020-01-24
Add a Loader extension mechanism

tlepley-cadence:
Summary: Add a loader extension mechanism to avoid 'overloading' the Loader class. Can be used for instance for doing IR instrumentation and more. 

Documentation: N/A

Part of #4020 

Test Plan: N/A. I've not seen any unit test that relates to image_classifier and Loader. We have used/tested this code internally.

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -234,6 +234,7 @@ static std::pair<Placeholder *, Placeholder *>
 buildAndCompileAndGetInAndOutPair(Loader &loader, PlaceholderBindings &bindings,
                                   TypeRef inputImageType) {
   auto LD = createProtobufLoader(loader, inputImageType);
+  loader.postModelLoad(bindings, *LD.get(), inputImageType->dims()[0]);
 
   // Allocate tensors to back all inputs and outputs.
   bindings.allocate(loader.getModule()->getPlaceholders());
@@ -542,6 +543,20 @@ setupContextPool(Placeholder *outputPH, Placeholder *inputImagePH,
   return contexts;
 }
 
+/// This function is a placeholder for registering Loader extensions to the
+/// loader. A loader extension in an instance of a loader extension class that
+/// must derive from class LoaderExtension. Any number of loader extensions
+/// can be registered to the Loader.
+static void registerLoaderExtensions(Loader &loader) {
+  // Example :
+  // class myLoaderExtension : public LoaderExtension {
+  //    // implement pure virtuel methods of class LoaderExtension.
+  // };
+  // loader.registerExtension(
+  //   std::unique_ptr<LoaderExtension>(new myLoaderExtension()));
+  (void)loader;
+}
+
 int main(int argc, char **argv) {
   // Verify/initialize command line parameters, and then loader initializes
   // the ExecutionEngine and Function.
@@ -613,6 +628,10 @@ int main(int argc, char **argv) {
           glow::make_unique<TraceContext>(TraceLevel::STANDARD));
     }
     Loader loader;
+
+    // Register Loader extensions.
+    registerLoaderExtensions(loader);
+
     // Used to make sure we only compile once, and run only once if not
     // streaming.
     bool isFirstRun = true;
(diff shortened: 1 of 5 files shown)


ayermolo:
@jfix71 can someone on your team or you, take a look at this?
After this is in I can refactor image-classifier as we discussed in https://github.com/pytorch/glow/issues/4020

tools/loader/Loader.h
```diff
@@ -45,6 +48,27 @@ bool profilingGraph();
...
 
+/// Loader extension interface from which to derive in order to extend the
+/// Loader driver object.
+class Loader;
+class ProtobufLoader;
+
+class LoaderExtension {
+public:
+  /// Called once after ONNX or Caffe2 model loading.
+  virtual void postModelLoad(Loader &, PlaceholderBindings &, ProtobufLoader &,
```
beicy: Is this an optional call or not after ONNX or Caffe2 model loading?

beicy:
@tlepley-cadence  Thanks for working on it! In general, it looks good to me. Could you please check this file https://github.com/pytorch/glow/blob/master/tests/unittests/Caffe2ImporterTest.cpp to see if you can add some test for the added function? Thanks!

tools/loader/Loader.h
tlepley-cadence: For all **registered** extensions (an extension is a class that derives from LoaderExtension), this method will be called. If for a particular extension, you don't plan any change after model load, then you just need to implement this method empty this extension. An extension may also decide to do something or not based on some command line options.

tlepley-cadence:
@beicy The loader extension mechanics in this MR is related to the Loader class, not to actual model loader classes (caffe2/onnx) that derive from ProtobufLoader. Then, it doesn't seem to me directly related to `Caffe2ImporterTest.cpp`

This execution of the Loader extension 'entry points' is driven by the application (like image_classifier). I've not seen any unit test for image_classifier and the Loader class and this is why I had not added any unit test in this MR. I can try to create a unit test for the Loader part of the code, but I guess it would better be in a separate unit test.

tools/loader/Loader.h
```diff
@@ -141,6 +167,18 @@ class Loader {
...
+  /// Register a loader extension.
+  Loader &registerExtension(std::unique_ptr<LoaderExtension> ext);
+  /// Called once after ONNX or Caffe2 model loading.
+  void postModelLoad(PlaceholderBindings &bindings, ProtobufLoader &protoLoader,
+                     size_t compilationBatchSize);
+  /// Called at the beginning of each mini-batch inference.
+  void inferInitMiniBatch(PlaceholderBindings &bindings, size_t minibatchIndex,
+                          size_t minibatchSize);
+  /// Called after the completion of each mini-batch inference.
+  void inferEndMiniBatch(PlaceholderBindings &, size_t minibatchIndex,
```
jfix71: I'm curious what exactly this one could be used for. Do you have any examples/ideas in mind, e.g. is this for processing results coming from a segmentation model or something? My first instinct was that this would be used as a replacement over `processAndPrintResults()` or something.

tlepley-cadence: Currently we are using Loader extension mainly for tensor instrumentation/dump and also for custom quantization profile generation.  `inferEndMiniBatch` is useful as soon as you want to scan the whole generated data and doing some reduction algorithm (the placeholder value of a minibatch execution will be lost when running the next mini-batch).

tools/loader/ImageClassifier.cpp
```diff
@@ -613,6 +614,9 @@ int main(int argc, char **argv) {
           glow::make_unique<TraceContext>(TraceLevel::STANDARD));
     }
     Loader loader;
+
+    // Register Loader extensions here.
```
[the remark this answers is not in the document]
tlepley-cadence: Done. Currently, we mainly use extension for ourselves. It's a way to extend the loader without to overload the Loader class (that is already IMO a bit too big). There are in some situations where we have  specific code that we don't plan to upstream and that we want physically separated from the Loader class for the sake of SW architecture.

tlepley-cadence:
@jfix71 All done. Please note that I managed to write a Loader extension test. This is the first unit test for Loader so I've created a new test class for it.

tlepley-cadence:
The PR is approved. Can we land it ?

beicy:
@tlepley-cadence  Sorry for the delay. Please run "./utils/format.sh" to fix the format issue, and after that, feel free to land it.

tlepley-cadence:
@beicy the formating issue is not in my PR but in the master branch. It has been fixed 2 days ago and I've not rebased my branch. Do I really need to rebase ?

beicy:
> @beicy the formating issue is not in my PR but was in the master branch. It has been fixed 2 days ago (see fix here : #4138) and I've not rebased my branch. Do I really need to rebase ?
> 
> EDIT: OK, just rebased to benefit the the _master_ fix

Thanks!

tlepley-cadence:
The master branch is broken again (folly dependency it seems) so rebasing my branch doesn't help.

PR #4058, opened 2020-01-24
Removed extraneous output from GraphOptimizer.cpp

mjanderson09:
Reviewed By: jfix71

Differential Revision: D19557678


opti-mix:
@mjanderson09 

Nit: Why `VLOG` and not `DEBUG_GLOW` as in most other places? It is probably not that important, but it would be nice to have more uniformity.

PR #4061, opened 2020-01-24
Disable constant folding in loader when called via checkGraphCompatibility

jfix71:
Summary: `checkGraphCompatibility()` builds a single Node with all Constant inputs, where the Constants are uninitialized. This causes problems for constant folding at loading time, as it was trying to constant fold with uninitialized data.

Reviewed By: jackm321, yinghai

Differential Revision: D19564071


jfix71:
Merged into https://github.com/pytorch/glow/pull/4027.

PR #4064, opened 2020-01-25
Some further progress in making backends more self-contained

opti-mix:
This PR provides some partial fixs for https://github.com/pytorch/glow/issues/3785

- Avoids explicit enumeration of all backend names in BackendTestUtils.h
- Moves backend-specific ONNX model writers into the sub-directories containing backend sources
   Yet another place in glow/master that had to know about all existing backends is removed.

### DIFF
--- a/tests/unittests/OperatorTest.cpp
+++ b/tests/unittests/OperatorTest.cpp
@@ -3113,21 +3113,21 @@ static void testFlip(glow::PlaceholderBindings &bindings, glow::Module &mod,
 
 /// Test Flip 1D with Int8.
 TEST_P(OperatorTest, Flip1D_Int8) {
-  ENABLED_BACKENDS(Interpreter, CPU);
+  ENABLED_BACKENDS("Interpreter", "CPU");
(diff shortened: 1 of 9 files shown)


tests/unittests/OperatorTest.cpp
```diff
@@ -3113,21 +3113,21 @@ static void testFlip(glow::PlaceholderBindings &bindings, glow::Module &mod,
 
 /// Test Flip 1D with Int8.
 TEST_P(OperatorTest, Flip1D_Int8) {
-  ENABLED_BACKENDS(Interpreter, CPU);
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
bertmaher: I thought we deprecated `ENABLED_BACKENDS` in favor of `CHECK_IF_ENABLED`?  (Am I wrong?)

opti-mix: I was not following this discussion closely. This PR only mechanically modified all existing places to use string literals instead of custom symbolic names introduced in `BackendTestUtils.h`. If we want to migrate `ENABLED_BACKENDS` to `CHECK_IF_ENABLED` we could do it in a separate PR for all the uses of `ENABLED_BACKENDS`.

PR #4066, opened 2020-01-27
[RFC] Added new bindContexts method to backend

gcatron:
Summary:
This is a proposal for a new bindContexts method to be on the backend. 
This method is a setup method that is meant to be called after the provisioner completes. 
Once called the backend is responsible for performing any setup needed for data transfer, this includes creating any copy commands or allocation of transfer space.
Any state generated from this method should either be stored in the DeviceManager or ExecutionContext. 
Once called the deviceManagers should be ready to execute the network with the bound contexts.
In the future this will enable backends to create P2P connection between devices and enable command chaining where execution can continue from one device to the next automatically. 

Documentation:

Test Plan: verify it builds

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -46,6 +46,9 @@ namespace runtime {
 class DeviceManager;
 struct DeviceInfo;
 struct DeviceConfig;
+struct ContextBinding;
+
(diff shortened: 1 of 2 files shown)


gcatron:
@jsubag

PR #4068, opened 2020-01-27
[Graph] Add a new method to NodeBuilder to support many results

SplitInfinity:
**Summary**
This commit adds a `hasExtraResults()` member to NodeBuilder that exposes `Node::addResult`
as a public member. This can be useful for creating nodes with an arbitrary number of outputs.

**Test Plan**
Used `hasExtraResults()` on existing node, examined generated files:
```
class SparseLengthsSumNode final : public Node {
...
  void addExtraResult(TypeRef T) { addResult(T); }
```

### DIFF
--- a/tools/ClassGen/NodeBuilder.cpp
+++ b/tools/ClassGen/NodeBuilder.cpp
@@ -559,6 +559,10 @@ void NodeBuilder::emitNodeClass(std::ostream &os) const {
      << "  Node* clone() const;\n"
      << "  bool verify() const;\n";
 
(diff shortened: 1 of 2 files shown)


SplitInfinity:
The motivation for this pull request is that I want to create a new, backend-specific node that has an arbitrary number of inputs. This is already possible (see `ConcatNode` for example), but I also want to differentiate this new node, which means I need to be able to create a new `Node` with an arbitrary number of outputs too.

SplitInfinity:
@jfix71 

What do you think now?

PR #4069, opened 2020-01-27
Explicitly unsupport unary SLS for perf purpose

yinghai:
Summary: Running unary SLS in NNPI doesn't make sense in terms of perf. But we don't want do blacklist all the SLS.

Differential Revision: D19588326


arunm-git:
CC: @jsubag

PR #4073, opened 2020-01-28
Fix yet another place where all backends had to be explicitly listed

opti-mix:
This is a partial fix for https://github.com/pytorch/glow/issues/3785


opti-mix:
@jfix71 You are too fast! :-D

PR #4075, opened 2020-01-29
[OpenCL] specialize convolutionK

pjaaskel:
Pass some of the properties as constants to enable more aggressive
constant propagation, strength reduction, loop removal, and other
optimizations.

This helps targets that use the non-local memory using convolution
since not having a local scratch pad memory.

Some numbers:

Intel SDK:
inception_v1: From 1.36 to 1.27 (6% drop)
with convolutionW total from 917 ms to 587 ms (36% drop)

mobilenetv2:  From 0.72 to 0.69 (4% drop)
with convolutionW total from 333 ms to 204 ms (39% drop)

resnet50:     From 2.46 to 2.38 (3% drop)
with convolutionW total from 2302 to 1687 ms (27% drop)

pocl:
inception_v1: From 0.57 s to 0.50 s (12% drop)
with convolutionW total from 480 ms to 437 ms (9% drop)

mobilenetv2:  From 0.22 s to 0.17 s (22% drop)
with convolutionW total from 207 ms to 158 ms (23% drop)

resnet50:    From 1.47 s to 1.36 s (7% drop)
with convolutionW total from 1369 ms to 1323 ms (5% drop)

All of these are “best out of 10 runs” numbers on an Intel i7-7500U
CPU OpenCL device.

Test Plan:

Tested with pocl and Intel OpenCL implementations on an Intel i7 CPU.

### DIFF
--- a/lib/Backends/OpenCL/OpenCL.cpp
+++ b/lib/Backends/OpenCL/OpenCL.cpp
@@ -74,6 +74,14 @@ static const unsigned char kernels_fwd_quantized_conv_cl_src[] = {
 static const size_t kernels_fwd_quantized_conv_cl_src_size =
     sizeof(kernels_fwd_quantized_conv_cl_src);
 
+// Non-local memory using convolution for aggressive compile time
+// specialization.
+static const unsigned char kernels_specialized_no_local_mem_conv_cl_src[] = {
+#include "glow/OpenCL/kernels_specialized_no_local_mem_conv.cl.inc"
+};
+static const size_t kernels_specialized_no_local_mem_conv_cl_src_size =
+    sizeof(kernels_specialized_no_local_mem_conv_cl_src);
+
 static llvm::cl::OptionCategory OpenCLBackendCat("Glow OpenCL Backend Options");
 
 llvm::cl::opt<unsigned>
@@ -88,6 +96,14 @@ llvm::cl::opt<bool> clDoProfile("opencl-profile",
                                 llvm::cl::desc("Profile OpenCL kernels"),
                                 llvm::cl::init(false),
                                 llvm::cl::cat(OpenCLBackendCat));
+
+// Since conv_forward_mem is always specialized, this actually affects only
+// the non-local memory using non-quantized one currently, but can be extended
+// later to cover also the quantized one.
+llvm::cl::opt<bool> clSpecializeConvolution(
+    "opencl-specialize-convolution",
+    llvm::cl::desc("Aggressively specialize convolution kernel launches."),
+    llvm::cl::init(false), llvm::cl::cat(OpenCLBackendCat));
 
 llvm::cl::opt<std::string> clDeviceProgramCacheDir(
     "opencl-program-cache-dir",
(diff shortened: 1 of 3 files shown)


pjaaskel:
The failure doesn't look like my fault.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1137,23 +1149,63 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
...
+        addIntOption(options, "CONVK_ODIM_C", odim.c);
+        addIntOption(options, "CONVK_PADS_TOP", pads.top);
+        addIntOption(options, "CONVK_PADS_LEFT", pads.left);
+        addIntOption(options, "CONVK_FILTER_W", kernelSize.w);
+        addIntOption(options, "CONVK_FILTER_H", kernelSize.h);
+        addIntOption(options, "CONVK_FILTER_C", kernelSize.c);
+        addIntOption(options, "CONVK_SPECIALIZED", 1);
+      }
+
+      cl_program specializedProg = createProgram(src, options, commands);
```
opti-mix: This will parse/compile the whole set of kernels, even though you try to specialize only one of them, right?

lib/Backends/OpenCL/kernels.cl
```diff
@@ -1027,50 +1034,147 @@ __kernel void softmaxgradW(__global void *mem, cl_uint32_t origDest,
   softmaxgradK(&mem[srcGrad], &mem[origDest], &mem[selected], sliceSize);
 }
 
+// Aggressive compile-time specialization of convolution parameters.
```
opti-mix: The idea of specializing kernels at the source-code level is really nice!

I wonder how to support it in a more principled way, so that we could easily apply the same trick to any of our kernels with a minimal effort. It may require introducing some machinery to support it, e.g. some macros, etc. Do you have any ideas how to expand it in this direction?

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: Yes, whatever is in that source string gets compiled, so this could be sped up somewhat if the specialized convolutionK was isolated to a separate file. I think the impact of this would be very minor as the optimizations (that are often performed only for the built kernel) take most of the compile time.

In any case, the compile time impact of the specialization is quite high for the first run anyways (and should be considered/discarded as initialization time) so I wouldn't do this solely for compile time, but possibly it would be cleaner if the specialized convolutionK was in another file.

pjaaskel:
> Have you tried to see if this PR results in any performance boost on GPUs? After all, the primary use-case for the OpenCL backend are GPUs. If it does not improve the GPU performance it is not clear if the introduced complexity is worth it.

Yes, I've tested on a GPU which benefits from this, but it's not a public one. This kernel is selected instead of conv_forward_mem in case local memory usage is not beneficial for the device at hand, so for such GPUs this patch should be beneficial. For conv_forward_mem there is already a similar specialization mechanism in place.

Also to me OpenCL is about more than GPUs; heterogeneous computing across "device categories", even though it is seen by most of the community as simply a GPU offload API. This improves (multicore) CPU performance via easier work-group SIMDization etc. and likely helps DSP targets / co-processors without scratch pad memories too.

lib/Backends/OpenCL/kernels.cl
pjaaskel: I tried to think a bit of a more generic specialization mechanism, but couldn't come up with a clean one so far. The problem is the fallback: if we want to partially specialize or not specialize at all using the same source code, the parameters must come from kernel arguments. The current code relies on kernel argument naming. Also since convolutions (and perhaps matmuls) are usually by far the biggest bottlenecks in inference, I didn't yet feel motivated to design a more complex mechanism for this.

And now that I look at it again after your other comment, it might make more sense at least in this simple convolutionK case to not even utilize the same source code for non-specialized and specialized versions for cleanup reasons, but add another source file for a (fully) specialized one and either specialize all paramters or none. Then the end result will look like conv_forward_mem which seems to be always specialized (it refers to it as "parametrized") via macros. I'll look into this.

opti-mix:
@pjaaskel I'll try your PR locally to check the performance on my GPU.

pjaaskel:
> @pjaaskel I'll try your PR locally to check the performance on my GPU.

Likely you won't see any difference since your GPU probably has proper local memory and thus the other conv_forward_mem gets used instead.

pjaaskel:
I separated the specialized conv to another file, which made the implementation cleaner. I also cleaned up the implementation a bit while at it (cl_uint32_t to unsigned, const and restrict qualifiers...).

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -508,6 +530,7 @@ void OpenCLFunction::enqueueKernel(llvm::StringRef name,
 void OpenCLFunction::executeNCHWConvolution(
     const ConvolutionInst *CC, ExecutionContext *executionContext,
     runtime::OpenCLDeviceBindings *devBindings) {
+
```
opti-mix: Why empty line here?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1137,36 +1160,77 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
 
       // This is a naive implementation that parallelizes using three dims:
       // the X and the Y in the output filter.
-      cl_kernel kernel = createKernel(kernelName, program);
-      setKernelArg(kernel, 0, deviceBuffer);
-      auto numArgs = setKernelArgsForBuffers(kernel, I, 1, runtimeBundle_);
-      auto odim = ShapeNHWC(CC->getDest()->getType()->dims());
+
```
opti-mix: Why empty lines around the idim's definition?

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1137,36 +1160,77 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
...
+        addIntOption(options, "CONVK_PADS_TOP", pads.top);
+        addIntOption(options, "CONVK_PADS_LEFT", pads.left);
+        addIntOption(options, "CONVK_FILTER_W", kernelSize.w);
+        addIntOption(options, "CONVK_FILTER_H", kernelSize.h);
+        addIntOption(options, "CONVK_FILTER_C", kernelSize.c);
+        src.append(reinterpret_cast<const char *>(
+                       specialized_no_local_mem_conv_cl_src),
+                   specialized_no_local_mem_conv_cl_src_size);
+      } else {
+        src.append(reinterpret_cast<const char *>(kernels_cl_src),
```
opti-mix: Why do we need create the source and then createProgram for the standard, non-specialized case? It seems like a waste of compute. Can we leave this execution path untouched and only pay the runtime cost for the specialized case?

lib/Backends/OpenCL/specialized_no_local_mem_conv.cl
```diff
@@ -0,0 +1,92 @@
...
+  const size_t inCperG = CONVK_IDIM_C / CONVK_GROUP;
+  const size_t outCperG = CONVK_ODIM_C / CONVK_GROUP;
+  const size_t inChannelOffset = d / outCperG * inCperG;
+
+  typedef int ssize_t;
+  // For each convolution 'jump' in the input tensor:
+  ssize_t x = -(ssize_t)CONVK_PADS_TOP + ax * CONVK_STRIDES_H;
+  ssize_t y = -(ssize_t)CONVK_PADS_LEFT + ay * CONVK_STRIDES_W;
+
+#define X_MIN -(ssize_t)CONVK_PADS_TOP
```
opti-mix: I'd suggest putting all macro definitions to the beginning of the file.

opti-mix:
@pjaaskel A very minor cosmetic request: Could you rename the `specialized_no_local_mem_conv.cl` file into `kernels_specialized_no_local_mem_conv.cl` to follow the naming convention of other `*.cl` files?

pjaaskel:
Done.

PR #4076, opened 2020-01-29
Add more inference timing stats to the ImageClassifier

pjaaskel:
When benchmarking long-running inference scenarions it is (sometimes)
interesting to:

a) Skip the first run from the statistics. This is interesting in case
there is aggressive specialization or other preperation in the first run
which won't recur in the later runs. This is the case at least with
OpenCL when applying kernel specialization which incurs several kernel
online builds for the first run, but the rest of the runs get the built
programs from an in-memory cache.

b) The best runtime. Especially when benchmarking on a desktop computer
with other background tasks that interfere with the execution, the best
runtime of N runs gives a better indication of the achievable peak
performance than the average performance.

This patch adds the ability to exclude the first warmup run from the
average run computation and adds a separate statistics that records
the best run out of all warmup runs.

Summary:

Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/tools/loader/ImageClassifier.cpp
+++ b/tools/loader/ImageClassifier.cpp
@@ -29,6 +29,7 @@
 #include "llvm/Support/raw_ostream.h"
 
 #include <atomic>
+#include <cfloat>
 #include <fstream>
 #include <future>
 #include <iostream>
@@ -154,6 +155,12 @@ llvm::cl::opt<unsigned>
            llvm::cl::desc("How many passes to do to warm everything up"),
            llvm::cl::init(0), llvm::cl::value_desc("W"),
            llvm::cl::cat(imageLoaderCat));
+
+llvm::cl::opt<unsigned> excludedFirstWarmupRuns(
+    "excluded-first-warmup-runs",
+    llvm::cl::desc("Exclude the time of the given number of first warmup runs "
+                   "from the total time"),
+    llvm::cl::Optional, llvm::cl::init(0), llvm::cl::cat(imageLoaderCat));
 
 std::mutex eventLock;
 std::unique_ptr<TraceContext> traceContext;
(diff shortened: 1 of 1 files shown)


tools/loader/ImageClassifier.cpp
```diff
@@ -155,6 +156,11 @@ llvm::cl::opt<unsigned>
            llvm::cl::init(0), llvm::cl::value_desc("W"),
            llvm::cl::cat(imageLoaderCat));
 
+llvm::cl::opt<bool> excludeFirstWarmupRun(
```
opti-mix: Maybe it should allow to skip first N warmup rounds? The default value would be 0.

pjaaskel: Done!

tools/loader/ImageClassifier.cpp
```diff
@@ -459,11 +466,31 @@ static void runInference(runtime::HostManager *hostManager, std::string name,
                          std::unique_ptr<ExecutionContext> batch,
                          std::promise<void> &runPromise,
                          std::atomic<unsigned> &inflight,
-                         std::atomic<int> &dispatched, unsigned warmUp) {
+                         std::atomic<int> &dispatched, unsigned warmUp,
```
opti-mix: Nit: The number of parameters for this function slowly becomes unwieldy and I have the feeling we may want to add more parameters in the future. May be some kind of a compound data structure should be used instead to pass the parameters? Or maybe we should introduce a proper class encapsulating all this data and make this function a method in this class?

pjaaskel: Perhaps, but converting to a class just moves the parameter list to a constructor so I'm not sure how much cleaner it would end up being. How about when there's time to add yet more parameters a refactoring could be in place?

opti-mix: Fair enough ;-)

opti-mix:
@pjaaskel OpenCL is failing for some reason on CI

opti-mix:
@pjaaskel Seems like this PR is not mergeable anymore. CI produces this error "fatal: refusing to merge unrelated histories". Could you rebase and push this PR again? Thanks!

PR #4078, opened 2020-01-29
ONNX import and export for ChannelwiseQuantizedConvolution

jackm321:
Summary:
Simplify ChannelwiseQuantizedConvolutionNode and write ONNXModelWriter case for it.

* Remove lowering for ChannelwiseQuantizedConvolutionNode, if we need this in the future, let's rewrite it using graph api so we don't need everything to be Constants.
* Change bias ElemKind to int32
* Add basic quantized type handling to onnx importer and exporter
* Add onnx export and import for ChannelwiseQuantizedConvolutionNode

Documentation: doxygen

Test Plan:
Added OnnxExporterTest case


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -711,7 +711,7 @@ void BoundInterpreterFunction::fwdChannelwiseQuantizedConvolutionInst(
             }
 
             // Scale the bias to match the scale of the matrix multiplication.
-            AccumulatorTy B = std::round(biasW.at({d}) / matMulScale);
+            AccumulatorTy B = biasW.at({d}) / matMulScale;
```
jackm321: @csummersea

lib/Graph/Graph.cpp
```diff
@@ -2481,15 +2481,28 @@ Convolution3DNode *Function::createConv3D(PlaceholderBindings &bindings,
...
     llvm::ArrayRef<unsigned_t> pads, unsigned_t group) {
   assertConvDims(input, filter, bias, kernels, strides, pads, group);
   auto OT = getParent()->uniqueType(*outTy);
+  auto biasType = bias.getElementType();
+  if (biasType == ElemKind::Int32QTy) {
+    // Nothing to do
+  } else if (biasType == ElemKind::FloatTy) {
+    auto biasType =
+        getParent()->uniqueType(glow::ElemKind::Int32QTy, bias.dims(), outTy->getScale(), 0);
+    bias = createQuantize("quantized_bias", bias, biasType);
```
jackm321: @csummersea

csummersea: The bias and biasType looks good to me.

PR #4092, opened 2020-01-31
Demotion: added Generic Transformation pass int64 --> int32

ayermolo:
Summary: Refactored index demotion from automatic during project creation to compile time pass.
Related to: github.com/pytorch/glow/issues/4071

Documentation:

[Optional Fixes #issue]

Test Plan: Unit tests

### DIFF
--- a/include/glow/Optimizer/GraphOptimizer/CompilationContext.h
+++ b/include/glow/Optimizer/GraphOptimizer/CompilationContext.h
@@ -116,6 +116,10 @@ struct OptimizationOptions {
   /// The number of cores to assign to non-SLS partition when using SparseNN
   /// partitioning scheme
   unsigned int sparseNNPartitioningSchemeNumCoresOther{1};
+
+  /// If true does int64 to int32 type demotion if backend supports for specific
+  /// nodes.
+  bool enableTypeDemotion{true};
 };
 
 /// Context for compilation.
--- a/lib/Backends/CPU/CPUBackend.cpp
+++ b/lib/Backends/CPU/CPUBackend.cpp
@@ -49,8 +49,12 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
         {ElemKind::FloatTy, ElemKind::Int32ITy, ElemKind::Int64ITy});
 
   case Kinded::Kind::AddNodeKind:
-  case Kinded::Kind::SubNodeKind:
   case Kinded::Kind::MulNodeKind:
+    return NI.allInputsAndOutputsHaveSameElemKind(
+        {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int32ITy,
+         ElemKind::Int64ITy});
+
+  case Kinded::Kind::SubNodeKind:
   case Kinded::Kind::MaxNodeKind:
   case Kinded::Kind::MinNodeKind:
   case Kinded::Kind::CPUMaxSplatNodeKind:
@@ -67,13 +71,15 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
     return NI.allInputsAndOutputsHaveSameElemKind(
                {ElemKind::FloatTy, ElemKind::Int8QTy}, {},
                {MaxPoolNode::ArgmaxIdx}) &&
-           (NI.getOutElemTy(MaxPoolNode::ArgmaxIdx) == IndexElemKind);
+           (NI.getOutElemTy(MaxPoolNode::ArgmaxIdx) == ElemKind::Int64ITy ||
+            NI.getOutElemTy(MaxPoolNode::ArgmaxIdx) == ElemKind::Int32ITy);
 
   case Kinded::Kind::ArgMaxNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
                {ElemKind::FloatTy, ElemKind::Int8QTy}, {},
                {ArgMaxNode::ArgmaxIdx}) &&
-           (NI.getOutElemTy(ArgMaxNode::ArgmaxIdx) == IndexElemKind);
+           (NI.getOutElemTy(ArgMaxNode::ArgmaxIdx) == ElemKind::Int64ITy ||
+            NI.getOutElemTy(ArgMaxNode::ArgmaxIdx) == ElemKind::Int32ITy);
 
   case Kinded::Kind::SaveNodeKind:
   case Kinded::Kind::ReshapeNodeKind:
@@ -89,17 +95,17 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
   case Kinded::Kind::ConcatNodeKind:
   case Kinded::Kind::SplatNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
-        {ElemKind::FloatTy, ElemKind::Int8QTy, IndexElemKind,
-         ElemKind::Int64ITy, ElemKind::BoolTy});
-
-  case Kinded::Kind::DivNodeKind:
+        {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int64ITy,
+         ElemKind::Int32ITy, ElemKind::BoolTy});
   case Kinded::Kind::SliceNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
         {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int32QTy,
-         ElemKind::Int64ITy});
+         ElemKind::Int32ITy, ElemKind::Int64ITy});
   case Kinded::Kind::SpaceToDepthNodeKind:
+  case Kinded::Kind::DivNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
-        {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int64ITy});
+        {ElemKind::FloatTy, ElemKind::Int8QTy, ElemKind::Int64ITy,
+         ElemKind::Int32ITy});
 
   case Kinded::Kind::TransposeNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
@@ -117,7 +123,9 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
                {ElemKind::FloatTy}, {SparseLengthsSumNode::IndicesIdx,
                                      SparseLengthsSumNode::LengthsIdx}) &&
            (NI.getInElemTy(SparseLengthsSumNode::IndicesIdx) ==
-            IndexElemKind) &&
+                ElemKind::Int64ITy ||
+            NI.getInElemTy(SparseLengthsSumNode::IndicesIdx) ==
+                ElemKind::Int32ITy) &&
            (NI.getInElemTy(SparseLengthsSumNode::LengthsIdx) ==
             ElemKind::Int32ITy);
 
@@ -127,7 +135,9 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
                {SparseLengthsWeightedSumNode::IndicesIdx,
                 SparseLengthsWeightedSumNode::LengthsIdx}) &&
            (NI.getInElemTy(SparseLengthsWeightedSumNode::IndicesIdx) ==
-            IndexElemKind) &&
+                ElemKind::Int64ITy ||
+            NI.getInElemTy(SparseLengthsWeightedSumNode::IndicesIdx) ==
+                ElemKind::Int32ITy) &&
            (NI.getInElemTy(SparseLengthsWeightedSumNode::LengthsIdx) ==
             ElemKind::Int32ITy);
 
@@ -135,8 +145,9 @@ bool CPUBackend::isOpSupported(const NodeInfo &NI) const {
     return NI.allInputsAndOutputsHaveSameElemKind(
                {ElemKind::FloatTy},
                {EmbeddingBagNode::IndicesIdx, EmbeddingBagNode::OffsetsIdx}) &&
-           (NI.getInElemTy(EmbeddingBagNode::IndicesIdx) == IndexElemKind) &&
-           (NI.getInElemTy(EmbeddingBagNode::OffsetsIdx) == IndexElemKind);
+           (NI.getInElemTy(EmbeddingBagNode::IndicesIdx) ==
+            ElemKind::Int64ITy) &&
+           (NI.getInElemTy(EmbeddingBagNode::OffsetsIdx) == ElemKind::Int64ITy);
 
   case Kinded::Kind::SparseLengthsWeightedSumGradNodeKind:
     // GradOfInputNamedIndicesIdx and GradOfInputNamedLengthsIdx do not need to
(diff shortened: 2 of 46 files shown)


lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -451,3 +487,19 @@ llvm::StringRef CPUBackend::getLibjitBitcode() const {
   return llvm::StringRef(reinterpret_cast<const char *>(libjit_bc),
                          libjit_bc_size);
 }
+
+/// Returns true if network supports Type Lowering from T1 to T2.
```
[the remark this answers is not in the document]
pjaaskel: So perhaps this is only a naming issue: canDoIndexTypeDemotion?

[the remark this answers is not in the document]
ayermolo: Well originally I was thinking this API can be used for other types of conversions, but yeah this might be an over kill. Can rename.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3411,6 +3451,28 @@ static void setFP16AccumSLS(Function *F,
   } while (nodeIt != stopIt);
 }
 
+static void transformIndexTypeDemotion(const Backend &B, Function *F,
```
pjaaskel: OK. Now that I look at this again, it is indeed generic as it relies on knowledge of the type's actual bitwidth. At least in its current state it does not differentiate the semantics of that integer data: it can be whatever as long as it's safe to clip the data without losing semantical information (actualSize()).

pjaaskel:
Yes, indeed that's the case. 32b dim_t mode breaks on OpenCL with this patch. 64b works.

ayermolo:
Thanks for the catch. Reverted kernels that they are now 64bit, since type demotion is disabled for OpenCL. OpenCLOperatorTest passes locally.

pjaaskel:
OK. After the 32b indices work again with OpenCL kernels, this looks good to me. Before that, it regresses back to before my original 32b dim_t commit, did I understand it correctly?

ayermolo:
Yes for the kernels themselves.

ayermolo:
@bertmaher @opti-mix @jfix71 Hi. Can you take a look at this MR, since you were involved in the original one @pjaaskel submitted.
Thanks

tests/unittests/OperatorTest.cpp
```diff
@@ -11458,5 +11510,305 @@ TEST_CAST_2WAYS(int64_t, int64_t, Int64ITy, Int64ITy, /* castIsNoOp */ true)
...
+  auto *save = F->createSave("save", N);
+  auto *saveTensor = bindings.allocate(save->getPlaceholder());
+
+  EE.compile(CompilationMode::Infer);
+  EE.run(bindings);
+
+  return saveTensor->getHandle<DataType>();
+}
+
+/// Check that the Mul operator behaves correctly with FP16.
```
jfix71: nit: comment says FP16 but appears to be int32.

tests/unittests/OperatorTest.cpp
```diff
@@ -11458,5 +11510,305 @@ TEST_CAST_2WAYS(int64_t, int64_t, Int64ITy, Int64ITy, /* castIsNoOp */ true)
...
+        for (dim_t f = 0; f < saveH.dims()[3]; ++f) {
+          EXPECT_EQ(xValues[counter] * yValues[counter],
+                    saveH.at({i, j, k, f}));
+          ++counter;
+        }
+      }
+    }
+  }
+}
+/// Check that the Mul operator behaves correctly with FP16.
```
jfix71: nit: comment says FP16 but appears to be float.

tests/unittests/OperatorTest.cpp
```diff
@@ -3003,7 +3003,40 @@ TEST_P(OperatorTest, TopK1) {
...
+  EXPECT_FLOAT_EQ(V.at({0, 0, 0}), 18);
+  EXPECT_EQ(I.at({0, 0, 0}), 1);
+  EXPECT_FLOAT_EQ(V.at({1, 0, 0}), 41);
+  EXPECT_EQ(I.at({1, 0, 0}), 3);
+  EXPECT_FLOAT_EQ(V.at({2, 0, 0}), 34);
+  EXPECT_EQ(I.at({2, 0, 0}), 2);
+}
+
+// Check the TopK operator for the special case of K=1.
+TEST_P(OperatorTest, TopK1int32) {
```
jfix71: nit: Appears to be nearly identical to `TopK1` -- can we templatize the original to consolidate them?

lib/Backends/Interpreter/tests/InterpreterOperatorTest.cpp
```diff
@@ -23,4 +23,5 @@ std::set<std::string> glow::backendTestBlacklist = {
     "RepeatedSLSWithPartialTensors/0",
     "SigmoidSweep_Float16/0",
     "TanHSweep_Float16/0",
+    "TopK1int32/0",
```
jfix71: In general we strive to keep support for everything in the Interpreter -- can we add support for this? Hopefully simple as templatizing the kernel in `InterpreterNodes.cpp`?

include/glow/Optimizer/GraphOptimizer/CompilationContext.h
```diff
@@ -116,6 +116,9 @@ struct OptimizationOptions {
   /// The number of cores to assign to non-SLS partition when using SparseNN
   /// partitioning scheme
   unsigned int sparseNNPartitioningSchemeNumCoresOther{1};
+
+  /// If true does int64 to int32 type demotion
```
jfix71: Can you expand here to make clear it will only do the conversion if the backend also supports it.

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -451,3 +487,19 @@ llvm::StringRef CPUBackend::getLibjitBitcode() const {
...
+/// Allows the backend to control node white/black list for optimization.
+bool CPUBackend::canDoIndexTypeDemotion(
```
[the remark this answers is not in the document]
ayermolo: Well canDoIndexTypeDemotion can rely on isOpSupported. Although I am not sure isOpsupported is representative enough. The way I see it is it answers a question: Does backend X support node Y for specific combination of types. Which worked because we were just operating on one type: float. Now we are operating on two independent types, in two passes. So there is a state in which the answer will be false.

To clarify let's say we have Gather Node. It's input is float and Index of type int64. Now backend X only supports Quantized input and index of type int32. When we query it for quantization we will be asking does it support Quantized and index int64. Because we are only intersted in quantization part. If isOpSupported implemented as it is intended the answer has to be no. Because it only supports index type of 32. 

Similar thing happens when we querry it for index type demotion.

If goal is to eliminate BE from being passed around, maybe a better approach is generatlize canDoIndexTypeDemotion back to canDoTypeDemotion. It then can return a black list of nodes that can't be transformed and it can be used by generic TypeAToTypeBFunctionConverter, and passed to FunctionQuantizer after BE is queried in quantizeFunction.

ayermolo:
@jfix71  thanks for looking at it! NP, I realize we are all busy. :)

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3411,6 +3451,28 @@ static void setFP16AccumSLS(Function *F,
...
+                                       CompilationContext &cctx) {
+  for (auto &n : F->getNodes()) {
```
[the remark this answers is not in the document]
ayermolo: Will do. Yep, just a coarse/sanity check. I don't think this is a common case, and if we do encounter it in a net can revisit and refine.

ayermolo:
@jfix71 Anything else I need to do, answered all your questions?

Right now it has my foly fix as part of it, in case other changes are needed. Once all is good I can rebase on pure master and push it in.

pjaaskel:
@ayermolo does this still break OpenCL with 32b dim_t?

ayermolo:
@pjaaskel Well that was fixed a while ago, but thanks for reminder. Double checked and fixed some issues with latest rebase (this morning). Now all unit tests pass in release on my system with -DTENSOR_DIMS_32_BITS=ON.

Which brings me to two points.
1) I think 32 bit should be added to CI, otherwise we might have future failures.
2) This MR has been open for almost a month now. It takes certain amount of time to keep things up to date and working. There doesn't seem to be any major objections. Can we get this through by end of week please?

pjaaskel:
@ayermolo I agree that 32b should be added to CI. It now only build tests with it, but doesn't run the tests. I'll see into it.

ayermolo:
@pjaaskel cool.

ayermolo:
@jfix71 ping. :)

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -455,3 +491,19 @@ llvm::StringRef CPUBackend::getLibjitBitcode() const {
   return llvm::StringRef(reinterpret_cast<const char *>(libjit_bc),
                          libjit_bc_size);
 }
+
+/// Returns true if network supports Type Lowering from T1 to T2.
```
jfix71: nit: `\returns`

lib/Backends/CPU/CPUBackend.cpp
```diff
@@ -455,3 +491,19 @@ llvm::StringRef CPUBackend::getLibjitBitcode() const {
...
+/// Allows the backend to control node white/black list for optimization.
```
jfix71: This is not clear -- this is returning a blacklist of ops that cannot be converted, right? I'd just say that specifically, otherwise it's not clear how this is used. Same for all places where this comment exists.

tests/unittests/OperatorTest.cpp
```diff
@@ -11981,5 +12014,305 @@ TEST_CAST_2WAYS(int64_t, int64_t, Int64ITy, Int64ITy, /* castIsNoOp */ true)
...
+  auto *save = F->createSave("save", N);
+  auto *saveTensor = bindings.allocate(save->getPlaceholder());
+
+  EE.compile(CompilationMode::Infer);
+  EE.run(bindings);
+
+  return saveTensor->getHandle<DataType>();
+}
+
+/// Check that the Mul operator behaves correctly with FP16.
```
jfix71: Stale comment, refers to wrong op/datatype.

tests/unittests/OperatorTest.cpp
```diff
@@ -11981,5 +12014,305 @@ TEST_CAST_2WAYS(int64_t, int64_t, Int64ITy, Int64ITy, /* castIsNoOp */ true)
...
+          EXPECT_EQ(xValues[counter] + yValues[counter],
+                    saveH.at({i, j, k, f}));
+          ++counter;
+        }
+      }
+    }
+  }
+}
+
+/// Check that the Mul operator behaves correctly with FP16.
```
jfix71: Stale comment, refers to wrong op/datatype.

tests/unittests/OperatorTest.cpp
```diff
@@ -11981,5 +12014,305 @@ TEST_CAST_2WAYS(int64_t, int64_t, Int64ITy, Int64ITy, /* castIsNoOp */ true)
...
+        for (dim_t f = 0; f < saveH.dims()[3]; ++f) {
+          EXPECT_EQ(xValues[counter] + yValues[counter],
+                    saveH.at({i, j, k, f}));
+          ++counter;
+        }
+      }
+    }
+  }
+}
+/// Check that the Mul operator behaves correctly with FP16.
```
jfix71: Stale comment, refers to wrong op/datatype.

include/glow/Optimizer/GraphOptimizer/CompilationContext.h
```diff
@@ -116,6 +116,10 @@ struct OptimizationOptions {
   /// The number of cores to assign to non-SLS partition when using SparseNN
   /// partitioning scheme
   unsigned int sparseNNPartitioningSchemeNumCoresOther{1};
+
+  /// If true does int64 to int32 type demotion if backend supports for specific
+  /// nodes
```
jfix71: nit: period.

tests/unittests/BackendTestUtils.cpp
```diff
@@ -310,6 +310,7 @@ void compareAgainstInterpreter(llvm::StringRef backendName,
   assert(resultTensors.size() == count &&
          "Should get the same number of Tensors back as count.");
 
+  cctxB.optimizationOpts.enableTypeDemotion = false;
```
jfix71: Why is this necessary? Maybe add comment?

tests/unittests/OperatorTest.cpp
```diff
@@ -10710,7 +10739,10 @@ TEST_P(OperatorTest, DivSizeT) {
   auto *result = F_->createSave("save", R);
   bindings_.allocate(result->getPlaceholder());
 
-  EE_.compile(CompilationMode::Infer);
+  CompilationContext cctx;
+  cctx.compMode = CompilationMode::Infer;
+  cctx.optimizationOpts.enableTypeDemotion = false;
```
jfix71: Also curious why this is necessary here?

ayermolo: The test specifically tests div of Int64ITy/size_t. So disabling optimization to preserve test intention.

tests/unittests/BackendTestUtils.cpp
ayermolo: I think originally I wanted to keep it same as Interpreter since there it's disabled, but on more thought I don't think it's necessary. Removed

ayermolo:
@jfix71 No problem. We are all busy with internal stuff. :) Thanks for looking at it. I think I addressed the comments, and pushed latest.

pjaaskel:
I fixed 32b dim_t issues in https://github.com/pytorch/glow/pull/4227. ninja check pass both with 32b and 64b with it (except TensorsTest, which fails for some reason in master too, but not when ran separately).

So if you rebase on top of that and run 'ninja check' with 32b dim_t and OpenCL enabled, it should reveal the most critical problems.

jfix71:
Thanks @ayermolo! @pjaaskel -- let's land this first and rebase https://github.com/pytorch/glow/pull/4227 after/fix any issues that come up. This PR has been waiting a while and is pretty big so want to avoid more merge conflicts.

ayermolo:
@jfix71  Thanks! 
Looks like there are now merge conflicts that are new since yesterday. Want me to rebase and push updated?

jfix71:
@ayermolo Don't worry about it, I will fix it internally and land it.

ayermolo:
Awesome, thanks.

pjaaskel:
OK, #4227 should now pass ninja check with dim_t 32b and 64b. However, the OpenCL is now regressed back to only supporting 64b indices.

@ayermolo did you already have some good ideas to implement int32 lowering for OpenCL in a generic manner, or should it just specialize with the two alternatives depending on the operand types?

ayermolo:
@pjaaskel I haven't thought of how to do it generically in openCL. On libjit CPU side we have this restriction on name mangling, I am not sure if it's same on openCL side or not.

PR #4094, opened 2020-01-31
ConvertTo should be a data parallel node

shajrawi:
Summary:

Probably an oversight - convert to should be data parallel graph node.

Test Plan:

ninja test

### DIFF
--- a/lib/Graph/TensorLayout.cpp
+++ b/lib/Graph/TensorLayout.cpp
@@ -602,11 +602,15 @@ bool TensorLayoutCommon::isSatisfiedBy(
   return true;
 }
 
-static std::string returnBaseReqOrNHWC(std::string baseReq) {
+static std::string returnBaseReqOrNHWC(std::string baseReq, const Node *node) {
   auto baseReqHelper = TensorLayoutDescription(baseReq);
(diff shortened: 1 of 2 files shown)


lib/Graph/TensorLayout.cpp
```diff
@@ -602,12 +602,16 @@ bool TensorLayoutCommon::isSatisfiedBy(
...
 
-static std::string returnBaseReqOrNHWC(std::string baseReq) {
+static std::string returnBaseReqOrNHWC(std::string baseReq, const Node *node) {
   auto baseReqHelper = TensorLayoutDescription(baseReq);
   if (!baseReqHelper.isSameLayout(
           CanonicalTensorLayout::getInstance().getLayoutsForDims()[4])) {
     return baseReq;
   }
+  if (CanonicalTensorLayout::getInstance().acceptsAnyLayout(node)) {
+    // These nodes accept any 4-D layout.
```
opti-mix: Is it important that the node has specifically a 4-D layout?

shajrawi: yes. the function is called `returnBaseReqOrNHWC` and in canonical mode only 4-D got special requirements

PR #4111, opened 2020-02-04
Adds allocateDeviceIOBuffer to the deviceManager 

gcatron:
Summary:
This adds a new api method to the Backend, allocateDeviceIOBuffer. This is used to allow a backend to allocate a buffer that satisfies any requirements needed for the buffer to be used for transfer to/from the device.
Documentation:

Test Plan:


gcatron:
@jsubag

PR #4113, opened 2020-02-04
Fix a number of places in the graph optimizer where alignments were lost during transformations

opti-mix:
Introduce new helpers for creating a type with dimensions and strides/alignments of another type.

Add tests checking that alignments are not lost during transformations.


lib/Graph/NodeValue.cpp
```diff
@@ -36,7 +36,17 @@ NodeValue::NodeValue(Node *N, unsigned resNo) {
 void NodeValue::replaceAllUsesOfWith(NodeValue v, const Function *F,
                                      Node *skipReplacement) const {
   if (v.getNode()) {
-    assert(getType() == v.getType() && "Replacing value with the wrong type");
+    if (getType() != v.getType()) {
```
jfix71: nit: could fold this condition into above if, and remove one indentation level here.

PR #4116, opened 2020-02-05
Gracefully fail when constant evaluation function can't be compiled or run

nickgg:
Summary: Coming back to the issue of constant folding: still having trouble with backend specific Nodes which are not implemented on the interpreter. This PR changes the behaviour so that if there is a terminal failure during CF it stops trying to constant fold. 

This usually isn't an issue in Glow since backend specific nodes are definitionally not standard and so don't often occur in input networks, but it comes up in unit tests for some backends.

Documentation: NFC

Test Plan: Ran tests in debug and release, ran my backends tests.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
+++ b/lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
@@ -154,13 +154,13 @@ static void bailOnNonCanonicalLayout(
 /// Evaluates a provided constant operation \p C using the provided \p backend
 /// and using the compilation context \p cctx.
 /// \returns constant results.
-std::vector<Constant *>
-evaluateConstantOperation(Backend &backend, CompilationContext &cctx, Node *C) {
+bool evaluateConstantOperation(Backend &backend, CompilationContext &cctx,
+                               Node *C, std::vector<Constant *> &constResults) {
   PlaceholderBindings bindings;
   assert(isConstantOperation(C, backend) && "Expected a constant expression");
   // Constants and splats do not need to be constant evaluated.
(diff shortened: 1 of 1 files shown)


lib/Optimizer/GraphOptimizer/ConstantFolding.cpp
```diff
@@ -311,7 +312,10 @@ bool glow::ConstantFold::run(Function *F, const CompilationContext &cctx) {
     }
 
     // Compute the constant value of the node.
-    std::vector<Constant *> constResults = constantFoldNodeImpl(*backend, N);
+    std::vector<Constant *> constResults;
```
opti-mix: Maybe use a SmallVector here to avoid heap allocations?

nickgg: We'd have to replace the vector everywhere in this file, but we could do it. I'm not sure theres a sensible number of results to use for the SmallVector though, it scales proportionally with the size of the network.

opti-mix: Yeah, let's keep it simple. Ignore my previous comment!

nickgg:
Don't think that works - we already have a check on `isOpSupported()`. To trigger this case you need to find a case where you're compiling for backend A which supports all nodes in the graph, but the Interpreter does not. Can you think of a case where that could happen? No backend supports that Select node configuration for example.

nickgg:
OK, having talked to @opti-mix and @jfix71 offline I have convinced them this is a) already covered implicitly and b) super complicated to test explicitly. I'll land and one day when im looking for good karma come back and write the mock backend that can handle this test.

PR #4117, opened 2020-02-05
[ONNXModelLoader] Update Onnx Matmul Operator to support inputs of dimension 3

sladwa:
Some of the recommendation systems when represented in onnx format have matmul operator whose inputs are of dimension size 3. Currently OnnxModelLoader does not support this requirement. loadMatmul API has been updated such that when dimension size is 3 then use BatchMatmulNode. 

Test Plan: A sample .onnxtxt has been added along with a test case in OnnxImporterTest


sladwa:
Issue reported by CircleCI is fixed and changes have been amended to same commit ID

PR #4127, opened 2020-02-07
[GraphOptz] Add verify check in TearDown

jfix71:
Summary: Add a verification check in the TearDown for GraphOptzTests. 

- Some tests created MatMul with invalid dimensions
- One converted to an incorrect type for fused quantized kinds
- One relied on ConvertTo optimizations removing converts for quantization, but we don't use ConvertTo for quantization, so moved it to indices (`ITy`).
- TODO: Some tests still fail for padding (see https://github.com/pytorch/glow/issues/4128)

Test Plan: Tested manually.

### DIFF
--- a/tests/unittests/BackendTestUtils.h
+++ b/tests/unittests/BackendTestUtils.h
@@ -183,6 +183,9 @@ protected:
                                              &optimizedBindings, allowedError));
   }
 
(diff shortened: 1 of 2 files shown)


opti-mix:
@jfix71 Thanks a lot for fixing this!

PR #4129, opened 2020-02-07
Add support for cloning IR functions and IR instructions

opti-mix:
Follow the example of the functionality available for the graph-level cloning:
- Introduce a clone method in the Instruction class and auto-generate the implementations of cloning functions by means of InstrGen.
- Add support for cloning IRFunctions

### DIFF
--- a/lib/IR/IR.cpp
+++ b/lib/IR/IR.cpp
@@ -609,6 +609,61 @@ std::string IRFunction::toString() const {
   return os.str();
 }
 
+IRFunction *
+IRFunction::clone(llvm::StringRef newName,
+                  llvm::DenseMap<const Value *, Value *> *map,
+                  llvm::DenseMap<const Value *, Value *> *currToNewMap) {
+  // Create a new function.
+  auto *newF = new IRFunction(getGraph());
+  newF->setName(newName);
+  return clone(newF, map, currToNewMap);
+}
+
+IRFunction *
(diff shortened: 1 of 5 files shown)


lib/IR/IR.cpp
```diff
@@ -359,6 +359,24 @@ bool Instruction::isDataParallel() const {
...
+  case Kinded::Kind::CLASS##Kind: {                                            \
+    auto *X = llvm::cast<const CLASS>(this);                                   \
+    return X->clone();                                                         \
+    break;                                                                     \
+  }
+#define DEF_BACKEND_SPECIFIC_INSTR(CLASS, NAME) DEF_INSTR(CLASS, NAME)
+#define DEF_VALUE(CLASS, NAME)
+#include "glow/AutoGenInstr.def"
+  }
+  return nullptr;
```
jfix71: Huh, we need this? Assumed default case with llvm_unreachable was sufficient.

opti-mix: Old habit! If a function has a return type different than void, it should return a value of this type ;-) Some (old) compilers always require that.

opti-mix:
@jfix71 

>  One thing I noticed is that we don't treat uniquing names the same between Node and Inst. Node name is always uniqued when we do Function::addNode(). But IRFunction::insertInstruction() does not. Maybe we should add uniquing there too?

Hmm. Maybe. But definitely not as part of this PR as this problem exists since ever  ;-)

PR #4130, opened 2020-02-08
Improved dynamic registration/unregistration of backends

opti-mix:
- Add a more dynamic way to register new backends and backend factories

   The name of the backend being registered is not hard-coded and can be provided as a parameter. The expression to generate new instances of the backend can also be dynamic.

- Unregister backend factories once they are being destructed

  This is useful e.g. for testing. You can define new backends inside your test and they will be automatically unregistered once the test (or the scope where they are declared) are finished.

- Allow instances of backends to have names 

   This is useful if you want to register multiple instances of the same backend as different backend factories with different names. getBackendName would still report the usual backend name as before unless you explicitly call setName for a backend instance.

- Add a test for dynamic backend factories registration

### DIFF
--- a/lib/Backends/CPU/CPUBackend.h
+++ b/lib/Backends/CPU/CPUBackend.h
@@ -38,7 +38,9 @@ public:
   ///@{
   virtual ~CPUBackend() override = default;
 
-  std::string getBackendName() const override { return getName(); }
+  std::string getBackendName() const override {
+    return Named::getName().empty() ? getName() : Named::getName().str();
+  }
   static std::string getName() { return "CPU"; }
   static unsigned numDevices();
 
--- a/tests/unittests/BackendTest.cpp
+++ b/tests/unittests/BackendTest.cpp
@@ -16,6 +16,7 @@
 #include "BackendTestUtils.h"
 
 #include "glow/Backend/BackendUtils.h"
+#include "glow/Backends/Interpreter/Interpreter.h"
 #include "glow/ExecutionEngine/ExecutionEngine.h"
 #include "glow/Graph/Graph.h"
 #include "glow/Graph/PlaceholderBindings.h"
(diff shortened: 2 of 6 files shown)


lib/Backends/CPU/CPUBackend.h
```diff
@@ -38,7 +38,9 @@ class CPUBackend : public LLVMBackend {
   ///@{
   virtual ~CPUBackend() override = default;
 
-  std::string getBackendName() const override { return getName(); }
+  std::string getBackendName() const override {
+    return Named::getName().empty() ? getName() : Named::getName().str();
```
opti-mix: I'm not quite happy about having a static `getName` in each backend and a dynamic `Named::getName()` used to set the name dynamically. But I was hesitant to completely re-factor the existing registration machinery.

jfix71: Hmm. Is it possible to ditch the static `getName()` in the backends and instead always use `Named`? e.g. `Interpreter() : Named("Interpreter") = {};` and then we could always just use `Named::getName()`?

tests/unittests/BackendTest.cpp
```diff
@@ -107,6 +108,41 @@ TEST(Interpreter, profileQuantizationForANetwork) {
...
+      // Check that a new backend is registered.
+      auto backends = getAvailableBackends();
+      EXPECT_NE(std::find(backends.begin(), backends.end(), backendName),
+                backends.end());
+      // The new backend factory will be destroyed at the end of this scope.
+    }
+    // Check that a new backend is not registered anymore after its factory was
+    // destroyed.
+    auto backends = getAvailableBackends();
+    EXPECT_EQ(std::find(backends.begin(), backends.end(), "CustomInterpreter"),
```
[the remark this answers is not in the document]
opti-mix: Oops. Good catch! Thanks!

lib/Backends/CPU/CPUBackend.h
opti-mix: Yes, I've thought about it and I'm not sure. I think it would interfere with the registration mechanism. Currently, we do this inside factory classes expanded from the macro:
```c++
std::string getRegistrationKey() const override { 
      return BackendClass::getName(); 
}
unsigned numDevices() const override { 
      return BackendClass::numDevices(); 
}   
```

It relies on the possibility of obtaining the name (and numDevices) of the backend without creating an instance of the backend. I don't know if we want to maintain this property. @rdzhabarov introduced this mechanism and probably knows reasons for that. 

If we decide it is fine to create an instance of the backend to obtain the name and number of devices, we can totally remove "static" from those methods.

opti-mix:
@jfix71 

> One question is if we use this in two unit tests with the same dynamic registration key/name, and then the tests are run multithreaded, could we encounter an issue? I don't recall/know exactly how tests are run multithreaded in sandcastle/gtest-parallel etc., but if part of the same address space then it seems like they could interfere?

Yes, I think these scenarios could create a race condition and they could probably do it even without this PR as our backend registration factory is a global object. But I don't think this scenarios are a real concern? Do we really expect a race condition related to the registration of backends in any realistic use-case? ;-)

lib/Backends/CPU/CPUBackend.h
jfix71: Ah right, OK. It seems like we could change the macro for `getRegistrationKey()` to return `#BackendClass`, though I suppose the class could then return a different name for `getName()`. I'm not sure we actually need `numDevices()` to be a static method, do we? Would this help?

jfix71:
>they could probably do it even without this PR as our backend registration factory is a global object

Well the factory is initialized statically, right? So when the program launches they're all registered, and then if it spawns threads they wouldn't be registered again. But here they're not initialized statically, so them being global seems like an issue.

>Do we really expect a race condition related to the registration of backends in any realistic use-case? ;-)

Not really, I just like to think of ways to break things 😈

lib/Backends/CPU/CPUBackend.h
opti-mix: > It seems like we could change the macro for getRegistrationKey() to return #BackendClass,

The name of the BackendClass is e.g. CPUBackend, but the name of the backend is "CPU". 

The numDevices could probably be a non-static method, though DeviceManager::numDevices uses it through a backend factory as far as I can see.

PR #4139, opened 2020-02-10
Add the ability to intercept the processing of instructions by the interpreter and implement custom handling

opti-mix:
This functionality is useful for testing and debugging, e.g. in the following cases:
- Changing the handling of specific instructions by the interpreter
- Performing certain actions before/after/instead of a usual instruction processing by the interpreter
- Extend the interpreter to support custom instructions without the need to change the interpreter source code

### DIFF
--- a/include/glow/IR/IR.h
+++ b/include/glow/IR/IR.h
@@ -225,6 +225,56 @@ protected:
   void dumpOperands(llvm::raw_ostream &os) const;
 };
 
+/// Different stages of processing an IR instruction. Transitions between stages
+/// always happen in the same order as they are defined below, i.e.
+/// PROCESSING -> POSTPROCESSING
+/// In particular, there is no possibility for any transition in the backwards
+/// direction.
+enum class IRInstructionProcessingStage {
+  /// Instruction is being processed.
+  PROCESSING,
+  /// Instruction was just processed.
+  POSTPROCESSING,
+};
+
+/// A function for processing instructions e.g. during a code generation or
+/// during a code execution of instructions by backends. A processing function
+/// takes an instruction \p I, a current instruction execution stage \p
+/// executionStage and current context \p ctx as inputs, processes the
+/// instruction and \returns true if the client should proceed to the next
+/// execution stage or false if the client should proceed with the current
+/// execution stage. The processing of the instruction will continue by the
+/// caller (e.g. by a backend) from this stage. For example, if this processing
+/// function has processed the instruction it may decide to return a stage
+/// indicating that the processing has finished so that the caller does not
+/// process it anymore. The passed \p ctx can be anything, e.g. a backend, an
+/// ExecutionContext, etc.
+using IRInstructionProcessingFn =
+    std::function<bool(const Instruction *I,
+                       IRInstructionProcessingStage executionStage, void *ctx)>;
+
+/// The interface class for IR instruction handlers. Useful for intercepting IR
+/// instructions processing.
+class IRInstructionProcessingHandler {
+public:
+  IRInstructionProcessingHandler() = default;
+  virtual ~IRInstructionProcessingHandler() = default;
+  /// Set the handler to be used for IR instruction processing.
+  virtual void
+  setIRInstructionProcessingHandler(IRInstructionProcessingFn hook) {
+    handler_ = hook;
+  }
+  /// \returns the handler to be used for IR instructions processing.
+  virtual const IRInstructionProcessingFn &
+  getIRInstructionProcessingHandler() const {
+    return handler_;
+  }
+
+protected:
+  /// The handler function to be used for instruction processing.
+  glow::IRInstructionProcessingFn handler_;
+};
+
 //===----------------------------------------------------------------------===//
 // TaggedListTraits for glow::Instruction
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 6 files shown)


include/glow/IR/IR.h
```diff
@@ -225,6 +225,54 @@ class Instruction : public Value, public TaggedListNode<Instruction> {
...
+  /// Instruction is being processed.
+  PROCESSING,
+  /// Instruction was just processed.
+  POSTPROCESSING,
+  /// Instruction processing is done.
+  DONE,
+};
+
+/// A function for processing instructions e.g. during a code generation or
+/// during execution of instructions by backends. A processing function takes an
```
SameerAsal: nit: "during code"

opti-mix:
@jfix71 

I addresses all your cosmetic comments.

> Or perhaps a compromise between the two would be for the handler to return a bool for success on whether to move onto the next stage? 

Good idea. I implemented it in the latest version. Also improved the docs about processing stages.

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -229,21 +231,55 @@ Error BoundInterpreterFunction::execute(IRFunction *F,
...
+        glow::errs() << "Invalid instruction: " << &I << "\n";
+        llvm_unreachable("Invalid instruction.");
+      }
+    }
+  } else {
+    auto &irInstructionProcessingHandler = getIRInstructionProcessingHandler();
+    // Dispatch the interpreter on each instruction in the program:
+    for (const auto &I : F->getInstrs()) {
+      // Perform pre-processing and proceed with standard processing if needed.
+      if (!irInstructionProcessingHandler(
```
SameerAsal: Having PREPROCESSING and PROCESSING seems a bit redundant to me. If you want to skip the default implementation you can just return true from your custom handler and implement your  logic for that Instr there.

If you need to do some extra work before the default work, you can just return false for skip.

The way it is listed here it seems that you only do PROCESSING if your preprocessing returns false. Seems a bit redundant. WDYT?

opti-mix: Yeah, I actually thought about it when I was defining processing stages.

But I guess my idea to explicitly distinguish between preprocessing, processing and post-processing comes from some past experiences with AOP (Aspect-oriented programming), where you explicitly have before/around/after advices.

Also, if we would allow for adding multiple hooks to the same backend, then you can get into a situation where you invoke a chain of hooks for the same instruction and the same processing stage and in that case you probably cannot always combine preprocessing with processing stages unless you are the last element of this chain, because the work done by these stages of the same hook would be separate in time (e.g. pre1 -> pre2 -> pre3 -> proc1 -> ...). Does it make sense?

jfix71: I'm a little confused here -- looks like if the handler returns true for PREPROCESSING here, then we skip PROCESSING entirely below? It seems like since there's no pre or post processing in the Interpreter currently then we would only need to check the handler's return from PROCESSING, to see if we skip the Interpreter's default processing.

include/glow/IR/IR.h
```diff
@@ -225,6 +225,60 @@ class Instruction : public Value, public TaggedListNode<Instruction> {
...
+/// direction.
+enum class IRInstructionProcessingStage {
+  /// Instruction is going to be processed.
+  PREPROCESSING,
+  /// Instruction is being processed.
+  PROCESSING,
+  /// Instruction was just processed.
+  POSTPROCESSING,
+  /// Instruction processing is done.
+  DONE,
```
jfix71: Do we still need DONE here since the caller manages the stage? I.e. the handler doesn't return a stage so not sure this would ever be used?

lib/Backends/Interpreter/InterpreterFunction.cpp
jfix71: RE: multiple hooks, the current design doesn't allow for this, right? IMO it makes more sense to leave it without the extra stages since they aren't currently useful, and in the future we can add them + multiple hooks etc. as needed. Especially because the current design using an enum to pass the current stage is easily extended to handle these cases.

opti-mix: @jfix71 

> It seems like since there's no pre or post processing in the Interpreter currently then we would only need to check the handler's return from PROCESSING, to see if we skip the Interpreter's default processing.

Oh, yes. I see your point. Will change.

opti-mix: > IMO it makes more sense to leave it without the extra stages since they aren't currently useful, and in the future we can add them + multiple hooks etc. as needed. Especially because the current design using an enum to pass the current stage is easily extended to handle these cases.

So, I removed DONE. Do you propose to remove PREPROCESSSING as well like Sameer has proposed ?

lib/Backends/Interpreter/InterpreterFunction.cpp
```diff
@@ -229,21 +231,55 @@ Error BoundInterpreterFunction::execute(IRFunction *F,
...
+    fwd##CLASS(llvm::cast<CLASS>(&I));                                         \
+    break;                                                                     \
+  }
+#define DEF_BACKEND_SPECIFIC_INSTR(CLASS, NAME)
+#include "glow/AutoGenInstr.def"
+
+        default:
+          glow::errs() << "Invalid instruction: " << &I << "\n";
+          llvm_unreachable("Invalid instruction.");
+        }
```
jfix71: nit: This logic is shared with the version without hooks. Could outline to a helper.

lib/Backends/Interpreter/InterpreterFunction.cpp
jfix71: I probably would remove it since there's no clear use for it as currently formulated, but no strong feelings.

opti-mix: OK, guys, you convinced me! :-) I removed PREPROCESSING.

Issue #4140, opened 2020-02-10
[torch_glow] Allow setting backend specific options from python


mneilly-et:
Host manager provides the --load-backend-specific-options command line argument which is used to provide backend specific settings via a YAML file.

AFAICT torch glow does not currently provide a way to pass backend specific options. It would be useful if the YAML file could be loaded from torch glow as well.

PR #4141, opened 2020-02-10
[torch_glow] Add ability to pass backend specific options

mneilly-et:
Add ability to pass backend specific options to backends from torch_glow

Summary:

Host manager already provides the --load-backend-specific-options command
line option which loads a YAML file containing backend options. When using
torch_glow, it is currently not possible ot pass this option.

This change adds torch_glow.loadBackendSpecificOptions() which takes the name
of a YAML file. CachingGraphRunner will pass the name of the YAML file to the
host manager via cctx.backendOpts.backendSpecificOpts and host manager will
load the options into the context which is already passed to backends.

Fixes:  #4140

### DIFF
--- a/include/glow/Backends/Interpreter/Interpreter.h
+++ b/include/glow/Backends/Interpreter/Interpreter.h
@@ -58,6 +58,8 @@ public:
 
   bool shouldLower(const Node *N) const override;
 
+  void parseBackendSpecificOptions(const BackendOptions &opts) const;
+
   /// @}
   //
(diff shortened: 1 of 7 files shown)


mneilly-et:
I was going to bring this current with master but saw the "@jackm321 has imported this pull request." message and wasn't sure at that point. Should I go ahead and update the PR?

jackm321:
@mneilly-et, I can rebase it

PR #4142, opened 2020-02-10
Only build network-debugger if GLOW_BUILD_TESTS is ON

### LINKED ISSUE #4135: [Glow build] Error when building with "-DGLOW_BUILD_TESTS=OFF"
@SameerAsal After the commit #4053 the Glow build fails when disabling the tests via the CMake option "-DGLOW_BUILD_TESTS=OFF":
- When the tests are disabled the "BackendTestUtils" library is not build. But this is a dependency for the "network-debugger" application which fails to link.

Therefore you should put a condition in the CMake to build the "network-debugger" only when the tests are also build (or remove the "BackendTestUtils" dependency entirely).

SameerAsal:
Remove unneeded dependencies network-debugger had.
  No longer needs Exporter to build.

Summary:

Documentation:
  None needed.

This fixes #4135

Test Plan:
  Built with  -DGLOW_BUILD_TESTS=OFF

### DIFF
--- a/tools/Debugger/CMakeLists.txt
+++ b/tools/Debugger/CMakeLists.txt
@@ -1,3 +1,4 @@
+if (GLOW_BUILD_TESTS)
 add_executable(network-debugger
                network-debugger.cpp
                NetworkComparator.cpp)
@@ -5,8 +6,7 @@ add_executable(network-debugger
 target_link_libraries(network-debugger
                       PRIVATE
                         BackendTestUtils
-                        HostManager
                         Importer
                         HostManager
-                        ExecutionEngine
-                        Exporter)
+                        ExecutionEngine)
+endif()


tools/Debugger/CMakeLists.txt
```diff
@@ -4,9 +4,6 @@ add_executable(network-debugger
 
 target_link_libraries(network-debugger
                       PRIVATE
-                        BackendTestUtils
-                        HostManager
                         Importer
                         HostManager
-                        ExecutionEngine
-                        Exporter)
+                        ExecutionEngine)
```
jfix71: nit: new line at end of file

SameerAsal:
I actually now remember why I added these dependencies, it was because of the shared build needed them. 

I might just not build network debugger if TESTS are off like the issue suggests.

PR #4147, opened 2020-02-11
NNPI: get rid of string lookups in InferenceThreadEnv::execute

tracelogfb:
Summary: From profiling, building up a hash map of tensor name to tensor pointer on every inference and doing all the lookups are slowing down the inference. Replace the unordered_map with vectors of placeholder pointers that match the order of host/net input/output lists. From testing some workloads, it improves throughput by up to 5% by allowing better overlap of copy and inference on the card.

Differential Revision: D19829996

### DIFF
--- a/lib/Backends/NNPI/InferencePool.cpp
+++ b/lib/Backends/NNPI/InferencePool.cpp
@@ -144,24 +144,81 @@ void InferenceThreadEnv::execute(RunIdentifierTy runId,
 
   // Pre inference input preparation.
   PlaceholderBindings &bindings = *ctx->getPlaceholderBindings();
-  ioTensors_.clear();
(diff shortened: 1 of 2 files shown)


tracelogfb:
cc: @jsubag

jsubag:
This optimization looks good (we saw similar overheads in our analysis).
However, this code was already refactored with a similar idea in mind and will be part of the next update. (In a nutshell the thread env class was replace with a context that should better align the Glow context and will also map placeholders to inputs/outputs during initialization).

Up to you if you prefer to push this now and we'll update on the next drop or wait for it.

tracelogfb:
Cool. I'll push this for now and we can replace it whole sale in the next NNPI drop.

Issue #4150, opened 2020-02-11
[torch_glow] PyTorchModelLoader is missing negative dimension support


mneilly-et:
I was unable to run the [Bert model test](https://github.com/huggingface/transformers/blob/master/tests/test_modeling_bert.py) from the huggingface/transformers repo as a result of missing negative dimension support in torch_glow's PyTorchModelLoader. 

When negative dimensions are encountered glow crashes python with the following error:

```
double free or corruption (out)                                                                                                                                                                
Fatal Python error: Aborted                                                                                                                                                                    
                                                                                                                                                                                               
Current thread 0x00007f6a12755740 (most recent call first):                                                                                                                                    
  File "/home/mneilly/.virtualenvs/mneilly-et/lib/python3.7/site-packages/torch/jit/__init__.py", line 644 in run_mod_and_filter_tensor_outputs                                                  File "/home/mneilly/.virtualenvs/mneilly-et/lib/python3.7/site-packages/torch/jit/__init__.py", line 682 in _check_trace                                                                     
  File "/home/mneilly/.virtualenvs/mneilly-et/lib/python3.7/site-packages/torch/autograd/grad_mode.py", line 15 in decorate_context                                                            
  File "/home/mneilly/.virtualenvs/mneilly-et/lib/python3.7/site-packages/torch/jit/__init__.py", line 1052 in trace_module                                                                    
  File "/home/mneilly/.virtualenvs/mneilly-et/lib/python3.7/site-packages/torch/jit/__init__.py", line 890 in trace                                                                            
  File "/home/mneilly/OSS/machine_learning/transformers/transformers/tests/modeling_common_test.py", line 232 in _create_and_check_torchscript                                                 
  File "/home/mneilly/OSS/machine_learning/transformers/transformers/tests/modeling_common_test.py", line 205 in test_torchscript                                                              
  File "/usr/lib/python3.7/unittest/case.py", line 615 in run 
```

The following examples of negative dimension use can be found in [modeling_bert.py](https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_bert.py#L227):

```
        # Take the dot product between "query" and "key" to get the raw attention scores.
        attention_scores = torch.matmul(query_layer, key_layer.transpose(-1, -2))
...
        # Normalize the attention scores to probabilities.
        attention_probs = nn.Softmax(dim=-1)(attention_scores)
```

PR #4151, opened 2020-02-11
[torch_glow] Add support for negative dimensions in PyTorchLoader

mneilly-et:
Add support for negative dimension indices.
    
Summary:
    
Add support for negative dimension indices for loadSize(), loadFusedConcat(), loadSoftmax() and loadTranspose(). When a dimension is negative, PyTorchModelLoader will convert the index to a positive index and check that it is in the valid range.
    
These changes enable running the huggingface/transformers Bert, Albert and Distilbert models.

Fixes:  #4150

TestPlan:

Added negative dimension tests to cat_test.py, transpose_test.py, size_test.py and softmax_test.py

### DIFF
--- a/torch_glow/tests/nodes/size_test.py
+++ b/torch_glow/tests/nodes/size_test.py
@@ -19,3 +19,25 @@ class TestSize(unittest.TestCase):
         x = torch.zeros([4], dtype=torch.int32)
 
         jitVsGlow(test_f, x, expected_fused_ops={"aten::size"})
+
+    @unittest.skip(reason="not ready")
+    def test_size_neg_dim(self):
+        """Test negative dimension index for the PyTorch aten::size Node on Glow."""
+
+        def test_f(a):
+            return a.size(-1)
+
+        x = torch.randn(2, 3, 4, dtype=torch.float32)
+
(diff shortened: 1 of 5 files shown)


torch_glow/tests/nodes/size_test.py
```diff
@@ -19,3 +19,25 @@ def test_f(a):
         x = torch.zeros([4], dtype=torch.int32)
 
         jitVsGlow(test_f, x, expected_fused_ops={"aten::size"})
+
+    @unittest.skip(reason="not ready")
```
yinghai: What's the issue with this?

mneilly-et: @yinghai, the existing test_size_basic was marked as skip "not ready" and when enabled it fails with COMPILE_UNSUPPORTED_NODE_AFTER_OPTIMIZE. When running test_size_neg_dim, there was an "AssertionError: Expected aten::size to be fused" regardless of the negative index which appeared to me to be related to the preceding error and not related to the negative index so I marked the new test as skip "not ready" as well. The test_size_oob_neg_dim test can run because the index error is raised by pytorch before reaching glow.

PR #4162, opened 2020-02-13
[NNPI] Move parallelization logic to open source

jfix71:
Summary: Note that setting `GlowDisableNNPIPrivateTransforms` to false will no longer disable parallelization, as it's no longer private. If desired we can add a flag to disable parallelization specifically.

Differential Revision: D19867766


jfix71:
CC: @jsubag

PR #4166, opened 2020-02-13
Adding cacheable networkExecutionState to Executor

gcatron:
Summary:
This adds a NetworkExecutionState class to the executor, it is similar to the ExecutionState object except that it is initialized once and re-used for multiple executions.
This will reduce the overhead in the Executor and allows for further optimizations such as pinned memory in contexts, and P2P support. 
Documentation:

Test Plan:
Ensure unit tests pass

### DIFF
--- a/include/glow/Runtime/Executor/NetworkExecutionState.h
+++ b/include/glow/Runtime/Executor/NetworkExecutionState.h
@@ -1,0 +1,158 @@
+/**
+ * Copyright (c) Glow Contributors. See CONTRIBUTORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef GLOW_RUNTIME_EXECUTOR_NETWORKEXECUTIONSTATE_H
+#define GLOW_RUNTIME_EXECUTOR_NETWORKEXECUTIONSTATE_H
+
+#include "glow/Runtime/RuntimeTypes.h"
+#include "glow/Support/TensorPool.h"
+#include "glow/Support/ThreadPool.h"
+
+#include <mutex>
+
+namespace glow {
+namespace runtime {
+
+/// This class keeps track of the state of execution for a run (identified
+/// by the runId).
+class NetworkExecutionState final {
+public:
+  /// Constructor.
+  explicit NetworkExecutionState(const DAGNode *root);
+
+  /// Destructor.
+  ~NetworkExecutionState();
+
+  /// Does the BFS traversal and initializes the NetworkExecutionState.
+  void init(const DeviceManagerMapTy &devices);
+
+  /// Binds the state to a new run. This moves the result ctx and cb to be owned
+  /// by the networkExecutionState for the duration of the run.
+  void bind(std::unique_ptr<ExecutionContext> resultCtx, ResultCBTy cb,
+            RunIdentifierTy runId);
+
+  /// \returns a unique pointer to an input bindings for \p node. This should
+  /// not be called at the same time as insertIntoNodeCtx().
+  std::unique_ptr<ExecutionContext>
+  getUniqueNodeContextPtr(const DAGNode *node);
+
+  /// Returns the intermediateContext back to the networkExecutionState after
+  /// completion so it can be re-used.
+  void returnUniqueNodeContextPtr(const DAGNode *node,
+                                  std::unique_ptr<ExecutionContext> ctx);
+
+  /// Increment the count of inflight nodes by \p increment (default is 1).
+  void incrementInflightNodes(unsigned increment = 1);
+
+  /// Decrement the count of inflight nodes by the \p decrement (default is 1).
+  /// \returns true if there are no nodes inflight after the decrement
+  /// operation.
+  bool decrementInflightNodes(unsigned decrement = 1);
+
+  /// Increment the count of completed parent nodes for \p node. \returns
+  /// true if all parents are done after the increment operation, false
+  /// otherwise.
+  bool incrementNodeParentsDone(const DAGNode *node, unsigned increment = 1);
+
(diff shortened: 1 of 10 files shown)


gcatron:
@jsubag

tracelogfb:
How does synchronization work on the cached state? It seems like we just round robin across object in the pool and wrap around. What's preventing from a thread from using an object that's still in use by another one?

It also seems like it's not a good idea for the ThreadPoolExecutor to block, right? That could easily lead to a dead lock situation if we ware not careful.

yinghai:
Should we review this or this is still wip?

gcatron:
> Should we review this or this is still wip?

Still WIP for now, I need to stitch in inputs/outputs from the request context. Also since we're going that route I don't think there's a need to keep executionState and NetworkExecutionState around so I'll fold them in. Should be ready later today.

tests/unittests/ThreadPoolExecutorTest.cpp
```diff
@@ -696,7 +698,7 @@ TEST_F(ThreadPoolExecutorTest, ConcurrentSingleNode) {
   constexpr RunIdentifierTy baseTestRunId = 10;
   constexpr DeviceIDTy testDeviceId = 111;
   constexpr unsigned deviceManagerThreads = 3;
-  unsigned numConcurrentRuns = 1000;
+  unsigned numConcurrentRuns = 10;
```
yinghai: Why this change?

include/glow/Runtime/Executor/NetworkExecutionState.h
```diff
@@ -0,0 +1,156 @@
...
+  bool initialized_{false};
+
+private:
+  /// The run identifier for this execution of a DAG.
+  RunIdentifierTy runId_;
+  /// The callback that should be called when execution is done.
+  ResultCBTy cb_;
+
+  /// Map of all buffers allocated for intermediate contexts.
+  std::unordered_map<Placeholder *, void *> buffers_;
```
yinghai: Can we group these members by things that are static (things we can determine when we create the state pool except for the tensors that we need to bind), and things are dynamic (mainly bookkeeping of the dag job status).

include/glow/Runtime/Executor/NetworkExecutionState.h
```diff
@@ -0,0 +1,156 @@
...
+  /// Thread Executor used for this run.
+  // ThreadExecutor *executor_;
+};
+
+class NetworkExecutionStatePool {
+public:
+  NetworkExecutionState *getNextNetworkExecutionState() {
+    unsigned current = currentState_;
+    currentState_ = (currentState_ + 1) % states_.size();
+    return states_[current].get();
```
yinghai: This is not thread safe. Two threads can read the same currentState_. Even with currentState_.fetch_and_add() it is still not safe. Here you can do what we did for the NNPI EnvPool, using a stack guarded by mutex. Or since we have access to boost now, we can use boost::lock_free::stack<NetExecutionState>.

lib/Runtime/HostManager/HostManager.cpp
```diff
@@ -228,6 +228,11 @@ Error HostManager::addNetwork(std::unique_ptr<Module> module,
     return err;
   }
 
+  // Create pool of cachedExecutionStates.
+  for (auto &node : nodeList) {
```
yinghai: This potentially wastes quite a bit memory. Let's keep it this way and see whether the trade-off is worth it.

yinghai: Actually do we need to set default as number of inflight requests? I'd think it should be number of workers that will feed up the backend? Say 2?

include/glow/Runtime/Executor/NetworkExecutionState.h
gcatron: I'm hoping to avoid a mutex since this is in the inference path, If I use fetch_add(), and move the modulus to line 142, It should be thread safe, states_.size() should not change after initialization. I can refactor create pool to make that more explicit.

yinghai: Modulus will wrap around.

gcatron: That's fine, since the size of the pool == maxActiveRequests. there won't be overlap.

yinghai: It's not safe. As I mentioned, having pool size equal to maxActiveRequests may not be a good idea. And relying on this implicitly may potentially lead to hard to debug issues. Mutex is actually not that bad and we have access to lockfree stack also. Such scheme has been proven to be efficient in large scale services.

tests/unittests/ThreadPoolExecutorTest.cpp
gcatron: set it to 10 for testing, changing it to 100. 1000 doesn't add much and really slows down testing particularly in debug.

yinghai: Cool. We do want the system to be over subscribed to flush out some race issue. 100 seems to be good.

lib/Runtime/HostManager/HostManager.cpp
gcatron: I think it needs to be maxActiveRequests for now. Since executor->run() modifies the networkExecutionState. say maxActive = 10 and workers = 2. If you have 10 requests, the first 2 would be enqueued and acted on immediately so we need at least 2. The next 8 would be dispatched to the executor, which would put them on the queue for the Device. Those 8 would need their networkExecutionState objects by then too.

yinghai: Yeah, it just doesn't necessarily need to be maxActiveRequests. One reason is that device resident tensors are going to consume resource in the cards. The other reason is that we can actually control the use of resource just by block waiting for the item to be available in the pool. Right now, let's put maxActiveRequests as an estimate. But we need to be aware that this number can change and system should still work.

lib/Runtime/Executor/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_library(Executor
               ExecutionState.cpp
```
yinghai: Where do we use ExecuteState anymore?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -207,6 +194,8 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
       }
     }
   }
+  // Return intermediateContext to executionState.
+  executionState->returnUniqueNodeContextPtr(node, std::move(ctx));
```
yinghai: I found this context check in and check out a bit awkward. Maybe if we use raw point in the lower level calls can improve the situation. You don't need to change this now.

PR #4167, opened 2020-02-13
Update NNPI Backend to v0.5.0.8

omromano:
Summary: Update of NNPI Backend to v0.5.0.8 branched out of commit 24df161c4d0eb716473eac43673b423777a9ab48.


jsubag:
@arunm-git @jfix71 Please take a look

lib/Backends/NNPI/InferenceContext.cpp
```diff
@@ -0,0 +1,433 @@
...
+  }
+  TRACE_EVENT_BEGIN(ctx->getTraceContext(), TraceLevel::COPY,
+                    TRACING_PRE_PROCESS);
+
+  // Pre-inference
+  auto &phTensorMap = bindings.pairs();
+  std::vector<void *> rawInputs, rawOutputs;
+  for (auto &in : inputResources_) {
+    if (in->GetUsage() != NNPIResource::ResourceUsage::StaticInputResource) {
+      Placeholder *ph = bindings.getPlaceholderByName(in->GetName());
```
yinghai: This seems to be a regression from recent positional bindings as here we are doing string lookup per inference per input again.

jsubag: Yes, we can apply a similar change as in #4147 here.

jsubag:
The refactor adds no new features.
It's a cleaner design and a step towards the context binding mechanism. In also gives some perf benefit in the pre/post processing before/after inferences.

lib/Backends/NNPI/InferenceContext.cpp
yinghai: Should we do it in this PR or a follow-up?

jsubag: Will try to do that on Sunday - otherwise can do it later in the week.

yinghai:
@jsubag I'm sendng https://github.com/pytorch/glow/pull/4194 for the follow-up.

PR #4168, opened 2020-02-14
Remove COVERAGE build

bertmaher:
Summary: This build has broken a couple times in the past week.  As far as I can tell no one really seems to get a lot of value out of it.  While I don't really like to decrease test coverage, those build breaks weren't actually indicative of a problem, they were just random things the compiler didn't support in that mode.


bertmaher:
Note I didn't remove the infra for supporting the coverage build.  So maybe this could come back some day, or maybe we can DCE the rest if it's never resurrected.

PR #4189, opened 2020-02-17
[OpenCL] Fuse convolution with ReLU

pjaaskel:
* Intel SDK (Intel i7-7500U CPU)

inception_v1: From 0.90 to 0.71 (22% drop)
mobilenetv2:  From 0.40 to 0.16 (60% drop)
resnet50:     From 1.41 to 1.25 (11% drop)

* Intel GPU (HD620)

inception_v1: From 0.14 to 0.14 (0% drop)
mobilenetv2:  From 0.21 to 0.20 (5% drop)
resnet50:     From 0.29 to 0.29 (0% drop)

* pocl / LLVM 9 (Intel i7-7500U CPU)

inception_v1: From 0.80 s to 0.52 s (35% drop)
mobilenetv2:  From 0.23 s to 0.15 s (35% drop)
resnet50:     From 1.97 s to 1.25 s (36% drop)

Test Plan:

Tested&benchmarked with the above OpenCL envs and nets.


lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3832,6 +3832,12 @@ Error glow::optimizeFunction(Function *F, const Backend &B,
   }
 
   RETURN_IF_ERR(optimizeFunctionBeforeLowering(F, cctx));
+
+  // Fold before lowering to enable cases which would not fold after
+  // lowering, particularly convolution+relu since relu will be lowered
+  // to max(0, x).
+  ::glow::fold(F, cctx, &B);
```
opti-mix: Isn't `glow::fold` called from `optimizeFunctionBeforeLowering`, which is invoked in line 3834? The only difference I see is that you pass the backend `B` as a parameter now. So, maybe `optimizeFunctionBeforeLowering` should take a backend as a parameter and pass it down to `fold` later?

lib/Backends/OpenCL/kernels.cl
```diff
@@ -1149,7 +1167,20 @@ convolution_i8W(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
   convolution_i8K(&mem[dest], &mem[src], &mem[filter], &mem[bias], kernelSizes,
                   strides, destOffset, destScale, srcOffset, srcScale,
                   filterOffset, filterScale, biasOffset, biasScale, pads, group,
-                  dilation, odim, idim, filterDim);
+                  dilation, odim, idim, filterDim, 0);
```
jfix71: nit: `/* fuseReLU */ 0`

lib/Backends/OpenCL/kernels.cl
```diff
@@ -1149,7 +1167,20 @@ convolution_i8W(__global void *mem, cl_uint32_t dest, cl_uint32_t src,
...
+    __global void *mem, cl_uint32_t dest, cl_uint32_t src, cl_uint32_t filter,
+    cl_uint32_t bias, ShapeHW kernelSizes, ShapeHW strides, PaddingTLBR pads,
+    cl_uint32_t group, cl_uint32_t dilation, ShapeNHWC odim, ShapeNHWC idim,
+    ShapeNHWC filterDim, cl_int32_t destOffset, float destScale,
+    cl_int32_t srcOffset, float srcScale, cl_int32_t filterOffset,
+    float filterScale, cl_int32_t biasOffset, float biasScale) {
+  convolution_i8K(&mem[dest], &mem[src], &mem[filter], &mem[bias], kernelSizes,
+                  strides, destOffset, destScale, srcOffset, srcScale,
+                  filterOffset, filterScale, biasOffset, biasScale, pads, group,
+                  dilation, odim, idim, filterDim, 1);
```
jfix71: nit: `/* fuseReLU */ 1`

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
jfix71: Ah, so I actually made that change in https://github.com/pytorch/glow/pull/3867 but I guess it got lost in the holiday shuffle and never landed 🙁 

Thinking more about it though, I don't think we actually can fuse until after `transformForPrecisionMode()` -- otherwise the graph may look differently and so the quantization profile may no longer fit the graph.

I think the right thing to do here is actually move the call to `foldActivations()` out of `fold()` to right after the call to `transformForPrecisionMode()`

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -586,6 +586,12 @@ void OpenCLFunction::executeNCHWConvolution(
   addStringOption(options, "v_pad_A", "0");
   addStringOption(options, "v_pad_B", "0");
 
+  if (CC->getFusedActivation() == FusedActivation::RELU) {
+    addIntOption(options, "v_fuse_relu", 1);
```
jfix71: nit: `/* fuseReLU */ 1`

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -586,6 +586,12 @@ void OpenCLFunction::executeNCHWConvolution(
...
+  } else {
+    addIntOption(options, "v_fuse_relu", 0);
```
jfix71: nit: `/* fuseReLU */ 0`

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: This doesn't seem to be consistent.

lib/Backends/OpenCL/OpenCL.cpp
pjaaskel: Nor this. Note that these are specialization macro parameters for the other convolution.

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
pjaaskel: Done. To avoid the relu to max lowering that hinders convolution+relu conversions happening later, I forbid relu to max lowering in the OpenCL backend and just implement RelU in the backend (using max). This should be beneficial also due to the constant scalar 0 input to the max (no need for splatted matrix just for this).

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -3850,6 +3851,11 @@ Error glow::optimizeFunction(Function *F, const Backend &B,
   // instrument with profiling nodes. This must be done after lowering.
   transformForPrecisionMode(B, F, cctx);
 
+  // Fold activations before lowering to enable cases which would not fuse after
+  // lowering. This cocerns particularly convolution&relu since relu will be
```
jfix71: typo: "cocerns"

pjaaskel:
I fixed the typo, but I (still) couldn't reproduce those failures. They seem relevant though, as they concern ReLU. I'll try a clean build.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -854,6 +860,16 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
                 quantization::quantizeScaleOffset32To8(resultScale, 0);
             setKernelArg(kernel, ++numArgs, resultScaleParams);
           }
+        } else if (isa<ReluInst>(I)) {
+          int32_t destOffset = I.getOperand(0).first->getType()->getOffset();
+          float destScale = I.getOperand(0).first->getType()->getScale();
+
+          auto SRCTy = I.getOperand(1).first->getType();
```
opti-mix: SRCTy -> srcTy

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -854,6 +860,16 @@ Error OpenCLFunction::execute(ExecutionContext *context) {
                 quantization::quantizeScaleOffset32To8(resultScale, 0);
             setKernelArg(kernel, ++numArgs, resultScaleParams);
           }
+        } else if (isa<ReluInst>(I)) {
```
opti-mix: Please use `if (auto *RI = dyn_cast<ReluInst>(I)) {...}` and use ReluInst specific getters like getDest(), etc.

lib/Backends/OpenCL/OpenCL.cpp
```diff
@@ -1987,7 +2011,10 @@ bool OCLBackend::verify(const Function &F, bool verbose) const {
 
 bool OCLBackend::verify(const IRFunction &IR) const {
   for (const auto &I : IR.getInstrs()) {
-    if (!checkNoFusionForInstr(I)) {
+    if (!(I.getKind() == Kinded::Kind::ConvolutionInstKind &&
```
opti-mix: Please add a comment.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -246,6 +246,8 @@ class OCLBackend final : public BackendUsingGlowIR {
     // The group convolution is supported in OpenCL slow convolution kernel.
     if (N->getKind() == Kinded::Kind::ConvolutionNodeKind)
       return false;
+    if (N->getKind() == Kinded::Kind::ReluNodeKind)
```
opti-mix: Please add a comment explaining why you don't want to lower ReLUs.

opti-mix: And BTW, we usually use {} even for a single line if/else/for/while bodies. I see that the line above was not following this style ;-) Please change it as well.

lib/Backends/OpenCL/OpenCL.h
```diff
@@ -255,6 +257,13 @@ class OCLBackend final : public BackendUsingGlowIR {
   runtime::DeviceManager *
   createDeviceManager(const runtime::DeviceConfig &deviceConfig) override;
 
+  /// \returns whether the backend supports fusing \p activation into \p parent.
+  bool supportsFusedActivation(Node *parent, Node *activation) const override {
+    bool V = parent->getKind() == Kinded::Kind::ConvolutionNodeKind &&
```
opti-mix: Please add a comment.

pjaaskel:
> I fixed the typo, but I (still) couldn't reproduce those failures. They seem relevant though, as they concern ReLU. I'll try a clean build.

_ninja clean && ninja && ninja check_  passes here... strange.

opti-mix:
@pjaaskel The CI error is:
```
./glow/AutoGenInstr.h:2607: void glow::ReluInst::verify() const: Assertion `getDest()->getType() == getSrc()->getType() && "Invalid Type"' failed.
```

It seems like there is a ReluInst with incorrect operand types being produced.

pjaaskel:
OK, it finally reproduced locally after a complete Glow rebuild from the scratch.

The ReLU in question here:
```
7 %fc_bias_res = allocactivation  { Ty: i8[S:0.1043 O:-104][-2.504,24.101]<2 x 16>} // size: 32 // Users: @out 19, @in 10, @out 8
... 
9 %relu_res = allocactivation  { Ty: i8[S:0.0945 O:-128][0.000,24.109]<2 x 16>} // size: 32 // Users: @out 20, @in 12, @out 10 
10 %relu = relu @out %relu_res, @in %fc_bias_res
```

Thus, the inputs/outputs have different quantization parameters. Perhaps this happens because previously it has been lowered to max which does not check for the type. Just guessing, as I didn't look deeper yet (I have to get back to this tomorrow), perhaps you have educated hints on how to fix this?

opti-mix:
@pjaaskel #4205 should help you once it lands and you rebase on top of it.

opti-mix:
@pjaaskel OK, all tests pass now after #4205 had landed.

@jfix71 Any further comments? Are you ok with the introduced foldActivations() call?

PR #4191, opened 2020-02-19
Start/stop device tracing interface.

omromano:
Summary:
Define an interface to start/stop device tracing.
Using this interface we will be able to capture device traces for end 2 end glow application execution when only one device tracing context is allowed.
Tracing context is owned by the HostManager and can be merge into a global trace context when the execution ends.
For an example of how to use, see the tools/loader/ImageClassifier.cpp suggested changes.

### DIFF
--- a/include/glow/Backends/DeviceManager.h
+++ b/include/glow/Backends/DeviceManager.h
@@ -229,6 +229,15 @@ public:
     DCHECK("Not Implemented");
     return false;
   }
+
+  /// Starts device tracing \returns Error if fails.
+  virtual Error startDeviceTrace(TraceContext *traceContext) {
+    return Error::success();
+  }
+  /// Stops device tracing \returns Error if fails.
+  virtual Error stopDeviceTrace(TraceContext *traceContext) {
+    return Error::success();
+  }
 };
 
 } // namespace runtime
--- a/include/glow/Runtime/HostManager/HostManager.h
+++ b/include/glow/Runtime/HostManager/HostManager.h
@@ -105,6 +105,8 @@ class HostManager final {
   /// Configuration parameters for this Runtime Host.
   const HostConfig config_{};
 
+  std::unique_ptr<TraceContext> hostTraceContext_;
+
   /// A map from a networkName to a network, which is represented by struct DAG.
   std::unordered_map<std::string, NetworkData> networks_;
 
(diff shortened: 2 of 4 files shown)


omromano:
@jfix71 Please take a look. 
CC: @jsubag

jfix71:
CC: @nickgg @tracelogfb

tracelogfb:
This makes sense to me in isolation. It doesn't play super well the glow tracing's default model of turning tracing on/off on a per request level. It's more along the line of, we want to turn tracing on for some amount of time (say 1 second) and everything happening within that window should get traced. 

I think that's a generally more useful mode of tracing. So in glow, we can do a little bit of work to make that mode a first class citizen. In that context, the start/stop tracing at the host manager / device manager looks good to me.

tracelogfb:
@omromano btw, does turning device tracing on/off require root privilege? If not, great. If so, it's a hard requirement for us to be able to do it without the application running as root.

tracelogfb:
Also to clarify, by 

> Using this interface we will be able to capture device traces for end 2 end glow application execution when only one device tracing context is allowed.

you mean that every inference happening during the start/stop on the device will be included in the trace context, right?

omromano:
> @omromano btw, does turning device tracing on/off require root privilege? If not, great. If so, it's a hard requirement for us to be able to do it without the application running as root.
Checking. 

> you mean that every inference happening during the start/stop on the device will be included in the trace context, right?
Yes.

include/glow/Runtime/HostManager/HostManager.h
```diff
@@ -228,6 +230,23 @@ class HostManager final {
...
+  TraceContext *getTraceContext() { return hostTraceContext_.get(); }
+
+  /// Sets the TraceContext and \returns the existing value.
+  std::unique_ptr<TraceContext>
+  setTraceContext(std::unique_ptr<TraceContext> traceContext) {
+    std::swap(hostTraceContext_, traceContext);
+    return traceContext;
+  }
+
+  /// Triggers start tracing of all active devices \returns Error if fails.
```
tracelogfb: The comment here says this will return an error if it fails, but it's returning void. I guess we should make it return a glow error or something.

include/glow/Backends/DeviceManager.h
```diff
@@ -222,6 +222,9 @@ class DeviceManager : public DeviceTensorTransferManager {
     DCHECK("Not Implemented");
     return false;
   }
+
+  virtual void startDeviceTrace(TraceContext *traceContext) {}
```
jfix71: Actually can you also add doxygen comments here too? @omromano

include/glow/Backends/DeviceManager.h
```diff
@@ -222,6 +222,9 @@ class DeviceManager : public DeviceTensorTransferManager {
...
+  virtual void stopDeviceTrace(TraceContext *traceContext) {}
```
jfix71: (And here)

omromano:
Hi @jfix71, @tracelogfb ,  Is this PR still planed to be merged? 
Thanks.
CC: @jsubag

yinghai:
@omromano Yes. Could you address the comments above? Thanks.

omromano:
@yinghai, Sure, will do it.

PR #4194, opened 2020-02-19
Put back the positional placeholder optimization that got lost in 0.5.0.8 upgrade

yinghai:
Summary: Basically put back D19829996.

Differential Revision: D19967323


yinghai:
@jsubag FYI

PR #4198, opened 2020-02-20
[Importer] Use Expected when loading vectors

jfix71:
See title. Not sure of a better way to check if the vector exists other than checking the size...?

### DIFF
--- a/include/glow/Importer/ProtobufLoader.h
+++ b/include/glow/Importer/ProtobufLoader.h
@@ -82,6 +82,7 @@ template <typename T> static Expected<std::string> loadStr(const T *arg) {
 template <typename ElemTy, typename AttrType>
 Expected<std::vector<ElemTy>> getShape(const AttrType *arg,
                                        bool allowEmptyShape = false) {
+  RETURN_ERR_IF_NOT(arg, "Node has no ints attribute with this name");
   if (!allowEmptyShape) {
     RETURN_ERR_IF_NOT(arg->ints_size() > 0, "Node has no ints values");
   }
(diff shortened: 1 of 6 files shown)


include/glow/Importer/ProtobufLoader.h
```diff
@@ -76,19 +76,23 @@ template <typename T> static Expected<std::string> loadStr(const T *arg) {
...
 }
 
-/// Load a floating record vector from \p arg. \returns a standard vector of
-/// floats.
-template <typename AttrType> std::vector<float> getFloats(const AttrType *arg) {
+/// Load a floating record vector from \p arg into a vector. \returns the vector
+/// if there is no error, or an error otherwise.
+template <typename AttrType>
+Expected<std::vector<float>> getFloats(const AttrType *arg) {
+  // RETURN_ERR_IF_NOT(arg->floats_size() > 0, "Node has no floats values");
```
jackm321: why is this commented out?

lib/Importer/Caffe2ModelLoader.cpp
```diff
@@ -1669,7 +1675,8 @@ Error Caffe2ModelLoader::loadWeight(const caffe2::OperatorDef &op) {
      *   }
      * }
      */
-    auto dim = getShape(dict["shape"]);
+    std::vector<dim_t> dim;
+    ASSIGN_VALUE_OR_RETURN_ERR(dim, getShape<dim_t>(dict.at("shape")));
```
jackm321: Not sure what's going on with this, it looks like it should be the same?

PR #4199, opened 2020-02-20
Add an option to set output trace file path.

omromano:
Summary: Add an option in the Repro tool to predefine the output trace file.


omromano:
Hi @jfix71, @tracelogfb 
Please take a look.
CC: @jsubag

tracelogfb:
LGTM, but I'll leave it for @jfix71 to merge it.

PR #4206, opened 2020-02-21
[Folly][VisualStudio]Getting GLOW to compile with Folly with Visual studio.

ayermolo:
Summary:
Some changes and documentation to get Folly to work with Visual Studio.
Kind of gave up trying to build Folly in glow from source. Changed it to just find a package and build is using vcpkg.
Right now there is an open issue in vcpkg that needs to be fixed. Opened and issue and proposed solution.
https://github.com/microsoft/vcpkg/issues/10169
Documentation:

[Optional Fixes #issue]

Test Plan:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,8 +185,11 @@ if(NOT EXISTS "${GLOW_THIRDPARTY_DIR}/fp16")
 else()
   include_directories(${GLOW_THIRDPARTY_DIR}/fp16/include)
 endif()
-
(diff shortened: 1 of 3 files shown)


Orvid:
Out of curiosity, when was the attempt at building Folly from source? Folly was broken with newer versions of the windows SDK for a couple of weeks, but it should be fixed now.

ayermolo:
@Orvid Last week. With latest Folly clone. Probably something in cmake or something since vcpkg can build -head.

PR #4215, opened 2020-02-25
Fix an LLVM issue in BundleSaver reported by LLVM debug builds

### LINKED ISSUE #4214: Bundle Saver issue with multiple functions
This is exposed by testSaveMultipleFunctions when asserts are enabled in LLVM.

Symptom is assert in LLVMIRGen::emitConstOffsetsArray when old global is being replaced by new global. It asserts because size of globals are mismatched.

Looking briefly at it seems like root issue is that Weights/allocas are keep being "appended" to valueNumbers data structure in allocationsInfo_. It's a map but valueIdx keeps getting incremented which is then used in emitConstOffsetsArray to find maxOffset.  

This happens in BundleSaver::performBundleMemoryAllocation() when it calls allocationsInfo_.allocateActivations(F);

Haven't looked further. Not sure if this appending behavior is intended or data structure needs to be cleared between each function. Since I don't know what assumptions are for bundlesaver with multiple functions.

opti-mix:
There was a problem with a specific replaceAllUsesWith, as it used a different type for the replacement. This is solved now by emitting all entry functions at once at the very end of the bundle emission process.

Fixes #4214

### DIFF
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -458,6 +458,10 @@ void BundleSaver::emitSymbolTable() {
 void BundleSaver::produceBundle() {
   DCHECK(!isSaved_) << "produceBundle can be invoked only once";
   isSaved_ = true;
+  // Emit entry functions.
+  for (auto &savedFunction : savedIRFunctions_) {
+    emitBundleEntryFunction(savedFunction);
+  }
   // Finish code generation.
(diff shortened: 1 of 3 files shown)


ayermolo:
@opti-mix awesome, will try tomorrow.

ayermolo:
@opti-mix that did the trick, thanks!

opti-mix:
@ayermolo 

> that did the trick, thanks!

Thanks for confirming it! Landing.

PR #4229, opened 2020-02-27
Refactored image classifier so that common parts can be used by other apps

ayermolo:
Summary:
This stems from: https://github.com/pytorch/glow/issues/4020

The main idea is to refactor image-classifier in to app specific part and common part that can be used by other apps.

Most of the changes are just mechanical movement of common helper functions in to ExecutorCoreHelperFunctions.cpp

A new class was introduced in ExecutorCore.h that encapsulates core of building and running networks with extensions mechanism introduced that is similar to Loader one.

Another change was to modify buildAndCompileAndGetInAndOutPair to return a map of placeholders instead of just one. This is to enable networks with multiple outputs. For example ObjectDetection networks.

Documentation:
In the Issue link.
[Optional Fixes #issue]

Test Plan:
Unit tests, image classification networks.

### DIFF
--- a/tools/loader/ExecutorCore.cpp
+++ b/tools/loader/ExecutorCore.cpp
@@ -1,0 +1,453 @@
+/**
+ * Copyright (c) Glow Contributors. See CONTRIBUTORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "glow/Base/Image.h"
+#include "glow/Converter/TypeAToTypeBFunctionConverter.h"
+#include "glow/Importer/Caffe2ModelLoader.h"
+#include "glow/Importer/ONNXModelLoader.h"
+
+#include "llvm/ADT/StringSwitch.h"
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Format.h"
+#include "llvm/Support/Timer.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <atomic>
+#include <cfloat>
+#include <fstream>
+#include <future>
+#include <iostream>
+#include <memory>
+#include <mutex>
+#include <queue>
+#include <sstream>
+#include <thread>
+
+#include "ExecutorCore.h"
+#include "ExecutorCoreHelperFunctions.h"
+
+extern llvm::cl::opt<unsigned> traceLevel;
+
+using namespace glow;
+
+class PostProcessingExecutor final {
+public:
+  /// Iterates over registered extensions for processing and printing results
+  /// and executes them.
+  /// \return accumulated errors. Value greater then 0 indicates one or more
+  /// errros have occured.
+  int processOutputs(const llvm::StringMap<Placeholder *> &PHM,
+                     PlaceholderBindings &bindings,
+                     llvm::ArrayRef<std::string> inputImageBatchFilenames);
+
+  /// Registers Post Processing Output extensions.
+  void registerPostProcessOutputExtensions(
+      const std::vector<
+          std::function<std::unique_ptr<PostProcessOutputDataExtension>()>>
+          &extVector);
+
+private:
+  std::vector<std::unique_ptr<PostProcessOutputDataExtension>> extensions_;
+};
+
+class PreProcessImageExecutor final {
+public:
+  /// Iterates over PreProcessInputDataExtension extensions and executes them
+  /// one by ene.
+  void preProcessInputData(Tensor &inputImageData, size_t startId, size_t endId,
+                           size_t batchSz);
+
+  /// Registers Input Data Preprocessing Extensions.
+  void registerInputDataPreProcessingExtension(
+      const std::vector<
+          std::function<std::unique_ptr<PreProcessInputDataExtension>()>>
+          &extVector);
+
+private:
+  std::vector<std::unique_ptr<PreProcessInputDataExtension>> extensions_;
+};
+
+void PostProcessingExecutor::registerPostProcessOutputExtensions(
+    const std::vector<
+        std::function<std::unique_ptr<PostProcessOutputDataExtension>()>>
+        &extVector) {
+  for (auto &f : extVector) {
+    extensions_.push_back(f());
+  }
+}
+
+/// Iterates over registered extensions for processing and Printing results
+/// and executes them.
+int PostProcessingExecutor::processOutputs(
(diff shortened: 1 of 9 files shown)


ayermolo:
Looks like clang format is failing for other reasons.

opti-mix:
@jfix71 Could you have a look?

tools/loader/ExecutorCore.cpp
```diff
@@ -0,0 +1,426 @@
...
+      t.join();
+    }
+  }
+
+  if (!tracePath.empty()) {
+    traceContext->dump(tracePath, appName_);
+  }
+
+  return numErrors;
+  return 0;
```
jfix71: nit: rm this

tools/loader/ImageClassifier.cpp
```diff
@@ -413,22 +206,39 @@ static int processAndPrintResultsImpl(Tensor *SMT,
...
 /// Given the output Softmax Tensor \p SMT and \p functionName, switch between
 /// the correct element type to print the results of inference as contained in
 /// \p SMT, \returns the number of found mismatches.
-static int processAndPrintResults(Tensor *SMT,
-                                  llvm::ArrayRef<std::string> imageList) {
+int ImageClassifierProcessResult::processOutputs(
+    const llvm::StringMap<Placeholder *> &PHM, PlaceholderBindings &bindings,
+    llvm::ArrayRef<std::string> imageList) {
+  if (PHM.size() > 1) {
+    LOG(INFO) << "Network has more then one output. Process results not "
```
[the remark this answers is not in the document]
ayermolo: Right now in image classifier there is this:

```
if (!profilingGraph()) {
        std::lock_guard<std::mutex> lock(ioMu);
        numErrors += processAndPrintResults(bindings.get(outputPH),
                                            inputImageBatchFilenames);
}
```

I believe this was added there because of object detection profiling:
https://github.com/pytorch/glow/pull/3766

Now in refactoring I didn't want to have that check there because it needs to be generic, and down the road a real object-detection app can be build on top if it, but I didn't want to break current functionality. Until app is up-streamed people will use image-classifier for object detection. So I made it as a LOG info and return 0. That was kind of my thinking. Not 100% if it's best way to address this.

ayermolo:
@jfix71 Thanks for feedback.
Yes you are right error return handling is not super clean right now. For processOutputs I wanted to keep it same as with original implementation of processAndPrintResults, but didn't quite refactor everything correctly. i.e. implementation of PostProcessingExecutor::processOutputs, leaving std::exit(1), etc.
Option 1) fix up issues with inconsistencies you have pointed out and leave it as is returning int. It's a simple way to deal with errors in multithreaded environment. We have an int we increment it, with a lock in that scope, Boom done. Downside for other apps like object-detection not sure it's super useful. The mAp evaluation is usually done offline by python scripts, my understanding, and output post processing we just want to dump raw output with appropriate labels.

Option 2) Do something similar that is done for addNetwork and return glowError from various levels of APIs, extend ErrorValue::ErrorCode with representative enum. On plus side it's a more generic way. Downside would be more complicated. What if only some instances of processImageRange return error codes? Need to think about how to handle it. 

In some ways maybe option 1 is better alternative because it is a simpler way to handle error propagation up the chain of apis to the app. On app level we just need to know that error has occurred, and low level details can be handled by the logging mechanism.

ayermolo:
* Addressed comments.
* Left int return for now, but cleaned things up the returns and error value propagation.
* Incorporated https://github.com/pytorch/glow/pull/4191

ayermolo:
Is it possible to restart pipeline without some kind of force push with changes? I don't think OpenCL is failing related to my changes.

tools/loader/ImageClassifier.cpp
```diff
@@ -413,22 +206,39 @@ static int processAndPrintResultsImpl(Tensor *SMT,
...
+                 "ran for ImageClassifier.";
```
jfix71: grammar nit: "ran" -> "run"

tools/loader/ImageClassifier.cpp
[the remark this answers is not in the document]
jfix71: Got it. So if we're trying to maintain current functionality then would it make more sense to check `profilingGraph()` here to skip `processAndPrintResultsImpl()` and return 0? And then we could also check the size after and `LOG(FATAL)` if so? WDYT?

tools/loader/ExecutorCore.cpp
```diff
@@ -0,0 +1,449 @@
...
+    const std::vector<
+        std::function<std::unique_ptr<PostProcessOutputDataExtension>()>>
+        &extVector) {
+  for (auto &f : extVector) {
+    extensions_.push_back(f());
+  }
+}
+
+/// Iterates over registered extensions for processing and Printing results
+/// and executes them.
```
jfix71: Can you update the doxygen here -- what does it return and what does that return represent?

tools/loader/ExecutorCore.cpp
```diff
@@ -0,0 +1,449 @@
...
+        // Build and compile the graph, and then get back the input Placeholder
+        // and output Placeholder.
+        std::pair<Placeholder *, llvm::StringMap<Placeholder *>>
+            inputOutputPair = buildAndCompileAndGetInAndOutPair(
+                loader, bindings, &inputImageData.getType());
+
+        // If in bundle mode, the bundle has been saved by the above call, so we
+        // can safely return.
+        if (emittingBundle()) {
+          return;
```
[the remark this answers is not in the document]
ayermolo: Not sure I follow. It's in processImageRange lambda which currently returns void. Do you want lambda to return int?

ayermolo:
Thanks. Updated with suggestions. Also in execution lambda replaced llvm::outs with LOG(INFO) when we return and update numErrors. One open issue remains about return code for emitBundle()

ayermolo:
@jfix71 Looks like TSAN is failing due to CI issues:
+ echo 'Error, '\''TSAN'\'' not valid mode; Please, check .circleci/test.sh for list of supported tests.'
+ exit 1

tools/loader/ImageClassifier.cpp
```diff
@@ -413,22 +206,45 @@ static int processAndPrintResultsImpl(Tensor *SMT,
...
+/// Given the output PlaceHolder StringMap \p PHM, of size 1, from SoftMax and
+/// \p functionName, switch between the correct element type to print the
+/// results of inference as contained in \p PHM, \returns the number of found
+/// mismatches.
+int ImageClassifierProcessResult::processOutputs(
+    const llvm::StringMap<Placeholder *> &PHM, PlaceholderBindings &bindings,
+    llvm::ArrayRef<std::string> imageList) {
+
+  if (profilingGraph()) {
+    LOG(INFO) << "Graph profiling is ON. Processing of output is disabed.";
```
jfix71: typo: "disabed" -> "disabled"

tools/loader/ExecutorCore.cpp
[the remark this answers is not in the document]
jfix71: Nevermind me 🙂

tools/loader/ImageClassifier.cpp
ayermolo: oops, fixed.

ayermolo:
@jfix71 updated. Thanks for reviewing.

PR #4239, opened 2020-03-03
Fix NetworkExecutionStatePool available state tracking.

gcatron:
Summary:
Changed getNextNetworkExecutionState to use a deque to track available states instead of round robin.
Documentation:

Test Plan:
ninja check

### DIFF
--- a/include/glow/Runtime/Executor/NetworkExecutionState.h
+++ b/include/glow/Runtime/Executor/NetworkExecutionState.h
@@ -32,6 +32,8 @@ public:
   /// Constructor.
   explicit NetworkExecutionState(const DAGNode *root);
 
+  const DAGNode *getRoot() { return root_; }
+
   /// Destructor.
   ~NetworkExecutionState();
 
@@ -138,18 +140,23 @@ private:
 class NetworkExecutionStatePool {
 public:
   NetworkExecutionState *getNextNetworkExecutionState() {
-    // Note: this assumes pool size is maxActiveRequests, otherwise it's
-    // possible to wrap around on a state currently in use.
-    unsigned current = currentState_.fetch_add(1);
-    return states_[current % states_.size()].get();
+    std::lock_guard<std::mutex> lock(stateLock_);
+    auto nextState = availableStates_.front();
+    availableStates_.pop_front();
+    return nextState;
   }
-  void addNewState(std::unique_ptr<NetworkExecutionState> state) {
-    states_.push_back(std::move(state));
+
+  void addNewState(std::unique_ptr<NetworkExecutionState> state);
+
+  void returnNetworkExecutionState(NetworkExecutionState *state) {
+    std::lock_guard<std::mutex> lock(stateLock_);
+    availableStates_.push_back(state);
   }
 
 private:
   std::vector<std::unique_ptr<NetworkExecutionState>> states_;
-  std::atomic<unsigned> currentState_;
+  std::deque<NetworkExecutionState *> availableStates_;
+  std::mutex stateLock_;
 };
 
 } // namespace runtime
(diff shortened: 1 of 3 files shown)


include/glow/Runtime/Executor/NetworkExecutionState.h
```diff
@@ -25,6 +25,30 @@
...
+  void addNewState(std::unique_ptr<NetworkExecutionState> state);
+
+  void returnNetworkExecutionState(NetworkExecutionState *state) {
+    std::lock_guard<std::mutex> lock(stateLock_);
+    availableStates_.push_back(state);
+  }
+
+private:
+  std::vector<std::unique_ptr<NetworkExecutionState>> states_;
+  std::deque<NetworkExecutionState *> availableStates_;
```
yinghai: It might be more efficient just to use stack pop/push.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -214,6 +214,7 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
     // will transfer. //executionState->transferOutputs();
     ResultCBTy cb = executionState->getCallback();
     DCHECK(cb != nullptr);
+    executionState->returnToPool();
```
yinghai: If there a way to use some sort of RAII to make sure we return it no matter what?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
```diff
@@ -214,6 +214,7 @@ void ThreadPoolExecutor::handleDeviceManagerResult(
...
     cb(executionState->getRunId(), executionState->getErrorContainer().get(),
```
yinghai: Looks like we are using it after we return it?

lib/Runtime/Executor/ThreadPoolExecutor.cpp
tracelogfb: One pattern that I've seen working well is to just have the pool give out the NetworkExecutionState wrapped in a unique_ptr. Then have the user return the unique_ptr back to the pool once it's done. Basically this models "whoever has the pointer has ownership". It doesn't do RAII to ensure we always return to the pool though.

One potential blocker to do this is that in c++ 14, std::function can't wrap a lambda that captures a unique_ptr because unique_ptr isn't copyable. folly::Function is a better solution that handles this case correctly. https://github.com/facebook/folly/blob/master/folly/docs/Function.md

gcatron:
> A nitpick, so feel free to leave as is.
> 
> It feels a little unsymmetrical that we call `states_[root]->getNextNetworkExecutionState();` to obtain a state, but call returnToPool() on the object here. It feels more symmetrical if we just do `states_[executionState->getRoot()]-> returnNetworkExecutionState(executionState);`
> 
> Then we can avoid having the NetworkExecutionState class know anything about NetworkExecutionStatePool and avoid somewhat of a circular dependency/pointer tracking from State to Pool.

That's fair, I went this route because I didn't always have a reference to the pool handy, but I think I should just plumb that and get rid of the references altogether and just use unique pointers to the ExecutionState.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
gcatron: I don't think we can use a unique_ptr here because execution can fork. If there are two sibling nodes in the DAG both calls of executeDAGNode need a reference to the state. So we could use a shared_ptr but that doesn't enforce ownership.

include/glow/Runtime/Executor/NetworkExecutionState.h
gcatron: Is stack more efficient than deque? If they are similar I prefer deque for follow on P2P work.

yinghai: Stack is more cache friendly as you get the one that is recently checked back. But if you prefer dequeue we can check this in and measure later.

lib/Runtime/Executor/ThreadPoolExecutor.cpp
tracelogfb: I see. Good point. raw pointer is good then. It's better than shared pointer.

PR #4255, opened 2020-03-04
optimize NetworkExecutionState::bind to avoid string hashing

tracelogfb:
Summary:
NetworkExecutionState::bind takes around 0.5ms for a sample workload. Most cpu intensive part of this is us removing the name from the nameMap_ and then immediately add it in again. So add an update function in PlaceholderBinding so we only swap out the underlying tensor without doing anything to the nameMap.

This reduced the overhead from ~0.5ms to ~0.2ms.

Differential Revision: D20228515

### DIFF
--- a/include/glow/Graph/PlaceholderBindings.h
+++ b/include/glow/Graph/PlaceholderBindings.h
@@ -105,6 +105,10 @@ public:
   /// \p P must be a valid Placeholder registered in the bindings.
   void erase(Placeholder *P);
 
+  /// Removes the existing Tensor backing Placeholder \p P; Bind \p T to \P.
+  /// \p P must be a valid Placeholder registered in the bindings.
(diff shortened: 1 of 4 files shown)


opti-mix:
@tracelogfb Not to say that my proposal is any better for this specific issue, but there is a well known trick of pre-computing the hash for a key once and then reusing it. This way, you don't need to compute it every time. E.g. something along the lines of this SO answer: https://stackoverflow.com/questions/33188513/how-to-pass-hash-value-into-unordered-map-to-reduce-time-lock-held

Some programming languages are actually really storing the hash of a string together with a string in the internal representation, e.g. Java and some others. This of course works for constant strings only as you'd need to recompute the hash on every mutation.

tracelogfb:
@opti-mix that's a very nice technique!

PR #4260, opened 2020-03-04
glow runtime: use folly cpu thread pool

tracelogfb:
Summary:
In glow we have our own implementation of a ThreadPool. Glow's implementation is a pool of thread each having their own work queue. New work is assigned to different threads round robin.

This kind of implementation is usually good if and only if
1. all the work items are guaranteed to complete with similar amount of time. Or
2. we implement work stealing.

Otherwise we can end up with load imbalance where some threads have a heavy workload and long queue while other threads are idle.

Now that we have folly, we can just use folly's thread pool which is implemented with a shared queue using high performance data structure and primitives (MPMC queue + LIfoSemaphore). Using this implementation can reduce load imbalance and tail latency.

In my testing, the avg latency and throughput doesn't really change. But the tail latency is much better. The reason is because of shorter queuing time between device manager thread being done with an inference and the host manager executor thread handling the result.

Previously, the handling of the device manager result is always scheduled onto a thread that was pre-determined via round robin. If this thread happens to be busy (say handling some other input), then the result handling has to wait while there may be other threads that are totally idle. After the change, the result will be handled by any worker thread that's free.

You can observe this change in the glow trace.

Reviewed By: yinghai

Differential Revision: D20203927


.circleci/build.sh
```diff
@@ -49,10 +49,11 @@ install_fmt() {
     popd
 }
 
-GLOW_DEPS="libpng-dev libgoogle-glog-dev libboost-all-dev libdouble-conversion-dev libgflags-dev libjemalloc-dev libpthread-stubs0-dev libevent-dev libssl-dev"
+GLOW_DEPS="libpng-dev libgoogle-glog-dev libboost-all-dev libdouble-conversion-dev libgflags-dev libjemalloc-dev libpthread-stubs0-dev libevent-dev libssl-dev g++7"
+
+sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
```
yinghai: I just noticed that if you do 
```
sudo update-alternatives ...
```
as listed in https://github.com/pytorch/glow#ubuntu section. It will also work.

.circleci/config.yml
```diff
@@ -108,7 +104,6 @@ workflows:
       - DEBUG
       - OPENCL
       - ASAN
-      - TSAN
```
yinghai: :)

PR #4264, opened 2020-03-05
Add "mul" and "add" (int32/int64) operator tests to HW black list.

omromano:
Summary: Add missing "mul" and "add" (int32/int64) operator tests to HW black list.


omromano:
Hi @jfix71 @arunm-git 
Please take a look.

CC: @jsubag

PR #4273, opened 2020-03-08
[GraphOptimizer]Bug fix for Conversion Optimizations in the case of constant input

garg1himanshu:
Due to optimize conversions, Conversion from Int32 to Int64 throws an error in the case of Constant Input. convertConstant API gives NodeValue default value in this case and it replaces the current node with empty NodeValue. This is an addition just like in OptimizeQuantization API. 

I have not added any test case as it is a bug fix. Please add constant folding tests for the Cast Operator too.

### DIFF
--- a/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
+++ b/lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
@@ -2838,7 +2838,9 @@ bool OptimizeConversions::run(Function *F, const CompilationContext &cctx) {
       if (auto *BN = llvm::dyn_cast<Constant>(CN->getInput())) {
         auto newConst =
             convertConstant(*F->getParent(), *BN, CN->getResult().getType());
-        LOG_ASSERT(newConst) << "Constant conversion failed.";
+        if (newConst == NodeValue()) {
+          continue;
+        }
         CN->getResult().replaceAllUsesOfWith(newConst, F);
         changed = true;
         continue;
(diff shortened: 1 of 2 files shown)


lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
```diff
@@ -2838,6 +2838,9 @@ bool OptimizeConversions::run(Function *F, const CompilationContext &cctx) {
       if (auto *BN = llvm::dyn_cast<Constant>(CN->getInput())) {
         auto newConst =
             convertConstant(*F->getParent(), *BN, CN->getResult().getType());
+        if (newConst == NodeValue()) {
+          continue;
+        }
         LOG_ASSERT(newConst) << "Constant conversion failed.";
```
jfix71: I think this line should have been triggering in this case. I'm wondering if it should have been `newConst.getNode()` here instead. Regardless, it should be removed since it will never be reached.

jfix71:
@garg1himanshu Also please sign the CLA, thanks!

lib/Optimizer/GraphOptimizer/GraphOptimizer.cpp
garg1himanshu: Yes, you are correct this Int32-Int64 should not throw this error. I was looking at an older commit where Int2ITy-Int64ITy case was not covered. So now I don't have the test case. 
But won't it be better to not optimize the node and move ahead instead of giving Log Assert? In the cases of plain conversions, this case might occur.

jfix71: @garg1himanshu Yes, please go ahead and remove the `LOG_ASSERT()` and we can land this.

garg1himanshu: I have made the requested changes.

Issue #4158, opened 2020-02-12
cifar10 default graph runtime error


mrrosen:
When trying to run the cifar10 example, I ran into the same problem as this poster when they were running the mnist example:

https://discuss.pytorch.org/t/glow-example-run-error/50461

The same fix worked, by changing all the connections from MaxPool layers to use the `getResult` method. I dont know enough about glow to understand what the difference is as all the other layers are connected "directly" by referencing the object output from `createXXX` while the MaxPool ones need this `getResult`.

Changes that fixed the problem:

    --- a/examples/cifar10.cpp
    +++ b/examples/cifar10.cpp
    @@ -65,15 +65,15 @@ static Placeholder *createDefaultModel(PlaceholderBindings &bindings,
       auto *RL0 = F->createRELU("relu", CV0);
       auto *MP0 = F->createMaxPool("pool", RL0, 2, 2, 0);
    
    -  auto *CV1 = F->createConv(bindings, "conv", MP0, 20, 5, 1, 2, 1);
    +  auto *CV1 = F->createConv(bindings, "conv", MP0->getResult(), 20, 5, 1, 2, 1);
       auto *RL1 = F->createRELU("relu", CV1);
       auto *MP1 = F->createMaxPool("pool", RL1, 2, 2, 0);
    
    -  auto *CV2 = F->createConv(bindings, "conv", MP1, 20, 5, 1, 2, 1);
    +  auto *CV2 = F->createConv(bindings, "conv", MP1->getResult(), 20, 5, 1, 2, 1);
       auto *RL2 = F->createRELU("relu", CV2);
       auto *MP2 = F->createMaxPool("pool", RL2, 2, 2, 0);
    
    -  auto *FCL1 = F->createFullyConnected(bindings, "fc", MP2, numLabels);
    +  auto *FCL1 = F->createFullyConnected(bindings, "fc", MP2->getResult(), numLabels);
       auto *SM = F->createSoftMax("softmax", FCL1, expected);
       auto *save = F->createSave("ret", SM);
       return save->getPlaceholder();

jfix71:
Thanks for pointing this out @mrrosen -- would you like to put up a PR to fix this?

[...]

PR #4278, opened 2020-03-09
[Examples] Fix MaxPool layer connections in cifar10 example

mrrosen:
Summary:

Due to some change in the MaxPool API, connections in the graph need to be made via getResult; the cifar10 example was not updated to fix this problem and caused a runtime error when using the default graph.

Fixes: 

https://github.com/pytorch/glow/issues/4158


mrrosen:
I ran the format script on the repo and it did change some formatting on file I modified, but now it looks like the check is failing on other files I didnt touch?

jfix71:
@mrrosen It's not relevant, master is broken but the fix is landed. Thanks!

Issue #4165, opened 2020-02-13
[ONNX GraphOptz] Fold MatMul and Add into FullyConnected


mciprian13:
The ONNX format does not have a straight FullyConnected node (although the computation might be done with a Gemm). Therefore a lot of ONNX models use the FullyConnected functionality as MatMul + Add like this:
![image](https://user-images.githubusercontent.com/29785908/74441763-12b3bd00-4e79-11ea-96de-e8b716f2bcda.png)
Some backends might have an atomic implementation of the FullyConnected therefore it is desirable for Glow to recognize the pattern and **fold the combination into a FullyConnected** (before the Backends decides if it will be lowered or not).

mciprian13:
Solved in #4280. Closing this issue.

PR #4280, opened 2020-03-10
[GraphOptz] Fold MatMul + Add into FullyConnected.

mciprian13:
**Summary**
[Title]. The pair is very common in ONNX models and without this fold some Backends cannot use (specialize) their atomic FullyConnected implementations (if present).

**Fixes**
#4165
Also add verification that the bias for FullyConnected node is 1D.
Fix FullyConnected instance in TypeAToTypeBFunctionConverterTest.


mciprian13:
Can someone trigger again the CI builds?

lib/Optimizer/GraphOptimizerPipeline/Pipeline.cpp
```diff
@@ -110,6 +110,11 @@ FunctionPassPipeline glow::createDefaultGraphOptimizationPassPipeline() {
        ConvergenceMode::OnePass,
        {CompilationMode::Infer}},
 
+      // Fold MatMul + Add into FullyConnected.
```
jfix71: This better fits in `createDefaultFoldPassPipeline()` below -- otherwise if a backend does some lowering to create this MatMul -> Add pattern, this optimization would undo it. Fold only happens once, at the start of the optimization pipeline.

mciprian13: I am not yet familiar with the pipeline organization so I followed your advice:)
Done.

mciprian13: The Clang format job fails for some PyTorch files. Not sure what is going on with the CI ...

jfix71: Not your fault, another PR broke it, will fix separately. Thanks for fixing!

PR #4287, opened 2020-03-11
[FrontEnd] Add generic model-profiler tool.

mciprian13:
**Summary**
Up until now the only tool used for profiling was the `image-classifier` which was tailored only for image-classification applications. The new tool is more generic:
- does not make assumptions about the number and types for the model inputs/outputs
- uses raw binary/text input files
This tool coupled with the `model-compiler` (which is also generic) allows the quantization of any model.

**Documentation**
None

**Test Plan**
Unit tests for the Tensor utilities.


include/glow/Base/Tensor.h
```diff
@@ -551,6 +551,22 @@ class Tensor final {
   /// Dump a textual representation of the shape of this Tensor to std::string.
   std::string getShapeToString() const;
 
+  /// Dump the data content of the tensor to a raw-binary file.
+  void dumpToRawBinaryFile(const char *filename);
```
opti-mix: StringRef instead of const char *?

lib/Base/Tensor.cpp
```diff
@@ -379,6 +380,60 @@ std::string Tensor::toString(unsigned maxNumElem) const {
   return os.str();
 }
 
+void Tensor::dumpToRawBinaryFile(const char *filename) {
```
opti-mix: I feel like dumping some meta-information about the type of the tensor could be a good idea. Otherwise there is a high-risk of misinterpreting the content of the dumped tensor later.

opti-mix: Also, this approach currently supports dumping a single tensor per file if I understand correctly. But what if we want to dump e.g. all tensors used during the computation. There could be thousands of them. I wonder if it makes sense to support scenarios where multiple tensors could be dumped into/read from the same file.

opti-mix: Nit: Maybe all these serialization related functions should be in a separate file like Serialization.cpp. 

And I'm not sure they need to be methods on Tensor, especially if we'd support dumping multiple tensors into the same file.

include/glow/Base/Tensor.h
opti-mix: Where do you invoke these methods? I see only examples in tests.

mciprian13: The tensor loading methods (binary/text) are used in ModelProfiler.cpp for feeding the model input placeholders with data during the profiling precedure. The methods were designed to provide an easy method for the user to feed with data potentially generated in other places (a Python script, a C code):
- the binary loading method expects a blob which has the exact tensor size in bytes
- the text loading method expects a single array of comma separated values
Having these methods to have a more specific format (Glow`ish) by also including meta-information (like tensor type/shape) does not sound easy for user to generate externally.
I did not need the dump methods but I wrote them to match the loading methods.

lib/Base/Tensor.cpp
mciprian13: See the comments above. The profiling application just needs the loading methods, one tensor per file, better without the meta info in order to be easier to be externally generated.
In this case, do you think I should move them in ModelProfiler.cpp?
I thought that maybe that simpler format would be reutilized in some other places, especially for the front-end tools.

opti-mix: I think that the functionality for dumping/reading tensors is generally useful. Not only for profiling purposes, but e.g. for numerical testing purposes, etc.

But I'm still very worried about a format that does not include any meta-information (e.g. payload type and dimensions; maybe optional tensor name) as such a format is very likely to result in a lot of troubles if the mapping between tensors to be read and the dumped content files get mixed up. 

@kunmingho You actually looked into the same issues. Would you like to comment based on your experience with this?

lib/Base/Tensor.cpp
```diff
@@ -379,6 +380,60 @@ std::string Tensor::toString(unsigned maxNumElem) const {
...
+  fs.open(filename, std::ios::out | std::ios::binary);
+  DCHECK(fs.is_open()) << "Failed to open file: " << filename;
+  auto handle = getHandle<>();
+  for (dim_t rdi = 0; rdi < actualSize(); rdi++) {
+    fs.write((char *)&handle.raw(rdi), sizeof(handle.raw(rdi)));
+  }
+  fs.close();
+}
+
+void Tensor::dumpToRawTextFile(const char *filename) {
```
opti-mix: Dumping into a text file should probably use hex to save the precise values. Dumping floating point values using a decimal format is not reliable and may loose precision.

lib/Base/Tensor.cpp
```diff
@@ -379,6 +380,60 @@ std::string Tensor::toString(unsigned maxNumElem) const {
...
+  fs.open(filename);
+  DCHECK(fs.is_open()) << "Failed to open file: " << filename;
+  auto handle = getHandle<>();
+  for (dim_t rdi = 0; rdi < actualSize(); rdi++) {
+    fs << handle.raw(rdi) << ", ";
+  }
+  fs.close();
+}
+
+void Tensor::loadFromRawBinaryFile(const char *filename) {
```
opti-mix: What happens in the loading methods if the file being read contains less elements (or more elements) than expected? It seems like it should be possible to detect such errors.

include/glow/Base/Tensor.h
mciprian13: Replaced const char * with llvm::StringRef.

lib/Base/Tensor.cpp
mciprian13: Added verification: when the binary file is loaded, the file size is checked against the tensor size.

lib/Base/Tensor.cpp
mciprian13: But textual representation serves the purpose of readability. That is why I added the two flavors: binary to promote bit exactness, text to promote readability. I enhanced the method description to point that.

lib/Base/Tensor.cpp
mciprian13: I would prefer these functions as-is. I enhanced the description of the methods (after all they contain the name "raw" in the method names which suggests no meta information). These methods are useful when the tensors are already created and part of the graph (loading model input tensors, dumping model output tensors).

For the mechanisms you mention I think other methods should be developed (much more complex). In order for the file to be sufficient for a tensor description, a lot of meta-information should be there: data type, shape, padding/strides, quantization parameters (scale/offset).

include/glow/Base/Tensor.h
```diff
@@ -551,6 +551,36 @@ class Tensor final {
   /// Dump a textual representation of the shape of this Tensor to std::string.
   std::string getShapeToString() const;
 
+  /// Dump the data content of the tensor to a raw-binary file without any
```
opti-mix: As we are going to have many different serialization/deserialization methods, e.g. raw and more complex ones, and most of them do not really need to access the internal representation of Tensors, I'd suggest moving this functionality into dedicated TensorSerialization.h/.cpp files.

tests/unittests/TensorsTest.cpp
```diff
@@ -1293,3 +1293,41 @@ TEST(Tensor, differentAlignment) {
   T2H.at({1, 0}) = 1;
   EXPECT_FALSE(T1.isEqual(T2));
 }
+
+// Check that both write / read of tensors data from / to raw-text files is
+// working properly.
+TEST(Tensor, accessToRawTextFile) {
+  Tensor Tref = {0.75f,  0.23f, 0.76f,  0.99f,  1.00f,
```
opti-mix: Glow usually uses camelCase. It would  mean `tRef` in this case.

tests/unittests/TensorsTest.cpp
```diff
@@ -1293,3 +1293,41 @@ TEST(Tensor, differentAlignment) {
...
+                 -0.78f, 0.23f, -0.97f, -0.37f, 0.00f};
+  Tref.dumpToRawTextFile("raw_text_file");
+  Tensor Ttest(ElemKind::FloatTy, {10});
```
opti-mix: Ttest -> tTest

tests/unittests/TensorsTest.cpp
```diff
@@ -1293,3 +1293,41 @@ TEST(Tensor, differentAlignment) {
...
+  Ttest.loadFromRawTextFile("raw_text_file");
+
+  auto Href = Tref.getHandle<>();
```
opti-mix: Href -> hRef
Htest -> hTest

tests/unittests/TensorsTest.cpp
```diff
@@ -1293,3 +1293,41 @@ TEST(Tensor, differentAlignment) {
...
+  EXPECT_EQ(Href.actualSize(), Htest.actualSize());
+  for (size_t rcnt = 0; rcnt < Ttest.actualSize(); rcnt++) {
+    EXPECT_FLOAT_EQ(Htest.raw(rcnt), Href.raw(rcnt));
+  }
+}
+
+// Check that both write / read of tensors data from / to raw-binary files is
+// working properly.
+TEST(Tensor, accessToRawBinaryFile) {
+  Tensor Tref = {0.75f,  0.23f, 0.76f,  0.99f,  1.00f,
```
opti-mix: Same here.

mciprian13: Done.

tests/unittests/TensorsTest.cpp
mciprian13: Done.

tests/unittests/TensorsTest.cpp
mciprian13: Done.

tests/unittests/TensorsTest.cpp
mciprian13: Done.

mciprian13:
I agree with your proposal. I did not want after all to make these serialization methods the "de-facto" methods for the Tensor object. Therefore I moved them inside different source files as suggested: TensorSerialization.cpp and TensorSerialization.h.
Additionaly:
- I repaired the camelCase namings
- I added more validation to the tensor serialization methods (when loading from text files I validate that no more and no less values are in the file)
- Added documentation about this tool in AOT.md to promote usage for any model types.

mciprian13:
Can we have this landed?

One (unrelated) question:
I'm interested next in working in adding in Glow the capability for node splitting, that is define graph operations to split nodes in order to make a graph feasible for (extremely) memory contrained devices.
Is this something one of you guys is interested? With whom should I talk about the (possible) design of this (proper IR level to do this, etc)?
Thanks!

lib/Base/Tensor.cpp
```diff
@@ -379,6 +380,126 @@ std::string Tensor::toString(unsigned maxNumElem) const {
...
+
+template <class ElemTy>
+static void loadFromRawTextFileImpl(Handle<ElemTy> handle,
+                                    llvm::StringRef filename) {
+  std::ifstream fs;
+  fs.open(filename.data());
+  DCHECK(fs.is_open()) << "Failed to open file: " << filename.data();
+  char ch;
+  float val;
+  for (dim_t idx = 0, e = handle.actualSize(); idx < e; idx++) {
```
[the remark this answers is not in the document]
mciprian13: Done.

include/glow/Base/Tensor.h
mciprian13: Done.

opti-mix:
@jfix71 WDYT about this PR?

opti-mix:
> I'm interested next in working in adding in Glow the capability for node splitting, that is define graph operations to split nodes in order to make a graph feasible for (extremely) memory constrained devices. Is this something one of you guys is interested? With whom should I talk about the (possible) design of this (proper IR level to do this, etc)?

Are you talking about splitting a graph into smaller subgraphs? If so, this is supported by the partitioner already. 

Or are you talking about splitting a single node into smaller nodes? This functionality is not quite supported yet and it could be interested to discuss this.

mciprian13:
I`m referring to splitting nodes in smaller pieces.

tools/loader/LoaderUtils.cpp
```diff
@@ -18,17 +18,24 @@
...
+  std::exit(1);
+}
+
+void glow::checkCond(bool cond, llvm::StringRef errMsg) {
   if (!cond) {
-    llvm::errs() << "GLOW ERROR: " << errMsg << "\n";
+    llvm::errs() << "ERROR: " << errMsg << "\n";
     std::exit(1);
   }
 }
```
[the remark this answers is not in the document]
mciprian13: For front-end purposes I find the stack trace pretty intimidating.

docs/AOT.md
```diff
@@ -141,18 +141,134 @@ The procedure used for quantizing the model is called **profile-guided quantizat
 details [here](./Quantization.md)). Before the model is quantized and compiled with the
 *model-compiler* tool, the quantization profile must be acquired.
 
-In order to compute the quantization profile, the **image-classifier** tool is used. This
-application is used for image classification models only and requires a set of images to do
-the inference and compute the profile:
+In order to compute the quantization profile, the **model-profiler** tool is used. This
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/LoaderUtils.h
```diff
@@ -33,6 +33,12 @@
 
 namespace glow {
 
+/// Print message and exit with return code.
```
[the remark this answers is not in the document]
mciprian13: Done.

tools/loader/LoaderUtils.h
```diff
@@ -50,10 +56,14 @@ using UnlabeledDataSet = std::vector<UnlabeledData>;
...
-/// are valid and \returns the unlabeled data set.
-UnlabeledDataSet readUnlabeledDataSet(llvm::StringRef dataSetFile,
-                                      llvm::StringRef dataSetDirPath);
+/// given by \p dataSetDirPath. If the \p dataSetDirPath is empty then it will
+/// not be used. This function validates that all the data paths are valid and
+/// \returns the unlabeled data set.
+UnlabeledDataSet readUnlabeledDataSetFromFile(llvm::StringRef dataSetFile,
+                                              llvm::StringRef dataSetDirPath);
+
+/// Read an unlabeled data set as all the files from the given directory.
```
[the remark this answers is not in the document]
mciprian13: Done.

mciprian13:
> RE: splitting Nodes -- we have some functionality here for splitting some Nodes, but for parallelization/perf purposes. See `parallelizeOps()` in GraphOptimizer.cpp. Could be interesting to hear what you were thinking to shrink the size of the model.

There are networks where just 1 FullyConnected node requires a couple of MB of data. When dealing with extremely limited memory in the usecase of accelerators which can only access directly a small memory, up to 1 MB, the computation is not feasible (even though the accelerator has enough compute power).
I`m thinking about having some graph transformations (node splitting techniques) such that the amount of memory required by a node (that is the sum of all the inputs and outputs) does not exceed a given amount. The assumption is of course is that the small memory used by the accelerator can exchange data with a bigger RAM (where the graph input and weights are expected to be).

jfix71:
>I`m thinking about having some graph transformations (node splitting techniques) such that the amount of memory required by a node (that is the sum of all the inputs and outputs) does not exceed a given amount. The assumption is of course is that the small memory used by the accelerator can exchange data with a bigger RAM (where the graph input and weights are expected to be).

Got it -- so then the parallelization pass I pointed to make work well with this, take a look 🙂

opti-mix:
@jfix71 LGTM! Please land it.

PR #4294, opened 2020-03-12
Add device pinning to executionContexts to enable P2P/DRT

gcatron:
Summary:
This PR adds the ability to pin an executionContext to a specific deviceManager. This allows it to be used to store additional DRT info for copy reduction and P2P.
To test this feature there is a new flag in compilationContext: enableStaticAssignment. This should work for all backends, but will allow devices that support DRT and P2P to enable those features.

Documentation:

Test Plan:
Added a new unittest to Hostmanager test.

### DIFF
--- a/include/glow/Backend/Backend.h
+++ b/include/glow/Backend/Backend.h
@@ -48,7 +48,7 @@ struct DeviceInfo;
 struct DeviceConfig;
 struct ContextBinding;
 
-struct DAG;
+struct DAGNode;
 
 } // namespace runtime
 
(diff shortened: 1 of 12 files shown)


gcatron:
@jsubag I revisited bindContexts and slight changed the API, same info available just slightly different packaging. 
This change does not assume pinned memory in the cached executionContexts, that will enabled with a different flag later.

gcatron:
@jsubag FYSA diff is landed.

jsubag:
@gcatron Thanks!

ayermolo:
@gcatron I was looking in to a failure of ConcatSweepTest.ConcatTest_Float on windows.
Looks like it runs out of memory and allgined_malloc fails.
Test creates 129 placeholders with dims 34x64x128 and type float.
During createPool it creates 48 states, default size of maxActiveRequests, and for each state allocates all the place holders.
In a test it's doubled because bother CPU and Interpreter execution engines are active at same time. Actual internal size tensor is greater, when PH->getType()->getSizeInBytes() is queried. 
So it ends up allocating per execution engine 268435456 * 48 bytes. Almost 13 gb.
In this test there are two execution engines. One for CPU and another for interpreter.

Not 100% what the correct solution would be. What do you think?

On side note for my own curiosity.
When ThreadPoolExecutor::run is executed it pops off a state when getNextNetworkExecutionState() is called and executes a network. What is the link between how many times ThreadPoolExecutor::run is called and number of networkStates? Can we end up in a situation when run is called and availableStates is empty?

gcatron:
Hey @ayermolo , good catch. Short term I think the easiest fix is to lower maxActive requests, better term fix would be to allow ExecutionEngine to set maxActiveRequests so we can adjust on a test by test basis. I'll look into it.
To your side note:
maxActiveRequests is the link, by allocating as many states as activeRequests we ensure there will always be a state available.

gcatron:
@ayermolo 
Decided to just lower the default to 12 [https://github.com/pytorch/glow/commit/86315a8899ac276ec2a755913434f8c4ba73761e] That should fix the OOMing. Let me know if you are still encountering issues.

ayermolo:
@gabagam It worked. Thanks!

PR #4306, opened 2020-03-17
Support int32 index for SLS

yinghai:
Summary: We need to support int32 indices type to save memory bandwidth.

Differential Revision: D20484977


lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -3248,6 +3248,7 @@ void BoundInterpreterFunction::fwdLengthsSumInst(const LengthsSumInst *I) {
                             I->getData()->getElementType(), I)
 }
 
+template <typename ElemTy>
```
jfix71: nit: for consistency with others, can you name this `TI` instead of `ElemTy`

lib/Backends/Interpreter/InterpreterNodes.cpp
```diff
@@ -3386,6 +3390,7 @@ void BoundInterpreterFunction::fwdSparseLengthsWeightedSumInstFloatImpl(
   }
 }
 
+template <typename ElemTy>
```
jfix71: nit: for consistency with others, can you name this `TI` instead of `ElemTy`

PR #4308, opened 2020-03-17
[ConstFoldLoader] Add a test case with Operator input names having numeric values only

sladwa:
Many of the onnx models have the operators whose input names have numeric values. For this purpose a test case is added with const+Relu node combination.  The test case is added so as to catch any bug if at all introduced in constant fold loader functionality. 

Test Plan: Ninja Test is run


sladwa:
@jfix71 Kindly help to review the changes.

PR #4324, opened 2020-03-20
[TensorLayout] Add TileNode to the operator list which can accept any Layout

sladwa:
Summary: When Tile Operator is followed by Transpose Operator (Tranpose -> Tile) and Transpose operator has layout change like NHWC2NCHW, then Tile Node tensor layout verification fails since it is not part of list of operators which can accept any layout.

In this commit we are adding Tile Operator as a part of operator list which can accept any layout

Test Plan: Ninja Test is run


opti-mix:
cc @shajrawi

Issue #2600, opened 2019-03-26
Keep min/max in profile file


ayermolo:
Right now only scale/offset is in the profile file. In addition to it we can also have min/max used to generate it. This would allow for
1) extra flexibility for custom backends to implement custom quantization algorithms.
2) deciding what schema to use when profile is loaded.

mciprian13:
Solved by #4326. Closing this issue!

Issue #2599, opened 2019-03-26
Add quantization profile to yaml file


ayermolo:
Right now there is no direct link between what quantization schema was used to generate profile, and what was is used when it is loaded. This can lead to incorrect classifications. It's up to user to specify same command line option.
We can encode the schema used when profile is generated in to yaml file, and read it from profile when we load it. If during loading user specifies incompatible -quantization-schema option an error can be thrown.

mciprian13:
No longer required due to #4326. This issue should be closed.

PR #4326, opened 2020-03-20
[Quantization][Refactor] Make profile include only min/max

mciprian13:
**Summary**

Until this point the profile generated after the profiling phase included the quantization parameters **scale** and **offset** which were computed for a given quantization configuration (schema, precision, precisionBias). This approach was extremely error prone (multiple times I had this problem) because in order for the flow to be consistent, same configuration had to be provided during the profiling (e.g. while using image-classifier or model-profiler) versus during the quantization (e.g. while using the model-compiler).

With this PR the profile includes only the profiling specific information (**min** and **max**) while the actual quantization parameters are computed during the quantization phase. This has the following benefits:
- the profiling phase becomes totally independent on the quantization configuration (schema, precision, precisionBias) thus making the usage of the front-end tools less error prone (for model-profiler or image-classifier no quantization parameters are required)
- the information in the **profile.yml** is much more readable in terms of the tensor dynamic range (which I think was the purpose of using the YAML format in the first place) because just having the **scale** and **offset** was not indicative on the tensor dynamic range because:
  - the information itself was not sufficient without the data precision (given offset & scale mean totally different things for int8 vs int32)
  - some information was lost because for (e.g.) Symmetric quantization schema the information about the original (min,max) pair was lost due to symmetrization
- the profiling information is closer to where the actual quantization is performed, allowing more elaborate and precise quantization schemes (joint operand quantization, etc)

The functionality of serializing the tensor quantization parameters (scale & offset) is still there and we might use it for debugging (although the scales and offset are readable in the dumped graph).

**Fixes**
#2600
#2599 (No longer required)

**Documentation**
Updated AOT.md.

**Test Plan**
The quantization unit tests are passing.
Added new unit test for the new **getNodeValueByName()** utility function.
I manually tested the LeNet MNIST model and verified the results of the quantized bundle are exactly the same (bit exact) before and after this change. This is possible because the serialization of the min/max preserves the floating-point bit-exactness (just as it did for the floating-point scale).

**Next**
We should also include in the profile the histogram (which is already computed) and use some smarter quantization (outlier removal, KL divergence minimization, etc).

### DIFF
--- a/include/glow/Quantization/Quantization.h
+++ b/include/glow/Quantization/Quantization.h
@@ -53,7 +53,7 @@ generateNodeProfilingInfos(PlaceholderBindings &bindings, const Function *F,
 /// is the lowering map obtained during the quantization phase and is used
 /// to find lowering patterns for the bias operands.
 std::vector<NodeQuantizationInfo> generateNodeQuantizationInfos(
-    const std::vector<NodeProfilingInfo> &profilingInfos, Function *F,
+    llvm::ArrayRef<NodeProfilingInfo> profilingInfos, Function *F,
     const LoweredInfoMap &loweredMap = {}, Schema schema = Schema::Asymmetric,
     ElemKind quantizationPrecision = ElemKind::Int8QTy,
     ElemKind quantizationPrecisionBias = ElemKind::Int32QTy);
--- a/include/glow/Quantization/Quantization.h
+++ b/include/glow/Quantization/Quantization.h
@@ -30,16 +30,30 @@ class Backend;
 
 namespace quantization {
 
-/// Generate NodeQuantizationInfo for all required nodes from function \p F
-/// using the method specified by \p schema and target quantization precision \p
-/// quantizationPrecision. Profiling values will be written into context \p
-/// bindings. \p loweredMap maps from the NodeOutputName of a NodeValue which
-/// was lowered to a vector of the original NodeOutputNames which it replaced;
-/// this map is used to generate infos for the original unlowered NodeValues
-/// which no longer exist in \p F. The quantization precision for the bias of
-/// Convolution and FullyConnected is given by \p quantizationPrecisionBias.
+/// Generate NodeProfilingInfo for all the required nodes from function \p F
+/// using the profiling information stored in \p bindings obtained after
+/// running the network in profiling mode. During the profiling phase all the
+/// nodes are lowered. The map \p loweredMap maps the vector of NodeOutputNames
+/// obtained after lowering a NodeValue to the NodeOutputName of the lowered
+/// NodeValue. This map is used to replicate the profiling information for the
+/// unlowered NodeValues such that during the actual quantization, depending
+/// on the backend decision to lower some nodes or not, a profile will be found
+/// for any NodeValue.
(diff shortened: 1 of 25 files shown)


mciprian13:
@jackm321 Can we have this reviewed? Thanks!

lib/Quantization/Quantization.cpp
```diff
@@ -702,15 +701,15 @@ class FunctionQuantizer : public FunctionConverter {
...
+  /// Creates a function quantizer for \p F using the profiling
+  /// parameters defined by \p profilingInfos and target quantization
   /// precision defined by \p quantizationPrecision.
   /// \p B and \p doNotQuantizeKinds are used to check which nodes shouldn't be
   /// converted. \p assertAllNodesQuantized is used as a debugging tool; if
   /// true then if the backend does not support a node as quantized for the
   /// given \p quantizationPrecision then the program will exit with an error.
   FunctionQuantizer(Function &F, const Backend &B, quantization::Schema schema,
-                    llvm::ArrayRef<NodeQuantizationInfo> quantizationInfos,
+                    const std::vector<NodeProfilingInfo> &profilingInfos,
```
jackm321: Why the change away from ArrayRef? Can generateNodeQuantizationInfos just accept ArrayRef<NodeProfilingInfo>?

include/glow/Quantization/Quantization.h
```diff
@@ -30,16 +30,30 @@ class Backend;
...
+/// the function \ref generateNodeProfilingInfos. The quantization schema is
+/// specified by \p schema and the target precision for quantization is given
+/// by \p quantizationPrecision for all the operands except the bias operands
+/// (used in nodes like Convolution and FullyConnected) for which the target
+/// precision is given by \p quantizationPrecisionBias. The map \p loweredMap
+/// is the lowering map obtained during the quantization phase and is used
+/// to find lowering patterns for the bias operands.
 std::vector<NodeQuantizationInfo> generateNodeQuantizationInfos(
-    PlaceholderBindings &bindings, const Function *F,
+    const std::vector<NodeProfilingInfo> &profilingInfos, Function *F,
```
jackm321: ArrayRef?

lib/Quantization/Quantization.cpp
mciprian13: I changed it back to llvm::ArrayRef.

include/glow/Quantization/Quantization.h
mciprian13: Changed it to llvm::ArrayRef.

mciprian13:
@jackm321 Can we have this landed? Thanks!

mciprian13:
Side question: why the preference for llvm::StringRef and llvm::ArrayRef instead of std::string and std::vector? What are the benefits?

jackm321:
@mciprian13 llvm::StringRef and llvm::ArrayRef are preferred just because they don't mandate the underlying container of the argument

mciprian13:
@jackm321 Thanks for the explanation. Can we have this landed?

mciprian13:
@jfix71 Can we have this landed? I`m working on other quantization refactorings and I need this landed. Thanks!

PR #4341, opened 2020-03-24
[OnnxModelLoader] Add support for Upsample Onnx Operator

sladwa:
Summary: Adding support of Upsample Onnx Operator. Also adding CPU backend implementation of Resize Nearest IR.

Test Plan: Ninja Test is run


sladwa:
@jfix71 Kindly help to review the changes

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1733,6 +1733,65 @@ Error ONNXModelLoader::loadArgMax(const ONNX_NAMESPACE::NodeProto &op,
...
+Error ONNXModelLoader::loadUpsample(const ONNX_NAMESPACE::NodeProto &op,
+                                    ArgumentDictionaryTy &dict) {
+
+  RETURN_ERR_IF_NOT(opsetVersion_ < 10,
+                    "Upsample Operator is deprecated for Opset 10 and above");
+
+  const std::string &opName = loadOperatorName(op);
+  NodeValue in;
+  ASSIGN_VALUE_OR_RETURN_ERR(in, getNodeValueByName(op.input(0)));
+  std::string mode;
```
jfix71: Looks like the default is `"nearest"` -- can we use that here if so?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1733,6 +1733,65 @@ Error ONNXModelLoader::loadArgMax(const ONNX_NAMESPACE::NodeProto &op,
...
+  }
+
+  /// Only Nearest Mode is supported
+  RETURN_ERR_IF_NOT(mode.compare("nearest") == 0,
+                    "Upsample Operator has nearest mode support only");
+
+  /// Scale is always float as per onnx documentation
+  std::vector<float> scales;
+
+  RETURN_ERR_IF_NOT(opsetVersion_ > 6,
```
jfix71: Move this to the top/merge with the check for `opsetVersion_ < 10`

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1733,6 +1733,65 @@ Error ONNXModelLoader::loadArgMax(const ONNX_NAMESPACE::NodeProto &op,
...
+                    "Upsample Operator is supported for Opset 7 and above");
+
+  if (opsetVersion_ == 7) {
+    if (dict.count("scales")) {
```
jfix71: What if `opsetVersion_ == 7 && !dict.count("scales")`? We'd never initialize `scales`. What default values should be used in this case?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1733,6 +1733,65 @@ Error ONNXModelLoader::loadArgMax(const ONNX_NAMESPACE::NodeProto &op,
...
+    }
+    auto constH = scale->getPayload().getHandle<float>();
+    for (dim_t i = 0; i < constH.size(); ++i) {
+      scales.push_back(constH.at({i}));
+    }
+  }
+
+  /// NCHW2NHWC. scales tensor format is NHWC.
+
+  auto channel = scales[1];
```
jfix71: Check that scales has the correct shape right before this point?

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -91,7 +91,8 @@ bool Interpreter::isOpSupported(const NodeInfo &NI) const {
   case Kinded::Kind::ResizeNearestNodeKind:
     return NI.allInputsAndOutputsHaveSameElemKind(
         {ElemKind::FloatTy, ElemKind::Float16Ty, ElemKind::Int8QTy,
-         ElemKind::Int16QTy, ElemKind::Int32QTy});
+         ElemKind::Int16QTy, ElemKind::Int32QTy, ElemKind::Int32ITy,
```
jfix71: Is `Int16QTy` actually supported? Doesn't look like it? Also, looks like it was broken before since `FloatTy` is listed but wasn't actually supported?

lib/Importer/ONNXModelLoader.cpp
```diff
@@ -1733,6 +1733,65 @@ Error ONNXModelLoader::loadArgMax(const ONNX_NAMESPACE::NodeProto &op,
...
+  auto height = scales[2];
+  auto weight = scales[3];
+
+  scales[1] = height;
+  scales[2] = weight;
+  scales[3] = channel;
+
+  auto *intr = G_->createTranspose(opName, in, NCHW2NHWC);
+  Node *node = G_->createResizeNearest(opName, intr, scales);
```
jfix71: nit: for consistency use `auto` here like the ones below and above.

tests/unittests/OnnxImporterTest.cpp
```diff
@@ -3476,3 +3476,47 @@ TEST_F(OnnxImporterTest, CustomGlowChannelwiseQuantizedGroupConvolution) {
...
+  std::string netFilename(GLOW_DATA_PATH
+                          "tests/models/onnxModels/upsampleOpset7.onnxtxt");
+  importUpsampleTest(netFilename);
+}
+
+TEST(onnx, importUpsampleOpset9) {
+  std::string netFilename(GLOW_DATA_PATH
+                          "tests/models/onnxModels/upsampleOpset9.onnxtxt");
+  importUpsampleTest(netFilename);
+}
```
jfix71: nit: new line at end of file please.

lib/LLVMIRCodeGen/LLVMIRGen.cpp
```diff
@@ -480,6 +480,18 @@ llvm::Value *LLVMIRGen::emitConstDimTArray(llvm::IRBuilder<> &builder,
   return emitConstArray(builder, elems, DimTType);
 }
 
+template <typename T>
+llvm::Value *LLVMIRGen::emitConstFloatArray(llvm::IRBuilder<> &builder,
+                                            llvm::ArrayRef<T> vals) {
+  assert(std::is_floating_point<T>() && "Can only emit const array of float");
```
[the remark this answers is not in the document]
sladwa: Done.

lib/Importer/ONNXModelLoader.cpp
sladwa: Done.

lib/Importer/ONNXModelLoader.cpp
sladwa: Done.

lib/Importer/ONNXModelLoader.cpp
sladwa: https://github.com/onnx/onnx/blob/master/docs/Changelog.md#Upsample-7 does not mention default values and it has the field as required. So whoever generates the onnx file when they run check_model API then the API should fail

lib/Importer/ONNXModelLoader.cpp
sladwa: Done.

lib/Importer/ONNXModelLoader.cpp
sladwa: Done.

tests/unittests/OnnxImporterTest.cpp
sladwa: Done.

lib/Backends/Interpreter/Interpreter.cpp
sladwa: Since the implementation is templatized. It is supported

lib/Importer/ONNXModelLoader.cpp
jfix71: Ok, but to be safe we should also check and fail here. Instead of `if (dict.count("scales"))` can you simply do `RETURN_ERR_IF_NOT(dict.count("scales"), "scales is required for Upsample");` or something?

lib/Backends/Interpreter/Interpreter.cpp
jfix71: Ah, I thought this was CPUBackend, nevermind 🙂

sladwa:
Done. Have taken care of another review comment

PR #4350, opened 2020-03-25
[Quantization] Refactor ChannelwiseQuantizedConvolution.

mciprian13:
**Summary**
Refactor the ChannelwiseQuantizedConvolution node:
- add **biasScales** and **biasOffsets** operands.
- added extra flexibility in the function **createChannelwiseQuantizedConv**:
  - allows quantized or float filter (if float then is quantized channelwise automatically)
  - allows quantized or float bias (if float then is quantized channelwise automatically)
  - **biasScales** and **biasOffsets** are optional. If NOT provided (nullptr) then the implicit quantization parameters are used.
  - this extra flexibility can be used by any backend and hence it has been removed/moved from the Interpreter.
- changed Interpreter implementation to match the implementation of similar nodes:
  - same bias special handling is required even though the bias is quantized element-wise because for small bias precisions (e.g. Int8) the bias data is saturated for the implicit quantization parameters biasScales[i]=inputScale * filterScales[i] and biasOffsets[i]=0.
  - previous numerical behavior for bias is preserved by using nullptr for **biasScales** and **biasOffsets** when using **createChannelwiseQuantizedConv()**:.
- having all the scales/offsets parameters (optionally) provided through **createChannelwiseQuantizedConv()** API will allow tuning the quantization parameters of the node (e.g. brute force)

Extra:
- Add CPU implementation.
- Add dilation parameter.
- Add support for int8/int32 bias.
- Add logic to map a Convolution node to ChannelwiseQuantizedConvolution during the quantization procedure.
- Added utility function to quantize a tensor with finer granularity (quantize along a dimension qDim with given step qStep)
- Move logic for bias type specialization in Base.cpp.
- Fixed Caffe2Importer test.
- Other refactorings.

Note:
- the ONNX unit test model for the Custom Glow Importer has been modified to include the new inputs **biasScales** and **biasOffsets**.
- this can be reversed by adding in NodeGen a flag to mark optional input operands for the importer like `.addInput("BiasScales", Importer::Optional)` so that the generated
importer code in `AutoGenNodesImport.h` would look like this:

```
  NodeValue BiasScales = nullptr;
  if (op.input_size() > 5) {
    ASSIGN_VALUE_OR_RETURN_ERR(BiasScales, getNodeValueByName(op.input(5)));
  }
```

**Documentation**
None

**Test Plan**
New unit tests in OperatorTest, QuantizationTest.


mciprian13:
@jackm321 It seems the ASAN is failing with the following message:
```
../include/glow/Quantization/Base/Base.h:174:55: runtime error: -nan is outside the range of representable values of type 'long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../include/glow/Quantization/Base/Base.h:174:55 in 
```
There is a `-NAN` value encountered in one of the unit tests I added and I want to know how to reproduce the problem locally on my machine so I can debug it.
I tried building Glow with `-DCMAKE_BUILD_TYPE=Release` and `-DGLOW_USE_SANITIZER='Address;Undefined'` but I cannot reproduce the error.
Anyways this is very strange because the unit test has well-behaved values.

mciprian13:
@jackm321 It seems the PR #4356 destroys this PR entirely. I think mixing the definitions of Convolution and Convolution3D together is not a good idea because they are different operators (even though they are similar).
Do you think you could refactor and create a different node 'ChannelwiseQuantizedConv3D' and a separate Interpreter implementation in order to separate the two?
Otherwise this PR is all gone to waste.
Thanks!

include/glow/Quantization/Base/Base.h
```diff
@@ -135,6 +135,9 @@ struct QuantizationConfiguration {
   /// Whether to use rowwise quantization when quantizing a Function.
   bool enableRowwise{false};
 
+  /// Whether to use channelwise quantization when quantizing a Function.
+  bool enableChannelwise{false};
```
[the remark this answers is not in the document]
mciprian13: I would use separate flags in order to have some flexibility.
After all the semantics behind "rowwise" and "channelwise" apply to different types of nodes.

include/glow/Graph/Graph.h
```diff
@@ -460,23 +460,44 @@ class Function final : public Named {
...
-      NodeValue scales, NodeValue offsets, TypeRef outTy,
-      llvm::ArrayRef<unsigned_t> kernels, llvm::ArrayRef<unsigned_t> strides,
-      llvm::ArrayRef<unsigned_t> pads, unsigned_t group);
+      NodeValue filterScales, NodeValue filterOffsets, NodeValue biasScales,
+      NodeValue biasOffsets, TypeRef outTy, llvm::ArrayRef<unsigned_t> kernels,
+      llvm::ArrayRef<unsigned_t> strides, llvm::ArrayRef<unsigned_t> pads,
+      unsigned_t group, unsigned_t dilation = 1,
+      quantization::Schema schema = quantization::Schema::Asymmetric,
+      ElemKind filterElemQTy = ElemKind::Int8QTy,
+      ElemKind biasElemQTy = ElemKind::Int32QTy);
```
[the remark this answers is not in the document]
mciprian13: The reason why I added support for bias types other than Int32QTy is the same reason why I added support for other nodes as well (like standard Convolution and FullyConnected):
- Multiple times in a row we were given a third party library with optimized implementations for quantized kernels for which **we had no control over** (we were only given an API and a library in binary format) and we had to integrate them in Glow. For some reason, the guys who implemented the kernels chose to use Int8 bias. I`m not saying this is a good choice (using int8 bias instead of int32 bias has almost zero gain in terms of memory usage) but we had to do something in order to make Glow capable to prepare data for such kernels.
- Another case when we had to do this is when we used Glow to dispatch (offload) work to an accelerator through a firmware API. Again, that accelerator chose Int8 bias for its quantized implementations. Again we had no control over those implementations.
- I like to think about Glow as a tool generic enough to allow integrating kernels with all sorts of constraints. This is the reason why I also added a long time ago the quantization schema **SymmetricWithPower2Scale** which prepared the data for some quantized kernels which only used a bias up-shift and an output down-shift.

include/glow/Graph/Graph.h
```diff
@@ -460,23 +460,44 @@ class Function final : public Named {
...
+  ///   biasScales[i] = inputScale * filterScales[i] and biasOffsets[i] = 0.
+  ///   To be noted that this case can handle safely only INT32 bias data type
+  ///   because for INT8 type the bias will almost certainly be saturated.
+  /// This function requires \p filter and \p bias operands to be constants.
   ChannelwiseQuantizedConvolutionNode *createChannelwiseQuantizedConv(
       llvm::StringRef name, NodeValue input, NodeValue filter, NodeValue bias,
-      NodeValue scales, NodeValue offsets, TypeRef outTy,
-      llvm::ArrayRef<unsigned_t> kernels, llvm::ArrayRef<unsigned_t> strides,
-      llvm::ArrayRef<unsigned_t> pads, unsigned_t group);
+      NodeValue filterScales, NodeValue filterOffsets, NodeValue biasScales,
```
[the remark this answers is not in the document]
mciprian13: The channelwise (element-wise) quantized bias has the property that it is exactly representable in the quantized domain (no quantization loss) as long the scale parameters are chosen properly, even when using bias types lower than Int32QTy:
- By "chosen properly" I mean that although the implicit selection `biasScales[i] = inputScale * filterScales[i]` is natural for Int32QTy, for lower precision types it will almost certainly saturate the bias data (when quantized by Glow from float to that respective type like Int8Qty for example) and the unit tests actually prove that. And bias quantization is very important since the bias values directly impact the output samples of the convolution, even more so than the weight values.
- Therefore for lower bias precision types the implementation must be slightly more flexibile/generic and NOT assume the implicit scales `biasScales[i] = inputScale * filterScales[i]`.
- I also don't like having assumptions embedded in the inner core of an implementation (like we had for the Interpreter) since that constraints the usage. That is why I moved the logic of choosing the "proper" bias quantization parameters in `Graph.cpp`. That way all the backends can benefit from that logic (and not just the Interpreter) like the CPU for which I also added an implementation
- In terms of consistency I also think the ChannelwiseQuantizedConvolution should have an implementation similar to the others (same bias treatment as normal quantized convolution, etc).
- Finally, the logic I added in `Graph.cpp` is flexible enough so that when dealing with Int32QTy bias, if not provided with explicit bias scales parameters it will select the implicit scale parameters and the numerical behavior will be the same as now.
Sorry for the long explanation, but the point I'm trying to make is that the Glow contributors (like myself) might deal with different use cases and we might want future-proof some of the implementations by making them more generic.
Thanks!

mciprian13:
> Hi @mciprian13,
> 
> I'm still in the process of reviewing this and will get more comments in today but I had some questions about the motivation for separate bias qparams that hopefully you can enlighten me on.
> 
> #4356 seems to generalize the conv2d so that it also works for conv3d, is there something about this code that would not also be generalizable?

I just thought that it would be more clean to have separate graph definitions (nodes) to maintain the same line as for the floating-point versions where we have separate `Convolution` and `Convolution3D` nodes.

jackm321:
> I just thought that it would be more clean to have separate graph definitions (nodes) to maintain the same line as for the floating-point versions where we have separate Convolution and Convolution3D nodes.

I'm not sure I see the benefits, could you show me some of them? I added `Convolution3D` and (especially after seeing how readily `ChannelwiseQuantizedConvolution` was extended for 3D) I'm not sure why another node for 3D is better. I think we should have the same interface for Convolution and ChannelwiseQuantizedConvolution but imo it would be cleaner if we just folded Convolution3D into Convolution rather than split ChannelwiseQuantizedConvolution into a new ChannelwiseQuantizedConvolution3D op. Thoughts?

mciprian13:
@jackm321 Right now it is confusing because there are two functions in Graph.cpp which both instantiate the same node which under the hood have pretty different implementations (the implementations in the Interpreter are each on a big branch of an "if"). To enforce consistency between the channelwise and float standard conv versions I see 2 ways:

**Way1**:
If we want to have a joint ChannelwiseQuantizedConv & ChannelwiseQuantizedConv3D:
- we should have a single function in Graph.cpp for both (and same node definition as it is now)
- we should also merge Conv and Conv3D which should also use same node definition.

**Way2**:
If not, then we need to split ChannelwiseQuantizedConv from ChannelwiseQuantizedConv3D

I don't have a strong preference for which way to go but we should have consistency all the way.
I think we should also check someone else's opinion? Can you ask people on your side possibly interested in this topic?
Either way I want to proceed to work on this to make everyone happy and then have this landed 😄 

Thanks!

jfix71:
@mciprian13 I agree that there may be some weirdness in optimizations, and other interfaces such as `isOpSupported()`. Adding a separate Node makes things less error prone, as you have to go out of your way to add support for something for 2D and 3D.

OTOH there is some elegance in having them shared in a single Node, as we can skip the boilerplate code in many places where we want to support both. And we already do this when checking whether a Conv is quantized or not (i.e. we don't have a `QuantizedConvNode` and a `FloatConvNode` -- the Type is used to check, just like it could be done here).

I'm a bit torn here. But I'd probably lean toward using the same Node. (Related/for consistency I actually would like to get rid of `FusedRowwiseQuantizedSparseLengthsWeightedSum` and just use `SparseLengthsWeightedSum` with `UInt8FusedQTy` for `data`. But that's a slightly different story 🙂 )

mciprian13:
@jfix I think the argument of not specializing nodes based on types (quantized/float) just convinced me.

@jackm321 I will accept having the same node definition ChannelwiseQuantizedConv for both 2D and 3D cases and I will take the following action items:

- I will merge the Graph.cpp functions when creating the two flavors in order to use same boilerplate code for specializing the filter/bias scales/offset. I think this is also cleaner since it makes clear that same node definition is used.
- I will separate though the kernel implementations from InterpreterNodes.cpp to make it easier to read (right now the conv2d/conv3d cases are part of separate big "if" branches). I will use 2 separate functions
`fwdChannelwiseQuantizedConvolutionInstImpl2D` and `fwdChannelwiseQuantizedConvolutionInstImpl3D` and dispatch them from `fwdChannelwiseQuantizedConvolutionInst`.
- I will also add CPU implementation for the 3D version of ChannelwiseQuantizedConv
- I think some extra unit tests will be required.

@jackm321 @jfix71 If you agree with these points and you give me green light, I will make these changes in the following week. Thanks!

jfix71:
@mciprian13 Sounds like a good plan, thanks!

mciprian13:
@jfix71 @jackm321 I made the updates I previously mentioned. Additionally I also renamed the libjit convolution functions to "conv2d" and "conv3d" to have more clarity in what they do.
Please have a look!
The remaining issues I`m struggling with:
1. The value "NaN" is issued by ASAN in one of the ChannelwiseQuantizedConv2d unit tests. I tried capturing the NaN value in the indicated place using `std::isnan()` but I did not succeed. All the values used in the unit tests are well-behaved (random values between -1.0 and 1.0) so it's very weird that this happens. Do you know how can I capture the NaN value for debugging?
2.  A channelwise quantized conv2d unit test from Pytorch fails because there is a single value mismatch between Pytorch and Glow. Is it expected to have identical results? If yes then how am I supposed to debug it?

jfix71:
@mciprian13 Sorry for the delay -- I'm guessing the issue may be related to fast-math like in the other case, since you said it disappears when you change the kernel. I'd suggest trying to add a loop after the inner loop kernel to check values instead of modifying the inner loop which changes the optimizations the compiler applies.

mciprian13:
@jfix71 I found the problem. It was a problem in the testing infrastructure in `BackendTestUtils.cpp`. When converting the placeholders before running the inference, the tensors associated to the output placeholders were also passed through `quantizeTensor()` and since the inference was not yet run the tensors were initialized with random values (in my unlucky testcase the tensors had NaN values).

I modified the logic in the function `convertBindingsToCorrectType` to:
- convert tensor type and values for input placeholders
- convert only the tensor type for the output placeholders

Can we have this reviewed?
Is there something wrong with the CI?

mciprian13:
I don't understand at all why the latest build fail:
- The `32B_DIM_T` build fails because of `Interpreter/MLTest.testFindPixelRegression/0` which has nothing to do with this PR.
- The `PYTORCH` build fails intermitently because of a single sample which has an error. Is there something random which is not seeded properly because sometimes it works and sometimes it doesn't.

mciprian13:
@jackm321 @jfix71 It seems the PYTORCH build fails intermitently also for other PRs.
Can we have this PR reviewed/merged?
Thanks!

tests/unittests/OperatorTest.cpp
```diff
@@ -7025,10 +7025,290 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+  auto refH = bindings.get(saveRef->getPlaceholder())->getHandle();
+  for (dim_t idx = 0; idx < refH.size(); idx++) {
+    float errVal = std::abs(refH.raw(idx) - outH.raw(idx));
+    EXPECT_TRUE(errVal < 0.05);
+  }
+}
+
+#define TEST_CWQCONV(testName, ...)                                            \
+  TEST_P(OperatorTest, testName) {                                             \
+    ENABLED_BACKENDS("Interpreter", "CPU");                                    \
```
jfix71, line 7195: nit: Please try to use `CHECK_IF_ENABLED()` on new operator tests.

tests/unittests/OperatorTest.cpp
```diff
@@ -7025,10 +7025,290 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+      F->createConv("Conv", input, filter, bias, outTy, kernels, strides, pads,
+                    group, dilation);
+  SaveNode *save = F->createSave("save", conv);
+  auto *outputTensor = bindings.allocate(save->getPlaceholder());
+  return std::make_pair(F, outputTensor);
+}
+
+/// Test Int8 ChannelwiseQuantizedConvolution2D with Int8 bias.
+TEST_P(OperatorStatelessTest, ChannelwiseQuantizedConv2D_Int8_BiasInt8) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71, line 7286: nit: Please try to use `CHECK_IF_ENABLED()` on new operator tests.

tests/unittests/OperatorTest.cpp
```diff
@@ -7025,10 +7025,290 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+      ElemKind::FloatTy, ElemKind::Int8QTy, 0.05f, parCloneCountOpt,
+      /* convertToRowwiseQuantization */ false,
+      quantization::Schema::Asymmetric, ElemKind::Int8QTy,
+      /* forceFP16AccumSLS */ false,
+      /* convertToChannelwiseQuantization */ true);
+}
+
+/// Test Int8 ChannelwiseQuantizedConvolution2D with Int32 bias.
+TEST_P(OperatorStatelessTest, ChannelwiseQuantizedConv2D_Int8_BiasInt32) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71, line 7298: nit: Please try to use `CHECK_IF_ENABLED()` on new operator tests.

tests/unittests/OperatorTest.cpp
```diff
@@ -7025,10 +7025,290 @@ TEST_P(OperatorTest, GroupConvolution) {
...
+      /* forceFP16AccumSLS */ false,
+      /* convertToChannelwiseQuantization */ true);
+}
+
 /// Test the functionality of channelwise quantized group convolution using
 /// ChannelwiseQuantizedConvNode.
-TEST_P(OperatorTest, ChannelwiseQuantizedGroupConvolution) {
-  CHECK_IF_ENABLED();
+TEST_P(OperatorTest, ChannelwiseQuantizedConv2D) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71, line 7311: Why the change?

tests/unittests/OperatorTest.cpp
```diff
@@ -7106,8 +7389,8 @@ TEST_P(OperatorTest, ChannelwiseQuantizedGroupConvolution) {
 
 /// Test the functionality of channelwise quantized group convolution using
 /// ChannelwiseQuantizedConvNode.
-TEST_P(OperatorTest, ChannelwiseQuantizedGroupConvolution3D) {
-  CHECK_IF_ENABLED();
+TEST_P(OperatorTest, ChannelwiseQuantizedConv3D) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71, line 7393: Why the change?

tests/unittests/OperatorTest.cpp
```diff
@@ -7239,8 +7522,8 @@ TEST_P(OperatorTest, DilatedConvolution) {
 
 /// Test the functionality of channelwise quantized group convolution using
 /// ChannelwiseQuantizedConvNode with non-zero offsets and biases.
-TEST_P(OperatorTest, ChannelwiseQuantizedGroupConvolutionNonZero) {
-  CHECK_IF_ENABLED();
+TEST_P(OperatorTest, ChannelwiseQuantizedConv_NonZero) {
+  ENABLED_BACKENDS("Interpreter", "CPU");
```
jfix71, line 7526: Why the change?

lib/Graph/Graph.cpp
```diff
@@ -2672,72 +2672,169 @@ Convolution3DNode *Function::createConv3D(PlaceholderBindings &bindings,
...
         << Type::getElementName(biasElemKind).str();
   }
 
+  // Validate filter precision.
+  auto filterElemKind = filter.getElementType();
+  if (filterElemKind != ElemKind::Int8QTy &&
+      filterElemKind != ElemKind::FloatTy) {
+    LOG(DFATAL)
+        << "Unsupported element type for ChannelwiseQuantizedConvolution "
+        << "filter: " << Type::getElementName(filterElemKind).str();
```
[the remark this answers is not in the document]
mciprian13: Done.

lib/Graph/Graph.cpp
```diff
@@ -2672,72 +2672,169 @@ Convolution3DNode *Function::createConv3D(PlaceholderBindings &bindings,
...
+          filterC->getPayload(), filterScalesC->getPayloadMutable(),
+          filterOffsetsC->getPayloadMutable(), schema, filterElemQTy, qDim,
+          qStep);
+      filterScales = NodeValue(filterScalesC);
+      filterOffsets = NodeValue(filterOffsetsC);
+    } else {
+      LOG(DFATAL)
+          << "ChannelwiseQuantizedConvolution: If the input filter is "
+          << "quantized then the filter scales/offsets must be provided!";
+    }
```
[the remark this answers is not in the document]
mciprian13: Done.

tests/unittests/OperatorTest.cpp
mciprian13: Done.

tests/unittests/OperatorTest.cpp
mciprian13: Done.

tests/unittests/OperatorTest.cpp
mciprian13: Done.

tests/unittests/OperatorTest.cpp
mciprian13: Up until now I did not fully understand how CHECK_IF_ENABLED() worked. Now I do.
I reverted back.

tests/unittests/OperatorTest.cpp
mciprian13: Reverted back to CHECK_IF_ENABLED.

tests/unittests/OperatorTest.cpp
mciprian13: Reverted back to CHECK_IF_ENABLED.

mciprian13:
@jfix71 Thanks a lot for reviewing this PR.

mciprian13:
@jackm321 Any other wishes/issues on your behalf? I would be glad to have this landed in the following days. Thanks!

mciprian13:
@jackm321 Kind reminder for you to tell me how we proceed further with this PR.
Thanks!

jackm321:
@mciprian13 my apologies, taking another look now and importing

jackm321:
Other than keeping bias quantization in backends control, it looks good to me. There are some minor things to fix in the NNPI backend but I can do that before landing

lib/Backends/Interpreter/Interpreter.cpp
```diff
@@ -772,89 +782,14 @@ static bool quantizeFloatBias(Function *F, FullyConnectedNode &fullyConnected) {
   return true;
 }
 
-/// Channelwise quantize the given float \p bias as int32 using \p inputScale,
-/// per-channel \p scales and \p offsets. \returns false if the bias was already
-/// quantized and thus no change was made and true otherwise.
-static bool channelwiseQuantizeFloatBias(
```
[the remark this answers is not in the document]
mciprian13: @jackm321 But this is the same logic to quantize the bias used in `Graph.cpp` when creating the node when provided with float bias and requested to quantize to Int32QTy. Do you plan to change the logic for the Interpreter? If not then this logic is duplicated.
Other backends can of course opt to provide an already quantized bias when creating the node with their custom logic.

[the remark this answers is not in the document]
jackm321: @mciprian13 what I'm saying is not moving this logic to graph.cpp but instead keeping it here (or rather in some helper function that can be called here and in other backends that need quantized bias), that way the person who is using loading the model into Glow can load it with quantized or unquantized biases and each backend will do the right thing for what they need. This is because some backends have their own quantization logic for the bias and so to support those backends, the model would have to have biases prequantized with their logic ahead of time.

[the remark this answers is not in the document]
mciprian13: @jackm321 Ok, I will bring back the logic used for the Interpreter in order for you to customize the bias quantization in Interpreter the way you want. But I will not remove anything from the node creator in `Graph.cpp` since currently has the flexibility of:
- being provided with float bias and apply a **default** quantization logic
- being provided directly with quantized bias and so it will do nothing with respect to quantization (and so other backends can do this in order to quantize the bias the way the want). The Interpreter will go through this path after I revert.
Does this sound ok to you?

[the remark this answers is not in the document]
jackm321: @mciprian13 maybe I'm not understanding but the thing I think is missing (or got removed) from Graph.cpp is the ability to pass in a FloatTy bias and have it _not_ get quantized right there and instead just get attached to the node as a float tensor so that the backend can quantize it instead.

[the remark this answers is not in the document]
mciprian13: @jackm321 
I'm sorry for being wooden-headed but only now I understood you requirements.
So this is what I did:
- I modified the `createChannelwiseQuantizedConv` to have two additional boolean flags: `quantizeFilter` and `quantizeBias` with which you can express if the function should actually quantize the filter/bias or not. If not, the filter/bias will be attached to the node without being touched/quantized such that the backends can do it themselves using their own custom logic.
- In all the importers (Caffe2, ONNX, Pytorch) I used the flag `quantizeBias` to `false` such that the bias operand is not touched at all (allowing therefore to be float).
- I restored the Interpreter logic for manually quantizing the bias.

Does this sound ok to you now?
Thanks!

mciprian13:
@jackm321 Have a look now! Hope it is good for you!
Thanks!

lib/Backends/Interpreter/Interpreter.cpp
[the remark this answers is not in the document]
jackm321: No worries at all, thank you for making those changes so that it works as before, really appreciate it!

jackm321:
@mciprian13 Thank you for making those changes it looks good to me! And thank you for adding this functionality. Just imported it and fixed a few merge conflicts, will land it soon hopefully, waiting for CI now.

jackm321:
@mciprian13 its causing some correctness issues in one of our backends so I'm debugging that now

mciprian13:
@jackm321 Thanks for landing this. I'm sorry to hear you had problems ... Did you solve them?
Thanks!

PR #4351, opened 2020-03-26
[Quantization] Options to quantize input/output placeholders.

mciprian13:
**Summary**
- Add options to quantize separately the input/output placeholders. Sometimes we don`t want to quantize all the placeholders (e.g. when the model ends with a Softmax which does not have a quantized implementation). Having separate options for inputs/ouputs adds extra flexibility.
- I needed to add in the **generateQuantizationInfos** the missing TQP for the output placeholders which do not have profiling information (the output nodes are not profiled). Without this the quantization of the output placeholders will not work because the TQP will not be found.
- Replaced in the bundle header generation the manual shape string derivation with **type.toString()** which already includes the shape and additionally for the quantized types the scales and offsets.

**Documentation**
Document parameters in AOT.md.

**Test Plan**
Unit test in QuantizationTest.

**Question**
Should I provide in the **quantizeFunction** also the **PlaceholderBindings** so that during the placeholder quantization, the backing tensors (if allocated) are also quantized?

### DIFF
--- a/docs/AOT.md
+++ b/docs/AOT.md
@@ -324,11 +324,12 @@ image-classifier 0_1009.png ... -image-mode=0to1 -model=lenet_mnist -model-input
 model-compiler -backend=CPU -model=lenet_mnist -emit-bundle=build -model-input=data,float,[1,1,28,28] -load-profile=profile.yml
 ```
 
-It is important to note that currently the quantization of a model is performed
+It is important to note that by default the quantization of a model is performed
 only for the intermediate nodes of the graph, without affecting the data types of
 the model inputs and outputs. In the examples above, the data type for the model input
 (image tensor) and the model output remains **float** even though the intermediate
-operators and tensors use **int8** data type.
+operators and tensors use **int8** data type. If you want to convert also the model
+input and output placeholders you can use the option `convert-placeholders`.
 
 When compiling a quantized bundle you can choose to disable the quantization of some
 of the graph operators which might be more susceptible to quantization by using
--- a/lib/LLVMIRCodeGen/BundleSaver.cpp
+++ b/lib/LLVMIRCodeGen/BundleSaver.cpp
@@ -284,31 +284,21 @@ void BundleSaver::saveHeader(llvm::StringRef headerFileName) {
   auto placeholders = findPlaceholders();
   for (auto &v : placeholders) {
     auto *w = cast<WeightVar>(getWeightForNode(v));
-    // Get placeholder shape as string.
-    std::string shapeStr = "[";
-    auto dims = w->getType()->dims();
-    for (size_t idx = 0; idx < dims.size(); idx++) {
-      if (idx < dims.size() - 1) {
-        shapeStr += strFormat("%" PRIuDIM ", ", dims[idx]);
-      } else {
-        shapeStr += strFormat("%" PRIuDIM "]", dims[idx]);
-      }
-    }
     // Get placeholder properties.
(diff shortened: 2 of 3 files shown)


lib/Quantization/Quantization.cpp
```diff
@@ -737,6 +737,45 @@ class FunctionQuantizer : public FunctionConverter {
...
+        break;
+      }
+    }
+    return isOutputPlaceholder;
+  }
+
+  /// Function to quantize input placeholders. If the placeholder \p bindings
+  /// is given then also the backing tensors (if allocated) will be quantized.
+  void quantizeInputPlaceholders(PlaceholderBindings *bindings) {
+    for (auto ph : mod_.getPlaceholders()) {
```
jfix71: nit: `auto *ph`

lib/Quantization/Quantization.cpp
```diff
@@ -737,6 +737,45 @@ class FunctionQuantizer : public FunctionConverter {
...
+      }
+    }
+    cleanUp();
+    assert(function_.verify() && "Conversion led to invalid function");
+  }
+
+  /// Function to quantize output placeholders. If the placeholder \p bindings
+  /// is given then also the backing tensors (if allocated) will be quantized.
+  void quantizeOutputPlaceholders(PlaceholderBindings *bindings) {
+    for (auto ph : mod_.getPlaceholders()) {
```
jfix71: nit: `auto *ph`

lib/Quantization/Quantization.cpp
```diff
@@ -1037,6 +1092,16 @@ void quantizeFunction(Function *F, const QuantizationConfiguration &quantConfig,
   if (quantConfig.enableRowwise) {
     quantizer.enableRowwise();
   }
+
+  // Quantize input placeholders.
+  if (quantConfig.quantizeInputPlaceholders) {
+    quantizer.quantizeInputPlaceholders(nullptr);
```
jfix71: nit: `/* bindings */ nullptr`

lib/Quantization/Quantization.cpp
```diff
@@ -1037,6 +1092,16 @@ void quantizeFunction(Function *F, const QuantizationConfiguration &quantConfig,
...
+  }
+
+  // Quantize output placeholders.
+  if (quantConfig.quantizeOutputPlaceholders) {
+    quantizer.quantizeOutputPlaceholders(nullptr);
```
jfix71: nit: `/* bindings */ nullptr`

mciprian13:
@jfix Actually that is exactly what I needed (if the network ends in a Softmax then the output will not be converted with the options which is what I need). I replaced the code to use the already available graph optimization passes with a CLI opt in Loader.
Have a look and if fine by you, let's have this landed!
Thanks

jfix71:
@mciprian13 I'm totally fine with reusing the optimization -- but I think this option is still going to cause issues when used without bundle mode right? E.g. if we merge a `QuantizeNode` into an input `Placeholder`, then we need to make sure the backing input Tensor is quantized in a pre-processing step if it was not already quantized. If this is not interesting/important to your use cases I'd at least like some sort of error printed/elegant exit when `!emittingBundle() && ! convertPlaceholdersOpt`.

mciprian13:
@jfix I understand the problem but I thought the optimization pass takes care of the tensor conversion. I have a lot on my plate lately so I think I`ll go (for now) with option 2. I inserted the verification you mentioned.

PR #4352, opened 2020-03-27
Add a facility to not use libpng library during glow compilation if need be.

sladwa:
Summary: When we are building any application where glow is one sub-system of the whole application then we have seen cases where there is runtime conflict of png library being used by glow and other applications who also uses png library. To avoid such scenarios we are adding a flag during glow compilation where we need not use png library for glow compilation especially for "Base" folder.

Test Plan: Ninja Test is run.

### DIFF
--- a/lib/Base/CMakeLists.txt
+++ b/lib/Base/CMakeLists.txt
@@ -17,7 +17,8 @@ if(PNG_FOUND)
                                ${PNG_INCLUDE_DIR})
   target_link_libraries(Base
                         PRIVATE
-                          ${PNG_LIBRARY})
+                          ${PNG_LIBRARY}
+                          z)
 endif()
 
 add_dependencies(Base AutoGen)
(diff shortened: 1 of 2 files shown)


sladwa:
@jfix71 kindly help to review the changes being done.

lib/Base/CMakeLists.txt
```diff
@@ -17,7 +17,8 @@ if(PNG_FOUND)
                                ${PNG_INCLUDE_DIR})
   target_link_libraries(Base
                         PRIVATE
-                          ${PNG_LIBRARY})
+                          ${PNG_LIBRARY}
+                          z)
```
jfix71: Is this relevant/intentional?

sladwa: @jfix71 While compiling glow in docker(Based out of ubuntu 16.04) we have some times seen issues where libpng was giving error of missing zlib functions. For that reason "z" was added. If this is not required we can remove this line of change.

jfix71: 🤷‍♂ Fine by me, just wanted to make sure it was intentional.

PR #4364, opened 2020-03-31
Enable toggling P2P and DRT independently

gcatron:
Summary: This change enables toggling P2P and DRT independently.

Differential Revision: D20724510


gcatron:
@jsubag I expanded the testing for P2P/DRT and enabling toggling them independently.

PR #4370, opened 2020-03-31
[NNPI] Propagate/setup CompilationHints

jfix71:
Summary: Now that we have backend-specific per-node options setup, we can propagate them into NNPI CompilationHints.

Differential Revision: D20754375


tests/unittests/NNPIOptPipelineTest.cpp
```diff
@@ -415,6 +422,7 @@ static void setupSplitParallelizationTestFCReluNNPI(
   nodeInfo[TN]["NNPI_extraEdges"].push_back(FC->getName().str() + "@7");
   nodeInfo[RN]["NNPI_extraEdges"].push_back(SN->getName().str());
   nodeInfo[TN]["NNPI_extraEdges"].push_back(SN->getName().str());
+  nodeInfo[AN]["NNPI_extraEdges"].push_back(TN->getName().str());
```
jfix71, line 425: @jsubag This edge does not appear to be respected in the HW graph final solution.

PR #4372, opened 2020-03-31
Add option to load network inputs from tensor blobs.

### LINKED ISSUE #4371: Add option to load raw tensor blobs as network inputs
We have numerous customer request to support various image formats. More importantly, we have requests that are beyond known image formats, like combining input tensors/images along the channel dimensions etc.

We are looking to enable these users using  a simple interface that is able to load arbitrary 4D tensors from blob files (Any preprocessing on images/tensors is out of our scope - we just want to unblock the users).
For example, such interface could use a new  option -`-input-tensor-list-file` that closely follows `--input-image-list-file` by providing a list of tensor blob files that are to be loaded into the network input tensor (and concatenated along the batch dimension).

vuzelac-cadence:
Add option `-input-tensor-list-file `which allows passing the list of tensor blobs that are to be loaded into network inputs. This is very similar to `-input-image-list-file`.

Also allows for registering a custom tensor loader in case e.g. backend wants to use it's own tensor blob format. The default used loader blob format is described in `TensorSerialization.h `

All the loaded blobs are concatenated along the batch dimensions.

Fixes #4371 

Added new tests to TensorsTests.cpp

### DIFF
--- a/include/glow/Base/TensorSerialization.h
+++ b/include/glow/Base/TensorSerialization.h
@@ -16,8 +16,10 @@
 #ifndef GLOW_BASE_TENSOR_SERIALIZATION_H
 #define GLOW_BASE_TENSOR_SERIALIZATION_H
 
+#include "glow/Base/Image.h"
 #include "glow/Base/Tensor.h"
 
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 
 namespace glow {
@@ -51,6 +53,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
 /// of data is NOT guaranteed to preserve data precision (bit exactness) upon
 /// round-trips (dump/load) but is used for human readability.
 void loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename);
+
+/// Load network input tensor (same one images are loaded into) from the tensor
+/// blobs files in \p filenames. All the loaded tensors are concatenated along
+/// the batch dimension. The default loader expected the following tensor blob
+/// format:
+///   -- first line: 4 space separated values representing NCHW dimensions (W is
+///   the fastest).
+///   -- second line: Space separated float value that are loaded into \p
+///   tensor.
+/// TODO: Default tensor loader could be extended to support various data types
+/// or tensor layout.
+void loadInputImageFromFileWithType(
+    const llvm::ArrayRef<std::string> &filenames, Tensor *tensor,
+    ImageLayout tensorLayout);
+/// Helper function to aid testing of loadFromFileWithShapeLayout.
+void dumpInputTensorToFileWithType(const llvm::ArrayRef<std::string> &filenames,
+                                   const Tensor &, ImageLayout);
+
+/// Input tensor loader function. The function needs to set \p tensor type
+/// according to the data provided in the tensor blob. Expected layout is
+/// provided in \p imageLayout.
+using InputTensorFileLoaderFn = std::function<void(
+    Tensor &tensor, llvm::StringRef filename, ImageLayout imageLayout)>;
+
+/// Register input tensor loader function.
+void registerInputTensorFileLoader(InputTensorFileLoaderFn loader);
 
 } // namespace glow
 
--- a/lib/Base/TensorSerialization.cpp
+++ b/lib/Base/TensorSerialization.cpp
@@ -15,12 +15,15 @@
  */
 
 #include "glow/Base/TensorSerialization.h"
+#include "glow/Base/Image.h"
+#include "glow/Graph/Graph.h"
 
+#include "llvm/Support/CommandLine.h"
 #include <fstream>
 
 using namespace glow;
 
-namespace {
+namespace glow {
 
 /// Helper method to dump the tensor content into a raw text file.
 template <class ElemTy>
@@ -33,6 +36,50 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
     fs << handle.raw(idx) << ", ";
   }
   fs.close();
+}
+
+/// Helper method to load tensor files into the model input tensor.
+static void loadTensorFromFileWithType(Tensor &T, llvm::StringRef filename,
+                                       ImageLayout imageLayout) {
+  std::ifstream infile(filename.str().c_str());
+  CHECK(infile.is_open()) << "Error opening file '" << filename.data() << "'!";
+  std::string line;
+  ShapeVector dims;
+
+  CHECK(std::getline(infile, line)) << "Failed to read 1st line";
+  std::stringstream ss(line);
+  for (dim_t i = 0; i < 4; i++) {
+    int val;
+    CHECK(ss >> val) << "Failed to read dimension " << i;
+    dims.push_back(val);
+  }
+  T.reset(ElemKind::FloatTy, dims);
+  // Now read the tensor.
+  CHECK(std::getline(infile, line)) << "Failed to read 2nd line";
+  auto H = T.getHandle<>();
+  std::stringstream ss2(line);
+  for (dim_t i = 0, e = H.size(); i < e; i++) {
+    float val;
+    CHECK(ss2 >> val) << "Error loading file " << filename.data()
+                      << " @ element " << i;
+    H.raw(i) = val;
+  }
+  // Convert to requested layout (tensor blob is in NCHW by default).
+  if (imageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    T.transpose(&transposed, NCHW2NHWC);
+    T = std::move(transposed);
+  }
+}
+
+/// Set default tensor loader.
+static InputTensorFileLoaderFn inputTensorFileLoader_ =
+    loadTensorFromFileWithType;
+
+} // namespace glow
+
+void glow::registerInputTensorFileLoader(InputTensorFileLoaderFn loader) {
+  inputTensorFileLoader_ = loader;
 }
 
 /// Helper method to load the tensor content from a raw text file.
@@ -66,7 +113,6 @@ static void loadFromRawTextFileImpl(Handle<ElemTy> handle,
                       << handle.actualSize() << " elements!";
   fs.close();
 }
-} // namespace
 
 void glow::dumpToRawBinaryFile(Tensor &tensor, llvm::StringRef filename) {
   std::ofstream fs;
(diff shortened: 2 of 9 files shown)


vuzelac-cadence:
@jfix71 , github option to request the review doesn't show up for me.

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
...
 
+/// Load network input tensor (same one images are loaded into) from the tensor
+/// blobs files in \p filenames. All the loaded tensors are concatanated along
+/// the batch dimension. The default loader expected the following tensor blob
+/// format:
+///   -- first line: 4 space separated values representing NCHW dimensions (W is
+///   the fastest).
+///   -- second line: Space separated float value that are loaded into \p
+///   tensor.
+/// TODO: Default tensor laoder could be extended to support various data types
```
jfix71, line 65: typo: laoder

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
 /// round-trips (dump/load) but is used for human readability.
 void loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename);
 
+/// Load network input tensor (same one images are loaded into) from the tensor
+/// blobs files in \p filenames. All the loaded tensors are concatanated along
```
jfix71, line 58: spelling: concatanated -> concatenated

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
...
+/// the batch dimension. The default loader expected the following tensor blob
+/// format:
+///   -- first line: 4 space separated values representing NCHW dimensions (W is
+///   the fastest).
+///   -- second line: Space separated float value that are loaded into \p
+///   tensor.
+/// TODO: Default tensor laoder could be extended to support various data types
+/// or image layout.
+void loadInputTensorFromFileWithType(
```
jfix71, line 67: Since this method is specific to images can we name it more appropriately?

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
...
+    const llvm::ArrayRef<std::string> &filenames, Tensor *tensor,
+    ImageLayout imageLayout);
+/// Helper function to aid testing of loadFromFileWithShapeLayout.
+void dumpInputTensorToFileWithType(const llvm::ArrayRef<std::string> &filenames,
```
jfix71, line 71: Since this method is specific to images can we name it more appropriately?

lib/Base/TensorSerialization.cpp
```diff
@@ -15,12 +15,13 @@
  */
 
 #include "glow/Base/TensorSerialization.h"
-
+#include "glow/Base/Image.h"
+#include "llvm/Support/CommandLine.h"
```
jfix71, line 19: nit: Style is to put a line break to separate glow, llvm, and system includes (otherwise clang-format will alphabetize them and may put them in an order we don't want).

lib/Base/TensorSerialization.cpp
```diff
@@ -35,6 +36,49 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
   fs.close();
 }
 
+static void loadTensorFromFileWithType(Tensor &T, std::string filename,
```
jfix71, line 39: nit: Doxygen

jfix71: nit: `llvm::StringRef` or `const std::string &`

lib/Base/TensorSerialization.cpp
```diff
@@ -35,6 +36,49 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
...
+  for (dim_t i = 0; i < H.size(); i++) {
+    float val;
+    CHECK(ss2 >> val) << "Error loading file " << filename << " @ element "
+                      << i;
+    H.raw(i) = val;
+  }
+  // Convert to requested layout (tensor blob is in NCHW by default).
+  if (imageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    T.transpose(&transposed, {0u, 2u, 3u, 1u});
```
jfix71, line 67: nit: Use `NCHW2NHWC` here.

lib/Base/TensorSerialization.cpp
```diff
@@ -163,3 +206,65 @@ void glow::loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename) {
...
+    IIDH.insertTensors(H, {n, 0, 0, 0});
+  }
+}
+
+/// Helper function for loadInputTensorFromFileWithType, to produce blob files.
+void glow::dumpInputTensorToFileWithType(
+    const llvm::ArrayRef<std::string> &filenames, Tensor &T,
+    ImageLayout imageLayout) {
+  CHECK_EQ(filenames.size(), 1) << "Dumping support single file only";
+  std::string filename = filenames[0];
```
jfix71, line 249: nit: `const std::string &filename`

lib/Base/TensorSerialization.cpp
```diff
@@ -163,3 +206,65 @@ void glow::loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename) {
...
+  Tensor localTensor = T.clone();
+  // Convert to requested layout (tensor blob is in NCHW by default).
+  if (imageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    localTensor.transpose(&transposed, {0u, 3u, 1u, 2u});
```
jfix71, line 254: nit: Use `NHWC2NCHW` here.

tests/unittests/TensorsTest.cpp
```diff
@@ -1322,6 +1322,78 @@ TEST(Tensor, accessToRawTextFile) {
...
+    FAIL() << "Failed to create temp file to write into.";
+  }
+  dumpInputTensorToFileWithType({path.str()}, tensorRef, outImageLayout);
+  //
+  Tensor tensorTest;
+  loadInputTensorFromFileWithType({path.str()}, &tensorTest, inImageLayout);
+
+  if (outImageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    tensorRef.transpose(&transposed, {0u, 3u, 1u, 2u});
```
jfix71, line 1345: nit: Use `NHWC2NCHW` here.

tests/unittests/TensorsTest.cpp
```diff
@@ -1322,6 +1322,78 @@ TEST(Tensor, accessToRawTextFile) {
...
+    tensorRef = std::move(transposed);
+  }
+
+  if (inImageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    tensorTest.transpose(&transposed, {0u, 3u, 1u, 2u});
```
jfix71, line 1351: nit: Use `NHWC2NCHW` here.

tests/unittests/TensorsTest.cpp
```diff
@@ -1322,6 +1322,78 @@ TEST(Tensor, accessToRawTextFile) {
...
+    tensorTest = std::move(transposed);
+  }
+
+  auto handleRef = tensorRef.getHandle<>();
+  auto handleTest = tensorTest.getHandle<>();
+  EXPECT_EQ(handleRef.size(), handleTest.size());
+  EXPECT_EQ(tensorRef.dims(), tensorTest.dims());
+  for (size_t rcnt = 0; rcnt < tensorTest.actualSize(); rcnt++) {
```
jfix71, line 1359: nit: `for (size_t rcnt = 0, e = tensorTest.actualSize(); rcnt < e; rcnt++)`

tests/unittests/TensorsTest.cpp
```diff
@@ -1322,6 +1322,78 @@ TEST(Tensor, accessToRawTextFile) {
...
+}
+
+TEST(Tensor, tensorInputWriterLoaderNHWC) {
+  tensorInputWriterLoader(ImageLayout::NHWC, ImageLayout::NHWC);
+}
+
+// Test custom input tensor loader
+TEST(Tensor, tensorCustomInputLoader) {
+  bool entered = false;
+  auto loader = [&entered](Tensor &T, std::string filename,
```
jfix71, line 1383: nit: `llvm::StringRef` or `const std::string &`

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
...
+                                   Tensor &, ImageLayout);
+
+/// Input tensor loader function. The function needs to set \p tensor type
+/// according to the data provided in the tensor blob. Expected layout is
+/// provided in \p imageLayout.
+using InputTensorFileLoaderFn = std::function<void(
+    Tensor &tensor, std::string filename, ImageLayout imageLayout)>;
```
jfix71, line 78: nit: `llvm::StringRef` or `const std::string &`

lib/Base/TensorSerialization.cpp
```diff
@@ -163,3 +206,65 @@ void glow::loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename) {
...
+    CHECK_EQ(dims0[3], dims[3]) << "Non batch dimensions must match";
+    batchSize += data[n].dims()[0];
+  }
+
+  // Input tensor size is know now.
+  inputData->reset(ElemKind::FloatTy, {batchSize, data[0].dims()[1],
+                                       data[0].dims()[2], data[0].dims()[3]});
+  auto IIDH = inputData->getHandle<>();
+  // Insert each loaded file (in data[] tensors) as the input tensor slices.
+  for (dim_t n = 0; n < data.size(); n++) {
```
jfix71, line 238: nit: `for (dim_t n = 0, e = data.size(); n < e; n++)`

lib/Base/TensorSerialization.cpp
```diff
@@ -163,3 +206,65 @@ void glow::loadFromRawTextFile(Tensor &tensor, llvm::StringRef filename) {
...
+    inputTensorFileLoader_(data[n], filenames[n], imageLayout);
+    auto dims0 = data[0].dims();
+    auto dims = data[n].dims();
+    CHECK_EQ(dims0[1], dims[1]) << "Non batch dimensions must match";
+    CHECK_EQ(dims0[2], dims[2]) << "Non batch dimensions must match";
+    CHECK_EQ(dims0[3], dims[3]) << "Non batch dimensions must match";
+    batchSize += data[n].dims()[0];
+  }
+
+  // Input tensor size is know now.
```
jfix71, line 233: typo: know -> known

lib/Base/TensorSerialization.cpp
```diff
@@ -35,6 +36,49 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
...
+    int val;
+    CHECK(ss >> val) << "Failed to read dimension " << i;
+    dims.push_back(val);
+  }
+  T.reset(ElemKind::FloatTy, dims);
+  // Now read the tensor.
+  CHECK(std::getline(infile, line)) << "Failed to read 2nd line";
+  auto H = T.getHandle<>();
+  std::stringstream ss2(line);
+  for (dim_t i = 0; i < H.size(); i++) {
```
jfix71, line 58: nit: `for (dim_t i = 0, e = H.size(); i < e; i++)`

include/glow/Base/TensorSerialization.h
jfix71, line 67: nit: The input `tensor` is always reset, so maybe it would be cleaner/make more sense to just return a Tensor here instead of passing a pointer to a Tensor?

include/glow/Base/TensorSerialization.h
```diff
@@ -52,6 +54,32 @@ void dumpToRawTextFile(Tensor &tensor, llvm::StringRef filename);
...
+///   -- second line: Space separated float value that are loaded into \p
+///   tensor.
+/// TODO: Default tensor laoder could be extended to support various data types
+/// or image layout.
+void loadInputTensorFromFileWithType(
+    const llvm::ArrayRef<std::string> &filenames, Tensor *tensor,
+    ImageLayout imageLayout);
+/// Helper function to aid testing of loadFromFileWithShapeLayout.
+void dumpInputTensorToFileWithType(const llvm::ArrayRef<std::string> &filenames,
+                                   Tensor &, ImageLayout);
```
jfix71, line 72: nit: `const Tensor &`

include/glow/Base/TensorSerialization.h
vuzelac-cadence: Actually I don't see it as image related - it's simply is loading a tensor just like ModelProfiler does or TensorSerializer existing functions.
We still didn't have the need to load multiple inputs (like ModelProfiler does) so I've left this out.  Also, could merge with existing TensorSerializer functions.

jfix71: Well the reason I said this is because you pass in `ImageLayout`. Although it doesn't do anything if not set to the default, it would feel weird to use this with non-image Tensors.

And yeah probably makes sense to merge with TensorSerialization

vuzelac-cadence: I've renamed imageLayout into tensorLayout. Layout can be specified for a tensor and the function also can be used to load images converted to tensors (so we don't have to provide various image format loaders).

vuzelac-cadence: I was trying to match `loadImagesAndPreprocess` which does the similar job. Please confirm you are ok with this one taking different signature.

vuzelac-cadence: Hi @jfix71 , pinging you about this questions -please let me know if I should change or leave as is.

jfix71: @vuzelac-cadence I'm fine either way, if you want to leave it for consistency fine by me.

vuzelac-cadence: Ok thanks, I prefer to leave as is.

lib/Base/TensorSerialization.cpp
```diff
@@ -35,6 +37,49 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
...
+    Tensor transposed;
+    T.transpose(&transposed, {0u, 2u, 3u, 1u});
+    T = std::move(transposed);
+  }
+}
+
+/// Set default tensor loader.
+static InputTensorFileLoaderFn inputTensorFileLoader_ =
+    loadTensorFromFileWithType;
+
```
ugaresuyash, line 76: nit: Remove empty line.

vuzelac-cadence:
@jfix71 , seems I missed many of you comments. All addressed now.

include/glow/Base/TensorSerialization.h
jfix71: I would still prefer this to be called something like `loadInputImageFromFileWithType()` -- even if `imageLayout` is named `tensorLayout` it doesn't change that this API seems image-specific. E.g. the input must have a specific number of dimensions as it expects to load images. What are your concerns with renaming it? Do you actually use this with non-images?

vuzelac-cadence:
@jfix71  Take for example a stereo image where now input has 6 channels. For us, that's not image anymore - it's just some tensor, for all we care.
Sure the input has to be 4D, but for me that's just the limitation of this tensor loader, as we don't need anything better now. If that's a reason to call this image-specific, please confirm, I'll rename.

include/glow/Base/TensorSerialization.h
vuzelac-cadence: @jfix71 Take for example a stereo image where now input has 6 channels. For us, that's not image anymore - it's just some tensor, for all we care.
Sure the input has to be 4D, but for me that's just the limitation of this tensor loader, as we don't need anything better now. If that's a reason to call this image-specific, please confirm, I'll rename.

vuzelac-cadence: @jfix71 , fixed anyway.

lib/Base/TensorSerialization.cpp
```diff
@@ -35,6 +38,50 @@ static void dumpToRawTextFileImpl(Handle<ElemTy> handle,
...
+  // Convert to requested layout (tensor blob is in NCHW by default).
+  if (imageLayout == ImageLayout::NHWC) {
+    Tensor transposed;
+    T.transpose(&transposed, NCHW2NHWC);
+    T = std::move(transposed);
+  }
+}
+
+/// Set default tensor loader.
+static InputTensorFileLoaderFn inputTensorFileLoader_ =
```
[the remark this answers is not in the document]
vuzelac-cadence: @jfix71 , did you mean as argument to `loadInputImageFromFileWithType` ?
Note that backends or other parts of program don't call `loadInputImageFromFileWithType` and e.g. a backend can override file loading function only through `registerInputTensorFileLoader`. Thus, a state is needed. Hope I didn't misinterpret your question ?

PR #4379, opened 2020-04-01
Fix SLSBench rowsize for 4bit embeddings

jsubag:
Summary:
This change fixes the rowsize calculation for 4bit embeddings, where the dimensions are in bytes instead of elements.


jsubag:
@jfix71 Please take a look and add Michael

mjanderson09:
Looks good, thank you. I will go ahead and land it, unless you're able to do that directly.

PR #4388, opened 2020-04-04
[Windows] Cmake updates for Windows MSVC build.

mciprian13:
**Summary**
Lately we also switched to MSVC based Windows build (from Msys2 MINGW).
Some CMake updates in order for the Windows build to work.

**Documentation**
I removed from documentation the Msys2 build since it is very outdated and from the introduction of the `folly` dependency it is impossible to build.

**Test Plan**
Manually tested the build works on both Linux and Windows.


mciprian13:
Could you add in the CI also a Windows build?
I personally earn a living from the Windows build and there's a lot of wasted time in maintaining the Windows build and fix errors every time somebody adds a new dependency or makes some modifications with only the Linux build in mind. I guess a lot of people are interested in a stable Windows build.

ayermolo:
@mciprian13 Thanks. I was going to upstream similar changes, but been busy internally.
For some reason on some systems it was able to find jemalloc, but on others not.
One down side with find_library is that it finds release build in vcpkg, but seems to work for both debug and release. so...
Yes, if we can restart the CI discussion I would be up for it and can help.

Issue #4263, opened 2020-03-05
[Bundle][Feature] Error handling in LIBJIT with a return error code.


mciprian13:
This is more like a nice to have feature of having LIBJIT functions to do some error handling
and dropping a value of 0 (no error) vs something else to report problems.
For example the main bundle entry function could have the following prototype:
`uint32_t resnet50(uint8_t *constantWeight, uint8_t *mutableWeight, uint8_t *activations);`
that is to return an error code if something happens.
For the main LIBJIT math functions this is not necessarily very useful but for custom LIBJIT functions (e.g. DMA transfers, Remote Procedure Calls, etc) this is extremely useful.

The main bundle function could have a local variable `uint32_t errorCode` initialized with 0 and maybe have an option in LLVMIRgen for `createCall` called `errorCheck` which does bitwise OR between `errorCode` and the `uint32_t` returned by the function (or this could be done unconditionally by `createCall` for all the libjit functions which return an `uint32_t`). Lastly the bundle would return `errorCode`.

opti-mix:
@mciprian13 

Implementing this feature shouldn't be too hard. But I'd like to understand a bit better your intent here.

1) Do you suggest all kernels/libjit functions should return an error code that is 0 (no-error) by default? Or do you suggest just some of them may return error codes?

2) IIUC, you want to perform some error checking inside the main function of the bundle to propagate the errors if needed. I suspect you want to return the error code as soon as it was returned by a libjit function? Or do you plan to collect all errors (by doing `errorCode |= libjitCall(...)`) and then return `errorCode` at the end.

3) Do you suggest to bail on any error or only on a subset of expected errors specified by the client?

mciprian13:
@opti-mix I thing it would make sense something like this:
```
uint32_t main(uint8_t *constantWeight, uint8_t *mutableWeight, uint8_t *activations) {

  // Call kernel which has a uint32_t return type in its API.
  if (uint32_t errCode = kernel1(args)) {
    return errCode;  
  }

  // Call kernel which does not have a return in its API.
  kernel2(args);

  return 0;
}
```
So to respond to your questions:
1. I think we should NOT force that all the kernels to have a return type. After all for 99% of the kernels (math related) the return code is not necessary (and we don't want to modify all the current libjit kernel APIs if not needed). The procedure which generates the function call could inspect the function API and:
   - if the kernel has a `uint32_t` return type then it generates the code to bail out immediately if the return code is different than 0.
[...]

PR #4392, opened 2020-04-06
[Debug] Dump tensors in separate txt/bin files.

mciprian13:
**Summary**
Besides the existing capability to dump all the tensors as text file in the console I also added the capability to dump as separate binary files.
I opted for using a simple name format for the dumped binary files ("data[idx].bin") for two reasons:
- using the DebugPrint instruction name as the dump file name is very hard to follow in any explorer:
FileDumpList
- is NOT safe because any filesystem has a maximum limit for a file name. I actually ran into this problem when dumping the tensors from MobileNet v1: half of them where not dumped because the `fopen` function returned error because the filenames were too long.

In order to associate a given binary dump to a given IR tensor I dump an extra meta file called `debug.info` which groups the information for all the instructions like this:
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
Output: data0002.bin : Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2_res
Input : data0003.bin : Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2_quantize_res
Input : data0004.bin : MobilenetV1_MobilenetV1_Conv2d_0_Conv2D_weight__1
Input : data0005.bin : conv_bias__27

Type: cpumaxsplat
Name: Clip_MobilenetV1_MobilenetV1_Conv2d_0_Relu6__1_minClip__1
Output: data0006.bin : Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2_res
Input : data0007.bin : Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2_res
```

I would have liked to print the operand names from the instruction definition (but currently there is not way to do it without modifying the InstrGen) which I think would have been the most readable way, like this:
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
Dest:  data0002.bin
Src:     data0003.bin
Filter:  data0004.bin
Bias:    data0005.bin
```

**Documentation**
Added section in AOT.md.

**Fixes**
#3845

**Test Plan**
Manually verified the dumps are correct for first 3 convolution layers in MobileNet v1 1.0 224 (this is actually why I needed this capability in the first place).


ayermolo:
@tlepley

vuzelac-cadence:
@mciprian13 @opti-mix 
JITTed code has no ability to share mutexes and such to allow e.g. concurrent access to the same files from different threads.
In addition  using fprintf from JITTed code causes issues on windows (hard to hit, we've seen it only on Windows Server 2016). This seems to be unsolvable issue in NTDLL, which is not upgrade-able.
I am sending you a MR that addresses these issues, that we use internally for various dumps mechanisms we use internally. #4395 . MR allows for using host provided print functionality. It would be good to merge this stuff.

opti-mix:
cc @kunmingho @SameerAsal

opti-mix:
@mciprian13 

> I would have liked to print the operand names from the instruction definition (but currently there is not way to do it without modifying the InstrGen)

The ability to ask for a name of given positional operand (in case of instructions) or input/output in case of nodes  could be a useful extension for InstrGen/NodeGen, which is very easy to implement. Could you file an issue about it?

mciprian13:
> @mciprian13
> 
> > I would have liked to print the operand names from the instruction definition (but currently there is not way to do it without modifying the InstrGen)
> 
> The ability to ask for a name of given positional operand (in case of instructions) or input/output in case of nodes could be a useful extension for InstrGen/NodeGen, which is very easy to implement. Could you file an issue about it?

@opti-mix I opened an issue here: #4396

mciprian13:
> @mciprian13 @opti-mix
> JITTed code has no ability to share mutexes and such to allow e.g. concurrent access to the same files from different threads.
> In addition using fprintf from JITTed code causes issues on windows (hard to hit, we've seen it only on Windows Server 2016). This seems to be unsolvable issue in NTDLL, which is not upgrade-able.
> I am sending you a MR that addresses these issues, that we use for various tensor dumps mechanisms we have internally . MR allows for using host provided print functionality. I'd like your input on #4395 usability in your use cases.

@vuzelac-cadence I am not versed in the JIT engine so I don't think I can provide useful feedback for your PR. All I can say is that I am almost entirely interested in using single-threaded applications using the Glow bundles (AOT).  This PR fits my needs so I would be grateful if this functionality is preserved as is. Do you think we could have this PR merged and then you could specialize the dumping mechanism for the JIT case? That would be awesome.
Thanks!

ayermolo:
@mciprian13 I think even in single threaded case you might hit an issue on some windows machines.

vuzelac-cadence:
@mciprian13, multi-threading will come automatically with e.g. `minibatch `option.
Also:
1. Any chance to add some tests (e.g. reading & verifying actual dumped files) ?
2. Did you verify this with `minibatch `? I am not sure if consecutive runs will dump to the same tensors, overwriting the existing content ? 
3. I'd suggest to print tensor shape/layout as the 1st line of either txt/bin files.
4. Maybe add directory path along with filename path.
As far as my JIRA is concerned, seems it might be an easy patch later, to take over fprintf & al functions.

opti-mix:
@mciprian13 #4398 should provide you with the required APIs to produces nicer meta files. I'd also suggest something like the following format:

```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
Dest (0):  data0002.bin
Src (1):     data0003.bin
Filter (2):  data0004.bin
Bias (3):    data0005.bin
```

The idea is to use the operand number after the operand name, so that when you read it back you can easily determine which operand is meant without implicitly relying on the fact that all operands are sorted in this file. WDYT?

Note: Alternatively, we could introduce `unsigned getOperandIdx(llvm::StringRef)` to get the operand index based on its name, but it seems like it could be too heavy weight eventually as the use-case for it would be very rare.

mciprian13:
> @mciprian13, multi-threading will come automatically with e.g. `minibatch `option.
> Also:
> 
> 1. Any chance to add some tests (e.g. reading & verifying actual dumped files) ?
> 2. Did you verify this with `minibatch `? I am not sure if consecutive runs will dump to the same tensors, overwriting the existing content ?
> 3. I'd suggest to print tensor shape/layout as the 1st line of either txt/bin files.
> 4. Maybe add directory path along with filename path.
>    As far as my JIRA is concerned, seems it might be an easy patch later, to take over fprintf & al functions.

@vuzelac-cadence 
1. I will add some unit tests for the dumps.
2. I did not verify instrumenting with debug a model running in multi-threaded minibatch mode (and I expect there to be issues when competing over the same dump files) but I'm not sure why would you want to instrument this way if the expected behavior is that the dump files are rewritten such that in the end only the last minibatch will have their data dumped? If you want that last minibatch dumped why not run inference on that alone? Conversely, why not dump all the minibatches in separate files by propagating to the IR (somehow) some minibatch index?
3. I'm not sure that fits my purpose since if I want to use the raw binaries to load them directly in some other application I would have to manually strip that meta information. But I am aware that we all have slightly different uses in mind so what I propose in the end is 4 possible formats:
   - raw binary/text: without meta information about shape/layout/etc
   - typed binary/text: with meta information about shape/layout/etc.
I think we should standardize these formats across all glow utilities (e.g. the logic from libjit/interpreter should match in the end the logic in TensorSerialization.cpp)
4. Got it, makes sense!


@opti-mix Thanks for adding the Instruction API for operands. I will use the format as you suggested ... with maybe a little twist?
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
[0] Dest:    data0002.bin
[1] Src:     data0003.bin
[2] Filter:  data0004.bin
[3] Bias:    data0005.bin
```

Let me know what you think.
Thanks!

mciprian13:
> @mciprian13 #4398 should provide you with the required APIs to produces nicer meta files. I'd also suggest something like the following format:
> 
> ```
> Type: convolution
> Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
> Dest (0):  data0002.bin
> Src (1):     data0003.bin
> Filter (2):  data0004.bin
> Bias (3):    data0005.bin
> ```
> 
> The idea is to use the operand number after the operand name, so that when you read it back you can easily determine which operand is meant without implicitly relying on the fact that all operands are sorted in this file. WDYT?
> 
> Note: Alternatively, we could introduce `unsigned getOperandIdx(llvm::StringRef)` to get the operand index based on its name, but it seems like it could be too heavy weight eventually as the use-case for it would be very rare.

@opti-mix Thanks for adding the Instruction API for operands. I will use the format as you suggested ... with maybe a little twist?
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
[0] Dest:    data0002.bin
[1] Src:     data0003.bin
[2] Filter:  data0004.bin
[3] Bias:    data0005.bin
```

vuzelac-cadence:
@mciprian
2.  multi-threaded minibatch mode: 
 _"I'm not sure why would you want to instrument this way if the expected behavior is that the dump files are rewritten such that in the end only the last minibatch will have their data dumped?"_

I didn't get this - why is this the expectation?
Minibatch is a part of microarchitecture - a way to speed up things, that should not be externally visible. What you get when running w/o minibatch is what you should get with minibatch.
A tensor X, that is say dumped as e.g. 2 20 20 3 for each minibatch (when --minibatch=2), should at the end be written as N 20 20 3, where N is the number of images. As a temp. solution files can be names with some postfix that always increases, to indicate consecutive dumps of the same tensor. It would be clunky to parse, but still - the final output information is not dropped.

3. Yes, but you can strip. But whoever needs that info cannot add it. So it's best to be there as instrument-debug should be generally usable.

opti-mix:
@mciprian13 Regarding the format of the meta info file: Yes, your twist looks totally fine! Go for it.

kunmingho:
> 3. I'm not sure that fits my purpose since if I want to use the raw binaries to load them directly in some other application I would have to manually strip that meta information. But I am aware that we all have slightly different uses in mind so what I propose in the end is 4 possible formats:
>    
>    * raw binary/text: without meta information about shape/layout/etc
>    * typed binary/text: with meta information about shape/layout/etc.
>      I think we should standardize these formats across all glow utilities (e.g. the logic from libjit/interpreter should match in the end the logic in TensorSerialization.cpp)

Having different formats is a good direction to goal as you already mentioned "we all have slightly different uses". For example, we do load binary tensors into other application and we "need" the meta info.

For binary output, if meta info will be included, I would suggest to also add "version" since the format could be changed in the future.

kunmingho:
If the reader APIs can also be provided, that will make this even more complete :-) And I think if unit tests will be created, the reader APIs are needed anyway.

mciprian13:
Overview:
- I opted for implementing only the **rawtxt** and **rawbin** dump formats. I don't have currently the bandwidth of implementing other formats. I suggest implementing other formats (**typedtxt** and **typedbin**) in other PR's.
- I added unit tests for the dumping mechanism (both CPU & Interpreter) and crosschecked with the current functions from **TensorSerialization.cpp**.
- I added the CLI option `-debug-instrument-dir` to choose the debug directory (default is **debug** where all the dumps and the meta file **debug.info** are written)
- Currently the meta file **debug.info** looks like this (after using the Instruction operand API):
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
[0] Dest:      data0002.bin
[1] Src:       data0003.bin
[2] Filter:    data0004.bin
[3] Bias:      data0005.bin
```
@opti-mix @vuzelac-cadence There are some remaining things for which I don't have time for right now (or I don't know how to do) like other dump formats (the ones with meta info which I'm not sure how they look or how do you want them to be) and the minibatch problem so my question is whether we can have this landed and maybe I will contribute again when I'll have some bandwidth? What do you say?
Thanks!

mciprian13:
So after all I added also the bin/txt formats with the tensor type as header. Overview: 
- I wrote an utility function `Type::fromString()` which can deserialize a Type object from string
- I updated the functions from `TensorSerialization.cpp` to accomodate both flavors of the formats (with/without tensor type)
- So currently there are 4 dump formats:
  - bin/txt (binary/txt with tensor type/meta) - to suit your needs
  - rawbin/rawtxt (binary/txt without tensor type/meta) - to suit my needs
- Added new unit tests
- I also added the operand type descriptions to the meta file as well so currently looks like this:
```
Type: convolution
Name: Conv_MobilenetV1_MobilenetV1_Conv2d_0_Conv2D__2
[0] Dest:      data0002.bin    i8[S:0.2500 O:0][-32.000,31.750]<1 x 112 x 112 x 32>
[1] Src:       data0003.bin    i8[S:0.0156 O:0][-2.000,1.984]<1 x 224 x 224 x 3>
[2] Filter:    data0004.bin    i8[S:0.0312 O:0][-4.000,3.969]<32 x 3 x 3 x 3>
[3] Bias:      data0005.bin    i32[S:0.0005 O:0][-1048576.000,1048576.000]<32>
```
Please have a look and let's have this reviewed.
Thanks!

vuzelac-cadence:
@mciprian13 , I have no complaints - just wanted the tensor type in the dump header in any format and that's it.
As for minibatch, at the moment I only suggest throwing an error so users don't get surprised.

mciprian13:
@vuzelac-cadence 
Regarding the minibatch I just want to make sure I'm doing the right thing: should I allow the debug-instrumentation only when the parameter `miniBatch` from `ExecutorCoreHelperFunctions.cpp` equals 0 (meaning that batch splitting is deactivated)? Where exactly in the code should I make this check and whether externalizing the symbol `instrumentDebug` from `IROptimizer.cpp` doesn't seem hackish?
Thanks!

vuzelac-cadence:
` should I allow the debug-instrumentation only when the parameter miniBatch from ExecutorCoreHelperFunctions.cpp equals 0 `

That was my idea, but could quickly verify if I am correct. The way I see it dumping to file should not work nicely when minibatching is enabled.

mciprian13:
@opti-mix @vuzelac-cadence I added the verification you requested.
Can we have this reviewed?
Thanks!

tools/loader/ExecutorCore.cpp
```diff
@@ -222,6 +222,13 @@ int Executor::executeNetwork(int argc, char **argv) {
       << "singleBatchRepeatedMode is not compatible with "
          "streamInputFilenamesMode";
 
+  // When the mini-batch mode is enabled do not allow debug instrumentation.
+  extern llvm::cl::opt<bool> instrumentDebug;
```
opti-mix, line 226: This is a bit unusual place for an `extern` declaration. Also, we have some CommandLine.h files for many libs/Backends/. Maybe we should have the same thing for the Optimizer, which would allow us to include such a file instead of sprinkling `extern` for different flags here and there.

lib/Optimizer/IROptimizer/IROptimizer.cpp
```diff
@@ -30,14 +30,38 @@
...
 
+#include <algorithm>
+#include <fstream>
 #include <unordered_map>
 #include <unordered_set>
 
 #define DEBUG_TYPE "ir-optimizer"
 
 namespace {
+static llvm::cl::opt<std::string> instrumentDebugDir(
```
opti-mix, line 45: Nit: Looks like IROptimizer.cpp contains a lot of instrumentation and serialization logic now. I almost feel like it would be nice to move into a separate file like `IRInstrumentation.cpp`

docs/AOT.md
```diff
@@ -468,6 +468,28 @@ option like in this:
...
+  (one tensor per file). Each file will contain ONLY the tensor data.
+  - `rawtxt` The operands (tensors) are dumped in text format in separate files
+  (one tensor per file). Each file will contain ONLY the tensor data.
+  The names of the dump files have a simple format `data[idx].bin` or `data[idx].txt`
+  but a separate meta file `debug.info` is dumped at compile-time which makes the
+  association between each binary file and the operand of the IR instruction to which
+  it belongs. All the dump files and the separate meta file `debug.info` are dumped in
+  a folder `debug` relative to the current directory at compile-time (if it does not
+  exist it is created automatically). You can choose a different directory with the
+  option `-instrument-debug-dir=<dir>`.
```
opti-mix, line 492: BTW, do we have an example of a textual format produced by these options anywhere in this doc?

include/glow/Base/TensorSerialization.h
```diff
@@ -24,35 +24,39 @@
...
-/// other meta information (data type and shape). The binary representation
-/// of data is guaranteed to preserve data precision (bit exactness) upon
-/// round-trips (dump/load).
-void dumpToRawBinaryFile(Tensor &tensor, llvm::StringRef filename);
+/// Dump the content of \p tensor to a binary file \p filename. You can choose
+/// to dump also the tensor type in the file with the flag \p dumpType. The
+/// binary representation of data is guaranteed to preserve data precision
+/// (bit exactness) upon round-trips (dump/load).
+void dumpTensorToBinaryFile(Tensor &tensor, llvm::StringRef filename,
+                            bool dumpType = true);
```
opti-mix, line 32: Nit: I foresee getting more and more parameters passes to serialization functions in the future and becoming a `SerializationOptions` or `SerializationOptions` ;-) Maybe we should go for it right away and have there only a single field called dumpType for now?

mciprian13: Sound reasonable but we have two types of functions: dumping and loading (one has a `dumpType` flag and the other has a `loadType` flag). Should we have two different types of structures (with `dumpType` and `loadType` flags) or just one with (say) a `withType` flag??

opti-mix: > Should we have two different types of structures (with dumpType and loadType flags) or just one with (say) a withType flag??

I don't have any strong opinions here. I personally would probably go with `withType` or `isTyped` flag. But I leave up to you to make this call.

tools/loader/ExecutorCore.cpp
mciprian13: Created `CommandLine.h` and `ComamndLine.cpp` files. Done.

lib/Optimizer/IROptimizer/IROptimizer.cpp
mciprian13: Done.

docs/AOT.md
mciprian13: No, therefore I put some simple examples for `txt` and `rawtxt` formats.

include/glow/Base/TensorSerialization.h
mciprian13: Created a single `TensorSerializationOptions` structure with the `withType` flag used for both dump/load functions.

mciprian13:
@opti-mix Have another look and if fine by you then let's have this landed.
Thanks!

lib/Optimizer/IROptimizer/IROptimizer.cpp
```diff
@@ -30,21 +30,17 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/FileSystem.h"
```
opti-mix, line 33: Do we still need these IO related headers in this file after you moved all the instrumentation logic into a separate file?

docs/AOT.md
```diff
@@ -468,6 +468,37 @@ option like in this:
...
+  - `rawbin` The operands (tensors) are dumped in binary format in separate files
+  (one tensor per file). Each file will contain ONLY the tensor data.
+  - `rawtxt` The operands (tensors) are dumped in text format in separate files
+  (one tensor per file). Each file will contain ONLY the tensor data. One such file
+  might look like this:
+    ```
+    1.1, 2.2, 3.3, 4.4, 5.5, 6.6,
+    ```
+  The names of the dump files have a simple format `data[idx].bin` or `data[idx].txt`
+  but a separate meta file `debug.info` is dumped at compile-time which makes the
```
opti-mix, line 496: Could you provide in the doc an example of how `debug.info` looks like?

lib/Optimizer/IROptimizer/IROptimizer.cpp
mciprian13: I cleaned up the unused header includes.

docs/AOT.md
mciprian13: Done.

opti-mix:
@mciprian13 Your PR has some linking issues when we build using the FB internal buck build system. Could you move the newly introduced CommandLine.cpp into IROptimizerPipeline sub-directory? It would solve the issue. And when you do this, could you also move the LLVM command-line flags defined in IRFunctonPassPipeline.cpp into CommandLine.cpp/.h as well?

Another really tiny nit: Could you replace current implicit invocations of constructors of TensorSerializationOptions which look like this `{/*withType*/ false}` into more explicit invocations of this constructor?

Thanks!

opti-mix:
@mciprian13 Just a quick ping in case you missed my last comment. Please let me know if you'd be able to do those minor changes and push the new version.

mciprian13:
@opti-mix Hopefully this time is good 😄
Side question: do you think you could have a look on #4350 because that PR is not progressing at all and I needed that like yesterday. Or at least start the review or something.
Thanks!

opti-mix:
> do you think you could have a look on #4350 because that PR is not progressing at all and I needed that like yesterday. 

I think @jfix71 would be a better candidate to review it than I. 

@jfix71 Could you have a look at #4350?

opti-mix:
@mciprian13 Thanks a lot for the updates. They fixed the last issue I reported. But now one more issue was detected on our internal CI. Please don't change BackendExecTest as you did. It breaks some tests on our side. Could you introduce another type like BackendExecStatelessTest and use it just for your new tests?

mciprian13:
@opti-mix I created a new BackendTest type `BackendExecStatelessTest` as you suggested.
Hope this time is ok.
Another side question: do you think you would be interested in doing #4263? I would do it myself but I`m not sure I am that skilled in LLVM IR.

mciprian13:
@opti-mix This PR seems to be very stubborn 😄 What problem is there now?

opti-mix:
@mciprian13 

> This PR seems to be very stubborn 😄 What problem is there now?

No problems. It is merged finally! :-) You've done a great job!

mciprian13:
@opti-mix Thanks for reviewing it!

Issue #3937, opened 2019-12-20
fprintf use in libijit on windows


ayermolo:
This is just in case someone runs in to this and searches here.
We ran in to an issue on windows when using fprintf in layer implemented in libijit. It ranged from sporadic to consistent depending on windows OS version. The root cause is still unknown, but a fix that worked was proposed by @vuzelac-cadence is to pass pointers to wrapper static functions around fopen/fclose/fprintf in struct, and use those internally.

Issue #2991, opened 2019-05-29
Problems accessing global variables in the host-process from the JITted code


opti-mix:
Let's say there is a global int variable `global_counter` defined somewhere in the host-process using Glow JIT. And something inside the `libjit` needs to access this variable.

The most natural way to do it would be to declare this variable as an external variable. To avoid any issues with the C++ name mangling, one would probably declare and use it like this:
```c++
extern "C" int global_counter;
void libjit_use_global_counter() {
   global_counter++;
}
```
And then define this variable in the host-process e.g. like this:
```c++
extern "C" int libjit_tot_counter = 0;
```

But if you try to build and run something like this using the `-cpu` backend you will discover that the address of `libjit_tot_counter` cannot be found and thus you cannot run the JITted code.

opti-mix:
The code responsible for resolving the addresses can be found in `GlowJIT.cpp`.
It looks like this:
```c++
resolver_(createLegacyLookupResolver(
          ES_,
          [this](const std::string &Name) -> JITSymbol {
            if (auto Sym = compileLayer_.findSymbol(Name, false))
              return Sym;
            else if (auto Err = Sym.takeError())
              return std::move(Err);
            if (auto SymAddr =
                    RTDyldMemoryManager::getSymbolAddressInProcess(Name))
              return JITSymbol(SymAddr, JITSymbolFlags::Exported);
            return nullptr;
          },
          [](Error Err) { cantFail(std::move(Err), "lookupFlags failed"); }))
```

As we can see, it first tries to ask the `compileLayer_` to resolve the symbol. But if it fails, it invokes `RTDyldMemoryManager::getSymbolAddressInProces` to resolve the symbol. And this call should be able to find `global_counter` because after all it is defined in the host process, right? Well, it turns out that just defining this variable in the host process is not enough, because by default all those global symbols are local, i.e. non-external, and `getSymbolAddressInProcess` looks for external symbols only. Thus to help `getSymbolAddressInProcess` find the symbol in question, one needs to define it with non-local visibility, i.e. essentially export it. This can be done e.g. like this:
```c++
extern "C" __attribute__ ((visibility("default"))) int libjit_tot_counter = 0;
```
If you try again now, everything works just fine.

**Note:** To check if a global variable is exported, you can e.g. use something like `nm your_executable | less libjit_tot_counter`. Without the visibility attribute, the output would be:
```
0000000102a41898 s _libjit_tot_counter
```
[...]

PR #4395, opened 2020-04-07
Add file printer for JIT backend use.

vuzelac-cadence:
Introduce host based file printer for backend JIT use.
It allows for concurrent writes to the same files and also avoids running into problems on windows with NTDLL failing on fprintf from JIT code.

Related to #4392, #3845 

Use like this in libjit.cpp for example:
```
#include <glow/LLVMIRCodeGen/JITFilePrinter.h>
extern "C" {
extern JITFileWriterIF _jitFileWriter;
}
```
Actual calls:
`_jitFileWriter.fopen();`

Testing: #4392 needs to provide testing, this MR will include replace libjit.cpp I/O calls, and the tests should pass.

### DIFF
--- a/include/glow/LLVMIRCodeGen/JITFilePrinter.h
+++ b/include/glow/LLVMIRCodeGen/JITFilePrinter.h
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) Glow Contributors. See CONTRIBUTORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This file contains the print functionality the JIT code can use. The backend
+// can call function to emit a global var that points to the host side struct
+// containing fopen/fprintf/fclose pointers, JIT code can call instead of it's
+// own fopen/fprintf/fclose.
+
+#include <condition_variable>
+#include <cstdarg>
+#include <unordered_map>
+
+/// Structure implements file I/O related functions. Currently,
+/// it just maps C-lib fopen/fprintf/fclose onto it's members.
+struct JITFileWriterIF {
+  FILE *(*fopen)(const char *, const char *);
+  int (*fprintf)(FILE *, const char *, ...);
+  int (*fclose)(FILE *);
+};
+
+struct JITFileWriter {
+  JITFileWriterIF IF;
+  /// Open a file. Operation, arguments and return value match standard fopen
+  /// function.
+  static FILE *JITfopen(const char *, const char *);
+  /// Writes the C string pointed by format to the stream. Operation, arguments
+  /// and return value match standard fprintf function.
+  static int JITfprintf(FILE *, const char *, ...);
+  /// Close a file. Operation, arguments and return value match standard fclose
+  /// function.
+  static int JITfclose(FILE *);
+  JITFileWriter() {
+    IF.fopen = &JITfopen;
+    IF.fprintf = &JITfprintf;
+    IF.fclose = &JITfclose;
+  }
+  std::unordered_map<std::string, bool> fileLock_;
+  std::unordered_map<FILE *, std::string> fileMap_;
+  std::condition_variable readyCV;
+  std::mutex mtx_;
+};
--- a/include/glow/LLVMIRCodeGen/LLVMIRGen.h
+++ b/include/glow/LLVMIRCodeGen/LLVMIRGen.h
@@ -371,6 +371,8 @@ public:
   /// Generates LLVM IR that materializes the string literal \p str.
   virtual llvm::Value *emitStringConst(llvm::IRBuilder<> &builder,
                                        llvm::StringRef str);
+  /// Emit symbols to JIT to allow it to use host side file printing.
+  void generateJITFileWriter();
   /// Register \p val as an argument that should not be specialized.
   virtual void markArgAsUnspecialized(llvm::Value *val);
   /// \returns bit-width of the target size_t.
--- a/lib/LLVMIRCodeGen/JITFilePrinter.cpp
+++ b/lib/LLVMIRCodeGen/JITFilePrinter.cpp
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) Glow Contributors. See CONTRIBUTORS file.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This file contains the print functionality the JIT code can use. The backend
+// can call function to emit a global var that points to the host side struct
+// containing fopen/fprintf/fclose pointers, JIT code can call instead of it's
+// own fopen/fprintf/fclose.
+
+#include "glow/LLVMIRCodeGen/JITFilePrinter.h"
+#include "glow/LLVMIRCodeGen/LLVMIRGen.h"
+
+#include "llvm/Support/DynamicLibrary.h"
+
+using namespace glow;
+
+#if defined _MSC_VER
+#define EXTERN __declspec(dllexport)
+#else
+#define EXTERN
+#endif
+
+static JITFileWriter jitfileWriter;
+
+int JITFileWriter::JITfprintf(FILE *f, const char *format, ...) {
+  va_list argptr;
+  va_start(argptr, format);
+  return vfprintf(f, format, argptr);
+}
+
+FILE *JITFileWriter::JITfopen(const char *path, const char *mode) {
+  // File access is locked using per file name lock. This lock is a simple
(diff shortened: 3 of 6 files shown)


opti-mix:
@vuzelac-cadence I wonder if the same functionality could have been sold using simpler methods, e.g. something described in #2991. One could have exposed in the host process custom versions of certain functions or expose some new APIs with the required functionality. WDYT?

include/glow/LLVMIRCodeGen/JITFilePrinter.h
```diff
@@ -0,0 +1,45 @@
...
+  JITFileWriter() {
+    IF.fopen = &JITfopen;
+    IF.fprintf = &JITfprintf;
+    IF.fclose = &JITfclose;
+  }
+  std::unordered_map<std::string, bool> fileLock_;
+  std::unordered_map<FILE *, std::string> fileMap_;
+  std::condition_variable readyCV;
+  std::mutex mtx_;
+};
```
opti-mix, line 45: New line is missing.

include/glow/LLVMIRCodeGen/JITFilePrinter.h
```diff
@@ -0,0 +1,45 @@
...
+// This file contains the print functionality the JIT code can use. The backend
+// can call function to emit a global var that points to the host side struct
+// containing fopen/fprintf/fclose pointers, JIT code can call instead of it's
+// own fopen/fprintf/fclose.
+
+#include <condition_variable>
+#include <cstdarg>
+#include <unordered_map>
+
+struct JITFileWriterIF {
```
opti-mix, line 25: Could you add Doxygen comments for the types and their members?

include/glow/LLVMIRCodeGen/LLVMIRGen.h
```diff
@@ -371,6 +371,8 @@ class LLVMIRGen {
   /// Generates LLVM IR that materializes the string literal \p str.
   virtual llvm::Value *emitStringConst(llvm::IRBuilder<> &builder,
                                        llvm::StringRef str);
+  /// Generate pointer to the host side file printer for JIT code to use. \p F.
```
opti-mix, line 374: The comment looks a bit strange at the end. I mean `\p F`. BTW, how is F used in the function?

lib/LLVMIRCodeGen/JITFilePrinter.cpp
```diff
@@ -0,0 +1,73 @@
...
+int JITFileWriter::JITfclose(FILE *f) {
+  std::lock_guard<std::mutex> lck(jitfileWriter.mtx_);
+  std::string file = jitfileWriter.fileMap_[f];
+  int ret = fclose(f);
+  jitfileWriter.fileLock_[file] = false;
+  jitfileWriter.readyCV.notify_all();
+  return ret;
+}
+
+void LLVMIRGen::generateJITFileWriter(llvm::Function *F) {
```
opti-mix, line 63: Please add comments explaining what this function does and how it is implemented.

lib/LLVMIRCodeGen/JITFilePrinter.cpp
```diff
@@ -0,0 +1,73 @@
...
+  llvm::GlobalVariable *writer = llmodule_->getNamedGlobal("_jitFileWriter");
+  if (!writer) {
+    return;
+  }
+  auto *ptrType = llmodule_->getTypeByName("struct.JITFileWriterIF");
+  llvm::Constant *writerV = builder_->getInt64((int64_t)&jitfileWriter.IF);
+  llvm::Constant *writerPtr = llvm::ConstantExpr::getIntToPtr(
+      writerV, llvm::PointerType::getUnqual(ptrType));
+  writer->setInitializer(writerPtr);
+}
```
opti-mix, line 73: New line missing.

lib/LLVMIRCodeGen/JITFilePrinter.cpp
```diff
@@ -0,0 +1,73 @@
...
+      lck, [&path] { return !jitfileWriter.fileLock_[std::string(path)]; });
+  jitfileWriter.fileLock_[std::string(path)] = true;
+  FILE *f = fopen(path, mode);
+  jitfileWriter.fileMap_[f] = std::string(path);
+  return f;
+}
+
+int JITFileWriter::JITfclose(FILE *f) {
+  std::lock_guard<std::mutex> lck(jitfileWriter.mtx_);
+  std::string file = jitfileWriter.fileMap_[f];
```
opti-mix, line 56: Don't you need to free the entry in fileMap_ afterwards?

Also don't you need to check if `f` is even in `fileMap_`?

vuzelac-cadence:
@opti-mix thanks for a quick review. I've just send out the commit we had internally to supplement my comments in  #4392 . Will cleanup.

vuzelac-cadence:
@opti-mix,  I was aware of 1st two approaches you mentioned in #2991. Now, using 
llvm::sys::DynamicLibrary::AddSymbol("global_counter", &counter);
certainly looks very sleek and I prefer it much more. I'll see to add my printer using this one.

ayermolo:
@vuzelac-cadence thanks for up streaming. :D
Related issue:
https://github.com/pytorch/glow/issues/3937

vuzelac-cadence:
@opti-mix , regarding` llvm::sys::AddSymbol`

My idea was not to export the `_jitFileWriter` but a pointer to it, thinking that will allow me to continue using the `weak `attribute on the JIT side in order to be able to check if Host has provided the printer:

const JITFileWriterIF *_jitFileWriter __attribute__((weak)) = nullptr;

But defining the symbol  using `AddSymbol` will not override the above symbol anymore.
Now I can explore more, but was thinking - who cares ?  JIT can still use use whatever printer it wants (through some other selection mechanism). We actually don't need this selection . I don't see the need for selection in #4392 either. Do you anticipate a need for such weak symbol definition ?.

lib/LLVMIRCodeGen/JITFilePrinter.cpp
vuzelac-cadence: Fixed both

lib/LLVMIRCodeGen/LLVMBackend.cpp
```diff
@@ -99,6 +99,8 @@ void LLVMBackend::emitJitMain(LLVMIRGen &irgen) const {
   irgen.createCall(builder, entryF, initFunctionCallArgs);
   // Terminate the function.
   builder.CreateRetVoid();
+  // Emit JIT file printer.
+  irgen.generateJITFileWriter(func);
```
opti-mix, line 103: LLVMBackend is used sometimes for AOT compilation, not just for a JIT. Thus, registering the JITFileWriter unconditionally could be not such a good idea in general. But I don't have a strong opinion here.

lib/LLVMIRCodeGen/JITFilePrinter.cpp
```diff
@@ -0,0 +1,73 @@
...
+  jitfileWriter.fileMap_.erase(f);
+  jitfileWriter.fileLock_[file] = false;
+  jitfileWriter.readyCV.notify_all();
+  return ret;
+}
+
+/// Expose host side file printer for JIT code to use.
+/// The function simply exposes the writer object as a global symbol
+/// that will be avaiable for JIT to use.
+void LLVMIRGen::generateJITFileWriter(llvm::Function *F) {
```
opti-mix, line 71: It seems like you don't use the `F` parameter at all. What is the purpose of `F` in a general case?

lib/LLVMIRCodeGen/LLVMBackend.cpp
vuzelac-cadence: I think of this "registering" as simply "providing". Backend is free not to use it. 
Note that we need CPU backend provided dumps as much as Xtensa dumps, throughout our compile and use flow. Therefore I moved registering here. We originally had a backend option to enabled this, however, we settled with the current approach merely due to it being one-liner, which allows for easier rebase until (and if ever) we upstream.

vuzelac-cadence:
@opti-mix I agree `HostEnvironment `is the ideal solution. But, we were trying to solve problem on hands and the approach looked not too patchy, if you think of it as a part of the JIT tensor dump infrastructure and nothing more. 
We don't have bandwidth to provide better implementation. We are ok if this is not upstreamed - the original intent of this MR is to backup the issues I brought up in #4392 . As is, #4392 will be buggy.
Note that we do have many host provided structures and functions we emit into Xtensa JIT, from Glow Xtensa Backend, which matches your expectation that backend should take care of this. However, this one is special as we need it for CPU Backend as well.

opti-mix:
@vuzelac-cadence Yeah, I see your point. And I really don't have a very strong opinion here. 

@jfix71 What do you think? Is this feature in its current form worth being upstreamed or do you think it will always stay very backend-specific or use-case specific?

jfix71:
Also no strong feelings. Agreed it's not the most general/ideal solution. But it seems potentially useful and pretty well self-contained, so I'd be fine with upstreaming it.

lib/LLVMIRCodeGen/JITFilePrinter.cpp
opti-mix: Could you address this comment?

opti-mix:
@vuzelac-cadence OK, let's land it. Could you please address the remaining comments and formatting issues?

vuzelac-cadence:
@opti-mix , @jfix71 , format check fails on these files:

	modified:   torch_glow/src/PyTorchModelLoader.cpp
	modified:   torch_glow/tests/nodes/rsub_test.py

I don't have them in my commit. I see they are indeed not formatted.

jfix71:
@vuzelac-cadence Yes, they are broken in master. Will get it fixed soon.

vuzelac-cadence:
@jfix71 , rebased to re-trigger the build. This is ready now

lib/LLVMIRCodeGen/JITFilePrinter.cpp
```diff
@@ -0,0 +1,73 @@
...
+  jitfileWriter.fileMap_[f] = std::string(path);
+  return f;
+}
+
+int JITFileWriter::JITfclose(FILE *f) {
+  std::lock_guard<std::mutex> lck(jitfileWriter.mtx_);
+  if (!jitfileWriter.fileMap_.count(f)) {
+    return -1;
+  }
+  std::string file = jitfileWriter.fileMap_[f];
```
jfix71, line 60: nit: `const std::string &file`

PR #4397, opened 2020-04-07
Update NNPI backend to 0.5.1.8

omromano:
Summary:
Update of NNPI Backend to v0.5.1.8.


jsubag:
@arunm-git , @jfix71 , please take a look.

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -1184,3 +1185,82 @@ Expected<bool> NNPIBackend::transformPostLowering(
...
+  }
+
+  // Need post order to ensure p2p dest resources are created before their
+  // source (since source will handle the copy command).
+  std::unordered_set<const runtime::DAGNode *> visited;
+  std::vector<const runtime::DAGNode *> postOrder;
+  traversePostOrder(root, visited, postOrder);
+  runtime::PlaceholderUsageMap phUsage;
+  // Collect placeholders usage count.
+  for (auto cb : bindings) {
```
jfix71, line 1223: nit: `const auto &cb`

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -1184,3 +1185,82 @@ Expected<bool> NNPIBackend::transformPostLowering(
...
+    runtime::NNPIDeviceManager *nnpiDM =
+        dynamic_cast<runtime::NNPIDeviceManager *>(cb.device);
+    LOG_IF_NOT_RETURN_LLVMERROR(nnpiDM, "Invalid device manager");
+    nnpiDM->addPlaceholderUsageCount(cb.networkName, phUsage);
+  }
+
+  for (auto usage : phUsage) {
```
jfix71, line 1230: nit: `const auto &usage`

lib/Backends/NNPI/NNPIResource.cpp
```diff
@@ -192,8 +307,36 @@ bool NNPIResource::init(const NNPIObjectName name,
...
   case ResourceUsage::StaticInputResource:
-    // Fallthrough.
+    // Fallthrough - the device resource doesn't need to be updated before
+    // inference.
+  case ResourceUsage::DRTInput:
+    // Fallthrough - the device resource doesn't need to be updated before
+    // inference.
+  case ResourceUsage::DRTOutput:
+    // Fallthrough - the device resource doesn't need to be copied out after
+    // inference.
```
jfix71, line 339: nit: can we add explicit breaks to these? Feels like it could be error prone if someone adds something to a case here in the future but doesn't realize others will fall in (even if it's something simple like logging, as it doesn't seem there's much functionality change that would happen here).

lib/Backends/NNPI/tests/NNPIOperatorTest.cpp
```diff
@@ -247,12 +245,10 @@ struct BlacklistInitializer {
              TestBlacklist::AnyDeviceAnyEngine},
             {"convTransposeConvolutionCompareSimpleK5S1P2I3/0",
              TestBlacklist::AnyDeviceAnyEngine},
-            {"ChannelwiseQuantizedGroupConvolution/0",
-             TestBlacklist::AnyDeviceAnyEngine},
             {"ChannelwiseQuantizedGroupConvolution3D/0",
              TestBlacklist::AnyDeviceAnyEngine},
             {"ChannelwiseQuantizedGroupConvolutionNonZero/0",
-             TestBlacklist::AnyDeviceAnyEngine},
+             TestBlacklist::AnyDeviceSWEngine},
```
jfix71, line 251: Seems a bit weird -- so ICEREF is broken for this test but device isn't?

lib/Backends/NNPI/InferenceContext.cpp
```diff
@@ -171,36 +168,74 @@ bool InferenceContext::init(
         staticPlaceholderMap->insert({PH, inputResources_.back()});
       }
     } else {
-      // Regular input resource - create it here.
+      // Dynamic input resource - create it here.
+      NNPIResource::ResourceUsage usage = NNPIResource::ResourceUsage::None;
+      if (!phUsage || (phUsage->at(name).numWriters == 0)) {
+        usage = NNPIResource::ResourceUsage::InputResource; // Net input
+      } else { // Some other context is writing to this placeholder --> P2P/DRT
+        switch (phUsage->at(name).devices.size()) {
```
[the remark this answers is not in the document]
jsubag: IIRC p2p only support 1-1 connections - i.e. no broadcast.
So for the case you mentioned you'll need to add another PH on A which is a copy of the existing one and connect one to B1 and the other to B2.

lib/Backends/NNPI/InferenceContext.cpp
```diff
@@ -287,10 +408,10 @@ void InferenceContext::execute(RunIdentifierTy runId,
   // outputResources_.
   if (netInputPlaceholders_.empty()) {
     for (const auto &in : inputResources_) {
-      if (in->GetUsage() == NNPIResource::ResourceUsage::StaticInputResource) {
+      if (in->getUsage() == NNPIResource::ResourceUsage::StaticInputResource) {
         continue;
       }
-      auto *placeholder = bindings.getPlaceholderByName(in->GetName());
+      auto *placeholder = bindings.getPlaceholderByName(in->getName());
```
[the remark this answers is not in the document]
jsubag: This happens only on the first time and is cached in netInputPlaceholders_ (see the condition block its in)

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -535,120 +540,149 @@ static void setupBasicParallelizationConfigs(
...
-        numChunks[FC] = numParallelChunks;
-        continue;
-      }
-      size_t M = FC->getInput().dims()[0];
-      if (M >= 256) {
-        parOpts[FC] = ParallelTransformKind::Data;
-        numChunks[FC] = numParallelChunks;
-        continue;
-      }
+  // Find all FC layers to split
```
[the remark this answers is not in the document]
jsubag: This looks like a merge error.
@omromano  please revert this.

lib/Backends/NNPI/NNPICompiledFunction.cpp
```diff
@@ -235,18 +253,32 @@ Error NNPICompiledFunction::compile(Function *F, const BackendOptions &opts) {
     return error;
   }
 
-  RETURN_IF_ERR(setupCompilationHints(F, newOpts.backendSpecificNodeInfo));
```
jsubag, line 238: This line looks like it was removed by mistake. (we should really stop merging every 5 min :) )
@jfix71 just keep this line as in master just above the "// Collect input/output names." comment.

lib/Backends/NNPI/NNPI.cpp
```diff
@@ -84,22 +88,22 @@ NNPIBackendOptions NNPIBackend::backendOptions_;
 NNPIAdapterContainer NNPIBackend::adapter_;
 
 unsigned NNPIBackend::numDevices() {
```
arunm-git, line 90: @omromano @jsubag 
This change to numDevices reverts a change I made to your last PR for 5.1.4.
numDevices as it is coded will leak memory if run on a host that doesn't have a card attached (adapterCreate will allocate, adapterGetInfo will fail and we return without calling adapterDestroy). numDevices() on github master fixes this - please update your internal sources for that change so we don't try to revert it on the next PR.

lib/Backends/NNPI/NNPIDeviceManager.cpp
```diff
@@ -63,8 +64,8 @@ NNPIDeviceManager::NNPIDeviceManager(
     std::shared_ptr<NNPIDeviceOptions> deviceOptions, NNPIAdapter adapter,
     unsigned numInferenceWorkers)
     : DeviceManager(config), numWorkersPerFunction_(numInferenceWorkers),
-      deviceId_(config_.deviceID), adapter_(adapter),
-      device_(NNPI_INVALID_NNPIHANDLE), deviceOptions_(deviceOptions) {
+      deviceId_(config_.deviceID), device_(NNPI_INVALID_NNPIHANDLE),
+      deviceOptions_(deviceOptions), adapter_(adapter) {
```
arunm-git, line 68: @omromano @jsubag 
This ordering was also an intentional change I made to your last PR. Without it there is a build warning/error due to incorrect initialization order. Same comment as above - please pull what's in github master into your tree so we don't try to revert it on the next PR. Thanks!

omromano: @arunm-git 
Sorry we missed that. Will make sure it updated in our code (and not reverted in the next PR).

